PR #20376, opened 2019-05-21
Cluster templates

prachidamle:
rancher/types API PR: https://github.com/rancher/types/pull/799

- CRUD ClusterTemplate and ClusterTemplateRevisions
- Create Cluster From Template
- Changes after design refactored
- Integration Tests added
- Update Cluster From Template


pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		if obj.Spec.Disabled {
+			return nil
+		}
+		toUpdate := obj.DeepCopy()
+		toUpdate.Spec.Disabled = true
+
+		_, err = w.ClusterTemplates.Update(toUpdate)
+		if err != nil {
+			return err
```
alena1108: Do we usually return a bare error in handlers, or wrap it up in apiContext.WriteResponse?

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -0,0 +1,124 @@
...
+		apiContext.WriteResponse(http.StatusNoContent, map[string]interface{}{})
+		return nil
+	default:
+		return fmt.Errorf("bad action for clusterTemplate %v", actionName)
+	}
+}
+func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHandler) error {
+	switch apiContext.Link {
+	case "revisions":
+		logrus.Infof("Revision link invoked")
```
alena1108: Assuming it was done for debug purpose, we should reduce the log level

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -0,0 +1,124 @@
...
+	return nil
+}
+
+func (w Wrapper) RevisionActionHandler(actionName string, action *types.Action, apiContext *types.APIContext) error {
+	var templateRevision client.ClusterTemplateRevision
+	if err := access.ByID(apiContext, &managementschema.Version, client.ClusterTemplateType, apiContext.ID, &templateRevision); err != nil {
+		return err
+	}
+	switch actionName {
+	case "disable":
```
alena1108: this note is for my own reference, and for other reviewers: we set Disable parameter via Action handler as opposed to regular update via POST because the intent is to disable all other updates for the template revision.

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -0,0 +1,124 @@
...
+	resource.Links["revisions"] = apiContext.URLBuilder.Link("revisions", resource)
+}
+
+func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiContext *types.APIContext) error {
+	var template client.ClusterTemplate
+	if err := access.ByID(apiContext, &managementschema.Version, client.ClusterTemplateType, apiContext.ID, &template); err != nil {
+		return err
+	}
+	switch actionName {
+	case "disable":
```
alena1108: We can set disable attribute on a template via regular update.

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (w Wrapper) RevisionActionHandler(actionName string, action *types.Action, apiContext *types.APIContext) error {
+	var templateRevision client.ClusterTemplateRevision
+	if err := access.ByID(apiContext, &managementschema.Version, client.ClusterTemplateType, apiContext.ID, &templateRevision); err != nil {
+		return err
+	}
+	switch actionName {
+	case "disable":
+		obj, err := w.ClusterTemplateRevisions.GetNamespaced(namespace.GlobalNamespace, templateRevision.Name, v1.GetOptions{})
```
alena1108: After talking to @prachidamle offline, found out that the intent of the action to turn on the flag was to prevent updates on all other fields. We can achieve similar functionality by setting noUpdate norman flag in rancher/types.

pkg/api/store/cluster/cluster_store.go
```diff
@@ -254,6 +352,29 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
...
+			return nil, err
+		}
+		clusterUpdate, err := loadDataFromTemplate(clusterTemplateRevision, data)
+		if err != nil {
+			return nil, err
+		}
+		updatedTemplateID := convert.ToString(clusterUpdate[managementv3.ClusterSpecFieldClusterTemplateID])
+		templateID := convert.ToString(existingCluster[managementv3.ClusterSpecFieldClusterTemplateID])
+		if !strings.EqualFold(updatedTemplateID, templateID) {
+			return nil, fmt.Errorf("cannot Update Cluster, Cluster cannot be changed to a new template")
```
alena1108: Nit: we should remove caps from the words in the middle of the sentence. Can you also check if error messages in Rancher API errors should start with cap

pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+//sync is called periodically and on real updates
+func (n *CTemplateRevController) sync(key string, obj *v3.ClusterTemplateRevision) (runtime.Object, error) {
+	if obj == nil || obj.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	//check if this revision has isDefault = true, if yes update the parent Template
```
alena1108: Do we ever validate that there is only one template revision with default attribute set to true? On the API level

pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,118 @@
...
+	if obj.Spec.IsDefault {
+		err := n.setDefaultOnTemplate(obj)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	//set OwnerRef on the revision
+	if len(obj.ObjectMeta.OwnerReferences) == 0 {
```
alena1108: It would be better if the owner reference is set on API level.

pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return fmt.Errorf("Error %v finding ClusterTemplate by name %v", err, obj.Spec.ClusterTemplateName)
+	}
+
+	if !strings.EqualFold(template.Spec.DefaultRevisionName, obj.Name) {
+		toUpdate := template.DeepCopy()
+		toUpdate.Spec.DefaultRevisionName = namespace.GlobalNamespace + ":" + obj.Name
+		_, err = n.clusterTemplates.Update(toUpdate)
+		if err != nil {
+			return fmt.Errorf("Error %v updating Default Revision on ClusterTemplate %v", err, obj.Spec.ClusterTemplateName)
```
alena1108: Nit: we should fix the error message. Either use camel case, or split words.

pkg/api/customization/clustertemplate/cluster_template.go
prachidamle: @alena1108 on second thoughts, maybe we should allow updates on ClusterTemplateRevision fields, unless it is being used by some clusters?

pkg/controllers/management/clustertemplate/revision_controller.go
prachidamle: @alena1108 I feel we should not have the isDefault flag on templateRevision, reason:
1)Right, we need to validate on API level that only one revision has isDefault true. This check will need to be done for both Create and Update templateRevision APIs
2) Also, for  Update templateRevision API, we are going to error out if some cluster is using this templateRevision.
3) Now when Update templateRevision API is called, to satisfy #1 and #2 above, we will have to check if isDefault is getting updated and do check #1. But for #2, we will have to scan the data and find if anything else is getting updated - and this is very inefficient.
4) Moreover, we can never guard against concurrent API requests in HA mode.

So I think we should not have this flag on templateRevision. Just having ClusterTemplate.DefaultRevision reference should be enough. Does this sound ok?

alena1108: I agree with your your points @prachidamle; the resolution sounds reasonable. We'll only have to run it with the UI team. TemplateId will always be set on a cluster, so I guess it should be easy to fetch the default templateRevision if needed.

alena1108:
LGTM

PR #20378, opened 2019-05-22
RBAC handler should handle sa kind PRTB

orangedeng:
**Problem:**
The project monitoring can't fetch metrics from the namespace moved
from the other project

**Solution:**
The RBAC handler should handle service account type PRTB
too.

**Issues:**

https://github.com/rancher/rancher/issues/20459
https://github.com/rancher/rancher/issues/20390

### DIFF
--- a/pkg/controllers/user/rbac/handler_base.go
+++ b/pkg/controllers/user/rbac/handler_base.go
@@ -394,8 +394,10 @@ func getPRTBProjectAndSubjectKey(prtb *v3.ProjectRoleTemplateBinding) string {
 		name = prtb.UserPrincipalName
 	} else if prtb.GroupName != "" {
 		name = prtb.GroupName
-	} else {
+	} else if prtb.GroupPrincipalName != "" {
 		name = prtb.GroupPrincipalName
+	} else if prtb.ServiceAccount != "" {
+		name = prtb.ServiceAccount
 	}
 	return prtb.ProjectName + "." + name
 }
--- a/pkg/controllers/user/rbac/namespace_handler.go
+++ b/pkg/controllers/user/rbac/namespace_handler.go
@@ -166,7 +166,7 @@ func (n *nsLifecycle) ensurePRTBAddToNamespace(ns *v1.Namespace) (bool, error) {
 			return false, errors.Wrapf(err, "object %v is not valid project role template binding", prtb)
 		}
 
-		if prtb.UserName == "" && prtb.GroupPrincipalName == "" && prtb.GroupName == "" {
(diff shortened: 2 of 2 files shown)


jiaqiluo:
Here is the issue:  https://github.com/rancher/rancher/issues/20390

loganhz:
@gitlawr please review it

gitlawr:
LGTM

pkg/controllers/user/rbac/handler_base.go
```diff
@@ -394,8 +394,10 @@ func getPRTBProjectAndSubjectKey(prtb *v3.ProjectRoleTemplateBinding) string {
 		name = prtb.UserPrincipalName
 	} else if prtb.GroupName != "" {
 		name = prtb.GroupName
-	} else {
+	} else if prtb.GroupPrincipalName != "" {
 		name = prtb.GroupPrincipalName
+	} else if prtb.ServiceAccount != "" {
+		name = prtb.ServiceAccount
```
prachidamle: @orangedeng Why is this change needed only in case a namespace is moved from one project to another? In usual case when the namespace belongs to a project from start don't we use serviceaccount?

orangedeng: For now, we only support create service account subject kind of PRTB for project monitoring app. We create this kind of PRTB only in the controller.
The PRTB lifecycle handler already can handle this when creating a new one.  The only missing place is moving namespace case.

prachidamle: @orangedeng can you point me to the code that adds the PRTB in the moving namespace case? From the changes done in this PR, I am not able to see how only these fixes will fix the moving namespace usecase.

orangedeng: Here https://github.com/rancher/rancher/blob/master/pkg/controllers/user/rbac/prtb_handler.go#L55

prachidamle:
LGTM

pkg/controllers/user/rbac/handler_base.go
prachidamle: Discussed offline, and this is the explanation added by @orangedeng "The PRTB for service account will generate rolebindings in the namespaces of the project. Then the service account can access to the pod/endpoints(to get the pod ips) in those namespaces. 
When moving namespace to a new project, we should generate the new rolebinding in this namespace for the service account if there is a PRTB in that project. We already have the logic for the user/group kind of PRTB so we only need to do that to service account kind of PRTB too.
The service account kind of PRTB don’t have user or group attributes in it so the original logic of ns handler won’t do anything for it. This change is to let that ns handler to handle the service account kind of PRTB also."

alena1108:
https://github.com/rancher/rancher/issues/20470

PR #20396, opened 2019-05-22
Fix broken jail

dramich:
Problem:
Jailed did not handle multi-threaded

Solution:
Lock the jailer runs

### DIFF
--- a/package/jailer.sh
+++ b/package/jailer.sh
@@ -24,6 +24,10 @@ cp -r /lib64 /opt/jail/$NAME
 cp /etc/ssl/certs/ca-certificates.crt /opt/jail/$NAME/etc/ssl/certs
 cp /etc/resolv.conf /opt/jail/$NAME/etc/
 
+if [ -d /var/lib/rancher/management-state/bin ]; then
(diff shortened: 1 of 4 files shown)


package/jailer.sh
```diff
@@ -24,6 +24,10 @@ cp -r /lib64 /opt/jail/$NAME
 cp /etc/ssl/certs/ca-certificates.crt /opt/jail/$NAME/etc/ssl/certs
 cp /etc/resolv.conf /opt/jail/$NAME/etc/
 
+if [ -d /var/lib/rancher/management-state/bin ]; then
+    cp -r /var/lib/rancher/management-state/bin /opt/drivers/management-state
```
daxmc99: Why do we need this?

dramich: Upgrade path, `/var/lib/rancher/management-state/bin` is where node drivers live in old rancher, on upgrade they need to be moved to our `new` install location to be pulled into the jail

daxmc99:
LGTM

cjellick:
LGTM

PR #20402, opened 2019-05-23
Update drivers

dramich:
Problem:
Node drivers are not updated even with url changes

Solution:
Validate driver urls, hashes, and docker-machine versions to allow
schemas of node-drivers to be updated

### DIFF
--- a/pkg/controllers/management/drivers/nodedriver/machine_driver.go
+++ b/pkg/controllers/management/drivers/nodedriver/machine_driver.go
@@ -3,6 +3,7 @@ package nodedriver
 import (
 	"context"
 	"fmt"
+	"os/exec"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


pkg/controllers/management/drivers/nodedriver/machine_driver.go
```diff
@@ -86,7 +87,9 @@ func (m *Lifecycle) download(obj *v3.NodeDriver) (*v3.NodeDriver, error) {
 		existingSchema, err = m.schemaLister.Get("", schemaName)
 	}
 
-	if driver.Exists() && err == nil {
+	anno := obj.Annotations[driverSchemaRefresh]
+
+	if driver.Exists() && err == nil && anno == "true" {
```
cjellick: not really an acceptable solution as it wont allow for future updates of drivers. We need to permanently fix node drivers so that they update on upgrade.
Here's what I think we can/should do: 
Create a new field on nodeDriverStatus called applied version. It will capture what the last version of the driver we downloaded and installed was. For built-in drivers, you can use `docker-machine --version`. For third-party drivers, we can use the calculated hash after download (not the user supplied hash bc that is optional)

Next time this runs, recalculate the appliedVersion (by calling docker-machine --version or hashing the downloaded driver again). If it differs from the appliedHash on the status object, then update the dynamic schema.

cjellick:
LGTM

PR #20434, opened 2019-05-24
add test for updating persistent volume 

kinarashah:
https://github.com/rancher/types/pull/835 
rancher/rancher#19237


alena1108:
LGTM

Issue #20389, opened 2019-05-22
Rancher v2.2.4-rc7  Rook Helm chart will not install


dsargent3220:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1.  In global Catalogs add the rook-release catalog: https://charts.rook.io/release/
2.  Create a project with an "Unrestricted" pod security policy and switch to it.
3.  In Apps search for rook and deploy v1.0.1  No changes to yaml files are necessary unless your nodes have taints.

**Result:**
In Rancher v2.2.3 three deployments are created and the rook-ceph operator is deployed
In Rancher v2.2.4-rc7 the following error is immediately thrown
```Failed to install app rook-ceph. Error: UPGRADE FAILED: Get https://localhost:443/k8s/clusters/c-l8rtd/api/v1/namespaces/rook-ceph/configmaps?labelSelector=NAME%!D(MISSING)rook-ceph%!C(MISSING)OWNER%!D(MISSING)TILLER%!C(MISSING)STATUS%!D(MISSING)DEPLOYED: dial tcp: lookup localhost on 10.100.0.11:53: no such host```
Note that 10.100.0.11 is our local DNS server.  It is referencing localhost for some reason.

**Other details that may be helpful:**
This worked in 2.2.3, then broken when I upgraded our test Rancher server to 2.2.4-rc7.  I reverted Rancher back to 2.2.3 and it immediately worked again.  
I use the same cleanup steps each time, including removing the rook-ceph namespaces and ensuring all global objects are removed and the rook folders on the target nodes are cleaned up if needed.

**Environment information**
- Rancher version v2.2.4-rc7 
- Installation option single on CentOs

**Cluster information**
- Cluster type Custom
- Machine type VM, CentOs, 3 workers, each worker is 4 cores and 16 gigs of RAM
- Kubernetes version (use `kubectl version`): 1.13.5

Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T21:04:45Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"windows/amd64"}
[...]

Issue #20289, opened 2019-05-16
Timeout issues with helm after upgrading to 2.2.4-rc2


akkie:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
* Upgrade to Rancher 2.2.4-rc2
* Try to install or upgrade a helm chart

**Result:**
```
Failed to install app cluster-monitoring. Error: UPGRADE FAILED: Get https://localhost:443/k8s/clusters/c-9nd4d/api/v1/namespaces/cattle-prometheus/configmaps?labelSelector=NAME%!D(MISSING)cluster-monitoring%!C(MISSING)OWNER%!D(MISSING)TILLER%!C(MISSING)STATUS%!D(MISSING)DEPLOYED: dial tcp 10.0.15.1:443: i/o timeout
```

**Other details that may be helpful:**
* 2.2.3-rc9 fails also
* If I switch back to 2.2.3-rc4, 2.2.3-rc5, 2.2.3-rc6, 2.2.3-rc7 or 2.2.3-rc8 it works as expected

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):  2.2.4-rc2
- Installation option (single install/HA): single docker install

bmdepesa:
This was a problem with system-charts branch being incorrect in `v2.2.4-rc2`. 

Deploying apps and enabling cluster monitoring works correctly in `v2.2.4-rc5`.

cjellick:
can we close this issue @bmdepesa ?

akkie:
For me the issue remains with v2.2.4-rc5. In my opinion, the issue isn't monitoring specific. We had already used the `dev` branch of the system-charts library.

The issue occurs also for every other chart I try to install:
```
Failed to install app aerospike. Error: UPGRADE FAILED: Get https://localhost:443/k8s/clusters/c-9nd4d/api/v1/namespaces/aerospike/configmaps?labelSelector=NAME%!D(MISSING)aerospike%!C(MISSING)OWNER%!D(MISSING)TILLER%!C(MISSING)STATUS%!D(MISSING)DEPLOYED: dial tcp 10.0.15.1:443: i/o timeout
```

bmdepesa:
@akkie can you give more details on your installation? I am not able to reproduce this at all. Any app (including from system-charts) installs and operates fine in my v2.2.4-rc5 setup.

[...]

PR #20438, opened 2019-05-24
Fix Helm install error

orangedeng:
**Problem:**
Create jailer for helm install but missing the `/etc/hosts`

**Solution:**
Copy the `/etc/hosts` to `/opt/jail/$NAME/etc/` to fix the network
namespace.

Related issues:
https://github.com/rancher/rancher/issues/20389
https://github.com/rancher/rancher/issues/20289


dramich:
lgtm

Issue #20436, opened 2019-05-24
[Windows] Rancher doesn't currently allow 2 certs in the serverca file


jambajaar:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Create a standalone rancher server by following the instructions at: https://rancher.com/docs/rancher/v2.x/en/installation/single-node/
-- For SSL we’re using the instructions described at OPTION B-BRING YOUR OWN CERTIFICATE: SELF-SIGNED
-- We’re using image: rancher/rancher:v2.3.0-alpha3

- Configure AD authentication

- Create a new custom cluster with Kubernetes v1.14.1:
-- Register one Ubuntu 16.04 node with Docker 18.9.6 using all the roles (etcd, control and worker). Wait until the node is active.
-- Register one Ubuntu 16.04 node with Docker 18.9.6 using only the worker role. Wait until the node is active.
-- Register one Windows Server 2019 node 18.9.6 using only the worker role.
--- When running the registration command as instructed in the Rancher UI, we get the error:

```
Status: Downloaded newer image for rancher/rancher-agent:v2.3.0-alpha3

ERRO[0000] The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters.
```

**Other Useful Info:**
```
C:\etc\kubernetes\ssl\certs\serverca

-----BEGIN CERTIFICATE-----
xxxx
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
xxxx
-----END CERTIFICATE-----
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.3.0-Alpha

thxCode:
I have submitted a PR to fix this: https://github.com/rancher/rancher/pull/20441, as a workaround for now, you could execute in `powershell` as below:
```
# go into rancher agent directory
cd c:\etc\rancher

# modify the run.ps1 file
notepad.exe run.ps1

###notepad###
#     #import the self-signed certificate#
#-    $cacertsBytes = [Convert]::FromBase64String((Get-Content "$SSL_CERT_DIR\serverca" | select -Skip 1 | select -SkipLast 1))
#-    $cacertsFile = "$env:TEMP\serverca.cer"
#-    Set-Content -Value $cacertsBytes -Path $cacertsFile -Encoding Byte
[...]

PR #20441, opened 2019-05-24
Support self-signed server CA chain on Windows

thxCode:
**Problem:**
Consider only the CA chain with only one CA

**Solution:**
Split multiple CAs from `$SSL_CERT_DIR\serverca`, then import one by one

**Issue:**
https://github.com/rancher/rancher/issues/20436


loganhz:
LGTM

PR #20460, opened 2019-05-24
Force update drivers based on condition

dramich:
Problem:
An already installed driver with an unknown condition can get stuck in
ithat condition

Solution:
If a condition is unkown force update the driver


cjellick:
LGTM

PR #20463, opened 2019-05-24
Check if dir is empty before copy

dramich:
Problem:
If the bin dir is empty stat fails on "*"

Solution:
Verify the dir has contents before attempting the copy


cjellick:
LGTM

PR #20234, opened 2019-05-14
Rancher23 istio metric graph

fyery-chen:
This PR support Istio metric graph.

1. Get Istio metric graph through API
2. Create Istio metric expression after enabling Istio
3. Some integration test about it

TODO: After Charts PR is merge, we need to the istio template id to Rancher system charts repo.
Related PR:
https://github.com/rancher/system-charts/pull/44

PR #20486, opened 2019-05-28
Istio metrics

orangedeng:
This PR support Istio metric graph.

1. Get Istio metric graph through API
2. Create Istio metric expression after enabling Istio
3. Some integration test about it


loganhz:
LGTM https://github.com/rancher/rancher/pull/20234#issuecomment-495586018

tests/integration/suite/test_app.py
```diff
@@ -86,6 +87,69 @@ def test_app_wordpress(admin_pc, admin_mc):
     wait_for_workload(client, ns.name, count=2)
 
 
+@pytest.mark.skip(reason="istio disabled")
```
cjellick: A disabled test does not have much use. If this cannot be legitimately tested as an integration test, then remove it.

cjellick:
I believe we agreed that Istio would be just an app deployment, completely independent of Rancher. This tightly couples the app deployment to backend rancher code. It creates a dependency between Rancher and Istio that we had hoped to avoid. 

We may need to revisit this design.

pkg/controllers/user/istio/metricExpression.go
```diff
@@ -0,0 +1,555 @@
...
+		return err
+	}
+	if err := json.Unmarshal(jsondata, obj); err != nil {
+		return err
+	}
+	return nil
+}
+
+var (
+	IstioMetricsTemplate = `
```
cjellick: this seems weird, next week, id like to discuss what it is and why its needed, but i can merge as-is, not a blocker

loganhz: got it

loganhz:
Types updated

cjellick:
lgtm

Issue #20487, opened 2019-05-28
[Windows] Cannot add Windows worker node from Aliyun


loganhz:
rancher/rancher:v2.3.0-alpha-4

Steps:

1. Create Windows Server 2019 in Aliyun
2. Install Docker EE
3. Add this node as a worker into Rancher k8s cluster

Results:

ERRO[0000] Illegal characters in path

kingsd041:
Tested  rancher/rancher:master (06/12)
According to the reproduce step test, I can add windows nodes on Aliyun

Issue #20235, opened 2019-05-14
[Windows] rancher-agent on windows 1809, x509 cert signed by unkown


curx:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
start rancher-agent on windows 1809 worker via run cmd provided from rancherui

**Result:**
```
ERRO[0000] https://192.168.100.131 is not accessible (The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.)
```

**Other details that may be helpful:**
test rancher installation with autogenerated tls certificate,
created on a linux
```
docker run -d --port 80:80 -p 443:443 -v /opt/rancher:/var/lib/rancher rancher/rancher:master
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
rancher/rancher:master, commit: b7071be
```
{
...
  "org.label-schema.build-date": "2019-05-14T04:29:13Z",
  "org.label-schema.vcs-ref": "b7071befc39087148a8660754f0e972b62c1d6a6",
  "org.label-schema.vcs-url": "https://github.com/rancher/rancher.git"
...
}
```
rancher/rancher-agent:master
```
{
...
   "org.label-schema.build-date": "2019-05-14T04:26:34Z",
   "org.label-schema.vcs-ref": "b7071befc39087148a8660754f0e972b62c1d6a6",
   "org.label-schema.vcs-url": "https://github.com/rancher/rancher.git",
...
}
```
- Installation option (single install/HA): **single**

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): **custom**
- Machine type (cloud/VM/metal) and specifications (CPU/memory): **VM**
- Kubernetes version (use `kubectl version`): **v1.14.1-rancher1-2**

- Docker version (use `docker version`):

``` docker version
Client: Docker Engine - Enterprise
 Version:           18.09.6
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        1578dcadd2
 Built:             05/04/2019 02:34:11
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Enterprise
 Engine:
  Version:          18.09.6
  API version:      1.39 (minimum version 1.24)
  Go version:       go1.10.8
  Git commit:       1578dcadd2
  Built:            05/04/2019 02:32:24
  OS/Arch:          windows/amd64
[...]

PR #20540, opened 2019-05-29
Fix could not add Windows node from Aliyun

thxCode:
**Problem:**
- Occurring error when adding Windows node from Aliyun, it seems we could
not find the Docker named pipe from Aliyun Windows node, however, the
Docker is running.
- `[Net.ServicePointManager]::CertificatePolicy` is now obsolete:
https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.certificatepolicy?view=netframework-4.8#System_Net_ServicePointManager_CertificatePolicy.

**Solution:**
- Give a warning msg to user when we could not find the Docker named pipe.
- According to http://huddledmasses.org/blog/validating-self-signed-certificates-properly-from-powershell/,
we can hack `[Net.ServicePointManager]::ServerCertificateValidationCallback`.

**Issue:**
#20487
#20235

### DIFF
--- a/package/windows/run.ps1
+++ b/package/windows/run.ps1
@@ -133,20 +133,24 @@ if (-not $currentPrincipal.IsInRole([System.Security.Principal.WindowsBuiltInRol
 
 # set http client #
 try {
-    add-type @"
-    using System.Net;
-    using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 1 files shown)


loganhz:
LGTM

cjellick:
LGTM

PR #20543, opened 2019-05-29
[Monitoring]Validate namespace in workload metrics query

orangedeng:
**Problem:**
It can fetch workload metrics from other project.

**Solution:**
Should validate the workload if it is in this project or not.

### DIFF
--- a/pkg/api/customization/monitor/handler_utils.go
+++ b/pkg/api/customization/monitor/handler_utils.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/rancher/pkg/controllers/user/workload"
 	"github.com/rancher/rancher/pkg/ref"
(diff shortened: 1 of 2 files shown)


alena1108:
@orangedeng can we link it to the corresponding github ticket

pkg/api/customization/monitor/handler_utils.go
```diff
@@ -223,6 +224,9 @@ func parseMetricParams(userContext *config.UserContext, nodeLister v3.NodeLister
 		if err != nil {
 			return newMetricParams, err
 		}
+		if !validateNS(newMetricParams, ns) {
+			return nil, httperror.NewAPIError(httperror.PermissionDenied, "can't access to namespace "+ns+" in this project")
```
alena1108: Shall we log the project name as well? We should also alter the error message slightly. "Can't access namespace <name> from project <>"

pkg/api/customization/monitor/handler_utils.go
```diff
@@ -266,6 +270,9 @@ func parseMetricParams(userContext *config.UserContext, nodeLister v3.NodeLister
 			return nil, fmt.Errorf("pod name is empty")
 		}
 		ns, name := ref.Parse(podName)
+		if !validateNS(newMetricParams, ns) {
```
alena1108: same here, need to change an error message

orangedeng:
Updated

aiwantaozi:
LGTM

cjellick:
lgtm, ill let @alena1108 have one more look before we decide to merge

alena1108:
LGTM

PR #20556, opened 2019-05-29
Refactor rbac for sharing global ns resources

mrajashree:
This PR introduces two changes:
1. Multiclusterapp, global dns providers & entries are created in the cattle-global-data namespace. They can be shared by users other than their creators through the Members field. The rbac controllers create roles containing the resourceNames, and rolebindings with these roles for the members.
Now these resources can also be shared by everyone, by typing in the character `*` in the search box of members. This will translate to the `system:authenticated` k8s rbac group on the rolebinding for a role containing this resource. This will allow all authenticated users to have access to this resource.

2. Adds the same sharing logic with Members and all authenticated users for clustertemplates

### DIFF
--- a/pkg/controllers/management/cloudcredential/controller.go
+++ b/pkg/controllers/management/cloudcredential/controller.go
@@ -44,10 +44,11 @@ func (n *Controller) ccSync(key string, cloudCredential *v1.Secret) (runtime.Obj
 		return cloudCredential, fmt.Errorf("cloud credential %v has no creatorId annotation", cloudCredential.Name)
 	}
 	if err := globalnamespacerbac.CreateRoleAndRoleBinding(
-		globalnamespacerbac.CloudCredentialResource, cloudCredential.Name, cloudCredential.UID, []v3.Member{}, creatorID,
-		n.managementContext, "*"); err != nil {
+		globalnamespacerbac.CloudCredentialResource, cloudCredential.Name, "v1", creatorID, []string{"*"}, cloudCredential.UID, []v3.Member{},
+		n.managementContext); err != nil {
 		return nil, err
 	}
+
 	return cloudCredential, nil
 }
 
--- a/pkg/controllers/management/clustertemplate/clustertemplate_rbac.go
+++ b/pkg/controllers/management/clustertemplate/clustertemplate_rbac.go
@@ -1,0 +1,115 @@
+package clustertemplate
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/rancher/rancher/pkg/controllers/management/globalnamespacerbac"
+	"github.com/rancher/rancher/pkg/namespace"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	creatorIDAnn = "field.cattle.io/creatorId"
+	ctLabel      = "io.cattle.field/clusterTemplateId"
+)
+
+type clusterTemplateController struct {
+	clusterTemplates              v3.ClusterTemplateInterface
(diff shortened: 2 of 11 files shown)


mrajashree:
Need to hash names of roles/rolebindings based on offline discussion with @prachidamle

pkg/controllers/management/clustertemplate/clustertemplate_rbac.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	//see if any revision exists and call CreateRoleAndRoleBinding for it so that if this is an update request adding/removing members,
+	//it reflects on the permissions for revisions too
+	revisions, err := ct.clusterTemplateRevisionLister.List(namespace.GlobalNamespace, labels.SelectorFromSet(map[string]string{ctLabel: clusterTemplate.Name}))
+	if err != nil && !apierrors.IsNotFound(err) {
+		return clusterTemplate, err
+	}
+	for _, rev := range revisions {
+		metaAccessor, err := meta.Accessor(rev)
```
alena1108: can we just enqueue the revision? And let the `func (ctr *clusterTemplateRevisionController) sync(key string...` to set the rbac rules on per revision basis

pkg/controllers/management/cloudcredential/controller.go
```diff
@@ -44,10 +44,11 @@ func (n *Controller) ccSync(key string, cloudCredential *v1.Secret) (runtime.Obj
 		return cloudCredential, fmt.Errorf("cloud credential %v has no creatorId annotation", cloudCredential.Name)
 	}
 	if err := globalnamespacerbac.CreateRoleAndRoleBinding(
-		globalnamespacerbac.CloudCredentialResource, cloudCredential.Name, cloudCredential.UID, []v3.Member{}, creatorID,
-		n.managementContext, "*"); err != nil {
+		globalnamespacerbac.CloudCredentialResource, cloudCredential.Name, "v1", creatorID, []string{"*"}, cloudCredential.UID, []v3.Member{},
+		n.managementContext); err != nil {
```
prachidamle: I did not get this change for On Edit Cluster, Answers can be updated, please can you explain a little

mrajashree: @prachidamle not sure I get the `Answers` you're referring to? I just changed the args of the method. Previously for cloud credentials it was setting apiversion to `v1` and apigroup to ['*'], so instead I'm passing it from this method

prachidamle: Answers was a typo. Updated

pkg/controllers/management/clustertemplate/clustertemplate_rbac.go
```diff
@@ -0,0 +1,115 @@
...
+	if !ok {
+		return clusterTemplate, fmt.Errorf("clusterTemplate %v has no creatorId annotation", metaAccessor.GetName())
+	}
+
+	if err := globalnamespacerbac.CreateRoleAndRoleBinding(globalnamespacerbac.ClusterTemplateResource, clusterTemplate.Name,
+		globalnamespacerbac.RancherManagementAPIVersion, creatorID, []string{globalnamespacerbac.RancherManagementAPIVersion},
+		clusterTemplate.UID,
+		clusterTemplate.Spec.Members, ct.managementContext); err != nil {
+		return nil, err
+	}
```
prachidamle: Should we call globalnamespacerbac.CreateRoleAndRoleBinding only when obj.Spec.Members has changed? Looks like we call on every sync

prachidamle: I guess no way to differentiate the updates?

mrajashree: @prachidamle yeah I don't think so. The function `CreateRoleAndRoleBinding` does compare the subjects of rolebindings before actually update rolebindings, so we should be fine I guess?

prachidamle: as discussed offline with @alena1108 @mrajashree - we cannot know if Members got updates, but have to do a full sync - we should use cache lookups in globalnamespacerbac.CreateRoleAndRoleBinding as much as possible

prachidamle:
LGTM

PR #20561, opened 2019-05-29
Use dynamic ssh key name to find key v2.2

dramich:
Problem: 
key name is hardcoded to 'id_rsa'

Solution:
Use the name of the file that gets generated when a user passes in an sshkey instead of using a generated one


cjellick:
lgtm

Issue #20568, opened 2019-05-30
Input fluentd configure in advance mode got crash


aiwantaozi:
version: v2.2.4-rc19
**Reproduce step**:
1. go to cluster tools -> logging
2. select syslog -> Edit as File
3. input fluentd configure for syslog as below got crash, expect 200 because below configure is valid format
```
  @type remote_syslog
  host rancher.com
  port 60514
  severity notice
  protocol udp
  packet_size 65535
  verify_mode 0
```

**Reason**
This bug was introduced when update systemAccountManager for system apps, systemAccountManager does not init for tester deployer

**Crash**
```
runtime/debug.Stack(0xc012bd4c70, 0x33f77e0, 0x6a97ee0)
	/usr/local/go/src/runtime/debug/stack.go:24 +0xa7
github.com/rancher/rancher/vendor/github.com/rancher/norman/api.(*Server).ServeHTTP.func1(0x41d72a0, 0xc036fba6c0)
	/go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/api/server.go:175 +0x67
panic(0x33f77e0, 0x6a97ee0)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/rancher/rancher/pkg/systemaccount.(*Manager).GetProjectSystemUser(0x0, 0xc027ae4466, 0x7, 0xf, 0xc01ec66d00, 0x0)
	/go/src/github.com/rancher/rancher/pkg/systemaccount/systemaccount.go:142 +0xbc
github.com/rancher/rancher/pkg/controllers/user/logging/deployer.(*TesterDeployer).Deploy(0xc027b77950, 0x3a9b181, 0x7, 0xc039db7a58, 0x7, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/rancher/rancher/pkg/controllers/user/logging/deployer/testerdeployer.go:51 +0xe4
github.com/rancher/rancher/pkg/api/customization/logging.(*Handler).dryRunLoggingTarget(0xc006528600, 0xc007ee0c60, 0x3a9b181, 0x7, 0xc039db7a58, 0x7, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/rancher/rancher/pkg/api/customization/logging/action.go:190 +0x82c
[...]

PR #20569, opened 2019-05-30
[Logging]Init systemAccountManager for logging tester deployer

aiwantaozi:
Problem:
Input fluentd configure in advance mode got crash
Solution:
This bug was introduced when update systemAccountManager for system apps, need to init systemAccountManager for tester deployer

Issue:
https://github.com/rancher/rancher/issues/20568

### DIFF
--- a/pkg/controllers/user/logging/deployer/testerdeployer.go
+++ b/pkg/controllers/user/logging/deployer/testerdeployer.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rancher/types/apis/core/v1"
 	mgmtv3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	projectv3 "github.com/rancher/types/apis/project.cattle.io/v3"
+	"github.com/rancher/types/config"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/rancher/pkg/namespace"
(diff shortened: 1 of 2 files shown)


orangedeng:
LGTM

pkg/controllers/user/logging/deployer/testerdeployer.go
```diff
@@ -48,7 +50,7 @@ func (d *TesterDeployer) Deploy(level, clusterName, projectID string, loggingTar
 	}
 
 	systemProjectID := ref.Ref(systemProject)
-	creator, err := d.systemAccountManager.GetProjectSystemUser(systemProject.Name)
+	creator, err := d.systemAccountManager.GetSystemUser(systemProject.Spec.ClusterName)
```
cjellick: can you explain why you need to switch to the cluster-wide user rahter than project user?

aiwantaozi: Hi, this app will deployed to system project, and the chart has clusterrole rbac, project user couldn't create clusterroles rbac.
https://github.com/rancher/system-charts/blob/dev/charts/rancher-logging/0.1.1/charts/fluentd-tester/templates/clusterrole.yaml

cjellick: That is a good explanatio. Can you add that to the commit message? You don't need to reference the chart via a github link, but just add something like:
```
This app will deployed to system project, and the chart has clusterrole rbac, project user couldn't create clusterroles rbac.
```

aiwantaozi: Updated. Thanks

cjellick:
lgtm

PR #20574, opened 2019-05-30
Helm Chart: Default to Pulbic certs when `--set tls=external`

### LINKED ISSUE #20573: Helm Chart: `--set tls=external` is using Self Signed certificates instead of External

Bug

Helm chart `--set tls=external` is using Self Signed certificates instead of External

**Steps to reproduce (least amount of steps as possible):**

Unpack the latest helm chart to a directory called `rancher`

```
helm template \
    --name rancher \
    --namespace cattle-system \
    --set hostname=rancher.example.com \
    --set tls=external \
    rancher
```

**Result:**

The deployment is using self-signed certs, even though the Ingress is not managing certs:

```
      containers:
      - image: rancher/rancher:v2.2.3
        name: rancher
        ports:
        - containerPort: 80
          protocol: TCP
        args:
        # Rancher self-signed - don't clear ca certs
        - "--http-listen-port=80"
        - "--https-listen-port=443"
        - "--add-local=auto"
```

It should be using Public Trust CA by default:

```
      containers:
      - image: rancher/rancher:v2.2.3
        name: rancher
        ports:
        - containerPort: 80
          protocol: TCP
        args:
        # Public trusted CA - clear ca certs
        - "--no-cacerts"
        - "--http-listen-port=80"
        - "--https-listen-port=443"
        - "--add-local=auto"
```

**Environment information**
- Rancher version: 2.2.3

caleblloyd:
- Fixes #20573 
- Check for `.Values.tls == ingress` in addition to `Values.ingress.tls.source == "rancher"` when determining if self-signed should be used

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -59,7 +59,7 @@ spec:
 {{- end }}
 {{- if .Values.privateCA }}
         # Private CA - don't clear ca certs
-{{- else if eq .Values.ingress.tls.source "rancher" }}
+{{- else if and (eq .Values.tls "ingress") (eq .Values.ingress.tls.source "rancher") }}
         # Rancher self-signed - don't clear ca certs
 {{- else }}
         # Public trusted CA - clear ca certs
--- a/chart/tests/deployment_test.yaml
+++ b/chart/tests/deployment_test.yaml
@@ -73,3 +73,32 @@ tests:
   - equal:
       path: spec.template.spec.containers[1].image
       value: my.private.repo:5000/rancher/busybox:1.0.1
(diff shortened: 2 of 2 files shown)


caleblloyd:
This change matches the conditionals found here:

https://github.com/rancher/rancher/blob/master/chart/templates/issuer-rancher.yaml#L1-L2

caleblloyd:
@drpebcak @superseb I see that you have experience with the chart, could you please review this?

drpebcak:
@caleblloyd looks right to me. Can you add some tests to make sure this doesn't change the behavior of any other tls configuration options?

caleblloyd:
@drpebcak I have added the tests

PR #20582, opened 2019-05-30
Update drivers on startup master

dramich:
Problem:
Node drivers are not updated even with url changes

Solution:
Validate driver urls, hashes, and docker-machine versions to allow
schemas of node-drivers to be updated


cjellick:
LGTM

Issue #20192, opened 2019-05-13
[Windows] iSCSI workflow failed to start


kingsd041:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a Rancher v2.3.0-alpha1 cluster and add a child cluster with a windows node(The firewall is off) in it.
2. Deploy iSCSI workflow using yaml below
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app: windows-iscsi-test
  name: windows-iscsi-test
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: windows-iscsi-test
      name: windows-iscsi-test
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values:
                - windows-iscsi-test
            topologyKey: "kubernetes.io/hostname"
      nodeSelector:
        beta.kubernetes.io/os: windows
      containers:
      - name: windowsiscsitest
        image: mcr.microsoft.com/powershell:nanoserver-1809
        command:
        - pwsh.exe
        args:
        - -NoLogo
        - -NonInteractive
        - -Command
        - '&{$hs=hostname; echo $hs >> c:/mnt/hosts; do { echo "==========="; cat c:/mnt/hosts; echo "==========="; echo ""; start-sleep -s 10; } while($True)}'
        volumeMounts:
        - name: iscsi-volume
          mountPath: c:/mnt
      volumes:
      - name: iscsi-volume
        flexVolume:
          driver: "microsoft.com/iscsi.cmd"
          fsType: "NTFS"
          secretRef:
            name: "iscsi-secret"
          readOnly: false
          options:
            chapAuthDiscovery:  "false"
            chapAuthSession:  "true"
            targetPortal:  "172.31.12.208"
            iqn:  "iqn.2018-08.com.example.storagesrv01.target01"
            lun:  "0"
            authType:  "ONEWAYCHAP"
```
3. Waiting for the workflow to start

**Result:**
Pod failed to start:
```
Events:
  Type     Reason       Age    From                      Message
  ----     ------       ----   ----                      -------
  Normal   Scheduled    3m27s  default-scheduler         Successfully assigned default/iscsi to ec2amaz-3a24kcc
[...]

PR #20586, opened 2019-05-31
[Windows] Fix Windows iSCSI workflow failed to start

thxCode:
**Problem:**
The `msiscsi` service enabled by flexvolume plugin will cause storage fail to mount.

**Solution:**
Start `msiscsi` service before running agent.

**Issue:**
#20192

### DIFF
--- a/package/windows/run.ps1
+++ b/package/windows/run.ps1
@@ -101,11 +101,6 @@ function set-env-var {
     [Environment]::SetEnvironmentVariable($Key, $Value, [EnvironmentVariableTarget]::Machine)
 }
 
-function get-agent-service {
-    $svcRancherAgt = Get-Service -Name "rancher-agent" -ErrorAction Ignore
-    return $svcRancherAgt
-}
-
 ## END main definitaion
 #########################################################################
 ## START main execution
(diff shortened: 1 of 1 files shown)


package/windows/run.ps1
```diff
@@ -152,6 +147,22 @@ public static class SkipServerCertificateValidation
...
+    throw "rancher agent is already on running"
+}
+
+# check msiscsi servcie running #
+$svcMsiscsi = Get-Service -Name "msiscsi" -ErrorAction Ignore
+if ($svcMsiscsi -and ($svcMsiscsi.Status -ne "Running")) {
+    Set-Service -Name "msiscsi" -StartupType Automatic
+    Start-Service -Name "msiscsi" -ErrorAction Ignore
+    if (-not $?) {
+        warn "Start msiscsi service failed"
```
orangedeng: Should we throw error here? The our flex volume function won't work without this service.
Or we should program that rancher-agent service depend on msiscsi service

thxCode: iSCSI is just one way to store support, we still support other ways (like SMB, cloud provided storage), so I think we could not throw an error to stop the running action.

how about changing the warning message is enough, like `warn "Start msiscsi service failed, you may not be able to use the iSCSI flexvolume properly."`?

orangedeng:
LGTM

cjellick:
lgtm

Issue #20180, opened 2019-05-12
[Windows] Could not use GCEPersistentDisk as Volume on Windows worker


thxCode:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Prepare 1 Linux Ubuntu 18.04 and 1 Windows 2019
1. Set up a cluster with GCE cloud provider:
    ```yaml
    cloud_provider:
      name: gce
      customCloudProvider: |-
        [Global]
        project-id=<your projcet ID, optional>
        network-name=<your network, optional if using default network>
        subnetwork-name=<your subnetwork of the above network, optional if using default network>
        node-instance-prefix=<your instance group name/your instance name specific prefix>
        node-tags=<your network tags, optional if not patching any tags>
    services:
      kube-controller:
        extra_args:
          configure-cloud-routes: true
    ```
1. Add a Linux control plane + etcd + worker node
1. Add a Windows node
1. Deploy a Windows workload with an existing GCEPersistentDisk, following https://kubernetes.io/docs/concepts/storage/volumes/#gcepersistentdisk

**Result:**
-. Workload event:
![image](https://user-images.githubusercontent.com/3394724/57578421-b4f3de00-74be-11e9-8bec-0f8a8a921744.png)

- kubelet logs from Windows:
![image](https://user-images.githubusercontent.com/3394724/57578422-c1783680-74be-11e9-8b8f-eaa0517a1be1.png)

**Other details that may be helpful:**
A related issue from Kubernetes: https://github.com/kubernetes/kubernetes/issues/74674

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): Cloud
- Kubernetes version (use `kubectl version`): 1.14.1

sowmyav27:
Closing this issue, as its a duplicate 
Tracking this issue in [https://github.com/rancher/rancher/issues/20839](url)

PR #20591, opened 2019-05-31
[Windows] Fix could not use GCEPersistentDisk on Windows

thxCode:
**Problem:**
K8S GCE cloud provider has replace `Get-GcePdName` with native windows
support of `Get-Disk`, but the GCE VMs haven't injected the new
method.

**Solution:**
Following
https://github.com/kubernetes/kubernetes/blob/f49fe2a7504c40684573f3e4928e8b0bfdfb2b30/cluster/gce/windows/k8s-node-setup.psm1#L782-L795,
we can patch the `Get-GcePdName`.

**Issue:**
https://github.com/rancher/rancher/issues/20180

### DIFF
--- a/package/windows/hyperkube.ps1
+++ b/package/windows/hyperkube.ps1
@@ -943,9 +943,39 @@ function init {
         try {
             $splited = $hostname -split "\."
             if ($splited.Length -lt 1) {
-                $NodeName = $splited[0]
+                $nn = $splited[0]
             }
         } catch {}
+
+        ## repair Get-GcePdName method ##
+        $getGcePodNameCommand = Get-Command -Name Get-GcePdName -ErrorAction Ignore
+        if (-not $getGcePodNameCommand) {
+            print "Can't find Get-GcePodName, try to repair ..."
+
+            $dllDownloadURL = "https://github.com/pjh/gce-tools/raw/master/GceTools/GetGcePdName/GetGcePdName.dll"
+            $dllPath = "$RancherDir\GetGcePdName.dll"
+            try {
+                Invoke-WebRequest -TimeoutSec 300 -UseBasicParsing -Uri $dllDownloadURL -OutFile $dllPath
+            } catch {}
+            if (-not $?) {
+                throw ("Failed to download GetGcePodName.dll from '{0}'" -f $dllDownloadURL)
+            }
+
+            print "Importing GetGcePodName.dll, wait a few seconds ..."
+
+            $profilePath = "$PsHome\profile.ps1"
+            if (-not (Test-Path $profilePath)) {
+                $nul = New-Item -Path $profilePath -Type file -ErrorAction Ignore
+            }
+            $appendProfile =
+@'
+Unblock-File -Path DLLPATH -ErrorAction Ignore
+Import-Module -Name DLLPATH -ErrorAction Ignore
+'@
+            Add-Content -Path $profilePath -Value $appendProfile.replace('DLLPATH', $dllPath) -ErrorAction Ignore
+
+            print ".................. OK"
+        }
     }
 
     set-env-var -Key "NODE_NAME" -Value $nn.ToLower()
--- a/package/windows/cleanup.ps1
+++ b/package/windows/cleanup.ps1
@@ -112,6 +112,11 @@ try {
         }
     }
 
+    # restore repair #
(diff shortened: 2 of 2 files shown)


orangedeng:
LGTM

package/windows/hyperkube.ps1
```diff
@@ -943,9 +943,39 @@ function init {
...
+                $nn = $splited[0]
             }
         } catch {}
+
+        ## repair Get-GcePdName method ##
+        $getGcePodNameCommand = Get-Command -Name Get-GcePdName -ErrorAction Ignore
+        if (-not $getGcePodNameCommand) {
+            print "Can't find Get-GcePodName, try to repair ..."
+
+            $dllDownloadURL = "https://github.com/pjh/gce-tools/raw/master/GceTools/GetGcePdName/GetGcePdName.dll"
```
alena1108: How will it work in airgapped installs?

package/windows/hyperkube.ps1
```diff
@@ -943,9 +943,39 @@ function init {
...
+            $dllPath = "$RancherDir\GetGcePdName.dll"
+            try {
+                Invoke-WebRequest -TimeoutSec 300 -UseBasicParsing -Uri $dllDownloadURL -OutFile $dllPath
+            } catch {}
+            if (-not $?) {
+                throw ("Failed to download GetGcePodName.dll from '{0}'" -f $dllDownloadURL)
+            }
+
+            print "Importing GetGcePodName.dll, wait a few seconds ..."
```
alena1108: we should put GetGcePodName.dll to a constant. It's been used quite a few times.

package/windows/hyperkube.ps1
```diff
@@ -943,9 +943,39 @@ function init {
...
             }
         } catch {}
+
+        ## repair Get-GcePdName method ##
+        $getGcePodNameCommand = Get-Command -Name Get-GcePdName -ErrorAction Ignore
+        if (-not $getGcePodNameCommand) {
+            print "Can't find Get-GcePodName, try to repair ..."
+
+            $dllDownloadURL = "https://github.com/pjh/gce-tools/raw/master/GceTools/GetGcePdName/GetGcePdName.dll"
+            $dllPath = "$RancherDir\GetGcePdName.dll"
```
loganhz: Is it only for GCE? Will you try to download it if your Windows server is in your local datacenter?

package/windows/hyperkube.ps1
alena1108: Unless the fix is for GCE specifically, and GCE implies non-airgapped installs. Then we should make sure this code is never called for non-gce cloud providers and on prem setups

package/windows/hyperkube.ps1
```diff
@@ -943,9 +943,39 @@ function init {
...
+            try {
+                Invoke-WebRequest -TimeoutSec 300 -UseBasicParsing -Uri $dllDownloadURL -OutFile $dllPath
+            } catch {}
```
alena1108: shall we log anything in catch()? or we don't care

thxCode: Yes, we don't care the exception.

package/windows/hyperkube.ps1
thxCode: They are not a variable, just a name, although they are repeated, they still seem to be very clear.

package/windows/hyperkube.ps1
thxCode: @alena1108 , the `Get-GcePodName` is only used to GCE. The Windows host of GCE [hasn't supported](https://github.com/kubernetes/kubernetes/issues/74674) `Get-GcePodName` yet, but the next Windows Kuberentes Cluster Service will [support this function](https://github.com/kubernetes/kubernetes/blob/f49fe2a7504c40684573f3e4928e8b0bfdfb2b30/cluster/gce/windows/k8s-node-setup.psm1#L782-L795). So I only deal with it when using the GCE cloud provider: 

1. Use `Get-Command -Name Get-GcePdName -ErrorAction Ignore` to detect `Get-GcePdName` existing or not
1. If not, download `GetGcePdName.dll` and import it

Should we need to consider AirGap on GCE? cc @loganhz

cjellick:
lgtm

PR #20108, opened 2019-05-09
Updates for global registry support

gitlawr:
1. Add `global-registry-enabled` setting, It is used to detect whether global registry is enabled for non-admin users.
2. Copy ca cert to `/etc/docker/certs.d/<server-host>[:port]/ca.crt` on the host([see reference](https://docs.docker.com/engine/security/certificates/)). So that nodes on clusters can connect to the global registry in custom CA root cases. CA cert locates in `/etc/kubernetes/ssl/certs/serverca` both inside node agents & on the hosts before, this keeps the same.
  In this solution, `/etc/docker/certs.d` instead of `/etc/docker/certs.d/<server-host>[:port]` on the host mounts node agent containers, due to two limitations:
  - Colon is not supported in kuberentes volume path, which may be needed in docker cert path. [see](https://github.com/kubernetes/kubernetes/issues/21681#issuecomment-187260705)
  - hostpath-subpath is broken, And looks like we need a deprecated kubelet flag for [that](https://github.com/rancher/rancher/issues/14836).

Depending on the system-chart PR: https://github.com/rancher/system-charts/pull/41

https://github.com/rancher/rancher/issues/20597

PR #20605, opened 2019-06-01
[Validation Tests]Add global registry validation tests

gitlawr:
Depending on pr: https://github.com/rancher/rancher/pull/20108

### DIFF
--- a/tests/validation/tests/v3_api/test_global_registry.py
+++ b/tests/validation/tests/v3_api/test_global_registry.py
@@ -1,0 +1,124 @@
+import pytest
+import errno
+import warnings
+from distutils.spawn import find_executable
+from .common import * # NOQA
+
+setup = {"p_client": None, "project": None,
+         "admin_client": None, "server_url": None,
+         "server_host": None}
+
+DISTRIBUTION_HEADER = "docker-distribution-api-version"
+REGISTRY_TEMP_ID = "catalog://?catalog=system-library&template=harbor" \
+                   "&version=1.7.5-rancher1"
+
+
+def test_global_registry_enable():
+    admin_client = setup["admin_client"]
+    p_client = setup["p_client"]
+    p = setup["project"]
+
+    setting = admin_client.by_id_setting(id="global-registry-enabled")
+    assert setting is not None, \
+        "Expected to find global-registry-enabled setting"
+
+    p_client.create_app(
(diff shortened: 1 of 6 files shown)


sangeethah:
@gitlawr , We want the validation test to cover the p1 basic workflow which involves the following:

```
1. Enable global registry with all necessary options including storage options.
2. Access Harbor UI.
3. Create an account.
4. Create a project in Harbor UI.
5. Push image to this project with appropriate tag.
6. In an existing cluster in Rancher server:
Add a project.
Add a registry in the project with the credentials from Step 3.
Deploy a workload using the image in step 5.
Workload should get deployed successfully.
```

tests/validation/tests/v3_api/test_global_registry.py
```diff
@@ -0,0 +1,124 @@
...
+        assert proc.returncode == 0, proc.stderr
+    else:
+        response = requests.get(setup["server_url"] + "/v2/",
+                                verify=False)
+        assert response.status_code == 401
+        assert response.headers[DISTRIBUTION_HEADER] == 'registry/2.0'
+
+
+def docker_exist():
+    return find_executable("docker") is not None
```
sangeethah: If we expect Docker to be present in the client from where the test cases are being executed , we should get this installed in the test container.

tests/validation/tests/v3_api/test_global_registry.py
```diff
@@ -0,0 +1,124 @@
...
+        externalId=REGISTRY_TEMP_ID,
+        targetNamespace="cattle-system",
+        projectId=p.id,
+        answers={
+            "expose.ingress.host": setup["server_host"],
+            "externalURL": setup["server_url"],
+            "persistence.enabled": "false",
+            "clair.enabled": "false",
+            "notary.enabled": "false"
+        },
```
sangeethah: We should pass all parameters that are exposed in the UI workflow for enabling global registry in answers and the parameter values have to be ideally the default ones presented in the UI. 
This would mean , we need to have persistence.enabled , clair.enabled and notary.enabled set to True.

gitlawr:
Updated.
Add tests to
    1. Access Harbor UI and create a test user/project.
    2. Push to Global registry
    3. Deploy workload using the image from Global Registry.

Update test setup to have docker & selenium utilities.

Issue #20666, opened 2019-06-05
Project member / owner is unable to create istio gateway or virtual service


bmdepesa:
rancher/rancher:master 6/3 8bfb1af5

A project member without the cluster owner role is not able to create an istio gateway or virtual service.

* As a cluster owner, create a cluster
* Enable Istio with a gateway
* Login as a project owner in the cluster, and import the yaml files here:
  * https://raw.githubusercontent.com/rancher/rancher/80bdad0ebb8bcd3d2fd9d55dc45c1851443bcf30/tests/validation/tests/v3_api/resource/istio_files/bookinfo.yaml
  * https://raw.githubusercontent.com/rancher/rancher/80bdad0ebb8bcd3d2fd9d55dc45c1851443bcf30/tests/validation/tests/v3_api/resource/istio_files/bookinfo-gateway.yaml

The second yaml will fail to import with an error: 
```
""message": "services "http:grafana:80" is forbidden: User "u-2wz9d" cannot get resource "services/proxy" in API group "\" in the namespace "istio-system","
```

Through kubectl the user is unable to list istio resources:
```
> kubectl get virtualservice -n istions2
Error from server (Forbidden): virtualservices.networking.istio.io is forbidden: User "u-tf87s" cannot list resource "virtualservices" in API group "networking.istio.io" in the namespace "istions2"
> kubectl get gateway -n istions2
Error from server (Forbidden): gateways.networking.istio.io is forbidden: User "u-tf87s" cannot list resource "gateways" in API group "networking.istio.io" in the namespace "istions2
```

The istio sidecars for the workloads are creating correctly.

bmdepesa:
rancher/rancher:master fed523b3 6/7

* Create cluster
* Enable istio
* Add user2 as project owner, user3 as project member
* Login as user2
* Import service yaml from test case
* View Service Mesh dashboard
* Links are accessible, traffic graph and metrics function.
* delete resources
* Login as user3
* View Service Mesh dashboard
* Links are accessible, traffic graph and metrics function.

PR #20668, opened 2019-06-05
[Istio]Add RBAC role for istio resource

orangedeng:
Grant istio related permissions to Project Member / Owner / Read-only.
All the added permission are namespaced scope

`networking.istio.io`, 
`config.istio.io`, 
`authentication.istio.io`
`rbac.istio.io`

Related issue:
https://github.com/rancher/rancher/issues/20666

### DIFF
--- a/app/role_data.go
+++ b/app/role_data.go
@@ -175,6 +175,10 @@ func addRoles(management *config.ManagementContext) (string, error) {
 		addRule().apiGroups("management.cattle.io").resources("catalogtemplateversions").verbs("*").
 		addRule().apiGroups("monitoring.cattle.io").resources("prometheus").verbs("view").
 		addRule().apiGroups("monitoring.coreos.com").resources("prometheuses", "prometheusRules", "serviceMonitors").verbs("*").
+		addRule().apiGroups("networking.istio.io").resources("destinationrules", "envoyfilters", "gateways", "serviceentries", "sidecars", "virtualservices").verbs("*").
+		addRule().apiGroups("config.istio.io").resources("apikeys", "authorizations", "checknothings", "circonuses", "deniers", "fluentds", "handlers", "kubernetesenvs", "kuberneteses", "listcheckers", "listentries", "logentries", "memquotas", "metrics", "opas", "prometheuses", "quotas", "quotaspecbindings", "quotaspecs", "rbacs", "reportnothings", "rules", "solarwindses", "stackdrivers", "statsds", "stdios").verbs("*").
+		addRule().apiGroups("authentication.istio.io").resources("policies").verbs("*").
+		addRule().apiGroups("rbac.istio.io").resources("rbacconfigs", "serviceroles", "servicerolebindings").verbs("*").
 		setRoleTemplateNames("admin")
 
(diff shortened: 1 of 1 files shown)


app/role_data.go
```diff
@@ -298,6 +301,12 @@ func addRoles(management *config.ManagementContext) (string, error) {
 		addRule().apiGroups("monitoring.cattle.io").resources("prometheus").verbs("view").
 		setRoleTemplateNames("view")
 
+	rb.addRoleTemplate("Manage Istio Network", "istionetwork-manage", "project", true, false, false, false).
```
gitlawr: s/Network/Networking

gitlawr:
LGTM

app/role_data.go
```diff
@@ -298,6 +301,12 @@ func addRoles(management *config.ManagementContext) (string, error) {
 		addRule().apiGroups("monitoring.cattle.io").resources("prometheus").verbs("view").
 		setRoleTemplateNames("view")
 
+	rb.addRoleTemplate("Manage Istio Networking", "istionetwork-manage", "project", true, false, false, false).
```
cjellick: Do we really need to build these in? From a practical perspective, who would ever use them?

cjellick:
remove the new istionetwork-manage and stionetwork-view roles and add the necessary rule to project-member.
Will be lgtm after that

app/role_data.go
```diff
@@ -175,6 +175,10 @@ func addRoles(management *config.ManagementContext) (string, error) {
 		addRule().apiGroups("management.cattle.io").resources("catalogtemplateversions").verbs("*").
 		addRule().apiGroups("monitoring.cattle.io").resources("prometheus").verbs("view").
 		addRule().apiGroups("monitoring.coreos.com").resources("prometheuses", "prometheusRules", "serviceMonitors").verbs("*").
+		addRule().apiGroups("networking.istio.io").resources("destinationrules", "envoyfilters", "gateways", "serviceentries", "sidecars", "virtualservices").verbs("*").
+		addRule().apiGroups("config.istio.io").resources("apikeys", "authorizations", "checknothings", "circonuses", "deniers", "fluentds", "handlers", "kubernetesenvs", "kuberneteses", "listcheckers", "listentries", "logentries", "memquotas", "metrics", "opas", "prometheuses", "quotas", "quotaspecbindings", "quotaspecs", "rbacs", "reportnothings", "rules", "solarwindses", "stackdrivers", "statsds", "stdios").verbs("*").
+		addRule().apiGroups("authentication.istio.io").resources("policies", "meshpolicies").verbs("*").
+		addRule().apiGroups("rbac.istio.io").resources("clusterrbacconfigs", "rbacconfigs", "serviceroles", "servicerolebindings").verbs("*").
```
gitlawr: @loganhz If we want to grant all istio resources, there are also `mesh/v1alpha1`, `policy/v1beta1` 
Can you check if we can use wildcards so that we don't miss anything, e.g. `apiGroups("*.istio.io").resources("*")`

loganhz: I would like to give the permission which kiali is using. So I don't want to use *

gitlawr:
lgtm

Issue #20339, opened 2019-05-20
Slack notification not sent when pod is restarted


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a slack notifier. I clicked the "Test" button and got: 
<img width="346" alt="image" src="https://user-images.githubusercontent.com/468462/58039724-74144e80-7ae8-11e9-9486-729196b612dd.png">
2. Created a one node custom cluster on AWS using t3a.medium, ros 1.5.1.
3. Turned on monitoring for the cluster.
4. Turned on monitoring for the default project.
5. Deployed a workload in the default project called "nginxtest" using the "nginx" image. Confirmed workload was running.
6. Created a new alert in the default project to trigger when the pod is restarted: <img width="1242" alt="image" src="https://user-images.githubusercontent.com/468462/58039577-ffd9ab00-7ae7-11e9-92c7-c9ff7e84d34b.png">
7. Edited the alert group and set my slack notifier as the notifier.
8. On the worker node, ran a "docker ps | grep nginxtest", then with the id, ran "docker inspect <id> | grep -i PID", then ran "sudo kill -9 <pid>". This caused the pod to restart: 
<img width="1239" alt="image" src="https://user-images.githubusercontent.com/468462/58039995-277d4300-7ae9-11e9-9012-70d7bef9b7be.png">

**Result:**
No slack notification was received. I would expect to receive one since I set an alert that should be triggered if the pod was restarted.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.2.1
- Installation option (single install/HA):
single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
Cloud - AWS t3a.medium
- Kubernetes version (use `kubectl version`):

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.7", GitCommit:"6f482974b76db3f1e0f5d24605a9d1d38fad9a2b", GitTreeState:"clean", BuildDate:"2019-03-25T02:52:13Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #20669, opened 2019-06-05
[Alerting]Fix pod restart not fire alerting after server restarted

aiwantaozi:
Problem:
pod restart count cache was stored in cache when alert create, when server restart the cache disappeared, so couldn't get pre pod restart count
Solution:
store current pod restart count in the cache when check pod restart

Issue:
https://github.com/rancher/rancher/issues/20339

### DIFF
--- a/pkg/controllers/user/alert/watcher/pod.go
+++ b/pkg/controllers/user/alert/watcher/pod.go
@@ -81,7 +81,6 @@ type ProjectAlertLifecycle struct {
 }
 
 func (l *ProjectAlertLifecycle) Create(obj *v3.ProjectAlertRule) (runtime.Object, error) {
-	l.podWatcher.podRestartTrack.Store(obj.Namespace+":"+obj.Name, make([]restartTrack, 0))
 	return obj, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/alert/watcher/pod.go
```diff
@@ -156,86 +156,82 @@ func (w *PodWatcher) watchRule() error {
...
 }
 
 func (w *PodWatcher) getRestartTimeFromTrack(alert *v3.ProjectAlertRule, curCount int32) int32 {
 	name := alert.Name
 	namespace := alert.Namespace
+	now := time.Now()
+	currentRestartTrack := restartTrack{Count: curCount, Time: now}
+	currentRestartTrackArr := []restartTrack{currentRestartTrack}
 
 	obj, ok := w.podRestartTrack.Load(namespace + ":" + name)
```
orangedeng: Should we use `LoadOrStore`?

pkg/controllers/user/alert/watcher/pod.go
```diff
@@ -156,86 +156,82 @@ func (w *PodWatcher) watchRule() error {
...
 	if !ok {
+		w.podRestartTrack.Store(namespace+":"+name, currentRestartTrackArr)
```
orangedeng: Do we really need to store the whole track in the map? I think you just need the latest one. So maybe we just need to store that.

orangedeng:
LGTM

pkg/controllers/user/alert/watcher/pod.go
```diff
@@ -156,86 +155,76 @@ func (w *PodWatcher) watchRule() error {
...
-			tracks = tracks[i:]
-			tracks = append(tracks, restartTrack{Count: curCount, Time: now})
-			w.podRestartTrack.Store(namespace+":"+name, tracks)
-			return track.Count
+	currentRestartTrack := restartTrack{Count: curCount, Time: now}
+	currentRestartTrackArr := []restartTrack{currentRestartTrack}
+
+	obj, loaded := w.podRestartTrack.LoadOrStore(namespace+":"+name, currentRestartTrackArr)
+	if loaded {
+		tracks := obj.([]restartTrack)
```
alena1108: Looks like formatting is off. It is more than 4 spaces

aiwantaozi: Hi Alena, it's 4 spaces, may the web preview looks like more than 4 spaces

cjellick:
will let @alena1108 finish the review

alena1108:
LGTM

PR #20678, opened 2019-06-05
publish on head of master and release branches

drpebcak:
This PR extends the functionality used to automatically build docker images when code is pushed to master so that `release/v*` branches also get created.

It also syncs the drone.yaml with the master branch to get changes that have been made there related to tagging conventions and manifests.

`master` builds are tagged `master`. `release/v*` branches have the `/` replaced with a `-` - for example `release-v2.2`

I have verified the following cases:
* We do NOT create docker images or manifests on push to branches that are not `master` or `release/v*`.
* When we tag a branch, it does NOT trigger the `head` steps.
* When we push to `master` or `release/v*` it DOES create and publish a `rancher/rancher` and `rancher/rancher-agent` image and manifest.

What I can't verify:
* Any steps that actually involve doing a release - since I can't replicate everything required to make that happen. These should still work as expected, but some changes have been made to the drone.yml that affects these, to bring this branch in line with master.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,6 +1,6 @@
 ---
 kind: pipeline
-name: default-amd64
+name: default-linux-amd64
 
 platform:
   os: linux
@@ -8,7 +8,6 @@ platform:
 
 steps:
 - name: build
-  pull: default
   image: rancher/dapper:v0.4.1
   commands:
   - dapper ci
@@ -23,7 +22,6 @@ steps:
     - tag
 
 - name: stage-binaries
-  pull: default
   image: rancher/dapper:v0.4.1
   commands:
   - "cp -r ./bin/* ./package/"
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -307,95 +319,143 @@ trigger:
...
 
 platform:
   os: linux
-  arch: amd64 
+  arch: amd64
 
 steps:
 - name: chart-promote
-  pull: default
   image: rancher/dapper:v0.4.1
```
cloudnautique: Why did we remove this?

drpebcak: The drone team indicated that it doesn't do anything. `pull: Always` is documented as an option, but setting it to `default` is not. As far as I can tell it is a holdover from the 0.8 syntax migration.

cjellick:
LGTM

PR #20697, opened 2019-06-06
implements external database configure for helm charts

whywaita:
Hi contributors!

This PR becomes to enable configure external Database value.

Now, Helm charts have not a data's persistent solution.
So, it should use a stateful layer outside.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -65,6 +65,16 @@ spec:
         # Public trusted CA - clear ca certs
         - "--no-cacerts"
 {{- end }}
+{{- if .Values.database.use }}
+        - "--db-host {{ .Values.database.host }}"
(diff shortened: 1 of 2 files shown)


vincent99:
Huh?  This chart is for rancher 2.x, which uses etcd and has no MySQL database.

whywaita:
oh, really? 🙇

whywaita:
@vincent99 btw, please teach me about "how to configure external etcd cluster" if you know.

vincent99:
The chart deploys Rancher **into** an existing Kubernetes cluster, which has its own etcd backing it, wherever it is.  Rancher 2.x's "database" is custom resource definitions (CRDs) within the Kubernetes API.

whywaita:
🆗 Thank you!

Issue #20683, opened 2019-06-05
Flakey Rancher Tests(PR)


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
push commit to rancher PR

**Result:**
sometimes test_project_owner and test_writing_config_to_disk fail:
https://drone-pr.rancher.io/rancher/rancher/2336/2/2
https://drone-pr.rancher.io/rancher/rancher/2317

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 
- Installation option (single install/HA):

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

PR #20706, opened 2019-06-06
Fix node test

rmweir:
Problem:

It is possible that cloud credential is not immediately available. This will cause the writing config to disk test to fail.

Solution:
Create will be retried if the credential is unavailable.

Issue:
https://github.com/rancher/rancher/issues/20683

### DIFF
--- a/tests/integration/suite/test_node.py
+++ b/tests/integration/suite/test_node.py
@@ -130,9 +130,20 @@ def test_writing_config_to_disk(admin_mc, wait_remove_resource):
         digitaloceancredentialConfig={"accessToken": "test"})
     wait_remove_resource(cloud_credential)
     userdata = "do cool stuff"
-    node_template = client.create_node_template(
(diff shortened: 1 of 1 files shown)


tests/integration/suite/test_node.py
```diff
@@ -130,9 +130,20 @@ def test_writing_config_to_disk(admin_mc, wait_remove_resource):
...
     userdata = "do cool stuff"
-    node_template = client.create_node_template(
-        digitaloceanConfig={'userdata': userdata}, name='danssweetassthing',
-        cloudCredentialId=cloud_credential.id)
+
+    def _node_template():
+        try:
+            return client.create_node_template(
+                digitaloceanConfig={'userdata': userdata},
+                name='danssweetassthing',
```
daxmc99: 👍

daxmc99:
LGTM

dramich:
lgtm

Issue #20709, opened 2019-06-06
Node-driver based provisioning does not respect proxy variables HTTP_PROXY and HTTPS_PROXY


Oats87:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

1. Create an environment that does not allow direct public access to the internet
2. Create a proxy that does allow access to the internet in that environment (tinyproxy is a good one for testing)
3. Start rancher server with the HTTP_PROXY and HTTPS_PROXY environment variables set so that it uses these settings for reaching the external internet (https://rancher.com/docs/rancher/v2.x/en/installation/single-node/proxy/)
4. Try to provision nodes using one of Rancher's node drivers (https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/node-pools/)

**Result:**
Nodes cannot be provisioned because `docker-machine` is not using the required proxies

sangeethah:
Nodes get stuck in "Provisioning" state for a long time and then gets to following error state:

<img width="1059" alt="Screen Shot 2019-06-06 at 8 00 34 PM" src="https://user-images.githubusercontent.com/4266958/59078848-c51c9480-8895-11e9-9f0b-0789ce1be9ca.png">

rancher server logs:
```
2019/06/07 02:54:53 [INFO] Creating jail for c-mf9dz
2019/06/07 02:54:53 [INFO] Provisioning node sangeetestme1
2019/06/07 02:54:53 [INFO] stdout: Creating CA: /management-state/node/nodes/sangeetestme1/certs/ca.pem
2019/06/07 02:54:53 [INFO] Creating CA: /management-state/node/nodes/sangeetestme1/certs/ca.pem
2019/06/07 02:54:53 [INFO] stdout: Creating client certificate: /management-state/node/nodes/sangeetestme1/certs/cert.pem
2019/06/07 02:54:53 [INFO] Creating client certificate: /management-state/node/nodes/sangeetestme1/certs/cert.pem
2019/06/07 02:54:53 [INFO] stdout: Running pre-create checks...
2019/06/07 02:54:53 [INFO] Running pre-create checks...


2019/06/07 02:57:57 [INFO] stdout: The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.
2019/06/07 02:57:57 [INFO] The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.
[...]

PR #20727, opened 2019-06-07
Whitelist envvars for jail

dramich:
Problem:
The jail clears the env and sets only ones required to run. This drops
envvars that might be needed for a proxy

Solution:
Add a whitelist of envvars to pull into the jail with a default list to
support proxies or user updated setting

Issue: https://github.com/rancher/rancher/issues/20709

### DIFF
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -8,11 +8,13 @@ import (
 	"os/user"
 	"path"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/rancher/rancher/pkg/settings"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/jailer/jailer_test.go
```diff
@@ -0,0 +1,46 @@
...
+		},
+		{
+			input:    []string{"BAZ=FOOBAZ"},
+			expected: []string{"BAZ=FOOBAZ", "ENVVAR_FOO=ALPHA"},
+			envValue: "ALPHA",
+		},
+	}
+
+	for _, tc := range testCases {
+		os.Setenv("ENVVAR_FOO", tc.envValue)
```
daxmc99: Ridiculous nit but could the test cases be ordered:
{
envValue: "",
input:"",
expected:""
}

dramich: Done

pkg/jailer/jailer.go
```diff
@@ -81,3 +83,20 @@ func GetUserCred() (*syscall.Credential, error) {
...
+
+func WhitelistEnvvars(envvars []string) []string {
+	wl := settings.WhitelistEnvironmentVars.Get()
+	envWhiteList := strings.Split(wl, ",")
+
+	if len(envWhiteList) == 0 {
+		return envvars
+	}
+
+	for _, wlVar := range envWhiteList {
```
alena1108: Do you think we should do `trimpSpace` on the each wlVar?

alena1108: Also do you know if Getenv is case sensitive? And can these variables be configured in a lower case on users' setups

dramich: Added trim space. Per discussion not going to do anything about case, user can utilize the setting if they need to change the case of the envvar being looked up.

alena1108:
LGTM

Issue #20696, opened 2019-06-06
rancher-save-images.sh fails v2.2.4


dengtuoyu:
https://github.com/rancher/rancher/releases/download/v2.2.4/rancher-images.txt
https://github.com/rancher/rancher/releases/download/v2.2.4/rancher-save-images.sh

docker pull jetstack/cert-manager-controller:v0.6.2
Error response from daemon: pull access denied for jetstack/cert-manager-controller, repository does not exist or may require 'docker login'

docker pull rancher/coredns-coredns:1.1.2
Error response from daemon: manifest for rancher/coredns-coredns:1.1.2 not found

docker pull rancher/prom-prometheus:v2.3.1
Error response from daemon: manifest for rancher/prom-prometheus:v2.3.1 not found

bentastic27:
Was able to confirm:

```
ben@ben-vm:~$ docker pull jetstack/cert-manager-controller:v0.6.2
Error response from daemon: pull access denied for jetstack/cert-manager-controller, repository does not exist or may require 'docker login'
ben@ben-vm:~$ docker pull rancher/coredns-coredns:1.1.2
Error response from daemon: manifest for rancher/coredns-coredns:1.1.2 not found
ben@ben-vm:~$ docker pull rancher/prom-prometheus:v2.3.1
Error response from daemon: manifest for rancher/prom-prometheus:v2.3.1 not found
```

Pulling from quay.io works for cert manager though, the image doesn't seem to exist on docker hub

```
ben@ben-vm:~$ docker pull quay.io/jetstack/cert-manager-controller:v0.6.2
v0.6.2: Pulling from jetstack/cert-manager-controller
78b595315e96: Pull complete 
2821a18ecc1f: Pull complete 
e506b56eda00: Pull complete 
3312e44d1788: Pull complete 
Digest: sha256:dab4def4ccb856dec0f62bdf96d2c3c9bbe17b8d569ef3f51c9a06b28db7a96a
Status: Downloaded newer image for quay.io/jetstack/cert-manager-controller:v0.6.2
```

The repos for rancher/coredns-coredns and prom-prometheus exist, but those tags aren't up.

wweldon:
When can we expect a fix for this?  We were hoping to get the 2.2.4 version in to address the 2.2.3 CVE .

alena1108:
@wweldon @bentastic27 @dengtuoyu we are looking into the fix. For now, you can remove all 3 images from rancher-images.txt file as non of them are required.

[...]

PR #20732, opened 2019-06-07
Read branch from the var for rancher images generation

alena1108:
https://github.com/rancher/rancher/issues/20696


rmweir:
LGTM

Issue #17885, opened 2019-02-06
Unable to set http_proxy config on Rancher notifier


garyellis:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

1. Create a slack notifier from Rancher server. (The notifier type is arbitrary)
2. Edit any alert  and set the alert to the notifier created in the previous step; wait for cattle-alertmanager resources to deploy in the current cluster context.
3. Edit the alertmanager secret and add the proxy URL under: `global.http_config.proxy_url`; alerts will work temporarily, but this is not a permanent nor a working solution. 

**Result:** 
alertmanager secret is reverted and alert send fails.

**Other details that may be helpful:**

Alertmanager configures http proxy in config exclusively. `http_proxy`, `https_proxy`, etc environment is not supported in the current version of alertmanager by design. 

See:
https://github.com/prometheus/alertmanager/issues/853
https://github.com/prometheus/alertmanager/pull/1222
https://github.com/prometheus/alertmanager/issues/1300

Alertmanager does not support multiple configuration files. A custom configuration file mounted is not an option.
See:
https://github.com/prometheus/alertmanager/issues/1150

To support http proxies, we need to allow users to configure `http_config.proxy_url per alert and/or at the global level or do something more generic like parameterize an empty http_config map for each alert.
See
https://prometheus.io/docs/alerting/configuration/

**Environment information**
- Rancher version: 2.1.6
- Single and HA installation

panho66:
slack_configs:
     - channel: application
       send_resolved: true
       text: "{{ range .Alerts }}{{ .Annotations.description }}\n{{ end }}"
       http_config:
         tls_config:
            insecure_skip_verify: YES
         proxy_url: http://proxy:8080


Please add insecure_skip_verify in tls_config as well. For some unknown reason, one of the environment I have need to skip tls in order to work around problem.

alena1108:
@aiwantaozi When we fix this issue, we have to make sure it is respected by all the notifiers, including the webhook

panho66:
Need a flexible option please, Ie proxy setting for selected notifiers.

aiwantaozi:
@alena1108 Hi Alena, backend pr will be ready today, then we going to integrate with UI. http_proxy could be added to slack, pageduty, webhook, wechat, except email, alertmanager doesn't allow to set smtp proxy.

bmdepesa:
Was able to reproduce and see the workaround working. 

[...]

PR #20742, opened 2019-06-10
[Alerting] Support set http_proxy config on Rancher notifier

aiwantaozi:
Problem:
Not support http_proxy setting now
Solution:
Add http_proxy configure for slack, pagerduty, wechat, webhook

Issue:
https://github.com/rancher/rancher/issues/17885

Types:
https://github.com/rancher/types/pull/855

Test:
https://github.com/rancher/rancher/pull/20743

### DIFF
--- a/pkg/controllers/user/alert/config/notifiers.go
+++ b/pkg/controllers/user/alert/config/notifiers.go
@@ -1,7 +1,9 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 )
@@ -173,6 +175,8 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type PagerdutyConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
 
+	HTTPConfig *HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
+
 	ServiceKey  Secret            `yaml:"service_key,omitempty" json:"service_key,omitempty"`
 	URL         string            `yaml:"url,omitempty" json:"url,omitempty"`
 	Client      string            `yaml:"client,omitempty" json:"client,omitempty"`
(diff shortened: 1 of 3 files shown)


orangedeng:
LGTM

pkg/controllers/user/alert/config/notifiers.go
```diff
@@ -430,3 +446,107 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	}
 	return checkOverflow(c.XXX, "pushover config")
 }
+
+// HTTPClientConfig configures an HTTP client.
+type HTTPClientConfig struct {
```
mrajashree: This and TLSConfig type and the method to generate http config from that seem to be exactly like the ones in [alertmanager](https://github.com/prometheus/alertmanager/pull/1222/files#diff-fbe65e96815b1559a386cd8521b5c79eR66) repo. Do you think we can vendor in and use them?

pkg/controllers/user/alert/configsyncer/configsyncer.go
```diff
@@ -461,6 +462,17 @@ func (d *ConfigSyncer) addRecipients(notifiers []*v3.Notifier, receiver *alertco
 					ServiceKey:  alertconfig.Secret(notifier.Spec.PagerdutyConfig.ServiceKey),
 					Description: `{{ template "rancher.title" . }}`,
 				}
+
+				if notifier.Spec.PagerdutyConfig.HTTPClientConfig != nil {
+					url, err := toAlertManagerURL(notifier.Spec.PagerdutyConfig.HTTPClientConfig.ProxyURL)
+					if err != nil {
+						logrus.Errorf("Parse pageduty proxy url %s failed, %v", notifier.Spec.PagerdutyConfig.HTTPClientConfig.ProxyURL, err)
```
mrajashree: nit: there's a typo in pagerduty spelling in this log (`pageduty`). Also can we change the error message to something like "Failed to parse pagerduty proxy url", for this and the other notifiers as well?

pkg/controllers/user/alert/config/notifiers.go
aiwantaozi: Hi @mrajashree alertmanager also include other go package in that notifier and config types, like github.com/go-kit/kit, we need to vendor alertmanager, go-kit, and the package go-kit dependence, so we directly copy the go struct. If we really want to vendor alertmanager, we can update in a separate pr, this PR should be easier to test.

pkg/controllers/user/alert/configsyncer/configsyncer.go
aiwantaozi: Thanks, error message updated

Issue #20721, opened 2019-06-07
[v2.2.4] Openstack cluster creating failed: ssh private key is filled with the path


antoine-rehm:
Version: v2.2.4

What kind of request is this (question/bug/enhancement/feature request):
Bug

Steps to reproduce (least amount of steps as possible):

Create a new cluster with openstack cloudprovider.
Waiting for instance creation.
The cluster creation failed due to waiting too long for ssh avability.

To understand what is going wrong, I try to download instance keys through rancher. It appears that the id_rsa is filled with the file path instead of the content of the file.

rollback to 2.2.3 corrected the issue.

sangeethah:
@antoine-rehm When creating openstack cluster, are you providing a privateKeyFile in the node template which points to a file location ?

antoine-rehm:
@sangeethah Yes, I provided an absolute path in the privateKeyFile it's work in rancher 2.2.3 but not in 2.2.4.
If i filled the privateKeyFile input with some random text, the instance keys zip have a id_rsa file with this random text.

sangeethah:
@antoine-rehm In rancher server version - v2.2.4 , we have changed all  file path related fields to content based fields in node template drivers. As part of this fix , privateKeyFile field was modified to start accepting actual contents of the private keys instead of an absolute file path containing the key content. We will have this change documented . The description field had been changed to reflect this change but we will have better documentation around the changes being made.
![image (1)](https://user-images.githubusercontent.com/4266958/59124993-f3da4f80-8915-11e9-8253-e7bf79a7b2e6.png)

If you have the actual key contents entered in privateKeyFile, you should be able to get the nodes to provision successfully.

antoine-rehm:
@sangeethah Ok, thanks you for the explanation about the new input form.
But to make it work I tried to simply paste the content of the privatekey but it didn't work. 
Next, I tried to make it one line it didn't work.
By now, the only solution that work is to edit the value through the api and replace new line with \n (if you paste \n in the form the value become \\n ).

cjellick:
We're working on a fix that will make this multi-line

cjellick:
@antoine-rehm what version of openstack are you using? I'm trying to test Rancher's openstack integration and need this as a data point.

cjellick:
@sangeethah this should have been moved over to test awhile ago

sangeethah:
UI support for Multiline fields has been validated as part of - https://github.com/rancher/rancher/issues/20789

[...]

PR #20751, opened 2019-06-10
fix invalied full path of alias path

whywaita:
This PR fix saved the path of SSH private key.

### before
- when devMode, save under 
  - key save under tempfile (`/tmp/~~~`)
  - config save under tempfile (`/tmp/~~~`)
- when not devMode (no option)
  - key save under jail (`/opt/jail/~~~`)
  - config save under root (`/~~~`)

### after
- when devMode
  - key save under tempfile (`/tmp/~~~`)
  - config save under tempfile (`/tmp/~~~`)
- when not devMode (no option)
  - key save under jail (`/opt/jail/~~~`)
  - config save under root (`/opt/jail/~~~`)


whywaita:
Hi, contributors!

Sorry, I don't understand why to fail CI.
What should I do?

dramich:
@whywaita The build is failing because this change is causing the tests to fail. This change is not correct as it would put the full path including the jail, for every file which is not the intended behavior. Do you have an issue this is related to?

whywaita:
sorry, I don't make related issue yet 🙇 

when not devMode, `config[driverField]` is not a correct file path. (like `/~~~`), correct path is under jail (`/opt/jail/~~~`).

whywaita:
@dramich fix test 💪 

please check me again 🙏

dramich:
@whywaita This change will break a real Rancher container. Dev mode is to enable local development and pathing of files and their locations. When not running in dev mode docker-machine is executed within a 'jail' which is why the path changes based off the dev mode flag. When running inside the jail docker-machine can not see the path for the tmp dir of the system as we build a new root using chroot which is why the files path given to the docker-machine command line args have `/` as the beginning of the path. We are writing these files to the base of the jail dir as seen by docker-machine.

whywaita:
I fixed this problem, The reason is https://github.com/rancher/rancher/issues/20721 🙇 .

I will close this PR 🙇

PR #20762, opened 2019-06-10
Add annotation on nodedriver for UI

dramich:
Problem:
Fields that were converted to take contents instead of a path are
showing as an input box but the UI doesn't know what fields to handle
differently.

Solution:
Add an annotation on the nodedriver to specify the fields and type the
UI should handle differently


cjellick:
LGTM

Issue #20496, opened 2019-05-28
[Windows] Removed nodes will try to add itself back after restart


loganhz:
rancher/rancher:v2.3.0-alpha3

Steps:

1. Create a Windows cluster
2. Add Windows node
3. Remove the node in Rancher UI
4. Restart the Windows node

Results:

The removed node will try to add itself back. 

![image](https://user-images.githubusercontent.com/21168270/58469229-ad486080-8171-11e9-97ed-e42e4f5d56e9.png)

Expected:

It should not try to add itself back

sowmyav27:
reproduced the issue on `v2.3.0-alpha3`
**Steps:**

1. Create a Windows cluster
2. Add Windows node
3. Remove the node in Rancher UI
4. Restart the Windows node

**Results:**

The removed node will try to add itself back.
And gets added successfully.

<img width="1452" alt="Screen Shot 2019-07-01 at 2 34 42 PM" src="https://user-images.githubusercontent.com/26032343/60468798-a2b93500-9c0f-11e9-9c59-971ef3cd5f34.png">

verified on `rancher:master-head` and works fine.
**Steps:**

1. Create a Windows cluster
2. Add Windows node
3. Remove the node in Rancher UI
4. Restart the Windows node

**Results:**
The removed node will not add itself back.

PR #20770, opened 2019-06-11
[Windows] Fix removed Windows node to re-add after restarting

thxCode:
**Problem:**
Add back after restarting the removed Windows node

**Solution:**
Return ErrNodeOrClusterNotFound error if could not found the node or cluster

**Issue:**
https://github.com/rancher/rancher/issues/20496

**result:**
- Windows agent will stop itself after removing from UI:
    image
- Restart the removed agent:
    image

### DIFF
--- a/pkg/agent/main_windows.go
+++ b/pkg/agent/main_windows.go
@@ -94,16 +94,12 @@ func getTokenAndURL() (string, string, error) {
 }
 
 func isConnected() bool {
-	_, err := os.Stat("connected")
+	_, err := os.Stat(`C:\etc\rancher\connected`)
 	return err == nil
 }
 
-func resetConnected() {
-	os.RemoveAll("connected")
-}
-
 func connected() {
-	f, _ := os.Create("connected")
+	f, _ := os.Create(`C:\etc\rancher\connected`)
 	defer f.Close()
 }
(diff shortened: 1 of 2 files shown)


pkg/agent/main_windows.go
```diff
@@ -94,17 +94,15 @@ func getTokenAndURL() (string, string, error) {
...
-func resetConnected() {
-	os.RemoveAll("connected")
-}
-
 func connected() {
-	f, _ := os.Create("connected")
-	defer f.Close()
+	f, err := os.Create(`C:\etc\rancher\connected`)
+	if err != nil {
+		f.Close()
```
orangedeng: Why do you only close the file pointer when err is not nil? Should we close it anyway?

thxCode: Rolled back, don't need to judge like this way, `File` has been processed internally:
```
// Close closes the File, rendering it unusable for I/O.
// On files that support SetDeadline, any pending I/O operations will
// be canceled and return immediately with an error.
func (file *File) Close() error {
	if file == nil {
		return ErrInvalid
	}
	return file.file.close()
}
```

orangedeng:
LGTM

Issue #20693, opened 2019-06-06
HPA stuck in pending state when HPA is created with metrics type pod/object/external


soumyalj:
Version: Master

Steps:
1. Create a deployment in a RKE cluster
2. Create a HPA pointing to a the deployment for the metrics type Pod/Object/External

The HPA is stuck in pending state with the below error in API
```
"the HPA was unable to compute the replica count: unable to get metric testpod: unable to fetch metrics from custom metrics API: no custom metrics API (custom.metrics.k8s.io) registered"

```

The custom metrics server associated with pod/object/external types is not available by default . 
Should these metrics types be disabled in UI ?

loganhz:
It looks like there is no way to detect if users deployed the custom metrics server

loganhz:
I think the error message in the HPA transition message helps?
I don't think we should disable it by default if we cannot detect custom metrics.

loganhz:
`The HPA was unable to compute the replica count: unable to get metric sadf: unable to fetch metrics from custom metrics API: no custom metrics API (custom.metrics.k8s.io) registered`

@orangedeng is there a way we can check if custom metrics API (custom.metrics.k8s.io) registered or not?

If we can do that, we disable it in UI when it is not registered

orangedeng:
Yes, but we need to introduce `apiservice.apiregistration.k8s.io` resource into Rancher API to do it.

loganhz:
We will add a warning for preview release first. https://github.com/rancher/rancher/issues/20866
But we should remove it once we get this issue fixed

loganhz:
For any cluster, check what version of HPA API is being used. List the version that's being used and that Rancher's requirements is v2beta2 (or whatever it is.

loganhz:
Added the warning for HPA created with metrics type pod/object/external.

soumyalj:
Tested with master-head build from Aug 1st. 
Create a deployment in a RKE cluster
Create a HPA pointing to a the deployment for the metrics type Pod/Object/External
During the HPA creation, there is a warning :

<img width="1283" alt="Screen Shot 2019-08-01 at 6 12 43 PM" src="https://user-images.githubusercontent.com/18536626/62294265-2d4bb900-b488-11e9-9d70-1ca798fef0bf.png">

<img width="1249" alt="Screen Shot 2019-08-01 at 6 14 44 PM" src="https://user-images.githubusercontent.com/18536626/62294331-50766880-b488-11e9-8547-90f8fe552c37.png">

[...]

PR #20775, opened 2019-06-11
Introduce APIService resource to Rancher

orangedeng:
**Problem:**
The custom metrics server associated with `pod/object/external`
types is not available by default, we need a way to tell UI we don't
support those kind of metrics for now.

**Solution:**
Introduce `apiservice.apiregistration.k8s.io` resource into Rancher.
This API can tell which api service is current available.

Related issue:
https://github.com/rancher/rancher/issues/20693

Related PR:
https://github.com/rancher/types/pull/857


orangedeng:
Update vendor and rebase to master

gitlawr:
LGTM

PR #20785, opened 2019-06-11
use driver metadata

kinarashah:
https://github.com/rancher/rancher/issues/20369
https://github.com/rancher/rancher/issues/21221 

https://github.com/rancher/kontainer-engine/pull/162


pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -0,0 +1,154 @@
...
+	for range ticker.Context(ctx, checkInterval) {
+		logrus.Infof("driverMetadata: checking rke-metadata-url every %v", checkInterval)
+		m.refresh(url)
+	}
+}
+
+func (m *MetadataController) refresh(url string) {
+	resp, err := http.Get(url)
+	if err != nil {
+		logrus.Errorf("driverMetadata err %v", err)
```
kinarashah: todo: on startup, add logic to load from vendor data if this errors out (for air gap setups)

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -0,0 +1,420 @@
...
+	k8sVersionSvcOptions := map[string]v3.KubernetesServicesOptions{}
+
+	for majorVersion, k8sVersion := range maxVersionForMajorK8sVersion {
+		k8sVersionRKESystemImages[k8sVersion] = nil
+		k8sVersionSvcOptions[k8sVersion] = data.K8sVersionServiceOptions[majorVersion]
+	}
+	return SaveSettings(k8sVersionRKESystemImages, k8sVersionSvcOptions, data.RancherDefaultK8sVersions, rancherVersion)
+}
+
+func SaveSettings(k8sCurrVersions map[string]interface{},
```
kinarashah: todo: update ui-settings as well

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -0,0 +1,420 @@
...
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	APIVersion                = "management.cattle.io/v3"
+	RancherVersionDev         = "2.3"
+	sendRKELabel              = "send"
```
kinarashah: todo: change label name to something sensible

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -0,0 +1,420 @@
...
+	maxVersionForMajorK8sVersion := map[string]string{}
+	rancherVersion := settings.ServerVersion.Get()
+	if rancherVersion == "dev" || rancherVersion == "master" {
+		rancherVersion = RancherVersionDev
+	}
+	for k8sVersion, systemImages := range data.K8sVersionRKESystemImages {
+		rancherVersionInfo, ok := data.K8sVersionInfo[k8sVersion]
+		if ok {
+			if rancherVersionInfo.DeprecateRancherVersion != "" && rancherVersion >= rancherVersionInfo.DeprecateRancherVersion {
+				// don't store
```
alena1108: do you think it's worth adding a debug statements to skip blocks?

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -0,0 +1,420 @@
...
+		}
+		return nil
+
+	}
+	if reflect.DeepEqual(sysImage.SystemImages, systemImages) {
+		return nil
+	}
+	sysImageCopy := sysImage.DeepCopy()
+	sysImageCopy.SystemImages = systemImages
+	// todo: add retry
```
alena1108: are we going to address dodo as a part of the original pr?

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -0,0 +1,181 @@
...
+	k8sVersionRKESystemImages := map[string]v3.RKESystemImages{}
+	k8sVersionWinSystemImages := map[string]v3.WindowsSystemImages{}
+	k8sVersionSvcOptions := map[string]v3.KubernetesServicesOptions{}
+
+	//rancherVersion := getRancherVersion()
+	maxVersionForMajorK8sVersion := map[string]string{}
+	for k8sVersion := range rkeSysImages {
+		if rancherVersionInfo, ok := rancherVersions[k8sVersion]; ok {
+			greaterThanMax := rancherVersionInfo.MaxRancherVersion != "" && rancherVersion > rancherVersionInfo.MaxRancherVersion
+			lowerThanMin := rancherVersionInfo.MinRancherVersion != "" && rancherVersion < rancherVersionInfo.MinRancherVersion
```
prachidamle: Use semver for version comparisons?

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -0,0 +1,181 @@
...
+			if greaterThanMax || lowerThanMin {
+				continue
+			}
+		}
+		majorVersion := util.GetTagMajorVersion(k8sVersion)
+		if curr, ok := maxVersionForMajorK8sVersion[majorVersion]; !ok || k8sVersion > curr {
```
prachidamle: semver comparison would be ideal

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
prachidamle: Shouldn't we also check DeprecateRancherVersion here?

pkg/controllers/management/clusterprovisioner/rke_store.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		if template != "" {
+			data[addonName] = template
+		}
+	}
+	return data
+}
+
+func (a *rkeStore) GetServiceOptions(k8sVersion string) v3.KubernetesServicesOptions {
+	logrus.Infof("KINARA entered getting service options; getting service options; getting service options;")
```
prachidamle: Need to remove this log statement :)

pkg/controllers/management/clusterprovisioner/rke_store.go
```diff
@@ -0,0 +1,60 @@
...
+	svcOptions, err := kd.GetRKEK8sServiceOptions(k8sVersion, a.SvcOptionLister, a.SvcOptions)
+	if err != nil {
+		logrus.Errorf("getK8sServiceOptions: k8sVersion %s [%v]", k8sVersion, err)
+	}
+	logrus.Infof("KINARA svcOptions %v", svcOptions)
```
prachidamle: same with this log

app/setting.go
```diff
@@ -2,9 +2,33 @@ package app
...
+	if rancherVersion == "dev" || rancherVersion == "master" {
+		rancherVersion = kd.RancherVersionDev
+	}
+	k8sCurrVersions := map[string]interface{}{}
+
+	k8sVersionToRKESystemImages, _, k8sVersionToSvcOptions := kd.GetK8sVersionInfo(rancherVersion,
+		metadata.DriverData.K8sVersionRKESystemImages,
+		metadata.DriverData.K8sVersionWindowsSystemImages,
+		metadata.DriverData.K8sVersionServiceOptions,
+		metadata.DriverData.K8sVersionInfo)
```
prachidamle: we seem to be doing this during server start twice, (main.go and addSetting) - maybe we could cache the results of kd.GetK8sVersionInfo for the current rancherVersion?

kinarashah: main.go is `./pkg/image/export/main.go`, which is called only when building dapper, and it prints out the images ... but not used in actual run

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
kinarashah: > Shouldn't we also check DeprecateRancherVersion here?

yeah, forgot adding latest changes here, thanks

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
kinarashah: updating to use semver for comparing rancher version info, but using string comparison for finding max since it doesn't respect semantic versioning

checked in 1.6, seems like we did something similar there too 
```
                long currentMaxVersion = Long.valueOf(url.substring(url.lastIndexOf(":") + 1, url.length()));
                if (currentMaxVersion >= maxVersion) {
                    maxVersion = currentMaxVersion;
```

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	return sysImage.SystemImages, err
+}
+
+func GetRKEAddonTemplate(k8sVersion string, addonName string, addonLister v3.RKEAddonLister, addons v3.RKEAddonInterface) (string, error) {
+	names := []string{
+		fmt.Sprintf("%s-%s", strings.ToLower(addonName), getName(k8sVersion)),
+		fmt.Sprintf("%s-%s", strings.ToLower(addonName), getName(util.GetTagMajorVersion(k8sVersion))),
+		fmt.Sprintf("%s-%s", strings.ToLower(addonName), "default"),
+	}
```
prachidamle: Why do we have 3 name formats? where do we add the addon crds?

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -0,0 +1,164 @@
...
+func loadData(url string) (Data, error) {
+	var data Data
+	resp, err := http.Get(url)
+	if err != nil {
+		return data, fmt.Errorf("driverMetadata err %v", err)
+	}
+	if resp.StatusCode != 200 {
+		return data, fmt.Errorf("driverMetadata statusCode %v", resp.StatusCode)
+	}
+	defer resp.Body.Close()
```
prachidamle: If resp.StatusCode!= 200, we would return without closing resp.Body; we should move this statement up

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -0,0 +1,164 @@
...
+
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(resp.Body)
```
prachidamle: handle error returned by buf.ReadFrom(resp.Body)?

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
kinarashah: adding in `createOrUpdateAddonCRD`, https://github.com/rancher/rancher/pull/20785/files#diff-e282be4392d4256b319d411aa68fbd79R310 

this is the preference order, if minor version specific template is available, we should use it, followed by major version and then last default.

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	for range ticker.Context(ctx, checkInterval) {
+		logrus.Infof("driverMetadata: checking rke-metadata-url every %v", checkInterval)
+		m.refresh(url)
+	}
+}
+
+func (m *MetadataController) refresh(url string) {
+	data, err := loadData(url)
+	if err != nil {
+		logrus.Errorf("error loading rke data, using stored defaults %v", err)
```
prachidamle: Do we need to even load anything from vendor if data has been synced once. I would say just keep using what has been synced. And not load older vendor on subsequent failures. 
- use vendor on first create only
- on every refresh pull from repo, if it fails do not reload from vendor

prachidamle:
LGTM

Issue #20796, opened 2019-06-11
Unable to add node driver 


daxmc99:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug QA (found on e0a97f4bbb40b1e3f884a86596183c29a40ba24c) 

**Steps to reproduce (least amount of steps as possible):**
Try to add a custom driver via the UI
Ex: the linode driver 
`https://github.com/linode/docker-machine-driver-linode/releases/download/v0.1.8/docker-machine-driver-linode_linux-amd64.zip`
UI: `https://linode.github.io/rancher-ui-driver-linode/releases/v0.2.0/component.js`

**Result:**

```
2019/06/11 20:26:41 [INFO] Copying management-state/machine-drivers/18206b37b5529c1138ddc9fe568a803b03ae3c39ebb8abe4ae1934aef49ba4bc-docker-machine-driver-linode => /opt/drivers/management-state/bin/docker-machine-driver-linode-tmp
2019/06/11 20:26:41 [ERROR] NodeDriverController nd-md7mk [node-driver-controller] failed with : Driver "" not found. Do you have the plugin binary accessible in your PATH?
```
**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): e0a97f4bbb40b1e3f884a86596183c29a40ba24c 
- Installation option (single install/HA):
Docker container and local

alena1108:
Will be available in v2.2.5-rc2

sowmyav27:
Verified on v2.2.5-rc2

Able to add custom driver via UI - Linode
the linode driver
`https://github.com/linode/docker-machine-driver-linode/releases/download/v0.1.8/docker-machine-driver-linode_linux-amd64.zip`
UI: `https://linode.github.io/rancher-ui-driver-linode/releases/v0.2.0/component.js`

Logs:
```
2019/06/13 21:07:19 [INFO] Download https://github.com/linode/docker-machine-driver-linode/releases/download/v0.1.8/docker-machine-driver-linode_linux-amd64.zip
2019/06/13 21:07:20 [INFO] Found driver docker-machine-driver-linode
2019/06/13 21:07:20 [INFO] Copying management-state/machine-drivers/18206b37b5529c1138ddc9fe568a803b03ae3c39ebb8abe4ae1934aef49ba4bc-docker-machine-driver-linode => /opt/drivers/management-state/bin/docker-machine-driver-linode-tmp
```
UI:
[...]

PR #20798, opened 2019-06-11
Refactor NodeDriver download

dramich:
Problem:
driver name is not always known before the driver is staged, this can cause path errors 

Solution:
Move the driver name creation back to after the driver has been staged and move out nodeDriver object updates until after the driver is successfully installed

Issue: https://github.com/rancher/rancher/issues/20796


cjellick:
LGTM

PR #20816, opened 2019-06-12
Add SMTX OS machine driver

fengye87:


fengye87:
Bump. Any feedback please?

fengye87:
Bump again. Any feedback please?

fengye87:
Bump again. This is a small and easy patch to just add a new driver to the node driver list. I'd be glad to get any feedback from you and make some proper changes to make the patch mergable.

vincent99:
The difficulty here is not with the size of the PR, but that it's for a product with no apparent way to use it other than requesting a demo.

So that leaves me with two concerns:

- There's no way for us to test that the driver is at least basically functional, which is I think a pretty low minimum bar we'd like to see before packaging a new one.

- We have enterprise customers too, but the vast majority users are the open-source community. 
 Are any/enough people going to have an installation of your (presumably high-cost/high-touch sales) enterprise product available to justify it being included?

fengye87:
For your first concern about basic functionality, we could provide an ISO of our product for you to test. Or, if that was too much trouble, we could provide some screen shots. Or maybe a live demo via Zoom? Any way would work for us, as long as it works for you.

For you second concern, we happen to be one of the main stream HCI vendors in China. We do have many enterprise customers in China, who would benefit a lot from this driver when they move to K8s and Rancher. Moreover, we are a strategical partner of Rancher China, this driver would ease the adoption of Rancher in our enterprise customers, and vice versa.

vincent99:
@loganhz Can you or someone there work with @fengye87 to test this?

fengye87:
@loganhz Any way you think we could make this work?

loganhz:
@fengye87  Let's talk about it via wechat or email. Can you loop George and Peng in the email, please?

Issue #20869, opened 2019-06-13
Flaky test_create_cluster_with_template 


kinarashah:
rancher/master 06/12

https://drone-pr.rancher.io/rancher/rancher/2507/1/2

```
==================================== ERRORS ====================================
____________ ERROR at teardown of test_create_cluster_with_template ____________
[gw0] linux -- Python 3.7.1 /go/src/github.com/rancher/rancher/tests/integration/.tox/py37/bin/python3.7
tp = <class 'rancher.ApiError'>, value = None, tb = None
def reraise(tp, value, tb=None):
try:
if value is None:
value = tp()
if value.__traceback__ is not tb:
raise value.with_traceback(tb)
> raise value
../.tox/py37/lib/python3.7/site-packages/six.py:693:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../.tox/py37/lib/python3.7/site-packages/six.py:693: in reraise
raise value
../.tox/py37/lib/python3.7/site-packages/six.py:693: in reraise
raise value
conftest.py:327: in clean
raise e
conftest.py:324: in clean
client.delete(resource)
../.tox/py37/lib/python3.7/site-packages/rancher.py:467: in delete
return self._delete(i.links.self)
../.tox/py37/lib/python3.7/site-packages/rancher.py:68: in wrapped
return fn(*args, **kw)
../.tox/py37/lib/python3.7/site-packages/rancher.py:336: in _delete
self._error(r.text)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <rancher.Client object at 0x7fd9463fa198>
text = '{"baseType":"error","code":"PermissionDenied","message":"Cannot delete the clusterTemplate until Clusters referring it are removed","status":403,"type":"error"}\n'
def _error(self, text):
> raise ApiError(self._unmarshall(text))
E rancher.ApiError: (ApiError(...), "PermissionDenied : Cannot delete the clusterTemplate until Clusters referring it are removed\n\t{'baseType': 'error', 'code': 'PermissionDenied', 'message': 'Cannot delete the clusterTemplate until Clusters referring it are removed', 'status': 403, 'type': 'error'}")
```

daxmc99:
[...]

PR #20875, opened 2019-06-14
Fix the test that can fail due to pending cluster delete

prachidamle:
Fixing the flaky test that can fail if a cluster delete remains pending, then removal of the cluster_template will error out during teardown of the test.

https://github.com/rancher/rancher/issues/20869

### DIFF
--- a/tests/integration/suite/test_clustertemplate.py
+++ b/tests/integration/suite/test_clustertemplate.py
@@ -1,6 +1,7 @@
 from .common import random_str
 from rancher import ApiError
 import pytest
+import time
 
 
 def test_create_cluster_template_with_revision(admin_mc, remove_resource):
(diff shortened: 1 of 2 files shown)


tests/integration/MANIFEST
```diff
@@ -1,39 +0,0 @@
-# file GENERATED by distutils, do NOT edit
```
alena1108: if we don't need this file, and it got accidentally checked in, lets add it to gitignore

tests/integration/suite/test_clustertemplate.py
```diff
@@ -46,7 +47,8 @@ def test_create_cluster_with_template(admin_mc, remove_resource):
         client.delete(cluster_template)
         assert e.value.error.status is None
 
-    remove_resource(cluster)
```
mrajashree: We can still leave the remove_resource call immediately after cluster creation in case anything in `wait_for_cluster_to_be_deleted` fails.

Issue #19103, opened 2019-03-22
RKE clusters: show auto generated certificates' expiration info in API/UI


alena1108:
Once support is added for it on RKE side https://github.com/rancher/rke/issues/1220, we have to surface it or implement similar logic in Rancher UI/API.

In 2.2, the check for cert expiry will be done via script.

Consider the issue for 2.2.1.

dnoland1:
As a nice to have, if the certificate is near expiration (maybe 30 days), can this be called out in a big red obvious font on the dashboard?

alena1108:
@dnoland1 should be doable. As long as UI has the info about certs expiration, it can generate the warning.

gburton1:
Would this include the ability to configure an alert/notifier based on time remaining before expiration?

Oats87:
@alena1108 @moelsayed does this show up for users that are not cluster owners (and thus don’t have the capability to rotate the certificates themselves)?

sowmyav27:
**Rancher server with custom 2.2.4 image --> upgrade to rancher:master-head --> wait for the certs to expire**

1. create a cluster in custom 2.2.4 image 
2. upgrade to rancher:master-head.
3. View in API does have certificatesExpiration block
4. expiry would be 30 minutes from the time of creation of cluster
5. wait for the cluster certs to expire.
6. Do Rotate Certificates.
7. The cluster will go into Updating state and will come up Active.
8. The cluster will have the certificatesExpiration in View in API. and the expiration date would now be close to 10 years from the time the action was performed.

**Fresh install of rancher:master-head**

1. Create an AWS cluster - 3 nodes - 1etcd, 1 control plane, 1 worker node
2. View in API has the certificatesExpiration block with the expiration 10 year from cluster creation time.
3. Perform a Rotate Certificates for all services separately and verify every one of them changes when you look at View in API.

PR #20902, opened 2019-06-15
Add CertificatesExpiration to cluster status 

moelsayed:
rancher/rancher#19103


pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,176 @@
...
+		certsExpInfo[certName] = info
+	}
+	cluster.Status.CertificatesExpiration = certsExpInfo
+	return c.ClusterClient.Update(cluster)
+
+}
+
+func (c Controller) getClusterCertificateBundle(clusterName string) (map[string]pki.CertificatePKI, error) {
+	// cluster has a state file ?
+	store := clusterprovisioner.NewPersistentStore(c.NamespaceClient, c.CoreV1)
```
kinarashah: why are we getting a new store every time? would it help to just have it in controller definition?

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+	for certName, certObj := range certBundle {
+		info, err := getCertExpiration(certObj.CertificatePEM)
+		if err != nil {
+			continue
+		}
+		certsExpInfo[certName] = info
+	}
+	cluster.Status.CertificatesExpiration = certsExpInfo
```
kinarashah: we should update this only if certsExpInfo has changed?

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,176 @@
...
+	return certs, nil
+}
+
+func getRKECurrentStateFromStore(store cluster.PersistentStore, clusterName string) (*rkecluster.State, error) {
+	cluster, err := store.Get(clusterName)
+	if err != nil {
+		return nil, err
+	}
+	var fullState rkecluster.FullState
+	stateStr, ok := cluster.Metadata["fullState"]
```
kinarashah: just a question, when will `fullState` not be present? in that case, do we need to look for it elsewhere?

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &fullState.CurrentState, nil
+}
+
+func getCertsFromUserCluster(c *config.UserContext) (map[string]pki.CertificatePKI, error) {
+	certs := map[string]pki.CertificatePKI{}
+
+	secrets, err := c.Core.Secrets("kube-system").List(metav1.ListOptions{})
```
kinarashah: can we use lister here?

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, s := range secrets.Items {
+
+		if strings.HasPrefix(s.GetName(), "kube-") &&
+			s.Type == corev1.SecretTypeOpaque {
+			name := s.GetName()
+			secret, err := c.Core.Secrets("kube-system").Get(name, metav1.GetOptions{})
```
kinarashah: same here, Controller.Lister.Get

pkg/controllers/management/certsexpiration/certsexpiration.go
kinarashah: also deepcopy before update

pkg/controllers/management/certsexpiration/certsexpiration.go
moelsayed: No, this is only available in 2.2.x clusters. If it's not found, it's 2.0.x or 2.1.x cluster, in that case we get the certificate bundle from the user cluster directly.

pkg/controllers/management/certsexpiration/certsexpiration.go
moelsayed: Unfortunately, no. It's a UserContext, in this case we hit the user cluster directly.

Issue #20894, opened 2019-06-14
Support Istio Virtual Service and Destination Rule as first class resources


loganhz:
Support Istio Virtual Service and Destination Rule as first class resources in Rancher UI just like Workloads.

We should use https://github.com/rancher/pkg/tree/rancher/apis/istio/v1alpha3 for types

loganhz:
UI is not merged. So we can only test it via API in master build now

bmdepesa:
rancher/rancher:master-head (8/9)

* Creating virtual service and destination rules work well through UI
* Functional checks:
  * Created a service using samples from istio test plan
  * Created canary deployment using virtual service and destination rules
  * Crated basic round robin deployment using virtual service and destination rules
* Checked CRDs after editing each field in the UI and validated that updates made from the UI are immediately reflected (since we are not able to easily test all Istio functionality)

Issue #17567, opened 2019-01-24
Failing test_cluster_catalog_templates_access tests


gitlawr:
**Rancher Version**
master 1/23

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Build fails randomly on Drone (https://drone7.rancher.io/rancher/rancher/1641/3)

```
=================================== FAILURES ===================================
____________________ test_cluster_catalog_templates_access _____________________
[gw8] linux -- Python 3.7.1 /go/src/github.com/rancher/rancher/tests/.tox/py37/bin/python3.7

admin_mc = <core.conftest.ManagementContext object at 0x7f33f8539278>
user_factory = <function user_factory.<locals>._create_user at 0x7f33f86b1f28>
remove_resource = <function remove_resource.<locals>._cleanup at 0x7f33f867b0d0>
admin_pc = <core.conftest.ProjectContext object at 0x7f33f41d5048>

    def test_cluster_catalog_templates_access(admin_mc, user_factory,
                                              remove_resource, admin_pc):
        # Cluster-owner,cluster-member, and all project-owners/members
        # in that cluster should have access to cluster catalog's templates
    
        # First add a user as cluster member to this cluster
        user1 = user_factory()
        remove_resource(user1)
        admin_client = admin_mc.client
        crtb_member = admin_client.create_cluster_role_template_binding(
            clusterId="local",
            roleTemplateId="cluster-member",
            userId=user1.user.id)
        remove_resource(crtb_member)
    
        wait_until(crtb_cb(admin_client, crtb_member))
    
        # Now create a cluster catalog
        name = random_str()
        catalog_name = "local:" + name
        url = "https://github.com/mrajashree/charts.git"
        cc = admin_client.create_cluster_catalog(name=name,
                                                 branch="onlyOne",
                                                 url=url,
                                                 clusterId="local",
                                                 )
        wait_for_clustercatalog_template_to_be_created(admin_client, catalog_name)
    
        # Now add a user to a project within this cluster as project-owner
        user2 = user_factory()
        remove_resource(user2)
        prtb_owner = admin_client.create_project_role_template_binding(
            userId=user2.user.id,
            roleTemplateId="project-owner",
            projectId=admin_pc.project.id,
        )
        remove_resource(prtb_owner)
        wait_until(prtb_cb(admin_client, prtb_owner))
    
        wait_until_available(admin_client, prtb_owner)
        project_owner_client = user2.client
[...]

PR #20910, opened 2019-06-17
Add Istio VirtualService and DestinationRule to Rancher API

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/20894
Depend on types PR: https://github.com/rancher/types/pull/870

Notes:
For project scoped resources we add `projectId` field in API layer in `https://github.com/rancher/norman/blob/master/api/handler/validate.go`. For native kubernetes resources like `service/deployment` apiserver will drop it in deserialization. But for CRDs like Istio resources, it is stored as part of the manifest, and fails the validation from pilot because it is an undefined field. Therefore we update the projectsetter store to drop the `projectId` field. It is unnecessary because the store reads `projectId` from the namespace object.


orangedeng:
LGTM

gitlawr:
Rebased

Also, encountered a flaky test in CI: https://github.com/rancher/rancher/issues/17567#issuecomment-509885850

orangedeng:
LGTM

cjellick:
merged the types pr. this one is lgtm once types gets updated

loganhz:
Close in favor of https://github.com/rancher/rancher/pull/21484

Issue #20919, opened 2019-06-17
backport Multi-cluster Apps - cannot get revisions


kinarashah:
backport issue for https://github.com/rancher/rancher/issues/20868

alena1108:
Can be tested with `v2.2.5-rc5`

jiaqiluo:
The bug fix is validated on Rancher: v2.2.5-rc6 (upgraded from v2.2.4)

The same steps as the original issue's:
- as a standard user, deploy the multi-cluster app mySQL with an old version v0.3.7
- wait for the app to be available
- upgrade the app to the latest version, v0.3.8
- wait for the app to be available
- click Rollback

Result:
- see the revisions, and can choose one to rollback successfully 
<img width="1000" alt="Screen Shot 2019-06-18 at 11 51 48 AM" src="https://user-images.githubusercontent.com/6218999/59711339-d03dc180-91bf-11e9-938c-c43557571393.png">

PR #20921, opened 2019-06-17
[backport] use creatorID of mcapp for mcapp revision

kinarashah:
https://github.com/rancher/rancher/issues/20919


alena1108:
LGTM

PR #20925, opened 2019-06-17
Changes to set Default Cluster Template Revision

prachidamle:
Corresponding Types PR: https://github.com/rancher/types/pull/872

Changes:
- When the very first Cluster Template Revision is created, we set the ClusterTemplate.DefaultRevisionId if it is not set
- When the Cluster Template Revision is being deleted, we error out incase it is the revision pointed by ClusterTemplate.DefaultRevisionId

### DIFF
--- a/pkg/controllers/management/clustertemplate/revision_controller.go
+++ b/pkg/controllers/management/clustertemplate/revision_controller.go
@@ -1,0 +1,81 @@
+package clustertemplate
+
+import (
+	"context"
+
+	"fmt"
+	"strings"
+
+	"github.com/rancher/rancher/pkg/namespace"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	RevisionController   = "mgmt-cluster-template-revision-controller"
+	clusterTemplateLabel = "io.cattle.field/clusterTemplateId"
+)
+
+type RevController struct {
+	clusterTemplates              v3.ClusterTemplateInterface
+	clusterTemplateLister         v3.ClusterTemplateLister
+	clusterTemplateRevisionLister v3.ClusterTemplateRevisionLister
+}
+
+func newRevController(ctx context.Context, mgmt *config.ManagementContext) *RevController {
+	n := &RevController{
+		clusterTemplates:              mgmt.Management.ClusterTemplates(namespace.GlobalNamespace),
+		clusterTemplateLister:         mgmt.Management.ClusterTemplates(namespace.GlobalNamespace).Controller().Lister(),
+		clusterTemplateRevisionLister: mgmt.Management.ClusterTemplateRevisions(namespace.GlobalNamespace).Controller().Lister(),
+	}
+	return n
+}
+
+func Register(ctx context.Context, management *config.ManagementContext) {
+	n := newRevController(ctx, management)
+	if n != nil {
+		management.Management.ClusterTemplateRevisions("").AddHandler(ctx, RevisionController, n.sync)
+	}
+}
+
+//sync is called periodically and on real updates
+func (n *RevController) sync(key string, obj *v3.ClusterTemplateRevision) (runtime.Object, error) {
+	if obj == nil || obj.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	if obj.Spec.ClusterTemplateName == "" {
+		return nil, nil
+	}
+	//load the template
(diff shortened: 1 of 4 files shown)


pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if obj.Spec.ClusterTemplateName != "" {
+		//load the template
+		split := strings.SplitN(obj.Spec.ClusterTemplateName, ":", 2)
+		if len(split) != 2 {
+			return nil, fmt.Errorf("error in splitting clusterTemplate name %v", obj.Spec.ClusterTemplateName)
+		}
+		templateName := split[1]
+		template, err := n.clusterTemplates.Get(templateName, metav1.GetOptions{})
+		if err != nil {
```
mrajashree: Do we want to return error even if the clustertemplate is not found?

tests/integration/suite/test_clustertemplate.py
```diff
@@ -15,6 +16,19 @@ def test_create_cluster_template_with_revision(admin_mc, remove_resource):
...
+    cluster_template = create_cluster_template(admin_mc, remove_resource)
+    templateId = cluster_template.id
+    first_revision = \
+        create_cluster_template_revision(admin_mc, templateId)
+    client = admin_mc.client
+    wait_for_default_revision(client, templateId, first_revision.id)
+    # delete the cluster template revision, it should error out
+    with pytest.raises(ApiError) as e:
+        client.delete(first_revision)
+        assert e.value.error.status is None
```
mrajashree: If this should error out why is the error none?

tests/integration/suite/test_clustertemplate.py
```diff
@@ -176,3 +189,17 @@ def getRKEConfig():
...
     return rke_config
+
+
+def wait_for_default_revision(client, templateId, revisionId, timeout=45):
+    updated = False
+    interval = 0.5
+    start = time.time()
+    while not updated:
+        if time.time() - start > timeout:
+            raise Exception('Timeout waiting for gdns entry to update')
```
mrajashree: nit: `clustertemplate` in place of `gdns` :)

pkg/controllers/management/clustertemplate/revision_controller.go
prachidamle: yes, since we are trying to update the template's defaultRevision reference

tests/integration/MANIFEST
```diff
@@ -1,39 +0,0 @@
-# file GENERATED by distutils, do NOT edit
```
alena1108: we should add this file to gitignore

pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+//sync is called periodically and on real updates
+func (n *RevController) sync(key string, obj *v3.ClusterTemplateRevision) (runtime.Object, error) {
+	if obj == nil || obj.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	if obj.Spec.ClusterTemplateName != "" {
```
alena1108: nit: we can avoid surrounding a big block of code with if {} by just doing:

```
if obj.Spec.ClusterTemplateName == "" {
return
}
```

pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,85 @@
...
+		//load the template
+		split := strings.SplitN(obj.Spec.ClusterTemplateName, ":", 2)
+		if len(split) != 2 {
+			return nil, fmt.Errorf("error in splitting clusterTemplate name %v", obj.Spec.ClusterTemplateName)
+		}
+		templateName := split[1]
+		template, err := n.clusterTemplates.Get(templateName, metav1.GetOptions{})
```
alena1108: we should fetch from cache

pkg/controllers/management/clustertemplate/revision_controller.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if template.Spec.DefaultRevisionName == "" {
+
+			//if default is not set, set the revision to revision if only one found
+			set := labels.Set(map[string]string{clusterTemplateLabel: template.Spec.DisplayName})
+			revisionList, err := n.clusterTemplateRevisions.List(metav1.ListOptions{LabelSelector: set.AsSelector().String()})
```
alena1108: we should fetch revisions from cache

tests/integration/MANIFEST
prachidamle: yes, this is being added by another PR open

alena1108:
LGTM

PR #20933, opened 2019-06-18
Check app update perms

rmweir:


pkg/api/customization/app/app.go
```diff
@@ -94,14 +94,22 @@ func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiConte
 		return err
 	}
 
+	var appMap map[string]interface{}
+	if err := access.ByID(apiContext, &projectschema.Version, projectv3.AppType, apiContext.ID, &appMap); err != nil {
+		return httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("unable to access app by id: %v", err))
+	}
+
+	if err := apiContext.AccessControl.CanDo(pv3.AppGroupVersionKind.Group, pv3.AppResource.Name, "update", apiContext, appMap, apiContext.Schema); err != nil {
```
kinarashah: q: why are we not using CanUpdate directly?

rmweir: This is my understanding of the CanUpdate vs CanDo after speaking with Craig: CanUpdate returns nil if the schema is generally updatable, while CanDo considers rbac.

kinarashah: oh, didn't know that, thank you!

pkg/api/customization/app/app.go
```diff
@@ -94,14 +94,22 @@ func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiConte
...
+		return httperror.NewAPIError(httperror.PermissionDenied, fmt.Sprintf("user does not have permission to update for action %s", actionName))
+	}
+
 	creatorNotFound := false
```
luthermonson: probably dont need this any more. just move it in to one if kicking the error now that the compound conditional is removed.

rmweir: reference in upgrade and rollback logic

luthermonson:
lgtm

cjellick:
lgtm

Issue #20939, opened 2019-06-18
[Windows] Not allow users to add  some nodes into Windows cluster


loganhz:
rancher/rancher:v2.3.0-alpha3

Windows agent should detect the server info and Docker version and not allow users to add Docker CE nodes into the cluster

Case 1: Docker CE

Case 2: 2 network interfaces without specifying node IP

Case 3: Using a Chinese language version of windows server

kingsd041:
Version: master master-head (8/30/19) 
Followed the same steps as the ticket describes and got the expected result.

Case 1: Docker CE
```
INFO: Detecting running permission ...
INFO: Detecting host CPU reservation ...
INFO: Detecting host RAM reservation ...
INFO: Detecting host DISK reservation ...
INFO: Detecting host system locale ...
INFO: Detecting host Docker name pipe existing ...
INFO: Detecting host Docker release version ...
template: :1:9: executing "" at <.Server.Platform.Nam...>: can't evaluate field Platform in type *types.Version
FATA: Only support with Docker EE
```

Case 2: 2 network interfaces without specifying node IP
```
INFO: Detecting running permission ...
INFO: Detecting host CPU reservation ...
INFO: Detecting host RAM reservation ...
INFO: Detecting host DISK reservation ...
INFO: Detecting host system locale ...
INFO: Detecting host Docker name pipe existing ...
INFO: Detecting host Docker release version ...
INFO: Detecting host network interface ...
WARN: More than 1 network interfaces are found: 172.17.132.246, 172.17.132.241
WARN: Please indicate --internal-address when adding failed
INFO: Configuring host Docker startup mode to automatical ...
INFO: Enabling host msiscsi service to support iscsi storage ...
WARN[2019-08-30T10:35:25+08:00] Service is not stopped, going to stop it
WARN[2019-08-30T10:35:28+08:00] Service is not stopped, going to stop it
FATA[2019-08-30T10:35:31+08:00] failed to register service: could not query status: The service has not been started.
a7b0e8d08ce4bb5179c3c786bab26a92df643a91f4085ca362f9fda66f5a762b
```

Case 3: Using a Chinese language version of windows server
```
Status: Downloaded newer image for rancher/rancher-agent:master-head
INFO: Detecting running permission ...
INFO: Detecting host CPU reservation ...
INFO: Detecting host RAM reservation ...
INFO: Detecting host DISK reservation ...
INFO: Detecting host system locale ...
FATA: Only support with English System Locale
```

[...]

PR #20941, opened 2019-06-18
[Windows] Add some pre-judgement for running Windows agent

thxCode:
**Problem:**
- User would like to run Windows agent with Docker CE
- Flannel could not start with Chinese System Locale
- Might select the unexpected internal IP if there are more than one
network interfaces in the host

**Solution:**
- Don't allow user to run with Docker CE
- Don't allow user to run with Chinese System Locale
- Warn for multiple network interfaces

**Issue:**
https://github.com/rancher/rancher/issues/20939


thxCode:
Closed, in favor of https://github.com/rancher/rancher/pull/21667

PR #18113, opened 2019-02-14
Don't log whole value of token since it's sensitive data

ukinau:
Currently Node Agent, Cluster Agent print whole token in log, but they are sensitive data, so we should not print whole value. we'd better to mask some part and show only first / last 4 characters.
rancher/rancher#18112

PR #20953, opened 2019-06-18
Remove logging of cluster token

kevinsimons-wf:
The secret token was being logged, since logs can be piped to public storage, we shouldn't log it.


superseb:
This is tracked in https://github.com/rancher/rancher/pull/18113

Issue #20964, opened 2019-06-19
Rancher crashing every 5 minutes with SIGSEGV in etcdbackup


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
unknown

**Result:**
Rancher server is crashing about every 5 minutes. Logs:
```
2019/06/18 23:26:41 [ERROR] CatalogController system-library [catalog] failed with : Timeout in HTTP GET to [https://git.rancher.io/system-charts/index.yaml], did not respond in 30s
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2accef2]

goroutine 6313 [running]:
github.com/rancher/rancher/pkg/controllers/management/etcdbackup.isRecurringBackupEnabled(...)
        /go/src/github.com/rancher/rancher/pkg/controllers/management/etcdbackup/etcdbackup.go:439
github.com/rancher/rancher/pkg/controllers/management/etcdbackup.shouldBackup(0xc0161af680, 0x0)
        /go/src/github.com/rancher/rancher/pkg/controllers/management/etcdbackup/etcdbackup.go:417 +0xb2
github.com/rancher/rancher/pkg/controllers/management/etcdbackup.(*Controller).doClusterBackupSync(0xc00e925680, 0xc0161af680, 0xc01f973f58, 0x1)
        /go/src/github.com/rancher/rancher/pkg/controllers/management/etcdbackup/etcdbackup.go:158 +0x6d
github.com/rancher/rancher/pkg/controllers/management/etcdbackup.(*Controller).clusterBackupSync(0xc00e925680, 0x41f1ba0, 0xc014c239c0, 0x45d964b800, 0xc00e925680, 0xc01bf6b380)
        /go/src/github.com/rancher/rancher/pkg/controllers/management/etcdbackup/etcdbackup.go:145 +0x14c
created by github.com/rancher/rancher/pkg/controllers/management/etcdbackup.Register
        /go/src/github.com/rancher/rancher/pkg/controllers/management/etcdbackup/etcdbackup.go:69 +0x69f
2019/06/18 23:26:53 [INFO] Rancher version v2.2.4 is starting
```

**Other details that may be helpful:**
Server was upgraded from v2.0 -> v2.2.0 -> v2.2.4

**Environment information**
[...]

PR #19623, opened 2019-04-15
Split manual and recurring backups (#18811) + review comments

kinarashah:
https://github.com/rancher/rancher/issues/18807
https://github.com/rancher/rancher/issues/18793

PR #20978, opened 2019-06-19
check for nil enabled

kinarashah:
https://github.com/rancher/rancher/issues/20964

bug was introduced when we refactored backup logic,  https://github.com/rancher/rancher/pull/19623/commits/37746abf15b948ba310a707ad1b821b0a17dc7b3


alena1108:
LGTM

PR #20981, opened 2019-06-19
CIS changes

leodotcloud:


pkg/api/customization/cluster/actions_cis.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("failed to create cis scan object"))
+	}
+
+	updatedCluster := cluster.DeepCopy()
+	updatedCluster.Annotations[cis.RunCISScanAnnotation] = cisScan.Name
+
+	_, err = a.ClusterClient.Update(updatedCluster)
+	if err != nil {
```
kinarashah: we should retry if it's just conflict error

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -0,0 +1,62 @@
...
+		return httperror.WrapAPIError(err, httperror.ServerError, "failed to update cluster annotation for cis scan")
+	}
+
+	cisScanJSON, err := json.Marshal(cisScan)
+	if err != nil {
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("failed to marshal cis scan object"))
+	}
+
+	logrus.Infof("CIS scan requested")
```
kinarashah: maybe add more info, cluster id or scan id

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -0,0 +1,62 @@
...
+	"time"
+
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+	"github.com/rancher/rancher/pkg/controllers/user/cis"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (a ActionHandler) runCISScan(actionName string, action *types.Action, apiContext *types.APIContext) error {
+	cluster, err := a.ClusterClient.Get(apiContext.ID, v1.GetOptions{})
```
kinarashah: maybe someone else can confirm too, but we normally use access.ByID in action handlers, that also makes sure of auth vs clusterClient just getting from db

pkg/api/customization/clusterscan/clusterscan.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (h Handler) LinkHandler(apiContext *types.APIContext, next types.RequestHandler) error {
+	var cs map[string]interface{}
+	if err := access.ByID(apiContext, apiContext.Version, apiContext.Type, apiContext.ID, &cs); err != nil {
+		return err
+	}
+
+	csID, _ := cs["id"].(string)
+	clusterID := strings.Split(csID, ":")[0]
+	clusterScanID := strings.Split(csID, ":")[1]
```
kinarashah: nit: `ref.Parse(convert.ToString(csID))`

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	userCtxNSClient              rcorev1.NamespaceInterface
+	clusterNamespace             string
+	systemAccountManager         *systemaccount.Manager
+	configMapsClient             rcorev1.ConfigMapInterface
+}
+
+func (csh *cisScanHandler) Create(cs *mgmtv3.ClusterScan) (runtime.Object, error) {
+	logrus.Debugf("cisScanHandler: Create: %+v", cs)
+	var err error
+	cluster, err := csh.mgmtCtxClusterClient.Get(cs.ClusterName, metav1.GetOptions{})
```
kinarashah: can use cache lister here

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return cs, err
+	}
+	if !v3.ClusterConditionReady.IsTrue(cluster) {
+		return cs, fmt.Errorf("cisScanHandler: Create: cluster not ready")
```
kinarashah: cluster id for easier debugging

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if !mgmtv3.ClusterScanConditionCreated.IsTrue(cs) {
+		logrus.Infof("cisScanHandler: Create: deploying helm chart")
+		// Deploy the system helm chart
+		if err := csh.deployApp(cs.Spec.ClusterID, cs.Name); err != nil {
+			logrus.Errorf("cisScanHandler: Create: error deploying app: %v", err)
```
kinarashah: can double check, mostly return err anyway prints this in logs with controller name

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error updating condition of cluster scan object: %v", cs.Name)
+		}
+	}
+	return cs, nil
+}
+
+func (csh *cisScanHandler) deployApp(clusterName, appName string) error {
+	appCatalogID := settings.SystemCISBenchmarkCatalogID.Get()
+	err := utils.DetectAppCatalogExistence(appCatalogID, csh.mgmtCtxTemplateVersionClient)
```
kinarashah: can use cache lister here

pkg/app/utils/utils.go
```diff
@@ -125,3 +126,32 @@ func DeployApp(mgmtAppClient projv3.AppInterface, projectID string, createOrUpda
...
+	_, err = cattleTemplateVersionsClient.GetNamespaced(templateVersionNamespace, templateVersionID, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to find catalog by ID %q", appCatalogID)
+	}
+
+	return nil
+}
+
+func DeleteApp(mgmtAppClient projv3.AppInterface, projectID, appName string) error {
+	app, err := mgmtAppClient.GetNamespaced(projectID, appName, metav1.GetOptions{})
```
kinarashah: q: why are we getting the app here if it's just for delete?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	err := csh.configMapsClient.Delete(cs.Name, &metav1.DeleteOptions{})
+	if err != nil {
+		return cs, fmt.Errorf("cisScanHandler: Delete: error deleting cm=%v", cs.Name)
+	}
+
+	return cs, nil
+}
+
+func (csh *cisScanHandler) Updated(cs *mgmtv3.ClusterScan) (runtime.Object, error) {
+	logrus.Debugf("cisScanHandler: Updated: %+v", cs)
```
kinarashah: can we remove this?

pkg/controllers/user/cis/podHandler.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	if v3.ClusterScanConditionCompleted.IsUnknown(cs) &&
+		!v3.ClusterScanConditionCompleted.IsFalse(cs) {
+		v3.ClusterScanConditionCompleted.False(cs)
+		_, err = ph.mgmtCtxClusterScanClient.Update(cs)
+		if err != nil {
+			return nil, fmt.Errorf("error updating condition of cluster scan object: %v", owner)
+		}
+		logrus.Infof("Marking CIS scan complete: %v", owner)
```
kinarashah: but we're updating condition complete False here, I don't understand

pkg/api/customization/clusterscan/clusterscan.go
```diff
@@ -0,0 +1,63 @@
...
+
+	clusterContext, err := h.ClusterManager.UserContext(clusterID)
```
kinarashah: possible to use userContext directly?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type cisScanHandler struct {
+	clusterLister                v3.ClusterLister
+	mgmtCtxClusterClient         v3.ClusterInterface
+	mgmtCtxProjClient            mgmtv3.ProjectInterface
+	mgmtCtxAppClient             projv3.AppInterface
```
kinarashah: nit: we normally just name them as `apps` for AppInterface and  `appLister` for AppLister

pkg/api/customization/cluster/actions_cis.go
leodotcloud: Since it's an API handler, I thought it's not a good idea to retry. But maybe a few retries for every few milliseconds should be ok.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: will fix

pkg/api/customization/cluster/actions_cis.go
kinarashah: if it's a controller, we rely on framework to retry by returning error.. but since this is in api handler, this error would be returned to user and - which shouldn't be if it's just a conflict error and not actual user fault

pkg/app/utils/utils.go
leodotcloud: To check for `app.DeletionTimestamp == nil`

pkg/controllers/user/cis/podHandler.go
leodotcloud: ClusterScanConditionCompleted is set to Unknown at start, it's set to false here and when the app is deleted it's set to true.

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: I used to do the same, but I liked this new approach used in monitoring: https://github.com/rancher/rancher/blob/master/pkg/controllers/user/monitoring/appHandler.go#L13 and took it to next level :)

pkg/api/customization/clusterscan/clusterscan.go
leodotcloud: Since this is in the context of an API handler, we don't have access to userContext directly.

kinarashah: ya, we should check for scan completed before proceeding

pkg/api/customization/cluster/actions_cis.go
leodotcloud: Added three retries with 5ms interval.

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return cs, nil
+}
+
+func (csh *cisScanHandler) Remove(cs *mgmtv3.ClusterScan) (runtime.Object, error) {
+	logrus.Debugf("cisScanHandler: Remove: %+v", cs)
+	// Delete the configmap associated with this scan
+	err := csh.configMapsClient.Delete(cs.Name, &metav1.DeleteOptions{})
+	if err != nil {
+		return cs, fmt.Errorf("cisScanHandler: Delete: error deleting cm=%v", cs.Name)
```
alena1108: we should log an error. The error itself will have a name of cs

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	return cs, nil
+}
+
+func (csh *cisScanHandler) Updated(cs *mgmtv3.ClusterScan) (runtime.Object, error) {
+	logrus.Debugf("cisScanHandler: Updated: %+v", cs)
+	if !mgmtv3.ClusterScanConditionCompleted.IsUnknown(cs) &&
+		mgmtv3.ClusterScanConditionCompleted.IsFalse(cs) {
+		// Delete the system helm chart
+		if err := csh.deleteApp(cs.Name); err != nil {
+			logrus.Errorf("cisScanHandler: error deleting app: %v", err)
```
alena1108: should just return the error. It's better to not log in the methods, and instead return the error back to the framework

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil, err
+		}
+
+		cluster, err := csh.clusterLister.Get("", csh.clusterNamespace)
+		if err != nil {
+			return nil, fmt.Errorf("cisScanHandler: error getting cluster %v", err)
+		}
+
+		updatedCluster := cluster.DeepCopy()
+		delete(updatedCluster.Annotations, RunCISScanAnnotation)
```
alena1108: we don't care about the delete returning an error?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+			logrus.Errorf("cisScanHandler: error deleting app: %v", err)
+			return nil, err
+		}
+
+		cluster, err := csh.clusterLister.Get("", csh.clusterNamespace)
+		if err != nil {
+			return nil, fmt.Errorf("cisScanHandler: error getting cluster %v", err)
+		}
+
+		updatedCluster := cluster.DeepCopy()
```
alena1108: we should only deepCopy/update when there are changes to the objects by comparing with the source. Do we do this comparison anywhere in this method?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+func (csh *cisScanHandler) deleteApp(appName string) error {
+	appDeployProjectID, err := utils.GetSystemProjectID(csh.mgmtCtxProjClient)
+	if err != nil {
+		return err
+	}
+
+	if err := utils.DeleteApp(csh.mgmtCtxAppClient, appDeployProjectID, appName); err != nil {
```
alena1108: we should add special handling for not found type of error

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (csh *cisScanHandler) deleteApp(appName string) error {
+	appDeployProjectID, err := utils.GetSystemProjectID(csh.mgmtCtxProjClient)
+	if err != nil {
```
alena1108: need to add a special handling for not found type of error

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: In the current code context, there is always an update, hence no need to compare.

pkg/api/customization/clusterscan/clusterscan.go
leodotcloud: Took care of that in the Formatter

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -0,0 +1,84 @@
...
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("failed to create cis scan object"))
+	}
+
+	updatedCluster := cluster.DeepCopy()
+	updatedCluster.Annotations[cis.RunCISScanAnnotation] = cisScan.Name
+
+	// Can't add either too many retries or longer interval as this an API handler
+	for i := 0; i < NumberOfRetriesForClusterUpdate; i++ {
+		_, err = a.ClusterClient.Update(updatedCluster)
```
alena1108: for the update retry attempt, you should fetch the new object with get inside of the loop. Otherwise it will always keep on failing if the object got changed in etcd

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	cluster, err := csh.clusterLister.Get("", csh.clusterNamespace)
+	if err != nil {
+		return nil, fmt.Errorf("cisScanHandler: Remove: error getting cluster %v", err)
+	}
+
+	if owner, ok := cluster.Annotations[RunCISScanAnnotation]; ok && owner == cs.Name {
+		updatedCluster := cluster.DeepCopy()
+		delete(updatedCluster.Annotations, RunCISScanAnnotation)
```
alena1108: there might be a chance when the update fails. And then it will never be retried because we return on the line 70.

pkg/api/customization/cluster/actions_cis.go
kinarashah: also we shouldn't retry for all errors, only update conflicts

alena1108:
LGTM

pkg/app/utils/utils.go
kinarashah: I don't get it, why make a call to db just to check if the object is already deleted, we could just issue another delete and check for notFound error

pkg/app/utils/utils.go
```diff
@@ -125,3 +123,32 @@ func DeployApp(mgmtAppClient projv3.AppInterface, projectID string, createOrUpda
...
 	return rtn, nil
 }
+
+func DetectAppCatalogExistence(appCatalogID string, cattleTemplateVersionsClient v3.CatalogTemplateVersionInterface) error {
+	templateVersionID, templateVersionNamespace, err := common.ParseExternalID(appCatalogID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse catalog ID %q", appCatalogID)
+	}
+
+	_, err = cattleTemplateVersionsClient.GetNamespaced(templateVersionNamespace, templateVersionID, metav1.GetOptions{})
```
kinarashah: Lister?

leodotcloud: That's existing code ;-) didn't want to touch it

Issue #20912, opened 2019-06-17
Cannot install Apps since 2.2.4


mwatche:
**What kind of request is this (question/bug/enhancement/feature request):**
Cannot install apps since 2.2.4. With 2.2.3 it runs.

**Steps to reproduce (least amount of steps as possible):**
Install an app for example artifactory-ha

**Result:**
Failed to install.
Log:
2019/06/17 06:26:13 [INFO] Creating roleBinding User u-zmkw5fpsdq Role project-member
[main] 2019/06/17 06:26:16 Starting Tiller v2.10+unreleased (tls=false)
[main] 2019/06/17 06:26:16 GRPC listening on :49892
[main] 2019/06/17 06:26:16 Probes listening on :59489
[main] 2019/06/17 06:26:16 Storage driver is ConfigMap
[main] 2019/06/17 06:26:16 Max history per release is 1
[tiller] 2019/06/17 06:26:17 getting history for release artifactory-ha
[storage] 2019/06/17 06:26:17 getting release history for "artifactory-ha"
2019/06/17 06:26:18 [INFO] Handling backend connection request [c-c54n5:m-c7b40a7757c5]
2019/06/17 06:26:28 [INFO] error in remotedialer server [400]: read tcp 172.17.0.2:443->172.24.15.250:56536: i/o timeout
2019/06/17 06:26:29 [INFO] Handling backend connection request [c-c54n5:m-62c10847a69e]
2019/06/17 06:26:39 [INFO] error in remotedialer server [400]: read tcp 172.17.0.2:443->172.24.249.5:39594: i/o timeout
[storage/driver] 2019/06/17 06:26:47 query: failed to query with labels: Get https://localhost:443/k8s/clusters/c-c54n5/api/v1/namespaces/artifactory-ha/configmaps?labelSelector=NAME%3Dartifactory-ha%2COWNER%3DTILLER: dial tcp 10.128.0.229:443: i/o timeout
[tiller] 2019/06/17 06:26:47 preparing update for artifactory-ha
[storage] 2019/06/17 06:26:47 getting deployed releases from "artifactory-ha" history
2019/06/17 06:26:49 [INFO] Handling backend connection request [c-c54n5:m-c7b40a7757c5]
2019/06/17 06:26:58 [INFO] Handling backend connection request [c-c54n5:m-62c10847a69e]
[...]

PR #20984, opened 2019-06-19
Update jailer script - master

dramich:
Problem:
ssh is not in the jail causing docker-machine to use the native go ssh package, this causes issues depending on the OS the system is running

Solution:
Pull in ssh and required supporting libs

https://github.com/rancher/rancher/issues/20985
https://github.com/rancher/rancher/issues/20987


cjellick:
probably also https://github.com/rancher/rancher/issues/20912

cjellick:
LGTM

PR #20989, opened 2019-06-19
Add CertificatesExpiration to cluster status

moelsayed:
#19103

### DIFF
--- a/pkg/controllers/user/certsexpiration/certsexpiration.go
+++ b/pkg/controllers/user/certsexpiration/certsexpiration.go
@@ -1,0 +1,163 @@
+package certsexpiration
+
+import (
+	"context"
+	"encoding/json"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/rancher/kontainer-engine/cluster"
+	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
+	rkecluster "github.com/rancher/rke/cluster"
+	"github.com/rancher/rke/pki"
+	v1 "github.com/rancher/types/apis/core/v1"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/cert"
+)
+
+type Controller struct {
+	ClusterName   string
+	ClusterLister v3.ClusterLister
+	ClusterClient v3.ClusterInterface
+	ClusterStore  cluster.PersistentStore
+	SecretLister  v1.SecretLister
+}
+
+func Register(ctx context.Context, userContext *config.UserContext) {
+	c := &Controller{
+		ClusterName:   userContext.ClusterName,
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,169 @@
...
+	for certName, certObj := range certBundle {
+		info, err := getCertExpiration(certObj.CertificatePEM)
+		if err != nil {
+			logrus.Debugf("failed to get expiration date for certificate [%s] for cluster [%s]:%v", certName, key, err)
+			continue
+		}
+		certsExpInfo[certName] = info
+	}
+	if !reflect.DeepEqual(cluster.Status.CertificatesExpiration, certsExpInfo) {
+		cluster.Status.CertificatesExpiration = certsExpInfo
```
kinarashah: we need to do deepcopy before updating cluster obj

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,169 @@
...
+		cleanCertificateBundle(currentState.CertificatesBundle)
+		return currentState.CertificatesBundle, nil
+	}
+
+	// No state file, let's try get the certs from the user cluster
+	// clusterContext, err := c.ClusterManager.UserContext(clusterName)
+	// if err != nil {
+	// 	return nil, err
+	// }
+	certs, err := getCertsFromUserCluster(c.UserContxt)
```
kinarashah: since we're already running as user controller, we don't need userContext, 
we can just have `secretsLister` in the Controller struct itself

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,169 @@
...
+	certs := map[string]pki.CertificatePKI{}
+	secrets, err := c.Core.Secrets("").Controller().Lister().List("kube-system", labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	for _, s := range secrets {
+		if strings.HasPrefix(s.GetName(), "kube-") &&
+			s.Type == corev1.SecretTypeOpaque {
+			name := s.GetName()
+			secret, err := c.Core.Secrets("").Controller().Lister().Get("kube-system", name)
```
kinarashah: question, don't we already have secret object since we listed them all on L118, or am I missing something?

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,169 @@
...
+		ClusterClient: userContext.Management.Management.Clusters(""),
+		ClusterStore:  clusterprovisioner.NewPersistentStore(userContext.Management.Core.Namespaces(""), userContext.Management.Core),
+		UserContxt:    userContext,
+	}
+
+	userContext.Management.Management.Clusters("").AddHandler(ctx, "certificate-expiration", c.sync)
+}
+
+func (c Controller) sync(key string, cluster *v3.Cluster) (runtime.Object, error) {
+	if key == "" || cluster == nil || cluster.DeletionTimestamp != nil {
```
kinarashah: since this is running as user controller, we'll now receive sync event in all clusters, but we need to run only for current cluster.. so we need to add `clusterName` in the Controller struct and check for name, else return. 
example, https://github.com/rancher/rancher/blob/master/pkg/controllers/user/monitoring/clusterHandler.go#L50

PR #20990, opened 2019-06-19
Cluster Template Enforcement

prachidamle:
This PR will change after https://github.com/rancher/rancher/pull/20556 is merged.

### DIFF
--- a/pkg/api/customization/cluster/validator.go
+++ b/pkg/api/customization/cluster/validator.go
@@ -1,21 +1,40 @@
 package cluster
 
 import (
+	"strings"
+
+	"fmt"
+
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
+	gaccess "github.com/rancher/rancher/pkg/api/customization/globalnamespaceaccess"
+	"github.com/rancher/rancher/pkg/namespace"
+	"github.com/rancher/rancher/pkg/settings"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	mgmtclient "github.com/rancher/types/client/management/v3"
 )
 
 type Validator struct {
-	ClusterLister v3.ClusterLister
+	ClusterLister                 v3.ClusterLister
+	ClusterTemplateLister         v3.ClusterTemplateLister
+	ClusterTemplateRevisionLister v3.ClusterTemplateRevisionLister
+	Users                         v3.UserInterface
+	GrbLister                     v3.GlobalRoleBindingLister
+	GrLister                      v3.GlobalRoleLister
 }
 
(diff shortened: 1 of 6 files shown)


pkg/api/customization/cluster/validator.go
```diff
@@ -49,3 +70,80 @@ func (v *Validator) validateLocalClusterAuthEndpoint(request *types.APIContext,
...
+		return httperror.NewFieldAPIError(httperror.MissingRequired, "", "A clusterTemplate is required to create a cluster")
+	}
+
+	enforced, err := v.isTemplateEnforced(spec)
+	if err != nil {
+		return err
+	}
+
+	if !enforced {
+		return httperror.NewFieldAPIError(httperror.PermissionDenied, "", "This is not a valid clusterTemplate to create a cluster")
```
alena1108: nit: we should change an error message to include the template name.

pkg/api/customization/cluster/validator.go
```diff
@@ -49,3 +70,80 @@ func (v *Validator) validateLocalClusterAuthEndpoint(request *types.APIContext,
...
+
+	clusterTemp, err := v.ClusterTemplateLister.Get(namespace.GlobalNamespace, spec.ClusterTemplateName)
+	if err != nil {
+		return false, err
+	}
+	if !clusterTemp.Spec.Enforced {
+		return false, nil
+	}
+
+	//make sure it is public
```
alena1108: can we just check the member attribute on the template? Would save us on the API call :)

pkg/api/customization/cluster/validator.go
```diff
@@ -49,3 +68,93 @@ func (v *Validator) validateLocalClusterAuthEndpoint(request *types.APIContext,
...
+
+	isAdmin, err := ma.IsAdmin(callerID)
+	if err != nil {
+		return err
+	}
+	if isAdmin {
+		return nil
+	}
+
+	if !strings.EqualFold(settings.ClusterTemplateEnforcement.Get(), "true") {
```
alena1108: do you think we can make this check before the caller id? So we save on IsAdmin call :)

pkg/api/customization/cluster/validator.go
```diff
@@ -49,3 +68,93 @@ func (v *Validator) validateLocalClusterAuthEndpoint(request *types.APIContext,
...
+		return nil
+	}
+	//enforcement is true, template is a must
+	if spec.ClusterTemplateID == "" || spec.ClusterTemplateRevisionID == "" {
```
alena1108: is this check already covered by checkClusterForEnforcement?

prachidamle: no, we need to make sure template is passed in case it is a create cluster call passing only rkeconfig

alena1108:
LGTM

PR #20995, opened 2019-06-19
Fix clustertemplate tests and update rancher/types

prachidamle:
This PR updates master to latest types(that has changes to the clustertemplate APIs) and fixes the clustertemplate tests
cc @alena1108


alena1108:
LGTM

PR #21000, opened 2019-06-20
replace the / in VERSION arg

drpebcak:
This should fix the issue with the rancher-agent docker image using an invalid tag name when using a `release-v*` docker image.


cjellick:
lgtm

PR #21009, opened 2019-06-20
Update jailer script v2.1

dramich:
Problem:
Jailer needs to use real ssh

Solution:
Pull in ssh and required libs to support it


dramich:
https://github.com/rancher/rancher/issues/21013

Issue #19222, opened 2019-03-27
Allow S3 Snapshot service to trust internally signed certs when talking to the S3 endpoint


jrmercier1:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement/feature request

Allow S3 Snapshot service to utilize internally signed certs when talking to the S3 endpoint.

We utilize internal S3 compatible services and would like to make use of the S3 Snapshot feature provided in 2.2.0.  We receive a trust type warning from the snapshot endpoint since our internal S3 endpoint utilizes an internally signed cert(private CA).

**Steps to reproduce (least amount of steps as possible):**

1. Edit Cluster -> Configure Internal S3 compatible endpoint that uses an internally signed cert (private CA).
2. Select cluster in UI -> Snapshot Now

**Result:**
Error in API view.

message": "Failed to take one-time snapshot, exit code [1]: time=\"2019-03-27T13:38:51Z\" level=fatal msg=\"faield to set s3 server: failed to check s3 bucket:mybucket, err:Head https://s3server:9021/mybucket/: x509: certificate signed by unknown authority\" \n",

**Other details that may be helpful:**

**Environment information**
- Rancher version: 2.2.0
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Kubernetes version (use `kubectl version`): 1.13.4

straffalli:
+1
We encountered the same problem, it could be nice to have the possibility to setup our internal CA when configuring the S3 endpoint for snapshots.

alena1108:
API/Backend support is in master. Separate ticket is filed for UI.

sowmyav27:
Verified in rancher:master-head -- API testing

- Launch a minio server with Self-signed Keys and Certificates.
- Created a cluster, by configuring the minio server as S3 snapshot details without passing the  certs in`customCa`. Error was seen on the rancher logs:

```
2019/06/27 22:17:24 [WARNING] Failed to take one-time snapshot, exit code [1]: time="2019-06-27T22:17:23Z" level=fatal msg="faield to set s3 server: failed to check s3 bucket:sowmya-test, err:Get https://x.x.x.x:9000/sowmya-test/?location=: x509: certificate signed by unknown authority" 
```

- Now create a cluster passing the valid certs in`customCa`field.
- Snapshot is taken successfully. Restore from snapshot works.

PR #21020, opened 2019-06-20
Add custom CA support for S3 backend

moelsayed:
rancher/rancher#19222


pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -373,6 +377,10 @@ func (c *Controller) deleteS3Snapshot(b *v3.EtcdBackup) error {
 	if err != nil {
 		return err
 	}
+	if b.Spec.BackupConfig.S3BackupConfig.EndpointCA != "" {
```
alena1108: shall we check for backup config, and s3 backup config to be not nil too? Or we are safe at this point

moelsayed: it's checked few lines earlier https://github.com/rancher/rancher/pull/21020/files/a6ec63d702a8c45b5620e2f9fb5987bc60b1854e#diff-ffdc9a1bfb3f66be2058cc55605965afR349

alena1108:
LGTM

PR #21060, opened 2019-06-21
Backport: Add permissions check for secrets

rmweir:
Backport: https://github.com/rancher/rancher/pull/21052

### DIFF
--- a/pkg/api/customization/secret/validator.go
+++ b/pkg/api/customization/secret/validator.go
@@ -1,0 +1,58 @@
+package secret
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/rancher/norman/api/access"
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+	v1 "github.com/rancher/types/apis/core/v1"
+)
+
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if request.Method == http.MethodPost {
+		id := ""
+
+		// extracting project name from data
+		if projectData, ok := data["projectId"].(string); ok {
+			if projectParts := strings.Split(projectData, ":"); len(projectParts) > 1 {
+				id = fmt.Sprintf("%s:%s", projectParts[1], data["name"])
+			}
+		}
+
+		// minimum info needed to use CanDo
+		secretState := map[string]interface{}{
+			"name":        data["name"],
+			"id":          id,
+			"namespaceId": data["namespaceId"],
+		}
+
+		// update is used here to avoid the application of general user permissions for secrets
+		if err := request.AccessControl.CanDo(v1.SecretGroupVersionKind.Group, v1.SecretResource.Name, "update", request, secretState, schema); err != nil {
+			return httperror.NewAPIError(httperror.PermissionDenied, "unauthorized")
+		}
+	} else if request.Method == http.MethodPut {
+		var secretState map[string]interface{}
+
+		if err := access.ByID(request, request.Version, request.Type, request.ID, &secretState); err != nil {
+			if httperror.IsNotFound(err) || isUnauthorized(err) {
+				return httperror.NewAPIError(httperror.NotFound, "not found")
+			}
+			return httperror.NewAPIError(httperror.ServerError, err.Error())
+		}
+		// this is unused but will be necessary if readonly users are ever given permission to view secrets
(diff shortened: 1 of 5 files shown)


pkg/api/customization/secret/validator.go
```diff
@@ -0,0 +1,48 @@
...
+		// minimum info needed to use CanDo
+		secretState := map[string]interface{}{
+			"name":        data["name"],
+			"id":          id,
+			"namespaceId": data["namespaceId"],
+		}
+
+		// update is used here to avoid the application of general user permissions for secrets
+		if err := request.AccessControl.CanDo(v1.SecretGroupVersionKind.Group, v1.SecretResource.Name, "update", request, secretState, schema); err != nil {
+			return httperror.NewAPIError(httperror.PermissionDenied, fmt.Sprintf("unable to validate: user does not have permission"))
```
cjellick: change "unable to validate: user does not have permission" to simply "unauthorized"

cjellick: "unable to validate" just isnt a useful error to display

pkg/api/customization/secret/validator.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+	} else if request.Method == http.MethodPut {
+		var secretState map[string]interface{}
+
+		if err := access.ByID(request, request.Version, request.Type, request.ID, &secretState); err != nil {
+			return httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("unable to validate: %v", err))
```
cjellick: you dont know this is a notFound error here.
Technically, the right thing to do would be to check to see if the error is a not found error and if it is, return NotFound, else return a 500 error with just the raw error message.

pkg/api/customization/secret/validator.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		// this is unused but will be necessary if readonly users are ever given permission to view secrets
+		if err := request.AccessControl.CanDo(v1.SecretGroupVersionKind.Group, v1.SecretResource.Name, "update", request, secretState, schema); err != nil {
+			return httperror.NewAPIError(httperror.PermissionDenied, "unable to validate: user does not have permission")
```
cjellick: just say unauthorized

tests/integration/suite/test_rbac.py
```diff
@@ -475,19 +476,147 @@ def test_member_can_perform_app_action(admin_mc, admin_pc, remove_resource,
...
+    prtb = admin_mc.client.create_project_role_template_binding(
+        name="prtb-" + random_str(),
+        userId=user_mc.user.id,
+        projectId=project.id,
+        roleTemplateId="read-only")
+    remove_resource(prtb)
+
+    wait_until_available(user_client, project)
+
+    proj_client = user_project_client(user_mc, project)
```
dramich: Nit but I think this should be renamed to be clear it's the user client

pkg/api/customization/secret/validator.go
```diff
@@ -0,0 +1,58 @@
...
+
+		// update is used here to avoid the application of general user permissions for secrets
+		if err := request.AccessControl.CanDo(v1.SecretGroupVersionKind.Group, v1.SecretResource.Name, "update", request, secretState, schema); err != nil {
+			return httperror.NewAPIError(httperror.PermissionDenied, "unauthorized")
+		}
+	} else if request.Method == http.MethodPut {
+		var secretState map[string]interface{}
+
+		if err := access.ByID(request, request.Version, request.Type, request.ID, &secretState); err != nil {
+			if httperror.IsNotFound(err) || isUnauthorized(err) {
```
dramich: why do we 404 on unauthorized?

dramich:
lgtm

pkg/api/customization/secret/validator.go
```diff
@@ -0,0 +1,58 @@
...
+	v1 "github.com/rancher/types/apis/core/v1"
+)
+
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if request.Method == http.MethodPost {
+		id := ""
+
+		// extracting project name from data
+		if projectData, ok := data["projectId"].(string); ok {
+			if projectParts := strings.Split(projectData, ":"); len(projectParts) > 1 {
```
alena1108: Should we return an error if by some reason len(projectParts) <=1?

alena1108:
lgtm

Issue #21067, opened 2019-06-22
Creating cluster templates as a "standard user" fails through API


soumyalj:
Version: Master

Steps
1. Login to the rancher setup as an admin
2. Create "user1" as a standard user
3. Login as "user1"
4. Create a cluster template T1using the API

Cluster template creation fails with the below error

```

HTTP/1.1 403
content-encoding: gzip
content-length: 200
content-type: application/json
date: Sat, 22 Jun 2019 03:26:32 GMT
expires: Wed 24 Feb 1982 18:42:00 GMT
x-api-schemas: https://157.230.157.78/v3/schemas

{

    "baseType": "error",
    "code": "Forbidden",
    "message": "clustertemplates.management.cattle.io is forbidden: User \"u-7vjmx\" cannot create resource \"clustertemplates\" in API group \"management.cattle.io\" in the namespace \"cattle-global-data\"",
    "status": 403,
    "type": "error"

}



```

prachidamle:
PR: https://github.com/rancher/rancher/pull/21090

soumyalj:
Tested with master-head build from June 25.
Cluster template(T1) /template revision(R1) and cluster creation with T1/R1 succeeds for a standard user.

PR #21090, opened 2019-06-24
Standard User role changes for clustertemplates resource

prachidamle:
As a standard user, one should be able to create clustertemplates and clustertemplaterevisions
https://github.com/rancher/rancher/issues/21067


alena1108:
LGTM

Issue #20927, opened 2019-06-17
Cluster remains in "Active" state when all 3 etcd nodes are powered OFF


soumyalj:
Rancher Version: v2.2.5-rc3
K8s version: v1.14.3-rancher1-1

Steps:
1. Create a custom cluster using S3 configuration with nodes as below:
N1 - control, worker
N2, N3, N4 - 3 etcds
2. Create a few workloads
3. Power off etcd nodes N2, N3, N4 

After powering off the etcd nodes, the cluster remains active in the UI and the cluster cannot be accessed through the UI. We cannot proceed further to delete the existing etcd nodes and restore the cluster from S3 . Cluster should to "Unavailable" state

Note: On the same setup, with K8s v1.13.7-rancher1-1, the cluster goes to "Unavailable" state . Cluster can still be accessed through UI. Cluster restore from S3 succeeded
This is true on a setup with v2.2.4 too.

moelsayed:
@soumyalj How long does it take the v1.13.7-rancher1-1 cluster to go to the Unavailable state?

soumyalj:
Tested with build master-head
K8s version : v1.14.3

Steps:
Create a custom cluster using S3 configuration with nodes as below:
N1 - control, worker
N2, N3, N4 - 3 etcds
Create a few workloads
Power off etcd nodes N2, N3, N4

After powering off the etcd nodes, the cluster changed to "Unavailable" state as expected.

<img width="1343" alt="Screen Shot 2019-06-26 at 9 47 40 PM" src="https://user-images.githubusercontent.com/18536626/60235793-c837ee80-985d-11e9-9947-281f398607d9.png">

sangeethah:
Fix for this issue resulted in #21216

soumyalj:
Tested with build master-head
K8s version : v1.14.3

Steps:
Create a custom cluster using S3 configuration with nodes as below:
N1 - control, worker
N2, N3, N4 - 3 etcds
Create a few workloads
Power off etcd nodes N2, N3, N4

After powering off the etcd nodes, the cluster changed to "Unavailable" state as expected.
Cluster was accessible through UI. etcd nodes were deleted and new etcd nodes were added and cluster was restored from S3.

Related issue is filed here: 
https://github.com/rancher/rancher/issues/21319

PR #21091, opened 2019-06-24
Fix componentstatus check for k8s v1.14

moelsayed:
https://github.com/rancher/rancher/issues/20927

### DIFF
--- a/pkg/controllers/user/healthsyncer/healthsyncer.go
+++ b/pkg/controllers/user/healthsyncer/healthsyncer.go
@@ -1,6 +1,7 @@
 package healthsyncer
 
 import (
+	"fmt"
 	"time"
 
 	"context"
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/healthsyncer/healthsyncer.go
```diff
@@ -74,6 +75,12 @@ func (h *HealthSyncer) getComponentStatus(cluster *v3.Cluster) error {
 	sort.Slice(cluster.Status.ComponentStatuses, func(i, j int) bool {
 		return cluster.Status.ComponentStatuses[i].Name < cluster.Status.ComponentStatuses[j].Name
 	})
+	// I am checking for individual componentstatus here to make sure I don't retrun before they are updated in the cluster object
```
alena1108: We should make a comment less personal :)

alena1108:
LGTM

PR #21095, opened 2019-06-24
Get clustertemplate name from spec in revision controller

mrajashree:


alena1108:
@mrajashree can we add a test for it?

alena1108:
LGTM

PR #21101, opened 2019-06-25
add print for more info on mcapp_revision_access test

kinarashah:
https://github.com/rancher/rancher/issues/21100


kinarashah:
Discussed with @mrajashree, she'll be marking `test_multiclusterapp_create_with_members` non parallel in one of her PRs, which seems to be the reason for this, since it shares mcapp with all authenticated users.

PR #21105, opened 2019-06-25
Add permissions check for secrets

rmweir:
Problem:
Any type of secret can be created by a read-only member

Solution:
Add validator to check project permissions on secrets

Issue:


alena1108:
LGTM

PR #21110, opened 2019-06-25
Disable operator when enable alert

aiwantaozi:
Problem:
System chart have bug that enable alertmanager will also deploy operator if not set false to enable operator

Solution:
Set enable operator to false, system charts bug will be fixed later

Issue:
https://github.com/rancher/rancher/issues/20975

### DIFF
--- a/pkg/controllers/user/alert/deployer/upgradeimpl.go
+++ b/pkg/controllers/user/alert/deployer/upgradeimpl.go
@@ -89,9 +89,6 @@ func (l *alertService) Upgrade(currentVersion string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if currentVersion == NewVersion {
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/alert/deployer/upgradeimpl.go
```diff
@@ -132,6 +129,7 @@ func (l *alertService) Upgrade(currentVersion string) (string, error) {
 	}
 	newApp := app.DeepCopy()
 	newApp.Spec.ExternalID = newCatalogID
+	newApp.Spec.Answers["operator.enabled"] = "false"
```
orangedeng: I don't like the way we only upgrade one of the answer value. Maybe we should apply all the answers we need in this version to the new app.

orangedeng:
LGTM

PR #21111, opened 2019-06-25
[2.2]Disable operator when enable alert

aiwantaozi:
Problem:
System chart have bug that enable alertmanager will also deploy operator if not set false to enable operator

Solution:
Set enable operator to false, system charts bug will be fixed later

Issue:
https://github.com/rancher/rancher/issues/21113


loganhz:
LGTM

PR #21114, opened 2019-06-25
embedded mode not use leader elect

better0332:


better0332:
It help for debug, otherwise it fault ` leaderelection lost for cattle-controllers`

dramich:
@better0332 you can set the envvar `CATTLE_DEV_MODE` which will give you hours before you hit this issue. https://github.com/rancher/norman/blob/master/leader/leader.go#L56

dramich:
Closing based off previous comment

PR #21121, opened 2019-06-25
Fix componentstatus check for k8s v1.14

alena1108:


alena1108:
https://github.com/rancher/rancher/issues/21122

PR #21151, opened 2019-06-26
Cluster Template Questions API

prachidamle:
Depends on types PR https://github.com/rancher/types/pull/890


pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -57,3 +68,61 @@ func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHan
 	}
 	return nil
 }
+
+func (w Wrapper) ClusterTemplateRevisionsActionHandler(actionName string, action *types.Action, apiContext *types.APIContext) error {
+	if actionName != "listquestions" {
```
mrajashree: Can we also add a check to see if caller of this action has access to this revision using access.ByID

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -57,3 +68,61 @@ func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHan
...
+
+	questionsOutput := v3.ClusterTemplateQuestionsOutput{}
+
+	if len(w.ClusterTemplateQuestions) == 0 {
+		w.ClusterTemplateQuestions = w.BuildQuestionsFromSchema(apiContext.Schemas.Schema(&managementschema.Version, client.ClusterSpecBaseType), apiContext.Schemas, "")
+	}
+	questionsOutput.Questions = w.ClusterTemplateQuestions
+	res, err := json.Marshal(questionsOutput)
+	if err != nil {
+		return fmt.Errorf("Error marshalling the Cluster Template questions output, %v", err)
```
mrajashree: Can we use httperror.WrapAPIError here?

pkg/api/customization/clustertemplate/cluster_template.go
prachidamle: This is a collectionAction and not an action on a particular revision. So I think the access check is not applicable here.

Issue #21113, opened 2019-06-25
Backport: No more alert after upgrade rancher server


loganhz:
https://github.com/rancher/rancher/issues/20975

loganhz:
For regression test,

Please make sure

1. Enable cluster alerting in 2.2.3 and upgrade to 2.2.5, cluster alerting app can be upgraded successfully
2. Enable cluster alerting in 2.2.4 and upgrade to 2.2.5, cluster alerting app can be upgraded successfully
3. Cluster alerting still functional after upgrade to 2.2.5
4. There is no prometheus operator deployed in cluster-alerting app
5. There is only one prometheus operator deployed in monitoring-operator app

jiaqiluo:
The bugfix is validated on the upgrade path from v2.2.3 -> v2.2-head
- Enable cluster alerting in 2.2.3 and upgrade to v2.2-head, cluster alerting app can be upgraded successfully
- Cluster alerting is still functional after upgrade to v2.2-head
- There is no Prometheus operator deployed in the cluster-alerting app
- There is only one Prometheus operator deployed in monitoring-operator app
<img width="1433" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/60133795-f977c800-9752-11e9-830b-eedbd7b02365.png">


---------

Failed to validate the bugfix on the upgrade path from v2.2.4 -> v2.2-head 

The following is as expected 
- Enable cluster alerting in 2.2.4 and upgrade to v2.2-head, cluster alerting app can be upgraded successfully
- Cluster alerting is still functional after upgrade to v2.2-head
- There is only one Prometheus operator deployed in monitoring-operator app

BUT
- **There is a  Prometheus operator deployed in cluster-alerting app**
<img width="1426" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/60133970-56737e00-9753-11e9-84d6-6f2c0a2e048a.png">

<img width="1413" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/60134933-7b68f080-9755-11e9-85ff-4a5a0ee6e103.png">

@loganhz @aiwantaozi Is it expected? Why one `cluster-alerting` has the `prometheus-operator-cluster-alerting` but another one does not? They are the same version.

davidnuzik:
@jiaqiluo please try changing ui-index (under settings) to https://releases.rancher.com/ui/2.2.79/index.html
Make sure browser cache is cleared.

jiaqiluo:
[...]

PR #21152, opened 2019-06-26
Remove version compare when upgrade alerting

aiwantaozi:
Problem:
Rancher 2.2.4 and 2.2.5 use monitoring chart 0.0.3, version check will
prevent app upgrade when version not change but we still want to update
the app

Solution:
Remove version check when upgrade alertmanager

Issue:
https://github.com/rancher/rancher/issues/21113


orangedeng:
LGTM

Issue #20805, opened 2019-06-11
Protect the required linux worker node in a windows cluster from windows workloads/deployments


deniseschannon:

loganhz:
![image](https://user-images.githubusercontent.com/21168270/62526733-6ca14d80-b86c-11e9-9412-a732e5e65af0.png)

@orangedeng sometimes I can see some linux pod failed to start up. But the toleration has been added. I have to remove and re-add them manually. Can you look at it, please?

sowmyav27:
**Verified with latest master-head.**

Windows AWS node: `Windows_Server-1903-English-Core-Base-2019.08.16`
Linux Node: `18.04 Ubuntu LTS`

**Steps:**
1. Launch a custom cluster 1 etcd and control plane linux node, 1 worker linux node and 1 worker windows node.
2. Linux node docker - 18.09, Windows node docker - 19.03, k8s version - 1.15.3
3. when the cluster comes up `Active`, the nginx-ingress-controller pods are scheduled on etcd/ control plane node also. 

Nodes show this taint 
<img width="1060" alt="Screen Shot 2019-08-23 at 10 52 50 AM" src="https://user-images.githubusercontent.com/26032343/63614150-33205000-c597-11e9-96ee-dbb518541f7b.png">

sangeethah:
Tested in latest master build
Create a linux cluster with 1 control plane , 1 etcd and 3 worker nodes.
nnginx-ingress-controller pods get scheduled only in the worker nodes as expected.

sowmyav27:
Verified with latest master-head. - latest - commit id: `d2ebb03eabe8`
**Setup:**
```
Windows AWS node: `Windows_Server-1903-English-Core-Base-2019.08.16`
Linux Node: `18.04 Ubuntu LTS`
Linux node docker and Windows node docker - 19.03, k8s version - 1.15.3
```
**Steps:**

- Launch a custom cluster
- Register 1 etcd and control plane linux node
- Register 1 worker linux node
- Register 2 worker windows node.
- when the cluster comes up Active, the nginx-ingress-controller pods are scheduled on worker node of linux only
- Nodes show this taint

<img width="931" alt="Screen Shot 2019-08-29 at 10 31 32 PM" src="https://user-images.githubusercontent.com/26032343/63995368-1dd28680-caad-11e9-9af8-03b8d4044d6e.png">

Issue #13972, opened 2018-06-13
No way to add taints to nodes


alexandrem:
Background:

There doesn't seem to be any way to add taints in Rancher or RKE at the moment.

Ideally, we should be allowed to define this in the node template.

If someone wants to create dedicated load balancer (e.g ingress controllers) nodes, then we can't mark those nodes with taints and use tolerations for specific workloads. 

This becomes a pain because you have to do this afterwards with `kubectl`, but you needlessly need to change taints to NoExecute to evict some containers that have been scheduled automatically to the nodes, where it should have been prevented at creation time.

Expected:

- [x] Support taints for node pool
- [x] Support taints for node template
- [x] Support taints for node
- [x] Support taints when adding custom node
- [x] Show taints in node list
- [x] Support adding tolerations when enable istio and monitoring for workloads

ekristen:
Please add this. I too would have the need to restrict what runs on what workers, not being able to do taints from within Rancher or the Node Templates makes this more difficult to manage.

ekristen:
@deniseschannon since you removed 2.0 version does that mean this isn't coming anytime soon?

vincent99:
It means nothing.  We're going through issues and removing the `version/2.0` label while doing so because (it's no longer 2.0 and) that's now the implied default instead of labeling both 1.x and 2.x.

garyhuang123:
I'm using vsphere providers in Rancher 2.2.1. Every time when i upgrade a node, i need to re-taint the node. It should be allowed to specify taints on nodetemplate. Mark this up!

loganhz:
The following stuff will be added in 2.3. Please let me know if we missed some requirements.

- Support taints for node pool
- Support taints for node
- Support taints when adding custom node
- Show taints in node list
- Support adding tolerations when enable istio and monitoring for workloads

ekristen:
@loganhz those sound great! Thanks for the hard work.

sangeethah:
Tracking validation for  - `Support adding tolerations when enable istio and monitoring for workloads` - https://github.com/rancher/rancher/issues/21487#issuecomment-527530723 and https://github.com/rancher/rancher/issues/22606

sowmyav27:
Verified on `master-head` the following scenarios.
The basic tests cases are working fine.
Any specific issue has been logged as a bug and is being tracked separately.

- Support taints for node pool
- Support taints for node template
- Support taints for node
- Support taints when adding custom node
- Show taints in node list

PR #21155, opened 2019-06-26
Add node taints logic

orangedeng:
About node taints enhancement:
- Parse node taints from node pool and node template to node when creating a new node.
- Configure node taints from node registration.

About Linux node taints in Windows cluster:
- Taint linux node in Windows preferred cluster.
- Add toleration to workloads with linux node scheduling rules(e.g. `NodeSelector: beta.kubernetes.io/os=linux`)

Related issue:

Related PRs:
https://github.com/rancher/types/pull/883


pkg/controllers/user/windows/workload_test.go
```diff
@@ -0,0 +1,341 @@
...
+	deploy   *v1beta2.Deployment
+	rc       *v1.ReplicationController
+	rs       *v1beta2.ReplicaSet
+	ss       *v1beta2.StatefulSet
+	ds       *v1beta2.DaemonSet
+	job      *batchv1.Job
+	cj       *batchv1beta1.CronJob
+}
+
+type workloadTestCase struct {
```
gitlawr: Add `Name` to testcase, and print it on failure. It would help to know what each case is for, and to track test failure. The same to the other tests.

pkg/controllers/user/windows/linux_node_taints.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/rancher/rancher/pkg/taints"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var (
+	nodeTaint = v1.Taint{
```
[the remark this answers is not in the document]
orangedeng: It should have a value.

pkg/controllers/user/windows/workload_toleration.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/kubernetes/pkg/apis/core/v1/helper"
+)
+
+var (
+	HostOSLabel = labels.Set(map[string]string{
+		"beta.kubernetes.io/os": "linux",
```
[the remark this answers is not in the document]
orangedeng: We should consider the backward compatibility.

pkg/controllers/user/windows/workload_toleration.go
```diff
@@ -0,0 +1,150 @@
...
+			return nil
+		}
+		newObj := ss.DeepCopy()
+		newObj.Spec.Template.Spec.Tolerations = append(newObj.Spec.Template.Spec.Tolerations, newOSTolerations())
+		if _, err := w.workloadController.StatefulSets.Update(newObj); err != nil {
+			logrus.Errorf("failed to update workload %s with os toleration", obj.Name)
+			return err
+		}
+	case job != nil:
+		// Do nothing for jobs. Some jobs may not be run again.
```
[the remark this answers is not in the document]
orangedeng: I am afraid that we update the job object and it will run twice.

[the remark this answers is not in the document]
orangedeng: I think we should also handle job. It won't get running if a linux workload gets scheduled in Windows node.

orangedeng:
Updated

pkg/controllers/user/nodesyncer/nodetaints.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	if obj.Spec.UpdateTaintsFromAPI == nil {
+		return obj, nil
+	}
+	node, err := nodehelper.GetNodeForMachine(obj, n.nodeLister)
+	if err != nil {
+		return obj, err
+	}
+	toAdd, toDel := taints.GetToDiffTaints(node.Spec.Taints, obj.Spec.DesiredNodeTaints)
```
gitlawr: It looks like the diff result is based on `DesiredNodeTaints `, which is cleaned up in L51. Then existing taints will become delete candidates in next sync when `DesiredNodeTaints ` is empty. Should we skip sync in empty desired taints

orangedeng: It won't, because we update `UpdateTaintsFromAPI` to nil and we will skip the sync if `UpdateTaintsFromAPI` == nil.

gitlawr:
LGTM

orangedeng:
Change types vendor based types PR is changed.

pkg/controllers/user/windows/linux_node_taints.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	taintSet := taints.GetTaintSet(obj.Spec.InternalNodeSpec.Taints)
+	// taint exists on nodes
+	if _, ok := taintSet[taints.GetTaintsString(nodeTaint)]; ok {
+		return obj, nil
+	}
+
+	newObj := obj.DeepCopy()
+	newObj.Spec.DesiredNodeTaints = append(newObj.Spec.DesiredNodeTaints, newObj.Spec.InternalNodeSpec.Taints...)
```
kinarashah: q: what is this for? I thought we'd only be adding `nodeTaint` here

orangedeng:
Rebased on master.

pkg/controllers/user/windows/linux_node_taints.go
orangedeng: The desired node taints field should be nil if the `obj.Spec.UpdateTaintsFromAPI` is nil. That is the normal state for node. It means that no need to update node taints.
When we want to update the taints, we will set both `DesiredNodeTaints` and `UpdateTaintsFromAPI`. Because we need to support the removing node taints case, we need to set the whole expected node taint set into `DesiredNodeTaints`.
That is the reason why I need to copy `InternalNodeSpec.Taints` to desired node taints and append our taint into that list.

kinarashah:
LGTM

Issue #21061, opened 2019-06-21
Azure character limits on resource groups


daxmc99:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Create a resource Group in azure with a very long  >60 resource name 
Attempt to use that resource group to provision a cluster

**Result:**
Azure cluster bring up returns a error about the name of one of the parameters exceeding a limit specified in https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions#general 

**Other details that may be helpful:**
Rancher should never send data to azure that we know will exceed a known naming limit
Error seen: 
Operationalinsights.WorkspacesClient#CreateOrUpdate: Invalid input: autorest/validation: validation failed: parameter=workspaceName constraint=MaxLength value="rancher-kube-cluster-dev2-c24b7341-10ad-4e4c-b3f7-f07c483184a9-WEU" details: value length must be less than or equal to 63; waiting on cluster-provisioner-controller (thanks @dbsanfte )

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
2.2.4

davidnuzik:
Duplicate of https://github.com/rancher/rancher/issues/20783 I believe.
We may want to close #20783 in favor of this issue.

davidnuzik:
Version: master-head (6/24/19 9am mst)
I was asked to reproduce the issue.

I see an error like:
`2019/06/24 18:34:42 [ERROR] NodeController c-lnzmj/m-wh9dz [node-controller] failed with : Error creating machine: Error in driver during machine creation: #: : StatusCode=%!!(MISSING)d(<nil>)`
When I attempt to create a cluster with a node template that defines a resource group with a string of longer than 63 chars.
So I am not seeing the exact same error, but perhaps it has something to do with the string used `david-sadgjasdlkfjalsdkjglksadjflkasdjlgkajsdklfjasdlkjgalksdjflkasjdlkgjasldkjfklasdgasdgasdjfkjskdjigjdkfjd`

daxmc99:
Current status:
the issue is caused by this line https://github.com/rancher/rancher/blob/7e115d273fbe56d069fe99aa399aa7e745700569/vendor/github.com/rancher/kontainer-engine/drivers/aks/aks_driver.go#L978
This concatenates 3 values, the resource Group, the subscription id and the workspace region code.
[...]

PR #21199, opened 2019-06-27
Fix logging analytics workspace naming length

daxmc99:
Addresses https://github.com/rancher/rancher/issues/21061

Dependency on: https://github.com/rancher/kontainer-engine/pull/165


cjellick:
lgtm

PR #21219, opened 2019-06-28
Check auth type before creating secret

mrajashree:
For auth providers with common secret field names, secrets
were getting generated irrespective of auth type being set,
example spKey is common for all SAML providers

### DIFF
--- a/pkg/api/store/auth/store.go
+++ b/pkg/api/store/auth/store.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
+	"github.com/rancher/norman/types/values"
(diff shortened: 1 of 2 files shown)


pkg/api/store/auth/store.go
```diff
@@ -38,7 +39,14 @@ type Store struct {
 }
 
 func (s *Store) Update(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, id string) (map[string]interface{}, error) {
+	authType, found := values.GetValue(data, "type")
+	if !found {
+		return nil, fmt.Errorf("invalid data for auth store update")
+	}
 	for kind, fields := range TypeToFields {
+		if kind != authType {
```
kinarashah: I think we can get fields directly from map using authType

PR #21245, opened 2019-07-01
[WIP]Don't use timestamp in monitoring app

orangedeng:
**Problem:**
The monitoring app helper updates app with newest timestamp in answers
cause the app revision grows every sync of cluster.

**Solution:**
Remove the timestamp in monitoring app answers.


orangedeng:
cc @loganhz

Issue #20387, opened 2019-05-22
Package local copy of system-charts into rancher/rancher container


cjellick:
This is to improve air-gap functionality. At build time, the catalog should be cloned into the container and if we cannot reach pull the public chart, we should fallback to the locally cloned version

dramich:
Per discussion only the catalogs `library` and `system-library` will be included in the image.

dramich:
A new setting will be added to tell Rancher to use internal resources vs attempting to fetch them (catalogs for now and more things in the future). This needs to be documented and exposed in the helm chart and single rancher installation instructions.

sangeethah:
@jiaqiluo With this fix, we should be able to enable monitoring in airgap setup without having to setup a private github repo for system charts.

jiaqiluo:
Tested in Rancher:v2.3.0-rc1 in air-gapped HA mode

we can pass the `CATTLE_SYSTEM_CATALOG` as the extra variable when rendering the templates

```
helm template ./rancher-2.3.0-rc1.tgz --output-dir . \
 --name rancher \
 --namespace cattle-system \
 --set hostname=my-hostname \
 --set rancherImage=my-registry/rancher/rancher \
 --set systemDefaultRegistry=my-registry \
 --set `extraEnv[0].name=CATTLE_SYSTEM_CATALOG`\
 --set `extraEnv[0].value=bundled`
```

PR #21260, opened 2019-07-01
Package catalog into build

dramich:
This adds a new setting `system-catalog` which changes the source of the catalogs. The two options are `bundled` and `external` with a default of `external`. If set to `bundled` the two rancher catalogs will only use the bundled version of the catalog from the build and will not attempt to fetch from the internet. If set to `bundled` the normal catalog will be disable by default but can be turned back on. This is to stop people from trying to install apps when the image may not be available to install. 

In order to update a catalog a new one could be copied in (not ideal) or the version of rancher updated in which case a new version could potentially be in the build. 

Issue: https://github.com/rancher/rancher/issues/20387

### DIFF
--- a/app/catalog_data.go
+++ b/app/catalog_data.go
@@ -5,8 +5,11 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strings"
 	"time"
 
+	"github.com/rancher/rancher/pkg/catalog/helm"
+	"github.com/rancher/rancher/pkg/settings"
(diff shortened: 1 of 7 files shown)


dramich:
Flag was updated to `system-catalog` to be specific.

erikwilson:
LGTM

daxmc99:
Could you check your commit message? Would be nice to have some more detail

Issue #21194, opened 2019-06-27
Change redirectURI being returned by Google OAuth provider login/configureTest api


mrajashree:
The redirectURI returned by the backend contains query parameters for a generated state, scopes and the uri to redirect to along with some other params specific to google oauth. Based on a discussion with @etokatlian backend should remove the following params:
1. state: since UI will generate and append it
2. redirectURI: so that UI can add path and customize it
3. scopes: UI will hardcode those 
```
openid
profile
email
https://www.googleapis.com/auth/admin.directory.user.readonly
https://www.googleapis.com/auth/admin.directory.group.readonly
```

mrajashree:
Closing as changes from both api and ui are in 
cc @etokatlian

PR #21261, opened 2019-07-01
Update google oauth redirectURL params

mrajashree:
Remove params for scope, state and the redirectURI since UI will append all these.

https://github.com/rancher/rancher/issues/21194


alena1108:
LGTM

PR #20375, opened 2019-05-21
Self Healing nodepools

daxmc99:
Verified following:

- [x]    Cluster created with cloud provider passed in (AWS). Check that the node is gone from Kubernetes only when it is removed from the cloud provider.
 - [x]   Cluster created without cloud provider option. Check at which point the node gets removed form Kubernetes.
 - [x]  `Kubectl delete node` is removed from Rancher UI as well. Rancher UI should reflect `kubectl get nodes` as closely as possible.
- [ ] Nodes are scaled down oldest first (sorted in the UI?)
- [x] Ensure no regression for this issue https://github.com/rancher/rancher/pull/13609

This fix allows nodes deleted via the `kubectl delete node` to actually be deleted in Rancher. If a cloud provider is used this means that `kubectl delete node` will actually destroy the underlying infrastructure. 

This PR also includes the feature of self-healing nodepools, where nodepools can have the self-healing option set to True & configure a timeout specified in seconds to delete and attempt to recreate the node. 

- [x] _Dependency_ Types PR: https://github.com/rancher/types/pull/868

PR #21269, opened 2019-07-01
Implement natural sorting for nodes in nodepool

daxmc99:
Dependency on: https://github.com/rancher/rancher/pull/20375

### DIFF
--- a/pkg/controllers/management/nodepool/util.go
+++ b/pkg/controllers/management/nodepool/util.go
@@ -1,34 +1,52 @@
 package nodepool
 
-import "regexp"
+import (
+	"strconv"
+
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+
+	"github.com/sirupsen/logrus"
+)
 
 // use pointer to avoid copying large number of structs
-// could also use strings here maybe?
-// type nodes []*v3.Node
 
-type stringSlice []string
-
-// implement sort interface
+type byNodeRequestedHostName []*v3.Node
 
 /*
  Expecting to see strings styled [Prefix] + [some decimal value] which means we can do some specific things
 Inspired by https://github.com/facette/natsort/blob/2cd4dd1e2dcba4d85d6d3ead4adf4cfd2b70caf2/natsort.go
 */
 
-func (s stringSlice) Len() int {
-	return len(s)
+func (s byNodeRequestedHostName) Len() int {
+	return len(s) //returns len of slice of structs
 }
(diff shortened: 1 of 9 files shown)


pkg/controllers/management/nodepool/util.go
```diff
@@ -0,0 +1,52 @@
...
+	return Compare(s[a].Spec.RequestedHostname, s[b].Spec.RequestedHostname)
+}
+
+func (s byNodeRequestedHostName) Swap(a, b int) {
+	s[a], s[b] = s[b], s[a]
+}
+
+// Compare returns true if the first string precedes the second one according to natural order
+func Compare(a, b string) bool {
+	s1 := nameRegexp.FindStringSubmatch(a)
```
vincent99: FWIW in the UI we compute (a string](https://github.com/rancher/ui/blob/3e647571b00187f38e791ea9ef15dfc75f4ba981/lib/shared/addon/utils/util.js#L213-L223) that is ascii-sortable once by zero-padding number suffixes (with regex) to a fixed long-enough length (i.e. `thing1` -> `thing0000000001`) and then let the normal native implementation sort that.

I have no idea how slow regex, atoi, etc are in Go but something like 2*n*log(n) extra operations the same strings many times is significant to the UI (although I do sorts much more often, and only have to compute the string once per instance of a resource for the life of the page)

pkg/controllers/management/nodepool/util.go
```diff
@@ -0,0 +1,52 @@
...
+	s2 := nameRegexp.FindStringSubmatch(b)
+
+	if len(s1) < 2 || len(s2) < 2 {
+		// fallback onto lexicographical sorting
+		logrus.Warnf("Unable to sort nodes %s  %s naturally, fallback onto lexicographical sort", a, b)
+		return a < b
+	}
+
```
vincent99: This is ignoring the prefix string (`s1[1]` / `s2[1]`) when sorting.. they're _normally_ the same, but you can rename a pool after the cluster exists and end up with a mixed set of names in one pool.

pkg/controllers/management/nodepool/util.go
daxmc99: https://github.com/rancher/rancher/blob/6a0e89c1012c33c520729929143b1c032e22c533/pkg/controllers/management/nodepool/nodepool.go#L258
It looks like the backend should be doing this as well... Since we follow that convention for naming

Issue #21200, opened 2019-06-27
Standard users should not be allowed to create a template with enforced flag set as true


soumyalj:
Version: Master-head

Steps:

1. Login to the setup as an admin and create a standard user "user1"
2. As global admin, set the "cluster-template-enforcement" setting as true
3. Login as "user1" 
4. Create a cluster template "T1" in API with enforcement flag set.
5. The template T1 gets created successfully which should not be allowed. 

Only global admins should be able to set the enforcement flag

soumyalj:
Tested with build from master-head on July 3rd.
Steps:

Login to the setup as an admin and create a standard user "user1"
As global admin, set the "cluster-template-enforcement" setting as true
Login as "user1"
Create a cluster template "T1" in API with enforcement flag set through API
The template T1  does not get created with enforcement flag set for user1 as expected
Error below:
```
{
"baseType": "error",
"code": "PermissionDenied",
"message": "ClusterTemplate's enforced field cannot be changed",
"status": 403,
"type": "error"
}
```

PR #21270, opened 2019-07-01
Only Admin can set or change the clustertemplate.enforced field

prachidamle:
https://github.com/rancher/rancher/issues/21200

API validation to ensure ClusterTemplate can be enforced only by an Admin

### DIFF
--- a/pkg/api/store/clustertemplate/store.go
+++ b/pkg/api/store/clustertemplate/store.go
@@ -9,6 +9,8 @@ import (
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
+	gaccess "github.com/rancher/rancher/pkg/api/customization/globalnamespaceaccess"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	managementv3 "github.com/rancher/types/client/management/v3"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/api/store/clustertemplate/store.go
```diff
@@ -9,26 +9,40 @@ import (
...
+		Users:     users,
+		GrbLister: grbLister,
+		GrLister:  grLister,
 	}
 	return storeWrapped
 }
 
 type Store struct {
 	types.Store
+	Users     v3.UserInterface
```
mrajashree: Maybe we don't need to export these?

pkg/api/store/clustertemplate/store.go
```diff
@@ -157,3 +177,39 @@ func isDefaultTemplateRevision(apiContext *types.APIContext, id string) (bool, e
...
+		}
+	}
+	if enforcedFlagChanged {
+		//only admin can set the flag
+		ma := gaccess.MemberAccess{
+			Users:     p.Users,
+			GrLister:  p.GrLister,
+			GrbLister: p.GrbLister,
+		}
+		//if user is admin, no checks needed
```
mrajashree: This comment is misplaced maybe? :)

pkg/api/store/clustertemplate/store.go
```diff
@@ -157,3 +177,39 @@ func isDefaultTemplateRevision(apiContext *types.APIContext, id string) (bool, e
...
+		callerID := apiContext.Request.Header.Get(gaccess.ImpersonateUserHeader)
+		isAdmin, err := ma.IsAdmin(callerID)
+		if err != nil {
+			return err
+		}
+		if !isAdmin {
+			return httperror.NewAPIError(httperror.PermissionDenied, fmt.Sprintf("ClusterTemplate's %v field cannot be changed", managementv3.ClusterTemplateFieldEnforced))
```
mrajashree: Can we change the error message to show that non-admin user made this request hence enforcement can't be set?

prachidamle: Yeah, deliberately keeping the error message generic, reason is to not reveal any unnecessary rbac information.

Issue #21164, opened 2019-06-26
Users with Read-Only access type for a template should not be able to create template revisions under that template


soumyalj:
Version: Master-head

Steps:
1. Login to the setup as an admin and create a standard user "user1"
2. Create a cluster template "T1" in API and provide read-only access to "user1" as below:
```
"members": [

    {
        "accessType": "read-only",
        "groupPrincipalId": null,
        "type": "/v3/schemas/member",
        "userPrincipalId": "local://u-p5wdg"
    }

]

```
3. Login as "user1" who has Read-Only access to the template T1
4. Create a cluster template revision "CTR1" with rancherKubernetesEngineConfig in API which succeeds

"user1" should not be allowed to create new template revisions for this template T1

soumyalj:
Version: Master-head

Steps:

Login to the setup as an admin and create a standard user "user1"
Create a cluster template "T1" in API and provide read-only access to "user1" as below:
"members": [

    {
        "accessType": "read-only",
        "groupPrincipalId": null,
        "type": "/v3/schemas/member",
        "userPrincipalId": "local://u-p5wdg"
    }

]

Login as "user1" who has Read-Only access to the template T1
Create a cluster template revision "CTR1" with rancherKubernetesEngineConfig in API w
Creating a template revision errors out as below. This is as expected:
```
{
"baseType": "error",
"code": "PermissionDenied",
"message": "read-only member of clustertemplate cannot create revisions for it",
"status": 403,
"type": "error"
}

```

This was tested with AD auth enabled.

PR #21273, opened 2019-07-02
Forbid readonly clustertemplate members from creating revisions

mrajashree:
https://github.com/rancher/rancher/issues/21164

Read-only members of a clusterTemplate should not be able to create revisions for it.
This is because there's a chance such a revision could get set as the default revision on clusterTemplate, thus updating it, which a read-only member should not be allowed to.

### DIFF
--- a/pkg/api/store/clustertemplate/store.go
+++ b/pkg/api/store/clustertemplate/store.go
@@ -1,29 +1,37 @@
 package clustertemplate
 
 import (
-	"strings"
-
 	"fmt"
+	"strings"
 
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
+	"github.com/rancher/norman/types/values"
 	gaccess "github.com/rancher/rancher/pkg/api/customization/globalnamespaceaccess"
+	rrbac "github.com/rancher/rancher/pkg/controllers/management/globalnamespacerbac"
+	"github.com/rancher/rancher/pkg/namespace"
+	"github.com/rancher/rancher/pkg/ref"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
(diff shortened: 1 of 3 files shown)


pkg/api/store/clustertemplate/store.go
```diff
@@ -157,3 +169,35 @@ func isDefaultTemplateRevision(apiContext *types.APIContext, id string) (bool, e
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	if rb == nil {
+		return nil
+	}
+
+	for _, sub := range rb.Subjects {
+		if sub.Name == userID {
```
prachidamle: what happens if a template is public? It will have a read-only rolebinding created for *, what if a user tries to create a revision for such a template?

pkg/api/store/clustertemplate/store.go
```diff
@@ -209,6 +222,53 @@ func (p *Store) canSetEnforce(apiContext *types.APIContext, data map[string]inte
...
+	// else not allowed since, either user has no access or only read-only access by either:
+	// 1. being added explicitly as read-only member OR
+	// 2. template is public, which gives everyone read-only access
+	canUpdate := false
+	for _, accessType := range []string{rrbac.OwnerAccess, rrbac.MemberAccess} {
+		rbName, _ := rrbac.GetRoleNameAndVerbs(accessType, clusterTemplateName, rrbac.ClusterTemplateResource)
+		// check if rolebinding with this name exists and has the user (caller of this request) as a subject
+		rb, err := p.rbLister.Get(namespace.GlobalNamespace, rbName)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return err
```
prachidamle: Use httperror.WrapAPIError before returning the err?

mrajashree: This will be k8s server error and unwrapped errors will throw 500

prachidamle:
LGTM

PR #21287, opened 2019-07-02
Fix flakey rbac tests

rmweir:
Problem:
Upgrade tests occasionally fail with timeout, unauthorized, and modified errors.

Solution:
Stop waiting for workloads as it is not required to upgrade. Wait for app to reach "installing" state
before attempting to upgrade.

Related build:
https://drone-pr.rancher.io/rancher/rancher/2912

### DIFF
--- a/tests/integration/suite/test_rbac.py
+++ b/tests/integration/suite/test_rbac.py
@@ -4,6 +4,7 @@ from rancher import ApiError
 import time
 
 from .common import random_str
+from .test_catalog import wait_for_template_to_be_created
 from .conftest import wait_until_available, \
     cluster_and_client, user_project_client, \
     kubernetes_api_client, wait_for, ClusterContext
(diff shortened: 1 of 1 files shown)


tests/integration/suite/test_rbac.py
```diff
@@ -396,18 +397,23 @@ def test_appropriate_users_can_see_kontainer_drivers(user_factory):
     assert len(kds) == 0
 
 
-def test_readonly_cannot_perform_app_action(admin_mc, admin_pc, user_factory,
```
daxmc99: Should we just rip out user_factory all together? Looks like it is the source of these errors.

luthermonson:
not sure i have context like dax does for removing user_factory but this looks good to me.

tests/integration/suite/test_rbac.py
```diff
@@ -457,19 +466,35 @@ def test_member_can_perform_app_action(admin_mc, admin_pc, remove_resource,
...
+        # a timeout here is okay, the intention of the wait_for is to reach a
+        # steady state, this test is not concerned with whether an app reaches
+        # installing state or not
+        assert "Timeout waiting for condition" in str(e)
 
     user.client.action(
         obj=app,
         action_name="upgrade",
         answers={"asdf": "asdf"})
+
```
daxmc99: Should we be asserting anything here about the response from `user.client.action` like a 200?

Issue #21142, opened 2019-06-25
Removing the cluster template and cluster template revision associated with the cluster should not be allowed


soumyalj:
Version: Master-head

Steps:

1. Login to the rancher setup as an admin
2. Create "user1" as a standard user
3. Login as "user1"
4. Create a cluster template T1using the API
5. Create a cluster template revision R1 using the API 
6. Create a cluster using T1/R1
7. Edit the cluster and remove the T1 and R1 associated with the cluster and make them NULL in the UI API . This operation succeeds.
```
"clusterTemplateId": null,
"clusterTemplateRevisionId": null,
```
This operation should not be allowed. 

Note: In step 7 after T1/R1 are removed and any of the rkeconfigs are changed, cluster update is triggered and it takes into effect

soumyalj:
Tested using master-head build from July 8th with the steps above:
Error message obtained is as below:
{
"baseType": "error",
"code": "MissingRequired",
"fieldName": "Cluster Template",
"message": "this cluster is created from a clusterTemplate, please pass the clusterTemplate",
"status": 422,
"type"
}

The message should inform users to pass a template revision as cluster template is not a necessary field to create a cluster

soumyalj:
Tested with build from master-head with the steps above.
Error message is as expected:
```
{
"baseType": "error",
"code": "MissingRequired",
"fieldName": "ClusterTemplateRevision",
"message": "this cluster is created from a clusterTemplateRevision, please pass the clusterTemplateRevision",
"status": 422,
"type": "error"
}

```

PR #21295, opened 2019-07-02
Cluster edit cannot remove clustertemplate and revision id

prachidamle:
https://github.com/rancher/rancher/issues/21142


pkg/api/store/cluster/cluster_store.go
```diff
@@ -389,6 +389,8 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
 		}
 
 		data = clusterUpdate
+	} else if existingCluster[managementv3.ClusterSpecFieldClusterTemplateID] != nil {
```
mrajashree: Do we need to do the same for revision as well?

Issue #21170, opened 2019-06-26
Members removed by editing cluster templates still have access to template revisions he had access before


soumyalj:
Version: Master-head
Steps:

1. Login to the setup as an admin and create a standard user "user1" and provide access to user1 and user2
2. Create a cluster template "T1" in API
```
"members": [
{
"accessType": "member",
"groupPrincipalId": null,
"type": "/v3/schemas/member",
"userPrincipalId": "local://u-bcdef"
}

{
"accessType": "member",
"groupPrincipalId": null,
"type": "/v3/schemas/member",
"userPrincipalId": "local://u-776jm"
}
]

```
3. Create a cluster template revision "R1"
4. Edit the cluster template "T1" in API and remove "user1" 
```
"members": [
{
"accessType": "member",
"groupPrincipalId": null,
"type": "/v3/schemas/member",
"userPrincipalId": "local://u-776jm"
}
]

```

It was observed that user1 still has access to revision R1. "user1" should not have access to R1

soumyalj:
Tested using master-head build from July 10th. 
Login to the setup as an admin and create a standard user "user1" , "user2" and provide access to user1 and user2
Create a cluster template "T1" in API
Create a cluster template revision "R1"
Edit the cluster template "T1" in API and remove "user1"

"user1" does not have access to template revision R1 as expected.

Issue #21167, opened 2019-06-26
New members added after editing cluster templates do not have access to previously created template revisions


soumyalj:
Version: Master-head
Steps:

1. Login to the setup as an admin and create a standard user "user1"
2. Create a cluster template "T1" in API 
3. Create a cluster template revision "R1"
4. Edit the cluster template "T1" in API and add "user1" and provide a member access to "user1"
```
members": [ 
{
"accessType": "member",
"groupPrincipalId": null,
"type": "/v3/schemas/member",
"userPrincipalId": "local://u-qqjrr"
}
],
```

After editing and adding "user1" , previously added revision R1 is not available for use by "user1". 
https://abc/v3/clusterTemplateRevisions
```
{

    "type": "collection",
    "links": {
        "self": "…/v3/clusterTemplateRevisions"
    },
    "createTypes": {
        "clusterTemplateRevision": "…/v3/clustertemplaterevisions"
    },
    "actions": { },
    "pagination": {
        "limit": 1000,
        "total": 0
    },
    "sort": {
        "order": "asc",
        "reverse": "…/v3/clusterTemplateRevisions?order=desc",
        "links": {
            "uuid": "…/v3/clusterTemplateRevisions?sort=uuid"
        }
    },
    "filters": {
        "clusterTemplateId": null,
        "created": null,
        "creatorId": null,
        "enabled": null,
        "name": null,
        "removed": null,
        "uuid": null
    },
    "resourceType": "clusterTemplateRevision",
    "data": [ ]

}



```
New revisions added later are available for use.

Note: When access type is changed, existing users continue to have the same access type as before for existing template but new users added to the template have the correct access type.
Ex: Lets say user1 has member access to template T1 to start with. If access-type of user1 is changed to read-only for template T1, user1 still continues have member access to T1. 
But if user2 is added to the template T1 with read-only access, he has access to T1 as a read-only user.

Also, Users continue to have the same access type as before for existing template revision, but for new template revision access type is correct for the same user. 
Ex: if new revision R2 is added to the template T1, user1 behaves has read-only for R2. But user1 still has member access to R1 which he had before changing the access type

soumyalj:
Tested using master-head build from July 9. 
Version: Master-head
Steps:

Login to the setup as an admin and create a standard user "user1"
Create a cluster template "T1" in API
Create a cluster template revision "R1"
Edit the cluster template "T1" in API and add "user1" and provide a member access to "user1"
members": [ 
{
"accessType": "member",
[...]

PR #21298, opened 2019-07-03
Enqueue revisions correctly on clustertemplate update

mrajashree:
https://github.com/rancher/rancher/issues/21167
https://github.com/rancher/rancher/issues/21170

Clustertemplate controller should enqueue all revisions of the template, so that if there were any changes in the members list of template, rbac for revisions would also get handled. This enqueuing had a bug which prevented permissions on the revisions from getting updated


tests/integration/MANIFEST
```diff
@@ -15,6 +15,7 @@ suite/test_clustertemplate.py
 suite/test_default_roles.py
 suite/test_dns.py
 suite/test_dynamic_schemas.py
+suite/test_features.py
 suite/test_globaldns.py
 suite/test_hpa.py
 suite/test_ingress.py
```
prachidamle: Can we delete this file, it should be added to .gitignore

prachidamle:
LGTM

PR #21313, opened 2019-07-03
sync metadata setting after ns gets created

kinarashah:
If global namespace isn't found, we miss the init logic and then have to wait for refresh interval to take place


alena1108:
LGTM

Issue #21315, opened 2019-07-03
Incorrect system-charts branch used for master


leodotcloud:
`rancher/rancher:master-head` should be using `dev` branch of system-charts, but it's using `master` instead. Needs to be fixed.

leodotcloud:
Also, the default branch needs to be updated in the source code as well. https://github.com/rancher/rancher/blob/master/app/catalog_data.go#L24

drpebcak:
This is already correct in the drone.yaml if you are using the tagged image:
https://github.com/rancher/rancher/blob/master/.drone.yml#L39

Verified:
```
root@fc14a8f0cdaf:/var/lib/rancher/management-state/catalog-cache/2315ce81daaf962f49597b7cca91be75f66bb5b0bfe53d76d545da33f06a
2752# git remote show origin
* remote origin
  Fetch URL: https://git.rancher.io/system-charts
  Push  URL: https://git.rancher.io/system-charts
  HEAD branch: dev
  Remote branch:
    dev tracked
  Local branch configured for 'git pull':
    dev merges with remote dev
  Local ref configured for 'git push':
    dev pushes to dev (up to date)
root@fc14a8f0cdaf:/var/lib/rancher/management-state/catalog-cache/2315ce81daaf962f49597b7cca91be75f66bb5b0bfe53d76d545da33f06a
2752# git branch
* dev
```

drpebcak:
I think this was resolved, so I'm going to close this now.

PR #21316, opened 2019-07-03
changing the default system charts branch to dev

leodotcloud:
Addresses: https://github.com/rancher/rancher/issues/21315
Problem: In a development environment, the system-charts repo is pointing incorrectly to `master` branch.
Fix: Change to `dev`


cjellick:
cant merge this. it has impacts on upgrades. there is some convoluted logic we needed to introduce in order to properly transition from a hard-coded branch to one supplied by the build

PR #21317, opened 2019-07-03
[Backport v2.2] Package catalog into build v2.2

dramich:
https://github.com/rancher/rancher/issues/21318


dramich:
Old and unsupported

PR #21320, opened 2019-07-03
Check dev mode for system charts

rmweir:
Problem:
users in dev mode were on the master system charts branch. If env
variable was forgotten, it would not have been immediately
apparent.

Solution:
Now, if in dev mode system charts default branch is set to dev.
If not in dev mode and no branch has been set to CATTLE_
_SYSTEM_CHART_DEFAULT_BRANCH, rancher will panic.


cjellick:
lgtm

Issue #21323, opened 2019-07-03
Altering - cluster-altering app shows error after upgrading Rancher


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- run Rancher: v2.2.4 container
- as admin, enable Github auth
- log in as Github user user1, add an RKE DO cluster
- as user1, adds cluster notifier and cluster alerting
- upgrade Rancher to v2.2.5-rc8

**Result:**
- the cluster altering still works, but `cluster-altering` app shows the following error
<img width="1480" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/60628988-c2379580-9da8-11e9-983b-e9268412a9f6.png">

<img width="1424" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/60628645-3cffb100-9da7-11e9-9988-802475bee160.png">

- and it will cause the cluster unavailable periodicity
<img width="1412" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/60628652-438e2880-9da7-11e9-9ba7-4fdd44ed6603.png">

```
2019/07/03 22:03:05 [ERROR] Failed to send alert: Post http://alertmanager-operated.cattle-prometheus.svc.cluster.local:9093/api/v1/alerts: dial tcp: lookup alertmanager-operated.cattle-prometheus.svc.cluster.local on 67.207.67.2:53: no suc
2019/07/03 22:03:05 [ERROR] Failed to send alert: Post http://alertmanager-operated.cattle-prometheus.svc.cluster.local:9093/api/v1/alerts: dial tcp: lookup alertmanager-operated.cattle-prometheus.svc.cluster.local on 67.207.67.2:53: no suc
2019/07/03 22:03:05 [ERROR] Failed to send alert: Post http://alertmanager-operated.cattle-prometheus.svc.cluster.local:9093/api/v1/alerts: dial tcp: lookup alertmanager-operated.cattle-prometheus.svc.cluster.local on 67.207.67.2:53: no suc
```

- and after the API server recovers and the cluster is `active`, the app `cluster-alerting` is still in the bad state as the above screenshot shows.

loganhz:
It looks like api server is unhealthy or it is due to network issue. All apps will hit this issue not just alerting.

jiaqiluo:
@loganhz shows me that the `cluster-alerting` app get redeployed successfully and the error is gone after editing and save it.

[...]

PR #21331, opened 2019-07-04
Check cluster ready before upgrade system service

aiwantaozi:
Problem:
got cluster not ready error when upgrade rancher server

Solution:
check cluster ready before upgrade system service

Issue:
https://github.com/rancher/rancher/issues/21323

### DIFF
--- a/pkg/controllers/user/alert/deployer/upgradeimpl.go
+++ b/pkg/controllers/user/alert/deployer/upgradeimpl.go
@@ -35,6 +35,7 @@ const (
 
 type alertService struct {
 	clusterName        string
+	clusterLister      v3.ClusterLister
 	apps               projectv3.AppInterface
 	oldClusterAlerts   v3.ClusterAlertInterface
 	oldProjectAlerts   v3.ProjectAlertInterface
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/alert/deployer/upgradeimpl.go
```diff
@@ -132,6 +134,15 @@ func (l *alertService) Upgrade(currentVersion string) (string, error) {
 	newApp.Spec.Answers["operator.enabled"] = "false"
 
 	if !reflect.DeepEqual(newApp, app) {
+		// check cluster ready before upgrade, because helm will not retry if got cluster not ready error
+		cluster, err := l.clusterLister.Get(metav1.NamespaceAll, l.clusterName)
+		if err != nil {
+			return "", fmt.Errorf("get cluster %s failed, %v", l.clusterName, err)
+		}
+		if !v3.ClusterConditionReady.IsTrue(cluster) {
```
orangedeng: Should we check this at the beginning? So we won't do anything when the cluster is not ready.

aiwantaozi: if we check on the beginning, it may throw several errors when server starts, so I only check the cluster ready status when we need to upgrade the app.

orangedeng:
LGTM

Issue #19915, opened 2019-05-01
[BUG] Custom ingress-ip-domain results in no automatic generated ingress hostnames


gabrielcalderon:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
* Change the default ingress-ip-domain from xip.io to nip.io in the global rancher settings.
* Deploy docker-registry from Rancher Charts Catalog using a layer-7 ingress with auto-generated nip.io hostname.

**Result:**
* Ingress gets created with nip.io/ instead of auto-generated path.
* Rancher works just fine with xip.io, changing to any other values results in no hostnames generated.

**Logs**
When setting an ingress with xip.io I can see a message in rancher server:
- [INFO] Updating ingress [docker-registry:docker-registry] with public endpoints [ json data here]

When setting an ingress with anything other than xip.io, for example: nip.io I get:
- [INFO] Updating ingress [docker-registry:docker-registry] with public endpoints [null]

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
Rancher 2.2.2
Running in bare-metal with 8443:443 and 8080:80
Single-Node Instance
I'm not setting the node IP, or internal cluster IP
```

- Docker version (use `docker version`):

```
docker version 18.09.5, build e8ff056
```

Other info:
* https://nip.io/ is a PowerDNS wildcard dns like xip.io but the backend runs Python instead of Bash.

azubizarretap:
Seeing the same behavior as @gabrielcalderon . I am also running rancher on bare metal.

zhmocean:
same on v2.2.3 stable

rootwuj:
Tested with rancher/rancher:master-head from July 11

This version uses the ingress-ip-domain other than xip.io, the ingress target can automatically generate the hostname.

PR #21339, opened 2019-07-05
Fix dns auto generate issue

orangedeng:
**Problem:**
No effect when `xip.io` suffix is changed to other suffix.

**Solution:**
The ingresshostgen controller is using the `IngressIPDomain` setting.
So the controller should run in Rancher side.

**Issue:**

https://github.com/rancher/rancher/issues/19915


loganhz:
LGTM

cjellick:
lgtm

Issue #13057, opened 2018-04-25
Cannot deploy catalog apps which use multiple namespaces?


kenfdev:
**Rancher versions:**
rancher/server: v2.0.0-beta4-rc2
rancher/agent: v2.0.0-beta4-rc2

**Infrastructure Stack versions:**
kubernetes (if applicable): 1.10.1-rancher1

**Docker version: (`docker version`,`docker info` preferred)**
Both Rancher Server/Node
```
Client:
 Version:      17.03.2-ce
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   f5ec1e2
 Built:        Tue Jun 27 03:35:14 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.03.2-ce
 API version:  1.27 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   f5ec1e2
 Built:        Tue Jun 27 03:35:14 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
Server
```
root@rancher-server:~# cat /etc/os-release
NAME="Ubuntu"
VERSION="16.04.1 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.1 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
root@rancher-server:~# uname -r
4.4.0-53-generic
```

Node
```
root@ubuntu1:~# cat /etc/os-release
NAME="Ubuntu"
VERSION="16.04.2 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.2 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

**Setup details: (single node rancher vs. HA rancher, internal DB vs. external DB)**
single node rancher and single worker node.

**Steps to Reproduce:**
[...]

PR #21340, opened 2019-07-05
ARGO-2037: Add namespace templates so that rancher's helm deploys out of spinnaker.

FayeGibbins:
ATM spinnaker can't deploy rancher via rancher's helm chart in anything but the spinnaker's default namespace because rancher's helm chart does not consistently conform to the requirements of namespace lines.
In fact the namespace lines are applied inconsistently leading to the deployment of rancher (via helm, in spinnaker) being non-viable.
This PR fixes this.
Here is the spinnaker docs regarding this: https://www.spinnaker.io/guides/user/kubernetes-v2/deploy-helm/#configure-the-bake-manifest-stage

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -2,6 +2,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: {{ template "rancher.fullname" . }}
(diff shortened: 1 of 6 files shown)


jburnitz:
Related to: https://github.com/rancher/rancher/issues/13057

PR #21347, opened 2019-07-05
Add cloudscale.ch driver.

alakae:
This adds the [cloudscale.ch](https://www.cloudscale.ch/en/) machine and ui-driver to the list of drivers available for installation with rancher.

Our drivers are maintained at:
 * https://github.com/cloudscale-ch/ui-driver-cloudscale
 * https://github.com/cloudscale-ch/docker-machine-driver-cloudscale

Screenshot

### DIFF
--- a/app/machinedriver_data.go
+++ b/app/machinedriver_data.go
@@ -51,6 +51,11 @@ func addMachineDrivers(management *config.ManagementContext) error {
 		[]string{"objects-east.cloud.ca"}, false, false, management); err != nil {
 		return err
 	}
+	if err := addMachineDriver("cloudscale", "https://github.com/cloudscale-ch/docker-machine-driver-cloudscale/releases/download/v1.0.0/docker-machine-driver-cloudscale_v1.0.0_linux_amd64.tar.gz",
+		"https://objects.cloudscale.ch/cloudscale-rancher-v2-ui-driver/component.js", "73eb95d58706b53076baf9baef8df720b97c06f7140f607fba6a65d0735a179c",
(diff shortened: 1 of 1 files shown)


alakae:
I am looking into why the CI failed.

alakae:
CI is now passing.

daxmc99:
Hi @alakae, 
Thanks for the PR. I am hoping to be able to take a look at it this week.

vincent99:
The UI component URL is loaded directly by the user's browser, so its hostname does not need to be whitelisted.  

The component _does_ make AJAX calls `https://api.cloudscale.ch` though.  It's making them directly to the API and therefore requires CORS, but no CORS headers are sent back with e.g. auth error responses (e.g. the token is wrong) so you will be unable to see the response body. (They are sent back for successful responses).

So that should either always return CORS headers, or be changed to go through the Rancher proxy `/meta/proxy`, in which case `api.cloudscale.ch` needs to be whitelisted.

vincent99:
What I said above is wrong, the UI CSS/JS does get proxied so the domain (`objects.cloudscale.ch`) does need to be whitelisted.

You do still probably want to fix CORS on errors or use the proxy for API calls.

alakae:
Hi Vincent, I am currently investigating how we can fix the „CORS on errors„ issue. I hope that we can roll-out the change next week. Thanks a lot for your help.

alakae:
@vincent99 Thanks again for brining our attention to the CORS misconfiguration. We have adapted the configuration of the web server. The CORS headers are now always returned. If an invalid token is specified the error message is now visible to the user. The merge request does not need to be changed.

PR #21363, opened 2019-07-08
Check cluster ready before upgrade system service

alena1108:
Problem:
got cluster not ready error when upgrade rancher server

Solution:
check cluster ready before upgrade system service


alena1108:
https://github.com/rancher/rancher/issues/21364

PR #21391, opened 2019-07-09
fix upgrade system image failed when manage multiple clusters

aiwantaozi:
Problem:
only one cluster system server upgrade success when there are multiple
cluster

Solution:
this cause by only one alert/logging/pipeline upgradeService was init,
when there are multiple cluster, the former clusterName will be
overwrite by the laster one. Now change handle multiple cluster when
init upgradeService

Issue:
https://github.com/rancher/rancher/issues/21455
https://github.com/rancher/rancher/issues/21323
https://github.com/rancher/rancher/issues/20975

### DIFF
--- a/pkg/controllers/user/alert/deployer/upgradeimpl.go
+++ b/pkg/controllers/user/alert/deployer/upgradeimpl.go
@@ -1,15 +1,14 @@
 package deployer
 
 import (
-	"context"
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/rancher/norman/controller"
 	alertutil "github.com/rancher/rancher/pkg/controllers/user/alert/common"
 	"github.com/rancher/rancher/pkg/controllers/user/helm/common"
-	"github.com/rancher/rancher/pkg/controllers/user/systemimage"
 	monitorutil "github.com/rancher/rancher/pkg/monitoring"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/ref"
@@ -26,16 +25,18 @@ import (
 )
 
 var (
-	serviceName = "alerting"
+	ServiceName             = "alerting"
+	waitCatalogSyncInterval = 60 * time.Second
 )
 
 const (
 	defaultGroupIntervalSeconds = 180
 )
 
-type alertService struct {
+type AlertService struct {
 	clusterName        string
 	clusterLister      v3.ClusterLister
+	catalogLister      v3.CatalogLister
 	apps               projectv3.AppInterface
 	oldClusterAlerts   v3.ClusterAlertInterface
 	oldProjectAlerts   v3.ProjectAlertInterface
@@ -45,14 +46,15 @@ type alertService struct {
 	projectAlertRules  v3.ProjectAlertRuleInterface
 	projectLister      v3.ProjectLister
 	namespaces         v1.NamespaceInterface
+	templateLister     v3.CatalogTemplateLister
 	appDeployer        *appDeployer
 }
 
-func init() {
-	systemimage.RegisterSystemService(serviceName, &alertService{})
+func NewService() *AlertService {
+	return &AlertService{}
 }
 
-func (l *alertService) Init(ctx context.Context, cluster *config.UserContext) {
+func (l *AlertService) Init(cluster *config.UserContext) {
 	ad := &appDeployer{
 		appsGetter:       cluster.Management.Project,
 		namespaces:       cluster.Management.Core.Namespaces(metav1.NamespaceAll),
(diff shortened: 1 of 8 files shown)


pkg/controllers/user/systemimage/systemimage.go
```diff
@@ -53,7 +59,10 @@ func (s *Syncer) Sync(key string, obj *v3.Project) (runtime.Object, error) {
 	}
 
 	changed := false
+	systemServices := getSystemService()
 	for k, v := range systemServices {
+		v.Init(s.userContext)
```
kinarashah: Question, why do we init every time? maybe we check if it's been already initialized, since this is on every project sync?

aiwantaozi: updated

orangedeng:
I would prefer you keep the current registration mechanism. Like logging upgrade implementation register to the upgrade service. I think it would be less code changed and easy to read.

orangedeng:
For example, I would prefer you modify `SystemService` interface like following:
```
type SystemService interface {
	Init(ctx context.Context, cluster *config.UserContext) SystemService
	Upgrade(currentVersion string) (newVersion string, err error)
	Version() (string, error)
}
```
And in the `system-image-upgrade-controller` `Register` function, you can do:
```
func Register(ctx context.Context, cluster *config.UserContext) {
	var sList []SystemService
	for _, v := range systemServices {
		sList = append(sList, v.Init(ctx, cluster))
	}

	projClient := cluster.Management.Management.Projects("")
	syncer := Syncer{
		clusterName:      cluster.ClusterName,
		projects:         projClient,
		projectLister:    projClient.Controller().Lister(),
		daemonsets:       cluster.Apps.DaemonSets(""),
		daemonsetLister:  cluster.Apps.DaemonSets("").Controller().Lister(),
		deployments:      cluster.Apps.Deployments(""),
		deploymentLister: cluster.Apps.Deployments("").Controller().Lister(),
		systemServices:   sList,
	}
	projClient.AddClusterScopedHandler(ctx, "system-image-upgrade-controller", cluster.ClusterName, syncer.Sync)

}
```

aiwantaozi:
Because when project init, the Version() function got called, it will return the init version for system service, at this time, we don't need to init those listers. So I create a NewService function. For the Init function return SystemService, Init function is called by SystemService itself, if it returns SystemService, may duplicate.

orangedeng:
Got it. LGTM

pkg/controllers/user/alert/deployer/upgradeimpl.go
```diff
@@ -140,17 +154,29 @@ func (l *alertService) Upgrade(currentVersion string) (string, error) {
...
+		if err != nil {
+			return "", fmt.Errorf("get catalog %s failed, %v", systemCatalogName, err)
+		}
+
+		if !v3.CatalogConditionUpgraded.IsTrue(systemCatalog) || !v3.CatalogConditionRefreshed.IsTrue(systemCatalog) || !v3.CatalogConditionDiskCached.IsTrue(systemCatalog) {
+			return "", fmt.Errorf("catalog %v not ready", systemCatalogName)
+		}
+
+		// wait for catalog syncing to latest version
+		time.Sleep(waitCatalogSyncInterval)
```
kinarashah: want to understand, what's this for? we already checked catalog has been refreshed in Line 165?

aiwantaozi: updated, also watch catalog, because latest catalog may not download completely the first time this controller execute after server upgraded

pkg/controllers/user/logging/deployer/upgradeimpl.go
```diff
@@ -84,7 +86,7 @@ func (l *loggingService) Upgrade(currentVersion string) (string, error) {
 		return "", fmt.Errorf("get cluster %s failed, %v", l.clusterName, err)
 	}
 	if !v3.ClusterConditionReady.IsTrue(cluster) {
-		return "", fmt.Errorf("upgrade system service %s failed, cluster %v not ready", serviceName, l.clusterName)
+		return "", fmt.Errorf("cluster %v not ready", l.clusterName)
```
kinarashah: should this logic be similar to alerting flow? I think that's better, check cluster and catalog conditions only if app actually changed 

or in any case, we should make it consistent for better understanding or comment why it's different

aiwantaozi: We used to check whether old version and latest version in temple for logging and alerting, but because of a bug in chart, we need to redeploy alerting chart without version change, the answer for alerting app is changed, so version comparis on is replaced by app comparison. So the pr keep normal version comparison in logging.

loganhz:
LGTM https://github.com/rancher/rancher/pull/21393#issuecomment-510095826

pkg/controllers/user/logging/deployer/upgradeimpl.go
kinarashah: got it, thank you

PR #21393, opened 2019-07-09
For 2.2: fix upgrade system image failed when manage multiple clusters

aiwantaozi:
Problem:
only one cluster system server upgrade success when there are multiple
cluster

Solution:
this cause by only one alert/logging/pipeline upgradeService was init,
when there are multiple cluster, the former clusterName will be
overwrite by the laster one. Now change handle multiple cluster when
init upgradeService

Issue:
https://github.com/rancher/rancher/issues/21458


orangedeng:
LGTM

orangedeng:
This PR got rebased, LGTM

Issue #20629, opened 2019-06-03
Backport: Add in rancher minimum and maximum versions to use with helm charts/catalog service


dramich:
https://github.com/rancher/rancher/issues/18469

deniseschannon:
v2.2.x has it's own dedicated branch for charts. The min/max version for 2.2 is not needed anymore.

PR #21408, opened 2019-07-09
[Backport v2.2] Rancher version verification

dramich:
https://github.com/rancher/rancher/issues/20629


alena1108:
If we add it to 2.2.x, we should merge it after 2.2.5 ships

Issue #19751, opened 2019-04-22
Workload's scheduler, priority and priority class name not working


gehring:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
When deploying some workload, set some value in any of the fields `Scheduler`, `Priority` or `Priority Class Name`, then deploy and check the generated yaml.

**Result:**
Those fields are ignored, no entries are added for those fields.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.2
- Installation option (single install/HA): Single

sangeethah:
Tested with latest build from master - 7f8c1191

Deploy workload by setting values for Scheduler, Priority or Priority Class Name.
These values are preserved in Edit workload page and in  "View/Edit YAML"
```
      priority: 3
      priorityClassName: high
      restartPolicy: Always
      schedulerName: testnew
```

Also they get set on the container spec as expected:
```
                    "spec": {
                        "containers": [
                            {
                                "image": "ubuntu:xenial",
                                "imagePullPolicy": "Always",
                                "name": "hello123",
                                "resources": {},
                                "securityContext": {
                                    "allowPrivilegeEscalation": false,
                                    "capabilities": {},
                                    "privileged": false,
                                    "readOnlyRootFilesystem": false,
                                    "runAsNonRoot": false
                                },
                                "stdin": true,
                                "terminationMessagePath": "/dev/termination-log",
                                "terminationMessagePolicy": "File",
                                "tty": true
                            }
                        ],
                        "dnsPolicy": "ClusterFirst",
                        "priority": 3,
                        "priorityClassName": "high",
                        "restartPolicy": "Always",
                        "schedulerName": "testnew",
                        "securityContext": {},
                        "terminationGracePeriodSeconds": 30
                    }
```

PR #21421, opened 2019-07-10
Update types and add workload scheduling integration tests

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/19751

Depend on types PR: https://github.com/rancher/types/pull/901

Note:
Only `scheduler` is used in the test, because the priority plugin is removed in the current v1.12.2-lite5 k3s vendor, thus `priority/priorityClassName` is not available in the embeded local cluster.


orangedeng:
LGTM

ibuildthecloud:
LGTM

Issue #21405, opened 2019-07-09
Backport: K8s upgrade isn't propagated to nodes manually tainted by users


alena1108:
Backport https://github.com/rancher/rancher/issues/16050

alena1108:
The fix is available in v2.2.5-rc12

sowmyav27:
@superseb 

verified the below scenario

1. Had Rancher setup on v2.2.3
2. Deployed a cluster - 1 control plane, 1 worker and 31 etcd nodes.
3. Added taints on all the three nodes.
4. Upgraded the rancher server to v2.2.4
5. Noticed the `cattle-node-agent` and `cattle-cluster-agent` fail to deploy. 
6. upgraded to v2.2.5-rc12(which has the fix)

**Observations:**
`cattle-cluster-agent` is stuck in `Scheduling` state.
`cattle-node-agent` has upgraded to correct version `rancher-agent:v2.2.5-rc12`
`kube-api-auth` has failed to deploy. there are no pods available.
<img width="1335" alt="Screen Shot 2019-07-09 at 4 08 28 PM" src="https://user-images.githubusercontent.com/26032343/60929089-b5151d80-a264-11e9-9c22-6bdb27dbdb05.png">

superseb:
Scenario 1:
- `cattle-cluster-agent` only tolerates etcd and controlplane taints, if those nodes are tainted manually, it won't be able to schedule. Before this it didnt have any tolerations, which made it schedule to workers only (and without workers, it wouldn't be scheduled)
- `cattle-node-agent` allows everything and will run everywhere
- `kube-api-auth` has the same tolerations as `cattle-cluster-agent`

Scenario 2:
Based on the information above, it all works as it is intended to in the logic of the PR.

@alena1108 2 options: allow all taints also for the `cattle-cluster-agent` and `kube-api-auth`, with preference for etcd/controlplane for `cattle-cluster-agent` or document that one of etcd or controlplane nodes can't be manually tainted to make sure the needed pods are scheduled.

sowmyav27:
### 2.2.3 to 2.2.4 to 2.2-head upgrade
**In upgraded 2.2.4**

1 etcd, 1 control plane and 1 worker node
```
cattle-cluster-agent — 1 pod - fails to deploy
cattle-node-agent — fails to deploy - 0 pods
Kube-api-auth — on 3 pods on all 3 nodes - etcd, control plane, worker nodes
```
3 etcd, 2 control plane and 2 worker nodes
```
[...]

PR #21436, opened 2019-07-10
[2.2 backport] Allow all taints for agents and kube-api-auth

alena1108:
Problem: Manually tainted etcd/controlplane nodes won't get cluster-agent or kube-api-auth scheduled

Solution: Allow all taints for agents and kube-api-auth, restrict kube-api-auth to controlplane nodes


alena1108:
https://github.com/rancher/rancher/issues/21405

Issue #21416, opened 2019-07-10
Altering - cluster-altering app shows errors after upgrading Rancher


jiaqiluo:
Steps:
- run Rancher: v2.2.3 container
- as admin, enable Github auth
- log in as Github user user1
- add the following three clusters: DO single-node, DO 3-node, and GKE
- adds cluster notifier and cluster alerting for all three clusters
- upgrade Rancher to v2.2.5-rc12

Results:
- for all three clusters, the cluster altering still work
- for all three clusters, adding a new alter group and new rule also works
- for all three clusters, the app `cluster-alerting` is stuck in `installing` state, like this:
<img width="1463" alt="from v2 2 3 to v2 2 5-rc12" src="https://user-images.githubusercontent.com/6218999/60932137-7f2a6600-a271-11e9-95f5-2754e7563f48.png">

- the app `cluster-alerting` in one cluster shows the following error
```
Failed to install app cluster-alerting. Error: UPGRADE FAILED: the server is currently unable to handle the request (get configmaps)
```
<img width="1479" alt="Screen Shot 2019-07-09 at 5 47 48 PM" src="https://user-images.githubusercontent.com/6218999/60932185-b13bc800-a271-11e9-9b6e-9bdc6c4f3c87.png">

- in another two clusters, it shows this error
```
Failed to install app cluster-alerting. Error: stat /rancher-monitoring: no such file or directory
```
<img width="1456" alt="from v2 2 3 to v2 2 5-rc12 " src="https://user-images.githubusercontent.com/6218999/60932219-d6c8d180-a271-11e9-96c9-4f6fc6a3fc97.png">

One more test:
- add a new DO single-node cluster in the same setup 
- enable the cluster altering 

Results:
- the cluster altering works

sangeethah:
To get the cluster-altering to active state , "Click upgrade and save".

cjellick:
most straight foward steps to repro:

1. Start rancher 2.2.3
2. Add 3~5 clusters in the cluster.
3. Add notifier in each cluster
4. Edit cluster alert group and add a recipient
5. Check if alerting is enabled by looking at apps in system project
6. Upgrade rancher to 2.2.5-rc12
7. Check the status of monitoring-operator apps 


Cannot repro if `CATTLE_DEV_MODE` is used to disable jailing.

loganhz:
Some error message might be useful. Not sure this is a Jail issue or helm controller issue.

```
[...]

PR #21445, opened 2019-07-10
Make helm jail name unique

dramich:
Problem:
The helm jail is based off the app name so if multiple apps with the
same name come through the jail can get deleted out from under other
running processes

Solution:
Make the jail name unique

https://github.com/rancher/rancher/issues/21416


cjellick:
LGTM

Issue #21112, opened 2019-06-25
Logging integration doesn't work if Docker Root is not default /var/lib/docker


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Provision a server with Docker, but use a path other than /var/lib/docker for Docker's root. For example /data/var/lib/docker
2. Create a new custom cluster and add host above.
3. Set the Docker Root Directory to /data/var/lib/docker
4. Enable the logging integration.

**Result:**
The rancher-fluentd-logger is using /var/lib/docker as the dockerroot and will likely not be able to follow the symbolic links in /var/lib/rancher/rke/log/ to access the logs.

![image](https://user-images.githubusercontent.com/468462/60071457-02bf5100-96cf-11e9-8a27-f14024cd1a55.png)

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.3.0-alpha5 (also reproducible on v2.2.4)
- Installation option (single install/HA):
single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
t3.medium
- Kubernetes version (use `kubectl version`):
1.14.3

- Docker version (use `docker version`):

```
$ docker version
Client:
 Version:           18.06.3-ce
 API version:       1.38
 Go version:        go1.10.4
 Git commit:        d7080c1
 Built:             Wed Feb 20 02:24:22 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.3-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       d7080c1
  Built:            Wed Feb 20 02:25:33 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

burakhoban:
Have you tried updating the pod via kubectl (or via UI)?

E.g. something like

- `kubectl get pod <container> -n cattle-logging --export -o yaml >> fluent.yaml`
- change path for dockerroot
- `kubectl apply -f fluent.yaml -n cattle-logging`

We had to do something similar when it came to cattle-agent falling over when our immediate CA certificates expired (there's no other nice way of doing it).

dnoland1:
[...]

Issue #21240, opened 2019-06-30
Pod logs folder was not generated


wongkimshing:
**What kind of request is this (question/bug/enhancement/feature request):**
Pod logs folder was not generated after first pod deployed. As being mentioned in the documentation, /var/log/containers folder should be available for gathering all pod logs, but it didn’t happen anyway. So it caused no log data could be sent to Elasticsearch after configuring logging for cluster even though connection testing is successful.

**Steps to reproduce (least amount of steps as possible):**
1. Install Rancher Server
2. Create custom type of Cluster
3. Install Rancher Agent in the other server
4. Deploy an application on the Cluster just created
5. Navigate Tools -> Logging and select Elasticsearch and configure properly with successful connection testing
6. Only see the connection testing log in Kibana and no application logs are received 

**Result:**
/var/log/containers was not generated after 1st pod deployed. Logging function was not available.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.4
- Installation option (single install/HA): single install
- OS: CentOS 7.6

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): Aliyun Cloud, 8 CPU/32G RAM
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):
17.03 ce
```
(paste the output here)
```

aiwantaozi:
Hi, @wongkimshing  ```/var/log/containers``` was generated by kubelet, could you check
1. whether this pod was scheduled to that node
2. whether docker logs $container_id return logs
3. check kubelet logs if there are any errors

wongkimshing:
Yes, I can see the folder now, but it still has not log data sending to Elasticsearch

wongkimshing:
Upgrading the system project -> apps rancher-logging and add answer fluentd.cluster.dockerRoot = you docker root dir can fix the issue manually.

wongkimshing:
@aiwantaozi I followed your instructions and resolved the log problem, but the other problem occurred. Path /var/lib/rancher/fluentd/buffer grows up very soon and it occupies most of the disk spaces. Could you please advise how can I change the parameters to config the buffer path of Fluentd and limit it to certain size? Thanks a lot.

aiwantaozi:
[...]

PR #21467, opened 2019-07-11
Fixing docker root not set in logging

aiwantaozi:
Problem:
logging not use docker root in cluster

Solution:
add docker root when enable logging

Issue:

### DIFF
--- a/pkg/controllers/user/logging/deployer/appdeployer.go
+++ b/pkg/controllers/user/logging/deployer/appdeployer.go
@@ -3,6 +3,7 @@ package deployer
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	loggingconfig "github.com/rancher/rancher/pkg/controllers/user/logging/config"
@@ -64,7 +65,7 @@ func (d *AppDeployer) cleanup(appName, appTargetNamespace, projectID string) err
 	return nil
 }
 
-func (d *AppDeployer) deploy(app *projectv3.App) error {
+func (d *AppDeployer) deploy(app *projectv3.App, systemWriteKeys []string) error {
 	if err := d.initNamespace(app.Spec.TargetNamespace); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/controllers/user/logging/deployer/appdeployer.go
```diff
@@ -83,6 +84,15 @@ func (d *AppDeployer) deploy(app *projectv3.App) error {
 		return errors.New("stale app " + app.Namespace + ":" + app.Name + " still on terminating")
 	}
 
+	app.Spec.AppRevisionName = current.Spec.AppRevisionName
+	if !reflect.DeepEqual(current.Spec, app.Spec) {
+		new := current.DeepCopy()
+		new.Spec = app.Spec
+		_, err = d.AppsGetter.Apps(app.Namespace).Update(new)
```
gitlawr: Now, every time cluster sync happens we will generate the logging app spec and update the current app to it. This can override any user custom answers. We only do create before and do not update the app, so would like to confirm if this is an expected change.

aiwantaozi: updated

gitlawr:
LGTM

wongkimshing:
@aiwantaozi What should I do to update the patch to my Production environment, please ?

aiwantaozi:
@wongkimshing You could upgrade the `system project` -> apps `rancher-logging` and add answer `fluentd.cluster.dockerRoot` = `you docker root dir` to fix the issue manully

aiwantaozi:
This issue may be related too. https://github.com/rancher/rancher/issues/21240

wongkimshing:
@aiwantaozi I have upgraded the system project -> apps rancher-logging and add answer fluentd.cluster.dockerRoot = my docker root dir, but the log still were not sent to Elasticsearch. Please advise. Thanks.

wongkimshing:
@aiwantaozi I followed your instructions and resolved the log problem, but the other problem occurred. Path **/var/lib/rancher/fluentd/buffer** grows up very soon and it occupies most of the disk spaces. Could you please advise how can I change the parameters to config the buffer path of Fluentd and limit it to certain size? Thanks a lot.

PR #21469, opened 2019-07-11
Forwardport: Whitelist envvars for jail

dramich:
Problem:
The jail clears the env and sets only ones required to run. This drops
envvars that might be needed for a proxy

Solution:
Add a whitelist of envvars to pull into the jail with a default list to
support proxies or user updated setting

https://github.com/rancher/rancher/issues/20709


cjellick:
lgtm

Issue #21401, opened 2019-07-09
Editing access type of members in the cluster template does not get reflected


soumyalj:
Version: Master-head
Steps:

1. Login to the setup as an admin and create a standard user "user1"
2. Create a cluster template "T1" in API
3. Create a cluster template revision "R1"
4. Edit the cluster template "T1" in API and add "user1" and provide a member access to "user1"
members": [ 
{
"accessType": "member",
"groupPrincipalId": null,
"type": "/v3/schemas/member",
"userPrincipalId": "local://u-qqjrr"
}
],
5. Edit the access type of user1 in the template to read-only as an admin
```
"members": [
{
"accessType": "read-only",
"groupPrincipalId": null,
"type": "/v3/schemas/member",
"userPrincipalId": "local://u-n66pv"
}
]
```
Now as "user1", edit the template T1 or revision R1. Both succeed. 
The read-only user should not have access to edit the template or revision.

For new revisions added, user1 behaves as a read-only member as expected.

soumyalj:
Tested with build from master-head. 
Tested the use case above where a member access is changed to read-only and verified the user has read-only access. 
The below access type change use cases were tested and it is functional:
1. read-only to member
2. read-only to owner
3. member to read-only
4. member to owner
5. owner to member
6. owner to read-only

PR #21479, opened 2019-07-11
Addresses a bug in updating rolebindings for clustertemplate

mrajashree:
https://github.com/rancher/rancher/issues/21401

If a clustertemplate was shared with a user giving "member" access, a rolebinding
was getting created with role containing "update" verb for that user. If the access
type was then switched to read-only, user still had update access. This is because
the rbac controller wasn't handling the case when there were no longer any members
for a particular access type


prachidamle:
LGTM

PR #21484, opened 2019-07-12
Add Istio VirtualService and DestinationRule to Rancher API

loganhz:
Address issue: https://github.com/rancher/rancher/issues/20894
Depend on types PR: https://github.com/rancher/types/pull/870

Notes:
For project scoped resources we add `projectId` field in API layer in `https://github.com/rancher/norman/blob/master/api/handler/validate.go`. For native kubernetes resources like `service/deployment` apiserver will drop it in deserialization. But for CRDs like Istio resources, it is stored as part of the manifest, and fails the validation from pilot because it is an undefined field. Therefore we update the projectsetter store to drop the `projectId` field. It is unnecessary because the store reads `projectId` from the namespace object.


loganhz:
Updated types to use master.

Based on https://github.com/rancher/rancher/pull/20910

Already lgtm in https://github.com/rancher/rancher/pull/20910

cjellick:
lgtm

PR #21486, opened 2019-07-12
Fix istio integration tests

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/21492

Problem:
Tests fail when istio tests are distributed to different workers. Module-scope fixtures/teardown can run multiple times in this case, causing race condition that the 1st test may clean up istio CRDs while the
2nd test is running.

Solution:
Run istio API tests in nonparallel.


orangedeng:
LGTM

Issue #21310, opened 2019-07-03
K8s metadata - change metadata url in settings


kinarashah:
Currently the URL to data.json includes branch name and that gets cached, so it doesn't immediately reflect the changes after pushing to repository. 

Workaround:
Editing the URL to different branch, or changing the metadata-refresh-interval-minutes, would reload everything. This way new data will be loaded

jiaqiluo:
The bug fix is validated with the rancher testing image v2.3.2

Saw the versions in the list is updated to the expected ones after either 
- editing the URL to a different branch
or 
- changing the `v3/settings/metadata-refresh-interval-minutes`

Issue #21296, opened 2019-07-02
Implement refresh action for k8s metadata


alena1108:

jiaqiluo:
The feature is tested by using testing Rancher images and forked `kontainer-driver-metadata` Github repo.

Steps:
- run Rancher-testing:v2.3.3
- in Rancher, check the k8s list: v1.12.4-rancher1-1 (default), v1.13.1-rancher1-1(experimental)
- in kontainer-driver-metadata, add a new k8s minor version v1.12.10-rancher1-1
- in kontainer-driver-metadata, add a new k8s major version v1.16.0-rancher1-1
- generate new metadata and push to Github
- in Rancher, refresh `v3/kontainerdrivers`

Result:
- the k8s list reflects the changes, meaning it shows newly-added v1.12.10-rancher1-1 as default and v1.16.0-rancher1-1 as experimental
- also checked the v3/settings and see proper values in the follow filed
```
- ui-k8s-supported-versions-range 
- ui-k8s-default-version-range 
- k8s-version-to-images
- k8s-versions-current
- k8s-version
- k8s-version-to-service-options
```

PR #21502, opened 2019-07-12
updates for driverMetadata

kinarashah:
add roles, refresh action and calculate url based on latest commit

rancher/rancher#21296
https://github.com/rancher/rancher/issues/21310
https://github.com/rancher/types/pull/905

### DIFF
--- a/app/role_data.go
+++ b/app/role_data.go
@@ -65,7 +65,10 @@ func addRoles(management *config.ManagementContext) (string, error) {
 		addRule().apiGroups("management.cattle.io").resources("multiclusterapps", "globaldnses", "globaldnsproviders", "clustertemplates", "clustertemplaterevisions").verbs("create").
 		addRule().apiGroups("project.cattle.io").resources("sourcecodecredentials").verbs("*").
 		addRule().apiGroups("project.cattle.io").resources("sourcecoderepositories").verbs("*").
-		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("get", "list", "watch")
+		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("rkek8ssystemimages").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("rkek8sserviceoptions").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("rkeaddons").verbs("get", "list", "watch")
 
 	rb.addRole("User Base", "user-base").addRule().apiGroups("management.cattle.io").resources("preferences").verbs("*").
 		addRule().apiGroups("management.cattle.io").resources("settings").verbs("get", "list", "watch").
--- a/pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
@@ -6,7 +6,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
+
+	"github.com/rancher/norman/types/convert"
+	"github.com/rancher/rancher/pkg/catalog/git"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -45,7 +49,6 @@ type Data struct {
 
 type TickerData struct {
 	cancelFunc context.CancelFunc
-	url        string
 	interval   time.Duration
 }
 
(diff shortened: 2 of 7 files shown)


app/role_data.go
```diff
@@ -65,9 +65,11 @@ func addRoles(management *config.ManagementContext) (string, error) {
 		addRule().apiGroups("management.cattle.io").resources("multiclusterapps", "globaldnses", "globaldnsproviders", "clustertemplates", "clustertemplaterevisions").verbs("create").
 		addRule().apiGroups("project.cattle.io").resources("sourcecodecredentials").verbs("*").
 		addRule().apiGroups("project.cattle.io").resources("sourcecoderepositories").verbs("*").
-		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("get", "list", "watch")
-
-	rb.addRole("User Base", "user-base").addRule().apiGroups("management.cattle.io").resources("preferences").verbs("*").
+		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("rkek8ssystemimages").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("rkek8sserviceoptions").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("rkeaddons").verbs("get", "list", "watch").
```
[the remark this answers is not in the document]
kinarashah: Yeah, that's the intended behavior 
will remove `.`

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -130,18 +131,49 @@ func (m *MetadataController) refresh(url string, init bool) {
 		return
 	}
 
+	logrus.Infof("driverMetadata: refresh data")
```
prachidamle: lMaybe change the log statement to Debug?

prachidamle:
As long as we do not update the existing CRDs on metadata refresh, we should be hopefully not have errors introduced due to potential parallel refresh calls. Right @kinarashah ?

kinarashah:
CRDs would get updated 
- if the same key has different values (in between RCs only, for final server versions - we just add new keys) 
- if the existLabel change, which would be when rancher upgrades and RKE now has the latest information, we update label on CRD 

I think in either case, multiple refresh would just result in k8s conflict errors, but data wouldn't get incorrect values

prachidamle:
LGTM

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -67,8 +70,11 @@ func Register(ctx context.Context, management *config.ManagementContext) {
 	}
 
 	// load values on startup and start ticker
-	m.refresh(settings.RkeMetadataURL.Get(), true)
-	m.initTicker(ctx)
+	url, interval, err := GetSettingValues()
```
alena1108: shall we check for error before executing line 74?

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -130,18 +133,49 @@ func (m *MetadataController) refresh(url string, init bool) {
 		return
 	}
 
+	logrus.Debug("driverMetadata: refresh data")
 	if err := m.createOrUpdateMetadata(data); err != nil {
 		logrus.Errorf("driverMetadata %v", err)
```
alena1108: I feel like we have to enhance this error message

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -98,27 +103,25 @@ func (m *MetadataController) sync(key string, test *v3.Setting) (runtime.Object,
...
+		url, _, err := GetSettingValues()
+		if err != nil {
+			logrus.Errorf("driverMetadata: error getting settings %v", err)
+		}
+		m.Refresh(url, false)
 	}
 }
 
-func (m *MetadataController) refresh(url string, init bool) {
+func (m *MetadataController) Refresh(url string, init bool) {
```
alena1108: I think we should add an error as a return value. And then let the caller decide on what to do with it.

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -130,18 +133,49 @@ func (m *MetadataController) refresh(url string, init bool) {
...
 		return "", 0, err
 	}
-	return settings.RkeMetadataURL.Get(), checkInterval, nil
+	urlData := map[string]interface{}{}
+	if err := json.Unmarshal([]byte(settings.RkeMetadataURL.Get()), &urlData); err != nil {
+		return "", 0, fmt.Errorf("unmarshal err %v", err)
+	}
+	url, ok := urlData["url"]
+	if !ok {
+		return "", 0, fmt.Errorf("url not present %s", settings.RkeMetadataURL.Get())
```
alena1108: do you think we can enhance the error message?

alena1108:
LGTM, can be merged once vendor.conf conflict is resolved

Issue #20819, opened 2019-06-12
etcd rolling backup naming scheme


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
feature

The new rolling backup naming scheme. When using rolling snapshots, the snapshot names are no longer meaningful. When passing in --once, I get a snapshot named '2019-06-07T23:05:09Z_etcd', which I can later use to parse and roll back to. With rolling snapshots, I get names like c-kk4hq-rl-2gd4z, ${cluster_id}-rl-${random-hash}. When we were using v0.1.17 rke, the rolling snapshots had the same naming scheme as the one-time snapshots. It would be nice to have the option to use the old naming scheme instead.

phomein:
Would like to add that this is with a k8s version of v1.13.5-rancher1-3 and a rancher agent version of v2.2.4 for the k8s cluster, with the rancher cluster created using rke v0.2.4

alena1108:
Can be tested with Rancher master build

moelsayed:
Prior to this change, we had: 
- backup name in kubernetes: `c-6jc29-ms-w547c`
- backup filename on disk/s3, which is a zipped file: `c-6jc29-ms-w547c.zip`

As of this change, the backup name in kubernetes will remain the same. However, the backup filename on disk/s3 will contain and addition timestamp similar to the one used by rke cli, eg: `c-6jc29-ms-w547c_2019-07-17T14:59:27Z.zip`.

sowmyav27:
### Verified on latest rancher:master-head
**Local backups on the UI are seen as below. `-ml-` - manual backups and `-rl-` is for recurring snapshots**
<img width="625" alt="Screen Shot 2019-07-23 at 10 36 26 AM" src="https://user-images.githubusercontent.com/26032343/61734059-01b52a00-ad36-11e9-9ab8-29bd1ad2ff24.png">

**On the node:**
<img width="614" alt="Screen Shot 2019-07-23 at 9 40 59 AM" src="https://user-images.githubusercontent.com/26032343/61734132-2dd0ab00-ad36-11e9-95b9-299df8aa8de4.png">

On doing a restore from snapshot(local), the file on the node, is unzipped, but is not deleted after. And when the restore happens from the same snapshot a second time, the unzipped file is overwritten. This is not the case with S3 snapshots, where there is no unzipped files(only zipped files exist in the S3 bucket)

Tracking this in another issue - https://github.com/rancher/rancher/issues/21710

[...]

PR #21505, opened 2019-07-12
Add timestamp to snapshot files

moelsayed:
https://github.com/rancher/rancher/issues/20819


pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -887,7 +889,14 @@ func (p *Provisioner) restoreClusterBackup(cluster *v3.Cluster, spec v3.ClusterS
 		return "", "", "", fmt.Errorf("snapshot [%s] is not a backup of cluster [%s]", backup.Name, cluster.Name)
 	}
 
-	api, token, cert, err = p.driverRestore(cluster, spec)
+	if filename, err := GetBackupFilenameFromURL(backup.Spec.Filename); err == nil { // s3 file
+		// need to remove extention
+		snapshot = strings.TrimSuffix(filename, path.Ext(filename))
+	} else if len(backup.Spec.Filename) != 0 { // not s3 url
```
superseb: Can you add a comment when this situation is hit? AFAIK only backups that have filename set is s3 with an URL, for local its not set. Is this to catch filenames with no URL, if so, what case is that hit?

moelsayed: There are 3 cases here:
1 - backup is s3
2 - backup is local (contains timestamp in the filename)
3 - backup is local (created with a previous version of rancher where filename is not used)

This condition handles case number 2.

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -225,8 +224,15 @@ func (c *Controller) etcdSaveWithBackoff(b *v3.EtcdBackup) (runtime.Object, erro
 			return b, err
 		}
 		var inErr error
+		snapshotName := b.Name
+		if filename, err := clusterprovisioner.GetBackupFilenameFromURL(b.Spec.Filename); err == nil { // s3 file
```
kinarashah: maybe we can extract this common logic of getting snapshotName?

Issue #21254, opened 2019-07-01
Use Grafana's Filesystem Usage metrics on Rancher's Disk Utilization dashboard for cluster metrics


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Enhancement

**Steps to reproduce (least amount of steps as possible):**
Create a cluster and enable monitoring metrics. If your node has NFS shares mounted, the discrepancies will be more apparent.

View and compare the following:
1. Rancher Dashboard - Cluster Metrics - Disk Utilization
2. Grafana Dashboard - Filesystem Usage

**Result:**
For a given node, the disk utilization or filesystem usage % are different numbers. Appears that the Rancher Dashboard will use all mounts that are not rootfs devices, including network mounts. Grafana will only use devices that start with /dev/.

**Other details that may be helpful:**
According to metrics team, the following expressions are used.
Rancher:
```
(sum(node_filesystem_size_bytes{device!="rootfs"}) by (instance) - sum(node_filesystem_free_bytes{device!="rootfs"}) by (instance)) / sum(node_filesystem_size_bytes{device!="rootfs"})  by (instance)
```
Grafana:
```
sum (container_fs_usage_bytes{device=~"^/dev/.*$",id="/"}) by (node) / sum (container_fs_limit_bytes{device=~"^/dev/.*$",id="/"}) by (node) * 100
```
Enhancement request is to use the expression (or something similar) in Rancher that is used in Grafana. What Grafana displays is likely more in line with what an end user would expect.

Appears there's also a similar issue with memory metrics, see https://github.com/rancher/rancher/issues/18827

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.2.4
- Installation option (single install/HA):
NA

jiaqiluo:
The bug is reproduced in Rancher:v2.2.6

Steps:
- run Rancher:v2.2.6 container
- add an RKE single-node cluster
- enable the cluster monitoring
- mount an NFS shared disk to the node
```
> df -H
167.71.106.250:/nfs   84G  1.6G   82G   2% /nfs 
```
- view and compare the following metrics:
```
- Rancher Dashboard - Cluster Metrics - Disk Utilization
- Grafana Dashboard - Filesystem Usage
```

Result:
- in the Rancher dashboard one, the value drops from ~11% to ~5% after mounting the NFS shared disk
[...]

PR #21518, opened 2019-07-15
Keep metrics graph consistancy

orangedeng:
**Problem:**
Cluster level graph `Disk Utilization` is not the same as grafana's
graph.

**Solution:**
Modify our metrics expression.

Related issue:
https://github.com/rancher/rancher/issues/21254


loganhz:
LGTM

Issue #21501, opened 2019-07-12
Cluster Templates - second global admin cannot clone revisions created by another global admin


davidnuzik:
Version: master-head (7/12/19) (`a20255d2`)

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- As admin, create a cluster template.
- Create a second user with global admin access in the UI.
- Login as this second admin user, try to clone a revision of the template the first admin made. Try to save.

**Result:**
The second global admin for some reason is not allowed to save a revision the first admin created. They get a message like this:
![image](https://user-images.githubusercontent.com/45179589/61151554-11618280-a49b-11e9-8384-2285d5325988.png)
Note, the second global admin can delete revisions and whole templates created by the first admin. They just can't make clones of revisions that first admin created.

sangeethah:
@davidnuzik Can we also test this case ?
As regular user , user1 create T1.
Share this template with another user , user2 as "owner" / "member".
As user2 , should be a able to create a template revision under T1 successfully.

davidnuzik:
Version: master-head (7/16/19)

**Validate originally described issue:**
Confirmed the second admin can now clone revisions of templaterevisions the first admin created. The second admin can do this both either via "New Revision from Default" or "Clone Revision" action.

**Validate Sangeetha's test case (from above comment):**
- Regular user, "user1" create a template "T1"
- Share the template with another user, "user2" as "owner" and also test with them as "member" too.
- Log in as "user2" and this user should be able to create a template revision under T1 without any issues (BOTH as an Owner and as a Member).

This worked without any issues.

PR #21531, opened 2019-07-15
Change logic to check permissions to create clustertemplate revision

mrajashree:
Instead of checking for rolebindings, we can use CanDo to check user's
access to the clusterTemplate with "update" verb. If that's not allowed
then creation of revision for that clusterTemplate shouldn't be allowed either.
This also does group membership based access check.

https://github.com/rancher/rancher/issues/21501

### DIFF
--- a/tests/integration/suite/test_clustertemplate.py
+++ b/tests/integration/suite/test_clustertemplate.py
@@ -332,8 +332,8 @@ def test_revision_creation_permission(admin_mc, remove_resource,
         create_cluster_template_revision(user_readonly.client, templateId)
     except ApiError as e:
         assert e.error.status == 403
-        assert "read-only member of clustertemplate cannot " \
-               "create revisions for it" in e.error.message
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_clustertemplate.py
```diff
@@ -332,8 +332,8 @@ def test_revision_creation_permission(admin_mc, remove_resource,
         create_cluster_template_revision(user_readonly.client, templateId)
     except ApiError as e:
         assert e.error.status == 403
-        assert "read-only member of clustertemplate cannot " \
-               "create revisions for it" in e.error.message
+        assert "user does not have permission to update clusterTemplate" \
+               in e.error.message
```
prachidamle: are we able to test group-based permissions via a test?

mrajashree: Not with local auth, will need to configure external auth provider in the tests to test group-based access

prachidamle:
LGTM

Issue #21084, opened 2019-06-24
Derive Rancher supported Kubernetes Version for a clustertemplate


prachidamle:
While creating a ClusterTemplate, template creator has a choice to

---- option a) Template creator can set exact K8sversion as value to this field. This will lock the k8sversion for this template revision. Values could be (1.14.3, 1.15.x)

---- option b) Template creator can turn this into a Question on the templaterevision (with a k8s range set using min/max). 

In both cases, we need backend processing to derive the Rancher supported Kubernetes version while creating a cluster from this template.

prachidamle:
For:
 option b) Template creator can turn this into a Question on the templaterevision (with a k8s range set using min/max).

We will be supporting setting the KubernetesVersion into a Question. But setting a min/max range will not be supported right now.

sangeethah:
Tested with rancher server version built from master - `f9aa0fe1`
We are now allowed to create clusterTemplateRevisions with versions in the following format - 
`{"rancherKubernetesEngineConfig": {"kubernetesVersion": "1.14.x"}}`
`{"rancherKubernetesEngineConfig": {"kubernetesVersion": "1.14.3"}}`

In both the cases we have cluster getting created with the `kubernetesVersion` set to correct image -`v1.14.3-rancher1-1`.

Related issue when creating clusters with invalid kubernetesVersion format - https://github.com/rancher/rancher/issues/21328
https://github.com/rancher/rancher/issues/21328

When having `kubernetesVersion` set to format like `1.15.x` , there could be a potential issue of cluster's k8s version getting upgraded when any update it attempted on the cluster.

Consider this scenario where a cluster was created with template revision that has following 
```
configuration:
"clusterConfig": {
"dockerRootDir": "/var/lib/docker",
},
"questions": [ 2 items
{
"default": "1.15.x",
"max": 0,
"maxLength": 0,
"min": 0,
"minLength": 0,
"required": true,
"type": "string",
"variable": "rancherKubernetesEngineConfig.kubernetesVersion"
},
{
"default": "false",
"max": 0,
"maxLength": 0,
"min": 0,
"minLength": 0,
"required": false,
"type": "boolean",
"variable": "rancherKubernetesEngineConfig.services.etcd.snapshot"
}
],
```
Create a cluster with this template revision with no answers.
Cluster gets created with snapshot set to "false" and k8s version set to `v1.15.0-rancher1-1` as expected.

[...]

PR #21539, opened 2019-07-16
DeriveKubernetesVersion from the generic major.minor.patch

prachidamle:
https://github.com/rancher/rancher/issues/21084

New changes to this ticket:
---- If a template specifies Kubernetes version of the format 1.14.x, a Question for the "rancherKubernetesEngineConfig.kubernetesVersion" should also be set - API will error out if a question is not added. 
---- When a cluster is created from a clusterTemplate that specifies the k8sversion of a format 1.14.x
backend will translate this to rancher supported kubernetes version on the create. If user does not provide an answer to the k8sversion Template Question, Backend will use the default value for translation.
---- On editing such a cluster, If the answer to the k8sversion Template Question is not different, no translation will be done again. This prevents the auto-upgrade to newer version as was seen on earlier tests.
---- If user wants to upgrade the cluster to newer k8sversion, he/she should set the new answer(with specific k8sversion or higher patch as 1.15.x) on cluster edit or change the templateRevision having a newer k8s support.

### DIFF
--- a/pkg/api/store/cluster/cluster_store.go
+++ b/pkg/api/store/cluster/cluster_store.go
@@ -174,7 +174,7 @@ func (r *Store) Create(apiContext *types.APIContext, schema *types.Schema, data 
 			return nil, err
 		}
 		clusterConfigSchema := apiContext.Schemas.Schema(&managementschema.Version, managementv3.ClusterSpecBaseType)
-		data, err = loadDataFromTemplate(clusterTemplateRevision, clusterTemplate, data, clusterConfigSchema)
+		data, err = loadDataFromTemplate(clusterTemplateRevision, clusterTemplate, data, clusterConfigSchema, nil)
 		if err != nil {
 			return nil, err
 		}
@@ -227,7 +227,7 @@ func transposeNameFields(data map[string]interface{}, clusterConfigSchema *types
 	return data
 }
 
-func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, clusterTemplate *v3.ClusterTemplate, data map[string]interface{}, clusterConfigSchema *types.Schema) (map[string]interface{}, error) {
+func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, clusterTemplate *v3.ClusterTemplate, data map[string]interface{}, clusterConfigSchema *types.Schema, existingCluster map[string]interface{}) (map[string]interface{}, error) {
 	dataFromTemplate, err := convert.EncodeToMap(clusterTemplateRevision.Spec.ClusterConfig)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/api/store/clustertemplate/store.go
```diff
@@ -295,5 +295,26 @@ func (p *Store) checkKubernetesVersionFormat(apiContext *types.APIContext, data
...
+		//ensure a question is added for "rancherKubernetesEngineConfig.kubernetesVersion"
+		templateQuestions, ok := data[managementv3.ClusterTemplateRevisionFieldQuestions]
+		if !ok {
+			return httperror.NewAPIError(httperror.MissingRequired, fmt.Sprintf("ClusterTemplateRevision must have a Question set for %v", clustertemplate.RKEConfigK8sVersion))
+		}
+		templateQuestionsSlice := convert.ToMapSlice(templateQuestions)
+		var foundQ bool
+		for _, question := range templateQuestionsSlice {
+			if question["variable"] == clustertemplate.RKEConfigK8sVersion {
+				foundQ = true
```
alena1108: should we break once found?

pkg/api/store/cluster/cluster_store.go
```diff
@@ -236,18 +236,25 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
 	dataFromTemplate[managementv3.ClusterSpecFieldClusterTemplateID] = ref.Ref(clusterTemplate)
 	dataFromTemplate[managementv3.ClusterSpecFieldClusterTemplateRevisionID] = convert.ToString(data[managementv3.ClusterSpecFieldClusterTemplateRevisionID])
-	dataFromTemplate[managementv3.ClusterSpecFieldClusterTemplateAnswers] = data[managementv3.ClusterSpecFieldClusterTemplateAnswers]
 
 	dataFromTemplate = transposeNameFields(dataFromTemplate, clusterConfigSchema)
 
 	//Add in any answers to the clusterTemplateRevision's Questions[]
-	answers := convert.ToMapInterface(data[managementv3.ClusterSpecFieldClusterTemplateAnswers])
-	allAnswers := convert.ToMapInterface(answers["values"])
+	allAnswers := convert.ToMapInterface(convert.ToMapInterface(data[managementv3.ClusterSpecFieldClusterTemplateAnswers])["values"])
```
kinarashah: q: why do we convert to map twice?

prachidamle: Because  it is a nested map structure. Example:
```
"answers" = {
           "values": {
	            "dockerRootDir": "/var/lib/docker123",
	            "rancherKubernetesEngineConfig.ignoreDockerVersion":
	            "false"
            }
          }
```

alena1108:
LGTM

Issue #21556, opened 2019-07-16
K8s metadata - behavior of  MaxRancherVersion and MinRancherVersion


jiaqiluo:
Bug

set the metadata such that for k8s version `v1.13.7-rancher1-1`:
```
		"v1.13": {
			MinRancherVersion: "2.3.3",
			MaxRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},
```

Expected Results:  

We expected to see the k8s `v1.13.7-rancher1-1` shows only in Rancher `rancher:v2.3.3`

Results:  

in fact `v1.13.7-rancher1-1` shows in other Rancher versions:
- rancher:v2.3.4
- rancher:v2.3.2
- rancher:v2.4.0

Question:

How to configure to make `v1.13.7-rancher1-1` shows only in Rancher `rancher:v2.3.3`? 
Is it possible? 

More Info:
- tested with the following config, but still not work as expected 
```
"v1.13": {
			MaxRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},
"v1.13.7-rancher1-1": {
			MinRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},
```

New finding:
- it turns out it will not work if setting the min and max the same value

jiaqiluo:
The bug fix is validated with the rancher testing image v2.3.2

set the metadata like this:
```
		"v1.13": {
			MinRancherVersion: "2.3.3",
			MaxRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},
```
The result is 
- k8s v1.13.x shows only in Rancher v2.3.3

Issue #21553, opened 2019-07-16
K8s metadata - unexpected experimental k8s version in the dropdown list  


jiaqiluo:
Bug

Set the testing metadata, specifically the pramether `MaxRancherVersion` and `MinRancherVersion`, as the following:
<img width="848" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/61311540-678c3980-a7ab-11e9-8165-ab158b7f276f.png">

```
		"v1.13": {
			MinRancherVersion: "2.3.3",
			MaxRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},

```
Result:

run Rancher testing image `rancher:v2.3.2`, seeing that `v1.13.7-rancher1-1`, `v1.14.3-rancher1-1` and `v1.15.0-rancher1-1` are shown as experimental versions, which is not expected. 
<img width="786" alt="Screen Shot 2019-07-16 at 9 27 01 AM" src="https://user-images.githubusercontent.com/6218999/61311835-0749c780-a7ac-11e9-9036-6bacb5f5063b.png">

jiaqiluo:
The bug fix is validated with the latest Rancher testing image v2.3.2

Steps:
- set the metadata that k8s v1.11 is the default, v1.12 is the experimental
- set the metadata that k8s v1.13, v1.14, and v1.15 do not show in Rancher v2.3.2
```
		"v1.11": {
			MaxRancherVersion: "2.3.2",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.12": {
			MaxRKEVersion:     "0.2.2",
		},
		"v1.12.3-rancher1-1": {
			MinRancherVersion: "2.3.2",
			MaxRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.12.4-rancher1-1": {
			MinRancherVersion: "2.3.3",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.13": {
			MinRancherVersion: "2.3.3",
			MaxRancherVersion: "2.3.4",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.13.1-rancher1-1": {
			MinRancherVersion: "2.3.3",
			MaxRancherVersion: "2.3.4",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.14": {
			MinRancherVersion: "2.3.4",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.15": {
[...]

PR #21593, opened 2019-07-17
driverMetadata: respect both minor and major k8s versions

kinarashah:
https://github.com/rancher/rancher/issues/21556
https://github.com/rancher/rancher/issues/21553


prachidamle:
LGTM

Issue #21300, opened 2019-07-03
Turn Linode on as default driver


deniseschannon:
Things to check:
* airgapped setups - we will not support linode working for airgapped setups
  - what shows up in the UI when airgapped?
  - will the icon appear? should we move the icon directly into rancher (probably will solve https://github.com/rancher/rancher/issues/20524)?
  - what happens in the logs? will it continually try to get the data?

kinarashah:
@alena1108 @deniseschannon 

For airgapped setups - 
- UI shows icon on cluster page only when driver is active, so nothing would show up because driver would be stuck in `downloading` state 

- Icon would show up only if they've set up mirror, currently it is in linode's repo. I'm not sure of the root cause of the above mentioned issue, but we should move it to rancher since we're officially turning it ON by default? 

- yes, it'll continuously try to fetch the url to download binary. they can just deactivate the driver in that case, is that good enough workaround if we document it?

alena1108:
As a part of the fix, we have to make sure that it doesn't break airgap and dev local setup. In airgap setup @vincent99 suggested including the driver to the rancher image, similar to how we do Packet.

kinarashah:
@alena1108 

The UI code for Linode doesn't follow the same logic as other builtin drivers. It needs to go through the `/meta/proxy` link, this has two benefits: 
- The whitelisted domains mentioned for the driver get respected 
- We can make use of cloud credentials support instead of actual passing of secret data from ui to linode api. 
https://github.com/rancher/ui/blob/master/lib/global-admin/addon/components/cru-cloud-credential/component.js#L188

alena1108:
@kinarashah missing UI functionality will be tracked in https://github.com/rancher/rancher/issues/21838

alena1108:
The scope of this issue is limited to:

* Making Linode driver enabled in UI by default.
* In airgapped setups, make data to be pulled from cache, with no error is presented in the UI/Rancher server logs

Those should be the steps to validate for qa in order to close the issue. For all other issues found along the way, new bugs have to be opened.

jiaqiluo:
> The scope of this issue is limited to:
> - Making Linode driver enabled in UI by default.
> - In airgapped setups, make data to be pulled from cache, with no error is presented in the UI/Rancher server logs

The first part is validated in Rancher:master-head 51769836 
[...]

PR #21602, opened 2019-07-17
mark linode active by default

kinarashah:
https://github.com/rancher/rancher/issues/21300


pkg/api/customization/node/node_driver.go
```diff
@@ -43,6 +43,17 @@ func (h *DriverHandlers) ActionHandler(actionName string, action *types.Action,
 	case "deactivate":
 		m.Spec.Active = false
 		v3.NodeDriverConditionInactive.Unknown(m)
+		var newConditions []v3.Condition
```
alena1108: @kinarashah can we put a comment in the code explaining why we remove these conditions on deactivate

alena1108:
LGTM

Issue #17471, opened 2019-01-17
Rancher CLI should expose Helm timeout flag


nickolaskraus-wf:
**What kind of request is this (question/bug/enhancement/feature request):**

🐞Bug

**Steps to reproduce (least amount of steps as possible):**

When attempting to deploy a Helm chart via Rancher that requires the `--wait` and `--timeout` flag, the deploy fails.

The Helm chart in question is [stable/sentry](https://github.com/helm/charts/tree/master/stable/sentry).

In order for this chart to deploy successfully, the following command is needed:

```bash
helm install -f values.yaml --name sentry --wait --timeout 600 .
```

Since Rancher does not allow for `--wait`, Rancher returns an error even thought the deployment is `active`.

**Result:**

```
$ rancher app install -n workiva -a helm/sentry/answers/inf-dev.yaml ./helm/sentry sentry
FATA[0149] timed out waiting for app notes, the app could still be installing. Run 'rancher apps' to verify
```

**Other details that may be helpful:**

From the Helm documentation:
```
helm install --help

This command installs a chart archive.

Usage:
  helm install [CHART] [flags]

Flags:
--timeout int              time in seconds to wait for any individual Kubernetes
                           operation (like Jobs for hooks) (default 300)

--wait                     if set, will wait until all Pods, PVCs, Services, and minimum
                           number of Pods of a Deployment are in a ready state before
                           marking the release as successful. It will wait for as long as
                           --timeout
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.1.5
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory): cloud, 16 cores/64GiB RAM
- Kubernetes version (use `kubectl version`):

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.2", GitCommit:"cff46ab41ff0bb44d8584413b598ad8360ec1def", GitTreeState:"clean", BuildDate:"2019-01-13T23:16:58Z", GoVersion:"go1.11.4", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #21604, opened 2019-07-18
Wait timeout fix

rmweir:
Problem:
 Users are unable to successfully install apps that take longer than the helm default of 300 seconds.

Solution:
Make timeout and wait configurable for app and pass values to helm.

Issue:
https://github.com/rancher/rancher/issues/17471

intended merge order:
types -> rancher -> cli

### DIFF
--- a/pkg/controllers/user/helm/common/common.go
+++ b/pkg/controllers/user/helm/common/common.go
@@ -119,12 +119,14 @@ func GenerateRandomPort() string {
 func InstallCharts(tempDirs *HelmPath, port string, obj *v3.App) error {
 	logrus.Debugf("InstallCharts - helm path info %+v\n", tempDirs)
 	extraArgs := GetExtraArgs(obj)
+	timeoutArgs := getTimeoutArgs(obj)
 	setValues, err := GenerateAnswerSetValues(obj, tempDirs, extraArgs)
 	if err != nil {
 		return err
 	}
 	commands := make([]string, 0)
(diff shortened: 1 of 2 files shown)


luthermonson:
lgtm

pkg/controllers/user/helm/common/common.go
```diff
@@ -160,6 +162,20 @@ func InstallCharts(tempDirs *HelmPath, port string, obj *v3.App) error {
...
 
+// getTimeoutArgs generates the appropriate arguments for helm depending on the app's timeout and wait fields
+func getTimeoutArgs(app *v3.App) []string {
+	var timeoutArgs []string
+
+	if app.Spec.Wait {
+		timeoutArgs = append(timeoutArgs, "--wait")
+	}
+	// the default timeout in helm is 300 seconds
+	if timeout := app.Spec.Timeout; timeout != 300 {
```
dramich: I think we can just append this regardless what the value is. This would be useful for debugging the commands to know what the expected timeout is.

tests/integration/suite/test_app.py
```diff
@@ -222,6 +222,73 @@ def test_app_namespace_annotation(admin_pc, admin_mc):
...
+                return False
+            return test_app
+
+        return wait_for(transition_error, timeout=15, fail_handler=lambda:
+                        "expected transitioning to fail")
+
+    app1 = wait_for_transition_error(app1)
+    assert "timed out waiting for the condition" in app1.transitioningMessage
+
+    app2 = client.create_app(
```
dramich: Drop from here down, this is testing creating an app which we already do.

dramich:
lgtm

PR #21638, opened 2019-07-19
Update to get desired replicas from spec

aiwantaozi:
Problem:
1. replicate number in state isn't the desired replicate number
2. convert threshold percentage to int32 will lose accurate in calculation

Solution:
get replicate number from spec and use float

Issue:
https://github.com/rancher/rancher/issues/21639

### DIFF
--- a/pkg/controllers/user/alert/watcher/workload.go
+++ b/pkg/controllers/user/alert/watcher/workload.go
@@ -2,7 +2,9 @@ package watcher
 
 import (
 	"context"
+	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/rancher/norman/controller"
@@ -10,8 +12,12 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/user/alert/manager"
 	"github.com/rancher/rancher/pkg/controllers/user/workload"
 	"github.com/rancher/rancher/pkg/ticker"
+	"github.com/rancher/types/apis/apps/v1beta2"
+	v1 "github.com/rancher/types/apis/core/v1"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config"
+
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/alert/watcher/workload.go
```diff
@@ -162,11 +181,63 @@ func (w *WorkloadWatcher) checkWorkloadCondition(wl *workload.Workload, alert *v
...
+		return o.Status.DesiredNumberScheduled, nil
+	case workload.StatefulSetType:
+		o, err := w.statefulsetLister.Get(namespace, name)
+		if err != nil {
+			return desiredReplicas, err
+		}
+		if o.Spec.Replicas != nil {
+			return *o.Spec.Replicas, nil
+		}
+	default:
```
gitlawr: The switch does not handle job/cronjob, they are considered as deployment kind in those cases.

aiwantaozi: This feature does not include job and cronjob, we return direclty when matching job/cronjob at line 150

gitlawr:
LGTM

pkg/controllers/user/alert/watcher/workload.go
```diff
@@ -162,11 +181,63 @@ func (w *WorkloadWatcher) checkWorkloadCondition(wl *workload.Workload, alert *v
...
 		}
 	}
 
 }
+
+func (w *WorkloadWatcher) getDesiredReplicas(workloadName string) (int32, error) {
+	var desiredReplicas int32
+	splitted := strings.Split(workloadName, ":")
+	if len(splitted) != 3 {
+		return desiredReplicas, errors.New("invalid workloadName: " + workloadName)
```
kinarashah: we always pass the correct string in L159, why do we need to check this?

pkg/controllers/user/alert/watcher/workload.go
```diff
@@ -162,11 +181,63 @@ func (w *WorkloadWatcher) checkWorkloadCondition(wl *workload.Workload, alert *v
...
+	}
+	workloadType := strings.ToLower(splitted[0])
+	namespace := splitted[1]
+	name := splitted[2]
+	switch workloadType {
+	case workload.ReplicationControllerType:
```
kinarashah: we already workloadController https://github.com/rancher/rancher/pull/21638/files#diff-2046287b69957361f7d174a68bcc0d66R62, could we use it to GetWorkload?

aiwantaozi: Hi @kinarashah workload struct isn't include desired replicas, just include current replicas, I afraid change workload struct affect other parts, so I get desired replicas here.

kinarashah: Unless I'm missing something, we have both desired and available replicas? 
https://github.com/rancher/rancher/blob/1a7a0b5329cabcff30fda8d5a75108a779846605//pkg/controllers/user/workload/workload_common.go#L262 

https://github.com/rancher/rancher/blob/1a7a0b5329cabcff30fda8d5a75108a779846605//pkg/controllers/user/workload/workload_common.go#L340

aiwantaozi: Hi @kinarashah , only the daemonSet use desired number, statefulSet, deployment, replicaSet use status replicas, it is the most recently observed number of replicas, so I finally give up function in workload_common.

cjellick:
lgtm

Issue #21302, opened 2019-07-03
Support exist condition for metric alert


aiwantaozi:
Now we only support compare condition and the threshold for metric alert
First: user can't input expression with condition and threshold by themselves.
Second: it can't support an expression combine with multiple parts, for example:
`predict_linear(node_filesystem_avail_bytes{app="node-exporter",fstype=~"ext.|xfs"}[6h], 4*60*60) <0`
`and`
`node_filesystem_avail_bytes{app="node-exporter",fstype=~"ext.|xfs"} / node_filesystem_size_bytes{app="node-exporter",fstype=~"ext.|xfs"} < 0.2
and
node_filesystem_readonly_bytes{app="node-exporter",fstype=~"ext.|xfs"} == 0
`

![image](https://user-images.githubusercontent.com/5936271/60562886-09635d80-9d8c-11e9-92bc-cf4b7a75c637.png)

loganhz:
cc @patrick0057

gitlawr:
Do we need a UI change/ticket for this?

aiwantaozi:
https://github.com/rancher/rancher/issues/21790 UI issue

jiaqiluo:
the feature enhancement is validated in Rancher: master-head 7277535d

**Test 1**: expression with conditions and thresholds  and multiple parts + set the condition to `Not Null`

**Expresssion**: `sum (kube_deployment_labels  {deployment = "test"} or kube_deployment_labels  {deployment = "test31"}) >= 2`

**Result**: see the alerts triggered and sent to Slack

![Screen Shot 2019-08-15 at 1 26 12 PM](https://user-images.githubusercontent.com/6218999/63125430-7dd01580-bf62-11e9-81f7-adc85b2f0d42.png)

**Test 2**:  expression **without** conditions and thresholds + set the condition to any except `Not Null`

<img width="1410" alt="Screen Shot 2019-08-15 at 1 48 42 PM" src="https://user-images.githubusercontent.com/6218999/63125826-73fae200-bf63-11e9-82d9-4482379d8d16.png">

**Result**: see the alerts triggered and sent to Slack

![Screen Shot 2019-08-15 at 1 50 12 PM](https://user-images.githubusercontent.com/6218999/63125913-9db40900-bf63-11e9-8f11-a6b5b09f8c16.png)

PR #21641, opened 2019-07-19
Support exist condition for metric alert

aiwantaozi:
Problem:
Now we only support compare condition and the threshold for metric alert

Solution:
Add exist condition for metric alert

Issue:
https://github.com/rancher/rancher/issues/21302

Types:
https://github.com/rancher/types/pull/914

### DIFF
--- a/pkg/controllers/user/alert/deployer/notification_template.go
+++ b/pkg/controllers/user/alert/deployer/notification_template.go
@@ -105,7 +105,11 @@ Project Name: {{ .Labels.project_name}}{{ end }}
 {{- if .Labels.pod_name }}
 Pod Name: {{ .Labels.pod_name}}{{ else if .Labels.pod -}}Pod Name: {{ .Labels.pod}}{{ end }}
 Expression: {{ .Labels.expression}}
+{{- if .Labels.threshold_value }}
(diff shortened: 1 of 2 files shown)


aiwantaozi:
Rebased

pkg/controllers/user/alert/deployer/notification_template.go
```diff
@@ -105,8 +105,12 @@ Project Name: {{ .Labels.project_name}}{{ end }}
 {{- if .Labels.pod_name }}
 Pod Name: {{ .Labels.pod_name}}{{ else if .Labels.pod -}}Pod Name: {{ .Labels.pod}}{{ end }}
 Expression: {{ .Labels.expression}}
+{{- if .Labels.threshold_value }}
+Description: The configured event happened for ({{ .Labels.duration}}): expression matched, datapoint value is {{ .Annotations.current_value}}
+{{- else}}
```
gitlawr: Looks like we print `Threshold Crossed ...` when `threshold_value` is empty. Should we do it the other way?

aiwantaozi:
Grafana has an alert for is null value for the expression result, if we use null would be a negative option like ```not null value```, so I change to opposite option ```has-value```

gitlawr:
LGTM

luthermonson:
LGTM, needs conflict resolved though

cjellick:
needs rebase but lgtm

aiwantaozi:
rebased, thanks

Issue #15788, opened 2018-09-24
Azure Availability Set for VM Managed Disks


ghost:
**Rancher versions:**
rancher/rancher: **v2.0.8**

**Type/provider of hosts:** 
Azure

**Steps to Reproduce:**

Deploy Azure VM using Azure cloud provider

Size: **Standard_D2s_V3**
Storage Type: **Premium LRS**
Availability Set: **massimo-avs**

Fill out the following in the azure cloud provider config

aadClientId*
aadClientSecret*
subscriptionId*
tenantId*

**Results:**

Rancher deploys Azure VM which does not use managed disks, even though I specified Premium LRS as the storage type and my own availability set.
```
'massimo1' is not using Managed Disks. Migrate to Managed Disks to get more benefits.
```
The odd thing is that if you create the availability set in Azure Portal as managed before deploying, Rancher will change it to unmanaged on the deployment of the Azure VM.

Am I missing a configuration option that allows me to set the availability set to managed on the node template for Azure? If not then can we implement an option to select Unmanaged (classic) vs Managed (Aligned) availability set?

https://docs.microsoft.com/en-us/rest/api/compute/availabilitysets/createorupdate#request-body
https://docs.microsoft.com/en-us/azure/virtual-machines/windows/tutorial-availability-sets#create-an-availability-set

erikwilson:
Notes: Testing on master initially produced an error like:
``
2019/01/21 16:09:11 [ERROR] NodeController c-jn98d/m-2rfl2 [node-controller] failed with : Error creating machine: Error in driver during machine creation: compute.VirtualMachinesClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="OperationNotAllowed" Message="Addition of a VM with managed disks to non-managed Availability Set or addition of a VM with blob based disks to managed Availability Set is not supported. Please create an Availability Set with 'Aligned' SKU in order to add a VM with managed disks to it."
``

Followed by:
``
[...]

PR #21654, opened 2019-07-19
Update machine v.0.15.0-rancher-9-1

rmweir:
Problem:
We do not support configuring azure VMs to use managed disks. This also block the user from provisioning with certain options we provide that require managed disks.

Solution:
Add managed disks to azure in docker machine

Issue:
https://github.com/rancher/rancher/issues/15788


alena1108:
LGTM

PR #21657, opened 2019-07-19
Remove example feature

rmweir:
Example feature, and its assignment to kontainerdriver was intended for demonstrations and testing. It is being taken out as it does not serve a purpose.


dramich:
lgtm

PR #21660, opened 2019-07-19
rke and kontainer-driver-metadata update

alena1108:


alena1108:
TF is unrelated, it happens in `test_user_manage_templates`, @prachidamle is going to look at it.

PR #21666, opened 2019-07-21
exclude lxcfs and tmpfs devices from node storage check

ahmadalli:
it fixes the issue that was described on https://forums.rancher.com/t/rancher-monitoring-is-falsely-alerting-on-var-lib-lxcfs-running-out-of-disk-space/14513

### DIFF
--- a/pkg/controllers/user/alert/initdata.go
+++ b/pkg/controllers/user/alert/initdata.go
@@ -239,7 +239,7 @@ var entries = []entry{
 					},
 					MetricRule: &v3.MetricRule{
 						Description: "Device on node is running full within the next 24 hours",
-						Expression:  `predict_linear(node_filesystem_free_bytes{mountpoint!~"^/etc/(?:resolv.conf|hosts|hostname)$"}[6h], 3600 * 24) < 0`,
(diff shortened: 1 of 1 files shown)


basroovers:
You can't use `or` in a promql query.
I think the final query should be:
```predict_linear(node_filesystem_files_free{mountpoint!~"^/etc/(?:resolv.conf|hosts|hostname)$", device!="lxcfs"}[6h], 3600 * 24)```

ahmadalli:
@basroovers I've fixed it on the second commit :)

ahmadalli:
@basroovers  `predict_linear(node_filesystem_files_free{mountpoint="/"}[6h], 3600 * 24)` what about the other devices (e.g. mounted on `/mnt/`)?

basroovers:
@ahmadalli I missed the 2nd commit there, much better. I think we should also filter out /boot/efi.

adampl:
Hi, could you merge it? This issue is really annoying.

ahmadalli:
I've changed the check to use `fstype` instead of mount point same of https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/rules/node.rules.yaml#L151

PR #21677, opened 2019-07-22
Revert "mark linode active by default"

alena1108:
This reverts commit 4dc6f7c9fcb92c65fa9acf758ed9bc49e18096b0.


alena1108:
Reverted the fix only temporarily till we fix the airgap setup issue.

Issue #20809, opened 2019-06-12
Failure to create/edit an Ingress object, with a name containing a period '.', via Rancher: `"Name" contains an invalid character: .`


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy the following ingress resource into a Rancher launched cluster via kubectl:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test.example.com
spec:
  rules:
  - host: test.example.com
    http:
      paths:
      - path: /testpath
        backend:
          serviceName: test
          servicePort: 80
```
- Attempt to edit the resource via the Rancher UI, to change the path for example, click 'Save'

**Result:**

- Observe error of the format `"Name" contains an invalid character: .`

**Environment information**
Rancher v2.2.4

davidnuzik:
I was able to replicate this on v2.2.4 and master (6/11/19).
You can also import via YAML without the UI/API complaining (until you edit and try to save changes)

I think the problem is that the schema has name field of type "dnsLabel" which if I recall correctly does not like periods in the string. For some reason only when we edit and try to save do we seem to be enforcing the schema.
There must be a reason but I was expecting when we at least import YAML we enforce schema (we can't do anything about kubectl, but we can probably change the type to support this).

bmdepesa:
rancher/rancher:master-head 7e82f85f

* Import yaml:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test.example.com
spec:
  rules:
  - host: test.example.com
    http:
      paths:
      - path: /testpath
        backend:
          serviceName: test
          servicePort: 80
```
* Ingress is created
* Edit the ingress
* Save is successful

alfonsodg:
Hi,

I am getting the same error with the latest rancher and just using the wordpress sample (my hostname: xip.io-wordpress

Is there any fix for this issue?

bmdepesa:
@alfonsodg I am not able to reproduce this in v2.2.8, can you provide more details?

alfonsodg:
Well... I just create 4 droplets in Digital Ocean, in droplet i have installed rancher and after that create the cluster with the other 3 nodes, after that i just simply install the wordpress application, when i try to change the hostname shows that message

`[rancher@prixtips-01 ~]$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                                      NAMES
[...]

PR #21681, opened 2019-07-22
Relaxed ingress name type

alena1108:
https://github.com/rancher/rancher/issues/20809

https://github.com/rancher/types/pull/915


alena1108:
The fix is in types only; this pr updates the types + adds the test.

PR #21684, opened 2019-07-22
Redact password in error message

mrajashree:


mrajashree:
Not able to add an integration test for this, as to repro this changing the git password followed by a refresh to catalog is needed. Also can't add unit test for this method.

pkg/catalog/helm/setup.go
```diff
@@ -193,6 +193,10 @@ func (h *Helm) updateGit(fetchLatest bool) (string, error) {
 		}
 		if changed {
 			if err = git.Update(h.LocalPath, commit); err != nil {
+				if strings.Contains(err.Error(), h.password) {
+					errStr := strings.Replace(err.Error(), h.password, "****", 1)
```
dramich: I think we should follow the convention from go [here](https://golang.org/src/net/http/client.go?s=30045:30300#L928) so more like `strings.Replace(err.Error(), h.password+"@", "***@", 1)`

dramich:
lgtm

Issue #21306, opened 2019-07-03
Bug: vSphere node provisioning results in  x509 error (Rancher 2.1.10)


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Create a new node in a vSphere cluster and download RancherOS vmware iso from URL with private CA.

**Result:**
You get an x509 error.

**Other details that may be helpful:**
This was working in 2.1.8, tls-ca has been checked and confirmed correct for the URL in question.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.1.10
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Provider vSphere
- Machine type (cloud/VM/metal) and specifications (CPU/memory):VPS

cjellick:
Can you provide relevant logs from the rancher server? Since it's a private CA, how are you ensuring it is known/trusted?

cc @dramich

contain-me-not:
we confirmed the trust via `openssl s_client` from the rancher pod using `cacerts.pem`.

contain-me-not:
the specific error is `Error creating machine: Error in driver during machine creation: Get https://artifactory.mydomain.com/rancheros/rancheros-vmware-v1.4.3.iso: x509: certificate signed by unknown authority; Timeout waiting for ssh key`

rancher log
```
rancher-54986d986-mjrv9 rancher 2019/07/03 16:24:29 [INFO] stdout: Creating CA: /management-state/node/nodes/kube-dev-03-w2/certs/ca.pem
rancher-54986d986-mjrv9 rancher 2019/07/03 16:24:29 [INFO] stdout: Creating client certificate: /management-state/node/nodes/kube-dev-03-w2/certs/cert.pem
rancher-54986d986-mjrv9 rancher 2019/07/03 16:24:30 [INFO] stdout: Running pre-create checks...
rancher-54986d986-mjrv9 rancher 2019/07/03 16:24:33 [INFO] stdout: Creating machine...
rancher-54986d986-mjrv9 rancher 2019/07/03 16:24:33 [INFO] stdout: (kube-dev-03-w2) Image cache directory does not exist, creating it at /management-state/node/nodes/kube-dev-03-w2/cache...
[...]

PR #21719, opened 2019-07-23
Pull in certs to jail

dramich:
Problem:
Additional certs passed to rancher are not pulled into the jail

Solution:
Pull in optional rancher cert files to stop issues with self signed and
internal CAs

Issue: https://github.com/rancher/rancher/issues/21306


vincent99:
image

(LGTM)

Issue #19342, opened 2019-04-02
User objects in kubeconfig are not per cluster friendly


Joeskyyy:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
* Sign in as user
* Create multiple clusters
* Generate kubeconfig for at least two clusters

**Result:**

* User object will have the same `name` field, but different `token` fields. Ideally this would be something like `user-ab11z.$clusterID` if the tokens are now generated per cluster.

**Other details that may be helpful:**
This is a change it seems in 2.2.0 after talking with some folks at rancher in how the token field is generated.

e.g.

Pre 2.2.0
```
kubeconfig-user-ab11z:YAYATOKEN
```

vs

2.2.0
```
kubeconfig-user-ab11z.c-a123b:YAYATOKEN
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.0
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory):AWS, m5.xlarge
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.2", GitCommit:"cff46ab41ff0bb44d8584413b598ad8360ec1def", GitTreeState:"clean", BuildDate:"2019-01-13T23:15:13Z", GoVersion:"go1.11.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.4", GitCommit:"c27b913fddd1a6c480c229191a087698aa92f0b1", GitTreeState:"clean", BuildDate:"2019-02-28T13:30:26Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):

```
Client:
 Version:      17.03.2-ce
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   f5ec1e2
 Built:        Tue Jun 27 02:21:36 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.03.2-ce
 API version:  1.27 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   f5ec1e2
[...]

PR #21728, opened 2019-07-23
Use ClusterName for KubeConfig User's name

vincent99:
Currently the user name is always the same for every config you generate
(`u-xxxxx`), so the kubeconfigs for 2 clusters cannot be merged together.

Instead, Generate kubeconfigs with the User, Context, and Cluster name all
set to the cluster `name`, so that the configs of any two clusters (with
different names...) can be merged together successfully by kubectl.

### DIFF
--- a/pkg/kubeconfig/templates.go
+++ b/pkg/kubeconfig/templates.go
@@ -16,7 +16,7 @@ clusters:
 {{- end}}
 
 users:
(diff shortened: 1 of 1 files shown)


vincent99:
rancher/rancher#19342

ibuildthecloud:
Lgtm

Issue #21731, opened 2019-07-23
Backport: Bug: vSphere node provisioning results in x509 error v2.1


dramich:
https://github.com/rancher/rancher/issues/21306

bmdepesa:
Duplicate of https://github.com/rancher/rancher/issues/21773

PR #21732, opened 2019-07-23
Pull in certs to jail v2.1

dramich:
Problem:
Additional certs passed to rancher are not pulled into the jail

Solution:
Pull in optional rancher cert files to stop issues with self signed and
internal CAs

Issue: https://github.com/rancher/rancher/issues/21731


deniseschannon:
https://github.com/rancher/rancher/issues/21773

Issue #21127, opened 2019-06-25
Alerts - alert group does not re-evaluate the notifier after the notifier is updated


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Run Rancher:master-head behind a proxy
- add a custom cluster which is also behind the proxy
- at the cluster level, add a Slack notifier without providing the Proxy URL
- at the cluster level, add a new alert group and an alert rule, select the Slack notifier as the receiver  
- when the alert is triggered, there is no message sent to Slack --> this is expected
- now edit the Slack notifier to add the Proxy URL
- when the alert is triggered, there is still no message sent to Slack --> this is NOT expected
- go to edit the alert group, change nothing but click saving 
- now messages are sent to Slack
 
It turns out that the alert group does not re-evaluate the receiver when the notifier is updated until we edit the alert group manually.

**Other details that may be helpful:**
- the bug is found while testing notifier behind a proxy, but it is a general bug that should also exist when not using a proxy

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 
master-head 259cfa71
- Installation option (single install/HA):
single install

jiaqiluo:
the bug fix is validated in Rancher:master-head 734bd953

The steps are the same as the issue describes:
- Run Rancher:master-head behind a proxy
- add a custom cluster which is also behind the same proxy
- at the cluster level, add a Slack notifier **without** providing the Proxy URL
- at the cluster level, add a new alert group and an alert rule, select the Slack notifier as the receiver  
- when the alert is triggered, there is no message sent to Slack --> this is expected
- now edit the Slack notifier to add the Proxy URL
- when the alert is triggered, there are messages including the old and new messages, sent to Slack 
- add a new alert group and an alert rule, select the Slack notifier as the receiver  - alters in the new group also works

PR #21737, opened 2019-07-24
Add notifier-config-syncer

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/21127

Problem:
When notifiers are updated, configurations of alert-manager are not
synced.

Solution:
Sync configs on notifier updates

We used to do the sync in [v2.1](https://github.com/rancher/rancher/blob/release/v2.1/pkg/controllers/user/alert/controller.go#L39). This is probably a regression if it is not intentionally dropped.


aiwantaozi:
LGTM

cjellick:
lgtm

Issue #21369, opened 2019-07-08
Global DNS configuration shows credentials in clear text


daanemanz:
**What kind of request is this (question/bug/enhancement/feature request):**
Feature request

**Steps to reproduce (least amount of steps as possible):**
* Configure global DNS provider for Route53 (via UI)
* Query the cluster for the new global DNS provider resource:
```
$ kubectl get globaldnsproviders.management.cattle.io route53 -n cattle-global-data -o yaml
apiVersion: management.cattle.io/v3
kind: GlobalDnsProvider
metadata:
  annotations:
    field.cattle.io/creatorId: <user>
  creationTimestamp: "2019-05-10T20:21:47Z"
  generation: 1
  labels:
    cattle.io/creator: norman
  name: route53
  namespace: cattle-global-data
  resourceVersion: "30597018"
  selfLink: /apis/management.cattle.io/v3/namespaces/cattle-global-data/globaldnsproviders/route53
  uid: <uuid>
spec:
  rootDomain: example.com
  route53ProviderConfig:
    accessKey: <IAM Access Key ID in clear text>
    secretKey: <IAM Secret Access Key in clear text>
```

**Result:**
The IAM credentials used to configure the Global DNS provider are shown in clear text.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
`2.2.4`
- Installation option (single install/HA):
`HA`

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

alena1108:
We can leverage "Storing credentials in secrets" feature for that.

alena1108:
@sangeethah as a part of testing, we should check the upgrade use case. Globaldns created prior to upgrade, should continue working.

sowmyav27:
Upgrade from 2.2.6 to master-head
In 2.2.6, create 2 clusters.
Add Global DNS provider and Global DNS entry pointing to cluster1 and project A.
Add a workload and ing pointing to the global DNS in  cluster 1 project A.
Verify global DNS provider resource. The secretKey is seen in plain text.

```
apiVersion: management.cattle.io/v3
kind: GlobalDnsProvider
metadata:
  annotations:
[...]

PR #21746, opened 2019-07-24
Save credentials as secrets for globaldnsprovider

prachidamle:
https://github.com/rancher/rancher/issues/21369

### DIFF
--- a/pkg/api/store/globaldns/globaldnsprovider_store.go
+++ b/pkg/api/store/globaldns/globaldnsprovider_store.go
@@ -1,9 +1,14 @@
 package globaldns
 
 import (
+	"strings"
+
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/convert"
+	"github.com/rancher/norman/types/values"
+	"github.com/rancher/rancher/pkg/namespace"
 	managementv3 "github.com/rancher/types/client/management/v3"
 )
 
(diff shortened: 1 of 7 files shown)


alena1108:
@prachidamle have we run the backwards compatibility test? Also can we add an integration/unit test for it. I believe there is a way to make direct api calls to bypass rancher apis in integration tests (@mrajashree did it for one of her prs); this way you can fetch global dns crd and check that the field has become a secret ref.

prachidamle:
@alena1108 yes ran the backwards compatibility test. On existing crds, any update to the globaldnsprovider will turn the keys to secrets and deploy new pod.

kinarashah:
> @prachidamle have we run the backwards compatibility test? Also can we add an integration/unit test for it. I believe there is a way to make direct api calls to bypass rancher apis in integration tests (@mrajashree did it for one of her prs); this way you can fetch global dns crd and check that the field has become a secret ref.

Just for reference, `integration/suite/test_password_store.py` already has similar tests for logging configs

alena1108:
LGTM, merge after integration test is in

pkg/controllers/management/globaldns/globaldnsprovider_secret_syncer.go
```diff
@@ -0,0 +1,86 @@
...
+//sync is called periodically and on real updates
+func (n *ProviderSecretSyncer) sync(key string, obj *corev1.Secret) (runtime.Object, error) {
+	if obj == nil || obj.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	return n.reconcileAllGlobalDNSProviders(key)
+}
+
+func (n *ProviderSecretSyncer) reconcileAllGlobalDNSProviders(key string) (runtime.Object, error) {
+	globalDNSProviderList, err := n.globalDNSProviderLister.List(namespace.GlobalNamespace, labels.NewSelector())
```
alena1108: @prachidamle you should only reconcile if the secret belongs to the cattle global namespace

pkg/api/store/globaldns/globaldnsprovider_store.go
```diff
@@ -46,3 +52,59 @@ func canDeleteProvider(apiContext *types.APIContext, id string) error {
...
+
+	if !passwordFound {
+		existingProvider, err := p.ByID(apiContext, schema, id)
+		if err != nil {
+			return nil, err
+		}
+		existingPassword, _ := values.GetValue(existingProvider, keys...)
+		if !strings.HasPrefix(convert.ToString(existingPassword), namespace.GlobalNamespace) {
+			values.PutValue(data, existingPassword, keys...)
+			logrus.Infof("ProviderPwdStore updated data %v", data)
```
alena1108: better log it in Debug

prachidamle: removed it, had added for my debugging

alena1108:
LGTM

Issue #19737, opened 2019-04-19
Catalogs cannot handle commas in your passowrd


Oats87:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Install a Catalog App with an automatically generated helm chart

**Result:**
Sometimes, a password may be generated with invalid characters in it, like `,`

When this happens, you will get the error message:
```
Helm template failed. Error: failed parsing --set data: key "gA92nmfSK" has no value (cannot end with ,) : exit status 1
```

where the password generated was `lhmn,gA92nmfSK`

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2
- Installation option (single install/HA): doesn't matter

loganhz:
Which catalog app? I guess we can fix it in the chart as `valid_chars ` works for password.

Related PR https://github.com/rancher/ui/pull/2505

Oats87:
@loganhz this is a custom catalog app

vincent99:
That's a perfectly valid random password, being not escaped properly somewhere along the line.  Answers should be able to have a comma in them.

fgeorgatos:
i also think character `#` gets mishandled. 

Could someone from the rancher team (@cjellick ?) state if such passwords are meant to be supported?

cjellick:
Yep, meant to be supported. We can address as part of this bug.

dramich:
https://github.com/helm/helm/issues/1556

dramich:
The issue comes from the helm parser splitting on commas when using the --set command as we do. So our command looks something like `--set "a=b,c=d` which is fine but if you include a comma in the value the parser breaks. Well, you can quote the pairs and escape all the commas but now something that _should_ work in helm, won't. For example the lists `a={a,b,c}` they allow for [set](https://github.com/helm/helm/blob/master/docs/using_helm.md#the-format-and-limitations-of---set). So this fix needs to take all of this into account - quoting the pairs and escaping commas but only when they should be escaped and won't break set.

cjellick:
Let's limit to escaping commas (and other prolematic characters) when WE generate passwords.
Lets add a general UI note that reference the fact that users must escape commas in their own answers. https://rancher.com/docs/rancher/v2.x/en/catalog/apps/?origin_team=T02RW4JDH#configuration-options

bmdepesa:
[...]

PR #21762, opened 2019-07-24
Escape commas in chart values

rmweir:
Problem:
Values containing commas result in an installation error. This is due to helm using commas as a delimiter.

Solution:
Values that are not arrays may contain commas and they will be passed to helm in an acceptable form.

Issue:
https://github.com/rancher/rancher/issues/19737

### DIFF
--- a/pkg/controllers/user/helm/common/common.go
+++ b/pkg/controllers/user/helm/common/common.go
@@ -179,10 +179,13 @@ func GenerateAnswerSetValues(app *v3.App, tempDir *HelmPath, extraArgs map[strin
 			if _, ok := extraArgs[k]; ok {
 				continue
 			}
-			values = append(values, fmt.Sprintf("%s=%s", k, v))
+			// helm will only accept escaped commas in values
+			escapedValue := fmt.Sprintf("%s=%s", k, escapeCommas(v))
+			values = append(values, escapedValue)
 		}
 		for k, v := range extraArgs {
-			values = append(values, fmt.Sprintf("%s=%s", k, v))
+			escapedValue := fmt.Sprintf("%s=%s", k, escapeCommas(v))
+			values = append(values, escapedValue)
 		}
 		setValues = append(setValues, "--set", strings.Join(values, ","))
 	}
@@ -218,4 +221,17 @@ func JailCommand(cmd *exec.Cmd, jailPath string) (*exec.Cmd, error) {
 	cmd.SysProcAttr.Chroot = jailPath
 	cmd.Env = jailer.WhitelistEnvvars(cmd.Env)
 	return cmd, nil
+}
+
+// escapeCommas will escape the commas in a string, unless helm would identify it as a list
+func escapeCommas(value string) string {
+	if len(value) == 0 {
+		return value
+	}
+	// as far as helm is concerned, an answer starting with '{' is a list
+	// commas in lists are not escaped for users as they are also used as delimiters
+	if strings.HasPrefix(value, "{") {
+		return value
+	}
+	return strings.Replace(value, ",", "\\,", -1)
 }
--- a/pkg/controllers/user/helm/common/common_test.go
+++ b/pkg/controllers/user/helm/common/common_test.go
@@ -1,0 +1,46 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type testCase struct {
+	input  string
+	expect string
+}
+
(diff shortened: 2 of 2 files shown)


pkg/controllers/user/helm/common/common.go
```diff
@@ -179,10 +179,13 @@ func GenerateAnswerSetValues(app *v3.App, tempDir *HelmPath, extraArgs map[strin
 			if _, ok := extraArgs[k]; ok {
 				continue
 			}
-			values = append(values, fmt.Sprintf("%s=%s", k, v))
+			// helm will only accept escaped commas in values
+			escapedValue := escapeCommas(v)
+			values = append(values, fmt.Sprintf("%s=%s", k, escapedValue))
```
luthermonson: seems pointless to set a variable to just pass it, just pass escapeCommans(v) as the 3rd param to sprintf since there is no error checking or anything. minor and not necessary for lgtm fi you want to ignore but throwing it out there.

pkg/controllers/user/helm/common/common.go
```diff
@@ -219,3 +222,17 @@ func JailCommand(cmd *exec.Cmd, jailPath string) (*exec.Cmd, error) {
...
+func escapeCommas(value string) string {
+	if len(value) == 0 {
+		return value
+	}
+	// as far as helm is concerned, an answer starting with '{' is a list
+	// commas in lists are not escaped for users as they are also used as delimiters
+	if string(value[0]) == "{" {
+		return value
+	}
+	escapedAnswer := strings.Replace(value, ",", "\\,", -1)
```
luthermonson: same thing on on setting param? why not just `return strings.Replace()`

pkg/controllers/user/helm/common/common.go
```diff
@@ -219,3 +222,17 @@ func JailCommand(cmd *exec.Cmd, jailPath string) (*exec.Cmd, error) {
...
 }
+
+// escapeCommas will escape the commas in a string, unless helm would identify it as a list
+func escapeCommas(value string) string {
+	if len(value) == 0 {
+		return value
+	}
+	// as far as helm is concerned, an answer starting with '{' is a list
+	// commas in lists are not escaped for users as they are also used as delimiters
+	if string(value[0]) == "{" {
```
luthermonson: no need to confirm an ending `}` and validate the input before returning? just assume user input is correct and move on? does helm break and send back proper errors if you pass in a list like this `{a,b,c` and don't close it?

luthermonson:
lgtm

pkg/controllers/user/helm/common/common.go
cbron: Had this same thought but assuming it was resolved.

pkg/controllers/user/helm/common/common_test.go
```diff
@@ -0,0 +1,45 @@
...
+		newTestCase("asdf}", "asdf}"),
+		newTestCase("asd}asdf", "asd}asdf"),
+		newTestCase("asd{asdf}dasf", "asd{asdf}dasf"),
+		newTestCase("as,df", "as\\,df"), // value not a list so commas should be escaped
+		newTestCase("asd,f{", "asd\\,f{"),
+		newTestCase("as,df}", "as\\,df}"),
+		newTestCase("asd}as,df", "asd}as\\,df"),
+		newTestCase("asd{asdf}da,sf", "asd{asdf}da\\,sf"),
+		newTestCase(",asdf", "\\,asdf"),
+		newTestCase("{,asdf", "{,asdf"), // helm would recognize as a list, commas should not be escaped
```
cbron: So `{a,b}` would not be escaped either ? Same below with `{{a,b}}`.

pkg/controllers/user/helm/common/common.go
```diff
@@ -219,3 +222,16 @@ func JailCommand(cmd *exec.Cmd, jailPath string) (*exec.Cmd, error) {
...
```
dramich: Why not use strings.HasPrefix

pkg/controllers/user/helm/common/common_test.go
```diff
@@ -0,0 +1,46 @@
...
+func newTestCase(input string, expect string) testCase {
+	return testCase{
+		input,
+		expect,
+	}
+}
+func TestEscapeCommas(t *testing.T) {
+	assert := assert.New(t)
+
+	invalidArguments := []testCase{
```
dramich: Nice testing

dramich:
lgtm

PR #21763, opened 2019-07-24
helm-regex-setup

cbron:


pkg/catalog/helm/setup.go
```diff
@@ -43,7 +46,11 @@ func NewNoUpdate(catalog *v3.Catalog) (*Helm, error) {
 	if catalog == nil || catalog.Name == "" {
 		return nil, errors.New("Catalog is not defined for helm")
 	}
-	return newCache(catalog), nil
+	cachedCatalog, err := newCache(catalog)
```
dramich: Just return this call, it's the same thing

pkg/catalog/helm/setup.go
```diff
@@ -64,7 +71,12 @@ func CatalogSHA256Hash(catalog *v3.Catalog) string {
 	return hex.EncodeToString(hashBytes[:])
 }
 
-func newCache(catalog *v3.Catalog) *Helm {
+func newCache(catalog *v3.Catalog) (*Helm, error) {
+	sanitizedURL, err := validateURL(catalog.Spec.URL)
+	if err != nil {
+		return nil, err
+	}
+	catalog.Spec.URL = sanitizedURL
```
dramich: Don't save this on the catalog, just use the `sanitizedURL` when creating the Helm struct

cbron: Thats what I did at first, but the problem with that approach is that `CatalogSHA256Hash` uses `catalog` with the non-sanitized URL first and therefore won't match. Also this has the benefit that it strips out weird chars from the UX.

pkg/catalog/helm/setup.go
cbron: fixed

dramich:
This is going to be moved into a validator

pkg/api/customization/catalog/validator.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/rancher/norman/types"
+)
+
+var (
+	controlChars   = regexp.MustCompile("[[:cntrl:]]")
+	controlEncoded = regexp.MustCompile("%[0-1][0-9,a-f,A-F]")
+)
+
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if url, ok := data["url"].(string); ok {
```
dramich: Let's also return an error if the URL is blank

pkg/api/customization/catalog/validator.go
```diff
@@ -0,0 +1,42 @@
...
+		sanitizedURL, err := validateURL(url)
+		if err != nil {
+			return err
```
dramich: Errors coming out of validators should be wrapped, use `httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("%v", err))`

dramich:
Some integration tests would be ideal for this.

pkg/api/customization/catalog/validator.go
cbron: fixed

pkg/api/customization/catalog/validator.go
cbron: fixed

pkg/api/customization/catalog/validator.go
cbron: Needed to move this whole validation to POST only, as an update would fail because it doesn't supply URL.

pkg/api/customization/catalog/validator.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/rancher/norman/types"
+)
+
+var (
+	controlChars   = regexp.MustCompile("[[:cntrl:]]")
+	controlEncoded = regexp.MustCompile("%[0-1][0-9,a-f,A-F]")
+)
+
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if request.Method == http.MethodPost || request.Method == http.MethodPut {
```
dramich: Validator only gets called on create and update so this check isn't needed

pkg/api/customization/catalog/validator.go
```diff
@@ -0,0 +1,52 @@
...
+		url, ok := data["url"].(string)
+		if ok && len(url) > 0 {
+			sanitizedURL, err := validateURL(url)
+			if err != nil {
+				return httperror.NewAPIError(httperror.InvalidFormat, fmt.Sprintf("%v", err))
```
dramich: Use httperror.InvalidBodyContent

pkg/api/customization/catalog/validator.go
```diff
@@ -0,0 +1,52 @@
...
+			}
+			data["url"] = sanitizedURL
+			return nil
+		} else if request.Method == http.MethodPost {
+			return httperror.NewAPIError(httperror.InvalidFormat, "Catalog URL not specified")
```
dramich: Use httperror.MissingRequired

tests/integration/suite/test_catalog.py
```diff
@@ -26,6 +27,55 @@ def test_catalog(admin_mc):
     wait_for_template_to_be_deleted(client, name2)
 
 
+def test_invalid_catalog(admin_mc):
+    client = admin_mc.client
+    name = random_str()
+    bad_url = "git://github.com/StrongMonkey/charts-1.git"
+    # POST: Bad URL
+    with pytest.raises(ApiError) as e:
+        client.create_catalog(name=name,
```
dramich: save the catalog and pass it to the `remove_resource` fixture in case it actually does get created.

tests/integration/suite/test_catalog.py
```diff
@@ -26,6 +27,55 @@ def test_catalog(admin_mc):
...
+    catalog = client.create_catalog(name=name,
+                                    branch="test",
+                                    url=good_url,
+                                    )
+    wait_for_template_to_be_created(client, name)
+    with pytest.raises(ApiError) as e:
+        catalog.url = bad_url
+        client.update_by_id_catalog(catalog.id, catalog)
+    assert e.value.error.status == 422
+    client.delete(catalog)
```
dramich: use `remove_resource` fixture

tests/integration/suite/test_catalog.py
```diff
@@ -26,6 +27,55 @@ def test_catalog(admin_mc):
...
+    wait_for_template_to_be_deleted(client, name)
+
+
+def test_invalid_catalog_chars(admin_mc):
+    client = admin_mc.client
+    name = random_str()
+    url = "https://github.com/%0dStrongMonkey%0A/charts-1.git"
+    catalog = client.create_catalog(name=name,
```
dramich: use `remove_resource` fixture.

pkg/api/customization/catalog/validator.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+
+	"github.com/rancher/norman/httperror"
+
```
dramich: Drop the blank line

dramich:
lgtm

PR #21774, opened 2019-07-25
package linode during build

kinarashah:
https://github.com/rancher/rancher/issues/21300

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -21,10 +21,14 @@ ENV CATTLE_MACHINE_VERSION v0.15.0-rancher9-1
 ENV LOGLEVEL_VERSION v0.1.2
 ENV TINI_VERSION v0.18.0
 ENV TELEMETRY_VERSION v0.5.6
-
+ENV DOCKER_MACHINE_LINODE_VERSION v0.1.7
+ENV LINODE_UI_DRIVER_VERSION v0.3.0-alpha.1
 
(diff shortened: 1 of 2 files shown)


alena1108:
@kinarashah can you check if we need to put changes for it to work with https://github.com/rancher/rancher/pull/21745

kinarashah:
@alena1108 Updated PR to use the correct uiURL. Now, UI would always fetch from the local files `/assets/rancher-ui-driver-linode/component.js`, unless the driver's custom URL is changed. 
Don't think I need any changes for proxy/whitelisting domains for the linked PR because they're already restrictive enough, `"linode.github.io", "api.linode.com"`

package/Dockerfile
```diff
@@ -61,6 +65,12 @@ VOLUME /var/log/auditlog
 RUN mkdir -p /usr/share/rancher/ui && \
     cd /usr/share/rancher/ui && \
     curl -sL https://releases.rancher.com/ui/${CATTLE_UI_VERSION}.tar.gz | tar xvzf - --strip-components=1 && \
+    cd assets && \
+    mkdir -p rancher-ui-driver-linode && \
```
dramich: use `mkdir -p assets/rancher-ui-driver-linode && \`
then `cd assets/rancher-ui-driver-linode` just in case assets isn't there

dramich:
lgtm

Issue #20204, opened 2019-05-13
Not able to see any revisions for the apps.


sangeethah:
Rancher server version - v2.2.2-patch1-rc2

Steps to reproduce the problem:
Create cluster1 as admin user.
Add a regular user as cluster owner.
Log in as this regular user and deploy app in any project in cluster1.
Once the app is deployed , user is not able see the revisions of this app when launching the "Rollback" view for the app.

Note - User is able to upgrade apps successfully . He is just not able to see the revisions of this app.

This behavior is also seen for a project owner.

kinarashah:
Note: not a regression, present in 2.1.x versions too

bmdepesa:
rancher/rancher:master-head bc3700f8

* Create cluster1 as admin user.
* Add a regular user as cluster owner.
* Deploy an app as admin
* Log in as this regular user and deploy app in any project in cluster1.
* User is able to view revisions in both apps

PR #21776, opened 2019-07-25
Cluster/project owners should be able to see app revisions

cbron:
Issue: https://github.com/rancher/rancher/issues/20204

* Adding to `projectManagmentPlaneResources` is required to allow users who are solely cluster owners to access revisions.
* Adding to role_data.go is required to allow project specific users to access revisions, it's matched with `apps` permissions for uniformity, since apps and their revisions should be 1:1. 

### QA

1. Create a user that is a cluster owner on a non-local cluster. Launch an app, after it loads you should be able to see revisions. 
2. Same as step 1 but as a project owner.

### DIFF
--- a/tests/integration/suite/test_rbac.py
+++ b/tests/integration/suite/test_rbac.py
@@ -495,9 +495,20 @@ def test_member_can_perform_app_action(admin_mc, admin_pc, remove_resource,
         action_name="upgrade",
         answers={"asdf": "asdf"})
 
-    '''
-        TODO: write rollback test, currently blocked by issue #20204
-    '''
+    def _app_revisions_exist():
+        a = admin_pc.client.reload(app)
+        return len(a.revision().data) > 0
+
+    wait_for(_app_revisions_exist, fail_handler=lambda: 'no revisions exist')
+    proj_user_client = user_project_client(user_mc, project)
+    app = proj_user_client.reload(app)
+    revID = app.revision().data[0]['id']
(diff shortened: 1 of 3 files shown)


tests/integration/suite/test_rbac.py
```diff
@@ -495,9 +495,19 @@ def is_installing():
...
 
-    '''
-        TODO: write rollback test, currently blocked by issue #20204
-    '''
+    def _app_revisions_exist():
+        a = admin_pc.client.reload(app)
+        return len(a.revision().data) > 0
+
+    wait_for(_app_revisions_exist, fail_handler=lambda: 'no revisions exist')
+    app = admin_pc.client.reload(app)
```
dramich: This is the admin, who can always see app revisions, this needs to be based off the user client

cbron: Technically the rollback would have failed anyway because the user can't access that revision, I tested and confirmed that, but I updated it to check here too. I'm leaving the app_revision check as `admin_pc` because a failure there would be ambiguous between a permissions issue or an infrastructure issue.

cbron: Technically the test would have failed anyway because the user can't see the rollback, I tested and confirmed that, but I updated this to check too. I'm leaving the `_app_revisions_exist` check as `admin_pc` because a failure there would be ambiguous, it could mean either a permissions issue or an infrastructure issue.

dramich:
lgtm

PR #21794, opened 2019-07-26
Fix node taints not showing

orangedeng:
**Problem:**
The node taints won't be set if there have duplicated taints in
node template and node pool.

**Solution:**
For the new node from node pool, the taints with same key and template
will respect the one from the node pool. The other taints from the node
template will be added to newly create node.

### DIFF
--- a/pkg/controllers/user/nodesyncer/nodetaints.go
+++ b/pkg/controllers/user/nodesyncer/nodetaints.go
@@ -2,11 +2,13 @@ package nodesyncer
 
 import (
 	"reflect"
+	"strings"
 
 	"github.com/pkg/errors"
 	nodehelper "github.com/rancher/rancher/pkg/node"
 	"github.com/rancher/rancher/pkg/taints"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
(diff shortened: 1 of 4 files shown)


pkg/controllers/user/nodesyncer/nodetaints.go
```diff
@@ -41,14 +43,21 @@ func (n *nodesSyncer) syncTaints(key string, obj *v3.Node) (runtime.Object, erro
 			taintList = append(taintList, taintStr)
 		}
 		newNode.Spec.Taints = taintList
-		if _, err := n.nodeClient.Update(newNode); err != nil {
+		if _, err := n.nodeClient.Update(newNode); err != nil && !isDuplicate(err) {
 			return obj, errors.Wrapf(err, "failed to update corev1.Node %s from v3.Node %s in node taint controller", node.Name, obj.Name)
-		}
-		if !reflect.DeepEqual(newObj.Spec.DesiredNodeTaints, newObj.Spec.InternalNodeSpec.Taints) {
+		} else if isDuplicate(err) {
```
kinarashah: I think this will error out even if there's one duplicate, do we want to set desired taints to nil in this case? 
Is there any way we can calculate taintList so that this conflict doesn't happen?

loganhz:
I tried to add taints for custom nodes. But it doesn't work. @orangedeng please fix it in this PR as well.

pkg/controllers/user/nodesyncer/nodetaints.go
orangedeng: The point for this duplicate check is that, in k8s >=1.14, they make key and effect as unique but only key in version <=1.13. And I don't want to block the following sync if this error happens.

kinarashah: we should add api validation in store update similar to what UI does so it's clearly visible to user

orangedeng: Updated

orangedeng:
Updated.

PR #21813, opened 2019-07-27
Fix marshal v1.Taint to yaml panic

orangedeng:
**Problem:**
Marshal v1.Taint with timeAdded nil value to yaml panic.

**Solution:**
Use our own taint struct with omitempty flag of timeAdded field.

Related PR: 
https://github.com/rancher/types/pull/923

Issues:


loganhz:
LGTM

pkg/controllers/management/node/controller.go
```diff
@@ -122,7 +122,7 @@ func (m *Lifecycle) setupCustom(obj *v3.Node) {
 		},
 	}
 
-	obj.Spec.DesiredNodeTaints = obj.Status.NodeConfig.Taints
+	obj.Spec.DesiredNodeTaints = taints.GetTaintsFromStrings(obj.Spec.CustomConfig.Taints)
```
kinarashah: isn't this supposed to be `obj.Status.NodeConfig`?

kinarashah: Oh nvm, I saw panic happens only for v3

alena1108:
@orangedeng can you confirm that we've tested the following case:

* While on rancher 2.2.x, user set the taints for nodes manually using kubectl
* Rancher upgraded to 2.3. User set taints should be preserved.
* Update cluster via UI. Make sure user set taints still remain.

orangedeng:
Yes. Tested this case.
The logic for `nodetaints` controller is to add/remove taints to current set. The edit node taints API expect a desired node taints set for input. We will compare them with current taints on node and update v1.node.
We also have a controller to sync taints from v1.Node to v3.Node to keep the current set updated.

alena1108:
LGTM

orangedeng:
Rebased.

Issue #21355, opened 2019-07-07
User with a cloned project owner role cannot enable project monitoring


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a cluster, add it to Rancher, create a "projectmon" project.
2. In the global view, go to security -> roles -> projects and select "clone" for the Project Owner role.
3. Name the project "projectownerclone" and change Project Creator Default to no. Keep other default selections. Click Create button.
4. Create a new Standard user called "usermon"
5. Go to the projectownerclone, click Add Member button.
6. Select "usermon" with "projectownerclone" role and click Create.
7. Log in as "usermon" in another browser.
8. Go to the "projectmon" project.
9. Go to Tools -> monitoring
10. Click on the Enable Monitoring button.

**Result:**
Seeing error `EnableMonitoring Error Unknown action: enableMonitoring` and `Error Unknown action: enableMonitoring`

**Other details that may be helpful:**
Expectation is that if project owner role is cloned and applied to a member in the project, that user should have ability to enable monitoring.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.2.4
- Installation option (single install/HA):
single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Imported

- Machine type (cloud/VM/metal) and specifications (CPU/memory):
k3d
- Kubernetes version (use `kubectl version`):

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T18:55:03Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1-k3s.4", GitCommit:"52f3b42401c93c36467f1fd6d294a3aba26c7def", GitTreeState:"clean", BuildDate:"2019-04-15T22:13+00:00Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):

```
$ docker version
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
[...]

PR #21836, opened 2019-07-29
Check inherited role for project/cluster member

mrajashree:
The auth controller creates rolebindings for a project member in the
project's cluster's namespace along with those in project's namespaces.
The rolebinding in cluster's ns, is either projectmember or projectowner,
depending on the prtb's roleTemplateName. This doesn't take into consideration
the inherited/cloned roles. This commit gets the parent builtin role
and creates projectmember or projectowner rb accordingly.
Same for cluster rb

https://github.com/rancher/rancher/issues/21355

The check against circular roles while getting parent roles recursively needs to be done at api level. We should disallow adding project/cluster members with roles that have circular reference here pkg/api/customization/roletemplatebinding/validator.go

Not adding this check in the controller logic, because by that point the prtb/crtb is already created, so adding the check won't help.
Opening a separate issue to track this: https://github.com/rancher/rancher/issues/21867

### DIFF
--- a/pkg/controllers/management/auth/manager.go
+++ b/pkg/controllers/management/auth/manager.go
@@ -762,3 +762,28 @@ func buildRule(resource string, verbs map[string]bool) v1.PolicyRule {
 		APIGroups: []string{"*"},
 	}
 }
+
+func (m *manager) checkReferencedRoles(roleTemplateName string) (bool, error) {
+	roleTemplate, err := m.rtLister.Get("", roleTemplateName)
+	if err != nil {
+		return false, err
+	}
+
+	if projectScopedAdminRoles[roleTemplate.Name] || roleTemplate.Administrative {
+		return true, nil
+	}
+	isOwnerRole := false
+	if len(roleTemplate.RoleTemplateNames) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/controllers/management/auth/manager.go
```diff
@@ -762,3 +762,18 @@ func buildRule(resource string, verbs map[string]bool) v1.PolicyRule {
...
+
+func (m *manager) getBuiltinParentRole(roleTemplateName string) (*v3.RoleTemplate, error) {
+	roleTemplate, err := m.rtLister.Get("", roleTemplateName)
+	if err != nil {
+		return nil, err
+	}
+	if !roleTemplate.Builtin && len(roleTemplate.RoleTemplateNames) > 0 {
+		// get referenced roletemplate
+		for _, rt := range roleTemplate.RoleTemplateNames {
+			return m.getBuiltinParentRole(rt)
```
kinarashah: q: what's the loop for, looks like we always return

mrajashree: Getting parent roles for the roles that the current roletemplate is referencing recursively

Issue #21636, opened 2019-07-19
AKS Cluster Deployment fails in South Central US


jambajaar:
Rancher version:
v2.2.5

Type:
Bug

Description:
deployment of AKS cluster on Azure South Central US failed w/ error below:

Operationalinsights.WorkspacesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="RequestDisallowedByPolicy" Message="Resource 'aksdevinternal01-southcentralus-rg-EUS' was disallowed by policy. 

It seems like Rancher is targeting resource group on Azure (AKS) Eastern US for South Central US deployment

davidnuzik:
Version: v2.2.5

I wasn't able to reproduce the issue. Keep in mind this doesn't invalidate your claims -- it's possible I couldn't reproduce due to some other factor(s). With our QA test account on Azure I was successfully able to bring up an Azure AKS cluster by picking the South Central US region.

The HTTP/API appears correct to me so frontend looks good. When I checked on the Azure web console the cluster and VMs are created in the correct region, South Central US. I didn't observe the error mentioned in the UI nor in the rancher logs.

I tested with a single-node rancher running v2.2.5 release.

Some guesses as to what may have gone wrong:
- Our Azure accounts differ somehow (not sure how/where)
- On the client side, browser wasn't fully updated (such as if this was an upgrade) and so for some reason UI has some stale cached files and so didn't provision correctly.

davidnuzik:
What was the exact "Cluster Resource Group" name you supplied as part of creating the AKS cluster? Was this done through the UI or API?

trananthony:
We are using an existing RG that is empty.  The app id has ownership of the RG but not the subscription.

trananthony:
aks cluster deployment through the UI

davidnuzik:
I reproduced the issue.

v2.2.5
1. Using Azure webconsole / cli create a Resource Group, make the location South Central US. (It should be empty).
2. Create an AKS cluster, use the name of the RG created in step 1.

Result:
The Log Analytics workspace and Solution objects are in East US region. They should be in the South Central US region...
![image](https://user-images.githubusercontent.com/45179589/61977336-0af6fe80-afa3-11e9-80ce-51f9e06ed8fa.png)

davidnuzik:
I took a deeper look, again with v2.2.5 although all versions should currently be affected. Here is some additional information:

[...]

PR #21841, opened 2019-07-30
Update kontainer-engine

rmweir:
Problem:
Logging resources are being created in the incorrect region

Solution:
update to kontainer-engine that includes correct region mappings for logging

Issue:
https://github.com/rancher/rancher/issues/21636


dramich:
lgtm

PR #21853, opened 2019-07-30
Change permissions on cluster and node agent

daxmc99:
The previous permissions defaulted to 644. The permissions are now
500 (octal) which is 320 in decimal.
P.S. When modifying defaultMode use decimal always to avoid conversion
errors. chmod 500 -> defaultMode: 320


dramich:
lgtm

Issue #20911, opened 2019-06-17
`Manage Cluster Backups` cluster role is not sufficient to display and create backups


nazarewk:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. assign `Manage Cluster Backups` to user
2. create a snapshot
3. log in with that user
4. either of:
	- try to restore snapshot
    - go to `v3/clusters/<CLUSTER_ID>/etcdbackups`

**Result:**
backups list is empty even though there are backups

**Other details that may be helpful:**
it's simple enough
I don't know whether it affects other versions of rancher

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.3
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): AWS EC2 instances
- Kubernetes version (use `kubectl version`): `1.13.5`

- Docker version (use `docker version`): `18.6.1`

nazarewk:
Update: Owner of the cluster also does not allow listing etcd backups

mrajashree:
@nazarewk Was the user created with the standard "user" global role or some custom role?

mrajashree:
Users created with "User" global role won't face this issue.
I was able to repro this with a user created with a custom global role granting login access only. Basically any role except for the "User" global role will face this issue unless that role explicitly grants etcdbackups permissions.

nazarewk:
User was created with `User Base` role so we have users that cannot create clusters by default.

mrajashree:
Also found these issues:
1. Project members can view etcdbackups
2. Users with "user-base" role added to a cluster with Manage Backups role cannot create etcdbackups

Expected:
1. Only cluster members can view etcdbackups
2. Users with either "user" or "user-base" global role should be able to create etcdbackups if they are added to a cluster either as "owner" or with "Manage backups" role

nazarewk:
@mrajashree any idea how can i grant permission to create backups on `2.2.6`?
Currently i've user that has `User Base` + following rules (created custom `GlobalRole` through kubernetest API on rancher cluster):
```json
{
  "rules": [
    {
      "apiGroups": [
        "management.cattle.io"
      ],
      "resources": [
        "etcdbackups"
      ],
      "type": "/v3/schemas/policyRule",
[...]

PR #21889, opened 2019-07-31
Create rolebindings for etcdbackups in cluster's ns

mrajashree:
https://github.com/rancher/rancher/issues/20911

Although etcdbackups are scoped to a cluster, they are rancher mgmt resources,
and need to be added to the `clusterManagmentPlaneResources` list. The crtb_handler.go
ensures all cluster members get access to resources in this list by creating
role bindings.
The "user" global role has get/list/watch permissions on etcdbackups, so cluster
members with 'user" role don't need this and will be able to see etcdbackups
even without this resource being in the `clusterManagmentPlaneResources` list.
But any user having "user-base" global role or created with any custom permissions
will require this change so they can see the etcdbackups in their cluster.


mrajashree:
Not sure how to add tests for this. The integration test will need an rke provisioned cluster, otherwise the option to take snapshot won't be available. Not sure how to add that, and don't see a proper place to add a unit test either.

dramich:
Add `management.cattle.io` for the value

dramich:
lgtm

Issue #20315, opened 2019-05-18
Azure shared disk PV fails to provision from storage class when storage account type is not set


bmdepesa:
If a storage account type is not entered when creating a shared unmanaged disk storage class on Azure Disk, the PV will fail to create. This occurs on k8s versions >= 1.12

To reproduce: 
* Deploy Rancher v2.2.4-rc5
* Create Azure cluster w/ cloud provider, k8s version >= 1.12
* Create a storage class for Azure Disk - shared unmanaged, without specifying the storage type
* Create workload with a new volume from the storage class
* The workload and PV will fail to create.

![Screen Shot 2019-05-17 at 7 08 12 PM](https://user-images.githubusercontent.com/21316915/57963298-283e8b00-78d7-11e9-87dd-f6c58e7ad05e.png)

* If the storage account type is specified, the PV will create regardless of k8s version.
* This does not occur on k8s v1.11.
* This does not occur with the dedicated unmanaged disk.
* This does not seem to be dependent on Rancher version:
  * Created a v1.11 cluster in v2.1.7 without storage account type set
  * PV provisioned after upgrade (before upgrading k8s version)
* This occurs in other versions of Rancher where clusters are created with k8s v1.12+ (v2.1.7, v2.2.2, etc..)
* This occurs in clusters that have been upgraded beyond k8s v1.11 (if a storage class was working in a v1.11 cluster, new volumes will fail to provision after the cluster is upgraded)

kinarashah:
Just to reiterate, this can be reproduced using rancher v2.1.7 too

alena1108:
Moving to v2.2 backlog as the issue is not a regression in 2.2.x. Will include the fix to one of our future releases.

sowmyav27:
### Reproduced the issue on `rancher:2.2.7`.
**Steps:**

- Deploy an Azure cluster - 1 node all roles, with cloud configuration enabled. (k8s v - 1.14)
- Create a storage class for Azure Disk - shared unmanaged, without specifying the storage type.
- Create workload with a new volume from the storage class
- The workload and PV will fail to create.

### Validated in `rancher:master-head-linux-amd64`.
**Steps:**

- Deploy an Azure cluster - 1 node all roles, with cloud configuration enabled.
- Create a storage class for Azure Disk - shared unmanaged, without specifying the storage type.
- Create workload with a new volume from the storage class
- Volume gets provisioned successfully
- Workload is deployed successfully with the volume attached to it.
- 
-

PR #21896, opened 2019-08-01
Azure Disk storage class default

rmweir:
**Problem:**
when creating an azure disk PVC, an error will be received if it references a storage class that is of kind "shared" and has no "storageaccounttype" value. This is a Kubernetes issue and occurs in newer versions (1.12+). This is because when making a storage account, Kubernetes now passes on the empty value, which is not a valid value for the storage account parameter. Kubernetes used to detect the empty value and pass on a default instead.

**Solution:**
Triage by assigning new storage classes the prior default "storageaccounttype" (Standard_LRS) and return an error if PVC references a storage class that does not have a "storageaccounttype".
Ultimately, this should be a fix in kubernetes that make a storage account of default type if one is not given. This was the behavior prior to 1.12.

**Related PRs:**
types: https://github.com/rancher/types/pull/930
kubernetes (requires no action): https://github.com/kubernetes/kubernetes/pull/80837

**Before merging:**
approve/merge types pr
update types in this pr to newest rancher/type commit

**Issue:**
https://github.com/rancher/rancher/issues/20315


pkg/api/store/storageclass/store_test.go
```diff
@@ -0,0 +1,79 @@
...
+	testStore := Wrap(dummyStore{})
+
+	// create with a storageaccounttype of nil should give the storage class a default value for the field
+	data, err := createWithFields(testStore, AzureDisk, "shared", nil)
+	assert.Nil(err)
+	assert.Equal("Standard_LRS", data["parameters"].(map[string]interface{})["storageaccounttype"])
+
+	// create with a kind that is not shared should not make a change to the storageaccounttype
+	data, _ = createWithFields(testStore, AzureDisk, "dedicated", nil)
+	assert.Nil(err)
```
dramich: The err is dropped but you are still asserting it, should just not drop the error.

pkg/api/server/userstored/setup.go
```diff
@@ -19,6 +20,8 @@ import (
 	"github.com/rancher/rancher/pkg/api/store/projectsetter"
 	"github.com/rancher/rancher/pkg/api/store/secret"
 	"github.com/rancher/rancher/pkg/api/store/service"
+
```
dramich: Extra line

pkg/api/store/storageclass/store.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"strings"
+
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+)
+
+const (
+	StorageAccountType = "storageaccounttype"
```
dramich: This doesn't look to be used outside the package

pkg/api/store/storageclass/store.go
```diff
@@ -0,0 +1,47 @@
...
+	storageKind        = "kind"
+	defaultStorageType = "Standard_LRS"
+	parameters         = "parameters"
+	provisioner        = "provisioner"
+	AzureDisk          = "kubernetes.io/azure-disk"
```
dramich: This doesn't look to be used outside the package

pkg/api/store/storageclass/store.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func Wrap(store types.Store) types.Store {
+	return &Store{
+		store,
+	}
+}
+
+type Store struct {
```
dramich: I don't think this needs to be exported

pkg/api/server/userstored/setup.go
```diff
@@ -95,6 +100,20 @@ func SetProjectID(schemas *types.Schemas, clusterManager *clustermanager.Manager
 	}
 }
 
+func StorageClass(schemas *types.Schemas) {
+	storageClassSchema := schemas.Schema(&clusterschema.Version, "storageClass")
+	storageClassSchema.Store = storageclass.Wrap(storageClassSchema.Store)
+}
+
+func PersistentVolumeClaim(cmanager *clustermanager.Manager, mgmt *config.ScaledContext, schemas *types.Schemas) {
```
dramich: mgmt isn't used in the function

dramich:
lgtm

Issue #20955, opened 2019-06-18
[HPA] Active, then Pending (with error) then Active again if create workload then somewhat quickly create HPA


davidnuzik:
Version: v2.3.0-alpha4
Version: master (v2.3) (6/18/19)
This has been an issue since introduction of HPA in Rancher. We are getting expected errors from kubectl. What we are concerned about is the user experience in the UI.

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Create a new workload and set CPU and memory reservation
- Within 20 seconds (a reasonable amount of time), create an HPA for this new workload (Resource type, either CPU or Memory, either Average Utilization or Average Value, any value)

**Result:**
HPA is initially in Active state, then Pending state with "error" transition (Red Pending state in UI due to API), then 10-60 seconds later everything is fine and it is back in the Active state again.

_This causes user confusion with the Red Pending (error) state seen in the UI. Also, it's confusing that it's Active at first then after a moment in Pending then back to Active. We expected to not see it in Active state at first._

We see in the API while Pending:
```
 "state": "pending",
"transitioning": "error",
"transitioningMessage": "the HPA was unable to compute the replica count: unable to get metrics for resource cpu: no metrics returned from resource metrics API",
```
While it is in the Pending state so the UI shows this as a red colored state implying there is an error. This is an expected message (k8s annotation on HPA object) when you first set up an HPA and query it before it is actually ready.

UI screenshots:

Add HPA, is Active at first:
![image](https://user-images.githubusercontent.com/45179589/59710890-c9627f00-91be-11e9-8863-e3e55d7860d8.png)

Several seconds later goes to Pending state and is red (error transition):
![image](https://user-images.githubusercontent.com/45179589/59710918-dd0de580-91be-11e9-811a-3c11eda38a63.png)

Then get a second error shortly thereafter:
![image](https://user-images.githubusercontent.com/45179589/59710954-edbe5b80-91be-11e9-8ed5-5175908f4244.png)

Then error will eventually go away and everything is fine:
![image](https://user-images.githubusercontent.com/45179589/59710974-fca50e00-91be-11e9-8f09-a9479ca2462c.png)

**Other details that may be helpful:**
The HPA will be in transitioning "error" state always. It does not matter if you wait for the workload to fully be Active (ready) or not. It does not matter if you wait a few minutes after deploying the workload or not.

[...]

PR #21542, opened 2019-07-16
Don't set active state to newly created HPA

orangedeng:
**Problem:**
The hpa object needs conditions to indicate the state,
if there are not any, we should set it to initializing.

Related issue:
https://github.com/rancher/rancher/issues/20955

Related PR:
https://github.com/rancher/types/pull/907

PR #21897, opened 2019-08-01
Start APIService controller for API usage

orangedeng:
**Problem:**
The apiservice lister can't return any data by using get/list function.
The problem is that we don't start the apiservice controller
in user/controller.

**Solution:**
Enable APIService controller in user controller.

Related issue:
https://github.com/rancher/rancher/issues/20955

### DIFF
--- a/pkg/controllers/user/controllers.go
+++ b/pkg/controllers/user/controllers.go
@@ -72,6 +72,9 @@ func Register(ctx context.Context, cluster *config.UserContext, clusterRec *mana
 	ingresshostgen.Register(ctx, cluster.UserOnlyContext())
 	windows.Register(ctx, clusterRec, cluster)
 
+	// register controller for API
+	cluster.APIAggregation.APIServices("").Controller()
+
 	if clusterRec.Spec.LocalClusterAuthEndpoint.Enabled {
 		err := clusterauthtoken.CRDSetup(ctx, cluster.UserOnlyContext())
 		if err != nil {


gitlawr:
LGTM

cjellick:
Can you explain in more detail how this fixes issue #20955? I do not see the direct connection

orangedeng:
The issue @davidnuzik pointed out has been fixed by https://github.com/rancher/rancher/pull/21542. Then the newly created HPA will be `initializing` state instead of `active`. After the `AbleToScale` condition appears in HPA and is set `true`, the HPA state will be set active.
But when Logan validated this issue, he found out that https://github.com/rancher/rancher/issues/20955#issuecomment-516855423. The second problem is fixed by PR `https://github.com/rancher/types/pull/929` And this PR is going to fix the first problem.

dramich:
This looks like a revert of https://github.com/rancher/rancher/commit/d9034c8e852eaf579f35360325aec6356c36da09#diff-1de1009377fbeb4507c945bdd07c8527. Why is this being changed back? Do all providers now support this API for the expected version?

orangedeng:
All the version in GKE doesn't support `v2beta2.autoscaling` api group. So I need to disable hpa in the GKE cluster.

orangedeng:
The reason we change back is in this PR message.
As we can't use api service controller in k3s 1.12.5-lite, the integration test will fail if we enable it. So we need to change back to the old way.

cjellick:
Where is the integration test that you are talking about? What cluster is that test running against?

I don't feel comfortable with this change. @loganhz can you setup a call sometime this week with me, @orangedeng, & @dramich to discuss this? Note, I'm not available thursday afternoon

pkg/api/store/k8sversion/store.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+type Store struct {
+	manager       *clustermanager.Manager
+	clusterLister v3.ClusterLister
+	types.Store
+}
+
+func (s *Store) Watch(apiContext *types.APIContext, schema *types.Schema, opt *types.QueryOptions) (chan map[string]interface{}, error) {
```
cjellick: why do you only care about watch? Why does this not impact list or get?

orangedeng:
As we discussed offline, we will hold this until Darren finished upgrading k3s to latest version.

orangedeng:
Updated

PR #21902, opened 2019-08-01
Handling HA install rancher local cluster may has empty value for field dockerRootDir

aiwantaozi:
Problem:
In HA installed rancher, local cluster is created by backend, not UI, field dockerRootDir may have an empty value

Solution:
Check dockerRootDir one more time when enable logging

Issue:

### DIFF
--- a/app/cluster_data.go
+++ b/app/cluster_data.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"github.com/rancher/rancher/pkg/settings"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,6 +18,9 @@ func addLocalCluster(embedded bool, adminName string, management *config.Managem
 		Spec: v3.ClusterSpec{
 			Internal:    true,
 			DisplayName: "local",
+			ClusterSpecBase: v3.ClusterSpecBase{
+				DockerRootDir: settings.InitialDockerRootDir.Get(),
+			},
 		},
 		Status: v3.ClusterStatus{
 			Driver: v3.ClusterDriverImported,
--- a/pkg/controllers/user/logging/deployer/deployer.go
+++ b/pkg/controllers/user/logging/deployer/deployer.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/project"
 	"github.com/rancher/rancher/pkg/ref"
+	"github.com/rancher/rancher/pkg/settings"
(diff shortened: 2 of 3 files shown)


app/cluster_data.go
```diff
@@ -17,6 +17,9 @@ func addLocalCluster(embedded bool, adminName string, management *config.Managem
 		Spec: v3.ClusterSpec{
 			Internal:    true,
 			DisplayName: "local",
+			ClusterSpecBase: v3.ClusterSpecBase{
+				DockerRootDir: "/var/lib/docker",
```
gitlawr: We don't need to set it here if it is the default value

pkg/controllers/user/logging/deployer/deployer.go
```diff
@@ -29,6 +29,10 @@ var (
 	windowNodeLabel = labels.Set(map[string]string{"beta.kubernetes.io/os": "windows"}).AsSelector()
 )
 
+var (
```
gitlawr: Use const

app/cluster_data.go
aiwantaozi: If the crd object is created by the UI, it will have the default value, when the crd object create by the controller directly, it doesn't have the default value

aiwantaozi:
@gitlawr Updated, thanks

gitlawr:
LGTM

app/cluster_data.go
cjellick: Do we really have any idea if the docker root dir of the local cluster is `/var/lib/docker`? This iseems like q too much of an assumption. What if it is legitimately _not_ that?

aiwantaozi: `/var/lib/docker` is just the default value, this value is the same as the cluster user created in UI. Because when creating the local cluster, we don't know the docker root, so we give a default value, allow user could edit it later.

cjellick: talked to @ibuildthecloud about this. can you make this a setting with /var/lib/docker as the default? that at least will let people deploying rancher in HA mode to set the corresponding environment variable to seed the setting and make this work out of the box.
test that this flow works end-to-end

aiwantaozi: Hi @cjellick Hi, the dockerRootDir setting is just for local cluster right? If we add this one in setting, the dockerRootDir for local cluster could be configured in two places, setting and cluster spec, they may be configured with different values, may cause more problem. 

This dockerRootDir just used in logging now, can we remove it? And the user could update rancher-logging app's answers if they want to change the default dockerRootDir.

cjellick: This would be just for the local cluster yes, so the setting could/should be called localClusterDockerRootDir.
We shouldn't remove this field at this time

aiwantaozi: Sure, this setting is just for initial setup.

PR #21908, opened 2019-08-01
Add windows node taint controller

orangedeng:
The linux node taints controller for Windows k8s cluster is not added in user controller. Fix it.

Issue


loganhz:
LGTM

luthermonson:
LGTM

Issue #21412, opened 2019-07-09
Google Auth - Hostname is not validated while enabling google Auth


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug 

**Steps to reproduce (least amount of steps as possible):**

1. Login to Rancher UI
2. Go to Global --> Authentication --> Google
3. Enter valid entries for Admin email, OAuth credentials and Service Account Credentials fields.
4. Enter invalid hostname
5. Click on Authenticate with Google. Enter valid username and password. Click on Allow.
6. Google auth is successfully enabled.

**Expected Result:**
An error message should be shown on the UI, regarding the invalid hostname

**Actual Result:**
Google auth is successfully enabled.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher:master-head
- Installation option (single install/HA): single

sowmyav27:
verified on master-head  - latest - commit id -`bec76ef3`
**Steps**

1. Login to Rancher UI
2. Go to Global --> Authentication --> Google
3. Enter valid entries for Admin email, OAuth credentials and Service Account Credentials fields.
4. Enter invalid hostname
5. Click on Authenticate with Google.
6. Enter valid username and password. Click on Allow.
User sees this on the UI.
<img width="1803" alt="Screen Shot 2019-08-02 at 2 09 01 PM" src="https://user-images.githubusercontent.com/26032343/62399170-c255bd00-b52f-11e9-97bb-4e5cc496a9a2.png">

CyberPoison:
I also facing the same issue how do you managed the way to workarround that issue ? 

Is that a error because of a issue in some config on google ?

PR #21915, opened 2019-08-01
Add hostname validation for google oauth

mrajashree:
https://github.com/rancher/rancher/issues/21412

This commit validates the hostname provided while configuring google oauth. When the test user logs into their google account, we get the hostname as one of the attributes of that user. So this commit checks that the hostname from user's attributes matches the hostname provided in the authconfig.


cjellick:
lgtm

PR #21944, opened 2019-08-02
[2.2] Return the icon url to the template to display in the UI

daxmc99:
Problem: The icon url was not returned to the template so the template
    would not display icons for charts with remote icons.

Solution: Do not fetch the icon but still return the url of the icon
    to the template


cjellick:
LGTM

PR #21949, opened 2019-08-02
Only modify the response if the icon is a url

daxmc99:
If the icon is a url then return the url to the browser. Otherwise,
 serve the content from Rancher and build the link.


cjellick:
lgtm

PR #21952, opened 2019-08-02
RBAC changes for etcdbackups

mrajashree:
This commit removes get/list/watch permissions on etcdbackups that "user"
globalRole had. Because this was also granting project members access to backups.

Also this commit checks if user has "create" permission for etcdbackup for the
backupEtcd actionHandler. Till now it checked if user has the permission to "update"
the cluster. Users added with "backups-manage" role won't have rights to update clusters,
hence won't be able to create etcdbackups either in spite of being granted this role.
This issue is resolved by checking for permissions on etcdbackups instead

https://github.com/rancher/rancher/issues/20911#issuecomment-518332555

### DIFF
--- a/pkg/api/customization/cluster/actionhandler.go
+++ b/pkg/api/customization/cluster/actionhandler.go
@@ -31,6 +31,15 @@ func (a ActionHandler) ClusterActionHandler(actionName string, action *types.Act
 		return apiContext.AccessControl.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", apiContext, cluster, apiContext.Schema) == nil
 	}
 
+	canBackupEtcd := func() bool {
+		etcdBackupSchema := types.Schema{ID: mgmtclient.EtcdBackupType}
+		// pkg/rbac/access_control.go:55 canAccess checks for the object's ID or namespace. The ns for etcdbackup will be the clusterID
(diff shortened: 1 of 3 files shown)


mrajashree:
The integration test only ensures the "user" globalRole does not have "etcdbackups" resource. I tried creating a v3 etcdbackup followed by remove_resource() for it in a test, but although the resource was removed, there was panic in the rancher logs...don't know if the same will happen in drone so decided not to add that in test.

pkg/api/customization/cluster/actionhandler.go
```diff
@@ -56,7 +65,7 @@ func (a ActionHandler) ClusterActionHandler(actionName string, action *types.Act
 		}
 		return a.disableMonitoring(actionName, action, apiContext)
 	case v3.ClusterActionBackupEtcd:
-		if !canUpdateCluster() {
+		if !canBackupEtcd() {
 			return httperror.NewAPIError(httperror.Unauthorized, "can not backup etcd")
 		}
 		return a.BackupEtcdHandler(actionName, action, apiContext)
```
cjellick: should restoreFromEtcdBackup be tied to cluster update or etcdbackup mgmt?

mrajashree: @cjellick I think cluster update since that will actually change the cluster state?

Issue #21321, opened 2019-07-03
feature run flag logs error even when not used


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
run rancher locally

**Result:**
error ill be logged 

> [ERROR] failed to apply feature args: feature argument should be of the form "features=feature1=bool,feature2=bool"

**Other details that may be helpful:**
just need to return nil when config features is empty

rmweir:
To validate: run locally before fix and after and check console for log error

cjellick:
@dramich can you validate

dramich:
Master 9/24: no longer see error when no feature flags are provided

PR #21956, opened 2019-08-03
Do not parse empty features run flag

rmweir:
Problem:
An error is logged if user did not use features run flag. Now, an error will not be logged as no input is valid.

Solution:
Do not attempt to parse features flag if it is empty.

Issue:
https://github.com/rancher/rancher/issues/21321


luthermonson:
lgtm

dramich:
lgtm

PR #21957, opened 2019-08-03
Handle upgrades with previously added charts

daxmc99:
If the chart has been added previously then the url of the icon will
not be included in the iconFilename. Instead of defaulting to serving
the icons, check if they have a url for the icon in the chart. If they
do, proceed to send the url to the UI and do not serve the icon data.

### DIFF
--- a/pkg/api/customization/catalog/template.go
+++ b/pkg/api/customization/catalog/template.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/rancher/norman/types/convert"
 
 	"github.com/rancher/norman/api/access"
@@ -20,14 +22,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func checkUpgrade(t TemplateWrapper, templateVersionName string) (string, error) {
-	templateVersion, err := t.TemplateVersion.Get(templateVersionName, metav1.GetOptions{})
+func getChartIconURL(t TemplateWrapper, templateVersionName string) (string, error) {
+	templateVersion, err := t.TemplateVersions.Get("", templateVersionName)
 	if err != nil {
 		return "", err
 	}
 	id := strings.Split(templateVersion.Name, "-")[1]
 	tag := templateVersion.Spec.Files[id+"/Chart.yaml"]
-	data, err := templatecontent.GetTemplateFromTag(tag, t.TemplateContentClient)
+	data, err := templatecontent.GetTemplateFromTag(tag, t.TemplateContentClients)
 	if err != nil {
 		return "", err
 	}
@@ -58,8 +60,10 @@ func findIconURL(s string) string {
 func indexInterfaceSlice(value interface{}, key string) string {
 	v := convert.Singular(value)
 	if v != nil {
-		v2 := v.(map[string]interface{})
-		return v2[key].(string)
+		v2, ok := v.(map[string]interface{})
+		if ok {
+			return v2[key].(string)
+		}
 	}
 	return ""
 }
@@ -69,10 +73,10 @@ func (t TemplateWrapper) TemplateFormatter(apiContext *types.APIContext, resourc
 	resource.Values["versionLinks"] = extractVersionLinks(apiContext, resource)
 
 	//icon
+	delete(resource.Values, "icon")
 	ic, ok := resource.Values["iconFilename"]
 	if ok {
 		if strings.HasPrefix(ic.(string), "http:") || strings.HasPrefix(ic.(string), "https:") {
-			delete(resource.Values, "icon")
 			resource.Links["icon"] = ic.(string)
 		} else {
 			// before marking an icon as from file, ensure this is not a chart from an upgrade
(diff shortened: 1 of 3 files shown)


pkg/api/customization/catalog/template.go
```diff
@@ -49,6 +112,7 @@ func TemplateFormatter(apiContext *types.APIContext, resource *types.RawResource
 
 type TemplateWrapper struct {
 	TemplateContentClient v3.TemplateContentInterface
+	TemplateVersion       v3.TemplateVersionInterface
```
cjellick: Our convention is for the client property to be plural TemplateVersions (I realize the one right above also does not follow the convention)

pkg/api/customization/catalog/template.go
```diff
@@ -3,20 +3,68 @@ package catalog
...
+	if i < 0 {
+		return ""
+	}
+	target = target[i:]
+	return target
+}
+
+// indexInterfaceSlice takes a list of interfaces, returns the first, asserts its a map[string]interface{}
+// and uses the provided index to retrieve the value
+func indexInterfaceSlice(value interface{}, key string) string {
```
cjellick: looks like we could have some danger of panicking at various points in the function if you get unexpected data, do you want to add some safety checks just so we dont have to worry about that?
like, i think if you simply change `v2 := v.(map[string]interface{})` to `v2, ok := v.(map[string]interface{})`, you prevent that type conversion from panicking

pkg/api/server/managementstored/setup.go
```diff
@@ -182,10 +182,12 @@ func Clusters(schemas *types.Schemas, managementContext *config.ScaledContext, c
 
 func Templates(schemas *types.Schemas, managementContext *config.ScaledContext) {
 	schema := schemas.Schema(&managementschema.Version, client.TemplateType)
-	schema.Formatter = catalog.TemplateFormatter
+
 	wrapper := catalog.TemplateWrapper{
 		TemplateContentClient: managementContext.Management.TemplateContents(""),
+		TemplateVersion:       managementContext.Management.TemplateVersions(""),
```
cjellick: lets change it to a lister. i think it should be fine

pkg/api/customization/catalog/template.go
```diff
@@ -3,20 +3,68 @@ package catalog
...
 	"github.com/rancher/norman/types"
 	"github.com/rancher/rancher/pkg/templatecontent"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	managementschema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
 	client "github.com/rancher/types/client/management/v3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func TemplateFormatter(apiContext *types.APIContext, resource *types.RawResource) {
+func checkUpgrade(t TemplateWrapper, templateVersionName string) (string, error) {
```
cjellick: checkUpgrade is a bad name with very little context in the future. Any idea for a better name?

daxmc99: How about `getChartIconURL` ?

pkg/api/customization/catalog/template.go
daxmc99: Changed type assertion to test if the underlying type exists

pkg/api/customization/catalog/template.go
```diff
@@ -3,35 +3,108 @@ package catalog
...
+				if err == nil {
+					resource.Links["icon"] = res
+					// also update the template itself
+					template, err := t.TemplateClients.Get(resource.ID, metav1.GetOptions{})
+					if err != nil {
+						return
+					}
+					template = template.DeepCopy()
+					template.Spec.IconFilename = res
+					_, err = t.TemplateClients.Update(template)
```
daxmc99: We still need to update the actual `template.IconFilename` for the icon handler to behave normally. This also has an affect of charts with remote icons don't need to be parsed every time.

cjellick: do we need to even get the err variable and return if it is not nil or can we simple completely ignore the returns of this update call?

pkg/api/customization/catalog/template.go
```diff
@@ -3,35 +3,110 @@ package catalog
...
+					if err != nil {
+						logrus.Warnf("unable to update template icon %s", template.Name)
+						return
```
cjellick: i dont want to return out of this function if error is not nil. just log the warning and move on

pkg/api/customization/catalog/template.go
```diff
@@ -3,35 +3,110 @@ package catalog
...
+			ver, ok := resource.Values["versions"]
+			if ok {
+				v := indexInterfaceSlice(ver, "version")
+				res, err := getChartIconURL(t, resource.ID+"-"+v)
+				if err == nil {
+					resource.Links["icon"] = res
+					// also update the template itself
+					template, err := t.TemplateClients.Get(resource.ID, metav1.GetOptions{})
+					if err != nil {
+						logrus.Warnf("unable to get template %s", err)
```
daxmc99: Need to use an if, else here to avoid issues if the value is nil and we try to deepcopy it

cjellick:
lgtm

Issue #21904, opened 2019-08-01
update rancher code to compatible with the prometheus operator crd validation


aiwantaozi:
Rancher server: master
After @guangbochen 's pr https://github.com/rancher/system-charts/pull/72 added validation to prometheus operator crd, we need to update rancher code to compatible with the validation

related issue:
https://github.com/rancher/rancher/issues/19362

Step to reproduce:
1. use dev system-charts, enable monitoring to deploy the crd include validation
2. add notifiers
3. add a cluster level event alert
4. then app rancher-alerting is deployed
5. check rancher server log and find errors
```
2019/08/01 08:53:56 [ERROR] ClusterAlertGroupController local/event-alert [cluster-alert-group-deployer] failed with : failed to get Alertmanager Deployment information, statefulsets.apps "alertmanager-cluster-alerting" not found, [cluster-alert-group-controller] failed with : create prometheus rule cattle-prometheus:local failed, PrometheusRule.monitoring.coreos.com "local" is invalid: []: Invalid value: map[string]interface {}{"kind":"PrometheusRule", "apiVersion":"monitoring.coreos.com/v1", "metadata":map[string]interface {}{"labels":map[string]interface {}{"source":"rancher-alert", "cattle.io/creator":"norman"}, "generation":1, "uid":"e5696d58-b439-11e9-b189-067d93cedea8", "name":"local", "namespace":"cattle-prometheus", "creationTimestamp":"2019-08-01T08:53:56Z"}, "spec":map[string]interface {}{"groups":[]interface {}{map[string]interface {}{"name":"local:event-alert", "rules":interface {}(nil)}}}}: validation failure list:
spec.groups.rules in body must be of type array: "null"
[...]

PR #21966, opened 2019-08-05
update rancher code to compatible with the prometheus operator crd validation

aiwantaozi:
Problem:
After added validation to prometheus operator crd, some crd will get validation failed

Solution:
Update crd objects to match operator crd validation

Issue:
https://github.com/rancher/rancher/issues/21904


orangedeng:
LGTM

Issue #15880, opened 2018-09-29
Deprecated Fluentd configuration


Gambler13:
**Rancher versions:**
rancher/server or rancher/rancher: v2.0.8

**Kubernetes version: v1.10.5**

**Docker version: 17.3.2**

**Steps to Reproduce:**
- Deploy a project which uses a logger with json output
- Enable Cluster or Project level logging with Elasticsearch
- Open generated cluster.config or project.config in ConfigMaps from namespace "Cattle-Logging"

Filter entry with kubernetes:metadata plugin
```
<filter c-7mg4k:project-z4wgs.**>
  @type kubernetes_metadata
  merge_json_log true
  preserve_json_log true
</filter>
```

- Execute Shell on fluentd container and execute
`gem list`
to list all installed fluentd plugins
```
...
fluent-config-regexp-type (1.0.0)
fluent-plugin-elasticsearch (2.11.2)
fluent-plugin-kafka (0.7.4)
**fluent-plugin-kubernetes_metadata_filter (2.1.2)**
fluent-plugin-prometheus (1.0.1)
....
```

Accorind to [fabric8io/fluent-plugin-kubernetes_metadata_filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter) are the values merge_json_log and preserve_json_log deprecated since 2.1.x.

**Results:**
I expect to extract the json values from the log tag string
Current:
```
{"source":"stdout","log":"{\"level\":\"warn\", \"msg\":\"this is a log message\"}","container_id":"3fd8678d487e540c7a303e1613101e746c5012f3317434eda93f24351c1928f7","container_name":"/angry_kalam"}
```

Expected:
```
{"source":"stdout","log":"{\"level\":\"warn\", \"msg\":\"this is a log message\"}","container_id":"3fd8678d487e540c7a303e1613101e746c5012f3317434eda93f24351c1928f7","container_name":"/angry_kalam", "level":"warn", "msg":"this is a log message"}
```

-------

Also tried with following filter after the kubernetes_metadata filter to extract the JSON. Without success...
`<filter c-48mmc:project-r8kk2.**>
  @type parser
  format json # apache2, nginx, etc...
  key_name log
  reserve_data true
</filter>`

kingsd041:
Version: master 2.2 (1/25/19) 
Confirmed fix. `fluent-plugin-kubernetes_metadata_filter` has been downgraded from `2.1.2` to `2.0.0`

**Latest format:**
```
{
[...]

PR #21969, opened 2019-08-05
Remove deprecated Fluentd config

kinolaev:
See https://github.com/rancher/rancher/issues/15880#issuecomment-517671647

Issue:

https://github.com/rancher/rancher/issues/22219

### DIFF
--- a/pkg/controllers/user/logging/generator/templatefilter.go
+++ b/pkg/controllers/user/logging/generator/templatefilter.go
@@ -31,8 +31,6 @@ var FilterTemplate = `
 {{define "filter-container"}}
 <filter  {{ .ContainerLogSourceTag }}.**>
   @type  kubernetes_metadata
-  merge_json_log  true
-  preserve_json_log  true
 </filter>
 {{end}}
 


aiwantaozi:
LGTM

aiwantaozi:
Hi @cjellick Above two configure is for plugin https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter. These two configures are remove since 2.1.x, the latest version is 2.1.6.

Here is the note from fluent-plugin-kubernetes_metadata_filter.
```
NOTE: As of the release 2.1.x of this plugin, it no longer supports parsing the source message into JSON and attaching it to the payload. The following configuration options are removed:

merge_json_log
preserve_json_log

One way of preserving JSON logs can be through the parser plugin
```
If the user still wants to parse the source message info JSON, can use the following ways.
1. Send log to target Fluentd A
2. Parse log inside the target Fluentd A, could use the suggested parser https://docs.fluentd.org/filter/parser, an example
```
<filter foo.bar>
  @type parser
  key_name log
  reserve_data true
  <parse>
    @type json
  </parse>
</filter>
```
3. Target Fluentd A send logs to Elasticsearch/Splunk etc.

aiwantaozi:
Related issue: https://github.com/rancher/rancher/issues/22219

kinolaev:
Thank you @aiwantaozi and @cjellick! What about backport to 2.2? Need I create separate issue and PR for it?

aiwantaozi:
Sure, can you create them? @kinolaev

kinolaev:
Hello @aiwantaozi! Done: issue #22320, PR #22321. Thank you!

PR #21970, opened 2019-08-05
Fix parsing of charts that contain hyphens

daxmc99:
Previously, if a chart contained multiple hyphens it would break the
looking up of that chart.
Now the external id is used which has a consistent format of
catalog://?catalog=<val>&template=<val>&version=<val> which allows for
predictable lookup using the value of the template to create the tag.
This will not break since '&' is an invalid character to name a chart.

### DIFF
--- a/pkg/api/customization/catalog/template.go
+++ b/pkg/api/customization/catalog/template.go
@@ -5,20 +5,20 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
-
-	"github.com/rancher/norman/types/convert"
-
+	"github.com/pkg/errors"
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/rancher/pkg/templatecontent"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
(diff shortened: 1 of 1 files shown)


pkg/api/customization/catalog/template.go
```diff
@@ -27,8 +27,15 @@ func getChartIconURL(t TemplateWrapper, templateVersionName string) (string, err
 	if err != nil {
 		return "", err
 	}
-	id := strings.Split(templateVersion.Name, "-")[1]
-	tag := templateVersion.Spec.Files[id+"/Chart.yaml"]
+	idStruct, err := url.ParseQuery(templateVersion.Spec.ExternalID)
```
dramich: Check the helm package for helpers that already do this.

pkg/api/customization/catalog/template.go
```diff
@@ -27,8 +27,15 @@ func getChartIconURL(t TemplateWrapper, templateVersionName string) (string, err
 	if err != nil {
 		return "", err
 	}
-	id := strings.Split(templateVersion.Name, "-")[1]
-	tag := templateVersion.Spec.Files[id+"/Chart.yaml"]
+	idStruct, err := url.Parse(templateVersion.Spec.ExternalID)
+	if err != nil {
+		return "", errors.WithMessage(err, "failed to parse external ID")
+	}
+	id := idStruct.Query().Get("template")
```
daxmc99: Compare with https://github.com/rancher/rancher/blob/0fce41539845482ef01acc9e9355bd21d089e6ce/pkg/controllers/user/helm/common/common.go#L51

dramich:
lgtm

PR #21982, opened 2019-08-05
troubleshoot issue with installing unittest plugin

drpebcak:


drpebcak:
I'm going to close this one. I don't think this will help. It's more likely that my PR against helm-unittest will help because we are hitting the github api rate limit.

Issue #19431, opened 2019-04-07
rancher removes wrong node when pool contains more than 10 nodes


contain-me-not:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
create node pool with 10 nodes
click the minus `-` on the UI to remove a node

**Result:**
node 9 is removed -- I expected node 10 to be removed

**Other details that may be helpful:**
appears to be a sorting issue where the last node, sorted alphanumerically, is removed instead of the last node number

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.1.8
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): vSphere
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM (any spec)
- Kubernetes version (use `kubectl version`): 1.12.7
- Docker version (use `docker version`): 18.06.1

tfiduccia:
Version - 2.2.x master
Was able to reproduce.  We should always remove the last node added if they select the -.

bmdepesa:
rancher/rancher:master-head-linux-amd64 a341755f

* Create a cluster
  * Worker nodepool has 9 nodes
* Scale up nodepool to 10 nodes
* After node provisions, scale down
* Most recently added node (10) is removed

PR #22020, opened 2019-08-06
Implement natural sort for nodepools

daxmc99:
Problem: Rancher would remove "node9" instead of "node10" when decreasing
nodepool size. This was due to the alphabetical sorting of nodepool names.

Solution: Implement natural sort of nodepool names.

Addresses: https://github.com/rancher/rancher/issues/19431


luthermonson:
LGTM

dramich:
lgtm

Issue #17231, opened 2018-12-26
It takes long time to show the pipeline config if there are many branches


loganhz:
Issues One:

It takes a long time to get the pipeline files for a repo if there are a lot of branches in the repo.

Issues Two:

If I have 100 branches in a repo, I can only see the first 50 branches in Rancher UI

Reason:

Rancher sends N + 1 API to github to get pipeline files. N is the number of the branches in the repo

Solution:

We need to provide an API for getting all branches. And an API to get pipeline file for a particular branch

loganhz:
cc @MrYuanZhen

gitlawr:
For issue 2, only 100 branches are listed in the reproduction.

jiaqiluo:
The bug is reproduced in Rancher:v2.2.7

When you click `Edit Config` in the pipeline 

![Screen Shot 2019-08-16 at 5 07 50 PM](https://user-images.githubusercontent.com/6218999/63204015-666b5800-c048-11e9-97b3-28ae293744aa.png)

Rancher send out a GET request to `https://ip/v3/projects/c-mr4nt:p-dg7mj/pipelines/p-dg7mj:p-xhmqj/configs` which take extremely long time to finish, in my test it takes 32.79s to load all branches 

<img width="1873" alt="Screen Shot 2019-08-16 at 5 10 13 PM" src="https://user-images.githubusercontent.com/6218999/63204036-c3ffa480-c048-11e9-81ce-5e5d60d75ca4.png">

jiaqiluo:
The bug fix is validated in Rancher:master-head

Doing the same test as the above comment describes, now Rancher sents out the GET request `/v3/project/<id>/pipelines/<id>/branches` to get all branches. And it takes much less time to get the response than in v2.2.7. In my test,
<img width="1874" alt="Screen Shot 2019-08-16 at 5 15 44 PM" src="https://user-images.githubusercontent.com/6218999/63204139-9c5d0c00-c049-11e9-9453-562678dcc52f.png">
 it takes **1.18s** to finish.

PR #22029, opened 2019-08-07
Fix fetching plenty of branches in pipeline

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/17231

Problem:
    1. Fails to list all branches when a pipeline repository has a large number of branches(>100).
    2. It takes a long time to fetch branches and pipeline files when there are plenty of branches.

Solution:
    1. Handle pagination.
    2. Update branches link to get all branches. So that we can reduce API calls from n+1 to 2, one for list branches and one for getting the pipeline file in a specific branch on demand.

### DIFF
--- a/pkg/pipeline/remote/bitbucketcloud/client.go
+++ b/pkg/pipeline/remote/bitbucketcloud/client.go
@@ -257,19 +257,21 @@ func (c *client) GetBranches(repoURL string, accessToken string) ([]string, erro
 		return nil, err
 	}
 
-	url := fmt.Sprintf("%s/repositories/%s/%s/refs/branches", apiEndpoint, owner, repo)
-
-	b, err := getFromBitbucket(url, accessToken)
-	if err != nil {
-		return nil, err
-	}
-	var branches PaginatedBranches
-	if err := json.Unmarshal(b, &branches); err != nil {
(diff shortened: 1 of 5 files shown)


orangedeng:
LGTM

pkg/pipeline/remote/bitbucketcloud/client.go
```diff
@@ -257,19 +257,21 @@ func (c *client) GetBranches(repoURL string, accessToken string) ([]string, erro
...
+		b, err := getFromBitbucket(nexturl, accessToken)
+		if err != nil {
+			return nil, err
+		}
+		var pageBranches PaginatedBranches
+		if err := json.Unmarshal(b, &pageBranches); err != nil {
+			return nil, err
+		}
+		for _, branch := range pageBranches.Values {
+			result = append(result, branch.Name)
```
cbron: Do you not need the `if b.Type != "BRANCH" ` check happening in Bitbucket server ?

gitlawr: It is a redundant check, given that the branches API always returns branch objects. Updated and removed the check.

gitlawr:
Updated

Issue #21714, opened 2019-07-23
k8s metadata: CNI template per k8s minor version support


alena1108:
In Rancher v2.0, v2.1, v2.2 CNI template can't diverge between k8s patch releases - v1.14.0 and v1.14.1 can't have different flannel version templates versions. It would be good to add support for it as a part of k8s metadata feature (ability to ship k8s w/o shipping Rancher). This way it would be possible to update flannel images along with the template for the latest patch release.

Required to fix https://github.com/rancher/rke/issues/1168 for k8s 1.13 and 1.14. Currently the new template has been made available to k8s 1.15 only.

jiaqiluo:
We used to define the `K8sVersionedTemplates` like this:
```
Calico: {
			"v1.15":   CalicoTemplateV115,
			"v1.14":   CalicoTemplateV113,
			"v1.13":   CalicoTemplateV113,
			"default": CalicoTemplateV112,
}
```

Now we change it to this:
https://github.com/rancher/kontainer-driver-metadata/blob/b813d099a1cdeb4c07fb0dc43b0c1d5ee871429a/rke/templates/templates.go#L43-L79
```
Calico: {
			">=1.16.0-alpha":         calicov116,
			">=1.15.0 <1.16.0-alpha": calicov115,
			">=1.13.0 <1.15.0": 	  calicov113,
			">=1.8.0 <1.13.0":  	  calicov18,
		},
```

This allows RKE/Rancher to pick the correct template depending on the k8s version.

For example, k8s `v1.16.0-beta.1-rancher1-1` will use the template `calicov116`, and k8s `v1.13.10-rancher1-1` will use the template `calicov113`

jiaqiluo:
reopen the bug. 

-------
Step 1: In the `kontainer-driver-metadata` project
- add a new k8s version in `rke/k8s_rke_system_images.go`, for example `v1.15.4`
- add a new canal template saying `CanalTemplateV154`
- edit `func LoadK8sVersionedTemplates()` and `func getTemplates() ` to add the new k8s and new template in `rke/templates/templates.go`  

Step 2:
 - run `Rancher:master-head` container
 - edit `v3/*` to use the above `kontainer-driver-metadata` repo, and refresh `v3/kontainerdrivers` to fetch new data
- add a cluster with k8s `v1.15.4` 
- wait for the cluster to be active 

Results:

[...]

PR #22046, opened 2019-08-07
driver metadata: use addon keys as reference on system images

kinarashah:
Each system image now has information on which addon template to fetch,
this is calculated on metadata refresh based on range information now
(instead of hardcoded k8s major versions). This will also allow providing
addon templates for minor versions.

Issue: 
https://github.com/rancher/rancher/issues/21714

Vendor PRs:
https://github.com/rancher/rke/pull/1530
https://github.com/rancher/kontainer-driver-metadata/pull/19


superseb:
With the changes on templates and having the ability to mark ranges now, I would prefer it to have all templates bound to a version (range), and default to error. This forces you to make sure all versions we cover have a template, and it will break when you add a new k8s version but not the templates. Also, the more variations are introduced, the more it becomes difficult to see what default gets applied to.

kinarashah:
> With the changes on templates and having the ability to mark ranges now, I would prefer it to have all templates bound to a version (range), and default to error. This forces you to make sure all versions we cover have a template, and it will break when you add a new k8s version but not the templates. Also, the more variations are introduced, the more it becomes difficult to see what default gets applied to.

https://github.com/rancher/rke/issues/1320

pkg/settings/setting.go
```diff
@@ -146,8 +146,8 @@ func NewSetting(name, def string) Setting {
 
 func getMetadataURL() string {
 	data := map[string]interface{}{
-		"url":    "https://github.com/rancher/kontainer-driver-metadata.git",
-		"branch": "master",
+		"url":    "https://github.com/kinarashah/kontainer-driver-metadata.git",
+		"branch": "tee",
```
prachidamle: this needs to be changed back?

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -45,41 +44,21 @@ func GetRKEWindowsSystemImages(k8sVersion string, sysImageLister v3.RKEK8sWindow
...
+	addon, err := addonLister.Get(namespace.GlobalNamespace, addonName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return "", err
 		}
-		if addon.Labels[sendRKELabel] == "false" {
-			return "", nil
+		addon, err = addons.GetNamespaced(namespace.GlobalNamespace, addonName, metav1.GetOptions{})
+		if !errors.IsNotFound(err) {
+			return "", err
```
prachidamle: We need to handle if err == errors.IsNotFound? Else we will hit panic below

pkg/controllers/management/clusterprovisioner/rke_store.go
```diff
@@ -1,48 +1,63 @@
...
 
 func (a *rkeStore) GetAddonTemplates(k8sVersion string) map[string]interface{} {
 	data := map[string]interface{}{}
-	for _, addonName := range addonsList {
-		template, err := kd.GetRKEAddonTemplate(k8sVersion, addonName, a.AddonLister, a.Addons)
+	sysImage, err := a.SystemImagesLister.Get(namespace.GlobalNamespace, k8sVersion)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			logrus.Errorf("getAddonTemplates: error finding system image for %s %v", k8sVersion, err)
+			return data
```
prachidamle: Don't we want to return the error?

kinarashah: this is getting called from kontainer engine to get data, in this case, we just log the error and move on with the assumption that either rke will have the template or rke will fail with proper msg

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
kinarashah: updated to return all errs if second Get fails

prachidamle:
LGTM

Issue #20881, opened 2019-06-14
[Feature Request] Azure China AKS Integration Support


pengjiang80:
AKS service is already launched in Azure China and some customers already use it. We need to support AKS service in Azure China, which includes:

1. Add China regions in location list of AKS creation page.
2. Call Azure China API endpoint when choosing China regions.
3. Test if there is any difference in API between Azure China and Global.

kingsd041:
Version: master master-head (8/29/19)
Passed the following test cases:
1. Ensure that the cluster can be successfully created in `China North 2` and `China East 2`region
2. Create a workload in the cluster and create a nodeport to ensure access to the workload via nodeport
3. Ensure the L4 Balancer and storage class are available
4. Modify the cluster

PR #22051, opened 2019-08-08
Add configurable endpoints to AKS meta handlers

gitlawr:
issue: https://github.com/rancher/rancher/issues/20881

Problem:
AKS meta handlers do not support talking to different cloud environment endpoints

Solution:
Add baseURL/authBaseURL params, like how we deal with those in [KE drivers](https://github.com/rancher/kontainer-engine/blob/7804cea8b58cd0c601c92617404092356a11fe2d/drivers/aks/aks_driver.go#L122)

### DIFF
--- a/server/capabilities/aks_version_endpoint.go
+++ b/server/capabilities/aks_version_endpoint.go
@@ -22,6 +22,10 @@ type AKSVersionHandler struct {
 }
 
 type regionCapabilitiesRequestBody struct {
+	// BaseURL specifies the Azure Resource management endpoint, it defaults "https://management.azure.com/".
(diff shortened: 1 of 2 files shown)


orangedeng:
LGTM

gitlawr:
For Azure cloud services, there are multiple cloud environments that use different endpoints, see [references](https://github.com/Azure/azure-cli/blob/468c654aa6500dd61c8c7dff8e741828c115c69b/src/azure-cli-core/azure/cli/core/cloud.py#L156-L243).
We make it configurable so as to keep consistent with AKS driver implementation.
UI will be responsible for passing these base URLs when specific regions are selected.

PR #22062, opened 2019-08-08
Fix pod scheduling problem in windows cluster

orangedeng:
**Problem:**
In windows preferred cluster, the pod create by linux node workload
won't be rescheduled immediately after we add toleration to the workload.

**Solution:**
1. Delete the pods which the pod doesn't have linux node toleration but
its owner workload has it.
2. In windows preferred cluster, we should set workload toleration in
api level. In this case, our controller won't need to add toleration again.

### DIFF
--- a/pkg/api/store/workload/toleration_test.go
+++ b/pkg/api/store/workload/toleration_test.go
@@ -1,0 +1,40 @@
+package workload
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSchedulingRuleFoundLinux(t *testing.T) {
+	type testcase struct {
+		name       string
+		rule       string
+		foundLinux bool
+	}
+	testcases := []testcase{
+		testcase{
+			name:       "rule operator equal",
+			rule:       "beta.kubernetes.io/os = linux",
+			foundLinux: true,
+		},
+		testcase{
+			name:       "rule operator not equal",
+			rule:       "beta.kubernetes.io/os != windows",
+			foundLinux: true,
+		},
+		testcase{
+			name:       "rule operator in",
+			rule:       "beta.kubernetes.io/os in (linux , darwin)",
+			foundLinux: true,
+		},
+		testcase{
+			name:       "rule key not match",
+			rule:       "test = test2",
+			foundLinux: false,
+		},
+	}
+	for _, c := range testcases {
+		assert.Equalf(t, c.foundLinux, schedulingRuleFoundLinux([]string{c.rule}), "case %s failed, function schedulingRuleFoundLinux is not as expected", c.name)
+	}
+}
--- a/pkg/api/store/workload/workload_store.go
+++ b/pkg/api/store/workload/workload_store.go
@@ -15,11 +15,13 @@ import (
 	"github.com/rancher/norman/types/values"
 	"github.com/rancher/rancher/pkg/api/customization/workload"
 	"github.com/rancher/rancher/pkg/clustermanager"
+	"github.com/rancher/rancher/pkg/taints"
 	managementschema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
 	"github.com/rancher/types/apis/project.cattle.io/v3/schema"
 	projectschema "github.com/rancher/types/apis/project.cattle.io/v3/schema"
 	managementv3 "github.com/rancher/types/client/management/v3"
 	projectclient "github.com/rancher/types/client/project/v3"
+	mapper "github.com/rancher/types/mapper"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 )
@@ -67,7 +69,9 @@ func (s *CustomizeStore) Create(apiContext *types.APIContext, schema *types.Sche
 	if err := setPorts(convert.ToString(data["name"]), data); err != nil {
 		return nil, err
 	}
-	setScheduling(apiContext, data)
+	if err := setScheduling(apiContext, data); err != nil {
+		return data, err
+	}
 	setStrategy(data)
 
 	err := s.validateStatefulSetVolume(schema, data)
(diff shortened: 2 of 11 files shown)


gitlawr:
We should investigate why the pod's spec is different from its owner workload. Is this an upstream issue?

orangedeng:
For those pods belong to statefulset, they won't get recreate if they are in `Unschedulable` condition and the owner statefulset already has the tolerates. So we still need to delete those pods without our tolerate configuration.

pkg/api/store/workload/workload_store.go
```diff
@@ -410,3 +433,82 @@ func imageUpdated(apiContext *types.APIContext, data map[string]interface{}) (bo
 	}
 	return false, nil
 }
+
+func getSchedulingRule(rule string) (string, string, string) {
```
gitlawr: Is it possible to extract and reuse [logic in scheduling mapper](https://github.com/rancher/types/blob/8c6adc8d2fb4c2d7c0f9b2709094b75f24e52005/mapper/scheduling.go#L148)? In case we miss some corner cases using a different parsing implementation

pkg/api/store/workload/workload_store.go
```diff
@@ -410,3 +433,82 @@ func imageUpdated(apiContext *types.APIContext, data map[string]interface{}) (bo
...
+	parts := strings.Split(rule, " ")
+	if len(parts) < 3 {
+		return "", "", ""
+	}
+	var key, operator, value string
+	key = parts[0]
+	i := 1
+	operator = parts[i]
+	i++
+	if operator == "not" {
```
gitlawr: We don't have the `not` operator, but a [notin](https://github.com/rancher/types/blob/master/mapper/scheduling.go#L175)

pkg/api/store/workload/workload_store.go
```diff
@@ -410,3 +433,82 @@ func imageUpdated(apiContext *types.APIContext, data map[string]interface{}) (bo
...
+		return false
+	}
+	switch operator {
+	case "=":
+		return strings.Contains(value, "linux")
+	case "in":
+		return strings.Contains(value, "linux")
+	case "!=":
+		return strings.Contains(value, "windows")
+	case "not in":
```
gitlawr: ditto

pkg/controllers/user/windows/pod_redeploy_controller.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	util "github.com/rancher/rancher/pkg/controllers/user/workload"
+	apicorev1 "github.com/rancher/types/apis/core/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type podRedeployController struct {
```
gitlawr: Can you add a description for the controller

pkg/api/store/workload/workload_store.go
```diff
@@ -150,7 +154,26 @@ func setScheduling(apiContext *types.APIContext, data map[string]interface{}) {
 		} else {
 			values.PutValue(data, "", "nodeId")
 		}
+
+		if requireRules, ok := values.GetStringSlice(data, "scheduling", "node", "requireAll"); ok {
```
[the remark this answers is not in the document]
orangedeng: Using requireAny seems like the user would have a second choice. I would prefer to use a resolute rule for it.

orangedeng:
Updated

orangedeng:
Updated. Need review again

pkg/api/store/workload/workload_store.go
```diff
@@ -150,7 +156,25 @@ func setScheduling(apiContext *types.APIContext, data map[string]interface{}) {
...
+			if schedulingRuleFoundLinux(requireRules) {
+				clusterName := clusterName(apiContext)
+				var cluster managementv3.Cluster
+				if err := access.ByID(apiContext, &managementschema.Version, "cluster", clusterName, &cluster); err != nil {
+					return err
+				}
+				// if !cluster.WindowsPreferedCluster || tolerationExists(data) {
+				// 	return nil
```
gitlawr: Fix the comment

gitlawr:
LGTM

pkg/api/store/workload/toleration_test.go
```diff
@@ -0,0 +1,40 @@
+package workload
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_getSchedulingRule(t *testing.T) {
```
rmweir: I think the function name should be pascal cased

pkg/api/store/workload/workload_store.go
```diff
@@ -410,3 +434,54 @@ func imageUpdated(apiContext *types.APIContext, data map[string]interface{}) (bo
 	}
 	return false, nil
 }
+
+func tolerationExists(workload map[string]interface{}) bool {
+	tolerations, ok := values.GetSlice(workload, "scheduling", "tolerate")
+	if !ok {
+		return false
+	}
+	for _, toleration := range tolerations {
```
rmweir: nit: I don't think you need to convert the below values before comparing them, it will just evaluate as false if they are not the same type.

orangedeng: If I don't convert the values below, I will need to use `reflect.DeepEqual` to do that, which is more slow way to compare.

pkg/api/store/workload/toleration_test.go
orangedeng: As my test target is `schedulingRuleFoundLinux` function, I rename it as `Test_schedulingRuleFoundLinux`.

orangedeng:
Updated.

pkg/api/store/workload/toleration_test.go
rmweir: I meant it should be of the form `TestSchedulingRuleFoundLinux`

pkg/api/store/workload/workload_store.go
daxmc99:
```suggestion
	for _, toleration := range tolerations {
	if toleration["key"] == taints.NodeTaint.Key &&
			toleration["effect"] == string(taints.NodeTaint.Effect) &&
			toleration["value"] == taints.NodeTaint.Value &&
			toleration["operator"] == string(corev1.TolerationOpEqual) {
			return true
```

orangedeng:
Addressed comments.

loganhz:
After discussions with Craig, we will use another way to fix this issue.

PR #22067, opened 2019-08-08
Rename ambiguous catalog 'utils' package to catutils

cbron:
This way the package name and searches or types on it will be unique.
Also see: https://blog.golang.org/package-names


dramich:
Seems like a lot of dead code in there based on this PR

cbron:
Looks like it. Want me to remove ?

cbron:
Going to remove dead code and rename everything to `version` since there aren't really any utils.

PR #22089, opened 2019-08-09
issue-18243 vendor only

cbron:
This is to run CI and make QA easier for: https://github.com/rancher/kontainer-engine/pull/180 - just the driver code. Will need to update vendor.conf before merging.


cbron:
Vendored separately.

Issue #21309, opened 2019-07-03
Export cluster produces an invalid and incomplete yaml


blackholegalaxy:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create a cluster on Azure using rancher UI.
- Run `rancher login` with appropriate token
 - Export cluster by running: `rancher clusters export CLUSTER`

**Result:**

Procuded yaml is incomplete. There is nothing useful here to be used to create a clone as stated here: https://rancher.com/docs/rancher/v2.x/en/cluster-admin/cloning-clusters/

```
Version: v3
clusters:
  product-dev-qa:
    dockerRootDir: /var/lib/docker
    enableNetworkPolicy: false
    localClusterAuthEndpoint: {}
```

**Environment information**
- Rancher version: 2.2.0
- Installation option (single install/HA): single install
- Rancher CLI version: 2.2.0
- cluster provider: Azure

We tested CLI on ubuntu and macOS.

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Azure AKS
- Machine type (cloud/VM/metal) and specifications (CPU/memory): cloud VM, 3 nodes, 2 cores/4GB each.
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.6", GitCommit:"ab91afd7062d4240e95e51ac00a18bd58fddd365", GitTreeState:"clean", BuildDate:"2019-02-26T12:59:46Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.9", GitCommit:"16236ce91790d4c75b79f6ce96841db1c843e7d2", GitTreeState:"clean", BuildDate:"2019-03-25T06:30:48Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```

blackholegalaxy:
@deniseschannon do you have any though on that?

deniseschannon:
@blackholegalaxy I'm wondering if something changed, it looks like the RKEConfig section isn't being exported. 

I'll see if I can have someone take a look.

dramich:
[...]

PR #22098, opened 2019-08-09
Only allow cluster export on RKE

dramich:
Problem:
Some clusters don't support being exported so an unusable config is
being returned

Solution:
Add an error message and remove export link from non-RKE clusters

Issue: https://github.com/rancher/rancher/issues/21309


luthermonson:
LGTM

Issue #21409, opened 2019-07-09
Google Auth - Errors messages can be made better for the user to read


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

1. Login to Rancher UI
2. Go to Global --> Authentication --> Google
3. Do not enter the domain name for the Admin EMail, just enter the name, and enter hostname, OAuth creds and Service account creds.
4. Click on Authenticate. Enter username and password. Click on Allow.

There is an error message seen on the UI

<img width="1823" alt="Screen Shot 2019-07-09 at 2 18 26 PM" src="https://user-images.githubusercontent.com/26032343/60927741-c4de3300-a25f-11e9-8817-e14d71e1757e.png">

When you provide invalid Admin email in step 3, and proceed to step 4. Users can see the below error message
<img width="1864" alt="Screen Shot 2019-07-09 at 2 17 25 PM" src="https://user-images.githubusercontent.com/26032343/60927782-f22ae100-a25f-11e9-8da1-4f27622d9751.png">

**Expected Result:**
User can be provided with better error messages on the UI.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher:master-head
- Installation option (single install/HA): single

sowmyav27:
@mrajashree When invalid service account creds are entered, and other input fields are valid inputs, this error message is seen.
`[Google OAuth] testAndApply: server error while authenticating: invalid character 's' after top-level value`
<img width="1595" alt="Screen Shot 2019-07-10 at 3 40 03 PM" src="https://user-images.githubusercontent.com/26032343/61009893-1ef6fb00-a329-11e9-9e8f-8dc1d36020d9.png">

mrajashree:
Putting the fix to provide better error message when service account creds are invalid. However for the admin email it's not possible to change the error message because the library we use returns only an error string, and at the point the call could have failed for multiple reasons so we can't say for sure it was because the admin email was wrong.

sowmyav27:
verified on master-head - commit id: `54432504`
On entering invalid service account creds, error seen on the UI
<img width="1502" alt="Screen Shot 2019-08-13 at 12 43 24 PM" src="https://user-images.githubusercontent.com/26032343/62972167-4463a280-bdc8-11e9-933b-dd3d6e48a7ae.png">

PR #22107, opened 2019-08-09
Provide better error msg for google auth

mrajashree:
https://github.com/rancher/rancher/issues/21409

Providing better error message for when service account credential is invalid.

The GH issue asks to do the same in case of admin email as well. But admin email is validated only when fetching user's groups during login. And in that case the golang admin library we use returns only an error string, and the error could have been due to multiple reasons at that point.

### DIFF
--- a/pkg/auth/providers/googleoauth/goauth_provider.go
+++ b/pkg/auth/providers/googleoauth/goauth_provider.go
@@ -401,12 +401,14 @@ func (g *googleOauthProvider) getDirectoryService(ctx context.Context, userEmail
 		// using JWTConfigFromJSON method
 		config, err := google.JWTConfigFromJSON(jsonCredentials, admin.AdminDirectoryUserReadonlyScope, admin.AdminDirectoryGroupReadonlyScope)
 		if err != nil {
-			return nil, err
+			logrus.Errorf("[Google OAuth] error unmarshaling service account creds: %v", err)
+			return nil, fmt.Errorf("invalid Service Account Credentials provided")
 		}
 		config.Subject = userEmail
 		srv, err := admin.NewService(ctx, option.WithTokenSource(config.TokenSource(ctx)))
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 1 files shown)


pkg/auth/providers/googleoauth/goauth_provider.go
```diff
@@ -401,12 +401,14 @@ func (g *googleOauthProvider) getDirectoryService(ctx context.Context, userEmail
 		// using JWTConfigFromJSON method
 		config, err := google.JWTConfigFromJSON(jsonCredentials, admin.AdminDirectoryUserReadonlyScope, admin.AdminDirectoryGroupReadonlyScope)
 		if err != nil {
-			return nil, err
+			logrus.Error(err)
+			return nil, fmt.Errorf("invalid Service Account Credentials provided")
```
kinarashah: seems like the func returns error for other reasons too? can we parse error to make sure?

mrajashree: `JWTConfigFromJSON` returns error either if parsing service account credentials fails, or if it's not of the type service account credential. So the reason is common for both that the service account creds are invalid.

pkg/auth/providers/googleoauth/goauth_provider.go
```diff
@@ -401,12 +401,14 @@ func (g *googleOauthProvider) getDirectoryService(ctx context.Context, userEmail
 		// using JWTConfigFromJSON method
 		config, err := google.JWTConfigFromJSON(jsonCredentials, admin.AdminDirectoryUserReadonlyScope, admin.AdminDirectoryGroupReadonlyScope)
 		if err != nil {
-			return nil, err
+			logrus.Error(err)
```
cjellick: The logged error isnt going to have any context to help people debugging the logs know where in the code it came from. Can you add a message that gives it context.

Same as below, but make sure the message is different for the two scenarios so that when we are looking at logs, we could distinguish between the two.

PR #22131, opened 2019-08-12
change type of ssh key fields

kinarashah:


pkg/controllers/management/drivers/nodedriver/machine_driver.go
```diff
@@ -105,6 +110,12 @@ func (m *Lifecycle) download(obj *v3.NodeDriver) (*v3.NodeDriver, error) {
 		}
 		userCredFields, pwdFields, defaults := getCredFields(obj.Annotations)
 		for name, field := range existingSchema.Spec.ResourceFields {
+			if _, ok := sshKeyFields[name]; ok {
+				if field.Type != "password" {
+					forceUpdate = true
```
dramich: Should this break here? If we hit this the driver needs an update

prachidamle:
lgtm

dramich:
lgtm

Issue #19589, opened 2019-04-12
Cluster Owner is not able to deploy catalog apps 


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
1. As an admin, create a cluster (rke DO).
2. Add authentication in Rancher example: Github.
3. Add another github user(say user1) as the cluster owner for the cluster created in step 1.
4. As an admin, create a project in the cluster.
5. As the github user - user1, deploy catalog app in project created in step 4.

**Expected Result:**
Cluster owner user should be able to create and deploy a catalog app.

**Actual Result:**
Cluster owner is not able to create a catalog app.
Error - `apps.project.cattle.io is forbidden: User "u-75s76v6b2h" cannot create resource "apps" in API group "project.cattle.io" in the namespace "p-hm6x5"
`

**Note:**
If the Cluster owner creates a Multicluster app in the project created in Step 4, He is not able to list the app under `Apps` but he is able to view the app under `Workloads`. 
Error in response - `{"baseType":"error","code":"Forbidden","message":"apps.project.cattle.io \"app03-p-b2sv5\" is forbidden: User \"u-75s76v6b2h\" cannot get resource \"apps\" in API group \"project.cattle.io\" in the namespace \"p-b2sv5\"","status":403,"type":"error"}` 

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher:v2.2.2-rc-x12
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): RKE cluster DO
- Kubernetes version (use `kubectl version`):

```
v1.13.5
```

- Docker version (use `docker version`):

```
18.9.4
```

sangeethah:
This issue is NOT reproducible in v2.1.8.
This issue is reproducible in v2.2.1.

mrajashree:
When a user is added to a cluster, we create a clusterroletemplatebinding, and the controller for crtb creates rolebindings for this user to grant access to them to project level resources in that cluster including apps, from [here](https://github.com/rancher/rancher/blob/release/v2.2/pkg/controllers/management/auth/crtb_handler.go#L155)
But in this case, the project is created in the cluster after the user is added as cluster-owner, and we don't seem to have the corresponding logic on project creation.

jiaqiluo:
[...]

Issue #15819, opened 2018-09-26
Cannot add members to a project when another cluster owner has created the project


tuxiem:
**Rancher versions:**
rancher version: v.2.0.8

**Problem:**
When another cluster owner has created a new project, without providing members. Other cluster owners, cannot add members. Denied with forbidden.

**Steps:**
(Only produced on AD joined cluster)
Person A and Person B is a member of the AD Group clusterAdmins
clusterAdmins is a member of cluster XYZ as role cluster owner
Person A creates a new project DEV in the cluster without adding any members
Person B, tries to add members to project DEV
Person B gets message "projectroletemplatebindings.management.cattle.io is forbidden: User "B" cannot create projectroletemplatebindings.management.cattle.io in the namespace "DEV"

**Results:**
Only the creator of the project is able to add members to the project.

**Expected:**
All cluster owners should have full access to projects within the cluster.

tfiduccia:
Version v2.2.0
I was able to reproduce this with the steps the user provided.  
1.  create cluster
2.  Add a group as owner with at least 2 users in it
3.  User 1 creates project
4.  User 2 tries to add to member to project user 1 created

This is only reproducible if the users are in a group.  If you add them individually it works fine.

tfiduccia:
Another issue related to group access in clusters:  I can't see the default namespace when i create a workload.  So:
1.  Create cluster
2.  Add group with 2 users as owner
3.  Log in as one of those users
4.  Create workload

The default namespace is not in the list of namespaces, and it should be.

cjellick:
> Another issue related to group access in clusters: I can't see the default namespace when i create a workload. So:

Unrelated. Can this be spun out to separate issue?

cjellick:
This is probably because we dont have a handler that is watching projects to trigger re-evaluating all CRTBs in that project's cluster. If you restarted rancher, it might fix itself bc that forces all CRTBs to be resynced (run through all handlers)

rmweir:
It is possible to repro this without using AD. (This worked fairly reliably for me but looked as though @dramich did not get the same behavior.) These were the steps followed:

1. create cluster
2. create 3 users: user1, user2, user3
3. separately add user1 and user2 as "Cluster Owner" to the cluster.
4. login to user1 and add a new project to the cluster
5. login to user2 and attempt to add user3 to the project as a member.

Result: should see same error as mentioned above

mrajashree:
[...]

PR #22133, opened 2019-08-12
Trigger crtb controller on project creation

mrajashree:
https://github.com/rancher/rancher/issues/19589
https://github.com/rancher/rancher/issues/15819

Problem: Users added as cluster-owners cannot create apps/add members in projects added to the
cluster after they have been made cluster-owners

Cause: The crtb controller lists all existing projects and creates rolebindings
in each project for the current crtb's user. So cluster-owners get access to resources
under `projectManagmentPlaneResources` for projects that exist already in the cluster when
they are being added. The same sync needs to happen on adding a new project to the cluster
as well which wasn't happening.

Fix: On project sync, enqueue clusterroletemplatebindings of that project's cluster.

### DIFF
--- a/pkg/controllers/management/auth/project_cluster_handler.go
+++ b/pkg/controllers/management/auth/project_cluster_handler.go
@@ -51,6 +51,7 @@ func newPandCLifecycles(management *config.ManagementContext) (*projectLifecycle
 		nsLister:           management.Core.Namespaces("").Controller().Lister(),
 		prtbLister:         management.Management.ProjectRoleTemplateBindings("").Controller().Lister(),
 		crtbLister:         management.Management.ClusterRoleTemplateBindings("").Controller().Lister(),
+		crtbClient:         management.Management.ClusterRoleTemplateBindings(""),
 		projectLister:      management.Management.Projects("").Controller().Lister(),
 		roleTemplateLister: management.Management.RoleTemplates("").Controller().Lister(),
 		clusterRoleClient:  management.RBAC.ClusterRoles(""),
@@ -69,7 +70,7 @@ type projectLifecycle struct {
 }
 
 func (l *projectLifecycle) sync(key string, orig *v3.Project) (runtime.Object, error) {
-	if orig == nil {
+	if orig == nil || orig.DeletionTimestamp != nil {
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -51,6 +51,7 @@ func newPandCLifecycles(management *config.ManagementContext) (*projectLifecycle
 		nsLister:           management.Core.Namespaces("").Controller().Lister(),
 		prtbLister:         management.Management.ProjectRoleTemplateBindings("").Controller().Lister(),
 		crtbLister:         management.Management.ClusterRoleTemplateBindings("").Controller().Lister(),
+		crtbs:              management.Management.ClusterRoleTemplateBindings(""),
```
rmweir: nit: maybe name this crtbClient to conform, or don't create a field at all and take it of l.mgr.mgt

pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -69,10 +70,14 @@ type projectLifecycle struct {
 }
 
 func (l *projectLifecycle) sync(key string, orig *v3.Project) (runtime.Object, error) {
-	if orig == nil {
+	if orig == nil || orig.DeletionTimestamp != nil {
 		return nil, nil
 	}
 
+	if err := l.enqueueCrtbs(orig); err != nil {
```
rmweir: do we want this to run before reconcileNamespace? Probably won't lead to an error but if there ends up being an error during reconcile we probably didn't need to enqueue, since the ns for it will not exist yet.

pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -100,6 +105,20 @@ func (l *projectLifecycle) sync(key string, orig *v3.Project) (runtime.Object, e
 	return nil, nil
 }
 
+func (l *projectLifecycle) enqueueCrtbs(project *v3.Project) error {
+	// get all crtbs in current project's cluster
+	clusterID := project.Namespace
+	crtbs, err := l.mgr.crtbLister.List(clusterID, labels.NewSelector())
```
rmweir: I think labels.Everything() is the proper selector here

pkg/controllers/management/auth/project_cluster_handler_test.go
```diff
@@ -0,0 +1,54 @@
+package auth
+
+import (
+	"testing"
+
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/apis/management.cattle.io/v3/fakes"
+	"github.com/stretchr/testify/assert"
```
rmweir: newline unnecessary

pkg/controllers/management/auth/project_cluster_handler.go
mrajashree: I added this because the test was initially initially in a different package where crtbInterface was added, and I can't call it in the test from l.mgr.mgt as I can't mock mgt. 
I think the name is fine since that's the convention we follow for all controllers?

pkg/controllers/management/auth/project_cluster_handler_test.go
mrajashree: Just added to organize imports, github vs k8s

pkg/controllers/management/auth/project_cluster_handler.go
mrajashree: makes sense, moved it

Issue #19133, opened 2019-03-25
[API] Inherited Roles can be deleted


n313893254:
**What kind of request is this (question/bug/enhancement/feature request):**

bug

It should fix in the API as well. https://github.com/rancher/rancher/issues/16890

PR #22158, opened 2019-08-13
Disable delete parent

rmweir:
Problem:
Roletemplate can be delete even if another roletemplate inherits from it.

Solution:
Check whether another roletemplate inherits from roletemplate being deleted and return error if so.

Issue:
https://github.com/rancher/rancher/issues/19133

### DIFF
--- a/pkg/api/customization/roletemplate/roletemplate.go
+++ b/pkg/api/customization/roletemplate/roletemplate.go
@@ -8,6 +8,8 @@ import (
 	"github.com/rancher/norman/types/convert"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	client "github.com/rancher/types/client/management/v3"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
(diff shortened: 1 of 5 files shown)


dramich:
This should also drop the delete link from the object

pkg/api/customization/roletemplate/roletemplate.go
```diff
@@ -49,3 +51,21 @@ func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data
 	}
 	return nil
 }
+
+func (w Wrapper) Formatter(apiContext *types.APIContext, resource *types.RawResource) {
+	roleTemplates, err := w.RoleTemplateLister.List("", labels.Everything())
+	if err != nil {
+		logrus.Warnf("Failed to list roletemplates. Error: %v", err)
```
dramich: Add a context here like `[roletemplate formatter]` so this can be traced back if needed

rmweir: oo good idea

dramich:
lgtm

Issue #21770, opened 2019-07-25
Upgrade rancher-logging chart api version to be k8s v1.16 compatible


guangbochen:
Need to update the Daemonset [api version](https://github.com/rancher/system-charts/blob/dev/charts/rancher-logging/0.1.1/charts/log-aggregator/templates/log-aggregator.yaml#L1) to `apps/v1` of the `rancher-logging v0.1.1` chart ,  since the `extensions/v1beta1` is deprecated in k8s v1.16.
 
Refer to [this helps.tpl](https://github.com/rancher/system-charts/blob/dev/charts/rancher-monitoring/v0.0.4/templates/_helpers.tpl#L76) as a good example.

bmdepesa:
rancher/rancher:master-head 41ce3ac49

* Deployed Rancher master-head
* Created 1.16.0 cluster
* Enabled cluster logging to Fluentd
* Performed some operations (create workloads / ingress)
* `cattle-logging` app version is `0.1.2`
* Log statements are transported correctly
* Daemonset has been updated to use new version
```
> kubectl get daemonset rancher-logging-fluentd-linux -n cattle-logging -o yaml
apiVersion: apps/v1
```

PR #22166, opened 2019-08-14
Add force upgrade for system apps

aiwantaozi:
Problem:
After version v1 release for deployment, statefulset, daemonset, there is some compatibility issue, like updates to deployment/statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden. So upgrade chart may get failed.

Solution:
Add force upgrade for system app

Issue:
https://github.com/rancher/rancher/issues/21770

Related Charts PR:
https://github.com/rancher/system-charts/pull/78


orangedeng:
LGTM

PR #22184, opened 2019-08-14
Label tokens

cbron:
**Depends on:** Types PR: https://github.com/rancher/types/pull/940

Adds a label to Rancher management created tokens to help categorize them. Examples: 

```
Helm token: "authn.management.cattle.io/kind": "helm",
Telemetry token: "authn.management.cattle.io/kind": "telemetry",
User login: "authn.management.cattle.io/kind": "session",
```

Note
* Won't add labels to user (custom) tokens. 
* Won't add labels for existing tokens.

### DIFF
--- a/pkg/auth/tokens/manager.go
+++ b/pkg/auth/tokens/manager.go
@@ -31,6 +31,7 @@ import (
 const (
 	userPrincipalIndex = "authn.management.cattle.io/user-principal-index"
 	UserIDLabel        = "authn.management.cattle.io/token-userId"
+	TokenKindLabel     = "authn.management.cattle.io/kind"
 	tokenKeyIndex      = "authn.management.cattle.io/token-key-index"
 	secretNameEnding   = "-secret"
 	secretNamespace    = "cattle-system"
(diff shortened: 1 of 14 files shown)


vendor.conf
```diff
@@ -43,7 +43,7 @@ github.com/rancher/norman                     043a1c919df3ada12d5592e7f0259404cb
 github.com/rancher/kontainer-engine           e5962e748ba4c0c020b45ed8ba050e912cc358ae
 github.com/rancher/rke                        66f7c26d396a841cd1f13fd19f20b84a9ba635bc
 github.com/rancher/kontainer-driver-metadata  821544f898e7bcbc0a475d2f0afccc6c272f691c
-github.com/rancher/types                      9097b6038b9ff351fdd2a699465b67c6c18e1168
+github.com/rancher/types                      e79b1fb936f48dac9ee6240a80d352d778508f81 https://github.com/cbron/types
```
cbron: Temporary

pkg/auth/tokens/manager.go
```diff
@@ -122,16 +123,14 @@ func (m *Manager) createToken(k8sToken *v3.Token) (v3.Token, error) {
 		return v3.Token{}, fmt.Errorf("failed to generate token key")
 	}
 
-	labels := make(map[string]string)
-	labels[UserIDLabel] = k8sToken.UserID
-
+	if len(k8sToken.ObjectMeta.Labels) == 0 {
```
daxmc99: 👍

tests/integration/suite/test_auth_configs.py
```diff
@@ -110,3 +111,12 @@ def key_secret_creation(k8sclient):
         if s.metadata.name == "pingconfig-spkey":
             return True
     return False
+
+
+def test_auth_label():
+    r = requests.post(AUTH_URL, json={
```
dramich: Let's not do this again, one it's hardcoding a password (again) and two the token most likely already exists so if you run this test against an existing setup it can fail. Create a new user and fetch their token explicitly.

pkg/auth/tokens/manager.go
dramich: This should be a nil check

tests/integration/suite/test_auth_configs.py
cbron: Fixed now. Also we aren't going to backfill existing tokens.

tests/integration/suite/test_auth_configs.py
```diff
@@ -110,3 +110,18 @@ def key_secret_creation(k8sclient):
         if s.metadata.name == "pingconfig-spkey":
             return True
     return False
+
+
+def test_auth_label(admin_mc, user_factory):
+    user = user_factory()
+    k8s_client = CustomObjectsApi(admin_mc.k8s_client)
+    tokens = k8s_client.list_cluster_custom_object(
```
dramich: There should be a wait_for in here to ensure we actually get the users token before attempting to access it.

cbron: Fixed now. Also re-vendored types, which included another change but it looks legit.

dramich:
lgtm

PR #22186, opened 2019-08-14
Hide clustertemplate answers on clusterspec 

prachidamle:
https://github.com/rancher/rancher/issues/21519

Answers to Questions set on ClusterTemplate are stored on the Cluster Spec when user creates the cluster with answers. If the properties set as answers are password/secret fields, we should not display such answers on API.

### DIFF
--- a/pkg/api/customization/cluster/formatter.go
+++ b/pkg/api/customization/cluster/formatter.go
@@ -7,11 +7,23 @@ import (
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	managementschema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
+	client "github.com/rancher/types/client/management/v3"
+	"github.com/rancher/types/config"
 	"github.com/sirupsen/logrus"
 )
 
 type Formatter struct {
 	KontainerDriverLister v3.KontainerDriverLister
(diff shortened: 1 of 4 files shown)


pkg/api/customization/cluster/formatter.go
```diff
@@ -71,6 +83,26 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 		setTrueIfNil(configMap, "associateWorkerNodePublicIp")
 		setIntIfNil(configMap, "nodeVolumeSize", 20)
 	}
+
+	if clusterTemplateAnswers, ok := resource.Values["answers"]; ok {
```
kinarashah: nit: we should move this logic to a separate func for templated related changes

tests/integration/suite/test_clustertemplate.py
```diff
@@ -479,6 +479,77 @@ def test_required_template_question(admin_mc, remove_resource):
...
+                  "default": "abcdClientId"
+                 },
+                 {
+                  "variable": azureClientSecret,
+                  "required": "true",
+                  "type": "string",
+                  "default": ""
+                 }]
+
+    rev = client.create_cluster_template_revision(clusterConfig=cconfig,
```
kinarashah: we prolly need remove_resource for template and revision too?

prachidamle: create_cluster_template will do the remove_resource on the template, which in turn deletes all the revisions.

Issue #21900, opened 2019-08-01
Java stack log multi-line matching collection


xiaoluhong:
**What kind of request is this (question/bug/enhancement/feature request):**

**enhancement**

**Steps to reproduce (least amount of steps as possible):**

In the current version, rancher-logging collects log information by line. In applications like Java, stack logs are split into multiple lines of output. So, what eventually gets sent to elasticsearch store and presented via Kibana are multiple logs.

<img width="1066" alt="image" src="https://user-images.githubusercontent.com/20707603/62260402-6cf0b180-b444-11e9-9fc0-94a374adc867.png">

<img width="946" alt="image" src="https://user-images.githubusercontent.com/20707603/62260411-7548ec80-b444-11e9-9649-e721e3547587.png">

gz#8786

gz#12614

gz#12648

gz#14981

rmalchow:
hi xiaohulong,

you might want to look into using some kind of JSON encoder (we use the logstash encoder) for your output. this encodes a single log object as a single-line java object. many log shipping tools can then handle it correctly down the line. other than that, this is a docker issue, not a rancher issue.

what we do to solve this: 

1.) add the logstash encoder
```
		<dependency>
			<groupId>net.logstash.logback</groupId>
			<artifactId>logstash-logback-encoder</artifactId>
		</dependency>
```
2.) let filebeat pick up the logs, and tell it to interpret the "log" field as a nested json object:

filebeat.yml:
```
          config:
            - type: docker
              containers.ids:
                - "${data.docker.container.id}"
              fields_under_root: true
              json.message_key: log
```

3. profit:

`{ "timestamp" : "3838849551" , "log" : "{ \"level\" : \"LOG\", \"message\" : \"foobar\", \"stacktrace\" : [ \"one\", \"two\", \"three\" ] }" } 
`
becomes:

`{ "timestamp" : "3838849551", "level" : "LOG", "message" : "foobar", "stacktrace" : [ "one", "two", "three" ] }`

with all the data that yo uhave in your initial log object neatly represented

aiwantaozi:
@loganhz this is an internal issue, should it be included in 2.3

ifeldshteyn:
In terms of handling multiline in rancher, check out this post.

deniseschannon:
@aiwantaozi @loganhz 

[...]

PR #22196, opened 2019-08-15
Add support for multiline logging

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/21900

Add a filter block for the concat plugin, do the validation accordingly.
Update `escapeString()` to address backslash issue.


pkg/controllers/user/logging/generator/validator.go
```diff
@@ -124,16 +127,5 @@ func filterRubyCode(s string) error {
...
-		"\r", `\\r`,
-		"\f", `\\f`,
-		"\b", `\\b`,
-		"\"", `\\\"`,
-		"\\", `\\\\`,
-	)
-
-	escapeString := escapeReplacer.Replace(postDoc)
-	return fmt.Sprintf(`"%s"`, escapeString)
+	return fmt.Sprintf("%q", postDoc)
```
aiwantaozi: This convert is the simulation from Fluentd. Base on the doc https://docs.fluentd.org/configuration/config-file#in-double-quoted-string-literal-is-escape-character and the test case https://github.com/fluent/fluentd/blob/v1.6.3/test/config/test_literal_parser.rb

For example, original user input is `firstline\nsecondline`, and we parse to `firstline\\nsecondline`, fluentd will parse back to `firstline\nsecondline`

pkg/controllers/user/logging/generator/templatefilter.go
```diff
@@ -36,6 +36,25 @@ var FilterTemplate = `
 </filter>
 {{end}}
 
+{{define "filter-concat"}}
+{{- if .EnableMultiLineFilter}}
+<filter {{ .ContainerLogSourceTag }}.**>
+  @type concat
+  key log
+  stream_identity_key tag
```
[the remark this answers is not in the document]
gitlawr: concat multiline logs using the same tag, i.e., aggregate those from the same container

pkg/controllers/user/logging/generator/validator.go
gitlawr: https://github.com/rancher/rancher/blob/85dedb19bbdfabc5e81a8258b868cc1ca8261b0d/pkg/controllers/user/logging/generator/validator.go#L128
Take the above as an example.
Please note that the backslash inside backquote in Golang is not the escape character. so `\\t` is literally two backslashes and a `t`. When it is quoted as `"\\t"` in following codes, it becomes an escaped backslash and a `t` from the fluentd parser's view. And it does not match the original "\t" that stands for a single horizontal tab.

aiwantaozi:
need rebase.

Issue #21720, opened 2019-07-23
[API] Cluster Templates - Member is able to reduce Access Type of other users on a cluster template


davidnuzik:
Version: master-head (7/23/19)

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Create a cluster template, share it with another user, make them a member.
- Share it with another member, make them an owner.
- Log in as the _member_ user, edit the cluster template.
- Notice, we can successfully change permissions on the owner user. We can limit their access to just Read-Only or remove them if we want.
- View in API - it indeed affected the clustertempalterevision.

**Result:**
Member can elevate access and change membership for any user including those with higher access.

mrajashree:
@davidnuzik @deniseschannon @alena1108 I don't think this is a bug. If the user was given "member" access type, they have the "update" permission on the clusterTemplate resource. So they can update the Members list on the clusterTemplate resource too. Only "read-only" members won't be able to change the access type of other members.

deniseschannon:
@mrajashree After discussing, we have agreed that this is the permission logic for members:

Ability to update templates, but explicitly exclude the ability to update membership of the template as that would provide potential privilege escalation (able to assign yourself owner permissions).

deniseschannon:
After further discussion, we have decided that the "member" role in cluster templates doesn't make sense. We will be removing this role and keep the following 2 roles:

1. Owner - Able to CRUD templates/revisions and share CT to other people (Note: a CT owner will also need to have the global `Create Cluster Template` role in order for it to make sense, and only admins can assign this role.)
2. Read-Only - People who will be dedicated to use the cluster templates. For better UX, the UI will be updated to `User` for when assigning this role. https://github.com/rancher/rancher/issues/22200

davidnuzik:
Version: master-head (v2.3) (8/19/19)

I can confirm now that we only have two access types to chose from in the UI when we are sharing a cluster template:

![image](https://user-images.githubusercontent.com/45179589/63308060-9e230b80-c2a5-11e9-893e-8db65a33316d.png)

I confirmed that the **Owner** member is able to create, read, update, delete the cluster template (assuming has the required `Create Cluster Template` global permission to create).

I confirmed that the **User** member is NOT able to create, update, or delete the cluster template (they can only read and leverage a cluster template to create a cluster).

[...]

PR #22203, opened 2019-08-15
Remove member access type for cluster templates

mrajashree:
https://github.com/rancher/rancher/issues/21720#issuecomment-521759473

### DIFF
--- a/tests/integration/suite/test_clustertemplate.py
+++ b/tests/integration/suite/test_clustertemplate.py
@@ -213,7 +213,8 @@ def test_create_cluster_template_with_members(admin_mc, remove_resource,
 
     # add * as member to share with all
     new_members = [{"userPrincipalId": "local://" + user_member.user.id,
-                    "accessType": "read-only"}, {"groupPrincipalId": "*"}]
+                    "accessType": "read-only"}, {"groupPrincipalId": "*",
+                                                 "accessType": "read-only"}]
     client.update(ct, members=new_members)
 
     split = cluster_template.id.split(":")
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_clustertemplate.py
```diff
@@ -309,16 +310,16 @@ def test_revision_creation_permission(admin_mc, remove_resource,
...
              fail_handler=fail_handler(rb_resource))
-    rb_name = name + "-ct-u"
+    rb_name = name + "-ct-a"
     wait_for(lambda: check_subject_in_rb(rbac, 'cattle-global-data',
-                                         user_member.user.id, rb_name),
+                                         user_owner.user.id, rb_name),
              timeout=60,
              fail_handler=fail_handler(rb_resource))
     templateId = cluster_template.id
     # user with accessType=member should not be able to create revision
```
prachidamle: nit: correction to the comment needed `accessType=owner`

tests/integration/suite/test_clustertemplate.py
```diff
@@ -331,9 +332,9 @@ def test_revision_creation_permission(admin_mc, remove_resource,
     # assign 'clustertemplates-create' role to the user with accessType=member
```
prachidamle: nit: correction to the comment needed `accessType=owner`

prachidamle:
LGTM

alena1108:
@mrajashree can we drop the member type from the api, so it's not available in schema?

mrajashree:
@alena1108 looks like we cannot remove "member" in schema, since that is an option in the enum for AccessType field from Members struct. This Members is used by MCapp and gdns as well and would affect them too.

Issue #20191, opened 2019-05-13
Use gitlab to configure cicd authentication, prompting net/ HTTP: request canceled (client.timeout while reading body)


xiaoluhong:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Steps to reproduce (least amount of steps as possible):**
1. Gitlab version  is 11.3.6，the repository has six pages, totaling nearly 600 repositories. 
2. The new gitlab account can be authorized normally.
3. When using root account for authorization, prompting net/ HTTP: request canceled (client.timeout while reading body)

**Result:**

Use gitlab to configure cicd authentication, prompting net/ HTTP: request canceled (client.timeout while reading body)

<img width="396" alt="image" src="https://user-images.githubusercontent.com/20707603/57605109-7599c000-7598-11e9-9f76-15959429dfec.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version:
`rancher/rancher:v2.2.2`

- Installation option :
HA

**Cluster information**
- Cluster type:
Custom

- Machine type:
VM

- Kubernetes version (use `kubectl version`):
v1.13.5
```

xiaoluhong:
@gitlawr

gitlawr:
Reproducible by following steps:
1. Set up a 4c16g Gitlab instance
2. Import 400+ repositories from Github using my access token.
3. Integrate with the Gitlab instance in tools-pipeline and get the error.

jiaqiluo:
The bug is reproduced in Rancher:v2.2.7

Steps:
- run Rancher:v2.2.7 single install
- add a cluster
- under the default project, go to tools -> pipeline
- add a Gitlab account
- wait for authentication to be finished 

Results:

<img width="1443" alt="Screen Shot 2019-08-20 at 7 46 27 PM" src="https://user-images.githubusercontent.com/6218999/63398957-021c0180-c384-11e9-91de-0c4934caf52d.png">
<img width="1321" alt="Screen Shot 2019-08-20 at 7 51 12 PM" src="https://user-images.githubusercontent.com/6218999/63398958-021c0180-c384-11e9-8514-b304670297ff.png">

jiaqiluo:
The bug fix is validated in Rancher: master-head 

Following the same steps as the previous comment, this time Rancehr loads all 300 repos from Gitlab successfully 

[...]

PR #22222, opened 2019-08-16
Fix pipeline repos timeout

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/20191

Problem:
Fetching repos results in HTTP client timeout when the number is large. The cause is that paginated responses are held and not proceeded until all of them are available.

Solution:
Proceed paginated response body early.
This pr also update the HTTP client timeout to 30s to have proper tolerance to high latency network.

### DIFF
--- a/pkg/pipeline/remote/gitlab/gitlab.go
+++ b/pkg/pipeline/remote/gitlab/gitlab.go
@@ -293,18 +293,13 @@ func (c *client) GetBranches(repoURL string, accessToken string) ([]string, erro
 	}
 	url := fmt.Sprintf(c.API+"/projects/%s/repository/branches", project)
 
-	responses, err := paginateGitlab(accessToken, url)
+	responseBodies, err := paginateGitlab(accessToken, url)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/pipeline/remote/gitlab/gitlab.go
```diff
@@ -474,25 +469,24 @@ func doRequestToGitlab(method string, url string, gitlabAccessToken string, opt
...
+		if err != nil {
+			return nil, err
+		}
+		body, err := ioutil.ReadAll(response.Body)
 		if err != nil {
-			return responses, err
+			return nil, err
 		}
-		responses = append(responses, response)
+		response.Body.Close()
```
kinarashah: should we add `defer response.Body.Close()` before `ReadAll` in case that returns an err?

gitlawr: It is in a for loop, therefore `Close()` is called directly instead of using defer. Updated to call `Close()` on `readAll` errors.

PR #22238, opened 2019-08-16
Update kontainer-engine

rmweir:
Updated kontainer-engine which now uses proper default k8s version for EKS clusters.


dramich:
lgtm

PR #22241, opened 2019-08-16
[2.2] Remove stuck Global Role bindings and RoleTemplates

daxmc99:
Problem: In the cluster scoped finalizer removal refactor some finalizers
    were not cleaned up. This was due to way they were being added without
    calling AddClusterScoped Finalizer

Solution: Write legacy sync handlers to remove orphaned resources and
    update norman code generation to add cluster scoped handlers to the
    resource map in the LifecycleAdapter function instead of inside controllers.
    This essentially moves the resource map add further up the call.

_Dependencies_:
https://github.com/rancher/norman/pull/296
https://github.com/rancher/types/pull/946

### DIFF
--- a/pkg/controllers/user/rbac/legacy_grb_cleaner.go
+++ b/pkg/controllers/user/rbac/legacy_grb_cleaner.go
@@ -1,0 +1,63 @@
+package rbac
+
+import (
+	"strings"
+	"time"
+
+	"github.com/rancher/norman/lifecycle"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type grbCleaner struct {
+	m *manager
+}
+
+func newLegacyGRBCleaner(m *manager) *grbCleaner {
+	return &grbCleaner{m: m}
+}
+
+func (p *grbCleaner) sync(key string, obj *v3.GlobalRoleBinding) (runtime.Object, error) {
+	if key == "" || obj == nil {
+		return nil, nil
+	}
+	if obj.DeletionTimestamp == nil || len(obj.Finalizers) == 0 {
+		return nil, nil
+	}
+	if time.Since(obj.DeletionTimestamp.Time) < 1*time.Hour {
+		return nil, nil
+	}
+	obj, err := p.removeStuckFinalizer(obj)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	}
+	return obj, nil
+
+}
+
+func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.GlobalRoleBinding, error) {
+	obj = obj.DeepCopy()
+	modified := false
+	md, err := meta.Accessor(obj)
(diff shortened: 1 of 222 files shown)


pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -0,0 +1,60 @@
+package rbac
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/errors"
```
rmweir: unneeded blank line I think

pkg/controllers/user/rbac/legacy_rt_cleaner.go
```diff
@@ -0,0 +1,60 @@
+package rbac
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/errors"
```
rmweir: same as above

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -0,0 +1,60 @@
...
+
+type grbCleaner struct {
+	m *manager
+}
+
+func newLegacyGRBCleaner(m *manager) *grbCleaner {
+	return &grbCleaner{m: m}
+}
+
+func (p *grbCleaner) sync(key string, obj *v3.GlobalRoleBinding) (runtime.Object, error) {
```
rmweir: nit: I feel like some of this logic could be repackaged and reused for both, otherwise looks fine

dramich:
Per discussion this needs a way to know if a finalizer is stale, suggested time is after 1 hour then take action.

pkg/controllers/user/rbac/legacy_rt_cleaner.go
daxmc99: Fixed thanks

pkg/controllers/user/rbac/legacy_grb_cleaner.go
daxmc99: While I think there is some repetition here I don't think its enough to warrant attempting to reduce the overlap. While the logic is similar the signatures provide some helpful type information.

pkg/controllers/user/rbac/legacy_rt_cleaner.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (p *rtCleaner) sync(key string, obj *v3.RoleTemplate) (runtime.Object, error) {
+	if key == "" || obj == nil {
+		return nil, nil
+	}
+	if obj.DeletionTimestamp == nil || len(obj.Finalizers) == 0 {
+		return nil, nil
+	}
+	if time.Since(obj.DeletionTimestamp.Time) < 1*time.Hour {
```
daxmc99: @dramich

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -0,0 +1,63 @@
...
+	if key == "" || obj == nil {
+		return nil, nil
+	}
+	if obj.DeletionTimestamp == nil || len(obj.Finalizers) == 0 {
+		return nil, nil
+	}
+	if time.Since(obj.DeletionTimestamp.Time) < 1*time.Hour {
```
daxmc99: @dramich

Issue #21598, opened 2019-07-17
k8s metadata - mismatch of default k8s version in UI and API


jiaqiluo:
Bug

Steps:
- run testing container Rancher:v2.3.3
- Edit the metadata to set the default k8s version for v2.3.3
```
func loadRancherDefaultK8sVersions() map[string]string {
    return map[string]string{
        ...
        "2.3.3": "v1.12.4-rancher1-1",
		...
        // rancher will use default if its version is absent
        "default": "v1.14.3-rancher1-1",
    }
}
```
- push the changes to GitHub repp
- refresh the `v3/kontainerdrivers` in Rancher v2.3.3

Result:
- in the UI, the default k8s version is `v1.12.9-rancher1-1`

More info:
- the value of `/v3/settings/ui-k8s-default-version-range` is `<=v1.12.x`, which indicates Rancher fetches the highest minor version as the default in UI
- the value of `/v3/settings/k8s-version` is `v1.12.4-rancher1-1`, which indicates the default k8s version will be the expected one if provisioning a cluster without providing a k8s version via API 

Expected:
- Option1: both UI and API should use the specific version from the metadata
- Option2: change the design of metadata such that it sets only the major k8s version, not the specific minor version, for the default k8s version.

jiaqiluo:
The bug fix is validated in rancher-master 8/23 with customized image 

Follow the same steps as the issue describes, and set the default k8s version for `rancher:v2.3.4` as `v1.13.1-rancher1-1`

Results:
- the value of `ui-k8s-default-version-range` == the value of `k8s-version` == `v1.13.10-rancher1-3`

PR #22261, opened 2019-08-18
make default k8s value same as UI 

kinarashah:
https://github.com/rancher/rancher/issues/21598
https://github.com/rancher/kontainer-driver-metadata/pull/24

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/data.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/data.go
@@ -489,9 +489,22 @@ func SaveSettings(k8sCurrVersions map[string]interface{},
 	if err := settings.KubernetesVersionToServiceOptions.Set(k8sSvcOptionData); err != nil {
 		return err
 	}
-	defaultK8sVersion, ok := rancherDefaultK8sVersions[rancherVersion]
-	if !ok || defaultK8sVersion == "" {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -450,9 +450,22 @@ func SaveSettings(k8sCurrVersions map[string]interface{},
...
-	if !ok || defaultK8sVersion == "" {
-		defaultK8sVersion = rancherDefaultK8sVersions["default"]
+	defaultK8sVersionRange, ok := rancherDefaultK8sVersions[rancherVersion]
+	if !ok || defaultK8sVersionRange == "" {
+		defaultK8sVersionRange = rancherDefaultK8sVersions["default"]
+	}
+	// get matching default k8s from k8s curr
+	toMatch := util.GetTagMajorVersion(defaultK8sVersionRange)
+	defaultK8sVersion := ""
+	for k8sCurr := range k8sCurrVersions {
```
prachidamle: does UI pick the default from k8sCurrVersions or from KubernetesVersionToSystemImages?

kinarashah: I believe the latter, but both have the same values

prachidamle:
LGTM

Issue #22057, opened 2019-08-08
Ability to use custom Environment Variables in pipeline's Publish Catalog Template step


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Enhancement

**Steps to reproduce (least amount of steps as possible):**

Currently it is not possible to reference custom environment variables in the pipeline's Publish Catalog Template step.

The UI does enable configuration of environment variables on this step, but these cannot be successfully referenced, a separate issue is open for that in https://github.com/rancher/rancher/issues/22031.

User would like the ability to reference custom environment variables in the pipeline's Publish Catalog Template step, in the same way they can use the pre-defined environment variables https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/pipelines/#pipeline-variable-substitution-reference

**Environment information**
- Rancher v2.2.4

eliaoggian:
To add a specific example, the **.rancher-pipeline.yml** file

This works:
```
stages:
- name: test
  steps:
  - runScriptConfig:
      image: busybox
      shellScript: echo $TEST
    env:
      TEST: ${CICD_GIT_URL}
```

This does not work:
```
- name: publish catalog tag
  steps:
  - publishCatalogConfig:
      path: ./chart
      catalogTemplate: app
      version: ${CICD_GIT_TAG}
      gitUrl: http://user@git.example.com/scm/ran/helm-catalog.git
      gitBranch: master
      gitAuthor: jenkins
      gitEmail: jenkins@example.com
    env:
      APP_VERSION: ${CICD_GIT_TAG}
      CATALOG_VERSION: ${CICD_GIT_TAG}
    envFrom:
    - sourceName: bitbucket-credentials
      sourceKey: USERNAME
      targetKey: USERNAME
    - sourceName: bitbucket-credentials
      sourceKey: PASSWORD
      targetKey: PASSWORD
```

I think that being able to set these environment vars would improve the customization level of the pipelines, and give more control to the customers.

gitlawr:
@eliaoggian 
A quick question for confirmation:
Regarding your example, do you expect to use `APP_VERSION /CATALOG_VERSION` in the chart files/templates?

eliaoggian:
Yes! I would like to use them to set different version of Chart.yaml using the same template.

gitlawr:
[...]

PR #22267, opened 2019-08-19
Support custom envvar substitution in pipeline publishcatalog steps

gitlawr:
Related issue: https://github.com/rancher/rancher/issues/22057

Depend on: https://github.com/rancher/types/pull/950 https://github.com/rancher/pipeline-tools/pull/16


orangedeng:
LGTM

gitlawr:
@cloudnautique There is no difference between those events. Do you mean we should handle them in different ways?

Issue #22231, opened 2019-08-16
v2.1.x project namespace gets stuck in terminating if they had alerts


cjellick:
Steps are somewhat theoretical, but I've seen it in a user's setup:

1. Create a project and project alerts in v2.1
2. Upgrade to v2.2
3. Delete the project
4. Using kubectl against the rancher control plane kubernetes, do `kubectl get ns`

You'll see that the namespace corresponding to the project you deleted is stuck in terminating.
If you do `kubectl -n <project id> get projectalerts -o yaml` you should see at least one alert that is stuck in deleting with a finalizer that looks like `clusterscoped.controller.cattle.io/project-alert-podtarget_c-123456`.

This is because the lifecycle for project alerts was dropped here https://github.com/rancher/rancher/blame/master/pkg/controllers/user/alert/watcher/pod.go#L65

Probably should backport fix to v2.2. Please check to see if any other lifecycles were dropped in that PR.

sowmyav27:
Validated the issue - upgraded from `rancher:v2.1.13` to `v2.3.0-alpha7`.

- Create a project and project alerts in v2.1.13
- Upgrade to v2.3.0-alpha7
- Delete the project
- Using kubectl against the rancher control plane kubernetes, do kubectl get ns
- namespace corresponding to the project deleted is stuck in terminating.
- do kubectl -n <project id> get projectalerts -o yaml, alerts are available in `Active` status. They are NOT in deleting status.

Validated the fix. Validated the issue - upgraded from `rancher:v2.1.13` to `master-head` (latest).
- Create a project and project alerts in v2.1.13
- Upgrade to master
- Delete the project
- Using kubectl against the rancher control plane kubernetes, do kubectl get ns
- namespace corresponding to the project is not available and has been deleted.

PR #22268, opened 2019-08-19
Removed legacy finalizer when upgrade alert

aiwantaozi:
Problem:
Because we change controller name since 2.2, when delete project included legacy projectAlert obj, it will stuck in getting legacy finalizer

Solution:
Remove legacy finalizer when upgrade alert

Issue:
https://github.com/rancher/rancher/issues/22231

### DIFF
--- a/pkg/controllers/user/alert/deployer/upgradeimpl.go
+++ b/pkg/controllers/user/alert/deployer/upgradeimpl.go
@@ -123,6 +123,11 @@ func (l *AlertService) Upgrade(currentVersion string) (string, error) {
 		}
 	}
 
+	//remove finalizer from legacy ProjectAlert
+	if err := l.removeFinalizerFromLegacyAlerting(); err != nil {
+		return "", err
+	}
+
 	//upgrade old app
 	defaultSystemProjects, err := l.projectLister.List(metav1.NamespaceAll, labels.Set(systemProjectLabel).AsSelector())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/alert/deployer/upgradeimpl.go
```diff
@@ -383,3 +387,22 @@ func (l *AlertService) removeLegacyAlerting() error {
 	}
 	return nil
 }
+
+func (l *AlertService) removeFinalizerFromLegacyAlerting() error {
+	oldProjectAlert, err := l.oldProjectAlerts.List(metav1.ListOptions{})
```
kinarashah: want to understand why we are not Lister here?

orangedeng: same here. We should use lister instead.

aiwantaozi: Because of upgrade first happening in the new version server set up, at this moment, the cache may not full sync, we may get nothing from Lister, then the range will be skipped. Using Lister may have some timing issue.

pkg/controllers/user/alert/deployer/upgradeimpl.go
```diff
@@ -121,6 +121,10 @@ func (l *AlertService) Upgrade(currentVersion string) (string, error) {
 		if err := l.removeLegacyAlerting(); err != nil {
 			return "", err
 		}
+
+		if err := l.removeFinalizerFromLegacyAlerting(); err != nil {
```
orangedeng: Should we remove those old alerts' finalizer no matter the alert version is latest or not? I think we should use a handler for them instead of removing them in upgrade logic.

orangedeng: Just move this logic out of `strings.Contains` and it will be fine.

aiwantaozi: updated.

orangedeng:
LGTM

PR #22289, opened 2019-08-20
Fix flaky test for clustertemplate global role assignment

prachidamle:
@alena1108 @cjellick 
adding a global role to a user - then trying an operation depending on that role from the test fails intermittently - the globalrole binding does reflect on the user crd always quickly
we even try to wait for a while - but it just fails randomly. Will try to debug this separately.


alena1108:
@prachidamle can we file an issue for this TF so we don't forget to debug/put back the failing area that is being temporarily removed with this pr?

prachidamle:
https://github.com/rancher/rancher/issues/22301
@alena1108 updated with the ticket

Issue #21221, opened 2019-06-28
Fix rancher-images generation to include the original image tag


alena1108:
Today on every Drone tag we generate rancher-images.txt file containing all the third party system images needed to launch Rancher microservices (code can be found here). Prior to generation, the images have to be tagged and pushed to Rancher repo - that is so they are all kept at the same location, and to make things easier to mirror for airgapped installs. This re-tag/pushToRancherDockerhub step is manual today. And it's better to offload it to Drone. For that, we have to generate the file containing an image source and destination in a following format:

rancherTag originalTag... (amd, arm, etc)

Example:
rancher/coreos-etcd:v1 cores/etcd:v1 cores/etcd:v1-arm64

alena1108:
Today on every Drone tag we generate rancher-images.txt file containing all the third party system images needed to launch Rancher microservices (code can be found here). Prior to generation, the images have to be tagged and pushed to Rancher repo - that is so they are all kept at the same location, and to make things easier to mirror for airgapped installs. This re-tag/pushToRancherDockerhub step is manual today. And it's better to offload it to Drone. For that, we have to generate the file containing an image source and destination in a following format:

rancherTag originalTag... (amd, arm, etc)

Example:
rancher/coreos-etcd:v1 cores/etcd:v1 cores/etcd:v1-arm64

alena1108:
@kinarashah that would require some code changes in https://github.com/rancher/rancher/blob/master/pkg/image/export/main.go based on the changes you are making for separating k8s images from rke/types.

alena1108:
https://github.com/rancher/rancher/issues/20765

sangeethah:
@jiaqiluo For validating this issue , following scenarios need to be validated in air gap setup.

1. Validate that we are able to deploy all supported K8s versions in the release successfully after the images from `rancher-images.txt` gets uploaded to the private registry used by airgap setup.

sangeethah:
@kinarashah When we introduce new k8s version support in existing rancher server setup, we would expect `rke-metadata-url` to be point to a github server that is accessible to the air gap setup and is kept in sync with rancher's default `rke-metadata-url` .
With this in place , k8s version support changes will be reflected in air gapped setups as expected. But we will still not have a way to get an updated `rancher-images.txt` with the supporting system images needed.
We should have a V3 settings exposed that will carry this information and get re-evaluated  everytime changes done to kontainer metatdata.

alena1108:
[...]

PR #22302, opened 2019-08-20
generate rancher-images from latest k8s data

kinarashah:
https://github.com/rancher/rancher/issues/21221


pkg/api/customization/kontainerdriver/actionhandler.go
```diff
@@ -70,3 +83,51 @@ func (a ActionHandler) setDriverActiveStatus(apiContext *types.APIContext, statu
...
+		if err != nil {
+			return err
+		}
+		rkeSysImages[k8sVersion] = rkeSysImg
+		// ui allows only >= 1.14 for windows
+		k8sMajor := util.GetTagMajorVersion(k8sVersion)
+		if k8sMajor <= "v1.13" {
+			continue
+		}
+		winSysImg, err := kd.GetRKEWindowsSystemImages(k8sVersion, lh.WinSysImageLister, lh.WinSysImages)
```
prachidamle: if we have UI allowing only >= 1.14 for windows, why do we generate the System Images for earlier versions?

pkg/api/customization/kontainerdriver/actionhandler.go
```diff
@@ -70,3 +83,51 @@ func (a ActionHandler) setDriverActiveStatus(apiContext *types.APIContext, statu
...
+		return err
+	}
+
+	b := []byte(strings.Join(targetImages, "\n"))
+	apiContext.Response.Header().Set("Content-Length", strconv.Itoa(len(b)))
+	apiContext.Response.Header().Set("Content-Type", "application/octet-stream")
+	apiContext.Response.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s.txt", "rancher-images"))
+	apiContext.Response.Header().Set("Cache-Control", "private")
+	apiContext.Response.Header().Set("Pragma", "private")
+	apiContext.Response.Header().Set("Expires", "Wed 24 Feb 1982 18:42:00 GMT")
```
prachidamle: Why do we add an Expires header?

pkg/api/customization/kontainerdriver/actionhandler.go
kinarashah: we don't, they have been removed from the metadata windows file

prachidamle: So do we need to have a conditional check here for 1.13 and below? Instead should kd.GetRKEWindowsSystemImages() return nil or some error indicating images unavailable? In future if we deprecate the images for some later versions, we wont have to change this code if we start returning empty/unavailable response.

prachidamle:
@kinarashah Can you remind me again here - How will the air-gapped setups get the updated rancher-images.txt? Seems like we generate that file on "refresh" action from kontainermetadata - how will airgapped setup get the updated metadata

pkg/api/customization/kontainerdriver/actionhandler.go
kinarashah: `kd.GetRKEWindowsSystemImages()` would return error in this case, but we have no way of knowing if this is because there is some actual error in fetching the image (in which case we want to return this to user) or it's just something we should not be fetching in the first place. 

There's only one `k8s-current` for both General and Windows and UI bypasses it by hardcoding it... so we need to sadly hardcode too?

pkg/api/customization/kontainerdriver/actionhandler.go
kinarashah: I think it's because browser doesn't cache the file

prachidamle:
LGTM

pkg/api/customization/kontainerdriver/actionhandler.go
```diff
@@ -70,3 +80,41 @@ func (a ActionHandler) setDriverActiveStatus(apiContext *types.APIContext, statu
...
+	apiContext.Response.WriteHeader(http.StatusOK)
+	_, err = apiContext.Response.Write(b)
+	if err != nil {
```
alena1108: nit: just `return err`

alena1108:
@kinarashah any way we can add the simplest test? Against the local cluster, just to make sure the link works, and doesn't return any error to the api.

tests/integration/suite/test_kontainer_drivers.py
```diff
@@ -271,6 +272,28 @@ def test_update_duplicate_driver_conflict(admin_mc, wait_remove_resource):
         assert "Driver URL already in use:" in e.error.message
 
 
+def test_kontainer_driver_links(admin_mc):
+    client = admin_mc.client
+    lister = client.list_kontainerDriver()
+    assert 'rancher-images' in lister.links
+    token = 'Bearer '+client.token
+    data = requests.get(
+            url="https://localhost:8443/v3/kontainerdrivers/rancher-images",
```
alena1108: If there is no library support for collection link, and we are building the request ourselves, can we build the url similar to how library does it? hostname in particular

kinarashah: Thank you, made me look at how we were generating base_url. Updated.

Issue #20467, opened 2019-05-24
Settings - no error message if no-admin user edits global settings


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- run Rancher: v2.2.4-rc12
- enable Github Auth
- log in as a Github user, and by default the user is a standard user
- go to `Settings`
- edit `ui-index` to another value
- save 
<img width="717" alt="screenshot" src="https://user-images.githubusercontent.com/6218999/58360993-912a8380-7e40-11e9-8c06-2d9fb288bb8e.png">

**Result:**
- the value of `ui-index` is not changed 
- no error message shows to indicate the Github user has no permission to change settings

**Expected Result:**
An error message should show up if the user has no permission to change settings

alena1108:
Assuming the issue is not a regression in 2.2.4, moving it to 2.3 cc @jiaqiluo

loganhz:
It has `links.update` is in API. I think we should fix it in backend.

mrajashree:
@loganhz @jiaqiluo  I do see API returning a 403. I even see the error in UI, but the error popup lasts for hardly a few seconds before the second api call to refresh settings page is done. This is the error in UI. 
<img width="1440" alt="Screen Shot 2019-08-16 at 11 02 25 AM" src="https://user-images.githubusercontent.com/10672776/63188097-630aa900-c015-11e9-8b1d-ace7ab8dd7ef.png">

Error in UI:
![Untitled](https://user-images.githubusercontent.com/10672776/63188091-60a84f00-c015-11e9-87ad-c4a18203af55.png)

loganhz:
@mrajashree can we remove the `update` link in API if the user doesn't have permission to update the setting.

We are doing this for other resources. So users don't have to try if they can edit or not.

mrajashree:
@loganhz okay will do

jiaqiluo:
The bug fix is validated in Rancher:master-head d66a45ecd

Follow the same steps as the issue describes, and now the Github user cannot see the `Edit` option for the `ui-index`

<img width="1466" alt="Screen Shot 2019-08-20 at 4 03 18 PM" src="https://user-images.githubusercontent.com/6218999/63390538-5911de80-c364-11e9-84f9-461fa92d7851.png">

PR #22311, opened 2019-08-20
Set "update" link on setting based on user permissions

mrajashree:
https://github.com/rancher/rancher/issues/20467
Remove update link on settings if user doesn't have update permissions. Without this commit, API still throws a 403 when regular user tries to update global settings. But with this fix the update link won't show in UI


pkg/api/customization/setting/setting.go
```diff
@@ -32,6 +33,12 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 	} else if slice.ContainsString(ReadOnlySettings, resource.ID) {
 		delete(resource.Links, "update")
 	}
+	setting := map[string]interface{}{
```
kinarashah: we should put this in else so we don't make api call if it's already deleted

PR #22316, opened 2019-08-21
Fix version for publishing

cjellick:
Agent image had the wrong version because of this mistake.
It is working correctly on v2.2 branch and this brings the two in line.


cjellick:
LGTM 🕶

Issue #20576, opened 2019-05-30
For nodes created with node template having “sshkeycontent” set,  ssh keys fetched using “Download keys” option is not named as id_rsa.


sangeethah:
Rancher server version - v2.2.4-rc20

Steps to reproduce the problem:
Create a DigitalOcean node template driver by providing sshKeyContents and sshKeyFingerprint.

Once nodes get created successfully , use “Download keys” option to download keys.

sshkey required to access the node is present in a file named - "5NCX7AJT4G . 

Expected Behavior:
sshkey required to access the node should be present in a file named - "id_rsa"

PR #22326, opened 2019-08-21
Update naming of files for node drivers

dramich:
Problem:
When using a nodetemplate and passing in file info, that file that gets
created has a hashed name which ends up coming out in the downloaded
keys for a node

Solution:
Use the hash to create the dir and name the file as related to the field
on the driver

https://github.com/rancher/rancher/issues/20576

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -17,6 +17,7 @@ import (
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
 	"github.com/rancher/rancher/pkg/api/customization/clusterregistrationtokens"
+	"github.com/rancher/rancher/pkg/controllers/management/drivers/nodedriver"
 	"github.com/rancher/rancher/pkg/encryptedstore"
 	"github.com/rancher/rancher/pkg/jailer"
 	"github.com/rancher/rancher/pkg/namespace"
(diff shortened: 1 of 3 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -329,16 +330,34 @@ func aliasToPath(driver string, config map[string]interface{}, ns string) error
...
 				hasher.Write([]byte(fileContents))
 				sha := base32.StdEncoding.WithPadding(-1).EncodeToString(hasher.Sum(nil))[:10]
-				fullPath := path.Join(baseDir, sha)
-				err := ioutil.WriteFile(fullPath, []byte(fileContents), 0600)
+				fileName := driverField
+				if ok := nodedriver.SSHKeyFields[schemaField]; ok {
+					fileName = "id_rsa"
+				}
+
+				fileDir := path.Join(baseDir, sha)
```
cbron: Nit: seems like weird spacing here.

pkg/controllers/management/node/controller.go
```diff
@@ -329,16 +330,34 @@ func aliasToPath(driver string, config map[string]interface{}, ns string) error
...
+
+				// Cleanup the old key file if it exists
```
cbron: Can we change _"Cleanup the old key file..."_ to _"Cleanup existing key file in deprecated format..."_. The former makes it sound like it was created on last run, but really it's tech debt code that can be removed in the future.

Issue #19018, opened 2019-03-19
Global Role Bindings orphaned after cluster is removed


cloudnautique:
With the addition and removal of clusters, over time the list call on grb resources ends up taking ~10 seconds when there are around 1200 records. 

Looking at the grb user, it can be seen that:
`displayName: System account for Cluster c-vwtbx`

Looking at the cluster:
`kubectl get cluster -o yaml c-vwtbx
Error from server (NotFound): clusters.management.cattle.io "c-vwtbx" not found`

It appears as though we aren't cleaning up the role bindings or users when the clusters are removed.

mrajashree:
The system account that gets created from clusterdeploy.go, results in creation of globalrolebindings. But on cluster removal, there's no deletion being called for this system account. That could be causing the previous grbs to still exist even after cluster is removed.

mrajashree:
Same with projects, system accounts created for them aren't removed, hence globalRoleBindings are also not being removed

alena1108:
Can be tested with `v0.2.3-rc1`

sowmyav27:
Reproduced this issue on `rancher:v2.2.2`
**Steps Taken:**
1. On Rancher v2.2.2, deploy a cluster.
2. Verify the role bindings, users created - `kubectl get clusterrolebinding`, `kubectl get users` and `kubectl get globalrolebindings`
3. Create a project and namespace 
4. Verify the role bindings, users created - `kubectl get clusterrolebinding`, `kubectl get users` and `kubectl get globalrolebindings`
5. Delete the project and namespace - You will see that the global role bindings created, will not be deleted.
6. Delete the cluster. You will see that the global bindings created for the cluster, will not be deleted.

Verified on `rancher:v2.2.3-rc3`
**Steps Taken:**
1. On Rancher v2.2.3-rc3, deploy a cluster.
2. Verify the role bindings, users created - `kubectl get clusterrolebinding`, `kubectl get users` and `kubectl get globalrolebindings`
3. Delete the cluster  and verify the role bindings.
4. The user role bindings are not deleted for the projects in the cluster.

alexandrem:
Is there anyone working on this? Is it going to be part of v2.3 or it may get delayed again?

We are deeply impacted by this in our environment.

cjellick:
@alexandrem what version of rancher are you on and in what way is it impacting you?
We are currently fixing a bug in this area, but I just want to make sure it will address your problem

alexandrem:
[...]

PR #22342, opened 2019-08-21
 Remove orphaned system account users

mrajashree:
https://github.com/rancher/rancher/issues/19018

A system user account is created for every cluster and project.
2.2.3 and prior versions did not delete these accounts when their clusters
or projects were deleted. 2.2.4 has the fix to delete these system users.
But the system users left behind from cluster/project deletion prior to 2.2.4
need to be cleaned up. This only needs to be done once on startup.

### DIFF
--- a/app/user_cleanup.go
+++ b/app/user_cleanup.go
@@ -1,6 +1,8 @@
 package app
 
 import (
+	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -22,7 +24,7 @@ type userCleanup struct {
 	projectLister v3.ProjectLister
 }
 
-func cleanupOrphanedSystemUsers(management *config.ManagementContext) {
+func cleanupOrphanedSystemUsers(ctx context.Context, management *config.ManagementContext) {
 	u := userCleanup{
 		users:         management.Management.Users(""),
 		userLister:    management.Management.Users("").Controller().Lister(),
@@ -30,40 +32,44 @@ func cleanupOrphanedSystemUsers(management *config.ManagementContext) {
 		projectLister: management.Management.Projects("").Controller().Lister(),
 	}
 
-	go wait.PollImmediate(time.Hour*24, 0, func() (bool, error) {
-		logrus.Debugf("Starting orphaned system users cleanup with exponentialBackoff")
-		steps := 5
-		backOffDuration := time.Minute * 10
-		factor := 2
-		var err error
-		for steps > 0 {
-			err = u.cleanup()
-			if err != nil {
-				time.Sleep(backOffDuration)
-				backOffDuration = time.Duration(factor) * backOffDuration
-			} else {
-				break
+	cleanupCtx, cleanupCancel := context.WithCancel(ctx)
+	go func(context.Context, context.CancelFunc) {
+		wait.PollImmediate(time.Hour*24, 0, func() (bool, error) {
+			logrus.Debugf("Starting orphaned system users cleanup with exponentialBackoff")
+			steps := 5
+			backOffDuration := time.Minute * 10
+			factor := 2
+			var err error
+			for steps > 0 {
+				err = u.cleanup()
+				if err != nil {
+					time.Sleep(backOffDuration)
+					backOffDuration = time.Duration(factor) * backOffDuration
+				} else {
+					break
+				}
+				steps--
 			}
-			steps--
-		}
-		if err != nil {
-			// returning false, nil because PollImmediate terminates on error
-			return false, nil
-		}
-		return true, nil
-	})
+			if err != nil {
+				// returning false & nil because PollImmediate terminates on error
+				logrus.Error(err)
+				return false, nil
+			}
+			// no error returned, user cleanup done, calling the child context's cancelfunc to terminate child context
+			cleanupCancel()
+			return true, nil
+		})
+	}(cleanupCtx, cleanupCancel)
 }
 
 func (u *userCleanup) cleanup() error {
 	users, err := u.userLister.List("", labels.Everything())
 	if err != nil {
-		logrus.Errorf("Error listing users during system account users cleanup: %v", err)
-		return err
+		return fmt.Errorf("error listing users during system account users cleanup: %v", err)
 	}
 	clusters, err := u.clusterLister.List("", labels.Everything())
 	if err != nil {
-		logrus.Errorf("Error listing clusters during system account users cleanup: %v", err)
-		return err
+		return fmt.Errorf("error listing clusters during system account users cleanup: %v", err)
 	}
 	var returnErr error
 	for _, user := range users {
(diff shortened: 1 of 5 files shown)


app/user_cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, user := range users.Items {
+		systemUser := false
+		for _, principal := range user.PrincipalIDs {
+			if strings.HasPrefix(principal, "system://") {
+				systemUser = true
+				break
+			}
+		}
+		if !systemUser {
+			return
```
kinarashah: should this be continue?

mrajashree: no we only want to continue for system  account users

app/user_cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+		clusters: management.Management.Clusters(""),
+		projects: management.Management.Projects(""),
+	}
+	u.cleanup()
+}
+
+func (u *userCleanup) cleanup() {
+	users, err := u.users.List(v1.ListOptions{})
+	if err != nil {
+		logrus.Errorf("Error listing users during system account users cleanup: %v", err)
```
kinarashah: since this happening only once, how would this get cleaned up if just return?

mrajashree: oh I had this all in a controller at first and tested it only with all system account users when switched to this one time logic, my bad

app/user_cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+				if err != nil && !errors.IsNotFound(err) && !errors.IsGone(err) {
+					logrus.Errorf("Error deleting system user %v for cluster %v: %v", user.Name, clusterID, err)
+					return
+				}
+			}
+		} else if strings.HasPrefix(user.DisplayName, systemaccount.ProjectSystemAccountPrefix) {
+			projectID := strings.TrimPrefix(displayName, systemaccount.ProjectSystemAccountPrefix)
+			// check if this project exists, if not, delete this user
+			// how to find the cluster ID of this project: go through all clusters
+			clusters, err := u.clusters.List(v1.ListOptions{})
```
kinarashah: could we do this only once and store cluster names?

app/user_cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+					break
+				}
+			}
+			if !projectExists {
+				// project not found, so delete the system user for this project
+				err := u.users.Delete(user.Name, &v1.DeleteOptions{})
+				if err != nil && !errors.IsNotFound(err) && !errors.IsGone(err) {
+					logrus.Errorf("Error deleting system user %v for project %v: %v", user.Name, projectID, err)
+					continue
+				}
```
prachidamle: Maybe we can have a common method to handle deletion and errors during delete?

app/user_cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+		for _, principal := range user.PrincipalIDs {
+			if strings.HasPrefix(principal, "system://") {
+				systemUser = true
+				break
+			}
+		}
+		if !systemUser {
+			continue
+		}
+
```
prachidamle: Maybe would be better to have separate method to check if the systemUser should be deleted, instead of having all logic inline here. And then we could do:
for... {
		if systemUser {
			checkCleanup()
		}
}

pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -71,6 +75,15 @@ type projectLifecycle struct {
 
 func (l *projectLifecycle) sync(key string, orig *v3.Project) (runtime.Object, error) {
 	if orig == nil || orig.DeletionTimestamp != nil {
+		projectID := ""
+		splits := strings.Split(key, "/")
+		if len(splits) == 2 {
+			projectID = splits[1]
```
kinarashah: what should happen if projectID doesn't get assigned?

mrajashree: I think it'll always get assigned since the key will be of the form clusterID/projectID.

app/user_cleanup.go
```diff
@@ -0,0 +1,135 @@
...
+			steps--
+		}
+		if err != nil {
+			// returning false, nil because PollImmediate terminates on error
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
```
prachidamle: Can we add some debug log statements when PollImmediate is invoked and when a systemuser is deleted by the poll. This will help in testing and validating cleanup.

prachidamle:
LGTM

app/user_cleanup.go
```diff
@@ -0,0 +1,137 @@
...
+func cleanupOrphanedSystemUsers(management *config.ManagementContext) {
+	u := userCleanup{
+		users:         management.Management.Users(""),
+		userLister:    management.Management.Users("").Controller().Lister(),
+		clusterLister: management.Management.Clusters("").Controller().Lister(),
+		projectLister: management.Management.Projects("").Controller().Lister(),
+	}
+
+	go wait.PollImmediate(time.Hour*24, 0, func() (bool, error) {
```
alena1108: we have to pass the context to the go routine - the examples can be found in the crone library

app/user_cleanup.go
```diff
@@ -0,0 +1,137 @@
...
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
+func (u *userCleanup) cleanup() error {
+	users, err := u.userLister.List("", labels.Everything())
+	if err != nil {
+		logrus.Errorf("Error listing users during system account users cleanup: %v", err)
```
alena1108: any reason for logging the error in the method? usually we just pass it back to the cleanup() caller, and log it in the caller stack

app/user_cleanup.go
```diff
@@ -0,0 +1,137 @@
...
+		return u.deleteSystemUser(user.Name)
+	} else if strings.HasPrefix(user.DisplayName, systemaccount.ProjectSystemAccountPrefix) {
+		projectID := strings.TrimPrefix(displayName, systemaccount.ProjectSystemAccountPrefix)
+		// check if this project exists, if not, delete this user
+		// how to find the cluster ID of this project: go through all clusters
+		projectExists := false
+		for _, cluster := range clusters {
+			project, err := u.projectLister.Get(cluster.Name, projectID)
+			if err != nil && !errors.IsNotFound(err) {
+				logrus.Errorf("Error finding project %v during system account users cleanup: %v", projectID, err)
```
alena1108: same here, for error

app/user_cleanup.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		// project not found, so delete the system user for this project
+		return u.deleteSystemUser(user.Name)
+	}
+	return nil
+}
+
+func (u *userCleanup) deleteSystemUser(userName string) error {
+	err := u.users.Delete(userName, &v1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) && !errors.IsGone(err) {
```
alena1108: didn't know about isGone! Care to share what this error means? :)

Issue #20889, opened 2019-06-14
Validate S3 credentials during cluster creation and editing


soumyalj:
Version: Master

Steps

1. Create DO cluster providing invalid credentials for S3 
2. The cluster gets created successfully without validating S3 credentials
3. Take a one time snapshot

Snapshot is stuck in "Activating" state

We should validate S3 credentials during cluster creation and editing a cluster as recurring backups and snapshot creation fails with invalid S3 credentials. 
Users will be expecting recurring backups to work as cluster creation succeeded and there was no notification that S3 credentials were invalid

PR #22344, opened 2019-08-22
Validate S3 credentials on cluster create/update

moelsayed:
https://github.com/rancher/rancher/issues/20889

### DIFF
--- a/pkg/api/store/cluster/cluster_store.go
+++ b/pkg/api/store/cluster/cluster_store.go
@@ -22,6 +22,7 @@ import (
 	"github.com/rancher/rancher/pkg/clustermanager"
 	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
 	"github.com/rancher/rancher/pkg/controllers/management/clusterstatus"
+	"github.com/rancher/rancher/pkg/controllers/management/etcdbackup"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/ref"
 	"github.com/rancher/rancher/pkg/settings"
@@ -36,6 +37,7 @@ import (
 const (
 	DefaultBackupIntervalHours = 12
 	DefaultBackupRetention     = 6
+	s3TransportTimeout         = 10
 )
 
 type Store struct {
@@ -210,7 +212,9 @@ func (r *Store) Create(apiContext *types.APIContext, schema *types.Schema, data 
 	if err = setInitialConditions(data); err != nil {
 		return nil, err
 	}
-
+	if err := validateS3Credentials(data); err != nil {
+		return nil, err
+	}
 	return r.Store.Create(apiContext, schema, data)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/store/cluster/cluster_store.go
```diff
@@ -633,6 +639,38 @@ func setBackupConfigSecretKeyIfNotExists(oldData, newData map[string]interface{}
 	}
 }
 
+func validateS3Credentials(data map[string]interface{}) error {
+	s3BackupConfig := values.GetValueN(data, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig", "s3BackupConfig")
+	if s3BackupConfig == nil {
+		return fmt.Errorf("Can't find S3 backup target configuration")
```
mrajashree: Do we need to return an error here or returning nil is fine since enabling s3 backups isn't mandatory?

pkg/api/store/cluster/cluster_store.go
```diff
@@ -633,6 +639,38 @@ func setBackupConfigSecretKeyIfNotExists(oldData, newData map[string]interface{}
...
+		Endpoint:  convert.ToString(configMap["endpoint"]),
+		Region:    convert.ToString(configMap["region"]),
+		CustomCA:  convert.ToString(configMap["customCa"]),
+	}
+
+	s3Client, err := etcdbackup.GetS3Client(sbc)
+	if err != nil {
+		return err
+	}
+	bucket := convert.ToString(configMap["bucketName"])
```
kinarashah: is this bucketName already present in configMap, we should do this test before getting client if so

pkg/api/store/cluster/cluster_store.go
```diff
@@ -633,6 +640,52 @@ func setBackupConfigSecretKeyIfNotExists(oldData, newData map[string]interface{}
 	}
 }
 
+func validateUpdatedS3Credentials(old, new map[string]interface{}) error {
+	newConfig := convert.ToMapInterface(values.GetValueN(new, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig", "s3BackupConfig"))
```
kinarashah: nit: new is keyword

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -389,6 +365,48 @@ func (c *Controller) deleteS3Snapshot(b *v3.EtcdBackup) error {
...
+	}).DialContext
+	endpoint := sbc.Endpoint
+	bucketLookup := getBucketLookupType(endpoint)
+
+	// no access credentials, we assume IAM roles
+	if sbc.AccessKey == "" ||
+		sbc.SecretKey == "" {
+		creds = credentials.NewIAM("")
+		if sbc.Endpoint == "" {
+			endpoint = s3Endpoint
```
kinarashah: should this be done before calculating bucket lookup type?

pkg/api/store/cluster/cluster_store.go
```diff
@@ -633,6 +640,52 @@ func setBackupConfigSecretKeyIfNotExists(oldData, newData map[string]interface{}
...
+	if newConfig == nil {
+		return nil
+	}
+	oldConfig := convert.ToMapInterface(values.GetValueN(old, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig", "s3BackupConfig"))
+	// remove "type" since it's added to the object by API, and it's not present in newConfig yet.
```
mrajashree: Is it possible no s3 backup was enabled initially and oldConfig is nil, so you could directly call `validateS3Credentials` avoiding calling DeepEqual?

PR #22366, opened 2019-08-22
Mirror Istio images

drpebcak:
Temporary until the work associated with #21221 is done.

This mirrors istio images to the rancher docker org.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -77,6 +77,26 @@ steps:
     event:
     - push
 
+- name: docker-mirror-istio
+  image: plugins/docker
+  environment:
+    PASSWORD:
+      from_secret: docker_password
+    USERNAME:
+      from_secret: docker_username
+  commands:
+  - docker login -u $USERNAME -p $PASSWORD
+  - cat scripts/istio-images-mirror | ./scripts/docker-pull-pusher.sh
+  settings:
+    custom_dns: 1.1.1.1
+  when:
+    ref:
+      include:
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -77,6 +77,26 @@ steps:
...
 
+- name: docker-mirror-istio
+  image: plugins/docker
+  environment:
+    PASSWORD:
+      from_secret: docker_password
+    USERNAME:
+      from_secret: docker_username
+  commands:
+  - docker login -u $USERNAME -p $PASSWORD
```
cloudnautique: I think you need to use DinD image or dapper even. The plugin entrypoint is the plugin-binary.

drpebcak: we did something similar here that seems to have worked: https://github.com/rancher/ingress-nginx/blob/46c70a1f6919537b25affa642ffbec6db10292be/.drone.yml#L20

Not sure if mounting in the docker socket was required.

cloudnautique: The image you are using here is plugins/docker that uses library docker
https://github.com/rancher/ingress-nginx/blob/46c70a1f6919537b25affa642ffbec6db10292be/.drone.yml#L21
so it brings the CLI uses the host docker socket.

drpebcak: ah, yes. alright let me switch over to that

cloudnautique: Volume mount for the host socket

cloudnautique: Nevermind

Issue #21698, opened 2019-07-23
[API] Cluster Templates - Growl error when attempt disable revision in use is confusing


davidnuzik:
This is a minor issue, but is a nice-to-have for v2.3. I think it will help to make the message more clear.

Version: master-head (7/22/19)

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Create a cluster tempalte with one revision.
- Leverage the template to create a cluster.
- Either while the cluster is still provisioning or is fully ready, in the UI attempt to disable this (in-use) cluster template revision.

**Result:**
You will get an error message like this:
![image](https://user-images.githubusercontent.com/45179589/61673520-a4bd6380-aca4-11e9-9f8c-0f6917637bec.png)
The wording is confusing to me. I expected to see something more like:
`Cannot disable the clusterTemplateRevision until any clusters are no longer using it`
(or if need to be more generic, change 2nd word form "disable" to "update")

prachidamle:
This should not apply any longer when this change is in: https://github.com/rancher/rancher/issues/22371

davidnuzik:
Version: master-head (v2.3) (8/29/19)

Now there is no error message shown at all when you disable a CT that is currently in use (cluster provisioning).

There is also no error when the cluster is fully active and then you disable.

PR #22370, opened 2019-08-22
Adding Enable Disabled actions on clustertemplaterevisions

prachidamle:
https://github.com/rancher/rancher/issues/21698

Depends on types PR: https://github.com/rancher/types/pull/959


pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -70,10 +87,101 @@ func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHan
...
+		}
+		return w.enableRevision(actionName, action, apiContext)
+	case "listquestions":
+		return w.listRevisionQuestions(actionName, action, apiContext)
+
 	}
+	return httperror.NewAPIError(httperror.NotFound, "not found")
+}
+
+func (w Wrapper) disableRevision(actionName string, action *types.Action, apiContext *types.APIContext) error {
```
kinarashah: `enableRevision` and `disableRevision` could be refactored into just one function passing the bool enabled value

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -70,10 +87,101 @@ func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHan
...
+	revisionCopy := revision.DeepCopy()
+	enabled := false
+	revisionCopy.Spec.Enabled = &enabled
+
+	_, err = w.ClusterTemplateRevisions.Update(revisionCopy)
+	if err != nil {
+		return httperror.WrapAPIError(err, httperror.ServerError, "failed to disable clusterTemplateRevision")
+	}
+
+	apiContext.WriteResponse(http.StatusNoContent, map[string]interface{}{})
```
kinarashah: we should send `http.StatusOK`?

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -33,6 +39,17 @@ func (w Wrapper) Formatter(apiContext *types.APIContext, resource *types.RawReso
 	resource.Links["revisions"] = apiContext.URLBuilder.Link("revisions", resource)
 }
 
+func (w Wrapper) RevisionFormatter(apiContext *types.APIContext, resource *types.RawResource) {
+
+	if err := apiContext.AccessControl.CanDo(v3.ClusterTemplateRevisionGroupVersionKind.Group, v3.ClusterTemplateRevisionResource.Name, "update", apiContext, resource.Values, apiContext.Schema); err == nil {
+		if convert.ToBool(resource.Values["enabled"]) {
+			resource.AddAction(apiContext, disableRevisionAction)
```
kinarashah: do we need to check if it's actually in use before allowing disable action?

pkg/api/customization/clustertemplate/cluster_template.go
```diff
@@ -70,10 +87,101 @@ func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHan
...
+		apiContext.WriteResponse(http.StatusNoContent, map[string]interface{}{})
+		return nil
+	}
+
+	revisionCopy := revision.DeepCopy()
+	enabled := false
+	revisionCopy.Spec.Enabled = &enabled
+
+	_, err = w.ClusterTemplateRevisions.Update(revisionCopy)
+	if err != nil {
```
mrajashree: Do we want to check for conflict error here and retry?

pkg/api/customization/clustertemplate/cluster_template.go
prachidamle: No we are allowing to disable even if the revision is in use by some clusters. After disabling the revision is not available for new cluster creation.

pkg/api/customization/clustertemplate/cluster_template.go
prachidamle: 204 os equivalent to 200 (OK) where it indicates update request is successful, additionally it indicates that there is no content returned in the response.

pkg/api/customization/clustertemplate/cluster_template.go
prachidamle: I don't think we handle conflict errors in usual update requests, but just return error on the client request indicating the operation failed.

prachidamle: will be handling the conflict by looking up from the store

Issue #19687, opened 2019-04-17
LB Ingress controller sets identical workloads on different xip.io hosts


rockoo:
Edit ingress controller:
- select workload target 1
- set hostname to workload.t1.1XX.1XX.1XX.1XX.xip.io
- add another rule
- select workload target 2
- set hostname to workload.t2.1XX.1XX.1XX.1XX.xip.io
- save changes

Both rules now show target 2 workload as selected.

![v1](https://i.imgur.com/DRPeGKx.png "Rules")
![v2](https://i.imgur.com/M3hFKEf.png "Rules Overridden")

---
| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.2.1` UI: `v2.2.19` |
|Route|`ingresses.run`|

tfiduccia:
Version: v2.2.2
I cannot reproduce this.

deniseschannon:
@tfiduccia can you reproduce this on 2.1.x?

leodotcloud:
Root cause:

The ingress hosts/rules/targets information is stored in under the annotation `field.cattle.io/ingressState` in the format of `key := fmt.Sprintf("%s/%s/%s/%s/%s", name, namespace, host, path, port)`. When building this key, currently for any host with xip.io, we ignore the actual value causing key conflicts. Hence the last available service/workload is stored.

```
func GetStateKey(name, namespace, host string, path string, port string) string {
	logrus.Infof("MP: GetStateKey: name: %v, namespace: %v, host: %v, path: %v, port: %v", name, namespace, host, path, port)
	//ipDomain := settings.IngressIPDomain.Get()
	//if ipDomain != "" && strings.HasSuffix(host, ipDomain) {
	//	host = ipDomain
	//}
	key := fmt.Sprintf("%s/%s/%s/%s/%s", name, namespace, host, path, port)
	logrus.Infof("MP key: %v", key)
	return base64.URLEncoding.EncodeToString([]byte(key))
}
```

Fix would be to not ignore the host blindly based on the suffix. But this would require code changes to handle this new key in multiple controllers. (ingress, ingresshostgen, store).

In case of upgrades, need to check for oldkey and see if the host info is missing and repopulate it accordingly.

PR #22388, opened 2019-08-23
fix problems with xip.io hostnames

leodotcloud:
https://github.com/rancher/rancher/issues/19687

**Problem:**
When an ingress is created/edited with multiple rules involving user-specified hostnames ending with xip.io (or configured domain), the targets are not created correctly.

**Root cause:**
The workloadID information is stored under the annotation `field.cattle.io/ingressState` of the ingress object with base64 encoded keys. The key is of the format `fmt.Sprintf("%s/%s/%s/%s/%s", name, namespace, host, path, port)`. When xip.io domain is found as a suffix, the hostname is defaulted to only xip.io, missing the user input.

**Fix:**
- Use the full hostname for the key
- Don't generate the expected services until the hostnames are generated for auto-generated case
- For the upgrade scenario, the state keys are updated and the values are moved over.


alena1108:
@leodotcloud is adding integration/unit test feasible?

leodotcloud:
@alena1108 Done, added test.

pkg/controllers/user/ingresshostgen/ingresshostgen.go
```diff
@@ -60,16 +62,42 @@ func (i *IngressHostGen) sync(key string, obj *v1beta1.Ingress) (runtime.Object,
 		}
 	}
 
-	if !changed {
+	ingressState := ingress.GetIngressState(obj)
+	requiredKeysUpdate := false
```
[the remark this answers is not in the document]
orangedeng: Sorry for the late reply.
1. We can also not generate the state key in the creation via api because we don't really need the ingress service if the `xip.io` is not get generated.
2. The separate controller is only for updating the state key generated with `xip.io` as host. The ingressHostGen controller logic will be updating the ingress rule with `xip.io` and set the state key for it. I think it will be easier to understand without the upgrading logic in it.
3. Need to make an example for this one. But We can go on fixing this issue without it.

leodotcloud:
Moved the test into a separate PR: https://github.com/rancher/rancher/pull/22619

pkg/controllers/user/controllers.go
```diff
@@ -70,6 +70,7 @@ func Register(ctx context.Context, cluster *config.UserContext, clusterRec *mana
 	istio.Register(ctx, cluster)
 	certsexpiration.Register(ctx, cluster)
 	ingresshostgen.Register(ctx, cluster.UserOnlyContext())
+	ingress.Register(ctx, cluster.UserOnlyContext())
```
dramich: Why is this getting moved up to management?

orangedeng:
LGTM

pkg/controllers/user/controllers.go
leodotcloud: We need access to settings, which is not available in user context

alena1108: Ingress controller was moved to the user context to address scalability issues. If we could find a way to propagate the setting to the user cluster (storing it in ingress annotations perhaps?), that would be great

dramich: So did this always just use xip.io then? It was using settings before, so perhaps an unseen bug?

leodotcloud: No, we added a check in this controller which requires reading the ip domain from the settings.

alena1108:
This pr will be ready to merge only when we ensure that the controller stays in the user cluster space.

PR #22394, opened 2019-08-23
Enqueue linked PRTBs when roleTemplate is updated

cbron:


pkg/controllers/management/auth/role_template_handler.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	roleTemplateController  = "mgmt-auth-role-template-controller"
+	prtbByRoleTemplateIndex = "management.cattle.io/prtb-by-role-template"
+)
+
+type roleTemplateLifecycle struct {
```
cjellick: dont call it a life cycle if its just a sync handler. lifecycles have Create Updated and Remove functions and have a lot of implications (kubeneretes finalizers), so its important that the name be accurate.

pkg/controllers/management/auth/role_template_handler.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	roleTemplateController  = "mgmt-auth-role-template-controller"
```
cjellick: the word handler would be more appropriate than controller (i realize previous examples contradict this). A controller is a k8s construct. a hander is a norman construct. 1 controller can execute (via the norman framework) many handlers

dramich:
lgtm

PR #22412, opened 2019-08-26
Fix system taints missing

orangedeng:
**Problem:**
Node taints syncer will remove `node-role.kubernetes.io/etcd` and
`node-role.kubernetes.io/controlplane` taints. It is because we set
taints both in RKE and Rancher side and the Rancher side will override
those taints added by RKE

**Solution:**
- Apply taints from node pool and node template to RKE config node so RKE
will add them when generating node.
- Linux node taints controller for Windows cluster add our taints to v1.Node
directly so we don't need to reuse the taints syncer logic of v3.Node.

Related issue:

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -73,6 +73,7 @@ func Register(ctx context.Context, management *config.ManagementContext) {
 		secretStore:               secretStore,
 		nodeClient:                nodeClient,
 		nodeTemplateClient:        management.Management.NodeTemplates(""),
+		nodePoolLister:            management.Management.NodePools("").Controller().Lister(),
 		nodeTemplateGenericClient: management.Management.NodeTemplates("").ObjectClient().UnstructuredClient(),
 		configMapGetter:           management.K8sClient.CoreV1(),
 		clusterLister:             management.Management.Clusters("").Controller().Lister(),
@@ -90,6 +91,7 @@ type Lifecycle struct {
 	nodeTemplateGenericClient objectclient.GenericClient
 	nodeClient                v3.NodeInterface
 	nodeTemplateClient        v3.NodeTemplateInterface
+	nodePoolLister            v3.NodePoolLister
 	configMapGetter           typedv1.ConfigMapsGetter
 	clusterLister             v3.ClusterLister
 	schemaLister              v3.DynamicSchemaLister
@@ -122,9 +124,6 @@ func (m *Lifecycle) setupCustom(obj *v3.Node) {
 			Address: obj.Status.NodeConfig.Address,
 		},
 	}
-
-	obj.Spec.DesiredNodeTaints = taints.GetTaintsFromStrings(obj.Spec.CustomConfig.Taints)
-	obj.Spec.UpdateTaintsFromAPI = &falseValue
 }
 
 func isCustom(obj *v3.Node) bool {
@@ -209,6 +208,11 @@ func (m *Lifecycle) Create(obj *v3.Node) (runtime.Object, error) {
 func (m *Lifecycle) getNodeTemplate(nodeTemplateName string) (*v3.NodeTemplate, error) {
 	ns, n := ref.Parse(nodeTemplateName)
 	return m.nodeTemplateClient.GetNamespaced(ns, n, metav1.GetOptions{})
+}
+
+func (m *Lifecycle) getNodePool(nodePoolName string) (*v3.NodePool, error) {
+	ns, p := ref.Parse(nodePoolName)
+	return m.nodePoolLister.Get(ns, p)
 }
 
 func (m *Lifecycle) Remove(obj *v3.Node) (runtime.Object, error) {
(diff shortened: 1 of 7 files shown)


pkg/controllers/user/nodesyncer/nodetaints_test.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func getNodeInterfaceUpdateFunc(t *testing.T, cases []*syncTaintsTestCase) func(*v1.Node) (*v1.Node, error) {
+	nodeSet := caseByNode(t, cases)
+	return func(in1 *v1.Node) (*v1.Node, error) {
+		c, ok := nodeSet[in1.Name]
+		if !ok {
+			return nil, httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("node %s not found", in1.Name))
+		}
+		if !c.nodeShouldUpdate {
```
gitlawr: This check looks redundant with that in L132.

pkg/controllers/management/node/controller.go
```diff
@@ -122,9 +124,6 @@ func (m *Lifecycle) setupCustom(obj *v3.Node) {
 			Address: obj.Status.NodeConfig.Address,
 		},
 	}
-
-	obj.Spec.DesiredNodeTaints = taints.GetTaintsFromStrings(obj.Spec.CustomConfig.Taints)
```
gitlawr: The `GetTaintsFromStrings()` is not used afterward. Should we remove it

pkg/controllers/management/node/controller.go
```diff
@@ -211,6 +210,11 @@ func (m *Lifecycle) getNodeTemplate(nodeTemplateName string) (*v3.NodeTemplate,
 	return m.nodeTemplateClient.GetNamespaced(ns, n, metav1.GetOptions{})
 }
 
+func (m *Lifecycle) getNodePool(nodePoolName string) (*v3.NodePool, error) {
+	ns, p := ref.Parse(nodePoolName)
+	return m.nodePoolLister.Get(ns, p)
```
gitlawr: It looks too trivial to be a standalone function which is only used once.

pkg/controllers/management/node/controller.go
```diff
@@ -499,6 +503,11 @@ func (m *Lifecycle) saveConfig(config *nodeconfig.NodeConfig, nodeDir string, ob
 		return obj, err
 	}
 
+	pool, err := m.getNodePool(obj.Spec.NodePoolName)
```
gitlawr: What happens when the nodePoolName is empty?

pkg/controllers/management/node/controller.go
orangedeng: I would prefer this function stays with `getNodeTemplate`.

pkg/controllers/management/node/controller.go
orangedeng: It won't. The `saveConfig` function will be only run for the node created by node pool. It's controlled by condition `NodeConditionProvisioned` and this condition will be set to true in custom node and cloud k8s node.

orangedeng:
Updated

pkg/controllers/management/node/controller.go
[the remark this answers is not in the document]
orangedeng: That's the first place we generate rke node config for v3.Node. So I added the logic here.

pkg/controllers/management/node/controller.go
```diff
@@ -521,17 +530,17 @@ func (m *Lifecycle) saveConfig(config *nodeconfig.NodeConfig, nodeDir string, ob
...
+	nodeSet := taints.GetKeyEffectTaintSet(pool.Spec.NodeTaints)
+	expectTaints := pool.Spec.NodeTaints
 
 	for key, ti := range templateSet {
 		// the desired node taints is set by the node pool already. so we don't need to set it by template because
 		// the taints from node pool should override the taints from template.
 		if _, ok := nodeSet[key]; !ok {
-			obj.Spec.DesiredNodeTaints = append(obj.Spec.DesiredNodeTaints, template.Spec.NodeTaints[ti])
-			obj.Spec.UpdateTaintsFromAPI = &falseValue
+			expectTaints = append(obj.Spec.DesiredNodeTaints, template.Spec.NodeTaints[ti])
```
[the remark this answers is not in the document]
orangedeng: Updated comment

pkg/controllers/user/windows/linux_node_taints.go
```diff
@@ -29,16 +29,16 @@ var (
 // NodeTaintsController This controller will only run on the cluster with windowsPreferred is true.
 // It will add taints to the nodes with label beta.kubernetes.io/os=linux.
 type NodeTaintsController struct {
-	nodeClient v3.NodeInterface
+	nodeClient apicorev1.NodeInterface
```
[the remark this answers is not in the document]
orangedeng: Updated

orangedeng:
No, we dont' deploy windows k8s workloads as part of rke. All of the k8s add-ons we need in windows has the specific scheduling rule to schedule into windows node. And the network plugins in windows won't run as a workload but as binary because the limitation for windows node.

PR #22443, opened 2019-08-27
Patch windows service options

thxCode:
- Save Windows RKE service options into CRD
- Output the appropriate version images

**Related PR:**
https://github.com/rancher/kontainer-engine/pull/185

**Issue:**
https://github.com/rancher/rancher/issues/22470


pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -105,47 +105,31 @@ func GetK8sVersionInfo(rancherVersion string, rkeSysImages map[string]v3.RKESyst
...
-	}
-	var k8sSvcOption *v3.KubernetesServicesOptions
-	for _, name := range names {
-		obj, err := svcOptionLister.Get(namespace.GlobalNamespace, name)
-		if err != nil {
-			if !errors.IsNotFound(err) {
-				return k8sSvcOption, err
-			}
+		// detect the svc options existing or not
+		if _, exist := svcOptions[majorVersion]; !exist {
```
kinarashah: we should first search for `svcOptions[k8sVersion]`, if it exists, store it for that version. If not, search for `majorVersion`. This is because we're supporting service options for both

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -105,47 +105,31 @@ func GetK8sVersionInfo(rancherVersion string, rkeSysImages map[string]v3.RKESyst
...
-	}
-	for _, name := range names {
-		obj, err := svcOptions.GetNamespaced(namespace.GlobalNamespace, name, metav1.GetOptions{})
-		if err != nil {
-			if !errors.IsNotFound(err) {
-				return k8sSvcOption, err
+
+		rkeSysImages := rkeSysImages[k8sVersion]
+		switch osType {
+		case Windows:
```
kinarashah: for windows we should also add a check for k8sVersion's. we should continue if `majorVersion < v1.14`

pkg/image/export/main.go
```diff
@@ -88,6 +86,9 @@ func run(images ...string) error {
 		targetImages = append(targetImages, image.Mirror(i))
 	}
 
+	k8sVersionToWindowsSystemImages, _ := kd.GetK8sVersionInfo(rancherVersion,
```
kinarashah: is there some way to combine this for both linux and windows?  Actual logic of calculating k8s versions is happening twice. This function gets called only by `pkg/image/export/main.go` so feel free to make changes as you like :-)

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
thxCode: Updated

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
thxCode: Yes, I use the existing checking of `svcOptions[majorVersion]` to avoid those unsupported versions.

pkg/image/export/main.go
thxCode: Updated

alena1108:
@thxCode rke pr has been merged, can you rebase this pr with the official vendor? and squash the changes into the code and vendor commits

thxCode:
@alena1108 , I have updated the rke vendor and squashed other commits, but this PR needs https://github.com/rancher/kontainer-engine/pull/185 support, I will test it with the steps described by @kinarashah , and then ping you later.

thxCode:
@kinarashah , @alena1108 , updated `rancher/kontianer-engine` vendor.

loganhz:
Close in favor of https://github.com/rancher/rancher/pull/22738

Issue #22438, opened 2019-08-27
Add windows node failed when rancher server url included port


aiwantaozi:
Rancher version: master-head
Step to reproduce: 
1. run rancher sever with the command:
`docker run -p 8443:443 -p 8080:80 rancher/rancher:master-head`
2. enable windows
3. add windows node
4. then the agent container in windows got errors
```
INFO: https://52.197.xxx.xxx:8443 is accessible
Copy-Item : The filename, directory name, or volume label syntax is incorrect : 'C:\etc\docker\certs.d\52.197.xxx.xxx:8443\ca.crt'
At C:\etc\rancher\execute.ps1:299 char:5
+     Copy-Item -Force -Path "$sslCertDir\serverca" -Destination "$dock ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [Copy-Item], IOException
+ FullyQualifiedErrorId : System.IO.IOException,Microsoft.PowerShell.Commands.CopyItemCommand
```

sowmyav27:
Verified on master-head commit id: `d7b425da7`
**Steps:**

- run rancher sever with the command:
```
docker run -p 8443:443 -p 8080:80 rancher/rancher:master-head
```
- Deploy a windows enabled cluster. Custom cluster, AWS nodes.
- Register 1 linux node, all roles
- add windows worker node 
- Windows node is registered successfully.
- User is able to deploy windows workload successfully

PR #22445, opened 2019-08-27
Fix Windows worker adding failure

thxCode:
**Problem:**
Windows file system doesn't allow path with colons

**Solution:**
Replace the colons with blank, details from:
https://github.com/moby/moby/issues/21189#issuecomment-234997515

**Issue:**
https://github.com/rancher/rancher/issues/22438


loganhz:
lgtm

Issue #21649, opened 2019-07-19
Enable support for k8s 1.16


kinarashah:
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

deniseschannon:
Review our charts/system-charts that we aren't using any of these deprecated APIs.

alena1108:
https://github.com/rancher/rancher/issues/21734

alena1108:
We have to add experimental support for k8s 1.16. As a part of it, following needs to be done:

- [ ]  Release v1.6 hyperkube image (alpha tag is fine)
- [ ] Revise v1.16 addons' images and templates, and bump up the versions if needed
- [ ] Test v1.16 k8s APIs from Rancher UI/API to make sure the vendored client we use, is compatible. If not, we should consider bumping it up.

sangeethah:
Rancher server version - Build from master - `24e4bdf`
Able to bring up a cluster with K8s version `1.16.0-beta1`  successfully.

Existing automation runs found 1 issue with rollback scenario that is tracked in https://github.com/rancher/rancher/issues/22583

PR #22457, opened 2019-08-27
change apis endpoints for system templates to work with k8s 1.16

galal-hussein:
https://github.com/rancher/rancher/issues/21649

### DIFF
--- a/pkg/systemtemplate/template.go
+++ b/pkg/systemtemplate/template.go
@@ -94,7 +94,7 @@ rules:
 
 ---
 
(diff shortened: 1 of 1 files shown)


cipianpascu:
just tried to upgrade to 1.17.0 and got the same error:
unable to recognize "/etc/config/rke-network-plugin.yaml": no matches for kind "DaemonSet" in version "extensions/v1beta1"

balrawat:
We're getting the same on 1.17.0. Is there any ETA for this .

Issue #15035, opened 2018-08-10
Validate agent when using node drivers


superseb:
* Installing Docker fails
* Running the agent should be retried/validated (this can break when the `docker run` fails due to host/network failures)
* If agent fails to connect to the server, retrieve error and show in UI.

ref https://github.com/rancher/rancher/issues/14933 https://github.com/rancher/rancher/pull/13278

cjellick:
@luthermonson I think this issue represents some of the general docker-machine hardening we previously discussed

cjellick:
Not sure why this is stuck in peer review, but we probably need to sync with daishan. I'm not sure that the merged coded addresses everythign mentioned in the original issue

PR #22461, opened 2019-08-28
[WIP] Agent Error Rebase

luthermonson:


StrongMonkey:
Make sure the logic in redeploying is to actually validate, which means we only want to deploy twice to retrieve the error in the second try.

alena1108:
https://github.com/rancher/rancher/issues/15035

Issue #22451, opened 2019-08-27
cattle-node-agent-windows workload seen on a linux cluster and pods are deployed on removing taints on control plane node


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
- Create a linux cluster - 1 etcd, 1 control plane and 1 worker nodes - DO cluster
**Issue#1**
- Go to Cluster-->System project -->cattle-node-agent-windows workload is seen, but with no pods.

**Issue#2**
- Add a taint on control plane node
```
Key: key1
Value: value1
Effect: NoSchedule
```
- Edit node, remove the taint.

**Note:**
This also happens when you add a taint on a worker node in a cluster.

**Expected Result:**
No pods should be deployed in the `cattle-node-agent-windows` workload

**Actual Result**
Pod is deployed in the `cattle-node-agent-windows` workload. And pod is in `Unavailable` state.
<img width="1532" alt="Screen Shot 2019-08-27 at 11 08 02 AM" src="https://user-images.githubusercontent.com/26032343/63798426-e605ec00-c8be-11e9-8049-ef9e16400160.png">
<img width="1486" alt="Screen Shot 2019-08-27 at 11 08 28 AM" src="https://user-images.githubusercontent.com/26032343/63798429-e7371900-c8be-11e9-9f20-0c5b612c45ac.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): DO rke
- Kubernetes version (use `kubectl version`):

```
1.15.3
```

- Docker version (use `docker version`):

```
18.9.8
```

sowmyav27:
Verified on master-head commit id: `d7b425da7`
**Steps:**
- Create a linux cluster - 1 etcd, 1 control plane and 1 worker nodes - DO cluster
- Go to Cluster-->System project 
- `cattle-node-agent-windows` workload is NOT seen.
- On adding or deleting of taints, `cattle-node-agent-windows` workload is NOT seen.

PR #22466, opened 2019-08-28
Determine if need to deploy Windows agent

thxCode:
**Problem:**
For pure Linux cluster, we don't want to deploy a Windows agent
DaemonSet.

**Solution:**
Deploy Windows agent DaemonSet with condition `WindowsPreferedCluster = true`

**Issue:**
https://github.com/rancher/rancher/issues/22451

### DIFF
--- a/pkg/api/customization/clusterregistrationtokens/import.go
+++ b/pkg/api/customization/clusterregistrationtokens/import.go
@@ -30,7 +30,7 @@ func ClusterImportHandler(resp http.ResponseWriter, req *http.Request) {
 		authImage = authImages[0]
 	}
 
-	if err := systemtemplate.SystemTemplate(resp, image.Resolve(settings.AgentImage.Get()), authImage, token, url); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/api/customization/clusterregistrationtokens/import.go
```diff
@@ -30,7 +30,7 @@ func ClusterImportHandler(resp http.ResponseWriter, req *http.Request) {
 		authImage = authImages[0]
 	}
 
-	if err := systemtemplate.SystemTemplate(resp, image.Resolve(settings.AgentImage.Get()), authImage, token, url); err != nil {
+	if err := systemtemplate.SystemTemplate(resp, image.Resolve(settings.AgentImage.Get()), authImage, token, url, false); err != nil {
```
orangedeng: Why always set it false here?

thxCode: For now, we could not support to import a Windows cluster, as we could not deploy the wins on the Windows host.

orangedeng:
LGTM

Issue #22472, opened 2019-08-28
[Windows] Adjust the check value of disk size


thxCode:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Steps to reproduce (least amount of steps as possible):**
- Prepare 2 Linux node and 1 `1903` "small disk" Windows node on Azure
- Setup rancher server
- Add Linux controlplane
- Add Windows worker

**Result:**
![image](https://user-images.githubusercontent.com/3394724/63844334-9efd2100-c9ba-11e9-990d-64f2d6380e11.png)

**Other details that may be helpful:**
- The small disk series OS only has 30GB system disk

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher/rancher:mater-head
- Installation option (single install/HA): single install

thxCode:
cc @davidnuzik

davidnuzik:
Version: master-head (v2.3) (8/28/19)

I tested again using the same image (only smalldisk is available on Azure website) and confirmed I no longer get the Fatal error. Instead, I only get a warning and the script runs without any issues.

I ensured my windows node is usable and can run Windows workloads.
I confirmed if I have more disk space (such as over 50GB) no warning messages occur (there is a separate warning about the minimum 50GB recommended to avoid k8s reporting disk pressure/low disk).

PR #22473, opened 2019-08-28
Adjust the check value of Windows disk size

thxCode:
**Problem:**
Could not setup agent on the Azure small disk series OS system.

**Solution:**
Reduce the amount of the check value.

**Issue:**
https://github.com/rancher/rancher/issues/22472


loganhz:
lgtm

Issue #22480, opened 2019-08-28
Snapshot Restore - etcd nodes are removed form the cluster on doing a restore form snapshot


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
on master-head - commit id: `0f081729c`

1. Create an rke cluster - AWS - 3 nodes - 1 etcd, 1 control plane and 1 worker node.
2. Configure local snapshot on the cluster.
3. A recurring snapshot is taken once the cluster comes up `Active`.
4. Deploy some workloads, and take a Snapshot.
5. Add two more etcd nodes - Scale up the number of etcd nodes by 2.
6. Now do a restore on the manual snapshot taken.

**Expected Result:**
The cluster should be restored successfully to that state.

**Actual Result:**
Etcd nodes, added in Step 5 (from above)  are removed from the cluster. Nodes will be seen in `Removing` state. And will eventually go to `Provisioning` state.

**Note:**
The same issue is seen for a cluster with same spec and s3 enabled snapshots.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke
- Kubernetes version (use `kubectl version`):

```
1.15.3
```

- Docker version (use `docker version`):

```
19.3.1
```

sowmyav27:
Verified on master-head commit id: `dbd970ae7`

- Create an rke cluster - AWS - 3 nodes - 1 etcd, 1 control plane and 1 worker node.
- Configure local snapshot on the cluster.
- A recurring snapshot is taken once the cluster comes up Active.
- Deploy some workloads, and take a Snapshot.
- Add two more etcd nodes - Scale up the number of etcd nodes by 2.
- Now do a restore on the manual snapshot taken.
- The cluster should be restored successfully to that state. No etcd nodes are deleted/re provisioned.
- On doing a `kubectl delete node <node-name>`(a worker node) the node is delete from the lcuster. in case of an rke cluster, it is re provisioned. And in case of a custom cluster, it is deleted form the cluster.

Issue #21857, opened 2019-07-30
etcd restore from S3 fails


soumyalj:
Tested with build from master-head
K8s version : v1.14.4

Steps:
1.Create a custom cluster using S3 configuration with nodes as below:
N1 - control, worker
N2, N3, N4 - 3 etcds
2. Create a few workloads
3. Power off etcd nodes N2, N3, N4

After powering off the etcd nodes, the cluster changed to "Unavailable" state as expected.
Cluster was accessible through UI. The etcd nodes were deleted and new etcd nodes were added. 
Perform restore  from S3. Restore succeeded the first time.
Perform a second restore. The second restore failed. 
Cluster is stuck in "updating" state. It was observed that the etcd nodes that were deleted after power OFF were still being reached and shown in the UI as below:
<img width="1114" alt="Screen Shot 2019-07-30 at 11 17 37 PM" src="https://user-images.githubusercontent.com/18536626/62152480-46395a80-b320-11e9-9857-7a304ff00e0e.png">

Rancher server logs below:
```
2019/07/30 17:38:41 [INFO] Handling backend connection request [c-5hjw4:machine-krkdw]
2019/07/30 17:38:42 [WARNING] Can't read container logs for container [kubelet]: error during connect: Get http://%!F(MISSING)var%!F(MISSING)run%!F(MISSING)docker.sock/v1.24/containers/kubelet/logs?stderr=1&stdout=1&tail=1: failed to find Session for client c-5hjw4:m-7c842e6d03e1
2019/07/30 17:38:42 [WARNING] Can't read container logs for container [kubelet]: error during connect: Get http://%!F(MISSING)var%!F(MISSING)run%!F(MISSING)docker.sock/v1.24/containers/kubelet/logs?stderr=1&stdout=1&tail=1: failed to find Session for client c-5hjw4:m-7c842e6d03e1
2019/07/30 17:38:42 [WARNING] Can't read container logs for container [kubelet]: error during connect: Get http://%!F(MISSING)var%!F(MISSING)run%!F(MISSING)docker.sock/v1.24/containers/kubelet/logs?stderr=1&stdout=1&tail=1: failed to find Session for client c-5hjw4:m-7c842e6d03e1
[...]

PR #22491, opened 2019-08-28
Stop nodesycner reconcile during cluster restore

moelsayed:
https://github.com/rancher/rancher/issues/22480
https://github.com/rancher/rancher/issues/21857


moelsayed:
#21857 is resolved here, too. Since in the case of the second restore the issue was caused because the nodesyncer would delete the node from rancher as soon as the backup loaded into etcd. When it's added again, it doesn't have any roles assigned and the cluster fails to provision since it doesn't have any etcd nodes in this case. 

With this fix, we will relay on the node configuration that is still stored in rancher state to re-add the node to the cluster with the correct roles.

Issue #19177, opened 2019-03-26
Rancher-created registry secret is missing the auth field


nanomad:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create a registry secret from the UI
2. Inspect it via kubectl
3. Decode the .dockerconfigjson field

**Result:**
```
{"auths":{"registry.example.com":{"password":"password","username":"username"}}}
```
**Expected Result:**
```
{"auths":{"registry.example.com":{"username":"username","password":"password","auth":"dXNlcm5hbWU6cGFzc3dvcmQ="}}}
```
**Other details that may be helpful:**

If the secret is created via the rancher-supplied kubectl then the secret has the auth field inside. Also, *technically speaking* the email field should be there as well

Possibly related to:
https://github.com/kubernetes/kubectl/issues/571

This breaks tools like 
https://github.com/weaveworks/flux/
That use the secret to list images sha and tags

**Environment information**
- Rancher version: 2.1.7
- Kubernetes Version: v1.13.4
- Installation option:  Single Node

calexandre:
+1
I have the same Rancher version and I can confirm this issue.
Flux needs the Auth field to work properly.

Also, there is a forum [issue](https://forums.rancher.com/t/dockerconfigjson-fields-missing-auth-on-2-1-7/14038) reporting this behavior.

cjellick:
@deniseschannon - ping to get this on our radar for consideration

ekristen:
I would vote for this being backported to 2.2.x too.

prachidamle:
The fix is based on this upstream kubectl fix, which generates this auth field: https://github.com/kubernetes/kubernetes/pull/72344/files#diff-3bf290ed2a11e649da5afa02a5bbfd3a

When we create a dockerCredential secret from Rancher UI, we post the data via K8s REST API - which does not seem to have this "auth" field generation like kubectl does. Hence we will handle the generation of this field while creating the secret.

We will not be supporting upgrade of existing secrets. Only new secrets created will have the "auth" field generated.

ekristen:
Any thoughts on backporting to 2.2?

ekristen:
I am assuming that since this is still open that this hasn't landed yet in 2.3?

prachidamle:
@ekristen not yet, it is scoped for the next rancher patch release 2.3.3

ekristen:
[...]

PR #22497, opened 2019-08-29
[2.3.3] Generate auth field for dockerCredential and support email field

prachidamle:
https://github.com/rancher/rancher/issues/19177

This PR generates the missing "auth" field for the RegistryCredentials - for both namespaced and project level secrets.

We don't seem to support edit registry secret anymore from UI. So the update would regenerate the auth field only if both username/password are present.

The fix is based on this upstream kubectl fix, which generates this auth field: https://github.com/kubernetes/kubernetes/pull/72344/files#diff-3bf290ed2a11e649da5afa02a5bbfd3a

When we create a dockerCredential secret from Rancher UI, we post the data via K8s REST API - which does not seem to have this "auth" field generation like kubectl does. Hence we will handle the generation of this field while creating the secret.

Update to PR:
Instead of generating "auth" value in rancher api, we will leverage the existing mapper in types repo. Also supporting "email" field of the registryCredential.

Depends on types: https://github.com/rancher/types/pull/972


alena1108:
@prachidamle can we add an integration test that would create the resource and verify that the field is set

prachidamle:
@alena1108 updated an existing test to check 'auth' field is set

prachidamle:
@alena1108 @kinarashah updates to the PR:

Instead of generating "auth" value in rancher api, we will leverage the existing RegistryCredentialMapper in types repo. Also supporting "email" field of the registryCredential.

Depends on types: https://github.com/rancher/types/pull/972

pkg/api/store/secret/store.go
```diff
@@ -18,10 +18,12 @@ type Store struct {
 }
 
 func (s *Store) Create(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}) (map[string]interface{}, error) {
+
```
alena1108: nit: doesn't look like new lines are needed here

pkg/api/store/secret/store.go
```diff
@@ -22,6 +22,7 @@ func (s *Store) Create(apiContext *types.APIContext, schema *types.Schema, data
 	t = strings.TrimPrefix(t, "namespaced")
 	t = convert.Uncapitalize(t)
 	data["kind"] = t
+
```
alena1108: no need for the new line

deniseschannon:
@prachidamle I've updated this to the `release/v2.3.x` branch - can you rebase it?

PR #22501, opened 2019-08-29
Ignore the first 60s error for HPA object

orangedeng:
**Problem:**
HPA with cpu rules for newly created workload will get
`transitioning=error` because the HPA controller can't get the cpu
metrics for newly created workload.

**Solutions:**
Ignore the transitioning error for the first 60s after HPA has been
created.

Related issue:
https://github.com/rancher/rancher/issues/20955

### DIFF
--- a/pkg/api/store/hpa/store.go
+++ b/pkg/api/store/hpa/store.go
@@ -1,0 +1,60 @@
+package hpa
+
+import (
+	"time"
+
+	"github.com/rancher/norman/store/transform"
+	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/values"
+)
+
+const (
+	condition = "ScalingActive"
+	status    = "False"
+	reason    = "FailedGetResourceMetric"
+)
+
+func NewIgnoreTransitioningErrorStore(store types.Store, d time.Duration, expectedState string) types.Store {
+	return &transform.Store{
+		Store: store,
+		Transformer: func(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
+			transitioning, _ := values.GetValue(data, "transitioning")
+			if transitioning != "error" {
+				return data, nil
+			}
+			conditions, ok := values.GetSlice(data, "conditions")
+			if !ok {
+				return data, nil
+			}
+			var found bool
+			for _, c := range conditions {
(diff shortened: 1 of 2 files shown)


gitlawr:
How do we decide the 60s boundary? I don't feel good to depend on time. Should we change this state to non-error one if it can possibly reconcile by itself?

orangedeng:
60s is default scrape interval for metrics server scraping cpu metrics from kubelet. Refer to https://github.com/kubernetes-incubator/metrics-server#flags.

pkg/api/store/ignoretransitioningerror/store.go
```diff
@@ -0,0 +1,49 @@
+package ignoretransitioningerror
```
gitlawr: Do you think it's better to name the package as `hpa`, and specifically handle HPA conditions? It looks like a rare case to be used by other resources.

orangedeng: Just want to specify the store usage with the package name.

orangedeng:
Updated.

orangedeng:
Updated. Only ignores the `ScalingActive` with `False` value and `FailedGetResourceMetric` reason.

pkg/api/store/hpa/store.go
```diff
@@ -0,0 +1,56 @@
...
+
+func NewIgnoreTransitioningErrorStore(store types.Store, d time.Duration, expectedState string) types.Store {
+	return &transform.Store{
+		Store: store,
+		Transformer: func(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
+			created, ok := values.GetValue(data, "created")
+			if !ok {
+				return data, nil
+			}
+			t, err := time.Parse(time.RFC3339, created.(string))
```
cjellick: Can you put a check for this to see if transitioning == "error" before parsing the date? You could even move parsing the date to below the point where you find the offending condition so that you only have to parse the date if you know you are in this scenario?

I don't know how expensive time.Parse is, but seems like it would be good to avoid doing it if we are not in a transitioning == error state

cjellick:
This PR is fine other than that one change, so i am a future-lgtm assuming the check for whether we are in the transitioning == "error" gets added to the top

orangedeng:
Updated

PR #22509, opened 2019-08-29
fix log message parameter order

moelsayed:


moelsayed:
Log before the fix:
```
time="2019-08-29T17:52:00Z" level=info msg="For process nginx-proxy, Args has changed from [CP_HOSTS=172.31.19.23,172.31.20.89] to [CP_HOSTS=172.31.20.24]"
```
After the fix:
```
time="2019-08-29T17:52:00Z" level=info msg="For process nginx-proxy, Args has changed from [CP_HOSTS=172.31.20.24] to [CP_HOSTS=172.31.19.23,172.31.20.89]"
```

Issue #22486, opened 2019-08-28
Global Rolebindings are not cleaned up when a cluster no longer exists


daxmc99:
**What kind of request is this (question/bug/enhancement/feature request):**
This is an extension of https://github.com/rancher/rancher/issues/22281

**Steps to reproduce (least amount of steps as possible):**
1. List globalrolebindings in a Rancher server with multiple users and deleted clusters prior to 2.2.8. 
2. Some globalrolebindings exist that do not have deletion timestamps but have finalizers for deleted clusters. 
3. These GRBs have the effect of keeping some users as global admins temporarily.

**Result:**
Persistent global admins that cannot have privileges reduced temporarily

**Other details that may be helpful:**
This only occurs on clusters coming from v2.2.7 and later. This should not occur in fresh clusters from v2.2.8 onward.

Cleanup script: https://gist.github.com/daxmc99/0ffb40ed1dd46f3a240d55abfa0954ea

sowmyav27:
reproduced on an update from `v2.2.7` to `v2.2.8`.
**Steps:**
- Create a global role binding as:
```
apiVersion: management.cattle.io/v3
globalRoleName: user
kind: GlobalRoleBinding
metadata:
  annotations:
    lifecycle.cattle.io/create.grb-sync_local: "true"
    lifecycle.cattle.io/create.mgmt-auth-grb-controller: "true"
  finalizers:
    - clusterscoped.controller.cattle.io/grb-sync_local
    - clusterscoped.controller.cattle.io/grb-sync_fake
    - controller.cattle.io/mgmt-auth-grb-controller
  generation: 1
  labels:
    cattle.io/creator: norman
  name: grb-test
userName: u-test
```

- Upgrade rancher to rancher:v2.2.8
- `kubectl get globalrolebindings grb-test -o json`
```
{
    "apiVersion": "management.cattle.io/v3",
    "globalRoleName": "user",
    "kind": "GlobalRoleBinding",
    .....
        "creationTimestamp": "2019-09-05T20:55:05Z",
        "finalizers": [
            "clusterscoped.controller.cattle.io/grb-sync_local",
            "clusterscoped.controller.cattle.io/grb-sync_fake",
            "controller.cattle.io/mgmt-auth-grb-controller",
            "clusterscoped.controller.cattle.io/grb-sync_c-5hn4s"
        ],
        "generation": 1,
        "labels": {
            "cattle.io/creator": "norman"
        },
        "name": "grb-test",
[...]

PR #22521, opened 2019-08-29
Cleanup legacy globalrolebindings finalizers to deleted clusters

daxmc99:
Previously, this only removed finalizers from deleted grbs with a 1 hour delay.
This meant that when a user had their admin access removed it would take one hour for
the change to propogate if they had a finalizer to a non-existent cluster in their list of finalizers.

Now, we premptively remove "bad" finalziers from grbs that point to non-existent clusters so that on
change they will not need to be cleaned. This only occurs for grbs with a creation timestamp greater than
one hour to prevent interference with normal handler operations.

Similar to: https://github.com/rancher/rancher/pull/22338
Addresses: https://github.com/rancher/rancher/issues/22486

### DIFF
--- a/pkg/controllers/user/rbac/legacy_grb_cleaner.go
+++ b/pkg/controllers/user/rbac/legacy_grb_cleaner.go
@@ -2,9 +2,9 @@ package rbac
 
 import (
 	"strings"
-	"time"
 
 	"github.com/rancher/norman/lifecycle"
+	grbstore "github.com/rancher/rancher/pkg/api/store/globalrolebindings"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -19,27 +19,31 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
 	return &grbCleaner{m: m}
 }
 
+// this function addresses issues with grb not being cleaned up that was an issue from v2.2.3 - v2.2.8
+// it works on previously created objects outside of the timeline of normal sync handlers to correct issues with finalizers not be removed on deletion and finalizers not being dropped on cluster deletion
 func (p *grbCleaner) sync(key string, obj *v3.GlobalRoleBinding) (runtime.Object, error) {
 	if key == "" || obj == nil {
 		return nil, nil
 	}
-	if obj.DeletionTimestamp == nil || len(obj.Finalizers) == 0 {
-		return nil, nil
+	if obj.Annotations[grbstore.GrbVersion] == "true" {
+		return obj, nil
 	}
-	if time.Since(obj.DeletionTimestamp.Time) < 1*time.Hour {
-		return nil, nil
-	}
-	obj, err := p.removeStuckFinalizer(obj)
+	obj, err := p.removeFinalizerFromNonExistentCluster(obj)
 	if err != nil && !errors.IsNotFound(err) {
-		return nil, err
+		return obj, err
 	}
-	return obj, nil
-
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+		obj.SetAnnotations(annotations)
+	}
+	obj.Annotations[grbstore.GrbVersion] = "true"
+	return p.m.workload.Management.Management.GlobalRoleBindings("").Update(obj)
 }
 
-func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.GlobalRoleBinding, error) {
+func (p *grbCleaner) removeFinalizerFromNonExistentCluster(obj *v3.GlobalRoleBinding) (*v3.GlobalRoleBinding, error) {
 	obj = obj.DeepCopy()
-	modified := false
+
 	md, err := meta.Accessor(obj)
 	if err != nil {
 		return obj, err
@@ -49,15 +53,14 @@ func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.Global
 	for i := len(finalizers) - 1; i >= 0; i-- {
 		f := finalizers[i]
 		if strings.HasPrefix(f, lifecycle.ScopedFinalizerKey) {
-			finalizers = append(finalizers[:i], finalizers[i+1:]...)
-			modified = true
+			s := strings.Split(f, "_")
+			// if cluster was not reported, its a deleted cluster and will cause the finalizer to hang in the future
+			if _, err = p.m.clusterLister.Get("", s[1]); errors.IsNotFound(err) {
+				finalizers = append(finalizers[:i], finalizers[i+1:]...)
+			}
 		}
 	}
 
-	if modified {
-		md.SetFinalizers(finalizers)
-		obj, e := p.m.workload.Management.Management.GlobalRoleBindings("").Update(obj)
-		return obj, e
-	}
+	md.SetFinalizers(finalizers)
 	return obj, nil
 }
--- a/pkg/controllers/user/rbac/legacy_rt_cleaner.go
+++ b/pkg/controllers/user/rbac/legacy_rt_cleaner.go
@@ -2,9 +2,9 @@ package rbac
 
 import (
 	"strings"
-	"time"
 
 	"github.com/rancher/norman/lifecycle"
+	"github.com/rancher/rancher/pkg/api/store/roletemplate"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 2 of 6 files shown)


pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -61,3 +81,42 @@ func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.Global
...
+			if _, found := clusterMap[s[1]]; !found {
+				finalizers = append(finalizers[:i], finalizers[i+1:]...)
+				modified = true
+			}
+		}
+	}
+
+	if modified {
+		md.SetFinalizers(finalizers)
+		obj, e := p.m.workload.Management.Management.GlobalRoleBindings("").Update(obj)
```
cbron: Can just return this directly. Feels weird to return e if you know that e is going to be `IsNotFound`, and then putting that check in the caller, although `removeStuckFinalizer` follows the same pattern.

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,19 +20,38 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
...
+		obj, err := p.removeStuckFinalizer(obj)
+		if err != nil && !errors.IsNotFound(err) {
+			return obj, err
+		}
+	} else {
+		if time.Since(obj.CreationTimestamp.Time) < 1*time.Hour {
+			return obj, nil
+		}
+
+		clusterList, err := p.m.clusterLister.List("", labels.Everything())
```
dramich: Don't use the cluster lister, when parsing the finalizers for the clusterID use a lookup by ID from the cache instead.

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,19 +20,38 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
...
+		if err != nil {
+			return obj, err
+		}
+		obj, err = p.removeFinalizerFromNonExistentCluster(obj, clusterList)
+		if err != nil && !errors.IsNotFound(err) {
+			return obj, err
+		}
```
dramich: Before leaving, a GRB that has been processed successfully should get an annotation to indicate it's already been cleaned. This annotation should be checked above so we don't process the same one repeatedly.

dramich:
Add/update a store for GRB to add the annotation on create so it doesn't need to be processed through this sync. Don't worry about the case where we use a client to create one.

dramich:
This could also use some tests. Python would be fairly easy with creating a grb with some bad finalizers.

pkg/controllers/user/rbac/legacy_grb_cleaner.go
daxmc99: Nice improvement 🎆

pkg/controllers/user/rbac/legacy_grb_cleaner.go
daxmc99: Added annotation

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,22 +19,28 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
...
-	obj, err := p.removeStuckFinalizer(obj)
+	obj, err := p.removeFinalizerFromNonExistentCluster(obj)
 	if err != nil && !errors.IsNotFound(err) {
-		return nil, err
+		return obj, err
 	}
-	return obj, nil
-
+	annotations := obj.GetAnnotations()
+	if len(annotations) == 0 {
```
dramich: Use a nil check, len of 0 can still be a valid map

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,22 +19,28 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
...
+		annotations = make(map[string]string)
+		obj.Annotations = annotations
```
dramich: This could use the helper as well since it was used above `SetAnnotations`

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,22 +19,28 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
...
+	}
+	obj.Annotations[grbstore.GrbVersion] = "true"
+	return p.m.workload.Management.Management.GlobalRoleBindings("").Update(obj)
 }
 
 func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.GlobalRoleBinding, error) {
```
dramich: Is this sill used?

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -56,8 +62,35 @@ func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.Global
...
+			if _, err = p.m.clusterLister.Get("", s[1]); errors.IsNotFound(err) {
+				finalizers = append(finalizers[:i], finalizers[i+1:]...)
+				modified = true
+			}
+		}
+	}
+
+	if modified {
+		md.SetFinalizers(finalizers)
+		return p.m.workload.Management.Management.GlobalRoleBindings("").Update(obj)
```
dramich: This isn't needed, call update after you add the annotation or update is going to get called twice in a row which is a good way to get conflict errors.

tests/integration/suite/test_users.py
```diff
@@ -16,3 +18,62 @@ def test_user_cant_deactivate_self(admin_mc):
...
+        },
+        "userName": "u-" + random_str()
+    }
+    grb_k8s = api.create_cluster_custom_object(group="management.cattle.io",
+                                               version="v3",
+                                               plural="globalrolebindings",
+                                               body=json
+                                               )
+    remove_resource(grb_k8s["metadata"]["name"])
+    grb_k8s = api.get_cluster_custom_object(group="management.cattle.io",
```
dramich: After it's created you should be able to use the normal rancher client to interact with it going forward, that would help cleanup the code

tests/integration/suite/test_users.py
```diff
@@ -16,3 +18,62 @@ def test_user_cant_deactivate_self(admin_mc):
         client.update(admin_mc.user, enabled=False)
 
     assert e.value.error.status == 422
+
+
+def test_globalrolebinding_create(admin_mc, remove_resource):
```
dramich: Nit: update the name to reflect this is part of checking cleanup. A comment describing that would be good as well. `test_globalrolebinding_finalizer_cleanup` or similar.

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,22 +19,28 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
 	return &grbCleaner{m: m}
 }
 
+// TODO  Remove this once no longer needed
```
cbron: Might add a note here saying "and its integration test". Will save some time in the future.

dramich: This comment should just be dropped, how do you know when it's not needed?

cbron: It is helpful to acknowledge that its code which is fixing a time-boxed issue though. Maybe re-phrase like _"handles bug that occured from version 2.y-2.z"_.

tests/integration/suite/test_users.py
daxmc99: I use the rancher client more in the code now, I still use the k8s client for access to the finalizers though

pkg/controllers/user/rbac/legacy_grb_cleaner.go
```diff
@@ -19,27 +19,32 @@ func newLegacyGRBCleaner(m *manager) *grbCleaner {
...
+		obj.SetAnnotations(annotations)
+	}
+	obj.Annotations[grbstore.GrbVersion] = "true"
+	return p.m.workload.Management.Management.GlobalRoleBindings("").Update(obj)
 }
 
-func (p *grbCleaner) removeStuckFinalizer(obj *v3.GlobalRoleBinding) (*v3.GlobalRoleBinding, error) {
+func (p *grbCleaner) removeFinalizerFromNonExistentCluster(obj *v3.GlobalRoleBinding) (*v3.GlobalRoleBinding, error) {
 	obj = obj.DeepCopy()
 	modified := false
```
dramich: Drop the modified check, just settings the finalizers regardless is fine and would be a no-op anyways if no changes happen.

daxmc99:
Only the most recent commit needs to be reviewed

pkg/controllers/user/rbac/legacy_rt_cleaner.go
```diff
@@ -49,15 +53,14 @@ func (p *rtCleaner) removeStuckFinalizer(obj *v3.RoleTemplate) (*v3.RoleTemplate
 	for i := len(finalizers) - 1; i >= 0; i-- {
 		f := finalizers[i]
 		if strings.HasPrefix(f, lifecycle.ScopedFinalizerKey) {
-			finalizers = append(finalizers[:i], finalizers[i+1:]...)
-			modified = true
+			s := strings.Split(f, "_")
+			// if cluster was not reported, its a deleted cluster and will cause the finalizer to hang in the future
+			if _, err = p.m.clusterLister.Get("", s[1]); errors.IsNotFound(err) {
```
cbron: Are we guaranteed every finalizer always has underscore ? If not then `s[1]` could error out.
Might also be a more descriptive var name here than `s`. The `f` var works because its short for finalizer but what does s mean.

tests/integration/suite/test_users.py
```diff
@@ -74,3 +74,53 @@ def check_annotation():
...
+            "name": "test-" + random_str(),
+        }
+    }
+    rt_k8s = api.create_cluster_custom_object(
+        group="management.cattle.io",
+        version="v3",
+        plural="roletemplates",
+        body=json,
+    )
+    remove_resource(rt_k8s["metadata"]["name"])
```
cbron: Does this work ? I only see objects being passed, and it looks like this is a string.

tests/integration/suite/test_users.py
```diff
@@ -74,3 +74,53 @@ def check_annotation():
...
+            return False
+    wait_for(check_annotation, fail_handler=lambda: "annotation was not added")
+    rt1 = api.get_cluster_custom_object(
+        group="management.cattle.io",
+        version="v3",
+        plural="roletemplates",
+        name=rt_k8s.id,
+    )
+    assert "clusterscoped.controller.cattle.io/grb-sync_fake" \
+        not in rt1["metadata"]["finalizers"]
```
cbron: Nice test, easy to understand. Spacing seems a little weird but meh.

pkg/controllers/user/rbac/legacy_rt_cleaner.go
daxmc99: For cluster scoped resources we have a defined format of `clusterscoped.controller.cattle.io/cluster-roletemplate-<clusterName>"`

dramich: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

pkg/controllers/user/rbac/legacy_rt_cleaner.go
```diff
@@ -2,7 +2,8 @@ package rbac
 
 import (
 	"strings"
-	"time"
+
+	"github.com/rancher/rancher/pkg/api/store/roletemplate"
```
dramich: Formatting

tests/integration/suite/test_users.py
dramich: No, it won't work. Move it down to where you get the obj.

tests/integration/suite/test_users.py
daxmc99: Black formatter + flake8

dramich:
lgtm

Issue #22513, opened 2019-08-29
Taints set for worker node in nodepool/node template are not available 


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
In master-head - commit id: `d7b425da7`

- Deploy an rke DO cluster - 1 etcd, 1 control plane and 1 worker node - a node pool for every node. 
- Add  taint - `NoExecute` on every node pool.
- The cluster comes up active.
- On the worker node, View in API at the node pool level, the taint is see
```
"nodeTaints": [
{
"effect": "NoExecute",
"key": "test3",
"type": "/v3/schemas/taint",
"value": "val3"
}
```
**Expected Result:**
Taint should be seen on the worker node. - in the UI and on Edit Node. It should be set on the node.

**Actual Result** 
The taint is NOT seen on the node on the UI.
kubectl get nodes <node-name> -o json -- does not have the taint set.

**Note:**
Same issue is seen for a custom cluster.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke
- Kubernetes version (use `kubectl version`):

```
1.15.3
```

- Docker version (use `docker version`):

```
19.3.1
```

sowmyav27:
Verified on latest master-head - commit id:`c464c7058 `
**Steps:**

- Deploy an rke DO cluster - 1 etcd, 1 control plane and 1 worker node - a node pool for every node.
- Add taint - NoExecute on every node pool.
- The cluster comes up active.
- On the worker node, View in API at the node pool level, the taint is see

"nodeTaints": [
{
"effect": "NoExecute",
"key": "test3",
"type": "/v3/schemas/taint",
"value": "val3"
}

- Taint is seen on the worker node. - in the UI and on Edit Node.

PR #22540, opened 2019-08-30
Fix worker taints not applied

orangedeng:
**Problem:**
Setting node pool taints and node template taints doesn't work.

**Solution:**
Add kubelet args `--register-with-taints` parameters to kubelet worker
process. So those taints will be added after kubelet is up and running.

Related issue:
https://github.com/rancher/rancher/issues/22513

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -537,7 +537,7 @@ func (m *Lifecycle) saveConfig(config *nodeconfig.NodeConfig, nodeDir string, ob
 		// the expect taints are based on the node pool. so we don't need to set taints with same key and effect by template because
 		// the taints from node pool should override the taints from template.
 		if _, ok := nodeSet[key]; !ok {
-			expectTaints = append(obj.Spec.DesiredNodeTaints, template.Spec.NodeTaints[ti])
+			expectTaints = append(expectTaints, template.Spec.NodeTaints[ti])
 		}
 	}
 	obj.Status.NodeConfig.Taints = taints.GetRKETaintsFromTaints(expectTaints)
--- a/pkg/rkenodeconfigserver/nodeserver.go
+++ b/pkg/rkenodeconfigserver/nodeserver.go
@@ -15,12 +15,14 @@ import (
 	"github.com/rancher/rancher/pkg/rkeworker"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/systemaccount"
+	"github.com/rancher/rancher/pkg/taints"
 	"github.com/rancher/rancher/pkg/tunnelserver"
 	rkehosts "github.com/rancher/rke/hosts"
 	rkeservices "github.com/rancher/rke/services"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -537,7 +537,7 @@ func (m *Lifecycle) saveConfig(config *nodeconfig.NodeConfig, nodeDir string, ob
 		// the expect taints are based on the node pool. so we don't need to set taints with same key and effect by template because
 		// the taints from node pool should override the taints from template.
 		if _, ok := nodeSet[key]; !ok {
-			expectTaints = append(obj.Spec.DesiredNodeTaints, template.Spec.NodeTaints[ti])
+			expectTaints = append(expectTaints, template.Spec.NodeTaints[ti])
```
gitlawr: Can you explain this change

pkg/rkenodeconfigserver/nodeserver.go
```diff
@@ -201,7 +202,7 @@ func (n *RKENodeConfigServer) nodeConfig(ctx context.Context, cluster *v3.Cluste
 				nc.Processes = enhanceWindowsProcesses(tempNode.Processes)
```
gitlawr: Is the kubelet param needed by windows workers?

pkg/controllers/management/node/controller.go
orangedeng: This is a bug. We should append `expectTaints` list instead of `obj.Spec.DesiredNodeTaints`.
We need to combine taints from node template and node pool.

orangedeng:
Updated

pkg/rkenodeconfigserver/nodeserver.go
orangedeng: Updated.

gitlawr:
Also, would like to know the reason for using this kubelet argument. If the taints are not added on registration, won't the framework add them in a later time?

orangedeng:
Unlike rke cli, rancher cluster provisioner only cares about control plane and etcd node and execute rke up after the configuration of those nodes is updated. That’s what we need to modify the rke worker to add taints to kubelet parameter.

pkg/rkenodeconfigserver/nodeserver.go
```diff
@@ -291,3 +294,12 @@ func enhanceWindowsProcesses(processes map[string]v3.Process) map[string]v3.Proc
 
 	return newProcesses
 }
+
+func appendTaintsToKubeletArgs(processes map[string]v3.Process, initialTaints []v3.RKETaint) map[string]v3.Process {
+	if kubelet, ok := processes["kubelet"]; ok && len(initialTaints) != 0 {
+		taintArgs := fmt.Sprintf("--register-with-taints=%s", strings.Join(taints.GetStringsFromRKETaint(initialTaints), ","))
```
alena1108: @superseb can you think of any implications of using this flag on a kubelet start

alena1108: @orangedeng would - and should - user be able to override this flag via passing custom kubelet arguments in cluster.yml config

superseb: We use the same on controlplane nodes, as long as it can take updates from the Rancher side (override), it should be fine

orangedeng: As kubelet can only accept only one `--register-with-taints` parameter, I think we should merge the taints from user custom taints and our taints.

orangedeng:
Updated PR.
1. Added tests for code
2. We will merge those taints in kubelet `extra_args` and node pool/template taints. And taints from pool/template will override the taints from kubelet `extra_args`.

orangedeng:
Addressed comments.

Issue #20122, opened 2019-05-09
[Windows Alerting] Expressions do not work correctly with Windows nodes so need to warn user


davidnuzik:
Version: Master (v2.3) (5/9/19)

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Create a cluster with windows nodes (1 or more)
- Peg the CPU on the windows nodes
- Enable monitoring
- Create an alert that uses expression to check CPU usage. Set it to alert at say 40%

**Result:**
Alert will not fire (assuming no workloads, etc) because it's not accounting for the windows nodes because we cannot (not possible today to deploy prometheus/monitoring to windows nodes).

Because of this we need to have some kind of warning in the UI explaining that metrics will not be captured for windows nodes (if and only if windows nodes are detected)

**Other details that may be helpful:**
If this is too much work for alpha:
> Because of this we need to have some kind of warning in the UI explaining that metrics will not be captured for windows nodes (if and only if windows nodes are detected)

Then  just showing any warning message regardless if okay.

davidnuzik:
Version: master-head (v2.3) (8/29/19)

I don't seem to be able to see my windows node listed:
`sum(node_load1) by (instance) / count(node_cpu_seconds_total{mode="system"}) by (instance) * 100`

I simply reused the default CPU load example we have with this expression. Only my linux master and linux worker node show up in the graph in the UI. The windows node is not listed.

Is there something wrong with this expression? Do we need to change the recommended default one in the UI?

Screenshot of the default one I'm talking about that I took this expression from:
![image](https://user-images.githubusercontent.com/45179589/63979751-79c9ea80-ca6e-11e9-8d0f-61db94eb9edd.png)

Oddly enough my CPU was above my threshold too and it never fired (even though not counting the windows node).

davidnuzik:
Reopening for review. Not sure if something is wrong with this expression or for some reason is not supported with Windows or something.

aiwantaozi:
Update pre-canned cluster level node alert to support both windows and linux. For cpu and memory, the expression could support both linux and windows. For disk running full, if your cluster has windows node, we will add one more alert for windows disk running full. If we use same disk alert for windows and linux, that expression is inaccuracy for linux, so we separate them.

davidnuzik:
Testing on hold due to https://github.com/rancher/rancher/issues/22615

davidnuzik:
[...]

PR #22544, opened 2019-08-30
Improve alert expression to work for linux and windows

aiwantaozi:
Problem:
Window cpu don't have system mode, so can use system mode to count the cpu cores, and windows file system is different

Solution:
Improve cpu load alert and add extra windows file alert

Issue:
https://github.com/rancher/rancher/issues/20122

### DIFF
--- a/pkg/controllers/user/alert/initdata.go
+++ b/pkg/controllers/user/alert/initdata.go
@@ -3,13 +3,26 @@ package alert
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
 	"github.com/rancher/rancher/pkg/controllers/user/alert/common"
 	"github.com/rancher/rancher/pkg/controllers/user/alert/manager"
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/alert/initdata.go
```diff
@@ -498,3 +511,65 @@ func getCommonRuleField(groupID, displayName, severity string) v3.CommonRuleFiel
...
+}
+
+func (l *windowsNodeSyner) Sync(key string, obj *corev1.Node) (runtime.Object, error) {
+	windowsNodes, err := l.nodeLister.List(metav1.NamespaceAll, windowNodeLabel)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to list nodes")
+	}
+
+	if len(windowsNodes) != 0 {
+		return obj, l.initClusterWindowsAlert()
```
alena1108: @aiwantaozi just for my own knowledge, where in rancher code do we create pre-canned/default linux alerts?

aiwantaozi: Hi @alena1108 when registering the all alert related controllers, the default linux alerts created, when windows node added, we created this windows alert

Issue #22483, opened 2019-08-28
Not able to use GitLab to host the kontainer-driver-metadata repo


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- host a Gitlab server in the air-gapped environment, mirror the repo https://github.com/rancher/kontainer-driver-metadata
- run rancher in air-gapped environment
- go to `https://server-ip/v3/settings/rke-metadata-url ` to change the branch and url to the mirror repo that is hosted inside the private network
- go to `https://server-ip/v3/kontainerdrivers` -> click action refresh -> send request

**Result:**
- Error in the response 
`
2019/08/27 21:39:18 [ERROR] SettingController rke-metadata-url [rke-metadata-handler] failed with : driverMetadata failed to refresh driverMetadata err Get https://raw.githubusercontent.com/root/kontainer-driver-metadata/902f69dd6bd78ab6a32c4a21bef81f90997a1951/data/data.json: dial tcp 151.101.188.133:443: i/o timeout
2019/08/27 21:39:33 [ERROR] Unknown error: driverMetadata err Get https://raw.githubusercontent.com/root/kontainer-driver-metadata/902f69dd6bd78ab6a32c4a21bef81f90997a1951/data/data.json: dial tcp 151.101.40.133:443: i/o timeout
`

We should support not only Github but all kind of git server and git server host.

jiaqiluo:
The bug fix is validated with self-built rancher images from rancher:mater-head 19377a560b660f293ec9ccbcc60969b9ddfd8361

The steps are the same as the issue provides, now we can use Gitlab to host the kontainer-driver-metadata repo without any problems.

Issue #22329, opened 2019-08-21
K8s metadata - error in Rancher's log in the air gap HA mode


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug 

**Steps to reproduce (least amount of steps as possible):**
- Install Rancher: v2.3.0-alpha7, airgap HA mode 

**Result:**

- the following error repeats in Rancher's log  

`
2019/08/21 18:42:25 [ERROR] SettingController rke-metadata-url [rke-metadata-handler] failed with : driverMetadata error getting settings fatal: unable to access 'https://github.com/rancher/kontainer-driver-metadata.git/': Failed to connect to github.com port 443: Connection timed out
`

jiaqiluo:
The bug fix is validated in Rancher:v2.3.0-rc1 in air-gapped HA mode

The mentioned error message does not show in the log after setting the `rke-metadata-refresh-interval-minutes` to 0 
 
If the user does not change anything about the metadata, Rancher will throw out the error every 1440 mins which is the default value.

PR #22557, opened 2019-08-30
driver metadata changes

kinarashah:
- sync only if interval!=0 and hash changes
- git clone instead of using raw github content
- updated log msgs

https://github.com/rancher/rancher/issues/22483
https://github.com/rancher/rancher/issues/22329

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
@@ -6,21 +6,21 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/rancher/pkg/catalog/git"
-
-	"k8s.io/apimachinery/pkg/runtime"
-
+	"github.com/rancher/rancher/pkg/randomtoken"
 	"github.com/rancher/rancher/pkg/settings"
-
 	"github.com/rancher/rancher/pkg/ticker"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
-	"github.com/sirupsen/logrus"
-
 	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 type MetadataController struct {
@@ -49,13 +49,32 @@ type TickerData struct {
 	interval   time.Duration
 }
 
+type URL struct {
+	//http path
+	path string
+	// branch set if .git path by user
+	branch string
+	// latestHash, isGit set in parseURL
+	latestHash string
+	isGit      bool
+}
+
 const (
-	rkeMetadataURL      = "rke-metadata-url"
-	rkeMetadataInterval = "rke-metadata-refresh-interval-minutes"
+	rkeMetadataConfig = "rke-metadata-config"
+	refreshInterval   = "refresh-interval-minutes"
+	fileLoc           = "data/data.json"
 )
 
-var tickerData *TickerData
-var prevURL string
+var (
+	httpClient = &http.Client{
+		Timeout: time.Second * 30,
+	}
+	dataPath    = filepath.Join("./management-state", "driver-metadata", "rke")
+	tickerData  *TickerData
+	prevHash    string
+	fileMapLock = sync.Mutex{}
+	fileMapData = map[string]bool{}
+)
 
 func Register(ctx context.Context, management *config.ManagementContext) {
 	mgmt := management.Management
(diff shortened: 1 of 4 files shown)


pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -284,3 +268,115 @@ func loadData(url string) (Data, error) {
...
+	if err != nil {
+		return data, err
+	}
+
+	if err := json.Unmarshal(buf, &data); err != nil {
+		return data, fmt.Errorf("error unmarshaling metadata contents %v", err)
+	}
+
+	if err := os.RemoveAll(path); err != nil {
+		logrus.Errorf("err removing %v", err)
```
mrajashree: nit: can we add path name to error message?

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -74,52 +98,56 @@ func Register(ctx context.Context, management *config.ManagementContext) {
...
-			prevURL = currURL
-		}
-		return setting, nil
+	settingValues, err := getSettingValues()
+	if err != nil {
+		return nil, err
 	}
 
-	interval, err := getTimeSetting()
+	interval, err := parseTime(settingValues[refreshInterval])
```
mrajashree: I think it makes sense to do parseTime check in a store create/update in addition to this as well. So even if the time is negative/zero controller won't keep returning error

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -6,9 +6,14 @@ import (
...
 	"io/ioutil"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
+	"github.com/rancher/rancher/pkg/randomtoken"
+
```
mrajashree: nit: can we remove extra blank line :)

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -74,52 +98,56 @@ func Register(ctx context.Context, management *config.ManagementContext) {
...
-			logrus.Debugf("driverMetadata: canceled counter")
-			tickerData.cancelFunc()
-			tickerData.interval = 0
+	url, urlErr := parseURL(settingValues)
+
+	if tickerData == nil {
+		// load from stored metadata on error
+		if interval == 0 || urlErr != nil {
+			logrus.Errorf("driverMetadata: error loading data, using stored defaults: interval %v, error %v", interval, urlErr)
+			if err := m.createOrUpdateMetadataDefaults(); err != nil {
```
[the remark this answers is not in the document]
kinarashah: I think the general idea is we won't make use of URL if the refresh is disabled(by setting interval 0), so even if url is changed, we'd consider it as a disabled action

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -74,52 +98,56 @@ func Register(ctx context.Context, management *config.ManagementContext) {
...
+	interval, err := parseTime(settingValues[refreshInterval])
 	if err != nil {
 		if err.Error() != "refresh disabled" {
 			return nil, err
 		}
 	}
 
-	if tickerData.interval != interval {
+	// don't sync if time is set to negative/zero
+	if interval == 0 && tickerData != nil && tickerData.interval != 0 {
```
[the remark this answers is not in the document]
kinarashah: same as below, 
> I think the general idea is we won't make use of URL if the refresh is disabled(by setting interval 0), so even if url is changed, we'd consider it as a disabled action

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
kinarashah: we don't keep returning error if time is negative/zero. but adding validation in store seems like a good idea, would add it in a separate pr?

pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -130,102 +153,77 @@ func (m *MetadataController) sync(key string, setting *v3.Setting) (runtime.Obje
...
-			logrus.Errorf("error loading rke data, using stored defaults %v", err)
-			if err := m.createOrUpdateMetadataDefaults(); err != nil {
-				return err
-			}
-			return nil
-		}
-		return err
+func (m *MetadataController) refresh(url *URL, init bool) error {
+	if !toSync(url) {
+		logrus.Infof("driverMetadata: skip sync, hash up to date %v", url.latestHash)
```
alena1108: shall we log is in debug? or having it in info is useful

Issue #22538, opened 2019-08-30
Fail to create PVCs using existing PVs


gitlawr:
**Rancher Version**
master 8/30 57508ebba

**Steps to reproduce (least amount of steps as possible):**
1. Create a single node cluster
2. Create a PV in the cluster level named `pv1`
3. Go to the default project, create a pvc named `pvc1` using the existing PV.

**Result:**
![image](https://user-images.githubusercontent.com/5697937/63991784-a425b180-cb1b-11e9-8075-b273f78ea223.png)

```
# here's the request body for the PVC creation API from the console
{
  "type": "persistentVolumeClaim",
  "namespaceId": "default",
  "accessModes": [
    "ReadWriteOnce"
  ],
  "name": "pvc1",
  "volumeId": "pv1",
  "storageClassId": null,
  "resources": {
    "requests": {
      "storage": "100Gi"
    }
  }
}
# here's the response body
{
  "baseType": "error",
  "code": "ServerError",
  "message": "resource name may not be empty",
  "status": 500,
  "type": "error"
}
```

Checked that in v2.2.7 it works with similar API request, so it should be an API issue.

gitlawr:
It also breaks binding new PVC on workload creation

cjellick:
maybe related to v1.16 upgrade? I dont know

cjellick:
I honestly have no idea on this one @luthermonson. Good luck!

luthermonson:
Looks like the validator for storage class were blocking it if you don't set a storage class: https://github.com/rancher/rancher/pull/22562

sowmyav27:
verified on latest `master-head`. Commit id: `7ed198d76`

**Steps:**

- Create a single node cluster
- Create a PV in the cluster level named pv1
- Go to the default project, create a pvc named pvc1 using the existing PV. 
- Volume is created successfully.
- Attach the volume in a workload.
- The workload fails to deploy.

**Error:**
```

Warning | FailedAttachVolume | AttachVolume.Attach failed for volume "pv1" : InvalidParameterValue: Value (v1) for parameter volumes is invalid. Expected: 'vol-...'. status code: 400, request id: 5aa82c06-7784-407b-89c3
Warning | FailedAttachVolume | (combined from similar events): AttachVolume.Attach failed for volume "pv1" : InvalidParameterValue: Value (v1) for parameter volumes is invalid. Expected: 'vol-...'. status code: 400, request id: 697475bc-9884-4a40-83aa-035d125118a7 
[...]

PR #22562, opened 2019-08-30
return if no storage class

luthermonson:


cjellick:
https://github.com/rancher/rancher/issues/22538

PR #22566, opened 2019-08-30
remove unused properties

cjellick:


pkg/controllers/user/systemimage/systemimage.go
```diff
@@ -21,15 +20,11 @@ import (
...
-	deployments      appsv1.DeploymentInterface
-	deploymentLister appsv1.DeploymentLister
-	projectLister    v3.ProjectLister
-	projects         v3.ProjectInterface
-	userContext      *config.UserContext
-	systemSercices   map[string]SystemService
+	clusterName    string
+	projectLister  v3.ProjectLister
+	projects       v3.ProjectInterface
+	userContext    *config.UserContext
```
dramich: This can be dropped as well.

PR #22595, opened 2019-09-03
[WIP] Add "Get User Info" global role

gitlawr:
There's no way to give related permissions to base-role users atm.
As a result, a project owner with custom roles is not able to add project members.


cjellick:
I think you can create custom global roles now, so this shouldnt be needed

PR #22619, opened 2019-09-04
Adding test

leodotcloud:
Separating the test from this PR: https://github.com/rancher/rancher/pull/22388


pkg/controllers/user/controllers.go
```diff
@@ -70,6 +70,7 @@ func Register(ctx context.Context, cluster *config.UserContext, clusterRec *mana
 	istio.Register(ctx, cluster)
 	certsexpiration.Register(ctx, cluster)
 	ingresshostgen.Register(ctx, cluster.UserOnlyContext())
+	ingress.Register(ctx, cluster.UserOnlyContext())
```
dramich: Why is this getting moved up to management?

leodotcloud:
This PR will be ready for merge once the original PR is fixed and merged. Otherwise, this will result in a test failure and drone will have issues.

Issue #17962, opened 2019-02-11
Rancher 2.2 Alpha 6 monitoring notification send_resolved: true


panho66:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Steps to reproduce (least amount of steps as possible):**
Goto monitoring, 'Add notify'
For option, 'Email', Slack and so on, there is no option to set  'send_resolved:true'. That cause Prometheus alertmanager to send alert [resolved]

**Result:**

**Other details that may be helpful:**
Email sample

   - name: 'wi-email'
     email_configs:
       - to: 'ApplicationSupport@somewhere.com.au'
         **send_resolved: true**

Slack sample
   - name: 'instalment'
     slack_configs:
     - channel: application-channel
       **send_resolved: true**
       text: "{{ range .Alerts }}{{ .Annotations.description }}\n{{ end }}"
       http_config:
         tls_config:
            insecure_skip_verify: ALERTMGR_SSL_SKIP_VERIFY
         proxy_url: http://proxy:8080

jiaqiluo:
In master-head 9ce31ab5e

The option `Send Resolved Alerts` is available for all supported notifiers. 

- add Slack as a notifier with the `Send Resolved Alerts` option enabled, 
- see the value `"sendResolved": true` in the API call's response and when viewing in API
- enable the cluster monitoring
- At the cluster-level alert section, configure the following alert
<img width="1457" alt="Screen Shot 2019-09-03 at 2 48 24 PM" src="https://user-images.githubusercontent.com/6218999/64211090-eae81400-ce59-11e9-86db-208b6424b214.png">

- launch an app which pods are deployed in the node `jiaqi-master-worker1`
- receive the alter in Slack: 
```
The CPU usage on the node jiaqi-master-worker1 is over 20%
Alert Name: a2
Severity: critical
Cluster Name: do (ID: c-7h9jt)
Used CPU: 1500 m
Total CPU: 4000 m
```
- delete the app
- wait for a while, receive the following message that contains only the title of the previous alter
```
The CPU usage on the node jiaqi-master-worker1 is over 20%
```


@gitlawr Is it a valid testing case? Is the second alert message (the one has only the title) the `resolved alert`?

gitlawr:
@jiaqiluo We should make the text more clear

jiaqiluo:
The bug fix is validated in rancher:master-head 41ce3ac49

[...]

PR #22621, opened 2019-09-04
Add "[Resolved]" to notification title on resolved

gitlawr:
Issue: https://github.com/rancher/rancher/issues/17962

So as to make resolved notifications recognizable.


aiwantaozi:
LGTM

cjellick:
in addition to adding "[Resolved]", this add a `-` to every if/else line. ideally, that would have been called at our explained in the commit message

gitlawr:
Noted. Will improve the details in further commits.
Regarding the `-`s, they are added to trim the space between `[Resolved]` and the original message, as explained in [gotemplate docs](https://golang.org/pkg/text/template/#hdr-Text_and_spaces).

Issue #22591, opened 2019-09-02
Can't send notifications to multiple recipients in Pipeline Configuration 


pgroot:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
I'm trying to add multiple recipients in the Pipeline Configuration.
![image](https://user-images.githubusercontent.com/1168948/64106581-0cd68d80-cdab-11e9-801e-c47a8a8dddb3.png)
However, by setting up the configuration as shown above, only the address 'huanghu@minder.co.uk' can get the notification email, it received the same notification THREE TIMES. 

It seems that only the bottom one can get the notification.

**Result:**
Only the last recipient can get the notification.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.2.7
- Installation option (single install/HA):
single 

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory): AWS t2-large
- Kubernetes version (use `kubectl version`):
![image](https://user-images.githubusercontent.com/1168948/64106856-a43be080-cdab-11e9-8ec8-7a81e9b30330.png)

- Docker version (use `docker version`):
19.3.1

tfiduccia:
Version - master
I was able to reproduce this.  Pipeline notification, will send the last recipient multiple emails dependent on the amount of recipients I added.  We should really fix this.  I checked the console and it's sending and posting properly.  It also looks like it's stored in the API properly.

bmdepesa:
rancher/rancher:master-head 41ce3ac49 

* Deployed Rancher master-head
* Created cluster
* Created an email notifier
* Created a pipeline with multiple notification recipients on success/failure 
* Executed pipeline
* Email notifications were sent properly to each individual recipient

PR #22622, opened 2019-09-04
Fix notifying multiple recipients in pipeline

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/22591

Problem:
When multiple recipients are configured in pipeline notification, one of
them get notified multiple times.

Solution:
Avoid using loop iterator variable in go routines.

### DIFF
--- a/pkg/controllers/user/pipeline/controller/pipelineexecution/pipelineexecution.go
+++ b/pkg/controllers/user/pipeline/controller/pipelineexecution/pipelineexecution.go
@@ -480,7 +480,8 @@ func (l *Lifecycle) doNotify(obj *v3.PipelineExecution) (runtime.Object, error) 
 		message = obj.Spec.PipelineConfig.Notification.Message
 	}
 	var g errgroup.Group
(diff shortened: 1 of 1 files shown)


orangedeng:
LGTM

pkg/controllers/user/pipeline/controller/pipelineexecution/pipelineexecution.go
```diff
@@ -480,7 +480,8 @@ func (l *Lifecycle) doNotify(obj *v3.PipelineExecution) (runtime.Object, error)
 		message = obj.Spec.PipelineConfig.Notification.Message
 	}
 	var g errgroup.Group
-	for _, toSendRecipient := range toSendRecipients {
+	for i := range toSendRecipients {
+		toSendRecipient := toSendRecipients[i]
```
cjellick: I do not understand how this fixes the bug.

cjellick: @rmweir pointed out that there's a goroutine down below

Issue #22615, opened 2019-09-03
[Alerts] Not working in Windows cluster as alertmanager-cluster-alerting system workload sets the matcher as kubernetes.io/os = linux


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Add Custom Windows Cluster with Azure cloud.
  - I used 1 master (etcd/control plane), one linux worker, two windows 1903 nodes .
- Activating alerts at cluster level.
- Setup an alert to trigger with some pod action. With notifier Slack or email.
- 

**Result:**

Alert fails:
`[ERROR] Failed to send alert: Post http://alertmanager-operated.cattle-prometheus.svc.cluster.local:9093/api/v1/alerts: dial tcp: lookup alertmanager-operated.cattle-prometheus.svc.cluster.local on 10.43.0.10:53: no such`

![Screen Shot 2019-09-03 at 3 44 42 PM](https://user-images.githubusercontent.com/27659/64213997-9d70a480-ce63-11e9-87c0-8afa0ef09a3b.png)

**Other details that may be helpful:**

The alerts workload **alertmanager-cluster-alerting** has:

`kubernetes.io/os = linux`

Linux Worker in the Custom cluster has:

`cattle.io/os=linux:NoSchedule`

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head (v2.3) (09/03/2019)

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom/Azure Cloud provider
- Machine type (cloud/VM/metal) and specifications (CPU/memory): cloud
- Kubernetes version (use `kubectl version`):

- Docker version (use `docker version`):

```
19.03.1
```

@sangeethah

cjellick:
@aiwantaozi @loganhz the associated pr also changed monitoring. this issue is only for alerting. So, make sure you call out both in the issue so that QA knows what they need to test

izaac:
Rancher version: `master-head` (v2.3) (09/05/2019)

Alerts are now flowing in the Windows cluster, I tested using a Slack notifier and got the notifications as expected.
**alertmanager-cluster-alerting** is now in available state and pods functional. 

It was deployed in the Linux worker with the tolerations working.

Monitoring and further Alerts test coverage is planned to be part of https://github.com/rancher/rancher/issues/16460. For now this is not longer blocking QA.

PR #22623, opened 2019-09-04
Remove monitoring and logging catalog id setting

aiwantaozi:
Problem:

When we deploy alerting, we are using the monitoring catalog setting to decide which version we should deploy. We should use the latest version (limited by rancher min and max version) instead of hard code version in Rancher server.

Solution:

Remove monitoring catalog from setting and remove legacy code

Issue:
https://github.com/rancher/rancher/issues/22615

### DIFF
--- a/pkg/catalog/utils/version.go
+++ b/pkg/catalog/utils/version.go
@@ -1,10 +1,12 @@
 package utils
 
 import (
+	"sort"
 	"strings"
 
 	"github.com/blang/semver"
 	mVersion "github.com/mcuadros/go-version"
+	"github.com/pkg/errors"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/rancher/pkg/catalog/utils/version"
 	"github.com/rancher/rancher/pkg/settings"
@@ -89,4 +91,38 @@ func ReleaseServerVersion(serverVersion string) bool {
 		return false
 	}
 	return true
+}
+
+func LatestAvailableTemplateVersion(template *v3.CatalogTemplate) (*v3.TemplateVersionSpec, error) {
+	versions := template.DeepCopy().Spec.Versions
+	if len(versions) == 0 {
+		return nil, errors.New("empty catalog template version list")
+	}
+
+	sort.Slice(versions, func(i, j int) bool {
+		val1, err := semver.ParseTolerant(versions[i].Version)
+		if err != nil {
+			return false
+		}
+
+		val2, err := semver.ParseTolerant(versions[j].Version)
+		if err != nil {
+			return false
+		}
+
+		return val2.LT(val1)
+	})
+
+	for _, templateVersion := range versions {
+		catalogTemplateVersion := &v3.CatalogTemplateVersion{
+			TemplateVersion: v3.TemplateVersion{
+				Spec: templateVersion,
+			},
+		}
+		if err := ValidateRancherVersion(catalogTemplateVersion); err == nil {
+			return &templateVersion, nil
+		}
+	}
+
+	return nil, errors.Errorf("template %s allowed rancher version not match current server", template.Name)
 }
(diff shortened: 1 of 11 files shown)


orangedeng:
LGTM

cjellick:
Won't this cause a problem with rancher-min-version/rancher-max-version? Like, if you're always getting latest and we've pushed a new version that is incompatible with your version of rancher, you are not taking min/max into account, so you would deploy an incompatible version.

loganhz:
We should make sure logging and alerting works well with `rancher-min-version/rancher-max-version`

Istio and monitoring should be fine as UI will send the version to backend.

deniseschannon:
Have we confirmed that istio and monitoring only show the versions for the rancher min/max version in the dropdown selection? @loganhz

loganhz:
> Have we confirmed that istio and monitoring only show the versions for the rancher min/max version in the dropdown selection? @loganhz

Yes, tested.

aiwantaozi:
Updated pr to support rancher min/max when deploying and upgrading

pkg/catalog/utils/version.go
```diff
@@ -90,3 +92,37 @@ func ReleaseServerVersion(serverVersion string) bool {
...
+		if err != nil {
+			return false
+		}
+
+		val2, err := semver.ParseTolerant(versions[j].Version)
+		if err != nil {
+			return false
+		}
+
+		return val2.LT(val1)
```
orangedeng: You are sorting the version from lowest to largest. After sorted, you will find the lowest version in the `for _, templateVersion := range versions {` below.

pkg/catalog/utils/version_test.go
```diff
@@ -129,3 +132,73 @@ func TestVersionSatifiesRange(t *testing.T) {
...
+					{
+						ExternalID:        "catalog://?catalog=library&template=artifactory-ha&version=0.12.16",
+						Version:           "0.12.16",
+						RancherMinVersion: "v2.2.0",
+						RancherMaxVersion: "v2.3.0",
+					},
+					{
+						ExternalID:        "catalog://?catalog=library&template=artifactory-ha&version=0.12.15",
+						Version:           "0.12.15",
+						RancherMinVersion: "v2.1.0",
```
orangedeng: You need to add test for same Rancher range with different version. So that you can make sure your function `LatestAvailableTemplateVersion` will return the latest one.

pkg/settings/setting.go
```diff
@@ -55,10 +55,7 @@ var (
 	UIKubernetesDefaultVersion        = NewSetting("ui-k8s-default-version-range", "<=1.14.x")
 	WhitelistDomain                   = NewSetting("whitelist-domain", "forums.rancher.com")
 	WhitelistEnvironmentVars          = NewSetting("whitelist-envvars", "HTTP_PROXY,HTTPS_PROXY,NO_PROXY")
-	SystemMonitoringCatalogID         = NewSetting("system-monitoring-catalog-id", "catalog://?catalog=system-library&template=rancher-monitoring&version=0.0.3")
-	SystemLoggingCatalogID            = NewSetting("system-logging-catalog-id", "catalog://?catalog=system-library&template=rancher-logging&version=0.1.1")
-	SystemExternalDNSCatalogID        = NewSetting("system-externaldns-catalog-id", "catalog://?catalog=system-library&template=rancher-external-dns&version=0.0.2")
-	SystemGlobalIstioCatalogID        = NewSetting("system-global-istio-catalog-id", "catalog://?catalog=system-library&template=rancher-istio&version=0.0.1")
+	SystemExternalDNSCatalogID        = NewSetting("system-externaldns-catalog-id", "catalog://?catalog=system-library&template=rancher-external-dns&version=0.0.1")
```
orangedeng: Why are we downgrading the external dns catalog id version?

dramich:
lgtm

Issue #18874, opened 2019-03-13
test_project_owner test failure


cjellick:
https://drone-publish.rancher.io/rancher/rancher/166/2/2

```
============================ FAILURES =================================== | 518s
-- | --
8657 | ______________________________ test_project_owner ______________________________ | 518s
8658 | [gw1] linux -- Python 3.7.1 /go/src/github.com/rancher/rancher/tests/.tox/py37/bin/python3.7 | 518s
8659 |   | 518s
8660 | admin_cc = <core.conftest.ClusterContext object at 0xffff754f0c50> | 518s
8661 | admin_mc = <core.conftest.ManagementContext object at 0xffff97c13160> | 518s
8662 | user_mc = <core.conftest.ManagementContext object at 0xffff94211320> | 518s
8663 | remove_resource = <function remove_resource.<locals>._cleanup at 0xffff7548a6a8> | 518s
8664 |   | 518s
8665 | def test_project_owner(admin_cc, admin_mc, user_mc, remove_resource): | 518s
8666 | """Tests that a non-admin member can create a project, create and | 518s
8667 | add a namespace to it, and can do workload related things in the namespace. | 518s
8668 |   | 518s
8669 | This is the first test written incorporating a non-admin user and the | 518s
8670 | kubernetes python client. It does a lot partially as an experiment and | 518s
8671 | partially as an example for other yet-to-be-written tests | 518s
8672 | """ | 518s
8673 | admin_client = admin_mc.client | 518s
8674 | admin_client.create_cluster_role_template_binding( | 518s
8675 | userId=user_mc.user.id, | 518s
8676 | roleTemplateId="cluster-member", | 518s
8677 | clusterId=admin_cc.cluster.id, | 518s
8678 | ) | 518s
8679 |   | 518s
8680 | user_client = user_mc.client | 518s
8681 |   | 518s
8682 | # When this returns, the user can successfully access the cluster and thus | 518s
8683 | # can create a project in it. We generally need this wait_until_available | 518s
8684 | # call when we are creating cluster, project, and namespaces as non-admins | 518s
8685 | # because until the rbac controllers have had a chance to run and the | 518s
8686 | # creator is bootstrapped into the resource, they will not be able to | 518s
8687 | # access it | 518s
8688 | wait_until_available(user_client, admin_cc.cluster) | 518s
8689 |   | 518s
8690 | proj_name = 'test-' + random_str() | 518s
8691 |   | 518s
8692 | def can_create_project(): | 518s
[...]

PR #22647, opened 2019-09-04
Fix flakey test_project_owner

cbron:
Issue: https://github.com/rancher/rancher/issues/18874


cbron:
This is 2.3-backlog so won't be merging yet.

Issue #22583, opened 2019-08-31
k8s 1.16 - Not able to rollback deployments.


sangeethah:
Rancher server version - Build from master - 24e4bdfe4

Steps to reproduce the problem:
Create  a K8s 1.16.0-beta1 cluster.
Deploy a workload with image "ubuntu:xenial"
Once the worklaod is active , edit and change the image to "nginx"
Wait for the upgrade to be done and try to rollback to the previous version ( with mage "ubuntu:xenial")

Following error is seen:

```
"ServerError : Error updating workload deployment:test-12761:default-13764 by rollback : the server could not find the requested resource\n\t{'baseType': 'error', 'code': 'ServerError', 'message': 'Error updating workload deployment:test-12761:default-13764 by rollback : the server could not find the requested resource', 'status': 500, 'type': 'error'
```

sangeethah:
Tested with latest build from master - `e78ddbbb9` .

Able to rollback deployments successfully.

PR #22649, opened 2019-09-04
API group changes for k8s 1.16 

mrajashree:
https://github.com/rancher/rancher/issues/22583


pkg/api/customization/workload/workload.go
```diff
@@ -124,6 +127,16 @@ func update(apiContext *types.APIContext, data map[string]interface{}, ID string
 
 func (a ActionWrapper) rollbackDeployment(apiContext *types.APIContext, clusterContext *config.UserContext,
 	actionName string, deployment projectclient.Workload, namespace string, name string) error {
+	split := strings.SplitN(apiContext.ID, ":", 3)
+	// apiContext.ID is of the form deployment:ns:name
+	if len(split) != 3 || split[0] != appsv1.DeploymentResource.SingularName {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("invalid deployment %s", apiContext.ID))
```
kinarashah: do we need to this again? we already do access.ByID for deployment type before calling the action and namespace and name is passed to this function

pkg/api/customization/workload/workload.go
```diff
@@ -124,6 +127,16 @@ func update(apiContext *types.APIContext, data map[string]interface{}, ID string
...
+	}
+	deploymentNamespace, deploymentName := split[1], split[2]
+	depl, err := clusterContext.Apps.Deployments("").Controller().Lister().Get(deploymentNamespace, deploymentName)
```
kinarashah: I think we should do this after validation of input, ie ParseAndValidateActionBody as well as validation on revision

pkg/api/customization/workload/workload.go
```diff
@@ -144,6 +157,30 @@ func (a ActionWrapper) rollbackDeployment(apiContext *types.APIContext, clusterC
 	if revision == "" {
 		return httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("ReplicaSet %s doesn't exist for deployment %s", rollbackInput.ReplicaSetID, deployment.ID))
 	}
+	deploymentVersion, err := k8sschema.ParseGroupVersion(depl.APIVersion)
```
kinarashah: we don't need to call L155 if we're already getting replicaSet ourselves, maybe this needs to be moved up a bit

pkg/api/customization/workload/workload.go
```diff
@@ -144,6 +157,30 @@ func (a ActionWrapper) rollbackDeployment(apiContext *types.APIContext, clusterC
...
+		// DeploymentRollback & RollbackTo are deprecated in apps/v1
+		// only way to rollback is update deployment podSpec with replicaSet podSpec
+		split := strings.SplitN(rollbackInput.ReplicaSetID, ":", 3)
+		if len(split) != 3 || split[0] != appsv1.ReplicaSetResource.SingularName {
+			return httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("Invalid ReplicaSet %s", rollbackInput.ReplicaSetID))
+		}
+		replicaNamespace, replicaName := split[1], split[2]
+		rs, err := clusterContext.Apps.ReplicaSets("").Controller().Lister().Get(replicaNamespace, replicaName)
+		if err != nil {
+			return err
```
kinarashah: httperror.NewAPIError

mrajashree:
Can't add integration tests, creation of pods for workloads isn't going through. And for the logic to get tested we'll need a replicaSet to be created properly.
Some of the methods called from the rollback actionHandler don't have mocks so unable to add unit tests.

Issue #22670, opened 2019-09-05
Scaling up the worker node should not give capability - "taintSupport": true on a cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

1. Have an rke cluster- AWS - 1 etcd, 1 control plane and 1 worker node. in v2.2.8
2. Upgrade to rancher master latest - `4c7e3263a` - commit id.
3. Scale up the number of worker nodes

**Expected Result:**
`"taintSupport": true` should NOT be set/seen on the cluster. (View in API, after the cluster has come up `Active`)

**Actual Result:**
`"taintSupport": true` should is set on the cluster.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.8 to rancher-master
- Installation option (single install/HA): single

mrajashree:
Tested the fix locally with [these](https://github.com/rancher/rancher/pull/22674#issuecomment-528996895) steps

sowmyav27:
Waiting for this issue to be fixed -- https://github.com/rancher/rancher/issues/22773

sowmyav27:
verified on upgrade from 2.2.8 to latest master-head, commit id: `ad92680c1`

**Steps:**

- Have an rke cluster- AWS - 1 etcd, 1 control plane and 1 worker node. in v2.2.8
- Upgrade to rancher master-head latest 
- Scale up the number of worker nodes
- "taintSupport": true is NOT set/seen on the cluster. (View in API, after the cluster has come up Active)

PR #22674, opened 2019-09-05
Add taints capability only on rke up

mrajashree:
https://github.com/rancher/rancher/issues/22670

Adding a worker node to an RKE cluster does not lead to rke up.
But the provisioner controller still calls the update, and sets the
taints capability after it. This commit checks if the update actually led
to rke up and sets taints capability based on that


mrajashree:
Tested with these steps:
1. Created a DO cluster on release/v2.2 branch. 
2. Upgraded rancher server by checking out this branch. Taints capability was NOT set and "Edit Taints" option was not available
3. Added a worker node. Taints capability was still NOT set and "Edit Taints" option was not available
4. Added a control plane node. Taints caps did get set and "Edit Taints" became available

Issue #22680, opened 2019-09-06
[Alerting] Node disk is running full within 24 hours default alert is unexpectedly firing


davidnuzik:
Version: master-head (v2.3) (9/5/19)

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Enable alerting

**Result:**
- After a while Node disk is running full within 24 hours alert may fire even though all linux nodes appear to have sufficient free space

davidnuzik:
@aiwantaozi I see the same issue as the Windows Expression issue. No metric graph data. And the expression results in No data Points in prometheus for me.

![image](https://user-images.githubusercontent.com/45179589/64465573-4583a900-d0c2-11e9-8969-88d710442f40.png)

aiwantaozi:
Yes, you could see it would alert when the value is not null, so it wouldn't have any value, this is an alert for the disk is running full within 24h, also an example to tell the user they can input an expression without threshold. There shouldn't have any graph when the predicted value of disk space is not less than 0, you could update the expression to another value, like 1073741824(1G) or other values

davidnuzik:
Thanks. I will create a separate issue to track this secondary issue as Logan has requested in https://github.com/rancher/rancher/issues/20122

davidnuzik:
Version: master-head (v2.3) (9/9/19)

I have been trying various things and can never get the rule to fire. I successfully got the Windows node disk alert rule to fire, but never the linux version.

To test I basically did the same as my windows test. I would create files about 1GB in size and every 10 mins create another. Eventually I got to 78% disk utilization in as little as an hour. The rule will never fire.

This is also true if I decrease the timeframe for `predict_linear(node_filesystem_files_free{mountpoint!~"^/etc/(?:resolv.conf|hosts|hostname)$"}[6h], 3600 * 24) < 0`

For example I tried the following. None of these worked (rule never fires)

`predict_linear(node_filesystem_files_free{mountpoint!~"^/etc/(?:resolv.conf|hosts|hostname)$"}[10m], 3600 * 24) < 0`
`predict_linear(node_filesystem_files_free{mountpoint!~"^/etc/(?:resolv.conf|hosts|hostname)$"}[20m], 3600 * 24) < 0`

and 30m, 1h, etc and tried changing timescale to all sorts of time ranges and made no difference.

[...]

PR #22681, opened 2019-09-06
Improve alert linux node disk running full expression

aiwantaozi:
Problem:
After monitoring charts upgraded, this expression value change from no data point to 0

Solution:
Update the thredshold for this expression

Issue:
https://github.com/rancher/rancher/issues/22680


orangedeng:
LGTM

Issue #22488, opened 2019-08-28
Update rancher system charts to work in k8s v1.16


cjellick:
See: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
Deprecated APIs are being removed from k8s 1.16. We need to make sure the charts in our catalogs will work with v1.16\

Relates to https://github.com/rancher/rancher/issues/22426

deniseschannon:
The system charts to check that work on k8s 1.16 is

external-dns
logging
monitoring
istio

Just need to make sure it deploys with k8s 1.16 due to the API deprecations.

bmdepesa:
rancher/rancher:master-head 4c7e3263a

Logging tested here: https://github.com/rancher/rancher/issues/21770

Monitoring app fails to deploy in 1.16 with error: `Failed to install app cluster-monitoring. Error: validation failed: unable to recognize “”: no matches for kind “PodSecurityPolicy” in version “extensions/v1beta1`

Istio enables successfully, and appears to function, but the Graph and Prometheus/Grafana links are unavailable as Grafana / Prometheus are not successfully deployed due to the monitoring chart needing updates.

jiaqiluo:
See the same error as @bmdepesa's comment in `Rancher:v2.3.0-rc1`, air-gapped HA mode
 
- enable the cluster monitoring in k8s 1.16.0 cluster

<img width="1490" alt="Screen Shot 2019-09-05 at 3 22 54 PM" src="https://user-images.githubusercontent.com/6218999/64387510-105b5600-cff1-11e9-8de6-dd4bf38cb800.png">

gitlawr:
Besides updating the API group for the newly added PSP resource, we also need to update the Prometheus operator.
see https://github.com/coreos/prometheus-operator/issues/2592

Here's the error logs from the Prometheus operator when enabling monitoring in 1.16 clusters:
```
ts=2019-09-06T03:00:39.595806765Z caller=main.go:169 msg="Starting Prometheus Operator version '0.29.0'."
level=info ts=2019-09-06T03:00:39.69272559Z caller=operator.go:213 component=alertmanageroperator msg="connection established" cluster-version=v1.16.0-beta.1
level=info ts=2019-09-06T03:00:39.693016093Z caller=operator.go:381 component=prometheusoperator msg="connection established" cluster-version=v1.16.0-beta.1
[...]

PR #22699, opened 2019-09-07
Mirror prometheus operator images

loganhz:
Mirror prometheus operator images
They will be used in https://github.com/rancher/system-charts/pull/86

Issue:

https://github.com/rancher/rancher/issues/22488

### DIFF
--- a/scripts/istio-images-mirror
+++ b/scripts/istio-images-mirror
@@ -8,3 +8,5 @@ istio/proxyv2 rancher/istio-proxyv2 1.2.5
 istio/proxy_init rancher/istio-proxy_init 1.2.5
 istio/node-agent-k8s rancher/istio-node-agent-k8s 1.2.5
 coredns/coredns rancher/coredns-coredns 1.1.2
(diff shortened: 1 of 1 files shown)


scripts/istio-images-mirror
```diff
@@ -8,3 +8,5 @@ istio/proxyv2 rancher/istio-proxyv2 1.2.5
 istio/proxy_init rancher/istio-proxy_init 1.2.5
 istio/node-agent-k8s rancher/istio-node-agent-k8s 1.2.5
 coredns/coredns rancher/coredns-coredns 1.1.2
+quay.io/coreos/prometheus-operator rancher/coreos-prometheus-operator v0.32.0
+quay.io/coreos/prometheus-config-reloader rancher/coreos-prometheus-config-reloader v0.32.0
```
alena1108: nit: new line @loganhz

drpebcak:
Does this have anything to do with istio? @alena1108 we need a more permanent solution for this instead of just putting more stuff in this file.

loganhz:
Per Bill, we can put them here for now even they are for monitoring.

Issue #22658, opened 2019-09-05
Clusters Stuck in Updating after Installing 2.3.0-rc1 -- argument list too long error


ekristen:
**What kind of request is this (question/bug/enhancement/feature request):**

bug

**Steps to reproduce (least amount of steps as possible):**

Unsure, but I had two existing clusters and installed 2.3.0-rc1

**Result:**

Clusters are stuck in updating with an error message of "argument list too long" exit code 1

**Other details that may be helpful:**

```
2019/09/05 14:59:15 [ERROR] ClusterController local [cluster-deploy] failed with : exit status 1, time="2019-09-05T14:59:13Z" level=fatal msg="argument list too long"

2019/09/05 14:59:19 [ERROR] ClusterController c-m74h6 exit status 1, time="2019-09-05T14:59:17Z" level=fatal msg="argument list too long"
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.3.0-rc1
- Installation option (single install/HA): HA

alena1108:
@ekristen what type of clusters do you have, and what k8s version do they have

soumyalj:
Tested with upgrade from v2.2.8 to v2.3.0-rc1. After upgrade, clusters are stuck in "Provisioning" state with the error below:
```
Exit status 1, time="2019-09-05T23:05:35Z" level=fatal msg="argument list too long"

```
<img width="1350" alt="Screen Shot 2019-09-05 at 3 53 40 PM" src="https://user-images.githubusercontent.com/18536626/64389325-12c0ae80-cff7-11e9-8d88-51b8cee02e44.png">

ekristen:
Baremetal. HA so local is imported on 1.14.5, one custom cluster on 1.14.3

sowmyav27:
Verified on an upgrade of a single node install from `v2.2.8` to `v2.3.0-rc1`
Cluster spec:
- Cluster#1 - 1 etcd, 1 control plane 1 worker node
- Cluster#2 - 3 nodes all-roles.

On an upgrade from `v2.2.8` to `v2.3.0-rc1`, cluster comes up `Active.
All System workloads, are Active.

leodotcloud:
Inspecting the logs there are few more error messages prior to the one mentioned in this issue, which could be of interest:
```
2019/09/06 22:49:54 [INFO] dynamic schema for kontainerdriver amazonelasticcontainerservice updating
[...]

PR #22703, opened 2019-09-07
Use upstream kubectl in server image

ibuildthecloud:
Addresses: https://github.com/rancher/rancher/issues/22658


cjellick:
Does this also control the kubectl running inside the kubectl UI shell?

Issue #19474, opened 2019-04-09
etcd alerts during backups


mattmattox:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- Create 3 node cluster using RKE 0.2.1 using k8s 1.13.5
- Deploy Rancher v2.2.1
- Configure etcd alert in Rancher UI using default
- Wait for scheduled etcd backup

**Result:**
The system component etcd is not running

Alert Name: Etcd is unavailable 
Severity: critical 
Cluster Name: ############# (ID: local) 
Logs: HTTP probe failed with statuscode: 503

**Other details that may be helpful:**
The issue looks to be that the Rancher etcd backups just take a bit longer than it takes for the alert to fire:

https://github.com/rancher/rancher/issues/14630

I got one at 0231 and another at 0831. On Rancher 2.1, I was able to edit the etcd alert and have it send the alert after 10 minutes instead of 3. I don't see a way to do that in 2.2.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.1
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VMware 2 core / 8GB
- Kubernetes version (use `kubectl version`): 1.13.5

gstorme:
You can still edit the tresholds using the API.
Select view in api -> edit -> change the intervals -> show request -> send request.

deniseschannon:
@aiwantaozi did we lose some functionality in etcd alerts with the change from alerts to alert groups in 2.1 to 2.2?

aiwantaozi:
In 2.2.1 UI not support advance option for alert rule level, but you can edit it in API. after 2.2.2 UI support advance option for alert rule level; the latest version also supports inherited advance option from alert group level.

aiwantaozi:
UI added the advance option for alert rule. For the etcd component alert is base on ```kubectl get componentstatus```, kubernetes enhance etcd health check after kubernetes 1.14.0, before kubenetes uses this member list command to check health, now it used health command to check health. It should be better. 
https://github.com/kubernetes/kubernetes/blob/v1.14.0/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go

deniseschannon:
[...]

PR #22709, opened 2019-09-08
Increase etcd system servce alert groupt wait seconds

aiwantaozi:
Problem:
Sometimes get etcd unhealth alert, but may cause by network or disk problem, actually etcd is health

Solution:
Improve group wait seconds to 10 mins as issue mentioned

Issue:
https://github.com/rancher/rancher/issues/19474

### DIFF
--- a/pkg/controllers/user/alert/initdata.go
+++ b/pkg/controllers/user/alert/initdata.go
@@ -138,6 +138,7 @@ func initClusterPreCanAlerts(clusterAlertGroups v3.ClusterAlertGroupInterface, c
 		logrus.Warnf("Failed to create precan rules for %s: %v", name, err)
 	}
 
+	inherited := false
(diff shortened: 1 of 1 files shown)


orangedeng:
LGTM

orangedeng:
I think we need to document the `10 mins` also.

alena1108:
@aiwantaozi @orangedeng will this adjustment be picked up automatically on upgraded Rancher setups having alert turned on prior to upgrade?

aiwantaozi:
@alena1108 This value won't adjust to 10 min when rancher upgraded, a user may adjust to the value they wanted, we can't do reset when the server upgrade.

alena1108:
@aiwantaozi ok, we can document/release note that the change will apply to the new clusters only

Issue #22721, opened 2019-09-09
Unsorted namespace in project logging cause useless configure reload


aiwantaozi:
Steps to reproduce:
1. create project first, add namespace f1,f2,f3,f4 for it
2. create project second, add namespace s1,s2,s3,s4 for it
3. create project third, add namespace t1,t2,t3,t4 for it
4. enable logging for the above projects
5. after logging app is deployed success, check secret
system project -> secret -> rancher-logging-fluentd -> project.conf
may include some configure like this 
```
<regexp>
    key namespace
    pattern (^t3$|^t2$|^t1$|^t4$)
 </regexp>
```
6. change one of the project logging configure, could change the interval, expect other parts not get changed, check the secret after 1min(wait for new configure to generate)
7. above configure also changed, may like this
```
<regexp>
    key namespace
    pattern (^t2$|^t3$|^t1$|^t4$)
 </regexp>
```
8. we expect the namespace is always sorted.

jiaqiluo:
The bug is reproduced in v2.2.8

Notice that before changing the project logging configuration, the pattern looks like this
```
<filter c-zlqrb:p-r8zds.**>
  @type grep
  <regexp>
    key namespace
    pattern (^a2$|^a3$|^a1$|^a4$)
  </regexp>
</filter>
```

after changing the project logging configuration, the pattern looks like this
```
<filter c-zlqrb:p-r8zds.**>
  @type grep
  <regexp>
    key namespace
    pattern (^a4$|^a1$|^a2$|^a3$)
  </regexp>
</filter>
```

jiaqiluo:
The bug fix is validated in rancher:master-head 

now the pattern is in a sorted order and will not change if we edit the project logging's configuration.

```
<filter c-pg4lk:p-lh6k4.**>
  @type grep
  <regexp>
    key namespace
    pattern (^a1$|^a2$|^a3$|^a4$)
  </regexp>
</filter>
```

PR #22722, opened 2019-09-09
sort grep namespace list for project logging

aiwantaozi:
Problem:
unsorted namespace cause useless configure update and reload

Solution:
sort namespace

Issue:
https://github.com/rancher/rancher/issues/22721


orangedeng:
LGTM

PR #22725, opened 2019-09-09
feat(chart): switcher to enable/disable ingress

rootsami:
This pull Request is mainly for Rancher Helm Chart.
Changes: 
- Have a switcher to turn off/on the ingress.
- Define service types. Loadbalancer, NodePort or ClusterIP

Use Cases:
- if you are using IstioGateways and you don't need an Ingress.
- if you need other service type like NodePort or LB using GKE Ingress associated with IAP service.
Just to have a bit of control

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.ingress.enabled }}
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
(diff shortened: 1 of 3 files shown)


rootsami:
@gitlawr @cjellick could you please take a look at this ? 😄
Thanks

juanpmarin:
@rootsami I was looking for how to change the service type beacause gcloud ingresses are not able to be used with ClusterIP services, I hope this get merged!
Thanks

rootsami:
cleanup old PRs

PR #22765, opened 2019-09-10
Note on image mirroring script

cjellick:


cjellick:
don't merge. will break the script. script needs logic to not process comment lines

Issue #22771, opened 2019-09-10
Show system-library in UI


cjellick:
In this bug, we hid the system-library: https://github.com/rancher/rancher/issues/22490

We want to show it for the following reasons:
- if users choose to manage their own mirror for system-charts, they still can. So, it should show in the UI for them
- If something breaks with the system-chart (like cant reach internet), we want to be able to surface that in the UI.

In addition to showing the system-charts catalog, we also want to do the following:
1. At the API level, prevent the deletion of the system-chart. This should do two things:
  - remove the delete link from the system chart catalog so the UI automatically knows it cannot be deleted
  - in a validator or a store, throw a validation error if the user still attempts to delete it (ie via `curl -X DELETE https:///.../system-chart
2. If the user has set the new environment variable that causes rancher to use the embedded version of the chart, then prevent editing the system-chart. This should be done similar to 1 above.

Once this is done, we'll also need to revert the UI PR that hid the chart AND ensure that the UI logic does automatically respect the absence of the delete and edit links so that the user doesnt even see those options

vincent99:
It was hidden because it made absolutely no sense showing up as a custom catalog, and decided nobody needed to turn it off in general because 2.3 bundles the charts with the image anyway.  If you want to show it then it needs its own dedicated control at the top like the other 3 special ones (and it's going to be weird that you specifically can't delete that one, when that is the normal method of overriding the other 3).

cjellick:
Just a couple small notes, though i dont want this to devolve to a GitHub comment argument. I just want to straighten out some things you said:

> ...and decided nobody needed to turn it off in general because 2.3 bundles the charts with the image anyway.

This is opt-in by setting an env-var when spinning up rancher. It is not the default functionality. The default functionality (being able to view and edit) the catalog is the same v2.2

> you don't even want to be able to edit it?

Only when you are in the opt-in env-var mode.

cjellick:
We've merged the backend to basically protect the system-chart catalog from editing or deletion. Logic is this:
- If its the system-chart catalog, we remove the "remove" link from the API and will raise an error if you send an http DELETE to that catalog's URL
- If its the system-cahrt catalog _and_ "bundled" via our new environment variable (used for airgap setups), we don't let you edit it in the same fashion (remove the "update" link and raise an error if you send a PUT)

So, we could do that backend regardless of whether we do the frontend change. I'm kicking over to team/ui to start the ~argument~ healthy debate on whether we should show in UI again.

[...]

Issue #22490, opened 2019-08-28
[UI] Hide system-library from Catalogs page


davidnuzik:
Now that the system-library is a part of the rancher/rancher container there is no longer a need for airgap installs to manually mirror the repo.

As such, there is no longer a need to show this entry on the catalogs page (it's also arguably dangerous as well as it allows deletion). We should hide this entry in the UI.

We just need to be careful with upgrade scenarios. If customer x had manually configured the system-library with catalog url y then the UI needs to detect that it's not a default URL and still show in the UI. 

Otherwise, hide the system-library entry on the Catalogs page if the URL is default.

cc: @sangeethah @jiaqiluo @vincent99

davidnuzik:
Version: master-head (v2.3) (8/29/19)

Now the system-library is hidden unless it's url is not the default (referene https://github.com/n313893254/ui/commit/a22c41f47f77236fbaca6f67dc8dae88e217e272#diff-c9c5fc5063654cc7ab0ca73da2a2945aR53)

On my fresh install the system-library is hidden from the catalog page:

![image](https://user-images.githubusercontent.com/45179589/63973898-93643580-ca60-11e9-8ff7-2e24daabf803.png)

I confirmed on my fresh install system-library apps still work. For example, monitoring still deploys successfully without any issues.

I changed my system-library url via the UI API at /v3/catalogs/system-library
After doing this, the UI now shows the system-library (as expected).
![image](https://user-images.githubusercontent.com/45179589/63974468-ca871680-ca61-11e9-9d71-e5914f610de8.png)

Last I changed it back to the way it was and it's hidden again.

Again, the reasoning for this is that if in an older version someone had an airgap install they may tweak the system-library url. So we should still show it post-upgrade if it's not the default git url.

The UI changes live to hide or unhide the catalog when you change the URL. There is no need to manually refresh, etc.

I didn't identify any issues while testing.

PR #22781, opened 2019-09-10
Disable removing system-catalog

rmweir:
Problem:
Prior, system-catalog was hidden to prevent these actions. This prevented user from viewing errors with system-catalog. Prior to that, user was able to delete system-catalog which could be dangerous.

Solution:
Allow user to view system catalog, but prevent them from deleting it.

Issue:
https://github.com/rancher/rancher/issues/22771

Related Issue:
https://github.com/rancher/rancher/issues/22490

### DIFF
--- a/pkg/api/customization/catalog/catalog.go
+++ b/pkg/api/customization/catalog/catalog.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/coreos/etcd/etcdserver/api/v3rpc/rpctypes"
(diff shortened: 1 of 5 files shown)


pkg/api/customization/catalog/store.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+}
+
+func (s *Store) Delete(apiContext *types.APIContext, schema *types.Schema, id string) (map[string]interface{}, error) {
+	catalog, err := s.ByID(apiContext, schema, id)
+	if err != nil {
+		return nil, err
+	}
+	if catalog["url"] == systemLibraryURL && catalog["name"] == "system-library" {
+		return nil, httperror.NewAPIError(httperror.Conflict, fmt.Sprint("[catalog store] Unathorized to delete system-library catalog"))
```
dramich: The error should be a `InvalidBodyContent`, `conflict` would most likely apply to a PUT request. The message itself should be updated to `not allowed to delete system-library`

pkg/api/customization/catalog/catalog.go
```diff
@@ -18,9 +18,16 @@ import (
...
 )
 
+const (
+	systemLibraryURL = "https://git.rancher.io/system-charts"
+)
+
 func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 	resource.AddAction(apiContext, "refresh")
 	resource.Links["exportYaml"] = apiContext.URLBuilder.Link("exportYaml", resource)
+	if resource.Values["url"] == systemLibraryURL && resource.Values["name"] == "system-library" {
```
dramich: Turn `"system-library" ` into a const as well.

pkg/api/customization/catalog/catalog.go
```diff
@@ -15,12 +15,36 @@ import (
...
 
-func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
+const (
+	SystemLibraryURL            = "https://git.rancher.io/system-charts"
+	SystemCatalogName           = "system-library"
+	embededSystemCatalogSetting = "system-catalog"
+)
+
+type Wrapper struct {
+	SettingLister v3.SettingLister
```
dramich: Use the settings package for this.

pkg/api/customization/catalog/validator.go
```diff
@@ -15,7 +16,17 @@ var (
 	controlEncoded = regexp.MustCompile("%[0-1][0-9,a-f,A-F]")
 )
 
-func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if request.Method == http.MethodPut {
```
dramich: Let's move this into a store update method then use a helper so this logic isn't duplicated

riaan53:
Will we still be able to edit the chart for air gap deployments? Currently we just delete and add the system-chart (in automated way) again with new values.

dramich:
@riaan53 we have bundled in the system-charts for air gap, so providing a flag to rancher will use that version vs having to get something external. This change does lock down system-charts so it can't be deleted. https://github.com/rancher/rancher/issues/20387

dramich:
lgtm

Issue #19560, opened 2019-04-11
Investigate if listenConfig changes need to be sent to the UI


cloudnautique:
ListenConfig updates are sent to the UI every 2 seconds and are approximately 32k a piece.

cjellick:
incomplete solution: https://github.com/rancher/rancher/pull/21979/files

vincent99:
My phone's tethering battery life thanks you for your service.

PR #22783, opened 2019-09-10
WIP: Reconcile certs in dynamicListener correctly

cbron:
Issue: #19560
Follow up to original WIP PR here: https://github.com/rancher/rancher/pull/21979

## Logic

Rather than overwriting all keys in any situation, only ever write keys that match this servers mapKey. 

## Testing

I tested both with and without this fix on separate double-server HA setups running 3 rancher pods each. In examples below I'm just using IP's for server-url and `localhost` as hostname, so those get used for keys.

* Login to both IP's via browser, open network tab, find websocket like: `wss://IP_ADDRESS:PORT/v3/subscribe?sockId=N`
* View messages on socket, filtering to those with baseType of "listenConfig"

Without the fix you can indeed see the waffling `listenconfigs/generatedCerts` node. At first you get entries for: 
```
Localhost:                                       
Local / 10.42.0.3:
Local /[ 10.42.0.3   45.79.35.48   45.79.59.176 ]: 
Local /[ 10.42.0.3   45.79.59.176 ]:                      
```
_format here is `[ pod-ip  node-ip  node-ip]`_

then a few seconds later entirely different pod IP's and no localhost line either: 
```
Local / 10.42.1.7:
Local /[ 10.42.1.7   45.79.35.48   45.79.59.176 ]:
Local /[ 10.42.1.7   45.79.59.176 ]:
```

With fix you get this, and no spamming in the UI's websocket:
```
Localhost:
Local / 10.42.0.10:  
Local / 10.42.1.10:  
Local / 10.42.1.9:   
Local /[ 10.42.0.10   45.79.51.199   45.79.52.11 ]:
Local /[ 10.42.0.10   45.79.51.199 ]:                 
Local /[ 10.42.1.10   45.79.51.199   45.79.52.11 ]:
Local /[ 10.42.1.10   45.79.51.199 ]:                  
Local /[ 10.42.1.9   45.79.51.199   45.79.52.11 ]: 
Local /[ 10.42.1.9   45.79.51.199 ]:                    
```

### DIFF
--- a/pkg/dynamiclistener/server.go
+++ b/pkg/dynamiclistener/server.go
@@ -13,7 +13,6 @@ import (
 	"log"
 	"net"
 	"net/http"
-	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -88,6 +87,7 @@ type Server struct {
 	handler             http.Handler
 	httpPort, httpsPort int
 	certs               map[string]*tls.Certificate
+	certKey             string
 	ips                 *lru.Cache
 
 	listeners    []net.Listener
@@ -130,7 +130,7 @@ func (s *Server) updateIPs(savedIPs map[string]bool) map[string]bool {
 		return savedIPs
 	}
 
-	err = s.reconcileCerts(cfg)
+	err = s.reconcileGeneratedCert(cfg)
 	if err != nil {
 		return savedIPs
 	}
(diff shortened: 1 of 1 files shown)


cbron:
Found an issue with this logic, will reopen when fixed.

pkg/dynamiclistener/server.go
```diff
@@ -130,20 +130,9 @@ func (s *Server) updateIPs(savedIPs map[string]bool) map[string]bool {
...
-	if !reflect.DeepEqual(certs, cfg.GeneratedCerts) {
-		cfg = cfg.DeepCopy()
-		cfg.GeneratedCerts = certs
-		cfg, err = s.listenConfigStorage.Update(cfg)
-		if err != nil {
-			return savedIPs
-		}
+	err = s.reconcileGeneratedCert(cfg)
+	if err != nil {
+		return savedIPs
```
cbron: This is done only as a backup, in case the original save fails. If we want to simplify we can just do the save on generation below.

pkg/dynamiclistener/server.go
```diff
@@ -173,6 +162,31 @@ func (s *Server) updateIPs(savedIPs map[string]bool) map[string]bool {
 	return allIPs
 }
 
+func (s *Server) reconcileGeneratedCert(cfg *v3.ListenConfig) error {
```
cbron: This is done only as a backup, in case the original save fails. If we want to simplify we can just do the save on generation below.

cbron:
This appears to be rewritten.

Issue #22644, opened 2019-09-04
Image lists are different when downloading from Rancher and Github Release Page


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug, `Rancher v2.3.0-rc1`

**Steps to reproduce (least amount of steps as possible):**
The image lists are different between the file `rancher-windows-images.txt` downloaded from Rancher UI and the file `rancher-windows-images.txt` downloaded from Github release page: https://github.com/rancher/rancher/releases/tag/v2.3.0-rc1

The image lists for Linux also do NOT match each other.

rmweir:
Would this be because the image was using the default system chart environment variable from package/Dockerfile which may have been more up to date than the one set and used in scripts/
package? If so this may be resolved by https://github.com/rancher/rancher/pull/22734

mrajashree:
@cjellick @rmweir This doesn't look like the same issue. The windows and linux images downloaded from rancher UI don't have rancher-agent images. Also the system tools images aren't included in the list from UI download link, whereas they are included while generating the list for the command from package script https://github.com/rancher/rancher/blob/master/pkg/image/export/main.go#L82

thxCode:
@jiaqiluo I think `rancher-windows-images.txt` should only keep the images using on Windows nodes, the following list includes the required images:
```
- from system images
  + rancher/hyperkube
  + rancher/rke-tools
  + rancher/kubelet-pause
  + rancher/rke-tools
  + rancher/rancher-agent
- from system chart images
  + logging:
    - rancher/fluentd
    - rancher/configmap-reload
  + monitoring:
    - rancher/wmi_exporter-package 
```

@mrajashree , I think we could address this issue and https://github.com/rancher/rancher/issues/22463 together in PR https://github.com/rancher/rancher/pull/22785, FYI:
- https://github.com/rancher/rancher/pull/22821 would drop the stale tool images for logging and alerting
- https://github.com/rancher/system-charts/pull/96 adds a new mark to indicate those system chart images using on Windows

jiaqiluo:
The bug fix is validated in Rancher:v2.3.0-rc4

The content of `rancher-windows-images.txt` from the UI and the Github release page are identical. 

the `rancher-linux-images.txt` from the Github release page has `rancher/rancher:v2.3.0-rc4` but the one from the UI does not.

It makes sense because if you can download the file from Rancher UI you should already have the `rancher/rancher` image.

PR #22785, opened 2019-09-10
Make image list from UI match release page list

mrajashree:
https://github.com/rancher/rancher/issues/22644

Types PR: https://github.com/rancher/types/pull/974

### DIFF
--- a/pkg/image/resolve.go
+++ b/pkg/image/resolve.go
@@ -1,13 +1,25 @@
 package image
 
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
 	"path"
+	"path/filepath"
 	"sort"
 	"strings"
 
 	"github.com/rancher/norman/types/convert"
+	libhelm "github.com/rancher/rancher/pkg/catalog/helm"
 	"github.com/rancher/rancher/pkg/settings"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	img "github.com/rancher/types/image"
+	"gopkg.in/yaml.v2"
 )
+
+var requiredImagesNotInSystemCharts = []string{
+	"busybox",
+}
 
 func Resolve(image string) string {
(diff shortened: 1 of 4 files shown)


pkg/image/resolve.go
```diff
@@ -44,6 +56,34 @@ func CollectionImages(objs ...interface{}) ([]string, error) {
...
+	if err != nil {
+		return []string{}, err
+	}
+	for _, i := range images {
+		targetImages = append(targetImages, img.Mirror(i))
+	}
+	return targetImages, nil
+}
+
+func GetWindowsImages(rkeSystemImages map[string]v3.RKESystemImages) ([]string, error) {
```
thxCode: We need to add the logic of getting windows images from charts in here.

pkg/image/resolve.go
```diff
@@ -54,3 +94,98 @@ func findStrings(obj map[string]interface{}, found map[string]bool) {
 		}
 	}
 }
+
+func getImagesFromCharts(path string) ([]string, error) {
```
thxCode: We could add another method or enhance this one to support get windows images from charts.

pkg/image/resolve.go
```diff
@@ -54,3 +94,98 @@ func findStrings(obj map[string]interface{}, found map[string]bool) {
...
+	if !repoOk || !tagOk {
+		return
+	}
+	repo, repoOk := r.(string)
+	tag, tagOk := t.(string)
+	if !repoOk || !tagOk {
+		return
+	}
+
+	output[fmt.Sprintf("%s:%s", repo, tag)] = struct{}{}
```
thxCode: Based on https://github.com/rancher/system-charts/pull/96, we could use the similar logic to judge the image:

```
os, osOk := inputMap["os"]
if osOk && os == "windows" {
  // should be a Windows image
} else {
  // should be a Linux image
}

```

PR #22792, opened 2019-09-11
Improve linux alert expression

aiwantaozi:
**Problem:**
`node_filesystem_files_free` returns the free inodes numbers, not free disk size.
**Solution:**
Using `node_filesystem_free_bytes` instead of `node_filesystem_files_free`.
**Issue**:
https://github.com/rancher/rancher/issues/22680


orangedeng:
LGTM

moschlar:
Does this get updated in existing alert rules for existing clusters automatically?

miraculixx:
@moschlar it does not get updated automatically afaict, but you can easily change it yourself: 

image

Issue #22676, opened 2019-09-06
[Windows] Windows nodes fail to upgrade k8s version from 1.14 to 1.15


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- On master-head - latest have ui-index: `https://releases.rancher.com/ui/2.3.3/index.html`
- Bring up a windows cluster using k8s `v1.14.6`
- Cluster spec:

```
1 linux - etcd and control plane
1 linux worker node
2 windows worker node
```

- Deploy workloads and ingresses.
- The windows nodes are registered only with internal addresses, even though while registering both the ips - internal and external addresses were given.
- Upgrade k8s version to `v1.15.3`

**Expected Result:**
All the nodes - Linux and windows nodes, the k8s version should upgrade. And the cluster should come up Active.

**Actual Result:**
The k8s version on the windows nodes do not change.
<img width="1508" alt="Screen Shot 2019-09-05 at 6 19 11 PM" src="https://user-images.githubusercontent.com/26032343/64394410-ea42af80-d00a-11e9-95a2-78db76d23676.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher-master- commit id: `f4b720b92`
- Installation option (single install/HA): single

loganhz:
We already have an issue for this https://github.com/rancher/rancher/issues/22599 .
I will close that one.

loganhz:
Maybe we can try to upgrade it from 1.15 to 1.16

It will be a bit slow as windows images is kind of large

sowmyav27:
Verified on `rancher:v2.3.0-rc5`.
Steps:
on rancher:v2.3.0-rc5, deploy a windows custom cluster with the following specs.
```
1 etcd/control plane node - 18.04 LTS Linux
1 worker node - 18.04 LTS Linux
2 worker nodes - Windows_Server-1903-English-Core-Base-2019.08.16
k8s - 1.15.3
```
Deploy some workloads - demonset, nodeport.
Verified they work.
Deploy ingress pointing to a workload. And verified it works.
Upgrade the k8s version to `v1.16.0-beta.1`.
You will notice the cluster go into Updating state and then Active. At this stage, the Linux control/etcd node's k8s would have upgraded to `v1.16.0-beta.1`

PR #22795, opened 2019-09-11
Update wins to support upgrading

thxCode:
**Problem:**
- Could not upgrade Kubernetes
- Could not upgrade wins

**Solution:**
- Upgrade wins version
- Adjust wins configuration

**Related PR:**
- https://github.com/rancher/rke/pull/1634
- https://github.com/rancher/rke-tools/pull/81

**Issue:**
#22676


orangedeng:
LGTM

package/windows/utils.psm1
```diff
@@ -156,6 +156,25 @@ function Invoke-HNSRequest
...
+    if (Test-Path -PathType leaf -Path $Dst) {
+        $dstHasher = Get-FileHash -Path $Dst
+        $srcHasher = Get-FileHash -Path $Src
+        if ($dstHasher.Hash -eq $srcHasher.Hash) {
+            return $false
+        }
+    }
+
+    $null = Copy-Item -Force -Path $Src -Destination $Dst
+    return $true
```
alena1108: don't know much about windows; do we return the actual result of the command execution? Would false be returned, and error logged, if the command had failed?

package/windows/Dockerfile.agent
```diff
@@ -12,7 +12,7 @@ RUN $URL = 'https://github.com/StefanScherer/docker-cli-builder/releases/downloa
     \
     Write-Host 'Complete.'
 # download wins
-RUN $URL = 'https://github.com/thxCode/wins/releases/download/v0.0.11/wins.exe'; \
+RUN $URL = 'https://github.com/thxCode/wins/releases/download/v0.1.0/wins.exe'; \
```
deniseschannon: Should this be moved from https://github.com/thxCode/wins to https://github.com/rancher/wins? And where is this hosted/being built?

package/windows/utils.psm1
thxCode: Calling `Copy-Item` without `-ErrorAction Ignore`, if it has failed(for example, the file handler of the destination file is using by another process), it will throw an error.

I will change here to wrap by a `try...catch...` block, then output some useful message.

Issue #22797, opened 2019-09-11
Logging couldn't delete in PSP restricted cluster


aiwantaozi:
Steps:
1. Enable logging in a PSP restricted cluster. Check that logging app is deployed but no fluentd pods are deployed due to PSP limitation.
2. Disable logging, and notice that the logging app is not removed.

Related issue: https://github.com/rancher/rancher/issues/22593

soumyalj:
Reproduced the issue with master-head build from Sep 11th. 
Tested with master-head build from Sep 13th (ca1fd72a34e5)

Enabled fluentd logging in a PSP enabled setup. Namespace cattle-logging and the fluentd workloads are shown in UI without any active pods as expected.

<img width="1256" alt="Screen Shot 2019-09-13 at 3 40 54 PM" src="https://user-images.githubusercontent.com/18536626/64898704-0662d400-d63d-11e9-90f4-afb568b7cd7d.png">

Disable  fluentd logging. Cattle-logging namespace and fluentd related containers get deleted as expected.

PR #22798, opened 2019-09-11
Improve logic for checking logging app deploy success

aiwantaozi:
Problem:
Enable logging in a PSP restricted cluster. Check that logging app is deployed but no fluentd pods are deployed due to PSP limitation, disable logging, and notice that the logging app is not removed.

Solution:
This is cause by wrong use of rancher.ticker, when the timeout, the ticker get stop, but the ticker channel isn't close cause loop

Issue:
https://github.com/rancher/rancher/issues/22797


orangedeng:
LGTM

Issue #22835, opened 2019-09-12
rancher_max_version set for older versions of monitoring and istio is not being applied


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
- Deploy a cluster in a new Rancher setup of `v2.3.0-rc3`

**Expected Result:**
Monitoring and Istio should show only the valid chart versions.

**Actual Result:**
When you navigate to cluster --> Tools --> Monitoring or Istio, all the 4 (Monitoring) / 2 (Istio) chart versions are seen.

**Note:**

This does not happen always.

Leave the setup around for sometime. DO NOT do a hard refresh of the page. Then again navigate to cluster --> Tools --> Monitoring or Istio. All the charts versions will be displayed..

ON doing a hard refresh of the page, only 1 chart version is displayed for both monitoring and Istio.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `v2.3.0-rc3`
- Installation option (single install/HA): single

dramich:
After discussing this with @vincent99 this should be a change in the UI to filter out the versions not compatible with that version of rancher.

vincent99:
The problem is there's no connection between asking for templates within this range in the initial list call and what comes down the socket.  So rather than having the server do that filtering (half the time) I suggested the UI just get them all and decide which to show on its own.

But that doesn't actually work, each templateVersion has a different valid range and the reason the server filters it is so that we can load just /v3/templates and get back the set of valid versionLinks for that range without having to load the entire (huge) set of /v3/templateVersions (or making n+1 calls to verify individual versions)

Soo... I think we need to go back to (stay on) server filtering:

- `/v3/templates` should only return `versionLinks` for `versions` which are satisfied by the current Rancher version.  You can add a secret `?all` flag if you want but I don't think there's any use case _requiring_ it.

- Drop/ignore the query param that the UI uses to pass in the server version.  You already know what version you are. (this is historical baggage from catalog being an independent microservice in 1.x)

- Suppress change events for any templateVersion that does not satisfy the current server version.

- /v3/templateVersions and /v3/templateVersions/:id should still return ALL the versions (so that if you have an App that already exists the UI can get info about the version and its upgrade links, even if it's now incompatible with the running Rancher version)

sowmyav27:
verified on `rancher:v2.3.0-rc4`.

[...]

PR #22863, opened 2019-09-13
Update template to use a store

dramich:
Problem:
Templates need their versionsLinks updated based off the rancher server
for both list and watch calls

Solution:
Add a Transformer to add the links and move the logic out of a
formatter

https://github.com/rancher/rancher/issues/22835

### DIFF
--- a/pkg/api/customization/catalog/store.go
+++ b/pkg/api/customization/catalog/store.go
@@ -1,0 +1,97 @@
+package catalog
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/rancher/norman/store/proxy"
+	"github.com/rancher/norman/store/transform"
+	"github.com/rancher/norman/types"
+	catUtil "github.com/rancher/rancher/pkg/catalog/utils"
+	hcommon "github.com/rancher/rancher/pkg/controllers/user/helm/common"
+	"github.com/rancher/rancher/pkg/settings"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	managementschema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
+	client "github.com/rancher/types/client/management/v3"
+	"github.com/rancher/types/config"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/api/customization/catalog/store.go
```diff
@@ -0,0 +1,97 @@
...
+	ts.Store = s
+
+	return ts
+}
+
+func (t *templateStore) extractVersionLinks(apiContext *types.APIContext, resource map[string]interface{}) map[string]interface{} {
+	schema := apiContext.Schemas.Schema(&managementschema.Version, client.TemplateVersionType)
+	r := map[string]interface{}{}
+	versionMap, ok := resource[client.CatalogTemplateFieldVersions].([]interface{})
+	if ok {
```
daxmc99: Is it ok if we return an empty map here?

cbron:
Commits should probably be squashed

tests/integration/suite/test_catalog.py
```diff
@@ -177,13 +177,6 @@ def test_template_version_links(admin_mc, admin_pc, custom_catalog,
 
     assert len(templates.data[0]['versionLinks']) == 0
 
-    # If no rancher version is set get back both versions
-    templates = client.list_template(catalogId=c_name)
-
-    assert len(templates.data[0]['versionLinks']) == 2
-    assert '1.6.0' in templates.data[0]['versionLinks']
-    assert '1.6.2' in templates.data[0]['versionLinks']
-
```
cbron: We could create a test here with an out-of-scope version right ?

pkg/api/customization/catalog/store.go
dramich: Yes, that's the same as having no versions matching your current rancher version.

tests/integration/suite/test_catalog.py
dramich: There is above this.

daxmc99:
LGTM

Issue #22836, opened 2019-09-12
Failed to prepare the private registry for both Windows and Linux images 


jiaqiluo:
BUG

I run a private registry in a Ubuntu box, tag and push the image `rancher/rancher-agent:v2.3.0-rc1` to it, now I run a windows box and  pull the same image, then the error occurs
```
PS C:\Users\Administrator> docker pull my-registry/rancher/rancher-agent:v2.3.0-rc1
v2.3.0-rc1: Pulling from rancher/rancher-agent
image operating system "linux" cannot be used on this platform
```

Cause:

The current `rancher-load-images.sh`, `rancher-save-images.sh` and the documentation does not provide the method to prepare the private registry for both Windows and Linux.

drpebcak:
This also needs to be fixed for `arm64`.

cloudnautique:
Procedurally, I'm onboard, it should look something like this:

Input file:
`rancher/image:version amd64;remote/repo:version windows;remote/repo:version arm;remote/repo:version`

Ideally only the first field is positional. 

Windows script would be run first. 
rancher-load-images.ps1 and rancher-save-images.ps1
It would use the input file and only process items that `windows;` items. Is it reasonable to assume that the version string will include the OS specifics? if so, would we tag that with `rancher/image:version-1903` for instance?

On linux:
Load both linux and arm.
Rancher save would push.

Then a 3rd script for manifesting... or flags on the linux would run with the ability to --skip-windows, --skip-armhf, or --skip-skip-arm64. Armhf might not be needed. I don't think there is any harm in manifesting things that don't have multiple archs/OSes.

drpebcak:
Yeah so the format that I think we agreed on was like this (for a somewhat concrete example, although I don't think etcd actually supports all of these versions):
`rancher/coreos-etcd:v1 amd64;coreos/etcd:v1 arm;coreos/etcd:v1-arm arm64;coreos/etcd:v1-arm64 windows-1903;coreos/etcd:v1-windows-1903`
and we will have the following tags in our org afterwards:
```
rancher/coreos-etcd:v1
rancher/coreos-etcd:v1-arm
rancher/coreos-etcd:v1-arm64
rancher/coreos-etcd:v1-windows-1903
```

So then from the user side, they can use this same file to pull these for each architecture (running from a computer that matches that arch).

If there is a windows entry on a line, pull the image like `rancher/coreos-etcd:v1-windows-1903`
[...]

Issue #22463, opened 2019-08-28
image for window is in the rancher-images.txt for linux


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- run Rancher:master-head 0f08172
- go to `v3/kontainerdrivers/rancher-images` to download the file `rancher-images.txt`

**Result:**
- the image `rancher/kubelet-pause:v0.1.3` is in the list, but it is not for Linux platform
```
root@localhost:~# docker pull rancher/kubelet-pause:v0.1.3
v0.1.3: Pulling from rancher/kubelet-pause
no matching manifest for linux/amd64 in the manifest list entries
```

We should not include this image if Rancher is running in a Linux environment.

jiaqiluo:
The bug fix is validated in rancher:master-head a98b70f17

Now the image `rancher/kubelet-pause:v0.1.3` is not in the downloaded `rancher-images.txt` file

jiaqiluo:
Reopen the issue because this image  is shown in the rancher-image.txt file downloaded from Github release page for v2.3.0-rc1 
https://github.com/rancher/rancher/releases/tag/v2.3.0-rc1

kinarashah:
This is getting fixed by https://github.com/rancher/rancher/pull/22443

jiaqiluo:
in v2.2.3-rc2

```
Error response from daemon: No such image: rancher/kubelet-pause:v0.1.3
Error response from daemon: No such image: rancher/wmi_exporter-package:v0.0.2
```

loganhz:
We are using manifest images for Windows.

`rancher/hyperkube:v1.15.3-rancher3` is a manifest 

In a Windows cluster, both linux node and windows node will pull it from the private registry. If we don't support manifest, we cannot do the air-gap install for Windows cluster.

I think we should delete `rancher-windows-images.txt`,  put all images into `rancher-images.txt` and support manifest images in `rancher-save-images.sh` and `rancher-load-images.sh`.

cc @cloudnautique @drpebcak

loganhz:
Can be validated in next RC

jiaqiluo:
The bug fix is validated in v2.3.0-rc5 that there are no Windows images in the `rancher-images.txt` file

PR #22880, opened 2019-09-16
Correct the image-list via OS type

thxCode:
**Problem:**
We can gain the wrong OS type images from image-list file.

**Solution:**
Adjust the logic to distinguish the images via os

**Issue:**
https://github.com/rancher/rancher/issues/22463
https://github.com/rancher/rancher/issues/22836


pkg/image/resolve_test.go
```diff
@@ -0,0 +1,369 @@
...
+				assert.NotContains(images, nc, cs.tName)
+			}
+		}
+	}
+}
+
+func TestAppendImagesFromRequirement(t *testing.T) {
+	testCases := []struct {
+		tName             string
+		iImages           []string
```
orangedeng: You don't have any test cases including `iImages` with content. Is that expected?

pkg/image/resolve_test.go
```diff
@@ -0,0 +1,369 @@
...
+				"prom/prometheus:v2.0.1",
+				"quay.io/coreos/flannel:v1.2.3",
+				"gcr.io/google_containers/kube-pause:v0.0.1",
+				"test.io/test:v0.0.1", // not in mirror list
+			},
+			oShouldContain: []string{
+				"rancher/coreos-flannel:v1.2.3",
+				"test.io/test:v0.0.1",
+			},
+			oShouldNotContain: []string{
```
orangedeng: nit, I think `oShouldEqual` would be enough. The `mirror`, `unique` and `sort` should work together and the output the a raw image list should be explicit.

pkg/image/resolve.go
```diff
@@ -188,3 +132,126 @@ func walkthroughMap(inputMap map[interface{}]interface{}, walkFunc func(map[inte
...
+func normalizeImages(rawImages []string) (images []string) {
+	// mirror
+	var m = img.Mirror
+	for _, image := range rawImages {
+		images = append(images, m(image))
+	}
+
+	// unique
+	var uImages = make([]string, 0, len(images))
+	var imagesSet = map[string]struct{}{}
```
mrajashree: could we use map[string]bool instead?

pkg/image/resolve.go
```diff
@@ -188,3 +132,126 @@ func walkthroughMap(inputMap map[interface{}]interface{}, walkFunc func(map[inte
...
+
+func appendImagesFromRequirement(images []string, osType OSType) []string {
+	switch osType {
+	case Linux:
+		images = append(images, linuxRequiredImagesNotInSystemCharts...)
+	}
+	return images
+}
+
+func normalizeImages(rawImages []string) (images []string) {
```
mrajashree: nit: maybe we don't need named return value here since we're as it is declaring uImages and imagesSet within this method?

pkg/image/resolve.go
```diff
@@ -134,7 +63,11 @@ func getChartAndVersion(path string) (map[string]string, error) {
 	return rtn, nil
 }
 
-func walkFunc(images map[string]struct{}, versions map[string]string, basePath, path string, info os.FileInfo, err error) error {
+func pickImagesFromValuesYAML(imagesSet map[string]struct{}, versions map[string]string, basePath, path string, info os.FileInfo, err error, osType OSType) error {
```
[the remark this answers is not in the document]
thxCode: Updated, Error handling placed outside.

pkg/image/resolve_test.go
```diff
@@ -0,0 +1,369 @@
...
+	assertlib "github.com/stretchr/testify/assert"
+)
+
+const rancherVersion = "v1.15.3-rancher3-1"
+
+func TestFetchImagesFromCharts(t *testing.T) {
+	systemChartPath := cloneSystemChart(t)
+
+	testCases := []struct {
+		tName             string
```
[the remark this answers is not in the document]
thxCode: Updated

pkg/image/resolve_test.go
thxCode: Using `getRequirementImages` instead of `appendImagesFromRequirement`, so remove this case.

pkg/image/resolve_test.go
thxCode: Updated

pkg/image/resolve.go
thxCode: Updated

pkg/image/resolve.go
thxCode: Updated

pkg/image/export/main.go
```diff
@@ -344,48 +338,200 @@ echo "Creating ${images} with $(echo ${pulled} | wc -w | tr -d '[:space:]') imag
...
-	.PARAMETER image-list
-	text file with list of images. 1 per line.
-	.PARAMETER images
-	tar.gz generated by docker save.
-#>
-param(
-	[string]${image-list}="rancher-windows-images.txt",
-	[string]$images="rancher-windows-images.tar.gz"
-)
+	windowsSaveScirpt = `$ErrorActionPreference = 'Stop'
```
mrajashree: nit: typo, should be `windowsSaveScript`?

thxCode: Updated

thxCode:
@mrajashree , rebased and updated.

pkg/image/export/main.go
```diff
@@ -344,48 +338,200 @@ echo "Creating ${images} with $(echo ${pulled} | wc -w | tr -d '[:space:]') imag
...
+
+$script_name = $MyInvocation.InvocationName
+$image_list = "rancher-windows-images.txt"
+$images = "rancher-windows-images.tar.gz"
+$os_release_id = $(Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\' | Select-Object -ExpandProperty ReleaseId)
+$help = $false
+
+function usage {
+    echo "USAGE: $script_name [--image-list rancher-windows-images.txt] [--images rancher-windows-images.tar.gz]"
+    echo "  [-l|--image-list path] text file with list of images. 1 per line."
```
alena1108: Nit: lets change to "text file with list of images; one image per line"

pkg/image/export/main.go
```diff
@@ -344,48 +338,200 @@ echo "Creating ${images} with $(echo ${pulled} | wc -w | tr -d '[:space:]') imag
...
+    echo "  [-i|--images path] tar.gz generated by docker save."
+    echo "  [-o|--os-release-id (1809|1903|...)] release id of OS, could detect automatically."
```
alena1108: `release id of OS, gets detected automatically if not passed."`

pkg/image/export/main.go
```diff
@@ -344,48 +338,200 @@ echo "Creating ${images} with $(echo ${pulled} | wc -w | tr -d '[:space:]') imag
...
+if (-not $registry)
+{
+	echo "Must input registry address !"
+    usage
+    exit 1
+}
+
+if (-not (Test-Path $images))
+{
+	echo "Could not find '$images' tar.gz !"
```
alena1108: do we need to have a space after '$images'?

pkg/image/export/main.go
```diff
@@ -344,48 +338,200 @@ echo "Creating ${images} with $(echo ${pulled} | wc -w | tr -d '[:space:]') imag
...
-    docker push $Registry/$item
+if ($help)
+{
+    usage
+    exit 0
 }
+
+if (-not $registry)
+{
+	echo "Must input registry address !"
```
alena1108: nit: "Registry address is required"

pkg/image/export/main.go
alena1108: also no need for explanation mark

pkg/image/export/main.go
alena1108: Once changed, should be done in usage command as well: https://github.com/rancher/rancher/pull/22880/files#diff-6bea58749a943397efd8c4a9b0e75351R350

pkg/image/export/main.go
```diff
@@ -344,48 +338,200 @@ echo "Creating ${images} with $(echo ${pulled} | wc -w | tr -d '[:space:]') imag
...
+
+if ($help)
+{
+    usage
+    exit 0
+}
+
+if (-not (Test-Path $image_list))
+{
+	echo "Could not find '$image_list' text file !"
```
alena1108: remove " !". Replace the message with `Could not find '$image_list' file`

Issue #22870, opened 2019-09-14
Custom cluster fails to deploy with GetSvcOptions: no service-option key present for v1.13.10-rancher1-3" error


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a custom cluster using Kubernetes version `1.13.10-rancher1-3` and default options.
2. Run agent on node with worker, controlplane, and etcd flags.

**Result:**
Cluster does not provision. See error `GetSvcOptions: no service-option key present for v1.13.10-rancher1-3` on UI and cluster is stuck in `Provisioning` state.

**Other details that may be helpful:**
Seems to work fine for other K8s versions, just busted for v1.13.

Logs from Rancher server:
```
2019/09/14 21:58:46 [INFO] [mgmt-cluster-rbac-delete] Creating namespace c-t6g29
2019/09/14 21:58:46 [INFO] [mgmt-cluster-rbac-delete] Creating Default project for cluster c-t6g29
2019/09/14 21:58:46 [WARNING] error updating ns c-t6g29 status: Operation cannot be fulfilled on namespaces "c-t6g29": the object has been modified; please apply your changes to the latest version and try again
2019/09/14 21:58:46 [INFO] [mgmt-cluster-rbac-delete] Creating System project for cluster c-t6g29
2019/09/14 21:58:46 [INFO] [mgmt-project-rbac-create] Creating namespace p-btf8n
2019/09/14 21:58:46 [INFO] [mgmt-cluster-rbac-delete] Updating cluster c-t6g29
2019/09/14 21:58:46 [INFO] [mgmt-project-rbac-create] Creating namespace p-5fbp9
2019/09/14 21:58:46 [INFO] [mgmt-project-rbac-create] Creating creator projectRoleTemplateBinding for user user-2jjcn for project p-btf8n
2019/09/14 21:58:46 [INFO] [mgmt-project-rbac-create] Creating creator projectRoleTemplateBinding for user user-2jjcn for project p-5fbp9
2019/09/14 21:58:46 [INFO] [mgmt-auth-crtb-controller] Setting InitialRolesPopulated condition on project p-btf8n
2019/09/14 21:58:46 [INFO] [mgmt-auth-prtb-controller] Creating clusterRole p-btf8n-projectowner
[...]

PR #22893, opened 2019-09-16
Fix missing svcOptions for windows nodes

moelsayed:
https://github.com/rancher/rancher/issues/22870

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/data_getter.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/data_getter.go
@@ -91,7 +91,12 @@ func GetRKEK8sServiceOptions(k8sVersion string, svcOptionLister v3.RKEK8sService
 		key = svcOptionWindowsKey
 	}
 	val, ok := sysImage.Labels[key]
-	if !ok {
+	// It's possible that we have a k8s version with no windows svcOptions. In this case, we just warn and return nil.
+	// if we have in fact windows nodes trying to use that version, the error will show in reknodeconfig server.
+	if !ok && osType == Windows {
+		logrus.Debugf("getSvcOptions: no service-option key present for %s", k8sVersion)
+		return k8sSvcOption, nil
+	} else if !ok {
 		return k8sSvcOption, fmt.Errorf("getSvcOptions: no service-option key present for %s", k8sVersion)
 	}
 	return getRKEServiceOption(val, svcOptionLister, svcOptions)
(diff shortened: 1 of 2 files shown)


pkg/rkenodeconfigserver/nodeserver.go
```diff
@@ -344,13 +346,15 @@ func (n *RKENodeConfigServer) getServiceOptions(k8sVersion string) (map[string]i
...
-		data["k8s-windows-service-options"] = svcOptionsWindows
+	if osType == "windows" {
+		svcOptionsWindows, err := kd.GetRKEK8sServiceOptions(k8sVersion, n.serviceOptionsLister, n.serviceOptions, n.sysImagesLister, n.sysImages, kd.Windows)
+		if err != nil {
+			logrus.Errorf("getK8sServiceOptionsWindows: k8sVersion %s [%v]", k8sVersion, err)
+			return data, err
+		}
+		if svcOptionsWindows != nil {
+			data["k8s-windows-service-options"] = svcOptionsWindows
+		}
```
leodotcloud: What happens if `svcOptionsWindows` is nil?

moelsayed: service options gets sent to GeneratePlan() which uses it in two places, both check if it's nil/not nil and handle it:
https://github.com/rancher/rancher/blob/master/vendor/github.com/rancher/rke/cluster/plan.go#L63-L66
https://github.com/rancher/rancher/blob/master/vendor/github.com/rancher/rke/cluster/plan.go#L999

pkg/controllers/management/kontainerdrivermetadata/data_getter.go
```diff
@@ -91,7 +91,12 @@ func GetRKEK8sServiceOptions(k8sVersion string, svcOptionLister v3.RKEK8sService
 		key = svcOptionWindowsKey
 	}
 	val, ok := sysImage.Labels[key]
-	if !ok {
+	// It's possible that we have a k8s version with no windows svcOptions. In this case, we just warn and return nil.
+	// if we have in fact windows nodes trying to use that version, the error will show in reknodeconfig server.
+	if !ok && osType == Windows {
+		logrus.Warnf("getSvcOptions: no service-option key present for %s", k8sVersion)
```
kinarashah: can we log this in debug? don't want users to be alarmed because of node config calling this very frequently 

nit: can do osType check in `if !ok` block

Issue #22895, opened 2019-09-16
Hide 'unsupported' storage providers and provisioners


tfiduccia:
**What kind of request is this (question/bug/enhancement/feature request):**
Hide 'unsupported' storage providers and provisioners

Pages affected:
Persistent Volume
Storage Class
Add Workload -> Volume -> Ephemeral

vincent99:
New feature flag to look at:  https://github.com/rancher/rancher/pull/22899

If false, then hide unsupported options for those types, unless you're editing an existing resource that is already using it.

**Supported:**

Persistent Volume volume plugin: `aws-ebs`, `azure-disk`, `azure-file`, `flex-volume-longhorn`,
`gce-pd`,`host-path`,`local`,`nfs`,`vsphere-volume`

StorageClass: `aws-ebs`, `azure-disk`, `azure-file`, `gce-pd`, `longhorn`, `local`, `vsphere-volume`

**Unsupported:** Everything else

izaac:
@vincent99 do we still prefer to wait for QA until this other PR is finished https://github.com/rancher/rancher/pull/22899

Or this is ready for validation after @codyrancher 's changes ?

sangeethah:
Reopening this issue for  #22899 to be merged.

rmweir:
appear to work fine if the page is refreshed, if the page is not refreshed then the user can still use one of these types. If we want to stop the creation then we should do it in validators for PV/SC:
feature off: 
![image](https://user-images.githubusercontent.com/19376037/65285602-d3fc1f80-daf1-11e9-96d5-ce0389003cb0.png)

feature on:
![image](https://user-images.githubusercontent.com/19376037/65285667-0279fa80-daf2-11e9-8ee5-082942210e19.png)

rmweir:
This appear to work when the value is changed via kubectl, but since using the flag or environment variable changes the default, it does not work

rmweir:
Arguments and environment variables appear to work now.

**form:** featurename1=bool,featurename2=bool....

**How to use arguments:**
`sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 rancher/rancher --features=unsupported-storage-drivers=bool`

**How to use environment variable:**
`sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 -e CATTLE_FEATURES='unsupported-storage-drivers=bool' rancher/rancher`

rmweir:
[...]

PR #22899, opened 2019-09-16
Unsupported storage & network drivers feature flag

vincent99:
For UI feature to toggle showing storage drivers that are not commercially supported

### DIFF
--- a/pkg/features/feature.go
+++ b/pkg/features/feature.go
@@ -16,6 +16,8 @@ var (
 	features = make(map[string]*feature)
 
 	// Features, ex.: ClusterRandomName = newFeature("cluster-randomizer", false)
+
+	UnsupportedStorageDrivers = newFeature("unsupported-storage-drivers", false)
 )
 
 type feature struct {


rmweir:
Is the intent of this just to start the feature flag? Are you going to check its value via API? It's not attached to a schema or anything.

vincent99:
They are going to continue to be visible and usable in the sky in this case, but hidden from the UI unless you go toggle the feature because customers are confused about which drivers are commercially supported.

We will build out a UI to toggle the features later.

rmweir:
> They are going to continue to be visible and usable in the sky in this case, but hidden from the UI unless you go toggle the feature because customers are confused about which drivers are commercially supported.
> 
> We will build out a UI to toggle the features later.

okay, lgtm

pkg/features/feature.go
```diff
@@ -16,6 +16,7 @@ var (
 	features = make(map[string]*feature)
 
 	// Features, ex.: ClusterRandomName = newFeature("cluster-randomizer", false)
```
rmweir: just need a newline here

codyrancher:
Did you verify the return value of the feature flag? When I applied this change locally the name and id fields were as expected but the value field was null.

cjellick:
https://github.com/rancher/rancher/issues/22895

Issue #22865, opened 2019-09-14
Redefine system images tag/push/mirror process


alena1108:
- [ ] Introduce a file containing all the images and their original/target tag in kontainer-driver-metadata, types, system-catalog repos. See if dedicated CI flow repo for publishing images makes sense. Automate the process of checking the image existence in the file before accepting chars/metadata changes referencing the image https://github.com/rancher/rancher/issues/21221

- [ ] Add scripts and Drone logic for tagging and pushing images and corresponding Manifests https://github.com/rancher/rancher/issues/22132

- [ ] Change `rancher-images.txt` format and generation logic around it. One file for amd/arm/windows instead of `rancher-images.txt` and `rancher-images-windows.txt`. Rancher min/max versions should be respected

- [ ] Change rancher-load-images.sh and rancher-save-images.sh to support new rancher-images.txt format (will support amd and arm). Introduce new scripts for windows. Consider renaming the scripts to  load to `rancher-pull-images.sh` and `rancher-push-images.sh`

PR #22906, opened 2019-09-17
Update rancher-load-images.sh to support manifest

gitlawr:
Issue: https://github.com/rancher/rancher/issues/22836

## For users who care about Linux only:

The script works exactly like before.

## For users who will use Windows cluster:

- **Step 1:** Users first use `rancher-save-images.ps1` and `rancher-load-images.ps1` to push windows images on Windows 1903 nodes with Docker installed. Recommended Disk Size: 50GB This step pushes the windows tagged images

- **Step 2:** Then users run `./rancher-load-images.sh --registry reg.example.com --image-list ./rancher-images.txt --windows-image-list ./rancher-windows-images.txt` This step pushes the linux tagged images as well as the manifest. 

### Notes
- You have to specify `--windows-image-list ./rancher-windows-images.txt` to indicate you would like to support Windows cluster.
- If you want to use images on 1809 and 1903 instead of 1903 only, they can use `--windows-versions 1809,1903` param


cloudnautique:
Is there an example for the input files?

loganhz:
example for the input files

https://github.com/rancher/rancher/releases/download/v2.3.0-rc4/rancher-images.txt
https://github.com/rancher/rancher/releases/download/v2.3.0-rc4/rancher-windows-images.txt



We would like to make it in one file in 2.3.x as it requires many changes.
Here is the issue for tracking it https://github.com/rancher/rancher/issues/22865

pkg/image/export/main.go
```diff
@@ -224,14 +224,31 @@ const (
...
 images="rancher-images.tar.gz"
 list="rancher-images.txt"
+windows_image_list=""
+windows_versions="1903"
 usage () {
     echo "USAGE: $0 [--images rancher-images.tar.gz] --registry my.registry.com:5000"
     echo "  [-l|--image-list path] text file with list of images. 1 per line."
     echo "  [-i|--images path] tar.gz generated by docker save."
     echo "  [-r|--registry registry:port] target private registry:port."
+    echo "  [--windows-image-list path] text file with list of images used in Windows. Windows image mirroring is skipped when it is empty"
```
cjellick: "when it is empty" is awkward english. "when this is empty" would be better
can be changed in a separate PR

Issue #22943, opened 2019-09-18
Owner is allowed to delete a cluster template in use


soumyalj:
version: master-head

Steps
1. Create user user1 in the rancher setup as admin
2. Create a cluster template/revision T1/R1 . Share the template with user1 as owner
3. Create a cluster using R1.
4. Delete the template as an admin and deletion is not allowed as expected
5. Login as user1 and delete the template T1 in UI
The template gets deleted even though the cluster is using the template
The cluster is still associated with the template

```

appliedSpec": {
"answers": {
"clusterId": null,
"projectId": null,
"type": "/v3/schemas/answer"
},
**"clusterTemplateId": "cattle-global-data:ct-p9sv5",
"clusterTemplateRevisionId": "cattle-global-data:ctr-2vjtk",**
"defaultClusterRoleForProjectMembers": null,
"defaultPodSecurityPolicyTemplateId": null,
"description": "",
"desiredAgentImage": "",
"desiredAuthImage": "",
"displayName": "donewtest",
"dockerRootDir": "/var/lib/docker",
"enableClusterAlerting": false,
"enableClusterMonitoring": false,
"enableNetworkPolicy": false,
"internal": false,
"localClusterAuthEndpoint": {
"enabled": true,
"type": "/v3/schemas/localClusterAuthEndpoint"
}

```
Cluster template list is empty

```
{
"type": "collection",
"links": {
"self": "…/v3/clustertemplates"
},
"createTypes": {
"clusterTemplate": "…/v3/clustertemplates"
},
"actions": { },
"pagination": {
"limit": 1000,
"total": 0
},
"sort": {
"order": "asc",
"reverse": "…/v3/clustertemplates?order=desc",
"links": {
"description": "…/v3/clustertemplates?sort=description",
"name": "…/v3/clustertemplates?sort=name",
"uuid": "…/v3/clustertemplates?sort=uuid"
}
},
"filters": {
"created": null,
"creatorId": null,
"defaultRevisionId": null,
"description": null,
"id": null,
"name": null,
"removed": null,
"uuid": null
},
"resourceType": "clusterTemplate",
"data": [ ]
}




```

soumyalj:
Tested with master-head build from Sep 19th (60d8d4f5ec85)
1. Create user user1 in the rancher setup as admin
2. Create a cluster template/revision T1/R1 . Share the template with user1 as owner
3. Create a cluster using R1.
4. Login as user1(owner of template T1) and delete the template T1 in UI
[...]

PR #22951, opened 2019-09-19
Use lister to list the clusters that are using a clustertemplate

prachidamle:
https://github.com/rancher/rancher/issues/22943

Instead of using the access.List to list the clusters using a given template/revision; we should rely on listers - since access.List would return only the results based on what the user can access/see.

### DIFF
--- a/pkg/api/store/clustertemplate/store.go
+++ b/pkg/api/store/clustertemplate/store.go
@@ -16,6 +16,7 @@ import (
 	mgmtSchema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
 	managementv3 "github.com/rancher/types/client/management/v3"
 	"github.com/rancher/types/config"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/api/store/clustertemplate/store.go
```diff
@@ -178,30 +181,30 @@ func setLabelsAndOwnerRef(apiContext *types.APIContext, data map[string]interfac
...
 	}
 
 	if len(clusters) > 0 {
-		return true, nil
+		for _, cluster := range clusters {
+			switch apiContext.Type {
+			case managementv3.ClusterTemplateType:
+				field = cluster.Spec.ClusterTemplateName
+			case managementv3.ClusterTemplateRevisionType:
+				field = cluster.Spec.ClusterTemplateRevisionName
```
mrajashree: Can we add check for a default case here?

pkg/api/store/clustertemplate/store.go
```diff
@@ -178,30 +181,30 @@ func setLabelsAndOwnerRef(apiContext *types.APIContext, data map[string]interfac
...
-		types.NewConditionFromString(field, types.ModifierEQ, []string{id}...),
-	}
-
-	if err := access.List(apiContext, apiContext.Version, managementv3.ClusterType, &types.QueryOptions{Conditions: conditions}, &clusters); err != nil {
+	clusters, err := p.clusterLister.List("", labels.NewSelector())
+	if err != nil {
 		return false, err
 	}
 
 	if len(clusters) > 0 {
```
mrajashree: nit: we won't need this if now

Issue #22970, opened 2019-09-19
Cluster creation through API call using cluster templates for which users have no access should not be allowed


soumyalj:
Version: Master-head

Steps:
1. Create a cluster template T1/R1 as admin with AD auth enabled
2. Login as testuser2 who does not have access to template T1
3. Create a cluster through API call using the template id of T1
Cluster creation succeeds with out any error.
```

curl --insecure -u "token-cmptg:2b65nq9hc9s22dnl8jllb2s88" -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -d '{"amazonElasticContainerServiceConfig":null, "answers":null, "azureKubernetesServiceConfig":null, "clusterTemplateRevisionId":"cattle-global-data:ctr-nlmlh", "dockerRootDir":"/var/lib/docker", "enableClusterAlerting":false, "enableClusterMonitoring":false, "googleKubernetesEngineConfig":null, "localClusterAuthEndpoint":null, "name":"testntemp1", "rancherKubernetesEngineConfig":null, "windowsPreferedCluster":false}' 'https://abcd/v3/clusters'

```
Cluster creation through API call using cluster templates for which users have no access should not be allowed

soumyalj:
Tested with build from master-head(410f7aa09388) . 

1. Create a cluster template T1/R1 as admin with AD auth enabled
2. Login as testuser2 who does not have access to template T1
3. Create a cluster through API call using the template id of T1

Cluster creation fails as expected

```
curl --insecure -u "token-cmptg:2b65nq9hc9s22dnl8jllb2s88" -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -d '{"amazonElasticContainerServiceConfig":null, "answers":null, "azureKubernetesServiceConfig":null, "clusterTemplateRevisionId":"cattle-global-data:ctr-nlmlh", "dockerRootDir":"/var/lib/docker", "enableClusterAlerting":false, "enableClusterMonitoring":false, "googleKubernetesEngineConfig":null, "localClusterAuthEndpoint":null, "name":"testntemp1", "rancherKubernetesEngineConfig":null, "windowsPreferedCluster":false}' 'https://abcd/v3/clusters'

```

```

{"baseType":"error","code":"NotFound","message":"The clusterTemplateRevision is not found","status":404,"type":"error"}

```

PR #22973, opened 2019-09-19
Access check for clustertemplaterevision while creating cluster

prachidamle:
https://github.com/rancher/rancher/issues/22970

- Adding access check to check if user has access to use the clustertemplaterevision to create cluster
- Changes to response codes in delete/update template when it is in use by a cluster


prachidamle:
Closing, have another PR for the same fix. https://github.com/rancher/rancher/pull/22993

PR #22993, opened 2019-09-20
Access check for clustertemplaterevision while creating cluster

prachidamle:
https://github.com/rancher/rancher/issues/22970

Trying to check the drone build with a new PR and also get more logs from the drone failure


prachidamle:
@alena1108 @mrajashree fresh PR for the same issue #22970. This passes the drone.

Issue #18630, opened 2019-03-06
Search filter under customize schema for openldap doesn't work.


bentastic27:
**What kind of request is this (question/bug/enhancement/feature request):**

bug

**Steps to reproduce (least amount of steps as possible):**

configure openldap on Rancher 2.x, go to customize schema, add a filter to the "Search Filter" and click save.

**Result:**

Clicking save will make it disappear.

**Other details that may be helpful:**

It seems as though the user search filter field isn't included in the API:

```
"userDisabledBitMask": 0,
"userLoginAttribute": "uid",
"userMemberAttribute": "memberOf",
"userNameAttribute": "cn",
"userObjectClass": "inetOrgPerson",
"userSearchAttribute": "uid|sn|givenName",
"userSearchBase": "dc=example,dc=com",
```

Only these messages show up in the logs when trying to add it:

```
2019/03/05 21:12:28 [DEBUG] Impersonating user user-b9p7h, groups [system:authenticated]
2019/03/05 21:12:28 [DEBUG] GET: 2.333876msauthconfigs
2019/03/05 21:12:28 [DEBUG] GET: 1.9112msauthconfigs
2019/03/05 21:12:28 [DEBUG] GET: 2.937752msauthconfigs
2019/03/05 21:12:28 [DEBUG] AuthConfigController calling handler authConfigController openldap
```

Dev tools in chrome also show the api endpoing returning a 200 for the request. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
2.1.6

- Installation option (single install/HA):
Either, I recreated in single, but the customer is on HA.

alena1108:
SearchFilter support was implemented for AD only https://github.com/rancher/rancher/issues/13656. We just have to extend it to include OpenLDAP.

mrajashree:
Same as https://github.com/rancher/rancher/issues/19436

soumyalj:
Tested with v2.3-head. 
Enabled FreeIPA auth .  "Search Filter" field option is presented for User and Groups in "Customize Schema" section.
Users has the option to set these fields when FreeIPA auth is enabled. When the field is set , this filter gets applied to the list of users/groups that get searched when adding users to add members to cluster/projects.
The following tests were done:
1. User Search Filter
(memberOf=cn=davidgroup1,cn=groups,cn=accounts,dc=rmfreeipa,dc=do,dc=rancher,dc=space)
2. Multiple user search filter
[...]

PR #22994, opened 2019-09-20
[2.3] Add custom search filter to ldap search queries

mrajashree:
https://github.com/rancher/rancher/issues/18630

Types PR: https://github.com/rancher/types/pull/982


pkg/auth/providers/ldap/ldap_client.go
```diff
@@ -355,13 +355,15 @@ func (p *ldapProvider) searchUser(name string, config *v3.LdapConfig, lConn *lda
...
+	// The user search filter will be added as another and clause
+	// and is expected to follow ldap syntax and enclosed in parenthesis
+	query += srchAttrs + ")" + config.UserSearchFilter + ")"
 	logrus.Debugf("%s searchUser query: %s", p.providerName, query)
 	return p.searchLdap(query, p.userScope, config, lConn)
 }
 
 func (p *ldapProvider) searchGroup(name string, config *v3.LdapConfig, lConn *ldapv2.Conn) ([]v3.Principal, error) {
-	query := fmt.Sprintf("(&(%v=*%v*)(%v=%v))", config.GroupSearchAttribute, name, ObjectClass, config.GroupObjectClass)
+	query := "(&(" + ObjectClass + "=" + config.GroupObjectClass + ")(" + config.GroupSearchAttribute + "=*" + name + "*)" + config.GroupSearchFilter + ")"
```
leodotcloud: Curious, do we have any kind of safe query verify functions that can be used here? Since we are using the string input from the user directly here, it can potentially be used as an attack vector.

go.mod
```diff
@@ -7,6 +7,7 @@ replace (
 	github.com/crewjam/saml => github.com/rancher/saml v0.0.0-20180713225824-ce1532152fde
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
+	github.com/rancher/types v0.0.0-20190917172018-55dc8a6c4865 => /Users/rajashree/go-work/src/github.com/rancher/types
```
leodotcloud: I am assuming you will fix this in the final draft?

pkg/auth/providers/ldap/ldap_client.go
mrajashree: I am not aware of a safe query verify function. This can be only configured by an admin while setting up auth. And that query is done against the admin's ldap server. It only adds an additional filter to the output being returned. So I don't see how it can be used as an attack against rancher either.

go.mod
mrajashree: yes..

mrajashree: the types pr isn't reviewed or merged yet, so yeah once it is merged I will be updating go.mod with the commit

pkg/auth/providers/ldap/ldap_client.go
leodotcloud: You are right, this is an admin config. Probably, it my comments are not relevant in this context. Anytime I see any text input from user being used, I get paranoid.

pkg/auth/providers/ldap/ldap_client.go
```diff
@@ -355,13 +355,15 @@ func (p *ldapProvider) searchUser(name string, config *v3.LdapConfig, lConn *lda
 	for _, attr := range srchAttributes {
 		srchAttrs += fmt.Sprintf("(%v=%v*)", attr, name)
 	}
-	query += srchAttrs + "))"
+	// The user search filter will be added as another and clause
+	// and is expected to follow ldap syntax and enclosed in parenthesis
+	query += srchAttrs + ")" + config.UserSearchFilter + ")"
```
prachidamle: di we validate the searchFilter input?

mrajashree: We don't validate it for AD either. If we want to validate, then during auth setup we can run ldapsearch with this filter, if it errors then we don't setup auth?

prachidamle: so seems like we let configure their own search filter, we should document it that admin needs to enter a valid filter.

prachidamle:
LGTM

leodotcloud:
LGTM. (Fix the types commit, once that PR is merged)

PR #23007, opened 2019-09-23
Change ticker write way

Masterlu1998:
Problem:
Some controller's Lifecycle or Handler doesn't work, when something out of time.
Solution:
I find Rancher.ticker used the wrong way. ticker.Stop will not close the channel and this function will be permanently blocked. I change the ticker write way.


Masterlu1998:
Sorry, it was my first time to send a PR, I'm a little nervous and made a lot of stupid things. Now everything is fine! : )

Issue #21695, opened 2019-07-22
K8s metadata -  deprecating a k8s version does not work


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- run `Rancher-testing: v2.3.2`
- modify the metadata to set the highest minor version for v1.11 is `v1.11.8-rancher1-1`
- create a cluster with k8s `v1.11.8-rancher1-1`
- upgrade the metadata to add `v1.11.9-rancher1-2`, and deprecate `v1.11.8-rancher1-1`
- refresh the kontainerdrivers in Rancher
- see that the k8s `v1.11.8-rancher1-1` is gone from the list
- edit the existing cluster to add a control plane node 

**Result:**
- the node is added and the cluster is updated successfully

**Expected Results:**

- We should NOT be able to add new etcd/ control plane node, 
- we should be able to upgrade the cluster's k8s to a higher version.

jiaqiluo:
The same bug for deprecating a major k8s version

**Steps:**

- run `Rancher-testing: v2.3.2`
- set the metadata that v2.3.2 supports k8s v1.11.x
- add a cluster with k8s v1.11.9
- modify the metadata to deprecate k8s v1.11

```
"v1.11": {
			MaxRancherVersion: "2.3.1",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.11.9-rancher1-1": {
			MaxRancherVersion: "2.3.2",
			MaxRKEVersion:     "0.2.2",
		},
		"v1.11.9-rancher1-2": {
			MaxRancherVersion: "2.3.2",
			MaxRKEVersion:     "0.2.2",
		},
```

- refresh the kontainerdrivers in Rancher
- edit the existing cluster to add 2 etcd nodes

**Result:**
- the node is added and the cluster is updated successfully

**Expected Results:**

- We should NOT be able to add new etcd/ control plane node, 
- we should be able to upgrade the cluster's k8s to a higher version.

jiaqiluo:
Reopen the bug, failed to validate in the setup upgraded from v2.2.8 to v2.3.0-rc4

Steps:
- run Rancher:v2.2.8 single install
- add a cluster with k8s version `v1.14.6-rancher1-1`
- upgrade to Rancher:v2.3.0-rc4
- edit the rke-metadata-config to point to my metadata git repo
- deprecate the k8s version in the metadata
```
"v1.14.6-rancher1-1": {
[...]

PR #23014, opened 2019-09-23
move isDeprecatedCheck from controller to store

kinarashah:
should return deprecated err only on cluster edits by user, not
on every sync which would cause clusters to be in a
failed updating state

https://github.com/rancher/rancher/issues/21695


prachidamle:
can we add an integration test now that the check is in store?

PR #23035, opened 2019-09-23
Escape ldap dn used in refresh group principals

mrajashree:


alena1108:
@mrajashree could you explain the fix?

mrajashree:
ExternalID is the distinguishedName of a user, which is being used as the baseDN
in ldap search query in RefetchGroupPrincipals. This doesn't need to be split on `\`.
That split is only for when we're using some field that has "logindomain\username"
format. Splitting incorrectly on `\` removed a part of the distinguishedName leading
to ldap errors

mrajashree:
@alena1108 I had included that in commit message but forgot updating PR description with it, updated now

prachidamle:
@mrajashree any idea why this change was present to start with - I wonder if there was any special case we were handling here

mrajashree:
@prachidamle It originally got added in the [commit](https://github.com/rancher/rancher/commit/a883cfbcf1d77e2d83656d2c13c943fb02c6a952) for refreshing group principals. I think that commit is accidentally splitting on principalID to follow what was done in loginUser. However, in loginUser we're splitting on `sAMAccountName`, and in this case the string is externalID of a principal which will always be the `distinguishedName`. It only fails for users with a comma or special characters in their DN, since it gets escaped by adding a `\`

prachidamle:
LGTM

Issue #23040, opened 2019-09-24
[Windows] add environment variable AGENT_IMAGE in the adding-node command


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- run v2.3.0-rc8 in airgap HA mode
- add a cluster with both Windows and Linux node
- run the command from Rancher UI in the Windows instance

**Result:**
- the Windows node tries to use the image  `rancher/rancher-agent:v2.3.0-rc8` although it has pulled the proper one from the private registry
```
v2.3.0-rc8: Pulling from rancher/rancher-agent
Digest: sha256:220c2e5580bbbc9b238a8577b57655bd3c93601ca3241d272c3eb7de8de883a1
Status: Downloaded newer image for ec2-54-215-219-6.us-west-1.compute.amazonaws.com:443/rancher/rancher-agent:v2.3.0-rc8
INFO: Detecting running permission ...
INFO: Detecting host CPU reservation ...
INFO: Detecting host RAM reservation ...
INFO: Detecting host DISK reservation ...
INFO: Detecting host system locale ...
INFO: Detecting host Docker name pipe existing ...
INFO: Detecting host Docker release version ...
INFO: Detecting host network interface ...
INFO: Configuring host Docker startup mode to automatical ...
INFO: Enabling host msiscsi service to support iscsi storage ...
Unable to find image 'rancher/rancher-agent:v2.3.0-rc8' locally
```

The workaround is to add the `-e AGENT_IMAGE=registry/rancher/rancher-agent:v2.3.0-rc8` to the command from the UI

 
So we should update the command in UI to include the `AGENT_IMAGE` part.

jiaqiluo:
The bug fix is validated in master-head d009d239b

Steps:
- run Rancher single install
- set the system default registry to a private registry
- add a new custom cluster with support for Windows nodes and private registry
- check the command for adding Windows nodes

Result:
- the command contains the `-e AGENT_IMAGE=...` part as expected

PR #23074, opened 2019-09-25
Fix generating incomplete Windows command in air-gapped

thxCode:
Patch the `AGENT_IMAGE` env back to the command when the setting side
agent image isn't equal to the resolved side.

**Issue:**
https://github.com/rancher/rancher/issues/23040

### DIFF
--- a/pkg/api/customization/clusterregistrationtokens/formatter.go
+++ b/pkg/api/customization/clusterregistrationtokens/formatter.go
@@ -15,7 +15,7 @@ const (
 	insecureCommandFormat = "curl --insecure -sfL %s | kubectl apply -f -"
 	nodeCommandFormat     = "sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run %s --server %s --token %s%s"
 
-	windowsNodeCommandFormat = `PowerShell -NoLogo -NonInteractive -Command "& {docker run -v c:/:c:/host %s bootstrap --server %s --token %s%s | iex}"`
+	windowsNodeCommandFormat = `PowerShell -NoLogo -NonInteractive -Command "& {docker run -v c:/:c:/host %s%s bootstrap --server %s --token %s%s | iex}"`
 )
 
 func Formatter(request *types.APIContext, resource *types.RawResource) {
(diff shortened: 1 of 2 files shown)


pkg/api/customization/clusterregistrationtokens/formatter.go
```diff
@@ -29,16 +29,27 @@ func Formatter(request *types.APIContext, resource *types.RawResource) {
...
+			agentImage,
+			rootURL,
 			token,
 			ca)
-		resource.Values["token"] = token
-		resource.Values["manifestUrl"] = url
+		// for windows
+		var agentImageDockerEnv string
+		if agentImage != agentImageInSetting {
+			agentImageDockerEnv = fmt.Sprintf("-e AGENT_IMAGE=%s ", agentImage)
```
gitlawr: Do we enter this condition when the system default registry is set?  Can we do the following which would be more intuitive:
```
systemDefaultRegistry:=settings.SystemDefaultRegistry.Get()
if systemDefaultRegistry != "" {
    ...
}
```

thxCode: Updated

pkg/api/customization/clusterregistrationtokens/formatter.go
```diff
@@ -15,7 +15,7 @@ const (
 	insecureCommandFormat = "curl --insecure -sfL %s | kubectl apply -f -"
 	nodeCommandFormat     = "sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run %s --server %s --token %s%s"
 
-	windowsNodeCommandFormat = `PowerShell -NoLogo -NonInteractive -Command "& {docker run -v c:/:c:/host %s bootstrap --server %s --token %s%s | iex}"`
+	windowsNodeCommandFormat = `PowerShell -NoLogo -NonInteractive -Command "& {docker run -v c:/:c:/host %s%s bootstrap --server %s --token %s%s | iex}"`
```
cjellick: Seems weird for there to be no space between `%s%s`

cjellick:
@loganhz @gitlawr @thxCode So the agent command looks like 
```
PowerShell -NoLogo -NonInteractive -Command "& {docker run -v c:/:c:/host -e AGENT_IMAGE=fake-registry.rancher.io:443/rancher/rancher-agent:test fake-registry.rancher.io:443/rancher/rancher-agent:test bootstrap --server %s --token %s%s | iex}"`
```
I don't understand why we need to specify the AGENT_IMAGE environment variable if the correct image is specified as the image.

Like, it seems like the agent should be intelligent enough to handle this scenario without needing AGENT_IMAGE set. Why doesn't Linux need to set the AGENT_IMAGE when using a private regirsty?

thxCode:
@cjellick ,

For Linux command, it is still one `docker run` command: 
```
sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run fake-registry.rancher.io:443/rancher/rancher-agent:test --server %s --token %s%s
```

but for Windows command, we could separate it to 2 part: 
```
PowerShell -NoLogo -NonInteractive -Command "& {

docker run -v c:/:c:/host %s%s bootstrap --server %s --token %s%s // **the docker run command**

| iex // **the powershell command**

}"
```
1. the `docker run command` would create some new directories on the host, put some util files into the host and generate an entry point PowerShell script
1. the `powershell command` would execute the above entry point script to detect some prerequisites, if everything is ok, it would `docker run` the agent container like that single Linux command: https://github.com/rancher/rancher/blob/58e0a2faf07eef2e104c9e4e2a941c1f86c1f064/package/windows/bootstrap.ps1#L269-L273

The starting point of the agent container is that `powershell command`, so the real question is how to query its Docker image name/ID inside a container? I found that we could use this to query the image ID on Linux: https://github.com/rancher/rke-tools/blob/8be3757c440e013a7049bfc64143aeb504684559/share-root.sh#L3-L4, but there is no Cgroup on Windows.

Maybe I could try to modify the `bootstrap.ps1` to query all container and compare each command with itself, I will ping you later.

pkg/api/customization/clusterregistrationtokens/formatter.go
cjellick: @thxCode what about this? is the lack of a space here a bug or intentional?

thxCode: The below logic is: 
```
                // for windows
		var agentImageDockerEnv string
		if settings.SystemDefaultRegistry.Get() != "" {
			// patch the AGENT_IMAGE env
			agentImageDockerEnv = fmt.Sprintf("-e AGENT_IMAGE=%s ", agentImage)
		}
		resource.Values["windowsNodeCommand"] = fmt.Sprintf(windowsNodeCommandFormat,
			agentImageDockerEnv,
			agentImage,
			rootURL,
			token,
			ca)
```
I want to make the format argument `agentImageDockerEnv` explicitly via `%s%s`.

thxCode:
@cjellick , If we modify the `bootstrap.ps1` to query all container and compare each command with itself, it looks like guess the image ID, as we could not ensure there is not another container with the same command. Although the chance is very low. So I think this works, but it is not very reliable. I prefer to use env to indicate image name not guess the image ID.

cjellick:
OK @thxCode I agree

PR #23076, opened 2019-09-25
Fix failed to add Windows worker with intermediary certificates

thxCode:
**Problem:**
certoc.exe only accept the first cert of the certificate file.

**Solution:**
Use a loop to iterate per certificate from the file.

P.S.: The logic of this part is lost during refactoring. Patching back as the same as https://github.com/rancher/rancher/commit/45af0e6c3ab6cf70b51a7f63d4b5381c72a64efc#diff-b7a68d3a8052546f5539c0f2661ae672 but using `certoc.exe` to import certificates instead of `Import-Certificate` in the nanoserver.

**Issue:**
Reopen: https://github.com/rancher/rancher/issues/20436

### DIFF
--- a/package/windows/execute.ps1
+++ b/package/windows/execute.ps1
@@ -286,9 +286,22 @@ if ($CATTLE_CA_CHECKSUM)
     $temp.MoveTo("$sslCertDir\serverca")
 
     # import the self-signed certificate
-    certoc.exe -addstore root "$sslCertDir\serverca" | Out-Null
-    if (-not $?) {
-        Log-Error "Failed to import rancher server certificates to Root"
+    $caBytes = $null
+    Get-Content "$sslCertDir\serverca" | % {
+        if ($_ -match '-+BEGIN CERTIFICATE-+') {
+            $caBytes = @()
+        } elseif ($_ -match '-+END CERTIFICATE-+') {
+            $caTemp = New-TemporaryFile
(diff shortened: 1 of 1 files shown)


package/windows/execute.ps1
```diff
@@ -286,9 +286,22 @@ if ($CATTLE_CA_CHECKSUM)
...
+            $caBytes = @()
+        } elseif ($_ -match '-+END CERTIFICATE-+') {
+            $caTemp = New-TemporaryFile
+            Set-Content -Value $caBytes -Path $caTemp.FullName -Encoding Byte
+            certoc.exe -addstore root $caTemp.FullName | Out-Null
+            if (-not $?) {
+                $caTemp.Delete()
+                Log-Error "Failed to import rancher server certificates to Root"
+            }
+            $caTemp.Delete()
```
gitlawr: Looks like it won't exit on Log-Error: https://github.com/rancher/rancher/blob/883c41e029a70a482f612f569983ae86d84ab0bf/package/windows/utils.psm1#L19
Will it do delete twice when an error occurs?

thxCode: Using `Log-Fatal` to make the logic exit when failed, so the user could get noticed. `$caTemp.Delete()` just drops the temp file, those imported certs would store in `Cert:\LocalMachine\Root\`. If getting an import error on the last one of the file, I think we don't need to delete those certs already imported.

Issue #23043, opened 2019-09-24
Update k8s 1.13, 1.14 and 1.15 versions on Rancher 2.2.x


alena1108:
Have to:

- [ ] Refresh k8s versions for 1.13, 1.14 and 1.15 lines on release/v2.2, in `rancher/types`
- [ ] Backport - forwardport? - versions to rancher master, `rancher/kontainer-driver-metadata`, for backwards compatibility

prachidamle:
- To take care of future upgrades of v2.2.9 to v2.3, we need to update kontainer-driver-metadata with system-images that are supported by v2.2.9 k8s versions.

- In v2.2.9, for versions v1.14.7-rancher1-1 and v1.15.4-rancher1-1, the "WindowsPodInfraContainer" image that v2.3 has does not exist.

- Hence we will be maintaining v1.14.7-rancher1-1 and v1.15.4-rancher1-1 without "WindowsPodInfraContainer" in kontainer-driver-metadata so that upon upgrade the existing clusters on v2.2.9 do not go to updating state.

- In kontainer-driver-metadata v1.14.7-rancher1-2 and v1.15.4-rancher1-2 will be having the windows-infra-pod image that will apply for v2.3 onwards.

jiaqiluo:
The bug fix is validated in Rancher:v2.2-head 9aea30e5

As a standard user, I add three EC2 clusters which have the following k8s versions
```
- 1.15.4 - default 
- 1.14.7
- 1.13.11
```
and each cluster has 
```
- 3 etcd
- 1 control plane
- 3 worker nodes. 
```

All automation tests are passed on those three clusters.

Issue #22945, opened 2019-09-18
Upgrade kubectl to 1.14.7


alena1108:
Upgrade our embedded kubectl to the latest version in k8s 1.13.x release line

https://github.com/rancher/rancher/issues/22938

prachidamle:
We will be upgrading kubectl version 1.14.7 and also support 1.15.x k8s version in Rancher UI
cc @alena1108

sangeethah:
Tested with latest build from `2.2-head` - `1375ef26`

```
> kubectl version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.7", GitCommit:"8fca2ec50a6133511b771a11559e24191b1aa2b4", GitTreeState:"clean", BuildDate:"2019-09-18T14:47:22Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4", GitCommit:"67d2fcf276fcd9cf743ad4be9a9ef5828adc082f", GitTreeState:"clean", BuildDate:"2019-09-18T14:41:55Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

PR #23086, opened 2019-09-25
v2.2.9: Update kubectl to 1.14.7 and support 1.15.x k8s version and update vendor

prachidamle:
https://github.com/rancher/rancher/issues/22945
https://github.com/rancher/rancher/issues/23043

Depends on PRs:
types: https://github.com/rancher/types/pull/993
rke: https://github.com/rancher/rke/pull/1674


alena1108:
@prachidamle we need to deprecate 1.12 in types, and make 1.15 a default

prachidamle:
> @prachidamle we need to deprecate 1.12 in types, and make 1.15 a default

Ok, I was not aware we need to update types, will look at that

Issue #21070, opened 2019-06-23
App upgrade failure: panic at go/src/k8s.io/helm/pkg/tiller/release_update.go:166


zhongguanqun:
What kind of request is this (question/bug/enhancement/feature request):
**panic: runtime error: invalid memory address or nil pointer dereference**
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x12e8cc8]

goroutine 15 [running]:
k8s.io/helm/pkg/tiller.(*ReleaseServer).performUpdateForce(0xc420819860, 0xc42052a8c0, 0x0, 0x0, 0x17b7b40)
	/go/src/k8s.io/helm/pkg/tiller/release_update.go:166 +0x188
k8s.io/helm/pkg/tiller.(*ReleaseServer).UpdateRelease(0xc420819860, 0x17d8960, 0xc420932210, 0xc42052a8c0, 0xc420819860, 0x16b98a6, 0xd)
	/go/src/k8s.io/helm/pkg/tiller/release_update.go:43 +0x245
k8s.io/helm/pkg/proto/hapi/services._ReleaseService_UpdateRelease_Handler.func1(0x17d8960, 0xc420932210, 0x16066c0, 0xc42052a8c0, 0x32, 0xc42052aa50, 0x1, 0x1681d11)
	/go/src/k8s.io/helm/pkg/proto/hapi/services/tiller.pb.go:1244 +0x86
k8s.io/helm/vendor/github.com/grpc-ecosystem/go-grpc-prometheus.(*ServerMetrics).UnaryServerInterceptor.func1(0x17d8960, 0xc420932210, 0x16066c0, 0xc42052a8c0, 0xc42094e6e0, 0xc42094e700, 0x48, 0xc42094e5c0, 0xc420a37a40, 0x4108c8)
	/go/src/k8s.io/helm/vendor/github.com/grpc-ecosystem/go-grpc-prometheus/server_metrics.go:86 +0xdb
k8s.io/helm/pkg/tiller.newUnaryInterceptor.func1(0x17d8960, 0xc420932210, 0x16066c0, 0xc42052a8c0, 0xc42094e6e0, 0xc42094e700, 0xc420655ab0, 0x4108c8, 0x50, 0x15d88a0)
	/go/src/k8s.io/helm/pkg/tiller/server.go:60 +0xab
[...]

PR #23099, opened 2019-09-26
Turn it up to 10 TILLER_HISTORY_MAX

dramich:
Problem:
Tiller can get stuck when an upgrade fails and we don't have any
versions to get back to

Solution:
Up the history count so tiller can find one that deployed successfully
to rollback to

https://github.com/rancher/rancher/issues/21070


alena1108:
LGTM

PR #23100, opened 2019-09-26
provision  a NFS server when running tests for PV and PVC

jiaqiluo:
An NFS server will be deployed in AWS for testing Persistent Volumes and Persistent Volume Claims.
A new configurable variable `RANCHER_DELETE_NFS` is introduced. The NFS server will be deleted automatically if `RANCHER_DELETE_NFS` is set to `True`, or be kept if `RANCHER_DELETE_NFS` is set to `False`


sangeethah:
LGTM

PR #23102, opened 2019-09-26
[2.3] Add check for app creation

mrajashree:
Add check for app creation

### DIFF
--- a/tests/integration/suite/test_project_catalog.py
+++ b/tests/integration/suite/test_project_catalog.py
@@ -1,4 +1,5 @@
-from .conftest import wait_until, wait_until_available
+from .conftest import wait_until, wait_until_available, wait_for_condition,\
+    user_project_client
 from rancher import ApiError
 from .common import random_str
 import time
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_project_catalog.py
```diff
@@ -211,6 +212,59 @@ def test_user_addition_after_creating_project_catalog(admin_mc,
...
+
+    name = random_str()
+    new_project = client.create_project(name=random_str(), clusterId='local')
+    remove_resource(new_project)
+    wait_for_condition('InitialRolesPopulated', 'True', client, new_project)
+    new_project = client.reload(new_project)
+
+    project_name = str.lstrip(new_project.id, "local:")
+    catalog_name = project_name + ":" + name
+    url = "https://github.com/mrajashree/charts.git"
```
prachidamle: Do we have a QA or test charts repo - we should have one

mrajashree: no we don't, but yeah we should

prachidamle: or we can use rancher charts repo itself

mrajashree: few tests were using rancher charts repo, but there was once a change in rancher charts that removed a version and all these tests started failing.

dramich: We do have one `DEFAULT_CATALOG = "https://github.com/rancher/integration-test-charts"`

mrajashree: okay, I wasn't aware of this and have been using my fork for charts in tests. updated it

prachidamle:
LGTM

alena1108:
@mrajashree what issue does it address? It's better to have a github ticket with the clear steps to reproduce for qa

PR #23109, opened 2019-09-26
k8s 1.16

kinarashah:
`unknown revision v0.0.0 errors` on kubernetes components while using `k8s.io/kubernetes v1.16.0` similar to `https://github.com/kubernetes/kubernetes/issues/79384`. Followed step mentioned in `https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505627280` to specifically mention all components with 1.16 tag in replace


ibuildthecloud:
LGTM

Issue #22556, opened 2019-08-30
podsecuritypolicytemplateprojectbindings does not validate template exists


dramich:
When creating a podsecuritypolicytemplateprojectbindings a `podSecurityPolicyTemplateId` is referenced. If that template does not exist the binding is still created successfully but the server will error saying the template is not found but no error is returned to the user. The binding sync attempts to create the psp in the downstream cluster but is unable too because the template does not exist. There should be validation to ensure the template exists before creating the binding resource. 

This happens in `pkg/controllers/user/rbac/podsecuritypolicy/binding.go`

To repro: 
Using a valid cluster/project ID send the following payload to rancher with a template ID that does not exist in rancher:

```
{
    "name": "testing",
    "namespaceId": "default",
     "podSecurityPolicyTemplateId": "doesntexist",
    "targetProjectId": "<valid projectID>"
}
```

No error is returned but the server will error with:
```
2019/08/30 11:13:11 [ERROR] PodSecurityPolicyTemplateProjectBindingController default/dddd [PodSecurityPolicyTemplateProjectBindingsLifecycleHandler] failed with : error creating policy: error getting pspt: podSecurityPolicyTemplate.management.cattle.io "doesntexist" not found
```

jiaqiluo:
The bug is reproduced in v2.3.1 

follow the steps from the issue to send a request, and see the error in Rancher's log 
```
2019/11/01 00:01:49 [ERROR] PodSecurityPolicyTemplateProjectBindingController c-8j9mg/psptpb-8zfcz [PodSecurityPolicyTemplateProjectBindingsLifecycleHandler] failed with : error creating policy: error getting pspt: podSecurityPolicyTemplate.management.cattle.io "abcdefg" not found
2019/11/01 00:01:49 [ERROR] PodSecurityPolicyTemplateProjectBindingController c-8j9mg/psptpb-8zfcz [PodSecurityPolicyTemplateProjectBindingsLifecycleHandler] failed with : error creating policy: error getting pspt: podSecurityPolicyTemplate.management.cattle.io "abcdefg" not found
2019/11/01 00:01:50 [ERROR] PodSecurityPolicyTemplateProjectBindingController c-8j9mg/psptpb-8zfcz [PodSecurityPolicyTemplateProjectBindingsLifecycleHandler] failed with : error creating policy: error getting pspt: podSecurityPolicyTemplate.management.cattle.io "abcdefg" not found
[...]

PR #23110, opened 2019-09-26
Add validation to psptBinding creation

dramich:
Problem:
When creating a psptBinding a template id is required but not validated
so an invalid value does not return to the user and causes the
controller to keep trying to create resources

Solution:
Validate the template ID belongs to a real template before creating the
binding

https://github.com/rancher/rancher/issues/22556


luthermonson:
lgtm

PR #23114, opened 2019-09-27
[2.3.2] enhance logging configure

aiwantaozi:
Problem:
1. Project match source use all file under /var/log/containers which include log files not belong to this project
2. When configuring read_from_head, if target file is large, it takes a long time and starting other plugins isn't executed until reading file is finished.
3. Default file chunk size is 256M, kafka plugin will iterate over each 256M file and deliver each 1024K, it takes a long time if file is too large

Solution:
1. Only tail file belong to this project
2. Remove read_from_head
3. reduce each chunk file size for kafka, set default kafka aggregation threshold higher

Issue:
https://github.com/rancher/rancher/issues/23303

### DIFF
--- a/pkg/api/customization/logging/validator.go
+++ b/pkg/api/customization/logging/validator.go
@@ -75,8 +75,10 @@ func validate(level, containerLogSourceTag string, loggingTargets v3.LoggingTarg
 		}
 	}
 
-	if err = generator.ValidateCustomTags(wrap); err != nil {
-		return err
+	if len(outputTags) != 0 {
+		if err = generator.ValidateCustomTags(wrap); err != nil {
+			return err
+		}
 	}
 
 	return generator.ValidateCustomTarget(wrap)
--- a/pkg/controllers/user/logging/config/config.go
+++ b/pkg/controllers/user/logging/config/config.go
@@ -102,3 +102,7 @@ func RancherLoggingSSLSecretName() string {
 func GetNamespacePattern(namespace string) string {
 	return fmt.Sprintf("^%s$", namespace)
 }
+
+func GetNamespacePathPattern(namespace string) string {
+	return fmt.Sprintf("/var/log/containers/*_%s_*.log", namespace)
+}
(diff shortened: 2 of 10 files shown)


orangedeng:
LGTM

aiwantaozi:
Update kafka configure

orangedeng:
LGTM

aiwantaozi:
collect kafka changes into one commit

pkg/api/customization/logging/validator.go
```diff
@@ -75,8 +75,10 @@ func validate(level, containerLogSourceTag string, loggingTargets v3.LoggingTarg
 		}
 	}
 
-	if err = generator.ValidateCustomTags(wrap); err != nil {
-		return err
+	if len(outputTags) != 0 {
+		if err = generator.ValidateCustomTags(wrap); err != nil {
```
alena1108: I see that the same pattern is being repeated here and on line 51. Would it make sense to include len(outputTags) != 0 verification in ValidateCustomTags method itself, and just return from the method immediately if ==0?

alena1108:
@aiwantaozi will there be a system-charts pr? or just rancher side changes

pkg/api/customization/logging/validator.go
aiwantaozi: I agree this `!=0` is being repeated, I check the method ValidateCustomTags, we pass interface{} as method's parameters, because we may pass ProjectLoggingTemplateWrap or ClusterLoggingTemplateWrap and use them the generate the configure. So if we may need to parse it back to map if we move the `!=` into ValidateCustomTags method, I prefer to keep it outside even it is being repeated

aiwantaozi:
@alena1108 Just rancher side changes

pkg/api/customization/logging/validator.go
alena1108: ok, it's fine, not a big deal

alena1108:
LGTM

PR #23118, opened 2019-09-27
Add mirror images

gitlawr:


scripts/istio-images-mirror
```diff
@@ -14,3 +14,8 @@ pstauffer/curl rancher/pstauffer-curl v1.0.3
 coredns/coredns rancher/coredns-coredns 1.1.2
 quay.io/coreos/prometheus-operator rancher/coreos-prometheus-operator v0.32.0
 quay.io/coreos/prometheus-config-reloader rancher/coreos-prometheus-config-reloader v0.32.0
+plugins/docker rancher/plugins-docker 18.09
```
alena1108: This image has the same source as jenkins-plugins-docker, which is already in Rancher dockerhub https://github.com/rancher/types/blob/master/image/mirror.go#L17. Is the intent to re-name the image?

gitlawr: Yes, it is mistakenly mirrored to `jenkins-plugins-docker` but it is not a jenkins plugin(drone plugin image indeed), so I think we should correct the name

scripts/istio-images-mirror
```diff
@@ -14,3 +14,8 @@ pstauffer/curl rancher/pstauffer-curl v1.0.3
...
+minio/minio rancher/minio-minio RELEASE.2019-09-25T18-25-51Z
+grafana/grafana rancher/grafana-grafana 6.3.6
+library/nginx rancher/nginx 1.17.4-alpine
+jenkins/jnlp-slave rancher/jenkins-jnlp-slave 3.35-4
```
deniseschannon: This was the only image that CICD actually pushed. I imagine the other repos in dockerhub were missing perms? @cloudnautique @drpebcak

deniseschannon: I added dockerhub permissions for the automated published on the other repos.

PR #23181, opened 2019-10-02
Add Metrics to Rancher

dramich:
This is based on the work done by @ukinau in [this pr](https://github.com/rancher/rancher/pull/21351)

The metrics endpoint will always be live and authed, it requires either admin permissions or the newly created role `view-rancher-metrics`. If the env var `CATTLE_PROMETHEUS_METRICS` is not set metrics will pull the generic Go metrics, similar to getting into the container and hitting the Go metrics endpoint. If the env var is set then all additional monitoring related to rancher will be enabled. 

Depends on:
https://github.com/rancher/norman/pull/310
https://github.com/rancher/remotedialer/pull/10


ukinau:
@dramich  Thanks for taking over that. Here is the grafana dashboard sample.
We have 2 different dashboards, "WebSocket Session Related Dashboard" and "Controller Related Dashboard".   

WebSocket Session Related Dashboard (https://gist.github.com/ukinau/11bb65380ebbbe82c557c4b2e20645a2)
スクリーンショット 2019-10-12 9 54 00
スクリーンショット 2019-10-12 9 54 09


Controller Related Dashboard (https://gist.github.com/ukinau/faf1363a383c1fae9805c793c04b8550)
スクリーンショット 2019-10-12 9 55 40
スクリーンショット 2019-10-12 9 55 58
スクリーンショット 2019-10-12 9 56 14

pkg/api/controllers/usercontrollers/usercontroller.go
```diff
@@ -144,7 +145,13 @@ func (u *userControllersController) amOwner(peers tpeermanager.Peers, cluster *v
 	scaled := int(ck) * len(peers.IDs) / math.MaxUint32
 	logrus.Debugf("%s(%v): (%v * %v) / %v = %v[%v] = %v, self = %v\n", cluster.Name, cluster.UID, ck,
 		uint32(len(peers.IDs)), math.MaxUint32, peers.IDs, scaled, peers.IDs[scaled], peers.SelfID)
-	return peers.IDs[scaled] == peers.SelfID
+	isOwner := peers.IDs[scaled] == peers.SelfID
+	if isOwner {
+		metrics.SetClusterOwner(peers.SelfID, cluster.Name)
```
luthermonson: fine leaving this here but seems weird to me that we'd set the metrics cluster owner in a bool check function. maybe better logic in peersSync that calls amOwner?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/rancher/rancher/pkg/settings"
+	"github.com/rancher/types/config"
+	authV1 "k8s.io/api/authorization/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var (
+	prometheusMetrics = false
+
+	gcInterval = time.Duration(60 * time.Second)
```
luthermonson: any benefit to let someone set this by config or envvar?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func NewMetricsHandler(scaledContext *config.ScaledContext, handler http.Handler) http.Handler {
+	return &metricsHandler{
+		k8sClient: scaledContext.K8sClient,
+		next:      handler,
+	}
+}
+
+func (h metricsHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
```
luthermonson: slept on it, maybe a pointer more efficient since this thing will be called a bunch, feeling this could fall within the when in doubt clause

pkg/metrics/garbagecollector.go
```diff
@@ -0,0 +1,183 @@
...
+	buildObservedLabelMaps(targetMetricsByIPForPeer, "peer", observedLabelsMap)
+	buildObservedLabelMaps([]interface{}{clusterOwner}, "cluster", observedLabelsMap)
+
+	removedCount := removeMetricsForDeletedResource(observedLabelsMap, observedResourceNames)
+
+	logrus.Debugf("[metrics-garbage-collector] Finished - removed %d items", removedCount)
+}
+
+func buildObservedLabelMaps(collectors []interface{}, targetLabel string, observedLabels map[string]map[interface{}][]map[string]string) int {
+	// {
```
alena1108: do we need lines 95-102

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+var (
+	prometheusMetrics = false
+
+	gcInterval = time.Duration(60 * time.Second)
+
+	clusterOwner = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: "cluster_manager",
+			Name:      "cluster_owner",
+			Help:      "Set 1 when cluster is owned, by this metrics you can find out which rancher server run user controllers for specific cluster",
```
alena1108: have to fix the grammar here :)

pkg/metrics/garbagecollector.go
dramich: It's the structure of the maps this thing builds. I found it helpful...

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	// Cluster Owner
+	prometheus.MustRegister(clusterOwner)
+
+	gc := metricGarbageCollector{
+		clusterLister:  context.Management.Clusters("").Controller().Lister(),
+		nodeLister:     context.Management.Nodes("").Controller().Lister(),
+		endpointLister: context.Core.Endpoints(settings.Namespace.Get()).Controller().Lister(),
+	}
+
+	go func() {
```
alena1108: we should pass the context to any go routine that we launch in rancher, so everything gets canceled when rancher server exists. This can be leveraged https://github.com/rancher/rancher/blob/master/pkg/controllers/management/catalog/controller.go#L30

pkg/api/controllers/usercontrollers/usercontroller.go
dramich: Good call. Moved.

pkg/metrics/metrics.go
dramich: Not that I can think of? I could be persuaded to do so but not sure it's actually useful.

pkg/metrics/garbagecollector.go
```diff
@@ -0,0 +1,183 @@
...
+		logrus.Errorf("[metrics-garbage-collector] failed to list clusters: %s", err)
+		return
+	}
+	for _, cluster := range clusters {
+		if _, ok := observedResourceNames[cluster.Name]; !ok {
+			observedResourceNames[cluster.Name] = true
+		}
+	}
+	// Get Nodes
+	nodes, err := gc.nodeLister.List("", labels.Everything())
```
alena1108: think we can call all listers before going over (nodes, clusters)? This way if any of them returns an error, we save cpu on iteration and memory on map allocation :)

pkg/metrics/garbagecollector.go
alena1108: ah, may be we can add a line like: `Example of the map`. So the purpose of the commented stuff is more clear. What do you think

pkg/metrics/garbagecollector.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+	return removedCount
+}
+
+func appendIfLabelIsNotInList(targetLabel map[string]string, labelList []map[string]string) []map[string]string {
+	found := false
+	for _, label := range labelList {
+		if reflect.DeepEqual(targetLabel, label) {
+			found = true
```
alena1108: shall we break here, if found?

pkg/metrics/garbagecollector.go
dramich: It only stops if cluster breaks, for the others it keeps going to at least clean clusters. If clusters aren't there there isn't much that can be done.

pkg/metrics/garbagecollector.go
dramich: Yes :)

pkg/metrics/garbagecollector.go
dramich: Agreed. Updated

pkg/metrics/metrics.go
dramich: Done

alena1108:
LGTM

Issue #12186, opened 2018-03-21
Admins should be able to create and control Node Templates


jgreat:
In a lot of enterprise environments allocation of resources is controlled by an operations group.  They are responsible for setting up and maintaining all the underlying infrastructure in the cloud, managing costs, etc...

Users of envionments (developers, qa, sec...) might not know and maybe should not need to know the details required to set up a node template.

As a Global Admin I want to set up node templates with  "Automation User" credentials and allow my Cluster Admins to use the templates when starting/stopping/sizing clusters.

I don't want users or cluster admins using personal cloud provider credentials in templates to create or modify clusters, because all that will break if a user changes jobs or leaves the company and I deprovision their cloud provider account.    

**Rancher versions:**
rancher/server: v2.0.0-alpha21

Overv:
We would also really like to have this feature to set up templates as admin for our users to apply.

billabongrob:
With 2-3 admins of the environment, it would be nice to share the node templates between team members.  This would help avoid confusion.

gillarda:
As other users suggested, this feature would be very appreciated in environments where multiple administrators manage clusters for various teams or customers.

DJAyth:
Very much interested in this.

mattskipton:
We would really like this feature as well

okunc:
+1

yuliujin:
We would like to have this feature as well.

JonathanParrilla:
+1

cloudnautique:
This issue will be used to track allowing an Rancher admin to administer all node templates. Today when a user leaves Rancher, no one can administer the cluster created from that template. The sharing component will be addressed in #22467 for the 2.4 timeframe.

cjellick:
@rmweir im taking wes off this. please make sure there is a ui issue for it

rmweir:
Suggested scenarios to test:

1. Node template Migration:
Node templates created from prior to this feature should be cloned into cattle-global-nt namespace, with an ID that includes the creator user id and old node template name. Old node template should then be deleted.

2. When a user creates a node template they should be able to view it. (Should contain annotation indicating that role bindings have been created for owner)

3. After creating a node template, it should have a field call "principalId" viewable from API.

4. Non-admin users cannot see node templates created by other non-admin users or admins.

5. Admin should be able to edit and delete node templates created by others.

6. Admin should be able to change cloud credential on node template and successfully provision cluster with it.

[...]

PR #23182, opened 2019-10-02
Admin node templates

rmweir:
**Problem:**
Admins should be able to CRUD node templates

**Solution:**
Make node templates rbac enforced so they can be exposed to admins, and migrate node templates to a global namespace.

**Depends on:**
https://github.com/rancher/norman/pull/314

**Issue:**
https://github.com/rancher/rancher/issues/12186


app/global_namespace.go
```diff
@@ -3,34 +3,44 @@ package app
...
 	} else if err != nil {
-		return fmt.Errorf("Error creating %v namespace: %v", namespace.GlobalNamespace, err)
+		return nil, fmt.Errorf("error creating %v namespace: %v", namespace, err)
 	}
 
-	logrus.Debugf("calling sync for driver metadata")
-	management.Management.Settings("").Controller().Enqueue("", settings.RkeMetadataConfig.Name)
-
-	return nil
+	return nil, nil
```
luthermonson: dont want to return the namesapce?

pkg/api/customization/cred/store.go
```diff
@@ -5,14 +5,16 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/labels"
+
```
luthermonson: eh?

rmweir: Separating the k8s imports from github imports, I've noticed some people prefer this so I am conforming.

luthermonson: well if the cool kids are doing it

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,281 @@
...
+
+	s := schema.GroupVersionResource{
+		Group:    "management.cattle.io",
+		Version:  "v3",
+		Resource: "nodetemplates",
+	}
+	ntDynamicClient := dynamicClient.Resource(s)
+
+	migratedNTName := fmt.Sprintf("nt-%s-%s", nodeTemplate.Namespace, nodeTemplate.Name)
+	if nodeTemplate.Namespace != namespace.NodeTemplateGlobalNamespace {
```
luthermonson: maybe since this only gonna technically happen once it could move to another function so the sync handler reads a little cleaner?

rmweir: good point, moved

app/global_namespace.go
```diff
@@ -3,34 +3,44 @@ package app
...
+		return err
+	}
+
+	logrus.Debugf("calling sync for driver metadata")
+	management.Management.Settings("").Controller().Enqueue("", settings.RkeMetadataConfig.Name)
+
+	return nil
+}
 
+func createNamespace(namespace string, management *config.ManagementContext) (*corev1.Namespace, error) {
```
dramich: Don't return the namespace, it isn't going to work anyways since the create call is eating the ns.

app/global_namespace.go
```diff
@@ -3,34 +3,44 @@ package app
...
-	_, err := lister.Get("", namespace.GlobalNamespace)
+	ns, err := lister.Get("", namespace)
 	if k8serrors.IsNotFound(err) {
-		ns := &corev1.Namespace{}
-		ns.Name = namespace.GlobalNamespace
+		ns = &corev1.Namespace{}
+		ns.Name = namespace
 		if _, err := management.Core.Namespaces("").Create(ns); err != nil {
-			return fmt.Errorf("Error creating %v namespace: %v", namespace.GlobalNamespace, err)
+			return nil, fmt.Errorf("error creating %v namespace: %v", namespace, err)
```
dramich: Check the error to see if it's an alreadyExists error.

pkg/api/customization/namespacedresource/namespaced_store.go
```diff
@@ -0,0 +1,39 @@
+package namespacedresource
```
dramich: The file should be renamed to `store.go`. This is normal convention and go files usually don't have an underscore unless it's denoting the architecture the file is for i.e. `store_windows.go`

pkg/api/customization/namespacedresource/namespaced_store.go
```diff
@@ -0,0 +1,39 @@
...
+		Store:              store,
+		NamespaceInterface: nsClient,
+		Namespace:          namespace,
+	}
+}
+
+func (s *namespacedStore) Create(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}) (map[string]interface{}, error) {
+	ns, ok := values.GetValue(data, client.PreferenceFieldNamespaceId)
+	if ok && !strings.EqualFold(convert.ToString(ns), s.Namespace) {
+		return nil, fmt.Errorf("Error creating namespaced resource, cannot assign to %v since already assigned to %v namespace", namespace.GlobalNamespace, ns)
```
dramich: nit: Error > error

pkg/api/customization/nodetemplate/formatter.go
```diff
@@ -28,30 +24,21 @@ func (ntf *Formatter) Formatter(request *types.APIContext, resource *types.RawRe
...
-		resource.ID = ""
-		resource.Values = map[string]interface{}{}
-		resource.Links = map[string]string{}
-		resource.Actions = map[string]string{}
-		resource.Type = ""
-
-		return false
+	user, err := ntf.UserLister.Get("", creatorID)
+	if err != nil {
+		logrus.Warnf("Failed to to get user associated with creatorId [%s]. Error: %v", creatorID, err)
```
dramich: I'm not sure we should warn here? This would most likely indicate the user is deleted which seems fine. Perhaps an error check to validate before the warning and if it's just a 404 return out.

pkg/api/store/nodetemplate/store.go
```diff
@@ -25,6 +26,24 @@ type Store struct {
...
+			Store:                 store,
+			NodePoolLister:        npLister,
+			CloudCredentialLister: secretLister,
+		},
+		Transformer: filterLegacyTemplates,
+	}
+}
+
+func filterLegacyTemplates(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
+	if data["namespaceId"] != namespace.NodeTemplateGlobalNamespace {
```
dramich: It could be useful to enqueue the nodetemplate to get it migrated if it's seen in here

pkg/controllers/management/rbac/rbac_common.go
```diff
@@ -1,11 +1,10 @@
-package globalnamespacerbac
+package rbac
```
dramich: Drop the `_common` from the filename

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
+package nodetemplate
```
dramich: Drop `_handler` from the filename

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	nsClient   v1.NamespaceInterface
+	mgmtCtx    *config.ManagementContext
+}
+
+func Register(ctx context.Context, mgmt *config.ManagementContext) {
+	nt := nodeTemplateController{
+		roleLister: mgmt.Management.GlobalRoles("").Controller().Lister(),
+		roleClient: mgmt.Management.GlobalRoles(""),
+		rbLister:   mgmt.Management.GlobalRoleBindings("").Controller().Lister(),
+		rbClient:   mgmt.Management.GlobalRoleBindings(""),
```
dramich: It doesn't look like the 4 things above this are being used?

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+		ntClient:   mgmt.Management.NodeTemplates(""),
+		ntLister:   mgmt.Management.NodeTemplates("").Controller().Lister(),
+		npLister:   mgmt.Management.NodePools("").Controller().Lister(),
+		npClient:   mgmt.Management.NodePools(""),
+		nsLister:   mgmt.Core.Namespaces("").Controller().Lister(),
+		nsClient:   mgmt.Core.Namespaces(""),
+		mgmtCtx:    mgmt,
```
dramich: Let's discuss this, looks like it's being saved for use with `CreateRoleAndRoleBinding`. The other places it's being used should be added to the struct instead, such as nodes.

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	metaAccessor, err := meta.Accessor(nodeTemplate)
+	if err != nil {
+		return nodeTemplate, err
+	}
+
+	creatorID, ok := metaAccessor.GetAnnotations()[rbac.CreatorIDAnn]
+	if !ok {
+		return nodeTemplate, fmt.Errorf("clusterTemplate %v has no creatorId annotation", metaAccessor.GetName())
```
dramich: `clusterTemplate`?

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	// Parts of the node template in k8s are dynamic, such as an azureConfig. If the dynamic client is not use then the
+	// dynamic fields, like azureConfig, are stripped when marshalling the data into the NodeTemplate struct. If changes
+	// are made, and that struct is committed to k8s, then the dynamic field is effectively removed.
+	restConfig := nt.mgmtCtx.RESTConfig
```
dramich: Can this be created and saved on the struct above? Then it won't need to be created on the fly and passed into the handlers

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+			if _, err := nt.mgmtCtx.Management.Nodes("").Update(nodeCopy); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// getDynamicAnnotations returns the annotations of the given unstructured object returns by dynamic, k8s client
+func getDynamicAnnotations(dynamicNodeTemplate *unstructured.Unstructured, nodeTemplateName string) (map[string]interface{}, error) {
```
dramich: This shouldn't be needed, use `dynamicNodeTemplate.GetAnnotations()`

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	annotations, ok := metadata["annotations"].(map[string]interface{})
+	if !ok {
+		annotations = make(map[string]interface{})
+	}
+
+	return annotations, nil
+}
+
+// writeDynamimcAnnotations adds the requested annotation to the given unstructured object returned by dynamic, k8s client
+func writeDynamimcAnnotations(dynamicNodeTemplate *unstructured.Unstructured, nodeTemplateName string, key string, value string) (*unstructured.Unstructured, error) {
```
dramich: This shouldn't be needed, use `dynamicNodeTemplate.SetAnnotations()`

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	dynamicNodeTemplate.Object["metadata"] = metadata
+
+	return dynamicNodeTemplate, nil
+}
+
+// createGlobalNodeTemplateClone returns the global clone of the given legacy node template. If one does not exist
+// it will be created
+func (nt *nodeTemplateController) createGlobalNodeTemplateClone(legacyName, cloneName string, dynamicNodeTemplate *unstructured.Unstructured,
+	client dynamic.NamespaceableResourceInterface) error {
+	if _, err := nt.nsLister.Get("", namespace.NodeTemplateGlobalNamespace); err != nil {
```
dramich: Drop the ns create, if this doesn't already exist something bad happened in startup

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+			return err
+		}
+	}
+
+	if _, err := nt.ntLister.Get(namespace.NodeTemplateGlobalNamespace, cloneName); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+
+		globalNodeTemplate := dynamicNodeTemplate.DeepCopy()
```
dramich: This isn't needed, you own this object from the dynamic client

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	mgmt.Management.NodeTemplates("").Controller().AddHandler(ctx, "nt-grb-handler", nt.sync)
+}
+
+func (nt *nodeTemplateController) sync(key string, nodeTemplate *v3.NodeTemplate) (runtime.Object, error) {
+	if nodeTemplate == nil || nodeTemplate.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
```
dramich: Check for the ownerBindingsAnno annotation, if exists return

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+	if nodeTemplate.Namespace != namespace.NodeTemplateGlobalNamespace {
+		// node template must be migrated to global namespace
+		if err := nt.migrateNodeTemplate(ntDynamicClient, nodeTemplate); err != nil {
+			return nil, err
+		}
+
+		// node template has been deleted- handler should return nil object
+		nodeTemplate = nil
+	} else {
```
dramich: Drop the else

dramich: The migration should give back a new version of the nodetemplate in the global ns, use that one to create the bindings and update the annotation once completed

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+				return nil, err
+			}
+
+			// owner bindings annotation is meant to prevent bindings from being created again if they have been removed from creator
+			dynamicNodeTemplate, err := writeDynamimcAnnotations(dynamicNodeTemplate, migratedNTName, ownerBindingsAnno, "true")
+			if err != nil {
+				return nil, err
+			}
+
+			if _, err := ntDynamicClient.Namespace(nodeTemplate.Namespace).Update(dynamicNodeTemplate, metav1.UpdateOptions{}); err != nil {
```
dramich: It should be OK to use the normal clients here for the updates.

pkg/controllers/management/nodetemplate/nodetemplate_handler.go
```diff
@@ -0,0 +1,294 @@
...
+// migrateNodeTemplate creates duplicate of node template in the global node template namespace, creates new role bindings
+// for duplicate, then deletes old node template
+func (nt *nodeTemplateController) migrateNodeTemplate(ntDynamicClient dynamic.NamespaceableResourceInterface, nodeTemplate *v3.NodeTemplate) error {
+	logrus.Infof("migrating node template [%s]", nodeTemplate.Spec.DisplayName)
+
+	migratedNTName := fmt.Sprintf("nt-%s-%s", nodeTemplate.Namespace, nodeTemplate.Name)
+	fullLegacyNTName := fmt.Sprintf("%s:%s", nodeTemplate.Namespace, nodeTemplate.Name)
+	fullMigratedNTName := fmt.Sprintf("%s:%s", namespace.NodeTemplateGlobalNamespace, migratedNTName)
+
+	dynamicNodeTemplate, err := ntDynamicClient.Namespace(nodeTemplate.Namespace).Get(nodeTemplate.Name, metav1.GetOptions{})
```
dramich: Move this get into `createGlobalNodeTemplateClone`

tests/integration/suite/test_node_templates.py
```diff
@@ -0,0 +1,116 @@
...
+                "field.cattle.io/creatorId": admin_mc.user.id
+            }
+        },
+        "kind": "NodeTemplate",
+        "apiVersion": "management.cattle.io/v3",
+    }
+
+    node_template = k8s_dynamic_client.create_namespaced_custom_object(
+        "management.cattle.io", "v3", ns.name, 'nodetemplates', body)
+    remove_resource(node_template)
```
dramich: This doesn't work with k8s client objects, get the obj with our client and pass it into remove_resource OR you can update the fixture to handle k8s object types.

tests/integration/suite/test_node_templates.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def node_template_migrate(id):
```
dramich: nit: function name should describe what it's doing

tests/integration/suite/test_node_templates.py
```diff
@@ -0,0 +1,116 @@
...
+    node_template_name = "nt-" + random_str()
+    body = {
+        "metadata": {
+            "name": node_template_name,
+            "annotations": {
+                "field.cattle.io/creatorId": admin_mc.user.id
+            }
+        },
+        "kind": "NodeTemplate",
+        "apiVersion": "management.cattle.io/v3",
```
dramich: You need a dynamic config in here to validate it gets migrated to the new NS

tests/integration/suite/test_node_templates.py
```diff
@@ -0,0 +1,116 @@
...
+            assert e.error.status == 403
+            return False
+
+    wait_for(lambda: node_template_migrate("cattle-global-nt:nt-" + ns.id +
+             "-" + node_template["metadata"]["name"]), fail_handler=lambda:
+             "failed waiting for node template to migrate")
+
+    wait_for(lambda: admin_client.by_id_node_template(id=ns.name + ":" +
+             node_template_name) is None, fail_handler="failed waiting for"
+             "old node template to delete")
```
dramich: Go get the new node template and compare to the old one to ensure the the dynamic fields are there and the new annotation showing bindings are created exist.

tests/integration/suite/test_node_templates.py
```diff
@@ -0,0 +1,116 @@
...
+    """asserts that node template is automatically created in
+    'cattle-global-nt' namespace"""
+    admin_client = admin_mc.client
+
+    node_template = admin_client.create_node_template(name="nt-" +
+                                                      random_str(),
+                                                      azureConfig={})
+    remove_resource(node_template)
+
+    assert node_template.id.startswith("cattle-global-nt")
```
dramich: Check the namespace?

tests/integration/suite/test_node_templates.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
+            nt = client.by_id_node_template(id=node_template.id)
+            return nt
+        except ApiError as e:
+            assert e.error.status == 403
+            return None
+
+    wait_for(can_view_template, fail_handler=lambda:
+             "creator was unable to view node template")
+
```
dramich: Use a 2nd normal user to attempt to get the same nodetemplate

app/global_namespace.go
rmweir: it already checks if the namespace exists above here

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func Register(ctx context.Context, mgmt *config.ManagementContext) {
+	// Parts of the node template in k8s are dynamic, such as an azureConfig. If the dynamic client is not use then the
+	// dynamic fields, like azureConfig, are stripped when marshalling the data into the NodeTemplate struct. If changes
+	// are made, and that struct is committed to k8s, then the dynamic field is effectively removed.
+	restConfig := mgmt.RESTConfig
+	dynamicClient, err := dynamic.NewForConfig(&restConfig)
+	if err != nil {
+		logrus.Errorf("[NodeTemplate Register] error creating dynamic client: %v", err)
```
dramich: Panic here

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+	"k8s.io/client-go/dynamic"
+)
+
+const (
+	ownerBindingsAnno = "ownerBindingsCreated"
+)
+
+type nodeTemplateController struct {
+	ntClient        v3.NodeTemplateInterface
+	ntLister        v3.NodeTemplateLister
```
dramich: This doesn't look like it's used

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+
+	mgmt.Management.NodeTemplates("").Controller().AddHandler(ctx, "nt-grb-handler", nt.sync)
+}
+
+func (nt *nodeTemplateController) sync(key string, nodeTemplate *v3.NodeTemplate) (runtime.Object, error) {
+	if nodeTemplate == nil || nodeTemplate.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	annotations := nodeTemplate.Annotations
```
dramich: Move the annotation check up to here `if nodeTemplate.Annotations !=nil && annotations[ownerBindingsAnno] == "true"`

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+
+	var templateName = nodeTemplate.Name
+	var migratedTemplate bool
```
dramich: Move these two down after all the checks are done, no point in allocating them if they won't be used

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// if owner bindings annotation is present, the node template is in the proper namespace and has had
+	// its creator rolebindings created
+	if ownerBindingsCreated := annotations[ownerBindingsAnno]; ownerBindingsCreated == "true" {
+		return nil, nil
```
dramich: Return the object

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+	annotations = dynamicNodeTemplate.GetAnnotations()
+	// owner bindings annotation is meant to prevent bindings from being created again if they have been removed from creator
+	annotations[ownerBindingsAnno] = "true"
+	dynamicNodeTemplate.SetAnnotations(annotations)
+
+	if _, err = nt.ntDynamicClient.Namespace(namespace.NodeTemplateGlobalNamespace).Update(dynamicNodeTemplate, metav1.UpdateOptions{}); err != nil {
+		return nil, err
+	}
+
+	// if a migrations has been performed, then the original node template object has not been updated
```
dramich: If we migrate doesn't that mean the original obj that came into the sync is deleted?

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	creatorID, ok := annotations[rbac.CreatorIDAnn]
+	if !ok {
+		return nodeTemplate, fmt.Errorf("nodeTemplate [%v] has no creatorId annotation", nodeTemplate.Name)
+	}
+
+	if nodeTemplate.Namespace != namespace.NodeTemplateGlobalNamespace {
+		// node template must be migrated to global namespace
+		if err := nt.migrateNodeTemplate(nt.ntDynamicClient, nodeTemplate); err != nil {
```
dramich: This should be getting back the new template to pass it into CreateRoleBindings.

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -0,0 +1,236 @@
...
+			return nil, err
+		}
+		templateName = fmt.Sprintf("nt-%s-%s", nodeTemplate.Namespace, nodeTemplate.Name)
+		migratedTemplate = true
+	}
+
+	// Create Role and RBs if they do not exist
+	if err := rbac.CreateRoleAndRoleBinding(rbac.NodeTemplateResource, nodeTemplate.Name, namespace.NodeTemplateGlobalNamespace,
+		rbac.RancherManagementAPIVersion, creatorID, []string{rbac.RancherManagementAPIVersion},
+		nodeTemplate.UID,
```
dramich: This is the old template

tests/integration/suite/conftest.py
```diff
@@ -375,6 +375,26 @@ def clean():
...
+    """Remove a k8s resource after a test finishes even if the test fails."""
+    def _cleanup(resource):
+        client = CustomObjectsApi(admin_mc.k8s_client)
+
+        def clean():
+            try:
+                metadata = resource["metadata"]
+                client.delete_namespaced_custom_object(
+                    "management.cattle.io", "v3", metadata["namespace"],
+                    'nodetemplates', metadata["name"], {})
```
dramich: hard coding `nodetemplates` makes this not work with anything else.

tests/integration/suite/conftest.py
```diff
@@ -375,6 +375,26 @@ def clean():
     return _cleanup
 
 
+@pytest.fixture
+def remove_k8s_resource(admin_mc, request):
```
dramich: The naming and comment of this should be updated to reflect that this only works with CRDs, not normal k8s objects.

tests/integration/suite/conftest.py
```diff
@@ -375,6 +375,26 @@ def clean():
...
+    """Remove a k8s resource after a test finishes even if the test fails."""
+    def _cleanup(resource, plural_kind):
```
dramich: Instead of passing in plural_kind it can be build off the object with `resource['kind'].lower()+'s'`

rmweir: Talked to @vincent99 about this before. Problem is this is not perfect, some things may require intervention. Quick example, some things ending with c require an s some don't. I went ahead added logic to just attach an s if the name ends with s or z and made it overridable.

dramich:
lgtm

Issue #23189, opened 2019-10-02
EKS - invalid  instance type in region me-south-1 in 2-2-head


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- run rancher v2.2-head, 89e6d3fc single install
- add an EKS cluster， use all default values: 
```
region: me-south-1
k8s version: 1.13
instance type: M4.large
```

**Result:**
- In Rancher UI the cluster is stuck in provisioning state with the error
```
Error creating stack with worker nodes template: stack failed to create: The specified instance type m4.large is not valid (Service: AmazonAutoScaling; Status Code: 400; Error Code: ValidationError; Request ID: 51b4f5f9-e532-11e9-aa48-e7d23e545d24)
```
<img width="1404" alt="Screen Shot 2019-10-02 at 1 31 48 PM" src="https://user-images.githubusercontent.com/6218999/66082121-47545700-e51e-11e9-94f0-4c901d0f3230.png">

- In Rancher, we cannot go to projects in the cluster 

- In AWS console the cluster is active
- the stack fails in AWS CloudFormation, it shows the same error as the on in Rancher UI
<img width="1832" alt="Screen Shot 2019-10-02 at 3 26 08 PM" src="https://user-images.githubusercontent.com/6218999/66086525-28a78d80-e529-11e9-98a3-96b0e49a1413.png">

More info:
- in master-head, I can create the EKS cluster in the same region, but the default k8s version and instance size are different:
```
- me-south-1
- 1.14
- m5.large 
```
 
Update 1:
- in the setup that found the bug, I can create another EKS cluster with  the following configurations
```
- me-south-1
- 1.13
- m5.large 
```
the only difference with the failed one is the instance type

rmweir:
I am checking to see if this is also an issue in master, as I would imagine it to be. Also, it appears another error shows for Bahrain in master: https://github.com/rancher/rancher/issues/22630

rmweir:
Apparently I don't have permissions but I can't imagine why it wouldn't be an issue in master.

jiaqiluo:
The bug fix is validated in v2.2-head 26d28803 with UI tag v2.2.97

Now the default instance type is m5.large , and the EKS is provisioned successfully and functional with the default values

PR #23205, opened 2019-10-03
ui 2.2.97

cjellick:


rmweir:
Issue: https://github.com/rancher/rancher/issues/23189

Issue #23206, opened 2019-10-03
Support k8s 1.16.1 as experimental


deniseschannon:

sangeethah:
Tested with latest build from master.

Able to bring up RKE cluster using version K8s version 1.16.1. Automation run on this cluster did not show any new issues.

sowmyav27:
**Verified on `rancher:v2.3.0-rc10`**

- k8s v1.16.1-rancher1-1 is available

**Verified on `master-head` latest ** 

- Able to bring up windows cluster - 1903 with k8s v1.16.1-rancher1-1
- Verified - Workload deployment, ingress, node port, curl on workloads work

Issue #22164, opened 2019-08-13
Update cli to a non-rc version in pipeline-tools


loganhz:
Update cli to a non-rc version in pipeline-tools when there is a non-rc cli ready

https://github.com/rancher/pipeline-tools/commit/cc560f65c391fe5ec0645b75b37933eae4084d0f

bmdepesa:
rancher/rancher:v2.3.0-rc11

* Reproduced in master-head (before RC11), saw Rancher CLI version as `2.3.0-rc7`
* Create a cluster
* Enable a pipeline
* Execute pipeline
* While pipeline is executing, exec shell into the executor/agent pod (with pipeline tools image)
* From shell, exec `rancher`
* Outputs:

```
bash-4.4# rancher
Rancher CLI, managing containers one UTF-8 character at a time

Usage: rancher [OPTIONS] COMMAND [arg...]

Version: v2.3.0

Options:
  --debug        Debug logging
  --help, -h     show help
  --version, -v  print the version
```

PR #23211, opened 2019-10-03
vendor update

kinarashah:


alena1108:
https://github.com/rancher/rancher/issues/23206
https://github.com/rancher/rancher/issues/22164

Issue #22308, opened 2019-08-20
Create a dev branch for kontainer-driver-metadata once Rancher v2.3 ships


alena1108:
After v2.3 ships, we should remember to create a dev (staging) branch for kontainer-driver-metadata repo. Also as a part of it, we would need to update rancher to use dev branch by default for the non official release images.

deniseschannon:
Should it be a dev branch or per release branch?

deniseschannon:
Spoke with @alena1108, no need to branch rke-tools. We branched kontainer-driver-metadata.

We have also branched rancher/rancher, so need to work on updating the master branch in rancher/rancher to use dev branch.

PR #23227, opened 2019-10-04
Update RKEMetadataConfig setting to use dev branch for kontainer-driver-metadata

prachidamle:
Changing setting to dev branch since we branched off for 2.3 and will use master branch only for releases.


deniseschannon:
https://github.com/rancher/rancher/issues/22308

Issue #23160, opened 2019-10-01
Make jailer script timeout configurable (or at least increased)


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Enhancement

**Steps to reproduce (least amount of steps as possible):**

When a customer is running Anti-virus software on their Rancher nodes, the jailer script can take take more than it's hard-coded 10second timeout to complete, and the Rancher container therefore fails to start (per https://github.com/rancher/rancher/issues/22379). The user would therefore like the jailer timeout to be configurable, or at least increased above 10 seconds.

**Environment information**
- Rancher v2.2.8

fizzyduck:
I'm also seeing this on Rancher v2.3.0-rc9 on hardware which is not running AV software but is otherwise heavily loaded.

dramich:
After some internal discussion this is going to be turned into a setting with a default of 60 seconds. We feel the 60 seconds will cover the majority of issues users are running into while also allowing operators to configure it with the understanding there could be other system impacts by doing so.

sangeethah:
Tested with 2.3-head - c45ec3648

Currently the default setting for jailer-timeout is 60 seconds.

To simulate the jailer scripts failing to complete before timeout , did the following steps:
In the jailer script (/usr/bin/jailer.sh) , induce a sleep of 15 sec.

Set jailer-timeout to 10 seconds ( < sleep set in jailer script).

Create a cluster with DO node driver.
Node provisioning fails with following error message that states the reason:

Following are the error message seen when jailer script fails to complete before timeout:

2019/10/11 23:48:26 [ERROR] NodeController c-ff8nn/m-dmf2n [node-controller] failed with : node create jail error: error running the jail command: timed out waiting for the script to complete: signal: killed

Set jailer-timeout to 20 seconds ( > sleep set in jailer script)

Try to Provision a new node. Provisioning of new node succeeds as expected.

fizzyduck:
Out of interest, is the test automated?

bmdepesa:
@fizzyduck this test is not automated

PR #23249, opened 2019-10-05
Increase jailer wait timeout to 60 seconds

fizzyduck:
On host machines which are running more slowly than expected due to factors outside our control (such as anti-virus, high CPU usage, older hardware) Rancher can be prevented from starting due to the current 10 second timeout.

I'd like to propose this change to increase the timeout to 60 seconds.

### DIFF
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -46,7 +46,7 @@ func CreateJail(name string) error {
 
 	logrus.Debugf("Creating jail for %v", name)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
 	defer cancel()
 
 	cmd := exec.CommandContext(ctx, "/usr/bin/jailer.sh", name)


fizzyduck:
@dramich Updated for master.

leodotcloud:
There might exist a platform where 60 seconds also might seem less. Does it make sense to add a CLI argument and make this configurable? (And leave the default at 10s)

fizzyduck:
I agree that there could be platforms where 60 seconds is not enough. I would have made it configurable but I'm not familiar enough with the codebase to do that.

What was the original rationale behind the timeout in the first place? Does jailer sometimes hang? Is there a way we could check that jailer is still doing its job without an arbitrary timeout?

fengzixu:
I think we don't need to modify this timeout in hard-code way for the official repository now. We can change it on our own codebase and to build the new image. 

Just make this timeout configurable otherwise it will confuse other developers.
@leodotcloud

leodotcloud:
> What was the original rationale behind the timeout in the first place? Does jailer sometimes hang? Is there a way we could check that jailer is still doing its job without an arbitrary timeout?

@dramich would you know the reason?

fengzixu:
> > What was the original rationale behind the timeout in the first place? Does jailer sometimes hang? Is there a way we could check that jailer is still doing its job without an arbitrary timeout?
> 
> @dramich would you know the reason?

I also want to know why we need to add the jailer.go and jailer.sh into the repository. I don't find any related issues to explain it.

paulsmithuk:
@fengzixu, while I appreciate the need for a root cause, this issue has been killing us since we upped to 2.2.8. Can we accept @fizzyduck 's patch for now, and fix it up later?

fengzixu:
> @fengzixu, while I appreciate the need for a root cause, this issue has been killing us since we upped to 2.2.8. Can we accept @fizzyduck 's patch for now, and fix it up later?

Actually you can modify the source code of Rancher to change the timeout of performing jailer script. But I think that we couldn't repair this problem with hard-code timeout. Because we cannot give an accurate timeout for all clients.

For example if we modify this timeout to 60s, some parts of the client perhaps will avoid this problem. But I am not sure if the remainder will face the same problem.

fizzyduck:
My personal feeling is that we should try and avoid a configuration parameter to control timeouts - Rancher will be more usable if it is able to manage itself in these sorts of scenarios so that users aren't having to scratch their heads, read logs and try and work out which timeout they should increase.

This PR really is a temporary fix to get things working for those that have hit the issue - I am certainly struggling with it daily as are others. Perhaps the timeout should be considerably longer? Perhaps 5 minutes? I don't know what implications this would have on the wider system which leads me to my next point.

Longer term, I'd like to understand what the timeout is trying to protect against, and whether there is another way of solving the problem without resorting to timeouts. To look in to that further, we need input from Rancher members who understand the architecture better than me.

dramich:
After some internal discussion this is going to be turned into a setting with a default of 60 seconds. We feel the 60 seconds will cover the majority of issues users are running into while also allowing operators to configure it with the understanding there could be other system impacts by doing so. Once my pull request is up I or @fizzyduck can close this one out and move the discussion to the issue https://github.com/rancher/rancher/issues/23160

dramich:
Closing in favor of https://github.com/rancher/rancher/pull/23309

fizzyduck:
Thank you.

PR #23263, opened 2019-10-07
Add Vsphere Lookup API

luthermonson:
Adding vSphere endpoints for the vSphere node template UI to call and help fill form fields with real answers from a vSphere server. This PR adds a new HttpHandler for /meta/vsphere and full documentation is in confluence.


pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,401 @@
...
+	case "data-centers":
+		data, err = listDataCenters(ctx, finder)
+	case "resource-pools":
+		data, err = listResourcePools(ctx, finder)
+	}
+
+	return data, err
+}
+
+func tagsLister(ctx context.Context, fieldName string, cc *v1.Secret, cat string) ([]string, error) {
```
cbron: There's 2 `...Lister` funcs and then a bunch of `list...` funcs, is there specific reason for that ? Seems like they both have similar functionality but only the `list..` ones are included in other methods.

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,401 @@
...
+func soapLister(ctx context.Context, fieldName string, cc *v1.Secret, dc string) ([]string, error) {
+	finder, err := getSoapFinder(ctx, cc, dc)
+	if err != nil {
+		return nil, err
+	}
+
+	var data []string
+	switch fieldName {
+	case "clusters":
+		data, err = listClusters(ctx, finder) //tbd
```
cbron: is this still tbd ? If so can you add some more info on what it is we are waiting for

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,190 @@
...
+	"port":     "vmwarevspherecredentialConfig-vcenterPort",
+}
+
+type Handler struct {
+	schemas       *types.Schemas
+	secretsLister v1.SecretLister
+}
+
+func NewVsphereHandler(scaledContext *config.ScaledContext) *Handler {
+	return &Handler{
```
rmweir: Is it possible for the return type to be http.Handler, so the local struct doesn't need to be exported?

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,401 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	mgr := tags.NewManager(rest.NewClient(soap.Client))
+	ui := url.UserPassword(string(cc.Data[dataFields["username"]]), string(cc.Data[dataFields["password"]]))
+	err = mgr.Login(ctx, ui)
+	if err != nil {
+		return nil, err
+	}
```
cbron: Feels like this ^ block could be abstracted so it's not repeated.

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,190 @@
...
+	return &types.APIContext{
+		Method:  req.Method,
+		Request: req,
+		Schemas: v.schemas,
+		Query:   map[string][]string{},
+	}
+}
+
+func validFieldName(s string) bool {
+	if _, ok := fieldNames[s]; !ok {
```
rmweir: suggestion: could just do
`_, ok := fieldNames[s]`
`return ok`

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,401 @@
...
+
+func listClusters(ctx context.Context, finder *find.Finder) ([]string, error) {
+	clusters, err := finder.ClusterComputeResourceList(ctx, "*")
+	if err != nil {
+		return nil, err
+	}
+
+	var data []string
+	fmt.Println("asdf")
+	fmt.Printf("%#v\n", clusters)
```
rmweir: did you meant to remove these?

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func getContentLibraryManager(ctx context.Context, cc *corev1.Secret) (*library.Manager, error) {
+	soap, err := getSoapClient(ctx, cc)
+	if err != nil {
+		return nil, err
+	}
+
+	mgr := library.NewManager(rest.NewClient(soap.Client))
+	ui := url.UserPassword(string(cc.Data[dataFields["username"]]), string(cc.Data[dataFields["password"]]))
```
rmweir: why is this called ui?

pkg/api/customization/vsphere/listers.go
luthermonson: oh goodness

pkg/api/customization/vsphere/listers.go
luthermonson: nope, it's not tbd any more. it's complete. ill remove

pkg/api/customization/vsphere/handler.go
luthermonson: yup, totally. switching

pkg/api/customization/vsphere/handler.go
luthermonson: yes, done

pkg/api/customization/vsphere/listers.go
luthermonson: talked it out a bit with @cbron will switch <blah>list to process<blah> and leave list<blah> as they are.

pkg/api/customization/vsphere/listers.go
luthermonson: it's of type Userinfo

pkg/api/customization/vsphere/listers.go
luthermonson: they're different types, probably not worth the headache with how govmomi sets up a tags.Manager and a tags.Library. im ok to look harder but it's only called twice.

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,187 @@
...
+	dc := req.FormValue("dataCenter")
+
+	if fieldName == "" || !validFieldName(fieldName) {
+		util.ReturnHTTPError(res, req, httperror.NotFound.Status, fmt.Sprintf("%s: invalid field name", httperror.NotFound.Code))
+		return
+	}
+
+	cc, err := v.getCloudCredential(req)
+	if err != nil {
+		util.ReturnHTTPError(res, req, httperror.NotFound.Status, fmt.Sprintf("%s: %s", httperror.InvalidBodyContent.Code, err.Error()))
```
dramich: The error code should match the message you are returning. This is true for all the cases below

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,187 @@
...
+	js, err := json.Marshal(map[string][]string{
+		"data": data,
+	})
+
+	if err != nil {
+		util.ReturnHTTPError(res, req, httperror.NotFound.Status, fmt.Sprintf("%s: %s", httperror.ServerError.Code, err.Error()))
+		return
+	}
+
+	res.WriteHeader(http.StatusOK)
```
dramich: `Write` will do this for you, it only needs to be explicit if you want something other than OK

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,187 @@
...
+	credID := req.FormValue("cloudCredentialId")
+	if credID == "" {
+		return nil, fmt.Errorf("cloud credential required")
+	}
+
+	var accessCred client.CloudCredential
+	if err := access.ByID(v.generateAPIContext(req), &schema.Version, client.CloudCredentialType, credID, &accessCred); err != nil {
+		if apiError, ok := err.(*httperror.APIError); ok {
+			if apiError.Code.Status == httperror.PermissionDenied.Status || apiError.Code.Status == httperror.NotFound.Status {
+				return nil, fmt.Errorf("cloud credential not found")
```
dramich: The errors in here need to be the same basically. Otherwise this will leak if a cloud cred exists by that ID or not. If you can't find it or they can't access it, the error should be the same. If there is a legit error that can be returned.

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,187 @@
...
+	res.Write(js)
+}
+
+func (v *handler) getCloudCredential(req *http.Request) (*corev1.Secret, error) {
+	credID := req.FormValue("cloudCredentialId")
+	if credID == "" {
+		return nil, fmt.Errorf("cloud credential required")
+	}
+
+	var accessCred client.CloudCredential
```
dramich: Can you add a comment above this to clarify why we don't need it? It holds no info, the secret holds our world etc.

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+func getSoapClient(ctx context.Context, cc *corev1.Secret) (*govmomi.Client, error) {
+	u, err := url.Parse(fmt.Sprintf("https://%s:%s/sdk", cc.Data[dataFields["host"]], cc.Data[dataFields["port"]]))
+	if err != nil {
+		return nil, err
+	}
+
+	u.User = url.UserPassword(string(cc.Data[dataFields["username"]]), string(cc.Data[dataFields["password"]]))
+	c, err := govmomi.NewClient(ctx, u, true)
```
dramich: This call can be returned

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,386 @@
...
+
+func getContentLibraryManager(ctx context.Context, cc *corev1.Secret) (*library.Manager, error) {
+	soap, err := getSoapClient(ctx, cc)
+	if err != nil {
+		return nil, err
+	}
+
+	mgr := library.NewManager(rest.NewClient(soap.Client))
+	ui := url.UserPassword(string(cc.Data[dataFields["username"]]), string(cc.Data[dataFields["password"]]))
+	err = mgr.Login(ctx, ui)
```
dramich: This call can be returned

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,386 @@
...
+
+func getTagsManager(ctx context.Context, cc *corev1.Secret) (*tags.Manager, error) {
+	soap, err := getSoapClient(ctx, cc)
+	if err != nil {
+		return nil, err
+	}
+
+	mgr := tags.NewManager(rest.NewClient(soap.Client))
+	ui := url.UserPassword(string(cc.Data[dataFields["username"]]), string(cc.Data[dataFields["password"]]))
+	err = mgr.Login(ctx, ui)
```
dramich: This call can be returned

pkg/api/customization/vsphere/listers.go
```diff
@@ -0,0 +1,386 @@
...
+	return mgr, nil
+}
+
+func getCustomFieldsManager(ctx context.Context, cc *corev1.Secret) (*object.CustomFieldsManager, error) {
+	soap, err := getSoapClient(ctx, cc)
+	if err != nil {
+		return nil, err
+	}
+
+	mgr, err := object.GetCustomFieldsManager(soap.Client)
```
dramich: This call can be returned

pkg/api/customization/vsphere/handler.go
```diff
@@ -0,0 +1,187 @@
...
+
+func NewVsphereHandler(scaledContext *config.ScaledContext) http.Handler {
+	return &handler{
+		schemas:       scaledContext.Schemas,
+		secretsLister: scaledContext.Core.Secrets(namespace.GlobalNamespace).Controller().Lister(),
+	}
+}
+
+func (v *handler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
+	res.Header().Set("Content-Type", "application/json")
```
dramich: This can be moved down to before the res.Write call

dramich:
Overall I think the logic looks fine but the organization could use some refactoring. In the `ServeHTTP` method build a struct that holds the required info going forward. i.e. throw in the soap client, rest client assuming it doesn't need a login, the request, and the field name. This stops from having to pass the same things everywhere and cleans up the calls to something more like `newclient. processSoapFinder()` since everything lives off the struct. `getSoapFinder` can return its own struct and hang methods off that. This will save more passing around of the same things that everything uses and also helps with separation of concerns.

Issue #22106, opened 2019-08-09
CRD managed pods not shown in the UI in Group by Namespace/Workload view


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Enhancement

**Steps to reproduce (least amount of steps as possible):**
Launch pods managed by a CRD and go to the Workloads page. Can use this example from Elasticsearch (https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-quickstart.html) by running:
```
$ kubectl apply -f https://download.elastic.co/downloads/eck/0.9.0/all-in-one.yaml
$ cat <<EOF | kubectl apply -f -
apiVersion: elasticsearch.k8s.elastic.co/v1alpha1
kind: Elasticsearch
metadata:
  name: quickstart
spec:
  version: 7.2.0
  nodes:
  - nodeCount: 1
    config:
      node.master: true
      node.data: true
      node.ingest: true
EOF
```
**Result:**
It appears the UI or API will filter out all pods that are not managed by ReplicaSets, DaemonSets, StatefulSets, or unmanaged. If a pod is managed by a CRD, it is not displayed in the UI. The above example will set the following in the pod:
```yaml
  ownerReferences:
  - apiVersion: elasticsearch.k8s.elastic.co/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Elasticsearch
    name: quickstart
    uid: 59253a84-ba36-11e9-a46f-0af107b47750
```

**Other details that may be helpful:**
After running above YAML and also launching an unmanaged pod, see this in CLI:
```
$ kubectl get pods
NAME                       READY   STATUS    RESTARTS   AGE
quickstart-es-cd28s99bh4   1/1     Running   0          22m
ubuntu-pod                 1/1     Running   0          17m
```
But see this in UI:
![image](https://user-images.githubusercontent.com/468462/62813077-f05e7280-bad6-11e9-9dc8-17983b7cf869.png)

For showing these pods in the UI, would be good to note them as CRD managed and possibly give information on which CRD is managing the pod.

This issue may be somewhat related - https://github.com/rancher/rancher/issues/13323
**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.2.7
- Installation option (single install/HA):
N/A

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Custom
[...]

PR #23272, opened 2019-10-07
Skip setting workloadId for pods with non-workload owner refs

prachidamle:
https://github.com/rancher/rancher/issues/22106

Looking at adding a test.


leodotcloud:
LGTM

Issue #23285, opened 2019-10-08
Upgrade to 2.3.0 causes Projects with malformed certs to fail to display in the UI


rajha-korithrien:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Start with a Rancher v2.2.8 cluster.
Create a certificate (which I think is malformed) with a format similar to the following.
* tls.cert:
```text
Bag Attributes
    friendlyName: broker
    localKeyID: 3A 5E EF 3F 97 4A DF 1F C8 55 8B 6C 96 46 17 3F 28 36 58 6D 
subject=/C=US/ST=strogbadia/O=burniator/CN=pulp-qpid.paas-pulp.svc.cluster.local
issuer=/C=US/ST=strogbadia/O=burniator/CN=trogdor
-----BEGIN CERTIFICATE-----
REDACTED
-----END CERTIFICATE-----
```
* tls.key:
```text
Bag Attributes
    friendlyName: broker
    localKeyID: 3A 5E EF 3F 97 4A DF 1F C8 55 8B 6C 96 46 17 3F 28 36 58 6D 
Key Attributes: <No Attributes>
-----BEGIN PRIVATE KEY-----
REDACTED
-----END PRIVATE KEY-----
```
This certificate does not display correctly in the UI in v2.2.8 via: `Resources --> Certificates` however the Project containing it does display as expected.

After the upgrade to v2.3.0, when navigating to the Project with the malformed certificates the following error is displayed in a red box in the upper right of the UI:
```text
Status [422 Unprocessable Entity]. Body: [code=InvalidBodyContent, fieldName=certs, message=failed to parse certificate: asn1: structure error: tags don't match (16 vs {class:0 tag:2 length:1 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} tbsCertificate @2, baseType=error]
```
**Result:**
After upgrading to v2.3.0 certain projects can not be displayed in the Rancher UI due to what seems to be existing malformed certificates.

**Other details that may be helpful:**
When I use `kubectl` to manually edit the certificates in question such that they consist of only:
```text
-----BEGIN CERTIFICATE-----
REDACTED
-----END CERTIFICATE-----
```
and:
```text
-----BEGIN PRIVATE KEY-----
REDACTED
-----END PRIVATE KEY-----
```
[...]

PR #23287, opened 2019-10-08
Log cert parse error instead of returning err

mrajashree:
The secrets store calls AddCertInfo to parse cert so it populates all the necessary
UI fields. But if the cert is malformed AddCertInfo throws an error. This also
prevents the project in which this cert exists from loading in UI.
This commit logs the error message instead of returning error. This way the cert
will still not display in UI properly since it is malformed, but this won't stop
the project from loading.

https://github.com/rancher/rancher/issues/23285


alena1108:
Can we add a test for it @mrajashree

mrajashree:
Closing in favor of https://github.com/rancher/rancher/pull/23585

Issue #21039, opened 2019-06-21
Override cluster capabilities' pspEnabled if annotation present


ChrisHaPunkt:
**What kind of request is this (question/bug/enhancement/feature request):**
BUG

**Steps to reproduce (least amount of steps as possible):**
1. Import a rke cluster with activated PSP in rancher 2.2.4
2. Create a new Project/ Edit an existing Project

**Result:**

See no Dropdown for selecting PodSecurityPolicies
![image](https://user-images.githubusercontent.com/4389395/59903922-8a247000-9402-11e9-9463-07e4d26ff583.png)

**Other details that may be helpful:**

In the API view the setting is present and adjustable though.
![image](https://user-images.githubusercontent.com/4389395/59903976-ac1df280-9402-11e9-9a5f-164d9e63a90c.png)

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
- Installation option (single install/HA): HA / 2.2.4

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory): vvm 32GB / 8Cores
- Kubernetes version (use `kubectl version`): v1.13.5

```
(paste the output here)
```

- Docker version (use `docker version`):
```
18.06.3
```

deniseschannon:
After internal discussion, we do not want to blindly show adding PSPs on projects unless we can specifically detect that an RKE cluster has PSP enabled on the cluster admission controller.

jiaqiluo:
The bug is reproduced in  v2.3.1

Steps:
- use RKE to provision a cluster with PSP is enabled 

the snippet of the `rancher-cluater.yml` file
```
services:
  kube-api:
    pod_security_policy: true
```
- run Rancher v2.3.1 single install 
- add the cluster as an imported cluster into Rancher
- add a project
- edit the project, notice that there is no option for PSP
- view the project in API, notice the action `setpodsecuritypolicytemplate` is available, but somehow I cannot send the request from the UI (Error: parser error), but copy and execute from terminal succeeds
- refresh the project API page, see the new value is applied

jiaqiluo:
In Rancher v2.3-head 1b51a5911 with ui index `latest-2.3`, the new annotation `capabilities.cattle.io/pspEnabled= true` cannot be set via the UI.

there will be a new PR to change the name of the annotation.

jiaqiluo:
[...]

PR #23289, opened 2019-10-08
Capability override

rmweir:
Problem:
Users want to use psp feature on imported cluster's that have it enabled. Before providing UI for this, we need a way of know whether an imported cluster has it enabled.

Solution:
Provide a way of overriding capabilities via annotations.

Depends on:
https://github.com/rancher/types/pull/998

Issue:
https://github.com/rancher/rancher/issues/21039

### DIFF
--- a/pkg/controllers/management/cluster/clustercontroller.go
+++ b/pkg/controllers/management/cluster/clustercontroller.go
@@ -2,18 +2,24 @@ package cluster
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 
 	errorsutil "github.com/pkg/errors"
 	"github.com/rancher/kontainer-engine/service"
 	"github.com/rancher/kontainer-engine/types"
+	normantypes "github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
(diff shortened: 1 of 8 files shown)


pkg/controllers/management/cluster/clustercontroller.go
```diff
@@ -99,8 +109,11 @@ func (c *controller) capsSync(key string, cluster *v3.Cluster) (runtime.Object,
 		}
 
 		capabilities = toCapabilities(k8sCapabilities)
-	} else {
```
rmweir: this was removed bc imported cluster have neither generic or rke config

pkg/controllers/management/cluster/clustercontroller.go
```diff
@@ -109,11 +123,77 @@ func (c *controller) capsSync(key string, cluster *v3.Cluster) (runtime.Object,
...
+		}
+		return capabilities, nil
+	}
+
+	return oldCapabilities, nil
+}
+
+// parseResourceInterface converts a capability annotation to the appropriate type based on given schema.
+// Json format is assumed if type is not bool, string, or integer.
+func parseResourceInterface(key string, annoValue string, schema *normantypes.Schema) (interface{}, error) {
```
rmweir: we didn't want to force the user to json encode simple values

cjellick: Yeah, fundamentally this is ok

Issue #22379, opened 2019-08-23
Improve log message when jailer script fails to complete before timeout


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Enhancement

**Steps to reproduce (least amount of steps as possible):**

If the jailer script fails to complete within the configured timeout of 10 seconds the Rancher log will show a message of the following format (https://github.com/rancher/rancher/blob/v2.2.8/pkg/jailer/jailer.go#L55):

`[FATAL] error running the jail command: : signal: killed`

The request is to add a specific log message indicating the script reached the timeout.

**Environment information**
- Rancher v2.2.8

fengzixu:
Same problem

fengzixu:
Are the 10 seconds sufficient for processing jailer script? @axeal

paulsmithuk:
+1

Our AV is causing the script to timeout...personally, I'd rather give it a little more time to complete, rather than having a nicer fail message!

sangeethah:
Tested with `2.3-head`  
Following are the error message seen when  jailer script fails to complete before timeout:

```2019/10/11 23:48:26 [ERROR] NodeController c-ff8nn/m-dmf2n [node-controller] failed with : node create jail error: error running the jail command: timed out waiting for the script to complete: signal: killed```

<img width="1048" alt="Screen Shot 2019-10-11 at 4 48 36 PM" src="https://user-images.githubusercontent.com/4266958/66690795-41a9f000-ec47-11e9-90ed-02eb2128ca3d.png">

PR #23309, opened 2019-10-09
Make jailer timeout a setting

dramich:
Problem:
On some systems the default time of 10 seconds can cause the initial
jail creation to fail and Rancher won't be able to start
The error message is also not clear as to what failed

Solution:
Make the timeout a setting and also increase the default to 60 seconds
Update error message to clarify the failure is coming from the timeout

https://github.com/rancher/rancher/issues/23160
https://github.com/rancher/rancher/issues/22379

### DIFF
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -2,7 +2,6 @@ package jailer
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"os/exec"
 	"os/user"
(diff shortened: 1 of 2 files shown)


fizzyduck:
Thank you @dramich - much more comprehensive than my fix :smile:

luthermonson:
lgtm

leodotcloud:
LGTM

Issue #23241, opened 2019-10-04
[forwardport] Jailer script enhancements


dramich:
https://github.com/rancher/rancher/issues/23160

https://github.com/rancher/rancher/issues/22379

sowmyav27:
**verified on master-head latest - commit id: `a45722bb4`**

Currently the default setting for jailer-timeout is `60 seconds`.

- To simulate the jailer scripts failing to complete before timeout , did the following steps:
In the jailer script (/usr/bin/jailer.sh) , add a sleep of 15 seconds.
- Set jailer-timeout to 10 seconds ( < sleep set in jailer script).
- deploy a cluster with DO node driver.
- Node provisioning fails with following error message that states the reason:
- Following are the error message seen when jailer script fails to complete before timeout:
- Error: `Node create jail error: error running the jail command: timed out waiting for the script to complete: signal: killed`
- Error in logs:
`2019/10/21 21:48:52 [ERROR] NodeController c-jss7z/m-gzv8t [node-controller] failed with : node create jail error: error running the jail command: timed out waiting for the script to complete: signal: killed`

<img width="1306" alt="Screen Shot 2019-10-22 at 3 21 02 AM" src="https://user-images.githubusercontent.com/26032343/67246018-fe3b5700-f47a-11e9-92da-d1253b196ba1.png">

- Set `jailer-timeout` to 20 seconds ( > sleep set in jailer script)
- Provision a new node in the cluster. Provisioning of new node succeeds as expected.

PR #23310, opened 2019-10-09
[forwardport]Make jailer timeout a setting - master

dramich:
Problem:
On some systems the default time of 10 seconds can cause the initial
jail creation to fail and Rancher won't be able to start
The error message is also not clear as to what failed

Solution:
Make the timeout a setting and also increase the default to 60 seconds
Update error message to clarify the failure is coming from the timeout

https://github.com/rancher/rancher/issues/23241

### DIFF
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -2,7 +2,6 @@ package jailer
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


luthermonson:
lgtm

PR #23332, opened 2019-10-10
Housekeeping

alena1108:
Unused code cleanup, grammar errors correction, etc


pkg/controllers/management/compose/wait.go
```diff
@@ -1,33 +1,10 @@
...
-	"time"
-
-	"github.com/pkg/errors"
 	"github.com/rancher/norman/clientbase"
-	managementClient "github.com/rancher/types/client/management/v3"
 )
 
 var (
 	// WaitCondition is a set of function that can be customized to wait for a resource
 	WaitCondition = map[string]func(baseClient *clientbase.APIBaseClient, id, schemaType string) error{}
```
dramich: I think this whole file can be dropped. The only place this is used is to check if one exists, it's never assigned to in `rancher/pkg/controllers/management/compose/controller.go`

dramich:
lgtm

Issue #22684, opened 2019-09-06
Workloads are not removed on app deletion in k8s v1.16


gitlawr:
**Rancher Version**
v2.3.0-rc1

**Steps to reproduce (least amount of steps as possible):**
1. Go to `Global - Catalogs`, add a catalog from `https://github.com/gitlawr/helm-charts.git, branch: k8s-1.16`(Current community stable charts are not compatible with k8s v1.16, so this is a stable mysql chart that tweaks the deployment resource for k8s v1.16)
2. Install an app with `persistence.enabled=false` answers to the Default project. Wait till the app is installed and active.
3. Delete the app.

**Result:**
The app is deleted from the `apps` page, But the mysql deployment remains in the workload page.

**Other details:**
1. It works in a v1.15 k8s cluster.
2. I can apply and delete these resources successfully from kubectl(v1.13.1):
```
$ helm template ./stable/mysql --name=mysql --namespace=mysql \
    --set persistence.enabled=false | kubectl apply -f -
$ helm template ./stable/mysql --name=mysql --namespace=mysql \
    --set persistence.enabled=false | kubectl delete -f -
```
3. The resources in the chart include secret, service, and deployment. The secret and the service are deleted as expected.
4. There are no errors but warnings from rancher server logs:
```
2019/09/06 07:53:37 [INFO] Creating roleBinding User u-wi7dorfzny Role edit
[main] 2019/09/06 07:53:39 Starting Tiller v2.10+unreleased (tls=false)
[main] 2019/09/06 07:53:39 GRPC listening on :41799
[main] 2019/09/06 07:53:39 Probes listening on :35662
[main] 2019/09/06 07:53:39 Storage driver is ConfigMap
[main] 2019/09/06 07:53:39 Max history per release is 1
[tiller] 2019/09/06 07:53:39 getting history for release mysql-from-chart
[storage] 2019/09/06 07:53:39 getting release history for "mysql-from-chart"
Release "mysql-from-chart" does not exist. Installing it now.
[tiller] 2019/09/06 07:53:39 preparing install for mysql-from-chart
[storage] 2019/09/06 07:53:39 getting release history for "mysql-from-chart"
[tiller] 2019/09/06 07:53:39 rendering mysql chart using values
2019/09/06 07:53:39 info: manifest "mysql/templates/configurationFiles-configmap.yaml" is empty. Skipping.
2019/09/06 07:53:39 info: manifest "mysql/templates/servicemonitor.yaml" is empty. Skipping.
[...]

PR #23363, opened 2019-10-11
Update helm version

dramich:
Problem:
Helm needs to be updated and renamed to stop name colisions

Solution:
Rebase helm, update all calls to use rancher-helm

https://github.com/rancher/rancher/issues/21070
https://github.com/rancher/rancher/issues/22684


alena1108:
LGTM

PR #23364, opened 2019-10-11
Vsphere Lookup API

luthermonson:
Adding vSphere endpoints for the vSphere node template UI to call and help fill form fields with real answers from a vSphere server. This PR adds a new HttpHandler for /meta/vsphere and full documentation is in confluence.


dramich:
lgtm

PR #23386, opened 2019-10-11
Use environment variables instead of arguments

moelsayed:


alena1108:
LGTM

package/shell-setup.sh
```diff
@@ -48,7 +44,7 @@ current-context: "Default"
 users:
 - name: "Default"
   user:
-    token: "${token}"
+    token: "${TOKEN}"
 EOF
```
luthermonson: maybe to be safe you should unset the 3 env vars when this script exits?

luthermonson:
please close in favor of https://github.com/rancher/rancher/pull/23749

Issue #23365, opened 2019-10-11
New single-node rancher installs using let's encrypt fail


cjellick:
If you use the Let's Encrypt certificate option for single node rancher installs, it will fail. I'm referring to the Let's Encrypt option specified in this doc:
https://rancher.com/docs/rancher/v2.x/en/installation/single-node/

The cause of this is that Let's Encrypt is EOLing their v1 API. This is detailed here:
https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430

And a schedule of "brown outs" is detailed here:
https://calendar.google.com/calendar/embed?src=letsencrypt.org_caqskun93lgiabjj4ba9cb1rek%40group.calendar.google.com

You see this error:
```
2019/10/10 23:33:10 [INFO] 2019/10/10 23:33:10 http: TLS handshake error from 63.232.146.98:58877: 403 urn:acme:error:unauthorized: Account creation on ACMEv1 is disabled. Please upgrade your ACME client to a version that supports ACMEv2 / RFC 8555. See https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430 for details.
```

cjellick:
@drpebcak found this, so just tagging him here.
The wording seems to indicate that _existing_ installs should continue to work until we have a fix, but we don't have a great way of verifying that.

cjellick:
Upstream: https://github.com/golang/go/issues/21081

codejamninja:
I'm having the same problem

https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430

jmhodges:
I don't use rancher, but I poked around since this ticket[1] was linked to another I answered on. It looks like the current autocert code rancher uses doesn't make this easy to workaround. 

If I've understood the code correctly, rancher relies on `autocert.Manager` to create the `acme.Client`, but, by default, it uses the constant for the API URL that's currently not been updated (see [1] again).

So, rancher will have to learn how to create it's own `acme.Client` and set the DirectoryURL like I describe in [a comment on the other ticket](https://github.com/golang/go/issues/21081#issuecomment-540855037). It'll also have to handle not reusing v01 accounts (if it saves the keys for them) because those accounts won't work in v02.

[1] https://github.com/golang/go/issues/21081

codejamninja:
Just curious, but why is this being worked on now when letsencrypt scheduled to deprecate it today?

alena1108:
[...]

PR #23387, opened 2019-10-11
Update crypto version round 2

dramich:


deniseschannon:
https://github.com/rancher/rancher/issues/23365

PR #23400, opened 2019-10-12
Revert "Revert "Update helm version""

deniseschannon:
Reverts rancher/rancher#23399 - This was backed out of the release branch for v2.3.1, so it needs to be added back in after v2.3.1 is released. 

Update helm version
    
    Problem:
    Helm needs to be updated and renamed to stop name colisions
    
    Solution:
    Rebase helm, update all calls to use rancher-helm


deniseschannon:
https://github.com/rancher/rancher/issues/22684

Issue #18638, opened 2019-03-06
Gometallinter will be deprecated on 04-07-2019


daxmc99:
See https://github.com/alecthomas/gometalinter/issues/590
While releases will still be available we should consider other linters.

dramich:
Suggested replacement is https://github.com/golangci/golangci-lint

daxmc99:
As of c0e17880540a115151c5edb8712e06dea1906da4 this is still tbd. 
golangci-lint has issues with memory consumption and would break local dev workflow with `make ci` taking a very long time to run locally

superseb:
https://github.com/golangci/golangci-lint/releases/tag/v1.20.0 contains multiple memory usage fixes

dramich:
Technically this was fixed by going back to standard go tools but with the memory fixes in golangci we are going to give it another go.

PR #23448, opened 2019-10-15
Switch back to golangci

dramich:
This switches rancher back to golangci and runs the same linters we were running before. It also adds a change to stop printing the rancher logs after the tests run if the tests do not fail. This saves ~6 min in build time.


dramich:
https://github.com/rancher/rancher/issues/18638

Dockerfile.dapper
```diff
@@ -53,12 +51,6 @@ RUN curl -sLf ${!HELM_URL} > /usr/bin/helm && \
 
 RUN wget -O - ${!DOCKER_URL} > /usr/bin/docker && chmod +x /usr/bin/docker
 
-# FIXME: Update to Rancher RKE when released
```
daxmc99: Why did we have this before?

dramich: This was added previously to spin up clusters in tests to validate the cluster auth endpoint, none of that is being used and won't be needed anyways with k3s.

alena1108:
LGTM

Issue #23427, opened 2019-10-14
Rancher's local system-library is empty with volume mount.


bentastic27:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

Simply spinning up Rancher with the following triggers it:

```
docker run -d --restart=unless-stopped -p 80:80 -p 443:443 -v /opt/rancher:/var/lib/rancher -e CATTLE_SYSTEM_CATALOG=bundled rancher/rancher:v2.3.0
```
You can see that it's empty with the following:

```
$ docker exec -ti <RANCHER CONTAINER> ls management-state/local-catalogs/system-library/
```

**Result:**

The system-library directory will be empty when you mount `/var/lib/rancher` and use the `CATTLE_SYSTEM_CATALOG` variable. Rancher will also complain about the directory being empty when managing system charts.

**Other details that may be helpful:**

This doesn't seem to happen when not mounting `/var/lib/rancher`.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 
2.3.0
- Installation option (single install/HA):
single

dnoland1:
This also seems to be an issue with a rancher upgrade which uses `--volumes-from rancher-data`.

sangeethah:
Scenario to validate:

Scenario 1:
Create a single node rancher server install using `v2.2.8` .
Create a 1 cluster and deploy apps.

Upgrade rancher server to v2.3.0 using  `-e CATTLE_SYSTEM_CATALOG=bundled`
```
docker run -d --volumes-from rancher-data \
  --restart=unless-stopped \
    -p 80:80 -p 443:443 \
    rancher/rancher:<RANCHER_VERSION_TAG> -e CATTLE_SYSTEM_CATALOG=bundled
```

Once rancher server is upgraded , we will not be able to see any system catalogs. ( Reproducing this issue).

Upgrade rancher server to `v2.3.1-rc5` using  `-e CATTLE_SYSTEM_CATALOG=bundled`
Once rancher server is upgraded , we will not be able to see system catalogs. 
Should be able to enable monitoring.

Scenario 2:
Create a single node rancher server install using `v2.2.8` .
Create a 1 cluster and deploy apps.
Upgrade rancher server to v2.3.1 using  `-e CATTLE_SYSTEM_CATALOG=bundled`
Once rancher server is upgraded , we will not be able to see system catalogs. 
[...]

PR #23450, opened 2019-10-15
Move location of bundled catalogs

dramich:
Problem:
When bind mounting `/var/lib/rancher` the catalog cache won't exist on the
local filesystem

Solution:
Move the catalog cache to a location we don't tell people to bind mount
like `/var/lib/rancher-data`

https://github.com/rancher/rancher/issues/23427


alena1108:
@dramich assuming the new way work out of the box for upgraded setups, but wanted to double check just in case.

PR #23487, opened 2019-10-16
[backport] Switch back to golangci

dramich:
This switches rancher back to golangci and runs the same linters we were running before. It also adds a change to stop printing the rancher logs after the tests run if the tests do not fail. This saves ~6 min in build time.

https://github.com/rancher/rancher/issues/18638


alena1108:
LGTM. Merge after 2.3.1 ships

Issue #23424, opened 2019-10-14
Compatibility issues with cert-manager v0.11.0


ahmedwaleedmalik:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug/request

**Steps to reproduce (least amount of steps as possible):**
Install cert-manager v0.11.0 by following the guidelines from https://docs.cert-manager.io/en/latest/getting-started/install/kubernetes.html

**Result:**
The issue being that in cert-manager v0.11.0 the label `certmanager.k8s.io` has been replaced with `cert-manager.io`. But this is yet to be handled the helm chart for rancher.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.3.0
- Installation option (single install/HA): HA

mschickervxob:
I updated cert-manager because of the advice from rancher regarding 1st of November. Cert-manager was on 0.5.2. Now I have 0.11.0 and cannot upgrade from rancher 2.2.8 to 2.2.9 via helm.

gregbacchus:
@mschickervxob if it helps, downgrading to cert-manager 0.10 should resolve this issue. It's caused by the changing of the cert-manager CRDs in v0.10. So if downgrading, make sure that you install the earlier CRDs as well.

mschickervxob:
Thanks @gregbacchus - but this also means to change all resources to the old format certmanager.k8s.io from cert-manager.io again and later again back which I prefer not to do.

gregbacchus:
Yes, I can understand that. It may be possible to run both versions of cert-manager, but I haven't tried (and probably wouldn't).

deniseschannon:
Update to the latest cert-manager version as this was written awhile ago and cert-manager updates very quickly. 

Create docs on how to upgrade to cert-manager v0.11.0+

Test the current docs for all installation and update wherever we indicate v0.9.1.
https://rancher.com/docs/rancher/v2.x/en/installation/

dnauck:
Currently it's not possible to install cert-manager v0.11, see #23850

cipianpascu:
Thank you for working on this, guys. And, don't forget to update rancher-latest helm chart also.
Kind regards,
Ciprian

deniseschannon:
On hold until docs are written of how to use the new cert-manager version for HA installs

jiaqiluo:
Follow the new documentation to install Rancher in HA mode successfully:

- Rancher version: v2.3.4-rc3
[...]

PR #23501, opened 2019-10-16
check capabilities for cert-manager changes

drpebcak:
Relies on https://github.com/rancher/helm-unittest/pull/5


drpebcak:
Relates to #23424

Issue #23303, opened 2019-10-09
[forwardport] fluentd pod CPU is high when enable plenty of project logging.


loganhz:
forwardport https://github.com/rancher/rancher/issues/23113

alena1108:
@aiwantaozi what kind of regression test qa need to do, in addition to validating the original issue?

aiwantaozi:
@alena1108 Hi, the step in this issue is enough for testing https://github.com/rancher/rancher/issues/23113, could test one more target link elasticsearch for regression test

bmdepesa:
rancher/rancher:master-head 48a2c4ffc

* Enabled cluster logging w/ fluentd target
* Enabled project monitoring w/ elasticsearch target on multiple projects
* All targets are receiving logs

PR #23502, opened 2019-10-17
[forwardport] Enhance log config

aiwantaozi:
Forward port pr for #23114

Problem:
1. Project match source use all file under /var/log/containers which include log files not belong to this project
2. When configuring read_from_head, if target file is large, it takes a long time and starting other plugins isn't executed until reading file is finished.
3. Default file chunk size is 256M, kafka plugin will iterate over each 256M file and deliver each 1024K, it takes a long time if file is too large

Solution:
1. Only tail file belong to this project
2. Remove read_from_head
3. reduce each chunk file size for kafka, set default kafka aggregation threshold higher

Issue:
https://github.com/rancher/rancher/issues/23303

### DIFF
--- a/pkg/api/customization/logging/validator.go
+++ b/pkg/api/customization/logging/validator.go
@@ -75,8 +75,10 @@ func validate(level, containerLogSourceTag string, loggingTargets v3.LoggingTarg
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


alena1108:
Forward port pr for https://github.com/rancher/rancher/pull/23436

PR #23515, opened 2019-10-17
bump to v2.3.1

fgeorgatos:
i'm pretty sure the PR workflow is somewhat different, but someone will notice it...


alena1108:
LGTM

PR #23526, opened 2019-10-17
Add wait in projectcatalog access check test

mrajashree:
All other tests that create prtb/crtb for a user have this wait check before using the user's client. So adding the same wait check here too will fix the flakiness


alena1108:
LGTM

Issue #23485, opened 2019-10-16
Make RC builds of rancher server point to "dev" branch of kontainer-driver-metadata.


sangeethah:
Make RC builds point to "dev" branch of kontainer-driver-metadata.

Currently it gets set to "master" branch and we have to manually change it  to "dev" branch once RC build is installed.

deniseschannon:
@alena1108 can you coordinate with @cloudnautique  on what would need to happen? If we can't do it in time for 2.3.2, let's push it out to 2.3.3

sangeethah:
Reopening to get this  PR merged https://github.com/rancher/rancher/pull/23606

sangeethah:
Rancher server built with latest build from `v2.3-head`  - rke-metadata-config points to "dev" branch.
Rancher server built with latest build from `v2.3-head` using server_version as `v2.3.2-rc1` - rke-metadata-config points to "dev" branch
Rancher server built with latest build from `v2.3-head` using server_version as `v2.3.2-alpha1` - rke-metadata-config points to "dev" branch

Rancher server built with latest build from `v2.3-head` using server_version as `v2.3.2` - rke-metadata-config points to "master" branch

PR #23545, opened 2019-10-18
use master branch of metadata only for release server versions

kinarashah:
https://github.com/rancher/rancher/issues/23485

### DIFF
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -3,9 +3,11 @@ package settings
 import (
 	"encoding/json"
 	"os"
+	"strconv"
 	"strings"
 
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/settings/setting.go
```diff
@@ -38,7 +38,7 @@ var (
 	PeerServices                      = NewSetting("peer-service", os.Getenv("CATTLE_PEER_SERVICE"))
 	RDNSServerBaseURL                 = NewSetting("rdns-base-url", "https://api.lb.rancher.cloud/v1")
 	RkeVersion                        = NewSetting("rke-version", "")
-	RkeMetadataConfig                 = NewSetting("rke-metadata-config", getMetadataConfig())
+	RkeMetadataConfig                 = NewSetting("rke-metadata-config", "")
```
alena1108: will it be reset on the upgraded setups?

pkg/settings/setting.go
```diff
@@ -143,19 +143,36 @@ func NewSetting(name, def string) Setting {
...
 		"url":                      "https://github.com/rancher/kontainer-driver-metadata.git",
-		"branch":                   "master",
+		"branch":                   branch,
 		"refresh-interval-minutes": "1440",
 	}
 	ans, err := json.Marshal(data)
-	if err == nil {
-		return string(ans)
+	if err != nil {
+		return ""
```
alena1108: do you think we should log an error? So it doesn't get lost

pkg/settings/setting.go
```diff
@@ -143,19 +143,36 @@ func NewSetting(name, def string) Setting {
...
 
-func GetEnvKey(key string) string {
-	return "CATTLE_" + strings.ToUpper(strings.Replace(key, "-", "_", -1))
+func releaseServerVersion(serverVersion string) bool {
+	if serverVersion == "dev" ||
+		serverVersion == "master" ||
+		serverVersion == "" ||
+		strings.HasSuffix(serverVersion, "-head") ||
+		strings.Contains(serverVersion, "-rc") {
+		return false
```
prachidamle: do we have any release versions like "alpha" / "beta"?

pkg/settings/setting.go
kinarashah: updated to log error

pkg/settings/setting.go
kinarashah: updated logic to calculate server version

alena1108:
LGTM

prachidamle:
LGTM

pkg/settings/setting.go
deniseschannon: does this account for `v2.4.0-alpha1` as that's what the alpha versions which are used for preview will be tagged.

Issue #23511, opened 2019-10-17
OpenStack Node Driver - "Waiting for SSH to be available" when using privateKeyFile


Paic:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

- Create a new cluster with OpenStack node provider using node template with the privateKeyFile specified.
- Wait for instance creation

**Result:**

- All nodes errored : "Too many retries waiting for SSH to be available"

**Other details that may be helpful:**

After some digging, it looks like the rancher server strip a new line character at the end of the `privateKeyFile` content (after `-----`), which make the private key invalid.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.3.0
- Installation option (single install/HA): HA

dramich:
This appears to be the UI dropping the last return character based on looking at the call the UI makes to the API.

Paic:
When I looked into the request sent (in v2.3.0), the last return character was present.
The bug is present in rancher v2.2.8 and the server return the privateKeyFile attribute without this character.

Paic:
Hey 👋
Thanks for the quick fix.

Is this included in v2.3.2 or is it planned for a next release ?

izaac:
Rancher version: `2.4 master-head (12/03/2019)` 5845c2d4a

I see the newlines at the end of the `cacert` and `userDataFile` getting truncated. `privateKeyFile` (id_rsa) looks good in latest master.

The expectation for both `cacert` and `userDataFile` is not to get their trailing end of line truncated. I tested this also in v2.3.3 and those are good in that version.

dramich:
The fix to apply this to all fields wasn't ported to master. It is now.

izaac:
I've checked this with Rancher versions:

Rancher version `v2.4 master-head (02/28/2020)` commit id: e59944099
Rancher version `v2.3.5`

The trailing new line is not truncated, and both versions behave identically now for files:

`cacert` `id_rsa`  `userDataFile`

Closing.

PR #23549, opened 2019-10-18
Add newlines to ssh keys for nodetemplates

dramich:
Problem:
When parsing nodetemplates initially new lines are dropped from the end
of ssh keys

Solution:
Add them back before writing them out for docker-machine usage

This is norman parsing the initial request so there isn't a good way to just not drop them as it would affect all other parsing. 

https://github.com/rancher/rancher/issues/23511


alena1108:
LGTM

PR #23551, opened 2019-10-19
Save a Running Cluster to a ClusterTemplate and ClusterTemplateRevision

prachidamle:
https://github.com/rancher/rancher/issues/22992

Depends on types PR: https://github.com/rancher/types/pull/1009


pkg/api/customization/cluster/actionhandler.go
```diff
@@ -40,6 +46,19 @@ func (a ActionHandler) ClusterActionHandler(actionName string, action *types.Act
...
 	}
 
+	canCreateClusterTemplate := func() bool {
+		ma := gaccess.MemberAccess{
+			Users:     a.Users,
+			GrLister:  a.GrLister,
+			GrbLister: a.GrbLister,
+		}
+		callerID := apiContext.Request.Header.Get(gaccess.ImpersonateUserHeader)
+		if canCreateTemplates, err := ma.CanCreateRKETemplate(callerID); err == nil {
```
alena1108: we can just return without checking for error. Unless we want to add logging for error or return it

pkg/api/customization/cluster/actionhandler.go
```diff
@@ -81,6 +100,14 @@ func (a ActionHandler) ClusterActionHandler(actionName string, action *types.Act
 		return a.RotateCertificates(actionName, action, apiContext)
 	case v3.ClusterActionRunCISScan:
 		return a.runCISScan(actionName, action, apiContext)
+	case v3.ClusterActionSaveAsTemplate:
+		if !canUpdateCluster() {
+			return httperror.NewAPIError(httperror.PermissionDenied, "can not save as an RKETemplate")
```
alena1108: nit: do you think we rephrase the error to "can not save cluster as an RKE template"?

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,159 @@
...
+			fmt.Sprintf("cluster with id %v doesn't exist", apiContext.ID))
+	}
+
+	if cluster.DeletionTimestamp != nil {
+		return httperror.NewAPIError(httperror.InvalidType,
+			fmt.Sprintf("cluster with id %v is being deleted", apiContext.ID))
+	}
+	if !v3.ClusterConditionReady.IsTrue(cluster) {
+		return httperror.WrapAPIError(err, httperror.ClusterUnavailable,
+			fmt.Sprintf("cluster not ready"))
```
alena1108: should we add an id to the error message?

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	//Copy the cluster rke spec to a template and revision object
+
+	actionInput, err := parse.ReadBody(apiContext.Request)
+	if err != nil {
+		return err
+	}
+	clusterTempName, ok := actionInput["clusterTemplateName"].(string)
```
alena1108: do you think it would make sense to validate input params before trying to fetch the cluster from api? So in case input is invalid, we can save on api call

alena1108: Also would be nice to split this method to multiple subcalls, as it's a bit long. I can think of validateActionInput, validateCluster, createClusterTemplateRevision, updateCluster.

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,159 @@
...
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("failed to create RKE Template object"))
+	}
+
+	newTemplateRevObj := newClusterTemplateRevision(clusterTempRevName, cluster, clusterTemplate, creatorID)
+	clusterTemplateRevision, err := a.ClusterTemplateRevisionClient.Create(newTemplateRevObj)
+	if err != nil {
+		//delete the clusterTemplate created above
+		deleteErr := a.ClusterTemplateClient.Delete(clusterTemplate.Name, &metav1.DeleteOptions{})
+		if deleteErr != nil {
```
alena1108: do you think we can add retries to delete call?

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,159 @@
...
+			logrus.Errorf("Failed to delete the RKE Template %v", clusterTemplate.Name)
+		}
+
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("failed to create RKE Template Revision object"))
+	}
+
+	updatedCluster := cluster.DeepCopy()
+	updatedCluster.Spec.ClusterTemplateRevisionName = ref.Ref(clusterTemplateRevision)
```
alena1108: Nit: do you think we can add it to the beginning of the loop? lines 84-86 are same as 100-102. So it can be something:

for number of retries:
 fetch cluster 
 update
 if error  - retry

tests/integration/suite/test_cluster_defaults.py
```diff
@@ -96,3 +96,41 @@ def test_rke_k8s_deprecated_versions(admin_mc, remove_resource):
                                     'v1.8.10-rancher1-1 is deprecated'
     client.update_by_id_setting(id=deprecated_versions_setting.id,
                                 value="")
+
```
alena1108: @prachidamle do you think it would make sense to add a test for the action handler? To see if it gets disabled when templateId is set on the cluster

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			logrus.Errorf("error fetching cluster with id %v: %v", apiContext.ID, err)
+			continue
+		}
+		updatedCluster = cluster.DeepCopy()
+		updatedCluster.Spec.ClusterTemplateRevisionName = ref.Ref(clusterTemplateRevision)
+		updatedCluster.Spec.ClusterTemplateName = ref.Ref(clusterTemplate)
+	}
+
+	apiContext.WriteResponse(http.StatusNoContent, map[string]interface{}{})
```
alena1108: should we return an explicit api error to the user if the update cluster with the template failed?

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,159 @@
...
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("failed to create RKE Template Revision object"))
+	}
+
+	updatedCluster := cluster.DeepCopy()
+	updatedCluster.Spec.ClusterTemplateRevisionName = ref.Ref(clusterTemplateRevision)
+	updatedCluster.Spec.ClusterTemplateName = ref.Ref(clusterTemplate)
+
+	// Can't add either too many retries or longer interval as this an API handler
+	for i := 0; i < NumberOfRetriesForClusterUpdate; i++ {
```
alena1108: I feel like we should run the template/revision cleanup logic after attempting the cluster.update failure. You can wrap the cleanup logic in its own cleanupcall, and call it in defer 

```
defer cleanup()
 if err := nil  {
 return
 }
if clusterRevisionId != nil {
clenaup revision
}
if clusterTemplateId != nil {
cleanupTemplate
}

```

pkg/api/customization/cluster/actionhandler.go
prachidamle: In case of error when figuring out if the user can create templates, this check will not let user have ability to invoke this action. Does that make sense?

tests/integration/suite/test_cluster_defaults.py
prachidamle: done

pkg/api/customization/cluster/actions_rketemplate.go
prachidamle: done

pkg/api/customization/cluster/actions_rketemplate.go
prachidamle: done

pkg/api/customization/cluster/actions_rketemplate.go
prachidamle: done

pkg/api/customization/cluster/actions_rketemplate.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/rancher/rancher/pkg/ref"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	mgmtclient "github.com/rancher/types/client/management/v3"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	NumberOfRetries   = 3
```
alena1108: nit: do you think it should be lowercase as it's used only within this file?

pkg/api/customization/cluster/actionhandler.go
prachidamle: got it, returning the result directly.

alena1108:
LGTM

PR #23561, opened 2019-10-21
Keep chart repo URL intact

sauber:
Don't change the repo URL given by user.

### DIFF
--- a/pkg/catalog/helm/helm.go
+++ b/pkg/catalog/helm/helm.go
@@ -108,8 +108,6 @@ func (h *Helm) request(pathURL string) (*http.Response, error) {
 }
 
 func (h *Helm) downloadIndex(indexURL string) (*RepoIndex, error) {
-	indexURL = strings.TrimSuffix(indexURL, "/")
-	indexURL = indexURL + "/index.yaml"
 	resp, err := h.request(indexURL)
 	if err != nil {
 		if e, ok := err.(net.Error); ok && e.Timeout() {


pkg/catalog/helm/helm.go
```diff
@@ -108,8 +108,6 @@ func (h *Helm) request(pathURL string) (*http.Response, error) {
 }
 
 func (h *Helm) downloadIndex(indexURL string) (*RepoIndex, error) {
-	indexURL = strings.TrimSuffix(indexURL, "/")
-	indexURL = indexURL + "/index.yaml"
```
dramich, line 112: Unfortunately this won't work with the other 99% of cases if we do the change. The url coming in is just where the catalog lives, users are not including index.yaml in that url and thus it will fail to get what we really need in the call.

sauber: @dramich , which other 99% are you referring to? On https://landscape.cncf.io/category=container-registry&format=card-mode, Harbor is the only graduated registry project. It has nearly as many github stars alone, as all the other projects combined. I acknowledge that registry server and chartmuseum is not necessarily the same, but chartmuseums do not yet have own category on cncf.

Could the code try both with and without the /index.yaml appendix, to support both scenarios?

dramich: @sauber This doesn't have anything to do with being harbor, the issue is dropping the `index.yaml` from the GET request. Per [helm docs](https://helm.sh/docs/topics/chart_repository/) there needs to be an `index.yaml` file at the base of the repo, which is the file Rancher requires in order to process the repo and grab all charts. I would suggest seeing where your `index.yaml` lives, then provide a url based off of that.

sauber: @dramich , you're right. Thanks for explaining the requirement. Indeed Harbor exposes the charts in index.yaml under https://core.harbor.domain/chartrepo . I withdraw my pull request.

Issue #23441, opened 2019-10-15
2.3.0 upgrade - cattle-cluster-agent & cattle-node-agent CrashLoopBackOff


modemhead:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Upgraded High Availability (HA) Install , 2.2.8 to 2.3.0 using helm.

helm upgrade rancher rancher-latest/rancher --set hostname=rancher-dev.corp****--set tls=external --set auditLog.level=3

**Result:**
cattle-cluster-agent & cattle-node-agent entered state CrashLoopBackOff 

**Other details that may be helpful:**
INFO: Environment: CATTLE_ADDRESS=10.42.2.68 CATTLE_CA_CHECKSUM=3c5745f051626ff302c53d93c*******e7f26896f518d4b653bbc2 CATTLE_CLUSTER=true CATTLE_INTERNAL_ADDRESS= CATTLE_K8S_MANAGED=true CATTLE_NODE_NAME=cattle-cluster-agent-7bbdc67b56-d97gq CATTLE_SERVER=https://rancher-dev.corp*********
INFO: Using resolv.conf: nameserver 10.43.0.10 search cattle-system.svc.cluster.local svc.cluster.local cluster.local  options ndots:5
INFO: https://rancher-dev.corp*********/ping is accessible
INFO: rancher-dev.corp********* resolves to ****
ERROR: The environment variable CATTLE_CA_CHECKSUM is set but there is no CA certificate configured at https://rancher-dev.corp*****/v3/settings/cacerts

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):2.3.0
- Installation option (single install/HA):HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke / centos 7 / vmware
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM 16gb RAM
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #23607, opened 2019-10-22
resync cluster on cacert change

kinarashah:
observed on ha setup: agents get deployed with old value before
cacerts update in settings, adding updating cluster with
forceDeploy annotation to redeploy agent.

https://github.com/rancher/rancher/issues/23441

### DIFF
--- a/pkg/api/controllers/dynamiclistener/listener_controller.go
+++ b/pkg/api/controllers/dynamiclistener/listener_controller.go
@@ -2,11 +2,13 @@ package dynamiclistener
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/rancher/pkg/cert"
+	"github.com/rancher/rancher/pkg/controllers/management/clusterdeploy"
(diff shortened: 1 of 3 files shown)


pkg/api/controllers/dynamiclistener/listener_controller.go
```diff
@@ -98,7 +104,11 @@ func (c *Controller) enable(listener *v3.ListenConfig) error {
 
 func (c *Controller) updateCurrent(listener *v3.ListenConfig) error {
 	caCerts := strings.TrimSpace(listener.CACerts)
-	settings.CACerts.Set(caCerts)
+	toUpdate := false
+	if settings.CACerts.Get() != caCerts {
```
alena1108: after an offline discussion, we have decided to not only call update when the values are different, but also if the something had failed during the update in lines 137-158

kinarashah: updated to place `toUpdate` outside `updateCurrent`, will be set to false on L158, only after the cluster sync logic successfully goes through

alena1108:
LGTM

PR #23625, opened 2019-10-23
[2.3.2] Change istio-virtual-service-ui to true by default

loganhz:
Change istio-virtual-service-ui to true by default

https://github.com/rancher/rancher/issues/23582


deniseschannon:
LGTM

Issue #23666, opened 2019-10-25
[forwardport] use master branch of metadata only for release server versions


kinarashah:
https://github.com/rancher/rancher/pull/23606

jiaqiluo:
The original issue https://github.com/rancher/rancher/issues/23485

jiaqiluo:
wait for the coming preview release to validate the fix, so we don't have to build customized images

jiaqiluo:
Build docker images from Rancher master branch c1c0f7f1d0d8f122d66c1951abfc53bc7cce5dfc

| server-version | metadata branch |
| --- | --- |
| v2.4.1 | master |
| v2.4.1-rc1 | dev |
| v2.4.1-alpha1 | dev |
| dev | dev |

PR #23667, opened 2019-10-25
use master branch of metadata only for release server versions

kinarashah:
https://github.com/rancher/rancher/issues/23666


alena1108:
LGTM

PR #23669, opened 2019-10-25
Change istio-virtual-service-ui feature flag to true by default

loganhz:
Change istio-virtual-service-ui to true by default

https://github.com/rancher/rancher/issues/23668


dramich:
lgtm

Issue #23684, opened 2019-10-25
Forwardport: Login fails for users with escaped comma in CN when using Active Directory LDAP authentication


mrajashree:
Forward port: https://github.com/rancher/rancher/issues/22192

aaronRancher:
The feature is validated in `rancher\rancher:master-head` `version d6a9d4509`.

Enable authentication with `Active Directory`.
Connect to AD management server via Microsoft Remote Desktop
Add users with escaped comma.
Login as user with comma in dn.
Lock back in as admin and check that a "User" account is created on the Global->Users page for this user with comma in dn.
While on the Global->Users page click on the "Refresh Group Memberships" button.
Check Rancher's log to make sure there are no errors.

![Screen Shot 2019-11-25 at 8 58 39 AM](https://user-images.githubusercontent.com/57409052/69556715-7b6a5800-0f62-11ea-9437-7daa2f5674e7.png)

PR #23685, opened 2019-10-25
Forward port: Remove invalid parsing of distinguishedName in AD refresh principals

mrajashree:
ExternalID is the distinguishedName of a user, which is being used as the baseDN
in ldap search query in RefetchGroupPrincipals. This doesn't need to be split on `\`.
That split is only for when we're using some field that has "logindomain\username"
format. Splitting incorrectly on `\` removed a part of the distinguishedName leading
to ldap errors

https://github.com/rancher/rancher/issues/23684

forward port of https://github.com/rancher/rancher/pull/23035


alena1108:
LGTM

PR #23718, opened 2019-10-28
Admin node templates 2.3

rmweir:
__Previously merged to 2.3.x__

**Problem:**
Admins should be able to CRUD node templates

**Solution:**
Make node templates rbac enforced so they can be exposed to admins, and migrate node templates to a global namespace.

**Issue:**
https://github.com/rancher/rancher/issues/12186


dramich:
lgtm

Issue #16216, opened 2018-10-20
Ability to create global roles and setting them as a default


mitchellmaler:
We are trying to setup the global user role so that user cannot create a cluster. I tried setting the default user to User Base just so they can log in and it seemed to work. The issue is now if this AD user tries to add a new Member to a project (cause they were added as a Project Owner) they cannot list any users and each AD user already added has an error. 

My goal is to give the user as minimum of access when the log in the first time but then can request access from either an Admin or Project/Cluster Owner to be added.

![2018-10-19 22_31_48-rancher](https://user-images.githubusercontent.com/4107699/47250849-e9069500-d3ee-11e8-9894-83507c073e45.png)

mitchellmaler:
The main issue is I cannot customize or create any new global roles.

loganhz:
I can reproduce this issue in master

Steps to reproduce:

1. Go to global role page and make `User Base` role as default role. And make the `User` role to non-default role.
2. Create two users `user1` and `user2` with custom global permission `Login` only.
3. Assign `user1` to `Default` project with `Project owner` role.
4. Login as `user1` and try to add `user2` as a `Project Member`

Results:

`user1` can not find `user2` when try to search and find it.
Also it shows an error for fetching the the default admin user info in project member page.

![image](https://user-images.githubusercontent.com/21168270/47263166-bf548780-d52e-11e8-92eb-784a6f7a3a35.png)

Expected:

`user1` should be able to add `user2` as a `Project Member` as `user1` is a `Project Owner`
Also, `user1` should be able to see the user info for all users in the project instead of seeing an error with `Unable to fetch user info`

cjellick:
@loganhz the Login role (aka User Base) behaving as expected. It is extremely limited permissions and we intend for someone with only that role to not be able to see/search other users.

@mitchellmaler to accomplish what you want, you basically need the ability to create a new global role. We don't let you edit the built-in ones because we need to keep those in-sync and in-line with what Rancher expects.
When we originally design this Global Roles feature, we intentionally locked them down to prevent admins from creating new ones. Our thinking/motivation was to prevent users from shooting themselves in the foot and that there should be no need to customize them.
I think this was wrong and I think we should/will open it up so that you can create new global roles. We will probably give you the ability to clone an existing one, so that you can basically clone the User role and remove whatever permissions you don't want in it.

[...]

PR #23734, opened 2019-10-29
Create global roles

rmweir:
**MERGE ORDER:**
merge types-> update rancher w/ types-> merge rancher
**Problem:**
Current global roles do not have desired set of permissions for default role

**Solution:**
Allow admin to create global roles.

**Dependencies:**
https://github.com/rancher/types/pull/1018

**Issue:**
https://github.com/rancher/rancher/issues/16216


dramich:
The validation of requests against builtin roles should follow the same pattern as roletemplates as seen [here](https://github.com/rancher/rancher/blob/3b22f8501526690bf470e721e6a2d592ef7880ca/pkg/api/customization/roletemplate/roletemplate.go#L19:18).

tests/integration/suite/test_global_roles.py
```diff
@@ -0,0 +1,24 @@
+def test_builtin_default_can_be_edited(admin_mc, remove_resource):
```
dramich: This test needs to undo anything it does, a fixture might be appropriate to ensure this happens

tests/integration/suite/test_global_roles.py
```diff
@@ -0,0 +1,24 @@
...
+    """Asserts admins can only edit a builtin global role's newUserDefault
+    field"""
+    admin_client = admin_mc.client
+
+    gr = admin_client.by_id_global_role(id="admin")
+    assert gr.builtin is True
+    assert "remove" not in gr.links.keys()
```
dramich: Assert on newUserDefault

tests/integration/suite/test_global_roles.py
```diff
@@ -0,0 +1,24 @@
...
+                                                   displayName="gr-test",
+                                                   description="asdf",
+                                                   rules=None,
+                                                   newUserDefault=not
+                                                   gr.newUserDefault,
+                                                   builtin=True)
+    assert new_gr.name == gr.name
+    assert new_gr.get("description") == gr.description
+    assert new_gr.rules is not None
+    assert new_gr.get("builtin") == gr.builtin is True
```
dramich: `assert new_gr.get("builtin") is True`

tests/integration/suite/test_global_roles.py
```diff
@@ -0,0 +1,24 @@
...
+
+    # newUserDefault is the only field that should editable
+    # for a builtin role
+    assert new_gr.newUserDefault == (not gr.newUserDefault)
```
dramich: `assert new_gr.newUserDefault is True`

tests/integration/suite/test_global_roles.py
```diff
@@ -0,0 +1,24 @@
...
+
+    gr = admin_client.by_id_global_role(id="admin")
+    assert gr.builtin is True
+    assert "remove" not in gr.links.keys()
+
+    new_gr = admin_client.update_by_id_global_role(id=gr.id,
+                                                   displayName="gr-test",
+                                                   description="asdf",
+                                                   rules=None,
+                                                   newUserDefault=not
```
dramich: Set this explicitly to True

pkg/api/store/globalrole/store.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func Wrap(s types.Store, grLister v3.GlobalRoleLister) types.Store {
+	return &store{
+		Store:    s,
+		grLister: grLister,
+	}
+}
+
+func (s *store) Create(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}) (map[string]interface{}, error) {
```
dramich: Create isn't needed if the only check is something norman already does.

tests/integration/suite/test_rbac.py
```diff
@@ -761,6 +761,72 @@ def try_create_ns_secret():
     )
 
 
+def test_only_admin_can_crud_global_roles(admin_mc, user_mc, remove_resource):
```
dramich: These two tests feel like they below under tests/integration/suite/test_global_roles.py.

dramich: Realizing we have similar things in here already, it looks like it's becoming the `common` or `utils` of tests

app/rolebuilder.go
```diff
@@ -54,6 +54,7 @@ func (rb *roleBuilder) addRole(displayName, name string) *roleBuilder {
 	if rb.name == "" {
 		rb.name = name
 		rb.displayName = displayName
+		rb.builtin = true
```
dramich: Lets update newRoleBuilder to default to true for builtin

app/rolebuilder.go
```diff
@@ -63,6 +64,7 @@ func (rb *roleBuilder) addRole(displayName, name string) *roleBuilder {
 	rb.next = &roleBuilder{
```
dramich: Use newRoleBuilder then set the fields on the returned role

app/rolebuilder.go
```diff
@@ -257,10 +259,12 @@ func (rb *roleBuilder) reconcileGlobalRoles(mgmt *config.ManagementContext) erro
 			return false, nil, errors.Errorf("unexpected type comparing %v and %v", have, want)
 		}
 
-		equal := haveGR.DisplayName == wantGR.DisplayName && reflect.DeepEqual(haveGR.Rules, wantGR.Rules)
+		builtin := haveGR.Builtin == wantGR.Builtin
+		equal := haveGR.DisplayName == wantGR.DisplayName && reflect.DeepEqual(haveGR.Rules, wantGR.Rules) && builtin
```
alena1108: nit: builtin can be placed as a first condition. This way if it's false, you save on other conditions' evaluation. DeepEquals in particular

pkg/api/customization/globalrole/validator.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/rancher/norman/types"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+)
+
+type Wrapper struct {
+	GlobalRoleLister v3.GlobalRoleLister
+}
+
+func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if request.Method != http.MethodPut && request.Method != http.MethodPost {
```
alena1108: can we split this line into 2 checks? As one of them repeats at line 19

app/rolebuilder.go
```diff
@@ -134,12 +135,12 @@ func (r *ruleBuilder) nonResourceURLs(u ...string) *ruleBuilder {
 	return r
 }
 
-func (r *ruleBuilder) addRoleTemplate(diplsayName, name, context string, builtin, external, hidden, administrative bool) *roleBuilder {
-	return r.rb.addRoleTemplate(diplsayName, name, context, builtin, external, hidden, administrative)
+func (r *ruleBuilder) addRoleTemplate(displayName, name, context string, builtin, external, hidden, administrative bool) *roleBuilder {
```
alena1108: builtin can be dropped from the method signature

pkg/api/customization/globalrole/validator.go
dramich: The check should just be `if request.Method != http.MethodPut`

alena1108:
LGTM

PR #23755, opened 2019-10-29
add region to ec2 cloud credential

kinarashah:
https://github.com/rancher/rancher/issues/23060

### DIFF
--- a/pkg/api/store/nodetemplate/store.go
+++ b/pkg/api/store/nodetemplate/store.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
+	"github.com/rancher/rancher/pkg/controllers/management/node"
 	"github.com/rancher/rancher/pkg/namespace"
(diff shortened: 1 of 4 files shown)


pkg/api/store/nodetemplate/store.go
```diff
@@ -10,12 +10,13 @@ import (
...
 	"github.com/rancher/norman/types/values"
+	"github.com/rancher/rancher/pkg/controllers/management/node"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/ref"
 	corev1 "github.com/rancher/types/apis/core/v1"
-	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/apis/management.cattle.io/v3"
 	managementschema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
-	client "github.com/rancher/types/client/management/v3"
+	"github.com/rancher/types/client/management/v3"
```
prachidamle: isnt there an import collision?

prachidamle:
Would be good to add integration tests to test: 
- the region is set when specified from the API
- nodetemplate can have a different region and ignores the cred's region

Issue #21674, opened 2019-07-22
Node-driver based provisioning does not respect NO_PROXY environment variable


janeczku:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create an environment that only allows internet access via an HTTP proxy and has access to a vSphere API endpoint in the local network.
2. Deploy Rancher server with the `HTTP_PROXY` and `HTTPS_PROXY` environment variables set to use the proxy and `NO_PROXY` set to exclude all private networks as documented [here](https://rancher.com/docs/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/#http-proxy).
3. Try to create a vSphere cluster according to [docs](https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/node-pools/vsphere/). Note that the vCenter IP address falls within the subnet CIDR specified in the `NO_PROXY` env variable.

**Result:**

Nodes fail to be provisioned because the machine driver attempts to connect to the vCenter via the internet proxy

**Expected Result:**

Nodes are provisioned. Machine driver respects the `NO_PROXY` CIDR values that exclude the vCenter IP from being proxied.

**Other details that may be helpful:**

1. > Rancher supports CIDR notation ranges in this list    

    See [docs](https://rancher.com/docs/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/#http-proxy)

2. Rancher docker-machine is build using [Go 1.9.x](https://github.com/rancher/machine-package/blob/master/Dockerfile.dapper#L14-L15) which doesn't support CIDR notation in the `NO_PROXY` env var (Go 1.10.x or later does).

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.5
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

bmdepesa:
rancher/rancher:v2.3-head 18fc2d261

* Created local Rancher server
  * Setup proxy vars, and NO_PROXY to exclude private network for vSphere
* Connect via VPN to access vSphere
* Created tunnel via ngrok & finish Rancher install
[...]

PR #23773, opened 2019-10-30
docker-machine version bump

luthermonson:


deniseschannon:
https://github.com/rancher/rancher/issues/21674

PR #23802, opened 2019-10-31
Change capability anno

rmweir:
**Problem:**
User cannot set capability override annotation through rancher because norman removes annotations containing cattle.io.

**Solution:**
Rename annotation to cattle.capabilities.io.

**Issue:**
https://github.com/rancher/rancher/issues/21039


dramich:
The tests need to be included in this pr

Issue #21548, opened 2019-07-16
Should not be able to assign pod security policy in project


janeczku:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug - It seems the GUI does indeed have support for disabling the PSP setting in project settings (see `local` cluster). 

**Steps to reproduce (least amount of steps as possible):**
1. Create RKE based cluster
2. Make sure `Pod Security Policy Support` is `disabled` in cluster option
3. Navigate to Project/Namespaces, click Add Project

**Result:**
User can select a pod security policy for the project. But this configuration will not take effect because the cluster was not configured with PSP support.

**Expected Result:**
When cluster is not configured to support PSP, use can not assign a PSP to a project.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.4
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

rmweir:
@jiaqiluo you will want to attempt to apply a template to a project by using the api.

jiaqiluo:
the bug is reproduced in v2.3.1 
Steps:
- run Rancher v2.3.1  single install
- add an ec2 cluster with NO psp enabled
- add a new project, notice that psp can be setted
<img width="1383" alt="Screen Shot 2019-11-04 at 10 07 43 AM" src="https://user-images.githubusercontent.com/6218999/68141549-f9e75300-feea-11e9-90fc-6fe841960897.png">
- set the psp to be restricted 
- deploy a workload in the project with setting privileged to be `Yes: container has full access to the host`
- the worload is deployed successfully, which indicates the psp is not applied actually. 
<img width="1388" alt="Screen Shot 2019-11-04 at 10 11 06 AM" src="https://user-images.githubusercontent.com/6218999/68141810-74b06e00-feeb-11e9-97c3-ba2d2998e3b2.png">

jiaqiluo:
The bug fix is validated in Rancher:v2.3-head da8cbdb71 with the default UI-index `latest-2.3`

1. When the cluster's psp is disabled, the project create/edit page does not show the option for psp anymore
[...]

PR #23815, opened 2019-11-01
Psp validator

rmweir:
**Problem:**
User can set psp template even if psp is not enabled.

**Solution:**
Return error in apply psp action is psp is disabled for the cluster

**Issue:**
https://github.com/rancher/rancher/issues/21548

### DIFF
--- a/pkg/controllers/management/cluster/clustercontroller.go
+++ b/pkg/controllers/management/cluster/clustercontroller.go
@@ -33,7 +33,7 @@ const (
 	AzureL4LB               = "Azure L4 LB"
 	NginxIngressProvider    = "Nginx"
 	DefaultNodePortRange    = "30000-32767"
-	capabilitiesAnnotation  = "capabilities.cattle.io/"
+	capabilitiesAnnotation  = "capabilities/"
 )
 
 type controller struct {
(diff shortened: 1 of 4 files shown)


pkg/controllers/management/cluster/clustercontroller.go
```diff
@@ -33,7 +33,7 @@ const (
 	AzureL4LB               = "Azure L4 LB"
 	NginxIngressProvider    = "Nginx"
 	DefaultNodePortRange    = "30000-32767"
-	capabilitiesAnnotation  = "capabilities.cattle.io/"
+	capabilitiesAnnotation  = "capabilities/"
```
leodotcloud: Was this change unintentional?

rmweir: yes, we want the user to be able to change this via rancher but currently they cannot because norman filters out annotations with cattle.io in them.

leodotcloud: I couldn't figure out the context based on the info provided :) ... You were fixing this as well I guess: https://github.com/rancher/rancher/issues/21039

rmweir: That is correct. The intention of that was to provide a way for users to override capabilities by using annotations. We decided it should be bundled in with this PR as it operates on the same area and requires that logic.

dramich:
lgtm

Issue #23743, opened 2019-10-29
[backport] OpenStack Node Driver - "Waiting for SSH to be available" when using privateKeyFile


dramich:
https://github.com/rancher/rancher/issues/23511

sangeethah:
@dan I am still able to reproduce the issue with new line character being stripped for userdata and cacert fields when testing with latest build from `v2.3-head` - `032f65c63`. 

New line character at the end of `privateKeyFile` field is not lost. But new line character is stripped from the end for `userdata` and `cacert` fields. We should be fixing this for all multi-line fields ?

dramich:
After discussing with @vincent99 we are going to add a new line to cacert and userdata as well.

sangeethah:
Able to reproduce the issue on `v2.3.1`. 
Trailing new line characters  in `privateKeyFile` , `userdata` and `cacert` field are lost when being passed as part of create cmd to driver machine.


Tested with latest build from `v2.3-head` - `75c231e51`
Trailing new line characters  in `privateKeyFile` , `userdata` and `cacert` field are NOT lost when being passed as part of create cmd to driver machine.

Note - Even when there are no Trailing new line characters  `privateKeyFile` , `userdata` and `cacert` field in the templates , the new line characters get appended to the values of these multi-line fields when being sent as part of create cmd to driver machine.

PR #23837, opened 2019-11-01
Add newline to all file content fields

dramich:
https://github.com/rancher/rancher/issues/23743


dramich:
> Should you have a test checking that \n is added if there isn't one?

It gets stripped so if there is or isn't a newline on the data it will get dropped and added back.

PR #23846, opened 2019-11-01
enhancement for testing around monitoring 

jiaqiluo:
The `action` function is used to enable cluster monitoring and project monitoring which is how the UI calls APIs for enabling monitoring.


sangeethah:
LGTM

PR #23877, opened 2019-11-04
adding vmwarevsphere validation tests

luthermonson:


tests/validation/tests/v3_api/test_vmwarevsphere_driver.py
```diff
@@ -0,0 +1,217 @@
...
+    rule = {"host": host,
+            "paths": [{"workloadIds": [workload.id], "targetPort": "80"}]}
+    p_client.create_ingress(name=name,
+                            namespaceId=ns.id,
+                            rules=[rule])
+    validate_ingress(namespace["p_client"], namespace["cluster"],
+                     [workload], host, path)
+
+
+def test_cluster_cleanup():
```
dramich: Put this into a pytest finalizers so it will run even if your tests bomb so clusters don't leak.

tests/validation/tests/v3_api/test_vmwarevsphere_driver.py
```diff
@@ -0,0 +1,217 @@
...
+        "quantity": 1,
+        "worker": True,
+    })
+
+    client.wait_success(master_pool)
+    client.wait_success(worker_pool)
+    wait_for_cluster_node_count(client, cluster, 2, 900, 5)
+
+
+def test_nodes_ready():
```
dramich: Call this from the `test_create_cluster` and `test_ingress` as well.

tests/validation/tests/v3_api/test_vmwarevsphere_driver.py
```diff
@@ -0,0 +1,217 @@
...
+        "etcd": False,
+        "hostnamePrefix": CLUSTER_NAME + "-worker",
+        "nodeTemplateId": nt.id,
+        "quantity": 1,
+        "worker": True,
+    })
+
+    client.wait_success(master_pool)
+    client.wait_success(worker_pool)
+    wait_for_cluster_node_count(client, cluster, 2, 900, 5)
```
dramich: Add some kind of asserts on the clusters to validate it's doing what is expected.

tests/validation/tests/v3_api/common.py
```diff
@@ -685,6 +694,7 @@ def wait_for_node_status(client, node, state):
     uuid = node.uuid
     start = time.time()
     nodes = client.list_node(uuid=uuid).data
+    print(node.state)
```
dramich: Drop the print

tests/validation/tests/v3_api/common.py
```diff
@@ -540,22 +540,31 @@ def wait_until_active(url, timeout=120):
...
             break
-        if client_pod is None:
-            curl_cmd = "curl " + cmd
-            result = run_command(curl_cmd)
-        else:
-            wget_cmd = "wget -qO- " + cmd
-            result = kubectl_pod_exec(client_pod, wget_cmd)
-            result = result.decode()
-        result = result.rstrip()
+        while result.find("503") != -1:
```
dramich: Instead of writing this here, let's add a helper that can be called by `validate_ingress` to do the initial check of the endpoint.

tests/validation/tests/v3_api/test_vmwarevsphere_driver.py
```diff
@@ -0,0 +1,217 @@
...
+    "vappIpallocationpolicy": "",
+    "vappIpprotocol": "",
+    "vappProperty": "",
+    "vappTransport": "",
+    "vcenter": "",
+    "vcenterPort": "443",
+}
+
+
+def test_valid_environment_variables():
```
dramich: Another option would be to something like [this](http://doc.pytest.org/en/latest/skipping.html#summary) which would stop all the tests from running anyways.

Issue #23783, opened 2019-10-30
GenerateKubeconfigActionHandler panics if LocalClusterAuthEndpoint is nil


dramich:
```
2019/10/30 23:20:47 [ERROR] Panic serving api request:
goroutine 879946 [running]:
runtime/debug.Stack(0xc037ea5000, 0x2d2e580, 0x5d5de80)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9d
github.com/rancher/rancher/vendor/github.com/rancher/norman/api.(*Server).ServeHTTP.func1(0x39e2a00, 0xc046484900)
	/go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/api/server.go:175 +0x7a
panic(0x2d2e580, 0x5d5de80)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/rancher/rancher/pkg/api/customization/cluster.ActionHandler.GenerateKubeconfigActionHandler(0x7ffac55c82f8, 0xc0003ef8c0, 0x3a64720, 0xc000acab40, 0x7ffac55c8318, 0xc0003ef8c0, 0x3a34d00, 0xc00017b380, 0xc000857600, 0x3a64ea0, ...)
	/go/src/github.com/rancher/rancher/pkg/api/customization/cluster/actions_kubeconfig.go:24 +0xbd
github.com/rancher/rancher/pkg/api/customization/cluster.ActionHandler.ClusterActionHandler(0x7ffac55c82f8, 0xc0003ef8c0, 0x3a64720, 0xc000acab40, 0x7ffac55c8318, 0xc0003ef8c0, 0x3a34d00, 0xc00017b380, 0xc000857600, 0x3a64ea0, ...)
	/go/src/github.com/rancher/rancher/pkg/api/customization/cluster/actionhandler.go:45 +0x107f
github.com/rancher/rancher/vendor/github.com/rancher/norman/api.handleAction(0xc0172ff720, 0xc09df06120, 0x0, 0x0)
	/go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/api/server.go:263 +0x62
github.com/rancher/rancher/vendor/github.com/rancher/norman/api.(*Server).handle(0xc00084d3f0, 0x39e2a00, 0xc046484900, 0xc072902800, 0xc072902800, 0xc00fe79900, 0xc08c9774a0)
[...]

PR #23912, opened 2019-11-05
Add nil check to generate kube config action

rmweir:
**Problem:**
The cluster generateKubeconfig action panics if cluster contains a nil localClusterAuthEndpoint.

**Solution:**
Nil check localClusterAuthEndpoint

**Issue:**
https://github.com/rancher/rancher/issues/23783

**Notes:**
This happens because to use access.byId the client cluster struct needs to be used, opposed to the regular management struct. Typically there is a default localClusterAuthEndpoint but all struct fields on the client cluster struct are pointers, and so the default is nil. This is a generated struct.

This only occurs when the request lacks a localClusterAuthEndpoint or it is explicitly set to null in the request. The UI is currently omitting the localClusterAuthEndpoint (see following issue):
https://github.com/rancher/rancher/issues/23913

### DIFF
--- a/pkg/api/customization/cluster/actions_kubeconfig.go
+++ b/pkg/api/customization/cluster/actions_kubeconfig.go
@@ -20,8 +20,8 @@ func (a ActionHandler) GenerateKubeconfigActionHandler(actionName string, action
 		token string
 		err   error
 	)
-
-	if cluster.LocalClusterAuthEndpoint.Enabled {
+	endpointEnabled := cluster.LocalClusterAuthEndpoint != nil && cluster.LocalClusterAuthEndpoint.Enabled
+	if endpointEnabled {
 		token, err = a.getClusterToken(cluster.ID, apiContext)
 	} else {
 		token, err = a.getToken(apiContext)
@@ -30,7 +30,7 @@ func (a ActionHandler) GenerateKubeconfigActionHandler(actionName string, action
 		return err
 	}
 
-	if cluster.LocalClusterAuthEndpoint.Enabled {
+	if endpointEnabled {
 		cfg, err = kubeconfig.ForClusterTokenBased(&cluster, apiContext.ID, apiContext.Request.Host, token)
 	} else {
 		cfg, err = kubeconfig.ForTokenBased(cluster.Name, apiContext.ID, apiContext.Request.Host, token)


tests/integration/suite/test_cluster_defaults.py
```diff
@@ -107,6 +107,17 @@ def test_import_initial_conditions(admin_mc, remove_resource):
     assert cluster.conditions is None
 
 
+def test_cluster_initial_kubeconfig(admin_mc, remove_resource):
```
dramich: This isn't going to test the fix. A default "empty" cluster gets a zero value for localClusterAuthEndpoint so it passes even without the fix.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -115,6 +115,9 @@ func ForTokenBased(clusterName, clusterID, host, token string) (string, error) {
 }
 
 func ForClusterTokenBased(cluster *managementv3.Cluster, clusterID, host, token string) (string, error) {
+	if cluster.LocalClusterAuthEndpoint == nil {
+		return "", fmt.Errorf("cluster [%s] LocalClusterAuthEndpoint cannot be nil", clusterID)
```
dramich: Let's update the error to assume nil == false and return `cluster auth endpoint disabled for cluster [%v]`. Saying the field is nil is getting into implementation details that an operator won't understand.

tests/integration/suite/test_cluster_defaults.py
rmweir: I cherry picked to master and it does test the fix. The value being omitted on create has the same effect and leads to the zero value of nil which causes the panic.

dramich:
We should drop the test as we don't have a good way to tell when cluster handles are done

pkg/kubeconfig/kubeconfig.go
```diff
@@ -115,6 +115,9 @@ func ForTokenBased(clusterName, clusterID, host, token string) (string, error) {
 }
 
 func ForClusterTokenBased(cluster *managementv3.Cluster, clusterID, host, token string) (string, error) {
+	if cluster.LocalClusterAuthEndpoint == nil {
```
kinarashah: q: do we ever reach here? we get to cluster token logic only if endpoint is enabled

rmweir: no, I added this just in case the function is used else where but I think it is unnecessary. will remove.

rmweir: no, I added this just in case the function is used else where but I think it is unnecessary. will remove.

dramich:
lgtm

PR #23927, opened 2019-11-06
v2.3 Adding vSphere Validation Tests

luthermonson:


tests/validation/tests/v3_api/common.py
```diff
@@ -1273,3 +1304,22 @@ def validate_app_deletion(client, app_id, timeout=DEFAULT_APP_DELETION_TIMEOUT):
         if len(app) == 0:
             break
 
+
+def get_admin_token(api_url=CATTLE_TEST_URL,
```
bmdepesa: First check if ADMIN_TOKEN is already set. In many cases when we run, this will be provided and the admin password will already be set, in which case this fails.

dramich:
lgtm

PR #23928, opened 2019-11-06
don't error out on missing template

kinarashah:
old constraint: each plugin template must be present for all k8s versions
this is not very convenient if a new plugin is introduced only for new
versions, changed from error to warning

https://github.com/rancher/kontainer-driver-metadata/pull/76


alena1108:
LGTM

kinarashah:
added for https://github.com/rancher/rke/issues/1050

Issue #23931, opened 2019-11-06
Invalid podSecurityPolicyTemplate  poisons Rancher's log 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- run Rancher v2.3.1
- add a cluster with `Pod Security Policy Support` enabled 
- go to the `project/namespace` page 
- add a new project
- view the project in API
- catch the request for the action `setpodsecuritypolicytemplate`

<img width="734" alt="Screen Shot 2019-11-05 at 6 30 38 PM" src="https://user-images.githubusercontent.com/6218999/68260479-64d37f80-fffa-11e9-8b8e-6d4dd89d2e06.png">

- copy the command, change the value of `podSecurityPolicyTemplateId` to an unexisting one, for example 
```
'{"podSecurityPolicyTemplateId":"doesNotExist"}'
```

- run the command from the terminal

Results:
- no error is returned in the terminal
- in Rancher's log, the following error keeps showing up

`
2019/11/06 01:25:23 [ERROR] PodSecurityPolicyTemplateProjectBindingController c-4x7mb/psptpb-6szfg [PodSecurityPolicyTemplateProjectBindingsLifecycleHandler] failed with : error creating policy: error getting pspt: podSecurityPolicyTemplate.management.cattle.io "doesNotExist" not found
`

jiaqiluo:
The bug fix is validated in v2.3.3-rc3

Steps: same as the issue
Result:
the following error returns
```
{"baseType":"error","code":"InvalidBodyContent","message":"podSecurityPolicyTemplate [restricted3] not found","status":422,"type":"error"}
```

sowmyav27:
**verified on `2.3.3-rc3`**

**Steps:**
- Run steps as mentioned here - https://github.com/rancher/rancher/issues/23931#issue-518184091
- Error in the terminal -`{"baseType":"error","code":"InvalidBodyContent","message":"podSecurityPolicyTemplate [restr] not found","status":422,"type":"error"}`
- No error is seen in the rancher logs.

PR #23932, opened 2019-11-06
add check for invalid podSecurityPolicyTemplateId

jiaqiluo:
**Problem:**
Project's action `setpodsecuritypolicytemplate` does not handle the invalid podSecurityPolicyTemplateId

**Solution:**
Return error if the podSecurityPolicyTemplate does not exist 

**Issue:**
https://github.com/rancher/rancher/issues/23931

### DIFF
--- a/pkg/api/customization/project/project_actions.go
+++ b/pkg/api/customization/project/project_actions.go
@@ -45,11 +45,12 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 }
 
 type Handler struct {
-	Projects       v3.ProjectInterface
-	ProjectLister  v3.ProjectLister
-	ClusterManager *clustermanager.Manager
-	ClusterLister  v3.ClusterLister
-	UserMgr        user.Manager
+	Projects          v3.ProjectInterface
+	ProjectLister     v3.ProjectLister
+	ClusterManager    *clustermanager.Manager
+	ClusterLister     v3.ClusterLister
+	UserMgr           user.Manager
+	PSPTemplateLister v3.PodSecurityPolicyTemplateLister
 }
 
 func (h *Handler) Actions(actionName string, action *types.Action, apiContext *types.APIContext) error {
@@ -287,6 +288,16 @@ func (h *Handler) setPodSecurityPolicyTemplate(actionName string, action *types.
 	podSecurityPolicyTemplateName, ok := providedPSP.(string)
 	if !ok && providedPSP != nil {
 		return fmt.Errorf("could not convert: %v", providedPSP)
+	}
+
+	if ok {
+		if _, err := h.PSPTemplateLister.Get("", podSecurityPolicyTemplateName); err != nil {
+			if apierrors.IsNotFound(err) {
+				return httperror.NewAPIError(httperror.InvalidBodyContent,
+					fmt.Sprintf("podSecurityPolicyTemplate [%v] not found", podSecurityPolicyTemplateName))
+			}
+			return err
+		}
 	}
 
 	schema := request.Schemas.Schema(&managementschema.Version, client.PodSecurityPolicyTemplateProjectBindingType)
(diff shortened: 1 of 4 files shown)


pkg/api/customization/project/project_actions.go
```diff
@@ -289,6 +290,15 @@ func (h *Handler) setPodSecurityPolicyTemplate(actionName string, action *types.
 		return fmt.Errorf("could not convert: %v", providedPSP)
 	}
 
+	if ok {
+		if _, err := h.PspTemplateLister.Get("", podSecurityPolicyTemplateName); err != nil {
```
cjellick: standard capitalization rule in golang is that abbreviations such as psp should be capitalized like PSP

jiaqiluo: fixed.

pkg/api/customization/project/project_actions.go
```diff
@@ -289,6 +290,15 @@ func (h *Handler) setPodSecurityPolicyTemplate(actionName string, action *types.
 		return fmt.Errorf("could not convert: %v", providedPSP)
 	}
 
+	if ok {
+		if _, err := h.PSPTemplateLister.Get("", podSecurityPolicyTemplateName); err != nil {
+			if apierrors.IsNotFound(err) {
+				return httperror.NewAPIError(httperror.InvalidBodyContent, "podSecurityPolicyTemplate not found")
```
rmweir: nit: might be good to include psp template name after psptemplate, otherwise this error is probably pretty similar to the raw one you get from the lister

rmweir: something like `fmt.Sprintf("podSecurityPolicyTemplate [%s] not found", podSecurityTemplateName)`

jiaqiluo: fixed

dramich:
An integration test should be added/updated to test this functionality

jiaqiluo:
@dan The tests are added. Could you review it?

tests/integration/suite/test_pod_security_policies.py
```diff
@@ -162,6 +162,16 @@ def check_psptpb():
         return False
 
     wait_for(check_psptpb, lambda: "PSPTB project binding not found")
+
+    # check that an invalid podSecurityPolicyTemplateId causes 422
```
dramich: The test looks good but lets move it out to it's own test. This one is already cramming a lot of functionality under the pretense of `delete`

jiaqiluo: fixed. the test is moved to its own file.

tests/integration/suite/test_pod_security_policies.py
```diff
@@ -43,6 +43,25 @@ def create_pspt(client):
     return pspt
 
 
+def set_psp_enabled(k8s_dynamic_client, value):
```
dramich: Including the `wait_for(lambda: api_client.by_id_cluster(id="local").capabilities.pspEnabled)` in here would clean up the other code since it's always done anyways.

dramich: Any test that uses this function should also have a fixture that sets the local cluster back to what it was before the tests started

tests/integration/suite/test_pod_security_policies.py
```diff
@@ -242,3 +245,70 @@ def test_pspt_binding(admin_mc, admin_pc, remove_resource):
         remove_resource(b)
     assert e.value.error.status == 422
     assert e.value.error.message == 'podSecurityPolicyTemplate not found'
+
+
+def test_project_action_set_pspt(admin_mc, admin_pc, remove_resource):
```
dramich: This test and the other one changing the psp need to be marked as non-parallel. If they run at the same time the flag could be flipped by the other test.

tests/integration/suite/test_pod_security_policies.py
dramich: Refer to the fixture `restore_rancher_version` as a good example

tests/integration/suite/test_pod_security_policies.py
```diff
@@ -43,6 +43,35 @@ def create_pspt(client):
     return pspt
 
 
+def set_psp_enabled(admin_mc, value):
```
dramich: Rename this function to drop the enabled, and move it to conftest.py so it can be integrated with a fixture for cleanup.

dramich:
lgtm

PR #23936, opened 2019-11-06
master adding vmwarevsphere driver validation test

luthermonson:


tests/validation/tests/v3_api/common.py
```diff
@@ -1338,3 +1369,23 @@ def get_user_token(user, cattle_auth_url=CATTLE_AUTH_URL):
     }, verify=False)
     print(r.json())
     return r.json()["token"]
+
+
+def get_admin_token(api_url=CATTLE_TEST_URL,
```
bmdepesa: First check if ADMIN_TOKEN is already set. In many cases when we run, this will be provided and the admin password will already be set, in which case this fails:

```
tests/v3_api/test_vmwarevsphere_driver.py:21: in <module>
    ADMIN_TOKEN = get_admin_token()
tests/v3_api/common.py:1317: in get_admin_token
    token = r.json()['token']
E   KeyError: 'token'
```

dramich:
lgtm

Issue #23930, opened 2019-11-06
[flakey test] test_pod_security_policy_template_del


rmweir:
https://drone-publish.rancher.io/rancher/rancher/1670/2/2

**Notes:**
Broke due to a modified k8s error, should retry as project is likely experience many updates in a small window of time

dramich:
https://drone-publish.rancher.io/rancher/rancher/1672/1/2

rmweir:
https://drone-pr.rancher.io/rancher/rancher/5453/2/2

PR #23952, opened 2019-11-07
Fix flaky PSP test

rmweir:
**Problem:**
If a 409 error is encountered the test will break. Encountering a 409 is expected to occur occasionally and should not lead to a failed test.

**Solution:**
Added retry logic to test_pod_security_policy_template_del

**Issue:**
https://github.com/rancher/rancher/issues/23930


prachidamle:
LGTM

Issue #23375, opened 2019-10-11
hostPort defined on workload via YAML lost after editing workload via the UI


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Spin up a default custom cluster with Rancher v2.2.8 or v2.3.0
2. Deploy the following simple nginx Deployment exposing a hostPort:
```
apiVersion: apps/v1beta2
kind: Deployment
metadata:
  labels:
    app: nginx
  name: nginx
  namespace: default
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
        ports:
        - containerPort: 80
          hostPort: 8080
          protocol: TCP
```
3. Observe nginx reachable via hostPort
4. Open Edit form for the Deployment in Rancher UI and observe port incorrectly shown as NodePort 0:
<img width="1374" alt="Screenshot 2019-10-11 at 10 48 31" src="https://user-images.githubusercontent.com/9771530/66639973-42775d80-ec18-11e9-9988-d74198a231aa.png">
5. Click upgrade in the Edit UI and observe the hostPort is removed from the Deployment.

**Environment information**
- Tested on Rancher v2.2.8 and v2.3.0

superseb:
Sounds like a duplicate of https://github.com/rancher/rancher/issues/17878

alena1108:
It's an API issue. Port created via yaml format in rancher api:

```
"ports": [
{
"containerPort": 80,
"protocol": "TCP",
"sourcePort": 0,
"type": "/v3/project/schemas/containerPort"
}
```

And here is what UI expects (example of the port configured via UI)

```
"ports": [
{
"containerPort": 80,
"dnsName": "nginx-hostport",
"kind": "HostPort",
"name": "foo",
"protocol": "TCP",
"sourcePort": 55,
"type": "/v3/project/schemas/containerPort"
}
],
```

Ideally we should figure out the mapper on the API side that would translate ports created natively, to the format UI understands.

sangeethah:
Able to reproduce the issue using rancher server version - `v2.3.1`

Tested with rancher server using laster - `v2.3-head` - `c915d7e34`

Create a workload with hostport using kubectl shell.
Open this workload in "Edit Workload" view . host port is displayed correctly.

[...]

PR #23974, opened 2019-11-07
Test for workload with host port

alena1108:
To validate https://github.com/rancher/rancher/issues/23375#issuecomment-550438582


prachidamle:
LGTM

PR #23975, opened 2019-11-07
[v2.3.3] Vsphere Legacy Normalizer

luthermonson:
Adds a normalizer which updates legacy vSphere node templates to proper vmware object paths

### DIFF
--- a/pkg/controllers/management/nodetemplate/nodetemplate.go
+++ b/pkg/controllers/management/nodetemplate/nodetemplate.go
@@ -3,6 +3,7 @@ package nodetemplate
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/rancher/rancher/pkg/controllers/management/rbac"
 	"github.com/rancher/rancher/pkg/namespace"
@@ -21,6 +22,7 @@ import (
 
 const (
 	ownerBindingsAnno = "ownerBindingsCreated"
+	vmwaredriver      = "vmwarevsphere"
 )
 
 type nodeTemplateController struct {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -221,10 +223,44 @@ func (nt *nodeTemplateController) createGlobalNodeTemplateClone(legacyName, clon
...
+/*
+	legacy vmwarevsphere nodetemplates were free form text fields
+	and this is an attempt to normalize the data to a valid vsphere path
+	e.g. "My Network" becomes "/DC_NAME/networks/My Network"
+*/
+func vsphereLegacyNormalizer(nt *unstructured.Unstructured) {
+	k := vmwaredriver + "Config"
+	c := nt.Object[k].(map[string]interface{})
+
+	dc := c["datacenter"].(string)
```
rmweir: dc, _ := will avoid panic

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -221,10 +223,44 @@ func (nt *nodeTemplateController) createGlobalNodeTemplateClone(legacyName, clon
...
+	if dc != "" && !strings.HasPrefix(dc, "/") {
+		dc = "/" + dc
+		c["datacenter"] = dc
+	}
+
+	ds := c["datastore"].(string)
```
rmweir: ds, _

pkg/controllers/management/nodetemplate/nodetemplate.go
luthermonson: ah yep.

pkg/controllers/management/nodetemplate/nodetemplate.go
luthermonson: ah yep.

PR #24006, opened 2019-11-08
Schema validation for IngressConfig fields

mrajashree:
https://github.com/rancher/rke/issues/1104
The fields ExtraEnvs, ExtraVolumes and ExtraVolumeMounts on the ingressConfig
refer the k8s native types. In types they have norman tag array[json] to prevent
unset subfields from showing up in "Edit as YAML" with their nil values.
This prevents schema validation when editing resource through UI "Edit as YAML"
and through API.
This commit does schema validation by converting the json in request to the
actual type, so invalid fields get removed.


pkg/api/store/cluster/cluster_store.go
```diff
@@ -825,3 +830,79 @@ func replaceWithOldSecretIfNotExists(oldData, newData map[string]interface{}, cl
...
+		return httperror.WrapAPIError(err, httperror.InvalidBodyContent, "ingressExtraVolumes spec conversion error")
+	}
+	for _, vol := range volumes {
+		mapForm, err := convert.EncodeToMap(vol)
+		if err != nil {
+			return err
+		}
+		mapVolumes = append(mapVolumes, mapForm)
+	}
+	values.PutValue(data, mapVolumes, "rancherKubernetesEngineConfig", "ingress", managementv3.IngressConfigFieldExtraVolumes)
```
kinarashah: we could directly put `volumes` instead of `mapVolumes` 

same for L885 and L906

alena1108:
@mrajashree can we add a test for validation part?

Issue #22578, opened 2019-08-31
Email notifier test sent from Rancher cluster, not managed Kubernetes cluster


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Set up Rancher server in a network that only has outbound SMTP access to smtp1.mycompany.com
2. Set up a Kubernetes server in a network that only has outbound SMTP access to smtp2.mycompany.com
3. Create an email notifier. Enter smtp1.mycompany.com as the host. Click Test. Note that it works.
4. Enter smtp2.mycompany.com as the host. Click Test. Note that it fails.

**Result:**
If you use smtp1.mycompany.com, the test will work, but the notifier will not because e-mails are sent out from the Kubernetes cluster, not Rancher.
If you use smtp2.mycompany.com, the test will fail, but the notifier will work.

**Other details that may be helpful:**
The test should use the Kubernetes cluster, not Rancher to test SMTP.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.2.6
- Installation option (single install/HA):
unknown

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
unknown
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
cloud (AWS), Ubuntu
- Kubernetes version (use `kubectl version`):
unknown

- Docker version (use `docker version`):
unknown

Coopyrightdmin:
Hi,
we made the same tests, but internally, smtp servers are protected against ip whitelists.
Could we imagine sending emails from Rancher instead of the cluster?

soumyalj:
Reproduced the issue with v2.3.3
Used Sendgrid and setup the email notifier in Rancher. Once the notifier test was done in rancher, rancher server IP was sent and observed in the IP access list of sendgrid. [Setup alert for a normal pod event. When an actual alert was sent, one of the cluster's node IP was sent as expected]

Tested on master-head. 
Used Sendgrid and setup the email notifier in Rancher. Once the notifier test was done in rancher, Kubernetes cluster IP(control node) was sent and observed in the IP access list of sendgrid. Rancher server IP was not sent as expected.

Tested on v2.3-head. 
[...]

Issue #22638, opened 2019-09-04
Webhook notifier test DNS failure for cluster service


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a test webhook deployment. Here's a YAML dump of the deployment and configmap I created:
```yaml
apiVersion: v1
data:
  hooks.json: |-
    [
      {
        "id": "test",
        "execute-command": "/etc/webhook/test.sh",
        "command-working-directory": "/etc/webhook",
        "include-command-output-in-response": true
      }
    ]
  test.sh: |-
    #!/bin/bash
    echo 'hello world'
kind: ConfigMap
metadata:
  annotations:
    field.cattle.io/creatorId: user-4j92f
  creationTimestamp: "2019-09-04T18:19:13Z"
  labels:
    cattle.io/creator: norman
  name: testwebhookconfig
  namespace: default
  resourceVersion: "89450"
  selfLink: /api/v1/namespaces/default/configmaps/testwebhookconfig
  uid: 7f70f43b-cf40-11e9-afb0-0aebccb4c79a
```
```yaml
apiVersion: apps/v1beta2
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "2"
    field.cattle.io/creatorId: user-4j92f
  creationTimestamp: "2019-09-04T18:20:32Z"
  generation: 2
  labels:
    cattle.io/creator: norman
    workload.user.cattle.io/workloadselector: deployment-default-testwebhooknew
  name: testwebhooknew
  namespace: default
  resourceVersion: "89718"
  selfLink: /apis/apps/v1beta2/namespaces/default/deployments/testwebhooknew
  uid: ae9b1065-cf40-11e9-afb0-0aebccb4c79a
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      workload.user.cattle.io/workloadselector: deployment-default-testwebhooknew
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 0
    type: RollingUpdate
  template:
    metadata:
      annotations:
        cattle.io/timestamp: "2019-09-04T18:21:39Z"
        field.cattle.io/ports: '[[{"containerPort":9000,"dnsName":"testwebhooknew","kind":"ClusterIP","name":"9000tcp02","protocol":"TCP","sourcePort":0}]]'
[...]

PR #24011, opened 2019-11-08
use cluster dialer

kinarashah:
https://github.com/rancher/rancher/issues/22578

### DIFF
--- a/pkg/api/customization/alert/notifier_action.go
+++ b/pkg/api/customization/alert/notifier_action.go
@@ -10,6 +10,8 @@ import (
 	"github.com/rancher/rancher/pkg/notifiers"
 	"github.com/rancher/rancher/pkg/ref"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	client "github.com/rancher/types/client/management/v3"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -42,9 +44,17 @@ func (h *Handler) testNotifier(actionName string, action *types.Action, apiConte
 		Message string
 		v3.NotifierSpec
 	}{}
+	clientNotifier := &struct {
+		client.NotifierSpec
+	}{}
+
 	if err = json.Unmarshal(data, input); err != nil {
-		return errors.Wrap(err, "unmarshaling input error")
+		return errors.Wrap(err, "unmarshalling input error")
 	}
+	if err = json.Unmarshal(data, clientNotifier); err != nil {
+		return errors.Wrap(err, "unmarshalling input error client")
+	}
+
 	notifier := &v3.Notifier{
 		Spec: input.NotifierSpec,
 	}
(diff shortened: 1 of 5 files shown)


pkg/controllers/user/pipeline/controller/pipelineexecution/pipelineexecution.go
```diff
@@ -491,7 +491,7 @@ func (l *Lifecycle) doNotify(obj *v3.PipelineExecution) (runtime.Object, error)
 			notifierMessage.Content = strings.Replace(message, "\n", "<br>\n", -1)
 		}
 		g.Go(func() error {
-			return notifiers.SendMessage(toSendRecipient.Notifier, toSendRecipient.Recipient, notifierMessage)
+			return notifiers.SendMessage(toSendRecipient.Notifier, toSendRecipient.Recipient, notifierMessage, nil)
```
gitlawr: Looks like we should use dialer here as well so that notification works in a consistent way.

pkg/api/customization/alert/notifier_action.go
```diff
@@ -42,15 +44,23 @@ func (h *Handler) testNotifier(actionName string, action *types.Action, apiConte
...
+		return errors.Wrap(err, "unmarshalling input error client")
 	}
+
 	notifier := &v3.Notifier{
 		Spec: input.NotifierSpec,
 	}
 	msg := input.Message
+	ns, id := ref.Parse(apiContext.ID)
 	if apiContext.ID != "" {
-		ns, id := ref.Parse(apiContext.ID)
```
gitlawr: These variables are only used inside the condition block, is it a better style to move this out?

pkg/notifiers/sender.go
```diff
@@ -325,6 +343,25 @@ func smtpInit(host string, port int) (*smtp.Client, error) {
 	return c, nil
 }
 
+func dialerWithTLSConfig(dialer dialer.Dialer, host, smartHost string) (dialer.Dialer, error) {
```
gitlawr: The error return value of the function looks unnecessary because it is always nil

kinarashah: oh, https://github.com/rancher/rancher/blob/37ecfd80d08309863d7d414163a1b2271cbab750/pkg/notifiers/sender.go#L354
and L359 return error?

gitlawr: Those are errors returned from `dialer.Dialer`, I mean the return value of the  `dialerWithTLSConfig` function

Tejeev:
> It is not mentioned in the original issue but I think we should do the same for webhook type notifiers. so that when a receiver is only accessible from the user cluster we don't hit a similar issue.

https://github.com/rancher/rancher/issues/22638

pkg/notifiers/sender.go
```diff
@@ -325,6 +343,25 @@ func smtpInit(host string, port int) (*smtp.Client, error) {
 	return c, nil
 }
 
+func dialerWithTLSConfig(dialer dialer.Dialer, host, smartHost string) dialer.Dialer {
+	tlsConfig := &tls.Config{
+		ServerName: host,
```
alena1108: is that the only default we have to set in tlsconfig? anything related to client cert verification? Like https://github.com/rancher/rancher/blob/master/pkg/clustermanager/manager.go#L279

PR #24014, opened 2019-11-09
Removed unused code

alena1108:


pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -241,28 +241,6 @@ type SimpleProxy struct {
 	overrideHostHeader bool
 }
 
-func NewSimpleProxy(host string, caData []byte, overrideHostHeader bool) (*SimpleProxy, error) {
```
dramich: It looks like `SimpleProxy` struct and method below can also be dropped

dramich:
lgtm

PR #24017, opened 2019-11-11
Update README.md to mention http to https port redirect conversion formula

KevinTHU:
Rancher would change external http port to https port by using conversion formula https://github.com/rancher/rancher/blob/979a592142c95ede885fcc54ceb99746bea58885/pkg/dynamiclistener/server.go#L529, example like `8080 to 8443, 30080 to 30443`, but this is not mentioned in any document, so sometimes if we werre using external http port without 80 (like 8080) and external https port 443, or some port not following the formula above, the http redirect would be confused. So, for user-friendly, adding a hint or trouble shooting in an obvious document would be better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To get automated notifications of our latest release, you can watch the announce
 
     sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 rancher/rancher
 
-Open your browser to https://localhost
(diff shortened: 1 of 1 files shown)


vincent99:
If you're not using 443 there is very little reason to bind 80 at all.  Nobody is going to seek out `http://host:[not 80]` in order to just be redirected to `https://host:[not 80  + 443 - 80]`.

KevinTHU:
> 
> 
> If you're not using 443 there is very little reason to bind 80 at all. Nobody is going to seek out `http://host:[not 80]` in order to just be redirected to `https://host:[not 80 + 443 - 80]`.

you are right, so maybe this port conversion formula is not quite necessary, or correct. the redirect is better only for 80 port and other port respone some hint like "you should us a https connect" but not a redirect? or just all http request response a hint like that?

Issue #23905, opened 2019-11-05
[CIS] Ability to specify skip tests


leodotcloud:
As a user, I should be able to specify certain tests (Level 1) to be skipped

sowmyav27:
@leodotcloud As discussed, reopening this issue, as there would be changes in this area.

leodotcloud:
The new skip value is no more an array of string. Here is a sample: {"skip": { "rke-cis-1.4": [ "1.1.1", "1.2.2"] }}

sowmyav27:
Reopening this issue.

Skip value in the API is to be given as `{"skip": { "rke-cis-1.4": [ "1.1.1", "1.2.2"] }}` 

But ideally it should just include `{ "rke-cis-1.4": [ "1.1.1", "1.2.2"] }` and not the `skip` again in its value

codyrancher:
@sowmyav27 I think that we should retain skip. This is used inside a generic config file so we should give appropriate names. Doing so makes the config easier to read and more extensible in the future.

Maybe we need to change the order so skip is nested inside the version but I don't think it should be omitted.

sowmyav27:
On master-head, currently this is how skip works from the API

1. runSecurityScan with no value passed in `skip`, the default skip values from the config.json (in the UI) are applied.
2. runSecurityScan with blank string `""` value passed in `skip`, the skip values from the config.json are ignored, and no tests are skipped.
3. runSecurityScan with skip values like `{"skip": { "rke-cis-1.4": [ "1.1.1", "1.2.2"] }}` value passed in `skip`, the tests mentioned are skipped.

@leodotcloud Question about the first scenario.

Should there be an API call to list of default skip values from the config.json or should the API rely on the config.json (from the UI)

sowmyav27:
**Verified on master head - commit id: `48a2c4ffc`** 

currently this is how skip works from the API

runSecurityScan with overrideSkip is an array of strings

- overrideSkip - [] - overrides the skips configured in the config.json
- overrideSkip - [""] - overrides the skips configured in the config.json
- overrideSkip - ["1.1.2"] - overrides the skips configured in the config.json and skips only "1.1.2" test

Issue #23904, opened 2019-11-05
[CIS] Modify the report format for new requirements


leodotcloud:
- Modify the summarizing logic to record all results, including pass, skipped, not just failures
- Change the results structure to include, fail, pass and skip section - 1

sowmyav27:
Verified on `master-head` latest.

- Report on the API summarizes the results. 
- It gives the total checks, passed/failed/mixed state.
- The details of different tests, with the status is also available in the result.
- The details on which node the test has to be run is available.
- The details about which nodes the test has failed in case the state of the test is "mixed" is available on the report

.

PR #24023, opened 2019-11-11
CIS changes for new report format and config changes

leodotcloud:
Addresses:
https://github.com/rancher/rancher/issues/23904
https://github.com/rancher/rancher/issues/23905


pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -116,43 +143,44 @@ func (csh *cisScanHandler) Updated(cs *v3.ClusterScan) (runtime.Object, error) {
...
 
 	app := &projv3.App{
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: map[string]string{creatorIDAnno: creator.Name},
-			Name:        appName,
+			Name:        appInfo.appName,
 			Namespace:   appDeployProjectID,
 		},
 		Spec: projv3.AppSpec{
 			Answers:         appAnswers,
```
luthermonson: as long as passing "skip" will do what you want later then the rest of this is good. merge types and update go mod for new hashes then it's lgtm

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -41,8 +51,30 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
...
+				appInfo.debugMaster = "true"
+			}
+			if cs.Spec.ScanConfig.CisScanConfig.DebugWorker {
+				appInfo.debugWorker = "true"
+			}
+		}
+		// Check if the configmap is populated
+		cm, err := csh.userCtxCMLister.Get(v3.DefaultNamespaceForCis, v3.ConfigMapNameForUserConfig)
+		if err != nil && !kerrors.IsNotFound(err) {
+			return cs, fmt.Errorf("error fetching cm: %v", err)
```
alena1108: can we say config map? :)

leodotcloud: `kubectl get cm` :) it's a well known acronym ... ha ha

leodotcloud: anyways, will change it :D

alena1108: on a second thought, we can just return the raw api error as it will be self explanatory

leodotcloud: Used full word and matched the format with the rest of the function.

alena1108:
LGTM

alena1108:
@leodotcloud was it built with go 1.13? I get a bunch of extra changes generated by `go mod vendor` on my machine, having 1.13.4

This pr can be merged only when vendor folder is updated with changes generated having go 1.13.4

leodotcloud:
@alena1108 https://github.com/rancher/rancher/blob/master/Dockerfile.dapper#L20, if I use newer version of go (1.13.x), then drone wouldn't be happy.

alena1108:
the pr is updated with go 1.13.4, so final LGTM

Issue #22572, opened 2019-08-30
user settings shouldn't get overridden by driver metadata


kinarashah:
Issue:
Currently metadata sync updates a couple of settings, `k8s-version`, `k8s-versions-current` and corresponding UI settings. If user updates these values, the next metadata refresh shouldn't override these values.

jiaqiluo:
here is the list of issues found during validating this issue:
- https://github.com/rancher/rancher/issues/24146
- https://github.com/rancher/rancher/issues/24152
- https://github.com/rancher/rancher/issues/24150

PR #24026, opened 2019-11-11
handle user updated metadata settings

kinarashah:
metadata settings related to k8s versions current, default and deprecated
shouldn't get overridden once user updates them

https://github.com/rancher/rancher/issues/22572

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/data.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/data.go
@@ -8,21 +8,17 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
-
-	"github.com/rancher/kontainer-driver-metadata/rke/templates"
-
-	"github.com/sirupsen/logrus"
-
-	"github.com/rancher/kontainer-driver-metadata/rke"
-
 	mVersion "github.com/mcuadros/go-version"
+	"github.com/rancher/kontainer-driver-metadata/rke"
+	"github.com/rancher/kontainer-driver-metadata/rke/templates"
+	"github.com/rancher/norman/types/convert"
+	setting2 "github.com/rancher/rancher/pkg/api/store/setting"
+	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rke/util"
-
-	"github.com/rancher/rancher/pkg/namespace"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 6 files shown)


alena1108:
@kinarashah would adding a test be possible? Also is there a way to restore to the default settings, once modified? If so, would be nice to add a test for it as well.

prachidamle:
What happens if user modifies settings and kontainer-driver-metadat gets updated to deprecate some k8s version - will that rancher setup still keep allowing the deprecated versions?

alena1108:
@prachidamle it's going to be up to user/admin to be aware of that and adjust the setting. That's an implication of using a custom setting values.

Think of it similar to UI tag setting - if user adjusts it to some custom value, and then something changes in the backend code and the custom UI no longer works. Same for default docker version and other settings.

kinarashah:
@alena1108 @prachidamle made the following changes: 

- earlier, no default value for settings was set since they were updated dynamically. Now, on every sync, we also update default value so user can see what's the current default value based on system sync. 
- if user updates value, we validate to make sure default k8s version is present in k8s-versions-current 
- for user updated values, we make sure not to override them and use them to calculate other values like `version-to-images`, `ui-default-range` etc 
- similar to other settings in rancher, when user updates value to empty, the setting would would fallback to default value. we set the label `io.cattle.user.updated` to false 
- added tests to cover these scenarios 
- refactored a bit of code to handle these changes

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -46,6 +41,23 @@ const (
 
 var existLabel = map[string]string{sendRKELabel: "false"}
 
+//settings corresponding to keys in setting2.MetadataSettings
+var userUpdateSettingMap = map[string]settings.Setting{
+	"k8s-version":             settings.KubernetesVersion,
```
prachidamle: Maybe would be good to declare these keys as const under say "rancher/pkg/settings" and reuse them across the store and kontainerdrivermetadata controller code?

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -387,97 +400,252 @@ func getWindowsName(str string) string {
...
 
-	k8sCurrVersionData, err := marshal(k8sCurrVersions)
-	if err != nil {
-		return err
+			_, err = md.Settings.Update(settingCopy)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
```
prachidamle: This code to update the default can be a separate method?

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -387,97 +400,252 @@ func getWindowsName(str string) string {
...
 	}
-	var versions []string
-	for k := range k8sCurrVersions {
-		versions = append(versions, k)
+
+	// normal setting update
+	for key, setting := range rancherUpdateSettingMap {
```
prachidamle: Maybe we should skip updating the settings updated by the user - by using the labels

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -387,97 +400,252 @@ func getWindowsName(str string) string {
...
+		if err != nil {
+			return err
+		}
+
+		if err := md.updateSettingFromFields(userUpdateSettings, false); err != nil {
+			return err
+		}
+	}
+
+	if err := md.updateSettingFromFields(updateSettings, true); err != nil {
```
prachidamle: Guess this call is to make sure defaults of all settings is updated. Do we need to update the `updateSettings` here?

pkg/controllers/management/kontainerdrivermetadata/data.go
prachidamle: @kinarashah can we do similar change for store as well?

pkg/controllers/management/kontainerdrivermetadata/data.go
prachidamle: @kinarashah does it make sense to update all default values at the start of this routine and then update the setting values? Wondering if that will be good to have defaults uptodate,  in case if we hit some error updating setting values.

kinarashah: don't think the order would matter either way, because we always read from value and not from default, but updated

prachidamle:
LGTM

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -387,97 +401,260 @@ func getWindowsName(str string) string {
...
+		return nil, err
+	}
+
+	minVersion := maxVersions[0]
+	maxVersion := util.GetTagMajorVersion(defaultK8sVersion)
+	uiSupported := fmt.Sprintf(">=%s.x <=%s.x", minVersion, maxVersion)
+	uiDefaultRange := fmt.Sprintf("<=%s.x", maxVersion)
+
+	return map[string]string{
+		"k8s-versions-current":            strings.Join(k8sVersionsCurrent, ","),
```
alena1108: shall we put all the keys in the constants somewhere in settings? it's being referenced in settings.go already I believe

Issue #23394, opened 2019-10-11
Allow user to specify that a node will never be healthy in the UI.


cloudnautique:
Some solutions create a node that is a "dummy" node that will never have kubelet or actually exist. In Rancher 2.2.8 this node would not show up in the UI. In Rancher 2.3 this node does show appear in the UI.

Since the node will never show up as healthy, or finish registering. the Rancher UI show the node as unhealthy and Red. In this situation the user will no longer be able to use the UI for noticing when something is actually broken.

As a user, I should be able to have Rancher ignore a particular node through a label.

Something like:
`cattle.rancher.io/node-status: ignore`

cloudnautique:
When the ignore label is applied to the node:
1. Rancher will show the node greyed out in UI as "Ignored".
 - there can be an action on the node object that can apply/remove the label
1. Rancher will not use this as overall node health calculation on the node dashboard.
1. This will not impact the way Rancher returns the node object through the API.
1. This will not impact kubernetes scheduling.

alena1108:
After talking offline with @cloudnautique, we are going to take the following approach:

* Don't register node created using kubectl, if it has a certain name or label. The result will be consistent with 2.2.x behavior that was reset in 2.3.x with https://github.com/rancher/rancher/pull/20375/files#diff-9ed547abad12a5cd862ad5e02f9e08b3L311 fix.
* Node name to skip will be defined in settings, and that's what we'd expect users to leverage if they want this functionality. 
* In addition to that, we'll expose similar behavior, only through specific label. If kubectl created node has `cattle.rancher.io/node-status: ignore` label, it will never be synced back by nodessyncer.

sangeethah:
Test cases to validate:

`v3/settings/ignore-node-name` related cases:

1. Set `v3/settings/ignore-node-name` to some nodename say  "deletenode".
2. Create  a new cluster.
3. Add a new node using kubectl with nodename - "deletenode". This node should not show up in ui in node list page.
4. Add a new node using kubectl with nodename - "node1". This node should  show up in ui in node list page and it will be in "Unavailable" state.
5. Add a node to a custom cluster using register command having node name override set to "deletenode" .  Should this get hidden from the UI ?
6. Set `v3/settings/ignore-node-name` to "node1". This will NOT result in "node1" (added in step4) to get removed from UI. This should NOT result in  "deletenode" ( added in step3) to show up in UI . It will continue to be hidden.

`cattle.rancher.io/node-status: ignore` label related cases:

[...]

PR #24050, opened 2019-11-12
Skip v1.node removal for objects with specific nodename

alena1108:
To resolve the removal use case https://github.com/rancher/rancher/issues/23394#issuecomment-552602152


prachidamle:
Changes are good, is it possible to add a test to add a node with the ignore label and attempt a delete?

alena1108:
@prachidamle unfortunately no. As it would require an actual node which our drone doesn't have :)

prachidamle:
@alena1108 LGTM

Issue #22707, opened 2019-09-08
Associate AD group with Rancher role


fmunteanu:
I have AD authentication enabled in Rancher 2.2.8, with 2 AD groups:
- rancher-admin
- rancher-user

I would like to automatically associate all `rancher-admin` users to Rancher `Admin` role and `rancher-user` users to Rancher `User` role. Is there a way to do this?

cloudnautique:
For any auth provider that Rancher has group support for, we want to be able to address this usecase and #22726.

tfiduccia:
Changes only to user screen:
1.  Add a button on the User list page that says Add Group.  It should only show if you are logged into an Auth provider that supports groups.

2.  Add group page:
     Drop-down with list of groups pulled from auth provider
     Global Permissions section
     Create/Cancel buttons

jiaqiluo:
The feature is tested in Rancher:master-head 5845c2d

Now admin can assign global roles to an AD group.

More tests are done and tracked in the test plan, and new issues will be made for bugs found during the validation.

PR #24058, opened 2019-11-12
Can assign groups to global roles

rmweir:
**Problem:**
Admins would like to assign AD groups to global roles so when a member is added that belongs to that group, they will inherit the global role permissions.

**Solution:**
Support assigning group principal IDs to global role bindings.

**Issue:**
https://github.com/rancher/rancher/issues/22707

### DIFF
--- a/pkg/api/customization/globalrolebinding/validator.go
+++ b/pkg/api/customization/globalrolebinding/validator.go
@@ -1,0 +1,23 @@
+package globalrolebinding
+
+import (
+	"net/http"
+
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+)
+
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if request.Method == http.MethodPut {
+		return nil
+	}
+
+	hasSingleSubject := (data["groupPrincipalId"] != nil && data["userId"] == nil) ||
+		(data["groupPrincipalId"] == nil && data["userId"] != nil)
+
+	if !hasSingleSubject {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, "must contain field [groupPrincipalId] "+
+			"OR field [userId]")
+	}
+	return nil
+}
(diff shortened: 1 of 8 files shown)


pkg/controllers/user/rbac/globalrole_handler.go
```diff
@@ -9,6 +9,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
+
+	"crypto/sha256"
```
dramich: Formatting

pkg/controllers/user/rbac/globalrole_handler.go
```diff
@@ -106,5 +104,38 @@ func grbByUserAndRole(obj interface{}) ([]string, error) {
 		return []string{}, nil
 	}
 
-	return []string{grb.UserName + "-" + grb.GlobalRoleName}, nil
+	return []string{getGRBTargetKey(grb) + "-" + grb.GlobalRoleName}, nil
+}
+
+// getGRBTargetKey returns a key that uniquely identifies the given GRB's target.
+// If a user is being targeted, then th.
```
dramich: Expand the comment

pkg/api/customization/globalrolebinding/validator.go
```diff
@@ -0,0 +1,23 @@
+package globalrolebinding
+
+import (
+	"net/http"
+
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+)
+
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
```
dramich: Should this also account for both subjects being empty since we don't allow updates to that field

rmweir: it does, it would return an error in that case, I tried to write the error so it would make sense in either scenario

dramich:
lgtm

Issue #24018, opened 2019-11-11
Ingress-After the custom domain name is accessed, the access returns 503, and the service is not automatically created.


MrYuanZhen:
* What kind of request is this
  * BUG
* Steps to reproduce
   * 1、Modify Global --> Settings --> ingress-ip-domain default xip.io domain name for other custom domain names；
   * 2、Create ingress, access returns 503,Service name is not created;

* Rancher Version
   * v2.2.8 +

guangbochen:
link PR to https://github.com/rancher/rancher/pull/24138

dewos:
@guangbochen please if possibile keep this inside the 2.3 roadmap. 

2.4 is too much distant in time. Thanks

guangbochen:
@dewos  we will backport this PR to 2.3 and this can be tracked with https://github.com/rancher/rancher/pull/24181, thanks

dewos:
@guangbochen thanks. Could you tell me the ETA for the next 2.3.* release?

dramich:
@guangbochen hold off on fixing this until user controllers are moved down to the agent level. This will allow us to fix this in a cleaner fashion and not impact performance in rancher.

PR #24067, opened 2019-11-13
Fix missing auto-generated ingress svc issue

loganhz:
**Problem:**
Ingress returns 503 due to auto-generated ingress svc is missing if we change the ingress domain to something other than xip.io

**Solution:**
The ingress controller is using the `IngressIPDomain` setting.
So the controller should run in Rancher side.

https://github.com/rancher/rancher/issues/24018


alena1108:
@loganhz I don't like the idea of moving the controller to the management side as it will have performance implications. It's exactly the reason why it was put to useronlycontext. We've had similar discussion here https://github.com/rancher/rancher/pull/22388#discussion_r321035540

The change should be - propagate the setting to user cluster. Store it in config map perhaps, and read from there instead reading it from settings directly

cc @dramich

loganhz:
Thanks! We will submit another PR

PR #24087, opened 2019-11-13
[master] docker-machine version bump

luthermonson:
https://github.com/rancher/rancher/issues/24101


luthermonson:
https://github.com/rancher/rancher/issues/24101

PR #24111, opened 2019-11-14
Rke/k-d-m vendor update

alena1108:


alena1108:
@kinarashah any idea of why windows images test fails on master, and passes on release/v2.3? The test didn't change, and the k-d-m commit is the same

kinarashah:
@alena1108 this is happening because release/v2.3 has the correct RancherVersionDev (2.4) vs master doesn't, so the loading `rkeSystemImages` has `1.13` and we get `rke-tools:v0.1.50` which isn't present in windows images (because windows only has >= 1.14)

we should update it here, 
https://github.com/rancher/rancher/blob/master/pkg/controllers/management/kontainerdrivermetadata/data.go#L38

Issue #23556, opened 2019-10-20
prometheus kube state exporter looks for  v1beta1 resources in 1.16.2


chrisbulgaria:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug likely

**Steps to reproduce (least amount of steps as possible):**
Fresh Rancher launched RKE cluster Kubernetes 1.16.1 , later upgraded to 1.16.2 , enable cluster monitoring.

**Result:**
Flooding of logs of 	exporter-kube-state-cluster-monitoring with below. No other symptoms it seems.

**Other details that may be helpful:**
020 07:53:14.845753 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.DaemonSet: the server could not find the requested resource
E1020 07:53:14.945528 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.ReplicaSet: the server could not find the requested resource
E1020 07:53:14.946056 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.Deployment: the server could not find the requested resource
E1020 07:53:15.846559 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.StatefulSet: the server could not find the requested resource
E1020 07:53:15.847481 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.DaemonSet: the server could not find the requested resource
E1020 07:53:15.946816 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.ReplicaSet: the server could not find the requested resource
E1020 07:53:15.947610 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.Deployment: the server could not find the requested resource
E1020 07:53:16.848946 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.StatefulSet: the server could not find the requested resource
E1020 07:53:16.849003 1 reflector.go:205] k8s.io/kube-state-metrics/pkg/collectors/builder.go:508: Failed to list *v1beta1.DaemonSet: the server could not find the requested resource
[...]

PR #24117, opened 2019-11-15
Mirror kube-state-metrics image

loganhz:
Problem:

There are some error logs in coreos-kube-state-metrics pod for monitoring

Solution:

Bump coreos-kube-state-metrics version to support k8s 1.16

Issue:

rancher/rancher#23556


deniseschannon:
LGTM

Issue #23368, opened 2019-10-11
Bad chart in repo causes constant refresh in UI and logs


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Create a git repo with a chart that has a bad chart name, for example (Chart.yaml):
```yaml
name: trying-a-name_with_underscores-and_dashes
version: 1.0.0
appVersion: 1.0.0
description: Blah blah
icon: 
keywords:
- testing
home: https://rancher.co/
```
Add the catalog with this chart. I used https://github.com/dnoland1/testcatalog.git master branch, global.

**Result:**
Catalog is stuck in "Refreshed" state and UI is constantly reporting it is syncing templates:
<img width="1260" alt="image" src="https://user-images.githubusercontent.com/468462/66615823-22e22580-eb83-11e9-8243-7027f8d058b3.png">

Logs show the chart error (expected), but reports templates are updated when they have not changed:
```
2019/10/10 23:51:04 [INFO] Updating global catalog testing
2019/10/10 23:51:14 [INFO] Catalog sync done. 1002 templates created, 0 templates updated, 0 templates deleted
2019/10/10 23:51:14 [ERROR] CatalogController testing [catalog] failed with : failed to update templates. Multiple error occurred: [failed to create template testing-trying-a-name_with_underscores-and_dashes: CatalogTemplate.management.cattle.io "testing-trying-a-name_with_underscores-and_dashes" is invalid: metadata.name: Invalid value: "testing-trying-a-name_with_underscores-and_dashes": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]
2019/10/10 23:51:15 [INFO] Updating global catalog testing
W1010 23:52:21.422624       6 reflector.go:270] github.com/rancher/norman/controller/generic_controller.go:175: watch of *v1.Endpoints ended with: too old resource version: 10713 (11164)
2019/10/10 23:53:08 [INFO] Catalog sync done. 1 templates created, 1001 templates updated, 0 templates deleted
[...]

PR #24127, opened 2019-11-15
Make errors not increase chart count

brendarearden:
Problem:
The count for catalog templates was increasing after create or update was called even if they threw an error. TemplateSpec value "UpgradeVersionLinks" was being set prior to DeepEquals being called during update causing the template to indicate that it needed updating when no values had changed. 

Solution:
Add bool to indicate if function completed without errors. Move "UpgradeVersionLinks" to be set only when it is needed.  

Issue: 
#23368

### DIFF
--- a/pkg/catalog/manager/catalog_sync.go
+++ b/pkg/catalog/manager/catalog_sync.go
@@ -1,12 +1,17 @@
 package manager
 
 import (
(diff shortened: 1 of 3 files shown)


brendarearden:
opening pull request with different branch

Issue #24115, opened 2019-11-14
[flakey test] test_only_admin_can_crud_global_roles


dramich:
https://drone-publish.rancher.io/rancher/rancher/1728/2/2

dramich:
https://drone-publish.rancher.io/rancher/rancher/1729/2/2

PR #24129, opened 2019-11-15
Fix flaky globalrole test

rmweir:
Problem:
Globalrole test results have been flakey. This is not due to the test. The validator is using a lister which does not always update quickly enough after object creation, leading to not found errors.

Solution:
Use client/interface instead of a lister.

Issue:
https://github.com/rancher/rancher/issues/24115

### DIFF
--- a/pkg/api/customization/globalrole/validator.go
+++ b/pkg/api/customization/globalrole/validator.go
@@ -3,8 +3,10 @@ package globalrole
 import (
 	"net/http"
 
+	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"k8s.io/apimachinery/pkg/api/errors"
 )
 
 type Wrapper struct {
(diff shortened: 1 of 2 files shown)


pkg/api/customization/globalrole/validator.go
```diff
@@ -5,18 +5,19 @@ import (
...
+	GlobalRoleClient v3.GlobalRoleInterface
 }
 
 func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
 	if request.Method != http.MethodPut {
 		return nil
 	}
 
-	gr, err := w.GlobalRoleLister.Get("", request.ID)
+	gr, err := w.GlobalRoleClient.Get(request.ID, v1.GetOptions{})
```
kinarashah: do you think it'd make sense to first check Lister and use client only if errors isNotFound?

dramich: This should go back to using the lister. The test is doing something atypical for how a grb is going to be used.

dramich:
This is a test issue. If the error returned is a 404 then it's known the lister isn't updated yet so the test can wait and try again.

dramich:
lgtm

PR #24145, opened 2019-11-18
Update types and add GRB test

rmweir:
Update types and add globalrolebinding test


tests/integration/suite/test_global_role_bindings.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def test_cannot_update_global_role(admin_mc, remove_resource):
+    """Asserts that globalRoleId field cannot be changed"""
+    admin_client = admin_mc.client
+
+    grb = admin_client.create_global_role_binding(
+        name="gr-" + random_str(),
+        userId=admin_mc.user.id,
+        globalRoleId="somerole")
```
dramich: Let's use real roles in case validation (or something else) stops using a random name

Issue #20029, opened 2019-05-07
Air-gapped cluster provisioning fails when private registry requires authentication


janeczku:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Install Rancher HA following the [HA air gapped installation docs](https://rancher.com/docs/rancher/v2.x/en/installation/air-gap-high-availability/).

2. In [step 2](https://rancher.com/docs/rancher/v2.x/en/installation/air-gap-high-availability/prepare-private-registry/), publish the images to a private registry (say, `private.registry.com`) that requires authentication to push/pull images.

3. Note how in [Step 5](https://rancher.com/docs/rancher/v2.x/en/installation/air-gap-high-availability/config-rancher-for-private-reg) one configures Rancher to `use the private registry in order to provision any Rancher launched Kubernetes clusters`.

4. Once Rancher is installed, create a vSphere node template

5. Create a cluster using the vSphere node template, and configuring the private registry from step 2 as "default system registry" while providing credentials allowing to pull the images.

**Result:**
Cluster creation fails. After nodes were provisioned the following error was logged:

```
Unable to find image 'private.registry.com/rancher/rancher-agent:v2.2.2' locally
 docker: Error response from daemon: Get https://private.registry.com/v2/:  unauthorized: authentication required
```

**Other details that may be helpful:**
While Rancher uses the credentials configured for the cluster private registry when pulling the RKE system images, it does not use the credentials when pulling the initial Rancher Agent image via the docker-machine ssh connection.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.2
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

moelsayed:
The issue is not specific to air gapped setups. This happens when:
- Rancher is configured with server level private registry(no configurable credentials).
- That private registry requires authentication. 
When rancher provisions a node, it will install docker and then try to pull the agent image from the server level private registry. Since the registry requires authentication, the pull will fail:
```
2019/08/27 17:46:10 [INFO] [node-controller-docker-machine] Installing Docker...
[...]

PR #24148, opened 2019-11-18
Private registry support for cluster/node agent

alena1108:
https://github.com/rancher/rancher/issues/20029


pkg/cluster/cluster_util.go
```diff
@@ -0,0 +1,59 @@
+package cluster
```
dramich: Nit: Can this file be renamed to privateregistry.go or similar?

dramich:
One small comment, otherwise lgtm

PR #24161, opened 2019-11-19
Move to Rancher Machine

luthermonson:
As part of a larger effort to rename `docker-machine` to `rancher-machine` and migrate the drone builds to rancher/machine from rancher/machine-package. This PR will updates both verbiage and RUNs in the Dockerfiles to get the right versions of the `rancher-machine` binaries.


dramich:
lgtm

PR #24170, opened 2019-11-20
Add timeout to exec command in node controller

takmatsu:
We sometimes faced the issue that executing command line is stuck indefinitely.
e.g) When docker-machine failed `yum update` for some reasons, node is stuck with `Provisioning centos...`
This patch adds timeout to execute command line.


StrongMonkey:
@takmatsu What if the provision process takes more than 10 minutes? Is it better to make this timeout configurable via a setting and also turn off by default? @dramich

takmatsu:
@StrongMonkey yes, we had same issues and changed it to configurable.
I think Rancher has default timeout in several controllers, It is better to have default timeout in node controller.

Issue #19724, opened 2019-04-19
Charts with long names cause catalog refreshes to time out


bentastic27:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

First setup chartmuseum or similar and set that up as a catalog in rancher. Create a package like the following with a really long name:

```
$ helm create areallylongnamelikereallyreallylongwestillneedmorez234dasdfasd
Creating areallylongnamelikereallyreallylongwestillneedmorez234dasdfasd

$ helm package areallylongnamelikereallyreallylongwestillneedmorez234dasdfasd/
Successfully packaged chart and saved it to: /Users/ben/tmp/areallylongnamelikereallyreallylongwestillneedmorez234dasdfasd-0.1.0.tgz
```

upload to chart museum:

```
$ curl -L --data-binary "@areallylongnamelikereallyreallylongwestillneedmorez234dasdfasd-0.1.0.tgz" http://chartmuseum-chartmuseum.chartmuseum.167.99.144.223.xip.io/api/charts
{"saved":true}
```

**Result:**

refreshing will timeout:

```
$ rancher catalog refresh stuff2 --wait
FATA[0061] catalog: timed out waiting for refresh
```

**Other details that may be helpful:**

This error will show up in the logs:

```
2019/04/19 14:12:03 [ERROR] CatalogController stuff2 recovered from panic "runtime error: invalid memory address or nil pointer dereference". (err=<nil>) Call stack:
/go/src/github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72
/go/src/github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:154
/usr/local/go/src/runtime/asm_amd64.s:522
/usr/local/go/src/runtime/panic.go:513
/usr/local/go/src/runtime/panic.go:82
/usr/local/go/src/runtime/signal_unix.go:390
/go/src/github.com/rancher/rancher/pkg/catalog/manager/crd.go:57
/go/src/github.com/rancher/rancher/pkg/catalog/manager/traverse.go:213
/go/src/github.com/rancher/rancher/pkg/catalog/manager/catalog_sync.go:47
/go/src/github.com/rancher/rancher/pkg/controllers/management/catalog/controller.go:86
[...]

PR #24176, opened 2019-11-20
Catalog log fix

brendarearden:
Problems:
The count for catalog templates was increasing after create or update was called even if they threw an error. TemplateSpec value "UpgradeVersionLinks" was being set prior to DeepEquals being called during update causing the template to indicate that it needed updating when no values had changed. Long label name was erring after template was being created. 

Solution:
Increment count only where there is not an error. Move "UpgradeVersionLinks" to be set only when it is needed. Add log to indicate files had a failure. Make errors indicate when the error occurred instead of being generic. Check for label length before allowing template to create. Delete template during delete if template version doesn't exist to ensure catalog delete occurs successfully. Reduce the number of times update is called on the catalog to reduce logging. UI no longer displays the individual templates that are syncing, but that the catalog is syncing. 

Issue:
#23368
#19724

### DIFF
--- a/pkg/catalog/manager/catalog_common.go
+++ b/pkg/catalog/manager/catalog_common.go
@@ -104,7 +104,7 @@ func (m *Manager) updateCatalogInfo(cmt *CatalogInfo, catalogType string, templa
 		}
 		v3.CatalogConditionRefreshed.Unknown(obj)
 		if templateName != "" {
-			v3.CatalogConditionRefreshed.Message(obj, fmt.Sprintf("syncing template %v", templateName))
+			v3.CatalogConditionRefreshed.Message(obj, fmt.Sprintf("syncing catalog %v", cmt.catalog.Name))
 		} else {
 			v3.CatalogConditionRefreshed.Message(obj, fmt.Sprintf(""))
 		}
--- a/pkg/catalog/manager/catalog_common.go
+++ b/pkg/catalog/manager/catalog_common.go
@@ -2,6 +2,7 @@ package manager
 
 import (
 	"fmt"
+	"strings"
 
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	client "github.com/rancher/types/client/management/v3"
@@ -48,6 +49,10 @@ func (m *Manager) deleteTemplates(key string, namespace string) error {
 	for _, t := range templates {
 		tvs, err := m.getTemplateVersion(t.Name, namespace)
 		if err != nil {
+			//if template version doesn't exist continue to delete template
+			if strings.Contains(err.Error(), "invalid label value") {
+				continue
+			}
 			return err
 		}
 		for k := range tvs {
--- a/pkg/catalog/manager/crd.go
+++ b/pkg/catalog/manager/crd.go
@@ -31,6 +31,10 @@ func (m *Manager) createTemplate(template v3.CatalogTemplate, catalog *v3.Catalo
 		template.Spec.Versions[i].AppReadme = ""
 	}
 	logrus.Debugf("Creating template %s", template.Name)
+	//check label length to ensure template doesn't get created without template versions
+	if len(template.Name) > 63 {
+		return errors.Errorf("%v must be no more than 63 characters", template.Name)
+	}
 	createdTemplate, err := m.templateClient.Create(&template)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create template %s", template.Name)
@@ -172,6 +176,14 @@ func (m *Manager) createTemplateVersions(catalogName string, versionsSpec []v3.T
 		templateVersion.Labels = map[string]string{
 			TemplateNameLabel: template.Name,
 		}
+		//help with garbage collection on delete
+		ownerRef := []metav1.OwnerReference{{
+			Name:       catalogName,
+			APIVersion: "v3",
+			UID:        template.UID,
+			Kind:       template.Kind,
+		}}
+		templateVersion.OwnerReferences = ownerRef
 
 		logrus.Debugf("Creating templateVersion %s", templateVersion.Name)
 		if _, err := m.templateVersionClient.Create(templateVersion); err != nil && !kerrors.IsAlreadyExists(err) {
--- a/pkg/catalog/manager/crd.go
+++ b/pkg/catalog/manager/crd.go
@@ -52,7 +52,10 @@ func (m *Manager) getTemplateMap(catalogName string, namespace string) (map[stri
 }
 
 func (m *Manager) updateTemplate(template *v3.CatalogTemplate, toUpdate v3.CatalogTemplate) error {
-	r, _ := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{template.Name})
+	r, err := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{template.Name})
+	if err != nil {
+		return err
+	}
 	templateVersions, err := m.templateVersionLister.List(template.Namespace, labels.NewSelector().Add(*r))
 	if err != nil {
 		return errors.Wrapf(err, "failed to list templateVersions")
@@ -145,7 +148,10 @@ func mergeLabels(set1, set2 map[string]string) map[string]string {
 
 func (m *Manager) getTemplateVersion(templateName string, namespace string) (map[string]struct{}, error) {
 	//because templates is a cluster resource now so we set namespace to "" when listing it.
-	r, _ := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{templateName})
+	r, err := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{templateName})
+	if err != nil {
+		return nil, err
+	}
 	templateVersions, err := m.templateVersionLister.List(namespace, labels.NewSelector().Add(*r))
 	if err != nil {
 		return nil, err
--- a/pkg/catalog/manager/crd.go
+++ b/pkg/catalog/manager/crd.go
@@ -33,7 +33,7 @@ func (m *Manager) createTemplate(template v3.CatalogTemplate, catalog *v3.Catalo
 	logrus.Debugf("Creating template %s", template.Name)
 	//check label length to ensure template doesn't get created without template versions
 	if len(template.Name) > 63 {
-		return errors.Errorf("%v must be no more than 63 characters", template.Name)
+		return errors.Errorf("failed %v must be no more than 63 characters", template.Name)
 	}
 	createdTemplate, err := m.templateClient.Create(&template)
 	if err != nil {
@@ -46,7 +46,7 @@ func (m *Manager) getTemplateMap(catalogName string, namespace string) (map[stri
 	r, _ := labels.NewRequirement(CatalogNameLabel, selection.Equals, []string{catalogName})
 	templateList, err := m.templateLister.List(namespace, labels.NewSelector().Add(*r))
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to list templates for %v", catalogName)
 	}
 	templateMap := map[string]*v3.CatalogTemplate{}
 	for _, t := range templateList {
@@ -58,7 +58,7 @@ func (m *Manager) getTemplateMap(catalogName string, namespace string) (map[stri
 func (m *Manager) updateTemplate(template *v3.CatalogTemplate, toUpdate v3.CatalogTemplate) error {
 	r, err := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{template.Name})
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to find template version with label %v for %v", TemplateNameLabel, template.Name)
 	}
 	templateVersions, err := m.templateVersionLister.List(template.Namespace, labels.NewSelector().Add(*r))
 	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/catalog/manager/traverse.go
```diff
@@ -133,8 +134,10 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 			v.VersionName = version.Name
 			v.VersionURLs = version.URLs
 
-			v.UpgradeVersionLinks = map[string]string{}
 			for _, versionSpec := range template.Spec.Versions {
+				if v.UpgradeVersionLinks == nil {
```
brendarearden: keep logic the same so get rid of nil check

brendarearden: update to make syntax

pkg/catalog/manager/traverse.go
```diff
@@ -41,7 +41,8 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 	newHelmVersionCommits := map[string]v3.VersionCommits{}
 	var errs []error
 	var terrors []error
```
brendarearden: rename terrors

pkg/catalog/manager/traverse.go
```diff
@@ -243,9 +255,20 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
+	v3.CatalogConditionUpgraded.True(catalog)
+	v3.CatalogConditionDiskCached.True(catalog)
 	cmt.catalog = catalog
 	cmt.projectCatalog = projectCatalog
 	cmt.clusterCatalog = clusterCatalog
+	if len(createErrors) > 0 {
+		if _, err := m.updateCatalogInfo(cmt, catalogType, "", false, true); err != nil {
+			return err
+		}
+		return errors.Errorf("failed to create templates. Multiple error occurred: %v", createErrors)
```
brendarearden: change errors.wrap

pkg/catalog/manager/traverse.go
```diff
@@ -258,10 +281,7 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 		commit = ""
 	}
 
-	v3.CatalogConditionUpgraded.True(catalog)
```
brendarearden: don't move that

brendarearden: In order for the templates that successfully created to not falsely show up in the logs, we have to set upgrade and discached true  even if there are errors, so it does in fact need to move

pkg/catalog/manager/crd.go
```diff
@@ -145,7 +153,10 @@ func mergeLabels(set1, set2 map[string]string) map[string]string {
 
 func (m *Manager) getTemplateVersion(templateName string, namespace string) (map[string]struct{}, error) {
 	//because templates is a cluster resource now so we set namespace to "" when listing it.
-	r, _ := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{templateName})
+	r, err := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{templateName})
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
brendarearden: determined that error being returned has information needed and wrapping is not necessary. it is returning an error string so it does not have a status and has a very specific error message

pkg/catalog/manager/crd.go
```diff
@@ -166,6 +176,14 @@ func (m *Manager) createTemplateVersions(catalogName string, versionsSpec []v3.T
 		templateVersion.Labels = map[string]string{
 			TemplateNameLabel: template.Name,
 		}
+		//help with garbage collection on delete
+		ownerRef := []metav1.OwnerReference{{
+			Name:       catalogName,
+			APIVersion: "v3",
+			UID:        template.UID,
+			Kind:       template.Kind,
```
rmweir: you would want catalog UID and Kind here

pkg/catalog/manager/traverse.go
```diff
@@ -40,8 +40,15 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 
 	newHelmVersionCommits := map[string]v3.VersionCommits{}
 	var errs []error
-	var terrors []error
-	createdTemplates, updatedTemplates, deletedTemplates := 0, 0, 0
+	var updateErrors []error
+	var createErrors []error
+	var createdTemplates, updatedTemplates, deletedTemplates, failedTemplates int
```
rmweir: nit: new line here might be good

pkg/catalog/manager/traverse.go
rmweir: this could potentially be a lot of info, might want to separate by new lines and maybe only show the actual errors in debug mode?

pkg/catalog/manager/traverse.go
```diff
@@ -243,25 +255,40 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
+	}
+	if len(updateErrors) > 0 {
+		v3.CatalogConditionRefreshed.False(catalog)
+		v3.CatalogConditionRefreshed.Message(catalog, fmt.Sprintf("Error syncing catalog %v", catalog.Name))
+		cmt.catalog = catalog
+		cmt.projectCatalog = projectCatalog
+		cmt.clusterCatalog = clusterCatalog
 		if _, err := m.updateCatalogInfo(cmt, catalogType, "", false, true); err != nil {
 			return err
 		}
```
rmweir: logic seems comparable here, may it can be put in a function to help with readability?

pkg/catalog/manager/catalog_common.go
```diff
@@ -99,10 +104,11 @@ func (m *Manager) updateCatalogInfo(cmt *CatalogInfo, catalogType string, templa
 		}
 		v3.CatalogConditionRefreshed.Unknown(obj)
 		if templateName != "" {
-			v3.CatalogConditionRefreshed.Message(obj, fmt.Sprintf("syncing template %v", templateName))
+			v3.CatalogConditionRefreshed.Message(obj, fmt.Sprintf("syncing catalog %v", cmt.catalog.Name))
 		} else {
 			v3.CatalogConditionRefreshed.Message(obj, fmt.Sprintf(""))
 		}
+
```
rmweir: ?

pkg/catalog/manager/crd.go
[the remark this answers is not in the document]
rmweir: I think this may require more context for someone who is reading logs, such as "failed to get template version [name]: improper label (or include label error)"

pkg/catalog/manager/crd.go
```diff
@@ -52,7 +56,10 @@ func (m *Manager) getTemplateMap(catalogName string, namespace string) (map[stri
 }
 
 func (m *Manager) updateTemplate(template *v3.CatalogTemplate, toUpdate v3.CatalogTemplate) error {
-	r, _ := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{template.Name})
+	r, err := labels.NewRequirement(TemplateNameLabel, selection.Equals, []string{template.Name})
+	if err != nil {
```
rmweir: same thing here with adding to the error, I don't think it'll make a lot of sense to a user, maybe something to indicate that it's probably not an issue with their catalog. Have you seen an error here at all?

pkg/catalog/manager/traverse.go
```diff
@@ -234,7 +246,7 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
 		deletedTemplates++
 	}
-	logrus.Infof("Catalog sync done. %v templates created, %v templates updated, %v templates deleted", createdTemplates, updatedTemplates, deletedTemplates)
+	logrus.Infof("Catalog sync done. %v templates created, %v templates updated, %v templates deleted, %v templates failed", createdTemplates, updatedTemplates, deletedTemplates, failedTemplates)
```
rmweir: I think including this in the log is a very good improvement. The option to see the errors may be quite useful for future debugging as well.

pkg/catalog/manager/crd.go
```diff
@@ -166,6 +176,14 @@ func (m *Manager) createTemplateVersions(catalogName string, versionsSpec []v3.T
 		templateVersion.Labels = map[string]string{
 			TemplateNameLabel: template.Name,
 		}
+		//help with garbage collection on delete
+		ownerRef := []metav1.OwnerReference{{
+			Name:       catalogName,
+			APIVersion: "v3",
+			UID:        template.UID,
+			Kind:       template.Kind,
+		}}
```
rmweir: I think you either want the name to be the template name or the UID/Kind to be the catalog's UID/Kind, depending on which you want to be the owner.

rmweir:
what are your thoughts on tests for this?

pkg/catalog/manager/catalog_common.go
brendarearden: I had to move the method that was calling catalog update outside the template loop, so it no longer updates for each template.  This means the sync message needs to be at a catalog level and not for each template.

pkg/catalog/manager/crd.go
brendarearden: if a template's template version failed to create, but the template successfully created this causes  r to be nill during update because templatenamelabel doesn't exist on that template. Then the next call `m.templateVersionLister.List(template.Namespace, labels.NewSelector().Add(*r))` was being called with a nil r and panicking.

pkg/catalog/manager/traverse.go
brendarearden: added function as well as unit test for new function

pkg/catalog/manager/traverse.go
brendarearden: The error information needs to be returned to the user instead of only during debug. Errors are separated by create, update and delete errors instead of new lines. Ensuring the backoff is happening we are reducing the space that logs will take up.

pkg/catalog/manager/crd.go
```diff
@@ -166,6 +176,14 @@ func (m *Manager) createTemplateVersions(catalogName string, versionsSpec []v3.T
 		templateVersion.Labels = map[string]string{
 			TemplateNameLabel: template.Name,
 		}
+		//help with garbage collection on delete
+		ownerRef := []metav1.OwnerReference{{
+			Name:       template.Name,
+			APIVersion: "v3",
```
dramich: This can come off the template as well.

pkg/catalog/manager/traverse.go
```diff
@@ -133,8 +141,8 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 			v.VersionName = version.Name
 			v.VersionURLs = version.URLs
 
-			v.UpgradeVersionLinks = map[string]string{}
 			for _, versionSpec := range template.Spec.Versions {
+				v.UpgradeVersionLinks = map[string]string{}
```
rmweir: don't you only want to do this map is null? otherwise it's being rewritten

pkg/catalog/manager/traverse_test.go
```diff
@@ -0,0 +1,62 @@
+package manager
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
rmweir: keep non standard imports together

pkg/catalog/manager/traverse.go
brendarearden: I decided to keep the original logic, just move it into the method that actually sets that value. Since we are building the value,  v.UpgradeVersionLinks will always be null until we set it so doing a null check was not needed.

brendarearden: wait, yes, you are correct - I will add the null check back in!

tests/integration/suite/test_catalog.py
```diff
@@ -283,3 +283,34 @@ def test_catalog_refresh(admin_mc):
     # It just needs to be more than none, other test can add/remove catalogs
     # so a hard count will break
     assert len(out['catalogs']) > 0, 'no catalogs in response'
+
+
+def test_invalid_catalog_charts(admin_mc, remove_resource):
+    """Test chart with too long name and chart with invalid
+     name in catalog error properly"""
+    client = admin_mc.client
+    name = "test"
```
dramich: Use the random string whenever creating resources so things don't collide

pkg/catalog/manager/traverse_test.go
```diff
@@ -0,0 +1,60 @@
...
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var cmt CatalogInfo
+var catalog v3.Catalog
+var projectCatalog v3.ProjectCatalog
+var clusterCatalog v3.ClusterCatalog
+
+func setupTest(t *testing.T) {
+	t.Log("Setting up test...")
```
dramich: Drop the t.log unless it's really needed

pkg/catalog/manager/traverse_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	clusterCatalog = v3.ClusterCatalog{
+		Namespaced:  types.Namespaced{},
+		Catalog:     v3.Catalog{},
+		ClusterName: "",
+	}
+}
+
+func Test_Set_Catalog_Error_State(t *testing.T) {
```
dramich: Add some more test cases, happy path and sad path, to cover more of the code.

dramich:
LGTM

Issue #24036, opened 2019-11-11
Can create PRTB and CRTB with group and user targets


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. create PRTB or CRTB with any combinations of groupId, groupPrincipalId, userId, userPrincipalId where at least two of them are not empty/null.

**Result:**
API returns successful, despite this being invalid and causing lifecycle to fail.

**Other details that may be helpful:**
Should be a validator that only allows one of the mentioned inputs.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
master 11/11/19

jiaqiluo:
The bug fix is validated in rancher:master-head dad58367e

Run Rancher container and enable the Github Auth


**Do the following validations on  `v3/clusterRoleTemplateBindings`**
Validation 1: create a clusterRoleTemplateBindings with both userId and userPrincipalId
```
{
"clusterId": "c-rchvh",
"groupId": "",
"groupPrincipalId": "",
"name": "",
"namespaceId": "",
"roleTemplateId": "clusterroletemplatebindings-manage",
"userId": "u-z9lhp",
"userPrincipalId": "github_team://1395580"
}
```
Result: 
```
{
"baseType": "error",
"code": "InvalidBodyContent",
"message": "must target a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]",
"status": 422,
"type": "error"
}
```

Validation 2: create a clusterRoleTemplateBindings with both userId and groupId
```
{
"clusterId": "c-rchvh",
"groupId": "",
"groupPrincipalId": "github_team://1395580",
"name": "",
"namespaceId": "",
"roleTemplateId": "clusterroletemplatebindings-manage",
"userId": "u-z9lhp",
"userPrincipalId": ""
}
```
Result: 
```
{
"baseType": "error",
"code": "InvalidBodyContent",
"message": "must target a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]",
"status": 422,
"type": "error"
}
```
Validation 3: create a clusterRoleTemplateBindings without any id 
```
{
"clusterId": "c-rchvh",
"name": "",
"namespaceId": "",
"roleTemplateId": "cluster-owner"
}
```
Result: 
```
{
"baseType": "error",
"code": "InvalidBodyContent",
[...]

PR #24179, opened 2019-11-21
Validate CRTB and PRTB targets

rmweir:
**Dependency:**
types (merge first): https://github.com/rancher/types/pull/1042

**Problem:**
CRTBs and PRTBs can currently target both, a user and a group, or neither. Those are invalid and will cause handlers to continuously fail.

**Solution:**
Add a validator to ensure CRTBs and PRTBs are target a user or (exclusive) a group.

**Issue:**
https://github.com/rancher/rancher/issues/24036

### DIFF
--- a/tests/integration/suite/test_role_template.py
+++ b/tests/integration/suite/test_role_template.py
@@ -55,68 +55,44 @@ def test_edit_builtin_role_template(admin_mc, remove_resource):
 
 def test_context_prtb(admin_mc, admin_pc, remove_resource,
                       user_mc):
-    """Tests that any user cannot edit a cluster role binding form the project
-    context
+    """Asserts that a projectroletemplatebinding cannot reference a cluster
+    roletemplate
     """
+    admin_client = admin_mc.client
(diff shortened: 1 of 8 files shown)


rmweir:
After some investigation it seems that changing the role ref of a PRTB or CRTB properly updates downstream cluster's CRBs. Changing the subject causes new CRBs to be created in the downstream cluster but does not delete the old ones. Therefore, we are blocking updates on subjects and allowing them for roles.

tests/integration/suite/test_role_template.py
```diff
@@ -55,68 +55,44 @@ def test_edit_builtin_role_template(admin_mc, remove_resource):
 
 def test_context_prtb(admin_mc, admin_pc, remove_resource,
                       user_mc):
-    """Tests that any user cannot edit a cluster role binding form the project
-    context
+    """Asserts that a projectroletemplatebinding cannot reference a project
```
dramich: `project` > `cluster`

PR #24191, opened 2019-11-21
Add GRB cleanup for deleted GR

rmweir:
**Problem:**
If a globalrole is deleted, globalrolebindings referencing it are left over.

**Solution:**
Add globalrole being referenced in GlobalRoleName field as an owner reference for globalrolebindings.

**Issue:**
https://github.com/rancher/rancher/issues/16216

### DIFF
--- a/pkg/api/store/globalrolebindings/globalrolebinding_store.go
+++ b/pkg/api/store/globalrolebindings/globalrolebinding_store.go
@@ -1,28 +1,60 @@
 package grbstore
 
 import (
+	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/values"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const GrbVersion = "field.cattle.io/grbUpgrade"
 
-func Wrap(store types.Store, grbLister v3.GlobalRoleBindingLister) types.Store {
+func Wrap(store types.Store, grLister v3.GlobalRoleLister) types.Store {
 	return &grbStore{
-		Store:     store,
-		grbLister: grbLister,
+		Store:    store,
+		grLister: grLister,
 	}
 }
 
 type grbStore struct {
 	types.Store
-	grbLister v3.GlobalRoleBindingLister
+
+	grLister v3.GlobalRoleLister
 }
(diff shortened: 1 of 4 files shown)


dramich:
Move this to a store to ensure the owner reference is applied

pkg/api/store/globalrolebindings/globalrolebinding_store.go
```diff
@@ -8,21 +8,48 @@ import (
...
+
+	grLister v3.GlobalRoleLister
 }
 
 func (s *grbStore) Create(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}) (map[string]interface{}, error) {
 
 	values.PutValue(data, "true", "metadata", "annotations", GrbVersion)
 
+	err := s.addOwnerReference(data)
+	if err != nil {
```
dramich: The error should be checked for not exists and an error code returned accordingly

pkg/api/store/globalrolebindings/globalrolebinding_store.go
```diff
@@ -8,21 +8,48 @@ import (
...
+		return err
+	}
+
+	ownerReference := map[string]interface{}{
+		"apiVersion": globalRole.APIVersion,
+		"kind":       globalRole.Kind,
+		"name":       globalRole.Name,
+		"uid":        globalRole.UID,
+	}
+	values.PutValue(data, []map[string]interface{}{ownerReference}, "ownerReferences")
```
dramich: `ownerReferences` is not a top level attribute, this needs to be under metadata and it's a list of objects.

dramich:
Some tests/test changes should be included in the PR.

pkg/api/store/globalrolebindings/globalrolebinding_store.go
```diff
@@ -8,21 +8,48 @@ import (
...
-		grbLister: grbLister,
+		Store:    store,
+		grLister: grLister,
 	}
 }
 
 type grbStore struct {
 	types.Store
-	grbLister v3.GlobalRoleBindingLister
+
```
dramich: Extra line here

pkg/api/store/globalrolebindings/globalrolebinding_store.go
dramich: After some discovery it appears things in the store should follow the view of being a client object - meaning everything at the top level. There are issues with doing metadata at both levels which causes things to be dropped once the obj reaches the toInternal stage.

dramich:
lgtm

PR #24208, opened 2019-11-22
CIS Scan custom cluster provisioning

sowmyav27:
Test to bring up a custom cluster with a CIS complaint cluster.yml (Can be modified when required)
No validation tests.
End of cluster provisioning, called this function - _validate_cluster_state_


sangeethah:
LGTM

Issue #22006, opened 2019-08-06
Provisioning of GKE-Cluster not working (configured proxy)


ChrisHaPunkt:
**What kind of request is this (question/bug/enhancement/feature request):**
BUG

**Steps to reproduce (least amount of steps as possible):**
Don't know if this is related to our proxy

1. Configure a http-proxy in Rancher Helm-Chart 
a. proxy: http://techproxy.<domain\>.local:80
b. noProxy: 127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
2. Generate a valid ServiceAccount in the GCP
3. Create GKE Cluster through UI (we disabled stackdriver monitoring and logging)

**Result:**
- "Provisioning cluster c-hb67b....."
- Cluster is up in GKE
![grafik](https://user-images.githubusercontent.com/4389395/62524083-51ffb200-b835-11e9-98d9-0d34e74d8330.png)
- "Waiting for API to be available"
- No cluster and node agent ressources are deployed in the GKE Cluster
- "Waiting for API to be available" infinitely
Is it prossible that the proxy is not applied to some sub-container during cluster creation or sth?
**Other details that may be helpful:**

- No Rancher cluster and node agent ressources are deployed to the GKE Cluster
![grafik](https://user-images.githubusercontent.com/4389395/62525283-6e9ce980-b837-11e9-8b74-dee7feb63abd.png)

- The endpoint url (from the logs below) can be curled and accessed from within the rancher pod. The proxy vars are set there correctly.
- The GKE Cluster is being deleted if I trigger the deletion in Rancher 

**Rancher logs flooding with the following**
> 19/08/06 08:31:42 [INFO] Stopping cluster agent for c-hb67b
> 2019/08/06 08:31:42 [ERROR] ClusterController c-z8rqx [user-controllers-controller] failed with : failed to start user controllers for cluster c-hb67b: failed to contact server: Get https://<gke-endpoint\>/api/v1/namespaces/kube-system?timeout=30s: proxyconnect tcp: waiting for cluster agent to connect
> 2019/08/06 08:31:43 [INFO] Stopping cluster agent for c-hb67b
> 2019/08/06 08:31:43 [ERROR] ClusterController c-vnrwk [user-controllers-controller] failed with : failed to start user controllers for cluster c-hb67b: failed to contact server: Get https://<gke-endpoint\>/api/v1/namespaces/kube-system?timeout=30s: proxyconnect tcp: waiting for cluster agent to connect
[...]

Issue #23293, opened 2019-10-09
Rancher cannot add EKS cluster in proxied environment


nickngch:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. have a proxied environment where outbound communication passes through a proxy
2. have proxy use a private CA
3. Try to add EKS cluster on Rancher UI

**Result:**
Error appeared on UI
`Error creating config map: Post https://xxx.yyy.eks.amazonaws.com/api/v1/namespaces/kube-system/configmaps: x509: certificate signed by unknown authority
`

**Other details that may be helpful:**
Tried:
1. import EKS CA into Rancher and Rancher agent pods.
2. import EKS CA cert into proxy VM
3. import ca-additional.pem into Rancher agent pods

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Hosted

gz#6166

soumyalj:
Reproduced the issue in Rancher v2.2.8. 
1. Created Proxy server without Certs 
2. Created rancher server behind the proxy server. 
3. Created EKS cluster . Cluster creation failed with logs as below:
```
2019/12/17 21:04:33 [ERROR] ClusterController c-mzfvh [user-controllers-controller] failed with : failed to start user controllers for cluster c-mzfvh: failed to contact server: Get https://F15B9900D2BA0662566B3E7B636730E3.yl4.us-west-2.eks.amazonaws.com/api/v1/namespaces/kube-system?timeout=30s: proxyconnect tcp: waiting for cluster agent to connect
2019/12/17 21:04:33 [INFO] Stopping cluster agent for c-mzfvh
2019/12/17 21:04:33 [ERROR] ClusterController c-mzfvh [user-controllers-controller] failed with : failed to start user controllers for cluster c-mzfvh: failed to contact server: Get https://F15B9900D2BA0662566B3E7B636730E3.yl4.us-west-2.eks.amazonaws.com/api/v1/namespaces/kube-system?timeout=30s: proxyconnect tcp: waiting for cluster agent to connect
2019/12/17 21:04:38 [INFO] kontainerdriver amazonelasticcontainerservice listening on address 127.0.0.1:10467
[...]

PR #24211, opened 2019-11-22
Hosted cluster dialer fix to work with proxy

alena1108:
https://github.com/rancher/rancher/issues/22006

In proxy setup, GKE cluster will fail registration with Rancher, at the user cluster controller registration point. 

The reason for the user controller registration failure is in the way we build the dialer to the hosted k8s clusters. All the requests should go directly to hosted k8s api endpoint, but this condition will never be true if proxy environment is set:

https://github.com/rancher/rancher/blob/master/pkg/dialer/factory.go#L139
address will be a proxy ip address, and hostPort is the k8s api endpoints, and they'll never match

### DIFF
--- a/pkg/dialer/factory.go
+++ b/pkg/dialer/factory.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/dialer/factory.go
```diff
@@ -136,7 +136,7 @@ func (f *Factory) clusterDialer(clusterName, address string) (dialer.Dialer, err
 
 	hostPort := hostPort(cluster)
 	logrus.Debugf("dialerFactory: apiEndpoint hostPort for cluster [%s] is [%s]", clusterName, hostPort)
-	if address == hostPort && isCloudDriver(cluster) {
+	if isCloudDriver(cluster) {
```
alena1108: @ibuildthecloud Do you have any security concerns
or any other concerns about hostPort/address check drop? Not sure what requests **do not go** to the apiEndpoint ip address in the hosted cluster, and you want them to go through the cluster agent

ibuildthecloud:
LGTM

deniseschannon:
https://github.com/rancher/rancher/issues/23293

Issue #24223, opened 2019-11-24
Automation - Implement RBAC cases for Workloads


jiaqiluo:
Task 1: Implement the infrastructure for implementing RBAC testing
Task 2: Add RBAC related testing cases for workloads:
- cluster member can create projects, deploy, edit and delete workloads in them, but cannot access to other projects that he has no access to in the same cluster 
- project member can deploy, edit and delete workloads in the project he has access to, but cannot access to other projects that he has no access to in the same cluster 
- project owner can deploy, edit and delete workloads in the project he has access to, but cannot access to other projects that he has no access to in the same cluster 
- Cluster owner can create new projects, and deploy, edit, delete workloads in any project in the cluster 
- Project Read-only user can not deploy new workload, edit or delete existing workloads in the project

sangeethah:
@jiaqiluo Could you also include RBAC cases for Cluster owner and project_readonly roles?

jiaqiluo:
@sangeethah  Sure, the issue is updated with new test cases.

sangeethah:
Project Read-only user can not deploy new workload, edit or delete existing workloads in the project -> Can you also add the check for this user to be able to view workloads in the project he has access to , but not in other projects.

PR #24224, opened 2019-11-24
Automation - RBAC for Workloads

jiaqiluo:
https://github.com/rancher/rancher/issues/24223

This introduces a new global variable `RANCHER_TEST_RBAC`, and RBAC tests will be executed when `RANCHER_TEST_RBAC` is set to be `True`


sangeethah:
LGTM

PR #24237, opened 2019-11-25
Don't loop n^2 times on stopping clusters that aren't connected

ibuildthecloud:

Issue #24204, opened 2019-11-22
Bug: Rancher pod memory leak, OOM


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Unknown at this time

**Result:**
Rancher pods getting killed OOM

**Other details that may be helpful:**
Might be related to https://github.com/rancher/rancher/issues/21361 but we scaled down all rancher node and cluster agents to 0 on all clusters except for the local cluster.

Ping me for logs, heap, thread dump and etcd snapshot.  

The only way we were able to fix this was by restoring to an etcd snapshot from before the problem started happening.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

deniseschannon:
We've put some fixes in around memory leaks into 2.3.3 but have not definitively concluded that this was the root cause for this issue. We want to keep this open for now to monitor it.

deniseschannon:
We will keep this around to track to see if someone else hits the OOM with the later releases.

harshsharma22:
> We will keep this around to track to see if someone else hits the OOM with the later releases.

facing same issue in v2.3.4

ukinau:
JFYI: as I left comment in #21361. 
there is memory leak problem when you have the clusters which is not responding via rancher cluster-agent. 

> we scaled down all rancher node and cluster agents to 0
This looks like pretty much causing the issue what I described in https://github.com/rancher/rancher/issues/21361#issuecomment-612500020
https://medium.com/uckey/memory-goroutine-leak-with-rancher-kubernetes-custom-controller-with-client-go-9e296c815209

maggieliu:
Closing this in favor of https://github.com/rancher/rancher/issues/29429.

PR #24245, opened 2019-11-26
Don't loop n^2 times on stopping clusters that aren't connected

cjellick:


cjellick:
New and improved https://github.com/rancher/rancher/pull/24237
This time with less linting errors!

deniseschannon:
https://github.com/rancher/rancher/issues/24204

PR #24302, opened 2019-11-27
fix the bug that Jenkins job fails because no kubeconfig file exists

jiaqiluo:
fix the bug that Jenkins job fails because no kubeconfig file exists


sowmyav27:
LGTM

Issue #15266, opened 2018-08-28
Index listing for /assets and /translations on rancher web server


sorinjulean:
*Rancher Webserver allows index listing for /assets /translations and subdirs*

Can it be disabled ?

Many thanks !

```
> curl https://<rancher-server-url>/translations/
<pre>
<a href="de-de.json">de-de.json</a>
<a href="en-us.json">en-us.json</a>
<a href="es-es.json">es-es.json</a>
<a href="fa-ir.json">fa-ir.json</a>
<a href="fr-fr.json">fr-fr.json</a>
<a href="hu-hu.json">hu-hu.json</a>
<a href="it-it.json">it-it.json</a>
<a href="ja-jp.json">ja-jp.json</a>
<a href="ko-kr.json">ko-kr.json</a>
<a href="nb-no.json">nb-no.json</a>
<a href="nl-nl.json">nl-nl.json</a>
<a href="none.json">none.json</a>
<a href="pt-br.json">pt-br.json</a>
<a href="ru-ru.json">ru-ru.json</a>
<a href="sh-hr.json">sh-hr.json</a>
<a href="sv-se.json">sv-se.json</a>
<a href="uk-ua.json">uk-ua.json</a>
<a href="vi-vn.json">vi-vn.json</a>
<a href="zh-hans.json">zh-hans.json</a>
<a href="zh-hant.json">zh-hant.json</a>
</pre>

```

```
> curl https://<rancher-server-url>/assets/
<pre>
<a href="fonts/">fonts/</a>
<a href="images/">images/</a>
<a href="intl/">intl/</a>
<a href="node-asset-manifest.js">node-asset-manifest.js</a>
<a href="ui-898547658eaad28cff27a5b72d3a11f7.js">ui-898547658eaad28cff27a5b72d3a11f7.js</a>
<a href="ui-8e94cbc7bf09ca375540e969018241ec.map">ui-8e94cbc7bf09ca375540e969018241ec.map</a>
<a href="ui-dark.css">ui-dark.css</a>
<a href="ui-dark.rtl.css">ui-dark.rtl.css</a>
<a href="ui-light.css">ui-light.css</a>
<a href="ui-light.rtl.css">ui-light.rtl.css</a>
<a href="ui.css">ui.css</a>
<a href="ui.rtl.css">ui.rtl.css</a>
[...]

PR #24330, opened 2019-12-02
[2.4] Disable index list

GGGitBoy:
Problem:

Rancher Webserver allows index listing for /assets /translations and subdirs

Solution:

Disable index list

Related Issue:

https://github.com/rancher/rancher/issues/15266

### DIFF
--- a/server/ui/ui.go
+++ b/server/ui/ui.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/rancher/norman/parse"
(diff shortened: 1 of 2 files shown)


vincent99:
This doesn't hurt anything (assuming it works), but also doesn't really accomplish anything.  The UI source and everything in the directory is public (and open-source).  There are no secret files to discover with directory listings.

loganhz:
Yes, it's safe as they are all public files. With this PR, users don't have to explain why directory listing is enabled to their security team. So I think we should disable it. :)

Issue #24347, opened 2019-12-02
Bug: all catalogs get x509 errors when using s3 snapshots on Rancher v2.3.2


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1) Setup HA Rancher installation v2.3.2 using your own set of private signed certificates.
2) create downstream cluster
3) setup s3 snapshots, use shorter snapshot periods and retention to reproduce faster

**Result:**
Observe when snapshots finally rotate through and delete the oldest snapshot that the catalog's in Rancher now have x509 errors.

**Other details that may be helpful:**
If you update the Rancher deployment it will delete the Rancher pods and fix the problem temporarily until the next time snapshots are cycled through.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.3.2
- Installation option (single install/HA): HA

sowmyav27:
**Reproduced the issue on `2.3.3`**

**Steps:**

1. Bring up a minio with self signed certs
2. On a rancher server with self-signed certificate that it generates, create an AWS cluster.
3. Enter the minio details for the S3 config in the cluster, and enter CA cert generated in Step 1 in field Custom CA Certificate 
4. When the cluster comes up Active, after manual delete of snapshots or the automatic deletion of snapshot, on ADDITION of a new catalog like "https://charts.rook.io/beta, branch:master" gives the error. On refresh of existing catalog, NO error is seen.
5. The issue is also seen when the S3 settings are modified, i.e. change the bucket name. On Listing Snapshots, NO issue was seen. 
6. Stop and start of Rancher server.
7. The catalogs recover and are seen Active. And Apps are available from these catalogs.


**Verified on `v2.3-head` commit id: `430dd0bb3`**

**Steps:**
1. Bring up a minio with self signed certs
2. On a rancher server with self-signed certificate that it generates, create an AWS cluster.
3. Enter the minio details for the S3 config in the cluster, and enter CA cert generated in Step 1 in field Custom CA Certificate 
4. When the cluster comes up Active, after manual delete of snapshots or the automatic deletion of snapshot, on ADDITION of a new catalog like "https://charts.rook.io/beta, branch:master" NO error is seen.
5. Modify S3 settings, i.e. change the bucket name in the cluster.
6. No error is seen on ADDITION of a new catalog like "https://charts.rook.io/release, branch:master" NO error is seen.

[...]

Issue #23231, opened 2019-10-04
Certificate error when attempting to Add a Catalog in v2.3.0-rc9


mattmattox:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Deploy Rancher v2.3.0-rc9

**Result:**
x509 certificate signed by unknown authority error for all catalogs minus git.rancher.io.
![image](https://user-images.githubusercontent.com/17145893/66229550-b315fb80-e6a7-11e9-9fe6-7b5de600403c.png)

**Other details that may be helpful:**
If you curl the catalog from inside Rancher, it works fine with not cert error.
Also, this environment is using an proxy. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.3.0-rc9
- Installation option (single install/HA): HA

dsargent3220:
I can confirm that this is still having issues in rc11.  I upgraded the cluster and Kubernetes.
I also removed the helm catalog and readded it, same errors as above.

dsargent3220:
This may be because we are behind a proxy, however, we have the same basic setup as our previous Rancher instances so something must be new that is causing this behavior.

dsargent3220:
This appears to be fixed in 2.3.1-rc4

dsargent3220:
OK, this is really frustrating.  I'm finishing up testing on our new cluster configs and deleted the current test cluster and readded it and now I'm getting the cert errors again.  This is on 2.3.1

bmdepesa:
I am unable to reproduce so far in v2.3.1

* HA install (fresh install v2.3.1)
* Rancher-generated self-signed certs
* Rancher behind proxy (Squid proxy)
* Create a cluster (1.15.5)
* Enable global Helm / Helm Incubator catalogs
  * Sync succeeded
* Add a few custom catalogs to global
  * Sync succeeded
* Add some project catalogs
  * Sync succeeded

I did not see any x509 errors at any point. 

@dsargent3220 are there any other details you can provide on your setup that might help in reproducing?

dsargent3220:
Both of the clusters I see this on were upgrades from 2.2.8.  Beyond that?  Not that I can think of
I'm free until 4 pm pacific if you want to do a Zoom.

bmdepesa:
Can you provide your proxy config (with sensitive info removed)?  

I will try in an upgraded setup as well on the next pass.

[...]

PR #24348, opened 2019-12-03
Etcd backup to use its own http transport

alena1108:
https://github.com/rancher/rancher/issues/24347
https://github.com/rancher/rancher/issues/23231

### DIFF
--- a/pkg/controllers/management/etcdbackup/etcdbackup.go
+++ b/pkg/controllers/management/etcdbackup/etcdbackup.go
@@ -31,7 +31,6 @@ const (
 	clusterBackupCheckInterval = 5 * time.Minute
 	compressedExtension        = "zip"
 	s3Endpoint                 = "s3.amazonaws.com"
-	defaultTransportTimeout    = 30
 )
 
 type Controller struct {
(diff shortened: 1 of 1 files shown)


dramich:
LGTM

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -322,7 +322,8 @@ func GetS3Client(sbc *v3.S3BackupConfig, timeout int) (*minio.Client, error) {
 	}
 	var s3Client = &minio.Client{}
 	var creds *credentials.Credentials
-	var tr = http.DefaultTransport
+	var tr http.RoundTripper = &http.Transport{}
+	copier.Copy(tr, http.DefaultTransport)
```
cjellick: Presumably this would copy pointers over such that the old transport and the new transport would point to the same underlying objects. It looks like that means this `tr` would share DefaultTransport's DialContext. Could that cause a future problem if we then modify the DialContext on `tr`?

alena1108: @cjellick we do modify it actually on line 327. Just to be safe, I'll better initialize the struct instead of copying it.

alena1108:
@cjellick can you review https://github.com/rancher/rancher/pull/24348/commits/d9375d6324fbe5c19906c5ab25186a2e8fbc91fd. Once approved, I'll remove the vendor commit (as it's no longer needed) before merging the pr

cjellick:
change looks good. one other concern: do you think there's any chance this could either leak or starve connetion resources?
Like, we are creating a new transport per handler execution. Are you relatively confident the underlying connections get cleaned up?

This area confuses me a bit, so I cannot answer that question definitely myself

alena1108:
@cjellick good point. Will see what can be done here

Issue #23983, opened 2019-11-08
[forwardport] Allow user to specify that a node will never be healthy in the UI


alena1108:
Forwardport https://github.com/rancher/rancher/issues/23394

izaac:
@alena1108 I see [this issue](https://github.com/rancher/rancher/issues/23394#issuecomment-552602152) in rancher `2.4 master-head (11/26/2019)` 781cd0fdc

We might've missed to forward port that fix to master? https://github.com/rancher/rancher/issues/23394 as I see that was fixed back in `2.3-head` and originally reported by @sangeethah

izaac:
Scenarios related to `v3/settings/ignore-node-name`  worked fine.

### I've reopened the issue due to this use case failing:

- Create a bipIp node from kubectl.
- Wait for it to get synced to rancher and get to "Unavailable" state.
- Using kubectl , add label - cattle.rancher.io/node-status: ignore. ( Did this step from kubectl , since the label change done from rancher UI did not take effect)
- Label gets synced to v3/nodes.
- Delete the node from rancher server using api ip. ( Attempted from api ui since UI does not have delete option)
- Node gets deleted from rancher side as well as from kubectl.

### Expected Behavior:

- Node gets deleted from rancher side .
- But node continues to be part of the cluster and can be listed from kubectl.

alena1108:
@izaac master branch has this commit 

`Skip v1.node removal for objects with specific nodename`

https://github.com/rancher/rancher/commit/9930781c8f2be0228ab273752ad090c1f55a9d65

I'll retest to see what's going on

alena1108:
@izaac the removal scenario works for exact name match, and broken for labels scenario only. The issue exists in both master and 2.3 though, as it looks like originally it was validated on 2.3 for the name match only https://github.com/rancher/rancher/issues/23394#issuecomment-554151930

izaac:
@alena1108 

This is working now in Rancher version `2.4 master-head (03/12/2019)` 41ffb6907

`cattle.rancher.io/node-status: ignore`
Ignore label is honored when deleting the node from the UI, I still see the node with kubectl.

The only thing I've noticed is related to `v3/settings/ignore-node-name`:

According to @sangeethah in this [acceptance criteria comment](https://github.com/rancher/rancher/issues/23394#issuecomment-552036024) I see the previously ignore-node-name node re-appearing in the UI after I set the settings value of `v3/settings/ignore-node-name` to a new one.

[...]

PR #24350, opened 2019-12-03
Respect ignore label on node removal

alena1108:
as we do respect the special name from settings

https://github.com/rancher/rancher/issues/23983

to fix the use case described in https://github.com/rancher/rancher/issues/23983#issuecomment-559210627


alena1108:
TF is unrelated; logged a separate issue for tracking https://github.com/rancher/rancher/issues/24365

Issue #23507, opened 2019-10-17
Support API Endpoint field for WeChat Notifier


WalkCloud:
Support API Endpoint field for WeChat Notifier.

Users should be able to use WeChat on-premise as notifier

PR #24353, opened 2019-12-03
[WIP] Support API URL for WeChat notifier

XianglongLuo:
Support API URL for WeChat notifier

https://github.com/rancher/rancher/issues/23507

Related PR:

https://github.com/rancher/types/pull/1044

### DIFF
--- a/pkg/controllers/user/alert/configsyncer/configsyncer.go
+++ b/pkg/controllers/user/alert/configsyncer/configsyncer.go
@@ -490,6 +490,7 @@ func (d *ConfigSyncer) addRecipients(notifiers []*v3.Notifier, receiver *alertco
 					AgentID:        notifier.Spec.WechatConfig.Agent,
 					CorpID:         notifier.Spec.WechatConfig.Corp,
 					Message:        `{{ template "wechat.text" . }}`,
+					APIURL:         notifier.Spec.WechatConfig.APIURL,
 				}
 
 				recipient := notifier.Spec.WechatConfig.DefaultRecipient
--- a/pkg/notifiers/sender.go
+++ b/pkg/notifiers/sender.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
+	"github.com/rancher/rancher/pkg/controllers/user/alert/config"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config/dialer"
 )
@@ -66,7 +67,7 @@ func SendMessage(notifier *v3.Notifier, recipient string, msg *Message, dialer d
 			recipient = s.DefaultRecipient
 		}
 		return TestWechat(notifier.Spec.WechatConfig.Secret, notifier.Spec.WechatConfig.Agent, notifier.Spec.WechatConfig.Corp, notifier.Spec.WechatConfig.RecipientType,
-			recipient, msg.Content, notifier.Spec.WechatConfig.HTTPClientConfig, dialer)
+			notifier.Spec.WechatConfig.APIURL, recipient, msg.Content, notifier.Spec.WechatConfig.HTTPClientConfig, dialer)
 	}
 
 	if notifier.Spec.WebhookConfig != nil {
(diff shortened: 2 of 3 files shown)


pkg/controllers/user/alert/configsyncer/configsyncer.go
```diff
@@ -490,6 +490,10 @@ func (d *ConfigSyncer) addRecipients(notifiers []*v3.Notifier, receiver *alertco
 					Message:        `{{ template "wechat.text" . }}`,
 				}
 
+				if notifier.Spec.WechatConfig.APIURL != "" {
+					wechat.APIURL = notifier.Spec.WechatConfig.APIURL
```
gitlawr: Can we put this  to the initialization block and get rid of the condition check

cjellick:
This PR should be marked as WIP until the types PR is updated and go.mod is no longer pointing at your fork. Don't want it to get accidentally merged.

XianglongLuo:
fix test_notifier

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,20 @@
+from .common import random_str
+
+
+def test_notifier_update(admin_mc, remove_resource):
```
gitlawr: The test name is too general concerning what's done in it. It would be better to change to test_notifier_wechat or test_notifier_wechat_apiurl

pkg/notifiers/sender.go
```diff
@@ -65,8 +66,7 @@ func SendMessage(notifier *v3.Notifier, recipient string, msg *Message, dialer d
 		if recipient == "" {
 			recipient = s.DefaultRecipient
 		}
-		return TestWechat(notifier.Spec.WechatConfig.Secret, notifier.Spec.WechatConfig.Agent, notifier.Spec.WechatConfig.Corp, notifier.Spec.WechatConfig.RecipientType,
-			recipient, msg.Content, notifier.Spec.WechatConfig.HTTPClientConfig, dialer)
+		return TestWechat(notifier.Spec.WechatConfig.Secret, notifier.Spec.WechatConfig.Agent, notifier.Spec.WechatConfig.Corp, notifier.Spec.WechatConfig.RecipientType, notifier.Spec.WechatConfig.APIURL, recipient, msg.Content, notifier.Spec.WechatConfig.HTTPClientConfig, dialer)
```
rmweir: nit: this is a very long line

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,20 @@
...
+        wechatConfig={"agent": "1111",
+                      "apiUrl": "test.com",
+                      "corp": "corp",
+                      "defaultRecipient": "2",
+                      "recipientType": "party",
+                      "secret": "231"})
+    remove_resource(n)
+    assert n is not None
+    assert (n["wechatConfig"]["apiUrl"]) == "test.com"
+    n = client.update(n, wechatConfig={"apiUrl": "test2.com"})
```
rmweir: what is the point of this part? just that it can be updated? We should probably test the default url too. I think maybe we can test the default url, then update it to a custom url and test that.

XianglongLuo:
@rmweir Thanks,updated.

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,21 @@
...
+        name=name,
+        wechatConfig={"agent": "1111",
+                      "apiUrl": defaultApiUrl,
+                      "corp": "corp",
+                      "defaultRecipient": "2",
+                      "recipientType": "party",
+                      "secret": "231"})
+    remove_resource(n)
+    assert n is not None
+    assert (n["wechatConfig"]["apiUrl"]) == defaultApiUrl
```
rmweir: You shouldn't have to set this right? The default url is set when this is blank, and that functionality should be tested.

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,21 @@
+from .common import random_str
+
+
+def test_notifier_wechat_apiurl(admin_mc, remove_resource):
```
rmweir: What is being tested? Shouldn't we attempt the "send" action? Otherwise, I don't think the added logic is being tested by this.

rmweir: I assume we should at least perform the action on two different inputs:
1. using a blank url- I would expect this would return as a success.
2. a url like test.com- I would expect this would return an error.

XianglongLuo: It needs proper tokens in order to get a successful test, and we don't want to include those in the tests. Does it make sense to you if we do a test to the default URL, and check an expected error?

loganhz: @XianglongLuo maybe you can try to add a unit test and mock `http` for it

Issue #23646, opened 2019-10-23
Allow option to bring back fluentd json log key parsing


cloudnautique:
The behavior json parsing of log data in the K8s metadata plugin changed between Rancher's packaging of Fluentd. The capabilities:

```
merge_json_log
preserve_json_log
``` 
no longer function. 

In order to bring this capability back for users that were relying on it, we need to run it through a fluentd filter block with parser.

```
<filter **>
  @type parser
  key_name log
  reserve_data true
  <parse>
    @type json
  </parse>
</filter>
```

This should be exposed as an option in both cluster and project.

cjellick:
@dramich to LOE

dramich:
Based on a conversation with @cloudnautique and looking at source we can potentially insert the above blob [here](https://github.com/rancher/rancher/blob/master/pkg/controllers/user/logging/generator/templatefilter.go#L35) which should make it work the same as pre upgrade.

dramich:
LOE 1 day

kinolaev:
Hello @cjellick,
just want to remind that parsing json log is bad default behavior https://github.com/rancher/rancher/issues/15880#issuecomment-517671647, sо if other clients relying on it, it should be exposed only as an option (as @cloudnautique suggest).

cjellick:
yeah, this would be the non-default behavior @kinolaev

dramich:
master pr: https://github.com/rancher/rancher/pull/24361
2.3 pr: https://github.com/rancher/rancher/issues/23646

izaac:
This option is not yet exposed in the UI: https://github.com/rancher/rancher/issues/24367

izaac:
UI was completed and ready To Test

izaac:
Rancher version: `2.2.9` `rancher/fluentd:v0.1.13` and gem version `fluent-plugin-kubernetes_metadata_filter (2.0.0)`

I was able to see how the following settings worked.

```
merge_json_log true
preserve_json_log true
```

The json sent to the fluend target (server) included the json log from an nginx workload log, preserved and merged.
```json
{
    "timestamp": "2019-12-11T18:51:35+00:00",
    "host": "x.x.x.x",
    "remote_addr": "10.42.2.0",
    "remote_user": "-",
    "body_bytes_sent": "612",
    "request_time": "0.000",
    "status": "200",
    "request": "GET / HTTP/1.1",
    "request_method": "GET",
    "http_referrer": "-",
[...]

PR #24361, opened 2019-12-03
[master] Add JSON parsing option to fluentd config

dramich:
Problem:
JSON parsing was dropped when the version of K8s metadata plugin was updated

Solution:
Add a flag to turn this option back on 

https://github.com/rancher/rancher/issues/23646


alena1108:
LGTM

Issue #24062, opened 2019-11-13
Update node-agent upgrade strategy


mrajashree:
For `node-agent` daemonset the `updateStrategy` is set to rollingUpdate. 
Configure it so that upgrade batchSize is 25% through rollingUpdate field `maxUnavailable`

Part of https://github.com/rancher/rancher/issues/23038

khushboo-rancher:
Node agent upgrade is working in batch of 25% when upgraded to master-head.
1. Rancher upgrade from v2.3.4 to master-head.
    Tested on a cluster of 10 nodes. 
    Node upgrade happened in batch of 25%. 2 Nodes upgraded at same time.  
    Yaml file - 
     updateStrategy:
           rollingUpdate:
                maxUnavailable: 25%
           type: RollingUpdate
2. Rancher upgrade from v2.3.4 to 2.3.5-rc2
     Tested on a cluster of 11 nodes.
      Node upgrade happened one by one.
3. Rancher rollbacked from master-head to v2.3.4
      Tested on a cluster of 11 nodes.
      Node upgrade happened one by one.
4. Tested for multiple clusters, functionality worked fine.

khushboo-rancher:
Node-agent upgrade strategy tested on HA env, functionality worked fine.

Tested on HA set up with 3 nodes(each node - 1etcd/1control/1worker)
ec2 cluster created with 8 nodes(1etcd/1control/6worker)
1. Upgraded from v2.3.4 to v2.3.5 - maxunavailable - 1 - upgraded one by one
2. Upgraded from v2.3.5 to rancher version ff4dfd753 - maxunavailable - 25% - upgrade 2 nodes at a time.

However, an issue #25154 is observed of failure cluser-agent pods.

PR #24363, opened 2019-12-03
Set node agent upgrade strategy to 25%

mrajashree:
https://github.com/rancher/rancher/issues/24062

### DIFF
--- a/pkg/systemtemplate/template.go
+++ b/pkg/systemtemplate/template.go
@@ -243,6 +243,8 @@ spec:
         name: docker-certs
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 25%
 
 {{- if .IsWindowsCluster}}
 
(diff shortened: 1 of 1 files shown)


prachidamle:
LGTM

dramich:
All three updated blocks have different formatting and fields, based off the docs they should all match:
```
updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 25%
```

pkg/systemtemplate/template.go
```diff
@@ -337,6 +339,8 @@ spec:
           type: DirectoryOrCreate
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+	  maxUnavailable: 25%
```
superseb: This has a tab before `maxUnavailable`

mrajashree: The tab was even before the column where updateStrategy starts and doesn't show up in the IDE :/

Issue #24273, opened 2019-11-27
Flaky test: test_app_namespace_annotation


mrajashree:
=================================== FAILURES ===================================
--
793 | ________________________ test_app_namespace_annotation _________________________
794 | [gw3] linux -- Python 3.7.5 /go/src/github.com/rancher/rancher/tests/integration/.tox/py37/bin/python3.7
795 |  
796 | admin_pc = <suite.conftest.ProjectContext object at 0x7fd167566c90>
797 | admin_mc = <suite.conftest.ManagementContext object at 0x7fd167b7ced0>
798 |  
799 | def test_app_namespace_annotation(admin_pc, admin_mc):
800 | client = admin_pc.client
801 | ns = admin_pc.cluster.client.create_namespace(name=random_str(),
802 | projectId=admin_pc.
803 | project.id)
804 | wait_for_template_to_be_created(admin_mc.client, "library")
805 | app1 = client.create_app(
806 | name=random_str(),
807 | externalId="catalog://?catalog=library&template=mysql&version=0.3.7"
808 | "&namespace=cattle-global-data",
809 | targetNamespace=ns.name,
810 | projectId=admin_pc.project.id,
811 | )
812 | wait_for_workload(client, ns.name, count=1)
813 |  
814 | external_id = "catalog://?catalog=library&template=wordpress" \
815 | "&version=1.0.5&namespace=cattle-global-data"
816 | app2 = client.create_app(
817 | name=random_str(),
818 | externalId=external_id,
819 | targetNamespace=ns.name,
820 | projectId=admin_pc.project.id,
821 | )
822 | wait_for_workload(client, ns.name, count=3)
823 | ns = admin_pc.cluster.client.reload(ns)
824 | ns = wait_for_app_annotation(admin_pc, ns, app1.name)
825 | ns = wait_for_app_annotation(admin_pc, ns, app2.name)
826 | client.delete(app1)
827 | wait_for_app_to_be_deleted(client, app1)
828 |  
829 | ns = admin_pc.cluster.client.reload(ns)
830 | assert app1.name not in ns.annotations['cattle.io/appIds']
831 | assert app2.name in ns.annotations['cattle.io/appIds']
832 |  
833 | client.delete(app2)
834 | wait_for_app_to_be_deleted(client, app2)
835 | ns = admin_pc.cluster.client.reload(ns)
836 | > assert 'cattle.io/appIds' not in ns.annotations
[...]

PR #24392, opened 2019-12-04
Update test_app_namespace_annotation

dramich:


dramich:
https://github.com/rancher/rancher/issues/24273

Issue #22900, opened 2019-09-17
Support S3 snapshot remove when rancher server is not configurated to access s3 bucket


moelsayed:
This is to support functional S3 backup for cases where IAM profiles are used to authenticate managed clusters cluster backup, while rancher server doesn't have access to use the same IAM profile. Use cases may include:
- Rancher server running outside of AWS, and cluster nodes are provisioned on AWS with IAM auth. 
- Rancher Server is running on AWS but in a different AWS account. 
- Rancher server nodes are not configured with the same IAM profile as the clusters for security reasons. 

An example use case: https://github.com/rancher/rancher/issues/22711#issuecomment-530177536

izaac:
Rancher version: `2.4 master-head (12/02/2019)` 1304c55a2

- Rancher server in Linode, the etcd nodes having IAM policy - cluster creation succeeds when access/secret key are not provided. Snapshot creation is successful
- Created a Daemonset workload, and an ingress.
- Triggered a manual snapshot.
- Clicked `Restore Snapshot`  in the cluster and selected the first snapshot created during cluster creation.

The cluster gets completly frozen in Rancher as it says the API is unavailable and the docker logs show lots for norman errors due to this:

`
[etcd] Failed to restore etcd snapshot: Failed to start [etcd-remove-snapshot] container on host [x.x.x.x]: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/opt/rke-tools/rke-etcd-backup\": stat /opt/rke-tools/rke-etcd-backup: no such file or directory": unknown
`
Reopening.

alena1108:
@izaac does restore work fine in Linode cluster that has API/Secret key set in the snapshot configuration section of a cluster? As opposed to being read from IAM creds on the node.

izaac:
@alena1108 I will check that scenario too. I was going through the first of these: 

https://github.com/rancher/rancher/issues/22711#issuecomment-531018019

izaac:
@alena1108 it does fail with credentials too. Same error message:

`[etcd] Failed to restore etcd snapshot: Failed to start [etcd-remove-snapshot] container on host [x.x.x.x]: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/opt/rke-tools/rke-etcd-backup\": stat /opt/rke-tools/rke-etcd-backup: no such file or directory": unknown`

alena1108:
[...]

PR #24398, opened 2019-12-04
use correct snapshot name on its removal

alena1108:
https://github.com/rancher/rancher/issues/22900

### DIFF
--- a/pkg/controllers/management/etcdbackup/etcdbackup.go
+++ b/pkg/controllers/management/etcdbackup/etcdbackup.go
@@ -200,13 +200,13 @@ func (c *Controller) etcdSaveWithBackoff(b *v3.EtcdBackup) (runtime.Object, erro
 		return b, err
 	}
 
+	snapshotName := clusterprovisioner.GetBackupFilename(b)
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -237,7 +237,8 @@ func (c *Controller) etcdRemoveSnapshotWithBackoff(b *v3.EtcdBackup) error {
 		return err
 	}
 	return wait.ExponentialBackoff(backoff, func() (bool, error) {
-		if inErr := c.backupDriver.ETCDRemoveSnapshot(c.ctx, cluster.Name, kontainerDriver, cluster.Spec, b.Name); inErr != nil {
+		snapshotName := clusterprovisioner.GetBackupFilename(b)
```
prachidamle: Can we move line#240 out of the ExponentialBackoff loop so it is invoked only once?

prachidamle:
LGTM

dramich:
LGTM

PR #24405, opened 2019-12-04
[v2.3] Catalog log fix

brendarearden:
**Problems:**
The count for catalog templates was increasing after create or update was called even if they threw an error. TemplateSpec value "UpgradeVersionLinks" was being set prior to DeepEquals being called during update causing the template to indicate that it needed updating when no values had changed. Long label name was erring after template was being created.

**Solution:**
Increment count only where there is not an error. Move "UpgradeVersionLinks" to be set only when it is needed. Add log to indicate files had a failure. Make errors indicate when the error occurred instead of being generic. Check for label length before allowing template to create. Delete template during delete if template version doesn't exist to ensure catalog delete occurs successfully. Reduce the number of times update is called on the catalog to reduce logging. UI no longer displays the individual templates that are syncing, but that the catalog is syncing.

**Issue:**
#23368
#19724

**Master PR**
https://github.com/rancher/rancher/pull/24176

### DIFF
--- a/pkg/catalog/manager/catalog_common.go
+++ b/pkg/catalog/manager/catalog_common.go
@@ -104,7 +104,7 @@ func (m *Manager) updateCatalogInfo(cmt *CatalogInfo, catalogType string, templa
 		}
 		v3.CatalogConditionRefreshed.Unknown(obj)
 		if templateName != "" {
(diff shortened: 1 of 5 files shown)


dramich:
LGTM

Issue #24234, opened 2019-11-25
Bug: nodes will not delete if aws key has been deleted/disabled even after updating nodetemplate.


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1) Create an ec2 cluster with some nodes
2) Deactivate/disable/delete AWS keys on AWS
3) Create new AWS keys on AWS
4) Update node template with new AWS keys
5) Attempt to delete node created before AWS keys were updated

**Result:**
Node will appear to be deleted but is actually still running in AWS

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

askldjd:
Just to provide a bit more information. Our Rancher node template were created before the existence of Cloud Credential. So the AWS access_key were directly embedded in the template. Here's the full sequence.

1. Deactivate existing AWS key
1. Create a new Cloud Credential with a new AWS key
1. Update each template to use the new Cloud Credential from step 2.
1. Delete a node that was created using the deactivated key. Notice that the node is never deleted in AWS. There is no error and the failure is silent. However, the node pool will immediately recreate the a new node using the new Cloud Credential.
1. Delete the newly created node, confirm that the node is actually deleted in AWS.

It seems that the AWS key that was used to create the node stays in the lifecycle of the node. Even if the node template is updated, it will not take into effect until the next lifecycle. This is very problematic for production because we rotate AWS keys often. As keys rotate, we are affecting stability of production because traffics are routed through these orphaned nodes.

askldjd:
Another update:

It turns out that it doesn't matter if you are using Access Keys in Node Template or Cloud Credential. The key that Rancher uses to create the node will always be the same key it uses to destroy the node. So whenever any AWS credentials gets deactivated, Rancher will lose the ability to delete all the nodes created by the deactivated key.

aaronRancher:
Able to reproduce bug using steps above in `v2.2.8` and `v2.3.3`.

dramich:
Master pr: https://github.com/rancher/rancher/pull/24435
2.3 pr: https://github.com/rancher/rancher/pull/24490

sangeethah:
@dramich Reopening this issue to get a fix in `2.3-release` branch.

aaronRancher:
[...]

PR #24435, opened 2019-12-05
Refresh credentials on node removal

dramich:
Problem:
The same credential used to create a node is used to delete it so if a
cloud cred is updated or keys rotated nodes will leak

Solution:
Before doing an update or delete of a node rebuild the credentials using
the latest cloud cred referenced in the template

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -43,10 +43,6 @@ const (
 	amazonec2               = "amazonec2"
 )
 
-var (
-	falseValue = false
-)
-
 // aliases maps Schema field => driver field
 // The opposite of this lives in pkg/controllers/management/drivers/nodedriver/machine_driver.go
 var aliases = map[string]map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -407,6 +394,9 @@ func (m *Lifecycle) ready(obj *v3.Node) (*v3.Node, error) {
 	}
 	defer config.Cleanup()
 
+	// Refresh first to save the compressed config, then Restore writes it to disk for machine to use
+	m.refreshNodeConfig(config, obj)
```
alena1108: Is it intended to not check for errors returned by refreshNodeConfig call?

pkg/controllers/management/node/controller.go
```diff
@@ -551,6 +541,56 @@ func (m *Lifecycle) saveConfig(config *nodeconfig.NodeConfig, nodeDir string, ob
...
+
+	rawTemplate, err := m.nodeTemplateGenericClient.GetNamespaced(template.Namespace, template.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	data := rawTemplate.(*unstructured.Unstructured).Object
+	rawConfig, ok := values.GetValue(data, template.Spec.Driver+"Config")
+	if !ok {
+		return fmt.Errorf("node config not specified")
```
alena1108: we should probably add more information to the log. like node name

pkg/controllers/management/node/controller.go
```diff
@@ -551,6 +541,56 @@ func (m *Lifecycle) saveConfig(config *nodeconfig.NodeConfig, nodeDir string, ob
...
+
+	newConfig := string(bytes)
+
+	currentConfig, err := nc.DriverConfig()
+	if err != nil {
+		return err
+	}
+
+	if currentConfig != newConfig {
+		err = nc.SetDriverConfig(string(bytes))
```
alena1108: nit: we can just `return nc.SetDriverConfig(string(bytes))`

dramich: If that doesn't error save is called below. Returning would stop from saving.

pkg/controllers/management/node/controller.go
dramich: No, updated all calls and error message.

pkg/controllers/management/node/controller.go
dramich: Updated

PR #24442, opened 2019-12-05
Drop clustertemplaterevisions-create from user-base and add new global role

prachidamle:
Drop clustertemplaterevisions-create from user-base roletemplate and add new global role "Create RKE Template Revisions" which can be assigned separately when needed.


alena1108:
@prachidamle is there a ticket? if not, do you think we have to create it for qa verification?

PR #24465, opened 2019-12-06
Automation Scenario for checking create/delete snapshots in S3 store

izaac:


soumyalj:
@izaac : Why is the s3 snapshot deete test in rke_cluster_provisioning.py ?

izaac:
> @izaac : Why is the s3 snapshot delete test in rke_cluster_provisioning.py ?

That's a good point I wanted to check, because this backup/restore requires a special cluster provisioning to set the S3 backups and the tests in that pytest file are using the ones in the default `rke_config`. And then I'm provisioning a cluster with the proper S3 options. 

Suggestions for a better structure are welcome.

izaac:
@soumyalj I was thinking in making those S3 cluster snapshot settings as defaults in the original `rke_config` Not sure if that would impact other tests though.

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -65,6 +79,50 @@ def test_bkp_restore():
     assert len(workload_list) == 0
 
 
+def test_s3_snapshot_delete(node_template_ec2_iam):
```
sowmyav27: Can we rename the method to include "create" scenario also, something like test_s3_snapshot_create_delete ?

izaac: good catch. Renamed

soumyalj:
LGTM for the existing test case.

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -1,10 +1,26 @@
...
+AWS_SECRET_ACCESS_KEY = os.environ.get("AWS_SECRET_ACCESS_KEY")
+AWS_REGION = os.environ.get("AWS_REGION")
+AWS_SUBNET = os.environ.get("AWS_SUBNET")
+AWS_VPC = os.environ.get("AWS_VPC")
+AWS_SG = os.environ.get("AWS_SG")
+AWS_ZONE = os.environ.get("AWS_ZONE")
+AWS_IAM_PROFILE = os.environ.get("AWS_IAM_PROFILE", "")
+AWS_S3_BUCKET_NAME = os.environ.get("AWS_S3_BUCKET_NAME", "")
+AWS_S3_BUCKET_FOLDER_NAME = os.environ.get("AWS_S3_BUCKET_FOLDER_NAME", "")
+LINODE_ACCESSKEY = os.environ.get('RANCHER_LINODE_ACCESSKEY', "None")
```
sangeethah: @izaac can we moved all the AWS and LINODE related env variables to common.py so that they can reside in 1 place. Also remove them from test_rke_cluster_provisioning.py.

tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
```diff
@@ -716,11 +714,25 @@ def validate_rke_dm_host_2(node_template,
...
     cluster_cleanup(client, cluster)
 
 
+def validate_rke_dm_host_2_no_cluster_cleanup(node_template,
+                           rancherKubernetesEngineConfig=rke_config):
+    client = get_user_client()
+    nodes = validate_rke_dm_host_2_build_nodes(node_template)
+    create_and_vaildate_cluster(client, nodes, rancherKubernetesEngineConfig)
+
+
```
sangeethah: There is no need for validate_rke_dm_host_2_no_cluster_cleanup method. cluster_cleanup method in existing method validate_rke_dm_host_2 that can be modified to not call cluster_cleanup by passing an additional parameter to this method - attemptDelete with default to "True" which can be overridden to false for your workflow.

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -65,7 +81,85 @@ def test_bkp_restore():
...
-@pytest.fixture(scope='module', autouse="True")
+def test_bkp_restore_s3_with_creds(create_project_client_and_cluster_s3_with_creds):
+    validate_s3_create_restore_delete()
+
+
+def test_bkp_restore_s3_with_iam(create_project_client_and_cluster_s3_with_iam):
+    validate_s3_create_restore_delete()
+
+
+def validate_s3_create_restore_delete():
```
sangeethah: Can the S3 checks in this method be folded into the existing validate_test_bkp_restore by making it take a mode parameter that can be set to local Vs s3?

tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
izaac: This one is really tricky as the cluster provisioning methods cleanup the cluster based on that variable and if I set it to `False` I will end with a cluster not being tear down at the end of the automation execution having to manually clean up the resources.

So this attempt was to make sure Im not removing the cluster after the original `validate_rke_dm_host_2` because the backup tests happen **after** the initial cluster validations. And cleanup the cluster after the backup tests.

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -1,21 +1,34 @@
...
+def validate_backup_create_restore_delete(backup_mode):
     p_client = namespace["p_client"]
     ns = namespace["ns"]
     client = get_user_client()
     cluster = namespace["cluster"]
     name = random_test_name("default")
 
-    if not hasattr(cluster, 'rancherKubernetesEngineConfig'):
-        assert False, "Cluster is not of type RKE"
-
```
sangeethah: @izaac We will need this check to prevent the test cases from running in a non-rke cluster in the case of test_bkp_restore_local

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -1,21 +1,34 @@
...
     con = [{"name": "test1",
-           "image": TEST_IMAGE}]
+            "image": TEST_IMAGE}]
```
sangeethah: This spacing change will result in flake8 error.

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -29,16 +42,25 @@ def test_bkp_restore():
...
     backupId = etcdbackupdata[0]['id']
-    wait_for_backup_to_active(client, cluster, backupname)
+
+    backupfilename = ""
+    if backup_mode == "s3":
+        backupfileurl = etcdbackupdata[0]['filename']
+        # Check the backup filename exists in S3
+        parseurl = urlparse(backupfileurl)
+        backupfilename = os.path.basename(parseurl.path)
+        assert AmazonWebServices().s3_backup_check(backupfilename)
```
sangeethah: Can we do a similar check for local snapshot as well ? In this case we should check for the snapshot to be present as s file in all  etcd nodes?

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -55,17 +77,23 @@ def test_bkp_restore():
...
     workload_list = p_client.list_workload(uuid=testworkload.uuid).data
     print(len(workload_list))
     assert len(workload_list) == 0
 
+    if backup_mode == "s3":
+        # Check the backup reference is deleted in Rancher and S3
+        client.delete(cluster.etcdBackups(id=backupId)['data'][0])
+        wait_for_backup_to_delete(cluster, backupname)
+        assert AmazonWebServices().s3_backup_check(backupfilename) is False
+    cluster_cleanup(client, cluster)
```
sangeethah: Can we do a similar delete check for local snapshot as well ?

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -1,21 +1,34 @@
...
+    validate_rke_dm_host_2
 
 namespace = {"p_client": None, "ns": None, "cluster": None, "project": None}
 
 
-def test_bkp_restore():
+def test_bkp_restore_local(create_project_client):
+    validate_backup_create_restore_delete("local")
+
+
```
sangeethah: Define a variable to skip S3 related cases , since we dont want them to be added to onTag job .

tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
```diff
@@ -1,31 +1,11 @@
...
-AWS_SG = os.environ.get("AWS_SG")
-AWS_ZONE = os.environ.get("AWS_ZONE")
-AWS_IAM_PROFILE = os.environ.get("AWS_IAM_PROFILE", "")
-AZURE_SUBSCRIPTION_ID = os.environ.get("AZURE_SUBSCRIPTION_ID")
-AZURE_CLIENT_ID = os.environ.get("AZURE_CLIENT_ID")
-AZURE_CLIENT_SECRET = os.environ.get("AZURE_CLIENT_SECRET")
-AZURE_TENANT_ID = os.environ.get("AZURE_TENANT_ID")
-worker_count = int(os.environ.get('RANCHER_STRESS_TEST_WORKER_COUNT', 1))
-HOST_NAME = os.environ.get('RANCHER_HOST_NAME', "testcustom")
-
```
sangeethah: Can you move out only the AWS and linode related variiables out to common.py and leave the rest of them here.

tests/validation/tests/v3_api/test_bkp_restore.py
izaac: That would require ssh into each of the etcd nodes' filesystem and look for the filename, that's beyond the scope of this PR. I can work on that in a separate PR as it doesn't sound a trivial addition.

tests/validation/tests/v3_api/test_bkp_restore.py
izaac: See above

tests/validation/tests/v3_api/test_bkp_restore.py
izaac:
```bash
pipenv run flake8 tests/v3_api/test_bkp_restore.py --ignore=F405,F401,F811
/Users/izaac/.local/share/virtualenvs/validation-pB2uESK5/lib/python3.7/site-packages/pycodestyle.py:113: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
```

I've changed it back but it wasn't throwing an error before or after:

sangeethah:
LGTM

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -55,17 +84,23 @@ def test_bkp_restore():
...
     workload_list = p_client.list_workload(uuid=testworkload.uuid).data
     print(len(workload_list))
     assert len(workload_list) == 0
 
+    if backup_mode == "s3":
+        # Check the backup reference is deleted in Rancher and S3
+        client.delete(cluster.etcdBackups(id=backupId)['data'][0])
+        wait_for_backup_to_delete(cluster, backupname)
+        assert AmazonWebServices().s3_backup_check(backupfilename) is False
+    cluster_cleanup(client, cluster)
```
sowmyav27: Can we have the "cluster_cleanup(client, cluster)" in a finalizer in the method where the cluster is created?

soumyalj:
LGTM

tests/validation/tests/v3_api/test_bkp_restore.py
izaac: I've moved the cleanup_cluster to the fixtures finalizers. It works I did testing `RANCHER_CLEANUP_CLUSTER=True` and `False. @sowmyav27

sowmyav27:
LGTM

PR #24467, opened 2019-12-07
Move user controllers to cluster agent

ibuildthecloud:


pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -227,6 +229,11 @@ func (r *RemoteService) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		er.Error(rw, req, err)
 		return
 	}
+
+	// TODO: THIS IS SO BAD!!!!
```
alena1108: :)

Issue #24335, opened 2019-12-02
Notifier email configure return plaintext password


aiwantaozi:
Notifier return plaintext password in API request

![image](https://user-images.githubusercontent.com/5936271/69937677-56538780-1516-11ea-9e1c-b8d2cc35962f.png)

@loganhz could you help add milestone

alena1108:
@aiwantaozi we have to make sure to:

1) add a test
2) validate that the subsequent update use case will work, from both the test + API. There were issues in the past when the password was reset on the crd when the UI was setting an empty value. Steps to take:
* Setup notifier with the password. Make sure the password is set on crd, and not shown in api
* Edit notifier. Change something non-password related, hit save. Make sure that the password in crd remains the same

bmdepesa:
rancher/rancher:master-head a70698d25

* Reproduced in v2.3.5
* Deploy`rancher/rancher:master-head`
* Create a cluster
* Create an email notifier w/ password set
* View notifier in API and no password field is present
* Check CRD on server and see that password is set correctly
* Update password in notifier and see it reflected in CRD
* Update some non-password field (username in this case) and see the field updated, and password is not changed

PR #24477, opened 2019-12-09
Test for setting email notifier SMTP password

XianglongLuo:
Problem: Notifier email configure return plaintext password
Solution: update type to password
Related PR:https://github.com/rancher/types/pull/1051
Issue: https://github.com/rancher/rancher/issues/24335

### DIFF
--- a/tests/integration/suite/test_notifier.py
+++ b/tests/integration/suite/test_notifier.py
@@ -1,0 +1,62 @@
+from kubernetes.client import CustomObjectsApi
+from .common import random_str
+
+
+def test_notifier_smtp_password(admin_mc, remove_resource):
+    client = admin_mc.client
+    name = random_str()
+    password = random_str()
+    notifier = client.create_notifier(clusterId="local",
+                                      name=name,
+                                      smtpConfig={
+                                          "defaultRecipient": "test",
+                                          "host": "test",
+                                          "port": "587",
+                                          "sender": "test",
+                                          "tls": "true",
+                                          "username": "test",
+                                          "password": password
+                                      })
+    remove_resource(notifier)
+    assert notifier is not None
+
+    # Test password not present in api
+    assert notifier['smtpConfig'].get('password') is None
+
+    crd_client = get_crd_client(admin_mc)
+    ns, name = notifier["id"].split(":")
+    # Test password is in k8s  after creation
+    verify_smtp_password(crd_client, ns, name, password)
+    # Test noop, password field should be as it is
+    notifier = client.update(notifier, smtpConfig=notifier['smtpConfig'])
+    verify_smtp_password(crd_client, ns, name, password)
+    # Test updating password
+    new_password = random_str()
+    notifier = client.update(notifier, smtpConfig={
+        "password": new_password})
+    verify_smtp_password(crd_client, ns, name, new_password)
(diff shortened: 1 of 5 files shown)


tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,62 @@
...
+    notifier = client.reload(notifier)
+    assert notifier["smtpConfig"]["username"] == "test2"
+
+    validate_password_exists_in_crd(admin_mc, password)
+
+
+def validate_password_exists_in_crd(admin_mc, password):
+    cluster, client = cluster_and_client('local', admin_mc.client)
+    cc = ClusterContext(admin_mc, cluster, client)
+    create_kubeconfig(cc.cluster)
```
kinarashah: we don't need to create kubeconfig to check the output of password, we can use k8s client, 
for example, crdClient used here: https://github.com/rancher/rancher/blob/master/tests/integration/suite/test_password_store.py#L175

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,62 @@
+from kubernetes.client import CustomObjectsApi
+from .common import random_str
+
+group = 'management.cattle.io'
+version = 'v3'
+namespace = 'local'
+plural = 'notifiers'
```
dramich: This set of things should be held in a dict, preferably inside of `verifySmtpPassword`, as they are specific to that function and having them global can cause potential issues in the future.

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,62 @@
...
+clusterId = "local"
```
dramich: This isn't used anywhere

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,62 @@
...
+    # Test the password in crd remains the same value after updating username
+    verifySmtpPassword(crdClient, ns, name, newPassword)
+
+
+def verifySmtpPassword(crdClient, ns, name, Password):
+    k8notifier = crdClient.get_namespaced_custom_object(
+        group, version, namespace, plural, name)
+    smtpPassword = k8notifier['spec']['smtpConfig']['password']
+
+    assert smtpPassword is not None
```
dramich: The None check isn't needed

tests/integration/suite/test_notifier.py
```diff
@@ -0,0 +1,62 @@
...
+    verifySmtpPassword(crdClient, ns, name, newPassword)
+    # Test updating field non-password related
+    notifier = client.update(notifier, smtpConfig={"username": "test2"})
+    notifier = client.reload(notifier)
+    assert notifier["smtpConfig"]["username"] == "test2"
+    # Test the password in crd remains the same value after updating username
+    verifySmtpPassword(crdClient, ns, name, newPassword)
+
+
+def verifySmtpPassword(crdClient, ns, name, Password):
```
dramich: `Password` should not be capitalized

dramich:
There are multiple places that camelCase is being used for vars and func declarations, this should follow our convention of using snake case. https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

XianglongLuo:
updated,Thanks.

PR #24482, opened 2019-12-09
Hosted cluster dialer fix to work with proxy

alena1108:
https://github.com/rancher/rancher/issues/22006


deniseschannon:
https://github.com/rancher/rancher/issues/23293

Issue #18927, opened 2019-03-14
FreeIPA group memberships not showing up in cluster membership auto complete list


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1) setup FreeIPA authentication
2) attempt to add a group to a cluster membership

**Result:**
No groups show up in the auto complete list

**Other details that may be helpful:**
Ping me for logs or additional information

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):v2.1.7
- Installation option (single install/HA): HA

mrajashree:
This fix worked for the setup (customer) that didn't have entryDN attribute. Our test setup returns that attribute. So we may not be able to reproduce the bug

deniseschannon:
Per @mrajashree's comments, our test setup aren't missing these attributes so we won't be able to test this exact use case, but we should run at least one pass to ensure FreeIPA setups are still working as expected.

soumyalj:
Tested with master-head branch(bdb8e67c907d)
Enabled auth with Free IPA. Adding groups/users to clusters and project members is functional. 
"Search Filter" field option is presented for User and Groups in "Customize Schema" section.
Users have the option to set these fields when FreeIPA auth is enabled. When the field is set , this filter gets applied to the list of users/groups that get searched when adding users to add members to cluster/projects.

PR #24486, opened 2019-12-09
Make freeipa group membership work without entrydn

mrajashree:
Some Freeipa servers seem to not return the entrydn attribute,
which we use in our query to get all details of the groups of a
logged in user. So in such cases, we will use queries with member
attribute to get all first level groups, and the nested logic we
have for openldap to retrieve nested parent groups.

https://github.com/rancher/rancher/issues/18927

### DIFF
--- a/pkg/auth/providers/ldap/ldap_client.go
+++ b/pkg/auth/providers/ldap/ldap_client.go
@@ -102,6 +102,7 @@ func (p *ldapProvider) getPrincipalsFromSearchResult(result *ldapv2.SearchResult
 	var nonDupGroupPrincipals []v3.Principal
 	var userScope, groupScope string
 	var nestedGroupPrincipals []v3.Principal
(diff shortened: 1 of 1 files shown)


prachidamle:
LGTM

pkg/auth/providers/ldap/ldap_client.go
```diff
@@ -140,6 +141,7 @@ func (p *ldapProvider) getPrincipalsFromSearchResult(result *ldapv2.SearchResult
 	}
 
 	userPrincipal = *user
+	userDN := result.Entries[0].DN
```
alena1108: can entries be ever empty? :)

mrajashree: it can't be because we ensure in the methods `loginUser` and `RefetchGroupPrincipals` that call `getPrincipalsFromSearchResult` that len(result.Entries) = 1

PR #24487, opened 2019-12-09
[2.3.4] Drop clustertemplaterevisions-create from user-base and add it as a separate global role

prachidamle:
[backport from  master] Drop clustertemplaterevisions-create from user-base and add it as a separate global role


prachidamle:
@alena1108 can you review/merge this backport PR

PR #24490, opened 2019-12-09
Refresh credentials on node removal

dramich:
Problem:
The same credential used to create a node is used to delete it so if a
cloud cred is updated or keys rotated nodes will leak

Solution:
Before doing an update or delete of a node rebuild the credentials using
the latest cloud cred referenced in the template


dramich:

PR #24493, opened 2019-12-10
when I review the code here, the state of mind has fallen

zehuaiWANG:
please remove the comment


dramich:
This is in a vendored package, the PR would need to happen against the norman repo.

Issue #24502, opened 2019-12-10
ResourceQuota sync controller deepEqual bug


Jason-ZW:
## Description
Using `reflect.DeepEqual` to compare `resource.Quantity` can lead to the compare results are not expected. This problem is particularly acute when setting CPU related quotas. [Code Line 1](https://github.com/rancher/rancher/blob/master/pkg/controllers/user/resourcequota/resource_quota_sync.go#L103-L109) [Code Line 2](
https://github.com/rancher/rancher/blob/master/pkg/controllers/user/resourcequota/resource_quota_sync.go#L145)

## Steps to reproduce
1. add `logrus` log to the code place
2. create project with cpu limit & cpu request quota
3. update project without change anything

## Debug Log:
```
requests.cpu:{i:{value:1 scale:0} d:{Dec:<nil>} s:1 Format:DecimalSI}
requests.cpu:{i:{value:1000 scale:-3} d:{Dec:<nil>} s: Format:DecimalSI}
```

```
2019/12/10 16:50:31 [INFO] src quota hard: map[limits.cpu:{i:{value:1500 scale:-3} d:{Dec:<nil>} s:1500m Format:DecimalSI} requests.cpu:{i:{value:1 scale:0} d:{Dec:<nil>} s:1 Format:DecimalSI} secrets:{i:{value:50 scale:0} d:{Dec:<nil>} s:50 Format:DecimalSI}]
2019/12/10 16:50:31 [INFO] dst quota hard: map[limits.cpu:{i:{value:1500 scale:-3} d:{Dec:<nil>} s:1500m Format:DecimalSI} requests.cpu:{i:{value:1000 scale:-3} d:{Dec:<nil>} s: Format:DecimalSI} secrets:{i:{value:50 scale:0} d:{Dec:<nil>} s:50 Format:DecimalSI}]
```

gitlawr:
Do you have steps to expose this bug from an end user's view? It can also help QA to validate the issue.

Jason-ZW:
## How to test
1. create a project with `limitCPU=1000` and `requestCPU=1000` milli  
2. watch the rancher logs (level: info)
3. update project without change anything

This log is incorrect if it appears: 
```
Updating default resource quota for namespace xxx
```

soumyalj:
Reproduced the issue with Rancher v2.3.3.
1. Created a Rancher server with CATTLE_DEV_MODE=true as below:
```
[...]

PR #24503, opened 2019-12-10
Fix resource_quota_sync deepEqual bug

Jason-ZW:
## Problem(#24502)
Using `reflect.DeepEqual` to compare `resource.Quantity` can lead to the compare results are not expected. This problem is particularly acute when setting CPU related quotas. [Code Line 1](https://github.com/rancher/rancher/blob/master/pkg/controllers/user/resourcequota/resource_quota_sync.go#L103-L109) [Code Line 2](
https://github.com/rancher/rancher/blob/master/pkg/controllers/user/resourcequota/resource_quota_sync.go#L145)

## Solution
Using `k8s.io/apimachinery/pkg/api/equality` [Sementic](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/equality/semantic.go#L31) Ignore formatting, only care that numeric value stayed the same. [References kubernetes quotaController code](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/resourcequota/resource_quota_controller.go#L214)

### DIFF
--- a/pkg/controllers/user/resourcequota/resource_quota_sync.go
+++ b/pkg/controllers/user/resourcequota/resource_quota_sync.go
@@ -16,6 +16,7 @@ import (
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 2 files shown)


prachidamle:
@Jason-ZW maybe your test can make sure no unnecessary updates happen due to project updates. 
Or @dramich suggests using go test if it is mainly on log watch

Jason-ZW:
@prachidamle @dramich @loganhz  already added a unit test.

Issue #23540, opened 2019-10-18
rancher doesnt detect chart updates on harbor


aemneina:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. update a chart on a harbor hosted helm repo
2. click refresh on the catalog within rancher

**Result:**
Rancher doesnt detect a hash change on harbor side, so doesnt think there are any new changes to pull. The workaround was to delete the local caches and pull again. Not sure if there are any better work arounds but thats tough in terms of having an automated process.

izaac:
Rancher version: `2.4 master-head (12/10/2019)`
Rancher version: `2.3.0`
Harbor Version: `v1.9.3-c80751b8`

I tried this by creating a public repo with one helm chart. Pushing new versions and updating from the Rancher catalog UI at Global level.

I wasn't able to reproduce.

brendarearden:
**Status Update**: We have added debug logging to help troubleshoot issues with helm charts. We are still working to track down this specific issue so this will remain open until we gather additional information.

maggieliu:
Status Update: Since the ZD ticket is closed & we cannot reproduce, moving this issue to unschedule, until customer creates another ZD ticket.

izaac:
Please open a new issue if a new ZD is created.

Issue #23256, opened 2019-10-07
catalog refreshes constantly


mattmattox:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Add a catalog from an internal repo served by Harbor for helm charts

**Result:**
kubectl logs -n cattle-system --tail 10 rancher-7fbb8ff996-j7ztv
2019/08/23 15:41:37 [INFO] Updating project catalog ***********
2019/08/23 15:41:37 [INFO] Updating project catalog ***********-***********
2019/08/23 15:41:37 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted
2019/08/23 15:41:37 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted
2019/08/23 15:41:38 [INFO] Updating project catalog ***********
2019/08/23 15:41:38 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted
2019/08/23 15:41:39 [INFO] Updating project catalog ***********
2019/08/23 15:41:39 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted
2019/08/23 15:41:39 [INFO] Updating project catalog ***********
2019/08/23 15:41:39 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted

**Other details that may be helpful:**
- This worked fine with v2.2.7 but after upgrading to v2.2.8 this issue started

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

splattner:
same here

```
2020/02/12 12:54:14 [INFO] Updating global catalog *****
2020/02/12 12:54:14 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted, 0 templates failed
2020/02/12 12:54:15 [INFO] Updating global catalog *****
2020/02/12 12:54:15 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted, 0 templates failed
2020/02/12 12:54:16 [INFO] Updating global catalog *****
2020/02/12 12:54:16 [INFO] Catalog sync done. 0 templates created, 0 templates updated, 0 templates deleted, 0 templates failed
2020/02/12 12:54:17 [INFO] Updating global catalog *****
[...]

PR #24531, opened 2019-12-11
[v2.3] Add debug for chart hash

brendarearden:
**Problem**
Creating a Rancher catalog using charts served from Harbor is causing issues where Rancher is not updating catalog/chart information as expected. 

**Solution**
This does not resolve the issue. This will assist with future debugging when charts are not updating as expected. 

**Issues**
#23540 
#23256

### DIFF
--- a/pkg/catalog/manager/catalog_sync.go
+++ b/pkg/catalog/manager/catalog_sync.go
@@ -41,6 +41,7 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 	if err != nil {
 		return m.updateCatalogError(catalog, err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/catalog/manager/traverse.go
```diff
@@ -25,6 +25,7 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 	catalogType := getCatalogType(cmt)
 
 	index, err := helm.LoadIndex()
+	logrus.Debugf("helm index hash: %v", index.Hash)
```
kinarashah: if this is only for debugging purposes, you think adding catalog/chart name/info would help?

dramich:
LGTM

PR #24532, opened 2019-12-11
[forwardport] Add debug for chart hash

brendarearden:
**Problem**
Creating a Rancher catalog using charts served from Harbor is causing issues where Rancher is not updating catalog/chart information as expected.

**Solution**
This does not resolve the issue. This will assist with future debugging when charts are not updating as expected.

**Related PR**
#24531 

**Issues**
#23540
#23256


dramich:
LGTM

PR #24533, opened 2019-12-11
Fix for cluster driver reset

alena1108:
There can be a situation when the `cluster.status.driver` field remains empty on rke clusters even after the cluster is provisioned. The issue was addressed by @rmweir in the following commit:

https://github.com/rancher/rancher/commit/03dd1256a789ccb11299c3edea23549466a0cb04

This particular PR is to address the issue when cluster-agent on its connect sets the `cluster.status.driver` field to "imported" if the field is empty. That results in provisioning code never being invoked on subsequent cluster updates for rke clusters. 

The fix introduces a check for figuring out if the cluster really qualifies for imported, before marking it this way. So if you are upgrading from the older version of rancher where the `cluster.status.driver`  is already set to "" on rke cluster(s), it will never get reset by the agent code.

### DIFF
--- a/pkg/tunnelserver/tunnel.go
+++ b/pkg/tunnelserver/tunnel.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/rancher/norman/types/convert"
+	provisioner "github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
 	"github.com/rancher/rancher/pkg/taints"
 	"github.com/rancher/remotedialer"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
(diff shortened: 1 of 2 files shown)


pkg/tunnelserver/tunnel.go
```diff
@@ -279,7 +282,13 @@ func (t *Authorizer) authorizeCluster(cluster *v3.Cluster, inCluster *cluster, r
 	}
 
 	changed := false
-	if cluster.Status.Driver == "" {
+
+	driver, err := provisioner.GetDriver(cluster, t.KontainerDriverLister)
```
ibuildthecloud: The old code would only set the cluster.Status.Driver field if is empty.  This will now always "reset" it, is that intended? Additionally will always set changed to true which may result in a etcd write per node checkin.

alena1108: @ibuildthecloud 

> This will now always "reset" it, is that intended?

It will only reset it if the driver is empty - line 290 has this check. The only diff this pr brings is  - `provisioner.GetDriver` call is more deterministic, and will never return empty value if RKE (or any other k8s provider config) is returned.

ibuildthecloud: Thanks

PR #24537, opened 2019-12-11
Modify machine config on disk

dramich:
Problem:
The machine config does not use updated cloud credentials for updates
and deletes

Solution:
Before running machine, update the config on disk with updated cloud
creds

### DIFF
--- a/pkg/nodeconfig/config.go
+++ b/pkg/nodeconfig/config.go
@@ -1,11 +1,12 @@
 package nodeconfig
 
 import (
+	"encoding/json"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
-	"encoding/json"
-
+	"github.com/pkg/errors"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
 	"github.com/rancher/rancher/pkg/encryptedstore"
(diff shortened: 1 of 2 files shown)


dramich:
The previous fix did update the config but that only counts for new nodes being created which was getting updated anyways. The issue comes from rancher-machine writing a config to disk that includes API keys and is not being updated. The machine remove command does not take any arguments and simply uses the already existing config that exists on disk. To update the config for a delete, the config needs to be picked up, modified and written back to disk before machine attempts to use it.

This fix only fixes AWS as a bigger fix is needed in general for all drivers which should live in machine itself. Modifying machine's config from Rancher is not ideal.

pkg/nodeconfig/config.go
```diff
@@ -167,6 +168,70 @@ func (m *NodeConfig) Restore() error {
...
+		return update, err
+	}
+
+	for _, file := range files {
+		if file.IsDir() {
+			configPath := filepath.Join(machines, file.Name(), "config.json")
+			b, err := ioutil.ReadFile(configPath)
+			if err != nil {
+				if os.IsNotExist(err) {
+					continue
```
alena1108: Nit: should we log the error, in debug at least?

dramich: The error indicates we don't have any machines yet which is a valid state

dramich:

Issue #24333, opened 2019-12-02
Always show certificates are about to expire if there are removed etcd hosts


xiaoluhong:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
![image](https://user-images.githubusercontent.com/20707603/69933605-3d44d980-150a-11ea-8937-76cb96e07875.png)

1. switch to global, click the ellipsis menu to the right of the cluster list, and then click Rotate Certificates > Rotate all service certificates.

  **Result:**  The UI still shows that the certificate is about to expire

2. View the cluster API certificatesExpiration field ，contains the master node that has changed, such as ip 192.168.0.68.

<img width="409" alt="image" src="https://user-images.githubusercontent.com/20707603/69932765-539d6600-1507-11ea-973f-5b495cdc0807.png">

3. Check the kube-system namespace under the business cluster system project. There is no etcd secrets corresponding to the 192.168.0.68 node. However, I only have three ETCD nodes

**Cause analysis**
Switch to the local cluster >system project >cattle-system namespace and enter the resource > secrets page，Export the secrets corresponding to the business cluster id. 

Analyzing the metadata.certs field, it was found that the certificate information of the master node that had been replaced was retained, so it was possible that the UI misjudged because of the residual certificate

**Other details that may be helpful:**

**Environment information**
- Rancher version :
`rancher/rancher:v2.3.2`
- Installation option (single install/HA):
HA
- Docker version 
18.09
- Kubernetes version 
v1.15.4

loganhz:
Reproduced in 2.3.3

Steps to reproduce:

1. Add a cluster with 3 etcd nodes
2. Remove one etcd node and add a new one 
3. Rotate certs
4. Click View API for the cluster. The cert for removed ectd node is still in API and the expirationDate not changed.

XianglongLuo:
We can find the same problem when we create  k8s cluster by rke:
Steps to reproduce:
1、Add three ALL(etcd,worker and controlplane) node and rke up,we can see three etcd cert in cluster.rkestate,one is:
![image](https://user-images.githubusercontent.com/37337250/71071588-dedd5380-21b7-11ea-9a85-b9e137ee23f1.png)
And no kubelet cert(generate serving certificate is disabled)(ref:https://rancher.com/docs/rke/latest/en/config-options/services/#kubelet):
[...]

Issue #23827, opened 2019-11-01
Etcd Certificate Rotate fails


maese83:
*Describe your issue here*
My Cluster has a soon expiring etcd certificate but I can't rotate it over the GUI. It does somthing in the back but afterwards I see still the same warning: This cluster has certs that are expiring or have expired. I have a single cluster installation. 

I tried with Kubernetes 
- v1.15.5: not working
- v1.16.2: not working

- Rancher v.2.3: not working
- Rancher v2.3.2: not working

sowmyav27:
@maese83 
I verified on 2.3.2 rancher, with k8s 1.15.
I am able to rotate certs from the UI and the certs do get rotated (get a new expiry date) on the etcd, control plane and worker nodes.

Would you be able to check if the right cert expiry is available in the API
Do a `View in API` of the cluster form the Clusters page in UI.
Check the section `certificatesExpiration` which will be displayed something like following.

```
"certificatesExpiration": {
"kube-apiserver": {
"expirationDate": "2029-10-29T16:56:27Z",
"type": "/v3/schemas/certExpiration"
},
"kube-ca": {
"expirationDate": "2029-10-29T16:49:31Z",
"type": "/v3/schemas/certExpiration"
},
"kube-controller-manager": {
"expirationDate": "2029-10-29T16:56:27Z",
"type": "/v3/schemas/certExpiration"
},
"kube-etcd-159-203-110-76": {
"expirationDate": "2029-10-29T16:56:28Z",
"type": "/v3/schemas/certExpiration"
},
"kube-node": {
"expirationDate": "2029-10-29T16:56:27Z",
"type": "/v3/schemas/certExpiration"
},
"kube-proxy": {
"expirationDate": "2029-10-29T16:56:27Z",
"type": "/v3/schemas/certExpiration"
},
"kube-scheduler": {
"expirationDate": "2029-10-29T16:56:27Z",
"type": "/v3/schemas/certExpiration"
}
},
```

maese83:
Hi @sowmyav27 
Here my expiry information. 2 etcd nodes have a 2019 expiration date, 1 has 2029.
`certificatesExpiration": {
"kube-apiserver": {
"expirationDate": "2029-10-29T13:30:54Z",
"type": "/v3/schemas/certExpiration"
},
"kube-ca": {
"expirationDate": "2028-11-08T22:05:19Z",
"type": "/v3/schemas/certExpiration"
},
"kube-controller-manager": {
[...]

PR #24548, opened 2019-12-12
Fix always show certificates are about to expire issue

XianglongLuo:
1. Update rke vendor to support deleting  unused ETCD and Kubelet certs when rotating certs

2. Support deleting  unused  certs when upgrade rancher

Related PR:https://github.com/rancher/rke/pull/1824
Issue: https://github.com/rancher/rancher/issues/24333 and https://github.com/rancher/rancher/issues/23827


pkg/controllers/user/certsexpiration/certexpiration_test.go
```diff
@@ -0,0 +1,282 @@
...
+							},
+						},
+					},
+				},
+			},
+			expectNewCerts: map[string]v3.CertExpiration{
+				"kube-node":               v3.CertExpiration{},
+				"kube-kubelet-172-17-0-4": v3.CertExpiration{},
+				"kube-kubelet-172-17-0-3": v3.CertExpiration{},
+				"kube-kubelet-172-17-0-5": v3.CertExpiration{},
```
gitlawr: Why is this kubelet cert present when `172.17.0.5` is an etcd only node?

XianglongLuo: When  generating kubelet certs,all hosts be used:
https://github.com/rancher/rke/blob/18548be050e7badc15dd31f9332aea80cc4db3ab/pki/services.go#L495

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -161,3 +164,39 @@ func getCertExpiration(c string) (v3.CertExpiration, error) {
...
+//deleteUnusedCerts removes unused certs and cleans up kubelet certs when GenerateServingCertificate is disabled
+func deleteUnusedCerts(certsExpInfo map[string]v3.CertExpiration, cluster *v3.Cluster) {
+	logrus.Info("Deleting unused certificates")
+	unusedCerts := make(map[string]bool)
+	for k := range certsExpInfo {
+		if strings.HasPrefix(k, pki.EtcdCertName) || strings.HasPrefix(k, pki.KubeletCertName) {
+			unusedCerts[k] = true
+		}
+	}
+	etcdHosts := hosts.NodesToHosts(cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.Nodes, services.ETCDRole)
```
superseb: Do we need a nil check here just in case? I only found the `cluster.Spec.RancherKubernetesEngineConfig` in `sync`. If this is caught else where, that's fine.

superseb:
Can we also bring the log message in sync, so RKE and Rancher log similar when something happens (both info and debug) as mentioned in the RKE PR. Logic looks good and it worked for me.

pkg/controllers/user/certsexpiration/certsexpiration.go
XianglongLuo: Updated.

XianglongLuo:
@superseb Thanks for your advice!Already updated，please help to re-review it

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -67,6 +72,9 @@ func (c Controller) sync(key string, cluster *v3.Cluster) (runtime.Object, error
 		}
 		certsExpInfo[certName] = info
 	}
+	logrus.Debugf("Checking and deleting unused kubelet certificates, current woker nodes are: %v", hosts.NodesToHosts(cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.Nodes, ""))
```
superseb: As this is for all nodes, we dont have to specify worker (woker)

XianglongLuo:
updated.Thanks

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -67,6 +72,9 @@ func (c Controller) sync(key string, cluster *v3.Cluster) (runtime.Object, error
 		}
 		certsExpInfo[certName] = info
 	}
+	logrus.Debugf("Checking and deleting unused kubelet certificates, current nodes are: %v", hosts.NodesToHosts(cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.Nodes, ""))
```
dramich: These log messages are going to be massive with anything other than a single node setup. We can't log all nodes in a message even under debug. If you need a debug make it specific to something useful like `deleting certs for cluster x`.

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -161,3 +169,32 @@ func getCertExpiration(c string) (v3.CertExpiration, error) {
 		ExpirationDate: date.Format(time.RFC3339),
 	}, nil
 }
+
+//deleteUnusedCerts removes unused certs and cleans up kubelet certs when GenerateServingCertificate is disabled
+func deleteUnusedCerts(certsExpInfo map[string]v3.CertExpiration, rancherKubernetesEngineConfig *v3.RancherKubernetesEngineConfig) {
+	logrus.Info("Checking and deleting unused certificates")
```
dramich: Add more info to this message, if will be too generic in a large setup, clusterID or other identification info.

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -161,3 +169,32 @@ func getCertExpiration(c string) (v3.CertExpiration, error) {
...
+		etcdName := pki.GetCrtNameForHost(host, pki.EtcdCertName)
+		delete(unusedCerts, etcdName)
+	}
+	if pki.IsKubeletGenerateServingCertificateEnabledinConfig(rancherKubernetesEngineConfig) {
+		for _, host := range allHosts {
+			kubeletName := pki.GetCrtNameForHost(host, pki.KubeletCertName)
+			delete(unusedCerts, kubeletName)
+		}
+	}
+	if unusedCerts != nil {
```
dramich: This check doesn't do anything as you made the map above. If there is nothing in the list the range is a no-op.

pkg/controllers/user/certsexpiration/certsexpiration.go
dramich: This can also leak security info related to nodes.

pkg/controllers/user/certsexpiration/certsexpiration.go
```diff
@@ -67,6 +72,9 @@ func (c Controller) sync(key string, cluster *v3.Cluster) (runtime.Object, error
 		}
 		certsExpInfo[certName] = info
 	}
+	logrus.Debugf("Checking and deleting unused kubelet certificates, current nodes are: %v", hosts.NodesToHosts(cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.Nodes, ""))
+	logrus.Debugf("Checking and deleting unused etcd certificates, current etcd nodes are: %v", hosts.NodesToHosts(cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.Nodes, services.ETCDRole))
```
dramich: Same as above comment.

gitlawr:
Updated

PR #24504, opened 2019-12-10
Exclude unrelated namespaces when validate namespace's quota

Jason-ZW:
## Problem (#24479)
The `data["projectId"]` value is empty when updates namespace with `ResourceQuota`, this cause `validateResourceQuota ` function aggregate some unexpected namespaces for validate.
## Solution
Change `data["projectId"]` to `projectID` variable.

## Validation Test PR

https://github.com/rancher/rancher/pull/24550

Issue #24479, opened 2019-12-09
If free-state namespace has quotas, quota checks will be affected when updating other namespaces that already belong to the project.


rootwuj:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Add project p01 and set quota: 
	* Resource Type: Config Maps (can be any resource type)
	* Project Limit: 20
	* Namespace Default Limit: 10
2. Add namespace ns1 under p01
3. Add project p02 and set quota: 
	* Resource Type: Config Maps 
	* Project Limit: 15
	* Namespace Default Limit: 15
4. Add namespace ns2 under p02
5. Move namespace ns2 to `none`
6. Edit the quota of namespace ns1 from 10 to 11

**Result:**
Edit failed. The error message is: 
```
Validation failed in API: resourceQuota exceeds projectLimit on fields: configMaps=26
```
The quota in the free-state is also counted, and the total exceeds the quota limit of the current project.

**Other details that may be helpful:**
The same problem exists with free-state namespace after re-import the cluster.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
Rancher: v2.3.3

soumyalj:
Reproduced the issue with v2.3.3
```
rancher.ApiError: (ApiError(...), "MaxLimitExceeded : exceeds projectLimit on fields: limitsCpu=41m\n\t{'baseType': 'error', 'code': 'MaxLimitExceeded', 'fieldName': 'resourceQuota', 'message': 'exceeds projectLimit on fields: limitsCpu=41m', 'status': 422, 'type': 'error'}")

```
Tested with master-head. 
1. Create a project P1 and set quota:
Project Limit: 20
Namespace Default Limit: 10
Create namespace ns1 under P1
2. Create project P2 and set quota:
Project Limit: 15
Namespace Default Limit: 15
Create namespace ns2 under P2
3. Move namespace ns2 to none

Edit the quota of namespace ns1 from 10 to 15. Editing is successful without any errors

PR #24550, opened 2019-12-12
Add validation test for namespace's quota

GGGitBoy:
Related Pr: https://github.com/rancher/rancher/pull/24504

Related Issue: https://github.com/rancher/rancher/issues/24479


loganhz:
Can you move it into test_project_quota.py ? I think it's better to be there

Issue #23658, opened 2019-10-24
Changes to role inheritance do not affect existing grants


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Provision a simple single node custom cluster from Rancher
2. Add a user `foo` to Rancher and grant this user the Project Member role on the Default Project.
3. Edit a Resource Grant on the Project Member role and observe the change is propagated to the ClusterRole in the downstream cluster immediately.
3. Edit the Project Member role to inherit from kubernetes admin or kubernetes view, or remove the role inheritance entirely from this role and observe that there is no change in the rolebindings on the default namespace for the `foo` user who continues to be granted the edit ClusterRole.
4. Remove and re-add Project Member role from user `foo` on Default Project and then observe the altered role inheritence takes effect.

**Environment information**
- Re-created in Rancher v2.2.8 and v2.3.1

dramich:
This should be fixed by the changes in https://github.com/rancher/rancher/pull/23641

axeal:
@dramich I just tested in v2.3.2 and still see this issue:

1. Provision a simple single node custom cluster from Rancher
2. Clone the Project Member role to 'test' role (since I couldn't edit the Project Member role in v2.3.2)
3. Add a user foo to Rancher and grant this user the 'test' role on the Default Project.
4. Edit a Resource Grant on the 'test' role and observe the change is propagated to the ClusterRole in the downstream cluster immediately.
5. Edit the 'test' role to inherit from kubernetes admin or kubernetes view, or remove the role inheritance entirely from this role and observe that there is no change in the rolebindings on the default namespace for the foo user who continues to be granted the edit ClusterRole.
6. Remove and re-add 'test' role from user foo on Default Project and then observe the altered role inheritance takes effect.

bmdepesa:
rancher/rancher:v2.3.3-rc4

Reproduced this in `v2.3.3-rc4` using the exact steps above:
```
1. Provision a simple single node custom cluster from Rancher
2. Clone the Project Member role to 'test' role (since I couldn't edit the Project Member role in v2.3.2)
3. Add a user foo to Rancher and grant this user the 'test' role on the Default Project.
4. Edit a Resource Grant on the 'test' role and observe the change is propagated to the ClusterRole in the downstream cluster immediately.
[...]

PR #24560, opened 2019-12-12
Update crb on roleTemplate inherited roles update

mrajashree:
https://github.com/rancher/rancher/issues/23658

ClusterRoleBindings are created for the inherited roles of a roleTemplate.
This commit updates them when the inherited roles on a roleTemplate are changed during the roleTemplate sync

### DIFF
--- a/pkg/controllers/user/rbac/roletemplate_handler.go
+++ b/pkg/controllers/user/rbac/roletemplate_handler.go
@@ -1,10 +1,17 @@
 package rbac
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/pkg/errors"
+	"github.com/rancher/norman/types/slice"
+	pkgrbac "github.com/rancher/rancher/pkg/rbac"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/user/rbac/roletemplate_handler.go
```diff
@@ -46,7 +53,7 @@ func (c *rtLifecycle) Updated(obj *v3.RoleTemplate) (runtime.Object, error) {
 		return nil, nil
 	}
 
-	err = c.syncRT(obj, hasPRTBs)
+	err = c.syncRT(obj, hasPRTBs, prtbs)
```
prachidamle: What is there are no prtbs, but there are crtbs present? can that be the case? hasPRTBs will be true and prtbs[] will be empty in that case

prachidamle: I guess we need to also pass the crtbs to syncRT?

pkg/controllers/user/rbac/roletemplate_handler.go
```diff
@@ -73,16 +81,109 @@ func (c *rtLifecycle) syncRT(template *v3.RoleTemplate, usedInProjects bool) err
...
+					delete(rolesToKeep, crb.RoleRef.Name)
+				}
+			}
+			for role := range rolesToKeep {
+				subject, err := pkgrbac.BuildSubjectFromRTB(prtb)
+				if err != nil {
+					return err
+				}
+				_, err = c.m.clusterRoleBindings.Create(&rbacv1.ClusterRoleBinding{
+					ObjectMeta: metav1.ObjectMeta{
```
prachidamle: Is there a way to avoid the clusterRoleBindings.Create incase the crb already exists - if the crb is present in "crbs" maybe we can skip the create?

pkg/controllers/user/rbac/roletemplate_handler.go
mrajashree: The prtb_handler sync wasn't getting called because the change was on roletemplate. But I have updated the pr with logic to pass crtbs too so the clusterrolebindings created for crtbs are updated

Issue #24294, opened 2019-11-27
OwnerReference APIVersion Invalid 


brendarearden:
OwnerReference APIVersion is hard coded to "v3" and needs to be changed to correct APIVersion
 
https://github.com/rancher/rancher/blob/6b41cb593ecb847a6771f6d9f6abf33e4f521678/pkg/controllers/management/globaldns/globaldns_handler.go#L143

https://github.com/rancher/rancher/blob/6b41cb593ecb847a6771f6d9f6abf33e4f521678/pkg/controllers/management/globaldns/globaldnsprovider_catalog_launcher.go#L220

jiaqiluo:
In v2.3.3
- add a `Global DNS Provider`
-check the object from kubetctl 
```
> kubectl get globaldnsproviders.management.cattle.io  test1 -n cattle-global-data  -o yaml
apiVersion: management.cattle.io/v3
```
- add a `Global DNS entry` 
-check the object from kubetctl 
```
> kubectl get globaldnses.management.cattle.io gd-jl89r -n cattle-global-data -o yaml
apiVersion: management.cattle.io/v3
```

jiaqiluo:
The bug fix is validated in v2.3-head 9c1ee616e

the steps are the same as my last comment provides, now we can see that the `metadata.ownerReferences.apiVersion` is set to  `management.cattle.io/v3`

```
➜ kubectl get ingress globaldns-ingress-gd-wbqtk -n cattle-global-data -o yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    external-dns.alpha.kubernetes.io/ttl: "300"
    field.cattle.io/publicEndpoints: '[{"addresses":["54.183.228.38","54.215.222.143"],"port":80,"protocol":"HTTP","serviceName":"cattle-global-data:http-svc-dummy","ingressName":"cattle-global-data:globaldns-ingress-gd-wbqtk","hostname":"test1.test.com","allNodes":false}]'
    kubernetes.io/ingress.class: rancher-external-dns
  creationTimestamp: "2019-12-19T22:42:57Z"
  generation: 1
  name: globaldns-ingress-gd-wbqtk
  namespace: cattle-global-data
  ownerReferences:
  - apiVersion: management.cattle.io/v3   <------- This line
    controller: true
    kind: GlobalDns
    name: gd-wbqtk
    uid: 15e455c0-c53a-462f-8fe6-d46fa0e01c1e
  resourceVersion: "64857"
[...]

PR #24585, opened 2019-12-13
[v2.3] Add api group name to apiversion

brendarearden:
**Problem**
APIVersions for `github.com/rancher/types/apis/management.cattle.io/v3` were being listed at `APIVersion: "v3"`. It is common practice to list `<GROUP_NAME>/<VERSION>` as the value for APIVersion. 

**Solution**
Update APIVersion to include group name in the version. Example `APIVersion: "v3"` >> `APIVersion: "management.cattle.io/v3`

**Issue**
#24294

### DIFF
--- a/pkg/controllers/management/globaldns/globaldns_handler.go
+++ b/pkg/controllers/management/globaldns/globaldns_handler.go
@@ -148,7 +148,7 @@ func (n *GDController) generateNewIngressSpec(globaldns *v3.GlobalDNS) *v1beta1.
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					Name:       globaldns.Name,
-					APIVersion: "v3",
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/globaldns/globaldns_handler.go
```diff
@@ -148,7 +148,7 @@ func (n *GDController) generateNewIngressSpec(globaldns *v3.GlobalDNS) *v1beta1.
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					Name:       globaldns.Name,
-					APIVersion: "v3",
+					APIVersion: "management.cattle.io/v3",
```
dramich: Can this be pulled off the object instead of hard coding it?

dramich:
LGTM

Issue #24421, opened 2019-12-05
runSecurityScans is missing RBAC


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
As a cluster member, do a view in API on the cluster, and click on `runSecurityScan` from the actions listed on the page.

**Expected Result:**
A cluster member should NOT be able to perform `runSecurityScan` action on the cluster

**Actual Result:**
A cluster member is able to perform `runSecurityScan` action on the cluster

**Note:** 
But the user is NOT able to view the scan reports.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - latest
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Kubernetes version (use `kubectl version`):

```
1.16
```

sowmyav27:
A project Owner and Project member are able to run security scan through the API.

With RBAC implemented, project Owner and Project member should NOT be able to run security scan through the API.

@leodotcloud We expected these use cases to be fix with the PR. Please confirm.

sowmyav27:
Expected RBAC:

```
Cluster Owner - Can run scans and list scans
Cluster Member -Should NOT be able to run scans but can list scans
Project Owner - Should NOT be able to run scans and Should NOT be able to list scans
Project Member -  Should NOT be able to run scans and Should NOT be able to list scans
Project ReadOnly - Should NOT be able to run scans and Should NOT be able to list scans
```

sowmyav27:
**Verified on master head - commit id: `48a2c4ffc`**

- The following users: cluster member/project member/project owner and user with Read only permissions(project) are not able to run security scans via the API. The following error is seen:

- API throws error: `{"baseType":"error","code":"PermissionDenied","message":"can not run security scan","status":403,"type":"error"}`

- API UI, the `runSecurityScan` option is disabled.

- Validation from Rancher UI is blocked because of this issue - https://github.com/rancher/rancher/issues/24644

sowmyav27:
Verified on master head - commit id: `48a2c4ffc`

- As a Cluster Owner, user is not able to run Scans via the API.
- In the API UI runSecurityScans is disabled.
- API throws error: {"baseType":"error","code":"PermissionDenied","message":"can not run security scan","status":403,"type":"error"}

[...]

PR #24586, opened 2019-12-13
add rbac for runSecurityScan action

leodotcloud:
https://github.com/rancher/rancher/issues/24421

RBAC changes:
- Remove "Manage Cluster Scan" role
- Allow only cluster owner to run the scan

Skip Changes:
- rename the field from skip to overrideSkip.
Here is how the logic works:
- There are two ways to pass skip information. One using the configmap security-scan-cfg and the other from the API using the field overrideSkip .
- The skip information is conveyed to the security-scan chart using a configmap name ... which gets mounted at /etc/kbs/config.json. 
- When security-scan-cfg  is mounted, the benchmark version is specified. But when overrideSkip is passed, instead of using the benchmark version as the key, "current" is used, as we don't have the version information in Rancher.


pkg/api/customization/cluster/formatter.go
```diff
@@ -47,7 +47,23 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 			resource.AddAction(request, v3.ClusterActionBackupEtcd)
 			resource.AddAction(request, v3.ClusterActionRestoreFromEtcdBackup)
 		}
-		resource.AddAction(request, v3.ClusterActionRunSecurityScan)
+		canRunScan := func() bool {
+			revision := map[string]interface{}{
+				"id": request.ID,
+			}
```
prachidamle: Is this the Cluster's Id? maybe just checking if "update" on cluster resource is allowed should suffice here

pkg/api/customization/cluster/formatter.go
```diff
@@ -47,7 +47,23 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+
+			return request.AccessControl.CanDo(
+				v3.ClusterScanGroupVersionKind.Group,
+				v3.ClusterScanResource.Name,
+				"create",
+				request,
+				revision,
+				request.Schema,
+			) == nil
```
prachidamle: I am not sure if we can check "create" ability using CanDo. That method sound like more for checking if an action/verb can be applied to an existing object as pointed by the obj map[string]interface{} parameter.

prachidamle: Maybe we need to do something like https://github.com/rancher/rancher/blob/master/pkg/api/customization/cluster/actionhandler.go#L48-L53

prachidamle:
@leodotcloud The ticket says "A cluster member should NOT be able to perform runSecurityScan action on the cluster"

But can a cluster member having "Manage Cluster Scans" role perform this action (on the cluster they are member of)?

pkg/api/customization/cluster/formatter.go
leodotcloud: modified the logic, we don't need the id.

pkg/api/customization/cluster/formatter.go
leodotcloud: As discussed offline, runSecurityScan behind the scenes creates a clusterscan object. Hence checking for "create" action.

leodotcloud:
Also, based on some offline discussions, we got rid of "Manage Cluster Scans" role too.

pkg/api/customization/cluster/formatter.go
```diff
@@ -47,7 +47,19 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+				v3.ClusterScanResource.Name,
+				"create",
+				request,
+				nil,
+				request.Schema,
+			) == nil
+		}
+		if canCreateClusterScanObject() {
+			resource.AddAction(request, v3.ClusterActionRunSecurityScan)
+		}
```
prachidamle: I think we should add this after checking if the user can update the given cluster - so only cluster-owner of *this* cluster should have this action enabled when the cluster is viewed in API - and not any other cluster-owner(who possibly has the clusterId of other cluster)

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -25,9 +23,22 @@ const (
...
+			"create",
+			apiContext,
+			nil,
+			apiContext.Schema,
+		) == nil
+	}
+	if !canCreateClusterScanObject() {
+		return httperror.NewAPIError(httperror.PermissionDenied, "can not run security scan")
+	}
+
```
prachidamle: We should also check for ability to update cluster since this action will also update the cluster object.
example: https://github.com/rancher/rancher/blob/393dadee4196a2ccbe31bde711df6a50f29a32ca/pkg/api/customization/cluster/actionhandler.go#L97

pkg/api/customization/cluster/formatter.go
leodotcloud: Good point. Will change.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: Will do

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -75,35 +87,42 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
...
+						logrus.Infof("cisScanHandler: Create: created skip override configmap %v with contents: %v", cm.Name, string(skipDataBytes))
+						success = true
+						break
+					}
+					time.Sleep(RetryIntervalInMilliseconds * time.Millisecond)
 				}
-				if err != nil {
-					logrus.Errorf("cisScanHandler: Create: (try: %v) error creating configmap: %v", i+1, err)
+				if !success {
+					logrus.Errorf("cisScanHandler: Create: error creating configmap: %v", err)
```
prachidamle: If !success, do you also need to make cm instance nil so that it is not set here https://github.com/rancher/rancher/blob/393dadee4196a2ccbe31bde711df6a50f29a32ca/pkg/controllers/user/cis/clusterScanHandler.go#L135 ?

leodotcloud: I don't think it should be an issue coz if Create fails, it should set cm to nil, but anyways I can set it to nil explicitly.

Issue #24446, opened 2019-12-06
[flakey test] test_invalid_catalog_charts


dramich:
https://drone-publish.rancher.io/rancher/rancher/1835/2/2

PR #24593, opened 2019-12-13
[flakey test][v2.3] Ensure error state

brendarearden:
**Problem** 
Test was occasionally getting information from chart when it was retrying and not in error state

**Solution**
Ensure the catalog is in refreshed error and not refreshed syncing when assertions on ran

**Issue**
#24446

### DIFF
--- a/tests/integration/suite/test_catalog.py
+++ b/tests/integration/suite/test_catalog.py
@@ -1,6 +1,7 @@
 import pytest
 import time
 from rancher import ApiError
+from .conftest import wait_for
 
 from .common import wait_for_template_to_be_created, \
     wait_for_template_to_be_deleted, random_str, wait_for_atleast_workload
@@ -291,7 +292,6 @@ def test_invalid_catalog_charts(admin_mc, remove_resource):
     client = admin_mc.client
     name = random_str()
     url = "https://github.com/rancher/integration-test-charts"
-
     catalog = client.create_catalog(name=name,
                                     branch="broke-charts",
                                     url=url,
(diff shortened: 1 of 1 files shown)


tests/integration/suite/test_catalog.py
```diff
@@ -298,19 +298,33 @@ def test_invalid_catalog_charts(admin_mc, remove_resource):
...
     assert catalog.transitioningMessage == "Error syncing catalog " + name
     # this will break if github repo changes
     assert len(templates) == 6
     # checking that the errored catalog can be deleted successfully
     client.delete(catalog)
     wait_for_template_to_be_deleted(client, name)
     assert not client.list_catalog(name=name).data
+
+
+def get_errored_catalog(client, catalog, timeout=45):
```
dramich: Use wait_for, this duplicates a lot of logic that already exists in there.

tests/integration/suite/test_catalog.py
```diff
@@ -292,21 +293,25 @@ def test_invalid_catalog_charts(admin_mc, remove_resource):
...
-    catalog = client.create_catalog(name=name,
-                                    branch="broke-charts",
-                                    url=url,
-                                    )
-    remove_resource(catalog)
-    wait_for_template_to_be_created(client, catalog.id)
-
-    catalog = client.reload(catalog)
+    def get_errored_catalog():
+        catalog = client.create_catalog(name=name,
```
dramich: This is going to keep trying to create the catalog every time wait_for calls this function, which I don't think is intended. Create the catalog outside this check then do the refresh/check in here

tests/integration/suite/test_catalog.py
```diff
@@ -291,22 +292,26 @@ def test_invalid_catalog_charts(admin_mc, remove_resource):
...
     catalog = client.create_catalog(name=name,
                                     branch="broke-charts",
                                     url=url,
                                     )
     remove_resource(catalog)
     wait_for_template_to_be_created(client, catalog.id)
 
-    catalog = client.reload(catalog)
+    def get_errored_catalog(catalog):
+        while catalog.transitioning != "error":
```
dramich: If catalog never goes to error this will never exit. This should refresh the catalog and return a bool/catalog based on the check of `catalog.transitioning`

PR #24594, opened 2019-12-13
Downstream Agent Image Tracking

luthermonson:
Adds a map of downstream cluster agents to check to keep them in sync. Especially important during upgrade/rollback process.


pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -209,6 +227,65 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
...
+		if cluster.Status.AgentImageOverride == cluster.Spec.DesiredAgentImage {
+			return nil
+		}
+		if cluster.Status.AgentImageOverride == cluster.Status.AgentImage {
+			return nil
+		}
+		clearAgentCaches(cluster.UID)
+		cluster.Status.AgentImage = ""
+		cluster.Spec.DesiredAgentImage = cluster.Status.AgentImageOverride
+		return fmt.Errorf("Rancher Agent Override specificied, setting desired image: %s",
```
kinarashah: why do we return err here and L283 if we're already handling it?

luthermonson: forces the process to save the cluster and get us back into the redeploy code (which is above this code) on the second call to doSync

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -20,14 +21,25 @@ import (
...
+var (
+	nodeAgentImages    map[k8stypes.UID]string
+	nodeMutex          sync.RWMutex
+	clusterAgentImages map[k8stypes.UID]string
+	clusterMutex       sync.RWMutex
+)
+
 func Register(ctx context.Context, management *config.ManagementContext, clusterManager *clustermanager.Manager) {
+	nodeAgentImages = make(map[k8stypes.UID]string)
+	clusterAgentImages = make(map[k8stypes.UID]string)
```
dramich: Instantiate the vars in the declaration, if someone calls register twice this will overwrite

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -20,14 +21,25 @@ import (
...
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
 const AgentForceDeployAnn = "io.cattle.agent.force.deploy"
 
+var (
+	nodeAgentImages    map[k8stypes.UID]string
```
dramich: This should be one map and one lock, the key should be cluster.ID, this will be much easier to debug if needed vs a UID. 
Use a type `map[string]map[string]string` > clusterID['nodeAgentImage']foo

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -20,14 +21,25 @@ import (
...
+	nodeMutex          sync.RWMutex
```
dramich: The Read part of the lock isn't being used, if not needed update to a standard Mutex

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -101,11 +113,17 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
 	if err != nil {
 		return err
 	}
+
 	err = cd.deployAgent(cluster)
 	if err != nil {
 		return err
 	}
 
+	err = cd.validateAgent(cluster)
```
dramich: Could this be called before deployagent then the correct one would be deployed in this run? I don't follow returning an error when it really isn't an error. If someone sets the override the controller should respect that and not throw an error in the logs.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -209,6 +227,65 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
...
+	if cluster.Status.AgentImage == "" {
+		//if empty we're waiting for something to deploy, valid until proven otherwise.
+		return nil
+	}
+
+	//check if the agent downstream is the same as the one locally in the crd
+	var na, ca string
+	var ok bool
+	var err error
+	if na, ok = nodeAgentImages[cluster.UID]; !ok {
```
dramich: Access the map requires the lock

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -209,6 +227,65 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
 	return err
 }
 
+func (cd *clusterDeploy) validateAgent(cluster *v3.Cluster) error {
```
dramich: The flow of this should be updated:
Get the cluster with the read lock
Do all validation on the cluster
Assuming no errors, grab a write lock and update the local cache
Currently this grabbing and dropping locks then would grab the lock again to delete from the cache.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -257,3 +334,51 @@ func (cd *clusterDeploy) getYAML(cluster *v3.Cluster, agentImage, authImage stri
...
+
+	if len(ds.Spec.Template.Spec.Containers) != 1 {
+		return "", fmt.Errorf("Expected one container in cattle-node-agent DaemonSet, found %d",
+			len(ds.Spec.Template.Spec.Containers))
+	}
+
+	return ds.Spec.Template.Spec.Containers[0].Image, nil
+}
+
+func clearAgentCaches(uid k8stypes.UID) {
```
dramich: Update to use defer once there is a single lock.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
dramich: This should also be integrated into the `redeployAgent` func.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
dramich: This isn't an error and should not be handled like one.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -209,6 +227,65 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
...
+		clearAgentCaches(cluster.UID)
+		cluster.Status.AgentImage = ""
+		cluster.Spec.DesiredAgentImage = cluster.Status.AgentImageOverride
+		return fmt.Errorf("Rancher Agent Override specificied, setting desired image: %s",
+			cluster.Status.AgentImageOverride)
+	}
+
+	if cluster.Status.AgentImage != na || cluster.Status.AgentImage != ca {
+		//downstream doesn't match and we should refresh
+		err = fmt.Errorf("Local agent image does not match downstream: %s locally, %s agent, %s cluster",
```
dramich: Same as above, this isn't an error

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -110,6 +129,22 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
...
+		}
+		if cluster.Status.AgentImageOverride == cluster.Status.AgentImage {
+			return false
+		}
+
+		//kick a redeploy with the overriden agent
+		clearAgentImages(cluster.Name)
+		cluster.Status.AgentImage = ""
+		cluster.Spec.DesiredAgentImage = cluster.Status.AgentImageOverride
+		return false
```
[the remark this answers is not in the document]
luthermonson: youre right but it was working and i think only because it saved and came back through on another cycle. ill change both of these to return true.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -130,10 +165,19 @@ func redeployAgent(cluster *v3.Cluster, desiredAgent, desiredAuth string) bool {
...
+			" private repo changed=%v", cluster.Name, forceDeploy, imageChange, repoChange)
 		return true
 	}
+
+	na, ca := getAgentImages(cluster.Name)
+	if cluster.Status.AgentImage != na || cluster.Status.AgentImage != ca {
+		//downstream agent does not match, kick a redeploy with settings agent
+		clearAgentImages(cluster.Name)
+		cluster.Status.AgentImage = ""
+		cluster.Spec.DesiredAgentImage = ""
```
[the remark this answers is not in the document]
luthermonson: youre right but it was working and i think only because it saved and came back through on another cycle. ill change both of these to return true.

PR #24595, opened 2019-12-13
[flakey test][forwardport] Ensure error state

brendarearden:
**Problem**
Test was occasionally getting information from chart when it was retrying and not in error state

**Solution**
Ensure the catalog is in refreshed error and not refreshed syncing when assertions on ran

**Issue**
#24446


brendarearden:
https://github.com/rancher/rancher/pull/24785  merging forwardport due to this approval

PR #24598, opened 2019-12-14
[forwardport] Agent tracking master

luthermonson:
Adds a map of downstream cluster agents to check to keep them in sync. Especially important during upgrade/rollback process.


brendarearden:
https://github.com/rancher/types/pull/1057 was merged in - Vendor changes from types to get changes

Issue #20159, opened 2019-05-10
Security, Authentication, LDAP:  rancher server error on principal search (groups, users)


donatasnicequestion:
**What kind of request is this (question/bug/enhancement/feature request):**
BUG

**Steps to reproduce (least amount of steps as possible):**
Configure Security, Provider Open LDAP 
Enter information  requested (the same like many other tools and systems)
Test-authentication: success. 

Afterwards

Select : "Allow members of Clusters, Projects, plus Authorized Users and Organizations" radio.
Start entering some user name into "Authorized Users and Organizations" input field

**Result:**
No users or groups available  -therefore  LDAP authorization ist not usable.
It is not possible to select users or groups and to assign them as a members of cluster.

Panic Errors in rancher server Logs:

```
2019/05/10 16:20:04 [ERROR] Panic serving api request:
goroutine 52240 [running]:
runtime/debug.Stack(0xc005c7a930, 0x33f6e40, 0x6a97ee0)
        /usr/local/go/src/runtime/debug/stack.go:24 +0xa7
github.com/rancher/rancher/vendor/github.com/rancher/norman/api.(*Server).ServeHTTP.func1(0x41d6e80, 0xc00739dcf0)
        /go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/api/server.go:175 +0x67
panic(0x33f6e40, 0x6a97ee0)
        /usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/rancher/rancher/pkg/auth/providers.SearchPrincipals(0x68fb7d6, 0x1, 0x0, 0x0, 0xc0116e47aa, 0x5, 0xc01139cfa0, 0x17, 0xc0116e47d0, 0xb, ...)
        /go/src/github.com/rancher/rancher/pkg/auth/providers/providers.go:137 +0x74
github.com/rancher/rancher/pkg/auth/principals.(*principalsHandler).actions(0xc0028148c0, 0xc009fa78ab, 0x6, 0xc00555e180, 0xc0059370e0, 0x0, 0x0)
        /go/src/github.com/rancher/rancher/pkg/auth/principals/handler.go:56 +0x3f7
github.com/rancher/rancher/pkg/auth/principals.(*principalsHandler).actions-fm(0xc009fa78ab, 0x6, 0xc00555e180, 0xc0059370e0, 0x0, 0x0)
        /go/src/github.com/rancher/rancher/pkg/auth/principals/schema.go:16 +0x52
[...]

PR #24605, opened 2019-12-16
Check token in SearhPrincipals to avoid panic

mrajashree:
https://github.com/rancher/rancher/issues/20159
The panic will occur only if the authProvider field on token is empty or the provider isn't configured.
Not sure why either of this would happen. But this commit will avoid panic and error will indicate what was causing it.


dramich:
LGTM

Issue #24334, opened 2019-12-02
Nested groups info sometimes lost with Google OAuth


yannrouillard:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

 * Configure Google Oauth with nested groups,
 * Create a user `UserA` and two groups  `GroupParent` and `GroupChild` in Google,
 * Add `UserA` as member of `GroupChild`, and add `GroupChild` as member of `GroupParent`
 * Authorize only `groupParent` to access a cluster (or a project)  
 * Log in with `UserA`

**Result:**

`User` can't access the cluster or project although he should because he belongs indirectly to `GroupParent`
If he log-out and re-log, he will then be able to properly see the cluster and the project.

The issue will happen at a subsequent login if there is enough delay with the previous login.

**Other details that may be helpful:**

I will post more details about the logs but basically I can clearly see that only the direct groups (`groupChild` in the example) is detected at first login and then the parent groups are detected by Rancher at the next logout/login.

**Environment information**
- Rancher version: 2.3.2
- Installation option: single install

yannrouillard:
Here is an excerpt of the logs when the problem happens with some info masked.
There was a log of logs so I chose to select the one that seemed relevant:

```
2019/12/02 05:15:38 [DEBUG] [Google OAuth] loginuser: Using code to get oauth token
2019/12/02 05:15:38 [DEBUG] [Google OAuth] loginuser: Exchanged code for oauth token
2019/12/02 05:15:38 [DEBUG] [Google OAuth] loginuser: Obtained userinfo using oauth access token
[...]
2019/12/02 05:15:40 [DEBUG] [Google OAuth] loginuser: Retrieved user's groups using admin directory
2019/12/02 05:15:40 [DEBUG] [Google OAuth] loginuser: Checking user's access to Rancher
2019/12/02 05:15:40 [DEBUG] [Google OAuth] loginuser: Returning principals and marshaled oauth token
[...]
2019/12/02 05:15:40 [DEBUG] Requesting user: u-fakeuserid, Requesting groups: [system:authenticated] Impersonate user:  Impersonate group: []
2019/12/02 05:15:40 [DEBUG] Triggering auth refresh on u-fakeuserid
[...]
2019/12/02 05:15:40 [DEBUG] Starting refresh process for u-fakeuserid
[...]

PR #24606, opened 2019-12-16
Preserve nested membership on refetch principals

mrajashree:
https://github.com/rancher/rancher/issues/24334
The refetch groups api for google auth wasn't checking if the nested
group membership option was enabled. This commit adds that check
to ensure nested group membership is preserved on refetch api


dramich:
LGTM

Issue #24190, opened 2019-11-21
Rancher Machine Rename


luthermonson:
Remove shim machine-pacakge repo and rename docker-machine to rancher-machine.

luthermonson:
machine
https://github.com/rancher/machine/pull/51
https://github.com/rancher/machine/pull/52
https://github.com/rancher/machine/pull/53

and 

rancher
https://github.com/rancher/rancher/pull/24161

sangeethah:
@izaac  As part of validating this issue , we need to make sure that all driver machine works as expected when provisioning nodes from all the supported node drivers:

Amazon EC2
Azure
Digital Ocean
Linode
Vsphere

izaac:
Could this other GKE issue be related to this renaming? https://github.com/rancher/rancher/issues/24218 @luthermonson

luthermonson:
@izaac not related and i do see you closed out that other issue. just thought i'd respond :)

izaac:
Tested with Rancher version "2.4 master-head (11/25/2019) addeb8399"

Clusters created (and also removed) for:

- Amazon EC2
- Azure
- Digital Ocean
- Linode
- vSphere

vSphere cluster:
![Screen Shot 2019-11-25 at 11 37 07 AM](https://user-images.githubusercontent.com/27659/69568182-5d0f5700-0f78-11ea-8e9f-82a1f5ba91c7.png)

The others:
![Screen Shot 2019-11-25 at 11 37 17 AM](https://user-images.githubusercontent.com/27659/69568199-64366500-0f78-11ea-8776-a9c7ccf5daee.png)

@luthermonson as per our conversation, please check if this coverage is enough to QA all the changes made in the renaming. If this is good we can close the issue.

luthermonson:
@izaac legit, this makes me happy to see. thanks!

luthermonson:
@izaac back ported this issue, will put it back to test when it's merged: https://github.com/rancher/rancher/pull/24618

luthermonson:
Version bump and rancher-machine rename in 2.3 now. Please run cluster creation automation against release/v2.3 and let's see how this goes.

izaac:
Tested with Rancher version `2.3-head` d73dff8df

Clusters created (and also removed) for:

- Amazon EC2 (Automation)
- Azure (Automation)
- Digital Ocean (Automation/Manual)
- Linode (Manual)
- vSphere (Manual)

[...]

PR #24618, opened 2019-12-16
[backport] Rancher Machine Rename

luthermonson:
https://github.com/rancher/rancher/issues/24190


go.mod
```diff
@@ -88,7 +86,7 @@ require (
 	github.com/sirupsen/logrus v1.4.2
 	github.com/stretchr/testify v1.4.0
 	github.com/tomnomnom/linkheader v0.0.0-20180905144013-02ca5825eb80
-	github.com/urfave/cli v1.21.0
```
daxmc99: 👀  Might need to pin this version if its a used by rancher machine

cjellick:
LGTM

go.mod
luthermonson: its the rancher dependency. this just gets it current to what is in master already.

PR #24634, opened 2019-12-17
[Forward port] Preserve nested membership on refetch principals

mrajashree:
Forward port for https://github.com/rancher/rancher/pull/24606

The refetch groups api for google auth wasn't checking if the nested
group membership option was enabled. This membership adds that check
to ensure nested group membership is preserved on refetch api


dramich:
LGTM

PR #24651, opened 2019-12-17
generate correct registry auth for private registry

kinarashah:
Issues: 
- incorrect value of registry auth is passed so worker nodes can't create process 
- nodes created via docker-machine cannot use private registry credentials, so new node creation fails 

PR fixes: 
- registry auth creation 
- deployAgent should only use System default registry to append to agent-image 

https://github.com/rancher/rancher/issues/20029

### DIFF
--- a/pkg/cluster/private_registry.go
+++ b/pkg/cluster/private_registry.go
@@ -4,9 +4,9 @@ import (
 	"encoding/base64"
 	"encoding/json"
 
+	"github.com/docker/docker/api/types"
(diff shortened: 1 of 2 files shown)


pkg/cluster/private_registry.go
```diff
@@ -4,9 +4,10 @@ import (
 	"encoding/base64"
 	"encoding/json"
 
+	"github.com/docker/docker/api/types"
+
```
luthermonson: probably dont need the space?

Issue #23848, opened 2019-11-02
Monitoring - app and namespace are left when deleting the project


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- run Rancher:v2.3-head
- add a cluster
- add a project
- enable the project monitoring 
- delete the project

**Result:**
- the project monitoring app and its namespace are left

jiaqiluo:
The bug fix is validated in master-head ff24cb9e3

Steps
- run Rancher:v2.3-head
- add a cluster
- add a project
- enable the project monitoring 
- delete the project

Result:
- the project monitoring app and its namespace are deleted along with the project

PR #24658, opened 2019-12-18
Fix the problem Monitoring - app and namespace are left when deleting the project

GGGitBoy:
Problem:

Monitoring - app and namespace are left when deleting the project

Solution:

If the namespace's creatorID is null, it wouldn't remove the namespace when deleting the project.
https://github.com/rancher/rancher/blame/fa26eccdb1ec6d9065877340f6b90ef374969adf/pkg/controllers/user/rbac/project_handler.go#L66
so add parameter creatorID for function EnsureAppProjectName can fix it.

Related Issue: 

https://github.com/rancher/rancher/issues/23848

### DIFF
--- a/pkg/app/utils/utils.go
+++ b/pkg/app/utils/utils.go
@@ -18,7 +18,11 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
-func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, clusterName, appTargetNamespace string) (string, error) {
+const (
+	creatorIDAnno = "field.cattle.io/creatorId"
+)
+
+func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, clusterName, appTargetNamespace, creatorID string) (string, error) {
 	// detect Namespace
 	deployNamespace, err := userNSClient.Get(appTargetNamespace, metav1.GetOptions{})
 	if err != nil && !kerrors.IsNotFound(err) {
@@ -33,7 +37,8 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 	} else {
 		deployNamespace = &corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: appTargetNamespace,
+				Name:        appTargetNamespace,
+				Annotations: map[string]string{creatorIDAnno: creatorID},
 			},
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/app/utils/utils.go
```diff
@@ -33,7 +33,8 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 	} else {
 		deployNamespace = &corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: appTargetNamespace,
+				Name:        appTargetNamespace,
+				Annotations: map[string]string{"field.cattle.io/creatorId": creatorID},
```
gitlawr: Use constant for the annotation key. Check if an existing one is appliable

gitlawr:
Does it work in upgrade scenario?

pkg/app/utils/utils.go
GGGitBoy: It works fine

GGGitBoy:
> Does it work in upgrade scenario?

The old Monitoring - namespace's creatorID is still null, namespace are left when deleting the project. 
The new Monitoring - namespace will be remove when we do the same

pkg/controllers/user/monitoring/projectHandler.go
```diff
@@ -132,7 +132,8 @@ func (ph *projectHandler) doSync(project *mgmtv3.Project, clusterName string) er
 }
 
 func (ph *projectHandler) ensureAppProjectName(appTargetNamespace string, project *mgmtv3.Project) (string, error) {
-	appProjectName, err := utils.EnsureAppProjectName(ph.app.agentNamespaceClient, project.Name, project.Spec.ClusterName, appTargetNamespace)
+	creator, err := ph.app.systemAccountManager.GetProjectSystemUser(project.Name)
```
kinarashah: handle err here

pkg/controllers/user/monitoring/clusterHandler.go
```diff
@@ -157,12 +157,13 @@ func (ch *clusterHandler) doSync(cluster *mgmtv3.Cluster) error {
 }
 
 func (ch *clusterHandler) ensureAppProjectName(clusterID, appTargetNamespace string) (string, error) {
+	creator, err := ch.app.systemAccountManager.GetSystemUser(ch.clusterName)
```
kinarashah: handle err

loganhz:
@kinarashah updated

dramich:
Can this be updated to use an owners reference? Tie the app and namespace to the project, if the project gets deleted everything goes with it

gitlawr:
An owning object must be in the same namespace as the dependent or be cluster-scoped.
And the problem here is that the app object is gone but the related namespace and resources are left in the user cluster.

Issue #24626, opened 2019-12-17
[CIS] - When a cluster scan fails to generate report, the scan is seen Active


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** 

**Steps to reproduce (least amount of steps as possible):**

- Run scan on a cluster deployed with k8s v 1.17
- The scan is seen Active

<img width="1303" alt="Screen Shot 2019-12-16 at 4 51 51 PM" src="https://user-images.githubusercontent.com/26032343/70955074-63976880-2024-11ea-8bf3-ed1dffb5832f.png">

**Expected Result**
The status of the scan should show Error.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Kubernetes version (use `kubectl version`): 

```
1.17
```

leodotcloud:
Earlier, the report was not generated as there was an exception in the summarizer. That part has been taken care of now. Outcome of the offline discussion with Sowmya was to add a timeout on top of that. The test case would be to delete the pod in the middle of the scan.

leodotcloud:
There could be other failure scenarios too. For example, the pod could get stuck for some reason. We need a comprehensive healthcheck to figure out the actual state of the health. That's a bigger enhancement.

leodotcloud:
As part of improving the error handling, this issue needed to be addressed: https://github.com/aquasecurity/kube-bench/issues/546. This got fixed in https://github.com/aquasecurity/kube-bench/releases/tag/v0.2.3 by PR: https://github.com/aquasecurity/kube-bench/pull/548

leodotcloud:
@codyrancher @maggieliu once this issue gets resolved, the UI needs to be able to show the error message (if present) from the condition.

Example from the API ClusterScan object:

```
{
"lastUpdateTime": "2020-02-06T17:01:54-08:00",
"message": "error running sonobuoy",
"status": "True",
"type": "Failed"
},
```

sowmyav27:
**Verified on rancher - master-head - commit id: `0f1c4b17d`**

- If a report fails to generate, CIS Scan is seen with `Error` status on the Rancher UI
[...]

Issue #24625, opened 2019-12-17
RBAC - Cluster owner is NOT able to list the cluster scan runs/reports


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):**  bug

**Steps to reproduce (least amount of steps as possible):**

1. As an admin, deploy a cluster.
2. Add another user 1 as the cluster owner for this cluster.
3. Login as user 1
4. Go to Tools --> CIS Scans and click on Run scans.

**Expected results:**
User should be able to perform the action.
User is NOT able to list the cluster scans in API/UI

**Actual Results:**
User is able to perform the action.
User is should be able to list the cluster scans in API/UI

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Kubernetes version (use `kubectl version`):

```
1.15
```

sowmyav27:
Expected RBAC:

```
Cluster Owner - Can run scans and list scans
Cluster Member -Should NOT be able to run scans but can list scans
Project Owner - Should NOT be able to run scans and Should NOT be able to list scans
Project Member -  Should NOT be able to run scans and Should NOT be able to list scans
Project ReadOnly - Should NOT be able to run scans and Should NOT be able to list scans
```

sowmyav27:
**Verified on master head - commit id: `48a2c4ffc`**

- As a Cluster Owner, user is not able to run Scans via the API.
- In the API UI `runSecurityScans` is disabled.

**Expected:** 
Cluster Owner - Can run scans and list scans

Reopening the issue.

sowmyav27:
**Verified on master head - commit id: `656781c34`**

- Verified from the UI .
- GO to Tools --> CIS Scans --> Run Scans
- As a Cluster Owner, user is able to run CIS Scans
- User lists the scans run on the CIS Scans page.
- User is able to list scans under /v3/clusterscans
- As a Cluster Owner, user is able to run Scans via the API.

Issue #24422, opened 2019-12-05
Support for different version of scan tool


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
Rancher must support different versions of scan tool

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

leodotcloud:
Using the API, it's possible to override the benchmark version used. By default we use `rke-cis-1.4` and we don't have any other version for RKE available yet. But for testing purposes, we can pass `cis-1.4` which will run the upstream tests directly (which will result in lot of failures for RKE)

sowmyav27:
**Reopening this issue**

**verified on latest master-head**

**Scenario#1**

- On entering overrideBenchmarkVersion : "cis-1.4" in the `runSecurityScan` API, the default `rke-cis-1.4` is run. 
- No additional failures are seen on the cluster scan as compared to Scan run from `rke-cis-1.4` .

**Expected:**
"cis-1.4" related Scan should be run on the cluster, which should have resulted in a lot more failures as mentioned in https://github.com/rancher/rancher/issues/24422#issuecomment-565579117


**Scenario#2**
- On entering invalid value for overrideBenchmarkVersion, the default `rke-cis-1.4` is run. 

**Expected:**
Users should be presented with an error, stating the value is invalid. 


**Scenario#3**
- overrideBenchmarkVersion should show the AVAILABLE values in the dropdown in the API UI for runSecurityScans.
- This can be achieved only when there is support for overrideBenchmarkVersion as a first class object.

sowmyav27:
Scenario#3 -- As per discussions, was decided to be as is.

Scenario#1 and Scenario#2 -- need to be addressed for alpha release.

sowmyav27:
**Verified on master head - commit id: `656781c34`**

**Scenario#1**
Verified the fix

- On entering overrideBenchmarkVersion : "cis-1.4" in the `runSecurityScan` API, the default `rke-cis-1.4` is run. 
- Additional failures are seen
- Download report.
- The report shows the scan tool version run is `Version: cis-1.4`


**Scenario#2**
- On entering invalid value for overrideBenchmarkVersion, say "1.1.1.1", the scan errors out. with status: `Error`
[...]

PR #24698, opened 2019-12-19
cis fixes

leodotcloud:
Partial fix for https://github.com/rancher/rancher/issues/24626
When there is a error marked by the security-scan-runner pod, the state wasn't reflecting correctly.

Addresses: https://github.com/rancher/rancher/issues/24625 and https://github.com/rancher/rancher/issues/24421
Fixed the rbac checks

https://github.com/rancher/rancher/issues/24422
(One of my commits in previous PRs missed somehow, fixing again) The overrideBenchmarkVersion was not being passed to answers. Fixed that.

### DIFF
--- a/app/role_data.go
+++ b/app/role_data.go
@@ -25,7 +25,8 @@ var defaultAdminLabel = map[string]string{defaultAdminLabelKey: defaultAdminLabe
 func addRoles(management *config.ManagementContext) (string, error) {
 	rb := newRoleBuilder()
 
-	rb.addRole("Create Clusters", "clusters-create").addRule().apiGroups("management.cattle.io").resources("clusters").verbs("create").
+	rb.addRole("Create Clusters", "clusters-create").
+		addRule().apiGroups("management.cattle.io").resources("clusters").verbs("create").
 		addRule().apiGroups("management.cattle.io").resources("templates", "templateversions").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("nodedrivers").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("kontainerdrivers").verbs("get", "list", "watch").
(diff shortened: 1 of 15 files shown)


pkg/api/customization/cluster/formatter.go
```diff
@@ -47,23 +47,36 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+			return request.AccessControl.CanDo(
+				v3.ClusterGroupVersionKind.Group,
+				v3.ClusterResource.Name,
+				"update",
+				request,
+				cluster,
+				resource.Schema) == nil
 		}
-		canCreateClusterScanObject := func() bool {
+		canCreateClusterScanObjectFn := func(request *types.APIContext, clusterID string) bool {
```
kinarashah: why do we need this? isn't just `canUpdateClusterFn` sufficient?

prachidamle: I think yes, isnt just checking for "update cluster" capability enough since only cluster-owner can edit a cluster? now that we don't have a separate "manage clusterscan" role.

leodotcloud: yup, left over from previous PR.

app/role_data.go
```diff
@@ -25,33 +25,49 @@ var defaultAdminLabel = map[string]string{defaultAdminLabelKey: defaultAdminLabe
 func addRoles(management *config.ManagementContext) (string, error) {
 	rb := newRoleBuilder()
 
-	rb.addRole("Create Clusters", "clusters-create").addRule().apiGroups("management.cattle.io").resources("clusters").verbs("create").
+	rb.addRole("Create Clusters", "clusters-create").
+		addRule().apiGroups("management.cattle.io").resources("clusters").verbs("create").
```
prachidamle: Is this adding an extra line?

leodotcloud: not an extra line, just moving to next line. (the extra space is due to the browser, if you zoom out, you can see there is not extra line)

prachidamle:
@leodotcloud do we have an integration test for this to check that only cluster-owner can run the scan and other user gets error?

app/role_data.go
prachidamle: yep, looks ok in zoomed out view

leodotcloud:
@prachidamle I will add the tests in a separate PR. Created an issue to track https://github.com/rancher/rancher/issues/24709

cjellick:
You should have one fix per commit and your commit messages should be more descriptive.

This is a good guide for writing better commit messages:
https://chris.beams.io/posts/git-commit/

But, another option I've seen people adopt is simpley:
```
<message subject>

Problem:
<brief description of the problem>

Solution:
<brief description of the solution>
```

leodotcloud:
@prachidamle @kinarashah could you re-lgtm? Added better commit messages to take care of @cjellick's comments.

deniseschannon:
@leodotcloud There is still 1 commit for all the fixes. @cjellick was implying it should be 1 commit per fix. Or can the commit not be broken up?

pkg/api/customization/cluster/formatter.go
```diff
@@ -47,23 +47,21 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 			resource.AddAction(request, v3.ClusterActionBackupEtcd)
 			resource.AddAction(request, v3.ClusterActionRestoreFromEtcdBackup)
 		}
-		canUpdateCluster := func(request *types.APIContext, clusterID string) bool {
+		canUpdateClusterFn := func(request *types.APIContext, clusterID string) bool {
```
cjellick: why is this an embedded function?

pkg/api/customization/cluster/formatter.go
```diff
@@ -47,23 +47,21 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
-				nil,
-				request.Schema,
-			) == nil
+				cluster,
+				resource.Schema) == nil
 		}
-		if canUpdateCluster(request, resource.ID) && canCreateClusterScanObject() {
+		canUpdateCluster := canUpdateClusterFn(request, resource.ID)
+		logrus.Debugf("user: %v, canUpdateCluster: %v", request.Request.Header.Get("Impersonate-User"), canUpdateCluster)
+		if canUpdateCluster {
```
cjellick: I'm just missing something with this fix that makes me not understand it. What bug are you fixing by making this check less restrictive? Is the TL;DR that you don't need to be able to create scan objects in order to run a scan? That doesn't seem right?

cjellick:
What part of these changes fixes https://github.com/rancher/rancher/issues/24421?
What makes it so that cluster-member cannot see scans?

pkg/api/customization/cluster/formatter.go
leodotcloud: I saw similar code style being used at multiple other places, so I followed the same. For example: https://github.com/rancher/rancher/blob/master/pkg/api/customization/cluster/actionhandler.go#L30

I guess the embedded function improves code readability?

pkg/api/customization/cluster/formatter.go
leodotcloud: In an earlier PR, `Manage Cluster Scan` role was removed, hence there is no need for a redundant check for that role. With the current requirement, only a cluster-owner needs to be able to run the scan and since a cluster can be updated by cluster-owner using that check.

leodotcloud:
>What part of these changes fixes #24421?
>What makes it so that cluster-member cannot see scans?

Addition of this line here: https://github.com/rancher/rancher/pull/24698/commits/5214f76d95a2906d5d53775fcbdb4554bb86538f#diff-ae24e970146d9acb523685da279e80c7R27 propagates the following lines in the cluster-owner role

```
kubectl -n c-jx8wp get roles cluster-owner -o yaml
<snip>
- apiGroups:
  - '*'
  resources:
  - clusterscans
  verbs:
  - '*'
<snip>
```

`kubectl -n c-jx8wp get roles cluster-member -o yaml | grep scan` doesn't have any rules related to clusterscans.

leodotcloud:
@deniseschannon @cjellick, as requested broken down into multiple commits.

pkg/api/customization/cluster/formatter.go
cjellick: ok makes sense now ty

Issue #24662, opened 2019-12-18
Support for Calico 3.10


Tejeev:
Rancher version:
v2.3.2

Description:
We need to work with Calico Ver 3.10 because of the new global selectors.
Currently the default is 3.8.1; Do we know when we may see an upgrade?

deniseschannon:
When reviewing calico 3.10, we should ensure we update the following images to the latest version: 

https://github.com/rancher/kontainer-driver-metadata/blob/b0b335ca8f3c664f77f54525cdafef3ef4596ca7/rke/k8s_rke_system_images.go#L1799-L1806

deniseschannon:
Add new images to be pushed in this file: https://github.com/rawmind0/rancher/blob/c13661520987f911fb6020347339969860912805/scripts/istio-images-mirror

deniseschannon:
Since k8s 1.17 was released out of band, make sure to increment to a new k8s 1.17 version `v1.17.0-rancher1-2`. 

Also, can you make sure to comment which Rancher release,  the k8s versions was added. 

Can you also add a comment to the `v1.17.0-rancher1-1` that this was released "out of band"

deniseschannon:
@brendarearden we need to also add this into k8s 1.16 due to some issues we're finding with upgrades with k8s 1.17. can you add it into k-d-m for k8s 1.16?

brendarearden:
Do you want me to upgrade it for the most recent version of k8s 1.16.3 or do I need to add changes for all of the v1.16 (1.16.3, 1.16.2, 1.16.1)? Currently the default version is v1.16.3.

brendarearden:
k8s v1.16.4-rancher1-1 and v1.17.0-rancher1-2 both have Calico 3.10.2 available for testing

sowmyav27:
**Verified on rancher 2.3-head- commit id: `a4f19bb0b`**

k8s - 1.16 and 1.17 have rancher/calico-node:v3.10.2

brendarearden:
this should be ready to test with the latest master build

sowmyav27:
**Verified on master-head - commit id: `fa2cbfc57`**

- k8s - 1.16 and 1.17 have rancher/calico-node:v3.10.2
- automation checks have passed on both clusters with k8s - 1.16 and 1.17

PR #24702, opened 2019-12-19
Vendor updates

brendarearden:
#24662


deniseschannon:
Don't we need to bump up RKE and k-d-m?

brendarearden:
dependent on https://github.com/rancher/rke/pull/1856

Issue #24675, opened 2019-12-19
kontainer-driver-metadata changes are not updated for rkeaddons


Oats87:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**

We were trying to modify the `rkeaddon` for `canal-v1.16` (see: https://github.com/HighwayofLife/kontainer-driver-metadata/blob/master/rke/templates/canal.go#L2414)

We generated a new `data.json` and specified this repository in the Settings and triggered off a `rke-metadata-config` refresh.

We could see in the logs that:
```
2019/12/19 01:35:05 [INFO] driverMetadata: refresh data
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.10.12-rancher1-1
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.12.7-rancher1-3
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.13.9-rancher1-1
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.9.7-rancher2-2
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.10.3-rancher2-1
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.10.5-rancher1-2
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.13.9-rancher1-2
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.12.10-rancher1-2
2019/12/19 01:35:05 [DEBUG] REST UPDATE apis/management.cattle.io/v3/cattle-global-data/rkek8ssystemimages/v1.13.1-rancher1-1
[...]

PR #24707, opened 2019-12-20
fix metadata sync

kinarashah:
1) metadata CRDs were only updated when `sendRKELabel` was true (if they were supposed to send to RKE). We now always update the content, so in case user wants to override the values, they can simply set the label true after refresh

2) labels comparison (causing multiple REST Updates in debug logs, but never updating actual object)
a) System Images: `"cattle.io/creator": "norman"` is present in our CRD labels and it wasn't considered for comparing object labels for update
b) Addons, Service Options: we check for a label (`sendRKELabel`)'s value, but if the label is not present, it should default to true 

3) added logs for easier debugging 

https://github.com/rancher/rancher/issues/24675

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/data.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/data.go
@@ -334,7 +334,7 @@ func getLabelMap(k8sVersion string, data map[string]map[string]string,
 	if err != nil {
 		return nil, fmt.Errorf("k8sVersion not sem-ver %s %v", k8sVersion, err)
 	}
-	labelMap := map[string]string{}
+	labelMap := map[string]string{"cattle.io/creator": "norman"}
 	for addon, addonData := range data {
 		if addon == templates.TemplateKeys {
 			continue
@@ -689,6 +689,9 @@ func labelEqual(labels map[string]string, exists bool) bool {
 	toSendValue := "true"
 	if exists {
 		toSendValue = "false"
+	}
+	if _, ok := labels[sendRKELabel]; !ok {
+		return toSendValue == "true"
 	}
 	return toSendValue == labels[sendRKELabel]
 }
--- a/pkg/controllers/management/kontainerdrivermetadata/data.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/data.go
@@ -206,15 +206,22 @@ func (md *MetadataController) createOrUpdateSystemImageCRD(k8sVersion string, sy
 			return err
 		}
 		return nil
-
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -314,7 +334,7 @@ func getLabelMap(k8sVersion string, data map[string]map[string]string,
 	if err != nil {
 		return nil, fmt.Errorf("k8sVersion not sem-ver %s %v", k8sVersion, err)
 	}
-	labelMap := map[string]string{}
+	labelMap := map[string]string{"cattle.io/creator": "norman"}
```
cjellick: Why do you need to do this?

cjellick:
Commit messages should be more descriptive of the problem and the solution

pkg/controllers/management/kontainerdrivermetadata/data.go
kinarashah: our crds always get this label, so if I don't consider it for comparison -- there's always need to update labels on sync

pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -206,15 +206,22 @@ func (md *MetadataController) createOrUpdateSystemImageCRD(k8sVersion string, sy
 			return err
 		}
 		return nil
-
 	}
-	if reflect.DeepEqual(sysImage.SystemImages, systemImages) && reflect.DeepEqual(sysImage.Labels, pluginsMap) {
+	dataEqual := reflect.DeepEqual(sysImage.SystemImages, systemImages)
```
cjellick: are these map or list comparisons (on my phone so hard to tell). If they are lists, deep equals is dangerous bc it takes into account order but we don’t really care about order of labels (do we?)

cjellick: I’m not sure on this one, just my reaction. So if you feel deepequals is safe in this (and the other below cases) respond with justification and we can move on

kinarashah: SystemImages is struct comparison with all fields string, so thought it should be ok? 

For labels, I use custom logic https://github.com/rancher/rancher/blob/b4ba60ce4408b75dda0f2f33a4ab5b0227d09c55/pkg/controllers/management/kontainerdrivermetadata/data.go#L688

cjellick: Ok

PR #24708, opened 2019-12-20
[2.3] fix metadata sync

kinarashah:
https://github.com/rancher/rancher/issues/24675

1) metadata CRDs were only updated when `sendRKELabel` was true (if they were supposed to send to RKE). We now always update the content, so in case user wants to override the values, they can simply set the label true after refresh

2) labels comparison (causing multiple REST Updates in debug logs, but never updating actual object)
a) System Images: `"cattle.io/creator": "norman"` is present in our CRD labels and it wasn't considered for comparing object labels for update
b) Addons, Service Options: we check for a label (`sendRKELabel`)'s value, but if the label is not present, it should default to true 

3) added logs for easier debugging


prachidamle:
@kinarashah Can you add some description for the changes needed for the fix

PR #24726, opened 2019-12-20
introduce env var rancher_metadata_branch 

kinarashah:
- allow metadata branch to be set via env var `RANCHER_METADATA_BRANCH` 
- need to return error if commit is not found while loading data 

https://github.com/rancher/rancher/issues/24743

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -14,6 +14,8 @@ WORKDIR /var/lib/rancher
 ARG ARCH=amd64
 ARG IMAGE_REPO=rancher
 ARG SYSTEM_CHART_DEFAULT_BRANCH=v2.3-dev
+# kontainer-driver-metadata branch to be set for specific branch other than dev/master, logic at rancher/rancher/pkg/settings/setting.go  
+ARG RANCHER_METADATA_BRANCH=
 
 ENV CATTLE_SYSTEM_CHART_DEFAULT_BRANCH=$SYSTEM_CHART_DEFAULT_BRANCH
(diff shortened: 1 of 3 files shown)


package/Dockerfile
```diff
@@ -14,6 +14,7 @@ WORKDIR /var/lib/rancher
 ARG ARCH=amd64
 ARG IMAGE_REPO=rancher
 ARG SYSTEM_CHART_DEFAULT_BRANCH=v2.3-dev
+ARG RANCHER_METADATA_BRANCH=
```
deniseschannon: since the default is assumed "", we should comment that the logic of picking dev vs master is handled somewhere else and that this is for if an explicit branch (besides dev and master) need to be used.

pkg/settings/setting.go
```diff
@@ -154,9 +154,12 @@ func GetEnvKey(key string) string {
 func getMetadataConfig() string {
 	key := GetEnvKey("server-version")
 	rancherVersion := os.Getenv(key)
-	branch := "dev"
-	if releaseServerVersion(rancherVersion) {
-		branch = "master"
+	branch := os.Getenv("RANCHER_METADATA_BRANCH")
```
deniseschannon: @cjellick are there any issues with trying to use a default value of an env var as empty?

deniseschannon: spoke with @cjellick offline and there is no downside to this concept

package/Dockerfile
kinarashah: updated with comment

Issue #21386, opened 2019-07-09
Cattle-cluster-agent and cattle-node-agent do not rollback to the previous version


soumyalj:
Version: v2.2.4 to v2.2.5-rc9

Steps:
1. Create a custom cluster in rancher v2.2.4 and create few workloads in the cluster
2. Upgrade to rancher v2.2.5-rc9. 
3. After upgrade, cattle-cluster-agent and cattle-node-agent get upgraded to v2.2.5-rc9
3. Rollback to v2.2.4

After rollback, the cattle-cluster-agent and cattle-node-agent do not rollback to v2.2.4.

![Screen Shot 2019-07-08 at 9.24.19 PM.png](https://zube.io/files/rancher/fd144f7e30ba17711fff7cb60cdb97f6-screen-shot-2019-07-08-at-9-24-19-pm.png)

The same issue is seen for the the below rollbacks too.
1. Upgrade from v2.0.15 to v2.0.16-rc1 and rollback to v2.0.15
2. Upgrade from v2.1.10 to v2.1.11-rc1 and rollback to v2.1.10

dramich:
I am not able to repro this. When doing an upgrade and rollback it can take some time for the cluster to run through the handlers and get updated. If checked immediately after upgrade they won't have had time to go through the controllers and update. It took about 30 seconds before I started seeing changes.

cjellick:
Sending back to QA for er-verification that it exists and to see if we missed anything important about how to reproduce.

soumyalj:
@dramich @cjellick : Reproduced the issue on an upgrade from v2.2.5 to master-head and rollback to v2.2.5. The agent continues to be master-head version

dramich:
@soumyalj can you provide the steps you are doing? How are you running rancher > upgrade > rollback. Where is rancher running? I see the cluster you created is DO.

soumyalj:
@dramich : 
1. Create a single node rancher setup running on a DO/AWS machine. 
2. Upgrade rancher using steps in the docs. 
https://rancher.com/docs/rancher/v2.x/en/upgrades/upgrades/single-node/
3. Rollback rancher using the steps 
https://rancher.com/docs/rancher/v2.x/en/upgrades/rollbacks/single-node-rollbacks/

Tejeev:
We have seen this issue occur with the following scenario:

1. Installed a Rancher HA Cluster with 2.3.0 and setup a downstream Kubernetes cluster with it. 
[...]

PR #24731, opened 2019-12-20
[2.3] Downstream Agent Tracking

luthermonson:
Adding a map in the clusterdeploy template to track the downstream agent incase of upgrade/rollback as per this bug report https://github.com/rancher/rancher/issues/21386

Read my writeup in that issue which tells you how to recreate this problem locally.

### DIFF
--- a/pkg/controllers/management/clusterdeploy/clusterdeploy.go
+++ b/pkg/controllers/management/clusterdeploy/clusterdeploy.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/rancher/norman/types"
@@ -20,12 +21,26 @@ import (
 	"github.com/rancher/types/config"
 	"github.com/rancher/types/user"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
-const AgentForceDeployAnn = "io.cattle.agent.force.deploy"
+const (
+	AgentForceDeployAnn = "io.cattle.agent.force.deploy"
+	nodeImage           = "nodeImage"
+	clusterImage        = "clusterImage"
+)
+
+var (
+	agentImagesMutex sync.RWMutex
+	agentImages      = map[string]map[string]string{
+		nodeImage:    map[string]string{},
+		clusterImage: map[string]string{},
+	}
+)
 
 func Register(ctx context.Context, management *config.ManagementContext, clusterManager *clustermanager.Manager) {
 	c := &clusterDeploy{
@@ -101,6 +116,13 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
 	if err != nil {
 		return err
 	}
+
+	if cluster.Status.AgentImage != "" && !agentImagesCached(cluster.Name) {
+		if err := cd.cacheAgentImages(cluster.Name); err != nil {
+			return err
+		}
+	}
+
 	err = cd.deployAgent(cluster)
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -130,10 +165,20 @@ func redeployAgent(cluster *v3.Cluster, desiredAgent, desiredAuth string) bool {
...
-			" private repo changed=%v", forceDeploy, imageChange, repoChange)
+		logrus.Infof("Redeploy Rancher Agents is needed for %s: forceDeploy=%v, agent/auth image changed=%v,"+
+			" private repo changed=%v", cluster.Name, forceDeploy, imageChange, repoChange)
 		return true
 	}
+
+	na, ca := getAgentImages(cluster.Name)
+	if cluster.Status.AgentImage != na || cluster.Status.AgentImage != ca {
+		//downstream agent does not match, kick a redeploy with settings agent
+		clearAgentImages(cluster.Name)
```
kinarashah: we already set the agent images in deploy logic, why do we need to clean it up here and L142? https://github.com/rancher/rancher/blob/b5409bab07aa3be1000ded896c05c74300a71956/pkg/controllers/management/clusterdeploy/clusterdeploy.go#L239

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -110,6 +129,22 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
...
+		}
+		if cluster.Status.AgentImageOverride == cluster.Status.AgentImage {
+			return false
+		}
+
+		//kick a redeploy with the overriden agent
+		clearAgentImages(cluster.Name)
+		cluster.Status.AgentImage = ""
+		cluster.Spec.DesiredAgentImage = cluster.Status.AgentImageOverride
+		return true
```
kinarashah: for debugging purposes, let's add log statement whenever redeploy returns true same as L168

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -101,6 +115,11 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
 	if err != nil {
 		return err
 	}
+
+	if cluster.Status.AgentImage != "" && !agentImagesCached(cluster.Name) {
+		cd.cacheAgentImages(cluster.Name)
```
kinarashah: we need to handle error returned from cacheAgentImages

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -257,3 +303,79 @@ func (cd *clusterDeploy) getYAML(cluster *v3.Cluster, agentImage, authImage stri
...
 }
+
+func (cd *clusterDeploy) getClusterAgentImage(name string) (string, error) {
+	uc, err := cd.clusterManager.UserContext(name)
+	if err != nil {
+		return "", err
+	}
+
+	d, err := uc.Apps.Deployments("cattle-system").Get("cattle-cluster-agent", v1.GetOptions{})
+	if err != nil {
```
kinarashah: is this logic also used to deploy the deployment/daemonset for the first time? if so, we need to handle the NotFound error, and return `""` ?

pkg/controllers/management/clusterdeploy/clusterdeploy.go
luthermonson: ah that was unintentional. sorry will fix

pkg/controllers/management/clusterdeploy/clusterdeploy.go
luthermonson: just clear logic just in case, would prefer they were blank at the time we know they should be blank. youre right tho it's not technically necessary and would drop it if you insist

pkg/controllers/management/clusterdeploy/clusterdeploy.go
luthermonson: i agree, will do

pkg/controllers/management/clusterdeploy/clusterdeploy.go
luthermonson: we should only ever hit this logic if we have just deployed a cluster or are recaching the agents after a restart. with our logic it shouldnt get a notfound but definite need to account for it just in case and will add, good catch.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
kinarashah: gotcha, fine by me, just wanted to understand if there was some use case we were addressing with this

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -257,3 +310,85 @@ func (cd *clusterDeploy) getYAML(cluster *v3.Cluster, agentImage, authImage stri
...
+	agentImagesMutex.RLock()
+	defer agentImagesMutex.RUnlock()
+	return agentImages[nodeImage][name], agentImages[clusterImage][name]
+}
+
+func clearAgentImages(name string) {
+	agentImagesMutex.Lock()
+	defer agentImagesMutex.Unlock()
+	delete(agentImages[nodeImage], name)
+	delete(agentImages[nodeImage], name)
```
cjellick: is this supposed to be `nodeImage`?

luthermonson: GGGGGGREEEAT catch, it is

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -110,6 +132,24 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
...
+		if cluster.Status.AgentImageOverride == cluster.Status.AgentImage {
+			return false
+		}
+
+		// kick a redeploy with the overriden agent
+		logrus.Infof("Redeploy Rancher Agents due to Agent Image Override for %s: was %s and will be %s",
+			cluster.Name, cluster.Status.AgentImage, cluster.Status.AgentImageOverride)
+		clearAgentImages(cluster.Name)
+		cluster.Status.AgentImage = ""
+		cluster.Spec.DesiredAgentImage = cluster.Status.AgentImageOverride
```
cjellick: This function used to be a side-effect free evaluator of whether or not you should redeploy an agent. It now has the side effect of setting two fields on the cluster object. I think this is a bad idea and I think the function should remain side-effect free. Its also not clear to me that setting this field is going to have the effect that you want. Will discuss in person.

cjellick:
LGTM but you have to get a types pr merged and properly updated, hence me holding back on official "approval" until then

PR #24760, opened 2019-12-24
Update CRB and RB if roleTemplate's inherited roles are edited

mrajashree:
https://github.com/rancher/rancher/issues/23658

When a user is added to project/cluster, based on the roleTemplate, rolebindings
and clusterrolebindings are created for that user.

When user is added to a  project, following clusterRoleBindings are created
* roleTemplate.Name-promoted: if storageClass, persistentVolumes and apiservices are
among the roles's resources
* project.Name-namespaces-edit and create-ns: if there is "create" or "*" permission on the
"namespaces" resource of any of the rules in the roleTemplate
* project.Name-namespaces-readonly: if there is no "create" permission on "namespaces" resources in any of the rules in roletemplate

Along with this, roleBindings are created in the namespaces belonging to that project:
* One with the clusterRole of the roleTemplate.Name
* One with the Kubernetes default role associated with the roleTemplate

The project-member role inherits the "edit" Kubernetes default role. So when a new roleTemplate
is created by cloning this role, that will also inherit the "edit" role.

The bug is that if a user is added to a project with this cloned role, on editing the cloned role and removing the "edit" inherited role, and replacing it with "view", the corresponding roleBindings for the user don't get updated.

This commit fixes it by calling the same logic that exists on the creation/edit of a prtb,
on the edit of a roleTemplate as well.

### DIFF
--- a/pkg/controllers/user/rbac/roletemplate_handler.go
+++ b/pkg/controllers/user/rbac/roletemplate_handler.go
@@ -1,14 +1,9 @@
 package rbac
 
 import (
-	"fmt"
-	"strings"
-
 	"github.com/pkg/errors"
-	"github.com/rancher/norman/types/slice"
-	pkgrbac "github.com/rancher/rancher/pkg/rbac"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
-	rbacv1 "k8s.io/api/rbac/v1"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


pkg/controllers/user/rbac/roletemplate_handler.go
```diff
@@ -92,114 +88,51 @@ func (c *rtLifecycle) syncRT(template *v3.RoleTemplate, usedInProjects bool, prt
 		}
 	}
 
+	for _, rtName := range template.RoleTemplateNames {
+		rt, err := c.m.rtLister.Get("", rtName)
+		if err != nil {
+			return err
+		}
+		roles[rtName] = rt
```
rmweir: what about the parents of parents and so on? in prtb handler we are gathering them recursively, should we be doing the same here?

mrajashree: syncRT also calles gatherRoles which gathers them recursively https://github.com/rancher/rancher/pull/24760/files#diff-e5bf9d33608d38443a803c450bc5a590L68

So this for loop isn't needed, removing it

rmweir: perfect

Issue #24545, opened 2019-12-12
Activating Cluster level logging sends lots of backslashes on fluentd target


izaac:
Activating Cluster level logging sends lots of backslashes making impossible to read the logs.

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Spin up a new cluster using rancher 2.3-head. I used `[INFO] Rancher version` 59110e936
Activate cluster logging using fluentd as target. I've used `fluentd:v1.7.4-debian-2.0`

`fluent.conf`:
```
<source>
  @type forward
  port 9881
  bind 0.0.0.0
</source>
<match **>
  @type stdout
</match>
```

**Result:**
Many backslashes send in the logged events to fluentd.

**Other details that may be helpful:**

**Environment information**
- Rancher version : Rancher version` 59110e936
- Installation option (single install/HA): Single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): GKE
- Machine type (cloud/VM/metal) and specifications (CPU/memory): cloud
- Kubernetes version (use `kubectl version`):

```
v1.14.8-gke.17
```

- Docker version (use `docker version`):

```
18.9.7
```

izaac:
Validated Rancher version: `2.3.3` I didn't reproduce the issue.

izaac:
This seems to happen more often with the `"enableJSONParsing": bool` cluster level settings enabling and disabling it with the API seems somehow trigger this behavior. @dramich

dramich:
https://github.com/rancher/rancher/issues/24157

izaac:
It doesn't seem to be the same scenario. The result is the same for sure but not enabling the enableJSONParsing in `2.3.4-rc3` Makes the logs to behave as expected but as soon as I turn it to `true` the backslash started flowing. @dramich

loganhz:
## Root Cause

Cluster logging will collect logs from many components not all of the logs are in JSON format.

So fluentd will throw a warning level log when we do the parse if it is not a JSON format log. This is expected.

However, rancher fluentd is picking up and processing its own logs, which is causing a feedback loop - every time it reads a log with JSON escaped strings, it has to escape it again before printing, which leads to lots of backslashes to escape other backslashes.

## Solution

Use `multi_format` plugin to avoid fluentd logging warning when parsing the log to JSON if it's not in JSON format.

Step 1: Release a new rancher/fluentd image with multi_format plugin installed
[...]

PR #24763, opened 2019-12-25
[2.3.4] Log json parsing format and sort namespaces

shanewxy:
**1. Fixed fluentd log parsing full of backslash**

Solution:
Use multi_format plugin to avoid fluentd logging warning when parsing the log to JSON if it's not in JSON format.

Related issue:
https://github.com/rancher/rancher/issues/24545

**2. Enable sorting grep system namespace for logging**

Related PR:
https://github.com/rancher/system-charts/pull/141

Need to merge https://github.com/rancher/system-charts/pull/141 first


luthermonson:
@shanewxy needs two commits squashed but gtg otherwise.

shanewxy:
rebased

PR #24777, opened 2019-12-27
Add redeploy to workload API and integration test

shanewxy:
Exposed redeploy REST API to the workload which set sets `cattle.io/timestamp` annotation to the current time.

Related issue: https://github.com/rancher/rancher/issues/23053
Related types PR: https://github.com/rancher/types/pull/1066

### DIFF
--- a/pkg/api/customization/workload/workload.go
+++ b/pkg/api/customization/workload/workload.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/rancher/norman/api/access"
@@ -28,6 +29,10 @@ import (
 const (
 	workloadRevisions    = "revisions"
 	DeprecatedRollbackTo = "deprecated.deployment.rollback.to"
+)
+
+var (
+	allowRedeployTypes = map[string]bool{"cronJob": true, "deployment": true, "replicationController": true, "statefulSet": true, "daemonSet": true, "replicaSet": true}
 )
 
 type ActionWrapper struct {
@@ -64,6 +69,8 @@ func (a ActionWrapper) ActionHandler(actionName string, action *types.Action, ap
 			return httperror.NewAPIError(httperror.InvalidAction, fmt.Sprintf("Pause deployment %s before resume", deployment.ID))
 		}
 		return updatePause(apiContext, false, deployment, "resume")
+	case "redeploy":
+		return updateTimestamp(apiContext, deployment)
 	}
 	return nil
 }
@@ -203,7 +210,19 @@ func (a ActionWrapper) rollbackDeployment(apiContext *types.APIContext, clusterC
 	}
 	return nil
 }
-
+func updateTimestamp(apiContext *types.APIContext, workload projectclient.Workload) error {
+	timestamp := time.Now().UTC().Format(time.RFC3339)
+	data, err := convert.EncodeToMap(workload)
+	if err != nil {
+		return httperror.WrapAPIError(err, httperror.ServerError, "Failed to parse workload")
+	}
+	values.PutValue(data, timestamp, "annotations", "cattle.io/timestamp")
+	err = update(apiContext, data, workload.ID)
+	if err != nil {
+		return httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("Error redeploying workload %s : %s", workload.ID, err.Error()))
+	}
+	return nil
+}
 func (h Handler) LinkHandler(apiContext *types.APIContext, next types.RequestHandler) error {
 	if apiContext.Link == workloadRevisions {
 		var deployment projectclient.Workload
@@ -224,6 +243,10 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 	resource.Links["self"] = apiContext.URLBuilder.ResourceLinkByID(workloadSchema, workloadID)
 	resource.Links["remove"] = apiContext.URLBuilder.ResourceLinkByID(workloadSchema, workloadID)
 	resource.Links["update"] = apiContext.URLBuilder.ResourceLinkByID(workloadSchema, workloadID)
+	//add redeploy action to the workload types that support redeploy
+	if _, ok := allowRedeployTypes[resource.Type]; ok {
+		resource.Actions["redeploy"] = apiContext.URLBuilder.ActionLinkByID(workloadSchema, workloadID, "redeploy")
+	}
 
 	delete(resource.Values, "nodeId")
 }
--- a/tests/integration/suite/test_workloads.py
+++ b/tests/integration/suite/test_workloads.py
@@ -1,7 +1,9 @@
 from .common import random_str
 from rancher import ApiError
+from .conftest import wait_for
 
 import time
+import datetime
 import pytest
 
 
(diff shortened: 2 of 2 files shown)


pkg/api/customization/workload/workload.go
```diff
@@ -64,6 +67,9 @@ func (a ActionWrapper) ActionHandler(actionName string, action *types.Action, ap
 			return httperror.NewAPIError(httperror.InvalidAction, fmt.Sprintf("Pause deployment %s before resume", deployment.ID))
 		}
 		return updatePause(apiContext, false, deployment, "resume")
+	case "redeploy":
+		timestamp := time.Now().UTC().Format(protocol.ISO8601TimeFormat)
```
gitlawr: Can we use a format in the standard library(https://golang.org/pkg/time/)
As the package name suggested `github.com/aws/aws-sdk-go/private/protocol` is meant to be used internally so we should not depend on it.

pkg/api/customization/workload/workload.go
```diff
@@ -236,6 +252,7 @@ func DeploymentFormatter(apiContext *types.APIContext, resource *types.RawResour
 	resource.Actions["pause"] = apiContext.URLBuilder.ActionLinkByID(workloadSchema, workloadID, "pause")
 	resource.Actions["resume"] = apiContext.URLBuilder.ActionLinkByID(workloadSchema, workloadID, "resume")
 	resource.Actions["rollback"] = apiContext.URLBuilder.ActionLinkByID(workloadSchema, workloadID, "rollback")
+	resource.Actions["redeploy"] = apiContext.URLBuilder.ActionLinkByID(workloadSchema, workloadID, "redeploy")
```
gitlawr: This only applies to deployments. We can also redeploy a statefulset/daemonset/replicaset in the UI. Can you check the complete list and make it work for all those types?

tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +311,34 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
...
+        annotations={
+            "cattle.io/timestamp":
+                datetime.datetime.now().isoformat()}, )
+    p_client.action(workload, "redeploy")
+    timestamp = p_client.list_workload(
+        uuid=workload.uuid).data[0].annotations["cattle.io/timestamp"]
+    assert timestamp != workload.annotations["cattle.io/timestamp"]
+    p_client.delete(workload)
+    admin_pc.cluster.client.delete(ns)
+    admin_mc.client.delete(project)
```
gitlawr: The admin_pc fixture handles the project creation/deletion so we don't need to do that in the tests

pkg/api/customization/workload/workload.go
```diff
@@ -203,7 +206,19 @@ func (a ActionWrapper) rollbackDeployment(apiContext *types.APIContext, clusterC
 	}
 	return nil
 }
-
+func updateTimestamp(apiContext *types.APIContext, workload projectclient.Workload) error {
+	timestamp := time.Now().UTC().Format(time.RFC3339)
+	data, err := convert.EncodeToMap(workload)
+	if err != nil {
+		return httperror.WrapAPIError(err, httperror.ServerError, "Failed to parse workload")
```
mrajashree: error message should be "Failed to redeploy workload"?

pkg/api/customization/workload/workload.go
```diff
@@ -203,7 +206,19 @@ func (a ActionWrapper) rollbackDeployment(apiContext *types.APIContext, clusterC
...
+	}
+	values.PutValue(data, timestamp, "annotations", "cattle.io/timestamp")
+	err = update(apiContext, data, workload.ID)
+	if err != nil {
+		return httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("Error updating workload %s by redeploy : %s", workload.ID, err.Error()))
```
mrajashree: can we say "Error redeploying workload" instead?

tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +310,31 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
         assert e.value.error.status == 422
 
 
+def test_workload_redeploy(admin_pc):
+    client = admin_pc.client
+    ns = admin_pc.cluster.client.create_namespace(
+        name=random_str(),
+        projectId=admin_pc.project.id)
```
mrajashree: can we add `remove_resource` for this namespace and the workload on line 319 too, we won't have to use client.delete

tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +310,31 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
...
+        scale=1,
+        containers=[{
+            'name': 'one',
+            'image': 'nginx',
+        }],
+        annotations={
+            "cattle.io/timestamp":
+                datetime.datetime.now().isoformat()}, )
+    client.action(workload, "redeploy")
+    timestamp = client.list_workload(
```
mrajashree: do we need some retry/sleep/check logic around this? might lead to error in drone

pkg/api/customization/workload/workload.go
```diff
@@ -224,6 +239,11 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 	resource.Links["self"] = apiContext.URLBuilder.ResourceLinkByID(workloadSchema, workloadID)
 	resource.Links["remove"] = apiContext.URLBuilder.ResourceLinkByID(workloadSchema, workloadID)
 	resource.Links["update"] = apiContext.URLBuilder.ResourceLinkByID(workloadSchema, workloadID)
+	resourceType := resource.Type
+	//add redeploy action to the workload types that support redeploy
```
mrajashree: maybe nit: can we have a map of workloads that allow redeploy instead and check with that?

shanewxy:
@mrajashree updated

shanewxy:
Updated to latest rancher types

tests/integration/suite/test_workloads.py
```diff
@@ -341,3 +369,19 @@ def wait_for_service_cluserip_reset(client, name, timeout=30):
         if time.time() - start > timeout:
             raise Exception('Timeout waiting for workload service')
     return services.data[0]
+
+
+def wait_for_workload_active(client, workload, timeout=30):
```
dramich: This is a duplication of code that already lives in `wait_for`, use that and pass in your callback and fail handler.

tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +310,34 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
...
+            "cattle.io/timestamp":
+                datetime.datetime.now().isoformat()}, )
+    remove_resource(workload)
+    workload = wait_for_workload_active(client, workload)
+    client.action(workload, "redeploy")
+    timestamp = client.list_workload(
+        uuid=workload.uuid).data[0].annotations["cattle.io/timestamp"]
+    assert timestamp != workload.annotations["cattle.io/timestamp"]
+    client.delete(workload)
+    admin_pc.cluster.client.delete(ns)
```
dramich: Both of these delete calls are unnecessary, `remove_resource` takes care of this and will handle errors.

shanewxy:
@dramich updated

tests/integration/suite/test_workloads.py
rmweir: I agree with @mrajashree this really needs retry as the annotation may not be set right away

tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +312,39 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
...
+        annotations={
+            "cattle.io/timestamp":
+                datetime.datetime.now().isoformat()}, )
+    remove_resource(workload)
+
+    def workload_active():
+        workloads = client.list_workload(uuid=workload.uuid).data
+        return workloads[0].state == 'active'
+
+    wait_for(workload_active,
```
rmweir: I think we can already observe that the default time in wait_for (45 seconds) is not long enough. See these two breaks: https://drone-pr.rancher.io/rancher/rancher/6875, https://drone-publish.rancher.io/rancher/rancher/2202

PR #24781, opened 2019-12-27
local filesystem test for clusters snapshots

izaac:
Adding support to check the local filesystem snapshot files in `/opt/rke/etcd-snapshots`
Creating a AWS custom cluster to be able to ssh into the `etcd` nodes.
Iterate over the `etcd` nodes to check each node's local filesystem.

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -53,10 +53,15 @@ AWS_S3_BUCKET_FOLDER_NAME = os.environ.get("AWS_S3_BUCKET_FOLDER_NAME", "")
 LINODE_ACCESSKEY = os.environ.get('RANCHER_LINODE_ACCESSKEY', "None")
 
 TEST_RBAC = ast.literal_eval(os.environ.get('RANCHER_TEST_RBAC', "False"))
-if_test_rbac = pytest.mark.skipif(TEST_RBAC is False, reason='rbac tests are skipped')
+if_test_rbac = pytest.mark.skipif(TEST_RBAC is False,
+                                  reason='rbac tests are skipped')
 
-TEST_S3 = ast.literal_eval(os.environ.get('RANCHER_TEST_S3', "False"))
-if_test_s3 = pytest.mark.skipif(TEST_S3 is False, reason='S3 tests are skipped')
+TEST_ALL_SNAPSHOT = ast.literal_eval(
+    os.environ.get('RANCHER_TEST_ALL_SNAPSHOT', "False")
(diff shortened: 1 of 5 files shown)


tests/validation/tests/v3_api/common.py
```diff
@@ -53,10 +53,15 @@
...
+                                  reason='rbac tests are skipped')
 
-TEST_S3 = ast.literal_eval(os.environ.get('RANCHER_TEST_S3', "False"))
-if_test_s3 = pytest.mark.skipif(TEST_S3 is False, reason='S3 tests are skipped')
+TEST_S3 = ast.literal_eval(
+    os.environ.get('RANCHER_TEST_ALL_SNAPSHOT', "False")
+)
+if_test_all_snapshot = \
+    pytest.mark.skipif(TEST_S3 is False,
+                       reason='Snapshots check tests are skipped')
```
sangeethah: Rename TEST_S3 to something that reflects RANCHER_TEST_ALL_SNAPSHOT .

tests/validation/tests/v3_api/common.py
```diff
@@ -940,7 +945,8 @@ def exec_shell_command(ip, port, cmd, password, user="root", sshKey=None):
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
     if sshKey:
-        ssh.connect(ip, username=user, key_filename=sshKey, port=port)
+        key = paramiko.RSAKey.from_private_key_file(sshKey)
+        ssh.connect(ip, username=user, pkey=key, port=port)
```
sangeethah: Is using key_filename not working in this case ? Would prefer not to introduce this change , since this code is being used as part of custom cluster creation and works as expected.

tests/validation/tests/v3_api/test_bkp_restore.py
```diff
@@ -118,8 +150,54 @@ def fin():
...
+    namespace["cluster"] = cluster
+    namespace["project"] = p
+    namespace["c_client"] = c_client
+
+    def fin():
+        client.delete(namespace["project"])
+        cluster_cleanup(client, cluster, aws_nodes)
+    request.addfinalizer(fin)
+
+
```
sangeethah: create_project_client_and_cluster_local fixture is not getting used in any place?

tests/validation/tests/v3_api/common.py
izaac: We're now using `execute_command` from `Node` class.

Issue #24869, opened 2020-01-06
Node annotations removed through UI/API cannot be re-added


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a cluster with at least one node.
2. Add an annotation to a node, example:
```
kubectl annotate node ip-10-0-2-14 hello=world
```
3. In the UI, navigate to the node, and select edit. Remove the hello=world annotation.
4. Add the annotation again by running same kubectl command:
```
kubectl annotate node ip-10-0-2-14 hello=world
```

**Result:**
The annotation does not exist on the node after re-adding:
```
$ kubectl get node  ip-10-0-2-14 -o custom-columns=name:.metadata.name,annotations:.metadata.annotations
name           annotations
ip-10-0-2-14   map[node.alpha.kubernetes.io/ttl:0 projectcalico.org/IPv4IPIPTunnelAddr:10.42.4.1 rke.cattle.io/external-ip:10.0.2.14 rke.cattle.io/internal-ip:10.0.2.14 volumes.kubernetes.io/controller-managed-attach-detach:true]
```

**Other details that may be helpful:**
The node has this, which is not getting cleared:
```
  metadataUpdate:
    annotations:
      delete:
        hello: true

```
Full YAML:
```
kubectl get nodes.management.cattle.io m-ee6a8368db15 -n c-xbbn6 -o yaml
apiVersion: management.cattle.io/v3
kind: Node
metadata:
  annotations:
    lifecycle.cattle.io/create.node-controller: "true"
    lifecycle.cattle.io/create.user-node-remove_c-xbbn6: "true"
    management.cattle.io/nodesyncer: "true"
  creationTimestamp: "2020-01-06T20:23:30Z"
  finalizers:
  - clusterscoped.controller.cattle.io/user-node-remove_c-xbbn6
  - controller.cattle.io/node-controller
  generation: 36
  labels:
    cattle.io/creator: norman
    management.cattle.io/nodename: ip-10-0-2-14
  name: m-ee6a8368db15
  namespace: c-xbbn6
  resourceVersion: "7737"
  selfLink: /apis/management.cattle.io/v3/namespaces/c-xbbn6/nodes/m-ee6a8368db15
  uid: 67294fb3-30c2-11ea-a2e0-0242ac110002
spec:
  controlPlane: false
  customConfig:
    address: 10.0.2.14
  desiredNodeTaints: null
  displayName: ""
[...]

PR #24872, opened 2020-01-06
[2.3] reset desired annotations after update

kinarashah:
#24869


pkg/controllers/user/nodesyncer/nodessyncer.go
```diff
@@ -196,7 +196,7 @@ func (m *nodesSyncer) updateAnnotations(node *corev1.Node, obj *v3.Node, nodePla
 
 	node, obj = node.DeepCopy(), obj.DeepCopy()
 	node.Annotations = finalMap
-	obj.Spec.MetadataUpdate.Labels = v3.MapDelta{}
+	obj.Spec.MetadataUpdate.Annotations = v3.MapDelta{}
```
deniseschannon: Is it an issue that we are removing the labels update?

kinarashah: No it was a typo, we have a separate function which handles labels

PR #24874, opened 2020-01-07
Add method to create custom cluster given set of node roles.

sangeethah:
Added method to create custom cluster given set of node roles.


izaac:
LGTM

Issue #24759, opened 2019-12-24
CIS Scan should not be run on windows cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Currently when CIS scan is run on a windows cluster, the scan workload is getting deployed on windows and is crashing. 

- We should not be allowed to run Scans on a windows cluster. - from the API

- And the UI should not allow the action to be performed on the cluster

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head / 2.4.0-alpha1
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Kubernetes version (use `kubectl version`):

```
1.16
```

sowmyav27:
**Verified on latest master - commit id: `4f1aff9b7`**

runSecurityScans options is disabled on a windows cluster - both VXLAN and Host Gateway network provider. from the API

But the action is allowed from the UI.
Error is seen when user clicks on Run Scans.
<img width="1502" alt="Screen Shot 2020-01-30 at 8 46 21 AM" src="https://user-images.githubusercontent.com/26032343/73470836-5a94f900-433d-11ea-8fec-fa4c873fa34c.png">

Expected:
The option should be disabled on the UI - similar to other cluster types - other than RKE.

sowmyav27:
Verified on master-head - commit id: `3fc6715ca`.

- Windows cluster - VXLan mode, has runSecurityScan disabled in the API
- This cluster does not have the options - Run CIS scan in the more option drop down on
- - Cluster dashboard page
- - Cluster page on the cluster
- Navigate to Cluster --> Tools --> CIS Scan -- > Run scan button is disabled.

Issue #24755, opened 2019-12-24
Error seen while navigating to CIS Scans page when the cluster is in Provisioning state


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy a cluster.
2. While the cluster is in Provisioning state, Navigate to Cluster --> Tools --> CIS Scans

**Error seen on the UI:**

<img width="641" alt="Screen Shot 2019-12-23 at 9 16 32 PM" src="https://user-images.githubusercontent.com/26032343/71395542-3ec66680-25cb-11ea-874e-c7f0ce893b6d.png">

From the API, when the cluster is Provisioning, when a User performs action "runSecurityScan", error seen:
```
{
"baseType": "error",
"code": "ClusterUnavailable",
"message": "cluster not ready",
"status": 503,
"type": "error"
}
```

**Expected Result:**
From the API, the action "runSecurityScan" be disabled until the cluster is Active.

UI be able to handle the navigation to CIS Scans page like other page navigations while the cluster is in provisioning state

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.0-alpha1
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke/custom
- Kubernetes version (use `kubectl version`):

```
1.16
```

sowmyav27:
**Verified on latest master-head - commit id: `47bf996c2`.**

- When a cluster is in provisioning state, the action "runSecurityScan" is seen disabled.

- But when the user navigates to Cluster --> Global --> Tools --> CIS Scans , User is presented with this error on the UI

<img width="594" alt="Screen Shot 2020-01-23 at 12 55 18 PM" src="https://user-images.githubusercontent.com/26032343/73024260-17d59d00-3de2-11ea-857f-fca173ea54d1.png">

leodotcloud:
@codyrancher I have disabled the action from the backend. Can you please look into the UI aspect of this?

codyrancher:
@sowmyav27 is this still an issue that you see? If so, what type of cluster are you creating and with what settings? Do you have to wait for the cluster to be in a specific stage?

This is what I see when I have a cluster provisioning in the latest.
[...]

PR #24944, opened 2020-01-13
disable runSecurityScan for non-active and windows clusters

leodotcloud:
https://github.com/rancher/rancher/issues/24755
https://github.com/rancher/rancher/issues/24759

### DIFF
--- a/pkg/api/customization/cluster/actions_cis.go
+++ b/pkg/api/customization/cluster/actions_cis.go
@@ -54,6 +54,10 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
 			fmt.Sprintf("cluster with id %v doesn't exist", apiContext.ID))
 	}
 
+	if cluster.Spec.WindowsPreferedCluster {
+		return httperror.WrapAPIError(err, httperror.InvalidAction,
+			fmt.Sprintf("cannot run scan on a windows cluster"))
+	}
 	if cluster.DeletionTimestamp != nil {
 		return httperror.NewAPIError(httperror.InvalidType,
 			fmt.Sprintf("cluster with id %v is being deleted", apiContext.ID))
--- a/pkg/api/customization/cluster/formatter.go
+++ b/pkg/api/customization/cluster/formatter.go
@@ -30,6 +30,22 @@ func NewFormatter(schemas *types.Schemas, managementContext *config.ScaledContex
 	return &clusterFormatter
 }
 
+func canUserUpdateCluster(request *types.APIContext, resource *types.RawResource) bool {
+	if request == nil || resource == nil {
+		return false
+	}
+	cluster := map[string]interface{}{
+		"id": resource.ID,
+	}
+	return request.AccessControl.CanDo(
+		v3.ClusterGroupVersionKind.Group,
+		v3.ClusterResource.Name,
+		"update",
+		request,
+		cluster,
+		resource.Schema) == nil
+}
+
 func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawResource) {
 	if convert.ToBool(resource.Values["internal"]) {
 		delete(resource.Links, "remove")
@@ -56,19 +72,7 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 			isWindowsCluster = true
 		}
 		if isActiveCluster && !isWindowsCluster {
-			canUpdateClusterFn := func(request *types.APIContext, clusterID string) bool {
-				cluster := map[string]interface{}{
-					"id": clusterID,
-				}
-				return request.AccessControl.CanDo(
-					v3.ClusterGroupVersionKind.Group,
-					v3.ClusterResource.Name,
-					"update",
-					request,
-					cluster,
-					resource.Schema) == nil
-			}
-			canUpdateCluster := canUpdateClusterFn(request, resource.ID)
+			canUpdateCluster := canUserUpdateCluster(request, resource)
 			logrus.Debugf("isActiveCluster: %v isWindowsCluster: %v user: %v, canUpdateCluster: %v", isActiveCluster, isWindowsCluster, request.Request.Header.Get("Impersonate-User"), canUpdateCluster)
 			if canUpdateCluster {
 				resource.AddAction(request, v3.ClusterActionRunSecurityScan)
--- a/pkg/api/customization/cluster/formatter.go
+++ b/pkg/api/customization/cluster/formatter.go
@@ -51,22 +51,28 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 		if resource.Values["state"] == "active" {
 			isActiveCluster = true
 		}
-		canUpdateClusterFn := func(request *types.APIContext, clusterID string) bool {
-			cluster := map[string]interface{}{
-				"id": clusterID,
-			}
-			return request.AccessControl.CanDo(
-				v3.ClusterGroupVersionKind.Group,
-				v3.ClusterResource.Name,
-				"update",
-				request,
-				cluster,
-				resource.Schema) == nil
+		isWindowsCluster := false
+		if resource.Values["windowsPreferedCluster"] == true {
+			isWindowsCluster = true
 		}
-		canUpdateCluster := canUpdateClusterFn(request, resource.ID)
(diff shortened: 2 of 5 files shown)


pkg/api/customization/cluster/formatter.go
```diff
@@ -47,6 +47,14 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
 			resource.AddAction(request, v3.ClusterActionRestoreFromEtcdBackup)
 		}
+		isActiveCluster := false
+		if resource.Values["state"] == "active" {
+			isActiveCluster = true
+		}
+		isWindowsCluster := false
+		if resource.Values["windowsPreferedCluster"] == true {
+			isWindowsCluster = true
+		}
```
prachidamle: Should we also add the Windows check to the action handler to deny calling this action on the cluster at all

leodotcloud: good point, will add.

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -62,6 +62,10 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
 		return httperror.WrapAPIError(err, httperror.ClusterUnavailable,
 			fmt.Sprintf("cluster not ready"))
 	}
+	if cluster.Spec.WindowsPreferedCluster {
+		return httperror.WrapAPIError(err, httperror.InvalidAction,
+			fmt.Sprintf("cannot run scan on a windows cluster"))
+	}
```
prachidamle: @leodotcloud do you think we can move this check at top and avoid all other checks sice with Windows we can right away exit

leodotcloud: 👍

prachidamle:
Also, is there any way to add integration tests for these APIs?

pkg/api/customization/cluster/formatter.go
```diff
@@ -60,8 +68,8 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 				resource.Schema) == nil
 		}
 		canUpdateCluster := canUpdateClusterFn(request, resource.ID)
-		logrus.Debugf("user: %v, canUpdateCluster: %v", request.Request.Header.Get("Impersonate-User"), canUpdateCluster)
-		if canUpdateCluster {
+		logrus.Debugf("isActiveCluster: %v isWindowsCluster: %v user: %v, canUpdateCluster: %v", isActiveCluster, isWindowsCluster, request.Request.Header.Get("Impersonate-User"), canUpdateCluster)
```
kinarashah: canUpdateClusterFn is an extra call, we should skip it after checking isActive  and isWindows

pkg/api/customization/clusterscan/clusterscan.go
```diff
@@ -21,7 +21,11 @@ const (
 
 func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 	if err := apiContext.AccessControl.CanDo(mgmtv3.ClusterScanGroupVersionKind.Group, mgmtv3.ClusterScanResource.Name, "read", apiContext, resource.Values, apiContext.Schema); err == nil {
-		status := resource.Values["status"].(map[string]interface{})
+		s, ok := resource.Values["status"]
+		if !ok {
+			return
+		}
+		status := s.(map[string]interface{})
```
kinarashah: nit: we also have convert.ToMapInterface

pkg/api/customization/cluster/formatter.go
leodotcloud: We need to check RBAC before doing anything else. Otherwise we would be leaking information.

pkg/api/customization/clusterscan/clusterscan.go
leodotcloud: 👍 will change

pkg/api/customization/cluster/formatter.go
kinarashah: This is just formatter, so I don't think we'd be leaking information here by checking RBAC only if cluster is active and not windows

leodotcloud: oops, you are right. I was looking at actions_cis.go file in Goland, will change.

leodotcloud:
@prachidamle added a test for making sure `runSecurityScan` is disabled for non-active clusters. I could have added a similar windows test, but it wouldn't be anyways be able to run coz the cluster is never active so the test wouldn't be testing the right condition anyways.

leodotcloud:
@prachidamle added a test for making sure `runSecurityScan` is disabled for non-active clusters. I could have added a similar windows test, but it wouldn't be anyways be able to run coz the cluster is never active so the test wouldn't be testing the right condition anyways.

tests/integration/suite/test_cluster_scan.py
```diff
@@ -0,0 +1,18 @@
...
+        rancherKubernetesEngineConfig={
+            "accessKey": "junk"
+        }
+    )
+    _, cluster_client = cluster_and_client(cluster.id, client)
+    try:
+        admin_mc.client.action(obj=cluster, action_name="runSecurityScan", )
+    except AttributeError as e:
+        assert e is not None
+    client.delete(cluster)
```
prachidamle: To ensure cluster object is cleaned up no matter of runtime errors, we should call remove_resource() like here https://github.com/rancher/rancher/blob/master/tests/integration/suite/test_clustertemplate.py#L688

prachidamle: also the test seems to have flake8 errors

leodotcloud: @prachidamle fixed.

tests/integration/suite/test_cluster_scan.py
```diff
@@ -0,0 +1,20 @@
...
+        rancherKubernetesEngineConfig={
+            "accessKey": "junk"
+        }
+    )
+    remove_resource(cluster)
+    _, cluster_client = cluster_and_client(cluster.id, client)
+    try:
+        admin_mc.client.action(obj=cluster, action_name="runSecurityScan", )
+    except AttributeError as e:
+        assert e is not None
```
dramich: This check should be specific to the error you are expecting, code and message. Update it to use pytest.raises, as seen [here](https://github.com/rancher/rancher/blob/master/tests/integration/suite/test_app.py#L366)

tests/integration/suite/test_cluster_scan.py
dramich: `client.delete(cluster)` is redundant with the `remove_resource` call above so it can be dropped.

pkg/api/customization/cluster/formatter.go
```diff
@@ -47,22 +47,32 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+		isWindowsCluster := false
+		if resource.Values["windowsPreferedCluster"] == true {
+			isWindowsCluster = true
 		}
-		canUpdateCluster := canUpdateClusterFn(request, resource.ID)
-		logrus.Debugf("user: %v, canUpdateCluster: %v", request.Request.Header.Get("Impersonate-User"), canUpdateCluster)
-		if canUpdateCluster {
-			resource.AddAction(request, v3.ClusterActionRunSecurityScan)
+		if isActiveCluster && !isWindowsCluster {
+			canUpdateClusterFn := func(request *types.APIContext, clusterID string) bool {
```
dramich: I realize this is just being carried over from before but the func isn't being used again and makes the code harder to follow vs just building the structs and making the access call.

tests/integration/suite/test_cluster_scan.py
leodotcloud: Will fix.

tests/integration/suite/test_cluster_scan.py
leodotcloud: @dramich The action is not available, hence the error is "AttributeError". The check that you have pointed will work when the action returns an actual error code/message. Nothing to be changed here.

dramich: Ah, ok, then it should be checking the links to validate it doesn't exist

pkg/api/customization/cluster/formatter.go
leodotcloud: moved it out to a separate function.

tests/integration/suite/test_cluster_scan.py
leodotcloud: I guess you meant actions, not links. Modified, please re-review.

PR #24977, opened 2020-01-16
Istio RBAC Tests

rancher-max:


tests/validation/tests/v3_api/test_istio.py
```diff
@@ -47,17 +65,86 @@ def test_istio_resources():
...
-        "gateways.istio-ingressgateway.ports[0].port": 80,
-        "gateways.istio-ingressgateway.ports[0].targetPort": 80,
-        "gateways.istio-ingressgateway.ports[0].name": "http2",
+        "global.members[0].kind": "Group",
+        "global.members[0].name": "system:authenticated",
+        "kiali.enabled": "true",
+        "tracing.enabled": "true",
+    }
+    update_istio_app(answers)
+
```
sangeethah: @rancher-max can we have each of the validation called out as  a new test case ? You can have the code that does the "allow all" settings moved to a fixture .

sangeethah: Should we also be testing for PROJECT_OWNER and  PROJECT_MEMBER ? Do they get access to in this case ?

tests/validation/tests/v3_api/test_istio.py
```diff
@@ -47,17 +65,86 @@ def test_istio_resources():
...
+    validate_access(kiali_url, cluster_owner)
+    validate_access(tracing_url, cluster_owner)
+    validate_access(grafana_url, cluster_owner)
+    validate_access(prometheus_url, cluster_owner)
+
+    validate_access(kiali_url, cluster_member)
+    validate_access(tracing_url, cluster_member)
+    validate_no_access(grafana_url, cluster_member)
+    validate_no_access(grafana_url, cluster_member)
+
```
sangeethah: Also include test cases for enabling/disable Istio to be allowed only for global admin users.

tests/validation/tests/v3_api/test_istio.py
rancher-max: @sangeethah They do get access, but I didn't include them as I figured CLUSTER_MEMBER was enough to cover for them as it is a member of the cluster but not of any of the projects, so technically has less access overall.

See most recent commit for split up validations. I split it by cluster role. Would you also prefer if I split by each URL check as well? If we do it as granular as by URL, I'd recommend we lump at least kiali + jaeger (tracing) together and grafana + prometheus together as that is how the access is currently granted.

sangeethah: @rancher-max Yes , we can go with the least kiali + jaeger (tracing) together  and  grafana + prometheus together .

tests/validation/tests/v3_api/test_istio.py
rancher-max: Added cases to try to disable and update istio using CLUSTER_MEMBER. Ignoring admin/owner use cases as we install and update istio in the tests using CLUSTER _OWNER token, so if there was an ApiError failure during the tests, we could safely assume it's due to something failing there and don't need an extra test case for it.

sangeethah:
LGTM

Issue #23013, opened 2019-09-23
Rancher 2.4 Scaling Targets


cloudnautique:
# Rancher aims to hit 2,000 clusters and 100,000 nodes under management. 

Based on Kubernetes published limits a single cluster can have 5000 nodes. If we were to run maxed out nodes, we could have (20) clusters with our 100,000 node target. At the opposite end of the spectrum, we would like to achieve a maximum configuration from a cluster stand point we could have 10,000 clusters with 50 nodes each. 

We need to determine in these tests, where etcd tips over. Once determined, we need to prove we can exceed these limits with k3s against alternate backends.

Test cases:
Hit a maximum of 10,000 single node clusters. (maximum of 10,000 clusters)
Attach 400 clusters with 250 nodes per cluster. (100,000 nodes)
Attach 20 clusters each with 5,000 nodes per cluster. (Maximum sized k8s clusters)
Hit a maximum of 2,000 clusters with 50 nodes in each cluster. (max clusters with max nodes)
Attach 1,000 clusters each with 100 nodes per cluster. (Mid point test)

Constraints:
This will be a test of Rancher in HA mode. A single Rancher server node **Should not** need more than 32vCPUs and 128GB of Ram.

For each of these tests:
Clusters will be imported.
All Rancher agents need to be deployed, operational and in a green state from the UI, and maintain status. 
Rancher management pods need to be able to go down to scale 0, and back up to the number of nodes.
Rancher management pods need to be upgraded from one version to the next. 
Rancher should be able to deploy a multi-cluster app to all clusters.
Rancher needs to be able to support 100 concurrent cluster provisions.

API SLOs / SLAs:
Any single object DELETE,POST,GET,PUT response should return in <200ms, in cases of large objects TTFB can be used.
Any list API call <5s
UI Load should <10s

## Deliverables:
- Enable Rancher server level prometheus metrics for increased visibility in the the Rancher controllers. This can be based on work of one of our community members: https://github.com/rancher/rancher/pull/21351
  - This should produce a Grafana dashboard for monitoring Rancher servers with Rancher metrics if deployed into the local cluster.

###
- UI Should introduce a concept of tagging/grouping to accommodate management of nodes and clusters.

This issue will likely spawn additional issues to address bottlenecks.

dramich:
On restart at 2k clusters rancher should:

[...]

PR #25006, opened 2020-01-17
Dedupe management

dramich:
Dedupe the management context and allow handlers to be added on the fly. 

This fix has two parts:

1. Don't create a new management context for every cluster created. This causes a linear growth in memory consumption per cluster along with running n number of additional management controllers.

2. Be able to add handlers on the fly. This was not supported previously so all controllers had to be registered before startup. Now a handler can be added at any time and should register and start processing with no issue. 

Issue: https://github.com/rancher/rancher/issues/23013

Relies on:
https://github.com/rancher/norman/pull/332
https://github.com/rancher/types/pull/1070


vendor/github.com/rancher/norman/controller/generic_controller.go
```diff
@@ -124,26 +124,37 @@ func (g *genericController) Enqueue(namespace, name string) {
...
 	h := &handlerDef{
 		name:       name,
 		generation: g.generation,
 		handler:    handler,
 	}
-	g.handlers = append(g.handlers, h)
-	g.Unlock()
 
-	go func() {
+	go func(gen int) {
```
daxmc99: What is the purpose of generation?

dramich: It's part of the ID and stops from duplicating the same handler

vendor/github.com/rancher/types/config/context.go
```diff
@@ -465,9 +474,10 @@ func NewUserContext(scaledContext *ScaledContext, config rest.Config, clusterNam
 
 func (w *UserContext) Start(ctx context.Context) error {
 	logrus.Info("Starting cluster controllers for ", w.ClusterName)
-	controllers := w.Management.controllers()
-	controllers = append(controllers, w.controllers()...)
-	return controller.SyncThenStart(ctx, 5, controllers...)
+	if err := controller.SyncThenStart(w.runContext, 50, w.Management.controllers()...); err != nil {
```
leodotcloud: Curious to know what these magic numbers 50 and 5 are?

vendor/github.com/rancher/types/config/context.go
```diff
@@ -180,6 +187,7 @@ type UserContext struct {
 	UnversionedClient rest.Interface
 	APIExtClient      clientset.Interface
 	K8sClient         kubernetes.Interface
+	runContext        context.Context
```
daxmc99: What is the purpose of a this new Context?

vendor/github.com/rancher/types/config/context.go
dramich: threadiness i.e. how many goroutines to spin up to process work for those queues.

vendor/github.com/rancher/types/config/context.go
dramich: It's not new, it's the main context rancher starts with and needs to be passed into the management controller startup, if using the user context then when a cluster is deleted management controllers would also shut down (was a fun issue to find)

dramich: To clarify, it's not a context like Ranchers (management, cluster, project) that's a go context usually seen as `ctx`

vendor/github.com/rancher/norman/controller/generic_controller.go
```diff
@@ -124,26 +124,37 @@ func (g *genericController) Enqueue(namespace, name string) {
...
-				handlers = append(handlers, h)
+			if handler.generation == gen {
+				continue
 			}
+			newHandlers = append(newHandlers, handler)
+		}
+		g.handlers = newHandlers
+	}(h.generation)
+
+	g.handlers = append(g.handlers, h)
```
daxmc99: Trying to wrap my head around this. It looks like we run through this sequentially and with the locking there is no way we could run through this sequentially. 

So, first time we run through the function we 
1. Acquire the lock 
2. Add the handlers to `g.Handlers` (see above) 
3. Check for sync
4. Unlock
4. Now, we `goto` line 138 and acquire the lock 
5. We iterate over the list of handlers skipping those that have the same generation
6. We overwrite `g.handlers` on  line 147 with handlers whose `generation != gen`

I think this could be simplified out. I'm not sure why we are using the goroutine here if it cannot execute concurrently due to the locking.

vendor/github.com/rancher/types/config/context.go
daxmc99: Ahh, its the context from https://github.com/rancher/rancher/blob/2c9f04baf3afcba07bfed94cf41a606a2ebd2a71/main.go#L204

vendor/github.com/rancher/norman/controller/generic_controller.go
daxmc99: https://play.golang.org/p/w6vzffRUqkl for testing

vendor/github.com/rancher/norman/controller/generic_controller.go
dramich: To clarify, it's for deleting the handler after the context closes

vendor/github.com/rancher/norman/controller/generic_controller.go
daxmc99: Followup based on conversation with @dramich, 
`<- ctx.Done()` is the critical piece here. This code will only run when the calling ctx is canelled. This goroutine essentially functions as _Deregister_ and removes the handler from the controller. 
The code change is required here because we are no longer duplicating the Mangement context (see changes in context.go)  which we were before.
This means that we need to ensure that we are not deleting that points to mgmt context

Issue #25028, opened 2020-01-21
Istio fails to deploy in airgapped setup in Rancher v2.3.4


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create an air gapped Rancher setup using `-e CATTLE_SYSTEM_CATALOG=bundled` option
2. Add a (large) cluster to the setup. I used EC2 and created 1 instance of t2.2xlarge. When adding the cluster in Rancher, be sure to specify all 3 roles.
3. Enable Istio

**Result:**
Istio fails to deploy. All of the images are stuck in `ImagePullBackOff` state, and they are using `docker.io` instead of the private registry.

![image](https://user-images.githubusercontent.com/55997940/72765679-da3aff00-3baa-11ea-9675-31ed5fe061a3.png)

**Other details that may be helpful:**
The istio images are not present in `rancher-images.txt`, so are not available in the private registry. Running `docker images` inside of the cluster only retrieves one istio image: `istio-kubectl`.

**Environment information**
- Rancher version: `v2.3.4`
- Installation option (single install/HA): single

**Cluster information**
- 1 Node with all 3 roles (cp, etcd, worker): EC2 t2.2xlarge
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:12:17Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Docker version (use `docker version`):

```
Client: Docker Engine - Community
 Version:           19.03.2
 API version:       1.40
 Go version:        go1.12.8
 Git commit:        6a30dfc
 Built:             Thu Aug 29 05:29:11 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.2
  API version:      1.40 (minimum version 1.12)
[...]

PR #25038, opened 2020-01-21
Istio images mirror for airgap install

cbron:
Issue: https://github.com/rancher/rancher/issues/25028


rancher-max:
Also address Issue: https://github.com/rancher/rancher/issues/25051

PR #25040, opened 2020-01-22
Update rke cli files 

soumyalj:
Updated few commands
Updated requirements.txt


soumyalj:
@sangeethah : Please review

PR #25043, opened 2020-01-22
Automate HA install w/ Helm 3 and supplied RKE version

bmdepesa:
Creates load balancer, target groups, nodes, and creates/updates hostname in Route53.

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -1,0 +1,184 @@
+import os
+import rancher
+import time
+
+from .common import create_config_file
+from .common import create_user
+from .common import random_test_name
+from .common import readDataFile
+from .common import run_command_with_stderr
+from .common import set_url_password_token
+
+from lib.aws import AmazonWebServices
+
+DATA_SUBDIR = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                           'resource')
+
+RANCHER_CHART_VERSION = os.environ.get("RANCHER_CHART_VERSION")
+test_run_id = random_test_name("auto")
+# if hostname is not provided, generate one ( to support onTag )
+RANCHER_HOSTNAME_PREFIX = os.environ.get("RANCHER_HOSTNAME_PREFIX", test_run_id)
+resource_suffix = test_run_id + "-" + RANCHER_HOSTNAME_PREFIX
+RANCHER_HA_HOSTNAME = RANCHER_HOSTNAME_PREFIX + ".qa.rancher.space"
+RANCHER_IMAGE_TAG = os.environ.get("RANCHER_IMAGE_TAG")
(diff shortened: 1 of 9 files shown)


tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -0,0 +1,181 @@
...
+    out = run_command_with_stderr(helm_init_cmd)
+    print(out)
+
+
+def install_rancher_self_signed():
+    helm_rancher_cmd = \
+        export_cmd + " && helm_v3 install rancher rancher-latest/rancher " + \
+        "--version " + RANCHER_CHART_VERSION + " " \
+        "--namespace cattle-system " + \
+        "--set hostname=" + RANCHER_HA_HOSTNAME
```
sangeethah: @bmdepesa Can we have the rancher repo also parameterized so we are able to deploy from rancher-stable/rancher-alpha/rancher-latest ?

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -0,0 +1,181 @@
...
+from lib.aws import AmazonWebServices
+
+DATA_SUBDIR = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                           'resource')
+
+RANCHER_CHART_VERSION = os.environ.get("RANCHER_CHART_VERSION")
+test_suffix = random_test_name("auto-ha")
+temp_hostname = test_suffix + ".qa.rancher.space"
+# if hostname is not provided, generate one ( to support onTag )
+RANCHER_HA_HOSTNAME = os.environ.get("RANCHER_HA_HOSTNAME", temp_hostname)
```
sangeethah: @bmdepesa Can the test_suffix be an environment variable that user can set and be used for both nlb and route53 entry that gets created ? ".qa.rancher.space" can be got as another environment variable .

bmdepesa: `qa.rancher.space` will never change as we use the hosted zone in Route53; I can parameterize the suffix and use that in place of the full host name (doing the concatenation in the test)

tests/validation/tests/v3_api/test_create_ha.py
bmdepesa: Added `RANCHER_HELM_REPO` -- user can supply one of `latest, stable, alpha`

tests/validation/tests/v3_api/test_create_ha.py
bmdepesa: Job now takes `RANCHER_HOSTNAME_PREFIX` and resources are now in format:
```
nlb-auto-21782-prefix
tg-80-auto-21782-prefix
tg-443-auto-21782-prefix
https://prefix.qa.rancher.space
```

sangeethah:
LGTM

Issue #25048, opened 2020-01-22
Grab project logs from start of file and not end with the fluentd collector


rayisbadat:
**What kind of request is this (question/bug/enhancement/feature request):**
Feature Request/Enhancement

**Steps to reproduce (least amount of steps as possible):**
1) Setup Project Logging via Rancher to remote collector.  Doesn't matter type.
2) Set flush interval to 1 second.
2) Deploy a workflow that generates lots of output ( I used my test image rayisbadat/infloop )
  
**Result:**
There is a delay before the fluentd log collector starts forwarding logs.  Delay ranges from 10-60 seconds, although had some outliers that went two minutes.   This causes us to lose the boot output, which we think will be critical to roll out and production.

**Other details that may be helpful:**
Adding two settings into auto generated `/opt/rke/var/lib/kubelet/pods/UUID/volumes/kubernetes.io~secret/config/project.conf`  will fix this problem in the auto generated  fluentd project.conf file.  

 `skip_refresh_on_startup true` Speeds up the collection and forwarding
 `read_from_head true` Causes fluentd collector to scrape entire file and not start with the end of file.  

PR: https://github.com/rancher/rancher/pull/25049

Timing from some simple tests (time in seconds)

|setting\Run # |1|2|3|4|5|min|avg|mean|max
|--------------|-|-|-|-|-|---|---|----|---
|default|58.203|30.385|23.806|37.457|13.32|13.32|32.6342|30.385|58.203
|skip_refresh_on_startup true|48.933|39.976|10.573|43.578|38.641|10.573|36.3402|39.976|48.933
|read_from_head true|55.677|58.089|56.444|12.87|30.588|12.87|42.7336|55.677|58.089
|Both|15.166|19.32|26.926|33.81|34.274|15.166|25.8992|26.926|34.274

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):v2.3.4
- Installation option (single install/HA): Single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom(Self Hoststed)
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM (Virtual Box), 2 Cores, 4GB of Ram
- Kubernetes version (use `kubectl version`): v1.17.0

- Docker version (use `docker version`):
```
Client: Docker Engine - Community
 Version:           19.03.4
 API version:       1.40
 Go version:        go1.12.10
[...]

PR #25049, opened 2020-01-22
adds skip_refresh_on_startup and read_from_head to fluentd collector

rayisbadat:
…f generated fluentd collector config file. https://github.com/rancher/rancher/issues/25048

### DIFF
--- a/pkg/controllers/user/logging/generator/templatesource.go
+++ b/pkg/controllers/user/logging/generator/templatesource.go
@@ -31,6 +31,8 @@ var SourceTemplate = `
   time_format  %Y-%m-%dT%H:%M:%S
   tag  {{ .ContainerLogSourceTag }}.*
   format  json
+  skip_refresh_on_startup true
+  read_from_head true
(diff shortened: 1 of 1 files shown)


rayisbadat:
I was fixated on project logs at the time, but yeah that makes sense to me.

> Why we only add `read_from_head true` in source-project-container section? How about source-container?

loganhz:
Hi @rayisbadat could you rebase it, please? The CI build failed due to the flaky test which has been fixed in `master` now.

aiwantaozi:
Could you check your commit message and update them, describe the problems and solutions. For example: 
Problem: ....
Solution: ...
Issue: ...

And when the refresh is skip in start up, when will the refresh happen?

Rebase master to solve the test failed. Thanks.

rayisbadat:
I wasn't sure what the difference between issue and problem was.  Do u need the commit message modified for both or just the latest?  Wasn't sure how to amend the first commit.

Issue #25042, opened 2020-01-22
get nodes -o wide returns same address on rancher cluster made from vagrant vms


jascase901:
I think there is a bug when using rancher/rancher in vagrant. 
 
I am using this to build my rancher server
> docker run -d --name $container_name --restart=unless-stopped -p 80:80 -p 443:443 -v /opt/rancher:/var/lib/rancher rancher/rancher:latest

In order to add nodes to my cluster I need to pass parameters to tell the nodes to use the ips
on eth1 instread of eth0. Heres an example

`sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run rancher/rancher-agent:v2.3.3 --server https://192.168.33.11 --token cr6cvjv949wrgf4jgxf5mdw5f9f44p72fbctgn9pjxlk7n9d4ddtzg --ca-checksum de16aef511787012e4ecce624a3b5bec2c312c895d14467dd3529dc35341dcfa --address 192.168.1.33 --internal-address 192.168.1.33 --etcd --controlplane --worker
`
This lets me add nodes to my cluster.

Unfortunately, these parameters aren't 100% honored because the internal ip from 
`get nodes -o wide` for all my
nodes is always the same one on eth0 (10.0.2.15). 

When building a kubelet there should be an argument passed called "node-ip" that
changes the internal ip address. However using docker inspect I can tell that this
argument wasn't passed.

I am guessing that this line is the issue
```

#rancher/vendor/github.com/rancher/rke/cluster/plan.go
if host.Address != host.InternalAddress {
		CommandArgs["node-ip"] = host.InternalAddress
	}

```

My hypotheses is the --address 192.168.33.33 changes the host.address, and the
--internal-address 192.168.33.33  changes the host.internal address. And because I am setting
them both to the same value.  The node-ip arg never gets added to the build kubelet cmd.

I tried building rancher from source to test this out, by typing make. But I don't know how to test my changes.

westonmyers:
Have you tried setting address to your interface name? Looking at the entrypoint for `rancher/rancher-agent`, it looks like you can just set `--address eth1`. (Looks like that's reinforced in [Rancher docs](https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/custom-nodes/agent-options/) too.)

I'm not saying it'll fix your issue, I'm just curious.

jascase901:
[...]

PR #25068, opened 2020-01-23
always pass node-ip parameter to kubelet

jascase901:
this fixes issue #25042. Where I started my rancher with the "internal-address"
and "address" both set to the ip address on eth1(192.168.33.0/24 in my case), to
make it not auto choose the ip adress eth0(10.0.2.15) which in my case was used
for NAT on vagrant. Because "address" and "internal-address" were both set to
the same value the "node-ip" parameter was never passed to the kublet, and the
kublet autoresolved its ip to the one on eth0.


superseb:
Vendored files need to be modified in their repositories, in this case `rancher/rke`. This was also considered in https://github.com/rancher/rke/issues/1533 but changing this could be breaking as specifying IPs that are not present on the node itself (EC2 public for instance), cant be set as `node-ip`. So this needs serious consideration before changing.

jascase901:
I kinda figured I was missing something there. 

What if we added a flag to the hosts struct that gets set to true if the internal address was manually assigned?

Issue #23923, opened 2019-11-05
cluster-roletemplate-sync runs excessively 


dramich:
This controller runs excessively and appears to be a good candidate to stop it from being a lifecycle and move it over to an owner reference relationship. 

`pkg/controllers/user/rbac/roletemplate_handler.go`

![Screen Shot 2019-11-05 at 4 45 51 PM](https://user-images.githubusercontent.com/13897607/68255901-c17b6e00-ffeb-11e9-947e-c9a969d0e8e3.png)

bmdepesa:
rancher/rancher:master-head 2/11

These issues re: scalability are hard to reproduce without testing at scale, so focused on regressions via automation: 

* Deployed Rancher
* Deployed multiple custom clusters
* Ran automation smoke tests and long running tests for a total of ~400 tests
* Automation runs succeeded with only a few known failures

PR #25136, opened 2020-01-28
Move roletemplate out of cluster lifecycle

dramich:
Problem:
cluster-roletemplate-sync was cluster scoped which means with 2k clusters there will 2k annotations and 2k finalizers on each roleTemplate. This causes slowdown when listing and also makes the object impossible to look at in general. 

Solution:
Move cluster-roletemplate-sync to a management level lifecycle to handle the downstream cluster deletion of the cluster roles and setup a cluster scoped handler to handle syncing the roleTemplate to clusterTemplate on changes.

issue: https://github.com/rancher/rancher/issues/23923


pkg/controllers/management/auth/role_template_handler.go
```diff
@@ -1,47 +1,105 @@
...
+			continue
+		}
+
+		err = userContext.RBAC.ClusterRoles("").Delete(b.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			// User context clusterRole doesn't exist
+			if !apierrors.IsNotFound(err) {
+				allErrors = append(allErrors, err)
+			}
+			continue
```
rmweir: nit: probably don't need continue

dramich: It's not technically necessary but it's there to be clear about intentions and if something is added below this it won't cause a bug. grb-sync follows a similar pattern.

dramich:
> Should there be an upgrade test?

The tests that were updated cover upgrade

pkg/api/store/globalrolebindings/globalrolebinding_store.go
```diff
@@ -9,7 +9,10 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const GrbVersion = "field.cattle.io/grbUpgrade"
+const (
+	GrbVersion    = "field.cattle.io/grbUpgradeCluster"
```
dramich: Update to use `cleanup.cattle.io`

pkg/api/store/roletemplate/store.go
```diff
@@ -10,7 +10,10 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
-const RTVersion = "field.cattle.io/rtUpgrade"
+const (
+	RTVersion    = "field.cattle.io/rtUpgradeCluster"
```
dramich: Update to use `cleanup.cattle.io`

pkg/controllers/management/auth/role_template_handler.go
```diff
@@ -1,47 +1,105 @@
...
-	prtbIndexer cache.Indexer
-	prtbClient  v3.ProjectRoleTemplateBindingInterface
+	prtbIndexer    cache.Indexer
+	prtbClient     v3.ProjectRoleTemplateBindingInterface
+	clusters       v3.ClusterInterface
+	clusterManager *clustermanager.Manager
 }
 
-func newRoleTemplateHandler(management *config.ManagementContext) *roleTemplateHandler {
+func newRoleTemplateHandler(management *config.ManagementContext, clusterManager *clustermanager.Manager) v3.RoleTemplateLifecycle {
```
cjellick: Since this is now a lifecycle, can we name the function accordingly?

cjellick: file name as well

cjellick: and the name `roleTemplateHandler` the functions are attached to

dramich:
Offline discussion: update the user controller sync to validate the clusterrole has a roletemplate in management and delete clusterrole if there isn't a corresponding roletemplate

pkg/controllers/user/rbac/handler_base.go
```diff
@@ -199,7 +195,8 @@ func (m *manager) createClusterRole(rt *v3.RoleTemplate) error {
 	logrus.Infof("Creating clusterRole for roleTemplate %v (%v).", rt.DisplayName, rt.Name)
 	_, err := m.clusterRoles.Create(&rbacv1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: rt.Name,
+			Name:        rt.Name,
+			Annotations: map[string]string{clusterRoleOwner: rt.Name},
 		},
```
mrajashree: v1.Roles are also created with the name of the roleTemplate, should we track them and delete when roleTemplate no longer exists by adding annotations for them too here? https://github.com/rancher/rancher/blob/master/pkg/controllers/management/auth/manager.go#L735

dramich: Well, those roletemplates are getting created in management so they could just be an owner reference but I don't know if that is intended behavior and outside the scope of this PR.

pkg/controllers/management/auth/role_template_handler.go
dramich: Renamed everything

PR #25160, opened 2020-01-29
[wip] Shibboleth

luthermonson:


pkg/auth/providers/saml/saml_provider.go
```diff
@@ -192,14 +256,29 @@ func (s *Provider) saveSamlConfig(config *v3.SamlConfig) error {
...
-	_, err = s.authConfigs.ObjectClient().Update(config.ObjectMeta.Name, config)
-	if err != nil {
+		ldapConfig, certpool, err := s.getLDAPConfig()
+		if err != nil {
+			return err
+		}
+
+		ldapConfigCopy := v3.LdapConfig{}
+		ldapConfig.DeepCopyInto(&ldapConfigCopy)
+		s.hasValidLdapConfig = ldaputils.ValidateLdapConfig(ldapConfigCopy, certpool)
```
rmweir: are you sure you don't want to return an error here?

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -141,6 +145,8 @@ func (p *ldapProvider) GetPrincipal(principalID string, token v3.Token) (v3.Prin
 		return v3.Principal{}, err
 	}
 
+	logrus.Debugln(externalID, scope)
```
rmweir: is this going to make sense on its own?

pkg/auth/providers/providers.go
```diff
@@ -148,7 +155,10 @@ func SearchPrincipals(name, principalType string, myToken v3.Token) ([]v3.Princi
 	if providers[myToken.AuthProvider] == nil {
 		return []v3.Principal{}, fmt.Errorf("[SearchPrincipals] authProvider %v not initialized", myToken.AuthProvider)
 	}
-	principals, err := providers[myToken.AuthProvider].SearchPrincipals(name, principalType, myToken)
+
+	var principals []v3.Principal
+	var err error
+	principals, err = providers[myToken.AuthProvider].SearchPrincipals(name, principalType, myToken)
```
rmweir: nit: necessary?

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -120,6 +136,52 @@ func PerformSamlLogin(name string, apiContext *types.APIContext, input interface
 	return nil
 }
 
+func (s *Provider) getLDAPConfig() (*v3.LdapConfig, *x509.CertPool, error) {
+	// TODO See if this can be simplified. also, this makes an api call everytime. find a better way
+	authConfigObj, err := s.authConfigs.ObjectClient().UnstructuredClient().Get(s.name, metav1.GetOptions{})
```
rmweir: nit: newline

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -120,6 +136,52 @@ func PerformSamlLogin(name string, apiContext *types.APIContext, input interface
...
+	storedLdapConfig := &v3.LdapConfig{}
+
+	mapstructure.Decode(storedLdapConfigMap, storedLdapConfig)
+	metadataMap, ok := storedLdapConfigMap["metadata"].(map[string]interface{})
+	if !ok {
+		return nil, nil, fmt.Errorf("failed to retrieve %s metadata, cannot read k8s Unstructured data", s.name)
+	}
+
+	objectMeta := &metav1.ObjectMeta{}
+	mapstructure.Decode(metadataMap, objectMeta)
```
rmweir: should these decodes be error checked?

pkg/auth/providers/ldap/ldap_provider.go
luthermonson: nope, just forgot to remove. thanks

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -223,17 +302,31 @@ func (s *Provider) toPrincipal(principalType string, princ v3.Principal, token *
...
+	}
+
 	return nil, errors.New("Not implemented")
 }
 
 func (s *Provider) SearchPrincipals(searchKey, principalType string, token v3.Token) ([]v3.Principal, error) {
 	var principals []v3.Principal
-
 	if principalType == "" {
 		principalType = "user"
```
mrajashree: I think when UI calls SearchPrincipals api `principalType` is empty. So this would always set `principalType` to user?

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -223,17 +302,31 @@ func (s *Provider) toPrincipal(principalType string, princ v3.Principal, token *
...
 	}
 
+	if principalType == s.groupType && s.hasLdapGroupSearch() {
```
mrajashree: We should call search principal for users too, till now since we couldn't perform search for saml we have been adding anything user types in as a user

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -192,14 +256,29 @@ func (s *Provider) saveSamlConfig(config *v3.SamlConfig) error {
...
+		}
+
+		ldapConfigCopy := v3.LdapConfig{}
+		ldapConfig.DeepCopyInto(&ldapConfigCopy)
+		s.hasValidLdapConfig = ldaputils.ValidateLdapConfig(ldapConfigCopy, certpool)
+		_, err = s.authConfigs.ObjectClient().Update(
+			config.ObjectMeta.Name,
+			&v3.ShibbolethConfig{
+				SamlConfig: samlConfig,
+				LdapFields: ldapConfigCopy.LdapFields,
```
mrajashree: maybe we should save ldap fields only if `hasValidLdapConfig` returns true?

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -243,21 +336,20 @@ func (s *Provider) SearchPrincipals(searchKey, principalType string, token v3.To
...
+	if externalID == "" && principalType == "" {
 		return v3.Principal{}, fmt.Errorf("SAML: invalid id %v", principalID)
 	}
-	principalType := parts[0]
-	externalID := strings.TrimPrefix(parts[1], "//")
-
 	if principalType != s.userType && principalType != s.groupType {
 		return v3.Principal{}, fmt.Errorf("SAML: Invalid principal type")
 	}
+	if principalType == s.groupType && s.hasLdapGroupSearch() {
```
mrajashree: GetPrincipal should also be called for users I think?

pkg/auth/providers/saml/saml_provider.go
luthermonson: I can't because they'll get cleared out and dropped from the UI so the user would need to re-enter them all.

pkg/auth/providers/saml/saml_provider.go
luthermonson: we are only trying to shim in the group functionality to a SAML provider and I'm trying to leave the user and login code mostly alone.

pkg/api/store/auth/store.go
```diff
@@ -23,6 +23,7 @@ var TypeToFields = map[string][]string{
 	client.ADFSConfigType:            {client.ADFSConfigFieldSpKey},
 	client.KeyCloakConfigType:        {client.KeyCloakConfigFieldSpKey},
 	client.OKTAConfigType:            {client.OKTAConfigFieldSpKey},
+	client.ShibbolethConfigType:      {client.ShibbolethConfigFieldSpKey},
```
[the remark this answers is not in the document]
luthermonson: you do!

pkg/auth/providers/saml/saml_client.go
```diff
@@ -222,15 +229,23 @@ func (s *Provider) getSamlPrincipals(config *v3.SamlConfig, samlData map[string]
...
-		for _, group := range groups {
-			group := v3.Principal{
-				ObjectMeta:    metav1.ObjectMeta{Name: s.groupType + "://" + group},
-				DisplayName:   group,
-				Provider:      s.name,
-				PrincipalType: "group",
-				MemberOf:      true,
+		if s.hasLdapGroupSearch() {
+			var err error
+			groupPrincipals, err = s.ldapProvider.SearchPrincipals(userPrincipal.ObjectMeta.Name, "group", v3.Token{})
```
mrajashree: This will eventually call https://github.com/rancher/rancher/blob/master/pkg/auth/providers/ldap/ldap_client.go#L359
Which means we will search for a group by the name `userPrincipal.ObjectMeta.Name`
Instead we should call SearchPrincipals api on each of the groups that we get from the saml assertion

pkg/auth/providers/common/ldap/ldap_util.go
```diff
@@ -57,6 +57,14 @@ func NewLDAPConn(servers []string, TLS bool, port int64, connectionTimeout int64
 	return lConn, nil
 }
 
+func Connect(config *v3.LdapConfig, caPool *x509.CertPool) (*ldapv2.Conn, error) {
+	servers := config.Servers
+	TLS := config.TLS
+	port := config.Port
+	connectionTimeout := config.ConnectionTimeout
+	return NewLDAPConn(servers, TLS, port, connectionTimeout, caPool)
```
mrajashree: nit: we can call NewLDAPConn(config.Servers, config.TLS, config.port, config.connectionTimeout, caPool) directly instead of assigning to new variables?

luthermonson: yup, totally can good idea. wasn't my func initially but will do that.

pkg/auth/providers/common/ldap/ldap_util.go
```diff
@@ -290,3 +308,27 @@ func Min(a int, b int) int {
...
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, err
+	}
+	pool.AppendCertsFromPEM([]byte(cert))
+	return pool, nil
+}
+
+func ValidateLdapConfig(ldapConfig *v3.LdapConfig, certpool *x509.CertPool) (bool, error) {
+	if len(ldapConfig.Servers) != 1 {
```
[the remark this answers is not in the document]
luthermonson: we do in other places, probably overkill for this func

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -61,6 +69,21 @@ func Configure(ctx context.Context, mgmtCtx *config.ScaledContext, userMGR user.
...
 		groupType:   name + "_group",
 	}
+
+	if samlp.hasLdapGroupSearch() {
+		samlp.ldapProvider = ldap.Configure(ctx, mgmtCtx, userMGR, tokenMGR, name)
+
+		// ignoring errors during boot if not configured
+		ldapConf, certpool, err := samlp.getLDAPConfig()
+		if err != nil {
+			logrus.Infof("getting %s ldap config failed: %s", samlp.name, err)
```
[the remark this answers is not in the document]
luthermonson: it's possible that someone has SAML setup but not ldap, i dont want to log at alll because it could make someone think they have to set this up when in reality it's optional

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -141,6 +145,47 @@ func (p *ldapProvider) GetPrincipal(principalID string, token v3.Token) (v3.Prin
...
+			userPrincipals, err := p.searchUser(externalID, config, lConn)
+			if err != nil {
+				return v3.Principal{}, err
+			}
+
+			if len(userPrincipals) != 1 {
+				return v3.Principal{}, fmt.Errorf("get principal for %s returned %d results",
+					externalID, len(userPrincipals))
+			}
+			return userPrincipals[0], nil
```
mrajashree: we should call `isThisUserMe` on the user

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -245,3 +278,11 @@ func (p *ldapProvider) getDNAndScopeFromPrincipalID(principalID string) (string,
 	externalID := strings.TrimPrefix(parts[1], "//")
 	return externalID, scope, nil
 }
+
+// if provider only enabled by another provider using ldap for group search e.g. SAML
+func (p *ldapProvider) groupSearchProvider() bool {
```
mrajashree: now that we use ldap for searching users too, can we rename `groupSearchProvider` to something that says that ldap is being used for saml search? maybe `calledForSamlSearch` ?

pkg/auth/providers/ldap/ldap_client.go
```diff
@@ -358,6 +357,7 @@ func (p *ldapProvider) searchPrincipals(name, principalType string, config *v3.L
 	if principalType == "" || principalType == "group" {
 		groupPrincipals, err := p.searchGroup(name, config, lConn)
 		if err != nil {
+			logrus.Debugf("group search failed %s", err)
```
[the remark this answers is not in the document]
luthermonson: yup sure

pkg/auth/providers/ldap/ldap_provider.go
luthermonson: samlSearchProvider

pkg/auth/providers/ldap/ldap_provider.go
luthermonson: yup done

Issue #25097, opened 2020-01-24
Automation - Implement RBAC cases for Istio CRDs


rancher-max:
Need to create automated tests to validate all Istio CRDs have proper RBAC. Cluster Owner, Project Owner, and Project Member should have access to read, create new, edit, and delete. Read Only user should only have access to read. Cluster Member should not have ANY access, including read.

CRD List:
```
policies.authentication.istio.io 
adapters.config.istio.io               
attributemanifests.config.istio.io     
handlers.config.istio.io   
httpapispecbindings.config.istio.io    
httpapispecs.config.istio.io    
instances.config.istio.io   
quotaspecbindings.config.istio.io   
quotaspecs.config.istio.io            
rules.config.istio.io                   
templates.config.istio.io       
destinationrules.networking.istio.io   
envoyfilters.networking.istio.io       
gateways.networking.istio.io       
serviceentries.networking.istio.io        
sidecars.networking.istio.io         
virtualservices.networking.istio.io                           
rbacconfigs.rbac.istio.io               
servicerolebindings.rbac.istio.io      
serviceroles.rbac.istio.io            
authorizationpolicies.security.istio.io

certificates.certmanager.k8s.io 
challenges.certmanager.k8s.io    
issuers.certmanager.k8s.io  
orders.certmanager.k8s.io  
```

For now can comment out CRDs that are known to fail. Refer to Issue: #25096

PR #25163, opened 2020-01-29
Add crd rbac tests and custom answers test

rancher-max:
Closes: #25097 

CRD RBAC tests added. Left the known failing ones commented out, and ignored the `certmanager` CRDs (that are also known to be failing in current state). These should be added when all of those are fixed.

Also created a test for validating all of the possible custom answers in Istio. It validates a few things:
1. Expected workloads, by name, are deployed successfully and all their pods are running.
2. There are no additional workloads deployed.
3. All of the workload containers have an image that starts with `rancher/` which is important for airgap setups.

### DIFF
--- a/tests/validation/tests/v3_api/test_istio.py
+++ b/tests/validation/tests/v3_api/test_istio.py
@@ -1,5 +1,6 @@
 import copy
 import os
+import re
 
 import pytest
 import time
@@ -8,15 +9,19 @@ from subprocess import CalledProcessError
 
 from rancher import ApiError
 
+from .common import apply_crd
 from .common import check_condition
+from .common import compare_versions
 from .common import CLUSTER_MEMBER
 from .common import CLUSTER_OWNER
 from .common import create_kubeconfig
 from .common import create_project_and_ns
 from .common import create_ns
 from .common import DEFAULT_TIMEOUT
+from .common import delete_crd
 from .common import execute_kubectl_cmd
 from .common import get_cluster_client_for_token
+from .common import get_crd
 from .common import get_project_client_for_token
 from .common import get_user_client
 from .common import get_user_client_and_cluster
@@ -25,10 +30,13 @@ from .common import PROJECT_MEMBER
 from .common import PROJECT_OWNER
 from .common import PROJECT_READ_ONLY
 from .common import random_test_name
+from .common import rbac_get_kubeconfig_by_role
+from .common import rbac_get_namespace
 from .common import rbac_get_user_token_by_role
 from .common import requests
 from .common import run_command as run_command_common
 from .common import USER_TOKEN
+from .common import validate_all_workload_image_from_rancher
 from .common import wait_for_condition
 from .common import wait_for_pod_to_running
 from .common import wait_for_pods_in_workload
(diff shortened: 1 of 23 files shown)


tests/validation/tests/v3_api/test_istio.py
```diff
@@ -326,18 +428,70 @@ def validate_no_access(url, user):
     return response
 
 
-def update_istio_app(answers, user):
+def validate_all_workloads(deployment_list=None, daemonset_list=None,
```
bmdepesa: I think we should move this to `common.py` so it can be used for monitoring, logging, etc..

tests/validation/tests/v3_api/test_istio.py
```diff
@@ -326,18 +428,70 @@ def validate_no_access(url, user):
...
+    if app is None:
+        app = namespace["istio_app"]
+    if ns is None:
+        ns = namespace["system_ns"]
+    if project is None:
+        project = namespace["system_project"]
     p_client = get_system_client(user)
     updated_answers = copy.deepcopy(DEFAULT_ANSWERS)
     updated_answers.update(answers)
     external_id = "catalog://?catalog=system-library" + \
```
bmdepesa: we use this url in 2 places -- can you make it a variable?

tests/validation/tests/v3_api/test_istio.py
```diff
@@ -310,6 +356,62 @@ def test_rbac_project_read_istio_disable():
...
+def test_rbac_project_read_crds(enable_certmanager, crd, manifest):
+    kubectl_context = rbac_get_kubeconfig_by_role(PROJECT_READ_ONLY)
+    assert str(apply_crd(manifest, kubectl_context)).startswith(
+        "Error from server (Forbidden)")
+    assert "Forbidden" not in get_crd(crd, kubectl_context)
+    assert str(delete_crd(manifest, kubectl_context)).startswith(
+        "Error from server (Forbidden)")
+
+
+def apply_crd(file, kubectl_context):
```
bmdepesa: move these crd methods to`common.py`, passing in the path as necessary

tests/validation/tests/v3_api/test_istio.py
```diff
@@ -70,6 +103,19 @@ def test_istio_resources():
     create_and_test_bookinfo_routing(app_client, app_ns, gateway_url)
 
 
+def test_istio_custom_answers(enable_all_options):
```
bmdepesa: are these the same deployments if we run this test with an older version of istio? Is there any other impact on running these tests w/ an older version?

tests/validation/tests/v3_api/test_istio.py
bmdepesa: Move these crd methods to `common.py`

tests/validation/tests/v3_api/test_istio.py
bmdepesa: This url is used in 2 places -- can you make it a variable?

tests/validation/tests/v3_api/test_istio.py
bmdepesa: Move this to `common.py` - rename to signify that it's checking `rancher` prefix

tests/validation/tests/v3_api/test_istio.py
rancher-max: I added a `skipif` tag to these now for previous versions. Good catch!

rancher-max:
Found in testing this against clusters without enough resources that when the tests fail due to not enough resources, they fail after either 2 or 4 minutes due to timeouts in `verify_istio_app_ready`. Generally it fails in the `installed` condition, so takes 2 minutes, but that is still unacceptable in my opinion. There are 75 tests here, and if they all fail (which is likely for a cluster without enough resources) then that is 2 and a half hours MINIMUM of waiting for failures.

Going to see what I can do to fix this before this PR is ready to merge I think..

PR #25177, opened 2020-01-30
Add RBAC tests for cert rotation

aaronRancher:
This PR adds RBAC tests for certificate rotation.

### DIFF
--- a/tests/validation/tests/v3_api/test_cert_rotation.py
+++ b/tests/validation/tests/v3_api/test_cert_rotation.py
@@ -1,17 +1,70 @@
 import pytest
 import datetime
 import time
+import os
+import ast
+from .common import rbac_get_user_token_by_role
+from .common import get_client_for_token
 from .common import get_user_client_and_cluster
 from .common import validate_cluster_state
 from .common import get_etcd_nodes
-from .common import wait_for_nodes_to_become_active
-from .common import wait_for_node_status
+from rancher import ApiError
+
 
 # Globals
 # Master list of all certs
 ALL_CERTS = ["kube-apiserver", "kube-controller-manager",
              "kube-node", "kube-proxy", "kube-scheduler",
              "kube-etcd", "kube-ca"]
+TEST_RBAC = ast.literal_eval(os.environ.get('RANCHER_TEST_RBAC', "False"))
+if_test_rbac = pytest.mark.skipif(TEST_RBAC is False,
+                                  reason='rbac tests are skipped')
+CLUSTER_NAME = os.environ.get("RANCHER_CLUSTER_NAME", "")
+# here are all supported roles for RBAC testing
+CLUSTER_MEMBER = "cluster-member"
+CLUSTER_OWNER = "cluster-owner"
+PROJECT_MEMBER = "project-member"
+PROJECT_OWNER = "project-owner"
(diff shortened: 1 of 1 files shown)


tests/validation/tests/v3_api/test_cert_rotation.py
```diff
@@ -1,18 +1,28 @@
 import pytest
 import datetime
 import time
-from .common import get_user_client_and_cluster
-from .common import validate_cluster_state
-from .common import get_etcd_nodes
-from .common import wait_for_nodes_to_become_active
-from .common import wait_for_node_status
+import os
+from .common import *  # NOQA
```
sowmyav27: Why are we importing all instead of the required ones?

tests/validation/tests/v3_api/test_cert_rotation.py
```diff
@@ -1,18 +1,28 @@
...
 
 # Globals
 # Master list of all certs
 ALL_CERTS = ["kube-apiserver", "kube-controller-manager",
              "kube-node", "kube-proxy", "kube-scheduler",
              "kube-etcd", "kube-ca"]
 
+# --------------------- rbac test -----------------------
+@if_test_rbac
+@pytest.mark.parametrize("role", [CLUSTER_MEMBER, PROJECT_OWNER, PROJECT_MEMBER, PROJECT_READ_ONLY])
```
sowmyav27: Can you run the flake8 check on this file?

sowmyav27: Can we also include "CLUSTER_OWNER" ?

aaronRancher: The default tests are already being run by cluster owner. But we can include it in the rbac tests.

tests/validation/tests/v3_api/test_cert_rotation.py
aaronRancher: This is because we end up needing many of the .common global variables like `TEST_RBAC = ast.literal_eval(os.environ.get('RANCHER_TEST_RBAC', "False"))` and `if_test_rbac = pytest.mark.skipif(TEST_RBAC is False,
                                  reason='rbac tests are skipped')`

sowmyav27: okay. But we could still import those rather than the whole file i believe. 
Its just best practice to import what is required rather than the whole file.

aaronRancher: Will do.

tests/validation/tests/v3_api/test_cert_rotation.py
aaronRancher: flake8 has been run. Only issues are that some methods are `undefined or defined from star imports: .common` This is resolved when specifying imports.

tests/validation/tests/v3_api/test_cert_rotation.py
aaronRancher: Only imported what was required.

tests/validation/tests/v3_api/test_cert_rotation.py
aaronRancher: Check for cluster_owner implemented.

Issue #25179, opened 2020-01-30
CIS scan does not run on an rke cluster, security scan pods are not deployed on the cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster.
- Click on run scans.
- The scan is stuck in `creating` state.
- No security scan deployements/pods are seen in Cluster --> System project --> workloads

**Error in rancher logs:**
```
2020/01/30 18:17:28 [ERROR] ClusterScanController c-cs5d8/cis-1580408233747682027 [cisScanHandler] failed with : error fetching default cis config: resource name may not be empty
2020/01/30 18:17:44 [INFO] cisScanHandler: Create: deploying helm chart
2020/01/30 18:17:44 [ERROR] ClusterScanController c-cs5d8/cis-1580408233747682027 [cisScanHandler] failed with : error fetching default cis config: resource name may not be empty
2020/01/30 18:18:16 [INFO] cisScanHandler: Create: deploying helm chart
2020/01/30 18:18:16 [ERROR] ClusterScanController c-cs5d8/cis-1580408233747682027 [cisScanHandler] failed with : error fetching default cis config: resource name may not be empty
2020/01/30 18:19:20 [INFO] cisScanHandler: Create: deploying helm chart
2020/01/30 18:19:20 [ERROR] ClusterScanController c-cs5d8/cis-1580408233747682027 [cisScanHandler] failed with : error fetching default cis config: resource name may not be empty
```

**Expected Result:**
Scan should run on the cluster and report should get generated successfully

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - latest - commit id -`4f1aff9b7`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke DO/AWS
- Kubernetes version (use `kubectl version`):

```
1.17.2
```

sowmyav27:
**Verified on latest master-head - commit id: `48783530d`**

- Run scans on an rke cluster.
- Scan is run successfully.
- And report is generated with the fails/passed/skipped cases.

PR #25183, opened 2020-01-30
use default when k8s version not found

leodotcloud:
Addresses: https://github.com/rancher/rancher/issues/25179

Resolve the short k8s version before, not during fetching the CIS config

### DIFF
--- a/pkg/controllers/user/cis/clusterScanHandler.go
+++ b/pkg/controllers/user/cis/clusterScanHandler.go
@@ -11,6 +11,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/management/kontainerdrivermetadata"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/systemaccount"
+	"github.com/rancher/rke/util"
 	rcorev1 "github.com/rancher/types/apis/core/v1"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	projv3 "github.com/rancher/types/apis/project.cattle.io/v3"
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -84,14 +85,15 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
 	if !v3.ClusterScanConditionCreated.IsTrue(cs) {
 		logrus.Infof("cisScanHandler: Create: deploying helm chart")
 		currentK8sVersion := cluster.Spec.RancherKubernetesEngineConfig.Version
+		shortK8sVersion := util.GetTagMajorVersion(currentK8sVersion)
 		cisConfigParams, err := kontainerdrivermetadata.GetCisConfigParams(
-			currentK8sVersion,
+			shortK8sVersion,
```
dramich: I am not seeing what this changes outside of where util.GetTagMajorVersion is getting called? Is this a refactor to get the debug log to have more information?

leodotcloud: Earlier, `GetCisConfigParams` was figuring out the k8sversion, but now it's just doing a get. Hence, the "default" is fetched with the change in logic.

dramich: But the value `name` being passed into the `GET` is the same in both versions

leodotcloud: https://github.com/rancher/rancher/pull/25183/files#diff-2bb8b7c6b4990566a18c3e0a0f28c683R98

leodotcloud: The value "default" was being used to get the short version and it would return an empty string, which leads to fetching the CRD without a name and the call craps out with the error mentioned in the original issue.

dramich: Ohh this is for the error case where the first call fails then `default` is needed to be passed and not ran through GetTagMajorVersion. Some go tests would really help out here.

leodotcloud: yes, so if we have a cluster with a k8s version that CIS is not aware of, it would use "default"

Issue #25164, opened 2020-01-29
exception in nodesyncer


leodotcloud:
Rancher                             v2.3.3
User Interface                v2.3.25
Helm                                  v2.14.3-rancher1
Machine                            v0.15.0-rancher20-1

```
kubectl logs rancher-747c45b4df-6n62n -n cattle-system -f

(..)

2020/01/27 23:43:59 [ERROR] NodeController c-xdqws/m-03f58edc8f45 recovered from panic "runtime error: invalid memory address or nil pointer dereference". (err=<nil>) Call stack:

goroutine 5057 [running]:

github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime.RecoverFromPanic(0xc002903db0)

        /go/src/github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:158 +0xb5

panic(0x3255340, 0x6bf26d0)

        /usr/local/go/src/runtime/panic.go:522 +0x1b5

github.com/rancher/rancher/pkg/controllers/user/nodesyncer.computeDelta(0xc0010f8b70, 0xc00bd078f0, 0x0, 0x0, 0x0, 0xc007c932f0, 0xc004867f00)

        /go/src/github.com/rancher/rancher/pkg/controllers/user/nodesyncer/nodessyncer.go:250 +0x2c0

github.com/rancher/rancher/pkg/controllers/user/nodesyncer.(*nodesSyncer).updateAnnotations(0xc000130c40, 0xc007b80000, 0xc0077bc480, 0xc0082a03a5, 0xb, 0xc00bd075c0, 0xc00bd07860, 0x1, 0x1, 0xc009ab2fc0, ...)

        /go/src/github.com/rancher/rancher/pkg/controllers/user/nodesyncer/nodessyncer.go:192 +0x69

github.com/rancher/rancher/pkg/controllers/user/nodesyncer.(*nodesSyncer).syncLabels(0xc000130c40, 0xc0063c38e0, 0x16, 0xc0077bc480, 0xc0000d8e00, 0x7f078c8826d0, 0x0, 0x0)

        /go/src/github.com/rancher/rancher/pkg/controllers/user/nodesyncer/nodessyncer.go:228 +0x1f9

[...]

PR #22866, opened 2019-09-14
Fix label/annotation sync logic

ibuildthecloud:
Depends on https://github.com/rancher/types/pull/978

PR #25185, opened 2020-01-31
pass canChangeValuePolicy to avoid panic error

kinarashah:
**Issue:** panic because canChangeValuePolicy is nil
https://github.com/rancher/rancher/issues/25164 

**Solution:** 

I'm using allowAll for annotations because didn't find any checks before refactor PR: https://github.com/rancher/rancher/pull/22866 

nodePlan initially sends the following annotations, https://github.com/rancher/rancher/blob/master/vendor/github.com/rancher/rke/cluster/plan.go#L131
```
"rke.cattle.io/internal-ip": host.InternalAddress
"rke.cattle.io/external-ip": host.Address
```

I don't know when these values on v1.Node could differ from initially set values from nodePlan..


ibuildthecloud:
LGTM

PR #25221, opened 2020-02-03
Create test cases for app deploy

khushboo-rancher:
Test reads json file to get app name and version.
Deploy apps to existing cluster and validate their name, version and answer.
Tested: http://jenkins-poc.rancher.io:8080/view/Khushboo/job/khushboo_app_test/


tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    user_client = cluster_info["user_client"]
+    project_client = cluster_info["project_client"]
+    cluster_client = cluster_info["cluster_client"]
+    cluster = cluster_info["cluster"]
+    project = cluster_info["project"]
+    result = {}
+
+    for app_name, app_version in app_data.items():
+        ns = create_ns(cluster_client, cluster, project, app_name)
```
sowmyav27: Why are we creating a namespace for each app?

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,100 @@
...
+        app_ext_id = create_catalog_external_id('library',
+                                                app_name, app_version)
+        try:
+            answer = get_defaut_question_answers(user_client, app_ext_id)
```
sowmyav27: Why do we need a try block here?

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,100 @@
...
+        except (AssertionError, RuntimeError):
+            answer = None
+            print("Failed to create answer for ", app_name)
+            result[app_name] = 'Failed'
+            traceback.print_exc()
+        if answer is None:
```
sowmyav27: Why are we proceeding if the answer is None? 
Does this mean we create an App with answers=None? 
Is there a scenario where there are no answers passed?

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,100 @@
...
+            continue
+        try:
+            app = project_client.create_app(
+                                    name=random_test_name(),
+                                    externalId=app_ext_id,
+                                    targetNamespace=ns.name,
+                                    projectId=ns.projectId,
+                                    answers=answer)
+            validate_catalog_app(project_client, app, app_ext_id, answer)
+        except (AssertionError, RuntimeError):
```
sowmyav27: Why do you expect the app deployment to fail?

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,100 @@
...
+        try:
+            answer = get_defaut_question_answers(user_client, app_ext_id)
+        except (AssertionError, RuntimeError):
+            answer = None
+            print("Failed to create answer for ", app_name)
+            result[app_name] = 'Failed'
+            traceback.print_exc()
+        if answer is None:
+            continue
+        try:
```
sowmyav27: "Try block" can include just the app creation/deployment.
validate_catalog_app --> waits for the app to get active, which has its own assert statement.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: If we deploy app manually, it creates namespace with same name of app. So I retained the same behavior.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: There are cases where it is fails at runtime, so the try block. It makes sure that the loop doesn't break and test continue with other app.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: If answer is None, this will not deploy the current app but proceed to next app from the json lis.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: To avoid runtime error and to proceed to next app deployment when validation fails

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: It could fail at runtime due to network issue or API not responding, resource constraint etc. So to handle such exceptions, Try and except block are used.

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,81 @@
...
+    creates relevant namespace and deploy them.
+    Validates status of the app, version and answer.
+    """
+    user_client = cluster_info["user_client"]
+    project_client = cluster_info["project_client"]
+    cluster_client = cluster_info["cluster_client"]
+    cluster = cluster_info["cluster"]
+    project = cluster_info["project"]
+
+    ns = create_ns(cluster_client, cluster, project, app_name)
```
jiaqiluo: Can you delete the ns at the end of the function?

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,81 @@
...
+    cluster_client = get_cluster_client_for_token(cluster, USER_TOKEN)
+    project = create_project(user_client, cluster,
+                             random_test_name("App-deployment"))
+    project_client = get_project_client_for_token(project, USER_TOKEN)
+
+    cluster_info["cluster"] = cluster
+    cluster_info["cluster_client"] = cluster_client
+    cluster_info["project"] = project
+    cluster_info["project_client"] = project_client
+    cluster_info["user_client"] = user_client
```
jiaqiluo: Can you add a finalizer to remove the project created for those tests?

tests/validation/tests/v3_api/common.py
```diff
@@ -1603,23 +1619,40 @@ def validate_catalog_app(proj_client, app, external_id, answer=None):
...
+    chart = parameters[len(parameters)-2].split("=")[1] + "-" + \
+            parameters[len(parameters)-1].split("=")[1]
+    app_name = parameters[len(parameters)-2].split("=")[1]
     workloads = proj_client.list_workload(namespaceId=ns).data
     for wl in workloads:
         assert wl.state == "active"
-        assert wl.workloadLabels.chart == chart, \
-            "the chart version is wrong"
+        chart_deployed = get_chart_info(wl.workloadLabels)
+        if app_name in chart_deployed and '-' in chart_deployed:
```
sowmyav27: why are we checking it like this?

tests/validation/tests/v3_api/common.py
```diff
@@ -1603,23 +1619,40 @@ def validate_catalog_app(proj_client, app, external_id, answer=None):
...
+def get_chart_info(workloadlabels):
+    """
+    This method finds either 'chart' tag or
+    'helm.sh/chart' tag from workload API
+    @param workloadlabels:
+    @return: chart value of workload e.g. 'app_name-version'
+    """
+    if "chart" in workloadlabels.keys():
+        return workloadlabels.chart
+    elif "helm.sh/chart" in workloadlabels.keys():
```
sowmyav27: When does this scenario happen?

tests/validation/tests/v3_api/common.py
```diff
@@ -1603,23 +1619,40 @@ def validate_catalog_app(proj_client, app, external_id, answer=None):
...
+        "Incorrect list of parameters from catalog external ID"
+    chart = parameters[len(parameters)-2].split("=")[1] + "-" + \
+            parameters[len(parameters)-1].split("=")[1]
+    app_name = parameters[len(parameters)-2].split("=")[1]
     workloads = proj_client.list_workload(namespaceId=ns).data
     for wl in workloads:
         assert wl.state == "active"
-        assert wl.workloadLabels.chart == chart, \
-            "the chart version is wrong"
+        chart_deployed = get_chart_info(wl.workloadLabels)
```
sowmyav27: when get_chart_info() returns the last scenario result `''` what is expected?

tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -0,0 +1,86 @@
...
+    answer = get_defaut_question_answers(user_client, app_ext_id)
+    try:
+        app = project_client.create_app(
+                                name=random_test_name(),
+                                externalId=app_ext_id,
+                                targetNamespace=ns.name,
+                                projectId=ns.projectId,
+                                answers=answer)
+        validate_catalog_app(project_client, app, app_ext_id, answer)
+    except (AssertionError, RuntimeError):
```
sowmyav27: Is there any specific reason why we have this try and except block?

tests/validation/tests/v3_api/common.py
khushboo-rancher: '-' to make sure app name and version exist in chart.
Added a comment in the code

tests/validation/tests/v3_api/common.py
khushboo-rancher: There are "helm.sh/chart" tag instead of 'chart' tag in some cases.

tests/validation/tests/v3_api/common.py
khushboo-rancher: If chart of helm.sh/chart not available, empty is returned and version check is skipped.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: Try block handles the scenario where app needs to be deleted if they fail in validation before next app test starts.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: Added the finalizer.

tests/validation/tests/v3_api/test_catalog_library.py
khushboo-rancher: Deleted the ns at the end

PR #25223, opened 2020-02-03
Added RKE addon status/image verification methods

soumyalj:
In this PR:
1. Removed the pre and post K8s upgrade related env variables from test_upgrade.py
2. In  validate_k8s_service_images, the expected images dictionary is built from the K8s version. 
3. Added addon verification related calls as part of validate_k8s_service_images

### DIFF
--- a/tests/validation/tests/rke/common.py
+++ b/tests/validation/tests/rke/common.py
@@ -1,8 +1,13 @@
 import time
 import os
+import json
 
-k8s_resurce_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), 
-                          "resources/k8s_ymls/")
+k8s_resource_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                               "resources/k8s_ymls/")
+
+# Global expectedimagesdict declared to store the images for a specific
+# k8s Version
+expectedimagesdict = {}
 
 
 def create_and_validate(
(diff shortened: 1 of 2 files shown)


tests/validation/tests/rke/common.py
```diff
@@ -1,8 +1,9 @@
 import time
 import os
+import json
 
 k8s_resurce_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
bmdepesa: typo here should be `k8s_resource_dir`

tests/validation/tests/rke/common.py
```diff
@@ -385,13 +523,15 @@ def teardown(self):
                 'namespace', service_info['namespace'])
 
 
-def validate_k8s_service_images(nodes, expectedimagesdict):
+def validate_k8s_service_images(nodes, k8s_version, rke_client, kubectl):
```
bmdepesa: We call this method ~8 times in the upgrade tests. I suggest storing the images for each k8s version as you grab them, and looking to see if the images are already present before generating the list from `rke` again

tests/validation/tests/rke/common.py
```diff
@@ -146,6 +147,143 @@ def wait_for_etcd_cluster_health(node, etcd_private_ip=False):
...
+                                   "rancher/calico-node")
+    verify_resource_with_kubectl(kubectl, namespace, selector, calico, flannel)
+
+
+def verify_resource_with_kubectl(kubectl, namespace, selector, resource1,
+                                 resource2=None):
+    # Method to verify addon status and images
+
+    # Sleep to allow the components to get to running state
+    time.sleep(30)
```
bmdepesa: Remove this sleep -- we already retry in the loop below to check if the pod is running.

tests/validation/tests/rke/common.py
```diff
@@ -146,6 +147,143 @@ def wait_for_etcd_cluster_health(node, etcd_private_ip=False):
...
+
+    flannel = get_component_version(rke_client, k8sversion,
+                                    "rancher/coreos-flannel")
+
+    calico = get_component_version(rke_client, k8sversion,
+                                   "rancher/calico-node")
+    verify_resource_with_kubectl(kubectl, namespace, selector, calico, flannel)
+
+
+def verify_resource_with_kubectl(kubectl, namespace, selector, resource1,
```
bmdepesa: This name should be updated to better reflect what it's doing, also pass a list of resources instead of just 1 or 2.

tests/validation/tests/rke/common.py
soumyalj: @bmdepesa : In the existing upgrade tests, we had the images stored in a dictionary for a K8s version by obtaining them as env variables. This was a change requested as part of refactoring.

bmdepesa: @soumyalj I understand. I'm saying to save the images after you build the list from rke -- if you need to get the images for the same k8s version in a subsequent, get them from the list you have already populated instead of running `rke config` again to get the same list.

tests/validation/tests/rke/common.py
soumyalj: This sleep is required for the components to get upgraded to the newer version. Else we catch them too early and the images are not upgraded. Will test a little more to see if the sleep time can be reduced.

bmdepesa: If we can't remove it completely, it should be moved outside of this method call.

tests/validation/tests/rke/common.py
soumyalj: Fixed the typo

tests/validation/tests/rke/common.py
soumyalj: Sleep is removed out of this method.

tests/validation/tests/rke/common.py
soumyalj: Renamed to verify_component_status_with_kubectl

tests/validation/tests/rke/common.py
soumyalj: Modified build_expectedimages_dict() to build the dictionary for a specific K8s version only if it does not exist. It will read from the global dictionary once populated.

tests/validation/tests/rke/common.py
```diff
@@ -146,6 +148,152 @@ def wait_for_etcd_cluster_health(node, etcd_private_ip=False):
...
+        for i in range(0, len(args)):
+            print(pod["status"]["containerStatuses"][i]["image"])
+            assert pod["status"]["containerStatuses"][i]["image"] == args[i]
+
+
+def get_system_images(rke_client, k8sversion):
+
+    # Method to obtain the system images for a k8s version  from rke cli
+
+    command = ("rke config --system-images --version " + k8sversion)
```
bmdepesa: This method is called from `get_component_version` which is called many times and will call `rke config` many times -- in this method `get_system_images` you should get the images from the common dictionary you populate with `build_expectedimages_dict`

soumyalj: Used  global systemimagesdict to store the required images list for a specific K8s version.

PR #25290, opened 2020-02-06
Always redeploy agent if agent deployment didn't succeed

ibuildthecloud:


dramich:
https://github.com/rancher/rancher/pull/25290

Issue #24235, opened 2019-11-25
Add User and Group Searching to Ping/OpenLDAP


luthermonson:
Sub Issue for https://github.com/rancher/rancher/issues/14404

When using ping with openldap we will add extra configuration to ping to use an openldap service account for user/group searching.

izaac:
This feature is available from 2.4 master-head commit id: e11d478

Bugs found when testing this feature are being tracked as separate issues.

PR #25299, opened 2020-02-06
Shibboleth Auth Provider + LDAP Search

luthermonson:
Adds a new SAML auth provider for Shibboleth but couples it with OpenLDAP searching for Principals. It operates as a standalone SAML provider with no needed extra configs but if you backend Shibboleth with OpenLDAP you can configure it like a regular OpenLDAP auth provider but only works for get/search so you can find and assign principals to Clusters/Projects/etc.

https://github.com/rancher/rancher/issues/24235
requires: https://github.com/rancher/types/pull/1080


pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -141,6 +145,53 @@ func (p *ldapProvider) GetPrincipal(principalID string, token v3.Token) (v3.Prin
 		return v3.Principal{}, err
 	}
 
+	if p.samlSearchProvider() {
+		config, caPool, err := p.getLDAPConfig()
```
mrajashree: we don't need to call getLDAPConfig here again since it's called on line 138

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -243,22 +334,24 @@ func (s *Provider) SearchPrincipals(searchKey, principalType string, token v3.To
...
 		return v3.Principal{}, fmt.Errorf("SAML: invalid id %v", principalID)
 	}
-	principalType := parts[0]
-	externalID := strings.TrimPrefix(parts[1], "//")
-
 	if principalType != s.userType && principalType != s.groupType {
 		return v3.Principal{}, fmt.Errorf("SAML: Invalid principal type")
 	}
 
+	logrus.Debugf("GET PRINCIPAL %s \n", principalID)
```
mrajashree: nit: is this in all caps by mistake? :)

pkg/auth/providers/ldap/ldap_client.go
```diff
@@ -521,6 +549,5 @@ func (p *ldapProvider) RefetchGroupPrincipals(principalID string, secret string)
 	if err != nil {
```
mrajashree: Since this will generate group principals using ldap's attributes, we'll need to change them according to the saml ones just like you're doing for searchPrincipals

pkg/auth/providers/ldap/ldap_client.go
```diff
@@ -423,8 +433,33 @@ func (p *ldapProvider) searchLdap(query string, scope string, config *v3.LdapCon
...
+		externalId := results.Entries[i].DN
 		entry := results.Entries[i]
-		principal, err := ldap.AttributesToPrincipal(entry.Attributes, results.Entries[i].DN, scope, p.providerName, config.UserObjectClass, config.UserNameAttribute, config.UserLoginAttribute, config.GroupObjectClass, config.GroupNameAttribute)
+
+		if p.samlSearchProvider() {
+			if strings.EqualFold("user", entityType) {
+				userLoginValues := ldap.GetAttributeValuesByName(entry.Attributes, config.UserLoginAttribute)
+				if len(userLoginValues) > 0 {
+					externalId = userLoginValues[0] // only support first
+				}
```
cjellick: so for user lookup to work properly, the user must ensure that the value of the userLoginAttribute field is the same as some value configured in the saml config, right?

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -141,6 +145,46 @@ func (p *ldapProvider) GetPrincipal(principalID string, token v3.Token) (v3.Prin
...
 
+	if p.samlSearchProvider() {
+		lConn, err := ldap.Connect(config, caPool)
+		if err != nil {
+			return v3.Principal{}, nil
+		}
+		defer lConn.Close()
+
+		if scope == p.userScope {
+			userPrincipals, err := p.searchUser(externalID, config, lConn)
```
cjellick: I dont think leveraging searchUser is really the right thing to be doing here. That function is for fuzzy searching, not for hard lookups, which this is. You should use something similar to the look up we do as part of login:

https://github.com/rancher/rancher/blob/7330335a207c6c0ab98d0b8606a5fdf8cde5bd12/pkg/auth/providers/ldap/ldap_client.go#L46-L49

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -141,6 +145,46 @@ func (p *ldapProvider) GetPrincipal(principalID string, token v3.Token) (v3.Prin
...
+			principal := &userPrincipals[0]
+			if p.isThisUserMe(token.UserPrincipal, *principal) {
+				principal.Me = true
+			}
+
+			return *principal, nil
+		}
+
+		if scope == p.groupScope {
+			groupPrincipals, err := p.searchGroup(externalID, config, lConn)
```
cjellick: Same as above. You are doing a hard lookup, not a fuzzy search. Should probably have a more explicit query.

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -31,21 +34,26 @@ const (
...
+	ctx                context.Context
+	authConfigs        v3.AuthConfigInterface
+	secrets            corev1.SecretInterface
+	userMGR            user.Manager
+	tokenMGR           *tokens.Manager
+	serviceProvider    *saml.ServiceProvider
+	name               string
+	userType           string
+	groupType          string
+	certs              string
```
cjellick: anything that is ldap specific, prefix it as such. But also, why arent the certs and caPool part of ldapProvider?

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -120,6 +145,52 @@ func PerformSamlLogin(name string, apiContext *types.APIContext, input interface
 	return nil
 }
 
+func (s *Provider) getLDAPConfig() (*v3.LdapConfig, *x509.CertPool, error) {
```
cjellick: Is this code just all copy and pasted from somewhere else? If yes, why did you need to do that rather than reuse the existing function (if there is no easy way to reuse, thats fine, just want to know).

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -223,13 +300,27 @@ func (s *Provider) toPrincipal(principalType string, princ v3.Principal, token *
 }
 
 func (s *Provider) RefetchGroupPrincipals(principalID string, secret string) ([]v3.Principal, error) {
-	// This should never be called
+	if s.hasValidLdapGroupSearch() {
```
cjellick: as discussed, removed refetch for saml

Issue #25262, opened 2020-02-04
Helm 3 CLI Support


brendarearden:
Add support for Helm 3 in the CLI 

Epic: rancher/rancher#20596

deniseschannon:
Needs to add ability to add catalogs so set what version of helm it is.

brendarearden:
available for testing in rancher/cli v2.4.3-rc1

rancher-max:
This option is available in the CLI, but is not fully working downstream.

Steps:
1. `rancher catalog add --helm-version v3 foo https://git.rancher.io/system-charts`
produces:
```
ID               NAME             URL                                    BRANCH    KIND       HELMVERSION
foo              foo              https://git.rancher.io/system-charts   master    helm       v3
```
But in the UI:
![image](https://user-images.githubusercontent.com/55997940/78943587-0448e280-7a71-11ea-86bf-3af64034e9f4.png)

Also `v3` is not currently a valid helmversion as it requires helm_v3 at the time of this testing. 


**Dependencies: https://github.com/rancher/rancher/issues/26560 and https://github.com/rancher/rancher/pull/26531**

rancher-max:
#### Validated above using `v2.4-2748-head` and `master-2751-head`

Working as expected and using `v3` uses helm_v3 when enabling the apps.

Issue #25261, opened 2020-02-04
Support for multi-cluster app Helm 3 


brendarearden:
Add being able to deploy multi-cluster apps with helm 3 while continue being able to support existing released multi-cluster apps using helm 2 

Epic: rancher/rancher#20596

jiaqiluo:
the feature is validated on master-head 63306835a

- helm 3 catalog can be added and used to install multi-cluster apps
- existing released multi-cluster apps still use helm 2

Detailed tests and results are tacked in the test plan.

Issue #25260, opened 2020-02-04
Add app deployment with Helm 3 and continue supporting existing apps with Helm 2


brendarearden:
Add in being able to deploy helm 3 while continue being able to support deploying existing apps using helm 2

Epic:  rancher/rancher#20596

deniseschannon:
Since we're changing the design to add catalogs as either a Helm 2 or Helm 3 catalog, this issue should test that Helm 2 apps that get upgraded still are marked as Helm 2.

jiaqiluo:
the feature is validated on master-head 6330683

- new catalog can be added as either helm 2 or helm 3 catalog
- the designated helm version is used to deploy, upgrade, rollback, delete apps 
- helm 2 is used for existing apps and catalogs when upgrading from v2.3.5 to master-head

Detailed tests and results are tacked in the test plan.

PR #25300, opened 2020-02-06
[2.4] Add Helm 3 app & multi-cluster deployment support

brendarearden:
**Problem**
Helm released Helm 3 and new applications need to be deployed using helm in order to support the newest features offered by helm 3 charts

**Solution** 
Add helm 3 to rancher while keeping currently deployed apps maintained with helm 2

**Issues** 
#25260 
#25261 
#21070 
#25262 

**Dependent on**
https://github.com/rancher/helm-unittest/pull/6
https://github.com/rancher/types/pull/1095
https://github.com/rancher/kustomizebuild/pull/1
https://github.com/rancher/types/pull/1106

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -11,6 +11,10 @@ ENV CATTLE_CHANNELSERVER_VERSION v0.3.0
 # version used by helm plugin install script
 ENV CATTLE_HELM_UNITTEST_VERSION v0.1.6-rancher1
 ENV GO111MODULE off
+# helm 3 version
+ENV HELM_VERSION v3.1.0
+ENV KUSTOMIZE_VERSION v3.5.4
+
 # kontainer-driver-metadata branch to be set for specific branch other than dev/master, logic at rancher/rancher/pkg/settings/setting.go
 ENV RANCHER_METADATA_BRANCH=dev-v2.4
 
@@ -34,9 +38,10 @@ ENV DOCKER_URL_amd64=https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 \
     DOCKER_URL_arm64=https://github.com/rancher/docker/releases/download/v1.10.3-ros1/docker-1.10.3_arm64 \
     DOCKER_URL=DOCKER_URL_${ARCH}
 
-ENV HELM_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm \
-    HELM_URL_arm64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm-arm64 \
-    HELM_URL=HELM_URL_${ARCH} \
+ENV HELM_URL_V2_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm \
+    HELM_URL_V2_arm64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm-arm64 \
+    HELM_URL_V2=HELM_URL_V2_${ARCH} \
+    HELM_URL_V3=https://get.helm.sh/helm-${HELM_VERSION}-linux-${ARCH}.tar.gz \
     TILLER_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-tiller \
     TILLER_URL_arm64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-tiller-arm64 \
     TILLER_URL=TILLER_URL_${ARCH} \
@@ -48,18 +53,38 @@ ENV HELM_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HE
     CHANNELSERVER_URL=CHANNELSERVER_URL_${ARCH} \
     ETCD_URL_amd64=https://github.com/etcd-io/etcd/releases/download/${CATTLE_ETCD_VERSION}/etcd-${CATTLE_ETCD_VERSION}-linux-amd64.tar.gz \
     ETCD_URL_arm64=https://github.com/etcd-io/etcd/releases/download/${CATTLE_ETCD_VERSION}/etcd-${CATTLE_ETCD_VERSION}-linux-arm64.tar.gz \
-    ETCD_URL=ETCD_URL_${ARCH}
+    ETCD_URL=ETCD_URL_${ARCH} \
+    KUSTOMIZE_URL_amd64=https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz \
+    KUSTOMIZE_URL_arm64=https://github.com/brendarearden/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_arm64.tar.gz \
+    KUSTOMIZE_URL=KUSTOMIZE_URL_${ARCH}
 
-RUN curl -sLf ${!HELM_URL} > /usr/bin/rancher-helm && \
+RUN if [ "${ARCH}" == "arm64" ]; then \
+    curl -sLf ${!KUSTOMIZE_URL} | tar xvzf - --strip-components=1 -C /usr/bin; else \
+    curl -sOL ${!KUSTOMIZE_URL} && \
+    sleep 1 && \
+    tar xzf kustomize_${KUSTOMIZE_VERSION}_linux_${ARCH}.tar.gz -C /usr/bin && \
+    rm kustomize_${KUSTOMIZE_VERSION}_linux_${ARCH}.tar.gz; \
+    fi
+
+# set up helm 2
+RUN curl -sLf ${!HELM_URL_V2} > /usr/bin/rancher-helm && \
     curl -sLf ${!TILLER_URL} > /usr/bin/rancher-tiller && \
-    curl -sLf ${!CHANNELSERVER_URL} > /usr/bin/channelserver && \
-    curl -sLf ${!K3S_URL} > /usr/bin/k3s && \
-    curl -sfL ${!ETCD_URL} | tar xvzf - --strip-components=1 -C /usr/bin/ etcd-${CATTLE_ETCD_VERSION}-linux-${ARCH}/etcd && \
-    chmod +x /usr/bin/rancher-helm /usr/bin/rancher-tiller /usr/bin/k3s /usr/bin/channelserver && \
+    chmod +x /usr/bin/rancher-helm /usr/bin/rancher-tiller && \
     ln -s /usr/bin/rancher-helm /usr/bin/helm && \
     ln -s /usr/bin/rancher-tiller /usr/bin/tiller && \
-    rancher-helm init -c && \
-    rancher-helm plugin install https://github.com/rancher/helm-unittest && \
+    helm init -c && \
+    helm plugin install https://github.com/rancher/helm-unittest
+
+# set up helm 3
+RUN mkdir /usr/tmp && \
+    curl ${HELM_URL_V3} | tar xvzf - --strip-components=1 -C /usr/tmp/ && \
+    mv /usr/tmp/helm /usr/bin/helm_v3 && \
+    chmod +x /usr/bin/kustomize
+
+RUN curl -sLf ${!CHANNELSERVER_URL} > /usr/bin/channelserver && \
+    curl -sLf ${!K3S_URL} > /usr/bin/k3s && \
+    curl -sfL ${!ETCD_URL} | tar xvzf - --strip-components=1 -C /usr/bin/ etcd-${CATTLE_ETCD_VERSION}-linux-${ARCH}/etcd && \
+    chmod +x /usr/bin/k3s /usr/bin/channelserver && \
     mkdir -p /go/src/github.com/rancher/rancher/.kube && \
     ln -s /etc/rancher/k3s/k3s.yaml /go/src/github.com/rancher/rancher/.kube/k3s.yaml && \
     curl -sLf https://releases.rancher.com/kontainer-driver-metadata/${RANCHER_METADATA_BRANCH}/data.json > /root/data.json
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.3
+	github.com/rancher/types => github.com/brendarearden/types v0.0.0-20200303192431-db66b084c040
 
 	k8s.io/api => k8s.io/api v0.17.2
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.17.2
@@ -87,7 +88,7 @@ require (
 	github.com/rancher/security-scan v0.1.7-0.20200222041501-f7377f127168
 	github.com/rancher/steve v0.0.0-20200302231302-020fd6252623
 	github.com/rancher/system-upgrade-controller v0.3.1
-	github.com/rancher/types v0.0.0-20200304001827-068a357fa053
+	github.com/rancher/types v0.0.0-20200303162837-300a04e6f743
 	github.com/rancher/wrangler v0.5.1-0.20200302190048-e60d4be8fc9b
 	github.com/rancher/wrangler-api v0.5.0
 	github.com/robfig/cron v1.1.0
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,6 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.3
-	github.com/rancher/types => github.com/brendarearden/types v0.0.0-20200303192431-db66b084c040
 
 	k8s.io/api => k8s.io/api v0.17.2
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.17.2
@@ -88,7 +87,7 @@ require (
 	github.com/rancher/security-scan v0.1.7-0.20200222041501-f7377f127168
 	github.com/rancher/steve v0.0.0-20200302231302-020fd6252623
 	github.com/rancher/system-upgrade-controller v0.3.1
-	github.com/rancher/types v0.0.0-20200303162837-300a04e6f743
+	github.com/rancher/types v0.0.0-20200305160044-e8e610dd9e01
 	github.com/rancher/wrangler v0.5.1-0.20200302190048-e60d4be8fc9b
 	github.com/rancher/wrangler-api v0.5.0
 	github.com/robfig/cron v1.1.0
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -30,6 +30,8 @@ ENV KUBECTL_VERSION v1.16.1
 ENV DOCKER_MACHINE_LINODE_VERSION v0.1.8
 ENV LINODE_UI_DRIVER_VERSION v0.3.0
 ENV RANCHER_METADATA_BRANCH ${RANCHER_METADATA_BRANCH}
+ENV HELM_VERSION v3.1.0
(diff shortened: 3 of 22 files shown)


package/Dockerfile
```diff
@@ -46,9 +48,10 @@ ENV TINI_URL_amd64=https://github.com/krallin/tini/releases/download/${TINI_VERS
     TINI_URL_arm64=https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-arm64 \
     TINI_URL=TINI_URL_${ARCH}
 
-ENV HELM_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm \
-    HELM_URL_arm64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm-arm64 \
-    HELM_URL=HELM_URL_${ARCH} \
+ENV HELM_URL_2_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-helm \
```
prachidamle: nit: Should this be of the format HELM_2_URL_xyz?

pkg/controllers/user/helm/common.go
```diff
@@ -142,8 +146,12 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
 	if err != nil {
 		return "", "", nil, err
 	}
-
-	commands := append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace}, setValues...)
+	if obj.Spec.HelmVersion != "helm_v3" {
+		helmName = "rancher-helm"
```
prachidamle: This will change the value of the global variable helmName and use "rancher-helm" henceforth. We should use a local scoped variable to decide the helm version for the current obj.Spec

pkg/controllers/user/helm/common.go
```diff
@@ -163,7 +175,11 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
 	}
 
 	// notes.txt
-	commands = append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace, "--notes"}, setValues...)
+	if obj.Spec.HelmVersion != "helm_v3" {
+		commands = append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace, "--notes"}, setValues...)
+	} else {
+		commands = append([]string{"install", obj.Name, appDir, "--dry-run", "--namespace", obj.Spec.TargetNamespace, "--kubeconfig", tempDir.KubeConfigInJail}, setValues...)
+	}
 	cmd = exec.Command(helmName, commands...)
```
prachidamle: Does helmName need to be "rancher-helm" for invoking this command as well, when obj.Spec.HelmVersion != "helm_v3"

pkg/controllers/user/helm/common/common.go
```diff
@@ -124,8 +135,18 @@ func InstallCharts(tempDirs *HelmPath, port string, obj *v3.App) error {
 	if err != nil {
 		return err
 	}
+	fmt.Print(tempDirs)
+	err = createKustomizeYaml(tempDirs, obj.Name)
```
prachidamle: When obj.Spec.HelmVersion != "helm_v3", do we need to createKustomizeYaml? Should this be invoked only for helm_v3?

pkg/controllers/user/helm/common/common.go
```diff
@@ -124,8 +135,18 @@ func InstallCharts(tempDirs *HelmPath, port string, obj *v3.App) error {
...
+	if err != nil {
+		return err
+	}
 	commands := make([]string, 0)
-	commands = append([]string{"upgrade", "--install", "--namespace", obj.Spec.TargetNamespace, obj.Name}, timeoutArgs...)
+	if obj.Spec.HelmVersion != "helm_v3" {
+		helmName = "rancher-helm"
```
prachidamle: This will change the global variable's value

pkg/controllers/user/helm/common.go
```diff
@@ -154,6 +162,10 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
 	if err != nil {
 		return "", "", nil, err
 	}
+	err = l.writeKubeConfig(obj, tempDir.KubeConfigFull, false)
+	if err != nil {
+		return "", "", nil, err
+	}
```
prachidamle: In case of helm2, is this `writeKubeConfig`  needed?

package/Dockerfile
brendarearden: I refactored all the helm 2 and helm 3 to have "v" for version

Dockerfile.dapper
```diff
@@ -40,17 +43,30 @@ ENV HELM_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HE
...
     ETCD_URL_amd64=https://github.com/etcd-io/etcd/releases/download/${CATTLE_ETCD_VERSION}/etcd-${CATTLE_ETCD_VERSION}-linux-amd64.tar.gz \
     ETCD_URL_arm64=https://github.com/etcd-io/etcd/releases/download/${CATTLE_ETCD_VERSION}/etcd-${CATTLE_ETCD_VERSION}-linux-arm64.tar.gz \
-    ETCD_URL=ETCD_URL_${ARCH}
+    ETCD_URL=ETCD_URL_${ARCH} \
+    KUSTOMIZE_URL_amd64=https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz \
+    KUSTOMIZE_URL_arm64=https://github.com/brendarearden/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_arm64.tar.gz \
+    KUSTOMIZE_URL=KUSTOMIZE_URL_${ARCH}
 
-RUN curl -sLf ${!HELM_URL} > /usr/bin/rancher-helm && \
+RUN if [ "${ARCH}" == "arm64" ]; then \
```
cbron: Nit: would be nice to have little one-line comments for these blocks of code that say: `Fetch kustomize for given arch`. Next block can be something like `Setup helm v2/v3, tiller and k3s`. Not necessary but its nice for skimming.

pkg/controllers/user/helm/common/common.go
```diff
@@ -169,7 +204,7 @@ func getTimeoutArgs(app *v3.App) []string {
 	if app.Spec.Wait {
 		timeoutArgs = append(timeoutArgs, "--wait")
 	}
-	timeoutArgs = append(timeoutArgs, "--timeout", strconv.Itoa(app.Spec.Timeout))
+	timeoutArgs = append(timeoutArgs, "--timeout", fmt.Sprintf("%ss", strconv.Itoa(app.Spec.Timeout)))
```
cbron: The double s is for seconds ?

tests/integration/suite/test_rbac.py
```diff
@@ -554,7 +554,7 @@ def test_member_can_perform_app_action(admin_mc, admin_pc, remove_resource,
     app = client.create_app(
         name="test-" + random_str(),
         externalId="catalog://?catalog=library&template"
-                   "=mysql&version=0.3.7&"
+                   "=mysql&version=1.3.1&"
```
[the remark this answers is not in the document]
brendarearden: Updated for version compatibility

Dockerfile.dapper
```diff
@@ -40,17 +43,30 @@ ENV HELM_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HE
     K3S_URL=K3S_URL_${ARCH} \
     ETCD_URL_amd64=https://github.com/etcd-io/etcd/releases/download/${CATTLE_ETCD_VERSION}/etcd-${CATTLE_ETCD_VERSION}-linux-amd64.tar.gz \
     ETCD_URL_arm64=https://github.com/etcd-io/etcd/releases/download/${CATTLE_ETCD_VERSION}/etcd-${CATTLE_ETCD_VERSION}-linux-arm64.tar.gz \
-    ETCD_URL=ETCD_URL_${ARCH}
+    ETCD_URL=ETCD_URL_${ARCH} \
+    KUSTOMIZE_URL_amd64=https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz \
+    KUSTOMIZE_URL_arm64=https://github.com/brendarearden/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_arm64.tar.gz \
```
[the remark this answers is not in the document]
cbron: We'll follow up on this later with linked kustomize-build pr.

package/jailer.sh
```diff
@@ -30,6 +31,9 @@ cp /etc/resolv.conf /opt/jail/$NAME/etc/
 cp /etc/passwd /opt/jail/$NAME/etc/
 cp /etc/hosts /opt/jail/$NAME/etc/
 cp /etc/nsswitch.conf /opt/jail/$NAME/etc/
+cp /bin/cat /opt/jail/$NAME/bin/
+cp /bin/bash /opt/jail/$NAME/bin/
+cp /bin/rm /opt/jail/$NAME/bin/
```
dramich: These are all copies, can we not link them like below with `ssh`?

pkg/controllers/user/helm/common.go
```diff
@@ -142,10 +145,17 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
 	if err != nil {
 		return "", "", nil, err
 	}
-
-	commands := append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace}, setValues...)
-
-	cmd := exec.Command(helmName, commands...)
+	if obj.Status.HelmVersion != "helm_v3" {
```
dramich: can we swap this to be an equals check? And it should use the const `== helmNameV2`

pkg/controllers/user/helm/common.go
```diff
@@ -83,6 +84,8 @@ func helmDelete(tempDirs *common.HelmPath, app *v3.App) error {
 
 func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.HelmPath, error) {
 	var appSubDir string
+	var commands []string
+	var cmd *exec.Cmd
```
dramich: Move these down to above the if block, instantiating where you need them helps when scanning the code

pkg/controllers/user/helm/common.go
```diff
@@ -163,8 +172,13 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
 	}
 
 	// notes.txt
-	commands = append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace, "--notes"}, setValues...)
-	cmd = exec.Command(helmName, commands...)
+	if obj.Status.HelmVersion != "helm_v3" {
```
dramich: Same as above for version check

pkg/controllers/user/helm/common/common.go
dramich: Was this a bug before? It doesn't seem we are distinguishing between helm versions where this is used

pkg/controllers/user/helm/common/common.go
```diff
@@ -249,3 +291,43 @@ func escapeCommas(value string) string {
...
+		}
+	} else {
+		kustomizeSh, err = ioutil.ReadFile("/usr/bin/kustomize.sh")
+		if err != nil {
+			return err
+		}
+	}
+	tmpString := strings.ReplaceAll(string(kustomizeSh), " . ", fmt.Sprintf(" %s ", tempDirs.InJailPath))
+	tmpString = strings.ReplaceAll(tmpString, "all.yaml", filepath.Join(tempDirs.InJailPath, "all.yaml"))
+	err = ioutil.WriteFile(filepath.Join(tempDirs.FullPath, "/kustomize.sh"), []byte(tmpString), 0755)
```
dramich: Is this still needed with fixing up the command call? The script should just be good to go and not require reading and modifying. If anything you can pass a path to the script if required. The file is also linked into this dir so if you modify it with something other than `/` it could be bad times for others.

pkg/controllers/user/helm/common/common.go
```diff
@@ -27,7 +28,9 @@ const (
 	base            = 32768
 	end             = 61000
 	tillerName      = "rancher-tiller"
-	helmName        = "rancher-helm"
+	helmNameV2      = "rancher-helm"
+	helmNameV3      = "helm_v3"
+	helmDefault     = "helm_v3"
```
dramich: Use the const to set this.

pkg/controllers/user/helm/common.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 const (
-	helmName    = "rancher-helm"
 	appLabel    = "io.cattle.field/appId"
 	failedLabel = "io.cattle.field/failed-revision"
+	helmNameV2  = "rancher-helm"
+	helmNameV3  = "helm_v3"
```
dramich: If you export these from common/common.go then you don't need to duplicate them here

Dockerfile.dapper
dramich: Does this get fixed with the move to our own kustomize release? Meaning same code for both architectures

Dockerfile.dapper
```diff
@@ -40,17 +43,30 @@ ENV HELM_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HE
...
-RUN curl -sLf ${!HELM_URL} > /usr/bin/rancher-helm && \
+RUN if [ "${ARCH}" == "arm64" ]; then \
+    curl -sLf ${!KUSTOMIZE_URL} | tar xvzf - --strip-components=1 -C /usr/bin; else \
+    curl -sOL ${!KUSTOMIZE_URL} && \
+    sleep 1 && \
+    tar xzf kustomize_${KUSTOMIZE_VERSION}_linux_${ARCH}.tar.gz -C /usr/bin && \
+    rm kustomize_${KUSTOMIZE_VERSION}_linux_${ARCH}.tar.gz; \
+    fi
+
+RUN curl -sLf ${!HELM_URL_V2} > /usr/bin/rancher-helm && \
```
dramich: Not required but doing one RUN for each helm version would be nice.

Dockerfile.dapper
brendarearden: yes, the arch code will go away

pkg/controllers/user/helm/common.go
```diff
@@ -163,8 +173,14 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
...
 
 	// notes.txt
-	commands = append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace, "--notes"}, setValues...)
-	cmd = exec.Command(helmName, commands...)
+	if obj.Status.HelmVersion == common.HELMNAMEV2 || obj.Status.HelmVersion == "" {
+		commands = append([]string{"template", appDir, "--name", obj.Name, "--namespace", obj.Spec.TargetNamespace, "--notes"}, setValues...)
+		cmd = exec.Command(common.HELMNAMEV2, commands...)
+	}
+	if obj.Status.HelmVersion == common.HELMNAMEV3 {
+		commands = append([]string{"upgrade", "--install", obj.Name, appDir, "--dry-run", "--namespace", obj.Spec.TargetNamespace, "--kubeconfig", tempDir.KubeConfigInJail}, setValues...)
```
cbron: dry run ?

pkg/controllers/user/helm/common/common.go
```diff
@@ -27,7 +28,8 @@ const (
 	base            = 32768
 	end             = 61000
 	tillerName      = "rancher-tiller"
-	helmName        = "rancher-helm"
+	HELMNAMEV2      = "rancher-helm"
```
dramich: Don't make them all caps, `HelmV2` is fine

pkg/controllers/user/helm/common/common.go
```diff
@@ -207,7 +243,13 @@ func GenerateAnswerSetValues(app *v3.App, tempDir *HelmPath, extraArgs map[strin
 }
 
 func DeleteCharts(tempDirs *HelmPath, port string, obj *v3.App) error {
-	cmd := exec.Command(helmName, "delete", "--purge", obj.Name)
+	var cmd *exec.Cmd
+	if obj.Status.HelmVersion == HELMNAMEV2 || obj.Status.HelmVersion == "" {
```
cbron: Feel like since we use this check so many times we could put into helper method: `isHelm3`, `isHelm2`

pkg/controllers/user/helm/common.go
brendarearden: helm 3 removed the template option, so there isn't a great way to create the notes.txt information

brendarearden: the dry run is returning more info than --notes previously did, so I do need to refactor that code. Good catch!

go.mod
```diff
@@ -9,6 +9,7 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.3
+	github.com/rancher/types => github.com/brendarearden/types v0.0.0-20200227175922-fb39870073d5
```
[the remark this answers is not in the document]
brendarearden: https://github.com/rancher/types/pull/1106. << I still have this one open

cbron:
Seems good to me, waiting to approve until types is ready to merge.

pkg/controllers/user/helm/common.go
```diff
@@ -174,13 +190,21 @@ func (l *Lifecycle) generateTemplates(obj *v3.App) (string, string, *common.Helm
...
+		return "", "", nil, errors.Wrapf(err, "start helm template rendering notes failed. %s", sbErr.String())
 	}
 	if err := cmd.Wait(); err != nil {
-		return "", "", nil, errors.Wrapf(err, "wait helm template --notes failed. %s", sbErr.String())
+		return "", "", nil, errors.Wrapf(err, "wait helm template rendering notes failed. %s", sbErr.String())
 	}
 	template := sbOut.String()
 	notes := noteOut.String()
+	if common.IsHelm3(obj.Status.HelmVersion) {
+		splitNotes := strings.SplitAfter(notes, "NOTES:")
```
dramich: Is it guaranteed nothing comes out after notes? Especially things that can leak keys

pkg/controllers/user/helm/controller.go
```diff
@@ -396,3 +404,15 @@ func isSame(obj *v3.App, revision *v3.AppRevision) bool {
 
 	return false
 }
+
+func (l *Lifecycle) getHelmVersion(obj *v3.App) (string, error) {
+	templateVersionID, templateVersionNamespace, err := hCommon.ParseExternalID(obj.Spec.ExternalID)
+	if err != nil {
+		return "", err
+	}
+	templateVersion, err := l.TemplateVersionClient.GetNamespaced(templateVersionNamespace, templateVersionID, metav1.GetOptions{})
```
dramich: This is using a client which is an API call, update to use a lister which will probably require adding one to the lifecycle

pkg/controllers/user/helm/common/common.go
```diff
@@ -249,3 +293,50 @@ func escapeCommas(value string) string {
...
+		return err
+	}
+	kpath := filepath.Join(tempDirs.FullPath, "/kustomization.yaml")
+	err = ioutil.WriteFile(kpath, y, 0644)
+	if err != nil {
+		return err
+	}
+	logrus.Debugf("successfully created kustomization.yaml in %s", kpath)
+
+	// set file link to ensure kustomize.sh works in dev move without haveing to copy
```
dramich: haveing > having

package/jailer.sh
```diff
@@ -55,15 +56,30 @@ cp -r -l /opt/drivers/management-state/bin /opt/jail/$NAME/var/lib/rancher/manag
...
 cp -l /usr/bin/rancher-helm /opt/jail/$NAME/usr/bin
 
+# Hard link helm_3 into the jail
+cp -l /usr/bin/helm_v3 /opt/jail/$NAME/usr/bin
+
 # Hard link tiller into the jail
 cp -l /usr/bin/rancher-tiller /opt/jail/$NAME/usr/bin
 
+# Hard link kustomize into the jail
+cp -l /usr/bin/kustomize /opt/jail/$NAME/usr/bin
```
dramich: kustomize.sh also needs to be added into the jail

pkg/controllers/user/helm/common.go
brendarearden: as you said, based on the source it is the last thing coming out. future changes could make this an issue.

PR #25328, opened 2020-02-08
check the graph etcd-peer-traffic only when there are more than one etcd node

jiaqiluo:
the graph `etcd-peer-traffic ` is available only when there are more than 1 etcd node, so 
- remove it from the list if only 1 etcd node exists
- otherwise, keep it


sangeethah:
LGTM

PR #25350, opened 2020-02-10
Add option to delete rancher sever

sangeethah:


tests/validation/tests/v3_api/Jenkinsfile_deploy_and_test
```diff
@@ -82,11 +86,29 @@ node {
...
+            echo "Error: " + err
+            currentBuild.result = 'FAILURE'
           }
         }
-
+      }
+      }
         stage('Test Report') {
           // copy and archive test results
           sh "docker cp ${setupContainer}:${rootPath}${setupResultsOut} ."
```
bmdepesa: Also copy the reports from the delete container if the server was deployed

sangeethah: Added report for delete container

Issue #25305, opened 2020-02-07
Cluster creation on rancher shows error - 'Failed while: Wait for condition: IntialRolesPopulated:True'


khushboo-rancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Install rancher version - master head- 1a03ecea3 single node
2. Click add cluster - custom, ec2, digital ocean.
3. Fill the details and click save/create.

**Result:**
UI shows error 'Failed while: Wait for condition: IntialRolesPopulated:True'.

**Other details that may be helpful:**
Even though it shows error custom, ec2 clusters get created. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head 1a03ecea3
- Installation option (single install/HA): Single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):v1.16.6
![cluster_creation_ss](https://user-images.githubusercontent.com/60111667/73992066-d9240480-4902-11ea-8987-d3bd35be82c8.png)

khushboo-rancher:
![cluster_creation_ss](https://user-images.githubusercontent.com/60111667/73992032-bb569f80-4902-11ea-8abf-fa9966c79395.png)

khushboo-rancher:
Tested in rancher version master-head - 1fa958804. It is not showing error.
Created a custom cluster, ec2 and digital ocean clusters. They came up active.

PR #25352, opened 2020-02-11
Don't use context based RBAC for watches

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/25305

Issue #25365, opened 2020-02-11
Using user auth token through API to a url allowing Group access fails on first attempt


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
**Prereqs**:
1. Enable authentication. I recreated this with freeIpa, openLdap, and activeDirectory.
2. Create a cluster large enough to enable Istio. I used 3 nodes, all roles, 8GB 4vCPU each.
3. Grant one auth user cluster owner or admin access.
4. Login as that user and enable Istio in the cluster. Provide **group** access to one of your auth groups. Ensure that it is enabled successfully.
5. Save a bearer auth token for this user.

**Issue steps -- NOT THROUGH UI**:
1. Login as user who is part of the auth group. _Save the userId and token from the response._  Example request:
```
POST https://<rancher server url>/v3-public/openLdapProviders/openldap?action=login
Request Body: 
{
    "username": "fakeuser",
    "password": "fakepass",
    "responseType": "json"
}
```
2. Grant user access to cluster: Example request:
```
POST https://<rancher server url>/v3/clusterroletemplatebindings
Bearer Auth Token: <owner/admin token>
Request Body:
{
    "clusterId": "<clusterId>",
    "roleTemplateId": "cluster-member",
    "userId": "<userId from step 1>"
}
```
3. Attempt to access kiali or jaeger url. Example request:
```
GET https://<rancher server url>/k8s/clusters/<cluster id>/api/v1/namespaces/istio-system/services/http:kiali:20001/proxy/
Bearer Auth Token: <token from step 1>
```

**Result:**
User is forbidden on the first try, even though it shouldn't be:
```
{
    "kind": "Status",
    "apiVersion": "v1",
    "metadata": {},
    "status": "Failure",
    "message": "services \"http:kiali:20001\" is forbidden: User \"<userId>\" cannot get resource \"services/proxy\" in API group \"\" in the namespace \"istio-system\"",
    "reason": "Forbidden",
    "details": {
        "name": "http:kiali:20001",
        "kind": "services"
    },
    "code": 403
}
```
_HOWEVER_., all following attempts of step (3) are successful. Can continuously recreate this by deleting the user from the Global scope and redoing steps 1-3.

**Other details that may be helpful:**
Turning on debug logs in rancher server led to this. Break is in between the FORBIDDEN attempt and the subsequent SUCCESS:
```
[...]

PR #25368, opened 2020-02-11
Added initial istio group rbac tests

rancher-max:
These new tests will fail until #25365 is resolved. We can code around it if we think that is the best solution. Supplying a single call to `get_client_for_token` before trying to use the token to validate istio url access is a workaround that I've validated.

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -98,6 +98,12 @@ rbac_data = {
     }
 }
 
+auth_rbac_data = {
+    "project": None,
+    "namespace": None,
+    "users": {}
+}
+
 # here are the global role templates used for
 # testing globalRoleBinding and groupRoleBinding
 TEMPLATE_MANAGE_CATALOG = {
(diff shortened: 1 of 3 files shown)


tests/validation/tests/v3_api/common.py
```diff
@@ -2224,4 +2252,55 @@ def login_as_auth_user(username, password):
...
+    auth_rbac_data["users"][user.username] = user
+
+
+def auth_resource_cleanup():
+    """ remove the project and namespace created for the AUTH tests"""
+    client, cluster = get_global_admin_client_and_cluster()
+    client.delete(auth_rbac_data["project"])
+    auth_rbac_data["project"] = None
+    auth_rbac_data["ns"] = None
+    crtbs = client.list_cluster_role_template_binding(clusterId=cluster.id)
```
sangeethah: Can we list all the cluster role template bindings by  each of the users in auth_rbac_data["users"] and delete them?

sangeethah:
LGTM

Issue #22691, opened 2019-09-06
AWS EC2 Node Driver Create should allow encryption of EBS volumes at launch time


Oats87:
In Rancher 2.2, it is not possible to (at launch time) create encrypted EBS volume backed instances.

Today, when launching a new instance via the AWS Launch Wizard, it is possible to "natively" encrypt the volumes at launch time, rather than having to snapshot, copy, encrypt, then create an AMI.

We should provide the capability to launch EC2 instances (via Node Template) and encrypt the EBS volumes for this instance at launch time.

mrajashree:
UI issue: https://github.com/rancher/rancher/issues/25266

aiyengar2:
PR containing rancher/machine changes: https://github.com/rancher/machine/pull/63

aiyengar2:
PR containing unit test in rancher/rancher to test the above change: https://github.com/rancher/rancher/pull/25370

khushboo-rancher:
Encryption of volumes is not working on upgraded rancher set up from v2.3.5 to v2.3-head. Created a bug https://github.com/rancher/rancher/issues/25546

khushboo-rancher:
Tested below scenarios, feature works fine in rancher version v2.3-head ```b5175db3f``` and master-head ```981263b53``` and in upgraded rancher set up from v2.3.5 to v2.3-head
1. Created EC2 node driver with encrypted EBS option and launched new instance with it.
2. Edit the node template and launched new instance with it.
3. Launched new instances with different node template(with/without encryption enabled) in same cluster.
4. Scale up the instances in existing cluster.
5. Impact on other fields in node template.
6. Create/edit/delete with different roles(Cluster owner, cluster member etc) in rancher.
7. Other functionality tested on instance created with encryption enabled - Workload deployment, ingress, service discovery, records, registry, secrets, connectivity, web socket etc.

PR #25370, opened 2020-02-11
Add unit test for encrypted EBS volume-backed instance

aiyengar2:
This commit contains a unit test that ensures that the amazonec2 driver contains a flag for encrypting an EBS volume.

PR for the change this unit test is testing: https://github.com/rancher/machine/pull/63

Related Issue: https://github.com/rancher/rancher/issues/22691


khushboo-rancher:
On an upgraded rancher server from v2.3.5 to v2.3-head, Encrypt EBS Volume is not working. Created a bug https://github.com/rancher/rancher/issues/25546

PR #25380, opened 2020-02-12
add tests for webSockets

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/25405

add the following tests:
- launching the kubectl
- exec into a pod to run commands
- check a pod's log

introduce one new keyword for automation:
- test_websocket

### DIFF
--- a/tests/validation/requirements_v3api.txt
+++ b/tests/validation/requirements_v3api.txt
@@ -18,4 +18,4 @@ python-digitalocean==1.13.2
 PyYAML==5.2
 requests==2.22.0
 rsa==4.0
-websocket-client==0.53.0
+websocket-client==0.57.0
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -13,6 +13,10 @@ from urllib.parse import urlparse
 from rancher import ApiError
 from lib.aws import AmazonWebServices
 from copy import deepcopy
+from threading import Lock
+from threading import Thread
+import websocket
+import base64
 
 DEFAULT_TIMEOUT = 120
 DEFAULT_MULTI_CLUSTER_APP_TIMEOUT = 300
@@ -2311,3 +2315,52 @@ def auth_resource_cleanup():
         user_crtbs = client.list_cluster_role_template_binding(userId=user.id)
         for crtb in user_crtbs:
             client.delete(crtb)
+
+
+class WebsocketLogParse:
+    """
+    the class is used for receiving and parsing the message
+    received from the websocket
+    """
+    def __init__(self):
+        self.lock = Lock()
+        self._last_message = ''
+
+    def receiver(self, socket, skip):
+        """
+        run a thread to receive and save the message from the web socket
+        :param socket: the socket connection
+        :param skip: if True skip the first char of the received message
+        """
+        while True:
+            try:
+                data = socket.recv()
+                # the message from the kubectl contains an extra char
+                if skip:
+                    data = data[1:]
+                if len(data) < 5:
+                    pass
+                data = base64.b64decode(data).decode()
+                self.lock.acquire()
+                self._last_message += data
+                self.lock.release()
+            except websocket.WebSocketConnectionClosedException:
+                print("Connection closed")
+                break
+
+    @staticmethod
+    def start_thread(target, args):
+        thread = Thread(target=target, args=args)
+        thread.daemon = True
+        thread.start()
+        time.sleep(1)
+
+    @property
+    def last_message(self):
+        return self._last_message
+
+    @last_message.setter
+    def last_message(self, value):
+        self.lock.acquire()
+        self._last_message = value
+        self.lock.release()
--- a/tests/validation/tests/v3_api/test_websocket.py
+++ b/tests/validation/tests/v3_api/test_websocket.py
@@ -1,0 +1,151 @@
+import urllib
+from .common import CATTLE_TEST_URL
+from .common import create_kubeconfig
+from .common import get_user_client_and_cluster
+from .common import get_project_client_for_token
+from .common import WebsocketLogParse
+from .common import USER_TOKEN
+import websocket
+import ssl
+import base64
+import time
+import pytest
+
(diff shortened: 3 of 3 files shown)


tests/validation/tests/v3_api/test_websocket.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def test_websocket_exec_shell():
+    global last_message
+    url = 'wss://' + CATTLE_TEST_URL[8:] + \
+        '/k8s/clusters/' + namespace["cluster"].id + \
+        '/api/v1/namespaces/' + namespace["ns"] + \
+        '/pods/' + namespace["pod"].name + \
+        '/exec?container=' + namespace["pod"].containers[0].name + \
+        '&stdout=1&stdin=1&stderr=1&tty=1&command=%2Fbin%2Fsh&command=-c' \
```
bmdepesa: Is it possible to have the url in plaintext and then urlencode afterwards? it would make it much easier to see the arguments passed into the endpoint

tests/validation/tests/v3_api/test_websocket.py
```diff
@@ -0,0 +1,154 @@
...
+    cmd_enc = base64.b64encode(command.encode('utf-8')).decode('utf-8')
+    ws_connection.send('0' + cmd_enc)
+    # sends the command to the webSocket
+    ws_connection.send('0DQ==')
+    time.sleep(3)
+
+
+def test_websocket_launch_kubectl():
+    global last_message
+    ws = websocket.create_connection(
```
bmdepesa: move connection and receiver daemon spawn to a common function - we use it multiple times w/ similar params

tests/validation/tests/v3_api/test_websocket.py
```diff
@@ -0,0 +1,154 @@
...
+            if skip:
+                data = data[1:]
+            if len(data) < 5:
+                pass
+            data = base64.b64decode(data).decode()
+            last_message += data
+        except websocket.WebSocketConnectionClosedException:
+            print("Connection closed")
+            break
+        except ssl.SSLZeroReturnError:
```
bmdepesa: When does this exception get raised?

dramich:
Setting and writing to a global will probably cause issues if these tests get ran in parallel

bmdepesa:
@dramich our tests do not run in parallel

tests/validation/tests/v3_api/test_websocket.py
jiaqiluo: now the parameters are listed and then added to the url.

tests/validation/tests/v3_api/test_websocket.py
jiaqiluo: I keep this function invocation but wrap the thread creation in a new function.

jiaqiluo: I keep this function invocation but wrap the thread creation in a new function.

tests/validation/tests/v3_api/test_websocket.py
jiaqiluo: I used to see this error but can not reproduce it anymore, so the exception is removed.

jiaqiluo:
@bmdepesa @dramich   A lock is added to make sure the message is modified by only one thread at a time.

dramich:
> @bmdepesa @dramich A lock is added to make sure the message is modified by only one thread at a time.

That is only part of the issue, if you have two tests running in parallel then you have two different things writing to the same place which means the global var is now dirty and has state from another test.

jiaqiluo:
update: added a new class to handle the message receiving and parsing.

tests/validation/tests/v3_api/common.py
```diff
@@ -2311,3 +2315,48 @@ def auth_resource_cleanup():
         user_crtbs = client.list_cluster_role_template_binding(userId=user.id)
         for crtb in user_crtbs:
             client.delete(crtb)
+
+
+class LogParse:
```
bmdepesa: rename to specify this is specific to websockets

tests/validation/tests/v3_api/test_websocket.py
bmdepesa: There's still a lot of duplicate code because we're passing the same options (except for url) in every test.

jiaqiluo: This is sovled by adding a new function

tests/validation/tests/v3_api/common.py
jiaqiluo: Renaming is done.

tests/validation/tests/v3_api/common.py
```diff
@@ -2311,3 +2315,52 @@ def auth_resource_cleanup():
...
+                data = base64.b64decode(data).decode()
+                self.lock.acquire()
+                self._last_message += data
+                self.lock.release()
+            except websocket.WebSocketConnectionClosedException:
+                print("Connection closed")
+                break
+
+    @staticmethod
+    def start_thread(target, args):
```
izaac: If we start threads, shouldn't we also make sure they terminate?

tests/validation/tests/v3_api/common.py
```diff
@@ -2311,3 +2315,52 @@ def auth_resource_cleanup():
...
+                    pass
+                data = base64.b64decode(data).decode()
+                self.lock.acquire()
+                self._last_message += data
+                self.lock.release()
+            except websocket.WebSocketConnectionClosedException:
+                print("Connection closed")
+                break
+
+    @staticmethod
```
izaac: We would probably remove the static of this method and make it non-static if we end up making a `finish_thread` method.

tests/validation/requirements_v3api.txt
```diff
@@ -18,4 +18,4 @@ python-digitalocean==1.13.2
 PyYAML==5.2
 requests==2.22.0
 rsa==4.0
-websocket-client==0.53.0
+websocket-client==0.57.0
```
izaac: We sure no other test was impacted by bumping this version?

jiaqiluo: I did not see any other places under `v3_api` except this PR that uses this library

Issue #25213, opened 2020-02-03
Add dynamic field to features flags


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Desired:**
Add dynamic field to feature flags. If dynamic is false, then Rancher will shutdown if the feature flags value is changed. This is ideal for features that need to be setup prior to runtime. If installed with kubernetes, rancher will reboot automatically.

ibuildthecloud:
This flag should be visible to the UI to indicate that if the user enables it the UI may become on responsive for a short period while the feature is updating.

sowmyav27:
Verified on rancher:master-head latest - commit id: `cc12e1d1d`

- Run docker command - fresh install to bring up server - `docker run -d -p 8080:80 -p 443:443 rancher/rancher:master-head`
- Go to /v3/features
- For `steve` feature flag , `dynamic: false` , change the value to True. 
- Rancher is shut off. 
- The container is in exited state.
- On starting the docker container aggain, UI shows `Steve` as enabled feature flag.
- <rancher-server>/dashboard, shows this
<img width="958" alt="Screen Shot 2020-02-14 at 1 03 28 PM" src="https://user-images.githubusercontent.com/26032343/74567615-0c7e1900-4f2b-11ea-88d2-cf85a1aadc42.png">

- <rancher-server>/v1 shows a bunch of this stuff
<img width="644" alt="Screen Shot 2020-02-14 at 1 05 12 PM" src="https://user-images.githubusercontent.com/26032343/74567621-0e47dc80-4f2b-11ea-8487-8c0f769831b5.png">

sowmyav27:
**Verified on a HA setup - 3 nodes**

- When the value for feature flag "steve" is changed, only the leader rancher server gets restarted.
- Logs in the rancher server:

```
2020/02/15 01:50:52 [INFO] kontainerdriver amazonelasticcontainerservice stopped
2020/02/15 01:50:52 [INFO] dynamic schema for kontainerdriver amazonelasticcontainerservice updating
2020/02/15 02:49:38 [FATAL] feature flag [steve] value has changed, rancher must be restarted
```

rmweir:
@sowmyav27 no it is not. Great catch.

sowmyav27:
**Verified on master-head - commit id: `7d221af9c`**

- HA setup - 3 nodes.

- When the value for feature flag "steve" is changed, all three rancher servers are shut off and restarted
- logs in all the three rancher server show:

[...]

PR #25382, opened 2020-02-12
Dynamic features

rmweir:
**Depends on:**
types (merge first): https://github.com/rancher/types/pull/1089 

**Problem:**
We cannot feature flag functionality that cannot be changed during runtime.

**Solution:**
Make all features either "dynamic" or "non-dynamic". Rancher must be restarted if the value of a "non-dynamic" feature is changed.

**Issue:**
https://github.com/rancher/rancher/issues/25213


pkg/features/feature.go
```diff
@@ -116,29 +145,32 @@ func applyArgumentDefaults(featureArgs string) error {
...
-	featureState, err := f.featuresLister.Get("", f.name)
-	if err != nil {
-		return false
-	}
-	if featureState.Value != nil {
-		return *featureState.Value
-	}
-	return featureState.Default
+func (f *feature) Set(val bool) {
+	defer f.lock.Unlock()
```
ibuildthecloud: I don't think the lock is needed.  If it is you need to lock on read too.

PR #25396, opened 2020-02-13
Adding shibboleth + OpenLDAP Search Auth Provider

luthermonson:
Adds a new SAML auth provider for Shibboleth but couples it with OpenLDAP searching for Principals. It operates as a standalone SAML provider with no needed extra configs but if you backend Shibboleth with OpenLDAP you can configure it like a regular OpenLDAP auth provider but only works for get/search so you can find and assign principals to Clusters/Projects/etc.

This PR explicitly avoids Refetching principals, please see this Issue to track potentially adding that functionality: https://github.com/rancher/rancher/issues/25397

**Overview**

- Added Shibboleth as a new Auth Provider
- Added OpenLdapFields to Shibboleth Auth Type to store params
- Cut into functionality in SAML provider to call out to ldap provider (GetPrincipal/SearchPrincipal)
- Added store/retrieve of secrets from sub field of auth config (see auth/store.go and getLdapConfig)

#24235
requires: rancher/types#1085

### DIFF
--- a/pkg/api/store/auth/store.go
+++ b/pkg/api/store/auth/store.go
@@ -13,18 +13,27 @@ import (
 	client "github.com/rancher/types/client/management/v3"
 )
 
-var TypeToFields = map[string][]string{
-	client.GithubConfigType:          {client.GithubConfigFieldClientSecret},
-	client.ActiveDirectoryConfigType: {client.ActiveDirectoryConfigFieldServiceAccountPassword},
-	client.AzureADConfigType:         {client.AzureADConfigFieldApplicationSecret},
-	client.OpenLdapConfigType:        {client.LdapConfigFieldServiceAccountPassword},
-	client.FreeIpaConfigType:         {client.LdapConfigFieldServiceAccountPassword},
-	client.PingConfigType:            {client.PingConfigFieldSpKey},
-	client.ADFSConfigType:            {client.ADFSConfigFieldSpKey},
-	client.KeyCloakConfigType:        {client.KeyCloakConfigFieldSpKey},
-	client.OKTAConfigType:            {client.OKTAConfigFieldSpKey},
-	client.GoogleOauthConfigType:     {client.GoogleOauthConfigFieldOauthCredential, client.GoogleOauthConfigFieldServiceAccountCredential},
-}
+var (
+	TypeToFields = map[string][]string{
+		client.GithubConfigType:          {client.GithubConfigFieldClientSecret},
+		client.ActiveDirectoryConfigType: {client.ActiveDirectoryConfigFieldServiceAccountPassword},
+		client.AzureADConfigType:         {client.AzureADConfigFieldApplicationSecret},
+		client.OpenLdapConfigType:        {client.LdapConfigFieldServiceAccountPassword},
+		client.FreeIpaConfigType:         {client.LdapConfigFieldServiceAccountPassword},
+		client.PingConfigType:            {client.PingConfigFieldSpKey},
+		client.ADFSConfigType:            {client.ADFSConfigFieldSpKey},
+		client.KeyCloakConfigType:        {client.KeyCloakConfigFieldSpKey},
+		client.OKTAConfigType:            {client.OKTAConfigFieldSpKey},
+		client.ShibbolethConfigType:      {client.ShibbolethConfigFieldSpKey},
+		client.GoogleOauthConfigType:     {client.GoogleOauthConfigFieldOauthCredential, client.GoogleOauthConfigFieldServiceAccountCredential},
+	}
+
+	SubTypeToFields = map[string]map[string][]string{
+		client.ShibbolethConfigType: {
+			client.ShibbolethConfigFieldOpenLdapConfig: {client.LdapConfigFieldServiceAccountPassword},
+		},
+	}
+)
 
 func Wrap(store types.Store, secrets corev1.SecretInterface) types.Store {
 	return &Store{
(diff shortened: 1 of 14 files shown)


pkg/api/store/auth/store.go
```diff
@@ -45,15 +54,50 @@ func (s *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
...
-			if err := common.CreateOrUpdateSecrets(s.Secrets, val, strings.ToLower(field), strings.ToLower(kind)); err != nil {
-				return nil, fmt.Errorf("error creating secret for %s:%s", kind, field)
+	fields, ok := TypeToFields[kind]
+	subFields, subOk := SubTypeToFields[kind]
+	if !ok && !subOk {
+		return s.Store.Update(apiContext, schema, data, id)
+	}
+
+	var err error
+	if ok {
```
mrajashree: nit: we don't need this check and the check for `if subOk`

pkg/api/store/auth/store.go
```diff
@@ -45,15 +54,50 @@ func (s *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
...
+				}
 			}
-			data[field] = fmt.Sprintf("%s:%s-%s", namespace.GlobalNamespace, strings.ToLower(kind), strings.ToLower(field))
 		}
 	}
+
+	// subfields for embedded configs, see saml group search using openldap
+	if subOk {
+		for subField, subFieldList := range subFields {
+			if subData, ok := data[subField]; ok {
```
mrajashree: I think here we can iterate over `subFieldList` and use GetValueN 
```
for _, field := range subFieldList {
	val := values.GetValueN(data, subField, field)
	if val != nil { CreateOrUpdateSecrets..}
}
```

pkg/auth/providers/saml/saml_actions.go
```diff
@@ -46,6 +48,18 @@ func (s *Provider) testAndEnable(actionName string, action *types.Action, reques
 		return err
 	}
 
+	if s.hasLdapGroupSearch() {
+		ldapConfig, certpool, err := ldap.GetLDAPConfig(s.ldapProvider)
+		if err != nil {
+			return err
+		}
```
mrajashree: `hasLdapGroupSearch` only checks the provider name and not the presence of ldapconfig fields, so this will return an error even if user didn't want to configure openldap support?

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -192,14 +220,18 @@ func (s *Provider) saveSamlConfig(config *v3.SamlConfig) error {
 	}
 
 	config.SpKey = common.GetName(config.Type, field)
+	if s.hasLdapGroupSearch() {
```
mrajashree: same question here, `hasLdapGroupSearch` will return true for shibboleth, and `combineSamlAndLdapConfig` will return err if it fails to get ldapConfig, should `hasLdapGroupSearch` check for actual presence of ldapfields?

pkg/auth/providers/saml/saml_actions.go
luthermonson: youre right, should just Debug log here and in the call under it just like we do in the Configure

pkg/api/store/auth/store.go
luthermonson: its an or situation tho, a provider could have secrets in the fields or subfields and since i had both in case the authprovider has no secrets i wrapped both or sections to be safe

pkg/auth/providers/common/ldap/ldap_util.go
```diff
@@ -97,6 +101,7 @@ func IsType(search []*ldapv2.EntryAttribute, varType string) bool {
 		if attrib.Name == "objectClass" {
 			for _, val := range attrib.Values {
 				if strings.EqualFold(val, varType) {
+					logrus.Debugf("Object is type: %s", varType)
```
rmweir: might need some more context in this log

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -178,23 +206,28 @@ func (p *ldapProvider) getLDAPConfig() (*v3.LdapConfig, *x509.CertPool, error) {
...
 	storedLdapConfig := &v3.LdapConfig{}
 
-	mapstructure.Decode(storedLdapConfigMap, storedLdapConfig)
-
-	metadataMap, ok := storedLdapConfigMap["metadata"].(map[string]interface{})
-	if !ok {
-		return nil, nil, fmt.Errorf("failed to retrieve %s metadata, cannot read k8s Unstructured data", p.providerName)
+	if p.samlSearchProvider() && ldapConfigKey[p.providerName] != "" {
+		if _, ok := storedLdapConfigMap[ldapConfigKey[p.providerName]]; ok {
+			storedLdapConfigMap = storedLdapConfigMap[ldapConfigKey[p.providerName]].(map[string]interface{})
```
rmweir: are you fairly sure this is safe?

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -243,5 +267,87 @@ func (p *ldapProvider) getDNAndScopeFromPrincipalID(principalID string) (string,
...
+	if p.providerName == ShibbolethName {
+		return true
+	}
+	return false
+}
+
+func (p *ldapProvider) samlSearchGetPrincipal(
+	externalID string, scope string, config *v3.LdapConfig, caPool *x509.CertPool) (*v3.Principal, error) {
+
+	if (scope != p.userScope) && (scope != p.groupScope) {
```
rmweir: nit: I don't think you need the parentheses

pkg/auth/providers/saml/saml_provider.go
```diff
@@ -278,6 +290,19 @@ func (s *Provider) isThisUserMe(me v3.Principal, other v3.Principal) bool {
 	return false
 }
 
+func (s *Provider) CanAccessWithGroupProviders(userPrincipalID string, groupPrincipals []v3.Principal) (bool, error) {
+	config, err := s.getSamlConfig()
+	if err != nil {
+		logrus.Errorf("Error fetching saml config: %v", err)
```
rmweir: is this err not going to bubble up?

luthermonson: this isn't actually any of my changes, the func is the original from the file when i got it just moved around or something and shows as my change. i think as a top level auth provider it will so ya this is redundant a bit but im not in a place to make the call

pkg/auth/providers/ldap/ldap_provider.go
luthermonson: fairly, but i rewrote to be extra safe

pkg/auth/providers/common/ldap/ldap_util.go
luthermonson: sure a bit, let me add

pkg/auth/providers/ldap/ldap_provider.go
luthermonson: rewrote a bit to be safe

thecrazyrussian:
Hi @luthermonson , this is super awesome. Not sure if you've been asked this before but we be able to add this via the api for okta for Active Directory?

PR #25409, opened 2020-02-13
Added test case to generate rke config

soumyalj:
Added a config file with 9 nodes

### DIFF
--- a/tests/validation/tests/v3_api/test_import_cluster.py
+++ b/tests/validation/tests/v3_api/test_import_cluster.py
@@ -1,39 +1,59 @@
 import os
 from lib.aws import AmazonWebServices
-from .common import *  # NOQA
+from .common import get_user_client
+from .common import run_command
+from .common import random_test_name
+from .common import run_command_with_stderr
+from .common import create_custom_host_registration_token
+from .common import validate_cluster
+from .common import cluster_cleanup
+from .common import readDataFile
+
+
 RANCHER_CLEANUP_CLUSTER = os.environ.get('RANCHER_CLEANUP_CLUSTER', "True")
 
 DATA_SUBDIR = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                            'resource')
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_import_cluster.py
```diff
@@ -65,15 +78,21 @@ def create_nodes():
...
 def create_rke_cluster_config(aws_nodes):
 
-    configfile = "cluster.yml"
+    # Create RKE Config file
+    print("Number of nodes ")
+    print(AWS_NODE_COUNT)
+    if (AWS_NODE_COUNT == 3):
+        configfile = "cluster.yml"
+    if (AWS_NODE_COUNT == 9):
+        configfile = "cluster_config_1.yml"
```
sangeethah: We should be able to generate the config file for any number of worker nodes without having to have any additional config yml. 
Existing config yml can be used for first 3 nodes and for the rest of the nodes build the configuration dynamically without relying on config yml.

soumyalj: Done

tests/validation/tests/v3_api/test_import_cluster.py
```diff
@@ -6,34 +6,44 @@
...
     client = get_user_client()
 
-    # Create nodes in AWS
-    aws_nodes = create_nodes()
+    # Create AWS nodes for the cluster
+    aws_nodes = \
+        AmazonWebServices().create_multiple_nodes(
+            AWS_NODE_COUNT, random_test_name("rketestcustom"),
+            wait_for_ready=True)
+    assert len(aws_nodes) == AWS_NODE_COUNT
```
sangeethah: We should not rely on AWS_NODE_COUNT since we can only support provisioning up to 3 nodes .

soumyalj: We can rely on AWS_NODE_COUNT and changes will be made to support it.

Issue #25184, opened 2020-01-31
Logging - Add more validation checks to detect when cluster/project level logs are not received.


sangeethah:
Logging - Add more validation checks to detect when cluster/project level logs are not received.
Current logging tests are not able to detect this failure -https://github.com/rancher/rancher/issues/24157#issuecomment-580519460

izaac:
This was completed.

PR #25414, opened 2020-02-14
[Automation] - fluentd logging e2e test

izaac:
This job deploys a fluentd aggregator app in a new project/namespace in Rancher and we setup cluster logging pointing to the aggregator host and port. And then:

- Activates cluster logging
- Parses the container view logs using websocket waiting for a `log_type` or `"{log"` string indicating that the aggregator is receiving the logs from Rancher
- Removes the aggregator project and disables cluster logging.

Then we do the same with a new aggregator app but we setup the `System` project logging  pointing to the aggregator's host and port, then:

- Activates project logging
- Parses the fluentd aggregator container view logs using websocket waiting for a `log_type` or `"{log"` string indicating that the aggregator is receiving the logs from Rancher.
- Removes the aggregator project and disables project logging in the `System` project.

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -3,6 +3,7 @@ import json
 import os
 import random
 import subprocess
+import ssl
 import time
 import requests
 import ast
@@ -2351,7 +2352,7 @@ class WebsocketLogParse:
         :param socket: the socket connection
         :param skip: if True skip the first char of the received message
         """
-        while True:
+        while True and socket.connected:
             try:
                 data = socket.recv()
                 # the message from the kubectl contains an extra char
@@ -2365,6 +2366,9 @@ class WebsocketLogParse:
                 self.lock.release()
             except websocket.WebSocketConnectionClosedException:
                 print("Connection closed")
+                break
+            except websocket.WebSocketProtocolException as wpe:
+                print("Error: {}".format(wpe))
                 break
 
     @staticmethod
@@ -2396,3 +2400,21 @@ def wait_for_cluster_delete(client, cluster_name, timeout=DEFAULT_TIMEOUT):
         time.sleep(.5)
         cluster = client.list_cluster(name=cluster_name).data
         cluster_count = len(cluster)
+
+
+def create_connection(url, subprotocols):
+    """
+    create a webscoket connection and check if it is connected
+    :param url: the url to connect to
+    :param subprotocols: the list of subprotocols
+    :return:
+    """
+    ws = websocket.create_connection(
+        url=url,
+        sslopt={"cert_reqs": ssl.CERT_NONE},
+        subprotocols=subprotocols,
+        timeout=10,
+        cookie="R_SESS=" + USER_TOKEN
+    )
+    assert ws.connected, "failed to build the websocket"
+    return ws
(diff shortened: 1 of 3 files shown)


izaac:
https://github.com/rancher/rancher/issues/25184

tests/validation/tests/v3_api/test_logging_e2e.py
```diff
@@ -0,0 +1,224 @@
+import pytest
```
bmdepesa: Alphabetize the imports

tests/validation/tests/v3_api/test_logging_e2e.py
```diff
@@ -0,0 +1,224 @@
...
+    name = random_test_name("fluentd")
+    return admin_client.create_cluster_logging(name=name,
+                                               clusterId=cluster.id,
+                                               fluentForwarderConfig=config,
+                                               enableJSONParsing=json_parsing,
+                                               outputFlushInterval=5
+                                               )
+
+
+def delete_cluster_logging(cluster_logging):
```
bmdepesa: This method and `delete_project_logging` are the same -- rename to `delete_logging` and remove one.

tests/validation/tests/v3_api/test_logging_e2e.py
```diff
@@ -0,0 +1,224 @@
...
+        if 'log_type' in wslog.last_message or '{"log"' in wslog.last_message:
+            found = True
+            wslog.last_message = ''
+            ws.close()
+            break
+        wslog.last_message = ''
+        ws.close()
+    assert found == True
+
+def create_connection(url, subprotocols):
```
bmdepesa: We already have this code somewhere -- we should re-use it.

tests/validation/tests/v3_api/test_logging_e2e.py
```diff
@@ -0,0 +1,224 @@
...
+def wait_for_match(wslog, url, timeout=DEFAULT_TIMEOUT):
+    start = time.time()
+    found = False
+
+    while not found:
+        if time.time() - start > timeout:
+            raise AssertionError(
+                "Timed out waiting for string to match in logs")
+        time.sleep(1)
+        ws = create_connection(url, ["base64.binary.k8s.io"])
```
bmdepesa: We shouldn't be continuously created and closing the connection while waiting for the logs. Open once, check for `last_message` in a loop, and close on a timeout.

tests/validation/tests/v3_api/common.py
```diff
@@ -2366,6 +2367,9 @@ def receiver(self, socket, skip):
             except websocket.WebSocketConnectionClosedException:
                 print("Connection closed")
                 break
+            except websocket.WebSocketProtocolException as wpe:
```
bmdepesa: This is catching an exception in the Jenkins output -- seems that it's still trying to call `socket.recv()` after the connection has been closed: try `while True && socket.connected`

tests/validation/tests/v3_api/test_logging_e2e.py
izaac: I just updated this, please check again. 👍

tests/validation/tests/v3_api/common.py
izaac: I added the validation the `OSError` is not always reproducible but I've added the `socket.connected` as suggested.

PR #25415, opened 2020-02-14
Add redeploy to workload API and integration test

shanewxy:
### The original PR has been reverted due to flaky test. https://github.com/rancher/rancher/pull/24777

Exposed redeploy REST API to the workload which set sets `cattle.io/timestamp` annotation to the current time.

Related issue: https://github.com/rancher/rancher/issues/23053
Related types PR: https://github.com/rancher/types/pull/1066

### DIFF
--- a/tests/integration/suite/test_workloads.py
+++ b/tests/integration/suite/test_workloads.py
@@ -1,5 +1,6 @@
 from .common import random_str
 from rancher import ApiError
+from .conftest import wait_for
 
 import time
 import pytest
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +312,43 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
...
+
+    workload = client.create_workload(
+        name=name,
+        namespaceId=ns.id,
+        scale=1,
+        containers=[{
+            'name': 'one',
+            'image': 'nginx',
+        }],
+        annotations={
```
rmweir: I don't think you should be adding the annotation like this. An annotation for timestamp won't normally be added during creation right?

tests/integration/suite/test_workloads.py
```diff
@@ -310,6 +312,43 @@ def test_statefulset_workload_volumemount_subpath(admin_pc):
...
+        workloads = client.list_workload(uuid=workload.uuid).data
+        return len(workloads) > 0 and workloads[0].annotations[
+            'cattle.io/timestamp'] != timestamp
+
+    wait_for(_timestamp_reset,
+             fail_handler=lambda: 'Timed out waiting for timestamp reset')
+
+    new_timestamp = client.list_workload(
+        uuid=workload.uuid).data[0].annotations['cattle.io/timestamp']
+    assert new_timestamp != timestamp
```
rmweir: nit: you are pretty much already asserting this since if it is not true then timestamp_reset will timeout and your fail_handler message will be shown.

shanewxy:
@rmweir updated

PR #25440, opened 2020-02-16
Move feature handler to scaled context

rmweir:
**Problem:**
Changing the effective value of non-dynamic features only causes a single rancher node to terminate.

**Solution:**
Move feature handler to scaled context.

**Issue:**
https://github.com/rancher/rancher/issues/25213


pkg/controllers/management/feature/feature_handler.go
```diff
@@ -11,8 +11,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
```
cbron: Should this move to `pkg/api/controllers/` ? Thats where it's registered now, and I don't see anything in `pkg/controllers/management` using scaledContext.

rmweir: whoops, good catch

Issue #25329, opened 2020-02-08
Zero Down: Add label to ignore nodes


maggieliu:

sowmyav27:
**verified on master-head - commit id - `926997017`**

**Steps:**

- Mark one worker node with label - `user.cattle.io/upgrade-policy=prevent`.
- Upgrade cluster by changing value

```
kubelet:
      extra_args:
        max-pods: '123'
```
- Cluster will go into `Updating` state.
- Worker node marked with label, does NOT get upgraded.
- Cluster comes back to `Active` state.

Issue #24073, opened 2019-11-13
worker node agents should report back when the upgrade is done


kinarashah:
With the introduction of upgradeStrategy, agents need to receive and send back some token to indicate the upgrade is complete.

kinarashah:
Implementation specific, would be tested as part of zero downtime feature.

Issue #23898, opened 2019-11-05
Add state transition for cluster when nodes are upgrading 


kinarashah:
Introduce logic for upgrade strategy and state transitions so that this information is easily available to be displayed in UI. 

https://github.com/rancher/rancher/issues/23777

sowmyav27:
**Verified on master-head - commit id:`926997017`**

- Update on cluster.yml
```
kubelet:
      extra_args: 
        max-pods: 2
```
- Cluster goes into `Updating` state
<img width="1442" alt="Screen Shot 2020-03-09 at 11 18 51 AM" src="https://user-images.githubusercontent.com/26032343/76244863-1d533f00-61f8-11ea-9989-33a071bf6a93.png">

- When the worker nodes are upgrading, the status of the cluster is `Upgrading`
<img width="1432" alt="Screen Shot 2020-03-09 at 11 18 58 AM" src="https://user-images.githubusercontent.com/26032343/76244867-1f1d0280-61f8-11ea-920a-72ea24e53143.png">

Issue #23897, opened 2019-11-05
Introduce upgrade strategy for workers 


kinarashah:
Worker nodes also respect some kind of upgrade strategy such that not all of them are updated simultaneously. It would include maxUnavailable count, default to 1 for control plane/etcd nodes. 

https://github.com/rancher/rancher/issues/23777

sowmyav27:
**On master-head - commit id: `926997017`**

- max_unavailable_worker: 10% is available by default on a cluster
- With default values upgrade_strategy, the sequence of how the upgrade happens in the cluster is given below
```
- etcd nodes - will upgrade one by one with the etcd components upgrading
- control plane nodes - will upgrade one by one - will be cordoned when this is happening. Control plane components - sidekick, kube-apiserver, kube-controller-manager, kube-scheduler and worker plane components - kubelet, kubeproxy
- etcd nodes will be cordoned - when the worker plane components - kubelet, kubeproxy are getting upgraded
- currently in rancher - worker plane nodes and add ons will get executed/upgraded in parallel.
- Default max_unavailable_worker: 10% so 1 worker node gets upgraded at a time. 
```
- If Drain is enabled, the nodes are drained.
- Else they go into cordoned state. and then Upgrade and to Active.
- State transition wrt to worker nodes from UI
- When Drain : false
```
-  Active --> Cordoned (Upgrading message) --> Active
```
- When Drain: True
```
-  Active --> Cordoned (Upgrading message) (might not be visible as it might happen really fast --> Draining --> Drained (Upgrading message) --> Active
```

PR #25451, opened 2020-02-18
rke worker upgrade logic

kinarashah:
Types: https://github.com/rancher/types/pull/1094 
rancher/rancher#23038

related issues: 
https://github.com/rancher/rancher/issues/24073
https://github.com/rancher/rancher/issues/23897
https://github.com/rancher/rancher/issues/25503
https://github.com/rancher/rancher/issues/25329
https://github.com/rancher/rancher/issues/23898

### DIFF
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -34,8 +34,7 @@ var (
 )
 
 const (
-	Token  = "X-API-Tunnel-Token"
-	Params = "X-API-Tunnel-Params"
+	Token = "X-API-Tunnel-Token"
 )
 
 func main() {
@@ -163,8 +162,8 @@ func run() error {
 	}
 
 	headers := map[string][]string{
-		Token:  {token},
-		Params: {base64.StdEncoding.EncodeToString(bytes)},
+		Token:                      {token},
+		rkenodeconfigclient.Params: {base64.StdEncoding.EncodeToString(bytes)},
 	}
 
 	serverURL, err := url.Parse(server)
@@ -269,7 +268,8 @@ func run() error {
 	onConnect := func(ctx context.Context) error {
 		connected()
 		connectConfig := fmt.Sprintf("https://%s/v3/connect/config", serverURL.Host)
-		if err := rkenodeconfigclient.ConfigClient(ctx, connectConfig, headers, writeCertsOnly); err != nil {
+		interval, err := rkenodeconfigclient.ConfigClient(ctx, connectConfig, headers, writeCertsOnly)
+		if err != nil {
 			return err
 		}
 
@@ -291,18 +291,23 @@ func run() error {
 		}
 
 		go func() {
-			logrus.Infof("Starting plan monitor")
+			logrus.Infof("Starting plan monitor, checking every %v seconds", interval)
+			tt := time.Duration(interval) * time.Second
 			for {
 				select {
-				case <-time.After(2 * time.Minute):
-					err := rkenodeconfigclient.ConfigClient(ctx, connectConfig, headers, writeCertsOnly)
+				case <-time.After(tt):
+					receivedInterval, err := rkenodeconfigclient.ConfigClient(ctx, connectConfig, headers, writeCertsOnly)
 					if err != nil {
 						logrus.Errorf("failed to check plan: %v", err)
+					} else if receivedInterval != 0 && receivedInterval != interval {
+						tt = time.Duration(receivedInterval) * time.Second
+						logrus.Infof("Plan monitor checking %v seconds", receivedInterval)
 					}
 				case <-ctx.Done():
 					return
 				}
 			}
+
 		}()
 
 		return nil
--- a/pkg/controllers/management/clusterprovisioner/provisioner.go
+++ b/pkg/controllers/management/clusterprovisioner/provisioner.go
@@ -43,6 +43,7 @@ type Provisioner struct {
 	ClusterController     v3.ClusterController
 	Clusters              v3.ClusterInterface
 	NodeLister            v3.NodeLister
+	Nodes                 v3.NodeInterface
 	engineService         *service.EngineService
 	backoff               *flowcontrol.Backoff
 	KontainerDriverLister v3.KontainerDriverLister
@@ -58,6 +59,7 @@ func Register(ctx context.Context, management *config.ManagementContext) {
 		Clusters:              management.Management.Clusters(""),
 		ClusterController:     management.Management.Clusters("").Controller(),
 		NodeLister:            management.Management.Nodes("").Controller().Lister(),
+		Nodes:                 management.Management.Nodes(""),
 		backoff:               flowcontrol.NewBackOff(30*time.Second, 10*time.Minute),
 		KontainerDriverLister: management.Management.KontainerDrivers("").Controller().Lister(),
 		DynamicSchemasLister:  management.Management.DynamicSchemas("").Controller().Lister(),
@@ -519,6 +521,10 @@ func (p *Provisioner) reconcileCluster(cluster *v3.Cluster, create bool) (*v3.Cl
 		cluster.Status.CACert = caCert
 		resetRkeConfigFlags(cluster, updateTriggered)
 
+		if cluster.Status.AppliedSpec.RancherKubernetesEngineConfig != nil {
+			cluster.Status.NodeVersion++
+		}
+
 		if cluster, err = p.Clusters.Update(cluster); err == nil {
 			saved = true
 			break
@@ -532,8 +538,17 @@ func (p *Provisioner) reconcileCluster(cluster *v3.Cluster, create bool) (*v3.Cl
 		return cluster, fmt.Errorf("failed to update cluster")
 	}
 
+	if cluster.Status.AppliedSpec.RancherKubernetesEngineConfig != nil {
+		logrus.Infof("Updated cluster [%s] with node version [%v]", cluster.Name, cluster.Status.NodeVersion)
+		p.reconcileForUpgrade(cluster.Name)
+	}
+
 	logrus.Infof("Provisioned cluster [%s]", cluster.Name)
 	return cluster, nil
+}
+
+func (p *Provisioner) reconcileForUpgrade(clusterName string) {
+	p.Nodes.Controller().Enqueue(clusterName, "upgrade_")
 }
 
 func (p *Provisioner) setGenericConfigs(cluster *v3.Cluster) {
--- a/pkg/controllers/management/rkeworkerupgrader/node.go
+++ b/pkg/controllers/management/rkeworkerupgrader/node.go
@@ -1,0 +1,181 @@
+package rkeworkerupgrader
+
+import (
+	"fmt"
+
+	nodehelper "github.com/rancher/rancher/pkg/node"
+	nodeserver "github.com/rancher/rancher/pkg/rkenodeconfigserver"
+	rkeservices "github.com/rancher/rke/services"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
+)
+
+func (uh *upgradeHandler) prepareNode(node *v3.Node, toDrain bool, nodeDrainInput *v3.NodeDrainInput) error {
+	var nodeCopy *v3.Node
+	if toDrain {
+		if node.Spec.DesiredNodeUnschedulable == "drain" {
+			return nil
+		}
+		nodeCopy = node.DeepCopy()
+		nodeCopy.Spec.DesiredNodeUnschedulable = "drain"
+		nodeCopy.Spec.NodeDrainInput = nodeDrainInput
+	} else {
+		if node.Spec.DesiredNodeUnschedulable == "true" || node.Spec.InternalNodeSpec.Unschedulable {
+			return nil
+		}
+		nodeCopy = node.DeepCopy()
+		nodeCopy.Spec.DesiredNodeUnschedulable = "true"
+	}
+
+	if _, err := uh.nodes.Update(nodeCopy); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (uh *upgradeHandler) processNode(node *v3.Node, cluster *v3.Cluster, msg string) error {
+	nodePlan, err := uh.getNodePlan(node, cluster)
+	if err != nil {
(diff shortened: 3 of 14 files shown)


pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -346,14 +335,112 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string) e
...
+func (uh *upgradeHandler) processNode(node *v3.Node, cluster *v3.Cluster) error {
+	nodePlan, err := uh.getNodePlan(node, cluster)
+	if err != nil {
+		return fmt.Errorf("setNodePlan: error getting node plan for [%s]: %v", node.Name, err)
+	}
+
+	nodeCopy := node.DeepCopy()
+	nodeCopy.Spec.NodePlan.Plan = nodePlan
+	nodeCopy.Spec.NodePlan.Version = cluster.Status.NodeVersion
+	nodeCopy.Spec.NodePlan.AgentCheckInterval = agentCheckIntervalDuringUpgrade
```
kinarashah: I'm modifying when to set this AgentCheckInterval, but that doesn't impact overall logic and PR ready for first  review

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,540 @@
...
+func unlock(name string) {
+	time.Sleep(5 * time.Second)
+	clusterLockMutex.RLock()
+	cMutex := clusterLock[name]
+	clusterLockMutex.RUnlock()
+
+	cMutex.Unlock()
+}
+
+func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string) error {
```
mrajashree: do we need to add check for cluster driver so we can return for non-rke?

mrajashree:
like discussed offline we'll also need logic in cluster store to set UpgradeStrategy and MaxUnavailable defaults since UpgradeStrategy is a pointer field and won't be set with defaults in spite of having them in types

pkg/controllers/management/rkeworkerupgrader/plan.go
```diff
@@ -0,0 +1,355 @@
...
+
+	if rkeConfig == nil {
+		rkeConfig = &v3.RancherKubernetesEngineConfig{}
+	}
+
+	rkeConfig = rkeConfig.DeepCopy()
+	rkeConfig.Nodes = []v3.RKEConfigNode{
+		*node.Status.NodeConfig,
+	}
+	rkeConfig.Nodes[0].Role = []string{rkeservices.WorkerRole, rkeservices.ETCDRole, rkeservices.ControlRole}
```
mrajashree: why do we need to build include etcd components in the plan?

pkg/controllers/management/rkeworkerupgrader/plan.go
```diff
@@ -0,0 +1,355 @@
...
+func (uh *upgradeHandler) workerPlan(node *v3.Node, cluster *v3.Cluster) (*v3.RKEConfigNodePlan, error) {
+	infos, err := librke.GetDockerInfo(node)
+	if err != nil {
+		return nil, err
+	}
+
+	hostAddress := node.Status.NodeConfig.Address
+	hostDockerInfo := infos[hostAddress]
+
+	rkeConfig := cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.DeepCopy()
```
mrajashree: need to check for RancherKubernetesEngineConfig being nil here too?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,540 @@
...
+	for _, node := range nodes {
+		if node.Status.NodeConfig == nil {
+			continue
+		}
+
+		if !workerOnly(node.Status.NodeConfig.Role) {
+			continue
+		}
+
+		if node.Spec.NodePlan == nil {
```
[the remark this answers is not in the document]
kinarashah: gets handled here, https://github.com/rancher/rancher/pull/25451/files#diff-a16c43ed026936b620e2fdb9ca6196b8R103

pkg/controllers/management/rkeworkerupgrader/upgrade.go
kinarashah: checking here, think this should suffice? https://github.com/rancher/rancher/pull/25451/files#diff-a16c43ed026936b620e2fdb9ca6196b8R74

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,540 @@
...
+			continue
+		}
+
+		// skip nodes marked for ignore by user
+		if node.Labels != nil && node.Labels[ignoreKey] == "true" {
+			continue
+		}
+
+		// skip provisioning nodes
+		if !v3.NodeConditionProvisioned.IsTrue(node) || !v3.NodeConditionRegistered.IsTrue(node) {
```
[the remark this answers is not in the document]
kinarashah: NodeConditionInitialized seems to be only for create, NodeConditionProvisioned won't be true

pkg/controllers/management/rkeworkerupgrader/plan.go
kinarashah: think this should cover it all, https://github.com/rancher/rancher/pull/25451/files#diff-a16c43ed026936b620e2fdb9ca6196b8R98

pkg/agent/main.go
```diff
@@ -329,6 +316,26 @@ func run() error {
...
+	tt := time.Duration(interval) * time.Second
+	for {
+		select {
+		case <-time.After(tt):
+			receivedInterval, err := rkenodeconfigclient.ConfigClient(ctx, connectConfig, headers, writeCertsOnly)
+			if err != nil {
+				logrus.Errorf("failed to check plan: %v", err)
+			}
+			if receivedInterval != interval {
+				go planMonitor(ctx, connectConfig, headers, writeCertsOnly, receivedInterval)
```
dramich: Can tt just be updated and continue vs starting up a new go routine?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,395 @@
...
+	)
+	if isNonWorkerOnly(node.Status.NodeConfig.Role) {
+		nodePlan, err = uh.nonWorkerPlan(node, cluster)
+	} else {
+		nodePlan, err = uh.workerPlan(node, cluster)
+	}
+	return nodePlan, err
+}
+
+func lock(name string) {
```
dramich: Looks like you just recreated https://github.com/moby/moby/tree/master/pkg/locker

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -519,6 +521,12 @@ func (p *Provisioner) reconcileCluster(cluster *v3.Cluster, create bool) (*v3.Cl
 		cluster.Status.CACert = caCert
 		resetRkeConfigFlags(cluster, updateTriggered)
 
+		if cluster.Status.AppliedSpec.RancherKubernetesEngineConfig != nil {
+			cluster.Status.NodeVersion++
```
dramich: Is it intended to have this update if we error? This loop runs 20 times potentially

dramich: I think this check should move below the update call, once that succeeds then call `reconcileForUpgrade`

dramich: To clarify, update the nodeVersion, call update, if that succeeds then call reconcileForUpgrade

pkg/controllers/management/rkeworkerupgrader/plan.go
```diff
@@ -0,0 +1,346 @@
...
+
+func sliceToMap(s []string) map[string]bool {
+	m := map[string]bool{}
+	for _, v := range s {
+		m[v] = true
+	}
+	return m
+}
+
+func myEqual(p v3.Process, o v3.Process) bool {
```
dramich: I don't see this being used anywhere

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,395 @@
...
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	WorkerThreads = 20
```
dramich: This isn't used

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,386 @@
...
+
+		if planChangedForUpgrade(nodePlan, node.Status.NodePlan.Plan) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func getNum(maxUnavailable string, nodes int) (int, error) {
```
superseb: Can we rename this function to describe what we are getting? I guess this is similar to `CalculateMaxUnavailable` in RKE?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -0,0 +1,386 @@
...
+				return err
+			}
+
+			logrus.Infof("cluster [%s] worker-upgrade: updated unavailable node [%s] to attempt upgrade", clusterName, name)
+		}
+	}
+
+	unavailable := upgrading + len(notReadyMap)
+
+	logrus.Debugf("cluster [%s] worker-upgrade: workerNodeInfo: nodes %v maxAllowed %v unavailable %v upgrading %v notReady %v "+
```
superseb: Can we move this up to directly after maxAllowed (line 236) so we directly see what we are dealing with before processing any logic?

Probably add a `len()` as well as with many nodes you don't want to be counting the list of nodes to know how many there are being used for calculation.

pkg/controllers/user/nodesyncer/cordonfieldssyncer.go
```diff
@@ -47,16 +47,22 @@ func (m *nodesSyncer) syncCordonFields(key string, obj *v3.Node) (runtime.Object
...
+		nodeCopy := obj.DeepCopy()
+		nodeCopy.Spec.DesiredNodeUnschedulable = ""
+
+		if !obj.Spec.InternalNodeSpec.Unschedulable {
+			removeDrainCondition(nodeCopy)
+		}
+
+		obj, err = m.machines.Update(nodeCopy)
+	} else {
+		logrus.Debugf("cordonNode: [%v] desired %v current %v", obj.Name, desiredValue, obj.Spec.InternalNodeSpec.Unschedulable)
```
superseb: Can we remove the else and just log this above the if?

pkg/controllers/management/rkeworkerupgrader/plan.go
```diff
@@ -0,0 +1,302 @@
...
+
+func planChangedForUpgrade(newPlan *v3.RKEConfigNodePlan, oldPlan *v3.RKEConfigNodePlan) bool {
+	if newPlan == nil || oldPlan == nil {
+		return true
+	}
+	newProcesses := newPlan.Processes
+	oldProcesses := oldPlan.Processes
+
+	if len(newProcesses) != len(oldProcesses) {
+		logrus.Infof("returning true from length comparison")
```
superseb: This needs more context

pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return nil
+}
+
+func (uh *upgradeHandler) filterNodes(nodes []*v3.Node, expectedVersion int) (map[string]*v3.Node, map[string]*v3.Node, map[string]*v3.Node, map[string]*v3.Node, int, int, int) {
+	done, upgrading, filtered := 0, 0, 0
+	toPrepareMap, toProcessMap, upgradedMap, notReadyMap := map[string]*v3.Node{}, map[string]*v3.Node{}, map[string]*v3.Node{}, map[string]*v3.Node{}
+
+	for _, node := range nodes {
+		if node.DeletionTimestamp != nil || node.Status.NodeConfig == nil || !workerOnly(node.Status.NodeConfig.Role) {
```
superseb: Can we add debug log for each of these conditions similar to when node is not ready so we know when a node is getting skipped and why?

```
			logrus.Debugf("node [%s] is not ready", node.Name)
```

pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -0,0 +1,162 @@
...
+func preparingNode(node *v3.Node) bool {
+	return node.Spec.DesiredNodeUnschedulable == "drain" || node.Spec.DesiredNodeUnschedulable == "true"
+}
+
+func preparedNode(node *v3.Node) bool {
+	return v3.NodeConditionDrained.IsTrue(node) || node.Spec.InternalNodeSpec.Unschedulable || v3.NodeConditionUpgraded.IsUnknown(node)
+}
+
+func workerOnly(roles []string) bool {
+	worker := false
```
superseb: Is `workerOnly` and `isNonWorkerOnly` written differently because you don't want to use `if ! slice.ContainString`?

superseb:
Just two things, rest LGTM

pkg/controllers/management/rkeworkerupgrader/node.go
kinarashah: `isNonWorkerOnly` was used https://github.com/rancher/rancher/blob/master/pkg/rkenodeconfigserver/nodeserver.go#L117:1 , so I just got it from there 😄 

I've updated to return in one iteration vs `ContainString` and renamed it to `IsNonWorker` since it checks for etcd/control but doesn't check for a worker role

pkg/controllers/management/rkeworkerupgrader/node.go
kinarashah: introduced a `skipNode` function for every condition

Issue #25122, opened 2020-01-27
Remove pspt lifecycle and move to user sync


dramich:
Cluster scoped lifecycles cause a lot of additional work for controllers and also make the object very hard to view when rancher is managing high numbers of clusters. For example, roleTemplates in a 2k cluster rancher setup will have 2k finalizers and 2k annotations. 

The below controllers should be investigated to determine if it's possible to rewrite them as non-cluster scoped lifecycles. 

Some options would be to split it into two controllers, a sync and a management lifecycle
An owners reference if an object need to be deleted.

```
grep -ir 'New.*LifecycleAdapter' pkg/controllers
pkg/controllers/user/rbac/podsecuritypolicy/template.go:	psptSync := v3.NewPodSecurityPolicyTemplateLifecycleAdapter("cluster-pspt-sync_"+context.ClusterName, true, pspti, lfc)
pkg/controllers/user/secret/secret.go:	sync := v1.NewSecretLifecycleAdapter(fmt.Sprintf("secretsController_%s", cluster.ClusterName), true,
```

EDIT: This issue is specifically for PodSecurityPolicyTemplate lifecycle getting moved out of being a cluster scoped lifecycle.

bmdepesa:
rancher/rancher:v2.4.0-rc7

* Ran long and short-running automation for regression validation

PR #25453, opened 2020-02-18
Remove pspt lifecycle and move to user sync

dramich:
Problem:
PSPT running a cluster scoped lifecycle makes pspt objects have
finalizers and annotations for all clusters

Solution:
Move pspt to a user scoped sync handler and add sync handlers to cleanup
psp and clusterRole when the parent pspt is removed

https://github.com/rancher/rancher/issues/25122


pkg/controllers/user/rbac/podsecuritypolicy/template.go
```diff
@@ -37,37 +38,41 @@ func RegisterTemplate(ctx context.Context, context *config.UserContext) {
...
 
-func (l *Lifecycle) Create(obj *v3.PodSecurityPolicyTemplate) (runtime.Object, error) {
-	return nil, nil
-}
+func (p *psptHandler) sync(key string, obj *v3.PodSecurityPolicyTemplate) (runtime.Object, error) {
+	if obj == nil {
+		return nil, nil
+	}
+
+	if _, ok := obj.Annotations[podSecurityPolicyTemplateCleanup]; !ok {
```
rmweir: nit: if it doesn't have cleanup annotation then cleanup? Cleaned would make more sense to me

tests/integration/suite/test_pod_security_policies.py
```diff
@@ -281,3 +281,65 @@ def check_project():
...
+        'metadata': {
+            'annotations': {
+                'serviceaccount.cluster.cattle.io/pod-security': 'doesntexist'
+            }
+        }
+    }
+    # Add the annotation the controller is looking for
+    psp = policy.patch_pod_security_policy(psp_name, anno)
+
+    # Controller will deleting the PSP as the parent PSPT doesn't exist
```
rmweir: nit: delete*

tests/integration/suite/test_pod_security_policies.py
```diff
@@ -281,3 +281,65 @@ def check_project():
...
+            "volumes": [
+                "*"
+            ]
+        }
+    }
+
+    psp = policy.create_pod_security_policy(args)
+    remove_resouce_func(policy.delete_pod_security_policy, psp_name)
+    psp = policy.read_pod_security_policy(psp_name)
+    assert psp is not None
```
rmweir: wouldn't it raise an ApiExcept if it couldn't find it? Also, should you retry this for some duration as even if it did delete it may not happen immediately right?

pkg/controllers/user/rbac/podsecuritypolicy/template.go
cbron: agree

tests/integration/suite/test_pod_security_policies.py
dramich: The issue with waiting is how long is long enough? The patch will fail if the psp is gone

Issue #25326, opened 2020-02-08
CIS Scan - Ability to Schedule Scan and Alert when there is failure


maggieliu:
CIS Scheduling:
* Need to make sure it scales with cluster
* Need to Parse the cron string and add 2 fields (Last run & next run time)
* Use Enqueueafter to schedule the run

CIS Alerting:
* New Alert group
* Create a new radio option for "Security Scan"
* The new option will allow for sending alert on Critical or Info
* The existing CIS scan controller will send a "Post" to the alert manager to trigger the alert
     * Need to make sure when the post fails and comes back, the controller won't do another scan just do a post.

izaac:
Rancher version: `2.4 master-head (03/09/2020)` commit id: 17bd84f8d

This is throwing me a panic error:

```log
2020/03/09 21:17:11 [INFO] Provisioned cluster [c-94hv8]
2020/03/09 21:17:11 [INFO] scheduledScanHandler: clusterSync: Scheduled Scan enabled for cluster c-94hv8: */1 * * * *
2020/03/09 21:17:11 [INFO] checking cluster [c-94hv8] for worker nodes upgrade
2020/03/09 21:18:00 [INFO] cisScanHandler: Create: deploying helm chart
2020/03/09 21:18:00 [ERROR] ClusterScanController c-94hv8/ss-cis-1583788680000279059 recovered from panic "runtime error: invalid memory address or nil pointer dereference". (err=<nil>) Call stack:
goroutine 19989 [running]:
github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime.RecoverFromPanic(0xc007563d30)
	/go/src/github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:158 +0xb5
panic(0x3664a00, 0x6bed090)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/rancher/rancher/pkg/controllers/user/cis.(*cisScanHandler).Create(0xc004555f00, 0xc005decc00, 0xc00982f818, 0x44618c, 0x60, 0x3b69d80)
	/go/src/github.com/rancher/rancher/pkg/controllers/user/cis/clusterScanHandler.go:185 +0xe7f
[...]

Issue #25325, opened 2020-02-08
CIS Scan - Ability to run Permissive & Hardened Profiles


maggieliu:
CIS Scan design summary:
* Types of CIS tests: Not Applicable, Skipped (Rancher), Skipped (User), Active
* There should be two profiles for CIS scan "Permissive" vs "Hardened"
        * By default, "Permissive" run contains Rancher recommended skip tests and it cannot be unskipped.
        * By default, "Hardened" run contains no skipped tests
* User can select either profile to run a CIS scan
         * If there are user skipped tests it will be applied on top of the profile selected.

 (i.e: if there are 5 user skipped tests and 10 Rancher skipped tests, when user picks "Permissive" to run, the run will contain 15 skipped test. If user selected "Hardened" to run, the run will contain 5 skipped tests)

* User skipped tests will be saved in a config-map file
* User can predefine skipped tests by calling the API and passing in the config map
* Terraform should be able to schedule CIS scan run with an option of which profile to run it with

sowmyav27:
**Verified on master-head commit id: `63306835a`**

- User is able to run cis scans using Permissive & Hardened Profiles from the API and the UI.
<img width="853" alt="Screen Shot 2020-03-06 at 9 37 19 PM" src="https://user-images.githubusercontent.com/26032343/76137509-cd575b00-5ff2-11ea-9f0f-832702ccc6a7.png">
<img width="1289" alt="Screen Shot 2020-03-06 at 9 37 09 PM" src="https://user-images.githubusercontent.com/26032343/76137510-ce888800-5ff2-11ea-8526-24a7758437a2.png">

Issue #24987, opened 2020-01-17
CIS scan tool support for cis v 1.5


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** enhancement

CIS Scan tool must support scan for CIS benchmark v 1.5

leodotcloud:
These are the current pending failing tests that need to be looked into. Some of them can use the fixes from rke-cis-1.4 directly, others are totally new.

  |   |   |   |   |  
-- | -- | -- | -- | -- | --
Open accordion | Fail | 1.1.13 | Ensure that the admin.conf file permissions are set to 644 or more restrictive (Scored)Remediation: Run the below command (based on the file location on your system) on the master node. For example, chmod 644 /etc/kubernetes/admin.conf | Skip |  
Open accordion | Fail | 1.1.14 | Ensure that the admin.conf file ownership is set to root:root (Scored)Remediation: Run the below command (based on the file location on your system) on the master node. For example, chown root:root /etc/kubernetes/admin.conf | Skip |  
Open accordion | Fail | 1.1.15 | Ensure that the scheduler.conf file permissions are set to 644 or more restrictive (Scored)Remediation: Run the below command (based on the file location on your system) on the master node. For example, chmod 644 /etc/kubernetes/scheduler.conf | Skip |  
Open accordion | Fail | 1.1.16 | Ensure that the scheduler.conf file ownership is set to root:root (Scored)Remediation: Run the below command (based on the file location on your system) on the master node. For example, chown root:root /etc/kubernetes/scheduler.conf | Skip |  
Open accordion | Fail | 1.1.17 | Ensure that the controller-manager.conf file permissions are set to 644 or more restrictive (Scored)Remediation: Run the below command (based on the file location on your system) on the master node. For example, chmod 644 /etc/kubernetes/controller-manager.conf | Skip |  
Open accordion | Fail | 1.1.18 | Ensure that the controller-manager.conf file ownership is set to root:root (Scored)Remediation: Run the below command (based on the file location on your system) on the master node. For example, chown root:root /etc/kubernetes/controller-manager.conf | Skip |  
Open accordion | Fail | 1.2.34 | Ensure that encryption providers are appropriately configured (Scored)Remediation: Follow the Kubernetes documentation and configure a EncryptionConfig file. In this file, choose aescbc, kms or secretbox as the encryption provider. | Skip |  
[...]

PR #25475, opened 2020-02-19
Changes for cis-1.5, scheduling, profiles, alerts

leodotcloud:


maggieliu:
Issues:
Schedule/Alert: rancher/rancher#25326
Profiles: rancher/rancher#25325
CIS 1.5 support: rancher/rancher#24987

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -82,6 +89,27 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
...
+			logrus.Errorf("cisScanHandler: Create: error listing pods: %v", err)
+		}
+		if len(pods) == 0 {
+			removed = true
+			break
+		}
+		time.Sleep(RetryIntervalInMilliseconds * time.Millisecond)
+	}
+	if !removed {
+		return cs, fmt.Errorf("cisScanHandler: Create: runner pod from previous scan not deleted yet, will retry in a bit")
```
kinarashah: isn't this for master pod?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -82,6 +89,27 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
 		v3.ClusterScanConditionFailed.Message(cs, "cannot run scan on a windows cluster")
 		return cs, nil
 	}
+
+	// Wait for complete pod removal from previous scan
+	removed := false
+	for retry := NumberOfRetriesForPodRemoval; retry > 0; retry-- {
+		pods, err := csh.podLister.List(
+			v3.DefaultNamespaceForCis,
+			labels.Set(SonobuoyMasterLabel).AsSelector(),
```
kinarashah: q: does this mean we can only have master across all clusters?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -251,6 +279,21 @@ func (csh *cisScanHandler) Updated(cs *v3.ClusterScan) (runtime.Object, error) {
 			}
 		}
 
+		// Wait for complete pod removal
+		removed := false
+		for retry := NumberOfRetriesForPodRemoval; retry > 0; retry-- {
+			time.Sleep(RetryIntervalInMilliseconds * time.Millisecond)
+			podName := fmt.Sprintf("security-scan-runner-%v", cs.Name)
+			_, err := csh.podLister.Get(v3.DefaultNamespaceForCis, podName)
+			if err != nil && kerrors.IsNotFound(err) {
```
kinarashah: return if err is !notFound

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,266 @@
...
+			cronSchedule := getCronSchedule(cluster)
+			clusterInfo = &scheduleInfo{
+				CronSchedule: cronSchedule,
+			}
+			logrus.Infof("scheduledScanHandler: clusterSync: Scheduled Scan enabled for cluster %v: %v",
+				cluster.Name, cronSchedule)
+			if err := scheduleScan(cluster, ssh.clusterScanClient, clusterInfo); err != nil {
+				return nil, fmt.Errorf("scheduledScanHandler: clusterSync: error scheduling scan: %v", err)
+			}
+			ssh.clustersMap[cluster.Name] = clusterInfo
```
kinarashah: multiple clusters could access the same map, we'd need to handle that using lock or syncMap

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,266 @@
...
+	if !ok {
+		logrus.Infof("scheduledScanHandler: sync: scheduled scan has been disabled, ignoring")
+	}
+	cluster, err := ssh.clusterLister.Get("", clusterId)
+	if err != nil {
+		return nil, fmt.Errorf("scheduledScanHandler: sync: error fetching cluster %v: %v", clusterId, err)
+	}
+
+	if !cluster.Spec.ScheduledClusterScan.Enabled {
+		logrus.Infof("scheduledScanHandler: sync: scheduled scan is disabled, ignoring the current one")
```
kinarashah: wondering if it'd be ok to delete this from clustersMap?

pkg/controllers/user/cis/register.go
```diff
@@ -3,61 +3,84 @@ package cis
...
 // Register initializes the controllers and registers
 func Register(ctx context.Context, userContext *config.UserContext) {
-	logrus.Infof("Registering CIS controller")
-
 	clusterName := userContext.ClusterName
-	clusterLister := userContext.Management.Management.Clusters(metav1.NamespaceAll).Controller().Lister()
+	clusterClient := userContext.Management.Management.Clusters(metav1.NamespaceAll)
+	cluster, err := clusterClient.Get(clusterName, metav1.GetOptions{})
+	if err != nil {
+		logrus.Errorf("error fetching cluster: %v", err)
```
kinarashah: 👀 👀 do we need retries? 
seeing this approach for the first time, afraid what if miss out on registering the controller totally

pkg/controllers/user/alert/watcher/cluster_scan.go
```diff
@@ -0,0 +1,161 @@
...
+	clusterAlertRuleLister := cluster.Management.Management.ClusterAlertRules(cluster.ClusterName).Controller().Lister()
+
+	clusterScanWatcher := &ClusterScanWatcher{
+		clusterName:            cluster.ClusterName,
+		clusterLister:          clusterLister,
+		clusterScanLister:      clusterScanLister,
+		clusterAlertRuleLister: clusterAlertRuleLister,
+		alertManager:           alertManager,
+	}
+	clusterScans.AddClusterScopedLifecycle(ctx, "cluster-scan-watcher", cluster.ClusterName, clusterScanWatcher)
```
kinarashah: q: why is this a Lifecycle?

pkg/controllers/user/alert/watcher/cluster_scan.go
```diff
@@ -0,0 +1,161 @@
...
+	data["rule_id"] = ruleID
+	data["group_id"] = alertRule.Spec.GroupName
+	data["alert_type"] = constants.AlertTypeClusterScan
+	data["alert_name"] = alertRule.Spec.DisplayName
+	data["severity"] = alertRule.Spec.Severity
+	data["cluster_name"] = clusterDisplayName
+	data["component_name"] = cs.Name
+	data["logs"] = csw.getAlertMessage(cs, alertRule)
+
+	logrus.Infof("ClusterScanWatcher: sendAlert: about to send alert")
```
kinarashah: nit: needs to have cluster+alert name

pkg/controllers/user/cis/scan.go
```diff
@@ -36,3 +38,60 @@ func NewCisScan(cluster *v3.Cluster, cisScanConfig *v3.CisScanConfig) *v3.Cluste
...
+		return nil, fmt.Errorf("failed to create cis scan object")
+	}
+
+	updatedCluster := cluster.DeepCopy()
+	updatedCluster.Annotations[v3.RunCisScanAnnotation] = cisScan.Name
+
+	// Can't add either too many retries or longer interval as this an API handler
+	for i := 0; i < retryCount; i++ {
+		_, err = clusterClient.Update(updatedCluster)
+		if err == nil {
```
kinarashah: retry only if conflict error?

pkg/controllers/user/alert/watcher/cluster_scan.go
```diff
@@ -0,0 +1,161 @@
...
+	alertSuccessful := true
+	for _, alertRule := range matchingAlertRules {
+		if e := csw.sendAlert(cs, alertRule); e != nil {
+			alertSuccessful = false
+			logrus.Errorf("ClusterScanWatcher: Sync: error sending alert: %v", e)
+			err = multierror.Append(err, e)
+		}
+	}
+	if !alertSuccessful {
+		return cs, err
```
kinarashah: on error for any matchingAlertRule, we'd be sending alert again even for successful ones on next sync? is that expected?

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: will do.

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: I have the delete logic in the clustersync when the schedule is disabled, not necessary to delete it here. It's just cleaner to have it in one place.

pkg/controllers/user/cis/scan.go
leodotcloud: 👍, will add that check.

pkg/controllers/user/alert/watcher/cluster_scan.go
leodotcloud: alert-manager kind of buffers the alerts for a short duration, this should be fine.

pkg/controllers/user/alert/watcher/cluster_scan.go
leodotcloud: actually, I had them while debugging, will remove the logs.

pkg/controllers/user/cis/register.go
leodotcloud: will add retries

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: one master per cluster ... hence we make sure there is only one scan running in a cluster at any given time.

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -82,6 +89,27 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
...
+
+	// Wait for complete pod removal from previous scan
+	removed := false
+	for retry := NumberOfRetriesForPodRemoval; retry > 0; retry-- {
+		pods, err := csh.podLister.List(
+			v3.DefaultNamespaceForCis,
+			labels.Set(SonobuoyMasterLabel).AsSelector(),
+		)
+		if err != nil {
+			logrus.Errorf("cisScanHandler: Create: error listing pods: %v", err)
```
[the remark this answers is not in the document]
leodotcloud: will fix

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: runner == master; worker == slave

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: will refactor this duplicated logic into a diff function

pkg/controllers/management/cis/register.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Register initializes the controllers and registers
+func Register(ctx context.Context, management *config.ManagementContext) {
+	logrus.Infof("Registering CIS scheduled scan controller")
```
prachidamle: Do we want this at Info?

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,272 @@
...
+	defer ssh.Unlock()
+	clusterInfo, ok := ssh.clustersMap[cluster.Name]
+	if cluster.Spec.ScheduledClusterScan.Enabled {
+		if !ok {
+			// Do stuff to enable scheduling
+			cronSchedule := getCronSchedule(cluster)
+			clusterInfo = &scheduleInfo{
+				cronSchedule: cronSchedule,
+			}
+			logrus.Infof("scheduledScanHandler: clusterSync: Scheduled Scan enabled for cluster %v: %v",
```
prachidamle: do we want to log this at debug?

pkg/controllers/management/cis/register.go
leodotcloud: This gets printed only once, should be ok.

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: This gets printed only when it's enabled, not always, so should be ok.

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+		return nil, nil
+	}
+	splits := strings.Split(strings.TrimPrefix(key, ScheduledScanKeyPrefix), separator)
+	clusterID := splits[0]
+	pastCronSchedule := splits[1]
+	ssh.Lock()
+	defer ssh.Unlock()
+	clusterInfo, ok := ssh.clustersMap[clusterID]
+	if !ok {
+		logrus.Infof("scheduledScanHandler: sync: scheduled scan has been disabled, ignoring")
```
prachidamle: We should change this to debug

pkg/controllers/management/cis/register.go
prachidamle: We don't really need this, the framework will log each controller that registers.

pkg/controllers/management/cis/scheduled_scan.go
prachidamle: missing a <return nil, nil> here?

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+	cluster, err := ssh.clusterLister.Get("", clusterID)
+	if err != nil {
+		return nil, fmt.Errorf("scheduledScanHandler: sync: error fetching cluster %v: %v", clusterID, err)
+	}
+
+	if err := cis.ValidateClusterBeforeLaunchingScan(cluster); err != nil {
+		return nil, err
+	}
+
+	if !cluster.Spec.ScheduledClusterScan.Enabled {
```
prachidamle: This check could be before  <cis.ValidateClusterBeforeLaunchingScan>, saves additional checks.

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+	if cluster.Spec.ScheduledClusterScan.ScanConfig == nil ||
+		cluster.Spec.ScheduledClusterScan.ScanConfig.CisScanConfig == nil {
+		cisScanConfig = &v3.CisScanConfig{
+			Profile: v3.CisScanProfileTypePermissive,
+		}
+	}
+
+	nextAt := clusterInfo.nextScanAt
+	now := time.Now().Round(time.Second)
+	if !nextAt.Equal(now) {
```
prachidamle: when you return from here, the next scan is not scheduled, should't we be scheduling the next one to maintain the schedule?

prachidamle: Also, I think we should be more tolerant of delays occuring in the scheduling, and not return when !nextAt.Equal(now) -  This should allow some offset or use >=

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+		logrus.Infof("scheduledScanHandler: checkAndLaunchScan: cluster %v next: %v, now: %v",
```
prachidamle: Change to debug here

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if _, err = checkAndLaunchScan(cluster, ssh.clusterClient, ssh.clusterScanClient, clusterInfo); err != nil {
+		return nil, fmt.Errorf("scheduledScanHandler: clusterSync: cluster %v: error launching scan: %v",
+			cluster.Name, err)
+	}
+
+	retention := getRetention(cluster)
+	clusterScanClient := ssh.managementClient.ClusterScans(clusterID)
+	if err := cleanOldScans(clusterID, retention, clusterScanClient, ssh.clusterScanLister); err != nil {
+		return nil, fmt.Errorf("scheduledScanHandler: clusterSync: error cleaning old scans: %v", err)
```
prachidamle: I am not sure we should return the error you get from <cleanOldScans> - this error is going to trigger another sync for the same key for which a scan has been already launched

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+		clusterClient,
+		clusterScanClient,
+		cis.RetryIntervalInMilliseconds,
+		cis.NumberOfRetriesForClusterUpdate,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("scheduledScanHandler: checkAndLaunchScan: error launching scan: %v", err)
+	}
+	clusterInfo.lastScanLaunchedAt = time.Now().Round(time.Second)
+	logrus.Infof("scheduledScanHandler: checkAndLaunchScan: launched scan for cluster %v at: %v",
```
prachidamle: log this at debug

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+func (ssh *scheduledScanHandler) sync(key string, _ *v3.ClusterScan) (runtime.Object, error) {
+	if !strings.HasPrefix(key, ScheduledScanKeyPrefix) {
+		return nil, nil
+	}
+	splits := strings.Split(strings.TrimPrefix(key, ScheduledScanKeyPrefix), separator)
+	clusterID := splits[0]
+	pastCronSchedule := splits[1]
+	ssh.Lock()
+	defer ssh.Unlock()
+	clusterInfo, ok := ssh.clustersMap[clusterID]
```
prachidamle: We should release the lock here after reading from the map. This method cleans up old scans which can potentially end up sleeping for retries and <defer ssh.Unlock()> wont be called until that returns. So better to unlock here.

prachidamle: Also keep the defer

pkg/controllers/user/alert/watcher/cluster_scan.go
prachidamle: Could this be just a sync handler?

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: During my testing, I found the variance/delay is generally in milliseconds, hence when rounded to Second, there is room to accommodate this.

pkg/controllers/management/cis/register.go
leodotcloud: Sure, will change to debug.

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: I will change to explicitly unlock, I don't think we can do both unlock and defer unlock

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: 👍

leodotcloud: Thanks for the catch!!!

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: yup, will move

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: This shouldn't be a problem because the checkandlaunchscan is idempotent. If we don't return an error, the clean up of the old scans would be retried until the next scheduled scan. The default is 24 hours, so no clean up for 24 hours.

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: 👍

pkg/controllers/management/cis/scheduled_scan.go
leodotcloud: 👍

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,279 @@
...
+) error {
+	cronSchedule := getCronSchedule(cluster)
+	key := fmt.Sprintf("%v%v%v%v", ScheduledScanKeyPrefix, cluster.Name, separator, cronSchedule)
+	timeAfter, next, err := getTimeAfterAndNext(cronSchedule)
+	if err != nil {
+		return err
+	}
+	clusterScanClient.Controller().EnqueueAfter("", key, timeAfter)
+	clusterInfo.nextScanAt = next.Round(time.Second)
+	logrus.Infof("scheduledScanHandler: scheduleScan: cluster %v now: %v setting nextScanAt: %v", cluster.Name, time.Now().String(), next.String())
```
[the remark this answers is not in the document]
leodotcloud: 👍

pkg/controllers/user/alert/watcher/cluster_scan.go
leodotcloud: Changed to sync handler

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,281 @@
...
+			Profile: v3.CisScanProfileTypePermissive,
+		}
+	}
+
+	nextAt := clusterInfo.nextScanAt
+	now := time.Now().Round(time.Second)
+	if !nextAt.Equal(now) && !now.After(nextAt) {
+		logrus.Debugf("scheduledScanHandler: checkAndLaunchScan: cluster %v next: %v, now: %v",
+			cluster.Name, nextAt.String(), now.String())
+		return nil, nil
```
prachidamle: when you return from here, the next scan is not scheduled, should't we be scheduling the next one to maintain the schedule?

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,281 @@
...
+		cluster.Spec.ScheduledClusterScan.ScanConfig.CisScanConfig == nil {
+		cisScanConfig = &v3.CisScanConfig{
+			Profile: v3.CisScanProfileTypePermissive,
+		}
+	}
+
+	nextAt := clusterInfo.nextScanAt
+	now := time.Now().Round(time.Second)
+	if !nextAt.Equal(now) && !now.After(nextAt) {
+		logrus.Debugf("scheduledScanHandler: checkAndLaunchScan: cluster %v next: %v, now: %v",
```
prachidamle: Can this log mention that scan is being not run or skipped

prachidamle: Oh I see that you have added !nextAt.Equal(now) && !now.After(nextAt) - With now >= nextAt, this looks good.

pkg/controllers/management/cis/scheduled_scan.go
prachidamle: Oh I see that you have added !nextAt.Equal(now) && !now.After(nextAt) - With now >= nextAt, this looks good.

prachidamle:
go.mod conflicts need to be resolved, Changes are fine.

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -86,35 +86,19 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
-	newCisScan := cis.NewCisScan(cluster, cisScanConfig)
-	cisScan, err := a.ClusterScanClient.Create(newCisScan)
+	isManual := true
+	cisScan, err := cis.LaunchScan(
+		isManual,
+		cisScanConfig,
+		cluster,
+		a.ClusterClient,
+		a.ClusterScanClient,
+		RetryIntervalInMilliseconds,
```
cjellick: knitpick: do we really need to pass in these two retry parameters? are they ever not the same thing?

pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, nil
+	}
+
+	if err := cis.ValidateClusterBeforeLaunchingScan(cluster); err != nil {
+		return nil, err
+	}
+
+	currentCronSchedule := getCronSchedule(cluster)
+	if currentCronSchedule != pastCronSchedule {
+		logrus.Infof("scheduledScanHandler: sync: cron schedule has changed for cluster: %v, hence skipping",
```
cjellick: did you mean for this to be info level or debu?

leodotcloud: Will change to debug.

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -86,35 +86,19 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
 		}
```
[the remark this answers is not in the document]
leodotcloud: I definitely had more descriptive commit messages ... they got lost somewhere during rebasing ... conflict resolution :(

pkg/controllers/user/cis/scan.go
```diff
@@ -36,3 +39,63 @@ func NewCisScan(cluster *v3.Cluster, cisScanConfig *v3.CisScanConfig) *v3.Cluste
...
+	nameTmpl := "cis-%v"
+	return NewCisScan(cluster, cisScanConfig, nameTmpl, v3.ClusterScanRunTypeManual)
+}
+
+func NewScheduledCisScan(cluster *v3.Cluster, cisScanConfig *v3.CisScanConfig) *v3.ClusterScan {
+	nameTmpl := "ss-cis-%v"
+	return NewCisScan(cluster, cisScanConfig, nameTmpl, v3.ClusterScanRunTypeScheduled)
+}
+
+func LaunchScan(
```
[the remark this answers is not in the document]
leodotcloud: Having all of them in one line goes beyond 120 char count and it's pretty hard to read code without scrolling to the right. With one param per line, the code is more readable and convenient to copy/paste/edit the lines to the place where the function is invoked. If we have some rancher coding style guide (like cattle days, couldn't find one for 2.x), I am totally fine changing/adhering to that.

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -82,6 +89,27 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
 		v3.ClusterScanConditionFailed.Message(cs, "cannot run scan on a windows cluster")
 		return cs, nil
 	}
+
+	// Wait for complete pod removal from previous scan
+	removed := false
+	for retry := NumberOfRetriesForPodRemoval; retry > 0; retry-- {
```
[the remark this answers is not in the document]
leodotcloud: Actually, I figured the framework itself will retry there is no need for EnqueueAfter. Will remove the explicit retry logic from code.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: Will clean up.

PR #25485, opened 2020-02-19
Fix automation failures

khushboo-rancher:


tests/validation/tests/v3_api/test_secrets.py
```diff
@@ -308,6 +313,7 @@ def test_rbac_secret_edit_cluster_member(remove_resource):
     assert e.value.error.code == 'NotFound'
 
     remove_resource(project)
+    remove_resource(ownerproject)
```
sowmyav27: Move all the remove_resource() before any asserts

sowmyav27: Move all the remove_resource() before any asserts

khushboo-rancher: Done

Issue #25484, opened 2020-02-19
Shibboleth - user input for cluster roles gets cleared, not allowing to add users manually


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Install Rancher 2.4 `master-head (02/19/2020)` commit id: e11d47814
2. Enable Shibboleth with OpenLDAP
3. Add a user and a group as cluster owners
4. Disable Shibboleth
5. Re-enable Shibboleth **without** OpenLDAP (OpenLDAP search will be disabled)
6. Login as a user in step 3.
7. Try to add a user search to LDAP won't work but we should be able to add an arbitrary username.

**Result:**

The dropdown input field for User/Group gets cleared out when trying to select the Cluster Role.
This doesn't happen with others SAML providers like Keycloak which allows us to add usernames manually (not selected from dropdown).

**Other details that may be helpful:**

Errors/warnings in Rancher logs

```
2020/02/19 21:30:52 [ERROR] Unknown error: invalid server config. only exactly 1 server is currently supported
2020/02/19 21:31:54 [WARNING] ldap search principals failed to connect to ldap: invalid server config. only exactly 1 server is currently supported
```

**Environment information**
- Rancher version: `2.4 master-head (02/19/2020)` commit id: e11d47814
- Installation option (single install/HA): single

izaac:
After further testing this also happens if we setup Shibboleth **without** OpenLDAP. There's no need to disable and re-enable to reproduce.

luthermonson:
https://github.com/rancher/rancher/pull/25489

luthermonson:
https://github.com/rancher/rancher/pull/25489

izaac:
This is no longer reproducible in Rancher version `2.4 master-head (02/21/2020)` commit id: 264d9235b

I was able to add users and groups when the Shibboleth is not configured with OpenLDAP search now.

PR #25489, opened 2020-02-20
shibboleth fallback to saml provider

luthermonson:
Only using the LDAP provider doesn't work everywhere when we're just using shibboleth as a SAML provider. This will short circuit that logic so we're always at least a SAML provider.

### DIFF
--- a/pkg/auth/providers/ldap/ldap_provider.go
+++ b/pkg/auth/providers/ldap/ldap_provider.go
@@ -32,6 +32,7 @@ const (
 )
 
 var (
+	errNotConfigured       = fmt.Errorf("not configured")
 	testAndApplyInputTypes = map[string]string{
 		FreeIpaName:  client.FreeIpaTestAndApplyInputType,
 		OpenLdapName: client.OpenLdapTestAndApplyInputType,
@@ -82,6 +83,10 @@ func GetLDAPConfig(authProvider common.AuthProvider) (*v3.LdapConfig, *x509.Cert
 	return ldapProvider.getLDAPConfig()
 }
 
+func IsNotConfigured(err error) bool {
+	return err == errNotConfigured
+}
+
 func (p *ldapProvider) GetName() string {
 	return p.providerName
 }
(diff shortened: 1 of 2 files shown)


cjellick:
https://github.com/rancher/rancher/issues/25484

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -82,6 +83,10 @@ func GetLDAPConfig(authProvider common.AuthProvider) (*v3.LdapConfig, *x509.Cert
 	return ldapProvider.getLDAPConfig()
 }
 
+func IsNotConfigured(err error) bool {
```
mrajashree: can we rename this to indicate Ldap search is not configured for saml?

pkg/auth/providers/ldap/ldap_provider.go
```diff
@@ -211,10 +222,16 @@ func (p *ldapProvider) getLDAPConfig() (*v3.LdapConfig, *x509.CertPool, error) {
...
-		if subLdapConfig, ok := storedLdapConfigMap[ldapConfigKey[p.providerName]]; ok {
-			storedLdapConfigMap = subLdapConfig.(map[string]interface{})
+		subLdapConfig, ok := storedLdapConfigMap[ldapConfigKey[p.providerName]]
+		if !ok {
+			return nil, nil, errNotConfigured
 		}
+
+		storedLdapConfigMap = subLdapConfig.(map[string]interface{})
 		mapstructure.Decode(storedLdapConfigMap, storedLdapConfig)
+		if len(storedLdapConfig.Servers) != 1 {
```
mrajashree: should we only return not configured error if `if len(storedLdapConfig.Servers) < 1 ` and let ldap.Connect return error if `len(storedLdapConfig.Servers) > 1`?

luthermonson: not my logic, grabbed from this: https://github.com/rancher/rancher/blob/master/pkg/auth/providers/common/ldap/ldap_util.go#L42-L44

pkg/auth/providers/ldap/ldap_provider.go
luthermonson: it could work for both, we just don't care in the non saml search thing. im a bit against renaming as we could add this same error into more parts of getLdapConfig

pkg/auth/providers/ldap/ldap_provider.go
mrajashree: yeah but since we're returning "NotConfigured" error, here we could return it if no server input was provided whereas NewLDAPConn stays as is, but not an important change :)

Issue #25462, opened 2020-02-18
Istio traffic routing tabs disappeared from UI in master


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Enable Istio in Rancher master
2. Navigate to Resources > Istio

**Result:**
There should be 5 tabs across the top of that page. It SHOULD look like:
![image](https://user-images.githubusercontent.com/55997940/74782966-ce059880-5261-11ea-82b1-2eea80be6217.png)

But what we actually see is:
![image](https://user-images.githubusercontent.com/55997940/74783128-3eacb500-5262-11ea-9b67-2e6b23e3c8e1.png)

In master, I do not see "Gateways", "Destination Rules", or "Virtual Services"

**Other details that may be helpful:**
First screenshot above is the correct view from Rancher v2.3.5. Changing ui-index in Rancher v2.3.5 to be the same as master (https://releases.rancher.com/ui/latest2/index.html) does NOT reproduce this issue.

**Environment information**
Single node install, master Rancher version `580ab5bb4`

**Cluster information**
3 nodes, all roles, 8gb 4vcpu each. Hosted on Linode.
 Kubernetes Version: v1.17.2

rancher-max:
**Interesting to note: using the URLs directly takes me to the appropriate pages:**
https://${RANCHER_SERVER_URL}/p/${CLUSTER:PROJECT IDs}/istio/project-istio/gateways
https://${RANCHER_SERVER_URL}/p/${CLUSTER:PROJECT IDs}/istio/project-istio/destination-rules
https://${RANCHER_SERVER_URL}/p/${CLUSTER:PROJECT IDs}/istio/project-istio/virtual-services

codyrancher:
The display of those tabs are now behind a feature flag
<img width="1272" alt="Screen Shot 2020-02-20 at 8 17 45 AM" src="https://user-images.githubusercontent.com/55104481/74948567-b7725500-53b9-11ea-95f1-e539d913b59e.png">

The code was committed in october https://github.com/rancher/ui/commit/5562d5f65d12ee2c65b944bc34ea5b3109179410

I don't think this is in error.

rancher-max:
[...]

PR #25494, opened 2020-02-20
Fix feature def

rmweir:
**Problem:**
Feature formatter assigns default from where default used to be located. This causes value to not be properly set to the default.

**Solution:**
Use default from feature status.

**Issue:**
https://github.com/rancher/rancher/issues/25462

### DIFF
--- a/pkg/api/customization/feature/feature.go
+++ b/pkg/api/customization/feature/feature.go
@@ -38,9 +38,22 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
 
 func Formatter(request *types.APIContext, resource *types.RawResource) {
 	if request.Method == http.MethodGet {
-		if resource.Values["value"] == nil {
-			// if value is nil, then this ensure default value will be used
-			resource.Values["value"] = resource.Values["default"]
-		}
+		resource.Values["value"] = getEffectiveValue(resource)
 	}
+}
+
+func getEffectiveValue(resource *types.RawResource) bool {
+	if val := resource.Values["value"]; val != nil {
+		val, _ := val.(bool)
+		return val
+	}
+
+	var val bool
+	// if value is nil, then this ensure default value will be usedq
+	status, ok := resource.Values["status"].(map[string]interface{})
+	if ok {
+		val, _ = status["default"].(bool)
+	}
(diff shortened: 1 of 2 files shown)


pkg/api/customization/feature/feature.go
```diff
@@ -40,7 +40,13 @@ func Formatter(request *types.APIContext, resource *types.RawResource) {
 	if request.Method == http.MethodGet {
 		if resource.Values["value"] == nil {
 			// if value is nil, then this ensure default value will be used
-			resource.Values["value"] = resource.Values["default"]
+			val := false
+			status, ok := resource.Values["status"].(map[string]interface{})
+			if ok {
+				val, _ = status["default"].(bool)
```
cbron: bool type is guaranteed here ?

rmweir: as long as _ is there, it will use the zero value (false) if it is not a bool.

pkg/api/customization/feature/feature.go
```diff
@@ -40,7 +40,13 @@ func Formatter(request *types.APIContext, resource *types.RawResource) {
 	if request.Method == http.MethodGet {
 		if resource.Values["value"] == nil {
 			// if value is nil, then this ensure default value will be used
-			resource.Values["value"] = resource.Values["default"]
+			val := false
```
dramich: update to use `var val bool`

tests/integration/suite/test_features.py
```diff
@@ -27,3 +27,9 @@ def test_can_list(admin_mc, user_mc, remove_resource):
     admin_client = admin_mc.client
     admin_client.list_feature()
     assert True
+
+
+def test_default_set_as_value(admin_mc):
+    """Asserts feature values reflect default prior to being toggled"""
+    istio_feature = admin_mc.client.by_id_feature("istio-virtual-service-ui")
+    assert istio_feature["value"] == istio_feature["status"]["default"]
```
dramich: Update to validate that the value is not None, this is not guaranteed though as this feature being flipped elsewhere will cause a false positive. Potentially marking as non-parallel and removing the value to ensure only the default will get filled in.

PR #25520, opened 2020-02-21
Enhance log message for kubectl calls

dramich:


superseb:
This is also done in https://github.com/rancher/rancher/pull/24953, can we decide what to keep so we can close the other?

dramich:
I prefer this style for the error and it also covers the delete message. The other PR should be kept to update the loop and adding logs.

PR #25521, opened 2020-02-21
Remove redundant checks

rmweir:
**Problem:**
CanAccess was is never evaluated, and that is where the error code was converted.

**Solution:**
Remove redundant check.  Return error with correct error code if access check fails.

### DIFF
--- a/pkg/api/customization/cluster/validator.go
+++ b/pkg/api/customization/cluster/validator.go
@@ -107,14 +107,14 @@ func (v *Validator) validateEnforcement(request *types.APIContext, data map[stri
 		return httperror.NewFieldAPIError(httperror.MissingRequired, "", "A clusterTemplateRevision to create a cluster")
 	}
 
-	canAccess, err := v.isTemplateAccessible(request, &spec)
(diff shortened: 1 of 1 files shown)


prachidamle:
@rmweir fix is good, build seems to have golint errors

pkg/api/customization/cluster/validator.go
```diff
@@ -107,14 +107,14 @@ func (v *Validator) validateEnforcement(request *types.APIContext, data map[stri
 		return httperror.NewFieldAPIError(httperror.MissingRequired, "", "A clusterTemplateRevision to create a cluster")
 	}
 
-	canAccess, err := v.isTemplateAccessible(request, &spec)
+	err = v.accessTemplate(request, &spec)
 	if err != nil {
+		if httperror.IsForbidden(err) || httperror.IsNotFound(err) {
+			return httperror.NewFieldAPIError(httperror.NotFound, "", "The clusterTemplateRevision is not found")
```
dramich: Use `NewAPIError` since field isn't being set

PR #25523, opened 2020-02-21
Update prometheus operator

daxmc99:
This updates the prometheus operator to avoid incompatibility errors
with regards to go modules

Depends on types: 
https://github.com/rancher/types/pull/1099


daxmc99:
Already in

Issue #25526, opened 2020-02-22
Not able to add group at project level


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Install Rancher 2.4 `master-head (02/21/2020)` commit id: 264d9235b
- Create a cluster
- Add a new user as Cluster Member for this cluster. (A user with a group)
- Login a the Cluster Member created
- Create a Project
- Edit the Project and add Member, click in the dropdown to list the group(s) the user belongs to.
- Select the Role
- Save

**Result:**

```
Validation failed in API: must target a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]
```

<img width="1710" alt="Screen Shot 2020-02-21 at 6 03 15 PM" src="https://user-images.githubusercontent.com/27659/75093569-0a761480-5540-11ea-9c5c-0fc0e2797898.png">

Happens with Shibboleth without LDAP where I pick the group from the arrow button.
Also this happens when selecting a group from the new Shibboleth OpenLDAP search dropdown.

**Other details that may be helpful:**

Adding group to cluster level works.

Headers
```
POST /v3/projectroletemplatebinding HTTP/1.1
Host: x.x.x.x
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:73.0) Gecko/20100101 Firefox/73.0
Accept: application/json
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate, br
content-type: application/json
x-api-action-links: actionLinks
x-api-no-challenge: true
x-api-csrf: a19f3bbbc0
Content-Length: 244
Origin: https://x.x.x.x
DNT: 1
Connection: keep-alive
Referer: https://96.126.101.112/c/c-sxcml/projects-namespaces/project/c-sxcml:p-4ffj4
Cookie: CSRF=a19f3bbbc0; R_SESS=token-7w7sm:wltsfm7pgnrnsghx449k8sz6txxftpcm58qh5fjw5hvshq85xm4rx2
```
POST

```json
{"type":"projectRoleTemplateBinding","subjectKind":"Group","userId":"","projectRoleTemplateId":"","projectId":"c-sxcml:p-4ffj4","groupPrincipalId":"shibboleth_group://cn=testgroup2,ou=groups,dc=example,dc=org","roleTemplateId":"project-member"}
```

Response:

```json
[...]

PR #25530, opened 2020-02-22
Convert PRTB/CRTB subject fields to strings

rmweir:
**Problem:**
Empty strings are treated as valid subjects. 

**Solution:**
Treat subject fields with empty strings as missing. Make null and empty values have identical results.

**Issue:**
https://github.com/rancher/rancher/issues/25526

### DIFF
--- a/pkg/api/customization/roletemplatebinding/validator.go
+++ b/pkg/api/customization/roletemplatebinding/validator.go
@@ -55,8 +55,13 @@ func (v *validator) validator(request *types.APIContext, schema *types.Schema, d
 		return nil
 	}
 
-	hasUserTarget := data["userId"] != nil || data["userPrincipalId"] != nil
-	hasGroupTarget := data["groupId"] != nil || data["groupPrincipalId"] != nil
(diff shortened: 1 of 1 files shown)


pkg/api/customization/roletemplatebinding/validator.go
```diff
@@ -55,8 +55,13 @@ func (v *validator) validator(request *types.APIContext, schema *types.Schema, d
 		return nil
 	}
 
-	hasUserTarget := data["userId"] != nil || data["userPrincipalId"] != nil
-	hasGroupTarget := data["groupId"] != nil || data["groupPrincipalId"] != nil
+	userID, _ := data["userId"].(string)
```
cbron: This will panic if interface is nil: https://play.golang.com/p/tncQH0k_qjh

rmweir: This is safe because it is doing the safety check. I am just throwing away the ok, but userID will get the zero value if the field is nil. See: https://play.golang.com/p/trIe8GXwDMV

Issue #22232, opened 2019-08-16
Allow Kubernetes downgrade when restoring etcd snapshot


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Feature request

**Description**
The Kubernetes upgrade documentation listed at https://rancher.com/docs/rancher/v2.x/en/cluster-admin/editing-clusters/#upgrading-kubernetes , it is recommended backing up (taking an etcd snapshot) before doing an upgrade. This implies that if you want to revert an upgrade, you should be able to restore the etcd snapshot and that will put back your Kubernetes cluster to the same version before the upgrade. However, if you attempt to restore the etcd cluster, it will not revert the upgrade, but only restore the state of the etcd data and Rancher will still attempt to upgrade the Kubernetes cluster.

User should be able to revert an upgrade and return their cluster to the version of Kubernetes  that corresponds to when the etcd snapshot was taken. For example, a user is on Kubernetes 1.14.0, takes an etcd snapshot, then upgrades to 1.14.5. The user should be able to restore the etcd snapshot and downgrade Kubernetes to v1.14.0. This would involve reverting all Kubernetes components - kube-apiserver, kube-controller, kube-proxy, kube-scheduler, and kubelet, to the previous version.

ajfriesen:
Did run into this today.

Wanted to test an upgrade. Upgrade failed due to some bug with kube-dns.
We had a snapshot and tried to restore that snapshot.

This did not work properly.

For the first few seconds `kubectl get nodes` did show the kubernetes version from the snapshot before the upgrade but suddenly the view switched to the newthe kubernetes version which we wanted to update to (testing at least).

Also `kubectl get nodes` and the rancher ui did show us different workes which I wonder about.
We habd workers in `kubectl get nodes` which we did not have in rancher ui. 
I thought it could not be different technically.

chrisbulgaria:
yes - this would be a great enhancement !

cloudnautique:
Distilling this down a bit:

1. A user needs to be able to optionally take a snapshot when initiating the upgrade. This snapshot would be tied to a specific Kubernetes version.
2. The user should have the option to rollback the upgrade. This means restoring the cluster, including all kubernetes components and etcd database back to the configuration pre-upgrade.

Since etcd snapshot will have the Kubernetes version data tied to it going forward, users should be able to see which version of k8s the backup was taken at.

soumyalj:
Tested with master-head branch. 
[...]

PR #25545, opened 2020-02-24
RKE Config Added to Backups and Restore Options for version/config

luthermonson:
The snapshot process only stores some basic metadata, this will add the entire cluster object in order to help track the state of the cluster when the backup is made. A restoreRkeConfig was added to the etcd restore action which supports "all" and "kubernetesVersion" meaning during the restore you can choose to restore the entirety of the cluster config or leave the cluster config as is and only restore the kubernetes version.
#22232 
types: https://github.com/rancher/types/pull/1101


pkg/api/customization/cluster/actions_etcd.go
```diff
@@ -106,8 +112,32 @@ func (a ActionHandler) RestoreFromEtcdBackupHandler(actionName string, action *t
 		}
 	}
 
+	if input.RestoreRkeConfig != "" && backup.Status.ClusterObject == "" {
```
cbron: Aren't you potentially incorrectly disallowing the kubeVersion part below since it doesn't use the status.clusterObject ?

luthermonson: both version and whole config are optional, pass in "" or no param at all if you want the original feature. that param not passed is original behavior, pass "all' you get whole rke config, pass "kubernetesVersion" for restoring just the vk8s version

luthermonson: also this is checking to see if you are trying to restore the version/config AND the backup was taken before this snapshot feature was upgraded

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -291,7 +305,65 @@ func NewBackupObject(cluster *v3.Cluster, manual bool) *v3.EtcdBackup {
...
+	gz := gzip.NewWriter(&gzCluster)
+
+	_, err = gz.Write([]byte(jsonCluster))
+	if err != nil {
+		return "", err
+	}
+
+	if err = gz.Flush(); err != nil {
+		return "", err
+	}
```
prachidamle: It is safer to call gz.Flush() in a defer func() to ensure flush is called even if there is an error in Write.

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -291,7 +305,65 @@ func NewBackupObject(cluster *v3.Cluster, manual bool) *v3.EtcdBackup {
...
+
+	if err := gz.Close(); err != nil {
```
prachidamle: Same as above, better to have this in defer

go.mod
```diff
@@ -86,7 +86,7 @@ require (
 	github.com/rancher/rke v1.1.0-rc4.0.20200205094053-56d4c1b93702
 	github.com/rancher/security-scan v0.1.5
 	github.com/rancher/steve v0.0.0-20200220230643-5b94b82bc6c8
-	github.com/rancher/types v0.0.0-20200224204201-0cbd16146caf
```
daxmc99: Where is the associated types PR?

pkg/api/customization/cluster/actions_etcd.go
```diff
@@ -38,7 +40,10 @@ func (a ActionHandler) BackupEtcdHandler(actionName string, action *types.Action
 		return errors.Wrapf(err, "failed to get Cluster by ID %s", apiContext.ID)
 	}
 
-	newBackup := etcdbackup.NewBackupObject(cluster, true)
+	newBackup, err := etcdbackup.NewBackupObject(cluster, true)
+	if err != nil {
+		return err
```
prachidamle: We should also return a response for the action using apiContext.WriteResponse  and return error using errors.Wrapf with a message that can help debugging

pkg/api/customization/cluster/actions_etcd.go
prachidamle: Is there a types PR to add the "RestoreRkeConfig" to input?

luthermonson: yes already merged https://github.com/rancher/types/pull/1101

prachidamle: Ah I see it in the PR desc, nevermind

pkg/api/customization/cluster/actions_etcd.go
luthermonson: yup ill add

pkg/controllers/management/etcdbackup/etcdbackup.go
luthermonson: ya i have a better fix removing the flush. agree.

pkg/api/customization/cluster/actions_etcd.go
```diff
@@ -106,8 +114,32 @@ func (a ActionHandler) RestoreFromEtcdBackupHandler(actionName string, action *t
...
 
+	if input.RestoreRkeConfig != "" && backup.Status.ClusterObject == "" {
+		// attempting to restore rke config and the backup does not contain data, probably pre 2.4 backup
+		return httperror.NewAPIError(httperror.MethodNotAllowed,
+			fmt.Sprintf("unable to restore RKE config, backup contains no cluster object: %s", input.EtcdBackupID))
+	}
+
+	// backup was taken in 2.4+ and has content
+	switch strings.ToLower(input.RestoreRkeConfig) {
+	case "kubernetesVersion":
```
dramich: This needs to be lower case

luthermonson: ha! nice catch.. thanks3

Issue #23543, opened 2019-10-18
Show expired cert message for local cluster


mattmattox:
**What kind of request is this (question/bug/enhancement/feature request):**
Feature Request

**Steps to reproduce (least amount of steps as possible):**
Show the message "This cluster has certs that are expiring or have expired." for the local cluster like we do for downstream custom clusters.

**Result:**
Message: This cluster has certs that are expiring or have expired.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

leodotcloud:
Also, when rancher is starting up, it should check if the certs being used are expiring soon or already expired and print a warning message in the logs.

cbron:
Fix is limited to RKE clusters, which is fine as thats all we officially support.

khushboo-rancher:
I've tested this feature in master-head ```37363ba67``` HA set up (3 nodes all roles) with a certificate expiring before 30 days as below steps
1. Created nodes and self signed certificate with less than 30 days expiration date.
2. Then ran command ``` rke up --custom-certs --cert-dir=custom_certs``` to make rke up for nodes.
3. Set up the HA env 
The cluster API has certificate details as below.
```
certificatesExpiration": {
"kube-apiserver": {
"expirationDate": "2021-02-28T00:17:00Z",
"type": "/v3/schemas/certExpiration"
},
"kube-apiserver-proxy-client": {
"expirationDate": "2021-02-28T00:22:00Z",
"type": "/v3/schemas/certExpiration"
},
"kube-ca": {
"expirationDate": "2020-03-20T00:20:37Z",
"type": "/v3/schemas/certExpiration"
},
"kube-controller-manager": {
"expirationDate": "2021-02-28T00:17:00Z",
"type": "/v3/schemas/certExpiration"
},
"kube-etcd-13-58-142-13": {
"expirationDate": "2021-02-28T00:20:00Z",
"type": "/v3/schemas/certExpiration"
},
"kube-etcd-18-222-215-23": {
"expirationDate": "2021-02-28T00:21:00Z",
"type": "/v3/schemas/certExpiration"
},
"kube-etcd-18-224-2-168": {
"expirationDate": "2021-02-28T00:25:00Z",
"type": "/v3/schemas/certExpiration"
},
"kube-node": {
"expirationDate": "2021-02-28T00:18:00Z",
"type": "/v3/schemas/certExpiration"
},
[...]

PR #25547, opened 2020-02-24
Adds cert expiration warnings for local cluster

cbron:
Issue: https://github.com/rancher/rancher/issues/23543

Implements management controller to check for RKE's `full-cluster-state` configMap for expiring certs. Re-used lots of the user-controller code which checked downstream clusters, so I moved that into pkg to share. 

### Blockers 
**Requires UI update**: The UI shows the warning but it also shows option to rotate certificates, which the local cluster can't do because it doesn't have an RKE Config. UI should be able to check and see if rotate action exists on the cluster, and if not then [hide the link](https://github.com/rancher/ui/blob/master/lib/monitoring/addon/components/cluster-dashboard/template.hbs#L1-L3). 

Screen Shot 2020-02-24 at 4 28 48 PM

### DIFF
--- a/pkg/controllers/management/certsexpiration/certsexpiration.go
+++ b/pkg/controllers/management/certsexpiration/certsexpiration.go
@@ -1,0 +1,82 @@
+package certsexpiration
+
+import (
+	"context"
+	"reflect"
+	"time"
+
+	"github.com/rancher/rancher/pkg/rkecerts"
+	rkeCluster "github.com/rancher/rke/cluster"
+	v1 "github.com/rancher/types/apis/core/v1"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// This controller handles cert expiration for local cluster only
+func Register(ctx context.Context, management *config.ManagementContext) {
+	c := &certsExpiration{
+		clusters:        management.Management.Clusters(""),
+		configMapLister: management.Core.ConfigMaps("kube-system").Controller().Lister(),
+	}
+	management.Management.Clusters("").AddHandler(ctx, "certificate-expiration", c.sync)
+}
+
+type certsExpiration struct {
+	clusters        v3.ClusterInterface
+	configMapLister v1.ConfigMapLister
+}
+
+func (c *certsExpiration) sync(key string, cluster *v3.Cluster) (runtime.Object, error) {
+	if cluster == nil || cluster.Name != "local" {
+		return cluster, nil // We are only checking local cluster
+	}
+	cm, err := c.configMapLister.Get("kube-system", rkeCluster.FullStateConfigMapName)
+	if err != nil {
+		if k8sErrors.IsNotFound(err) {
+			return cluster, nil // not an rke cluster, nothing we can do
+		}
+		return cluster, err
+	}
+	certBundle, err := rkecerts.CertBundleFromConfig(cm)
+	if err != nil {
+		return cluster, err
+	}
+	rkecerts.CleanCertificateBundle(certBundle)
+
+	certsExpInfo := map[string]v3.CertExpiration{}
+	for certName, certObj := range certBundle {
+		info, err := rkecerts.GetCertExpiration(certObj.CertificatePEM)
+		if err != nil {
+			logrus.Debugf("failed to get expiration date for certificate [%s] for local cluster: %v", certName, err)
+			continue
+		}
+		certsExpInfo[certName] = info
+		err = logCertExpirationWarning(certName, info)
+		if err != nil {
+			logrus.Warnf("certificate [%s] from local cluster has or will expire and date is corrupted: %v", certName, err)
+			continue
+		}
+	}
+	// Update certExpiration on cluster obj in order for it to display in API, and the UI if expiring
+	if !reflect.DeepEqual(cluster.Status.CertificatesExpiration, certsExpInfo) {
+		cluster.Status.CertificatesExpiration = certsExpInfo
+		return c.clusters.Update(cluster)
+	}
+	return cluster, nil
+}
+
+func logCertExpirationWarning(name string, certExp v3.CertExpiration) error {
+	date, err := time.Parse(time.RFC3339, certExp.ExpirationDate)
+	if err != nil {
+		return err
+	}
+	if time.Now().UTC().After(date) { // warn if expired
+		logrus.Warnf("Certificate from local cluster has expired: %s", name)
+	} else if time.Now().UTC().AddDate(0, 1, 0).After(date) { // warn if within a month
+		logrus.Warnf("Certificate from local cluster will expire soon: %s", name)
+	}
+	return nil
+}
--- a/pkg/rkecerts/expiration.go
+++ b/pkg/rkecerts/expiration.go
@@ -1,0 +1,59 @@
+package rkecerts
+
+import (
+	"encoding/json"
+	"errors"
+	"strings"
+	"time"
+
+	rkeCluster "github.com/rancher/rke/cluster"
+	"github.com/rancher/rke/pki"
+	"github.com/rancher/rke/pki/cert"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	v1 "k8s.io/api/core/v1"
+)
+
+func CleanCertificateBundle(certs map[string]pki.CertificatePKI) {
+	for name := range certs {
+		if strings.Contains(name, "token") || strings.Contains(name, "header") || strings.Contains(name, "admin") {
+			delete(certs, name)
+		}
+	}
+}
+
+func GetCertExpiration(c string) (v3.CertExpiration, error) {
+	date, err := GetCertExpirationDate(c)
+	if err != nil {
+		return v3.CertExpiration{}, err
+	}
+	return v3.CertExpiration{
+		ExpirationDate: date.Format(time.RFC3339),
+	}, nil
+}
+
+func GetCertExpirationDate(c string) (*time.Time, error) {
+	certs, err := cert.ParseCertsPEM([]byte(c))
+	if err != nil {
+		return nil, err
+	}
+	if len(certs) == 0 {
+		return nil, errors.New("no valid certs found")
+	}
+	return &certs[0].NotAfter, nil
+}
+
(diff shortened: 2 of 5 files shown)


pkg/rkecerts/expiration.go
```diff
@@ -0,0 +1,56 @@
...
+		ExpirationDate: date.Format(time.RFC3339),
+	}, nil
+}
+
+func GetCertExpirationDate(c string) (*time.Time, error) {
+	certs, err := cert.ParseCertsPEM([]byte(c))
+	if err != nil {
+		return nil, err
+	}
+	return &certs[0].NotAfter, nil
```
StrongMonkey: Probably index check for safety

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,77 @@
...
+type certsExpiration struct {
+	clusterClient   v3.ClusterInterface
+	configMapGetter coreV1.ConfigMapsGetter
+}
+
+func (c *certsExpiration) sync(key string, cluster *v3.Cluster) (runtime.Object, error) {
+	if cluster.Name != "local" {
+		return cluster, nil // We are only checking local cluster
+	}
+	cm, err := c.configMapGetter.ConfigMaps("kube-system").Get(rkeCluster.FullStateConfigMapName, metaV1.GetOptions{})
```
StrongMonkey: Use cache maybe?

cbron: @StrongMonkey all comments fixed

pkg/rkecerts/expiration.go
```diff
@@ -0,0 +1,59 @@
...
+	rkeCluster "github.com/rancher/rke/cluster"
+	"github.com/rancher/rke/pki"
+	"github.com/rancher/rke/pki/cert"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	v1 "k8s.io/api/core/v1"
+)
+
+func CleanCertificateBundle(certs map[string]pki.CertificatePKI) {
+	for name := range certs {
+		if strings.Contains(name, "client") ||
```
luthermonson: nit: little hard to read what this is doing, maybe put the boolean checks in another func

pkg/rkecerts/expiration.go
luthermonson: he already checked if len == 0, this should be guaranteed shouldnt it? do we need to handle if len > 1?

luthermonson:
controller logic seems good, nit picks against the code you copy/pasted.. sorry :/

pkg/rkecerts/expiration.go
cbron: I fixed it after his comment, this is resolved.

cbron:
@galal-hussein can you do a quick review here just for the overall approach with rke and certs on local cluster

galal-hussein:
> @galal-hussein can you do a quick review here just for the overall approach with rke and certs on local cluster

Sure will review today

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,76 @@
...
+	management.Management.Clusters("").AddHandler(ctx, "certificate-expiration", c.sync)
+}
+
+type certsExpiration struct {
+	clusterClient   v3.ClusterInterface
+	configMapLister v1.ConfigMapLister
+}
+
+func (c *certsExpiration) sync(key string, cluster *v3.Cluster) (runtime.Object, error) {
+	if cluster.Name != "local" {
```
galal-hussein: this can cause a panic, we need to make sure that cluster isn't nil first

galal-hussein: also a similar check should be added here i think https://github.com/rancher/rancher/blob/master/pkg/controllers/user/certsexpiration/certsexpiration.go#L47

pkg/rkecerts/expiration.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, err
+	}
+	if len(certs) == 0 {
+		return nil, errors.New("certificate contains invalid dates")
+	}
+	return &certs[0].NotAfter, nil
+}
+
+func CertBundleFromConfig(cm *v1.ConfigMap) (map[string]pki.CertificatePKI, error) {
+	rawCerts, ok := cm.Data[rkeCluster.FullStateConfigMapName]
```
galal-hussein: this also can cause a panic i think if the cm is nil

pkg/rkecerts/expiration.go
```diff
@@ -0,0 +1,59 @@
...
+	if !ok {
+		return nil, errors.New("full-cluster-state configmap does not contain data")
+	}
+	rkeFullState := &rkeCluster.FullState{}
+	err := json.Unmarshal([]byte(rawCerts), rkeFullState)
+	if err != nil {
+		return nil, err
+	}
+	return rkeFullState.DesiredState.CertificatesBundle, nil
```
galal-hussein: we should return Current state not the desired state since it reflects the current certificates actually deployed on the nodes

pkg/rkecerts/expiration.go
galal-hussein: this actually was another bug, basically api-proxy-client is a legit cert and shouldn't be skipped from the check

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,76 @@
...
+	if !reflect.DeepEqual(cluster.Status.CertificatesExpiration, certsExpInfo) {
+		toUpdate := cluster.DeepCopy()
+		toUpdate.Status.CertificatesExpiration = certsExpInfo
+		return c.clusterClient.Update(toUpdate)
+	}
+	return cluster, nil
+}
+
+func logCertExpirationWarning(name string, certExp v3.CertExpiration) {
+	date, _ := time.Parse(time.RFC3339, certExp.ExpirationDate)
```
galal-hussein: nit: the code should check for parse error just in case

pkg/controllers/management/certsexpiration/certsexpiration.go
cbron: This one yes, but linked user one should be ok since it's checking `cluster == nil `, unless you wan to check `c.ClusterName` ? That code has been in prod for a long time so should be fine.

pkg/rkecerts/expiration.go
cbron: removed client from check

pkg/controllers/management/certsexpiration/certsexpiration.go
galal-hussein: thats cool then, thanks!

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,84 @@
...
+	v1 "github.com/rancher/types/apis/core/v1"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// This controller handles cert expiration for local cluster only
+func Register(ctx context.Context, management *config.ManagementContext, clusterManager *clustermanager.Manager) {
```
dramich: clusterManager isn't used

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,84 @@
...
+	c := &certsExpiration{
+		clusterClient:   management.Management.Clusters(""),
+		configMapLister: management.Core.ConfigMaps("kube-system").Controller().Lister(),
+	}
+	management.Management.Clusters("").AddHandler(ctx, "certificate-expiration", c.sync)
+}
+
+type certsExpiration struct {
+	clusterClient   v3.ClusterInterface
```
dramich: Update to `clusters`, the naming conventions should be `<type>s` for the interface or `<type>Lister`

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,84 @@
...
+		certsExpInfo[certName] = info
+		err = logCertExpirationWarning(certName, info)
+		if err != nil {
+			logrus.Warnf("certificate [%s] from local cluster has or will expire and date is corrupted: %v", certName, err)
+			continue
+		}
+	}
+	// Update certExpiration on cluster obj in order for it to display in API, and the UI if expiring
+	if !reflect.DeepEqual(cluster.Status.CertificatesExpiration, certsExpInfo) {
+		toUpdate := cluster.DeepCopy()
```
dramich: No need to deepCopy here, the cluster didn't come out of a cache

pkg/rkecerts/expiration.go
```diff
@@ -0,0 +1,59 @@
...
+	}, nil
+}
+
+func GetCertExpirationDate(c string) (*time.Time, error) {
+	certs, err := cert.ParseCertsPEM([]byte(c))
+	if err != nil {
+		return nil, err
+	}
+	if len(certs) == 0 {
+		return nil, errors.New("certificate contains invalid dates")
```
dramich: Can the error be updated to clarify no valid certs were found and it's not a date issue

pkg/rkecerts/expiration.go
dramich: What about Luther's question? If len > 1 what do?

dramich:
Thanks for the tests

pkg/rkecerts/expiration.go
cbron: Technically we could loop and check all of them, but theres no reason one pem would have multiple certs with diff expiration dates, rke requires [specific separated pems](https://rancher.com/docs/rke/latest/en/installation/certs/#using-custom-certificates). This code was pre-existing, I just moved it.

pkg/controllers/management/certsexpiration/certsexpiration.go
```diff
@@ -0,0 +1,82 @@
...
+func Register(ctx context.Context, management *config.ManagementContext) {
+	c := &certsExpiration{
+		clustersClient:  management.Management.Clusters(""),
+		configMapLister: management.Core.ConfigMaps("kube-system").Controller().Lister(),
+	}
+	management.Management.Clusters("").AddHandler(ctx, "certificate-expiration", c.sync)
+}
+
+type certsExpiration struct {
+	clustersClient  v3.ClusterInterface
```
dramich: This should just be `clusters`

dramich:
LGTM

PR #25555, opened 2020-02-25
Add support for multi-arc image mirrorig

StrongMonkey:


StrongMonkey:
Address https://github.com/rancher/rancher/issues/21221.

StrongMonkey:
@cloudnautique @paraglade changes to pushed to address a bug with tagging multiple arch.

scripts/docker-pull-pusher.sh
```diff
@@ -1,21 +1,30 @@
 #!/bin/sh
 set -e
+set -x
```
cloudnautique: I don't think this should be on by default.

StrongMonkey:
Per @ibuildthecloud 's suggestion. Going to take a different approach.

PR #25598, opened 2020-02-26
self-healing nodepool tests

izaac:
Added 4 tests.

Using cloud provider:

-  Terminate node in cloud provider should remove node from Rancher with cloud provider. (deleteNotReadyAfterSecs = 0)
- Stop node in cloud provider should remove node from Rancher with cloud provider. (deleteNotReadyAfterSecs = 10)

Without cloud provider:

- Stop node in cloud provider should not remove node from Rancher without cloud provider. 
(deleteNotReadyAfterSecs = 0)
- Stop node in cloud provider should remove node from Rancher without cloud provider. (deleteNotReadyAfterSecs = 10)


izaac:
I will revisit this in a new PR.

Issue #25504, opened 2020-02-20
Keycloak - principalType group not present in principals on first login. After a second login the group principalType is added.


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Setup a new Rancher 2.4 install. I used commit id: 8a68bd75a
- Configure Keycloak auth with `testadmin1`
- Make sure there's another user `testuser2` which is part of a group (check the Keycloak management console)
- Logout from Rancher `testadmin1`
- Login using `testuser2`

**Result:**

- checking `/v3/principals` there's no `principalType` group for the user.
- Log out and login as `testuser2` again.
- `/v3/principals` for the user will show the group `principalType`

izaac:
Rancher version: v2.4.0-rc1 (ca71db485)

- Validated the principalType: group are retrieved as expected on users first login in Keycloak Auth setup.
- Users members of single and multiple groups are shown respectively, on group entry per Keycloak group.

Issue #25479, opened 2020-02-19
Shibboleth - principalType group not present in principals on first login. After a second login the group principalType is added.


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Setup a new Rancher 2.4 install. I used commit id: e11d47814
- Configure Shibboleth with OpenLDAP
- Create a cluster and add `group1` as owner. 
- logout from the Shibboleth admin and login as `testuser2` or other member of `group1`
  - This would be the first time testuser2 logs into Rancher

**Result:**

- At this point you won't be able to see the cluster even though `testuser2` belongs to `group1`
- checking `/v3/principals` there's no `principalType` group for the user.
- Log out and login as `testuser2` again.
- The cluster will be visible this time
- `/v3/principals` for the user will show the group `principalType`

**Other details that may be helpful:**

First login `/v3/principals`

```json
{

    "type": "collection",
    "links": {
        "self": "…/v3/principals"
    },
    "actions": {
        "search": "…/v3/principals?action=search"
    },
    "pagination": {
        "limit": 1000
    },
    "sort": {
        "order": "asc",
        "reverse": "…/v3/principals?order=desc",
        "links": {
            "loginName": "…/v3/principals?sort=loginName",
            "name": "…/v3/principals?sort=name",
            "principalType": "…/v3/principals?sort=principalType",
            "profilePicture": "…/v3/principals?sort=profilePicture",
            "profileURL": "…/v3/principals?sort=profileURL",
            "provider": "…/v3/principals?sort=provider",
            "uuid": "…/v3/principals?sort=uuid"
        }
    },
    "filters": {
        "created": null,
        "creatorId": null,
        "id": null,
        "loginName": null,
        "me": null,
        "memberOf": null,
        "name": null,
        "principalType": null,
        "profilePicture": null,
        "profileURL": null,
        "provider": null,
        "removed": null,
        "uuid": null
    },
    "resourceType": "principal",
    "data": [
        {
            "baseType": "principal",
            "created": null,
            "creatorId": null,
            "id": "shibboleth_user://testuser2",
            "links": {
                "self": "…/v3/principals/shibboleth_user:%2F%2Ftestuser2"
            },
            "loginName": "testuser2",
            "me": true,
            "memberOf": false,
            "name": "testuser2",
            "principalType": "user",
            "provider": "shibboleth",
            "type": "principal"
        }
    ]

}
```

Second login:

[...]

PR #25602, opened 2020-02-26
SAML Logins Group Principals Not Created on First Login

luthermonson:
**Problem**
When a first time user logs in via SAML their group principals are not set, subsequent logins it does work. You can easily recreate this by logging in with a new user and going to /v3/principals, logout and login again and now groups will show up. You can delete the user via an Admin user and recreate this every time.

**Solution**
Logic is a little off in tokens.UserAttributeCreateOrUpdate. The create logic is under the update logic and the SAML user will get caught in the update logic and ultimately return without saving the UserAttribute.

### DIFF
--- a/pkg/auth/tokens/manager.go
+++ b/pkg/auth/tokens/manager.go
@@ -573,23 +573,19 @@ func (m *Manager) UserAttributeCreateOrUpdate(userID, provider string, groupPrin
 		return err
 	}
 
-	// Exists, just update if necessary
-	attribs = attribs.DeepCopy()
-
-	if m.UserAttributeChanged(attribs, provider, groupPrincipals) {
+	if needCreate {
 		attribs.GroupPrincipals[provider] = v3.Principals{Items: groupPrincipals}
-		if !needCreate {
-			_, err := m.userAttributes.Update(attribs)
-			if err != nil {
-				return err
-			}
+		_, err := m.userAttributes.Create(attribs)
+		if err != nil {
+			return err
 		}
-
 		return nil
 	}
 
-	if needCreate {
-		_, err := m.userAttributes.Create(attribs)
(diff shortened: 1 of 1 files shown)


luthermonson:
https://github.com/rancher/rancher/issues/25479
https://github.com/rancher/rancher/issues/25504

pkg/auth/tokens/manager.go
```diff
@@ -573,23 +573,21 @@ func (m *Manager) UserAttributeCreateOrUpdate(userID, provider string, groupPrin
 		return err
 	}
 
-	// Exists, just update if necessary
-	attribs = attribs.DeepCopy()
-
-	if m.UserAttributeChanged(attribs, provider, groupPrincipals) {
+	if needCreate {
+		attribs = attribs.DeepCopy()
```
prachidamle: We need not invoke the DeepCopy() here since when needCreate=true, a new instance is just created here https://github.com/rancher/rancher/blob/647b6fa7d15cf3420a2ee82677e20fd0e4069b3e/pkg/auth/tokens/manager.go#L550

luthermonson: yup, totally agree. thanks.

luthermonson: fixed

dramich:
If the return is moved up into the brackets won't that fix the issue? [linky](https://github.com/rancher/rancher/pull/25602/files#diff-9362863182562aef3b69af11364beb80L588)

prachidamle:
@dramich when needCreate=true this fix saves invoking this https://github.com/rancher/rancher/blob/1d21c437cf793c7a2b984a5323c40a59c95da70b/pkg/auth/tokens/manager.go#L601

dramich:
LGTM, saving a call always wins.

PR #25603, opened 2020-02-26
Update role calls to use listers

dramich:
Problem:
On project sync API calls are being used to validate something exists

Solution:
Update to use a cache to ease the load on the API

Issue: https://github.com/rancher/rancher/issues/25605

### DIFF
--- a/pkg/systemaccount/systemaccount.go
+++ b/pkg/systemaccount/systemaccount.go
@@ -25,6 +25,7 @@ func NewManager(management *config.ManagementContext) *Manager {
 		crtbs:       management.Management.ClusterRoleTemplateBindings(""),
 		crts:        management.Management.ClusterRegistrationTokens(""),
 		prtbs:       management.Management.ProjectRoleTemplateBindings(""),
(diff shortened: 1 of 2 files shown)


pkg/systemaccount/systemaccount.go
```diff
@@ -116,28 +119,32 @@ func (s *Manager) GetOrCreateSystemClusterToken(clusterName string) (string, err
...
 				Name:      bindingName,
 				Namespace: projectName,
 			},
 			ProjectName:      projectID,
-			UserName:         user.Name,
+			UserName:         u.Name,
 			RoleTemplateName: projectMemberRole,
-		})
+		}
+		if _, err := s.prtbs.Create(prtb); err != nil && !errors2.IsAlreadyExists(err) {
```
cbron: Why would we need to add this alreadyExists check if we just found that prtb doesn't exist, is it for HA or something ?

dramich: It does't exist in the cache, a cache miss will still cause an error here if it already exists in the DB.

dramich: In general create call errors should get protected with an IsAlreadyExists check.

PR #25629, opened 2020-02-27
[2.4] Fix chart when specify audit-log level

rawmind0:


deniseschannon:
Don't we need this fix for anything that uses an int in the chart? https://rancher.com/docs/rancher/v2.x/en/installation/options/chart-options/

auditlog.level
auditlog.maxAge
auditlog.maxbackups
auditlog.maxSize

rawmind0:
Not really.. Just `auditlog.level` has if with integer comparisons in the chart templates.

deniseschannon:
What branch is this? I thought we already added it to master with this PR: 

https://github.com/rancher/rancher/pull/25628

PR #25646, opened 2020-02-28
Nodedriver Schemas, Defaulting array[string] to Nullable

luthermonson:
array[string] in the dynamicschemas were a little annoying to edit in the api ui because they default to null when an empty array `[]` is sent over. On edit you would have to change each one to [] or "" to make it save properly.


luthermonson:
@sangeethah per a conversation with @vincent99 he's approving setting array[string] to nullable and would like to make sure testers are aware in case anything unusual happens.

PR #25653, opened 2020-02-28
Increase wait time for rancher server to become available

sangeethah:
Increase wait time for rancher server to become available from 2 minutes to 5 minutes


sowmyav27:
LGTM

Issue #25371, opened 2020-02-12
K3s integration to Rancher


maggieliu:
What kind of request is this:
Feature request/Backend development

Description:
Backend requirements for k3s integration to Rancher.

Additional Information:
rancher/rancher#25141

daxmc99:
results of k3s sync-up @cjellick, @dweomer , @galal-hussein @rmweir:

**How will k3s version be determined**:
- client-go K8sClient.Discovery().ServerVersion 
https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go#L120
- @daxmc99 to determine if timeout concerns on this method are valid 

**How will Plans be used to upgrade the cluster** 
- 2 plans will be used: Master/ Non-master
- "Ordering" will be preserved by using the *Prepare* spec within the worker plan
- - ie: Master plans proceeds and worker plan has a *Prepare* spec to wait until master plan completes 
- - Goal of _not_ having Rancher run the upgrade, controller + plans should be self-sufficient. 
- - @galal-hussein to work with @dweomer to add *Prepare* logic to k3s-upgrade image

**Users who have previously used plans + system-upgrade-controller**
-  Rancher will create plans that disable all other plans
-  for example `plane.upgrade.cattle.io/<plan name>: <digest>` exists
- - rancher plan will have 
```
spec { 
         nodeSelector {
                            plane.upgrade.cattle.io/disable : true
```
- - rancher applies label `plane.upgrade.cattle.io/disable : true` to all nodes it sees

- If a user removes cluster from rancher they simply need to (preferably)  delete the label  or 
use a plan that specifically selects the disable label

dweomer:
@daxmc99 @cjellick I thought that we decided that `upgrade.cattle.io/disable=true` on nodes was not necessary given plans `p0` and `p1` with `p0` being legacy aka pre-import and provided that we are able to leverage the assumption that `plan.upgrade.cattle.io/p1!=disabled` is the enabling label for `p1` which means that we put the equivalent of `!plan.upgrade.cattle.io/p1` aka `- {key: plan.upgrade.cattle.io/p1, op: DoesNotExist}` in the `spec.nodeSelector` of `p0`

cjellick:
@dweomer, no - I came away with (almost) the same takeaway as @daxmc99.
When Rancher creates a plan, it first adds `upgrade.cattle.io/disable: true` to all non-rancher manager plans. I like this approach because the documentation explaining what we did is very simple and straightforward, something like:
[...]

PR #25667, opened 2020-02-28
K3s Integration

daxmc99:
This adds the ability for Rancher to import k3s clusters and upgrade them

Addresses: https://github.com/rancher/rancher/issues/25371


pkg/controllers/management/k3supgrade/template.go
```diff
@@ -0,0 +1,163 @@
...
+metadata:
+  name: k3s-master-plan 
+  namespace: system-upgrade
+spec:
+  concurrency: 1
+  version: v1.17.2-k3s1
+  nodeSelector:
+    matchExpressions:
+      - {key: k3s-upgrade, operator: Exists}
+  serviceAccountName: default
```
dweomer:
```suggestion
  serviceAccountName: system-upgrade # or cattle-upgrade
```

daxmc99: https://github.com/rancher/rancher/pull/25667/files#diff-f70d9c6e95a58b98b200416cd10ef05eR54 
These plans are actually unused in their current form

pkg/api/customization/kontainerdriver/actionhandler.go
```diff
@@ -137,12 +137,12 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
 	var targetImages []string
 	switch apiContext.ID {
 	case linuxImages:
-		targetImages, err = image.GetImages(systemCatalogChartPath, []string{}, rkeSysImages, image.Linux)
+		targetImages, err = image.GetImages(systemCatalogChartPath, []string{}, []string{}, rkeSysImages, image.Linux)
```
rmweir: this may have been me, but I believe this was a mistake. Should be rkeSysimages still.

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,209 @@
...
+		return err
+	}
+	planClient := planConfig.Plans(metav1.NamespaceAll)
+
+	planList, err := planClient.List(metav1.ListOptions{})
+	if err != nil {
+		// may need to handle this error, if there is no Plan CRD what should we do?
+		if errors.IsNotFound(err) {
+			// no plan CRD exists
+			logrus.Warnf("plan CRD does not exist: %s", err)
```
rmweir: nit: maybe just make this the err that is returned in this case and say, the crd does not exist _yet_

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,209 @@
...
+				return err
+			}
+			if !cmp(masterPlan, newMaster) {
+				planClient = planConfig.Plans(systemUpgradeNS)
+				_, err = planClient.Update(&newMaster)
+				if err != nil {
+					return err
+				}
+			} else {
+				fmt.Println("master plan is the same, not updating")
```
rmweir: should this be an info log?

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,209 @@
...
+				return err
+			}
+			if !cmp(workerPlan, newWorker) {
+				planClient = planConfig.Plans(systemUpgradeNS)
+				_, err = planClient.Update(&newWorker)
+				if err != nil {
+					return nil
+				}
+			} else {
+				fmt.Println("worker plan is the same, not updating")
```
rmweir: same as L141

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+		}
+
+	} else { // create the plans
+		planClient = planConfig.Plans(systemUpgradeNS)
+		masterPlan, err = generateMasterPlan(cluster.Spec.K3sConfig.Version,
```
rmweir: do we not want to error check this?

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,209 @@
...
+			cluster.Spec.K3sConfig.ServerConcurrency)
+		_, err = planClient.Create(&masterPlan)
+		if err != nil {
+			return err
+		}
+		workerPlan, err = generateWorkerPlan(cluster.Spec.K3sConfig.Version,
```
rmweir: same as L163

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,209 @@
...
+			cluster.Spec.K3sConfig.WorkerConcurrency)
+		_, err = planClient.Create(&workerPlan)
+		if err != nil {
+			return nil
+		}
+		fmt.Println("Deployed plans into cluster")
```
rmweir: info log?

pkg/controllers/management/k3supgrade/template.go
```diff
@@ -0,0 +1,163 @@
...
+     image: rancher/k3s-upgrade:latest
+     args: ["prepare","k3s-master-plan"]
+  nodeSelector:
+    matchExpressions:
+    - {key: k3s-worker-upgrade, operator: Exists}
+  serviceAccountName: system-upgrade
+  drain:
+    force: true
+  upgrade:
+    image: rancher/k3s-upgrade`
```
rmweir: If we don't use these plan constants can we just delete them?

pkg/controllers/management/k3supgrade/template.go
```diff
@@ -0,0 +1,163 @@
...
+		Upgrade: &planv1.ContainerSpec{
+			Image: upgradeImage,
+		},
+	},
+	Status: planv1.PlanStatus{},
+}
+
+func generateMasterPlan(version string, concurrency int) (planv1.Plan, error) {
+	masterPlan := genericPlan
+	masterPlan.Name = k3sMasterPlanName
```
rmweir: do we need use selectors if we already know the name? and will there be an issue ever with us hardcoding the name? There should never be two?

pkg/controllers/management/k3supgrade/template.go
```diff
@@ -0,0 +1,163 @@
...
+			Values:   []string{"true"},
+		}},
+	}
+
+	return masterPlan, nil
+}
+
+func generateWorkerPlan(version string, concurrency int) (planv1.Plan, error) {
+	workerPlan := genericPlan
+	workerPlan.Name = k3sWorkerPlanName
```
rmweir: same as L86

pkg/image/export/main.go
```diff
@@ -77,12 +77,14 @@ func run(systemChartPath string, imagesFromArgs []string) error {
...
-	targetImages, err := img.GetImages(systemChartPath, imagesFromArgs, linuxInfo.RKESystemImages, img.Linux)
+	k3sUpgradeImages := getK3sUpgradeImages(data.K3S)
+
+	targetImages, err := img.GetImages(systemChartPath, k3sUpgradeImages, imagesFromArgs, linuxInfo.RKESystemImages, img.Linux)
 	if err != nil {
 		return err
 	}
 
-	targetWindowsImages, err := img.GetImages(systemChartPath, []string{getWindowsAgentImage()}, windowsInfo.RKESystemImages, img.Windows)
+	targetWindowsImages, err := img.GetImages(systemChartPath, k3sUpgradeImages, []string{getWindowsAgentImage()}, windowsInfo.RKESystemImages, img.Windows)
```
rmweir: k3sUpgradeImages should probably be empty. Even though a windows cluster will be a mixture of windows and linux nodes, the windows-images textfile is just to explain the images needed for the windows nodes I believe, as the linux ones will be downloaded anyway. See: https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/air-gap/populate-private-registry/#linux-steps
Could have been me but I thought I changed this.

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -966,10 +966,10 @@ func k3sClusterConfig(cluster *v3.Cluster) error {
 			Err:    fmt.Errorf("waiting for full cluster configuration"),
 			Reason: "Pending"}
 	}
-	// TODO Rancher may support upgrading the local cluster if its k3s in the future
-	if cluster.Name == v3.ClusterDriverLocal || cluster.Status.Driver == v3.ClusterDriverK3s {
-		return nil
+	if cluster.Status.Driver == v3.ClusterDriverK3s {
```
galal-hussein: I am not sure I understand why did you remove the exception for local clusters here

pkg/controllers/management/k3supgrade/template.go
daxmc99: Selectors are required to activate a plan (cc @dweomer) and we do not want to label every node to activate a plan. There should only ever be two _rancher-managed_ plans. One master and one worker

pkg/controllers/management/k3supgrade/register.go
daxmc99: I don't think it matters what the error here is actually. We should just requeue on any error.

daxmc99: In other controllers we may not return the 'NotFound' error to the framework but here we know that the system-upgrade-controller should be installed

pkg/controllers/management/k3supgrade/register.go
daxmc99: Not sure if this is worth a log

pkg/controllers/management/k3supgrade/register.go
daxmc99: Same here, going to drop the log

pkg/api/customization/kontainerdriver/actionhandler.go
daxmc99: This is still `rkeSysImages`. Current func signature is `GetImages(systemChartPath string, k3sUpgradeImages, imagesFromArgs []string, rkeSystemImages map[string]v3.RKESystemImages, osType OSType) ([]string, error)` 
arg[3] is and should be rkeSysImage ✨

pkg/image/export/main.go
```diff
@@ -218,6 +220,20 @@ func getWindowsAgentImage() string {
 	return fmt.Sprintf("%s/rancher-agent:%s", repo, tag)
 }
 
+func getK3sUpgradeImages(k3sData map[string]interface{}) []string {
```
cjellick: This function is fairly funky in what it does. I think it needs a comment

pkg/controllers/management/clusterprovisioner/provisioner.go
daxmc99: @cjellick Asked that we be able to recognize local clusters if they are k3s. Do you see any issues here?

pkg/controllers/management/k3supgrade/template.go
daxmc99: done

pkg/controllers/management/k3supgrade/register.go
daxmc99: dropped error creation in `generateMasterPlan`

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
```diff
@@ -0,0 +1,174 @@
...
+
+// deployK3sUpgradeController creates a rancher k3s upgrader controller if one does not exist.
+// Updates k3s upgrader controller if one exists and is not the newest available version.
+func (h *handler) deployK3sUpgradeController(clusterName string) error {
+	userCtx, err := h.manager.UserContext(clusterName)
+	if err != nil {
+		return err
+	}
+
+	projectLister := userCtx.Management.Management.Projects("").Controller().Lister()
```
dramich: This can just live on the handler so it doesn't need to be created here

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
```diff
@@ -0,0 +1,174 @@
...
+	systemProject, err := project.GetSystemProject(clusterName, projectLister)
+	if err != nil {
+		return err
+	}
+
+	templateID := k3sUpgraderCatalogName
```
dramich: Why the new var?

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,190 @@
...
+	templateLister         v3.CatalogTemplateLister
+	systemAccountManager   *systemaccount.Manager
+	manager                *clustermanager.Manager
+}
+
+const (
+	systemUpgradeNS        = "cattle-system"
+	rancherManagedPlan     = "rancher-managed"
+	upgradeDisableLabelKey = "plan.upgrade.cattle.io/disable"
+	k3sUpgraderCatalogName = "system-library-rancher-k3s-upgrader"
```
dramich: This isn't used in this file, can it be moved?

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return err
+	}
+
+	appLister := userCtx.Management.Project.Apps("").Controller().Lister()
+	appClient := userCtx.Management.Project.Apps("")
+
+	latestVersionID := latestTemplateVersion.ExternalID
+
+	app, err := appLister.Get(systemProjectName, "rancher-k3s-upgrader")
```
dramich: "rancher-k3s-upgrader" can move into a const

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
```diff
@@ -0,0 +1,174 @@
...
+	_, systemProjectName := ref.Parse(systemProjectID)
+
+	nsClient := userCtx.Core.Namespaces("")
+	appProjectName, err := utils2.EnsureAppProjectName(nsClient, systemProjectName, clusterName, systemUpgradeNS, creator.Name)
+	if err != nil {
+		return err
+	}
+
+	appLister := userCtx.Management.Project.Apps("").Controller().Lister()
+	appClient := userCtx.Management.Project.Apps("")
```
dramich: Why don't these live on handler?

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
```diff
@@ -0,0 +1,174 @@
...
+			Spec: v32.AppSpec{
+				Description:     "Upgrade controller for k3s clusters",
+				ExternalID:      latestVersionID,
+				ProjectName:     appProjectName,
+				TargetNamespace: systemUpgradeNS,
+			},
+		}
+
+		// k3s upgrader doesn't exist yet, so it will need to be created
+		if _, err = appClient.Create(desiredApp); err != nil {
```
dramich: Check the error isn't an alreadyExists error to save from races

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
```diff
@@ -0,0 +1,174 @@
...
+			return err
+		}
+	} else {
```
dramich: Else block isn't needed

dramich: Eh guess it is if you don't return after the create...

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// create a client for GETing Plans in the downstream cluster
+	planConfig, err := planClientset.NewForConfig(&clusterCtx.RESTConfig)
+	if err != nil {
+		return err
+	}
+	planClient := planConfig.Plans(metav1.NamespaceAll)
+
+	planList, err := planClient.List(metav1.ListOptions{})
```
dramich: Ideally this client can be moved onto the user context so we don't need to create it here. In the future other things will probably need it.

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
rmweir: because the handler on the management cluster but the app is being deployed in the downstream cluster, which will be cluster specific.

dramich: That doesn't make sense. App is a rancher CRD and this pathing is calling management, which gets back to the management context. This is not doing anything with the downstream cluster.

pkg/controllers/management/k3supgrade/k3supgrade_handler.go
rmweir: if it receives an already exists error then the handler should work on the next attempt

pkg/api/customization/cluster/validator.go
```diff
@@ -118,6 +125,58 @@ func (v *Validator) validateEnforcement(request *types.APIContext, data map[stri
...
+	if request.Method == http.MethodPost {
+		return nil
+	}
+
+	if spec.K3sConfig == nil {
+		// only applies to k3s clusters
+		return nil
+	}
+
+	// must wait for original spec version to be set
```
cjellick: this comment is weird. you arent waiting on anything; you're just ensuring the user didnt send a ""

cjellick: and more important, returning `upgradeNotReadyErr` is wrong in this case

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -0,0 +1,190 @@
...
+		systemAccountManager:   systemaccount.NewManager(mgmtCtx),
+		manager:                manager,
+	}
+
+	wContext.Mgmt.Cluster().OnChange(ctx, "k3s-upgrade-controller", h.onClusterChange)
+}
+
+// deployPlans creates a master and worker plan in the downstream cluster to instrument
+// the system-upgrade-controller in the downstream cluster
+func (h *handler) deployPlans(cluster *v3.Cluster) error {
```
cjellick: weird that this function is in register.go

pkg/controllers/management/k3supgrade/template.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+		Upgrade: &planv1.ContainerSpec{
+			Image: upgradeImage,
+		},
+	},
+	Status: planv1.PlanStatus{},
+}
+
+func generateMasterPlan(version string, concurrency int) planv1.Plan {
+	masterPlan := genericPlan
```
cjellick: does this copy the plan?

pkg/controllers/management/k3supgrade/template.go
```diff
@@ -0,0 +1,121 @@
...
+func generateWorkerPlan(version string, concurrency int) planv1.Plan {
+	workerPlan := genericPlan
+	workerPlan.Name = k3sWorkerPlanName
+	workerPlan.Spec.Version = version
+	workerPlan.Spec.Concurrency = int64(concurrency)
+
+	// worker plans wait for master plans to complete
+	workerPlan.Spec.Prepare = &planv1.ContainerSpec{
+		// TODO: the image version doesn't matter here, needs something recent
+		Image: upgradeImage + ":" + "v1.17.3-k3s1",
```
cjellick: is this going to go to production like this or do we have a beter solution?

cjellick:
failures are unrelated. im merging

pkg/controllers/management/k3supgrade/template.go
daxmc99: Ideally this would be fixed in system-upgrade-controller here https://github.com/rancher/system-upgrade-controller/issues/41

pkg/controllers/management/k3supgrade/register.go
daxmc99: Should be moved

Issue #25660, opened 2020-02-28
Helm-Unittest failures due to api limit


brendarearden:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Run drone build
-OR-
curl https://api.github.com/repos/$PROJECT_GH/releases/latest a few times

**Result:**
See dapper step `rancher-helm plugin install https://github.com/rancher/helm-unittest` fail occasionally
-OR-
`{
  "message": "API rate limit exceeded for <ip>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}`

**Example Drone Failure** 
https://drone-pr.rancher.io/rancher/rancher/7332/2/2

jiaqiluo:
wait for https://github.com/rancher/rancher/pull/25740 to be merged.

brendarearden:
fix is working in master and 2.3 branches as both builds passed successfully and continue to pass.

PR #25673, opened 2020-02-28
Add helm unitttest version env var

brendarearden:
**Problem** 
Hitting rate limit for github api calls when helm-unittest plugin was being install

**Solution** 
Rewrite plugin install script to pull the binary down directly based on a new `HELM_UNITTEST_VERSION` env var and existing `ARCH` env var from dapper file

**Issue** 
#25660 

**Dependent on**
https://github.com/rancher/helm-unittest/pull/7


brendarearden:
Drone build pointed at helm-unittest changes: https://drone-pr.rancher.io/rancher/rancher/7393

Issue #25506, opened 2020-02-21
Cluster agent and Node agent pods stuck in removing state on a rancher server upgrade and fails to get upgraded


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster - 3 nodes - 1 etcd, 1 control plane and 1 worker node in 2.3.5
- Deploy a workload on the cluster
- Upgrade rancher to latest master-head - commit id: `1ce03e735`
- The Cluster agent and Node agent pods stuck in removing state on the control plane node.
- cattle node agent is seen upgraded on the worker node

**Expected Result:**
Cluster agent and Node agent pods should get upgraded successfully

**Note:**

- CLuster with all roles - 1 node. Upgrade was successful from 2.3.5 to master head.
- Cluster agent and Node agent pods are upgraded successfully
- System Project --> Pods are all active and up.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `1ce03e735`
- Installation option (single install/HA): single node

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke do
- Kubernetes version (use `kubectl version`):

```
v1.15.9
```

soumyalj:
Reproduced the issue above:
1. Create a 3 node cluster with K8s version : v1.15.9
N1 : control 
N2: etcd
N3 : worker
2. Deploy few workloads
3. Upgrade Rancher to master-head(Rancher version dad58367e (dad58367e))
After upgrading rancher, the node-agent on worker node was stuck in "Removing" state and failed to get upgraded.
node-agent on contolplane node got upgraded. 
<img width="911" alt="Screen Shot 2020-02-21 at 11 23 48 AM" src="https://user-images.githubusercontent.com/18536626/75065276-5fad1a00-549d-11ea-8505-cc74206c271a.png">

The cluster agent got upgraded successfully.
<img width="923" alt="Screen Shot 2020-02-21 at 11 30 50 AM" src="https://user-images.githubusercontent.com/18536626/75065394-a3a01f00-549d-11ea-943f-154643dc62f5.png">

sowmyav27:
Even "kube-api-auth" workload pods are stuck in `Removing` state. Pods on the control plane nodes are stuck in removing state
[...]

PR #25690, opened 2020-03-02
Restart kubelet after sharemnt is run

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/25506

superseb:
Another option would be if we could detect it is a running cluster which didn't need it anyway (as it wouldnt be able to come up properly if it would fail), and just don't run it.

StrongMonkey:
@superseb With the change introduced ion share-mnt I think it is safe to restart kubelet once share-mnt is run. I still need to understand why kubelet restart fix it, but force deploy doesn't look like a better option. Restarting kubelet shouldn't impact existing workloads, But I wonder if share-mnt is still needed to be run on every host.

StrongMonkey:
After talking to @ibuildthecloud, this approach seems better https://github.com/rancher/share-mnt/pull/14. First we check if root directory is mounted as shared, and if it is, skip stage2.(stage2 is to mark directory as shared).

StrongMonkey:
@superseb ^^

Issue #25144, opened 2020-01-29
Adding controlplane or etcd node with --label resets labels on other nodes


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Provision a three node custer cluster (k8s v1.15.9), 1 worker, 1 controlplane, and 1 etcd node. For each node use `--label color=red` in the docker run command to label each node with the color red.
2. In the UI, edit each node of the three nodes and change the color label to blue.
3. Add a fourth controlplane node to the cluster and use `--label color=red`.

**Result:**
After adding the fourth controlplane node, the original controlplane node and etcd node labels are changed back to their original labels of color=red. This is not expected behavior, the labels should not have changed.

**Other details that may be helpful:**
The label on the worker node does not change. Adding more workers does not reset labels. Adding more etcd or controlplane nodes does reset labels on both etcd and controlplane nodes. Deletion of an etcd or controlplane node also resets labels of the remaining etcd and controlplane nodes.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.3.5

- Installation option (single install/HA):
single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
t3a.medium in aws
- Kubernetes version (use `kubectl version`):

```
> kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.9", GitCommit:"2e808b7cb054ee242b68e62455323aa783991f03", GitTreeState:"clean", BuildDate:"2020-01-18T23:24:23Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):

```
$ docker version
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea838
 Built:             Wed Nov 13 07:22:05 2019
[...]

PR #25719, opened 2020-03-02
Fix labels add/remove when labels overlap with plan

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/25144

PR #25725, opened 2020-03-03
WIP: Add test for deleting a cluster

sangeethah:


tests/validation/tests/v3_api/common.py
```diff
@@ -1131,7 +1131,7 @@ def delete_cluster(client, cluster):
             nodes = client.list_node(clusterId=cluster.id).data
             filters = [
                 {'Name': 'tag:Name',
-                 'Values': ['testcustom*', 'teststess*']}]
+                 'Values': ['testcustom*', 'teststress*', 'testsa*']}]
```
bmdepesa: Should we have an option to pass this list of prefixes as a param?

PR #25726, opened 2020-03-03
Update types to include pipeline image revert

StrongMonkey:


StrongMonkey:
@deniseschannon This is not needed anymore

PR #25733, opened 2020-03-03
WIP: Token hashing test

cbron:


pkg/systemtokens/systemtokens.go
```diff
@@ -143,8 +146,19 @@ func (t *SystemTokens) createOrUpdateSystemToken(tokenName, description, kind, u
 		logrus.Infof("Creating system token for %v, token: %v", userName, tokenName)
 		token, err = t.tokenClient.Create(token)
 		if err != nil {
+			// race condition on create, someone else just created so return that val instead
+			if apierrors.IsAlreadyExists(err) {
+				if tokenVal := t.getTokenValue(tokenName); tokenVal != "" {
+					return tokenVal, nil
+				}
```
dramich: If we miss the cache then it should return an error of your choosing, returning the alreadyexists error below is not accurate as to what happened

Issue #25735, opened 2020-03-03
kontainer-driver-metadata - auto-refreshing does not use new url after editing


jiaqiluo:
Bug 

Rancher:master-head 2e6041d

after changing the URL from `url-1` to `url-2`, there is a refreshing trigged as expected, but Rancher log shows it still uses the `url-1`, and this can also be confirmed from the UI/API that new values are not applied.

PR #25738, opened 2020-03-03
Use object from sync method as setting

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/25735.

Issue #25744, opened 2020-03-03
Failed while: Wait for Condition: InitialRolesPopulated: True when creating a cluster with a Standard User


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

- Install Rancher `2.4 master-head (03/03/2020)` commit id: 3623201ad
- Login as admin, create an Standard User (`standarduser1`)
- Logout and login as `standarduser1`
- Start creating a cluster, I used a one/all roles node on DigitalOcean
- Click Create

**Result:**

`Failed while: Wait for Condition: InitialRolesPopulated: True`

<img width="1119" alt="Screen Shot 2020-03-03 at 2 00 18 PM" src="https://user-images.githubusercontent.com/27659/75820355-6e2ae980-5d59-11ea-8c87-c718f4b4f1a8.png">

**Other details that may be helpful:**

- This is not reproducible in Rancher v2.3.5
- Clicking in Create again after getting the error succeeds in creating the cluster.

**Environment information**
- Rancher version: `2.4 master-head (03/03/2020)` commit id: 3623201ad
- Installation option (single install/HA): single

izaac:
Rancher version: `2.4 master-head (03/05/2020)` commit id: f6872eb2c

Validated this is no longer happening when deploying a cluster with an Standard User.

JannikZed:
I'm facing the same issue in Rancher 2.5.3 - saving it again does not succeed.. I'm an admin user.

JannikZed:
..and for Rancher 2.6.3 as well. I'm trying to create a new cluster with the custom "hetzner driver"

amagura:
...and for Rancher 2.5.8 as well.

vullankib:
Do we know why this is happening and any resolution to this? Not sure why this is Marked as closed.

ArthurMcTool:
Happening for me in 2.5.7. Would be nice to add a new cluster...

ArthurMcTool:
I was getting this error in cattle-system pods:
[...]

PR #25782, opened 2020-03-05
Don't cache the AccessSet for subscribe

ibuildthecloud:


rmweir:
https://github.com/rancher/rancher/issues/25744

Issue #25214, opened 2020-02-03
CIS Scan - State should reflect Running/Pass/Fail info


codyrancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Enhancement

**Result:**
CIS Scan state should reflect the results of the tests that were run.
We don't want to see `Active` etc. We want to see `Running/Pass/Fail`.

**Notes:**
Referring to the state displayed here.
![Screen Shot 2020-02-03 at 10 37 07 AM](https://user-images.githubusercontent.com/55104481/73676205-3739b900-4671-11ea-98b6-abafcb077e5d.png)

This was originally reported at rancher/rancher#25159 but this is a backend issue so I'm splitting it into a separate issue.

sowmyav27:
**Verified on latest master-head - commit id: `bfa3e1c8a` ui tag: latest2.**

Fail is seen in Red and Pass in Green on UI.
<img width="1333" alt="Screen Shot 2020-03-08 at 9 42 34 AM" src="https://user-images.githubusercontent.com/26032343/76167078-3ccb6880-6121-11ea-9fe9-7f983c8bc5ac.png">

State Transitions currently happening -- Creating --> Running --> Active --> Fail/Pass/Error

For Active state showing, logged bug -- https://github.com/rancher/rancher/issues/25856

PR #25796, opened 2020-03-05
Fix CIS state and review comments

leodotcloud:
Addresses: https://github.com/rancher/rancher/issues/25214
Addresses comments from: https://github.com/rancher/rancher/pull/25475

### DIFF
--- a/pkg/api/customization/clusterscan/clusterscan.go
+++ b/pkg/api/customization/clusterscan/clusterscan.go
@@ -56,6 +56,7 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 			if err != nil {
 				logrus.Errorf("error converting failInterface to int: %v", err)
 				resource.Values["state"] = errorState
+				return
 			}
 			if fail > 0 {
 				resource.Values["state"] = failedState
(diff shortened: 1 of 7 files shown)


pkg/api/customization/clusterscan/clusterscan.go
```diff
@@ -33,11 +40,28 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
...
+			if !ok {
+				resource.Values["state"] = errorState
+				return
+			}
+			cisScanStatus := convert.ToMapInterface(cisScanStatusInterface)
+			failInterface := cisScanStatus["fail"]
+			fail, err := convert.ToNumber(failInterface)
+			if err != nil {
+				logrus.Errorf("error converting failInterface to int: %v", err)
+				resource.Values["state"] = errorState
```
luthermonson: should you return after this? if err is != nil likely fail == 0 meaning the the else would mark it passed right?

leodotcloud: yup, let me fix it.

pkg/controllers/user/cis/utils.go
```diff
@@ -32,22 +32,14 @@ func createConfigMapWithRetry(configMapsClient rcorev1.ConfigMapInterface, cm *v
...
-		)
-		if err != nil {
-			return fmt.Errorf("error listing pods: %v", err)
-		}
-		if len(pods) == 0 {
-			removed = true
-			break
-		}
-		time.Sleep(RetryIntervalInMilliseconds * time.Millisecond)
+	pods, err := podLister.List(
```
dramich: We should prefer to use indexes vs a lister with a selector. The lister will loop over all objects to build your return and on something that will most likely have a lot of things it's resource intensive.

leodotcloud: This is an ephemeral pod so not sure if having an index would be of any benefit. Also, there is only one pod running in a cluster with this particular label set.

PR #25808, opened 2020-03-05
Do not add labels into plan if plan doesn't have the key

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/25144#issuecomment-595386365

StrongMonkey:
Tests are failing because of hash function.

Issue #25786, opened 2020-03-05
"fatal error: runtime: out of memory" panic seen in rancher master


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
This happened on an upgraded set up from 2.3.5 to master-head latest - commit `53e0430e1`.

**What was done in this setup:** 
- In 2.3.5 have a cluster and run pre upgrade automation tests on the cluster (as a std user, enable Auth)
- Create few API tokens from the UI.
- Verified that the tokens are in plain text on the management plane.
- Remote kubectl with kubeconfig (from the rancher UI) works. Save this config to be used after upgrade.
- Upgrade rancher to master-head.
- Existing tokens have the token hashed, they are not seen in plain text on the management plane anymore.
- Post upgrade automation tests work fine.
- Edit cluster, change the k8s version.
- The cluster goes into updating state. And comes up "Active"

**Actual Result:**
Rancher server has restarted, and is up and running
Rancher logs:
```
[storage] 2020/03/05 01:00:04 updating release "mysql.v1"
2020/03/05 01:00:04 [INFO] Creating roleBinding User user-2mpdb Role admin
2020/03/05 01:00:05 [INFO] Creating roleBinding User u-irdieh3onn Role edit
2020/03/05 01:00:05 [INFO] Creating roleBinding User u-irdieh3onn Role project-member
NAME:   mysql
[storage] 2020/03/05 01:00:05 getting last revision of "mysql"
[storage] 2020/03/05 01:00:05 getting release history for "mysql"
I0305 01:00:08.093288      26 trace.go:116] Trace[510403320]: "GuaranteedUpdate etcd3" type:*core.Endpoints (started: 2020-03-05 01:00:06.520831741 +0000 UTC m=+2550.150603306) (total time: 1.430985888s):
Trace[510403320]: [467.046237ms] [436.694507ms] Transaction prepared
Trace[510403320]: [1.415706004s] [948.659767ms] Transaction committed
I0305 01:00:08.881819      26 trace.go:116] Trace[1332190086]: "Update" url:/api/v1/namespaces/kube-system/endpoints/kube-scheduler,user-agent:k3s/v1.17.2+k3s1 (linux/amd64) kubernetes/cdab19b/leader-election,client:127.0.0.1 (started: 2020-03-05 01:00:06.389796989 +0000 UTC m=+2550.019568544) (total time: 2.151701335s):
[...]

PR #25810, opened 2020-03-05
Token hashing uses sha256

cbron:
Issue: https://github.com/rancher/rancher/issues/25786

Notes: 
* Can't use Sscanf for parsing format because go stumbles on multiple strings not delimited by spaces, so had to go the split route.
* hashedSecretKey is actually a `[]uint8` but thats basically a byte so comparison is ok. The `[:]` converts array to slice.

### DIFF
--- a/pkg/auth/tokens/sha256.go
+++ b/pkg/auth/tokens/sha256.go
@@ -1,0 +1,67 @@
+package tokens
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+	"crypto/subtle"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+const hashFormat = "$%d:%s:%s" // $version:salt:hash -> $1:abc:def
+const Version = 1
+
+// CreateSHA256Hash can be used for basic key hashing, includes a random salt
+func CreateSHA256Hash(secretKey string) (string, error) {
+	salt := make([]byte, 8)
+	_, err := rand.Read(salt)
+	if err != nil {
+		return "", err
+	}
+	hash := sha256.Sum256([]byte(fmt.Sprintf("%s%s", salt, secretKey)))
+	encSalt := base64.RawStdEncoding.EncodeToString(salt)
+	encKey := base64.RawStdEncoding.EncodeToString(hash[:])
+	return fmt.Sprintf(hashFormat, Version, encSalt, encKey), nil
+}
+
+// VerifySHA256Hash takes a key and compares it with stored hash, including its salt
+func VerifySHA256Hash(hash, secretKey string) error {
+	if !strings.HasPrefix(hash, "$") {
+		return errors.New("hash format invalid")
+	}
+	splitHash := strings.SplitN(strings.TrimPrefix(hash, "$"), ":", 3)
+	if len(splitHash) != 3 {
+		return errors.New("hash format invalid")
+	}
+
+	version, err := strconv.Atoi(splitHash[0])
+	if err != nil {
+		return err
+	}
+	if version != Version {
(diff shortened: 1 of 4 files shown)


pkg/auth/tokens/token_util.go
```diff
@@ -106,7 +109,7 @@ func VerifyToken(storedToken *v3.Token, tokenName, tokenKey string) (int, error)
 	if storedToken.ObjectMeta.Name != tokenName {
 		return 422, errors.New("Invalid auth token value")
 	}
-	if err := common.VerifyHash(storedToken.Token, tokenKey); err != nil {
+	if err := VerifyHash(storedToken.Token, tokenKey); err != nil {
```
daxmc99: I would prefer a rename here, find by symbol is impacted since now we have 2 `VerifyHash` functions in the codebase. We should also clarify when to use each. 
```suggestion
	if err := VerifySHA256Hash(storedToken.Token, tokenKey); err != nil {
```

pkg/auth/tokens/token_util.go
```diff
@@ -119,17 +122,55 @@ func VerifyToken(storedToken *v3.Token, tokenName, tokenKey string) (int, error)
...
+		token.Token = CreateHash(token.Token)
 		if token.Annotations == nil {
 			token.Annotations = map[string]string{}
 		}
 		token.Annotations[TokenHashed] = "true"
 	}
 	return nil
 }
+
+const hashFormat = "$%d:%s:%s" // $version:algorithm:hash -> $1:sha256:abc
```
daxmc99: This looks like the code in `common/crypto`. I think this all should be contained in its own file. It will likely be used in the future.

pkg/auth/tokens/token_util.go
```diff
@@ -119,17 +122,55 @@ func VerifyToken(storedToken *v3.Token, tokenName, tokenKey string) (int, error)
...
+const Version = 1
+const Algorithm = "sha256"
+
+func CreateHash(secretKey string) string {
```
daxmc99: I think we should document when to use this instead of the other VerifyHash. Which tokens require this vs the other

pkg/auth/tokens/token_util.go
```diff
@@ -119,17 +122,55 @@ func VerifyToken(storedToken *v3.Token, tokenName, tokenKey string) (int, error)
...
+	hash := sha256.Sum256([]byte(secretKey))
+	enc := base64.RawStdEncoding.EncodeToString(hash[:])
+	return fmt.Sprintf(hashFormat, Version, Algorithm, enc)
+}
+
+func VerifyHash(hash, secretKey string) error {
```
daxmc99: Same as above, I think this should have a commet about when to use it

pkg/auth/tokens/sha256.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return err
+	}
+	decodedSalt, err := base64.RawStdEncoding.DecodeString(salt)
+	if err != nil {
+		return err
+	}
+	// compare the two
+	hashedSecretKey := sha256.Sum256([]byte(string(decodedSalt) + secretKey))
+	if !bytes.Equal(decodedKey, hashedSecretKey[:]) {
```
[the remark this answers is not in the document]
vincent99: This "should never happen" but Darren requested a sanity check here that both values are non-zero length, so that there's no way a mistake leads to auth being `if ( "" == "")` and always succeeding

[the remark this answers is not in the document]
vincent99: https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare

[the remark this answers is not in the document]
cbron: updated both

Issue #23254, opened 2019-10-07
Runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized


mattmattox:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Deploy Rancher v2.2.8
- Create custom cluster with default settings
- Add node with missing firewall rules so node will only been partially added
- remove node from Rancher UI and try to rejoin

**Result:**
Runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialize

**Other details that may be helpful:**
This looks very similar to https://github.com/rancher/rancher/issues/13484

**Workaround**
Copy /etc/cni/net.d/10-canal.conflist and calico-kubeconfig from working node to new node. Then join node to cluster.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): 2/4
- Kubernetes version (use `kubectl version`):

```
v1.14.6-rancher1
```

- Docker version (use `docker version`):

```
17.03.2-ce
```

alena1108:
For RKE clusters provisioned with Rancher, there is a special handling for worker nodes. Node update/delete is handled by rancher agent, and is not included in the main rke reconcile loop. For the ref, that's where this exclusion happens https://github.com/rancher/rancher/blob/master/pkg/controllers/management/clusterprovisioner/driver.go#L165.

On the node creation, rancher agent gets a node plan from rke/rancher, places the configuration on the node and start all the k8s components https://github.com/rancher/rancher/blob/master/pkg/controllers/user/noderemove/nodes.go#L30
On node removal, agent simply removes the v1.node, and then rancher kills agent container https://github.com/rancher/rancher/blob/master/pkg/controllers/management/node/controller.go#L255.

We have to add a cleanup part to the node removal logic on Rancher/agent side, and remove all the necessary configs before proceeding with the agent removal.

[...]

PR #25812, opened 2020-03-05
Fix kubelet needs to be restart if node is deleted in UI

StrongMonkey:
Problem:
Today if node is deleted from rancher, we immdiately delete v1.node from
kubernetes. Since we don't clean up agent(or did a poor job in cleaning
agent). All the system containers keep running including kubelet. This
will cause kubelet keep throwing node not found error. This is because
kubelet only register new node at start up, and without restart kubelet
it will not re-register node again. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L52

Solution:
In our agent code, add kubelet restart logic. Notice that this logic is
only added into unmanaged agent(which is only added through docker run
command). This is because once managed agents are added, kubelet should
be ready since managed agents are pods from daemonset which mean the
node is schedulable now. Once managed agents are added unmanaged agents
will be cleaned up, then this logic will not be needed anymore.

### DIFF
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -39,6 +39,10 @@ const (
 )
 
 func main() {
+	if _, err := reconcileKubelet(context.Background()); err != nil {
+		logrus.Warnf("failed to reconcile kubelet, error: %v", err)
+	}
+
 	logrus.SetOutput(colorable.NewColorableStdout())
 	logserver.StartServerWithDefaults()
 	if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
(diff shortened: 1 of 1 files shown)


StrongMonkey:
Potential address https://github.com/rancher/rancher/issues/23254. Tested the issue with this change and could not reproduce it anymore.

pkg/agent/main.go
```diff
@@ -348,3 +366,38 @@ func certinfo(cert *x509.Certificate) {
...
 	logrus.Infof("PublicKeyAlgorithm: %+v", cert.PublicKeyAlgorithm)
 }
+
+func reconcileKubelet(ctx context.Context) (bool, error) {
+	if os.Getenv("CATTLE_K8S_MANAGED") == "true" {
+		return true, nil
+	}
+
+	nodeName := os.Getenv("CATTLE_NODE_NAME")
+	logrus.Infof("node %v is not registered, try restarting kubelet now", nodeName)
```
cjellick: "try restarting" is a little odd. can you drop try and say "restarting kubelet now"

pkg/agent/main.go
```diff
@@ -299,6 +299,24 @@ func run() error {
...
 						logrus.Errorf("failed to check plan: %v", err)
 					}
+
+				case <-ctx.Done():
+					return
+				}
+			}
+		}()
+
+		go func() {
```
cjellick: can you add a comment here explaining why we need this

cjellick:
per slack convo, add a call to reconcileKubelet to the beginning of the agent startup. this should allow the node to recover more quickly as opposed to having to wait a minute for the first reconcile to trigger

pkg/agent/main.go
```diff
@@ -299,6 +299,24 @@ func run() error {
...
+			for {
+				select {
+				case <-time.After(time.Minute):
```
cjellick: since you are adding the reconcile to the beginning, i think we should increase the retrigger to three minutes

PR #25813, opened 2020-03-05
Add Helm 3 testing

brendarearden:
Adding testing for #25300

### DIFF
--- a/tests/integration/suite/test_app.py
+++ b/tests/integration/suite/test_app.py
@@ -3,7 +3,7 @@ import pytest
 from rancher import ApiError
 from .test_catalog import wait_for_template_to_be_created
 from .common import random_str
-from .conftest import set_server_version, wait_for
+from .conftest import set_server_version, wait_for, DEFAULT_CATALOG
 
 
 def test_app_mysql(admin_pc, admin_mc):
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_app.py
```diff
@@ -591,6 +591,63 @@ def _app_fail():
...
+def test_app_has_helmversion(admin_pc, admin_mc, remove_resource):
+    """Test that app is using specified helm version"""
+    app_client = admin_pc.client
+    catalog_client = admin_mc.client
+    catalog_name1 = random_str()
+    catalog_name2 = random_str()
+    app_name1 = random_str()
+    app_name2 = random_str()
+    url1 = "https://github.com/rancher/integration-test-charts"
+    url2 = "https://github.com/rancher/integration-test-charts"
```
superseb: Do we need two vars for the same value?

superseb:
Besides ^^, LGTM

tests/integration/suite/test_app.py
brendarearden: we do not :)

superseb:
Needs a rebase

tests/integration/suite/test_app.py
```diff
@@ -591,6 +591,62 @@ def _app_fail():
...
 
+def test_app_has_helmversion(admin_pc, admin_mc, remove_resource):
+    """Test that app is using specified helm version"""
+    app_client = admin_pc.client
+    catalog_client = admin_mc.client
+    catalog_name1 = random_str()
+    catalog_name2 = random_str()
+    app_name1 = random_str()
+    app_name2 = random_str()
+    url = "https://github.com/rancher/integration-test-charts"
```
dramich: This already exists at the top of the file as `CATALOG_URL`

tests/integration/suite/test_app.py
```diff
@@ -591,6 +591,62 @@ def _app_fail():
...
+    catalog1 = catalog_client.create_catalog(name=catalog_name1,
+                                             branch="master",
+                                             url=url,
+                                             )
+    catalog2 = catalog_client.create_catalog(name=catalog_name2,
+                                             branch="master",
+                                             url=url,
+                                             helmVersion="helm_v3"
+                                             )
+    remove_resource(catalog1)
```
dramich: Keep the `remove_resource` right under the create call in case the next create call explodes we will still clean up the first resource.

tests/integration/suite/test_app.py
dramich: tests/integration/suite/conftest.py

tests/integration/suite/conftest.py
```diff
@@ -20,6 +20,8 @@
     create_cluster, \
     import_cluster
 
+# This is where all charts for testing are
+CATALOG_URL = "https://github.com/rancher/integration-test-charts"
```
dramich: This already exists in this file https://github.com/rancher/rancher/blob/master/tests/integration/suite/conftest.py#L46

Issue #25754, opened 2020-03-04
Add Helm 3 chart catalog to repo


brendarearden:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement/feature

**Explanation :**
Helm versions are determined at the catalog level, so new Rancher charts that are helm 3 compatible will need to be in their own catalog.  

**Result:**
New helm 3 catalog (https://github.com/rancher/helm3-charts)  needs to be added https://git.rancher.io
and needs to be added into Rancher

paraglade:
helm chart updated to pull in helm3-charts repo and production upgraded:
 #https://github.com/rancher/eio-charts/pull/32
```
┌─[paraglade][angus][~/brain/tmp]
└─▪ git clone https://git.rancher.io/helm3-charts
Cloning into 'helm3-charts'...
remote: Counting objects: 9, done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 9 (delta 0), reused 0 (delta 0)
Unpacking objects: 100% (9/9), done.
┌─[paraglade][angus][~/brain/tmp]
└─▪ ls -la helm3-charts
total 32
drwxr-xr-x 4 paraglade paraglade 4096 Mar  4 13:42 .
drwxr-xr-x 5 paraglade paraglade 4096 Mar  4 13:42 ..
-rw-r--r-- 1 paraglade paraglade  335 Mar  4 13:42 .drone.yml
drwxr-xr-x 8 paraglade paraglade 4096 Mar  4 13:42 .git
-rw-r--r-- 1 paraglade paraglade   10 Mar  4 13:42 .gitignore
-rw-r--r-- 1 paraglade paraglade 5656 Mar  4 13:42 README.md
drwxr-xr-x 2 paraglade paraglade 4096 Mar  4 13:42 test
┌─[paraglade][angus][~/brain/tmp]
```

deniseschannon:
@brendarearden can you add helm3-library pointing to `git.rancher.io/helm3-charts` in this section?
https://github.com/rancher/rancher/blob/master/app/catalog_data.go

brendarearden:
Yes, I will add that in and get a PR up

jiaqiluo:
Rancher master-head 9269970

confirm that
- a new helm 3 catalog named `helm3-library` is enabled by default
- and it points to `git.rancher.io/helm3-charts`

[...]

PR #25818, opened 2020-03-06
Add helm3 catalog data

brendarearden:
**Problem:**
With helm version being set at the catalog level, we need a helm 3 specific catalog

**Solution**
Add helm 3 specific catalog to Rancher

**Issue**
#25754

### DIFF
--- a/app/catalog_data.go
+++ b/app/catalog_data.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/rancher/rancher/pkg/catalog/helm"
+	"github.com/rancher/rancher/pkg/controllers/user/helm/common"
 	"github.com/rancher/rancher/pkg/settings"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config"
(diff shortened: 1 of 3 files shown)


app/catalog_data.go
```diff
@@ -27,6 +27,11 @@ const (
 	systemLibraryBranch = "master"
 	systemLibraryName   = "system-library"
 	defSystemChartVer   = "management.cattle.io/default-system-chart-version"
+
+	helm3LibraryURL    = "https://git.rancher.io/helm3-charts"
+	helm3LibraryBranch = "master"
+	helm3LibraryName   = "helm3-library"
+	helm3HelmVersion   = "helm_v3"
```
dramich: Is there a const in `pkg/catalog/helm` that can be used here? This should be defined in one place to keep it consistent for future changes.

app/catalog_data.go
```diff
@@ -154,7 +167,7 @@ func syncCatalogs(management *config.ManagementContext) error {
 	)
 }
 
-func doAddCatalogs(management *config.ManagementContext, name, url, branch string, bundledMode bool) error {
+func doAddCatalogs(management *config.ManagementContext, name, url, branch string, helmVersion string, bundledMode bool) error {
```
dramich: Update to drop the `string` after `branch`

brendarearden:
@cbron confirmed with Denise that the naming helm3-library is expected

Issue #25785, opened 2020-03-05
Panic: creating CRD store etcdserver: request timed out


soumyalj:
Version: master-head - Rancher version 53e0430e1

Steps:
1. Imported a RKE cluster into the Rancher setup and deleted it. 
2. Created a DO cluster
Rancher server crashed with the below panic:
```
2020/03/05 01:03:46 [INFO] Creating system token for u-4cuf72xlif, token: agent-u-4cuf72xlif-jbw72
2020/03/05 01:03:46 [INFO] Registering project network policy
2020/03/05 01:03:46 [INFO] Registering CIS controller
2020/03/05 01:03:46 [INFO] registering podsecuritypolicy cluster handler for cluster c-kjp8x
2020/03/05 01:03:46 [INFO] registering podsecuritypolicy project handler for cluster c-kjp8x
2020/03/05 01:03:46 [INFO] registering podsecuritypolicy namespace handler for cluster c-kjp8x
2020/03/05 01:03:46 [INFO] registering podsecuritypolicy serviceaccount handler for cluster c-kjp8x
2020/03/05 01:03:46 [INFO] registering podsecuritypolicy template handler for cluster c-kjp8x
2020/03/05 01:03:46 [INFO] Registering monitoring for cluster "c-kjp8x"
2020/03/05 01:03:46 [INFO] Registering istio for cluster "c-kjp8x"
2020/03/05 01:03:46 [INFO] Creating CRD clusterauthtokens.cluster.cattle.io
panic: creating CRD store etcdserver: request timed out

goroutine 57731 [running]:
github.com/rancher/rancher/vendor/github.com/rancher/norman/store/crd.(*Factory).BatchCreateCRDs.func1(0xc0004b7600, 0xc0004b7620, 0x2, 0x2, 0xc00135e820, 0x6b1d1e0, 0x462d220, 0xc0014d17d0, 0x3dce065, 0x4)
	/go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/store/crd/init.go:65 +0x2db
created by github.com/rancher/rancher/vendor/github.com/rancher/norman/store/crd.(*Factory).BatchCreateCRDs
	/go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/store/crd/init.go:50 +0xce


```

cjellick:
This is interesting because that would actually be the downstream cluster that is failing to create the CRDs. What were the specs of the nodes in the downstream cluster? Like iops, cpu, memory

[...]

PR #25820, opened 2020-03-06
Don't panic when CRD creation fails

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/25785

PR #25824, opened 2020-03-06
[FEATURE] Post a elasticsearch test message with a bulk request.

BobVanB:
**problem**
With elasticsearch the test button is posting to `<cluster>/<index>/container_log`
For ES6 this is fine, but ES7 and higher types are not allowed anymore.

**suggestion**
* Fluentd ES test button with POST _bulk
* Fluentd ES test button removes the _type containerlog

**support**
ES6: type and id can be mixed
ES7: still allows a single type
ES8: types are going bye bye

**bulk request**
fluentd is posting with _bulk requests on root  
Because we use a loadbalancer that only allows _bulk requests,  
i would like to test my route with a bulk request
```
	// elasticsearch <= 6
	url.Path = path.Join(url.Path, index, "/container_log")
	bulkTestData = httpTestData
	
	// elasticsearch 7
	url.Path = path.Join(url.Path, "/_bulk")
	bulkTestData = []byte(`{"index":{"_index":"` + index + `","_type":"container_log"}}\n` + httpTestData)

	// elasticsearch 8
	url.Path = path.Join(url.Path, "/_bulk")
	bulkTestData = []byte(`{"index":{"_index":"` + index + `"}\n` + httpTestData)
	
	req, err := http.NewRequest(http.MethodPost, url.String(), bytes.NewReader(bulkTestData))
```
**better**
I Don't know what the plans are with elasticsearch 8, but i would suggest to use the one without `_type`. Because the only thing that is tested is if the index exists. Better test would be to send a request parsed by fluentd. This way it will also build a correct mapping file.

### DIFF
--- a/pkg/controllers/user/logging/utils/elasticsearch.go
+++ b/pkg/controllers/user/logging/utils/elasticsearch.go
@@ -44,10 +44,12 @@ func (w *elasticsearchTestWrap) TestReachable(ctx context.Context, dial dialer.D
 	}
 
 	index := getIndex(w.DateFormat, w.IndexPrefix)
+	url.Path = path.Join(url.Path, "/_bulk")
 
-	url.Path = path.Join(url.Path, index, "/container_log")
-
-	req, err := http.NewRequest(http.MethodPost, url.String(), bytes.NewReader(httpTestData))
+	bulkTestData := []byte(`{"index":{"_index":"` + index + `","_type":"container_log"}}` + "\n")
+	bulkTestData = append(bulkTestData, httpTestData...)
+	bulkTestData = append(bulkTestData, "\n"...)
+	req, err := http.NewRequest(http.MethodPost, url.String(), bytes.NewReader(bulkTestData))
 	if err != nil {
 		return errors.Wrap(err, "create request failed")
 	}


maggieliu:
related to: https://github.com/rancher/rancher/issues/26994

BobVanB:
Hi @aiwantaozi , i dont know which version we had. It was early version 7. I will spin up a few clusters and try them. I will mention you when i've done it. With kind regards Bob.

BobVanB:
@aiwantaozi Hi, i had not tested it. I am currently doing it now. But compiling, starting and testing takes some time. I had indeed an error in it. The _bulk request needs to end with a newline.
I running the test now and will send an update as soon as possible.

```
	bulkTestData := []byte(`{"index":{"_index":"` + index + `"}}` + "\n")
	append(bulkTestData, httpTestData...)
	append(bulkTestData, byte("\n"))
	req, err := http.NewRequest(http.MethodPost, url.String(), bytes.NewReader(bulkTestData))
```

BobVanB:
@aiwantaozi Finally tested the current change.
```
Rancher type message
latest oss/7.5.2 Type is depricated
notype oss/7.5.2 OK
notype oss/7.6.2 OK
notype oss/6.8.6 Type is missing
notype oss/6.8.9 Type is missing
notype 7.5.2 OK
notype 7.6.2 OK
notype 6.8.9 Type is missing
type oss/7.6.2 [types removal] Specifying types in bulk requests is deprecated.
type oss/6.8.9 OK
```
My conclusion is that there has to be some advanced logic to ensure the correct message is used.
Or support for ES 6.X has to go.
For now i return the type and used the _bulk request in this pull request.

Hope this will help for now.

paynejacob:
Tested with es6 and es7 LGTM

BobVanB:
@cbron yes. Thank you. I was already looking for a way to use the same code from fluent. Because that is first retrieving the version. If ES8 is coming i will probably make a new pull.
Thanks for all.

PR #25826, opened 2020-03-06
Save created obj

dramich:
Fixes panic seen at https://drone-pr.rancher.io/rancher/rancher/7685/1/2 line 6974


pkg/controllers/user/logging/deployer/appdeployer.go
```diff
@@ -84,10 +84,10 @@ func (d *AppDeployer) deploy(app *projectv3.App, systemWriteKeys []string) error
...
-
-	if current.DeletionTimestamp != nil {
-		return errors.New("stale app " + app.Namespace + ":" + app.Name + " still on terminating")
+	} else {
+		if current.DeletionTimestamp != nil {
+			return errors.New("stale app " + app.Namespace + ":" + app.Name + " still on terminating")
+		}
 	}
 
 	new := current.DeepCopy()
```
cbron: Wouldn't this be a nil pointer too. If somehow it's getting through error check but error exists _and_ current is nil that would be case here too. Would need to either check `current != nil` or re-assign in create.

Issue #25790, opened 2020-03-05
Rancher server keeps restarting


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
- Deploy a cluster in 1.16 and upgrade cluster to 1.17 k8s version
- Rancher server keeps restarting every now and then. Logs are as follows
**Rancher logs:**
```
2020/03/05 00:48:59 [INFO] Creating roleBinding User user-q94zv Role project-owner
2020/03/05 00:48:59 [ERROR] NamespaceController kube-system [namespace-auth] failed with : clusterroles.rbac.authorization.k8s.io "p-khxrd-namespaces-readonly" already exists
2020/03/05 00:49:00 [INFO] Creating roleBinding User user-q94zv Role admin
2020/03/05 00:49:00 [INFO] Creating clusterRoleBinding for project access to global resource for subject user-q94zv role create-ns.
2020/03/05 00:49:00 [INFO] Creating roleBinding User user-q94zv Role admin
2020/03/05 00:49:00 [INFO] Creating clusterRoleBinding for project access to global resource for subject user-q94zv role create-ns.
2020-03-05 00:49:05.941269 W | etcdserver: read-only range request "key:\"/registry/mutatingwebhookconfigurations\" range_end:\"/registry/mutatingwebhookconfigurationt\" count_only:true " with result "range_response_count:0 size:5" took too long (125.833865ms) to execute
2020/03/05 00:49:05 [INFO] Updating workload [ingress-nginx/nginx-ingress-controller] with public endpoints [[{"nodeName":"c-lrtzs:m-q2clc","addresses":["134.209.166.225"],"port":80,"protocol":"TCP","podName":"ingress-nginx:nginx-ingress-controller-47fs5","allNodes":false},{"nodeName":"c-lrtzs:m-q2clc","addresses":["134.209.166.225"],"port":443,"protocol":"TCP","podName":"ingress-nginx:nginx-ingress-controller-47fs5","allNodes":false}]]
2020/03/05 00:49:06 [INFO] [mgmt-auth-prtb-controller] Creating clusterRole p-hjd4z-projectmember
2020/03/05 00:49:08 [ERROR] ProjectController c-lrtzs/p-hjd4z [system-image-upgrade-controller] failed with : upgrade cluster c-lrtzs system service logging failed: cluster c-lrtzs not ready
[...]

PR #25828, opened 2020-03-06
Fix panic in access control lookup

ibuildthecloud:
When multiple goroutines are spawned per request and access control is checked across multiple contexts there ends up being multiple locks for the same map. The fix is to have the lock instantiated at the same time as the protected map.


deniseschannon:
https://github.com/rancher/rancher/issues/25790

PR #25831, opened 2020-03-06
vendor update types

kinarashah:


tests/integration/suite/test_node.py
```diff
@@ -365,4 +367,4 @@ def test_no_node_template(user_mc):
             clusterId="local")
     assert e.value.error.status == 404
     assert e.value.error.message == \
-        "unable to find node template [%s]" % invalid_template_id
+        "unable to find node template [%s]" % invalid_template_id
```
rmweir: new line at bottom? probably will break tox

daxmc99: Hmm, looks like flake8 is failing

Issue #25817, opened 2020-03-06
Change the feature flag name for Steve and add descriptions


maggieliu:
Need to change the feature flag name to "Dashboard" for Steve and update the description.

deniseschannon:
@rmweir We also need to change it so that the description of feature flags come from the API instead of the UI.

rancher-max:
#### Validated in master-2408-head `Rancher version e78e43fea`
![image](https://user-images.githubusercontent.com/55997940/76337759-0f172880-62b5-11ea-9331-eab6f7894c5f.png)
- `dashboard` is now shown instead of `steve`
- Related to #25909 where more detail was added to validations for descriptions.

PR #25848, opened 2020-03-06
Change feature flag name and add description

rmweir:
MERGE TYPES FIRST: https://github.com/rancher/types/pull/1120

Change feature flag name from steve to dashboard.

**Issue:** https://github.com/rancher/rancher/issues/25817

### DIFF
--- a/pkg/features/feature.go
+++ b/pkg/features/feature.go
@@ -15,18 +15,31 @@ import (
 var (
 	features = make(map[string]*Feature)
 
-	// Features, ex.: ClusterRandomName = newFeature("cluster-randomizer", false)
+	// Features, ex.: ClusterRandomName = newFeature("cluster-randomizer", "Randomizes clusters.", false, false)
 
-	UnsupportedStorageDrivers = newFeature("unsupported-storage-drivers", false, true)
-	IstioVirtualServiceUI     = newFeature("istio-virtual-service-ui", true, true)
-	Steve                     = newFeature("dashboard", true, false)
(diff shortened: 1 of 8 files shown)


deniseschannon:
@rmweir as part of this request, we had discussed wanting to also update so feature flags take descriptions from the API instead of from the UI. Can you make a separate issue and PR for that request?

cbron:
Don’t have full context here, but shouldn’t the description be getting used/set somewhere ?

rmweir:
> Don’t have full context here, but shouldn’t the description be getting used/set somewhere ?

you're right, also have to make another change, will reping when ready

deniseschannon:
@rmweir can you update the other 2 feature flags and pull the description from the UI. Ping @vincent99 for best text for description of Steve.

pkg/features/feature.go
```diff
@@ -87,6 +100,15 @@ func InitializeFeatures(ctx *config.ScaledContext, featureArgs string) {
 				}
 			}
 
+			// checks if developer has changed description value from previous rancher version
+			if featureState.Status.Description != f.description {
+				newFeatureState := featureState.DeepCopy()
+				newFeatureState.Status.Description = f.description
+				if featureState, err = ctx.Management.Features("").Update(newFeatureState); err != nil {
+					logrus.Errorf(errMsg, f.name, err)
```
cbron: Should either return nil here or not assign to same variable as it could be mangled/empty coming back, and then the function would continue with that.

rmweir: good catch

Issue #25834, opened 2020-03-06
Helm 3 should not need to start Tiller


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy rancher server. Tail logs
2. Add a cluster
3. Add a helm_v3 catalog. I used `https://argoproj.github.io/argo-helm`
4. Deploy an app from the helm_v3 catalog.

**Result:**
In the rancher server logs, it shows:
```
[INFO] Installing chart using helm version: helm_v3
[main] 2020/03/06 18:22:25 Starting Tiller v2.16+unreleased (tls=false)
```

Helm 3 does not use Tiller so shouldn't need to start it.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `master-2373-head Rancher version beba5247e`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Linode
- Machine type (cloud/VM/metal) and specifications (CPU/memory): 3 nodes, all roles, each 4gb 2cpu
- Kubernetes version (use `kubectl version`): `v1.17.3`

rancher-max:
#### Validated on master-2399-head `Rancher version 926997017`

Logs now only show `[INFO] Installing chart using helm version: helm_v3` and don't mention Tiller when using helm3. This is the case for install, upgrade, and rollback scenarios. The apps still perform those operations successfully. Delete scenario shows `2020/03/09 16:56:45 [INFO] Deleting chart using helm version: helm_v3` and also correctly deletes. 

When using helm2, it does correctly still use Tiller and shows this in the logs for install, upgrade, and rollback scenarios:
```
[INFO] Installing chart using helm version: rancher-helm
[main] 2020/03/09 16:52:55 Starting Tiller v2.16.3-rancher1 (tls=false)
```
Delete scenario for helm2 has the same logs but shows "Deleting" instead of "Installing" and correctly deletes the app.

PR #25854, opened 2020-03-07
Add start tiller helm version check

brendarearden:
**Problem**
Tiller was starting even when running helm3 commands

**Solution**
Add check to determine if tiller should be started or not

**Issue**
#25834

### DIFF
--- a/pkg/controllers/user/helm/common.go
+++ b/pkg/controllers/user/helm/common.go
@@ -58,12 +58,14 @@ func helmInstall(tempDirs *common.HelmPath, app *v3.App) error {
 	cont, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	addr := common.GenerateRandomPort()
-	go func() {
-		err := common.StartTiller(cont, tempDirs, addr, app.Spec.TargetNamespace)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/helm/common.go
```diff
@@ -58,25 +58,31 @@ func helmInstall(tempDirs *common.HelmPath, app *v3.App) error {
...
 	defer cancel()
 	addr := common.GenerateRandomPort()
-	go func() {
-		err := common.StartTiller(cont, tempDirs, addr, app.Spec.TargetNamespace)
-		if err != nil {
-			logrus.Errorf("got error while stopping tiller, error message: %s", err.Error())
-		}
-	}()
+	hVersion := app.Status.HelmVersion
+	if hVersion == common.HelmV2 || hVersion == "" {
```
cbron: Didn’t we add helper func for this check ?

kinarashah: possible to use existing `IsHelm3`?

kinarashah: sorry, didn't see @cbron's comment till I refreshed

Issue #25554, opened 2020-02-25
Custom cluster provisioning stuck in "provisioning" state with error " Failed to bring up Etcd Plane: etcd cluster is unhealthy: hosts [<ip>] failed to report healthy"


sangeethah:
Rancher server version - Build from master - `60f0a4c4c`

Rancher server  - Single node rancher install

Steps to reproduce the problem:
Create a custom cluster with following configuration:
1 control node
1 etcd node
3 worker nodes

Custom cluster provisioning stuck in "provisioning" state with error  - " Failed to bring up Etcd Plane: etcd cluster is unhealthy: hosts [<ip>] failed to report healthy"

![imageedit__2070187990](https://user-images.githubusercontent.com/4266958/75221933-cf4e2e00-5757-11ea-9fee-3ee03d8020e2.png)

etcd container logs:
```
2020-02-25 05:48:38.765716 W | pkg/flags: unrecognized environment variable ETCD_UNSUPPORTED_ARCH=x86_64
[WARNING] Deprecated '--logger=capnslog' flag is set; use '--logger=zap' flag instead
2020-02-25 05:48:38.765815 I | etcdmain: etcd Version: 3.4.3
2020-02-25 05:48:38.765828 I | etcdmain: Git SHA: 3cf2f69b5
2020-02-25 05:48:38.765833 I | etcdmain: Go Version: go1.12.12
2020-02-25 05:48:38.765838 I | etcdmain: Go OS/Arch: linux/amd64
2020-02-25 05:48:38.765843 I | etcdmain: setting maximum number of CPUs to 2, total number of available CPUs is 2
[WARNING] Deprecated '--logger=capnslog' flag is set; use '--logger=zap' flag instead
2020-02-25 05:48:38.765982 I | embed: peerTLS: cert = /etc/kubernetes/ssl/kube-etcd-xxxx.pem, key = /etc/kubernetes/ssl/kube-etcd-xxxx-key.pem, trusted-ca = /etc/kubernetes/ssl/kube-ca.pem, client-cert-auth = true, crl-file =
2020-02-25 05:48:38.766546 I | embed: name = etcd-ip-xxxx
2020-02-25 05:48:38.766620 I | embed: data dir = /var/lib/rancher/etcd/
2020-02-25 05:48:38.766634 I | embed: member dir = /var/lib/rancher/etcd/member
2020-02-25 05:48:38.766639 I | embed: heartbeat = 500ms
[...]

Issue #25524, opened 2020-02-21
Sometimes custom cluster provisioning stuck in "Updating" state with "401" unauthorized error 


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Install Rancher master-head in HA using k3s nodes.
https://staging.rancher.com/docs/rancher/v2.x/en/installation/k8s-install/create-nodes-lb/
Rancher install command:
```
helm install rancher rancher-alpha/rancher --namespace cattle-system --set hostname=<hostname> --version 2.4.0-alpha1  --set rancherImageTag=master-head
```
In master-head, provision a custom 3 node cluster using Linux instances (3 nodes, all roles each)

**Result:**
Cluster fails to be provisioned and user cannot access.
Fails with:
```
	[etcd] Failed to bring up Etcd Plane: Container [etcd-fix-perm] did not complete in time on host [18.222.106.119]; [etcd] Failed to bring up Etcd Plane: Container [etcd-fix-perm] did not complete in time on host [18.222.106.119]; [controlPlane] Failed to bring up Control Plane: [Failed to verify healthcheck: Service [kube-apiserver] is not healthy on host [18.222.106.119]. Response code: [401], response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401} , log: W0221 23:15:04.288799 1 clientconn.go:1120] grpc: addrConn.createTransport failed to connect to {https://172.31.13.3:2379 0 <nil>}. Err :connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"kube-ca\")". Reconnecting...]
```
![Screen Shot 2020-02-21 at 4 15 11 PM](https://user-images.githubusercontent.com/57409052/75079570-64d28f00-54c6-11ea-943c-cd690693132c.png)

User cannot access cluster and is blocked by 503 error:
![Screen Shot 2020-02-21 at 4 15 33 PM](https://user-images.githubusercontent.com/57409052/75079551-55ebdc80-54c6-11ea-975d-87242137e87c.png)
Dev console:
![Screen Shot 2020-02-21 at 4 15 23 PM](https://user-images.githubusercontent.com/57409052/75079565-62703500-54c6-11ea-892b-9461949584c3.png)

[...]

PR #25898, opened 2020-03-09
Modify GenericEncryptedStore set secret logic

mrajashree:
Throughout cluster provisioning kontainer-engine saves the cluster with different statuses by calling GenericEncryptedStore.Set
Sometimes creating/updating these secrets wasn't saving the cluster with the desired status, leading to some intermittent errors during cluster provisioning.
This commit adds the following changes to ensure cluster gets saved with the correct status
1. If cluster secret create returns IsAlreadyExists, get the secret object and update
it with current expected value
2. When updating a secret, update the object obtained from DeepCopy and not the one returned
from lister.

https://github.com/rancher/rancher/issues/25554
https://github.com/rancher/rancher/issues/25524

### DIFF
--- a/pkg/encryptedstore/secrets_store.go
+++ b/pkg/encryptedstore/secrets_store.go
@@ -2,11 +2,14 @@ package encryptedstore
 
 import (
 	"reflect"
+	"time"
 
 	v1 "github.com/rancher/types/apis/core/v1"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


mrajashree:
@cbron The previous logic was retrying 5 times, so I've kept the retries for exponentialBackoff same as that. And ideally since the first retry will get object from db the update should succeed, but I can reduce the backoff duration

pkg/encryptedstore/secrets_store.go
```diff
@@ -63,38 +66,84 @@ func (g *GenericEncryptedStore) getKey(name string) string {
...
+	}
+	return wait.ExponentialBackoff(backoff, func() (bool, error) {
+		// fetch secret from the db when retrying due to IsConflict/IsAlreadyExists error
+		secret, err := g.secrets.GetNamespaced(g.namespace, g.getKey(name), metav1.GetOptions{})
+		if err != nil {
+			logrus.Errorf("[GenericEncryptedStore]: error getting secret %v from db: %v", g.getKey(name), err)
+			return false, err
 		}
+		secToUpdate := prepareSecretForUpdate(secret, data)
+		if !reflect.DeepEqual(secToUpdate, secret) {
```
cjellick: I know this DeepEqual existed in the old code, so you didn't introduce it, but since we are in here, we should consider taking the opportunity to improve it.
DeepEquals on k8s objects compare the entire object, which is often more than we care about. I've seen this lead to funky bugs. Do you need to compare the whole object or can you just deepEqual on secret.Data?

mrajashree: makes sense to compare just the secret.Data, changed it

Issue #25877, opened 2020-03-08
CIS scan with Profile=NULL runs RKE-CIS-1.5 Permissive scan but the results are incorrect


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- From the API run scan with default values for all input values except `Profile=NULL`
- The scan run on the Rancher UI shows it is RKE-CIS-1.5 Permissive scan run.

<img width="1271" alt="Screen Shot 2020-03-07 at 9 31 47 PM" src="https://user-images.githubusercontent.com/26032343/76157299-b2045280-60bb-11ea-9009-cf2fe7e65761.png">

**Scan run with Profile=NULL**
<img width="1323" alt="Screen Shot 2020-03-07 at 9 35 53 PM" src="https://user-images.githubusercontent.com/26032343/76157302-c0526e80-60bb-11ea-94fd-6fca57af4de1.png">

**Actual RKE-CIS-1.5 Permissive run on the cluster**
<img width="1312" alt="Screen Shot 2020-03-07 at 9 36 03 PM" src="https://user-images.githubusercontent.com/26032343/76157307-cb0d0380-60bb-11ea-8d2c-626192e9af27.png">

**Expected Result:**
@leodotcloud Question:
When Profile=NULL, what should be the expected CIS scan run? Is it RKE-CIS-1.5 Permissive scan by default? 

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `63306835a`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke DO 
- Kubernetes version (use `kubectl version`):

```
1.17
```

leodotcloud:
@sowmyav27 yes, if no profile is specified, permissive is used. Will fix the issue.

sowmyav27:
@leodotcloud @deniseschannon @cloudnautique At the API level, when no profile is specified, should Permissive be the default profile to be used for the scan?

maggieliu:
Permissive should be the default.

izaac:
Rancher version: `2.4 master-head (03/10/2020)` commit id: e78e43fea

Using NULL in the API now triggers the`Permissive` profile by default and benchmark version 1.5. 
Triggering with null matches the results of running the `rke-cis-1.5` Permissive test results.

Issue #25855, opened 2020-03-07
cronSchedule does not have a default value


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- From the UI, Give the `Recurring Security Scan Interval (cron)` = <blank> while creating the cluster.
- The cron value is NOT set for the cluster.

**Expected Result:**

- When no value is sent for the `Recurring Security Scan Interval (cron)`, API should have the default value in place for the `cronSchedule` param
- UI should send the default value for the field when the cluster is being created.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head latest - commit id: `63306835a`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom rke
- Kubernetes version (use `kubectl version`):

```
1.17
```

izaac:
#25933

izaac:
Rancher version: `2.4 master-head (03/10/2020)` commit id: 8ede17d10

Default value from API configuration is now `0 0 0 * *`  Validated this is getting set properly.

Issue #25838, opened 2020-03-06
Panic in rancher logs after a scheduled CIS scan run


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Schedule a CIS scan in a cluster, through the API
- Give in `cisScanConfig` through View in API --> Edit --> `scheduledClusterScan`

```
cisScanConfig": {
"overrideBenchmarkVersion": "rke-cis-1.4",
"profile": "permissive"
},
```
- ScheduleConfig set to 
```
"scheduleConfig": {
"cronSchedule": "*/3 * * * *",
"retention": 2,
"type": "/v3/schemas/scheduledClusterScanConfig"
},
```
- When the Scan is run on the cluster, Panic is seen, and the Scan is stuck in `Creating` state.

**Result:**

- Panic seen in rancher logs
- Rancher logs:

```
2020/03/06 20:33:09 [INFO] Marking CIS scan complete: ss-cis-1583526780000249907
2020/03/06 20:33:09 [ERROR] ClusterScanController c-p2t4r/ss-cis-1583526780000249907 [cisScanHandler] failed with : cisScanHandler: Updated: runner pod not yet deleted, will retry
2020/03/06 20:33:09 [INFO] Deleting chart using helm version: rancher-helm
[main] 2020/03/06 20:33:10 Starting Tiller v2.16+unreleased (tls=false)
[main] 2020/03/06 20:33:10 GRPC listening on :39688
[main] 2020/03/06 20:33:10 Probes listening on :38630
[main] 2020/03/06 20:33:10 Storage driver is ConfigMap
[main] 2020/03/06 20:33:10 Max history per release is 10
[storage] 2020/03/06 20:33:10 getting release history for "ss-cis-1583526780000249907"
[tiller] 2020/03/06 20:33:10 uninstall: Deleting ss-cis-1583526780000249907
[tiller] 2020/03/06 20:33:10 executing 0 pre-delete hooks for ss-cis-1583526780000249907
[tiller] 2020/03/06 20:33:10 hooks complete for pre-delete ss-cis-1583526780000249907
[storage] 2020/03/06 20:33:10 updating release "ss-cis-1583526780000249907.v1"
2020/03/06 20:33:10 [ERROR] ClusterScanController c-p2t4r/ss-cis-1583526780000249907 [cisScanHandler] failed with : cisScanHandler: Updated: runner pod not yet deleted, will retry
[...]

Issue #25837, opened 2020-03-06
Config maps left behind after a failed CIS Scan run has been deleted


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Run a CIS scan on a cluster.
- Scan gets stuck in "Creating" state, because of a known issue.
- Delete the CIS scan.
- The config maps related to this scan are NOT deleted.

**Expected Result:**
Config maps related the deleted failed CIS scan must get cleaned up

<img width="1441" alt="Screen Shot 2020-03-06 at 1 45 20 PM" src="https://user-images.githubusercontent.com/26032343/76125867-95c8be80-5fb2-11ea-998e-e6f77b664d41.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - latest - commit id: `beba5247e`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom cluster
- Kubernetes version (use `kubectl version`):

```
1.17
```

izaac:
Rancher version: `2.4 master-head (03/10/2020)` commit id: e78e43fea

I was able to get the scan stuck in a "Running" state then after a minute I've just deleted the scan, all its related config maps were removed after doing this.

PR #25899, opened 2020-03-09
cis bug fixes

leodotcloud:
Addresses:
- https://github.com/rancher/rancher/issues/25877
- https://github.com/rancher/rancher/issues/25838
- https://github.com/rancher/rancher/issues/25855
- https://github.com/rancher/rancher/issues/25837

### DIFF
--- a/pkg/api/store/cluster/cluster_store.go
+++ b/pkg/api/store/cluster/cluster_store.go
@@ -22,6 +22,7 @@ import (
 	ccluster "github.com/rancher/rancher/pkg/api/customization/cluster"
 	"github.com/rancher/rancher/pkg/api/customization/clustertemplate"
 	"github.com/rancher/rancher/pkg/clustermanager"
+	"github.com/rancher/rancher/pkg/controllers/management/cis"
 	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
 	"github.com/rancher/rancher/pkg/controllers/management/clusterstatus"
 	"github.com/rancher/rancher/pkg/controllers/management/etcdbackup"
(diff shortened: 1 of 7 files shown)


pkg/api/store/cluster/cluster_store.go
```diff
@@ -669,6 +673,44 @@ func enableLocalBackup(data map[string]interface{}) {
...
+	if cisScanConfig["profile"] == "" {
+		values.PutValue(data, v3.CisScanProfileTypePermissive, "scheduledClusterScan", "scanConfig", "cisScanConfig", "profile")
+	}
+}
+
+func handleScheduledScan(data map[string]interface{}) {
+	scheduledClusterScan, ok := values.GetValue(data, "scheduledClusterScan")
+	if ok && scheduledClusterScan != nil {
+		enabled := values.GetValueN(data, "scheduledClusterScan", "enabled")
+		if enabled == nil || (enabled != nil && !enabled.(bool)) {
```
rmweir: nit: you can drop `enabled != nil`, if it were nil this wouldn't be evaluated

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -385,3 +399,42 @@ func (csh *cisScanHandler) deleteApp(appInfo *appInfo) error {
 
 	return nil
 }
+
+func (csh *cisScanHandler) ensureCleanup(cs *v3.ClusterScan) error {
```
rmweir: can owner references be used instead to clean any of this up?

pkg/api/store/cluster/cluster_store.go
leodotcloud: I did think about that, I felt having that additional condition made the code easier to read.

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: The installation tools (helm, sonobuoy) already use owner references. This is a fallback cleaning mechanism in case something bad happens.

Issue #25885, opened 2020-03-08
Lots of info logs relating to "nodeplan in nil" when provisioning cluster with node drivers.


sangeethah:
Rancher Server version - `1c78296da`

Steps to reproduce the problem:
Provision a DO cluster.

Lots of info logs relating to "nodeplan in nil" when provisioning cluster with node drivers seen in rancher logs:

```2020/03/08 20:32:23 [INFO] Generating and uploading node config sanw1
2020/03/08 20:32:23 [INFO] nodePlan is nil for node m-bwlrv
2020/03/08 20:32:24 [INFO] Handling backend connection request [c-gmkpj:m-v4mhl]
2020/03/08 20:32:25 [INFO] nodePlan is nil for node m-rvvlx
2020/03/08 20:32:28 [INFO] nodePlan is nil for node m-jbtjr
2020/03/08 20:32:28 [INFO] nodePlan is nil for node m-bwlrv
2020/03/08 20:32:30 [INFO] Provisioning node sane1 done
2020/03/08 20:32:30 [INFO] Generating and uploading node config sane1
2020/03/08 20:32:30 [INFO] Provisioning cluster [c-gmkpj]
2020/03/08 20:32:30 [INFO] Creating cluster [c-gmkpj]
2020/03/08 20:32:30 [INFO] nodePlan is nil for node m-rvvlx
2020/03/08 20:32:31 [INFO] Handling backend connection request [c-gmkpj:m-sfp6q]
2020/03/08 20:32:31 [INFO] nodePlan is nil for node m-sfp6q
2020/03/08 20:32:33 [INFO] nodePlan is nil for node m-jbtjr
2020/03/08 20:32:33 [INFO] nodePlan is nil for node m-bwlrv
2020/03/08 20:32:35 [INFO] kontainerdriver rancherkubernetesengine listening on address 127.0.0.1:41459
2020/03/08 20:32:35 [INFO] cluster [c-gmkpj] provisioning: Initiating Kubernetes cluster
2020/03/08 20:32:35 [INFO] cluster [c-gmkpj] provisioning: [dialer] Setup tunnel for host [165.22.33.72]
2020/03/08 20:32:36 [INFO] cluster [c-gmkpj] provisioning: [dialer] Setup tunnel for host [167.71.163.5]
2020/03/08 20:32:36 [INFO] cluster [c-gmkpj] provisioning: [dialer] Setup tunnel for host [167.71.174.150]
[...]

PR #25901, opened 2020-03-09
add more context for node plan nil log msg

kinarashah:
https://github.com/rancher/rancher/issues/25885


kinarashah:
@superseb Thanks, moved it to trace. The node plan is nil when node's provisioning or hasn't been updated in rkeConfig so rke doesn't return a plan for it and I wanted to not keep on getting requests so kept it at 10s. But since it's just for node creation, I agree with the reasoning that we want nodes to come up faster. I've updated it to 2s, in line with how agent earlier used to retry every 2s if it didn't get a plan.

kinarashah:
Based on offline discussion, 
before this PR: server returns the error and agent would retry with a 5 second sleep and I see agent retrying at least 2-3 times
now: server doesn't return the error, it'll just log in Trace and just asking agent to retry... so I'm going to change it to 15 seconds

Issue #25575, opened 2020-02-25
OPA Gatekeeper - Upon Deleting Gatekeeper App, delete built in contrainttemplates


prachidamle:
OPA Gatekeeper app is deployed via system chart: https://github.com/rancher/rancher/issues/23753

The chart also installs the ConstraintTemplate CRDs and few builtin ConstraintTemplates.

After Gatekeeper is deleted from the cluster,  CRDs and CRs installed by helm are not cleaned up via helm. 

Rancher needs to handle deletion of these resources after deleting Gatekeeper.

prachidamle:
The issue with constraintTemplate CRD and constraints left behind after deleting gatekeeper seems to be an issue with gatekeeper. 

It has been documented on gatekeeper and they ask to manually cleanup the ConstraintTemplate CRD and resources before deleting gatekeeper: https://github.com/open-policy-agent/gatekeeper#uninstallation

I have tested gatekeeper on a vanilla kubernetes cluster on GKE:

1) Installed gatekeeper using the helm installation instructions on https://github.com/open-policy-agent/gatekeeper
Chart used: https://github.com/open-policy-agent/gatekeeper/tree/master/chart/gatekeeper-operator

2) This helm chart from gatekeeper installed the constrainttemplate CRD on the cluster. 
I then also installed few constraints.

3) Upon uninstalling gatekeeper via helm, the constrainttemplate CRD and constraints were not cleaned up.

khushboo-rancher:
Tested on master-2450-head- 73f603f9f
On deleting gatekeeper app, constraint template isn't getting deleted if we have constraint under it.

Steps followed.
1. Enabled OPA gatekeeper for a cluster.
2. Create a constraint under 'k8srequiredlabels' template.
sample : 
```
apiVersion: constraints.gatekeeper.sh/v1beta1
kind: K8sRequiredLabels
metadata:
  name: ingress-must-have-test-label
spec:
  match:
    kinds:
    - apiGroups:
      - extensions
      kinds:
      - Ingress
  parameters:
    message: All Ingress must have an `test` label
    labels:
    - key: test
      allowedRegex: "^[a-zA-Z]"
```
3. Delete the app using disable OPA gatekeeper option.
4.  'k8srequiredlabels' doesn't get deleted.

khushboo-rancher:
Tested on master-head 888029fb0.
Disabling Gatekeeper deletes built in templates and user created templates too.

Steps:
1. Enable gatekeeper
[...]

PR #25919, opened 2020-03-10
Deleting OPA Gatekeeper constrainttemplates when gatekeeper app is removed

prachidamle:
https://github.com/rancher/rancher/issues/25575

Adding in logic to clean up OPA Gatekeeper constrainttemplates during removal of the app as part of the Helm lifecycle remove.

Not adding an extra controller since that would need us to add a finalizer on the app, and finalizer is known to possibly cause blocking issues with deletion.
Instead adding this part of the existing lifecycle handler.

### DIFF
--- a/pkg/controllers/user/helm/controller.go
+++ b/pkg/controllers/user/helm/controller.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"net/url"
 	"os"
 	"reflect"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/helm/controller.go
```diff
@@ -257,6 +261,17 @@ func (l *Lifecycle) Remove(obj *v3.App) (runtime.Object, error) {
 	if err != nil {
 		return obj, err
 	}
+
+	if l.isGatekeeperApp(obj) {
+		logrus.Debugf("Removing GatekeeperApp: %v", obj.Name)
+		//remove the constrainttemplates.
+		err := l.handleConstraintTemplateCleanup(obj)
+		if err != nil {
+			logrus.Errorf("error while deleting gatekeeper constraints for cluster %v: %v", err, l.ClusterName)
```
StrongMonkey: I think if you throw error the framework will print it out. It should be better to use error.Wrapf(). Otherwise the same error will print out twice

cjellick:
I don't really like the approach of putting this very specific logic into the existing helm controller. This controller is already complex and brittle. Setting the precedent of putting app specific logic into it would make it even worse.

I don't yet have a fully formed alternative solution. I acknowledge it is a tricky problem.

prachidamle:
Closing this in favor of: https://github.com/rancher/system-charts/pull/182

Issue #19279, opened 2019-03-30
info pointed by catalog URL makes rancher crash at ~2 minutes interval


fgeorgatos:
**What kind of request is this (question/bug/enhancement/feature request):**

bug: unexpected container crash (i.e. a heavy slowdown)

**Steps to reproduce (least amount of steps as possible):**

add catalog https://swissdatasciencecenter.github.io/helm-charts/

**Result:**

DoS, with this log visible:

```
2019/03/30 00:16:03 [ERROR] CatalogController renku-helm-charts [catalog] failed with : failed to update templates. Multiple error occurred: [CatalogTemplateVersion.management.cattle.io "renku-helm-charts-renku-notebooks-0.2.1-gputest-" is invalid: metadata.name: Invalid value: "renku-helm-charts-renku-notebooks-0.2.1-gputest-": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')] | stdout
-- | --
```

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.2.1
- Installation option (single install/HA): single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM
- Kubernetes version (use `kubectl version`): 1.11.9 & 1.13.5 - verified in both

```
2019/03/30 00:16:03 [ERROR] CatalogController renku-helm-charts [catalog] failed with : failed to update templates. Multiple error occurred: [CatalogTemplateVersion.management.cattle.io "renku-helm-charts-renku-notebooks-0.2.1-gputest-" is invalid: metadata.name: Invalid value: "renku-helm-charts-renku-notebooks-0.2.1-gputest-": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')] | stdout
-- | --
```

- Docker version (use `docker version`):

```
[...]

Issue #24914, opened 2020-01-10
[helm]: Reject repos with local:// URLs in index.yaml


daenney:
When using Artifactory with Helm it has 2 types of repositories, local, and virtual. Virtual are the ones you have to use with Rancher, and will generate the `index.yaml` with full URLs that can then be used to fetch the charts.

If you add a Artifactory local Helm Chart by accident Rancher doesn't catch this but errors out when trying to deploy an app, with an "unsupported protocol local://". Now you have a broken app in installing state that you can't even delete (that's really 2 bugs).

When adding a catalog to Rancher it would be helpful to sanity check the URLs in the `index.yaml` that gets retrieved and error out, or warn the user, that this won't work.

jef:
This also doesn't work with `file://`. I'm getting a stalled page like mentioned in https://github.com/rancher/rancher/issues/24927.

bmdepesa:
rancher/rancher:v2.3.5

**First pass**
I didn't reproduce this exactly, but added 2 catalogs containing an index.yaml with `file://` and `local://` urls. Using these catalogs, the UI goes into an indefinite loading state (spinner spinning and UI grayed out).

brendarearden:
https://github.com/rancher/rancher/issues/24863. << this is likely fixed by the PR related to this issue

daenney:
@brendarearden How do you mean? It's a fairly different issue, and there doesn't seem to be a PR associated with it?

brendarearden:
My bad - I linked them the wrong way around - the refactoring completed to address this issue may have fixed a few other catalog issues we have been facing. I am going to remove my initial comment on this one link it the correct way.

bmdepesa:
rancher/rancher:master-head

* Add a catalog with a chart containing a `local://` and `file://` urls
* The chart does not sync and an error message is shown in the UI

daenney:
For those looking for the code, #25920.

PR #25920, opened 2020-03-10
Check chart urls are valid

brendarearden:
**Problem**
Some apps are attempted to be deployed with invalid urls which results in an error

**Solution**
Add an error on the catalog and forget the errors for invalid chart urls to ensure it does not keep refreshing through it

**Issue:**
https://github.com/rancher/rancher/issues/24914

**Dependent on:**
https://github.com/rancher/wrangler/pull/70
https://github.com/rancher/types/pull/1121
https://github.com/rancher/integration-test-charts/pull/12

### DIFF
--- a/pkg/catalog/manager/catalog_preprocess.go
+++ b/pkg/catalog/manager/catalog_preprocess.go
@@ -1,0 +1,92 @@
+package manager
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+
+	"github.com/rancher/rancher/pkg/catalog/helm"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+)
+
+const invalidURLErrMsg string = "a valid url must consist of an http, https, or git scheme"
+
+type ChartsWithErrors struct {
+	Template *helm.ChartVersion
+	error    InvalidChartError
+}
+
+type InvalidChartError struct {
+	Err          string
+	TemplateName string
+}
+
+func (e *InvalidChartError) Error() string {
+	return fmt.Sprintf("failed %v %s", e.TemplateName, e.Err)
+}
+
+// Go through each catalog entry and check for errors in values used for template name and labels
+// due to restrictions for kubernetes objects. Do not create templates with errored charts.
+func (m *Manager) preprocessCatalog(catalog *v3.Catalog, catalogEntries map[string]helm.ChartVersions) ([]ChartsWithErrors, map[string]helm.ChartVersions) {
+	var invalidEntries []ChartsWithErrors
+	for chart, metadata := range catalogEntries {
+		for _, version := range metadata {
+			var versionErrors []string
+			// get a chart name that matches helm general conventions
+			templateName := getValidTemplateName(catalog.Name, chart)
+			// check that template name could be used as a kubernetes sub domain
+			errorString := validation.IsDNS1123Subdomain(templateName)
+			if len(errorString) > 0 {
+				versionErrors = append(versionErrors, fmt.Sprintf("template name %s: [%s]", templateName, strings.Join(errorString, ";")))
+			}
+			// check that template name can be used as a label
+			errorString = validation.IsValidLabelValue(templateName)
+			if len(errorString) > 0 {
+				versionErrors = append(versionErrors, fmt.Sprintf("template name %s: [%s]", templateName, strings.Join(errorString, ";")))
+			}
+			templateVersionName := getValidTemplateNameWithVersion(templateName, version.Version)
+			// validate template version name can be used as a kubernetes sub domain
+			errorString = validation.IsDNS1123Subdomain(templateVersionName)
+			if len(errorString) > 0 {
+				versionErrors = append(versionErrors, fmt.Sprintf("template version name %s: [%s]", templateVersionName, strings.Join(errorString, ";")))
+			}
+			if version.URLs != nil {
+				for _, url := range version.URLs {
+					// check to see if url is http, https or git scheme
+					if isInvalidVersionURL(url) {
+						versionErrors = append(versionErrors, fmt.Sprintf("url %s is invalid. %s", url, invalidURLErrMsg))
+					}
+				}
+			}
+			// a chart can have multiple version, so errors are collected per version
+			// this allows valid chart versions to be processed for use
+			if len(versionErrors) > 0 {
+				invalidEntries = append(invalidEntries, ChartsWithErrors{
+					Template: version,
+					error: InvalidChartError{
+						Err:          strings.Join(versionErrors, ";"),
+						TemplateName: version.Name,
+					},
+				})
+			}
+		}
+	}
+	// remove invalid entry from catalog entries so that they are not processed into templates
+	for _, version := range invalidEntries {
+		delete(catalogEntries, version.Template.Name)
+	}
+	return invalidEntries, catalogEntries
+}
+
+func isInvalidVersionURL(versionURL string) bool {
+	url, err := url.Parse(versionURL)
+	if err != nil {
+		return false
+	}
+	if url.Scheme == "file" || url.Scheme == "local" {
+		return true
+	}
+	return false
+}
--- a/pkg/catalog/manager/catalog_sync.go
+++ b/pkg/catalog/manager/catalog_sync.go
@@ -37,6 +37,13 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 		}
 	}
 
+	// if the catalog was processed but some templates had errors due to local/file urls or chart names
+	// that cannot be used as labels - the catalog is up to date, but had errors so we don't refresh
+	// to give time for the user to make the necessary corrections.
+	if v3.CatalogConditionProcessed.IsFalse(catalog) && v3.CatalogConditionRefreshed.IsTrue(catalog) {
+		return nil, nil
+	}
+
 	commit, helm, err := helmlib.NewForceUpdate(catalog)
 	if err != nil {
 		return m.updateCatalogError(catalog, err)
--- a/pkg/catalog/manager/catalog_sync.go
+++ b/pkg/catalog/manager/catalog_sync.go
@@ -37,13 +37,6 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 		}
 	}
 
-	// if the catalog was processed but some templates had errors due to local/file urls or chart names
-	// that cannot be used as labels - the catalog is up to date, but had errors so we don't refresh
-	// to give time for the user to make the necessary corrections.
-	if v3.CatalogConditionProcessed.IsFalse(catalog) && v3.CatalogConditionRefreshed.IsTrue(catalog) {
-		return nil, nil
-	}
-
 	commit, helm, err := helmlib.NewForceUpdate(catalog)
 	if err != nil {
 		return m.updateCatalogError(catalog, err)
--- a/pkg/catalog/manager/cluster_catalog_sync.go
+++ b/pkg/catalog/manager/cluster_catalog_sync.go
@@ -31,13 +31,6 @@ func (m *Manager) ClusterCatalogSync(key string, obj *v3.ClusterCatalog) (runtim
 		return nil, err
 	}
 
-	// if the catalog was processed but some templates had errors due to local/file urls or chart names
-	// that cannot be used as labels - the catalog is up to date, but had errors so we don't refresh
-	// to give time for the user to make the necessary corrections.
-	if v3.CatalogConditionProcessed.IsFalse(clusterCatalog) && v3.CatalogConditionRefreshed.IsTrue(clusterCatalog) {
-		return nil, nil
-	}
-
 	commit, helm, err := helmlib.NewForceUpdate(&clusterCatalog.Catalog)
 	if err != nil {
 		return m.updateClusterCatalogError(clusterCatalog, err)
--- a/pkg/catalog/manager/cluster_catalog_sync.go
+++ b/pkg/catalog/manager/cluster_catalog_sync.go
@@ -31,6 +31,13 @@ func (m *Manager) ClusterCatalogSync(key string, obj *v3.ClusterCatalog) (runtim
 		return nil, err
 	}
 
+	// if the catalog was processed but some templates had errors due to local/file urls or chart names
+	// that cannot be used as labels - the catalog is up to date, but had errors so we don't refresh
+	// to give time for the user to make the necessary corrections.
+	if v3.CatalogConditionProcessed.IsFalse(clusterCatalog) && v3.CatalogConditionRefreshed.IsTrue(clusterCatalog) {
+		return nil, nil
+	}
+
 	commit, helm, err := helmlib.NewForceUpdate(&clusterCatalog.Catalog)
 	if err != nil {
 		return m.updateClusterCatalogError(clusterCatalog, err)
--- a/pkg/catalog/manager/crd.go
+++ b/pkg/catalog/manager/crd.go
@@ -3,7 +3,6 @@ package manager
 import (
 	"fmt"
 	"reflect"
-	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/rancher/pkg/catalog/utils"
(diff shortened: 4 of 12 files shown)


pkg/catalog/manager/traverse.go
```diff
@@ -273,6 +292,20 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
+				maxTemplateNames = forgetErrorsLen
+			}
+			errorMessage := fmt.Sprintf("%s .... %v charts with invalid template url or name", strings.Join(errTemplateNames[:maxTemplateNames], " ; "), forgetErrorsLen)
+			setCatalogIgnoreErrorErrorState(cmt, catalog, projectCatalog, clusterCatalog, fmt.Sprintf("Error syncing catalog: %v", errorMessage))
+			if _, err := m.updateCatalogInfo(cmt, catalogType, "", false, true); err != nil {
+				return err
+			}
+			logrus.Error(fmt.Sprintf("failed to sync templates. Multiple error(s) occured: %v", forgetErrors))
+			return &controller.ForgetError{Err: errors.Errorf(fmt.Sprintf("failed to sync templates. Multiple error(s) occured: %v", forgetErrors))}
+		}
```
cbron: So if there is a different error and an error here, this block would be skipped, is that ok ?

pkg/catalog/manager/crd.go
```diff
@@ -12,13 +13,23 @@ import (
...
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 const (
 	CatalogNameLabel  = "catalog.cattle.io/name"
 	TemplateNameLabel = "catalog.cattle.io/template_name"
 )
 
+type ForgettableError struct {
```
cbron: Interesting idea here. I don't think this name is most appropriate though. Something more like `InvalidChartError` ?

pkg/catalog/manager/traverse.go
```diff
@@ -314,7 +347,43 @@ type catalogYml struct {
...
+	}
+	return true
+}
+
+func getVersionsWithValidURL(urls []string) []string {
+	var validURLs []string
+	for _, url := range urls {
+		if validVersionURL(url) {
+			validURLs = append(validURLs, url)
+		}
```
cbron: Shouldn't you bubble up an error here ? I see part above that has if 0 valid urls create an error, but do they need to know which error ?

tests/integration/suite/test_catalog.py
```diff
@@ -313,11 +313,52 @@ def get_errored_catalog(catalog):
...
     client.delete(catalog)
     wait_for_template_to_be_deleted(client, name)
     assert not client.list_catalog(name=name).data
+
+
+def test_invalid_catalog_chart_urls(admin_mc, remove_resource):
+    """Test chart with file:// and local:// url paths"""
+    client = admin_mc.client
+    name = random_str()
+    url = "https://github.com/brendarearden/integration-test-charts"
```
cbron: This will need to get switched to rancher

tests/integration/suite/test_catalog.py
```diff
@@ -313,11 +313,52 @@ def get_errored_catalog(catalog):
...
+        catalog = client.reload(catalog)
+        if catalog.transitioning == "error":
+            return catalog
+        return None
+    catalog = wait_for(lambda: get_errored_catalog(catalog),
+                       fail_handler=lambda:
+                       "catalog was not found in error state")
+    templates = client.list_template(catalogId=catalog.id).data
+
+    assert "aks-goodbyeworld" not in templates
```
cbron: Is this a bad url in the chart ?

pkg/catalog/manager/traverse.go
```diff
@@ -314,7 +347,43 @@ type catalogYml struct {
...
 	v3.CatalogConditionRefreshed.False(catalog)
 	v3.CatalogConditionRefreshed.Message(catalog, fmt.Sprintf("Error syncing catalog %v", catalog.Name))
+	v3.CatalogConditionProcessed.True(catalog)
 	cmt.catalog = catalog
 	cmt.projectCatalog = projectCatalog
 	cmt.clusterCatalog = clusterCatalog
 }
+
+func validVersionURL(versionURL string) bool {
+	if strings.HasPrefix(versionURL, "local://") || strings.HasPrefix(versionURL, "file://") {
```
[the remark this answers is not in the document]
cbron: Follow up here, this logic I linked is primarily for catalog not charts. Charts get deployed from app lifecycle, and they follow different pattern than catalogs. They could potentially use http/https/git protocols, so we can check for those specifically here or leave as-is.

pkg/catalog/manager/traverse.go
brendarearden: https://github.com/rancher/rancher/blob/2e612b488c6ff6cb4bc3313411f34e16b641a0fb/pkg/catalog/manager/crd.go#L95. It gets bubbled here

pkg/catalog/manager/traverse.go
brendarearden: yes because each `InvalidChartError` also gets added to the create and update errors, so they will get logged even if this portion gets skipped. It just ensures that if there is a weird error, we retry catalog refresh

tests/integration/suite/test_catalog.py
brendarearden: It is a bard url in the index.yaml of the chart repository - so its the url the deploy app is going to use to get the chart files - I added comments to help a little

pkg/catalog/manager/crd.go
```diff
@@ -28,12 +39,21 @@ func (m *Manager) createTemplate(template v3.CatalogTemplate, catalog *v3.Catalo
 	for i := range template.Spec.Versions {
 		template.Spec.Versions[i].Files = nil
 		template.Spec.Versions[i].Readme = ""
+
```
cbron: nit: unecessary space

pkg/catalog/manager/traverse.go
```diff
@@ -314,7 +348,47 @@ type catalogYml struct {
...
+	v3.CatalogConditionProcessed.Message(catalog, message)
+	v3.CatalogConditionRefreshed.Message(catalog, "")
+	v3.CatalogConditionProcessed.ReasonAndMessageFromError(catalog, errors.New(message))
+	v3.CatalogConditionRefreshed.True(catalog)
+	cmt.catalog = catalog
+	cmt.projectCatalog = projectCatalog
+	cmt.clusterCatalog = clusterCatalog
+}
+
+func onlyForgetErrors(forgetErrors, createErrors, updateErrors []error) bool {
```
cbron: forgot to change this name

pkg/catalog/manager/crd.go
```diff
@@ -72,6 +91,9 @@ func (m *Manager) updateTemplate(template *v3.CatalogTemplate, toUpdate v3.Catal
 		For each templateVersion in toUpdate, if spec doesn't match, do update
 		For version that doesn't exist, create a new one
 	*/
+	if len(toUpdate.Spec.Versions) == 0 || toUpdate.Spec.Versions == nil {
```
rmweir: nit: could probably do this before grabbing and iterating template versions

pkg/catalog/manager/catalog_sync.go
```diff
@@ -37,6 +37,11 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 		}
 	}
 
+	// check to see if catalog was processed successfully, but had errors while processing
+	if v3.CatalogConditionProcessed.IsFalse(catalog) && v3.CatalogConditionRefreshed.IsTrue(catalog) {
+		return nil, nil
```
rmweir: I think the comment above can be reworded, saying that it processed successfully but had errors is confusing. I also think there should be a comment explaining why you would return without an error despite process failing.

pkg/catalog/manager/traverse.go
```diff
@@ -273,6 +293,20 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 		errstrings = append(errstrings, fmt.Sprintf("failed to update templates. Multiple error(s) occurred: %v", updateErrors))
 	}
 	if len(errstrings) > 0 {
+		if isChartInvalidErrors(InvalidChartErrors, createErrors, updateErrors) {
+			maxTemplateNames := 5
+			forgetErrorsLen := len(InvalidChartErrors)
+			if forgetErrorsLen < 5 {
+				maxTemplateNames = forgetErrorsLen
+			}
+			errorMessage := fmt.Sprintf("%s .... %v charts with invalid template url or name", strings.Join(errTemplateNames[:maxTemplateNames], " ; "), forgetErrorsLen)
```
rmweir: maybe say actual number of templates failed, but 5 shown if forget is more than max

pkg/catalog/manager/catalog_sync.go
```diff
@@ -37,6 +37,13 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 		}
 	}
 
+	// if the catalog was processed but some templates had errors due to local/file urls or chart names
+	// that cannot be used as labels - the catalog is up to date, but had errors so we don't refresh
+	// to give time for the user to make the necessary corrections.
+	if v3.CatalogConditionProcessed.IsFalse(catalog) && v3.CatalogConditionRefreshed.IsTrue(catalog) {
```
dramich: Is this needed? We should be processing the whole catalog and collecting errors then updating the status and commit. The commit check below should stop the catalog from coming through again

pkg/catalog/manager/cluster_catalog_sync.go
```diff
@@ -31,6 +31,13 @@ func (m *Manager) ClusterCatalogSync(key string, obj *v3.ClusterCatalog) (runtim
 		return nil, err
 	}
 
+	// if the catalog was processed but some templates had errors due to local/file urls or chart names
+	// that cannot be used as labels - the catalog is up to date, but had errors so we don't refresh
+	// to give time for the user to make the necessary corrections.
+	if v3.CatalogConditionProcessed.IsFalse(clusterCatalog) && v3.CatalogConditionRefreshed.IsTrue(clusterCatalog) {
```
dramich: Same comment as above

pkg/catalog/manager/crd.go
```diff
@@ -12,14 +13,28 @@ import (
...
+	TemplateName string
+}
+
+func (e *InvalidChartError) Error() string {
+	return fmt.Sprintf("failed %v %s", e.TemplateName, e.Err)
+}
+
 func (m *Manager) createTemplate(template v3.CatalogTemplate, catalog *v3.Catalog) error {
+	// check to see if the template has valid urls - do not process if none exist
+	if len(template.Spec.Versions) == 0 || template.Spec.Versions == nil {
```
dramich: This check should be swapped around so the nil check is first.

pkg/catalog/manager/crd.go
```diff
@@ -12,14 +13,28 @@ import (
...
+		return &InvalidChartError{"must have one or more valid urls", template.Name}
```
dramich: The error doesn't convey what really happened, there isn't any data, not that it's invalid.

pkg/catalog/manager/crd.go
```diff
@@ -56,6 +76,10 @@ func (m *Manager) getTemplateMap(catalogName string, namespace string) (map[stri
 }
 
 func (m *Manager) updateTemplate(template *v3.CatalogTemplate, toUpdate v3.CatalogTemplate) error {
+	// check to see if the version has valid urls - do not process if none exist
+	if len(toUpdate.Spec.Versions) == 0 || toUpdate.Spec.Versions == nil {
+		return &InvalidChartError{"must have one or more valid urls", template.Name}
```
dramich: The error doesn't convey what really happened, there isn't any data, not that it's invalid.

pkg/catalog/manager/catalog_sync.go
brendarearden: not relying on commit to determine if refresh needs to occur again

pkg/catalog/manager/crd.go
brendarearden: removed in favor of catalog preprocess

brendarearden:
Changed to look for errors prior to processing catalog entries into templates. Added name and label processing in according with helm general conventions to ensure we are accepting what helm accepts.

pkg/catalog/manager/crd.go
brendarearden: removed as part of new implementation

pkg/catalog/manager/crd.go
brendarearden: removed as part of new implementation

brendarearden:
Per conversation with Dan I will be fixing the following items: 
1. remove the use of chart related naming conventions in favor of template and template version in the new pre process func 
2. do not truncate template name and template label strings and return error if larger than len 63 to keep inline with how it currently is
3. Return all errors in the UI instead of only truncated chart names to ensure all users can troubleshoot
4. Add url to the invalid url error message 
(if I missed any, feel free to add them)

pkg/catalog/manager/crd.go
```diff
@@ -172,7 +168,7 @@ func (m *Manager) createTemplateVersions(catalogName string, versionsSpec []v3.T
 		templateVersion := &v3.CatalogTemplateVersion{}
 		templateVersion.Spec = spec
 		templateVersion.Status = v3.TemplateVersionStatus{HelmVersion: template.Status.HelmVersion}
-		templateVersion.Name = fmt.Sprintf("%s-%v", template.Name, spec.Version)
+		templateVersion.Name = getValidHelmChartLabel(template.Name, spec.Version)
```
dramich: Should this be passing through the label validation?

brendarearden: I added label validation to the template name and remove it from the template version name

pkg/catalog/manager/traverse.go
```diff
@@ -311,10 +324,19 @@ type catalogYml struct {
...
+	var invalidChartErrors []string
+	for _, errorInfo := range entriesWithErrors {
+		invalidChartErrors = append(invalidChartErrors, errorInfo.error.Error())
+	}
+
+	return strings.Join(invalidChartErrors, " ; ")
+}
+
+// Min returns the smaller of x or y.
+func Min(x, y int) int {
```
dramich: This doesn't appear to be used

pkg/catalog/manager/catalog_preprocess.go
```diff
@@ -0,0 +1,92 @@
...
+				versionErrors = append(versionErrors, fmt.Sprintf("template name %s: [%s]", templateName, strings.Join(errorString, ";")))
+			}
+			// check that name can be used as a label
+			errorString = validation.IsValidLabelValue(templateName)
+			if len(errorString) > 0 {
+				versionErrors = append(versionErrors, fmt.Sprintf("template name %s: [%s]", templateName, strings.Join(errorString, ";")))
+			}
+			// validate label that follows general conventions and can be used as kubernetes label
+			templateVersionName := getValidTemplateNameWithVersion(templateName, version.Version)
+			errorString = validation.IsDNS1123Subdomain(templateVersionName)
```
dramich: Is this the correct check here? Should it also be IsQualifiedName? And the comment above this needs to be updated

pkg/catalog/manager/traverse.go
```diff
@@ -311,10 +324,19 @@ type catalogYml struct {
 	Labels     map[string]string `yaml:"labels,omitempty"`
 }
 
-func setCatalogErrorState(cmt *CatalogInfo, catalog *v3.Catalog, projectCatalog *v3.ProjectCatalog, clusterCatalog *v3.ClusterCatalog) {
-	v3.CatalogConditionRefreshed.False(catalog)
-	v3.CatalogConditionRefreshed.Message(catalog, fmt.Sprintf("Error syncing catalog %v", catalog.Name))
-	cmt.catalog = catalog
-	cmt.projectCatalog = projectCatalog
-	cmt.clusterCatalog = clusterCatalog
+func processInvalidChartErrors(entriesWithErrors []ChartsWithErrors) string {
```
dramich: Use a string builder with the loop, it will be cleaner/faster/less memory https://golang.org/pkg/strings/#Builder

pkg/catalog/manager/catalog_preprocess.go
brendarearden: both checks need to be the same and they are not fully qualified

brendarearden:
Also resolves https://github.com/rancher/rancher/issues/19279

Issue #25914, opened 2020-03-09
Upgrading rancher causes cluster-alerting app to upgrade and subsequently fail


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Add a notifier (can use dummy values)
2. Add an alert group in a project (see screenshot at end for my configuration in System project)
3. Verify the cluster-alerting and monitoring-operator apps are fully deployed and are on version 0.0.7
4. Upgrade to master-head
5. Navigate to the System > Apps page. 

**Result:**
cluster-alerting has been upgraded to 0.0.8 and failed upgrade.
monitoring-operator is still 0.0.7 and is running successfully.

**Other details that may be helpful:**
cluster-alerting should upgrade by default, and that upgrade should pass. I see this in the server logs:
```
2020/03/09 22:13:49 [INFO] Installing chart using helm version: rancher-helm
[main] 2020/03/09 22:13:49 Starting Tiller v2.16.3-rancher1 (tls=false)
[main] 2020/03/09 22:13:49 GRPC listening on :50852
[main] 2020/03/09 22:13:49 Probes listening on :43292
[main] 2020/03/09 22:13:49 Storage driver is ConfigMap
[main] 2020/03/09 22:13:49 Max history per release is 10
[tiller] 2020/03/09 22:13:49 getting history for release cluster-alerting
[storage] 2020/03/09 22:13:49 getting release history for "cluster-alerting"
2020/03/09 22:13:49 http: proxy error: dial tcp 127.0.0.1:6443: connect: connection refused
[storage/driver] 2020/03/09 22:13:49 query: failed to query with labels: an error on the server ("") has prevented the request from succeeding (get configmaps)
[tiller] 2020/03/09 22:13:49 preparing update for cluster-alerting
[storage] 2020/03/09 22:13:49 getting deployed releases from "cluster-alerting" history
2020/03/09 22:13:49 http: proxy error: dial tcp 127.0.0.1:6443: connect: connection refused
[storage/driver] 2020/03/09 22:13:49 query: failed to query with labels: an error on the server ("") has prevented the request from succeeding (get configmaps)
[tiller] 2020/03/09 22:13:49 failed to prepare update: an error on the server ("") has prevented the request from succeeding (get configmaps)
[tiller] 2020/03/09 22:13:49 performing force update for cluster-alerting
[storage] 2020/03/09 22:13:49 getting last revision of "cluster-alerting"
[...]

Issue #25842, opened 2020-03-06
Endless loop rancher server logs on helm3 install failure


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Add catalog:  	https://argoproj.github.io/argo-helm specify helm2 name it helm2
2. Add catalog  	https://argoproj.github.io/argo-helm specify helm3 name it helm3
3. Using helm2, install app argo-cd and name it foo
4. Delete foo
5. Using helm3, install app argo-cd and name it foo

**Result:**
This fails as expected due to helm3 not having the `crd-install` hook option. However, there are endless logs on the rancher server every 1 second:
```
2020/03/06 22:28:37 [ERROR] AppController p-kctnb/foo [helm-controller] failed with : wait helm template rendering notes failed. manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: , name: foo-argocd-application-controller, existing_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole, new_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole
: exit status 1
```

**Other details that may be helpful:**
This happened for me in an upgraded setup. I previously had deployed a few apps in the setup before upgrading (using helm2) and used the same chart URL on my new catalog after upgrading, but had that one use helm3. 

Can workaround this by deleting the clusterroles and clusterrolebindings that were created in this case: `kubectl delete clusterroles foo-argocd-application-controller foo-argocd-server`
`kubectl delete clusterrolebindings foo-argocd-application-controller foo-argocd-server`

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `Rancher version 63306835a` (`master-2375-head`)
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): 3 Linode nodes all roles
- Machine type (cloud/VM/metal) and specifications (CPU/memory): 2cpu 4gb each node
- Kubernetes version (use `kubectl version`):

```
[...]

Issue #25742, opened 2020-03-03
Istio - steve blocks istio from finishing installation


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Run rancher:mater-head 3623201 single install
2. Enable Istio (with default options)

**Result:**
Navigating to System > Apps > cluster-istio after some time shows a failed Installing state:
![image](https://user-images.githubusercontent.com/55997940/75817361-29508400-5d54-11ea-96c4-2c79c60a2896.png)

**Other details that may be helpful:**
All pods appear to be running, and sidecar injection works properly even though cluster-istio is not shown as active. Istio Traffic Metrics page shows graphs correctly. Traffic Graphs page and links (prometheus, grafana, kiali, jaeger) all **do not work** and give 404 errors for all of their JS and CSS resources. The pages generally do not load.

Disabling Steve fixes the problems. However, the Kiali link and Traffics Graph page are still broken.

Likely related to https://github.com/rancher/rancher/issues/25741

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): DO 8gb 4vcpu 3 nodes all roles for each.
- Kubernetes version (use `kubectl version`): 

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

davidnuzik:
When I try this in latest master I actually get rancher server to crash. Reproduced twice. Simple enable Istio and you'll see tons of panics in rancher server logs (seems like infinite loop) then rancher server will eventually crash.

cbron:
Unable to replicate on master-head `53e0430e1` with HA setup. Both monitoring and istio install correctly in downstream cluster. I can share my setup, but might just retry it. 
[...]

Issue #25741, opened 2020-03-03
Monitoring - steve blocks monitoring from finishing installation 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- run rancher:mater-head 3623201ad single install
- add a cluster 
- enable the cluster monitoring 

**Result:**
- both the app `cluster-monitoring` and `monitoring-operator` are stuck in installing state 

<img width="1400" alt="Screen Shot 2020-03-03 at 1 05 49 PM" src="https://user-images.githubusercontent.com/6218999/75817628-b3005180-5d54-11ea-9ac9-f22823180f08.png">
<img width="1455" alt="Screen Shot 2020-03-03 at 1 33 58 PM" src="https://user-images.githubusercontent.com/6218999/75817955-4a65a480-5d55-11ea-9143-373e8963b7e6.png">
<img width="1429" alt="Screen Shot 2020-03-03 at 1 34 08 PM" src="https://user-images.githubusercontent.com/6218999/75817960-4b96d180-5d55-11ea-89e6-3077fa214076.png">

- Metrics show on the cluster dashboard, node page, etc, but the Grafana and Prometheus links are broken

**Other details that may be helpful:**
- disabling steve fixes the problem partially: Grafana and Prometheus are accessible, but apps show new errors
<img width="1413" alt="Screen Shot 2020-03-03 at 1 06 02 PM" src="https://user-images.githubusercontent.com/6218999/75817632-b4317e80-5d54-11ea-918a-d6f31f7bb248.png">
<img width="1419" alt="Screen Shot 2020-03-03 at 1 06 12 PM" src="https://user-images.githubusercontent.com/6218999/75817637-b562ab80-5d54-11ea-95fc-68c7e8dedcf7.png">

Rancher logs:
```
2020/03/03 20:05:24 [ERROR] AppController p-tjr5h/cluster-monitoring [helm-controller] failed with : failed to install app cluster-monitoring. Error: UPGRADE FAILED: a release named cluster-monitoring is in use, cannot re-use a name that is still in use
```

jiaqiluo:
New findings on rancher:master-head 17a40f502

- the cluster monitoring apps are deployed successfully
- but links to Grafana and Prometheus still not work
- after disabling the cluster monitoring, apps are stuck in uninstalling state 

[...]

PR #25926, opened 2020-03-10
Disable steve proxy by default for the majority of requests

ibuildthecloud:


maggieliu:
Related to issues:
#25914
#25842
#25742
#25741

Issue #25856, opened 2020-03-07
State Transition of the Cluster scan report on the UI is confusing


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- On a hardened cluster (CIS 1.4 compliant), , Go to Cluster --> Tools --> CIS Scans
- Click on Run Scan and select Run scan 1.4 (Permissive) and click on Run
- The scan report entry is added on the page
- State Transition of the report seen is Creating --> Running --> Active --> Fail
- When the state is Active
<img width="1297" alt="Screen Shot 2020-03-06 at 4 51 26 PM" src="https://user-images.githubusercontent.com/26032343/76133210-f4e6fd00-5fcb-11ea-8c17-0eec742c1555.png">

- When the state is Fail
<img width="1290" alt="Screen Shot 2020-03-06 at 4 51 37 PM" src="https://user-images.githubusercontent.com/26032343/76133212-f6182a00-5fcb-11ea-8cc9-20241ee3473f.png">

**Expected Result:**
- State `Active` seems confusing here. 
- Could we add a state to show, a report is being generated?

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - latest - commit id: 63306835a UI tag: latest2
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom rke
- Kubernetes version (use `kubectl version`):

```
1.17
```

sowmyav27:
**Verified on latest -master-head - commit id: `5ee70d014`**

- Deployed a 1 node cluster
- On running CIS 1.5 Permissive/Hardened cluster - the state transition is: Creating --> Running --> Reporting --> PASS/FAIL

Issue #25010, opened 2020-01-17
CIS Scan - UI - config.json does not accept single quotes for its value


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug 

**Steps to reproduce (least amount of steps as possible):**
- Deploy a cluster
- Through the API run CIS Scan on the cluster with the config map value set as 
`{"config.json": "{'skip':{'rke-cis-1.4':['1.1.2']}}"}`

**Actual Result:**
The scan does NOT skip the test.

**Expected Result:**
The Scan should skip the test - 1.1.2

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): single

codyrancher:
@maggieliu this is a backend issue because it's being run through the API and not the UI.

> Through the API run CIS Scan on the cluster with the config map value set as
{"config.json": "{'skip':{'rke-cis-1.4':['1.1.2']}}"}

As a side note. I don't think the expected result is correct. I think the api should respond with an error because the supplied configuration is not valid json. Valid json requires the use of double quotes.

vincent99:
Yes that is not JSON.  The test could be changed to (in the backend) if the config is unparseable instead of (apparently) proceeding with no custom config, so that you know something is wrong.

leodotcloud:
With the latest changes around using scan profiles, not applicable/skip checks from k-d-m, this issue is not relevant anymore.

sowmyav27:
**On latest master-head - commit id:`63306835a`**

- From the UI, pass config.json like this
<img width="1184" alt="Screen Shot 2020-03-06 at 10 09 04 PM" src="https://user-images.githubusercontent.com/26032343/76137962-71db9c00-5ff7-11ea-909e-0e3d1c6d8c4c.png">
- On running a Scan for cis 1.4 Hardened profile, the report generated is erred out
<img width="1283" alt="Screen Shot 2020-03-06 at 10 12 34 PM" src="https://user-images.githubusercontent.com/26032343/76137990-d139ac00-5ff7-11ea-8535-e687f83cf48d.png">

- Rancher logs:
```
2020/03/07 06:07:20 [INFO] Marking CIS scan complete: cis-1583561232635481655
[...]

PR #25927, opened 2020-03-10
cis bug fixes

leodotcloud:
addresses: https://github.com/rancher/rancher/issues/25856
~addresses: https://github.com/rancher/rancher/issues/25010~

### DIFF
--- a/pkg/api/customization/cluster/actions_cis.go
+++ b/pkg/api/customization/cluster/actions_cis.go
@@ -19,11 +19,6 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	NumberOfRetriesForClusterUpdate = 3
-	RetryIntervalInMilliseconds     = 5
-)
-
 func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiContext *types.APIContext) error {
 	var err error
 
--- a/pkg/api/customization/clusterscan/clusterscan.go
+++ b/pkg/api/customization/clusterscan/clusterscan.go
@@ -31,6 +31,7 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 		status := convert.ToMapInterface(s)
 		failed := false
 		completed := false
+		runCompleted := false
 		for _, cond := range convert.ToMapSlice(status["conditions"]) {
 			if cond["type"] == string(mgmtv3.ClusterScanConditionCompleted) && cond["status"] == "True" {
 				completed = true
@@ -38,6 +39,13 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 			if cond["type"] == string(mgmtv3.ClusterScanConditionFailed) && cond["status"] == "True" {
 				failed = true
 			}
+			if cond["type"] == string(mgmtv3.ClusterScanConditionRunCompleted) && cond["status"] == "True" {
+				runCompleted = true
+			}
+		}
+		if runCompleted {
+			resource.Values["state"] = "reporting"
+			resource.Values["transitioning"] = "yes"
 		}
 		if failed {
 			resource.Values["state"] = errorState


pkg/api/customization/cluster/actions_cis.go
```diff
@@ -73,16 +73,51 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
-			if kerrors.IsNotFound(err) {
+	validBenchmarkVersions, err := a.getValidBenchmarkVersions()
+	if err != nil {
+		logrus.Errorf("error fetching benchmark versions: %v", err)
+		return httperror.WrapAPIError(err, httperror.ServerError, "error fetching cis benchmark versions %v")
+	}
+
+	if cisScanConfig.OverrideBenchmarkVersion != "" &&
+		!validBenchmarkVersions[cisScanConfig.OverrideBenchmarkVersion] {
+		return httperror.WrapAPIError(err, httperror.InvalidOption,
```
rmweir: wouldn't err be nil?

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -73,16 +73,51 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
+			fmt.Sprintf("invalid override benchmark version specified"))
+	}
+
+	userContext, err := a.ClusterManager.UserContext(apiContext.ID)
+	if err != nil {
+		logrus.Errorf("error fetching user context for cluster %v: %v", apiContext.ID, err)
```
rmweir: nit: maybe add something about cis to these logs so its clear where they are coming from

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -73,16 +73,51 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
+	if ok {
+		// check for content
+		skipInfo := map[string]map[string][]string{}
+		if err := json.Unmarshal([]byte(skipInfoStr), &skipInfo); err != nil {
+			logrus.Errorf("error unmarshaling cis skip info from configmap: %v", err)
+			return httperror.WrapAPIError(err, httperror.InvalidFormat, fmt.Sprintf("invalid skip info found in configmap: %v", v3.ConfigMapNameForUserConfig))
+		}
+		versionsMap, ok := skipInfo["skip"]
+		if !ok {
+			return httperror.WrapAPIError(err, httperror.InvalidOption,
```
rmweir: Do we want to wrap err here?

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -73,16 +73,51 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
+				fmt.Sprintf("missing 'skip' key in configmap %v", v3.ConfigMapNameForUserConfig))
+		}
+		// check for versions
+		for version := range versionsMap {
+			if !validBenchmarkVersions[version] {
 				return httperror.WrapAPIError(err, httperror.InvalidOption,
```
rmweir: and here?

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -73,16 +73,51 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
+
+	userContext, err := a.ClusterManager.UserContext(apiContext.ID)
+	if err != nil {
+		logrus.Errorf("error fetching user context for cluster %v: %v", apiContext.ID, err)
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("error fetching user contect for cluster %v", apiContext.ID))
+	}
+
+	cm, err := userContext.Core.ConfigMaps(v3.DefaultNamespaceForCis).Controller().Lister().Get(v3.DefaultNamespaceForCis, v3.ConfigMapNameForUserConfig)
+	if err != nil && !kerrors.IsNotFound(err) {
```
prachidamle: Handle kerrors.IsNotFound(err)? We need to either retry or error out otherwise line#102 will panic

pkg/api/customization/cluster/actions_cis.go
```diff
@@ -73,16 +73,51 @@ func (a ActionHandler) runCisScan(actionName string, action *types.Action, apiCo
...
+	}
+
+	userContext, err := a.ClusterManager.UserContext(apiContext.ID)
+	if err != nil {
+		logrus.Errorf("error fetching user context for cluster %v: %v", apiContext.ID, err)
+		return httperror.WrapAPIError(err, httperror.ServerError,
+			fmt.Sprintf("error fetching user contect for cluster %v", apiContext.ID))
+	}
+
+	cm, err := userContext.Core.ConfigMaps(v3.DefaultNamespaceForCis).Controller().Lister().Get(v3.DefaultNamespaceForCis, v3.ConfigMapNameForUserConfig)
```
cjellick: I don't think that getting a lister from a userContext will work properly and consistently in an HA setup. I don't think you'll have access to the cache on all api server nodes.

pkg/api/customization/cluster/actionhandler.go
```diff
@@ -26,6 +27,7 @@ type ActionHandler struct {
 	ClusterTemplateRevisionClient v3.ClusterTemplateRevisionInterface
 	SubjectAccessReviewClient     v1.SubjectAccessReviewInterface
 	CisBenchmarkVersionLister     v3.CisBenchmarkVersionLister
+	cmLister                      rcorev1.ConfigMapLister
```
cjellick: not used?

cjellick:
How does this interact with scheduled scans? The actual test run needs to fail in a way that clearly states malformed json in that config map is the cause.

I don't really like parsing and retrieving a configmap in a downstream cluster as part of an API call. I'd rather just see the test fail in a clear and concise way.

pkg/api/customization/cluster/actionhandler.go
leodotcloud: Yup, left over. Cleaned up.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: copy, paste error, thanks for the catch.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: done.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: Thanks for the deep insight, I didn't know about this. I have changed it to Get instead of Lister.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: Changed the logic. Thanks for the catch.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: copy, paste error, thanks for the catch.

pkg/api/customization/cluster/actions_cis.go
leodotcloud: copy, paste error, thanks for the catch.

leodotcloud:
@cjellick updated to remove the fix for https://github.com/rancher/rancher/issues/25010

PR #25994, opened 2020-03-12
validate number of notReady nodes on cluster update

kinarashah:
With introduction of maxUnavailable, we want to make sure we check if there are ready nodes to upgrade or else just error out. 

https://github.com/rancher/rancher/issues/26005

### DIFF
--- a/pkg/api/store/cluster/cluster_store.go
+++ b/pkg/api/store/cluster/cluster_store.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/definition"
+	"github.com/rancher/norman/types/slice"
 	"github.com/rancher/norman/types/values"
 	ccluster "github.com/rancher/rancher/pkg/api/customization/cluster"
 	"github.com/rancher/rancher/pkg/api/customization/clustertemplate"
@@ -26,9 +27,12 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
 	"github.com/rancher/rancher/pkg/controllers/management/clusterstatus"
 	"github.com/rancher/rancher/pkg/controllers/management/etcdbackup"
+	"github.com/rancher/rancher/pkg/controllers/management/rkeworkerupgrader"
 	"github.com/rancher/rancher/pkg/namespace"
+	nodehelper "github.com/rancher/rancher/pkg/node"
 	"github.com/rancher/rancher/pkg/ref"
 	"github.com/rancher/rancher/pkg/settings"
+	rkeservices "github.com/rancher/rke/services"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	managementschema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
 	managementv3 "github.com/rancher/types/client/management/v3"
@@ -51,6 +55,8 @@ type Store struct {
 	KontainerDriverLister         v3.KontainerDriverLister
 	ClusterTemplateLister         v3.ClusterTemplateLister
 	ClusterTemplateRevisionLister v3.ClusterTemplateRevisionLister
+	NodeLister                    v3.NodeLister
+	ClusterLister                 v3.ClusterLister
 }
 
 type transformer struct {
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -187,3 +188,47 @@ func workerOnly(roles []string) bool {
...
+		if slice.ContainsString(node.Status.NodeConfig.Role, rkeservices.ControlRole) {
+			if nodehelper.IsMachineReady(node) {
+				controlReady++
+			} else {
+				controlNotReady++
+			}
+			continue
+		}
+		if slice.ContainsString(node.Status.NodeConfig.Role, rkeservices.ETCDRole) {
+			continue
```
cbron: If a box can be all 3 roles, wouldn't this logic skip some ? IE: box-a roles are `[ctrl, etcd, worker]` and box-b roles are `[worker]`, this would only check worker status on box-b.

kinarashah: This would count box-a as a `ctrl` node and box-b as `worker`. To clarify, if a node has all 3 roles, it's only considered as control node. I will rename the worker to `workerOnly`, does that answer your example?

pkg/api/store/cluster/cluster_store.go
```diff
@@ -941,3 +946,37 @@ func replaceWithOldSecretIfNotExists(oldData, newData map[string]interface{}, cl
...
+	}
+	spec, err := getRkeConfig(data)
+	if err != nil || spec == nil {
+		return err
+	}
+	status, err := getRkeConfig(existingData)
+	if err != nil || status == nil {
+		return err
+	}
+	if reflect.DeepEqual(status, spec) {
```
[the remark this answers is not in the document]
cbron: Answer: yes, node upgrade would only come from cluster spec change.

pkg/controllers/management/rkeworkerupgrader/node.go
cbron: yes, as long as we are ok with that scenario, and scenario where box is `[etcd, worker]`

dramich:
As this sits it would stop any update to the cluster if that check fails, even something as simple as updating the description.

kinarashah:
Doing node check only if rkeConfig changes, so updating description shouldn't be an issue https://github.com/rancher/rancher/pull/25994/files#diff-ad5bf6e8e88eb832a2fba190c8e1273eR986 

But I also added cluster condition checks, so we don't block a cluster which is already provisioning or upgrading

pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -185,3 +186,48 @@ func workerOnly(roles []string) bool {
 	}
 	return worker
 }
+
+func CanUpgrade(nodes []*v3.Node, upgradeStrategy *v3.NodeUpgradeStrategy) error {
```
cjellick: nit (doesnt block merging): why not have this function in the `pkg/api/store/clsuter` package? that's the only place it is used.

pkg/api/store/cluster/cluster_store.go
```diff
@@ -941,3 +948,47 @@ func replaceWithOldSecretIfNotExists(oldData, newData map[string]interface{}, cl
...
+	if err != nil {
+		return fmt.Errorf("error getting cluster, try again %v", err)
+	}
+	// no need to validate if cluster's already provisioning or upgrading
+	if v3.ClusterConditionProvisioned.IsUnknown(cluster) ||
+		v3.ClusterConditionUpgraded.IsUnknown(cluster) ||
+		v3.ClusterConditionUpdated.IsUnknown(cluster) {
+		return nil
+	}
+	getRkeConfig := func(data map[string]interface{}) (*v3.RancherKubernetesEngineConfig, error) {
```
cjellick: is there a reason to write this as a function literal? I feel the significantly increase the complexity and reduce the readability of the code.

pkg/controllers/management/rkeworkerupgrader/node.go
kinarashah: I didn't want to export `calculateMaxUnavailable` and include other node imports in store, since I already had them in here, moved it

Issue #25999, opened 2020-03-13
Nodes get upgraded even when it encounters error during drain.


sangeethah:
Rancher Server version - `dc0a89013`

Steps to reproduce the problem:
Create a cluster with following configuration with k8s version - `1.15`:
1 control
1 etcd
3 worker nodes.

Deploy a workload using an "Ephimeral Volume"  using "Empty Dir volume" option.

Upgrade  k8s version to `1.16` with Drain option on  (with safe option) and `Maxiumum Worker Nodes Unavailable` set to 1.

Control and etcd nodes get upgraded.
When worker node on which the pods (from the above workload )  resides are getting drained , 
drain fails with following error (which is expected) :
```error: cannot delete Pods with local storage (use --delete-local-data to override): default/santest1-546fc4f984-p9n7c```

I see this error happen twice after which , the node gets upgraded . Following this upgrade on other worker nodes succeed.

Expected Behavior:
In this case, we expect the worker node were the drain fails to get stuck in "Draining" state and not be upgraded. 

rancher server logs:

```
2020/03/13 02:49:56 [INFO] Draining node sanw1 in c-qfkxg with flags --delete-local-data=false --force=false --grace-period=0 --ignore-daemonsets=true --timeout=60s
2020/03/13 02:49:57 [ERROR] nodeDrain: kubectl error draining node [sanw1] in cluster [c-qfkxg]: 
 sanw1
error: cannot delete Pods with local storage (use --delete-local-data to override): default/santest1-546fc4f984-p9n7c

2020/03/13 02:49:57 [INFO] cluster [c-qfkxg] worker-upgrade: call upgrade to reconcile for node [m-f8lzr]
2020/03/13 02:49:57 [INFO] cluster [c-qfkxg] worker-upgrade: updated node [m-f8lzr] to drain
2020/03/13 02:49:57 [INFO] Draining node sanw1 in c-qfkxg with flags --delete-local-data=false --force=false --grace-period=0 --ignore-daemonsets=true --timeout=60s
2020/03/13 02:49:57 [INFO] cluster [c-qfkxg] worker-upgrade: call upgrade to reconcile for node [m-r77k8]
2020/03/13 02:49:57 [ERROR] NodeController c-qfkxg/m-r77k8 [rke-worker-upgrader] failed with : cluster [c-qfkxg] worker-upgrade: more than allowed nodes upgrading for cluster: unavailable 2 maxUnavailable 1
[...]

Issue #25946, opened 2020-03-11
Worker node is stuck in upgrading when another worker node is manually drained/cordoned during upgrade process


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy cluster - 1 etcd, 3 control 3 worker node (maxUnavailable_control = 2, maxUnavailable_worker =1, drain=True)
- Upgrade k8s version from 1.16 to 1.17
- Control and etcd node goes through upgrade process.
- When the worker node W1 starts upgrading -- is in stated Drained Upgrading, manually drain/cordon worker node W2
- Upgrade on W1 goes  thourgh as the k8s verison on the node is changed, But the node ist stuck in Upgrading state
- W2 is in Drained state.

<img width="1317" alt="Screen Shot 2020-03-10 at 7 41 17 PM" src="https://user-images.githubusercontent.com/26032343/76381822-d65a6c00-6313-11ea-8d76-2bc2c8c57e25.png">

**Expected Result:**
The node W1 should have finished upgrading and be available in Active state.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `570461823`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke DO
- Kubernetes version (use `kubectl version`):

```
1.16 to 1.17
```

sowmyav27:
**Verified on latest master-head - commit id: `e651c7a81`**

**Steps:**

- Deploy cluster - 1 etcd, 3 control 3 worker node (maxUnavailable_control = 2, maxUnavailable_worker =1, drain=True)
- Upgrade cluster - change max-pods value
- Control and etcd node goes through upgrade process.
- When the worker node W1 starts upgrading -- is in stated Drained Upgrading, manually drain worker node W2. W2 will be in Drained status
- Upgrade on W1 finishes. 
- W2 - node gets picked up and gets upgraded. 
- The other node - W3 also gets upgraded. And the cluster is Active

Expected:
W2 node should be in Drained state. Cluster should be in Upgrading state

PR #26016, opened 2020-03-13
more bugs fixes

kinarashah:
https://github.com/rancher/rancher/issues/25999: nodes get upgraded even if drain fails, this happens because node was considered prepared if node is cordoned/drained. If drain is set to true in cluster spec, only that should be considered. see https://github.com/rancher/rancher/pull/26016/files#diff-2eabb733bb0dba004b0c4c6d065eeb8cL170

https://github.com/rancher/rancher/issues/25946: even if unavailable nodes >= maxUnavailable nodes, we should still mark the upgraded nodes as active. there's no point in letting them stuck in upgrading. see https://github.com/rancher/rancher/pull/26016/files#diff-a16c43ed026936b620e2fdb9ca6196b8L273

https://github.com/rancher/rancher/issues/26023: all the fields should be read from cluster status, not spec. I am correctly doing this for rkeConfig, but maxUnavailable is getting read from spec. This ensures that we work with rkeConfig that has already gone through rke up. Fix is changing that`cluster.Spec` to `cluster.Status`.  

https://github.com/rancher/rancher/issues/25885: Changing log level for create/update node plan to debug.

### DIFF
--- a/pkg/controllers/management/rkeworkerupgrader/node.go
+++ b/pkg/controllers/management/rkeworkerupgrader/node.go
@@ -104,7 +104,7 @@ func skipNode(node *v3.Node) bool {
 	return false
 }
 
-func (uh *upgradeHandler) filterNodes(nodes []*v3.Node, expectedVersion int) (map[string]*v3.Node, map[string]*v3.Node, map[string]*v3.Node, map[string]*v3.Node, int, int, int) {
+func (uh *upgradeHandler) filterNodes(nodes []*v3.Node, expectedVersion int, drain bool) (map[string]*v3.Node, map[string]*v3.Node, map[string]*v3.Node, map[string]*v3.Node, int, int, int) {
 	done, upgrading, filtered := 0, 0, 0
 	toPrepareMap, toProcessMap, upgradedMap, notReadyMap := map[string]*v3.Node{}, map[string]*v3.Node{}, map[string]*v3.Node{}, map[string]*v3.Node{}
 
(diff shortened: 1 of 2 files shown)


StrongMonkey:
Can we separate into two PRs? Or are those commits related?

kinarashah:
@StrongMonkey I've put the first commit to a separate PR, and added explanation for each issue in the description

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -234,16 +237,31 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string) e
 		return err
 	}
 
-	toPrepareMap, toProcessMap, upgradedMap, notReadyMap, filtered, upgrading, done := uh.filterNodes(nodes, cluster.Status.NodeVersion)
+	upgradeStrategy := cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.UpgradeStrategy
```
kinarashah: thinking if it'd make sense to short circuit if cluster condition is updating

pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -162,12 +162,21 @@ func (uh *upgradeHandler) filterNodes(nodes []*v3.Node, expectedVersion int) (ma
...
+	return node.Spec.DesiredNodeUnschedulable == "true"
 }
 
-func preparedNode(node *v3.Node) bool {
-	return v3.NodeConditionDrained.IsTrue(node) || node.Spec.InternalNodeSpec.Unschedulable || v3.NodeConditionUpgraded.IsUnknown(node)
+func preparedNode(node *v3.Node, drain bool) bool {
+	if v3.NodeConditionUpgraded.IsUnknown(node) {
+		return true
+	}
+	if drain {
```
mrajashree: should we do this check first before `if v3.NodeConditionUpgraded.IsUnknown(node)`?

Issue #20209, opened 2019-05-13
vSphere Node vApp options not sticking


alexholliz:
**What kind of request is this (question/bug/enhancement/feature request): Bug**

**Steps to reproduce (least amount of steps as possible):**

- Create a Network Protocol Profile associated with your desired network with a pool of IPs for vApps to be assigned
- Create vsphere node template
- Fill out all fields as appropriate, set the network to the network that your network protocol profile is associated with, and leaving Host field blank for DRS enabled cluster
- Click the radio button under vApp options: Use vApp to configure networks with network protocol profiles
- Save Node Template

**Result:**
Nodes do not deploy with IP from the range specified, they default to DHCP

**Other details that may be helpful:**
I have attempted to modify the Network Protocol Profile to remove DHCP, as well as several redeployments of nodes from new templates and the result is the same. I also tried to "Provide a custom vApp config" and fill out the three present fields with:

- com.vmware.guestInfo
- IPv4
- fixedAllocated

Which did not work. I also tried to fill out the interface 0 details as represented in this issue: https://github.com/rancher/rancher/issues/15817
![image](https://user-images.githubusercontent.com/273599/57654370-9d3d5800-7599-11e9-9924-eaadf48989fb.png)

That also did not work. Checking the node in the API, it looks like those options aren't being set at all. When I open the edit page for the node template, it's defaulted back to "Do no use vApp".

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
![image](https://user-images.githubusercontent.com/273599/57654711-792e4680-759a-11e9-9758-df3bfd20f3b6.png)

- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type: Infrastructure Provider: vSphere
- Machine type: VM 
- Specifications: 8 vCPU 4GB memory
- Kubernetes version (use `kubectl version`): 

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-26T00:04:52Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #26017, opened 2020-03-13
[2.3] rancher machine version bump

luthermonson:


luthermonson:
https://github.com/rancher/rancher/issues/20209

PR #26018, opened 2020-03-13
bumping rancher-machine version

luthermonson:


luthermonson:
https://github.com/rancher/rancher/issues/20209

Issue #25944, opened 2020-03-10
CIS Scan - scheduled scans shouldn't remove manual scans


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Enable scheduled scans set the retention count to 1
- Execute 1 CIS scan manually
- Wait for the scheduled scan to get triggered.

**Result:**

The user will end up with one scheduled scan listed in the CIS Scan list page. The manually generated CIS scan gets deleted.

Manually executed CIS Scans shouldn't get touched by the scheduled scans.

**Other details that may be helpful:**

**Environment information**
- Rancher version: `2.4 master-head (03/10/2020)` commit id: 8ede17d10
- Installation option (single install/HA): single

sowmyav27:
**Verified on master-head - commit id: `330a4bed0`**

1. Deploy a cluster - 1 etcd, 1 control and 1 worker.
2. Enable CIS schedule scans and give cron time = 10 * * * *. (for example) and retention count = 2
3. Scheduled scan will run on the 10th minute that hour. 
4. Run few manual scans
5. Edit cluster change cron time = 12 * * * *. 
6. Scheduled scan will run on the 12th minute that hour. 
7. Edit cluster change cron time = 13 * * * *. 
8. Scan run in Step 3 will be deleted. And a new Scheduled scan will be generated for the cluster.
9. Manual scans do not get deleted.

Issue #25943, opened 2020-03-10
CIS Scan - scheduled scan keeps trying to execute after the cluster is deleted


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Create a cluster and enable scheduled CIS Scans
- Delete the cluster

**Result:**

After some time the Scheduled scan tries to get executed in the deleted cluster:

```
2020/03/10 23:23:03 [ERROR] ClusterScanController scheduledscan/c-vk7cw:23 * * * * [scheduledScanHandler:sync] failed with : scheduledScanHandler: sync: error fetching cluster c-vk7cw: cluster.management.cattle.io "c-vk7cw" not found
```

**Other details that may be helpful:**

**Environment information**
- Rancher version: `2.4 master-head (03/10/2020)` commit id: 8ede17d10
- Installation option (single install/HA): single

sowmyav27:
**Verified on master-head - commit id: `330a4bed0`**

- Deploy a cluster.
- Enable Scheduled CIS scan on the cluster. 
- Deleted the cluster
- Scheduled scan does not run on a deleted cluster.

PR #26029, opened 2020-03-14
cis bug fixes

leodotcloud:
https://github.com/rancher/rancher/issues/25943
https://github.com/rancher/rancher/issues/25944


pkg/controllers/management/cis/scheduled_scan.go
```diff
@@ -255,17 +255,25 @@ func cleanOldScans(
 	clusterScanClient v3.ClusterScanInterface,
 	clusterScanLister v3.ClusterScanLister,
 ) error {
+	logrus.Debugf("scheduledScanHandler: cleanOldScans: checking")
```
prachidamle: do we need this log? Does not seem to log any debug information

leodotcloud: will clean up.

Issue #25647, opened 2020-02-28
Ignore zero downtime constraints during atomic rollback


kinarashah:
With the introduction of zero downtime feature in 2.4, 
Upgrade on cluster will 
- consider only maxUnavailable nodes simultaneously
- do few prechecks, cordon/drain node, upgrade and uncordon the node back

We want to disable these extra steps during etcd restore action on the cluster. Restore flag is set on the rke config before restore operation (https://github.com/rancher/rancher/blob/master/pkg/api/customization/cluster/actions_etcd.go#L143) and gets reset back later (https://github.com/rancher/rancher/blob/master/pkg/controllers/management/clusterprovisioner/provisioner.go#L561) - so we should be able to rely on the same flag to check. 

https://github.com/rancher/rancher/issues/22232

mrajashree:
It looks like RKE already handles this, because ClusterUp calls zero downtime upgrade logic based on the clusterState.CurrentState https://github.com/rancher/rke/blob/master/cmd/up.go#L148 
`RestoreEtcdSnapshot` before calling ClusterUp sets the CurrentState to nil and writes this to the statefile https://github.com/rancher/rke/blob/master/cmd/etcd.go#L136-L137
So the `reconcileCluster` flag never gets set to true and DeployControlPlane and DeployWorkerPlane don't call the functions that carry out zero downtime upgrades

sowmyav27:
**Verified on latest master-head - comiit id: `888029fb0`**

**Steps:**

- Have a cluster - 1 etcd , 1 control plane and 5 worker nodes
- Max unavailable worker = 2, max Unavailable control = 1, drain = false
- take a snapshot
- Power down one worker plane node W1
- Upgrade k8s cluster from 1.16 to 1.17
- The cluster is stuck in upgrading state, as there is one worker node - powered down W1.
- other worker nodes get upgraded.
- restore to snapshot taken in Step 3.
- The powered down node is taken up for Upgrading, it is Cordoned. The node is then stuck in Cordoned state.
- one more node, worker node, is also stuck in cordoned state. 
- Cluster is stuck in Upgrading state 
- All worker nodes k8s version is 1.16
<img width="1281" alt="Screen Shot 2020-03-17 at 4 24 32 PM" src="https://user-images.githubusercontent.com/26032343/76910709-f934d500-686b-11ea-93f6-dd2e82efdcac.png">

sowmyav27:
**Verified on master-head - latest - commit id: `0582d678c`**

[...]

PR #26032, opened 2020-03-14
snapshot restore for worker nodes

kinarashah:
https://github.com/rancher/rancher/issues/25647

### DIFF
--- a/pkg/controllers/management/rkeworkerupgrader/upgrade.go
+++ b/pkg/controllers/management/rkeworkerupgrader/upgrade.go
@@ -4,16 +4,23 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/docker/docker/pkg/locker"
+	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
 	nodeserver "github.com/rancher/rancher/pkg/rkenodeconfigserver"
 	"github.com/rancher/rancher/pkg/systemaccount"
+	"github.com/rancher/rke/util"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 type upgradeHandler struct {
@@ -62,6 +69,12 @@ func (uh *upgradeHandler) Sync(key string, node *v3.Node) (runtime.Object, error
 		if cluster.DeletionTimestamp != nil || cluster.Status.AppliedSpec.RancherKubernetesEngineConfig == nil {
 			return nil, nil
 		}
+
+		if cluster.Annotations[clusterprovisioner.RkeRestoreAnnotation] == "true" {
+			// rke completed etcd snapshot restore, need to update plan for nodes so worker nodes reset to old rkeConfig
+			return node, uh.restore(cluster)
+		}
+
 		logrus.Infof("checking cluster [%s] for worker nodes upgrade", cluster.Name)
 
 		if ok, err := uh.toUpgradeCluster(cluster); err != nil {
@@ -90,6 +103,10 @@ func (uh *upgradeHandler) Sync(key string, node *v3.Node) (runtime.Object, error
 
 	if node.Status.NodePlan == nil {
 		return uh.updateNodePlan(node, cluster, true)
+	}
+
+	if v3.ClusterConditionUpdated.IsUnknown(cluster) || cluster.Annotations[clusterprovisioner.RkeRestoreAnnotation] == "true" {
+		return node, nil
 	}
 
 	if v3.ClusterConditionUpgraded.IsUnknown(cluster) {
@@ -273,7 +290,7 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string) e
 				continue
 			}
 
-			if err := uh.processNode(node, cluster, false); err != nil {
+			if err := uh.setNodePlan(node, cluster, false); err != nil {
 				return err
 			}
 
@@ -293,7 +310,7 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string) e
 			continue
 		}
 
-		if err := uh.processNode(node, cluster, true); err != nil {
+		if err := uh.setNodePlan(node, cluster, true); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


mrajashree:
Adding a separate approach for restoring rancher worker nodes, and not reusing the restore flag for rke config is a good idea, and using annotation sounds good to me. When resetting that annotation back to false, do we need retries for cluster update?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -62,6 +65,14 @@ func (uh *upgradeHandler) Sync(key string, node *v3.Node) (runtime.Object, error
 		if cluster.DeletionTimestamp != nil || cluster.Status.AppliedSpec.RancherKubernetesEngineConfig == nil {
 			return nil, nil
 		}
+
+		if cluster.Annotations[clusterprovisioner.RkeRestoreAnnotation] == "true" {
+			if err := uh.restore(cluster); err != nil {
```
StrongMonkey:
```suggestion
			return node, uh.restore(cluster)
```

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -369,6 +384,45 @@ func (uh *upgradeHandler) toUpgradeCluster(cluster *v3.Cluster) (bool, error) {
...
+func (uh *upgradeHandler) restore(cluster *v3.Cluster) error {
+	nodes, err := uh.nodeLister.List(cluster.Name, labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	// allowed to use goroutine?
+	var errgrp errgroup.Group
+	nodesQueue := util.GetObjectQueue(nodes)
+	for w := 0; w < 5; w++ {
```
StrongMonkey: What's the reason to use goroutinue here? Do you need to execute the logic fives times and wait for all go routinues be to finished?

kinarashah: each goroutine reads one node from the queue as and when it's done. 
ProcessNode gets the plan and updates the node so I didn't want to do that in a normal for loop, but also spinning one goroutine for each node seemed like a bit too much. So I am creating 5 goroutines and wait till entire queue is done

kinarashah:
@mrajashree Yeah, I'll add retries for cluster update

pkg/controllers/management/rkeworkerupgrader/upgrade.go
daxmc99: This is fundamental change. Should the handler still return the object on err?

kinarashah: this logic is under our own custom key enqueue, and has no object here to return  https://github.com/rancher/rancher/blob/78f5953f7a2e2cafcfe3bc290ddd9fa14e36a5e4/pkg/controllers/management/rkeworkerupgrader/upgrade.go#L62

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -369,6 +384,45 @@ func (uh *upgradeHandler) toUpgradeCluster(cluster *v3.Cluster) (bool, error) {
...
+		errgrp.Go(func() error {
+			var errList []error
+			for node := range nodesQueue {
+				node := node.(*v3.Node)
+				if node.Status.NodeConfig != nil && workerOnly(node.Status.NodeConfig.Role) {
+					if node.Status.NodePlan.Version != cluster.Status.NodeVersion {
+						if err := uh.processNode(node, cluster, false); err != nil {
+							errList = append(errList, err)
+						}
+						logrus.Infof("cluster [%s]: updated node [%s] for restore, plan version [%v]", cluster.Status.NodeVersion)
```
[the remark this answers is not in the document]
kinarashah: Yep build was failing 😸

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -374,6 +393,79 @@ func (uh *upgradeHandler) toUpgradeCluster(cluster *v3.Cluster) (bool, error) {
 	return false, nil
 }
 
+func (uh *upgradeHandler) restore(cluster *v3.Cluster) error {
+	nodes, err := uh.nodeLister.List(cluster.Name, labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	// allowed to use goroutine?
```
daxmc99: Drop before merge?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -374,6 +393,79 @@ func (uh *upgradeHandler) toUpgradeCluster(cluster *v3.Cluster) (bool, error) {
...
+	var errgrp errgroup.Group
+	nodesQueue := util.GetObjectQueue(nodes)
+	for w := 0; w < 5; w++ {
+		errgrp.Go(func() error {
+			var errList []error
+			for node := range nodesQueue {
+				node := node.(*v3.Node)
+				if node.Status.NodeConfig != nil && workerOnly(node.Status.NodeConfig.Role) {
+					if node.Status.NodePlan.Version != cluster.Status.NodeVersion {
+						if err := uh.processNode(node, cluster, false); err != nil {
```
daxmc99: I think a comment should be added to `processNode` or it should be renamed to better clarify what it does. 
My understanding of what it does is:
* deep copies the node
* Sets the status nodePlan.Plan field
* Sets the status nodePlan.Version field 

It is hard to understand what / why those need to be set

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -374,6 +393,79 @@ func (uh *upgradeHandler) toUpgradeCluster(cluster *v3.Cluster) (bool, error) {
...
+				if errors.IsConflict(err) {
+					return false, nil
+				}
+				return false, err
+			}
+		}
+		return true, nil
+	})
+
+}
```
daxmc99: This is going to try multiple times and it replicates what the norman framework would already do. Why do we need another backoff beyond what the norman framework provides?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
daxmc99: I'm not sure I follow... 
`return node, err` still works here

pkg/controllers/management/rkeworkerupgrader/upgrade.go
kinarashah: It's only going to retry only if the retries aren't sufficient from norman. Despite these retries being present in our framework, there are multiple times I've seen these conflict errors in logs

pkg/controllers/management/rkeworkerupgrader/upgrade.go
kinarashah: I had been using processNode during the upgrade logic consistently, as two stages of upgrade (preparing, then processing node) so missed that it has little context in this part of code 
- updated it to `setNodePlan`
- added a comment to restore function about why we need to update plan

pkg/controllers/management/rkeworkerupgrader/upgrade.go
kinarashah: not a logic change, the object is nil so changing code to `return node, uh.restore(cluster)`

rmweir:
Ik we already do it but is there something we can do instead of gating with conditions more?

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -374,6 +391,78 @@ func (uh *upgradeHandler) toUpgradeCluster(cluster *v3.Cluster) (bool, error) {
...
+		Steps:    7,
+	}
+
+	return wait.ExponentialBackoff(backoff, func() (bool, error) {
+		cluster, err := uh.clusters.Get(name, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		if cluster.Annotations[clusterprovisioner.RkeRestoreAnnotation] != "false" {
+			clusterCopy := cluster.DeepCopy()
```
daxmc99: This may perform a deep copy in a loop. 
This object has already been copied once on https://github.com/rancher/rancher/pull/26032/files#diff-a16c43ed026936b620e2fdb9ca6196b8R424 (line 424)

pkg/controllers/management/rkeworkerupgrader/upgrade.go
dramich: I think this is fine, if the cluster comes back through it's going to reprocess all nodes and on a large cluster will eat a large chunk of time and CPU.

pkg/controllers/management/rkeworkerupgrader/upgrade.go
dramich: No need to copy as the cluster obj is coming from the API

kinarashah:
@rmweir I understand, but probably not at this point since I've been using cluster/node conditions heavily for this feature

Issue #26024, opened 2020-03-13
K3S - First time save from upgrade, the status blue box should show immediately


maggieliu:
K3s Backend change:
* First time save from upgrade, the status blue box should show immediately

sangeethah:
Tested with `v2.4.0-rc5`
Import a k3s Cluster.
Upgrade k3 version.
Able to see the upgrade status message immediately which is as expected:

<img width="1263" alt="Screen Shot 2020-03-16 at 6 07 41 PM" src="https://user-images.githubusercontent.com/4266958/76812520-62f0a880-67b2-11ea-9ff8-e001517728d8.png">

But when the node are getting cordoned now we see the message relating to that node show up for a very short time and then it gets back to the standard message of your "cluster is upgrading" instead of sticking to show the node specific upgrade message.

<img width="1237" alt="Screen Shot 2020-03-16 at 6 08 06 PM" src="https://user-images.githubusercontent.com/4266958/76812592-95020a80-67b2-11ea-90dc-fed0078f2ff0.png">

Reopening this issue to fix this.

aaronRancher:
The bug fix is verified with Rancher server `rancher\rancher:master-2474-head` `version 888029fb0`

Update box shows up immediately upon cluster update. Box message changes with sequential nodes being updated.

PR #26043, opened 2020-03-16
Set upgrading status sooner

daxmc99:
Problem: Previously the updating state would be set after plans were
deployed into the downstream cluster which meant that the whole helm
install would need to occur before the cluster entered upgrading.

Solution: When an upgrade is determined to be necessary, the cluster
state is updating. This is before the system-upgrade controller is
deployed.

Addresses https://github.com/rancher/rancher/issues/26024

### DIFF
--- a/pkg/controllers/management/k3supgrade/deployPlans.go
+++ b/pkg/controllers/management/k3supgrade/deployPlans.go
@@ -156,6 +156,12 @@ func (h *handler) modifyClusterCondition(cluster *v3.Cluster, masterPlan, worker
 	// implement a simple state machine
 	// NotUpgraded => MasterPlanUpgrading || WorkerPlanUpgrading =>  NotUpgraded
 
+	if masterPlan.Name == "" && workerPlan.Name == "" {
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/k3supgrade/deployPlans.go
```diff
@@ -156,6 +156,13 @@ func (h *handler) modifyClusterCondition(cluster *v3.Cluster, masterPlan, worker
 	// implement a simple state machine
 	// NotUpgraded => MasterPlanUpgrading || WorkerPlanUpgrading =>  NotUpgraded
 
+	if masterPlan.Name == "" && workerPlan.Name == "" {
+		v3.ClusterConditionUpgraded.Unknown(cluster)
+		genericUpgradeMessage := fmt.Sprintf("cluster is being upgraded")
```
mrajashree: nit: we don't need this var and can directly pass the string to `v3.ClusterConditionUpgraded.Message`? rest LGTM

daxmc99: Sounds like a good change

Issue #25990, opened 2020-03-12
CIS Scan - Choose a profile dropdown is empty when trying to run as Cluster Owner


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create an RKE cluster
2. Create an **Standard User** in Rancher
3. Assign the user from Step 2 as the cluster owner of the cluster of Step 1
4. Logout from Admin and Login as the Cluster Owner added in Step 3
5. In the Cluster navigate to Run CIS Scan

**Result:**

The profile dropdown will be empty. As admin I can see it populated.

<img width="607" alt="Screen Shot 2020-03-12 at 1 46 47 PM" src="https://user-images.githubusercontent.com/27659/76566333-c8b6fa80-6469-11ea-84e3-036057d10f75.png">

**Additional Information:**

**/v3/cisconfigs** is empty for this Standard User.

As an admin this is the respond coming from /v3/cisconfigs:

```json
CisConfigParams": {
  "default": {
   "benchmarkVersion": "rke-cis-1.4"
  },
  "v1.15": {
   "benchmarkVersion": "rke-cis-1.4"
  },
  "v1.16": {
   "benchmarkVersion": "rke-cis-1.4"
  },
  "v1.17": {
   "benchmarkVersion": "rke-cis-1.4"
  },
  "v1.18": {
   "benchmarkVersion": "rke-cis-1.4"
  }
```

**Environment information**
- Rancher version: `2.4 master-head (03/12/2020)` commit id: d47b0ec4c
- Installation option (single install/HA): single

izaac:
Rancher version: `v2.4.0-rc8 (03/20/2020)`

I created `admin1` as admin user in rancher.
Login with admin1 and created a cluster.
Added `user1` and `user2` standard Rancher users.
Added user1 as Cluster Member to the cluster.
Added user2 as Cluster Owner to the cluster.

![Screen Shot 2020-03-20 at 10 20 10 AM](https://user-images.githubusercontent.com/27659/77189439-cae60e00-6a94-11ea-963b-9089ba7193e6.png)


user1 (Cluster Member) can't see the Run CIS scan option in dropdown or click in Run in Tools -> CIS Scans.
user2 (Cluster Owner) can click the Run CIS scan but the CIS Profile dropdown is empty. And neither from Tools -> CIS Scan

[...]

PR #26048, opened 2020-03-16
allow cluster-owner users to access cis info

leodotcloud:
addresses: https://github.com/rancher/rancher/issues/25990

Since UI fetches cis scan related information from the API and
the user didn't have privileges, empty data was returned, hence
the drop down was empty.

Fixed the problem by adding a controller that handles the RBAC
correctly for cluster-owner users.

### DIFF
--- a/pkg/controllers/user/cis/rbac.go
+++ b/pkg/controllers/user/cis/rbac.go
@@ -1,0 +1,134 @@
+package cis
+
+import (
+	"fmt"
+
+	"github.com/rancher/rancher/pkg/namespace"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	rbacv1 "github.com/rancher/types/apis/rbac.authorization.k8s.io/v1"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	cisReadOnlyRoleName = "cis-readonly"
+)
+
+type cisScanRBACHandler struct {
+	roles             rbacv1.RoleInterface
+	roleLister        rbacv1.RoleLister
+	roleBindings      rbacv1.RoleBindingInterface
+	roleBindingLister rbacv1.RoleBindingLister
+}
+
+func (rh *cisScanRBACHandler) Sync(key string, crtb *v3.ClusterRoleTemplateBinding) (runtime.Object, error) {
+	if crtb == nil {
+		return nil, nil
+	}
+	if crtb.DeletionTimestamp != nil {
+		return rh.handleDelete(key, crtb)
+	}
+	if crtb.RoleTemplateName != "cluster-owner" || crtb.ClusterName == "local" {
+		return nil, nil
+	}
+	logrus.Debugf("cisScanRBACHandler: Sync: crtb: %+v", crtb)
+
+	if err := rh.createRoles(); err != nil {
+		return nil, err
+	}
+	if err := rh.createRoleBindings(crtb); err != nil {
+		return nil, err
+	}
+	return nil, nil
+}
+
+func (rh *cisScanRBACHandler) handleDelete(_ string, crtb *v3.ClusterRoleTemplateBinding) (runtime.Object, error) {
+	logrus.Debugf("cisScanRBACHandler: handleDelete: crtb: %+v", crtb)
+	rbName, err := getRoleBindingName(crtb)
+	if err != nil {
+		return nil, err
+	}
+	err = rh.roleBindings.Delete(rbName, &metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+	logrus.Debugf("cisScanRBACHandler: handleDelete: deleting rb of user: %v for cluster: %v", crtb.UserName, crtb.ClusterName)
+	return nil, nil
+}
+
+func (rh *cisScanRBACHandler) createRoles() error {
+	_, err := rh.roleLister.Get(namespace.GlobalNamespace, cisReadOnlyRoleName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			rule := v1.PolicyRule{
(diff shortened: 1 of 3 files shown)


pkg/tunnelserver/tunnel.go
```diff
@@ -154,7 +154,7 @@ func (t *Authorizer) Authorize(req *http.Request) (*Client, bool, error) {
 
 func (t *Authorizer) getMachine(cluster *v3.Cluster, inNode *client.Node) (*v3.Node, error) {
 	machineName := machineName(inNode)
-	logrus.Debugf("getMachine: looking up machine [%s] in cluster [%s]", machineName, cluster.Name)
+	//logrus.Debugf("getMachine: looking up machine [%s] in cluster [%s]", machineName, cluster.Name)
```
cjellick: you probably want to uncomment this?

pkg/controllers/user/cis/rbac.go
```diff
@@ -0,0 +1,134 @@
...
+	roleBindings      rbacv1.RoleBindingInterface
+	roleBindingLister rbacv1.RoleBindingLister
+}
+
+func (rh *cisScanRBACHandler) Sync(key string, crtb *v3.ClusterRoleTemplateBinding) (runtime.Object, error) {
+	if crtb == nil {
+		return nil, nil
+	}
+	if crtb.DeletionTimestamp != nil {
+		return rh.handleDelete(key, crtb)
```
cjellick: We like to avoid finalizers as much as we can, but for RBAC, its important to guarantee that permissions are correct. You probably need to convert this to a lifecycle so that you have a finalizer and an explicit delete function (or tack it into an existing CRTB lifecycle.

pkg/controllers/user/cis/rbac.go
```diff
@@ -0,0 +1,134 @@
...
+					Name:      rbName,
+					Namespace: namespace.GlobalNamespace,
+				},
+				Subjects: []v1.Subject{subject},
+				RoleRef: v1.RoleRef{
+					Kind: "Role",
+					Name: cisReadOnlyRoleName,
+				},
+			}
+			_, err := rh.roleBindings.Create(rb)
```
daxmc99:
```suggestion
			_, err = rh.roleBindings.Create(rb)
```

pkg/controllers/user/cis/rbac.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (rh *cisScanRBACHandler) Sync(key string, crtb *v3.ClusterRoleTemplateBinding) (runtime.Object, error) {
+	if crtb == nil {
+		return nil, nil
+	}
+	if crtb.DeletionTimestamp != nil {
+		return rh.handleDelete(key, crtb)
+	}
+	if crtb.RoleTemplateName != "cluster-owner" || crtb.ClusterName == "local" {
```
cjellick: This will eventually bite us, but it might be good enough for now. You can't hard code the check to a role name. Operators can create their own RoleTemplates that are functionally equivalent to cluster-owner (or close enough).

cjellick: The correct solution would be to do something similar to
https://github.com/rancher/rancher/blob/4a72409ef5122d3902faee2b3e051799573ad179/pkg/controllers/management/auth/manager.go#L379-L382
which calculates whether you should get a certain permission based on the rules in the RoleTemplate, rather than the name of the role.

In fact, if you were to parameterize the namespace used in that method, you _might_  be able to re-use it.

pkg/controllers/user/cis/rbac.go
```diff
@@ -0,0 +1,134 @@
...
+func (rh *cisScanRBACHandler) createRoleBindings(crtb *v3.ClusterRoleTemplateBinding) error {
+	rbName, err := getRoleBindingName(crtb)
+	if err != nil {
+		return err
+	}
+	_, err = rh.roleBindingLister.Get(namespace.GlobalNamespace, rbName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			subject := v1.Subject{
+				Kind: "User",
```
cjellick: what if the CRTB is for a group rather than a user?

pkg/controllers/user/cis/rbac.go
```diff
@@ -0,0 +1,134 @@
+package cis
```
cjellick: You put this in the `user` package. A user handler is one that writes to the downstream cluster. This one does not. It should be in the `management` package.

pkg/controllers/user/cis/rbac.go
cjellick: @deniseschannon I want to leave the "call" for this one up to you as it is as much about time and effort as anything else. tl;dr is we are hard coding a check to a roleTemplate name `cluster-owner`, but users can define their own. If they have, then they wont get the permission that lets them see cisconfigs and cisbenchmarkversions.

Of the roles that people actually replace, I would expect cluster-owner to be fairly low on the list because it simply means you have full operational control over the cluster. I do recall someone wanting to create, basically, a "security read-only" role. That could be used to allow a sec team to view a cluster, but not modify it. With that use case, they wouldn't be able to see the two mentioned resources with how this is coded now.

deniseschannon: Basically only cluster owners can see these resources and there is no ability to add them to a custom role? @cjellick @leodotcloud

If that is the case, then this is fine for for 2.4.0 and we could add in the ability to manage cisconfigs/cisbenchmarkversions as a separate role in a later release.

pkg/tunnelserver/tunnel.go
leodotcloud: Yeah, I will skip this file. For some reason, it was printing insane number of debug messages in my setup.

pkg/controllers/user/cis/rbac.go
leodotcloud: 👍

pkg/controllers/user/cis/rbac.go
leodotcloud: Good point.

pkg/controllers/user/cis/rbac.go
leodotcloud: @deniseschannon that is correct. Only cluster owners can see these CIS info related resources.

pkg/controllers/user/cis/rbac.go
leodotcloud: Will modify.

Issue #25599, opened 2020-02-26
Accept maxUnavailable for control plane nodes during upgrade 


mrajashree:
maxUnavailable can be passed for control plane nodes too during zero downtime upgrade.
Default will be 1.

sowmyav27:
**Verified in `rancher:v2.4.0-rc1`**

- In cluster.yml, following is available:
```
    upgrade_strategy:
    drain: false
    max_unavailable_controlplane: '1'
    max_unavailable_worker: 10%
```

mrajashree:
Scenario 1: This is expected. Since the node is not labeled to be ignored, provisioning is incomplete because one node could not be contacted and we will try to contact it.

mrajashree:
Are there logs for scenario 2?

mrajashree:
In scenario 2, when upgrading k8s version through UI, maxUnavailable is being reset. This UI issue will address it.
For now editing through api will fix it
https://github.com/rancher/rancher/issues/25670

sowmyav27:
As per discussion with @kinarashah 
```
Have cluster - 1 etcd, 3 control plane and 1 worker node in k8s version - 1.15
edit cluster - cluster.yml - change max_unavailable_control plane = 2
power down one control plane node
edit cluster - cluster.yml - make sure max_unavailable_control plane = 2 and upgrade k8s version to 1.16
the cluster upgrades - etcd and control plane whcih are active - get upgraded.
But the upgrade fails with error: Provisioning incomplete, host(s) [<control-node>] skipped because they could not be contacted; Provisioning incomplete, host(s) [<control-node>] skipped because they could not be contacted 
```

**Expected Behavior:**
- The worker nodes should also get upgrade

kinarashah:
After offline discussion around feature requirements and user expectations, Rancher is going to assume any error from RKE as an error or a failure state (yes, even if number of failed nodes is less than maxUnavailableControlPlane). 

Rancher will keep retrying to reconcile (with some backoff). User has two options to resolve this: 
- fix whatever is preventing the control node from upgrading successfully
- add the ignore label to the node so RKE ignores the node

Worker nodes won't proceed in this use case.

sowmyav27:
**Verified on latest-master-head - commit id: `fa96c0c3d`**

**Steps:**

- Cluster spec - 1 etcd, 3 control planes, 1 worker nodes
- Power down one control plane node
- maxUnavailable_controlplane - 2
[...]

PR #26058, opened 2020-03-16
For WrapTransport in RKE use cluster dialer is node dialer fails

ibuildthecloud:


mrajashree:
Addresses: https://github.com/rancher/rancher/issues/25599

PR #26068, opened 2020-03-16
Check err not nil before calling clusterDialer

mrajashree:
https://github.com/rancher/rancher/issues/26071

### DIFF
--- a/pkg/ref/parse.go
+++ b/pkg/ref/parse.go
@@ -11,6 +11,9 @@ import (
 var NodeNotFound = "can not build dialer to"
 
 func IsNodeNotFound(err error) bool {
+	if err == nil {
+		return false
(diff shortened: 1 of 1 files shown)


pkg/rkedialerfactory/rke_dialer.go
```diff
@@ -68,10 +68,12 @@ func (t *RKEDialerFactory) WrapTransport(config *v3.RancherKubernetesEngineConfi
 						address = strings.Replace(address, ip, privateIP, 1)
 					}
 					conn, err := dialer(network, address)
-					if ref.IsNodeNotFound(err) {
```
ibuildthecloud: The IsNodeNotFound method should deal with nil errors.

PR #26086, opened 2020-03-17
Remain in upgrading state

daxmc99:
Problem: The cluster upgrading state would momentarily flicker during an
upgrade.
Solution: Once the upgrade state is entered, the only state that can
exit 'Upgrading' are states were worker or master plans are being
applied. This prevents the state flip flopping.

Addresses https://github.com/rancher/rancher/issues/26024

### DIFF
--- a/pkg/controllers/management/k3supgrade/deployPlans.go
+++ b/pkg/controllers/management/k3supgrade/deployPlans.go
@@ -154,12 +154,19 @@ func cmp(a, b planv1.Plan) bool {
 func (h *handler) modifyClusterCondition(cluster *v3.Cluster, masterPlan, workerPlan planv1.Plan) (*v3.Cluster, error) {
 
 	// implement a simple state machine
-	// NotUpgraded => MasterPlanUpgrading || WorkerPlanUpgrading =>  NotUpgraded
+	// UpgradedTrue => GenericUpgrading =>  MasterPlanUpgrading || WorkerPlanUpgrading =>  UpgradedTrue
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/k3supgrade/deployPlans.go
```diff
@@ -154,12 +154,19 @@ func cmp(a, b planv1.Plan) bool {
...
-	// NotUpgraded => MasterPlanUpgrading || WorkerPlanUpgrading =>  NotUpgraded
+	// UpgradedTrue => GenericUpgrading =>  MasterPlanUpgrading || WorkerPlanUpgrading =>  UpgradedTrue
 
 	if masterPlan.Name == "" && workerPlan.Name == "" {
-		v3.ClusterConditionUpgraded.Unknown(cluster)
-		v3.ClusterConditionUpgraded.Message(cluster, "cluster is being upgraded")
-		return h.clusterClient.Update(cluster)
+		// enter upgrading state
+		if v3.ClusterConditionUpgraded.IsTrue(cluster) {
+			v3.ClusterConditionUpgraded.Unknown(cluster)
```
cbron: Missing context here, why would we set upgraded condition to unknown if its currently true, because plans are missing ?

daxmc99: True is base, non-error case. Unknown triggers the Upgrading UI change. False indicates an error while upgrading. So it starts in True and will return to True if successful

Issue #25861, opened 2020-03-07
More than required number of nodes getting drained/cordoned and cluster gets stuck in upgrading state


kinarashah:
(not always reproducible) tested with rancher/rancher:master-head 

1. DO cluster with 11 worker nodes 
2. NodeUpgradeStrategy: drain=true, maxUnavailableWorker=50%
3. Upgraded from 1.16.7 to 1.17.3 

Bug: 
6 nodes got drained instead of 5! 

Workaround:
Uncordon or stop a draining node

Logs: 
```
2020/03/07 03:33:03 [INFO] cluster [c-nzj6q] worker-upgrade: updated node [m-tdlqg] to drain
2020/03/07 03:33:03 [INFO] Draining node kinarasaworkernode4 in c-nzj6q with flags --delete-local-data=false --force=false --grace-period=30 --ignore-daemonsets=true --timeout=60s
2020/03/07 03:33:03 [INFO] cluster [c-nzj6q] worker-upgrade: updated node [m-4kvc2] to drain
2020/03/07 03:33:03 [INFO] cluster [c-nzj6q] worker-upgrade: updated node [m-hnvdl] to drain
2020/03/07 03:33:03 [INFO] Draining node kinarasaworkernode7 in c-nzj6q with flags --delete-local-data=false --force=false --grace-period=30 --ignore-daemonsets=true --timeout=60s
2020/03/07 03:33:03 [INFO] cluster [c-nzj6q] worker-upgrade: updated node [m-j2zjz] to drain
2020/03/07 03:33:03 [INFO] cluster [c-nzj6q] worker-upgrade: updated node [m-mwkh7] to drain
2020/03/07 03:33:03 [INFO] Draining node kinarasaworkernode5 in c-nzj6q with flags --delete-local-data=false --force=false --grace-period=30 --ignore-daemonsets=true --timeout=60s
2020/03/07 03:33:03 [INFO] cluster [c-nzj6q] worker-upgrade: updated node [m-9sglz] to drain
2020/03/07 03:33:03 [INFO] Draining node kinarasaworkernode10 in c-nzj6q with flags --delete-local-data=false --force=false --grace-period=30 --ignore-daemonsets=true --timeout=60s
2020/03/07 03:33:03 [INFO] Draining node kinarasaworkernode2 in c-nzj6q with flags --delete-local-data=false --force=false --grace-period=30 --ignore-daemonsets=true --timeout=60s
[...]

PR #26096, opened 2020-03-18
use sorted list to store node upgrade status

kinarashah:
Problem:
Lister might not always give latest updated node object.
If nodes are picked for cordon/drain out of order,
we might cordon/drain extra nodes, violating the maxUnavailable.

Solution:
Using a sorted list ensures we always pick the same node.

https://github.com/rancher/rancher/issues/25861

### DIFF
--- a/pkg/controllers/management/rkeworkerupgrader/node.go
+++ b/pkg/controllers/management/rkeworkerupgrader/node.go
@@ -2,6 +2,7 @@ package rkeworkerupgrader
 
 import (
 	"fmt"
+	"sort"
 
 	nodehelper "github.com/rancher/rancher/pkg/node"
 	nodeserver "github.com/rancher/rancher/pkg/rkenodeconfigserver"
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -260,17 +260,19 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string) e
...
 
-	maxAllowed, err := CalculateMaxUnavailable(upgradeStrategy.MaxUnavailableWorker, filtered)
+	maxAllowed, err := CalculateMaxUnavailable(upgradeStrategy.MaxUnavailableWorker, status.filtered)
 	if err != nil {
 		return err
 	}
 
-	logrus.Debugf("cluster [%s] worker-upgrade: workerNodeInfo: nodes %v maxAllowed %v upgrading %v notReady %v "+
-		"toProcess %v toPrepare %v done %v", cluster.Name, filtered, maxAllowed, upgrading, len(notReadyMap), keys(toProcessMap), keys(toPrepareMap), keys(upgradedMap))
+	logrus.Infof("cluster [%s] worker-upgrade: workerNodeInfo: nodes %v maxAllowed %v upgrading %v notReady %v "+
```
dramich: This can't be at Info, it's going to explode logs in a large cluster. Even then I would say this might be a bad log message, imagine 10k nodes in this message.

pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -474,6 +480,14 @@ func (uh *upgradeHandler) retryClusterUpdate(name string) error {
 
 }
 
+func keys(nodes []*v3.Node) []string {
+	keys := []string{}
```
dramich: Create this with the exact length you need so the underlying array won't need to be grown, this will mainly affect large clusters.

dramich: `keys := make([]string, len(nodes))`

pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -10,6 +11,27 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type UpgradeStatus struct {
```
dramich: This is a great change

dramich: But it doesn't need to be exported

pkg/controllers/management/rkeworkerupgrader/upgrade.go
kinarashah: Oh oops!! moved it to info just for testing, forgot changing back, thank you

PR #26116, opened 2020-03-18
Add all k3s images to airgap images

rmweir:
We usually download files used for things like image export in dapper, but in this scenarios we do not know what k3s version are relevant until data.json is parsed. This could be done in bash but its already being done in export/main.go.

### DIFF
--- a/pkg/image/export/main.go
+++ b/pkg/image/export/main.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/rancher/rancher/pkg/controllers/management/k3supgrade"
@@ -12,6 +14,7 @@ import (
 	img "github.com/rancher/rancher/pkg/image"
 	"github.com/rancher/types/image"
 	"github.com/rancher/types/kdm"
+	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/image/export/main.go
```diff
@@ -265,13 +269,69 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{})
...
+		}
+
+		supportingImages := strings.Split(images, "\n")
+		if supportingImages[len(supportingImages)-1] == "" {
+			supportingImages = supportingImages[:len(supportingImages)-1]
+		}
+
+		for _, imageName := range supportingImages {
+			// ignore original registry and repo name
+			imageParts := strings.Split(imageName, "/")
```
cjellick: error out if length isnt 3

cjellick: if len() != 3{
logrus.Fatalf
}

pkg/image/export/main.go
```diff
@@ -265,13 +269,69 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{})
...
+			aliasedImageName := fmt.Sprintf(imageFormat, imageParts[1]+"-"+imageParts[2])
```
cjellick: dont manipulate image names here. just remove registry if it is `docker.io` (remove prefix)

cjellick:
https://github.com/rancher/rancher/issues/26200

Issue #26119, opened 2020-03-19
Clusters become unavailable after Rancher upgrade from 2.2.10


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Install Rancher `v2.2.10`
Provision a custom cluster
Upgrade Rancher to `v2.4.0-rc6`

**Result:**
Clusters become stuck in unavailable
<img width="1019" alt="Screen Shot 2020-03-18 at 6 35 11 PM" src="https://user-images.githubusercontent.com/57409052/77022684-8862eb00-6947-11ea-87b6-8a48b5d6dd2c.png">

**Other details that may be helpful:**
The old rancher-agent has exited with no new agent being created
```
$ docker ps -a | grep agent
42860439dca2        422cade64eb5                                      "run.sh"                 About an hour ago   Up About an hour                                   k8s_agent_cattle-node-agent-82tm6_cattle-system_c3f2b84d-86fd-48ca-9196-56c9f0cd954f_0
a7520208bf26        rancher/pause:3.1                                 "/pause"                 About an hour ago   Up About an hour                                   k8s_POD_cattle-node-agent-82tm6_cattle-system_c3f2b84d-86fd-48ca-9196-56c9f0cd954f_0
688692c3cbb1        rancher/rancher-agent:v2.2.10                     "run.sh -- share-roo…"   About an hour ago   Exited (0) About an hour ago                       share-mnt
```
Logs of agent:
```
time="2020-03-19T01:45:13Z" level=error msg="Failed to connect to proxy" error="x509: certificate is valid for 127.0.0.1, 172.17.0.2, not <Rancher ip>"
time="2020-03-19T01:45:13Z" level=error msg="Failed to connect to proxy" error="x509: certificate is valid for 127.0.0.1, 172.17.0.2, not <Rancher ip>"
time="2020-03-19T01:45:23Z" level=info msg="Connecting to wss://<Rancher ip>/v3/connect with token 86hh2zdqqqgf7s6qqpspdvhlhjqc56rlbkkpwclhxblx7g25qg7k2j"
time="2020-03-19T01:45:23Z" level=info msg="Connecting to proxy" url="wss://<Rancher ip>/v3/connect"
```
**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `v2.4.0-rc6`
- Installation option (single install/HA):
single, self-signed certs

[...]

PR #26127, opened 2020-03-19
On upgrade from < 2.4 ensure we keep the known SANs

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/26119

Issue #26089, opened 2020-03-17
modifying altermanager.yaml injects empty proxy_url option


flrichar:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- enable monitoring
- modify secrets for alertmanager.yaml
- remove proxy_url line

**Result:**
- proxy_url comes back with null value "" and is parsed incorrectly

**Other details that may be helpful:**
user does have a proxy but does not need it for internal custom webhook.  the monitoring will work for a while before the controller replaces the proxy_url line, breaking the monitoring.

**Environment information**
- Rancher version v2.3.5
- Installation option Single

**Cluster information**
- Cluster type: Custom
- Machine type: VM
- Kubernetes version 1.15
- Docker version 18.09.3

loganhz:
Workaround 1 : 

- Remove the webhook and add it again without proxy url input

Workaround 2: 

- Click View API for the broken notifier. 
- Click "Edit" and remove the proxyurl part in webhookConfig. 
- Click "Show request". 
- Copy the `curl` command.
- Append `?_replace=true` in the curl command
- Send the command via curl

![image](https://user-images.githubusercontent.com/21168270/76917576-cee02780-68fe-11ea-8586-4759e225c5cf.png)
![image](https://user-images.githubusercontent.com/21168270/76917722-41510780-68ff-11ea-8b69-b3ed5f54e5ce.png)

gmolkvk:
I have also found that activating the Proxy URL field will insert an empty string into the webhookConfig.proxy_url, which will end up in the AlertManager. 

By extension, this will cause "connection refused" errors in the AlertManager on triggered Webhook alerts, because "" (empty string) is not a valid hostname for a proxy.

jiaqiluo:
Some observations on around notifiers in v2.3.5
- we can edit an existing notifier to change the value of the proxy URL to another valid value or an empty string, but cannot remove the proxy URL once it was set
- if we enter spaces on the proxy URL filed when creating a notifier, the value will be set to an empty string which can be seen when viewing the notifier in API.

<img width="675" alt="Screen Shot 2020-04-06 at 4 08 58 PM" src="https://user-images.githubusercontent.com/6218999/78613552-0157b300-7821-11ea-93e0-a17cfd435370.png">

jiaqiluo:
The bug fix is validated in Rancher:v2.4-master 0865a2cc9 

[...]

PR #26129, opened 2020-03-19
Fix the issue that users cannot remove proxyUrl of notifier

GGGitBoy:
**Problem:**

After users set a proxy url for a notifier, they can not really remove it in alert manager config. 
Rancher will always set an empty proxyUrl in alert manger config for the notifier. So alert manager cannot send out alerts due to the empty proxy url setting.

**Solution:**

So we need to check if  it's empty.

**Related Issue:**

https://github.com/rancher/rancher/issues/26089

**Related PR:**

https://github.com/rancher/ui/pull/3854

### DIFF
--- a/pkg/controllers/user/alert/configsyncer/configsyncer.go
+++ b/pkg/controllers/user/alert/configsyncer/configsyncer.go
@@ -16,6 +16,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/user/alert/deployer"
 	"github.com/rancher/rancher/pkg/controllers/user/alert/manager"
 	monitorutil "github.com/rancher/rancher/pkg/monitoring"
+	notifierutil "github.com/rancher/rancher/pkg/notifiers"
 	"github.com/rancher/rancher/pkg/project"
 	"github.com/rancher/rancher/pkg/ref"
 	v1 "github.com/rancher/types/apis/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/controllers/user/alert/configsyncer/configsyncer.go
```diff
@@ -556,7 +556,7 @@ func (d *ConfigSyncer) addRecipients(notifiers []*v3.Notifier, receiver *alertco
 					slack.Channel = r.Recipient
 				}
 
-				if notifier.Spec.SlackConfig.HTTPClientConfig != nil {
+				if notifier.Spec.SlackConfig.HTTPClientConfig != nil && notifier.Spec.SlackConfig.HTTPClientConfig.ProxyURL != "" {
```
gitlawr: Given that this check happens many times I think we can have a utility function for it.
e.g.:
```
func IsHTTPClientConfigSet(cfg *v3.HTTPClientConfig) bool{}
```

GGGitBoy:
Updated

pkg/notifiers/sender.go
```diff
@@ -516,7 +516,7 @@ func NewClientFromConfig(cfg *v3.HTTPClientConfig, dialer dialer.Dialer) (*http.
 		Timeout: time.Second * 10,
 	}
 
-	if cfg != nil {
+	if cfg != nil && cfg.ProxyURL != "" {
```
gitlawr: The utility function can be used here as well. We can define it in `pkg/notifiers` and use it in `pkg/controllers`

pkg/notifiers/sender_test.go
```diff
@@ -0,0 +1,29 @@
+package notifiers
+
+import (
+	"testing"
+
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsHTTPClientConfigSet(t *testing.T) {
```
dramich, line 10: The tests should be setup with a list of cases to process, this helps make the tests easier to read and add additional tests later. An example can be seen here https://github.com/rancher/rancher/blob/master/pkg/rbac/common_test.go

GGGitBoy:
Updated

deniseschannon:
@GGGitBoy Can you make a backport PR into the release/v2.4 branch?

GGGitBoy:
> @GGGitBoy Can you make a backport PR into the release/v2.4 branch?

release/v2.4 branch PR is here
https://github.com/rancher/rancher/pull/26393

Issue #26108, opened 2020-03-18
CIS Scan - Scans not running on a cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

1. Deploy a cluster - 1 etcd, 1 control and 1 worker.
2. Enable CIS schedule scans and give cron time = 10 * * * *. (for example) and retention count = 2
3. Scheduled scan will run on the 10th minute that hour. 
4. Run few manual scans
5. Edit cluster change cron time = 12 * * * *. 
6. Scheduled scan will run on the 12th minute that hour. 
7. Edit cluster change cron time = 13 * * * *. 
8. Scan run in Step 3 will be deleted. And a new Scheduled scan will be generated for the cluster.
9. Manual scans do not get deleted
10. Do Step 4, 5 and 6 a couple of times.

**Expected Result:**
Scheduled scan and run scan should be run on the cluster

**Actual Result:**

- Scheduled scan and run scan are not on the cluster. Error seen while Run scan is clicked from UI - `CIS scan already running on cluster` 
- No workloads/pods related to scan running in System project -- Security scan namespace
- Logs:

```
2020/03/18 17:41:37 [INFO] scheduledScanHandler: clusterSync: cron schedule modified for cluster c-8q8rv: 42 * * * *
2020/03/18 17:41:37 [DEBUG] scheduledScanHandler: scheduleScan: cluster c-8q8rv now: 2020-03-18 17:41:37.201276888 +0000 UTC m=+7113.246273901 setting nextScanAt: 2020-03-18 17:42:00 +0000 UTC
2020/03/18 17:41:37 [DEBUG] Checking and deleting unused certificates for cluster c-8q8rv
2020/03/18 17:41:53 [DEBUG] machineName: returning [m-031e08b51fd9] for node with RequestedHostname [ip-172-31-8-218]
2020/03/18 17:41:53 [DEBUG] getMachine: looking up machine [m-031e08b51fd9] in cluster [c-dxlgg]
2020/03/18 17:42:00 [ERROR] ClusterScanController scheduledscan/c-8q8rv:42 * * * * [scheduledScanHandler:sync] failed with : CIS scan already running on cluster
2020/03/18 17:42:00 [ERROR] ClusterScanController scheduledscan/c-8q8rv:42 * * * * [scheduledScanHandler:sync] failed with : CIS scan already running on cluster
[...]

PR #26135, opened 2020-03-19
Move annotation to status field

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/26108.

Issue #25996, opened 2020-03-13
CIS Scan - Rancher should prevent running CIS Scans in clusters with k8s version lower than 1.15


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Cluster after upgrading from 2.2.9 or 2.3.x with k8s lower than 1.15
- `runSecurityScan` from API or Run CIS Scan from UI. 

**Result:**

The CIS Scan gets executed even if the k8s version is lower than 1.15

**Expectation:**

Prevent the user from running scans in clusters with k8s versions for CIS 1.4 Profile

**Environment information**
- Rancher version: `2.4 master-head (03/12/2020)` commit id: d47b0ec4c
- Installation option (single install/HA): single

izaac:
Rancher version: `v2.4 master-head (03/20/2020)` commit id: 3e6e9ddee Upgraded from Rancher v2.2.10.

I tried this with three clusters versions upgraded from v2.2.10:

- v1.13.12
- v1.14.8
- v1.15.5

Versions lower than 1.15.x shown the following message when tried to run the CIS Scan:

![Screen Shot 2020-03-20 at 5 12 42 PM](https://user-images.githubusercontent.com/27659/77214930-a6a72300-6ace-11ea-8313-88d10173ced5.png)

But I was able to trigger a CIS Scan from the API, but it just got stuck in the `Creating` state:

![Screen Shot 2020-03-20 at 5 19 39 PM](https://user-images.githubusercontent.com/27659/77214983-ff76bb80-6ace-11ea-80f2-722a0405e3c4.png)


The difference we noticed when doing from UI and API was the UI is sending the Action POST as:

```json
{
	"failuresOnly": false,
	"skip": null,
	"profile": "permissive",
	"overrideBenchmarkVersion": "rke-cis-1.4"
}
```

And the API action (with defaults) as:

```json
{

    "debugMaster": false,
    "debugWorker": false,
    "overrideBenchmarkVersion": null,
    "overrideSkip": [ ],
    "profile": "permissive"
}
```

I was able to run scans from the UI and API for k8s version 1.15, 1.16 and 1.17. Clusters with k8s 1.16 and 1.17 were created after upgrade.

Reopening because I was able to trigger scans from the API for clusters with k8s versions 1.13 and 1.14

izaac:
[...]

PR #26137, opened 2020-03-19
verify min k8s version before running scan

leodotcloud:
addresses: https://github.com/rancher/rancher/issues/25996

Minimum k8s version required for a benchmark version is available in the CRD. Added checks to compare that with the cluster version when the scan is initiated.

### DIFF
--- a/pkg/controllers/user/cis/clusterScanHandler.go
+++ b/pkg/controllers/user/cis/clusterScanHandler.go
@@ -9,10 +9,8 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/rancher/rancher/pkg/app/utils"
-	"github.com/rancher/rancher/pkg/controllers/management/kontainerdrivermetadata"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/systemaccount"
-	"github.com/rancher/rke/util"
 	"github.com/rancher/security-scan/pkg/kb-summarizer/report"
 	appsv1 "github.com/rancher/types/apis/apps/v1"
 	rcorev1 "github.com/rancher/types/apis/core/v1"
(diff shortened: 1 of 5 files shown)


pkg/controllers/user/cis/utils.go
```diff
@@ -107,3 +110,61 @@ func ValidateClusterBeforeLaunchingScan(cluster *v3.Cluster) error {
...
+	if err != nil {
+		return "", false, fmt.Errorf("cisScanHandler: error fetching benchmark version info %v: %v",
+			bv, err)
+	}
+	benchmarkK8sVersionSemVer, err := util.StrToSemVer(ConvertToSemVerStr(benchmarkInfo.MinKubernetesVersion))
+	if err != nil {
+		return "", false, fmt.Errorf("cisScanHandler: error getting sem version for benchmark k8s version %v: %v",
+			benchmarkInfo.MinKubernetesVersion, err)
+	}
+	clusterK8sSemVerStr := ConvertToSemVerStr(shortK8sVersion)
```
prachidamle: What if the benchmarkInfo is derived using "default" when kontainerdrivermetadata.GetCisConfigParams("shortK8sVersion"...) has failed? In that case  should we compare with "default"

pkg/controllers/user/cis/utils.go
```diff
@@ -107,3 +110,61 @@ func ValidateClusterBeforeLaunchingScan(cluster *v3.Cluster) error {
...
+	clusterK8sSemVer, err := util.StrToSemVer(clusterK8sSemVerStr)
+	if err != nil {
+		return "", false, fmt.Errorf("cisScanHandler: error getting sem version for cluster k8s version %v: %v",
+			clusterK8sSemVer, err)
+	}
+	logrus.Debugf("cisScanHandler: checking if cluster version %v is less than min version: %v",
+		clusterK8sSemVerStr, benchmarkInfo.MinKubernetesVersion)
+	if clusterK8sSemVer.LessThan(*benchmarkK8sVersionSemVer) {
+		return "", false, fmt.Errorf("minimum k8s version %v needed for running cis scan", benchmarkInfo.MinKubernetesVersion)
+	}
```
prachidamle: Can we to put the semver and version comparison in a separate method.

pkg/controllers/user/cis/utils.go
leodotcloud: Only CisConfig has "default" which would return a benchmark version (along with min k8s version). There is no "default" in benchmark. So I think we are good here. Let me know if you still have concerns.

pkg/controllers/user/cis/utils.go
leodotcloud: 👍

pkg/controllers/user/cis/utils.go
```diff
@@ -107,3 +110,68 @@ func ValidateClusterBeforeLaunchingScan(cluster *v3.Cluster) error {
...
+		return "", false, fmt.Errorf("cisScanHandler: error fetching benchmark version info %v: %v",
+			bv, err)
+	}
+	if err := isClusterVersionSufficient(shortK8sVersion, benchmarkInfo.MinKubernetesVersion); err != nil {
+		return "", false, err
+	}
+	return bv, benchmarkInfo.Managed, nil
+}
+
+func isClusterVersionSufficient(shortClusterK8SVersion, benchmarkMinK8SVersion string) error {
```
StrongMonkey: Would it be better if this returns (bool, error). Then we can tell if the version comparison succeeded or there are errors parsing versions?

pkg/controllers/user/cis/utils.go
```diff
@@ -107,3 +110,68 @@ func ValidateClusterBeforeLaunchingScan(cluster *v3.Cluster) error {
 	}
 	return nil
 }
+
```
StrongMonkey: since this is a util method, would you mind adding a comment on the top explain this function?

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -111,40 +109,24 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
...
-			csh.cisBenchmarkVersionLister,
-			csh.cisBenchmarkVersionClient,
+		bv, bvManaged, err := GetBenchmarkVersionToUse(overrideBenchmarkVersion, currentK8sVersion,
+			csh.cisConfigLister, csh.cisConfigClient,
+			csh.cisBenchmarkVersionLister, csh.cisBenchmarkVersionClient,
 		)
 		if err != nil {
-			return nil, fmt.Errorf("cisScanHandler: Create: error fetching benchmark version info %v: %v",
-				cisConfigParams.BenchmarkVersion, err)
+			return cs, err
```
StrongMonkey: When version comparison fails, does this keep returning error? Is there any gated logic to prevent invalid cluster scan being re-enqueued into controller?

pkg/controllers/user/cis/utils.go
leodotcloud: will make the change.

pkg/controllers/user/cis/utils.go
leodotcloud: Sure.

pkg/controllers/user/cis/clusterScanHandler.go
leodotcloud: It's gated at the create time itself to reject the scan.

leodotcloud: actions_cis.go

prachidamle: This means we will never remove any benchmarkversion data from kdm - i that correct @leodotcloud? because if we change kdm, there will be clusterscans that might become invalid and get re-enqueued?

prachidamle: Thus even the "MinKubernetesVersion" in the"default" entry in kdm should never change?

deniseschannon: Correct, we will ever remove any benchmark data versions from kdm. Typically anything in kdm, once it's released, we can never remove it. 

But I cannot guarantee that the min k8s versions would never change, but most likely not.

leodotcloud: As discussed offline, older k8s versions will pick the "default" entry and use the corresponding benchmark version which will have a min k8s version which is 1.15. The default is mainly to allow newer k8s versions to use the latest available benchmark version. The current PR enforces the min k8s version so that older k8s versions don't allow the scan.

PR #26138, opened 2020-03-19
Replace `+` with`-` for name compliance

brendarearden:
Initially replaced `+` with `_` but should be `-` to be valid kubernetes name

**Related PR:** 
https://github.com/rancher/rancher/pull/25920


dramich:
Can we update the catalog tests to catch this case?

Issue #26028, opened 2020-03-14
CIS Scan does not run on a cluster with powered off nodes


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster - 3 etcd, 2 control and 4 worker nodes
- Power off one worker node.
- Worker node status - `Unavailable` in the cluster
- Go to cluster --> Tools --> CIS scan --> Run scan --> Select profile
- The Scan is seen running.
- Report is not generated
- `security-scan-runner` Pod is seen running in System project

**Expected Result:**
The report should be generated for the cluster

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head latest - commit id: `547c8dd3d`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke do
- Kubernetes version (use `kubectl version`):

```
v1.16.8
```

sowmyav27:
Was able to reproduce this issue when a control node was powered down.
Scheduled scan - failed to run on the cluster.

sowmyav27:
**Verified on master-head - commit id: `546b039e4`**

- Deploy a cluster - 2 control, 3 etcd, 3 worker nodes.
- Power off worker node W1
- Run CIS scan. Verified the scan report does not report W1 in the list of nodes.
- Power on worker node W1
- Run CIS scan. Verified the scan reports  W1 in the list of nodes.

Similar tests wrt to powering off/on control plane and etcd were verified. 
Verified the same for schedule scan also

StrongMonkey:
CIS can't be run if the following is broken:
1. Metric server is not running. There is a known issue that helm will not work if metric api is not available. CIS app will fail to deploy. https://github.com/helm/helm/issues/6361
2. Coredns is not running. CIS relies on DNS internally so if dns is broken CIS run can't be finished.

sowmyav27:
**Verified on latest master-head - commit id: `371ec137f`**

- Cluster spec - 1 etcd, 1 control plane and 5 worker nodes.
- Power off W5, cre dns and metrics server are NOT deployed on this worker node.
- Run CIS scan on the cluster
- The scan runs successfully. Report is generated.
- Report does not have the powered off node in its list of nodes.

```
"nodes":{
   "etcd":[
[...]

PR #26168, opened 2020-03-20
Add tolerations to CIS 

StrongMonkey:


StrongMonkey:
Related chart changes: https://github.com/rancher/system-charts/pull/190. Fix https://github.com/rancher/rancher/issues/26028.

prachidamle:
After offline discussion with @StrongMonkey , here is outline of the fix:
- we need to make sure kubebench gets scheduled to  only active nodes so it will not block generating report.

So the idea of the fix is:
- read all nodes and gather the taints; skip any node.kubernetes.io taints. This will skip any nodes having `node.kubernetes.io/not-ready` taint which is added by kubernetes when node goes down.
- build matching tolerations and pass them while launching KB app
- It will ensure  that we look at only active nodes and KB launches on them since tolerations match

pkg/controllers/user/cis/clusterScanHandler.go
```diff
@@ -380,6 +389,29 @@ func (csh *cisScanHandler) deployApp(appInfo *appInfo) error {
 	return nil
 }
 
+// collectTaints collect all taints on kubernetes nodes except node.kubernetes.io/*
+func (csh *cisScanHandler) collectTaints() (map[string]string, error) {
+	r := map[string]string{}
+	selector := labels.NewSelector()
+	nodes, err := csh.nodeLister.List("", selector)
```
prachidamle: Here we are using Lister to list all nodes; however we are relying on nodelister to be updated when node goes down - so there could be some delay for lister cache getting updated?
What if CIS scan gets scheduled when lister is not yet updated? KB will get launched to the nonactive node and report generation will get stuck. Can the user cancel the CIS scan in this case?

prachidamle: Confirmed with @sowmyav27 and @StrongMonkey that a user can delete the current running scan if it gets stuck in the reporting state because of the above timing.

prachidamle:
Changes look good to me; build is failing though

Issue #26148, opened 2020-03-19
Upgrade message shown during upgrade k3s cluster doesn't show all nodes if concurrency is more than 1


khushboo-rancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Import k3s cluster (1 control, 6 workers) - version 1.16.3
2. Edit cluster and upgrade it to 1.16.7 with concurrency 3

**Result:**
While upgrading 3 worker nodes simultaneously, upgrade message shows only 1 node.
<img width="1342" alt="Screen Shot 2020-03-19 at 3 43 32 PM" src="https://user-images.githubusercontent.com/60111667/77124182-6b89ee80-69ff-11ea-8eef-b31941a4e888.png">

**Expectation**
While upgrading 3 worker nodes simultaneously, upgrade message should show all nodes.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-2486-head- cf6d0c979
- Installation option (single install/HA): Single

khushboo-rancher:
Tested on Rancher version master-2523-head- bd2eb0cf9

Upgrade message shows all the nodes which are getting upgraded. 
Tested for concurrency 1 and concurrency 3.
Sample 
```
This cluster is currently Upgrading.
Worker node [ip-xx-xx-xx-xx, ip-yy-yy-yy, ip-zz-zz-zz-zz] being upgraded.
```

PR #26177, opened 2020-03-21
Report every node that is being upgraded

daxmc99:
Previously, the UI would only list one node when it was being upgraded.
Now the UI will show the top c nodes of the applying plan where c is
equal to the concurrency

Addresses https://github.com/rancher/rancher/issues/26148

### DIFF
--- a/pkg/controllers/management/k3supgrade/deployPlans.go
+++ b/pkg/controllers/management/k3supgrade/deployPlans.go
@@ -3,6 +3,7 @@ package k3supgrade
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	planv1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
 	planClientset "github.com/rancher/system-upgrade-controller/pkg/generated/clientset/versioned/typed/upgrade.cattle.io/v1"
@@ -10,6 +11,8 @@ import (
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const MaxDisplayNodes = 10
 
 // deployPlans creates a master and worker plan in the downstream cluster to instrument
 // the system-upgrade-controller in the downstream cluster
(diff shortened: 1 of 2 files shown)


daxmc99:
@dweomer This assumes the if a concurrency of 2 is set on a plan then the `applying` list node[0] and node[1] are those nodes where a plan is being applied

pkg/controllers/management/k3supgrade/deployPlans.go
```diff
@@ -190,3 +195,28 @@ func (h *handler) modifyClusterCondition(cluster *v3.Cluster, masterPlan, worker
 	return h.clusterClient.Update(cluster)
 
 }
+
+func upgradingMessage(concurrency int, nodes []string) string {
```
dweomer: This code looks fine but I don't understand the need for it if the results are equivalent to `strings.Join(nodes, ", ")`. However I do see value in this function if applies a truncating criteria (e.g. number of applying nodes, or length of the output string) with a conditionally appended ellipses to indicate truncation. For a large `plan.spec.concurrency` values it seems a bit much to burden a condition message (or is it reason?) with so much text.

Additionally, this message is rendered as-is in the UI correct? Something to know about the `plan.status.applying` slice is that the values therein are sorted in lexical order and not the order in which they were added to the list (which is essentially randomized per `plan.status.latestHash`). This means that for moderate to large `plan.spec.concurrency` values one is likely to see a bit more hopping around visually as nodes come in and out rather than what one might expect (e.g. an approximate fifo-like slide from right to left).

dweomer: Something that I missed initially is that the passed `concurrency` is a truncating criteria but its value is always sourced from the `plan.spec.concurrency` which should rarely if ever be less than the length of the `plan.status.applying` by design.

pkg/controllers/management/k3supgrade/deployPlans.go
```diff
@@ -190,3 +197,15 @@ func (h *handler) modifyClusterCondition(cluster *v3.Cluster, masterPlan, worker
...
+	// concurrency max can be very large
+	if concurrency > len(nodes) {
+		concurrency = len(nodes)
+	}
+	if concurrency > MAX_DISPLAY_NODES {
```
daxmc99: This limits the current number of nodes being upgraded in the UI to `10`

PR #26193, opened 2020-03-22
Add conditions for sidecar auditLog.destination in deployment.yaml

dkeightley:
Create the sidecar container only when `auditLog.destination: sidecar` is passed as a value without a `auditLog.hostPath`.

This achieves the outcome expected as per our documentation, where destination `hostPath` and `sidecar` values are mutually exclusive.

A user can still make these mutually inclusive by passing a `hostPath`, for eg:

```yaml
auditLog:
  destination: sidecar
  hostPath: /var/log/testing/
  level: 2
```

Tested chart with Rancher v2.3.5:
- `auditLog.destination: hostPath` no longer creates a sidecar
- `auditLog.destination: sidecar` and `auditLog.hostPath` creates a sidecar and logs to a specified directory

gz#16192
gz#17090


chart/templates/deployment.yaml
```diff
@@ -167,7 +169,7 @@ spec:
           secretName: tls-ca
 {{- end }}
 {{- if gt (int .Values.auditLog.level) 0 }}
-  {{- if eq .Values.auditLog.destination "hostPath" }}
+  {{- if .Values.auditLog.hostPath }}
```
loganhz, line 172: Please just keep `if eq .Values.auditLog.destination "hostPath" `. We don't need to change this line.

there is a default value of `.Values.auditLog.hostPath`. If you do so, no matter if user is using `hostPath` mode, Rancher will write audit logs to the default hostPath on the node.

dkeightley: Done

loganhz (dismissed):

rawmind0 (approved):
LGTM

rawmind0:
CI is failing on master branch due to PR https://github.com/rancher/rancher/pull/35157 was not back ported

`Package 'docker-20.10.6_ce-6.49.3' not found.`

rawmind0:
As master branch is not active anymore, this PR is superseded by: 
* 2.6: https://github.com/rancher/rancher/pull/35805 
* 2.5: https://github.com/rancher/rancher/pull/35806

Issue #26204, opened 2020-03-23
Baidu Cluster Driver freezes UI when navigating to Add Cluster Baidu CCE


izaac:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Activate Baidu Cluster Driver
2. Activate Browser Devtools and open the Terminal Tab.
3. Navigate to the Add Cluster page
4. Click Baidu CCE

**Result:**

A GET request is made to:

`https://x.x.x.x/meta/proxy/https:/cluster-driver.oss-cn-shenzhen.aliyuncs.com/baidu/ui/component.js `

Resulting in:
net::ERR_ABORTED 502 (Bad Gateway)

`' because its MIME type ('') is not a supported stylesheet MIME type, and strict MIME checking is enabled.`

Clicking in the Baidu CCE results in:

`Uncaught (in promise) TypeError: Cannot read property 'manager' of null` and UI is unresponsive.

**Environment information**
- Rancher version: `2.4 master-head (03/23/2020)` commit id: 371ec137f
- Installation option (single install/HA): single

izaac:
Rancher version: `2.4 master-head (03/24/2020)` commit id: 4aa97c7be

Validated the UI is no longer freezing and no more browser console or network errors.

PR #26224, opened 2020-03-24
Move china cloud provider drivers to aliyun oss with global acceleration

loganhz:
Move china cloud provider drivers to aliyun oss with global acceleration

Most of China users cannot download drivers on github due to network issues.

Also users in other places sometimes cannot download some drivers or drivers' UI (such as Baidu cluster driver UI) hosted in aliyun oss

https://github.com/rancher/rancher/issues/26204


orangedeng:
Should we rename the url to make kontainer engine driver and node driver have the same format? Like `kontainer-engine-driver-xxx` and `node-driver-xxx`.

loganhz:
Updated

Issue #26219, opened 2020-03-23
CIS Scan - Schedule scan report is seen in "Creating" state when a schedule scan is run on cluster with k8s version less than 1.15


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Bring up a cluster with k8s 1.13 in rancher:v2.2.10
- Upgrade rancher to `master-head` latest
- Edit cluster --> Advanced options --> Enable Schedule CIS scan --> Set it to True.
- Set an appropriate cron time.
- Click on Save.

**Actual Result:**
- Schedule scan is started on the cluster. Report is stuck in "Creating" State
<img width="1392" alt="Screen Shot 2020-03-23 at 3 35 22 PM" src="https://user-images.githubusercontent.com/26032343/77369468-ef9fe700-6d1b-11ea-8a69-688f368cb524.png">

- But error is seen in the logs . Logs:
```
2020/03/23 22:30:30 [DEBUG] cisScanHandler: checking if cluster version v1.13.0 is less than min version: 1.15
2020/03/23 22:30:30 [ERROR] ClusterScanController c-pdbvf/ss-cis-1585002120000225789 [cisScanHandler] failed with : minimum k8s version 1.15 needed for running cis scan
```

**Expected Result:**
User should NOT be allowed to enable a schedule scan OR when user enables schedule scan, it should not be taken into consideration for a cluster with k8s version < 1.13

**Other details that may be helpful:**
Related to issue: https://github.com/rancher/rancher/issues/25996

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `07f6b9a33`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke
- Kubernetes version (use `kubectl version`):

```
1.13
```

sowmyav27:
**Verified on master-head commit id: `759736e81`**

- Upgrade cluster setup from 2.2.10 to master-head
- User is not able to Edit cluster and enabled Schedule CIS scan

<img width="1334" alt="Screen Shot 2020-03-24 at 2 16 26 PM" src="https://user-images.githubusercontent.com/26032343/77478100-aa8eba00-6dda-11ea-920e-aea5d3cdac0c.png">

- Edit cluster through API: Give in value for `scheduledClusterScan`
```
[...]

PR #26228, opened 2020-03-24
add version check when enabling scheduled scan

leodotcloud:
https://github.com/rancher/rancher/issues/26219

### DIFF
--- a/pkg/api/customization/cluster/validator.go
+++ b/pkg/api/customization/cluster/validator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher/norman/types/convert"
 	gaccess "github.com/rancher/rancher/pkg/api/customization/globalnamespaceaccess"
 	"github.com/rancher/rancher/pkg/controllers/management/k3supgrade"
+	"github.com/rancher/rancher/pkg/controllers/user/cis"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/settings"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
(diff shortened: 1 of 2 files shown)


pkg/api/customization/cluster/validator.go
```diff
@@ -48,17 +53,29 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
...
+	overrideBenchmarkVersion := ""
+	if spec.ScheduledClusterScan.ScanConfig.CisScanConfig != nil {
+		overrideBenchmarkVersion = spec.ScheduledClusterScan.ScanConfig.CisScanConfig.OverrideBenchmarkVersion
+	}
+	_, _, err := cis.GetBenchmarkVersionToUse(overrideBenchmarkVersion, currentK8sVersion,
+		v.CisConfigLister, v.CisConfigClient,
+		v.CisBenchmarkVersionLister, v.CisBenchmarkVersionClient,
+	)
+	if err != nil {
+		return err
```
StrongMonkey: We should return a HTTP error in here

StrongMonkey:
Also build is failing.

pkg/api/customization/cluster/validator.go
leodotcloud: I will return a http error, the message is already coming from the function being called.

pkg/api/customization/cluster/validator.go
```diff
@@ -48,18 +53,33 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 		return err
 	}
 
-	if err := validateScheduledClusterScan(&spec); err != nil {
-		return err
+	if err := v.validateScheduledClusterScan(&spec); err != nil {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
```
StrongMonkey: If we are wrapping the error, would the message have duplicated http code? since you are wrapping one http error with another http error that could possibly be returned in `validateScheduledClusterScan` method.

PR #26259, opened 2020-03-25
updating default k8s from rke

luthermonson:


deniseschannon:
we should update RKE along with kdm to keep the commits in sync.

PR #26299, opened 2020-03-26
[2.2x] Bump helm unit-test version

cbron:
Backport fix for changes in helm unit-test


cbron:
Very weird drone logs on that passing CI run, maybe I'm not used to 2.2 logs.

Issue #26260, opened 2020-03-25
coredns add on job fails after an addition of worker node and a cluster upgrade


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy a cluster - 1 etcd, 1 worker, 1 control node.
2. Add a worker node.
3. Upgrade the cluster, by changing the max-pods of kubelet. (in cluster.yml)

**Expected Result:**
rke coredns add on job should have deployed and pod should be in "Active" state

**Actual Result:**
The are in error state and Unavailable
<img width="1291" alt="Screen Shot 2020-03-25 at 1 18 01 PM" src="https://user-images.githubusercontent.com/26032343/77582446-a4f9a880-6e9c-11ea-8193-3b0cbc75dc91.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancmaster-head - commit id: `ae28b67cc`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke do
- Kubernetes version (use `kubectl version`):

```
1.17
```

mrajashree:
@sowmyav27 I'm not able to repro with the above steps, commit id `e783315279c389c73ed4b465b261d0004ffbf6a0`

sowmyav27:
On master-head 
- First edit on a cluster, cluster.yaml file, change max-pods, dns params are :
```
  dns:
    linear_autoscaler_params: {}
    node_selector: null
    nodelocal:
      ip_address: ''
      node_selector: null
      update_strategy: {}
    reversecidrs: null
    stubdomains: null
    update_strategy: {}
    upstreamnameservers: null
```
- Request sent:
```
dns: {nodelocal: {updateStrategy: {}, nodeSelector: null, ipAddress: ""}, reversecidrs: null,…}
nodelocal: {updateStrategy: {}, nodeSelector: null, ipAddress: ""}
updateStrategy: {}
nodeSelector: null
ipAddress: ""
reversecidrs: null
stubdomains: null
upstreamnameservers: null
updateStrategy: {}
```
- Response received:
```
"dns":{"nodelocal":{"ipAddress":"","type":"/v3/schemas/nodelocal"},"type":"/v3/schemas/dnsConfig"}
```
[...]

PR #26300, opened 2020-03-26
Update types and rke

mrajashree:


deniseschannon:
https://github.com/rancher/rancher/issues/26260

Issue #26304, opened 2020-03-27
k8s version is not available sometimes in the dropdown while creating a cluster in a HA setup


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
- Bring up a HA cluster on master-head
- Deploy a DO cluster.
- k8s version dropdown is empty
<img width="910" alt="Screen Shot 2020-03-26 at 5 36 05 PM" src="https://user-images.githubusercontent.com/26032343/77710056-3f83e580-6f8a-11ea-9989-72badcee1223.png">

- On a refresh of kubernetes metadata, the k8s version were available in the dropdown during cluster creation

**Expected Result:**
k8s version should be available in the dropdown

**Other details that may be helpful:**
Note: Not always reproducible (hit this issue 1 out of 3 ha setups on master-head)

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `a195fe7c6`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke do
- Kubernetes version (use `kubectl version`):

```
1.17
```

deniseschannon:
@sowmyav27 How quickly are you going to start creating clusters? I wonder if it's a timing thing?

StrongMonkey:
@sowmyav27  if you reproduced it next time, can you share the setup? I'd like to know if version never shows up until you refresh or it is just a timing thing.

sowmyav27:
@deniseschannon @StrongMonkey 
Logs:
```
2020/03/27 00:25:44 [INFO] Refreshing driverMetadata in 1440 minutes
2020/03/27 00:25:44 [INFO] driverMetadata: refreshing data from upstream https://releases.rancher.com/kontainer-driver-metadata/dev-v2.4/data.json
2020/03/27 00:25:45 [INFO] kontainerdriver googlekubernetesengine listening on address 127.0.0.1:42833
2020/03/27 00:25:45 [INFO] kontainerdriver googlekubernetesengine stopped
2020/03/27 00:25:45 [INFO] kontainerdriver amazonelasticcontainerservice listening on address 127.0.0.1:44585
2020/03/27 00:25:45 [INFO] kontainerdriver amazonelasticcontainerservice stopped
[...]

PR #26339, opened 2020-03-28
Force Enqueue after sync, change log level to info

StrongMonkey:


StrongMonkey:
Potentially fix https://github.com/rancher/rancher/issues/26304. I couldn't produce it locally but I think this should fix it. Also changed log level to better understand when controller are skipping refreshing.

PR #26343, opened 2020-03-29
Force Enqueue after sync, change log level to info

StrongMonkey:


deniseschannon:
In https://github.com/rancher/rancher/pull/26344

Issue #26399, opened 2020-03-31
Upgrade from Rancher 2.3.x -> 2.4.0 caused all Kubelets to constantly be restarted


Oats87:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**

Upgrade Rancher installation from 2.3.x to 2.4.0

**Result:**

Observe that an "unmanaged" node agent is started on the nodes, and is constantly restarting the Kubelet with log messages like:
```
time="2020-03-31T15:55:34Z" level=info msg="node fe-downstream-ingress-0 is not registered, restarting kubelet now"
time="2020-03-31T15:55:52Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:56:27Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:57:02Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:57:37Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:58:12Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:58:35Z" level=info msg="node fe-downstream-ingress-0 is not registered, restarting kubelet now"
time="2020-03-31T15:58:47Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:59:22Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T15:59:57Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T16:00:32Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T16:01:07Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T16:01:35Z" level=info msg="node fe-downstream-ingress-0 is not registered, restarting kubelet now"
time="2020-03-31T16:01:42Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T16:02:18Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T16:02:53Z" level=info msg="Plan monitor checking 35 seconds"
time="2020-03-31T16:03:28Z" level=info msg="Plan monitor checking 35 seconds"
```

When looking at the containers running on the node: 
```
[root@fe-downstream-ingress-0 ~]# docker ps
[...]

PR #26409, opened 2020-03-31
Stop reconciling if managed agent is found

StrongMonkey:
This double checks if manage agent is found and unmanaged agent can't be
cleaned up by managed agent due to race condition. Once managed agent
is found then the loop is immediately terminated.

### DIFF
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/mattn/go-colorable"
+	"github.com/rancher/norman/types/slice"
 	"github.com/rancher/rancher/pkg/agent/clean"
 	"github.com/rancher/rancher/pkg/agent/cluster"
 	"github.com/rancher/rancher/pkg/agent/node"
@@ -320,6 +321,7 @@ func run() error {
 		// launching reconcileKubelet in a goroutine to make sure kubelet will be restarted
 		// every three minutes until managed agent comes up and take over
 		go func() {
+		loop:
 			for {
 				select {
 				case <-time.After(3 * time.Minute):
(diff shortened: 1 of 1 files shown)


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/26399.

StrongMonkey:
@cjellick Can you help review this PR? since you were the original reviewer.

Oats87:
@StrongMonkey I believe the `cattle-cluster-agent` is also "managed" and would false trigger the loop to quit

StrongMonkey:
@Oats87 you are right, I should add another check for node-agent.

pkg/agent/main.go
```diff
@@ -422,3 +433,32 @@ func reconcileKubelet(ctx context.Context) (bool, error) {
...
+		return false, err
+	}
+
+	for _, container := range containers {
+		con, err := c.ContainerInspect(ctx, container.ID)
+		if err != nil {
+			return false, err
+		}
+		// this check if container is a managed node-agent
+		if _, ok := container.Labels["io.kubernetes.pod.namespace"]; ok &&
```
kinarashah, line 455: not sure, but do we need to also check for container state?

pkg/agent/main.go
```diff
@@ -328,7 +330,7 @@ func run() error {
 						logrus.Errorf("failed to reconcile kubelet: %v", err)
 					}
 					if done {
-						break
+						break loop
```
dramich, line 333: Can this return instead?

StrongMonkey: yeah that would work

pkg/agent/main.go
StrongMonkey: I think checking the existence of the container is sufficient, don't want the state impact this logic tho.

StrongMonkey: what's your concern?

Issue #26286, opened 2020-03-26
UI shows Upgrading even after the cluster is already upgraded in a single node k3s cluster


ShylajaDevadiga:
**Steps to reproduce (least amount of steps as possible):**
Import a single node k3s cluster with v1.0.1
Upgrade to the cluster to v1.17.4+k3s1

**Result:**
Cluster is upgraded to the selected version. UI shows still Upgrading

**Other details that may be helpful:**
This issue os not seen in multinode setup.

**Environment information**
- Rancher version: master-head (ImageID: ecd553b460c9)

Screenshot attached:  
![image](https://user-images.githubusercontent.com/56045581/77683871-54df1c80-6f56-11ea-91d6-7803de52033e.png)

daxmc99:
Unable to replicate this bug on master-head (image id: eb09dcb7680b)

daxmc99:
Upgraded to latest-master head, again, unable to replicate.

sangeethah:
@daxmc99 Can you have this merged for 2.4 release as well ?

deniseschannon:
@ShylajaDevadiga @sangeethah this is ready to test in v2.4 now.

ShylajaDevadiga:
Issue mentioned above is resolved. Verified using 2.4-head as well as master-head(Image ID:b0e67872d010), Single control and multi control node. With Drain checked/unchecked.
2.4-head
![image](https://user-images.githubusercontent.com/56045581/78399503-d9124f00-75a9-11ea-92af-aab16df1a3ac.png)

master-head
![image](https://user-images.githubusercontent.com/56045581/78399648-27275280-75aa-11ea-99db-8bc6dfeb6dca.png)

khushboo-rancher:
K3s upgrade verified on rancher server ```master-2689-head - d18a0b6b4``` and ```v2.4-2686-head-00827641e``` worked fine.

Scenarios tested:
1. Single node
2. Multi nodes - combination of single/multiple control plane and workers.
3. Drain option enabled and disabled.
4. Concurrency for control/worker nodes as 1 and 2

PR #26411, opened 2020-03-31
Enqueue controller if still in upgrading state

daxmc99:
While upgrading k3s clusters the controller may cause updates that do not change the cluster object. This means that the controller will not be run again to exit the updating state. In these cases the controller should re-enqueue itself.

Addresses https://github.com/rancher/rancher/issues/26286

### DIFF
--- a/pkg/controllers/management/k3supgrade/register.go
+++ b/pkg/controllers/management/k3supgrade/register.go
@@ -2,6 +2,7 @@ package k3supgrade
 
 import (
 	"context"
+	"time"
 
 	"github.com/rancher/rancher/pkg/clustermanager"
 	"github.com/rancher/rancher/pkg/systemaccount"
@@ -23,6 +24,7 @@ type handler struct {
 	nodeLister             v3.NodeLister
 	systemAccountManager   *systemaccount.Manager
 	manager                *clustermanager.Manager
+	clusterEnqueueAfter    func(name string, duration time.Duration)
 }
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/k3supgrade/register.go
```diff
@@ -37,6 +38,7 @@ func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.M
 		systemUpgradeNamespace: systemUpgradeNS,
 		clusterCache:           wContext.Mgmt.Cluster().Cache(),
 		clusterClient:          wContext.Mgmt.Cluster(),
+		enqueue:                wContext.Mgmt.Cluster().Enqueue,
```
dramich, line 41: Let's use engueueAfter just so we aren't stuck in a loop of checking even though we know the clusters is still ugprading.

pkg/controllers/management/k3supgrade/register.go
```diff
@@ -37,6 +39,7 @@ func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.M
 		systemUpgradeNamespace: systemUpgradeNS,
 		clusterCache:           wContext.Mgmt.Cluster().Cache(),
 		clusterClient:          wContext.Mgmt.Cluster(),
+		enqueueAfter:           wContext.Mgmt.Cluster().EnqueueAfter,
```
dramich, line 42: nit: prefix with the type in case we need more later.

daxmc99: Added prefix

daxmc99:
> LGTM with a nit: could probably explain enqueue or update reasoning in a comment

Do you think the commit message is enough?

Issue #25835, opened 2020-03-06
Cannot fully rotate AWS credentials for EKS clusters


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. create eks cluster
2. let it provision
3. change access keys and secret key via the UI or kubectl
4. in the API/UI it will appear the access key is change
5. delete the old key from AWS, and attempt to update EKS version or delete the cluster

**Result:**
will not work due to invalid credentials

**Other details that may be helpful:**
I believe this is happening because the driver store the state for the cluster in a secret. Since the only thing that can be changed in actual AWS for the EKS cluster is the version, the driver will only update version in the secret it maintains for the clusters current state. If you decode the secret you will see the key has not changed.

rmweir:
Created backport issue, this will be used to track master, the other will be for 2.4. There is no milestone for 2.5 yet, so it will stay as 2.4.

rancher-max:
#### Validated in both master-2689-head (`d18a0b6b4`) and v2.4-2696-head (`04159d027`)

Followed the same steps as noted above: 

- Keys can be rotated and the cluster continues to work. It allows updating, continues to function properly, and deletion. 

- The secret that Rancher stores (mentioned in the original issue comment) is not updated. Leaving this open for @rmweir to investigate further.

rancher-max:
The secret is updated correctly. The correct secret to verify is in namespace `cattle-system` and is like `c-c-*****`. Above was a user-error and this is fully functional.

Issue #24171, opened 2019-11-20
Cluster Upgrade stuck if EKS Cluster was updated manually


ms-dennis:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
* Create EKS Cluster with Rancher(use a previous version to be able to upgrade ot the newest one)
* Manually Update the EKS Cluster via AWS Console
* Edit the Cluster in Rancher and set k8s Version to the most recent one

**Result:**
Cluster is stuck in upgrade process:

InvalidParameterException: Unsupported Kubernetes minor version update from 1.14 to 1.14 status code: 400, request id: 0ccda22c-e6ad-47ba-a0e4-4bdbf1b0f4ce

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
* Cluster was initially created with Rancher 2.2.x
* Updated the Cluster with Rancher 2.3.2 to k8s Version 1.14
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): EKS
- Machine type (cloud/VM/metal) and specifications (CPU/memory): c5.4xlarge (16/32G)
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.2", GitCommit:"f6278300bebbb750328ac16ee6dd3aa7d3549568", GitTreeState:"archive", BuildDate:"2019-08-29T18:43:18Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6-eks-5047ed", GitCommit:"5047edce664593832e9b889e447ac75ab104f527", GitTreeState:"clean", BuildDate:"2019-08-21T22:32:40Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):

```
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a215d7133c34aa18e3b72b4a21fd0c6136
 Built:             Mon Mar  4 21:25:41 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
[...]

PR #26412, opened 2020-03-31
Eks edit other

rmweir:
**Issues:**
https://github.com/rancher/rancher/issues/25835
https://github.com/rancher/rancher/issues/24171

### DIFF
--- a/pkg/controllers/management/clusterprovisioner/provisioner.go
+++ b/pkg/controllers/management/clusterprovisioner/provisioner.go
@@ -813,14 +813,6 @@ func (p *Provisioner) getSpec(cluster *v3.Cluster) (*v3.ClusterSpec, error) {
 		return nil, err
 	}
 
-	// Version is the only parameter that can be updated for EKS, if they is equal we do not need to update
-	// TODO: Replace with logic that is more adaptable
-	if cluster.Spec.GenericEngineConfig != nil && (*cluster.Spec.GenericEngineConfig)["driverName"] == "amazonelasticcontainerservice" &&
-		cluster.Status.AppliedSpec.GenericEngineConfig != nil && (*cluster.Spec.GenericEngineConfig)["kubernetesVersion"] ==
-		(*cluster.Status.AppliedSpec.GenericEngineConfig)["kubernetesVersion"] {
-		return nil, nil
-	}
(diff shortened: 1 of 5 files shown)


pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -813,14 +813,6 @@ func (p *Provisioner) getSpec(cluster *v3.Cluster) (*v3.ClusterSpec, error) {
...
 	}
 
-	// Version is the only parameter that can be updated for EKS, if they is equal we do not need to update
-	// TODO: Replace with logic that is more adaptable
-	if cluster.Spec.GenericEngineConfig != nil && (*cluster.Spec.GenericEngineConfig)["driverName"] == "amazonelasticcontainerservice" &&
-		cluster.Status.AppliedSpec.GenericEngineConfig != nil && (*cluster.Spec.GenericEngineConfig)["kubernetesVersion"] ==
-		(*cluster.Status.AppliedSpec.GenericEngineConfig)["kubernetesVersion"] {
-		return nil, nil
-	}
-
```
cbron, line 823: This must be removed to allow client id and secret to be updated right ? 

For the upgrade-when-downstream-is-already-upgraded scenario, it previously never would have gotten hit because it returned an error above ? Only when the cluster legitimately didn't need to be upgraded. And now in that same scenario the cluster will be refreshed from update call above, so the standard checks will work. That sound right ?

rmweir: It actually still got hit because as far as rancher was concerned the version were not equal. Scenario:
your desired and current version is 1.13.
in eks you change to 1.14.
in rancher you upgrade to 1.14, now current is 1.13 and desired is 1.14.
controller explodes.

This was the same prior and is the second issue this PR addresses.

cbron: So to confirm new approach in that same scenario: everything is the same except when you upgrade it doesn't error out and status/spec would both equal 1.14. And also client id/secret can be updated.

rmweir: Correct.

Issue #26416, opened 2020-04-01
Project owner cannot see PVC in 2.4.0


loganhz:
rancher/rancher:v2.4.0

Docker install

Steps:

1. Create a standard user `user1`
2. Add `user1` into Default project as project owner
3. Login as `user1`
4. Import a PVC

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: mypvc
  namespace: default
spec:
  accessModes: ["ReadWriteMany"]
  resources:
    requests:
      storage: 6Gi
```

Results:

Project owner can import and create PVC but can not see it in Rancher UI. 

Other:

It works in 2.3.5 and 2.3.6

cjellick:
i've now seen this with:
- persistenVolumeClaim
- configMap
- statefulSet
- cronJob
- daemonSet

I'm guesing it is a case-sensitivity issue in https://github.com/rancher/rancher/tree/master/vendor/github.com/rancher/steve/pkg/accesscontrol

sangeethah:
Also see with HPA.

sowmyav27:
**Verified with 2.4.2-rc1**

Verified "list" is working for project member/project owner/cluster member users for

- Persistent Volume
- persistenVolumeClaim
- configMap
- statefulSet
- cronJob
- daemonSet
- HPA

PR #26424, opened 2020-04-01
Lowercase resource set in proxyStore

ibuildthecloud:
The resource name is takes from the Norman schema PluralName field. In norman this is a plural form of the ID field which is typically camelCase.  A Kubernetes resource field is always lowerscase. In Rancher <=2.3 the RBAC system would do case insensitive checks against all strings. In Rancher 2.4 the RBAC system was switch from Norman to Steve.  Steve's implementation is much faster and more proper and as a result does not do case insensitive checks. It is expected that all input to the RBAC system is proper. This is the one place in Rancher where we weird not generating a proper resource name.


deniseschannon:
https://github.com/rancher/rancher/issues/26416

deniseschannon:
https://github.com/rancher/rancher/issues/26582

Issue #26457, opened 2020-04-02
Self signed TLS fails to generate after v2.4.2 upgrade, UI won't load


andruwa13:
[ERROR] MultiClusterAppRevisionController cattle-global-data/mcapprevision-t7zmd [management-multiclusterapp-revisions-rbac] failed with : multiClusterApp.management.cattle.io "cattle-global-data/cert-manager" not found

[INFO] failed to create TLS cert for: 172.17.0.2

run command:

docker run -d -v /home/rancher/data:/var/lib/rancher --restart=unless-stopped -p 8080:80 -p 8443:443  --name rancher-server rancher/rancher

os Ubuntu 18.04.4
Docker version 19.03.8, build afacb8b7f0

mshade:
I had the same issue. I switched to [providing my own self-signed cert to get going](https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/single-node-docker/#option-b-bring-your-own-certificate-self-signed)

If I restart rancher server without the provided self-signed cert, the problem returns. Logs look like this:

```
rancher_1  | 2020/04/03 00:01:46 [INFO] Creating new TLS secret for serving-cert (count: 16): map[listener.cattle.io/redacted listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/redacted listener.cattle.io/redacted:8443 listener.cattle.io/cn-172.22.0.2:172.22.0.2 listener.cattle.io/cn-redacted listener.cattle.io/cn-hidden listener.cattle.io/redacted listener.cattle.io/redacted listener.cattle.io/cn-localhost:localhost listener.cattle.io/cn-redacted listener.cattle.io/redacted listener.cattle.io/redacted listener.cattle.io/redacted listener.cattle.io/cn-redacted listener.cattle.io/redacted listener.cattle.io/hash:0cd54ff8cbe4580f42c47283d84c3803465313dee49c616668ff4cd1c0aca82f]
rancher_1  | 2020/04/03 00:01:46 [INFO] failed to create TLS cert for: 172.22.0.2
```

How is it auto-detecting these hostnames? Some of the `redacted` names are unknown to me...

ibuildthecloud:
If anyone hits this issue you can run the below in the kubernetes cluster run Rancher.  If this is a single node setup just `docker exec -it CONTAINER bash` first and then run the below.

```
kubectl -n cattle-system delete secret serving-cert
kubectl delete listenconfigs cli-config
```
[...]

Issue #26282, opened 2020-03-26
User is thrown out of Rancher UI after few minutes of launching new rancher master setup


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Launch Rancher server - master
- After entering a password and setting the rancher server URL, user is navigate to the Rancher UI
- Deploy a do CLuster. When the cluster is in provisioning state, user will be thrown out of this UI.
- He will be thrown to this page - Warning: Your connection is not private is seen
- Click on Advanced --> proceed ( Chrome web page)
- Logs show something similar to:

```
raft2020/03/26 17:33:40 INFO: 8e9e05c52164694d is starting a new election at term 1
raft2020/03/26 17:33:40 INFO: 8e9e05c52164694d became candidate at term 2
raft2020/03/26 17:33:40 INFO: 8e9e05c52164694d received MsgVoteResp from 8e9e05c52164694d at term 2
raft2020/03/26 17:33:40 INFO: 8e9e05c52164694d became leader at term 2
raft2020/03/26 17:33:40 INFO: raft.node: 8e9e05c52164694d elected leader 8e9e05c52164694d at term 2
2020-03-26 17:33:40.928199 I | etcdserver: published {Name:default ClientURLs:[http://localhost:2379]} to cluster cdf818194e3a8c32
2020-03-26 17:33:40.928334 I | etcdserver: setting up the initial cluster version to 3.4
2020-03-26 17:33:40.928400 I | embed: ready to serve client requests
2020-03-26 17:33:40.929202 N | embed: serving insecure client requests on 127.0.0.1:2379, this is strongly discouraged!
2020-03-26 17:33:40.929585 N | etcdserver/membership: set the initial cluster version to 3.4
2020-03-26 17:33:40.929671 I | etcdserver/api: enabled capabilities for version 3.4
2020/03/26 17:33:41 [INFO] Waiting for k3s to start
time="2020-03-26T17:33:41Z" level=info msg="Preparing data dir /var/lib/rancher/k3s/data/7c4aaa633ac3ff4849166ba2759da158a70beb5734940e84b6e67011a35f4c59"
2020/03/26 17:33:42 [INFO] Waiting for k3s to start
2020/03/26 17:33:43 [INFO] Waiting for k3s to start
time="2020-03-26T17:33:43.552090192Z" level=info msg="Starting k3s v1.17.2+k3s1 (cdab19b0)"
[...]

PR #26464, opened 2020-04-03
Lock dynamic TLS to server-url value

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/26282
https://github.com/rancher/rancher/issues/26457

PR #26481, opened 2020-04-03
Add new suse and sles images

aaronRancher:
Added images for `suse-12-sp5-docker-19.03.1-kernel-4.12.14-122.17.1` and `sles15-sp1-docker-19.03.1`


aaronRancher:
We will remove private AMI functionality from aws.py. Will switch to AMI utilization solely on Jenkins. Closing PR as a result.

Issue #14420, opened 2018-07-06
Preferences show other user's name and username instead of the logged in user


divyangjp:
**Rancher versions:**
rancher/server or rancher/rancher: v2.0.4
rancher/agent or rancher/rancher-agent:

**Infrastructure Stack versions:**
healthcheck:
ipsec:
network-services:
scheduler:
kubernetes (if applicable): 1.10.3

**Docker version: 17.03.2

**Operating system and kernel: Ubuntu 16.04

**Type/provider of hosts: Openstack

**Setup details: Rancher HA with 3 nodes and internal DB

**Environment Template: Kubernetes

**Steps to Reproduce:**
Create new Admin user say `xyz@abc.com`. 
Using any admin user create standard user say `stuser@abc.com`. 
Now login with standard user.
Go to `Preferences`. 
**Results:**
Check that `My Account` contains name and username of Admin user `xyz@abc.com`

**Expected Result**
Name and username should be of logged in user and not the admin user.

loganhz:
Hi @divyangjp , I can not reproduce this issue.

Can you provide the screenshot like this, please?

![image](https://user-images.githubusercontent.com/21168270/42438311-47b4add8-8392-11e8-9548-3a76e2fbe158.png)

loganhz:
Steps to reproduce:
1. Create an admin user `user1`.
2. Login as `user1` and create a `standard` user `user2`.
3. Change the role of `user1` to standard user.
4. Logout `user1` and Login as `user2`

API `/v3/users?me=true` returns `user1`.

I think we need to fix this issue in server side.

cjellick:
Thanks for the steps @loganhz

almoore:
I am still seeing this issue in the v2.0.6 release. I am using the OpenLDAP setup and have only the admin user locally. All the users see the Preferences user info as the admin user.

erikwilson:
It appears this was fixed in v2.0.8

loganhz:
API `/v3/users?me=true` returns the multiple users in `v2.0.4`
![image](https://user-images.githubusercontent.com/21168270/50672616-543a2d80-1014-11e9-86f0-2f55a082f45b.png)


API `/v3/users?me=true` returns the current user in `v2.1.4`
![image](https://user-images.githubusercontent.com/21168270/50672593-1f2ddb00-1014-11e9-9b8e-fc5a4de46f1b.png)

This issue has been fixed in current rancher/rancher:stable version

[...]

PR #26493, opened 2020-04-04
[2.4] Fix the issue that preferences show other user's name 

loganhz:
**Problem:**

Preferences show other user's name and username instead of the logged in user

**Solution:**

Users can set `Me` field of another user to true in Rancher UI.
And it will be always true even after logout and login.
Set `Me` field to nocreate and noupdate

**Related Issue:**

rancher/rancher#14420

**Related PR:**

https://github.com/rancher/types/pull/1131


GGGitBoy:
@deniseschannon Updated

Issue #26448, opened 2020-04-02
Upgrade K3OS cluster fails


DarrenRatcliffe:
**What kind of request is this (question/bug/enhancement/feature request):**
Question / Bug?

**Steps to reproduce (least amount of steps as possible):**
In Rancher 2.4.2 edit imported K3OS cluster (K3OS v0.9.1) pick latest release and save

**Result:**
system upgrade controller deploys but fails

Rancher server stuck in 'Cluster is being upgraded'

controller container log output:

`W0402 15:40:10.608696       1 client_config.go:543] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
time="2020-04-02T15:40:10Z" level=fatal msg="Error starting: namespaces \"kube-system\" is forbidden: User \"system:serviceaccount:cattle-system:system-upgrade\" cannot get resource \"namespaces\" in API group \"\" in the namespace \"kube-system\""`

**Other details that may be helpful:**

**Environment information**
- Rancher version: v2.4.2
- Installation option: single install
- K3OS: v0.9.1
- K3S: v1.17.2+k3s1

sangeethah:
Able to upgrade a imported K3s cluster on version `v1.0.1` to `v1.17.4+k3s1` successfully using rancher server version - `v2.4.2`.

cjellick:
We'll have to clarify this in our docs and possibly prevent it at the API level, but the integrated upgrades are not intended for k3os, just k3s installed into vanilla linux os like ubuntu (currently the only officially supported os for k3s).

We haven't focused on a k3os integration because that is still very much a non-produciotn/non-GA project.

DarrenRatcliffe:
> We'll have to clarify this in our docs and possibly prevent it at the API level, but the integrated upgrades are not intended for k3os, just k3s installed into vanilla linux os like ubuntu (currently the only officially supported os for k3s).
> 
> We haven't focused on a k3os integration because that is still very much a non-produciotn/non-GA project.

Ah OK, I figured if it worked for K3S then by default it should work for K3OS

DarrenRatcliffe:
I just went back and checked the K3OS github docs and in the automatic upgrade section it specifically states using the Rancher/system-upgrade-controller https://github.com/rancher/k3os to upgrade K3OS?

cjellick:
[...]

PR #26501, opened 2020-04-06
Add introductory support for recognizing k3os clusters

daxmc99:
Addresses #26448 

Merge types PR first:  https://github.com/rancher/types/pull/1135

How to test:
* clone k3os repo
* `./scripts/run-qemu k3os.mode=install k3os.install.device=/dev/vda k3os.install.silent k3os.install.power_off=true`
* `./scripts/run-qemu k3os.password=rancher` 
* import single node k3os cluster into Rancher


prachidamle:
> what's up with the non types go mod changes that were pulled in?

yeah seems some go mod vendor changes?

daxmc99:
@rmweir was able to replicate the vendor changes locally. I'm not exactly sure of the cause (either it wasn't vendored in last time it was updated or one of our dependencies changed the version it requires). This should not be a blocker for merging

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -973,20 +979,29 @@ func GetBackupFilename(backup *v3.EtcdBackup) string {
 	return snapshot
 }
 
-// transform an imported cluster into a k3s cluster using its discovered version
-func k3sClusterConfig(cluster *v3.Cluster) error {
+// transform an imported cluster into a k3s or k3os cluster using its discovered version
+func k3sClusterConfig(cluster *v3.Cluster, nodes []*v3.Node) error {
 	// version is not found until cluster is provisioned
-	if cluster.Status.Driver == "" || cluster.Status.Version == nil {
+	if cluster.Status.Driver == "" || cluster.Status.Version == nil || len(nodes) == 0 {
```
cbron, line 985: Is it possible for `nodes == 0` check to block the normal `k3s` cluster from coming through here ? I'm not familiar here, but assuming k3s would still register 1 node even if it has no workers, and then it'd be ok.

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -973,20 +979,29 @@ func GetBackupFilename(backup *v3.EtcdBackup) string {
...
-	if cluster.Status.Driver == v3.ClusterDriverK3s {
+	if cluster.Status.Driver == v3.ClusterDriverK3s ||
+		cluster.Status.Driver == v3.ClusterDriverK3os {
 		return nil //no-op
 	}
 
 	if strings.Contains(cluster.Status.Version.String(), "k3s") {
-		cluster.Status.Driver = v3.ClusterDriverK3s
+		for _, node := range nodes {
+			if _, ok := node.Status.NodeLabels["k3os.io/mode"]; ok {
```
dramich, line 997: Can we confirm with @dweomer that this is the best way to determine if this is a k3os cluster? Also, can we end up in some hybrid setup with k3s management running some k3os downstream workers?

dramich:
I don't think this fixes the issue though, this is just marking the cluster type as k3os, the cluster was getting marked as k3s before which allowed the upgrade to happen. Should there be a split in the logic to stop the upgrade from happening?

pkg/controllers/management/clusterprovisioner/provisioner.go
daxmc99: Might slow down provisioning but I don't think it would block

pkg/controllers/management/clusterprovisioner/provisioner.go
dweomer: So, the presence of the `k3os.io/mode` label on a node is the most reliable way to know if it is a k3OS node but, as per the second part of @dramich's comment, does one k3OS node make the entire (possibly hetergeneous) cluster a k3OS cluster? It might make sense to include in the `k3s-[master|worker]` plans to exclude anything with the `k3os.io/mode` label

dramich:
Marking it as k3os will stop the upgrade here https://github.com/rancher/rancher/blob/master/pkg/controllers/management/k3supgrade/k3supgrade_handler.go#L28

pkg/controllers/management/clusterprovisioner/provisioner.go
dramich: Based on an offline (online) conversation with @daxmc99 we can't handle k3os yet, so if one node is k3os that would stop the upgrade from completing thus getting us back into the same place. Safe bet until this can be fully integrated is call the whole cluster a k3os cluster.

Issue #26496, opened 2020-04-05
Cannot add label to node in imported cluster


loganhz:
rancher 2.4.0 and rancher 2.3.6

1. Import a k3s  cluster or other any cluster
2. go to node page
3. Try to add a label to a node

Result:

Cannot add it

aaronRancher:
The bug is reproduced with Rancher server `rancher\rancher:v2.0.4`
**Steps:**
Import a k3s cluster or other any cluster
go to node page
Try to add a label to a node
**Result:** 
Node label is not saved

The bug fix is verified with Rancher server `rancher\rancher:master-2720-head` `version 3c6eb2b74` and `rancher\rancher:v2.4-2718-head` `version 64cf5ecd0`
Labels can be added and saved.
![Screen Shot 2020-04-07 at 4 17 42 PM](https://user-images.githubusercontent.com/57409052/78728433-29601880-78ec-11ea-879b-80230fef3ee8.png)
![Screen Shot 2020-04-07 at 4 17 54 PM](https://user-images.githubusercontent.com/57409052/78728250-b6ef3880-78eb-11ea-84bd-693607e75c98.png)

PR #26509, opened 2020-04-06
Check nil pointer for node.status.NodeConfig

StrongMonkey:
UpdateLabel method is called for any type of node, even if nodePlan is
passed into the method. This commit fixed the bug where
node.status.NodeConfig could be nil if it is a imported node.


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/26496.

loganhz:
`2.3` https://github.com/rancher/rancher/pull/26525
`2.4` https://github.com/rancher/rancher/pull/26524

PR #26511, opened 2020-04-06
Validation - Support HA install for most cert types

bmdepesa:
Support most cert types for HA installation: 
* Rancher self-signed
* Bring your own valid cert
* Bring your own self-signed cert
* LetsEncrypt

Certificate values are stored in Jenkins credentials

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -1,4 +1,6 @@
+import base64
 import os
+import pytest
 import rancher
 import time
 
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -121,17 +142,37 @@ def add_repo_create_namespace(repo=RANCHER_HELM_REPO):
...
+        helm_rancher_cmd = helm_rancher_cmd + \
+            " --set rancherImageTag=" + RANCHER_IMAGE_TAG
+
+    
+    if "byo-" in type:
+        valid_cert = True
+        if type == "byo-self-signed":
+            valid_cert = False
+
+        create_tls_secrets(valid_cert)
```
jiaqiluo, line 175: Can you simplify the logic here since there are only 2 cases:
```
if type == "byo-self-signed":
	create_tls_secrets(valid_cert=False)
elif type == "byo-valid":
	create_tls_secrets(valid_cert=True)
```

jiaqiluo:
Can you add some checks for necessary variables before staring running tests? Maybe a module-scope auto-run method? 
- when RANCHER_HA_CERT_OPTION == byo-valid, check whether RANCHER_VALID_TLS_CERT and RANCHER_VALID_TLS_KEY exist
- when RANCHER_HA_CERT_OPTION == byo-self-signed, check whether RANCHER_BYO_TLS_CERT, RANCHER_BYO_TLS_KEY, and RANCHER_PRIVATE_CA_CERT exist

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -140,6 +181,38 @@ def install_rancher_self_signed(repo=RANCHER_HELM_REPO):
...
+    ca_command = export_cmd + " && kubectl -n cattle-system create secret generic tls-ca --from-file=" + ca_path
+
+    print(tls_command)
+    out = run_command_with_stderr(tls_command)
+    print(out)
+
+    if valid_cert == False:
+        print(ca_command)
+        out = run_command_with_stderr(ca_command)
+        print(out)
```
rancher-max, line 208: I think we can remove all 4 of these print statements since `run_command_with_stderr` already outputs those by default

bmdepesa:
> Can you add some checks for necessary variables before staring running tests? Maybe a module-scope auto-run method?
> 
> * when RANCHER_HA_CERT_OPTION == byo-valid, check whether RANCHER_VALID_TLS_CERT and RANCHER_VALID_TLS_KEY exist
> * when RANCHER_HA_CERT_OPTION == byo-self-signed, check whether RANCHER_BYO_TLS_CERT, RANCHER_BYO_TLS_KEY, and RANCHER_PRIVATE_CA_CERT exist

Addressed

tests/validation/tests/v3_api/test_create_ha.py
bmdepesa: Removed all of the same pattern from the entire file. Didn't realize we had that in the method already.

tests/validation/tests/v3_api/test_create_ha.py
bmdepesa: Done

Issue #25227, opened 2020-02-03
[cli] Listing a lot of clusters takes a long time


dramich:
When listing clusters, there is a secondary call for every cluster to get a node count for that cluster, this is obviously bad in a setup running a high number of clusters. 

This should be moved into a formatter for the cluster so the CLI can stop doing the extra calls.

dramich:
This is available in cli v2.4.3-rc1 2.4 master

dramich:
Testing should include the CLI to validate node counts, and the API by looking at the cluster object which now has a node count. 

Without running a lot of clusters the slowness won't be visible.

rancher-max:
#### Ran Benchmarks on a Rancher v2.4-head server with 25 clusters
Previously, using cli v2.4.0:
```
Time (mean ± σ):      6.902 s ±  0.379 s   [User: 135.6 ms, System: 33.3 ms]
Range (min … max):    6.405 s …  7.325 s   
```

With v2.4.3-rc1:
```
Time (mean ± σ):      4.734 s ±  0.583 s    [User: 75.0 ms, System: 14.6 ms]
Range (min … max):    3.697 s …  5.044 s
```

- Time has been reduced by approx. 31%
- Correct node counts are still returned in the `rancher cluster ls` call.
- Also validated in cluster API there is a field called `nodeCount` which lists the correct number of nodes

rancher-max:
#### Ran Benchmarks on a Rancher v2.4-head server with 120 clusters

Previously, using cli v2.4.0:
```
Benchmark #1: rancher cluster ls
  Time (mean ± σ):     41.345 s ±  0.610 s    [User: 281.0 ms, System: 81.3 ms]
  Range (min … max):   40.322 s … 41.882 s    5 runs
```

With v2.4.3-rc1:
```
Benchmark #1: rancher cluster ls
  Time (mean ± σ):     24.802 s ±  0.662 s    [User: 110.8 ms, System: 20.9 ms]
  Range (min … max):   23.670 s … 25.421 s    5 runs
```

- Time reduction is more significant as we increase cluster size. 
This has been fixed

PR #26514, opened 2020-04-06
Add nodeCount to formatter

dramich:
Problem:
The CLI displays a node count for clusters which required the CLI to make multiple API calls

Solution:
Add a new nodeCount field based off a lister and included in the formatter to help the CLI 

Depends on https://github.com/rancher/types/pull/1133

Issue https://github.com/rancher/rancher/issues/25227

### DIFF
--- a/pkg/api/customization/cluster/formatter.go
+++ b/pkg/api/customization/cluster/formatter.go
@@ -12,11 +12,13 @@ import (
 	client "github.com/rancher/types/client/management/v3"
 	"github.com/rancher/types/config"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
 	v1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 )
 
 type Formatter struct {
 	KontainerDriverLister     v3.KontainerDriverLister
+	nodeLister                v3.NodeLister
 	clusterSpecPwdFields      map[string]interface{}
 	SubjectAccessReviewClient v1.SubjectAccessReviewInterface
 }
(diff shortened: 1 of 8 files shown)


pkg/api/customization/cluster/formatter.go
```diff
@@ -145,6 +148,12 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 		}
 	}
 
+	nodes, err := f.nodeLister.List(resource.ID, labels.Everything())
```
prachidamle, line 151: resource.ID will the clusterID, does it also act as the namespaceID for the List call?

dramich: Yes, the v3.nodes live in the cluster namespace

pkg/api/customization/cluster/formatter.go
```diff
@@ -145,6 +148,12 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+	if err != nil {
+		logrus.Warnf("error getting node list for cluster %s: %s", resource.ID, err)
+	} else {
+		resource.Values["nodeCount"] = len(nodes)
```
prachidamle, line 155: We will be showing the nodeCount, irrespective of the cluster state; Also we list all nodes irrespective of the node status. Just confirming we want is this way? Also,  can we add an integration test to check the nodeCount of a cluster created?

dramich: That is the current behavior when using the CLI

PR #26521, opened 2020-04-07
Only set TLS when tls dialer is not nil

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/26282
https://github.com/rancher/rancher/issues/26457

deniseschannon:
this is not an issue in `release/v2.4` and only an issue in master

PR #26530, opened 2020-04-07
[Backport v2.3]Check nil pointer for node.status.NodeConfig

StrongMonkey:
UpdateLabel method is called for any type of node, even if nodePlan is
passed into the method. This commit fixed the bug where
node.status.NodeConfig could be nil if it is a imported node.


StrongMonkey:
Fix https://github.com/rancher/rancher/pull/26525

Issue #26450, opened 2020-04-02
Helm version checks are too strict


brendarearden:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
use a random value for helm version
or
use a shortened version of helm version (example: "v2", "v3")

**Result:**
* isHelm2 check is too strict so tiller wont start. 
* isHelm3 check  is too strict so shortened values are not accepted

**Solution**
Tiller always checks to see if it is helm v2 before starting by checking param version against the v2 string, but users are not aware of the exact v2 string or might throw in something more generic like the shortened version (or just garbage). In this case we need to default to helm 2. 

Similarly, users specifying another version of v3 should end up using helm v3.

rancher-max:
Need a merge in 2.4 branch as well

cbron:
@rancher-max fixed, let me know if you need anything else.

rancher-max:
This is needed for the cli to allow passing `v2` or `v3` when adding a catalog and specifying helm version. Valid values are now `helm_v2`, `helm_v3`, `v2`, or `v3`. This has been validated in both `master-2750-head` and `v2.4-2745-head` Rancher.


Passing in garbage works in some cases. See two scenarios below:
- Pass in garbage v3 value
```
$ rancher catalog add --helm-version myv3 badversion-test https://github.com/rancher-max/helm-foo
FATA[0000] Bad response statusCode [422]. Status [422 Unprocessable Entity]. Body: [baseType=error, code=InvalidBodyContent, message=Invalid helm 3 version] from [https://173.230.145.148/v3/catalogs]
```
- Pass in garbage v2 value
```
$ rancher catalog add --helm-version myv2though badversion-test https://github.com/rancher-max/helm-foo
# Success!
```

I believe the error case is okay as it gives a descriptive error, and using `rancher catalog add --help` specifies to use `v2` or `v3` explicitly. @brendarearden and @cbron If you feel differently let me know. I'll leave this open pending your judgement, but if you agree we can close.

cbron:
@rancher-max yes thats by design. We wanted to give users who try to pass something v3-like to be warned that it failed. If the mangle v2 then we just use v2 anyway so we don't error out. If thats all then good to close.

PR #26531, opened 2020-04-07
Helm version checks

brendarearden:
**Problem**
Helm version checks were too strict causing tiller not to start for helm2 consistently
Helm version checks were too strict so helm3 could only be one value

**Solution**
Allow for helm3 to be "helm_v3" or "v3"
Add validation for helm 3 since it needs to be specific values
Change the isHelm2 check to !isHelm3 to ensure helmVersion for helm 2 can be more flexible

**Issue**
#26450

### DIFF
--- a/pkg/api/customization/catalog/validator.go
+++ b/pkg/api/customization/catalog/validator.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
+	"github.com/rancher/rancher/pkg/controllers/user/helm/common"
 )
 
 var (
@@ -24,11 +25,16 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
 			u.Scheme = strings.ToLower(u.Scheme) // git commands are case-sensitive
 			data["url"] = u.String()
 		}
-		return nil
 	} else if request.Method == http.MethodPost {
 		return httperror.NewAPIError(httperror.MissingRequired, "Catalog URL not specified")
 	}
 
+	if helmVersion, ok := data["helmVersion"]; ok {
+		toLowerHelmVersion := strings.ToLower(helmVersion.(string))
+		if strings.Contains(toLowerHelmVersion, "v3") && !common.IsHelm3(toLowerHelmVersion) {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "Invalid helm 3 version")
+		}
+	}
 	return nil
 }
 
--- a/pkg/api/customization/catalog/validator.go
+++ b/pkg/api/customization/catalog/validator.go
@@ -28,9 +28,10 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
 	} else if request.Method == http.MethodPost {
 		return httperror.NewAPIError(httperror.MissingRequired, "Catalog URL not specified")
 	}
-
+	//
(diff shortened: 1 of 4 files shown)


pkg/api/customization/catalog/validator.go
```diff
@@ -38,3 +45,10 @@ func validateURL(pathURL string) error {
 	}
 	return nil
 }
+
+func validateHelm3Version(helmVersion string) error {
+	if common.IsHelm3(helmVersion) {
+		return nil
+	}
+	return httperror.NewAPIError(httperror.InvalidBodyContent, "Invalid helm 3 version. 'v3' or 'helm_v3' are valid values")
```
dramich, line 53: Just return an error, the caller should decide what to do with it.

pkg/api/customization/catalog/validator.go
```diff
@@ -24,11 +25,17 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
...
 		}
-		return nil
 	} else if request.Method == http.MethodPost {
 		return httperror.NewAPIError(httperror.MissingRequired, "Catalog URL not specified")
 	}
 
+	helmVersion := strings.ToLower(data["helmVersion"].(string))
+	if strings.Contains(helmVersion, "v3") {
+		if err := validateHelm3Version(helmVersion); err != nil {
+			return err
```
dramich, line 35: Do the httperror here

pkg/api/customization/catalog/validator.go
brendarearden: change error message verbiage

pkg/api/customization/catalog/validator.go
```diff
@@ -24,11 +25,15 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
...
 			data["url"] = u.String()
 		}
-		return nil
 	} else if request.Method == http.MethodPost {
 		return httperror.NewAPIError(httperror.MissingRequired, "Catalog URL not specified")
 	}
 
+	if helmVersion, ok := data["helmVersion"]; ok {
+		if validateHelm3Version(strings.ToLower(helmVersion.(string))) {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "Invalid helm 3 version")
```
rmweir, line 34: I'm confused here. If your validate functions returns true, then the helm 3 version is invalid?

pkg/api/customization/catalog/validator_test.go
```diff
@@ -43,3 +43,33 @@ func TestValidateURL(t *testing.T) {
...
+		},
+		{
+			name:        "Invalid helm 3 value",
+			helmVersion: "sdfasdfasdv3",
+			wantErr:     true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotErr := validateHelm3Version(tt.helmVersion)
```
rmweir, line 71: this isn't an err?

pkg/api/customization/catalog/validator.go
```diff
@@ -38,3 +43,10 @@ func validateURL(pathURL string) error {
 	}
 	return nil
 }
+
+func validateHelm3Version(helmVersion string) bool {
```
rmweir: I think it would make more sense to return true if the version is valid or to name the function something like invalidHelm3Version

brendarearden: Sounds good - will do!

PR #26558, opened 2020-04-08
Fix kubelet needs to be restart if node is deleted in UI (#25812)

StrongMonkey:
Problem:
Today if node is deleted from rancher, we immdiately delete v1.node from
kubernetes. Since we don't clean up agent(or did a poor job in cleaning
agent). All the system containers keep running including kubelet. This
will cause kubelet keep throwing node not found error. This is because
kubelet only register new node at start up, and without restart kubelet
it will not re-register node again. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L52

Solution:
In our agent code, add kubelet restart logic. Notice that this logic is
only added into unmanaged agent(which is only added through docker run
command). This is because once managed agents are added, kubelet should
be ready since managed agents are pods from daemonset which mean the
node is schedulable now. Once managed agents are added unmanaged agents
will be cleaned up, then this logic will not be needed anymore.

### DIFF
--- a/pkg/agent/main.go
+++ b/pkg/agent/main.go
@@ -41,6 +41,10 @@ const (
 )
 
 func main() {
+	if _, err := reconcileKubelet(context.Background()); err != nil {
+		logrus.Warnf("failed to reconcile kubelet, error: %v", err)
+	}
+
 	logrus.SetOutput(colorable.NewColorableStdout())
 	logserver.StartServerWithDefaults()
 	if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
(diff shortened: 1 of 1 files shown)


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/23254

deniseschannon:
@StrongMonkey where's the follow up commit for v2.4 for this: https://github.com/rancher/rancher/pull/26557

deniseschannon:
Never mind. I see that you merged the 2 commits into 1 when backporting. 

Can you make 2 separate commits in this 1 PR so it's better tracked of what was changed in master as well?

StrongMonkey:
@deniseschannon Updated with two "revert" commits.

Issue #26440, opened 2020-04-02
[2.4.2] [Bug] Monitoring is not deployed correctly on new cluster creation.


mitchellmaler:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Create a new cluster using the terraform provider (might not be relevant) with monitoring enabled. 
**Result:**
The monitoring apps do not deploy correctly. Looking in the clusters System Apps both show this error:
```
Failed to install app cluster-monitoring. Error: Could not get apiVersions from Kubernetes: unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```

**Other details that may be helpful:**
Here are the rancher logs. It might be they are created before the CRDs are installed by the agent? If I kick them off again by forcing an upgrade they seem to install fine so seems like a timing issue. Seems like the monitoring-operator does automatically get kicked off again and deploys correctly, not sure if cluster-monitoring will.

```
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [INFO] Create app /
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [INFO] Create app /
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [INFO] clusterHandler: calling sync to create network policies for cluster c-4b2p4
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [ERROR] ClusterController c-4b2p4 [cluster-monitoring-handler] failed with : failed to get cattle-prometheus/prometheus-operated endpoints: endpoints "cattle-prometheus/prometheus-operated" not found
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [INFO] cluster [c-4b2p4] worker-upgrade: updating node [m-qj99n] with node-version 1
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [INFO] cluster [c-4b2p4] worker-upgrade: sending node-version for node [] version 1
rancher-687cf6dc6b-2m2pq rancher 2020/04/02 04:33:54 [INFO] clusterHandler: calling sync to create network policies for cluster c-4b2p4
[...]

PR #26564, opened 2020-04-08
[2.4] Force redeploy cluster-monitoring if workload doesn't exist

mrajashree:
https://github.com/rancher/rancher/issues/26440
Problem: Enabling rancher cluster-monitoring launches 2 apps, `cluster-monitoring` and `monitoring-operator`. If there's an error when installing these apps, monitoring controller has logic to force redeploy only the `monitoring-operator` app. So if the error is resolved, only `monitoring-operator` gets deployed but not `cluster-monitoring`

Solution: This commit adds the same logic to force deploy `cluster-monitoring` app, by checking whether the workload it should have created exists or not

### DIFF
--- a/pkg/controllers/user/monitoring/clusterHandler.go
+++ b/pkg/controllers/user/monitoring/clusterHandler.go
@@ -19,6 +19,7 @@ import (
 	mgmtv3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	v3 "github.com/rancher/types/apis/project.cattle.io/v3"
 	k8scorev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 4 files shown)


pkg/controllers/user/monitoring/clusterHandler.go
```diff
@@ -419,7 +420,17 @@ func (ch *clusterHandler) deployApp(appName, appTargetNamespace string, appProje
 		},
 	}
 
-	_, err = utils.DeployApp(ch.app.cattleAppClient, appDeployProjectID, app, false)
+	// redeploy cluster-monitoring App forcibly if its workload doesn't exist
+	var forceRedeploy bool
+	appWorkload, err := ch.app.agentStatefulSetClient.GetNamespaced(appTargetNamespace, fmt.Sprintf("prometheus-%s", appName), metav1.GetOptions{})
```
prachidamle, line 425: why should we use agentStatefulSetClient and not lister here?

pkg/controllers/user/monitoring/clusterHandler.go
```diff
@@ -419,7 +420,17 @@ func (ch *clusterHandler) deployApp(appName, appTargetNamespace string, appProje
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, errors.Wrapf(err, "failed to get statefulset %s/prometheus-%s", appTargetNamespace, appName)
+	}
+	if appWorkload == nil || appWorkload.Name == "" || appWorkload.DeletionTimestamp != nil {
```
prachidamle, line 429: Also checking apierrors.IsNotFound(err) as well would be decisive in most cases?

pkg/controllers/user/monitoring/clusterHandler.go
mrajashree: I followed what monitoring-operator did with deploymentClient, but it should be okay to use lister, switched to lister

pkg/controllers/user/monitoring/clusterHandler.go
mrajashree: Line 426 already checks for when error is not NotFound and returns it

prachidamle: Yes that is the assumption that appWorkload will be nil; but I have seen with client-go for ingress object that along with NotFound error the object returned is non-nil at times. Hence feel that also checking for the "error is NotFound" would be reliable. If your tests show that object returned is always nil, that is good.

prachidamle: Did you see a nil object always? @mrajashree

mrajashree: @prachidamle I was going by [this](https://github.com/rancher/types/blob/master/apis/apps/v1/zz_generated_stateful_set_controller.go#L116-L134) and it always returns nil if there's an error. 
But I also checked in my setup and the object is nil for NotFound error.

I think listers return nil with error, maybe not the client's [GetNamespaced](https://github.com/rancher/rancher/blob/master/vendor/github.com/rancher/norman/objectclient/object_client.go#L142-L158) call

prachidamle: great, thanks for digging it up

PR #26584, opened 2020-04-09
add rancher/calico-kube-controllers:v3.4.0

kinarashah:


scripts/istio-images-mirror
```diff
@@ -28,6 +28,7 @@ quay.io/jetstack/cert-manager-controller rancher/jetstack-cert-manager-controlle
 quay.io/calico/node rancher/calico-node v3.13.0
 quay.io/calico/cni rancher/calico-cni v3.13.0
 quay.io/calico/kube-controllers rancher/calico-kube-controllers v3.13.0
+quay.io/calico/kube-controllers rancher/calico-kube-controllers v3.4.0
```
deniseschannon, line 31: You need to only replace version in the same line as we just push the image after this is merged.

PR #26585, opened 2020-04-09
Enable auth during test run for istio

rancher-max:


tests/validation/tests/v3_api/conftest.py
```diff
@@ -16,6 +16,57 @@ def pytest_unconfigure(config):
...
+def get_tls(certificate):
+    if len(certificate) != 0:
+        tls = True
+    else:
+        tls = False
+    return tls
+
+
+@pytest.fixture(scope="module")
+def enable_ad():
```
izaac, line 29: What's the reasoning behind enabling AD from `conftest.py`?

rancher-max: I did it so that we can use the fixture from any test file we need it in

tests/validation/tests/v3_api/conftest.py
```diff
@@ -16,6 +16,57 @@ def pytest_unconfigure(config):
     if TEST_RBAC and CATTLE_TEST_URL:
         rbac_cleanup()
 
+
+def get_tls(certificate):
```
jiaqiluo, line 20: There is the same method in `test_auth.py`
why not move it to common.py?

tests/validation/tests/v3_api/conftest.py
izaac: there's an `enable_ad` function in `test_auth` we could either import that function in other test files or refactor `test_auth` to use this new fixture but that's harder as one common test case is to enable or **disable** the auth.

rancher-max: I did the suggested and used the function from `test_auth` but had to change it a little bit. Shouldn't affect the current tests in that file.

tests/validation/tests/v3_api/conftest.py
rancher-max: Resolved. I used that function

tests/validation/tests/v3_api/common.py
```diff
@@ -162,6 +162,10 @@
 # the shared password for all auth users
 AUTH_USER_PASSWORD = os.environ.get('RANCHER_AUTH_USER_PASSWORD', "")
 
+# Admin Auth Info
+AUTH_USERNAME = os.environ.get("RANCHER_AUTH_USERNAME")
```
izaac, line 166: @rancher-max in `test_auth` the `enable_ad` method makes use of admin_user which comes from the `v3_api/resource/*.json` files loaded in `test_auth`'s fixture. Do we we really need these variables or are these to set the build-in `admin` user password? @sangeethah

izaac: It seems the requirement in `test_auth` is to have the `AUTH` (AD/OpenLDAP/FreeIPA) admin already setup in the Rancher server, and these two variables are specific for Istio. Please correct me if I'm wrong.

rancher-max: We would definitely need at least password. 

Admin user could technically come from the `*.json` files but I wanted to allow it to be managed differently since our internal document for enabling AD auth setup uses a different admin user than the one noted in `v3_api/resource/activedirectory.json`

The other problem was that `test_auth.py` requires admin user and other auth users to have the same password, which I don't think is the case anymore for at least our AD setup so I needed to allow them to be separate.

rancher-max: I made this better now actually so it is just the password!

jiaqiluo: @rancher-max 
I saw this line multiple times when enabling the auth
```
AUTH_ADMIN_USER = load_setup_data()["admin_user"]
```

So we are not using this global variable `AUTH_ADMIN_USERNAME`? reserved for future use?

rancher-max: We could use that. This way grabs it from the *.json file. It might make more sense to grab it and then store it to use later

tests/validation/tests/v3_api/test_group_grb.py
```diff
@@ -305,7 +305,10 @@ def create_project_client(request):
     client, cluster = get_user_client_and_cluster()
     create_kubeconfig(cluster)
 
-    if AUTH_PROVIDER == "activeDirectory":
+    admin_client = get_admin_client()
+    ad_enabled = admin_client.by_id_auth_config("activedirectory").enabled
+
+    if AUTH_PROVIDER == "activeDirectory" and ad_enabled:
```
jiaqiluo, line 311: should it be `not ad_enabeld` ?

Issue #26583, opened 2020-04-09
[Forwardport] - Upgrade from Rancher 2.3.x -> 2.4.0 caused all Kubelets to constantly be restarted


sangeethah:
Forward port for - https://github.com/rancher/rancher/issues/26399

StrongMonkey:
It is already ported into master. Available to test in master head.

sangeethah:
https://github.com/rancher/rancher/commit/a6907ab73a595a05e5e80f84a4fe3efec440d5df https://github.com/rancher/rancher/commit/bfa3e1c8a45931cc138198be6323c2cff8867e22

jiaqiluo:
The bug fix is validated in Rancher:master-head 8bd924599 HA upgraded from Rancher:v2.3.6

Steps:
- run rancher:v2.3.6 in HA mode
- add two clusters (DO and ec2), each has 1 etcd, 1 control plan, and 2 workers
- upgrade rancher to master-head 8bd92459
- ssh to each node

Results:
- the container `kubelet` does not restart in any of those nodes
```
~$ docker ps | grep kubelet
b475af6168cc        rancher/hyperkube:v1.17.4-rancher1                "/opt/rke-tools/entr…"   30 minutes ago      Up 29 minutes                           kubelet
```
```
~$ docker inspect kubelet | grep RestartCount
        "RestartCount": 0,
```

jiaqiluo:
from https://github.com/rancher/rancher/issues/26399#issuecomment-616728493

Rancher version: `master-head (04/20/2020)` commit id: 8bd924599

- HA Install with self-signed certs.
- Upgrade from `v2.3.5`
- **All nodes checked and the kubelet wasn't restarting as reported in this issue.**

3 clusters tested

DO
- 1 etcd
- 1 controlplane
- 2 workers

Linode
- 1 etcd
- 1 controlplane
- 2 workers

Custom from EC2
- 1 etcd
- 1 controlplane
- 1 workers

PR #26618, opened 2020-04-13
Only restart kubelet once, do not check managed agent

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rancher/issues/23254. Talked with @ibuildthecloud and we kind of agree that restart just once at the beginning would be the best solution. This will cause kubelet to be restarted if `docker run` register command is run from the host.

jiaqiluo:
fix https://github.com/rancher/rancher/issues/26583

deniseschannon:
https://github.com/rancher/rancher/issues/26399

Issue #24652, opened 2019-12-17
API has editable fields for fields that are not-editable for EKS clusters


izaac:
EKS edit, I can change the subnets and I see the changes applied from Rancher but if I inspect the AWS EKS web console I do not see the Subnets edited.

**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

Rancher version 2.4 master-head (12/17/2019) e8fbc0c

1. I create a cluster with a VPC which has **4 subnets**.
2. Cluster gets created.
3. I am able to edit the cluster and the VPC can be selected now.
4. After selecting the VPC, I **remove** one of the subnets.
5. After saving I see the subnet I unselected in the previous step is no longer listed in the UI.
6. Save the security group.
7. Finally save settings.

After all that I now edit the cluster again and I see the cluster Rancher is listing **3 subnets**. But in AWS EKS web console I see the **original 4 subnets** from step 1. So this not seems to be applying the settings in the EKS cluster.

**Result:**

Changes made in the EKS cluster from Rancher are actually applied in the EKS cluster.

**Other details that may be helpful:**

https://github.com/rancher/rancher/issues/23879

**Environment information**
- Rancher version: `2.4 master-head (12/17/2019)` e8fbc0c
- Installation option (single install/HA): Single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): EKS 
- Machine type (cloud/VM/metal) and specifications (CPU/memory):  cloud
- Kubernetes version (use `kubectl version`):

```
1.14 (AWS)
```

daxmc99:
This is part of a larger issue with EKS in general. The current way we provision EKS clusters uses cloud formation and doesn't always clean up nicely. 

Per https://github.com/rancher/rancher/issues/23879#issuecomment-566785911
>
>The backend should change the schemas to not be updatable if they are not available to >update.

This is my planned course of action with this issue. These fields will no longer be editable on the backend

deniseschannon:
Overall, the issue is that the schema makes it seem like you can edit fields that you can't. 

We've updated the UI in 2.4 to make these fields read only, but since the subnet is being saved as `null` and being returned from AWS as `null`, there is a UI fix to just send an empty array instead of properly fixing the backend. 

When we fix the API schema to make these fields not editable we should also investigate why the API is returning a null value for when an empty array is sent. Also the field is non-nullable, which makes it extra weird.

[...]

PR #26640, opened 2020-04-15
Add EKS specific subnet validation

daxmc99:
Previously, subnets were editable on EKS clusters which does not represent the architecture of
Kontainer Driver which does not allow for editing of subnets after the cluster has been provisioned.

This was done as a validator due to the difficulty of adding additional fields to the kontainer
engine type since this requires regenerating the protobuff files for Kontainer Driver.

Addresses https://github.com/rancher/rancher/issues/24652

### DIFF
--- a/pkg/api/customization/cluster/validator.go
+++ b/pkg/api/customization/cluster/validator.go
@@ -3,9 +3,11 @@ package cluster
 import (
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 	"time"
 
+	"github.com/rancher/kontainer-engine/service"
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
@@ -54,6 +56,10 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 	}
 
 	if err := v.validateScheduledClusterScan(&spec); err != nil {
+		return err
+	}
+
+	if err := v.validateGenericEngineConfig(request, &spec); err != nil {
 		return err
 	}
 	return nil
@@ -268,5 +274,49 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
 		return err
 	}
 
+	return nil
+}
+
+// validateGenericEngineConfig allows for additional validation of clusters that depend on Kontainer Engine or Rancher Machine driver
+func (v *Validator) validateGenericEngineConfig(request *types.APIContext, spec *v3.ClusterSpec) error {
+
+	if request.Method == http.MethodPost {
+		return nil
+	}
+
+	if spec.AmazonElasticContainerServiceConfig != nil {
+		// compare with current cluster
+		clusterName := request.ID
+		prevCluster, err := v.ClusterLister.Get("", clusterName)
+		if err != nil {
+			return httperror.WrapAPIError(err, httperror.InvalidBodyContent, err.Error())
+		}
+
+		err = validateEKS(*prevCluster.Spec.GenericEngineConfig, *spec.AmazonElasticContainerServiceConfig)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+
+}
+
+func validateEKS(prevCluster, newCluster map[string]interface{}) error {
+	// check config is for EKS clusters
+	if driver, ok := prevCluster["driverName"]; ok {
+		if driver != service.AmazonElasticContainerServiceDriverName {
+			return nil
+		}
+	}
+
+	// don't allow for updating subnets
+	if prev, ok := prevCluster["subnets"]; ok {
+		if new, ok := newCluster["subnets"]; ok {
+			if !reflect.DeepEqual(prev, new) {
+				return httperror.NewAPIError(httperror.InvalidBodyContent, "cannot modify EKS subnets after creation")
+			}
+		}
+	}
 	return nil
 }
--- a/tests/integration/suite/test_cluster_defaults.py
+++ b/tests/integration/suite/test_cluster_defaults.py
@@ -55,6 +55,51 @@ def test_generic_initial_conditions(admin_mc, remove_resource):
     assert 'exportYaml' not in cluster.actions
 
 
+def test_eks_cluster_immutable_subnets(admin_mc, remove_resource):
+    cluster = admin_mc.client.create_cluster(
+        name=random_str(), amazonElasticContainerServiceConfig={
+            "accessKey": "asdfsd",
+            "secretKey": "verySecretKey",
+            "subnets": [
+                "subnet-045bfaeca7d3f1cb3",
+                "subnet-02388a166136f98c4"
+            ]})
+    remove_resource(cluster)
+
+    # try to edit cluster subnets
+    with pytest.raises(ApiError) as e:
+        admin_mc.client.update_by_id_cluster(
+            id=cluster.id,
+            amazonElasticContainerServiceConfig={
+                 "accessKey": "asdfsd",
+                 "secretKey": "verySecretKey",
+                 "subnets": [
+                     "subnet-045bfaeca7d3f1cb3"
+                 ]})
+
+    assert e.value.error.status == 422
+    assert e.value.error.message == 'cannot modify EKS subnets after creation'
+
+    # tests updates still work
+    new = admin_mc.client.update_by_id_cluster(
+       id=cluster.id,
+       name=cluster.name,
+       description="update",
+       amazonElasticContainerServiceConfig={
+           # required field when updating KE clusters
+           "driverName": "amazonelasticcontainerservice",
+           "accessKey": "asdfsd",
+           "secretKey": "verySecretKey",
+           "subnets": [
+               "subnet-045bfaeca7d3f1cb3",
+               "subnet-02388a166136f98c4"
+           ]})
+
+    assert new.id == cluster.id
+    assert not hasattr(cluster, "description")
+    assert hasattr(new, "description")
+
+
 def test_rke_initial_conditions(admin_mc, remove_resource):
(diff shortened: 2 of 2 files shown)


pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +276,79 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+// validateGenericEngineConfig allows for additional validation of clusters that depend on Kontainer Engine or Rancher Machine driver
+func (v *Validator) validateGenericEngineConfig(request *types.APIContext, spec *v3.ClusterSpec) error {
+
+	if request.Method == http.MethodPost {
+		return nil
+	}
+
+	clusterName := request.ID
+	fmt.Printf("%v", time.Now())
+	fmt.Printf("%v", spec.AmazonElasticContainerServiceConfig)
```
rmweir, line 289: Should remove these prints

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +276,79 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+
+	// eks specific logic
+	prevCluster, err := v.ClusterLister.Get("", clusterName)
+	if err != nil {
+		if errors.IsNotFound(err) { //nothing to do, cluster is being created
```
rmweir, line 294: How would the cluster not exist?

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +276,79 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+			if !ok {
+				return fmt.Errorf("unable to convert subnets to interace slice")
+			}
+			newSlice, ok := new.([]interface{})
+			if !ok {
+				return fmt.Errorf("unable to convert subnets to interface slice")
+			}
+
+			if !testEq(prevSlice, newSlice) {
+				return fmt.Errorf("cannot modify EKS subnets after creation")
```
rmweir, line 331: validators should return API errors with code, these would probably be "invalid body content"

rmweir: also, this is probably better but should note that if we were using norman tags (ik they can't be for this) I believe the fields would just be dropped and the update would go through

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +276,79 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+			}
+		}
+	}
+	return nil
+}
+
+func testEq(a, b []interface{}) bool {
```
rmweir, line 338: nit: would the subnets ever be out of order? could use a map

daxmc99: It’s also a string slice at the EKS level. I believe if a update ever got through it would stay in an endless update loop between the actual cluster and the proposed one. Subnets should be considered immutable after provisioning.

pkg/api/customization/cluster/validator.go
daxmc99: Or first request to create the cluster

rmweir: Wouldn't that be a Post then?

daxmc99: Dropped this err check

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +275,72 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+			return nil
+		}
+	}
+
+	// don't allow for updating subnets
+	if prev, ok := prevCluster["subnets"]; ok {
+		if new, ok := newCluster["subnets"]; ok {
+			prevSlice, ok := prev.([]interface{})
+			if !ok {
+				return httperror.NewAPIError(httperror.InvalidBodyContent, "unable to convert subnets to interace slice")
```
cbron, line 315:
```suggestion
				return httperror.NewAPIError(httperror.InvalidBodyContent, "unable to convert subnets to interface slice")
```

cbron: This is a weird case too, that those subnets exist but can't be typecast ? I almost wonder if we should just return nil here, because its not what we expected.

daxmc99:
> LGTM, Is it possible to add integration test for this case? @daxmc99 

Hmm, this somewhat depends on if we have a test that creates an EKS cluster. Would a go test work here?

pkg/api/customization/cluster/validator.go
daxmc99: I think this provides a useful case for clients and api users. Although I do find it hard to imagine a malformed request will make it this far, since it has to have passed the object conversion steps.

cbron: In this case it's not the request thats malformed though, its the original cluster. Trying to catch the case where perhaps there is some old cluster out there that never got this type set on it or something.

prachidamle:
> > LGTM, Is it possible to add integration test for this case? @daxmc99
> 
> Hmm, this somewhat depends on if we have a test that creates an EKS cluster. Would a go test work here?

@daxmc99 It need not be a functional EKS cluster; as long as it is in provisioning state we could test that the subnet data cannot be changed? Create an EKS cluster as in here with some subnet data and try to edit that? https://github.com/rancher/rancher/blob/master/tests/integration/suite/test_cluster_defaults.py#L40-L43

tests/integration/suite/test_cluster_defaults.py
```diff
@@ -55,6 +55,32 @@ def test_generic_initial_conditions(admin_mc, remove_resource):
     assert 'exportYaml' not in cluster.actions
 
 
+def test_eks_cluster_immutable_subnets(admin_mc, remove_resource):
```
rmweir, line 58: should add test case where update has same subnets

pkg/api/customization/cluster/validator.go
dramich: Two parts here.
1. The error is not a useful error to a user. If I got that error what would I do with it? At some point that value was put into an AWS config, which means it was valid and even if it's empty it would get the zero value for the field, an empty array. 
2. I don't think we should block the update if Rancher did something bad in the past with this field.

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +275,72 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+			}
+			newSlice, ok := new.([]interface{})
+			if !ok {
+				return httperror.NewAPIError(httperror.InvalidBodyContent, "unable to convert subnets to interace slice")
```
dramich, line 319: Same comment as above. Has it been tested that we can shove through something that would actually hit this error? The test cases below don't test that.

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +275,72 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+
+			if !testEq(prevSlice, newSlice) {
+				return httperror.NewAPIError(httperror.InvalidBodyContent, "cannot modify EKS subnets after creation")
+			}
+		}
+	}
+	return nil
+}
+
+func testEq(a, b []interface{}) bool {
```
dramich, line 330: Why are we rewriting `DeepEqual`?

pkg/api/customization/cluster/validator.go
```diff
@@ -270,3 +275,72 @@ func (v *Validator) accessTemplate(request *types.APIContext, spec *mgmtclient.C
...
+
+// validateGenericEngineConfig allows for additional validation of clusters that depend on Kontainer Engine or Rancher Machine driver
+func (v *Validator) validateGenericEngineConfig(request *types.APIContext, spec *v3.ClusterSpec) error {
+
+	if request.Method == http.MethodPost {
+		return nil
+	}
+
+	clusterName := request.ID
+	prevCluster, err := v.ClusterLister.Get("", clusterName)
```
dramich, line 287: This should move after the check for the `AmazonElasticContainerServiceConfig`. No reason to error here if this is the wrong driver anyways.

dramich:
Why are we not doing this in KE where we just added other comparison logic? https://github.com/rancher/kontainer-engine/blob/release/v2.4/drivers/eks/eks_driver.go#L978

daxmc99:
This change really is to make the API more inline with what the UI is doing by setting these values to read-only. We really don't want to ever update the subnet, even if it isn't applied

pkg/api/customization/cluster/validator.go
daxmc99: Avoids the use of reflection when we know the type

pkg/api/customization/cluster/validator.go
daxmc99: I think it is unlikely this will occur but I would prefer to catch it instead of letting it panic. Returning no error might allow a potentially badly form request to get further through the api

pkg/api/customization/cluster/validator.go
daxmc99: moved

pkg/api/customization/cluster/validator.go
cbron: Was there an outcome to this ?

daxmc99: I think we decided on not catching errors here and just returning nil

pkg/api/customization/cluster/validator.go
daxmc99: Switched to use `reflect.DeepEqual`

pkg/api/customization/cluster/validator.go
daxmc99: Dropped

Issue #26505, opened 2020-04-06
Add 'cert-manager.io/issuer-kind: Issuer' to ingress.yaml in Rancher chart to mitigate https://github.com/jetstack/cert-manager/issues/2533 


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Enhancement

**Steps to reproduce (least amount of steps as possible):**

If a user configures a default issuer when installing cert-manager into the Kubernetes cluster during the installation of Rancher v2.x HA they will encounter the cert-manager issue documented at https://github.com/jetstack/cert-manager/issues/2533 when installing the Rancher chart. This can be mitigated by updating the ingress.yaml template in the Rancher chart to include `cert-manager.io/issuer-kind: Issuer` per the below:

```
{{- if or (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") (and (eq (int $certmanagerVer._0) 0) (ge (int $certmanagerVer._1) 11)) }}
    cert-manager.io/issuer: {{ template "rancher.fullname" . }}
    cert-manager.io/issuer-kind: Issuer
```

**Environment information**
- Rancher v2.3.5

cbron:
Do we want to backport to 2.2 ?

rancher-max:
#### Reproduced in v2.3.5 and in v2.x-head (where x is all of 2, 3 and 4) and in master-head

Steps:
1. Setup kubernetes cluster for HA: 
`rke up --config ./rancher-cluster.yml`
2. After everything is running, grab the latest rancher helm chart: 
`helm repo add rancher-latest https://releases.rancher.com/server-charts/latest`
3. Install cert-manager with a default ClusterIssuer:
```
$ kubectl apply -f https://raw.githubusercontent.com/jetstack/cert-manager/release-0.12/deploy/manifests/00-crds.yaml && \
kubectl create namespace cert-manager && \
helm repo add jetstack https://charts.jetstack.io && helm repo update

$ helm install \
  cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --version v0.12.0 \
  --set ingressShim.defaultIssuerName=letsencrypt-staging \
  --set ingressShim.defaultIssuerKind=ClusterIssuer \
  --set ingressShim.defaultIssuerGroup=cert-manager.io
```
4. After all cert-manager pods are running, apply a default clusterissuer:
`kubectl apply -f clusterissuer.yaml`
```yaml
# clusterissuer.yaml
[...]

Issue #2533, opened 2015-11-04
[ui] Service links disappear when switching environment


ndelitski:
Open environment A in new tab(links are shown), switch to B(no links). Reload page. There are links in B, switch to A(no links again)
```
Rancher	v0.43.1
Cattle	v0.106.0
User Interface	v0.63.0
Rancher Compose	v0.5.0
```

deniseschannon:
This is a duplicate of this. https://github.com/rancher/rancher/issues/1389

Sometimes links aren't showing up when there are multiple environments.

PR #26655, opened 2020-04-16
Updated rancher chart to mitigate cert-manager issue #2533

rawmind0:
Fix issue #26505

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -19,6 +19,7 @@ metadata:
     {{- $certmanagerVer :=  split "." .Values.certmanager.version -}}
     {{- if or (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") (and (eq (int $certmanagerVer._0) 0) (ge (int $certmanagerVer._1) 11)) }}
     cert-manager.io/issuer: {{ template "rancher.fullname" . }}
+    cert-manager.io/issuer-kind: Issuer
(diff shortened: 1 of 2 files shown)


chart/templates/ingress.yaml
```diff
@@ -19,6 +19,7 @@ metadata:
     {{- $certmanagerVer :=  split "." .Values.certmanager.version -}}
     {{- if or (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") (and (eq (int $certmanagerVer._0) 0) (ge (int $certmanagerVer._1) 11)) }}
     cert-manager.io/issuer: {{ template "rancher.fullname" . }}
+    cert-manager.io/issuer-kind: Issuer
     {{- else if or (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") (and (eq (int $certmanagerVer._0) 0) (lt (int $certmanagerVer._1) 11)) }}
     certmanager.k8s.io/issuer: {{ template "rancher.fullname" . }}
```
cbron, line 24: This appears to check if it's below version 0.11, and issues says its 0.11, so confirming that we don't need to add here right ?

rawmind0: It checks if cert-manager version is greater or equal than 0.11 to add the `cert-manager.io/issuer-kind` annotation

rawmind0:
@cbron , updated ingress_test.yaml to pass ci

PR #26658, opened 2020-04-16
[2.3] Updated rancher chart to mitigate cert-manager issue #2533

rawmind0:
Fix issue #26505


maggieliu:
related to: https://github.com/rancher/rancher/issues/27085

Issue #26605, opened 2020-04-10
User is not able to create a PVC using a PV created in Longhorn


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster - RKE do - 1 node
- Deploy Longhorn app in the Default project
- Navigate to the Longhorn UI once it is deployed
- Deploy a Volume and create a PV say pv1
- Navigate to Rancher UI --> Cluster --> Default Project --> Resources --> Volumes
- Create a volume
- Select "Use an existing persistent volume".
- From the dropdown select pv1
- Enter other details and Click on Save.

**Actual Result:**
Error is seen `Error: storageclasses.storage.k8s.io "longhorn-static" not found` and PVC is not created.

**Expected Result:**
PVC should be created successfully

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.2
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): rke do
- Kubernetes version (use `kubectl version`):

```
1.17.4
```

yasker:
I don't think we should check for the existence of StorageClass object here. StorageClass doesn't need to exist for PV and PVC to match. It will function as a label for matching if it doesn't exist as an object.

yasker:
Example to explain and reproduce the issue without Longhorn, using following two yamls.

```
# random_pv.yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: random-pv
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 2Gi
  hostPath:
    path: /opt/test-tmp
    type: DirectoryOrCreate
  persistentVolumeReclaimPolicy: Delete
  storageClassName: random-storage-class
```
```
# random_pvc.yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: random-pvc
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
  storageClassName: random-storage-class
  volumeMode: Filesystem
```

In Kubernetes, create `random_pv.yaml`, then create `random_pvc.yaml`. Now you can see PVC has been bound to PV.

[...]

PR #26660, opened 2020-04-16
No need to check for storageClass existence while creating a pvc from pv

prachidamle:
https://github.com/rancher/rancher/issues/26605
Dropping the api validation to check if storageClass exists while creating a pvc from an existing pv, since it is not needed and the resources can be created without an actual storageclass existing.

### DIFF
--- a/pkg/api/customization/persistentvolumeclaim/validator.go
+++ b/pkg/api/customization/persistentvolumeclaim/validator.go
@@ -7,6 +7,7 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/rancher/pkg/api/store/storageclass"
 	"github.com/rancher/rancher/pkg/clustermanager"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -28,7 +29,10 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 
 	storageClass, err := c.Storage.StorageClasses("").Get(storageClassID, v1.GetOptions{})
 	if err != nil {
-		return err
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		return nil
 	}
 
 	// if the referenced storage class does not have a storageaccounttype, storage account creation will fail in k8s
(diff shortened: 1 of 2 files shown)


pkg/api/customization/persistentvolumeclaim/validator.go
```diff
@@ -27,10 +28,14 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 	}
 
 	storageClass, err := c.Storage.StorageClasses("").Get(storageClassID, v1.GetOptions{})
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 
+	if apierrors.IsNotFound(err) || storageClass == nil {
```
mrajashree, line 35: We don't need storageClass only if the specified persistentVolume exists right, so before returning nil should we add a check to see if the PV corresponding to the `volumeId` from the request exists or not?

prachidamle: No a persistentVolumeClaim can be created without either the storageClass or an existing persistentVolume - both of these are used in order to match the pvc to a pv if it exists. If none of them are specified, the pvc remains in "pending"/"unbound" state.

prachidamle: I added an integration test for this case.

prachidamle:
Added an integration test to check pvc creation goes through even if no storageClass or VolumeId is provided.

tests/integration/suite/test_persistent_volume_claim.py
```diff
@@ -104,3 +104,27 @@ def test_can_create_azure_any_accountstoragetype(admin_pc, admin_cc,
             }
         })
     remove_resource(pvc2)
+
+
+def test_can_create_pvc_no_storage_no_vol(admin_pc, admin_cc,
```
rmweir, line 109: nit: looks like you're not using admin_cc or admin_mc, so they can be removed

pkg/api/customization/persistentvolumeclaim/validator.go
```diff
@@ -27,10 +28,14 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 	}
 
 	storageClass, err := c.Storage.StorageClasses("").Get(storageClassID, v1.GetOptions{})
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
```
dramich, line 31: A nil chick on the `storageClass` isn't needed, if there is an error process accordingly, otherwise we assume we have a real thing. 

```suggestion
	if err != nil {
		if !apierrors.IsNotFound(err) {
			return err
		}
		return nil
	}
```

PR #26664, opened 2020-04-16
Add Oracle Container Engine (OKE) as a builtin cluster driver.

jlamillan:
This pull request adds the most recent [Oracle Container Engine (OKE) cluster driver](https://github.com/rancher-plugins/kontainer-engine-driver-oke) to be `builtin` (but `inactive` by default for the time being). 

Notes:
- Link to a related [PR in the UI repo](https://github.com/rancher/ui/pull/3931) also up for review.

### DIFF
--- a/app/kontainerdriver_data.go
+++ b/app/kontainerdriver_data.go
@@ -88,6 +88,16 @@ func addKontainerDrivers(management *config.ManagementContext) error {
 	); err != nil {
 		return err
 	}
+	if err := creator.addCustomDriver(
+		"oraclecontainerengine",
+		"https://github.com/rancher-plugins/kontainer-engine-driver-oke/releases/download/v1.4.0/kontainer-engine-driver-oke-linux",
+		"a42bd8b14e780edc426119f143acbc54a93a07f8b62f193b9e261ddd85db6e9c",
+		"",
+		false,
+		"*.oraclecloud.com",
+	); err != nil {
+		return err
+	}
 
 	return nil
 }
--- a/tests/integration/suite/test_rbac.py
+++ b/tests/integration/suite/test_rbac.py
@@ -471,14 +471,14 @@ def ns_count(client, count):
 
 def test_appropriate_users_can_see_kontainer_drivers(user_factory):
     kds = user_factory().client.list_kontainer_driver()
-    assert len(kds) == 8
+    assert len(kds) == 9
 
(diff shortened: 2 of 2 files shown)


app/kontainerdriver_data.go
```diff
@@ -88,6 +88,16 @@ func addKontainerDrivers(management *config.ManagementContext) error {
 	); err != nil {
 		return err
 	}
+	if err := creator.addCustomDriver(
+		"oraclecontainerengine",
+		"https://github.com/rancher-plugins/kontainer-engine-driver-oke/releases/download/v1.3.0/kontainer-engine-driver-oke-linux",
+		"",
```
dramich, line 94: A checksum needs to be added, this is how Rancher knows we have downloaded the correct driver and not an imposter.

app/kontainerdriver_data.go
```diff
@@ -88,6 +88,16 @@ func addKontainerDrivers(management *config.ManagementContext) error {
...
+		"https://cdn.jsdelivr.net/gh/rancher-plugins/ui-cluster-driver-oke@v1.3.2/dist/component.js",
+		false,
+		"*.jsdelivr.net",
```
dramich, line 97: Is this an oracle controller domain? The whitelist is to ensure we don't download random content from the internet and attempt to run it. If this is oracle controlled then it can be updated to `cdn.jsdelivr.net` and if not then it needs to be moved somewhere oracle controlled.

app/kontainerdriver_data.go
jlamillan: Ill add a checksum.

app/kontainerdriver_data.go
jlamillan: Spoke with @vincent99 this morning about this. We decided to avoid the need of a CDN by moving the UI component into the [Rancher UI repository](https://github.com/rancher/ui/).

jlamillan: checksum added

jlamillan:
> This looks fine but I want to wait to merge until we get confirmation that the UI won't need access to an oracle API to get values that are currently hard coded.

@dramich 

@vincent99 and I agreed that pulling values from the Oracle API is desirable, but not required and therefore will be submitted as a subsequent enhancement (soon).

deniseschannon:
Approved based on the decision to move pulling from Oracle API directly in a later PR.

Issue #26586, opened 2020-04-09
Panic after adding a worker node to custom cluster


soumyalj:
OS: Ubuntu 18.04

Steps :
1. Deploy a custom cluster in 2.0.10 with k8s 1.11.6-rancher1-1 and run pre upgrade checks
2. Upgrade from rancher v2.0.10 to rancher v2.2.9 and run post upgrade checks
3. Upgrade to rancher 2.2.11 and run post upgrade checks
4. Upgrade k8s to 1.14.10 and run post upgrade checks and also add a node
5. Upgrade to 2.4.2 and run post upgrade checks and add a node

The node addition failed and was stuck in "Registering" state :
```
2020/04/09 21:38:08 [ERROR] NodeController c-dkzqr/m-db790889d42b [rke-worker-upgrader] failed with : getNodePlan error for node [m-db790889d42b]: failed to find plan for abcdef
2020/04/09 21:38:08 [ERROR] NodeController c-dkzqr/m-db790889d42b [rke-worker-upgrader] failed with : getNodePlan error for node [m-db790889d42b]: failed to find plan for abcdef
2020/04/09 21:38:09 [INFO] Provisioning cluster [c-dkzqr]
2020/04/09 21:38:09 [INFO] Updating cluster [c-dkzqr]
2020/04/09 21:38:09 [ERROR] NodeController c-dkzqr/m-db790889d42b [rke-worker-upgrader] failed with : getNodePlan error for node [m-db790889d42b]: failed to find plan for abcdef
```
Kubelet does not start on this node. 
```
ubuntu@ip-172-31-11-123:~$ sudo docker ps -a
CONTAINER ID        IMAGE                                 COMMAND                  CREATED             STATUS              PORTS               NAMES
ef2cdc5bb2ea        rancher/hyperkube:v1.14.10-rancher1   "/opt/rke-tools/entr…"   2 hours ago         Up 2 hours                              kube-proxy
9879f038b419        rancher/rancher-agent:v2.4.2          "run.sh --server htt…"   2 hours ago         Up 2 hours                              zen_moore
d49ea28ae1aa        rancher/rke-tools:v0.1.50             "nginx-proxy CP_HOST…"   2 hours ago         Up 2 hours                              nginx-proxy
24edb76888db        rancher/rancher-agent:v2.4.2          "run.sh -- share-roo…"   2 hours ago         Up 2 hours                              share-mnt
[...]

PR #26685, opened 2020-04-17
check for nil upgradeStrategy

kinarashah:
We initialize node upgrade strategy in cluster store, but on rancher upgrade if a cluster automatically gets upgraded, it'll panic because of nil upgradeStrategy

https://github.com/rancher/rancher/issues/26586

### DIFF
--- a/pkg/controllers/management/rkeworkerupgrader/upgrade.go
+++ b/pkg/controllers/management/rkeworkerupgrader/upgrade.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/management/clusterprovisioner"
 	nodeserver "github.com/rancher/rancher/pkg/rkenodeconfigserver"
 	"github.com/rancher/rancher/pkg/systemaccount"
+	rkedefaults "github.com/rancher/rke/cluster"
 	"github.com/rancher/rke/util"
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -228,12 +229,25 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string, p
 	uh.clusterLock.Lock(clusterName)
 	defer uh.clusterLock.Unlock(clusterName)
 
-	logrus.Debugf("cluster [%s] upgrading condition: [%v] plan changed: [%v]", clusterName, v3.ClusterConditionUpgraded.IsUnknown(cluster), planChanged)
-	if !v3.ClusterConditionUpgraded.IsUnknown(cluster) || planChanged {
+	upgradeStrategy := cluster.Status.AppliedSpec.RancherKubernetesEngineConfig.UpgradeStrategy
+	toUpdate := !v3.ClusterConditionUpgraded.IsUnknown(cluster) || planChanged || upgradeStrategy == nil
```
mrajashree, line 233: I don't think we need to check for `upgradeStrategy == nil` while evaluating `toUpdate`, the check for `if upgradeStrategy == nil {` on line 243 should suffice?

PR #26702, opened 2020-04-18
Changes to get test_auth working with AD auth provider

izaac:
The original structure and logic is untouched.
Most of the work was to remove some repetitive code like `enable_ad_nestedgroups` and `enable_openldap_nestedgroups`.

Pre-requirements are now automatically provisioned if they are not prepared. Including enabling auth and provisioning two clusters.
Updated the requirements file to be compatible with both `pip` and `pipenv` because `pipenv` is more strict and required `#egg=<app name>`

Everything else is just renaming to follow pep8 rules and making sure `flake8` report is ok.

### DIFF
--- a/tests/validation/tests/v3_api/test_auth.py
+++ b/tests/validation/tests/v3_api/test_auth.py
@@ -1,13 +1,24 @@
+import json
+import os
 import pytest
 import requests
-import os
 
-from .common import *  # NOQA
+from .common import ADMIN_TOKEN
+from .common import AUTH_PROVIDER
+from .common import AUTH_USER_PASSWORD
+from .common import CATTLE_TEST_URL
+from .common import cluster_cleanup
+from .common import create_project_and_ns
+from .common import get_admin_client
+from .common import get_client_for_token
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_auth.py
```diff
@@ -815,10 +765,16 @@ def create_project_client(request):
...
         client.delete(setup["project1"])
         client.delete(setup["project2"])
         delete_cluster_users()
+        if cluster_total == 0:
+            if aws_nodes1 and aws_nodes2:
+                cluster_cleanup(client, cluster1, aws_nodes1)
+                cluster_cleanup(client, cluster2, aws_nodes2)
+        if cluster_total == 1:
+            if aws_nodes1:
+                cluster_cleanup(client, cluster1, aws_nodes1)
```
rancher-max, line 777: I think this is good, but do we potentially want a way to keep the clusters around?

izaac: in that case just set `RANCHER_CLEANUP_CLUSTER` to false 👍

Issue #19615, opened 2019-04-15
Automatically lower case rancher catalog template names 


Amos-85:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
add jenkins-x catalog url: https://github.com/jenkins-x/jenkins-x-platform

exception in rancher logs:
`[ERROR] CatalogController jenkinsx [catalog] failed with : failed to update templates. Multiple error occurred: [TemplateVersion.management.cattle.io "jenkinsx-jenkins-x-platform-0.0.1-SNAPSHOT" is invalid: metadata.name: Invalid value: "jenkinsx-jenkins-x-platform-0.0.1-SNAPSHOT": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]
2019/04/15 09:07:59 [INFO] Catalog sync done. 0 templates created, 4 templates updated, 0 templates deleted`

**Result:** catalog doesn't sync when chart version in chart.yaml is in uppercase characters the issue reproduced.

**expected behavior :**
manipulate chart.yaml version to lower case to avoid the issue.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.1.8 stable
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type: Hosted
- Machine type: Metal
- Kubernetes version: 1.13.5

Amos-85:
@deniseschannon ,
is there any estimation to fix this issue?
sill happen in version 2.2.8

stearz:
This Pull Request should fix the problem:
https://github.com/rancher/rancher/pull/26703

stearz:
New PR https://github.com/rancher/rancher/pull/26943

rancher-max:
#### Reopening due to unable to deploy app

The catalog is now able to be added, but when navigating to Apps > Launch and trying to deploy the app from this catalog, it fails with the same error:
<img width="302" alt="image" src="https://user-images.githubusercontent.com/55997940/81445885-9c98bc80-912e-11ea-84aa-900bb918aed5.png">

This happens in both `master-2939-head` and `v2.4-2941-head`

rancher-max:
[...]

PR #26703, opened 2020-04-18
Support uppercase characters in helm chart version

stearz:
SemVer supports uppercase characters, so I updated the regex. Would be great if you could merge the pull request as it blocks me from using rancher in production


brendarearden:
The original error in the related issue `a DNS-1123 subdomain must consist of lower case alphanumeric
characters, '-' or '.', and must start and end with an alphanumeric
character (e.g. 'example.com', regex used for validation is
'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]` is an error being thrown by the kubernetes api as the chart is being stored as a kubernetes object for rancher to use - I don't believe this PR will resolve the root cause, but adding a tolower here might be the result we are looking for https://github.com/rancher/rancher/blob/6a60f3386484e841480b3686e2cd6b189fa77b55/pkg/catalog/manager/catalog_common.go#L215 or adding some code here to ensure the version will work as a subdomain in kubernetes 
https://github.com/rancher/rancher/blob/6a60f3386484e841480b3686e2cd6b189fa77b55/pkg/catalog/manager/catalog_preprocess.go

stearz:
I put a `ToLower` in `getValidTemplateNameWithVersion` as there exist other conversions to return a valid label.

stearz:
Is there anything I can do to get this PR merged?

brendarearden:
the root cause will be fixed by the tolower so we should not need to update the regex. Can we get the regex and tests reverted and then commits squashed?

maggieliu:
related to https://github.com/rancher/rancher/issues/19615

stearz:
successor to this PR is here https://github.com/rancher/rancher/pull/26943

Issue #26562, opened 2020-04-08
Add Customizable Header/Footer/'Classification Banners' to Rancher - Backend


maggieliu:
Original ticket: https://github.com/rancher/rancher/issues/25694

kinarashah:
removed to/test label, needs to be backported to release/v2.4

kinarashah:
Available to test in `2.4.3-rc1`

jiaqiluo:
The enhancement is validated on v2.4.3-rc1 with change UI index to `https://releases.rancher.com/ui/latest-2.4/index.html` 

Now we can set the header and banner and showing either of them or both 
 
```
{
  "banner": {
    "textColor": "#313131",
    "background": "#27AE5F",
    "text": "Testing banner and header on v2.4.3-rc1, with UI index latest-2.4"
  },
  "showHeader": "true",
  "showFooter": "true"
}
```
Result:

![Screenshot_2020-04-21 Rancher](https://user-images.githubusercontent.com/6218999/79927734-87652380-83f5-11ea-8cc4-7adefcbee8d1.png)

PR #26706, opened 2020-04-19
Add ui-banners setting

aiyengar2:
This PR adds the ui-banners setting to Rancher, which will be used to facilitate customizable header / footer banners on the Rancher UI.

Note: GET requests to the setting are unauthenticated so that they can be queried from the login page.

Related Issue: https://github.com/rancher/rancher/issues/26562


aiyengar2:
last force push was a nit fix which moved the UI banner setting with the other UI settings

Issue #21744, opened 2019-07-24
Logging: AWS Elasticsearch: Cannot get new connection from pool


nlowe:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create an Elasticsearch Domain in AWS
2. Configure Rancher Cluster Logging to ship to the Elasticsearch endpoint

**Result:**

After some period of time, new logs stop showing up in Elasticsearch. Checking the logs of the fluentd container shows the following error:

```ruby
2019-07-24 06:58:22 +0000 [warn]: #0    failed to flush the buffer. retry_time=9 next_retry_seconds=2019-07-24 07:02:13 +0000 chunk="58e67ce51dc61232b625376ad93f6e7a" error_class=Elasticsearch::Transport::Transport::Error error="Cannot get new connection from pool."
  2019-07-24 06:58:22 +0000 [warn]: #0    /var/lib/gems/2.3.0/gems/elasticsearch-transport-6.1.0/lib/elasticsearch/transport/transport/base.rb:258:in `perform_request'
  2019-07-24 06:58:22 +0000 [warn]: #0    /var/lib/gems/2.3.0/gems/elasticsearch-transport-6.1.0/lib/elasticsearch/transport/transport/http/faraday.rb:20:in `perform_request'
  2019-07-24 06:58:22 +0000 [warn]: #0    /var/lib/gems/2.3.0/gems/elasticsearch-transport-6.1.0/lib/elasticsearch/transport/client.rb:131:in `perform_request'
  2019-07-24 06:58:22 +0000 [warn]: #0    /var/lib/gems/2.3.0/gems/elasticsearch-api-6.1.0/lib/elasticsearch/api/actions/bulk.rb:95:in `bulk'
  2019-07-24 06:58:22 +0000 [warn]: #0    /var/lib/gems/2.3.0/gems/fluent-plugin-elasticsearch-2.12.5/lib/fluent/plugin/out_elasticsearch.rb:639:in `send_bulk'
  2019-07-24 06:58:22 +0000 [warn]: #0    /var/lib/gems/2.3.0/gems/fluent-plugin-elasticsearch-2.12.5/lib/fluent/plugin/out_elasticsearch.rb:541:in `block in write'
[...]

PR #26734, opened 2020-04-21
[Master] Set reload_connections to false by default for es

loganhz:
2.4 PR https://github.com/rancher/rancher/pull/26733

Issue:

fluent-plugin-elasticsearch reloads connection after 10000 requests. Sometimes this reloading functionality bothers users to send events with ES plugin. Fluentd will stop to send logs to ES.

Solution:

Set it to false by following https://github.com/uken/fluent-plugin-elasticsearch#stopped-to-send-events-on-k8s-why . It works in users env with this setting. So users would like to have this configured by default. This is also the default value in upstream fluentd `kubernetes-daemonset`

https://github.com/rancher/rancher/issues/21744


aiwantaozi:
LGTM

PR #26741, opened 2020-04-21
Update kdm - single version k8s patch+nodelocal

mrajashree:


prachidamle:
No need to update rke (since it is vendored into rancher) with latest kdm?

Issue #26752, opened 2020-04-22
RKE loading wrong ingress template during cluster provisioning


ryansann:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. In Rancher UI, navigate to settings
2. Scroll down to the `rke-metadata-config` section
3. Modify the config to use the following:
```json
{
  "refresh-interval-minutes": "1",
  "url": "https://raw.githubusercontent.com/ryansann/kontainer-driver-metadata/fa58036b8a5c7a2496444d4ac93e1983d86a0d91/data/data.json"
}
```
4. Add a cluster with RKE (I am using Digital Ocean though this isn't specific to a cloud provider)
5. Choose your node template
6. In Kubernetes Options, for the Kubernetes Version, choose: `v1.16.8-rancher1-4`
7. Create Cluster

**Result:**
The ingress daemon set is provisioned using the wrong template

**Other details that may be helpful:**
I have added a new ingress template for an updated ingress container, (see [data.json](https://raw.githubusercontent.com/ryansann/kontainer-driver-metadata/fa58036b8a5c7a2496444d4ac93e1983d86a0d91/data/data.json)). The new template is expected to be used based on the configuration in the rke-metadata-config, but a different template is being used. 

The way you can distinguish that the correct template is being used is by running the following command: `kubectl get configmap rke-ingress-controller -n kube-system -o yaml | grep runAsUser` against the cluster that RKE provisioned. The output should show a value of `101` for the runAsUser field. If the value is `33`, the wrong template is being used.

Per @kinarashah "it's a regression in 2.4 when kdm logic went through a refactor"

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.2
- Installation option (single install/HA): single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Infrastructure Provider
- Machine type (cloud/VM/metal) and specifications (CPU/memory): Cloud (Digital Ocean Droplet) - 2 vCPUs / 4GB RAM
- Kubernetes version (use `kubectl version`):

```
[...]

PR #26761, opened 2020-04-22
read vendor data from local metadata

kinarashah:
Problem:
Rancher only sends non-vendored addon templates to RKE, if vendor
data is calculated from external kdm's data, all templates get
considered as vendored (sendRKELabel: "false")

https://github.com/rancher/rancher/issues/26752

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/data.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/data.go
@@ -62,6 +62,9 @@ var rancherUpdateSettingMap = map[string]settings.Setting{
 }
 
 func (md *MetadataController) loadDataFromLocal() (kdm.Data, error) {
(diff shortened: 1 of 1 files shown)


ryansann:
I validated that these changes fix #26752

StrongMonkey:
hmm @kinarashah thanks for putting the fix, I did change the code. What confuses me is that why do we need two different source of data.json(local and remote) to calculate template?

kinarashah:
@StrongMonkey 
(It could have changed) but the earlier design was rancher won't send service options and addons for every k8s version to RKE. 
In order for rancher to know which ones to send, we rely on the vendor/local/in-built data because it's guaranteed RKE already knows about them. 
So while we load the external/kdm data, we compare it with local data and update label(`sendRKELabel`/`io.cattle.rke_store`) on CRD -- this way Rancher sends the new template

Issue #26160, opened 2020-03-20
Investigate nodetemplate ID options


cloudnautique:
The data migration to support Rancher admins being able to edit them involved a data migration. The result was the IDs changed on the node templates, and clients like the Terraform provider can no longer access the node template by the old IDs. The Terraform provider believes it needs to create a new one, even though one already exists. 

To address this, we need engineering to investigate the following options:
1. Is there something that can be done to make the resource ids reachable post-upgrade?
2. If the ids are manually modified post-upgrade with the pre-upgrade values would that be supportable? (though not ideal)

rawmind0:
Submitted and tested PR https://github.com/terraform-providers/terraform-provider-rancher2/pull/288 to fix issue on terraform provider

rmweir:
Just leaving these notes here per what we have discussed.

* First, we discussed a solution where if a nodetemplate is requested from rancher that has been replaced, then the new template is returned instead if user has RBAC for it. This should not break clients consuming the API.

* The first proposed solution is abnormal and unpredictable design for an API. A more idiomatic solution would be to return a redirect response. However, this would break clients that are not prepared to handle a redirect response.

izaac:
I was able to test the workflow for Node Templates:

Reproduced the issue with the TF provider `master` branch:

`v2.3.2 -> v2.3.3`

Then validated the fix with:

`v2.3.2 -> v2.3.3 -> v2.3.6 -> v2.4.1` using the TF provider `288` branch for testing. 
Working on an scenario to include more Rancher objects at the moment.

izaac:
Tested the following E2E scenario:

1. Deploy rancher server - v2.3.2
2. Using terraform scripts:
3. Create Node Template -T1
4. Deploy a cluster using C1
5. Upgrade rancher server to 2.3.6
6. Using terraform scripts:
7. **Scale up a node pool in cluster C1 that uses T1.**

In the step 7 above `terraform apply` increasing only the node pool `quantity` and nothing else gets the node pool count increased but a new Node Template gets created and this new template is used in the new node in the nodepool.

izaac:
After further investigation and testing with @rmweir we identified I had a binary cached by terraform that was using the latest stable release code (old code without the patch).

We validated my previously [commented scenario](https://github.com/rancher/rancher/issues/26160#issuecomment-607547252) passed for both:

- `v2.3.0` to `v2.3.3`
- `v2.3.2` to `v2.3.6`

rmweir:
[...]

PR #26806, opened 2020-04-24
Return migrated node template

rmweir:
**Problem:**
Node template migration breaks API for any client referring to legacy node templates.

**Solution:**
Forward all request for legacy node templates to their corresponding migrated node template.

**Issue:**
https://github.com/rancher/rancher/issues/26160

### DIFF
--- a/pkg/api/store/nodetemplate/store.go
+++ b/pkg/api/store/nodetemplate/store.go
@@ -24,43 +24,114 @@ type nodeTemplateStore struct {
 	types.Store
 	NodePoolLister        v3.NodePoolLister
 	CloudCredentialLister corev1.SecretLister
-	NodeTemplateClient    v3.NodeTemplateInterface
+}
+
+type transformer struct {
+	NodeTemplateClient v3.NodeTemplateInterface
+}
+
+type nodeTemplateIDs struct {
+	fullOriginalID string
+	originalID     string
+	originalNs     string
+	fullMigratedID string
+	migratedID     string
+	migratedNs     string
 }
 
 func Wrap(store types.Store, npLister v3.NodePoolLister, secretLister corev1.SecretLister, ntClient v3.NodeTemplateInterface) types.Store {
-	s := &nodeTemplateStore{
-		Store:                 store,
+	t := &transformer{
+		NodeTemplateClient: ntClient,
+	}
+	s := &transform.Store{
+		Store:       store,
+		Transformer: t.filterLegacyTemplates,
+	}
+	return &nodeTemplateStore{
+		Store:                 s,
 		NodePoolLister:        npLister,
 		CloudCredentialLister: secretLister,
-		NodeTemplateClient:    ntClient,
-	}
-	return &transform.Store{
-		Store:       s,
-		Transformer: s.filterLegacyTemplates,
 	}
 }
 
-func (s *nodeTemplateStore) filterLegacyTemplates(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
+func (t *transformer) filterLegacyTemplates(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
 	ns, _ := data["namespaceId"].(string)
-	name, _ := data["name"].(string)
+	id, _ := data["id"].(string)
 	if ns != namespace.NodeTemplateGlobalNamespace {
-		s.NodeTemplateClient.Controller().Enqueue(ns, name)
+		// nodetemplate may not have been migrated yet, enqueue
+		t.NodeTemplateClient.Controller().Enqueue(ns, strings.TrimPrefix(id, ns+":"))
 		return nil, nil
 	}
 	return data, nil
 }
 
+func (s *nodeTemplateStore) ByID(apiContext *types.APIContext, schema *types.Schema, id string) (map[string]interface{}, error) {
+	ids := getAllIDs(id)
+	data, err := s.Store.ByID(apiContext, schema, ids.fullMigratedID)
+	if err != nil {
+		return nil, replaceIDInError(err, ids.migratedID, ids.originalID, ids.migratedNs, ids.originalNs)
+	}
+
+	data["namespaceId"] = ids.originalNs
+	data["id"] = ids.fullOriginalID
+	return data, nil
+}
+
+// getAllIDs returns the namespace and trimmed ID of the original ID
+// and its corresponding migrated ID
+func getAllIDs(id string) nodeTemplateIDs {
+	originalNs, originalID := ref.Parse(id)
+	fullMigratedID := getMigratedID(originalNs, originalID)
+	migratedNs, migratedID := ref.Parse(fullMigratedID)
+	return nodeTemplateIDs{
+		fullOriginalID: id,
+		originalID:     originalID,
+		originalNs:     originalNs,
+		fullMigratedID: fullMigratedID,
+		migratedID:     migratedID,
+		migratedNs:     migratedNs,
+	}
+}
+
+func getMigratedID(ns, id string) string {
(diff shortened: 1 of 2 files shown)


pkg/api/store/nodetemplate/store.go
```diff
@@ -24,43 +24,95 @@ type nodeTemplateStore struct {
...
+	return fmt.Sprintf("%s:nt-%s-%s", namespace.NodeTemplateGlobalNamespace, ns, id)
+}
+
+func getNsAndID(id string) (string, string) {
+	idParts := strings.SplitN(id, ":", 2)
+	if len(idParts) == 1 {
+		return "", idParts[0]
+	}
+	return idParts[0], idParts[1]
+}
```
prachidamle, line 87: we can use ref.Parse() utility method https://github.com/rancher/rancher/blob/master/pkg/ref/parse.go#L32

pkg/api/store/nodetemplate/store.go
```diff
@@ -24,43 +24,95 @@ type nodeTemplateStore struct {
...
 	return data, nil
 }
 
+func (s *nodeTemplateStore) ByID(apiContext *types.APIContext, schema *types.Schema, id string) (map[string]interface{}, error) {
+	originalNs, originalID := getNsAndID(id)
+	fullMigratedID := getMigratedID(originalNs, originalID)
+	migratedNs, migratedID := getNsAndID(fullMigratedID)
+	data, err := s.Store.ByID(apiContext, schema, fullMigratedID)
+	if err != nil {
+		return nil, replaceIDInError(err, migratedID, originalID, migratedNs, originalNs)
```
prachidamle, line 65: if s.Store.ByID returns say k8serror IsNotFound, their can be details about the legacy ID in the error, do we handle them too in replaceIDInError?

pkg/api/store/nodetemplate/store.go
```diff
@@ -24,43 +24,95 @@ type nodeTemplateStore struct {
...
-		Store:       s,
-		Transformer: s.filterLegacyTemplates,
 	}
 }
 
-func (s *nodeTemplateStore) filterLegacyTemplates(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
+func (t *transformer) filterLegacyTemplates(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, opt *types.QueryOptions) (map[string]interface{}, error) {
 	ns, _ := data["namespaceId"].(string)
-	name, _ := data["name"].(string)
+	id, _ := data["id"].(string)
```
prachidamle, line 50: can you pls explain why is this change needed?

pkg/api/store/nodetemplate/store.go
```diff
@@ -24,43 +24,95 @@ type nodeTemplateStore struct {
...
+		msg = strings.Replace(msg, ns, newNs, -1)
+		err = httperror.NewAPIError(apiError.Code, msg)
+	}
+	return err
+}
+
 func (s *nodeTemplateStore) Delete(apiContext *types.APIContext, schema *types.Schema, id string) (map[string]interface{}, error) {
+	originalNs, originalID := getNsAndID(id)
+	fullMigratedID := getMigratedID(originalNs, originalID)
+	migratedNs, migratedID := getNsAndID(fullMigratedID)
```
prachidamle, line 101: Can we move the three steps to a separate method and call that instead in all the Store API's?

pkg/api/store/nodetemplate/store.go
```diff
@@ -24,43 +24,95 @@ type nodeTemplateStore struct {
...
+		if pool.Spec.NodeTemplateName == fullMigratedID {
 			return nil, httperror.NewAPIError(httperror.MethodNotAllowed, "Template is in use by a node pool.")
 		}
 	}
-	return s.Store.Delete(apiContext, schema, id)
+	data, err := s.Store.Delete(apiContext, schema, fullMigratedID)
+	if err != nil {
+		return nil, replaceIDInError(err, migratedID, originalID, migratedNs, originalNs)
+	}
+	return data, nil
```
prachidamle, line 115: replace back the originalId and NS in the response data?

pkg/api/store/nodetemplate/store.go
rmweir: name was the display name, so nothing which is not valid for enqueueing

pkg/api/store/nodetemplate/store.go
rmweir: yes, id is being replaced to not break anything that may rely or reuse response id- as request by @ibuildthecloud and @cloudnautique

prachidamle: right understood, but I meant why are we not replacing it back here for Delete() like its done in Update()

rmweir: oh- because data return by deletes is nil anyway. I can just return a nil to be safe

pkg/api/store/nodetemplate/store.go
rmweir: I believe in the proxy store k8s errors (which fit the interface Status) are translated to norman api errors: https://github.com/rancher/norman/blob/07a7713a06d92f6a2aee25cb2b86c6670577febc/store/proxy/error_wrapper.go#L46

pkg/api/store/nodetemplate/store.go
```diff
@@ -24,43 +24,98 @@ type nodeTemplateStore struct {
...
+	}
+
+	data["namespaceId"] = originalNs
+	data["id"] = id
+	return data, nil
+}
+
+// getAllIDs returns the namespace and trimmed ID of the original ID
+// and its corresponding migrated ID
+func getAllIDs(id string) (string, string, string, string, string) {
```
dramich, line 73: I think this could just return a struct with the fields, returning 5 strings from one func is pretty rough to reason about.

tests/integration/suite/test_node_templates.py
```diff
@@ -47,16 +52,57 @@ def migrated_template_exists(id):
...
+    def legacy_template_deleted():
+        try:
+            k8s_dynamic_client.get_namespaced_custom_object(
+                "management.cattle.io", "v3", ns.name, 'nodetemplates',
+                legacy_id)
+            return False
+        except ApiException as e:
+            assert e.status == 404
+            print(e)
+            return True
```
dramich, line 76: In place of the True, do a check on the error code, if it's 404 return true, if not let it keep going and time out. Also drop the print

tests/integration/suite/test_node_templates.py
```diff
@@ -47,16 +52,57 @@ def migrated_template_exists(id):
...
+
+    # wait for legacy node template to be deleted
+    wait_for(lambda: legacy_template_deleted(), fail_handler=lambda:
              "failed waiting for old node template to delete")
 
+    # retrieve node template via legacy id
+    nt = admin_client.by_id_node_template(id=full_legacy_id)
```
dramich, line 83: Validate using both IDs get the same data back before attempting to change it.

tests/integration/suite/test_node_templates.py
```diff
@@ -47,16 +52,57 @@ def migrated_template_exists(id):
...
+    nt2 = admin_client.by_id_node_template(id=id)
+    # assert node template being updated is migrated node template
+    assert nt2.azureConfig.customData == new_config.customData
+
+    # delete node template via legacy id
+    admin_client.delete(nt)
+    wait_for(lambda: admin_client.by_id_node_template(id) is None,
+             fail_handler=lambda:
+             "failed waiting for migrate node template to delete")
+    admin_client.by_id_node_template(id="something:soment")
```
dramich, line 104: This looks unfinished, is there going to be another check?

rmweir:
Tested using 1.8.1 of rancher2 provider. When upgrading from 2.3.2 to master and using `terraform apply`, the node template is recreated because it thinks it does not exist. When upgrading from 2.3.2 to this PR and using `terraform apply`, nothing changes

PR #26814, opened 2020-04-25
stabilize test_auth when used in Jenkins

izaac:
Make sure to only remove the delete container if Rancher was deployed and `DELETE_RANCHER_SERVER` is `true`

Fix the `cluster` and `aws_nodes` references when cleaning up the test_auth resources.

Use the admin context when `DELETE_RANCHER_SERVER` is true to delete the clusters before deleting the Rancher server.

Added `validate` parameter for `create_and_validate_custom_host` for making the cluster validation optional. This will save resources as we won't need to validate workloads/ingresses availability for tests that don't need them for validation, in this case `test_auth` and thus no need to deploy VM instances larger than `t2.medium` in AWS.

### DIFF
--- a/tests/validation/tests/v3_api/test_auth.py
+++ b/tests/validation/tests/v3_api/test_auth.py
@@ -17,7 +17,7 @@ from .test_rke_cluster_provisioning import create_and_validate_custom_host
 Prerequisite:
 1. testautoadmin as your admin user, if the fixture detects the auth
    is disabled it will be enabled automatically.
-2. Two clusters in your setup, if none or one are detected by the fixture 
+2. Two clusters in your setup, if none or one are detected by the fixture
    will create clusters to match two
 '''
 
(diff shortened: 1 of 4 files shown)


tests/validation/tests/v3_api/test_auth.py
```diff
@@ -741,21 +741,24 @@ def create_project_client(request):
...
     aws_nodes1 = cluster1 = None
     aws_nodes2 = cluster2 = None
     if cluster_total == 0:
-        cluster1, aws_nodes1 = create_and_validate_custom_host(node_roles, True)
-        cluster2, aws_nodes2 = create_and_validate_custom_host(node_roles, True)
+        cluster1, aws_nodes1 = create_and_validate_custom_host(**kargs)
+        cluster2, aws_nodes2 = create_and_validate_custom_host(**kargs)
     if cluster_total == 1:
-        cluster1, aws_nodes1 = create_and_validate_custom_host(node_roles, True)
+        cluster2, aws_nodes2 = create_and_validate_custom_host(**kargs)
```
jiaqiluo, line 753: Is it possible that cluster_total > 1 ?

izaac: Yes, if you prepare the requirements yourself and execute the `needs_cluster` jenkins job then there wouldn't be need for creating clusters. There's a check for this further down in the code where the test asserts `clusters >= 2`

Issue #26807, opened 2020-04-24
Worker node cordoned after adding new worker in custom rke cluster


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Deploy a custom cluster in Rancher `v2.4.3-rc2` k8s version `1.17.5-rancher1-1` (1 node, all roles)
Wait for cluster to become available.
Add a worker node.
Continue adding worker nodes until previous nodes become cordoned.

**Result:**
![Screen Shot 2020-04-24 at 12 05 15 PM](https://user-images.githubusercontent.com/57409052/80249860-700d7c80-8627-11ea-8f3e-c263d6197072.png)

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `v2.4.3-rc2`
- Installation option (single install/HA): Single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom

Related issue:
https://github.com/rancher/rancher/issues/26790

StrongMonkey:
@aaronRancher I assume this is the same issue that was happening on https://github.com/rancher/rancher/issues/26791?

aaronRancher:
@StrongMonkey This is part of the same issue, but created to specifically note the "cordoned" behavior. We've seen nodes become temporarily "cordoned" when a worker node is successfully provisioned. Can close issue if deemed too redundant.

StrongMonkey:
I am not aware of node being cordoned automatically though. @kinarashah ^^ maybe you have some idea?

sowmyav27:
Verified on a custom image from 2.4-release and master branch 

**Steps:**

- Create a custom cluster - 1 node all roles
- Add a worker node - The worker node will be stuck in registering state because of this issue - https://github.com/rancher/rancher/issues/26790
- Add more worker nodes 
- No panic is seen
- The cluster goes into `Updating` state for about 3-5 seconds and comes back to `Active`.
- Existing "Active" worker nodes do NOT go into "Cordoned" state.

Issue #26790, opened 2020-04-24
Worker node fails to register in a custom rke cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a custom cluster in 2.3.6 k8s - 1.17.4-rancher1-2
- Upgrade rancher server to 2.4.3-rc2
- Upgrade k8s version to 1.17.5
- Add a worker node to the cluster
- the node is stuck in registering state
- Logs of `share-mnt` container on the node

```
Error: failed to start containers: kubelet
+ sleep 2
+ docker start kubelet
Error response from daemon: {"message":"No such container: kubelet"}
Error: failed to start containers: kubelet
+ sleep 2
```

**Expected Result:**
Kubelet should be deployed on the node
And the worker node should get registered.

**Other details that may be helpful:**

- This was not reproducible on an addition of a worker node in a node driver AWS cluster.
- Reproduced this issue on an addition of the Second worker node in an upgraded cluster in a single node install. This was right after rancher server upgrade. No k8s upgrade was done on the cluster. K8s on this cluster - v1.17.4

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.3.6 to 2.4.3-rc2
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom-rke
- Kubernetes version (use `kubectl version`):

```
1.17.5
```

- Docker version (use `docker version`):

```
(paste the output here)
```

mcmcghee:
I'm seeing this same issue on 2.4.2 and 1.17.4. It happens completely randomly on the provision of a new cluster and only on 2-3 nodes.

izaac:
I experienced this after upgrading from `v2.4.2` to `v2.4.3-rc2` in a Custom cluster.

zbialik:
I experienced this after upgrading from `v2.3.2` to `v2.4.2` in a Custom cluster. 

I wrote [this issue post](https://github.com/rancher/rancher/issues/26776) outlining some more specific information.

aaronRancher:
Reproducible in a single node fresh install of v2.4.3-rc2.
Steps:
Provision 1 node custom cluster (all roles)
Add 1 worker node (succeeded)
Add 1 worker node (failed, stuck provisioning)
Error from `share-mnt`:
```
[...]

Issue #24994, opened 2020-01-17
Add TRACE loglevel


superseb:
After adding a significant amount of debug logging, we could move our very verbose debug lines to a TRACE level to make sure the debug logging is still readable. Debug would then include printing single line variables/singe line diffs between larger objects/certain behavior being executed, and TRACE would include dumping the structs/objects. This would also help in leaving debug on for a longer time for a certain issue (else it would either fill the disk or app could become sluggish)

superseb:
To test Rancher:

- Executing `loglevel --set trace` in `rancher/rancher` container (PASS)
- Executing `loglevel --set invalid` in `rancher/rancher` container (FAIL)
- Executing `loglevel --set debug` in `rancher/rancher` container (PASS)
- Executing `loglevel --set info` in `rancher/rancher` container (PASS)

To test RKE:

- Executing `rke help` should show `--trace` as flag
- Executing `rke --trace up` should show `TRAC[0000] Loglevel set to [trace]`
- Executing `rke --debug up` should still work and show `DEBU[0000] Loglevel set to [debug]`

aaronRancher:
This enhancement is verified with Rancher server `rancher\rancher:master-head` `version beba5247e`

Executing `loglevel --set trace` in `rancher/rancher` container (PASS)
```
root@cf0e798bffa9:/var/lib/rancher# loglevel --set trace
OK
```

Executing `loglevel --set invalid` in `rancher/rancher` container (FAIL)
```
root@cf0e798bffa9:/var/lib/rancher# loglevel --set invalid
Error when setting loglevel: Failed to parse loglevel: not a valid logrus Level: "invalid"
```

Executing `loglevel --set debug` in `rancher/rancher` container (PASS)
```
root@cf0e798bffa9:/var/lib/rancher# loglevel --set debug
OK
```

Executing `loglevel --set info` in `rancher/rancher` container (PASS)
```
root@cf0e798bffa9:/var/lib/rancher# loglevel --set info
OK
```

This enhancement is verified with RKE version `v1.1.0-rc11`
`rke help` shows `--trace` as a flag
```
GLOBAL OPTIONS:
   --debug, -d    Debug logging
   --quiet, -q    Quiet mode, disables logging and only critical output will be printed
   --trace        Trace logging
   --help, -h     show help
   --version, -v  print the version
```
Executing `rke --trace` up shows:
```
Aarons-MBP:documents aaronyee$ ./rke --trace up
[...]

PR #26815, opened 2020-04-26
[2.4] wait for dockerInfo to be set before creating node plan

kinarashah:
Problem:
creating node plan with empty dockerRootDir results into wrong kubelet's process

https://github.com/rancher/rancher/issues/26790

Problem:
change in node plan of registering nodes triggers cluster upgrade for other worker nodes

https://github.com/rancher/rancher/issues/26807

### DIFF
--- a/pkg/controllers/management/rkeworkerupgrader/plan.go
+++ b/pkg/controllers/management/rkeworkerupgrader/plan.go
@@ -30,6 +30,8 @@ func (uh *upgradeHandler) nonWorkerPlan(node *v3.Node, cluster *v3.Cluster) (*v3
 	hostAddress := node.Status.NodeConfig.Address
 	hostDockerInfo := infos[hostAddress]
 
+	logrus.Debugf("getDockerInfo for node [%s] dockerInfo [%s]", node.Name, hostDockerInfo.DockerRootDir)
+
 	svcOptions, err := uh.getServiceOptions(rkeConfig.Version, hostDockerInfo.OSType)
 	if err != nil {
 		return nil, err
@@ -85,6 +87,8 @@ func (uh *upgradeHandler) workerPlan(node *v3.Node, cluster *v3.Cluster) (*v3.RK
 	if err != nil {
 		return nil, err
 	}
+
+	logrus.Debugf("getDockerInfo for node [%s] dockerInfo [%s]", node.Name, hostDockerInfo.DockerRootDir)
 
 	np := &v3.RKEConfigNodePlan{}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/rkeworkerupgrader/plan.go
```diff
@@ -30,6 +30,8 @@ func (uh *upgradeHandler) nonWorkerPlan(node *v3.Node, cluster *v3.Cluster) (*v3
 	hostAddress := node.Status.NodeConfig.Address
 	hostDockerInfo := infos[hostAddress]
 
+	logrus.Tracef("getDockerInfo for node [%s] dockerInfo %#v", node.Name, hostDockerInfo.DockerRootDir)
```
StrongMonkey, line 33: Good to know this. What does tracef do? :smiley:

pkg/controllers/management/rkeworkerupgrader/plan.go
```diff
@@ -72,6 +74,11 @@ func (uh *upgradeHandler) workerPlan(node *v3.Node, cluster *v3.Cluster) (*v3.RK
 	hostAddress := node.Status.NodeConfig.Address
 	hostDockerInfo := infos[hostAddress]
 
+	if hostDockerInfo.Driver == "" || hostDockerInfo.DockerRootDir == "" {
```
StrongMonkey, line 77: What's the workflow in here if error is returned?

pkg/controllers/management/rkeworkerupgrader/plan.go
kinarashah: we've recently started printing out whole objects in trace than debug so debug becomes more readable https://github.com/rancher/rancher/issues/24994

pkg/controllers/management/rkeworkerupgrader/plan.go
kinarashah: it would return back to the sync, and keep retrying till the docker info gets assigned

kinarashah:
@superseb @StrongMonkey 

Changing the approach a little. 
DockerInfo gets synced on the node each time agent requests for a node plan. There's no point in returning error in controllers or in `GetDockerInfo` because retry there isn't going to fix anything + spam logs. I've changed logic to wait for `Node.Status.DockerInfo` to be set on before getting and setting the node plan. 

The second commit is for issue that's happening because of this wrong node plan. First node is stuck in `registering` and on addition of other nodes, rkeConfig changes and `toUpgradeCluster` becomes true because registering node now has the right node plan (dockerInfo got synced). 

This issue shouldn't happen now, but let's say node is stuck in registering state for some other reason, any subsequent `rke up` shouldn't result into workers upgrade because of change in this node plan -- so I'm ignoring registering nodes.

superseb:
Makes sense to me, should we add a logline for each found condition to trace down the cause of not getting a node plan?

StrongMonkey:
Also build failed

Issue #26792, opened 2020-04-24
Cluster goes into updating state when adding a worker node and a panic is seen on the cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

1. This is on an upgraded rancher from 2.3.6 to 2.4.3-rc2
2. Deploy a custom cluster - 3 etcd, 2 control, 3 worker nodes in 2.3.6
3. upgrade rancher to 2.4.3-rc2
4. Add a worker node. 
5. Worker node is stuck in "Registering" state
6. Add another worker node.
7. This is also stuck in Registering state, 
8. But an existing worker node in cluster is in "Cordoned" state
9. Cluster goes into "Updating" state.
10. An rke up is triggered.
11. Now the cluster is stuck in "Updating" state
12. There is a panic in the cluster

```
Updated cluster [c-lr9nb] with node version [2]
2020/04/24 01:15:52 [INFO] Provisioned cluster [c-lr9nb]
2020/04/24 01:15:52 [INFO] checking cluster [c-lr9nb] for worker nodes upgrade
2020/04/24 01:15:52 [ERROR] NodeController c-lr9nb/upgrade_ recovered from panic "runtime error: invalid memory address or nil pointer dereference". (err=<nil>) Call stack:
goroutine 18685 [running]:
github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime.RecoverFromPanic(0xc008911d30)
	/go/src/github.com/rancher/rancher/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:158 +0xb5
panic(0x3686aa0, 0x6c38100)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/rancher/rancher/pkg/controllers/management/rkeworkerupgrader.(*upgradeHandler).upgradeCluster(0xc00b5af340, 0xc0052e9700, 0xc0064a26a0, 0x10, 0x0, 0x0, 0x0)
	/go/src/github.com/rancher/rancher/pkg/controllers/management/rkeworkerupgrader/upgrade.go:262 +0x630
github.com/rancher/rancher/pkg/controllers/management/rkeworkerupgrader.(*upgradeHandler).Sync(0xc00b5af340, 0xc0064a26a0, 0x10, 0x0, 0xc000abc380, 0x7f63c16cc6d0, 0x0, 0x0)
[...]

PR #26834, opened 2020-04-27
Fix panic on adding worker node

mrajashree:
https://github.com/rancher/rancher/issues/26792


kinarashah:
We'd also need to update the cluster object, since this is only initializing the upgradeStrategy. Can close this in lieu of https://github.com/rancher/rancher/pull/26847

mrajashree:
cool, thanks @kinarashah

Issue #26791, opened 2020-04-24
Upgrade from v2.4.2 to v2.4.3-rc2 the K8s version dropdown in edit cluster is empty sometimes


izaac:
**What kind of request is this:**
Bug

**Steps to reproduce:**

- Upgrade from Rancher v2.4.2 to v2.4.3-rc2 HA
  - Local imported cluster k8s: `v1.17.4`
  - Build your own certificate with Recognized CA
  - RKE clusters used 
    - Azure `v1.15.11-rancher1-3`
    - EC2 `1.16.8-rancher1-3`
    - Custom `Windows 1.16.8-rancher1-3`
  - EKS `v1.14.9-eks-f459c0`

- Install command:
   - `helm_v3 install rancher rancher-latest/rancher --version v2.4.2 --namespace cattle-system --set hostname=<hostname> --set ingress.tls.source=secret --set rancherImageTag=v2.4.2`

- Upgrade command:
  - `helm_v3 upgrade rancher rancher-latest/rancher --version v2.4.3-rc2 --namespace cattle-system --set hostname=<hostname> --set ingress.tls.source=secret --set rancherImageTag=v2.4.3-rc2`

**Result:**

Going to Edit cluster in each of the above clusters shown the `Kubernetes version` dropdown empty.

**Other details that may be helpful:**

I had to navigate and click: `Global -> Tools -> Drivers -> Refresh Kubernetes Metadata` and after that the Dropdown was populated again with k8s versions.

**Environment information**
- Rancher version: `v2.4.3-rc2`
- Installation option: HA with local k8s `v1.17.4` cluster

StrongMonkey:
Related https://github.com/rancher/rancher/issues/24091

StrongMonkey:
#24091 is not addressed, This issue is not easy to reproduce during the normal setup. 
@izaac You should try to following setup and should be able to reproduce this bug:
1. Create a HA cluster with 3 nodes, wait for kubernetes dropdown to be popuated.
2. Edit rancher deployment to increase one replica, and then check kubernetes dropdown. You should see empty value.

I put an fix and will need to verify with @kinarashah about the potential regression. Once fix is put in, try the steps above and verify the bug.

Also I would like verify that no regression is introduced. You might also need to verify the following setting is populated correctly. You can go to https://server-url/v3/settings and check each value.
1. k8s-version
2. k8s-version-to-service-options
3. k8s-version-to-images
4. k8s-versions-current
5. k8s-versions-deprecated
[...]

PR #26838, opened 2020-04-28
Fix kubernete dropdown empty issue

StrongMonkey:


StrongMonkey:
@kinarashah I think I found the bug where setting is wiped out to empty occasionally during upgrade and bootstrap. In here we are modifying the default setting, which is conflicted with the code path here https://github.com/rancher/rancher/blob/be429b344df99c082bbc03fb8db0cd67e3cccb5a/pkg/api/controllers/settings/settings.go#L113. SetAll() is called from SetProvider(), which is called during the bootstrap on every node no matter it is leader or not. So it raises a race condition so that the default value can be reverted to empty if SetAll() is called after driver metadata is refreshed. This can be easily reproduced in HA cluster by just adding one replica. I looked at the code and think there is no need to update default value of the setting, but want to sync with you to check if default value is being used somewhere.

StrongMonkey:
Fix https://github.com/rancher/rancher/issues/26791.

PR #26847, opened 2020-04-29
[2.4] fix panic on empty worker upgrade strategy

kinarashah:
I don't really know why this panic is happening, all scenarios should have been covered by L239 block. Issue was observed during an incorrect cluster upgrade on adding a worker node, but since this is a panic, arranging logic a little to always check for empty upgradeStrategy. 

#26792


pkg/controllers/management/rkeworkerupgrader/upgrade.go
```diff
@@ -230,26 +230,34 @@ func (uh *upgradeHandler) upgradeCluster(cluster *v3.Cluster, nodeName string, p
 	defer uh.clusterLock.Unlock(clusterName)
 
 	logrus.Debugf("cluster [%s] upgrading condition: [%v] plan changed: [%v]", clusterName, v3.ClusterConditionUpgraded.IsUnknown(cluster), planChanged)
+
+	var (
+		clusterCopy *v3.Cluster
+		err         error
+	)
 	if !v3.ClusterConditionUpgraded.IsUnknown(cluster) || planChanged {
 		clusterCopy := cluster.DeepCopy()
```
mrajashree, line 239: since we continue to use same clusterCopy object line 244 onwards, `clusterCopy := cluster.DeepCopy()` should be changed to `clusterCopy = cluster.DeepCopy()`?

PR #26853, opened 2020-04-29
Preliminary support for LE DNS-01 challenge

### LINKED ISSUE #26850: [Feature Request] LetsEncrypt DNS-01 challenge type
Feature Request

It would be nice to have support for DNS-01 challenge requests for LetsEncrypt during Rancher install.

Use Case:

1) Security - Would not have to expose Rancher to the internet directly in order to get a valid certificate, this increases overall security and limits the attack surface.

2) Security - Using wildcard certificates obscures the names of (potentially) internal services from showing up in certificate transparency logs, which is mostly security through obscurity, but could be useful in some cases.

3) Convenience - DNS-01 is more convenient to use since it doesn't require changing firewall rules, something which can be an issue for businesses.

Just-Insane:
This PR adds preliminary support for Let's Encrypt DNS-01 challenge type.

Closes #26850

### DIFF
--- a/chart/templates/issuer-letsEncrypt.yaml
+++ b/chart/templates/issuer-letsEncrypt.yaml
@@ -26,9 +26,16 @@ spec:
       name: letsencrypt-{{ .Values.letsEncrypt.environment }}
     {{- if or (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") (and (eq (int $certmanagerVer._0) 0) (ge (int $certmanagerVer._1) 11)) }}
     solvers:
+    {{- if eq .Values.letsEncrypt.method "DNS-01" }}
+    - dns01:
+{{- if .Values.extraletsEncryptArgs }}
+{{ toYaml .Values.extraletsEncryptArgs | indent 6 }}
+{{- end }}
+    {{- else }}
     - http01:
         ingress:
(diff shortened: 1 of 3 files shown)


whywaita:
@Just-Insane Hi! Thank you for your PR. :)
I want DNS-01 mode also.

why don't set reviewers?
(and maybe, you need to fix conflict!)

erSitzt:
Is this PR dead ?
Would love to see DNS-01 support without manual steps as this is my only option to use letsencrypt

Just-Insane:
For all intents and purposes this PR is dead. There is a solution in the related issue about setting up DNS01 validation with LE and replacing the certificate after cluster install.

Technically you could setup cert-manager before installing rancher and just select that certificate by default.

erSitzt:
Ok, so just set up an ClusterIssuer and populate the secret used by rancher by default like described in #26850
Still not sure which way feels cleaner though...

Just-Insane:
Yea, that’s what I’ve been doing.

It is probably cleaner from a cluster standpoint, one less certificate to have to store and renew, but it feels sloppy from a rancher standpoint.

I want to be able to `rke up` and then follow the rancher install docs without any other intervention on the cluster, but that is not currently possible.

PR #26859, opened 2020-04-29
add support for installing Rancher in k3s cluster

jiaqiluo:
Now we can provide the existing hostname and k3s cluster's kubeconfig file for installing Rancher and provisioning a custom cluster on it. 

all resources (instances, target group, load balancer) will be named with a unified prefix

New parameters in Jenkins:
- RANCHER_HA_HOSTNAME
- RANCHER_HA_KUBECONFIG

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -4,14 +4,18 @@ import pytest
 import rancher
 import time
 
-from .common import create_config_file
 from .common import create_user
+from .common import create_config_file
+from .common import get_custom_host_registration_cmd
 from .common import random_test_name
+from .common import random_name
 from .common import readDataFile
 from .common import run_command_with_stderr
 from .common import set_url_password_token
-
+from .common import validate_cluster
+from .common import wait_until_active
 from lib.aws import AmazonWebServices
+from .test_rke_cluster_provisioning import rke_config
 
 DATA_SUBDIR = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                            'resource')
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -42,42 +47,75 @@
 kubeconfig_path = DATA_SUBDIR + "/kube_config_cluster-ha-filled.yml"
 export_cmd = "export KUBECONFIG=" + kubeconfig_path
 
+rke_config = {"authentication": {"type": "authnConfig", "strategy": "x509"},
```
sowmyav27, line 50: We have an rke_config defined in common.py. Can we not use that here instead?

jiaqiluo: I import the `rek_config` from the `test_rke_cluster_provisioning.py` instead of moving it to `common.py` because I see the `rke_config` is imported into multiple files and I don't want to change all of them in this PR. 
I can have another PR for moving the `rke_config` and need to validate it is not breaking anything.

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -43,41 +48,67 @@
...
-
     if "byo-" in RANCHER_HA_CERT_OPTION:
         cm_install = False
+    print("The hostname is: {}".format(RANCHER_HA_HOSTNAME))
+
+    # prepare an RKE cluster and other resources
+    # if no kubeconfig file is provided
+    if RANCHER_HA_KUBECONFIG is None:
+        nodes = create_resources()
+        rke_config = create_rke_cluster_config(nodes)
```
sowmyav27, line 61: This is same as the global variable `rke_config` --> `from .test_rke_cluster_provisioning import rke_config` . Do we have any chances that it will be overridden here?

jiaqiluo: the variable is renamed to `config_path` which makes more sense here (`create_rke_cluster_config(nodes)` returns the path to the filled config)

PR #26860, opened 2020-04-29
Scripts to create rancher, clusters, and run validation tests against them

rancher-max:
This initial PR introduces a new Jenkinsfile that has three functional stages:
1. Deploy Rancher server
2. Deploy clusters into that rancher server
3. Run validation tests in parallel on those clusters

This is initially working with RKE clusters and will be expanded for EKS, GKE, and AKS very soon. Custom Clusters can also be added eventually.

### DIFF
--- a/tests/validation/tests/v3_api/test_deploy_clusters.py
+++ b/tests/validation/tests/v3_api/test_deploy_clusters.py
@@ -1,0 +1,70 @@
+from .common import *  # NOQA
+from .test_rke_cluster_provisioning import create_and_validate_custom_host
+
+if_not_auto_deploy_rke = pytest.mark.skipif(
+    ast.literal_eval(
+        os.environ.get(
+            'RANCHER_TEST_DEPLOY_RKE', "False")) is False,
+    reason='auto deploy RKE tests are skipped')
+if_not_auto_deploy_eks = pytest.mark.skipif(
+    ast.literal_eval(
+        os.environ.get(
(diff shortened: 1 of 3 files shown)


tests/validation/tests/v3_api/test_deploy_clusters.py
```diff
@@ -0,0 +1,53 @@
...
+        node_roles = [["controlplane"], ["etcd"],
+                      ["worker"], ["worker"], ["worker"]]
+        cluster, aws_nodes = create_and_validate_custom_host(
+            node_roles, random_cluster_name=True, version=k8s_version)
+        env_details += cluster.name
+        print("Successfully deployed {} with kubernetes version {}".format(
+            cluster.name, k8s_version))
+    env_details += "'"
+    print("Env details: {}".format(env_details))
+    create_config_file(env_details)
```
jiaqiluo, line 41: Why do we need to create the config file here?

rancher-max: So that the new environment variable is available (`env.CLUSTER_NAMES`) and the Jenkinsfile can load it in: 
```
// copy file containing CLUSTER_NAMES and load into environment variables
sh "docker cp ${clusterSetupContainer}:${rootPath}${testsDir}${rancherConfig} ."
load rancherConfig
```

sangeethah:
LGTM

Issue #26323, opened 2020-03-28
Rancher UI gives 404 backend when using helm chart 


MrAmbiG:
**What kind of request is this (question/bug/enhancement/feature request):**
BUG

**Steps to reproduce (least amount of steps as possible):**
any one of 3
`helm install rancher rancher-stable/rancher \
  --namespace cattle-system \
  --set hostname=rancher.amd.com \
  --set ingress.tls.source=secret

helm install rancher rancher-stable/rancher \
  --namespace cattle-system \
  --set hostname=rancher.amd.com
  
  
helm install rancher rancher-stable/rancher \
  --namespace cattle-system \
  --set hostname=rancher.amd.com \
  --set ingress.tls.source=letsEncrypt \
  --set letsEncrypt.email=myemail@amd.com `

**Result:**
default backend - 404

**Other details that may be helpful:**
`kubectl get ingress -n cattle-system`    
NAME      HOSTS                  ADDRESS                       PORTS     AGE    
rancher   rancher.amd.com   ip_addr1,ip_addr2                  80, 443   48m    

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher server with helm/stable
- Installation option (single install/HA): 2 node

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):Hosted
- Machine type (cloud/VM/metal) and specifications (CPU/memory):metal
- Kubernetes version (use `kubectl version`):

```
1.18
```

- Docker version (use `docker version`): 

```
19.03.5
```

superseb:
1.18 is not yet validated in either Rancher or RKE, on what kind of cluster are you installing Rancher on? Also, please include output from the troubleshooting docs here: https://rancher.com/docs/rancher/v2.x/en/troubleshooting/

MrAmbiG:
@superseb 
k8s cluster --> RKE cluster. 
- I followed rancher's tutorial on how to install k8s using rke. The version of k8s was an automatic decision from rke. I did not specify or have not rigid requirement of any version.
- I wanted rancher server on rke cluster. So i followed rancher website's installation guide. `helm install rancher rancher-stable/rancher` as per their site. Here too the choice of version was left to helm ( latest).
Goal = Deploy 2 node rke cluster + deploy rancher server on rke cluster.

superseb:
[...]

PR #26870, opened 2020-04-30
Add ingress class option to helm chart

galal-hussein:
Removing ingress Class nginx so that any ingress controller can use the ingress defined for lets encrypt

https://github.com/rancher/rancher/issues/26323

### DIFF
--- a/chart/templates/issuer-letsEncrypt.yaml
+++ b/chart/templates/issuer-letsEncrypt.yaml
@@ -28,7 +28,7 @@ spec:
     solvers:
     - http01:
         ingress:
-          class: nginx
+          class: {{ .Values.letsEncrypt.ingress.class }}
     {{- else if or (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") (and (eq (int $certmanagerVer._0) 0) (lt (int $certmanagerVer._1) 11)) }}
     http01: {}
     {{- end }}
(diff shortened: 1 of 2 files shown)


chart/values.yaml
```diff
@@ -65,6 +65,8 @@ ingress:
 letsEncrypt:
   # email: none@example.com
   environment: production
+  # options: nginx, traefik
+  ingressClass: nginx
```
rawmind0, line 69: May be it's better add class variable nested to ingress , `ingress.class` instead of `letsEncrypt.ingressClass` ??

chart/templates/issuer-letsEncrypt.yaml
```diff
@@ -28,7 +28,7 @@ spec:
     solvers:
     - http01:
         ingress:
-          class: nginx
+          class: {{ .Values.letsEncrypt.ingressClass }}
```
rawmind0, line 31: May be better `ingress.class`??

```
-          class: {{ .Values.letsEncrypt.ingressClass }}
+          class: {{ .Values.ingress.class }}
```

catherineluse:
This has been documented here https://github.com/rancher/docs/pull/2484

catherineluse:
@cjellick so if it will work out of the box with no changes, does that mean these changes don't need to be made? https://github.com/rancher/docs/pull/2484

galal-hussein:
@catherineluse there is still changes need to be made to the docs to clarify ifbthe the user need to restrict the resolver ingress to certain provider

PR #26963, opened 2020-05-06
Add OCI meta handler API for OKE and OCI drivers, bump OKE driver version

jlamillan:
Co-authored-by: Jesse Millan <jesse.millan@oracle.com>
Co-authored-by: Joe Rosinski <joe.rosinski@oracle.com>

This PR:
* adds `/meta/oci` endpoint to Rancher for the OKE cluster driver and OCI node template/cloud-credential UI's to validate credentials and retrieve values such as shapes and versions dynamically (requires [OCI Golang SDK](https://github.com/oracle/oci-go-sdk)).
* updates the OKE driver to the latest version, which is backwards compatible
* adds integration tests for the OKE cluster driver and additionally exercises the meta endpoint for OCI.

This is followup work from the previous PRs [1](https://github.com/rancher/ui/pull/3931), [2](https://github.com/rancher/ui/pull/3905), which did not validate credentials and used static maps for the content. 

Corresponding PRs for the driver UI's to use the `/meta/oci` endpoint:
- OCI node driver: https://github.com/rancher/ui/pull/3952
- OKE cluster driver: https://github.com/rancher/ui/pull/3953


deniseschannon:
https://github.com/rancher/rancher/issues/27051

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,217 @@
...
+	credID := req.URL.Query().Get("cloudCredentialId")
+	if credID != "" {
+		ns, name := ref.Parse(credID)
+		if ns == "" || name == "" {
+			writer.WriteHeader(http.StatusUnprocessableEntity)
+			logrus.Debugf("invalid cloud credential ID %s", credID)
+			return fmt.Errorf("invalid cloud credential ID %s", credID)
+		}
+
+		cc, err := handler.secretsLister.Get(namespace.GlobalNamespace, name)
```
dramich, line 145: There needs to be a check here that validates the user sending the request has access to the cloud credential being referenced.

dramich: This should follow the same pattern the vspehere handler is doing

jlamillan:
@dramich - thank you for the detailed feedback.

> For the debug messages, add a context to where they are coming from i.e. [oci-handler] then the message

We'll prefix all debug messages with `[oci-handler]`


> Debugs are inconsistent in the case statement.

We'll add the missing debug statements to those last few cases.


> The error codes being returned need to be based off what actually happened in the request. Just defaulting all errors to a 500 is not how an API works and is not accurate to what is actually happening. If the user provided bad info the error code and message need to represent that.

It's not necessary to be condescending to get your point across here is it?  We figured a pass/fail approach would suffice considering the OCI driver UI code is the _only_ intended client of the API and the error message itself was preserved. But we can look into passing the response from the OCI API downstream for posterity. 

> Imports should be grouped into two groups, go standard lib and everything else. Use goimports to help with this.

We'll run `goimports` to reorganize the imports for you

> Break into two commits, one for rancher changes and a 2nd one for the go.mod changes, this makes it easier to review.

We'll move the go.mod / vendor/ changes to a second commit.

> Tests need to be added to validate the new endpoint. This should at minimum be added into the integration tests to ensure some basic functionality, like rejecting for bad creds or missing, not providing a valid resourceType

Do you mean tests that do validation _locally_, or are you talking about tests that make outbound calls to the OCI API?

> Go tests can also be included to test helper functions that are not attempting to call out to an API.

Noted.

pkg/api/customization/oci/handler.go
jlamillan: This is complete.

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,217 @@
...
+
+func NewOCIHandler(scaledContext *config.ScaledContext) http.Handler {
+	return &Handler{
+		secretsLister: scaledContext.Core.Secrets(namespace.GlobalNamespace).Controller().Lister(),
+	}
+}
+
+func (handler *Handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
+
+	if req.Method != http.MethodPost {
```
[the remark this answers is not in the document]
jlamillan: It is ultimately a `GET` on the Oracle API side. This was a holdover from developing it against the cluster driver, which does not support cloud-credentials and therefore must _always_ pass credentials in the body, which people don't expect on a `GET`.

In any event, I've updated to allow `GET` in the case where cloud-creds are passed via query param and POST for cluster driver scenario.

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,217 @@
...
+		writer.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+	writer.Header().Set("Content-Type", "application/json")
+
+	// New credential every call
+	creds := Credentials{}
+	err := handler.extractCreds(writer, req, &creds)
+	if err != nil {
+		util.ReturnHTTPError(writer, req, httperror.ServerError.Status, err.Error())
```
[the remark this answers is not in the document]
jlamillan: I believe the first part of this is fixed. The co-author will push an update soon to pass the response code (on error) from `oci/listers.go`

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,217 @@
...
+		return
+	}
+
+	err = handler.validateCreds(&creds)
+	if err != nil {
+		util.ReturnHTTPError(writer, req, httperror.ServerError.Status, err.Error())
```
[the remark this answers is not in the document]
jlamillan: fixed

jlamillan:
@dramich validation tests have been added as well, which I believe covers the surface area you recommended (error scenarios) of the meta endpoint for OCI in addition to exercising the driver to provision a cluster if credentials are supplied.

tests/validation/tests/v3_api/test_oke_cluster.py
```diff
@@ -0,0 +1,203 @@
...
+                                          'cannot create cluster')
+
+
+def test_error_get_oke_latest_versions_missing_key():
+    oci_cred_body = {
+        "tenancyOCID": "ocid1.tenancy.oc1..aaaaaa",
+        "userOCID": "ocid1.user.oc1..aaaaaaaa",
+        "region": "us-ashburn-1",
+        "fingerprint": "xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx",
+    }
```
jiaqiluo, line 27: So the values can be any random values as long as they follow the pattern? 

Why does this one not have the `privateKey` item comparing with those in the other two methods?

jlamillan: It's missing the private key because it's a negative/error scenario test `test_error_get_oke_latest_versions_missing_key`.

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/rancher/pkg/auth/util"
+	"github.com/rancher/rancher/pkg/ref"
+	v1 "github.com/rancher/types/apis/core/v1"
+	"github.com/rancher/types/apis/management.cattle.io/v3/schema"
+	"github.com/rancher/types/config"
+	"github.com/rancher/types/namespace"
+
+	"encoding/json"
+)
```
dramich, line 25:
```suggestion
import (
	"encoding/json"
	"fmt"
	"io/ioutil"
	"net/http"

	"github.com/gorilla/mux"
	"github.com/oracle/oci-go-sdk/common"
	"github.com/rancher/norman/api/access"
	"github.com/rancher/norman/httperror"
	"github.com/rancher/norman/types"
	"github.com/rancher/rancher/pkg/auth/util"
	"github.com/rancher/rancher/pkg/ref"
	v1 "github.com/rancher/types/apis/core/v1"
	"github.com/rancher/types/apis/management.cattle.io/v3/schema"
	client "github.com/rancher/types/client/management/v3"
	"github.com/rancher/types/config"
	"github.com/rancher/types/namespace"
	"github.com/sirupsen/logrus"
)
```

dramich: If the spaces are deleted then goimports is ran, the output should look like this. Our normal style is two groups, standard library and everything else.

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,238 @@
...
+// Cloud Credential Secret Fields
+var requiredDataFields = map[string]string{
+	"fingerprint":        "ocicredentialConfig-fingerprint",
+	"tenancyId":          "ocicredentialConfig-tenancyId",
+	"userId":             "ocicredentialConfig-userId",
+	"privateKeyContents": "ocicredentialConfig-privateKeyContents",
+}
+
+type Handler struct {
+	Action        string
```
dramich, line 46: These don't need to be exported

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,238 @@
...
+			return
+		}
+		writer.Write(serialized)
+	default:
+		util.ReturnHTTPError(writer, req, httperror.NotFound.Status, "invalid endpoint "+resourceType)
+	}
+}
+
+// extractCreds attempts to extract the credentials from a given request either from the body or cloud credentials.
+func (handler *Handler) extractCreds(writer http.ResponseWriter, req *http.Request, creds *Credentials) (httperror.ErrorCode, error) {
```
dramich, line 139: `writer` is not used

pkg/api/customization/oci/handler.go
```diff
@@ -0,0 +1,238 @@
...
+	creds := Credentials{}
+	errCode, err := handler.extractCreds(writer, req, &creds)
+	if err != nil {
+		util.ReturnHTTPError(writer, req, errCode.Status, err.Error())
+		return
+	}
+
+	err = handler.validateCreds(&creds)
+	if err != nil {
+		util.ReturnHTTPError(writer, req, httperror.ServerError.Status, err.Error())
```
dramich, line 72: This needs to be updated to `InvalidBodyContent`, a user passing in bad/incomplete data is not a server error.

pkg/api/customization/oci/listers.go
```diff
@@ -0,0 +1,238 @@
...
+
+	"github.com/oracle/oci-go-sdk/common"
+	"github.com/oracle/oci-go-sdk/containerengine"
+	"github.com/oracle/oci-go-sdk/core"
+	"github.com/oracle/oci-go-sdk/identity"
+
+	"encoding/json"
+)
+
+func processVcns(provider common.ConfigurationProvider, compartment string) ([]byte, httperror.ErrorCode, error) {
```
dramich, line 20: All the helpers in here should be updated to return an int in place of httperror.ErrorCode, then in the error handling above that can be passed strait into the helpers.

pkg/api/customization/oci/listers.go
```diff
@@ -0,0 +1,238 @@
...
+	logrus.Debugf("[oci-handler] listing VCNs in compartment: %s", compartment)
+	virtualNetworkClient, err := core.NewVirtualNetworkClientWithConfigurationProvider(provider)
+	if err != nil {
+		logrus.Debugf("[oci-handler] error creating OCI Virtual Network client: %v", err)
+		return nil, httperror.ErrorCode{}, err
```
dramich, line 25: Using an empty struct is going to attempt to use a status of 0, all error returns need a valid code based on what happened.

pkg/api/customization/oci/listers.go
```diff
@@ -0,0 +1,238 @@
...
+		logrus.Debugf("[oci-handler] error listing VCNs with OCI Virtual Network client: %v", err)
+		return nil, httpErr, err
+	}
+
+	var vcnDisplayNames []string
+	for _, item := range vcnResponse.Items {
+		vcnDisplayNames = append(vcnDisplayNames, *(item.DisplayName))
+	}
+
+	data, err := json.Marshal(vcnDisplayNames)
```
dramich, line 48: The error isn't being checked here so the return below would be 0 if there is an error.

server/server.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"net/http"
 
+	"github.com/rancher/rancher/pkg/api/customization/oci"
```
dramich, line 7: Import formatting needs updated.

pkg/api/customization/oci/handler.go
jlamillan: I ran `goimports`. The CI job is also not complaining. Is there a particular version or flags I should be using?

```
jemillan-mac:rancher jemillan$ goimports -d pkg/api/customization/oci/handler.go  
jemillan-mac:rancher jemillan$ 
```

pkg/api/customization/oci/listers.go
```diff
@@ -0,0 +1,257 @@
...
+
+	return data, http.StatusOK, err
+}
+
+func processOkeVersions(provider common.ConfigurationProvider) ([]byte, int, error) {
+	logrus.Debug("[oci-handler] listing OKE versions")
+	containerClient, err := containerengine.NewContainerEngineClientWithConfigurationProvider(provider)
+	if err != nil {
+		logrus.Debugf("[oci-handler] error creating OCI ContainerEngine client: %v", err)
+		return nil, http.StatusOK, err
```
dramich, line 59: This is an error state, it should return an error

pkg/api/customization/oci/listers.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	getClusterOptionsReq := containerengine.GetClusterOptionsRequest{
+		ClusterOptionId: common.String("all"),
+	}
+	getClusterOptionsResp, err := containerClient.GetClusterOptions(context.Background(), getClusterOptionsReq)
+	if err != nil {
+		httpErr := httperror.ErrorCode{}
+		if getClusterOptionsResp.RawResponse != nil {
+			httpErr.Status = getClusterOptionsResp.RawResponse.StatusCode
+			httpErr.Code = getClusterOptionsResp.RawResponse.Status
```
dramich, line 69: This code isn't used so there is no reason to set it

pkg/api/customization/oci/listers.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	shapeResponse, err := computeClient.ListShapes(context.Background(), shapeRequest)
+	if err != nil {
+		logrus.Debugf("[oci-handler] error listing shapes with OCI Compute client: %v", err)
+		return nil, getErrorCode(shapeResponse.RawResponse), err
+	}
+
+	var nodeShapes []string
+	for _, item := range shapeResponse.Items {
+		if listContains(nodeShapes, *(item.Shape)) == false {
```
dramich, line 162:
```suggestion
		if !listContains(nodeShapes, *(item.Shape)) {
```

PR #26974, opened 2020-05-07
[2.3] Fix resource_quota_sync deepEqual bug

dramich:


maggieliu:
related to this issue: https://github.com/rancher/rancher/issues/26976

PR #26982, opened 2020-05-08
Added Basic volume create, edit and delete, rbac cases

soumyalj:
Please review

### DIFF
--- a/tests/validation/tests/v3_api/test_volumes.py
+++ b/tests/validation/tests/v3_api/test_volumes.py
@@ -1,0 +1,547 @@
+import pytest
+from rancher import ApiError
+from .common import *  # NOQA
+
+namespace = {"p_client": None,
+             "ns": None,
+             "cluster": None,
+             "project": None,
+             "pv": None,
+             "pvc": None}
+
+# this is the path to the mounted dir in the pod(workload)
+MOUNT_PATH = "/var/nfs"
+# if True then delete the NFS after finishing tests, otherwise False
+DELETE_NFS = eval(os.environ.get('RANCHER_DELETE_NFS', "True"))
+
+
+rbac_role_list = [
+    (CLUSTER_OWNER),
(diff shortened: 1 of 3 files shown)


tests/validation/tests/v3_api/test_volumes.py
```diff
@@ -0,0 +1,581 @@
...
+        print(testdata)
+        while pvcname in testdata[0]:
+            if time.time() - start > timeout:
+                raise AssertionError("Timed out waiting for deletion")
+            time.sleep(.5)
+            pvcdict = client.list_persistentVolumeClaim(name=pvcname)
+            testdata = pvcdict.get('data')
+        assert True
+    if len(pvcdict.get('data')) == 0:
+        assert True
```
jiaqiluo, line 483: Can we change it to something like
```
assert len(pvcdict.get('data')) == 0, "failed to delete the pvc"

File: tests/validation/tests/v3_api/test_volumes.py [hunk 2]
```diff
@@ -0,0 +1,581 @@
...
+    command = "get pvc {} --namespace {}".format(pvc['name'], ns.name)
+    print("Command to obtain the pvc")
+    print(command)
+    result = execute_kubectl_cmd(command, json_out=False, stderr=True)
+    print(result)
+
+    print("Verify that the pvc does not exist "
+          "and the error code returned is non zero ")
+    if result != 0:
+        assert True
```
jiaqiluo, line 495: something like 
```
assert result != 0, "expect to not see 0"
```

tests/validation/tests/v3_api/test_volumes.py
soumyalj: Don

tests/validation/tests/v3_api/test_volumes.py
soumyalj: Done

PR #27001, opened 2020-05-08
support removing rancher ha setup from AWS

jiaqiluo:
This PR enables us to delete the rancher ha setup from AWS
it does the following:
- delete clusters within Rancher server 
- delete instances, target groups, load balancers, database, and route53 records in AWS

the only parameter needed is `CATTLE_TEST_URL`

Update:
add support for removing a k3s setup from AWS by resource name prefix  `RANCHER_RESOURCE_NAME`

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -1147,14 +1147,11 @@ def get_cluster_type(client, cluster):
 
 def delete_cluster(client, cluster):
     nodes = client.list_node(clusterId=cluster.id).data
-    # Delete Cluster
-    client.delete(cluster)
     # Delete nodes(in cluster) from AWS for Imported and Custom Cluster
     if len(nodes) > 0:
         cluster_type = get_cluster_type(client, cluster)
         print(cluster_type)
         if get_cluster_type(client, cluster) in ["Imported", "Custom"]:
-            nodes = client.list_node(clusterId=cluster.id).data
             filters = [
                 {'Name': 'tag:Name',
                  'Values': ['testcustom*', 'teststress*', 'testsa*']}]
@@ -1170,9 +1167,17 @@ def delete_cluster(client, cluster):
             assert len(ip_filter) > 0
             print(ip_filter)
             aws_nodes = AmazonWebServices().get_nodes(filters)
-            for node in aws_nodes:
-                print(node.public_ip_address)
+            if aws_nodes is None:
+                # search instances by IPs in case names do not follow patterns
+                aws_nodes = AmazonWebServices().get_nodes(filters=[ip_filter])
+            if aws_nodes is None:
+                print("no instance is found in AWS")
+            else:
+                for node in aws_nodes:
+                    print(node.public_ip_address)
                 AmazonWebServices().delete_nodes(aws_nodes)
+    # Delete Cluster
+    client.delete(cluster)
 
 
 def check_connectivity_between_workloads(p_client1, workload1, p_client2,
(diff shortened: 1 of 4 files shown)


tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -48,6 +53,60 @@
...
+
+    # delete nodes
+    node_filter = [{
+        'Name': 'tag:Name',
+        'Values': [resource_prefix + "-*"]
+    }]
+    nodes = AmazonWebServices().get_nodes(filters=node_filter)
+    print("deleting the following instances: {}"
+          .format([node.public_ip_address for node in nodes]))
+    AmazonWebServices().delete_nodes(nodes)
```
soumyalj, line 77: After the node deletion, can we assert the length of nodes list is zero ? 
If the node deletion fails, can we retry after a short time and exit with a proper message ?

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -48,6 +53,60 @@
...
+    # delete rds
+    db_name = resource_prefix + "-multinode-db"
+    print("deleting the database: {}".format(db_name))
+    AmazonWebServices().delete_db(db_name)
+
+    # delete the route 53 record
+    record_name = CATTLE_TEST_URL.split("//")[1] + "."
+    print("deleting the route53 record: {}".format(record_name))
+    AmazonWebServices().delete_route_53_record(record_name)
+
```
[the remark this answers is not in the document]
jiaqiluo: same here, the best effort we can do is to send out the deletion request to AWS, we don't need to check if the deletion succeeds or not.

tests/validation/tests/v3_api/test_create_ha.py
jiaqiluo: The best effort we can do is to send out the deletion request to AWS, we don't need to check if the deletion succeeds or not.

soumyalj:
LGTM

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -48,6 +53,60 @@
 export_cmd = "export KUBECONFIG=" + kubeconfig_path
 
 
+def test_remove_rancher_ha():
+    resource_prefix = \
+        CATTLE_TEST_URL.split(".qa.rancher.space")[0].split("//")[1]
```
ShylajaDevadiga, line 58: CATTLE_TEST_URL is an IP eg https://1.2.3.4

ShylajaDevadiga: Would using RESOURCE_PREFIX instead of CATTLE_TEST_URL help in deleting k3s clusters too using same method?

jiaqiluo: this method is mainly used for deleting all resources coming from the auto-created HA setup.
We can have a separate method and job to handle the deletion of a k3s cluster.

tests/validation/tests/v3_api/common.py
```diff
@@ -1170,9 +1170,11 @@ def delete_cluster(client, cluster):
             assert len(ip_filter) > 0
             print(ip_filter)
             aws_nodes = AmazonWebServices().get_nodes(filters)
+            if aws_nodes is None:
+                aws_nodes = AmazonWebServices().get_nodes(filters=[ip_filter])
```
ShylajaDevadiga, line 1174: Could we add a check here if nodes are found?

soumyalj: assert  len(aws_nodes) > 0

tests/validation/tests/v3_api/common.py
```diff
@@ -2525,3 +2530,55 @@ def create_pv(client, nfs_ip):
...
     assert pv_object['type'] == 'persistentVolume'
     return pv_object
+
+
+def delete_resource_in_AWS_by_prefix(resource_prefix):
+    """
+    :param resource_prefix: the prefix of resource name
+    :return: None
+    """
+    # delete nodes
```
soumyalj, line 2540: Can we add comment that nodes of the local cluster are also deleted here ?

tests/validation/tests/v3_api/common.py
jiaqiluo: @soumyalj an assertion may break the workflow. Instead, it will just print no instance is found in AWS and skip deleting instances, and go to the next step.

tests/validation/tests/v3_api/common.py
jiaqiluo: comment is added

ShylajaDevadiga:
LGTM

PR #27006, opened 2020-05-09
Add aks, gke, and eks to support matrix cluster deploy tests

rancher-max:


tests/validation/tests/v3_api/test_aks_cluster.py
```diff
@@ -53,26 +58,21 @@ def get_aks_version():
         print("JSON RESPONSE IS")
         print(response.content)
         json_response = json.loads(response.content)
-        versionarray_length = len(json_response)
-        aksclusterversion = json_response[versionarray_length-1]
-        print(aksclusterversion)
+
+        if multiple_versions:
+            aksclusterversion = [json_response[0], json_response[-1]]
```
jiaqiluo, line 63: Is it possible there is only one version in the response? Should we handle that case?

tests/validation/tests/v3_api/test_deploy_clusters.py
```diff
@@ -51,20 +60,99 @@ def test_deploy_rke():
...
-    env_details += "'"
     create_config_file(env_details)
 
 
 @if_not_auto_deploy_eks
 def test_deploy_eks():
     print("Deploying EKS Clusters")
+    global env_details
+    errors = []
+    k8s_versions = [EKS_K8S_VERSIONS[0], EKS_K8S_VERSIONS[-1]]
```
jiaqiluo, line 71: So we try only the first and the last version in the list?

rancher-max: Yes for the support matrix testing we only want to do the first and last versions for these 3 providers (gke, eks, aks).

tests/validation/tests/v3_api/test_aks_cluster.py
rancher-max: You're right, it probably won't ever happen but I think might be a possibility. I'll update all 3 of them to account for that!

tests/validation/tests/v3_api/test_deploy_clusters.py
```diff
@@ -51,20 +60,99 @@ def test_deploy_rke():
...
+                "refresh-interval-minutes": "1440",
+                "url": KDM_URL
+            })
+        })
+        assert r.ok
+
+        # Refresh Kubernetes Metadata
+        kdm_refresh_url = CATTLE_API_URL + "/kontainerdrivers?action=refresh"
+        response = requests.post(kdm_refresh_url, verify=False, headers=header)
+        assert response.ok
```
soumyalj, line 153: Can we obtain response.content after the post request and add asserts to check for the right values ?
json.loads(response.content) can be used if required.

tests/validation/tests/v3_api/test_deploy_clusters.py
```diff
@@ -51,20 +60,99 @@ def test_deploy_rke():
...
+        kdm_url = CATTLE_API_URL + "/settings/rke-metadata-config"
+        r = requests.put(kdm_url, verify=False, headers=header, json={
+            "name": "rke-metadata-config",
+            "value": json.dumps({
+                "branch": KDM_BRANCH,
+                "refresh-interval-minutes": "1440",
+                "url": KDM_URL
+            })
+        })
+        assert r.ok
```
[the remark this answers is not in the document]
rancher-max: Fixed!

tests/validation/tests/v3_api/test_deploy_clusters.py
rancher-max: This value has `None` in response.content. I think for this one just checking that the response is 200 is enough since in the UI flow just a button click is all it is and it doesn't really give any "success" message. Let me know your thoughts!

tests/validation/tests/v3_api/test_aks_cluster.py
rancher-max: Fixed!

tests/validation/tests/v3_api/test_deploy_clusters.py
soumyalj: ok

soumyalj:
LGTM

Issue #26969, opened 2020-05-07
2.4.3 upgrade monitoring to 0.1.0 fails


eroji:
Just upgraded Rancher to 2.4.3 from 2.3.6. The clusters I have were previously running on 1.7.4 and I have upgraded them to 1.7.5. It indicated there is a monitoring upgrade to 0.1.0. I was previously running 0.7.0 without any issues. However, when proceeded with the upgrade, the `prometheus-cluster-monitoring` statefulset never comes up on and is constantly created then removed. It's not quite clear what the issue is. All the sidecars in the pod comes up healthy except `prometheus-agent`. I'm running CentOS 7 VMs and Docker 19.3.1/19.3.8.

I am using persistent storage. Of the clusters I have that I tried performing the upgrade, one is using Rook/Ceph and one is on AWS using EBS. Both exhibit the same exact problem. However, if I turn off persistent storage it comes online fine.

teknologista:
Same issue here:

Logs from the pod:
`level=info ts=2020-05-07T17:32:26.461Z caller=main.go:329 msg="Starting Prometheus" version="(version=2.11.1, branch=HEAD, revision=e5b22494857deca4b806f74f6e3a6ee30c251763)"
level=info ts=2020-05-07T17:32:26.461Z caller=main.go:330 build_context="(go=go1.12.7, user=root@d94406f2bb6f, date=20190710-13:51:17)"
level=info ts=2020-05-07T17:32:26.461Z caller=main.go:331 host_details="(Linux 4.15.0-76-generic #86-Ubuntu SMP Fri Jan 17 17:24:28 UTC 2020 x86_64 prometheus-cluster-monitoring-0 (none))"
level=info ts=2020-05-07T17:32:26.461Z caller=main.go:332 fd_limits="(soft=1048576, hard=1048576)"
level=info ts=2020-05-07T17:32:26.461Z caller=main.go:333 vm_limits="(soft=unlimited, hard=unlimited)"
level=info ts=2020-05-07T17:32:26.463Z caller=main.go:652 msg="Starting TSDB ..."
level=info ts=2020-05-07T17:32:26.463Z caller=web.go:448 component=web msg="Start listening for connections" address=127.0.0.1:9090
level=info ts=2020-05-07T17:32:26.488Z caller=main.go:667 fs_type=EXT4_SUPER_MAGIC
level=info ts=2020-05-07T17:32:26.488Z caller=main.go:668 msg="TSDB started"
[...]

PR #27032, opened 2020-05-11
Update monitoring images

StrongMonkey:


StrongMonkey:
Address https://github.com/rancher/rancher/issues/26969. We are going to bump monitoring charts to match upstream image version.

Issue #26692, opened 2020-04-17
Monitoring - Monitoring API is not ready after apps are active 


jiaqiluo:
this is found during the testing this issue: https://github.com/rancher/rancher/issues/26440

Steps:
- run `Rancher:v2.4-head` single install 
- add a cluster with cluster monitoring enabled by editing the cluster as YAML file
- add a second cluster, enable the cluster monitoring after the cluster is active

Results:

When enabling the cluster monitoring in the provisioning configuration, the UI will keep showing the `Monitoring API is not ready` message after the apps are active, and it takes **another 3 to 5 mins** to see the message disappear and metrics section shows up. --> This behavior is **not** seen on `rancher:v2.3.6` single install. 

While if the cluster monitoring is enabled after the cluster is active, the `Monitoring API is not ready` message will disappear and the metrics section shows up as soon as the apps are active. --> This behavior is the same on `rancher:v2.3.6` single install.

lots0logs:
I'm using v2.4.2 and while I can see that the monitoring apps are running, the dashboard continues to show "Monitoring API is not ready" message. It's been over 24 hours and the message is still there. I even tried rebooting the hosts. Please let me know what data you need to identify the issue.

jiaqiluo:
This happens in an imported multiple-node k3s cluster (v1.17.4+k3s1) in a `rancher:v2.4.3-rc2` single-install setup. 

it takes a long time for the metrics showing in the UI after the apps are active

StrongMonkey:
https://github.com/rancher/rancher/issues/26811#issuecomment-627679570

jiaqiluo:
The bug fix is validated on Rancher
- `v2.4-3036-head` and
-  `master-3034-head` 

with monitoring chart version `0.1.1`  


The cluster monitoring is tested on clusters with different k8s versions and different kinds of clusters on different scenarios while validating other monitoring-related issues. The metrics show up in Rancher UI right after the monitoring apps are active at most times. 

So I am confident to say that the bug is fixed.

deviops-com:
Hi,

Issue #26811, opened 2020-04-25
Imported K3s cluster - Monitoring API is not ready after apps are active


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- run Rancher v2.4.3-rc2 single install
- imported a k3s cluster `v1.17.4+k3s1`
- enable the cluster monitoring 

**Result:**
- after the apps are available, the UI still shows waiting for the API to be ready 
- and it waits a long time to show the graphs in the UI 
- this is the same behavior we saw here: https://github.com/rancher/rancher/issues/26692

**Other details that may be helpful:**
- this is not seen in an imported k3s cluster with an older version (v1.16.3-k3s.2)

StrongMonkey:
@jiaqiluo I can't reproduce it when testing 2.4-head. Are you always able to reproduce this?

StrongMonkey:
Also notice that there are failure scenario where the controller takes maximum 2 minutes to re-sync. So any status cloud be refreshed after 2 minutes.

StrongMonkey:
Actually I am still struggling at reproducing this issue. But https://github.com/rancher/rancher/pull/27054 should potentically fix it. I think the general issue right here is not related to the underlying k3s version. The controller inside rancher has to fetch the download prometheus instances and update cluster status accordingly. While the controller follow the back-off algorithm to sync status, it will eventually fallback to 2 minutes(or other maximum interval). So there is a delay between the real status and the status showing on UI.

@jiaqiluo Let me know if you see that the cluster stays in `not ready` states and never returns to a normal state. My understanding is that it will recover after a certain period. This could happen a lot when cluster is not ready(for example, launched with automation scripts). 

cc @maggieliu

jiaqiluo:
The bug fix is validated on Rancher
- `v2.4-3036-head` and
-  `master-3034-head` 

with monitoring chart version `0.1.1`  

Steps:
- run Rancher server, single install
- edit the system chart to point to branch `dev-v2.5`
- import a k3s cluster into Rancher, the k3s cluster has 2 servers and 2 workers, and it version is `v1.17.4+k3s1`
- enable the cluster monitoring  `0.1.1`  

Results:
The cluster page shows up graphs right after the monitoring apps are available under system projects 

![Screenshot_2020-05-26 Rancher](https://user-images.githubusercontent.com/6218999/82948800-54afcd00-9f57-11ea-8054-7437598329be.png)

PR #27054, opened 2020-05-13
Always return errors if status is not ready

StrongMonkey:
This commit fixed an issue where an error will always be returned so
that controller will try to re-sync the status on the next run


StrongMonkey:
Potentially fix https://github.com/rancher/rancher/issues/26692 and https://github.com/rancher/rancher/issues/26811.

thxCode:
In fact, there is an endpoint detection of cluster-level Prometheus in https://github.com/rancher/rancher/blob/dd102af1773e574f40a5811a85151431c3e3ec4b/pkg/controllers/user/monitoring/cluster_monitoring_enabled_handler.go#L25-L43, so we don't return an error to enqueue again.

StrongMonkey:
@thxCode thanks, didn't know this. I ll close this PR.

PR #27056, opened 2020-05-13
Default serviceaccounts of system namespaces should not be used

prachidamle:
This change is made for the following CIS 1.5 test:
5.1.5 - Ensure that default service accounts are not actively used.

We need to make sure that an RKE cluster out of the box does not use the default serviceaccounts. Thus rancher system services (launched in any of the system namespaces) should not use default serviceaccount. 
We will exclude kube-system and default namespace since they can be used by Kubernetes and we need not control the SA usage in them.

Adding new controller to make sure default serviceaccounts of system namespaces have automountServiceAccountToken flag set to false.

### DIFF
--- a/pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
+++ b/pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
@@ -1,0 +1,82 @@
+package nsserviceaccount
+
+import (
+	"context"
+	"strings"
+
+	"github.com/rancher/rancher/pkg/settings"
+	rv1 "github.com/rancher/types/apis/core/v1"
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type namespaceSvcAccountHandler struct {
+	serviceAccountLister rv1.ServiceAccountLister
+	serviceAccounts      rv1.ServiceAccountInterface
+}
+
+func Register(ctx context.Context, cluster *config.UserContext) {
+	logrus.Debugf("Registering namespaceSvcAccountHandler for checking default serviceaccount")
+	nsh := &namespaceSvcAccountHandler{
+		serviceAccountLister: cluster.Core.ServiceAccounts("").Controller().Lister(),
+		serviceAccounts:      cluster.Core.ServiceAccounts(""),
+	}
+	cluster.Core.Namespaces("").AddHandler(ctx, "namespaceSvcAccountHandler", nsh.Sync)
+}
+
+func (nsh *namespaceSvcAccountHandler) Sync(key string, ns *corev1.Namespace) (runtime.Object, error) {
+	if ns == nil || ns.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	logrus.Debugf("namespaceSvcAccountHandler: Sync namespace: key=%v", key)
+
+	//handle default svcAccount of system namespaces
+	if err := nsh.handleSystemNS(key); err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: Sync: error handling default ServiceAccount of namespace key=%v, err=%v", key, err)
+	}
+	return nil, nil
+}
+
+func (nsh *namespaceSvcAccountHandler) handleSystemNS(namespace string) error {
+	if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
+		return nil
+	}
+	defSvcAccnt, err := nsh.serviceAccountLister.Get(namespace, "default")
+	if err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: error listing serviceaccount flag: Sync: key=%v, err=%+v", namespace, err)
+		return err
+	}
+
+	if defSvcAccnt.AutomountServiceAccountToken != nil && *defSvcAccnt.AutomountServiceAccountToken == false {
+		return nil
+	}
+
+	defSvcAccntCopy := defSvcAccnt.DeepCopy()
+	automountServiceAccountToken := false
+	defSvcAccntCopy.AutomountServiceAccountToken = &automountServiceAccountToken
+	logrus.Debugf("namespaceSvcAccountHandler: updating default serviceaccount key=%v", defSvcAccntCopy)
+	_, err = nsh.serviceAccounts.Update(defSvcAccntCopy)
+	if err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: error updating serviceaccnt flag: Sync: key=%v, err=%+v", namespace, err)
+		return err
+	}
+	return nil
+}
+
+func (nsh *namespaceSvcAccountHandler) isSystemNS(namespace string) bool {
+	systemNamespacesStr := settings.SystemNamespaces.Get()
+	if systemNamespacesStr == "" {
+		return false
+	}
+
+	systemNamespaces := make(map[string]bool)
+	splitted := strings.Split(systemNamespacesStr, ",")
+	for _, s := range splitted {
+		ns := strings.TrimSpace(s)
+		systemNamespaces[ns] = true
+	}
+
+	return systemNamespaces[namespace]
+}
--- a/pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
+++ b/pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strings"
 
+	"github.com/rancher/norman/types/slice"
 	"github.com/rancher/rancher/pkg/settings"
 	rv1 "github.com/rancher/types/apis/core/v1"
 	"github.com/rancher/types/config"
@@ -12,71 +13,59 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-type namespaceSvcAccountHandler struct {
-	serviceAccountLister rv1.ServiceAccountLister
-	serviceAccounts      rv1.ServiceAccountInterface
+type defaultSvcAccountHandler struct {
+	serviceAccounts rv1.ServiceAccountInterface
 }
 
 func Register(ctx context.Context, cluster *config.UserContext) {
-	logrus.Debugf("Registering namespaceSvcAccountHandler for checking default serviceaccount")
-	nsh := &namespaceSvcAccountHandler{
-		serviceAccountLister: cluster.Core.ServiceAccounts("").Controller().Lister(),
-		serviceAccounts:      cluster.Core.ServiceAccounts(""),
+	logrus.Debugf("Registering defaultSvcAccountHandler for checking default service account of system namespaces")
+	nsh := &defaultSvcAccountHandler{
+		serviceAccounts: cluster.Core.ServiceAccounts(""),
 	}
-	cluster.Core.Namespaces("").AddHandler(ctx, "namespaceSvcAccountHandler", nsh.Sync)
+	cluster.Core.ServiceAccounts("").AddHandler(ctx, "defaultSvcAccountHandler", nsh.Sync)
 }
 
-func (nsh *namespaceSvcAccountHandler) Sync(key string, ns *corev1.Namespace) (runtime.Object, error) {
-	if ns == nil || ns.DeletionTimestamp != nil {
+func (nsh *defaultSvcAccountHandler) Sync(key string, sa *corev1.ServiceAccount) (runtime.Object, error) {
+	if sa == nil || sa.DeletionTimestamp != nil {
 		return nil, nil
 	}
-	logrus.Debugf("namespaceSvcAccountHandler: Sync namespace: key=%v", key)
-
-	//handle default svcAccount of system namespaces
-	if err := nsh.handleSystemNS(key); err != nil {
-		logrus.Errorf("namespaceSvcAccountHandler: Sync: error handling default ServiceAccount of namespace key=%v, err=%v", key, err)
+	logrus.Debugf("defaultSvcAccountHandler: Sync service account: key=%v", key)
+	//handle default svcAccount of system namespaces only
+	if err := nsh.handleIfSystemNSDefaultSA(sa); err != nil {
+		logrus.Errorf("defaultSvcAccountHandler: Sync: error handling default ServiceAccount of namespace key=%v, err=%v", key, err)
 	}
 	return nil, nil
 }
 
-func (nsh *namespaceSvcAccountHandler) handleSystemNS(namespace string) error {
+func (nsh *defaultSvcAccountHandler) handleIfSystemNSDefaultSA(defSvcAccnt *corev1.ServiceAccount) error {
+	//check if sa is "default"
+	if defSvcAccnt.Name != "default" {
+		return nil
+	}
+	//check if ns is a system-ns
+	namespace := defSvcAccnt.Namespace
 	if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
 		return nil
 	}
-	defSvcAccnt, err := nsh.serviceAccountLister.Get(namespace, "default")
-	if err != nil {
-		logrus.Errorf("namespaceSvcAccountHandler: error listing serviceaccount flag: Sync: key=%v, err=%+v", namespace, err)
-		return err
-	}
-
 	if defSvcAccnt.AutomountServiceAccountToken != nil && *defSvcAccnt.AutomountServiceAccountToken == false {
 		return nil
 	}
-
-	defSvcAccntCopy := defSvcAccnt.DeepCopy()
 	automountServiceAccountToken := false
-	defSvcAccntCopy.AutomountServiceAccountToken = &automountServiceAccountToken
-	logrus.Debugf("namespaceSvcAccountHandler: updating default serviceaccount key=%v", defSvcAccntCopy)
-	_, err = nsh.serviceAccounts.Update(defSvcAccntCopy)
+	defSvcAccnt.AutomountServiceAccountToken = &automountServiceAccountToken
+	logrus.Debugf("defaultSvcAccountHandler: updating default service account key=%v", defSvcAccnt)
+	_, err := nsh.serviceAccounts.Update(defSvcAccnt)
 	if err != nil {
-		logrus.Errorf("namespaceSvcAccountHandler: error updating serviceaccnt flag: Sync: key=%v, err=%+v", namespace, err)
+		logrus.Errorf("defaultSvcAccountHandler: error updating default service account flag for namespace: %v, err=%+v", namespace, err)
 		return err
 	}
 	return nil
 }
 
-func (nsh *namespaceSvcAccountHandler) isSystemNS(namespace string) bool {
+func (nsh *defaultSvcAccountHandler) isSystemNS(namespace string) bool {
 	systemNamespacesStr := settings.SystemNamespaces.Get()
 	if systemNamespacesStr == "" {
 		return false
 	}
-
-	systemNamespaces := make(map[string]bool)
-	splitted := strings.Split(systemNamespacesStr, ",")
-	for _, s := range splitted {
-		ns := strings.TrimSpace(s)
-		systemNamespaces[ns] = true
-	}
-
-	return systemNamespaces[namespace]
+	systemNamespaces := strings.Split(systemNamespacesStr, ",")
+	return slice.ContainsString(systemNamespaces, namespace)
 }
--- a/tests/integration/suite/test_system_project.py
+++ b/tests/integration/suite/test_system_project.py
@@ -7,7 +7,8 @@ initial_system_namespaces = set(["kube-node-lease",
                                  "kube-system",
                                  "cattle-system",
                                  "kube-public",
-                                 "cattle-global-data"])
+                                 "cattle-global-data",
+                                 "cattle-global-nt"])
 loggingNamespace = "cattle-logging"
 
 
--- a/tests/integration/suite/test_system_project.py
+++ b/tests/integration/suite/test_system_project.py
@@ -1,5 +1,7 @@
 import pytest
 from rancher import ApiError
+from kubernetes.client import CoreV1Api
+from .conftest import wait_for
 
 systemProjectLabel = "authz.management.cattle.io/system-project"
 defaultProjectLabel = "authz.management.cattle.io/default-project"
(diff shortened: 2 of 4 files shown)


pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: Sync: error handling default ServiceAccount of namespace key=%v, err=%v", key, err)
+		return nil, nil
+	}
+	return nil, nil
+}
+
+func (nsh *namespaceSvcAccountHandler) handleSystemNS(namespace string) error {
+
+	if (namespace != "kube-system" && namespace != "default") && nsh.isSystemNS(namespace) {
```
aiyengar2, line 46: nit: Can we change this to:
```
if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
    return nil
}
// Remaining code unindented
```

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,84 @@
...
+		var svcAccounts clientv1.ServiceAccountInterface
+		svcAccounts = nsh.clusterCtx.K8sClient.CoreV1().ServiceAccounts(namespace)
+
+		defSvcAccnt, err := svcAccounts.Get("default", metav1.GetOptions{})
+		if err != nil {
+			logrus.Errorf("namespaceSvcAccountHandler: error listing serviceaccount flag: Sync: key=%v, err=%+v", namespace, err)
+			return err
+		}
+
+		if defSvcAccnt.AutomountServiceAccountToken == nil || *defSvcAccnt.AutomountServiceAccountToken == true {
```
aiyengar2, line 56: nit: can we change this to
```
if defSvcAccnt.AutomountServiceAccountToken != nil && *defSvcAccnt.AutomountServiceAccountToken == false {
    return nil
}
// Remaining code unindented
```

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+	if ns == nil {
+		return nil, nil
+	}
+	logrus.Debugf("namespaceSvcAccountHandler: Sync: key=%v, ns=%+v", key, *ns)
+
+	//handle default svcAccount of system namespaces
+	err := nsh.handleSystemNS(key)
+	if err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: Sync: error handling default ServiceAccount of namespace key=%v, err=%v", key, err)
+		return nil, nil
```
aiyengar2, line 39: nit: Should this be returning an error? If not, we can just omit this line as it already returns `nil, nil`

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (nsh *namespaceSvcAccountHandler) Sync(key string, ns *corev1.Namespace) (runtime.Object, error) {
+	if ns == nil {
+		return nil, nil
+	}
+	logrus.Debugf("namespaceSvcAccountHandler: Sync: key=%v, ns=%+v", key, *ns)
+
+	//handle default svcAccount of system namespaces
+	err := nsh.handleSystemNS(key)
+	if err != nil {
```
aiyengar2, line 37: nit:
```
if err := nsh.handleSystemNS(key); err != nil {
```

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+func Register(ctx context.Context, cluster *config.UserContext) {
+	logrus.Debugf("Registering namespaceHandler for checking default serviceaccount")
+	nsh := &namespaceSvcAccountHandler{
+		clusterCtx: cluster,
+	}
+	cluster.Core.Namespaces("").AddHandler(ctx, "namespaceHandler", nsh.Sync)
+}
+
+func (nsh *namespaceSvcAccountHandler) Sync(key string, ns *corev1.Namespace) (runtime.Object, error) {
+	if ns == nil {
```
mrajashree, line 30: Do we need to check for deletion timestamp?

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+func (nsh *namespaceSvcAccountHandler) handleSystemNS(namespace string) error {
+
+	if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
+		return nil
+	}
+
+	var svcAccounts clientv1.ServiceAccountInterface
+	svcAccounts = nsh.clusterCtx.K8sClient.CoreV1().ServiceAccounts(namespace)
+
+	defSvcAccnt, err := svcAccounts.Get("default", metav1.GetOptions{})
```
mrajashree, line 53: can we use lister?

prachidamle: Lister does not work; had tried using it - mostly because I am working with core V1 resource; so also have to use K8sClient() to make the requests.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: Will add, mostly not needed since we only update if its system ns- but sounds correct to add.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: Yeah will remove this; we need not return error here since its not something we should try over and over aggressively - the periodic controller sync should call it again.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if defSvcAccnt.AutomountServiceAccountToken != nil && *defSvcAccnt.AutomountServiceAccountToken == false {
+		return nil
+	}
+
+	automountServiceAccountToken := false
+	defSvcAccnt.AutomountServiceAccountToken = &automountServiceAccountToken
+	defSvcAccnt.Namespace = namespace
+	logrus.Debugf("namespaceSvcAccountHandler: updating default serviceaccount key=%v", defSvcAccnt)
```
cbron, line 64: Perhaps this should be logrus.info ? If they setup a serviceAccount and then we change it they could get confused or it could break something. Although they shouldn't be doing that in system ns. But at least if it's in the logs they would get a warning. Up for debate. 

Also might want to specify what changed, ie:
> setting automount to false on default serviceaccount...

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: error updating serviceaccnt flag: Sync: key=%v, err=%+v", namespace, err)
+		return err
+	}
+
+	return nil
+}
+
+func (nsh *namespaceSvcAccountHandler) isSystemNS(namespace string) bool {
+	systemNamespacesStr := settings.SystemNamespaces.Get()
```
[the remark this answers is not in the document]
dramich: Well, sure. But what does the check get you that just processing the string doesn't? Either way we return false and don't touch the namespace. If someone sets it to blank they have much bigger problems.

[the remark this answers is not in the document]
cbron: It would buy you an error as opposed to a false negative. But agree that it's not really a bad outcome and there would be bigger problems than this, hence the nit.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: @cbron We are only changing "default" serviceaccounts or rancher system namespaces. So I think logrus.debugf should be fine since we are not changing any user namespace or serviceaccount?

cbron: Yea, its very unlikely to affect a customer. If they do mess with it they should know we are prone to change our own system.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type namespaceSvcAccountHandler struct {
+	clusterCtx *config.UserContext
```
dramich, line 17: The whole context should not be stored, update to use ServiceAccountLister

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func Register(ctx context.Context, cluster *config.UserContext) {
+	logrus.Debugf("Registering namespaceHandler for checking default serviceaccount")
+	nsh := &namespaceSvcAccountHandler{
+		clusterCtx: cluster,
+	}
+	cluster.Core.Namespaces("").AddHandler(ctx, "namespaceHandler", nsh.Sync)
```
dramich, line 25: `namespaceHandler` is a duplicate name

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (nsh *namespaceSvcAccountHandler) Sync(key string, ns *corev1.Namespace) (runtime.Object, error) {
+	if ns == nil || ns.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	logrus.Debugf("namespaceSvcAccountHandler: Sync: key=%v, ns=%+v", key, *ns)
```
dramich, line 32: Can just the name be printed here? This controller is really looking at the serviceAccount so not sure what the value of printing the entire NS is

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+
+	//handle default svcAccount of system namespaces
+	if err := nsh.handleSystemNS(key); err != nil {
+		logrus.Errorf("namespaceSvcAccountHandler: Sync: error handling default ServiceAccount of namespace key=%v, err=%v", key, err)
+	}
+	return nil, nil
+}
+
+func (nsh *namespaceSvcAccountHandler) handleSystemNS(namespace string) error {
+
```
dramich, line 42: Extra space here

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+	if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
+		return nil
+	}
+
+	var svcAccounts clientv1.ServiceAccountInterface
+	svcAccounts = nsh.clusterCtx.K8sClient.CoreV1().ServiceAccounts(namespace)
```
dramich, line 48: Update to use a lister

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
dramich: And a client for the update call

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	if defSvcAccnt.AutomountServiceAccountToken != nil && *defSvcAccnt.AutomountServiceAccountToken == false {
+		return nil
+	}
+
+	automountServiceAccountToken := false
+	defSvcAccnt.AutomountServiceAccountToken = &automountServiceAccountToken
+	defSvcAccnt.Namespace = namespace
```
dramich, line 62: This field should already be set to the correct namespace

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: LIster does not work for the v1.ServiceAccount object _ i tried using it earlier

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: done

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: lister did not work here

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: removed

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: changed

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
prachidamle: Specifically the update does not work without using the 
clientv1.ServiceAccountInterface svcAccounts = clusterCtx.K8sClient.CoreV1().ServiceAccounts(namespace)

Also notice that since we need to use this v1 interface, to get the  default serviceaccount of a particular namespace by name, we have to pass in the "namespace" and create a new instance of this interface each time so that the following Get call works.

defSvcAccnt, err := svcAccounts.Get(context.TODO(), "default", metav1.GetOptions{})

dramich: Why doesn't a lister work?

prachidamle: @dramich I tested again changing to lister and types interface and was able to see the updates go through! Not sure why it did not work earlier - I think possibly I was using the userOnlyContext during that test. But now updating to use rancher types instead of client-go K8sClient.CoreV1()

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return nil, nil
+}
+
+func (nsh *namespaceSvcAccountHandler) handleSystemNS(namespace string) error {
+	if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
+		return nil
+	}
+	defSvcAccnt, err := nsh.serviceAccountLister.Get(namespace, "default")
+	if err != nil {
```
cbron, line 47: We aren't checking for NotFound error here, would there be some system-namespaces where the default SA won't exist ?

prachidamle: Even if there is any such system ns where the default SA does not exist, there is no need to error out - we will just skip that ns. The main function of the controller is to update AutomountServiceAccountToken=false for any default SA found for any system NS.

cbron: But if a system-ns without a default SA exists, won't that error out here with a NotFound error ? Since the handler is off of the NS type and not off of the service-account type, we won't know until we get here.

prachidamle: I think we can check if it is a NotFound error and just log this at INFO. We have these cases that fall into this category:
a) a system NS with NO default SA at all- we will just log this at INFO in each sync and not log at error.
b) a system NS which has a default SA but lister failed to list - We will be logging at INFO and hoping next periodic sync should list this when lister cache gets updated.

cbron: Never mind on this whole thread, I was under false assumption that a default SA might not exist. Looks like it always does.

tests/integration/suite/test_system_project.py
```diff
@@ -71,3 +73,14 @@ def test_system_project_cant_be_deleted(admin_mc, admin_cc):
         admin_mc.client.delete(system_project)
     assert e.value.error.status == 405
     assert e.value.error.message == 'System Project cannot be deleted'
+
+
+def test_system_namespaces_default_svc_account(admin_mc):
```
dramich, line 78: The test should use the rancher setting `system-namespaces` so we know what rancher considers a system namespace. This will also help with upkeep of the tests as the setting changes.

tests/integration/suite/test_system_project.py
```diff
@@ -71,3 +73,14 @@ def test_system_project_cant_be_deleted(admin_mc, admin_cc):
...
+    k8sclient = CoreV1Api(admin_mc.k8s_client)
+    saccounts = k8sclient.list_service_account_for_all_namespaces()
```
dramich, line 80: This func can accept a `field_selector` to filter the results to just the name required which would save the loop and name check below

tests/integration/suite/test_system_project.py
```diff
@@ -71,3 +73,14 @@ def test_system_project_cant_be_deleted(admin_mc, admin_cc):
...
+    for sa in saccounts.items:
+        if sa.metadata.name == "default":
+            sa_ns = sa.metadata.namespace
+            if sa_ns in initial_system_namespaces and sa_ns != "kube-system":
```
dramich, line 84: Should this also ignore `default`?

tests/integration/suite/test_system_project.py
```diff
@@ -71,3 +73,14 @@ def test_system_project_cant_be_deleted(admin_mc, admin_cc):
...
+                flag = sa.automount_service_account_token
+                assert flag is False
```
dramich, line 86: To ensure the test isn't marked as a failure before the controller is done, use a wait_for on the check which will give the controllers time to complete their work.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,79 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type defaultSvcAccountHandler struct {
+	serviceAccounts rv1.ServiceAccountInterface
+}
+
+func Register(ctx context.Context, cluster *config.UserContext) {
+	logrus.Debugf("Registering defaultSvcAccountHandler for checking default serviceaccount of system namespaces")
```
dramich, line 20:
```suggestion
	logrus.Debugf("Registering defaultSvcAccountHandler for checking default service account of system namespaces")
```

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,79 @@
...
+		serviceAccounts: cluster.Core.ServiceAccounts(""),
+	}
+	cluster.Core.ServiceAccounts("").AddHandler(ctx, "defaultSvcAccountHandler", nsh.Sync)
+}
+
+func (nsh *defaultSvcAccountHandler) Sync(key string, sa *corev1.ServiceAccount) (runtime.Object, error) {
+	if sa == nil || sa.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	logrus.Debugf("defaultSvcAccountHandler: Sync serviceaccount: key=%v", key)
```
dramich, line 31:
```suggestion
	logrus.Debugf("defaultSvcAccountHandler: Sync service account: key=%v", key)
```

tests/integration/suite/test_system_project.py
prachidamle: ```default``` is not a system namespace.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	//check if ns is a system-ns
+	namespace := defSvcAccnt.Namespace
+	if namespace == "kube-system" || namespace == "default" || !nsh.isSystemNS(namespace) {
+		return nil
+	}
+	if defSvcAccnt.AutomountServiceAccountToken != nil && *defSvcAccnt.AutomountServiceAccountToken == false {
+		return nil
+	}
+	defSvcAccntCopy := defSvcAccnt.DeepCopy()
```
dramich, line 53: No need to deepCopy here, we aren't working on an object from a lister.

pkg/controllers/user/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,79 @@
...
+	if systemNamespacesStr == "" {
+		return false
+	}
+
+	systemNamespaces := make(map[string]bool)
+	splitted := strings.Split(systemNamespacesStr, ",")
+	for _, s := range splitted {
+		ns := strings.TrimSpace(s)
+		systemNamespaces[ns] = true
+	}
```
dramich, line 76: This can updated to use slice.ContainsString, no need to make a new map, split the string and pass in the values.

tests/integration/suite/test_system_project.py
```diff
@@ -71,3 +74,27 @@ def test_system_project_cant_be_deleted(admin_mc, admin_cc):
...
+                if sa.automount_service_account_token is False:
+                    return True
+                else:
+                    return False
+            else:
+                if sa.automount_service_account_token is None:
+                    return True
+                else:
+                    return False
+        wait_for(_check_system_sa_flag)
```
dramich, line 100: Use a `fail_handler` to provide context on the failure, include the SA and what is wrong so if this test fails it's clear what is happening (or not happening)

tests/integration/suite/test_system_project.py
```diff
@@ -71,3 +74,27 @@ def test_system_project_cant_be_deleted(admin_mc, admin_cc):
...
+        ns = sa.metadata.namespace
+
+        def _check_system_sa_flag():
+            if ns in system_namespaces and ns != "kube-system":
+                if sa.automount_service_account_token is False:
+                    return True
+                else:
+                    return False
+            else:
+                if sa.automount_service_account_token is None:
```
dramich, line 96: This should check for both None and True or the test will fail here.

PR #27068, opened 2020-05-13
Add to lower on template name and version

brendarearden:
**Problem**
catalog was storing template version with toLower but using original name and version that was not lower case to try to retrieve the templates

**Solution**
Add toLower on template name and template version so its stored and called the same way

**Issue**
#19615


cbron:
lgtm, but let's get everyones review again, also can you update your solution in description. 

When testing we should ensure that this doesn't negatively affect previous templates/versions.

mrajashree:
Makes sense to create templates with lowercase names. But for the existing templates should we still have the `get` calls check for lowercase? because their names won't get updated?

brendarearden:
@mrajashree currently you cannot have uppercase chart names as it would not pass the subdomain regex check

Issue #25117, opened 2020-01-27
Support k8s 1.18


superseb:
```
The 1.18 release cycle is proposed as follows:

Monday, January 06: Week 1 - Release cycle begins
Tuesday, January 28: Week 4 - Enhancements Freeze
Thursday, March 05: Week 9 - Code Freeze
Monday, March 16: Week 11 - Docs must be completed and reviewed
Tuesday, March 24: Week 12 - Kubernetes v1.18.0 released
```

https://github.com/kubernetes/sig-release/tree/master/releases/release-1.18

deniseschannon:
@superseb I assigned this to you since the nodelocal dns changes will need to go along with a new k8s version.

deniseschannon:
@StrongMonkey to spend the time to add k8s 1.18 into kontainer-driver-metadata, as k8s 1.18 alpha still has the hyperkube release artifact. 

@superseb will investigate what's going to happen with the moving hyperkube out of kubernetes/kubernetes and refactor RKE/Rancher to figure out how to handle the breaking out of hyperkube into the separate images.

Oats87:
@superseb Can we not move to https://github.com/kubernetes/kubernetes/pull/84662 instead?

StrongMonkey:
Feel like there isn't too much change we need to do on our side. They will still release hyperkube images and it will just have all the binaries inside container. They are not maintaining hyperkube as a single binary but instead a shell script which just calls other binaries.

superseb:
The script based one is what we are using today as that is what went in 1.17.

The whole hyperkube discussion in the kubernetes repo is a bit confusing, as the original issue was not really the end solution. After the script solution was introduced, the step to release hyperkube builds was also added back. Also basically why we could use 1.17 as-is (besides the increase of image size because of the script based approach). The changelog for 1.17 indicates that kubeadm deprecates the use of the hyperkube image which apparently led to more distros to move off hyperkube images because "they are deprecated" while they are still being released.

The other confusion is filed as https://github.com/kubernetes/kubernetes/issues/86808, where the per component images are based on Debian Buster (with mostly new binaries that help us with some usecases, and a major drop in size) and the hyperkube ones are not:

```
$ docker run --entrypoint sh -ti gcr.io/google-containers/kube-apiserver:v1.18.0-alpha.5 -c "cat /etc/os-release"
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
[...]

PR #27070, opened 2020-05-13
Update to k8s 1.18

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/25117


pkg/controllers/management/kontainerdrivermetadata/data.go
```diff
@@ -32,7 +32,7 @@ const (
 
 const (
 	APIVersion           = "management.cattle.io/v3"
-	RancherVersionDev    = "2.4"
+	RancherVersionDev    = "2.4.3-rc100"
```
aiyengar2, line 35: Required as the minimum rancher version for k8s 1.18 is set to 2.4.3-rc99 in KDM:

https://github.com/aiyengar2/kontainer-driver-metadata/blob/afe5b43a12112a86eccee8281f89fd4d2e50ae4e/rke/k8s_version_info.go#L121

prachidamle: Then the value should be 2.4.3-rc99 I think

mrajashree: From offline discussion 2.4.3-rc100 was still showing 1.18 version, but changing ti to 2.4.3-rc99 seems correct

aiyengar2: Updated in the latest commit!

Issue #25515, opened 2020-02-21
istio destination rule cookie based hash error


stephen-dahl:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. create new destination rule.
2. set loadbalancing to cookie based hash
![image](https://user-images.githubusercontent.com/6061073/75060563-e6f59000-5493-11ea-8f29-706d90c09186.png)
4. save

**Result:**
```
Validation failed in API: DestinationRule.networking.istio.io "test" is invalid: [: Invalid value: "": "spec.trafficPolicy.loadBalancer" must validate one and only one schema (oneOf). Found none valid, : Invalid value: "": "spec.trafficPolicy.loadBalancer.consistentHash" must validate one and only one schema (oneOf). Found 2 valid alternatives]
```

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.3.5
- Installation option (single install/HA): HA

cbron:
Workaround here is to use `kubectl apply`.

cbron:
I reproduced this a few weeks back, but now I can't on either Istio 1.4.7 or 1.4.8. Tested on Rancher master branch (2783-head), so it may have already been resolved.

rancher-max:
I was able to recreate this on v2.4.3 and v2.4-head both using Istio v1.4.700 and 1.4.800 respectively, following the same steps mentioned in the issue.

cbron:
There is an issue where our API is validating openAPI schema differently than k8s.
This works when applied with kubectl: 
``` apiVersion: networking.istio.io/v1alpha3
 kind: DestinationRule
 metadata:
   name: bookinfo-ratings
 spec:
   host: ratings.prod.svc.cluster.local
   trafficPolicy:
     loadBalancer:
       consistentHash:
         httpCookie:
           name: user
           ttl: 0s
```

The equivalent does not, when going to our api: 
```
{
  "type": "destinationrule",
  "namespaceId": "default",
  "trafficPolicy": {
    "loadBalancer": {
      "consistentHash": {
        "httpCookie": {
          "ttl": "0s",
          "name": "user"
        }
      }
    }
  }
}
```

cbron:
[...]

PR #27098, opened 2020-05-15
Allow istios dest rule LB to be nullable

cbron:
Vendoring and integration test for: https://github.com/rancher/types/pull/1142 
Issue: https://github.com/rancher/rancher/issues/25515


tests/integration/suite/test_istio.py
```diff
@@ -66,6 +66,45 @@ def test_destination_rule(admin_pc):
...
+        }
+    ),
+    destinationRules = client.list_destinationRule(
+        namespaceId=ns.id
+    )
+    assert len(destinationRules) == 1
+    assert destinationRules.data[0].trafficPolicy.loadBalancer\
+        .consistentHash.httpCookie.name == cookie_name
+    client.delete(destinationRules.data[0])
+    client.delete(ns)
```
rmweir, line 105: Should use remove_resource. In current state, if something fails they won't be deleted.

tests/integration/suite/test_istio.py
```diff
@@ -66,6 +66,45 @@ def test_destination_rule(admin_pc):
     client.delete(ns)
 
 
+@pytest.mark.nonparallel
+def test_destination_rule_on_cookie(admin_pc):
```
rmweir, line 70: a small summary here to explain what this addresses or why it's being tested would be good

tests/integration/suite/test_istio.py
```diff
@@ -66,6 +66,45 @@ def test_destination_rule(admin_pc):
     client.delete(ns)
 
 
+@pytest.mark.nonparallel
```
rmweir, line 69: Do we need it to be nonparallel?

cbron: Probably not for a destination rule, for virtual services they were likely testing similar things so they could mess with each other. I'll update.

tests/integration/suite/test_istio.py
```diff
@@ -66,6 +66,48 @@ def test_destination_rule(admin_pc):
...
+            "loadBalancer": {
+                "consistentHash": {
+                    "httpCookie": {
+                        "ttl": "0s",
+                        "name": cookie_name,
+                    }
+                }
+            }
+        }
+    ),
```
rmweir, line 101: nit: remove_resource on destinationRule too, although it should be fine since ns will be deleted

cbron: fixed

tests/integration/suite/test_istio.py
```diff
@@ -66,6 +66,48 @@ def test_destination_rule(admin_pc):
     client.delete(ns)
 
 
+# consistentHash has an "anyOf" openAPI validation on it,
```
rmweir, line 69: do you mean oneOf? AnyOf would allow both to be filled in.

cbron: dang, you're right

cbron: fixed

PR #27104, opened 2020-05-15
Remove coredns from istio-images-mirror

aiyengar2:
`core-dns` needs a multi-arch build and can't be included in the istio mirror images script


aiyengar2:
https://github.com/rancher/rancher/pull/27122 will be used for coredns and similar images in the future

Issue #26999, opened 2020-05-08
CRDs not available to helm template when installing with Helm 3


brendarearden:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Add a catalog with URL https://github.com/PennyScissors/ohmyglb
Set the catalog helm version to helm 3
Attempt to install the ohmyglb app from the newly added catalog

**Result:**
```Wait helm template rendering notes failed. manifest_sorter.go:192: info: skipping unknown hook: "crd-install" Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "EtcdCluster" in version "etcd.database.coreos.com/v1beta2" : exit status 1```

**Other details that may be helpful:**
https://github.com/rancher/rancher/blob/01a8242b828d8797e33e4f2b8b097fc2e1d98812/pkg/controllers/user/helm/common.go#L156
The template command for helm 3 does not currently use the `--include-crds` flag that was added as part of https://github.com/helm/helm/pull/7138 so the template is not able to render properly which results in an error being thrown

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
- Installation option (single install/HA):
Rancher v2.4.2 single install

rancher-max:
Reopening because still seeing the same issue following the same steps on both v2.4-head and master-head

brendarearden:
https://github.com/rancher/rancher/blob/dd102af1773e574f40a5811a85151431c3e3ec4b/pkg/controllers/user/helm/common.go#L180  << line that is causing error. `--dry-run` flag does not allow crds to be included. Using `helm install --dry-run` was a work around since helm 3 does not support `helm template --notes` for rendering notes.

rancher-max:
#### Validated in v2.4-head and master-head

- This chart is able to be deployed using helm3 and the error does not occur
<img width="824" alt="image" src="https://user-images.githubusercontent.com/55997940/82934648-a6009200-9f40-11ea-81b9-01247f9b12b9.png">

- Notes are working properly when this app is deployed
- App deploys successfully using non-default options and when setting custom answers

dnoland1:
Hit this issue about 30 minutes ago and have also noticed this causes significant spikes in CPU and load on the node running Rancher:
<img width="794" alt="image" src="https://user-images.githubusercontent.com/468462/85049657-d9060200-b149-11ea-811e-7c572f49a795.png">

PR #27105, opened 2020-05-15
Refactor render notes to use get notes

brendarearden:
**Problem**
App install is failing when trying to install app that has CRDs with helm 3 as the command being used to generate the notes does not support CRDs to be rendered

**Solution**
Change notes logic to use `helm get notes` after install is completed.

**Additional Info** 
If install fails, the app revision available will have empty notes. This was not the behavior previously

**Issue**
https://github.com/rancher/rancher/issues/26999

### DIFF
--- a/pkg/controllers/user/helm/common.go
+++ b/pkg/controllers/user/helm/common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"math"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/helm/common.go
```diff
@@ -66,7 +67,16 @@ func helmInstall(tempDirs *common.HelmPath, app *v3.App) error {
 			}
 		}()
 	}
-	return common.InstallCharts(tempDirs, addr, app)
+	err := common.InstallCharts(tempDirs, addr, app)
+	if err != nil {
+		return "", err
+	}
+	// rendering notes here so tiller does not need to be started twice
+	notes, err := renderNotes(app, tempDirs, addr)
```
dramich, line 75: This can be returned then the error check isn't needed

brendarearden: now returning renderNotes and not doing error check

Issue #27028, opened 2020-05-11
Allow Github integration to capture the Github user information


maggieliu:
Related to https://github.com/rancherlabs/managed-rancher/issues/353

izaac:
@kinarashah I am getting a 500 when I click Login with Github after enabling Github auth and logging out

{"baseType":"error","code":"ServerError","message":"Server error while authenticating","status":500,"type":"error"}

![Screen Shot 2020-05-26 at 12 19 54 PM](https://user-images.githubusercontent.com/27659/82941624-84f16e80-9f4b-11ea-9f17-fe2f88705130.png)

I used one of the test accounts and I think it has 2FA enabled.

Rancher version `v2.4-head (05/26/2020)` commit aae934198

izaac:
This is in the Rancher server logs:

```
2020/05/26 19:37:47 [ERROR] Github getGithubUser: GET email url https://api.github.com/user/emails received error from github, err: request failed, got status code: 404. Response: {
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user"
}

2020/05/26 19:37:47 [ERROR] API error response 500 for POST /v3-public/githubProviders/github?action=login. Cause: request failed, got status code: 404. Response: {
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user"
}
```

mrajashree:
Login works from API, but when "Login with github" is clicked in UI, the query URL has scope:undefined, possible from [this](https://github.com/rancher/ui/pull/4006/files#diff-f3328c3858d233291fe2019e2bcc36c7R69) line. This could be causing the problem since we don't want ui to add scope param

mrajashree:
Moving to-test since the UI fix is merged

izaac:
Validated this using Github in new installs for

Rancher version `v2.4-head (05/28/2020)` 24c684289
Rancher version `master-head (05/28/2020)` 56a464049

I also smoke tested Google Auth  continued working (I was able to enable the auth and login afterwards.


Before in Rancher v2.4.3 installs the userattribute store shown none with the following `kubectl` command (for local and Github users):

```
kubectl get userattribute -o custom-columns='Name:UserName,Email:UserPrincipal.extraInfo.Email,Location:UserPrincipal.extraInfo.Location,Company:UserPrincipal.extraInfo.Company,Username:UserPrincipal.loginName'
[...]

PR #27143, opened 2020-05-18
 Store userinfo in userattribute

mrajashree:
https://github.com/rancher/rancher/issues/27028
Depends on https://github.com/rancher/types/pull/1145

This commit adds a field UserInfo on to UserAttribute CRD. The fields in
UserInfo vary per provider and can be set as a map. For github we want to save the
company, location, email and github ID.
The userInfo api endpoint of github already returns all these fields.
This commit adds 3 new fields to github Account type so we can get the
values for company,location and email from [github api response](https://developer.github.com/v3/users/#response). And saves
all these fields on the userAttribute crd object of that user

Although userinfo endpoint returns email too, it only does so for emails
set on the profile. For other users we need to request email from user/emails
endpoint. So this commit also adds the scope "user:email" for this request.
Now on UI won't add the existing "read:org" scope either, the api will add that too

### DIFF
--- a/pkg/auth/providers/github/github_account.go
+++ b/pkg/auth/providers/github/github_account.go
@@ -16,6 +16,9 @@ type Account struct {
 	AvatarURL string `json:"avatar_url,omitempty"`
 	HTMLURL   string `json:"html_url,omitempty"`
 	Type      string `json:"type,omitempty"`
+	Email     string `json:"email,omitempty"`
+	Company   string `json:"company,omitempty"`
+	Location  string `json:"location,omitempty"`
 }
 
 //Team defines properties a team on github has
@@ -24,6 +27,12 @@ type Team struct {
 	Organization map[string]interface{} `json:"organization,omitempty"`
 	Name         string                 `json:"name,omitempty"`
 	Slug         string                 `json:"slug,omitempty"`
+}
+
+// EmailResponse is used for getting response from the emails endpoint. https://developer.github.com/v3/users/emails/#response .We only need the Email and Primary fields
+type EmailResponse struct {
+	Email   string `json:"email,omitempty"`
+	Primary bool   `json:"primary,omitempty"`
 }
 
 func (t *Team) toGithubAccount(url string, account *Account) {
--- a/pkg/auth/providers/github/github_client.go
+++ b/pkg/auth/providers/github/github_client.go
@@ -76,7 +76,27 @@ func (g *GClient) getUser(githubAccessToken string, config *v3.GithubConfig) (Ac
 		logrus.Errorf("Github getGithubUser: error unmarshalling response, err: %v", err)
 		return Account{}, err
 	}
-
+	// If the UserInfo endpoint does not return the email of a user, we need to query the emails endpoint
+	if githubAcct.Email == "" {
+		emailURL := g.getURL("USER_EMAIL", config)
+		emailResp, _, err := g.getFromGithub(githubAccessToken, emailURL)
+		if err != nil {
+			logrus.Errorf("Github getGithubUser: GET email url %v received error from github, err: %v", emailURL, err)
+			return Account{}, err
+		}
+		var userEmails []EmailResponse
+		if err := json.Unmarshal(emailResp, &userEmails); err != nil {
+			logrus.Errorf("Github getGithubUser: error unmarshalling email response, err: %v", err)
+			return Account{}, err
+		}
+		// This api endpoint doesn't accept query parameter to filter on primary, so we get all emails and loop through it
+		for _, userEmail := range userEmails {
+			if userEmail.Primary {
+				githubAcct.Email = userEmail.Email
+				break
+			}
+		}
+	}
 	return githubAcct, nil
 }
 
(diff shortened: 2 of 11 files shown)


pkg/auth/providers/github/github_client.go
```diff
@@ -76,7 +76,27 @@ func (g *GClient) getUser(githubAccessToken string, config *v3.GithubConfig) (Ac
...
+			logrus.Errorf("Github getGithubUser: error unmarshalling email response, err: %v", err)
+			return Account{}, err
+		}
+		// This api endpoint doesn't accept query parameter to filter on primary, so we get all emails and loop through it
+		for _, userEmail := range userEmails {
+			if userEmail.Primary {
+				githubAcct.Email = userEmail.Email
+				break
+			}
+		}
```
aiyengar2, line 98: Seems like this might be required by the following as well: https://github.com/rancher/rancher/blob/774dc55f6e1c4dba46906fbbcf42ffaeba336509/pkg/auth/providers/github/github_client.go#L242-L258

Can we create a new function `getUserEmailID` so it can be called by both?

mrajashree: I hadn't seen that, and I'm not sure why we call userinfo endpoint for org

pkg/auth/providers/github/github_account.go
```diff
@@ -26,6 +29,12 @@ type Team struct {
 	Slug         string                 `json:"slug,omitempty"`
 }
 
+// EmailResponse is used for getting response from the emails endpoint. https://developer.github.com/v3/users/emails/#response .We only need the Email and Primary fields
+type EmailResponse struct {
+	Email   string `json:"email,omitempty"`
+	Primary bool   `json:"primary,omitempty"`
+}
```
prachidamle, line 36: Does this also list user's private email addresses?

pkg/auth/providers/github/github_client.go
```diff
@@ -76,7 +76,27 @@ func (g *GClient) getUser(githubAccessToken string, config *v3.GithubConfig) (Ac
...
+			return Account{}, err
+		}
+		var userEmails []EmailResponse
+		if err := json.Unmarshal(emailResp, &userEmails); err != nil {
+			logrus.Errorf("Github getGithubUser: error unmarshalling email response, err: %v", err)
+			return Account{}, err
+		}
+		// This api endpoint doesn't accept query parameter to filter on primary, so we get all emails and loop through it
+		for _, userEmail := range userEmails {
+			if userEmail.Primary {
```
prachidamle, line 94: Should we be also reading only Public?

pkg/auth/providers/github/github_account.go
mrajashree: yes, but only after user approves the user:email scope we request during login

pkg/auth/providers/github/github_client.go
prachidamle: This is not calling userinfo endpoint for org - It is just a confusing method name. I think is a method to get User Or Org (group) By ID

pkg/auth/providers/github/github_account.go
prachidamle: Will user be approving it on the configuration page for github? What if user does not approve it

pkg/auth/providers/github/github_client.go
prachidamle: This call should only be returning public info of the user/ since this can be called by user search

prachidamle: So if we are going to add in email; it should not be private email. Its probably better to not add email data in here.

mrajashree: But I don't think we should get email at all here, since this is not what we store on userattribute. So it's an additional api call that we can avoid

pkg/auth/providers/github/github_account.go
mrajashree: User will approve it during login. If user doesn't click "authorize" then login won't take place.  It's similar to the read:org access we already request

pkg/auth/providers/github/github_client.go
mrajashree: There can be only one primary email, and it's not necessary that's the one user marks as public. It could either be us storing the only primary email or the first public email. But I think either is fine since everything's returned by the github call response. I don't really have an opinion on which one to store, maybe @dnoland1 can tell?

pkg/auth/providers/github/github_client.go
mrajashree: Yeah better to not add email data here, since this is not what we store on userattribute. So it's an additional api call that we can avoid

pkg/auth/providers/github/github_client.go
mrajashree: It's possible user has only one email saved and that is set to "private". So if we look for only the public one, we discard all emails that github api has returned to us. Plus on login page Github shows this to the user so user's aware of giving us read-only access to private email. Based on offline discussion with @deniseschannon and @will-chan we will store the primary one whether or not it's public
new

prachidamle: great, thanks

pkg/auth/providers/github/github_account.go
prachidamle: ok

PR #27152, opened 2020-05-19
[v2.3] set kdm and system-chart branch to dev-v2.3

kinarashah:


go.mod
```diff
@@ -70,7 +70,7 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v1.4.0
 	github.com/prometheus/common v0.6.0
-	github.com/rancher/kontainer-driver-metadata v0.0.0-20200505175632-890843de6b2a
+	github.com/rancher/kontainer-driver-metadata v0.0.0-20200422181828-23522875e520
```
deniseschannon, line 73: what's this for?

deniseschannon:
@kinarashah I'm not sure why you're bumping KDM commit IDs. Ideally, you should have a PR that is just the changes for the branches so you can easily revert later to set it back to the release branches. 

If the KDM commits are just to switch from the dev branch to the release branch, they don't need to be done in this PR. fundamentally, the content is the same.

kinarashah:
@deniseschannon yep the commit IDs change was for switching the branches, wanted commit to also reflect that just as a reminder when new commits go to dev-v2.3 because that's where the airgap data comes from. 

But agreed on the revert commit part for the release, updated PR to include just the branch change.

package/Dockerfile
```diff
@@ -13,10 +13,10 @@ WORKDIR /var/lib/rancher
 
 ARG ARCH=amd64
 ARG IMAGE_REPO=rancher
-ARG SYSTEM_CHART_DEFAULT_BRANCH=release-v2.3
+ARG SYSTEM_CHART_DEFAULT_BRANCH=dev-v2.3
 
 # kontainer-driver-metadata branch to be set for specific branch other than dev/master, logic at rancher/rancher/pkg/settings/setting.go
-ARG RANCHER_METADATA_BRANCH=
+ARG RANCHER_METADATA_BRANCH=dev-v2.3
```
deniseschannon, line 19: pretty sure you don't need this but we can keep it in.

Issue #26754, opened 2020-04-22
cant edit description on clusters created with rke templates


aemneina:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. create cluster and select an existing rke template
2. after cluster is provisioned, edit cluster, modifying description
3. click save

**Result:**
the description doesnt get updated or saved. custom clusters work fine, its just when you use rke templates, the description change is omitted.

aemneina:
UI doesnt seem to preserve annotations added via the UI as well.

soumyalj:
Reproduced the issue with v2.4.3. 
1. Created a RKE template "template1"
2. Create a cluster using "template1"
3. Edit the cluster and add description, labels and annotations and save
The description/labels/annotation do not get updated.

soumyalj:
Validated the fix with master-head build (dea919fb2)

1. Create a RKE template "template1"
2. Create a cluster using "template1"
Edit the cluster and add description, labels and annotations and save
The description, labels and annotations are updated correctly.
API also reflects the description, labels, annotations. 
```
created": "2020-05-21T00:32:57Z",
"createdTS": 1590021177000,
"creatorId": "user-nd9k6",
"defaultClusterRoleForProjectMembers": null,
"defaultPodSecurityPolicyTemplateId": null,
"description": "testdesc",
"desiredAgentImage": null,
"desiredAuthImage": null,
"dockerRootDir": "/var/lib/docker",
"driver": "rancherKubernetesEngine",
"enableClusterAlerting": false,
"enableClusterMonitoring": false,
"enableNetworkPolicy": false,
"id": "c-zpq5b",
"internal": false,
"istioEnabled": false,
"labels": {
"cattle.io/creator": "norman",
"l1": "l2"
},

```
```
"annotations": {
"a1": "a2",
"authz.management.cattle.io/creator-role-bindings": "{\"created\":[\"cluster-owner\"],\"required\":[\"cluster-owner\"]}",
"lifecycle.cattle.io/create.cluster-agent-controller-cleanup": "true",
"lifecycle.cattle.io/create.cluster-provisioner-controller": "true",
"lifecycle.cattle.io/create.cluster-scoped-gc": "true",
"lifecycle.cattle.io/create.mgmt-cluster-rbac-remove": "true",
"provisioner.cattle.io/ke-driver-update": "updated"
},


```

PR #27154, opened 2020-05-19
[ForwardPort] For a cluster created from RKE template, update description, labels, annotations

prachidamle:
For a cluster created from RKE template, update description, labels and annotations separately

#26754


prachidamle:
@deniseschannon thanks, I remember that this was the process to have only committers approval on forwardport/backport; dint know we could use any approval. Thanks!

PR #27167, opened 2020-05-20
Add job to run sonobuoy

ShylajaDevadiga:


tests/validation/tests/v3_api/test_run_sonobuoy.py
```diff
@@ -0,0 +1,50 @@
...
+        f.write(config)
+
+    cmd = "sonobuoy run --mode={0} --kube-conformance-image-version={1} --kubeconfig={2}".format(RANCHER_SONOBUOY_MODE, RANCHER_K8S_VERSION, kubeconfig)
+    status = run_command(cmd)
+    time.sleep(60)
+
+    cmd = "sonobuoy status --kubeconfig={0}".format(kubeconfig)
+    status = run_command(cmd)
+    print(status)
+    while "running" in status or "Pending" in status:
```
izaac, line 27: Could we get this in a separate method in this file? like `wait_for_sonobouy_results`

tests/validation/tests/v3_api/test_run_sonobuoy.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_sonobuoy_results():
+    config = base64.b64decode(RANCHER_KUBECONFIG).decode("utf-8")
+    kubeconfig = "./kubeconfig"
+
+    with open(kubeconfig, 'w') as f:
+        f.write(config)
+
+    cmd = "sonobuoy run --mode={0} --kube-conformance-image-version={1} --kubeconfig={2}".format(RANCHER_SONOBUOY_MODE, RANCHER_K8S_VERSION, kubeconfig)
```
izaac, line 20: Each of this commands perform a different action, could we wrap them around methods so the main test looks more as reading steps?
```python
def test_sonobuoy_results():
    run_sonobuouy_conformance()
    wait_for_sonobouy_results() # <- in this method we can call get_sonobuoy_status()
    retrieve_sonobuoy_results()
    unpack_sonobuoy_results()
    show_sonobuoy_results()
    delete_sonobuoy_from_cluster()
```

tests/validation/tests/v3_api/test_run_sonobuoy.py
```diff
@@ -0,0 +1,50 @@
...
+    cmd = "tar xzf {0}".format(result)
+    status = run_command(cmd, log_out=False)
+
+    filepath = "./plugins/e2e/results/global/e2e.log"
+    is_file = os.path.isfile(filepath)
+    assert is_file
+
+    cmd = "tail -10 " + filepath
+    result = run_command(cmd)
+    print(result)
```
izaac, line 47: Assert the results after printing, we should get the pass/fail if a result is other than passed.

Issue #24996, opened 2020-01-17
Improve clusterdeploy agent deployment logic


superseb:
The loop for deploying the needed Rancher resources in a cluster is using a basic loop to try 3 times (https://github.com/rancher/rancher/blob/master/pkg/controllers/management/clusterdeploy/clusterdeploy.go#L209).

This can also cause the cluster provisioning to error out with a massive error with Forbidden responses from the cluster.

As we execute the command and get an exit code and response in return, we can probably enhance this logic to retry based on the response and not just 3 times. And if it still fails, return that as an error and not possibly just log the response.

sowmyav27:
No regressions on regular automation runs on 2.4-head, for fresh installs and upgrades.

PR #27174, opened 2020-05-20
[2.4] Extend kubectl agent deploy retries, improve log

deniseschannon:
Issue(s):
* No logging available around clusterdeploy/kubectl actions
* Randomly an error is seen during cluster provisioning when applying agent YAML

Solution:
* Add (debug) logging on the clusterdeploy/kubectl actions
* Extend retries and sleep for kubectl apply for agent YAML to give it a bit more time to be applied, also added logging (and redacted the token in the logging)


deniseschannon:
https://github.com/rancher/rancher/issues/24996

Issue #25846, opened 2020-03-06
catalog apps - duplicated revisions when deploying an app 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- deploy an app from the builtin catalog, in my case it is WordPress
- wait the app the be active, go to rollback the app

**Result:**
- there are 2 duplicated revisions on the list
<img width="977" alt="Screen Shot 2020-03-06 at 4 23 06 PM" src="https://user-images.githubusercontent.com/6218999/76130718-0d094d00-5fc9-11ea-9919-52ff0f64ca5b.png">

**Other details that may be helpful:**
- this does not happen to the multi-cluster apps 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
master-head 63306835a
- Installation option (single install/HA):
single install

jiaqiluo:
the same issue happens for the WordPress app from the builtin `library` catalog which uses helm 2.

Steps:
- go to a project
- deploy the `WordPress` app from the builtin `library` catalog 
- check revisions after the app is active

jiaqiluo:
See the same issue when using a helm 3 catalog

steps:
- go to a project
- add this catalog with helm 3 `https://argoproj.github.io/argo-helm`
![image](https://user-images.githubusercontent.com/6218999/76245971-1decd500-61fa-11ea-904c-8b2278786978.png)

- deploy the app `argo` from the catalog 
- check revisions after the app is active
![image](https://user-images.githubusercontent.com/6218999/76246027-352bc280-61fa-11ea-9611-8649fa472feb.png)

results:
- two revisions show in the list and they are identical 
![image](https://user-images.githubusercontent.com/6218999/76246050-3c52d080-61fa-11ea-8f53-dc17ed4c7a4c.png)

brendarearden:
Running from this commit  `6e723b723`  prior to helm changes, I am seeing this same behavior for helm 2 to, so it seems like something else introduced after v2.3.5 that is causing these issues unrelated to helm 3 changes.

brendarearden:
**TLDR**
Changing to an api call instead of using cache does not solve root problem of how we are checking for app obj changes. Need to resolve how changes are being detected

**Initial fix attempt**
[...]

PR #27179, opened 2020-05-21
Use AppGetter to get current obj

brendarearden:
**Problem**
Duplicate app revisions are randomly occurring 

**Solution**
AppLister seems to be returning a stale object, so changed method from lister to use AppGetter interface. This does have performance tradeoffs as it will not be using the cache. 

**Issue**
#25846

### DIFF
--- a/pkg/controllers/user/helm/controller.go
+++ b/pkg/controllers/user/helm/controller.go
@@ -57,7 +57,6 @@ func Register(ctx context.Context, user *config.UserContext, kubeConfigGetter co
 		ClusterName:           user.ClusterName,
 		AppRevisionGetter:     user.Management.Project,
 		AppGetter:             user.Management.Project,
-		AppsLister:            user.Management.Project.Apps("").Controller().Lister(),
 		NsLister:              user.Core.Namespaces("").Controller().Lister(),
 		NsClient:              user.Core.Namespaces(""),
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/helm/controller.go
```diff
@@ -116,7 +116,7 @@ func (l *Lifecycle) Updated(obj *v3.App) (runtime.Object, error) {
 	// always refresh app to avoid updating app twice
 	_, projectName := ref.Parse(obj.Spec.ProjectName)
 	var err error
```
cbron, line 118: Can you add a comment here explaining why this can't be a lister, so someone doesn't switch in the future trying to make performance improvement.

ibuildthecloud:
There's basically a solution for this general problem. If you know you locally modify a resource and then immediately act upon the result from somewhere else that uses a cache, you can setup a MutationCache.  I don't think it's straight forward to use, so maybe out of scope here, but this is something I've been thinking of integrating in as a simpler pattern to use.

My thought is that all our clients from wrangler/norman are setup to use mutation caches so that if we update an object we locally store the mutation until the cache updates.  Food for thought.

ibuildthecloud:
My comments might be out of context.

brendarearden:
This solution does not resolve root cause -  this PR will be closed and additional information has been added to the issue.

PR #27200, opened 2020-05-22
fix the failure where no route53 record is found

jiaqiluo:
fix the failure where no route53 record is found


ShylajaDevadiga:
LGTM

PR #27225, opened 2020-05-22
Ingress with no path will cause a panic

ibuildthecloud:
Ingress with no path will cause a panic

If no path is set the HTTP field is nil causing a panic. The below yaml cause Rancher to panic.

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
spec:
  rules:
  - host: hostname
```


deniseschannon:
https://github.com/rancher/rancher/issues/27230

PR #27229, opened 2020-05-23
[2.4] Ingress with no path will cause a panic

deniseschannon:
If no path is set the HTTP field is nil causing a panic. The below yaml
cause Rancher to panic.

apiVersion: extensions/v1beta1
kind: Ingress
spec:
  rules:
  - host: hostname


deniseschannon:
Backport https://github.com/rancher/rancher/pull/27225

https://github.com/rancher/rancher/issues/27230

PR #27408, opened 2020-06-03
Adding code for including k8s version while creating a windows cluster

sowmyav27:


sowmyav27:
No need for this as this already exists.

PR #27426, opened 2020-06-03
change the default EC2 instance type to t3a.medium

jiaqiluo:
change the default EC2 instance type to t3a.medium


sangeethah:
LGTM

PR #27445, opened 2020-06-04
[2.4]Correct the OwnerReferences of Role and Rolebindings to use correct resource Kind and apiVersion

prachidamle:
This PR fixes the OwnerReferences of Role and Rolebindings created for ClusterTemplate, ClusterTemplateRevision, GlobalDNS, MulticlusterApp and NodeTemplate resources.

The OwnerReferences were using wrong kind and apiVersion for these resources. 

Example for ClusterTemplate:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  creationTimestamp: "2020-06-04T21:06:26Z"
  labels:
    cattle.io/creator: norman
  name: ct-nlcgc-ct-a
  namespace: cattle-global-data
  ownerReferences:
  - apiVersion: management.cattle.io
    kind: clustertemplates
    name: ct-nlcgc
    uid: 0eb0f6ac-6e91-4a04-b9cf-8693bb702f36
  resourceVersion: "25988"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/cattle-global-data/rolebindings/ct-nlcgc-ct-a
  uid: 7f7c1263-9e01-4859-961e-099f532c0a65
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: ct-nlcgc-ct-a
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: user-gjrll
```
The ownerreference here is wrong.  The apiVersion is wrong (it is missing the version v3) and the kind is set to the resource name (clustertemplates) instead of kind (ClusterTemplate)

This caused the issue where after deleting a ClusterTemplate, the dependent ClusterTemplateRevisions were not deleted and orphaned because the Kubernetes GC was stuck in a loop trying to delete the dependent Role/Rolebinding that were using invalid OwnerReferences. So GC never finished cleaning up the bindings or the ClusterTemplateRevisions leading to the orphaned data.

https://github.com/rancher/rancher/issues/26861

### DIFF
--- a/pkg/controllers/management/rbac/rbac.go
+++ b/pkg/controllers/management/rbac/rbac.go
@@ -27,7 +27,8 @@ const (
 	ClusterTemplateRevisionResource = "clustertemplaterevisions"
 	CloudCredentialResource         = "secrets"
 	CreatorIDAnn                    = "field.cattle.io/creatorId"
-	RancherManagementAPIVersion     = "management.cattle.io"
+	RancherManagementAPIVersion     = "management.cattle.io/v3"
+	RancherManagementAPIGroup       = "management.cattle.io"
 	NodeTemplateResource            = "nodetemplates"
 )
 
@@ -37,7 +38,7 @@ var subjectWithAllUsers = k8srbacv1.Subject{
 	APIGroup: rbacv1.GroupName,
 }
 
-func CreateRoleAndRoleBinding(resource, name, namespace, apiVersion, creatorID string, apiGroup []string, UID types.UID, members []v3.Member,
+func CreateRoleAndRoleBinding(resource, kind, name, namespace, apiVersion, creatorID string, apiGroup []string, UID types.UID, members []v3.Member,
 	mgmt *config.ManagementContext) error {
 	/* Create 3 Roles containing the CRD, and the current CR in resourceNames list
 	1. Role with owner verbs (Owner access, includes creator); name multiclusterapp.Name + "-ma" / (globalDNS.Name + "-ga")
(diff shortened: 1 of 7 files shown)


pkg/controllers/management/rbac/rbac.go
```diff
@@ -160,20 +161,20 @@ func createRole(resourceType, resourceName, namespace, roleAccess, apiVersion st
...
 	roleName, _ := GetRoleNameAndVerbs(roleAccess, resourceName, resourceType)
 	// we can define the rolebinding first, since if it's not already present we can call create. And if it's present then we'll
 	// still need to compare the current members' list
 	sort.Slice(subjects, func(i, j int) bool { return subjects[i].Name < subjects[j].Name })
-	return createRoleBinding(resourceType, resourceName, namespace, roleName, apiVersion, mgmt, UID, subjects)
+	return createRoleBinding(resourceType, resourceKind, resourceName, namespace, roleName, apiVersion, mgmt, UID, subjects)
 }
 
-func createRoleBinding(resourceType, resourceName, namespace, roleName, apiVersion string, mgmt *config.ManagementContext,
+func createRoleBinding(resourceType, resourceKind, resourceName, namespace, roleName, apiVersion string, mgmt *config.ManagementContext,
```
dramich, line 173: `resourceType` is no longer used in this func, or `createRoleBindingForMembers` so it should be dropped (renamed really)

prachidamle: Still used by createRoleBindingForMembers; dropped it from createRoleBinding

Issue #27258, opened 2020-05-26
Flakey test: test_eks_cluster_immutable_subnets


cbron:
Added here: https://github.com/rancher/rancher/pull/26640/commits/83fea749f62c10b649f2d26558b972796484fc4d

Examples:

* https://drone-pr.rancher.io/rancher/rancher/9124/2/2
* https://drone-pr.rancher.io/rancher/rancher/9143/2/2
* https://drone-pr.rancher.io/rancher/rancher/9154/2/2
* https://drone-pr.rancher.io/rancher/rancher/9163/2/2

Looks like they are all happening on arm, not amd, so might be a resources/timing issue.

cbron:
https://github.com/rancher/rancher/blob/master/tests/integration/suite/test_cluster_defaults.py#L59-L81
Need to wait on line 68 until cluster is created. Potentially on line 80/98 for update to occur.

PR #27460, opened 2020-06-05
flakey test: EKS subnet test failing because cluster is slow to create on arm

cbron:
https://github.com/rancher/rancher/issues/27258


cbron:
@rmweir ran this 50 times parallel without fix and it failed 4 times, with fix it failed 0 times.

PR #27501, opened 2020-06-09
Add test for deploying airgapped rancher

rancher-max:


tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -0,0 +1,384 @@
...
+        targets.append(aws_node.provider_node_id)
+
+    # Register the nodes to the target groups
+    targets_list = [dict(Id=target_id, Port=80) for target_id in targets]
+    AmazonWebServices().register_targets(targets_list, tg80_arn)
+    AmazonWebServices().register_targets(targets_list, internal_tg80_arn)
+    targets_list = [dict(Id=target_id, Port=443) for target_id in targets]
+    AmazonWebServices().register_targets(targets_list, tg443_arn)
+    AmazonWebServices().register_targets(targets_list, internal_tg443_arn)
+    return lb_dns
```
jiaqiluo, line 377: what is the internal lb used for?

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -0,0 +1,384 @@
...
+    random_test_name, RANCHER_SERVER_VERSION, HOST_NAME, AGENT_REG_CMD
+)
+from .common import (
+    AmazonWebServices, run_command, TEST_IMAGE, TEST_IMAGE_NGINX,
+    TEST_IMAGE_OS_BASE
+)
+
+PRIVATE_REGISTRY_USERNAME = os.environ.get("RANCHER_BASTION_USERNAME")
+PRIVATE_REGISTRY_PASSWORD = os.environ.get("RANCHER_BASTION_PASSWORD")
+PROVIDER_ID = os.environ.get("RANCHER_PROVIDER_ID", "")
```
jiaqiluo, line 14: In the case where I want to get both the bastion server and a rancher server in one job, should this value be set by the job itself?

tests/validation/lib/aws.py
```diff
@@ -144,19 +144,24 @@ def create_node(self, node_name, ami=AWS_AMI, ssh_user=AWS_USER,
 
         if wait_for_ready:
             node = self.wait_for_node_state(node)
-            node.ready_node()
+            if public_ip:
+                node.ready_node()
+            else:
+                time.sleep(60)
```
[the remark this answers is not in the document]
rancher-max: Sleep in both places is because there is no ssh access, so no way to really validate `ready_node()` other than wait for a bit. I noticed that without the sleep there was some flakiness in the initial connections to the nodes. It worked most of the time at 30 seconds but I saw 2 failures in my tests when only having the sleep for 30 seconds so did it for 60 just to be safe.

tests/validation/tests/v3_api/test_airgap.py
rancher-max: This is used in conjunction with the Route53 CNAME to set for the Rancher Server URL so that adding clusters will work properly.

The other LB is used to access the URL without setting SOCKS proxy or anything.

tests/validation/tests/v3_api/test_airgap.py
rancher-max: `PRIVATE_REGISTRY_USERNAME` and `PRIVATE_REGISTRY_PASSWORD` are set by the job, and `PROVIDER_ID` isn't necessary in this case as the test to create both bastion server and rancher already knows about the bastion server.

jiaqiluo: @rancher-max  thanks for the explanation. 

Why it is called `PROVIDER_ID`? Could we make the name more self-explaining?

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -0,0 +1,455 @@
...
+NUMBER_OF_INSTANCES = int(os.environ.get("RANCHER_AIRGAP_INSTANCE_COUNT", "1"))
+IMAGE_LIST = os.environ.get("RANCHER_IMAGE_LIST", ",".join(
+    [TEST_IMAGE, TEST_IMAGE_NGINX, TEST_IMAGE_OS_BASE])).split(",")
+
+RANCHER_AG_INTERNAL_HOSTNAME = HOST_NAME + "-internal.qa.rancher.space"
```
jiaqiluo, line 24: Can you add the deletion of the route53 record for `RANCHER_AG_INTERNAL_HOSTNAME ` here: 
https://github.com/rancher/rancher/blob/0d7c892c00fdd62e0d0673ef98c98bd802bb7a4e/tests/validation/tests/v3_api/common.py#L2648-L2651

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -0,0 +1,383 @@
...
+
+    # Copy SSH Key to Bastion and local dir and give it proper permissions
+    write_key_command = "cat <<EOT >> {}.pem\n{}\nEOT".format(
+        bastion_node.ssh_key_name,  bastion_node.ssh_key)
+    bastion_node.execute_command(write_key_command)
+    local_write_key_command = \
+        "mkdir -p {} && cat <<EOT >> {}/{}.pem\n{}\nEOT".format(
+            SSH_KEY_DIR, SSH_KEY_DIR,
+            bastion_node.ssh_key_name, bastion_node.ssh_key)
+    run_command(local_write_key_command, log_out=False)
```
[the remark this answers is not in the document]
rancher-max: I feel like the writing files commands didn't need to be shown in the logs. Mostly just for log cleanliness!

PR #27503, opened 2020-06-09
Automation s3 etcd recover test

izaac:
This tests create 1 cluster with s3 backups.

- 1 ControlPlane/Worker Node
- 3 etcd nodes

Then it snapshots the cluster.
Stops the etcd nodes in ec2
Waits for the cluster to go into unavailable state
Removes all 3 etcd nodes
Creates 3 new etcd nodes and waits until the cluster asks to restore from backup.
Restores from backup
Cluster is validated after it gets in Active state.
Removes cluster


tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
```diff
@@ -1076,7 +1076,8 @@ def register_host_after_delay(client, cluster, node_role, delay):
 
 
 def create_and_validate_custom_host(node_roles, random_cluster_name=False,
-                                    validate=True, version=K8S_VERSION):
+                                    validate=True, version=K8S_VERSION,
+                                    rke_config=rke_config):
```
soumyalj, line 1080: We are including rke_config in the list of variables for the method, how did it work before? 
Are there other calls to this method (in other files) which will need modification ?

tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
```diff
@@ -1095,7 +1097,7 @@ def create_and_validate_custom_host(node_roles, random_cluster_name=False,
 
 def create_custom_host_from_nodes(nodes, node_roles,
                                   random_cluster_name=False, windows=False,
-                                  version=K8S_VERSION):
+                                  version=K8S_VERSION, rke_config=rke_config):
```
soumyalj, line 1100: We are including rke_config in the list of variables for the method, how did it work before?
Are there other calls to this method which will need modification ?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,158 @@
+import pytest
+from .common import *  # NOQA
+from .test_rke_cluster_provisioning import rke_config, \
+    create_and_validate_custom_host, node_template_ec2
+
+namespace = {"p_client": None, "ns": None, "cluster": None, "project": None,
+             "nodes": []}
+backup_info = {"backupname": None, "backup_id": None, "workload": None,
+               "backupfilename": None, "etcdbackupdata": None}
+HOST_NAME = os.environ.get('RANCHER_HOST_NAME', "testcustom")
```
sowmyav27, line 10: we can import this from `test_rke_cluster_provisioning`

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,158 @@
...
+
+rke_config["services"]["etcd"]["backupConfig"] = {
```
sowmyav27, line 12: Maybe we can put it in a common place and import than instead ? I see this used in other files like `test_bkp_restore_s3_with_creds.py`, etc

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,158 @@
...
+    cluster = namespace["cluster"]
+    client = namespace["client"]
+    ns, b_info = validate_backup_create(namespace, backup_info, "s3")
+    etcd_nodes = get_etcd_nodes(client, cluster)
+    assert len(etcd_nodes) > 0
+    ips_to_remove = []
+    for etcd_node in etcd_nodes:
+        ips_to_remove.append(etcd_node.externalIpAddress)
+        client.delete(etcd_node)
+    cluster = wait_until_cluster_is_updating(client, cluster)
```
sowmyav27, line 42: Why do we wait until the state is `Updating` here ?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,158 @@
...
+    for ip_to_remove in ips_to_remove:
+        delete_node_from_ec2(ip_to_remove)
+    ips_to_remove.clear()
+    cluster = add_new_etcd_nodes(client, cluster)
+    cluster = client.reload(cluster)
+    message = "Please restore your cluster from backup"
```
sowmyav27, line 56: Do we always expect to get this message ?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,158 @@
...
+    request.addfinalizer(fin)
+
+
+def wait_for_cluster_transitioning_message(client, cluster, message):
+    start = time.time()
+    while message not in cluster.transitioningMessage:
+        print(cluster.transitioningMessage)
+        time.sleep(5)
+        cluster = client.reload(cluster)
+        if time.time() - start > MACHINE_TIMEOUT:
```
sowmyav27, line 109: In Worst case scenario, we wait for `MACHINE_TIMEOUT` which is 1200 seconds here? Is that needed? How long does it usually take for the message to appear?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+        {'Name': 'ip-address',
+         'Values': [address]}
+    ]
+    aws_node = AmazonWebServices().get_nodes(filters)
+    AmazonWebServices().delete_node(aws_node[0])
+
+
+def restore_backup_to_recover(client, cluster, b_info):
+    cluster.restoreFromEtcdBackup(etcdBackupId=b_info["backup_id"])
+    validate_cluster(client, cluster, intermediate_state="updating",
```
sowmyav27, line 157: This returns a cluster. which can be return back to the original function ?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+
+    namespace["p_client"] = p_client
+    namespace["ns"] = ns
+    namespace["cluster"] = cluster
+    namespace["project"] = p
+    namespace["c_client"] = c_client
+    namespace["client"] = client
+
+    def fin():
+        cluster_cleanup(client, cluster)
```
sowmyav27, line 99: We should pass the `aws_nodes` as well here for them to be cleaned up ?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+    cluster = add_new_etcd_nodes(client, cluster)
+    cluster = client.reload(cluster)
+    # This message is expected to appear after we add new etcd nodes
+    # The cluster will require the user to perform a backup to recover
+    # this is appears in the cluster object in cluster.transitioningMessage
+    message = "Please restore your cluster from backup"
+    cluster = wait_for_cluster_transitioning_message(client, cluster, message)
+    etcd_nodes = get_etcd_nodes(client, cluster)
+    assert len(etcd_nodes) == 3, "Make sure the cluster now has 3 etcd nodes"
+    restore_backup_to_recover(client, cluster, b_info)
```
sowmyav27, line 60: `restore_backup_to_recover` should return cluster object ?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+    etcd_nodes = get_etcd_nodes(client, cluster)
+    master_node = get_worker_nodes(client, cluster)
```
sowmyav27, line 62: Small thing, but can we name it worker_node instead?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+    # validate the ingress that was created in the first backup
+    # after restoring and recovering the cluster
+    validate_ingress(namespace["p_client"], cluster,
+                     [b_info["workload"]], ns["host"],
+                     "/name.html")
+
+    # Cleanup task
+    if RANCHER_CLEANUP_CLUSTER:
```
sowmyav27, line 70: Can we have this in the finalizer in here, as this will not get executed if any of the above assert statements fail
```
@pytest.fixture(scope='module', autouse=True)
def create_project_client_and_cluster_s3_three_etcd(request):
```

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: `with_creds` and `with_iam` use different configs, I'm using the same as `with_creds` but this could allow us to be more flexible with buckets configurations if needed.

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: Yes

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: This is more complex than that, as we're removing nodes on demand not after all the tests are passed, but I will think on a better way to track the nodes.

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: I will use `DEFAULT_TIMEOUT`

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: Not of much use at that stage of the scenario but I will store the returned value.

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: Not of much use at that stage of the scenario but I will store the returned value.

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: Yes

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: Yes It's a way more complex than the cleanup of the other scenarios I've done but I will keep track of the on demand node deletes.

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: I was just making sure the cluster didn't go into an Error state or anything but I can assume it goes immediately to `Updating`

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
izaac: Yes

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,156 @@
...
+    request.addfinalizer(fin)
+
+
+def wait_for_cluster_transitioning_message(client, cluster, message):
+    start = time.time()
+    while message not in cluster.transitioningMessage:
+        print(cluster.transitioningMessage)
+        time.sleep(5)
+        cluster = client.reload(cluster)
+        if time.time() - start > DEFAULT_TIMEOUT:
```
soumyalj, line 105: Is DEFAULT_TIMEOUT(2 mins) always sufficient? In my testing I have seen it take longer than that.

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+    def fin():
+        cluster_cleanup(client, cluster, namespace["nodes"])
+    request.addfinalizer(fin)
+
+
+def wait_for_cluster_unavailable(client, cluster):
+    return wait_for_condition(client, cluster,
+                              lambda x: x.state == "unavailable",
+                              lambda x: 'State is: ' + x.state,
+                              timeout=240)
```
sowmyav27, line 102: Why a timeout of 240 seconds? Why not the DEFAULT_TIMEOUT of 120 seconds?

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -0,0 +1,169 @@
...
+
+def wait_for_cluster_transitioning_message(client, cluster, message):
+    start = time.time()
+    while message not in cluster.transitioningMessage:
+        print(cluster.transitioningMessage)
+        time.sleep(5)
+        cluster = client.reload(cluster)
+        # We are waiting 4 minutes for the transitioning message to appear
+        # this could be impacted by environmental factors
+        if time.time() - start > 240:
```
sowmyav27, line 113: Why a timeout of 240 seconds? Why not the DEFAULT_TIMEOUT of 120 seconds?

Issue #27403, opened 2020-06-02
Pulling/Pushing Images for Rancher


deniseschannon:
Within Rancher, images are maintained in Rancher's Dockerhub to support users without access to other registry domains as well as the air gap installation. 

The original image or image source can be either a direct image or a manifest (to support multiple architectures while using the same image name). 

Requirements: 
1. Organize a new repo that will be responsible for pulling/pushing images from the image source into Rancher's Dockerhub repo. 
2. Support being able to pull/push images and manifests 
3. Automate so that the pull/push is automated with Drone.

Current WIP parts:

## Image Pulling/Pushing in Rancher/Rancher

In rancher/rancher, pulling/pushing images (not manifests) exist today with the `istio-images-mirror` script. 
[Isito-Images-Mirror File](https://github.com/rancher/rancher/blob/master/scripts/istio-images-mirror) - This file lists source image, destination image and image tag. 
When changes are made to said file, the [drone pipeline](https://github.com/rancher/rancher/blob/master/.drone.yml#L80) uses that in conjunction with the [docker-pull-pusher.sh](https://github.com/rancher/rancher/blob/master/scripts/docker-pull-pusher.sh) script to upload images into Rancher's DockerHub.

> Note: Before any new repo can support this, the ci-publisher needs to be added to Rancher's Dockerhub for that specific image. 

## Manifest Pulling/Pushing

There are several versions of manifest pulling scripts that a couple of people who have run locally, but these need to be combined to create something similar to the image pulling/pushing.

deniseschannon:
STage 1:
- [ ] Pull out image pulling/pushing from rancher/rancher into a new repo

Stage 2:
- [ ] Figure out how to add multi-arch support into the new repo

PR #27512, opened 2020-06-09
[2.3] Remove image generation step from drone pipeline

yogeshmundada:
GH id: https://github.com/rancher/rancher/issues/27403

Cherry picking commit 4af6ca6 from master

Removes the image-mirroring step in drone pipeline that consumes resources


deniseschannon:
Instead of prefixing with `[backward port]` or `[forwardport]`, pelase prefix with the Rancher release associated with it. I've fixed it with this PR.

Issue #26226, opened 2020-03-24
Cannot use Aliyun virtual kubelet feature


chenchi2038:
* Follow Aliyun virtual kubelet docs, the node doesn't show up correctly in Rancher UI.
* Users can only use `nodeName` to schedule pod to the virtual kubelet . `nodeSelector` doesn't work for virtual kubelet .
*  After specifying the `nodeName` via YAML, users can not update it via Rancher UI as  `nodeName` will be removed if we update it via workload edit page

Aliyun virtual kubelet docs:

https://help.aliyun.com/document_detail/118970.html

---
| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.3.5` UI: `v2.3.36` |
|Route|`containers.index`|

loganhz:
If you would like to ask questions in Chinese, please add `rancher2` to join 
Rancher WeChat group. Thanks

maggieliu:
@guangbochen Please update the PRs since "Types" is got merged into Rancher in master

GGGitBoy:
@maggieliu PR has been updated

guangbochen:
link to PR https://github.com/rancher/rancher/pull/27523

DoNotSayYes:
**Version:** `rancher/rancher: master-head` 
**Commit:** [3ff065f](https://github.com/rancher/rancher/commit/3ff065fb6d1c448a5bd9343460d80a44e7d6cfaf)

**Steps to verify:**
1. Add an Aliyun ack cluster on Rancher UI
2. After the cluster is ready, add a virtual node on the Aliyun dashboard
3. Check on Rancher UI

**Result:**
The virtual node can be displayed correctly and users can schedule pods to it
![image](https://user-images.githubusercontent.com/57614736/93187589-881a9f80-f772-11ea-9cc5-cb5e485e4a9f.png)

PR #27523, opened 2020-06-10
Fix node.status.capacity.cpu conversion issue

GGGitBoy:
**Problem:**

Aliyun virtual kubelet node cannot be displayed in the Rancher UI because the node.status.capacity.cpu of the virtual kubelet is `1M`, which is not a number.

`1000m`, `1`, `100`, `1M` are all valid values for node.status.capacity.cpu field in Kubernetes.

However, in Rancher types, the filed is int, and we will not convert it to number if they are not int in types.

**Solution:**

So we need to handle other cases

**Related Issue:**

https://github.com/rancher/rancher/issues/26226

### DIFF
--- a/pkg/schemas/mapper/os_info.go
+++ b/pkg/schemas/mapper/os_info.go
@@ -6,6 +6,7 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 type OSInfo struct {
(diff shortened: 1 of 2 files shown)


pkg/schemas/mapper/os_info.go
```diff
@@ -15,16 +16,16 @@ func (o OSInfo) FromInternal(data map[string]interface{}) {
 	if data == nil {
 		return
 	}
-	cpuInfo := map[string]interface{}{
-		"count": values.GetValueN(data, "capacity", "cpu"),
+	cpuInfo := map[string]interface{}{}
+	cpuNum, err := resource.ParseQuantity(convert.ToString(values.GetValueN(data, "capacity", "cpu")))
+	if err == nil {
+		cpuInfo["count"] = cpuNum.Value()
 	}
```
gitlawr, line 23: Can we fall back on the original handling when the parse failed?

gitlawr (commented):
Can you add some unit tests?

pkg/schemas/mapper/os_info.go
GGGitBoy: Use `resource.ParseQuantity` can cover the original handling. 

before, we can handle values
memory: `Ki`
Cpu:  `“”`

Now we can handle all valid values in Kubernetes
memory: `Ki`, `Mi`, `Gi`, `Ti`, `Pi`, `Ei`
Cpu: `n`, `u`, `m`, `“”`, `k`, `M`, `G`, `T`, `P`, `E`

GGGitBoy:
Update, add the unit tests

gitlawr (approved):
LGTM

dramich (approved):

PR #27572, opened 2020-06-12
[2.3] Increase rancher helm/tiller version

brendarearden:
Increase rancher helm/tiller version


brendarearden:
no longer needs back ported

Issue #27297, opened 2020-05-28
Rancher helm chart does not create self-signed certificate


jmgilman:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
* Install k3s to a 3-node cluster
* Deploy cert-manager helm chart
* Deploy Rancher helm chart with the default values
* Attempt to connect to Rancher and get invalid certificate errors

Note that in the above case I was following the respective guides in the official Rancher docs.

**Result:**
Traefik is returning the default certificate (*.example.com) which, as expected, causes Chrome to complain. 

**Other details that may be helpful:**
```
$> kubectl logs traefik-758cd5fc85-4rk64 -n kube-system
{"level":"error","msg":"Error configuring TLS for ingress rancher/rancher: secret rancher/tls-rancher-ingress contains the following empty TLS data entries: tls.crt","time":"2020-05-28T04:35:31Z"}
{"level":"error","msg":"Error configuring TLS for ingress rancher/rancher: secret rancher/tls-rancher-ingress contains the following empty TLS data entries: tls.crt","time":"2020-05-28T04:35:33Z"}
{"level":"error","msg":"Error configuring TLS for ingress rancher/rancher: secret rancher/tls-rancher-ingress contains the following empty TLS data entries: tls.crt","time":"2020-05-28T04:35:35Z"}
{"level":"error","msg":"Error configuring TLS for ingress rancher/rancher: secret rancher/tls-rancher-ingress contains the following empty TLS data entries: tls.crt","time":"2020-05-28T04:35:37Z"}
```

```
$> kubectl logs -l app=rancher -n rancher
2020/05/28 04:33:52 [INFO] Adding peer wss://10.42.2.9/v3/connect, 10.42.2.9
2020/05/28 04:33:52 [ERROR] Failed to connect to peer wss://10.42.2.9/v3/connect [local ID=10.42.1.9]: websocket: bad handshake
2020/05/28 04:33:57 [ERROR] Failed to connect to peer wss://10.42.2.9/v3/connect [local ID=10.42.1.9]: websocket: bad handshake
2020/05/28 04:34:00 [INFO] Adding peer wss://10.42.0.12/v3/connect, 10.42.0.12
2020/05/28 04:34:00 [ERROR] Failed to connect to peer wss://10.42.0.12/v3/connect [local ID=10.42.1.9]: websocket: bad handshake
[...]

Issue #27160, opened 2020-05-20
Log "error running the jail command: exit status 2" seen when using CRI-O with default permissions


kubealex:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Fresh k8s installation
Follow the steps on the setup documentation:
- Install cert-manager via helm chart
- Install rancher via helm chart 
- Use 'rancher' cert creation with cert-manager

**Result:**

tls-rancher secret not created.

Events in the cattle-system namespace:

99s         Warning   ErrGetKeyPair       issuer/rancher                                      Error getting keypair for CA issuer: secret "tls-rancher" not found
99s         Warning   ErrInitIssuer       issuer/rancher                                      Error initializing issuer: secret "tls-rancher" not found

Logs in rancher pods:

2020/05/20 08:38:26 [INFO] Rancher version v2.4.3 (684884c00) is starting
2020/05/20 08:38:26 [INFO] Rancher arguments {ACMEDomains:[] AddLocal:auto Embedded:false HTTPListenPort:80 HTTPSListenPort:443 K8sMode:auto Debug:false Trace:false NoCACerts:false AuditLogPath:/var/log/auditlog/rancher-api-audit.log AuditLogMaxage:10 AuditLogMaxsize:100 AuditLogMaxbackup:10 AuditLevel:0 Features:}
2020/05/20 08:38:26 [INFO] Listening on /tmp/log.sock
2020/05/20 08:38:26 [INFO] Starting API controllers
2020/05/20 08:38:26 [INFO] Running in clustered mode with ID 10.85.0.52, monitoring endpoint cattle-system/rancher
2020/05/20 08:38:27 [INFO] Starting API controllers
2020/05/20 08:38:27 [INFO] Starting rbac.authorization.k8s.io/v1, Kind=Role controller
2020/05/20 08:38:27 [INFO] Starting rbac.authorization.k8s.io/v1, Kind=ClusterRole controller
2020/05/20 08:38:28 [FATAL] error running the jail command: exit status 2

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): stable 
- Installation option (single install/HA): single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```

[...]

PR #27582, opened 2020-06-14
Make rancher compatible with cri-o

kubealex:
Adding MKNOD capability has no effect in docker and containerd where it is enabled by default, but makes rancher compatible with cri-o.

Solves:
https://github.com/rancher/rancher/issues/27297
https://github.com/rancher/rancher/issues/27160

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -55,6 +55,9 @@ spec:
       - image: {{ .Values.rancherImage }}:{{ default .Chart.AppVersion .Values.rancherImageTag }}
         imagePullPolicy: {{ default "IfNotPresent" .Values.rancherImagePullPolicy }}
         name: {{ template "rancher.name" . }}
+        capabilities:
+          add:
+            - MKNOD
         ports:
         - containerPort: 80
           protocol: TCP


ibuildthecloud:
This honestly seems like a bug in cri-o.  Is there an issue in cri-o to reference? I don't have any issue with merging this, but I'd really like to understand why there is a different between docker/containerd and cri-o.

ibuildthecloud:
@kubealex Do you have any idea of what Rancher is doing that requires mknod? The only thing I can think of is https://github.com/rancher/rancher/blob/master/package/jailer.sh which doesn't seem related to the issues referenced.

kubealex:
> @kubealex Do you have any idea of what Rancher is doing that requires mknod? The only

The jailer is the problem, when it comes to creating the jail, this is the output:

+ cd /dev
+ tar cf - --exclude=mqueue --exclude=shm --exclude=pts .
+ cd /opt/jail/driver-jail/dev
+ tar xfp -
tar: ./urandom: Cannot mknod: Operation not permitted
tar: ./zero: Cannot mknod: Operation not permitted
tar: ./tty: Cannot mknod: Operation not permitted
tar: ./full: Cannot mknod: Operation not permitted
tar: ./random: Cannot mknod: Operation not permitted
tar: ./null: Cannot mknod: Operation not permitted
tar: Exiting with failure status due to previous errors

markc43:
I think it's good to explicitly declare any capabilities you're using that may be uncommon, even if they are the defaults for most container runtimes as they too could change in the future.

kubealex:
So?

kubealex:
@ibuildthecloud  any news about this?

rwdj:
Are you sure that's supposed to be 
```
        capabilities:
          add:
          - MKNOD 
```
And not
```
         securityContext:
          capabilities:
            add:
            - MKNOD 
```

krouma:
> Are you sure that's supposed to be
> 
> ```
>         capabilities:
>           add:
>           - MKNOD 
> ```
> 
> And not
> 
> ```
>          securityContext:
>           capabilities:
>             add:
>             - MKNOD 
> ```

I agree, it should be in `securityContext`.

kubealex:
Closing this ;)

PR #27585, opened 2020-06-15
[Master] fix fluentd mounted docker root isn't updated

aiwantaozi:
Problem:

When user updated docker root in cluster page, fluentd mounted docker root isn't updated. This bug is introduced in 2.3.
We need to update rancher logging's app config in this case.

Solution:
Update rancher logging app's config

Issue:


izaac:
Reproduced this issue in Rancher `v2.4.5` dockerroot volume doesn't get updated if we change the  Docker Root cluster value.

I checked this in an HA setup

Check workload `rancher-logging-fluentd-linux` in `System` project after provisioning/edit a cluster. 
App sets the answer `fluentd.cluster.dockerRoot` correctly in a custom cluster.
The app in the local HA cluster doesn't update the `fluentd.cluster.dockerRoot` answer in the `rancher-logging` App in my testing is that expected @aiwantaozi ?

Issue #23832, opened 2019-11-01
rancher app install from local directory fails with 404 Not Found


rextca:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Create a helm chart in directory mychart.
cd mychart
login to rancher server: rancher login -t token-g4htb:glfbmnb48g7q2dblln645etc https://ustr-rancher.myco.com
rancher --debug app install . my-name

**Result:**
DEBU[0000] Namespace create wait - Name: mychart-u9nhi, State: activating, Transitioning: yes
DEBU[0001] Namespace create wait - Name: mychart-u9nhi, State: activating, Transitioning: yes
FATA[0001] Bad response statusCode [404]. Status [404 Not Found]. Body: [baseType=error, code=NotFound, message=failed to find resource by id] from [https://ustr-rancher.myco.com/v3/project/c-trxcm:p-2ssw8/apps]

**Other details that may be helpful:**
This also fails with Rancher 2.3.2 and CLI 2.0.6.  It worked on Rancher 2.3.0 with CLI 2.0.6.

I tried various paths to the chart, like ./mychart while in the parent of mychart and fully specifying the path with no success.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.3.2
Rancher CLI version: v2.3.1
- Installation option (single install/HA):
Single

- Machine type (cloud/VM/metal) and specifications (CPU/memory): docker container
- Kubernetes version (use `kubectl version`):
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-17T17:16:09Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}

- Docker version (use `docker version`):

Client:
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        6247962
 Built:             Tue Feb 26 23:56:24 2019
[...]

PR #27591, opened 2020-06-15
Do not check user access for user's locally stored charts

brendarearden:
**Problem**
When using the cli to install an app from a local chart (ex:`rancher app install . my-name`) an error is returned from the [ProjectClient.App.Create(app)](https://github.com/rancher/cli/blob/v2.4.3/cmd/app.go#L721) method stating "failed to find resource by id". Since the externalID is empty, any checks that require data from the externalID is failing on ByID checks. externalID is added to a template when the catalog it belongs to is being traversed to make those templates available [here](https://github.com/rancher/rancher/blob/b9812df065178f31eb9c5b17af79bc3106f7b8da/pkg/catalog/manager/traverse.go#L159). There is no catalog information available when installing an app from a local chart through the cli since it is a standalone chart. 

**Solution**
return nil when externalID is empty prior to ByID checks

**Issue**
#23832

### DIFF
--- a/tests/integration/suite/test_app.py
+++ b/tests/integration/suite/test_app.py
@@ -760,6 +760,28 @@ def test_app_externalid_target_project_verification(admin_mc,
     p1_client.update(app, update_data)
 
 
+def test_local_app_can_deploy(admin_pc, admin_mc, remove_resource):
+    """Test that an app without an externalId can be deployed
+    successfully to simulate a local app deployed through cli"""
+    app_client = admin_pc.client
+    app_name = random_str()
+    ns = admin_pc.cluster.client.create_namespace(name=random_str(),
+                                                  projectId=admin_pc.
+                                                  project.id)
+    remove_resource(ns)
+
+    # create app without the externalId value set
+    app = app_client.create_app(
+        name=app_name,
+        targetNamespace=ns.name,
+        projectId=admin_pc.project.id,
+    )
+    remove_resource(app)
+    wait_for(lambda: app_client.by_id_app(app.id) is not None,
+             fail_handler=lambda:
+             "app could not be found")
+
+
 def wait_for_workload(client, ns, timeout=60, count=0):
     start = time.time()
     interval = 0.5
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_app.py
```diff
@@ -760,6 +760,26 @@ def test_app_externalid_target_project_verification(admin_mc,
...
+    remove_resource(ns)
+    # create app without the externalId value set
+    app = app_client.create_app(
+        name=app_name,
+        targetNamespace=ns.name,
+        projectId=admin_pc.project.id,
+    )
+    remove_resource(app)
+    # check that app is listed as an available app
+    assert app_client.list_app(id=app.id)
```
rmweir, line 780: The only other places I see list used this way, it uses name instead of id. I think even if it doesn't exist, a list will not fail, you would need to assert the length or maybe that it isn't none. It may be easier to do a call to by_id_app.

tests/integration/suite/test_app.py
```diff
@@ -760,6 +760,27 @@ def test_app_externalid_target_project_verification(admin_mc,
...
+    remove_resource(ns)
+
+    # create app without the externalId value set
+    app = app_client.create_app(
+        name=app_name,
+        targetNamespace=ns.name,
+        projectId=admin_pc.project.id,
+    )
+    remove_resource(app)
+    wait_until(lambda: len(app_client.by_id_app(app.id)) > 0)
```
dramich, line 780: Use `wait_for` and include a `fail_handler` so it's clear what went wrong. A length check would be good for a list, but this is getting an object so the check should use `is not None`

tests/integration/suite/test_app.py
```diff
@@ -760,6 +760,27 @@ def test_app_externalid_target_project_verification(admin_mc,
...
+    assert len(app_client.by_id_app(app.id)) > 0
```
dramich, line 781: The assert isn't needed unless it's going to check something specific, it will always have the same result as the wait_for.

PR #27596, opened 2020-06-15
Modify setup method to enable monitoring with storage class set to longhorn

soumyalj:
**Steps:**
In the setup method, 
1. Install Longhorn app 
2. Use longhorn storage class and enable monitoring 

All monitoring tests will be run with this initial setup.

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -1805,15 +1805,20 @@ def validate_catalog_app(proj_client, app, external_id, answer=None):
     chart = chart_prefix + "-" + chart_suffix
     app_name = parameters[len(parameters) - 2].split("=")[1]
     workloads = proj_client.list_workload(namespaceId=ns).data
+
+    # For longhorn app, only active state of workloads is verified as longhorn
+    # workloads do not have the field workloadLabels
+    # For all other apps active state of workloads & chart version are verified
+
     for wl in workloads:
         print("Workload {} , state - {}".format(wl.id, wl.state))
         assert wl.state == "active"
-        chart_deployed = get_chart_info(wl.workloadLabels)
-        print("Chart detail of app - {}".format(chart_deployed))
-        # '-' check is to make sure chart has both app name and version.
-        if app_name in chart_deployed and '-' in chart_deployed:
-            assert chart_deployed == chart, "the chart version is wrong"
-
+        if "longhorn" not in app.externalId:
+            chart_deployed = get_chart_info(wl.workloadLabels)
+            print("Chart detail of app - {}".format(chart_deployed))
+            # '-' check is to make sure chart has both app name and version.
+            if app_name in chart_deployed and '-' in chart_deployed:
+                assert chart_deployed == chart, "the chart version is wrong"
     # Validate_app_answers
     assert len(answers.items() - app["answers"].items()) == 0, \
         "Answers are not same as the original catalog answers"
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_monitoring.py
```diff
@@ -459,28 +465,49 @@ def create_project_client(request):
...
     if cluster["enableClusterMonitoring"] is False:
         rancher_client.action(cluster, "enableMonitoring",
                               answers=C_MONITORING_ANSWERS,
                               version=MONITORING_VERSION)
     validate_cluster_monitoring_apps()
 
-    # wait 2 minute for all graphs to be available
-    time.sleep(60 * 2)
+    # Wait 3 minutes for all graphs to be available
+    time.sleep(60 * 3)
```
izaac, line 506: Could we poll a url or api endpoint in the deployed monitoring app (prometheus?) to make sure the graph is up instead of waiting a fixed amount of time?

tests/validation/tests/v3_api/test_monitoring.py
```diff
@@ -86,7 +87,9 @@
     "workload": workload_graph_list,
     "node": node_graph_list,
 }
+STORAGE_CLASS = "longhorn"
```
izaac, line 90: Not too familiar with storage/volumes but shouldn't we also make this `STORAGE_CLASS` available through environment variables?

tests/validation/tests/v3_api/test_monitoring.py
izaac: I'm ok if we need to give Rancher more time for collecting data from the monitoring service and start populating the graphs.

soumyalj: The wait is to ensure there is enough data for graphs to be available.
The endpoint "URL/v3/clusters/c-cj7ng/clustermonitorgraphs.data" provides info only on the graphs but not if there is enough data to plot a graph.

tests/validation/tests/v3_api/test_monitoring.py
soumyalj: At this point we are testing only storage class set to longhorn.

izaac: Sounds good then 👍

tests/validation/tests/v3_api/common.py
```diff
@@ -1805,15 +1805,20 @@ def validate_catalog_app(proj_client, app, external_id, answer=None):
     chart = chart_prefix + "-" + chart_suffix
     app_name = parameters[len(parameters) - 2].split("=")[1]
     workloads = proj_client.list_workload(namespaceId=ns).data
+    '''
```
jiaqiluo, line 1808: can you switch to use # for inline comments?

tests/validation/tests/v3_api/test_monitoring.py
```diff
@@ -450,7 +454,9 @@ def test_rbac_cluster_owner_control_cluster_monitoring():
 
 
 @pytest.fixture(scope="module", autouse="True")
-def create_project_client(request):
+def setup_monitoring(request):
+    # Initialize projects, clients, install longhorn app and enable monitoring
+    # with persistence storageClass set to longhorn
```
jiaqiluo, line 459: Can you use """ for method description?

tests/validation/tests/v3_api/common.py
soumyalj: Done

tests/validation/tests/v3_api/test_monitoring.py
soumyalj: Done

Issue #26920, opened 2020-05-05
Incorrect helm version on App 


batiati:
**Problem**: A Helm 3 app is installed using Helm 2 + tiller
**Rancher: v2.4.2**
Cause: TemplateVersion objects missing it's helm version. 
Just for templateVersions added after the catalog creation, all templateVersions existing on creation have the correct version;
Workarround: Delete the catalog and recreate a new one for each chart update.

Details:

Rancher container's log
```
2020/05/05 15:54:57 [INFO] Installing chart using helm version: rancher-helm
```

Rancher catalog object on API .../v3/catalog/
```
...
"description": "",
"helmVersion": "helm_v3",
"id": "mycatalog",
"kind": "helm",
...
```
Rancher template object on API .../v3/templates
```
...
"status": {
"helmVersion": "helm_v3",
"type": "/v3/schemas/templateStatus"
}
...
```

Rancher templateVersion object on API .../v3/templateVersions/ just after it is added

```
...
"status": {
"type": "/v3/schemas/templateVersionStatus"
}
...
```

TemplateVersion for a object added during the template creation:

```
"status": {
"helmVersion": "helm_v3",
"type": "/v3/schemas/templateVersionStatus"
},
```

Source code [controller.go](https://github.com/rancher/rancher/blob/master/pkg/controllers/user/helm/controller.go) expecting the field **HelmVersion** on **TemplateVersion.Status** in order to choose the helm version.

```
func (l *Lifecycle) getHelmVersion(obj *v3.App) (string, error) {
	templateVersionID, templateVersionNamespace, err := hCommon.ParseExternalID(obj.Spec.ExternalID)
	if err != nil {
		return "", err
	}
	templateVersion, err := l.TemplateVersionLister.Get(templateVersionNamespace, templateVersionID)
	if err != nil {
		return "", err
	}
	return templateVersion.TemplateVersion.Status.HelmVersion, nil
}
```

batiati:
Workarround:

1. Create an empty index
2. Refresh the catalog in order to remove all templates
3. Deploy the updated index.yaml 
4. Refresh the catalog again to create new templates with the correct helm version

I included a simple script on my deployment pipeline:

```
cp -f ./empty.yaml index.yaml
rancher catalog refresh mycatalog -w
helm package mychart
helm repo index .
rancher catalog refresh mycatalog -w
```

Issue #27252, opened 2020-05-26
[2.4.3] Error: apiVersion 'v2' is not valid. The value must be "v1" : exit status 1 on Helm 3 Catalog


Lebvanih:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
The version of helm to create the charts in the next steps is:
```
./helm version
version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d418ded79a", GitTreeState:"clean", GoVersion:"go1.13.10"}
```
1 - Create a new application using helm: `./helm create rancher-v3-issue`. It should directly create a Helm 3, but ensure in the Charts.yaml that apiVersion is set to v2.
2 - Create a new git repository and place that application in it. By default the version is 0.1.0 so the current layout for me was:
```
.
└── charts
    └── rancher-v3-issue
        └── 0.1.0
            ├── charts
            ├── Chart.yaml
            ├── templates
            │   ├── deployment.yaml
            │   ├── _helpers.tpl
            │   ├── hpa.yaml
            │   ├── ingress.yaml
            │   ├── NOTES.txt
            │   ├── serviceaccount.yaml
            │   ├── service.yaml
            │   └── tests
            │       └── test-connection.yaml
            └── values.yaml
```
3 - Add a new catalog in rancher using that git repository. Make sure to select Helmv3. Then deploy the application. No issue should be encountered
4 - Duplicate the 0.1.0 directory into 0.1.1 and edit the Chart.yaml file to reflect the change of value. In my case:
```
.
└── charts
    └── rancher-v3-issue
        ├── 0.1.0
        │   ├── charts
        │   ├── Chart.yaml
        │   ├── templates
        │   │   ├── deployment.yaml
        │   │   ├── _helpers.tpl
        │   │   ├── hpa.yaml
        │   │   ├── ingress.yaml
        │   │   ├── NOTES.txt
        │   │   ├── serviceaccount.yaml
        │   │   ├── service.yaml
        │   │   └── tests
        │   │       └── test-connection.yaml
        │   └── values.yaml
        └── 0.1.1
            ├── charts
            ├── Chart.yaml
            ├── templates
            │   ├── deployment.yaml
            │   ├── _helpers.tpl
            │   ├── hpa.yaml
            │   ├── ingress.yaml
            │   ├── NOTES.txt
            │   ├── serviceaccount.yaml
[...]

PR #27599, opened 2020-06-15
Add helm version to toCreate on update

brendarearden:
**Problem**
When a new version is added to a helm 3 catalog that is already in rancher, update is triggered instead of create, so the new version is not getting the helm version assigned to it

**Solution**
Add helm version to the update method to ensure new versions created on update have the helm version assigned

**Issue** 
#27252

### DIFF
--- a/tests/integration/suite/test_app.py
+++ b/tests/integration/suite/test_app.py
@@ -647,6 +647,107 @@ def test_app_has_helmversion(admin_pc, admin_mc, remove_resource):
     assert app2.helmVersion == "helm_v3"
 
 
+def test_app_upgrade_has_helmversion(admin_pc, admin_mc, remove_resource):
+    """Test helm version exists on new chart versions when added to an
+    existing catalog and that the helm version carries through template,
+    templateVersion and app on upgrade"""
+    app_client = admin_pc.client
+    catalog_client = admin_mc.client
+    catalog_name = random_str()
+    app1_name = random_str()
+    app2_name = random_str()
+    helm_3 = 'helm_v3'
+    cat_base = "catalog://?catalog=" + catalog_name + \
+               "&template=rancher-v3-issue&version="
+
+    helm3_catalog = catalog_client.create_catalog(
+        name=catalog_name,
+        branch="helmversion-onupdate-1v",
+        url=DEFAULT_CATALOG,
+        helmVersion=helm_3
+    )
+    remove_resource(helm3_catalog)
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_app.py
```diff
@@ -647,6 +647,92 @@ def test_app_has_helmversion(admin_pc, admin_mc, remove_resource):
...
+    # catalog
+    catalog_data = {
+      'name': catalog_name,
+      'branch': "helmversion-onupdate-2v",
+      'url': "https://github.com/brendarearden/integration-test-charts",
+      'helmVersion': "helm_v3"
+    }
+    helm3_catalog = catalog_client.update(helm3_catalog, catalog_data)
+
+    def ensure_udpated_catalog(catalog):
```
rmweir, line 696: udpated -> updated

batiati:
Fixes #26920

tests/integration/suite/test_app.py
```diff
@@ -647,6 +647,92 @@ def test_app_has_helmversion(admin_pc, admin_mc, remove_resource):
...
+    catalog_client = admin_mc.client
+    catalog_name = random_str()
+    app1_name = random_str()
+    app2_name = random_str()
+    cat_base = "catalog://?catalog=" + catalog_name +\
+        "&template=rancher-v3-issue&version="
+    helm3_catalog = catalog_client.create_catalog(
+      name=catalog_name,
+      branch="helmversion-onupdate-1v",
+      url="https://github.com/brendarearden/integration-test-charts",
```
dramich, line 663: This needs to live under a rancher repo

tests/integration/suite/test_app.py
```diff
@@ -647,6 +647,92 @@ def test_app_has_helmversion(admin_pc, admin_mc, remove_resource):
...
+        templatesString = ','.join([str(i) for i in templates])
+        if "0.1.1" in templatesString:
+            return catalog
+        return None
+    helm3_catalog = wait_for(
+      lambda: ensure_updated_catalog(helm3_catalog),
+      fail_handler=lambda:
+      "Timed out waiting for catalog to stop transitioning")
+    project_client = user_project_client(admin_pc, admin_pc.project)
+    # update existing app with new version to ensure correct
```
dramich, line 708: Ideally the template should be checked as that is really our source of truth, the app could be holding old data from pre upgrade thus giving a false positive.

dramich: Both can be checked and this also tests the upgrade in general.

PR #27601, opened 2020-06-16
Enable /v1/clusters for setups w/ no local cluster

ibuildthecloud:


cjellick:
Does an issue need opened for this so that QA can test it? Not obvious from the title and my 2 second glance at the code

Issue #27618, opened 2020-06-16
Panic seen in rancher server and rancher server keeps restarting


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster through API (Used Jenkins job)
**rke config of cluster created through automation:**
```
rke_config = {
    "addonJobTimeout": 30,
    "authentication":
    {"strategy": "x509",
     "type": "authnConfig"},
    "ignoreDockerVersion": True,
    "ingress":
        {"provider": "nginx",
         "type": "ingressConfig"},
    "monitoring":
        {"provider": "metrics-server",
         "type": "monitoringConfig"},
    "network":
        {"plugin": "canal",
         "type": "networkConfig",
         "options": {"flannel_backend_type": "vxlan"}},
    "services": {
        "etcd": {
            "extraArgs":
                {"heartbeat-interval": 500,
                 "election-timeout": 5000},
            "snapshot": False,
            "backupConfig":
                {"intervalHours": 12, "retention": 6, "type": "backupConfig"},
            "creation": "12h",
            "retention": "72h",
            "type": "etcdService"},
        "kubeApi": {
            "alwaysPullImages": False,
            "podSecurityPolicy": False,
            "serviceNodePortRange": "30000-32767",
            "type": "kubeAPIService"}},
    "sshAgentAuth": False}
```
- Wait for the cluster to come up.
- Edit cluster, set `Drain : True`
- Save the changes made
- The rancher server panics. And keeps restarting

**Rancher logs:**
```
2020/06/16 21:10:17 [INFO] Removing container [etcd-fix-perm] on host [ip], try #1







2020/06/16 21:10:17 [INFO] cluster [c-mvsr4] provisioning: [remove/etcd-fix-perm] Successfully removed container on host [ip]
2020/06/16 21:10:17 [INFO] Image [rancher/rke-tools:v0.1.58] exists on host [ip]
2020/06/16 21:10:18 [INFO] Starting container [rke-log-linker] on host [ip], try #1







2020/06/16 21:10:18 [INFO] cluster [c-mvsr4] provisioning: [etcd] Successfully started [rke-log-linker] container on host [ip]
2020/06/16 21:10:18 [INFO] Removing container [rke-log-linker] on host [ip], try #1







2020/06/16 21:10:18 [INFO] cluster [c-mvsr4] provisioning: [remove/rke-log-linker] Successfully removed container on host [ip]
[...]

PR #27629, opened 2020-06-17
[2.4] vendor rke

cbron:


mrajashree:
https://github.com/rancher/rancher/issues/27618

Issue #26543, opened 2020-04-07
[BUG] in Rancher Logging, Tests are not indicative of whether the config will work or not


Tejeev:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

1. Select ES as target for cluster logging
1. Use `Edit as File` mode and input the correct config
2. Click `Dry Run`

**Result:**
Dry run fails.  Even the config is correct.

**Environment information**
- Rancher: v2.3.5

**Tested on:**
Rancher Launched (Provider) EC2 cluster
- Kubernetes: v1.17.2
- Elasticsearch: standalone docker container elasticsearch:7.6.0
- Logging: v0.1.4

Rancher Launched (Provider) Google GKE cluster
- Kubernetes: v1.15.9-gke.24
- Elasticsearch: standalone docker container elasticsearch:7.6.0
- Logging: v0.1.4

Tejeev:
This is a result of #26559 so I am closing this issue as a duplicate

aiwantaozi:
@Tejeev could you paste the configure you just input in UI here?

loganhz:
I cannot reproduce it locally. But it did happen in the user's env. The config looks good.

aiwantaozi:
Not sure what tj got, but after I try it, and find we need to update some code base on the lastest rancher code and new fluentd version.

I got this error after click dry-run
![image](https://user-images.githubusercontent.com/5936271/85124710-efe94900-b25c-11ea-9a66-8154d49690ad.png)

configure
```
        @type elasticsearch
	include_tag_key  true
	hosts http://18.163.101.228:9200    
	logstash_prefix "ec2-2"
	logstash_format true
	logstash_dateformat  %Y-%m-%d
	type_name  "container_log"
```

and got fluentd warning [Deprecate top-level match](https://docs.fluentd.org/deployment/logging#deprecate-top-level-match)

izaac:
Rancher version `v2.4-head (07/24/2020)` _29be104e6_
Rancher version `master-head (07/24/2020)` _80542c31b_

This is now showing an error in the UI if the dry run failed. 

This was previously reproduced in `v2.4.5` https://github.com/rancher/rancher/issues/27695#issuecomment-648973825

cbron:
@aiwantaozi fluentD v0.1.20 has been tagged, can you PR that in. Afterwards @izaac we should re-test with those changes.

izaac:
[...]

PR #27660, opened 2020-06-19
[Master] Fix logging dry-run function return error

aiwantaozi:
Problem:
1. This PR bring in some crash https://github.com/rancher/rancher/commit/4dcd2b751f1b2fecf21351b082935f8e12ca9446. it adds appLister but forget to init it here https://github.com/rancher/rancher/blob/master/pkg/controllers/user/logging/deployer/testerdeployer.go#L30. Then when we use dry run, and server got crash, return 500, but UI still show success.

2. After fluentd 1.x <match fluent.**> will get deprecated warning.

Solution:
1. Init appLister
2. Add latest fluentd suggested format <label @FLUENT_LOG>

Issue:

### DIFF
--- a/pkg/api/norman/customization/logging/action.go
+++ b/pkg/api/norman/customization/logging/action.go
@@ -41,12 +41,21 @@ var (
 	fluentdCommand      = "fluentd -q --dry-run -i "
 	cleanCertDirCommand = "rm -rf "
 	tmpCertDirPrefix    = "/fluentd/etc/config/tmpCert"
+	/* Default configuration <match fluent.*> in old version rancher/fluentd is deprecated,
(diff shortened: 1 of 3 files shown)


pkg/api/norman/customization/logging/action.go
```diff
@@ -171,7 +181,7 @@ func (h *Handler) dryRunLoggingTarget(apiContext *types.APIContext, level, clust
 
 	podLister := context.Core.Pods(loggingconfig.LoggingNamespace).Controller().Lister()
```
cbron, line 182: This PR is different from the 2.4 one. Doesn't this line need to change to `metav1.NamespaceAll` or is master different ? Same for change to user/controllers.go in other PR.

PR #27675, opened 2020-06-20
Move branches back to dev

cbron:


pkg/settings/setting.go
```diff
@@ -174,7 +174,7 @@ func GetEnvKey(key string) string {
 func getMetadataConfig() string {
 	branch := os.Getenv("RANCHER_METADATA_BRANCH")
 	if branch == "" {
-		branch = "release-v2.4"
+		branch = "dev-v2.4"
```
deniseschannon, line 177: This branch we don't typically change. It's not in the checklist

cbron: It's the last step in parts 6/17. Was dev this morning before I changed it for release: https://github.com/rancher/rancher/pull/27671/files#diff-8cae2b257a070262d733c79cc16f917bR177

PR #27680, opened 2020-06-22
[2.4.6] Fix logging dry-run function return error

aiwantaozi:
Problem:
1. This PR brings in some crash https://github.com/rancher/rancher/commit/4dcd2b751f1b2fecf21351b082935f8e12ca9446. It added appLister but forgot to init it here https://github.com/rancher/rancher/blob/master/pkg/controllers/user/logging/deployer/testerdeployer.go#L30. Then when we do a dry run, server crashes and returns 500, but UI still shows a success.
2. Dry run fails because it can't find the fluentd-test pod. The reason is that userContext from clustermanager does not start controller now, so we can't get pods from controller.
3. After fluentd 1.x <match fluent.**> will get deprecated warning.

Solution:
1. Init appLister
2. Use pod interface client instead of pod controller
3. Add latest fluentd suggested format <label @FLUENT_LOG>

Issue:

Master PR:

https://github.com/rancher/rancher/pull/27660

### DIFF
--- a/pkg/api/customization/logging/action.go
+++ b/pkg/api/customization/logging/action.go
@@ -38,12 +38,21 @@ var (
 	fluentdCommand      = "fluentd -q --dry-run -i "
 	cleanCertDirCommand = "rm -rf "
 	tmpCertDirPrefix    = "/fluentd/etc/config/tmpCert"
+	/* Default configuration <match fluent.*> in old version rancher/fluentd is deprecated,
+	append the latest configuration to avoid the warning from old version rancher/fluentd. */
+	fluentdLabelConfig = `
+<label @FLUENT_LOG>
+    <match fluent.*>
+      	@type null
+    </match>
+</label>`
 )
 
 type Handler struct {
 	clusterManager       *clustermanager.Manager
 	dialerFactory        dialer.Factory
 	appsGetter           projectv3.AppsGetter
+	appLister            projectv3.AppLister
 	projectLister        mgmtv3.ProjectLister
 	templateLister       mgmtv3.CatalogTemplateLister
 	projectLoggingLister mgmtv3.ProjectLoggingLister
@@ -54,6 +63,7 @@ func NewHandler(management *config.ScaledContext, clusterManager *clustermanager
 		clusterManager:       clusterManager,
 		dialerFactory:        management.Dialer,
 		appsGetter:           management.Project,
+		appLister:            management.Project.Apps("").Controller().Lister(),
 		projectLister:        management.Management.Projects("").Controller().Lister(),
 		templateLister:       management.Management.CatalogTemplates("").Controller().Lister(),
 		projectLoggingLister: management.Management.ProjectLoggings("").Controller().Lister(),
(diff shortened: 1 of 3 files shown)


pkg/controllers/user/logging/deployer/appdeployer.go
```diff
@@ -121,12 +121,15 @@ func (d *AppDeployer) isDeploySuccess(targetNamespace string, selector map[strin
 	for {
 		select {
 		case <-ticker.C:
-			pods, err := d.PodLister.List(targetNamespace, labels.Set(selector).AsSelector())
+			pods, err := d.Pods.List(metav1.ListOptions{
```
gitlawr, line 124: Use `ListNamespaced()`

aiwantaozi: Updated

pkg/api/customization/logging/action.go
```diff
@@ -38,12 +38,19 @@ var (
 	fluentdCommand      = "fluentd -q --dry-run -i "
 	cleanCertDirCommand = "rm -rf "
 	tmpCertDirPrefix    = "/fluentd/etc/config/tmpCert"
+	fluentdLabelConfig  = `
+<label @FLUENT_LOG>
+    <match fluent.*>
+      	@type null
+    </match>
+</label>`
```
cbron, line 46: Not very familiar with fluent, but wouldn't this get updated in [`MatchTemplate`](https://github.com/rancher/rancher/blob/master/pkg/controllers/user/logging/generator/templatematch.go#L5), why instead append to `dryRunConfigBuf` ?

pkg/api/customization/logging/action.go
```diff
@@ -166,9 +174,9 @@ func (h *Handler) dryRunLoggingTarget(apiContext *types.APIContext, level, clust
 		return err
 	}
 
-	podLister := context.Core.Pods(loggingconfig.LoggingNamespace).Controller().Lister()
+	pods := context.Core.Pods(loggingconfig.LoggingNamespace)
```
dramich, line 177: This should continue to use a lister, if it's not being initialized then the initialization can be moved to ensure the cache is primed once we get here.

pkg/api/customization/logging/action.go
aiwantaozi: Dry run will deploy fluentd-tester which use the default fluentd configure here. https://github.com/rancher/fluentd/blob/master/package/fluent.conf#L4 , now fluentd will return warning when we use <match fluent.*>, so append fluentd label configure to avoid this warning. Because new release rancher may with old fluentd-tester deployed, so update the configure in rancher.

Related fluentd doc: 
https://docs.fluentd.org/deployment/logging#capture-fluentd-logs

pkg/api/customization/logging/action.go
aiwantaozi: Found that 2.4 is different with master, in 2.4 we don't start the controller in userContext here https://github.com/rancher/rancher/blob/release/v2.4/pkg/clustermanager/manager.go#L428, so we can't use lister here, no sure we can simply start the controller or need more changes.

pkg/api/customization/logging/action.go
cbron: Can we add it in fluent.conf also, leaving both ? If so would be good to leave comments around reasoning for having both.

aiwantaozi: Sure, add a comment [here](https://github.com/rancher/rancher/pull/27680/files#diff-5cc90d611c97b997b67fbc56fe215abdR41) and create a fluentd pr https://github.com/rancher/fluentd/pull/30

cbron: lgtm now, adding jacob as reviewer to second PR

pkg/api/customization/logging/action.go
dramich: That link goes to where we do start the controller, is there another issue being seen?

aiwantaozi: Yes, the start function passes false to the third parameter, so I can't get pods by Lister from the userContext.
```
func (m *Manager) start(ctx context.Context, cluster *v3.Cluster, controllers, clusterOwner bool) (*record, error) 
```

dramich: Expand `RegisterFollower` to include the resource needed, that will allow the cache to be ready.

aiwantaozi: Updated, thank you

PR #27688, opened 2020-06-22
Do not require "local://" principal on users

ibuildthecloud:
Do enable a more "gitops" friendly flow we'd rather not prefer this
redudany field was required.  If one uses kubectl apply to create a
user with an external principal and then the controller adds
"local://" is modifies the principal list which is hard to reconcile
later on a subsequent apply.

### DIFF
--- a/pkg/auth/providers/common/usermanager.go
+++ b/pkg/auth/providers/common/usermanager.go
@@ -619,7 +619,18 @@ func userByPrincipal(obj interface{}) ([]string, error) {
 		return []string{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
When steve runs we do not run any auth controllers.  This means in the steve context the "local://" principal is never auto added.  If Rancher is ran we still add the "local://" principal ID because I think it might be used for another purpose.  It would be good to remove the behavior of auto-adding "local://" in general.  Right now this is just a patch to make it work with Steve.

Issue #20565, opened 2019-05-30
Cluster logging not shipping control plane and etcd logs


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Install Rancher and prepare two hosts.
2. Create custom cluster using two hosts, 1 etcd/control plane, 1 worker
3. Install splunk (can use one of the hosts or separate server). Use:
```
docker run -d -p 8000:8000 -p 8088:8088 -e "SPLUNK_START_ARGS=--accept-license" -e "SPLUNK_PASSWORD=Pass0word!" --name splunk splunk/splunk:latest
```
4. Create event collector and token in Splunk. See https://docs.splunk.com/Documentation/Splunk/7.0.0/Data/UsetheHTTPEventCollector#Configure_HTTP_Event_Collector_on_Splunk_Enterprise
5. Go to Logging option of newly created cluster. Select Splunk, enter endpoint URL with port 8088 and token created in step 4. You may need to disable SSL in global settings to get port 8088 up. 

**Result:**
The logging pods are only created on the worker and not the etcd/control plane node. The logging pod daemon sets should be created on *all* nodes.

```
> kubectl get nodes
NAME           STATUS   ROLES               AGE     VERSION
ip-10-0-0-10   Ready    controlplane,etcd   9m59s   v1.13.5
ip-10-0-0-11   Ready    worker              8m9s    v1.13.5
> kubectl get pods -n cattle-logging -o wide
NAME                                   READY   STATUS    RESTARTS   AGE   IP          NODE           NOMINATED NODE   READINESS GATES
rancher-logging-fluentd-f74t9          2/2     Running   0          81s   10.42.1.7   ip-10-0-0-11   <none>           <none>
rancher-logging-log-aggregator-lz4bq   1/1     Running   0          81s   10.42.1.8   ip-10-0-0-11   <none>           <none>
> kubectl get ds -n cattle-logging
NAME                             DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
rancher-logging-fluentd          1         1         1       1            1           <none>          2m42s
rancher-logging-log-aggregator   1         1         1       1            1           <none>          2m42s
```
Appears to be caused by incorrect tolerations set on the pod for the taints on the nodes:
```
> kubectl get node ip-10-0-0-10 -o yaml | grep -A6 taints
  taints:
  - effect: NoSchedule
[...]

PR #27690, opened 2020-06-22
add support for configuring log aggregator tolerations

paynejacob:
For: #20565 

When logging is deployed tolerations are captured from the logging configuration and passed to the helm chart.

### DIFF
--- a/pkg/controllers/user/logging/deployer/appdeployer.go
+++ b/pkg/controllers/user/logging/deployer/appdeployer.go
@@ -141,10 +141,38 @@ func (d *AppDeployer) isDeploySuccess(targetNamespace string, selector map[strin
 	}
 }
 
-func rancherLoggingApp(appCreator, systemProjectID, catalogID, driverDir, dockerRoot string) *projectv3.App {
+func rancherLoggingApp(appCreator, systemProjectID, catalogID, driverDir, dockerRoot string, tolerations []k8scorev1.Toleration) *projectv3.App {
 	appName := loggingconfig.AppName
 	namepspace := loggingconfig.LoggingNamespace
 	_, systemProjectName := ref.Parse(systemProjectID)
+
+	answers := map[string]string{
+		//compatible with old version
+		"fluentd.enabled":              "true",
+		"fluentd.cluster.dockerRoot":   dockerRoot,
+		"log-aggregator.enabled":       "true",
+		"log-aggregator.flexVolumeDir": driverDir,
+
+		//new version
+		"fluentd.fluentd-linux.enabled":                     "true",
+                "fluentd.fluentd-linux.cluster.dockerRoot":          dockerRoot,
+		"log-aggregator.log-aggregator-linux.enabled":       "true",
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/logging/deployer/deployer.go
```diff
@@ -150,7 +151,25 @@ func (d *Deployer) deployRancherLogging(systemProjectID, appCreator string) erro
 		return err
 	}
 
-	app := rancherLoggingApp(appCreator, systemProjectID, templateVersion.ExternalID, driverDir, dockerRootDir)
+	var clusterLogging *mgmtv3.ClusterLogging
+	if cls, err := d.clusterLoggingLister.List("", labels.NewSelector()); err == nil {
+		for _, cl := range cls {
+			if cl.Spec.ClusterName == cluster.Name {
+				clusterLogging = cl
+			}
```
cbron, line 159: * Do we also need to also check projectLogging, not just clusterLogging ? Seems like issue only calls out cluster, but I see that types have it on project also. Was thought that its not bad idea to add to project anyway to make uniform, or was there some other reason ? 
* Instead of doing this loop, would think it will be fast to just pass the object in from sync. That would also solve the project case. Could also pass key and do a `Get` but thats extra call and you'd have to do on both project/cluster lister.

cbron: Looking at part 2 again, sync is called without the logger key/obj in `NodeSync` and `ClusterSync`, so maybe this is the best way.

pkg/controllers/user/logging/deployer/appdeployer.go
```diff
@@ -141,11 +142,40 @@ func (d *AppDeployer) isDeploySuccess(targetNamespace string, selector map[strin
...
+		"log-aggregator.log-aggregator-linux.flexVolumeDir": driverDir,
+	}
+
+	if tolerations != nil {
+		for i, toleration := range *tolerations {
+			// linux
+			answers[fmt.Sprintf("log-aggregator.log-aggregator-linux.tolerations[%d].key", i)] = toleration.Key
+			answers[fmt.Sprintf("log-aggregator.log-aggregator-linux.tolerations[%d].value", i)] = toleration.Value
+			answers[fmt.Sprintf("log-aggregator.log-aggregator-linux.tolerations[%d].effect", i)] = string(toleration.Effect)
+			answers[fmt.Sprintf("log-aggregator.log-aggregator-linux.tolerations[%d].operator", i)] = string(toleration.Operator)
```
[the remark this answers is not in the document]
paynejacob: This will not fix the problem but allows users to configure tolerations in a way that schedules them on the control plane and etcd nodes.  This will require a UI change.

pkg/controllers/user/logging/deployer/deployer.go
paynejacob: Project logging just modifies the fluentd configuration, this changes only affects the deployment of containers which is handled by cluster logging.

paynejacob:
This functionality exists in logging v2 on rancher 2.5.0. This version of logging is deprecated.

Issue #25383, opened 2020-02-12
Deploying applications through local disk to Rancher yields installation unusable


patrickdappollonio:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug (on Rancher 2.2.9, haven't tried 2.3.x)

**Steps to reproduce (least amount of steps as possible):**

* Create a cluster through Rancher
* Clone to a local directory any helm chart that requires dependencies, in our
    testing, the Helm Chart for Kong
    [at this specific commit](https://github.com/helm/charts/commit/0580e453e)
    was tested, but the situation can also be replicated by using a helm chart
    that's broken (for example, a template syntax error) or that has
    dependencies, and then `cd` into the folder `charts/stable/kong`.

 ```bash
# Clone helm chart
cd $(mktemp -d) && git clone git@github.com:helm/charts.git

# Reset to specific commit (YMMV testing with others, see below)
cd charts && git reset --hard 0580e453e

# cd into the specific chart
cd charts/stable/kong

# Install it as an app
rancher app install . kong
```

* With a `rancher` CLI properly configured pointing at the cluster created with
    the first step, install the app by executing `rancher apps install . kong`.
    To accelerate the process, you can create a few other copies, like `kong2`,
    `kong3`, and so on.

**Expected Result:**
* Rancher will display the fact that the installation failed because the
    Helm Chart dependencies weren't installed (`helm dep update`) before
    sending the app to install.

**Actual Result:**

* This is a chain reaction of events:
  * The application shows correctly as broken in the UI
  * After a couple of minutes, the Rancher UI becomes unresponsive with the
      error: `etcdserver: mvcc: database space exceeded`
  * Rancher can't be used anymore from the UI, the CLI or directly using
      `kubectl`.

The reason behind the issue is that each helm install failure seems to cause
Rancher to attach the payload of the repository cloned in the CRD
`apps.project.cattle.io` (since it's the way to be able to store it due to being
installed from a local directory). In an etcd using multi-version concurrency
control, etcd ends up holding to several large etcd document revisions (which
include each file from the folder, base64-encoded in the list of files) which
causes a massive increase in the etcd database size, preventing it from running
compaction and defragment tasks in-time resulting in the cluster hitting the
maximum database size and then switching to read-only mode.

https://github.com/etcd-io/etcd/blob/master/Documentation/faq.md#what-does-mvcc-database-space-exceeded-mean-and-how-do-i-fix-it

**Other details that may be helpful:**

[...]

PR #27714, opened 2020-06-23
Change app status if true

brendarearden:
**Problem**
When `helm template` or `helm install` fails, the app continues to retry installing and the error backoff never happens so disk space fills up. This appears to be occurring due to the app status getting set to false and then set back to unknown which updates the app continuously.

**Solution**
Only change the app install status back to unknown if the app install status is true. This will allow for the controller back off to occur

**Issue**
#25383

### DIFF
--- a/pkg/controllers/user/helm/controller.go
+++ b/pkg/controllers/user/helm/controller.go
@@ -228,7 +228,7 @@ func (l *Lifecycle) Updated(obj *v3.App) (runtime.Object, error) {
 func (l *Lifecycle) DeployApp(obj *v3.App) (*v3.App, error) {
 	obj = obj.DeepCopy()
 	var err error
-	if !v3.AppConditionInstalled.IsUnknown(obj) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/helm/controller.go
```diff
@@ -228,7 +228,7 @@ func (l *Lifecycle) Updated(obj *v3.App) (runtime.Object, error) {
 func (l *Lifecycle) DeployApp(obj *v3.App) (*v3.App, error) {
 	obj = obj.DeepCopy()
 	var err error
-	if !v3.AppConditionInstalled.IsUnknown(obj) {
+	if !v3.AppConditionInstalled.IsUnknown(obj) && !v3.AppConditionInstalled.IsFalse(obj) {
```
dramich, line 231: How about:
```suggestion
	if v3.AppConditionInstalled.IsTrue(obj) {
```

Issue #27031, opened 2020-05-11
Cluster owners can't scale nodepool if the node template has been created by a different user


spiette:
**Bug**

**Steps to reproduce (least amount of steps as possible):**
- As user *A*, create nodetemplates, and create a cluster with nodepool using this nodetemplates
- As user *A*, add user *B* as cluster owner
- As user *B*, try to scale up the node pool 

**Result:**

Error updating node pool scale: unable to find node template [cattle-global-nt:nt-u-t2k5pklxw3-nt-f9sw5]: undefined

**Other details that may be helpful:**

**Environment information**
- Rancher version (2.3.6 and 2.4.2)
- Installation option (single install/HA): HA
- Kubernetes 1.17

gzrancher/rancher#10033

aaronRancher:
The bug is reproduced with Rancher server `rancher\rancher:v2.3.6`

Steps:
* As user A, create node template. Create a cluster with node pool using new node template.
* As user A, add user B as cluster owner
* As user B, try to scale up the node pool

Result:
User B is unable to scale up node pool
```
unable to find node template [cattle-global-nt:nt-7s2qq]: undefined
```
![Screen Shot 2020-06-24 at 4 25 09 PM](https://user-images.githubusercontent.com/57409052/85637164-51336280-b637-11ea-836a-3a4880be0d26.png)

aaronRancher:
The bug fix is verified with Rancher server `rancher\rancher:v2.4-head` `version fb2c5fb2d`

Reopening for a fix in master-head

luthermonson:
https://github.com/rancher/rancher/pull/27959

aaronRancher:
The bug fix is verified with Rancher server `rancher\rancher:master-head` `version 16616de9c`

PR #27715, opened 2020-06-24
[2.4] Update NodePool Validator

luthermonson:
**Problem:** User added to cluster via cluster role binding can not update nodepool quantity.
**Cause:** Nodepool validator was only checking if the user has access to the entire nodetemplate. Since the added user had no access to the nodetemplate created by another use the error was being thrown.
**Fix:** Updated the validator to do the following.
First, check the user has access to the nodepool, this will pass for both original and added user.
Second, pull the nopepool from the lister. If the user is attempting to make a change to pool's template, validate the user has access to the new pool before letting them change it.
Validate everything else, -/+ to the nodepool quantity will validate as it is a PUT with the existing nodetemplate but a different quantity value passed.

https://github.com/rancher/rancher/issues/27031

### DIFF
--- a/pkg/api/customization/nodepool/validator.go
+++ b/pkg/api/customization/nodepool/validator.go
@@ -2,22 +2,58 @@ package nodepool
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	mgmtSchema "github.com/rancher/types/apis/management.cattle.io/v3/schema"
 	mgmtclient "github.com/rancher/types/client/management/v3"
 )
 
-func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
-	nodetemplateID, _ := data["nodeTemplateId"].(string)
+type Validator struct {
+	NodePoolLister v3.NodePoolLister
+}
+
+func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	// validate access to nodetemplate
+	nodetemplateID, ok := data["nodeTemplateId"].(string)
+	if !ok {
+		// nodetemplate not passed, nothing to check
+		return nil
+	}
+	if request.ID == "" {
+		// creating new pool, confirm access to template
+		return checkNodetemplateAccess(request, nodetemplateID)
+	}
+
+	// validate request ID is in the right format
+	split := strings.SplitN(request.ID, ":", 2)
+	if len(split) != 2 {
+		return httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("unable to find nodepool [%s]", request.ID))
+	}
+	cluster, nodepool := split[0], split[1]
+	np, err := v.NodePoolLister.Get(cluster, nodepool)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/api/customization/nodepool/validator.go
```diff
@@ -2,22 +2,51 @@ package nodepool
...
-func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
-	nodetemplateID, _ := data["nodeTemplateId"].(string)
-	if err := access.ByID(request, &mgmtSchema.Version, mgmtclient.NodeTemplateType, nodetemplateID, nil); err != nil {
+type Validator struct {
+	NodePoolLister v3.NodePoolLister
+}
+
+func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	// does we have access to the nodepool?
+	if err := access.ByID(request, &mgmtSchema.Version, mgmtclient.NodePoolType, request.ID, nil); err != nil {
```
rmweir, line 21: You shouldn't have to check if the user has access to the nodepool in the nodepool validator. I believe an auth check is already being done for every resource request. We usually do this when another resource is being referenced in the body of the request..

pkg/api/customization/nodepool/validator.go
```diff
@@ -2,22 +2,51 @@ package nodepool
...
+	//check if nodetemplate changing
+	split := strings.SplitN(request.ID, ":", 2)
+	cluster, nodepool := split[0], split[1]
 
+	np, err := v.NodePoolLister.Get(cluster, nodepool)
+	if err != nil {
+		return err
 	}
+
+	nodetemplateID, _ := data["nodeTemplateId"].(string)
```
rmweir, line 37: I think it's possible for nodeTemplate to not be in the request if its an update. This should also be allowed

pkg/api/customization/nodepool/validator.go
```diff
@@ -2,22 +2,51 @@ package nodepool
...
+	if np.Spec.NodeTemplateName != nodetemplateID {
+		//update attempt to the nodetemplate, confirm access
+		if err := access.ByID(request, &mgmtSchema.Version, mgmtclient.NodeTemplateType, nodetemplateID, nil); err != nil {
+			if httperror.IsNotFound(err) {
+				return httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("unable to find node template [%s]", nodetemplateID))
+			}
+			if httperror.IsForbidden(err) {
+				return httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("unable to access node template [%s]", nodetemplateID))
+			}
```
rmweir, line 46: It's advisable to return the same error message for NotFound and Forbidden. This makes it more difficult for people to confirm the existence of resources they do not have access to.

tests/integration/suite/test_node.py
```diff
@@ -273,6 +275,79 @@ def test_user_access_to_other_template(user_factory, remove_resource):
...
+                                           user_factory,
+                                           remove_resource,
+                                           wait_remove_resource):
+    """Test that a cluster created by the admin is accessible by another user
+    added as a cluster-owner, validate nodepool changing and switching
+    nodetemplate"""
+
+    # make an admin and user client
+    admin_client = admin_mc.client
+    k8sclient = CoreV1Api(admin_mc.k8s_client)
```
dramich, line 288: The k8s client shouldn't be needed, the admin_client has access to namespaces.

luthermonson:
https://github.com/rancher/rancher/issues/27031

PR #27741, opened 2020-06-26
Add remaining workload types to RBAC tests

aaronRancher:
Adds stateful set, cronjob, and job workload types to workload rbac tests.

### DIFF
--- a/tests/validation/tests/v3_api/test_workload.py
+++ b/tests/validation/tests/v3_api/test_workload.py
@@ -6,6 +6,7 @@ from rancher import ApiError
 namespace = {"p_client": None, "ns": None, "cluster": None, "project": None}
 RBAC_ROLES = [CLUSTER_OWNER, PROJECT_MEMBER, PROJECT_OWNER,
               PROJECT_READ_ONLY, CLUSTER_MEMBER]
+WORKLOAD_TYPES = ["daemonSet", "statefulSet", "cronJob", "job"]
 
 if_check_lb = os.environ.get('RANCHER_CHECK_FOR_LB', "False")
 if_check_lb = pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


tests/validation/tests/v3_api/test_workload.py
```diff
@@ -995,83 +982,70 @@ def test_rbac_wl_daemonset_update(role, remove_resource):
...
+        workload = create_workload_by_type(p_client, name, con, ns, config)
         wait_for_wl_to_active(p_client, workload)
         with pytest.raises(ApiError) as e:
             user_token = rbac_get_user_token_by_role(role)
             p_client = get_project_client_for_token(project, user_token)
             con = [{"name": "test1", "image": os.environ.get('RANCHER_TEST_IMAGE',
                                                          "nginx")}]
             p_client.update(workload, containers=con)
+            wait_for_pods_in_workload(p_client, workload)
+            validate_workload(p_client, workload, "deployment", ns.name)
```
jiaqiluo, line 994: the type is not only deployment.

tests/validation/tests/v3_api/test_workload.py
```diff
@@ -995,83 +982,70 @@ def test_rbac_wl_daemonset_update(role, remove_resource):
...
 @pytest.mark.parametrize("role", RBAC_ROLES)
-def test_rbac_wl_daemonset_update_negative(role, remove_resource):
+@pytest.mark.parametrize("config", WORKLOAD_TYPES)
+def test_rbac_wl_parametrize_update_negative(role, remove_resource, config):
     if role == CLUSTER_OWNER:
-        # cluster owner can edit daemonsets in any project in the cluster
-        project = rbac_get_project()
+        # workloads of type job cannot be edited
+        if config == "job":
+            return
```
jiaqiluo, line 1007: how about when the role is not cluster owner and config is job?

tests/validation/tests/v3_api/test_workload.py
aaronRancher: Changed to `validate_workload(p_client, workload, config, ns.name)`
This will validate per workload type `["daemonSet", "statefulSet", "cronJob", "job"]`

tests/validation/tests/v3_api/test_workload.py
aaronRancher: Moved this check to the beginning of the method. This should be the first thing that is checked regardless of RBAC role.

Issue #25955, opened 2020-03-11
Can't create resources with . in their names


aemneina:
**What kind of request is this (question/bug/enhancement/feature request):**
bug possibly

**Steps to reproduce (least amount of steps as possible):**
in the 2.2 line users were able to create resources with a `.` in the name. In 2.3 this is no longer the case which throws errors when managing existing resources that have the dot. In 2.3 one cant manage or create resources with a dot in the name. Since `.` is explicitly accepted, along with `-` in the k8s naming convention, I feel this is a regression.

gzrancher/rancher#8937

vincent99:
@aemneina What type of resource are you referring to here?  It's definitely wrong in some places but I don't think there's anything new about this?

There are several different rules for what's allowed as a name depending on the resource.  Most are a DNS hostname (`[a-z.-]`, <= 253 chars) but some are a single DNS label where a dot is correctly not allowed (& <= 63 chars) or have other restrictions.

Those are supposed to be reflected in the Rancher schema, which has types for `dnsLabel`, `hostname`, and `dnsLabelRestricted`, but these are not all applied correctly to each type, or at all (lots are just `string`).

So someone should go through and fix them to match the rule k8s actually has for each type (among the other validation info that's missing and could be getting used).

There's additional complication for some types like Workloads, because we create multiple things from the one name:
  - A deployment (which should be a `hostname`)
  - A service (`dnsLabel`)
  - A container in the deployment (`dnsLabel`)

Current mapping in master (for a project):
```
 "alertmanager": "dnsLabel"
  "alertmanagerEndpoints": "string"
  "app": "dnsLabel"
  "appRevision": "dnsLabel"
  "applyAppConfig": "string"
  "basicAuth": "dnsLabel"
  "bitbucketCloudPipelineConfig": "dnsLabel"
  "bitbucketCloudProvider": "dnsLabel"
  "bitbucketServerPipelineConfig": "dnsLabel"
  "bitbucketServerProvider": "dnsLabel"
  "certificate": "dnsLabel"
  "configMap": "dnsLabel"
  "configMapEnvSource": "string"
  "configMapKeySelector": "string"
  "configMapProjection": "string"
  "configMapVolumeSource": "string"
  "container": "string"
  "containerPort": "string"
  "containerStatus": "string"
  "cronJob": "dnsLabelRestricted"
  "crossVersionObjectReference": "string"
  "daemonSet": "dnsLabelRestricted"
  "deployment": "dnsLabelRestricted"
[...]

PR #27751, opened 2020-06-26
Update types dependency

ryansann:
Types PR:
- rancher/types#1159

Issue:
- rancher/rancher#25955

### DIFF
--- a/vendor/github.com/rancher/rancher/pkg/auth/providers/common/usermanager.go
+++ b/vendor/github.com/rancher/rancher/pkg/auth/providers/common/usermanager.go
@@ -619,7 +619,18 @@ func userByPrincipal(obj interface{}) ([]string, error) {
 		return []string{}, nil
 	}
 
-	return u.PrincipalIDs, nil
+	match := false
+	for _, id := range u.PrincipalIDs {
+		if strings.HasPrefix(id, "local://") {
+			match = true
+			break
+		}
+	}
+
+	if match {
+		return u.PrincipalIDs, nil
+	}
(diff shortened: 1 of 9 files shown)


vendor/github.com/rancher/rancher/pkg/auth/providers/common/usermanager.go
```diff
@@ -619,7 +619,18 @@ func userByPrincipal(obj interface{}) ([]string, error) {
 		return []string{}, nil
 	}
 
-	return u.PrincipalIDs, nil
+	match := false
+	for _, id := range u.PrincipalIDs {
+		if strings.HasPrefix(id, "local://") {
+			match = true
```
yogeshmundada, line 625: Just trying to understand this part, would the logic change if we just return at this point?

vendor/github.com/rancher/rancher/pkg/auth/providers/local/local_auth_provider.go
```diff
@@ -141,6 +141,9 @@ func getLocalPrincipalID(user *v3.User) string {
 			principalID = p
 		}
 	}
+	if principalID == "" {
+		return Name + "://" + user.Name
```
yogeshmundada, line 145: Nit: I would just use fmt.Sprintf()

vendor/github.com/rancher/rancher/pkg/auth/providers/common/usermanager.go
ryansann: This got pulled in because of `go mod vendor`, my changes were for the `rancher/types` related stuff

PR #27768, opened 2020-06-28
Adding changes to include a test for passing a specific `K8S_VERSION` to the cluster.

sowmyav27:
Adding changes to include a test for passing a specific `K8S_VERSION` to the cluster.

### DIFF
--- a/tests/validation/lib/rke_client.py
+++ b/tests/validation/lib/rke_client.py
@@ -12,6 +12,7 @@ DEBUG = os.environ.get('DEBUG', 'false')
 
 DEFAULT_CONFIG_NAME = 'cluster.yml'
 DEFAULT_NETWORK_PLUGIN = os.environ.get('DEFAULT_NETWORK_PLUGIN', 'canal')
+K8S_VERSION = os.environ.get('RANCHER_K8S_VERSION', "")
 
 
 class RKEClient(object):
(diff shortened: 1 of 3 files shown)


tests/validation/tests/rke/test_install_config.py
```diff
@@ -6,7 +6,7 @@ def test_install_config_1(test_name, cloud_provider, rke_client, kubectl):
     """
     Node Address specified as just IP and using only this in the node spec
```
soumyalj, line 7: Could you add a comment that a specific K8s version is being tested in this test case ? We are replacing cluster_install_config_1.yml.j2 with cluster_install_config_11.yml.j2 . Do we still want to keep cluster_install_config_1.yml.j2 ?

tests/validation/lib/rke_client.py
```diff
@@ -54,7 +55,9 @@ def build_rke_template(self, template, nodes, **kwargs):
         """
         render_dict = {
             'master_ssh_key_path': self.master_ssh_key_path,
-            'network_plugin': DEFAULT_NETWORK_PLUGIN}
+            'network_plugin': DEFAULT_NETWORK_PLUGIN,
+            'k8s_version': K8S_VERSION
+        }
```
soumyalj, line 60: Could we also run the other test_install_config tests to see if they are working ? Its a change in the method build_rke_template()  of rke_client.py. Does this change affect other tests ?

tests/validation/tests/rke/test_install_config.py
sowmyav27: Done.
Yes `cluster_install_config_1.yml.j2` is still available as its being used by another test.

tests/validation/lib/rke_client.py
sowmyav27: Does not affect other tests.

sangeethah:
LGTM

PR #27780, opened 2020-06-30
Adding support matrix for RKE import clusters

sowmyav27:


sangeethah:
In test_generate_rke_config() method, modify call to create_rke_cluster_config()  to include filename.

tests/validation/tests/v3_api/test_deploy_clusters.py
```diff
@@ -60,6 +68,40 @@ def test_deploy_rke():
...
+
+@if_not_auto_deploy_rke_import
+def test_deploy_rke_import():
+    print("Deploying RKE import Clusters")
+    global env_details
+    global cluster_details
+    errors = []
+    # get rke k8s versions
+    rkecommand = "rke config --list-version -a"
+    print(rkecommand)
```
rancher-max, line 83: Nit: remove this print since the command is printed in the following function call

tests/validation/tests/v3_api/test_import_cluster.py
```diff
@@ -76,13 +90,14 @@ def test_generate_rke_config():
...
-    rkeconfigpath = create_rke_cluster_config(aws_nodes)
+    cluster_filename = random_test_name("cluster")
+    rkeconfigpath = create_rke_cluster_config(aws_nodes, cluster_filename)
     rkeconfig = run_command("cat " + rkeconfigpath)
     print("RKE Config file generated\n")
     print(rkeconfig)
 
 
-def create_rke_cluster_config(aws_nodes):
+def create_rke_cluster_config(aws_nodes, cluster_filename):
```
rancher-max, line 100: I think this function is used in `test_create_ha.py` as well. Will need to update there to pass in the `cluster_filename` or set a default for it in this function.

rancher-max: Oh wait... ignore this one. It uses its own definition of the function.

tests/validation/tests/v3_api/test_deploy_clusters.py
sowmyav27: done

PR #27786, opened 2020-06-30
add checks for the count of kubelet restart times after upgrading Rancher

jiaqiluo:


jiaqiluo:
close as the PR is not needed anymore.

PR #27787, opened 2020-06-30
set the CPUCredits to standard to avoid extra charge if CPU utilization stays above the baseline

jiaqiluo:


jiaqiluo:
close as the PR is not needed anymore.

PR #27800, opened 2020-07-01
Support signing global aws requests

rmweir:
**Problem:**
An endpoint format is being assumed that is not compatible with public endpoints.

**Solution:**
us-east-1 will be returned for global endpoints, as mentioned in aws docs: https://docs.aws.amazon.com/general/latest/gr/sigv4_elements.html

**Issue:**
https://github.com/rancher/rancher/issues/27792

### DIFF
--- a/pkg/httpproxy/sign.go
+++ b/pkg/httpproxy/sign.go
@@ -13,11 +13,15 @@ import (
 	"strings"
 	"time"
 
-	"github.com/rancher/norman/httperror"
-
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	v4 "github.com/aws/aws-sdk-go/aws/signer/v4"
+	"github.com/rancher/norman/httperror"
 	v1 "github.com/rancher/types/apis/core/v1"
+)
+
+const (
+	defaultAWSRegion = "us-east-1"
 )
 
 var requiredHeadersForAws = map[string]bool{"host": true,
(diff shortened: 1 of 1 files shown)


yogeshmundada:
LGTM

pkg/httpproxy/sign.go
```diff
@@ -101,9 +101,41 @@ func (a awsv4) sign(req *http.Request, secrets v1.SecretInterface, auth string)
...
+		service, region = partitionServiceAndRegion(partition, host)
+		// empty region is valid, but if one is found it should be assumed correct
+		if region != "" {
+			return service, region
+		}
+	}
+
+	// if no region is found, global endpoint is assumed. In this case us-east-1 should be used for signing:
+	// https://docs.aws.amazon.com/general/latest/gr/sigv4_elements.html
+	return service, "us-east-1"
```
yogeshmundada, line 116: nit: might want to convert this to something like 
DEFAULT_AWS_REGION = "us-east-1"

pkg/httpproxy/sign.go
```diff
@@ -101,9 +101,41 @@ func (a awsv4) sign(req *http.Request, secrets v1.SecretInterface, auth string)
...
+		}
+	}
+
+	if service == "" {
+		return "", ""
+	}
+
+	host = strings.Trim(host, service)
+	for _, part := range strings.Split(host, ".") {
+		if id := partition.Services()[service].Regions()[part].ID(); id != "" {
```
cbron, line 134: Same here, if there is a non-existent region in `part` the `ID` call would error.

pkg/httpproxy/sign.go
```diff
@@ -101,9 +101,41 @@ func (a awsv4) sign(req *http.Request, secrets v1.SecretInterface, auth string)
...
+		if id := partition.Services()[part].ID(); id != "" {
+			service = id
+			break
+		}
+	}
+
+	if service == "" {
+		return "", ""
+	}
+
```
cbron, line 131: Would be good to add comment below here explaining that you are trying to get the region. Not obvious on first glance.

pkg/httpproxy/sign.go
```diff
@@ -101,9 +101,41 @@ func (a awsv4) sign(req *http.Request, secrets v1.SecretInterface, auth string)
...
+	// if no region is found, global endpoint is assumed. In this case us-east-1 should be used for signing:
+	// https://docs.aws.amazon.com/general/latest/gr/sigv4_elements.html
+	return service, "us-east-1"
+}
+
+func partitionServiceAndRegion(partition endpoints.Partition, host string) (string, string) {
+	service := ""
+	for _, part := range strings.Split(host, ".") {
+		if id := partition.Services()[part].ID(); id != "" {
```
cbron, line 122: * If host string is incorrectly set, and `part` doesn't exist in the map string, then the `.ID()` call here would fail. Should be able to solve with a second check. 
* Would be better to set `Services()` method above so you don't have to call it every loop.

pkg/httpproxy/sign.go
rmweir: I believe the map would return an empty Service which would have an ID of "", so this should not error out.

pkg/httpproxy/sign.go
rmweir: Changing to use a constant but leaving the comment for clarity.

rmweir: Changing to use constant but leaving comment for clarity.

pkg/httpproxy/sign.go
cbron: You're right, double checked and its fine

PR #27819, opened 2020-07-02
Adding RBAC test cases

sowmyav27:
Member access : `"owner", "member", "read-only"`
Verify if users with roles are able to update mcapp and add users as members to mcapp

### DIFF
--- a/tests/validation/tests/v3_api/test_multi_cluster_app.py
+++ b/tests/validation/tests/v3_api/test_multi_cluster_app.py
@@ -1,20 +1,5 @@
-from .common import create_catalog_external_id
-from .common import create_project_and_ns
-from .common import get_admin_client
-from .common import get_defaut_question_answers
-from .common import get_project_client_for_token
-from .common import get_user_client
-from .common import get_user_client_and_cluster
 from .test_rke_cluster_provisioning import create_and_validate_custom_host
-from .common import random_test_name
-from .common import random_name
-from .common import USER_TOKEN
-from .common import validate_app_deletion
-from .common import validate_response_app_endpoint
-from .common import wait_for_app_to_active
-from .common import wait_for_catalog_active
-from .common import wait_for_mcapp_to_active
-from .common import delete_node
+from .common import *
 import pytest
 import time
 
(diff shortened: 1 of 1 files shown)


tests/validation/tests/v3_api/test_multi_cluster_app.py
```diff
@@ -602,3 +653,109 @@ def validate_answer_override(multiclusterapp, id,
...
+                    "read-only members cannot update multiclusterapp"
+    return mcapp
+
+
+def get_member_list(access, user):
+    member = dict()
+    member["accessType"] = access
+    member["displayName"] = user["username"]
+    member["displayType"] = user["type"]
+    member["displayType"] = "member"
```
jiaqiluo, line 726: the second line will overwrite the value so it will always be `member`, right?

tests/validation/tests/v3_api/test_multi_cluster_app.py
```diff
@@ -362,6 +347,70 @@ def test_multi_cluster_all_answer_override():
...
+    answer_values = get_defaut_question_answers(
+        admin_client,
+        MYSQL_EXTERNALID_131
+    )
+    mcapp = create_mcapp_with_member(targets, answer_values, member)
+    admin_client.reload(mcapp)
+    time.sleep(5)
+    admin_client.reload(mcapp)
+    # verfiy rbac CLUSTER_OWNER can edit/upgrade the mcapp
+    new_user, token = create_user(admin_client)
```
jiaqiluo, line 370: the `create_user` method creates a new local user, is it the right role we need here?

tests/validation/tests/v3_api/test_multi_cluster_app.py
```diff
@@ -362,6 +347,70 @@ def test_multi_cluster_all_answer_override():
     client.delete(mcapp)
 
 
+@if_test_rbac
+@pytest.mark.parametrize("role", ["owner", "member", "read-only"])
```
jiaqiluo, line 351: since `["owner", "member", "read-only"]` appears in multiple places, can we make it a variable?

Issue #27753, opened 2020-06-26
Updating a cluster using rke template revision with "enable_cluster_monitoring:false" disables monitoring


soumyalj:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create a rke template/ revision T1/R1 with **enable_cluster_monitoring:false**

2. Create a cluster using T1/R1. When a cluster is launched using this template revision, Rancher does NOT deploy the monitoring components.

3. After the cluster is active, enable monitoring on the cluster using UI.

4. Create another revision R2  for this RKE template T1, with **enable_cluster_monitoring:false**

5. Edit the cluster and update rke template revision to R2

**Result:**

Cluster gets updated, but  monitoring is disabled. 

Cluster update should succeed, but monitoring should not be disabled. enable_cluster_monitoring flag from the template should be ignored

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): **v2.4.5**
- Installation option (single install/HA): Single node

**Note: Test on v2.3.8**
 Tested the same steps on Rancher v2.3.8. The issue is not reproduced and monitoring is not disabled after step 5.

soumyalj:
Tested with v2.4-head(bf06dcf73)

Steps:
1. As a standard user, create a rke template/ revision T1/R1 with K8s version v1.16.10-rancher2-3  with `enable_cluster_monitoring:false` 
2. Create a cluster using T1/R1. When a cluster is launched using this template revision, Rancher does NOT deploy the monitoring components.
3. After the cluster is active, enable monitoring on the cluster using UI.
4. Create another revision R2 with K8s version v1.18.3-rancher2-3 for this RKE template T1, with `enable_cluster_monitoring:false`
5. Edit the cluster and update rke template revision to R2

Cluster gets updated to revision R2 and monitoring is still enabled as expected.

PR #27881, opened 2020-07-09
[2.4] Do not turn off monitoring/alerting from templateRevision during cluster updates

prachidamle:
This is a forwardport PR from release/v2.3: https://github.com/rancher/rancher/pull/24194

The fix is to ignore the template revision's "EnableClusterAlerting or EnableClusterMonitoring" flag when they are false during update cluster usecase for a cluster using a clusterTemplate.

This will allow the cluster updates (edit cluster to change overrides or update revision) to go through but keep monitoring/alerting intact.

https://github.com/rancher/rancher/issues/27753

### DIFF
--- a/tests/integration/suite/test_clustertemplate.py
+++ b/tests/integration/suite/test_clustertemplate.py
@@ -786,9 +786,11 @@ def test_cluster_desc_update(admin_mc, list_remove_resource):
     wait_for_cluster_to_be_deleted(client, cluster.id)
 
 
-def test_update_cluster_monitoring(admin_mc, remove_resource):
-    cluster_template = create_cluster_template(admin_mc,
-                                               remove_resource, [], admin_mc)
+def test_update_cluster_monitoring(admin_mc, list_remove_resource):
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_clustertemplate.py
```diff
@@ -786,6 +786,73 @@ def test_cluster_desc_update(admin_mc, list_remove_resource):
...
+            }
+        },
+        "enableClusterMonitoring": "false",
+        "defaultPodSecurityPolicyTemplateId": "restricted",
+    }
+    rev2 = client.create_cluster_template_revision(clusterConfig=cconfig2,
+                                                   name="v2",
+                                                   clusterTemplateId=tId,
+                                                   enabled="true")
+
```
mrajashree, line 834: looks like we don't have remove_resource for rev1 and rev2?

prachidamle: revisions will be deleted with the template due to owner-references

PR #27898, opened 2020-07-10
Enqueue CRTBs on RoleTemplate changes

ryansann:
Currently when a RoleTemplate changes in the management cluster, only the ProjectRoleTemplateBindings associated with that RoleTemplate are enqueued for processing. This needs to be extended to ClusterRoleTemplateBindings associated with the RoleTemplate that is changing. This fixes an issue where RoleBindings in the management cluster are not created when a RoleTempalte with cluster context is updated. Enqueueing the CRTBs triggers the CRTB Handler logic which in turn creates the required RoleBindings in the management cluster. 

Previous related PR: https://github.com/rancher/rancher/pull/22394

### DIFF
--- a/pkg/controllers/management/auth/role_template_lifecycle.go
+++ b/pkg/controllers/management/auth/role_template_lifecycle.go
@@ -15,25 +15,36 @@ import (
 const (
 	roleTemplateLifecycleName = "mgmt-auth-roletemplate-lifecycle"
 	prtbByRoleTemplateIndex   = "management.cattle.io/prtb-by-role-template"
+	crtbByRoleTemplateIndex   = "management.cattle.io/crtb-by-role-template"
 )
 
 type roleTemplateLifecycle struct {
 	prtbIndexer    cache.Indexer
 	prtbClient     v3.ProjectRoleTemplateBindingInterface
+	crtbIndexer    cache.Indexer
+	crtbClient     v3.ClusterRoleTemplateBindingInterface
 	clusters       v3.ClusterInterface
 	clusterManager *clustermanager.Manager
 }
 
 func newRoleTemplateLifecycle(management *config.ManagementContext, clusterManager *clustermanager.Manager) v3.RoleTemplateLifecycle {
-	informer := management.Management.ProjectRoleTemplateBindings("").Controller().Informer()
-	indexers := map[string]cache.IndexFunc{
+	prtbInformer := management.Management.ProjectRoleTemplateBindings("").Controller().Informer()
+	prtbIndexers := map[string]cache.IndexFunc{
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/auth/crtb_handler.go
```diff
@@ -49,20 +50,25 @@ type crtbLifecycle struct {
 }
 
 func (c *crtbLifecycle) Create(obj *v3.ClusterRoleTemplateBinding) (runtime.Object, error) {
+	logrus.Debugf("%s lifecycle: create", ctrbMGMTController)
```
cbron, line 53: Need this ? I would think we don't. Doesn't look like anything in this file needs to change, besides maybe the misspelling.

pkg/controllers/management/auth/role_template_lifecycle.go
```diff
@@ -44,13 +55,19 @@ func (rtl *roleTemplateLifecycle) Create(obj *v3.RoleTemplate) (runtime.Object,
 	if err := rtl.enqueuePrtbs(obj); err != nil {
 		return nil, err
 	}
+	if err := rtl.enqueueCrtbs(obj); err != nil {
```
cbron, line 58: Should be able to assert here if its project or cluster and enqueue relevant one: `if obj.Context == "project"`

pkg/controllers/management/auth/manager.go
```diff
@@ -436,6 +441,9 @@ func (m *manager) grantManagementPlanePrivileges(roleTemplateName string, resour
 		}
 	}
 
+	desiredBytes, _ := json.MarshalIndent(desiredRBs, "", "    ")
+	logrus.Debugf("desiredRBs: %v", string(desiredBytes))
```
cbron, line 445: same

pkg/controllers/management/auth/crtb_handler.go
ryansann: This is not needed, removing

pkg/controllers/management/auth/manager.go
ryansann: Looks like I committed some other changes I made. These other files were not supposed to get added, going to remove them

cbron:
Lgtm, would like some other reviewers

cjellick:
@ryansann (and @dramich) informed me of the question around roleTempalte.Context
```
 Question, roleTemplate has a context field, we don't enforce this being set but it's normally expected to be project or cluster. In the code we have a mix, some places don't care and just process anyways and some places are switching off that field and ignoring a roleTemplate without it set, what say you? Ignore if not set?
```

This field's original intent was to just be a UI hint, so I think it is ok to ignore if not et.

PR #27912, opened 2020-07-13
Add k3s airgap and initial rancher server setup to airgap automation

rancher-max:


tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -299,11 +390,89 @@ def prepare_airgap_node(bastion_node, number_of_nodes):
...
+    return results
+
+
+def deploy_airgap_k3s_cluster(bastion_node, ag_nodes):
+    token = ""
+    server_ip = ag_nodes[0].private_ip_address
+    for num, ag_node in enumerate(ag_nodes):
+        if num == 0:
+            # Install k3s server
+            install_k3s_command = \
```
ShylajaDevadiga, line 463: Could we change install_k3s_server instead?

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -299,11 +390,89 @@ def prepare_airgap_node(bastion_node, number_of_nodes):
...
+                'INSTALL_K3S_SKIP_DOWNLOAD=true ./install.sh && ' \
+                'sudo chmod 644 /etc/rancher/k3s/k3s.yaml'
+            run_command_on_airgap_node(bastion_node, ag_node,
+                                       install_k3s_command)
+            token_command = 'sudo cat /var/lib/rancher/k3s/server/node-token'
+            token = run_command_on_airgap_node(bastion_node, ag_node,
+                                               token_command)[0].strip()
+        else:
+            install_k3s_work = \
```
ShylajaDevadiga, line 472: install_k3s_worker to be consistent with variable names

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -72,6 +78,58 @@ def test_deploy_airgap_nodes():
...
+                                                  RANCHER_K3S_VERSION)
+    assert failures == [], "Failed to add images: {}".format(failures)
+    ag_nodes = prepare_airgap_k3s(bastion_node, NUMBER_OF_INSTANCES)
+    assert len(ag_nodes) == NUMBER_OF_INSTANCES
+
+    print(
+        '{} airgapped k3s instance(s) created.\n'
+        'Connect to these and run commands by connecting to bastion node, '
+        'then connecting to these:\n'
+        'ssh -i {}.pem {}@NODE_PRIVATE_IP'.format(
```
ShylajaDevadiga, line 100: Is  NODE_PRIVATE_IP is supposed to be a variable?

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -299,11 +390,89 @@ def prepare_airgap_node(bastion_node, number_of_nodes):
 
 
 def deploy_airgap_nodes(bastion_node, ag_nodes):
```
ShylajaDevadiga, line 392: What is the purpose of this method? Including comments above each method to describe the purpose would be helpful

tests/validation/tests/v3_api/test_airgap.py
rancher-max: No it's just supposed to print like that, so it shows up like:

```
Airgapped K3S Instance Details:
NAME: max-k3s-airgap-79619-0
PRIVATE IP: a.b.c.d

1 airgapped k3s instance(s) created.
Connect to these and run commands by connecting to bastion node, then connecting to these:
ssh -i thekey.pem ubuntu@NODE_PRIVATE_IP
```

tests/validation/tests/v3_api/test_airgap.py
rancher-max: Oops, I changed some things around with that function so that was supposed to get deleted

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -298,12 +412,87 @@ def prepare_airgap_node(bastion_node, number_of_nodes):
...
-                                                   AGENT_REG_CMD)
-        assert "Downloaded newer image for {}/rancher/rancher-agent".format(
-            bastion_node.host_name) in deploy_result[1]
+        if AGENT_REG_CMD.startswith("docker") or \
+                AGENT_REG_CMD.startswith("sudo docker"):
+            deploy_result = run_command_on_airgap_node(
+                bastion_node, ag_node, AGENT_REG_CMD)
+        else:
+            deploy_result = run_command_on_airgap_node(
+                bastion_node, ag_node, AGENT_REG_CMD)
```
jiaqiluo, line 471: Does `run_command_on_airgap_node` handles both cases?

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -295,13 +295,13 @@ def write_kubeconfig():
...
-    admin_client = rancher.Client(url=server_url + "/v3",
+def set_url_and_password(rancher_url, server_url=None):
+    admin_token = set_url_password_token(rancher_url, server_url)
+    admin_client = rancher.Client(url=rancher_url + "/v3",
                                   token=admin_token, verify=False)
-    auth_url = server_url + "/v3-public/localproviders/local?action=login"
+    auth_url = rancher_url + "/v3-public/localproviders/local?action=login"
     user, user_token = create_user(admin_client, auth_url)
-    env_details = "env.CATTLE_TEST_URL='" + server_url + "'\n"
+    env_details = "env.CATTLE_TEST_URL='" + rancher_url + "'\n"
```
jiaqiluo, line 304: in the case that `server_url` is NOT None, should we use `rancher_url` or `server_url`?

rancher-max: `rancher_url` should be used in these places as that is still where rancher is accessible and `.../v3` uri lives.

jiaqiluo: what is the difference between the `rancher_url` and the `server_url`?

rancher-max: `rancher_url` is the url used to access rancher -- the public IP of the rancher server.

`server_url` is the default rancher install url. All of the nodes in a cluster must be able to reach that url and is used as the access point to rancher when adding clusters/nodes. 

Generally these are the same, but in the case of airgap testing we have them different so that we have an exposed rancher_url without doing SOCKS proxies, but clusters are still able to be added because they look at the other (private) URL

Issue #23333, opened 2019-10-10
[ERROR] nodeDrain (v2.2.8)


LincolnBruce:
**What kind of request is this (question/bug/enhancement/feature request):**
Node Draining

**Steps to reproduce (least amount of steps as possible):**
执行Node驱散

**Result:**
`
2019/10/10 01:15:39 [ERROR] nodeDrain: [iZ2zefj6k9vb8bfodcfg0wZ] in cluster [c-24bcj] :  <nil>
2019/10/10 01:15:40 [INFO] Draining node iZ2zefj6k9vb8bfodcfg0wZ in c-24bcj
2019/10/10 01:15:40 [ERROR] nodeDrain: [iZ2zefj6k9vb8bfodcfg0wZ] in cluster [c-24bcj] :  <nil>
2019/10/10 01:15:40 [INFO] Draining node iZ2zefj6k9vb8bfodcfg0wZ in c-24bcj
2019/10/10 01:15:40 [ERROR] nodeDrain: [iZ2zefj6k9vb8bfodcfg0wZ] in cluster [c-24bcj] :  <nil>
2019/10/10 01:15:40 [INFO] Draining node iZ2zefj6k9vb8bfodcfg0wZ in c-24bcj
2019/10/10 01:15:40 [ERROR] nodeDrain: [iZ2zefj6k9vb8bfodcfg0wZ] in cluster [c-24bcj] :  <nil>
`

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
- Installation option (single install/HA):
Rancher v2.2.8 single install
![image](https://user-images.githubusercontent.com/7814560/66531871-ec36eb80-eb3f-11e9-86c8-1054274dc7f7.png)

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):
Cluster type: Custom
Machine type: aliyun cloud 
![image](https://user-images.githubusercontent.com/7814560/66531647-0d4b0c80-eb3f-11e9-8e13-161df84bea88.png)

```
(paste the output here)
```

Rancher Docker version (use `docker version`):

Client:
 Version:           18.09.0
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        4d60db4
 Built:             Wed Nov  7 00:49:01 2018
[...]

PR #27957, opened 2020-07-16
use v1.node name for draining node

kinarashah:
Forward port https://github.com/rancher/rancher/pull/27778 

#23333

### DIFF
--- a/pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go
+++ b/pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go
@@ -81,15 +81,28 @@ func (d *nodeDrain) drainNode(key string, obj *v3.Node) (runtime.Object, error) 
 			nodeMapLock.Unlock()
 			return nil, nil
 		}
-		ctx, cancel := context.WithCancel(d.ctx)
-		d.nodesToContext[obj.Name] = cancel
-		go d.drain(ctx, obj, cancel)
 		nodeMapLock.Unlock()
+
+		node, err := nodehelper.GetNodeForMachine(obj, d.nodeLister)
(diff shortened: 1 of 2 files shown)


pkg/controllers/user/nodesyncer/cordonfieldssyncer.go
```diff
@@ -75,21 +75,26 @@ func (d *nodeDrain) drainNode(key string, obj *v3.Node) (runtime.Object, error)
...
 		}
+		node, err := nodehelper.GetNodeForMachine(obj, d.nodeLister)
+		if err != nil {
+			return nil, err
+		}
+		if node == nil {
+			return nil, fmt.Errorf("nodeDrain: error finding node [%s]", obj.Spec.RequestedHostname)
+		}
 		ctx, cancel := context.WithCancel(d.ctx)
 		d.nodesToContext[obj.Name] = cancel
```
luthermonson, line 93: i know it's cleaner with the defer but wrapping the lock around the if/else means it locks the lister call which you dont need to do since the lock was only for protecting this map. i think you should go back to it being inside the if/else and move locking in the if under the lister.

maggieliu:
Master PR for: #23333

pkg/controllers/user/nodesyncer/cordonfieldssyncer.go
kinarashah: yep you're right, was trying to make it cleaner. but lister blocking might be an issue when multiple nodes are drained simultaneously if it takes a long time. updated!

will create a backport to update 2.4 once approved

luthermonson:
this version of the lock is getting into 2.4 too right?

Issue #27979, opened 2020-07-17
[KEv2] Rancher eks-operator integration


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
feature

**Expected:**
The following should be possible through the Rancher API:

Can provision an EKS cluster with the following configurable options:
  * Logging- can have any or none of the following: audit, scheduler, authenticator, controllerManager, api
  * Private access- false/true
  * Public access- false/true
  * Public Access Sources- can be empty or contain cider block such as 0.0.0.0/16
  * Kubernetes Version- string version i.e. 1.15
  * Tags- a map of string/string
  * Amazon Credential Secret (cloud crendential)
  * Display Name- this should correspond to the desired name in EKS
  * Subnets and security groups- either both of these should be provided or neither. If neither they will be generated
  * Service Role- this is the role that EKS will use to communicate with other parts of AWS. If empty it will be generated
  * Region- this should control which region the EKS cluster is provisioned in
  * Imported- This should be false and will be covered below
  * Node groups- array of objects with the follow options
    * GPU- false/true
    * Node group Name- name of the nodegroup. Should be unique
    * Disk Size- size of EBS volume, int, can be empty
    * InstanceType- ec2 instance to use, i.e. t3.medium
    * Labels- map of string/string
    * Ec2 Ssh key- name of the key pair in amazon to use, can be left blank in which case remote access to nodes will be disabled
    * Source Security Groups- only needed if ec2 ssh key is provided, provides security group for remote access (ssh) to worker nodes
    * Desired Size- how many nodes cluster should have
    * Min size- min nodes (not relevant until cluster auto scaler is installed)
    * Max nodes- max nodes (note relevant until cluster auto scaler is installed)
    * Subnets-subnets of worker nodes. If none provided then cluster's subnets will be used.
    * Tags- a map of string/string
    * Version- kubernetes version of the worker node. If left blank worker nodes will automatically keep up with control-plane's kubernetes version. If not blank must initially be the same as control plane, then as control plane upgrade, node group will stay at its version until incremented.

Of the above fields, the following can be updated:
  * Logging
  * Private Access
  * Public Access
  * Public Access Source
  * Kubernetes Version
  * Tags
  * Amazon Credential Secret
  * Node groups- can be added or deleted, the follow fields can be updated on individual node groups
  * Version

Can import EKS clusters- must provide following fields:
  * Display Name- name of cluster in EKS to import
  * Region- region that the cluster is in
[...]

PR #27981, opened 2020-07-17
Add EKSv2 to Rancher

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/27979


pkg/api/customization/cluster/validator.go
```diff
@@ -318,6 +324,41 @@ func (v *Validator) validateGenericEngineConfig(request *types.APIContext, spec
...
+	if !isNewImported && len(spec.EKSConfig.NodeGroups) == 0 {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("must have at least one node group"))
+	}
+
+	if request.Method != http.MethodPost {
+		return nil
+	}
+
+	allNetworkingFields := len(spec.EKSConfig.Subnets) == 0 && len(spec.EKSConfig.SecurityGroups) == 0
+	nonetworkingFields := len(spec.EKSConfig.Subnets) != 0 && len(spec.EKSConfig.SecurityGroups) != 0
```
cbron, line 352:
```suggestion
	noNetworkingFields := len(spec.EKSConfig.Subnets) != 0 && len(spec.EKSConfig.SecurityGroups) != 0
```

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"time"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	apimgmtv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
```
cbron, line 13: Only real changes in this file are the package rename, and getting it to skip any eksV2 ?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+	eksAPIGroup         = "eks.cattle.io"
+	eksV1               = "eks.cattle.io/v1"
+	managementV3        = "management.cattle.io/v3"
+	eksOperatorTemplate = "system-library-rancher-eks-operator"
+	eksOperator         = "rancher-eks-operator"
+	localCluster        = "local"
+	creatorAnno         = "field.cattle.io/creatorId"
+	importedAnno        = "eks.cattle.io/imported"
+	awsAccessKey        = "amazonec2credentialConfig-accessKey"
+	awsSecretKey        = "amazonec2credentialConfig-secretKey"
```
cbron, line 54: Can we pull some of these from more global const's ? Seems weird to redefine all as strings.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+	nsClient             corev1.NamespaceInterface
+	clusterClient        mgmtv3.ClusterInterface
+	systemAccountManager *systemaccount.Manager
+	dynamicClient        dynamic.NamespaceableResourceInterface
+}
+
+func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.ManagementContext, manager *clustermanager.Manager) {
+	dynamicClient, err := dynamic.NewForConfig(wContext.RestConfig)
+	if err != nil {
+		panic(fmt.Sprintf("[EKSClusterController Register] error creating dynamic client: %v", err))
```
cbron, line 73: Why panic ? Why not just error out ? Thats big impact.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+		if err != nil {
+			return cluster, err
+		}
+		eksClusterConfigDynamic = &unstructured.Unstructured{
+			Object: eksClusterConfigMap,
+		}
+		eksClusterConfigDynamic, err = e.dynamicClient.Namespace(globalNS).Create(context.TODO(), eksClusterConfigDynamic, v1.CreateOptions{})
+		if err != nil {
+			return cluster, err
+		}
```
cbron, line 163: Confused about this block, we create here and then update below ?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+			}
+
+			cluster = cluster.DeepCopy()
+			cluster.Status.APIEndpoint = endpoint
+			cluster.Status.CACert = ca
+			cluster.Status.ServiceAccountToken = saToken
+			return e.clusterClient.Update(cluster)
+		}
+
+		// check for unauthorized error
```
cbron, line 285: Why is this check so far below ? Because we just got the SA token/endpoint ?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+			}
+		}
+
+		if !apimgmtv3.ClusterConditionUpdated.IsTrue(cluster) {
+			cluster = cluster.DeepCopy()
+			apimgmtv3.ClusterConditionUpdated.True(cluster)
+			return e.clusterClient.Update(cluster)
+		}
+	case "updating":
+		failureMessage, _ := status["failureMessage"].(string)
```
cbron, line 303: Nit: this can go below since first conditional block doesn't use it.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+			}
+		}
+
+		if failureMessage == "" {
+			logrus.Infof("waiting for cluster EKS [%s] to update", cluster.Name)
+		} else {
+			logrus.Infof("waiting for cluster EKS [%s] update failure to be resolved", cluster.Name)
+		}
+
+		e.clusterEngqueueAfter(cluster.Name, 10*time.Second)
```
cbron, line 343: Typo `clusterEngqueueAfter`, should try to do a test where this actually gets called to ensure it works as expected.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,561 @@
...
+				return e.clusterClient.Update(cluster)
+			}
+		} else {
+			logrus.Infof("waiting for cluster EKS [%s] to create", cluster.Name)
+		}
+		e.clusterEngqueueAfter(cluster.Name, 20*time.Second)
+		return cluster, nil
+	}
+
+	eksClusterConfigMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&cluster.Spec.EKSConfig)
```
cbron, line 360: In what case does this section get hit ? Looks like the cases above all return.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: The EKS Cluster Config is created and then the rancher cluster is updated with the applied spec. It will then end up in the default case and be reenqueued.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: The logic above covers clusters that have just become active for the first time. This block could be relevant anytime as the eks-operator attempts to create a session prior to any update.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: that the EnqueueAfter function works? Neither controller would work if it didn't. When running the controller the logs can be observed on this time interval.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: active and updating cluster do not return by this point unless an error was encountered. Went ahead and remove some unnecessary returns after updates for readability.

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -725,8 +742,12 @@ func GetDriver(cluster *v3.Cluster, driverLister v3.KontainerDriverLister) (stri
 		}
 	}
 
+	if cluster.Spec.EKSConfig != nil {
+		return "EKS", nil
```
dramich, line 746: This should not be hardcoded and should live along with the other driver types in apimgmtv3

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -331,17 +339,26 @@ func (p *Provisioner) machineChanged(key string, machine *v3.Node) (runtime.Obje
...
+	if cluster.Spec.EKSConfig != nil {
+		return cluster, nil
+	}
+
 	var err error
 
+	// eksclustercontroller will oversee the creation of EKSv2 clusters
+	if cluster.Spec.EKSConfig != nil {
+		return cluster, nil
+	}
```
dramich, line 352: This is the same change twice

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -126,7 +130,11 @@ func (p *Provisioner) Remove(cluster *v3.Cluster) (runtime.Object, error) {
 }
 
 func (p *Provisioner) Updated(cluster *v3.Cluster) (runtime.Object, error) {
-	obj, err := v32.ClusterConditionUpdated.Do(cluster, func() (runtime.Object, error) {
+	if cluster.Spec.EKSConfig != nil {
```
dramich, line 133: I think adding a debug in all of these skips to indicate Provisioner will not be operating on the cluster will help debugging later

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: changed the ones with exiting global vars

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: If this fails during register they won't be able to provision EKS clusters.

pkg/controllers/management/clusterprovisioner/provisioner.go
rmweir: Correct

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,562 @@
...
+		cluster = cluster.DeepCopy()
+		cluster.Status.Driver = apimgmtv3.ClusterDriverEKS
+		var err error
+		cluster, err = e.clusterClient.Update(cluster)
+		if err != nil {
+			return cluster, err
+		}
+	}
+
+	eksClusterConfigDynamic, err := e.dynamicClient.Namespace(namespace.GlobalNamespace).Get(context.TODO(), cluster.Name, v1.GetOptions{})
```
paynejacob, line 122: TODO Context

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,562 @@
...
+					cluster.Status.EKSStatus.GeneratedSubnets = append(cluster.Status.EKSStatus.GeneratedSubnets, val.(string))
+				}
+				for _, val := range generatedSecurityGroups {
+					cluster.Status.EKSStatus.GeneratedSecurityGroups = append(cluster.Status.EKSStatus.GeneratedSecurityGroups, val.(string))
+				}
+				return e.clusterClient.Update(cluster)
+			}
+		}
+		logrus.Infof("waiting for EKS cluster [%s] to finish creating", cluster.Name)
+		e.clusterEnqueueAfter(cluster.Name, 20*time.Second)
```
paynejacob, line 206: 20s seems specific, if there is a reason can you add a comment

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,562 @@
...
+				}
+			}
+		}
+
+		if cluster.Spec.EKSConfig.Imported {
+			// cluster spec has been copied and cluster is no longer pending
+			if !apimgmtv3.ClusterConditionPending.IsTrue(cluster) {
+				cluster = cluster.DeepCopy()
+				apimgmtv3.ClusterConditionPending.True(cluster)
+				cluster, err = e.clusterClient.Update(cluster)
```
dramich, line 237: Missing an error check

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,562 @@
...
+			}
+		}
```
dramich, line 239: Can this whole block move up under the same if check above then we can do just one update call to the cluster

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: we use this all over to deal with client-go 1.18 requiring contexts when we don't have one to worth passing. Could probably hand one down from Register but that applies to everywhere.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+		ca := string(caSecret.Data["ca"])
+		saToken, err := generateSAToken(sess, cluster.Spec.EKSConfig.DisplayName, endpoint, ca)
+		if err != nil {
+			return cluster, err
+		}
+
+		cluster = cluster.DeepCopy()
+		cluster.Status.APIEndpoint = endpoint
+		cluster.Status.CACert = ca
+		cluster.Status.ServiceAccountToken = saToken
```
dramich, line 342: Is this like a secret?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+			return e.setUnknown(cluster, apimgmtv3.ClusterConditionUpdated, "")
+		case <-timeout.C:
+			return cluster, nil
+		}
+	}
+}
+func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+	// service account token and API endpoint are need to deploy cluster agent, should be able to retrieve
+	// from secret
+	for i := 1; i < 15; i++ {
```
dramich, line 314: use the kubernetes backoff, start at like .5 seconds and it can still max at 15*20 seconds.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+			if failureMessage != "" {
+				apimgmtv3.ClusterConditionUpdated.False(cluster)
+			}
+			cluster, err = e.clusterClient.Update(cluster)
+			if err != nil {
+				return cluster, fmt.Errorf("failure setting updating condition for cluster [%s]: %v", cluster.Name, err)
+			}
+		}
+
+		e.clusterEnqueueAfter(cluster.Name, 20*time.Second)
```
dramich, line 245: Can lower this to 5 seconds

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+	switch phase {
+	case "creating":
+		// set provisioning to unknown
+		cluster, err = e.setUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "")
+		if err != nil {
+			return cluster, err
+		}
+
+		logrus.Infof("waiting for EKS cluster [%s] to finish creating", cluster.Name)
+		e.clusterEnqueueAfter(cluster.Name, 20*time.Second)
```
dramich, line 174: Can lower to 5 seconds

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+		if cluster.Spec.EKSConfig.Imported {
+			cluster, err = e.setUnknown(cluster, apimgmtv3.ClusterConditionPending, "")
+			if err != nil {
+				return cluster, err
+			}
+			logrus.Infof("waiting for cluster import [%s] to start", cluster.Name)
+		} else {
+			logrus.Infof("waiting for cluster create [%s] to start", cluster.Name)
+		}
+		e.clusterEnqueueAfter(cluster.Name, 20*time.Second)
```
dramich, line 262: Can lower this 5 seconds

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+		cluster.Annotations = make(map[string]string)
+	}
+	cluster.Annotations[importedAnno] = "true"
+	if !apimgmtv3.ClusterConditionPending.IsTrue(cluster) {
+		apimgmtv3.ClusterConditionPending.True(cluster)
+	}
+	return e.clusterClient.Update(cluster)
+}
+
+func buildEKSCCObject(cluster *mgmtv3.Cluster) (*unstructured.Unstructured, error) {
```
dramich, line 368: Clarify the name of this function

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+
+			// this enqueue is necessary to ensure that the controller is reentered with the updating phase
+			e.clusterEnqueueAfter(cluster.Name, 10*time.Second)
+			return e.setUnknown(cluster, apimgmtv3.ClusterConditionUpdated, "")
+		case <-timeout.C:
+			return cluster, nil
+		}
+	}
+}
+func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
```
dramich, line 311: Missing space

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+	// service account token and API endpoint are need to deploy cluster agent, should be able to retrieve
```
dramich, line 312: Use this as the function description

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+		systemAccountManager: systemaccount.NewManager(mgmtCtx),
+		dynamicClient:        eksCCDynamicClient,
+	}
+
+	wContext.Mgmt.Cluster().OnChange(ctx, "eks-operator-controller", e.onClusterChange)
+}
+
+func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+	if cluster == nil || cluster.DeletionTimestamp != nil {
+		return nil, nil
```
dramich, line 98: Return cluster instead of nil

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+	list, err := e.dynamicClient.Namespace(namespace.GlobalNamespace).List(context.TODO(), v1.ListOptions{})
+	if err != nil {
+		return cluster, err
+	}
+	selector := fields.OneTermEqualSelector("metadata.name", cluster.Name)
+	w, err := e.dynamicClient.Namespace(namespace.GlobalNamespace).Watch(context.TODO(), v1.ListOptions{ResourceVersion: list.GetResourceVersion(), FieldSelector: selector.String()})
+	if err != nil {
+		return cluster, err
+	}
+	eksClusterConfigDynamic.Object["spec"] = oldSpec
```
dramich, line 279: Is this intended?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -0,0 +1,573 @@
...
+	}
+
+	eksClusterConfigMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&cluster.Spec.EKSConfig)
+	if err != nil {
+		return cluster, err
+	}
+
+	// check for changes between EKS spec on cluster and the EKS spec on the EKSClusterConfig object
+	if !reflect.DeepEqual(eksClusterConfigMap, eksClusterConfigDynamic.Object["spec"]) {
+		// if cluster is imported make sure it's original spec has been copied copied before mutating
```
dramich, line 155:
```suggestion
		// if cluster is imported make sure it's original spec has been copied before mutating
```

PR #28032, opened 2020-07-23
Rancher- rke2 integration 

MonzElmasry:
This pr allows rancher to detect and upgrade rke2 clusters.
- By exposing rke2 channels available from kdm/data.json with key= rke2
- Rke2 releases are available at v1-rke2-release/release
- K3s releases are changed to v1-k3s-release/release
-  bump channel server to v0.4.0 to be able to set a prefix to each listening port 
-  added cluster detection and upgrade functionality on import 
 
issue : https://github.com/rancher/rke2/issues/42


pkg/apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -74,6 +74,7 @@ const (
 	ClusterDriverRKE      = "rancherKubernetesEngine"
 	ClusterDriverK3s      = "k3s"
 	ClusterDriverK3os     = "k3os"
+	ClusterDriverRke2     = "rke2"
```
cbron, line 77: nit: `ClusterDriverRKE2`. Not a big deal, and/or can be done later.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -213,20 +213,21 @@ func (v *Validator) validateEnforcement(request *types.APIContext, data map[stri
...
 	}
 
-	if spec.K3sConfig == nil {
+	if spec.K3sConfig == nil && spec.Rke2Config == nil {
 		// only applies to k3s clusters
 		return nil
 	}
-
+	isK3s := spec.K3sConfig != nil
+	isrke2 := spec.Rke2Config != nil
```
cbron, line 228: nit: this check is done 3 lines above, you could move this higher

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -1008,15 +1010,26 @@ func k3sClusterConfig(cluster *v3.Cluster, nodes []*v3.Node) error {
...
+func setTypeConfig(config interface{}, version string) {
+	switch c := config.(type) {
+	case *v32.K3sConfig:
+		c.Version = version
+		c.SetStrategy(1, 1)
+	case *v32.Rke2Config:
+		c.Version = version
+		c.SetStrategy(1, 1)
+	}
+}
```
cbron, line 1035: Couldn't you just do this above: 
```
cluster.Spec.K3sConfig = &v32.K3sConfig{
  Version: cluster.Status.Version.String(),
}
cluster.Spec.K3sConfig.SetStrategy(1,1)
```

Seems simpler. Passing a pointer to an interface and then switching feels weird. Actually I liked the old way best, without the `SetStrategy` setter method where you just hardcoded everything inline. Less boilerplate code and passing pointers around.

pkg/controllers/management/k3sbaseupgrade/deployPlans.go
```diff
@@ -59,13 +79,12 @@ func (h *handler) deployPlans(cluster *v3.Cluster) error {
 
 		} else {
 
-			switch name := plan.Name; name {
-			case k3sMasterPlanName:
+			if name := plan.Name; strings.Contains(name, "master-plan") {
```
cbron, line 82: Keep case switch and just add new case: 
`case k3sMasterPlanName, rke2MasterPlanName:`
Less brittle that way, if someone adds a third and doesn't want to use this "master-plan" sub-string.

pkg/controllers/management/k3sbaseupgrade/k3sbase_upgrade_handler.go
```diff
@@ -133,7 +150,7 @@ func (h *handler) deployK3sUpgradeController(clusterName string) error {
 				},
 			},
 			Spec: v33.AppSpec{
-				Description:     "Upgrade controller for k3s clusters",
+				Description:     "Upgrade controller for k3s base clusters",
```
cbron, line 153: Nit: I think I would have preferred the phrase `k3s based clusters`, as then its more obvious `based` is referring to k3s and not the cluster. Another example is that `k3sBaseUpgrade` sounds like you're doing a base upgrade of k3s, not that its a k3s-based cluster upgrade. Not sure we need to rewrite everything, but I think in this description line you should use `based`.

pkg/apis/management.cattle.io/v3/cluster_types.go
MonzElmasry: used it the detection process, should mark rke2 clusters.

PR #28107, opened 2020-07-28
Adding test for the new hosted EKS cluster creation

sowmyav27:
Currently creates a 1 node EKS cluster.

### DIFF
--- a/tests/validation/tests/v3_api/test_hosted_eks_cluster.py
+++ b/tests/validation/tests/v3_api/test_hosted_eks_cluster.py
@@ -1,0 +1,144 @@
+import os
+from .common import get_user_client, validate_cluster, \
+    random_test_name, AWS_SSH_KEY_NAME
+import pytest
+
+EKS_ACCESS_KEY = os.environ.get('RANCHER_EKS_ACCESS_KEY', "")
+EKS_SECRET_KEY = os.environ.get('RANCHER_EKS_SECRET_KEY', "")
+EKS_REGION = os.environ.get('RANCHER_EKS_REGION', "us-east-2")
+EKS_K8S_VERSION = os.environ.get('RANCHER_EKS_K8S_VERSION', "1.17")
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_hosted_eks_cluster.py
```diff
@@ -0,0 +1,89 @@
+from .common import *  # NOQA
```
bmdepesa, line 1: Can you use specific imports instead of importing all?

tests/validation/tests/v3_api/test_hosted_eks_cluster.py
```diff
@@ -0,0 +1,91 @@
...
+        "subnets": [],
+        "tags": {},
+        "type": "eksclusterconfigspec",
+        "nodeGroups":[{
+            "desiredSize": 1,
+            "diskSize": 20,
+            "gpu": False,
+            "instanceType": "m5.large",
+            "maxSize": 1,
+            "minSize": 1,
```
sangeethah, line 72: @sowmyav27 Can we have the node group size set to 2 ,so that inter host networking gets tested as part of validate_cluster() call.
Also can we validate that we did get 2 worker nodes created as part of the EKS cluster ?

tests/validation/tests/v3_api/test_hosted_eks_cluster.py
```diff
@@ -0,0 +1,191 @@
...
+    return logging_types
+
+
+def validate_eks_cluster(client, cluster):
+    time.sleep(5)
+    cluster = wait_for_condition(
+        client, cluster,
+        lambda x: x.state == "active",
+        lambda x: 'State is: ' + x.state,
+        timeout=DEFAULT_TIMEOUT_EKS)
```
sangeethah, line 155: @sowmyav27 Can we try to use the validate_cluster() in common.py ?

Issue #28120, opened 2020-07-28
Review Default k8s versions for hosted k8s providers (EKS, GKE, AKS)


maggieliu:
Review the default k8s versions for all hosted k8s providers and update the versions to make sure we support the latest one.

DoNotSayYes:
**Version to test:** `rancher/rancher2.4-head` off 2019-07-29

**Steps to test:**
1. Add an Amazon EKS cluster on rancher
2. Choose `cn-north-1` region and configure the credentials
3. Check Kubernetes version

**Results:**
On the configuration page, Rancher UI just provides `1.15` as the latest Kubernetes version, however, the latest available version on Amazon is `1.17`, users can update the Kubernetes version on Amazon and it can be synchronized to Rancher
![image](https://user-images.githubusercontent.com/57614736/88776032-0964a580-d1b8-11ea-86b6-db93276d3f31.png)
![image](https://user-images.githubusercontent.com/57614736/88777742-27330a00-d1ba-11ea-922e-0fe07a7e52ae.png)

soumyalj:
**Validated on v2.4-head (fdbd0245a)** 

- Verified that K8s versions  **v1.17, v1.16 and v1.15** are displayed when creating EKS cluster as below:

![Screen Shot 2020-07-30 at 8 38 34 PM](https://user-images.githubusercontent.com/18536626/88997486-b1f03380-d2a4-11ea-91e3-eeddb9ecbf2b.png)

- Created EKS clusters with K8s versions v1.17, v1.16 and v1.15 in different regions. Clusters provisioned successfully and were functional

![Screen Shot 2020-07-30 at 8 41 19 PM](https://user-images.githubusercontent.com/18536626/88997687-1612f780-d2a5-11ea-8488-1fbc600849ac.png)

DoNotSayYes:
**Validated on v2.4-head** ([`335be4b`](https://github.com/rancher/rancher/commit/335be4bf046d80b933c591ee7aa098499c277b3b))

**Validation steps**

1. Add an Amazon EKS cluster on rancher
2. Choose cn-north-1 region and configure the credentials
3. Check Kubernetes version

**Results:**
Clusters can be provisioned successfully and functional by using any of the available kubernetes versions.
![image](https://user-images.githubusercontent.com/57614736/89246930-c46cc300-d63e-11ea-9257-079dffc77153.png)
[...]

PR #28123, opened 2020-07-28
Update EKS AMIs

rmweir:
Reference: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html


maggieliu:
related to: https://github.com/rancher/rancher/issues/28120

PR #28161, opened 2020-07-30
Add KMS endpoint to accepted listed

rmweir:
**Problem:**
Cannot make request in UI to amazon KMS.

**Solution:**
Add the service to the accepted domain list.

**Issue:**
https://github.com/rancher/rancher/issues/28141


vincent99:
Is there no kms in China?

rmweir:
> Is there no kms in China?

good catch

Issue #28031, opened 2020-07-23
Minio deployment in the pipeline fails to get to active state


soumyalj:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a DO cluster
2. Enable pipeline in the cluster using default example repository: https://github.com/rancher/pipeline-example-go.git
3. Run the pipeline

**Result:**
The minio deployment fails to get to active state
![Screen Shot 2020-07-22 at 5 10 11 PM](https://user-images.githubusercontent.com/18536626/88241212-5091ea00-cc3e-11ea-80db-52788f3432b6.png)

![Screen Shot 2020-07-22 at 5 06 42 PM](https://user-images.githubusercontent.com/18536626/88241261-77502080-cc3e-11ea-957b-6cd10a1a85e4.png)

**Other details that may be helpful:**
Sometimes it gets to active state, but gets into crashloop backoff state after sometime

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head(f7e83e3)
- Installation option (single install/HA): Single node

**Cluster information**

- Kubernetes version (use `kubectl version`): V1.18.6-rancher1-2

gitlawr:
The Minio pod is OOMKilled. This newer version of Minio may exceed the current memory limit on bootstrap.

gitlawr:
Also noted we are not able to pull images from the embedded docker registry on master, which is tracked in https://github.com/rancher/rancher/issues/28167

soumyalj:
Validated with master-head(3d7c3b3c1)

Steps:
1. Create a DO cluster
2. Enable pipeline in the cluster using default example repository: https://github.com/rancher/pipeline-example-go.git
3. Run the pipeline

Minio deployment is active :

![Screen Shot 2020-08-04 at 11 21 20 AM](https://user-images.githubusercontent.com/18536626/89330134-ac983d80-d644-11ea-9505-55d10c89cf2b.png)

- Verified that the memory limit for minio deployment was updated to 500M

![Screen Shot 2020-08-04 at 11 19 27 AM](https://user-images.githubusercontent.com/18536626/89330014-7b1f7200-d644-11ea-8997-71504ec73ad2.png)


kubectl 
[...]

PR #28166, opened 2020-07-30
Bump memory limit for pipeline Minio.

gitlawr:
Issue: https://github.com/rancher/rancher/issues/28031

Problem:
Newer version Minio consumes more memory. It may exceed the current limit.

Solution:
Bump memory limit from 200M to 500M and handle upgrade.


dramich:
Should this be user configurable with a default value of this updated value?

gitlawr:
This is the default. Users can customize it if they want.

Issue #27011, opened 2020-05-09
"TestAndApply Error" when setting up GitHub application for pipelines


PBXg33k:
Just deployed a new rancher environment on a new kubernetes environment and was in the process of setting up Rancher to automagically deploy applications from my public & private repositories.

In order for pipelines to work i registered a GitHub application for rancher to use and provided Rancher with the Client ID & Client Secret. The OAuth process went smoothly untill i was sent back to the callback uri (rancher).

Everytime i try to setup a GitHub application in Rancher i am presented with the following error:
`"failed to decode the config, error: 1 error(s) decoding:↵↵* 'ManagedFields[0].Time' expected a map, got 'string'"`

---
| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.4.3` UI: `v2.4.22` |
|Route|`/p/local:p-v4zkc/pipeline?group=none`|
|Installation method|Helm rancher|
|Cluster Type|Custom|
|Machine Type|VPS (Hetzner Cloud)|
|Kubernetes Version|v1.18.2|

masteredward:
Same problem here! Rancher 2.4.3, Helm Install K3s 1.18.2.

houtianhong:
Rancher 2.4.3 has the same problem when adding the authorization of private gitlab-ce account. Click the authorization button and the error will pop up immediately:
failed to decode the config, error: 1 error(s) decoding: * 'ManagedFields[0].Time' expected a map, got 'string'

icanhazbeer:
Also ran into this issue running K3S 1.18.2

| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.4.3` UI: `v2.4.22` |

icanhazbeer:
We just tested with a reinstall of K3S/Rancher using K3S 1.17.5 and has **_success_** with this feature:

| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.4.3` UI: `v2.4.22` |
|Kubernetes Version in k3s| `**1.17.5**` |

What broke in the 1.18 version?

raynic35:
Same problem
Useful | Info
-- | --
Versions | Rancher v2.4.4
Kubernetes Version in k3s | 1.18.3, v1.17.6+k3s1
Any news on that?

fbrbovic:
Same problem on my end as well, any update on this?

khatrig:
rancher 2.4.5 does not work as well

sammasamadhi:
[...]

PR #28169, opened 2020-07-30
[2.4] Fix unstructured decoded with managedfields

gitlawr:
Issue: https://github.com/rancher/rancher/issues/27011

Problem:
Decoding unstructured pipeline provider configs fails in k8s 1.18+ with managedFields introduced.

Solution:
Decode time fields properly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -123,4 +123,5 @@ require (
 	k8s.io/kubectl v0.18.0
 	k8s.io/kubernetes v1.18.0
 	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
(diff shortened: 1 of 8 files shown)


YaoC:
I dont think deleting all fields containing `time.Time` solve the problem once for all. The bug may occur again when `ObjectMeta` changing in the future. `mapstructure` provides a  [`DecodeHookFunc`](https://godoc.org/github.com/mitchellh/mapstructure#DecodeHookFunc) interface for transforming data. Here is a example mitchellh/mapstructure#159.

gitlawr:
@YaoC Make sense. Thanks for the advice.

PR #28152, opened 2020-07-29
Validation - Added test to create an EKS cluster using boto API

bmdepesa:
Requires: 
```
RANCHER_EKS_ACCESS_KEY
RANCHER_EKS_SECRET_KEY
RANCHER_EKS_REGION
RANCHER_EKS_K8S_VERSION
AWS_SUBNET
RANCHER_EKS_ROLE_ARN
RANCHER_EKS_WORKER_ROLE_ARN
RANCHER_CLUSTER_NAME
```

To use the generated kubeconfig locally requires `aws` cli installed and configured w/ access key & secret key used to create the cluster

PR #28180, opened 2020-07-30
Rancher local HA installed in EKS cluster

izaac:
Based on https://github.com/rancher/rancher/pull/28152

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -1,4 +1,5 @@
 from .common import *  # NOQA
+from .test_boto_create_eks import get_eks_kubeconfig
 from .test_import_k3s_cluster import create_multiple_control_cluster
 from .test_rke_cluster_provisioning import rke_config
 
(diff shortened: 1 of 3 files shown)


tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -201,6 +224,23 @@ def install_cert_manager():
...
+def set_route53_with_eks_ingress():
+    kubectl_ingress = "kubectl get ingress -n cattle-system -o " \
+                      "jsonpath=\"" \
+                      "{.items[0].status.loadBalancer.ingress[0].hostname}\""
+    ingress_address = run_command_with_stderr(export_cmd
+                                              + " && " +
+                                              kubectl_ingress).decode()
+    AmazonWebServices().upsert_route_53_record_cname(RANCHER_HA_HOSTNAME,
+                                                     ingress_address)
+    time.sleep(20)
```
sangeethah, line 241: @izaac We should wait for the DNS resolution to succeed instead of a sleep of 20 sec. In the last couple of runs we have done , seems like 20 seconds is not enough.

PR #28187, opened 2020-07-31
Remove duplicated configure for cluster logging

aiwantaozi:
Problem:
PR for collect containerd logs bring in dulicated configure cause dry-run return warning

Solution:
remove duplicated configure

Issue:

### DIFF
--- a/pkg/controllers/managementuser/logging/generator/templatesource.go
+++ b/pkg/controllers/managementuser/logging/generator/templatesource.go
@@ -17,9 +17,7 @@ var SourceTemplate = `
   @type  tail
   path  /var/log/containers/*.log
   pos_file  /fluentd/log/{{ .ContainerLogPosFilename}}
-  time_format  %Y-%m-%dT%H:%M:%S.%N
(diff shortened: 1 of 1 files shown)


cbron:
This removal won’t affect anything because it falls into the multi-format ? Both of those have timezone at the end so what happens if time stamp doesn’t have timezone, does it still match ?

aiwantaozi:
Sorry for reply late, docker json-file logs and containerd logs format are different, but the time both field generate with RFC 3339Nano timestamp, related [doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/kubelet-cri-logging.md). Fluentd use [strptime](https://docs.ruby-lang.org/en/2.4.0/Time.html#method-c-strptime) to parse time_format, fluentd can handle the timezone 

Here is a similar [PR](https://github.com/kubernetes/kubernetes/pull/54777/files) in Kubernetes repo.

cbron:
Need to backport to 2.4 also

PR #28190, opened 2020-07-31
add the support for validating rbac for cluster roles

jiaqiluo:
This PR contains the utilities for testing rabc of the newly-introduced cluster roles by Monitoring, Istio, and Logging.  
Instead of running against the Rancher server as we did before, we run tests against the k8s' native API directly. 

Basic logic of the tests:
- create a new service account, and generate a kubeconfig file from it.
- run kubectl command using the newly-created kubeconfig, and see Forbidden error in the output 
- bind the service account with the cluster role
- run the same command, and it should work now, ie. no previous error in the output

one new parameter is introduced:
- RANCHER_CONTROL_PLANE_IP --> this is used as the server IP in the kubeconfig file for the service account 

How to use:
- a test method for Pytest to recognize and run. see `tests/validation/tests/v3_api/test_rbac_2.py` as an example 
- ~a JSON file contains test cases in the format of  `cluster roles + kubectl commands to run + expected result`. see `tests/validation/tests/v3_api/resource/rbac/monitoring/monitoring_rbac.json` as an example~
- a JSON file contains test cases for each clusterrole. see `tests/validation/tests/v3_api/resource/rbac/monitoring/monitoring_rbac.json` as an example
- some supporting files which are used in commands like `kubectl create -f `, `kubectl apply -f `, and `kubectl delete -f `

Update on 8/3:

some changes to address @aiyengar2  comments:
- simplified the formate of content in the `monitoring_rbac.json` file
- rename the CRDs' yaml file 
- added test cases to cover `create`, `update`, `list`, `get`, and  `delete`


tests/validation/tests/v3_api/test_rbac_2.py
```diff
@@ -0,0 +1,16 @@
+from .common import *  # NOQA
```
bmdepesa, line 1: Can we use specific imports instead of importing all?

tests/validation/tests/v3_api/resource/rbac/monitoring/crd-prometheus.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: prometheus
+spec:
+  serviceAccountName: prometheus
+  serviceMonitorSelector:
+    matchLabels:
+      team: frontend
+  resources:
+    requests:
+      memory: 400Mi
+  enableAdminAPI: true
```
izaac, line 13: nit: New line

tests/validation/tests/v3_api/test_rbac_2.py
```diff
@@ -0,0 +1,23 @@
...
+    create_kubeconfig(cluster)
+
+
+@pytest.mark.parametrize("cluster_role, command, authorization",
+                         test_reader(os.path.join(
+                             DATA_SUBDIR,
+                             'rbac/monitoring/monitoring_rbac.json')))
+def test_monitoring_v2_rbac(cluster_role, command, authorization):
+    validate_cluster_role_rbac(cluster_role, command, authorization)
+    return
```
izaac, line 23: Why do we need a return here?

tests/validation/tests/v3_api/resource/rbac/monitoring/crd-alert-manager.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+metadata:
+  name: crd-alert-manager
+spec:
+  replicas: 1
```
izaac, line 6: nit: New line

tests/validation/tests/v3_api/test_rbac_2.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+import os
+from .common import \
+    create_kubeconfig, \
+    DATA_SUBDIR, \
```
izaac, line 5: I prefer to have `from .common import x` for each import instead of breaking one line in multiple.

All caps variables go first too in the `.common` group.

tests/validation/tests/v3_api/resource/rbac/monitoring/monitoring_rbac.json
```diff
@@ -0,0 +1,261 @@
...
+      "command": "get prometheuses.monitoring.coreos.com",
+      "authorization": true
+    },
+    {
+      "command": "delete -f {resource_root}/rbac/monitoring/crd-prometheus.yaml",
+      "authorization": false
+    }
+  ]
+}
+
```
izaac, line 261: Data Driven I like it.

tests/validation/tests/v3_api/common.py
```diff
@@ -2735,3 +2735,111 @@ def get_node_details(cluster, client):
...
+    with open(file_path) as reader:
+        test_cases = json.loads(reader.read().replace("{resource_root}",
+                                                      DATA_SUBDIR))
+    output = []
+    for cluster_role, checks in test_cases.items():
+        for check in checks:
+            output.append((cluster_role,
+                           check['command'],
+                           check['authorization']))
+    return output
```
aiyengar2, line 2758: In general for structuring the test cases input, it might be easier to have something with the following format:
```
roles: [
    {
        name: "monitoring-admin" # role name
        apply_before: [] # any test cases to apply with cluster-admin permissions first in case resources should already exist before other test cases are run
        should_pass: ["create-configmap", "get-configmap", ...] # test cases that need to pass
        should_fail: [] # test cases that should fail
    },
    {
        name: "monitoring-edit"
        apply_before: []
        should_pass: ["create-configmap", "get-configmap", ...]
        should_fail: []
    },
    {
        # What we expect from monitoring-view before an admin creates resources in a cluster
        name: "monitoring-view"
        apply_before: []
        should_pass: []
        should_fail: ["create-configmap", ...]
    },
    {
        # What we expect from monitoring-view after an admin creates resources in a cluster
        name: "monitoring-view"
        apply_before: ["create-configmap", ...]
        should_pass: ["get-configmap", ...]
        should_fail: []
    }
],
test_cases: {
    "create-configmap": {
        command: "create configmap my-config --from-literal=key1=config1"
    },
    "get-configmap": {
        command: "get configmap"
    },
    ....
}
```

Within `test_reader`, you can then convert each role entry into the corresponding:
- `cluster_role` (`roles[i].name`)
- `command` (`test_cases[roles[i].should_pass[j]` or `test_cases[roles[i].should_fail[j]`)
- `authorization` (whether you are iterating through test cases in `should_pass` or `should_fail`)
- `test_state` (list of `test_cases[roles[i].apply_before[j]]`)

The current structure ends up repeating the same underlying test cases with just a different `authorization` flag, which makes it harder to catch if minor changes will be necessary in the future. I'm also not sure how you would be able to test getting a specific resource (which tests the GET permissions as opposed to LIST permissions) without having some initial cluster state that needs to exist for view permissions.

tests/validation/tests/v3_api/resource/rbac/monitoring/crd-service-monitor.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: monitoring.coreos.com/v1
```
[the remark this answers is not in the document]
jiaqiluo: the resource filename is updated to the format of `cr-<resource>.yaml`

tests/validation/tests/v3_api/resource/rbac/monitoring/monitoring_rbac.json
```diff
@@ -0,0 +1,207 @@
...
+      "get podmonitors.monitoring.coreos.com",
+      "get podmonitors.monitoring.coreos.com test-example-1",
+      "get prometheusrules.monitoring.coreos.com",
+      "get prometheusrules.monitoring.coreos.com test-example-1",
+      "get servicemonitors.monitoring.coreos.com",
+      "get servicemonitors.monitoring.coreos.com test-example-1",
+      "get alertmanagers.monitoring.coreos.com",
+      "get alertmanagers.monitoring.coreos.com test-example-1",
+      "get prometheuses.monitoring.coreos.com",
+      "get prometheuses.monitoring.coreos.com test-example-1"
```
aiyengar2, line 99: Each role's suite of test cases seems to rely on the previous role having successfully passed each test case in order to return the correct results.

For example:
1. `monitoring-view` tests rely on the tester having previously (and successfully) ran either `monitoring-admin` or `monitoring-edit` tests to create the resources in the first place
2. `monitoring-edit` relies on `monitoring-admin` being able to successfully delete resources so that its create doesn't fail because the resource already exists (even if they do have the permissions to create).

If there is a requirement for these roles to be independently testable (e.g. wanting to run tests for roles in parallel or individually), there should be another field to the test cases that runs a set of commands to add or remove resources from cluster before running the tests for a given role.

Not sure if this is something that needs to be addressed, feel free to resolve this comment if this requirement is not necessary.

tests/validation/tests/v3_api/resource/rbac/monitoring/monitoring_rbac.json
```diff
@@ -0,0 +1,207 @@
...
+      "delete -f {resource_root}/rbac/monitoring/cr_prometheus_rule_upgrade.yaml",
+      "create -f {resource_root}/rbac/monitoring/cr_service_monitor_base.yaml",
+      "get servicemonitors.monitoring.coreos.com",
+      "get servicemonitors.monitoring.coreos.com test-example-1",
+      "apply  -f {resource_root}/rbac/monitoring/cr_service_monitor_upgrade.yaml",
+      "delete -f {resource_root}/rbac/monitoring/cr_service_monitor_upgrade.yaml",
+      "get alertmanagers.monitoring.coreos.com",
+      "get alertmanagers.monitoring.coreos.com test-example-1",
+      "get prometheuses.monitoring.coreos.com",
+      "get prometheuses.monitoring.coreos.com test-example-1"
```
aiyengar2, line 32: When are these resources created within the cluster?

tests/validation/tests/v3_api/resource/rbac/monitoring/monitoring_rbac.json
jiaqiluo: We are more focused on validating if the cluster role gets the right set of permissions, rather than if the object is created/updated/got/listed/deleted with target values. That is also why we check the particular `Error from server (Forbidden)` error in the response of API (kubectl command) calls.

tests/validation/tests/v3_api/common.py
```diff
@@ -2735,3 +2735,113 @@ def get_node_details(cluster, client):
...
+        if name in item:
+            secret_name = item.split("/")[1]
+            break
+    res = execute_kubectl_cmd(cmd="get secret {}".format(secret_name))
+    ca = res["data"]["ca.crt"]
+    token = res["data"]["token"]
+    token = base64.b64decode(token).decode()
+
+    # get the server ip, which is the control plane's ip
+    ip = os.environ.get('RANCHER_CONTROL_PLANE_IP', "None")
```
izaac, line 2759: Can we get this info from the `cluster` object? This is to be able to go end-to-end in Jenkins:

- Create cluster
- Get CP info
- Run tests
- Results
- Teardown

tests/validation/tests/v3_api/common.py
```diff
@@ -2735,3 +2735,113 @@ def get_node_details(cluster, client):
...
+    # create the cluster role binding if it does not exist
+    result = execute_kubectl_cmd(cmd='get clusterrolebinding {}'.format(name),
+                                 json_out=False,
+                                 stderr=True).decode('utf_8')
+    if "NotFound" in result:
+        cmd = "create clusterrolebinding {} " \
+              "--clusterrole {} " \
+              "--serviceaccount {}".format(name, cluster_role,
+                                           "default:" + name)
+        execute_kubectl_cmd(cmd, json_out=False)
```
sangeethah, line 2835: @jiaqiluo Why cant we do the clusterrolebindings along with service account creation ?

jiaqiluo: you are right. 
This is moved out.

tests/validation/tests/v3_api/common.py
jiaqiluo: good point. 
Now we get the ip from the control plane node.

sangeethah:
LGTM

Issue #28162, opened 2020-07-30
Cluster stuck in Waiting state


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create a hosted EKS cluster from Rancher
- Enter the following in the cluster

```
publicAccessSources: ["0.0.0.0/16","1.1.1.1/32"]
"privateAccess": true,
"publicAccess": true
```
- Other values in the cluster are default. Add a nodegroup. And save the cluster
- The cluster is seen in Active EKS console the Access values are seen set in the EKS console on the cluster.
- Options are seen set on View in API of the cluster
- The cluster is seen in `Waiting`: `Waiting for API to be available` state in Rancher
- eks-controller logs:
```
time="2020-07-30T00:24:08Z" level=info msg="waiting for cluster [c-hh5hb] to update nodegroups [sow-new]"
time="2020-07-30T00:24:39Z" level=info msg="waiting for cluster [c-hh5hb] to update nodegroups [sow-new]"
time="2020-07-30T00:25:09Z" level=info msg="cluster [c-hh5hb] finished updating"
```

**Expected Result:**
Cluster should be in Active state

**Note:**
When only the following values are set, cluster comes ups Active (publicAccessSources is NOT set)
```
"privateAccess": true,
"publicAccess": true
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `6d6199121`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Hosted EKS
- Kubernetes version (use `kubectl version`):

```
1.17
```

sowmyav27:
Verified on master-head - commit id: ``

- Create a hosted EKS cluster from Rancher
- Enter the following in the cluster

```
publicAccessSources: ["0.0.0.0/16","1.1.1.1/32"]
"privateAccess": true,
"publicAccess": true
```
- publicAccessSources CIDRs are incorrect (rancher will not be able to connect to the eks cluster)
- Other values in the cluster are default. Add a nodegroup. And save the cluster
- The cluster is seen in Active EKS console the Access values are seen set in the EKS console on the cluster.
- Cluster is seen in `Waiting` state with an error 
[...]

PR #28209, opened 2020-08-04
Set waiting to false if SA generation fails

rmweir:
**Problem:**
Prior, cluster would appear to be in waiting state without error despite service account generation failing.

**Solution:**
Now, waiting condition will be set to false with an error message set.

**Issue:**
https://github.com/rancher/rancher/issues/28162

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -219,6 +219,12 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		if cluster.Status.ServiceAccountToken == "" || cluster.Status.APIEndpoint == "" {
 			cluster, err = e.generateAndSetServiceAccount(cluster)
 			if err != nil {
+				var statusErr error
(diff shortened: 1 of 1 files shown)


brendarearden (dismissed):

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -219,6 +219,12 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		if cluster.Status.ServiceAccountToken == "" || cluster.Status.APIEndpoint == "" {
 			cluster, err = e.generateAndSetServiceAccount(cluster)
 			if err != nil {
+				var statusErr error
+				cluster, statusErr = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting,
+					fmt.Sprintf("failed to generate service account, check networking settings: %v", err))
```
dramich, line 224:
```suggestion
					fmt.Sprintf("failed to generate service account: %v", err))
```
This should be more general, the call could be failing for any number of reasons, auth, networking outage. The error message coming back from `generateAndSetServiceAccount` should hopefully be useful enough to troubleshoot.

dramich (approved):

PR #28220, opened 2020-08-04
Remove duplicated configure for cluster logging

paynejacob:
Problem:
PR for collect containerd logs bring in dulicated configure cause dry-run return warning

Solution:
remove duplicated configure


cbron:
backport for issue https://github.com/rancher/rancher/issues/26543

PR #28227, opened 2020-08-04
Automation tests for Zero Down Time 

tiffanytfmao:


tests/validation/tests/v3_api/test_zero_downtime.py
```diff
@@ -0,0 +1,328 @@
...
+    wait_for_node_upgrade(nodes, p_client, cluster, workload, node_ver)
+    # Validate update has went through
+    assert len(client.list_node(clusterId=cluster.id).data) == original + 1
+    for node in nodes:
+        node = client.reload(node)
+        assert node["info"]["kubernetes"]["kubeletVersion"] == node_ver
+
+
+def validate_node_state(nodes, workload, drain=False, timeout=600,
+                        max_unavailableWorker=1):
```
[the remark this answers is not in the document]
kinarashah: the default is set to in env var 10% but this is where we need the actual count of worker nodes. maybe we can remove the default altogether since it's a helper func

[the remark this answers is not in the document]
tiffanytfmao: there is a default since I combined both the cordon and drain function, only worker nodes/drain needs to a max unavailable to be entered. This is a helper function so max_unavailable must be parameterized as an integer.

tests/validation/tests/v3_api/common.py
```diff
@@ -643,18 +645,31 @@ def wait_for_pod_to_running(client, pod, timeout=DEFAULT_TIMEOUT):
...
-    for node in nodes:
-        if node.worker and (not node.unschedulable):
-            for key, val in node.labels.items():
-                # Either one of the labels should be present on the node
-                if key == 'kubernetes.io/os' or key == 'beta.kubernetes.io/os':
-                    if val == os_type:
-                        schedulable_nodes.append(node)
+    if activeNodes:
+        for node in nodes:
+            if node.worker and (not node.unschedulable) and node.state == "active":
```
kinarashah, line 655: instead of having if and else separately with same logic, we can add `activeNodes` in this condition. 
so it becomes, 
```
// by default, assumption is all nodes are active unless the bool variable says to actually check 
nodeActive = True
if activeNodes:
   nodeActive = node.state == "active" 
if nodeActive and node.worker and ... 
```

tests/validation/tests/v3_api/test_zero_downtime.py
```diff
@@ -0,0 +1,352 @@
...
+MAX_UNAVAILABLE_WORKER = os.environ.get("RANCHER_MAX_UNAVAILABLE_WORKER", "10%")
+MAX_UNAVAILABLE_CONTROLPLANE = os.environ.get("RANCHER_MAX_UNAVAILABLE_CONTROLPLANE", "1")
+NODE_COUNT_CLUSTER = os.environ.get("RANCHER_NODE_COUNT_CLUSTER", 3)
+host = "test" + str(random_int(10000, 99999)) + ".com"
+path = "/name.html"
+namespace = {"p_client": None, "ns": None, "cluster": None, "project": None,
+             "nodes": []}
+backup_info = {"backupname": None, "backup_id": None, "workload": None,
+               "backupfilename": None, "etcdbackupdata": None}
+K8S_VERSION_PREUPGRADE = os.environ.get("RANCHER_K8S_VERSION_PREUPGRADE", "v1.15.12-rancher2-2")
```
kinarashah, line 18: nit: move this up so all the env vars are first, followed by other lowercase vars

tests/validation/tests/v3_api/test_zero_downtime.py
```diff
@@ -0,0 +1,352 @@
...
+K8S_VERSION_UPGRADE = os.environ.get("RANCHER_K8S_VERSION_UPGRADE", "v1.16.10-rancher2-1")
+
+
+def test_zero_downtime():
+    client, cluster, workload, ingress, rke_config = get_and_validate_cluster()
+    # Update Cluster to k8 version + upgrade strategy maxUnavailable worker
+    rke_updated_config = rke_config.copy()
```
kinarashah, line 25: let's also put this in a common func, so it becomes, 
```
rke_config = get_default_rke_config()
# test drain true 
rke_config["upgradeStrategy"]["drain"] = True
```

kinarashah: `get_default_rke_config` would be the default config that we have when creating a cluster

tests/validation/tests/v3_api/test_zero_downtime.py
```diff
@@ -0,0 +1,352 @@
...
+    rke_updated_config["kubernetesVersion"] = K8S_VERSION_UPGRADE
+    rke_updated_config["upgradeStrategy"] = {
+        'drain': True,
+        'maxUnavailableWorker': MAX_UNAVAILABLE_WORKER,
+        'maxUnavailableControlplane': MAX_UNAVAILABLE_CONTROLPLANE,
+        'type': '/v3/schemas/nodeUpgradeStrategy'}
+    cluster = client.update(cluster,
+                            name=cluster.name,
+                            rancherKubernetesEngineConfig=rke_updated_config)
+    completed_nodes = validate_nodes_after_upgrade(cluster, workload)
```
kinarashah, line 35: nit: rename completed_nodes to `upgraded_nodes`

tests/validation/tests/v3_api/test_zero_downtime.py
```diff
@@ -0,0 +1,352 @@
...
+    nodes = client.list_node(clusterId=cluster.id).data
+    assert len(completed_nodes) == len(nodes), "Not all Nodes Upgraded"
+    node_ver = K8S_VERSION_UPGRADE.split("-")[0]
+    for node in nodes:
```
kinarashah, line 39: maybe I am missing something, why didn't we include this in some validate func as well?

tests/validation/tests/v3_api/test_zero_downtime.py
```diff
@@ -0,0 +1,352 @@
...
+        return 1
+    return max_unavailable
+
+
+def validate_nodes_after_upgrade(cluster, workload):
+    completed_upgrade = []
+    cp_nodes = get_role_nodes(cluster, "control")
+    etcd_nodes = get_role_nodes(cluster, "etcd")
+    worker_nodes = get_role_nodes(cluster, "worker")
+    # Validate Node Upgrades by job
```
kinarashah, line 215: nit: do you mean `by role` ?

PR #28249, opened 2020-08-05
expose multiple configs with a single instance of channelserver 

MonzElmasry:
This pr allows rancher to get multiple configs from kdm, using a single instance of channelserver


pkg/channelserver/channelserver.go
```diff
@@ -38,20 +38,22 @@ func GetURLAndInterval() (string, string) {
 
 }
 
-func run(driver, port, prefix string) chan error {
+func run() chan error {
```
briandowns, line 41: Is there a context we can pass in or create a context and use the cancel pattern here with it rather than making our own context pattern?

pkg/channelserver/handler.go
```diff
@@ -6,16 +6,10 @@ import (
 	rurl "net/url"
 )
 
-var k3sURL = "http://localhost:8115"
-var rke2URL = "http://localhost:8116"
+var rawURL = "http://localhost:8115"
```
briandowns, line 9: If this is not used anywhere else besides the function below, can you move it there and make it a const?

davidnuzik:
Needs at least committer review such as from @cjellick before we can merge this in.

cjellick:
I want @dramich to review this

pkg/channelserver/channelserver.go
dramich, line 41: Instead of dropping the inputs and hard coding everything this function can accept something like `cmdArgs []string` then expand when creating the command. This would give a lot more flexibility if this does need to be called elsewhere and also allows for testing around this function. 

`config-key` `path-prefix` and `listen-address` can come in this way and the others can be left alone as they are based off settings that can be manipulated externally.

PR #28252, opened 2020-08-05
print the kubeconfig file after the cluster is accessible

jiaqiluo:
print the kubeconfig file after the cluster is accessible


sangeethah:
LGTM

Issue #28158, opened 2020-07-30
User can disable both public and private access in edit EKS cluster


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Import an EKS cluster
Edit cluster
Uncheck `public access` and `private access` under the `VPC and Subnet` section.
Save

Edit cluster
Add a new node group
Save

**Result:**
Cluster will go into updating state and errors with:
`InvalidParameterException: Both private and public endpoint access cannot be false { RespMetadata: { StatusCode: 400, RequestID: "0e2a1dbb-83a0-4120-8e64-fc19cfc3ab7a" }, ClusterName: "aaron-eks6", Message_: "Both private and public endpoint access cannot be false" }`

In eks-operator logs:
```
{
  RespMetadata: {
    StatusCode: 400,
    RequestID: "3a75a343-e4bb-49d8-a010-45268f6f32c9"
  },
  ClusterName: "aaron-eks5",
  Message_: "Both private and public endpoint access cannot be false"
}, requeuing
```

Users should not be able to make an edit where both Public Access and Private Access are unchecked.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `master-3369-head`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported

aaronRancher:
The bug fix is verified with Rancher server `rancher\rancher:master-head` `version 8a60de75d` `master-8a60de75db5324c73535b71c087c3e4bd5f36d29-head`

When editing eks cluster, user is unable to disable both public and private access.
![Screen Shot 2020-08-24 at 11 26 39 AM](https://user-images.githubusercontent.com/57409052/91081988-d4077800-e5fc-11ea-9287-f804947d1800.png)

PR #28270, opened 2020-08-06
Add validation for public and private access

rmweir:
**Problem:**
A user could create an EKS cluster that did not have public or private access enabled. This is invalid and inevitably leads to an error in the eks-operator. This is because with neither enabled there is no way to communicate with the cluster. Prior, a user could create an EKS cluster that did not have public or private access enabled. This is invalid and inevitably leads
to an error in the eks-operator.

**Solution:**
An error will be returned if neither public nor private access is
enabled.

**Issue:**
https://github.com/rancher/rancher/issues/28158

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -68,7 +68,7 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 		return err
 	}
 
-	if err := v.validateEKSConfig(request, &clusterSpec); err != nil {
+	if err := v.validateEKSConfig(request, data); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/api/norman/customization/cluster/validator.go
```diff
@@ -329,13 +329,16 @@ func (v *Validator) validateGenericEngineConfig(request *types.APIContext, spec
...
 
-func (v *Validator) validateEKSConfig(request *types.APIContext, spec *v32.ClusterSpec) error {
-	if spec.EKSConfig == nil {
+func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[string]interface{}) error {
+	eksConfig, ok := cluster["eksConfig"].(map[string]interface{})
+	if !ok {
 		return nil
 	}
 
+	amazonCredential, _ := eksConfig["amazonCredentialSecret"].(string)
```
cbron, line 338: Don't want to return an error if the secret is empty ?

rmweir: we could although an error should be returned by access.ByID

PR #28305, opened 2020-08-08
Set system-chart branch to dev-v2.4

deniseschannon:


cbron:
Fixed conflicts in https://github.com/rancher/rancher/pull/28314

Issue #24203, opened 2019-11-22
Imported AKS clusters show in error.  Can we mark Imported AKS to silence alarms?


slessardjr:
**What kind of request is this (question/bug/enhancement/feature request): Enhancement/Bug**

**Steps to reproduce (least amount of steps as possible):**

1. Create AKS Cluster in Azure Portal
2. Create Cluster in Rancher, select import
3. Run kubectl command to import AKS Cluster in Rancher

**Result:**
1. Rancher will show the cluster in error.  Specifically the Controller Manager and Scheduler are shown as unhealthy.  This is already a known upstream k8s bug from rancher/rancher#23451 

**Other details that may be helpful:**
1.  If you create an AKS cluster, Rancher seems to know that it's managed and will gray out those errors and ignore. Ideally, there would be a way to annotate if the imported cluster is from a managed service so the same "feature" of ignoring the error could be used.
2.  This could affect other cloud providers EKS or GKE.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.3.1
- Installation option (single install/HA): HA

gz#11712

gz#12343

cjellick:
@deniseschannon I had a user point this out at KubeCon as well. Apparently, this did not happen in v2.2, but it started happening in v2.3

mikearruda:
Glad to hear I am not the only one facing this issue. Looking forward to the fix.

davesykeselateral:
Yes, happening to me too.

ThEWiZ76:
v2.3.3 same issue

codyrancher:
I did digging to see if we could carry over the AKS Cluster fix to an AKS Import Cluster. Unfortunately there doesn't appear to be any information we can use to determine the provider of an Import Cluster. Without the provider information we can't mute these alerts without muting them for all providers.

@deniseschannon I think we either need someone from the backend to tell us a way to determine the provider or we may need to close this issue.

iboware:
Having the same problem with the version 2.4.1

elvishsu66:
I have this problem for a while and found this issue now. Can't UI team check out a good way to mute the alert and warning? Or check out the nodes' system info?

slessardjr:
[...]

PR #28310, opened 2020-08-09
Added abiity to drop componentstatuses based on additional annotation

ibrokethecloud:
During the import of managed k8s clusters, such as AKS, the component status fails because scheduler and controller-manager cs are not valid.

This cause the Rancher dashboard to show a red warning in the main dashboard page.

For certain customers who are only using managed k8s clusters, this can send the wrong message every time they login to the rancher ui.

The tweak allows users to specify a new annotation in the Rancher UI during the process to create "import cluster" type clusters.

An annotation of the form:

`lifecycle.cattle.io/ignorecslist: scheduler, controller-manager`

Will cause the healthsyncer controller to drop the CS for scheduler and controller-manager.

This will default the UI to drop the error and show a healthy cluster.

Ideally the cs for missing components should be greyed out but this will need a UI fix.

Related issue: https://github.com/rancher/rancher/issues/24203

### DIFF
--- a/pkg/controllers/user/healthsyncer/healthsyncer.go
+++ b/pkg/controllers/user/healthsyncer/healthsyncer.go
@@ -1,6 +1,7 @@
 package healthsyncer
 
 import (
+	"strings"
 	"time"
 
 	"context"
@@ -23,7 +24,8 @@ import (
 )
 
 const (
-	syncInterval = 15 * time.Second
+	syncInterval            = 15 * time.Second
+	managedK8SAnnotationKey = "lifecycle.cattle.io/ignorecslist"
 )
 
 type ClusterControllerLifecycle interface {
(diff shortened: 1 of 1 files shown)


mrvarmazyar (commented):
Supposedly it needs to assignee to someone in rancher team, I've just tested it, and it works fine!

mohag:
There is [this deprecation of the API](https://github.com/kubernetes/kubernetes/pull/93570) as well:

Eric-Fontana-Tjx:
Bump - Why is this not merged, it looks REALLY bad for Rancher when you import AKS clusters.   Need this ASAP.

mohag:
This would be a nice option to fix the kubeadm clusters showing broken as well

ibrokethecloud:
2.6.x handles this. Closing this issue as its not needed.

mohag:
In 2.6, they still show broken on a kubeadm cluster (which disables the HTTP port which is used by componentstatuses)
image

PR #28315, opened 2020-08-10
change upgrade app name to reflect cluster type

MonzElmasry:
This pr changes the cluster-upgrade app name to match cluster type
issue : https://github.com/rancher/rke2/issues/154

### DIFF
--- a/pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
+++ b/pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
@@ -80,7 +80,7 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 	}
 
 	// create or update k3supgradecontroller if necessary
-	if err = h.deployK3sBasedUpgradeController(cluster.Name); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -94,7 +94,7 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 
 // deployK3sBaseUpgradeController creates a rancher k3s/rke2 upgrader controller if one does not exist.
 // Updates k3s upgrader controller if one exists and is not the newest available version.
-func (h *handler) deployK3sBasedUpgradeController(clusterName string) error {
+func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isRke2 bool) error {
```
brandond, line 97: We're using two bools to set a single value, but there's no possible case where both would be true. Why not create a DistrubutionType enum with values 'k3s' and 'rke2'?

Although I don't even necessarily like that, as it leaks rke2-specific code upstream into k3s. Maybe just pass appName through from an argument to deployK3sBasedUpgradeController, and k3s or rke2 can set is as needed?

Issue #25108, opened 2020-01-25
Bug: Unable to set prefix_path in windows clusters


patrick0057:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1) Create windows cluster
2) Attempt to set prefix_path in cluster yml on windows cluster

**Result:**
Windows agents cannot set it properly because there is a difference in how paths are done in windows and linux.

```
time="2020-01-25T00:55:57Z" level=info msg="For process service-sidekick, Binds has changed from [c:/opt:c:/host/opt c:/etc/cni/net.d:c:/host/etc/cni/net.d c:/etc/kube-flannel:c:/host/etc/kube-flannel c:/run:c:/host/run //./pipe/rancher_wins://./pipe/rancher_wins] to [/opt/rke/opt:c:/host/opt /opt/rke/etc/cni/net.d:c:/host/etc/cni/net.d /opt/rke/etc/kube-flannel:c:/host/etc/kube-flannel /opt/rke/run:c:/host/run //./pipe/rancher_wins://./pipe/rancher_wins]"
time="2020-01-25T00:55:58Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T00:57:58Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T00:59:58Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T01:01:58Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T01:03:58Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T01:05:59Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T01:07:59Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
time="2020-01-25T01:09:59Z" level=error msg="failed to check plan: Error response from daemon: invalid volume specification: '/opt/rke/opt:c:/host/opt'"
```

[...]

PR #28344, opened 2020-08-11
[v2.4] Win Prefix Path

luthermonson:
Windows agent bootstrap updated to support win prefix


luthermonson:
https://github.com/rancher/rancher/issues/25108

Issue #27777, opened 2020-06-29
[KEv2] Rework EKS managed node groups upgrade


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Steps to reproduce (least amount of steps as possible):**
upgrade nodegroup version

**Result:**
Takes ~50 mins for 1 node, this is due to the method AWS uses for upgrading managed node groups: https://docs.aws.amazon.com/eks/latest/userguide/managed-node-update-behavior.html

`The Auto Scaling group maximum size and desired size are incremented by twice the number of distinct Availability Zones of the Auto Scaling group to ensure at least one new instance comes up in every Availability Zone of your node group.`

Nodegroups will no longer automatically upgrade. Once a cluster is upgraded a user will be able to upgrade the nodegroup manually or they can create a nodegroup and delete the old node group.

**Other details that may be helpful:**
EKSctl instructs the user to create a new node group with the desired version number, then delete the old node group:
https://eksctl.io/usage/cluster-upgrade/

rmweir:
some scenarios and their expected behaviors after PR is in:
1. create regular cluster with proper kubernetes version and proper nodegroup version
success
2. create cluster with unsupported kubernetes version but valid format
if kubernetes version doesn't match a nodegroup version a controller error will quickly be shown for that. Otherwise once cluster is being created error will show.
3. create cluster with invalid kubernetes version format
same as 2.
4. create cluster with no kubernetes version
quick controller error
5. create cluster with empty kubernetes version
api level error saying kubernetes version cannot be empty string
6. create nodegroup with no version
quick controller error
7. create nodegroup with empty version
api level error saying kubernetes version cannot be empty string
8. create nodegroup with invalid version format
quick controller error saying nodegroup version must match cluster's kubernetes version

deniseschannon:
Based on what I can see, this looks fixed with eks-operator v0.1.0-rc30 and can be tested now

aaronRancher:
Tested with Rancher server `rancher\rancher:master-head` `version 67e24f2`

Tested import EKS cluster on the following scenarios:
- create nodegroup with no version
quick controller error
![Screen Shot 2020-09-22 at 5 07 27 PM](https://user-images.githubusercontent.com/57409052/93949739-464fa300-fcf6-11ea-96ee-c23f4a17ecc6.png)
```
{
"baseType": "error",
"code": "InvalidFormat",
"fieldName": "eksConfig",
[...]

PR #28357, opened 2020-08-12
Nodegroup upgrade

rmweir:
**Problem:**
Users need the option to upgrade nodegroups the amazon way or to create new nodegroups and delete ones that are outdated.

**Solution:**
Nodegroup versions now must be provide. On create they must match the control plane's kubernetes version. Control plane and nodegroup versions will be validated against supported versions. More in depth validation will be performed in the eks-operator.

**Issue:**
https://github.com/rancher/rancher/issues/27777
https://github.com/rancher/rancher/issues/28968

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"time"
 
-	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
@@ -71,7 +71,7 @@ func (v *Validator) Validator(request *types.APIContext, schema *types.Schema, d
 		return err
 	}
 
-	if err := v.validateEKSConfig(request, data); err != nil {
+	if err := v.validateEKSConfig(request, data, &clusterSpec); err != nil {
 		return err
 	}
 
@@ -332,7 +332,7 @@ func (v *Validator) validateGenericEngineConfig(request *types.APIContext, spec 
 
 }
 
-func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[string]interface{}) error {
+func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[string]interface{}, clusterSpec *v32.ClusterSpec) error {
 	eksConfig, ok := cluster["eksConfig"].(map[string]interface{})
 	if !ok {
 		return nil
@@ -353,21 +353,24 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
 
 	createFromImport := request.Method == http.MethodPost && eksConfig["imported"] == true
 
-	var prevSpec *eksv1.EKSClusterConfigSpec
+	var prevCluster *v3.Cluster
 
 	if request.Method == http.MethodPut {
-		prevCluster, err := v.ClusterLister.Get("", request.ID)
+		var err error
+		prevCluster, err = v.ClusterLister.Get("", request.ID)
 		if err != nil {
 			return err
 		}
-		prevSpec = prevCluster.Spec.EKSConfig
 	}
 
 	if !createFromImport {
-		if err := validateEKSNodegroups(request, eksConfig, prevSpec); err != nil {
+		if err := validateEKSKubernetesVersion(clusterSpec, prevCluster); err != nil {
 			return err
 		}
-		if err := validateEKSAccess(request, eksConfig, prevSpec); err != nil {
+		if err := validateEKSNodegroups(clusterSpec); err != nil {
+			return err
+		}
+		if err := validateEKSAccess(request, eksConfig, prevCluster); err != nil {
 			return err
 		}
 	}
@@ -420,15 +423,15 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
 	return nil
 }
 
-func validateEKSAccess(request *types.APIContext, eksConfig map[string]interface{}, prevSpec *eksv1.EKSClusterConfigSpec) error {
+func validateEKSAccess(request *types.APIContext, eksConfig map[string]interface{}, prevCluster *v3.Cluster) error {
 	publicAccess, _ := eksConfig["publicAccess"]
 	privateAccess, _ := eksConfig["privateAccess"]
 	if request.Method != http.MethodPost {
 		if publicAccess == nil {
-			publicAccess = prevSpec.PublicAccess
+			publicAccess = prevCluster.Spec.EKSConfig.PublicAccess
 		}
 		if privateAccess == nil {
-			privateAccess = prevSpec.PrivateAccess
+			privateAccess = prevCluster.Spec.EKSConfig.PrivateAccess
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


paynejacob (dismissed):

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -37,6 +37,10 @@ type Validator struct {
 	CisBenchmarkVersionLister     v3.CisBenchmarkVersionLister
 }
 
+var (
+	availableVersions = []string{"1.17", "1.16", "1.15"}
```
prachidamle, line 41: These should be read from some system settings/config. Do we need to hardcode here?

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -408,6 +413,61 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
...
+		ngMap, _ := ng.(map[string]interface{})
+		name, _ := ngMap["nodegroupName"].(string)
+		if name == "" {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("nodegroupName cannot be an empty"))
+		}
+
+		ngVersion, ok := ngMap["version"].(string)
+		if !ok {
+			if create {
+				errors = append(errors, fmt.Sprintf("verison for nodegroup [%s] is required", name))
```
prachidamle, line 451: nit: typo verison

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -408,6 +413,61 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
...
+			}
+			continue
+		}
+
+		for _, version := range availableVersions {
+			if ngVersion == version {
+				continue nodegroups
+			}
+		}
+		errors = append(errors, fmt.Sprintf("invalid kubernetes version [%s] for nodegroup [%s]. Supported versions: %v", ngVersion, name, availableVersions))
```
prachidamle, line 461: Instead of appending error, shouldn't we fail-fast and error out as we find some invalid nodegroup?

rmweir: I feel it is better UX to receive all nodegroup version errors at once instead of resubmitting the request multiple times until all errors are resolved.

pkg/api/norman/customization/cluster/validator.go
rmweir: Thinking of removing this entirely since this is already done by EKS api and keeping it prevents us from updating version support without a rancher release.

paynejacob (approved):

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -440,26 +443,52 @@ func validateEKSAccess(request *types.APIContext, eksConfig map[string]interface
...
-		if prevSpec.NodeGroups != nil && len(prevSpec.NodeGroups) == 0 {
-			return noNodegroupsError
-		}
-	} else if len(nodegroups) == 0 {
-		return noNodegroupsError
+// validateEKSKubernetesVersion checks whether a kubernetes version is provided and if it is supported
+func validateEKSKubernetesVersion(eksConfig map[string]interface{}) error {
+	clusterVersion, _ := eksConfig["kubernetesVersion"]
+	if clusterVersion == nil {
+		return nil
```
prachidamle, line 450: a nil defaults to some default version?

rmweir: nil means the field has not been managed through rancher yet and its effective value is its upstream value. The upstream value can be found on the status.

prachidamle (dismissed):
LGTM

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -264,13 +264,6 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			}
 		}
 
-		// check for unauthorized error
-		if failureMessage != "" {
-			if strings.Contains(failureMessage, "403") {
-				return e.setFalse(cluster, apimgmtv3.ClusterConditionUpdated, "cannot access EKS, check cloud credential")
-			}
-		}
-
```
cbron, line 273: Why'd we remove this error for create ? The below addition is in the upgrading block.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -440,26 +443,52 @@ func validateEKSAccess(request *types.APIContext, eksConfig map[string]interface
...
+			errors = append(errors, fmt.Sprintf("nodegroup [%s] version cannot be empty string", name))
+			continue
 		}
 	}
+
+	if len(errors) != 0 {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf(strings.Join(errors, ";")))
+	}
 	return nil
 }
```
cbron, line 494: > On create they must match the control plane's kubernetes version.

I don't see this check. I assumed the `validateEKSNodegroups` func would return the version to check against the control plane version.

rmweir: it's check in validateKubernetesVersion

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: eks-operator will now be in a failed updating state if this or any other error is encountered, but I do need to convert this error regardless of state, so I moved that to the top

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -440,25 +443,73 @@ func validateEKSAccess(request *types.APIContext, eksConfig map[string]interface
...
+		nodegroups = prevCluster.Status.EKSStatus.UpstreamSpec.NodeGroups
+	}
+
+	currentVersion := prevCluster.Spec.EKSConfig.KubernetesVersion
+	if currentVersion == nil {
+		currentVersion = prevCluster.Status.EKSStatus.UpstreamSpec.KubernetesVersion
+	}
+	for _, ng := range nodegroups {
+		if aws.StringValue(currentVersion) != aws.StringValue(ng.Version) {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "nodegroups' versions must match cluster's kubernetes version before cluster upgrade")
```
cbron, line 475: How come the `clusterVersion` from spec is never checked here ? All we do is make sure its not blank

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -440,25 +443,73 @@ func validateEKSAccess(request *types.APIContext, eksConfig map[string]interface
...
 		}
-		if prevSpec.NodeGroups != nil && len(prevSpec.NodeGroups) == 0 {
-			return noNodegroupsError
+		nodegroups = prevCluster.Status.EKSStatus.UpstreamSpec.NodeGroups
+	}
+
+	currentVersion := prevCluster.Spec.EKSConfig.KubernetesVersion
+	if currentVersion == nil {
+		currentVersion = prevCluster.Status.EKSStatus.UpstreamSpec.KubernetesVersion
+	}
```
cbron, line 472: this idea of `a) check spec, b) check status`, is that for import ?

pkg/api/norman/customization/cluster/validator.go
rmweir: All we are doing here is validating that the nodegroup version and current cluster version are equal before allowing an attempted upgrade. It being in the request tells us an upgrade is being attempted. If the version is unsupported or something, a controller error will appear.

pkg/api/norman/customization/cluster/validator.go
rmweir: Not necessarily, it just means it is not being updated. If the version is not in the request it will result in a nil value in the struct.

cbron: Ok, that makes sense, you would still check current version if its not being updated in this request.

pkg/api/norman/customization/cluster/validator.go
cbron: I was confused about purpose here, we are really checking the old versions to make sure they line up first, and if those line up we can proceed. Might be nice to comment that inline somewhere.

cbron (dismissed):
One nit, otherwise lgtm

cbron (approved):
extra version check was removed, lgtm

Issue #28094, opened 2020-07-27
Editing a roletemplate to add new rules does not grant expected permissions to users with that role


mrajashree:
1. On 2.4.5 create a project scoped role `cloned-project-owner` by cloning project-owner
2. Add a user `testuser` with this `cloned-project-owner` role to a project
3. The `testuser` can enable project monitoring and edit project resource quota 
4. Upgrade to 2.4-head.
5. The builtin project-owner role will have a new rule for the resource "project" with verb "own" that gives it ownership. The cloned role `cloned-project-owner` will not have this rule upon upgrade.
6. Edit `cloned-project-owner` through API to add the same rule as the project-owner for "own" verb on "project" resource.
7. On editing the roletemplate, its corresponding k8s role in the project's namespace also gets updated to have this new rule
7. Login as `testuser`.

**Result**
testuser cannot enable monitoring or edit project resource quota

**Expected**
testuser should be able to perform everything a project-owner can, including monitoring actions and editing project quota

**Other details that may be helpful:**
The corresponding kubernetes role in the project's namespace gets updated as expected. And the rolebinding for this user binding it to the k8s role is intact upon upgrade. So this should have given the `testuser` expected permissions

gzrancher/rancher#9345

ryansann:
This issue has been hard to track down. I am still actively working on it, and wanted to provide an update with my findings so far. The issue is not related to the changes that @mrajashree made having to do with project owner permissions. I was able to reproduce the same issue in Rancher `v2.4.5` by adding and removing the inherited `Project Owner` role template to the `cloned-project-owner` role template. I also ran the same test in Rancher `v2.3.8` and did not find this issue happening. We were not using the `membershipBindingOwnerIndex` in that version and were directly retrieving RoleBindings from the api-server as opposed to using Kubernetes caching mechanisms as we have from `v2.4.0` onward.

[...]

PR #28371, opened 2020-08-12
[2.4] Fix cache inconsistencies for RoleBindings/ClusterRoleBindings

ryansann:
Issue addressed by this PR:
- rancher/rancher#28094

We were unintentionally removing important labels from cached objects. Any modifications should go through the api server so their deltas can be consumed and processed by the controller's underlying SharedIndexInformer and cache/indices can be updated accordingly. I added a deep copy to mitigate this issue. The cache no longer contains stale keys that resolve to nil objects with this change. I also addressed an issue with the `membershipBindingOwnerIndex` where we were returning once we found a single owner. Technically there can be multiple owners, so this function now returns a list of all owner keys.

### DIFF
--- a/pkg/controllers/management/auth/indexes.go
+++ b/pkg/controllers/management/auth/indexes.go
@@ -1,6 +1,8 @@
 package auth
 
 import (
+	"strings"
+
 	v1 "k8s.io/api/rbac/v1"
 	meta2 "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/auth/indexes.go
```diff
@@ -37,20 +37,24 @@ func rbRoleSubjectKeys(roleName string, subjects []v1.Subject) []string {
...
-			}
+	meta, err := meta2.Accessor(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	var keys []string
+	for k, v := range meta.GetLabels() {
+		if v == membershipBindingOwner {
+			keys = append(keys, meta.GetNamespace()+"/"+k)
```
mrajashree, line 53: nit: we can save namespace returned by meta.GetNamespace() before the for loop and use it?

Issue #28355, opened 2020-08-12
User is not able to edit an EKS cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create a hosted EKS cluster
- Wait for the cluster to come up active
- Edit cluster
- User is NOT able to edit the cluster

<img width="1201" alt="Screen Shot 2020-08-11 at 9 12 49 PM" src="https://user-images.githubusercontent.com/26032343/89974364-b8bd6580-dc17-11ea-977c-1ca27c6e5af2.png">

**Expected Result:**
User should be able to edit the cluster

**Other details that may be helpful:**
This happens for an imported EKS cluster also

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `d19f179c4`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): hosted and imported EKS clusters
- Kubernetes version (use `kubectl version`):

```
1.17
```

sowmyav27:
**Verified on latest master-head - commit id: `d19f179c4` ui tag: `latest2`**

- Deploy an EKS cluster
- Edit cluster. User is able to Edit the cluster --> Add nodegroup --> Save changes made
- Cluster goes into Updating state --> Comes to Active state

PR #28377, opened 2020-08-13
Renamed provider json key name

rmweir:
Renamed cluster's provider field's json key from "provider" to
"clusterProvider". This is because the UI computes values for
provider which are overridden when it is provided. This causes
the EKS edit UI to not properly show. The values from this
field cannot always be used in place of the computed provider.
For example, EKSv1 and EKSv2 would have the same provider
value in the backend but need to be handled differently in the
UI.

**Issue:**
https://github.com/rancher/rancher/issues/28355


ibuildthecloud:
@westlywright Can we just change the UI field?

Issue #28128, opened 2020-07-29
Import EKS cluster does not work from rancher when the EKS cluster has no nodegroups


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create an EKS cluster with no nodegroups in the EKS console.
- Import the cluster into Rancher.
- The cluster will be stuck in `Provisioning` state.

**Expected Result:**
User should not be allowed to import a cluster with no nodegroups

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `ca3f705a4 `
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Hosted EKS
- Kubernetes version (use `kubectl version`):

```
1.17
```

- Docker version (use `docker version`):

```
(paste the output here)
```

sowmyav27:
**On master-head - commit id: `ad697207c`**

1. Import an EKS cluster with no Nodegroups. 
2. Cluster is seen stuck with error - `Cluster health check failed: deployments.apps "cattle-cluster-agent" not found; Cannot deploy agent without nodegroups. Add a nodegroup.`
3. From the EKS console, add a nodeGroup to the cluster.
4. The nodegroup reflects in the UI and on the Edit cluster but the cluster is seen stuck in `Waiting` state with error `Cannot deploy agent without nodegroups. Add a nodegroup.`

Will test Step 3 when issue - https://github.com/rancher/rancher/issues/28422 is fixed and will log a separate bug if any issue is found.

sowmyav27:
**On master-head - commit id: `53cf961efce`**

- Import an EKS cluster with no Nodegroups.
- Cluster is seen stuck with error - Cluster health check failed: deployments.apps "cattle-cluster-agent" not found; Cannot deploy agent without nodegroups. Add a nodegroup.
- From the EKS console, add a nodeGroup to the cluster.
- Cluster comes up Active on rancher

PR #28380, opened 2020-08-13
Fail wait on import if cluster has no nodegroups

rmweir:
**Problem:**
EKS cluster stuck in waiting if it does not have nodegroups.

**Solution:**
Fail waiting and show error message recommending the user adds a nodegroup.

**Issue:**
https://github.com/rancher/rancher/issues/28128

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -183,6 +183,22 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 					return cluster, err
 				}
 			}
+			addNgMessage := "Cannot deploy agent without nodegroups. Add a nodegroup."
+			if len(cluster.Spec.EKSConfig.NodeGroups) == 0 {
+				cluster, err = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting, addNgMessage)
+				if err != nil {
+					return cluster, err
+				}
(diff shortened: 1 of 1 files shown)


mrajashree (approved):

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -183,6 +183,22 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
 				}
 			}
+			addNgMessage := "Cannot deploy agent without nodegroups. Add a nodegroup."
+			if len(cluster.Spec.EKSConfig.NodeGroups) == 0 {
+				cluster, err = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting, addNgMessage)
+				if err != nil {
+					return cluster, err
+				}
+			} else {
+				if apimgmtv3.ClusterConditionWaiting.GetMessage(cluster) == addNgMessage {
```
cbron, line 193: This seems somewhat brittle. What if it gets into a state where the first part of conditional executes and error message is set and then rancher gets restarted. Would first part execute again and update error message ? Otherwise this is good case for a constant error code.

rmweir: The error would be updated. The SetFalse function checks if the current message matches the inputted message and updates it if not.

cbron (approved):

Issue #28130, opened 2020-07-29
User is able to import the same EKS cluster multiple times on Rancher UI


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Import cluster A (created in EKS) as test-1
- Import cluster A (created in EKS) as test-2
- Add a nodegroup ng1 in test-1
- The node gets added in the node page for test-2 cluster also
- Edit test-2 and add a node group ng2.
- The cluster goes into Updating state. ng1 gets deleted and ng2 gets added in test-2

**Expected Result:**
To avoid the above scenario, we must not allow for the same EKS cluster to be imported multiple times

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `ca3f705a4 `
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): EKS custom
- Kubernetes version (use `kubectl version`):

```
1.17
```

aaronRancher:
The bug fix is verified with Rancher server `rancher\rancher:master-head` `version d5e1b21b`

Steps:
Import cluster A (created in EKS) as test-1
Import cluster A (created in EKS) as test-2

Result:
User is unable to import the same cluster twice:
![Screen Shot 2020-08-27 at 4 37 20 PM](https://user-images.githubusercontent.com/57409052/91505194-d92a2880-e883-11ea-9977-09ad92a8a8e2.png)

PR #28390, opened 2020-08-14
Multi import fix

rmweir:
**Problem:**
User can create multiple Rancher clusters referencing the same EKS cluster.

**Solution:**
Add validation to check that a Rancher cluster does not already exist with the given displayName.

**Issue:**
https://github.com/rancher/rancher/issues/28130

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -22,9 +22,11 @@ import (
 	mgmtSchema "github.com/rancher/rancher/pkg/schemas/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/robfig/cron"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type Validator struct {
+	ClusterClient                 v3.ClusterInterface
 	ClusterLister                 v3.ClusterLister
 	ClusterTemplateLister         v3.ClusterTemplateLister
 	ClusterTemplateRevisionLister v3.ClusterTemplateRevisionLister
@@ -335,6 +337,7 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
 		return nil
 	}
 
+	// check user's access to cloud credential
 	amazonCredential, _ := eksConfig["amazonCredentialSecret"].(string)
 
 	var accessCred mgmtclient.CloudCredential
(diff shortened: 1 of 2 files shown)


pkg/api/norman/customization/cluster/validator.go
```diff
@@ -347,14 +350,14 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
 		return httperror.WrapAPIError(err, httperror.ServerError, fmt.Sprintf("error accessing cloud credential"))
 	}
 
-	if request.Method == http.MethodPost && eksConfig["imported"] == true {
-		// cluster is imported and has not copied spec from upstream yet, validation below does not apply
-		return nil
-	}
+	importCreate := request.Method == http.MethodPost && eksConfig["imported"] == true
```
cbron, line 353:
```suggestion
	createFromImport := request.Method == http.MethodPost && eksConfig["imported"] == true
```
or 
```suggestion
	createFromExisting := request.Method == http.MethodPost && eksConfig["imported"] == true
```

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -384,8 +388,10 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
 	}
 
 	if !publicAccess && !privateAccess {
-		return httperror.NewAPIError(httperror.InvalidBodyContent,
-			"public access, private access, or both must be enabled")
+		if !importCreate {
+			return httperror.NewAPIError(httperror.InvalidBodyContent,
+				"public access, private access, or both must be enabled")
```
cbron, line 393: could make the case that this should still fail

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -394,15 +400,41 @@ func (v *Validator) validateEKSConfig(request *types.APIContext, cluster map[str
 
 	// validation for creates only
```
cbron, line 402: any reason this check is below public/private check, but above networking ?

rmweir: Public/private can be updated while the other networking fields cannot.

pkg/api/norman/customization/cluster/validator.go
rmweir: Could you? When importing it is typical to only provide the few fields that are necessary to import the cluster since the spec will be overwritten anyway. This will partially change soon.

Issue #28129, opened 2020-07-29
Rancher UI does not show errors when a cluster fails to come up


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster (selected subnets and security group). Cluster was seen in `Provisioning` state on the UI
- But the cluster provisioning failed.
- EKS controller logs (in the local cluster of the HA setup) showed:

```
1 controller.go:166] error syncing 'cattle-global-data/c-jkbb5': handler eks-controller: error creating stack with VPC template: stack failed to create: The maximum number of VPCs has been reached. (Service: AmazonEC2; Status Code: 400; Error Code: VpcLimitExceeded; Request ID: <>), requeuing
```
- This error was not seen on the UI

**Result:**
It will good to have the errors seen on the rancher UI while the cluster is Provisioning/Updating
Currently all errors are seen only in the eks-controller logs. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `ca3f705a4 `
- Installation option (single install/HA):  HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Hosted EKS
- Kubernetes version (use `kubectl version`):

```
1.17
```

sowmyav27:
**on master-head - commit id: `b1abc4156fd`**

- Error is seen on the UI for a hosted EKS cluster- `Error creating stack with VPC template: stack failed to create: The maximum number of internet gateways has been reached. (Service: AmazonEC2; Status Code: 400; Error Code: InternetGatewayLimitExceeded; Request ID: 196b0b59-1765-439e-9977-25e2ed38700a; Proxy: null)`

PR #28406, opened 2020-08-14
Record provisioning errors from eks-operator

rmweir:
**Problem:**
Errors are not displayed to user during provisioning.

**Solution:**
Copy failure message from EKS cluster config object to Rancher cluster. If it is not empty set provisioning to false.

**Issue:**
https://github.com/rancher/rancher/issues/28129

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -162,6 +162,7 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 	// get EKS Cluster Config's phase
 	status, _ := eksClusterConfigDynamic.Object["status"].(map[string]interface{})
 	phase, _ := status["phase"]
+	failureMessage, _ := status["failureMessage"].(string)
 
 	switch phase {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -171,9 +172,13 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			return cluster, err
 		}
 
-		logrus.Infof("waiting for EKS cluster [%s] to finish creating", cluster.Name)
 		e.clusterEnqueueAfter(cluster.Name, 20*time.Second)
-		return cluster, nil
+		if failureMessage == "" {
+			logrus.Infof("waiting for cluster EKS [%s] to finish creating", cluster.Name)
+			return e.setUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "")
```
mrajashree, line 178: why is there a difference between the provisioned condition being set to unknown here vs false if the message exists?

mrajashree (approved):

cbron (approved):

Issue #28411, opened 2020-08-14
Node provisioning stuck on `Waiting for ssh` when provisioning from an airgapped environment with proxy


bmdepesa:
rancher/rancher:v2.4.5

* Create an airgapped single-node rancher setup which is configured to use a proxy for outbound connections
  * https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/air-gap/
  * https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/single-node-docker/proxy/
* Create a nodedriver cluster (ec2,do)
* The node is created in the cloud provider, but provisioning hangs on `Waiting for SSH to be available`

Expected:
* The node provisions successfully

Notes:
* The Rancher instance is accessible through a public NLB to external machines
* Outbound connections from Rancher are configured to go through the proxy
* `curl` commands from within the Rancher container go through the proxy successfully
* `ssh` from within the Rancher container fails unless using a jumphost
  * exec into the Rancher container
  * `ssh` to some machine (in this case, some random machine on ec2 in a different VPC)
  *  https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts
```
root@35d3c0b70171:/var/lib/rancher# ssh -i mykey.pem -o ProxyCommand="ssh -W %h:%p host.compute.amazonaws.com:3131" ubuntu@host
root@host.compute.amazonaws.com: Permission denied (publickey).
ssh_exchange_identification: Connection closed by remote host

# this example above fails due to the key error, but typically this would time out 

root@35d3c0b70171:/var/lib/rancher# ssh ubuntu@host
ssh: connect to host <host> port 22: Connection timed out
```
* The same configuration in a non-airgapped setup (but still using a proxy) allows a cluster to be provisioned successfully, though no `ssh` traffic is seen traversing the proxy

Jono-SUSE-Rancher:
Does this now work with the proxy changes that were made in 2.5.6?

bmdepesa:
@slickwarren can you retest this?

slickwarren:
Rancher/rancher:v2.5.6  I was able to reproduce this issue:

* Create an airgapped node, then configure it to use a proxy for outbound connections. Deploy rancher via single-install setup instructions
  * https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/air-gap/
[...]

Issue #25488, opened 2020-02-19
Rancher K8s Auth Proxy does not utilize forward proxy (if defined) to proxy communication


Oats87:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**
Spin up Rancher HA in semi-airgapped environment (with HTTP/HTTPS forward proxies and no ability for Rancher to directly access the internet without utilizing these forward proxies)
Attempt to provision a ClusterDriver cluster (GKE, EKS, AKS for example)
Observe that while the cluster is in fact able to provision (kontainer-engine is able to use the proxies when provisioning as they are passed in via environment variables), the `cattle-node-agent` and `cattle-cluster-agent` are unable to deploy into the cluster and the cluster never finishes provisioning. Eventually you will see an error message similar to: 
```
unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request unable to recognize "management-statefile_path_redacted": the server is currently unable to handle the request
``` 

in the Rancher UI for this cluster.

**Result:**
Rancher is unable to communicate with the underlying cluster when it attempts to deploy its agents to the cluster, because the K8s auth proxy does not forward traffic properly through a defined forward proxy for clusters that are cluster driver clusters.

Tracing this down, it is seen that these agents are deployed using the clusterdeploy controller here: https://github.com/rancher/rancher/blob/v2.3.5/pkg/controllers/management/clusterdeploy/clusterdeploy.go#L212 

When looking at the `kubectl` package, it can be seen that the kube config context that is passed to the `kubectl` command contains the server as `https://localhost:443/k8s/clusters/<cluster-id>/`, which means it will proxy through the K8s authentication proxy. Due to the fact that the authentication proxy is not properly utilizing the underlying proxy (and this is a cluster driver cluster which does not utilize the underlying agent tunnels for communication), this breaks down and the agents are never properly deployed.

[...]

PR #28412, opened 2020-08-15
[v2.3] utilize proxy in proxy handler

kinarashah:
https://github.com/rancher/rancher/issues/25488 

**Setup:** 
Spin up Rancher HA in semi-airgapped environment (with HTTP/HTTPS forward proxies and no ability for Rancher to directly access the internet without utilizing these forward proxies)
Attempt to provision a ClusterDriver cluster (GKE, EKS, AKS for example)

**Note** 
Only tested for ClusterDriver clusters, but I'm guessing it'd also break for RKE clusters -- Couldn't test that out because of https://github.com/rancher/rancher/issues/28411 

**Issues:** 
(1) agents are unable to deploy into the cluster and the cluster never finishes provisioning 
This is because the request doesn't go through proxy and directly to cluster's endpoint. 

(2) kubectl exec shell and view logs from UI doesn't work 
For upgrade requests, the `UpgradeAwareHandler` doesn't use transport.Proxy but instead relies on the dialer. We pass only native dialer for cloud drivers which don't utilize proxy. 

**Solution:**
I could think of 2 ways to fix this: This PR implements (a) 
(a) Setting transport.Proxy for all requests + Custom proxy for upgrade requests 
- Applies to all clusters, Setting `transport.Proxy = http.ProxyFromEnvironment` on transport from cluster dialer fixes the issue. While this applies to all clusters, I think it makes sense that all requests go through the configured proxy unless no_proxy is set.
- Custom proxy for upgrade requests is similar to `UpgradeAwareHandler` avoiding the upgrade path 

(b) Change the cloud driver's dialer function to handle the connect to proxy 
- Applies to only hosted providers, sample implementation - https://github.com/rancher/rancher/pull/28413 (uses code for http proxy from an existing gist I found online, not sure of the licensing)

### DIFF
--- a/pkg/clusterrouter/proxy/proxy_server.go
+++ b/pkg/clusterrouter/proxy/proxy_server.go
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 	"strings"
 	"sync"
@@ -14,6 +15,8 @@ import (
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	"github.com/rancher/types/config/dialer"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/httpstream"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/proxy"
 	"k8s.io/client-go/rest"
 )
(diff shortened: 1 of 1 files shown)


pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -227,6 +231,13 @@ func (r *RemoteService) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		er.Error(rw, req, err)
 		return
 	}
+
+	if httpstream.IsUpgradeRequest(req) {
```
StrongMonkey, line 235: @kinarashah  What's the purpose of this code?

kinarashah: It's to check if it's a request that has `Connection`: `Upgrade` in its headers, so this applies only to websocket requests

kinarashah, line 235: @superseb @StrongMonkey both suggested this offline - going to add a check here so that it doesn't break normal setups without proxy if there's some issue with new proxy.  `isProxySet && IsUpgradeRequest `

kinarashah:
Discussed with @dramich and a couple of concerns/questions:  

- #28413 feels like the right direction to go, where we pass our dialer through proxy but the PR seems to be doing more than required, is there any other way where we can wrap it around native dialer? 
(I couldn't think of any way to do this, going to look into this more) 

- Right now RKE clusters don't come up, but this is going to affect all clusters? 
(the new proxy that I introduced would work for all clusters, but I am not sure what other cases we could be missing or what else needs to be handled) 

- Where else do we need to fix the proxy logic? This could be a bigger problem that might need a framework fix than fixing it on per case basis.

PR #28416, opened 2020-08-15
Adding validation for node count in a cluster

sowmyav27:


tests/validation/tests/v3_api/common.py
```diff
@@ -1065,7 +1065,8 @@ def validate_dns_record_deleted(client, dns_record, timeout=DEFAULT_TIMEOUT):
 
 
 def wait_for_nodes_to_become_active(client, cluster, exception_list=[],
-                                    retry_count=0):
+                                    retry_count=0,node_count=0,
```
bmdepesa, line 1068: Remove `check_node_count`, and rename `node_count` to `expected_node_count`, then check if `expected_node_count` is not 0 to perform the assertion.

sowmyav27: done

Issue #28370, opened 2020-08-12
Replace PRTB/CRTB UID with their name+ns for use as labels and annotations


mrajashree:
The projectroletemplatebinding and clusterroletemplatebinding UIDs are used as label keys for
clusterrolebindings and rolebindings in the management cluster as follows

CRTB.UID is the label key for a CRB, CRTB.UID=memberhsip-binding-owner
CRTB.UID is label key for the RB, CRTB.UID=crtb-in-project-binding-owner
PRTB.UID is the label key for a CRB, PRTB.UID=memberhsip-binding-owner
PRTB.UID is the label key for RB, PRTB.UID=memberhsip-binding-owner
PRTB.UID is also the label key for RB, PRTB.UID=prtb-in-cluster-binding-owner

PRTB/CRTB are also used as label values for clusterrolebindings and rolebindings in downstream user clusters

CRTB.UID is label value for CRB, authz.cluster.cattle.io/rtb-owner=CRTB.UID
PRTB.UID is label key for CRB, PRTB.UID=owner-user

Starting 2.5, instead of using UID, use namespace+name(+gvk if needed) as the label.
This is needed when during backup/restore UID is changed we don't lose the relation between PRTB/CRTB and RB/CRB respectively

sowmyav27:
**On 2.4.8** (single node install)

- Add a user1 as the cluster owner of the cluster 1
- Add user2 as project owner of cluster2
- Labels set on clusterrolebindings :

- kubectl get clusterrolebinding  -o yaml | grep -B7 -A33 "authz.cluster.cattle.io/rtb-owner: a16f26e7-206a-4840-8223-af45d807b43e"
```
labels:authz.cluster.cattle.io/rtb-owner: a16f26e7-206a-4840-8223-af45d807b43e
      
```
- Add a user as a project owner -- 
- kubectl get clusterrolebinding -o yaml | grep -B5 -A30 "00d2791f-d99a-4698-a73b-15d02a710200: owner-user"
```
labels:
      00d2791f-d99a-4698-a73b-15d02a710200: owner-user
```


**Upgrade to master-head - commit id: `89dd0fe25`**

CTRB:
```
labels:
      authz.cluster.cattle.io/rtb-owner-updated: c-btcsr_crtb-rgf7w
      cattle.io/creator: norman 
```

PRTB:
- kubectl get clusterrolebindings -o yaml | grep -B5 -A30 "00d2791f-d99a-4698-a73b-15d02a710200: owner-user"
[...]

PR #28423, opened 2020-08-17
Add labels to link PRTB/CRTB to RB/CRB  by namespace+name

mrajashree:
The projectroletemplatebinding and clusterroletemplatebinding UIDs are used as label keys for
clusterrolebindings and rolebindings in the management cluster as follows
* CRTB.UID is the label key for a CRB, CRTB.UID=memberhsip-binding-owner
* CRTB.UID is label key for the RB, CRTB.UID=crtb-in-project-binding-owner (in the namespace of each project in the cluster that the user has access to)
* PRTB.UID  is the label key for a CRB, PRTB.UID=memberhsip-binding-owner (with role c-xxxx-clustermember)
* PRTB.UID is the label key for RB, PRTB.UID=memberhsip-binding-owner (with role denoting the permissions in project)
* PRTB.UID is also the label key for RB, PRTB.UID=prtb-in-cluster-binding-owner (with role=exact roletemplate ID)

PRTB/CRTB are also used as label values for clusterrolebindings and rolebindings in user clusters
* CRTB.UID is label value for CRB, authz.cluster.cattle.io/rtb-owner=CRTB.UID
* PRTB.UID is label key for CRB, PRTB.UID=owner-user
* PRTB.UID is the label value for RBs, authz.cluster.cattle.io/rtb-owner=PRTB.UID

With the backup-restore-operator, there are two cases in which using UIDs as labels on rolebindings/clusterrolebindings will not give users correct permissions:
1. Migrating rancher to a new cluster: In this case, all PRTB/CRTBs will get newly created on the new server, and so their UIDs will be newly generated. But the rolebindings/clusterrolebindings will be still using the older UIDs as labels, and there's no way to update them.
2. Restoring on the same cluster with the intention to add back project/cluster members that were removed: An example of this use case:
   1. Add a user user1 as cluster-owner
   2. Take a backup
   3. Remove user1 from the cluster
   4. Restore from backup
In this case, the backup saves the RB/CRB linked to the CRTB with the label containing CRTB.UID, and once the user is removed, its CRTB will be deleted. So on restore, CRTB will be newly created so its UID changes, and there's no way to update the label on the corresponding CRB/RB.

As a solution, this commit adds a label for the RB/CRBs associated with any CRTB/PRTB with key "CRTB.Namespace_CRTB.Name" OR "PRTB.Namespace_PRTB.Name", in addition to the CRTB.UID and PRTB.UID respectively.
So 2.5 fresh install onwards the only label will be "CRTB.Namespace_CRTB.Name" OR "PRTB.Namespace_PRTB.Name"
But on upgraded setups this new label will be in addition to the label containing CRTB.UID or PRTB.UID to cover the rollback scenario
https://github.com/rancher/rancher/issues/28370
https://github.com/rancher/backup-restore-operator/issues/25

### DIFF
--- a/pkg/controllers/management/auth/crtb_handler.go
+++ b/pkg/controllers/management/auth/crtb_handler.go
@@ -11,6 +11,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
@@ -22,6 +23,8 @@ const (
 	rbByOwnerIndex              = "auth.management.cattle.io/rb-by-owner"
 	rbByRoleAndSubjectIndex     = "auth.management.cattle.io/crb-by-role-and-subject"
 	ctrbMGMTController          = "mgmt-auth-crtb-controller"
+	rtbLabelUpdated             = "auth.management.cattle.io/rtb-label-updated"
+	rtbCrbRbLabelsUpdated       = "auth.management.cattle.io/crb-rb-labels-updated"
 )
 
 var clusterManagmentPlaneResources = map[string]string{
@@ -63,12 +66,15 @@ func (c *crtbLifecycle) Updated(obj *v3.ClusterRoleTemplateBinding) (runtime.Obj
 	if err != nil {
 		return nil, err
 	}
+	if err := c.reconcileLabels(obj); err != nil {
+		return nil, err
+	}
 	err = c.reconcileBindings(obj)
 	return obj, err
 }
 
 func (c *crtbLifecycle) Remove(obj *v3.ClusterRoleTemplateBinding) (runtime.Object, error) {
-	if err := c.mgr.reconcileClusterMembershipBindingForDelete("", string(obj.UID)); err != nil {
+	if err := c.mgr.reconcileClusterMembershipBindingForDelete("", getRTBLabelKey(obj.ObjectMeta)); err != nil {
 		return nil, err
 	}
 	err := c.removeMGMTClusterScopedPrivilegesInProjectNamespace(obj)
@@ -142,7 +148,7 @@ func (c *crtbLifecycle) reconcileBindings(binding *v3.ClusterRoleTemplateBinding
 	if err != nil {
 		return err
 	}
-	if err := c.mgr.ensureClusterMembershipBinding(clusterRoleName, string(binding.UID), cluster, isOwnerRole, subject); err != nil {
+	if err := c.mgr.ensureClusterMembershipBinding(clusterRoleName, getRTBLabelKey(binding.ObjectMeta), cluster, isOwnerRole, subject); err != nil {
 		return err
 	}
 
@@ -168,8 +174,9 @@ func (c *crtbLifecycle) removeMGMTClusterScopedPrivilegesInProjectNamespace(bind
 	if err != nil {
 		return err
 	}
+	bindingKey := getRTBLabelKey(binding.ObjectMeta)
 	for _, p := range projects {
-		set := labels.Set(map[string]string{string(binding.UID): crtbInProjectBindingOwner})
+		set := labels.Set(map[string]string{bindingKey: crtbInProjectBindingOwner})
 		rbs, err := c.mgr.rbLister.List(p.Name, set.AsSelector())
 		if err != nil {
 			return err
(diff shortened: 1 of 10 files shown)


mrajashree:
I'm working on two things:
* Updating labels in downstream clusters when upgrading from 2.4 to master
* Updating tests that are failing because of the changes

pkg/controllers/managementuser/rbac/prtb_handler.go
```diff
@@ -190,6 +194,11 @@ func (m *manager) ownerExists(uid interface{}) (bool, error) {
 	return len(prtbs) > 0, err
 }
 
+func (m *manager) ownerExistsByNsName(nsAndName interface{}) (bool, error) {
```
dramich, line 197: This isn't used anywhere

pkg/controllers/managementuser/rbac/prtb_handler.go
```diff
@@ -297,3 +306,63 @@ func buildRule(resource string, verbs map[string]bool) rbacv1.PolicyRule {
...
+		if crb.Labels[rtbLabelUpgraded] == "true" {
+			continue
+		}
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			crbToUpdate, updateErr := p.m.clusterRoleBindings.Get(crb.Name, metav1.GetOptions{})
+			if updateErr != nil {
+				return updateErr
+			}
+			crbToUpdate.Labels[binding.Namespace+"_"+binding.Name] = owner
```
dramich, line 333: Create a helper function for the `binding.Namespace+"_"+binding.Name` part. If it ever needs to change that will be a bummer to find all the places it's used.

pkg/controllers/managementuser/rbac/prtb_handler.go
```diff
@@ -297,3 +306,63 @@ func buildRule(resource string, verbs map[string]bool) rbacv1.PolicyRule {
...
+
+func (p *prtbLifecycle) reconcilePRTBUserClusterLabels(binding *v3.ProjectRoleTemplateBinding) error {
+	/* Prior to 2.5, for every PRTB, following CRBs are created in the user clusters
+		1. PRTB.UID is the label key for a CRB, PRTB.UID=owner-user
+		2. PRTB.UID is the label value for RBs with authz.cluster.cattle.io/rtb-owner: PRTB.UID
+	Using this labels, list the CRBs and update them to add a label with ns+name of CRTB
+	*/
+
+	set := labels.Set(map[string]string{string(binding.UID): owner})
+	userCRBs, err := p.m.clusterRoleBindings.List(metav1.ListOptions{LabelSelector: set.String()})
```
dramich, line 318: Look into doing inequality in the list, that will stop from getting all the ones you already updated back in the response https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement

pkg/controllers/managementuser/rbac/reconcile_roletemplate.go
```diff
@@ -75,7 +75,8 @@ func (m *manager) reconcileProjectAccessToGlobalResources(binding *v3.ProjectRol
 		}
 	}
 
-	rtbUID := string(binding.UID)
+	//oldRtbUID := string(binding.UID)
```
dramich, line 78: Delete if not needed

pkg/controllers/managementuser/rbac/prtb_handler.go
```diff
@@ -297,3 +306,63 @@ func buildRule(resource string, verbs map[string]bool) rbacv1.PolicyRule {
...
+	if err != nil {
+		return err
+	}
+
+	for _, crb := range userCRBs.Items {
+		if crb.Labels[rtbLabelUpgraded] == "true" {
+			continue
+		}
+
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
dramich, line 328: Nice find, this is really cool 🥇

pkg/controllers/managementuser/rbac/prtb_handler.go
mrajashree: good point! will add this in the next commit

pkg/controllers/managementuser/rbac/prtb_handler.go
mrajashree: ah I hadn't pushed the above changes that use this yet

pkg/controllers/management/auth/crtb_handler.go
```diff
@@ -183,3 +190,78 @@ func (c *crtbLifecycle) removeMGMTClusterScopedPrivilegesInProjectNamespace(bind
...
+			if updateErr != nil {
+				return updateErr
+			}
+			crbToUpdate.Labels[bindingKey] = membershipBindingOwner
+			crbToUpdate.Labels[rtbLabelUpdated] = "true"
+			_, err := c.mgr.crbClient.Update(crbToUpdate)
+			return err
+		})
+		if retryErr != nil {
+			returnErr = append(returnErr, retryErr)
```
dramich, line 228: Consider using multierror in these cases, it helps formatting

dramich (approved):

PR #28433, opened 2020-08-18
Adding additional validation tests for hosted EKS cluster

sowmyav27:
Adding two test cases
- Edit cluster
- Delete cluster

Including validations from EKS console end.

### DIFF
--- a/tests/validation/tests/v3_api/test_hosted_eks_cluster.py
+++ b/tests/validation/tests/v3_api/test_hosted_eks_cluster.py
@@ -1,6 +1,8 @@
 import os
-from .common import get_user_client, validate_cluster, \
-    random_test_name, AWS_SSH_KEY_NAME
+from .common import  get_user_client
+from .common import random_test_name
+from .common import  validate_cluster
+from .common import  wait_for_cluster_delete
 from .test_create_ha import resource_prefix
 from lib.aws import AmazonWebServices
 import pytest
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_hosted_eks_cluster.py
```diff
@@ -1,9 +1,10 @@
 import os
 from .common import get_user_client, validate_cluster, \
-    random_test_name, AWS_SSH_KEY_NAME
+    random_test_name, AWS_SSH_KEY_NAME, wait_for_cluster_delete
```
izaac, line 3: can we add a line for each import from `.common`? this multi import line is getting a bit long.

tests/validation/tests/v3_api/test_hosted_eks_cluster.py
```diff
@@ -187,3 +223,177 @@ def get_logging_types():
...
+               == eks_nodegroup["nodegroup"]["diskSize"], \
+            "diskSize is incorrect on the nodes"
+        # check ec2SshKey
+        if "ec2SshKey" in nodegroup.keys():
+            assert nodegroup["ec2SshKey"] \
+                   == eks_nodegroup["nodegroup"]["remoteAccess"]["ec2SshKey"], \
+                "Ssh key is incorrect on the nodes"
+
+
+def wait_for_cluster_delete_eks(cluster_name, timeout=DEFAULT_TIMEOUT_EKS):
```
izaac, line 391: Could we use an AWS (boto) waiter instead of our own custom?

there are a few waiters, I know we're using one for `nodegroups` there might be one too for cluster 

`waiter_ng = self._eks_client.get_waiter('nodegroup_deleted')`

tests/validation/tests/v3_api/test_hosted_eks_cluster.py
sowmyav27: This is within the 80 character limit. Do you think that should be ok ?

tests/validation/tests/v3_api/test_hosted_eks_cluster.py
izaac: Oh you're deleting from Rancher, this might not be possible but I'm not completely sure.

tests/validation/tests/v3_api/test_hosted_eks_cluster.py
izaac: Yeah I noticed it is in conformance of the line length but the import line is just getting multiline split and not sorted alphabetically, and an all caps variable in the middle. It just looks a bit odd.

PR #28445, opened 2020-08-18
[2.4] Changes for detecting admin & cluster ownership

ryansann:
A role template with cluster ownership permissions can be either:
1. Builtin `Cluster Owner`
2. Have rules for "own" on "clusters" resources
3. Inherit from a role template that has 1. or 2.

When a user is added to a cluster with any of 1, 2, or 3, there should not be any namespaced roles created, this PR fixes an issue where that was happening and leading to errors. 

When a user is created and associated to a role, if they are associated to a global role with admin permissions that isn't admin, they are not getting the proper permissions / capabilities. Changes to globalrole_handler.go fix this issue. 

Previous related PR: 
- rancher/rancher#27978

### DIFF
--- a/pkg/controllers/user/rbac/globalrole_handler.go
+++ b/pkg/controllers/user/rbac/globalrole_handler.go
@@ -1,10 +1,12 @@
 package rbac
 
 import (
+	"github.com/rancher/norman/types/slice"
 	"github.com/rancher/rancher/pkg/rbac"
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 	rbacv1 "github.com/rancher/types/apis/rbac.authorization.k8s.io/v1"
 	"github.com/rancher/types/config"
+	"github.com/sirupsen/logrus"
 	v12 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


ryansann:
One thing I am not positive of here is whether the logic for detecting a Global Role with admin permissions is correct. Currently the `admin` role's rules looks like this: 
```yaml
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - '*'
  verbs:
  - '*'
```
The changes I made in pkg/controllers/user/rbac/globalrole_handler.go look for the first rule in order to detect if roles other than `admin` should be treated as admin. Perhaps I should be checking for the second rule as well? Input appreciated

pkg/controllers/user/rbac/globalrole_handler.go
```diff
@@ -70,9 +84,31 @@ func (c *grbHandler) sync(key string, obj *v3.GlobalRoleBinding) (runtime.Object
...
+	if err != nil {
+		return false, err
+	}
+
+	// global role is admin
+	if gr.Name == "admin" {
+		return true, nil
+	}
+
+	for _, rule := range gr.Rules {
```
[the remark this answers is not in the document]
ryansann: I changed this to check for both rules

pkg/controllers/user/rbac/globalrole_handler.go
```diff
@@ -70,9 +84,42 @@ func (c *grbHandler) sync(key string, obj *v3.GlobalRoleBinding) (runtime.Object
...
 
+// isAdminRole detects whether a GlobalRole has admin permissions or not.
+func (c *grbHandler) isAdminRole(rtName string) (bool, error) {
+	gr, err := c.grLister.Get("", rtName)
+	if err != nil {
+		return false, err
+	}
+
+	// global role is admin
+	if gr.Name == "admin" {
```
mrajashree, line 99: can we also add a check for "Builtin" like we do for cluster-owner?

ryansann: Ahh yes, good point, let me add that

Issue #27192, opened 2020-05-21
UI and API slowness in Rancher v2.4.3


rosskirkpat:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

When using rancher as the admin user on 2.4.3, it is now taking a considerably longer time than in 2.3 due to the UI waiting on `/v3/clusters?limit=-1&sort=name` endpoint on page loads/refreshes. This call can take up to 3 minutes to complete on certain occasions.

The entire cluster list can be queried directly via CRDs in less than 2 seconds.
```
$ time kubectl get cluster -o json > /dev/null
kubectl get cluster -o json > /dev/null 1.17s user 0.08s system 94% cpu 1.315 total
```
But the API call takes more than a minute to complete.
```
$ time curl -fSso /dev/null -u “token-***:***” ‘https:/xxxx.com/v3/clusters?limit=-1&sort=name’
curl -fSso /dev/null -u 0.02s user 0.01s system 0% cpu 1:17.02 total
```

env:
rancher v2.4.3
k8s v1.17.2
3 node control plane, 5 node etcd v3.3.8

gzrancher/rancher#10245

sowmyav27:
**Reproduced the issue on 2.4.5**

- On 2.3.8 - Deploy 200 clusters. Load the UI page. UI loads all the clusters in about 1 or 2 seconds. 
- Upgrade to 2.4.5
- Load the UI page. UI loads all the clusters in about 39 seconds.
- time curl -fSso /dev/null -u token-<token> "https://<rancher-server>/v3/clusters" --insecure -- took about `real	0m39.726s`

sowmyav27:
**Verified on 2.4-head -commit id: `beca318de`**

- Deploy 200 empty custom clusters. Load the UI page. UI loads all the clusters in about 39 seconds.
- Upgrade to 2.4-head
- Load the UI page. UI loads all the clusters in about 1.2 seconds.
- time curl -fSso /dev/null -u token- "https:///v3/clusters" --insecure -- took about `real	0m1.272s`

PR #28458, opened 2020-08-19
Update auth check to use cache

dramich:
Problem:
When listing large amounts of clusters (200) the call takes approx 40
seconds to complete vs 1 second when using kubectl directly

Solution:
The formatter uses a hard API call to check auth on the clusterTemplate
perms for the user. This adds 200ms per cluster when running locally.
Swap this call to use the cached access through the request
AccessControl

Issue: https://github.com/rancher/rancher/issues/27192

### DIFF
--- a/pkg/api/customization/cluster/formatter.go
+++ b/pkg/api/customization/cluster/formatter.go
@@ -6,7 +6,6 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
-	gaccess "github.com/rancher/rancher/pkg/api/customization/globalnamespaceaccess"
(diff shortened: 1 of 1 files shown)


dramich:
The test file changes are a lot of formatting changes, the real change is the test added https://github.com/rancher/rancher/pull/28458/files#diff-50d971bd3fe396984fcd0f35042a4119R757

dramich:
Dropped the test (running into timing issues around namespaces) to get this merged and over to QA.

prachidamle:
@sangeethah Things to check is access to "SaveAsTemplate" Cluster action that results in a new template and revision created.
Only global admin and cluster-admins having "Create ClusterTemplate" and "Create ClusterTemplateRevision" roles assigned should be able to do this action.

PR #28494, opened 2020-08-20
Update norman

rmweir:


tests/integration/suite/test_cluster_defaults.py
```diff
@@ -157,7 +157,7 @@ def test_import_initial_conditions(admin_mc, remove_resource):
     cluster = admin_mc.client.create_cluster(name=random_str())
     remove_resource(cluster)
 
-    assert cluster.conditions is None
```
paynejacob, line 160: are you sure this is never None?  `assert not cluster.conditions` might be safer

rmweir: good catch, fixed

PR #28516, opened 2020-08-22
[2.4] Fix error due to cluster owner rules being applied to namespaced roles

ryansann:
These changes fix an issue where a cloned cluster owner RoleTemplate was not being recognized as a cluster owner role before  the `clusters+own` rule was added via the API. This was leading to an error like:
```
... rules[0].nonResourceURLs: Invalid value: []string{"*"}: namespaced rules cannot apply to non-resource URLs
```
which was backing up the workqueue for the RoleTemplate in question and leading to delays in processing time. 

Issue:
- rancher/rancher#28094

### DIFF
--- a/pkg/controllers/user/rbac/handler_base.go
+++ b/pkg/controllers/user/rbac/handler_base.go
@@ -208,11 +208,9 @@ func (m *manager) createClusterRole(rt *v3.RoleTemplate) error {
 }
 
 func (m *manager) ensureNamespacedRoles(rt *v3.RoleTemplate) error {
-	isClusterOwner, err := m.isClusterOwner(rt.Name)
-	if err != nil {
+	// if the RoleTemplate has cluster owner rules, don't update Roles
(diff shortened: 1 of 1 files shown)


pkg/controllers/user/rbac/handler_base.go
```diff
@@ -237,42 +235,24 @@ func (m *manager) ensureNamespacedRoles(rt *v3.RoleTemplate) error {
...
-				return true, nil
-			}
+		// cluster + own rule that indicates cluster owner permissions
+		if slice.ContainsString(rule.Resources, "clusters") && slice.ContainsString(rule.Verbs, "own") {
+			return true, nil
 		}
-	}
-
-	isOwner := false
-	if len(rt.RoleTemplateNames) > 0 {
```
mrajashree, line 261: why do we need to remove this?

ryansann: I was thinking that Cluster RoleTemplates can't have inherited role templates due to the option not being available in the UI, but they can if you add them through the api, so I will add this logic back in. Good catch

Issue #28422, opened 2020-08-17
[KEv2]: EKSv2 Add syncing


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Description:**
If fields for an EKS cluster are updated in the EKS console, they should be displayed in rancher and should not be over written on update unless it is the field being changed.

**Explanation of approach:**
All fields in the EKSConfig will be nillable. If a field is nilled then it is considered "unmanaged". Fields that are "managed" have been modified in Rancher at least once before.

There is a new field on status called UpstreamSpec. This is updated every 5 minutes and represents the state of the cluster in EKS. Most or all fields will not be nil. When this is updated it will check to see if the EKS cluster is currently being updated by Rancher. This is evaluated by checking if spec and applied spec are equal. If it is not updating then all fields in EKSConfig that are no nil will be overwritten with the fields in upstreamSpec. If it is updating then this step is skipped.

The UI will display fields from the upstreamSpec. If a field is edited then it will be sent in the update.

sowmyav27:
**Validated on master-head - commit id: `e919796c8`**

- Deployed a hosted EKS cluster through Rancher
- Verified the following scenarios and issues:
```
- Logging set in EKS console does not sync back to Rancher
- Cluster level Tags set do not reflect
- Endpoint-access-update does NOT reflect
- Labels and tags on Nodegroups, does NOT reflect
- Cluster is in Updating state in EKS, in Rancher, the state is Active state
- Scale up the nodes in a nodegroup - added nodes reflect in the Rancher nodes page but on an edit cluster, the node group is NOT available
- Adding new nodegroup - added nodes reflect in the Rancher nodes page but on an edit cluster, the node group is NOT available
```

sowmyav27:
**Verified on master-head - commit id: `33f755f18`**

- Basic syncing use cases work for a hosted EKS and an imported EKS cluster
- Tracking issues related to this separately.
- Closing this issue

PR #28524, opened 2020-08-23
Add upstream syncing

rmweir:
Added upstream syncing and the concept of managed fields

**Issue:**
https://github.com/rancher/rancher/issues/28422


paynejacob (approved):

dramich (changes requested):
I think a better approach would be to use the daemon/cron approach. Using the controller, every time the cluster is touched it will end up in the enqueueAfter bucket and with a 5 min window, that could be a lot of the same cluster hanging out.

pkg/controllers/management/eks/eks_upstream_refresh.go
```diff
@@ -0,0 +1,178 @@
...
+		logrus.Infof("updating cluster [%s], upstream change detected", cluster.Name)
+		cluster = cluster.DeepCopy()
+		cluster.Status.EKSStatus.UpstreamSpec = upstreamSpec
+		cluster, err = e.clusterClient.Update(cluster)
+		if err != nil {
+			return cluster, err
+		}
+	}
+
+	if !reflect.DeepEqual(cluster.Spec.EKSConfig, cluster.Status.AppliedSpec.EKSConfig) {
```
dramich, line 89: Would it make sense to do this check first before attempting to pull the EKS config from AWS?

dramich (changes requested):

pkg/controllers/management/wrangler.go
```diff
@@ -14,5 +14,6 @@ import (
 func RegisterWrangler(ctx context.Context, wranglerContext *wrangler.Context, management *config.ManagementContext, manager *clustermanager.Manager) error {
 	k3sbasedupgrade.Register(ctx, wranglerContext, management, manager)
 	eks.Register(ctx, wranglerContext, management)
+	eks.RegisterRefresh(wranglerContext)
```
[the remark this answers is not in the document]
rmweir: providerrefresh does not use wrangler context but I have moved it to what seems to be a near equivalent for wrangler

pkg/controllers/management/eks/eks_upstream_refresh.go
rmweir: No, because we still want to write the upstream state to the status we just do not want to overwrite spec.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -7,6 +7,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
+
```
dramich, line 12: Extra line

pkg/controllers/management/wrangler.go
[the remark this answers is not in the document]
rmweir: nvm was looking at the steve start logic

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -0,0 +1,195 @@
...
+	eksUpstreamRefresher.clusterCache.AddIndexer(isEKSIndexer, func(obj *apimgmtv3.Cluster) ([]string, error) {
+		if obj.Spec.EKSConfig == nil {
+			return []string{}, nil
+		}
+		return []string{"true"}, nil
+	})
+
+	schedule, err := cron.ParseStandard(settings.EKSUpstreamRefreshCron.Get())
+	if err != nil {
+		logrus.Errorf("Error parsing EKS upstream cluster refresh cron: %v", err)
```
dramich, line 53: Can we add that the cron will not start?

dramich (approved):

dramich (dismissed):

deniseschannon (approved):

Issue #27487, opened 2020-06-08
Update rancher-images.txt generation for airgap


cbron:
With our movement towards cloud-native-packages / dev-charts, we need to use it to generate rancher-images.txt in a similar manner to how we do today with system-charts. 

The new generator must have a feature that allows certain packages to provide their images via a standard text file, rather than through a search. The package must provide the text file to the generator through a standard naming convention. For example if the generator see's _"images.amd.txt"_ or similar it will use that and skip the search. 

Related: 
* https://github.com/rancher/rancher/issues/19366
* https://github.com/rancher/rancher/issues/22865 
* https://github.com/rancher/rancher/issues/27403

cbron:
Looks like K3s images are missing

jiaqiluo:
This is validated in rancher: `v2.5.0-alpha3`, docker install in airgap environment

k3s images are bundled in the rancher image.

cbron:
Not all the k3s images are bundled with the docker repo. I think we need those to upgrade downstream clusters on a new release, is that right @davidnuzik ? 
That being said I'm not sure if this ticket covers that side of the house or not. Might create separate issue for that.

davidnuzik:
I'm not sure @cbron - my understanding was all I need to do is get k3s in KDM like as per https://github.com/rancher/kontainer-driver-metadata/pull/356 and the k3s images should show up. We are just waiting on approval from @kinarashah to get that in. I've requested re-review as it may not have been clear in the PR we are waiting.
cc: @cjellick @galal-hussein

cbron:
My main question is where are we tracking airgap testing for k3s upgrade since its not currently working. @jiaqiluo @sangeethah do we usually test for that ?

PR #28554, opened 2020-08-25
Update rancher-images.txt generation for airgap

PennyScissors:
**This PR includes the following:**

The logic behind generating a file listing all images we use (`rancher-images.txt`) for airgap has been updated to support extracting the names of images used by charts in `rancher/charts` in a similar way it is done for `rancher/system-charts`. The main difference is system-charts is not a helm repo so the charts are contained in local directories and a virtual index is created to walk through them, whereas dev charts is a helm repo with an index we can use, but the charts are packaged into tgz files that will need to be extracted before the walk.
A flag `isLocalHelmRepo` has been added to distinguish between the two cases, and in the case we have a local helm repo (flag value is true), the tgz of the latest version of a chart will be extracted before the walk.

**IMPORTANT:**
Since `rancher/charts` will not be a catalog in ECM, the option to generate a `rancher-images.txt` when you hit the version button in the bottom left corner of the ember UI will not include images from `rancher/charts`.

This PR will not pass the CI tests because there are charts using the following images which aren't mirrored yet and they
prevent the generator from creating a `rancher-images.txt` successfully since they are treated as errors.

Images in `rancher/charts` in need of mirrors:
- ibuildthecloud/fleet:v0.3.0-alpha2                         
- banzaicloud/logging-operator:3.4.0                         
- jettech/kube-webhook-certgen:v1.2.1                        
- docker.io/jimmidyson/configmap-reload:v0.3.0               
- curlimages/curl:7.70.0                                     
- busybox:1.31.1                                             
- kiwigrid/k8s-sidecar:0.1.151                               
- quay.io/prometheus/node-exporter:v1.0.0                    
- quay.io/coreos/prometheus-config-reloader:v0.38.1          
- jettech/kube-webhook-certgen:v1.0.0                        
- k8s.gcr.io/hyperkube:v1.16.12                              
- directxman12/k8s-prometheus-adapter-amd64:v0.6.0           
- ibuildthecloud/fleet-agent:v0.3.0-alpha2                   
- quay.io/kubernetes-ingress-controller/nginx-                                                           
- grafana/grafana:7.0.5                                      
- quay.io/coreos/kube-state-metrics:v1.9.7                   
- quay.io/coreos/prometheus-operator:v0.38.1                 
- quay.io/prometheus/prometheus:v2.18.1                      
- squareup/ghostunnel:v1.5.2                                 
- k8s.gcr.io/defaultbackend-amd64:1.5                        
- quay.io/prometheus/alertmanager:v0.20.0  

**Related PR:** https://github.com/rancher/rancher/issues/27487

### DIFF
--- a/pkg/catalog/helm/helm.go
+++ b/pkg/catalog/helm/helm.go
@@ -47,16 +47,17 @@ var (
 )
 
 type Helm struct {
-	LocalPath   string
-	IconPath    string
-	catalogName string
-	Hash        string
-	Kind        string
-	url         string
-	branch      string
-	username    string
-	password    string
-	lastCommit  string
+	IsLocalHelmRepo bool
+	LocalPath       string
+	IconPath        string
+	catalogName     string
+	Hash            string
+	Kind            string
+	url             string
+	branch          string
+	username        string
+	password        string
+	lastCommit      string
 }
 
 func (h *Helm) lock() {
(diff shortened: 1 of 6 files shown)


pkg/catalog/helm/helm.go
```diff
@@ -239,6 +239,41 @@ func (h *Helm) fetchTgz(helmURL string) ([]v32.File, error) {
 	return files, nil
 }
 
+func (h *Helm) fetchTgz(helmURL string) ([]v32.File, error) {
+	var files []v32.File
+	logrus.Debugf("Helm fetching file %s", helmURL)
+
+	if h.IsLocalHelmRepo {
+		tgzPath := filepath.Join(h.LocalPath, helmURL)
+		r, err := os.Open(tgzPath)
```
cbron, line 248: On first glance I don't like the idea of adding this into the helm package for rancher itself, opening files on the local cluster seems like it could be security hole. I haven't looked to deep into this but how hard would it be to keep all this logic in the separate image go program ? 

Can you just open up all the tgz files that exist in the assets and read them without caring about the index.yaml ? Or do you need to know sub-charts and stuff.

PennyScissors: @Caleb The index.yaml is used to figure out the latest version of a chart and the location of its tgz. 
What's happening here is we open the tgz to provide `readTgz()` with a reader so that it can be extracted to a separate directory. Then we update the version `version.Dir` with the path to that directory which is used in the image package to pick images from all values.yaml.
I will need to investigate a bit more, but I think we can move this logic to image. Although this will result in duplicated code for reading and extracting a tgz to a separate folder.

cbron: I think duplication would be fine, we don't have to recreate whole helm dir package, just have to make it work for this use-case.
@StrongMonkey do you have opinion here ?

StrongMonkey: @cbron I feel like it is ok, in helm we store these content in local disk so that we don't have to re-fetch it from internet. Besides that this code will only be called in image gen code.

StrongMonkey:
@PennyScissors After reviewing this PR, I feel like this might be too much regarding to the changes to image generator code. Can we do the following? 
1. In https://github.com/rancher/rancher/blob/78963d00c4ab87af90aad4ebbec21ab6a5282bd4/scripts/package#L44, after you clone the repo, go to `assets` folder and untar all the tarball in its current directory using shell.
2. Add `TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go build/charts/assets $IMAGE $AGENT_IMAGE` in the end of `scripts/package`.

PennyScissors:
Closed in favor of:  #28619

PR #28590, opened 2020-08-27
2.5 alpha1 - bump ui and chart versions

cbron:
### branches

charts: https://github.com/rancher/charts/tree/v2.5-preview1
partner-charts: https://github.com/rancher/partner-charts/tree/v2.5-preview1
system-charts: https://github.com/rancher/system-charts/tree/v2.5-preview1
kdm: https://github.com/rancher/kontainer-driver-metadata/tree/v2.5-preview1


package/Dockerfile
```diff
@@ -104,9 +104,9 @@ RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
     chmod +x /usr/bin/tini /usr/bin/telemetry /usr/bin/k3s /usr/bin/kubectl /usr/bin/channelserver && \
     mkdir -p /var/lib/rancher-data/driver-metadata
 
-ENV CATTLE_UI_VERSION 2.4.10
+ENV CATTLE_UI_VERSION 2.5.0-alpha1
 ENV CATTLE_DASHBOARD_UI_VERSION latest
```
deniseschannon, line 108: Don't we still need to set the 2.5.0-alpha1 tag for the UI version?

cbron: Thought we decided on latest because it doesn't update once you start up, its always stuck there. Unless you just meant revert after this alpha is out.

cbron: pinned

PR #28614, opened 2020-08-28
revert alpha1 branch changes

cbron:


package/Dockerfile
```diff
@@ -105,7 +105,7 @@ RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
     mkdir -p /var/lib/rancher-data/driver-metadata
 
 ENV CATTLE_UI_VERSION 2.5.0-alpha1
```
deniseschannon, line 107: if the logic has been dropped to no longer allow to pick up from `latest2` does this also need to change so that ember changes can be seen in ember?

cbron: That logic still exists, it's just in a different spot. As I understand it this value would still supply the static files if you turned `ui-offline-preferred` to true.

deniseschannon: I don't quite understand what this needs to be or how it will affect QA, but we can move forward to get the branches updated.

PR #28619, opened 2020-08-28
Add support for rancher/charts in rancher-images.txt generator

PennyScissors:
**This PR includes the following:**

Add support in image list generator for picking images in rancher/charts so they are included in the generated `rancher-images.txt`. The `rancher/charts` repo is cloned and pre-processed so that images are picked the same way is done for system charts. The pre-processing includes deleting the index files so that a virtual one is built, and finding and extracting the latest version's tgz of each chart. 

Functions in the image and main package were modified to take in an extra path to a local `rancher/charts` repo.
The main reason behind doing that versus just calling main.go twice (one for each chart repo) is that we would be adding in unnecessary complexity because we overwrite the generated images text file, and images are normalized before is created. 

**IMPORTANT:**
Since rancher/charts will not be a catalog in ECM, the option to generate a rancher-images.txt when you hit the version button in the bottom left corner of the ember UI will not include images from rancher/charts.

This PR will not pass the CI tests because there are charts using the following images which aren't mirrored yet and they
prevent the generator from creating a rancher-images.txt successfully since they are treated as errors.

Images in rancher/charts in need of mirrors:
```
location: nginx-ingress/values.yaml                                       
- quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.30.0        
- jettech/kube-webhook-certgen:v1.0.0                                          
- k8s.gcr.io/defaultbackend-amd64:1.5

location: rancher-kiali-server/values.yaml                         
- kiali/kiali:v1.22.0

location: rancher-logging/values.yaml                                   
- banzaicloud/logging-operator:3.4.0

location: rancher-monitoring/charts/grafana/values.yaml              
- kiwigrid/k8s-sidecar:0.1.151                                                 
- grafana/grafana:7.0.5                                                        
- curlimages/curl:7.70.0                                                       
- busybox:1.31.1

location: rancher-monitoring/charts/kube-state-metrics/values.yaml
- quay.io/coreos/kube-state-metrics:v1.9.7

location: rancher-monitoring/charts/prometheus-adapter/values.yaml
- directxman12/k8s-prometheus-adapter-amd64:v0.6.0

location: rancher-monitoring/charts/prometheus-node-exporter/values.yaml
- quay.io/prometheus/node-exporter:v1.0.0

location: rancher-monitoring/values.yaml
- jettech/kube-webhook-certgen:v1.2.1
- directxman12/k8s-prometheus-adapter-amd64:v0.6.0
```
**Issue** #27487
**Previous PR:** https://github.com/rancher/rancher/pull/28554

### DIFF
--- a/pkg/image/resolve_test.go
+++ b/pkg/image/resolve_test.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -18,22 +21,33 @@ import (
 )
 
 const (
-	testSystemChartBranch = "dev"
-	testSystemChartCommit = "7c183b98ceb8a7a50892cdb9a991fe627a782fe6"
+	testChartRepoURL       = "https://github.com/rancher/charts.git"
+	testChartBranch        = "dev-v2.5"
+	testChartCommit        = "3887f667bf6d8663032b7bd298d74d46018183ca"
+	testSystemChartRepoURL = "https://github.com/rancher/system-charts.git"
+	testSystemChartBranch  = "dev"
+	testSystemChartCommit  = "7c183b98ceb8a7a50892cdb9a991fe627a782fe6"
 )
 
 func TestFetchImagesFromCharts(t *testing.T) {
-	systemChartPath := cloneTestSystemChart(t)
+	systemChartPath := cloneChartRepo(t, testSystemChartRepoURL, testSystemChartBranch, testSystemChartCommit)
+	chartRepoPath := cloneChartRepo(t, testChartRepoURL, testChartBranch, testChartCommit)
+	chartPath := prepareTestCharts(t, chartRepoPath)
 
-	bothImages := []string{
+	systemChartBothImages := []string{
 		"rancher/fluentd:v0.1.16",
 	}
-	linuxImagesOnly := []string{
+	systemChartLinuxImagesOnly := []string{
 		"rancher/prom-alertmanager:v0.17.0",
 	}
-	windowsImagesOnly := []string{
+	systemChartWindowsImagesOnly := []string{
 		"rancher/wmi_exporter-package:v0.0.2",
 	}
+	chartLinuxImagesOnly := []string{
+		"rancher/istio-kubectl:1.5.8",
+		"rancher/opa-gatekeeper:v3.1.0-rc.1",
+	}
+	chartWindowsImagesOnly := []string{}
 
 	testCases := []struct {
 		caseName               string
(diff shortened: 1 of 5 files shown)


pkg/image/resolve_test.go
```diff
@@ -317,6 +345,76 @@ func cloneTestSystemChart(t *testing.T) string {
...
+		t.Fatalf("failed to checkout chart: %s, %v", string(gitCheckoutOutput), err)
+	}
+	fmt.Printf("Checked out charts to %s\n", testChartCommit)
+
+	prepareTestCharts(t, tempDir)
+
+	return tempDir
+}
+
+func prepareTestCharts(t *testing.T, chartDir string) {
```
cbron, line 387: unfortunate that we can't test the extractor bash script in here somehow

scripts/package
```diff
@@ -38,4 +49,16 @@ if [ ! -d build/system-charts ]; then
...
+
+if [ ! -d build/charts ]; then
+    mkdir -p build
+    git clone --branch $CHART_DEFAULT_BRANCH https://github.com/rancher/charts build/charts
+
+    # Iterate through chart directories and execute callback to extract latest version tgz
+    find build/charts/assets -type d -maxdepth 1 -exec bash -c 'extract_latest_version_tgz {}' \;
+
+    # Remove index to force building a virtual index like system charts
+    rm -f build/charts/index.yaml build/charts/assets/index.yaml
```
cbron, line 61: Didn't realize the walker script actually generated a virtual index

scripts/package
```diff
@@ -38,4 +49,16 @@ if [ ! -d build/system-charts ]; then
...
+fi
+
+TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go build/system-charts build/charts $IMAGE $AGENT_IMAGE
```
cbron, line 64: I'm assuming this whole PR does actually work, do you know what happens if an image exists in both system-charts and charts ? If so we should unique-ify it, but if thats hard it isn't an immediate need, shouldn't be hard though.

pkg/image/resolve_test.go
```diff
@@ -317,6 +345,76 @@ func cloneTestSystemChart(t *testing.T) string {
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.TODO(), 2*time.Minute)
+	defer cancel()
+
+	gitCloneCmd := exec.CommandContext(ctx, "git", "clone",
+		"--branch", testChartBranch,
+		"--single-branch",
+		"https://github.com/rancher/charts.git",
+		tempDir,
```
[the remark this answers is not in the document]
PennyScissors: Agreed, I'll update in next commit.

pkg/image/resolve_test.go
PennyScissors: I thought the same thing, but I figured this would be good enough considering the circumstances.

scripts/package
PennyScissors: We build a virtual index (in helm package) when one isn't found. When building it we set `version.Dir` for each version to it's location, which is later used to pick the images. Since charts is a helm repo, `version.Dir` is empty and `version.URLs` contains the location of the tgz instead. This was the cleanest way to reuse all the logic for system-charts. A little hacky so left a comment for the future.

scripts/package
PennyScissors: We gather all images and append them to a list. Then they all get normalized (Check for non-mirrors, remove duplicates, and sort ) and written to a text file. Since they all get normalized together, this shouldn't be a problem.

scripts/package
```diff
@@ -1,6 +1,17 @@
 #!/bin/bash
 set -e
 
+extract_latest_version_tgz() {
+    # Takes in a path to a chart's directory, finds the latest version's tgz and extracts it
+    # All crd tgz are ignored
+    # Max depth is set to prevent extracting a tgz contained within another tgz, which is the case for charts containing a helm repo
```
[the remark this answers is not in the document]
PennyScissors: This should cover sub-charts as long as they are not packaged into a tgz like `parent-chart/charts/subchart.tgz`, which I don't think we do since all the sub-charts I found were in their own separate directories as `parent-chart/charts/subchart/` after extraction. 
The main reason I implemented this constraint is because OPA's gatekeeper has a helm repo inside the chart. It has an index.yaml and a tgz which will get extracted and non-mirrored images will get picked. Right now, our process would only allow deleting that tgz through the patch which will pretty much make it unreadable.

StrongMonkey (dismissed):
I like the current approach right now, much cleaner. :fire: thanks for all the work!

PennyScissors:
@StrongMonkey Thank you!! I appreciate it.

cbron (approved):

pkg/image/resolve_test.go
```diff
@@ -18,49 +21,75 @@ import (
...
+	chartLinuxImagesOnly := []string{
+		"rancher/istio-kubectl:1.5.8",
+		"rancher/opa-gatekeeper:v3.1.0-rc.1",
+	}
+	chartWindowsImagesOnly := []string{}
 
 	testCases := []struct {
 		caseName               string
 		inputPath              string
+		isLocalHelmRepo        bool
```
StrongMonkey, line 55: Do you still need this?

PennyScissors: Good catch, I'll remove it.

cbron (approved):

StrongMonkey (approved):

cbron (approved):

Issue #28135, opened 2020-07-29
Airgap


maggieliu:
* Need to support airgap for dev-charts 
* How to handle partner charts

StrongMonkey:
After talking to darren, this should actually be done through client code. For catalog v2, rancher just sends helm commands from clients and run them in a pod. Client needs to read systemDefaultProperties and set extra values in helm command `--set systemDefaultRegistry=foo`.

PR #28621, opened 2020-08-28
Add system global registry to helm chart

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/28135

cbron (approved):
I don't have wider context on catalogv2, but lgtm. Only nit is that there might be several things to inject long term, so we could move this to a standalone function like how `extraArgsFuncs` worked on v1.

StrongMonkey:
@cbron After talking to darren, this is actually the wrong place to update. For catalog v2, rancher just sends helm commands from clients and run them in a pod. So theoratically this could be added in client code(UI) not backend code.Client will need to send extra set value `--set systemDefaultRegistry=foo`.

PR #28641, opened 2020-08-30
WIP: Enable local docker install to run pods

cbron:
This allows running pods on the local cluster.

Todo: 

* vendoring norman after merge

Need to doc 2 things: 

* Setting local to false disables features in docker install
* You must run with privileged flag


go.mod
```diff
@@ -9,6 +9,7 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20200712062324-13d1f37d2d77
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.3
+	github.com/rancher/norman => github.com/cbron/norman v0.0.0-20200831152730-0814837e9375
```
cbron, line 12: todo

package/Dockerfile
```diff
@@ -117,7 +117,6 @@ ENV AUDIT_LOG_MAXAGE 10
 ENV AUDIT_LOG_MAXBACKUP 10
 ENV AUDIT_LOG_MAXSIZE 100
 ENV AUDIT_LEVEL 0
-VOLUME /var/log/auditlog
```
cbron, line 120: I'm assuming we can remove this volume because it's covered below, not sure if this has repercussions for upgrade scenario.

main.go
```diff
@@ -75,8 +75,8 @@ func main() {
 		},
 		cli.StringFlag{
 			Name:        "add-local",
-			Usage:       "Add local cluster (true, false, auto)",
-			Value:       "auto",
+			Usage:       "Add local cluster (true, false)",
+			Value:       "true",
```
ibuildthecloud, line 79: I just want to confirm that if for some reason "auto" is passed as a value Rancher will treat that as "true."  For example an upgrade of a chart that had is set to "auto" might still retain the value of "auto" in the new version.

cbron: Ahh, good point. I need to update the `localClusterEnabled` function to allow `auto` or `true`. Right now it only accepts true and otherwise sets it to false.

cbron: Updated

README.md
```diff
@@ -18,7 +18,7 @@ To get automated notifications of our latest release, you can watch the announce
 
 ## Quick Start
 
-    sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 rancher/rancher
+    sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 --privileged rancher/rancher
```
dramich, line 21: Can we add a quick explanation why `--privileged` is needed?

dramich (approved):

README.md
cbron: Sure, I'd prefer to put in the docs page we link below though, or lower in the readme. The quick start is just 2 lines and I feel adding it here removes some of the wow factor. I have to do a separate docs PR anyway.

Issue #28515, opened 2020-08-21
Update build script to run go mod tidy


maggieliu:
Now we no longer vendor starting Rancher 2.5, we need to update the build script to:
* Run go mod tidy

paynejacob:
https://drone-publish.rancher.io/rancher/rancher/3663

PR #28645, opened 2020-08-31
added go mod tidy to the validate script

paynejacob:
Added go mod tidy to the validate script.  This will remove any unused dependencies before we build rancher.

https://github.com/rancher/rancher/issues/28515

### DIFF
--- a/scripts/validate
+++ b/scripts/validate
@@ -11,5 +11,8 @@ fi
 echo Running: golangci-lint
 golangci-lint run
 
+echo Tidying up modules
+go mod tidy
+
 echo Verifying modules
 go mod verify
--- a/scripts/validate
+++ b/scripts/validate
@@ -16,3 +16,8 @@ go mod tidy
 
 echo Verifying modules
 go mod verify
+
+if [[ $(git diff --stat) != '' ]]; then
+  echo "Encountered dirty repo!"
+  exit 1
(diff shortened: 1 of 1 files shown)


rmweir (dismissed):
LGTM

cbron (dismissed):
lgtm. One thing is that this could turn the git status to dirty if it changes go.sum, which changes the tag in the version script, but I don't think that should actually cause any problems.

dramich:
The build needs to fail if `go mod tidy` causes any changes

dramich (approved):

scripts/validate
```diff
@@ -11,5 +11,13 @@ fi
 echo Running: golangci-lint
 golangci-lint run
 
+echo Tidying up modules
+go mod tidy
+
 echo Verifying modules
 go mod verify
+
+if [[ $(git diff --stat) != '' ]]; then
```
cbron, line 20: Any reason we used `git diff` ? Typically we use [`git status --porcelain`](https://github.com/search?q=org%3Arancher+porcelain&type=code). Not a big difference in functionality, but could be parsing subtleties.

paynejacob: I don't have a preference, ill switch it to porcelain to make it consistent.

cbron (approved):

Issue #28668, opened 2020-09-01
Rancher 2.4.6 - Api Key max duration


quadrinho:
**What kind of request is this (question/bug/enhancement/feature request): question

**Steps to reproduce (least amount of steps as possible): Create a Api Key in Rancher UI

**Result:You can create a Api Key with Max 1 Day expiration

**Other details that may be helpful:**

**Is there a Rancher API that can be used to refresh the Bearer Token so that it doesn't expire every day?**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): Rancher 2.4.6
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

martin-kieliszek:
+1 

Curious to get advice on what the recommended flow is, now that those who have migrated away from previous long lived API keys can only get maximum 1 day keys. Definitely would like to automate it (where previous automation wasn't needed as it was just a longer expiry than 1 day)

toolbar23:
+1 

I use api keys to let another system do stuff on my behalf. 

And most of the time I want that system to do it's job not just for 24 hours but for eternity (though I am ok with renewing these api keys every now and then).

vincent99:
This is something that was intended to be optional, and not on by default.  You can edit `/v3/settings/auth-token-max-ttl-minutes` for the moment.

quadrinho:
> This is something that was intended to be optional, and not on by default. You can edit `/v3/settings/auth-token-max-ttl-minutes` for the moment.

Thanks @vincent99 . On my side is usefully change that parameter.

maggieliu:
Need to update the AuthTokenMaxTTLMinutes to default to 0

jiaqiluo:
the following tests are done on v2.4.7-rc2: 


### 2.4.5 -> 2.4.6 -> 2.4.7
- run v2.4.5
- create tokens t1(never), t2(a day), t3(a month), and t4(a year) -> confirm that all tokens get the expected expiration date  
- upgrade to v2.4.6
- confirm that all tokens keep the existing expiration date  
- by default, auth-token-max-ttl-minutes = 1440
[...]

PR #28674, opened 2020-09-01
[v2.4] update AuthTokenMaxTTLMinutes default to 0

kinarashah:
The old behavior will be preserved by default. It's configurable for opt-in by changing the value. 

https://github.com/rancher/rancher/issues/28668

### DIFF
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -18,7 +18,7 @@ var (
 
 	AgentImage                        = NewSetting("agent-image", "rancher/rancher-agent:master-head")
 	AuthImage                         = NewSetting("auth-image", v3.ToolsSystemImages.AuthSystemImages.KubeAPIAuth)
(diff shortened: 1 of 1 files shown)


kinarashah:
@deniseschannon nope, our settings never override the set values on upgrade. since we're changing the default value itself, it'll only get applied when the setting is newly introduced or the value is empty

deniseschannon (dismissed):

pkg/settings/setting.go
```diff
@@ -18,7 +18,7 @@ var (
 
 	AgentImage                        = NewSetting("agent-image", "rancher/rancher-agent:master-head")
 	AuthImage                         = NewSetting("auth-image", v3.ToolsSystemImages.AuthSystemImages.KubeAPIAuth)
-	AuthTokenMaxTTLMinutes            = NewSetting("auth-token-max-ttl-minutes", "1440") // 24 hours
+	AuthTokenMaxTTLMinutes            = NewSetting("auth-token-max-ttl-minutes", "0")
```
deniseschannon, line 21: Maybe we should comment why/what 0 means

PR #28698, opened 2020-09-02
Check if CIS v2 CRD is present and error out while running CIS v1 clusterscan

prachidamle:
If CIS v2 is enabled via dashboard on a cluster, we want to disallow users from running a CIS v1 scan. 

Reason is that we do have a limitation of running only one CIS scan at any given time for a cluster - there cannot be parallel scans running. This limitation is due to the sonobuoy tool. 

Even if we want to just stall the scans when another one is running, it is complicated to predict the run times of scheduled scans if present on the cluster. Hence we just want to disable the v1 scans if user has enabled CIS v2 feature.

### DIFF
--- a/pkg/controllers/managementuser/cis/clusterScanHandler.go
+++ b/pkg/controllers/managementuser/cis/clusterScanHandler.go
@@ -1,6 +1,7 @@
 package cis
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -24,6 +25,7 @@ import (
 	"github.com/rancher/security-scan/pkg/kb-summarizer/report"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -69,6 +71,7 @@ type cisScanHandler struct {
 	dsClient                     appsv1.DaemonSetInterface
 	dsLister                     appsv1.DaemonSetLister
 	templateLister               v3.CatalogTemplateLister
+	apiExtensionsClient          clientset.Interface
 }
 
 type appInfo struct {
@@ -108,6 +111,16 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
 	if cluster.Spec.WindowsPreferedCluster {
 		v32.ClusterScanConditionFailed.True(cs)
 		v32.ClusterScanConditionFailed.Message(cs, "cannot run scan on a windows cluster")
+		return cs, nil
+	}
+
+	cisv2, err := csh.isCISv2Enabled(cluster)
+	if err != nil {
+		return cs, fmt.Errorf("cisScanHandler: Create: Error while checking CIS v2 CRD presence %v, will retry", err)
+	}
+	if cisv2 {
+		v32.ClusterScanConditionFailed.True(cs)
+		v32.ClusterScanConditionFailed.Message(cs, "cannot run scan on cluster, CIS v2 feature is enabled")
 		return cs, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/cis/clusterScanHandler.go
```diff
@@ -461,3 +487,39 @@ func (csh *cisScanHandler) getCISBenchmarkCatalogID() (string, error) {
 func (csh *cisScanHandler) getRancherCISBenchmarkTemplateID() string {
 	return fmt.Sprintf("%s-%s", cutils.SystemLibraryName, templateName)
 }
+
+func (csh *cisScanHandler) isCISv2Enabled(cluster *v3.Cluster) (bool, error) {
+	cfg, err := csh.getKubeConfig(cluster)
```
dramich, line 492: On the UserContext there is a `APIExtClient` which should be used to determine what CRDs are in the downstream cluster.

pkg/kubectl/kubectl.go
```diff
@@ -64,6 +64,22 @@ func Delete(yaml []byte, kubeConfig *clientcmdapi.Config) ([]byte, error) {
 	return runWithHTTP2(cmd)
 }
 
+func Get(kubeConfig *clientcmdapi.Config, resourceName string, args []string) ([]byte, string, error) {
```
dramich, line 67: This change wont be needed with the switch to using `APIExtClient`

pkg/controllers/managementuser/cis/clusterScanHandler.go
```diff
@@ -111,6 +117,16 @@ func (csh *cisScanHandler) Create(cs *v3.ClusterScan) (runtime.Object, error) {
 		return cs, nil
 	}
 
+	cisv2, err := csh.isCISv2Enabled(cluster)
+	if err != nil {
+		return cs, fmt.Errorf("cisScanHandler: Create: %v, will retry", err)
+	}
+	if cisv2 {
+		v32.ClusterScanConditionFailed.True(cs)
+		v32.ClusterScanConditionFailed.Message(cs, "cannot run scan on this cluster since CIS v2 feature is enabled")
```
dramich, line 126:
```suggestion
		v32.ClusterScanConditionFailed.Message(cs, "cannot run scan on cluster, CIS v2 feature is enabled")
```

pkg/controllers/managementuser/cis/clusterScanHandler.go
prachidamle: Do I need to initialize APIExtClient? I dont see APIExtClient exposing methods...

dramich: `user.APIExtClient.ApiextensionsV1().CustomResourceDefinitions().Get()`

pkg/controllers/managementuser/cis/clusterScanHandler.go
```diff
@@ -304,6 +317,16 @@ func (csh *cisScanHandler) Updated(cs *v3.ClusterScan) (runtime.Object, error) {
...
 		v32.ClusterScanConditionAlerted.Unknown(cs)
+	} else if v32.ClusterScanConditionFailed.IsTrue(cs) {
+		cluster, err := csh.clusterLister.Get("", csh.clusterNamespace)
+		if err != nil {
+			return nil, fmt.Errorf("cisScanHandler: Updated: error getting cluster %v", err)
+		}
+		updatedCluster := cluster.DeepCopy()
+		updatedCluster.Status.CurrentCisRunName = ""
+		if _, err := csh.clusterClient.Update(updatedCluster); err != nil {
+			return nil, fmt.Errorf("cisScanHandler: Updated: failed to update cluster about CIS scan completion")
```
dramich, line 328: Include the error message from the update call

prachidamle: done

pkg/controllers/managementuser/cis/clusterScanHandler.go
```diff
@@ -461,3 +484,16 @@ func (csh *cisScanHandler) getCISBenchmarkCatalogID() (string, error) {
 func (csh *cisScanHandler) getRancherCISBenchmarkTemplateID() string {
 	return fmt.Sprintf("%s-%s", cutils.SystemLibraryName, templateName)
 }
+
+func (csh *cisScanHandler) isCISv2Enabled(cluster *v3.Cluster) (bool, error) {
+	cisv2CRD := "clusterscanprofiles.cis.cattle.io"
+	_, err := csh.apiExtensionsClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.TODO(), cisv2CRD, metav1.GetOptions{})
+	if err != nil && kerrors.IsNotFound(err) {
```
dramich, line 491:
```suggestion
	if err != nil {
		if kerrors.IsNotFound(err) {
			return false, nil
		}
		return false, err
	}
	return true, nil
```

dramich (approved):

prachidamle:
I need an approval to merge, the approval got dismissed since I re-pushed to restart drone.

aiyengar2 (approved):

Issue #28131, opened 2020-07-29
Installing Rancher (Single Cluster (rancherd))


maggieliu:
* Installation available on RancherD CLI
** Please refer to the internal RancherD design doc on Confluence for details

StrongMonkey:
Here is the detailed implementation of rancherd:

Rancherd is a single binary released along with rancher. Master build of rancherd can be downloaded through https://releases.rancher.com/rancherd/rancherd-master. Rancherd is essentially rke2 + rancherd chart.

[Rancherd chart](https://github.com/rancher/rancher/tree/master/cmd/rancherd/chart) is a simplified version of rancher server chart we have today. 

-  rancher is deployed through **daemonset** instead of deployment in master node. This makes scaling rancher easier as it scales when you add more master nodes.

- rancher is exposed through hostport. This avoids applying ingress and ingress controller on top of rancher. To provide outbound access an external LB has to be configured to connect to hostports on server nodes. 

- cert-manager is no longer used to provision certs for rancher. In order to bring your own certs, the certs have to be provided on the following location on serve nodes. Rancherd chart will automatically mount these as hostPath into rancher pods.

 ```
 Private key: /etc/rancher/ssl/key.pem
 Certificate: /etc/rancher/ssl/cert.pem
 CA Certificate(self-signed): /etc/rancher/ssl/cacerts.pem
 Additional CA Certificate: /etc/ssl/certs/ca-additional.pem
```

- rancherd chart is customized through [helmChartConfig](https://github.com/rancher/rancher/tree/master/cmd/rancherd/chart#customize-chart) instead of overriding values.yaml in helm command. 

### Security improvement

Rancherd deploys rancher in a way so that no default admin will be generated on start-up, and in order to bootstrap default admin and password `rancherd reset-admin` has to be run on one of the server nodes. `rancherd reset-admin` provides the ability to bootstrap/reset default admin user and a random generated password. It will also print out a URL that redirects to UI to ask user to change password.(Note: this part has to be implemented in UI)

### QA

Rancherd can be installed throught package script that @dweomer is working on. Once rancherd is installed, run `rancherd reset-admin` to bootstrap admin and password. It should print out a URL, username and password. Click on the URL and fill out username and password and rancher should be able to use. Be sure to test the following use cases:

1. Make sure rancher is deployed to only master node. Rancher is also scalable with the number of master node in the cluster
2. Make sure rancher can be configured to access by an external LB that connects master node through 8080 and 8443.
[...]

PR #28722, opened 2020-09-02
rancherd and new RancherChart

StrongMonkey:
https://github.com/rancher/rancher/issues/28131


StrongMonkey:
This PR rewrite rancher charts and pin rke2 to the latest version.

The following changes are done through rancher charts:
1. Remove the dependency of cert-manager. This removes cert-manager issuer types as the certs have to be provided on disk/secret in the future.
2. Remove ingress. Rancher is exposed on hostPort by default.
3. Remove different certs option. Now the certs have to be provided on disk/secrets.
   

- If provided on disk, it has to be pre-mount on disk at the following location:

   certificate file: `/etc/rancher/ssl/cert.pem`
   private key file: `/etc/rancher/ssl/key.pem`
   cacerts(if self-signed): `/etc/rancher/ssl/cacerts`

- if provided by secrets, a secret name `tls-certs` on `cattle-system` namespace must pre-exist and contain the following data:

`kubectl create secret generic tls-certs --from-file=key.pem=/path/to/key.pem --from-file=cert.pem=/path/to/cert.pem`

4. Rancher deployment is now a daemonset instead of deployment.
  
Other rancher chart options remain unchanged. In order to configure custom value.yaml, HelmChartCRD must be leveraged. 
For example:

```
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: rancher
  namespace: kube-system
spec:
  valuesContent: |
     # custome value.yaml...
```

This has to be pre-written into `/var/lib/rancher/rke2/server/manifest` before rancherd is launched

StrongMonkey:
Per darren's request, the following changes have been made:
1. Certs can only be provided through hostpath. No secret hookup.
2. This is an internal rancherd charts that only lives in /cmd/rancher/charts.

StrongMonkey:
The certs workflow works in the same way with single docker install https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/single-node-docker/, which allows user to provide certs under `/etc/rancher/ssl` on the host and we use hostpath to mount these into container. That will be the only way to provide certs.

ibuildthecloud (approved):

Issue #26688, opened 2020-04-17
Error relating to "kubeclt apply failed" presented for around 5 minutes during cluster provisioning 


soumyalj:
Version: master-head(00ff15976)

Steps:
1. Create a DO cluster in master-head
2. Provisioning error is displayed as below for around 5 minutes. 
![Screen Shot 2020-04-17 at 11 25 00 AM](https://user-images.githubusercontent.com/18536626/79602560-8bb0da00-809f-11ea-8008-1076de8ffaca.png)

It recovers after 5 minutes

izaac:
This happened to me today

Rancher version `master-head (08/12/2020)` _d19f179c4_

I was using DO driver to provision a cluster. Usually after deploying Rancher and provisioning a cluster shortly after I get this error *occasionally*

soumyalj:
Tested with master-head(master-76c26fb0f9)

Deployed few node driver and custom clusters. 
The error as mentioned above was not observed during provisioning.

PR #28751, opened 2020-09-04
Only use the node dialer if the host is a control plane node or ready

ibuildthecloud:
During cluster provisioning you could end up with a node being first in the
   list that is worker only and wait to be provisioned because the control
   plane is still provisioning. In this situation we chose the worker node for
   the node dialer which fails because the nginx proxy does not exist yet. 
   Because the node dialer fails we can't deploy the cluster agent and get
   stuck in a failing loop.

### DIFF
--- a/pkg/dialer/factory.go
+++ b/pkg/dialer/factory.go
@@ -178,7 +178,8 @@ func (f *Factory) clusterDialer(clusterName, address string) (dialer.Dialer, err
 	}
 
 	for _, node := range nodes {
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
https://github.com/rancher/rancher/issues/26688

rmweir (approved):
nit: might be easier to read if we did continue's if cases aren't met but otherwise LGTM

dramich (approved):

Issue #28728, opened 2020-09-03
RFE: Flag to change the number of Helm Version Histories we keep for Charts deployed by Rancher


leflambeur:
**What kind of request is this (question/bug/enhancement/feature request):**
Enhancement

**Description**

At the moment we hard code the number of helm releases we store by default in a cluster for system charts at 10:

https://github.com/rancher/rancher/blob/master/pkg/controllers/managementuser/helm/common/common.go#L34

For clusters with limited resources or at the opposite end with lots of deployments, this assumption can cause instability

We should give customers the option to finetune this value so that they can either have more or less to meet their needs, probably as a flag in cluster.yml so it can be set for specific tenants in any customer environments at a cluster level. 

Globally is less preferred but I appreciate that is where the helm controller lives.

We should also lower the default from 10 to 5 so any soft recommendation we make has a less adverse effect (most customers would not use more than 1-2 rollbacks during troubleshooting anyway) and those customers who use this feature extensively can then choose to increase it.

dramich:
@maggieliu @catherineluse This needs to be added to the support kb, release notes and best practices guide.

bmdepesa:
rancher/rancher:master-head 76c26fb

* Rancher HA install
* Deploy downstream cluster
* Helm v2
  * Check `/v3/settings/helm-max-history` and see the default set to `10`
  * Deploy an app from a Helm v2 catalog and upgrade 10 times
  * Check ConfigMaps for the project where the app was deployed and see 10 ConfigMaps for each version of the app
  * Change `/v3/settings/helm-max-history` to `2` and upgrade the app again
  * Check ConfigMaps and see 2 ConfigMaps listed for the app version history
  * Upgrade again and see only 2 ConfigMaps listed with the appropriate versions
* Helm v3
  * Check `/v3/settings/helm-max-history` and see the value set to `2`, change it to `4`
  * Deploy an app from a Helm v3 catalog and upgrade 5 times
  * Check Secrets for the project where the app was deployed and see 5 secrets for each version of the app
  * Change `/v3/settings/helm-max-history` to `2` and upgrade the app again
  * Check Secrets and see 2 secrets listed for the app version history
  * Upgrade again and see only 2 secrets listed with the appropriate versions

bmdepesa:
* Boundary testing:
  * negative values (-1) and values above max int (2147483647) are accepted -- it doesn't break anything functionally, and seems to set to the default `256` but I haven't confirmed the exact value. We should probably validate this value though.

[...]

PR #28754, opened 2020-09-04
Make helm max history a setting

dramich:
Problem:
HelmMaxHistory is hardcoded to 10, in a setup with many helm apps this
causes a lot of secrets to be created

Solution:
Make helmMaxHistory a setting so it can be set based on customer needs

https://github.com/rancher/rancher/issues/28728


mrajashree (approved):

brendarearden (approved):
looks good!

Issue #28223, opened 2020-08-04
Launching kubectl fails with a blank screen/error


soumyalj:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a HA setup using master-head with ECM disabled
2. Create a user cluster 
3. Launch kubectl on the user cluster and local cluster

**Result:**
kubectl fails to launch as below:
**User cluster:**

![Screen Shot 2020-08-04 at 11 06 06 AM](https://user-images.githubusercontent.com/18536626/89328684-99846e00-d642-11ea-8999-3da2e4758bcd.png)

**Local cluster**

![Screen Shot 2020-08-04 at 11 08 21 AM](https://user-images.githubusercontent.com/18536626/89328859-d7819200-d642-11ea-9fcd-db69865faeea.png)

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head(3d7c3b3c1)
- Installation option (single install/HA): HA

Kubernetes Version: V1.18

sowmyav27:
**Verified on master-head - commit id: `d19f179c4`**

- Launch kubectl fails to load. Same issue as the original error is seen.
- Reopening the issue

izaac:
This is not yet merged.

izaac:
Rancher version `master-head (08/20/2020)` _56ecd6510_ 

In an HA setup I can get the kubectl console from the emberUI in the downstream clusters but the `local` cluster shows `Closed Code: 1006`

jiaqiluo:
on rancher:`master-f7f8e8bb0649e6f163bd6ba2118622a9889236ce-head` HA setup:
- admin can launch kubectl shell on Embler UI, and open the dashboard 
- but cluster member can NOT do neither 

on rancher:`master-4ff98cc603ea47528062ee6994d2a5e58a29ed5a-head` HA setup:
- even admin can NOT launch kubectl shell on Embler UI, or open the dashboard

prachidamle:
For admin user, kubectl shell is working fine on an HA setup launched using  master-ed2749841db0b3323e1e284ca3d2388a53f4bafd-head

For cluster-member, it is not working with this latest version yet

sowmyav27:
**On master-head - `0ccce450`**

Cluster owner is able to do a kubectl shell through ember UI and dashboard.

prachidamle:
[...]

PR #28755, opened 2020-09-04
Do not run on HTTPS because it might redirect improperly

ibuildthecloud:
If the x-forwarded-proto header is not set properly it will redirect
to HTTPS but the path will be wrong.  So just don't run on https as it's
not ever used.


ibuildthecloud:
https://github.com/rancher/rancher/issues/28223

PR #28759, opened 2020-09-04
Set node role to worker if nodepool has no role assigned

ibuildthecloud:
This makes the logic consist where no role means worker role.  For custom this is how we do it, but for node pools it seems to be different.

https://github.com/rancher/rancher/issues/28758


ibuildthecloud:
It's not really decided yet if this is the right behavior though.

PR #28797, opened 2020-09-08
rancherd: package installer

dweomer:
This adds rancherd-installer.linux-amd64.run to the built artifacts that
will be uploaded during CI builds. This is installer is effectively a
self-extracting tarball + shell script rolled into a single artifact.

Addresses #28131.


scripts/package
```diff
@@ -39,3 +39,18 @@ if [ ! -d build/system-charts ]; then
     git clone --branch $SYSTEM_CHART_DEFAULT_BRANCH https://github.com/rancher/system-charts build/system-charts
 fi
 TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go build/system-charts $IMAGE $AGENT_IMAGE
+
+mkdir -p ./build/installer/bin
+rsync -a ./rancherd ./build/installer/bin/
+rsync -a ../cmd/rancherd/installer/ ./build/installer/
+makeself.sh \
```
ibuildthecloud, line 46: So I run `./scripts/package` from the host quite a bit for development.  Can we gate this with a check that makeself.sh exists.  If it doesn't don't run this?  Since in development you most likely won't have makeself installed on your host.

dweomer: Yeah, I can skip this entire block if makeself.sh isnt available on the path. Alternatively we could install it into `bin/build/tools` or the like.

dweomer:
@ibuildthecloud for this I have copied, with slight modifications, the scripts from rke2 for install/killall/uninstall. I wonder if it would make more sense to download and unpack the rke2-installer (the makeself-generated files and run arbitrary tar commands on the bundled content), or clone the rke2 repo, and copy these scripts from there. It would take a little more effort to properly parameterize on the rke2 side but it would allow us to have one source of truth for these scripts.

scripts/package
dweomer: Resolved by https://github.com/rancher/rancher/commit/39724451fe6d744777ca4dc458e2654224b03598

dweomer:
Closing in favor of #28922

PR #28825, opened 2020-09-10
Don't use http proxy when sending requests to steve

ibuildthecloud:


ibuildthecloud:
https://github.com/rancher/rancher/issues/28850

Issue #25702, opened 2020-03-02
Drop share-mnt process


StrongMonkey:
Share-mnt process is the container we ran to make some directories shared in order for kubelet container to propagate mounts. Since we are starting kubelet with rshared volumes, container will fail to start if mountpoint is not rshared. Since most of OSs today mount their root dir as shared, it is not needed anymore.(It previously only applied for boot2docker, which is not supported in our matrix anymore). This issue tracks the work to drop share-mnt container.

Things to be noticed:
We still need share-mnt container to be run in order to write certs into rancherOS. Only the actual share-mnt process needs to be dropped.

bmdepesa:
rancher/rancher:master-head bcc8261

Fresh installs - 

* Fresh install in HA
* Create an RKE cluster
  * Cluster & node agents do not have `share-mnt` binary (exec shell into agent and `share-mnt`, and no container is running
* Create RKE ROS cluster
  * Deploys successfully
* Automation runs successfully against both clusters

soumyalj:
Tested with upgrade from v2.4.0 to master-head(bcc8261487ecd)

Steps:

- Created RKE DO  and ROS clusters in v2.4.0
- Cluster and node agents had the share-mnt binary
- Upgraded to Rancher to master-head
- Verified that share-mnt binary does not exist in the cluster and node agents

```
root@cattle-cluster-agent-59d4c7f5fd-zggjf:/var/lib/rancher# share-mnt --help
bash: share-mnt: command not found
root@cattle-cluster-agent-59d4c7f5fd-zggjf:/var/lib/rancher# 
```

```
root@soumyarostest3:/var/lib/rancher# share-mnt --help
bash: share-mnt: command not found
root@soumyarostest3:/var/lib/rancher# 

```
- Basic automation runs succeed on the clusters

PR #28845, opened 2020-09-10
Drop share mount process

dramich:
Problem:
The share-mount process is ran on every install/upgrade. This was used for boot2Docker which is no longer supported. 

Solution:
Drop the share-mount binary from the agent image and stop running it. This does not stop the rke process `share-mnt` from running in order to write certs. 
These are two different things with similar names that perform different functions. 

https://github.com/rancher/rancher/issues/25702

### DIFF
--- a/package/share-root.sh
+++ b/package/share-root.sh
@@ -3,11 +3,6 @@ set -x
 
 trap 'exit 0' SIGTERM
 ID=$(grep :devices: /proc/self/cgroup | head -n1 | awk -F/ '{print $NF}' | sed -e 's/docker-\(.*\)\.scope/\1/')
-IMAGE=$(docker inspect -f '{{.Config.Image}}' $ID)
 bash -c "$1"
 
-docker run --privileged --net host --pid host -v /:/host --rm --entrypoint /usr/bin/share-mnt $IMAGE "${@:2}" -- norun
-while ! docker start kubelet; do
-    sleep 2
-done
 docker kill --signal=SIGTERM $ID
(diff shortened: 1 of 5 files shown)


package/share-root.sh
```diff
@@ -3,10 +3,8 @@ set -x
 
 trap 'exit 0' SIGTERM
 ID=$(grep :devices: /proc/self/cgroup | head -n1 | awk -F/ '{print $NF}' | sed -e 's/docker-\(.*\)\.scope/\1/')
-IMAGE=$(docker inspect -f '{{.Config.Image}}' $ID)
 bash -c "$1"
 
-docker run --privileged --net host --pid host -v /:/host --rm --entrypoint /usr/bin/share-mnt $IMAGE "${@:2}" -- norun
 while ! docker start kubelet; do
```
StrongMonkey, line 8: you might want to remove this restart kubelet also

dramich: Well, this is a start call, so if it's already running it should be a no-op, but if it isn't then we start it (not sure why it would be stopped though)

dramich: Oh guessing it would be failed before share-mount ran, after share-mount finishes then we can start it again. So I would agree, should be able to drop it.

StrongMonkey: yeah, kubelet container is stopped because the mount is not shared. Once share-mnt is run, it tries to restart kubelet then. Without this logic we can safely drop it.

StrongMonkey (approved):
LGTM, would like to confirm with @ibuildthecloud so that we are comfortable dropping this since all the Oses rancher supports would no longer need share-mnt.

ibuildthecloud (approved):

PR #28865, opened 2020-09-11
Add shell image to images.txt

ibuildthecloud:


pkg/image/resolve.go
```diff
@@ -9,15 +9,14 @@ import (
...
-
-	rketypes "github.com/rancher/rke/types"
-
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/types/convert"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	libhelm "github.com/rancher/rancher/pkg/catalog/helm"
 	util "github.com/rancher/rancher/pkg/cluster"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/settings"
```
cbron, line 18: This resolve.go is run off of a separate go program, don't we have to register the settings provider ?

cbron: Think this worked actually

cbron (approved):

Issue #28907, opened 2020-09-14
[2.3] backport flag provided but not defined seen in share-mnt container


kinarashah:
Backport https://github.com/rancher/rancher/issues/25244

deniseschannon:
Closing as 2.3 is EOM.

PR #28908, opened 2020-09-14
[2.3] Add node-name param to the end of nodecommand

kinarashah:
https://github.com/rancher/rancher/issues/28907


superseb (commented):
Pretty sure we agreed on that this wasn't the root cause for the issue and it was solved after. This was a fix that would only help on OS's that need root shared which are unsupported anyway. So unless this is a cosmetic fix (to not see the error), this is most likely not fixing the issue.

kinarashah:
@superseb Yeah we agreed we don't need this fix anymore in 2.3.x. @maggieliu do we still need this?

maggieliu:
Ok, I didn't know we no longer need this fix anymore. @kinarashah can you close the PR and the issue and comment on it that it's no longer needed?

Issue #28879, opened 2020-09-12
Airgap - Built-in repos & charts are not available 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement	

**Steps to reproduce (least amount of steps as possible):**
- prepare a private registry for Rancher's images 
- run Rancher `v2.5.0-alpha2` single install in airgap environment with the option `-e CATTLE_SYSTEM_CATALOG=bundled` in the docker command 
- Rancher is deployed successfully
- go to local cluster -> dashboard -> apps 

**Result:**
No chart is available, since the repos are not accessible 
<img width="1866" alt="Screen Shot 2020-09-11 at 6 16 35 PM" src="https://user-images.githubusercontent.com/6218999/92983975-a07d7800-f45b-11ea-899f-d761c3df8f91.png">

 
<img width="1859" alt="Screen Shot 2020-09-11 at 6 05 50 PM" src="https://user-images.githubusercontent.com/6218999/92983971-9d828780-f45b-11ea-9fdc-d3bdeeac0bd0.png">

Expected:
When the option `CATTLE_SYSTEM_CATALOG=bundled` is set, we expect to see the builtin repos are available in the airgap environment.

cbron:
v1 charts (system-charts) are being bundled, just not v2 charts (rancher/charts). Looks like its just missing from the catalog_data package: https://github.com/rancher/rancher/blob/master/pkg/data/management/catalog_data.go#L145-L169

jiaqiluo:
the bug fix is validated on rancher:`master-01c13f69cb277e037f6d58323cb0859300e8af5d-head`

Steps are the same as the issue provides, and now we can see the charts and apps loaded in the dashboard 

<img width="1860" alt="Screen Shot 2020-09-15 at 3 19 48 PM" src="https://user-images.githubusercontent.com/6218999/93271811-fbb7af00-f768-11ea-8314-9a6e2f8d81c1.png">

![image](https://user-images.githubusercontent.com/6218999/93271828-08d49e00-f769-11ea-8ff6-e63ca70ba146.png)

PR #28915, opened 2020-09-15
Don't error out on updating bundled helm repos

ibuildthecloud:
If we fail to update a repo or the system catalog is set to bundled we do
   not record any error.


cbron:
issue: https://github.com/rancher/rancher/issues/28879

cbron (approved):

PR #28921, opened 2020-09-15
Fix eks import test

aaronRancher:
Changed the eks_config to remove unnecessary public and private access fields.

### DIFF
--- a/tests/validation/tests/v3_api/test_hosted_eks_cluster.py
+++ b/tests/validation/tests/v3_api/test_hosted_eks_cluster.py
@@ -148,10 +148,7 @@ def test_eks_v2_create_import_cluster():
         "amazonCredentialSecret": ec2_cloud_credential.id,
         "displayName": display_name,
         "imported": True,
-        "privateAccess": False,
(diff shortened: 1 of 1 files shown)


sowmyav27 (changes requested):
`"secretsEncryption": False,` can also be removed
Can you double check what UI is sending while adding an imported cluster?

aaronRancher:
Full EKS config from UI:
```
eksConfig: {amazonCredentialSecret: "cattle-global-data:cc-xlr4z", displayName: "test18643-ekscluster",…}
 amazonCredentialSecret: "cattle-global-data:cc-xlr4z"
 displayName: "test18643-ekscluster"
 imported: true
 region: "us-east-2"
 type: "/v3/schemas/eksClusterConfigSpec"
```
Will delete extra fields in automation eks_config.

aaronRancher:
Updated PR to match EKS config from UI.

sowmyav27 (approved):
LGTM

sangeethah (approved):
LGTM

PR #28922, opened 2020-09-15
rancherd: installation tarball

dweomer:
Intended to be unpacked at `/usr/local` the tarball provides:
- `bin/{rancherd,rancherd-killall.sh,rancherd-uninstall.sh}`
- `lib/systemd/system/rancherd-{agent,server}.service`
- `share/rancherd/rke2-cis-sysctl.conf`

Typical installation should be as simple as:
```sh
tar xzf rancherd-amd64.tar.gz -C /usr/local
systemctl enable --now rancherd-server
```

Assets under cmd/rancherd/bundle were copied from rke2/bundle and
adapted for rancherd.

Addresses #28131


cmd/rancherd/auth/auth.go
```diff
@@ -29,7 +30,8 @@ var (
 	defaultAdminLabel      = map[string]string{defaultAdminLabelKey: defaultAdminLabelValue}
 )
 
-func ResetAdmin(c *cli.Context) error {
+func ResetAdmin(_ *cli.Context) error {
```
StrongMonkey, line 33: why do we drop to not use context from root command?

dweomer: RKE2 reverted back to urfave/cli v1 which doesn't provide a context.Context implementation nor member.

dweomer:
Seems like a misconfigured test? https://drone-pr.rancher.io/rancher/rancher/10946/2/2

cjellick:
Sometimes you gotta give it The Fonz treatment

StrongMonkey (approved):
I have no domain knowledge around install script, LGTM then :smile:

cmd/rancherd/bundle/share/rancherd/rke2-cis-sysctl.conf
```diff
@@ -0,0 +1,5 @@
+vm.panic_on_oom=0
```
StrongMonkey, line 1: Should this file name be `rancherd-cis-sysctl`? not sure if you want to rename all these to rancherd

dweomer: It didn't seem necessary because `rancherd` is a very thin wrapper around `rke2` so I left it. This has the benefit of having the same file name that is mentioned in the rke2 readme (althought it is under share/rancherd instead of share/rke2). Happy to change if someone objects.

Issue #28900, opened 2020-09-14
User is not able to import an EKS cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create a cluster in the EKS console. Create one nodegroup - 2 nodes (default values)
- As a standard user, Import this cluster in Rancher after the cluster is seen in Active state in the EKS console. Give in the AWS cloud creds and select the cluster from the dropdown.
- Error seen on the cluster

<img width="1434" alt="Screen Shot 2020-09-14 at 10 11 48 AM" src="https://user-images.githubusercontent.com/26032343/93117262-6c7c9f80-f673-11ea-9b28-f2ddc437e918.png">

- Error in eks-operator logs: `time="2020-09-14T17:11:00Z" level=error msg="error syncing 'cattle-global-data/c-wtw6p': handler eks-controller: improper version format for cluster [c-wtw6p]: , requeuing"`
- eks operator version dpeloyed - `eks-operator:v0.1.0-rc29`

**Expected Result:**
User should be able to import the cluster into Rancher

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `e919796c8`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): EKS hosted
- Kubernetes version (use `kubectl version`):

```
1.17
```

sowmyav27:
**On master-head - commit id: `ad697207`**

- Import an EKS cluster on Rancher
- Cluster is stuck in `waiting` state with error

<img width="1412" alt="Screen Shot 2020-09-15 at 11 10 19 AM" src="https://user-images.githubusercontent.com/26032343/93248416-a36eb600-f744-11ea-8ec2-be1f96106a10.png">

sowmyav27:
**On master-head - commit id: `53cf961efc`**

- User is able to import an EKS cluster in rancher successfully.
- Cluster comes up successfully.

PR #28927, opened 2020-09-15
Fix nodegroup length check

rmweir:
**Problem:**
Prior, nodegroups were only being checked for imported clusters. This check should be performed regardless of a cluster being imported or not as neither type will be able to deploy the rancher agent without one. Prior, only the spec was being checked for nodegroups, but an imported cluster may not have nodegroups as a managed field. In this case the nodegroups should be reflected on the upstream spec.

**Solution:**
Now, it will be evaluated if a cluster has more than one nodegroup whether it is imported or not. Upstream spec will be checked for nodegroups if the field is nil on spec.

**Issue:**
https://github.com/rancher/rancher/issues/28900
https://github.com/rancher/rancher/issues/28422

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -185,33 +185,37 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		logrus.Infof("waiting for cluster EKS [%s] create failure to be resolved", cluster.Name)
 		return e.setFalse(cluster, apimgmtv3.ClusterConditionProvisioned, failureMessage)
 	case "active":
-		if apimgmtv3.ClusterConditionPending.IsUnknown(cluster) {
-			cluster = cluster.DeepCopy()
-			apimgmtv3.ClusterConditionPending.True(cluster)
-			cluster, err = e.clusterClient.Update(cluster)
-			if err != nil {
-				return cluster, err
-			}
-		}
-
 		if cluster.Spec.EKSConfig.Imported {
 			if cluster.Status.EKSStatus.UpstreamSpec == nil {
+				// non imported clusters will have already had upstream spec set
 				return e.setInitialUpstreamSpec(cluster)
 			}
-			addNgMessage := "Cannot deploy agent without nodegroups. Add a nodegroup."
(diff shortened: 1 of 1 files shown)


brendarearden (dismissed):
Looks good!

paynejacob (approved):

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -185,33 +185,37 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
-					if err != nil {
-						return cluster, err
-					}
+			}
+		}
+
+		addNgMessage := "Cannot deploy agent without nodegroups. Add a nodegroup."
+		noNodeGroupsOnSpec := len(cluster.Spec.EKSConfig.NodeGroups) == 0
+		noNodeGroupsOnUpstreamSpec := len(cluster.Status.EKSStatus.UpstreamSpec.NodeGroups) == 0
+		if cluster.Spec.EKSConfig.NodeGroups != nil && noNodeGroupsOnSpec || noNodeGroupsOnUpstreamSpec {
```
cbron, line 207: Nit: parentheses aren't necessary here, but would make it less confusing. `(true && false) || true`

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -185,33 +185,37 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		logrus.Infof("waiting for cluster EKS [%s] create failure to be resolved", cluster.Name)
 		return e.setFalse(cluster, apimgmtv3.ClusterConditionProvisioned, failureMessage)
 	case "active":
-		if apimgmtv3.ClusterConditionPending.IsUnknown(cluster) {
```
cbron, line 188: Do we not need this section anymore for rancher-created clusters ?

rmweir: Rancher created ones go: Provisioning -> waiting -> active
Imported: pending -> waiting -> active

cbron (dismissed):

cbron (approved):
lgtm

Issue #28764, opened 2020-09-05
Registering EKS cluster crashes Rancher when local cluster is off


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Install Rancher with local cluster disabled. Helm option is `addLocal` = `off`
2. Create a EKS cluster in the AWS console. Default options are fine.
3. In Rancher UI, create a cloud credential for the same AWS account used in step 2.
4. Click the Add Cluster button, select Register option, then EKS.
5. Enter a cluster name, select the cloud credentials from step 3, select the EKS cluster created in step 2. Click Save.

**Result:**
Rancher pods go into a crash loop. Seeing the following panic in the logs:
```
E0904 23:48:21.423243       7 runtime.go:78] Observed a panic: can't find system project (can't find system project) 
goroutine 1267 [running]: 
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x3a6a480, 0xc002160e40) 
	/go/pkg/mod/k8s.io/apimachinery@v0.18.8/pkg/util/runtime/runtime.go:74 +0xa3 
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0) 
	/go/pkg/mod/k8s.io/apimachinery@v0.18.8/pkg/util/runtime/runtime.go:48 +0x82 
panic(0x3a6a480, 0xc002160e40) 
	/usr/local/go/src/runtime/panic.go:969 +0x166 
github.com/rancher/rancher/pkg/controllers/management/eks.(*eksOperatorController).deployEKSOperator(0xc007c03320, 0x4097ab, 0xc0049d6000) 
	/go/src/github.com/rancher/rancher/pkg/controllers/management/eks/eks_cluster_handler.go:416 +0x7e3 
github.com/rancher/rancher/pkg/controllers/management/eks.(*eksOperatorController).onClusterChange(0xc007c03320, 0xc000d618b0, 0x7, 0xc00473e000, 0xc00473e000, 0x0, 0x0) 
	/go/src/github.com/rancher/rancher/pkg/controllers/management/eks/eks_cluster_handler.go:105 +0x8f 
[...]

PR #28932, opened 2020-09-16
Remove panic when deploying eks-operator fails

rmweir:
**Problem:**
Prior, rancher would fail if deploying eks-operator returned an error. This disrupted setups that did not have a local cluster.

**Solution:**
Now, cluster condition will be set to false and error will be displayed.

**Issue:**
https://github.com/rancher/rancher/issues/28764

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -103,6 +103,19 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 	}
 
 	if err := e.deployEKSOperator(); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -461,8 +474,9 @@ func (e *eksOperatorController) deployEKSOperator() error {
 
 	systemProject, err := project.GetSystemProject(localCluster, e.projectCache)
 	if err != nil {
-		panic(err)
```
ibuildthecloud, line 464: Well goodness, when you said "remove panic" I didn't think it was that literal :)

dramich (approved):

Issue #28916, opened 2020-09-15
Airgap - workload coredns does not pull from the private registry in the local cluster in docker install 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- prepare a private registry for Rancher's images
- run Rancher `v2.5.0-alpha2` single install in airgap environment with the option -e CATTLE_SYSTEM_CATALOG=bundled in the docker command
- Rancher is deployed successfully
- go to `local` cluster -> system project 

**Result:**
- Workload `coredns` does not pull from the private registry
![image](https://user-images.githubusercontent.com/6218999/93242113-2723a500-f73b-11ea-9299-4dd4dc88c220.png)

maggieliu:
Only affects local docker cluster

jiaqiluo:
The bugfix is validated on rancher: `v2.5.0-alpha3`, docker install in airgap env
 
![image](https://user-images.githubusercontent.com/6218999/93531786-a6a8a400-f8f4-11ea-8548-602aeac4cfef.png)

The image is included in the Rancher image, so it does not need to pull from the private registry.

PR #28938, opened 2020-09-16
Add k3s airgap images to build

dramich:
https://github.com/rancher/rancher/issues/28916

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -23,7 +23,7 @@ ENV CATTLE_SYSTEM_CHART_DEFAULT_BRANCH=$SYSTEM_CHART_DEFAULT_BRANCH
 ENV CATTLE_CHART_DEFAULT_BRANCH=$CHART_DEFAULT_BRANCH
 ENV CATTLE_PARTNER_CHART_DEFAULT_BRANCH=$PARTNER_CHART_DEFAULT_BRANCH
 ENV CATTLE_HELM_VERSION v2.16.8-rancher1
-ENV CATTLE_K3S_VERSION v1.17.2+k3s1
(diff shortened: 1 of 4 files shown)


package/Dockerfile
```diff
@@ -100,6 +103,8 @@ RUN curl ${HELM_URL_V3} | tar xvzf - --strip-components=1 -C /usr/bin && \
 RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
     curl -sLf ${!CHANNELSERVER_URL} > /usr/bin/channelserver && \
     curl -sLf ${!K3S_URL} > /usr/bin/k3s && \
+    mkdir -p /var/lib/rancher/k3s/agent/images/ && \
+    curl -sLf ${!K3S_IMAGE_TAR} > /var/lib/rancher/k3s/agent/images/${CATTLE_K3S_VERSION}-images.tar && \
```
cbron, line 107: Shouldn't the final path be `.../k3s-airgap-images-$ARCH.tar` ? Thats how docs have it, but idk if k3s just picks up whatever tar is in that dir.

cbron (commented):
Thought we needed Norman PR to update k3s server setting, but guess not, should pick these up automatically.

package/Dockerfile
dramich: The name is not important, any tar file in the dir will get extracted.

cbron (approved):
lgtm

ibuildthecloud:
We only really need coredns and pause.  So it's bummer to pull in a couple hundred mb for that.

ibuildthecloud (approved):

cbron (approved):
makes sense to me

Issue #28954, opened 2020-09-17
When alerts only configured notifiers that are not webhook forwarded, rancher server returns secrets "Webhook -receiver" not found error


DoNotSayYes:
**What kind of request is this (question/bug/enhancement/feature request):**
bug/enhancement

**Steps to reproduce (least amount of steps as possible):**
1. Create a new alert group, add an alert rule, and configure a notifier that is not webhook forwarded. (slack, or email).
2. Make the alert rule trigger alert, then check on the rancher server master container log

**Result:**
Users can still receive alert messages normally but receiving these error logs, which can be annoying
```
2020/09/17 01:42:53 [ERROR] ClusterAlertGroupController c-tz5cn/cag-s5kr2 [cluster-alert-group-controller] failed with : Update Webhook Receiver Config: Get secret: secrets "webhook-receiver" not found
2020/09/17 01:42:53 [ERROR] ClusterAlertGroupController c-tz5cn/cag-s5kr2 [cluster-alert-group-controller] failed with : Update Webhook Receiver Config: Get secret: secrets "webhook-receiver" not found
2020/09/17 01:42:53 [ERROR] ClusterAlertRuleController c-tz5cn/car-74thf [cluster-alert-rule-controller] failed with : Update Webhook Receiver Config: Get secret: secrets "webhook-receiver" not found
2020/09/17 01:42:54 [ERROR] ClusterAlertGroupController c-tz5cn/cag-s5kr2 [cluster-alert-group-controller] failed with : Update Webhook Receiver Config: Get secret: secrets "webhook-receiver" not found
2020/09/17 01:42:54 [ERROR] ClusterAlertRuleController c-tz5cn/car-74thf [cluster-alert-rule-controller] failed with : Update Webhook Receiver Config: Get secret: secrets "webhook-receiver" not found
```
![image](https://user-images.githubusercontent.com/57614736/93414193-61c04580-f8d3-11ea-8bdc-3c2b2f67945a.png)

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 
v2.4.8
- Installation option (single install/HA): 
HA

gz#13134

gz#13273

gz#13110

mfbrj:
The error log also happens when a Slack notifier is configured in a existing group alert such as "A set of alerts for node"
Rancher: v2.4.8
Install: HA

izaac:
Reproduced on `v2.4.8`

[...]

PR #28956, opened 2020-09-17
Fix return error when secret webhook-receiver is not found

GGGitBoy:
**Problem:**

User has not configured alert that requires webhook-receiver, but will still receiver the error log `secret webhook-receiver is not found`.

**Solution:**

when the secret webhook-receiver cannot be found, it will not be output as an error

**Relate issue:**

https://github.com/rancher/rancher/issues/28954

### DIFF
--- a/pkg/controllers/managementuser/alert/configsyncer/configsyncer.go
+++ b/pkg/controllers/managementuser/alert/configsyncer/configsyncer.go
@@ -123,7 +123,7 @@ func (d *ConfigSyncer) sync() error {
 	}
 
 	systemProjectName := project.Name
-	isDeployed, err := d.isAppDeploy(systemProjectName)
+	isDeployed, webhookReceiverEnabled, err := d.isAppDeploy(systemProjectName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


aiwantaozi:
If we enable MS team or other notifiers need this secret, and sever can't find this secret, and we directly return nil at this situation, it may cause problem.

aiwantaozi (commented):
If we enable MS team or other notifiers need this secret, and sever can't find this secret, and we directly return nil at this situation, it may cause a problem.

orangedeng (approved):
LGTM

orangedeng:
Please check the build failure.

aiwantaozi:
Looks like this will only sync receivers when webhookReceiverEnabled, when the user only use notifiers no need to deploy webhookReceiver, the receivers won't update right?

pkg/controllers/managementuser/alert/configsyncer/configsyncer.go
```diff
@@ -255,8 +255,10 @@ func (d *ConfigSyncer) sync() error {
 		logrus.Debug("The config stay the same, will not update the secret")
 	}
 
-	if err := d.syncReceiver(notifiers, cAlertGroupsMap, pAlertGroupsMap); err != nil {
-		return errors.Wrapf(err, "Update Webhook Receiver Config")
+	if webhookReceiverEnabled {
+		if err := d.syncReceiver(notifiers, cAlertGroupsMap, pAlertGroupsMap); err != nil {
```
aiwantaozi, line 259: This function is used for syncing configure for webhook-receiver, can we rename it?

aiwantaozi (approved):

aiwantaozi (approved):

dramich (approved):

aiyengar2 (commented):
lgtm

Issue #28926, opened 2020-09-15
Airgap - helm-operation pod keeps creating in the local cluster 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- run the latest rancher:`master-01c13f69cb277e037f6d58323cb0859300e8af5d-head`, single install in airgap environment 
- go to the local cluster -> system project 

**Result:**
- there are more and more pods for helm-operation created 
- and they might be related to the rancher-webhook app  

<img width="1408" alt="Screen Shot 2020-09-15 at 3 25 16 PM" src="https://user-images.githubusercontent.com/6218999/93271661-a9768e00-f768-11ea-9351-68a0bdbe6a23.png">

![image](https://user-images.githubusercontent.com/6218999/93271644-a2e81680-f768-11ea-8974-2921aab4f586.png)

maggieliu:
@ibuildthecloud should the helm-operation pod be more aggressive in cleaning up pods? When we error out it fills up with pods every quickly.

ibuildthecloud:
@maggieliu Yes, we need to not try so much.  I'll address this.

jiaqiluo:
The bugfix is validated on rancher: `v2.5.0-alpha3`, docker install in airgap env

jiaqiluo:
The fix is validated on rancher:`v2.5.0-alpha4` 
the steps are the same as the issue provides.

![image](https://user-images.githubusercontent.com/6218999/93690563-92001380-fa8e-11ea-838f-86e53133204f.png)

PR #28959, opened 2020-09-17
Only try to install system charts every 15 minutes

ibuildthecloud:


dramich:
https://github.com/rancher/rancher/issues/28926

Issue #28930, opened 2020-09-16
Dashboard feature flag remains deactivated after upgrade 


soumyalj:
**What kind of request is this (question/bug/enhancement/feature request):**Bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy rancher v2.4.8 single node setup
2. At the global level, navigate to settings --> Feature Flags.
3. Deactivate Dashboard
4. Deploy few clusters in the setup and verify dashboard button is not available 
5. Upgrade rancher to master-head(master-01c13f69cb277) with --privileged flag

**Result:**
After upgrade it was observed that dashboard feature flag was still "Deactivated"  and the users do not have the option to switch to Dashboard view
<img width="1357" alt="Screen Shot 2020-09-15 at 5 41 49 PM" src="https://user-images.githubusercontent.com/18536626/93279701-bef5b300-f77c-11ea-87f9-d84d6589282f.png">

The dashboard feature flag should be enabled on a upgrade to master-head setup

Note: The dashboard feature flag can be manually activated and users can switch to dashboard view from the clusters.

Question: Should the dashboard feature flag be visible in "Settings" after upgrade or in a fresh install setup?

soumyalj:
Tested with upgrade from v2.4.8 to master-head

1. Deploy rancher v2.4.8 single node setup
2. At the global level, navigate to settings --> Feature Flags.
3. Deactivate Dashboard
4. Deploy few clusters in the setup and verify dashboard button is not available 
5. Upgrade rancher to master-head(d671d58a30606) with --privileged flag

After upgrade, the dashboard feature flag is still "Deactivated" , but the dashboard button is available to be used in each of the clusters.

soumyalj:
Related issue: https://github.com/rancher/rancher/issues/28928

deniseschannon:
Backend is working as expected and we will hide the feature flag using this ticket: https://github.com/rancher/rancher/issues/28928

Issue #28928, opened 2020-09-16
“dashboard" feature flag should not visible in the UI


sangeethah:
Rancher server version - Latest build from master - 

Fresh install of 2.5 server.
Go to settings->Feature Flags
“dashboard” feature flag is available and enabled by default. 
We are allowed to disable this as well. 

Expected behavior:
We should not have “dashboard" feature flag.

soumyalj:
Tested on master-head(bcc8261487ecdd3776)

**Steps:**

- Deploy a fresh install  HA setup 
- Navigate to Settings -> Feature flags
- The dashboard feature flag is still displayed and activated
- We can still deactivate the feature flag. Once deactivated, the HA setup is unavailable only for few seconds and becomes available again
After deactivating the feature flag, the UI button to access the dashboard is not available in each of the clusters. But we can access the dashboard on local and downstream clusters using the call :
https://xyz/dashboard/c/local/explorer
https://xyz/dashboard/c/c-jsdfd/explorer

deniseschannon:
Is this still in the API? @dramich

sangeethah:
@dramich We see the dashboard feature flag is  still in fresh install .

dramich:
@sangeethah May need to check what version you are running, the feature was literally deleted from the code so that isn't possible.

sangeethah:
On Fresh install of master-head - `bd88c` , we dont see the dashboard feature flag enabled anymore as expected.

<img width="1272" alt="Screen Shot 2020-09-19 at 6 34 14 PM" src="https://user-images.githubusercontent.com/4266958/93692371-35f5b900-faa7-11ea-8747-7176d2641b3a.png">

sangeethah:
This issue now tracks the UI fixes needed for hiding the dashboard feature flag on upgrade cases -  https://github.com/rancher/rancher/issues/28930#issuecomment-694633565

sangeethah:
Tested with Rancher server version -` v2.5.0-rc2`
ui-index - `https://releases.rancher.com/ui/latest2/index.html`

Deploy rancher v2.4.8 single node setup
At the global level, navigate to settings --> Feature Flags.
Deactivate Dashboard
Deploy few clusters in the setup and verify dashboard button is not available
Upgrade rancher to ` v2.5.0-rc2` with --privileged flag
After upgrade, the dashboard feature flag is still "disabled" , but the dashboard button is available to be used in each of the clusters.

[...]

PR #28964, opened 2020-09-17
Drop dashboard from being a feature flag

dramich:
This will enable dashboard permanently. The feature CR will still exist but it just won't do anything to dashboard. This is being done so that rollback will not be broken if the CR is dropped on upgrade. 

https://github.com/rancher/rancher/issues/28928
https://github.com/rancher/rancher/issues/28930

### DIFF
--- a/pkg/api/steve/proxy/proxy.go
+++ b/pkg/api/steve/proxy/proxy.go
@@ -127,13 +127,7 @@ func (h *Handler) MatchNonLegacy(prefix string, force bool) gmux.MatcherFunc {
 		}
 		match.Vars["clusterID"] = clusterID
 
-		cluster, err := h.clusters.Get(clusterID)
-		if err != nil {
-			return true
-		}
-
-		// No steve means we are upgrading a node that doesn't have the right proxy
-		return cluster.Status.AgentFeatures[features.Steve.Name()]
+		return true
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/api/steve/proxy/proxy.go
```diff
@@ -119,21 +119,7 @@ func (h *Handler) MatchNonLegacy(prefix string, force bool) gmux.MatcherFunc {
 		if !features.SteveProxy.Enabled() && !force {
 			return false
 		}
-
-		clusterID := strings.TrimPrefix(req.URL.Path, prefix)
```
ibuildthecloud, line 123: Oh crap. This might fail if we are upgrade from 2.3

ibuildthecloud: Thinking..... Thinking....

ibuildthecloud: I think we are fine.

pkg/api/steve/proxy/proxy.go
```diff
@@ -119,21 +119,7 @@ func (h *Handler) MatchNonLegacy(prefix string, force bool) gmux.MatcherFunc {
...
-		clusterID = strings.SplitN(clusterID, "/", 2)[0]
-		if match.Vars == nil {
-			match.Vars = map[string]string{}
-		}
-		match.Vars["clusterID"] = clusterID
```
ibuildthecloud, line 128: This is mutating state, so it could be important.  It's setting "clusterID" which is scoped to the current request.  I'd probably leave it unless you can prove it's not needed.

ibuildthecloud (approved):

Issue #28867, opened 2020-09-11
Repeated "HTTP2 has been explicitly disabled" log statements in the rancher logs


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster in the latest master.
- Rancher logs show repeated log statements

```
I0910 22:26:13.125428   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:13.125710   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:13.125721   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:28.125588   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:28.126647   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:28.127171   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:43.125688   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:43.126727   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:43.127235   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:58.125517   22 http.go:116] HTTP2 has been explicitly disabled
I0910 22:26:58.126350   22 http.go:116] HTTP2 has been explicitly disabled
```

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id : `ab6598e5a`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Kubernetes version (use `kubectl version`):

```
1.18
```

maggieliu:
Disabled here: https://github.com/rancher/rancher/issues/26624

sowmyav27:
**On master-head - commit id: `8fd81a9b`**

- Rancher logs do NOT show these log statements anymore
```
HTTP2 has been explicitly disabled
```

PR #28990, opened 2020-09-18
Don't use DISABLE_HTTP2 env var

ibuildthecloud:
By properly configure the NextProtos to disable http2. Using the
DISABLE_HTTP2 env var causes a repeated log message.


maggieliu:
related to: https://github.com/rancher/rancher/issues/28867

Issue #28861, opened 2020-09-11
Chart - no charts available for non-cluster-owner role


jiaqiluo:
bug rancher `6a6f13` HA installation 

As a standard user, cluster member/ project owner/ project member, go to dashboard -> apps 

<img width="1305" alt="Screen Shot 2020-09-11 at 10 15 25 AM" src="https://user-images.githubusercontent.com/6218999/92955472-08ab6a00-f41a-11ea-9d2e-bb8171a7d4a0.png">

------

It turns out only admin/ cluster owner can see charts. 
All other roles ( cluster member, project owner, project member) can NOT see charts.

```
> k get clusterrepos.catalog.cattle.io -A
Error from server (Forbidden): clusterrepos.catalog.cattle.io is forbidden: User "u-q84bt" cannot list resource "clusterrepos" in API group "catalog.cattle.io" at the cluster scope
```

dramich:
None of the new types have RBAC associated with them yet which is why this only works for admins. Per discussion with @ibuildthecloud we aren't ready to implement this yet so a fix is TBD.

deniseschannon:
Still discussing the exact level of permissions for each type, but need to be implemented to add access to these CRD types to our built-in roles. 

Type | Cluster Owner | Cluster Member | Project Owner | Project Member
---|---|---|---|---
ClusterRepos | CRUD | CRUD | R | R
Operations | R | R | R | R 
Releases | R | R | R | R

jiaqiluo:
on rancher:`master-9addfe297070f85eb3020d0e54571adaa438ccec-head`, docker install 

As a standard user, cluster member/ project owner/ project member can see apps in the dashboard now.

jiaqiluo:
This issue can be closed since the original bug in the ticket is fixed.

A new issue is made for the RBAC table.

sangeethah:
Related issue - https://github.com/rancher/rancher/issues/29038

PR #28991, opened 2020-09-18
RBAC for v2 apps

mrajashree:
**Problem**: Users with role cluster-member, project-owner and project-member cannot list Charts
**Solution**: Add rules for the `catalog.cattle.io` CRDs to these roles
With this change, 
* Cluster-member:
 1. Can list all Charts
 2. Can list installed Apps only of their projects (if they belong to a project or have created one)

* Project-owner and Project-member 
 1. Can list all Charts
 2. Can list only those Apps that are installed in namespaces belonging to their projects. (By default I can't select what namespace to install an app in in the dashboard, so I tested this by first installing an app then moving the namespace to a project)

* This commit also adds rules for the project scoped Read-only role:
1. Can list all Charts
2. Can list only those Apps that are installed in namespaces belonging to their projects
we do the same for v1 apps

https://github.com/rancher/rancher/issues/28861

### DIFF
--- a/pkg/controllers/managementuser/rbac/prtb_handler.go
+++ b/pkg/controllers/managementuser/rbac/prtb_handler.go
@@ -29,6 +29,9 @@ var globalResourcesNeededInProjects = map[string]map[string]bool{
 	"apiservices": {
 		"apiregistration.k8s.io": true,
 	},
+	"clusterrepos": {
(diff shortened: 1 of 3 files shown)


mrajashree:
TFs are unrelated

dramich (approved):

Issue #28984, opened 2020-09-18
Settings must be propagated to downstream cluster


cbron:
Problem: When settings change they don't get propagated to downstream cluster. 

Solution: If settings change we need to propagate the settings, and also redeploy the downstream agents.

Blocking: https://github.com/rancher/rancher/issues/28923

jiaqiluo:
this is closed, https://github.com/rancher/rancher/issues/28923 is tracking the progress.

PR #29016, opened 2020-09-19
Sync install-uuid and system-default-registry

ibuildthecloud:
Set values in downstream clusters.


ibuildthecloud:
https://github.com/rancher/rancher/issues/28984

Issue #28999, opened 2020-09-18
Error while upgrading the k8s version of an EKS v1 cluster upgraded from 2.4.8 


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- deploy an EKS v1 on 2.4.8 in 1.16 k8s version
- Upgrade to master-head
- Edit cluster, upgrade k8s version of the cluster to 1.17
- Save changes made
- Error on the UI
- API Request sent has an error in the response
- Response: `{"baseType":"error","code":"InvalidBodyContent","message":"cannot modify EKS subnets after creation","status":422,"type":"error"}`

**Expected Result:**
User should be able to edit a v1 EKS cluster

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `8fd81a9b`
- Installation option (single install/HA): single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): EKS v1
- Kubernetes version (use `kubectl version`):

```
1.16 to 1.17
```

aaronRancher:
The bug fix is verified with Rancher server `rancher\rancher:master-head` `version e9a98a3`

Steps:
deploy an EKS v1 on 2.4.8 in 1.16 k8s version
Upgrade to master-head
Edit cluster, upgrade k8s version of the cluster to 1.17
Save

Result:
Cluster can be edited and upgraded:
![Screen Shot 2020-09-23 at 11 30 59 AM](https://user-images.githubusercontent.com/57409052/94054285-51efa800-fd90-11ea-88b6-ab47a8c040b3.png)

PR #29027, opened 2020-09-20
Dont mutate eksv1 subnets

rmweir:
**Problem:**
Prior, if subnets were different an error would be returned. Before v2.5 sending a request with an empty array would result in a nil field. Now it results in an empty array. This is causing the error to return given the same cluster body.

**Solution:**
To prevent field updates sometimes the "noupdate" tag is used. These were not used before because the subnets are in a dynamic config. Now, logic has been added to the cluster store to implement the same behavior as the "noupdate" tag which simply does not mutate its fields value.

**Issue:**
https://github.com/rancher/rancher/issues/28999

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"fmt"
 	"net/http"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


dramich (changes requested):
I think there are a couple of options to address this issue. The most correct would be to add logic [here-ish](https://github.com/rancher/rancher/blob/master/pkg/controllers/management/drivers/kontainerdriver/kontainerdriver.go#L178) to check if the driver is EKS and if so set the field update to false. 
Another option would be to extend the `validateEKS` check to see if the len of the subnets is 0, if so return nil. 
My concern with the current fix is we are dropping a field that we say you can update but also provide no error. I think this would be a worse user experience than returning an error on a field we say you can update. I think the best option is fix the schema.

rmweir:
@dramich I understand the concern with the no error behavior but we use the noupdate tag heavily and this is how it behaves. Setting default to false would be the same. I think putting it there would make it harder to fine and confuse someone while debugging.

dramich:
@rmweir But then it's expected behavior. Adding logic to a store to drop a field is not expected and would be much harder to understand vs the schema explicitly stating the field is a no update field. The simplest fix at this point is to just fix the validator and in the interest of time for the release should probably be the path forward.

rmweir:
Closed in favor of https://github.com/rancher/rancher/pull/29092

Issue #28968, opened 2020-09-17
EKS cluster is stuck when updating mismatched k8s version in both cluster and node group


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Import a k8s 1.15 EKS cluster with a k8s 1.15 node group
From EKS console side, upgrade cluster to 1.16
Wait for update/sync on Rancher side
From Rancher side, upgrade k8s 1.17 and upgrade node group to 1.16 in the same edit

**Result:**
Cluster is stuck in updating state:
![image](https://user-images.githubusercontent.com/57409052/93514046-90421e80-f8db-11ea-933e-026c1617a070.png)
User is unable to edit k8s version to correct this:
![Screen Shot 2020-09-17 at 11 47 50 AM](https://user-images.githubusercontent.com/57409052/93514093-a0f29480-f8db-11ea-9362-ee269955df69.png)

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `rancher\rancher:master-head` `version 53cf961`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported

aaronRancher:
Deleting the cluster from Rancher and reimporting it will reset the k8s versions on edit.
Cluster is no longer stuck updating on reimport.

cbron:
@aaronRancher did the first upgrade finish completely ? Or was it still updating.

aaronRancher:
@cbron Neither upgrade completed. The cluster did not enter updating when viewed in EKS console. From the console side, the cluster had not changed at all.

sowmyav27:
A similar issue - **On master-head - commit id: `33f755f1`**

- Deploy a hosted EKS cluster - with k8s 1.15
- When the cluster comes up Active, Edit the k8s version from EKS console of the cluster to 1.16
- When the sync has finished to Rancher, the cluster in Rancher is now in 1.16 but the node groups are in 1.15 k8s version
- Edit cluster and change the k8s version to 1.17
- Error seen on the cluster
<img width="1453" alt="Screen Shot 2020-09-17 at 4 10 59 PM" src="https://user-images.githubusercontent.com/26032343/93537781-5421b480-f901-11ea-9658-43313388d89b.png">

- User is NOT able to come out of this state

[...]

PR #29029, opened 2020-09-20
Validate upgrades for EKS clusters

rmweir:
**Problem:**
EKS cluster version can be upgraded twice in a row without upgrading nodegroups. This is not valid in EKS. There is validation for this in the eks-operator but not stopping the user at the API level puts the cluster in a state that cannot be fixed through the UI.

**Solution:**
Add validation to the API that does not let the cluster upgrade unless all nodegroup versions are equal to the current kubernetes version.

**Issue:**
https://github.com/rancher/rancher/issues/28968

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -277,13 +277,6 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			}
 		}
 
-		// check for unauthorized error
-		if failureMessage != "" {
-			if strings.Contains(failureMessage, "403") {
-				return e.setFalse(cluster, apimgmtv3.ClusterConditionUpdated, "cannot access EKS, check cloud credential")
-			}
-		}
-
 		cluster, err = e.recordAppliedSpec(cluster)
 		if err != nil {
 			return cluster, err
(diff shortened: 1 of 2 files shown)


aiyengar2 (approved):

dramich (changes requested):
`validateEKSConfig` needs to accept the `var clusterSpec v32.ClusterSpec` created at the top of the validator, this will stop the need for all the type conversions and the eating of type checks. The cluster is already put into a real ClusterSpec so all data is already available.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -302,6 +295,9 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			return e.setUnknown(cluster, apimgmtv3.ClusterConditionUpdated, "")
 		}
 		logrus.Infof("waiting for cluster EKS [%s] update failure to be resolved", cluster.Name)
+		if strings.Contains(failureMessage, "403") {
+			failureMessage = "cannot access EKS, check cloud credential"
```
dramich, line 299: Why does this drop the original message? a 403 is much more explicit to being an auth issue, saying `cannot access EKS` seems ambiguous and leans more towards a networking issue.

Issue #28863, opened 2020-09-11
Cluster goes into error state after deploying an app from a custom repo


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a custom cluster - 5 nodes - 1 etcd, 1 control, 3 worker in a HA setup.
- In the dashboard deploy a repo - type `http` - `https://kubernetes-charts.storage.googleapis.com`
- In the Dashboard go to charts and deploy an app (Mysql) from this repo. Request sent to load the chart fails
```
Req URL: https://<rancher-server>/k8s/clusters/c-d8qd5/v1/catalog.cattle.io.clusterrepos/test-new?link=info&chartName=mysql&version=1.6.7
Request Method: GET
Status Code: 500
```
- Cluster will be in `Error` state - `Cluster health check failed: Failed to communicate with API server: Get "https://<>:6443/api/v1/namespaces/kube-system?timeout=45s": context deadline exceeded`

**Expected Result:**
Cluster should not throw an error

**Other details that may be helpful:**
Was able to see this issue on two different setups
In one of the setups the cluster recovers, but goes bad while trying to deploy the app from the repo again.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `6a6f13cec3`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Kubernetes version (use `kubectl version`):

```
1.18.8
```

StrongMonkey:
Don't test it, need https://github.com/rancher/rancher/pull/29043 to go in.

sowmyav27:
**On master-head - commit id: `f1a410a97`**

- Deploy a custom cluster - 5 nodes - 1 etcd, 1 control, 3 worker in a HA setup.
- In the dashboard deploy a repo - type http - `https://kubernetes-charts.storage.googleapis.com`
- In the Dashboard go to charts and deploy an app (Mysql) from this repo, persistence set as false.
- App is deployed successfully.
- Cluster does NOT go into a bad state after the app is deployed or the custom repo is added.

PR #29030, opened 2020-09-20
Cache index repo file

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/28863

pkg/catalogv2/content/content.go
```diff
@@ -76,6 +86,15 @@ func (c *Manager) Index(namespace, name string) (*repo.IndexFile, error) {
 		return nil, err
 	}
 
+	c.lock.RLock()
+	if cache, ok := c.IndexCache[fmt.Sprintf("%s/%s", r.status.IndexConfigMapNamespace, r.status.IndexConfigMapName)]; ok {
+		if bytes.Compare(cache.binaryData, cm.BinaryData["content"]) == 0 {
```
ibuildthecloud, line 91: Why don't you just keep the revision number of the config map. Faster comparison.

pkg/catalogv2/content/content.go
```diff
@@ -95,11 +114,18 @@ func (c *Manager) Index(namespace, name string) (*repo.IndexFile, error) {
...
+		index:      *index,
+		binaryData: cm.BinaryData["content"],
+	}
+	c.lock.Unlock()
+
 	return c.filterReleases(index), nil
 }
 
 func (c *Manager) filterReleases(index *repo.IndexFile) *repo.IndexFile {
-	index.SortEntries()
```
ibuildthecloud, line 102: Was this redundant? It's really important the data is sorted.

pkg/catalogv2/content/content.go
StrongMonkey: oh you are right

pkg/catalogv2/content/content.go
StrongMonkey: https://github.com/rancher/rancher/blob/417bdfd9874f8903c5d9923f78859fb1e313a098/pkg/controllers/dashboard/helm/repo.go#L180 you did sorting at here

Issue #29021, opened 2020-09-19
Cluster member can't create namespaces in Project that it owns


izaac:
**What kind of request is this:**
bug

**Steps to reproduce (least amount of steps as possible):**

- Install Rancher and create 2 Standard user `user1` and `user2`
- `user1` provisions a cluster and adds `user2` as `Cluster Member`
- `user2` Create a Project and then try creating a namespace in this Project.

**Result:**

`projects.management.cattle.io "p-smwqg" is forbidden: User "u-p9lb7" cannot get resource "projects" in API group "management.cattle.io" in the namespace "c-k9s8v"`

Error when creating the Project (the project gets created regardless):

![Screen Shot 2020-09-19 at 12 39 14 PM](https://user-images.githubusercontent.com/27659/93687923-b5b75f80-fa76-11ea-9a86-8b7e258731d8.png)

**Project owners are not getting access to the cluster either.**

**Other details that may be helpful:**

**Environment information**
- Rancher version: `master-head (09/19/2020)` _bd88c23_
- Installation option: HA k8s

Tried in downsteam clusters EKS and imported RKE2, K3s.

izaac:
Project owners are not getting access to the cluster either.

sowmyav27:
**On master-head - commit id: `db8a864f`**

- Install Rancher and create 2 Standard user user1 and user2
- user1 provisions a cluster (Created a single node custom cluster)
- As User1 add user2 as Cluster Member
- user2 --> Create a Project --> Create a namespace in this Project
- User2 is able to create a namespace

StrongMonkey:
This issue can only be reproduced if local cluster's k8s is v1.19 above

izaac:
Rancher version `master-head (09/21/2020)` _50572e1_ HA local k8s `v1.19.2`
Rancher version `master-head (09/21/2020)` _50572e1_ HA local k8s `v1.18.8`


RBAC is looking good with k8s 1.1.9 local cluster  I am able to create namespaces in projects that I am member or owner.
Project Owner has no issues either.

Quick pass in HA setup with local cluster using k8s 1.18. Looks good.

PR #29048, opened 2020-09-21
Fix role/clusterRole creation

StrongMonkey:
This commit fixed an issue where role/clusterRole are sent to the same
role Client for creation. This used to work before v1.19, but since
v1.19 is API server does validation which no longer allows it.


StrongMonkey:
https://github.com/rancher/rancher/issues/29021

StrongMonkey:
```
2020/09/21 17:46:27 [ERROR] error syncing 'p-kgqzp/creator-project-owner': handler mgmt-auth-prtb-controller: ClusterRole in version "v1" cannot be handled as a Role: converting (v1.ClusterRole) to (rbac.Role): unknown conversion, requeuing
```

dramich (approved):

Issue #28977, opened 2020-09-17
Adding a new restricted admin role


maggieliu:
For Rancher 2.5 local cluster needs to be enabled for most of the features to work. In order to address the security concern around admin having access to local cluster, there is a need to introduce a new restricted admin role that has access to all downstream clusters except local cluster.

Scenarios to consider:
1. I turn on the new restricted-admin role
2. That role gives people with access to it the permission to create GlobalRoleBindings and GlobalRoles
3. If a user with this role creates a GlobalRole that gives access to “/” (or simply “pods” for a more direct example) and then assigns that role to someone (or themselves) and they get access to those additional resources in the local cluster, it is a bad privilege escalation bug.

Similarly, if the restricted-admin user is able to create clusterRole-template-bindings in the local cluster, they would be able to assign themselves as cluster-owner in the cluster and it is again privilege escalation.

we might need a restrict-local-cluster-rbac flag that the user supplies in place of turning off local cluster. Its function would be to cause our rbac controllers to not “realize” crtb/prtbs in the local cluster

dramich:
There is a new restricted-admin role added to rancher. It will have admin access to all rancher CRD types except PRTBs and CRTBs with limited access to being able to create GRBs. For GRBs it can only assign a GRB that has rules the restricted admin already possesses. This will allow them to create new users and add them to rancher but not much else. 

The restricted-admin DOES NOT have access to any native Kubernetes types in the local cluster. This includes but is not limited to secrets, namespaces, RBAC objects etc. 

The restricted-admin DOES have full admin access in all downstream clusters. 

The restricted-admin can create other restricted-admins as that would be equal to their level of access.

dnoland1:
Started testing and hit an issue deploying helm chart with this option turned on. See https://github.com/rancher/rancher/issues/29186 . So wasn't able to get too far in trying this out.

dnoland1:
Hit https://github.com/rancher/rancher/issues/29279 while testing. Need to make sure we prevent the restricted admin from being able to modify or delete any of the rancher related workloads, namespaces, etc. in the local cluster.

dnoland1:
Getting an error when creating a cluster as a restricted admin. Filed here - https://github.com/rancher/rancher/issues/29324

dnoland1:
[...]

PR #29050, opened 2020-09-21
Add new restricted-admin role

dramich:
Problem:
With local cluster now being required the admin role will have full
access to the local cluster which is not good from a security
perspective

Solution:
Add a new restricted-admin role with limited access to the local cluster
but cluster-admin access to all downstream clusters. The
restricted-admin will not be able to create anything rbac related in the
local cluster but should have access otherwise.
Add new setting and update helm chart with the option to set the default admin created on first startup to the restricted-admin role. 

https://github.com/rancher/rancher/issues/28977

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -104,6 +104,10 @@ spec:
         - name: CATTLE_SYSTEM_CATALOG
           value: bundled
 {{- end }}
+{{- if .Values.restrictedAdmin }}
+        - name: CATTLE_RESTRICTED_DEFAULT_ADMIN
+          value: true
+{{- end}}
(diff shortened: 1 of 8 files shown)


cjellick (approved):
LGTM but can you make sure the UI will support displaying the rules as they are defined for your new role (where api group matters and resource and verbs are *. If not, UI might need tweaked)

dramich:
> LGTM but can you make sure the UI will support displaying the rules as they are defined for your new role (where api group matters and resource and verbs are *. If not, UI might need tweaked)

Confirmed that the UI is looking as it should.

cbron (approved):
Helm chart, syntax, and logic all look good to me. I have limited knowledge of rbac system though.

PR #29054, opened 2020-09-21
Use cattle-system instead of dashboard-shells for shell

StrongMonkey:
This commit switched to use cattle-system instead of dashboard-shells
for shell containers, in order to avoid patching dashboard-shells for
CIS compliance.

### DIFF
--- a/pkg/api/steve/clusters/clusters.go
+++ b/pkg/api/steve/clusters/clusters.go
@@ -17,7 +17,7 @@ import (
 func Register(ctx context.Context, server *steve.Server) error {
 	shell := &shell{
 		cg:           server.ClientFactory,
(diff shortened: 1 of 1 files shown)


StrongMonkey:
https://github.com/rancher/fleet/issues/94

prachidamle:
@StrongMonkey change is good, but making sure with QA about the `cattle-system` namespace which we will use instead. @sowmyav27 does `cattle-system` fail the hardened scan?

prachidamle:
@sowmyav27 confirmed that only `dashboard-shells` is failing the scan. We are good with `cattle-system`

prachidamle (approved):
LGTM

dramich (approved):

PR #29075, opened 2020-09-22
Set local rke2 cluster to rancherd, and prevent upgrades through rancher

MonzElmasry:
**Proposed change** 
- Add a separate driver for rancherd, which can be set when the local cluster is rke2
- Set the config for rancherd cluster to nil
issue : https://github.com/rancher/rke2/issues/201

### DIFF
--- a/pkg/controllers/management/clusterprovisioner/provisioner.go
+++ b/pkg/controllers/management/clusterprovisioner/provisioner.go
@@ -10,19 +10,15 @@ import (
 	"strings"
 	"time"
 
-	v1 "github.com/rancher/rancher/pkg/generated/norman/apps/v1"
-
-	apimgmtv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
-	rketypes "github.com/rancher/rke/types"
-
 	"github.com/mitchellh/mapstructure"
 	"github.com/rancher/norman/controller"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/slice"
 	"github.com/rancher/norman/types/values"
+	apimgmtv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	util "github.com/rancher/rancher/pkg/cluster"
 	kd "github.com/rancher/rancher/pkg/controllers/management/kontainerdrivermetadata"
+	v1 "github.com/rancher/rancher/pkg/generated/norman/apps/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/kontainer-engine/drivers/rke"
 	"github.com/rancher/rancher/pkg/kontainer-engine/service"
@@ -31,6 +27,7 @@ import (
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/rancher/rke/services"
+	rketypes "github.com/rancher/rke/types"
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


cjellick (approved):

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -1037,6 +1039,10 @@ func k3sBasedClusterConfig(cluster *v3.Cluster, nodes []*v3.Node) error {
 			cluster.Spec.K3sConfig.SetStrategy(1, 1)
 		}
 	} else if strings.Contains(cluster.Status.Version.String(), "rke2") {
+		if cluster.Name == "local" {
```
StrongMonkey, line 1042: How can we tell if user just install rancher on rke2 cluster but not from rancherd? does the provider also become rancherd?

cjellick: oh right. hmph.

StrongMonkey: Not sure if installing rancher on rke2 will be a use case, if so we need to tell the difference between rke2 cluster and a rancherd rke2 cluster. Right now we can say if it is rke2 then it is always going to be rancherd rke2.

cjellick: I would like to see this fixed in the initial release. As is, an rke2 cluster will look like rancherd and then if we fix it in a later release, the cluster's type would change from rancherd to rke2 upon upgrade, which may cause an issue.

Is there some simple thing we can look for in the cluster to "know" it was a rancherd2 install? Heck, if we just added a check for a daemonset named rancher, that would probably be a good enough indication. (This is because rancher is typically installed as a deployment, not a daemonset, but in the new chart that @StrongMonkey developed for rancherd, we changed that to a daemonset)

StrongMonkey: :+1:

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	v1 "github.com/rancher/rancher/pkg/generated/norman/apps/v1"
+
```
StrongMonkey, line 14: nit: empty space. In general if you use IDE like goland it will do this. You have to manually check imports to remove it

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -1037,6 +1042,14 @@ func k3sBasedClusterConfig(cluster *v3.Cluster, nodes []*v3.Node) error {
 			cluster.Spec.K3sConfig.SetStrategy(1, 1)
 		}
 	} else if strings.Contains(cluster.Status.Version.String(), "rke2") {
+
+		daemonsets, _ := p.DaemonsetLister.List("cattle-system", labels.Everything())
```
StrongMonkey, line 1046: Can you do a `GET` request in here? also return error

```
if error is not found error:
  set driver to rke2
else error is not nil:
  return error
else error is nil:
  set driver to rancherd
```

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -1037,7 +1039,16 @@ func k3sBasedClusterConfig(cluster *v3.Cluster, nodes []*v3.Node) error {
...
-		cluster.Status.Driver = apimgmtv3.ClusterDriverRke2
+
+		_, err := p.DaemonsetLister.Get("cattle-system", "rancher")
+		if apierrors.IsNotFound(err) {
+			cluster.Status.Driver = apimgmtv3.ClusterDriverRke2
+		} else if err != nil {
+			return err
+		} else {
+			cluster.Status.Driver = apimgmtv3.ClusterDriverRancherD
+			return nil
```
StrongMonkey, line 1050: do you need to return here? You will still need to set `cluster.Spec.Rke2Config` if it is rancherd, correct?

MonzElmasry: noo, the rke2-config only contains upgrade strategy, and we need to prevent upgrades so i think there should not be a config assigned to this cluster. 
https://github.com/rancher/rancher/blob/3ccd4d9164bd09640e1cd6ad059008e2a06f778d/pkg/api/norman/customization/cluster/validator.go#L231 
specially with this starting upgrades if the rke2Config exists

StrongMonkey (approved):
LGTM

cjellick (approved):

Issue #29085, opened 2020-09-22
Cannot deploy an EKS cluster, Panic seen in Rancher


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a hosted EKS cluster from Rancher (default values choosen except for Subnet and Security group)
- API request on a created gives a 500 Error
- Panic in rancher logs:

```
2020/09/22 16:34:09 [ERROR] Panic serving api request: 
goroutine 21157 [running]:
runtime/debug.Stack(0xc001aadcd0, 0x3bf9d00, 0x732a500)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9d
github.com/rancher/norman/api.(*Server).ServeHTTP.func1(0x4e39b40, 0xc00559fe40)
	/go/pkg/mod/github.com/rancher/norman@v0.0.0-20200908202416-992a35eef40f/api/server.go:175 +0x7a
panic(0x3bf9d00, 0x732a500)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/rancher/rancher/pkg/api/norman/customization/cluster.(*Validator).validateEKSConfig(0xc004130420, 0xc0029590e0, 0xc0060001e0, 0xc0074c9800, 0x0, 0x0)
	/go/src/github.com/rancher/rancher/pkg/api/norman/customization/cluster/validator.go:354 +0xd20
github.com/rancher/rancher/pkg/api/norman/customization/cluster.(*Validator).Validator(0xc004130420, 0xc0029590e0, 0xc004d230e0, 0xc0060001e0, 0x0, 0xc0060001e0)
	/go/src/github.com/rancher/rancher/pkg/api/norman/customization/cluster/validator.go:75 +0x3e9
github.com/rancher/norman/parse/builder.(*Builder).Construct(0xc0060001b0, 0xc004d230e0, 0xc006000000, 0x44ab469, 0x6, 0x44a7e7d, 0x4, 0xc006d09501)
	/go/pkg/mod/github.com/rancher/norman@v0.0.0-20200908202416-992a35eef40f/parse/builder/builder.go:58 +0xe2
github.com/rancher/norman/api/handler.ParseAndValidateBody(0xc0029590e0, 0xc001aaee01, 0x2a9ed3e, 0xc001d6dd78, 0xc0029590e0)
[...]

PR #29087, opened 2020-09-22
Fix nil in credential check

rmweir:
**Problem:**
Previous cluster is being use without checking if it is nil.

**Solution:**
Check if previous cluster is nil.

**Issue:**
https://github.com/rancher/rancher/issues/29085

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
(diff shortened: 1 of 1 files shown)


paynejacob (approved):

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -455,16 +454,16 @@ func validateEKSKubernetesVersion(spec *v32.ClusterSpec, prevCluster *v3.Cluster
...
+	credentialErr := "error accessing cloud credential"
 	if err := access.ByID(request, &mgmtSchema.Version, mgmtclient.CloudCredentialType, credential, &accessCred); err != nil {
 		return httperror.NewAPIError(httperror.NotFound, credentialErr)
 	}
 
-	if prevSpec != nil {
+	if prevCluster != nil {
 		// validate the user has access to the old cloud credential before allowing them to change it
-		credential = prevSpec.AmazonCredentialSecret
+		credential = prevCluster.Spec.EKSConfig.AmazonCredentialSecret
```
cbron, line 466: EKSConfig can be nil also, we have removed that check

cbron (approved):

Issue #28983, opened 2020-09-18
Allow one main Rancher install to manage multiple downstream Rancher installs with local cluster


maggieliu:
There is a need to all for a main Rancher install that manages multiple downstream Rancher installs with local cluster.

dnoland1:
I've hit a few issues with having Rancher v2.5.0-rc2 manage Rancher 2.4.8 clusters which I have reported here:

- https://github.com/rancher/rancher/issues/29238
- https://github.com/rancher/rancher/issues/29239

dnoland1:
Having better luck with using Rancher 2.4.8 to manage Rancher 2.5.0-rc clusters.

dnoland1:
I'm okay closing this out for now and will go with the strategy of using Rancher v2.4.8 to manage Rancher v2.4.x/v2.5.x clusters. Once all managed clusters are on 2.5.x, will upgrade main Rancher to v2.5.x. If I run into any issues with us, will open new GH issues.

PR #29094, opened 2020-09-22
Set leader election id differently for agent

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/28983

PR #29120, opened 2020-09-23
add a parameter for cert-manager version

jiaqiluo:
this PR allows us to pass in the cert-manager's version in automation

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -196,6 +196,7 @@ if_test_group_rbac = pytest.mark.skipif(
 test_run_id = "test" + str(random.randint(10000, 99999))
 RANCHER_HOSTNAME_PREFIX = os.environ.get("RANCHER_HOSTNAME_PREFIX",
                                          test_run_id)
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/common.py
```diff
@@ -196,6 +196,7 @@
 test_run_id = "test" + str(random.randint(10000, 99999))
 RANCHER_HOSTNAME_PREFIX = os.environ.get("RANCHER_HOSTNAME_PREFIX",
                                          test_run_id)
+CERT_MANAGER_VERSION = os.environ.get("RANCHER_CERT_MANAGER_VERSION", "v1.0.1")
```
sowmyav27, line 199: @jiaqiluo Is the default one going to be this `"v1.0.1"` or `0.15.0` ?

sangeethah (approved):
LGTM

tests/validation/tests/v3_api/common.py
jiaqiluo: @sowmyav27 
1.0.1 will be supported in the coming rancher v2.4.9 and v2.5.0, so it should be ok for us to leave the default to be 1.0.1

we need to pass the value 0.15.0 for installing older version of Rancher

sowmyav27 (approved):
LGTM

PR #29129, opened 2020-09-24
Add permission to execute shell for cluster member

StrongMonkey:
This commit add a permission to allow cluster member to access
v1/management.cattle.io/clusters since it is required for executing
shells

### DIFF
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -140,7 +140,8 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext) 
 		addRule().apiGroups("management.cattle.io").resources("clustermonitorgraphs").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("catalogtemplates").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("catalogtemplateversions").verbs("get", "list", "watch").
-		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch")
+		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("clusters").resourceNames("local").verbs("get")
 
 	rb.addRoleTemplate("Create Projects", "projects-create", "cluster", false, false, false).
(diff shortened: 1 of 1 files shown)


StrongMonkey:
@dramich I tested with a user token and tried to access local cluster, it wouldn't give any permission.

```
➜  rancher-operator git:(fix-server-url) ✗ k get clusters.management.cattle.io local -o yaml
Error from server (Forbidden): clusters.management.cattle.io "local" is forbidden: User "u-w79md" cannot get resource "clusters" in API group "management.cattle.io" at the cluster scope
```

I feel like if adding these would affect the upstream cluster(local), a lot of things could screw up. 

For example:
https://github.com/rancher/rancher/blob/d4176fba2ed888a7ba2678bb38b5e2eb22de05f5/pkg/data/management/role_data.go#L130

This could allow access to all the nodes in upstream cluster.

So I think this only impacts downstream cluster, which is the cluster that you are assigned as a member. :thinking:

StrongMonkey:
https://github.com/rancher/rancher/issues/28223

pkg/data/management/role_data.go
```diff
@@ -140,7 +140,8 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext)
 		addRule().apiGroups("management.cattle.io").resources("clustermonitorgraphs").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("catalogtemplates").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("catalogtemplateversions").verbs("get", "list", "watch").
-		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch")
+		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch").
+		addRule().apiGroups("management.cattle.io").resources("clusters").resourceNames("local").verbs("get")
```
mrajashree, line 144: won't this give member of any cluster permission to get the local cluster? plus cluster-member could launch shell on 2.4 without this permission

ibuildthecloud (approved):

pkg/data/management/role_data.go
StrongMonkey: @mrajashree This will only impacts downstream cluster, see https://github.com/rancher/rancher/pull/29129#issuecomment-698053373

Issue #29099, opened 2020-09-22
Changes to the labels for PRTB/CRTB


mrajashree:
Most of the review comments are addressed in this PR https://github.com/rancher/rancher/pull/28423
Some of the pending ones that don't affect functionality are:

- [x]  Using multierror

- [x]  Renaming the labels in some places of the code

- [x]  Moving the helper func to a common pkg. 

 
Based on offline discussion with @dramich opening a separate issue to track these code changes

mrajashree:
Closing as the PR is merged

PR #29133, opened 2020-09-24
Adds a new label to CRB/RB created for a CRTB/PRTB

mrajashree:
**Problem**: Upgrading to 2.5 adds a new label containing the CRTB/PRTB namespace and name
to the CRB/RBs created for the it. Both labels have the same value "memberhsip-binding-owner".
When a CRTB/PRTB is deleted, the indexer gets all RTBs with this label value. If there are more
than one labels with this value, the CRB/RB is not deleted. On upgrade there will be two labels with
the same value, one is the existing label with UID as key and the other is the new label with ns+name

**Solution**: Change the label value. With this, on an upgraded setup, the controllers will only check if
the RB/CRB has a label with the new value. There will be only one such label, so on CRTB/PRTB deletion
the RB/CRB will get deleted. This is also required when migrating an upgraded setup using backup-restore-operator

https://github.com/rancher/rancher/issues/28370
Also addresses remaining comments from https://github.com/rancher/rancher/pull/28423
https://github.com/rancher/rancher/issues/29099

### DIFF
--- a/pkg/controllers/management/auth/crtb_handler.go
+++ b/pkg/controllers/management/auth/crtb_handler.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	pkgrbac "github.com/rancher/rancher/pkg/rbac"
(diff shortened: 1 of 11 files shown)


pkg/controllers/management/auth/crtb_handler.go
```diff
@@ -15,16 +16,17 @@ import (
...
-	crtbInProjectBindingOwner   = "crtb-in-project-binding-owner"
-	prtbInClusterBindingOwner   = "prtb-in-cluster-binding-owner"
-	rbByOwnerIndex              = "auth.management.cattle.io/rb-by-owner"
-	rbByRoleAndSubjectIndex     = "auth.management.cattle.io/crb-by-role-and-subject"
-	ctrbMGMTController          = "mgmt-auth-crtb-controller"
-	rtbLabelUpdated             = "auth.management.cattle.io/rtb-label-updated"
-	rtbCrbRbLabelsUpdated       = "auth.management.cattle.io/crb-rb-labels-updated"
+	clusterResource              = "clusters"
+	membershipBindingOwnerLegacy = "memberhsip-binding-owner"
+	membershipBindingOwner       = "membership-binding-owner"
```
dramich, line 21: Can you add comments here to clarify the two labels and call out that they are indeed different

tests/integration/suite/test_rbac.py
```diff
@@ -280,6 +279,70 @@ def fail_handler():
...
+
+    # Find the expected k8s clusterRoleBinding
+    crbs = api_instance.list_cluster_role_binding(
+        label_selector=prtb_key + "=" + mbo)
+
+    assert len(crbs.items) == 1
+
+    # edit this CRB to add in the legacy label to simulate an upgraded setup
+    crb = crbs.items[0]
+    crb.metadata.labels[prtb.uuid] = mbo_legacy
```
dramich, line 323: This isn't being updated in k8s

tests/integration/suite/test_rbac.py
```diff
@@ -280,6 +279,70 @@ def fail_handler():
...
+    def fail_handler():
+        return "failed waiting for cluster role binding to be deleted"
+
+    wait_for(crb_callback, fail_handler=fail_handler)
+
+    try:
+        cluster = user_mc.client.by_id_cluster(admin_cc.cluster.id)
+        assert cluster is None
+    except ApiError as e:
+        assert e.error.status == 403
```
dramich, line 343: Use `with pytest.raises(ApiError) as e:` if you are expecting a call to error

dramich (approved):

PR #29138, opened 2020-09-24
Adding a check to validate all workloads have come up active in System project

sowmyav27:
Adding a check to validate all workloads have come up active in System project before running the hardening steps on the clusters.


izaac (changes requested):
This looks good but this is duplicate code that's also in the `validate_cluster` method.

It's a short piece of code, but it could prove helpful on other use cases so it makes sense to move this to its own method like `check_system_project_workloads_active`. (It's perhaps a long name but just to get the idea).

PR #29140, opened 2020-09-24
Match the rancher helm and kubectl versions to the shell

cbron:
https://github.com/rancher/shell/blob/c1f4422edb59e33aa08c58410d76f9165077211d/package/Dockerfile

I tried to check the support matrix but [kube docs](https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl) say that kubectl only supports one version above and below. So if we are 1.19->1.16 then 1.18 should likely be most correct.


brendarearden (approved):
👍 looks good!

dramich (approved):

StrongMonkey (approved):

PR #29141, opened 2020-09-24
upload rancherd tarball as release artifact

dweomer:
This fixes an issue that was creating the tarball under dist/ instead of
bin/ which prevented it from being uploaded as a release artifact.

Supplemental to rancher/rancher#28922
Addresses rancher/rancher#28131

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -64,11 +64,11 @@ fi
 
 TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go build/system-charts build/charts $IMAGE $AGENT_IMAGE
 
-if [ "$ARCH" = "amd64" ]; then
+if [ ${ARCH} == amd64 ]; then
     # rancherd tarball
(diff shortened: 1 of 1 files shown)


MonzElmasry (approved):

scripts/package
```diff
@@ -64,11 +64,11 @@ fi
 
 TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go build/system-charts build/charts $IMAGE $AGENT_IMAGE
 
-if [ "$ARCH" = "amd64" ]; then
+if [ ${ARCH} == amd64 ]; then
```
cbron, line 67: Curious if this is a fix or just a best practice

StrongMonkey (approved):

scripts/package
dweomer: Alignment with what the script is doing above.

cbron (approved):

Issue #29143, opened 2020-09-24
air-gap - Fleet-related pods do not use the private registry in HA air-gap installation 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug 

**Steps to reproduce (least amount of steps as possible):**
- install Rancher: `v2.5.0-rc1` in airgap env in HA mode 
- here is the helm command to generate the template
```
 helm template rancher ./rancher-2.5.0-rc1.tgz --output-dir . \
    --namespace cattle-system \
    --set hostname=jiaqi0235-15555-internal.qa.rancher.space \
    --set rancherImage=ec2-18-222-107-174.us-east-2.compute.amazonaws.com/rancher/rancher \
    --set ingress.tls.source=secret \
    --set systemDefaultRegistry=ec2-18-222-107-174.us-east-2.compute.amazonaws.com \
    --set useBundledSystemChart=true
```
- and install rancher 
```
kubectl create namespace cattle-system
kubectl -n cattle-system apply -R -f ./rancher
```

**Result:**

The following pods do not use the private registry so fail to pull images.
They are all related to Fleet.

```
> kubectl get pods -A | grep -i ImagePullBackOff
cattle-system             rancher-webhook-5764f986cf-9x7q9           0/1     ImagePullBackOff   0          28m
fleet-system              fleet-controller-5cd4bc9cd7-c78kw          0/1     ImagePullBackOff   0          27m
fleet-system              gitjob-568d66bc7f-9n6tz                    0/1     ImagePullBackOff   0          27m
rancher-operator-system   rancher-operator-75499d5fcb-qvfmq          0/1     ImagePullBackOff   0          29m
```

one example:
```
  Back-off pulling image "rancher/rancher-webhook:v0.1.0-alpha8"
```

StrongMonkey:
This is because the fleet-components are launched in rancher instead of dashboard and needs to be applied with default_registry_settings.

jiaqiluo:
the bug fix is validated on rancher:v2.5.0-rc2 HA airgap setup 

we see fleet-controller, gitjob, and rancher-operator pulls the image from the private registry 

but rancher-webhook does not get deployed, which is tracked by this issue https://github.com/rancher/rancher/issues/29189

PR #29174, opened 2020-09-25
Fix airgap for fleet

StrongMonkey:
This commit add global registry setting and pass it to the fleet charts
that are installed by controller, since these settings can't be passed
from UI.

### DIFF
--- a/pkg/controllers/dashboard/fleetcharts/controller.go
+++ b/pkg/controllers/dashboard/fleetcharts/controller.go
@@ -46,7 +46,8 @@ func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error
 	}
 
 	if setting.Name != settings.ServerURL.Name &&
-		setting.Name != settings.CACerts.Name {
+		setting.Name != settings.CACerts.Name &&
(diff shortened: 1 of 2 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/29143

pkg/controllers/dashboard/fleetcharts/controller.go
```diff
@@ -55,9 +55,15 @@ func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error
 		return setting, err
 	}
 
+	systemGlobalRegistry := map[string]interface{}{
+		"cattle": map[string]interface{}{
+			"systemDefaultRegistry": settings.SystemDefaultRegistry.Get(),
```
ibuildthecloud, line 60: This needs to be on L49 also

StrongMonkey: added

cbron (approved):
I don't have wider context here but seems like it could work

ibuildthecloud (approved):

Issue #29172, opened 2020-09-25
App upgrade fails with error "has no deployed releases"


sowmyav27:
On master-head - commit id: `4ac3a87d7`

- Deploy logging chart with version 3.4.0
- Upgrade app to 3.4.6 once it finishes deploying.
- App fails to upgrade with error

```
helm --debug upgrade --history-max=5 --timeout=10m0s --values=/home/shell/helm/values-rancher-logging-3.6.000.yaml --version=3.6.000 --wait=true rancher-logging /home/shell/helm/rancher-logging-3.6.000.tgz
--
Fri, Sep 25 2020 1:20:21 pm | upgrade.go:121: [debug] preparing upgrade for rancher-logging
Fri, Sep 25 2020 1:20:21 pm | Error: UPGRADE FAILED: "rancher-logging" has no deployed releases
Fri, Sep 25 2020 1:20:21 pm | helm.go:94: [debug] "rancher-logging" has no deployed releases
```

Happens with cis v2 scan and backup restore app as well.

sowmyav27:
**On master-head commit id: `ec04f7878`**

- Deploy logging chart with version 3.4.0
- Upgrade app to 3.4.6 once it finishes deploying.
- App upgrade succeeds. 
- Validated using backup-restore and CIS benchmark apps also. Upgrade works

PR #29190, opened 2020-09-26
Add annotations to chart install/upgrade and fix namespace param

ibuildthecloud:


maggieliu:
related to: https://github.com/rancher/rancher/issues/29172

Issue #29183, opened 2020-09-23
Change events don't work reliably in general


vincent99:
There are many issues with the same root-cause of change events aren't sent in general, or contain incorrect data.  We don't really need more reports of different areas, it can affect any of the hundreds of types.

rancher/dashboard#1339              CIS
rancher/dashboard#1315              Schemas/missing links
rancher/dashboard#1266              Namespaces
rancher/dashboard#1246              Backup
rancher/dashboard#1065              Configmaps/counts
rancher/dashboard#1012              Deployments
rancher/dashboard#371              Gatekeeper constraints

sowmyav27:
**On master-head - commit id: `27000be7e`**

State of CIS scans do not change. Refresh of the page is required

sowmyav27:
CIS scan state change work sometimes and sometimes it does not.

maggieliu:
There is an UI part that needs to be fixed also before moving this to test.

maggieliu:
UI PR: https://github.com/rancher/dashboard/pull/1534

sowmyav27:
**On master-head - commit id: `d4b3bc89`**

**Still seen as an Issue:**
- State of CIS scans do not change. Refresh of the page is required
- https://github.com/rancher/dashboard/issues/1266 is also reproducible

Reopening this issue

maggieliu:
@sowmyav27 Can you try to reproduce again and capture the websocket messages in network Tab of devTools

sowmyav27:
**On master-head - commit id: `d4b3bc89`**

- CIS scan state change works sometimes. When it works notice resource.create calls in WS messages tab (dev tools). When state of cluster scan does not change - WS messages do not contain resource.create for `resourceType:"cis.cattle.io.clusterscanreport"` 
- Create namespace --> No WS messages related to resource.create seen for namespace creation.

vincent99:
Found & fixed another issue which explains the randomness a lot easier than the edge cases we previously fixed with the server side and what happens in error conditions.

If you switch clusters, the dashboard closed the connection to the old cluster like it's supposed to, and opened a new connection...  But still to the old cluster, not the new one.  So if you switch clusters, no events until you reload the page.

sowmyav27:
**On master-head - commit id: `cb9e8a105c1`**
- State of cluster scan does not change - WS messages do not contain resource.create for `resourceType:"cis.cattle.io.clusterscanreport"` 
[...]

PR #29200, opened 2020-09-28
Ensure resource.error is sent when revision is too old

ibuildthecloud:


maggieliu:
related to: https://github.com/rancher/rancher/issues/29183

PR #29215, opened 2020-09-28
Do not look at the automount_service_account_token flag of default SA in non-system namespaces

prachidamle:
The test test_system_project.py is failing for `fleet-system` namespace. It is not present in system_namespaces setting, hence should the test should not error out for it irrespective of the value of the automount_service_account_token flag of the default SA.

Also adding fleet-system to system-namespaces setting.

### DIFF
--- a/tests/integration/suite/test_system_project.py
+++ b/tests/integration/suite/test_system_project.py
@@ -10,7 +10,8 @@ initial_system_namespaces = set(["kube-node-lease",
                                  "cattle-system",
                                  "kube-public",
                                  "cattle-global-data",
-                                 "cattle-global-nt"])
+                                 "cattle-global-nt",
+                                 "fleet-system"])
 loggingNamespace = "cattle-logging"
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/suite/test_system_project.py
```diff
@@ -93,12 +94,7 @@ def _check_system_sa_flag():
                 else:
                     return False
             else:
-                if sa.automount_service_account_token is None:
```
dramich, line 96: I think we should leave this in, if it wasn't for the test we wouldn't have noticed fleet-system being missing

prachidamle: you mean we keep the test as-is and error out on non-system ns if the flag is found to be false? But I think that seems wrong - if the namespace is not in system_ns, we should not care for the automount_service_account_token really.

prachidamle: CIS scan tests will start failing for all the out-of-the-box namespaces that don't set this flag and that way we will know?

dramich (approved):

tests/integration/suite/test_system_project.py
dramich: Yeah that sounds good.

PR #29224, opened 2020-09-29
Update test for webhook check

dramich:
Problem:
The webhook now enforces that only users with `**` in the local cluster
can create PRTBs so this causes the test to fail

Solution:
Update the test to do listing actions on PRTBs

Example: https://drone-pr.rancher.io/rancher/rancher/11319/1/2


mrajashree (approved):
LGTM! 👍

Issue #29205, opened 2020-09-18
403 forbidden error for editing a deployment is seen in the API request/response and not in the UI


sowmyav27:
**On master-head - commit id: `8fd81a9b46`**

- Add user1 with Project Read only permissions to a project in a cluster.
- Login as the user, navigate to the dashboard page.
- Edit an existing Deployment.
- The error is seen in the console/Network tab and not in the UI
- Error `{"type":"error","links":{},"code":"Forbidden","message":"deployments.apps \"test\" is forbidden: User \"u-4cnrgx47ty\" cannot update resource \"deployments\" in API group \"apps\" in the namespace \"default\"","status":403,"type":"error"}`

Can we display the error in the UI ?

sowmyav27:
**On master-head - commit id: `c3bc230f`**

- Add user1 with Project Read only permissions to a project in a cluster.
- Login as the user, navigate to the dashboard page.
- Edit option is NOT available on the deployment for this user.

<img width="275" alt="Screen Shot 2020-09-29 at 12 37 10 PM" src="https://user-images.githubusercontent.com/26032343/94607401-b60ee200-0250-11eb-8de6-5c8f6e0433a0.png">

PR #29241, opened 2020-09-29
Don't put update link if update isn't allowed

ibuildthecloud:


ibuildthecloud:
https://github.com/rancher/rancher/issues/29205

Issue #29222, opened 2020-09-28
[rancherd] Cannot access clusters as admin user with generated kubeconfig


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Bring up a rancherd cluster
2. Add a cluster to the rancher setup using a non-admin user
3. Use the generated kubeconfig to try to access the cluster as admin user 

**Result:**
Get forbidden access errors when performing any kubectl command

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.0-rc1
- Installation option (single install/HA): HA with rancherd, installed using binary

sowmyav27:
**On master-head - commit id: `c3bc230fe`**

- deploy rancher - HA setup
- As a standard user, deploy a downstream cluster - DO 1 node master and 1 worker node
- Global admin is able to access the cluster/resources through kubectl shell. No issues seen

StrongMonkey:
@rancher-max Please re-test this with the latest rancherd from master.

rancher-max:
#### Validated using master rancherd build with v2.5.0-rc4 runtime image

Steps to run rancherd from master:
```
wget -O rancherd https://releases.rancher.com/rancherd/rancherd-master

sudo RKE2_RUNTIME_IMAGE=rancher/rancher-runtime:v2.5.0-rc4 rancherd server --write-kubeconfig-mode 644 2>&1 &

sudo rancherd reset-admin
```
- Validated global admin user can access local cluster through kubectl shell and generated kubeconfig file
- Created a downstream cluster with a different user (standard user) and validated both that user and global admin could access it through kubectl shell and generated kubeconfig files
- Validated that a third standard user who was a project member to downstream cluster could access it through kubectl shell and generated kubeconfig file

PR #29257, opened 2020-09-30
Fix rancherd not setting admin on local clusters

StrongMonkey:
This commit fixed an issue where local admin is not set as the cluster
owner of local cluster. This is because when we disable the default
admin, all the controller won't be able to rerun twice since all the
conditions have been set to true. Thus in reset-admin, when we first
bootstrap the password, we need to set creatorId on the annotation and
also revert the condition to be false so that controller can be re-run
to add corresponding roles/rolebindings

### DIFF
--- a/cmd/rancherd/auth/auth.go
+++ b/cmd/rancherd/auth/auth.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"context"
 	"fmt"
+	"os"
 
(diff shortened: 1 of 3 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/29222

dramich (approved):

MonzElmasry (approved):

Issue #29231, opened 2020-09-29
old container show up as unavailable Node when upgrading using docker install


tfiduccia:
Steps:
1.  Create a vm
2.  Install rancher using docker
3.  Log into Rancher and go to local nodes (which always seems to be here now, even if I don't use the flag?)
4.  Go back to terminal and rm the docker container and upgrade to new master
5.  Log into Rancher and go to Nodes

Results: Container I removed is still showing up as a node and now my local cluster thinks it's broken cause it can't find the old node.
ember:
![image](https://user-images.githubusercontent.com/11514927/94538751-41f51f80-01f9-11eb-958c-1d16975df51c.png)

vue:
![image](https://user-images.githubusercontent.com/11514927/94538842-5cc79400-01f9-11eb-8717-068535c779cb.png)

Comment:  The container is also still showing up in all the node drop-downs for workloads.

sowmyav27:
**Reproduced the issue on master-head - commit id: `ec04f7878`**

- On a single node install -- `docker run -v /home/ubuntu/state:/var/lib/rancher -d -p 80:80 -p 443:443 --restart=unless-stopped --privileged rancher/rancher:master-head`
- Rancher deploys. Local cluster is deployed. looks fine.
- In the VM, stop and delete the above container.
- Run a new docker run command ^^^ on this VM

<img width="1450" alt="Screen Shot 2020-09-29 at 10 59 08 AM" src="https://user-images.githubusercontent.com/26032343/94599576-7a224f80-0245-11eb-94f7-868b6d4b8374.png">

- VUE UI:
<img width="1905" alt="Screen Shot 2020-09-29 at 11 06 05 AM" src="https://user-images.githubusercontent.com/26032343/94599584-7d1d4000-0245-11eb-85ef-fb0e334d4787.png">

- Unavailable node shows up in scheduling while deploying a workload:
<img width="418" alt="Screen Shot 2020-09-29 at 11 11 57 AM" src="https://user-images.githubusercontent.com/26032343/94599605-873f3e80-0245-11eb-9810-d7ba54e48513.png">

cbron:
Workaround is just deleting old node by hand, so moving to 2.5.1. Ideally we can handle there automatically, which might be somewhat tricky. We could also [allow to delete in UI](https://github.com/rancher/rancher/blob/b4891bc9ab51711b9947aa6b697ed98758298d47/pkg/api/norman/customization/node/node.go#L47-L56) but thats still annoying.

[...]

PR #29259, opened 2020-09-30
Set node-name to local-node in the Docker container

ibuildthecloud:


ibuildthecloud:
https://github.com/rancher/rancher/issues/29231

Issue #29209, opened 2020-09-28
rancherd install script needs to use channel-server at update.rancher.io


cjellick:
It currently, presumably, would just use the "latest" release on github. But Rancher will quickly have a latest release from an older branch, so we need better control over what it uses.

rancher-max:
This is validated for rancherd using the following options:

```
$ curl -sfL https://raw.githubusercontent.com/rancher/rancher/master//cmd/rancherd/install.sh | sudo INSTALL_RANCHERD_CHANNEL=latest sh -

$ curl -sfL https://raw.githubusercontent.com/rancher/rancher/master//cmd/rancherd/install.sh | sudo INSTALL_RANCHERD_CHANNEL=testing sh -

$ curl -sfL https://raw.githubusercontent.com/rancher/rancher/master//cmd/rancherd/install.sh | sudo INSTALL_RANCHERD_CHANNEL=stable sh -
```

Stable returns 404 currently as it attempts pull from: `[INFO]  downloading checksums at https://github.com/rancher/rancher/releases/download/stable/sha256sum.txt`

Testing pulls from latest rc: `[INFO]  downloading checksums at https://github.com/rancher/rancher/releases/download/v2.5.0-rc3/sha256sum.txt`

Latest pulls from latest actual released version: `[INFO]  downloading checksums at https://github.com/rancher/rancher/releases/download/v2.4.8/sha256sum.txt`

**I don't think stable will ever work? If this is expected behavior though, we can close this.**

cjellick:
A couple things:
1. You should be test this using get.rancher.io rather than the raw url. But that url isn't resolving to the right thing right now. I've opened an issue with EIO to fix that. @davidnuzik please track that ticket and make sure max knows when its done.
2.  @MonzElmasry can you make the default channel in the script "latest" so that the script will work without having to supply a channel at all.
3. Stable won't work until we introduce a "stable channel. But, we don't need to do that until this graduates from preview.

cjellick:
for now, latest will be funky until 2.5.0 is release. thats why latest resolves to 2.4.8. once 2.5.0 is out, something in the 2.5 release will always be “latest”.

We should make “latest” be the same as “testing” until 2.5.0 is officially released.

I need to have the install script working by default before the official release, without extra parameters, even if that means it installs an RC.

On the day of the 2.5 release, we’ll need to update the latest channel in channels.yaml back to having excludeRegexp: ^[^+]+-

[...]

PR #29277, opened 2020-09-30
set latest channel as default for rancherd install.sh

MonzElmasry:
This pr sets the latest channel -if not specified via INSTALL_RANCHERD_CHANNEL- as the default install channel. 
issue : https://github.com/rancher/rancher/issues/29209

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,7 +1,6 @@
 channels:
 - name: latest
   latestRegexp: .*
-  excludeRegexp: ^[^+]+-
 - name: testing
   latestRegexp: .*
 github:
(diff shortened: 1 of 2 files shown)


dweomer (approved):

StrongMonkey (approved):

BloodyIron:
So why wasn't the installation documentation updated to actually state this, and then provide a method to switch channels in the same documentation? I've had to go down so many rabbit holes just to find this single merge, and have had zero success getting RancherD working after 3x attempts so far.

Gee, I wonder why...

vincent99:
Stable and latest are the same version almost all the time except right after a major release (e.g. 2.4.x -> 2.5.0).  This is not related to your problem,

BloodyIron:
Then why even have two branches?

vincent99:
Because during those times we want the community using the newest by default while more conservative users can opt for the older more stable one.

Issue #18221, opened 2019-02-19
vSphere deletes PVs when deleting single node


horihel:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug Report

**Steps to reproduce (least amount of steps as possible):**
* Create a K8S-Cluster on vSphere (inkl. steps to enable vSphere storage provisioning)
* Create a workload with a volume claim
* Wait for it to be deployed
* Delete the node (in node view) where the pod is running on

**Result:**
* VM gets deleted
* All volumes connected to this VM get deleted as well
* Rancher recreates the VM according to node template
* Rancher/K8S tries to redeploy the workload, but fails, because the volume has been deleted earlier
* PVC needs to be deleted in order to proceed to a working state again
* data in the original volume is lost

**Other details that may be helpful:**
From my limited understanding it looks like Rancher is telling Vsphere to "Delete from disk" the VM, which causes vSphere to delete the VM and all VMDKs currently attached to it. But the volumes created by the provisioner should not be deleted.
This is also a user error, as the node should be drained before deleting it, but in the case of a node not responding, this might not be possible any more.
Maybe Rancher should disconnect all volumes before deleting the VM.

**Environment information**
- Rancher version 2.1.6
- Installation option : single install

**Cluster information**
- Cluster type: Infrastructure Provider
- Machine type: VM and specifications: 7 VMs, 2 CPUs each and 4GB RAM each
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.5", GitCommit:"32ac1c9073b132b8ba18aa830f46b77dcceb0723", GitTreeState:"clean", BuildDate:"2018-06-21T11:46:00Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.6", GitCommit:"b1d75deca493a24a2f87eb1efde1a569e52fc8d9", GitTreeState:"clean", BuildDate:"2018-12-16T04:30:10Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):

```
17.3.2
```

gzrancher/rancher#8983

gzrancher/rancher#11020

gzrancher/rancher#12087

gzrancher/rancher#11675

gzrancher/rancher#12210

LinAnt:
[...]

PR #29320, opened 2020-10-02
[v2.4] Vsphere Drain on Delete for Cloud Providers

luthermonson:
vSphere Clusters which use the intree cloud provider we setup via RKE were deleting the mounted volumes when nodes were deleted. To fix this we will be forcing a drain before the node is deleted, this will force the cloud provider to unmount the volume before the delete is called. Additionally, the v1 node deletion which happens in the user-node-remove controller was moved to the management node-controller and user-node-remove is now gone completely with node Updated now containing logic to strip the old finalizer.

https://github.com/rancher/rancher/issues/18221


paynejacob (dismissed):

pkg/controllers/management/node/controller.go
```diff
@@ -717,3 +744,81 @@ func (m *Lifecycle) updateRawConfigFromCredential(data map[string]interface{}, r
...
+			DeleteLocalData: false,
+			GracePeriod:     60,
+			Timeout:         60,
+		}
+	}
+
+	ctx, cancel := context.WithTimeout(m.ctx, time.Duration(node.Spec.NodeDrainInput.Timeout)*time.Second)
+	defer cancel()
+
+	_, msg, err := kubectl.Drain(ctx, kubeConfig, node.Status.NodeName, nodehelper.GetDrainFlags(node))
```
rmweir, line 809: Are you sure the error should not be checked first?

pkg/node/node.go
```diff
@@ -173,3 +178,91 @@ func IsMachineReady(machine *v3.Node) bool {
...
 }
+
+func DrainBeforeDelete(node *v3.Node, cluster *v3.Cluster) bool {
+	// never drain an etcd node, you will regret it
+	if IsEtcd(node.Status.NodeConfig) {
+		return false
+	}
+
+	// vsphere nodes with cloud providers need to be drained to unmount vmdk files as vsphere deletes them automatically
+	if nil != cluster.Spec.RancherKubernetesEngineConfig.CloudProvider.VsphereCloudProvider {
```
rmweir, line 189: Is it possible for CloudProvider to ever be nil here?

pkg/controllers/management/node/controller.go
```diff
@@ -76,13 +81,17 @@ func Register(ctx context.Context, management *config.ManagementContext) {
 		clusterLister:             management.Management.Clusters("").Controller().Lister(),
 		schemaLister:              management.Management.DynamicSchemas("").Controller().Lister(),
 		credLister:                management.Core.Secrets("").Controller().Lister(),
+		userManager:               management.UserManager,
+		clusterManager:            clusterManager,
+		nodeInterface:             management.Core.Nodes(""),
```
rmweir, line 86: I don't think nodeInterface is ever use, can delete if that's the case.

rmweir (changes requested):

pkg/controllers/management/node/controller.go
luthermonson: good eye

pkg/controllers/management/node/controller.go
luthermonson: there were cases i've now fixed a bit (e.g. trying to drain an etcd node) that put me into a really bad place. ultimately eating this error puts us right back to where we were when we started this so felt it was best to guarantee we move past the drain and ultimately make the delete call to vsphere. but if the drain never works and you want the node gone i dont think it should block the deleteNode call. If the drain never works and we're catching this error the node will get stuck in a loop and never delete at all. i'm open to retrying 3x times or something? maybe the error bubbles up and you go fix it and the node drains the second time?

pkg/node/node.go
luthermonson: ya it's not a pointer, that's why it worked and VsphereCloudProvider is a pointer hence the nil check.

pkg/controllers/management/node/controller.go
rmweir: Yeah but that would require a fairly quick manual intervention, it can probably stay as is in that case.

pkg/controllers/management/node/controller.go
```diff
@@ -717,3 +742,81 @@ func (m *Lifecycle) updateRawConfigFromCredential(data map[string]interface{}, r
...
+
+	if node.Spec.NodeDrainInput == nil {
+		logrus.Debugf("node [%s] has no NodeDrainInput, creating one with 60s timeout",
+			node.Spec.RequestedHostname)
+		node.Spec.NodeDrainInput = &v3.NodeDrainInput{
+			Force:           false,
+			DeleteLocalData: false,
+			GracePeriod:     60,
+			Timeout:         60,
+		}
```
rmweir, line 801: Node object needs to be updated and should be using a copy.

rmweir: I can see why you might not need to update the node for this to work, perhaps you shouldn't assign it to the obj if it's not used anywhere else. Something like `nodeDrainInput = node.Spec.NodeDrainInput; if nodeDrainInput == nil ...`

pkg/controllers/management/node/controller.go
```diff
@@ -717,3 +742,81 @@ func (m *Lifecycle) updateRawConfigFromCredential(data map[string]interface{}, r
...
+	select {
+	case <-ctx.Done():
+		if ctx.Err() != nil {
+			logrus.Errorf("node [%s] drain failed: %s", node.Spec.RequestedHostname, ctx.Err())
+		}
+		if err != nil {
+			// kubectl failed continue on with delete any way
+			logrus.Errorf("node [%s] kubectl drain error: %s", node.Spec.RequestedHostname, err)
+		}
+	default:
```
rmweir, line 817: I don't think this empty default is necessary. If ctx is not done on select this will skip it, removing the effect of the timeout.

pkg/controllers/management/node/utils.go
```diff
@@ -320,3 +321,17 @@ func setEc2ClusterIDTag(data interface{}, clusterID string) {
 		}
 	}
 }
+
+func (m *Lifecycle) getKubeConfig(cluster *v3.Cluster) (*clientcmdapi.Config, error) {
```
rmweir, line 325: nit: there maybe be a function for this already but if this one better fits the use case then it's fine.

rmweir (changes requested):

pkg/controllers/management/node/utils.go
luthermonson: there are but they're elsewhere and need a lot of things inthe controller to make work so i left a local one

pkg/controllers/management/node/controller.go
luthermonson: this is just a niceity, it should be set but incase it's not i quickly set it... after this the node gets deleted so there's no need to persist

pkg/controllers/management/node/controller.go
luthermonson: great point

pkg/controllers/management/node/controller.go
luthermonson: oh and it needs to be on the Node for GetDrainFlags to work, was just easier to put there as other functions in nodesyncer had to change to move that to pkg/node properly.

paynejacob (dismissed):

pkg/controllers/management/node/controller.go
rmweir: and there are no other finalizers and you don't think other will be added? because this could mess up the cache.

luthermonson: if it was nil before it can stay nil, it's not a problem. i just need it set quickly to run the drain.

rmweir (approved):
LGTM

paynejacob (approved):

Issue #29144, opened 2020-09-25
extensions/v1beta1 Ingress in Rancher chart is deprecated


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Steps to reproduce (least amount of steps as possible):**
- install Rancher: `v2.5.0-rc1` in airgap env in HA mode 
- here is the helm command to generate the template
```
 helm template rancher ./rancher-2.5.0-rc1.tgz --output-dir . \
    --namespace cattle-system \
    --set hostname=jiaqi0235-15555-internal.qa.rancher.space \
    --set rancherImage=ec2-18-222-107-174.us-east-2.compute.amazonaws.com/rancher/rancher \
    --set ingress.tls.source=secret \
    --set systemDefaultRegistry=ec2-18-222-107-174.us-east-2.compute.amazonaws.com \
    --set useBundledSystemChart=true
```
- and install rancher 
```
kubectl create namespace cattle-system
kubectl -n cattle-system apply -R -f ./rancher
```

**Result:**
```
>  kubectl -n cattle-system apply -R -f ./rancher
clusterrolebinding.rbac.authorization.k8s.io/rancher created
deployment.apps/rancher created
Warning: extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
ingress.extensions/rancher created
service/rancher created
serviceaccount/rancher created
```

we should update the Rancher helm chart to use `networking.k8s.io/v1` for Ingress

brendarearden:
is available in the most recent master and 2.5 image tags for testing

jiaqiluo:
wait for a v2.5.2-rc to validate this fix in the helm chart.

mrajashree:
Can be tested with v2.5.2-rc1

jiaqiluo:
in `v2.5.2-rc1` HA air gap installation 

the abovementioned warning for `extensions/v1beta1` is gone, but the following new warning appears

> Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

and the recommended version `networking.k8s.io/v1` is the same 

the full output:
```
>  kubectl -n cattle-system apply -R -f ./rancher
clusterrolebinding.rbac.authorization.k8s.io/rancher unchanged
deployment.apps/rancher created
[...]

PR #29336, opened 2020-10-02
[2.5.2] Use newer apiVersion for ingress

brendarearden:
**Problem**
Starting in 1.16 Ingress was no longer in the `extensions/v1beta1` api, so it is not recognized 

**Solution**
Update the apiVersion to a version that is supported in the following kubernetes versions
```
1.19.2 (default)
1.18.9
1.17.12
1.16.15
```
**Issue** 

**Additional Information**
The use of `networking.k8s.io/v1beta1 ` started in **1.16** and will be removed in **v1.22** .

https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
```
The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

Ingress in the extensions/v1beta1 API version will no longer be served
Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.
```

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: {{ template "rancher.fullname" . }}


nickgerace (approved):
Confirmed the apiVersion change to ```networking.k8s.io/v1beta1```. Ship it!

rmweir (commented):
What if a cluster pre 1.16 is being used?

brendarearden:
@rmweir https://v1-15.docs.kubernetes.io/docs/setup/release/notes/#api-changes per this the `networking.k8s.io/v1beta1` api has been available since 1.14. and then based on the https://kubernetes.io/docs/reference/using-api/deprecation-policy/ Do we need to support versions for longer than kubernetes supports them?

hajism:
i have same error 
image
image

chart/templates/ingress.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
```
prachidamle, line 1: This takes care of the rancher's ingress. But there are a lot other ingress usages withing rancher code - how are we transitioning them?

brendarearden: since we aren't wanting to make significant changes to the code base for 2.5.2, we are just addressing the rancher chart with this issue, and then I have opened this issue https://github.com/rancher/rancher/issues/29482 for the rancher code

dramich (approved):

chart/templates/ingress.yaml
dramich: There is already an issue for this (that I can't find)

prachidamle: Ok, lgtm

prachidamle (approved):

rmweir:
@brendarearden no probably not.

rmweir (approved):
LGTM

PR #29377, opened 2020-10-05
Add channels.yaml file

cjellick:
In preparation for having an install script for rancherd, we need this
channels.yaml server so that we can stand up a channel-server that
points at it.


cjellick:
canceled build as this change doesnt get hit by CI at all

Issue #29381, opened 2020-10-05
Panic in rancher when upgrading k8s version of cluster with drain enabled


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- (this was an upgraded rancher HA setup from 2.4.8 to 2.5.0-rc7)
- Upgrade k8s version of the cluster from 1.18.8 to 1.18.9 (custom rke cluster - 5 nodes, 1 etcd, 1 control and 3 worker nodes)
- Panic in rancher server and rancher is seen restarting
- Rancher logs:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2994c5d]

goroutine 10154 [running]:
github.com/rancher/rancher/pkg/controllers/managementuser/nodesyncer.getFlags(0xc001c458c0, 0xc008763b00, 0x4690790, 0x0)
	/go/src/github.com/rancher/rancher/pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go:243 +0x1cd
github.com/rancher/rancher/pkg/controllers/managementuser/nodesyncer.(*nodeDrain).drain.func1(0x44474e0, 0xc003515b00, 0x44c8a6c, 0x7)
	/go/src/github.com/rancher/rancher/pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go:161 +0x13d
github.com/rancher/norman/condition.Cond.doInternal(0x44c8a6c, 0x7, 0x4e25301, 0x4e253e0, 0xc003515b00, 0xc002becf38, 0x6, 0x3906a00, 0xc003499db0, 0x198)
	/go/pkg/mod/github.com/rancher/norman@v0.0.0-20200930000340-693d65aaffe3/condition/condition.go:212 +0x97
github.com/rancher/norman/condition.Cond.do2(0x44c8a6c, 0x7, 0x44c8a01, 0x4e253e0, 0xc003515b00, 0xc002becf38, 0x4065ea, 0xc0047997a0, 0x0, 0xc008702000, ...)
	/go/pkg/mod/github.com/rancher/norman@v0.0.0-20200930000340-693d65aaffe3/condition/condition.go:181 +0x1d8
github.com/rancher/norman/condition.Cond.do(...)
[...]

PR #29388, opened 2020-10-06
[release/v2.5] set default value for ignoreDaemonSets

kinarashah:
https://github.com/rancher/rancher/issues/29381

### DIFF
--- a/pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go
+++ b/pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go
@@ -236,11 +236,17 @@ func (d *nodeDrain) getKubeConfig() (*clientcmdapi.Config, error) {
 }
 
 func getFlags(input *v32.NodeDrainInput) []string {
+	var ignoreDaemonSets bool
+	if input.IgnoreDaemonSets == nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/managementuser/nodesyncer/cordonfieldssyncer.go
```diff
@@ -236,6 +236,10 @@ func (d *nodeDrain) getKubeConfig() (*clientcmdapi.Config, error) {
 }
 
 func getFlags(input *v32.NodeDrainInput) []string {
+	defaultIgnoreDaemonSets := true
+	if input.IgnoreDaemonSets == nil {
+		input.IgnoreDaemonSets = &defaultIgnoreDaemonSets
```
luthermonson, line 241: youre setting this, if someone passes in nil on purpose it'll come back as true. perhaps my attempt at this in another PR is a better idea? https://github.com/rancher/rancher/blob/master/pkg/controllers/managementuser/noderemove/nodes.go#L30-L40

kinarashah: yep makes sense, updated

luthermonson (approved):

dramich (approved):

Issue #29070, opened 2020-09-22
EKS cluster fails to come up in an air gapped proxy setup


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy an air gapped proxy setup
- Deploy EKS cluster
- EKS cluster fails to come up with error

```
time="2020-09-22T01:22:06Z" level=error msg="error syncing 'cattle-global-data/c-b9x2f': handler eks-controller: error creating stack with VPC template: error creating master: RequestError: send request failed\ncaused by: Post https://cloudformation.us-west-1.amazonaws.com/: dial tcp 176.32.118.28:443: i/o timeout, requeuing"
time="2020-09-22T01:22:06Z" level=info msg="Bringing up vpc"
time="2020-09-22T01:22:44Z" level=error msg="error syncing 'cattle-global-data/c-k9454': handler eks-controller: error creating stack with service role template: error creating master: RequestError: send request failed\ncaused by: Post https://cloudformation.us-west-2.amazonaws.com/: dial tcp 54.240.251.124:443: i/o timeout, requeuing"
time="2020-09-22T01:22:44Z" level=info msg="Creating service role"
```

**Expected Result:**
EKS cluster should come up successfully.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head - commit id: `ed2749841db0`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): hosted EKS
- Kubernetes version (use `kubectl version`):

```
1.17
```

paynejacob:
I traced this down to the container running the eks operator it's self.  I am able to resolve dns inside the container but it cannot route to the aws api server.  I am able to route to the aws api server from the node and k3s container.

sowmyav27:
reopening the issue for a fix in 2.5-head

sowmyav27:
**On 2.5-head commit id: `b722847200` and on master-head - commit id: `4916c38e4`**

- Deploy an air gapped proxy setup
- Deploy EKS cluster
- EKS cluster comes up Active. No error seen

PR #29458, opened 2020-10-09
Add proxy support to eks-operator integration

rmweir:
**Problem:**
EKS cluster not provisioning in airgap proxy setup because eks-operator requests to AWS API are timing out.

**Solution:**
Rancher will now pass HTTPS_PROXY, HTTP_PROXY, and NO_PROXY env variables to eks-operator. App will be redeployed when a change is detected in any of the aforementioned variables.

**Issue:**
https://github.com/rancher/rancher/issues/29070

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 	"time"
@@ -44,6 +45,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/token"
+	"sigs.k8s.io/yaml"
 )
 
 const (
@@ -56,6 +58,12 @@ const (
 	enqueueTime         = time.Second * 5
 	importedAnno        = "eks.cattle.io/imported"
 )
+
+type eksOperatorValues struct {
+	HTTPProxy  string `json:"httpProxy,omitempty"`
+	HTTPSProxy string `json:"httpsProxy,omitempty"`
+	NoProxy    string `json:"noProxy,omitempty"`
+}
 
 type eksOperatorController struct {
 	clusterEnqueueAfter  func(name string, duration time.Duration)
@@ -496,6 +504,11 @@ func (e *eksOperatorController) deployEKSOperator() error {
 	systemProjectID := ref.Ref(systemProject)
 	_, systemProjectName := ref.Parse(systemProjectID)
 
+	valuesYaml, err := generateValuesYaml()
+	if err != nil {
+		return err
+	}
+
 	app, err := e.appLister.Get(systemProjectName, eksOperator)
 	if err != nil {
 		if !errors.IsNotFound(err) {
@@ -528,18 +541,21 @@ func (e *eksOperatorController) deployEKSOperator() error {
 			},
 		}
 
+		desiredApp.Spec.ValuesYaml = valuesYaml
+
 		// k3s upgrader doesn't exist yet, so it will need to be created
 		if _, err = e.appClient.Create(desiredApp); err != nil {
 			return err
 		}
 	} else {
-		if app.Spec.ExternalID == latestVersionID {
+		if app.Spec.ExternalID == latestVersionID && app.Spec.ValuesYaml == valuesYaml {
 			// app is up to date, no action needed
 			return nil
 		}
 		logrus.Info("updating EKS operator in local cluster's system project")
 		desiredApp := app.DeepCopy()
 		desiredApp.Spec.ExternalID = latestVersionID
+		desiredApp.Spec.ValuesYaml = valuesYaml
 		// new version of k3s upgrade available, update app
 		if _, err = e.appClient.Update(desiredApp); err != nil {
 			return err
@@ -634,4 +650,23 @@ func notFound(err error) bool {
 		return awsErr.Code() == eks.ErrCodeResourceNotFoundException
 	}
 	return false
+}
+
+// generateValuesYaml generates a YAML string containing any
+// necessary values to override defaults in values.yaml. If
+// no defaults need to be overwritten, an empty string will
+// be returned.
+func generateValuesYaml() (string, error) {
+	values := eksOperatorValues{
+		HTTPProxy:  os.Getenv("HTTP_PROXY"),
+		HTTPSProxy: os.Getenv("HTTPS_PROXY"),
+		NoProxy:    os.Getenv("NO_PROXY"),
+	}
+
+	valuesYaml, err := yaml.Marshal(values)
+	if err != nil {
+		return "", err
+	}
+
+	return string(valuesYaml), nil
 }


pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -635,3 +651,31 @@ func notFound(err error) bool {
...
+		HttpsProxy: httpsProxy,
+		NoProxy:    noProxy,
+	}
+
+	yaml, err := yaml2.Marshal(proxySettings)
+	if err != nil {
+		return "", err
+	}
+
+	return string(yaml), nil
```
nickgerace, line 680: What's the purpose in converting the resulting ```[]byte``` type to ```string```? I see the string used on lines 551 to 558, but am curious if the YAML bytestream could be used instead.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -44,6 +45,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/token"
+	yaml2 "sigs.k8s.io/yaml"
```
nickgerace, line 48: Cosmetic comment/question: the [latest YAML release](https://pkg.go.dev/sigs.k8s.io/yaml) is ```1.2.0```. Is there a reason why we use ```yaml2``` here?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -528,18 +541,21 @@ func (e *eksOperatorController) deployEKSOperator() error {
...
 	} else {
-		if app.Spec.ExternalID == latestVersionID {
+		if app.Spec.ExternalID == latestVersionID && app.Spec.ValuesYaml == valuesYaml {
 			// app is up to date, no action needed
 			return nil
 		}
 		logrus.Info("updating EKS operator in local cluster's system project")
 		desiredApp := app.DeepCopy()
 		desiredApp.Spec.ExternalID = latestVersionID
+		desiredApp.Spec.ValuesYaml = valuesYaml
```
brendarearden, line 558: So based on how the valuesYaml works, this should get overlaid on top of the existing values.yaml. And this would only happen if the user didn't set environment variables for it?

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: It is because the valuesYAML field on app is a string, which is used to feed into the helm CLI.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: I think GoLand did this to avoid colliding with the variable at L675. Will change, good catch.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -635,3 +651,31 @@ func notFound(err error) bool {
...
+		return "", nil
+	}
+
+	proxySettings := proxyValues{
+		HttpProxy:  httpsProxy,
+		HttpsProxy: httpsProxy,
+		NoProxy:    noProxy,
+	}
+
+	valuesYaml, err := yaml.Marshal(proxySettings)
```
aiyengar2, line 675: If the `proxyValues` struct corresponds to the `values` YAML that we're going to pass in, should `proxyValues` be renamed `environmentOverrides` (or something similar) in case we need to submit non-proxy overrides to the `values.yaml` in the future?

rmweir: I imagine we can just change the name once that is the case, shouldn't effect marshaling at all. I am using valuesYaml everywhere else so it could make sense to change it to be valuesYaml if you are in favor of that.

aiyengar2: I agree that we could always refactor, but unless there's ever a reason why we would use `proxyValues` directly outside of `generateValuesYaml`, I'd be in favor of just changing the struct name to `valuesYaml` to be more explicit about what it contains / might be updated to contain in the future.

rmweir: okay, fixed the struct to be "eksOperatorValues" and the local variable to be called "values".

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -635,3 +651,31 @@ func notFound(err error) bool {
...
+// be returned.
+func generateValuesYaml() (string, error) {
+	httpProxy := os.Getenv("HTTP_PROXY")
+	httpsProxy := os.Getenv("HTTPS_PROXY")
+	noProxy := os.Getenv("NO_PROXY")
+
+	if httpProxy == "" && httpsProxy == "" && noProxy == "" {
+		// no proxy related environment variables set
+		return "", nil
+	}
```
aiyengar2, line 667: nit: is there any reason why we need to specifically pass in `""` here after checking the values instead of just letting it pass in:
```
{
 	"httpProxy": "",
 	"httpsProxy": "",
 	"noProxy": ""
 }
```

aiyengar2: If there's no specific reason for this, I think it would be cleaner to just pass in the `os.Getenv` calls directly while creating the struct.

aiyengar2 (dismissed):
lgtm, minor comment

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: I just didn't see the point in marshaling and passing a values yaml to the cli for no reason. I think your suggestion would be clean as well but view these approaches about equal: either we pass a redundant values yaml to cli or we have a few extra lines.

pkg/controllers/management/eks/eks_cluster_handler.go
brendarearden: Oop! I did read the env parsing incorrectly so you can ignore that second half of my statement

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: Went ahead and changed this, since if more env variables are eventually added it may be unreasonable to check if they are all empty anyway.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: Regarding the first part, that is correct.

aiyengar2 (approved):

brendarearden (approved):
LGTM

nickgerace (approved):

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -635,3 +651,22 @@ func notFound(err error) bool {
...
+		HTTPSProxy: os.Getenv("HTTPS_PROXY"),
+		NoProxy:    os.Getenv("NO_PROXY"),
+	}
+
+	valuesYaml, err := yaml.Marshal(values)
+	if err != nil {
+		return "", err
+	}
+
+	return string(valuesYaml), nil
```
cbron, line 671: nit: could just return `yaml.Marshal`, same thing.

cbron (approved):

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: At some point it needs to be converted from []byte to string. Otherwise, that would be the same.

cbron:
Needs to get backported to release/v2.5 branch

Issue #28230, opened 2020-08-04
Ingress Status Load Balancer IPs Not Being Updated


ekristen:
**What kind of request is this (question/bug/enhancement/feature request):** 

Bug

**Steps to reproduce (least amount of steps as possible):**

Unknown?

**Result:**

I have several clusters setup with dedicated edge nodes with nginx ingress deployment, however the `status.loadBalancer` is not being updated to reflect the IP addresses of all the nginx ingress controller nodes.

**Other details that may be helpful:**

1. Amazon EC2 v1.16.10 (cloud provider OFF)
2. Amazon EC2 v1.16.10 (cloud provider on)
3. Amazon EC2 v1.17.6 (cloud provider OFF)

I have multiple Ingress resources defined and the `status.loadBalancer` shows only a single ip address where there are 2+ nodes WITH external IP with the nginx ingress controller deployed.

I found this because I was looking my global DNS and I saw a few records only had a single IP for the DNS record when I knew there was more than one ingress controller and edge node deployed.

---

Question 1 -- who/what is responsible for updating the `status.loadBalancer` section of the ingress in Rancher 2? Is that k8s, something special about Rancher? I can't seem to find any details?

Question 2 -- has anyone seen this before? 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.4
- Installation option (single install/HA): HA

gzrancher/rancher#11807

ekristen:
This looks like it's a bug in the nginx ingress controller. There are lots of reports where the status stops getting updated.

ekristen:
A restart of the ingress controller that has the lock seems to fix it, but it's less than ideal. Supposedly this is fixed after k8s 1.14 and after 0.23 of the controller but I clearly am hitting it.

https://github.com/kubernetes/ingress-nginx/issues/3033
https://github.com/kubernetes/ingress-nginx/pull/4996

ekristen:
Upon further investigation this appears to be a problem with having Project Isolation enabled. Circumstantial evidence would suggesting what when Project Isolation is enabled that only the IP of the ingress controller that has leader election will be used to update the `status.loadBalancer.ingress.{}` ip addresses.  This needs testing to confirm, but after disabling project isolation and restarting the controllers IPs for every node started appearing on ingress status fields.

kinarashah:
[...]

Issue #26677, opened 2020-04-17
[BUG] Rancher breaks ingress-nginx by labelling its Pods ⚠️


al45tair:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Install Rancher.

**Result:**
At some point, Rancher adds the label "field.cattle.io/podName" to the ingress controller Pods. This is unfortunate, because ingress-nginx asks Kubernetes to find all the Pods running ingress-nginx *with the same label as the running pod* in order to find the set of IP addresses that it should use to annotate ingresses (in the loadBalancer part of the Ingress status).

The upshot is that once Rancher has done this, Ingresses are massively less robust than they should be, because they only have the IP address of a single node, rather than all the addresses of all of the nodes running ingress controller instances.

This feeds through into DNS if you're using external-dns (which lots of us are).

This is the cause of #25492; rather than add to that Issue, I filed a new one explaining the actual bug.

**Other details that may be helpful:**
The code in ingress-nginx that does this is visible in internal/ingress/status/status.go, in runningAddresses() and isRunningMultiplePods() (I was looking at revision 6e8c68d8881212 of [kubernetes/ingress-nginx](https://github.com/kubernetes/ingress-nginx)). Adding the label breaks this code.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.4.2
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): Azure, 2 cores, 8GB
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4", GitCommit:"8d8aa39598534325ad77120c120a22b3a990b5ea", GitTreeState:"clean", BuildDate:"2020-03-12T20:55:23Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

[...]

PR #29480, opened 2020-10-12
don't add hostport podname label for system ns pods

kinarashah:
Problem:
We add label field.cattle.io/podName to pods with host ports when network policy is enabled.
This is an issue for nginx ingress controller finds running pods by filtering pods
using label selector.

Solution:
We shouldn't add this label to pods of any system ns, no network policies get implemented
for system namespaces.

https://github.com/rancher/rancher/issues/28230
https://github.com/rancher/rancher/issues/26677


paynejacob (approved):

kinarashah:
https://github.com/rancher/rancher/issues/29714

Issue #28677, opened 2020-09-01
Cluster in Error state when attempting recover from s3 backup scenario 


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

        - This test create 1 custom cluster with s3 backups enabled
            - 1 ControlPlane/worker node
            - 2 worker nodes
            - 3 etcd nodes
        - Creates an Ingress pointing to a workload
        - Snapshots the cluster and checks the backup is in S3
        - Stops the etcd nodes in ec2
        - Waits for the cluster to go into unavailable state.              *****
        - Removes all 3 etcd nodes.                                        *****
        - Creates 3 new etcd nodes and waits until the cluster             *****
          asks to restore from backup.
        - Restores from S3 backup
        - Cluster is validated after it gets in Active state.
        - Checks the Ingress created before the backup is functional after restore.

In the above scenario master-head usually fails between `Waits for the cluster to go into unavailable state` and `Creates 3 new etcd nodes and waits until the cluster asks to restore from backup.` (Lines with *****)

**Result:**

![Screen Shot 2020-09-01 at 9 27 11 AM](https://user-images.githubusercontent.com/27659/91889766-3686f680-ec43-11ea-927f-1156acad5a19.png)

**Other details that may be helpful:**

This scenario works in **v2.4.6** exact same same HA setup and custom downstream cluster.

**Environment information**
- Rancher version: `master-head (09/01/2020)` _c428bc9_
- Installation option: HA k8s

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): AWS t3a.medium
- Kubernetes version (use `kubectl version`): `v1.18.6-rancher1-2

kinarashah:
@izaac 
Could you test this again on master-head? 

I tried this multiple times with both HA and single node, following the steps: I am not able to reproduce the issue. 
rancher: `master-9addfe297070f85eb3020d0e54571adaa438ccec-head`. 

- create custom cluster with 3 etcd nodes, 1 control and 1 worker with s3 enabled backup
- create a workload and ingress
- take a snapshot
- stop 3 etcd nodes
- wait for cluster to be in error state, saying it’s unhealthy
- delete nodes from nodes page
- register 3 new etcd nodes
- cluster complains about stop provisioning, restore from backup
- restored from backup
- cluster stays in the last red error state, but cluster provisioning starts in logs
[...]

PR #29487, opened 2020-10-13
[Automation] - Cluster state fixes for the backup and recover scenario

izaac:
Starting with Rancher `v2.5` there was a change in the cluster states transition behavior described here: #28677

This is a fix to make the test handle this error and ensuring the test keeps working on older Rancher versions.


sowmyav27 (approved):
LGTM

PR #29491, opened 2020-10-13
Fix default value for noProxy helm chart variable

bashofmann:
Starting with Rancher 2.5, the rancher pods fail with an error like

```
2020/10/13 09:34:41 [ERROR] Failed to connect to peer wss://10.42.1.13/v3/connect [local ID=10.42.2.8]: websocket: bad handshake
2020/10/13 09:34:41 [ERROR] Failed to connect to peer wss://10.42.0.11/v3/connect [local ID=10.42.2.8]: websocket: bad handshake
```

if an http proxy is configured and the default noProxy setting from the helm chart is used. This fixes this problem by adding `.svc` and `.cluster.local` to the default noProxy value.

See also for reference:
* https://github.com/rancher/k3s/issues/2176
* https://github.com/rancher/docs/issues/2725


cbron:
Issue: https://github.com/rancher/rancher/issues/29495

rmweir (approved):
LGTM

dramich (approved):

Issue #29403, opened 2020-10-06
Rancher not starting - "Observed a panic: "assignment to entry in nil map" [v2.5.0]


mikekuzak:
Flatcar CoreOS latest stable. 
Rancher standalone docker upgrade from working 2.4.8 to 2.5.0

Changed startup command to include --privileged
Docker container is crashing and rebooting ..

```
2020/10/06 10:08:11 [INFO] Starting /v1, Kind=ResourceQuota controller
2020/10/06 10:08:11 [INFO] Creating roleBinding User u-a4ujwkgdme Role edit
2020/10/06 10:08:11 [INFO] Starting apps/v1, Kind=StatefulSet controller
I1006 10:08:11.680424      25 controller.go:606] quota admission added evaluator for: projectalerts.management.cattle.io
2020/10/06 10:08:11 [INFO] Creating roleBinding User u-a4ujwkgdme Role edit
I1006 10:08:11.737710      25 controller.go:606] quota admission added evaluator for: nodes.management.cattle.io
E1006 10:08:11.891054       8 runtime.go:78] Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)
goroutine 5758 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x3c139e0, 0x4d2e320)
        /go/pkg/mod/github.com/rancher/apimachinery@v0.19.0-rancher1/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /go/pkg/mod/github.com/rancher/apimachinery@v0.19.0-rancher1/pkg/util/runtime/runtime.go:48 +0x82
panic(0x3c139e0, 0x4d2e320)
        /usr/local/go/src/runtime/panic.go:969 +0x166
github.com/rancher/rancher/pkg/controllers/managementuser/rbac.(*crtbLifecycle).reconcileCRTBUserClusterLabels.func2(0xc0020fae78, 0x4513cf)
        /go/src/github.com/rancher/rancher/pkg/controllers/managementuser/rbac/crtb_handler.go:142 +0x100
k8s.io/client-go/util/retry.OnError.func1(0xc0057a7a00, 0xf, 0xc0057a7a00)
        /go/pkg/mod/github.com/rancher/client-go@v1.19.0-rancher.1/util/retry/util.go:51 +0x3c
[...]

PR #29496, opened 2020-10-13
Fix panic while adding new label on CRTB

mrajashree:
https://github.com/rancher/rancher/issues/29403

This PR has two commits, first is to fix the panic and to update the right set of labels
Second is to add the nil check for labels in other rbac controllers too to avoid panic. It covers the four controllers that were modified with this PR https://github.com/rancher/rancher/pull/28423

### DIFF
--- a/pkg/controllers/management/auth/crtb_handler.go
+++ b/pkg/controllers/management/auth/crtb_handler.go
@@ -227,6 +227,9 @@ func (c *crtbLifecycle) reconcileLabels(binding *v3.ClusterRoleTemplateBinding) 
 			if updateErr != nil {
 				return updateErr
 			}
+			if crbToUpdate.Labels == nil {
+				crbToUpdate.Labels = make(map[string]string)
+			}
 			crbToUpdate.Labels[bindingKey] = membershipBindingOwner
 			crbToUpdate.Labels[rtbLabelUpdated] = "true"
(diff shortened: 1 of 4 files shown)


aiyengar2 (dismissed):
LGTM

prachidamle (dismissed):

pkg/controllers/management/auth/crtb_handler.go
```diff
@@ -205,7 +205,7 @@ func (c *crtbLifecycle) reconcileLabels(binding *v3.ClusterRoleTemplateBinding)
 	    2. CRTB.UID is label key for the RB, CRTB.UID=crtb-in-project-binding-owner (in the namespace of each project in the cluster that the user has access to)
 	Using above labels, list the CRB and RB and update them to add a label with ns+name of CRTB
 	*/
-	if binding.Labels[rtbCrbRbLabelsUpdated] == "true" {
+	if binding.Labels != nil && binding.Labels[rtbCrbRbLabelsUpdated] == "true" {
```
dramich, line 208: These nil checks are redundant, you will get the zero value from the map trying to access the nil map - see https://play.golang.org/p/1nGFExa4DwJ

mrajashree: oh thanks!

dramich (approved):

PR #29502, opened 2020-10-13
Add rancher.cattle-system to tls san

StrongMonkey:
This is used for fleet-controller to be able to talk to rancher API(and
downstream cluster) using k8s service name(Service IP). This is a more
efficient way and can bypass proxy if rancher is installed behind a
proxy


StrongMonkey:
Related PR https://github.com/rancher/rancher-operator/pull/11

PR #29515, opened 2020-10-14
Add rke2 airgap deployment methods

rancher-max:
Currently there are two ways to deploy RKE2 in an airgap environment: 
  - Through a private registry that is NOT password protected but does require TLS
  - Tarball method

See rke2 docs for more info on these. This just automates the process for us for easier testing. For tarball method, a bastion server is required to already be present and user must supply the ID. For private registry method, it will create the bastion server itself.

A future PR will enable importing these clusters to Rancher, but this lays the groundwork for that as well by providing options for the necessary additional server arguments as well.

### DIFF
--- a/tests/validation/tests/v3_api/test_airgap.py
+++ b/tests/validation/tests/v3_api/test_airgap.py
@@ -217,23 +217,7 @@ def deploy_bastion_server():
     node_name = AG_HOST_NAME + "-bastion"
     # Create Bastion Server in AWS
     bastion_node = AmazonWebServices().create_node(node_name)
-
-    # Copy SSH Key to Bastion and local dir and give it proper permissions
-    write_key_command = "cat <<EOT >> {}.pem\n{}\nEOT".format(
-        bastion_node.ssh_key_name, bastion_node.ssh_key)
-    bastion_node.execute_command(write_key_command)
-    local_write_key_command = \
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -624,21 +810,35 @@ def run_command_on_airgap_node(bastion_node, ag_node, cmd, log_out=False):
     return result
 
 
-def wait_for_airgap_pods_ready(bastion_node, ag_nodes):
+def wait_for_airgap_pods_ready(bastion_node, ag_nodes,
+                               kubectl='kubectl', kubeconfig=None):
+    if kubeconfig:
+        node_cmd = "{} get nodes --kubeconfig {}".format(kubectl, kubeconfig)
```
ShylajaDevadiga, line 816: Can we skip passing kubectl as param?

ShylajaDevadiga: pls ignore

ShylajaDevadiga:
LGTM

bmdepesa (changes requested):
Can we move the RKE2 airgap tests to their own file?

bmdepesa (approved):

jiaqiluo (approved):
LGTM

Issue #29055, opened 2020-09-21
Slow memory leak in Rancher nodes (related to AD client and leaking goroutines?)


Tejeev:
Slow memory leak in Rancher pods related to Azure client. Dumps show most of the leaking goroutines are coming from http conn.serve which could indicate an open connection.

**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
See: https://github.com/rancher/rancher/issues/27605
This is a continuation of the above investigation.

**Environment information**
Architecture:
Provider is Azure AD
Built at v2.2.9 with RKE v0.2.9
After 5 days upgraded to 2.4.2
After a few more days upgraded to 2.4.3.
We upgraded to 2.4.5 to resolve rancher/rancher#27605 

There is no network isolation on this cluster.
![image](https://user-images.githubusercontent.com/8906089/93818764-b3116180-fc52-11ea-9e8b-3cae6bdeb41e.png)

gzrancher/rancher#12369
gzrancher/rancher#10142

dramich:
@izaac For testing, can you run regression for the AzureAD provider? One way to tell my changes are working is to turn on debug and watch the logs, after the cache is populated you should see the debug messages seen [here](https://github.com/rancher/rancher/pull/29521) drop in time significantly. One way to force the log is login-logout-login with the same user.

izaac:
Rancher version: `master-head (10/19/2020)` 

I validated this is working. The only case where I've noticed a slow down was the first time I logged into Rancher where I assume the cache was built:

`[INFO] [AZURE_PROVIDER] Completed gathering users groups, took 45.200841592s, keys in cache:1001`

Subsequent login requests were really quick:

`[INFO] [AZURE_PROVIDER] Completed gathering users groups, took 848.69µs, keys in cache:1001`

Also covered: 

- Site access P1 cases
- Cluster owner/member P1s
- Project owner/member P1s
- Refresh group membership

I will check `v2.5-head` and `v2.4-head` backports next

izaac:
I performed the same regression [tests done](https://github.com/rancher/rancher/issues/29055#issuecomment-712461266) (see above) in `master-head` In the following Rancher versions:

`v2.5-head (63e99b6) (10/19/2020)`
[INFO] [AZURE_PROVIDER] Completed gathering users groups, took 11.976168432s, keys in cache:1002
[...]

PR #29521, opened 2020-10-14
Add a cache to the azureAD group retrieval

dramich:
Problem:
In large azureAD setups user sync and even on login causes a large
amount of API calls to azureAD to fetch users groups.

Solution:
Cache the group principals as this is static data so it's causing
unnecessary load on rancher and the azureAD API
User syncs are an order of magnitude faster after the cache is populated

Users before and after the cache is populated:
```
2020/10/15 12:03:48 [INFO] [AZURE_PROVIDER] Completed gathering users groups, took 228.658551ms, keys in cache:7
2020/10/15 12:03:51 [INFO] [AZURE_PROVIDER] Completed gathering users groups, took 8.517µs, keys in cache:7
```
```
2020/10/15 12:03:56 [INFO] [AZURE_PROVIDER] Completed gathering users groups, took 1.934538826s, keys in cache:1008
2020/10/15 12:03:59 [INFO] [AZURE_PROVIDER] Completed gathering users groups, took 612.79µs, keys in cache:1008
```

### DIFF
--- a/pkg/auth/providers/azure/azure_provider.go
+++ b/pkg/auth/providers/azure/azure_provider.go
@@ -4,23 +4,25 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/go-autorest/autorest/adal"
+	lru "github.com/hashicorp/golang-lru"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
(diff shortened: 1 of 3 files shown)


mrajashree (dismissed):
LGTM! 👍

mrajashree (dismissed):

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -50,12 +51,16 @@ func Configure(
 	tokenMGR *tokens.Manager,
 ) common.AuthProvider {
 
+	// Setup the groupCache, ignore the error, it will only error on a non-positive number
+	c, _ := lru.New(10000)
```
ibuildthecloud, line 55: Should this be a settings?  Seems like we always hit this with LRU that when you end up with N+1 possibilities we end up trashing the cache.

dramich: Good call. Updated

mrajashree (approved):

dramich:
https://github.com/rancher/rancher/issues/29055

Issue #29362, opened 2020-10-04
ui-offline-preferred as "dynamic" does not work does not work in 2.5-head


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
- `ui-offline-preferred` in v3/settings is `dynamic` by default and the ui-index - `https://releases.rancher.com/ui/latest2/index.html` and dashboard index `"https://releases.rancher.com/dashboard/latest/index.html"` does not pull the latest UI changes. 
- For example: On 2.5-head with `ui-offline-preferred` is dynamic, the fixes for issue -- https://github.com/rancher/dashboard/issues/1602 is NOT available. When `ui-offline-preferred` is false, the fixes are seen. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5-head - commit id: `5d2dd90`
- Installation option (single install/HA): Single node

cbron:
Believe this is a dupe of https://github.com/rancher/rancher/issues/29307 ?

cbron:
> For example: On 2.5-head with ui-offline-preferred is dynamic, the fixes for issue is NOT available. When ui-offline-preferred is false, the fixes are seen.

This is actually correct because the dockerhub tag is v2.5-head, and that happens to match the release pattern of `v2.*`, it would only pull the latest on master-head. That is not what we actually want though, so we should adjust the `IsRelease` function.

nickgerace:
Opened PR for backport to release/v2.5: #29572

sowmyav27:
**On 2.5-head commit id: `b7228` and master-head - commit id: `b7228` with ui-offline-preferred - dynamic**

This issue is seen fixed - https://github.com/rancher/dashboard/issues/1583 . i.e: Default scan selected for an RKE node driver cluster - `rke-profile-permissive`

When ui-offline-preferred - false, Default scan selected for an RKE node driver cluster - `rke-profile-permissive`

PR #29525, opened 2020-10-14
Add check and unit test for release semver 

nickgerace:
Add check and unit test for release semver. The check parses for a "head" substring.
The IsRelease function needs to be able to discern a v2.X-head branch from a release.
Add unit test for the IsRelease function.

Related issue: #29362

### DIFF
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -108,7 +108,7 @@ func PrefixPrivateRegistry(image string) string {
 }
 
 func IsRelease() bool {
-	return releasePattern.MatchString(ServerVersion.Get())
+	return !strings.Contains(ServerVersion.Get(), "head") && releasePattern.MatchString(ServerVersion.Get())
 }
 
 func init() {
--- a/pkg/settings/setting_test.go
+++ b/pkg/settings/setting_test.go
@@ -1,0 +1,31 @@
+package settings
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsRelease(t *testing.T) {
+	inputs := map[string]bool{
+		"dev":         false,
+		"master-head": false,
+		"master":      false,
+		"v2.5.2":      true,
+		"v2":          true,
+		"v2.0":        true,
+		"v2.x":        true,
+		"v2.5-head":   false,
+		"2.5":         false,
+		"2.5-head":    false,
+	}
+	a := assert.New(t)
+	for key, value := range inputs {
+		if err := ServerVersion.Set(key); err != nil {
+			t.Errorf("Encountered error while setting temp version: %v\n", err)
+		}
+		result := IsRelease()
+		a.Equal(value, result, fmt.Sprintf("Expected value [%t] for key [%s]. Got value [%t]", value, key, result))
+	}
+}


nickgerace:
Spoke to Darren about using ```IsValid``` from the official semver library. That may be an option for the future, but this fix is non-intrusive. Added unit test.

nickgerace:
Unit test...

````

**Comment by nickgerace** [comment]
Unit test output...
```bash
[nickgerace at rancherbook in ~/git/nickgerace/rancher-root/rancher/src/github.com/rancher/rancher/pkg/settings] (0) (issue-29362)
% go test -v -run TestIsRelease
=== RUN   TestIsRelease
--- PASS: TestIsRelease (0.00s)
PASS
ok  	github.com/rancher/rancher/pkg/settings	0.263s
```

pkg/settings/setting.go
```diff
@@ -108,7 +108,11 @@ func PrefixPrivateRegistry(image string) string {
 }
 
 func IsRelease() bool {
-	return releasePattern.MatchString(ServerVersion.Get())
+	return IsReleaseHelper(ServerVersion.Get())
+}
+
+func IsReleaseHelper(input string) bool {
+	return !strings.Contains(input, "head") && releasePattern.MatchString(input)
```
aiyengar2, line 115: Can we just avoid the new helper function? I see that you need the helper function to define a unit test for this, but since the logic here is a simple string comparision which might be entirely [replaced with `isValid` based on your comment](https://github.com/rancher/rancher/pull/29525#issuecomment-709437839) I'm not sure if we need the unit test.

Without the helper function, this just becomes a 2 line change.

```suggestion
        server := ServerVersion.Get()
        return !strings.Contains(server, "head") && releasePattern.MatchString(server)
}
```

brendarearden (approved):
looks good!

aiyengar2 (approved):
This PR looks good to me, but imo we should drop the unit test and the helper function unless anyone feels strongly about it (see comment in review)

Approving this PR otherwise since either option is fine!

pkg/settings/setting.go
aiyengar2: cc: @ibuildthecloud if you had a specific reason for recommending it?

nickgerace: Using ```semver.IsValid``` would have been my first choice, but I believe the goal was to avoid unintended collateral damage. Therefore, just checking for the ```head``` substring was enough.

As far as the unit test goes, alternatively, we could use ```IsRelease(optionalInput ...string) bool``` without creating the helper function that the test relies on. However, I wanted to avoid changing the original implementation since ```ServerVesion.Get()``` is hardcoded. There are multiple ways to tackle this, so long as we think a unit test is necessary.

cbron: Agree that we should avoid helper function just for unit test. I would try to mock the setting. If that doesn't work then this is basic enough of a change that we don't need a unit test.

pkg/settings/setting_test.go
```diff
@@ -0,0 +1,26 @@
...
+		"master":      false,
+		"v2.5.2":      true,
+		"v2":          true,
+		"v2.0":        true,
+		"v2.x":        true,
+		"v2.5-head":   false,
+		"2.5":         false,
+		"2.5-head":    false,
+	}
+	for key, value := range inputs {
```
dramich, line 20: Add `ServerVersion.Set(key)` here

pkg/settings/setting.go
dramich: Keep the test, drop the helper and set your server version in the tests. And thank you for adding tests

pkg/settings/setting_test.go
```diff
@@ -0,0 +1,26 @@
...
+		result := IsReleaseHelper(key)
+		if result != value {
```
dramich, line 22: You can use `"github.com/stretchr/testify/assert"` here instead

pkg/settings/setting.go
nickgerace: @dramich @cbron acknowledged, and will move accordingly.

pkg/settings/setting_test.go
nickgerace: Ah good idea.

pkg/settings/setting_test.go
nickgerace: Ahhhhhh this is what I was looking for. @aiyengar2: we found our helper function killer.

pkg/settings/setting.go
nickgerace: Just published the new PR. @aiyengar2 kept your recommendations but did not want to introduce another state, so I called the function twice instead (simple getter method).

pkg/settings/setting_test.go
```diff
@@ -0,0 +1,35 @@
...
+		if err != nil {
+			t.Errorf("Encountered error while setting temp version: %v\n", err)
+		}
+
+		// We use "!assert.Equal" here instead of "assert.NotEqual" because we want the testing harness to know that
+		// the desired outcome is for both objects to be equal.
+		result := IsRelease()
+		if !assert.Equal(t, value, result) {
+			t.Errorf("Expected value [%t] for key [%s]. Got value [%t]", value, key, result)
+		}
```
dramich, line 33: Using both `assert` and `t` is redundant
```suggestion
		assert.Equal(t, value, result, fmt.Sprintf("Expected value [%t] for key [%s]. Got value [%t]", value, key, result))
```

pkg/settings/setting_test.go
```diff
@@ -0,0 +1,35 @@
...
+		"master-head": false,
+		"master":      false,
+		"v2.5.2":      true,
+		"v2":          true,
+		"v2.0":        true,
+		"v2.x":        true,
+		"v2.5-head":   false,
+		"2.5":         false,
+		"2.5-head":    false,
+	}
```
dramich, line 21: Creating a `a := assert.New(t)` here will allow you to stop passing in T down below into the test.

nickgerace: Did not know this. Thanks for the suggestion.

dramich (dismissed):

pkg/settings/setting_test.go
```diff
@@ -0,0 +1,31 @@
...
+	a := assert.New(t)
+	for key, value := range inputs {
+		if err := ServerVersion.Set(key); err != nil {
```
nickgerace, line 25: Simplified this line.

dramich (approved):

nickgerace:
Merging after successful build after talking to Dan.

Issue #29412, opened 2020-10-06
local k3s cluster should not be upgradable in Rancher single install mode 


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug 

**Steps to reproduce (least amount of steps as possible):**
- run rancher:v2.5.0, single install 
- edit the local cluster, and there are new k3s version available in the dropdown list 

 
<img width="1428" alt="Screen Shot 2020-10-06 at 11 56 44 AM" src="https://user-images.githubusercontent.com/6218999/95247710-0f37c200-07cb-11eb-905a-f586ef31ef40.png">

- choose a newer version and upgrade 

**Result:**
The local cluster starts the upgrading process which will never finish 
<img width="1441" alt="Screen Shot 2020-10-06 at 11 35 01 AM" src="https://user-images.githubusercontent.com/6218999/95246018-c1ba5580-07c8-11eb-9ecd-03f764343ba7.png">
 
**Expected Result:**

We do not support upgrading the local k3s cluster in a rancher single-install setup.

aleksey005:
joy 
send one unvalidated byte and the system will crash
scream

jiaqiluo:
This is not an issue in v2.4.x in the docker-install mode 
 
![screencapture-64-225-121-169-c-local-edit-2020-10-08-10_25_55](https://user-images.githubusercontent.com/6218999/95493129-e5a8a300-0950-11eb-8fcb-665cfdca3e9b.png)

jiaqiluo:
The bugfix is validated on rancher:`v2.5-7a10a58973a03121a49f8f4efb1f52292a446b7d-head`, docker install 

Now there is no upgrade option for the local cluster on the edit page 

<img width="1467" alt="Screen Shot 2020-10-16 at 11 35 18 AM" src="https://user-images.githubusercontent.com/6218999/96296731-998add80-0fa4-11eb-91f1-13b2b9925be9.png">

jiaqiluo:
reopen to wait for the fix available in master-head

cbron:
Just merged in master.

jiaqiluo:
The bugfix is validated in rancher:`master-35e2ab8aad5a06f89c4dde7daeb8ca179c27ae77-head` 

No upgrade option for the local cluster in docker-install mode.

alexfok007:
i hit this bug, may i know the solution to stop upgrading the local cluster ?

cbron:
[...]

PR #29551, opened 2020-10-15
Disable kubernetes upgrades on embedded cluster

cbron:
https://github.com/rancher/rancher/issues/29412

This PR stops the k3s portion of provisioner from setting the K3sConfig on an embedded cluster.  UI shown below. Upgrading the cluster is then not possible in the API as k3sConfig is gone. 

Embedded logic is here: https://github.com/rancher/rancher/blob/master/pkg/data/dashboard/cluster_data.go#L40-L42

~Potential caveats: Is it bad for the cluster not to be labeled as k3s ? As far as I can tell that just stops some validating logic around upgrade version, and stops k3s-upgrade-controller, which is what we want. The version on an embedded cluster should never be upgraded.~ It's labeled as K3s now. Only downside now is that we don't have a flag to see if its embedded for other features. 

Screen Shot 2020-10-15 at 3 22 38 PM

### DIFF
--- a/pkg/controllers/management/clusterprovisioner/provisioner.go
+++ b/pkg/controllers/management/clusterprovisioner/provisioner.go
@@ -1020,6 +1020,7 @@ func (p *Provisioner) k3sBasedClusterConfig(cluster *v3.Cluster, nodes []*v3.Nod
 		cluster.Status.Driver == apimgmtv3.ClusterDriverRancherD {
 		return nil //no-op
 	}
+	isEmbedded := cluster.Status.Driver == apimgmtv3.ClusterDriverLocal
 
 	if strings.Contains(cluster.Status.Version.String(), "k3s") {
 		for _, node := range nodes {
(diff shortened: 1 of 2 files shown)


dramich (dismissed):
Could put the if check in the block above

dramich:
If the driver shows `local` maybe the UI can say something different than imported?

mrajashree:
> Is it bad for the cluster not to be labeled as k3s ?

@cbron I'm not entirely sure, but I've seen monitoring v2 has some additional logic for k3s clusters. So if the local cluster is not detected as k3s, we might miss out on that? cc @aiyengar2

dramich:
@cbron I think the safer route is to set the driver but leave the k3s config as nil. If you check the upgrade code it will skip if the k3s config under spec is nil. https://github.com/rancher/rancher/blob/master/pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go#L52

cbron:
I went Dan's route and that seems to work the same way. UI/API still can't upgrade, and that should solve the potential monitoring issues as not its labeled as K3s driver. Had to move around some conditionals to make it work. Only downside is that other features might not be able to tell if its embedded or not, but they could check to see if config is nil, and this whole area will be deprecated in the end anyway. I'll talk to Wes about "imported" phrase in UI but that would be fixed separately.

dramich (approved):

mrajashree (approved):

Issue #29212, opened 2020-09-28
Unable to integrate Keycloak SAML SSO when Client ID field is reserved


joshrwolf:
**What kind of request is this (question/bug/enhancement/feature request):**

enhancement

**Steps to reproduce (least amount of steps as possible):**

Configure Keycloak (SAML) within Rancher, but do not set the `Client ID` filed as the required `https://yourRancherHostURL/v1-saml/keycloak/saml/metadata`.

**Result:**

Rancher fails to test authentication with keycloak, returning "Access was not authorized".  

**Other details that may be helpful:**

Users with reserved format/entries for Keycloak `Client ID`s are unable to set up Keycloak SAML because the saml metadata endpoint is parsed from the client ID.  Adding the metadata endpoint as a separate field, or reconstructing it from the existing `Rancher API Host` field would allow for generic `Client ID`s.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.4.5
- Installation option (single install/HA): HA

mrajashree:
As per Keycloak docs,
```
Client ID: This value must match the issuer value sent with AuthNRequests. Keycloak will pull the issuer from the Authn SAML request and match it to a client by this value.
```
When setting up Keycloak with Rancher, the rancher metadata URL will be the issuer for AuthNRequests. So I don't think it can be changed.

joshrwolf:
I think it would have to be a change with how rancher is constructing the issuer from the `Rancher API Host` (`https://${rancherAPIHost}/v1-saml/keycloak/saml/metadata`).  Instead, providing a user defined field for `issuer` (or the expected sp entity id) would allow for non standard `clientIDs`.  This seems to be a common option amongst other tools supporting SAML.

mrajashree:
Will move this to test once [this](https://github.com/rancher/rancher/issues/29553#issuecomment-710682731) is fixed

izaac:
I validated this configuration and included some P1s checks after enabling Keycloak auth.

- Rancher version: `master-head (10/19/2020)` _e4dc25135_
- Rancher version: `v2.5-head (10/19/2020)` _52583ea7a_


First I enabled Keycloak in Rancher by using the same configuration settings as documented in the docs, and leaving the Entity ID Field blank so that Rancher will generate one automatically. Then checked users auth and site access.


[...]

PR #29555, opened 2020-10-15
Make Keycloak entity ID configurable

mrajashree:
Problem: When configuring SAML auth for Rancher, Rancher (service provider) sets the rancher metadata URL as the authentication request issuer (service provider entity ID). This is how the saml library we use sets it, and is a common way of setting the issuer. SAML identity providers (keycloak, adfs etc) have fields that accept the service provider entity ID, which must match the auth request issuer. For Keycloak this field is called `ClientID`. Recently there has been a change in the saml library, which allows users to configure this entity ID, and set it to some custom value instead of the metadata URL. We use a fork of this library, and the version we are on does not include this change. So Keycloak can only be configured with rancher metadata URL as the clientID and users cannot configure it with some custom reserved clientID

Solution: To include this fix in 2.5.2, update rancher's fork of saml to include this change. This change was added with this commit https://github.com/rancher/saml/commit/0fa95beb269fc67ac0fce223e96ccb14fb4207b2.
Add a configurable entity ID field to the saml config. 

Post 2.5.2 we will stop using rancher's fork and use the upstream saml library, this is tracked in a [separate task](https://github.com/rancher/tasks/issues/112)
https://github.com/rancher/rancher/issues/29212

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 replace (
 	git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
-	github.com/crewjam/saml => github.com/rancher/saml v0.0.0-20180713225824-ce1532152fde
+	github.com/crewjam/saml => github.com/rancher/saml v0.0.3-rancher1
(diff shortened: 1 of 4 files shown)


dramich (approved):

maggieliu:
related to : https://github.com/rancher/rancher/issues/29650

Issue #29346, opened 2020-09-25
Add in support to limit k8s versions for system-charts


maggieliu:
Need to be able to restrict different chart versions based on k8s version. Example where this is needed is for Monitoring v1 chart. If we update the prometheus operator image to v0.39, the API is only supported in k8s 1.16. Given that 2.3 and 2.4 both still support k8s 1.15, this will cause problem.

jiaqiluo:
Failed to validate the enhancement in rancher:`v2.5-7eb1792946639251f51bd109b630f9c1cf8940c2-head` single-install 
Steps:
- clone the rancher's system-chart https://github.com/rancher/system-charts
- create new Monitoring charts version `0.1.5` `0.1.6`
- add the following kubeVersion requirements to the charts
```
chart 0.1.5 - kubeVersion: >= 1.16.0 < 1.18.10
chart 0.1.6 - kubeVersion: >1.18.0
```
https://github.com/jiaqiluo/system-charts/commit/50831a799a2523b54ec7f2b521ad9b8e2d60021a
- in Rancher change the system-library to point to my repo and branch. 
![image](https://user-images.githubusercontent.com/6218999/97369781-cf647780-186a-11eb-9c86-6c825d49d2a7.png)
- Rancher's logs show the chart is updated
```
2020/10/27 22:34:59 [INFO] Updating global catalog system-library
2020/10/27 22:35:00 [INFO] Catalog sync done. 0 templates created, 1 templates updated, 0 templates deleted, 0 templates failed
```
- provision a cluster with k8s `v1.19.3`
![image](https://user-images.githubusercontent.com/6218999/97369868-ff137f80-186a-11eb-82ec-ab113c67eb46.png)

- go to the cluster monitoring page

results:
- It is expected to see the chart version `0.1.6` in the list but it does not show up 
![image](https://user-images.githubusercontent.com/6218999/97369977-3aae4980-186b-11eb-9a2e-8dd04ca67946.png)

rmweir:
Constraints that includes spaces such as `>= 1.16.0 < 1.18.10` must be wrapped in quotes or they will not be properly parsed. Moving back to test.

izaac:
Rancher version: `master-head (10/29/2020)` _d61a1a4_
Rancher version: `v2.5-head (10/29/2020)` _0465f04_

I created new charts for monitoring 0.1.5 and 0.1.6
[...]

PR #29559, opened 2020-10-16
Add kubeVersion checks for catalogv1

rmweir:
**Problem:**
Some charts will only be compatible with certain kubernetes versions

**Solution:**
Validate kubernetes version meets kubeVersion constraint in chart if one is set.

**Issue:**
https://github.com/rancher/rancher/issues/29346

### DIFF
--- a/pkg/api/norman/customization/multiclusterapp/store.go
+++ b/pkg/api/norman/customization/multiclusterapp/store.go
@@ -12,7 +12,7 @@ import (
 	"github.com/rancher/norman/types/values"
 	"github.com/rancher/rancher/pkg/auth/providers"
 	"github.com/rancher/rancher/pkg/auth/requests"
-	catUtil "github.com/rancher/rancher/pkg/catalog/utils"
+	"github.com/rancher/rancher/pkg/catalog/manager"
 	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	"github.com/rancher/rancher/pkg/clusterrouter"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
@@ -65,6 +65,7 @@ func SetMemberStore(ctx context.Context, schema *types.Schema, mgmt *config.Scal
 		templateVersionLister: mgmt.Management.CatalogTemplateVersions("").Controller().Lister(),
 		users:                 mgmt.Management.Users(""),
 		rtLister:              mgmt.Management.RoleTemplates("").Controller().Lister(),
+		catalogManager:        mgmt.CatalogManager,
 	}
 
 	schema.Store = s
@@ -80,10 +81,11 @@ type Store struct {
 	templateVersionLister v3.CatalogTemplateVersionLister
 	users                 v3.UserInterface
 	rtLister              v3.RoleTemplateLister
+	catalogManager        manager.CatalogManager
 }
 
 func (s *Store) Create(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}) (map[string]interface{}, error) {
-	if err := s.validateRancherVersion(data); err != nil {
+	if err := s.validateChartCompatibility(data); err != nil {
 		return nil, err
 	}
 
@@ -101,7 +103,7 @@ func (s *Store) Create(apiContext *types.APIContext, schema *types.Schema, data 
 }
 
 func (s *Store) Update(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}, id string) (map[string]interface{}, error) {
-	if err := s.validateRancherVersion(data); err != nil {
+	if err := s.validateChartCompatibility(data); err != nil {
 		return nil, err
 	}
 
@@ -165,7 +167,7 @@ func (s *Store) ensureRolesNotDisabled(apiContext *types.APIContext, data map[st
 	return nil
 }
 
-func (s *Store) validateRancherVersion(data map[string]interface{}) error {
+func (s *Store) validateChartCompatibility(data map[string]interface{}) error {
 	rawTempVersion, ok := data["templateVersionId"]
 	if !ok {
 		return nil
@@ -183,7 +185,25 @@ func (s *Store) validateRancherVersion(data map[string]interface{}) error {
 		return err
 	}
 
-	return catUtil.ValidateRancherVersion(template)
+	if err := s.catalogManager.ValidateRancherVersion(template); err != nil {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
+	}
+
+	targets, _ := data["targets"].([]interface{})
+	for _, target := range targets {
+		targetMap, ok := target.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		parts := strings.SplitN(targetMap["projectId"].(string), ":", 2)
+		if len(parts) != 2 {
+			continue
+		}
+		if err := s.catalogManager.ValidateKubeVersion(template, parts[0]); err != nil {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
+		}
+	}
+	return nil
 }
 
 func checkDuplicateTargets(data map[string]interface{}) error {
--- a/pkg/api/norman/store/app/store.go
+++ b/pkg/api/norman/store/app/store.go
@@ -8,7 +8,7 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	pv3app "github.com/rancher/rancher/pkg/api/norman/customization/app"
-	catUtil "github.com/rancher/rancher/pkg/catalog/utils"
+	"github.com/rancher/rancher/pkg/catalog/manager"
 	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	hcommon "github.com/rancher/rancher/pkg/controllers/managementuser/helm/common"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
@@ -23,6 +23,8 @@ type Store struct {
 	types.Store
 	Apps                  pv3.AppLister
 	TemplateVersionLister v3.CatalogTemplateVersionLister
+	CatalogManager        manager.CatalogManager
+	ClusterLister         v3.ClusterLister
 }
 
 func (s *Store) Create(apiContext *types.APIContext, schema *types.Schema, data map[string]interface{}) (map[string]interface{}, error) {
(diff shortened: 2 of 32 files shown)


pkg/api/norman/customization/multiclusterapp/store.go
```diff
@@ -183,7 +185,22 @@ func (s *Store) validateRancherVersion(data map[string]interface{}) error {
...
 	}
 
-	return catUtil.ValidateRancherVersion(template)
+	if err := catUtil.ValidateRancherVersion(template); err != nil {
+		return err
+	}
+
+	targets, _ := data["targets"].([]interface{})
+	for _, target := range targets {
+		targetMap, _ := target.(map[string]interface{})
```
mrajashree, line 194:
```suggestion
		targetMap, ok := target.(map[string]interface{})
		if !ok {
		    continue
		}
```

pkg/catalog/utils/version.go
```diff
@@ -63,6 +65,41 @@ func VersionGreaterThan(a, b string) bool {
...
+	k8sVersion, err := semver.Parse(strings.TrimPrefix(cluster.Status.Version.String(), "v"))
+	if err != nil {
+		return err
+	}
+	if !constraint(k8sVersion) {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("incompatible kubernetes version [%s] for template [%s]", k8sVersion.String(), template.Name))
+	}
+	return nil
+}
+
```
cbron, line 102: Trying to wrap my head around the flows going on here, can you summarize which methods are primary ones and how they are being used. Somewhat confusing as there are several with the same name.

rmweir: This method, ValidateKubeVersion, in the utils package is the main one being added. Before, most charts were just being checked for their compatibility with the Rancher server version with a method called ValidateRancherVersion. 

Now both of these methods, ValidateKubeVersion and ValidateRancherVersion are in a new method called ValidateChartCompatibility. Most occurrences of ValidateRancherVersion have been replaced with this function. Any other functions called validateChartCompatibility are helper functions in their own respective packages that prepare their data for the functions and do any additional processing. This is because compatibility is validated in stores, validators, controllers, and action handlers which may all vary a bit in how they extract the information needed.

pkg/api/norman/store/app/store.go
```diff
@@ -57,11 +64,16 @@ func (s *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
 		return nil, err
 	}
 
-	if err := s.validateRancherVersion(data); err != nil {
+	if err := s.validateForMultiClusterApp(id, "update"); err != nil {
 		return nil, err
 	}
 
-	if err := s.validateForMultiClusterApp(id, "update"); err != nil {
+	clusterName, err := s.parseClusterID(apiContext.Request.URL.Path)
```
brendarearden, line 71: nit: var name is clusterName, but method is parseClusterID. not super familiar with this area, so my bad if those are commonly interchangeable.

brendarearden (approved):
LGTM

pkg/catalog/utils/version.go
```diff
@@ -63,6 +65,41 @@ func VersionGreaterThan(a, b string) bool {
...
+func ValidateKubeVersion(template *v3.CatalogTemplateVersion, clusterlister v3.ClusterLister, clusterName string) error {
+	if template.Spec.KubeVersion == "" {
+		return nil
+	}
+	constraint, err := semver.ParseRange(template.Spec.KubeVersion)
+	if err != nil {
+		logrus.Errorf("failed to parse constraint for kubeversion %s: %v", template.Spec.KubeVersion, err)
+		return nil
```
mrajashree, line 85: q: why don't we return the error here?

pkg/catalog/utils/version.go
```diff
@@ -63,6 +65,41 @@ func VersionGreaterThan(a, b string) bool {
...
+	}
+
+	cluster, err := clusterlister.Get("", clusterName)
```
mrajashree, line 88: nit: `clusterLister` instead of `clusterlister`

pkg/catalog/utils/version.go
rmweir: Because an error in this function insinuates that the cluster version is not compatible with the chart, but if the KubeVersion constraint is not readable I think it's best to just treat it as though it's empty and install the app. Installing an app that isn't compatible won't be catastrophic.

mrajashree (dismissed):

pkg/api/norman/store/app/store.go
rmweir: They are but I change it anyway

mrajashree (approved):
lgtm

brendarearden (approved):

pkg/api/norman/store/app/store.go
```diff
@@ -175,3 +187,20 @@ func (s *Store) parseAppExternalID(data map[string]interface{}) (string, string,
...
+	}
+	if _, err := s.ClusterLister.Get("", parts[0]); err != nil {
+		if errors.IsNotFound(err) {
+			return "", httperror.NewAPIError(httperror.NotFound, fmt.Sprintf("cannot find target cluster [%s]: %v", parts[0], err))
+		}
+		return "", httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("error getting target custer [%s]: %v", parts[0], err))
+
+	}
+	return parts[0], nil
+}
```
cbron, line 206: Can we not get project name at all from apiContext and then use `ref.parse` ? There is also a `GetClusterID` method that might work. Adding all these things feels dangerous.

pkg/catalog/utils/version.go
```diff
@@ -95,7 +132,7 @@ func ReleaseServerVersion(serverVersion string) bool {
 	return true
 }
 
-func LatestAvailableTemplateVersion(template *v3.CatalogTemplate) (*v32.TemplateVersionSpec, error) {
+func LatestAvailableTemplateVersion(template *v3.CatalogTemplate, clusterLister v3.ClusterLister, clusterName string) (*v32.TemplateVersionSpec, error) {
```
cbron, line 135: This `LatestAvailableTemplateVersion` subs as a `ValidateChartCompatibility` parent ? We should at least clarify that with a comment above.

cbron (commented):
Some minor things. In general this approach feels a bit jumbled, we are now passing around a lot of objects throughout a lot of method signatures just to eventually hand them off to a util method down the line that is mostly unrelated to the functionality of the feature. But I don't have a better idea right now.

pkg/api/norman/store/app/store.go
rmweir: it's because this is a project scoped api

pkg/catalog/utils/version.go
rmweir: This is technically finds the latest available compatible template version, it did the same prior but only checked Rancher version constraints.

pkg/api/norman/store/app/store.go
dramich: @rmweir I don't understand your answer, why can't `GetClusterID` be used here?

dramich (changes requested):
There are a couple of options we can do with this:
1. Move version utils and let them live on the catalog manager struct, then add that to the scaled/management context, similar to what we do with `UserManager`, then wherever things are being created and registered grab the catalogManager struct. This will save a lot of work in the future if we ever need to extend what this is doing and need access to more controllers/listers. 
2. "Register" the version piece on its own by creating a var at the top which holds the controllers/listers and all the currently exposed functions could then use

I think option 1 is better and follows patterns we already have in place.

pkg/api/norman/store/app/store.go
rmweir: @dramich that function is intended to be used on cluster requests. It does not work because the url in this case is of the form: `/v3/project(s)/<clusterName>:<projectName>/app`, but it is expecting `/v3|k8s/cluster(s)/<clusterName>`. I could modify GetClusterID so it will work.

pkg/types/config/context.go
```diff
@@ -72,6 +75,14 @@ type ScaledContext struct {
...
 }
 
+type CatalogManager interface {
+	ValidateChartCompatibility(template *managementv3.CatalogTemplateVersion, clusterName string) error
+	ValidateKubeVersion(template *managementv3.CatalogTemplateVersion, clusterName string) error
+	ValidateRancherVersion(template *managementv3.CatalogTemplateVersion) error
+	LatestAvailableTemplateVersion(template *managementv3.CatalogTemplate, clusterName string) (*v32.TemplateVersionSpec, error)
+	GetSystemAppCatalogID(templateVersionID, clusterName string) (string, error)
+}
+
```
dramich, line 85: This should be in its own package or under catalog, it doesn't belong in context

pkg/catalog/manager/manager.go
```diff
@@ -2,16 +2,24 @@ package manager
...
+	"sort"
 	"strings"
 	"time"
 
+	"github.com/blang/semver"
+	mVersion "github.com/mcuadros/go-version"
+	"github.com/rancher/norman/httperror"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/catalog/utils"
+
```
dramich, line 14: extra space here

pkg/catalog/manager/manager.go
```diff
@@ -184,3 +194,108 @@ func (m *Manager) deleteBadCatalogTemplates() []error {
...
+	if err != nil {
+		return err
+	}
+
+	k8sVersion, err := semver.Parse(strings.TrimPrefix(cluster.Status.Version.String(), "v"))
+	if err != nil {
+		return err
+	}
+	if !constraint(k8sVersion) {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("incompatible kubernetes version [%s] for template [%s]", k8sVersion.String(), template.Name))
```
dramich, line 228: None of the functions in here should return an APIerror, these helpers are getting called from all over and the caller should decide if they need to wrap as an APIerror

dramich (dismissed):

cbron (approved):

pkg/catalog/manager/manager_test.go
```diff
@@ -0,0 +1,81 @@
...
+					{
+						ExternalID: "catalog://?catalog=library&template=artifactory-ha&version=0.12.14",
+						Version:    "0.12.14",
+					},
+				},
+			},
+		},
+	}
+
+	testLatestAvailableTemplateVersion(t, "v2.1.0", "0.12.15", template)
```
dramich, line 60: Build up a map of your test cases then in your test func loop over the map and process the cases
In the test func you can create one assert at the top with assert.New(t)

dramich: Oh this already existed...

dramich (approved):

Issue #29518, opened 2020-10-14
Panic seen in isAlertRuleMatching after upgrading to v2.5.2


superseb:
From https://github.com/rancher/rancher/issues/29460#issuecomment-708278892:

I believe I have the same issue with Rancher 2.5.2 (standalone Docker container), used to manage a single bare-metal cluster (K8s v1.19.2-rancher1-1).

@rbq If you can provide more info on this, that would help. We can fix the panic based on code but any info would be nice.

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x289aa4a]

goroutine 7030 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/rancher/apimachinery@v0.19.0-rancher1/pkg/util/runtime/runtime.go:55 +0x105
panic(0x3c138c0, 0x73706a0)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/rancher/rancher/pkg/controllers/managementuser/alert/watcher.(*ClusterScanWatcher).isAlertRuleMatching(...)
	/go/src/github.com/rancher/rancher/pkg/controllers/managementuser/alert/watcher/cluster_scan.go:110
github.com/rancher/rancher/pkg/controllers/managementuser/alert/watcher.(*ClusterScanWatcher).Sync(0xc00d96bc70, 0xc01224f830, 0x22, 0xc0222d2300, 0x44c3ef8, 0x4, 0x44d3772, 0xb)
	/go/src/github.com/rancher/rancher/pkg/controllers/managementuser/alert/watcher/cluster_scan.go:79 +0x30a
github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3.(*clusterScanController).AddClusterScopedHandler.func1(0xc01224f830, 0x22, 0x442dfa0, 0xc0222d2300, 0xc02224fb20, 0x3, 0x3, 0x0)
	/go/src/github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3/zz_generated_cluster_scan_controller.go:182 +0x79
[...]

PR #29566, opened 2020-10-16
Check CisStatusScan Fail field before acccesing

nickgerace:
Check if the CisStatusScan Fail field is initialized before accessing
it. If Fail is "nil", a panic will occur.

Issue: #29518

### DIFF
--- a/pkg/controllers/managementuser/alert/watcher/cluster_scan.go
+++ b/pkg/controllers/managementuser/alert/watcher/cluster_scan.go
@@ -53,6 +53,12 @@ func (csw *ClusterScanWatcher) Sync(_ string, cs *v3.ClusterScan) (runtime.Objec
 	if cs.DeletionTimestamp != nil {
 		return cs, nil
 	}
+	// Cis sends alerts when the cluster scan has completed, and when it has completed with failures.
+	// In both scenarios, the "CisScanStatus" needs to be set in order to send an alert, so we skip
+	// sending alerts until then.
+	if cs.Status.CisScanStatus == nil {
+		return cs, nil
+	}
 	// Start with Unknown, True if there is/are a matching alert rule(s), else False
 	if !(v32.ClusterScanConditionAlerted.IsUnknown(cs) &&
 		v32.ClusterScanConditionCompleted.IsTrue(cs)) {
--- a/pkg/controllers/managementuser/cis/clusterScanHandler.go
+++ b/pkg/controllers/managementuser/cis/clusterScanHandler.go
@@ -313,6 +313,7 @@ func (csh *cisScanHandler) Updated(cs *v3.ClusterScan) (runtime.Object, error) {
 				NotApplicable: r.NotApplicable,
 			}
 
+			cs = cs.DeepCopy()
 			cs.Status.CisScanStatus = cisScanStatus
 		}
 		v32.ClusterScanConditionCompleted.True(cs)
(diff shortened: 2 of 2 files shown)


prachidamle (dismissed):
LGTM

nickgerace:
CC: @dramich @prachidamle...

I believe we may either want to...

1. keep the PR "as-is" (or, at most, refactor the existing logic) because the ```ClusterScanRule``` is checked before calling ```isAlertRuleMatching```.
2. move the ```alertRule.Spec.ClusterScanRule == nil``` check inside ```isAlertRuleMatching```.
3. put a new ```cs.Status.CisScanStatus == nil``` check in the outer function (```Sync```).

Here's the check for the ```ClusterScanRule```...
```go
if alertRule.Status.AlertState == "inactive" || alertRule.Spec.ClusterScanRule == nil {
	continue
}
```

The last option would reduce number of calls on the function call stack. We may not even need the ```isAlertRuleMatching``` in that case. Here's the logic for that function...
```go
if alertRule.Spec.ClusterScanRule.ScanRunType != cs.Spec.RunType {
	return false
}
if alertRule.Spec.ClusterScanRule.FailuresOnly && !(cs.Status.CisScanStatus.Fail > 0) {
	return false
}
return true
```

So then the final design could incorporate that logic inside ```Sync```.
```go
for _, alertRule := range clusterAlertRules {
    // insert isAlertRuleMatching logic here
}
```

Reference: https://github.com/rancher/rancher/blob/master/pkg/controllers/managementuser/alert/watcher/cluster_scan.go#L76

rmweir (dismissed):
LGTM

nickgerace:
Need to change this PR in a oder to check if CisScanStatus is not nil before calling the function in question.

nickgerace:
Changed the PR after discussing with @prachidamle and @dramich. Cis sends alerts when the scan has completed and when the scan has failures. In both scenarios, the Status should be set, so I've added a skip at the top of the controller that triggers is Status is nil.

pkg/controllers/managementuser/alert/watcher/cluster_scan.go
```diff
@@ -50,7 +50,7 @@ func (csw *ClusterScanWatcher) Sync(_ string, cs *v3.ClusterScan) (runtime.Objec
 	if cs == nil {
 		return nil, nil
 	}
-	if cs.DeletionTimestamp != nil {
+	if cs.DeletionTimestamp != nil || cs.Status.CisScanStatus == nil {
```
dramich, line 53: Let's add a comment on why we skip the object if `CisScanStatus` is nil

nickgerace: Will do.

nickgerace: Done.

dramich (dismissed):

pkg/controllers/managementuser/alert/watcher/cluster_scan.go
```diff
@@ -50,7 +50,10 @@ func (csw *ClusterScanWatcher) Sync(_ string, cs *v3.ClusterScan) (runtime.Objec
 	if cs == nil {
 		return nil, nil
 	}
-	if cs.DeletionTimestamp != nil {
+	// Cis sends alerts when the cluster scan has completed, and when it has completed with failures.
+	// In both scenarios, the "CisScanStatus" needs to be set in order to send an alert, so we skip
+	// sending alerts until then.
+	if cs.DeletionTimestamp != nil || cs.Status.CisScanStatus == nil {
```
rmweir, line 56: Do we know if the object will be modified once it is finished or should we be reenqueueing the object?

rmweir, line 56: nit: I'd probably just separate these checks since the comment only apply to one of them.

rmweir (approved):
LGTM, one nit

pkg/controllers/managementuser/alert/watcher/cluster_scan.go
nickgerace: Hmmmmmmmm, I concur, Ricky. Will do.

nickgerace: Not the cleanest thing ever, but I think the readability is more important here, and agree with you.

rmweir (approved):
LGTM

nickgerace:
Merging after blue, green, and Dan approval and "go ahead" to merge. Changes have been cosmetic since Dan's approval.

Issue #29139, opened 2020-06-16
Reserved/Used resources are not updated when k3s cluster is local


wsaca:
I installed rancher on kubernetes but the current resources are not displayed on the new UI.
![imagen](https://user-images.githubusercontent.com/3788624/84831636-50744e00-aff1-11ea-83b1-4d6aec3e731a.png)
![imagen](https://user-images.githubusercontent.com/3788624/84831938-e4deb080-aff1-11ea-9608-70ddf999c8b0.png)

Environment:
OS: Ubuntu 20.04 WSL
Rancher / Replicas: 2.4.4 / 1
Nodes: 1

gzrancher/rancher#9733

cjellick:
This has only discussed the local cluster. Do you see the same with a downstream cluster?

cjellick:
This data being displayed is on the rancher cluster CRD.

It gets there by aggregating data from v3.Nodes (rancher crd).

The data on v3.Nodes gets there through the “nodesyncer” handler that syncs it from the v1.Nodes object (k8s resource).

A super quick look at a k3s 1.18 setup I just did shows that the v1.Node object does have the relevant data (status.Allocatable and status.Capacity). But to be fair, that was just a quick k3d cluster.

If QA can get a setup that has the bug going and then look at the k8s Nodes object to see if they’re reporting expected capacity and allocatable data, that woudl help narrow it down.

Coud QA also see if it is happening just for the local cluster or downstream k3s clusters as well?

Also, since this is calling out k3s as the local cluster, I'd like to know what the other cluster types tested were. Thats not in the issue. Was it RKE1 or something else where the data was showing up correctly? if only two types were tried (rke1 and k3s) then it could be on either side. Does the data show properly when EKS is the local cluster?

cjellick:
Actually @vincent99 @codyrancher - I just realized I made some assumptions abou the Vue UI. Doe it in fact read this data from the cluster object at `/v1/management.cattle.io.clusters/local`?

izaac:
Rancher version: `master-head (09/25/2020)` _df9345c_

HA k3s local cluster shows all the resource graphs properly.
RKE downstream cluster shows all the resource graphs properly.

Imported k3s clusters (one openrc and one systemd) resource graphs are not even present in the Dashboard cluster overview.

izaac:
Related: https://github.com/rancher/rancher/issues/28594

cjellick:
[...]

PR #29567, opened 2020-10-16
[ v2.5.2 ] Default to worker if no other role

brendarearden:
**Problem**
Not all nodes are registering in the UI stats due to the lack of role / role label applied. Role labels are not required. 

**Solution**
Check the existing labels on the node, ensure the Spec values reflect the label values, and default to registering it as a worker node if no labels are applied. 

**Issue**

**Additional Information**
Similar logic for determining node labels / status values  is also used [here](https://github.com/rancher/rancher/blob/cd2258c4293be86c9f3b1cb06f8a00f9e195d53a/pkg/controllers/managementuser/nodesyncer/nodessyncer.go#L719)
K3s has expected behavior where these [labels will not be applied to the node](https://github.com/rancher/k3s/issues/379#issuecomment-570933692) from requiring these label values
Kubernetes has [deprecated the use of node-role.kubernetes.io/*](https://github.com/kubernetes/kubernetes/issues/84912)

### DIFF
--- a/pkg/controllers/management/clusterstats/statsaggregator.go
+++ b/pkg/controllers/management/clusterstats/statsaggregator.go
@@ -9,6 +9,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 
+	"github.com/pkg/errors"
 	"github.com/rancher/rancher/pkg/clustermanager"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config"
@@ -63,6 +64,10 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 	var machines []*v3.Node
 	// only include worker nodes
 	for _, m := range allMachines {
+		// if none are set, then nodes syncer has not completed
+		if !m.Spec.Worker && !m.Spec.ControlPlane && !m.Spec.Etcd {
+			return errors.Errorf("node role cannot be determined because node %s has not finished syncing. retrying...", m.Status.NodeName)
+		}
 		if m.Spec.Worker && !m.Spec.InternalNodeSpec.Unschedulable {
 			machines = append(machines, m)
 		}
--- a/pkg/controllers/managementuser/nodesyncer/nodessyncer.go
+++ b/pkg/controllers/managementuser/nodesyncer/nodessyncer.go
@@ -715,24 +715,9 @@ func (m *nodesSyncer) convertNodeToMachine(node *corev1.Node, existing *v3.Node,
 	for name, quantity := range limits {
 		machine.Status.Limits[name] = quantity
 	}
-
-	if node.Labels != nil {
-		_, etcd := node.Labels["node-role.kubernetes.io/etcd"]
-		machine.Spec.Etcd = etcd
-		_, control := node.Labels["node-role.kubernetes.io/controlplane"]
-		_, master := node.Labels["node-role.kubernetes.io/master"]
-		if control || master {
-			machine.Spec.ControlPlane = true
-		}
-		_, worker := node.Labels["node-role.kubernetes.io/worker"]
-		machine.Spec.Worker = worker
-		if !machine.Spec.Worker && !machine.Spec.ControlPlane && !machine.Spec.Etcd {
-			machine.Spec.Worker = true
-		}
-	}
-
-	machine.Status.NodeAnnotations = node.Annotations
 	machine.Status.NodeLabels = node.Labels
+	determineNodeRoles(machine)
+	machine.Status.NodeAnnotations = node.Annotations
 	machine.Status.NodeName = node.Name
 	machine.APIVersion = "management.cattle.io/v3"
 	machine.Kind = "Node"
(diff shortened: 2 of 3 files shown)


pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -220,3 +221,20 @@ func (s *StatsAggregator) machineChanged(key string, machine *v3.Node) (runtime.
 	}
 	return nil, nil
 }
+
+func determineNodeRole(machine *v3.Node) {
```
rmweir, line 225: I'm thinking maybe this logic should go where the role is usually set, nodessyncer. I wouldn't expect statsaggregator to set the role. Another possible option would be to not set it but infer that it is a worker here and have the determineNodeRole function return a bool.

rmweir (changes requested):

pkg/controllers/management/clusterstats/statsaggregator.go
cbron: Agree. Also in that same line of thinking I'm not sure we even want to update the spec on a node. Using status or a util method to check both seems cleanest.

rmweir (approved):
LGTM

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -220,3 +220,13 @@ func (s *StatsAggregator) machineChanged(key string, machine *v3.Node) (runtime.
 	}
 	return nil, nil
 }
+
+func isWorkerNode(machine *v3.Node) bool {
+	if machine.Spec.Worker {
+		return true
+	}
+	if !machine.Spec.Worker && !machine.Spec.ControlPlane && !machine.Spec.Etcd {
```
cbron, line 228: First part seems redundant but thats ok I guess. So we aren't checking labels/annotations at all anymore like original PR ? Is it possible box has a label/annotation of control plane ?

cbron: I see those linked PR's now. I guess this works. Might be nice to explain with comment here why we assume its worker.

brendarearden: it makes sense not to re-check the labels/annotations and rely on the Spec as a source of truth - unless there is a reason that I am missing the nodesync stuff wouldn't have occurred for a machine

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
```diff
@@ -851,3 +836,18 @@ func (m *nodesSyncer) isClusterRestoring() (bool, error) {
...
+		_, master := machine.Status.NodeLabels["node-role.kubernetes.io/master"]
+		_, worker := machine.Status.NodeLabels["node-role.kubernetes.io/worker"]
+		machine.Spec.Etcd = etcd
+		machine.Spec.Worker = worker
+		machine.Spec.ControlPlane = control || master
+		if !machine.Spec.Worker && !machine.Spec.ControlPlane && !machine.Spec.Etcd {
+			machine.Spec.Worker = true
+		}
+	}
+}
```
cbron, line 853: Is this method split out just so its more testable ? I don't see any real changes

cbron (commented):
import error is failing CI

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
brendarearden: yeah, same steps - just added tests

pkg/controllers/management/clusterstats/statsaggregator.go
brendarearden: so it seems like its just that race condition between the states getting updated prior to nodesync happening, so instead of updating the spec or lables / annotations, just going to be erroring if one if the specs is not set since that indicates the nodesync hasn't happened

pkg/controllers/management/clusterstats/statsaggregator.go
brendarearden: this method was removed

prachidamle (dismissed):

cbron (dismissed):
lgtm

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -63,6 +65,10 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 	var machines []*v3.Node
 	// only include worker nodes
 	for _, m := range allMachines {
+		// if none are set, then nodes syncer has not completed
+		if !m.Spec.Worker && !m.Spec.ControlPlane && !m.Spec.Etcd {
+			return errors.Errorf("Not able to determine node role for %s.", m.Status.NodeName)
```
rmweir, line 70: Include something about it not being ready yet since it is probably not in a stuck state.

rmweir: Also, I thought the linter didn't like errors that start with a capital but maybe not.

rmweir (changes requested):

pkg/controllers/management/clusterstats/statsaggregator.go
brendarearden: removed capital and expanded error message to include indication of retry

rmweir (approved):
LGTM

cbron (approved):

Issue #29379, opened 2020-10-05
restrictedAdmin - cannot assign global role to users


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- run rancher:v2.5.0-rc8/ master-head, docker install
- add user1 as a restricted admin
- log in as user1, try to create a new local user with the role `Standard User`

**Result:**
Failed to assign the role Standard User to the user

![screencapture-64-225-32-226-g-security-accounts-add-2020-10-05-11_17_13](https://user-images.githubusercontent.com/6218999/95116965-a97d0480-06fc-11eb-83b9-e329f9c49383.png)

But the user account is created
![screencapture-64-225-32-226-g-security-accounts-detail-u-d627p-2020-10-05-11_17_53](https://user-images.githubusercontent.com/6218999/95116975-ad108b80-06fc-11eb-90bb-587804214467.png)

deniseschannon:
@jiaqiluo does this apply to all global roles or just standard user?

dramich:
Just standard, their perm is incorrect and needs to be fixed.

jiaqiluo:
the bugfix is validated in `v2.5-52583ea7a00f64b0ea15190566ae2cef9540d19c-head` 

the steps are the same as the issue provides. 

Result: the new local user with the role `Standard User` is created successfully, and I can log in as the new user.

jiaqiluo:
the bugfix is validated in  `master-e4dc25135dbe9bad376dc7bd62360b87273cdbde-head`

the steps are the same as the issue provides.

Result: the new local user with the role Standard User is created successfully, and I can log in as the new user.

PR #29575, opened 2020-10-16
Fix secrets creation rule for standard user

nickgerace:
Add secrets creation rule for restrictedAdmin. This applies to all
apiGroups.

Issue: #29379

### DIFF
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -94,7 +94,7 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext) 
 		addRule().apiGroups("management.cattle.io").resources("kontainerdrivers").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("podsecuritypolicytemplates").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("nodetemplates").verbs("create").
-		addRule().apiGroups("*").resources("secrets").verbs("create").
+		addRule().apiGroups("").resources("secrets").verbs("create").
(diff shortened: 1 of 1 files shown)


nickgerace:
Admittedly, I am curious if we can restricted the wildcard down to something smaller. However, the admissions webhook seems to explicitly require...
```
apiGroup: "*"
resources: "secrets"
verbs: "create"
```

rmweir (dismissed):
Not sure why empty wouldn't work but it seems this is how we add this rule everywhere else. LGTM.

nickgerace:
@rmweir I was confused as well, but I followed the repro steps from the issue, and it worked.

nickgerace:
Had it backwards. The ```user```'s permissions need to be changed instead.

pkg/data/management/role_data.go
```diff
@@ -94,7 +94,7 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext)
 		addRule().apiGroups("management.cattle.io").resources("kontainerdrivers").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("podsecuritypolicytemplates").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("nodetemplates").verbs("create").
-		addRule().apiGroups("*").resources("secrets").verbs("create").
+		addRule().apiGroups("").resources("secrets").verbs("create").
```
mrajashree, line 97: This is changing the existing User role's permissions. If we want to allow restricted admin to create standard user, then [this](https://github.com/rancher/rancher/pull/29575/files#diff-1233e0b4cb129841936b478407cce09e86417b3125fd9e68731243adec2d6003R84) permission should be changed to 
```
addRule().apiGroups("*").resources("secrets").verbs("create").
```

mrajashree (approved):

dramich (approved):

Issue #29127, opened 2020-09-24
[EKSv2] If eks-operator deploy fails, it is not retried


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Set system-charts catalog branch to a branch that does not contain the eks-operator chart.
If eks-operator app is already running be sure to delete it.
Create eks cluster
Observe error in the UI
Change branch to dev-v2.4

**Result:**
App deploy is never retried.

sowmyav27:
**Reproduced this issue on 2.5.1**

- Set system-charts catalog branch to a branch that does not contain the eks-operator chart - `dev` branch.
- Create eks cluster
- Error seen on cluster - `Failed to deploy eks-operator: catalogtemplates.management.cattle.io "system-library-rancher-eks-operator" not found` and cluster is seen in `Provisioning` state.
- Change branch to dev-v2.4
- The eks operator app does NOT get deployed.

**On 2.5-head commit id: `52583ea7a0` and master-head - commit id: `e4dc2513`**

- Set system-charts catalog branch to a branch that does not contain the eks-operator chart - `dev` branch.
- Create eks cluster
- Error seen on cluster - `Failed to deploy eks-operator: catalogtemplates.management.cattle.io "system-library-rancher-eks-operator" not found` and cluster is seen in `Provisioning` state.
- Change branch to dev-v2.4
- eks operator workload is deployed. 
- Error message on the cluster is NOT seen anymore.
- Cluster starts provisioning. 
- Clusters come up active.

RogerSik:
Same issue here. But i'm worried that i will get into troubles when switching vom branch 2.6 to branch 2.5

PR #29576, opened 2020-10-16
Evaluate proper error

rmweir:
**Problem:**
Prior, the error from deploying the eks-operator was evaluated but the condition update error was returned. The condition
update error is usually nil. This caused the controller to not reenqueue the cluster. Since the cluster was not reenqueued the
app was not redeployed unless a separate event caused the cluster or another cluster to reenqueue. Now, the condition error is evaluated and the original error is returned if the condition error is nil.

**Solution:**
Now, the condition error is evaluated and the original error is returned if the condition error is nil.

**Issue:**
https://github.com/rancher/rancher/issues/29127


thedadams (approved):

prachidamle (approved):

cbron (approved):
Needs backport

PR #29578, opened 2020-10-16
Update rke, cli, ui tags for 2.5.2-rc1

mrajashree:


pkg/apis/go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rancher/eks-operator v1.0.1
 	github.com/rancher/norman v0.0.0-20200930000340-693d65aaffe3
-	github.com/rancher/rke v1.2.0
+	github.com/rancher/rke v1.2.2-rc1
```
cbron, line 14: No go.sum changes ?

cbron (commented):
Once this merges we can bump it in master too. Thats not part of release checklist but good to keep them somewhat inline.

pkg/apis/go.mod
mrajashree: there are go.sum changes too=> https://github.com/rancher/rancher/pull/29578/files#diff-3295df7234525439d778f1b282d146a4f1ff6b415248aaac074e8042d9f42d63R1029

cbron: I mean in `pkg/apis/`

mrajashree: ah, thanks!

cbron (approved):

Issue #29557, opened 2020-10-16
Installing Loki chart from URL, "Error: gzip: invalid header"


zhcsyncer:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

1. Add loki repository https://grafana.github.io/loki/charts
2. Install Loki 

**Result:**
![image](https://user-images.githubusercontent.com/13214840/96203930-a5bb6e80-0f95-11eb-9825-2d6132e5308c.png)

**Other details that may be helpful:**

```log
rancher-7c7b6bd945-45zp2 rancher 2020/10/16 01:53:40 [ERROR] Unknown error: gzip: invalid header
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.1
- Installation option (single install/HA):

```bash
~ (⎈ infra-cluster|cattle-system) ❯❯❯ helm get values rancher -n cattle-system
USER-SUPPLIED VALUES:
hostname: r.xxx.com
ingress:
  extraAnnotations:
    kubernetes.io/ingress.class: edge-nginx
    kubernetes.io/ingress.provider: nginx
    nginx.ingress.kubernetes.io/proxy-connect-timeout: "30"
    nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
    nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
  tls:
    source: secret
replicas: 1
```

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-13T18:08:14Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.9-aliyun.1", GitCommit:"4f7ea78", GitTreeState:"", BuildDate:"2020-05-08T07:29:59Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

Fgruntjes:
[...]

PR #29590, opened 2020-10-17
[2.5] Add trailing slash to helm repo url

thedadams:
If the helm repo url does not end with a '/' and the chart url is relative, then the end of the helm repo path was being removed.

For example, if the helm repo was `https://helm.example.com/repo/charts` and the chart url was `chart-v1.0.0.tgz`, then the url was being resolved to `https://helm.example.com/repo/chart-v1.0.0.tgz`. This was incorrect.

Ensuring that the repo url ends with a '/' fixes this problem.

Issue:
https://github.com/rancher/rancher/issues/29557

To elaborate on how this fixes the issue: when the path was being altered in the undesired way as above, the response was HTML with a 404 status. When the `gzip` reader tried to decompress the expected response, it produced and `invalid header error`because it was reading HTML instead of gzip.

### DIFF
--- a/pkg/controllers/dashboard/helm/repo.go
+++ b/pkg/controllers/dashboard/helm/repo.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"strings"
 	"time"
 
 	catalog "github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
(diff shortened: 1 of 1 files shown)


rmweir (approved):
LGTM

mrajashree (approved):

cbron (approved):

PR #29601, opened 2020-10-17
Adding P1 test cases for CIS v2 charts

sowmyav27:
Added the following tests:
- Run Permissive scan on a cluster
- Run Default scan on a cluster
- Create a scan profile and run scan
- Create a benchmark version and run scan

### DIFF
--- a/tests/validation/tests/v3_api/test_v2_cis_scan.py
+++ b/tests/validation/tests/v3_api/test_v2_cis_scan.py
@@ -1,15 +1,19 @@
 import pytest
 import os
+import time
+import json
 from .common import USER_TOKEN
 from .common import get_cluster_client_for_token_v1
 from .common import execute_kubectl_cmd
 from .common import get_user_client_and_cluster
 from .common import wait_until_app_v2_deployed
 from .common import check_v2_app_and_uninstall
+from .common import check_v2_app_installed
+from .common import random_test_name
 
 CIS_CHART_VERSION = os.environ.get('RANCHER_CIS_CHART_VERSION', "1.0.100")
 SCAN_PROFILE = os.environ.get('RANCHER_SCAN_PROFILE', "rke-profile-permissive")
-
+DEFAULT_SCAN_RUN_TIMEOUT = 120
 cluster_detail = {"cluster": None}
 cis_annotations = \
     {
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_v2_cis_scan.py
```diff
@@ -68,10 +78,203 @@ def test_install_v2_cis_benchmark():
...
+        )
+    rancherrepo = \
+        client.list_catalog_cattle_io_clusterrepo(id="rancher-charts")
+    cluster_id = cluster_detail["cluster"]["id"]
+    cluster_name = cluster_detail["cluster"]["name"]
+    rancher_repo = rancherrepo["data"][0]
+
+    # check if CIS is already installed and uninstall the app
+    if not check_v2_app_installed(client,CHART_NAME):
+        install_cis_v2(client, cluster_id, cluster_name, rancher_repo)
```
jiaqiluo, line 232: The comment message is misleading: do you uninstall the app here? 
(this comment appears at multiple places)

jiaqiluo (changes requested):

tests/validation/tests/v3_api/test_v2_cis_scan.py
sowmyav27: Sorry about that. Will change it.

tests/validation/tests/v3_api/test_v2_cis_scan.py
```diff
@@ -68,10 +78,203 @@ def test_install_v2_cis_benchmark():
...
+    rancher_repo = rancherrepo["data"][0]
+
+    # install CIS if not installed
+    if not check_v2_app_installed(client,CHART_NAME):
+        install_cis_v2(client, cluster_id, cluster_name, rancher_repo)
+
+    client = \
+        get_cluster_client_for_token_v1(
+            cluster_detail["cluster"]["id"], USER_TOKEN
+        )
```
jiaqiluo, line 103: Why do you get the client again which is the same as line 86 - 89?

tests/validation/tests/v3_api/test_v2_cis_scan.py
```diff
@@ -68,10 +78,203 @@ def test_install_v2_cis_benchmark():
...
+    # create cluster scans
+    scan_profile = ""
+    scan = create_clusterscan(client, scan_profile)
+
+    if cluster_detail["cluster"]["provider"] == "rke":
+        scan_profile = "rke-profile-permissive"
+    benchmark_version = cis_scan_profile_benchmark[scan_profile]
```
jiaqiluo, line 151: What will happen if the provider is not rke? Will this test report a failure?

jiaqiluo (changes requested):

tests/validation/tests/v3_api/test_v2_cis_scan.py
sowmyav27: The first client does not have context to "Apps" which gets installed right before this client statement, because of which my cases were failing. Reload of the client object somehow didnt seem to work. (I am not sure if that is valid)

tests/validation/tests/v3_api/test_v2_cis_scan.py
sowmyav27: Good question. I will write some code up for when its not rke.

tests/validation/tests/v3_api/test_v2_cis_scan.py
jiaqiluo: @sowmyav27  It makes sense. Could you add a comment in the code, others may have the same confusion in the future.

jiaqiluo (changes requested):
@sowmyav27 
The current tests are hard-coded for the RKE cluster, right? 

I notice that there are GKE and EKS in the list of `cis_scan_report_results`, how will we run the test on GKE and EKS? 

We can consider adding logic to determine the type of the cluster and use the corresponding expected result from `cis_scan_report_results`.  

We already have the method to get the cluster type.
https://github.com/rancher/rancher/blob/ad28e26612968098ea1420bc81333ae14df0f4bd/tests/validation/tests/v3_api/common.py#L1186-L1201

sowmyav27:
@jiaqiluo I have added code to support other types of clusters also.

jiaqiluo:
unassigned myself. please add me back if needed.

PR #29609, opened 2020-10-19
[2.4] Add a cache to the azureAD group retrieval

dramich:
Problem:
In large azureAD setups user sync and even on login causes a large
amount of API calls to azureAD to fetch users groups.

Solution:
Cache the group principals as this is static data so it's causing
unnecessary load on rancher and the azureAD API
User syncs are an order of magnitude faster after the cache is populated

backport of: https://github.com/rancher/rancher/pull/29521


mrajashree (approved):

dramich:
https://github.com/rancher/rancher/issues/29055

PR #29644, opened 2020-10-20
[ v2.5.2 ] Use capabilities to check if has apiVersion

brendarearden:
**Problem**
Deprecation warning is still visible

**Solution**
Determine if Ingress is available at the networking/v1 apiVersion and default to v1beta1 if it is not

**Issue**
#29144

**Additional Info**
Checking apiVersion has resource is available starting at helm 2.15


rmweir (approved):
LGTM

dramich (approved):

Issue #29327, opened 2020-10-02
Fleet agent doesn't deploy in downstream cluster when Rancher is behind proxy


bmdepesa:
rancher/rancher:v2.5.0-rc4

* Deploy Rancher behind a proxy 
* Custom cluster in same network as Rancher,  with docker configured to use proxy - but no proxy vars passed to agents
* Custom cluster becomes active
* Navigate to Fleet in Dashboard for the downstream cluster
* Click `Clusters`
* Fleet agent is not ready for the downstream custom cluster with error: 
```
Get "https://<host>/k8s/clusters/c-n6pv4/version?timeout=32s": dial tcp <internal_host>:443: i/o timeout
```

maggieliu:
This will work if in the setup server-url is accessible to fleet.

maggieliu:
Will only be fixed for HA setup.

izaac:
Rancher version: `v2.5-head (10/22/2020)` _1599dc1_
Rancher version: `master-head (10/22/2020)` _9124f6e_

HA k8s local cluster behind Squid Proxy

I see the `fleet-agent` active in the downstream cluster (Custom RKE) in both `v2.5-head` and `master-head`

StrongMonkey:
@izaac Can you retest this? the fleet is just merged in rancher with version v0.3.1-rc2 and v0.1.1 rancher-operator. We should verify that there is no regression in both normal and airgap install and clusters are abled to registered in fleet for both cases.

izaac:
@StrongMonkey I will take a look at this today and validate with the latest fleet

izaac:
Rancher version: `v2.5-head (10/26/2020)`  _0b71b59_
Rancher version: `master-head (10/26/2020)` _a3a4c87_

HA k8s local cluster behind Squid Proxy

I see the `fleet-agent` active in the downstream cluster (Custom RKE) in both `v2.5-head` and `master-head`
P1 use case executed and worked in both branches too.

PR #29655, opened 2020-10-20
Listen tls on a different port with unique certs

StrongMonkey:
Open a different port in rancher for tls so that fleet can talk to this
internal tls port with ca


StrongMonkey:
https://github.com/rancher/rancher/issues/29327

PR #29659, opened 2020-10-20
Publish rancherd-upgrade images 

MonzElmasry:
This pr adds a dispatch pipeline for rancher drone to create a new rancherd-upgrade image when a new release tag is created


briandowns (dismissed):

.drone.yml
```diff
@@ -896,4 +896,43 @@ volumes:
...
+type: docker
+name: dispatch
+
+platform:
+  os: linux
+  arch: amd64
+
+steps:
+  - name: dispatch
+    image: rancher/dapper:v0.5.0
```
dramich, line 910: This is a pretty old version, the rest of the file appears to use 0.5.2, is there a specific reason to use this version?

MonzElmasry: no, will update

.drone.yml
```diff
@@ -896,4 +896,43 @@ volumes:
...
+type: docker
+name: dispatch
+
+platform:
+  os: linux
+  arch: amd64
+
+steps:
+  - name: dispatch
+    image: rancher/dapper:v0.5.2
```
luthermonson, line 910: this is just a curl in a shell script, do we really need dapper? just use alpine or something small?

luthermonson: if you did that you wouldnt need the volume mount or anything, just adds an extra container overhead thats unnecessary

scripts/dispatch
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -ex
+
+REPO="https://api.github.com/repos/rancher/rancherd-upgrade/dispatches"
+
+# send dispatch event to REPO
+curl -XPOST -u "${PAT_USERNAME}:${PAT_TOKEN}" \
```
luthermonson, line 7: you sure we need a username? most github api stuff only requires a token, i worry we dont have the this pat_username in secrets and it is pointless to add. might want to double check you can get around this with just the token

scripts/dispatch
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
luthermonson, line 1: if you switch to alpine or a smaller image you might want to do sh instead of bash to be extra safe

luthermonson:
side note.. you're using github workflows and drone in rancherd-upgrade? can we just stick with drone and stay consistent?

Additionally, calling the dispatch seems flimsy, is there a reason why this has to be in another repo and not just all embedded into rancher/rancher if we're trying to keep version parity? we're doing so with agent today with multiple builds, perhaps we can just house it in rancher/rancher instead of relying on a dispatch + additional drone builds?

scripts/dispatch
dweomer: We use this pattern for k3s-upgrade and rke2-upgrade. So long as the secrets are setup correctly in drone we should be fine. I believe that @MonzElmasry has leveraged a past PR for another project as a template for this very thing.

.drone.yml
dweomer: Dapper is already in scope for the build in drone. We can slim it down at a later date with some trial and error but if this has worked elsewhere we should be able to leverage it here.

scripts/dispatch
dweomer: These scripts are copy-pasta from another project, rke2-upgrade I think. Bash is already assumed and know to work.

cmd/rancherd/package/.drone.yml
```diff
@@ -0,0 +1,78 @@
+---
```
luthermonson, line 1: probably dont need this?

.drone.yml
```diff
@@ -231,6 +231,24 @@ steps:
     - drone-publish.rancher.io
     status:
     - failure
+- name: docker-publish-rancherd
```
luthermonson, line 234: probably should do this before slack notify

cmd/rancherd/package/.drone.yml
```diff
@@ -0,0 +1,78 @@
...
+kind: pipeline
+name: manifest
+
+platform:
+  os: linux
+  arch: amd64
+
+steps:
+  - name: manifest
```
luthermonson, line 53: do you still need the manifest step? youre not doing it in the maine .drone.yml

.drone.yml
```diff
@@ -231,6 +231,24 @@ steps:
...
+  image: plugins/docker
+  settings:
+    build_args:
+      - ARCH=amd64
+      - "TAG=${DRONE_TAG}"
+    context: cmd/rancherd/package
+    custom_dns: 1.1.1.1
+    dockerfile: cmd/rancherd/package/Dockerfile
```
luthermonson, line 242: if you have this in a drone pipeline do you need the extra docker build in scripts/package?

MonzElmasry: removed

cmd/rancherd/package/.drone.yml
MonzElmasry: That was for if we support multiple architectures in rancherd-upgrade, since we dont, removed

.drone.yml
MonzElmasry: moved

cmd/rancherd/package/.drone.yml
MonzElmasry: removed

MonzElmasry:
Closing this pr as it's not needed now, a different path of upgrades will be supported

Issue #29652, opened 2020-10-20
Error Provisioning Cluster - Cluster health check failed: cluster agent is not ready


izaac:
**What kind of request is this:**
bug

**Steps to reproduce (least amount of steps as possible):**

- Install v2.5.2-rc2, v2.5-head or master-head
- Try to provision a custom or RKE cluster

**Result:**

Agents are in active state but the cluster shows this error

`Cluster health check failed: cluster agent is not ready`

![Screen Shot 2020-10-20 at 9 45 40 AM](https://user-images.githubusercontent.com/27659/96640082-82ffc180-12d7-11eb-86bb-b44fa16eaa2b.png)

**Other details that may be helpful:**

**Environment information**
- Rancher version: `v2.5.2-rc2
- Installation option: HA RKE k8s `byo-valid` certificates option

izaac:
At first I thought this was #29620 but here the nodes are Active but the cluster is in Error state.

bmdepesa:
rancher/rancher:v2.5-head

Kubelet logs on one of my clusters' controlplane nodes are showing:
```
E1020 20:13:23.157228   12224 pod_workers.go:191] Error syncing pod ee7e29e6-ebee-4c9a-8448-8f6051466c0a ("cattle-cluster-agent-67dc69955d-nktfm_cattle-system(ee7e29e6-ebee-4c9a-8448-8f6051466c0a)"), skipping: failed to "StartContainer" for "cluster-register" with CrashLoopBackOff: "back-off 5m0s restarting failed container=cluster-register pod=cattle-cluster-agent-67dc69955d-nktfm_cattle-system(ee7e29e6-ebee-4c9a-8448-8f6051466c0a)"
I1020 20:13:24.001867   12224 setters.go:77] Using node IP: "172.31.11.103"
```

Himansh-ilink:
I am also facing this issue in my cluster,  can someone please look into it from rancher team.**This cluster is currently Error; areas that interact directly with it will not be available until the API is ready.

Cluster health check failed: cluster agent is not ready**

izaac:
Rancher version: `master-head (10/21/2020)` _0390ff5_

Letsencrypt and byo-valid certs install methods worked fine now.
I was able to provision Custom and Driver RKE clusters.

sowmyav27:
**On 2.5-head - commit id: `56bf444`**

- Deployed a DO cluster - 1 etcd, 1 control and 2 worker nodes.
- Cluster has come up active

[...]

PR #29663, opened 2020-10-21
Fix cluster agent failed to start if no ca setting

StrongMonkey:
Do not setup listener if httpsPort is zero. This is because when agent
deploys it only listens to httpsPort which requires cacerts setting to
be set. Agent will not install any settings because of the usecase for
rancher in rancher(otherwise setting will be overriden)


StrongMonkey:
https://github.com/rancher/rancher/issues/29652

ibuildthecloud (approved):

Issue #29599, opened 2020-10-17
airgap - rancher/rancher-operator:v0.1.0 is used but rancher/rancher-operator:v0.1.0-alpha8 is in the images.txt file


izaac:
**What kind of request is this:**
bug

**Steps to reproduce (least amount of steps as possible):**

- Install Rancher `v2.5.2-rc1` Airgap.
- Check the logs and Rancher will be trying to pull `rancher/rancher-operator:v0.1.0`

**Result:**

`Error syncing pod 80a1d956-ef65-45f8-bdb3-30ca650b229f ("rancher-operator-8667f7c985-48zcf_rancher-operator-system(80a1d956-ef65-45f8-bdb3-30ca650b229f)"), skipping: failed to "StartContainer" for "rancher-operator" with ImagePullBackOff: "Back-off pulling image \"<private_registry>/rancher/rancher-operator:v0.1.0\""`

**Other details that may be helpful:**

`rancher/rancher-operator:v0.1.0-alpha8` is in `rancher-images.txt`

`rancher-operator-system   rancher-operator-8667f7c985-48zcf   0/1     ImagePullBackOff   0          9m57s`

cbron:
This looks like a duplicate of https://github.com/rancher/rancher/issues/29591  @izaac / @jiaqiluo

jiaqiluo:
@cbron yep, the same symptom for a different image.

mrajashree:
Can be tested with v2.5.2-rc3

izaac:
Rancher version: `v2.5.2-rc3 (10/21/2020)`

rancher-images.txt has now `rancher/rancher-operator:v0.1.0`.
No errors in logs after fresh install as previously noticed.

Issue #29591, opened 2020-10-17
airgap - rancher/fleet:v0.3.0 is used but rancher/fleet:v0.3.0-rc3 is in the images.txt file


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- install rancher `v2.5.2-rc1` in HA mode in airgap environment 

**Result:**
- in the local cluster, the fleet-controller uses the image `rancher/fleet:v0.3.0` which is not in the `images.txt` file, instead, `rancher/fleet:v0.3.0-rc3` is included. 

 
<img width="1481" alt="96323631-c81fad00-0fd2-11eb-871e-79efe240a6a5" src="https://user-images.githubusercontent.com/6218999/96503684-824c2a00-1208-11eb-9cde-353208add5fe.png">

```
> cat rancher-images.txt | grep -i fleet
rancher/fleet-agent:v0.3.0-rc3
rancher/fleet:v0.3.0-rc3
```

Which image should rancher `v2.5.2-rc1` use?

mrajashree:
Can be tested with v2.5.2-rc3

jiaqiluo:
the bug fix is validated in rancher:`v2.5.2-rc3` airgap installation 



![image](https://user-images.githubusercontent.com/6218999/96793961-2d4c1780-13b2-11eb-8e49-b2a0281abe4b.png)

```
> cat rancher-images.txt | grep -i fleet
rancher/fleet-agent:v0.3.0
rancher/fleet:v0.3.0
```

PR #29664, opened 2020-10-21
[2.5] Fix image list for airgap

PennyScissors:
**Problem:**
The tarballs of the latest version of each chart needs to be extracted before the `rancher-images.txt` for airgap setups can be created. The previous logic that does this pre-processing leverages on `sort -Vr` (reverse version-sort) to find the latest version of each chart, but it's proven to be unreliable at times as shown in the related issues below.

**Solution:**
Modify pre-processing logic to read the repo's `index.yaml` using `yq` to find the latest version of each chart. Since we don't allow modifying old charts, it's safe to assume that `index.yaml` will always contain a list of chart sorted in the correct order.

Ran `make ci` locally and the generated `rancher-images.txt` contained the correct tags for `rancher/fleet` and `rancher/rancher-operator`

**Related Issues:**
https://github.com/rancher/rancher/issues/29591
https://github.com/rancher/rancher/issues/29599

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -91,6 +91,9 @@ RUN curl -sLf https://github.com/rancher/k3s/releases/download/${CATTLE_K3S_VERS
 
 RUN wget -O - ${!DOCKER_URL} > /usr/bin/docker && chmod +x /usr/bin/docker
 
+ENV YQ_URL=https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_${ARCH}
+RUN wget -O - ${YQ_URL} > /usr/bin/yq && chmod +x /usr/bin/yq
+
 ENV KUBECTL_URL=https://storage.googleapis.com/kubernetes-release/release/v1.11.0/bin/linux/${ARCH}/kubectl
 RUN wget -O - ${KUBECTL_URL} > /usr/bin/kubectl && chmod +x /usr/bin/kubectl
 
--- a/scripts/package
+++ b/scripts/package
@@ -1,23 +1,14 @@
 #!/bin/bash
 set -e
 
-extract_latest_version_tgz() {
-    # Takes in a path to a chart's directory, finds the latest version's tgz and extracts it
-    # All crd tgz are ignored
-    # Max depth is set to prevent extracting a tgz contained within another tgz, which is the case for charts containing a helm repo
-    LATEST_VERSION_TGZ_PATH=$(find $1 -maxdepth 1 -name "*.tgz" ! -name "*crd*.tgz" -print | sort -Vr | head -1)
-    if [[ $LATEST_VERSION_TGZ_PATH ]]; then
-        tar -xvf $LATEST_VERSION_TGZ_PATH -C $(dirname $LATEST_VERSION_TGZ_PATH)
-    fi
-}
-export -f extract_latest_version_tgz
-
 ./k3s-images.sh
 
 source $(dirname $0)/version
 
 ARCH=${ARCH:-"amd64"}
+SYSTEM_CHART_REPO_DIR=build/system-charts
 SYSTEM_CHART_DEFAULT_BRANCH=${SYSTEM_CHART_DEFAULT_BRANCH:-"dev-v2.4"}
+CHART_REPO_DIR=build/charts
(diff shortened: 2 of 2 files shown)


Dockerfile.dapper
```diff
@@ -91,6 +91,9 @@ RUN curl -sLf https://github.com/rancher/k3s/releases/download/${CATTLE_K3S_VERS
 
 RUN wget -O - ${!DOCKER_URL} > /usr/bin/docker && chmod +x /usr/bin/docker
 
+ENV YQ_URL=https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_${ARCH}
```
cbron, line 94: Does this work on a mac ? I thought host arch would be darwin, but maybe since we are in docker its already linux, not sure how `DAPPER_HOST_ARCH` works. The download for mac is `yq_darwin_amd64`.

scripts/package
```diff
@@ -49,20 +40,32 @@ cd ../bin
...
 if [ ! -d build/charts ]; then
-    git clone --branch $CHART_DEFAULT_BRANCH https://github.com/rancher/charts build/charts
-
-    # Iterate through chart directories and execute callback to extract latest version tgz
-    find build/charts/assets -type d -maxdepth 1 -exec bash -c 'extract_latest_version_tgz {}' \;
+    INDEX_PATH=$CHART_REPO_DIR/index.yaml
+    git clone --branch $CHART_DEFAULT_BRANCH https://github.com/rancher/charts $CHART_REPO_DIR
+
+    # Get list of paths to tarballs of the latest version of each chart from Helm repo's index.yaml
+    LATEST_TGZ_PATHS=$(yq r $INDEX_PATH "entries.*.[0].urls[0]")
```
cbron, line 51: I suppose urls should never be blank, but if it is would yq error out here ? Same for entries actually, I wonder how a blank index.yaml reacts, you can make one with `helm repo index .`

cbron (commented):
Logic and approach looks good. This feels like a cleaner solution than before.

scripts/package
PennyScissors: For blank index.yaml (completely blank file), entries or urls, the result would be that nothing or that specific url will not get extracted and images will not get included in the images list because it doesn't exist or is invalid (missing url). yq will not throw an error.

Dockerfile.dapper
PennyScissors: I'm using a mac, but as you say it is probably linux already since we are using docker. I assumed we didn't need darwin because we don't use darwin for kubectl although there is one available.
https://github.com/rancher/rancher/blob/0390ff5bc542936c93969624c3644ddefb139b52/Dockerfile.dapper#L94

scripts/package
```diff
@@ -49,20 +40,32 @@ cd ../bin
...
+    git clone --branch $SYSTEM_CHART_DEFAULT_BRANCH https://github.com/rancher/system-charts $SYSTEM_CHART_REPO_DIR
 fi
 
 if [ ! -d build/charts ]; then
-    git clone --branch $CHART_DEFAULT_BRANCH https://github.com/rancher/charts build/charts
-
-    # Iterate through chart directories and execute callback to extract latest version tgz
-    find build/charts/assets -type d -maxdepth 1 -exec bash -c 'extract_latest_version_tgz {}' \;
+    INDEX_PATH=$CHART_REPO_DIR/index.yaml
+    git clone --branch $CHART_DEFAULT_BRANCH https://github.com/rancher/charts $CHART_REPO_DIR
```
StrongMonkey, line 48: nit: with `--depth=1 --no-tags` it should faster

scripts/package
cbron: Ok, I just want to ensure it doesn't error and break build. Same would go for `urls[0]` ? Should be easy enough to manually edit the index file and check by hand.

StrongMonkey (dismissed):
LGTM except one small nit @PennyScissors

scripts/package
PennyScissors: I tested it with a blank index.yaml, empty entries as in `entries: {}` which is what `helm repo index .` creates, and a chart without a `urls` key. There were no error from yq. In the missing/empty entries case nothing was extracted. In the missing urls case, only that specific chart didn't get extracted.

cbron (approved):

StrongMonkey (approved):

PR #29709, opened 2020-10-22
Change ingress spec when using newest apiVersion

brendarearden:
**Problem** 
Spec changes needed for new apiVersion

**Solution**
Spec has been updated to match new apiVersion requirements

**Issue**

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -1,4 +1,4 @@
-{{- if (.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress") }}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
 apiVersion: networking.k8s.io/v1
 {{- else }}
 apiVersion: networking.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


chart/templates/ingress.yaml
```diff
@@ -38,11 +38,20 @@ spec:
...
+            port:
+              number: 80
+          {{- else }}
           serviceName: {{ template "rancher.fullname" . }}
           servicePort: 80
+          {{- end }}
 {{- if eq .Values.tls "ingress" }}
   tls:
   - hosts:
+    {{- if .Values.hostname }}
```
rmweir, line 53: is hosts being included but empty valid?

brendarearden: `if` will check for both "" and `nil` so if you have `hostname: ` or `hostname: ""` in your values.yaml, it wont evaluate as true for the if check

mrajashree: I think this would still leave `hosts` as an empty key right? so maybe this `if` should be before `hosts`, else this is what it will be when hostname is nil/empty:
```
tls:
  - hosts:
    secretName: tls-rancher-ingress
```

brendarearden: Oh, I see what you are saying - from what I can tell, it would not be a valid - it seems like for the rules section you don't provide a `-host` field if you don't have a host name either, so we have already set it up kind of dependent on a host name existing.... for sure the tls `-host` has to match the rules `-host` or it won't work. I can remove the if statement around that section- our rancher install instructions tell you to provide a hostname so install just wont complete if they don't provide it

rmweir: Okay, that all makes sense.

rmweir (approved):
LGTM

cbron (approved):

mrajashree (approved):

PR #29710, opened 2020-10-22
Handle rollback actions

nickgerace:


tests/integration/suite/test_workloads.py
```diff
@@ -372,3 +415,14 @@ def wait_for_service_cluserip_reset(client, name, timeout=30):
         if time.time() - start > timeout:
             raise Exception('Timeout waiting for workload service')
     return services.data[0]
+
+
+def wait_for_workload(client, name, timeout=30):
+    start = time.time()
+    curr = client.list_workload()["data"][0]
```
cbron, line 422: Does that not error out right away with an index error ?  Feel like we should get a length count first to ensure its above 0. The line below would error out too. 
 https://github.com/rancher/rancher/blob/master/tests/integration/suite/test_app.py#L886-L897

tests/integration/suite/test_workloads.py
```diff
@@ -341,6 +341,49 @@ def _timestamp_reset():
...
+
+    wait_for_workload(client, workload_name)
+
+    with pytest.raises(ApiError) as e:
+        user.client.action(
+            obj=workload,
+            action_name="rollback",
+            answers={"asdf": "asdf"})
+    assert e.value.error.status == 404
+    client.delete(ns)
```
cbron, line 384: are we removing this since its already removed above on line 354 ?

tests/integration/suite/test_multi_cluster_app.py
```diff
@@ -614,6 +615,48 @@ def test_mcapp_rollback_validation(admin_mc, admin_pc, custom_catalog,
...
+        roleTemplateId="read-only")
+    remove_resource(prtb)
+
+    wait_until_available(user.client, project)
+
+    mcapp_name = random_str()
+    mcapp_user_read_only = "local://" + user.user.id,
+    mcapp = client.create_multi_cluster_app(
+        name=mcapp_name,
+        templateVersionId="cattle-global-data:library-wordpress-1.0.5",
```
cbron, line 643: Wordpress ?

nickgerace: This was the templateVersionId used from the other tests. @dramich suggested using ```docker-registry``` (less bloat), but I kept receiving an 500 HTTP error code. Need to double check this.

tests/integration/suite/test_workloads.py
nickgerace: Yes, good catch.

tests/integration/suite/test_workloads.py
nickgerace: I'll fix that, good catch.

tests/integration/suite/test_multi_cluster_app.py
nickgerace: Fixed.

tests/integration/suite/test_workloads.py
```diff
@@ -372,3 +424,20 @@ def wait_for_service_cluserip_reset(client, name, timeout=30):
...
+        """Return an empty string if the workload is not initialized yet."""
+        if len(client.list_workload()["data"]) > 0:
+            return client.list_workload()["data"][0]
+        return ""
+    start = time.time()
+    curr = _get_curr_workload()
+    while (curr["name"] != name) and (curr["status"] != "active"):
+        time.sleep(.5)
+        curr = _get_curr_workload()
+        if time.time() - start > timeout:
```
cbron, line 441: nit: this could be in while conditional.

cbron (dismissed):
Lgtm. 500 errors in test checks work for now until we can fix up later.

pkg/api/norman/customization/workload/workload.go
```diff
@@ -35,17 +40,31 @@ var (
...
+	// Create a RawResource with a minimal config. This will be used for the eventual "CanDo" method, which parses
+	// the name and namespace from a RawResource object.
+	workloadType := getTypeFromAPIContext(apiContext)
+	rawResource := &types.RawResource{
+		Values: map[string]interface{}{
+			"id":          name,
+			"namespaceId": namespace,
+		},
+	}
+	if a.canUpdateWorkload(apiContext, rawResource, workloadType) != nil {
```
nickgerace, line 65: Self-review: do not need a ```workloadType``` variable.

pkg/api/norman/customization/workload/workload.go
```diff
@@ -274,3 +296,49 @@ func splitID(id string) (string, string) {
...
+		rbac.ObjFromContext(apiContext, resource), a.Schemas[workloadType])
+}
+
+// Use with *types.Resource.ID with format "type:namespace:name".
+// Return the same ID that was passed in if the format is incorrect.
+func getTypeFromRawResource(resource *types.RawResource) string {
+	return strings.Split(resource.ID, ":")[0]
+}
+
+// Use with *types.Resource.ID with format "type:namespace:name".
```
nickgerace, line 340: Self-review: should say ```Use with *types.APIContext.ID```.

tests/integration/suite/test_workloads.py
```diff
@@ -341,6 +341,76 @@ def _timestamp_reset():
...
+    wait_for_workload(client, workload.id)
+
+    # Admin user updates the workload to yield a rollback option. We change the
+    # name below.
+    workload = client.reload(workload)
+    workload.containers = [{
+        'name': 'bar',
+        'image': 'nginx:1.19',
+    }]
+    wait_for_workload(client, workload.id)
```
dramich, line 397: extra wait

tests/integration/suite/test_multi_cluster_app.py
```diff
@@ -614,6 +616,79 @@ def test_mcapp_rollback_validation(admin_mc, admin_pc, custom_catalog,
...
+    remove_resource(mcapp)
+    wait_for_app(admin_pc, mcapp_name)
+
+    # Admin user updates the MCA to yield a rollback option. We change the
+    # image version below.
+    mcapp = client.reload(mcapp)
+    original_rev = mcapp.revisions().data[0].name
+    mcapp.templateVersionId = (
+        "cattle-global-data:library-docker-registry-1.8.1")
+    wait_for_app(admin_pc, mcapp_name)
```
dramich, line 676: extra wait, already did this above

pkg/api/norman/customization/workload/workload.go
```diff
@@ -35,17 +39,30 @@ var (
...
+
+	// Create a RawResource with a minimal config. This will be used for the eventual "CanDo" method, which parses
+	// the name and namespace from a RawResource object.
+	rawResource := &types.RawResource{
+		Values: map[string]interface{}{
+			"id":          name,
+			"namespaceId": namespace,
+		},
+	}
+	if a.canUpdateWorkload(apiContext, rawResource, getTypeFromAPIContext(apiContext)) != nil {
```
dramich, line 63: We are missing some error handling here. `canUpdateWorkload` returns errors but they are never used. If the type/id is invalid a user will never know and get a 404. That isn't terribly useful. After calling `getTypeFromID` do a check on a.Schemas to validate it's a valid type, if not give back a real error. 
Or update splitID to give back all three pieces then do a check on the type here and ignore the type in other places

dramich (changes requested):

pkg/api/norman/customization/workload/workload.go
```diff
@@ -274,3 +294,47 @@ func splitID(id string) (string, string) {
...
+
+// Return the same ID that was passed in if the format ("type:namespace:name") is incorrect.
+func getTypeFromRawResource(resource *types.RawResource) string {
+	return strings.Split(resource.ID, ":")[0]
+}
+
+// Return the same ID that was passed in if the format ("type:namespace:name") is incorrect.
+func getTypeFromAPIContext(apiContext *types.APIContext) string {
+	return strings.Split(apiContext.ID, ":")[0]
+}
```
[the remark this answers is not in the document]
nickgerace: Will do

tests/integration/suite/test_workloads.py
nickgerace: Adding

tests/integration/suite/test_workloads.py
```diff
@@ -341,6 +341,76 @@ def _timestamp_reset():
...
+        containers=[{
+            'name': 'foo',
+            'image': 'nginx:1.18',
+        }])
+    remove_resource(workload)
+    wait_for_workload(client, workload.id)
+
+    # Admin user updates the workload to yield a rollback option. We change the
+    # name below.
+    workload = client.reload(workload)
```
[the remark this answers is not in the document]
nickgerace: Removing

tests/integration/suite/test_multi_cluster_app.py
nickgerace: Removing

pkg/api/norman/customization/workload/workload.go
nickgerace: Okay I can add that. If the type is invalid, is that worth returning an API error for?

tests/integration/suite/test_workloads.py
```diff
@@ -372,3 +442,21 @@ def wait_for_service_cluserip_reset(client, name, timeout=30):
         if time.time() - start > timeout:
             raise Exception('Timeout waiting for workload service')
     return services.data[0]
+
+
+def wait_for_workload(client, id, timeout=30):
```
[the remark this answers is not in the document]
nickgerace: I tried to find one, but I'll double check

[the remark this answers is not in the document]
nickgerace: https://github.com/rancher/rancher/blob/6291329896885ef6844426b97f36a39777e4c049/tests/validation/tests/v3_api/common.py#L602

This helper? It's in the validation library, but I can use it.

pkg/api/norman/customization/workload/workload.go
nickgerace: Actually, I do not know if we need to add extra error handling functionality given that we have the switch block handle whether or not the type is valid.

https://github.com/rancher/rancher/pull/29710/files/3fdf884e780f3d4c883b10c917209bcf6a061390#diff-78e6871c27609dec2cd98962aa49b9d15cb5dd9e35923aa279735c148aeefde9R325

What do you think about returning (bool, err) for this function, and having the ActionHandler return an API error based on the result? Example...

```go
if err := a.canUpdateWorkload(...); err != nil {
    if strings.Contains(err.Error(), "invalid workload type provided") {
        // return non-404 error
    } else {
        // return 404 error
    }
}

Comment (nickgerace) [reply pkg/api/norman/customization/workload/workload.go:63]: We can also change this function to return (bool, error) as well. Thoughts?

Comment (dramich) [reply pkg/api/norman/customization/workload/workload.go]: If you update canUpdateWorkload to return an error make it a typed error so you can just do a type check vs attempting to read the string

File: tests/integration/suite/test_workloads.py (hunk 12 above)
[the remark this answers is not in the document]
Comment (nickgerace) [reply tests/integration/suite/test_workloads.py:445]: I'll take another pass, but I believe we do need this new function. Resolving this.

File: pkg/api/norman/customization/workload/workload.go (hunk 9 above)
Comment (nickgerace) [reply pkg/api/norman/customization/workload/workload.go]: Agree with that. Will do

File: tests/integration/suite/test_workloads.py [hunk 13]
```diff
@@ -372,3 +440,18 @@ def wait_for_service_cluserip_reset(client, name, timeout=30):
...
             raise Exception('Timeout waiting for workload service')
     return services.data[0]
+
+
+def wait_for_workload(client, workload_id, timeout=30):
+    """Wait for a given workload to be active."""
+    def _get_result():
+        try:
+            return client.by_id_workload(
+                workload_id)["deploymentStatus"]["conditions"][-1]["status"]
```
dramich, line 450: This should check that the `state` is active, just grabbing the last condition could change meaning at any time.

nickgerace: Agreed. Done.

dramich (approved):

cbron (approved):

PR #29744, opened 2020-10-23
Fix test_writing_config_to_disk timeout issues

thedadams:
This test was timing out. Instead of waiting for the file to be written
to disk, the tests waits for the node to have a certain condition. This
indicates that the test is ready to check for the file.

Related Issue:
https://github.com/rancher/tasks/issues/1

### DIFF
--- a/tests/integration/suite/test_node.py
+++ b/tests/integration/suite/test_node.py
@@ -4,7 +4,7 @@ import pytest
 from rancher import ApiError
 from kubernetes.client import CoreV1Api
 from .common import auth_check, random_str, string_to_encoding
-from .conftest import wait_for
+from .conftest import wait_for, wait_for_condition
 import time
 
 
@@ -139,9 +139,9 @@ def test_writing_config_to_disk(admin_mc, wait_remove_resource):
         digitaloceancredentialConfig={"accessToken": "test"})
     wait_remove_resource(cloud_credential)
 
-    data = {'userdata': 'do cool stuff\n',
+    data = {'userdata': 'do cool stuff' + random_str() + '\n',
             # This validates ssh keys don't drop the ending \n
-            'id_rsa': 'some\nfake\nstuff\n'
+            'id_rsa': 'some\nfake\nstuff\n' + random_str() + '\n'
             }
 
     def _node_template():
@@ -166,6 +166,15 @@ def test_writing_config_to_disk(admin_mc, wait_remove_resource):
         nodeTemplateId=node_template.id,
         hostnamePrefix="test1",
         clusterId="local")
+
+    def node_available():
+        node = client.list_node(nodePoolId=node_pool.id)
+        if len(node.data):
+            return node.data[0]
+        return None
+
+    node = wait_for(node_available)
+    wait_for_condition("Saved", "False", client, node)
     wait_remove_resource(node_pool)
 
     for key, value in data.items():


tests/integration/suite/test_node.py
```diff
@@ -160,13 +160,21 @@ def _node_template():
...
-    wait_remove_resource(node_template)
 
     node_pool = client.create_node_pool(
         nodeTemplateId=node_template.id,
         hostnamePrefix="test1",
         clusterId="local")
+
+    def node_available():
+        if len(client.list_node(nodePoolId=node_pool.id).data):
+            return client.list_node(nodePoolId=node_pool.id).data[0]
```
nickgerace, line 171: If ```list_node``` makes an API call rather than grabs an object locally, I'd cache it a line above...

```python
node = client.list_node(nodePoolId=node_pool.id)
if len(node.data):
    return node.data[0]
```

nickgerace (approved):
LGTM! (I think I am using that right...)

tests/integration/suite/test_node.py
```diff
@@ -160,13 +160,22 @@ def _node_template():
...
     node_pool = client.create_node_pool(
         nodeTemplateId=node_template.id,
         hostnamePrefix="test1",
         clusterId="local")
+
+    def node_available():
+        node = client.list_node(nodePoolId=node_pool.id)
+        if len(node.data):
+            return node.data[0]
+        return None
```
aiyengar2, line 173: nit: Python returns `None` by default so this isn't necessary (unless you added it for readability?)
```suggestion
```

tests/integration/suite/test_node.py
```diff
@@ -139,9 +139,9 @@ def test_writing_config_to_disk(admin_mc, wait_remove_resource):
         digitaloceancredentialConfig={"accessToken": "test"})
     wait_remove_resource(cloud_credential)
 
-    data = {'userdata': 'do cool stuff\n',
+    data = {'userdata': 'do cool stuff' + random_str() + '\n',
```
aiyengar2, line 142: Is `userdata` always expected to be just one line or does multiline input need to be tested here?

If multi-line input should be tested, should there be a `/n` at the end of `do cool stuff` (or whatever it is changed to)?

aiyengar2 (dismissed):
Left two minor comments for review but the PR looks good to me either way!

tests/integration/suite/test_node.py
thedadams: `userdata` was already one line. I added the `random_str()` in the middle be testing on my local machine would always pass this test because the contents of the file was not changing.

tests/integration/suite/test_node.py
```diff
@@ -160,13 +160,22 @@ def _node_template():
     node_template = wait_for(_node_template,
                              fail_handler=lambda:
                              'failed to create node template')
-    wait_remove_resource(node_template)
```
cbron, line 163: This should stay put, because if a lower line fails for some reason it still gets cleaned up post-method completion. If this is below that might not happen. It won't remove it early.

tests/integration/suite/test_node.py
```diff
@@ -160,13 +160,22 @@ def _node_template():
...
         clusterId="local")
+
+    def node_available():
+        node = client.list_node(nodePoolId=node_pool.id)
+        if len(node.data):
+            return node.data[0]
+        return None
+
+    node = wait_for(node_available)
+    wait_for_condition("Saved", "False", client, node)
```
cbron, line 176: Why false ?

thedadams: The node that is created with a Digital Ocean driver, but the SSH information is fake and will cause the "save" to fail even though the files are created.

cbron (approved):
Good to merge, but only into master. Open backport PRs into release/v2.5 and 2.4 for later merging.

PR #29748, opened 2020-10-24
Fix nil pointer dereference in EKS controller

thedadams:
The  catalog manager was not initialized with the eksOperatorController
and this caused a nil pointer dereference when it was used.

Caused by:
https://github.com/rancher/rancher/pull/29559


rmweir (approved):
LGTM

dramich (approved):

PR #29759, opened 2020-10-26
Add check for taints to determine if pod can run

brendarearden:
**Problem**
Even when some nodes are labelled as control plane, they still have the ability to run pods

**Solution**
Check for taints to determine if pods can be executed on that node and count it towards worker stats if it can

**Issue**

### DIFF
--- a/pkg/controllers/management/clusterstats/statsaggregator.go
+++ b/pkg/controllers/management/clusterstats/statsaggregator.go
@@ -68,7 +68,7 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 		if !m.Spec.Worker && !m.Spec.ControlPlane && !m.Spec.Etcd {
 			return errors.Errorf("node role cannot be determined because node %s has not finished syncing. retrying...", m.Status.NodeName)
 		}
-		if m.Spec.Worker && !m.Spec.InternalNodeSpec.Unschedulable {
+		if (m.Spec.Worker || isNotTainted(m)) && !m.Spec.InternalNodeSpec.Unschedulable {
 			machines = append(machines, m)
 		}
 	}
@@ -224,4 +224,13 @@ func (s *StatsAggregator) machineChanged(key string, machine *v3.Node) (runtime.
 		s.Clusters.Controller().Enqueue("", machine.Namespace)
 	}
 	return nil, nil
+}
+
+func isNotTainted(m *v3.Node) bool {
+	for _, taint := range m.Spec.InternalNodeSpec.Taints {
+		if taint.Value == "true" && (taint.Effect == "NoSchedule" || taint.Effect == "NoExecute") {
+			return false
+		}
+	}
+	return true
 }


pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -225,3 +225,12 @@ func (s *StatsAggregator) machineChanged(key string, machine *v3.Node) (runtime.
 	}
 	return nil, nil
 }
+
+func isNotTainted(m *v3.Node) bool {
+	for _, taint := range m.Spec.InternalNodeSpec.Taints {
+		if taint.Value == "true" && (taint.Effect == "NoSchedule" || taint.Effect == "NoExecute") {
```
cbron, line 231: I was curious about the `true` part here, and when I looked at k3s and I don't actually see a taint on any of the nodes.

brendarearden: what kind of nodes were they? 
I've seen this added to nodes: 
```
"taints": [
{
"effect": "NoExecute",
"key": "node-role.kubernetes.io/etcd",
"type": "/v3/schemas/taint",
"value": "true"
}
],
```
and this one as well
```
"taints": [
{
"effect": "NoSchedule",
"key": "node-role.kubernetes.io/controlplane",
"type": "/v3/schemas/taint",
"value": "true"
}
],
```

brendarearden: I assumed that with a value true or false, then that means you could have a taint, but disable it. If that is not the case, then the remove check could be removed

cbron: I spun up a k3d cluster with 2 workers and all 3 of the nodes don't have any taints. Are your taints from RKE cluster or k3s ?

aiyengar2: According to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#taint-v1-core, `value` is not necessarily required and based on https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/, removing a taint is equivalent to setting the `value` to empty. Maybe the check here should instead be:
```suggestion
		if taint.Value != "" && (taint.Effect == "NoSchedule" || taint.Effect == "NoExecute") {
```

brandond: Taints are complicated. Like labels, they have a key, value, and effect. Kubeadm for example sets
`node-role.kubernetes.io/master=:NoExecute` which is `{Key="node-role.kubernetes.io/master", Value="", Effect="NoSchedule"}`.  Removing this from the command line would NOT involve setting it to an empty string, instead you put a - at the end of the effect: `kubectl taint node node1 node-role.kubernetes.io/master=:NoExecute-`

A toleration for this would normally just use the Exists operator, as opposed for only tolerating a specific value:
```
  - key: "example-key"
    operator: "Exists"
    effect: "NoSchedule"
```

cbron:
closing in favor of other PR

Issue #26366, opened 2020-03-30
Cluster Private Registry setting not adhered to when pulling rancher-agent image during Node Driver cluster provisioning


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

_Edited to clarify reproduction steps, and call-out this appears to be Node Driver provisioning related, not Azure specific._

> The following steps assume you have an internet-accessible private Docker registry populated with the Rancher [v2.4.8 rancher-images.txt](https://github.com/rancher/rancher/releases/v2.4.8) images and a Rancher v2.4.8 instance (I used https://github.com/superseb/tf-do-rancher2/ for the latter). In the tests below I have replaced the actual private registry URL with registry.example.com:

The issue is that the initial rancher-agent container, started by Rancher, on a node during a Node Driver cluster provisioning does not adhere to the Private Registry configured on the cluster. Instead it will attempt to pull the image from Docker Hub (by default) or the `system-default-registry` (if configured). As a result cluster provisioning for Node Driver clusters will fail if Docker Hub, or the `system-default-registry` (if configured), is un-reachable from the node. Below are the three tests I performed for this:

#### Working as expected: Custom cluster with Private Registry configured

1. Create a custom cluster with Kubernetes v1.18.6-rancher1-2, specifying a Private Registry in the cluster configuration with the URL, user and password for your private registry.
2. Observe the agent command, generated by Rancher, for the cluster correctly references the private registry:
```
sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run registry.example.com/rancher/rancher-agent:v2.4.8 --server https://161.35.32.125 --token <token> --ca-checksum 25710285359810f586aefef46cd3673efdb97be0131d4942dc0e94489b933032 --worker --etcd --controlplane
```
3. On a host with Docker installed, login to the private registry `docker login registry.example.com` and execute the agent command. Observe the node is successfully provisioned with all Kubernetes component and System Project workload images pulled from the private registry.

#### Not working as expected: Digital Ocean Node Driver cluster with Private Registry configured

1. Create a single node Digital Ocean Node Driver cluster with Kubernetes v1.18.6-rancher1-2, specifying a Private Registry in the cluster configuration with the URL, user and password for your private registry.
[...]

PR #29769, opened 2020-10-26
[2.4] Enable agent image to be pulled from private registry

ryansann:
Issue:
- rancher/rancher#26366 

Solution:
- Resolve the agent image using the cluster object so that it is qualified with the private registry url
- If there's a cluster private registry defined, execute a docker login command on the downstream node via `rancher-machine` before the agent is deployed
- Escape special characters in password since login is via command line

Testing:
- Setup a private registry w/ rancher images
- Launch a cluster w/ a private registry
- tested w/ passwords:
```
testpassword
a$b"'@123`*\
```

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -17,6 +17,7 @@ import (
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
 	"github.com/rancher/rancher/pkg/api/customization/clusterregistrationtokens"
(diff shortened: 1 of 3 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	util "github.com/rancher/rancher/pkg/cluster"
+
```
luthermonson, line 16: nit

luthermonson (approved):

PR #29772, opened 2020-10-27
remove ambiguous methods.

jiaqiluo:
Problem:
The method name `add_cluster_to_rancher` is confusing. It sounds like it will add a cluster to Rancher while it actually runs the registry command on a list of nodes. 

Solution:
Remove this method since it is being used at only 3 places and 2 of them are passed in only 1 node.


rancher-max (approved):
LGTM

sowmyav27 (approved):
LGTM

PR #29789, opened 2020-10-27
Fix helm repo constant refreshing

thedadams:
The helm repo was being refreshed because `repoSpec.URL` and `status.URL` were different. They were different because we were ensuring that `status.URL` ended with a '/'. Now we only add the slash if it is needed to download a chart.

The issue is that `base.ResolveReference(u)` will remove the last part of the path unless it ends in a '/'.
You can see the difference in this playground: https://play.golang.org/p/DbXvDC_S6ZX

### DIFF
--- a/pkg/catalogv2/http/download.go
+++ b/pkg/catalogv2/http/download.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strings"
 
 	"github.com/rancher/wrangler/pkg/schemas/validation"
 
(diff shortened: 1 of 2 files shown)


aiyengar2 (approved):

cbron:
https://github.com/rancher/rancher/issues/29557

cbron (commented):
lgtm, with note that we might have to fix icon code, waiting for other reviewers too

cbron (approved):
lgtm, needs a backport after everyone approves

aiyengar2 (approved):

rmweir (approved):
LGTM

Issue #29754, opened 2020-10-26
Auto Replace does not work (v2.5.1 on rke)


Negashev:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):** 

1) install k8s by rke 1.2.1 on one node
2) install rancher 2.5.1 by helm
3) create cluster with hetzner node driver
4) add pool with auto replace node after 1-10 minutes
5) go to hetzner console and stop on node from pool (with auto replace)

**Result:**
Rancher see that kubelet stop, but nothing happend with node in next 30+ minutes

**Other details that may be helpful:**

I have only one node with k8s for rancher
and test cluster with one node for etcd and control panel, and 2 nodes (node pool with auto-replace)

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.1
- Installation option (single install/HA): HA with one node k8s

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): hetzner docker-driver
- Machine type (cloud/VM/metal) and specifications (CPU/memory): CX21
- Kubernetes version (use `kubectl version`): 

v1.19.3

- Docker version (use `docker version`):

 19.3.13

sowmyav27:
On 2.4.8
Deploy a DO cluster node driver on a single node rancher install.
Worker node's `Auto Replace` value is 5 minutes.
When the cluster comes up Active, power off the worker nod ein the cluster
the node will be seen as `Unavailable`
After 5 minutes, the node will be removed and another worker node is seen in `Provisioning` state.

**On 2.5.1**

- Deploy a DO cluster node driver on a single node rancher install.
- Worker node's `Auto Replace` value is 5 minutes.
- When the cluster comes up Active, power off the worker node in the cluster
- the node will be seen as `Unavailable`
- A new node is not provisioned after 5 minutes.

mrajashree:
This affects only k8s 1.19 clusters, because the node doesn't get the taint `node.kubernetes.io/unreachable:NoExecute`

sowmyav27:
**On 2.5-head commit id: `a90aa3ca` and master-head commit id: `b7e8c0d3`**

- Deploy a DO cluster node driver on a single node rancher install.
- Worker node's Auto Replace value is 5 minutes.
- When the cluster comes up Active, power off the worker node in the cluster
- the node will be seen as Unavailable
[...]

PR #29790, opened 2020-10-27
Match unreachable taint key without comparing effect

mrajashree:
Problem: Till k8s 1.18.10, the k8s node lifecycle controller sets these two taints on an unreachable node (or a node with Ready=Unknown): `node.kubernetes.io/unreachable:NoSchedule` and `node.kubernetes.io/unreachable:NoExecute`.
In k8s 1.19.3, the node controller is only setting the taint with NoSchedule effect for such a node.
The rancher nodepool controller checks for the presence of a taint with key `node.kubernetes.io/unreachable` and effect `NoExecute` when determining if a node is unreachable and needs to be replaced. Since the node in 1.19 does not get this taint, nodepool controller does not consider it as unreachable.

Solution: Modify the check in nodepool controller to consider a node unreachable even if only the taint key matches
 `node.kubernetes.io/unreachable`. The taint effect should not impact how nodepool processes this node.

https://github.com/rancher/rancher/issues/29754

### DIFF
--- a/pkg/controllers/management/nodepool/nodepool.go
+++ b/pkg/controllers/management/nodepool/nodepool.go
@@ -24,12 +24,7 @@ import (
 )
 
 var (
-	nameRegexp       = regexp.MustCompile("^(.*?)([0-9]+)$")
-	unReachableTaint = v1.Taint{
-		Key:    "node.kubernetes.io/unreachable",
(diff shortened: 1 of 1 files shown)


aiyengar2 (approved):
LGTM

dramich (approved):

PR #29793, opened 2020-10-27
Add check for taints to determine if pod can run

rmweir:
Version2 of: https://github.com/rancher/rancher/pull/29759

**Problem**
Even when some nodes are labelled as control plane, they still have the ability to run pods

**Solution**
Check for taints to determine if pods can be executed on that node and count it towards worker stats if it can

**Issue**

### DIFF
--- a/pkg/controllers/management/clusterstats/statsaggregator.go
+++ b/pkg/controllers/management/clusterstats/statsaggregator.go
@@ -18,6 +18,13 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+const (
+	nodeRoleControlPlane       = "node-role.kubernetes.io/controlplane"
+	nodeRoleControlPlaneHyphen = "node-role.kubernetes.io/control-plane"
+	nodeRoleETCD               = "node-role.kubernetes.io/etcd"
+	nodeRoleMaster             = "node-role.kubernetes.io/master"
+)
+
 type StatsAggregator struct {
 	NodesLister    v3.NodeLister
 	Clusters       v3.ClusterInterface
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -68,7 +68,7 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 		if !m.Spec.Worker && !m.Spec.ControlPlane && !m.Spec.Etcd {
 			return errors.Errorf("node role cannot be determined because node %s has not finished syncing. retrying...", m.Status.NodeName)
 		}
-		if m.Spec.Worker && !m.Spec.InternalNodeSpec.Unschedulable {
+		if (m.Spec.Worker || isNotTainted(m)) && !m.Spec.InternalNodeSpec.Unschedulable {
 			machines = append(machines, m)
 		}
```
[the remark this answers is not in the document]
rmweir: I think we were going to keep this for now just in case, although it is probably covered.

brandond (approved):
LGTM!

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -18,6 +18,11 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+const (
+	nodeRoleControlPlane = "node-role.kubernetes.io/controlplane"
+	nodeRoleETCD         = "node-role.kubernetes.io/etcd"
```
cbron, line 23: These are the Rancher taints, but currently [kubeadmn uses `master`](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#control-plane-node-isolation). However, [it is moving to `control-plane`](https://github.com/kubernetes/enhancements/issues/2067). So my guess is we should also add both of those.

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -225,3 +236,16 @@ func (s *StatsAggregator) machineChanged(key string, machine *v3.Node) (runtime.
 	}
 	return nil, nil
 }
+
+func isTaintedNoExecuteNoSchedule(m *v3.Node) bool {
+	for _, taint := range m.Spec.InternalNodeSpec.Taints {
+		if taint.Key != nodeRoleControlPlane && taint.Key != nodeRoleETCD &&
+			taint.Key != nodeRoleControlPlaneHyphen && taint.Key != nodeRoleMaster {
+			continue
+		}
```
cbron, line 245: nit: feel like this is confusing conditional. You could do `notWorker := a || b || c` then `if notWorker && taint == NoSchedule; return true`

cbron (approved):

pkg/controllers/management/clusterstats/statsaggregator.go
rmweir: @cbron Did something equivalent but without the inverted logic. Also, I try to avoid return the true case outside of conditionals as it's more idiomatic per: https://dave.cheney.net/2020/02/23/the-zen-of-go, but I think this reads fine now.

cbron (approved):
looks great

Issue #27080, opened 2020-05-14
Add support for GovCloud regions for EKS


superseb:
https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-eks-now-available-in-the-aws-govcloud-us-regions/

```
Amazon Elastic Kubernetes Service (EKS) is now available in AWS GovCloud (US) Regions.
```

gzrancher/rancher#12101

sowmyav27:
**On master-head - commit id - `33f755f1809`**

- Create a cloud cred with region - us-gov-west-1
- Deploy a hosted EKS cluster by giving in the creds
- Error seen in console and UI

<img width="1674" alt="Screen Shot 2020-09-16 at 8 53 21 PM" src="https://user-images.githubusercontent.com/26032343/93418764-effae400-f85f-11ea-8cbc-5eff9187d0b2.png">

thedadams:
After some investigation with @rmweir, it seems that maybe the hostname is incorrect in the request. When we make a request manually to `iam.us-gov-west-1.amazonaws.com`, we immediately get a `401` response.

When we make a request to `iam.us-gov.amazonaws.com` (the one in the error screenshot above), the request appears to timeout.

Is this perhaps an issue with the UI making a request to the wrong hostname?

thedadams:
According to https://docs.amazonaws.cn/en_us/general/latest/gr/iam-service.html, the above is incorrect.

sowmyav27:
**On 2.5-head commit id: `393468fd` and master-head commit id: `5aa3918`**

Deployed hosted EKS clusters in `us-gov-west-1` region.
Clusters came up Active. Validation checks works fine on the clusters

sowmyav27:
Blocked because of issue - https://github.com/rancher/rancher/issues/29103#issuecomment-733993630

sowmyav27:
**On 2.5-head - commit id: `b2953cbc` and master-head - commit id: `278c1f988`**

- Able to import an EKS gov cloud. Edit cluster--> Add node group -->Save changes made.
- Node group is seen added and the cluster comes up Active.

PR #29795, opened 2020-10-27
Add support for Amazon us-gov regions

thedadams:
The Amazon IAM requests use the "global" region. However, the global
region for us-gov is different from standard AWS. The global region and
host are added for us-gov.

Issue:
https://github.com/rancher/rancher/issues/27080

### DIFF
--- a/pkg/httpproxy/sign.go
+++ b/pkg/httpproxy/sign.go
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	defaultAWSRegion = "us-east-1"
+	defaultAWSRegion      = "us-east-1"
+	defaultUSGovAWSRegion = "us-gov-west-1"
 )
 
 var requiredHeadersForAws = map[string]bool{"host": true,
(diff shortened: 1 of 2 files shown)


rmweir (dismissed):
LGTM

pkg/httpproxy/sign.go
```diff
@@ -115,8 +116,13 @@ func (a awsv4) getServiceAndRegion(host string) (string, string) {
...
 	}
 
-	// if no region is found, global endpoint is assumed. In this case us-east-1 should be used for signing:
+	// if no region is found, global endpoint is assumed.
 	// https://docs.aws.amazon.com/general/latest/gr/sigv4_elements.html
+	if strings.Contains(host, "us-gov") {
+		// For us-gov, the default region is us-gov-west-1
+		return service, defaultUSGovAWSRegion
+	}
+	// In this case us-east-1 should be used for signing
```
rmweir, line 125: nit: ik I wrote it originally but could remove mention of us-gov-west-1 and us-east-1 in comments since if the values change they will be outdated, plus the docs are linked for reference which should be sufficient.

thedadams: Done

mrajashree (approved):

pkg/httpproxy/sign.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	defaultAWSRegion = "us-east-1"
+	defaultAWSRegion      = "us-east-1"
+	defaultUSGovAWSRegion = "us-gov-west-1"
```
cbron, line 25: Any reason we went with west ? I'll ask PM if they have preference.

cbron (approved):

PR #29809, opened 2020-10-28
Treat wss as https

### LINKED ISSUE #25061: Accept wss in addition to https for x-forwarded-proto

Bugfix

**Steps to reproduce (least amount of steps as possible):**

Access Rancher behind a reverse proxy that sets the `x-forwarded-proto` header to `wss` when a websocket request is made to it over HTTPS.

**Result:**

Rancher looks for a value of `https` in the `x-forwarded-proto` header at

https://github.com/rancher/rancher/blob/7a898ff87276951d95310b4dacb87020aa9e0de9/pkg/dynamiclistener/server.go#L505-L507

to determine whether to serve the request; otherwise it falls through to a redirect on error condition. Rancher should also accept `wss` as a valid forwarded protocol. For reference, Traefik sets its value for this header at https://github.com/containous/traefik/blob/a20a6636b4582e4e7b562fc01f64567775fc8a9a/pkg/middlewares/forwardedheaders/forwarded_header.go#L143-L149

**Other details that may be helpful:**

This is a continuation of the self-closed ticket at https://github.com/rancher/rancher/issues/24202, and a much more in-depth background is offered by that user on Traefik's issue queue at https://github.com/containous/traefik/issues/5533#issuecomment-553823512.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `2.3.3`
- Installation option (single install/HA): HA, behind Traefik 2.0 (as noted above)

bradjones1:
Fixes #25061

### DIFF
--- a/pkg/clusterrouter/proxy/proxy_server.go
+++ b/pkg/clusterrouter/proxy/proxy_server.go
@@ -213,7 +213,7 @@ func (r *RemoteService) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	proto := req.Header.Get("X-Forwarded-Proto")
 	if proto != "" {
-		// Certain reverse proxies may use the non-standard wss for websockets; allow https to remain the default.
+	    // Certain reverse proxies may use the non-standard wss for websockets; allow https to remain the default.
 	    if proto == "wss" {
 	        proto = "https"
 	    }
--- a/pkg/clusterrouter/proxy/proxy_server.go
+++ b/pkg/clusterrouter/proxy/proxy_server.go
@@ -212,12 +212,12 @@ func (r *RemoteService) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	u.RawQuery = req.URL.RawQuery
 
 	proto := req.Header.Get("X-Forwarded-Proto")
-	if proto != "" {
+	req.URL.Scheme = "https"
+	// Certain reverse proxies may use the non-standard wss for websockets; allow https to remain the default.
+	if proto != "" && proto != "wss" {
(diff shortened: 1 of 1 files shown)


bradjones1:
Test failure looks unrelated?

pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -212,12 +212,12 @@ func (r *RemoteService) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	u.RawQuery = req.URL.RawQuery
 
 	proto := req.Header.Get("X-Forwarded-Proto")
-	if proto != "" {
+	req.URL.Scheme = "https"
+	// Certain reverse proxies may use the non-standard wss for websockets; allow https to remain the default.
+	if proto != "" && proto != "wss" {
```
ibuildthecloud, line 217: If proto == 'proto' and req.TLS == nil then the scheme will be "http".  We want to only consider the req.TLS parameter if proto isn't set.  This is because TLS is typically terminated early to https connections will req.TLS == nil.

ibuildthecloud: I'm not totally sure my comment it correct, but I think the logic is off.

bradjones1: @ibuildthecloud Thanks for the review. I think you're right. I've updated this with a specific override to transliterate `wss` to `https` - it's not as clean but I think given the issues you raise and the intent of this logic, that's the only way to do what we're attempting here.

alexmnyc (approved):

Issue #22790, opened 2019-09-11
Ability to restrict user to deploy only from catalog


janeczku:
**What kind of request is this (question/bug/enhancement/feature request):**
Enhancement

There are use cases that involve having Rancher manage Kubernetes clusters that are exlusively used to host applications deployed by *project users* from catalogs managed and provided by the *global admin team*.

This involves the requirement to limit *project users* to only be able to deploy and manage resources that are part of a catalog provided by the *global admin team* while not being able to create any project/workload resources on their own.

The required behavior of Rancher App deployment would more closely mirror the behavior of vanilla helm/tiller, where user needs permissions to deploy the chart (via tiller) but the permissions to create the deployed Kubernetes resources are derived from the role assigned to tiller itself. Similary in Rancher a user could have only permissions to deploy apps from a catalog, while Rancher ensures that all resources are created within the current project, thus maintaining tenancy.

maggieliu:
Code for this is merged. However, we don't have enough time to validate this to ensure it's bug free. The testing/validation/bug fixing for this will be done in 2.5.1

cloudnautique:
To test this functionality the high-level steps are:

1. create a service account and bind it to a cluster-admin role. 
2. Create a repository from YAML in the apps and marketplace.
  - Add the `serviceAccount` and `serviceAccountNamespace` to the spec.
3. log in as a read-only user.
4. Deploy an application from the repository added with a service account.

Result: 
1. The application should deploy without errors even though the logged-in user has no permissions to create resources.

For the repository service account, it would be good to test non-cluster admin-level permissions as well. The permission for the service account would need to be able to deploy the application.

The logged-in read-only user should not be able to install apps from system or library charts.

sowmyav27:
On 2.5.0 - 

- create a service account and bind it to a cluster-admin role n a downstream cluster.
```
kubectl create sa sa-1
kubectl create clusterrolebinding test-clusterrole-1 --clusterrole cluster-admin --serviceaccount sa-1:default
```
- Create a repository from YAML in the apps and marketplace. Add the serviceAccount and serviceAccountNamespace to the spec.
- log in as a read-only user.
- Deploy an application from the repository added with a service account.
- The helm operation pod is in Reconciling state. and Logs of it show: 
```
Waiting for Kubernetes API to be available
Waiting for Kubernetes API to be available
Waiting for Kubernetes API to be available
```
- User is NOT able to deploy apps.

sowmyav27:
[...]

PR #29816, opened 2020-10-28
[2.5]Fix service account doesn't passed into uninstall and insecureSkipTLS doesn't work in git

StrongMonkey:


ibuildthecloud (approved):

StrongMonkey:
https://github.com/rancher/rancher/issues/22790

PR #29850, opened 2020-10-30
Use annotation to fetch cluster repo info for uninstall

StrongMonkey:
For uninstall action it is called on app we need to fetch cluster repo
from annotation.

### DIFF
--- a/pkg/catalogv2/helmop/operation.go
+++ b/pkg/catalogv2/helmop/operation.go
@@ -92,7 +92,7 @@ func (s *Operations) Uninstall(ctx context.Context, user user.Info, namespace, n
 		return nil, err
 	}
 
-	user, err = s.getUser(user, namespace, name)
+	user, err = s.getUser(user, namespace, name, true)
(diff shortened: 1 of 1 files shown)


StrongMonkey:
https://github.com/rancher/dashboard/issues/1234

vincent99:
Uh, what now?  Why does uninstalling an app need a reference to the chart that installed it?

The `catalog.cattle.io/ui-source-repo[-type]` are meant as hints from the UI to itself to help prefer the same chart again on upgrade.  They were not intended to be read or relied on by the backend, and won't be present if you do helm CLI install, etc.

StrongMonkey:
@vincent99  talked with @ibuildthecloud and he thinks that it is ok to use this.... But I agree we need to put info in the backend to poplute this information. Clusterrepo is needed for this issue https://github.com/rancher/rancher/issues/22790 (we need to serviceaccount info on cluster repo) to perform helm install

ibuildthecloud (approved):

Issue #29292, opened 2020-09-24
High system load alert fails to trigger on 1.19.2


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a k3s v1.19.2 two node cluster. Used 2 VMs with 2vCPU and 8GB RAM.
2. Import the cluster into Rancher.
3. Enable monitoring
4. Create a Notifier
5. Edit the alert group for High CPU Load alert rule and add the notifier
6. Create a workload that causes high cpu load on the cluster. I used the Folding at Home App in the Rancher catalog and set the CPU limit to 1500m 

**Result:**
The alert is never triggered, even though monitoring graphs showed CPU load well above 1.

**Other details that may be helpful:**
Dug a little more into this and saw that k3s 1.18 clusters are exporting the `machine_cpu_cores` metric but k3s 1.19 clusters are not.

k3s 1.18.8:
![image](https://user-images.githubusercontent.com/468462/94209011-19c79280-fe80-11ea-9a5f-9e2e19e45e14.png)
k3s 1.19.2:
![image](https://user-images.githubusercontent.com/468462/94209037-2946db80-fe80-11ea-8e2e-8203b832d4df.png)

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.4.8
- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
Imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
AWS EC2 m5a.large (2vCPU, 8GB RAM)

- Kubernetes version (use `kubectl version`):
```
> kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.6", GitCommit:"d32e40e20d167e103faf894261614c5b45c44198", GitTreeState:"clean", BuildDate:"2020-05-20T13:16:24Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2+k3s1", GitCommit:"d38505b124c92bffd45f6e0654adb9371cae9610", GitTreeState:"clean", BuildDate:"2020-09-21T17:00:07Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

[...]

PR #29860, opened 2020-10-30
Fix for machine_cpu_cores being removed from cadvisor metrics

dnoland1:
See also https://github.com/rancher/rancher/issues/29292

### DIFF
--- a/pkg/controllers/managementuser/alert/initdata.go
+++ b/pkg/controllers/managementuser/alert/initdata.go
@@ -284,7 +284,7 @@ var entries = []entry{
 					},
 					MetricRule: &v32.MetricRule{
 						Description:    "The cpu load is higher than 100",
-						Expression:     `sum(node_load1) by (node)  / sum(machine_cpu_cores) by (node) * 100`,
+						Expression:     `sum(node_load1) by (node)  / sum(kube_node_status_capacity_cpu_cores) by (node) * 100`,
 						Comparison:     manager.ComparisonGreaterThan,
 						Duration:       "3m",
 						ThresholdValue: 100,
--- a/pkg/controllers/managementuser/alert/initdata.go
+++ b/pkg/controllers/managementuser/alert/initdata.go
@@ -284,7 +284,7 @@ var entries = []entry{
 					},
 					MetricRule: &v32.MetricRule{
 						Description:    "The cpu load is higher than 100",
-						Expression:     `sum(node_load1) by (node)  / sum(kube_node_status_capacity_cpu_cores) by (node) * 100`,
+						Expression:     `sum(node_load1) by (node)  / sum(kube_node_status_capacity{resource="cpu"}) by (node) * 100`,
 						Comparison:     manager.ComparisonGreaterThan,
 						Duration:       "3m",
 						ThresholdValue: 100,


pkg/controllers/managementuser/alert/initdata.go
```diff
@@ -284,7 +284,7 @@ var entries = []entry{
 					},
 					MetricRule: &v32.MetricRule{
 						Description:    "The cpu load is higher than 100",
-						Expression:     `sum(node_load1) by (node)  / sum(machine_cpu_cores) by (node) * 100`,
+						Expression:     `sum(node_load1) by (node)  / sum(kube_node_status_capacity_cpu_cores) by (node) * 100`,
```
aiyengar2, line 287: This is a deprecated metric that comes from `kube-state-metrics`. It will officially be removed in v2.x based on https://github.com/kubernetes/kube-state-metrics/issues/974.

However, this should be okay since:
- The change only impacts Alerting V1, not Alerting V2. Alerting V2 does not use MetricRules afaik (@codyrancher we aren't using this for https://github.com/rancher/dashboard/issues/812 right?)
- Alerting V2 is still on version 1.9.x of kube-state-metrics and is deprecated, so that is unlikely to change.

@guangbochen I'm adding you as a reviewer to also get a sign-off since this is a Monitoring / Alerting V1 issue. I want to confirm that using this metric is okay and that we have no plans of upgrading the `kube-state-metrics` version of Monitoring / Alerting V1.

aiyengar2 (approved):
@dnoland1 LGTM but we also need a backport to 2.4 since this will primarily only affect users who are in 2.4 (as users in 2.5 are expected to use Alerting V2, which doesn't use MetricRules at all).

Also, we should wait on a sign-off from the China team about whether this metric is okay to be used.

dnoland1:
@aiyengar2 Since this is only an issue on k8s 1.19+ and v2.4.x only supports up to k8s 1.18, is it really necessary to backport?

aiyengar2:
> @aiyengar2 Since this is only an issue on k8s 1.19+ and v2.4.x only supports up to k8s 1.18, is it really necessary to backport?

Hmm so this brings into question whether this fix is necessary at all. If we were planning to backport to 2.4, it would make sense to forward port it to 2.5. However, if it's only going to impact users who are using Monitoring / Alerting V1 in Rancher 2.5 (which is considered deprecated in favor of Monitoring / Alerting V2), then should we push out a fix at all or just encourage users to switch to V2 for k8s 1.19+ clusters?

@maggieliu are we still actively supporting Monitoring / Alerting V1 in Rancher 2.5?

maggieliu:
@dnoland1 @aiyengar2 Don't need to backport to 2.4 because k8s 1.19 is not supported in 2.4. As for whether or not we need this fix at all: technically we are not actively support v1 anymore because it's deprecated. However, if there is a high priority request from customer, then we will still put in fixes.

aiyengar2:
> @dnoland1 @aiyengar2 Don't need to backport to 2.4 because k8s 1.19 is not supported in 2.4. As for whether or not we need this fix at all: technically we are not actively supporting v1 anymore because it's deprecated. However, if there is a high priority request from customer, then we will still put in fixes.

@dnoland1 is this a high priority fix? Otherwise, I think we should avoid changing the Monitoring V1 for just Rancher 2.5, especially since this is a soon-to-be-removed metric from kube-state-metrics.

dnoland1:
It's nice to have. The CPU load out of box v1 alert will not work on k8s 1.19. Users can manually workaround the issue by replacing the metric in their alert.

pkg/controllers/managementuser/alert/initdata.go
guangbochen: this is LGTM since for v1 monitoring the latest v1 monitoring chart is using v1.9.7 `kube-state-metrics` and the metric is supported.
https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-monitoring/v0.2.0/values.yaml#L187

guangbochen (approved):

aiwantaozi (dismissed):

pkg/controllers/managementuser/alert/initdata.go
aiyengar2: @dnoland1, based on https://github.com/kubernetes/kubernetes/issues/95204#issuecomment-719445180, should the correct, non-deprecated call be the following?
```suggestion
						Expression:     `sum(node_load1) by (node)  / sum(kube_node_status_capacity{resource="cpu"}) by (node) * 100`,
```

aiyengar2 (changes requested):
Unsure if this change should be merged, waiting for confirmation on whether Monitoring V1 should be supported in k8s 1.19 clusters (which is what this fix is for)

aiyengar2 (approved):
LGTM, seems like we still want to support V1 tools in Rancher 2.5 so this should be good to merge as long as it has been tested + we get a committer review.

@dnoland1 can you make sure that the newly changed expression `kube_node_status_capacity{resource="cpu"}` outputs the same thing you expected from the deprecated metric? Not sure if you already tested it again before pushing out the commit I suggested.

dnoland1:
I tried putting the expression directly into the UI and worked as expected.

aiyengar2:
@dnoland1 similar to https://github.com/rancher/rancher/pull/30339, we need this PR to be opened from `dnoland1:somebranch` to `master` to avoid running into build failures because of Drone. Can you open another PR from your fork and close out this PR?

dnoland1:
Done - https://github.com/rancher/rancher/pull/30343

PR #29862, opened 2020-10-30
Use annotation to fetch cluster repo info for uninstall

StrongMonkey:
For uninstall action it is called on app we need to fetch cluster repo
from annotation.


mrajashree:
https://github.com/rancher/dashboard/issues/1234
Forward port of https://github.com/rancher/rancher/pull/29850

PR #29876, opened 2020-10-31
Ensure CatalogManager fields are set

rmweir:
Prior, nil pointer errors were being encountered due to
CatalogManager fields not being set.


cbron (approved):

mrajashree (approved):

rmweir:
https://github.com/rancher/rancher/issues/29875

Issue #29874, opened 2020-10-31
Helm error on Rancher upgrade from 2.5.1 to 2.5.2-rc4 and also on fresh install of 2.5.2-rc4


izaac:
**What kind of request is this:**
bug

**Steps to reproduce (least amount of steps as possible):**

- Install Rancher RKE HA v2.5.1 (Bring your own valid certs method)
- Here I configured a few apps like logging, monitoring, istio, etc.
- Upgrade HA to v2.5.2-rc4

**Result:**

`Error: UPGRADE FAILED: cannot patch "rancher" with kind Ingress: Ingress.extensions "rancher" is invalid: spec.rules[0].http.paths[0].pathType: Required value: pathType must be specified`

![Screen Shot 2020-10-30 at 5 22 12 PM](https://user-images.githubusercontent.com/27659/97766632-7ac43500-1ad4-11eb-95f6-428bda946ae0.png)

**Other details that may be helpful:**

```
 k get po -n cattle-system
NAME                               READY   STATUS             RESTARTS   AGE
helm-operation-cp6qz               0/2     Completed          0          12m
helm-operation-kxph9               0/2     Completed          0          11m
helm-operation-ntqxn               0/2     Completed          0          11m
helm-operation-pwgzb               0/2     Completed          0          10m
helm-operation-qcvwn               0/2     Completed          0          10m
rancher-589757dd7-ggmv2            0/1     CrashLoopBackOff   6          13m
rancher-589757dd7-h84dg            0/1     CrashLoopBackOff   6          13m
rancher-589757dd7-tc5l7            0/1     CrashLoopBackOff   6          13m
rancher-webhook-698dd9dc76-9vcb6   1/1     Running            0          11m
```

**Environment information**
- Rancher version: `v2.5.1` upgrade to `v2.5.2-rc4`
- Installation option: HA

**Cluster information**
- Cluster type: Infrastructure Provider Linode Driver RKE
- Machine type: VM Worker nodes 8 GB RAM and 4vcpu
- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:49:28Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #29878, opened 2020-10-31
add ingress pathtype

cbron:
https://github.com/rancher/rancher/issues/29874

Recreated the error, and successfully fixed it with this patch, using the default `ImplementationSpecific`: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#better-path-matching-with-path-types

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -47,6 +47,9 @@ spec:
           serviceName: {{ template "rancher.fullname" . }}
           servicePort: 80
           {{- end }}
+        {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
+        pathType: ImplementationSpecific
+        {{- end }}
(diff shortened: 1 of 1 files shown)


chart/templates/ingress.yaml
```diff
@@ -37,7 +37,8 @@ spec:
   - host: {{ .Values.hostname }}  # hostname to access rancher server
     http:
       paths:
-      - backend:
+      - pathType: ImplementationSpecific
```
mrajashree, line 40: I think this field is new in the networking.k8s api, so should we add a check for the api version?
`{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}`

mrajashree (approved):
don't have enough context on ingress pathType but the default options sounds right, LGTM

rawmind0 (approved):
LGTM

PR #29889, opened 2020-11-02
[Backport]2.4: Fix return err when secret webhook-receiver is not found

GGGitBoy:
**Problem:**

The user has not configured an alert that requires webhook-receiver, but will still receive the error log `secret webhook-receiver is not found`.

**Solution:**

Execute `syncWebhookConfig` only when `webhook-receiver.enabled` is `true`

**Relate issue:**

https://github.com/rancher/rancher/issues/28954


guangbochen (approved):
LGTM

aiyengar2 (approved):
LGTM

dramich (approved):

maggieliu:
https://github.com/rancher/rancher/issues/30088

Issue #23715, opened 2019-10-28
[Feature Request] Support Load-Balancer-SKU selection in AKS cluster creation


pengjiang80:
AKS now supports two SKU of LBs in AKS cluster creation: basic and standard. More information can be found at https://docs.microsoft.com/en-us/azure/aks/load-balancer-standard 

The option is available in az cli cluster creation like below with --load-balancer-sku parameter

az aks create \
    --resource-group myResourceGroup \
    --name myAKSCluster \
    --node-count 1 \
    --load-balancer-sku standard \

Note: Not sure the Azure UI in global Azure, but it seems it's not available in Azure China UI yet.

We may need to add an option to let user choose basic or standard for load-balancer-sku in rancher cloud driver UI.

---
| Useful | Info |
| :-- | :-- |
|Versions|Rancher `v2.3.1` UI: `v2.3.18` |
|Route|`global-admin.clusters.index`|

gzrancher/rancher#13068

mscbpi:
Any update on that ? Azure official documentation will suggest SKU Standard is in use by default, but Rancher creates cluster with Basic LB.

then if you follow the doc for [deploying the nginx ingress](https://docs.microsoft.com/en-us/azure/aks/ingress-static-ip) you will fail because it creates a Public IP with Standard SKU and both SKU must match. 

Alternative is to create a Basic SKU IP Address with fewer feature (as well as fewer feature / lower SLA in the Basic LB)

junkiebev:
I have ticket #13068 in for this - it's a fairly pressing need.

cbron:
@sowmyav27 when we test this we want to double check that existing clusters from earlier versions of rancher upgrade without issue or change. So if a AKS cluster is provisioned with Rancher 2.4.5 then upgraded to this, and then an upgrade occurs on the cluster, that the LB stays the same and cluster continues to work, feel free to reach out to @nickgerace with any questions.

sowmyav27:
**Validated on 2.4-head commit id: `a61621e57`**

- Deploy an AKS cluster - set `loadBalancerSku` to Basic
<img width="822" alt="Screen Shot 2020-11-03 at 2 34 57 PM" src="https://user-images.githubusercontent.com/26032343/98053176-d6503480-1dec-11eb-83d9-a3c61d3012d6.png">

- When the cluster comes up, verified in the backend LB is set as basic.
```
Traffic routing
Load balancer : Basic
```

junkiebev:
[...]

PR #29901, opened 2020-11-02
Add LoadBalancerSKU support to AKS creation

nickgerace:
Add LoadBalancerSKU support to AKS creation. We default to "Standard" by
default. Add unit test for the new functionality. Update existing code
to work with new Azure SDK module. Deprecate AgentStorageProfile.

Issue: #23715

```bash
[nickgerace at rancherbook in ~/git/rancher/pkg/kontainer-engine/drivers/aks] (0) (issue-23715)
% go test -v -run TestLoadBalancerSKU
=== RUN   TestLoadBalancerSKU
--- PASS: TestLoadBalancerSKU (0.00s)
PASS
ok  	github.com/rancher/rancher/pkg/kontainer-engine/drivers/aks	0.300s
```

Continuing work from @GGGitBoy: https://github.com/rancher/kontainer-engine/pull/201

### DIFF
--- a/pkg/kontainer-engine/drivers/aks/aks_driver.go
+++ b/pkg/kontainer-engine/drivers/aks/aks_driver.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2018-03-31/containerservice"
+	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2019-10-01/containerservice"
 	"github.com/Azure/azure-sdk-for-go/services/preview/operationalinsights/mgmt/2015-11-01-preview/operationalinsights"
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2017-05-10/resources"
 	"github.com/Azure/go-autorest/autorest"
@@ -76,10 +76,10 @@ type state struct {
 	AgentName string `json:"agentPoolName,omitempty"`
 	// AgentOsdiskSizeGB specifies the disk size for every machine in the agent pool. [optional only when creating]
 	AgentOsdiskSizeGB int64 `json:"agentOsdiskSize,omitempty"`
-	// AgentStorageProfile specifies the storage profile in the agent pool. [optional only when creating]
-	AgentStorageProfile string `json:"agentStorageProfile,omitempty"`
 	// AgentVMSize specifies the VM size in the agent pool. [optional only when creating]
 	AgentVMSize string `json:"agentVmSize,omitempty"`
+	// LoadBalancerSku specifies the LoadBalancer SKU of the cluster. [optional only when creating]
+	LoadBalancerSku string `json:"loadBalancerSku,omitempty"`
 	// VirtualNetworkResourceGroup specifies the Azure Virtual Network located int which resource group. Composite of agent virtual network subnet ID. [optional only when creating]
 	VirtualNetworkResourceGroup string `json:"virtualNetworkResourceGroup,omitempty"`
 	// VirtualNetwork specifies an existing Azure Virtual Network. Composite of agent virtual network subnet ID. [optional only when creating]
@@ -235,15 +235,15 @@ func (d *Driver) GetDriverCreateOptions(ctx context.Context) (*types.DriverFlags
 		Usage: `GB size to be used to specify the disk for every machine in the agent pool. If you specify 0, it will apply the default according to the "agent vm size" specified.`,
 	}
 
-	driverFlag.Options["agent-storage-profile"] = &types.Flag{
-		Type:  types.StringType,
-		Usage: fmt.Sprintf("Storage profile specifies what kind of storage used on machine in the agent pool. Chooses from %v.", containerservice.PossibleStorageProfileTypesValues()),
-		Value: string(containerservice.ManagedDisks),
-	}
 	driverFlag.Options["agent-vm-size"] = &types.Flag{
 		Type:  types.StringType,
 		Usage: "Size of machine in the agent pool.",
-		Value: string(containerservice.StandardD1V2),
+		Value: string(containerservice.VMSizeTypesStandardD1V2),
+	}
+	driverFlag.Options["load-balancer-sku"] = &types.Flag{
+		Type:  types.StringType,
+		Usage: "The LoadBalancer SKU of the cluster.",
+		Value: string(containerservice.Standard),
 	}
 	driverFlag.Options["virtual-network-resource-group"] = &types.Flag{
 		Type:  types.StringType,
@@ -429,7 +429,6 @@ func getStateFromOptions(driverOptions *types.DriverOptions) (state, error) {
 	state.AgentMaxPods = options.GetValueFromDriverOptions(driverOptions, types.IntType, "max-pods", "maxPods").(int64)
 	state.AgentName = options.GetValueFromDriverOptions(driverOptions, types.StringType, "agent-pool-name", "agentPoolName").(string)
 	state.AgentOsdiskSizeGB = options.GetValueFromDriverOptions(driverOptions, types.IntType, "agent-osdisk-size", "agentOsdiskSize", "os-disk-size", "osDiskSizeGb").(int64)
-	state.AgentStorageProfile = options.GetValueFromDriverOptions(driverOptions, types.StringType, "agent-storage-profile", "agentStorageProfile").(string)
 	state.AgentVMSize = options.GetValueFromDriverOptions(driverOptions, types.StringType, "agent-vm-size", "agentVmSize").(string)
 	state.VirtualNetworkResourceGroup = options.GetValueFromDriverOptions(driverOptions, types.StringType, "virtual-network-resource-group", "virtualNetworkResourceGroup").(string)
 	state.VirtualNetwork = options.GetValueFromDriverOptions(driverOptions, types.StringType, "virtual-network", "virtualNetwork").(string)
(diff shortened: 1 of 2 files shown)


rmweir (commented):
Should a user be able to update this?

nickgerace:
Good question @rmweir. No, since current guidance appears to require rebuilding the cluster: https://github.com/MicrosoftDocs/azure-docs/issues/43114

pkg/kontainer-engine/drivers/aks/aks_driver.go
```diff
@@ -796,26 +805,20 @@ func (d *Driver) createOrUpdate(ctx context.Context, options *types.DriverOption
 			osDiskSizeGBPointer = to.Int32Ptr(int32(driverState.AgentOsdiskSizeGB))
 		}
 
-		agentStorageProfile := containerservice.ManagedDisks
-		if driverState.AgentStorageProfile != "" {
-			agentStorageProfile = containerservice.StorageProfileTypes(driverState.AgentStorageProfile)
-		}
-
-		agentVMSize := containerservice.StandardD1V2
```
rmweir, line 804: Why are we no longer setting this? Was it removed from the API?

rmweir (changes requested):

pkg/kontainer-engine/drivers/aks/aks_driver.go
nickgerace: Correct. The type still exists, but it is no longer a field in the `ManagedClusterAgentPoolProfile`.

Reference: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-09-01/containerservice#ManagedClusterAgentPoolProfile

pkg/kontainer-engine/drivers/aks/aks_driver.go
```diff
@@ -753,7 +761,7 @@ func (d *Driver) createOrUpdate(ctx context.Context, options *types.DriverOption
 			driverState.Subnet,
 		))
 
-		networkProfile = &containerservice.NetworkProfile{
+		networkProfile = &containerservice.NetworkProfileType{
```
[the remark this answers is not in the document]
nickgerace: Good catch. Will change

pkg/kontainer-engine/drivers/aks/aks_driver.go
```diff
@@ -243,7 +245,12 @@ func (d *Driver) GetDriverCreateOptions(ctx context.Context) (*types.DriverFlags
 	driverFlag.Options["agent-vm-size"] = &types.Flag{
 		Type:  types.StringType,
 		Usage: "Size of machine in the agent pool.",
-		Value: string(containerservice.StandardD1V2),
+		Value: string(containerservice.VMSizeTypesStandardD1V2),
```
brendarearden, line 248: Curious why this changed as well

nickgerace: API change. They are the same value.

pkg/kontainer-engine/drivers/aks/aks_driver.go
nickgerace: Thinking about this a little more, I'm wondering if we should deprecate `agentStorageProfile` since the API has done so with the `ManagedClusterAgentPoolProfile`. Thoughts @rmweir?

rmweir: @nickgerace if we don't use it for anything else than yes. I think I may have worked on this in the past; users wanted to be able to use managed disk storage. I was told in a kubernetes PR for the same issue that the managed disk was going to become the only option I believe.

rmweir: You'll want to create a UI issue for this so the option can be removed.

brendarearden (dismissed):
Looks good!

pkg/kontainer-engine/drivers/aks/aks_driver.go
nickgerace: Created issue: https://github.com/rancher/rancher/issues/29908

brendarearden (dismissed):

rmweir (approved):
LGTM

go.mod
```diff
@@ -39,7 +39,7 @@ replace (
 )
 
 require (
-	github.com/Azure/azure-sdk-for-go v43.0.0+incompatible
+	github.com/Azure/azure-sdk-for-go v48.0.0+incompatible
```
cbron, line 42: It looks like this was available back in 36.2, is the upgrade to 48.0 necessary ? If not we should revert to 43 being this close to a release.

go.sum
```diff
@@ -30,6 +30,8 @@ github.com/Azure/azure-sdk-for-go v36.1.0+incompatible h1:smHlbChr/JDmsyUqELZXLs
 github.com/Azure/azure-sdk-for-go v36.1.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
 github.com/Azure/azure-sdk-for-go v43.0.0+incompatible h1:/wSNCu0e6EsHFR4Qa3vBEBbicaprEHMyyga9g8RTULI=
 github.com/Azure/azure-sdk-for-go v43.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
```
cbron, line 32: (Ignore if above question changes approach) Did go tidy get run here ? Would have thought 43 goes away.

pkg/kontainer-engine/drivers/aks/aks_driver.go
```diff
@@ -775,6 +773,7 @@ func (d *Driver) createOrUpdate(ctx context.Context, options *types.DriverOption
 		}
 	}
 
+	networkProfile.LoadBalancerSku = containerservice.LoadBalancerSku(driverState.LoadBalancerSku)
```
cbron, line 776: How does this react with older clusters that didn't have one set ? Does it just stay blank and nothing happens.

go.sum
nickgerace: I ran `go mod tidy` before pushing. It does not, but I've seen that before in other libraries using Go modules.

go.mod
nickgerace: It may not be necessary. I'll check.

nickgerace: As far as the `containerservice` module goes, the current version in this file does _not_ have `LoadBalancerSku` available. However, I am unsure how the path `github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2018-03-31/containerservice` relates to `github.com/Azure/azure-sdk-for-go <semver>`. Regardless, the current module does not have the functionality we need.

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2018-03-31/containerservice

pkg/kontainer-engine/drivers/aks/aks_driver.go
nickgerace: I have not tested this on older clusters, but this functionality should only be exposed during cluster creation. AKS only allows the standard LB to be configured at creation time.

nickgerace: I think we could add a check to ensure that the LoadBalancerSku is only being configured during cluster creation.
```go
// Only configure the LoadBalancerSku during cluster creation.
if sendRBAC {
	networkProfile.LoadBalancerSku = containerservice.LoadBalancerSku(driverState.LoadBalancerSku)
}
```

go.mod
cbron: Not sure about that link, but its available in the 36.2 codebase: https://github.com/Azure/azure-sdk-for-go/blob/v36.2.0/profiles/preview/preview/containerservice/mgmt/containerservice/models.go#L46-L50

nickgerace: Confirmed that the YYYY-MM-DD module is a wrapper around a SemVer module for the Azure SDK. While `v43` is technically used, the `2018-03-31` module behaves differently, exposing different parts of the underlying library. We need to update the Azure API as a result.

pkg/kontainer-engine/drivers/aks/aks_driver.go
nickgerace: New idea: we always set the Value to "standard" currently. I can make it so that the default is `Value: ""` , and then we can have logic to set it like so...
```
if driverState.LoadBalancerSku != "" {
    	networkProfile.LoadBalancerSku = containerservice.LoadBalancerSku(driverState.LoadBalancerSku)
}
```

nickgerace: Going with this change for the time being since it strikes a balance between being backwards-compatible and non-intrusive.

go.mod
cbron: Makes sense, although I think we should keep that date a bit lower then. We've seen instability with AKS SDK upgrading to quickly. Lets stick with their original `2019-10-01`

nickgerace:
Confirmed that pre-existing clusters are unaffected since they have the Standard LB set anyway. The option is only exposed during cluster creation as well.

pkg/kontainer-engine/drivers/aks/aks_driver.go
nickgerace: Undid this as it was unnecessary. The pre-existing clusters were unaffected by the new change.

brendarearden (approved):

cbron (approved):
lgtm

binaryPUNCH:
What release of Rancher was this fixed in? We have an AKS cluster created by rancher where the load balancer is basic - super annoying that we can't change it

nickgerace:
Hi @binaryPUNCH. This PR made it into Rancher v2.4.9, and I'm unsure of which v2.5.x release, but I know it is in v2.5.5 at the very least. Our [support matrix](https://rancher.com/support-maintenance-terms/) can help let you know which releases are active and may work best for you. You can also visit our [user slack](https://slack.rancher.io/) if you have more questions. Hope this helped!

cbron:
This was released in 2.5.2.

Issue #29826, opened 2020-10-29
[EKSv2]: Additional security groups should not be automatically generated.


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Description:**
With EKS a security group is created for the cluster by default. You can also provide additional security groups, but the default is still used.

In Rancher 2.5.1, the user is only able to provide an additional security group when they have selected their own VPC. Additionally, they cannot proceed after selecting a VPC without providing a security group. These selections should be decoupled.

**Solution**

* the user should not be required to select an additional security group regardless of networking section
* ~allow users to provide additional security groups, regardless of networking section~ security groups must be created with existing VPC, cannot be added to a different VPC
* Change description in UI to portray that it is additional

aaronRancher:
Tested with with Rancher server `rancher\rancher:master-head` `version b72b399` and `rancher\rancher:master-v2.5-head` `version 71da156`

Select custom VPC and subnets
![Screen Shot 2020-11-13 at 4 50 07 PM](https://user-images.githubusercontent.com/57409052/99132214-a36b2500-25d2-11eb-983a-e64620822276.png)
The user is prompted to select security groups
![Screen Shot 2020-11-13 at 4 50 16 PM](https://user-images.githubusercontent.com/57409052/99132215-a49c5200-25d2-11eb-96d1-0a1abed5186b.png)
The user cannot proceed until a security group is chosen
![Screen Shot 2020-11-13 at 4 50 36 PM](https://user-images.githubusercontent.com/57409052/99132217-a6661580-25d2-11eb-92e0-e5c0bd7d8487.png)

thedadams:
@aaronRancher I created a UI ticket to make selecting security groups optional. You should be able to test this with the API for now, if you'd like.

sowmyav27:
Waiting for this [issue](https://github.com/rancher/rancher/issues/30083) to be fixed

aaronRancher:
Validated this issue from the UI: https://github.com/rancher/rancher/issues/30083

The enhancement is verified with Rancher server `rancher\rancher:master-head` `version a549b47` and `rancher\rancher:v2.5-head` `version 9ee39b9`

Steps:
Create an EKS cluster
Select a custom VPC and subnets
Do not select a security group
Set node group
Create cluster

Result:
Cluster is provisioned successfully.
[...]

PR #29930, opened 2020-11-04
Make security groups optional when VPC is provided

thedadams:
Previously, if a user provided a VPC ID, they were also required to
provide at least one security group. However, this requirement is
extraneous. So a user is able to provide a VPC ID and not provide any
security groups.

Conversely, if a user does provide security groups, then those security
groups are already attached to a VPC. Therefore, the VPC should be
provided as well.

Issue:
https://github.com/rancher/rancher/issues/29826

EKS Operator PR:
https://github.com/rancher/eks-operator/pull/21


rmweir (approved):
LGTM

aiyengar2 (approved):
LGTM

cbron (approved):

Issue #29773, opened 2020-10-27
Add filter option for template GET request


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Description:**
When making a request to get a template, links are added to the response for version that are compatible with server version. This request is for being able to accept a query param for cluster kubernetes version so that only version links compatible with that cluster, on top of being compatible with rancher version, will be included. To achieve this, the new ValidateChartCompatibility can be used in the extractVersionLinks method of the template store.

jiaqiluo:
This is enhancement is validated on 
- rancher:master-head `203033a` 
- rancher:v2.5-head `f671292`


Steps:
- add the following repo which charts contains the `kebeVersion` filed
```
URL: https://github.com/izaac/system-charts.git
branch: k8stest
```
- open the url  `https://server-url/v3/templateversions` in the browser 

Resutls:
- `kubeVersion` is an option for filtering 
<img width="345" alt="Screen Shot 2020-11-20 at 2 27 07 PM" src="https://user-images.githubusercontent.com/6218999/99852184-b4211b00-2b3d-11eb-8891-53c997382aec.png">

- We can use the filter to find all `templateversions` which `kubeVersion` is '>1.18.0'
 
![screencapture-178-128-8-172-v3-templateversions-2020-11-20-14_38_25](https://user-images.githubusercontent.com/6218999/99852362-17ab4880-2b3e-11eb-993d-2333fa4dfd09.png)

jiaqiluo:
Reopen this issue because all charts with the field `kubeVersion` in `Chart.yaml` are now not returned from the backend. Also, we are not passing the ClusterName when getting the version links of a template.

The following is done on rancher:master-head 6f96683ba

**Steps:**
- run rancher with passing the parameter `CATTLE_SERVER_VERSION`
```
sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 -e CATTLE_SERVER_VERSION=2.5.3 --privileged rancher/rancher:master-head
```
- add clusters with different k8s versions 

![image](https://user-images.githubusercontent.com/6218999/101083128-c18ccb00-3569-11eb-9c8f-07b7fc5986e1.png)

- add the following repo as a global repo 
```
repo:     https://github.com/jiaqiluo/system-charts
branch: k8s-limit
---
for 0.1.5 I set: kubeVersion: ">=1.16.0 <1.18.10" in Chart.yaml
[...]

PR #29933, opened 2020-11-04
Add clusterName filter to templates GET and tests

rmweir:
**Problem:**
Version displayed by dropdown in UI do not filter out template versions that are incompatible with targeted cluster.

**Solution:**
The templates API is not cluster scoped so a query param can be optionally passed to indicate which cluster is being targeted.

**Issue:**
https://github.com/rancher/rancher/issues/29773


brendarearden (approved):
LGT💵

aiyengar2 (approved):
LGTM

cbron (approved):
One question, want to make sure we standardize there. Otherwise lgtm.

PR #29935, opened 2020-11-04
Set LoadBalancerSKU to empty string for AKS

nickgerace:
Set LoadBalancerSKU to empty string for AKS. This allows for the field
to be omitted for older AKS clusters.

Original Issue: https://github.com/rancher/rancher/issues/23715


rmweir (approved):
LGTM. I think this is fine in favor of time, but my hunch is that we do not need to remove the default to make this work. I think that would be good because these defaults get added to the schema which could help UI.

cbron (approved):

PR #29946, opened 2020-11-05
[2.4] Ignore Accessor error in role index

dramich:
Problem:
If an object does not implement the correct interface the binding
indexer will error and no rbac changes will be processed

Solution:
Dont return the error because the indexer only gives back valid objects
anyways so this will allow rbac changes to continue despite an invalid
object in the cache

https://github.com/rancher/rancher/issues/29947

### DIFF
--- a/pkg/controllers/management/auth/indexes.go
+++ b/pkg/controllers/management/auth/indexes.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/rbac/v1"
 	meta2 "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/auth/indexes.go
```diff
@@ -39,14 +40,21 @@ func rbRoleSubjectKeys(roleName string, subjects []v1.Subject) []string {
...
+	meta, err := meta2.Accessor(ro)
 	if err != nil {
-		return nil, err
+		var gvk string
+		if g := ro.GetObjectKind().GroupVersionKind().String(); g != "" {
+			gvk = g
+		} else {
+			gvk = "no GVK available"
+		}
+		logrus.Warnf("[indexByMembershipBindingOwner] unexpected object type: %s, err: %v", gvk, err.Error())
```
ibuildthecloud, line 56: Do you want to log the type itself?  I think that's `%T`

ibuildthecloud (approved):

PR #30014, opened 2020-11-10
Update tags and README

mrajashree:


package/Dockerfile
```diff
@@ -13,7 +13,7 @@ WORKDIR /var/lib/rancher
 
 ARG ARCH=amd64
 ARG IMAGE_REPO=rancher
-ARG SYSTEM_CHART_DEFAULT_BRANCH=release-v2.5
+ARG SYSTEM_CHART_DEFAULT_BRANCH=dev-v2.5
```
cbron, line 16: This needs to be done in a few other places like scripts/package, maybe another place.

cbron (approved):
Assuming charts and KDM were always dev on here, lgtm

PR #30061, opened 2020-11-13
fix the bug that test fails because 'NoneType' object has no attribute 'rstrip'

jiaqiluo:
1. fix the bug that test fails because 'NoneType' object has no attribute 'rstrip'
The error in logs:
```
>           result = result.rstrip()
E           AttributeError: 'NoneType' object has no attribute 'rstrip'

tests/v3_api/common.py:894: AttributeError
```

2. print the full RANCHER_SERVER_CMD

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -891,10 +891,11 @@ def validate_http_response(cmd, target_name_list, client_pod=None,
                 wget_cmd = "wget -qO- " + cmd
             result = kubectl_pod_exec(client_pod, wget_cmd)
             result = result.decode()
-        result = result.rstrip()
-        assert result in target_name_list
-        if result in target_hit_list:
-            target_hit_list.remove(result)
+        if result is not None:
+            result = result.rstrip()
+            assert result in target_name_list
+            if result in target_hit_list:
+                target_hit_list.remove(result)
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/common.py
```diff
@@ -891,10 +891,11 @@ def validate_http_response(cmd, target_name_list, client_pod=None,
                 wget_cmd = "wget -qO- " + cmd
             result = kubectl_pod_exec(client_pod, wget_cmd)
             result = result.decode()
-        result = result.rstrip()
-        assert result in target_name_list
-        if result in target_hit_list:
-            target_hit_list.remove(result)
+        if result is not None:
```
sowmyav27, line 894: What when the result is None? Can we print something, because this happens a lot of time (not sure why). Do we know why the result is None sometimes?

sowmyav27 (changes requested):

tests/validation/tests/v3_api/common.py
jiaqiluo: From my observation, it happens when the curl command is sent right after the URL is accessible but the pod is not ready to respond the request.

aaronRancher (approved):
LGTM

sowmyav27 (approved):
LGTM

Issue #29907, opened 2020-11-02
[EKSv2]: Rancher should be able to fully communicate with private EKS clusters


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Steps to reproduce (least amount of steps as possible):**
1. Create or register private EKS cluster (a cluster that only has Private Access enabled).

**Result:**
Error: "Failed to communicate with cluster.
Even if a cluster is created as Public/Private then has public disabled, the agent will connect but health checks will error.

**Desire:**
Registering a private EKS cluster should work and rancher should be able to communicate with it.
Creating a private EKS cluster should work. This flow will still require the user to use a command to deploy the agent, similar to regular imported clusters, described here:  https://github.com/rancher/rancher/issues/28356.

**Notes:**
It is possible to import a private cluster through regular import. This is because imported clusters tunnel any communication between rancher and the downstream server through the agent and communicate with the kubernetes service IP (which is private) instead of the public endpoint. The solution for this should probably use this same approach for EKS clusters. Requests should be tunneled and instead of using the public Kube API endpoint, the kubernetes service ip should be used. Part of this is handled in authorizeCluster function located in pkg/tunnelserver/tunnel.go.

This is blocking: https://github.com/rancher/rancher/issues/28356

sowmyav27:
**On 2.5-head - commit id: `395b4fe77`**

Validated the following use cases:

**Use case#1**

- Deploy an EKS cluster with only Private access enabled.
- The cluster comes to Waiting to register state
- Import the cluster using import command from Rancher UI ( rancher server is on the same subnet/security group as the EKS cluster).
- Eks cluster come sup Active.

**Usecase#2**

- Edit the above created cluster.
- Add a nodegroup
- Save changes made.
- Cluster goes into Updating state and nodes get added and the cluster comes up Active.

**Usecase#3**

- Edit the above cluster. Update the k8s version. Save the changes made.
- Cluster goes into Updating state

aaronRancher:
This enhancement is verified with Rancher server `rancher\rancher:master-head` `version e129f92` and `rancher\rancher:master-head` `version 6ad4dce72`

**Hosted EKS cases:**
Use case#1
- Deploy an EKS cluster with only Private access enabled.
- The cluster comes to Waiting to register state
- Import the cluster using import command from Rancher UI ( rancher server is on the same subnet/security group as the EKS cluster).
- Eks cluster come sup Active.

Use case#2
[...]

PR #30120, opened 2020-11-17
Allow communication with EKS clusters with private API endpoint

thedadams:
Rancher is able to import and create clusters with private API endpoint, but is not able to communicate with them afterwards. Now, as long as the agent is deployed, rancher is able to tunnel to communicate with the API.

Issue:
https://github.com/rancher/rancher/issues/29907

### DIFF
--- a/pkg/clusterrouter/proxy/proxy_server.go
+++ b/pkg/clusterrouter/proxy/proxy_server.go
@@ -11,10 +11,9 @@ import (
 	"strings"
 	"sync"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
 	"github.com/rancher/norman/httperror"
-	factory "github.com/rancher/rancher/pkg/dialer"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	dialer2 "github.com/rancher/rancher/pkg/dialer"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config/dialer"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -173,7 +172,7 @@ func (r *RemoteService) getTransport() (http.RoundTripper, error) {
 			return nil, err
 		}
 		transport.DialContext = d
-		if factory.IsCloudDriver(newCluster) {
+		if dialer2.IsPublicCloudDriver(newCluster) {
 			transport.Proxy = http.ProxyFromEnvironment
 		}
 	}
--- a/pkg/controllers/management/clusterdeploy/clusterdeploy.go
+++ b/pkg/controllers/management/clusterdeploy/clusterdeploy.go
@@ -10,12 +10,12 @@ import (
 	"sync"
 	"time"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/types"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	util "github.com/rancher/rancher/pkg/cluster"
 	"github.com/rancher/rancher/pkg/clustermanager"
+	"github.com/rancher/rancher/pkg/dialer"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/image"
 	"github.com/rancher/rancher/pkg/kubectl"
@@ -219,6 +219,11 @@ func getDesiredImage(cluster *v3.Cluster) string {
 }
 
 func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
+	if dialer.HasPublicAPIEndpoint(cluster) {
+		logrus.Debugf("clusterDeploy: deployAgent: cluster [%s] is private so agent cannot be deployed automatically", cluster.Name)
+		return nil
+	}
+
 	if cluster.Spec.Internal {
 		return nil
 	}
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rancher/rancher/pkg/catalog/manager"
 	"github.com/rancher/rancher/pkg/controllers/management/eksupstreamrefresh"
 	"github.com/rancher/rancher/pkg/controllers/management/rbac"
+	"github.com/rancher/rancher/pkg/dialer"
 	v3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	corev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
@@ -31,6 +32,7 @@ import (
 	"github.com/rancher/rancher/pkg/ref"
 	"github.com/rancher/rancher/pkg/systemaccount"
 	"github.com/rancher/rancher/pkg/types/config"
+	typesDialer "github.com/rancher/rancher/pkg/types/config/dialer"
 	"github.com/rancher/rancher/pkg/wrangler"
 	wranglerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
@@ -44,6 +46,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/token"
 	"sigs.k8s.io/yaml"
 )
@@ -77,6 +80,7 @@ type eksOperatorController struct {
 	catalogManager       manager.CatalogManager
 	systemAccountManager *systemaccount.Manager
 	dynamicClient        dynamic.NamespaceableResourceInterface
+	clientDialer         typesDialer.Factory
 }
 
 func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.ManagementContext) {
@@ -99,6 +103,7 @@ func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.M
 		catalogManager:       mgmtCtx.CatalogManager,
 		systemAccountManager: systemaccount.NewManager(mgmtCtx),
 		dynamicClient:        eksCCDynamicClient,
+		clientDialer:         mgmtCtx.Dialer,
 	}
 
 	wContext.Mgmt.Cluster().OnChange(ctx, "eks-operator-controller", e.onClusterChange)
@@ -281,6 +286,14 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			cluster, err = e.generateAndSetServiceAccount(cluster)
 			if err != nil {
 				var statusErr error
+				if strings.Contains(err.Error(), fmt.Sprintf(dialer.WaitForAgentError, cluster.Name)) {
+					// In this case, the API endpoint is private and rancher is waiting for the import cluster command to be run.
+					cluster, statusErr = e.setUnknown(cluster, apimgmtv3.ClusterConditionWaiting, "waiting for cluster agent to be deployed")
+					if statusErr == nil {
+						e.clusterEnqueueAfter(cluster.Name, enqueueTime)
+					}
+					return cluster, statusErr
+				}
 				cluster, statusErr = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting,
 					fmt.Sprintf("failed to communicate with cluster: %v", err))
 				if statusErr != nil {
@@ -391,9 +404,8 @@ func (e *eksOperatorController) updateEKSClusterConfig(cluster *mgmtv3.Cluster, 
 	}
 }
 
-// generateAndSetServiceAccount reads the EKSClusterConfig's secret once available and uses its fields to generate a
-// service account token. The token, CA cert, and API endpoint are then copied to the cluster status.
-func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+// getCAAndAPIEndpoint reads the EKSClusterConfig's secret once available. The CA cert and API endpoint are then copied to the cluster status.
+func (e *eksOperatorController) getCAAndAPIEndpoint(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
 	backoff := wait.Backoff{
 		Duration: 2 * time.Second,
 		Factor:   2,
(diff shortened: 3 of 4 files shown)


pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -281,8 +285,14 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			cluster, err = e.generateAndSetServiceAccount(cluster)
 			if err != nil {
 				var statusErr error
-				cluster, statusErr = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting,
-					fmt.Sprintf("failed to communicate with cluster: %v", err))
+				if strings.Contains(err.Error(), fmt.Sprintf(dialer.WaitForAgentError, cluster.Name)) {
+					cluster, statusErr = e.setUnknown(cluster, apimgmtv3.ClusterConditionWaiting, "waiting for cluster agent to be deployed")
+					err = nil
+					e.clusterEnqueueAfter(cluster.Name, enqueueTime)
```
rmweir, line 291: Since an error is being returned we shouldn't need to enqueue manually.

pkg/dialer/factory.go
```diff
@@ -54,15 +58,24 @@ func (f *Factory) ClusterDialer(clusterName string) (dialer.Dialer, error) {
...
+	if cluster.Spec.Internal ||
+		cluster.Status.Driver == "" ||
+		cluster.Status.Driver == v32.ClusterDriverImported ||
+		cluster.Status.Driver == v32.ClusterDriverRKE ||
+		cluster.Status.Driver == v32.ClusterDriverK3s ||
+		cluster.Status.Driver == v32.ClusterDriverK3os ||
+		cluster.Status.Driver == v32.ClusterDriverRke2 ||
+		cluster.Status.Driver == v32.ClusterDriverRancherD {
+		return false
+	}
```
rmweir, line 71: could this not be simplified to cluster.Status.Driver != EKS {return false}, or is it possible the driver is not set yet?

pkg/dialer/factory.go
```diff
@@ -54,15 +58,24 @@ func (f *Factory) ClusterDialer(clusterName string) (dialer.Dialer, error) {
...
+
+	return !IsPrivateCloudDriver(cluster)
+}
+
+func IsPrivateCloudDriver(cluster *v3.Cluster) bool {
```
rmweir, line 76: I think the name is a little confusing on this and IsPublicCloudDriver. A name like IsPrivateCloudCluster, and possibly a comment to explain this refers to a cluster with only private access opposed to one that has at least private access.

rmweir (changes requested):

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -281,8 +285,14 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			cluster, err = e.generateAndSetServiceAccount(cluster)
 			if err != nil {
 				var statusErr error
-				cluster, statusErr = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting,
-					fmt.Sprintf("failed to communicate with cluster: %v", err))
+				if strings.Contains(err.Error(), fmt.Sprintf(dialer.WaitForAgentError, cluster.Name)) {
+					cluster, statusErr = e.setUnknown(cluster, apimgmtv3.ClusterConditionWaiting, "waiting for cluster agent to be deployed")
+					err = nil
```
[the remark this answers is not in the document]
thedadams: I did that to avoid code duplication. Will refactor.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: The error shouldn't be returned in this case. Will refactor to make it more obvious.

pkg/dialer/factory.go
thedadams: The original was written this way. This is to protect against new drivers being added (like GKE or AKS). That is, if a new cloud driver is added, then we don't have to come here to alter this code as well.

pkg/dialer/factory.go
thedadams: Will update.

pkg/dialer/factory.go
rmweir: Maybe we should keep a slice of included or excluded drivers and just have a function that iterates through them, we could put it in a package for the cloud providers.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -281,8 +285,15 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
-					fmt.Sprintf("failed to communicate with cluster: %v", err))
+				if strings.Contains(err.Error(), fmt.Sprintf(dialer.WaitForAgentError, cluster.Name)) {
+					// In this case, the API endpoint is private and rancher is waiting for the import cluster command to be run.
+					cluster, statusErr = e.setUnknown(cluster, apimgmtv3.ClusterConditionWaiting, "waiting for cluster agent to be deployed")
+					if statusErr == nil {
+						e.clusterEnqueueAfter(cluster.Name, enqueueTime)
+					}
+					return cluster, statusErr
+				}
+				cluster, statusErr = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting, fmt.Sprintf("failed to communicate with cluster: %v", err))
```
cbron, line 296: Nit: Since this line didn't actually change we could leave as-is.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -281,8 +285,15 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			cluster, err = e.generateAndSetServiceAccount(cluster)
 			if err != nil {
 				var statusErr error
-				cluster, statusErr = e.setFalse(cluster, apimgmtv3.ClusterConditionWaiting,
-					fmt.Sprintf("failed to communicate with cluster: %v", err))
+				if strings.Contains(err.Error(), fmt.Sprintf(dialer.WaitForAgentError, cluster.Name)) {
+					// In this case, the API endpoint is private and rancher is waiting for the import cluster command to be run.
+					cluster, statusErr = e.setUnknown(cluster, apimgmtv3.ClusterConditionWaiting, "waiting for cluster agent to be deployed")
```
cbron, line 290: Unknown because its still "pending" ?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -419,22 +429,33 @@ func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Clu
...
 	}
 
-	logrus.Infof("generating service account token for cluster [%s]", cluster.Name)
+	cluster = cluster.DeepCopy()
+	cluster.Status.APIEndpoint = string(caSecret.Data["endpoint"])
+	cluster.Status.CACert = string(caSecret.Data["ca"])
+	return e.clusterClient.Update(cluster)
+}
+
+// generateAndSetServiceAccount users the API endpoint and CA cert to generate a service account token. The token is then copied to the cluster status.
```
cbron, line 438:
```suggestion
// generateAndSetServiceAccount uses the API endpoint and CA cert to generate a service account token. The token is then copied to the cluster status.
```

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -391,9 +402,8 @@ func (e *eksOperatorController) updateEKSClusterConfig(cluster *mgmtv3.Cluster,
 	}
 }
 
-// generateAndSetServiceAccount reads the EKSClusterConfig's secret once available and uses its fields to generate a
-// service account token. The token, CA cert, and API endpoint are then copied to the cluster status.
-func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+// getCAAndAPIEndpoint reads the EKSClusterConfig's secret once available. The CA cert and API endpoint are then copied to the cluster status.
+func (e *eksOperatorController) getCAAndAPIEndpoint(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
```
cbron, line 406: Whats the reason for splitting them up, just readability ? Seems like they could all be inline.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -112,6 +112,11 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
 		return nil
 	}
 
+	if !clouddrivers.HasPublicAPIEndpoint(cluster) {
+		logrus.Tracef("clusterDeploy: doSync: cluster [%s] is private so agent cannot be deployed automatically", cluster.Name)
```
cbron, line 116: Probably info or debug here, not trace.

thedadams: All of the other logs in this method are Trace. Would you still suggest info or debug?

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Yes.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: We can get the `APIEndpoint` and `CACert` before the cluster agent is deployed. Since we can get those things, we need to make sure that they are added to Status (which is what happens at the end of this method). It simplified some error checking and made sure the items were set in Status in case a later error was produced.

pkg/dialer/factory.go
```diff
@@ -54,17 +58,6 @@ func (f *Factory) ClusterDialer(clusterName string) (dialer.Dialer, error) {
...
 
-func IsCloudDriver(cluster *v3.Cluster) bool {
-	return !cluster.Spec.Internal &&
-		cluster.Status.Driver != "" &&
-		cluster.Status.Driver != v32.ClusterDriverImported &&
-		cluster.Status.Driver != v32.ClusterDriverRKE &&
-		cluster.Status.Driver != v32.ClusterDriverK3s &&
-		cluster.Status.Driver != v32.ClusterDriverK3os &&
-		cluster.Status.Driver != v32.ClusterDriverRke2 &&
-		cluster.Status.Driver != v32.ClusterDriverRancherD
```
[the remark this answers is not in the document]
thedadams: The only "Driver" from `v32` (here) that isn't checked is EKS. This seems to protect against other cloud drivers being added (GKE/AKS) and not being added in this `dialer` package. Admittedly, there might be other reasons, but this seems like the most obvious reason.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
dramich: I think debug is fine

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -112,6 +112,11 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
 		return nil
 	}
 
+	if !clouddrivers.HasPublicAPIEndpoint(cluster) {
```
dramich, line 115: This should move inside `deployAgent`

pkg/controllers/management/drivers/cloud/cloud.go
```diff
@@ -0,0 +1,23 @@
+package clouddrivers
```
dramich, line 1: Move this back, create a new helper that wraps the current `IsCloudDriver ` and adds a case for EKS publicaccess checking

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -419,22 +430,33 @@ func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Clu
 		return cluster, fmt.Errorf("failed waiting for cluster [%s] secret: %s", cluster.Name, err)
 	}
 
-	logrus.Infof("generating service account token for cluster [%s]", cluster.Name)
+	cluster = cluster.DeepCopy()
+	cluster.Status.APIEndpoint = string(caSecret.Data["endpoint"])
+	cluster.Status.CACert = string(caSecret.Data["ca"])
+	return e.clusterClient.Update(cluster)
```
dramich, line 436: Use retryOnConflict here

pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -11,10 +11,9 @@ import (
 	"strings"
 	"sync"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
 	"github.com/rancher/norman/httperror"
-	factory "github.com/rancher/rancher/pkg/dialer"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	dialer2 "github.com/rancher/rancher/pkg/dialer"
```
cbron, line 16: Is this changing because we use the factory var in other places ? Like arg to NewRemote

cbron (commented):
Seems fine to me now, would like @dramich approval first.

pkg/clusterrouter/proxy/proxy_server.go
thedadams: No, I'm afraid the reason is much simpler than that: that is what Goland chose when I moved everything back.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Reverted.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Fixed.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
thedadams: Done.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
thedadams: Done.

pkg/clusterrouter/proxy/proxy_server.go
dramich: It's because `	"github.com/rancher/rancher/pkg/types/config/dialer"` is already imported

pkg/controllers/management/drivers/cloud/cloud.go
thedadams: Moved back to `pkg/dialer/factory.go`.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Done.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -219,6 +219,11 @@ func getDesiredImage(cluster *v3.Cluster) string {
 }
 
 func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
+	if dialer.HasPublicAPIEndpoint(cluster) {
+		logrus.Debugf("clusterDeploy: doSync: cluster [%s] is private so agent cannot be deployed automatically", cluster.Name)
```
dramich, line 223:
```suggestion
		logrus.Debugf("clusterDeploy: deployAgent: cluster [%s] is private so agent cannot be deployed automatically", cluster.Name)
```

thedadams: Done.

dramich (approved):

Issue #29318, opened 2020-10-01
Alerts - rule deployer errors when enabling alerts


izaac:
**What kind of request is this:**
bug

**Steps to reproduce (least amount of steps as possible):**

1. Add a notifier
2. Add a new Alert group and add a new alert of type node to it (i used 1% CPU use)
3. Select the notifier
4. After a while I started receiving alerts properly but error in logs.

**Result:**

- Rancher logs show these alerts deployer alert when alerts are enabled.

```
2020/10/01 22:55:42 [ERROR] error syncing 'c-4m9vp/car-8fvng': handler cluster-alert-rule-deployer: failed to get Alertmanager Deployment information, statefulsets.apps "alertmanager-cluster-alerting" not found, handler cluster-alert-rule-controller: Failed to get service for alertmanager, services "cattle-prometheus/alertmanager-operated" not found, requeuing
2020/10/01 22:56:28 [ERROR] error syncing 'c-4m9vp/car-8fvng': handler cluster-alert-rule-deployer: failed to get Alertmanager Deployment information, statefulsets.apps "alertmanager-cluster-alerting" not found, handler cluster-alert-rule-controller: Failed to get service for alertmanager, services "cattle-prometheus/alertmanager-operated" not found, requeuing
2020/10/01 22:57:55 [ERROR] error syncing 'c-4m9vp/car-8fvng': handler cluster-alert-rule-deployer: failed to get Alertmanager Deployment information, statefulsets.apps "alertmanager-cluster-alerting" not found, handler cluster-alert-rule-controller: Failed to get service for alertmanager, services "cattle-prometheus/alertmanager-operated" not found, requeuing
2020/10/01 23:00:44 [ERROR] error syncing 'c-4m9vp/car-8fvng': handler cluster-alert-rule-deployer: failed to get Alertmanager Deployment information, statefulsets.apps "alertmanager-cluster-alerting" not found, handler cluster-alert-rule-controller: Failed to get service for alertmanager, services "cattle-prometheus/alertmanager-operated" not found, requeuing
2020/10/01 23:04:22 [ERROR] error syncing 'c-4m9vp/car-8fvng': handler cluster-alert-rule-deployer: failed to get Alertmanager Deployment information, statefulsets.apps "alertmanager-cluster-alerting" not found, handler cluster-alert-rule-controller: Failed to get service for alertmanager, services "cattle-prometheus/alertmanager-operated" not found, requeuing
```

**Environment information**
[...]

PR #30124, opened 2020-11-18
Fix missing handle error cause prometheus-operator dosen't deploy

aiwantaozi:
Problem:
We used to just log error when failed to deploy/withdraw systemMonitor, if it failed to deploy prometheus-operator, no error return cause operator isn't get deployed until next time cluster sync.

Solution:
Return error when failed to deploy/withdraw systemMonitor

Issue:
https://github.com/rancher/rancher/issues/29318

### DIFF
--- a/pkg/controllers/managementagent/monitoring/operatorHandler.go
+++ b/pkg/controllers/managementagent/monitoring/operatorHandler.go
@@ -5,21 +5,18 @@ import (
 	"reflect"
 	"strings"
 
-	app2 "github.com/rancher/rancher/pkg/app"
-
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v33 "github.com/rancher/rancher/pkg/apis/project.cattle.io/v3"
-
-	"k8s.io/apimachinery/pkg/labels"
-
-	"github.com/pkg/errors"
+	app2 "github.com/rancher/rancher/pkg/app"
 	"github.com/rancher/rancher/pkg/catalog/manager"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	projectv3 "github.com/rancher/rancher/pkg/generated/norman/project.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/monitoring"
-	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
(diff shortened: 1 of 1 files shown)


orangedeng (approved):
LGTM

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +40,37 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if updateErr != nil {
+		return nil, updateErr
+	}
+
+	return cpy, nil
```
aiyengar2, line 73: Instead of doing the error checking at the end, we should check and return the error when the error is created (e.g. replace the `logrus.Warnf(args)` with `return nil, fmt.Errorf(args)`).

The framework for controllers should automatically take care of requeuing the cluster and triggering the handlers (including this one) when an error is returned.

aiyengar2 (changes requested):

pkg/controllers/managementagent/monitoring/operatorHandler.go
aiwantaozi: I think the old code doesn't return err directly because it needs to update the error reason and status to the cluster object, so I check error after update cluster.

aiyengar2: The code you linked seems to be using `_, updateErr = cd.clusters.Update(cluster)` since it wants to return the error created by `Update` at the end. However, here it seems like the `updateErr` is being entirely discarded either way (since you call `return cpy, nil` right after the `updateErr` ever gets a non-nil value).

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +40,37 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
 
-	var newCluster *mgmtv3.Cluster
-	var err error
+	cpy := obj.DeepCopy()
+	var err, updateErr error
 	//should deploy
 	if obj.Spec.EnableClusterAlerting || obj.Spec.EnableClusterMonitoring {
-		newObj, err := v32.ClusterConditionPrometheusOperatorDeployed.Do(obj, func() (runtime.Object, error) {
-			cpy := obj.DeepCopy()
+		newObj, err := v32.ClusterConditionPrometheusOperatorDeployed.Do(cpy, func() (runtime.Object, error) {
```
dramich, line 47: This `err` shadows the one created above so the if check below will never get an error from here. This can be fixed by either calling this error something else and setting the outer scope to this one or declaring `newObj` at the outer scope so the `:=` can be dropped

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +40,37 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
 			return cpy, deploySystemMonitor(cpy, h.app, h.catalogManager, h.clusterName)
 		})
 		if err != nil {
 			logrus.Warnf("deploy prometheus operator error, %v", err)
```
dramich, line 51: We should not be logging this or the one below, the contents of the log is what should be returned at the end in the error.

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +40,37 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
 		}
 	}
 
-	if newCluster != nil && !reflect.DeepEqual(newCluster, obj) {
-		if newCluster, err = h.clusters.Update(newCluster); err != nil {
-			return nil, err
-		}
-		return newCluster, nil
+	if !reflect.DeepEqual(cpy, obj) {
+		cpy, updateErr = h.clusters.Update(cpy)
```
dramich, line 61: Use `RetryOnConflict` on here https://github.com/rancher/rancher/blob/master/pkg/controllers/management/auth/crtb_handler.go#L225
Then if it still fails use `multierror.Append` to return both errors.

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +40,37 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
+		return cpy, nil
```
dramich, line 62: The update error is never checked so this will always return.

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +40,37 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
 	}
-	return obj, nil
+
+	if err != nil {
+		return nil, err
+	}
+
+	if updateErr != nil {
+		return nil, updateErr
+	}
```
dramich, line 71: Move up to the call, see previous comment about using multierror there

dramich (changes requested):

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -40,32 +42,31 @@ func (h *operatorHandler) syncCluster(key string, obj *mgmtv3.Cluster) (runtime.
...
-		}
+		returnErr = withdrawSystemMonitor(cpy, h.app)
 	}
 
-	if newCluster != nil && !reflect.DeepEqual(newCluster, obj) {
-		if newCluster, err = h.clusters.Update(newCluster); err != nil {
-			return nil, err
+	if !reflect.DeepEqual(cpy, obj) {
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			_, err := h.clusters.Update(cpy)
```
orangedeng, line 60: Do we really need `RetryOnConflict` to retry updating cluster? I think the `clusters.Update` already include the retry logic.

orangedeng (dismissed):

pkg/controllers/managementagent/monitoring/operatorHandler.go
dramich: @orangedeng There is not a retry built into `clusters.Update`, the framework for controllers handles this by enqueuing the cluster again when an error is returned which causes all handles to run on the cluster again. 

Because of all the changes being done to the cluster object inside the helper functions I agree we shouldn't use it here.

pkg/controllers/managementagent/monitoring/operatorHandler.go
```diff
@@ -5,21 +5,19 @@ import (
...
 
-	app2 "github.com/rancher/rancher/pkg/app"
-
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v33 "github.com/rancher/rancher/pkg/apis/project.cattle.io/v3"
+	app2 "github.com/rancher/rancher/pkg/app"
 
-	"k8s.io/apimachinery/pkg/labels"
-
+	"github.com/hashicorp/go-multierror"
```
dramich, line 12: Delete the blank line and run goimports to organize the imports.

dramich (commented):
Update the imports formatting and it looks good.

aiyengar2 (approved):
LGTM

dramich (approved):

Issue #30143, opened 2020-11-18
Panic in latest 2.5-head manager.go


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Install singler Rancher docker
- Provision an RKE DO cluster.

**Result:**

Rancher keeps restarting, I can't see the Cluster getting provisioned.

```
E1118 23:10:03.910999       7 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 3089 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x3c27180, 0x7392980)
	/go/pkg/mod/github.com/rancher/apimachinery@v0.19.0-rancher1/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/rancher/apimachinery@v0.19.0-rancher1/pkg/util/runtime/runtime.go:48 +0x82
panic(0x3c27180, 0x7392980)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/rancher/rancher/pkg/catalog/manager.(*Manager).ValidateKubeVersion(0xc001c54000, 0xc005765028, 0xc000d54dc4, 0x7, 0xc00a52f801, 0xc00a4f7800)
	/go/src/github.com/rancher/rancher/pkg/catalog/manager/manager.go:229 +0x1c1
github.com/rancher/rancher/pkg/catalog/manager.(*Manager).ValidateChartCompatibility(0xc001c54000, 0xc005765028, 0xc000d54dc4, 0x7, 0x0, 0x0)
	/go/src/github.com/rancher/rancher/pkg/catalog/manager/manager.go:208 +0x74
github.com/rancher/rancher/pkg/catalog/manager.(*Manager).LatestAvailableTemplateVersion(0xc001c54000, 0xc001db6ec0, 0xc000d54dc4, 0x7, 0x21, 0xc001db6ec0, 0x0)
	/go/src/github.com/rancher/rancher/pkg/catalog/manager/manager.go:288 +0x211
github.com/rancher/rancher/pkg/controllers/managementuser/alert/deployer.(*AlertService).Upgrade(0xc005e76900, 0xc009ed0de0, 0x2e, 0x8, 0xc009f43828, 0x0, 0x0)
[...]

PR #30154, opened 2020-11-19
Add nil check for version on status

rmweir:
**Problem:**
Prior, version could be accessed before it was set. This would occur on clusters that were newly provisioned or in the process of being provisioned and did not have version set yet.

**Solution:**
Added nil check for version on status. Now, if the version is not set an error will be returned so that controllers can retry until version is available.

**Issue:**
https://github.com/rancher/rancher/issues/30143

### DIFF
--- a/pkg/catalog/manager/manager.go
+++ b/pkg/catalog/manager/manager.go
@@ -226,6 +226,10 @@ func (m *Manager) ValidateKubeVersion(template *v3.CatalogTemplateVersion, clust
 		return err
 	}
 
+	if cluster.Status.Version == nil {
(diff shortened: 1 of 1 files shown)


pkg/catalog/manager/manager.go
```diff
@@ -226,6 +226,10 @@ func (m *Manager) ValidateKubeVersion(template *v3.CatalogTemplateVersion, clust
 		return err
 	}
```
cbron, line 227: And `cluster` would never be nil because you'd get a NotFound error above right ?

cbron (approved):

pkg/catalog/manager/manager.go
rmweir: correct

brendarearden (approved):
LGTM

Issue #28391, opened 2020-08-14
FR: GlobalDNS/externaldns Allow setting of '--aws-batch-change-size' via Rancher UI


Rancheroo:
**What kind of request is this (question/bug/enhancement/feature request):**
feature-request

**Steps to reproduce (least amount of steps as possible):**
No need to reproduce the default aws-batch-setting is 1000
![image](https://user-images.githubusercontent.com/56899019/90220300-48972580-de4b-11ea-854b-a644cf9aaf12.png)
![image](https://user-images.githubusercontent.com/56899019/90220318-5056ca00-de4b-11ea-875c-9cf2954bc9c3.png)

**Result:**
However, with large deploys this can cause issues when an UPSERT operation is triggered (this doubles the transactions to surpass 1000)

**Other details that may be helpful:**
Currently, the workaround is to lower this batch size to 500 which will allow for a 500 batch upsert operation.

![image](https://user-images.githubusercontent.com/56899019/90220361-6795b780-de4b-11ea-8c71-1b371e5ebe18.png)

However when any updates to the underlying GlobalDNS app are made the change is reverted.

Having this setting configurable, or even defaulting lower due to the UPSERT situation would be ideal. Definitely prefer allowing us to configure as needed.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.5
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): GlobalDNS
- Machine type (cloud/VM/metal) and specifications (CPU/memory): 
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

gzrancher/rancher#10967

gz#13293

prachidamle:
@maggieliu  @cbron  The PR above will expose the parameter as a chart value and ensure it is always set to a lower value when GlobalDNS is deployed.

 But looks like in this ticket, we want to expose the parameter via Rancher GlobalDNS U and be able to change it while deploying it via Rancher UI? If that is what is desired, we will need to change Rancher API to expose this parameter. 

https://github.com/rancher/system-charts/pull/370#issuecomment-729860601

maggieliu:
@Rancheroo @dkeightley Is the intention to make the parameter a fixed value or something user should be able to configure when deploying the chart?

dkeightley:
[...]

PR #30161, opened 2020-11-20
Adds options to global DNS

cbron:
https://github.com/rancher/rancher/issues/28391


pkg/apis/management.cattle.io/v3/globaldns_types.go
```diff
@@ -64,6 +64,7 @@ type Route53ProviderConfig struct {
 	RoleArn         string `json:"roleArn,omitempty"`
 	Region          string `json:"region" norman:"default=us-east-1"`
 	ZoneType        string `json:"zoneType" norman:"default=public"`
+	BatchChangeSize string `json:"batchChangeSize" norman:"default=100"`
```
prachidamle, line 67: I think instead of adding this specific field only to the config, it would be useful to add a Values map[string]string to catch all such needs of custom external-dns deployment. 
Since we anyway need to change UI to expose this - we might as well expose a key-value text field to user and append them to answers during the chart deployment.

prachidamle: Such request to add some fields has been made before and it requires us to change Rancher API and UI - accepting key-value pairs would solve such usecases.

prachidamle: Also better to add this key-value input for all providers along with route53

cbron: updated

prachidamle (dismissed):
LGTM - we will need a UI ticket for this change

dramich (approved):

Issue #28673, opened 2020-09-01
k8s 1.19 removes basic auth but it's in use by kontainer-engine


daenney:
Kubernetes 1.19 is removing support for Basic Auth. However, it looks as though kontainer-engine is actually using it: https://github.com/rancher/kontainer-engine/blob/65780a8839e471d3842580d5d9075ae9a3b70a02/drivers/gke/gke_driver.go#L901-L909 in the GKE driver. This would appear to mean that it won't be possible to use kontainer-engine to create k8s 1.19+ clusters in GCP, as it'll fail to bootstrap its service account.

E-mail from Google:

> We are writing to let you know that upstream Kubernetes open source has removed support for basic authentication in version 1.19 (v1.19). Once Google Kubernetes Engine (GKE) upgrades your clusters to v1.19, you will no longer be able to use basic authentication. You will still be able to authenticate to GKE using other methods.
>
>What do I need to know?
>
> Kubernetes has announced that basic authentication will no longer be available starting with Kubernetes v1.19. GKE clusters will also no longer support basic authentication as they gradually upgrade to Kubernetes v1.19:
>
>    * Clusters subscribed to the rapid release channel will begin upgrades on or after January 2021
>    * Clusters subscribed to the regular release channel will begin upgrades on or after March 2021.
>    * Clusters subscribed to the stable release channel and non-release-channel clusters will be upgraded later in 2021; a reminder will be sent before these clusters are upgraded to v1.19

hypnotortoise:
This is also impacting GKE cluster edit operations in my case which are stuck in constant reconciling until I re-enable basic authentication for the cluster via GKE UI / gcloud.  

# Reproducible Test Scenario
1. Disable basic-auth as recommened via `gcloud container clusters update CLUSTER_NAME --no-enable-basic-auth`
2. In Rancher UI -> Edit Cluster (using same SA credentials as on initial cluster setup)
3. Trigger some operational change (e.g. update master cluster version)
4. Rancher will trigger the update (verified in GKE UI), but is stuck with failedSpec error-msg:
`Namespaces is forbidden: User "system:anonymous" cannot create resource "namespaces" in API group "" at the cluster scope`
5. Rancher triggers constant reconciling cluster updates which lead to unavailability of the cluster for management
6. Re-enable basic-auth via `gcloud container clusters update CLUSTER_NAME --enable-basic-auth`
7. Reconciliation stops and Rancher will recover to active state

sadiapoddar:
[...]

PR #30174, opened 2020-11-20
Remove basic auth from GKE driver

thedadams:
The basic was used to generate an access token. However, the access
token was also being generated when getting a service client. Instead of
using the basic auth, we now use the access token.

Original PR:
https://github.com/rancher/rancher/pull/29683

kontainer-engine PR:
https://github.com/rancher/kontainer-engine/pull/247

Issue:


cbron (commented):
Looks fine, but can't merge until 2.4.11 goes out.

cbron (approved):

Issue #29162, opened 2020-09-25
[rancherd] URL from reset-admin is always the private ip


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug/enhancement

**Steps to reproduce (least amount of steps as possible):**
1. Bring up a rancherd cluster
2. Run `rancherd reset-admin`

**Result:**
The URL that is printed is always using the Private IP, even if a Public IP is available.

**Other details that may be helpful:**
Ideally this would even use a value given with tls-san if this was present during the cluster install.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.0-rc1
- Installation option (single install/HA): HA

StrongMonkey:
This is designed to fetch external IP first then internal IP, but without setting cloud provider it is hard to get external IP from v1.nodes. Probably won't fix this.

rancher-max:
This works okay if server-url is set to use the load balancer. Also per @StrongMonkey I believe this is fine so we can close this out.

ebauman:
This is going to flummox folks who try and stand this up on instances w/o having prior knowledge of `server-url`. If possible, can we add an env var or argument to the install script to allow for server url override?

rancher-max:
Reopening per comment above to encourage more discussion on this.

izaac:
Rancher version: `v2.5.2` 

Deploying Rancher with `rancherd` on a Linode VM with private IP is working fine.
Deploying Rancher with `rancherd` on an EC2 VM with private IP I am able to reproduce this issue.

StrongMonkey:
I put a PR https://github.com/rancher/rancher/pull/30175 to make it more user friendly to print serverURL.

It will first fetch server-url setting from rancher if it is already been set. Then fallback to rke2 config file in case user already created a fixed address/domain for all the nodes. If both are not found, use externalNodeIP/internalNodeIP. Then if all of these can't detect the server URL it will just print a message saying rancher is listening on http/8080 and https/8443 and it is up to user to find the right way to talk to it(by creating LB/round-robin DNS).
@rancher-max @ebauman Let me know what you think and feel free to comment on the PR.

ebauman:
LGTM thanks for working on this

izaac:
Rancherd version `v2.5.4-rc1` I am still getting the private IP with EC2 instances.
Ubuntu `20.04`
Instance type: `t3a.xlarge`

```
[...]

PR #30175, opened 2020-11-20
Add more serverURL print logic

StrongMonkey:
This commit adds logic to fetch tls-san from rke2 config.yaml and print
it if found. Also give better output when serverURL can't be found.


StrongMonkey:
https://github.com/rancher/rancher/issues/29162

MonzElmasry (approved):

PR #30189, opened 2020-11-21
Automation - stabilize backup/restore recover test

izaac:


jiaqiluo (approved):
LGTM

tests/validation/tests/v3_api/test_bkp_restore_s3_recover.py
```diff
@@ -45,7 +45,6 @@ def test_bkp_restore_s3_recover_validate():
     for etcd_node in etcd_nodes:
         ips_to_remove.append(etcd_node.customConfig['internalAddress'])
         client.delete(etcd_node)
-        wait_for_node_to_be_deleted(client, etcd_node)
```
sowmyav27, line 48: Why are we removing this wait statement ? It takes sometime for the etcd nodes to get deleted from the cluster after selecting the node and clicking on delete.

izaac: This was ok but I observed that sometimes waiting for each etcd to get removed we ended up not being able to delete all three during testing. This was originally added to circumvent a 2.5 specific issue which was fixed in RKE.

Issue #30188, opened 2020-11-21
Enabling monitoring on new cluster gives intermittent 500 errors and object has been modified error


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create Rancher HA setup
2. Create API keys and plug into script below
3. Run the following script (delete cluster and re-run until you hit the problem.
```
export CATTLE_ACCESS_KEY=token-xxxxx
export CATTLE_SECRET_KEY=xxxxxxxxxxxxxxxxxxxxxxx
export HOST=myrancherserver.something.com

JSON="{\"name\":\"test-cluster\",
       \"description\":\"My Test Cluster\",
       \"rancherKubernetesEngineConfig\":{
         \"type\":\"rancherKubernetesEngineConfig\",
         \"version\":\"v1.19.4-rancher1-2\"
       },
       \"type\":\"cluster\"
      }"

echo json is $JSON

RESPONSE=`curl -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" \
-H 'content-type: application/json' \
-H 'Accept: application/json'  \
-X POST \
-d "${JSON}" \
https://${HOST}/v3/clusters`

echo response is $RESPONSE
echo ""

CLUSTER_ID=`echo $RESPONSE | jq -r .id`

echo cluster id is $CLUSTER_ID
echo ""

curl -vu "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" \
-H 'content-type: application/json' \
-H 'Accept: application/json'  \
-X POST \
-d "{\"version\": null}" \
https://${HOST}/v3/clusters/${CLUSTER_ID}?action=enableMonitoring
```

**Result:**
Sometimes this works, get 204 return code. But sometimes get:
```
{"baseType":"error","code":"ServerError","message":"failed to enable monitoring","status":500,"type":"error"}
```
And error in Rancher logs is:
```
2020/11/21 18:36:46 [ERROR] API error response 500 for POST /v3/clusters/c-lw2f9?action=enableMonitoring. Cause: Operation cannot be fulfilled on clusters.management.cattle.io "c-lw2f9": the object has been modified; please apply your changes to the latest version and try again
```

**Other details that may be helpful:**
Should be able to reproduce if you run it 3-4 times. Suspect you need a Rancher HA setup, did not test on single server setup.

**Environment information**
[...]

PR #30239, opened 2020-11-25
Retry on conflict for enable monitoring actions

aiyengar2:
This commit introduces changes to try harder to prevent the following error from appearing:
```
2020/11/25 15:53:58 [ERROR] API error response 500 for POST /v3/clusters/c-l4bdt?action=enableMonitoring. Cause: Operation cannot be fulfilled on clusters.management.cattle.io "c-l4bdt": the object has been modified; please apply your changes to the latest version and try again
```

The root cause of this error is that we do not normally retry on a conflict, so before it was appearing even after spinning up just 4 clusters in succession and trying to immediately enable monitoring.

~Now the error still appears, but it appears extremely infrequently (around 1 in every 12 if you just hammer the API at the same time).~ **(fixed)**

Related Issue: https://github.com/rancher/rancher/issues/30188

This will be backported to 2.4 / 2.5 line after merge.

### DIFF
--- a/pkg/api/norman/customization/cluster/actions_monitoring.go
+++ b/pkg/api/norman/customization/cluster/actions_monitoring.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
@@ -72,10 +73,10 @@ func (a ActionHandler) editMonitoring(actionName string, action *types.Action, a
 		return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
 	}
 
-	cluster = cluster.DeepCopy()
-	cluster.Annotations = monitoring.AppendAppOverwritingAnswers(cluster.Annotations, string(data))
-
-	_, err = a.ClusterClient.Update(cluster)
+	err = updateClusterWithRetryOnConflict(a.ClusterClient, cluster, func(cluster *v3.Cluster) *v3.Cluster {
+		cluster.Annotations = monitoring.AppendAppOverwritingAnswers(cluster.Annotations, string(data))
+		return cluster
+	})
 	if err != nil {
 		return httperror.WrapAPIError(err, httperror.ServerError, "failed to upgrade monitoring")
 	}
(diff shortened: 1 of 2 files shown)


pkg/api/norman/customization/cluster/utils.go
```diff
@@ -151,3 +153,10 @@ func CanCreateRKETemplate(callerID string, subjectAccessReviewClient clientauthv
 	logrus.Debugf("CanCreateRKETemplate: %v", result)
 	return result.Status.Allowed, nil
 }
+
+func updateClusterWithRetryOnConflict(clusterClient v3.ClusterInterface, cluster *v3.Cluster) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
aiyengar2, line 158: @dramich [`retry.DefaultRetry`](https://github.com/kubernetes/client-go/blob/fb61a7c88cb9f599363919a34b7c54a605455ffc/util/retry/util.go#L26-L33) currently tries to retry 5 times with 10ms in between. Should this be modified? Right now, this makes us somewhat more resilient but the error still appears if you hammer the API enough. The Terraform provider will also be modified to do retries based on @rawmind0 's comment [here](https://github.com/rancher/rancher/issues/30188#issuecomment-733211942).

It will only ever need to retry if the underlying object is being modified at the same time.

dramich: The function needs to refresh the cluster, otherwise it's trying with the old version of the object which is why the error still appears.

aiyengar2: good catch. I updated the function to take in an `updateFunc` that allows the caller to provide a function that performs the changes on the latest version of the cluster.

pkg/api/norman/customization/cluster/actions_monitoring.go
```diff
@@ -72,10 +73,11 @@ func (a ActionHandler) editMonitoring(actionName string, action *types.Action, a
 		return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
 	}
 
-	cluster = cluster.DeepCopy()
-	cluster.Annotations = monitoring.AppendAppOverwritingAnswers(cluster.Annotations, string(data))
-
-	_, err = a.ClusterClient.Update(cluster)
+	err = updateClusterWithRetryOnConflict(a.ClusterClient, cluster, func(cluster *v3.Cluster) *v3.Cluster {
+		cluster = cluster.DeepCopy()
```
mrajashree, line 77: I don't think we need to deepcopy objects that are retrieved with client.Get call from etcd/data store.
I think it's only needed for objects from cache

aiyengar2: updated!

mrajashree (approved):

dramich (approved):

PR #30265, opened 2020-11-30
[release/v2.4] revert "set to dev-2.x for RC"

kinarashah:
This reverts commit 1125a1e0c22576fbe4e915a7b7dd0b63a64b9fae.

Not including new changes for emergency release v2.4.11


cbron (approved):
Thought you missed one, but looks like it never got switched back to dev: https://github.com/rancher/rancher/blob/release/v2.4/pkg/settings/setting.go#L182

Issue #26088, opened 2020-03-17
Unable to set helm value/answer to YAML string "true" in editMonitoring


nikkelma:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Enable monitoring on a cluster
2. Call cluster's `editMonitoring` action with all default answers and two extra answers:
  - ~JSON: `"grafana.extraEnv[0].name": "TEST_VAR"`~
  - ~JSON: `"grafana.extraEnv[0].value": "<value>"` where `<value>` is `true` or `\"true\"` or `'true'`~
 - 2020-11-09: JSON: `"grafana.extraVars[0].name": "TEST_VAR"`
 - 2020-11-09: JSON: `"grafana.extraVars[0].value": "<value>"` where `<value>` is `true` or `\"true\"` or `'true'`

**Result:**
- Scenario A, where `<value>` is `true`: grafana-cluster-monitoring is not deployed, as `env.value` expects a string but gets a bool
- Scenario B, where `<value>` is `\"true\"` or `'true'`: grafana-cluster-monitoring is deployed but `env.value` for `env.name: TEST_VAR` is wrapped in single quotes - YAML `'"true"'` or `'''true'''`, respectively

**Other details that may be helpful:**
- rancher/rancher#17346 is similar, a potential workaround (providing `values.yaml` instead of an `answers` map) is not possible as the `monitoringInput` API object does not have this field
- rancher/rancher#17346, https://github.com/rancher/rancher/issues/13158#issuecomment-418910405, rancher/rancher#13158's fix, and rancher/rancher#15043 indicates that some stability may be needed for helm answers in general

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): **v2.3.5**
- Installation option (single install/HA): **HA**

gz#14692

nikkelma:
Potential fixes:
1. Add ValuesYaml from [`App.Spec`](https://github.com/rancher/types/blob/e8e610dd9e0169e93090b9f881f47043738ee3ec/apis/project.cattle.io/v3/app_types.go#L37) to [`MonitoringInput`](https://github.com/rancher/types/blob/37f04dbaa0a5358e6fc4e99d28c220d953bd98ad/apis/management.cattle.io/v3/cluster_types.go#L286-L289) (UI support would be nice, too)
[...]

Issue #22532, opened 2019-08-29
[CLI ]Catalog automatically converts a numbered string into number instead of https://github.com/rancher/rancher/issues/22532keeping it as string


loganhz:
Fix https://github.com/rancher/rancher/issues/13158 in CLI

anupama2501:
Verified on master-head f33188d node driver rke1 cluster with 3 worker nodes, 1 cp and 1 etcd

**Single app:**

- Installed docker-registry single app in the default project by passing the values as follows:

```
./rancher apps install docker-registry --version 1.8.1 --set var=true --set-string varstr="true" --no-prompt
```

- verified the app created in the default project with the values reflected in `answer` and `answerSetString`:

<img width="400" alt="Screen Shot 2021-07-18 at 12 54 21 PM" src="https://user-images.githubusercontent.com/38144301/126080708-9581e95d-68c0-49cb-a827-da9cfe5845dd.png">
 
**Multicluster app:**
Verified by passing values to the multi cluster and the app is installed in local default project and  cluster default project:
```
./rancher multiclusterapps install --no-prompt --set-string c-clj9f:ans-str1=1 --set-string  c-clj9f:p-glql6:ans-str2=2 --set local:p-gvrj9:ans1=1 --set-string local:ans2=2 --target c-clj9f:p-glql6 --target local:p-gvrj9 --version 1.8.1 docker-registry docker-reg
```
cluster-default project
<img width="400" alt="Screen Shot 2021-07-18 at 1 19 42 PM" src="https://user-images.githubusercontent.com/38144301/126081084-4b8b33f5-fb4c-4b1c-b470-f6f54e3df942.png">

local-default project:
<img width="400" alt="Screen Shot 2021-07-18 at 1 21 41 PM" src="https://user-images.githubusercontent.com/38144301/126081142-598a335c-9c68-42ba-8d65-98a8b2561680.png">

anupama2501:
**Upgrade test case:**

- Verified on v2.5.8, create a node driver rke1 3 worker, 1 cp, 1 etcd cluster
- Install docker-registry in the cluster `./rancher apps install docker-registry --version 1.8.1 --set var=true --set-string varstr="true" --no-prompt`
- Upgrade the cluster to master-head f33188d
- Via cli, upgrade the docker-registry and set the value in `--set-string` :
- `./rancher app upgrade --set foo=bar --set-string baz=bunk docker-registry 1.8.1`
- Verified the values are applied in the cluster --> apps --> docker-registry

PR #30298, opened 2020-12-02
Add ability to force an App Answer to string

nikkelma:
Add field to AppSpec allowing forcing answers to use the equivalent of `--set-string` in the Helm CLI. No breaking API changes, all changed behaviors are opt-in.

Additional commit removes references to norman's deprecated App type, replacing with proper v3 App type.

Linked issue #26088.

### DIFF
--- a/pkg/controllers/managementuserlegacy/monitoring/clusterHandler.go
+++ b/pkg/controllers/managementuserlegacy/monitoring/clusterHandler.go
@@ -103,7 +103,7 @@ func (ch *clusterHandler) doSync(cluster *mgmtv3.Cluster) error {
 			}
 		}
 
-		_, err = ch.deployApp(appName, appTargetNamespace, appProjectName, cluster, etcdTLSConfigs, systemComponentMap)
+		_, _, err = ch.deployApp(appName, appTargetNamespace, appProjectName, cluster, etcdTLSConfigs, systemComponentMap)
 		if err != nil {
 			v32.ClusterConditionMonitoringEnabled.Unknown(cluster)
 			v32.ClusterConditionMonitoringEnabled.Message(cluster, err.Error())
@@ -278,7 +278,14 @@ func (ch *clusterHandler) getExporterEndpoint() (map[string][]string, error) {
 	return endpointMap, nil
 }
 
-func (ch *clusterHandler) deployApp(appName, appTargetNamespace string, appProjectName string, cluster *mgmtv3.Cluster, etcdTLSConfig []*etcdTLSConfig, systemComponentMap map[string][]string) (map[string]string, error) {
+func (ch *clusterHandler) deployApp(
+	appName string,
+	appTargetNamespace string,
+	appProjectName string,
+	cluster *mgmtv3.Cluster,
+	etcdTLSConfig []*etcdTLSConfig,
+	systemComponentMap map[string][]string,
+) (map[string]string, map[string]bool, error) {
 	_, appDeployProjectID := ref.Parse(appProjectName)
 	clusterAlertManagerSvcName, clusterAlertManagerSvcNamespaces, clusterAlertManagerPort := monitoring.ClusterAlertManagerEndpoint()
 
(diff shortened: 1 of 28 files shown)


nikkelma:
Also addresses #22532

nikkelma:
Looks like the `AnswersForceString` field should be accounted for across controllers, will update accordingly.

PennyScissors:
Overall lgtm, just concerned about how the UI should make use of this.

nikkelma:
I'm currently limiting scope to only types and methods that directly impact the correct behavior of Apps, including `AppRevisions` and `AppUpgradeConfig`.

Should multi-cluster apps be included as well? I have that work in-progress, as it makes sense but could be moved to a separate effort.

I'm treating v1 monitoring, v1 logging, v1 CIS as out of scope.

pkg/controllers/managementuser/helm/controller.go
```diff
@@ -396,14 +397,14 @@ func (l *Lifecycle) writeKubeConfig(obj *v3.App, kubePath string, remove bool) e
 
 func isSame(obj *v3.App, revision *v3.AppRevision) bool {
 	if obj.Spec.ExternalID != "" {
-		if revision.Status.ExternalID == obj.Spec.ExternalID && reflect.DeepEqual(revision.Status.Answers, obj.Spec.Answers) && reflect.DeepEqual(revision.Status.ValuesYaml, obj.Spec.ValuesYaml) {
+		if revision.Status.ExternalID == obj.Spec.ExternalID && reflect.DeepEqual(revision.Status.Answers, obj.Spec.Answers) && reflect.DeepEqual(revision.Status.AnswersForceString, obj.Spec.AnswersForceString) && reflect.DeepEqual(revision.Status.ValuesYaml, obj.Spec.ValuesYaml) {
```
nikkelma, line 400: `isSame` should consider `AnswersForceString`, but should there be any normalization performed to ensure only upgrading when needed?

For example ...
```
answersForceString1 = map[string]bool{}
answersForceString2 = map[string]bool{"test-key":false}
```
Should the above be considered to be changed, or because the key `"test-key"` returns false for both cases, should this be a no-op?

PennyScissors: > Should multi-cluster apps be included as well? I have that work in-progress, as it makes sense but could be moved to a separate effort.
>
 I suppose multi-cluster apps could be included.

> Should the above be considered to be changed, or because the key "test-key" returns false for both cases, should this be a no-op?
>
If you are ultimately comparing the same value, either way would be fine with me. 

What are your thoughts @dramich?

PennyScissors: Just following up on this. Do you have any comments on the above @dramich? (I'm not too familiar with this code)

PennyScissors:
Assuming we are not including multi-cluster and the normalization mentioned above, would this be ready for merge?

nikkelma:
Ready to merge - as commit messages show, covered the following functionality:
- Apps support an optional `AnswersForceString` field
- Multi-cluster Apps now support an optional `ValuesForceString` field in each `Answer`
- v1 Monitoring and Alerts account for `AnswersForceString` when deploying their Apps
- cleaned up some code blocks as made sense

Tested the following scenarios:
- `enableMonitoring` and `editMonitoring` respect values forced to strings
- Apps respect values forced to strings on create and update
- Multi-cluster Apps respect values forced to strings on create and update, propagating configs down to each individual App

# KNOWN ISSUE(S)
When editing an API object with these new fields through the UI, _**the relevant new fields are sent to the server as empty, clobbering any changes made to these fields.**_ I believe the fix is as simple as changing all relevant UI screens to store and send the new fields.

PennyScissors (dismissed):
Approving this, but it should still get reviewed by @dramich before merging. I tested multiple cases including multi cluster apps and works as expected. It will require UI changes, but I'll take care of creating a task for that.

pkg/controllers/managementuserlegacy/monitoring/clusterHandler.go
```diff
@@ -348,9 +355,33 @@ func (ch *clusterHandler) deployApp(appName, appTargetNamespace string, appProje
...
+			if k == "true" || k == "false" {
+				notForcedStringKeys = append(notForcedStringKeys, k)
+			}
+			// if any default values are numbers, add a check here to ensure those number values are NOT forced as a string
+		}
+		return notForcedStringKeys
+	}
+
+	knownUnforcedStringKeys = append(knownUnforcedStringKeys, detectNotForcedStringKeys(optionalAppAnswers)...)
+	knownUnforcedStringKeys = append(knownUnforcedStringKeys, detectNotForcedStringKeys(mustAppAnswers)...)
```
dramich, line 375: This section looks incomplete based on the comments. I don't follow why we expose the ability to force things but now we are deleting things a user tells us to force? And this is looking at keys? 

This code is also triplicated so if it's necessary we should create a helper somewhere

nikkelma: Helper makes sense, didn't know where to put it so just need to determine that. I didn't want to impose the number-parsing overhead if not necessary, but it does make sense to add it in.

This functionality you're questioning is analogous to required App settings in the monitoring application. All the surrounding logic here relates to v1 monitoring and extra Helm answers, and specific Answers _must_ have certain values, so therefore certain Answers that are intended to be boolean or numbers _must_ be unable to be forced as a string.

nikkelma: Just added d07994658e22b4e8a0943f81e7fd30aa62784cf5 that addresses the duplicated code and detects numbers as well.

nikkelma:
Closing as this isn't the path being pursued to resolve the target issue. Unfortunate display of a lack of transparency and not encouraging outside collaboration.

Issue #30292, opened 2020-12-01
GKE - Errors after some time Cluster health check failed: Failed to communicate with API server: Unauthorized


izaac:
**What kind of request is this:**

Bug

**Steps to reproduce:**

- Provision a GKE cluster with json key using Rancher
- The only thing I did was to install Rancher logging V2 from Dashboard Apps
- Wait for some time

**Result:**

### 

`Cluster health check failed: Failed to communicate with API server: Unauthorized`

![Screen Shot 2020-12-01 at 1 47 15 PM](https://user-images.githubusercontent.com/27659/100796210-83fb3700-33dd-11eb-8b5a-8190bde9619c.png)

**Other details that may be helpful:**

```
2020/12/01 15:50:37 [INFO] error in remotedialer server [400]: websocket: close 1006 (abnormal closure): unexpected EOF
2020/12/01 15:51:37 [INFO] error in remotedialer server [400]: websocket: close 1006 (abnormal closure): unexpected EOF
2020/12/01 15:54:07 [INFO] error in remotedialer server [400]: websocket: close 1006 (abnormal closure): unexpected EOF
2020/12/01 18:51:22 [ERROR] error syncing 'p-2s9mc/creator-project-owner': handler mgmt-auth-prtb-controller: clusterroles.rbac.authorization.k8s.io "c-spv97-clustermember" already exists, requeuing
2020/12/01 18:55:19 [ERROR] error parsing azure-group-cache-size, skipping update strconv.Atoi: parsing "": invalid syntax
2020/12/01 18:55:19 [ERROR] error syncing 'system-default-registry': handler copy-settings: the server could not find the requested resource, requeuing
2020/12/01 18:55:19 [ERROR] error syncing 'install-uuid': handler copy-settings: the server could not find the requested resource, requeuing
2020/12/01 18:55:19 [ERROR] error syncing 'system-default-registry': handler copy-settings: the server could not find the requested resource, requeuing
2020/12/01 18:55:19 [ERROR] error syncing 'install-uuid': handler copy-settings: the server could not find the requested resource, requeuing
2020/12/01 18:55:19 [ERROR] error syncing 'system-default-registry': handler copy-settings: the server could not find the requested resource, requeuing
2020/12/01 18:55:19 [ERROR] error syncing 'install-uuid': handler copy-settings: the server could not find the requested resource, requeuing
[...]

Issue #30170, opened 2020-11-20
[2.4 backport] k8s 1.19 removes basic auth but it's in use by kontainer-engine


cbron:
https://github.com/rancher/rancher/issues/28673

deniseschannon:
This seems to only be needed for IF we add in k8s 1.19 to 2.4, but there are no plans to do so.

cbron:
@deniseschannon this was the situation where its possible cloud providers like GKE could end up above what RKE/Rancher supports.

sadiapoddar:
Verified the fix on 2.4-head, commit ca0c2b9f4
Steps
Created a 3 nodes gke cluster
From google console, Disabled basic-auth as recommended via gcloud container clusters update CLUSTER_NAME --no-enable-basic-auth
In Rancher, UI -> Edited the Cluster (using same SA credentials as on initial cluster setup)
Triggered the operational (e.g. update master cluster Kubernetes version from 1.17.14-gke.1600 to 1.17.17-gke-400)
Rancher triggered the update and the cluster return healthy state after the update
From the console re enabled the basic-auth and cluster updated without any error

<img width="1357" alt="Screen Shot 2020-12-28 at 11 12 55 AM" src="https://user-images.githubusercontent.com/72409237/103237853-c5f49b80-48fd-11eb-8208-2b7e2694a83f.png">

PR #30311, opened 2020-12-02
[2.4] Revert "Remove basic auth from GKE driver"

thedadams:
This reverts commit 9ced45f6160cf43cfa3de96cee95fa2f3969588c.

The tokens being used expire after 1 hour and are not being refreshed. A mechanism for token refresh should be implemented.

Original PR:
https://github.com/rancher/rancher/pull/30174

Issues:
https://github.com/rancher/rancher/issues/30292
https://github.com/rancher/rancher/issues/30170

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,7 +79,7 @@ require (
 	github.com/prometheus/common v0.9.1
 	github.com/prometheus/tsdb v0.8.0 // indirect
 	github.com/rancher/dynamiclistener v0.2.1-0.20200418023342-52ede5ec9234
-	github.com/rancher/kontainer-engine v0.0.4-dev.0.20201120185903-87b9f0bb2861
+	github.com/rancher/kontainer-engine v0.0.4-dev.0.20201202223022-33de45824b08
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -79,7 +79,7 @@ require (
 	github.com/prometheus/common v0.9.1
 	github.com/prometheus/tsdb v0.8.0 // indirect
 	github.com/rancher/dynamiclistener v0.2.1-0.20200418023342-52ede5ec9234
-	github.com/rancher/kontainer-engine v0.0.4-dev.0.20201120185903-87b9f0bb2861
+	github.com/rancher/kontainer-engine v0.0.4-dev.0.20201106160106-4442123b5eaa
```
cbron, line 82: Now that I think about it, it would be safer if you pull in the reverted KE changes just in case, rather than reverting this. If there is another KE change down the line we don't want any accidental things to pop up that the revert-commit introduced.

cbron (approved):

Issue #30325, opened 2020-12-03
"Cluster health check failed: deployments.apps "cattle-cluster-agent" not found" error seen while deploying clusters


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a cluster - DO/Custom - 1etcd/control + 1 worker/1 all roles
- Cluster is seen in error state with error:

<img width="1276" alt="Screen Shot 2020-12-03 at 6 31 32 AM" src="https://user-images.githubusercontent.com/26032343/101041239-355eb180-3531-11eb-8ee0-a0cc61971e30.png">

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5-head - commit id: `82a0a33b`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): DO/custom
- Kubernetes version (use `kubectl version`):

```
1.19.4
```

sowmyav27:
On 2.5-head - commit id: `e2e530c19`

- Able to bring up clusters with no errors.

PR #30328, opened 2020-12-03
Fix deploy agents

thedadams:
A mistake was made in the code for clusters with private API endpoint that caused the cluster/node agents to not be deployed.

Issue:
https://github.com/rancher/rancher/issues/30325

### DIFF
--- a/pkg/dialer/factory.go
+++ b/pkg/dialer/factory.go
@@ -69,17 +69,17 @@ func IsCloudDriver(cluster *v3.Cluster) bool {
 }
 
 func IsPublicCloudDriver(cluster *v3.Cluster) bool {
-	return IsCloudDriver(cluster) && HasPublicAPIEndpoint(cluster)
+	return IsCloudDriver(cluster) && !HasOnlyPrivateAPIEndpoint(cluster)
 }
 
-func HasPublicAPIEndpoint(cluster *v3.Cluster) bool {
+func HasOnlyPrivateAPIEndpoint(cluster *v3.Cluster) bool {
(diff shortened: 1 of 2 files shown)


pkg/dialer/factory.go
```diff
@@ -69,17 +69,17 @@ func IsCloudDriver(cluster *v3.Cluster) bool {
...
-func HasPublicAPIEndpoint(cluster *v3.Cluster) bool {
+func HasOnlyPrivateAPIEndpoint(cluster *v3.Cluster) bool {
 	switch cluster.Status.Driver {
 	case v32.ClusterDriverEKS:
-		return cluster.Status.EKSStatus.UpstreamSpec == nil ||
-			cluster.Status.EKSStatus.UpstreamSpec.PublicAccess == nil ||
-			*cluster.Status.EKSStatus.UpstreamSpec.PublicAccess
+		return cluster.Status.EKSStatus.UpstreamSpec != nil &&
+			cluster.Status.EKSStatus.UpstreamSpec.PublicAccess != nil &&
+			!*cluster.Status.EKSStatus.UpstreamSpec.PublicAccess
```
cbron, line 80: I wonder if you should check private access too. I think public should be enough because if public is off then private should be on. Or roughly, public being off means it _is_ private.

cbron (approved):

pkg/dialer/factory.go
thedadams: For EKS, you can't create a cluster with both private and public off.

PR #30358, opened 2020-12-04
Remove private endpoint check when deploying agent

thedadams:
When an EKS cluster is created with private-only API endpoint, the user is responsible for deploying the cluster agent. There was a check to try to ensure that the cluster agent was not deployed by Rancher. However, this check is not needed because, in this scenario, there are no nodes detected before the cluster agent is deployed.

Issue:
https://github.com/rancher/rancher/issues/29907

Related PRs:
https://github.com/rancher/rancher/pull/30328
https://github.com/rancher/rancher/pull/30120


nickgerace (approved):
LGTM

cbron (approved):
lgtm, but would like dans approval

dramich (approved):

PR #30371, opened 2020-12-07
Remove basic auth with GKE clusters

thedadams:
Basic authentication is deprecated and will be removed. In preparation for this, basic authentication is removed for GKE clusters.

Issue:

### DIFF
--- a/pkg/clustermanager/manager.go
+++ b/pkg/clustermanager/manager.go
@@ -23,6 +23,7 @@ import (
 	"github.com/rancher/rancher/pkg/clusterrouter"
 	clusterController "github.com/rancher/rancher/pkg/controllers/managementuser"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/kontainer-engine/drivers/gke"
 	"github.com/rancher/rancher/pkg/rbac"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/types/config"
@@ -31,6 +32,7 @@ import (
 	"github.com/rancher/steve/pkg/accesscontrol"
 	rbacv1 "github.com/rancher/wrangler/pkg/generated/controllers/rbac/v1"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/oauth2"
 	"golang.org/x/sync/semaphore"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/clustermanager/manager.go
```diff
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/oauth2"
```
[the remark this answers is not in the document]
thedadams: Moved.

pkg/clustermanager/manager.go
```diff
@@ -315,6 +318,13 @@ func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Con
 					ht.DialTLS = tlsDialer
 				}
 			}
+			if cluster.Status.Driver == "googleKubernetesEngine" {
+				ts, _ := gke.GetTokenSource(context.RunContext, (*cluster.Spec.GenericEngineConfig)["credential"].(string))
```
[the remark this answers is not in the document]
thedadams: Yes, I get complication errors without these parenthesis. Without them, it tries to dereference `cluster.Spec.GenericEngineConfig["credential"]`. However, I am trying to dereference `cluster.Spec.GenericEngineConfig` and then look at the `credential` entry.

pkg/clustermanager/manager.go
```diff
@@ -315,6 +317,15 @@ func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Con
 					ht.DialTLS = tlsDialer
 				}
 			}
+			if cluster.Status.Driver == "googleKubernetesEngine" {
+				ts, err := gke.GetTokenSource(context.RunContext, (*cluster.Spec.GenericEngineConfig)["credential"].(string))
+				if err != nil {
```
mrajashree, line 322: why don't we return the error here? should we log the error?

pkg/kontainer-engine/drivers/gke/gke_driver.go
```diff
@@ -786,7 +781,12 @@ func (d *Driver) PostCheck(ctx context.Context, info *types.ClusterInfo) (*types
 		return nil, err
 	}
 
-	svc, err := d.getServiceClient(ctx, state)
+	ts, err := GetTokenSource(ctx, state.CredentialContent)
```
mrajashree, line 784: this can be replaced with `getServiceClient(ctx, state.CredentialContent)` like in the update method?

thedadams: We need the token source later for the `ServiceAccountToken`.

pkg/clustermanager/manager.go
thedadams: Will log the error.

mrajashree (dismissed):

pkg/clustermanager/manager.go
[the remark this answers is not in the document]
rmweir: ah that makes sense

rmweir (approved):
LGTM, 1 nit

pkg/kontainer-engine/drivers/gke/gke_driver.go
```diff
@@ -986,7 +954,7 @@ func (d *Driver) getClusterStats(ctx context.Context, info *types.ClusterInfo) (
 		return nil, err
 	}
 
-	svc, err := d.getServiceClient(ctx, state)
+	svc, err := getServiceClient(ctx, state.CredentialContent)
```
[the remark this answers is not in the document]
thedadams: Removed as many as I could find in this file.

pkg/clustermanager/manager.go
```diff
@@ -315,6 +317,16 @@ func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Con
 					ht.DialTLS = tlsDialer
 				}
 			}
+			if cluster.Status.Driver == "googleKubernetesEngine" {
+				ts, err := gke.GetTokenSource(context.RunContext, (*cluster.Spec.GenericEngineConfig)["credential"].(string))
```
cbron, line 321: `spec.GenericEngineConfig` is a pointer, we should check for nil otherwise you can get a panic here. Possible that driver is set but spec is not.

cbron (approved):

Issue #29429, opened 2020-10-07
Rancher 2.4.8 Memory Leak


thecrazyrussian:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Currently, the issue exists in 1 production cluster somewhat replicated to a test cluster.
We have around ~60 downstream clusters that are often disconnected. We have many catalog apps but so does our test environment which does not have these issues. 

**Result:**
After running for a couple of days, the pods continue to use more memory until they exhaust all of the nodes' memory and cause a node crash.

This is a grafana time-laps of the events, note the continuous increase in memory consumption.
![Rancher Pod Memory 2](https://user-images.githubusercontent.com/3221320/95366567-3957b580-0899-11eb-997a-d836542d218e.jpg)
![Rancher Pod Memory 1](https://user-images.githubusercontent.com/3221320/95366573-39f04c00-0899-11eb-9439-c31db19049c1.jpg)

**Mitigation**
To prevent the nodes from crashing, we set a limit on the rancher pods that would prevent them from bringing down the cluster at 6Gig per pod. That has allowed the pod to crash and restart without taking the node down.

**Other details that may be helpful:**
We saw a couple of other issues related to memory leaks that might be relevant since we also use azure ad and catalogs.

**Environment information**
2 x k3s,4cpu 16gb nodes in Azure
Rancher 2.4.8
Hosted MySQL
Azure AD Rbac

There is already a disconnected cluster issue from last year.
https://github.com/rancher/rancher/issues/24204 (open)

As well as 2 other open leak Issues that may not be related but since we use both Azure AD and catalogs, it is worth looking into. 
https://github.com/rancher/rancher/issues/29055 (azure ad - open)
https://github.com/rancher/rancher/issues/22949 (catalog - open)

gz#13272

dramich:
To test, create a rancher setup, enable monitoring and add clusters to it. I used 15 but the more you have the fast you can see the leak. Then, delete both agent deployments/daemonsets from the downstream clusters. This will cause rancher to try and reconnect to the clusters in a loop. If done properly the logs should show `Stopping cluster agent for <ID>` every two minutes which is expected and how the leak is reproduced. 

```
kubectl delete deployment -n cattle-system cattle-cluster-agent &&
kubectl delete daemonsets.apps -n cattle-system cattle-node-agent
```

sowmyav27:
[...]

PR #30379, opened 2020-12-08
Move creation of accesscontrol

dramich:
Problem:
When a cluster agent is diconnected rancher tries to start the
controllers in a slow loop, but this causes a memory leak. The leak
stems from adding a handler in client-go which has no method to
deregister the handler so a fixed size array leaks every time.

Solution:
Don't create access control until Rancher has verified a connection to
the cluster which is a good indication controllers will start. There is
technically a small leak here every time we start these controllers.

https://github.com/rancher/rancher/issues/29429


ibuildthecloud:
Oops, I meant to review and I merged instead.  Hopefully that's fine.

Issue #30021, opened 2020-11-10
Update EKSClusterConfig syncing to include launch template information


thedadams:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**New Result:**
Syncing EKS clusters needs to take into account the template ids on each node group. This also includes the writing the `eksclusterconfig` CRD properly to reflect the new values. There is some subtlety here: there may be a launch template on a node group that was specified by Rancher and not by the user. Therefore, in this case, the template id and version should not be written to `Spec`.

**Other details that may be helpful:**
The `eks-operator` should be updated here to include the new fields on the `EKSClusterConfig` type. A `go generate` may be required.

thedadams:
https://github.com/rancher/rancher/issues/28803

aaronRancher:
The enhancement is verified with Rancher server `rancher\rancher:v2.5-head` `version 8201e08`

EKS Config syncing includes template information. Template IDs are seen on each node group when using a preexisting launch template from EKS console.
Node group config:
```
"nodeGroups": [ 3 items
{
"desiredSize": 2,
"diskSize": null,
"ec2SshKey": null,
"gpu": false,
"imageId": null,
"instanceType": null,
"labels": { },
"launchTemplate": {
"id": "lt-01cf5f5cd-----",
"name": "sowmya-test",
"type": "/v3/schemas/launchTemplate",
"version": 2
},
"maxSize": 2,
"minSize": 2,
"nodegroupName": "aaron-ng-byolt",
"requestSpotInstances": false,
"resourceTags": null,
"spotInstanceTypes": null,
"subnets": [ 3 items
"subnet-",
"subnet-",
"subnet-"
],
"tags": { },
"type": "/v3/schemas/nodeGroup",
"userData": null,
"version": "1.18"
},

When a node group uses a Rancher Managed template, the launch template ID is omitted from the node group config:
```
"desiredSize": 2,
"diskSize": 20,
"ec2SshKey": null,
"gpu": false,
"imageId": null,
"instanceType": "t3.medium",
"labels": { },
"launchTemplate": null,
"maxSize": 2,
"minSize": 2,
"nodegroupName": "aaron-test-gpu",
"requestSpotInstances": false,
"resourceTags": { },
"spotInstanceTypes": null,
"subnets": [ 3 items
"subnet-",
"subnet-",
"subnet-"
],
"tags": { },
"type": "/v3/schemas/nodeGroup",
"userData": null,
"version": "1.18"
```

[...]

PR #30384, opened 2020-12-08
Add support for EKS launch templates

thedadams:
The most significant part are the changes to syncing. The syncing
algorithm more carefully crawls the cluster spec to only sync the parts
that are not nil in the config. This is useful because it allows
tracking of the Rancher managed launch template in the upstream spec,
but this will not sync this template to the spec (the launch
template will be nil because it is not provided by the user).

The rancher managed launch template ID is also tracked in status for
updates.

This PR will need to be updated once the related PR is merged.

Related PR:
https://github.com/rancher/eks-operator/pull/27

Issue:
https://github.com/rancher/rancher/issues/30021


pkg/apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -344,8 +344,9 @@ type SaveAsTemplateOutput struct {
...
 type EKSStatus struct {
-	UpstreamSpec   *eksv1.EKSClusterConfigSpec `json:"upstreamSpec"`
-	VirtualNetwork string                      `json:"virtualNetwork"`
-	Subnets        []string                    `json:"subnets"`
-	SecurityGroups []string                    `json:"securityGroups"`
+	UpstreamSpec     *eksv1.EKSClusterConfigSpec `json:"upstreamSpec"`
+	VirtualNetwork   string                      `json:"virtualNetwork"`
+	Subnets          []string                    `json:"subnets"`
+	SecurityGroups   []string                    `json:"securityGroups"`
+	LaunchTemplateID string                      `json:"launchTemplateId"`
```
rmweir, line 351: there could be multiple launch template ID's right? or is this just the one that we create?

go.mod
```diff
@@ -46,7 +46,7 @@ require (
 	github.com/DataDog/zstd v1.4.5 // indirect
 	github.com/Masterminds/semver/v3 v3.1.0
 	github.com/Masterminds/sprig/v3 v3.1.0
-	github.com/aws/aws-sdk-go v1.33.5
```
aiyengar2, line 49: Could we split the go mod changes + generated changes into a separate commit(s)?

pkg/client/generated/management/v3/zz_generated_node_group.go
```diff
@@ -1,32 +1,40 @@
...
+	NodeGroupFieldInstanceType   = "instanceType"
+	NodeGroupFieldLabels         = "labels"
+	NodeGroupFieldLaunchTemplate = "launchTemplate"
+	NodeGroupFieldMaxSize        = "maxSize"
+	NodeGroupFieldMinSize        = "minSize"
+	NodeGroupFieldNodegroupName  = "nodegroupName"
+	NodeGroupFieldResourceTags   = "resourceTags"
+	NodeGroupFieldSubnets        = "subnets"
+	NodeGroupFieldTags           = "tags"
+	NodeGroupFieldUserData       = "userData"
```
aiyengar2, line 19: Should this be `userdata`?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -303,6 +303,13 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			}
 		}
 
+		clusterLaunchTemplateID := status["launchTemplateID"].(string)
```
aiyengar2, line 306:
```suggestion
		clusterLaunchTemplateID := status[EKSStatusFieldLaunchTemplateID].(string)
```

pkg/apis/management.cattle.io/v3/cluster_types.go
thedadams: This is just the one we create and keep track of.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Technically, these are different fields. The `EKSStatusFieldLaunchTemplateID` refers to a field of the `cluster` object, but we are looking for a field of the `eksclusterconfig` object.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -151,7 +153,16 @@ func (e *eksRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluste
...
-	_, eksService, err := controller.StartAWSSessions(secretsCache, *cluster.Spec.EKSConfig)
+	switch cluster.Status.Driver {
+	case apimgmtv3.ClusterDriverEKS:
+		return BuildEKSUpstreamSpec(secretsCache, cluster)
+	default:
+		return nil, fmt.Errorf("unsupported cloud driver")
+	}
+}
+
+func BuildEKSUpstreamSpec(secretsCache wranglerv1.SecretCache, cluster *mgmtv3.Cluster) (*v1.EKSClusterConfigSpec, error) {
```
rmweir, line 164: Don't need to export function
```suggestion
func buildEKSUpstreamSpec(secretsCache wranglerv1.SecretCache, cluster *mgmtv3.Cluster) (*v1.EKSClusterConfigSpec, error) {
```

rmweir (changes requested):
One change, otherwise LGTM.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -182,10 +207,26 @@ func GetComparableUpstreamSpec(secretsCache wranglerv1.SecretCache, cluster *mgm
 		}
 
 		nodeGroupStates = append(nodeGroupStates, ng)
+		var nodeGroupMustBeDeleted string
+		if len(ng.Nodegroup.Health.Issues) != 0 {
+			var issueMessages []string
+			for _, issue := range ng.Nodegroup.Health.Issues {
+				issueMessages = append(issueMessages, aws.StringValue(issue.Message))
+				if !controller.NodeGroupIssueIsUpdatable(aws.StringValue(issue.Code)) {
```
rmweir, line 215: why not just check if the status is "degraded"?

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -232,6 +232,10 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			}
 		}
 
+		if apimgmtv3.ClusterConditionUpdated.IsFalse(cluster) && strings.HasPrefix(apimgmtv3.ClusterConditionUpdated.GetMessage(cluster), "[Syncing error") {
+			return cluster, nil
```
rmweir, line 236: I understand that coming back in may not do anything until the cluster has been updated, I can't help but feel it is appropriate to return the error here, for debugging purposes and because it is simply not in a good state. @cbron thoughts?

rmweir (changes requested):

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
thedadams: A node group can be degraded for a lot of reasons, and some of them can be taken care of by an update. Here, we are checking to see if the degraded status is unrecoverable.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -95,14 +98,26 @@ func (e *eksRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluste
...
+			return cluster, statusErr
+		}
 		return cluster, err
 	}
 
 	if !reflect.DeepEqual(cluster.Status.EKSStatus.UpstreamSpec, upstreamSpec) {
 		logrus.Infof("updating cluster [%s], upstream change detected", cluster.Name)
 		cluster = cluster.DeepCopy()
 		cluster.Status.EKSStatus.UpstreamSpec = upstreamSpec
+		if err != nil {
```
cbron, line 117: This section is quite confusing to me, is this err from `GetComparableUpstreamSpec` but upstreamSpec isn't nil and it's not equal ? Can we push this up outside of the conditional ?

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -182,10 +207,26 @@ func GetComparableUpstreamSpec(secretsCache wranglerv1.SecretCache, cluster *mgm
...
+			for _, issue := range ng.Nodegroup.Health.Issues {
+				issueMessages = append(issueMessages, aws.StringValue(issue.Message))
+				if !controller.NodeGroupIssueIsUpdatable(aws.StringValue(issue.Code)) {
+					nodeGroupMustBeDeleted = ": node group cannot be updated, must be deleted and recreated"
+				}
+			}
+			errs = append(errs, fmt.Sprintf("health error for node group [%s]: %s%s",
+				aws.StringValue(ng.Nodegroup.NodegroupName),
+				strings.Join(issueMessages, "; "),
+				nodeGroupMustBeDeleted,
```
cbron, line 222: This will always append an empty string to the errs

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -95,14 +98,26 @@ func (e *eksRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluste
 	}
 
 	upstreamSpec, err := GetComparableUpstreamSpec(e.secretsCache, cluster)
-	if err != nil {
+	if err != nil && upstreamSpec == nil {
+		var statusErr error
+		cluster = cluster.DeepCopy()
+		apimgmtv3.ClusterConditionUpdated.False(cluster)
+		apimgmtv3.ClusterConditionUpdated.Message(cluster, "[Syncing error: syncing broken] "+err.Error())
```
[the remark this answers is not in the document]
thedadams: Yes, I agree that we need a better phrase.

There are two reasons for a "Syncing error":
- When we sync, we noticed that one or more of the node groups are in a state that cannot be recovered from. The syncing of the EKS state still occurs here.
- When we sync, that actual syncing does not occur because of some type of problem.

Perhaps "syncing failed" would be better?

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
thedadams: The last `nodeGroupMustBeDelete` is in the `fmt.Sprintf` call. The phrase `: node group cannot be updated, must be deleted and recreated` is being added to the error string if the check `nodeGroupIssueIsUpdatable` fails.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: There are two reasons why I didn't return the error here:
1. The error does not occur here. It is just being checked here.
2. If the Updated message begins with syncing error, then the syncer needs to run again or there must be action by the user to rectify the situation. That is, the user must either update the node groups or create new ones and delete old ones. This will not be fixed by returning an error here and re-entering the controller.

thedadams: I'm happy to change it if my thinking is incorrect. I was only answering your question.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
thedadams: If `err != nil && upstreamSpec == nil`, then the syncing failed for some reason, so it is reported to the user in the Updated message.

If `upstreamSpec != nil`, then the syncing occurred as expected. If `err != nil` in this case, then there were health issues with one (or more) of the node groups that needs to be reported to the user. These are health issues like "the node group is degraded because the user deleted the launch template." These types of health issues are not reported to the user at any other time because there are not Kubernetes health issues.

cbron: My question is why wouldn't this be on line 116, as just `if err != nil && upstreamSpec != nil`. In this block we are also checking if its the same as spec. So where is this case being handled ?

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
cbron: Yes, bad phrasing on my part. If `nodeGroupMustBeDeleted` is not set on line 221, then this will still append an empty string. Does that end up in ux/logs somehow ?

pkg/controllers/management/eks/eks_cluster_handler.go
cbron: > If the Updated message begins with syncing error, then the syncer needs to run again or there must be action by the user to rectify the situation. That is, the user must either update the node groups or create new ones and delete old ones. This will not be fixed by returning an error here and re-entering the controller.

Would the user find out about it if you don't return the error here ?

thedadams: The error string is set as the message on the "Updated" condition and appears in the UI.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
thedadams: Ah yes. I believe I understand what you are saying, and this does need to be moved.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
thedadams: This error string does eventually show up in the UI.

If `nodeGroupMustBeDeleted` is empty, then it will be put into the last format string verb and will be undetectable in the formatted string.

pkg/controllers/management/eks/eks_cluster_handler.go
rmweir: I mainly like the persistence logs of returning the error even if it will not be able to fix itself

rmweir: It seems like we should return an error if the cluster is broken. The error was encountered in syncing but now this handler should error because it cannot work without the syncing working. At the very least syncing cron job will log the error but doing both might make sense.

pkg/controllers/management/eksupstreamrefresh/eks_upstream_refresh.go
```diff
@@ -182,10 +215,27 @@ func GetComparableUpstreamSpec(secretsCache wranglerv1.SecretCache, cluster *mgm
 		}
 
 		nodeGroupStates = append(nodeGroupStates, ng)
+		var nodeGroupMustBeDeleted string
+		if len(ng.Nodegroup.Health.Issues) != 0 {
+			var issueMessages []string
+			for _, issue := range ng.Nodegroup.Health.Issues {
+				issueMessages = append(issueMessages, aws.StringValue(issue.Message))
+				if !controller.NodeGroupIssueIsUpdatable(aws.StringValue(issue.Code)) {
```
rmweir, line 223: The template missing is the only one that totally stops syncing from working right? I'm wondering now if we should continue treating them all the same since the eks handler is going to essentially shut down until it's resolved.

rmweir (commented):
left comments

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: OK. The handler will return the error if it sees a syncing error.

rmweir (approved):
LGTM

cbron (approved):

PR #30397, opened 2020-12-09
Allow Helm installation without ingress

Gallardo994:
Currently, you cannot install Helm-version of Rancher without an ingress object. If the cluster has no ingress controller, Ingress object will be stuck in Initialized state, e.g.:
image

For installations without ingress controllers, it makes sense to have an ability to disable creation of ingress object completely, for example if it's LoadBalancer-only cluster, or you use some specific software (for example like us, Traefik with IngressRoute only)

This PR adds an ability to use `ingress.enabled: false` to prevent Helm from installing Ingress to the cluster. By default, it is set to `true` so nothing breaks on old installations or upgrades.

https://github.com/rancher/rancher/issues/30535


paynejacob (approved):

dramich:
@Gallardo994 Thanks for the PR. I also back ported this to the 2.5 line so it will be in an upcoming release after 2.5.4 ships.

Issue #30012, opened 2020-11-10
Validate Airgap configuration using Rancherd


sowmyav27:
Validate Airgap configuration. Take notes so this can get documented (sync notes with Catherine)

Related issue: https://github.com/rancher/rancher/issues/29797

To Validate:
- Deploy rancherd in airgap setup
- Verify all P1 use cases in an air gap setup.

rancher-max:
### Attempted in v2.5.4-rc2

**Results:**
RKE2 server comes up successfully. 
Rancher helm job fails with ImagePullBackOff, as it is trying to pull the public `rancher/klipper-helm` image instead of the one from my private registry:
```
  Warning  Failed            12s                  kubelet            Failed to pull image "rancher/klipper-helm:v0.3.0": rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/rancher/klipper-helm:v0.3.0": failed to resolve reference "docker.io/rancher/klipper-helm:v0.3.0": failed to do request: Head https://registry-1.docker.io/v2/rancher/klipper-helm/manifests/v0.3.0: dial tcp 54.236.131.166:443: i/o timeout
  Normal   BackOff           1s (x2 over 88s)     kubelet            Back-off pulling image "rancher/klipper-helm:v0.3.0"
  Warning  Failed            1s (x2 over 88s)     kubelet            Error: ImagePullBackOff
```


**Steps:**
Note:
- `$FQDN` = Public FQDN of my public facing private registry instance
- `$PRIVATEIP` = Private IP of my airgapped (no public ip available) instance

1. Create a private registry that uses tls but is not password protected:
```
mkdir -p certs
openssl req \
  -newkey rsa:4096 -nodes -sha256 -keyout certs/domain.key \
  -x509 -days 365 -out certs/domain.crt -subj "/C=US/ST=AZ/O=Rancher QA/CN=$FQDN"

sudo docker run -d \
  --restart=always \
  --name registry \
  -v "$(pwd)"/certs:/certs \
  -e REGISTRY_HTTP_ADDR=0.0.0.0:443 \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
  -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
  -p 443:443 \
  registry:2
```
2. Ensure docker recognizes the private registry:
```
sudo mkdir -p /etc/docker/certs.d/$FQDN && \
[...]

PR #30409, opened 2020-12-09
Split rancher.yaml into multiple doc

StrongMonkey:
Due to the limitation https://github.com/rancher/rke2/issues/527 we can
not use multi-doc yaml for airgap installation

### DIFF
--- a/scripts/build-rancherd
+++ b/scripts/build-rancherd
@@ -31,3 +31,5 @@ cd ../..
 ./scripts/chart/package
 
 sed -e "s|%{CHART_CONTENT}%|$(base64 -w0 ./bin/chart/*/rancher-${CHART_VERSION}.tgz)|g" ./scripts/rancher.yaml.tmpl > ./bin/rancher.yaml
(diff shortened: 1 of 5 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/30012

scripts/build-rancherd
```diff
@@ -31,3 +31,5 @@ cd ../..
 ./scripts/chart/package
 
 sed -e "s|%{CHART_CONTENT}%|$(base64 -w0 ./bin/chart/*/rancher-${CHART_VERSION}.tgz)|g" ./scripts/rancher.yaml.tmpl > ./bin/rancher.yaml
+cp ./scripts/rancher-namespace.yaml.tmpl ./bin/rancher-namespace.yaml
```
brandond, line 34: is it going to surprises someone later on down the road that you're just copying this template file as-is instead of actually templating it?

brandond (approved):

scripts/build-rancherd
StrongMonkey: Yeah, Just following the same naming conversion in case we need to template. Think It should be fine in here :)

Issue #29565, opened 2020-10-16
rancherd reset-password can panic on startup


superseb:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
```
curl -sL https://get.rancher.io | sh
systemctl  daemon-reload
systemctl start rancherd-server
while true; do rancherd reset-admin; sleep 1; done
```

**Result:**
```
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
FATA[0000] Get "https://127.0.0.1:6443/apis/management.cattle.io/v3/users?labelSelector=authz.management.cattle.io%2Fbootstrapping%3Dadmin-user": dial tcp 127.0.0.1:6443: connect: connection refused 
[...]

PR #30429, opened 2020-12-10
Reorganize reset-admin to make sure everthing get created

StrongMonkey:
Refactored logic so that only if admin user, globalrolebinding,
clusterrolebinding are created before we create the configmap. The
previous logic gated on whether user has been created which could fail
mid-way if cluster is not fully bootstraped


StrongMonkey:
https://github.com/rancher/rancher/issues/29565

MonzElmasry (approved):

Issue #30403, opened 2020-12-09
Private EKS import command should not be needed when Rancher is in the same VPC


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Create a single node Rancher instance in EC2.
From Rancher, deploy an EKS cluster .
Set cluster to private network access. Launch the cluster in the same VPC and subnet as the Rancher instance.
Create cluster.

**Result:**
Rancher displays an import command. The cluster is unreachable until the command is run and the cluster agent is deployed.
For clusters deployed in the same VPC and subnet, we would expect the cluster to connect automatically.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
`rancher\rancher:master-head` `version ff16903` and `rancher\rancher:v2.5-head` `version b2953cb`
- Installation option (single install/HA): Single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Hosted

aaronRancher:
Tested with Rancher server `rancher\rancher:master-head` `version 495f7a1` and `rancher\rancher:v2.5-head` `version e2295c6`

Steps:
Create a single node Rancher instance in EC2.
From Rancher, deploy an EKS cluster .
Set cluster to private network access. Launch the cluster in the same VPC and subnet as the Rancher instance.
![Screen Shot 2020-12-15 at 1 16 37 PM](https://user-images.githubusercontent.com/57409052/102269675-37822200-3eda-11eb-9a21-7e798faa8dd7.png)
Create cluster.

Result:
The cluster is automatically imported into rancher.
![Screen Shot 2020-12-15 at 1 34 36 PM](https://user-images.githubusercontent.com/57409052/102269754-52549680-3eda-11eb-8a8a-51da7feb71ed.png)

aaronRancher:
But on master-head, I am unable to import a private EKS cluster from a Rancher setup which is not in the same VPC and subnet.
**Steps:**
Create a single node Rancher instance in EC2.
From Rancher, deploy an EKS cluster .
Set cluster to private network access. Launch the cluster in a different VPC and subnet as the Rancher instance.

Result:
Import command is never displayed. Cluster is stuck in provisioning state:
[...]

PR #30439, opened 2020-12-11
Check for cluster connection when public access disabled

thedadams:
If public access is disabled, it is no longer assumed that Rancher cannot communicate with the cluster API. Instead, Rancher attempts to contact the API. If this is successful, then Rancher can communicate with the cluster as normal (through the API access endpoint) instead of tunneling.

This is most important on cluster creation/import. In this case, since Rancher can still communicate with the API endpoint, the user does not need to run the import command that is displayed.

Issue:
https://github.com/rancher/rancher/issues/30403

### DIFF
--- a/pkg/apis/management.cattle.io/v3/cluster_types.go
+++ b/pkg/apis/management.cattle.io/v3/cluster_types.go
@@ -344,8 +344,9 @@ type SaveAsTemplateOutput struct {
 }
 
 type EKSStatus struct {
-	UpstreamSpec   *eksv1.EKSClusterConfigSpec `json:"upstreamSpec"`
-	VirtualNetwork string                      `json:"virtualNetwork"`
-	Subnets        []string                    `json:"subnets"`
-	SecurityGroups []string                    `json:"securityGroups"`
+	UpstreamSpec          *eksv1.EKSClusterConfigSpec `json:"upstreamSpec"`
+	VirtualNetwork        string                      `json:"virtualNetwork"`
+	Subnets               []string                    `json:"subnets"`
+	SecurityGroups        []string                    `json:"securityGroups"`
+	PrivateRequiresTunnel *bool                       `json:"privateRequiresTunnel"`
 }
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -3,14 +3,16 @@ package eks
 import (
 	"context"
 	"encoding/base64"
+	stderrors "errors"
 	"fmt"
+	"net"
 	"os"
 	"reflect"
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/rancher/eks-operator/controller"
 	eksv1 "github.com/rancher/eks-operator/pkg/apis/eks.cattle.io/v1"
@@ -197,15 +199,13 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		}
 
 		if cluster.Status.EKSStatus.UpstreamSpec == nil {
-			if cluster.Status.EKSStatus.UpstreamSpec == nil {
-				cluster, err = e.setInitialUpstreamSpec(cluster)
-				if err != nil {
-					if !notFound(err) {
-						return cluster, err
-					}
+			cluster, err = e.setInitialUpstreamSpec(cluster)
+			if err != nil {
+				if !notFound(err) {
+					return cluster, err
 				}
-				return cluster, nil
 			}
+			return cluster, nil
 		}
 
 		e.clusterEnqueueAfter(cluster.Name, enqueueTime)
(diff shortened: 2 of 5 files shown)


pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -480,6 +485,27 @@ func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Clu
 	return e.clusterClient.Update(cluster)
 }
 
+func (e *eksOperatorController) canUsePublicAPIEndpoint(cluster *mgmtv3.Cluster) (string, *bool) {
```
rmweir, line 488: If you are generating the SA token here then the name should be changed to more accurately reflect what is happening.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -282,7 +280,21 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			}
 		}
 
-		if cluster.Status.ServiceAccountToken == "" || cluster.Status.APIEndpoint == "" {
+		if cluster.Status.APIEndpoint == "" {
+			return e.getCAAndAPIEndpoint(cluster)
```
rmweir, line 284: does this function record something on the cluster? If so, the name should be changed or record should happen outside of it. I know it's not part of this PR but I usually do not expect mutations from get functions. Disclaimer: totally possible I wrote it 😅

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -480,6 +485,27 @@ func (e *eksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Clu
...
+
+	netDialer := net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+
+	serviceToken, err := generateSAToken(cluster.Status.APIEndpoint, cluster.Status.CACert, accessToken, netDialer.DialContext)
+
+	if err == nil {
+		publicAccess = true
```
rmweir, line 503: Can we check for a specific error, like a timeout? otherwise maybe we just need to retry, in which case maybe this function should just return the err so the parent function can handle it.

rmweir (changes requested):

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Yes, that is a bad name.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: Yes, that is a bad name.

pkg/controllers/management/eks/eks_cluster_handler.go
thedadams: If Rancher can't contact the API because it is private-only, would this error always be a timeout?

thedadams: It's a `net.DNSError`, and it is handled more gracefully.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -608,21 +619,34 @@ func (e *eksOperatorController) deployEKSOperator() error {
...
 	if err != nil {
-		return "", err
+		return "", nil
 	}
 
-	generator, err := token.NewGenerator(false, false)
+	netDialer := net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
```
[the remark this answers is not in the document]
thedadams: No, this is trying without the tunnel. I can try a shorter timeout, but I chose this because this is the default in the `dialer` package.

pkg/apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -344,8 +344,9 @@ type SaveAsTemplateOutput struct {
...
 type EKSStatus struct {
-	UpstreamSpec   *eksv1.EKSClusterConfigSpec `json:"upstreamSpec"`
-	VirtualNetwork string                      `json:"virtualNetwork"`
-	Subnets        []string                    `json:"subnets"`
-	SecurityGroups []string                    `json:"securityGroups"`
+	UpstreamSpec           *eksv1.EKSClusterConfigSpec `json:"upstreamSpec"`
+	VirtualNetwork         string                      `json:"virtualNetwork"`
+	Subnets                []string                    `json:"subnets"`
+	SecurityGroups         []string                    `json:"securityGroups"`
+	PrivateCanUsePublicAPI *bool                       `json:"privateCanUsePublicAPI"`
```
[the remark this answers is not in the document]
rmweir: privateRequiresTunnel, tunnelfPrivate, or something similar and use inversed values.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -282,7 +282,25 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
+			// the cluster has private-only access
+			serviceToken, usePublicEndpoint := e.generateSATokenWithPublicAPI(cluster)
+			if usePublicEndpoint != nil {
+				cluster = cluster.DeepCopy()
+				cluster.Status.EKSStatus.PrivateCanUsePublicAPI = usePublicEndpoint
+				cluster.Status.ServiceAccountToken = serviceToken
+				return e.clusterClient.Update(cluster)
+			}
+			e.clusterEnqueueAfter(cluster.Name, enqueueTime)
+			return cluster, nil
```
rmweir, line 300: It would be more conventional to just eval the error and if it's not dns then return the error since I believe that is the only time usePublicEndpoint would be nil. That way the error is logged somewhere and the controller will automatically retry. You could also just create an error here and return it if you do not think the error would contain anything useful. Manual enqueues should be avoided unless needed.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -608,21 +619,34 @@ func (e *eksOperatorController) deployEKSOperator() error {
...
 	}
 
-	generator, err := token.NewGenerator(false, false)
+	netDialer := net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+
+	serviceToken, err := generateSAToken(cluster.Status.APIEndpoint, cluster.Status.CACert, accessToken, netDialer.DialContext)
+
```
rmweir, line 635: remove newline

rmweir (changes requested):

pkg/controllers/management/eks/eks_cluster_handler.go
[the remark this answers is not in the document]
cbron: What I mean is this is how long we have to wait to see if public fails ?

cbron (commented):
Makes sense to me, would like @rmweir approval first. Also how will this look in the UI ? I’m assuming the import command appears either way and you have to exit that box for VPC situation.

pkg/controllers/management/eks/eks_cluster_handler.go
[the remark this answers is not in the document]
thedadams: No. The DNS error occurs fairly quickly.

thedadams:
Yes, the import command will pop up, but it can be ignored by the user if they have setup the VPC correctly.

rmweir (approved):
LGTM

cbron (approved):

Issue #30484, opened 2020-12-15
Prometheus CRD requires two apply calls to update a new field


aiyengar2:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Get a fresh Rancher install and create a downstream cluster. Check the CRDs and see that prometheuses.monitoring.coreos.com exists as a CRD initially.

Install a newer version of the CRD with a new field (such as `disableMountSubPath`) e.g. 
`kubectl apply -f https://raw.githubusercontent.com/rancher/system-charts/dev-v2.5/charts/rancher-monitoring/v0.2.0/charts/operator-init/files/crd-prometheus.yaml`

(Note: You can alternatively update all the CRDs to the latest upstream via `kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/master/bundle.yaml`. Same result)

Try to create a Prometheus resource with the new field e.g.

```yaml
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  name: my-monitoring
  namespace: default
spec:
  image: rancher/prom-prometheus:v2.18.2
  storage:
    disableMountSubPath: true
```

**Result:**

The Prometheus that is created does not have the new field. However, on a second `kubectl apply -f` it magically has the new field. If you then delete and repeat, it again takes 2 tries to get it working.

**Other details that may be helpful:**

Initially came from investigating the bug post-workaround for https://github.com/rancher/rancher/issues/29149

I launched a [kind](https://github.com/kubernetes-sigs/kind) cluster, installed the older versions of the CRDs, added the Prometheus resource without the field, installed the newer versions of the CRDs, updated the Prometheus resource, and only one apply was needed.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5-head
- Installation option (single install/HA): single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): DigitalOcean
- Machine type (cloud/VM/metal) and specifications (CPU/memory): Various (e.g. you could do two 8GB-4CPU-16GB Droplets with one agent and one server)
- Kubernetes version (use `kubectl version`): v1.19.4 (also happened in k8s 1.16)

```
(paste the output here)
```

- Docker version (use `docker version`): 19.3.14

```
(paste the output here)
```

aiyengar2:
[...]

Issue #29149, opened 2020-09-25
Request to add disableMountSubPath option in Prometheus operator (monitoring system chart)


suryatejaboorlu:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement 

**Steps to reproduce (least amount of steps as possible):**
* On the NFS server, export a shared folder accessible by a specific user (Say username=kubeusr, uid=2001, gid=3001).
* Ensure the specific user exists on the nodes of the cluster.
* Mount the NFS share on the nodes of the cluster at `/pv`.
* Deploy local-path-provisioner (https://github.com/rancher/local-path-provisioner/tree/v0.0.17#installation) using `/pv`. This installation will take care of installing the StorageClass too.
* In the enable/disable page of monitoring, enable persistent storage using the local-path-storage class, and use the following 
answers under the advanced options: 
 prometheus.securityContext.runAsUser = 2001
 prometheus.securityContext.runAsGroup = 3001
 prometheus.securityContext.runAsNonRoot = true

**Result:**
Prometheus operator container will not succeed to run, and the logs point to "permission" errors.  

**Root Cause**
https://github.com/rancher/local-path-provisioner/issues/4

**Environment information**
- Rancher version: v2.3.5
- Installation option (single install/HA): Single and HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom 
- Machine type (cloud/VM/metal) and specifications (CPU/memory): Cloud
- Kubernetes version : v1.16.15

aiyengar2:
As noted in https://github.com/rancher/system-charts/pull/345#issuecomment-698650663, the Prometheus Operator version that contains the ability to `disableMountSubPath` is only supported in k8s 1.16+. Since there is no Rancher release that exclusively supports k8s 1.16+ at the moment and we only have the ability to scope system charts to Rancher versions (as opposed to having something like a min kubernetes version), we can't support this change at the moment.

I'm closing out https://github.com/rancher/system-charts/pull/345 and https://github.com/rancher/image-mirror/pull/40 till we have the ability to support different system charts for different versions of k8s.

aiyengar2:
https://github.com/rancher/rancher/issues/29346 tracks support for this in 2.5 and https://github.com/rancher/rancher/issues/29867 tracks the backport to add support for this in 2.4

suryatejaboorlu:
I've reproduced the issue as in the instructions above and found the same permission error as below:

[...]

PR #30485, opened 2020-12-15
Upgrade Prometheus Operator to v0.44.1

aiyengar2:
Need to upgrade the Prometheus Operator version so that https://github.com/rancher/rancher/issues/30484 doesn't happen when installing the new Monitoring chart.

See note in https://github.com/rancher/rancher/issues/30484#issuecomment-745642510.

Related Issues: https://github.com/rancher/rancher/issues/30484, https://github.com/rancher/rancher/issues/29149


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.14
 replace (
 	git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible
+	github.com/coreos/prometheus-operator => github.com/coreos/prometheus-operator v0.39.0
```
mrajashree, line 8: why do we need this in the replace block? can we directly update line 56

aiyengar2: We could directly update it, but I believe it would require changes over a number of different files that import it. Since it's just a name change of the same underlying repository, I thought it made more sense to use a replace block instead of propagating those changes to files that can be left untouched.

I don't feel too strongly about it though, so if it would be better to make those changes I can re-run go generate

dramich: Couple of comments:
1. Why not just fixup the imports? I think that is the better option, if the upstream changed we should just change as well. 
2. Why are we updating to such an old version? 0.44.1 is the most recent, vs 39 which was released in may

aiyengar2: @dramich I can update to 0.44.1, but I was trying to make the minimal update required for the original issue with the Monitoring v0.2.0 chart (which is just going to version 0.39.0).

Will update in the next commit, and will fixup the imports when I run go generate this time.

dramich (approved):

PR #30509, opened 2020-12-16
[2.5] Revert sort app

StrongMonkey:


StrongMonkey:
Reverting the hack to sort based on appVersion since changes are made through charts https://github.com/rancher/charts/pull/897

StrongMonkey:
@dramich I reverted this change since we still need to generate the right version otherwise it will break airgap image.txt. So the fix is patched in charts.

dramich (approved):

PR #30658, opened 2020-12-31
Allow automation to run on cis hardened clusters

rancher-max:
I've built public image: `maxross/mytestcontainer:unprivileged` with the Dockerfile here. 

When running on a CIS hardened cluster, it should be run with the following parameters:
```
PYTEST_OPTIONS=-k "test_wl or test_dns_record or test_rbac or test_connectivity or test_ingress or test_secrets or test_service_discovery or test_websocket"
RANCHER_TEST_IMAGE=maxross/mytestcontainer:unprivileged
RANCHER_TEST_IMAGE_NGINX=nginxinc/nginx-unprivileged
RANCHER_TEST_IMAGE_PORT=8080
RANCHER_HARDENED_CLUSTER=True
RANCHER_SKIP_PING_CHECK_TEST=True
```
Note PYTEST_OPTIONS may include other tests as well, but I've listed what works for BOTH rke2 and rke1. For rke1, you should be able to run bkp_restore tests and CIS tests as well, though running with those options has not been fully validated with the changes here.
For rke2, run `test_ingress_xip_io` instead of `test_ingress`

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -41,8 +41,11 @@ kube_fname = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                           "k8s_kube_config")
 MACHINE_TIMEOUT = float(os.environ.get('RANCHER_MACHINE_TIMEOUT', "1200"))
 
+HARDENED_CLUSTER = ast.literal_eval(
+    os.environ.get('RANCHER_HARDENED_CLUSTER', "False"))
 TEST_OS = os.environ.get('RANCHER_TEST_OS', "linux")
 TEST_IMAGE = os.environ.get('RANCHER_TEST_IMAGE', "sangeetha/mytestcontainer")
+TEST_IMAGE_PORT = os.environ.get('RANCHER_TEST_IMAGE_PORT', "80")
 TEST_IMAGE_NGINX = os.environ.get('RANCHER_TEST_IMAGE_NGINX', "nginx")
 TEST_IMAGE_OS_BASE = os.environ.get('RANCHER_TEST_IMAGE_OS_BASE', "ubuntu")
 if TEST_OS == "windows":
@@ -950,7 +953,8 @@ def validate_cluster(client, cluster, intermediate_state="provisioning",
         path = "/name.html"
         rule = {"host": host,
                 "paths":
-                    [{"workloadIds": [workload.id], "targetPort": "80"}]}
+                    [{"workloadIds": [workload.id],
+                      "targetPort": TEST_IMAGE_PORT}]}
         ingress = p_client.create_ingress(name=name,
                                           namespaceId=ns.id,
                                           rules=[rule])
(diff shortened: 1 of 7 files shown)


tests/validation/tests/v3_api/common.py
```diff
@@ -990,30 +994,36 @@ def check_cluster_state(etcd_count):
...
+def validate_dns_entry(pod, host, expected, port=TEST_IMAGE_PORT):
     if is_windows():
         validate_dns_entry_windows(pod, host, expected)
         return
 
     # requires pod with `dig` available - TEST_IMAGE
-    cmd = 'ping -c 1 -W 1 {0}'.format(host)
-    ping_output = kubectl_pod_exec(pod, cmd)
+    if HARDENED_CLUSTER:
+        cmd = 'curl -vs {}:{} 2>&1'.format(host, port)
```
jiaqiluo, line 1011: why can not we use `ping` for the hardened cluster?

jiaqiluo (changes requested):

tests/validation/tests/v3_api/common.py
rancher-max: The `ping` utility requires the binary to be owned by root, and since we have to run this image specifically as unprivileged -- not root -- it means ping doesn't actually work in it

rancher-max: We can do ownership changes but it still fails on the ping socket unfortunately, so we just use curl in this case

jiaqiluo: ok, it makes sense. Thanks for the explanation.

tests/validation/tests/v3_api/common.py
```diff
@@ -1256,22 +1266,32 @@ def check_connectivity_between_workload_pods(p_client, workload):
...
+        assert pod_ip in str(response)
     if allow_connectivity:
         if is_windows():
             assert " (0% loss)" in str(response)
         else:
-            assert " 0% packet loss" in str(response)
+            if HARDENED_CLUSTER:
+                assert " 200 OK" in str(response)
+            else:
+                assert " 0% packet loss" in str(response)
```
jiaqiluo, line 1286: Can we simplify the logic to 
```
if is_windows(): 
	...
elif HARDENED_CLUSTER:
	...
else:
	...
```

tests/validation/tests/v3_api/common.py
```diff
@@ -1256,22 +1266,32 @@ def check_connectivity_between_workload_pods(p_client, workload):
...
     else:
         if is_windows():
             assert " (100% loss)" in str(response)
         else:
-            assert " 100% packet loss" in str(response)
+            if HARDENED_CLUSTER:
+                assert " 200 OK" not in str(response)
+            else:
+                assert " 100% packet loss" in str(response)
```
jiaqiluo, line 1294: the same as the above. 
Can we simplify the logic to 
```
if is_windows(): 
	...
elif HARDENED_CLUSTER:
	...
else:
	...
```

jiaqiluo (changes requested):

jiaqiluo:
> For rke2, run test_ingress_xip_io instead of test_ingress

Why does `test_ingress` not work on rke2?

rancher-max:
> > For rke2, run test_ingress_xip_io instead of test_ingress
> 
> Why does `test_ingress` not work on rke2?

@jiaqiluo In the `validate_ingress` function called for the other `test_ingress` tests, it validates the ingress is available from all nodes: `nodes = get_schedulable_nodes(cluster, os_type="linux")`. However, this is only the case when the ingress provider (nginx-ingress in most of our cases) is deployed as a daemonset or a deployment but tainted for all nodes. In rke2's case, it is deployed only as a deployment on the initial node, so ingress should only be available from the initial node. 

I have created https://github.com/rancher/rke2/issues/649 to address this, but if this functionality remains in rke2 then I'll make another update later on to get the ingress endpoints instead of schedulable nodes and iterate through that.

jiaqiluo (approved):
LGTM

tests/validation/tests/v3_api/common.py
```diff
@@ -1256,20 +1266,28 @@ def check_connectivity_between_workload_pods(p_client, workload):
...
     if is_windows():
         cmd = 'ping -w 1 -n 1 {0}'.format(pod_ip)
+    elif HARDENED_CLUSTER:
+        cmd = 'curl -I {}:{}'.format(pod_ip, TEST_IMAGE_PORT)
+    else:
+        cmd = "ping -c 1 -W 1 " + pod_ip
 
     response = kubectl_pod_exec(pod1, cmd)
-    assert pod_ip in str(response)
+    if not HARDENED_CLUSTER:
```
sowmyav27, line 1277: Do we need checks for when HARDENED_CLUSTER=True ?

sowmyav27 (changes requested):

tests/validation/tests/v3_api/common.py
rancher-max: We are just relying on the below `allow_connectivity` (or disallow) checks in this case. This is simply because `ping` returns the IP that a domain name resolves to but `curl` does not, and there isn't a similar check in the case of `curl` so I'm just skipping it.

sowmyav27: okay.

sowmyav27 (approved):
LGTM

rancher-max:
Recent push was to change maintainer of the docker image to myself

sangeethah:
LGTM

sangeethah (approved):
LGTM

PR #30668, opened 2021-01-04
Add LKE as a cluster driver

phillc:


pkg/data/management/kontainerdriver_data.go
```diff
@@ -100,6 +100,16 @@ func addKontainerDrivers(management *config.ManagementContext) error {
 	); err != nil {
 		return err
 	}
+	if err := creator.addCustomDriver(
+		"linodekubernetesengine",
+		"https://github.com/linode/kontainer-engine-driver-lke/releases/download/v0.0.3/kontainer-engine-driver-lke-linux-amd64",
+		"02fa95d24a1c6f9c520307e24a543c1777ed21fc3a4f060434e067806578e647",
+		"https://github.com/linode/ui-cluster-driver-lke/releases/download/v0.0.2/component.js",
+		false,
+		"",
```
dramich, line 109: It looks like the UI is doing calls directly to the API, these should be proxied through rancher and a host would need to be added here.

dramich (changes requested):
The build is failing due to a test needing to be updated. See https://github.com/rancher/rancher/commit/e57d0aef40b82304a8a7af9e39af6c11b235d8b2 as for what to update.

phillc:
I've fixed the api requests to go through the proxy. Luckily was able to piggyback the work from the node driver https://github.com/rancher/ui/blob/master/lib/shared/addon/linode/service.js

I haven't made the release for the component.js yet, as my manual testing passes on Rancher 2.4.11, but not master-head due to `Cluster health check failed: cluster agent is not ready` so I need to dig a little deeper.

dramich:
@phillc Sounds good, just ping me when you are ready.

dramich (approved):

PR #30673, opened 2021-01-05
Add ingressClass to globalDNS ingress if the provider has the annotationFilter for it specified

prachidamle:
Adding ability to filter out globalDNS entries when  there two globalDNSProviders having same root domain (one for public hostedzone and other for private), by using an annotation-filter parameter of external-dns.

By adding an "annotation-filter" to the globalDNSProvider a user can specify which ingress will this provider look for - rancher will place that annotation on the ingresses created for the globalDNS entries using this provider.

### DIFF
--- a/pkg/controllers/management/globaldns/globaldns_handler.go
+++ b/pkg/controllers/management/globaldns/globaldns_handler.go
@@ -27,18 +27,22 @@ import (
 const (
 	GlobaldnsController    = "mgmt-global-dns-controller"
 	annotationIngressClass = "kubernetes.io/ingress.class"
+	annotationFilter       = "annotation-filter"
 	annotationDNSTTL       = "external-dns.alpha.kubernetes.io/ttl"
+	defaultIngressClass    = "rancher-external-dns"
 )
 
 type GDController struct {
-	ingresses         clientv1beta1.IngressInterface //need to use client-go IngressInterface to update Ingress.Status field
-	managementContext *config.ManagementContext
+	ingresses               clientv1beta1.IngressInterface //need to use client-go IngressInterface to update Ingress.Status field
+	managementContext       *config.ManagementContext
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/globaldns/globaldns_handler.go
```diff
@@ -176,6 +189,51 @@ func (n *GDController) generateNewIngressSpec(globaldns *v3.GlobalDns) *v1beta1.
...
+	if options != nil {
+		ingressClassAnnotation, ok := options[annotationFilter]
+		if ok {
+			prefix := annotationIngressClass + "="
+			if strings.HasPrefix(ingressClassAnnotation, prefix) {
+				return strings.TrimPrefix(ingressClassAnnotation, prefix), nil
+			}
+		}
+	}
+	return defaultIngressClass, nil
```
mrajashree, line 225: question, as per [this doc](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/public-private-route53.md) linked in the issue, the annotation-filter's value can only have external-dns or internal-dns as the ingress class, so if the annotation is not present, why do we have to return rancher-external-dns as the ingress class? does it get picked as external or internal?

prachidamle: The annotation-filter's value can be any thing that user wants to put. It just means that external-dns will only choose the ingresses that carry this annotation

Now we  add this annotation `kubernetes.io/ingress.class: rancher-external-dns` to all the dummy ingresses we create as part of GlobalDNS controller so that our Rancher's nginx ingress-controller does not program it by mistake.
So now if user provides an annotation filter for the provider, we place that or if user does not we default to `kubernetes.io/ingress.class: rancher-external-dns`

dramich (approved):

cbron:
Is this good to merge ?

Issue #30606, opened 2020-12-23
forwardport: Removing in-tree vSphere vCenter server entries does not propagate


deniseschannon:
forwardport: https://github.com/rancher/rancher/issues/27306

rawmind0:
Submitted PR #30694  to fix the issue

sadiapoddar:
Verified the fix on 2.5-head, commit id, 23a29c30

1. Created a cluster on vSphere cloud provider by choosing the "custom" provider option and added the clouder provider information in YMAL file

2. Once the cluster is up checked all the config info and verified info. 
root@sadia-control1:# docker inspect kube-apiserver |grep cloud

3. Edited the cluster configuration and added a second non-existing vCenter information in the cluster config and saved it
<img width="843" alt="102672878-f17dc600-4146-11eb-9f71-61acb5f28a66" src="https://user-images.githubusercontent.com/72409237/105412308-69795900-5be9-11eb-8efc-e79b56e75e47.png">

4. Noticed a health error during updating the cluster with the newly added Center. 

5. Re-edited the cluster config to remove the newly added vCenter info 
6. Checked after saving it 
The cloud configuration doesn't have the second vCenter information in the config
<img width="669" alt="Screen Shot 2021-01-21 at 1 04 47 PM" src="https://user-images.githubusercontent.com/72409237/105412371-7d24bf80-5be9-11eb-8667-89190ce6a74f.png">

7. Cluster became healthy. 
<img width="1349" alt="Screen Shot 2021-01-21 at 1 01 54 PM" src="https://user-images.githubusercontent.com/72409237/105412005-fff94a80-5be8-11eb-8ae0-35bfdb39b85f.png">

Closing the bug as vCenter's in-tree server entries are propagating as expected

sowmyav27:
Reopening this for a PR for master-head

sadiapoddar:
The fix is not merged in the master branch.

rawmind0:
Submitted PR to master, https://github.com/rancher/rancher/pull/30948

sadiapoddar:
Verified the fix on master-head, 41045a8

- Created a cluster on vSphere cloud provider by choosing the "custom" provider option and added the clouder provider information in YMAL file

- Once the cluster is up checked all the config info and verified info.

- Edited the cluster configuration and added a second non-existing vCenter information in the cluster config and saved it
- Noticed a health error during updating the cluster with the newly added Center.

- Re-edited the cluster config to remove the newly added vCenter info
[...]

PR #30694, opened 2021-01-06
[2.5] Fix to properly remove vSphere vCenter server entries

rawmind0:
Fix issue https://github.com/rancher/rancher/issues/30606

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -538,6 +538,25 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data 
 	if err := r.validateUnavailableNodes(data, existingCluster, id); err != nil {
 		return nil, err
 	}
+
+	// Replace rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter value to properly update virtualCenter removal
+	if newVCenter, ok := values.GetValue(data, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter"); ok && newVCenter != nil {
+		oldVCenter, ok := values.GetValue(existingCluster, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter")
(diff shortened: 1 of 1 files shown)


superseb (commented):
So we have to reset (set to nil) before we can store the actual value, this means we are calling `store.Update` twice but this is unavoidable because the scope of changing `store.Update` is too big I assume?

Did we find the code that is responsible for this behavior? I assume it is because it finds an existing key and then merges values under it?

Last question would be if there is a specific reason why we didn't add a `setX` function like we use for the other "manipulations" before updating versus doing this directly in the flow as it is now.

rawmind0:
> So we have to reset (set to nil) before we can store the actual value, this means we are calling `store.Update` twice but this is unavoidable because the scope of changing `store.Update` is too big I assume?

Yes, changing `store.Update` scope would be much bigger 

> Did we find the code that is responsible for this behavior? I assume it is because it finds an existing key and then merges values under it?

The issue is caused by [`r.Store.Update`](https://github.com/rancher/rancher/blob/master/pkg/api/norman/store/cluster/cluster_store.go#L541) function. 
Data is arriving fine to update function but the function is not overwriting empty virtual_center entries. Old and new values are merged instead of being replaced. 

This behaviour is caused by [VirtualCenter](https://github.com/rancher/rke/blob/91aed199f04cc345275f6e694b0859f967fb92bd/types/rke_types.go#L652)  field definition with `omitempty` tag and the [generic function](https://github.com/rancher/norman/blob/master/clientbase/ops.go#L205) used by `store.Update` 

> Last question would be if there is a specific reason why we didn't add a `setX` function like we use for the other "manipulations" before updating versus doing this directly in the flow as it is now.

The main reason to not use `setX` function is notice that this process is distinct than others due to it's setting some values on `data` but it's also calling `store.Update` function if needed, affecting the flow. It can be implemented as `setX` function if you think is better.

superseb (approved):
I think this should be fine, maybe not ideal but with keeping the scope small and the behavior of Update, this is probably the best we can do. I think Kinara has experience in this part so she will have input on this.

cbron (approved):

Issue #30134, opened 2020-11-18
Support k8s 1.20


maggieliu:
Need to support the new k8s version 1.20

StrongMonkey:
There are Azure library changes when bumping k8s 1.20 in rancher, make sure azureAD auth is tested to avoid any regression bug.

StrongMonkey:
Related changes:

- [ ] Rancher hyperkube is updated to v1.20.2. Hyperbase is updated to v0.0.2. https://github.com/rancher/hyperkube/pull/168 https://github.com/rancher/hyperkube/pull/167
- [ ] Etcd version is updated to 3.4.14.
- [ ] Corednd is updated to 1.8.0. (Due to https://github.com/coredns/coredns/issues/4418 we can't update to latest version 1.8.0 until this is fixed)
- [ ] Calico and canal CNI is updated to v3.17.1
- [ ] Weave CNI is updated to 2.8.1
- [ ] Metric-server is updated to v0.4.1
- [ ] Ingress-Nginx-controller is updated to v0.43.0. https://github.com/rancher/ingress-nginx/pull/37 (Drone build is failing)

Available to test on this fork. 

https://raw.githubusercontent.com/StrongMonkey/kontainer-driver-metadata/k8s.120/data/data.json

sowmyav27:
Will wait for the fixes to be merged to dev-v2.5 to test. Reopening it.

StrongMonkey:
v1.20.2-rancher2-1 is available to test on 2.5-head/2.5 RCs.

sowmyav27:
**On 2.5.6-rc4**

- Deployed clusters with k8s v1.20.2-rancher2-1 and using different network providers.
- The version of images seen:
```
- rancher/hyperkube - v1.20.2-rancher2
- etcdctl version: 3.4.14
- rancher/coredns-coredns:1.8.0
- rancher/calico-cni:v3.17.1
- rancher/calico-node:v3.17.1
- rancher/metrics-server:v0.4.1
- rancher/nginx-ingress-controller:nginx-0.43.0-rancher1
- weaveworks/weave-kube:2.8.1
```
- Network checks on all 4 clusters look good.

StrongMonkey:
Calico and Canal is upgraded to the latest version v3.17.2. Please verify this also.

StrongMonkey:
Kubectl version in shell(ember and dashboard) will also be updated to v1.19.7. Please verify that later.

sowmyav27:
[...]

PR #30757, opened 2021-01-08
added kube version restrictions to rancher chart

paynejacob:


chart/Chart.yaml
```diff
@@ -3,6 +3,7 @@ name: rancher
 description: Install Rancher Server to manage Kubernetes clusters across providers.
 version: %VERSION%
 appVersion: %APP_VERSION%
+kubeVersion: < 1.20.0
```
cbron, line 6: How are we going to keep track of this over time ? Question for wider team.

deniseschannon: As part of the introducing new minor versions into Rancher, we should add this as a step.

deniseschannon: @StrongMonkey As part of adding k8s 1.20 into Rancher v2.5 line, please update this kubeVersion. Can you also add it into the process docs about introducing new k8s versions. 

@paynejacob Can you please make a PR into the release/v2.5 branch and master branch?

deniseschannon (approved):

cbron (approved):

PPCM:
What is the goal of this restriction?
Currently k3s has a kubversion of 1.20 and it is not possible to install Rancher on fresh install since this morning
We have to downgrade k3s version to 1.19.5

deniseschannon:
It's exactly what you pointed out. Rancher doesn't support k8s 1.20 which is why we have added the k8s version restriction to the helm chart.

PPCM:
OK thanks, I didn't see explanation in the release note, I only saw the restriction without reason

guemidiborhane:
will rancher support k8s >= 1.20 someday?

superseb:
It is tracked in https://github.com/rancher/rancher/issues/30134

kubealex:
what is exactly tracking aside from some tests that should be made on azuread? I mean, can we help in any way to move forward?

cbron:
@kubealex #30134 is tracking adding Rancher support for Kubernetes 1.20. If you have a problem with the K8s version restrictions on the helm chart make sure you are on Rancher v2.5.5 and that the K8s version is <=1.19.x

PR #30761, opened 2021-01-08
[v2.5] added kube version restrictions to rancher chart

paynejacob:


paynejacob:
conflicts with https://github.com/rancher/rancher/pull/30776/files

PR #30762, opened 2021-01-08
added kube version restrictions to rancher chart

paynejacob:


superseb (approved):

GProulx:
Hi, this modification seems to to that bug: rancher/k3d/issues/459 

Can you explain why that limitation was added?

vincent99:
Because k8s 1.20 isn't supported by the current Rancher version, and has things that actually don't work correctly.

GProulx:
Thanks @vincent99  I will try to downgrade k3s then and should be good to go.  Will update the related issue too if it works.

Steven-Beasley:
It's strange that the latest v4.0 of k3d would use 1.20 if rancher doesn't support it yet.

GProulx:
I totally agree!  Many some unexpected problems that show up late in the release process.

vincent99:
K3d follows k3s releases and k3s follows kubernetes ones.  None of those are tightly-coupled to Rancher (the product) releases, nor are they necessarily used together.

Rancher has large enterprise users which can't necessarily patch frequently, so it is updated a little more conservatively.

Most of the time this doesn't _really_ matter, because it will work anyway, but in this case 1.20 drops an API field that Rancher uses. So this change was to let you know earlier instead of installing and then finding out.  Until a version that does support 1.20 is available.

Issue #30291, opened 2020-10-15
Switch to using the upstream saml library directly instead of rancher's fork of it


mrajashree:
Rancher uses a fork of crewjam/saml library maintained at https://github.com/rancher/saml
It was forked earlier because the upstream library was not working for PingFederate, so the fork contained changes for that. The library's latest version has the required changes for pingFederate, and also has changes for some other saml issues/requests. 

- [ ] Switch to using the upstream library, stop maintaining & using our fork

- [ ] Test that this change works with all existing SAML providers: Shibboleth, PingFederate, Keycloak, ADFS

cardoso81priscila:
#26195 
![20201130_163940](https://user-images.githubusercontent.com/75869019/103001893-a24fe080-450c-11eb-8adb-5871f05da0e6.jpg)
@cardoso81priscila

mrajashree:
As discussed offline, we should also check if IDP initiated SSO works with the latest library changes

izaac:
Working on 2.5-head but we also need this in master.

izaac:
Rancher version: `v2.5-head` _4cb2ebe_

P1 tests done in the following auths

- Keycloak
- Okta
- Shibboleth

Found this issue https://github.com/rancher/rancher/issues/31163
But it seems pre-existing: https://github.com/rancher/rancher/issues/31163#issuecomment-775090599


Work on master-head in progress.

izaac:
Rancher version: `master-head` _86600eaff_

P1 tests done in the following auths

- Keycloak
- Okta
- Shibboleth

izaac:
Rancher version `v2.5-head (02/12/2021)` _dba3321_

Upgrade path `v2.5.5` Keycloak enabled to `v2.5-head`.

Spot checked authentication and authorization still works after upgrade.

izaac:
Rancher version `v2.5-head (02/24/2021)` _30e53bd_
Rancher version `master-head (02/24/2021)` _465bb3b_

P1 tests done using ADFS.

izaac:
Found an issue with groups membership assignments on Ping
https://github.com/rancher/rancher/issues/31519

izaac:
Rancher version `v2.5-head (02/26/2021)` _67228a3_
Rancher version `master-head (02/26/2021)` _465bb3b_

Ping

Validated P1 test cases of user access and login.

- Three different levels of site access.
- Cluster Access
- Project Access




[...]

Issue #30306, opened 2020-12-02
ADFS Auth fails with Invalid SAML Attributes


leflambeur:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Enable auth with AD then disable and enable auth with ADFS

**Result:**
Invalid SAML Attribute messages, SAML Trace shows that SAML Attributes are in fact fine

Rancher logs indicate a failure to get UPN (also fine in the SAML trace)

No other info is shown to be discovered and the environment has been validated several times for correct settings

**Other details that may be helpful:**
Debug logging in Rancher showed no extra information, equivalent environments configured with ADFS OOTB work

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.4.8

gz#12724

paynejacob:
Validated on 238e9f21be0554b8c237a1226e337bea0a972c56.  I first configured AD auth and was able to log in.  AD auth was disabled and ADFS was configured.  I was able to login with an ADFS user.

anupama2501:
On 2.5.5 was able to reproduce.
From the global --> authentication --> enable AD auth --> disable AD ---> Enable the ADFS auth and when we try to save, it fails with the error "Invalid SAML Attributes"

anupama2501:
On v2.5-4e48b8a9fbedad498fa21fd00e6e426ab3e13770-head
From Auth providers --> Enable AD and save --> Disable AD and enable ADFS by providing all the details --> Enabling goes through without any error --> Logout and login with ADFS credentials --> Authentication goes through and users are allowed to login

anupama2501:
Changing the label to-test as @paynejacob has provided the commit ID for master-head https://github.com/rancher/rancher/pull/31106

anupama2501:
on master-2dca0c67b48331c23c05c025208bacfc258d45bd-head
From Auth providers --> Enable AD and save --> Disable AD and enable ADFS by providing all the details --> Enabling goes through without any error --> Logout and login with ADFS credentials --> Authentication goes through and users are allowed to login
Log in as standard user --> Able to create a cluster

anupama2501:
Rancher before upgrade v2.5.6
After upgrade : v2.5-fa087c81628de057406608f077d0952bb03cea56-head

Upgrade testcase:

- Create rancher server on v2.5.6
[...]

PR #30858, opened 2021-01-15
[2.5] switch to upstream saml library

paynejacob:
#30291 

PROBLEM:

Rancher was maintaining a fork of crewjam/saml to fix a since resolved issue.  This has put us behind the latest version of the upstream library.  Now that the original issue is resolved we can switch back to the upstream.

SOLUTION:

Switched back to the upstream library in `go.mod`.  In the latest version of crewjam/saml the generic cookie database interface was remove in favor of a more concise approach for storing data. [link](https://github.com/crewjam/saml/commit/695c7b1abe5223dd32db3fa60915b58116880b29) Rancher depended on this generic interface to store some extra rancher specific data. I pulled the removed code into the rancher saml package.

TESTING:
I tested this with a keycloak setup and was able to configure keycloak and log in as a keycloak user.


superseb (changes requested):
Build failure

superseb (approved):

paynejacob:
#30306

superseb (approved):

luthermonson (approved):

PR #30955, opened 2021-01-22
Deploy secret with more permissions for hardened

rancher-max:
Allow non-root user to access the secret value in the case of hardened cluster


rancher-max:
This isn't a necessary change

PR #30971, opened 2021-01-24
Ability to have rancher itself listen on https with certificates rather than only http

mrproper:
- add cert.pem key.pem volumes to rancher deployment
- make all tls secrets configurable via tlsSecrets rather than hardcoded and default them to the existing locations
- move tls: ingress|external to tls.offload to make more sense (this will obviously affect anyone using tls: "external"


KittyKatt:
Would there be any interest in getting this rebased on current code to allow it to be merged? This code would greatly help us with a customer that wants to be able to do E2E SSL in their clusters.

PR #31022, opened 2021-01-28
Fix k3s upgrade dev image export

rmweir:
Prior, using export script with TAG set to dev, master, or a
value equal to kd.RancherVersionDev would result in no k3s
upgrade or related images being added to the images text.
This is because kd.RancherVersionDev does not contain a
valid semver version. This happens to dev and master because
those values are overwritten with the value of
kd.RancherVersionDev if detected. Now, versions are used for
comparison that match the minor and major version of
kd.RancherVersionDev and will match any constraints that
could be met by any version that would also have matching
major and minor versions.

### DIFF
--- a/pkg/image/export/main.go
+++ b/pkg/image/export/main.go
@@ -267,35 +267,46 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{}) 
 			continue
 		}
 
-		if rancherVersion != "dev" {
-			maxVersion, _ := releaseMap["maxChannelServerVersion"].(string)
-			maxVersion = strings.TrimPrefix(maxVersion, "v")
-			if maxVersion == "" {
-				continue
-			}
-			minVersion, _ := releaseMap["minChannelServerVersion"].(string)
-			minVersion = strings.Trim(minVersion, "v")
-			if minVersion == "" {
-				continue
-			}
+		maxVersion, _ := releaseMap["maxChannelServerVersion"].(string)
+		maxVersion = strings.TrimPrefix(maxVersion, "v")
+		if maxVersion == "" {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


pkg/image/export/main.go
```diff
@@ -267,35 +267,44 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{})
...
-			maxVersion = strings.TrimPrefix(maxVersion, "v")
-			if maxVersion == "" {
-				continue
-			}
-			minVersion, _ := releaseMap["minChannelServerVersion"].(string)
-			minVersion = strings.Trim(minVersion, "v")
-			if minVersion == "" {
-				continue
-			}
+		maxVersion, _ := releaseMap["maxChannelServerVersion"].(string)
```
nickgerace, line 270: Is there value in at least logging the error if we encounter it?

pkg/image/export/main.go
```diff
@@ -267,35 +267,44 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{})
...
+		maxVersion = strings.TrimPrefix(maxVersion, "v")
+		if maxVersion == "" {
+			continue
+		}
+		minVersion, _ := releaseMap["minChannelServerVersion"].(string)
```
nickgerace, line 275: Same for here as above

pkg/image/export/main.go
rmweir: This is usually ran in a CI process and since it doesn't cause a build failure it likely won't be noticed but can't hurt.

rmweir: @nickgerace actually what I said applies to the actual errors but this is just an ok check to keep it safe.

nickgerace: Sounds goo d to me. Thanks

PR #31025, opened 2021-01-28
Change dev version to be 2.6

StrongMonkey:


StrongMonkey:
@kinarashah Is it safe to change the dev version to be 2.6? What I am seeing in the current behavior is that if the k8s min rancher version is 2.5.6-rc0, rancher will skip this version since 2.5 < 2.5.6-rc0 in the current compare logic

kinarashah:
@StrongMonkey yep, it should be safe to update it for the master branch. do you see this behavior in `2.5-head` also?

StrongMonkey:
@kinarashah I haven't tested on the 2.5-head branch. Since it is sharing the same code I would expect it to have the same probelm.

kinarashah (approved):

PR #31034, opened 2021-01-28
Single click rke2-upgrade for jenkins

rancher-max:
This enables a single-click upgrade job that will optionally deploy a rancher server, import rke2 clusters of different versions, and then run preupgrade job, upgrade the clusters, and postupgrade job. 

It works for hardened as well, but will have a bit of failures due to the nature of those jobs failing expectedly in a hardened run.


bmdepesa:
Can you squash the commits on this?

ShylajaDevadiga (approved):
LGTM

bmdepesa (approved):
LGTM

PR #31037, opened 2021-01-28
Update rancher CLI version to v2.4.11-rc2

rmweir:


package/Dockerfile
```diff
@@ -109,7 +109,7 @@ RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
 
 ENV CATTLE_UI_VERSION 2.5.6-rc3
 ENV CATTLE_DASHBOARD_UI_VERSION v2.5.6-rc3
-ENV CATTLE_CLI_VERSION v2.4.10
+ENV CATTLE_CLI_VERSION v2.4.11-rc2
```
cbron, line 112: Why are we bumping this in 2.5 ? It should stay as 2.4.10.

cbron (approved):
lgtm, as long as this is rc with it reverted

PR #31041, opened 2021-01-28
Fix k3s upgrade dev image export

rmweir:
Prior, using export script with TAG set to dev, master, or a
value equal to kd.RancherVersionDev would result in no k3s
upgrade or related images being added to the images text.
This is because kd.RancherVersionDev does not contain a
valid semver version. This happens to dev and master because
those values are overwritten with the value of
kd.RancherVersionDev if detected. Now, versions are used for
comparison that match the minor and major version of
kd.RancherVersionDev and will match any constraints that
could be met by any version that would also have matching
major and minor versions.

**Issue:**
https://github.com/rancher/rancher/issues/31061

### DIFF
--- a/pkg/image/export/main.go
+++ b/pkg/image/export/main.go
@@ -267,35 +267,46 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{}) 
 			continue
 		}
 
-		if rancherVersion != "dev" {
-			maxVersion, _ := releaseMap["maxChannelServerVersion"].(string)
-			maxVersion = strings.TrimPrefix(maxVersion, "v")
-			if maxVersion == "" {
-				continue
-			}
-			minVersion, _ := releaseMap["minChannelServerVersion"].(string)
-			minVersion = strings.Trim(minVersion, "v")
-			if minVersion == "" {
-				continue
-			}
+		maxVersion, _ := releaseMap["maxChannelServerVersion"].(string)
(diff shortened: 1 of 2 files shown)


rmweir:
Closed old PR in an attempt to fix drone: https://github.com/rancher/rancher/pull/31022

pkg/image/export/main.go
```diff
@@ -267,35 +267,46 @@ func getK3sUpgradeImages(rancherVersion string, k3sData map[string]interface{})
...
-			if err != nil {
-				continue
-			}
-			if rancherVersion != minVersion && !versionGTMin {
-				// rancher version not equal to or greater than minimum supported rancher version
-				continue
-			}
+		var versionGTMin bool
+		var err error
+		if rancherVersion == kd.RancherVersionDev {
```
cbron, line 283: So `rancherVersion` would never be `dev` right ? It always gets translated to rancherVersionDev

cbron (approved):
Think it makes sense to me, but def need a QA test to ensure it doesn't break existing images.txt.

pkg/image/export/main.go
rmweir: Correct, as it is right now this function only gets used in one place where dev and master are translated.

cbron: ok cool

PennyScissors (approved):

Issue #30934, opened 2021-01-21
Add minFleetVersion to rancher


StrongMonkey:
Rancher auto-install/upgrade system charts(fleet,rancher-operator,rancher-webhook), which is not desirable. We need to add setting `min$(system-chart-name)Version` in rancher, so that if rancher detects that the current system has chart installed and meets minVersion, it should not attempt to upgrade.

slickwarren:
I was able to test the following on 2.5-head (03b8153) and master-head (3134a9a) on a single install cluster: 
* CATTLE_RANCHER_WEBHOOK_MIN_VERSION
* CATTLE_RANCHER_OPERATOR_MIN_VERSION
* CATTLE_FLEET_MIN_VERSION

**note:** each chart deploys with the latest available version by default when creating a new rancher server. I was able to install older versions of each chart by going to `cluster explorer` -> Marketplace & Apps, selecting the app to modify, and changing the version there. 
To trigger the min_chart_version, I stopped rancher from running (as if doing a backup), and reran with the environment variable set (-e <APP>_MIN_VERSION=X.Y.Z for single install, --set 'extraEnv[0].name=<APP>_MIN_VERSION' --set 'extraEnv[0].value=X.Y.Z' for HA installs.

the following tests were performed:
* set <app> version to lower than the specified min version -- result: upgrades to latest chart version -- PASS
* set <app> version to higher or equal to specified min version -- result: remains on old version -- PASS
* set min version to nonexistent version lower than <app> version -- result: remains on old version -- PASS
* set min version to nonexistent version higher than <app> version -- result: upgrades to latest version -- PASS
* set min version to  invalid entry (ex. x.0y.123) -- result: logs show `[ERROR] Failed to install system chart <app>: Invalid Semantic Version` -- PASS
* set min version to null -- result: logs show `[ERROR] Failed to install system chart <app>: Invalid Semantic Version` -- PASS
* set multiple app versions at once -- result: older and newer chart versions behaved as stated in the above cases -- PASS

slickwarren:
reopening for discussion (cc @sowmyav27 )

PR #31066, opened 2021-01-31
Add minRequiredVersion for fleet, rancher-operator and webhook

StrongMonkey:

### DIFF
--- a/pkg/catalogv2/system/system.go
+++ b/pkg/catalogv2/system/system.go
@@ -38,8 +38,9 @@ var (
 )
 
 type desiredKey struct {
-	namespace string
-	name      string
+	namespace  string
+	name       string
+	minVersion string
 }
 
 type desired struct {
@@ -104,7 +105,7 @@ func (m *Manager) runSync() {
 func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
 	for key, values := range charts {
 		for {
-			if err := m.install(key.namespace, key.name, values); err == repo.ErrNoChartName || apierrors.IsNotFound(err) {
+			if err := m.install(key.namespace, key.name, key.minVersion, values); err == repo.ErrNoChartName || apierrors.IsNotFound(err) {
 				logrus.Errorf("Failed to find system chart %s will try again in 5 seconds: %v", key.name, err)
 				time.Sleep(5 * time.Second)
 				continue
@@ -116,12 +117,13 @@ func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
 	}
 }
 
-func (m *Manager) Ensure(namespace, name string, values map[string]interface{}) error {
+func (m *Manager) Ensure(namespace, name, minVersion string, values map[string]interface{}) error {
 	go func() {
 		m.sync <- desired{
 			key: desiredKey{
-				namespace: namespace,
-				name:      name,
+				namespace:  namespace,
+				name:       name,
+				minVersion: minVersion,
 			},
 			values: values,
 		}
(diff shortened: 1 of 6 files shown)


pkg/catalogv2/system/system.go
```diff
@@ -284,6 +286,17 @@ func (m *Manager) isInstalled(namespace, name, version string, values map[string
 			return false, err
 		}
 
+		if minVersion != "" {
+			min, err := semver.NewVersion(minVersion)
+			if err != nil {
+				return false, err
+			}
+			if min.LessThan(ver) || min.Equal(ver) {
```
dramich, line 294: So how do you upgrade if you want to? Doesn't this mean that even if you upgrade rancher to a newer version these things will never upgrade?

StrongMonkey: oh right, I think this will not upgrade even if you upgrade rancher.. I wonder if we should just disable that and always require user to manually upgrade.

StrongMonkey: @dramich I am thinking that we can bump the version in here https://github.com/rancher/rancher/pull/31066/files#diff-c55d1ee74f32cd4b9eeeceaec981a180f52e761acb158a5b0b0e088c7725c9caR41 so that when upgrading rancher it will force to upgrade because min Version changed.

dramich (dismissed):

dramich (dismissed):

pkg/catalogv2/system/system.go
```diff
@@ -284,6 +286,17 @@ func (m *Manager) isInstalled(namespace, name, version string, values map[string
...
+				logrus.Infof("Skipping installing/upgrading desired version %v for release %v, current version %v is greater or equal to minimal required version %v", desired.String(), name, ver.String(), minVersion)
```
ibuildthecloud, line 295: This will be annoying to see in the logs every 15 minutes.

pkg/catalogv2/system/system.go
```diff
@@ -284,6 +286,17 @@ func (m *Manager) isInstalled(namespace, name, version string, values map[string
...
+				return true, nil
+			}
+		}
+
 		if (desired.LessThan(ver) || desired.Equal(ver)) && equality.Semantic.DeepEqual(values, release.Config) {
```
ibuildthecloud, line 300: The new logic is not considering the situation where we have changed the values file.  If the values has changed, but the version is sufficient, we should do an upgrade to the exact same version installed but with new values.

pkg/catalogv2/system/system.go
StrongMonkey: Will move this to debug.

StrongMonkey:
@dramich @ibuildthecloud Added upgrade current version logic. Also added a superset method so that we don't revert any setting for user.

pkg/catalogv2/system/system.go
```diff
@@ -176,7 +179,7 @@ func (m *Manager) install(namespace, name string, values map[string]interface{})
 		Charts: []types.ChartUpgrade{
 			{
 				ChartName:   name,
-				Version:     chart.Version,
+				Version:     desiredVersion,
```
ibuildthecloud, line 182: I'm not sure this logic is right.  If min version is 1 and nothing is install and version 2 is available, version 2 should be installed.

pkg/catalogv2/system/system.go
```diff
@@ -257,52 +259,76 @@ func podDone(chart string, newPod *corev1.Pod) (bool, error) {
...
+		if err != nil {
+			return false, "", nil, err
+		}
+
+		patchedJSON, err := jsonpatch.MergePatch(actualValueJSON, desiredValuesJSON)
+		if err != nil {
+			return false, "", nil, err
+		}
+
+		if err := json.Unmarshal(patchedJSON, &desiredValue); err != nil {
```
ibuildthecloud, line 302: You need to do `desiredValue = map[string]interface{}` otherwise you might be aggregating expected values.

ibuildthecloud (approved):

Issue #30869, opened 2021-01-15
Can not login as a standard user (who is part of 1000 groups) with AD auth without trying many times. 


Ellenqs:
**What kind of request is this (question/bug/enhancement/feature request):**
question/bug

**Steps to reproduce (least amount of steps as possible):**
1. Set up AD authentication
2. Create a standard user
3. login as standard user

**Result:**
Got login error many times before login successfully.

<img width="427" alt="Screen Shot 2021-01-15 at 11 47 53 AM" src="https://user-images.githubusercontent.com/75277058/104772339-1d7d6e80-5728-11eb-97b0-9720142b118d.png">

API error : 
```
{baseType: "error", code: "ServerError", message: "server error while authenticating", status: 500,…}
baseType: "error"
code: "ServerError"
message: "server error while authenticating"
status: 500
type: "error"
```

Rancher log:

```
2021/01/15 22:36:07 [ERROR] Unknown error: Error creating ssl connection: LDAP Result Code 200 "": x509: cannot validate certificate for ******** because it doesn't contain any IP SANs
2021/01/15 22:36:48 [INFO] Updating user user-4pwnt. Adding principal
2021/01/15 22:38:22 [INFO] Creating new GlobalRoleBinding for GlobalRoleBinding grb-qtl2c
2021/01/15 22:38:22 [INFO] [mgmt-auth-grb-controller] Creating clusterRoleBinding for globalRoleBinding grb-qtl2c for user u-4rg9h with role cattle-globalrole-user
2021/01/15 22:39:02 [ERROR] API error response 500 for POST /v3-public/activeDirectoryProviders/activedirectory?action=login. Cause: ldap: connection timed out
2021/01/15 22:39:33 [INFO] Creating user for principal activedirectory_user://CN=testuser2,CN=Users,DC=testad,DC=rancher,DC=io
2021/01/15 22:39:33 [INFO] Creating globalRoleBindings for u-npjx57vxal
2021/01/15 22:39:33 [INFO] Creating new GlobalRoleBinding for GlobalRoleBinding grb-pf8l7
2021/01/15 22:39:33 [INFO] [mgmt-auth-grb-controller] Creating clusterRoleBinding for globalRoleBinding grb-pf8l7 for user u-npjx57vxal with role cattle-globalrole-user
2021-01-15 22:39:45.555813 I | mvcc: store.index: compact 8110
[...]

PR #31067, opened 2021-02-01
We need not pull the memberof attribute of a group when listing groups during user login

prachidamle:
When a user logs in, we pull all the groups that he is a memberOf. While pulling details of each group, the attribute "memberOf" of a group that lists all the parent groups of this group, need not be synced from the AD as we only need the group DN and some important attributes.
This saves some time needed to sync the groups of a user during login, particularly significant if the user is part of large number(1000) of groups.

The query to gather parent groups when nested groups are enabled, remains unchanged by this fix.

https://github.com/rancher/rancher/issues/30869


aiyengar2 (approved):
LGTM

dramich (approved):

Issue #30594, opened 2020-12-22
[Catalog v2] can not change the url or branch of the builtin catalog `Rancher`


jiaqiluo:
**What kind of request is this (question/bug/enhancement/feature request):**
question / bug 
I am not sure if this is the expected behavior or built-in charts.

**Steps to reproduce (least amount of steps as possible):**
- run rancher:`v2.5-head`/`v2.5.3`  single-install
- go to the `cluster explorer` page of the local cluster 
- edit the builtin catalog `Rancher` to the same repo url and branch as the `partner` one, as the following screenshot shows 
![image](https://user-images.githubusercontent.com/6218999/102930959-b0452900-445a-11eb-83f7-2b6ea6bc4ec4.png)

**Result:**
Go to the `charts` page, we still see the original charts under the `Rancher` catalog even refreshing the catalog manually 
<img width="1860" alt="Screen Shot 2020-12-22 at 1 40 00 PM" src="https://user-images.githubusercontent.com/6218999/102931226-382b3300-445b-11eb-84f4-ccda3a84d3e4.png">

**Expected**:
I expected to see the same set of charts under both catalogs.

jiaqiluo:
The same behavior happens on the downstream cluster on v2.5.3 single-install setup.

StrongMonkey:
@jiaqiluo I think there are some caches in browser. If you switch to incognite mode I think both will show the same.

StrongMonkey:
Also we should not change the url for rancher charts. This will break the existing system charts.

slickwarren:
can confirm the charts don't refresh even after emptying the cache/opening in incognito mode
* rancher 2.5-head (2ae4cf1)
<img width="1792" alt="Screen Shot 2021-02-01 at 7 08 38 AM" src="https://user-images.githubusercontent.com/16691014/106476902-689fbd00-645c-11eb-8727-fa18af467672.png">

StrongMonkey:
@slickwarren I just tried 2.5-head and it is working for me... Do you have a setup that I can look at?

StrongMonkey:
@slickwarren Available to test on the latest 2.5-head.

StrongMonkey:
@slickwarren Make sure you tested the airgap. This touches the airgap code so we want to make sure it doesn't break airgap install(airgap install should have the right access to catalog still)

slickwarren:
[...]

PR #31074, opened 2021-02-01
Use unique url hash for built-in catalog git dir

StrongMonkey:


StrongMonkey:
This PR fixes https://github.com/rancher/rancher/issues/30594. But we should discuss the behavior that whether user is allowed to change the system charts url.

dramich (approved):

PR #31080, opened 2021-02-01
[master]Allow k8s 1.20 in rancher helm chart

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/30134

cbron (approved):
Thanks. We had forgot to forwardport old fix too :)

Issue #30406, opened 2020-12-09
Validate Docker 20.10.x


superseb:
https://docs.docker.com/engine/release-notes/#version-2010
https://github.com/moby/moby/releases/tag/v20.10.1

* Update `install-docker` script to latest and use `VERSION` environment variable to set version
* Validate all setups with Docker with the latest version

immanuelfodor:
In the meantime, I used `rke up --ignore-docker-version` as a temporal workaround to bring up the cluster after the Docker 20.10.0 update. It works fine on Centos Stream (upgraded from CentOS 8) with Docker packages from https://download.docker.com/linux/centos/8/x86_64/stable/Packages/ and `firewalld` disabled.

```bash
$ containerd -v
containerd containerd.io 1.4.3 269548fa27e0089a8b8278fc4fc781d7f65a939b

$ docker version
Client: Docker Engine - Community
 Version:           20.10.0
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        7287ab3
 Built:             Tue Dec  8 18:59:27 2020
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.0
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       eeddea2
  Built:            Tue Dec  8 18:57:25 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.3
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc:
  Version:          1.0.0-rc92
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

$ cat /etc/os-release 
NAME="CentOS Stream"
VERSION="8"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Stream 8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
```

[...]

PR #31125, opened 2021-02-03
[2.5] Add support to docker 20.10 and set it as default install script

rawmind0:
Issue https://github.com/rancher/rancher/issues/30406


tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
```diff
@@ -15,7 +15,7 @@
 AZURE_TENANT_ID = os.environ.get("AZURE_TENANT_ID")
 worker_count = int(os.environ.get('RANCHER_STRESS_TEST_WORKER_COUNT', 1))
 HOST_NAME = os.environ.get('RANCHER_HOST_NAME', "testcustom")
-engine_install_url = "https://releases.rancher.com/install-docker/19.03.sh"
+engine_install_url = "https://releases.rancher.com/install-docker/20.10.sh"
```
sowmyav27, line 18: Even though we dont usually run automation from this branch, we should not be changing values in files under tests/validation as we dont always run automation tests on 2.5.6+

sowmyav27, line 18: Even though we dont usually run automation from this branch, we should not be changing values in files under tests/validation as we dont always run automation tests on 2.5.6+

superseb (approved):

Issue #26390, opened 2020-03-31
rancher UI view/edit yaml , the request return error


hachi-zzq:
**What kind of request is this (question/bug/enhancement/feature request):**

view / edit a  namespace yaml , then the api return error 。

**Steps to reproduce (least amount of steps as possible):**

step 1 : select a workloads 
step 2: click view/edit yaml

**Result:**

the "/v3/cluster/c-br6qn/namespaces/xxx/yaml" api return "500 Internal Server Error" and return the response 
"baseType: error
code: ServerError
message: invalid character '\x1f' looking for beginning of value
status: 500
type: error" 

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
- Installation option (single install/HA):

rancher Version: rancher/rancher:v2.3.5      
Installation option: single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```
![image](https://user-images.githubusercontent.com/5241713/77981451-926bed00-733c-11ea-8257-84a74141958d.png)

![image](https://user-images.githubusercontent.com/5241713/77981456-98fa6480-733c-11ea-885b-edb069d9f470.png)

- Docker version (use `docker version`):

docker verison : Docker version 19.03.1, build 74b1e89
```
(paste the output here)
```

skaven81:
We are seeing the same thing at Qualcomm.  Exact same error, even the same invalid `\\x1f` character.  Rancher support ticket opened.

ranchersvet:
gz#12501

sudevva:
up. Exact same error, but only for big yaml's. Small yaml opens ok.

Leon-pan:
#30881 same problem

hachi-zzq:
i use kubectl CLI tool to get the yml

william0503:
I figure out that this error is occuring when Rancher API is compacting the yaml.

`2022-04-21 19:51:29.222170 I | mvcc: store.index: compact 100754132
2022-04-21 19:51:29.228664 I | mvcc: finished scheduled compaction at 100754132 (took 1.619617ms)
2022/04/21 19:53:08 [ERROR] Unknown error: invalid character '\x1f' looking for beginning of value`

[...]

PR #31131, opened 2021-02-04
[2.5] Fix unable to view namespace yaml

GGGitBoy:
**Problem:**

some namespace gets error `invalid character '\x1f' looking for beginning of value` when editing YAML.

**Solution:**

If the content of the resource is too large, the returned content will be compressed by gzip, add the request header `Accept-Encoding: identity` to prevent compression.

**Relate issue:**

https://github.com/rancher/rancher/issues/26390


Leon-pan (approved):
I think it can fix my problem.

william0503 (approved):

william0503:
My problem is to export the YAML file and this fix can solve it.
Currently, I'm doing a workaround to export the YAML file, using Postman to request rancher api, adding this header, and it works properly.

Issue #29932, opened 2020-11-04
Duplicate roleBindings being created


dramich:
@mrajashree has seen this in a downstream cluster and there have been at least two internal reports of seeing this in larger setups. 

It appears to affect multiple versions of rancher but at this point there are no clear steps to repro but needs to be investigated thoroughly to determine if rancher really is creating duplicate roleBindings.

There are multiple data sets that can be used to look at this which I have. 

gz#12526
gz#13400
gz#14235
gz#13826
gz#14439
gz#14985

gz#15411

mrajashree:
yup I added a user to a cluster as owner, and the CRTB created 3 CRBs in the downstream cluster for the same user with role `cluster-owner`, and same rbac label `authz.cluster.cattle.io/rtb-owner`

mrajashree:
I added a user to a cluster as owner, and the CRTB created 3 CRBs in the downstream cluster for the same user with role `cluster-owner`, and same rbac label `authz.cluster.cattle.io/rtb-owner` . 
We use generated names for those clusterrolebindings and that could be the reason.
This does not repro always, happened only once

Tejeev:
Hi @mrajashree, we have a cluster where this seems to happen often.  Please DM me if you would like access (customer has offered)

ryansann:
I am now able to reliably reproduce this issue with the following steps:

1. Deploy Rancher HA (3 replicas)
2. Create 3-5 users
3. Deploy downstream cluster
4. Add users as members to project=default in downstream cluster
5. Add users as members to downstream cluster
6. Dump the duplicate RoleBindings and ClusterRoleBindings
7. Restart rancher pods
8. Wait for startup / initialization
9. Dump duplicates again
10. Observe that the number of duplicates has increased

Duplicates appear to be created when cache retrieval operations via the indicies in the management/auth controllers don't return results for existing bindings. I am continuing to investigate.

deniseschannon:
After the design discussions, the proper fix for this is to move to a deterministic names instead of generated names.

anupama2501:
Able to reproduce this on v2.4.15 by following @Ryan steps:
1. On HA rancher server, create a three users from global --> security --> users
2. Create a cluster and add the users as members in the cluster [from default project]
3. From the kubectl commands, get the count of the role-bindings and cluster role-bindings
4. Restart the rancher pods
5. Once the cluster is active, from the same kubectl commands, observer the count of role-bindings and cluster role-bindings increase. [increased by 10] 

[...]

PR #31135, opened 2021-02-04
Add cleanup to the agent

dramich:
This will add the ability for the agent to be ran as a job using a script and k8s manifest to cleanup duplicate roleBindings and clusterRoleBindings. 

There is nothing added to rancher yet to use this code, it can only be used externally through the script. 

Related to https://github.com/rancher/rancher/issues/29932 but this does not fix the issue, just allows a cleanup to be done on the management cluster.

### DIFF
--- a/cleanup/binding-clean.sh
+++ b/cleanup/binding-clean.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+# set -x
+set -e
+
+CLEAR='\033[0m'
+RED='\033[0;31m'
+
+# Location of the yaml to use to deploy the cleanup job
+yaml_url=https://raw.githubusercontent.com/rancher/rancher/master/cleanup/binding-clean.yaml
+
+# 120 is equal to a minute as the sleep is half a second
+timeout=120
+
+# Agent image to use in the yaml file
+agent_image="$1"
+
+show_usage() {
+  if [ -n "$1" ]; then
+    echo -e "${RED}👉 $1${CLEAR}\n";
+  fi
+	echo -e "Usage: $0 [AGENT_IMAGE] [FLAGS]"
+	echo "AGENT_IMAGE is a required argument"
+	echo ""
+	echo "Flags:"
+	echo -e "\t-dry-run Display the resources that would will be updated without making changes"
+}
+
+if [ $# -lt 1 ]
+then
+	show_usage "AGENT_IMAGE is a required argument"
+	exit 1
+fi
+
+if [[ $1 == "-h" ||$1 == "--help" ]]
+then
+	show_usage
+	exit 0
+fi
+
+# Pull the yaml and replace the agent_image holder with the passed in image
+yaml=$(curl --insecure -sfL $yaml_url | sed -e 's=agent_image='"$agent_image"'=')
+
+if [ "$2" = "-dry-run" ]
+then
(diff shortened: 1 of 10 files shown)


cleanup/binding-clean.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# set -x
+set -e
+
+CLEAR='\033[0m'
+RED='\033[0;31m'
+
+# Location of the yaml to use to deploy the cleanup job
+yaml_url=https://raw.githubusercontent.com/rancher/rancher/master/cleanup/binding-clean.yaml
```
prachidamle, line 9: Is this assuming that we will merge this to master after this 2.5 PR is merged? otherwise this will 404

cleanup/binding-clean.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - get
+      - delete
+  - apiGroups:
+      - batch
+    resources:
+      - jobs
+    verbs:
+      - list
+      - get
+      - delete
```
prachidamle, line 87: needs newline?

pkg/agent/clean/binding.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// The label's key and value changes depending on the rancher version
+	var rancher25 bool
+
+	// Check if we have the updated label, this indicates we are running on rancher 2.5+
+	if len(crtbs.Items) > 0 {
+		if _, ok := crtbs.Items[0].Labels[auth.RtbCrbRbLabelsUpdated]; ok {
+			rancher25 = true
+		}
+	} else if len(prtbs.Items) > 0 {
```
prachidamle, line 109: should we check both crtbs and prtbs for label existance? "else if" will skip prtbs and decide only on crtbs label

pkg/agent/clean/binding.go
```diff
@@ -0,0 +1,295 @@
...
+func (bc *bindingsCleanup) dedupeCRB(bindings []k8srbacv1.ClusterRoleBinding) error {
+	// Sort by creation timestamp so we keep the oldest
+	sort.Sort(crbByCreation(bindings))
+
+	// Leave the first one alone, we only need the duplicates
+	duplicates := bindings[1:]
+
+	for _, binding := range duplicates {
+		if !dryRun {
+			bc.clusterRoleBindings.Delete(binding.Name, &metav1.DeleteOptions{})
```
prachidamle, line 212: should we handle/log delete error?

cleanup/binding-clean.sh
dramich: Yup, master next.

cleanup/binding-clean.yaml
dramich: No need

pkg/agent/clean/binding.go
dramich: If one has it and the other doesn't we are in a really bad state for RBAC, so I just need to check one.

pkg/agent/clean/binding.go
dramich: Logging the error on both delete calls now.

prachidamle (approved):
LGTM

ryansann (approved):

Issue #30978, opened 2021-01-25
Fleet fails to deploy on downstream clusters with error "secret not found"


Ellenqs:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Bring up rancher v2.5-head
2. Create a downstream cluster with 1 controlplane, 1 etcd, 3 workers (ubuntu/flatcar)
3. Go to local cluster exployer -> contiuous delivery -> clusters

**Result:**
Cluster can not find secret and "wait check in".

**Other details that may be helpful:**

**Environment information**
- Rancher version: v2.5-head commit: 23a29c3
- Installation option (single install/HA):single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Provider/custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): AWS ec2
- Kubernetes version (use `kubectl version`):v1.19.7

```
(paste the output here)
```
![Screen Shot 2021-01-25 at 11 00 05 AM](https://user-images.githubusercontent.com/75277058/105754816-5c1ce100-5eff-11eb-9120-ca115345e4a2.png)

Related: https://github.com/rancher/fleet/issues/83

cbron:
@sowmyav27 since this code has reverted in 2.5, can you test this quickly (only on 2.5) to ensure it is not a bug in 2.5. Then we can move this to 2.6 as the code is still there, and likely still exists there.

Ellenqs:
Tested on v2.5-head, cluster is active, bundle deployed successfully.

![Screen Shot 2021-01-27 at 2 05 30 PM](https://user-images.githubusercontent.com/75277058/106061173-e2752680-60a9-11eb-8dce-c58c50f07162.png)
![Screen Shot 2021-01-27 at 2 05 59 PM](https://user-images.githubusercontent.com/75277058/106061179-e43eea00-60a9-11eb-9729-c4a74cc123ba.png)

Ellenqs:
This issue is gone on v2.5-head, but still exits on rancher `master-head (fbe2c30)`.

cbron:
@Ellenqs yes once this PR merges it will be working again: https://github.com/rancher/rancher/pull/31169
But this issue should stay open because the underlying issue itself is not fixed.

cbron:
@rmweir can this go to test now ?

virus2016:
I get

[...]

PR #31169, opened 2021-02-05
Revert token hashing

rmweir:


package/Dockerfile
```diff
@@ -109,7 +109,7 @@ RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
 
 ENV CATTLE_UI_VERSION 2.5.2-rc8.1
 ENV CATTLE_DASHBOARD_UI_VERSION v2.5.2-rc5
-ENV CATTLE_CLI_VERSION v2.4.11-rc1
+ENV CATTLE_CLI_VERSION v2.4.11-rc2
```
cbron, line 112: This is the reverted version right ?

cbron (approved):
1 question, if that is right then we are good.

package/Dockerfile
rmweir: Yep

rmweir:
Revert due to https://github.com/rancher/rancher/issues/30978, plan on fixing that issue in the future and pulling this back in.

ibuildthecloud:
I'd rather this not be reverted.  I'm going to fix rancher-operator for this.

Issue #30049, opened 2020-11-12
Cluster stuck "currently Upgrading, updating worker node" on Cordoned workers


asemen2608:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

How to recreate:
1. Create a cluster with 3 master and a
In that state different issues happen:t least 2 workers.
2. Once the cluster is ready, cordon at least 2 workers.
3. Delete master #1, wait for it to be replaced.
4. Repeat step 3 with master rancher/rancher#2 and #3.
5. Cluster state is still "This cluster is currently Upgrading – Updating worker nodes", some other side effects might also be seen. 

**Result:**
The current bug is that if more than 2 worker nodes are cordoned, the cluster will be stuck in a "This cluster is currently Upgrading – Updating worker nodes" state forever until we uncordon the nodes.

**Other details that may be helpful:**

When a Controlplane node get replaced, the worker nodes need to be updated with the new IP address of the master nodes.

* Sometimes it will take quite a while (more than 30 minutes) to recreate the deleted master node.
* Kubelet is down on multiple nodes (cordoned or not)

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.8
- Installation option (single install/HA): HA

```
kubectl 1.17.2 
```

![screenshot2_](https://user-images.githubusercontent.com/61939253/98946602-603a8480-24f4-11eb-9bd5-5bad2c2c90e1.png)

![screenshot1_](https://user-images.githubusercontent.com/61939253/98946506-413bf280-24f4-11eb-8e96-5805c77f1da3.png)

gz#13331

asemen2608:
A side effect with a cordoned worker node and the replacement of a master node is that the worker node gets automatically uncordoned.

How to recreate:
1. Create a cluster with 3 masters and at least 1 worker.
2. Once the cluster is ready, cordon 1 worker.
3. Delete any master node.
4. The worker that was cordoned is now uncordoned.

asemen2608:
Will there be also a fix for v2.4.x?

kinarashah:
@asemen2608 Thanks for opening the issue, I'll backport the fix so it's available in `v2.4.x` as well.

kinarashah:
[...]

PR #31267, opened 2021-02-11
[v2.5] uncordon upgraded nodes stuck in cordoned state

kinarashah:
Problem:
An unavailable node is always reconciled to desired state bypassing the usual cordon->upgrade->uncordon
path. If they become ready later on, node could have already been upgraded before it's marked for upgrade
(NodeConditionUpgraded is never set). So the node never gets uncordoned since we rely on node conditions.

https://github.com/rancher/rancher/issues/30049

### DIFF
--- a/pkg/controllers/management/rkeworkerupgrader/node.go
+++ b/pkg/controllers/management/rkeworkerupgrader/node.go
@@ -24,6 +24,8 @@ type upgradeStatus struct {
 	toProcess []*v3.Node
 	// upgrading => upgraded => uncordon
 	upgraded []*v3.Node
+	// notReady => stuck in cordoned (unavailable nodes get new plan without NodeConditionUpgraded)
+	toUncordon []*v3.Node
 	// unavailable nodes
 	notReady []*v3.Node
 	// upgraded active nodes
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/rkeworkerupgrader/node.go
```diff
@@ -161,6 +163,10 @@ func (uh *upgradeHandler) filterNodes(nodes []*v3.Node, expectedVersion int, dra
 			} else {
 				// node hasn't un-cordoned, so consider it upgrading in terms of maxUnavailable count
 				status.upgrading++
+				// node has already upgraded, but condition is not unknown, so uncordon it
+				if !v32.NodeConditionUpgraded.IsUnknown(node) && node.Spec.DesiredNodeUnschedulable != "false" {
```
dramich, line 167: Can `NodeConditionUpgraded` be false? If so would we still want to attempt to un-cordon?

kinarashah: It can't be false, this is because we wanted to be consistent with the old approach of how there's no errors when node plan isn't sent or applied correctly -> we just continue trying, and node would stay in upgrading state forever.

dramich (approved):

Issue #31269, opened 2021-02-11
Cluster repo - failed to clone a private git repo with ssh key


Ellenqs:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

- Clone the rancher charts repo (https://github.com/rancher/charts) to private repo.
- Deploy rancher single install.
- Go to cluster explorer -> Apps & marketplace ->  create charts repo -> authentication ->
   select ssh key.
- Provide git repo url, public key, private key.
- Create repo.

**Result:**
Chart repo stuck in downloading state. See the error message below.

<img width="1256" alt="Screen Shot 2021-02-10 at 9 39 19 AM" src="https://user-images.githubusercontent.com/75277058/107669325-77e5ee00-6c46-11eb-8d93-35e107678b74.png">

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `v2.5-head,3c4c625`
- Installation option (single install/HA): single install

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): AWS
- Kubernetes version (use `kubectl version`): v1.20.2

Ellenqs:
Validated on Rancher `v2.5-head, 05fbcbb`.

**Steps**

- Deploy rancher single install.
- Create a downstream cluster.
- Go to cluster explorer -> Apps & marketplace -> create charts repo -> Authentication -> select ssh key.
- Provide git repo url, public key, private key.
- Create repo.

**Results**

Charts repo created successfully.

<img width="1046" alt="Screen Shot 2021-02-16 at 3 01 56 PM" src="https://user-images.githubusercontent.com/75277058/108132775-4dda6480-7068-11eb-90de-902abf9e157c.png">

Ellenqs:
Wait for the PR to be merged into master-head.

Ellenqs:
Validated on Rancher `master-head, bda4456`.

**Steps**

Deploy rancher single install.
Create a downstream cluster.
Go to cluster explorer -> Apps & marketplace -> create charts repo -> Authentication -> select ssh key.
Provide git repo url, public key, private key.
Create repo.

**Results**
Charts repo created successfully.

<img width="1160" alt="Screen Shot 2021-02-17 at 9 26 21 AM" src="https://user-images.githubusercontent.com/75277058/108243726-41efb080-7103-11eb-9d32-ab2deb6eb9ba.png">

PR #31276, opened 2021-02-11
Add ssh to checkout git repo with ssh key

StrongMonkey:
https://github.com/rancher/rancher/issues/31269

### DIFF
--- a/package/Dockerfile.agent
+++ b/package/Dockerfile.agent
@@ -11,7 +11,7 @@ ENV DOCKER_URL_amd64=https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 \
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 ENV KUBECTL_VERSION v1.18.2
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion unzip sysstat acl && \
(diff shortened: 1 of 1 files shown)


package/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM ubuntu:18.04
-RUN apt-get update && apt-get install -y git curl ca-certificates unzip xz-utils && \
+RUN apt-get update && apt-get install -y git curl ca-certificates unzip xz-utils ssh && \
```
dramich, line 2: I don't think we need to add this here, https://packages.ubuntu.com/groovy/git

package/Dockerfile.agent
```diff
@@ -11,7 +11,7 @@ ENV DOCKER_URL_amd64=https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 \
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 ENV KUBECTL_VERSION v1.18.2
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion unzip sysstat acl && \
+    apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion unzip sysstat acl ssh && \
```
dramich, line 14: Can we just drop the `--no-install-recommends`?

dramich (approved):

Issue #22871, opened 2019-09-15
Choose which node to delete when scaling down


pthrasher:
**What kind of request is this (question/bug/enhancement/feature request):**
question

**Steps to reproduce (least amount of steps as possible):**
1. Navigate to nodes page
2. Scale down a node pool

**Result:**
The one I didn't want deleted was deleted.

**Expected Result:**
To be able to choose which node to remove.

**Question:**
Is there some alternate way for me to delete a node and scale down at the exact same time? I need to scale down by 1 and I need to choose the node that gets deleted.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Infrastructure Provider (aws)

gz#9849
gz#12232
gz#11020

gz#14529

dkeightley:
Additionally, when not choosing a node, Rancher defaults to the oldest node. This is not ideal in all circumstances, for eg: if it's the only good node in the cluster at the time. The oldest of the non-Active nodes should take precedence when selecting a node to remove in a scale down.

deniseschannon:
This enhancement should be implemented as:

* Ability to select which node(s) to delete when scaling down based on node ID(s)/node name(s)
* Ability to select whether or not to drain the node before terminating the nodes

kinarashah:
https://github.com/rancher/rancher/issues/27458

ricky-sb:
@pthrasher, do you have a workaround for this?

sowmyav27:
**On 2.5.6-rc5**

- Deployed a node driver cluster - 3 worker node, enable Drain on delete
- Deploy a deployment with replicas on W1, W2 and W3
- Select W1, W3 and "Scale Down"
- Nodes W1 and W3 get drained. The pods are rescheduled on W2.
- And the nodes W1 and W3 are deleted
- Scale up Worker nodes by 2
- W1, W2, W3 are available.

luthermonson:
Feature adds new member to the v3 node spec called 'scaledownTime' which is defaulted to empty. When set as an rfc3339 time stamp the node pool will delete the node and decrement the pool removing the node from the pool. This allows you to select which node to remove from the pool.

[...]

PR #31301, opened 2021-02-12
[v2.5] add scaledown nodepool feature

luthermonson:
Problem: scaling down a nodepool only allows for the +/- buttons or spec.quantity field to change, the nodepool reconciles and adds/removes nodes numerically from low to high. Allow for the ability to flag nodes for scaledown and remove them from the pool and decrement the quantity.
Solution: added node.spec.scaledown which is an RFC3339 date for when the node should be pulled from the pool. added a scaledown action which sets the scaledown time to now. nodepool will reconcile and pull the node marked for scaledown after the scaledown time. 

https://github.com/rancher/rancher/issues/22871


pkg/api/steve/clusters/clusters.go
```diff
@@ -47,6 +48,7 @@ func Register(ctx context.Context, server *steve.Server) error {
 				// By pass authorization for local shell because the user might not have
 				// GET granted for local cluster
 				if request.Name == "local" && request.Link == "shell" {
+					fmt.Println("fdsa fdsa fdsa fdsa")
```
paynejacob, line 51: rm

paynejacob (commented):
1 minor comment

pkg/api/steve/clusters/clusters.go
luthermonson: bahahah good catch

pkg/apis/management.cattle.io/v3/machine_types.go
```diff
@@ -238,6 +238,7 @@ type NodeSpec struct {
 	DesiredNodeUnschedulable string          `json:"desiredNodeUnschedulable,omitempty"`
 	NodeDrainInput           *NodeDrainInput `json:"nodeDrainInput,omitempty"`
 	MetadataUpdate           MetadataUpdate  `json:"metadataUpdate,omitempty"`
+	Scaledown                string          `json:"scaledown,omitempty"`
```
kinarashah, line 241: maybe it's just me but could we name this better? it sounds like a bool to me 

also, do we want to drop this from api view (similar to desiredNodeUnschedulable) so it's not available for people to edit?

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -7,9 +7,13 @@ import (
 	"regexp"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+
```
kinarashah, line 16: nit: go fmt

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -56,15 +62,73 @@ func (c *Controller) Create(nodePool *v3.NodePool) (runtime.Object, error) {
...
+		logrus.Errorf("[nodepool] reconcile erorr, create or check nodes: %s", err)
+	}
+
+	var np *v3.NodePool
+	if qty != nodePool.Spec.Quantity {
+		// pool was scaled down, need to update qty
+		np = nodePool.DeepCopy()
+		np.Spec.Quantity = qty
+	} else {
+		np = nodePool
```
kinarashah, line 86: why do we need to update node pool if the quantity isn't changed?

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -56,15 +62,73 @@ func (c *Controller) Create(nodePool *v3.NodePool) (runtime.Object, error) {
...
+			}
+		} else if anno == "updated" {
+			nodePool.Annotations[ReconcileAnnotation] = ""
+			return c.NodePools.Update(nodePool)
+		} else if strings.HasPrefix(anno, "updating/") {
+			// gate updating the node pool to every 20s
+			pieces := strings.Split(anno, "/")
+			t, err := time.Parse(time.RFC3339, pieces[1])
+			if err != nil || int(time.Since(t)/time.Second) > 20 {
+				nodePool.Annotations[ReconcileAnnotation] = ""
```
kinarashah, line 116: is the main purpose of this update to call the reconcile check again? 
Since this func only gets called on an update sync, I'm confused why we need to clean up the annotation every 20s vs checking this on L98 itself: 

`if anno == "" || timeToGateUpdate`

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -86,6 +150,47 @@ func (c *Controller) Remove(nodePool *v3.NodePool) (runtime.Object, error) {
...
+		Duration: 500 * time.Millisecond,
+		Factor:   1,
+		Jitter:   0,
+		Steps:    6,
+	}
+
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		newPool, err := c.NodePools.GetNamespaced(nodePool.Namespace, nodePool.Name, metav1.GetOptions{})
+		if err != nil {
+			logrus.Info(err)
```
kinarashah, line 164: nit: prefix error w [nodepool] as other err msgs, same L179

kinarashah, line 164: nit: prefix err msg with [nodepool] as other errors + L179

pkg/apis/management.cattle.io/v3/machine_types.go
luthermonson: Definitely want it in the api, re naming I'm open to suggestions

pkg/controllers/management/nodepool/nodepool.go
luthermonson: To clear the annotation

pkg/controllers/management/nodepool/nodepool.go
luthermonson: the annotation has a timestamp set to it. if something failed (server restarted or anything) and the go routine bombed we need to clear the annotation and start again. it's just safety to give things another try if they failed. similar code is in the cluster provisioned controller

pkg/controllers/management/nodepool/nodepool.go
luthermonson: good idea

pkg/apis/management.cattle.io/v3/machine_types.go
deniseschannon: what is the name of this supposed to be?

deniseschannon: Err. what is the value of this supposed to represent? I agree that the name doesn't immediately indicate what it is.

luthermonson: it's an RFC3339 timestamp for when the node will be pulled from the pool. the action sets it to now but you can set it via the api to a time in the future and the nodepool will pull the node after that time

deniseschannon: So it's the timestamp of when to scale down? 

So could it be `ScaledownTimestamp`? I'm not great at naming items.

luthermonson: we do `LastUpdateTime` on norman conditions which are also rfc3339 timestamps, how about `ScaledownTime` ?

kinarashah: sounds better than `scaledown`

pkg/api/norman/customization/node/node.go
```diff
@@ -99,7 +102,24 @@ func (a ActionWrapper) ActionHandler(actionName string, action *types.Action, ap
...
+	return nil
+}
+
+func scaledownNode(actionName string, apiContext *types.APIContext) error {
+	node, schema, err := getNodeAndSchema(apiContext)
+	if err != nil {
+		return err
 	}
+
+	values.PutValue(node, time.Now().Format(time.RFC3339), "scaledown")
```
kinarashah, line 117: would need to change to scaledowntime now

pkg/controllers/management/node/controller.go
```diff
@@ -499,6 +501,28 @@ outer:
...
+	if obj.Spec.ScaledownTime == "" {
+		return nil
+	}
+
+	scaledown, err := time.Parse(time.RFC3339, obj.Spec.ScaledownTime)
+	if err != nil {
+		return fmt.Errorf("[node] failed to parse scaledown time, is it in RFC3339? %s", obj.Spec.ScaledownTime)
+	}
+
+	if scaledown.Before(time.Now()) {
```
kinarashah, line 514: wouldn't we need to `EnqueueAfter` if the time is set to sometime later?  (we could maybe rely on the node updates which happen often, but wanted to ask if that's the decision)

pkg/api/norman/customization/node/node.go
luthermonson: great catch, thanks for that

PR #31314, opened 2021-02-12
Skip app tests on hardened clusters

rancher-max:
Most apps will not deploy on a hardened cluster due to the nature of its security, so better to just skip these tests during the pre/post upgrade checks.


bmdepesa (approved):
LGTM

sowmyav27 (approved):

PR #31317, opened 2021-02-13
Added support for configuring the Service object type in the Helm chart

danielmittelman:
This PR adds the option to configure the Kubernetes Service object type for the Rancher Server. This is primarily needed to expose Rancher via a load balancer (like the AWS Application Load Balancer) outside the cluster, by setting the Service type to `NodePort` and adding the proper annotations to the Ingress object.

If unspecified, the type value defaults to `ClusterIP`, thus maintaining backward compatibility with previous installations that did not explicitly define it.

Example of exposing Rancher via an internal AWS ALB with this change: 
```yaml
...
service:
  type: NodePort
...
ingress:
  enabled: true
  extraAnnotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/scheme: internal
    alb.ingress.kubernetes.io/backend-protocol: HTTP
    alb.ingress.kubernetes.io/target-type: instance
    alb.ingress.kubernetes.io/ssl-policy: "ELBSecurityPolicy-TLS-1-2-Ext-2018-06"
    alb.ingress.kubernetes.io/listen-ports: "[{\"HTTP\": 80}, {\"HTTPS\":443}]"
    alb.ingress.kubernetes.io/certificate-arn: "..."
```


funes79:
Is this still open? Can somebody please merge? I would like to use ALB as well, and it looks like this is exactly what I need. 
Thanks

Issue #31338, opened 2021-02-16
Project network isolation can be enabled if RKE network plugin supports network policies


janeczku:
**What kind of request is this (question/bug/enhancement/feature request):**
Enhancement

**Steps to reproduce (least amount of steps as possible):**
When configuring an RKE network provider that supports Kubernetes network policies, such as the Cisco ACI CNI plugin, Rancher should allow the cluster option to enable Project Network Isolation to be set (`enable_network_policy: true`).

Currently, when enabling the network isolation with such a compatible network provider the cluster provisioning fails with a validation error.

The validation is implemented here:

https://github.com/rancher/rancher/blob/41045a8bd742ca24c69dc3a9d54a1888686c0159/pkg/api/norman/store/cluster/cluster_store.go#L713-L714

**Other details that may be helpful:**

Rancher 2.5.5

ryansann:
# Support Matrix

Rancher RKE Cluster Project Network Isolation (PNI) support per CNI
| CNI | Supports PNI |
| ----------- | ----------- |
| Calico | Yes |
| Canal | Yes |
| Weave | Yes |
| ACI | Yes |
| None/Custom | Yes |
| Flannel | No |

# Testing

Currently Rancher only allows for PNI to be enabled when selecting the Canal Network Provider when provisioning an RKE cluster. These changes allow for PNI to be enabled when using other CNIs. 

Until [this UI issue](FILL ME IN) is complete, you can provision a cluster with PNI enabled by Selecting your CNI, editing the cluster as YAML, and flipping the top-level `enable_network_policy` field to true in the RKE config.

## Allowed Network Paths
With PNI enabled, and the below testing setup, these are the network paths that should be allowed
![pni-allowed](https://user-images.githubusercontent.com/8824599/114619804-c96b2300-9c5f-11eb-9c0e-7b7a4235034d.png)

## Disallowed Network Paths
And these are the network paths that should not be allowed
![pni-not-allowed](https://user-images.githubusercontent.com/8824599/114619836-cf610400-9c5f-11eb-81ff-66579a34e87d.png)

## Steps

1. Launch a cluster with a CNI that supports PNI (see support matrix above) and with PNI enabled
2. Create a project: `pni-testing` and a namespace within it `testing`
3. Launch the Rancher Hello World example as a workload in both the `default` and `pni-testing` projects, e.g. 
[...]

PR #31344, opened 2021-02-16
Permit use of Project Network Isolation with other compatible RKE network providers

janeczku:
This PR will allow the use of the **Project Network Isolation** feature for any RKE network provider that by default enforces Kubernetes network policies. In order to use the feature with plugins other than `canal` the built-in ingress provider must be disabled (`ingress.provider=none`) or configured with `ingress.networkMode=hostPort|none`.

### Details

The Project Network Isolation feature (`enableNetworkPolicy`) is implemented via a controller that maintains Kubernetes network policies configured to either permit or allow traffic between namespaces based on whether the namespace is assigned to a particular project.

Being based on native Kubernetes network policies that feature is compatible with any CNI plugin that enforces Kubernetes network policies. This includes the following RKE network providers:
- Canal
- Calico
- Weave
- ACI

The only CNI plugin-specific functionality included in the network policy management controller is a Canal-specific workaround that allows the creation of network policy ingress rules for the NGINX ingress controller in the case where it's running in `hostNetwork`  mode (which requires the use of an IP-based policy rule).

https://github.com/rancher/rancher/issues/31338

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -54,6 +54,14 @@ const (
 	s3TransportTimeout         = 10
 )
 
+// RKE network providers with default support of Kubernetes network policies
+var rkeNetworkPolicyControllerPlugins = map[string]bool{
+	"canal":  true,
+	"calico": true,
+	"weave":  true,
+	"aci":    true,
+}
+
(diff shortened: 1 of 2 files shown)


janeczku:
Closing in favour of broader solution as discussed

PR #31359, opened 2021-02-17
[master]Use empty map if value is nil

StrongMonkey:


pkg/catalogv2/system/system.go
```diff
@@ -283,6 +283,12 @@ func (m *Manager) isInstalled(namespace, name, version, minVersion string, desir
 		if release.Info.Status != release2.StatusDeployed {
 			continue
 		}
+		if desiredValue == nil {
+			desiredValue = map[string]interface{}{}
+		}
+		if release.Config == nil {
+			release.Config = map[string]interface{}{}
```
ibuildthecloud, line 290: can you not modify the original object, make a new variable for it.

StrongMonkey:

ibuildthecloud (approved):

PR #31360, opened 2021-02-17
[2.5]Use empty map if value is nil

StrongMonkey:


StrongMonkey:

ibuildthecloud (approved):

PR #31084, opened 2021-02-01
Use 2.5.99 for dev version

StrongMonkey:

Issue #31061, opened 2021-01-29
Includes k3s upgrade images when uses image export script in dev, master, or non semver tag that is equal to kd.RancherVersionDev


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Use export script with TAG set to dev, master, or a value equal to kd.RancherVersionDev.

**Result:**
Using export script with TAG set to dev, master, or a value equal to kd.RancherVersionDev will result in no k3s
upgrade or related images being added to the images text.

**Other details that may be helpful:**
Nothing should be observably different when rancher-images texts are produced for tagged releases. QA can verify that an rc tag has a rancher-images text that includes k3s-upgrade images for compatible k3s versions in https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/dev-v2.5/data/data.json. Each compatible version will cause some supporting images to be added as well. Can confirm by making sure klipper-helm images are add to images-text.

cbron:
Aiming for 2.5.6 on this one as I am assuming it will merge before then, but it can slip to 2.5.7 without a problem.

jiaqiluo:
This is validated on rancher:`v2.5.6-rc4`

Steps:
- get the compatible k3s versions from the [here](https://raw.githubusercontent.com/rancher/kontainer-driver-metadata/dev-v2.5/data/data.json)
- get the `rancher-images.txt` from rancher:`v2.5.6-rc4`'s [release page](https://github.com/rancher/rancher/releases/tag/v2.5.6-rc4)

Results:
- there are four compatible k3s versions for rancher:v2.5.6
```
v1.17.17+k3s1
v1.18.15+k3s1
v1.19.7+k3s1
v1.20.2+k3s1
```
- the `klipper-helm` images are in the `rancher-images.txt` file 
```
rancher/klipper-helm:v0.2.3
rancher/klipper-helm:v0.2.7
rancher/klipper-helm:v0.3.0
rancher/klipper-helm:v0.4.3
```
- the `k3s-upgrade` images are in the `rancher-images.txt` file  
```
rancher/k3s-upgrade:v1.17.17-k3s1
rancher/k3s-upgrade:v1.18.15-k3s1
rancher/k3s-upgrade:v1.19.7-k3s1
rancher/k3s-upgrade:v1.20.2-k3s1
```

jiaqiluo:
[...]

PR #31376, opened 2021-02-18
Revert "Merge pull request #31041 from rmweir/k3s-upgrade-dev-export"

rmweir:
**Problem:**
This fix no longer works because it depends on the default dev version being an incomplete semver version but a recent change has updated it to a complete version. The aforementioned change also solves the same issue that this fix did, the only difference being that it will now only include images that have constraints compatible with a .99 patch version.

**Solution:**
Revert fix and rely on the new fix instead: https://github.com/rancher/rancher/pull/31084

**Notes:**
Original fix was to address: https://github.com/rancher/rancher/issues/31061


StrongMonkey:
Should you revert the actual commit instead of merge commit?

rmweir:
@StrongMonkey fixed

StrongMonkey (approved):
:100:

thedadams (approved):

cbron (approved):
Seems fine, just need to re-open issue and test this.

Issue #31370, opened 2021-02-18
Extend agentEnvVars support to rancher agents across all cluster types, including the node run command for cluster and node driver clusters


axeal:
**Steps to reproduce (least amount of steps as possible):**

Support for adding proxy settings to the Rancher cluster agent in imported clusters is being added via the addition of an `agentEnvVars` cluster setting in https://github.com/rancher/rancher/issues/24876

However, this implementation is limited to imported clusters only and does not enable configuration of proxy settings on the agents in other cluster types. Lack of support for agent proxy configuration prevents the provisioning of clusters where a forward proxy is required for access to Rancher from the downstream cluster, as well as causing an error for catalogs in downstream clusters in Rancher v2.5, where a proxy is required for internet access from the downstream cluster (as tracked in https://github.com/rancher/rancher/issues/29961).

Extending `agentEnvVars` support across agents for all cluster types will mitigate https://github.com/rancher/rancher/issues/29961 and provide consistent support for proxy configuration across clusters.

In addition to cluster and node agents, the `agentEnvVars` settings would need to be propagated to the agent node run command used (manually) to join custom cluster nodes and (automatically) to join node driver cluster nodes, to permit successful provisioning of these clusters in environments requiring a forward proxy for access to the Rancher server.

**Environment information**
- Rancher v2.5.5

kinarashah:
@axeal Thanks for opening the issue, I am currently working on #29961 and the fix is going to be exactly what you mentioned - extending `agentEnvVars` for RKE clusters. Running into a few issues at the moment, but I'll let you know once the PRs are in so you could test it out and see if it's working as expected.

Jono-SUSE-Rancher:
Let me know once the PRs are tested and if it's working as expected, I'll likely close this one as a duplicate of #29961 once we're satisfied with the results.

sowmyav27:
As part of this issue - validate on custom cluster (the agent command should have the env variables) and node driver rke clusters. Check cattle cluster agent and cattle node agent. --> should have the env variables set

slickwarren:
the following has been tested on 2.5-head (30e53bd4927449b67b707164a3c9557fac2d9194) and master-head (465bb3b80cb9f2d8c002fa9d7c37be20d38ed411)
* [imported clusters](https://github.com/rancher/rancher/issues/29993)
* custom clusters ([params added through the UI](https://github.com/rancher/rancher/issues/31023))
[...]

Issue #29961, opened 2020-11-06
Public Helm Repo's (marketplace) not pulled in anymore due to proxy issues.


MG2R:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
- set up a Rancher 2.5.1 behind an HTTP proxy (with the `proxy` value set in Helm at installation time)
- Try adding https://helm.elastic.co as a legacy catalog (Cluster Manager)
- Try adding https://helm.elastic.co as a Repository in the new "Apps & Marketplace" Cluster Explorer UI

**Result:**
Legacy way -> just works
New way -> ERROR `Get "https://helm.elastic.co/index.yaml": dial tcp: lookup helm.elastic.co on 10.43.0.10:53: no such host`

**Other details that may be helpful:**
When I refresh the new type of Helm Repo's, I can see log entries in the cluster-agent of the managed RKE cluster (i.e. the one I'm trying to add the new Repo for). When I refresh the legacy catalog, I get no log entries anywhere.

This leads me to believe the old catalogs were fully managed on the Rancher Server, which was started with the correct proxy information by Helm, while the new ones are managed by the individual clusters' cluster-agent, which does not get the correct proxy settings from Rancher.

Another weird thing here is that the built-in Helm Repo's (in the new UI) do actually work, even though they're getting information from an external source (git.rancher.io). These also do not generate logs in the clusters' cluster agent, so I'm guessing those are managed by Rancher Server itself? 

There's almost zero documentation regarding all of this, the way it works, or which settings there are for it. Is there a way to provision clusters and set the correct proxy information for Rancher components in that cluster? I have a workaround available (locally caching external repo's), but I'd still like to properly understand what's going on here. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.1
- Installation option (single install/HA): HA. Rancher Server running on RKE, the managed clusters are 

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM, 4 cores, 8 GB or 16 GB (etcd vs workers)
- Kubernetes version (use `kubectl version`):

```
[...]

Issue #24876, opened 2020-01-07
Rancher agent  does not use proxy environment in callback to  rancher


chrisbulgaria:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Import existing cluster into Rancher - with a proxy between that cluster and Rancher. Add to the agent yaml the required HTTP_PROXY, HTTPS_PROXY, NO_PROXY envs

Expected : Rancher cluster/node can contact Rancher and import is successful

**Result:**
Rancher agent wss calls fail, log 
time="2019-12-19T09:39:08Z" level=info msg="Connecting to wss://rancher.foobar.com with token $token"
time="2019-12-19T09:39:08Z" level=info msg="Connecting to proxy" url="wss://rancher.foobar.com/v3/connect/register"
time="2019-12-19T09:39:18Z" level=error msg="Failed to connect to proxy. Empty dialer response" error="dial tcp :443: i/o timeout"
time="2019-12-19T09:39:18Z" level=error msg="Remotedialer proxy error" error="dial tcp $IP_OF_RANCHER:443: i/o timeout"

**Other details that may be helpful:**
This is the same issue as rancher/rancher#24681 and rancher/rancher#22750 , sorry for opening a new one, but those are probably not clear enough.
The problem is in remotedialer/client.go line 26
dialer = &websocket.Dialer{HandshakeTimeout:HandshakeTimeOut}
should read instead
dialer = &websocket.Dialer{HandshakeTimeout:HandshakeTimeOut,Proxy: http.ProxyFromEnvironment}

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
2.3.3 single install

gz#13523

chrisbulgaria:
This  bug still exists in 2.3.4

rshiva777:
@ loganhz, is the issue resolved. it am still facing the same proxy issue in v.2.3.4

chrisbulgaria:
No , it’s not resolved unfortunately . I ‘ll continue to try to make rancher people see it’s important ... here is the pull request fixing it : https://github.com/rancher/remotedialer/pull/12

sangeethah:
Re opening this issue for validation.

bmdepesa:
rancher/rancher:v2.3-head
rancher/rancher:master-head

* Created Rancher behind proxy
* Created nodedriver ec2 cluster
[...]

PR #31381, opened 2021-02-18
changes for agent env vars

kinarashah:
- added `AgentEnvVars` to node-agent template, same as cluster-agent 
- added `AppliedAgentEnvVars` to cluster status and redeploy agents if changed from spec 
- passing env vars to node registration command generated for custom clusters 
- passing env vars for deployAgent by node driver clusters 
- passing clusterID to the import url generated for imported clusters so `http://server_IP/v3/import/{token}_{clusterID}.yaml` has the env vars added by user on the cluster object 

https://github.com/rancher/rancher/issues/24876
https://github.com/rancher/rancher/issues/29961
https://github.com/rancher/rancher/issues/31370

### DIFF
--- a/pkg/api/norman/customization/clusterregistrationtokens/formatter.go
+++ b/pkg/api/norman/customization/clusterregistrationtokens/formatter.go
@@ -3,6 +3,7 @@ package clusterregistrationtokens
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
 	"github.com/rancher/norman/types"
 	util "github.com/rancher/rancher/pkg/cluster"
@@ -18,7 +19,7 @@ import (
 const (
 	commandFormat            = "kubectl apply -f %s"
 	insecureCommandFormat    = "curl --insecure -sfL %s | kubectl apply -f -"
-	nodeCommandFormat        = "sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run %s --server %s --token %s%s"
+	nodeCommandFormat        = "sudo docker run -d --privileged --restart=unless-stopped --net=host -v /etc/kubernetes:/etc/kubernetes -v /var/run:/var/run %s %s --server %s --token %s%s"
 	loginCommandFormat       = "echo \"%s\" | sudo docker login --username %s --password-stdin %s"
 	windowsNodeCommandFormat = `PowerShell -NoLogo -NonInteractive -Command "& {docker run -v c:\:c:\host %s%s bootstrap --server %s --token %s%s%s | iex}"`
 )
@@ -54,6 +55,7 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 		agentImage := image.ResolveWithCluster(settings.AgentImage.Get(), cluster)
 		// for linux
 		resource.Values["nodeCommand"] = fmt.Sprintf(nodeCommandFormat,
+			AgentEnvVars(cluster),
 			agentImage,
 			rootURL,
 			token,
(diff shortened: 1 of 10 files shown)


pkg/api/norman/customization/clusterregistrationtokens/formatter.go
```diff
@@ -41,19 +42,21 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 	}
 
 	token, _ := resource.Values["token"].(string)
+	clusterID := fmt.Sprintf("%v", resource.Values["clusterId"])
```
StrongMonkey, line 45: nit: there is a `convert.ToString` method that can handle this better

pkg/api/norman/customization/clusterregistrationtokens/import.go
```diff
@@ -6,15 +6,22 @@ import (
...
 
-func ClusterImportHandler(resp http.ResponseWriter, req *http.Request) {
+type ClusterImport struct {
+	Clusters v3.ClusterInterface
+}
+
+func (ch *ClusterImport) ClusterImportHandler(resp http.ResponseWriter, req *http.Request) {
 	resp.Header().Set("Content-Type", "text/plain")
 	token := mux.Vars(req)["token"]
+	clusterID := mux.Vars(req)["clusterId"]
```
StrongMonkey, line 24: hmm How does this work? ClusterId is not a query parameter on the URL?

pkg/api/norman/customization/clusterregistrationtokens/import.go
```diff
@@ -30,8 +37,13 @@ func ClusterImportHandler(resp http.ResponseWriter, req *http.Request) {
 		authImage = authImages[0]
 	}
 
+	var cluster *v3.Cluster
+	if clusterID != "" {
+		cluster, _ = ch.Clusters.Get(clusterID, metav1.GetOptions{})
```
StrongMonkey, line 42: If there is err should we return a http error instead of continuing?

pkg/api/norman/customization/clusterregistrationtokens/import.go
kinarashah: we updated that as well, https://github.com/rancher/rancher/pull/31381/files#diff-68120b5b2bc57e41b2dee9f83f5022c4487dd82cb597c132152c58ffc4286e5bR169

pkg/api/norman/customization/clusterregistrationtokens/import.go
kinarashah: the earlier code always generated agent template, so I wanted to keep that behavior as is even if cluster get fails. the only thing we get from cluster obj is agent env vars, so didn't seem enough to fail, what do you think?

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -237,6 +237,21 @@ func redeployAgent(cluster *v3.Cluster, desiredAgent, desiredAuth string, desire
 		return true
 	}
 
+	agentEnvVarsChanged := false
+	desiredAgentEnvVars := cluster.Spec.AgentEnvVars
+	currentAgentEnvVars := cluster.Status.AppliedAgentEnvVars
+
+	if (desiredAgentEnvVars == nil && currentAgentEnvVars != nil) || (desiredAgentEnvVars != nil && currentAgentEnvVars == nil) {
```
StrongMonkey, line 244: I think `reflect.DeepEqual(desiredAgentEnvVars, currentAgentEnvVars)` should handle all the cases?

StrongMonkey: So this can be 
```
if !reflect.DeepEqual(cluster.Spec.AgentEnvVars, cluster.Status.AppliedAgentEnvVars) {
    logrus.Infof("clusterDeploy: redeployAgent: redeploy Rancher agents due to agent env vars mismatched for [%s], was [%v] and will be [%v]", cluster.Name, currentAgentEnvVars, desiredAgentEnvVars)
    return true
}
```

pkg/api/norman/customization/clusterregistrationtokens/import.go
StrongMonkey: :+1:

StrongMonkey (dismissed):

StrongMonkey (approved):

Issue #31222, opened 2021-02-09
[v2.5] KDM release: k8s feb patches


kinarashah:
This KDM release is going out with Rancher `v2.5.6` release for new k8s versions: v1.17.18, v1.18.16, v1.19.8 and v1.20.3.

- There are existing k8s versions in `dev-v2.5` that haven't been released (absent in `release-v2.5`), so the same versions can be updated. Following versions need to be updated with their compatible rancher server versions: 

| update [version](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/k8s_rke_system_images.go#L3590) from | to | update [rancher version constraints]( https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.4/rke/k8s_version_info.go#L278) 
| --------------- | --------------- | -----------| 
| v1.20.2-rancher2-1 | v1.20.3-rancher1-1 | 2.5.6+ | 
| v1.19.7-rancher2-1 | v1.19.8-rancher1-1 |  2.5.0+ | 
| v1.18.15-rancher2-1 | v1.18.16-rancher1-1 | 2.4.5+ | 
|v1.17.17-rancher2-1 | v1.17.18-rancher1-1 | 2.4.4+ | 

- Following [system images](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/k8s_rke_system_images.go) need to be updated: 

| Name | Kubernetes Version |  Version |
| ------ | ------ | ------| 
| windowsPodInfra | v1.20.3 | v0.1.6 | 

- [Update constraints for all addon versions](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/templates/templates.go#L78) so the new Kubernetes versions use the right template. Refer to the table below to confirm the template name. 

<details>
<summary> Confirm the versions for System Images and Addon templates are correct. Update this table if new versions get introduced. Click to view the full list. </summary>

| Name | Kubernetes Version |  Version | [Update constraints for addon versions](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/templates/templates.go#L78) if required | 
| ------ | ------ | ------| ---- | 
| etcd |  1.17.18, v1.18.16 | v3.4.3-rancher1 | 
| etcd | v1.19.8 | v3.4.13-rancher1 | 
| etcd | v1.20.3 | v3.4.14-rancher1 | 
|  | 
[...]

PR #31383, opened 2021-02-18
[master] Vendor rke

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/31222


cbron (approved):
We should bump UI once we get final tags

PR #31428, opened 2021-02-23
[2.5] Verify context for own check on roleTemplate

dramich:


pkg/controllers/management/auth/manager.go
```diff
@@ -770,18 +775,24 @@ func buildRule(resource string, verbs map[string]bool) v1.PolicyRule {
...
 	if err != nil {
 		return false, err
 	}
 
+	// Only check if we are in the same context, if the roleTemplate is from a different context then
+	// it can't possibly be a owner in the callers context.
+	if roleTemplate.Context != roleTemplateContext {
+		return false, nil
+	}
+
```
prachidamle, line 789: This checks the "project" or "cluster" context and ensures that a project-owner does not get owner rights for the cluster.
1) But should a cluster-owner actually be able to delete a project? This will block a cluster-owner from managing a project in that cluster and is that ok?
2) This just blocks based on the context, where do we block that a user who is a project-owner for project A cannot get "owner" access to some other project B?

dramich: Discussed offline. This change is working as intended and does not stop cluster-owner.

prachidamle (approved):
LGTM

PR #31444, opened 2021-02-23
Add OAuth2 for GKE in proxy server

thedadams:
Google (GKE specifically) must use OAuth2 for authentication. In order to do that efficiently, the google oauth2 library should be used for requests made to GKE. In this case, the proxy server was missing this.

Issue:
https://github.com/rancher/rancher/issues/31434

### DIFF
--- a/pkg/clustermanager/manager.go
+++ b/pkg/clustermanager/manager.go
@@ -13,13 +13,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/rancher/wrangler/pkg/ratelimit"
-
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
 	"github.com/rancher/lasso/pkg/controller"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/clusterrouter"
 	clusterController "github.com/rancher/rancher/pkg/controllers/managementuser"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
@@ -31,8 +28,8 @@ import (
 	"github.com/rancher/rke/pki/cert"
 	"github.com/rancher/steve/pkg/accesscontrol"
 	rbacv1 "github.com/rancher/wrangler/pkg/generated/controllers/rbac/v1"
+	"github.com/rancher/wrangler/pkg/ratelimit"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/oauth2"
 	"golang.org/x/sync/semaphore"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/clustermanager/manager.go
```diff
@@ -319,14 +316,10 @@ func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Con
...
-				if err == nil {
-					return &oauth2.Transport{
-						Source: ts,
-						Base:   rt,
-					}
+				rt, err = gke.Oauth2Transport(context.RunContext, rt, cred)
+				if err != nil {
+					logrus.Errorf("unable to retrieve token source for GKE oauth2: %v", err)
 				}
-				logrus.Errorf("unable to retrieve token source for GKE oauth2: %v", err)
```
cbron, line 329: So no changes here, just consolidating.

pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -220,7 +222,20 @@ func (r *RemoteService) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	}
+
+	if r.cluster.Status.Driver == "googleKubernetesEngine" && r.cluster.Spec.GenericEngineConfig != nil {
+		cred, _ := (*r.cluster.Spec.GenericEngineConfig)["credential"].(string)
+		transport, err = gke.Oauth2Transport(context.Background(), transport, cred)
+		if err != nil {
+			er.Error(rw, req, fmt.Errorf("unable to retrieve token source for GKE oauth2: %v", err))
+			return
+		}
+	} else if r.auth == nil {
```
cbron, line 238: Is there a reason we should have this not run when the cluster is GKE ? 
I don't get why we didn't leave the `getTransport` call where it was and put this new code below that. Do we always want Oauth to skip the `auth()` calls here ?

pkg/clustermanager/manager.go
thedadams: Correct. Since we are doing this in two different places I created a function to be called.

pkg/clusterrouter/proxy/proxy_server.go
thedadams: It is to emphasize that we are NOT using the bearer token. The transport is handling it for us.

cbron: Assuming you don't have to delete the `Authorization` header because it never existed ? Ok.

cbron (approved):
Think this makes sense, but wait for @rmweir

pkg/clustermanager/manager.go
```diff
@@ -319,14 +316,10 @@ func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Con
...
 				cred, _ := (*cluster.Spec.GenericEngineConfig)["credential"].(string)
-				ts, err := gke.GetTokenSource(context.RunContext, cred)
-				if err == nil {
-					return &oauth2.Transport{
-						Source: ts,
-						Base:   rt,
-					}
+				rt, err = gke.Oauth2Transport(context.RunContext, rt, cred)
+				if err != nil {
+					logrus.Errorf("unable to retrieve token source for GKE oauth2: %v", err)
```
rmweir, line 321: What does rt's RoundTrip method do if there was an error here?

thedadams: If there is an error, then `gke.Oauth2Transport` will return the `rt` it was given so execution can continue. This means that the request will likely fail because of a bad token.

rmweir (approved):
LGTM

PR #31495, opened 2021-02-25
Drop go-cache from dapper volumes

dramich:


cbron (approved):
Seems fine if it doesn't slow things down. If it does we should try Daishan's dual-volume theory.

PR #31530, opened 2021-02-27
[v2.5] regenerate nodeCommand to set updated env vars for agent

kinarashah:
https://github.com/rancher/rancher/issues/31529

### DIFF
--- a/pkg/api/norman/customization/clusterregistrationtokens/formatter.go
+++ b/pkg/api/norman/customization/clusterregistrationtokens/formatter.go
@@ -101,7 +101,10 @@ func AgentEnvVars(cluster *v3.Cluster) string {
 	var agentEnvVars []string
 	if cluster != nil {
 		for _, envVar := range cluster.Spec.AgentEnvVars {
-			agentEnvVars = append(agentEnvVars, fmt.Sprintf("-e %s=%s", envVar.Name, envVar.Value))
+			// envVar.ValueFrom isn't supported currently for docker run command https://github.com/rancher/docs/pull/3075
+			if envVar.Name != "" && envVar.Value != "" {
+				agentEnvVars = append(agentEnvVars, fmt.Sprintf("-e %s=%s", envVar.Name, envVar.Value))
+			}
 		}
 	}
 	return strings.Join(agentEnvVars, " ")


luthermonson (approved):

pkg/api/norman/customization/clusterregistrationtokens/formatter.go
```diff
@@ -75,6 +78,22 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 			token,
 			ca,
 			getWindowsPrefixPathArg(cluster.Spec.RancherKubernetesEngineConfig))
+	} else {
+		// refresh nodeCommand if agent image or env vars changed
+		cluster, err := f.ClusterLister.Get("", clusterID)
+		if err != nil && errors.IsNotFound(err) {
+			cluster, _ = f.Clusters.Get(clusterID, metav1.GetOptions{})
```
prachidamle, line 85: why not handle this error returned by f.Clusters.Get?

kinarashah: I wanted to preserve the old behavior, the UI always showed the node registration command. I didn't want to return an error in case cluster doesn't load. Users might not always set cluster env vars, so would be weird to get an error in that case...if they did, worst case would be getting an old command. 

What do you think?

dramich: An error can't be returned here anyways, this is inside a formatter. The best that could happen is an error in rancher logs which won't really help a user.

pkg/api/norman/customization/clusterregistrationtokens/formatter.go
```diff
@@ -75,6 +78,22 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 			token,
 			ca,
 			getWindowsPrefixPathArg(cluster.Spec.RancherKubernetesEngineConfig))
+	} else {
+		// refresh nodeCommand if agent image or env vars changed
+		cluster, err := f.ClusterLister.Get("", clusterID)
```
dramich, line 83: I think we should just do an API call here, if the intention is to be updating things getting the most recent version seems like a safer option and we really aren't going to hit this code much.

pkg/api/norman/customization/clusterregistrationtokens/formatter.go
```diff
@@ -75,6 +78,22 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+			cluster, _ = f.Clusters.Get(clusterID, metav1.GetOptions{})
+		}
+		if cluster != nil {
+			rootURL := getRootURL(request)
+			agentImage := image.ResolveWithCluster(settings.AgentImage.Get(), cluster)
+			resource.Values["nodeCommand"] = fmt.Sprintf(nodeCommandFormat,
+				AgentEnvVars(cluster),
+				agentImage,
+				rootURL,
+				token,
```
dramich, line 94: This would end up being empty string

pkg/api/norman/customization/clusterregistrationtokens/formatter.go
prachidamle: So why not log it, it is anyway useful to debug?

kinarashah:
Thanks for reviewing @prachidamle @dramich! As @dramich pointed out, we already generated node command correctly and this change isn't really required. I am not sure if I reproduced it wrong or my eyes tricked me but UI will fix this by saving cluster and reloading clusterregistration tokens. 

The only change now in this PR is making sure we don't append env vars with empty name and value, if both name and value is empty it adds `-e ` to the command. Also, this means we're dropping valueFrom from node command.

dramich (approved):

pkg/api/norman/customization/clusterregistrationtokens/formatter.go
```diff
@@ -101,7 +101,9 @@ func AgentEnvVars(cluster *v3.Cluster) string {
 	var agentEnvVars []string
 	if cluster != nil {
 		for _, envVar := range cluster.Spec.AgentEnvVars {
-			agentEnvVars = append(agentEnvVars, fmt.Sprintf("-e %s=%s", envVar.Name, envVar.Value))
+			if envVar.Name != "" && envVar.Value != "" {
```
dramich, line 104: Can you add a comment we are explicitly dropping valueFrom and link to the issue for the docs as it being currently unsupported?

prachidamle (approved):

pkg/api/norman/customization/clusterregistrationtokens/formatter.go
kinarashah: ✅

dramich (approved):

Issue #30981, opened 2021-01-25
Add GCP cloud credential option


cbron:
For GKEv2 we need to make a cloud credential that works with GCP. This should not introduce a node driver to GCP.

prabalsharma:
I was able to create CC on 2.5-head-7426ba83e using UI and checked the created secret using kubectl

prabalsharma:
was tested on master-head https://zube.io/rancher/rancher/c/37660

prabalsharma:
invalid json bug: https://github.com/rancher/rancher/issues/32020

prabalsharma:
2.5-head-7426ba83e: deletion of CC from UI, cleans the secret as well.

PR #31548, opened 2021-03-01
Add support for creating a GKE cloud credential

cmurphy:
This patch adds the ability to create and store a CloudCredential via
the /v3/cloudcredentials API that can be used to drive GKE clusters.

By adding it to the machine driver controller, Rancher can expose this
credential type as a dynamic schema in the same way that other cloud
credentials are supported. However, since GKE is not a real machine
driver, we have to fake it by creating dummy flags for a pretend machine
driver binary.

The credential is identified by the private field "authEncodedJson"
where the value is an *escaped* string representing a JSON object for
the service account data obtained from Google. If the credential object
is provided as a regular JSON object rather than an escaped string,
Rancher will end up converting into a Go map and then to a string
representation of the map before storing it in a Secret, rather than
directly storing the JSON object that is needed to authenticate with
Google. The result is that creating a cloud credential of this type will
look like this:

```
HTTP/1.1 POST /v3/cloudcredentials

{
  "googlecredentialConfig": {
    "authEncodedJson": "{\"type\": \"service_account\",
      \"project_id\": \"my-project\",
      \"private_key_id\": \"abcdef\",
      \"private_key\": \"-----BEGIN PRIVATE KEY-----\\nABCdef==\\n-----END PRIVATE KEY-----\\n\",
      \"client_email\": \"user@my-project.iam.gserviceaccount.com\",
      \"client_id\": \"123456\",
      \"auth_uri\": \"https://accounts.google.com/o/oauth2/auth\",
      \"token_uri\": \"https://oauth2.googleapis.com/token\",
      \"auth_provider_x509_cert_url\": \"https://www.googleapis.com/oauth2/v1/certs\",
      \"client_x509_cert_url\": \"https://www.googleapis.com/robot/v1/metadata/x509/user%40project.iam.gserviceaccount.com\"}"
  }
}
```

#30981

### DIFF
--- a/pkg/controllers/management/drivers/nodedriver/machine_driver.go
+++ b/pkg/controllers/management/drivers/nodedriver/machine_driver.go
@@ -39,6 +39,7 @@ var (
 		"otc":           map[string]string{"privateKeyFile": "privateKeyFile"},
 		"packet":        map[string]string{"userdata": "userdata"},
 		"vmwarevsphere": map[string]string{"cloud-config": "cloudConfig"},
+		"google":        map[string]string{"authEncodedJson": "authEncodedJson"},
 	}
 	SSHKeyFields = map[string]bool{
 		"sshKeyContents": true,
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -67,6 +67,7 @@ var aliases = map[string]map[string]string{
 	"otc":           map[string]string{"privateKeyFile": "privateKeyFile"},
 	"packet":        map[string]string{"userdata": "userdata"},
 	"vmwarevsphere": map[string]string{"cloudConfig": "cloud-config"},
+	"google":        map[string]string{"authEncodedJson": "authEncodedJson"},
 }
 
 func Register(ctx context.Context, management *config.ManagementContext, clusterManager *clustermanager.Manager) {
--- a/pkg/data/management/machinedriver_data.go
+++ b/pkg/data/management/machinedriver_data.go
@@ -28,6 +28,7 @@ const (
 	RackspaceDriver    = "rackspace"
 	SoftLayerDriver    = "softlayer"
 	Vmwaredriver       = "vmwarevsphere"
+	GoogleDriver       = "google"
 )
 
 var driverData = map[string]map[string][]string{
@@ -44,6 +45,7 @@ var driverData = map[string]map[string][]string{
 	RackspaceDriver:    {"privateCredentialFields": []string{"apiKey"}},
 	SoftLayerDriver:    {"privateCredentialFields": []string{"apiKey"}},
 	Vmwaredriver:       {"publicCredentialFields": []string{"username", "vcenter", "vcenterPort"}, "privateCredentialFields": []string{"password"}},
+	GoogleDriver:       {"privateCredentialFields": []string{"authEncodedJson"}},
 }
 
 var driverDefaults = map[string]map[string]string{
@@ -92,6 +94,9 @@ func addMachineDrivers(management *config.ManagementContext) error {
 		return err
 	}
 	if err := addMachineDriver(ExoscaleDriver, "local://", "", "", []string{"api.exoscale.ch"}, false, true, management); err != nil {
+		return err
+	}
+	if err := addMachineDriver(GoogleDriver, "local://", "", "", nil, true, true, management); err != nil {
 		return err
 	}
 	linodeBuiltin := true
--- a/pkg/controllers/management/drivers/nodedriver/utils.go
+++ b/pkg/controllers/management/drivers/nodedriver/utils.go
@@ -69,6 +69,21 @@ func ToLowerCamelCase(nodeFlagName string) (string, error) {
 }
 
 func getCreateFlagsForDriver(driver string) ([]cli.Flag, error) {
+	var flags []cli.Flag
+	// NOTE(cmurphy): There is not currently a real google machine driver, but
+	// we still want to be able to create a Google cloud credential to use with
+	// GKE. This fake driver flag allows us to go through the motions of
+	// handling the credential fields without actually needing to run the
+	// machine driver binary.
+	if driver == "google" {
+		flags = []cli.Flag{
+			&cli.StringFlag{
+				Name: "google-auth-encoded-json",
+			},
+		}
+		return flags, nil
+	}
+
 	logrus.Debug("Starting binary ", driver)
 	p, err := localbinary.NewPlugin(driver)
 	if err != nil {
@@ -93,8 +108,6 @@ func getCreateFlagsForDriver(driver string) ([]cli.Flag, error) {
 	defer rpcclient.Close()
 
 	c := rpcdriver.NewInternalClient(rpcclient)
-
-	var flags []cli.Flag
 
 	if err := c.Call(".GetCreateFlags", struct{}{}, &flags); err != nil {
 		return nil, fmt.Errorf("Error getting flags err=%v", err)


pkg/data/management/machinedriver_data.go
```diff
@@ -28,6 +28,7 @@ const (
 	RackspaceDriver    = "rackspace"
 	SoftLayerDriver    = "softlayer"
 	Vmwaredriver       = "vmwarevsphere"
+	GKEDriver          = "gke"
```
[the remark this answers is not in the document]
thedadams: The name of the driver should be `google` in case we add more general support in the future.

pkg/controllers/management/drivers/nodedriver/machine_driver.go
```diff
@@ -179,7 +181,17 @@ func (m *Lifecycle) download(obj *v3.NodeDriver) (*v3.NodeDriver, error) {
...
 	}
 
-	flags, err := getCreateFlagsForDriver(driverName)
+	var flags []cli.Flag
+	if driverName == "gke" {
+		flags = []cli.Flag{
+			&cli.StringFlag{
+				Name: "gke-credential",
+			},
+		}
```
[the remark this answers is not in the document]
thedadams: I believe this is the only place the driver is called here. This is just to get the flags for creation.

The name of the flag should be `google-auth-encoded-json` to make it compatible with machine (in case we add more general support).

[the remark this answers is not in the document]
thedadams: If the changes do get merged into `rancher-machine`, then this block can be removed.

thedadams (changes requested):

pkg/controllers/management/drivers/nodedriver/machine_driver.go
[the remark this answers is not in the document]
cmurphy: Where does the string `google-auth-encoded-json` come from? I used `credential` because that's currently how kontainer-engine identifies it

https://github.com/rancher/kontainer-engine/blob/89626b028c6a0cb1eeb399c9e7dc5404286efbb4/drivers/gke/gke_driver.go#L323 

https://github.com/rancher/kontainer-engine/blob/89626b028c6a0cb1eeb399c9e7dc5404286efbb4/drivers/gke/gke_driver.go#L846

pkg/data/management/machinedriver_data.go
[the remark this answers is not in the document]
cbron: `google` or `gcp` ? I'd vote second one.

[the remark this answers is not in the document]
cmurphy: +1 for `gcp`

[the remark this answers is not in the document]
thedadams: It's not a choice, unfortunately. This code uses rancher-machine, so if we are going to support GCP in the future then we have to use the driver's name in rancher-machine which is `google`.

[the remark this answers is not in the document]
cmurphy: It's not using rancher-machine though, it's purposely working around using rancher-machine. Don't we want to be able to decouple cloud credentials from machine provisioning?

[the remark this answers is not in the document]
cbron: We do, but there's a decent chance we'll add GCP VM's to rancher-machine in near future.

cmurphy:
Fixed the driver name and the flag name. Also note the changed commit message and PR description: the JSON token credential needs to be provided as an escaped string, not a direct JSON object, otherwise the GKE operator will have a hard time extracting the object from the Secret.

thedadams (dismissed):

cbron (commented):
lgtm, but waiting on review from Kinara

pkg/controllers/management/node/controller.go
```diff
@@ -67,6 +67,7 @@ var aliases = map[string]map[string]string{
 	"otc":           map[string]string{"privateKeyFile": "privateKeyFile"},
 	"packet":        map[string]string{"userdata": "userdata"},
 	"vmwarevsphere": map[string]string{"cloudConfig": "cloud-config"},
+	"google":        map[string]string{"authEncodedJson": "authEncodedJson"},
```
kinarashah, line 70: note: this won't do anything at the moment. @dramich mentioned long term goal is to add a driver so can keep it as is

pkg/data/management/machinedriver_data.go
```diff
@@ -94,6 +96,9 @@ func addMachineDrivers(management *config.ManagementContext) error {
 	if err := addMachineDriver(ExoscaleDriver, "local://", "", "", []string{"api.exoscale.ch"}, false, true, management); err != nil {
 		return err
 	}
+	if err := addMachineDriver(GoogleDriver, "local://", "", "", []string{"www.googleapis.com"}, true, true, management); err != nil {
```
kinarashah, line 99: we don't need to whitelist `local://` since we're not going call rancher for google driver (thanks @dramich)

dramich: @cmurphy To expand on this the whitelist is URLs that rancher will proxy for the UI. The UI calls to providers through rancher to talk to the provider API. Depending on the provider it could be getting instance sizes, zones, etc. For Google, there will not be a UI portion that needs to talk to google so we don't want to whitelist a URL that rancher will not be using.

pkg/controllers/management/drivers/nodedriver/machine_driver.go
```diff
@@ -179,7 +181,22 @@ func (m *Lifecycle) download(obj *v3.NodeDriver) (*v3.NodeDriver, error) {
...
+	// we still want to be able to create a Google cloud credential to use with
+	// GKE. This fake driver flag allows us to go through the motions of
+	// handling the credential fields without actually needing to run the
+	// machine driver binary.
+	if driverName == "google" {
+		flags = []cli.Flag{
+			&cli.StringFlag{
+				Name: "google-auth-encoded-json",
+			},
+		}
```
dramich, line 195: Can we move this inside the `getCreateFlagsForDriver` call? We still want the flags for the "driver" so shortcut in there instead of here.

pkg/data/management/machinedriver_data.go
cbron: Do you know if this whitelist is only for machine ? I believe the UI will need to make calls to gcp to get data for GKE, but that might be a different mechanism.

dramich: Same whitelist technically. If the UI needs to do that then something needs to be whitelisted, but it should probably be `container.googleapis.com` based on https://cloud.google.com/kubernetes-engine/docs/reference/rest

vincent99: Machine is running arbitrary code and can call whatever it wants; the whitelist is only for the /meta/proxy the UI uses

pkg/data/management/machinedriver_data.go
```diff
@@ -44,6 +45,7 @@ var driverData = map[string]map[string][]string{
 	RackspaceDriver:    {"privateCredentialFields": []string{"apiKey"}},
 	SoftLayerDriver:    {"privateCredentialFields": []string{"apiKey"}},
 	Vmwaredriver:       {"publicCredentialFields": []string{"username", "vcenter", "vcenterPort"}, "privateCredentialFields": []string{"password"}},
+	GKEDriver:          {"privateCredentialFields": []string{"credential"}},
```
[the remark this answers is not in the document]
cbron: Follow up: yes UI will handle validation (both that field is required and creds are valid), but through meta proxy api calls.

pkg/data/management/machinedriver_data.go
cbron: We are aiming at using the GKE specific meta proxy for now which does not need this whitelist, only the generic `meta/proxy` does. We can leave this for now as it shouldn't hurt anything and we may support GCP VMs in future.

dramich: If we don't need it we should delete it. It's also not the correct URL for GCE.

cbron: That works also. Is that not right URL though ? Minus the `www` part:  `compute.googleapis.com`

pkg/controllers/management/drivers/nodedriver/machine_driver.go
cmurphy: done

pkg/data/management/machinedriver_data.go
cmurphy: It depends on which API is being used, I had been thinking this whitelist would be used to proxy auth requests and I had thought that `www.googleapis.com` would be used for that, but it might actually be `accounts.google.com` or `oauth2.googleapis.com`, I'm not sure. Removing it for now.

pkg/data/management/machinedriver_data.go
```diff
@@ -94,6 +96,9 @@ func addMachineDrivers(management *config.ManagementContext) error {
 	if err := addMachineDriver(ExoscaleDriver, "local://", "", "", []string{"api.exoscale.ch"}, false, true, management); err != nil {
 		return err
 	}
+	if err := addMachineDriver(GoogleDriver, "local://", "", "", []string{}, true, true, management); err != nil {
```
dramich, line 99:
```suggestion
	if err := addMachineDriver(GoogleDriver, "local://", "", "", nil, true, true, management); err != nil {
```

cmurphy: fixed

dramich (approved):

cbron (approved):
Lgtm. @kinarashah we are going to merge now to un-block some other work, lmk if you see anything in here you'd like addressed.

Issue #30280, opened 2020-12-01
Support Keycloak using OIDC


maggieliu:
Please list the breakdown tickets below:

https://github.com/rancher/rancher/issues/30675
https://github.com/rancher/rancher/issues/30676

cbron:
@izaac @richard-cox backend has merged, UI already merged AFAIK. There may be some backend changes in future but we will track what those are and how to test them, we should start QA now/soon to see if there are larger problems.

izaac:
https://github.com/rancher/rancher/issues/33036

izaac:
Remaining issues update: 

https://github.com/rancher/rancher/issues/33337
https://github.com/rancher/rancher/issues/33314

izaac:
We're missing here the secure http certificate ready keycloak to test the certificates fields part of the Dashboard issue: https://github.com/rancher/dashboard/issues/2191
Most of the testing was done as part of https://github.com/rancher/rancher/issues/30280
Latest rancher build tested: `2.6 master-head (07/01/2021)` _c1151f6_

izaac:
We want this issue to be fixed first to do a final round of QA using only the Dashboard UI
https://github.com/rancher/dashboard/issues/3588

izaac:
Rancher version: `2.6-head (08/10/2021)` _723f747_

The dashboard issue is not a blocker and it was validated today, a validation pass was done to check site, cluster and project access.

A nestedgroup issue was identified in the process and tracked as a separate issue as it's not Keycloak specific.
https://github.com/rancher/rancher/issues/34142

PR #31572, opened 2021-03-03
OIDC change

brendarearden:
**Problem**
Keylcoak has OIDC support, but we currently only offer authentication with keycloak using SAML

**Solution**
Add OIDC support for Keycloak so that users can be authenticated with a keycloak client that uses OIDC protocol

**Issue**
https://github.com/rancher/rancher/issues/30280


pkg/apis/management.cattle.io/v3/authn_types.go
```diff
@@ -407,3 +407,32 @@ type ShibbolethConfig struct {
...
+	Scopes           string `json:"scope", norman:"required,default=openid|profile|email,notnullable,required"`
+	ResponseType     string `json:"responseType", norman:"required"`
+	GrantType        string `json:"grantType", norman:"type=enum,default=authorization_code,notnullable,required"`
+	TokenEndpoint    string `json:"tokenEndpoint,omitempty"`
+	AuthEndpoint     string `json:"authEndpoint,omitempty"`
+	UserInfoEndpoint string `json:"userInfoEndpoint,omitempty"`
+	Issuer           string `json:"issuer" norman:"required,notnullable"`
+	Certificate      string `json:"certificate,omitempty"`
+	PrivateKey       string `json:"privateKey" norman:"type=password"`
+	RancherUrl       string `json:"rancherUrl" norman:"required,notnullable"`
```
prachidamle, line 427: should be RancherURL

paynejacob (dismissed):

pkg/auth/providers/oidc/oidc_client.go
```diff
@@ -0,0 +1,210 @@
...
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("[generic oidc] unable to read response body: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("[generic oidc] response Status: %s. Body: %s", resp.Status, body)
+	}
+
```
prachidamle, line 153: The code line # 127 to line # 152 can be moved to a method and called from all the above functions that make a request and unmarshal the response body

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,418 @@
...
+	userPrincipal = o.toPrincipal(UserType, userInfo, nil)
+	logrus.Debugf("[generic oidc] loginuser: Checking user's access to Rancher")
+	allowed, err := o.userMGR.CheckAccess(config.AccessMode, config.AllowedPrincipalIDs, userPrincipal.Name, nil)
+	if err != nil {
+		return userPrincipal, groupPrincipal, "", err
+	}
+	if !allowed {
+		return userPrincipal, groupPrincipal, "", httperror.NewAPIError(httperror.Unauthorized, "unauthorized")
+	}
+	return userPrincipal, groupPrincipal, "", nil
```
prachidamle, line 128: so is there no groupPrincipals found here?

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,418 @@
...
+	} else {
+		p.PrincipalType = GroupType
+		if token != nil {
+			p.MemberOf = o.tokenMGR.IsMemberOf(*token, p)
+		}
+	}
+	return p
+}
+
+func (o *oidcProvider) toPrincipalFromPrincipal(principalType string, princ v3.Principal, token *v3.Token) v3.Principal {
```
prachidamle, line 235: maybe we should name this getPrincipalFromToken?

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: So I was able to add the user's groups using the original method we talked about, but there is still no way just to search groups - I am getting groups from the user's info

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: renamed

paynejacob (approved):

pkg/auth/providers/oidc/oidc_provider.go
prachidamle: yes groups list from the user info during login only is fine, all SAML providers do that anyway since searching groups is not possible.

scripts/ci
```diff
@@ -5,6 +5,6 @@ cd $(dirname $0)
 
 ./validate
 ./build
-./test
+#./test
 ./package
-./chart/ci
+#./chart/ci
```
prachidamle: This needs to be removed from the commit

pkg/client/generated/management/v3public/zz_generated_oidc_provider.go
```diff
@@ -0,0 +1,28 @@
...
+	Created         string            `json:"created,omitempty" yaml:"created,omitempty"`
+	CreatorID       string            `json:"creatorId,omitempty" yaml:"creatorId,omitempty"`
+	Labels          map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
+	Name            string            `json:"name,omitempty" yaml:"name,omitempty"`
+	OwnerReferences []OwnerReference  `json:"ownerReferences,omitempty" yaml:"ownerReferences,omitempty"`
+	RedirectURL     string            `json:"redirectUrl,omitempty" yaml:"redirectUrl,omitempty"`
+	Removed         string            `json:"removed,omitempty" yaml:"removed,omitempty"`
+	Type            string            `json:"type,omitempty" yaml:"type,omitempty"`
+	UUID            string            `json:"uuid,omitempty" yaml:"uuid,omitempty"`
+}
```
prachidamle, line 28: Add Generated files in a separate commit

cbron (changes requested):

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return userPrincipal, groupPrincipal, "", err
+	}
+
+	userInfo.Claims(&claimInfo)
+	userPrincipal = o.userToPrincipal(userInfo, claimInfo)
+
+	for _, group := range claimInfo.Groups {
+		groupPrincipal = append(groupPrincipal, o.groupToPrincipal(group))
+	}
```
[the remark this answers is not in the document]
brendarearden: So this is implemented similarly to how SAML is. We do get which groups a user belongs to, but we do not have the ability to search all of the groups available within keycloak with the current implementation.

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,353 @@
...
+	p[publicclient.OIDCProviderFieldRedirectURL] = o.getRedirectURL(authConfig)
+	return p, nil
+}
+
+func (o *oidcProvider) getRedirectURL(config map[string]interface{}) string {
+	return fmt.Sprintf(
+		"%s?client_id=%s&response_type=code&redirect_uri=%s",
+		config["authEndpoint"],
+		config["clientId"],
+		config["rancherUrl"],
```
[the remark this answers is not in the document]
brendarearden: Yes it does get set by the UI. I do not think we need validation as this is just a helper string builder method.

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,353 @@
...
+
+func (o *oidcProvider) userToPrincipal(userInfo *oidc.UserInfo, info ClaimInfo) v3.Principal {
+	displayName := info.Name
+	if displayName == "" {
+		displayName = userInfo.Email
+	}
+	p := v3.Principal{
+		ObjectMeta:    metav1.ObjectMeta{Name: Name + "_" + UserType + "://" + userInfo.Subject},
+		DisplayName:   displayName,
+		LoginName:     userInfo.Email,
```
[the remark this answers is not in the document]
brendarearden: with the go-oidc package, you get the generic user info response and on that response it has all of the claims it could retrieve from keycloak. So to get these more specific claim values, I parsed them into the ClaimInfo object so that I could get the user's name - so info.Name is whatever the name value is in keycloak when creating a user. Yes, this does  mean that the user's email will be in the users profile.

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	return userPrincipal, groupPrincipal, oauth2Token.AccessToken, nil
+}
+
+func (o *oidcProvider) SearchPrincipals(searchValue, principalType string, token v3.Token) ([]v3.Principal, error) {
+	var principals []v3.Principal
+	if principalType == "" {
+		principalType = UserType
+	}
+
```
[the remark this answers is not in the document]
brendarearden: Again, this is how SAML is set up - since we don't have the ability to query all the users in the user management system at this time, then all we can return is a non search result. We could always opt for a "no implemented" error instead if that would be preferred.

pkg/auth/providers/oidc/oidc_actions.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/rancher/norman/api/handler"
+	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
+	managementschema "github.com/rancher/rancher/pkg/schemas/management.cattle.io/v3"
+
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+
+	"github.com/pkg/errors"
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+	"github.com/rancher/rancher/pkg/auth/providers/common"
```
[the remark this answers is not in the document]
brendarearden: I have sorted these

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	return false
+}
+
+func (o *oidcProvider) GetUserExtraAttributes(token *v3.Token) map[string][]string {
+	extras := make(map[string][]string)
+	extras["principalid"] = []string{token.UserPrincipal.Name}
+	extras["username"] = []string{token.UserPrincipal.LoginName}
+	return extras
+}
```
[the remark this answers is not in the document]
brendarearden: I have added a verifier for the id token as part of the recent commit

pkg/client/generated/management/v3public/zz_generated_oidc_provider.go
brendarearden: I now have a separate commit for the generated files

tests/integration/suite/test_auth_configs.py
```diff
@@ -15,7 +15,7 @@ def test_auth_configs(admin_mc):
 
     configs = client.list_auth_config()
 
-    assert configs.pagination.total == 12
+    assert configs.pagination.total == 13
```
cbron, line 18: wouldn't this be 14 now ?

pkg/schemas/management.cattle.io/v3/schema.go
```diff
@@ -618,7 +618,43 @@ func authnTypes(schemas *types.Schemas) *types.Schemas {
...
+		}).
+		MustImport(&Version, v3.OIDCApplyInput{}).
+		MustImport(&Version, v3.OIDCTestOutput{}).
+		//KeyCloakOIDC Config
+		MustImportAndCustomize(&Version, v3.KeyCloakOIDCConfig{}, func(schema *types.Schema) {
+			schema.BaseType = "authConfig"
+			schema.ResourceActions = map[string]types.Action{
+				"disable": {},
+				"configureTest": {
+					Input:  "keycloakoidcConfig",
```
cbron, line 646: Confused about casing here, why is it all lowercase but config ?

pkg/auth/providers/publicapi/login.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/rancher/rancher/pkg/auth/providers/keycloakoidc"
+
+	"github.com/rancher/rancher/pkg/auth/providers/oidc"
```
cbron, line 13: lint

pkg/apis/management.cattle.io/v3/public_authn_types.go
```diff
@@ -174,3 +174,24 @@ type SamlLoginInput struct {
...
+}
+
+type OIDCLogin struct {
+	GenericLogin `json:",inline"`
+	Code         string `json:"code" norman:"type=string,required"`
+}
+
+type KeyCloakOIDCProvider struct {
+	OIDCProvider `json:",inline"`
+}
```
cbron, line 197: Do we need KeyCloak Login ?

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/sirupsen/logrus"
+
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+)
+
+type searchResults struct {
+	Items []Account `json:"items"`
+}
+
+//Account defines properties an account on github has
```
cbron, line 22: copy paste error

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	var searchResult Account
+	if err := json.Unmarshal(b, &searchResult); err != nil {
+		logrus.Errorf("[keycloak oidc]: received error unmarshalling search results, err: %v", err)
+		return searchResult, err
+	}
+	return searchResult, nil
+
+}
+func (k *KClient) getSearchUrl(issuer string) string {
```
cbron, line 69: lint

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+//Account defines properties an account on github has
+type Account struct {
+	ID       int      `json:"id,omitempty"`
+	Username string   `json:"username,omitempty"`
+	Enabled  bool     `json:"enabled,omitempty"`
+	Name     string   `json:"firstName,omitempty"`
+	LastName string   `json:"lastName,omitempty"`
+	Groups   []string `json:"groups"`
```
cbron, line 29: Shouldn't email be in here ?

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	return u.String()
+}
+
+func (k *KClient) getFromKeyCloak(accessToken string, url string) ([]byte, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("Authorization", "token "+accessToken)
```
cbron, line 93: We are doing a GET request with an access token ? Can you not do a POST ? Or is the accessToken not sensitive ?

pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -0,0 +1,205 @@
...
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/types/config"
+	"github.com/rancher/rancher/pkg/user"
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	Name      = "keycloakOidc"
```
cbron, line 25: weird casing again

pkg/auth/providers/oidc/oidc_provider.go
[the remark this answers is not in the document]
cbron: I'm assuming this is outdated now.

cbron (changes requested):

pkg/auth/providers/keycloakoidc/keycloak_client.go
dramich: Why would a POST be used to GET data? A POST doesn't hide the access token

cbron: It does if it's in the body. I suppose this depends how keycloak can receive it.

pkg/apis/management.cattle.io/v3/public_authn_types.go
brendarearden: The keycloak configuration is going to be using the OIDCLogin struct for it so we shouldn't need to make more structs just for keycloak

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: I added email with my recent changes.

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: This is resolved in my recent updates.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	defer resp.Body.Close()
+	// Check the status code
+	switch resp.StatusCode {
+	case 200:
+	case 201:
+	default:
+		var body bytes.Buffer
+		io.Copy(&body, resp.Body)
+		return nil, fmt.Errorf("[keycloak oidc] request failed, got status code: %d. Response: %s",
+			resp.StatusCode, body.Bytes())
```
cbron, line 175: Would we want to return the body on error ? It could end up in various logs, and we don't know what data will be exposed. Possible that the error message can contain user data.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	resp, err := k.httpClient.Do(req)
+	if err != nil {
+		logrus.Errorf("[keycloak oidc] received error from keycloak: %v", err)
+		return nil, err
+	}
+	defer resp.Body.Close()
+	// Check the status code
+	switch resp.StatusCode {
+	case 200:
+	case 201:
```
cbron, line 170: So we only accept 200/201 and otherwise fail ? Thats somewhat strict but it might work. Does keycloak specifically talk about those 2 codes ? I'm worried if you get a 301 in the middle you'll end up in the error section, but perhaps that case would be incorrectly configured.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	if len(group.Subgroups) > 0 {
+		for _, sub := range group.Subgroups {
+			groups = append(groups, sub)
+			groups = append(groups, getSubGroups(sub)...)
+		}
+	}
+	return groups
+}
+
+func (k *KClient) keycloakGroupSearch(searchTerm, searchType string, accessToken string, config *v32.OIDCConfig) ([]Group, error) {
```
cbron, line 103: This isn't used anywhere ?

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	var searchResult Account
+	if err := json.Unmarshal(b, &searchResult); err != nil {
+		logrus.Errorf("[keycloak oidc]: received error unmarshalling search results, err: %v", err)
+		return searchResult, err
+	}
+	return searchResult, nil
+
+}
+func (k *KClient) getSearchURL(issuer string) string {
+	splitIssuer := strings.SplitAfter(issuer, "/auth/")
```
cbron, line 136: Do we have validation that issuer always has this format ?

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,389 @@
...
+	var err error
+
+	if config == nil {
+		config, err = o.GetOIDCConfig()
+		if err != nil {
+			return userPrincipal, nil, "", err
+		}
+	}
+
+	if config.Certificate != "" && config.PrivateKey != "" {
```
cbron, line 97: Wouldn't this also need to go in the `getFromKeycloak` stuff ? It looks like those are the two cases where Rancher is reaching out to keycloak server.

tests/integration/suite/test_auth_configs.py
```diff
@@ -55,9 +57,12 @@ def test_auth_configs(admin_mc):
             googleoauth = c
         elif c.type == "shibbolethConfig":
             shibboleth = c
+        elif c.type == "oidcConfig":
+            oidc = c
```
cbron, line 61: Needs keycloakOIDC too

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	var accounts []Account
+
+	if principalType == "" || principalType == UserType {
+		var userAccounts []Account
+		searchURL := fmt.Sprintf("%s/%ss?search=%s", k.getSearchURL(config.Issuer), UserType, searchTerm)
+		search := URLEncoded(searchURL)
+
+		b, err := k.getFromKeyCloak(accessToken, search)
+		if err != nil {
+			// no match on search returns an error. do not log
```
prachidamle, line 55: Is there a specific error returned when there is no match? It is better to match against specific error code or string and decide to ignore, for other errors we need to bubble the error up.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	if principalType == "" || principalType == GroupType {
+		var groups []Group
+		searchURL := fmt.Sprintf("%s/%ss?search=%s", k.getSearchURL(config.Issuer), GroupType, searchTerm)
+		search := URLEncoded(searchURL)
+
+		b, err := k.getFromKeyCloak(accessToken, search)
+		if err != nil {
+			// no match on search returns an error. do not log
+			return nil, nil
```
prachidamle, line 75: same comment as above on specific error code or message returned

pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+type keycloakoidcProvider struct {
+	keycloakoidcClient *KClient
+	oidcProvider       oidc.Provider
+}
+
+func Configure(ctx context.Context, mgmtCtx *config.ScaledContext, userMGR user.Manager, tokenMGR *tokens.Manager) common.AuthProvider {
+	keycloakoidcClient := &KClient{
+		httpClient: &http.Client{},
```
prachidamle, line 37: Do we have a way to set  request timeout? Also why don't we configure cert and private key on the client?

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,389 @@
...
+		caCertPool.AppendCertsFromPEM([]byte(config.Certificate))
+		oidcClient := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					Certificates: []tls.Certificate{cert},
+					RootCAs:      caCertPool,
+				},
+			},
+		}
+		ctx = oidc.ClientContext(ctx, oidcClient)
```
prachidamle, line 112: Should this be handled for each connection? Maybe need a createConnection utility method to set all the properties every time?

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return accounts, nil
+}
+
+func getSubGroups(group Group) []Group {
+	var groups []Group
+	if len(group.Subgroups) > 0 {
+		for _, sub := range group.Subgroups {
+			groups = append(groups, sub)
+			groups = append(groups, getSubGroups(sub)...)
```
prachidamle, line 97: Does the above ""getFromKeyCloak search request return all group detail needed for all nested groups? Is there any limit/level upto which subgroups are listed back?

prachidamle: Shouldn't getSubGroups(sub) call the provider to search for the subgroup attributes?

prachidamle: Seems like we need some limit configured for this recursive call, is there a max level of nested groups set by keycloak? If not maybe we should add a default limit in our config

brendarearden: Yeah I was also wondering about this when I coded it - I checked in the google provider to get an idea for a default amount and I didn't see one in there for the subgroup call - I did not kind any keycloak documentation with a limit identified for how many subgroups are possible either. Is there a specific default amount in mind that we think would be good?

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: From what I have seen, when a search returns nothing, it is a 401 not found. Either way I am ignoring the error messages later so if this seems too strict I can remove the case
```		
b, err := k.getFromKeyCloak(accessToken, search)
if err != nil {
  // no match on search returns an error. do not log
  return nil, nil
}
```

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: I added validation that the issuer was in expected format prior to splitting it to ensure the search url function works as expected. since it is keycloak specific search we can expect the issuer to always be in this specific format.

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: This was left behind after a refactor - I have removed it as part of the recent commit.

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: I think you and @cbron had similar questions - the error returned when nothing is found is a 401, with a message saying "nothing was found". We can match against that error code if we would like and then error error when other codes are retuned.

pkg/schemas/management.cattle.io/v3/schema.go
brendarearden: changed the casing - there was no specific need for the casing to be all lower

pkg/auth/providers/keycloakoidc/keycloak_client.go
cbron: This would fail before create happens, so we don't need to check here.

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: Based on how the other providers are handling their get requests, it seems fairly standard to keep it is currently.

pkg/auth/providers/keycloakoidc/keycloak_provider.go
brendarearden: After looking into this a bit more, postman had been adding the cert/key on my other calls, so I added the cert and key to all calls. 

There isn't a way to set request timeout. Do we have a specific timeout limit we want to add?

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: Added this as a common util so other providers could use it if they needed.

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: see comment above

pkg/schemas/management.cattle.io/v3/schema.go
```diff
@@ -618,7 +618,43 @@ func authnTypes(schemas *types.Schemas) *types.Schemas {
...
+				},
+				"testAndApply": {
+					Input: "keyCloakOidcApplyInput",
+				},
+			}
+			schema.CollectionMethods = []string{}
+			schema.ResourceMethods = []string{http.MethodGet, http.MethodPut}
+		}).
+		MustImport(&Version, v3.OIDCApplyInput{}).
+		MustImport(&Version, v3.OIDCTestOutput{})
```
dramich, line 657: These two are extra

pkg/auth/providers/publicapi/login.go
```diff
@@ -8,6 +8,10 @@ import (
...
+
```
dramich, line 14: Fix formatting

pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -0,0 +1,206 @@
...
+	return k.oidcProvider.GetOIDCConfig()
+}
+
+func (k *keycloakoidcProvider) isThisUserMe(me v3.Principal, other v3.Principal) bool {
+	return k.oidcProvider.IsThisUserMe(me, other)
+}
+
+func (k *keycloakoidcProvider) GetUserExtraAttributes(token *v3.Token) map[string][]string {
+	return k.oidcProvider.GetUserExtraAttributes(token)
+}
```
dramich, line 206: Some of these are needed, to call the OIDC provider, the ones that are not exported should not be here

pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -0,0 +1,206 @@
...
+
+const (
+	Name      = "keycloakoidc"
+	UserType  = "user"
+	GroupType = "group"
+)
+
+type keycloakoidcProvider struct {
+	keycloakoidcClient *KClient
+	oidcProvider       oidc.Provider
```
dramich, line 33: This could potentially be embedded then the below wrapper functions won't be needed

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -0,0 +1,385 @@
...
+	} else {
+		princ.PrincipalType = GroupType
+		if token != nil {
+			princ.MemberOf = o.TokenMGR.IsMemberOf(*token, princ)
+		}
+	}
+	return princ
+}
+
+func (o *Provider) SaveOIDCConfig(config *v32.OIDCConfig) error {
```
dramich, line 300: This doesn't need to be exported

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/rancher/rancher/pkg/auth/providers/common"
+
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
+)
+
+type searchResults struct {
```
dramich, line 19: This isn't used

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	Items []Account `json:"items"`
+}
+
+//Account defines properties an account in keycloak has
+type Account struct {
+	ID            int    `json:"id,omitempty"`
+	Email         string `json:email,omitempty`
+	EmailVerified string `json:emailVerified,omitempty`
```
dramich, line 27: Missing quotes for the json tags

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	Name          string `json:"firstName,omitempty"`
+	LastName      string `json:"lastName,omitempty"`
+	Type          string
+}
+
+//Group defines properties a group in keycloak has
+type Group struct {
+	ID        int     `json:"id,omitempty"`
+	Name      string  `json:"name,omitempty"`
+	Subgroups []Group `json:subGroups,omitempty`
```
dramich, line 39: Missing quotes for the json tags

pkg/auth/providers/keycloakoidc/keycloak_client.go
dramich: We ned to handle the errors here so we don't eat them. It can be more intelligent and check error types but in general we need to return something.

pkg/auth/providers/common/util.go
```diff
@@ -0,0 +1,40 @@
+package common
```
dramich, line 1: Can this move into a provider specific client file?

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return u.String()
+}
+
+func (k *KClient) getFromKeyCloak(accessToken, url, certificate, key string) ([]byte, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	err = common.GetClientWithCertKey(k.httpClient, certificate, key)
```
dramich, line 155: Can we do this one time when we initialize `KClient` with the assumption that login would still need to do this

pkg/auth/providers/keycloakoidc/keycloak_client.go
dramich: We need to handler errors properly

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -0,0 +1,179 @@
...
+	case 200:
+	case 201:
+	default:
+		var body bytes.Buffer
+		io.Copy(&body, resp.Body)
+		return nil, fmt.Errorf("[keycloak oidc] request failed, got status code: %d. Response: %s",
+			resp.StatusCode, body.Bytes())
+	}
+	b, err := ioutil.ReadAll(resp.Body)
+	return b, err
```
dramich, line 178: Just return `b, err := ioutil.ReadAll(resp.Body)` and let the called handler the data and the error.

dramich (changes requested):

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: I have changed the code to just return the body, error code and error now and handle the additional checks where I call this method

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: removed the functions from being exported that were not longer used outside the oidc package

pkg/auth/providers/keycloakoidc/keycloak_provider.go
brendarearden: I have now embedded this and do not make the extra functions as I was previously for the wrapper functionality. I am only making the specific functions for the keycloak specific searches

pkg/auth/providers/keycloakoidc/keycloak_provider.go
brendarearden: I removed the calls from being exported that did not need to be

pkg/auth/providers/publicapi/login.go
brendarearden: fixed

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: errors are now being handled where this func is called

pkg/schemas/management.cattle.io/v3/schema.go
brendarearden: removed the extra "MustImport"s

pkg/auth/providers/common/util.go
brendarearden: moved this code to an oidc_client file

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: because we need the configuration information, we are not able to do it as part of the init, but I did move this code so that it is only happening in twice in the keycloak_provider code - once for search principals and once for get principals.

PR #31585, opened 2021-03-04
Add cloudprovider tolerations to helm install pod

StrongMonkey:


pkg/catalogv2/helmop/operation.go
```diff
@@ -766,6 +766,12 @@ func (s *Operations) createPod(secretData map[string][]byte) (*v1.Pod, *podimper
 					Value:    "true",
 					Effect:   "NoExecute",
 				},
+				{
+					Key:      "node.cloudprovider.kubernetes.io/uninitialized",
+					Operator: corev1.TolerationOpEqual,
```
deniseschannon, line 771: why isn't this "Equal" like the other tolerations?

deniseschannon (approved):

Issue #31461, opened 2021-02-24
GKE v2 Provisioner: add Register Cluster option


gaktive:
For the GKE v2 Provisioner, we need to add a function to register a cluster. This should match what we do in EKS currently (let us know if you need details on that).

cmurphy:
This will need the UI to be done to fully test. In the UI, you should be able to select a cloud credential, and then be able to select from a list of clusters that already exist in GKE. Should test that it gets registered successfully, that you have kubectl access to it, that properties like the kubernetes version are visible from the rancher UI, and that changes made from Rancher or from the GKE console don't overwrite each other.

prabalsharma:
This was tested via test plan

Issue #31110, opened 2021-02-03
GKE Provisioner: set number of pods per node


cbron:
We should support editing the number of allowed pods per node. By default this is 110. This requires 2 parts: 
1. The CIDR field on the node
2. The kubelet max pod setting

More info: https://cloud.google.com/kubernetes-engine/docs/how-to/flexible-pod-cidr

prabalsharma:
PASSED

Tested on gke-operator rc4

```
ClusterConfig:
    clusterIpv4CidrBlock: 10.100.0.0/16
    createSubnetwork: true
    nodeIpv4CidrBlock: 10.101.0.0/16
    servicesIpv4CidrBlock: 10.102.0.0/16
name: example-node-pool-1
    maxPodsConstraint: 55
name: example-node-pool-2
    maxPodsConstraint: 110


gcloud container clusters describe:
  clusterIpv4Cidr: 10.100.0.0/16
  clusterIpv4CidrBlock: 10.100.0.0/16
  clusterSecondaryRangeName: gke-prabal-rc4-cluster-b-pods-23bb97f7
  createSubnetwork: true
  servicesIpv4Cidr: 10.102.0.0/16
  servicesIpv4CidrBlock: 10.102.0.0/16
  servicesSecondaryRangeName: gke-prabal-rc4-cluster-b-services-23bb97f7
  subnetworkName: new-example-subnet-for-cluster-b
name: example-node-pool-1
    maxPodsPerNode: '55'
    podIpv4CidrSize: 25
name: example-node-pool-2
    maxPodsPerNode: '110'
    podIpv4CidrSize: 24
```
```
Cluster Config:
maxPodsConstraint: 8
clusterIpv4Cidr: ""

gcloud:
maxPodsPerNode: '8'
podIpv4CidrSize: 28
  clusterIpv4Cidr: 10.76.0.0/14
  clusterIpv4CidrBlock: 10.76.0.0/14
  servicesIpv4Cidr: 10.80.0.0/20
  servicesIpv4CidrBlock: 10.80.0.0/20
```
```
Cluster Config:
maxPodsConstraint: 220
Error 400: Maximum pods per node must be at least 8 and at most 110, current number is 220
```

cbron:
@cmurphy can you link PRs in issues
@prabalsharma if the PRs go to rancher ensure that it is in both master and release/v2.5 branches

cmurphy:
How to test:

On GKEClusterConfigSpec set:

```
nodePools:
[...]

Issue #31011, opened 2021-01-27
GKE Provisioner: Handler


gaktive:
We need another controller that manages the translation from GCP to Rancher. This would replicate what's in place for EKS

Issue #31010, opened 2021-01-27
GKE Provisioner: Sync


gaktive:
This is to work on a controller that manage syncing data between GKE and Rancher

gaktive:
EKS did something like this (I think): https://github.com/rancher/rancher/issues/28422

prabalsharma:
sync works in general, on 2.5-head
added to https://github.com/rancher/rancher/issues/32140 for master testing

PR #31625, opened 2021-03-08
Integrate GKE operator

cmurphy:
Add GKE v2 integration controller and cluster refresher.

https://github.com/rancher/rancher/issues/31011
https://github.com/rancher/rancher/issues/31010
https://github.com/rancher/rancher/issues/31461
https://github.com/rancher/rancher/issues/31895
https://github.com/rancher/rancher/issues/31110


cmurphy:
Marking this as ready for review so I can start working on feedback even though it is awaiting final approval of https://github.com/rancher/rancher/pull/31674

rmweir (changes requested):
Skipped any parts related to syncing as that is subject to change. We should probably create a package contaaining logic that is used in multiple controllers such as deploy<provider>Operator, setUnknown, setTrue, setFalse, aand generateValuesYaml.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -534,3 +538,173 @@ func validateEKS(prevCluster, newCluster map[string]interface{}) error {
...
+	}
+
+	// check user's access to cloud credential
+	googleCredential, ok := gkeConfig["googleCredentialSecret"].(string)
+	if !ok {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, "must provide googleCredentialSecret")
+	}
+	if err := validateGKECredentialAuth(request, googleCredential, prevCluster); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
cmurphy: Why would this be different from the EKS credential? https://github.com/rancher/rancher/blob/1031efc0fcf9078037c841c4d8bcc3af92b78360/pkg/api/norman/customization/cluster/validator.go#L355-L360

I could move it down past line 582 to validate for creates only but I think the credential still needs to be present and validated in the import case?

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -534,3 +538,173 @@ func validateEKS(prevCluster, newCluster map[string]interface{}) error {
...
+	for _, np := range nodepools {
+		name := *np.Name
+		if name == "" {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("node pool name cannot be an empty string"))
+		}
+
+		version := np.Version
+		if version == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
cmurphy: It would end up getting rejected by the operator. Fixing here.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -534,3 +538,173 @@ func validateEKS(prevCluster, newCluster map[string]interface{}) error {
...
+	zone := spec.GKEConfig.Zone
+	msgSuffix := fmt.Sprintf("in region [%s]", region)
+	if region == "" {
+		msgSuffix = fmt.Sprintf("in zone [%s]", spec.GKEConfig.Zone)
+	}
+
+	// cluster client is being used instead of lister to avoid the use of an outdated cache
+	clusters, err := client.List(metav1.ListOptions{})
+	if err != nil {
+		return httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("failed to confirm clusterName is unique among Rancher GKE clusters "+msgSuffix, region))
```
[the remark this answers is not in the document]
cmurphy: Yes. Fixing.

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -123,6 +123,11 @@ func (cd *clusterDeploy) doSync(cluster *v3.Cluster) error {
 		return nil
 	}
 
+	_, err := cd.systemAccountManager.GetOrCreateSystemClusterToken(cluster.Name)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
cmurphy: Thanks for bring this up because I was confused about it, see the commit message:

```
Create system cluster token on first sync attempt

If a cluster provisioner has deployed a remote cluster with a private
control plane endpoint, it will need to have the cluster agent deployed
manually. Without this patch, it will not be able to serve the agent
YAML manifest yet because doSync exits early since it cannot find any
nodes. In order to be able to generate the agent manifest URL, this
change goes ahead and generates the cluster registration token as soon
as the cluster is deployed.
```

In short, when deploying a private cluster in GKE I couldn't do the manual agent install step without this. I don't know why it's a problem here and not a problem for EKS, do you?

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,706 @@
...
+			if err != nil {
+				return cluster, err
+			}
+		}
+
+		if cluster.Spec.GKEConfig.Imported {
+			if cluster.Status.GKEStatus.UpstreamSpec == nil {
+				// non imported clusters will have already had upstream spec set
+				return e.setInitialUpstreamSpec(cluster)
+			}
```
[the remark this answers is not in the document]
cmurphy: Oops, fixed

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,706 @@
...
+			// Check to see if we can still use the control plane endpoint even though
+			// the cluster has private-only access
+			serviceToken, mustTunnel, err := e.generateSATokenWithPublicAPI(cluster)
+			if mustTunnel {
+				cluster = cluster.DeepCopy()
+				cluster.Status.GKEStatus.PrivateRequiresTunnel = &mustTunnel
+				cluster.Status.ServiceAccountToken = serviceToken
+				return e.clusterClient.Update(cluster)
+			}
+			if err != nil {
```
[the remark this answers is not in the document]
cmurphy: Copypasta from eks_cluster_handler. In this case `generateSATokenWithPublicAPI` never returns an error if mustTunnel is true but I'll fix it to future-proof it.

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,701 @@
...
+
+		e.clusterEnqueueAfter(cluster.Name, enqueueTime)
+		if failureMessage == "" {
+			logrus.Infof("waiting for cluster GKE [%s] to finish creating", cluster.Name)
+			return e.setUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "")
+		}
+		logrus.Infof("waiting for cluster GKE [%s] create failure to be resolved", cluster.Name)
+		return e.setFalse(cluster, apimgmtv3.ClusterConditionProvisioned, failureMessage)
+	case "active":
+		if cluster.Status.GKEStatus.UpstreamSpec == nil {
```
cbron, line 214: This is slight departure from EKS where we always check imported first. We are saying now that any type of cluster where upstream is nil we want to SetInitialUpstreamSpec

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,701 @@
...
+			if apimgmtv3.ClusterConditionPending.IsUnknown(cluster) {
+				cluster = cluster.DeepCopy()
+				apimgmtv3.ClusterConditionPending.True(cluster)
+				cluster, err = e.clusterClient.Update(cluster)
+				if err != nil {
+					return cluster, err
+				}
+			}
+		}
+
```
cbron, line 241: Shouldnt we be making sure there is at least one nodegroup here ?

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,701 @@
...
+		return "", err
+	}
+
+	restConfig := &rest.Config{
+		Host: endpoint,
+		TLSClientConfig: rest.TLSClientConfig{
+			CAData: decodedCA,
+		},
+		WrapTransport: func(rt http.RoundTripper) http.RoundTripper {
+			return &oauth2.Transport{
```
cbron, line 624: Good thing we aren't using bearer

pkg/controllers/management/clusterdeploy/clusterdeploy.go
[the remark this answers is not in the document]
cbron: @thedadams can you review this, and also the last two commits in this PR for private cluster approach.

cbron (commented):
It's in description but this depends on integration PR first. 
Kinda sad we have to copy so much code and not use shared package but oh well.

go.mod
```diff
@@ -10,6 +10,7 @@ replace (
 	github.com/rancher/rancher/pkg/apis => ./pkg/apis
 	github.com/rancher/rancher/pkg/client => ./pkg/client
 	golang.org/x/sys => golang.org/x/sys v0.0.0-20190921190940-14da1ac737cc // This is because docker/docker/pkg/archive has a dependency on an old x/sys
+	google.golang.org/grpc => google.golang.org/grpc v1.29.1
```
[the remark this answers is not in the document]
cmurphy: `go mod tidy` fails without it due to this https://github.com/etcd-io/etcd/issues/12124

Adding a comment

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: Fixed

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: The validator in `pkg/api/norman/customization/cluster/validator.go` ensures there is at least one node pool.

pkg/controllers/management/clusterprovisioner/provisioner.go
```diff
@@ -108,6 +108,10 @@ func (p *Provisioner) Remove(cluster *v3.Cluster) (runtime.Object, error) {
 		logrus.Debugf("EKS cluster [%s] will be managed by eks-operator-controller, skipping remove", cluster.Name)
 		return cluster, nil
 	}
+	if cluster.Spec.GKEConfig != nil {
+		logrus.Debugf("GKE cluster [%s] will be managed by gke-operator-controller, skipping remove", cluster.Name)
+		return cluster, nil
+	}
```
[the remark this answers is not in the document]
cmurphy: Added a `skipOperatorCluster` function for this.

cmurphy:
> We should probably create a package contaaining logic that is used in multiple controllers such as deployOperator, setUnknown, setTrue, setFalse, aand generateValuesYaml.

Added a commit to do this.

pkg/systemtemplate/import.go
```diff
@@ -75,7 +75,7 @@ func SystemTemplate(resp io.Writer, agentImage, authImage, namespace, token, url
 		tolerations = templates.ToYAML(taints)
 	}
 
-	if cluster != nil && cluster.Spec.AgentEnvVars != nil {
+	if cluster != nil && cluster.Spec.AgentEnvVars != nil && len(cluster.Spec.AgentEnvVars) > 0 {
```
thedadams, line 78: A `nil` array has length 0.
```suggestion
	if cluster != nil && len(cluster.Spec.AgentEnvVars) > 0 {
```

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -534,3 +538,170 @@ func validateEKS(prevCluster, newCluster map[string]interface{}) error {
...
+		}
+	}
+
+	if len(errors) != 0 {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf(strings.Join(errors, ";")))
+	}
+	return nil
+}
+
+func validateGKEClusterName(client v3.ClusterInterface, spec *v32.ClusterSpec) error {
```
thedadams, line 675: There is already a name-collision validation check for all clusters. A user cannot have two clusters with the exact same name: https://github.com/rancher/rancher/blob/9a3562666ec2022358153d2c7a9193964ec57db3/pkg/api/norman/store/cluster/cluster_store.go#L613

pkg/controllers/management/clusterdeploy/clusterdeploy.go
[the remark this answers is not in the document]
thedadams: The way this works in the UI now is the following:
1. A user creates a private cluster.
2. Once the cluster is running, the UI gives the user the option to see the import command.
3. If the user clicks to see the import command, the UI makes a GET request to `/v3/clusterregistrationtokens` to see if the cluster has any registration tokens.
3a. If the cluster does not have any registration tokens, the UI makes a POST request to `/v3/clusterregistrationtokens` to create a token and the repeats step (3).

You are essentially doing (3) and (3a) automatically for every cluster that is created in Rancher. This is not the correct flow because not every cluster needs a registration token.

I believe removing this code and letting the UI handle this as above is the correct way to do it. I'm happy to help if you would like to test this with your development environment.

pkg/dialer/factory.go
```diff
@@ -77,6 +77,14 @@ func HasOnlyPrivateAPIEndpoint(cluster *v3.Cluster) bool {
 		}
 		return cluster.Status.EKSStatus.PrivateRequiresTunnel != nil &&
 			*cluster.Status.EKSStatus.PrivateRequiresTunnel
+	case v32.ClusterDriverGKE:
+		if cluster.Status.GKEStatus.UpstreamSpec != nil &&
+			cluster.Status.GKEStatus.UpstreamSpec.PrivateClusterConfig != nil &&
+			cluster.Status.GKEStatus.UpstreamSpec.PrivateClusterConfig.EnablePrivateEndpoint == false {
```
thedadams, line 83:
```suggestion
			!cluster.Status.GKEStatus.UpstreamSpec.PrivateClusterConfig.EnablePrivateEndpoint {
```

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,519 @@
...
+}
+
+func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+	if cluster == nil || cluster.DeletionTimestamp != nil {
+		return cluster, nil
+	}
+
+	if cluster.Spec.GKEConfig == nil {
+		return cluster, nil
+	}
```
thedadams, line 88: nit:
```suggestion
	if cluster == nil || cluster.DeletionTimestamp != nil || cluster.Spec.GKEConfig == nil {
		return cluster, nil
	}
```

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,519 @@
...
+	upstreamSpec, err := clusterupstreamrefresher.BuildGKEUpstreamSpec(e.SecretsCache, cluster)
+	if err != nil {
+		return cluster, err
+	}
+	cluster.Status.GKEStatus.UpstreamSpec = upstreamSpec
+	return e.ClusterClient.Update(cluster)
+}
+
+// updateGKEClusterConfig updates the GKEClusterConfig object's spec with the cluster's GKEConfig if they are not equal.
+func (e *gkeOperatorController) updateGKEClusterConfig(cluster *mgmtv3.Cluster, gkeClusterConfigDynamic *unstructured.Unstructured, spec map[string]interface{}) (*mgmtv3.Cluster, error) {
```
thedadams, line 308: It looks like this function could have a more generic name and moved to `clusteroperator/utils.go`. This would also require the EKS version to be deleted in favor of the more generic version.

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,519 @@
...
+			if err != nil {
+				return cluster, err
+			}
+			return cluster, nil
+		}
+	}
+}
+
+// recordCAAndAPIEndpoint reads the GKEClusterConfig's secret once available. The CA cert and API endpoint are then copied to the cluster status.
+func (e *gkeOperatorController) recordCAAndAPIEndpoint(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
```
thedadams, line 350: It also looks like this function could be moved to `clusteroperator/utils.go`.

pkg/controllers/management/clusteroperator/utils.go
```diff
@@ -0,0 +1,199 @@
+package clusteroperator
```
thedadams, line 1: This is excellent!

thedadams (changes requested):

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: This one is a little tricky because it invokes `e.recordAppliedSpec` which is not generic.

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: done

thedadams (dismissed):

pkg/api/norman/customization/cluster/validator.go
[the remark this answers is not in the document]
rmweir: In the EKS example nothing is happening if ok is false. In your code an error is being returned. I think I was incorrect about the non-import part, both need it but only on a POST request. It is important to consider that when a user is just updating a single field, that is the only field that will be sent instead of sending the entire cluster/config.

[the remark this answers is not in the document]
cmurphy: Got it, fixed by moving it down

thedadams (approved):

pkg/api/norman/customization/cluster/validator.go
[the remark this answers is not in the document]
rmweir: The difference is that in EKS if ok is false, then it is not checked. I was incorrect on the import part. We need to validate POST requests, be them for an import or non import cluster, or any PUT requests that are changing the actual field. It is important to note that when updating something in the UI that only the field(s) that are being changed may be sent.

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -0,0 +1,461 @@
...
+		return e.updateGKEClusterConfig(cluster, gkeClusterConfigDynamic, gkeClusterConfigMap)
+	}
+
+	// get gke Cluster Config's phase
+	status, _ := gkeClusterConfigDynamic.Object["status"].(map[string]interface{})
+	phase, _ := status["phase"]
+	failureMessage, _ := status["failureMessage"].(string)
+	if strings.Contains(failureMessage, "403") {
+		failureMessage = fmt.Sprintf("cannot access gke, check cloud credential: %s", failureMessage)
+	}
```
rmweir, line 152: Did you test if this works for gke? Like if you add a cloud credential, then delete it in gke, does this catch that?

rmweir (changes requested):
LGTM on everything non-syncing related besides 1 comment in this review and separate comment about credential validation. @mjura is nearly finished with syncing and it has changed quite a bit since being included in here.

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: I had not tested it. It does not work, but not because of the message string: https://github.com/rancher/gke-operator/pull/17

pkg/api/norman/customization/cluster/validator.go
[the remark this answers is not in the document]
cmurphy: Changed this to look like the EKS validator, it should now activate on either POST or PUT but only if the field is actually present in the config.

pkg/controllers/management/gke/gke_cluster_handler.go
rmweir: Okay, once that other PR is merged it is something that should be checked.

rmweir (approved):
LGTM, made comment about checking error for auth errors. Can either wait for related PR to be merged and test it out before merging this one or just merge this PR and check later.

thedadams (dismissed):

pkg/controllers/management/clusteroperator/utils.go
```diff
@@ -0,0 +1,254 @@
...
+	NoProxy    string `json:"noProxy,omitempty"`
+}
+
+func (e *OperatorController) DeployOperator(operator, operatorTemplate, shortName string) error {
+	template, err := e.TemplateCache.Get(namespace.GlobalNamespace, operatorTemplate)
+	if err != nil {
+		return err
+	}
+
+	latestTemplateVersion, err := e.CatalogManager.LatestAvailableTemplateVersion(template, "local")
```
cbron, line 66: We can use `localCluster` const here

pkg/controllers/management/gke/gke_cluster_handler.go
cbron: That other PR is ready to merge, does that mean we can test this now ?

cbron (commented):
minor things, otherwise lgtm, but need to wait for refresher PR

pkg/controllers/management/clusteroperator/utils.go
cmurphy: done

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: Sorry, I should have been more clear. I *had* not tested it but I have now. With the other pull request, it works - there is a string `403` in the error message that can be matched against.

cbron (dismissed):

cbron:
Needs to be rebased

cbron (approved):

cmurphy:
Backport: https://github.com/rancher/rancher/pull/31907

cbron (approved):

thedadams (approved):

Issue #22093, opened 2019-08-09
Annotations on Rancher defined Pod Security Policies are not applied on Kubernetes cluster PSP


gillarda:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create a new PSP via Global > Security > Pod Security Policies > Add Policy.
2. Add annotations to PSP, for example : 
`"seccomp.security.alpha.kubernetes.io/defaultProfileName": "runtime/default"`
3. Apply the new PSP to a project

**Expected Result:**

User specified annotations should be set on PSP on Kubernetes clusters, in order to be enforced.

**Actual Results:**

PSP on Kubernetes clusters lack the annotations.

**Other details that may be helpful:**

**Environment information**
- Rancher version : 2.2.6
- Installation option (single install/HA): HA

**Cluster information**
- Kubernetes version (use `kubectl version`):
```
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:36:19Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

**PSP state in Rancher API**
```
{
  "allowPrivilegeEscalation": false,
  "allowedCapabilities": null,
  "allowedFlexVolumes": null,
  "allowedHostPaths": null,
  "allowedProcMountTypes": null,
  "allowedUnsafeSysctls": null,
  "annotations": {
    "lifecycle.cattle.io/create.cluster-pspt-sync_c-jrxhd": "true",
    "lifecycle.cattle.io/create.cluster-pspt-sync_local": "true",
    "lifecycle.cattle.io/create.mgmt-pspt-lfc-handler": "true",
    "seccomp.security.alpha.kubernetes.io/allowedProfileNames": "runtime/default",
    "seccomp.security.alpha.kubernetes.io/defaultProfileName": "runtime/default"
  },
  "baseType": "podSecurityPolicyTemplate",
  "created": "2019-08-09T08:28:46Z",
  "createdTS": 1565339326000,
  "creatorId": "user-kj7gd",
  "defaultAddCapabilities": null,
  "defaultAllowPrivilegeEscalation": false,
  "description": null,
  "forbiddenSysctls": null,
  "id": "restricted-enhanced",
  "labels": {
[...]

PR #31651, opened 2021-03-10
[v2.5] Added downstream k8s cluster psp annotations from Rancher psp template

rawmind0:
Issue https://github.com/rancher/rancher/issues/22093

This PR add the feature to apply annotations defined at Rancher psp template to applied downstream cluster psp. Just annotations that its name doesn't contain `.cattle.io/` will be applied.

### DIFF
--- a/pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
+++ b/pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
@@ -3,6 +3,7 @@ package podsecuritypolicy
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
```diff
@@ -131,6 +132,13 @@ func (l *lifecycle) createPolicy(obj *v3.PodSecurityPolicyTemplateProjectBinding
 	objectMeta.Annotations[podSecurityPolicyTemplateParentAnnotation] = template.Name
 	objectMeta.Annotations[podSecurityPolicyTemplateVersionAnnotation] = template.ResourceVersion
 
+	// Setting annotations that doesn't contains podSecurityPolicyTemplateFilterAnnotation
+	for k, v := range template.Annotations {
```
StrongMonkey, line 136: My two cents would be that we don't override any existing annotation that were set in the code. So:
 ```suggestion
	for k, v := range template.Annotations {
	  if _, ok := objectMeta.Annotations[k]; ok {
	    continue
	  }
	  objectMeta.Annotations[k] = v
```

rawmind0: Good point @StrongMonkey, that's why just annotations that doesn't contains `.cattle.io/` are copied. Just "user" annotations should be overwritten with the value coming from psp template. 

Also, we don't want to copy annotations prefixed with`clusterscoped.controller.cattle.io/cluster-pspt-sync_` nor  `lifecycle.cattle.io/create.cluster-pspt-sync_` as they are maintained with the template `cleanPSPT` function, but with your proposal they would be copied.

StrongMonkey (approved):

Issue #31404, opened 2021-02-20
[2.4.x 2.5.x] After the certificate expires, the certificate cannot be automatically updated


kingsd041:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
1. Launch a Rancher server using docker
```
docker run -itd -p 80:80 -p 443:443 --restart=unless-stopped --privileged rancher/rancher:v2.5.5
```
2. Enter the rancher container and check the expiration time of Rancher and K3s certificates
```
root@34f342ec8475:/var/lib/rancher# date
Sat Feb 20 09:00:33 UTC 2021

root@34f342ec8475:/var/lib/rancher# for i in `ls /var/lib/rancher/k3s/server/tls/*.crt`; do echo $i; openssl x509 -enddate -noout -in $i; done
/var/lib/rancher/k3s/server/tls/client-admin.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-auth-proxy.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-ca.crt
notAfter=Feb 18 08:44:14 2031 GMT
/var/lib/rancher/k3s/server/tls/client-cloud-controller.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-controller.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-k3s-controller.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-kube-apiserver.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-kube-proxy.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/client-scheduler.crt
notAfter=Feb 20 08:44:14 2022 GMT
/var/lib/rancher/k3s/server/tls/request-header-ca.crt
notAfter=Feb 18 08:44:14 2031 GMT
/var/lib/rancher/k3s/server/tls/server-ca.crt
notAfter=Feb 18 08:44:14 2031 GMT
/var/lib/rancher/k3s/server/tls/serving-kube-apiserver.crt
notAfter=Feb 20 08:44:14 2022 GMT


[...]

PR #31654, opened 2021-03-10
[master]Update built-in k3s version to v1.19.8+k3s1

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/31404

dramich (dismissed):

StrongMonkey:
Changed to pull from rancher/k3s which is the non-self-extracted binary per @brandond's suggestion.

StrongMonkey:
Since rancher/k3s is multi-arch image so this should also work for arm

dramich (approved):

Issue #31604, opened 2021-03-05
Implement ACE for RKE2/K3S clusters provisioned in Rancher


thedadams:
**What kind of request is this (question/bug/enhancement/feature request):**
feature

**Result:**
ACE should be installed on an RKE2/K3S cluster provisioned in Rancher.

Martin-Weiss:
Will this also work for imported RKE2 clusters? Today we deploy and import RKE2 - but we need ACE once available and we would not want to redeploy all the clusters...

rancher-max:
Using https://github.com/rancher/rancher/issues/32388 for rke2. Blocked on trying with k3s due to https://github.com/rancher/rancher/issues/33443

rancher-max:
When using ACE for both rke2 and k3s on v2.6-head commit `e3316c9`, the downloaded kubeconfig is duplicating the direct node IP(s). This gives an error when trying to use it: `error: error loading config file "~/.kube/config": error converting *[]NamedCluster into *map[string]*api.Cluster: duplicate name "maxrke2ace-ip-<redacted-ip>" in list:` 

The downloaded kubeconfig, with some redacted information, looks like:
```
apiVersion: v1
kind: Config
clusters:
- name: "maxrke2ace"
  cluster:
    server: "https:/<rancher_server>/k8s/clusters/c-m-81d7f8c5"
    certificate-authority-data: "<redacted>"
- name: "maxrke2ace-ip-<redacted-ip>"
  cluster:
    server: "https://<redacted-ip>:6443"
    certificate-authority-data: "<redacted>"
- name: "maxrke2ace-ip-<redacted-ip>"
  cluster:
    server: "https://<redacted-ip>:6443"
    certificate-authority-data: "<redacted>"

users:
- name: "maxrke2ace"
  user:
    token: "<redacted-token>"


contexts:
- name: "maxrke2ace"
  context:
    user: "maxrke2ace"
    cluster: "maxrke2ace"
- name: "maxrke2ace-ip-<redacted-ip>"
  context:
    user: "maxrke2ace"
    cluster: "maxrke2ace-ip-<redacted-ip>"
- name: "maxrke2ace-ip-<redacted-ip>"
  context:
    user: "maxrke2ace"
    cluster: "maxrke2ace-ip-<redacted-ip>"

current-context: "maxrke2ace"
```
where `<redacted-ip>` is the same ip.

thedadams:
[...]

PR #31655, opened 2021-03-10
Add ACE support for Rancher provisioned RKE2/K3S clusters

thedadams:
In order to supoort ACE for RKE2/K3S clusters, the v3 Cluster object
needs to track which v3 Nodes are associated to it. NodesFromCAPI is
used for this purpose. Nodes are added/removed from rancher-operator.

Also, RKE uses the label `controlplane` for the control plane role,
while RKE2/K3S uses `control-plane`. A matchExpresion is added so that
the kube-api-auth daemon set schedules a pod on a node with the
`control-plane` role.

Issue:
https://github.com/rancher/rancher/issues/31604


pkg/api/norman/customization/cluster/actions_kubeconfig.go
```diff
@@ -64,6 +64,9 @@ func (a ActionHandler) GenerateKubeconfigActionHandler(actionName string, action
 		}
 
 		cfg, err = kubeconfig.ForClusterTokenBased(&cluster, apiContext.ID, apiContext.Request.Host, tokenKey)
+		if err != nil {
+			return err
+		}
```
rmweir, line 69: An error check is already being done on L73.

rmweir (commented):
One comment, otherwise LGTM.

pkg/api/norman/customization/cluster/actions_kubeconfig.go
thedadams: The `err` variable is being shadowed on line 46, so the value of `err` on line 73 is different.

dramich: Trying to hold that error all the way through is a bad pattern, it takes too long to parse this function to determine where those errors are being set and as Donnie found, we can accidentally shadow it and never catch the error. We should be checking them when we create them.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -150,6 +146,17 @@ func ForClusterTokenBased(cluster *managementv3.Cluster, clusterID, host, token
 			}
 			nodes = append(nodes, clusterNode)
 		}
+	} else {
+		for _, v3Node := range cluster.NodesFromCAPI {
+			if v3Node.ControlPlane {
+				nodes = append(nodes, node{
+					ClusterName: v3Node.RequestedHostname,
```
dramich, line 153: Is the name format being changed on purpose? I would think it should be same as above to keep the kubeconfig consistent.

thedadams: Yes, the `RequestedHostname` already has the cluster name in it. For example, `donnie-test-cluster-nodepool-pool1-d82r8`.

dramich (dismissed):

PR #31656, opened 2021-03-11
Adding changes to windows cluster creation

sowmyav27:
Parameterizing cloud provider for windows so that we can create vxlan and Host GW with and without CP enabled.

### DIFF
--- a/tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
+++ b/tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
@@ -16,6 +16,7 @@ AZURE_TENANT_ID = os.environ.get("AZURE_TENANT_ID")
 worker_count = int(os.environ.get('RANCHER_STRESS_TEST_WORKER_COUNT', 1))
 HOST_NAME = os.environ.get('RANCHER_HOST_NAME', "testcustom")
 engine_install_url = "https://releases.rancher.com/install-docker/20.10.sh"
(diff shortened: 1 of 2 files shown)


tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
```diff
@@ -1139,15 +1146,20 @@ def create_and_validate_custom_host(node_roles, random_cluster_name=False,
...
-                                  version=K8S_VERSION):
+                                  version=K8S_VERSION, 
+                                  cloudprovider=CLOUDPROVIDER_ENABLED):
     client = get_user_client()
     cluster_name = random_name() if random_cluster_name \
         else evaluate_clustername()
 
     if windows:
-        if windows_flannel_backend == "host-gw":
+        if windows_flannel_backend == "host-gw" and cloudprovider:
```
izaac, line 1156: Isn't `cloudprovider` always be boolean `True` here ? `CLOUDPROVIDER_ENABLED` is a string and non-empty will always return True.

izaac (commented):
A comment about the `CLOUDPROVIDER_ENABLED`

tests/validation/tests/v3_api/test_rke_cluster_provisioning.py
sowmyav27: i think you are right. Changing it and testing it set to false. I checked for "CLOUDPROVIDER_ENABLED" = True earlier.

sowmyav27: Works now with the latest changes. Tested it locally with cloudprovider = True

PR #31665, opened 2021-03-11
[release/v2.4] Post Release

ryansann:
Switch back to dev branches (post-release step)


luthermonson (dismissed):

luthermonson (commented):
https://github.com/rancher/rancher/blob/release/v2.4/pkg/settings/setting.go#L182

kinarashah (approved):

PR #31666, opened 2021-03-11
[release/v2.5] Post Release

ryansann:
Switch branches back to dev as a post-release step


luthermonson:
https://github.com/rancher/rancher/blob/release/v2.5/pkg/settings/setting.go#L223 forgot one

kinarashah (approved):

Issue #30924, opened 2021-01-21
helm uninstall rancher leaves fleet, rancher operator, rancher webhook and helm operation remnants 


izaac:
**What kind of request is this:**
enhancement

**Steps to reproduce:**

- Install Rancher HA RKE with helm
- Once installed perform an uninstall with helm

**Result:**

```
k  get all -n cattle-system
NAME                                  READY   STATUS      RESTARTS   AGE
pod/helm-operation-4k9fg              0/2     Completed   0          26m
pod/helm-operation-kbpsb              0/2     Completed   0          27m
pod/helm-operation-rp2jq              0/2     Completed   0          26m
pod/helm-operation-x65dn              0/2     Completed   0          26m
pod/helm-operation-zghhz              0/2     Completed   0          27m
pod/rancher-webhook-b5b7b76c4-ftbqs   1/1     Running     0          27m
NAME                      TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)   AGE
service/rancher-webhook   ClusterIP   10.43.97.95   <none>        443/TCP   27m
NAME                              READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/rancher-webhook   1/1     1            1           27m
NAME                                        DESIRED   CURRENT   READY   AGE
replicaset.apps/rancher-webhook-b5b7b76c4   1         1         1       27m



k get all -n fleet-system
NAME                                   READY   STATUS    RESTARTS   AGE
pod/fleet-agent-76fc77855-dq7jq        1/1     Running   0          25m
pod/fleet-controller-ccc95b8cd-vs8mb   1/1     Running   0          26m
pod/gitjob-5997858b9c-2l4fr            1/1     Running   0          26m
NAME             TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)   AGE
service/gitjob   ClusterIP   10.43.90.64   <none>        80/TCP    26m
NAME                               READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/fleet-agent        1/1     1            1           26m
deployment.apps/fleet-controller   1/1     1            1           26m
deployment.apps/gitjob             1/1     1            1           26m
NAME                                         DESIRED   CURRENT   READY   AGE
replicaset.apps/fleet-agent-76fc77855        1         1         1       25m
replicaset.apps/fleet-agent-7bbdcffc68       0         0         0       26m
replicaset.apps/fleet-controller-ccc95b8cd   1         1         1       26m
[...]

Issue #10, opened 2014-11-27
Host stats are reported until after the first container on the host is deployed


ibuildthecloud:
We detect the node IP on first container start.  We need to detect the host IP on register.  Since we don't know the node IP the browser can't connect to it.

PR #31670, opened 2021-03-12
add a Helm post-delete-hook for Rancher

jiaqiluo:
Add a post-delete-hook for cleaning up the apps left after uninstalling Rancher.
As of now, it will delete 
- the `helm-operation-*` pods
- the following apps from the cluster: 
```
- fleet
- fleet-agent
- rancher-operator
- rancher-webhook
```

see this comment for the behavior when the job fails: https://github.com/rancher/rancher/pull/31670#issuecomment-800437206

issue: https://github.com/rancher/rancher/issues/30924

### DIFF
--- a/chart/scripts/post-delete-hook.sh
+++ b/chart/scripts/post-delete-hook.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+set -e
+
+namespaces="${NAMESPACES}"
+timeout="${TIMEOUT}"
+ignoreTimeoutError="${IGNORETIMEOUTERROR}"
+
+if [[ -z ${namespaces} ]]; then
+  echo "No namespace is provided."
+  exit 1
+fi
+
+if [[ -z ${timeout} ]]; then
+  echo "No timeout value is provided."
+  exit 1
+fi
+
+if [[ -z ${ignoreTimeoutError} ]]; then
+  echo "No ignoreTimeoutError value is provided."
+  exit 1
+fi
+
+succeeded=()
+failed=()
+
+get_pod_count() {
+  kubectl get pods --selector app="${1}" -n "${2}" -o json | jq '.items | length'
+}
+
+echo "Uninstalling Rancher resources in the following namespaces: ${namespaces}"
+
+for namespace in ${namespaces}; do
+  for app in $(helm list -n "${namespace}" -q); do
+    if [[ ${app} =~ .crd$ ]]; then
+      echo "--- Skip the app [${app}] in the namespace [${namespace}]"
+      continue
+    fi
+    echo "--- Deleting the app [${app}] in the namespace [${namespace}]"
+    if [[ ! $(helm uninstall "${app}" -n "${namespace}") ]]; then
+      failed=("${failed[@]}" "${app}")
+      continue
+    fi
+
+    t=0
+    while true; do
+      if [[ $(get_pod_count "${app}" "${namespace}") -eq 0 ]]; then
+        echo "successfully uninstalled [${app}] in the namespace [${namespace}]"
+        succeeded=("${succeeded[@]}" "${app}")
+        break
+      fi
+      if [[ ${t} -ge ${timeout} ]]; then
+        echo "timeout uninstalling [${app}] in the namespace [${namespace}]"
+        failed=("${failed[@]}" "${app}")
+        break
+      fi
+      # by default, wait 120 seconds in total for an app to be uninstalled
+      echo "waiting 5 seconds for pods of [${app}] to be terminated ..."
+      sleep 5
+      t=$((t + 5))
+    done
+  done
+
+  # delete the helm operator pods
+  for pod in $(kubectl get pods -n "${namespace}" -o name); do
+    if [[ ${pod} =~ ^pod\/helm-operation-* ]]; then
+      echo "--- Deleting the pod [${pod}] in the namespace [${namespace}]"
+      kubectl delete "${pod}" -n "${namespace}"
+    fi
+  done
+done
+
+echo "------ Summary ------"
+if [[ ${#succeeded[@]} -ne 0 ]]; then
+  echo "Succeeded to uninstall the following apps:" "${succeeded[@]}"
+fi
+
+if [[ ${#failed[@]} -ne 0 ]]; then
+  echo "Failed to uninstall the following apps:" "${failed[@]}"
+  if [[ "${ignoreTimeoutError}" == "false" ]]; then
+    exit 2
+  fi
+else
+  echo "Cleanup finished successfully."
+fi
--- a/chart/templates/post-delete-hook-job.yaml
+++ b/chart/templates/post-delete-hook-job.yaml
@@ -1,0 +1,44 @@
+{{- if .Values.postDelete.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "rancher.fullname" . }}-post-delete
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "rancher.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "3"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  backoffLimit: 3
+  template:
+    metadata:
+      name: {{ template "rancher.fullname" . }}-post-delete
+      labels: {{ include "rancher.labels" . | nindent 8 }}
+    spec:
+      serviceAccountName: {{ template "rancher.fullname" . }}-post-delete
+      restartPolicy: OnFailure
+      containers:
+        - name: {{ template "rancher.name" . }}-post-delete
+          image: "{{ .Values.systemDefaultRegistry }}{{ .Values.postDelete.image.repository }}:{{ .Values.postDelete.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            runAsUser: 0
+          command:
+            - /scripts/post-delete-hook.sh
+          volumeMounts:
+            - mountPath: /scripts
+              name: config-volume
+          env:
+            - name: NAMESPACES
+              value: {{ .Values.postDelete.namespaceList | join " " | quote }}
+            - name: TIMEOUT
+              value: {{ .Values.postDelete.timeout | quote }}
+            - name: IGNORETIMEOUTERROR
+              value: {{ .Values.postDelete.ignoreTimeoutError | quote }}
+      volumes:
+        - name: config-volume
+          configMap:
+            name: {{ template "rancher.fullname" . }}-post-delete
+            defaultMode: 0777
+{{- end }}
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -137,3 +137,18 @@ customLogos:
   # storageClass: "-"
   accessMode: ReadWriteOnce
   size: 1Gi
+
+# Rancher post-delete hook
+postDelete:
+  enabled: true
+  image:
+    repository: %POST_DELETE_IMAGE_NAME%
+    tag: %POST_DELETE_IMAGE_TAG%
+  namespaceList:
+    - fleet-system
+    - cattle-system
+    - rancher-operator-system
+  # Number of seconds to wait for an app to be uninstalled
+  timeout: 120
+  # by default, the job will fail if it fail to uninstall any of the apps
+  ignoreTimeoutError: false
--- a/scripts/chart/build
+++ b/scripts/chart/build
@@ -14,3 +14,18 @@ cp -rf ${1} build/chart/rancher
 
 sed -i -e "s/%VERSION%/${CHART_VERSION}/g" build/chart/rancher/Chart.yaml
 sed -i -e "s/%APP_VERSION%/${VERSION}/g" build/chart/rancher/Chart.yaml
+
+# get the value of shell-image, such as rancher/shell:v0.1.6, from the file pkg/settings/setting.go
+post_delete_base=$(grep -i shell-image pkg/settings/setting.go | cut -d "," -f 2 | sed -e 's/"//g' | sed -e 's/)//g' | sed -e 's/ //g') || ""
+post_delete_image_name=$(echo "${post_delete_base}" | cut -d ":" -f 1) || ""
+post_delete_image_tag=$(echo "${post_delete_base}" | cut -d ":" -f 2) || ""
+if [[ ! ${post_delete_image_name} =~ ^rancher\/.+ ]]; then
+  echo "The image name [$post_delete_image_name] is invalid. Its prefix should be rancher/"
+  exit 1
+fi
+if [[ ! ${post_delete_image_tag} =~ ^v.+ ]]; then
+  echo "The image tag [$post_delete_image_tag] is invalid. It should start with the letter v"
+  exit 1
+fi
+sed -i -e "s@%POST_DELETE_IMAGE_NAME%@${post_delete_image_name}@g" build/chart/rancher/values.yaml
+sed -i -e "s/%POST_DELETE_IMAGE_TAG%/${post_delete_image_tag}/g" build/chart/rancher/values.yaml
--- a/chart/templates/post-delete-hook-cluster-role-binding.yaml
+++ b/chart/templates/post-delete-hook-cluster-role-binding.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.postDelete.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "rancher.fullname" . }}-post-delete
+  labels: {{ include "rancher.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "rancher.fullname" . }}-post-delete
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "rancher.fullname" . }}-post-delete
+    namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/chart/templates/post-delete-hook-cluster-role.yaml
+++ b/chart/templates/post-delete-hook-cluster-role.yaml
@@ -1,0 +1,27 @@
+{{- if .Values.postDelete.enabled }}
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ template "rancher.fullname" . }}-post-delete
+  labels: {{ include "rancher.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+rules:
+  - apiGroups: [ "extensions","apps" ]
+    resources: [ "deployments" ]
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "rbac.authorization.k8s.io" ]
+    resources: [ "clusterroles", "clusterrolebindings", "roles", "rolebindings" ]
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "" ]
+    resources: [ "serviceaccounts", "pods", "secrets", "services", "configmaps" ]
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "networking.k8s.io" ]
+    resources: [ "networkpolicies" ]
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "admissionregistration.k8s.io" ]
+    resources: [ "validatingwebhookconfigurations" ]
+    verbs: [ "get", "list", "delete" ]
+{{- end }}
--- a/chart/templates/post-delete-hook-config-map.yaml
+++ b/chart/templates/post-delete-hook-config-map.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.postDelete.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "rancher.fullname" . }}-post-delete
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "rancher.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+data:
+  post-delete-hook.sh: |-
+{{ $.Files.Get "scripts/post-delete-hook.sh" | indent 4 }}
+{{- end }}
--- a/chart/templates/post-delete-hook-service-account.yaml
+++ b/chart/templates/post-delete-hook-service-account.yaml
@@ -1,0 +1,12 @@
+{{- if .Values.postDelete.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "rancher.fullname" . }}-post-delete
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "rancher.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+{{- end }}


chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,34 @@
...
+  app_list=$(helm list -n "$namespace" -q)
+  for app in $app_list; do
+    if [[ $app =~ .crd$ ]]; then
+      echo "--- skip" $app
+      continue
+    fi
+    echo "--- deleting the app [$app] (namespace: $namespace)"
+    helm uninstall "$app" -n "$namespace"
+    c=0
+    # wait 120 sections in total for an uninstallation to be finished
```
rawmind0, line 24: typo here

```
-    # wait 120 sections in total for an uninstallation to be finished
+    # wait 120 seconds in total for an uninstallation to be finished
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,34 @@
...
+    while [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]] && [[ $c -lt 20 ]]; do
+      echo "waiting for pods of" "$app" "to terminate..."
+      sleep 6
+      c=$((c+1))
+    done
+    echo "uninstalled" "[$app]" "in the namespace" "[$namespace]"
```
rawmind0, line 30: The scripts is waiting up to 120 seconds for every single app being uninstalled. I think that the exit of the `while` loop should be controlled to check if the app has been properly uninstalled or has got timeout

```
-    echo "uninstalled" "[$app]" "in the namespace" "[$namespace]"
+    if [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]]; then
+      echo "timeout uninstalling" "[$app]" "in the namespace" "[$namespace]"
+      continue
+    fi
+    echo "uninstalled" "[$app]" "in the namespace" "[$namespace]"
```

chart/templates/post-delete-hook-configMap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "rancher.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": hook-succeeded
+data:
+  post-delete-hook.sh: |-
+{{ $.Files.Get "scripts/post-delete-hook.sh" | indent 4 }}
```
rawmind0, line 14: Shouldn't this template rendering be conditioned by `Values.postDelete.enabled` value??

```
+{{- if .Values.postDelete.enabled }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Values.postDelete.name}}-script
  namespace: {{ .Release.Namespace }}
  labels:
{{ include "rancher.labels" . | indent 4 }}
  annotations:
    "helm.sh/hook": post-delete
    "helm.sh/hook-weight": "1"
    "helm.sh/hook-delete-policy": hook-succeeded
data:
  post-delete-hook.sh: |-
{{ $.Files.Get "scripts/post-delete-hook.sh" | indent 4 }}
+{{ end }}
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,34 @@
...
+  for app in $app_list; do
+    if [[ $app =~ .crd$ ]]; then
+      echo "--- skip" $app
+      continue
+    fi
+    echo "--- deleting the app [$app] (namespace: $namespace)"
+    helm uninstall "$app" -n "$namespace"
+    c=0
+    # wait 120 sections in total for an uninstallation to be finished
+    while [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]] && [[ $c -lt 20 ]]; do
```
rawmind0, line 25: Have you considered to set retries configurable from values?? I mean, here is hardcoded 20 retries every 6 seconds, 120 seconds. Would make sense to add a a new values var `.Values.postDelete.retries = 20` to make it configurable??

```
-    while [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]] && [[ $c -lt 20 ]]; do
+    while [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]] && [[ $c -lt $retries ]]; do
```

chart/templates/post-delete-hook-job.yaml
```diff
@@ -0,0 +1,42 @@
...
+          securityContext:
+            runAsUser: 0
+          command:
+            - /script/post-delete-hook.sh
+          volumeMounts:
+            - mountPath: /script
+              name: config-volume
+          env:
+            - name: namespaces
+              value: {{ .Values.postDelete.namespaceList | join " " }}
```
rawmind0, line 36: Adding retries values if make sense

```
+            - name: retries
+              value: {{ .Values.postDelete.retries }}
```

chart/values.yaml
```diff
@@ -136,3 +136,15 @@ customLogos:
...
+postDelete:
+  enabled: true
+  name: rancher-post-delete
+  image:
+    repository: rancher/shell
+    tag: v0.1.6
+  namespaceList:
+    - fleet-system
+    - cattle-system
+    - rancher-operator-system
```
rawmind0, line 150: Adding retries values if make sense

```
+  # Number of retries to unsinstall apps. Post delete hook will retry every 6 seconds, 120 seconds total by default
+  retries: 20
```

rawmind0 (commented):
@jiaqiluo , my 2 cents on this PR. Please, take a look to review comments.

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed.

chart/templates/post-delete-hook-configMap.yaml
jiaqiluo: Yes, you are right. The condition was added.

jiaqiluo:
@rawmind0  Thank you for your feedback! I have addressed all of them.

Instead of adding a variable for the count of retries, I added a variable for the total waiting time (in seconds). After the script runs the `helm uninstall` command, it repeats to check if all pods for the app are deleted every 5s until it hits the timeout.

Another thing to mention is that, in the current design, the bash script will always exit with code 0, so the job will always succeed and then delete itself. This behavior is controlled by the following annotation in the job
```
"helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```

Do you think the script should exit with a non-zero value if it failed to uninstall any app? If the script fails, should we restart the job and keep it in the cluster if it still fail eventually?

rawmind0:
> Instead of adding a variable for the count of retries, I added a variable for the total waiting time (in seconds). After the script runs the helm uninstall command, it repeats to check if all pods for the app are deleted every 5s until it hits the timeout.

No problem, it's more clear in the way you did it.  :)

> Do you think the script should exit with a non-zero value if it failed to uninstall any app? If the script fails, should we restart the job and keep it in the cluster if it still fail eventually?

I see. I think that the script should exit with a non-zero value. The uninstallation process should be run "manually" again if fails. If unsure, may be making it configurable would be a good option. It'd be just add a new bool var `postDelete.ignoreTimeoutError`, then exit non zero if false or continue if true at the script.

jiaqiluo:
@rawmind0 Thank you for your feedback!

Here is what I am planning to do:
- add a new bool var `postDelete.ignoreTimeoutError`. When a timeout happens, the script will exit non-zero if `postDelete.ignoreTimeoutError` is false, or continue if it is true 
- at the Job level, if `postDelete.ignoreTimeoutError` is false and the pod exits at the failed/error state, the `job` (where the pod comes from) and the `configMap` (where the script comes from) will be kept for future investigation. 

@cbron What do you think? Does that sound good to you?

chart/values.yaml
```diff
@@ -136,3 +136,17 @@ customLogos:
...
   accessMode: ReadWriteOnce
   size: 1Gi
+
+# Rancher post-delete hook
+postDelete:
+  enabled: true
+  name: rancher-post-delete
+  image:
+    repository: rancher/shell
+    tag: v0.1.6
```
cbron, line 146: This will not get pulled into images.txt if the primary shell gets updated. That is from here: 
https://github.com/rancher/rancher/blob/master/pkg/settings/setting.go#L94
https://github.com/rancher/rancher/blob/master/pkg/image/resolve.go#L248

Can we use the agent ? I can't think of a way to get the shell into the chart consistently. Unless we hardcode it into the requirements struct there but that feels bad.

chart/templates/post-delete-hook-ServiceAccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.postDelete.enabled }}
```
cbron, line 1: can you lowercase the filename, uppercase looks weird.

chart/templates/post-delete-hook-ServiceAccount.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.postDelete.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "rancher.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": hook-succeeded
```
cbron, line 12: Interesting, I hadn't thought of using the delete policy on random objects like this. Should work.

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+set -e
+
+echo "Preparing for uninstalling apps left by Rancher in the following namespaces:" "$namespaces"
```
cbron, line 5:
```suggestion
echo "Uninstalling Rancher resources in the following namespaces:" "$namespaces"
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,38 @@
...
+for namespace in $namespaces; do
```
cbron, line 6: Does namespaces actually get inherited from the env here ? I'm surprised its in the shells context.

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,38 @@
...
+    fi
+    echo "--- deleting the app [$app] in the namespace [$namespace]"
+    helm uninstall "$app" -n "$namespace"
+    t=0
+    # by default, wait 120 seconds in total for an app to be uninstalled
+    while [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]] && [[ $t -lt "$timeout" ]]; do
+      echo "waiting for pods of [$app] to be terminated ..."
+      sleep 5
+      t=$((t + 5))
+    done
```
cbron, line 29: This code looks familiar :)

cbron (commented):
Yes but I would track which of all apps failed and list them below and exit 1, not just exit 1 as soon as one is hit. Alternative may be to use k8s job spec to delete job either way after 1 hour, that was in beta and non-working by default last I checked. 
Also what is value flag default, false ? I'd say false. 
They can't really re-run the job manually, they would only use it for error logs to see which job failed.

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,38 @@
...
+    if [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]]; then
+      echo "timeout uninstalling [$app] in the namespace [$namespace]"
+    else
+      echo "uninstalled [$app] in the namespace [$namespace]"
+    fi
+  done
+done
+
+echo "The job is done."
```
cbron, line 38: This should list failed jobs if any, or:
```suggestion
echo "Cleanup finished successfully."
```

chart/templates/post-delete-hook-ServiceAccount.yaml
jiaqiluo: Existing files in the directory (/chart/templates) also use CamelCase, so I thought it is the pattern I should follow.  

here is the list of existing ones: 
```
clusterRoleBinding.yaml
issuer-letsEncrypt.yaml
serviceAccount.yaml
```
Should I correct them as well?

cbron: Ok if its a pattern just follow those.

chart/values.yaml
cbron: Actually now that I think about it, I'm not sure the agent image exists on the local cluster in a default (non-airgap/public) setup, do you know ? It would suck to have to pull an image to start a post-delete hook. Its likely that the shell image would have been pulled into local, and if not its 1/4th the size of the agent. Is there anything else we can guarantee is on the local cluster ?

jiaqiluo: This image `rancher/shell` should already be in the local cluster because the `helm-operation` pod uses it to install Helm charts in the local cluster. 
```
> k -n cattle-system describe pod helm-operation-jp4m2 | grep -i image
    Image:         rancher/shell:v0.1.6
    Image ID:      docker.io/rancher/shell@sha256:bc884e3b37a737a7e411766c42705fbe477d6219c72ec781efd8438d8fb267d0
    Image:         rancher/shell:v0.1.6
    Image ID:      docker.io/rancher/shell@sha256:bc884e3b37a737a7e411766c42705fbe477d6219c72ec781efd8438d8fb267d0
  Normal  Pulled     6m58s  kubelet            Container image "rancher/shell:v0.1.6" already present on machine
  Normal  Pulled     6m58s  kubelet            Container image "rancher/shell:v0.1.6" already present on machine
```

Still, the problem is that how we can get the image dynamically instead of hardcoded in the value.yaml file.

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed.

chart/scripts/post-delete-hook.sh
jiaqiluo: yes, the script gets the environment variable from the container.

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed.

jiaqiluo:
@rawmind0 @cbron 
With setting `job.sepc.backoffLimit = 3` and `job.spec.template.spec.restartPolicy = Never`, the job will not restart failed pod, instead, it creates a new pod up to 3 times. If all three pods fail, the job will fail and keep all existing pods for users to check logs. 

Here are the results of a failed run:
(I simulate a failure by adding the `--dry-run` flag to the `helm uninstall` command and setting `timeout = 9` ) 

- helm fails to run the post-delete job (but Rancher itself is uninstalled successfully)
```
> helm uninstall rancher -n cattle-system
Error: uninstallation completed with 1 error(s): job failed: BackoffLimitExceeded
```

- the job and its pods are left in the cluster 
```
> k get job -n cattle-system
NAME                  COMPLETIONS   DURATION   AGE
rancher-post-delete   0/1           4m7s       4m7s

> k get pods -n cattle-system
NAME                              READY   STATUS    RESTARTS   AGE
rancher-post-delete-phrbc         0/1     Error     0          17m
rancher-post-delete-v9plr         0/1     Error     0          18m
rancher-post-delete-xzvqg         0/1     Error     0          17m
```
- we can check the logs for failures
```
> k -n cattle-system logs rancher-post-delete-xzvqg
Uninstalling Rancher resources in the following namespaces: fleet-system cattle-system rancher-operator-system
--- Deleting the app [fleet] in the namespace [fleet-system]
release "fleet" uninstalled
successfully uninstalled [fleet] in the namespace [fleet-system]
--- Skip the app [fleet-crd] in the namespace [fleet-system]
--- Deleting the app [rancher-webhook] in the namespace [cattle-system]
release "rancher-webhook" uninstalled
waiting 5 seconds for pods of [rancher-webhook] to be terminated ...
waiting 5 seconds for pods of [rancher-webhook] to be terminated ...
timeout uninstalling [rancher-webhook] in the namespace [cattle-system]
--- Deleting the app [rancher-operator] in the namespace [rancher-operator-system]
release "rancher-operator" uninstalled
waiting 5 seconds for pods of [rancher-operator] to be terminated ...
waiting 5 seconds for pods of [rancher-operator] to be terminated ...
timeout uninstalling [rancher-operator] in the namespace [rancher-operator-system]
--- Skip the app [rancher-operator-crd] in the namespace [rancher-operator-system]
------ Summary ------
Succeeded to uninstall the following apps: fleet
Failed to uninstall the following apps: rancher-webhook rancher-operator
```

chart/values.yaml
jiaqiluo: > This will not get pulled into images.txt if the primary shell gets updated. That is from here:
> [https://github.com/rancher/rancher/blob/master/pkg/settings/setting.go#L94](https://github.com/rancher/rancher/blob/master/pkg/settings/setting.go?rgh-link-date=2021-03-16T02%3A36%3A58Z#L94)
> [https://github.com/rancher/rancher/blob/master/pkg/image/resolve.go#L248](https://github.com/rancher/rancher/blob/master/pkg/image/resolve.go?rgh-link-date=2021-03-16T02%3A36%3A58Z#L248)
> 
> Can we use the agent? I can't think of a way to get the shell into the chart consistently. Unless we hardcode it into the requirements struct there but that feels bad.


I have 2 possible solutions so far:
1. add some logic to the file [scripts/chart/build](https://github.com/rancher/rancher/blob/master/scripts/chart/build) to get the image and tag from the following line in [pkg/settings/setting.go](https://github.com/rancher/rancher/blob/2dca0c67b48331c23c05c025208bacfc258d45bd/pkg/settings/setting.go#L94)
```
	ShellImage                        = NewSetting("shell-image", "rancher/shell:v0.1.6")
```

2. add a comment to the above line in `pkg/settings/setting.go`:  we need to update the `postDelete.image` section in the file `/chart/values.yaml` for any change on this line.

cbron: 1. could work but it needs to load the entire rancher codebase for it to work, so that would take good amount of time.
2. If someone forgets then airgap won't work for this, much safer to add to requirements struct and also a comment there. That seems like best solution for now.

pkg/settings/setting.go
```diff
@@ -92,6 +92,7 @@ var (
 	PartnerChartDefaultBranch         = NewSetting("partner-chart-default-branch", "main")
 	FleetDefaultWorkspaceName         = NewSetting("fleet-default-workspace-name", "fleet-default") // fleetWorkspaceName to assign to clusters with none
 	ShellImage                        = NewSetting("shell-image", "rancher/shell:v0.1.6")
+	HelmPostDeletionImage             = NewSetting("helm-post-deletion-image", "rancher/shell:v0.1.6") // need to update the value in chart/values.yaml for any change in this line
```
jiaqiluo, line 95: @cbron this is to address your [comment](https://github.com/rancher/rancher/pull/31670/files#r595497658):
> 1. could work but it needs to load the entire rancher codebase for it to work, so that would take good amount of time.
> 2. If someone forgets then airgap won't work for this, much safer to add to requirements struct and also a comment there. That seems like best solution for now.

cbron:
I think a single pod with `restartPolicy: OnFailure` is probably best. That will try a bunch of times but only use one pod and if last try fails they still get logs.

pkg/settings/setting.go
cbron: Need to run gofmt on this line. Also I think we should add here and say to update this line: https://github.com/rancher/rancher/blob/master/pkg/image/resolve.go#L248
That resolve.go line needs to always be in sync with values.yaml for this hook, that is our failsafe if someone changes this setting and not the values.yaml.

jiaqiluo:
> I think a single pod with `restartPolicy: OnFailure` is probably best. That will try a bunch of times but only use one pod and if last try fails they still get logs.

@cbron The problem of setting the pod with `restartPolicy: OnFailure` is that the pod will be deleted when the job hits its `backoffLimit` and fails, which gives us no chance to check its logs. 

Quote from https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy

> Note: If your job has restartPolicy = "OnFailure", keep in mind that your container running the Job will be terminated once the job backoff limit has been reached. This can make debugging the Job's executable more difficult. We suggest setting restartPolicy = "Never" when debugging the Job or using a logging system to ensure output from failed Jobs is not lost inadvertently.

jiaqiluo:
Based on the conversation with @cbron, we decided to set the pod with `restartPolicy: OnFailure`.

rawmind0 (dismissed):
LGTM

scripts/chart/build
```diff
@@ -14,3 +14,9 @@ cp -rf ${1} build/chart/rancher
 
 sed -i -e "s/%VERSION%/${CHART_VERSION}/g" build/chart/rancher/Chart.yaml
 sed -i -e "s/%APP_VERSION%/${VERSION}/g" build/chart/rancher/Chart.yaml
+
+post_delete_base=$(grep -i shell-image pkg/settings/setting.go | cut -d "," -f 2 | sed -e 's/"//g' | sed -e 's/)//g' | sed -e 's/ //g')
+post_delete_image_name=$(echo ${post_delete_base} | cut -d ":" -f 1)
+post_delete_image_tag=$(echo ${post_delete_base} | cut -d ":" -f 2)
```
cbron, line 20: Let's add some checks here to ensure the name and tag are what we expect. We _could_ hardcode the image name but thats probably not necessary as it could change down the line. But checking that tag starts with v or that "shell" is in image name, something like that.

jiaqiluo: The following 2 changes are made to address your comments:
1. For the line 18-20, those commands will not exit with code 0, they return an empty string and exit code 0 if fail.
2. The checks for the image name and tag are added at lines 21-28.

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+set -e
+
+succeeded=()
+failed=()
+echo "Uninstalling Rancher resources in the following namespaces:" "$namespaces"
```
aiyengar2, line 7:
```suggestion
echo "Uninstalling Rancher resources in the following namespaces: ${namespaces}"
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+set -e
+
+succeeded=()
```
aiyengar2, line 5: Need to add a check to ensure that the expected env vars exists. Example:

```suggestion
if [ -z "${namespaces}" ]; then
  echo "No namespaces provided."
  exit 0
fi

succeeded=()
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
...
+failed=()
+echo "Uninstalling Rancher resources in the following namespaces:" "$namespaces"
+for namespace in $namespaces; do
+  app_list=$(helm list -n "$namespace" -q)
+  for app in $app_list; do
```
aiyengar2, line 10: nit:
```suggestion
  for app in $(helm list -n "${namespace}" -q); do
```

chart/scripts/post-delete-hook.sh
aiyengar2: Can you change all env var references to `${ENVVAR}` instead of `$ENVVAR`? 

The bracket notation is less ambiguous and imo is generally better practice in bash scripts since you don't encounter issues like `${foo}bar` v.s. `$foobar`.

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
...
+    if [[ $app =~ .crd$ ]]; then
+      echo "--- Skip the app [$app] in the namespace [$namespace]"
+      continue
+    fi
+    echo "--- Deleting the app [$app] in the namespace [$namespace]"
+    if [[ ! $(helm uninstall "$app" -n "$namespace") ]]; then
```
aiyengar2, line 16: Does the following work instead?

```suggestion
    if ! helm uninstall "${app}" -n "${namespace}"; then
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
...
+      t=$((t + 5))
+    done
+    if [[ "$(kubectl get pods --selector app="$app" -n "$namespace" -o json | jq '.items | length')" != "0" ]]; then
+      echo "timeout uninstalling [$app] in the namespace [$namespace]"
+      failed=("${failed[@]}" "$app")
+    else
+      echo "successfully uninstalled [$app] in the namespace [$namespace]"
+      succeeded=("${succeeded[@]}" "$app")
+    fi
+  done
```
aiyengar2, line 34: Can we drop this in favor of just adding the `--timeout` flag on the helm uninstall?

https://helm.sh/docs/helm/helm_uninstall

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
...
+  # delete the helm operator pods
+  pod_list=$(kubectl get pods -n "$namespace" -o name --ignore-not-found)
+  for pod in $pod_list; do
```
aiyengar2, line 37: `--ignore-not-found` is only needed for kubectl delete right? Not for `kubectl get`

```suggestion
  for pod in $(kubectl get pods -n "$namespace" -o name); do
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+if [[ ${#failed[@]} -ne 0 ]]; then
+  echo "Failed to uninstall the following apps:" "${failed[@]}"
+  if [[ "$ignoreTimeoutError" == "false" ]]; then
+    exit 1
+  fi
+else
+  echo "Cleanup finished successfully."
+fi
```
aiyengar2, line 57: Any reason why we want to collect them and then output it at the end as opposed to just printing an error statement and continuing in the for loop itself? Seems like collecting and dumping at the end makes the code more complicated to maintain (because Bash arrays aren't easy to look at 😅 ).

chart/templates/post-delete-hook-ServiceAccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.postDelete.enabled }}
+apiVersion: v1
```
aiyengar2, line 2: File naming shouldn't be capitalized and have dashes. We should probably just stick to dashes for all of these files for style.

chart/templates/post-delete-hook-clusterRole.yaml
```diff
@@ -0,0 +1,28 @@
...
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "" ]
+    resources: [ "serviceaccounts", "pods", "secrets", "services" ]
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "networking.k8s.io" ]
+    resources: [ "networkpolicies" ]
+    verbs: [ "get", "list", "delete" ]
+  - apiGroups: [ "admissionregistration.k8s.io" ]
+    resources: [ "validatingwebhookconfigurations" ]
+    verbs: [ "get", "list", "delete" ]
```
aiyengar2, line 27: nit: to avoid mixing JSON / YAML, I'd personally prefer if this was something like
```yaml
- apiGroups:
  - extensions
  - apps
- resources:
  - deployments
verbs:
  - get
  - list
  - delete
```

If you disagree, feel free to resolve this comment.

chart/templates/post-delete-hook-clusterRole.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.postDelete.enabled }}
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Values.postDelete.name }}
```
aiyengar2, line 5: Why does the user need to provide `.postDelete.name`? Let's use something like `{{ template "rancher.fullname" . }}-post-delete` instead if possible and remove the provided name entirely from the values.yaml.

chart/templates/post-delete-hook-clusterRole.yaml
```diff
@@ -0,0 +1,28 @@
...
+  labels:
+{{ include "rancher.labels" . | indent 4 }}
```
aiyengar2, line 7: nit:
```suggestion
  labels: {{ include "rancher.labels" . | nindent 4 }}
```

chart/templates/post-delete-hook-clusterRole.yaml
aiyengar2: same comment for everywhere else this is used.

chart/templates/post-delete-hook-clusterRole.yaml
aiyengar2: same comment for everywhere else this is used.

chart/templates/post-delete-hook-configMap.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+{{ include "rancher.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+data:
+  post-delete-hook.sh: |-
+{{ $.Files.Get "scripts/post-delete-hook.sh" | indent 4 }}
+{{ end }}
```
aiyengar2, line 16:
```suggestion
{{- end }}
```

aiyengar2: same comment for wherever else the prefix `-` is omitted.

While it's not a big deal, if someone were to add a resource under this it could potentially leave weird newlines in the helm template produced.

chart/templates/post-delete-hook-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+    "helm.sh/hook-weight": "3"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  backoffLimit: 3
+  template:
+    metadata:
+      name: {{ .Values.postDelete.name }}
+      labels:
+        heritage: {{ .Release.Service }}
+        release: {{ .Release.Name }}
```
aiyengar2, line 20:
```suggestion
      labels: {{ include "rancher.labels" . | nindent 8 }}
```

chart/templates/post-delete-hook-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+      containers:
+        - name: {{ .Values.postDelete.name }}
+          image: "{{ .Values.systemDefaultRegistry }}{{ .Values.postDelete.image.repository }}:{{ .Values.postDelete.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            runAsUser: 0
+          command:
+            - /script/post-delete-hook.sh
+          volumeMounts:
+            - mountPath: /script
```
aiyengar2, line 33: In case we add more scripts in the future. Will also need to change the command to `/scripts/post-delete-hook.sh`

```suggestion
            - mountPath: /scripts
```

chart/templates/post-delete-hook-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+              name: config-volume
+          env:
+            - name: namespaces
+              value: {{ .Values.postDelete.namespaceList | join " " | quote }}
+            - name: timeout
+              value: {{ .Values.postDelete.timeout | quote }}
+            - name: ignoreTimeoutError
+              value: {{ .Values.postDelete.ignoreTimeoutError | quote }}
```
aiyengar2, line 41: environment variables that come from outside the bash script should be capitalized, e.g. `NAMESPACES`.

Also, are the `| quote` steps necessary?

chart/values.yaml
```diff
@@ -136,3 +136,19 @@ customLogos:
...
+postDelete:
+  enabled: true
+  name: rancher-post-delete
+  image:
+    repository: %POST_DELETE_IMAGE_NAME%
+    tag: %POST_DELETE_IMAGE_TAG%
+  namespaceList:
+    - fleet-system
+    - cattle-system
+    - rancher-operator-system
```
aiyengar2, line 150: any reason why we expose this instead of hard-coding the namespaceList in the chart deployment? Is there a use case where a suer can install stuff onto non-default namespaces that we want to support?

scripts/chart/build
```diff
@@ -14,3 +14,17 @@ cp -rf ${1} build/chart/rancher
 
 sed -i -e "s/%VERSION%/${CHART_VERSION}/g" build/chart/rancher/Chart.yaml
 sed -i -e "s/%APP_VERSION%/${VERSION}/g" build/chart/rancher/Chart.yaml
+
+post_delete_base=$(grep -i shell-image pkg/settings/setting.go | cut -d "," -f 2 | sed -e 's/"//g' | sed -e 's/)//g' | sed -e 's/ //g') || ""
```
aiyengar2, line 18: I see what you are doing here, but I'm really not in favor of running sed through Go code to infer the base image of this chart. IMO, we should be just hard-coding the base image in the chart instead of relying on build magic.

The Version / AppVersion are different since they are small sed commands, but this one is really hard to parse through (because parsing a Go file is not something bash scripts can / should do). This is also relatively fragile since it forces us to never add something with the word `shell-image` before [this](https://github.com/rancher/rancher/blob/b0d3575cc216c5191590b39f23f2888f4e731a69/pkg/settings/setting.go#L97) in our Go code, which is a weird constraint to have on our code.

Was this a strict requirement for this issue?

aiyengar2 (changes requested):

chart/scripts/post-delete-hook.sh
cbron: We'd have to trap the exit code so we can list all failures. Or do away with that idea.

chart/values.yaml
cbron: Yes I think Raul requested that. Can also be used to clean up other resources, or not clean up fleet.

scripts/chart/build
cbron: No it was just best solution we came up with, grepping for the image. There's no easy way to get to it from this point in build process. However I do like idea of just hardcoding the image name and then parsing for image version. We had thought of that but went this route so that if imaged changed it would catch it, but you are right its very fragile.

chart/templates/post-delete-hook-ServiceAccount.yaml
jiaqiluo: yes, I followed the pattern of existing files. But it is a good time to switch to dashes :)

chart/templates/post-delete-hook-clusterRole.yaml
jiaqiluo: I followed the format of [examples](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-example) from the K8bernetes documentation, such like the following one:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  namespace: default
  name: pod-reader
rules:
- apiGroups: [""] # "" indicates the core API group
  resources: ["pods"]
  verbs: ["get", "watch", "list"]
```

chart/templates/post-delete-hook-clusterRole.yaml
jiaqiluo: fixed.

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed all occurrences

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed.

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed.

chart/scripts/post-delete-hook.sh
jiaqiluo: fixed.

chart/templates/post-delete-hook-clusterRole.yaml
jiaqiluo: fixed all occurrences.

chart/templates/post-delete-hook-configMap.yaml
jiaqiluo: fixed all occurrences.

chart/templates/post-delete-hook-job.yaml
jiaqiluo: fixed.

chart/templates/post-delete-hook-job.yaml
jiaqiluo: fixed.

chart/scripts/post-delete-hook.sh
cbron: Talked with Jack offline and we are going to not use timeout as it doesn't buy us anything but extending default timeout of 5 min.

scripts/chart/build
jiaqiluo: the PR is updated. Now the image name is hardcoded, and only the image tag is grepped from the go file.

chart/scripts/post-delete-hook.sh
jiaqiluo: I'd prefer to keep the `[[ ]]` to keep the consistency of the format.

chart/templates/post-delete-hook-job.yaml
jiaqiluo: > environment variables that come from outside the bash script should be capitalized, e.g. NAMESPACES. 

This is fixed. 

> Also, are the | quote steps necessary?

THe `helm uninstall rancher -n cattle-system` will throw the following error if there are no quotes around those env vars

`>  helm uninstall rancher -n cattle-system
Error: uninstallation completed with 1 error(s): warning: Hook post-delete rancher/templates/post-delete-hook-job.yaml failed: Job in version "v1" cannot be handled as a Job: v1.Job.Spec: v1.JobSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found 1, error found in #10 byte of ...|,"value":120},{"name|..., bigger context ...|cher-operator-system"},{"name":"TIMEOUT","value":120},{"name":"IGNORETIMEOUTERROR","value":false}],"|...
`

chart/scripts/post-delete-hook.sh
cbron: I'm good with this either way. It actually does print out as it works or fails. This is just summary.

cbron (dismissed):
lgtm, once arvind approves we are good to merge and backport to 2.5

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+if [ -z "${ignoreTimeoutError}" ]; then
+  echo "No ignoreTimeoutError value is provided."
+  exit 1
+fi
+
+succeeded=()
+failed=()
+
+get_pod_count() {
+  kubectl get pods --selector app="${1}" -n "${2}" -o json | jq '.items | length'
```
aiyengar2, line 28: Can we assume rancher/shell has jq in it?

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+
+echo "Uninstalling Rancher resources in the following namespaces: ${namespaces}"
+
+for namespace in ${namespaces}; do
+  for app in $(helm list -n "${namespace}" -q); do
+    if [[ ${app} =~ .crd$ ]]; then
```
aiyengar2, line 35:
```suggestion
    if [[ "${app}" =~ ".crd$" ]]; then
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+      continue
+    fi
+    echo "--- Deleting the app [${app}] in the namespace [${namespace}]"
+    if [[ ! $(helm uninstall "${app}" -n "${namespace}") ]]; then
+      failed=("${failed[@]}" "${app}")
+      continue
+    fi
+    t=0
+    # by default, wait 120 seconds in total for an app to be uninstalled
+    while [[ $(get_pod_count "${app}" "${namespace}") != "0" ]] && [[ "${t}" -lt "${timeout}" ]]; do
```
aiyengar2, line 46:
```suggestion
    while [[ "$(get_pod_count ${app} ${namespace})" -ne 0 ]] && [[ "${t}" -lt "${timeout}" ]]; do
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+echo "Uninstalling Rancher resources in the following namespaces: ${namespaces}"
+
+for namespace in ${namespaces}; do
+  for app in $(helm list -n "${namespace}" -q); do
+    if [[ ${app} =~ .crd$ ]]; then
+      echo "--- Skip the app [${app}] in the namespace [${namespace}]"
+      continue
+    fi
+    echo "--- Deleting the app [${app}] in the namespace [${namespace}]"
+    if [[ ! $(helm uninstall "${app}" -n "${namespace}") ]]; then
```
aiyengar2, line 40: Check if this works:

```suggestion
    if ! helm uninstall "${app}" -n "${namespace}"; then
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+      continue
+    fi
+    t=0
+    # by default, wait 120 seconds in total for an app to be uninstalled
+    while [[ $(get_pod_count "${app}" "${namespace}") != "0" ]] && [[ "${t}" -lt "${timeout}" ]]; do
+      echo "waiting 5 seconds for pods of [${app}] to be terminated ..."
+      sleep 5
+      t=$((t + 5))
+    done
+    if [[ $(get_pod_count "${app}" "${namespace}") != "0" ]]; then
```
aiyengar2, line 51:
```suggestion
    if [[ "$(get_pod_count ${app} ${namespace})" -ne 0 ]]; then
```

aiyengar2: If the deletion is successful the first time around, this seems like an unnecessary second call to be made. Could we move the failure logic directly into the previous while statement?

e.g. 

```shell
while : ; do
    pod_count=$(get_pod_count ${app} ${namespace})
    if [[ "${pod_count}" -eq 0 ]]; then
        echo "successfully uninstalled [${app}] in the namespace [${namespace}]"
        succeeded=("${succeeded[@]}" "${app}")
        break
    fi
    if [[ "${t}" -ge "${timeout}" ]]; then
        echo "timeout uninstalling [${app}] in the namespace [${namespace}]"
        failed=("${failed[@]}" "${app}")
        break
    fi
    echo "waiting 5 seconds for pods of [${app}] to be terminated ..."
    sleep 5
    t=$((t + 5))
done
```

(note: `while : ` is bash for `while true`)

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+      echo "timeout uninstalling [${app}] in the namespace [${namespace}]"
+      failed=("${failed[@]}" "${app}")
+    else
+      echo "successfully uninstalled [${app}] in the namespace [${namespace}]"
+      succeeded=("${succeeded[@]}" "${app}")
+    fi
+  done
+  # delete the helm operator pods
+  for pod in $(kubectl get pods -n "${namespace}" -o name); do
+    if [[ ${pod} =~ ^pod\/helm-operation-* ]]; then
```
aiyengar2, line 61:
```suggestion
    if [[ "${pod}" =~ ^pod\/helm-operation-* ]]; then
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+      echo "--- Deleting the pod [${pod}] in the namespace [${namespace}]"
+      kubectl delete "${pod}" -n "${namespace}"
+    fi
+  done
+done
+
+echo "------ Summary ------"
+if [[ ${#succeeded[@]} -ne 0 ]]; then
```
aiyengar2, line 69:
```suggestion
if [[ "${#succeeded[@]}" -ne 0 ]]; then
```

chart/scripts/post-delete-hook.sh
```diff
@@ -0,0 +1,80 @@
...
+  echo "Succeeded to uninstall the following apps:" "${succeeded[@]}"
+fi
+
+if [[ ${#failed[@]} -ne 0 ]]; then
```
aiyengar2, line 73:
```suggestion
if [[ "${#failed[@]}" -ne 0 ]]; then
```

scripts/chart/build
aiyengar2: If we are anyways grepping the image tag, we're imposing a constraint on the Go code regardless of whether we get the image name or not. So IMO we might as well go back to grepping the image name itself as well; my initial preference was to remove this sed script entirely in favor of hard-coding both the image and image tag.

However, I do have one idea: since rancher/rancher is using [Go 1.16](https://github.com/rancher/rancher/blob/baa461856167e28afea67c12667e92b3f242f467/go.mod#L3) which supports [`go:embed`](https://blog.carlmjohnson.net/post/2021/how-to-use-go-embed/), maybe the solution here is to set

```go
const (
        //go:embed shell_image.txt
	ShellDockerImage string
        //go:embed shell_image_tag.txt
        ShellDockerTag string
)

var (
// ...
	ShellImage                        = NewSetting(ShellDockerImage, ShellDockerTag)
// ...
)
```

That way we can just `cat pkg/settings/shell_image.txt` and `cat pkg/settings/shell_image_tag.txt` to get the image and tag for this script instead of sedding.

Otherwise, let's go back to grepping both the image tag + image name.

Any thoughts @cbron @jiaqiluo?

aiyengar2 (changes requested):

chart/scripts/post-delete-hook.sh
jiaqiluo: yes, `jq` is included in the rancher/shell image

https://github.com/rancher/shell/blob/49ecd409727764d202026ab7f28b6820c2e28b5e/package/Dockerfile#L15
```
RUN apk add -U --no-cache bash bash-completion jq
```

chart/scripts/post-delete-hook.sh
jiaqiluo: The double brackets `[[` prevents the word splitting by default, so the quotes around the string variable are optional. 
I replaced other `[`s with `[[`s in the file and no quotes to make use of the good feature of  `[[`.

chart/scripts/post-delete-hook.sh
jiaqiluo: 1. The double brackets `[[` prevents the word splitting by default, so the quotes around the string variable are optional. 
I replaced other `[`s with `[[`s in the file and no quotes to make use of the good feature of  `[[`. 

2. `"$(get_pod_count ${app} ${namespace})"` returns a string not a number, so we need to compare two strings here.

chart/scripts/post-delete-hook.sh
jiaqiluo: The double brackets `[[` prevents the word splitting by default, so the quotes around the string variable are optional. 
I replaced other `[`s with `[[`s in the file and no quotes to make use of the good feature of  `[[`.

chart/scripts/post-delete-hook.sh
jiaqiluo: The double brackets `[[` prevents the word splitting by default, so the quotes around the string variable are optional. 
I replaced other `[`s with `[[`s in the file and no quotes to make use of the good feature of  `[[`.

chart/scripts/post-delete-hook.sh
jiaqiluo: The double brackets `[[` prevents the word splitting by default, so the quotes around the string variable are optional. 
I replaced other `[`s with `[[`s in the file and no quotes to make use of the good feature of  `[[`.

chart/scripts/post-delete-hook.sh
jiaqiluo: it works with or without the double brackets, but I'd prefer to keep the brackets to keep the style consistent.

chart/scripts/post-delete-hook.sh
jiaqiluo: Good point! The script is updated to use this `while` loop.

chart/scripts/post-delete-hook.sh
aiyengar2: 1. Sounds good! Thanks for the clarification
2. Bash doesn’t have the concept of types so strings or numbers don’t make a difference. A number is a “string” with just digits

scripts/chart/build
jiaqiluo: We grep the image name and tag from `pkg/settings/setting.go` because we want to avoid inconsistency and extra dependencies. Using `go:embed` means we will have 2 more files to maintain which is not an ideal solution. 

Going back to grepping both the image tag + image name looks like an acceptable workaround to me.

scripts/chart/build
```diff
@@ -14,3 +14,11 @@ cp -rf ${1} build/chart/rancher
...
+post_delete_image_tag=$(echo "${post_delete_base}" | cut -d ":" -f 2) || ""
+if [[ ! $post_delete_image_tag =~ ^v.* ]]; then
```
aiyengar2, line 20:
```suggestion
if [[ ! ${post_delete_image_tag} =~ ^v.* ]]; then
```

scripts/chart/build
aiyengar2: Agreed

scripts/chart/build
```diff
@@ -14,3 +14,11 @@ cp -rf ${1} build/chart/rancher
 
 sed -i -e "s/%VERSION%/${CHART_VERSION}/g" build/chart/rancher/Chart.yaml
 sed -i -e "s/%APP_VERSION%/${VERSION}/g" build/chart/rancher/Chart.yaml
+
+post_delete_base=$(grep -i shell-image pkg/settings/setting.go | cut -d "," -f 2 | sed -e 's/"//g' | sed -e 's/)//g' | sed -e 's/ //g') || ""
```
aiyengar2, line 18: nit: add a comment on top of this gigantic sed command for humans to read 😄

aiyengar2 (approved):
I have two nit comments and @jiaqiluo needs to add back in the image stuff in https://github.com/rancher/rancher/pull/31670/files#r603693702, but LGTM.

Great work!! 🎉 🎉

scripts/chart/build
jiaqiluo: A comment is added

scripts/chart/build
cbron: Thats very interesting solution Arvind. We could also just use a const there with a string to make grepping settings.go easier. But I am good with this solution for now.

cbron (approved):

PR #31674, opened 2021-03-12
Add generic cloud config upstream refresher

mjura:
Refreshing and checking differences in upstream config spec for all
new cloud drivers can be done by generic service. Code base from existing
eksupstreamrefresher should be reused and prepared for adding new cloud drivers.


pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,226 @@
...
+func (c *clusterRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+	if cluster == nil || cluster.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	// for other cloud drivers, please edit HERE
+	// if obj.Spec.EKSConfig == nil && obj.Spec.GKEConfig == nil && ...
+	if cluster.Spec.EKSConfig == nil {
+		return cluster, nil
+	}
```
thedadams, line 97: I don't think this is necessary if the index key is more descriptive. Just trying to descrease the number of places that need to be modified when we add new hosted providers.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,226 @@
...
+	clusterUpstreamRefresher.clusterClient = wContext.Mgmt.Cluster()
+	clusterUpstreamRefresher.clusterCache = wContext.Mgmt.Cluster().Cache()
+
+	clusterUpstreamRefresher.clusterCache.AddIndexer(isCKSIndexer, func(obj *apimgmtv3.Cluster) ([]string, error) {
+		// for other cloud drivers, please edit HERE
+		// if obj.Spec.EKSConfig == nil && obj.Spec.GKEConfig == nil && ...
+		if obj.Spec.EKSConfig == nil {
+			return []string{}, nil
+		}
+		return []string{"true"}, nil
```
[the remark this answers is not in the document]
rmweir: Yeah, I actually think we should return which provider it is and then pass the key into `c.refreshClusterUpstreamSpec` instead nil checking config fields.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,226 @@
...
+		logrus.Errorf("Error parsing upstream cluster refresh cron. Upstream state will not be refreshed: %v", err)
+		return
+	}
+	clusterUpstreamRefresher.refreshCronJob.Schedule(schedule, cron.FuncJob(clusterUpstreamRefresher.refreshAllUpstreamStates))
+	clusterUpstreamRefresher.refreshCronJob.Start()
+}
+
+func (c *clusterRefreshController) refreshAllUpstreamStates() {
+	logrus.Debugf("Refreshing clusters' upstream states")
+	clusters, err := c.clusterCache.GetByIndex(isCKSIndexer, "true")
```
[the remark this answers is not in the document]
rmweir: Get clusters for each driver then append the lists.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
rmweir: We could also do the upstream nil check here and set the key to "not ready" or no key if that's the case to help condense logic.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	logrus.Infof("checking cluster [%s] upstream state for changes", cluster.Name)
+
+	// for other cloud drivers, please edit HERE
+	// if (cluster.Spec.EKSConfig != nil && cluster.Status.EKSStatus.UpstreamSpec == nil) ||
+	//		(cluster.Spec.GKEConfig != nil && cluster.Status.GKEStatus.UpstreamSpec == nil) ...
+	if cluster.Spec.EKSConfig != nil && cluster.Status.EKSStatus.UpstreamSpec == nil {
+		logrus.Infof("initial upstream spec for cluster [%s] has not been set by cluster handler yet, skipping", cluster.Name)
+		return cluster, nil
```
rmweir, line 106: Should be able to remove this after adding to the indexer.

rmweir (changes requested):
I was brainstorming ways we could use interfaces to limit the number of times we evaluate which cloud provider is used but I think after removing the two at the top of `refreshClusterUpstreamSpec` they will be sufficiently reduced.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
mjura: Please look and review on new

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
mjura: Please look and review on new

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
mjura: Please look and review on new

mjura:
@rmweir @thedadams I have added also mechanism for cleaning up old setting object.

cmurphy:
lgtm, this is working great for me in https://github.com/rancher/rancher/pull/31625

mjura:
@cbron Who can restart Drone or can we get privileges for restarting it ?

Currently CI is broken because drone failed, needs to be restarted:

`default-windows-1809: Error response from daemon: failed during hnsCallRawResponse: hnsCall failed in Win32: Element not found. (0x490)`

mjura:
It was rebased

rmweir (approved):
LGTM, great job! In the future we can consider changing things so that adding a new driver requires less individual additions but I think with how you've written it this would be the easiest part to add a new driver to.

thedadams (approved):

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+		}
+	}
+}
+
+func (c *clusterRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluster, cloudDriver string) (*mgmtv3.Cluster, error) {
+	if cluster == nil || cluster.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
```
cbron, line 101: Was looking for the error check here but looks like it moved above to the start function.

mjura: Yes, it was moved to `func StartClusterUpstreamCronJob()`

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/rancher/rancher/pkg/wrangler"
+	wranglerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/robfig/cron"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+//isCloudKubernetesServiceIndexer
+const (
+	isCKSIndexer = "clusters.management.cattle.io/is-cks"
```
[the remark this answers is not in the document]
mjura: Ok, I have renamed this variable on `isKEIndexer`

pkg/settings/setting.go
```diff
@@ -148,6 +151,7 @@ type Provider interface {
 	Set(name, value string) error
 	SetIfUnset(name, value string) error
 	SetAll(settings map[string]Setting) error
+	Delete(name string) error
```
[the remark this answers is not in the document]
mjura: I would like to point that `Set` and `SetAll` functions are also destructive. All settings will be recreated when removed after restarting Rancher in opposite to situation if they would be modified by `Set`. They will stay in broken state. 

Please explain how users can use it from outside Rancher?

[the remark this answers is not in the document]
cbron: From outside of rancher they can't update it. Users would need to have permission to delete, but we try to avoid anything where they can accidentally mess up the cluster. If they delete it and don't restart Rancher, then Rancher will break. Also if there were custom settings those would be lost. They could overwrite using update but that is also a legitimate thing they need to do.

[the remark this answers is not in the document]
mjura: I have added one more condition to `Delete`, which is preventing from removing settings based on ENV_VARIABLE. Regarding removing settings, I have noticed that users can `Edit` and `Delete` settings from Rancher API. I am not sure if any operation or check are needed to prevent users to break cluster settings. The best would be to do not delete anything if we are worry that something can be broken. As I have checked users can `Update` and `Delete` any setting through Rancher API and this part doesn't change anything with it.

pkg/settings/setting.go
```diff
@@ -95,7 +95,7 @@ var (
 	IgnoreNodeName                    = NewSetting("ignore-node-name", "") // nodes to ignore when syncing v1.node to v3.node
 	NoDefaultAdmin                    = NewSetting("no-default-admin", "")
 	RestrictedDefaultAdmin            = NewSetting("restricted-default-admin", "false") // When bootstrapping the admin for the first time, give them the global role restricted-admin
-	EKSUpstreamRefreshCron            = NewSetting("eks-refresh-cron", "*/5 * * * *")
+	ClusterUpstreamRefreshCron        = NewSetting("eks-refresh-cron", "*/5 * * * *")
```
cbron, line 98: I know you are addressing this later, but to be clear we'll need to revert this as either of our solutions leave eks with its own refresh setting.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,323 @@
...
+
+// migrateEksRefreshCronSetting migrate deprecated EKS refresh cron setting to new format
+// and remove deprecated eks-refresh-cron setting object
+func (c *clusterRefreshController) migrateEksRefreshCronSetting(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+	eksCronSetting, err := c.settingsCache.Get(eksRefreshCronDeprecated)
+	if err != nil {
+		return cluster, err
+	}
+
+	eksSchedule, err := cron.Parse(eksCronSetting.Default)
```
cbron, line 171: Should use value not default or you are not getting the users setting. But see this conversation as it will change this code: https://github.com/rancher/rancher/pull/31771#discussion_r599707655

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return cluster, err
+	}
+	next := eksSchedule.Next(time.Now())
+	refreshTime := int64(eksSchedule.Next(next).Sub(next)/time.Second) * 60
+
+	eks := "eks"
+	eksRefreshSetting, err := c.settingsCache.Get(fmt.Sprintf(refreshCronSettingFormat, eks))
+	eksRefreshSetting.Default = string(refreshTime)
```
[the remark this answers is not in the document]
cbron: Also you should be using the Set function on the setting.

cbron (commented):
Needs to address comments from original PR

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,323 @@
...
+	if provider == apimgmtv3.ClusterDriverEKS &&
+		(cluster.Annotations == nil || cluster.Annotations[eksRefreshCronMigrated] == "") {
+		cluster, err = c.migrateEksRefreshCronSetting(cluster)
+		if err != nil {
+			return cluster, err
+		}
+		cluster.Annotations[eksRefreshCronMigrated] = "true"
+		if cluster, err = c.clusterClient.Update(cluster); err != nil {
+			return cluster, err
+		}
```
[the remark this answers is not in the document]
rmweir: It feels bad but I'm sure it' not an expensive call. Annotation seems heavy too. Can we really not remove settings? If we could only remove deprecated settings none of this would have to be weird.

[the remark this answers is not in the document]
rmweir: Either way it wouldn't make sense to put the annotation on the cluster since this needs to be performed one time, not once for each cluster.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
rmweir: In situations where you have access tot he setting like this, you can just use <setting>.Get() if you need the effective value.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
rmweir: Set() will do it probably but to clarify, each setting has a default and a value. The value field may change but the default field shouldn't ever be set as default should remain whatever it is set to in settings.go. A Get() basically returns the value unless it is empty, then it returns the default.

pkg/settings/setting.go
```diff
@@ -238,3 +238,13 @@ func getMetadataConfig() string {
 	}
 	return string(ans)
 }
+
+// getSettingByID returns a setting that is stored with the given id. If none is found
+// an error is return
+func GetSettingByID(id string) (Setting, error) {
+	setting, ok := settings[id]
```
rmweir, line 245: As @dramich  mentioned on my PR:
```suggestion
	setting, ok := provider.Get(id)
```

rmweir (changes requested):

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
cbron: As an update, the annotation will now go away entirely with our new plan.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,323 @@
...
+		return time.Time{}, fmt.Errorf("unable to parse last KEv2 refresh time [%s]: %v", lastRefreshTime, err)
+	}
+
+	return time.Unix(lastRefreshUnix, 0).Add(refreshInterval), nil
+}
+
+// migrateEksRefreshCronSetting migrate deprecated EKS refresh cron setting to new format
+// and remove deprecated eks-refresh-cron setting object
+func (c *clusterRefreshController) migrateEksRefreshCronSetting(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
+	eksCronSetting, err := c.settingsCache.Get(eksRefreshCronDeprecated)
```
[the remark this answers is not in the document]
mjura: I have re-implemented this part, if setting is `NotFound` then return and continue without error

pkg/settings/setting.go
```diff
@@ -98,7 +98,7 @@ var (
 	IgnoreNodeName                    = NewSetting("ignore-node-name", "") // nodes to ignore when syncing v1.node to v3.node
 	NoDefaultAdmin                    = NewSetting("no-default-admin", "")
 	RestrictedDefaultAdmin            = NewSetting("restricted-default-admin", "false") // When bootstrapping the admin for the first time, give them the global role restricted-admin
-	EKSUpstreamRefreshCron            = NewSetting("eks-refresh-cron", "*/5 * * * *")
+	EKSUpstreamRefresh                = NewSetting("eks-refresh", "300")
```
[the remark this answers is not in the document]
mjura: Cool

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
mjura: I have re-implemented this part, please check it. If `Value` was set for `eks-cron-refresh` then it is taken, if not then `Default`. For setting I used `Set()`

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
mjura: I agree, annotation is not needed. I have re-implemented this once again

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
mjura: If` Value` was set for `eks-cron-refresh` setting then it is taken, if not then `Default` is used.

pkg/settings/setting.go
mjura: After recent changes `func GetSettingByID(id string)` is not needed any more, it can be removed.

mjura:
I have addressed @dramich comments:

- [ ] https://github.com/rancher/rancher/pull/31771#pullrequestreview-618776507
- [ ] https://github.com/rancher/rancher/pull/31771#pullrequestreview-618788856
- [ ] https://github.com/rancher/rancher/pull/31771#pullrequestreview-618791980
- [ ] https://github.com/rancher/rancher/pull/31771#pullrequestreview-618796203

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
cbron: This isn't strictly needed anymore but it works if we do remove that setting in future.

pkg/settings/setting.go
```diff
@@ -98,7 +98,7 @@ var (
 	IgnoreNodeName                    = NewSetting("ignore-node-name", "") // nodes to ignore when syncing v1.node to v3.node
 	NoDefaultAdmin                    = NewSetting("no-default-admin", "")
 	RestrictedDefaultAdmin            = NewSetting("restricted-default-admin", "false") // When bootstrapping the admin for the first time, give them the global role restricted-admin
-	EKSUpstreamRefreshCron            = NewSetting("eks-refresh-cron", "*/5 * * * *")
```
cbron, line 101: Leave this setting as-is

pkg/settings/setting.go
cbron: Lets remove it then.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,137 @@
...
+	if errors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	eksCron := eksCronSetting.Value
+	if eksCronSetting.Value == "" {
+		eksCron = eksCronSetting.Default
+	}
```
cbron, line 117: Just check the Value and if Value is "" then return. We don't want to migrate if its just a default.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrate deprecated EKS refresh cron setting to new format
+// and remove deprecated eks-refresh-cron setting object, if setting is not found then return
+// This function will be run only once in pkg/multiclustermanager/app.go
+func MigrateEksRefreshCronSetting(wContext *wrangler.Context) error {
+	settingsClient := wContext.Mgmt.Setting()
```
cbron, line 105: As I understand it we can't do it this way because the provider (storage) isn't used. We can do how `StartEKSUpstreamCronJob` is done today, just import settings package and check there. Though interestingly we can't get just the Value from there. So I suppose we will have to check if `Get()` and `Default` are different and then if they are we assume its set, and if they are not we assume its default.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
cbron: See comment above^

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if err = settings.EKSUpstreamRefresh.Set(fmt.Sprint(refreshTime)); err != nil {
+		return err
+	}
+
+	// remove deprecated eks-refresh-cron setting
+	err = settingsClient.Delete(eksRefreshCronDeprecated, &metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
```
cbron, line 134: Not deleting

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,137 @@
...
+			cluster.Name,
+			strings.Join(errs, "\n"))
+	}
+
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrate deprecated EKS refresh cron setting to new format
+// and remove deprecated eks-refresh-cron setting object, if setting is not found then return
+// This function will be run only once in pkg/multiclustermanager/app.go
```
cbron, line 103: Comment needs to be updated

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,269 @@
...
+// refreshing a cluster created by the given cloud provider.
+func getProviderRefreshInterval(provider string) (time.Duration, error) {
+	var refreshInterval int
+
+	// for other cloud drivers, please edit HERE
+	switch provider {
+	case apimgmtv3.ClusterDriverEKS:
+		refreshInterval = settings.EKSUpstreamRefresh.GetInt()
+	default:
+		return 0, nil
```
cbron, line 122: nit: default to 300 to be safe

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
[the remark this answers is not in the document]
cbron: Set looks right now.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
cbron: Actually I think I am wrong here

mjura: After bringing back `EKSUpstreamRefreshCron            = NewSetting("eks-refresh-cron", "*/5 * * * *")` to `pkg/settings/setting.go` we can get setting value in both ways.

mjura: According to our discussion on Slack it stayed as it is

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Done

pkg/settings/setting.go
mjura: Done

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Ok, not deleting

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Updated

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
mjura: Updated

pkg/settings/setting.go
cbron: Functionally its fine, but add comment at end of setting like it is for others.
Also looks like there are other changes here, lets revert those. HideLocalCluster setting should not change. The only change here should be new setting and added comment.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,128 @@
...
+		// If there are errors here, we can still sync, but there are problems with the nodegroups that should be reported
+		err = fmt.Errorf("error for cluster [%s]: %s",
+			cluster.Name,
+			strings.Join(errs, "\n"))
+	}
+
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrate deprecated eks-refresh-cron setting to new
```
cbron, line 101:
```suggestion
// MigrateEksRefreshCronSetting migrates the deprecated eks-refresh-cron setting to new
```

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,128 @@
...
+// setting format only if default setting was changed
```
cbron, line 102:
```suggestion
// setting only if default setting was changed
```

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,128 @@
...
+	settingsClient := wContext.Mgmt.Setting()
+
+	eksCronSetting, err := settingsClient.Get(eksRefreshCronDeprecated, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	eksCron := eksCronSetting.Value
```
cbron, line 114: nit: there is no point in making this variable, just use `eksCronSetting.Value`

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,128 @@
...
+
+	eksSchedule, err := cron.Parse(eksCron)
+	if err != nil {
+		return err
+	}
+	next := eksSchedule.Next(time.Now())
+	refreshTime := int(eksSchedule.Next(next).Sub(next)/time.Second) * 60
+
+	err = settings.EKSUpstreamRefresh.Set(fmt.Sprint(refreshTime))
+	return err
```
cbron, line 127: nit: return the function directly

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrate deprecated eks-refresh-cron setting to new
+// setting format only if default setting was changed
+// This function will be run only once during startup by pkg/multiclustermanager/app.go
+func MigrateEksRefreshCronSetting(wContext *wrangler.Context) error {
+	settingsClient := wContext.Mgmt.Setting()
```
rmweir, line 105: @cbron can we use the cache here or is it too early? @mjura if we can use that instead

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,128 @@
...
+	if eksCronSetting.Value == "" {
+		return nil
+	}
+
+	eksSchedule, err := cron.Parse(eksCron)
+	if err != nil {
+		return err
+	}
+	next := eksSchedule.Next(time.Now())
+	refreshTime := int(eksSchedule.Next(next).Sub(next)/time.Second) * 60
```
rmweir, line 124: What is the `* 60` for? Does that not convert it to minutes?

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
rmweir: @cbron the error returned from this function isn't (and probably shouldn't be) used, should we just log them and then return nothing?

rmweir (changes requested):

thedadams (commented):
I think it will be good to go once the current comments are addressed.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: It seems that golang Cron has different format https://pkg.go.dev/github.com/robfig/cron than Linux CronJob, I have changed parse function and used `cron.ParseStandard`

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: As I have tested, on that moment Cache() is already populated so it can be used. I have changed it

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: It was changed according to comment above

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Ok, fixed

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Fixed

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Fixed

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,129 @@
...
+
+// MigrateEksRefreshCronSetting migrates the deprecated eks-refresh-cron setting to new
+// setting only if default setting was changed
+// This function will be run only once during startup by pkg/multiclustermanager/app.go
+func MigrateEksRefreshCronSetting(wContext *wrangler.Context) {
+	settingsClient := wContext.Mgmt.Setting().Cache()
+
+	//eksCronSetting, err := settingsClient.Get(eksRefreshCronDeprecated, metav1.GetOptions{})
+	eksCronSetting, err := settingsClient.Get(eksRefreshCronDeprecated)
+	if errors.IsNotFound(err) || eksCronSetting.Value == "" {
```
rmweir, line 109: Can also return early if the new setting already has a value. We don't want to overwrite it if that is the case.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,129 @@
...
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrates the deprecated eks-refresh-cron setting to new
+// setting only if default setting was changed
+// This function will be run only once during startup by pkg/multiclustermanager/app.go
+func MigrateEksRefreshCronSetting(wContext *wrangler.Context) {
+	settingsClient := wContext.Mgmt.Setting().Cache()
+
+	//eksCronSetting, err := settingsClient.Get(eksRefreshCronDeprecated, metav1.GetOptions{})
```
rmweir, line 107: Remove this.

rmweir (changes requested):

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: ok, good point, I have added additional check
```if settings.EKSUpstreamRefresh.Default != settings.EKSUpstreamRefresh.Get()```

File: pkg/settings/setting.go (hunk 15 above)
Comment (mjura) [reply pkg/settings/setting.go:101]: ok, only two lines are touched

File: pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go (hunk 28 above)
Comment (mjura) [reply pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go]: Removed

File: pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go (hunk 27 above)
Comment (rmweir) [reply pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go:112]: I think you should explicitly check the value like you are doing here for eksCronSetting. It is possible someone would change value then change it back to match default so this doesn't tell you whether they've ever changed it before.

File: pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go [hunk 29]
```diff
@@ -0,0 +1,132 @@
...
+
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrates the deprecated eks-refresh-cron setting to new
+// setting only if default setting was changed
+// This function will be run only once during startup by pkg/multiclustermanager/app.go
+func MigrateEksRefreshCronSetting(wContext *wrangler.Context) {
+	// if the new setting already has a custom value don't overwrite it
+	if settings.EKSUpstreamRefresh.Default != settings.EKSUpstreamRefresh.Get() {
```
rmweir, line 106: Get the setting and check value. It's possible that they would be equal even if the value had been set before.

rmweir (changes requested):

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: I have changed it to `if errors.IsNotFound(err) || eksRefreshSetting.Value != "" `

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: I have checked it to `if errors.IsNotFound(err) || eksRefreshSetting.Value != ""`

rmweir (dismissed):
LGTM

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	nextRefreshTime, err := getNextProviderRefreshTime(providerRefreshInterval, lastRefreshTime)
+	if err != nil {
+		return cluster, err
+	}
+
+	if !time.Now().After(nextRefreshTime) {
+		return cluster, err
+	}
```
dramich, line 84: The last time.Now check should move inside getNextProviderRefreshTime then rename getNextProviderRefreshTime to something like `shouldRefresh` and return a bool and error. This logic fits better inside shouldRefresh and nextRefreshTime isn't used here.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,269 @@
...
+	default:
+		return 300, nil
+	}
+
+	return time.Duration(refreshInterval) * time.Second, nil
+}
+
+// getNextProviderRefreshTIme returns the next time a KEv2 cluster is due to refresh given
+// its last refresh time and cloud provider.
+func getNextProviderRefreshTime(refreshInterval time.Duration, lastRefreshTime string) (time.Time, error) {
```
dramich, line 130:
```suggestion
func shouldRefresh(refreshInterval time.Duration, lastRefreshTime string) (bool, error) {
```

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,136 @@
...
+			strings.Join(errs, "\n"))
+	}
+
+	return upstreamSpec, err
+}
+
+// MigrateEksRefreshCronSetting migrates the deprecated eks-refresh-cron setting to new
+// setting only if default setting was changed
+// This function will be run only once during startup by pkg/multiclustermanager/app.go
+func MigrateEksRefreshCronSetting(wContext *wrangler.Context) {
```
dramich, line 105: The migration should only happen once unless there is an error. To accomplish this the deprecated setting needs to have an annotation added showing the migration was completed. This will serve a 2 purposes: 1. It gives a fast exit - if the old setting has the anno no more processing is required. 2. It shows that rancher did do the migration which is important for testing and from a users/support perspective showing rancher did its job of attempting a 1 time migration.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,136 @@
...
+	settingsClient := wContext.Mgmt.Setting().Cache()
+	eksRefreshSetting, err := settingsClient.Get(eksUpstreamRefresh)
+	if errors.IsNotFound(err) || eksRefreshSetting.Value != "" {
```
dramich, line 108: Errors should always be checked first as standard practice in Go.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	} else if err != nil {
+		logrus.Errorf("Error getting %s setting: %v", eksUpstreamRefresh, err)
```
dramich, line 111: The errors in this section need to be explicit on where they are coming from and what they mean. This is important since we only do this once on startup, so the above error would just print once but as an operator it won't mean anything. 
```suggestion
		logrus.Errorf("Unable to complete Eks cron migration, will attempt at next rancher startup. Error getting %s setting: %v", eksUpstreamRefresh, err)
```

dramich: This is an example of wording but should convey the intent

dramich (changes requested):

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
mjura: Annotation was done

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
mjura: I renamed it and changed func logic

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
mjura: I moved it to `func shouldRefreshCluster()`

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
rmweir: Oh I thought we were just going to check values but this is good too.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/robfig/cron"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	eksUpstreamRefresh       = "eks-refresh"
+	eksRefreshCronDeprecated = "eks-refresh-cron"
+	eksRefreshCronAnnotation = "settings.management.cattle.io/migration-is-done"
```
cbron, line 26: nit: 
```suggestion
	eksRefreshCronAnnotation = "settings.management.cattle.io/migrated"
```

mjura: I have changed this

thedadams (approved):

cbron (commented):
lgtm, just waiting on dan

rmweir (approved):

cbron (approved):

cbron:
Hard-merging, @dramich will post-review.

PR #31680, opened 2021-03-12
[2.5] Add support for creating a GKE cloud credential

cmurphy:
This patch adds the ability to create and store a CloudCredential via
the /v3/cloudcredentials API that can be used to drive GKE clusters.

By adding it to the machine driver controller, Rancher can expose this
credential type as a dynamic schema in the same way that other cloud
credentials are supported. However, since GKE is not a real machine
driver, we have to fake it by creating dummy flags for a pretend machine
driver binary.

The credential is identified by the private field "authEncodedJson"
where the value is an *escaped* string representing a JSON object for
the service account data obtained from Google. If the credential object
is provided as a regular JSON object rather than an escaped string,
Rancher will end up converting into a Go map and then to a string
representation of the map before storing it in a Secret, rather than
directly storing the JSON object that is needed to authenticate with
Google. The result is that creating a cloud credential of this type will
look like this:

```
HTTP/1.1 POST /v3/cloudcredentials

{
  "googlecredentialConfig": {
    "authEncodedJson": "{\"type\": \"service_account\",
      \"project_id\": \"my-project\",
      \"private_key_id\": \"abcdef\",
      \"private_key\": \"-----BEGIN PRIVATE KEY-----\\nABCdef==\\n-----END PRIVATE KEY-----\\n\",
      \"client_email\": \"user@my-project.iam.gserviceaccount.com\",
      \"client_id\": \"123456\",
      \"auth_uri\": \"https://accounts.google.com/o/oauth2/auth\",
      \"token_uri\": \"https://oauth2.googleapis.com/token\",
      \"auth_provider_x509_cert_url\": \"https://www.googleapis.com/oauth2/v1/certs\",
      \"client_x509_cert_url\": \"https://www.googleapis.com/robot/v1/metadata/x509/user%40project.iam.gserviceaccount.com\"}"
  }
}
```

(cherry picked from commit 7269a38314ec9628153f7e6d7819974d88d8e666)

### DIFF
--- a/pkg/controllers/management/drivers/nodedriver/machine_driver.go
+++ b/pkg/controllers/management/drivers/nodedriver/machine_driver.go
@@ -39,6 +39,7 @@ var (
 		"otc":           map[string]string{"privateKeyFile": "privateKeyFile"},
 		"packet":        map[string]string{"userdata": "userdata"},
 		"vmwarevsphere": map[string]string{"cloud-config": "cloudConfig"},
+		"google":        map[string]string{"authEncodedJson": "authEncodedJson"},
 	}
 	SSHKeyFields = map[string]bool{
 		"sshKeyContents": true,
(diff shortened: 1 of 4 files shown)


cbron (approved):
backport lgtm

cbron:
issue: https://github.com/rancher/rancher/issues/30981

mjura (approved):
LGTM

Issue #31600, opened 2021-03-05
Using Private Registry without Authentication fails


meerkampdvv:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

- Setup a internal Registry for airgapped deployments without the need to authenticate. 
- Setup rancher with CATTLE_SYSTEM_DEFAULT_REGISTRY Environment Variable to use the internal Registry
- Create a new Cluster

**Result:**

Node Creation fails with : 
Error: Cannot perform an interactive login from a non TTY device
Process exited with status 1

**Other details that may be helpful:**

in the node log i see the following cmd fail :
`docker login --username --password-stdin`
no username is set. 

version 2.5.5 did not show this behavior.

**Current Workaround:**
When we create a new Cluster we enable the same Private Registry with a user and a password.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):  v2.5.6
- Installation option (single install/HA): HA Helm deployment on 3 rke nodes. `systemDefaultRegistry` value is set to internal Registry.

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): VSphere 
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM 8 Core/16 GiB

gz#15692

mitchellmaler:
We just ran into the same issue using the vSphere Driver after upgrading to Rancher 2.5.6. The only workaround was to disable private registries. 

```
[INFO] [node-controller-rancher-machine] private registry detected, authenticating node_name to ourproxy.com
rancher-7996b6d87b-7sgbb rancher 2021/03/08 21:46:30 [ERROR] [node] enqueing node pool c-p4b99:node-pool
rancher-7996b6d87b-7sgbb rancher 2021/03/08 21:46:30 [ERROR] error syncing 'c-p4b99/m-m2qzr': handler node-controller: Error: Cannot perform an interactive login from a non TTY device
rancher-7996b6d87b-7sgbb rancher Process exited with status 1
rancher-7996b6d87b-7sgbb rancher : exit status 1, requeuing
```

panho66:
I ran into same issue using AWS EC2 Node template Rancher 2.4.14 a fresh installation.

2021/03/10 22:57:52 [ERROR] NodeController c-q7pj8/m-zhgrq [node-controller] failed with : Error: Cannot perform an interactive login from a non TTY device
[...]

PR #31681, opened 2021-03-12
[release/v2.5.7] Private Registry Regression Fix

ryansann:
This adds a fix to check for the case where no credentials are provided to the private registry configuration for a cluster.

Issue: 
- https://github.com/rancher/rancher/issues/31600

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -443,8 +443,9 @@ func (m *Lifecycle) deployAgent(nodeDir string, obj *v3.Node) error {
 // this enables the agent image to be pulled from the private registry
 func (m *Lifecycle) authenticateRegistry(nodeDir string, node *v3.Node, cluster *v3.Cluster) error {
 	reg := util.GetPrivateRepo(cluster)
-	if reg == nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -443,8 +443,9 @@ func (m *Lifecycle) deployAgent(nodeDir string, obj *v3.Node) error {
 // this enables the agent image to be pulled from the private registry
 func (m *Lifecycle) authenticateRegistry(nodeDir string, node *v3.Node, cluster *v3.Cluster) error {
 	reg := util.GetPrivateRepo(cluster)
-	if reg == nil {
-		return nil // if there is no private registry defined, return since auth is not needed
+	// if there is no private registry defined or there is a registry without credentials, return since auth is not needed
+	if reg == nil || reg.User == "" || reg.Password == "" {
```
nickgerace, line 447: Do password-less registries exist? (e.g. you can specify a username but not a password)

ryansann: (discussed offline)

dramich (approved):

Issue #31166, opened 2021-02-05
User ID Tracking in Audit Logs


prachidamle:
This ticket is for the feature/enhancement: "User level audit log traceability."

Description:

Rancher needs to include the Identity Provider name(common name of the user in the external Auth provider) in both the Rancher and downstream Kubernetes audit logs. 

Problem:
A Rancher Admin can not trace an event from the Rancher audit logs and into the Kubernetes audit logs without knowing the mapping of Identity Provider user → u-xXXX id mapping.

Learning the Identity Provider user and u-XXXX mapping requires Rancher API, UI, or local cluster access. For users delivering KaaS without allowing Rancher users access to Rancher there is no way to know the mapping.

KaaS operators are in the critical path of answering questions from cluster owners “who did X in the cluster?”. This inhibits the self-service model.

Requirements:
A downstream cluster admin should be able to look at the Kubernetes audit logs and know that external IDP user X performed an action without needing to view anything in Rancher.

If the audit logs are shipped off of the cluster, a user of the logging system should be able to identify the user in the Identity Provider system.

A Rancher Admin should be able to view Rancher audit logs and follow through to the Kubernetes audit log by Identity Provider username.

Related tickets:
https://rancher.zendesk.com/agent/tickets/7326

gz#17013

prachidamle:
Steps to test this fix for Kubernetes Audit log:

1. On a rancher setup, deploy a cluster and enable audit log feature via the cluster.yml, example this is the config needed at minimum to enable Kubernetes audit logs
`services:
  kube-api:
    audit_log:
      enabled: true
 `
 Details for more config here: https://rancher.com/docs/rke/latest/en/config-options/audit-log/#example-configurations
 
 2. Enable Authentication by configuring some external auth provider like github/LDAP etc.
 
 3. Login as an  admin or a standard user and deploy some workload in say the cluster's default project
 
 4. Check the Kubernetes audit logs on the cluster's node at `/var/log/kube-audit/audit-log.json` path 
 
 5. Grep for the "impersonatedUser" entry in the log and check the "extra" field under it. The username and principalid of the user from the external auth provider should be logged.
 
Sample Snippet of Audit logs as seen for a rancher setup with OpenLDAP auth enabled, when a user scales up a workload:

`root@prachi-doaudit1:~# tail -f /var/log/kube-audit/audit-log.json | grep impersonat
[...]

PR #31711, opened 2021-03-15
Audit changes for logging user info from auth provider

prachidamle:
Adding in "Impersonate-Extra-" headers to log user's principalId and username from the external IDP. 
This needs a fix in norman proxy_store as well. So includes a vendor update for new norman change https://github.com/rancher/norman/pull/396

https://github.com/rancher/rancher/issues/31166

### DIFF
--- a/pkg/auth/providers/activedirectory/activedirectory_provider.go
+++ b/pkg/auth/providers/activedirectory/activedirectory_provider.go
@@ -228,3 +228,10 @@ func (p *adProvider) getDNAndScopeFromPrincipalID(principalID string) (string, s
 	externalID := strings.TrimPrefix(parts[1], "//")
 	return externalID, scope, nil
 }
+
+func (p *adProvider) GetUserExtraAttributes(token *v3.Token) map[string][]string {
+	extras := make(map[string][]string)
+	extras["principalid"] = []string{token.UserPrincipal.Name}
+	extras["username"] = []string{token.UserPrincipal.LoginName}
+	return extras
+}
--- a/pkg/auth/providers/common/provider.go
+++ b/pkg/auth/providers/common/provider.go
@@ -16,4 +16,5 @@ type AuthProvider interface {
 	TransformToAuthProvider(authConfig map[string]interface{}) (map[string]interface{}, error)
 	RefetchGroupPrincipals(principalID string, secret string) ([]v3.Principal, error)
 	CanAccessWithGroupProviders(userPrincipalID string, groups []v3.Principal) (bool, error)
+	GetUserExtraAttributes(token *v3.Token) map[string][]string
 }
--- a/pkg/auth/requests/authenticate.go
+++ b/pkg/auth/requests/authenticate.go
@@ -8,10 +8,12 @@ import (
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/rancher/pkg/auth/providerrefresh"
+	"github.com/rancher/rancher/pkg/auth/providers"
 	"github.com/rancher/rancher/pkg/auth/tokens"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/rancher/steve/pkg/auth"
+	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apiserver/pkg/authentication/user"
@@ -23,18 +25,30 @@ var (
 )
 
 type Authenticator interface {
-	Authenticate(req *http.Request) (authed bool, user string, groups []string, err error)
+	Authenticate(req *http.Request) (*AuthenticatorResponse, error)
 	TokenFromRequest(req *http.Request) (*v3.Token, error)
+}
+
+type AuthenticatorResponse struct {
+	IsAuthed      bool
+	User          string
+	UserPrincipal string
+	Groups        []string
+	Extras        map[string][]string
 }
 
 func ToAuthMiddleware(a Authenticator) auth.Middleware {
 	f := func(req *http.Request) (user.Info, bool, error) {
-		authed, u, groups, err := a.Authenticate(req)
+		authResp, err := a.Authenticate(req)
+		if err != nil {
+			return nil, false, err
+		}
 		return &user.DefaultInfo{
-			Name:   u,
-			UID:    u,
-			Groups: groups,
-		}, authed, err
+			Name:   authResp.User,
+			UID:    authResp.User,
+			Groups: authResp.Groups,
+			Extra:  authResp.Extras,
+		}, authResp.IsAuthed, err
 	}
 	return auth.ToMiddleware(auth.AuthenticatorFunc(f))
 }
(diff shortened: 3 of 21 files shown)


pkg/auth/providers/activedirectory/activedirectory_provider.go
```diff
@@ -228,3 +228,10 @@ func (p *adProvider) getDNAndScopeFromPrincipalID(principalID string) (string, s
 	externalID := strings.TrimPrefix(parts[1], "//")
 	return externalID, scope, nil
 }
+
+func (p *adProvider) GetUserAttributesForAuditLog(token *v3.Token) map[string][]string {
+	extras := make(map[string][]string)
+	extras["principalId"] = []string{token.UserPrincipal.Name}
```
dramich, line 234: I think these need to be lowercase keys based on the description in the code and https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy
https://github.com/kubernetes/apiserver/blob/master/pkg/authentication/user/user.go#L31

pkg/auth/requests/authenticate.go
```diff
@@ -2,6 +2,8 @@ package requests
 
 import (
 	"context"
+	"github.com/rancher/rancher/pkg/auth/providers"
+	"github.com/sirupsen/logrus"
```
dramich, line 6: Imports should be sorted, built in at the top with an empty line and third party imports below

pkg/auth/requests/authenticate.go
```diff
@@ -81,31 +92,37 @@ func tokenKeyIndexer(obj interface{}) ([]string, error) {
...
 }
 
-func (a *tokenAuthenticator) Authenticate(req *http.Request) (bool, string, []string, error) {
+func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResponse, error) {
+	authResp := &AuthenticatorResponse{
+		IsAuthed: false,
+		User: "",
+		UserPrincipal: "",
+		Groups: []string{},
+	}
```
dramich, line 101: This is initializing all the fields to the zero value, which we don't need to do, but it should be done for the map. A nil map will cause a panic if it is accessed.

pkg/auth/requests/authenticate.go
```diff
@@ -81,31 +92,37 @@ func tokenKeyIndexer(obj interface{}) ([]string, error) {
 	return []string{token.Token}, nil
 }
 
-func (a *tokenAuthenticator) Authenticate(req *http.Request) (bool, string, []string, error) {
+func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResponse, error) {
```
dramich, line 95: if the return is not a pointer then the errors below can return nil without having to create an object

pkg/auth/requests/authenticate.go
```diff
@@ -137,7 +154,14 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (bool, string, []st
 		go a.userAuthRefresher.TriggerUserRefresh(token.UserID, false)
 	}
 
-	return true, token.UserID, groups, nil
+	authResp.IsAuthed = true
+	authResp.User = token.UserID
+	authResp.UserPrincipal = token.UserPrincipal.Name
+	authResp.Groups = groups
+	logrus.Debugf("Auth filter calling provider to get extra info %v", token.AuthProvider)
+	authResp.Extras = providers.GetAttributesForAuditLog(token.AuthProvider, token)
```
dramich, line 162: Is there a possibility this ends up being a nil map? If so we should guard against that

pkg/auth/requests/chained.go
```diff
@@ -16,14 +16,19 @@ type chainedAuth struct {
 	auths []Authenticator
 }
 
-func (c *chainedAuth) Authenticate(req *http.Request) (authed bool, user string, groups []string, err error) {
+func (c *chainedAuth) Authenticate(req *http.Request) (*AuthenticatorResponse, error) {
```
dramich, line 19: Same comment as above about returning the pointer and the zero initialization below

pkg/auth/requests/filter.go
```diff
@@ -1,6 +1,7 @@
 package requests
 
 import (
+	"github.com/sirupsen/logrus"
```
dramich, line 4: Imports need to be sorted

pkg/auth/requests/filter.go
```diff
@@ -40,6 +44,14 @@ func (h authHeaderHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 		req.Header.Add("Impersonate-Group", group)
 	}
 
+	for key, extras := range userInfo.GetExtra() {
+		for _, s := range extras {
+			req.Header.Add("Impersonate-Extra-"+key, s)
+		}
+	}
+
+	logrus.Debugf("Rancher Auth Filter ##headers %v: ", req.Header)
```
dramich, line 53: This should be at trace since it has auth info in it.

pkg/auth/requests/filter.go
```diff
@@ -27,10 +28,13 @@ func (h authHeaderHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 		return
 	}
 
-	// clean extra
+	//clean extra that is not part of userInfo
 	for header := range req.Header {
 		if strings.HasPrefix(header, "Impersonate-Extra-") {
-			req.Header.Del(header)
+			key := strings.TrimPrefix(header, "Impersonate-Extra-")
+			if _, ok := userInfo.GetExtra()[key]; !ok {
```
dramich, line 35: I don't think we can rely on the data that came from the request to do the filtering here, we need a list of ones we accept

pkg/auth/requests/impersonate.go
```diff
@@ -29,6 +29,7 @@ func (h *impersonatingAuth) Authenticate(req *http.Request) (k8sUser.Info, bool,
 	}
 	user := userInfo.GetName()
 	groups := userInfo.GetGroups()
+	extra := userInfo.GetExtra()
```
dramich, line 32: This can be moved down where it's needed, the other two vars are used in the below code blocks

pkg/auth/providers/common/provider.go
```diff
@@ -16,4 +16,5 @@ type AuthProvider interface {
 	TransformToAuthProvider(authConfig map[string]interface{}) (map[string]interface{}, error)
 	RefetchGroupPrincipals(principalID string, secret string) ([]v3.Principal, error)
 	CanAccessWithGroupProviders(userPrincipalID string, groups []v3.Principal) (bool, error)
+	GetUserAttributesForAuditLog(token *v3.Token) map[string][]string
```
dramich, line 19:
```suggestion
	GetUserExtraAttributes(token *v3.Token) map[string][]string
```

pkg/auth/requests/authenticate.go
prachidamle: they can return nil even when  return is a pointer, right?

pkg/auth/requests/chained.go
prachidamle: changed the initialization to only include the map. But not sure why pointer is a problem

pkg/auth/requests/filter.go
prachidamle: Added a static map to list the currently supported attributes

dramich (approved):

dramich (approved):

Issue #30135, opened 2020-11-18
User created network policies are deleted when disabling Network project isolation on the cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create user network policies following this example [here](https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource) on a cluster
- Enable NPI on the cluster
- hn-nodes and np-default seen additionally
- Disable NPI on the cluster.
- All the network policies are deleted.
```
> kubectl get networkpolicies -A
No resources found
```

**Expected Result:**
User created Network policies should not get deleted.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.7 and 2.5-head - commit id: `71da1569`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): DO RKE
- Kubernetes version (use `kubectl version`):

```
1.18/1.19 (any)
```

kinarashah:
Note: This isn't a regression, behavior is present in earlier versions as well. Ideally we shouldn't be removing user's network policies @leodotcloud what do you think?

leodotcloud:
Agree with @kinarashah. Unfortunately, this has been the behavior since the beginning. The function [here](https://github.com/rancher/rancher/blob/0db277711f3f7a5fbc3801343448707e8f952d69/pkg/controllers/managementuser/networkpolicy/clusterHandler.go#L132) needs to be fixed by not deleting all the network policies, only the ones created by Rancher. (np-*, hp-*, etc)

kinarashah:
Thanks for confirming @leodotcloud! 

@maggieliu this is not a regression and this has been since the beginning of project network isolation feature. Going to move this out of 2.5.3. cc @deniseschannon

kravciak:
I reproduced this on master-head (440b4e5), I will re-test after the fix is merged.

aaronRancher:
This issue is reproduced on Rancher `v2.5.7`.

Steps:
Create user network policies following this example [here](https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource) on a cluster
```
Aarons-MBP:Test yaml aaronyee$ kc apply -f networkpolicy.yaml
networkpolicy.networking.k8s.io/test-network-policy created
[...]

PR #31756, opened 2021-03-18
[release/v2.5] PNI - only delete Rancher created NetworkPolicy resources

ryansann:
When disabling PNI for a cluster, we should only delete NetworkPolicy resources in the downstream cluster that Rancher created. This change implements that behavior so that we don't also delete user created NetworkPolicy resources when disabling PNI.

I opted for a label based approach here as opposed to using NetworkPolicy names or some other approach.

I've tested this with user network policies created via kubectl and via Cluster Explorer UI. In both cases any user created NetworkPolicy resources are not deleted when PNI is disabled and the Rancher created policies are deleted, as expected.

Issue:
- https://github.com/rancher/rancher/issues/30135

### DIFF
--- a/pkg/controllers/managementuser/networkpolicy/clusterHandler.go
+++ b/pkg/controllers/managementuser/networkpolicy/clusterHandler.go
@@ -3,9 +3,8 @@ package networkpolicy
 import (
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


leodotcloud (approved):
LGTM. Thanks for taking care of this.

Issue #31678, opened 2021-03-12
GKE operator: Update proxy calls


cbron:
- [x] Split v1 and v2 calls as necessary
- [x] Make v2 calls work with new cloud credentials

cbron:
Needs main PR merged and 2.5 backport

mjura:
2.6 PR: #31759 
2.5 PR: #31894

cbron:
@prabalsharma or @sowmyav27 need to ensure when testing this one or GKEv2 in general, that GKEv1 clusters continue to be editable. This PR specifically updates the API for them, which should be backwards compatible, but need to double check.

prabalsharma:
@aaronRancher I guess you have tested this as part of the test plan.

aaronRancher:
Validations were done as part of the GKE test plan.

Tested GKEv1 compatibility with the following test cases:
- From Rancher v2.5.7, deploy a GKEv1 cluster. Upgrade to v2.5-head and check compatibility.
- Edit a GKEv1 cluster in a v2.5-head setup.

Tested V2 cloud credentials with the following test cases:
- Create, edit, and delete Google cloud credentials.
- Update Google cloud credentials in a cluster edit. Verify cluster can be edited with new valid credentials.
- Test GKEv2 functionality with Google cloud credentials.
- Verify a cluster edit cannot be made with invalid creds (errors with: `Get "https://container.googleapis.com/...": oauth2: cannot fetch token: 400 Bad Request Response: {"error":"invalid_grant","error_description":"Invalid grant: account not found"}`).

PR #31759, opened 2021-03-19
Refactor GKE proxy

mjura:
GKE Kontainer Engine v2 will use cloud credentials. For checking GKE cluster capabilities cloud credentials should be also used.

issue: https://github.com/rancher/rancher/issues/31678


pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,185 @@
...
+			util.ReturnHTTPError(writer, req, errCode, err.Error())
+			return
+		}
+		writer.Write(serialized)
+	default:
+		util.ReturnHTTPError(writer, req, httperror.NotFound.Status, "invalid endpoint "+resourceType)
+	}
+}
+
+func (h *handler) getCloudCredentials(req *http.Request, cap *Capabilities, credID string) error {
```
cmurphy, line 122:
```suggestion
func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credID string) error {
```

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,185 @@
...
+			return
+		}
+
+	} else if req.Method == http.MethodPost {
+		err := h.getCredentialsFromBody(writer, req, capa)
+		if err != nil {
+			return
+		}
+
+		if capa.Region == "" {
```
cmurphy, line 61: I'm confused how this relates to the cloud credential, why is this check here rather than in the subnetwork handler the way it was before?

cmurphy:
I was able to test this and it didn't quite work. Here's how I tested it:

1) In the API view of Rancher, navigate to `/v3/cloudcredentials`, click create, and in the `googlecredentialConfig` field enter `{"authEncodedJson": "<escaped service account JSON>"}` with the double quotes escaped, the newlines within the private key escaped, and other newlines removed.
2) Note the ID e.g. `"id": "cattle-global-data:cc-n6dwv"`
3) Create an API key in the web UI and copy the bearer token
4) Make a request:

```
curl -i -k -u $TOKEN "https://localhost:8443/meta/gkeSubnetworks?cloudCredentialId=cattle-global-data:cc-n6dwv&region=us-west1
```

It was giving me 500 errors. I made a PR here https://github.com/mjura/rancher/pull/1 to fix the issues, feel free to squash it into your commit.

pkg/api/norman/customization/gke/handler.go
mjura: I renamed it, question about `getCredentialsFromBody` function name. Should it stay as it is or also rename ?

pkg/api/norman/customization/gke/handler.go
mjura: I am also confused, I moved it to proper place

mjura:
Kudos to @cmurphy for testing and helping to make it work

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	if err = json.Unmarshal(raw, &cap); err != nil {
+		writer.WriteHeader(http.StatusBadRequest)
+		return fmt.Errorf("cannot parse request body: " + err.Error())
+	}
+
+	if cap.Region == "" {
+		writer.WriteHeader(http.StatusBadRequest)
+		return fmt.Errorf("invalid region name")
```
cmurphy, line 171: I think this is only required if querying `/meta/gkeSubnetworks` https://github.com/rancher/rancher/pull/31759/files#diff-68792714fef67a0819b8cccf08dbf2e76502031d314c37c390435358d240220aL31 so probably best to leave the validation in the `listSubnetworks` function

pkg/multiclustermanager/capabilities/gke_version_endpoint.go
```diff
@@ -1,49 +0,0 @@
...
-func (g *GKEVersionHandler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
-	body := preCheck(writer, req)
-	if body == nil {
-		return
-	}
-
-	zone := body.Zone
-	if zone == "" {
-		writer.WriteHeader(http.StatusBadRequest)
-		handleErr(writer, fmt.Errorf("invalid zone"))
```
cmurphy, line 34: This validation is missing now, it should be added to `listZones`

cmurphy (changes requested):

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+}
+
+func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
+
+	writer.Header().Set("Content-Type", "application/json")
+
+	capa := &Capabilities{}
+
+	if credID := req.URL.Query().Get("cloudCredentialId"); credID != "" {
```
cbron, line 50: If gkeV2 we are using a GET call ? Why not a POST ?

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,184 @@
+package gke
```
cbron, line 1: Just scanning through these, do any of them have changes for gkev2 ? For instance where is shared VPC's ? Any differences between v1 and v2 will have to be handled separately, as v1 is not changing at all so it must work the same as it does today.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,194 @@
...
+	var serialized []byte
+	var errCode int
+	var err error
+
+	resourceType := mux.Vars(req)["resource"]
+
+	switch resourceType {
+	case "gkeMachineTypes":
+		if serialized, errCode, err = listMachineTypes(req.Context(), capa); err != nil {
+			logrus.Debugf("[gke-handler] error getting machine types: %v", err)
```
cbron, line 73: Looks like we changed how the call gets returned. Specifically that postCheck isn't called anymore. Are we still writing necessary headers ?

pkg/api/norman/customization/gke/handler.go
mjura: I moved it to proper functions, thank you for pointing it

pkg/multiclustermanager/capabilities/gke_version_endpoint.go
mjura: I have added it to proper functions

pkg/api/norman/customization/gke/handler.go
mjura: It is getting `cloudCredentialId` parameter from HTTP request URL

mjura: We still use POST method

pkg/api/norman/customization/gke/listers.go
mjura: Both `gkev1` and `gkev2` were integrated here

pkg/api/norman/customization/gke/handler.go
mjura: It works for gke v1 and gke v2 with Cloudcredentials

pkg/api/norman/customization/gke/listers.go
mjura: I have to checked shared VPC function @cmurphy do you have idea how I get shared VPC list ?

pkg/api/norman/customization/gke/handler.go
cmurphy: I guess the problem is figuring out whether to use the new cloud credential method versus the raw credential JSON? It is a little weird to use POST with URL query parameters. What about using POST but adding `CloudCredentialID` to the `Capabilities` struct and checking that one of either `CloudCredentialID` or `Credentials` is populated?

pkg/api/norman/customization/gke/listers.go
cmurphy: That's moving the goal posts on this PR a little, isn't it? I thought this PR was about using cloud credentials on the existing proxy methods.

There's no all-encompassing GET API for shared VPCs. You can get networks for a project and you can get the [shared VPC host project for a project](https://cloud.google.com/compute/docs/reference/rest/v1/projects/getXpnHost). What do you imagine this proxy method would return, how would the UI look? We could make recursive calls to build a tree of projects and networks...we should have that discussion as a new feature in a new PR, not as an addon to this one.

I don't see any other methods needed here for gke v2.

pkg/multiclustermanager/capabilities/gke_version_endpoint.go
cmurphy: :+1: `listZones` was wrong, you added them to the right place

pkg/api/norman/customization/gke/handler.go
mjura: This was implementation was done based on `oci` and `vsphere` module. So probably this would be done on UI site.

pkg/api/norman/customization/gke/listers.go
mjura: I agree that implementation is done.

mjura:
I have re-tested it with RAW credential JSON and with Cloud Credentials defined. It works as expected.

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,197 @@
...
+	return data, http.StatusOK, err
+}
+
+func listNetworks(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
+	client, err := getComputeServiceClient(ctx, cap.Credentials)
+	if err != nil {
+		return nil, httperror.ServerError.Status, err
+	}
+
+	result, err := client.Networks.List(cap.ProjectID).Do()
```
cmurphy, line 101: Should validate that projectId isn't empty before using it, here and for every other list function that uses it

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,197 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+	return service, nil
+}
+
+func listMachineTypes(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
+	if cap.Zone == "" {
+		return nil, httperror.ServerError.Status, fmt.Errorf("zone is required")
```
cmurphy, line 74: is `ServerError` the right error to use here? This would be caused by a malformed request so it would normally be a 4xx error.

pkg/api/norman/customization/gke/listers.go
mjura: Ok, I have added it

mjura: ok, I have added it

pkg/api/norman/customization/gke/listers.go
mjura: Ok, you are right, I propose `422` which is `httperror.InvalidOption`

https://pkg.go.dev/github.com/rancher/norman/httperror@v0.0.0-20210225010917-c7fd1e24145b#InvalidAction

pkg/multiclustermanager/routes.go
```diff
@@ -97,12 +98,7 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 
 	authed.Path("/meta/aksVersions").Handler(capabilities.NewAKSVersionsHandler())
 	authed.Path("/meta/aksVirtualNetworks").Handler(capabilities.NewAKSVirtualNetworksHandler())
-	authed.Path("/meta/gkeMachineTypes").Handler(capabilities.NewGKEMachineTypesHandler())
-	authed.Path("/meta/gkeNetworks").Handler(capabilities.NewGKENetworksHandler())
-	authed.Path("/meta/gkeServiceAccounts").Handler(capabilities.NewGKEServiceAccountsHandler())
-	authed.Path("/meta/gkeSubnetworks").Handler(capabilities.NewGKESubnetworksHandler())
-	authed.Path("/meta/gkeVersions").Handler(capabilities.NewGKEVersionsHandler())
-	authed.Path("/meta/gkeZones").Handler(capabilities.NewGKEZonesHandler())
+	authed.Path("/meta/{resource}").Handler(gke.NewGKEHandler(scaledContext))
```
thedadams, line 101: I believe this will match a lot more paths than desired. For example, this will match `/meta/anything-not-matched-by-others`.

Perhaps a different approach would be:
```suggestion
	authed.Path("/meta/{resource:gke.+}").Handler(gke.NewGKEHandler(scaledContext))
```
This will match anything that starts with `gke`.

pkg/multiclustermanager/capabilities/http_request_body.go
```diff
@@ -5,8 +5,29 @@ import (
...
+	asJSON, err := json.Marshal(resp)
+
+	if err != nil {
+		logrus.Error("error while marshalling error message '" + originalErr.Error() + "' error was '" + err.Error() + "'")
+		writer.Write([]byte(err.Error()))
+
+		return
+	}
+
+	writer.Write([]byte(asJSON))
```
thedadams, line 28: `json.Marshal` returns `[]byte`.
```suggestion
	writer.Write(asJSON)
```

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+			util.ReturnHTTPError(writer, req, httperror.InvalidBodyContent.Status, err.Error())
+			return
+		}
+	} else {
+		util.ReturnHTTPError(writer, req, httperror.Unauthorized.Status, "cannot access Google API without credentials to authenticate")
+	}
+
+	var serialized []byte
+	var errCode int
+	var err error
```
thedadams, line 66: nit: remove these since each case is handled in its own scope.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+	if err := access.ByID(h.generateAPIContext(req), &schema.Version, client.CloudCredentialType, credID, &accessCred); err != nil {
+		if apiError, ok := err.(*httperror.APIError); ok {
+			if apiError.Code.Status == httperror.PermissionDenied.Status || apiError.Code.Status == httperror.NotFound.Status {
+				return fmt.Errorf("cloud credential not found")
+			}
+		}
+		return err
+	}
+
+	cc, err := h.secretsLister.Get(namespace.GlobalNamespace, name)
```
thedadams, line 135: Should `ns` from above be used as the namespace here?

thedadams (changes requested):

pkg/api/norman/customization/gke/handler.go
mjura: this I would like keep it, when I removed it `goland` doesn't like it

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return oauth2.NewClient(ctx, ts.TokenSource), nil
+}
+
+func getComputeServiceClient(ctx context.Context, credentialContent string) (*compute.Service, error) {
+	client, err := getOAuthClient(ctx, credentialContent)
+
```
thedadams, line 29: Remove line so that error check is directly below `err` assignment. Here and other places in this file.

thedadams (commented):
One minor comment about error checks. Otherwise, looks good.

pkg/api/norman/customization/gke/listers.go
mjura: Ahh, you are right. Done

thedadams (dismissed):

cbron:
issue: https://github.com/rancher/rancher/issues/31678

cmurphy (dismissed):
You could squash the commits into one if you like

pkg/multiclustermanager/routes.go
```diff
@@ -97,12 +98,7 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 
 	authed.Path("/meta/aksVersions").Handler(capabilities.NewAKSVersionsHandler())
 	authed.Path("/meta/aksVirtualNetworks").Handler(capabilities.NewAKSVirtualNetworksHandler())
-	authed.Path("/meta/gkeMachineTypes").Handler(capabilities.NewGKEMachineTypesHandler())
-	authed.Path("/meta/gkeNetworks").Handler(capabilities.NewGKENetworksHandler())
-	authed.Path("/meta/gkeServiceAccounts").Handler(capabilities.NewGKEServiceAccountsHandler())
-	authed.Path("/meta/gkeSubnetworks").Handler(capabilities.NewGKESubnetworksHandler())
-	authed.Path("/meta/gkeVersions").Handler(capabilities.NewGKEVersionsHandler())
-	authed.Path("/meta/gkeZones").Handler(capabilities.NewGKEZonesHandler())
+	authed.Path("/meta/{resource:gke.+}").Handler(gke.NewGKEHandler(scaledContext))
```
cbron, line 101: We need to guarantee that v1 requests haven't changed at all.

cbron (changes requested):
Needs some adjustments

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	} else {
+		util.ReturnHTTPError(writer, req, httperror.Unauthorized.Status, "cannot access Google API without credentials to authenticate")
+	}
+
+	var serialized []byte
+	var errCode int
+	var err error
+
+	resourceType := mux.Vars(req)["resource"]
```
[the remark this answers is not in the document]
thedadams: Reading from a map, even a `nil` map, is valid in Go: https://play.golang.org/p/SbJwNy1XOvH

pkg/multiclustermanager/routes.go
mjura: URL requests were not changed fro v1, double checked

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	var accessCred client.CloudCredential //var to check access
+	if err := access.ByID(h.generateAPIContext(req), &schema.Version, client.CloudCredentialType, credID, &accessCred); err != nil {
+		if apiError, ok := err.(*httperror.APIError); ok {
+			if apiError.Code.Status == httperror.PermissionDenied.Status || apiError.Code.Status == httperror.NotFound.Status {
+				return fmt.Errorf("cloud credential not found")
+			}
+		}
+		return err
```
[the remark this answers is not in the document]
mjura: I have added `handleError` function

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
+
+	writer.Header().Set("Content-Type", "application/json")
+
+	capa := &Capabilities{}
+
+	if credID := req.URL.Query().Get("cloudCredentialId"); credID != "" {
+		if err := h.getCloudCredential(req, capa, credID); err != nil {
+			util.ReturnHTTPError(writer, req, httperror.InvalidBodyContent.Status, err.Error())
```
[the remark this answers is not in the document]
mjura: I have added `handleError` function

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return nil, httperror.ServerError.Status, err
+	}
+
+	result, err := client.MachineTypes.List(cap.ProjectID, cap.Zone).Do()
+	if err != nil {
+		return nil, httperror.ServerError.Status, err
+	}
+
+	data, err := json.Marshal(&result)
```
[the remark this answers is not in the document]
mjura: Added helper function

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,206 @@
...
+	return oauth2.NewClient(ctx, ts.TokenSource), nil
+}
+
+func getComputeServiceClient(ctx context.Context, credentialContent string) (*compute.Service, error) {
+	client, err := getOAuthClient(ctx, credentialContent)
+	if err != nil {
+		return nil, err
+	}
+
+	service, err := compute.New(client)
```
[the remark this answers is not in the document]
mjura: I have reimplemented deprecated functions

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		logrus.Debugf("[GKE] error accessing cloud credential %s", credID)
+		return fmt.Errorf("error accessing cloud credential %s", credID)
+	}
+	cap.Credentials = string(cc.Data["googlecredentialConfig-authEncodedJson"])
+
+	projectID := req.URL.Query().Get("projectId")
+	if projectID != "" {
+		cap.ProjectID = projectID
+	}
```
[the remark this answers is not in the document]
mjura: It was changed

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,189 @@
...
+	if cap.ProjectID == "" {
+		writer.WriteHeader(http.StatusBadRequest)
+		return fmt.Errorf("invalid projectId")
+	}
+
+	return nil
+}
+
+func (h *handler) generateAPIContext(req *http.Request) *types.APIContext {
```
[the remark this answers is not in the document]
mjura: I think this same

cmurphy (dismissed):

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	return nil
+}
+
+func (h *handler) getCredentialsFromBody(writer http.ResponseWriter, req *http.Request, cap *Capabilities) error {
+	raw, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		writer.WriteHeader(http.StatusBadRequest)
+		return fmt.Errorf("cannot read request body: " + err.Error())
```
thedadams, line 167: nit, similar to below:
```suggestion
		return fmt.Errorf("cannot read request body: %v", err)
```

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,208 @@
...
+		Query:   map[string][]string{},
+	}
+}
+
+func handleErr(writer http.ResponseWriter, errorCode int, originalErr error) {
+	resp := errorResponse{originalErr.Error()}
+
+	asJSON, err := json.Marshal(resp)
+	if err != nil {
+		logrus.Error("error while marshalling error message '" + originalErr.Error() + "' error was '" + err.Error() + "'")
```
thedadams, line 201: Use `logrus.Errorf` with formatting verbs.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,208 @@
...
+		Request: req,
+		Schemas: h.schemas,
+		Query:   map[string][]string{},
+	}
+}
+
+func handleErr(writer http.ResponseWriter, errorCode int, originalErr error) {
+	resp := errorResponse{originalErr.Error()}
+
+	asJSON, err := json.Marshal(resp)
```
thedadams, line 199: nit: format to `[]byte` directly and avoid the necessity of error checking and the `errorReponse` type.
```suggestion
	asJSON := []byte(fmt.Sprintf(`{"error":"%v"}`, originalErr))
```

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,208 @@
...
+		cap.Zone = zone
+	}
+
+	return nil
+}
+
+func (h *handler) getCredentialsFromBody(writer http.ResponseWriter, req *http.Request, cap *Capabilities) error {
+	raw, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		writer.WriteHeader(http.StatusBadRequest)
```
thedadams, line 166: I believe these errors are getting passed to the `handleError` function which calls `writer.WriteHeader` again. This should probably only be called once to avoid confusion.

Below instances as well.

thedadams (changes requested):

pkg/api/norman/customization/gke/listers.go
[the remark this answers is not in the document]
cbron: Can you clarify what you mean or did here ? What I meant was to upgrade the client in go.mod: https://github.com/rancher/rancher/blob/master/go.mod#L118

pkg/api/norman/customization/gke/handler.go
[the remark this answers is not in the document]
cbron: handleErr isn't used here. You don't want to return the actual err as it could give away information to user. Lets just follow precedent of oci and do 404: https://github.com/rancher/rancher/blob/master/pkg/api/norman/customization/oci/handler.go#L152

pkg/api/norman/customization/gke/handler.go
cbron: can you check this @mjura

pkg/api/norman/customization/gke/handler.go
cbron: Good catch

pkg/api/norman/customization/gke/handler.go
mjura: Fixed

pkg/api/norman/customization/gke/handler.go
mjura: fixed

pkg/api/norman/customization/gke/handler.go
mjura: thank you, it looks great

pkg/api/norman/customization/gke/handler.go
mjura: done, all errors handled by `handleError` function

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,193 @@
...
+		cap.Zone = zone
+	}
+
+	return nil
+}
+
+func (h *handler) getCredentialsFromBody(writer http.ResponseWriter, req *http.Request, cap *Capabilities) error {
+	raw, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		return fmt.Errorf("cannot read request body: %v", err.Error())
```
thedadams, line 162: nit: don't need to call `Error()` on the errors. The `%v` verb will take care of it.

One more below.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,193 @@
...
+		writer.Write(serialized)
+	case "gkeZones":
+		if serialized, errCode, err = listZones(req.Context(), capa); err != nil {
+			logrus.Debugf("[gke-handler] error getting zones: %v", err)
+			handleErr(writer, errCode, err)
+			return
+		}
+		writer.Write(serialized)
+	default:
+		handleErr(writer, httperror.NotFound.Status, fmt.Errorf("invalid endpoint "+resourceType))
```
thedadams, line 113: nit:
```suggestion
		handleErr(writer, httperror.NotFound.Status, fmt.Errorf("invalid endpoint %v", resourceType))
```

thedadams (dismissed):
Two small nits. I'm fine with this getting merged without addressing them.

pkg/api/norman/customization/gke/listers.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return service, nil
+}
+
+func listMachineTypes(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
+	if cap.ProjectID == "" || cap.Zone == "" {
+		return nil, httperror.InvalidOption.Status, fmt.Errorf("projectId and zone are required")
```
cbron, line 71: This is not same status code as before. Need to either confirm they are different with Wes or keep the same.

pkg/api/norman/customization/gke/listers.go
cbron: These need to be whatever they were in v1.

pkg/api/norman/customization/gke/handler.go
mjura: ok, thank you, done

pkg/api/norman/customization/gke/handler.go
mjura: done

pkg/api/norman/customization/gke/handler.go
mjura: @westlywright can you please judge on this ? do you know if HTTP error CODE is checked somewhere in UI? As I have checked and tested it is not, but please confirm. According to comment from @cmurphy we would like to use more proper HTTP error codes https://github.com/rancher/rancher/pull/31759#discussion_r603460276

pkg/api/norman/customization/gke/listers.go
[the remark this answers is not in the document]
mjura: It was already upgraded in https://github.com/rancher/rancher/pull/31625

pkg/api/norman/customization/gke/handler.go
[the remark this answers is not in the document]
mjura: now, it will return `return httperror.NotFound.Status, err` which will be passed to `handleErr()`

pkg/api/norman/customization/gke/listers.go
mjura: @westlywright can you please judge on this ? do you know if HTTP error CODE is checked somewhere in UI? As I have checked and tested it is not, but please confirm. According to comment from @cmurphy we would like to use more proper HTTP error codes #31759 (comment)

mjura: I have added back old error code response

pkg/api/norman/customization/gke/handler.go
mjura: I have added back old error code response

cmurphy: My comment before was about returning a 5xx server error for an invalid request, which wouldn't make sense from an API perspective and is also backwards incompatible with how it was before https://github.com/rancher/rancher/pull/31759/files#diff-d86bb7d4d3eda030b6a6ce369760bcf4f949ea0eda69b9d841f1f741c18a5d69L29

So http.StatusBadRequest is fine, that's how it was before and makes sense from an API usability perspective. http.InvalidOption would not have made any sense.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,193 @@
...
+			handleErr(writer, errCode, err)
+			return
+		}
+	} else if req.Method == http.MethodPost {
+		if errCode, err := h.getCredentialsFromBody(writer, req, capa); err != nil {
+			handleErr(writer, errCode, err)
+			return
+		}
+	} else {
+		handleErr(writer, httperror.Unauthorized.Status, fmt.Errorf("cannot access Google API without credentials to authenticate"))
```
cmurphy, line 60: I think this is still not backwards compatible, this is what the error was before https://github.com/rancher/rancher/pull/31759/files#diff-d86bb7d4d3eda030b6a6ce369760bcf4f949ea0eda69b9d841f1f741c18a5d69L66

cmurphy: Never mind, I guess it wouldn't make sense to return Method Not Allowed any more now that the method isn't checked on line 49. But then this should probably return Bad Request rather than Unauthorized I think.

mjura: I have changed it to `http.StatusMethodNotAllowed`

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		logrus.Debugf("[GKE] error accessing cloud credential %s", credID)
+		return httperror.InvalidBodyContent.Status, fmt.Errorf("error accessing cloud credential %s", credID)
+	}
+	cap.Credentials = string(cc.Data["googlecredentialConfig-authEncodedJson"])
+
+	cap.ProjectID = req.URL.Query().Get("projectId")
+	if cap.ProjectID == "" {
+		logrus.Debugf("[GKE] error getting projectId")
+		return httperror.InvalidBodyContent.Status, fmt.Errorf("error getting projectId")
```
cmurphy, line 145: Why `InvalidBodyContent` rather than `BadRequest`? It's inconsistent with the other validation https://github.com/rancher/rancher/pull/31759/files#diff-3791cf107be82e12b9080e5e2ad658a28e794a29522d2a446f9d833c571102daR173

mjura: I have changed it to `http.StatusBadRequest`

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,194 @@
...
+			handleErr(writer, errCode, err)
+			return
+		}
+	} else if req.Method == http.MethodPost {
+		if errCode, err := h.getCredentialsFromBody(writer, req, capa); err != nil {
+			handleErr(writer, errCode, err)
+			return
+		}
+	} else {
+		handleErr(writer, http.StatusMethodNotAllowed, fmt.Errorf("cannot access Google API without credentials to authenticate"))
```
cmurphy, line 60: See last comment https://github.com/rancher/rancher/pull/31759#discussion_r607984638 I'm not sure anymore that this is the right error here

cmurphy (approved):
Still not 100% sure about the error codes but I think it at least won't break things, someone else can weigh in

thedadams (approved):

pkg/api/norman/customization/gke/handler.go
cmurphy: As @cbron pointed out this should probably have a `return` after this line, otherwise it would fall through to the switch statement

cmurphy (changes requested):

pkg/api/norman/customization/gke/handler.go
mjura: Indeed, I have changed it to `http.StatusBadRequest` and added `return`

cmurphy (approved):

pkg/api/norman/customization/gke/handler.go
cbron: Looks like its badRequest now

cbron (approved):

cbron (approved):
lgtm

PR #31771, opened 2021-03-20
Generic refresher

rmweir:


pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/rancher/rancher/pkg/wrangler"
+	wranglerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/robfig/cron"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	noKEv2Provider           = "none"
+	lastRefreshTime          = "clusters.management.cattle.io/ke-last-refresh"
```
mjura, line 25: I would propose to rename it `keLastRefreshTime`

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	if !ready {
+		logrus.Infof("initial upstream spec for cluster [%s] has not been set by cluster handler yet, skipping", cluster.Name)
+		return cluster, nil
+	}
+
+	var lastRefreshTime string
+	if cluster.Annotations != nil {
+		lastRefreshTime = cluster.Annotations[lastRefreshTime]
```
mjura, line 70:
```suggestion
		lastRefreshTime = cluster.Annotations[keLastRefreshTime]
```

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(specMap, cluster.Spec.EKSConfig)
+	}
+	if err != nil {
+		return cluster, err
+	}
+
+	if cluster.Annotations == nil {
+		cluster.Annotations = make(map[string]string)
+	}
+	cluster.Annotations[lastRefreshTime] = strconv.FormatInt(time.Now().Unix(), 10)
```
mjura, line 257:
```suggestion
	cluster.Annotations[keLastRefreshTime] = strconv.FormatInt(time.Now().Unix(), 10)
```

mjura (commented):
We have to change `lastRefreshTime` const name

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
cbron: It's done both ways but typically we have done dashes in annotations.

pkg/settings/setting.go
```diff
@@ -95,7 +95,7 @@ var (
 	IgnoreNodeName                    = NewSetting("ignore-node-name", "") // nodes to ignore when syncing v1.node to v3.node
 	NoDefaultAdmin                    = NewSetting("no-default-admin", "")
 	RestrictedDefaultAdmin            = NewSetting("restricted-default-admin", "false") // When bootstrapping the admin for the first time, give them the global role restricted-admin
-	EKSUpstreamRefreshCron            = NewSetting("eks-refresh-cron", "*/5 * * * *")
+	ClusterUpstreamRefreshCron        = NewSetting("eks-refresh-cron", "*/5 * * * *")
```
cbron, line 98: As discussed in meeting we'll have EKS and GKE and AKS, but take out cron name and cron format. There will be conversion method from cron to int representing seconds in the controller.

pkg/controllers/management/wrangler.go
```diff
@@ -17,7 +17,7 @@ func RegisterWrangler(ctx context.Context, wranglerContext *wrangler.Context, ma
 	if features.Legacy.Enabled() {
 		k3sbasedupgrade.Register(ctx, wranglerContext, management, manager)
 		eks.Register(ctx, wranglerContext, management)
+		clusterupstreamrefresher.Register(ctx, wranglerContext)
```
cbron, line 20: This will need to move outside of legacy block in 2.6. But we may need to leave inside block just to test.

pkg/controllers/management/clusterupstreamrefresher/eks_upstream_spec.go
```diff
@@ -0,0 +1,89 @@
+package clusterupstreamrefresher
```
cbron, line 1: This file looks like its just copied

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+	wranglerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/robfig/cron"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	noKEv2Provider           = "none"
+	lastRefreshTime          = "clusters.management.cattle.io/ke-last-refresh"
+	refreshCronSettingFormat = "%s-refresh-cron"
```
cbron, line 26: This should change to: 
```suggestion
	refreshCronSettingFormat = "%s-refresh"
```

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+			logrus.Infof("initial upstream spec for cluster [%s] has not been set by cluster handler yet, skipping", cluster.Name)
+			return apimgmtv3.ClusterDriverEKS, false
+		}
+		return apimgmtv3.ClusterDriverEKS, true
+	default:
+		return noKEv2Provider, false
+	}
+}
+
+// getProviderRefreshInterval returns the duration that should pass between
```
cbron, line 113: As discussed, this should be revamped to use seconds. If cron format is found in old setting we migrate to seconds.

mjura:
This change was cherry picked and fixed in PR https://github.com/rancher/rancher/pull/31759

pkg/settings/setting.go
```diff
@@ -235,3 +235,13 @@ func getMetadataConfig() string {
...
+
+// getSettingByID returns a setting that is stored with the given id. If none is found
+// an error is return
+func GetSettingByID(id string) (Setting, error) {
+	setting, ok := settings[id]
+	if !ok {
+		return Setting{}, fmt.Errorf("%s is not a valid setting", id)
+	}
+	return setting, nil
+}
```
dramich, line 247: This needs to use the provider, if there is one, otherwise you will be reading the value as it lives in this file, not using values set through the env OR in the k8s setting CRD.

pkg/settings/setting.go
dramich: So what is the upgrade path for a user that already used the old setting? Changing the name creates a new setting CR.

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	provider, ready := getProviderAndReadyStatus(cluster)
+	if provider == noKEv2Provider {
+		// not a KEv2 cluster
+		return cluster, nil
+	}
+
+	if !ready {
+		logrus.Infof("initial upstream spec for cluster [%s] has not been set by cluster handler yet, skipping", cluster.Name)
```
dramich, line 64: This is being logged twice, it exists in `getProviderAndReadyStatus`

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+		return time.Time{}, fmt.Errorf("unable to parse last KEv2 refresh time [%s]: %v", lastRefreshAnno, err)
+	}
+
+	return time.Unix(lastRefreshUnix, 0).Add(refreshInterval), nil
+}
+
+func (c *clusterRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluster, cloudDriver string) (*mgmtv3.Cluster, error) {
+	if cluster == nil || cluster.DeletionTimestamp != nil {
+		return nil, nil
+	}
```
dramich, line 148: This is redundant, the check is already done when the cluster first enters the handler

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		if reflect.DeepEqual(specMap[key], value) {
+			continue
+		}
+		updateClusterConfig = true
+		specMap[key] = value
+	}
+
+	if !updateClusterConfig {
+		logrus.Infof("cluster [%s] matches upstream, skipping spec sync", cluster.Name)
```
dramich, line 241: Should this be a debug message? Seeing this every 5 minutes seems like a lot of logs with not much value

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	// for other cloud drivers, please edit HERE
+	switch cloudDriver {
+	case apimgmtv3.ClusterDriverEKS:
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(specMap, cluster.Spec.EKSConfig)
+	}
+	if err != nil {
+		return cluster, err
+	}
```
dramich, line 252: Can the error be checked where it's created? This looks like it could get into the same issue @thedadams found previously.

pkg/controllers/management/wrangler.go
dramich: How do we decide if this is legacy?

cbron: Legacy is system-charts

pkg/settings/setting.go
cbron: Yes we will be migrating old setting in cron format to new setting in seconds format, then deleting the old setting.

pkg/settings/setting.go
cbron: That is as simple as returning `setting.Get()` right ?

PR #31813, opened 2021-03-26
[release/v2.5] Switch to deterministic naming for RBs and CRBs

ryansann:
Issue:
- rancher/rancher#29932

### DIFF
--- a/pkg/controllers/management/auth/manager.go
+++ b/pkg/controllers/management/auth/manager.go
@@ -145,20 +145,31 @@ func (m *manager) ensureClusterMembershipBinding(roleName, rtbNsAndName string, 
 
 	if len(objs) == 0 {
 		logrus.Infof("[%v] Creating clusterRoleBinding for membership in cluster %v for subject %v", m.controller, cluster.Name, subject.Name)
-		_, err := m.mgmt.RBAC.ClusterRoleBindings("").Create(&v1.ClusterRoleBinding{
+		roleRef := v1.RoleRef{
+			Kind: "ClusterRole",
+			Name: roleName,
+		}
+		crbName := pkgrbac.NameForClusterRoleBinding(roleRef, subject) // use deterministic name for crb
+		_, err = m.mgmt.RBAC.ClusterRoleBindings("").Create(&v1.ClusterRoleBinding{
 			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: "clusterrolebinding-",
+				Name: crbName,
 				Labels: map[string]string{
 					rtbNsAndName: MembershipBindingOwner,
 				},
 			},
 			Subjects: []v1.Subject{subject},
-			RoleRef: v1.RoleRef{
-				Kind: "ClusterRole",
-				Name: roleName,
-			},
+			RoleRef:  roleRef,
 		})
-		return err
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+
+		// if the binding exists but was not found in the index, manually retrieve it so that we can add appropriate labels
+		crb, err := m.mgmt.RBAC.ClusterRoleBindings("").Get(crbName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		objs = append(objs, crb)
 	}
 
 	crb, _ = objs[0].(*v1.ClusterRoleBinding)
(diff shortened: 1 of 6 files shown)


pkg/controllers/management/auth/manager.go
```diff
@@ -218,9 +220,10 @@ func (m *manager) ensureProjectMembershipBinding(roleName, rtbNsAndName, namespa
 
 	if len(objs) == 0 {
 		logrus.Infof("[%v] Creating roleBinding for membership in project %v for subject %v", m.controller, project.Name, subject.Name)
+		rbName := strings.Join([]string{"rolebinding", pkgrbac.HashFromKey(key)}, "-")
 		_, err := m.mgmt.RBAC.RoleBindings(namespace).Create(&v1.RoleBinding{
 			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: "rolebinding-",
+				Name: rbName,
```
prachidamle, line 226: One question: Since the scope of a roleBinding is limited to a namespace, is it possible to have 2 roleBindings with same Name in different namespaces? If yes then should the hashed key consider namespace as well for roleBindings?

ryansann: Yes, this is possible with the current hashing logic, though I see this as desired functionality. RoleBindings are considered to be duplicates if they have the same: 
1. Subject
2. Role
3. Namespace

But since RoleBindings are namespaced resources, we only need to consider the Subject and Role for the hash. For example: 
```
rb1 = RoleBinding{
  Name: rolebinding-fe3adc2e
  Namespace: ns-1
  Subject: user-1
  Role: member
}
rb2 = RoleBinding{
  Name: rolebinding-fe3adc2e
  Namespace: ns-2
  Subject: user-1
  Role: member
}
```
Here rb1 and rb2 have the same Name due to having the same Subject and Role, but due to being in different namespaces, they will not conflict with each other. 

This is how I've thought through it, but I would not mind getting @dramich input on this as well.

ryansann: NOTE: this also lines up with how deduping is carried out [here](https://github.com/dramich/rancher/blob/2a59dbca0009e2846099e076ee914a56ea1f986f/pkg/agent/clean/binding.go#L231)

pkg/controllers/managementuser/rbac/handler_base.go
```diff
@@ -495,10 +495,11 @@ func (m *manager) ensureBindings(ns string, roles map[string]*v3.RoleTemplate, b
 
 func bindingParts(roleName, parentNsAndName string, subject rbacv1.Subject) (string, metav1.ObjectMeta, []rbacv1.Subject, rbacv1.RoleRef) {
 	crbKey := rbRoleSubjectKey(roleName, subject)
+	crbName := strings.Join([]string{"clusterrolebinding", pkgrbac.HashFromKey(crbKey)}, "-")
 	return crbKey,
 		metav1.ObjectMeta{
-			GenerateName: "clusterrolebinding-",
-			Labels:       map[string]string{rtbOwnerLabel: parentNsAndName},
+			Name:   crbName,
+			Labels: map[string]string{rtbOwnerLabel: parentNsAndName},
```
dramich, line 502: This is causing roleBindings to end up with the name `clusterrolebinding-`, this should also be addressed as part of this pr.

pkg/controllers/management/auth/manager.go
ryansann: I've implemented these changes, resolving

pkg/controllers/managementuser/rbac/handler_base.go
```diff
@@ -469,16 +469,17 @@ func (m *manager) ensureBindings(ns string, roles map[string]*v3.RoleTemplate, b
 	}
 
 	for key, rb := range desiredRBs {
-		logrus.Infof("Creating roleBinding %v", key)
 		switch roleBinding := rb.(type) {
 		case *rbacv1.RoleBinding:
+			logrus.Infof("Creating roleBinding %v", key)
```
prachidamle, line 474: change this to debug?

ryansann: It was Info before, so I just kept it at the same level

prachidamle (approved):

superseb (changes requested):
Can we add debug logging to the name generation parts so we dont have to code diving to find what is generated for what object?

superseb (dismissed):
I think debug logging would still be nice

superseb (approved):

prachidamle (approved):

Issue #30864, opened 2021-01-15
Google Cloud Platform UI to be able to work with Shared VPC


jambajaar:
Currently, Rancher UI is unable to create clusters when a project is a "service project" on a shared VPC.

In this scenario, the service project must use a network that's provided to it by the Host Project, which owns all organization level networks.

Current Rancher UI only lists "local networks" (local to the project), not offering a way to select "Networks shared with my project".

Rancher2 terraform provider is able to create clusters with this configuration. Being able to do so confirms the problem's scope is the UI.

Attached are GCP screens showing how their interface lists the relevant objects and a simple terraform file showing how the objects are configured to be sent to rancher.

fabrinator:
I'm curious for view the attached files. Can you post them here? I don't have access to the link that you provided.

Kind Regards

mtterry-qfree:
I'm running into this issue currently. we have a requirement that we have a shared VPC in which we deploy to. I'm looking into simply "importing" my gke cluster, but i'm sure that's not ideal.

cbron:
This can only be truly tested through the UI

prabalsharma:
I was able to create a public cluster on a `service project` using shared vpc from a `host project`, on 2.5-head UI

PR #31829, opened 2021-03-29
Add /meta/gkeSharedSubnets and /meta/gkeClusters routes

cmurphy:
Add /meta/gkeClusters route

Add a route to list GKE clusters for the purpose of selecting one to
import.

The function can list either regional or zonal clusters. It rejects the
request if neither region nor zone are provided, and also if both
region and zone are provided.

https://github.com/rancher/rancher/issues/31461

---

Add /meta/gkeSharedSubnets route

Add a proxy route /meta/gkeSharedSubnets for querying usable subnets
shared to a service project, where the project ID provided in the URL
query is the service project. This call looks up the host project for
the provided service project, and uses that to look up the subnets that
are shared with the service project. See the `gcloud` command line
reference example[1].

The result of this query will resemble the following:

```
$ curl -k -u $TOKEN "https://localhost:8443/meta/gkeSharedSubnets?cloudCredentialId=cattle-global-data:cc-abcde&projectId=service-project-id" | jq .
{
  "subnetworks": [
    {
      "ipCidrRange": "10.1.0.0/24",
      "network": "projects/vpc-host-309518/global/networks/vpc-host-network",
      "secondaryIpRanges": [
        {
          "ipCidrRange": "10.2.0.0/21",
          "rangeName": "pods",
          "status": "UNUSED"
        },
        {
          "ipCidrRange": "10.3.0.0/21",
          "rangeName": "services",
          "status": "UNUSED"
        }
      ],
      "subnetwork": "projects/vpc-host-309518/regions/us-west1/subnetworks/vpc-host-subnet"
    }
  ]
}
```

The UI needs to use the full path of the network and subnetwork fields
returned from this response, as well as the names and CIDR ranges of the
secondary IP ranges, to create a cluster using a shared VPC subnet.

If the project is not part of a shared VPC structure, the request will
succeed but return null. The caller needs to call /meta/gkeNetworks and
/meta/gkeSubnetworks to get the project's own list of networks and
subnetworks.

[1] https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#verify_usable_subnets

https://github.com/rancher/rancher/issues/30864

---

Notes for reviewers:

~1) I haven't fully tested this yet because I don't have access to a GCP project with VPCs shared to it~ This is now tested
1 edit) Changed /meta/gkeSharedNetworks to /meta/gkeSharedSubnets
2) This is based on @mjura's work in https://github.com/rancher/rancher/pull/31759 - I could make this PR directly into that one if that would make things easier
3) @cbron I confirmed service projects [can only have one host project](https://cloud.google.com/vpc/docs/shared-vpc#concepts_and_terminology) "You can create and use multiple host projects; however, each service project can only be attached to a single host project."
4) Now also adds /meta/gkeClusters

### DIFF
--- a/pkg/api/norman/customization/gke/handler.go
+++ b/pkg/api/norman/customization/gke/handler.go
@@ -111,6 +111,20 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 
 		}
 		writer.Write(serialized)
+	case "gkeClusters":
+		if serialized, errCode, err = listClusters(req.Context(), capa); err != nil {
+			logrus.Debugf("[gke-handler] error getting clusters: %v", err)
+			handleErr(writer, errCode, err)
+			return
+		}
+		writer.Write(serialized)
+	case "gkeSharedSubnets":
+		if serialized, errCode, err = listSharedSubnets(req.Context(), capa); err != nil {
+			logrus.Debugf("[gke-handler] error getting shared subnets: %v", err)
+			handleErr(writer, errCode, err)
+			return
+		}
+		writer.Write(serialized)
 	default:
 		handleErr(writer, httperror.NotFound.Status, fmt.Errorf("invalid endpoint %v", resourceType))
 	}
--- a/pkg/api/norman/customization/gke/listers.go
+++ b/pkg/api/norman/customization/gke/listers.go
@@ -174,6 +174,66 @@ func listZones(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
 	return encodeOutput(result)
 }
 
+func listClusters(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
+	if cap.Region == "" && cap.Zone == "" {
+		return nil, http.StatusBadRequest, fmt.Errorf("either region or zone is required")
+	}
+	if cap.Region != "" && cap.Zone != "" {
+		return nil, http.StatusBadRequest, fmt.Errorf("only one of region or zone can be provided")
+	}
+
+	var location string
+	if cap.Region != "" {
+		location = cap.Region
+	} else {
+		location = cap.Zone
+	}
+
+	client, err := getContainerServiceClient(ctx, cap.Credentials)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	parent := "projects/" + cap.ProjectID + "/locations/" + location
+	result, err := client.Projects.Locations.Clusters.List(parent).Do()
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	return encodeOutput(result)
+}
+
+func listSharedSubnets(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
+	computeClient, err := getComputeServiceClient(ctx, cap.Credentials)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+	containerClient, err := getContainerServiceClient(ctx, cap.Credentials)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	hostProject, err := computeClient.Projects.GetXpnHost(cap.ProjectID).Do()
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	// If there is no host project for this project, the fields in this returned object will be empty.
+	// In this case, we will return a null object indicating there are no subnets explicitly shared to this project.
+	// The caller will need to call /meta/gkeNetworks and /meta/gkeSubnetworks to get the project's own network and subnet list.
+	var result *container.ListUsableSubnetworksResponse
+	if hostProject.Name != "" {
+		parent := "projects/" + cap.ProjectID
+		filter := "networkProjectId=" + hostProject.Name
+		result, err = containerClient.Projects.Aggregated.UsableSubnetworks.List(parent).Filter(filter).Do()
+		if err != nil {
+			return nil, http.StatusInternalServerError, err
+		}
+	}
+
+	return encodeOutput(result)
+}
+
 func encodeOutput(result interface{}) ([]byte, int, error) {
 	data, err := json.Marshal(&result)
 	if err != nil {


pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,196 @@
...
+			util.ReturnHTTPError(writer, req, httperror.InvalidBodyContent.Status, err.Error())
+			return
+		}
+	} else {
+		util.ReturnHTTPError(writer, req, httperror.Unauthorized.Status, "cannot access Google API without credentials to authenticate")
+	}
+
+	var serialized []byte
+	var errCode int
+	var err error
```
thedadams, line 66: nit: remove these since each case is handled in its own scope.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -0,0 +1,196 @@
...
+	if err := access.ByID(h.generateAPIContext(req), &schema.Version, client.CloudCredentialType, credID, &accessCred); err != nil {
+		if apiError, ok := err.(*httperror.APIError); ok {
+			if apiError.Code.Status == httperror.PermissionDenied.Status || apiError.Code.Status == httperror.NotFound.Status {
+				return fmt.Errorf("cloud credential not found")
+			}
+		}
+		return err
+	}
+
+	cc, err := h.secretsLister.Get(namespace.GlobalNamespace, name)
```
thedadams, line 142: Should `ns` from above be used as the namespace here?

thedadams (changes requested):

cmurphy:
@thedadams sorry for the confusion, I based this pull request on top of https://github.com/rancher/rancher/pull/31759 and only intended to address shared VPC functionality in this one. Can you make your comments on @mjura's PR?

thedadams:
Sorry about that, @cmurphy. I made my comments on the other PR. This looks good otherwise.

rmweir (dismissed):
LGTM

thedadams (approved):

pkg/api/norman/customization/gke/listers.go
```diff
@@ -195,3 +195,36 @@ func listZones(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
 
 	return data, http.StatusOK, err
 }
+
+func listSharedNetworks(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
+	if cap.ProjectID == "" {
+		return nil, httperror.ServerError.Status, fmt.Errorf("projectId is required")
+	}
```
cbron, line 203: This might be moved to helper, will need to check that sort of thing when we rebase.

pkg/api/norman/customization/gke/listers.go
```diff
@@ -228,3 +229,41 @@ func listSharedNetworks(ctx context.Context, cap *Capabilities) ([]byte, int, er
...
+	} else {
+		location = cap.Zone
+	}
+
+	client, err := getContainerServiceClient(ctx, cap.Credentials)
+	if err != nil {
+		return nil, httperror.ServerError.Status, err
+	}
+
+	parent := "projects/" + cap.ProjectID + "/locations/" + location
```
cbron, line 257: So if region is present this list gets more specific ? Does Wes know that they can pass in either/both ?

cbron (commented):
Seems fine, but need to wait on base.

pkg/api/norman/customization/gke/listers.go
cmurphy: It's not "more specific", it's either/or. A cluster is either regional or zonal. It would not make sense to pass both in. However, I'm proposing that in case both are passed in, we default to looking up clusters for the region and ignoring the zone. The other option is to return an error if both region and zone are provided. @westlywright what do you think?

cbron: Lets just do region in the UI then, thats what eks does. We can leave API accepting either, and prefer region.

cmurphy: Sorry, no, I wasn't clear - I don't mean "either/or" as in it doesn't matter, I mean "either/or" as in it's either one or the other, not both. It would make no sense to GKE to only be able to select region in the UI. You need to be able to select one or the other. A cluster in region "us-west1" is not the same as a cluster in the zone "us-west1-a".

image

thedadams (dismissed):

cbron (approved):
lgtm as long as we are communicating all this to Wes.

pkg/api/norman/customization/gke/listers.go
cbron: Oh wait, these should be handled now before this point

cbron (changes requested):

pkg/api/norman/customization/gke/listers.go
cmurphy: Pretty sure this needs to be handled here, same as the others https://github.com/rancher/rancher/pull/31829/files#diff-18b89e110c42b6e252c7b898f1c84f058011c923d7af5f01125f4085acba8550R87

cbron: It should be handled inside getCloudCredential, before this point: https://github.com/rancher/rancher/blob/master/pkg/api/norman/customization/gke/handler.go#L144

cmurphy: Oops, fixed

cbron (approved):

pkg/api/norman/customization/gke/listers.go
```diff
@@ -182,3 +182,73 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
...
+			return nil, http.StatusInternalServerError, err
+		}
+	}
+
+	data, err := json.Marshal(&result)
+	if err != nil {
+		return data, http.StatusInternalServerError, err
+	}
+
+	return data, http.StatusOK, err
```
rmweir, line 253: nit: could just return nil since err would be nil here

rmweir (approved):
LGTM: 1 nit

thedadams (approved):

cbron (approved):

rmweir (approved):
LGTM

mjura (approved):

thedadams (approved):

Issue #31109, opened 2021-02-03
Unable to pass boolean values to Rancher app


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Add the rancher catalog to a rancher install - https://releases.rancher.com/server-charts/latest
2. Deploy rancher as an app to a downstream cluster. To hide the local cluster, pass the following two answers:
variable: extraEnv[0].name   value: CATTLE_HIDE_LOCAL_CLUSTER
variable: extraEnv[0].value   value: true

**Result:**
Deployment fails with error:
```
Failed to install app rancher. Error: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found t, error found in          #10       byte of ...|,"value":true}],"ima|..., bigger context ...|ue"},{"name":"CATTLE_HIDE_LOCAL_CLUSTER","value":true}],"image":"rancher/rancher:v2.5.2-rc4","imageP|...
```

**Other details that may be helpful:**
Related to:
https://github.com/rancher/rancher/issues/26088
https://github.com/rancher/rancher/issues/29325#issuecomment-719864450

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
- v2.4.13 (deploying Rancher v2.5.5 downstream using catalog app)
- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
k3s
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
AWS m5a.large (2vCPU, 8GB RAM)
- Kubernetes version (use `kubectl version`):

```
> kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.6", GitCommit:"d32e40e20d167e103faf894261614c5b45c44198", GitTreeState:"clean", BuildDate:"2020-05-20T13:16:24Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #31837, opened 2021-03-30
[master] Fix Helm changing app answers types

PennyScissors:
Problem:
When we install an app using Helm install, we pass all user provided answers as args to Helm's `--set` which changes will change a value of a quoted string `"true"` to a boolean `true` and break if the app expects a quoted string.

Solution:
Pass the answers that should keep the quoted string type as args to Helm's `--set-string`. Since both changing to the correct type and preserving the value as a quoted string are desirable outcomes depending on the situation, it should be up to the user to decide. For that reason the answers has been split into `answers` (keeps current behavior) and `answersSetString` (keeps its quoted string type). The expectation from the UI perspective is that the user will be able to choose which flag to use (A checkbox next to each answer text field maybe?), and the UI will add the value to the correct map.

Issues: https://github.com/rancher/rancher/issues/26088, https://github.com/rancher/rancher/issues/31109, https://github.com/rancher/rancher/issues/22532

### DIFF
--- a/pkg/api/norman/customization/app/app.go
+++ b/pkg/api/norman/customization/app/app.go
@@ -92,6 +92,15 @@ func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data
 	return nil
 }
 
+func copyAnswers(specAnswers map[string]string, answers interface{}) {
+	m, ok := answers.(map[string]interface{})
+	if ok {
+		for k, v := range m {
+			specAnswers[k] = convert.ToString(v)
+		}
+	}
+}
+
 func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiContext *types.APIContext) error {
 	var app projectv3.App
 	if err := access.ByID(apiContext, &projectschema.Version, projectv3.AppType, apiContext.ID, &app); err != nil {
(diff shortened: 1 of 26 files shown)


StrongMonkey (dismissed):
Changes look good to me. Should we use the `answersString` instead? It sounds more straightfoward.

PennyScissors:
> Changes look good to me. Should we use the `answersString` instead? It sounds more straightfoward.

The reason I went for `answersSetString` over `anwersString` is that I named it after the `--set-string` flag rather than the type because `answers` can still take strings so it might be a little misleading. I don't mind changing the name if you think `answersString` is still better though.

Jono-SUSE-Rancher:
@brendarearden, @dramich, @cbron, @PennyScissors - Even if the review for this is good to go, can we please hold off on merging this until we have an opportunity to further consult the UI team on their end, just to ensure proper bandwidth and availability.

PennyScissors:
@Jono-SUSE-Rancher I converted the PR to a draft to prevent it from getting merged in the meantime, but still allow getting it reviewed. Please let me know once it's good to merge.

brendarearden (approved):
LGTM!!

dramich (commented):
@PennyScissors Can you provide an example json blobs where this is used so we can see the difference from an API perspective? 

This will also require CLI and UI changes to support this

PennyScissors:
@dramich 

Example of new behavior:
User input in UI
```
"A"="1"  setString=false
"B"="2"  setString=false
"C"="3"  setString=true
```

Expected json:
```
{
  "answers": {
    "A":"1",
    "B":"2"
  },
  "answersSetString": {
    "C": "3"
  },
  ...
}
```

dramich (approved):

dramich:
@PennyScissors Can you writeup everywhere this needs to be tested by QA? App, multi-cluster apps, monitoring etc.

dramich (approved):

brendarearden (approved):

PR #31856, opened 2021-03-31
Move feature charts to 2.6

cbron:


aiyengar2 (commented):
LGTM, forgot that we need all the charts available in dev-v2.6 so I reverted the change in https://github.com/rancher/charts/commit/20fb8f030f5f87bbd46867ef7486559e40e1811e and pushed to dev-v2.6

PR #31880, opened 2021-04-01
Add workaround for networkmanager

rancher-max:
As documented in https://github.com/rancher/rke2/issues/786


sangeethah (dismissed):
LGMT

ShylajaDevadiga (approved):
LGTM

Oats87:
@rancher-max this is not going to work, it needs to use the `flannel` interface name, not the calico standard ones

Oats87 (changes requested):
I think there may need to be a network manager config reload forced as well, I'm not entirely sure how to do this, otherwise the changes won't take effect

Oats87:
I think adding the contents from: https://github.com/rancher/rke2/issues/786

and doing a subsequent restart of NetworkManager would work...

Oats87 (approved):
LGTM

Issue #31453, opened 2021-02-24
[2.5.x] auditlog can not be recorded


jianghang8421:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Set AUDIT_LEVEL=3 when install Rancher
2. Check the auditlog

**Result:**
Almost all requests cannot be recorded by auditlog.

**Other details that may be helpful:**

In Rancher2.5, almost all requests cannot be recorded by auditlog. When auditlog tries to write bytes to a file, it can not work as `json.Compact` in [Write function](https://github.com/rancher/rancher/blob/465bb3b80cb9f2d8c002fa9d7c37be20d38ed411/pkg/auth/audit/audit.go#L130) failed. The error logs are as follows: `compact audit log json failed: invalid character '\x1f' looking for beginning of value`

However, the auditlog can work well in 2.4. I suspect it is caused by the order of http middleware processing. The key point may be the gzip responsewriter. If the autilog writer is executed before gzip writer, there will be no problem, just like 2.4. However, in 2.5, I tried to print the logs and found that auditlog is executed after gzip. If the response body is compressed by gzip, auditlog cannot parse and write the bytes to the file.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5-87006de30625c046523a329b03ea2580db35fb57-head
- Installation option (single install/HA):single node

jiaqiluo:
I cannot reproduce the bug on rancher/rancher:v2.5-head  `v2.5-87006de30625c046523a329b03ea2580db35fb57-head`

Steps: 
- run Rancher with the following command 
```
sudo docker run -d 
--privileged 
--restart=unless-stopped 
-p 80:80 -p 443:443 
-v /var/log/rancher/auditlog:/var/log/auditlog 
-e AUDIT_LEVEL=3 
rancher/rancher:v2.5-head
```

-  do normal operations on the Rancher setup 


Results:
- the audit log file is created at `/var/log/rancher/auditlog` on the host and being updated consistently.
```
> ls /var/log/rancher/auditlog/
rancher-api-audit.log
```

- there is no error message related to `compact audit log json failed` in Rancher's logs 


@jianghang8421 Do you know how to trigger the bug?

jianghang8421:
[...]

PR #31881, opened 2021-04-01
[2.5] Try to gunzip resbody when writing auditlogs

niusmallnan:
Due to the change of the v2.5 handler chain, auditlog may get the data compressed by the gzip handler, which will cause the json conversion to fail.

https://github.com/rancher/rancher/issues/31453

### DIFF
--- a/pkg/auth/audit/audit.go
+++ b/pkg/auth/audit/audit.go
@@ -2,6 +2,7 @@ package audit
 
 import (
 	"bytes"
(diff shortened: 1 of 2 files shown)


niusmallnan:
close this PR as no reviewers.

niusmallnan:
reopen as this issue was still not fixed.

ibuildthecloud (approved):

Issue #31789, opened 2021-03-22
Rewrite GKEv2 cluster refresher


cbron:
In order to accommodate all the kev2 settings and clusters, we are generalizing the cluster refresher. This requires a re-testing of the EKS syncing mechanism. 
The cron job is getting retired and moved to a int seconds setting. A prior cronjob setting should be migrated to that setting.

cbron:
https://github.com/rancher/rancher/pull/31674

aaronRancher:
Tested and validated the following test cases on Rancher `master-head` `version 89c4d85` and `v2.5-head` `version bdec2f2` 
EKS operator version `v1.0.8-rc2`

Scenario 1:
From Rancher create a hosted EKS cluster.
After cluster becomes available, edit cluster from Rancher.
Validate the following field edits are seen in EKS console:
- SSH Key
- Instance Type
- User Data
- Resource Tag
- Node group Tag
Add node group and verify it is available in EKS console.

Scenario 2:
From EKS console, create a cluster.
From Rancher, import the EKS cluster.
From Rancher, scale the node group.
Verify node group is scaled from Rancher and EKS console.
From Rancher, add a node group. Verify node group is available from EKS console.

Scenario 3:
From Rancher create a hosted EKS cluster.
After cluster becomes available, edit cluster from EKS console.
Because hosted EKS clusters use launch templates, edit the launch template from EKS console. 
- SSH Key
- Instance Type
- User Data
- Resource Tag
- Node group Tag

From EKS console, add a cluster. Verify nodes appear on cluster node view. From Rancher, the node group created from EKS console will not appear on cluster edit. This is consistent with behavior from `v2.5.5`.

There are additional upgrade tests that are needed.

aaronRancher:
The syncing interval can be changed by running `kubectl edit setting eks-refresh` on the local cluster. The value field (not the default field) can be edited to change the sync timing. The refresher will produce logs similar to: `checking cluster [<cluster-name>] upstream state for changes`. As the refresher code is all in rancher, EKS-Operator does not need an update.

aaronRancher:
Tested and validated the following upgrade use cases:

**Scenario 1:**
Deploy Rancher `v2.5.7`.
Create a hosted EKS cluster.
Upgrade Rancher setup to `v2.5-head`
From Rancher, edit the EKS cluster.
Validate the following field edits are seen in EKS console:
- SSH Key
- Instance Type
- User Data
- Resource Tag
- Node group Tag

[...]

PR #31885, opened 2021-04-01
[release/v2.5] Add generic cloud config upstream refresher

mjura:
Refreshing and checking differences in upstream config spec for all
new cloud drivers can be done by generic service. Code base from existing
eksupstreamrefresher should be reused and prepared for adding new cloud drivers.

(cherry-pick from https://github.com/rancher/rancher/pull/31674)

issue: https://github.com/rancher/rancher/issues/31789


cbron (approved):
lgtm, we should have squashed to single commit on old PR, but it's ok to keep this one the same. Also make sure issue is linked, I added.

Issue #11496, opened 2018-02-20
Upstream Kubernetes Deprecation: Controller Manager & Scheduler show as Unhealthy in clusters that do not have component status for them (Azure, Tectonic among others)


tfiduccia:
**Rancher versions:** 2.0 master 2/20

**Steps to Reproduce:**
1.  Create a Azure AKE cluster
2.  Wait for it to activate and go to dashboard

**Results:**  The Controller Manager and Scheduler stay in unhealthy state.
![image](https://user-images.githubusercontent.com/11514927/36444885-c156ba80-1639-11e8-84a2-5872d5e1f8a8.png)

gz#11712
gz#12343

nathan-jenan-rancher:
This is a bug with Azure that we can't fix ourselves, here is the issue we opened with them:

https://github.com/Azure/AKS/issues/173

JasonvanBrackel:
This appears to be an upsteam kubernetes issue, that is affecting AKS by proxy affecting Rancher.

@nathan-jenan-rancher linked to this https://github.com/kubernetes/kubernetes/issues/19570 in the conversation on https://github.com/Azure/AKS/issues/173.  This is complicated by the fact that componentstatus is broken and likely be deprecated before it's fixed.

See https://github.com/kubernetes/features/issues/553
and
https://github.com/kubernetes/kubernetes/issues/18610

codedevote:
As far as Is understand, there is currently no way to create an AKS cluster using rancher that will actually work as expected due to this issue. Is there any workaround or advice on how to overcome this limitation?

JasonvanBrackel:
@codedevote These clusters are fully functional, the health reporting what is wrong.  If you were to create an AKS cluster from the Azure Portal and import it into rancher, this cluster would be completely functional, but the controller manager and scheduler will show unhealthy in Rancher.  Rancher can still manage workload on the cluster as designed.

hadifarnoud:
I'm not on Azure and still have the same issue

JasonvanBrackel:
It's a Kubernetes problem.  It's surfaced by any implementation that do not host the scheduler and the kubeapi-server on the same node.  The underlying code makes incorrect assumptions around this layout as it's calling localhost in areas where that's no longer a sound assumptions.  It doesn't surface on most implementation as most implementations use a traditional Kubernetes "master" node, but it's not a requirement.  Hetzner is likely architected similar to Azure in that manner to surface there error.

ablekh:
[...]

PR #31890, opened 2021-04-01
exclude some component status checks from some providers

paynejacob:
ISSUE: https://github.com/rancher/rancher/issues/11496

PROBLEM: 
Not all providers implement component status correctly, e.g. AKS shows the scheduler in an error state. We need to ignore some component statuses as they can be misleading.

SOLUTION:

Added a filter to the health syncer that does not return some component statues based on provider.

TESTING:

I was unable to get access to an AKS cluster but, I could filter components out of a k3os cluster by adding it to the exclusion map.


pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -85,6 +93,19 @@ func (h *HealthSyncer) getComponentStatus(cluster *v3.Cluster) error {
...
+			for ii := 0; ii < len(cluster.Status.ComponentStatuses); ii++ {
+				if cluster.Status.ComponentStatuses[ii].Name == excludedComponents[i] {
+					cluster.Status.ComponentStatuses = append(cluster.Status.ComponentStatuses[:ii], cluster.Status.ComponentStatuses[ii+1:]...)
+					break
+				}
+			}
+
+		}
+	}
+
```
PennyScissors, line 108: Hey, just took a quick look and I think you can probably simplify this more by doing something along these lines (untested).

```
func init() {
	excludedComponentMap[providers.AKS] = map[string]struct{}{"controller-manager": struct{}{}, "scheduler": struct{}{}}
}

if excludedComponents, ok := excludedComponentMap[cluster.Status.Driver]; ok {
    for i, status := range cluster.Status.ComponentStatuses {
        if _, ok := excludedComponents[status.Name]; ok {
            cluster.Status.ComponentStatuses = append(cluster.StatusComponentStatuses[:i], cluster.Status.ComponentStatuses[i+1:]...)
            break
        }
}
```

paynejacob: ill refactor it to use a map thanks!

paynejacob: @PennyScissors I looked into refactoring to using a map and remembered why I did it this way.  If you delete an index from an array you are traversing you will get an Index error because the length will have decreased.  While we do traverse the list multiple times we reset our length in the event of a delete.  Most clusters only have 1 - 3 statues so this  should not be a performance hit.

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -25,6 +27,14 @@ const (
...
 )
 
+var excludedComponentMap = map[string][]string{}
+
+func init() {
+	excludedComponentMap[providers.AKS] = []string{"controller-manager", "scheduler"}
+	excludedComponentMap["tencenttke"] = []string{"controller-manager", "scheduler"}
+	excludedComponentMap["linodelke"] = []string{"controller-manager", "scheduler"}
+}
+
```
dramich, line 37: The init here is not needed, the map can be declared and initialized in the var. See https://github.com/rancher/rancher/blob/master/pkg/controllers/managementuser/rbac/prtb_handler.go#L25 for an example.

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -85,6 +95,19 @@ func (h *HealthSyncer) getComponentStatus(cluster *v3.Cluster) error {
...
+		for i := 0; i < len(excludedComponents); i++ {
+			for ii := 0; ii < len(cluster.Status.ComponentStatuses); ii++ {
+				if cluster.Status.ComponentStatuses[ii].Name == excludedComponents[i] {
+					cluster.Status.ComponentStatuses = append(cluster.Status.ComponentStatuses[:ii], cluster.Status.ComponentStatuses[ii+1:]...)
+					break
+				}
+			}
+
+		}
+	}
```
dramich, line 109:
```suggestion
	if excludedComponents, ok := excludedComponentMap[cluster.Status.Driver]; ok {
		var newStatus []v32.ClusterComponentStatus
		for _, status := range cluster.Status.ComponentStatuses {
			if !slice.ContainsString(excludedComponents, status.Name) {
				newStatus = append(newStatus, status)
			}
		}
		cluster.Status.ComponentStatuses = newStatus
	}
```

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/rancher/kubernetes-provider-detector/providers"
+
```
dramich, line 10: Extra space here

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
dramich, line 109: This could be dropped and add the below at line 91, this should save us some processing/loops
```
for _, cs := range cses.Items {
		if excludedComponents, ok := excludedComponentMap[cluster.Status.Driver]; ok && !slice.ContainsString(excludedComponents, cs.Name) {
			clusterCS := convertToClusterComponentStatus(&cs)
			cluster.Status.ComponentStatuses = append(cluster.Status.ComponentStatuses, *clusterCS)
		}
	}
```

dramich (dismissed):

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -2,6 +2,7 @@ package healthsyncer
 
 import (
 	"context"
+	"github.com/rancher/norman/types/slice"
```
dramich, line 5: update the sorting of imports

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -78,13 +85,19 @@ func (h *HealthSyncer) getComponentStatus(cluster *v3.Cluster) error {
...
 	cluster.Status.ComponentStatuses = []v32.ClusterComponentStatus{}
+	excludedComponents, ok := excludedComponentMap[cluster.Status.Driver]
 	for _, cs := range cses.Items {
+		if ok && slice.ContainsString(excludedComponents, cs.Name) {
+			continue
+		}
 		clusterCS := convertToClusterComponentStatus(&cs)
 		cluster.Status.ComponentStatuses = append(cluster.Status.ComponentStatuses, *clusterCS)
 	}
+
```
dramich, line 96: Drop the extra lines

PennyScissors (dismissed):

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -78,13 +86,23 @@ func (h *HealthSyncer) getComponentStatus(cluster *v3.Cluster) error {
 		return condition.Error("ComponentStatsFetchingFailure", errors.Wrap(err, "Failed to communicate with API server"))
 	}
 	cluster.Status.ComponentStatuses = []v32.ClusterComponentStatus{}
+	clusterType := cluster.Status.Provider // the provider detector is more accurate but we can fall back on the driver is some cases
```
PennyScissors, line 89: typo in the comment

PennyScissors (dismissed):

paynejacob:
These changes will be switched to master for 2.6 in a separate pr.

PR #31898, opened 2021-04-02
Fix for lastRefreshTime inclusterupstreamrefresher

mjura:
For the first time when clusterLastRefreshTime annotation doesn't
exist yet, don't report error about lastRefreshTime.


mjura:
This will be continued in https://github.com/rancher/rancher/pull/31901

PR #31939, opened 2021-04-06
[v2.5] Fix GKE integration with clusterupstreamrefresher

mjura:
Add missing gke-refresh setting and missing bits for GKE
cloud provider in clusterupstreamrefresher.

(cherry picked from commit d3959066923992147a448a132f9a3849b42e6f37)

### DIFF
--- a/pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
+++ b/pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
@@ -23,6 +23,7 @@ import (
 const (
 	noKEv2Provider         = "none"
 	clusterLastRefreshTime = "clusters.management.cattle.io/ke-last-refresh"
+	refreshSettingFormat   = "%s-refresh"
 )
 
 type clusterRefreshController struct {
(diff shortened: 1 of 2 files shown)


cmurphy (approved):

thedadams (dismissed):

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -120,6 +126,8 @@ func getProviderRefreshInterval(provider string) (time.Duration, error) {
 	switch provider {
 	case apimgmtv3.ClusterDriverEKS:
 		refreshInterval = settings.EKSUpstreamRefresh.GetInt()
+	case apimgmtv3.ClusterDriverGKE:
```
rmweir, line 129: I think we should go with the suggestion here instead of having a switch in this function: https://github.com/rancher/rancher/pull/31939/files#diff-ec482fc6c00713153755d85cb92dd1a3f9c76b416d548446a5f5d5728d4afec0R129-R130

rmweir (changes requested):

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
cbron: @rmweir the link points back here

mjura: But what if GKE and EKS settings will have different refresh times ?

cbron: He is saying to do `getSettingById(fmt.sprintf("%s-refresh"), provider))` so it gets auto detected and pulled for each provider.

mjura: according to our discussion, I have included suggestions

thedadams (approved):

rmweir (approved):
LGTM

cbron (approved):

cbron:
backfix for: https://github.com/rancher/rancher/issues/31011

Issue #31264, opened 2021-02-11
Feature charts should have rancher min/max version


cbron:
We can do in similar way to system-charts: https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-monitoring/v0.1.0/questions.yml

Note that internal apps such as fleet and operator have their own system, see: https://github.com/rancher/rancher/issues/30934

brendarearden:
https://github.com/rancher/rancher/blob/8476503a1a631c25ead03293736fbfb85d69c906/pkg/catalogv2/content/content.go#L212 This appears to already be added here, as part of this commit https://github.com/rancher/rancher/commit/fa26506587a029a7774c5bb3756772c3e95175d4. I am working on adding tests, and will update documentation and chart annotations as necessary.

cbron:
@brendarearden to help QA, can you setup one feature chart to have this in dev and then revert after they test. Or find some other way to QA.

brendarearden:
I have an example branch that can be used as a repository in rancher to test against: 
Type: Github Repo
Url: https://github.com/brendarearden/charts
Branch: testing-branch
Chart with the annotation: rancher-istio
The test is: depending on the rancher version that you are on, the chart should show or not show up available in the drop down. 
Set-up: have a rancher setup that you have set the rancher-server version to a release version. 

If you would like, you can fork the branch or make your own - the test does not need valid charts, it just needs a valid index.yaml that lists charts with the `atalog.cattle.io/rancher-version` annotation on it.

jiaqiluo:
The enhancement is validated on rancher:v2.5-head 9e412a346

The following repo is used:
```
Url: https://github.com/brendarearden/charts
Branch: testing-branch
Chart with the annotation: rancher-istio
```

The annotation on each version in the [index.yaml](https://github.com/brendarearden/charts/blob/testing-branch/index.yaml) file
Istio Version | Annotation  
------------- | ----------  
1.7.700 | ~2.6.0               
1.7.600 | <2.6.0                
1.7.500 | >= 2.5.0-alpha < 2.6.0 
1.7.400 | >= 2.4.1-alpha < 2.5.8 
1.7.103 |   
1.7.102 | >= 2.4.1 < 2.5.7
1.7.101 | >= 2.4.3



The following command is used to change the version of Rancher by setting the environment variable `CATTLE_SERVER_VERSION`
```
sudo docker run -d --privileged --restart=unless-stopped -p 80:80 -p 443:443 \
[...]

PR #31959, opened 2021-04-07
Add tests for min/max rancher version chart annotation

brendarearden:
**Problem**
Catalog V2 needs to have a way to not show certain charts for a min, max or min/max version of rancher. 

**Solution** 
As part of completing this issue, I have created tests to validate the various semver comparison strings that could be used to filter or not filter out a chart from being available for a specific rancher version. 

As part of https://github.com/rancher/rancher/commit/fa26506587a029a7774c5bb3756772c3e95175d4, a chart annotation  was added that can be used to set a min, max, or min/max version of rancher. The version comparison string is taken from the chart annotation and is used to determine if the current rancher server version meets the criteria.  https://github.com/rancher/rancher/commit/fa26506587a029a7774c5bb3756772c3e95175d4

**Issue** 
https://github.com/rancher/rancher/issues/31264

### DIFF
--- a/pkg/catalogv2/content/content_test.go
+++ b/pkg/catalogv2/content/content_test.go
@@ -1,0 +1,204 @@
+package content
+
+import (
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/rancher/rancher/pkg/settings"
+	"github.com/stretchr/testify/assert"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/repo"
+)
+
+func TestFilterReleases(t *testing.T) {
+	tests := []struct {
+		testName               string
+		chartVersionAnnotation string
+		rancherVersion         string
+		expectedPass           bool
+	}{
+		{
+			"rancher version in range comparison with `>= <`style comparison",
+			">= 2.5.0-alpha3 <2.6.0",
+			"v2.5.0+123", //SemVer comparisons using constraints without a prerelease comparator will skip prerelease versions
+			true,
+		},
+		{
+			"rancher version in range comparison with `> <`style comparison",
+			">2.5.0 <2.6.0",
+			"v2.5.7",
+			true,
+		},
+		{
+			"rancher version in range comparison with `> <=`style comparison",
+			">2.5.0-rc1 <=2.6.0-0",
+			"v2.5.0-rc2", //SemVer comparisons using constraints with prerelease will be evaluated using an ASCII sort order, per the spec
+			true,
+		},
+		{
+			"rancher version in range comparison with `>= <=`style comparison",
+			">=2.5.0-alpha2 <=2.6.0",
+			"v2.5.0", //Pre-release versions would be skipped with this comparison
+			true,
+		},
+		{
+			"rancher version in range comparison with `~` style comparison",
+			"~2.5.x", //equivalent to ">= 2.5.0 < 2.6.0"
+			"v2.5.7",
+			true,
+		},
+		{
+			"rancher version in range comparison with `<` style comparison",
+			"<2.6.001",
+			"v2.6.000",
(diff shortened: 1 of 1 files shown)


StrongMonkey (approved):
:100:

rmweir (commented):
Should a case be added for a constraint like "^2.5.1"? Also do hyphenated ranges and or ranges work? This is what I'm looking at for ideas: https://devhints.io/semver

PennyScissors:
Should this include cases for RCs and non dot separated pre-release identifiers to account for how use `-rc1` and `-alpha1` rather than `-rc.1` and `-alpha.1`?

brendarearden:
@rmweir I was using this as a reference http://masterminds.github.io/sprig/semver.html - looks like all of our cases could be covered with basic comparisons, but since I did test the tilde range comparison it does make sense to add caret and hyphen range tests as well. I will get those updated. 

@PennyScissors I will add those in! Side note, I don't think alpha1 is technically the same semver as a pre-release alpha.1? I don't think its technically a valid semver pre-release tag. I don't think it ultimately matters https://jubianchi.github.io/semver-check/#/%3E%3D%201.0.0-rc5%20%3C%201.0.0-rc9/1.0.0-rc7<< because we see weird behavior like this when it comes to pre-release comparisons so I think it makes sense to test what we use.

rmweir:
@brendarearden may also want to try a version that has metadata added to it with "+".

rmweir (approved):
LGTM, great job on tests.

aiyengar2 (approved):
LGTM

pkg/catalogv2/content/content_test.go
```diff
@@ -0,0 +1,204 @@
...
+			"2.5 - 2.6.3", //equivalent to ">= 2.5 <= 2.6.3"
+			"v2.4.9",
+			false,
+		},
+		{
+			"rancher version out range comparison with `^` style comparison",
+			"^2.7.5", //equivalent to ">= 2.7.5 < 3.0.0"
+			"v3.0.0",
+			false,
+		},
```
cbron, line 152: Out of curiosity I wonder about 2 cases: 

* One that is 000, the invalid semver, to see how it reacts. like `"2.5.000 - 2.6.005"`
* A dash metadata: '>=v1.14.0-r0'

Neither of those would likely end up as rancher version but wonder if they break this.

pkg/catalogv2/content/content_test.go
```diff
@@ -0,0 +1,204 @@
...
+		testName               string
+		chartVersionAnnotation string
+		rancherVersion         string
+		expectedPass           bool
+	}{
+		{
+			"rancher version in range comparison with `>= <`style comparison",
+			">= 2.5.0-alpha3 <2.6.0",
+			"2.5.0-alpha2",
+			true,
```
cbron, line 25: Shouldn't this be false ? `2.5.0-alpha3 - 2.6.0`. Either I'm mistaken or this is passing incorrectly.

brendarearden: Good catch and you are correct. I had missed the `v` in the version, so it was not getting past the ```	if !settings.IsRelease() {
		return index
	}``` check so it was never actually evaluating the comparison. I have made changes to correct this.

pkg/catalogv2/content/content_test.go
brendarearden: I have added a few random scenarios for `000` and for random meta data `r8` - its all evaluated with ASCII sort order - https://github.com/Masterminds/semver/blob/49c09bfed6adcffa16482ddc5e5588cffff9883a/constraints_test.go#L262 << here are additional tests from the library. These are the tests for the specific methods we are calling to filter out our charts.

cbron (approved):
lgtm

PR #31975, opened 2021-04-07
Add k3sconfig validation

rmweir:
**Problem:**
 Prior, it could be
set which is invalid because it is expected it will be initialized by a controller when it is detected that a cluster is a k3s cluster and the cluster is not embedded. This made it possible to set the k3sConfig and attempt upgrade an embedded k3s cluster, which is not possible.

**Solution:**
Validation has been added to prevent a k3sConfig being set on a cluster that does not have one set already.

**Issue:**
https://github.com/rancher/rancher/issues/30977

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -248,6 +248,12 @@ func (v *Validator) validateK3sBasedVersionUpgrade(request *types.APIContext, sp
 		return err
 	}
 
+	if isK3s && cluster.Spec.K3sConfig == nil {
+		// prevents embedded cluster from have k3sConfig set. Embedded cluster cannot be upgraded. Non-embedded
+		// clusters' config will be set my controller.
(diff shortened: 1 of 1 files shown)


pkg/api/norman/customization/cluster/validator.go
```diff
@@ -248,6 +248,10 @@ func (v *Validator) validateK3sBasedVersionUpgrade(request *types.APIContext, sp
 		return err
 	}
 
+	if (isK3s && cluster.Spec.K3sConfig == nil) || (isrke2 && cluster.Spec.Rke2Config == nil) {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, "k3sConfig cannot be changed from nil")
```
thedadams, line 252: Should we separate these checks so the error message corresponds to the field that was changed?

rmweir: Actually, embedded cluster is k3s only so I will remove rke check.

cbron (approved):
lgtm, I don't think this can affect normal k3s clusters. Don't forget to forwardport.

thedadams (approved):

PR #32018, opened 2021-04-09
Take apiGroups into account when processing auth objects

dramich:
Problem:
apiGroups are not taken into account when creating downstream k8s auth
objects. This causes rancher to fail comparisons and have incorrect
permissions

Solution:
Update to use the apiGroup when creating/validating roles and
clusterRoles

### DIFF
--- a/pkg/controllers/management/auth/manager.go
+++ b/pkg/controllers/management/auth/manager.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"fmt"
 	"reflect"
+	"sort"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/objectclient"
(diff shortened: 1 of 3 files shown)


pkg/controllers/management/auth/manager.go
```diff
@@ -713,17 +714,19 @@ func checkGroup(apiGroup string, rule v1.PolicyRule) bool {
...
 	if role, err := m.rLister.Get(namespace, rt.Name); err == nil && role != nil {
 		newRole := role.DeepCopy()
 		for resource, newVerbs := range resourceToVerbs {
-			currentVerbs := map[string]bool{}
+			currentVerbs := map[string]string{}
 			for _, rule := range role.Rules {
 				if slice.ContainsString(rule.Resources, resource) {
 					for _, v := range rule.Verbs {
-						currentVerbs[v] = true
+						if rule.APIGroups[0] == newVerbs[v] {
```
nickgerace, line 727: This is because we only want to evaluate the new verb against the single rule, right?

nickgerace: Resolved by offline convo LGTM

prachidamle: I had question on the same line - do we always have only 1 APIGroup entry under the rule?

dramich: Yes because these are based off our tole templates.

dramich: The long term goal is to refactor this to use `[]rbacv1.PolicyRule` everywhere which will get rid of a lot of this complexity such as this.

prachidamle (approved):
LGTM

Issue #31964, opened 2021-04-07
Not able to access cluster explorer for downstream clusters


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):**  bug

**Steps to reproduce (least amount of steps as possible):**

- On master-head commit id: `077fbf6`
- Deploy a downstream cluster
- when the cluster is up and Active, navigate to Cluster explorer of this downstream cluster
- Request: `https://rancher-server/k8s/clusters/c-56zbb/v1/schemas`
- Response: `failed to find Session for client steve-cluster-c-56zbb`

```
HTTP Error 500: Internal Server Error
from /k8s/clusters/c-56zbb/v1/schemas
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `077fbf6`
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): DO RKE
- Kubernetes version (use `kubectl version`):

```
1.20 (Any)
```

izaac:
Rancher version: `master-head (04/12/2021)` _b202f20_
`500` when accessing the Dashboard on a downstream RKE custom cluster.

deniseschannon:
@sowmyav27 Can you test this now?

sangeethah:
Tested with master-head - `2d7fa0eab3e89d3ca`

Issue is still reproducible.

anupama2501:
On master-head [ebad1df](url) 
Issue is still seen with internal error 500

sowmyav27:
**On master-head commit id: `360edae`**

This is fixed, however, it takes ~2-3 minutes for the cluster explorer to come up active, else it fails with 500 Internal server error. Logging another bug to track it and closing this one.

PR #32022, opened 2021-04-09
Don't let the version get longer than 63 chars

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/31964

Issue #32052, opened 2021-04-12
Remove Docker 1.10 Cli from Agent


luthermonson:
per request of @ibuildthecloud and some feedback we received from docker, we need to stop abusing the docker 1.10 client binaries.

kravciak:
Works fine on master-head (66c31dd) and 2.5-head (e9c4413)
Docker is removed from linux agent container, windows containers still have it (as expected)

I am not creating separate ticket for 2.5-head since it will be probably reverted.

PR #32053, opened 2021-04-12
[v2.5] remove docker 1.10 cli from agent

luthermonson:
We are abusing the downloads of the docker 1.10 cli as we only use it in one scenario which is basically deprecated. Basically this was a fix for non-shared mounted root directories. We are more or less convinced this isn't a thing any more so we're dropping this shared-mount.sh fix as it was the only thing actually using the docker cli in the agent.

To confirm your linux distro is fine you do a quick 

```
findmnt -o TARGET,PROPAGATION /
```

and you should see

```
TARGET PROPAGATION
/      shared
```

https://github.com/rancher/rancher/issues/32052

### DIFF
--- a/package/share-root.sh
+++ b/package/share-root.sh
@@ -2,7 +2,4 @@
 set -x
 
 trap 'exit 0' SIGTERM
-ID=$(grep :devices: /proc/self/cgroup | head -n1 | awk -F/ '{print $NF}' | sed -e 's/docker-\(.*\)\.scope/\1/')
 bash -c "$1"
-
-docker kill --signal=SIGTERM $ID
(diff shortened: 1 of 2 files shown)


luthermonson:
https://github.com/rancher/rancher/issues/32052

StrongMonkey (dismissed):
:100:

package/share-root.sh
```diff
@@ -4,5 +4,3 @@ set -x
 trap 'exit 0' SIGTERM
 ID=$(grep :devices: /proc/self/cgroup | head -n1 | awk -F/ '{print $NF}' | sed -e 's/docker-\(.*\)\.scope/\1/')
 bash -c "$1"
-
-docker kill --signal=SIGTERM $ID
```
StrongMonkey, line 8: If we are sure that the share-mnt process exit with 0 the I am good with the change.

Issue #32065, opened 2021-04-12
[GKE v2] Deleting registered clusters in Rancher may delete the cluster from GKE


cmurphy:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):**

1. Register an existing GKE cluster in Rancher.
2. Update an updatable field in the cluster in GKE, such as the kubernetes version.
3. After the update is complete in GKE, wait up to five minutes for the change to be reflected in Rancher.
4. Delete the cluster in Rancher.

**Result:**

GKE starts deleting the cluster.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5-head 6f71d4c7346c63a60c671d59185e1595fe826ac7
- Installation option (single install/HA):

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

cbron:
Can this move to test ?

cmurphy:
done

prabalsharma:
passed on v2.5-4e70ed29ccc5de7c61516ba0367c7c4904989b56-head

prabalsharma:
added to https://github.com/rancher/rancher/issues/32140

PR #32066, opened 2021-04-12
[v2.5] Add Imported field to GKE upstream spec builder

cmurphy:
Without this patch, if a GKE cluster is imported to Rancher, and then a
change from GKE is synced back to the cluster in Rancher, "imported"
will flip back to false, and the guard in gke-operator that prevents
sending a DELETE request to imported clusters that are removed from
Rancher gets thwarted. This change fixes the upstream spec builder
function in the cluster refresher to always use the config spec value
for the Imported attribute.

https://github.com/rancher/rancher/issues/32065


cmurphy:
forward-port: https://github.com/rancher/rancher/pull/32069

thedadams (approved):

cbron (approved):

Issue #32087, opened 2021-04-13
GKEv2 cluster in error state in GCE console, switches between Provisioning state and Error state in Rancher.


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a hosted GKEV2 cluster 
- The cluster fails to provision in the GKEv2 console. 
- In rancher, the cluster is seen switching between Provisioning and Error state.
- Error in GKE operator:

```
time="2021-04-13T01:53:33Z" level=error msg="error syncing 'cattle-global-data/c-wnhn7': handler gke-controller: creation failed for cluster sowmya-gke-restore, requeuing"
time="2021-04-13T02:10:14Z" level=error msg="error syncing 'cattle-global-data/c-wnhn7': handler gke-controller: creation failed for cluster sowmya-gke-restore, requeuing"
```

**Expected Result:**
Can this error be better handled instead of continuous switching between "Provisioning" state and "Error" state

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5-head commit id: `6bb6c97`
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Hosted GKE v2

sowmyav27:
**On 2.5-head commit id: `41f173ae9` and on master-head - commit id: `1c3629fbe`**

- Deploy a cluster - hosted GKE in rancher
- Deploy another cluster using the same CIDR range for the pods
- The cluster is in the error state. There is no flipping between statuses 

<img width="322" alt="Screen Shot 2021-04-15 at 1 24 37 PM" src="https://user-images.githubusercontent.com/26032343/114934012-38768200-9dee-11eb-9bc3-5ab208730a64.png">

PR #32099, opened 2021-04-13
[v2.5] Do not set unknown condition for creating cluster

cmurphy:
Without this change, if a GKE cluster creation fails for some GKE-side
reason, the state of the cluster object will appear to rapidly flip
between provisioning and failed, and the Rancher logs will be filled
with a constant stream of messages:

  2021/04/13 10:15:43 [ERROR] error syncing 'c-tcc2p': handler gke-operator-controller: failed setting cluster [] condition Provisioned false with message: creation failed for cluster cluster-name, requeuing

This is a confusing UI behavior and the volume of logs emitted has the
potential to generate large log files unexpectedly quickly.

This change removes the initial status update during the "creating"
phase. The Provisioning condition will be set a few lines later to
either Unknown or False, so the initial status update request is
superfluous. Removing it prevents a case where the cluster object might
be have been modified and the condition status cannot be set, which
causes the onClusterChange handler to error out without calling
`e.ClusterEnqueueAfter(cluster.Name, enqueueTime)` and so causes the the
handler to re-run repeatedly without a delay.

The problem was observed for GKE clusters but this fix applies to both
EKS and GKE.

This change also fixes the SetUnknown/SetFalse/SetTrue helper functions
to return the update error if there is one, instead of returning an
error string with the context of the error but not the error itself, so
that when update conflicts happen it's no longer hidden.

https://github.com/rancher/rancher/issues/32087

### DIFF
--- a/pkg/controllers/management/gke/gke_cluster_handler.go
+++ b/pkg/controllers/management/gke/gke_cluster_handler.go
@@ -153,12 +153,6 @@ func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 
 	switch phase {
 	case "creating":
-		// set provisioning to unknown
-		cluster, err = e.SetUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "")
-		if err != nil {
-			return cluster, err
-		}
-
 		if cluster.Status.GKEStatus.UpstreamSpec == nil {
 			cluster, err = e.setInitialUpstreamSpec(cluster)
 			if err != nil {
--- a/pkg/controllers/management/clusteroperator/utils.go
+++ b/pkg/controllers/management/clusteroperator/utils.go
@@ -147,7 +147,7 @@ func (e *OperatorController) SetUnknown(cluster *mgmtv3.Cluster, condition condi
 	var err error
 	cluster, err = e.ClusterClient.Update(cluster)
 	if err != nil {
-		return cluster, fmt.Errorf("failed setting cluster [%s] condition %s unknown with message: %s", cluster.Name, condition, message)
+		return cluster, err
 	}
 	return cluster, nil
 }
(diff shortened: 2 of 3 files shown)


cmurphy:
Another option to solve this is to add retry loops to `SetUnknown`, `SetFalse`, and `SetTrue`

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -154,9 +154,11 @@ func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
 		// set provisioning to unknown
-		cluster, err = e.SetUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "")
-		if err != nil {
-			return cluster, err
+		if failureMessage == "" {
+			cluster, err = e.SetUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "")
+			if err != nil {
+				return cluster, err
+			}
 		}
```
thedadams, line 162: Not sure if this is needed because the same call is also made below. @rmweir: What are your thoughts?

cmurphy: I'm not sure why it's here except that it's the same in EKS

cmurphy:
I went ahead and removed the line since I don't see any reason to keep it. If I'm mistaken I can add it back.

If this is the right direction, should I do the same for EKS?

dramich:
If we have the same pattern in EKS that should be updated as well

cmurphy:
> If we have the same pattern in EKS that should be updated as well

Done, although I don't have access to EKS to validate it

dramich (approved):

rmweir (approved):
LGTM

cmurphy:
Forward port: https://github.com/rancher/rancher/pull/32133

PR #32109, opened 2021-04-13
[release/v2.5] Allow PNI to be enabled for additional RKE Cluster Network Providers

ryansann:
Issue: 
- https://github.com/rancher/rancher/issues/31338

For information on testing, [go here](https://github.com/rancher/rancher/issues/31338#issuecomment-819057626)

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -9,10 +9,6 @@ import (
 	"strings"
 	"sync"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
-	rketypes "github.com/rancher/rke/types"
-
 	"github.com/blang/semver"
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/api/access"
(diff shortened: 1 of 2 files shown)


pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -716,9 +713,6 @@ func validateNetworkFlag(data map[string]interface{}, create bool) error {
 			}
 			return fmt.Errorf("enableNetworkPolicy should be false for non-RKE clusters")
 		}
-		if plugin != "canal" {
```
StrongMonkey, line 719: Do we not want to run validation so that if user enable PNI with flannel, we can raise an error?

ryansann: There are 3 options for Flannel:
1. Disable in UI and backend
2. Disable in UI and not backend
3. Allow in UI and backend

I've opted for `2` here per some conversations I've had around this. With this approach, it's still technically possible to enable PNI for Flannel if you manually set `enable_network_policy: true` in the RKE Config. Enabling NetworkPolicy for a CNI that doesn't support it shouldn't have any adverse effects, the policies just don't get enforced.

ryansann: UI Issue [is here](https://github.com/rancher/rancher/issues/32106) for reference @StrongMonkey

kinarashah (dismissed):

kinarashah (approved):

StrongMonkey (approved):

Issue #31568, opened 2021-03-03
Update rancher-images.txt generator to handle multiple chart version


PennyScissors:
Problem:
Currently, the rancher-images.txt generator extracts images from the latest version of each chart. That logic doesn't work for monitoring v1 because in some cases we may need to update an old chart to add support for a k8s version, which results in the images missing from the rancher-images.txt.

Solution:
Update the rancher-images.txt generator to extract images from charts in the `rancher/system-charts` repository based on the `rancher_min_version` stated in the charts `questions.yaml`. E.g. if we are releasing Rancher v2.5.6, images in charts with `rancher_min_version: v2.5.6-rc0` and over should be included in the rancher-images.txt

Notes:
This change will be applicable to all charts at one point, so it would be good to take that into consideration when implementing a fix for it.

cbron:
cc @rmweir this will affect your stuff

rmweir:
@cbron @PennyScissors That's true. I would wait for https://github.com/rancher/rancher/pull/31492 to get merged before working on this.

cbron:
Since this got bumped to 2.6, do we need to do anything special for monitoring in 2.5.8 ?

PennyScissors:
Closing in favor of this [issue](https://github.com/rancher/rancher/issues/33037)

PR #32138, opened 2021-04-14
[master] Update rancher-images generator to support system charts min/max

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/31568

Some definitions before the explanation to make sure it's clear
chart = rancher-monitoring, rancher-logging, etc
version = rancher-monitoring/0.1.0, rancher-logging/v0.1.0

This PR adds support to the rancher-images.txt generator to pick images in system charts based on the `rancher_min_version` and `rancher_max_version` set in the questions.yaml.

Old behavior:
Pick images from the latest version of each chart only

New behavior:
Pick images from the latest version of each chart regardless of min/max and rancherVersion.
If the rancherVersion is a valid semver, then pick images of the remaining versions based on their min/max. A questions.yaml with a `rancher_min_version` is now required, whereas `rancher_max_version` remains optional. We change the rancherVersion to 2.6.99 (set somewhere in kdm) when it runs as dev or master, so the only case this doesn't execute is when rancherVersion is an empty string or a commit hash.

Testing:
Ran a local drone build with tag v2.5.8-rc4 before and after my changes (including changes from [this PR](https://github.com/rancher/system-charts/pull/444) to update missing min/max in system charts): [diff here](https://www.diffchecker.com/Tgukkypq) (expires in 30 days from 4/15/2021). Images added are from rancher-monitoring v1 v0.1.5

Commit `Update rancher images generator logging`: Changes in this commit can be removed if you as a reviewer don't like it. I added a log a while back so that we could see in the drone's logs which images are being added to the file, but this is not helpful because we are logging them right before we write to the rancher-images.txt file and you could just check the file. The new log will print the images found in each chart and subchart, and the path to the `values.yaml` where they were found which I believe to be a lot more helpful when debugging issues related to missing images.
Old log example:
```
2021/04/12 17:29:47 Image: rancher/pipeline-jenkins-server:v0.1.4
2021/04/12 17:29:47 Image: rancher/pipeline-tools:v0.1.15
2021/04/12 17:29:47 Image: rancher/prom-alertmanager:v0.21.0
```
New log example:
```
2021/04/06 15:23:52 Images found in testbuild/system-charts/charts/rancher-external-dns/v0.1.2/values.yaml: [rancher/kubernetes-external-dns:v0.7.3]
2021/04/06 15:23:53 Images found in testbuild/system-charts/charts/rancher-gatekeeper-operator/0.1.0/values.yaml: [rancher/istio-kubectl:1.4.6 rancher/opa-gatekeeper:v3.1.0-beta.7]
2021/04/06 15:23:55 Images found in testbuild/system-charts/charts/rancher-gke-operator/1.0.100/values.yaml: [rancher/gke-operator:v1.0.1-rc4]
2021/04/06 15:23:55 Images found in testbuild/system-charts/charts/rancher-istio/1.5.901/charts/certmanager/values.yaml: []
```

### DIFF
--- a/pkg/api/norman/customization/kontainerdriver/actionhandler.go
+++ b/pkg/api/norman/customization/kontainerdriver/actionhandler.go
@@ -158,16 +158,17 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
 	}
 	systemCatalogHash := helmlib.CatalogSHA256Hash(systemCatalog)
 	systemCatalogChartPath := filepath.Join(helmlib.CatalogCache, systemCatalogHash)
+	rancherVersion := kd.GetRancherVersion()
 
 	var targetImages []string
 	switch apiContext.ID {
 	case linuxImages:
-		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", []string{}, []string{}, rkeSysImages, image.Linux)
+		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", rancherVersion, []string{}, []string{}, rkeSysImages, image.Linux)
 		if err != nil {
 			return httperror.WrapAPIError(err, httperror.ServerError, "error getting image list for linux platform")
 		}
 	case windowsImages:
-		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", []string{}, []string{}, rkeSysImages, image.Windows)
+		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", rancherVersion, []string{}, []string{}, rkeSysImages, image.Windows)
 		if err != nil {
 			return httperror.WrapAPIError(err, httperror.ServerError, "error getting image list for windows platform")
 		}
--- a/pkg/image/export/main.go
+++ b/pkg/image/export/main.go
@@ -91,12 +91,12 @@ func run(systemChartPath, chartPath string, imagesFromArgs []string) error {
 
 	k3sUpgradeImages := getK3sUpgradeImages(rancherVersion, data.K3S)
 
-	targetImages, targetImagesAndSources, err := img.GetImages(systemChartPath, chartPath, k3sUpgradeImages, imagesFromArgs, linuxInfo.RKESystemImages, img.Linux)
+	targetImages, targetImagesAndSources, err := img.GetImages(systemChartPath, chartPath, rancherVersion, k3sUpgradeImages, imagesFromArgs, linuxInfo.RKESystemImages, img.Linux)
 	if err != nil {
 		return err
 	}
 
-	targetWindowsImages, targetWindowsImagesAndSources, err := img.GetImages(systemChartPath, chartPath, []string{}, []string{getWindowsAgentImage()}, windowsInfo.RKESystemImages, img.Windows)
+	targetWindowsImages, targetWindowsImagesAndSources, err := img.GetImages(systemChartPath, chartPath, rancherVersion, []string{}, []string{getWindowsAgentImage()}, windowsInfo.RKESystemImages, img.Windows)
 	if err != nil {
 		return err
 	}
--- a/pkg/image/export/main.go
+++ b/pkg/image/export/main.go
@@ -226,8 +226,6 @@ func imagesText(arch string, targetImages []string) error {
 		if err != nil {
 			return err
 		}
-
-		log.Println("Image:", image)
 		fmt.Fprintln(save, image)
 	}
 
--- a/pkg/image/resolve.go
+++ b/pkg/image/resolve.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/types/convert"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
@@ -23,14 +24,10 @@ import (
 
 type OSType int
 
-type chart struct {
-	dir     string
-	version string
-}
-
 const (
 	Linux OSType = iota
 	Windows
+	SystemChartsRepoDir = "build/system-charts"
 )
 
 func Resolve(image string) string {
@@ -50,8 +47,8 @@ func ResolveWithCluster(image string, cluster *v3.Cluster) string {
 	return image
 }
 
-func getChartAndVersion(path string) (map[string]chart, error) {
-	rtn := map[string]chart{}
+func getChartVersions(path, rancherVersion string) (libhelm.ChartVersions, error) {
+	chartVersions := libhelm.ChartVersions{}
 	helm := libhelm.Helm{
 		LocalPath: path,
 		IconPath:  path,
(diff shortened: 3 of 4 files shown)


PennyScissors:
Build failure expected until this PR is merged: https://github.com/rancher/system-charts/pull/444

paynejacob (dismissed):

cbron:
lmk when ready for review

PennyScissors:
> lmk when ready for review

Will do, I'm just trying to figure out the best way to test this.

pkg/image/resolve.go
```diff
@@ -83,9 +132,9 @@ func pickImagesFromValuesYAML(imagesSet map[string]map[string]bool, charts map[s
 		return err
 	}
 	var chartNameAndVersion string
-	for name, chart := range charts {
-		if strings.HasPrefix(relPath, chart.dir) {
-			chartNameAndVersion = fmt.Sprintf("%s:%s", name, chart.version)
+	for _, v := range chartVersions {
+		if strings.HasPrefix(relPath, v.Dir) {
+			chartNameAndVersion = fmt.Sprintf("%s:%s", v.Name, v.Version)
```
cbron, line 138: I'd prefer to keep this whole section how it was, more readable before. Is there a difference between `chart.name` and `name` ?

pkg/image/resolve.go
```diff
@@ -61,20 +58,72 @@ func getChartAndVersion(path string) (map[string]chart, error) {
...
 
-func pickImagesFromValuesYAML(imagesSet map[string]map[string]bool, charts map[string]chart, basePath, path string, info os.FileInfo, osType OSType) error {
+func isInMinMaxRange(rancherSemVer *semver.Version, min, max string) (bool, error) {
+	var minSemVer, maxSemVer *semver.Version
+	var isInRange bool
+	var err error
+	if minSemVer, err = semver.NewVersion(strings.TrimSpace(min)); err == nil {
+		if maxSemVer, err = semver.NewVersion(strings.TrimSpace(max)); err == nil {
+			isInRange = (rancherSemVer.GreaterThan(minSemVer) || rancherSemVer.Equal(minSemVer)) &&
+				(rancherSemVer.LessThan(maxSemVer) || rancherSemVer.Equal(maxSemVer))
```
cbron, line 118: There's no `greaterThanOrEqual` ?

pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -162,12 +162,12 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
 	var targetImages []string
 	switch apiContext.ID {
 	case linuxImages:
-		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", []string{}, []string{}, rkeSysImages, image.Linux)
+		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", "", []string{}, []string{}, rkeSysImages, image.Linux)
```
cbron, line 165: Need to test this is handled gracefully. It will also not be totally accurate anymore, can't we grab version from inside Rancher ?

pkg/image/export/main.go
```diff
@@ -91,12 +91,12 @@ func run(systemChartPath, chartPath string, imagesFromArgs []string) error {
 
 	k3sUpgradeImages := getK3sUpgradeImages(rancherVersion, data.K3S)
 
-	targetImages, targetImagesAndSources, err := img.GetImages(systemChartPath, chartPath, k3sUpgradeImages, imagesFromArgs, linuxInfo.RKESystemImages, img.Linux)
+	targetImages, targetImagesAndSources, err := img.GetImages(systemChartPath, chartPath, rancherVersion, k3sUpgradeImages, imagesFromArgs, linuxInfo.RKESystemImages, img.Linux)
```
cbron, line 94: @rmweir can rancherVersion still be `DEV` ?

pkg/image/resolve.go
PennyScissors: Unfortunately, no

pkg/image/resolve.go
PennyScissors: We created a new type chart that gets populated from values of Helm's chartVersion types. To me that seemed like adding unnecessary complexity when you could just use the chartVersion type directly. The code itself is also confusing because we've been using chart and version interchangeably when naming functions and variables, but they are not the same. 

```
type chart struct {
	dir     string
	version string
}

Helm's:
type ChartVersion struct {
	ChartMetadata `yaml:",inline"`
	Dir           string   `json:"-" yaml:"-"`
	LocalFiles    []string `json:"-" yaml:"-"`
	URLs          []string `json:"urls" yaml:"urls"`
	Digest        string   `json:"digest,omitempty" yaml:"digest,omitempty"`
}
type ChartMetadata struct {
	Name        string   `json:"name,omitempty" yaml:"name,omitempty"`
	Sources     []string `json:"sources,omitempty" yaml:"sources,omitempty"`
	Version     string   `json:"version,omitempty" yaml:"version,omitempty"`
	KubeVersion string   `json:"kubeVersion,omitempty" yaml:"kubeVersion,omitempty"`
	Description string   `json:"description,omitempty" yaml:"description,omitempty"`
	Keywords    []string `json:"keywords,omitempty" yaml:"keywords,omitempty"`
	Icon        string   `json:"icon,omitempty" yaml:"icon,omitempty"`
}
```

pkg/api/norman/customization/kontainerdriver/actionhandler.go
kinarashah: @PennyScissors Does `settings.ServerVersion.Get()` work here?

PennyScissors: @kinarashah Nice, I think that should work.

kinarashah: We have a couple of these GetServer funcs which use setting and then filter the RC/dev versions as needed by the feature. Could check if we could use any of these or need a new one. 
https://github.com/rancher/rancher/blob/master/pkg/controllers/management/kontainerdrivermetadata/data_getter.go#L208:6 
https://github.com/rancher/rancher/blob/master/pkg/catalog/utils/version.go#L58

PennyScissors: I'll update this in the next commit

PennyScissors:
The build is currently failing, but that's expected until PR https://github.com/rancher/system-charts/pull/444 to update missing min/max in system charts is merged.

Error:
```
2021/04/26 20:32:41 failed to fetch images from system charts: failed to get chart and version from "build/system-charts": no questions file in chart rancher-external-dns:0.1.1
```

pkg/image/export/main.go
rmweir: @cbron if it is "dev" it will be converted to 2.x.99 by kd.GetRancherVersion().

pkg/image/resolve.go
```diff
@@ -61,20 +59,72 @@ func getChartAndVersion(path string) (map[string]chart, error) {
...
+					for _, v := range versions[1:] {
+						questions, err := getQuestions(filepath.Join(path, v.Dir))
+						if err != nil {
+							return nil, fmt.Errorf("no questions file in chart %s:%s", v.Name, v.Version)
+						}
+						min, ok := questions["rancher_min_version"].(string)
+						if !ok {
+							return nil, fmt.Errorf("no rancher_min_version set in chart %s:%s", v.Name, v.Version)
+						}
+						// No ok check, max is optional
```
rmweir, line 79: Aren't both optional?

pkg/image/resolve.go
```diff
@@ -61,20 +59,72 @@ func getChartAndVersion(path string) (map[string]chart, error) {
...
+		return nil, err
+	}
+	return questions, nil
+}
+
+func isInMinMaxRange(rancherSemVer *semver.Version, min, max string) (bool, error) {
+	var minSemVer, maxSemVer *semver.Version
+	var isInRange bool
+	var err error
+	if minSemVer, err = semver.NewVersion(strings.TrimSpace(min)); err == nil {
```
rmweir, line 115: If min is optional then the two checks should be decoupled and the case where min is empty should be considered.

pkg/image/resolve.go
```diff
@@ -61,20 +59,72 @@ func getChartAndVersion(path string) (map[string]chart, error) {
...
+		if maxSemVer, err = semver.NewVersion(strings.TrimSpace(max)); err == nil {
+			isInRange = (rancherSemVer.GreaterThan(minSemVer) || rancherSemVer.Equal(minSemVer)) &&
```
rmweir, line 117: Not less than is equivalent to `greatherThanOrEqual`:
```suggestion
			isInRange = !rancherSemVer.LessThan(minSemVer) &&
```

pkg/image/resolve.go
rmweir: See above.

pkg/image/resolve.go
```diff
@@ -61,20 +59,72 @@ func getChartAndVersion(path string) (map[string]chart, error) {
...
+				(rancherSemVer.LessThan(maxSemVer) || rancherSemVer.Equal(maxSemVer))
+			return isInRange, nil
+		}
+		isInRange = rancherSemVer.GreaterThan(minSemVer) || rancherSemVer.Equal(minSemVer)
+		return isInRange, nil
```
rmweir, line 122: Same as above:
```suggestion
		return !rancherSemVer.LessThan(minSemVer)
```

pkg/image/resolve.go
```diff
@@ -61,20 +59,72 @@ func getChartAndVersion(path string) (map[string]chart, error) {
...
+						isInRange, err := isInMinMaxRange(rancherSemVer, min, max)
+						if err != nil {
+							return nil, err
+						}
+						if isInRange {
+							chartVersions = append(chartVersions, v)
+						}
+					}
+				}
+			}
```
rmweir, line 90: Would be more idiomatic to continue when co0nditions aren't met than it is to nest conditionals. Should make it easier to follow.

pkg/image/resolve.go
```diff
@@ -97,29 +147,31 @@ func pickImagesFromValuesYAML(imagesSet map[string]map[string]bool, charts map[s
...
 		return err
 	}
 
-	walkthroughMap(dataInterface, func(inputMap map[interface{}]interface{}) {
-		generateImages(chartNameAndVersion, inputMap, imagesSet, osType)
+	var imagesFound []string
+	walkthroughMap(chartValues, func(inputMap map[interface{}]interface{}) {
+		generateImages(chartNameAndVersion, inputMap, imagesSet, &imagesFound, osType)
 	})
+	log.Printf("Images found in %s: %+v", path, imagesFound)
```
rmweir, line 159: I think this is a good replacement for the log that is being removed. A slice value should just be returned from the function here, pass it as a pointer is unnecessary and not idiomatic in this scenario: https://blog.golang.org/slices. I also think instead of passing this info back you could just print this where it is being appended.

rmweir (changes requested):

pkg/image/resolve.go
PennyScissors: No, min isn't option, but max is.

pkg/image/resolve.go
PennyScissors: I'll update it to return a list rather than print each individual image. The reason for this is that I think printing a list of images (including empty for no images found) and the path to the chart it was found (including subcharts) is more useful compared to printing each individual image that was generated since then it would be harder to distinguish which images are from the chart and subchart.

pkg/image/resolve.go
PennyScissors: In my opinion the one I have is longer, but it's much easier to read since you can almost read it like a sentence in plain english, whereas your suggestion cleans it up at the expense of readability. If you don't agree I can change it, but I figured you should know why I chose one over the other first.

PennyScissors:
Closing in favor of https://github.com/rancher/rancher/pull/33576

Issue #32161, opened 2021-04-15
GKEv2UI- can have a better way to handle updates to a cluster with wrong parameters 


prabalsharma:
**What kind of request is this (question/bug/enhancement/feature request):** feature request

**Steps to reproduce (least amount of steps as possible):**
1. Create a cluster on gcloud with 2 NP's; one with  Linux image and the other with Windows image
2. Register the cluster on Rancher 
3. Go to edit and delete the only Linux NP

**Result:**
Google will throw error and Rancher will show cluster in Updating state along with the error
<img width="1372" alt="Screen Shot 2021-04-15 at 11 37 58 AM" src="https://user-images.githubusercontent.com/3749546/114921112-0ad60c80-9ddf-11eb-89c8-739d0934f432.png">

All this ^^ is as expected. 

Now, if the user wants to correct his mistake and goes back to edit cluster, the NP with Linux image is gone and there is not way for him to undo his mistake.
If the Edit were to show the current spec of the GKE cluster, instead of the last user provided spec, it may solve this issue as then the user will just have to go to Edit (where both NPs will show up) and just click save (without changing anything) to get rid of the above state. 

**Other details that may be helpful:**
The same state will occur if one changes Linux image to Windows image, 

<img width="1214" alt="Screen Shot 2021-04-15 at 11 48 08 AM" src="https://user-images.githubusercontent.com/3749546/114922408-753b7c80-9de0-11eb-8938-ff057b862609.png">

but here the reversal of state is easy as no data is lost. User can just go to edit and change the Image back to Linux from Windows.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5-4e70ed29ccc5de7c61516ba0367c7c4904989b56-head
- Installation option (single install/HA):

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

prabalsharma:
passed on v2.5-a5d9602ce1f264764b22e26a2425f99cb812cf97-head
merged to master
added to https://github.com/rancher/rancher/issues/32140

PR #32171, opened 2021-04-15
[v2.5] Add linux node pool GKE cluster validator

cmurphy:
Windows is a supported image type for GKE node pools, but it can't be
the only type of node pool in the cluster. Without this change, it's
possible to send a cluster create or update request for an invalid
cluster with only Windows node pools. For create requests this is not
much of a problem, because GKE will send an HTTP 400 response and the
user can recreate the cluster with the correct configuration. It causes
more of a problem if the user starts with a valid cluster and then sends
an update request to delete the non-Windows node pools. Because of how
syncing works, the state of the cluster object will only reflect the
invalid Windows-only node pool configuration instead of the real state
of the cluster, so the mistake cannot be corrected. The solution is to
ensure this kind of invalid request never gets made by ensuring the
cluster object has at least one non-Linux node pool.

https://github.com/rancher/rancher/issues/32161

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -656,6 +656,7 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
 	}
 
 	var errors []string
+	hasRequiredLinuxPool := false
 
 	for _, np := range nodepools {
 		name := *np.Name
@@ -668,6 +669,15 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
 			errors = append(errors, fmt.Sprintf("node pool [%s] version cannot be empty", name))
 			continue
 		}
+
+		// Windows images are WINDOWS_LTSC or WINDOWS_SAC. The cluster must have at least one non-Windows node pool.
+		if !hasRequiredLinuxPool && !strings.Contains(strings.ToLower(np.Config.ImageType), "windows") {
+			hasRequiredLinuxPool = true
+		}
+	}
+
+	if !hasRequiredLinuxPool {
+		errors = append(errors, fmt.Sprintf("at least 1 Linux node pool is required"))
 	}
 
 	if len(errors) != 0 {


pkg/api/norman/customization/cluster/validator.go
```diff
@@ -665,13 +666,23 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
...
 		version := np.Version
 		if version == nil || *version == "" {
-			errors = append(errors, fmt.Sprintf("node pool [%s] version cannot be empty", name))
+			versionErrors = append(versionErrors, fmt.Sprintf("node pool [%s] version cannot be empty", name))
 			continue
 		}
+
+		image := strings.ToLower(np.Config.ImageType)
+		// Windows images are WINDOWS_LTSC or WINDOWS_SAC. The cluster must have at least one non-Windows node pool.
+		if !strings.Contains(image, "windows") {
```
thedadams, line 675: Do we need to be concerned with the capitalization here? I noticed that your comment is in all caps.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -665,13 +666,23 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
...
 	}
 
-	if len(errors) != 0 {
-		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf(strings.Join(errors, ";")))
+	if len(versionErrors) != 0 {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf(strings.Join(versionErrors, ";")))
+	}
+
+	if !hasRequiredLinuxPool {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("at least 1 Linux node pool is required"))
```
thedadams, line 685: Is there a reason for separating these errors? For example, would it be bad to keep the name of the variable as `errors` and just append all errors to the slice, check the length of this slice, and return the error if any error is present?

thedadams (commented):
Just a few questions.

pkg/api/norman/customization/cluster/validator.go
cmurphy: It's converted to lower case on the line above. Both upper and lowercase are valid. The UI sends it as uppercase.

pkg/api/norman/customization/cluster/validator.go
cmurphy: That sounds better, I'll do that

pkg/api/norman/customization/cluster/validator.go
thedadams: I'm sorry I missed that.

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -668,6 +669,16 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
 			errors = append(errors, fmt.Sprintf("node pool [%s] version cannot be empty", name))
 			continue
 		}
+
+		// Windows images are WINDOWS_LTSC or WINDOWS_SAC. The cluster must have at least one non-Windows node pool.
+		image := strings.ToLower(np.Config.ImageType)
+		if !strings.Contains(image, "windows") {
```
thedadams, line 675: nit:
```suggestion
		if !hasRequiredLinuxPool && !strings.Contains(image, "windows") {
```

thedadams (approved):
One nit. Otherwise, looks good!

pkg/api/norman/customization/cluster/validator.go
cmurphy: done

thedadams (approved):

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -668,6 +669,16 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
 			errors = append(errors, fmt.Sprintf("node pool [%s] version cannot be empty", name))
 			continue
 		}
+
+		// Windows images are WINDOWS_LTSC or WINDOWS_SAC. The cluster must have at least one non-Windows node pool.
+		image := strings.ToLower(np.Config.ImageType)
+		if !hasRequiredLinuxPool && !strings.Contains(image, "windows") {
+			hasRequiredLinuxPool = true
+		}
```
cbron, line 677: nit: not necessary to have the `image` variable

cbron (approved):

pkg/api/norman/customization/cluster/validator.go
cmurphy: done

cmurphy:
Forward port https://github.com/rancher/rancher/pull/32200

cbron (approved):

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -668,6 +669,15 @@ func validateGKENodePools(spec *v32.ClusterSpec) error {
 			errors = append(errors, fmt.Sprintf("node pool [%s] version cannot be empty", name))
 			continue
 		}
+
+		// Windows images are WINDOWS_LTSC or WINDOWS_SAC. The cluster must have at least one non-Windows node pool.
+		if !hasRequiredLinuxPool && !strings.Contains(strings.ToLower(np.Config.ImageType), "windows") {
```
rmweir, line 674: nit: checking hasRequiredLinuxPool isn't necessary.
```suggestion
		if !strings.Contains(strings.ToLower(np.Config.ImageType), "windows") {
```

rmweir (approved):
LGTM, 1 nint

pkg/api/norman/customization/cluster/validator.go
cmurphy: Well I don't disagree but this was @thedadams's suggestion https://github.com/rancher/rancher/pull/32171#discussion_r614437854

Issue #26545, opened 2020-04-07
Removed nodes not correctly cleaned up


Tejeev:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy a 3 node custom cluster with roles `all` and wait for the cluster to stabilize
2. Delete one of the nodes

**Result:**
- Docker and k8s containers continue to run on the deleted node.
- etcd is still trying to reconnect to its peers, and in one test even able to do so and continued on as the master (though we've not reproduced this since). 
- `/etc/kubernetes/ssl` has not been cleared.
- When I restarted the node it rejoined the cluster as a worker

```
root@ip-10-0-4-253:/home/ubuntu# docker ps
CONTAINER ID        IMAGE                                 COMMAND                  CREATED             STATUS                          PORTS               NAMES
3d658f8790c4        70f311871ae1                          "/coredns -conf /etc…"   45 seconds ago      Up 44 seconds                                       k8s_coredns_coredns-7c5566588d-m585c_kube-system_bd1af8ea-f647-43ca-8c80-57020d261e1e_3
887b4fdf0ccb        rancher/kube-api-auth                 "/bin/sh -c 'kube-ap…"   29 hours ago        Up 29 hours                                         k8s_kube-api-auth_kube-api-auth-6l5ms_cattle-system_25c77d6e-59ef-40a6-9b52-4527bb14a53f_0
87d115b009f9        069831ff8196                          "run.sh"                 29 hours ago        Up 29 hours                                         k8s_agent_cattle-node-agent-r4lp7_cattle-system_93160240-7dac-4bf5-ae1a-51b1d4b349c4_0
e52cd9e1081c        rancher/pause:3.1                     "/pause"                 29 hours ago        Up 29 hours                                         k8s_POD_kube-api-auth-6l5ms_cattle-system_25c77d6e-59ef-40a6-9b52-4527bb14a53f_0
e859429fd2cf        rancher/pause:3.1                     "/pause"                 29 hours ago        Up 29 hours                                         k8s_POD_cattle-node-agent-r4lp7_cattle-system_93160240-7dac-4bf5-ae1a-51b1d4b349c4_0
[...]

PR #32181, opened 2021-04-16
[wip] node cleanup

luthermonson:
https://github.com/rancher/rancher/issues/26545


phillipsj (approved):
LGTM!

rosskirkpat (approved):
LGTM

cmd/agent/main.go
```diff
@@ -42,33 +43,72 @@ const (
...
 
-	logrus.SetOutput(colorable.NewColorableStdout())
-	logserver.StartServerWithDefaults()
-	if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
-		logrus.SetLevel(logrus.DebugLevel)
-	}
+	if len(os.Args) > 1 {
+		err = runArgs()
+	} else {
+		if _, err := reconcileKubelet(context.Background()); err != nil {
```
aiyengar2, line 51: Are we shadowing the error here?
```suggestion
		if _, err = reconcileKubelet(context.Background()); err != nil {
```

cmd/agent/main.go
```diff
@@ -42,33 +43,72 @@ const (
...
+	switch os.Args[1] {
+	case "clean":
+		return runClean()
+	default:
+		return run()
+	}
+}
+
+func runClean() error {
+	if len(os.Args) > 2 {
```
aiyengar2, line 87: nit: Is it valid to pass in more than 2 `os.Args`? If not, we should return an error.

package/windows/agent.ps1
```diff
@@ -0,0 +1,2 @@
+Start-Process -NoNewWindow -Wait -FilePath "c:\etc\rancher\agent.exe" -ArgumentList $args
```
aiyengar2, line 1: Do we need to worry about prefixPath here?

package/windows/bootstrap.ps1
```diff
@@ -89,7 +89,6 @@ try
 {
     Copy-Item -Force -Destination "$($hostPrefixPath)etc\rancher" -Path @(
         "c:\etc\rancher\utils.psm1"
-        "c:\etc\rancher\cleanup.ps1"
```
aiyengar2, line 92: Should we still leave this script in in case someone wants to run manual cleanup?

package/windows/bootstrap.ps1
```diff
@@ -286,6 +288,7 @@ catch
             "$($CATTLE_PREFIX_PATH)etc\nginx\nginx.exe"
             "$($CATTLE_PREFIX_PATH)opt\bin\flanneld.exe"
             "$($CATTLE_PREFIX_PATH)etc\rancher\wins\wins-upgrade.exe"
+            "$($CATTLE_PREFIX_PATH)etc\rancher\powershell.exe"
```
aiyengar2, line 291: We need to add this path to the wins-upgrader chart defaults:

https://github.com/rancher/wins/blob/036e27d9a3254471868e29e165710f5a6ba24546/charts/rancher-wins-upgrader/values.yaml#L27-L33

pkg/agent/clean/node.go
```diff
@@ -0,0 +1,349 @@
...
+func Job() error {
+	logrus.Infof("Starting clean container job: %s", ContainerName)
+
+	c, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return err
+	}
+	defer c.Close()
+
+	ctx := context.Background()
```
aiyengar2, line 36: I think a general good practice would be to create the context in main and pass it in via the function signature right?

pkg/agent/clean/node.go
```diff
@@ -0,0 +1,349 @@
...
+				logrus.Infof("container named %s already exists, exiting.", ContainerName)
+				return nil
+			}
+		}
+	}
+
+	binds := []string{
+		"/var/run/docker.sock:/var/run/docker.sock",
+		"/:/host:z",
+	}
```
aiyengar2, line 55: For all of these static strings (shown throughout the PR), can we make them all constants or vars declared on the top of the file when possible?

e.g. even some of the bash commands should be declared as constants at the top IMO.

If this creates too much stuff at the top of the file, maybe this file should also be split into multiple files for each one of the calls (e.g. firewall.go, paths.go, etc.)

pkg/agent/clean/node.go
```diff
@@ -0,0 +1,349 @@
...
+func stopContainers(ctx context.Context, c *client.Client) error {
+	containers, err := c.ContainerList(ctx, types.ContainerListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, container := range containers {
+		isCleanup := false
+		for _, n := range container.Names {
+			if strings.HasPrefix(n, "/"+ContainerName) {
```
aiyengar2, line 228: Can we get the current container by ID instead of by name?

pkg/agent/clean/node.go
```diff
@@ -0,0 +1,349 @@
...
+
+		if isCleanup {
+			continue // don't stop the cleanup container!
+		}
+
+		config, err := c.ContainerInspect(ctx, container.ID)
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(config.Config.Image, "rancher/") {
```
aiyengar2, line 242: Just wanted to clarify; for private registry images, does the image name still start with `rancher/` always? Or could they change the image name as part of mirroring?

Wondering if this should be exposed as an env var option.

pkg/agent/clean/node.go
```diff
@@ -0,0 +1,349 @@
...
+
+func getPaths() []string {
+	pathPrefix := getPrefixPath()
+
+	// all paths on the hosts to rm -rf
+	return []string{
+		hosts.ToCleanCNIConf, hosts.ToCleanCNIBin,
+		"/opt/rke", "/run/secrets/kubernetes.io",
+		"/run/calico", "/run/flannel", "/var/log/containers", "/var/log/pods",
+		hosts.ToCleanCalicoRun,
```
aiyengar2, line 339: why don't these have pathPrefix prefixed on them? Don't we place these in the path prefix as well?

pkg/agent/clean/node_windows.go
```diff
@@ -0,0 +1,293 @@
...
+		return err
+	}
+	return nil
+}
+
+func writeScript() error {
+	// add a null file to the container for wins to find and make a hash
+	psPath := getPowershellPath()
+	if !fileExists(psPath) {
+		psHostPath := strings.Replace(psPath, "c:\\", "c:\\host\\", 1)
```
aiyengar2, line 107: maybe this should be in a helper function `getHostPathFromContainer` (returns the string with host prepended) and `getContainerPath` (returns the exact string back) to make it obvious to read?

Would also be nice if we can add `getHostPathFromHost` which handles the prefixPath logic there itself.

cmd/agent/main.go
```diff
@@ -42,33 +43,72 @@ const (
...
+		err = runArgs()
+	} else {
+		if _, err := reconcileKubelet(context.Background()); err != nil {
+			logrus.Warnf("failed to reconcile kubelet, error: %v", err)
+		}
 
-	initFeatures()
+		logrus.SetOutput(colorable.NewColorableStdout())
+		logserver.StartServerWithDefaults()
+		if os.Getenv("CATTLE_DEBUG") == "true" || os.Getenv("RANCHER_DEBUG") == "true" {
```
aiyengar2, line 57: Instead of having a ton of `os.Getenv` calls, could we just parse all of this into an `AgentConfig` struct that parses all of the environment variables before running an actual command that gets passed into the commands itself?

This would avoid having to do a ton of stuff around helper functions defined below since we can just reference the value from methods defined on `AgentConfig` itself, e.g. `getHostPathFromContaienr`, `getHostPathFromHost`, and `getContainerPath`.

aiyengar2 (changes requested):
Nits and clarifications

pkg/agent/clean/node.go
luthermonson: great question.... all our mirror scripts put them into rancher/* repos as far as i understand... wouldnt mind a second opinion on that tho

pkg/agent/clean/node.go
luthermonson: you'd have to read rke for history on that.. this was copy/pasted from existing rke functions and then added some too... that's where RKE does it that's where we have to clean them up from.

cmd/agent/main.go
luthermonson: ya good call this should have no colon

package/windows/agent.ps1
luthermonson: this is inside the container, it will always be in c:\etc\rancher

package/windows/bootstrap.ps1
luthermonson: and this is why i wanted you to look at this ;)

package/windows/bootstrap.ps1
luthermonson: it doesnt really work and the new one can be pulled from the container with 1 command and will be documented but the gist is `docker run rancher/rancher-agent -- agent clean scripts > cleanup.ps1` will get you what you need

pkg/agent/clean/node.go
luthermonson: let's add an env var to let people toggle this, will default to rancher/ but will let people set via CATTLE_REPO_PREFIX

cmd/agent/main.go
luthermonson: 👍

package/run.sh
```diff
@@ -26,7 +26,7 @@ if [ "$1" = "--" ]; then
     exec "$@"
 fi
 
-if [ "$CLUSTER_CLEANUP" = true ]; then
+if [ "$CLUSTER_CLEANUP" = true ] || [ "$NODE_CLEANUP_JOB" = true ] || [ "$NODE_CLEANUP" = true ]; then
```
luthermonson, line 29: this isnt used any more.. derp

pkg/agent/clean/node_windows.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+	} else {
+		logrus.Infof("powershell.exe already exists: %s", psPath)
+	}
+
+	// write one to the host for wins cli to call
+	scriptBytes := []byte(PowershellScript)
+	hostScriptPath := strings.Replace(getScriptPath(), "c:\\", "c:\\host\\", 1)
+	if !fileExists(hostScriptPath) {
+		logrus.Infof("writing file to host: %s", hostScriptPath)
```
luthermonson, line 132: add to logs about running script by hand if something fails

PR #32202, opened 2021-04-17
[v2.5.8 only] Update rancher-operator-min-version Setting

cmurphy:
Without this change, if Rancher is upgraded from an older version to the
current version, rancher-operator does not get upgraded with it. This
causes problems for the new GKE provisioner which requires bugfixes in
the latest releases of rancher-opreator. Updating the environment
variable in the Dockerfile ensures the rancher-operator-min-version
Setting resource gets the latest version and the rancher-operator
Deployment will get updated on upgrade.

https://github.com/rancher/rancher/issues/32124


rmweir (dismissed):
LGTM

thedadams (approved):

cbron (approved):

PR #32217, opened 2021-04-19
Fix Chart Repo shouldRefresh logic for inconsistent repo definitions

bashofmann:
When the RepoSpec of a chart repo contains a git branch and an http url, shouldRefresh returns always true, because the unnecessary branch is never set in the RepoStatus. As a result the agent refreshes the repo every 4 seconds. This puts a high load on the agent as well as etcd and causes significant network traffic.

This can happen when a Repo is created manually through the Kubernetes API or with Terraform. The rancher2 terraform provider currently even sets gitBranch to "master" by default, even if a http url is set.

Issue: https://github.com/rancher/rancher/issues/23256.


brendarearden (approved):
LGTM

PR #32219, opened 2021-04-19
Add GKE validator for IP ranges

cmurphy:
The primary and secondary IP ranges of a GKE cluster are immutable, so
they need to be correct when they are first provided. This change adds
some surface-level validation for the node, pod, and services IP ranges
if provided for the cluster. The valid ranges are based on the
documented requirements[1]. This will not be able to validate whether an
IP range is overlapping with the range of an existing cluster, and it
does not check whether the node range is valid for the number of nodes
in the cluster, which would need to take into account the number of
nodes in each node pool as well as the node autoscaling configuration.

[1] https://cloud.google.com/kubernetes-engine/docs/concepts/alias-ips

https://github.com/rancher/rancher/issues/32207

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"reflect"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmurphy:
Wes is going to work on this in UI

thedadams (approved):
Looks good. Maybe add a comment on the integer comparisons because it does look strange with the inverted nature of the CIDR blocks.

cmurphy:
Closing in favor of handling this in the UI

PR #32228, opened 2021-04-19
Update private registry logic

dbason:
Check if the url field exists in privateRegistry settings.  This allows for reusing password where the url isn't set.

Still need to do a bit more dev testing but these are the preliminary changes


pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -1099,3 +1099,10 @@ func canUpgrade(nodes []*v3.Node, upgradeStrategy *rketypes.NodeUpgradeStrategy)
 	}
 	return nil
 }
+
+func comparePrivateRegistryUrls(newConfig map[string]interface{}, oldConfig map[string]interface{}) bool {
```
superseb, line 1103: Is this function name based on another one in this file? The function is called `comparePrivateRegistryUrls` but the return is a boolean, this probably needs an expression so that the boolean confirms or denies the expression

Did we also check if `replaceWithOldSecretIfNotExists` cannot be used?

dbason: Sorry about the bad naming, I will fix that if we use the new method.

I haven't checked `replaceWithOldSecretIfNotExists`, but it looks like that is explicitly expecting a cloudProvider at the moment.  The idea behind a new method was to keep the logic as close as possible to what was there already.

superseb (changes requested):
This needs to be rebased for CI to pass.

dbason:
Rebased, and the linting error is fixed.

HarrisonWAffel:
This PR is no longer needed as other secret migration logic has addressed this issue indirectly.

PR #32236, opened 2021-04-20
[2.5] Pin webhook to 0.1.000 in dockerfile

StrongMonkey:


package/Dockerfile
```diff
@@ -40,7 +40,7 @@ ENV KUSTOMIZE_VERSION v3.5.4
 # System charts minimal version
 ENV CATTLE_FLEET_MIN_VERSION=0.3.500
 ENV CATTLE_RANCHER_OPERATOR_MIN_VERSION=0.1.300
-ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=0.1.0-beta900
+ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=0.1.000
```
dramich, line 43: Should this be https://github.com/rancher/charts/blob/dev-v2.5/charts/rancher-webhook/rancher-webhook/0.1.000-rc02/Chart.yaml#L11

StrongMonkey: It doesn't have to be. 0.1.000 > 0.1.000-rc2 so it will just choose the latest rc chart until official 0.1.000 is released.

dramich: So this envvar doesn't do what it says then. If I say my min version is 0.1.000 I would not expect to get an rc of that version installed

StrongMonkey: if your min version is 0.1.000, and none of version meets this criteria, it will just pick up the latest version. It is just a way to make things easier so you don't have to update it multiple times.

StrongMonkey: Setting this so that in 2.5-head/master-head, you always get latest version of chart. Once the chart is released, it starts pinning down the version and prevent uprgade.

dramich (approved):

PR #32270, opened 2021-04-21
Images txt should not source blank image strings

cbron:
If you don't pass in a imageArg (like the agent) and don't set REPO env param you will error out the script. But this could happen for other reasons, and if so we should ignore the image.


rmweir (approved):
LGTM, although not sure how necessary it is since this won't occur if the expected arguments are passed (rancher and agent images).

cbron:
Not backporting, but will merge master.

Issue #32231, opened 2021-04-20
Add ability to pass files to machine-provisioning job


thedadams:
When machines are provisioned in v2, we need a way to pass files to the pod for things like `userdata`, `ssh`, and `creds`.

izaac:
We will wait once the RKE2 OCI is implemented as this will take much effort to do through api objects.

UI issue:
https://github.com/rancher/dashboard/issues/2767

snasovich:
@kinarashah or @rmweir , can one of you pick up validation of this since I cannot assign validation to Donnie as he was the one who implemented it?
@cbron, FYI.

cbron:
Assigning @rmweir for dev-validate

rmweir:
Was able to validate this with DO.
1. Added following to a do config:
```
userdata: |-
    #cloud-config
    final_message: "CLOUD INIT WAS ADDED"
```
2. Added key to `sshKeyContents`
3. Added fingerprint of public key to `sshKeyFingerprint`
4. Created RKE2 cluster with config.
5. After cluster provisioned successfully, ssh'd into cluster using key provided in 2, and checked `/var/log/cloud-init-output.log` for "CLOUD INIT WAS ADDED".

**Results:**
Cluster provisioning being successfully should mean the ssh key was stored successfully, being able to ssh into the node means the correct public key was used by do. `/var/log/cloud-init-output.log` contained "CLOUD INIT WAS ADDED" which means the cloud-init was used.

Attempted to follow the same steps for AWS but encountered errors each time, often different ones while using the same input. Some mentioned that accessKey and secretKey were not included, but those should be provided by the cloud credential. Another printed all the flags that could be passed to rancher-machine.

thedadams:
The issues encountered with AWS were not related to the passing of files to rancher-machine. Since Ricky was able to get both Digital Ocean and EC2 working, this issue is validated.

PR #32275, opened 2021-04-21
Support passing files into machine provisioning job

thedadams:
Some of the arguments to rancher-machine are paths to files needed for
provisioning. These changes add support for passing these files into the
pod when necessary.

Issue:
https://github.com/rancher/rancher/issues/32231


thedadams:
The implementation is changing from using a configmap to using a secret. This update in in progress.

PR #32277, opened 2021-04-21
[2.6] exclude some status checks for some providers

paynejacob:
2.6 version of https://github.com/rancher/rancher/pull/31890

ISSUE: #11496

PROBLEM:
Not all providers implement component status correctly, e.g. AKS shows the scheduler in an error state. We need to ignore some component statuses as they can be misleading.

SOLUTION:

Added a filter to the health syncer that does not return some component statues based on provider.

TESTING:

I was unable to get access to an AKS cluster but, I could filter components out of a k3os cluster by adding it to the exclusion map.

### DIFF
--- a/pkg/controllers/managementuser/healthsyncer/healthsyncer.go
+++ b/pkg/controllers/managementuser/healthsyncer/healthsyncer.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/condition"
+	"github.com/rancher/norman/types/slice"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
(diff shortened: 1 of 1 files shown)


dramich (approved):

PennyScissors (approved):

mohag:
Will this actually work for imported clusters (of those types) and modern clusters that disable HTTP on the components? (e..g anything created with a current kubeadm version)

(It feels, for a quick fix, like an option to disable those statuses that can be set on any cluster might be useful. In some cases, where the stuff is running, but without the insecure port that component-statuses demand, the option to designate a filter to find the pods' health checks to monitor instead might be better, but harder to implement...)

glego:
Hi, 

With AKS it's ignoring the Controller Manager and Scheduler just fine. But like @mohag is mentioning, it would be nice to also apply this for imported clusters.

Is it somehow possible?

image

Issue #31014, opened 2021-01-27
AKS Provisioner: Cloud Credentials


gaktive:
We need to look at using cloud credentials with the AKS Provisioner.
https://github.com/rancher/rancher/issues/31014

Fix: https://github.com/rancher/rancher/pull/32288

cbron:
This one is just the integration to the cluster object, as the azure cloud credentials themselves already exist.

gaktive:
Need to add the TenantID still along with some other bits of code.

mjura:
After discussion we agreed to not add tenantId to Cloud Credentials, because it might brake rancher machine for RKE.

cbron:
Reopening, we are looking at adding again.

cbron:
QA cases to test: 
* that it doesn't break existing setups without a tenant
* fresh clusters with and without a tenant
* not sure this is valid use case, but see if you can convert existing cloud-credential cluster to one with a tenant

cbron:
We also need to test plain old azure node driver with and without tenant, and upgrading that from 2.5

prabalsharma:
> QA cases to test:
> 
> * that it doesn't break existing setups without a tenant

Does not, as AKSv1 don't use stored CC

> * fresh clusters with and without a tenant

This is not possible, as cc created without tenantID does not show up in list of CCs while provisioning AKSv2 cluster

> * not sure this is valid use case, but see if you can convert existing cloud-credential cluster to one with a tenant

No possible. V1 cluster does not use stored CC

PR #32288, opened 2021-04-22
Add tenatdId to AKSv2 cloud credentials

mjura:
Part of AKSv2 cloud credentials is also tenantId, which has to be added.

rancher/rancher#31014


kinarashah (approved):

thedadams (approved):

mjura:
We would like to drop this change, `tenatId` will be catched in UI panel for AKS options

cbron:
@mjura do you mean in the proxy calls ? We can't just use UI as terraform/API needs to still work.

cmurphy (approved):
This won't be useful until there is a flag in the machine driver, but I think it's all that's needed for the Rancher side.

cbron (approved):
Agree, just wait to merge

thedadams (approved):

Issue #32290, opened 2021-04-22
Feature charts with Windows images do not have those included in rancher-windows-images.txt


nickgerace:

nickgerace:
**Details pertaining to #32291**: The ideal solution would be to remove the concept of OSType since the existing code creates rancher-images.txt and then rancher-windows-images.txt (n^2 runtime complexity). However, this solution aims to add the comma-delineated list functionality to the existing structure, rather than uprooting it. Moreover, there are only two OSType constants, so the runtime complexity is minimally impacted (realistically around 2n).

nickgerace:
Solution Explanation: https://github.com/rancher/rancher/pull/32291#issuecomment-825115713

nickgerace:
**Note to QA:** validating this issue will focus on ensuring the expected images are in `rancher-windows-images.txt`.

izaac:
Rancher version `v2.5.8-rc10`

Logging passed the install and k8s objects creation.
Monitoring: https://github.com/rancher/rancher/issues/32265 (failed, missing image)
Wins upgrader: https://github.com/rancher/rancher/issues/32402 (failed, missing image)
Fleet: pending.

PR #32291, opened 2021-04-22
Use comma-delineated list for OS in image list generation

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/32290
Original solution basis: https://play.golang.org/p/_Dj725132Lm
Refactor basis: https://play.golang.org/p/k-6IyqTPE66
Solution Explanation: https://github.com/rancher/rancher/pull/32291#issuecomment-825115713

### DIFF
--- a/pkg/image/resolve.go
+++ b/pkg/image/resolve.go
@@ -119,21 +119,34 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
 		return
 	}
 
-	// distinguish images by os
-	os := inputMap["os"]
-	switch os {
-	case "windows": // must have indicate `os: windows` if the image is using in Windows cluster
-		if osType != Windows {
-			return
+	imageName := fmt.Sprintf("%s:%v", repo, t)
+
+	// By default, images are added to the generic images list ("linux"). For Windows and multi-OS
+	// images to be considered, they must use a comma-delineated list (e.g. "os: windows",
+	// "os: windows,linux", and "os: linux,windows").
+	if osList, ok := inputMap["os"].(string); ok {
+		for _, os := range strings.Split(osList, ",") {
+			switch strings.TrimSpace(strings.ToLower(os)) {
+			case "windows":
+				if osType == Windows {
+					addSourceToImage(output, imageName, chartNameAndVersion)
+					return
+				}
+			case "linux":
+				if osType == Linux {
+					addSourceToImage(output, imageName, chartNameAndVersion)
+					return
+				}
+			}
 		}
-	default:
-		if osType != Linux {
-			return
+	} else {
+		if inputMap["os"] != nil {
+			panic(fmt.Sprintf("Field 'os:' for image %s contains neither a string nor nil", imageName))
+		}
+		if osType == Linux {
+			addSourceToImage(output, imageName, chartNameAndVersion)
 		}
 	}
-
-	imageName := fmt.Sprintf("%s:%v", repo, t)
-	addSourceToImage(output, imageName, chartNameAndVersion)
 }
 
 func addSourceToImage(imagesSet map[string]map[string]bool, image string, sources ...string) {


thedadams (changes requested):

pkg/image/resolve.go
```diff
@@ -119,20 +119,39 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
...
+	if inputMap["os"] != nil {
+		switch inputMap["os"].(interface{}).(type) {
+		case string:
+			foundMatchingOSType := false
+			for _, os := range strings.Split(inputMap["os"].(string), ",") {
+				switch strings.TrimSpace(strings.ToLower(os)) {
+				case "windows":
+					if osType == Windows {
+						foundMatchingOSType = true
+						break
```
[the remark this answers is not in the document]
nickgerace: Ah good catch. Artifact from previous commit

pkg/image/resolve.go
```diff
@@ -119,20 +119,39 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
...
-	default:
-		if osType != Linux {
-			return
+	imageName := fmt.Sprintf("%s:%v", repo, t)
+
+	// By default, images are added to the generic images list ("linux"). For Windows and multi-OS
+	// images to be considered, they must use a comma-delineated list (e.g. "os: windows",
+	// "os: windows,linux", and "os: linux,windows").
+	if inputMap["os"] != nil {
+		switch inputMap["os"].(interface{}).(type) {
```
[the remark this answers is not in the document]
nickgerace: Good idea.

thedadams (dismissed):

nickgerace:
Rebased after Donnie's approval

aiyengar2 (approved):
LGTM

pkg/image/resolve.go
```diff
@@ -119,20 +119,37 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
 		return
 	}
 
-	// distinguish images by os
```
aiyengar2, line 122: https://github.com/rancher/rancher/blob/f6dfe7ca192c4f87c4d197ab95888a1bf15d6dab/pkg/image/resolve.go#L168-L170

Can we add another if statement here that says:

```go
 if v, ok := value.([]interface{}); ok {
    for _, elem := range v {
        walkthroughMap(v, walkFunc) 
    }
 }
```

I can make this in a separate PR, but it's a really small change. This currently still breaks wins-upgrader, since the image exists within a list as seen here:

https://github.com/rancher/charts/blob/99131a36a40414789a9f3de3930bf7988a23a85e/charts/rancher-wins-upgrader/rancher-wins-upgrader/0.0.100-rc01/values.yaml#L19-L22

aiyengar2 (changes requested):

pkg/image/resolve.go
nickgerace: I'm fine with this change so long as @cbron is. However, this would allow for more options to our charts' images format.

pkg/image/resolve.go
```diff
@@ -119,20 +119,37 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
...
-	default:
-		if osType != Linux {
-			return
+	imageName := fmt.Sprintf("%s:%v", repo, t)
+
+	// By default, images are added to the generic images list ("linux"). For Windows and multi-OS
+	// images to be considered, they must use a comma-delineated list (e.g. "os: windows",
+	// "os: windows,linux", and "os: linux,windows").
+	if inputMap["os"] != nil {
+		switch osValue := inputMap["os"].(type) {
```
cbron, line 128: Just check if string and fail if not.

pkg/image/resolve.go
```diff
@@ -119,20 +119,37 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
...
+		case string:
+			foundMatchingOSType := false
```
cbron, line 130: Why not just call `addSourceToImage` when OS is found and return.

cbron (changes requested):

pkg/image/resolve.go
aiyengar2: Will move into another PR

aiyengar2 (approved):

pkg/image/resolve.go
nickgerace: I believe you need a switch statement when checking the type. Plus this field is allowed to be `nil`.
You will receive the following error when trying to evaluate the type:
```
use of .(type) outside type switch compiler BadTypeKeyword
```

If we try to use `nil` as a switch case, then we can face a runtime error.
Here is an example from an older run: https://drone-pr.rancher.io/rancher/rancher/14913/1/2

Thus, this PR is formatted to:

1. Check if the `interface{}` is `nil` with an if statement.
2. Check if the `interface{}` is `string` with a switch statement.

pkg/image/resolve.go
nickgerace: Good point. That's a much better idea.

pkg/image/resolve.go
nickgerace: Now, I think we could refactor this to be more readable, so I will do that.

I wrote this to help illustrate that plan: https://play.golang.org/p/k-6IyqTPE66

nickgerace: I've made two commits to illustrate the idea. The first commit is the optimized version of the original idea, and the second commit is a new idea that might be more readable, but also more verbose. Which do you prefer?

- e63f8e8c56312472dc99f401b49efd3bfbbe7c18
- 3f1f87b41d58b8ba3fdc09fbb12b2ffed9e695d7

aiyengar2 (dismissed):

pkg/image/resolve.go
cbron: Can we do the 2nd with this: https://play.golang.org/p/Qfe11eUA2NE

pkg/image/resolve.go
```diff
@@ -119,21 +119,32 @@ func generateImages(chartNameAndVersion string, inputMap map[interface{}]interfa
...
+
+	// By default, images are added to the generic images list ("linux"). For Windows and multi-OS
+	// images to be considered, they must use a comma-delineated list (e.g. "os: windows",
+	// "os: windows,linux", and "os: linux,windows").
+	if osList, ok := inputMap["os"].(string); ok {
+		for _, os := range strings.Split(osList, ",") {
+			switch strings.TrimSpace(strings.ToLower(os)) {
+			case "windows":
+				if osType == Windows {
+					addSourceToImage(output, imageName, chartNameAndVersion)
```
cbron, line 132: Don't you need return here, so it doesn't get added multiple times ? Also need to squash commits

pkg/image/resolve.go
aiyengar2: https://github.com/rancher/rancher/pull/32324

pkg/image/resolve.go
nickgerace: No, because the osType will be the same throughout the loop. Returning is optional

cbron (approved):

PR #32299, opened 2021-04-22
Validate actual cluster names in GKE

cmurphy:
Currently, the validateGKEClusterName validator checks the new cluster
name against the names and locations of clusters that Rancher already
knows about, and does not account for any potential clusters that exist
in GKE but have not been registered with Rancher. If the user creates a
cluster with the same name as another cluster in the same region or zone
that isn't known to Rancher, it will start to provision it but meet a
400 response from GKE. Since the clusterName field isn't editable, the
problem isn't fixable after the fact and the only option is to delete
and recreate the cluster. This commit changes the validator to actually
query GKE for current clusters instead of just the cluster list in
Rancher. This will provide a more accurate view of the current clusters
and will allow the user to correct the mistake before provisioning has
started.

https://github.com/rancher/rancher/issues/32292

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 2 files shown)


thedadams:
I don't remember if this logic is already in GKE, but that is where it belongs.

What follows is my opinion and I am happy to be overruled by anyone involved:

Adding this logic in norman creates complexity and GKE-specific logic in a general space. I think it is fine that a user would have to delete the cluster and create another if they create a cluster with the same name of one that already exists in GKE. If this code is in gke-operator, then it will get caught quickly and the user can perform the manual actions. This is the way EKS works as well.

cmurphy:
> I think it is fine that a user would have to delete the cluster and create another

tbh I don't disagree but since this was a bug raised by QA maybe a broader discussion is needed

> This is the way EKS works as well.

I can't find anything similar in EKS, the closest I can find is this:

https://github.com/rancher/eks-operator/blob/ef42f2f60025bbd7c63fa6f287e10d013f92c87b/controller/eks-cluster-config-handler.go#L521-L529

which is just doing the same thing this code was doing before this change, i.e. using the Cluster list to check names but not actually checking all the clusters in EKS. It also would not help here, because even if it returns an error, the DisplayName field is immutable:

https://github.com/rancher/eks-operator/blob/ef42f2f60025bbd7c63fa6f287e10d013f92c87b/pkg/apis/eks.cattle.io/v1/types.go#L37

which means it would face the same problem, if you try to go to Edit the cluster the update to that field will get ignored and the name will stay the same on the Cluster object.

I think it's probably not a good idea to remove the `norman:"noupdate"` tag from this field, that would be an easy fix for this problem but I'm guessing would cause other problems.

thedadams:
Line 534 of the file you referenced in eks-operator lists the clusters from EKS (we only want to do this check for non-imported clusters)

cmurphy:
Closing this in favor of providing a nicer error message in gke-operator and potentially addressing in UI later.

PR #32309, opened 2021-04-23
Set condition messages for "Provisioned" condition

cmurphy:
Without this patch, the changes in how wrangler summarizes
conditions[1][2] causes the condition type, in this case "Provisioned",
to be rendered as the cluster transition message. Since this string
doesn't match the display state "Provisioning", the transition message
is displayed underneath the display state. Setting an non-empty message
on the condition makes this slightly less confusing.

[1] https://github.com/rancher/wrangler/commit/3eba78f45e7d61364efd11d73334641c17305b4c
[2] https://github.com/rancher/wrangler/commit/3533ace946aaf8cc64ae4be9ead3221c35975884

https://github.com/rancher/rancher/issues/32165


ibuildthecloud:
We could probably just revert the changes in wrangler. It was an attempt to derive some useful status from CAPI, but I don't think it's really worked out.

ibuildthecloud:
So I'm fine if you'd rather want to just revert the changes in pkg/summary/condition.go.

cmurphy:
Cool https://github.com/rancher/wrangler/pull/162

Issue #32280, opened 2021-04-21
[rancher-wins-upgrader] rancher/wins image doesn't show up in air gapped setups


aiyengar2:
rancher/wins doesn't show up in rancher-images.txt, probably because image and repository is a field within a list as seen [here](https://github.com/rancher/charts/blob/6cbef79083183ff1f0d1e3070de12b343dc61ff3/released/charts/rancher-wins-upgrader/rancher-wins-upgrader/0.0.100/values.yaml#L21-L22).

Needs to be resolved to support wins upgrades in air gapped setups. Alternatively, a user could workaround this issue by using the rancher/windows_exporter-package image instead, although that contains an extra unnecessary binary.

sowmyav27:
@aiyengar2 is this related to this issue - https://github.com/rancher/rancher/issues/32265 ?

aiyengar2:
I think this is unrelated, but I'm not sure. The wins-upgrader chart is different than other charts since the image and repository tag exist in a list, so it's possible our scripts just don't handle that at this point.

But I'm not familiar enough with the scripts to know that off-hand, so may need to investigate.

nickgerace:
I actually do think they are related. I think feature charts Windows images are not being added to `rancher-windows-images.txt`

aiyengar2:
Reassigned to @nickgerace since the fix for this is being tackled in https://github.com/rancher/rancher/issues/32290

cc: @cbron @SheilaghM

aiyengar2:
After talking to @nickgerace , I re-assigned this back to me since chart changes are required to add the `os` tag to each image.

slickwarren:
this appears to be working as of 2.5.8-rc11 (the chart installs successfully)
<img width="929" alt="Screen Shot 2021-04-27 at 5 22 19 PM" src="https://user-images.githubusercontent.com/16691014/116328159-41b40700-a77d-11eb-9c09-a769e2ed376b.png">

PR #32324, opened 2021-04-23
Allow walkthroughMap to walk through list entries

aiyengar2:
Allows Rancher to pick up images embedded in lists like rancher-wins-upgrader [here](https://github.com/rancher/charts/blob/6cbef79083183ff1f0d1e3070de12b343dc61ff3/released/charts/rancher-wins-upgrader/rancher-wins-upgrader/0.0.100/values.yaml#L21-L22).

Related Issue: https://github.com/rancher/rancher/issues/32280

### DIFF
--- a/pkg/image/resolve.go
+++ b/pkg/image/resolve.go
@@ -146,26 +146,18 @@ func addSourceToImage(imagesSet map[string]map[string]bool, image string, source
 }
 
 func walkthroughMap(data interface{}, walkFunc func(map[interface{}]interface{})) {
-	inputMap, isMap := data.(map[interface{}]interface{})
-	if isMap {
+	if inputMap, isMap := data.(map[interface{}]interface{}); isMap {
 		// Run the walkFunc on the root node and each child node
 		walkFunc(inputMap)
 		for _, value := range inputMap {
 			walkthroughMap(value, walkFunc)
 		}
-		return
-	}
-	// Check if data is a list
-	inputList, isList := data.([]interface{})
-	if isList {
+	} else if inputList, isList := data.([]interface{}); isList {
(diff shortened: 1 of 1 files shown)


pkg/image/resolve.go
```diff
@@ -145,13 +145,27 @@ func addSourceToImage(imagesSet map[string]map[string]bool, image string, source
...
+	inputList, isList := data.([]interface{})
+	if isList {
+		// Run the walkFunc on each element in the root node, ignoring the root itself
+		for _, elem := range inputList {
+			walkthroughMap(elem, walkFunc)
+		}
+		return
 	}
+	// Element is neither a map nor a list
+	return
```
thedadams, line 168: nit for fewer `return`s:
```suggestion
	if inputMap, isMap := data.(map[interface{}]interface{}); isMap {
		// Run the walkFunc on the root node and each child node
		walkFunc(inputMap)
		for _, value := range inputMap {
			walkthroughMap(value, walkFunc)
		}
	} else if inputList, isList := data.([]interface{}); isList {
		// Run the walkFunc on each element in the root node, ignoring the root itself
		for _, elem := range inputList {
			walkthroughMap(elem, walkFunc)
		}
	}
```

thedadams (approved):
One nit just for code flow.

pkg/image/resolve.go
nickgerace, line 168: Is this return necessary?

aiyengar2: Nope, just for style. Can remove

nickgerace (approved):

thedadams (approved):

dramich (approved):

Issue #32298, opened 2021-04-22
Cannot import k3s v1.21


rancher-max:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Create a k3s v1.21.0-rc1+k3s1 cluster
2. Attempt to import it into rancher

**Result:**
It fails to successfully import:
![image](https://user-images.githubusercontent.com/55997940/115793684-66753c80-a381-11eb-804b-a4067fab90a7.png)

**Other details that may be helpful:**
The k3s apiserver logs with --v=9 show the following:
```
pr 22 15:33:14 centos01 k3s: E0422 15:33:14.764538    7904 authentication.go:63] "Unable to authenticate the request" err="[invalid bearer token, Token has been invalidated]"
Apr 22 15:33:14 centos01 k3s: I0422 15:33:14.764604    7904 httplog.go:89] "HTTP" verb="GET" URI="/version" latency="4.719321ms" userAgent="rancher/v0.0.0 (linux/amd64) kubernetes/$Format cluster c-ptdl5" srcIP="10.42.0.10:46508" resp=401
--
Apr 22 15:33:18 centos01 k3s: E0422 15:33:18.952774    7904 authentication.go:63] "Unable to authenticate the request" err="[invalid bearer token, Token has been invalidated]"
Apr 22 15:33:18 centos01 k3s: I0422 15:33:18.952863    7904 httplog.go:89] "HTTP" verb="GET" URI="/api/v1/namespaces/kube-system?timeout=45s" latency="4.536598ms" userAgent="rancher/v0.0.0 (linux/amd64) kubernetes/$Format cluster c-ptdl5" srcIP="10.42.0.10:46508" resp=401
```

- No errors in the `cattle-cluster-agent` pod logs. It actually appears to be running. 
- Potentially related to 1.21 changes with service account issuer discovery - it’s mandatory that you provide an OpenID provider endpoint now, and upstream kubeadm just points it at the provider built in to the apiserver which is what k3s 1.21 also now does. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.7 and v2.5.8-rc8
- Installation option (single install/HA): single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
[...]

PR #32332, opened 2021-04-24
Pull token from secret to support k8s 1.21

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/32298

Issue #30683, opened 2021-01-05
[Cluster Manager UI] Update UI to support new env var structure


PennyScissors:
Issue: https://github.com/rancher/rancher/issues/16148

The Cluster Manager UI needs to be updated to support the new Name/Value Name/ValueFrom structure.
Only affects Cluster Manager UI.

codyrancher:
@paynejacob

As one of the reviewers for the backend would you mind confirming my understanding here?


Is this the new interface of the `environment` field within the container schema?
```
[
  {
      name: "variableName",
      valueFrom: "secret, config or container if used",
      value: "value for key value pairs or the name of the key within the resource pointed to by valueFrom"
  }
]
```

paynejacob:
Yes it is returned with the container object https://github.com/rancher/rancher/pull/30634/files#diff-7db0a06f1c829c0dd9501d5a34504ab4094a0884a09c6ce671a6add4f1bffa5e

PennyScissors:
@superseb @paynejacob @codyrancher 
Did the following to test how this change will impact workloads created with the previous structure. 

1. Ran Rancher locally, master branch.

2. Deployed a workload in "local" cluster with the following env vars:

Edit YAML View:
```
- name: A
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: status.hostIP
- name: B
  value: $(A):5000
```
API UI:
![image](https://user-images.githubusercontent.com/8951271/103953439-0373d880-50f7-11eb-9999-c49b3e21202d.png)

3. Stopped Rancher, pulled in the commits with my changes, and ran Rancher again

4. The workload shows up, env vars are still present in the container.
API UI after pulling in my changes:
![image](https://user-images.githubusercontent.com/8951271/103953807-a9274780-50f7-11eb-9e99-b2dcb82ee9b4.png)

5. UI's environment variables drop down is broken (expected):
![image](https://user-images.githubusercontent.com/8951271/103954098-58641e80-50f8-11eb-9d3a-5bd60cfca759.png)

6. Redeployed workload and added a new variable.  Order was preserved and variables were present in the container and API UI.
```
- name: A
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: status.hostIP
- name: B
  value: $(A):5000
- name: C
  value: $(B)
```
![image](https://user-images.githubusercontent.com/8951271/103954385-fa840680-50f8-11eb-8ac0-a6cffbb0dc0d.png)

[...]

Issue #16148, opened 2018-10-16
Rancher UI environment variables interpolation order


gferreux:
**Rancher versions:**
rancher/rancher:v2.0.6

**Environment Template: Kubernetes**

**Steps to Reproduce:**
Rancher ui order (and reorder) environment variables as simple (value=x) first then special (that use valueFrom) at the end.
If you define a simple variable that use interpolation of a special variable like:

```
- name: MEMCACHED_HOST
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: status.hostIP
- name: MEMCACHED_URL
  value: $(MEMCACHED_HOST):5000
```

the order (that is necessary for the interpolation to work) is preserved if you Edit YAML, but when you Upgrade via the UI, the YAML will become:

```
- name: MEMCACHED_URL
  value: $(MEMCACHED_HOST):5000
- name: MEMCACHED_HOST
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: status.hostIP
```

**Results:**
The interpolation fail and the container will see: `MEMCACHED_URL="$(MEMCACHED_HOST):5000"` without interpolation

gz#9725
gz#10539
gz#16638

gz#16888

cjellick:
@StrongMonkey is this a UI issue or a backend catalog issue?

StrongMonkey:
@cjellick most likely a type mapper issue in workload api. we don't sort the map base on name.

ivanfilippov:
Looks like this was an issue in kubectl as well, but was fixed, see here: https://github.com/kubernetes/kubernetes/issues/40373

According to that issue, constant ordering of env variables is part of the kubernetes API spec.

matfiz:
I could also reproduce it, still an issue in 2.2.1

isikemre:
It's still an issue on v2.2.3.
After saving environment variables from the UI, the order changes.
Then k8s cannot use $(SOME_ENV_VAR) (because the it's decelerated above)

switchboardOp:
Still an issue in v2.2.4

digger18:
Will this ever be fixed?

junkiebev:
Still in issue on 2.3.2

cjellick:
I think it fell off our radar. Thanks for the bump. I'll ping the necessary ppl to try to get this prioritized.

cjellick:
@vincent99 @westlywright do you know of some technical reason why we have to reorder these?

vincent99:
[...]

PR #32415, opened 2021-04-27
[master] Deprecate container env vars fields in API

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/16148
UI issue: https://github.com/rancher/rancher/issues/30683

The fields `environment` and `environmentFrom` has been deprecated in favor of the k8s native types. This is due to our types not being able to preserve the order of the variables and break interpolation. I went for the validation approach instead of migrating them for the user because the migration could result in the wrong order (We are basically doing this already since we map these to the k8s types and vice versa and the order isn't preserved) and removed the mappers since then it won't be possible to use the k8s types directly as they get overwritten by ours. The expectation after this change is for users to use the k8s types instead.


pkg/api/norman/store/workload/workload_store.go
```diff
@@ -116,6 +118,22 @@ func (s *CustomizeStore) validateStatefulSetVolume(schema *types.Schema, data ma
...
+// validateNoDeprecatedEnvVars validates that the now deprecated fields `environment` and `environmentFrom` are not being used.
+// The fields were deprecated in favor of k8s native types for being unable to preserve the order of the variables and break interpolation.
+func (s *CustomizeStore) validateNoDeprecatedEnvVars(data map[string]interface{}) error {
+	if containers, _ := values.GetSlice(data, "containers"); len(containers) > 0 {
+		for _, c := range containers {
+			if !convert.IsAPIObjectEmpty(values.GetValueN(c, "environment")) {
+				return httperror.NewAPIError(httperror.InvalidOption, "field `environment` is deprecated, please use Kubernetes native types for the container's environment variables")
+			}
+			if !convert.IsAPIObjectEmpty(values.GetValueN(c, "environmentFrom")) {
+				return httperror.NewAPIError(httperror.InvalidOption, "field `environmentFrom` is deprecated, please use Kubernetes native types for the container's environment variables")
```
PennyScissors, line 130: I used `httperror.InvalidOption` (Code 422) for this because I couldn't find anything more suitabled [here](https://github.com/rancher/norman/blob/8e6ffc77a819d067456dcec666553daf929eba33/httperror/error.go#L8-L34), but please let me know if there is another code more suitable for this case and I'll update this.

StrongMonkey (commented):
Will the validator break if user has existing workload using environment or environmentFrom? Should we only validate if user is creating new workload?

PennyScissors:
@StrongMonkey The problem is that to be able to use the k8s native types we need to get rid of the mappers, and without the mappers any existing workload using environment and environmentFrom will break because it won't get converted to the k8s native types or vice versa. This is expected to be a breaking API change and will be documented in the release notes.

PennyScissors:
Expectations after deprecation:
Creating a new workload or modifying an existing:
* Creating a new workload with environment variables or modifying the variables of an existing one through the UI (key value text fields in Ember) should remain the same. However, the UI is currently using the deprecated types so it will go through the validation and fail until this issue is resolved. As for the edit as YAML view, the k8s native types are expected so no changes expected.
* Creating a new workload with environment variables or modifying the variables of an existing one through the API View will go through validation to make sure the k8s native types are used.
* When upgrading from v2.5 to v2.6 with existing workloads, the environment variables will not be mapped to our custom types anymore, and will instead show up in the API with the k8s native type.
* Workloads with environment variables created using the UI (key value text fields), and API View will go through validation and should fail if the deprecated fields are used. Those created using the UI(Edit as YAML) or kubectl will not go through the validation because we don't map those to our own types, instead the k8s native type is expected and k8s validates them.

Upgrade example:
Environment variables of workload created in v2.5
Screen Shot 2021-06-02 at 5 43 21 PM

Environment variables of the same workload after upgrading to v2.6
Screen Shot 2021-06-02 at 5 36 53 PM

pkg/api/norman/customization/workload/validator.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/rancher/norman/types/values"
+)
+
+// Validates deprecated fields `environment` and `environmentFrom` are not being used. These fields were deprecated in favor
+// of the k8s native types `env` and `envFrom` for being unable to preserve the order of the variables and break interpolation.
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if containers, _ := values.GetSlice(data, "containers"); len(containers) > 0 {
+		for _, c := range containers {
+			if !convert.IsAPIObjectEmpty(values.GetValueN(c, "environment")) {
+				return httperror.NewAPIError(httperror.InvalidBodyContent, "field `environment` is deprecated, please use Kubernetes native types for the container's environment variables")
```
kinarashah, line 16: could we update msg to say `please use Kubernetes native type env...`? and `envFrom` for L19

kinarashah (dismissed):
🎉

StrongMonkey (dismissed):

pkg/api/norman/customization/workload/validator.go
```diff
@@ -0,0 +1,24 @@
+package workload
+
+import (
+	"github.com/rancher/norman/httperror"
+	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/convert"
+	"github.com/rancher/norman/types/values"
+)
+
+// Validates deprecated fields `environment` and `environmentFrom` are not being used. These fields were deprecated in favor
```
dramich, line 10:
```suggestion
// Validator validates deprecated fields `environment` and `environmentFrom` are not being used. These fields were deprecated in favor
```

pkg/api/norman/customization/workload/validator.go
```diff
@@ -0,0 +1,24 @@
...
+// of the k8s native types `env` and `envFrom` for being unable to preserve the order of the variables and break interpolation.
```
dramich, line 11:
```suggestion
// of the k8s native types `env` and `envFrom`. See https://github.com/rancher/rancher/issues/16148
```

pkg/api/norman/customization/workload/validator.go
```diff
@@ -0,0 +1,24 @@
...
+func Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	if containers, _ := values.GetSlice(data, "containers"); len(containers) > 0 {
+		for _, c := range containers {
+			if !convert.IsAPIObjectEmpty(values.GetValueN(c, "environment")) {
```
dramich, line 15: Nice job using all the helpers to access data fields.

PennyScissors: Thanks!

dramich (approved):

PR #32429, opened 2021-04-28
Fix helm chart default values and documentation

pgdagenais:
Rancher default chart values are incorrect, since we are merging the values for the `extraAnnotations` it is not easy to overwrite the default one if we are using the upstream Rancher chart repository. Since we are not using the nginx ingress controller we do not need these values.

It also doesn't match the expected values from the default tab: https://github.com/rancher/rancher/blame/master/chart/README.md#L181

I also fixed an incorrect value in the readme tab. Which should match the configuration value in the deployment file: https://github.com/pgdagenais/rancher/blob/master/chart/templates/deployment.yaml#L94

Thank you very much!


pgdagenais:
Related issue: https://github.com/rancher/rancher/issues/32773

dramich:
> ... Since we are not using the nginx ingress controller we do not need these values.

@pgdagenais I don't really understand the change. If you are not using the nginx ingress, meaning you set `enabled: false` then those annotations should not affect your ingress. This change also drops the defaults for anyone running with it set to true. 
I think the README defaults need to be updated to include these values vs commenting them out in the values.yaml

pgdagenais:
Sorry, what I meant is that we do not use the nginx ingress-controller on our infra (we use HAproxy) so we still need the ingress and since helm values are additive, this will configure both annotations. 

For example:

```
$ cat test-values.yaml
ingress:
  enabled: true
  extraAnnotations:
    kubernetes.io/ingress.class: "haproxy"
    ingress.kubernetes.io/secure-backends: "false"
    ingress.kubernetes.io/maxconn-server: "10000"
  tls:
    source: secret
$ helm3 template . -s templates/ingress.yaml -f test-values.yaml
---
# Source: rancher/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: RELEASE-NAME-rancher
  labels:
    app: RELEASE-NAME-rancher
    chart: rancher-1.0.0
    heritage: Helm
    release: RELEASE-NAME
  annotations:
    ingress.kubernetes.io/maxconn-server: "10000"
    ingress.kubernetes.io/secure-backends: "false"
    kubernetes.io/ingress.class: haproxy
    nginx.ingress.kubernetes.io/proxy-connect-timeout: "30"
    nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
    nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
spec:
  rules:
  - host:   # hostname to access rancher server
    http:
      paths:
      - backend:
          serviceName: RELEASE-NAME-rancher
          servicePort: 80
  tls:
  - hosts:
    -
    secretName: tls-rancher-ingress
```

I do agree this cause an issue with the users using the default values. Not sure how we can solve that. Maybe this is something that could be released with Rancher 2.6? But I don't know.

dramich:
OK I follow. I think moving the defaults to their own block and then making an option to turn them off would solve the issue. Something like `includeDefaultExtraAnnotations: true` then in the chart do an if check to include them

pgdagenais:
@dramich Good idea!

I also added a new test to validate that new configuration value.

Thank you very much

aiyengar2 (approved):
LGTM

dramich (approved):

dramich:
@pgdagenais Can you update your initial PR description to reflect our changes?

pgdagenais:
@dramich Sure! Just updated it.

Thanks!

dramich:
https://github.com/rancher/rancher/issues/33254

PR #32432, opened 2021-04-28
Switching windows images to the rancher hub account

luthermonson:
looks like drone isn't going to get their build hosts any time soon, moving the images I made for 2004/20H2 from my personal docker up to rancher's.


superseb (approved):
LETS GOOOOOOO

Issue #32436, opened 2021-04-28
rancher-post-delete workload missing `/` in image, causes fail to pull image


slickwarren:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
* setup an airgap HA cluster (private registry, offline nodes, etc.)
* use RKE to configure the HA cluster (latest RC is 1.2.8-rc4)
* install rancher/certs on the cluster

**Result:**
* when viewing the system project or kubectl, one of the pods fails, due to a missing slash before `rancher`:
`...spacerancher/shell:v0.1.6`

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.8-rc12
- Installation option (single install/HA): Airgap HA

slickwarren:
on v2.5.8-rc17 in an airgapped setup, I was able to successfully deploy rancher.  `rancher-post-delete` is successful:
```
cattle-system             rancher-post-delete-wn6qt                  0/1     Completed   0          6m33s
```

<img width="1536" alt="Screen Shot 2021-05-03 at 10 13 06 AM" src="https://user-images.githubusercontent.com/16691014/116908848-3c7d0f00-abf8-11eb-8599-0b769c09f8fd.png">
opening forwardport for 2.6 (need rc)

PR #32438, opened 2021-04-28
[2.5.8] handle the system default registry in post-delete-hook-job better

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/32436

Add a template function to better handle the system default registry in post-delete-hook-job 

Note:
A forward PR to rancher/master will be made after this PR is approved.

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -33,3 +33,13 @@ chart: {{ .Chart.Name }}-{{ .Chart.Version }}
 heritage: {{ .Release.Service }}
 release: {{ .Release.Name }}
 {{- end }}
+
+{{- define "system_default_registry" -}}
+{{- if .Values.systemDefaultRegistry -}}
+  {{- if hasSuffix "/" .Values.systemDefaultRegistry -}}
+    {{- printf "%s" .Values.systemDefaultRegistry -}}
+  {{- else -}}
+    {{- printf "%s/" .Values.systemDefaultRegistry -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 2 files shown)


chart/templates/_helpers.tpl
```diff
@@ -33,3 +33,9 @@ chart: {{ .Chart.Name }}-{{ .Chart.Version }}
 heritage: {{ .Release.Service }}
 release: {{ .Release.Name }}
 {{- end }}
+
+{{- define "rancher.systemDefaultRegistry" -}}
```
cbron, line 37: Either: 
```suggestion
{{- define "system_default_registry" -}}
```
or
```suggestion
{{- define "rancher.system_default_registry" -}}
```
I don't know why we put `rancher.` in there

chart/templates/_helpers.tpl
```diff
@@ -33,3 +33,9 @@ chart: {{ .Chart.Name }}-{{ .Chart.Version }}
...
+{{- if .Values.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.systemDefaultRegistry -}}
```
cbron, line 39: Since this is user input, ideally we would check for slash somehow, they might add it. Whereas other charts it gets added automatically.

cbron (changes requested):

chart/templates/_helpers.tpl
jiaqiluo: I was trying to follow what the [Helm documentation](https://helm.sh/docs/chart_template_guide/named_templates/) recommends:

> By using the specific chart name as a prefix we can avoid any conflicts that may arise due to two different charts that implement templates of the same name.

But since no conflict will happen here, I changed it to `system_default_registry`

chart/templates/_helpers.tpl
jiaqiluo: the logic is updated to handle this case.

cbron (approved):
lgtm, wait for 1 more to merge

nickgerace (approved):

cbron (approved):

PR #32470, opened 2021-04-29
Add context to Ready condition while updating

cmurphy:
When zonal GKE clusters are updating the control plane, the control
plane usually becomes unreachable. This is expected and normal
behavior[1]. However, when the HealthSyncer controller runs, it cannot
reach the API and sets ClusterConditionReady to false with an alarming
error message. This change does not eliminate or ignore the error but
adds more context to the error message to help users understand that
this is normal.

[1] https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#upgrading_manually

https://github.com/rancher/rancher/issues/32473

### DIFF
--- a/pkg/controllers/management/gke/gke_cluster_handler.go
+++ b/pkg/controllers/management/gke/gke_cluster_handler.go
@@ -246,6 +246,18 @@ func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		e.ClusterEnqueueAfter(cluster.Name, enqueueTime)
 		if failureMessage == "" {
 			logrus.Infof("waiting for cluster GKE [%s] to update", cluster.Name)
+
+			// If the HealthSyncer runs while upgrading a zonal cluster, the control plane may not be reachable.
+			// This adds additional context to the error message to help explain that this is normal.
+			readyMsg := apimgmtv3.ClusterConditionReady.GetMessage(cluster)
+			helpMsg := ": control plane may be unavailable while it is being upgraded"
+			if apimgmtv3.ClusterConditionReady.IsFalse(cluster) && strings.Contains(readyMsg, "connect: connection refused") && !strings.Contains(readyMsg, helpMsg) {
+				msg := apimgmtv3.ClusterConditionReady.GetMessage(cluster) + helpMsg
+				// return here; ClusterConditionUpdated is most likely already set, and
+				// if not will be set on the next loop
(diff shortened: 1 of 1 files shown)


cmurphy:
Example of the UI change 
image

It's still red and still in "Error" but if you read closely there's some more information

thedadams (approved):

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -246,6 +246,18 @@ func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		e.ClusterEnqueueAfter(cluster.Name, enqueueTime)
 		if failureMessage == "" {
 			logrus.Infof("waiting for cluster GKE [%s] to update", cluster.Name)
+
+			// If the HealthSyncer runs while upgrading a zonal cluster, the control plane may not be reachable.
+			// This adds additional context to the error message to help explain that this is normal.
+			readyMsg := apimgmtv3.ClusterConditionReady.GetMessage(cluster)
```
rmweir, line 252: Would this message not be found on the updating condition?

rmweir (changes requested):
Are there things we can easily check that would help us determine if it is going through an upgrade?

pkg/controllers/management/gke/gke_cluster_handler.go
cmurphy: No, it's only the Ready condition that gets the health check message set

https://github.com/rancher/rancher/blob/f5ff8e1fbe76e63d2197185c0d7adda72bb1db70/pkg/controllers/managementuser/healthsyncer/healthsyncer.go#L118
https://github.com/rancher/rancher/blob/f5ff8e1fbe76e63d2197185c0d7adda72bb1db70/pkg/controllers/managementuser/healthsyncer/healthsyncer.go#L80

cmurphy:
> Are there things we can easily check that would help us determine if it is going through an upgrade?

It is checking the gkeclusterconfig phase and only setting this condition during the "updating" phase.

cbron (approved):
Downside of this is if a cluster really does go down when updating the user may think its ok. But this is a GKE trait, so not much else we can do but tell people that.

rmweir (approved):
LGTM. My concern was that just checking if it's updating may not be specific enough and appending that message if it isn't upgrading but updating something else may be misleading. This is probably fine as it is assuming the aforementioned doesn't happen often.

cmurphy:
forward port https://github.com/rancher/rancher/pull/32497

Issue #31375, opened 2021-02-18
[rancherd] Bump RKE2 version - 1.20


davidnuzik:
Once the Rancher team is ready, we should plan to update rancherd to RKE2 version 1.20.x (latest stable).

davidnuzik:
Related issue for 1.19.x: https://github.com/rancher/rancher/issues/30492 - this issue has more context and and example covering the processs if you look at the linked PR and comments.

davidnuzik:
Utilize v1.20.5+k3s1 now instead of v1.20.4+k3s1 due to a CVE called out in v1.20.5 k3s release notes.

bmdepesa:
Rancherd is failing to start. It looks to be using the `rke2-runtime:latest` tag instead of `rke2-runtime:v1.20.6-rke2r1`

```
> curl -sfL https://get.rancher.io | sudo INSTALL_RANCHERD_VERSION=v2.5.8-rc12 sh -
> systemctl enable rancherd-server.service && systemctl start rancherd-server.service
> journalctl -eu rancherd-server -f
...
rancherd[2608]: time="2021-04-28T16:43:49Z" level=fatal msg="Runtime image index.docker.io/rancher/rke2-runtime:latest is not a not a reference to a digest or version tag matching pattern ^v[0-9]"
...
```

bmdepesa:
rancher/rancher:v2.5.8-rc15

Testing manually with the possible workaround we see another issue where the `rancher` and `rancher-namespace` manifests are not included in the RancherD installation, and Rancher never deploys. 

In v2.5.8-rc15 RancherD installation:
```
/var/lib/rancher/rke2/server/manifests$ ls
rke2-canal.yaml  rke2-coredns.yaml  rke2-ingress-nginx.yaml  rke2-kube-proxy.yaml  rke2-metrics-server.yaml
```

vs. v2.5.7 installation:
```
/var/lib/rancher/rke2/server/manifests# ls
rancher-namespace.yaml  rke2-canal.yaml    rke2-ingress-nginx.yaml  rke2-metrics-server.yaml
rancher.yaml            rke2-coredns.yaml  rke2-kube-proxy.yaml
```

rancher-max:
Validated rancherd is using v1.20.5+rke2r1 and installs correctly in both airgap and not.

PR #32494, opened 2021-04-30
Correct runtime image for rancherd

thedadams:
There was some mixups with the RKE2 changes that caused the runtime
image to be set incorrectly.

Also, there were changes in RKE2 that caused problems with containerd
not being able to find the $HOME environment variable. That is also
fixed here.

Issue:
https://github.com/rancher/rancher/issues/31375

### DIFF
--- a/cmd/rancherd/bundle/lib/systemd/system/rancherd-agent.env
+++ b/cmd/rancherd/bundle/lib/systemd/system/rancherd-agent.env
@@ -1,0 +1,1 @@
+HOME=/root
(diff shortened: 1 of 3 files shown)


cbron (approved):
Don't fully understand whats going on here but it seems ok. Wait for @brandond.

brandond (approved):

brandond:
RancherD wraps the rke2-runtime image with some extra charts. You need to use your runtime image with your build tag, not the rke2 image, otherwise you're missing out on all the special sauce.

Issue #32520, opened 2021-05-04
v2 cluster management has cluster greyed out even after they are active


anupama2501:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Spin up rancher on the latest master-head
2. Create a cluster infrastructure provider/import/custom form cluster explorer --> cluster management 
3. On the tope left corner, click on the three vertical lines next to local and under local, all the clusters are greyed out even when they are active

**Result:**
Clusters should not be greyed out and users should be able to access as for local cluster

**Other details that may be helpful:**
<img width="300" alt="Screen Shot 2021-05-03 at 6 27 43 PM" src="https://user-images.githubusercontent.com/38144301/116951670-555be380-ac3d-11eb-8248-fb81e2dd707d.png">

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 
master-90dc82bfc7ff26f3b3f8922595d491cbf8c9d3dd-head
- Installation option (single install/HA): Single

slickwarren:
still seeing this on master-head (91a6f51)
<img width="266" alt="Screen Shot 2021-05-04 at 4 48 53 PM" src="https://user-images.githubusercontent.com/16691014/117083082-b1377280-acf8-11eb-9ebe-e07078100433.png">

anupama2501:
Verified on master-360edaec8f7e236611fdd30e6798874bac267d5b-head
1. Spin up rancher on the version of master-head mentioned above.
2. Create a custom cluster and on an AWS node run the registration command with all roles enabled.
3. Once the node gets active, from the explore cluster --> click on the cluster created above and is redirected to the cluster dashboard
4. Create a DO cluster 1 node all roles node driver, once it gets active, repeated step3

PR #32524, opened 2021-05-04
tls call is blocking, move watch before

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/32520

Issue #32588, opened 2021-05-06
Add username at login to Rancher API Audit log at metadata (level 1)


prachidamle:
Rancher API Audit log have been enhanced by this fix https://github.com/rancher/rancher/issues/31166 to log username and principalid data for the authenticated user for every request that gets logged at Metadata level (audit log level=1).

This is the "extra" that is logged for an authenticated user with above fix:
"extra":{"principalid":["openldap_user://uid=testuser1,ou=dev,dc=us-west-2,dc=compute,dc=internal"],"username":["testuser1"]}}

But when there is a login  failure,  there is no username logged at Metadata level, since there is no authenticated user entity present. (If the Audit level is set to 2 and above, entire request body is logged, but it is beneficial to log the username at Metadata level so that service providers can track login failures easily)

This enhancement is to log `extra.username` in the Rancher API audit logs at Metadata level for a Login request - the username field of the login form is read and logged in the Rancher API audit log.

This is supported only for authproviders which work with a username/password form of login like AD,LDAP, local. For other token/SAML based providers there will be no username logged for login resuests at Metadata level.

gz#7326

prachidamle:
This is how the extra.username gets added to Rancher API audit log at level 1

[...]

PR #32603, opened 2021-05-06
Add username from login request to rancher api audit log at level 1 (metadata) and above

prachidamle:
https://github.com/rancher/rancher/issues/32588

This enhancement is to log extra.username in the Rancher API audit logs at Metadata level for a Login request - the username field of the login form is read and logged in the Rancher API audit log.

This is supported only for authproviders which work with a username/password form of login like AD,LDAP, local. For other token/SAML based providers there will be no username logged for login requests at Metadata level.

### DIFF
--- a/pkg/auth/audit/audit.go
+++ b/pkg/auth/audit/audit.go
@@ -6,10 +6,13 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/pborman/uuid"
 	"github.com/pkg/errors"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apiserver/pkg/endpoints/request"
 )
@@ -53,6 +56,7 @@ type log struct {
 	ResponseHeader    http.Header  `json:"responseHeader,omitempty"`
 	RequestBody       []byte       `json:"requestBody,omitempty"`
 	ResponseBody      []byte       `json:"responseBody,omitempty"`
+	UserLoginName     string       `json:"userLoginName,omitempty"`
 }
 
 var userKey struct{}
@@ -76,6 +80,16 @@ func getUserInfo(req *http.Request) *User {
 	}
 }
 
+func getUserNameForBasicLogin(body []byte) string {
+	input := &v32.BasicLogin{}
+	err := json.Unmarshal(body, input)
+	if err != nil {
+		logrus.Debugf("error unmarshalling input, cannot add login info to audit log: %v", err)
+		return ""
+	}
+	return input.Username
+}
+
 func FromContext(ctx context.Context) (*User, bool) {
 	u, ok := ctx.Value(userKey).(*User)
 	return u, ok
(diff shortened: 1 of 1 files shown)


pkg/auth/audit/audit.go
```diff
@@ -76,6 +80,16 @@ func getUserInfo(req *http.Request) *User {
 	}
 }
 
+func getUserNameForBasicLogin(body []byte) (string, bool) {
+	input := &v32.BasicLogin{}
+	err := json.Unmarshal(body, input)
+	if err != nil {
+		logrus.Debugf("Not a BasicLogin input, cannot add login info to audit log: %v", err)
```
dramich, line 87: I don't think we are going to see an error here, the struct will just continue to be empty if no fields match from the input.

dramich (changes requested):

pkg/auth/audit/audit.go
```diff
@@ -94,12 +108,23 @@ func newAuditLog(writer *LogWriter, req *http.Request) (*auditLog, error) {
...
+			if loginReq {
+				loginName := getUserNameForBasicLogin(reqBody)
+				if loginName != "" {
+					auditLog.log.UserLoginName = loginName
+				}
+			}
+			if writer.Level >= levelRequest {
+				auditLog.reqBody = reqBody
+			}
 		}
```
aiyengar2, line 127: can we restructure this? e.g.
```go
	if exists, ok := bodyMethods[req.Method]; !exists || !ok {
		return auditLog, nil
	}
	if contentType != contentTypeJSON {
		return auditLog, nil
	}
	if writer.Level >= levelRequest {
                var err error
		auditLog.reqBody, err = readBodyWithoutLosingContent(req)
		if err != nil {
			return nil, err
		}
	}
	loginReq := isLoginRequest(req.RequestURI)
	if loginReq {
		auditLog.log.UserLoginName = getUserNameForBasicLogin(reqBody)
	}
	return auditLog, nil
```

pkg/auth/audit/audit.go
```diff
@@ -110,6 +135,13 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
 	a.log.RequestHeader = filterOutHeaders(reqHeaders, sensitiveRequestHeader)
 	a.log.ResponseHeader = filterOutHeaders(resHeaders, sensitiveResponseHeader)
 	a.log.ResponseCode = resCode
+	if a.log.UserLoginName != "" {
+		if a.log.User.Extra == nil {
+			a.log.User.Extra = make(map[string][]string)
+		}
+		a.log.User.Extra["username"] = []string{a.log.UserLoginName}
+		logrus.Debugf("Added username for login request to audit log %v", a.log.User)
```
aiyengar2, line 143: Do we want to print out the whole User object or just the added username + UserLoginName?

aiyengar2 (changes requested):

pkg/auth/audit/audit.go
prachidamle: no,  since we need to readBodyWithoutLosingContent in both cases  when it is  levelRequest || loginReq, and we need to ensure the UserLoginName is set only for loginReq irrespective of the writer.Level

pkg/auth/audit/audit.go
```diff
@@ -139,6 +171,10 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
 	return err
 }
 
+func isLoginRequest(uri string) bool {
+	return strings.Contains(uri, "?action=login")
```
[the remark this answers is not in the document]
prachidamle: we were using it before in two places, so it was a helper method. Nevertheless I think it makes the code readable.

pkg/auth/audit/audit.go
prachidamle: changed the debug log to log the UserLoginName

dramich (approved):

aiyengar2 (approved):
LGTM

PR #32649, opened 2021-05-10
fix dropped gke cluster error

alrs:
This fixes a dropped `err` variable in `pkg/controllers/management/gke`.


dramich:
Close as this was addressed in https://github.com/rancher/rancher/commit/c534673723195e8d1565ea1326ffddc2dc1e74ef

Issue #32304, opened 2021-04-23
Nodes do not get deleted in the backend during cluster creation when nodepools have same name


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy a node driver cluster with 2 nodepools with same name
- nodepool 1 --> etcd/control plane
- nodepool 2 --> Worker node roles - 1 node each
- One or both the nodes get deleted and recreated from Rancher 
- But the node does not get deleted in the backend. (After the cluster comes up Active)

**Other details that may be helpful:**

- Happens on 2.5.7 also
- Does not happen on 2.5.5

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.8-rc8
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): RKE DO node driver
- Kubernetes version (use `kubectl version`):

```
1.20 (Any)
```

sowmyav27:
**See this issue on master-head commit id: `5d49b92`**

- Deploy a DO RKE1 cluster from Ember UI with 2 nodepools
```
sowmya-issue-1 --> etcd/control plane - 1 node
sowmya-issue-1 --> worker role - 1 node
```

- Worker node gets the IP and master node fails with error `Error creating machine: Error detecting OS: Too many retries waiting for SSH to be available. Last error: Maximum number of retries (60) exceeded; lstat /opt/jail/c-hv456/management-state/node/nodes/sowmya-issue-1: no such file or directory` and re provisions.
- When it is re-provisioning the worker node is deleted in the backend instead of the control plane node. (both have the same nodename)
- Worker node is seen stuck in Registering state in Rancher UI

**Expected Result:**
Worker node should not be deleted and the master node should be from the backend.

**Note:**
- Happens always with ^^^ these steps
- See the same issue on 2.5.5

Tracked in this issue - https://github.com/rancher/rancher/issues/32302

sowmyav27:
Reopening to see if this issue is fixed by this PR https://github.com/rancher/rancher/pull/32662 as the related PR is for a Custom cluster.

sowmyav27:
[...]

PR #32662, opened 2021-05-10
Custom RKE1 Node Cleanup

luthermonson:
Problem: When creating a custom RKE1 cluster the expectation was to take a fresh server with docker, copy/paste a docker run command and add a new node to a cluster. However, if the node was removed from the cluster via the Rancher UI a node remove call is made to the kubernetes api but we would leave RKE1 tooling around on the node and mangled networking/firewall configs. This led to a lot of work for support to create cleanup scripts to undo the mess. 

Solution: Upon v3 Node delete in rancher we will pause the v1 node delete until we leave a cleanup process on the deleting node. We will target a k8s job to the node with the docker socket mounted who then starts a cattle-node-cleanup container. This container is outside of k8s and won't be reaped when the v1 node is deleted and kubelet is told to shutdown. The new cleanup process will pause and wait for kubelet to remove all k8s_ containers and when that's completed it will proceed to wipe everything from the server (containers, directories, networking, firewall rules, etc). This now means you can add a node with the docker command and delete it through the UI and immediately add it back to the cluster with no further intervention.

There are also utility methods now build into the agent to allow you to get a cleanup script from the agent container and run it yourself by hand this way we can document the cleanup process using only the agent image. This solution works on both custom linux and windows nodes.

Additional: This PR addresses the old docker cli version we had in the agent, we used it for the share-mnt to patch ephemeral /etc directories in things like RancherOS. This removes the docker cli and makes docker go client calls instead within the agent.

https://github.com/rancher/rancher/issues/26545
https://github.com/rancher/rancher/issues/32304
https://github.com/rancher/rancher/issues/32052


pkg/agent/clean/node_notwindows.go
```diff
@@ -0,0 +1,590 @@
...
+const (
+	DockerPipe = "/var/run/docker.sock"
+	HostMount  = "/host"
+)
+
+func Run(ctx context.Context, args []string) error {
+	if len(args) > 3 {
+		fmt.Println(usage())
+		return nil
+	}
```
aiyengar2, line 34: Any reason why we aren't returning an error?

aiyengar2: also, if `len(args) < 3`, why is `Node(ctx)` (cleaning?) the default behavior? Seems like a weird thing to do if the args are `agent job`.

pkg/controllers/management/node/cleanup.go
```diff
@@ -0,0 +1,372 @@
...
+			return nil // gave it 30s and it timed out so move to delete v1 node
+		}
+		return err
+	}
+
+	// remove the job to clean up
+	return userContext.K8sClient.BatchV1().Jobs(job.Namespace).Delete(m.ctx, job.Name, metav1.DeleteOptions{})
+}
+
+func (m *Lifecycle) createCleanUpJob(userContext *config.UserContext, node *v3.Node) (*batchV1.Job, error) {
```
aiyengar2, line 193: nit: cleanup is one word.
```suggestion
func (m *Lifecycle) createCleanupJob(userContext *config.UserContext, node *v3.Node) (*batchV1.Job, error) {
```

aiyengar2 (changes requested):
Mostly looks good! Just a couple of minor comments.

pkg/agent/clean/node_notwindows.go
luthermonson: cause it prints the usage and assumes you dont know what youre doing? possible usage should be the default behavior... let me thing about this.

pkg/controllers/management/node/cleanup.go
luthermonson: how does one give dagger eyes on this thing?

aiyengar2 (dismissed):
LGTM

phillipsj (dismissed):

aiyengar2 (approved):
LGTM

pkg/rkeworker/execute.go
```diff
@@ -29,9 +30,11 @@ func ExecutePlan(ctx context.Context, nodeConfig *NodeConfig, writeCertOnly bool
 
 	f := fileWriter{}
 	for _, file := range nodeConfig.Files {
+		logrus.Infof("writing file %s", file.Name)
 		f.write(file.Name, file.Contents)
 	}
 	if writeCertOnly {
+		logrus.Info("writing certificates only, no need to continue executing plan")
```
dramich, line 37: Can we move these to debug?

pkg/agent/clean/node_notwindows.go
```diff
@@ -0,0 +1,591 @@
...
+	node - cleans the entire node and performs all actions below and is the default command, requires docker socket to clean up dockerd
+	links - clean the interface links created on the host, requires --privileged and --network host
+	paths - clean the paths added by the host like /etc/rancher, /var/lib/rancher, etc. requires -v /:/host 
+	firewall - clean the iptables firewall rules, needs --privileged and --network host to work 
+	script - prints a bash script you can run to clean up the node from the cli
+	help - print this help message
+
+other commands for automation:
+	job - used by the k8s batch job to start the ` + NodeCleanupContainerName + ` container to watch kubelet cleanup and wait to clean the node
+`
```
dramich, line 415: These will all print out indented since there is a tab at the beginning.

dramich (approved):
Couple small comments, looking forward to having some clean nodes

dramich (approved):

pkg/agent/clean/node_notwindows.go
luthermonson: good point.

dramich (approved):

dramich (approved):

kinarashah:
https://github.com/rancher/rancher/pull/32899

PR #32700, opened 2021-05-12
Add volume variable to rke2

rancher-max:


tests/validation/tests/v3_api/test_import_rke2_cluster.py
```diff
@@ -25,7 +25,7 @@
 RANCHER_RKE2_SERVER_FLAGS = os.environ.get("RANCHER_RKE2_SERVER_FLAGS", "server")
 RANCHER_RKE2_WORKER_FLAGS = os.environ.get("RANCHER_RKE2_WORKER_FLAGS", "agent")
 RANCHER_RKE2_OPERATING_SYSTEM = os.environ.get("RANCHER_RKE2_OPERATING_SYSTEM")
-
+RANCHER_VOLUME_SIZE = os.environ.get("RANCHER_VOLUME_SIZE", "20")
```
ShylajaDevadiga, line 28: Let's keep default to disk size to minimum 8G and depending on the need we could use a larger size.

rancher-max: rke2 requires default minimum of 20

rancher-max: Leaving rke2 at 20 but updated k3s to use 8

ShylajaDevadiga (approved):
LGTM

bmdepesa (approved):
LGTM

Issue #31690, opened 2021-03-13
[Backport] exec shell on GKE cluster fails with 1006 error


sowmyav27:
Backport of https://github.com/rancher/rancher/issues/31434

thedadams:
A related problem to this is the "Launch kubectl" button does not work an hour after the cluster is created.

slickwarren:
currently blocked by https://github.com/rancher/rancher/issues/33241

slickwarren:
on v2.4-5860-head:
executing into the shell both immediately and after 1+hours is now working as expected

PR #32716, opened 2021-05-13
Add Google OAuth to proxy

thedadams:
The removal of the Google basic auth means that the service account
token expires after 60 minutes. This was fixed in other areas, but the
proxy still used the service account token on the cluster object instead
of getting one from the token source. Getting the token from the token
source gets a renewed token that is usable for 60 minutes.

Issue:
https://github.com/rancher/rancher/issues/31690

### DIFF
--- a/pkg/clusterrouter/proxy/proxy_server.go
+++ b/pkg/clusterrouter/proxy/proxy_server.go
@@ -1,6 +1,7 @@
 package proxy
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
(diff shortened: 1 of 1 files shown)


cmurphy (approved):

rmweir (approved):
LGTM

pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -123,6 +125,19 @@ func NewRemote(cluster *v3.Cluster, clusterLister v3.ClusterLister, factory dial
...
+			if err != nil {
+				return "", fmt.Errorf("unable to retrieve token source for GKE oauth2: %v", err)
+			}
+			token, err := tokenSource.Token()
+			if err != nil {
+				return "", fmt.Errorf("unable to retrieve token from token source for GKE oauth2: %v", err)
+			}
+			return "Bearer " + token.AccessToken, nil
+		}
+
```
cbron, line 140: Curious on the "usable for 60 minutes" part, will it bypass this method in the future and reuse that token ? Or does every api call come in here and generate a new one.

thedadams: The `TokenSource` only retrieves a new token if it is necessary.

cbron (approved):

Issue #32726, opened 2021-05-13
Single worker node fails to register to kubernetes cluster - DO


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Rancher HA RKE1 2.6 _351f7ca_ 
- Downstream RKE1 DO cluster provisioned (Ember or Dashboard)

Worker node should be 1 in its own node pool
Other nodes should not have the worker role.

Examples
- Nodepool `#1` - With 1 etcd node
- Nodepool `#2` - With 1 control plane node
- Nodepool `#3` - With 1 worker node

or:

- Nodepool `#1` - With 1 node of etcd and control roles.
- Nodepool `#2` - With 1 worker node

**Result:**

Worker node is stuck provisioning

**Other details that may be helpful:**

- RKE2 feature disabled

```
docker logs -f competent_fermi


INFO: Arguments: --server https://<REDACTED> --token REDACTED -r -n m-95j7x
INFO: Environment: CATTLE_ADDRESS=<REDACTED> CATTLE_AGENT_CONNECT=true CATTLE_INTERNAL_ADDRESS= CATTLE_NODE_NAME=m-95j7x CATTLE_SERVER=https://izmle1.qa.rancher.space CATTLE_TOKEN=REDACTED
INFO: Using resolv.conf: nameserver <REDACTED> nameserver <REDACTED>
INFO: https://<REDACTED>/ping is accessible
INFO: <REDACTED> resolves to <REDACTED>
time="2021-05-13T20:30:01Z" level=warning msg="failed to reconcile kubelet, error: Error response from daemon: client version 1.41 is too new. Maximum supported API version is 1.40"
time="2021-05-13T20:30:01Z" level=info msg="Listening on /tmp/log.sock"
time="2021-05-13T20:30:01Z" level=info msg="Rancher agent version 351f7ca04 is starting"
time="2021-05-13T20:30:01Z" level=info msg="Option worker=false"
time="2021-05-13T20:30:01Z" level=info msg="Option requestedHostname=m-95j7x"
time="2021-05-13T20:30:01Z" level=info msg="Option customConfig=map[address:<REDACTED> internalAddress: label:map[] roles:[] taints:[]]"
time="2021-05-13T20:30:01Z" level=info msg="Option etcd=false"
time="2021-05-13T20:30:01Z" level=info msg="Option controlPlane=false"
[...]

PR #32747, opened 2021-05-14
use docker client with api version negotiation

kinarashah:
https://github.com/rancher/rancher/issues/32726

### DIFF
--- a/pkg/agent/node/node.go
+++ b/pkg/agent/node/node.go
@@ -32,6 +32,19 @@ func Params() map[string]interface{} {
 		"requestedHostname": os.Getenv("CATTLE_NODE_NAME"),
 	}
 
+	dclient, err := client.NewClientWithOpts(client.WithAPIVersionNegotiation(), client.FromEnv)
+	if err != nil {
+		logrus.Errorf("Error getting docker client: %v", err)
+	} else {
+		defer dclient.Close()
+		info, err := dclient.Info(context.Background())
+		if err != nil {
+			logrus.Errorf("Error getting docker info: %v", err)
+		} else {
+			params["dockerInfo"] = info
+		}
+	}
+
 	for k, v := range params {
 		if m, ok := v.(map[string]string); ok {
 			for k, v := range m {
@@ -39,15 +52,6 @@ func Params() map[string]interface{} {
 			}
 		} else {
 			logrus.Infof("Option %s=%v", k, v)
-		}
-	}
-
-	dclient, err := client.NewEnvClient()
-	if err == nil {
-		defer dclient.Close()
-		info, err := dclient.Info(context.Background())
-		if err == nil {
-			params["dockerInfo"] = info
 		}
 	}
 
--- a/pkg/tunnelserver/mcmauthorizer/tunnel.go
+++ b/pkg/tunnelserver/mcmauthorizer/tunnel.go
@@ -208,6 +208,7 @@ func (t *Authorizer) authorizeNode(register bool, cluster *v3.Cluster, inNode *c
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/agent/node/node.go
```diff
@@ -32,17 +32,15 @@ func Params() map[string]interface{} {
...
-	for k, v := range params {
-		if m, ok := v.(map[string]string); ok {
-			for k, v := range m {
-				logrus.Infof("Option %s=%s", k, v)
-			}
-		} else {
-			logrus.Infof("Option %s=%v", k, v)
+	dclient, err := client.NewClientWithOpts(client.WithAPIVersionNegotiation())
+	if err != nil {
+		logrus.Errorf("Error getting client with API version negotiation, using default client from Env: %v", err)
```
superseb, line 37: Can we use Docker client here to make sure everyone gets what is meant by client?

pkg/agent/node/node.go
```diff
@@ -32,17 +32,15 @@ func Params() map[string]interface{} {
...
+		dclient, err = client.NewClientWithOpts(client.FromEnv)
+		if err != nil {
+			logrus.Errorf("Error getting default client from Env: %v", err)
 		}
 	}
```
superseb, line 43: Can we log the retrieved Docker info in trace so we can compare the local value with the remote value (if something goes wrong here)

kinarashah: I moved the params logging (https://github.com/rancher/rancher/pull/32747/files#diff-93fe6e130b2e58fb15b3ec4070b80b9e71e37b443b225179c32dc7a1f1cd864dR52) after [setting dockerInfo in params](https://github.com/rancher/rancher/pull/32747/files#diff-93fe6e130b2e58fb15b3ec4070b80b9e71e37b443b225179c32dc7a1f1cd864dL50) so now it'll get logged similar to other options

StrongMonkey: Can we use client.NewClientWithOpts(client.WithAPIVersionNegotiation(), client.FromEnv)? By doing this it won't break if user is using the env variables for their docker socket.

kinarashah: Yeah @superseb and I did discussed that, but ugh now I don't remember why I skipped it. Adding FromEnv there as well

pkg/agent/node/node.go
```diff
@@ -32,17 +32,15 @@ func Params() map[string]interface{} {
...
-	for k, v := range params {
-		if m, ok := v.(map[string]string); ok {
-			for k, v := range m {
-				logrus.Infof("Option %s=%s", k, v)
-			}
-		} else {
-			logrus.Infof("Option %s=%v", k, v)
+	dclient, err := client.NewClientWithOpts(client.WithAPIVersionNegotiation(), client.FromEnv)
+	if err != nil {
+		logrus.Errorf("Error getting docker client with API version negotiation, using default client from Env: %v", err)
```
StrongMonkey, line 37: Hmm, do we still need to do fallback?

kinarashah: I think so, because ClientWithOpts errors out if the first option fails https://github.com/moby/moby/blob/master/client/client.go#L133

StrongMonkey: But client.WithAPIVersionNegotiation() doesn't return an error...

kinarashah: oh, hm that's weird...so ok it'll always get a valid api version, that will work - removing the default client

StrongMonkey (approved):

superseb (dismissed):

superseb:
@kinarashah I killed the build after `80:46:36`, you need to rebase to remove builds for Windows versions 1903 and 1909 I believe.

kinarashah:
Rebased, Thank you @superseb!

superseb (approved):

PR #32749, opened 2021-05-14
Generate a random password when rancher starts up

StrongMonkey:
Adds ability to specify bootstrap password when rancher start. This adds a new Environmental variable `CATTLE_BOOTSTRAP_PASSWORD`.

1. For single docker install, user can specify password by passing `-e CATTLE_BOOTSTRAP_PASSWORD=password` to docker. If it is not specified, the password will generated randomly and user needs to grap it from docker logs.

2. For helm HA install, user can specify that with .Values.bootstrapPassword. The password will be stored in k8s secrets and there is instruction of how to retrieve password on note.txt after rancher is installed. Same that if user don't specify in values.yaml, it will be randomly generated.

### DIFF
--- a/chart/templates/NOTES.txt
+++ b/chart/templates/NOTES.txt
@@ -6,4 +6,8 @@ Check out our docs at https://rancher.com/docs/rancher/v2.x/en/
 
 Browse to https://{{ .Values.hostname }}
 
+If this is the first time you installed Rancher, run this command to get link to setup initial password:
+
+$ echo https://{{ .Values.hostname }}/dashboard/?setup=$(kubectl get secret --namespace {{ .Release.Namespace }} bootstrap-secret -o jsonpath="{.data.bootstrapPassword}" | base64 --decode)
+
 Happy Containering!
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -2,6 +2,7 @@ package management
 
 import (
 	"context"
+	"os"
 	"reflect"
 	"sort"
 	"sync"
@@ -13,6 +14,7 @@ import (
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/rancher/rancher/pkg/wrangler"
+	"github.com/rancher/wrangler/pkg/randomtoken"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/bcrypt"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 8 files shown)


chart/templates/NOTES.txt
```diff
@@ -6,4 +6,9 @@ Check out our docs at https://rancher.com/docs/rancher/v2.x/en/
 
 Browse to https://{{ .Values.hostname }}
 
+If this is your first time login, use boostrap username and password below:
```
superseb, line 9: boostrap typo. Probably better to rephrase to "If this is the first time you installed Rancher, use username and password as shown below:"

pkg/data/management/role_data.go
```diff
@@ -481,6 +490,9 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
 		if err != nil && !apierrors.IsAlreadyExists(err) {
 			return "", errors.Wrap(err, "can not ensure admin user exists")
 		}
+		if err == nil {
+			logrus.Infof("Default username/password created, username: admin, password: %s", bootstrapPassword)
```
superseb, line 494: Probably more safe to write the password to a file in the container? If logs are forwarded they end up somewhere. In general I think we avoid logging sensitive data unless its in TRACE.

vincent99: It's not sensitive any more than knowing the current hardcoded "admin" is, you have to change it on setup.

(ok change password doesn't    _actually_ require it to be different, you can change that if you want)

This should be as easy to use as possible, which means a `https://<server-url or hostname>/?setup=<password>` URL (that terminals will make clickable).

StrongMonkey: Yeah, this is changed to https://<server-url or hostname>/dashboard/?setup=<password>, looks like UI respect this setup query and auto-forward to change password page.

chart/templates/NOTES.txt
```diff
@@ -6,4 +6,8 @@ Check out our docs at https://rancher.com/docs/rancher/v2.x/en/
 
 Browse to https://{{ .Values.hostname }}
 
+If this is the first time you installed Rancher, run this command to get link to setup initial password:
+
+$ echo https://{{ .Values.hostname }}/dashboard/?setup=$(kubectl get secret --namespace {{ .Release.Namespace }} bootstrap-secret -o jsonpath="{.data.bootstrapPassword}" | base64 --decode)
```
vincent99, line 11: Can we print out the actual password from templating so this can actually be a directly clickable link?  This is your first interaction with the product, so the less friction the better.

StrongMonkey: Tried many times... but helm doesn't seem to have a good way to define a random number elsewhere and import it to note.txt, also that will get re-generated everytime everytime chart is installed/upgraded.

superseb (approved):

Jono-SUSE-Rancher:
@paynejacob  - Hey Jacob, can you shepherd this through and just make sure it gets in? Thanks.

Issue #31385, opened 2021-02-19
Custom encryption config is case sensitive in 2.5.x


WalterDurkee:
**What kind of request is this (question/bug/enhancement/feature request):**
Question/bug

**Steps to reproduce (least amount of steps as possible):**
We've upgraded our cluster from 2.3.8 to 2.5.5. local cluster is not encrypted but for our managed cluster we use following config provided via rke templates:
```
rancher_kubernetes_engine_config:
...
  services:
...
      secrets_encryption_config:
        custom_config:
          apiVersion: apiserver.config.k8s.io/v1
          kind: EncryptionConfiguration
          resources:
          - providers:
              - aescbc:
                  keys:
                    - name: key1
                      secret: xxxx
              - identity: {}
            resources:
              - secrets
        enabled: true
```
And this used to work on 2.3.x without any problems. With 2.5.5 `/etc/kubernetes/ssl/encryption.yaml` is zeroed and looks like:
```
apiVersion: apiserver.config.k8s.io/v1
kind: EncryptionConfiguration
resources: []
```
which basically causes kube-apiserver not to start. 

**Result:**
kube-apiserver not starting

**Other details that may be helpful:**
Tried without custom_config like:
```
rancher_kubernetes_engine_config:
...
  services:
...
      secrets_encryption_config:
        enabled: true
```
and the result was the same. 

Tried also spinning literally one node aio cluster with new template and behaviour was the same.

**Environment information**
- Rancher version: 2.5.5
- Installation option: H/A

**??**
Are we doing something that we shouldn't? Did something change? Config matches with the docs.

gz#15889
gz#16142

dramich:
When using a [custom configuration for encrypting data at rest](https://rancher.com/docs/rke/latest/en/config-options/secrets-encryption/), there is a bug relating to the format that exists in Rancher 2.5.0+.

In 2.4.x, when configuring `secrets_encryption_config` Rancher would take any case for the keys and adjust them to be the case it needed, and that cycle repeated. In 2.5 however, the cases have to match on submit. The RKE docs linked above display the configuration keys all starting with lowercase, and with this issue that format will not work.

If you set this configuration in 2.4 and you upgraded to 2.5, everything should continue to work, as the configuration's case will still be correct. But if you edit the configuration and change the case of the keys, the cluster will become unavailable. It appears some workloads may continue to run, but the cluster's API server will be unresponsive.

[...]

PR #32760, opened 2021-05-17
Migrate encryption config on startup

dramich:
Problem:
Rancher 2.5 clients are case sensitive based on the rke config that does
not use a properly tagged struct

Solution:
Update rke to use a json tagged struct for the config and migrate all
clusters to use the proper casing of the configs

https://github.com/rancher/rancher/issues/31385

### DIFF
--- a/pkg/rancher/rancher.go
+++ b/pkg/rancher/rancher.go
@@ -2,14 +2,18 @@ package rancher
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
 
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
 	responsewriter "github.com/rancher/apiserver/pkg/middleware"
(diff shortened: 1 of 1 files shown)


pkg/rancher/rancher.go
```diff
@@ -263,3 +279,62 @@ func localClusterEnabled(opts *Options) bool {
...
+		}
+		cluster.Annotations[encryptionConfigUpdate] = "true"
+
+		u, err := unstructured.ToUnstructured(cluster)
+		if err != nil {
+			allErrors = multierror.Append(err, allErrors)
+			continue
+		}
+
+		_, err = clusterDynamicClient.Update(ctx, u, metav1.UpdateOptions{})
```
luthermonson, line 334: probably need backoff/retry on this.. maybe ignore errors until some reasonable amount of retries and error out only at the end?

dramich: Added a retry loop

PR #18911, opened 2019-03-14
Configurable Helm chart TLS secret name

madddi:
Our TLS secret is used by different tools and it would help us to have the possibility to configure the secret used by the ingress.

PR #32763, opened 2021-05-17
Support for existing ingress secret

digiserg:
This is a small change to avoid using a hard coded name for the Ingress secret and allow using an existing one with the TLS certs.


dramich:
Closing in favor of https://github.com/rancher/rancher/pull/18911

Issue #25273, opened 2020-02-05
Native Private Links AKS Cluster Mgmt


leflambeur:
**What kind of request is this (question/bug/enhancement/feature request):**
Feature Request/Enhancement

**Request**
Rancher Support for Azure Private Links at GA (feature is in Public Preview)

Can Rancher include creating/managing AKS Private Link clusters in a future version when the feature becomes GA in Azure, it is understood it is out of scope until then.

It is highly beneficial and increasingly standardised for secure networking environments that customers be allowed to use Private Links Networks for provisioning and managing their data and infrastructure when using the cloud.

Rancher supporting this natively would enable customers who have these concerns to operate with confidence while expanding the customer base who need this complexity to meet stringent audit requirements.

gz#11713
gz#8164

janeczku:
Azure Private Link for AKS is now GA.
https://azure.microsoft.com/en-us/updates/azure-private-link-for-aks-is-now-generally-available/

prabalsharma:
Tested private cluster on master-7a2fc74, operator rc 11

PR #32771, opened 2021-05-18
Add smarter private-requires-tunnel detection for AKS

thedadams:
The previous implementation would always return `false` for
privateRequiresTunnel for any error that occurred. This may lead to some
false positives for errors that can be recovered from.

Now, privateRequiresTunnel will only be set to `false` when we detect
the proper error.

### DIFF
--- a/pkg/controllers/management/aks/aks_cluster_handler.go
+++ b/pkg/controllers/management/aks/aks_cluster_handler.go
@@ -186,9 +186,9 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			if err != nil {
 				return cluster, err
 			}
-			if mustTunnel {
+			if mustTunnel != nil {
 				cluster = cluster.DeepCopy()
-				cluster.Status.AKSStatus.PrivateRequiresTunnel = &mustTunnel
+				cluster.Status.AKSStatus.PrivateRequiresTunnel = mustTunnel
 				cluster.Status.ServiceAccountToken = serviceToken
 				return e.ClusterClient.Update(cluster)
 			}
@@ -387,9 +387,13 @@ func (e *aksOperatorController) deployAKSOperator() error {
 	return e.DeployOperator(aksOperator, aksOperatorTemplate, aksShortName)
 }
 
-func (e *aksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, bool, error) {
+func (e *aksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, *bool, error) {
 	restConfig, err := e.getKubeConfig(cluster)
+	if err != nil {
+		return "", nil, err
+	}
 
+	requiresTunnel := new(bool)
 	netDialer := net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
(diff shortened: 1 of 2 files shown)


mjura (dismissed):
LGTM

mjura:
There is missing entry in `go.sum`
```
github.com/rancher/aks-operator@v1.0.1-rc4: missing go.sum entry; to add it:
go mod download github.com/rancher/aks-operator
```

cmurphy (commented):
Do we need to do the same for gke? https://github.com/rancher/rancher/blob/55f0bb22ce58f477dab4f81b208300c0e74fc44f/pkg/controllers/management/gke/gke_cluster_handler.go#L420

thedadams:
Related Issue: https://github.com/rancher/rancher/issues/25273

mjura (dismissed):
LGTM

thedadams:
@cmurphy Yes, I also changed the implementation for GKE. These changes will have to be backported to 2.5

cmurphy (approved):

thedadams:
Both reviewer approved, but I had to resolver merge conflicts in `go.mod`.

mjura (approved):
LGTM

cmurphy (approved):

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -387,9 +387,13 @@ func (e *aksOperatorController) deployAKSOperator() error {
 	return e.DeployOperator(aksOperator, aksOperatorTemplate, aksShortName)
 }
 
-func (e *aksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, bool, error) {
+func (e *aksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, *bool, error) {
```
cbron, line 390: So if I understand this method correctly, for `requiresTunnel` we only return true or false if everything works and we are certain of the cluster case. Nil means error.

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -203,9 +203,9 @@ func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			if err != nil {
```
cbron, line 203: > I also changed the implementation for GKE. These changes will have to be backported to 2.5

Can you clarify that part, I see you changed the gke cluster handler but why does it need to be backported, just because its a bug ?

pkg/controllers/management/aks/aks_cluster_handler.go
thedadams: Yes, that's the overall gist.

pkg/controllers/management/gke/gke_cluster_handler.go
thedadams: Yes. We only want to set `PrivateRequiresTunnel` to a non-nil value if we know the answer for sure.

cbron: Ok, so this might be a performance/overhead problem because we are tunneling when we don't need to be. Can you make a backport issue and put into 2.5.10 and link backport PR there.

cbron (approved):

thedadams:
@cbron I believe this does the opposite. If `PrivateRequiresTunnel == nil`, then we don't tunnel. So this will protect against tunneling unless we are able to determine that we must do so.

pkg/controllers/management/gke/gke_cluster_handler.go
thedadams: I believe this does the opposite. If `PrivateRequiresTunnel == nil`, then we don't tunnel. So this will protect against tunneling unless we are able to determine that we must do so.

PR #32454, opened 2021-04-29
Refactor AKS UI proxy calls

mjura:
AKS Kontainer Engine v2 will use cloud credentials. For checking AKS
cluster capabilities cloud credentials should be also used.
AKS Kontainer Engine v1 should works as before.

Dependency https://github.com/rancher/rancher/pull/31333

PR #32805, opened 2021-05-20
Move Tenant ID to Cloud Credentials

mjura:


mjura:
It will be squashed into https://github.com/rancher/rancher/pull/32454

Issue #32808, opened 2021-05-20
Allow user to disable fleet gitops in 2.6


StrongMonkey:
Rancher has feature flag fleet in 2.5, which can be disabled. In 2.6 fleet is required as part of provisioing process and can't be disabled. However the gitops part of fleet can be disabled since it is not used internally by provisioning. We need to add ability to allow user to disable fleet gitops.

SheilaghM:
Daishan was working on this and almost done. Seemed sensible to let him finish.

StrongMonkey:
Rancher 2.6 will be adding a new feature flag called `continous-delivery`. This contains the gitops feature of fleet.
1. For rancher 2.6, `fleet` feature is required to be turned on. `continous-delivery` is turn on by default too, and can be disabled if user doesn't want gitops feature of fleet.

2. For upgrade from rancher 2.5:

a). If user has `fleet` feature turned **on** in 2.5 and upgrade to 2.6, nothing will change. Both `fleet` and `continous-delivery` will be turned on

b) If user has `fleet` feature turned **off** in 2.5 and upgrade to 2.6. `fleet` feature will be forced to turned on since it is required. `continous-delivery` feature will be set to false.

izaac:
Rancher version: `2.6 master-head (06/15/2021)` _2b50005 _
Rancher HA RKE1
Fleet `100.0.0+up0.3.6-rc2`
UI Issue found when the feature is disabled: https://github.com/rancher/dashboard/issues/3246

I was able to deploy with the feature flag enabled and disabled with helm.
When the feature is `disabled` after helm install I am still able to provision `rke2` clusters (DO).

Will continue testing of Upgrade paths.

izaac:
Rancher version: `2.6 master-head (06/15/2021)` _2b50005_

### Upgrades

Upgrade from 2.5.8 to master-head Default values (rke2 provisioning works) - PASS
Upgrade from 2.5.8 to master-head with options set to true (fleet and continuous-delivery) - PASS

### Pending Upgrade scenarios


Upgrade from 2.5.8 (Fleet disabled) to master-head with defaults (expecting `fleet` and `continuous delivery` _enabled_)
Upgrade from 2.5.8 (Fleet disabled) to master-head with `continuous delivery` disabled (expecting `fleet` _enabled_ and `continuous delivery` _disabled_)


I am doing an RKE2 provisioning as sanity check but that's blocked by: https://github.com/rancher/rancher/issues/33207

[...]

Issue #32688, opened 2021-05-11
Fleet feature gets enabled after upgrade to 2.6 from an upgraded 2.5 without fleet enabled but pods get restarted a few times


izaac:
**What kind of request is this:**
bug

**Steps to reproduce (least amount of steps as possible):**

- Install rancher 2.5.8 HA with fleet feature disabled.
  - `--set 'extraEnv[0].name=CATTLE_FEATURES' --set 'extraEnv[0].value=fleet=false'`
- Upgrade the HA to 2.6 without any features variables in the `helm`

**Result:**

Rancher gets upgraded to 2.6
pods get restarted a few times in the process.

**Other details that may be helpful:**

```
Log:
[FATAL] feature flag [fleet] value has changed, rancher must be restarted


NAME                               READY   STATUS      RESTARTS   AGE
rancher-7984cc8b4-ftd97            1/1     Running     3          55m
rancher-7984cc8b4-v2pkm            1/1     Running     0          50m
rancher-7984cc8b4-xf224            1/1     Running     3          55m
```

In this case two of the pods restarted 3 times.

Related issue: https://github.com/rancher/rancher/issues/32680

**Environment information**
- Rancher version: 2.6 _1e1ed67_
- Installation option: HA RKE

Jono-SUSE-Rancher:
This is expected behavior as fleet is required for operation in 2.6. We will make sure to document in release notes.

izaac:
Rancher version `2.6 master-head (05/26/2021)` _6395fbccd_
Upgrade from 2.5.8
RKE1 HA

[before_upgrade_fleet_false.log](https://github.com/rancher/rancher/files/6550340/before_upgrade_fleet_false.log) - 2.5.8
[after_upgrade_fleet_false.log](https://github.com/rancher/rancher/files/6550337/after_upgrade_fleet_false.log) - master-head

Pods restarted, logs with `--previous`
Pod3 wasn't restarted, so it has no previous container logs.
[after_upgrade_pod2.log](https://github.com/rancher/rancher/files/6550338/after_upgrade_pod2.log)
[after_upgrade_pod1.log](https://github.com/rancher/rancher/files/6550339/after_upgrade_pod1.log)

izaac:
@StrongMonkey I've attached the logs in my previous comment. Of an upgrade from 2.5.8 with fleet disabled to master-head

What's the reason only two of the pods were restarted. And why the third pod wasn't restarted.

izaac:
[...]

PR #32814, opened 2021-05-20
Add gitops feature flag, force fleet to be turned on

StrongMonkey:
https://github.com/rancher/rancher/issues/32808
https://github.com/rancher/rancher/issues/32688

### DIFF
--- a/pkg/controllers/dashboard/fleetcharts/controller.go
+++ b/pkg/controllers/dashboard/fleetcharts/controller.go
@@ -7,6 +7,7 @@ import (
 
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/catalogv2/system"
+	"github.com/rancher/rancher/pkg/features"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/wrangler"
 )
@@ -67,6 +68,10 @@ func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error
 		"apiServerURL": settings.ServerURL.Get(),
 		"apiServerCA":  settings.CACerts.Get(),
 		"global":       systemGlobalRegistry,
+	}
+
+	fleetChartValues["gitops"] = map[string]interface{}{
+		"enabled": features.Gitops.Enabled(),
 	}
 
 	gitjobChartValues := make(map[string]interface{})
--- a/pkg/features/feature.go
+++ b/pkg/features/feature.go
@@ -47,6 +47,12 @@ var (
 		true,
 		false,
 		true)
+	Gitops = newFeature(
+		"continuous-delivery",
+		"Gitops components in fleet",
+		true,
+		false,
+		true)
 	Auth = newFeature(
 		"auth",
 		"Enable authentication",
@@ -170,6 +176,24 @@ func InitializeFeatures(featuresClient managementv3.FeatureClient, featureArgs s
 			f.Set(*featureState.Spec.Value)
 		}
 	}
+}
+
+func SetFeature(featuresClient managementv3.FeatureClient, featureName string, value bool) error {
+	if featuresClient == nil {
+		return nil
+	}
+
+	featureState, err := featuresClient.Get(featureName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	featureState.Spec.Value = &[]bool{value}[0]
+	if _, err = featuresClient.Update(featureState); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // applyArgumentDefaults reads the features arguments and uses their values to overwrite
--- a/pkg/rancher/rancher.go
+++ b/pkg/rancher/rancher.go
@@ -2,7 +2,6 @@ package rancher
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 
 	responsewriter "github.com/rancher/apiserver/pkg/middleware"
@@ -29,6 +28,7 @@ import (
 	steveauth "github.com/rancher/steve/pkg/auth"
 	steveserver "github.com/rancher/steve/pkg/server"
 	"github.com/rancher/wrangler/pkg/k8scheck"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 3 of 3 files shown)


pkg/controllers/dashboard/fleetcharts/controller.go
```diff
@@ -69,6 +70,12 @@ func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error
 		"global":       systemGlobalRegistry,
 	}
 
+	if !features.Gitops.Enabled() {
+		fleetChartValues["gitops"] = map[string]interface{}{
+			"enabled": false,
+		}
```
nickgerace, line 76: Are we using a map in anticipation of future values being added?

pkg/rancher/rancher.go
```diff
@@ -104,7 +104,14 @@ func New(ctx context.Context, clientConfg clientcmd.ClientConfig, opts *Options)
 	}
 
 	if features.MCM.Enabled() && !features.Fleet.Enabled() {
-		return nil, fmt.Errorf("multi-cluster-management features requires feature fleet=true")
+		logrus.Info("fleet can't be turned off when MCM is enabled. Turning on fleet and turning off continuous-delivery")
+		// fleet needs to be turned on. If fleet is not and MCM is on, turn on fleet and turn off continuous-delivery
```
nickgerace, line 108: nit: I think the info statement makes the comment not necessary

nickgerace (dismissed):
LGTM barring nits

rmweir (changes requested):

pkg/controllers/dashboard/fleetcharts/controller.go
StrongMonkey: actually I can change it to 
```
fleetChartValues["gitops"] = map[string]interface{}{			
  "enabled": features.Gitops.Enabled(),		
}
```

pkg/rancher/rancher.go
StrongMonkey: @rmweir We need to be able to handle the upgrade case where user has turned off fleet in 2.5 and upgrade to 2.6. If we are still returning the error the upgrade will be broken for these users and it will require extra steps to recover the setup.

StrongMonkey: See this comment https://github.com/rancher/rancher/issues/32808#issuecomment-845480007

pkg/features/feature.go
```diff
@@ -160,6 +166,25 @@ func InitializeFeatures(featuresClient managementv3.FeatureClient, featureArgs s
 	}
 }
 
+func SetFeature(featuresClient managementv3.FeatureClient, feature *Feature, value bool) error {
```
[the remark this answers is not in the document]
nickgerace: Agreed here ^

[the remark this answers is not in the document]
rmweir: No response to this so far.

pkg/rancher/rancher.go
```diff
@@ -104,7 +104,13 @@ func New(ctx context.Context, clientConfg clientcmd.ClientConfig, opts *Options)
 	}
 
 	if features.MCM.Enabled() && !features.Fleet.Enabled() {
-		return nil, fmt.Errorf("multi-cluster-management features requires feature fleet=true")
+		logrus.Info("fleet can't be turned off when MCM is enabled. Turning on fleet and turning off continuous-delivery")
+		if err := features.SetFeature(wranglerContext.Mgmt.Feature(), features.Fleet, true); err != nil {
```
rmweir, line 108: Can we just remove the fleet feature flag if this is the case? This should solve https://github.com/rancher/rancher/issues/32688.

pkg/rancher/rancher.go
```diff
@@ -104,7 +104,13 @@ func New(ctx context.Context, clientConfg clientcmd.ClientConfig, opts *Options)
...
+			return nil, err
+		}
+		if err := features.SetFeature(wranglerContext.Mgmt.Feature(), features.Gitops, false); err != nil {
```
rmweir, line 111: Is it necessary that we assume the desired value and set this on their behalf? If we remove the fleet feature flag we can just say that it has been removed and a new feature flag for gitops has been added which the user can disable if they do not desire it.

rmweir (changes requested):

pkg/rancher/rancher.go
StrongMonkey: We can't remove this flag, since downstream cluster stilll needs the ability to disable fleet.

pkg/features/feature.go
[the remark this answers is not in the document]
StrongMonkey: https://github.com/rancher/rancher/pull/32814/files#diff-2f6213e2bd15bb256be4ed430e664c08612e86384445be797a20d5f56d48d101R184 It stills needs the ability to set in-memory map since it can be accessed in-memory. I just followed the code because it updates in both crd and in-memory struct.

pkg/rancher/rancher.go
StrongMonkey: The downstream cluster runs mini rancher so it needs the ability to disable fleet

pkg/features/feature.go
```diff
@@ -166,6 +172,25 @@ func InitializeFeatures(featuresClient managementv3.FeatureClient, featureArgs s
...
+	if err != nil {
+		return err
+	}
+
+	featureState.Spec.Value = &[]bool{value}[0]
+	if _, err = featuresClient.Update(featureState); err != nil {
+		return err
+	}
+
+	feature.Set(value)
```
rmweir, line 190: There is a handler that will set this automatically. It's important we let the handler do it because it is on the ScaledContext and handles restarting each Rancher node when necessary. Doing this will bypass that restart for whichever node it is used on. Set is something that must occur on every rancher node while feature update only needs to happen once.

rmweir (changes requested):

pkg/features/feature.go
StrongMonkey: I see, will change this.

pkg/rancher/rancher.go
StrongMonkey: Same, can't remove fleet feature since downstream cluster will need to set it to false.

rmweir (changes requested):
I think feature flag not really being a toggleable feature anymore for main rancher install and us auto setting gitops feature flag under certain conditions to override default needs to be documented. Can you make a docs issue and link it? Are we sure gitops feature flag default needs to be overwritten? May be easier to explain that it's a new feature flag that is not the same as fleet and can be disabled by them if they wish.

StrongMonkey:
@rmweir Yes documentation will be added. https://github.com/rancher/docs/issues/3313

> Are we sure gitops feature flag default needs to be overwritten?

Yes, due to the fact that fleet is required in 2.6, any previous install that disabled it will need to be overriden, otherwise the install will be broken. This should be a hard requirement and user will not be allowed to override. That's why there are issues that hide the flag in UI/backend. Instead like you said we should document that a new flag is introduced to only disable the part and old fleet feature is no longer available. This is to solve the upgrade bug as there is no other way around. There are already discussion around that, I can get you in the thread.

rmweir (approved):
LGTM

nickgerace (approved):

dramich (approved):

Issue #32689, opened 2021-05-11
Cannot delete clusters  created in new cluster explorer from Ember UI 


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create clusters (had about 4-5 clusters) from Dashboard or the new UI - RKE clusters.
- When the clusters are up and running, delete the clusters by selecting them all from Ember UI.
- The clusters are seeing stuck in pending state.
- Errors in Rancher logs:

[clusterDeleteLogs.txt](https://github.com/rancher/rancher/files/6462207/clusterDeleteLogs.txt)

**Note:**
Select all and retry deleting them, they get removed, but are seen on the UI again.
Continuous errors in rancher logs:
```
E0511 22:04:57.403112       7 reflector.go:139] pkg/mod/github.com/rancher/client-go@v0.21.0-rancher.1/tools/cache/reflector.go:168: Failed to watch *v1.Node: failed to list *v1.Node: {"Code":{"Code":"Forbidden","Status":403},"Message":"clusters.management.cattle.io \"c-m-ed6faf9c\" is forbidden: User \"u-2ydmfgoyz5\" cannot get resource \"clusters\" in API group \"management.cattle.io\" at the cluster scope","Cause":null,"FieldName":""} (get nodes)
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `1e1ed673`
- Installation option (single install/HA): Single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): RKE2 clusters - DO and import RKE from new cluster explorer UI
- Kubernetes version (use `kubectl version`):

```
Any
```

sowmyav27:
On master-head commit id: `f90b36e`

Create clusters  2 RKE2 DO clusters created from cluster explorer and 1 RKE1 cluster created from Ember UI
When the clusters are up and running, delete the clusters by selecting them all from Ember UI.
The clusters get deleted
The nodes of RKE2 clusters are _not_ deleted from backend. Logged this issue to track it https://github.com/rancher/rancher/issues/32868

Issue #32663, opened 2021-05-10
Deleting an RKE cluster from Dashboard does not delete the cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Create a cluster using new Dashboard
- Create an RKE cluster (toggle and choose RKE from RKE vs RKE2)
- Wait for the cluster to come up successfully.
- It will be available in the Cluster Management page (From left Panel --> Cluster Management)
- API Request sent: Method: `DELETE ` and request URL: `https://<server>/v1/provisioning.cattle.io.clusters/fleet-default/c-cjc9c`
- However, the cluster is seen on the `Left Pane` under Cluster Explorer but is disabled _or_ Greyed out.

**Expected Result:**
The cluster should be deleted in Rancher.

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `8c014ef`
- Installation option (single install/HA): Single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): RKE EC2/any
- Kubernetes version (use `kubectl version`):

```
Any
```

izaac:
Rancher version `v2.6 master-head (05/18/2021) _55f0bb2_

This also happens with DO clusters.
https://github.com/rancher/rancher/issues/32663

sowmyav27:
**On master-head commit id: `f90b36e`**

- Create a cluster using new Dashboard
- Create an RKE cluster (toggle and choose RKE from RKE vs RKE2)
- Wait for the cluster to come up successfully.
- It will be available in the Cluster Management page (From left Panel --> Cluster Management)
- Delete cluster, cluster is deleted successfully from Cluster management page and the nodes are deleted from the backend.

PR #32835, opened 2021-05-21
Mutually assured destruction of v1 and v3.Cluster

ibuildthecloud:


kinarashah:
https://github.com/rancher/rancher/issues/32663 
https://github.com/rancher/rancher/issues/32689

PR #32840, opened 2021-05-22
Audit Log at Login - Check the content-type header having charset as well which is set by Dashboard UI

prachidamle:
https://github.com/rancher/rancher/issues/32588

### DIFF
--- a/pkg/auth/audit/audit.go
+++ b/pkg/auth/audit/audit.go
@@ -110,7 +110,7 @@ func newAuditLog(writer *LogWriter, req *http.Request) (*auditLog, error) {
 	contentType := req.Header.Get("Content-Type")
 	loginReq := isLoginRequest(req.RequestURI)
 	if writer.Level >= levelRequest || loginReq {
-		if bodyMethods[req.Method] && contentType == contentTypeJSON {
+		if bodyMethods[req.Method] && strings.HasPrefix(contentType, contentTypeJSON) {
 			reqBody, err := readBodyWithoutLosingContent(req)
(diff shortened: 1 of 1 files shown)


pkg/auth/audit/audit.go
```diff
@@ -110,7 +111,7 @@ func newAuditLog(writer *LogWriter, req *http.Request) (*auditLog, error) {
 	contentType := req.Header.Get("Content-Type")
 	loginReq := isLoginRequest(req.RequestURI)
 	if writer.Level >= levelRequest || loginReq {
-		if bodyMethods[req.Method] && contentType == contentTypeJSON {
+		if bodyMethods[req.Method] && (contentType == contentTypeJSON || contentType == contentTypeJSONCharset) {
```
aiyengar2, line 114: Any reason why we can't just check the prefix?
```suggestion
		if bodyMethods[req.Method] && strings.HasPrefix(contentType, contentTypeJSON) {
```

aiyengar2 (changes requested):

pkg/auth/audit/audit.go
prachidamle: Can do that, dint want to change it since this code seems to be existing for quite a while. But makes sense to check prefix.

dramich: I agree with @aiyengar2's suggestion.

dramich (dismissed):

dramich (approved):

aiyengar2 (approved):
LGTM

PR #32851, opened 2021-05-24
change loglevel to debug

leodotcloud:


ibuildthecloud:
Is the intention to not print the token?  Can we just drop the token, or make printing the token only at debug level.  This message is useful to know when things aren't connecting as it repeats.  The logging is so light in the agent I'd hate to have even less information.

leodotcloud:
>Is the intention to not print the token?

yes

@ibuildthecloud updated the PR with requested changes.

cmd/agent/main.go
```diff
@@ -331,7 +331,8 @@ func run() error {
 		if !isConnect() {
 			wsURL += "/register"
 		}
-		logrus.Infof("Connecting to %s with token %s", wsURL, token)
+		logrus.Infof("Connecting to %s", wsURL)
```
deniseschannon, line 334: Could we do a substring of the actual token as part of the info statement?

snasovich:
Replaced by https://github.com/rancher/rancher/pull/33042

Issue #32586, opened 2021-05-06
GKE/AKS Windows compatability


cbron:
See if: 
a) we can easily create clusters with Windows nodes
b) if your features deploy correctly on them. I am assuming there will be problems with taints and tolerations.

cmurphy:
Additional fields were needed in the spec in order to create Windows nodes https://github.com/rancher/aks-operator/pull/5

Besides that, AKS requires at least one node pool to be a "System" type and "System" node pools are Linux-only, similar to the way GKE requires at least one non-windows node pool. So there's not an issue with taints, and actually none of the nodes in any node pools seem to have any taints. The cluster agent gets deployed successfully and the cluster explorer comes up fine.

prabalsharma:
tested on aks operator rc4
> See if:
> a) we can easily create clusters with Windows nodes
Works
> b) if your features deploy correctly on them. I am assuming there will be problems with taints and tolerations.

monitoring v2 installs fine. However, when I go to Grafana, **the monitoring data is populated only for linux nodes** (not windows nodes)

cmurphy:
Update: I haven't entirely gotten it to work yet, but I think it /should/ work without needing any changes in aks-operator.

What I've done so far:
1. followed [this guide](https://docs.microsoft.com/en-us/azure/aks/rdp) to set up an RDP connection to the windows nodes in the cluster (the `az` commands did not work, but you can reproduce most of it via the web UI).
2. installed [wins](https://github.com/rancher/wins) on the nodes. 
3. installed the monitoring chart through Rancher, setting `global.windows.enabled=true` in `values.yaml`. For some reason this never seems to take on the first install, I have to go upgrade it and change it again.
4. Eventually you should see a healthy `rancher-monitoring-windows-exporter` daemonset running on the nodes using the `kubernetes.io/os=windows` selector.

At that point things seem to look fine but there are still no metrics in grafana for those nodes. In the Prometheus "Targets" UI I see the error "Get "http://10.42.0.112:9796/metrics": context deadline exceeded" for the `cattle-monitoring-system/rancher-monitoring-windows-exporter/0 (0/2 up)` target, so I think it's something wrong with the VNet or something wrong with how I'm running wins.exe.

cbron:
Hmm. We should talk about this on Monday. Don't think you need to put any more work on it for now. It's mostly a question of what Rancher typically does for windows nodes, and how to align expectations with that. But that last part about VNet we may need to dive into at some point.

[...]

PR #32869, opened 2021-05-25
Validate against using windows node pools for AKS

cmurphy:
Windows node pools do not currently work with Rancher charts, so add a
validator to ensure "Windows" is not set for the node pools OS Type, and
update the schema to remove windows-related fields.

Depends on https://github.com/rancher/aks-operator/pull/17

https://github.com/rancher/rancher/issues/32586

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -449,13 +449,16 @@ func validateAKSNodePools(spec *v32.ClusterSpec) error {
 		return nil
 	}
 	if len(nodePools) == 0 {
-		return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("must have at least one nodepool"))
+		return httperror.NewAPIError(httperror.InvalidBodyContent, "must have at least one nodepool")
 	}
 
 	for _, np := range nodePools {
 		name := np.Name
 		if to.String(name) == "" {
-			return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("nodePool Name cannot be an empty string"))
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "nodePool Name cannot be an empty string")
+		}
+		if np.OsType == "Windows" {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "windows node pools are not supported")
 		}
 	}
 
--- a/pkg/apis/go.mod
+++ b/pkg/apis/go.mod
@@ -2,7 +2,9 @@ module github.com/rancher/rancher/pkg/apis
 
 go 1.14
 
-replace k8s.io/client-go => k8s.io/client-go v0.21.0
+replace (
+	k8s.io/client-go => k8s.io/client-go v0.21.0
+)
 
 require (
 	github.com/pkg/errors v0.9.1


pkg/api/norman/customization/cluster/validator.go
```diff
@@ -432,13 +432,16 @@ func validateAKSNodePools(spec *v32.ClusterSpec) error {
...
 	}
 
 	for _, np := range nodePools {
 		name := np.Name
 		if to.String(name) == "" {
-			return httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("nodePool Name cannot be an empty string"))
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "nodePool Name cannot be an empty string")
+		}
+		if np.OsType == "Windows" {
+			return httperror.NewAPIError(httperror.InvalidBodyContent, "windows node pools are not supported")
```
cbron, line 444: This means it will take effect on the API layer, but you could technically make one with a plain cluster CR and you could also register one. Registration case I hadn't considered yet. We could either error or send warning.

cmurphy: If you made a plain cluster CR you would run into this https://github.com/rancher/aks-operator/pull/17/files#diff-1aa54d4941ca80708078867e1b130aff086719689479dea38683fded2f3c7852R391-R392

I'm not sure what to do about registration.

mjura (dismissed):
LGTM

pkg/api/norman/customization/cluster/validator.go
thedadams: I see two potential options:
1. Release note it saying it is possible to import a cluster with Windows nodes, but it is not supported.
2. We could do some import validation of the AKS cluster in `aks-operator`. We don't do this for imported cluster in any of the the three operators, but we could do it in this case if it is deemed important enough.

cbron: > Release note it saying it is possible to import a cluster with Windows nodes, but it is not supported.

Would that work though if all these fields are deleted ? The operator/rancher code would treat it like Linux nodes right ? And then who knows what happens. I think we should reject it.

cmurphy: The windowsProfile fields would be ignored when the cluster is imported, and they wouldn't be manageable from rancher. But there's nothing in the operator code that would treat the node pools differently. It's just that the monitoring chart still won't work on them.

cbron: Interesting. So we go that route and then QA it to see if we can break it ? Do linux nodes in aks get the default linux node label by default ? https://github.com/rancher/charts/blob/dev-v2.6/charts/rancher-logging/rancher-logging/3.9.401%2Bup3.9.4/values.yaml#L40-L41

cmurphy: Yes, linux nodes get the `kubernetes.io/os=linux` label and windows nodes get `kubernetes.io/os=windows` in AKS. AKS does not allow creating a cluster with windows nodes without also having linux nodes, so the logging chart will always have some linux nodes available where that selector will work.

cbron: K, lets give that a shot and then release note it.

thedadams (dismissed):

mjura:
@cmurphy it seems that there are some issues with go.mod
```
Diff found between ./go.mod and ./pkg/apis/go.mod
```

cmurphy:
We need a new aks-operator tag in order to fix go.mod

go.mod
```diff
@@ -11,6 +11,7 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20200712062324-13d1f37d2d77
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc10
+	github.com/rancher/aks-operator => github.com/cmurphy/aks-operator v0.0.0-20210525201153-687a9fd5a1fb // FIXME
```
cbron, line 14: This can be tagged now right ?

cbron (commented):
lgtm, just need to tag

go.mod
mjura: It was pushed `v1.0.1-rc7`

mjura: I pushed also change to system-charts https://github.com/rancher/system-charts/pull/479

mjura (dismissed):
LGTM

thedadams (approved):

pkg/api/norman/customization/cluster/validator.go
rmweir: I would imagine we could check the upstreamSpec on status and add validation to a controller right?

rmweir (approved):
LGTM. Including the name of the nodepool in the error could be useful but if there were multiple windows nodepools it would only show a single name, which could be confusing, unless this was changed more.

pkg/api/norman/customization/cluster/validator.go
cmurphy: upstreamSpec status doesn't get set until after the cluster object is created so I don't think that validation would be doable here in the norman pre-create validator, we'd have to put it somewhere else where it wouldn't be useful to the UI.

mjura (approved):
LGTM

PR #32885, opened 2021-05-26
Fix check for AKS cluster version in norman

mjura:
AKS cluster version must be validated only,
when it is provided in config.

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -381,6 +381,20 @@ func (v *Validator) validateAKSConfig(request *types.APIContext, cluster map[str
 		}
 	}
 
+	if request.Method != http.MethodPost {
+		return nil
+	}
+
+	// validation for creates only
+	if err := validateAKSClusterName(v.ClusterClient, clusterSpec); err != nil {
+		return err
+	}
+
+	region, regionOk := aksConfig["resourceLocation"]
+	if !regionOk || region == "" {
+		return httperror.NewAPIError(httperror.InvalidBodyContent, "must provide region")
+	}
+
 	return nil
 }
 
@@ -416,6 +430,9 @@ func validateAKSCredentialAuth(request *types.APIContext, credential string, pre
 // validateAKSKubernetesVersion checks whether a kubernetes version is provided
 func validateAKSKubernetesVersion(spec *v32.ClusterSpec, prevCluster *v3.Cluster) error {
 	clusterVersion := spec.AKSConfig.KubernetesVersion
+	if clusterVersion == nil {
+		return nil
+	}
 
 	if to.String(clusterVersion) == "" {
 		return httperror.NewAPIError(httperror.InvalidBodyContent, "cluster kubernetes version cannot be empty string")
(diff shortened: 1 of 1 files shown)


cmurphy (changes requested):
I think this should follow the model from EKS and GKE where the main validation method exits early before validating create-only fields: https://github.com/rancher/rancher/blob/629146a3fbbc8d5677aecece78e627580dd7c95d/pkg/api/norman/customization/cluster/validator.go#L485-L489

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -416,6 +430,9 @@ func validateAKSCredentialAuth(request *types.APIContext, credential string, pre
 // validateAKSKubernetesVersion checks whether a kubernetes version is provided
 func validateAKSKubernetesVersion(spec *v32.ClusterSpec, prevCluster *v3.Cluster) error {
 	clusterVersion := spec.AKSConfig.KubernetesVersion
+	if clusterVersion == nil {
+		return nil
+	}
```
cmurphy, line 435: I'm sorry, I did not mean that you should add a whole new method to validate the cluster name. What I meant was that this section here should not be here. If cluster version is nil during a create request it is invalid, it should error here. Instead of returning nil on a case-by-case basis, the call to `validateAKSKubernetesVersion` should come after a check for `if request.Method != http.MethodPost` in the main `validateAKSConfig`.

cmurphy (changes requested):

pkg/api/norman/customization/cluster/validator.go
mjura: ok, I moved it after `if request.Method != http.MethodPost` , but I want to point that for GKE and for EKS this validation check is before, should we fix also there ?

cmurphy:
I was mistaken on this, on closer look the original implementation was consistent with EKS and GKE so we should revert to that. I'm sorry for the misdirection.

mjura:
@cmurphy ok, I have updated it

thedadams (approved):

cmurphy (approved):

pkg/api/norman/customization/cluster/validator.go
```diff
@@ -445,6 +462,34 @@ func validateAKSNodePools(spec *v32.ClusterSpec) error {
...
+
+	for _, cluster := range clusters.Items {
+		if cluster.Spec.AKSConfig == nil {
+			continue
+		}
+		if name != cluster.Spec.AKSConfig.ClusterName {
+			continue
+		}
+		if region != "" && region != cluster.Spec.AKSConfig.ResourceLocation {
```
rmweir, line 484: What if the regions are both empty?

mjura: Cluster with empty `region` can't be created, since it is hard requirement and validated.

rmweir: Is the `region != ""` not needed then? And if it were wouldn't the error that is returned be misleading?

rmweir (approved):
One question that is more of a nit, otherwise LGTM.

PR #32899, opened 2021-05-26
use docker client with api negotiation

kinarashah:
https://github.com/rancher/rancher/issues/32726


ryansann (approved):
based on comments in the issue, fix lgtm

Issue #31846, opened 2021-03-30
KEv2 Operators (EKS/AKS/GKE) should get the same ca-additional as the Rancher deployment


bentastic27:
There are some users that put Rancher behind a proxy that get SSL errors when creating EKS clusters in the eks-operator pod:

```
time="2021-03-10T14:40:21Z" level=error msg="error syncing 'cattle-global-data/c-clq89': handler eks-controller: error getting role: RequestError: send request failed\ncaused by: Post https://iam.amazonaws.com/: x509: certificate signed by unknown authority, requeuing"
time="2021-03-10T14:49:09Z" level=warning msg="cluster [c-clq89] never advanced to creating status, will not delete EKS cluster"
```

A fix is to mount the ca-addtional.pem from the [tls-ca-addtional](https://rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-k8s/chart-options/#additional-trusted-cas) secret inside /etc/ssl/certs on the pod.

gz#15594
gz#15618

bentastic27:
I made a PR for the EKS one, but perhaps the same changes need to apply in all of the provider kubernetes operators (AKS/GKS) as well.

ranchersvet:
Even after the workaround is applied, the Rancher pods log this error:

2021/03/31 13:07:31 [ERROR] error syncing 'c-sh8jx/m-h6hg4': handler node-controller: unable to refresh config for node m-h6hg4: nodetemplates.management.cattle.io "nt-pg6jb" not found, requeuing
2021/03/31 13:10:00 [INFO] checking cluster [c-rz2lr] upstream state for changes
2021/03/31 13:10:00 [INFO] cluster [c-rz2lr] currently updating, skipping spec sync
2021/03/31 13:15:00 [INFO] checking cluster [c-rz2lr] upstream state for changes
2021/03/31 13:15:00 [INFO] cluster [c-rz2lr] currently updating, skipping spec sync
2021/03/31 13:19:13 [ERROR] error syncing 'c-rz2lr': handler eks-operator-controller: Post "https://XXXXXXXXXXXXXXXXXXXXXX.gr7.us-east-2.eks.amazonaws.com/api/v1/namespaces": x509: certificate signed by unknown authority, requeuing
2021/03/31 13:20:00 [INFO] checking cluster [c-rz2lr] upstream state for changes
2021/03/31 13:20:00 [INFO] cluster [c-rz2lr] currently updating, skipping spec sync
[...]

PR #32900, opened 2021-05-26
Use additional CA for hosted providers

thedadams:
When the call is made to get a service account for the hosted providers,
the CA for the k8s API is used. If Rancher is running behind a proxy or
the like that uses its own CA, then this call fails.

This change will add the additional CAs to the rest config so the CAs
are used in the call to get the service account token.

Issue:
https://github.com/rancher/rancher/issues/31846

### DIFF
--- a/pkg/controllers/management/aks/aks_cluster_handler.go
+++ b/pkg/controllers/management/aks/aks_cluster_handler.go
@@ -5,6 +5,7 @@ import (
 	stderrors "errors"
 	"fmt"
 	"net"
+	"net/url"
 	"reflect"
 	"strings"
 	"time"
@@ -18,11 +19,9 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/management/rbac"
 	"github.com/rancher/rancher/pkg/dialer"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
-	"github.com/rancher/rancher/pkg/kontainer-engine/drivers/util"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/systemaccount"
 	"github.com/rancher/rancher/pkg/types/config"
-	typesDialer "github.com/rancher/rancher/pkg/types/config/dialer"
 	"github.com/rancher/rancher/pkg/wrangler"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -31,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
@@ -180,6 +178,7 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 
 		if cluster.Status.AKSStatus.PrivateRequiresTunnel == nil &&
 			to.Bool(cluster.Status.AKSStatus.UpstreamSpec.PrivateCluster) {
+			// In this case, the API endpoint is private and it has not been determined if Rancher must tunnel to communicate with it.
 			// Check to see if we can still use the control plane endpoint even though
 			// the cluster has private-only access
 			serviceToken, mustTunnel, err := e.generateSATokenWithPublicAPI(cluster)
@@ -317,7 +316,7 @@ func (e *aksOperatorController) updateAKSClusterConfig(cluster *mgmtv3.Cluster, 
 
 // generateAndSetServiceAccount uses the API endpoint and CA cert to generate a service account token. The token is then copied to the cluster status.
 func (e *aksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
-	restConfig, err := e.getKubeConfig(cluster)
+	restConfig, err := e.getRestConfig(cluster)
 	if err != nil {
 		return cluster, fmt.Errorf("error getting service account token: %v", err)
 	}
@@ -327,7 +326,8 @@ func (e *aksOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Clu
 		return cluster, err
 	}
 
-	saToken, err := generateSAToken(restConfig, clusterDialer)
+	restConfig.Dial = clusterDialer
+	saToken, err := clusteroperator.GenerateSAToken(restConfig)
 	if err != nil {
 		return cluster, fmt.Errorf("error getting service account token: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -390,30 +389,25 @@ func (e *aksOperatorController) deployAKSOperator() error {
...
-	clientSet, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		return "", fmt.Errorf("error creating clientset: %v", err)
 	}
 
-	return util.GenerateServiceAccountToken(clientSet)
+	return serviceToken, false, err
 }
 
 func (e *aksOperatorController) getKubeConfig(cluster *mgmtv3.Cluster) (*rest.Config, error) {
```
cmurphy, line 413: Not mission critical but maybe this could be renamed to `getRestConfig` for consistency with the other handlers

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -347,17 +347,23 @@ func (e *gkeOperatorController) updateGKEClusterConfig(cluster *mgmtv3.Cluster,
 
 // generateAndSetServiceAccount uses the API endpoint and CA cert to generate a service account token. The token is then copied to the cluster status.
 func (e *gkeOperatorController) generateAndSetServiceAccount(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
-
 	ctx := context.Background()
 	ts, err := controller.GetTokenSource(ctx, e.SecretsCache, cluster.Spec.GKEConfig)
```
cmurphy, line 351: This could potentially move into `getRestConfig` for consistency with the equivalent method in the EKS handler

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -390,30 +389,25 @@ func (e *aksOperatorController) deployAKSOperator() error {
 func (e *aksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, bool, error) {
 	restConfig, err := e.getKubeConfig(cluster)
 
-	netDialer := net.Dialer{
-		Timeout:   30 * time.Second,
-		KeepAlive: 30 * time.Second,
-	}
-	serviceToken, err := generateSAToken(restConfig, netDialer.DialContext)
+	// TODO: there will be a merge conflict here with another PR
```
cmurphy, line 392: Which one? Is this still a concern?

cmurphy (commented):
LGTM aside from the TODOs. I like the new clusteroperator GenerateSAToken. Left some minor thoughts but feel free to ignore.

pkg/controllers/management/aks/aks_cluster_handler.go
thedadams: Yes. I just merged it, so I will fix them and update.

pkg/controllers/management/gke/gke_cluster_handler.go
thedadams: Done.

pkg/controllers/management/aks/aks_cluster_handler.go
thedadams: Done.

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -489,52 +489,33 @@ func (e *eksOperatorController) deployEKSOperator() error {
 
 func (e *eksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, *bool, error) {
 	var publicAccess *bool
-	accessToken, err := e.getAccessToken(cluster)
+
+	// TODO: this will have merge conflicts with another PR.
```
cmurphy, line 493: ☝️

thedadams: Removed. Thanks!

cmurphy (dismissed):

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -388,41 +387,35 @@ func (e *aksOperatorController) deployAKSOperator() error {
 }
 
 func (e *aksOperatorController) generateSATokenWithPublicAPI(cluster *mgmtv3.Cluster) (string, *bool, error) {
```
rmweir, line 389: Could all KEv2 clusters use the same `generateSATokenWIthPublicAPI` if the restConfig were passed as a parameter?

pkg/controllers/management/gke/gke_cluster_handler.go
```diff
@@ -418,37 +418,45 @@ func (e *gkeOperatorController) deployGKEOperator() error {
...
-
-	netDialer := net.Dialer{
-		Timeout:   30 * time.Second,
-		KeepAlive: 30 * time.Second,
-	}
 	publicAccess := new(bool)
-	serviceToken, err := generateSAToken(cluster.Status.APIEndpoint, cluster.Status.CACert, ts, netDialer.DialContext)
+	serviceToken, err := clusteroperator.GenerateSAToken(restConfig)
 	if err != nil {
+		*publicAccess = true
```
rmweir, line 428: Doesn't this actually mean publicAccess should be false? Should this variable be renamed to `mustTunnel`?

rmweir (changes requested):
A couple comments, overall I think the logic is probably fine but I'm worried that serviceaccount logic in general is becoming more complicated to read, not particularly bc of this PR. Do you agree or have thoughts on how we can tackle that? Part of the complexity for me is understanding the difference between GenerateSAToken, generateSATokenWithPublicAPI, and generateAndSetServiceAccount, along with why each one is used in certain areas. Another area is this and how the 2 code blocks relate to one another, and when one is expected to be entered. It's possible more comments could clarify this: https://github.com/rancher/rancher/blob/master/pkg/controllers/management/gke/gke_cluster_handler.go#L197-L233. I'm not sure it's clear

pkg/controllers/management/clusteroperator/utils.go
```diff
@@ -51,9 +55,10 @@ type OperatorController struct {
 }
 
 type operatorValues struct {
-	HTTPProxy  string `json:"httpProxy,omitempty"`
-	HTTPSProxy string `json:"httpsProxy,omitempty"`
-	NoProxy    string `json:"noProxy,omitempty"`
+	HTTPProxy    string `json:"httpProxy,omitempty"`
+	HTTPSProxy   string `json:"httpsProxy,omitempty"`
+	NoProxy      string `json:"noProxy,omitempty"`
+	AdditionalCA bool   `json:"additionalTrustedCAs"`
```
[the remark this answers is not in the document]
thedadams: This is to make it consistent with the Rancher chart: https://github.com/rancher/rancher/blob/master/chart/templates/deployment.yaml#L136-L145

pkg/controllers/management/gke/gke_cluster_handler.go
thedadams: Yes, it should be `mustTunnel`. Will fix.

rmweir (approved):
LGTM. Hope we switch to the bastion method to simplify some of the surrounding logic.

cmurphy (dismissed):

pkg/controllers/management/clusteroperator/utils.go
```diff
@@ -251,3 +264,39 @@ func (e *OperatorController) RecordCAAndAPIEndpoint(cluster *mgmtv3.Cluster) (*m
...
+	if err != nil {
+		return "", fmt.Errorf("error creating clientset: %v", err)
+	}
+
+	return util.GenerateServiceAccountToken(clientSet)
+}
+
+func addAdditionalCA(secretsCache wranglerv1.SecretCache, caCert string) (string, error) {
+	additionalCA, err := getAdditionalCA(secretsCache)
+	if err != nil || additionalCA == nil {
```
cbron, line 279: yet another very minor nit: this would be more readable split up. I know its more lines of code and another operation, but this looks like a plain error return, when really its error return _or_ a no-op.

pkg/controllers/management/clusteroperator/utils.go
```diff
@@ -251,3 +264,39 @@ func (e *OperatorController) RecordCAAndAPIEndpoint(cluster *mgmtv3.Cluster) (*m
...
+	secret, err := secretsCache.Get(namespace.System, "tls-ca-additional")
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if secret == nil {
+		return nil, nil
+	}
+
+	return secret.Data["ca-additional.pem"], nil
```
cbron, line 301: If they mess up secret and don't have this key, will this return an empty string or nil ? Might want to do this check on line 297.

cbron (commented):
basically lgtm, bunch of nits

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -387,42 +387,48 @@ func (e *aksOperatorController) deployAKSOperator() error {
 	return e.DeployOperator(aksOperator, aksOperatorTemplate, aksShortName)
 }
 
+// generateSATokenWithPublicAPI tries to get a service account token from the cluster using the public API endpoint.
+// This function is called if the cluster has only privateEndpoint enabled.
+// If Rancher is able to communicate with the cluster with the API endpoint even though it is private, then this function will retrieve
```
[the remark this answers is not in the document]
thedadams: Done.

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -387,42 +387,48 @@ func (e *aksOperatorController) deployAKSOperator() error {
 	return e.DeployOperator(aksOperator, aksOperatorTemplate, aksShortName)
 }
 
+// generateSATokenWithPublicAPI tries to get a service account token from the cluster using the public API endpoint.
+// This function is called if the cluster has only privateEndpoint enabled.
```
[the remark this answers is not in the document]
thedadams: Done.

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -387,42 +387,48 @@ func (e *aksOperatorController) deployAKSOperator() error {
...
 		return "", nil, err
 	}
 
 	requiresTunnel := new(bool)
-	netDialer := net.Dialer{
-		Timeout:   30 * time.Second,
-		KeepAlive: 30 * time.Second,
-	}
-	serviceToken, err := generateSAToken(restConfig, netDialer.DialContext)
+	restConfig.Dial = (&net.Dialer{Timeout: 30 * time.Second, KeepAlive: 30 * time.Second}).DialContext
```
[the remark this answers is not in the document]
thedadams: Done.

pkg/controllers/management/clusteroperator/utils.go
thedadams: Done.

pkg/controllers/management/clusteroperator/utils.go
thedadams: It returns `nil`. So it's possible this function returns `nil, nil` in a couple different places and the caller would have to handle these possibilities. There are two places where this function is used, and they both handle these possibilities.

cmurphy (approved):

cbron (approved):

Issue #32887, opened 2021-05-26
[ERROR] could not convert gke config to map


mjura:
During updating Rancher cluster from master branch, I am getting following error:

```
2021/05/26 15:00:34 [WARNING] Catalog-manager error deleting old template content: no kind "DeleteOptions" is registered for version "provisioning.cattle.io/v1" in scheme "k8s.io/apiextensions-apiserver/pkg/apiserver/apiserver.go:52"
2021/05/26 15:01:55 [ERROR] could not convert gke config to map
2021/05/26 15:01:55 [ERROR] could not convert gke config to map
2021/05/26 15:01:55 [ERROR] could not convert gke config to map
2021/05/26 15:01:55 [INFO] clusterDeploy: redeployAgent: redeploy Rancher agents due to agent env vars mismatched for [c-xf8vp], was [[]] and will be [[]]
```

Fix for this issue https://github.com/rancher/rancher/pull/32912

mjura:
Fix was merged

prabalsharma:
Repro steps from Michal:

1. Deploy AKS cluster
2. Wait until it will finish successfully
3. Change configuration for example by adding new tags
4. In Rancher logs you will find [ERROR] could not convert gke config to map

prabalsharma:
Tested on master-8cdbffe7fcbd433177ec2e0effd160190d36fa54-head, aks-operator:v1.0.1-rc6

PR #32912, opened 2021-05-27
Fix check for gke and eks configs

mjura:
In Formatter function keys googleKubernetesEngineConfig
and amazonElasticContainerServiceConfig exist, but values
are for them are nil.

Issue https://github.com/rancher/rancher/issues/32887


thedadams (approved):

cmurphy (approved):
I'm not clear on why this happens but it seems like a reasonable safety check.

rmweir (approved):
LGTM

Issue #31214, opened 2021-02-09
Add feature flag for token hashing and make it opt-in functionality


rmweir:
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement

**Description:**
* ~~Add a setting for opting into token hashing.~~ 
* Set opt-in setting to false.
* Implement logic to only hash tokens if user has opted in.
* Once opted in, token hashing should be permanent and non reversible.

sowmyav27:
**On master-head commit id: `4d41301`**

- Deploy a downstream cluster and a kubeconfig token.
- Activate Token hashing
- All tokens get hashed

```
authn.management.cattle.io/token-hashed: "true"
      authn.management.cattle.io/token-userId: u-mo773yttt4
            f:authn.management.cattle.io/token-hashed: {}
            f:authn.management.cattle.io/token-userId: {}
        f:token: {}
  token: $2:6PNSxvFY5hA:h9mD2JTEiOAnuOwuI7fVOxr9NTi7sZpe6nymay7rTN8
```
- Deactivate the token hashing feature. Error seen - `{"baseType":"error","code":"InvalidBodyContent","message":"feature flag cannot be changed from current value: true","status":422,"type":"error"}`

PR #32975, opened 2021-06-01
Tokens toggle

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/31214


pkg/controllers/management/feature/feature_handler.go
```diff
@@ -0,0 +1,100 @@
...
+
+// setLockedValue evaluates whether a value should be written to the lockedValue
+// field on status and records the value if so.
+func (h *handler) setLockedValue(obj *v3.Feature) (*v3.Feature, error) {
+	lockedValue := LockedValue(obj)
+
+	if lockedValue == nil && obj.Status.LockedValue == nil {
+		return obj, nil
+	}
+	update := (lockedValue == nil && obj.Status.LockedValue != nil) || (lockedValue != nil && obj.Status.LockedValue == nil) ||
```
cbron, line 68: Can you comment the cases here, I can't understand the first case. If lockedValue returns nil the feature can be changed (according to that method) so how would you ever get here. The last case is also confusing me.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -0,0 +1,100 @@
...
+	featureCopy := obj.DeepCopy()
+	featureCopy.Status.LockedValue = lockedValue
+	return h.featuresClient.Update(featureCopy)
+}
+
+// LockedValue evaluates whether updates to a feature's effective value
+// shouldl be prevented. If so LockedValue returns the value that should
+// be pinned to a feature. If nil is returned, the features value can be
+// changed and those changes should toggle the associated behavior.
+func LockedValue(obj *v3.Feature) *bool {
```
cbron, line 84: This name is very generic and also used as a var, how about `isFeatureLocked`

rmweir: `isFeatureLocked` is not accurate. It is returning the locked value not whether it is locked. For example if it returns `false`, then the locked value is `false`. `isFeatureLocked` would imply `false` to mean that the feature is not locked.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -0,0 +1,100 @@
...
+	h := handler{
+		featuresClient: wContext.Mgmt.Feature(),
+		tokensLister:   wContext.Mgmt.Token().Cache(),
+		tokenEnqueue:   wContext.Mgmt.Token().EnqueueAfter,
+	}
+	wContext.Mgmt.Feature().OnChange(ctx, "feature-handler", h.sync)
+}
+
+func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
+	if obj == nil || obj.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
rmweir: Locked value seems like a field that should be on status, especially since we have no intention of allowing a user to change it. I don't think putting it on spec would remove the need for the locked value logic unless we counted on the UI to change it which would mean the API wouldn't set it. If we set it in the store we would need logic just like this. This also works if the value is set by kubectl.

pkg/controllers/management/feature/feature_handler.go
rmweir: There is no telling what locked logic other feature flags may use; you can't change spec if there's a locked value due to validator but you could change something else. This is just making sure that LockedValue on status matches the value returned from the function.

pkg/features/feature.go
```diff
@@ -71,12 +71,18 @@ var (
...
+		false,
+		true,
+		true)
 )
 
 type Feature struct {
 	name        string
 	description string
-	// val is the effective value- it is equal to default until explicitly changed
+	// val is the effective value. The order of precedence is lockedValue > value > default
```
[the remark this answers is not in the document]
rmweir: Sure, although the logic is such that there could be a scenario where a feature flag is locked prior to any changes, if such a scenario was added to the LockedValue method, in which case this is false.

pkg/systemtokens/systemtokens.go
```diff
@@ -77,7 +77,7 @@ func (t *systemTokens) createOrUpdateSystemToken(tokenName, description, kind, u
 					tokens.TokenKindLabel: kind,
 				},
 			},
-			TTLMillis:    86400000, // 24 hours, token purge daemon will cleanup
+			TTLMillis:    3600000, // 1 hour, token purge daemon will cleanup
```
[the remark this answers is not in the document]
rmweir: This was part of the original ask, just forgot to do it.

pkg/controllers/management/feature/feature_handler.go
kinarashah: So this makes sense to me: 
- Validator is simply to validate if status.LockedValue is not nil, we allow spec.Value to change only if it's nil.
- setLockedValue is actually setting obj.StatusLockedValue from spec.Value, and we we keep them in sync.  

But I feel like I don't fully grasp what we're allowing features to do with LockedValue. Is the intention that they can introduce and some other field on spec....but on status, it won't be a corresponding new field and it'll reflect on LockedValue? In that case, is LockedValue really locked then?

rmweir: Another reason for this logic is upgrade, it's possible the LockedValue function is updated in a new version and the LockedValue will need to be updated.

pkg/systemtokens/systemtokens.go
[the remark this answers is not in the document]
cbron: Boy, I forgot history of this thing. Originally it was 1 hour, then it got moved to 24, and now back to 1. You remember why we moved it to 24 ? 
https://github.com/rancher/rancher/commit/a213e857b2d5d5931ae74804532133140ad8068d#diff-cc51e2bb0cce2c268bac8a7d6681166b658a7edaca22454b1eedc277e8a4da41R80

pkg/features/feature.go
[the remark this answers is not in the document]
cbron: Ok maybe say `it is equal to default until explicitly changed or locked upon initialization`

pkg/controllers/management/feature/feature_handler.go
cbron: I'm also confused at the real goal of all of this. It seems to me that its all around pinning a spec once a certain value is hit. But the real problem we are trying to solve is: users need to know token-hashing is a one-way feature and they shouldn't be able to break that.
If we can make it so it doesn't matter if they flip the switch back and forth, and "one way" is in the description that would also satisfy the problem. Meaning: On disable this feature we log an error. On enabling this feature if a v3 token has the `hashFormat` we know its already on and we can't redo it and we also log an error. Would UX for that be any different than this ? 
We can go this route if we need to but it seems complicated.

rmweir: The user is already prevented from toggling the value in the UI due to the validator. This is just making it so if for some reason the LockedValue should be different it updates it accordingly. This could happen due to an upgrade.

pkg/systemtokens/systemtokens.go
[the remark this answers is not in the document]
rmweir: May because of the kubectl shell? Which it turns out doesn't use tokens anymore?

pkg/controllers/management/feature/feature_handler.go
cbron: Add something like this in the comment
```
Return values
*  nil - not currently locked
* false - currently locked and false value
* true - currently locked and true value
```
And func rename to something like: `EvaluateLockedValueFromSpec`

pkg/controllers/management/feature/feature_handler.go
cbron: Talked offline and the real reason is so we can flip the code on/off based on the feature flag, rather than based on token string.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, nil
+	}
+
+	obj, err := h.setLockedValue(obj)
+	if err != nil {
+		return obj, err
+	}
+
+	if obj.Name == features.TokenHashing.Name() {
```
cbron, line 40: It looks below like you call `refreshTokens` even if it wasn't updated ? 
```
	if !update {		return obj, nil	}
```
error is nil so this still gets hit, why refresh ?

rmweir: Because the update check is for for LockedValue where you want to refresh tokens if the effective value of the feature flag is true. This could be the regular Value field or LockedValue.

cbron (dismissed):
lgtm, we'll see what QA says about the systemToken TTL

kinarashah (approved):

nickgerace:
Rebased this PR due to conflict at the request of @cbron

kinarashah (approved):

cbron (approved):

Issue #32938, opened 2021-05-28
Reduce size of v3.Catalog objects


nickgerace:
## Summary

**This issue targets `v3.Catalog` objects.**

- sibling issue targeting `v3.App` objects: https://github.com/rancher/rancher/issues/32939
- parent issue that has been closed to focus on the child issues: https://github.com/rancher/rancher/issues/32418

### What kind of request is this? (question/bug/enhancement/feature request)
enhancement

### Description
Catalogs and apps can be enormous in size causes performance issues and other bugs. The [bitnami](https://charts.bitnami.com/bitnami) chart for example is about 1mb in size. For catalog size is largely due to recording the hash of each version of each charts on its status, at least for non git catalogs. One solution is to gzip certain fields, like chart version hashes, or only record the hash for the latest version of each chart on the status (both examples apply to catalogs specifically). Large object size in kubernetes 1.20 is leading to timeouts and high cpu consumption spikes, see related issues.

### Related issues
https://github.com/rancher/rancher/issues/32175
https://github.com/rancher/rancher/issues/31940
https://github.com/rancher/rancher/issues/31614
https://github.com/rancher/rancher/issues/32295

### Other details that may be helpful
Once solutions are implemented may want to tryout adding the bitnami catalog while using a 1.20 management cluster.

nickgerace:
## QA Validation Recommendations

The resolution of this issue will not require checking how much the `v3.Catalog` CR sizes have been reduced. Rather, validation of the solution will be dependent on catalogs working as intended. If catalogs, project catalogs, and cluster catalogs work as intended via creation, updates, deletion, and app usage, then the solution is valid.

nickgerace:
This will require a backport to 2.5.x.

nickgerace:
Related: https://github.com/rancher/rancher/issues/32419#issuecomment-858136129

kravciak:
Adding / removing catalogs and installing apps from them works fine for both local and rke1 clusters.

Tested on master-head (3dae3e2)

anupama2501:
Verified on v2.5.9-rc8

- Created a catalog at global level , cluster level  and project level 
- Installed mySql app from cluster, project, global in 3 different projects.
- All apps have come up fine and active in all the projects.

PR #32992, opened 2021-06-02
Deprecate helmVersionCommits from v3.Catalog

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/32938


nickgerace:
All builds pass except for Windows 2004. Appears to be an outlier: https://drone-pr.rancher.io/rancher/rancher/16036/5/2

pkg/catalog/manager/traverse.go
```diff
@@ -20,7 +20,36 @@ import (
...
+		if projectCatalog.Status.HelmVersionCommits == nil {
+			return nil
+		}
+		projectCatalog.Status.HelmVersionCommits = nil
+	}
+	_, err := m.updateCatalogInfo(cmt, catalogType, "", true, true)
+	if err != nil {
+		return err
+	}
+	return nil
```
nickgerace, line 48: This can be simplified to `return err`, but might want to be kept for readability. Leaving this open for reviewers' thoughts here.

pkg/apis/management.cattle.io/v3/catalog_types.go
```diff
@@ -35,11 +35,13 @@ type CatalogSpec struct {
...
-	LastRefreshTimestamp string `json:"lastRefreshTimestamp,omitempty"`
-	Commit               string `json:"commit,omitempty"`
-	// helmVersionCommits records hash of each helm template version
+	LastRefreshTimestamp string             `json:"lastRefreshTimestamp,omitempty"`
+	Commit               string             `json:"commit,omitempty"`
+	Conditions           []CatalogCondition `json:"conditions,omitempty"`
+
+	// Deprecated: should no longer be in use. If a Catalog CR is encountered with this field
+	// populated, it will be set to nil.
 	HelmVersionCommits map[string]VersionCommits `json:"helmVersionCommits,omitempty"`
```
nickgerace, line 44: @StrongMonkey recommended that we remove these deprecation notices since the `status` is not strictly part of the user-consumable API (`spec`). Leaving this open for reviewer's thoughts.

pkg/apis/management.cattle.io/v3/catalog_types.go
```diff
@@ -64,6 +66,7 @@ type CatalogCondition struct {
 	Message string `json:"message,omitempty"`
 }
 
+// Deprecated: CatalogStatus.HelmVersionCommits has been deprecated, which is the only consumer of this type.
```
nickgerace, line 69: Same as: https://github.com/rancher/rancher/pull/32992#discussion_r645145209
Leaving this open for reviewer's thoughts.

pkg/catalog/manager/traverse.go
```diff
@@ -54,35 +90,39 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
+				return true
+			}
 		}
+		return false
+	}
 
-		if !hasChanged && hasAllUpdates(catalog) {
+	entriesWithErrors, entriesToProcess := m.preprocessCatalog(catalog, index.IndexFile.Entries)
+	for chart, metadata := range entriesToProcess {
+		if !hasChartChanged(catalogTemplateMap[getValidTemplateName(catalogName, chart)], metadata) && hasAllUpdates(catalog) {
```
StrongMonkey, line 125: This is going to print logs multiple times. We should save it as a boolean and print log in the end.

StrongMonkey: Also you don't have to call `hasAllUpdates(catalog)` in the loop

pkg/catalog/manager/traverse.go
```diff
@@ -54,35 +90,39 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
-		newHelmVersionCommits[chart] = v32.VersionCommits{
-			Value: map[string]string{},
+	catalogTemplateMap, err := m.getTemplateMap(catalogName, templateNamespace)
+	if err != nil {
+		return err
+	}
+
+	// Check if a version has been deleted from a template or if an existing template version has changed.
+	hasChartChanged := func(template *v3.CatalogTemplate, metadata helmlib.ChartVersions) bool {
+		if template == nil || metadata == nil {
```
StrongMonkey, line 100: We should return true here? If a version exist in metadata but not in template, the chart has changed(an entry has been deleted)

StrongMonkey (commented):
Overall approach looks good and I really like it. There are small nits that need to be addressed. I am leaning towards deleting the field from status, unless people have concerns.

pkg/catalog/manager/traverse.go
nickgerace: > This is going to print logs multiple times. We should save it as a boolean and print log in the end.

The anonymous function does not print any logs unless I am missing something.

> you don't have to call hasAllUpdates(catalog) in the loop

The `hasAllUpdates` call was from the previous code. I will call it before the loop begins and store the result in a variable for optimization.

pkg/catalog/manager/traverse.go
nickgerace: This is a good point. I will upgrade this logic for that scenario.

nickgerace:
> Overall approach looks good and I really like it. There are small nits that need to be addressed. I am leaning towards deleting the field from status, unless people have concerns.

Thank you! What happens to existing users that rely on the field though? The current code accounts for that by dropping any values found.

StrongMonkey:
For existing catalogs that uses it, during the upgrade it will just use new way to detect changes. During the upgrade all the data in that field is going to be wiped out, so it doesn't make a difference whether we keep it or not.

pkg/catalog/manager/traverse.go
```diff
@@ -20,7 +20,36 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+func (m *Manager) dropDeprecatedFields(cmt *CatalogInfo, catalogType string) error {
```
snasovich, line 23: Does it make sense to move this new helper func to somewhere after `traverseAndUpdate`? Another helper func called by `traverseAndUpdate` (`processInvalidChartErrors`) is all the way at the bottom of the file.

pkg/catalog/manager/traverse.go
StrongMonkey: https://github.com/rancher/rancher/pull/32992/files/63dd97c65fa60840f208a8edb74cf5c27eba5742#diff-756f710464390ef227414411a01a1f9d7fca7694ffdfd046c22248272412e479R126 this line actually prints it if you turn on debug. We should just print it once

pkg/catalog/manager/traverse.go
nickgerace: I will move it

pkg/catalog/manager/traverse.go
nickgerace: Ah good catch. This was from the original code, but I'll move it.

pkg/catalog/manager/traverse.go
nickgerace: Decided to make the optimization

pkg/catalog/manager/traverse.go
nickgerace: Fixed both

pkg/catalog/manager/traverse.go
nickgerace: The length check covers your case, but this line should be `&&` instead of `||`.

pkg/catalog/manager/traverse.go
```diff
@@ -54,36 +62,42 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
-			Value: map[string]string{},
+	catalogTemplateMap, err := m.getTemplateMap(catalogName, templateNamespace)
+	if err != nil {
+		return err
+	}
+
+	// Check if a version has been deleted from a template or if an existing template version has changed.
+	hasChartChanged := func(template *v3.CatalogTemplate, metadata helmlib.ChartVersions) bool {
+		if template == nil && metadata == nil {
+			return false
```
StrongMonkey, line 73: Hmm should this return true as we discussed? If template is missing(which means template is nil or metadata == nil) then return true. Also you will likely get a panic if template is nil but metadata is not nil. I think this should be `or` condition?

StrongMonkey: I am thinking that we should the logic should be...
If template == nil:
Then means a new template has been added. Metadata is not likely to be nil I think. but if it is nil we can probably return false because a chart has been deleted from upstream then we can skip it and do nothing.

nickgerace: Spoke offline and agree on this approach. `metadata` should never be `nil`, so an additional "safety" check has been added to the parent function.

StrongMonkey (dismissed):

pkg/catalog/manager/traverse.go
```diff
@@ -316,6 +334,31 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
+		}
+		clusterCatalog.Status.HelmVersionCommits = nil
+	case client.ProjectCatalogType:
+		projectCatalog := cmt.projectCatalog
+		if projectCatalog.Status.HelmVersionCommits == nil {
+			return nil
+		}
+		projectCatalog.Status.HelmVersionCommits = nil
+	}
+	_, err := m.updateCatalogInfo(cmt, catalogType, "", true, true)
```
rmweir, line 358: Due to the way the function is written, this should be fine but I think if a function returns an updated version of an object you're passing in you should usually set it. In this case you may was to assign it to another variable and then do something like `*cmt = *cmt2`.

pkg/catalog/manager/traverse.go
```diff
@@ -54,36 +62,42 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
-			digest, ok := existingHelmVersionCommits[version.Version]
-			if !ok || digest != version.Digest {
-				hasChanged = true
-			}
-			if ok {
-				versionNumber++
+		for _, templateVersion := range template.Spec.Versions {
+			// If the digest is not the same between the actual and desired version, or if the version does not
+			// exist in the metadata, then we know that the chart has changed. In addition, we do not have to check
+			// if metadata versions exist because the length of each slice is the same at this point.
```
rmweir, line 83: Couldn't a version be removed and another added leading to equal lengths: `because the length of each slice is the same at this point.`?

rmweir (dismissed):
Some small comments, otherwise looking clean 👍 .

pkg/catalog/manager/traverse.go
```diff
@@ -54,36 +62,42 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
...
+			digest, ok := metadataVersions[templateVersion.Version]
+			if digest != templateVersion.Digest || !ok {
```
nickgerace, line 85: Move `ok` check to be first

pkg/catalog/manager/traverse.go
```diff
@@ -236,15 +252,19 @@ func (m *Manager) traverseAndUpdate(helm *helmlib.Helm, commit string, cmt *Cata
 				updatedTemplates++
 			}
 		}
-		if err != nil {
-			delete(newHelmVersionCommits, template.Spec.DisplayName)
-		}
+	}
+	if skippedCharts != nil {
+		logrus.Debugf("skipped generating templates for charts that have not been changed: %v", skippedCharts)
 	}
```
nickgerace, line 258: Remove check around statement. Logging `nil` value is fine.

pkg/apis/management.cattle.io/v3/catalog_types.go
nickgerace: Spoke to @dramich: we agree that these are fine to leave.

pkg/apis/management.cattle.io/v3/catalog_types.go
nickgerace: Resolved by: https://github.com/rancher/rancher/pull/32992/files#r647739404

pkg/catalog/manager/catalog_common.go
```diff
@@ -201,6 +201,16 @@ func setTraverseCompleted(catalog *v3.Catalog) {
 	v32.CatalogConditionProcessed.Reason(catalog, "")
 }
 
+// This function is the inverse of "getValidTemplateName". It does not check for correctness.
```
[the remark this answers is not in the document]
nickgerace: I'll condense the function name to something that stands on its own and ensure it meets godoc compatibility.

pkg/catalog/manager/traverse.go
nickgerace: Is this not solved by line 72?

pkg/catalog/manager/catalog_common.go
```diff
@@ -201,6 +201,16 @@ func setTraverseCompleted(catalog *v3.Catalog) {
...
+// An empty string will be returned the catalogName input is invalid.
+func getChartNameFromValidTemplateName(catalogName, templateName string) string {
```
nickgerace, line 206: Rename this to something that stands on its own

pkg/catalog/manager/traverse.go
nickgerace: When we set `updateOnly` to true, the catalog is eaten by the `updateCatalogInfo` method. This is also done multiple times throughout the traversal function. This is intentional, but I would normally agree if this was another circumstance.

Confirmed with @dramich

dramich (approved):

nickgerace:
Thanks @dramich

Issue #32179, opened 2021-04-16
Remove rancher-operator 


thedadams:
rancher-operator is no longer needed in Rancher v2.6+

sowmyav27:
To test:
- Fresh install of master-head
- Upgrade from 2.5.8 to master-head

sowmyav27:
@thedadams Can you let us know what resources make use of rancher-operator so that we can test an upgrade usecase and also the rollback from 2.6 to 2.5.8 (for instance) works?

sangeethah:
from @cbron on workflows that uses rancher-operator:
Cluster is successfully registered in fleet once cluster is provisioned.
Clusters is assigned to the right fleet workspace(by default it is fleet-default), local cluster should be in fleet-local.
Cluster label from rancher gets populated into fleet cluster.

anupama2501:
Verified on master-head commit d60543a
Fresh install
1. Once the cluster is created, from cluster explorer --> continuous delivery --> clusters --> fleet-default has the clusters registered but not the local cluster
2. cluster explorer --> continuous delivery --> clusters --> fleet-local has the local cluster and the default cluster group populated in the cluster groups
<img width="500" alt="Screen Shot 2021-05-24 at 10 22 23 AM" src="https://user-images.githubusercontent.com/38144301/119384718-fefd2600-bc79-11eb-9dbc-38e5b3c68baf.png">

anupama2501:
@thedadams I have verified the above mentioned scenarios on a fresh install. can you let me know if I need to verify any other test cases?

anupama2501:
Re-opening the issue:
On an upgrade from rancher version v2.5.8 to master-head commit d60543a:
rancher-operator is still seen in the local cluster workloads.

from cluster explorer --> local --> workloads --> rancher_operator should be removed after upgrading from v2.5.8
<img width="945" alt="Screen Shot 2021-05-24 at 11 21 01 AM" src="https://user-images.githubusercontent.com/38144301/119393149-6ec4de00-bc85-11eb-92b2-2196f7d391b9.png">

anupama2501:
On v2.5.8 --> Created 2 rke1 node driver clusters from ember UI

- Verified rancher-operator is present in the local --> workloads

Upgraded to master-head d44698d:
- Verified the `rancher-operator` is removed from the workloads 

But the rke1 clusters are in `Modified` state with the following error in conditions:
```
[...]

PR #32996, opened 2021-06-02
Uninstall objects from charts in non-default namespaces

thedadams:
If the objects of a chart were installed in a non-default namespace
without the namespace being listed in the object's metadata, then the
uninstalled would seem successful but the helm is not able to find the
objects. The non-namespaced items would be removed, but any namespaced
items would be leftover.

It should be noted that the namespace provided allowed helm to find the
release secret, but would not namespace the call to delete the objects
in the release.

Issue:
https://github.com/rancher/rancher/issues/32179

### DIFF
--- a/pkg/catalogv2/system/system.go
+++ b/pkg/catalogv2/system/system.go
@@ -123,10 +123,19 @@ func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
 }
 
 func (m *Manager) Uninstall(namespace, name string) error {
+	if ok, err := m.hasStatus(namespace, name, action.ListDeployed|action.ListFailed); err != nil {
+		return err
+	} else if !ok {
+		return nil
+	}
+
 	uninstall, err := json.Marshal(types.ChartUninstallAction{
 		Timeout: &metav1.Duration{Duration: 5 * time.Minute},
 	})
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
 		return err
 	}
 
@@ -174,22 +183,10 @@ func (m *Manager) install(namespace, name, minVersion string, values map[string]
 		return nil
 	}
 
-	if ok, err := m.isPendingUninstall(namespace, name); err != nil {
+	if ok, err := m.hasStatus(namespace, name, action.ListPendingInstall); err != nil {
 		return err
 	} else if ok {
-		uninstall, err := json.Marshal(types.ChartUninstallAction{
-			Timeout: &metav1.Duration{Duration: 5 * time.Minute},
-		})
-		if err != nil {
-			return err
-		}
-
-		op, err := m.operation.Uninstall(m.ctx, installUser, namespace, name, bytes.NewBuffer(uninstall))
-		if err != nil {
-			return err
-		}
-
-		if err := m.waitPodDone(op); err != nil {
+		if err = m.Uninstall(namespace, name); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/catalogv2/system/system.go
```diff
@@ -122,27 +122,26 @@ func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
...
 
-	if len(releases) == 0 {
-		return nil
+	if wait {
+		if err = m.waitPodDone(op); err != nil {
+			return err
+		}
 	}
 
-	uninstall := action.NewUninstall(helmcfg)
```
StrongMonkey, line 143: Does it mean that when using this library to install, it doesn't clean up namespaced items? I am not sure if there is a difference between using library or using pod to uninstall.

thedadams: If the objects in the helm release had the namespace in their metadata, then it would clean them up.

However, if the objects did not have a namespace in their metadata and were installed in a non-default namespace, then objects were not getting cleaned up.

There is a way to use the library, but this seemed cleaner. If you think the library should continued to be used, then I can update this PR with that approach instead.

StrongMonkey: Ok fine by either way. Using library might be fast since it doesn't need to involve creating pods, but it is just one time uninstall so it shouldn't matter that much.

StrongMonkey (dismissed):

nickgerace (dismissed):

pkg/catalogv2/system/system.go
StrongMonkey: @thedadams Actually I think we need to check if errors.isnotfound(). I am getting 
```
error syncing 'rancher-charts': handler bootstrap-charts: apps.catalog.cattle.io "rancher-operator" not found, requeuing
```
when trying to use your method. I happen to test something else with your PR.

thedadams: @StrongMonkey Added check for "not found" error.

StrongMonkey (approved):

pkg/catalogv2/system/system.go
```diff
@@ -123,26 +123,19 @@ func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
...
+		Timeout: &metav1.Duration{Duration: 5 * time.Minute},
+	})
 	if err != nil {
 		return err
 	}
 
-	if len(releases) == 0 {
-		return nil
+	op, err := m.operation.Uninstall(m.ctx, installUser, namespace, name, bytes.NewBuffer(uninstall))
+	if err != nil && !apierrors.IsNotFound(err) {
```
StrongMonkey, line 134: Actually this should be 
```
if err != nil {
  if apierrors.IsNotFound(err) {
    return nil
  }
  return err
}

Comment (thedadams) [reply pkg/catalogv2/system/system.go]: Done.

File: pkg/catalogv2/system/system.go [hunk 3]
```diff
@@ -123,26 +123,22 @@ func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
 }
 
 func (m *Manager) Uninstall(namespace, name string) error {
-	helmcfg := &action.Configuration{}
```
cbron, line 126: Why remove all the list/check code ? The install method below still checks that inside `isPendingUninstall`, which I'm assuming is there so it doesn't error out. Must be some reason it existed.

cbron (approved):

pkg/catalogv2/system/system.go
```diff
@@ -123,10 +123,19 @@ func (m *Manager) installCharts(charts map[desiredKey]map[string]interface{}) {
 }
 
 func (m *Manager) Uninstall(namespace, name string) error {
+	if ok, err := m.hasStatus(namespace, name, action.ListDeployed|action.ListFailed); err != nil {
```
rohitsakala, line 126: Why was this added ? this condition ?

Issue #32941, opened 2021-05-28
Keycloak OIDC - Can't login with users after the auth is enabled


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Enable Keycloak OIDC on latest master build (used _62aef76_ (05/27/2021))
- Open a separate browser and with a valid user click Login with Keycloak

**Result:**

- Blank page after attempting login received a 401 from `/v1/userpreferences`

URL stuck here

`/dashboard/auth/verify?state=<hash>&code=uss.<more code>`

**Other details that may be helpful:**

TRACE LOG:

`[ERROR] API error response 500 for POST /v3-public/keyCloakOIDCProviders/keycloakoidc?action=login. Cause: Get "/.well-known/openid-configuration": unsupported protocol scheme ""`

**Environment information**
- Rancher version: `2.6 master-head (05/27/2021)` _62aef76_
- Installation option: HA RKE1

izaac:
Rancher version `2.6 master-head (06/03/2021)` _be03c8bd1_

I am able to login with standard user and mapped admin principal

Issue #32936, opened 2021-05-28
Can't disable Keycloak OIDC action not available


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Enable Keycloak OIDC
- Then click `Disable` 

**Result:**

Action not available

`POST /v3/keyCloakOIDCConfigs/keycloakoidc?action=disable`

`{"baseType":"error","code":"ActionNotAvailable","message":"","status":404,"type":"error"}`

**Other details that may be helpful:**

![Screen Shot 2021-05-28 at 6 29 24 AM](https://user-images.githubusercontent.com/27659/119991393-40454c80-bf7e-11eb-8039-cc39a1bfcfe8.png)

**Environment information**
- Rancher version: `2.6 master-head (05/27/2021)` _62aef76_
- Installation option: HA RKE1

Related: 
https://github.com/rancher/rancher/issues/30280
https://github.com/rancher/dashboard/issues/2191

izaac:
Rancher version `2.6 master-head (06/03/2021)` _be03c8bd1_

I am able to enable, disable and re-enable the auth on a build from today.

PR #33002, opened 2021-06-02
Use name and type string from struct for keycloak oidc & oidc

brendarearden:
**Issue(s)**
https://github.com/rancher/rancher/issues/32990
https://github.com/rancher/rancher/issues/32941
https://github.com/rancher/rancher/issues/32936

**Problem**
The type and name are not being set properly with `.GetName()` and `.GetType()` which is resulting in both Keycloak OIDC and OIDC providers to become enabled. Both of these being enabled at the same time causes

1. The login screen not to know which auth provider to show to the user
2. Once they become enabled, there is no way to disable them because the disable doesn't have the correct Name of the provider to disable it. 

**Solution**
Set the name and type values on the OIDCProvider struct so that the values are correct at time of configuration and it doesn't rely on inheritance to call the correct function.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_provider.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_provider.go
@@ -36,6 +36,8 @@ func Configure(ctx context.Context, mgmtCtx *config.ScaledContext, userMGR user.
 			httpClient: &http.Client{},
 		},
 		oidc.OpenIDCProvider{
+			Name:        Name,
+			Type:        client.KeyCloakOIDCConfigType,
 			CTX:         ctx,
 			AuthConfigs: mgmtCtx.Management.AuthConfigs(""),
 			Secrets:     mgmtCtx.Core.Secrets(""),
@@ -47,10 +49,6 @@ func Configure(ctx context.Context, mgmtCtx *config.ScaledContext, userMGR user.
 
 func (k *keyCloakOIDCProvider) GetName() string {
 	return Name
-}
-
-func (k *keyCloakOIDCProvider) GetType() string {
-	return client.KeyCloakConfigType
 }
 
 func (k *keyCloakOIDCProvider) SearchPrincipals(searchValue, principalType string, token v3.Token) ([]v3.Principal, error) {
--- a/pkg/auth/providers/oidc/oidc_actions.go
+++ b/pkg/auth/providers/oidc/oidc_actions.go
@@ -22,7 +22,7 @@ func (o *OpenIDCProvider) Formatter(apiContext *types.APIContext, resource *type
 }
 
 func (o *OpenIDCProvider) ActionHandler(actionName string, action *types.Action, request *types.APIContext) error {
-	handled, err := common.HandleCommonAction(actionName, action, request, o.GetName(), o.AuthConfigs)
+	handled, err := common.HandleCommonAction(actionName, action, request, o.Name, o.AuthConfigs)
 	if err != nil {
 		return err
 	}
@@ -42,7 +42,7 @@ func (o *OpenIDCProvider) ActionHandler(actionName string, action *types.Action,
 func (o *OpenIDCProvider) ConfigureTest(actionName string, action *types.Action, request *types.APIContext) error {
 	//verify body has all required fields
 	input, err := handler.ParseAndValidateActionBody(request, request.Schemas.Schema(&managementschema.Version,
-		o.getType()))
+		o.Type))
 	if err != nil {
 		return err
 	}
--- a/pkg/auth/providers/oidc/oidc_provider.go
+++ b/pkg/auth/providers/oidc/oidc_provider.go
@@ -33,6 +33,8 @@ const (
 )
 
 type OpenIDCProvider struct {
+	Name        string
+	Type        string
 	CTX         context.Context
 	AuthConfigs v3.AuthConfigInterface
 	Secrets     corev1.SecretInterface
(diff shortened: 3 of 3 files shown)


pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -255,10 +255,10 @@ func (o *OpenIDCProvider) userToPrincipal(userInfo *oidc.UserInfo, info claimInf
 		displayName = userInfo.Email
```
cbron, line 255: Nit: use arg variable name of `claiminfo` instead of `info`. As both those variables have info in their name.

pkg/auth/providers/oidc/oidc_actions.go
```diff
@@ -42,7 +42,7 @@ func (o *OpenIDCProvider) ActionHandler(actionName string, action *types.Action,
 func (o *OpenIDCProvider) ConfigureTest(actionName string, action *types.Action, request *types.APIContext) error {
 	//verify body has all required fields
 	input, err := handler.ParseAndValidateActionBody(request, request.Schemas.Schema(&managementschema.Version,
-		o.getType()))
+		o.Type))
```
cbron, line 45: So this all gets inherited by keycloak OIDC ? And therefore O is of keycloak type.

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -169,11 +169,11 @@ func (o *OpenIDCProvider) SearchPrincipals(searchValue, principalType string, to
...
 
 	p := v3.Principal{
-		ObjectMeta:    metav1.ObjectMeta{Name: o.GetName() + "_" + principalType + "://" + searchValue},
+		ObjectMeta:    metav1.ObjectMeta{Name: o.Name + "_" + principalType + "://" + searchValue},
 		DisplayName:   searchValue,
 		LoginName:     searchValue,
 		PrincipalType: principalType,
-		Provider:      o.GetName(),
+		Provider:      o.Name,
 	}
```
cbron, line 177: Would be nice to re-use the `toPrinciples` methods below but might be too complicated here since it looks like you are using both and you don't have necessary info. 
Unrelated to this PR, but I am confused why you just return this directly with the search value included, its just a placeholder ?

paynejacob (dismissed):

pkg/auth/providers/oidc/oidc_actions.go
brendarearden: Yes, the `o` at this point is the `OpenIDCProvider` on the `keyCloakOIDCProvider` struct. This issue that this fixes was  was when functions in the OpenIDCProvider interface were being called, it had no idea about the "overriden"  `o.getType()` that I had created in `keyCloakOIDCProvider` so it wasn't returning OpenIDCProvider values and not the keyCloakOIDCProvder values. This meant some types and names were getting set wrong ultimately leading to both providers getting enabled with the rest of the problems that come with it. By setting the value as string, it ensures the value will be available regardless of if its called in the interface or if I change the func definition in the `keyCloakOIDCPRovider` struct. There could be better ways to resolve this, but this seemed like a good solution for now.

pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -87,10 +85,10 @@ func (k *keyCloakOIDCProvider) toPrincipal(principalType string, acct account, t
 		displayName = acct.Username
 	}
 	princ := v3.Principal{
-		ObjectMeta:  metav1.ObjectMeta{Name: k.GetName() + "_" + principalType + "://" + strconv.Itoa(acct.ID)},
+		ObjectMeta:  metav1.ObjectMeta{Name: k.Name + "_" + principalType + "://" + strconv.Itoa(acct.ID)},
 		DisplayName: displayName,
 		LoginName:   acct.Username,
-		Provider:    k.GetName(),
+		Provider:    k.Name,
```
prachidamle, line 91: k.GetName() returns the Name itself, so why this change?

pkg/auth/providers/oidc/oidc_provider.go
prachidamle: I think this is because  OIDC does not support searching?

pkg/auth/providers/keycloakoidc/keycloak_provider.go
brendarearden: I changed it here to match how I am using Name and Type in the oidc_provider. But I do like the idea of using GetName() in the specific providers so I will revert this.

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: updated `info` to `claimInfo` to match the `userInfo` variable

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: Correct - This is how our SAML providers are currently handling `SearchPrincipals` because they do not support search functionality for the various SAML providers.

The other option would be a `return nil, errors.New("[generic oidc]: not implemented")` instead of implementing it.

pkg/auth/providers/oidc/oidc_actions.go
cbron: I had to brush up my memory here to remember: https://play.golang.org/p/IGMYjG9mQbQ

> Embedding types introduces the problem of name conflicts but the rules to resolve them are simple. First, a field or method X hides any other item X in a more deeply nested part of the type. If log.Logger contained a field or method called Command, the Command field of Job would dominate it.

https://golang.org/doc/effective_go#embedding

If you added a keycloak.printer there it would only return that one. So in reverse if you dropped the GetType from Keycloak it would refer back to the OIDC one. 

Using strings avoids this whole confusion.

cbron (approved):
lgtm

prachidamle (approved):
LGTM

Issue #31592, opened 2021-03-04
Move KEv2 operators from system-charts to feature-charts


cbron:
With the migration from v1 to v2 for app deployments, we need to move the these operators from system-charts to feature-charts. 

We need to support the upgrade from system-charts to feature-charts

aaronRancher:
On Rancher `master-head` `version bfdeb22e9`, I am unable to deploy an AKS cluster.

From Ember or Dashboard UI, navigate to AKS cluster create.
Enter cluster resource group and DNS Prefix.
Leave remaining fields default
Create cluster.

Rancher server panics when user tries to deploy an AKS cluster:
```
k8s.io/apimachinery/pkg/util/wait.Until(0xc0032e9420, 0x3b9aca00, 0xc0017016e0)
	/go/pkg/mod/k8s.io/apimachinery@v0.21.0/pkg/util/wait/wait.go:90 +0x4d
created by github.com/rancher/lasso/pkg/controller.(*controller).run
	/go/pkg/mod/github.com/rancher/lasso@v0.0.0-20210616224652-fc3ebd901c08/pkg/controller/controller.go:134 +0x33b
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x34c9f1b]

goroutine 4280 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.21.0/pkg/util/runtime/runtime.go:55 +0x109
panic(0x4899ac0, 0x83202f0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
```

aaronRancher:
On Rancher `master-head` `version 6757965`, I see Rancher panic when creating a GKE cluster.

From Ember or Dashboard UI, navigate to GKE cluster create.
Provide cloud credential and node pool name.
Leave all other fields default.

Rancher server panics with the same error as above:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x34ca1bb]
```
[...]

PR #33004, opened 2021-06-03
Migrate kev2 to feature charts

StrongMonkey:
This PR migrates kev2 install from system-chart to feature charts with ability to support on-demand install and min version. 

On-demand install: operator will only get installed if kev2 cluster is created.
Min version:  https://github.com/rancher/rancher/issues/30934

~Note: This PR relies on a fork of helm to add ability to adopt resources from helm v2 in order to handle upgrade scenario https://github.com/StrongMonkey/helm/commits/v3.5.4-rancher. I will try to get this into shell images.~ Helm changes are merged and included in rancher/shell:v1.0.7

https://github.com/rancher/rancher/issues/31592

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -46,6 +46,9 @@ ENV CATTLE_SYSTEM_AGENT_UPGRADE_IMAGE rancher/system-agent:${CATTLE_SYSTEM_AGENT
 # System charts minimal version
 ENV CATTLE_FLEET_MIN_VERSION=100.0.0+up0.3.6
 ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=100.0.0+up0.2.0-alpha01
+ENV CATTLE_AKS_OPERATOR_MIN_VERSION=100.0.0+1.0.100
+ENV CATTLE_EKS_OPERATOR_MIN_VERSION=100.0.0+1.1.100
+ENV CATTLE_GKE_OPERATOR_MIN_VERSION=100.0.0+1.1.100
 
 RUN mkdir -p /var/lib/rancher-data/local-catalogs/system-library && \
     mkdir -p /var/lib/rancher-data/local-catalogs/library && \
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -46,9 +46,6 @@ ENV CATTLE_SYSTEM_AGENT_UPGRADE_IMAGE rancher/system-agent:${CATTLE_SYSTEM_AGENT
 # System charts minimal version
 ENV CATTLE_FLEET_MIN_VERSION=100.0.0+up0.3.6
 ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=100.0.0+up0.2.0-alpha01
-ENV CATTLE_AKS_OPERATOR_MIN_VERSION=100.0.0+1.0.100
-ENV CATTLE_EKS_OPERATOR_MIN_VERSION=100.0.0+1.1.100
-ENV CATTLE_GKE_OPERATOR_MIN_VERSION=100.0.0+1.1.100
 
 RUN mkdir -p /var/lib/rancher-data/local-catalogs/system-library && \
     mkdir -p /var/lib/rancher-data/local-catalogs/library && \
--- a/pkg/api/steve/catalog/types/rest.go
+++ b/pkg/api/steve/catalog/types/rest.go
@@ -47,6 +47,7 @@ type ChartUpgradeAction struct {
 	DisableHooks             bool             `json:"noHooks,omitempty"`
 	DisableOpenAPIValidation bool             `json:"disableOpenAPIValidation,omitempty"`
 	Force                    bool             `json:"force,omitempty"`
+	ForceAdopt               bool             `json:"forceAdopt,omitempty"`
 	MaxHistory               int              `json:"historyMax,omitempty"`
 	Install                  bool             `json:"install,omitempty"`
 	Namespace                string           `json:"namespace,omitempty"`
--- a/pkg/api/steve/catalog/types/rest.go
+++ b/pkg/api/steve/catalog/types/rest.go
@@ -53,7 +53,6 @@ type ChartUpgradeAction struct {
 	Namespace                string           `json:"namespace,omitempty"`
 	CleanupOnFail            bool             `json:"cleanupOnFail,omitempty"`
 	Charts                   []ChartUpgrade   `json:"charts,omitempty"`
-	ForceAdopt               bool             `json:"forceAdopt,omitempty"`
 }
 
 type ChartUpgrade struct {
--- a/pkg/controllers/dashboard/controller.go
+++ b/pkg/controllers/dashboard/controller.go
@@ -8,7 +8,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/dashboard/clusterregistrationtoken"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/fleetcharts"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/helm"
-	"github.com/rancher/rancher/pkg/controllers/dashboard/hostedkubernetecharts"
+	"github.com/rancher/rancher/pkg/controllers/dashboard/hostedcluster"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/kubernetesprovider"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/mcmagent"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/scaleavailable"
@@ -36,7 +36,7 @@ func Register(ctx context.Context, wrangler *wrangler.Context) error {
 	if err := systemcharts.Register(ctx, wrangler); err != nil {
 		return err
 	}
-	hostedkubernetecharts.Register(ctx, wrangler)
+	hostedcluster.Register(ctx, wrangler)
 
 	if features.Fleet.Enabled() {
 		if err := fleetcharts.Register(ctx, wrangler); err != nil {
--- a/pkg/controllers/dashboard/controller.go
+++ b/pkg/controllers/dashboard/controller.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/dashboard/clusterregistrationtoken"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/fleetcharts"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/helm"
+	"github.com/rancher/rancher/pkg/controllers/dashboard/hostedkubernetecharts"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/kubernetesprovider"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/mcmagent"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/scaleavailable"
@@ -35,6 +36,7 @@ func Register(ctx context.Context, wrangler *wrangler.Context) error {
 	if err := systemcharts.Register(ctx, wrangler); err != nil {
 		return err
 	}
+	hostedkubernetecharts.Register(ctx, wrangler)
 
 	if features.Fleet.Enabled() {
 		if err := fleetcharts.Register(ctx, wrangler); err != nil {
--- a/pkg/controllers/dashboard/hostedcluster/controller.go
+++ b/pkg/controllers/dashboard/hostedcluster/controller.go
@@ -16,34 +16,28 @@ import (
 
 var (
 	AksCrdChart = chart.Definition{
-		ReleaseNamespace:  "cattle-system",
-		ChartName:         "rancher-aks-operator-crd",
-		MinVersionSetting: settings.AKSOperatorMinVersion,
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-aks-operator-crd",
 	}
 	AksChart = chart.Definition{
-		ReleaseNamespace:  "cattle-system",
-		ChartName:         "rancher-aks-operator",
-		MinVersionSetting: settings.AKSOperatorMinVersion,
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-aks-operator",
 	}
 	EksCrdChart = chart.Definition{
-		ReleaseNamespace:  "cattle-system",
-		ChartName:         "rancher-eks-operator-crd",
-		MinVersionSetting: settings.EKSOperatorMinVersion,
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-eks-operator-crd",
 	}
 	EksChart = chart.Definition{
-		ReleaseNamespace:  "cattle-system",
-		ChartName:         "rancher-eks-operator",
-		MinVersionSetting: settings.EKSOperatorMinVersion,
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-eks-operator",
 	}
 	GkeCrdChart = chart.Definition{
-		ReleaseNamespace:  "cattle-system",
-		ChartName:         "rancher-gke-operator-crd",
-		MinVersionSetting: settings.GKEOperatorMinVersion,
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-gke-operator-crd",
 	}
 	GkeChart = chart.Definition{
-		ReleaseNamespace:  "cattle-system",
-		ChartName:         "rancher-gke-operator",
-		MinVersionSetting: settings.GKEOperatorMinVersion,
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-gke-operator",
 	}
 )
 
@@ -84,7 +78,7 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster, 
 		return cluster, nil
 	}
 
-	if err := h.manager.Ensure(toInstallCrdChart.ReleaseNamespace, toInstallCrdChart.ChartName, toInstallCrdChart.MinVersionSetting.Get(), nil, true); err != nil {
+	if err := h.manager.Ensure(toInstallCrdChart.ReleaseNamespace, toInstallCrdChart.ChartName, "", nil, true); err != nil {
 		return cluster, err
 	}
 
@@ -107,7 +101,7 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster, 
 		"additionalTrustedCAs": additionalCA != nil,
 	}
 
-	if err := h.manager.Ensure(toInstallChart.ReleaseNamespace, toInstallChart.ChartName, toInstallChart.MinVersionSetting.Get(), chartValues, true); err != nil {
+	if err := h.manager.Ensure(toInstallChart.ReleaseNamespace, toInstallChart.ChartName, "", chartValues, true); err != nil {
 		return cluster, err
 	}
 
--- a/pkg/controllers/dashboard/hostedcluster/controller.go
+++ b/pkg/controllers/dashboard/hostedcluster/controller.go
@@ -1,4 +1,4 @@
-package hostedkubernetecharts
+package hostedcluster
 
 import (
 	"context"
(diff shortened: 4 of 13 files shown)


StrongMonkey:
Assigning forks that worked on kev2 for reviews. In case you found something that needs to be addressed.

thedadams (commented):
Looks good. Will have conflicts with https://github.com/rancher/rancher/pull/32900

Also, can remove the following:
https://github.com/rancher/rancher/blob/master/pkg/controllers/management/clusteroperator/utils.go#L53
https://github.com/rancher/rancher/blob/master/pkg/controllers/management/clusteroperator/utils.go#L59
https://github.com/rancher/rancher/blob/master/pkg/controllers/management/clusteroperator/utils.go#L189

cmurphy:
And also these:

https://github.com/rancher/rancher/blob/113c35317ea3524a8d6e9da6659c772377d33f2b/pkg/controllers/management/aks/aks_cluster_handler.go#L384-L388
https://github.com/rancher/rancher/blob/113c35317ea3524a8d6e9da6659c772377d33f2b/pkg/controllers/management/eks/eks_cluster_handler.go#L484-L488
https://github.com/rancher/rancher/blob/113c35317ea3524a8d6e9da6659c772377d33f2b/pkg/controllers/management/gke/gke_cluster_handler.go#L414-L418

StrongMonkey:
@thedadams I will rebase on your changes if your PR gets merged earlier.

StrongMonkey:
@thedadams @cmurphy Unused functions are removed.

thedadams (approved):

rmweir (commented):
Looking through the install code it looks like this will try to keep the charts up to date with the latest version. What if the rancher version in not compatible with the latest charts version? Will it try to make sure rancher has the latest compatible version even if perhaps it cannot use the latest? If it does do this then LGTM after conflicts have been resolved.

StrongMonkey:
@rmweir There is a minimal version setting for each operator so that we make sure for upgrading setup, if the current operator version meets the min version of the setting, it will not try to upgrade to the latest version. The behavior is defined in here #30934. This is to prevent that we don't intentionally upgrade user's operator if min version is satisfied.

But if there is a fresh install it will always try to get the latest version, this is not a one-operator-version-per-rancher-version model(2.6.0 -> 1.1.0, 2.6.1 -> 1.1.1), as it will always try to install the latest version of the chart in feature charts(so 2.6.x -> latest version of operator). But if there is a newer rancher-operator version that won't work with old rancher version, we can use rancher min/max version on catalog to filter out the latest charts. @cbron Thoughts?

StrongMonkey:
Also there is also a point that for fresh install, user always install the latest version which should get the most compatiable version of operator at the time of the install.

rmweir (changes requested):
@StrongMonkey do have parity with the way KEv2 currently works, it should always install the latest compatible version.

pkg/settings/setting.go
```diff
@@ -113,6 +113,9 @@ var (
 
 	FleetMinVersion          = NewSetting("fleet-min-version", "")
 	RancherWebhookMinVersion = NewSetting("rancher-webhook-min-version", "")
+	AKSOperatorMinVersion    = NewSetting("aks-operator-min-version", "")
+	EKSOperatorMinVersion    = NewSetting("eks-operator-min-version", "")
+	GKEOperatorMinVersion    = NewSetting("gke-operator-min-version", "")
```
cbron, line 118: Also curious about this part, would we need to update every time ? Older version used `LatestAvailableTemplateVersion` to grab newest one. 

edit: 
Oh I missed your comment a few days back: 

> There is a minimal version setting for each operator so that we make sure for upgrading setup, if the current operator version meets the min version of the setting, it will not try to upgrade to the latest version. The behavior is defined in here #30934. This is to prevent that we don't intentionally upgrade user's operator if min version is satisfied.

> But if there is a fresh install it will always try to get the latest version, this is not a one-operator-version-per-rancher-version model(2.6.0 -> 1.1.0, 2.6.1 -> 1.1.1), as it will always try to install the latest version of the chart in feature charts(so 2.6.x -> latest version of operator). But if there is a newer rancher-operator version that won't work with old rancher version, we can use rancher min/max version on catalog to filter out the latest charts.

So to confirm we can use min/max to filter this ? Ideally we would always want the KEv2 operators to have the latest compatible chart for their version line. 
So 2.6.4 gets latest release in release-2.6 branch that is compatible with 2.6.4 (maybe newer one exists but min is 2.6.5). 
If we have to specify that in env vars that is possible, but not the best process. I'm surprised fleet uses this model, I know we don't want the _actual_ latest chart, but latest that uses min/max would be fine no ?

package/Dockerfile
```diff
@@ -43,6 +43,9 @@ ENV CATTLE_SYSTEM_AGENT_UPGRADE_IMAGE rancher/system-agent:${CATTLE_SYSTEM_AGENT
 # System charts minimal version
 ENV CATTLE_FLEET_MIN_VERSION=100.0.0+up0.3.6
 ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=100.0.0+up0.2.0-alpha01
+ENV CATTLE_AKS_OPERATOR_MIN_VERSION=100.0.0+1.0.100
+ENV CATTLE_EKS_OPERATOR_MIN_VERSION=100.0.0+1.1.100
+ENV CATTLE_GKE_OPERATOR_MIN_VERSION=100.0.0+1.1.100
```
cbron, line 48: These haven't been reverted yet in charts repo, but they'll stay at 2.0.0 etc...

pkg/api/steve/catalog/types/rest.go
```diff
@@ -47,6 +47,7 @@ type ChartUpgradeAction struct {
 	DisableHooks             bool             `json:"noHooks,omitempty"`
 	DisableOpenAPIValidation bool             `json:"disableOpenAPIValidation,omitempty"`
 	Force                    bool             `json:"force,omitempty"`
+	ForceAdopt               bool             `json:"forceAdopt,omitempty"`
```
cbron, line 50: This exists just to adopt the system-charts install right ?

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return cluster, err
+	}
+
+	chartValues := map[string]interface{}{
+		"global":               systemGlobalRegistry,
+		"httpProxy":            os.Getenv("HTTP_PROXY"),
+		"httpsProxy":           os.Getenv("HTTPS_PROXY"),
+		"noProxy":              os.Getenv("NO_PROXY"),
+		"additionalTrustedCAs": additionalCA,
```
thedadams, line 109: This should a `bool` value. The operators only care that there are additional CAs. They will get them the same way the Rancher chart does.

pkg/settings/setting.go
StrongMonkey: @cbron Yes, min/max can also be used in feature chart, It will fetch the latest chart in the release branch(release-2.6) that is compatible with the min/max setting. The only difference in this model is that for existing setup, we don't upgrade to the latest compatiable model if the min version is satified. So user won't get an upgrade unless they upgrade rancher, which changes min version settings. Fleet uses this model and I was told to apply it to kev2 charts. But if that's not the case I can remove this min version so it always install the latest compatible version which is the same as how it currently works.

package/Dockerfile
StrongMonkey: Oh I see, this can be modified as long as we make a decision on versioning...

pkg/api/steve/catalog/types/rest.go
StrongMonkey: Yes, for upgrade case, to adopt resources that were created by helm 2.

thedadams (changes requested):

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
StrongMonkey: Oh good catch :+1:

pkg/controllers/dashboard/controller.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/dashboard/clusterindex"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/fleetcharts"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/helm"
+	"github.com/rancher/rancher/pkg/controllers/dashboard/hostedkubernetecharts"
```
nickgerace, line 10: Is this spelling intentional?

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
```diff
@@ -0,0 +1,106 @@
+package hostedkubernetecharts
```
nickgerace, line 1: nit: It seems to be intentional. Perhaps, `hostedkubecharts` or `hostedcharts`? I might have missed the conversations leading to this.

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	GkeChart = chartDef{
+		ReleaseNamespace: "cattle-system",
+		ChartName:        "rancher-gke-operator",
+	}
+)
+
+type chartDef struct {
+	ReleaseNamespace string
+	ChartName        string
+}
```
nickgerace, line 46: I wonder if we should standardize and/or centralize charts references across our controllers.

- https://github.com/rancher/rancher/blob/a56d10297a277dcc04aa48c6dfd45022dda859e3/pkg/controllers/dashboard/systemcharts/controller.go#L17-L50
- https://github.com/rancher/rancher/blob/a56d10297a277dcc04aa48c6dfd45022dda859e3/pkg/controllers/dashboard/fleetcharts/controller.go#L15-L33

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
```diff
@@ -0,0 +1,106 @@
...
+		minVersion = settings.EKSOperatorMinVersion.Get()
+	} else if cluster.Spec.GKEConfig != nil {
+		toInstallCrdChart = &GkeCrdChart
+		toInstallChart = &GkeChart
+		minVersion = settings.GKEOperatorMinVersion.Get()
+	}
+
+	if toInstallCrdChart == nil || toInstallChart == nil {
+		return cluster, nil
+	}
```
nickgerace, line 87: Might we need to continue if the other chart is not nil?

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
StrongMonkey: This was just a random name I gave for the package name, I am up for any good suggestion.

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
StrongMonkey: hmm both should be nil given the if condition above...

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
StrongMonkey: That's a good one, I am going to standardize it.

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
nickgerace: Re-read it. Agreed

pkg/controllers/dashboard/hostedkubernetecharts/controller.go
nickgerace: I would personally prefer `hostedkubecharts` or `hostedclustercharts` if we want to keep it similar.

pkg/controllers/dashboard/controller.go
nickgerace: Resolving due to https://github.com/rancher/rancher/pull/33004#discussion_r652315459

nickgerace (approved):
LGTM, but will likely need reviews from others.

rmweir:
To clarify, I think we want to make sure the operator is as up to date as possible right?

thedadams:
If I have understood the code correctly: if we don't set a min version for the operators in the `Dockerfile`, then the operators would always get updated to the latest version.

StrongMonkey:
@rmweir @thedadams That's correct, if there is no min version defined, the behavior should be alighed with the old system-chart way.

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -0,0 +1,128 @@
...
+
+	chartValues := map[string]interface{}{
+		"global":               systemGlobalRegistry,
+		"httpProxy":            os.Getenv("HTTP_PROXY"),
+		"httpsProxy":           os.Getenv("HTTPS_PROXY"),
+		"noProxy":              os.Getenv("NO_PROXY"),
+		"additionalTrustedCAs": additionalCA != nil,
+	}
+
+	if err := h.manager.Ensure(toInstallChart.ReleaseNamespace, toInstallChart.ChartName, toInstallChart.MinVersionSetting.Get(), chartValues, true); err != nil {
```
rmweir, line 110: Is there a reason why ensure needs to be used twice?

StrongMonkey: One is for CRD chart and one is for operator chart since these two are separated now.

rmweir:
@StrongMonkey can we remove the min version so they are always updated?

package/Dockerfile
rmweir: @StrongMonkey can we remove the min versions so they are always updated?

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -76,23 +76,6 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
-			}
-		} else {
-			cluster, conditionErr = e.SetFalse(cluster, apimgmtv3.ClusterConditionProvisioned, fmt.Sprintf(failedToDeployAKSOperatorErr, err))
-			if conditionErr != nil {
-				return cluster, conditionErr
-			}
-		}
-		return cluster, err
-	}
-
```
rmweir, line 95: Nit: Should we handle the scenario where the CRD isn't installed yet more gracefully or wait to proceed until it is installed? Otherwise, errors are likely to occur on L87. We could also return an error saying "waiting for CRD to be created.

StrongMonkey: Since the chart operator is based on `v3.cluster`. Once chart is installed it will change v3.cluster which trigger this controller, so it should eventually go through? But yeah I can check if the crd exists and give a more graceful error.

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -260,6 +247,22 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
 }
 
+// check if any resources in aks.cattle.io/v1
+func (e *aksOperatorController) checkCrdReady(cluster *mgmtv3.Cluster) (err error) {
+	resources, err := e.Discovery.ServerResourcesForGroupVersion("aks.cattle.io/v1")
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	if errors.IsNotFound(err) || len(resources.APIResources) == 0 {
+		cluster, err = e.SetUnknown(cluster, apimgmtv3.ClusterConditionProvisioned, "Waiting on aks crd to be initialized")
```
rmweir, line 257: I think just return the error is sufficient and setting the message may not be necessary. It's nice but depending on if it is registered or not the expected condition is different here I believe. Also, if we were to update the entire object we probably want to return it from this function and reassign so that the cluster object in the parent function is up to date. This comment applies to all 3 occurrences of this function definition.

rmweir (changes requested):

pkg/controllers/management/aks/aks_cluster_handler.go
StrongMonkey: Yeah, setting the condition in here so that if crd chart install failed, cluster could have a better way to display the information in UI showing the cluster is waiting on crd. For the cluster object it is passed as a pointer and reassigned in the method, it should get what's updated.

pkg/settings/setting.go
cbron: Seems like you've done that above, do you still need these settings ? 
If its because if you set these you get the "fleet model" there should be comment saying thats how it works, as its somewhat counterintuitive.

pkg/controllers/management/aks/aks_cluster_handler.go
thedadams: Reassigning a pointer in a function doesn't bubble up to the caller: https://play.golang.org/p/rdD6A6alGIE

StrongMonkey: Oh......

StrongMonkey: @rmweir @thedadams This is fixed.

rmweir (approved):
LGTM after function comment fixes.

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -260,6 +248,22 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 	}
 }
 
+// check if any resources in aks.cattle.io/v1
```
rmweir, line 251: Can you fix this comment?
```suggestion
// checkCRDReady checks whether necessary CRD, AKSConfig, has been created yet
```

rmweir (changes requested):

pkg/controllers/management/aks/aks_cluster_handler.go
StrongMonkey: :+1: fixed

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -260,6 +248,22 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 	}
 }
 
+// checkCRDReady checks whether necessary CRD, AKSConfig, has been created yet
+func (e *aksOperatorController) checkCrdReady(cluster *mgmtv3.Cluster) (*mgmtv3.Cluster, error) {
```
thedadams, line 252: This method can be moved to `pkg/controllers/management/clusteroperator/utils.go` and changed to take a string parameter as well: `checkCrdReady(cluster *mgmtv3.Cluster, provider string)`.

Then each handler could pass a string, like: `checkCrdReady(cluster, "aks")`.

We try to keep the common code in `pkg/controllers/management/clusteroperator/utils.go`.

thedadams (changes requested):

pkg/controllers/management/aks/aks_cluster_handler.go
StrongMonkey: Sounds good, let me change that.

StrongMonkey: :+1: Changed to common function.

thedadams (dismissed):

rmweir (dismissed):
LGTM

pkg/settings/setting.go
cbron: @thedadams @rmweir if we remove these we will have to set min/max on every feature chart version, otherwise every newer release that goes into the release branches will get pulled into old ranchers, which could crash things. Whats your preferred route, updating these all the time or updating min/max in the chart ? 
The chart is better way IMO, but with feature charts its hard to adjust old ones, you have to replace tarball. So we'll either have to go route of: 
a) set a max on every release and continuously update that to bump it
b) not set a max and then add one when there is a new release.
Either ways are error prone. Which then leads me back to using this min version, as it would be obvious if we forgot it.

cbron: @StrongMonkey I think we should remove these and then we will just put a min on every feature chart version that matches rancher version.

StrongMonkey: @cbron Settings are also removed.

cbron (approved):

PR #33007, opened 2021-06-03
Patchset to support SLE BCI

mvarlese:
Support for SUSE BCI

Removed Leap 15.3 repo

All packages are now available via the BCI repository hence the Leap
15.3 repo added temporarily to support migration can now be removed.

Explicitly requesting gawk

awk is installed via the package gawk which is available via
patterns-base-minimal_base. When moving away from Leap 15.3 repo
that got lost hence we have to explicitly call its installation.

Explicitely call glibc-devel installation

Added glibc-devel-static to buld rancherd

Moved to SLE BCI image and repository

Use zypper internals

Use zypper internal product and architecture handling instead of hardcoding
them in repository URL. This will facilitate easy handling of newly
available versions.

Do not run 'zypper update'

In first attempt of translating from Ubuntu to SUSE there was a mistake made
where 'apt-get update' was translated to 'zypper update'. However, that
should have translated to 'zypper refresh'.
This patch fixes that.

add parameter to support SLE_BCI

Update package/Dockerfile.agent

improve zypper call

Co-authored-by: Dirk Mueller <dmueller@suse.com>

Update package/Dockerfile

improve zypper call

Co-authored-by: Dirk Mueller <dmueller@suse.com>

Fixup install command

Set default to SLE BCI image

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,5 +1,4 @@
-FROM ubuntu:20.04
-# FROM arm=armhf/ubuntu:20.04 arm64=arm64v8/ubuntu:20.04
+FROM registry.suse.com/suse/sle15:15.3
 
 ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
@@ -16,9 +15,8 @@ ENV KUSTOMIZE_VERSION v3.10.0
 # kontainer-driver-metadata branch to be set for specific branch other than dev/master, logic at rancher/rancher/pkg/settings/setting.go
 ENV CATTLE_KDM_BRANCH=dev-v2.6
 
-RUN apt-get update && \
-    apt-get install -y gcc ca-certificates git wget curl vim less file xz-utils unzip iproute2 && \
-    rm -f /bin/sh && ln -s /bin/bash /bin/sh
+RUN zypper ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI && \
+    zypper -n install gcc binutils glibc-devel-static ca-certificates git-core wget curl tar vim less file xz gzip sed gawk iproute2 iptables
 
 RUN curl -sLf https://github.com/rancher/machine/releases/download/${CATTLE_MACHINE_VERSION}/rancher-machine-${ARCH}.tar.gz | tar xvzf - -C /usr/bin
 
@@ -28,7 +26,11 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH_arm64=arm64 GOLAN
 RUN curl -sLf https://storage.googleapis.com/golang/go1.16.4.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local
 
 RUN if [ "${ARCH}" == "amd64" ]; then \
-    curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.40.1; \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.40.1; \
+    fi
+# workaround for https://bugzilla.suse.com/show_bug.cgi?id=1183043
+RUN if [ "${ARCH}" == "arm64" ]; then \
+        zypper -n install binutils-gold ; \
     fi
 
 ENV DOCKER_URL_amd64=https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 \
@@ -82,8 +84,7 @@ RUN curl -sLf ${YQ_URL} -o /usr/bin/yq && chmod +x /usr/bin/yq
 ENV KUBECTL_URL=https://storage.googleapis.com/kubernetes-release/release/v1.21.0/bin/linux/${ARCH}/kubectl
 RUN curl -sLf ${KUBECTL_URL} -o /usr/bin/kubectl && chmod +x /usr/bin/kubectl
 
-RUN apt-get update && \
-    apt-get install -y tox python3.8 python3-dev python3.8-dev libffi-dev libssl-dev
+RUN zypper install -y python3-tox python3-base python3 libffi-devel libopenssl-devel
 
 ENV HELM_HOME /root/.helm
 ENV DAPPER_ENV REPO TAG DRONE_TAG SYSTEM_CHART_DEFAULT_BRANCH
--- a/package/Dockerfile.agent
+++ b/package/Dockerfile.agent
@@ -2,17 +2,18 @@ ARG RANCHER_TAG=dev
 ARG RANCHER_REPO=rancher
 FROM ${RANCHER_REPO}/rancher:${RANCHER_TAG} as rancher
 
-FROM ubuntu:18.04
+FROM registry.suse.com/suse/sle15:15.3
 ARG ARCH=amd64
 
-RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 ENV KUBECTL_VERSION v1.20.7
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion sysstat acl ssh iptables && \
+# sysstat is left out until systemd dependency is stripped
+RUN zypper -n ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI && \
+    zypper -n install --no-recommends curl ca-certificates jq hostname iproute2 vim-small less bash-completion acl openssh-clients tar gzip xz gawk && \
+    zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* && \
+    rpm -e libsolv-tools zypper libzypp container-suseconnect ; \
     curl -sLf https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl > /usr/bin/kubectl && \
-    chmod +x /usr/bin/kubectl && \
-    DEBIAN_FRONTEND=noninteractive apt-get autoremove -y && \
-    DEBIAN_FRONTEND=noninteractive apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    chmod +x /usr/bin/kubectl
+
 ENV LOGLEVEL_VERSION v0.1.3
 
 RUN curl -sLf https://github.com/rancher/loglevel/releases/download/${LOGLEVEL_VERSION}/loglevel-${ARCH}-${LOGLEVEL_VERSION}.tar.gz | tar xvzf - -C /usr/bin
--- a/package/jailer.sh
+++ b/package/jailer.sh
@@ -22,11 +22,12 @@ mkdir -p /opt/jail/$NAME/bin
 # Copy over required files to the jail
 if [[ -d /lib64 ]]; then
   cp -r /lib64 /opt/jail/$NAME
+  cp -r /usr/lib64 /opt/jail/$NAME/usr
 fi
 
 cp -r /lib /opt/jail/$NAME
 cp -r /usr/lib /opt/jail/$NAME/usr
-cp /etc/ssl/certs/ca-certificates.crt /opt/jail/$NAME/etc/ssl/certs
+cp /var/lib/ca-certificates/ca-bundle.pem /opt/jail/$NAME/etc/ssl/certs
 cp /etc/resolv.conf /opt/jail/$NAME/etc/
 cp /etc/passwd /opt/jail/$NAME/etc/
 cp /etc/hosts /opt/jail/$NAME/etc/
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,14 +1,18 @@
-FROM ubuntu:18.04
-RUN apt-get update && apt-get install -y git curl ca-certificates unzip xz-utils && \
+FROM registry.suse.com/suse/sle15:15.3
+
+RUN zypper -n ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI && \
+    zypper -n install git-core curl ca-certificates unzip xz gzip sed tar shadow gawk && \
+    zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* && \
+    rpm -e libsolv-tools zypper libzypp container-suseconnect ; \
     useradd rancher && \
     mkdir -p /var/lib/rancher/etcd /var/lib/cattle /opt/jail /opt/drivers/management-state/bin && \
     chown -R rancher /var/lib/rancher /var/lib/cattle /usr/local/bin
+
 RUN mkdir /root/.kube && \
     ln -s /etc/rancher/k3s/k3s.yaml /root/.kube/k3s.yaml  && \
     ln -s /etc/rancher/k3s/k3s.yaml /root/.kube/config && \
     ln -s /usr/bin/rancher /usr/bin/reset-password && \
-    ln -s /usr/bin/rancher /usr/bin/ensure-default-admin && \
-    rm -f /bin/sh && ln -s /bin/bash /bin/sh
+    ln -s /usr/bin/rancher /usr/bin/ensure-default-admin
 WORKDIR /var/lib/rancher
 
 ARG ARCH=amd64


package/Dockerfile.agent
```diff
@@ -1,18 +1,25 @@
...
 ARG ARCH=amd64
 
 RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 ENV KUBECTL_VERSION v1.20.7
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion unzip sysstat acl ssh iptables && \
+RUN if [ -x /usr/bin/zypper ]; then \
+     zypper ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI && \
+     zypper -n install --no-recommends git-core curl ca-certificates jq iproute2 vim less bash-completion unzip sysstat acl openssh-clients && \
+     SUSE_FRONTEND=zypper clean && rm -rf /tmp/* /var/tmp/* ; \
```
dirkmueller, line 14:
```suggestion
     zypper clean -a && rm -rf /tmp/* /var/tmp/* ; \
```

dirkmueller: the SUSE_FRONTEND is not needed

package/Dockerfile.agent
```diff
@@ -1,18 +1,25 @@
 ARG RANCHER_TAG=dev
 ARG RANCHER_REPO=rancher
+ARG BASE_DISTRIBUTION=registry.suse.com/suse/sle15:15.3
 FROM ${RANCHER_REPO}/rancher:${RANCHER_TAG} as rancher
 
-FROM ubuntu:18.04
```
dirkmueller, line 5: note to self: we need to understand why this remained on 18.04 while others were 20.04

package/Dockerfile.agent
```diff
@@ -1,18 +1,25 @@
...
+     zypper -n install --no-recommends git-core curl ca-certificates jq iproute2 vim less bash-completion unzip sysstat acl openssh-clients && \
+     zypper -n clean -a && rm -rf /tmp/* /var/tmp/* ; \
+    else \
+     apt-get update && \
+     apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion unzip sysstat acl ssh && \
+     DEBIAN_FRONTEND=noninteractive apt-get autoremove -y && \
+     DEBIAN_FRONTEND=noninteractive apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* ; \
+    fi && \
     curl -sLf https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl > /usr/bin/kubectl && \
     chmod +x /usr/bin/kubectl && \
```
dirkmueller, line 22:
```suggestion
    chmod +x /usr/bin/kubectl
```

Dockerfile.dapper
```diff
@@ -85,8 +91,12 @@ RUN wget -O - ${YQ_URL} > /usr/bin/yq && chmod +x /usr/bin/yq
 ENV KUBECTL_URL=https://storage.googleapis.com/kubernetes-release/release/v1.21.0/bin/linux/${ARCH}/kubectl
 RUN wget -O - ${KUBECTL_URL} > /usr/bin/kubectl && chmod +x /usr/bin/kubectl
 
-RUN apt-get update && \
-    apt-get install -y tox python3.8 python3-dev python3.8-dev libffi-dev libssl-dev
+RUN if [ -x /usr/bin/zypper ]; then \
+     zypper install -y tox python39 python39-devel libffi-devel libopenssl-devel ; \
```
fcrozat:
```suggestion
     zypper install -y python3-tox libffi-devel libopenssl-devel ; \
```
no need for python3.9 and use proper name for tox package

dirkmueller: thanks, manually merged

fcrozat (approved):

fcrozat (approved):

ibuildthecloud:
What is the difference in the size of the images produced? I'd imagine the size difference is negligible.

mvarlese:
> What is the difference in the size of the images produced? I'd imagine the size difference is negligible.

I have no clue. Can't that be seen from the CI where you build that?

dirkmueller:
> What is the difference in the size of the images produced? I'd imagine the size difference is negligible.

There is still quite a bit of difference: 

ubuntu based:

```
REPOSITORY                     TAG       IMAGE ID       CREATED         SIZE
rancher                        HEAD      64bc232385ab   5 seconds ago   1.28GB
```

with this PR (SLE based):

```
dirk@magnolia:~> docker images
REPOSITORY                     TAG       IMAGE ID       CREATED          SIZE
rancher                        HEAD      188fbd203e0c   4 minutes ago    1.62GB

```

ibuildthecloud:
The images we really care about are rancher/rancher and rancher/agent.  The "rancher" image you referenced is the one created by Dockerfile.dapper which is the build environment and we don't really about the size.  I tried running master vs this PR and I got:

bci:

```
rancher/rancher-agent            master-head            cfec73cbad84   About a minute ago   556MB
rancher/rancher                  dev                    f30476eb483d   2 minutes ago        1.08GB
```

ubuntu:

```
rancher/rancher-agent            master-head             ff28c7ddb24b   2 minutes ago    463MB
rancher/rancher                  dev                     b3343f286c35   3 minutes ago    1.07GB
```

So the BCI images are about 100mb bigger.

dirkmueller:
> So the BCI images are about 100mb bigger.

we're aware of pending issues that hit these two images: 

* git-core pulls in perl-full
* there is no vim-tiny/vim-small in SLES
* sysstat pulls in systemd / udev / kbd
* zypper is pretty hefty (and there is no zypper-minimal in SLES)
* container-suseconnect is build with full debug

when I "avoid" those for Dockerfile.agent with this change:

```diff

--- a/package/Dockerfile.agent
+++ b/package/Dockerfile.agent
@@ -10,8 +9,9 @@ RUN rm -f /bin/sh && ln -s /bin/bash /bin/sh
 ENV KUBECTL_VERSION v1.20.7
 RUN if [ -x /usr/bin/zypper ]; then \
         zypper -n ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI && \
-        zypper -n install --no-recommends git-core curl ca-certificates jq iproute2 vim less bash-completion unzip sysstat acl openssh-clients tar gzip xz && \
-        zypper -n clean -a && rm -rf /tmp/* /var/tmp/* ; \
+        zypper -n install --no-recommends curl ca-certificates jq iproute2 less bash-completion unzip acl openssh-clients tar gzip xz && \
+        zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages && \
+        rpm -e libsolv-tools zypper libzypp container-suseconnect libyaml-cpp0_6 libglib-2_0-0 boost-license1_66_0 libboost_system1_66_0 libboost_thread1_66_0 ; \
     else \
         apt-get update && \
         apt-get install -y --no-install-recommends git curl ca-certificates jq iproute2 vim-tiny less bash-completion unzip sysstat acl ssh && \
```

by exactly 128MB. for most of the above listed changes, we have pending maintenance updates that will fix those, so image size will shrink over time. 

Also I can take a look at slimming things down further if wanted.

mvarlese:
@ibuildthecloud CI for arm64 is failing by what it looks like being a timeout? Can anybody with right privileges kick the build off again so we ensure all is GREEN again, please?

Dockerfile.dapper
```diff
@@ -28,7 +34,11 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH_arm64=arm64 GOLAN
 RUN wget -O - https://storage.googleapis.com/golang/go1.16.4.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local
 
 RUN if [ "${ARCH}" == "amd64" ]; then \
-    curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.27.0; \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.27.0; \
+    fi
+# fugly workaround for https://bugzilla.suse.com/show_bug.cgi?id=1183043
+RUN if [ -x /usr/bin/zypper -a "${ARCH}" == "arm64" ]; then \
+        rpm -Uvh http://download.opensuse.org/distribution/leap/15.3/repo/oss/aarch64/binutils-gold-2.35.1-7.18.1.aarch64.rpm ; \
```
fcrozat, line 41: SLE_BCI now ships binutils-gold for aarch64, let's use instead
```suggestion
        zypper -n install -y binutils-gold ; \
```

dirkmueller:
Here's the update. the agent image is now with this patch set 30 MB smaller than with ubuntu before, which compensates for the 10MB increase in rancherd. The 10 MB increase in rancherd will be reduced in a followon patchset.

mvarlese:
@ibuildthecloud would it be possible to kick off the build again? It fails (randomly) due to timeouts. Thanks.

mvarlese:
@ibuildthecloud would it be possible to re-spin the build for anybody with the right privileges? It looks like something is failing due to timeout...

package/run.sh
```diff
@@ -156,7 +156,7 @@ if [ "$CATTLE_CLUSTER" != "true" ]; then
 fi
 
 if [ -z "$CATTLE_NODE_NAME" ]; then
-    CATTLE_NODE_NAME=$(hostname -s)
+    CATTLE_NODE_NAME=$HOSTNAME
```
ibuildthecloud, line 159: Is this guaranteed to be the "short" hostname, only the portion before the first dot.

mvarlese: Here the output:

mvarlese@marco-pc-lavoro:~> hostname -s
marco-pc-lavoro

mvarlese@marco-pc-lavoro:~> hostname
marco-pc-lavoro.fritz.box

dirkmueller: the man page says `HOSTNAME  Automatically set to the name of the current host`, so it is ambiguous. I tested it within a k8s cluster and outside and it seems to behave equivalent. I'll try to get hostname(1) embedded in the SLES base image as well, so we can change this. here's my testing:

```
$ hostname -f
postgresql-0.hub-1621432880-postgresql-headless.default.svc.cluster.local
$ hostname -s
postgresql-0
$ echo $HOSTNAME
postgresql-0
$ uname -n
postgresql-0

another one:

$ hostname -f
wotan.suse.de
$ hostname -s
wotan
$ echo $HOSTNAME
wotan
$ uname -n
wotan
```

dirkmueller: I see this affects documented behavior: 

https://github.com/rancher/docs/blame/master/content/rancher/v2.x/en/cluster-provisioning/rke-clusters/custom-nodes/agent-options/_index.md#L20

so I'll have to work harder on getting hostname(1) into the image it seems..

mvarlese: The "hostname" executable is now available via the BCI repository (not installed in the base image by default yet). 
You could revert the latest patch and add the package hostname where needed.

dirkmueller: done.

Dockerfile.dapper
```diff
@@ -16,9 +17,14 @@ ENV KUSTOMIZE_VERSION v3.10.0
 # kontainer-driver-metadata branch to be set for specific branch other than dev/master, logic at rancher/rancher/pkg/settings/setting.go
 ENV CATTLE_KDM_BRANCH=dev-v2.6
 
-RUN apt-get update && \
-    apt-get install -y gcc ca-certificates git wget curl vim less file xz-utils unzip iproute2 && \
-    rm -f /bin/sh && ln -s /bin/bash /bin/sh
+RUN if [ -x /usr/bin/zypper ]; then \
+        zypper ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI && \
```
pdostal, line 21: Variables `$releasever_major`, `$releasever_minor` and `$basearch` probably should be defined.

fcrozat: No, those variable are internal to zypper (see man zypper) and resolved when zypper is run. This allows the repository url to be version and arch agnostic.

pdostal: @fcrozat thank you, this is useful! This thread can be marked as resolved.

fcrozat: Last comment in this particular topic: the zypper ar call will be removed once SLE Base Image is updated to have SLE_BCI pre-configured.

pdostal:
Hello,

> Support for SUSE BCI
> 
> Removed Leap 15.3 repo

may I ask you what is the plan with openSUSE? Will this new container be usable on openSUSE host?

fcrozat:
> Hello,
> 
> > Support for SUSE BCI
> > Removed Leap 15.3 repo
> 
> may I ask you what is the plan with openSUSE? Will this new container be usable on openSUSE host?

This container can be used on ANY host, including openSUSE. This is not a new container but the regular SLE Base Image (the only upcoming change is preconfiguration of SLE BCI repository in it, which doesn't require registration). When host is used on a registered SLES system, SLE BCI repository is not used and instead, SLES repositories from SUSE Customer Center are used.

ibuildthecloud:
Can we just remove the apt-get stuff and assume zypper? We are going to default to sle and nobody will ever manually switch it back to ubuntu, so basically we will just end up accidentally breaking the ubuntu support.

deniseschannon:
Can we remove the ability to change the base image as an argument so that we only have the option to use the new base image?

mvarlese:
> Can we just remove the apt-get stuff and assume zypper? We are going to default to sle and nobody will ever manually switch it back to ubuntu, so basically we will just end up accidentally breaking the ubuntu support.

Of course; we did not know your intentions so this patchset was the most flexible approach.

mvarlese:
> Can we remove the ability to change the base image as an argument so that we only have the option to use the new base image?

Sure; as per comment above: we didn't know this was a welcomed approach hence wrote the patch to be offer a flexible configuration via command line parameter.

fcrozat (dismissed):

dirkmueller:
did a very minimal rebase to resolve conflicts, would like to squash some of the commits to gether as this is quite messy now. will take  a bit of time

deniseschannon:
@dirkmueller When building this image locally and then trying this out as a docker install, it doesn't successfully come up. 

Have you guys tested this out?

snasovich:
@dirkmueller , Rancher fails to start due to error running `jailer.sh` script which (among other things) does the following:
```cp /etc/ssl/certs/ca-certificates.crt /opt/jail/$NAME/etc/ssl/certs```

The image produced by updated dockerfile doesn't have `/etc/ssl/certs/ca-certificates.crt`. Here is a quick test:
```
➜  rancher git:(mvarlese-master) ✗ docker run --rm -it --entrypoint bash registry.suse.com/suse/sle15:15.3           
ef15b43e539c:/ # ls
bin  boot  dev  etc  home  lib  lib64  mnt  opt  proc  root  run  sbin  selinux  srv  sys  tmp  usr  var
ef15b43e539c:/ # zypper ar --refresh --priority 100 'https://updates.suse.com/SUSE/Products/SLE-BCI/$releasever_major-SP$releasever_minor/$basearch/product/' SLE_BCI
Adding repository 'SLE_BCI' ......................................................................................................................................................................................[done]
Repository 'SLE_BCI' successfully added

URI         : https://updates.suse.com/SUSE/Products/SLE-BCI/15-SP3/x86_64/product/
Enabled     : Yes
GPG Check   : Yes
Autorefresh : Yes
Priority    : 100 (lowered priority)

Repository priorities are without effect. All enabled repositories share the same priority.
ef15b43e539c:/ # zypper -n install ca-certificates
Refreshing service 'container-suseconnect-zypp'.
Problem retrieving the repository index file for service 'container-suseconnect-zypp':
[container-suseconnect-zypp|file:/usr/lib/zypp/plugins/services/container-suseconnect-zypp] 
Warning: Skipping service 'container-suseconnect-zypp' because of the above error.
Retrieving repository 'SLE_BCI' metadata .........................................................................................................................................................................[done]
Building repository 'SLE_BCI' cache ..............................................................................................................................................................................[done]
Loading repository data...
Reading installed packages...
'ca-certificates' is already installed.
No update candidate for 'ca-certificates-2+git20210309.21162a6-2.1.noarch'. The highest available version is already installed.
Resolving package dependencies...
Nothing to do.
ef15b43e539c:/ # cd /etc/ssl
ef15b43e539c:/etc/ssl # ls
ca-bundle.pem  certs  ct_log_list.cnf  ct_log_list.cnf.dist  openssl.cnf  private
ef15b43e539c:/etc/ssl # 
```

Previous `ubuntu:20.04`-based image got `/etc/ssl/certs/ca-certificates.crt` file in after `apt-get install ca-certificates`. Is it possible to update Dockerfile to have the same on this `registry.suse.com/suse/sle15:15.3`-based image?

**Comment by snasovich** [comment]
Another option may be to update `jailer.sh` to copy all files from `/etc/ssl/certs` to `/opt/jail/$NAME/etc/ssl/certs`. I've tried this locally and `make ci` produces `rancher/rancher` image that at least starts successfully run as
```docker run -d --restart=unless-stopped \
  -p 80:80 -p 443:443 \
  --privileged \
  rancher/rancher:<commithash>```

**Comment by snasovich** [comment]
As expected the fix is incomplete as per https://github.com/rancher/rancher/issues/33127#issuecomment-859144930

**Comment by dirkmueller** [comment]
> Have you guys tested this out?

Yes, I did.. maybe my last rebase activity was suboptimal. let me figure it out.

**Comment by deniseschannon** [comment]
@dirkmueller How have you guys been testing this? By creating a local image @snasovich was able to create an image and tried to bring up Rancher using the docker install method. 

The Rancher server never starts, which would have completely blocked anyone from using the `master-head` builds and caused a complete blocker to our QA team. 

Please confirm that you've tested this locally yourself and you are able to hit the UI after you do the Docker install. 

This is the Docker command to start up Rancher, which you can do on your local machine. Afterwards, you should be able to navigate to https://localhost/ and see the Rancher UI. 

```
docker run -d --restart=unless-stopped  -p 80:80 -p 443:443 --privileged  rancher/rancher:latest
```

Can you confirm that your latest updates have been locally tested before we try and send it to QA to a round of automation tests on an image?

**Comment by snasovich** [comment]
@deniseschannon , we were able to get past the issue of container not starting yesterday. And QA did some quick testing and reported the following issue: https://github.com/rancher/rancher/issues/33127#issuecomment-859144930
My understanding is that @dirkmueller fixed it by changing `jailer.sh` to copy additional libs. Is it correct, @dirkmueller ?

@dirkmueller , how do you test currently? For example, do you include creation of new downstream clusters?

**Comment by deniseschannon** [comment]
@dirkmueller Because of the time zone differences, would it make more sense to set up a call to discuss how we would minimally want this tested on your end before Rancher picks it up for additional testing?

**Comment by dirkmueller** [comment]
> @deniseschannon , we were able to get past the issue of container not starting yesterday. And QA did some quick testing and reported the following issue: [#33127 (comment)](https://github.com/rancher/rancher/issues/33127#issuecomment-859144930)
> My understanding is that @dirkmueller fixed it by changing `jailer.sh` to copy additional libs. Is it correct, @dirkmueller ?

yes, as reported here https://github.com/rancher/rancher/issues/33127#issuecomment-859346781

> @dirkmueller , how do you test currently? For example, do you include creation of new downstream clusters?

I was hoping somebody could help us understand how to run the validation suite. I am testing rancher manually but this is more like an untrained monkey clicking. I did find a few small issues that were easy to resolve. 

the easiest way would be to give us help in running an elaborate testsuite or set aside somebody that can report issues.

**Comment by snasovich** [comment]
@dirkmueller , I'm helping with the initial smoke testing and QA will run their validation testing after.

Following the fix applied by you I was able to deploy a cluster based on an image built from this PR.
However, provisioning a new cluster in Digital Ocean was getting stuck with `Cluster must have at least one etcd plane host: failed to connect to the following etcd host(s) [167.71.94.72]` message in the UI.
Long story short, it came down to `rancher-agent` container failing due to missing `awk` command used by: https://github.com/rancher/rancher/blob/490011ea432d4dde4533c246cb5302fe5724d1ec/package/run.sh#L243
Looks like `awk` is not a part of the base sle15:15.3 image (it's built into ubuntu:20.04 by default). I've added `awk` to `zypper install` command in my local `Dockerfile.agent` and was able to successfully provision a cluster after that.
@dirkmueller , I'm not sure if there are any reasons we shouldn't have awk there. If there are no issues, please apply the change to this PR.

On a side note, when you navigate to https://registry.suse.com/ it shows that 15 SP3 is _intended for testing only. Their use in production systems is not recommended nor supported._
Do we want to downgrade to `sle15:15.2` or you know that 15.3 will be deemed production ready really soon?

**Comment by dirkmueller** [comment]
@snasovich  we're good on using 15.3. it is done, and it will be shipped June 22nd, 2021.

**Comment by dirkmueller** [comment]
> Looks like `awk` is not a part of the base sle15:15.3 image (it's built into ubuntu:20.04 by default). I've added `awk` to `zypper install` command in my local `Dockerfile.agent` and was able to successfully provision a cluster after that.

Great finding, let's do that then. I have added a commit to add awk. Thank you for troubleshooting this!

BTW, removing use of awk is just as easy. I propose an alternative PR for this.

**Comment by snasovich** [comment]
@dirkmueller , thank you for all the suggestions. For this one we can keep `gawk` and then as part of #33159 we can remove it. Also, I'm not sure we need it in `Dockerfile.dapper` - it's used by for build environment isolation and wasn't failing without gawk for me.
And thank you for the clarifications on SP3 - I figured the release may be just around the corner, but wanted to double-check.

**Comment by mvarlese** [comment]
> @dirkmueller , I'm helping with the initial smoke testing and QA will run their validation testing after.
> 
> Following the fix applied by you I was able to deploy a cluster based on an image built from this PR.
> However, provisioning a new cluster in Digital Ocean was getting stuck with `Cluster must have at least one etcd plane host: failed to connect to the following etcd host(s) [167.71.94.72]` message in the UI.
> Long story short, it came down to `rancher-agent` container failing due to missing `awk` command used by:
> 
> https://github.com/rancher/rancher/blob/490011ea432d4dde4533c246cb5302fe5724d1ec/package/run.sh#L243
> 
> 
> Looks like `awk` is not a part of the base sle15:15.3 image (it's built into ubuntu:20.04 by default). I've added `awk` to `zypper install` command in my local `Dockerfile.agent` and was able to successfully provision a cluster after that.

Thanks for your testing, investigation and findings!

> @dirkmueller , I'm not sure if there are any reasons we shouldn't have awk there. If there are no issues, please apply the change to this PR.
> On a side note, when you navigate to https://registry.suse.com/ it shows that 15 SP3 is _intended for testing only. Their use in production systems is not recommended nor supported._
> Do we want to downgrade to `sle15:15.2` or you know that 15.3 will be deemed production ready really soon?

-SP3 is upcoming and we should not be based on anything else for this.

**Comment by dirkmueller** [comment]
> Also, I'm not sure we need it in `Dockerfile.dapper` - it's used by for build environment isolation and wasn't failing without gawk for me.

so in `Dockerfile.dapper` it is needed (scripts/build-* are using awk in various places). Thinking about it, you were maybe referring to the other package/Dockerfiles and I agree, it looks like some of the scripts might require awk. I've added it there as well, in the assumption be be better safe than sorry. we can always clean this up when the dust has settled.

File: Dockerfile.dapper [hunk 8]
```diff
@@ -1,5 +1,5 @@
-FROM ubuntu:20.04
 # FROM arm=armhf/ubuntu:20.04 arm64=arm64v8/ubuntu:20.04
```
snasovich, line 1: Might as well remove this commented-out line as it doesn't seem we're going back to ubuntu following this change.

Dockerfile.dapper
```diff
@@ -28,7 +27,11 @@ ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm=armv6l GOLANG_ARCH_arm64=arm64 GOLAN
 RUN curl -sLf https://storage.googleapis.com/golang/go1.16.4.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local
 
 RUN if [ "${ARCH}" == "amd64" ]; then \
-    curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.40.1; \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.40.1; \
+    fi
+# fugly workaround for https://bugzilla.suse.com/show_bug.cgi?id=1183043
```
snasovich, line 32: Recommend removing `fugly` word, just `workaround for ` should be enough. :)

package/jailer.sh
```diff
@@ -54,6 +55,7 @@ fi
 cp -r -l /opt/drivers/management-state/bin /opt/jail/$NAME/var/lib/rancher/management-state
 
 # Hard link rancher-machine into the jail
+
```
snasovich, line 58: Major nitpick, but we don't need this new empty line.

snasovich:
@dirkmueller , my bad - I've missed that installation of `gawk` was already part of `Dockerfile.dapper` before my last round of testing.

Meanwhile, we have received positive validation from QA: https://github.com/rancher/rancher/issues/33127#issuecomment-860921695. So, please get this PR ready for merging. My understanding is the current process for working with PRs in this repo is to force-push a single commit containing all the changes (I've also added some minor comments to address). Before then might as well get the source branch up to date with the destination as it's a bit behind now.

I see one item in the build is failing - there are some flaky tests, so let's see how it goes when you push the next one.

Dockerfile.dapper
mvarlese: Fixed in latest patchset.

Dockerfile.dapper
mvarlese: Fixed in latest patchset.

package/jailer.sh
mvarlese: Fixed in latest patchset.

mvarlese:
> @dirkmueller , my bad - I've missed that installation of `gawk` was already part of `Dockerfile.dapper` before my last round of testing.
> 
> Meanwhile, we have received positive validation from QA: [#33127 (comment)](https://github.com/rancher/rancher/issues/33127#issuecomment-860921695). So, please get this PR ready for merging. My understanding is the current process for working with PRs in this repo is to force-push a single commit containing all the changes (I've also added some minor comments to address). Before then might as well get the source branch up to date with the destination as it's a bit behind now.

I think I have addressed your comments with my latest patchset. Please, have a second look and let us know any further thoughts.

> 
> I see one item in the build is failing - there are some flaky tests, so let's see how it goes when you push the next one.

I am not sure this new patchset would fix that as it's mostly nitpicks...

snasovich:
@mvarlese , thank you for addressing my comments. The build actually succeeded - https://drone-pr.rancher.io/rancher/rancher/16419. That's what I meant by "flaky tests" - they sometimes fail randomly and pushing even trivial changes causes build to rerun and it may succeed then. It's also possible to just rerun the build if it fails.

Now, my understanding of current Rancher PR process is that we want a single commit on the PR before it's merged. So, could you please squash all commits into one and force-push to this branch?

@cbron , please correct me if I'm wrong here.

snasovich (dismissed):

ibuildthecloud (approved):

mvarlese:
@ibuildthecloud I am not sure if the failing CI will stop the merge from happening but it looks like a failure not connected to this patch to me. CI was green just before the squash so not expecting it to be the breakage here. Worth to kick off the CI again (I can't do that tho)?

snasovich:
@mvarlese , I've kicked off CI ~20 minutes ago - but now it already failed on Windows build (which was succeeding previously). I'm going to kick if off again.
On a side note, make sure you're logged in to drone - "Log In" button in top right corner. I've had the same issue of not being able to restart builds until noticing I'm not logged in.

snasovich:
@mvarlese @dirkmueller , there were some issues with Windows builds that prevented CI builds from succeeding. These were addressed per https://github.com/rancher/rancher/pull/33199. Could you please merge these changes in and make the same dance with squasing the changes and force-pushing?
Once it's done and a build succeeds, feel free to merge as you have approvals.

mvarlese:
@snasovich CI is green now; I guess you guys need to re-approve before this PR can be merged anyway.

snasovich:
@mvarlese and @dirkmueller , thank you gents. I'm just coordinating with QA before we merge it.

snasovich (approved):

Issue #30628, opened 2020-12-24
Nodes fail to update, with cluster stuck in Upgrading state when attempting to update private registry details in Rancher provisioned cluster


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Provision a Rancher instance (I tested with v2.4.0, v2.4.8 and v2.5.3) and a custom cluster with 3 all role nodes and 3 worker nodes. I used https://github.com/superseb/tf-do-rancher2
2. Once the custom cluster is successfully provisioned edit the cluster to add URL, username and password for a valid private registry with the applicable set of images.

**Result:**

Some nodes will successfully update to use images from the private registry, but on others the node-agent will be stuck with ImagePullBackoff due to a lack of credentials for the registry. On these nodes, it seems the node-agent Pod is updated to use the new image from the private registry before Rancher has established a connection to the node to run RKE reconciliation (and thus add registry credentials to the node). In testing with 3 all role nodes and 3 worker nodes, this seemed only to occur for worker nodes, due to the order of the updates. However, I was also able to trigger this condition testing the process with a single all role node custom cluster.

**Other details that may be helpful:**

This issue was initially reported in comments on a separate UI issue (https://github.com/rancher/rancher/issues/25147). In my testing, I have reproduced by enabling a private registry on a cluster provisioned without one, as I only had a single registry to-hand. @janeczku also reported reproduction when updating the settings for a cluster originally provisioned with a private registry in https://github.com/rancher/rancher/issues/25147#issuecomment-671348263

**Environment information**
- Rancher v2.4.0, v2.4.8 and v2.5.3
- For my testing, I had k8s version pinned to v1.17.0-rancher1-2

gz#8083

deniseschannon:
@axeal When attempting to update the private registry, you are referring to the registry as part of the cluster itself, correct?

axeal:
@deniseschannon Correct, the Private Registry in the Cluster Options per https://rancher.com/docs/rancher/v2.x/en/admin-settings/config-private-registry/#setting-a-private-registry-with-credentials-when-deploying-a-cluster

aaronRancher:
Tried to reproduce this issue on Rancher `v2.5.3`.

Steps:
Create a private registry with images for Rancher `v2.5.3`.
Create a custom cluster with 3 nodes all roles and 3 worker nodes.
Edit the cluster to use the private registry.
[...]

Issue #27306, opened 2020-05-28
Removing in-tree vSphere vCenter server entries does not propagate


DH-Rancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Initial config:

```yaml
rancher_kubernetes_engine_config:
  cloud_provider:
    name: vsphere
    vsphereCloudProvider:
      disk: {}
      global:
        insecure-flag: true
        soap-roundtrip-count: 0
      network: {}
      virtual_center:
        172.16.10.22:
          datacenters: /Homelab
          soap-roundtrip-count: 0
          user: administrator@vsphere.local
      workspace:
        datacenter: /Homelab
        folder: /Homelab/vm/K8S
        server: 172.16.10.22
```

Kube-apiserver container info:

```bash
root@vsp-intree1:~# docker inspect kube-apiserver | grep cloud
            "--cloud-config=/etc/kubernetes/cloud-config",
            "--cloud-provider=vsphere",
                "--cloud-config=/etc/kubernetes/cloud-config",
                "--cloud-provider=vsphere",
root@vsp-intree1:~# 
```

Contents of /etc/kubernetes/cloud-config

```bash
root@vsp-intree1:~# cat /etc/kubernetes/cloud-config

[Global]
user = ""
password = ""
insecure-flag = "true"
soap-roundtrip-count = "0"


[VirtualCenter "172.16.10.22"]
        user = "administrator@vsphere.local"
        password = "Removed"
        datacenters = "/Homelab"
        soap-roundtrip-count = "0"

[Workspace]
        server = "172.16.10.22"
        datacenter = "/Homelab"
        folder = "/Homelab/vm/K8S"
        default-datastore = ""
        resourcepool-path = ""

[Disk]

[Network]
```

Amend config to include second vcenter, point the workspace to it

```yaml
rancher_kubernetes_engine_config:
  cloud_provider:
    name: vsphere
    vsphereCloudProvider:
      disk: {}
      global:
        insecure-flag: true
        soap-roundtrip-count: 0
      network: {}
      virtual_center:
        172.16.10.22:
          datacenters: /Homelab
          soap-roundtrip-count: 0
          user: administrator@vsphere.local
        172.16.10.23:
          datacenters: /Homelab2
          soap-roundtrip-count: 0
          user: administrator@vsphere.local
      workspace:
        datacenter: /Homelab
        folder: /Homelab/vm/K8S
[...]

PR #33009, opened 2021-06-03
[2.5] Fix to properly remove vSphere vCenter server entries

rawmind0:
Issue https://github.com/rancher/rancher/issues/27306
Issue https://github.com/rancher/rancher/issues/30628
Issue https://github.com/rancher/rancher/issues/33128

Related to issue comment, https://github.com/rancher/rancher/issues/27306#issuecomment-832985239, the previous fix were not working fine on edge cases. When a cluster has been updated (updating rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter data), eventually the changes where not propagated properly to all k8s nodes. Also at affected nodes, the rancher agent became stopped without automatic recover. If rancher agent is manually run at affected nodes, the cluster use to recover itself. Similar behaviour observed for issues, https://github.com/rancher/rancher/issues/30628, https://github.com/rancher/rke/issues/2496 (trying to modify other rancherKubernetesEngineConfig data) 

The root cause of the issues seems to be related with the way that rancher agents are updated. Eventually, the rancher agent becomes updated in the middle of a node plan execution, making that the node plan execution aborts and the rancher agent is never restarted. Also, we found another related rancher agent config issue when using private registry with auth. The `cattle-private-registry` secret containing `.dockerconfigjson`  is wrong formatted when generated, so rancher agent is unable to pull docker image from private registry by itself. It needs private registry auth config at rke cluster level before work.

This PR fix proposal:
* <del>Fix rancher agent docker auth config when used private registry
  Create/Update `cattle-private-registry` secret containing `.dockerconfigjson` using proper format so rancher agent is able to pull docker image from private registry by itself</del> This code has been splitted into PR https://github.com/rancher/rancher/pull/33165
* <del>Wait for rancher agent daemonset rollout before update cluster
  Check that the `cattle-node-agent` daemonset has been properly updated and rollout, before continue with the cluster update. This is assuring that the rancher agent is updated and running before the cluster provision to avoid its updated during a node plan execution. This also add some security to the cluster update process due to if the new rancher agent config is not working properlly the cluster will not be updated and still ready and working</del> This code has been splitted into PR https://github.com/rancher/rancher/pull/33166
* Fix to properly remove vSphere vCenter server entries:
  The previous fix PR https://github.com/rancher/rancher/pull/30694, has been refactored to <del>just replace `rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter` entries if any virtualCenter has been removed.</del> to use k8s dynamic client to properly replace cluster data, if any `rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter` has been removed. 

The fix has been build and tested on all 3 mentioned issues scenarios and seems to work fine.

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -43,8 +43,10 @@ import (
 	rkeservices "github.com/rancher/rke/services"
 	rketypes "github.com/rancher/rke/types"
 	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/dynamic"
 )
 
 const (
@@ -63,6 +65,7 @@ type Store struct {
 	NodeLister                    v3.NodeLister
 	ClusterLister                 v3.ClusterLister
 	DialerFactory                 dialer.Factory
+	ClusterClient                 dynamic.ResourceInterface
 }
 
 type transformer struct {
@@ -136,6 +139,14 @@ func GetClusterStore(schema *types.Schema, mgmt *config.ScaledContext, clusterMa
 		NodeLister:                    mgmt.Management.Nodes("").Controller().Lister(),
 		DialerFactory:                 mgmt.Dialer,
 	}
+
+	dynamicClient, err := dynamic.NewForConfig(&mgmt.RESTConfig)
+	if err != nil {
+		logrus.Warnf("GetClusterStore error creating K8s dynamic client: %v", err)
+	} else {
+		s.ClusterClient = dynamicClient.Resource(v3.ClusterGroupVersionResource)
+	}
+
 	schema.Store = s
 	return s
 }
@@ -544,21 +555,19 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data 
 		return nil, err
 	}
 
-	// Replace rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter value to properly update virtualCenter removal
+	// When any rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter has been removed, updating cluster using k8s dynamic client to properly
+	// replace cluster spec. This is required due to r.Store.Update is merging this data instead of replacing it, https://github.com/rancher/rancher/issues/27306
 	if newVCenter, ok := values.GetValue(data, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter"); ok && newVCenter != nil {
-		oldVCenter, ok := values.GetValue(existingCluster, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter")
-		if ok && oldVCenter != nil && !reflect.DeepEqual(newVCenter, oldVCenter) {
-			newData := map[string]interface{}{}
-			for k, v := range data {
-				newData[k] = v
+		if oldVCenter, ok := values.GetValue(existingCluster, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter"); ok && oldVCenter != nil {
+			if oldVCenterMap, oldOk := oldVCenter.(map[string]interface{}); oldOk && oldVCenterMap != nil {
+				if newVCenterMap, newOk := newVCenter.(map[string]interface{}); newOk && newVCenterMap != nil {
+					for k := range oldVCenterMap {
+						if _, ok := newVCenterMap[k]; !ok && oldVCenterMap[k] != nil {
+							return r.updateClusterByK8sclient(apiContext.Request.Context(), id, updatedName, data)
+						}
+					}
+				}
 			}
-			// Update virtualCenter value to nil
-			values.PutValue(newData, nil, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter")
-			if _, err := r.Store.Update(apiContext, schema, newData, id); err != nil {
-				return nil, err
-			}
-			// Set virtualCenter value to newVCenter
-			values.PutValue(data, newVCenter, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter")
 		}
 	}
 
@@ -607,6 +616,35 @@ func (r *Store) getDynamicField(data map[string]interface{}) (string, error) {
 	}
 
 	return "", nil
+}
+
+func (r *Store) updateClusterByK8sclient(ctx context.Context, id, name string, data map[string]interface{}) (map[string]interface{}, error) {
+	logrus.Tracef("Updating cluster [%s] using K8s dynamic client", id)
+	if r.ClusterClient == nil {
+		return nil, fmt.Errorf("Error updating the cluster: k8s client is nil")
+	}
+
+	object, err := r.ClusterClient.Get(ctx, id, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Error updating the cluster: %v", err)
+	}
+
+	// Replacing name by displayName to properly update cluster using k8s format
+	values.PutValue(data, name, "displayName")
+	values.RemoveValue(data, "name")
+
+	// Setting data as cluster spec before update
+	object.Object["spec"] = data
+	object, err = r.ClusterClient.Update(ctx, object, metav1.UpdateOptions{})
+	if err != nil || object == nil {
+		return nil, fmt.Errorf("Error updating the cluster: %v", err)
+	}
+
+	// Replacing displayName by name to properly return updated data using API format
+	values.PutValue(object.Object, name, "spec", "name")
+	values.RemoveValue(object.Object, "spec", "displayName")
+
+	return object.Object["spec"].(map[string]interface{}), nil
 }
 
 func canUseClusterName(apiContext *types.APIContext, requestedName string) error {


rawmind0:
With the patch applied, the cluster update behaviour can be changed based on `agent-rollout-wait` setting:
* `agent-rollout-wait="false"`: Current behaviour
  cluster updated by user -> agent deployment NO wait for rollout (client tunnel) -> cluster update (rke up - eventually, agent rollout occurs in the middle of this execution causing the issue) -> cluster active/error
* `agent-rollout-wait="true"`: Proposed behaviour
  cluster updated by user -> agent deployment wait for rollout (client tunnel) -> cluster update (rke up - agent is already rollout) -> cluster active
  On tests done on this scenario, the cluster update seems safer and "easier" to recover in that way. When update a cluster, if a new agent config is not working, the agent rollout is timeout and the cluster update is stopped. The cluster became on error on ui, but the k8s is perfectly working due to has not been updated. To recover the state, editing the cluster undoing the last changes and saving, makes that the agent is redeployed with previous working config.

pkg/settings/setting.go
```diff
@@ -20,6 +20,8 @@ var (
 	InjectDefaults string
 
 	AgentImage                        = NewSetting("agent-image", "rancher/rancher-agent:master-head")
+	AgentRolloutTimeout               = NewSetting("agent-rollout-timeout", "300s")
+	AgentRolloutWait                  = NewSetting("agent-rollout-wait", "true")
```
superseb, line 24: This is changing existing behavior and might have side effects so we need docs on this so users know what they can configure if something happens.

superseb (approved):
I think overall this is brining more reliability but could have some side effects so the earlier we can get it in for testing, the better.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -544,21 +544,34 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
...
+					}
+					// Update just if any virtualcenter has been removed
+					if oldRowRemoved {
+						newData := map[string]interface{}{}
+						for k, v := range data {
+							newData[k] = v
+						}
+						// Update virtualCenter value to nil
+						values.PutValue(newData, nil, "rancherKubernetesEngineConfig", "cloudProvider", "vsphereCloudProvider", "virtualCenter")
+						if _, err := r.Store.Update(apiContext, schema, newData, id); err != nil {
```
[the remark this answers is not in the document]
rawmind0: @superseb , you are right, this is causing the cluster to go into provision state twice. But the root cause of the issue seems to be at the agent restart more than the second update. Anyway, here i'm thinking to add a cluster annotation to avoid the first cluster update, removing it when come bak and update again, what do you think??

pkg/settings/setting.go
rawmind0: Yes, this is changing the cluster update behaviour as mentioned at previous comments. Agreed that might have side effects, so we should deep test it. This was one of the main reasons to add the parameters, to disable/config the "new" behaviour if needed.

pkg/api/norman/store/cluster/cluster_store.go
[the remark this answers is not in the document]
rawmind0: @superseb , I've added a `io.cattle.rke.no.update` cluster annotation to skip for `r.Store.Update`. Could you please take a view??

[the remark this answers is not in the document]
superseb: Just to be clear here, this approach is needed because with the current API behavior, we cannot delete an entry because it will merge the given input, correct? So to work around this, we empty the value (first `Store.Update`), and then put the correct value (second/last `Store.Update`). Because `Store.Update` will trigger cluster provisioning, we add an annotation to make sure that it doesn't trigger provisioning on the first `Store.Update`. And this approach was preferred because we didn't want to change the API because of scope of impact/unexpected behavioral change?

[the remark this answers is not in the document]
rawmind0: That's it.

superseb (approved):

dramich (changes requested):
This PR should be split into at least two, possibly three PRs. This one should focus on just the vsphere fix, the other pieces should move to their own. Doing this will help keep this PR cleaner, make it easier for QA to test each change individually and also in the case of a revert, make it cleaner. 

For the fix for vsphere, we should not call update twice, and with a blank config. This causes issue as we have seen since the cluster will then run through ALL associated handlers. 

A better option would be to to the comparison as this does and if `oldRowRemoved` ends up true then amend the request to have `replace=true` which should skip the merge logic in the store for updates. 

From testing, the UI sends this and we don't hit the issue. When an API request is made without `replace=true` then the bug is encountered.

rawmind0:
> This PR should be split into at least two, possibly three PRs. This one should focus on just the vsphere fix, the other pieces should move to their own. Doing this will help keep this PR cleaner, make it easier for QA to test each change individually and also in the case of a revert, make it cleaner.

Splitted PR into 3:
* this PR for vsphere fix
* https://github.com/rancher/rancher/pull/33165 for private registry auth fix
* https://github.com/rancher/rancher/pull/33166 for wait for agent rollout fix

> For the fix for vsphere, we should not call update twice, and with a blank config. This causes issue as we have seen since the cluster will then run through ALL associated handlers.
> 
> A better option would be to to the comparison as this does and if oldRowRemoved ends up true then amend the request to have replace=true which should skip the merge logic in the store for updates.
> 
> From testing, the UI sends this and we don't hit the issue. When an API request is made without replace=true then the bug is encountered.

@dan , on my tests, adding the `_replace=true` to the request (as UI do) seems not working for this case. Which Rancher version are you using for testing?? 
For me, the UI is hitting the same issue for Rancher v2.5.5 and lower and dev version, built from current `release/v2.5.9` branch but removing the previous patch. Just Rancher v2.5.6, v2.5.7 and v2.5.8 UI is not hitting the same issue, but on these versions the previous version of this patch is already applied.

rawmind0:
@dan, updated PR to use k8s dynamic client to properly replace cluster data, if any `rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter` has been removed. 

Thanks for your guidance on this.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -609,6 +613,31 @@ func (r *Store) getDynamicField(data map[string]interface{}) (string, error) {
 	return "", nil
 }
 
+func (r *Store) updateClusterByK8sclient(id, name string, data map[string]interface{}) (map[string]interface{}, error) {
+	logrus.Tracef("Updating cluster [%s] using K8s dynamic client", id)
+	object, err := r.ClusterClient.Get(context.TODO(), id, metav1.GetOptions{})
+	if err != nil {
```
superseb, line 619: Are the errors returned here OK for user facing or should we wrap them?

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -544,21 +551,18 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
 		return nil, err
 	}
 
-	// Replace rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter value to properly update virtualCenter removal
+	// When any rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter has been removed, updating cluster using k8s dynamic client
```
superseb, line 554: I think we should add why we are doing this using the dynamic client now so a person touching this code later knows why its not using Store.Update

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -126,6 +129,8 @@ func GetClusterStore(schema *types.Schema, mgmt *config.ScaledContext, clusterMa
 		ClusterManager: clusterManager,
 	}
 
+	dynamicClient, _ := dynamic.NewForConfig(&mgmt.RESTConfig)
```
superseb, line 132: We dont want to check for error here?

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: What do you mean??  I was getting some errors on testing, and they were showed properly to user facing

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Updated

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Updated

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Updated

pkg/api/norman/store/cluster/cluster_store.go
superseb: I was actually referring to why we are not using `Store.Update` but we can use Git history for that

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -609,6 +615,35 @@ func (r *Store) getDynamicField(data map[string]interface{}) (string, error) {
...
+
+	// Replacing name by displayName to properly update cluster using k8s format
+	values.PutValue(data, name, "displayName")
+	values.RemoveValue(data, "name")
+
+	// Setting data as cluster spec before update
+	object.Object["spec"] = data
+	object, err = r.ClusterClient.Update(context.TODO(), object, metav1.UpdateOptions{})
+	if err != nil || object == nil {
+		return nil, fmt.Errorf("Updating cluster [%s] using K8s dynamic client: %v", id, err)
```
superseb, line 637: Can we change to `Error updating cluster ...` so that we indicate there was an error

superseb (dismissed):

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Updated

superseb (dismissed):

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -609,6 +616,35 @@ func (r *Store) getDynamicField(data map[string]interface{}) (string, error) {
 	return "", nil
 }
 
+func (r *Store) updateClusterByK8sclient(id, name string, data map[string]interface{}) (map[string]interface{}, error) {
+	logrus.Tracef("Updating cluster [%s] using K8s dynamic client", id)
+	if r.ClusterClient == nil {
+		return nil, fmt.Errorf("Error updating the cluster: k8s client is nil")
+	}
+
+	object, err := r.ClusterClient.Get(context.TODO(), id, metav1.GetOptions{})
```
dramich, line 625: Pass in the apiContext here and use it vs doing a TODO
```suggestion
	object, err := r.ClusterClient.Get(apiContext.Request.Context(), id, metav1.GetOptions{})
```

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -136,6 +139,12 @@ func GetClusterStore(schema *types.Schema, mgmt *config.ScaledContext, clusterMa
 		NodeLister:                    mgmt.Management.Nodes("").Controller().Lister(),
 		DialerFactory:                 mgmt.Dialer,
 	}
+
+	dynamicClient, err := dynamic.NewForConfig(&mgmt.RESTConfig)
```
dramich, line 143: I don't think we could hit an error here since other places will fatal on a bad mgmt.RESTConfig but we should at least log the error at a warning level so we have a record in the logs that we don't have the dynamic client

pkg/api/norman/store/cluster/cluster_store.go
dramich: And in the other spot TODO is used

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -63,6 +65,7 @@ type Store struct {
 	NodeLister                    v3.NodeLister
 	ClusterLister                 v3.ClusterLister
 	DialerFactory                 dialer.Factory
+	ClusterClient                 dynamic.NamespaceableResourceInterface
```
dramich, line 68: Swap to ResourceInterface, since a cluster is not in a namespace

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Updated

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Updated logging a warning if fails

pkg/api/norman/store/cluster/cluster_store.go
rawmind0: Done

dramich (approved):

superseb (approved):

Issue #32878, opened 2021-05-26
Rancher chart post-delete hook fails on hardened cluster


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Install Rancher on CIS 1.5 compliant cluster RKE1
  -  Details on pre-requirements [here](https://rancher.com/docs/rancher/v2.x/en/security/rancher-2.4/hardening-2.4/).
- Uninstall the Rancher chart `helm uninstall rancher -n cattle-system`

**Result:**

```
Events:
  Type     Reason        Age                From            Message
  ----     ------        ----               ----            -------
  Warning  FailedCreate  18s (x3 over 48s)  job-controller  Error creating: pods "rancher-post-delete-" is forbidden: PodSecurityPolicy: unable to admit pod: []
```

**Other details that may be helpful:**

**Environment information**
- Rancher version: `2.6 master-head (05/25/2021)` _f90b36e_
- Installation option: HA RKE1

izaac:
Rancher version `2.6 master-head (06/06/2021)` _b0c0eb4_

This is failing now with:


```
k describe job rancher-post-delete -n cattle-system
Name:           rancher-post-delete
Namespace:      cattle-system
Selector:       controller-uid=65e6336a-20c8-4984-a894-657184783487
Labels:         app=rancher
                chart=rancher-2.5.8
                heritage=Helm
                release=rancher
Annotations:    helm.sh/hook: post-delete
                helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
                helm.sh/hook-weight: 3
Parallelism:    1
Completions:    1
Pods Statuses:  0 Running / 0 Succeeded / 0 Failed
Pod Template:
  Labels:           app=rancher
                    chart=rancher-2.5.8
                    controller-uid=65e6336a-20c8-4984-a894-657184783487
                    heritage=Helm
                    job-name=rancher-post-delete
                    release=rancher
  Service Account:  rancher-post-delete
  Containers:
   rancher-post-delete:
    Image:      rancher/shell:v0.1.6
    Port:       <none>
    Host Port:  <none>
    Command:
      /scripts/post-delete-hook.sh
    Environment:
      NAMESPACES:          fleet-system cattle-system rancher-operator-system
      TIMEOUT:             120
      IGNORETIMEOUTERROR:  false
    Mounts:
      /scripts from config-volume (rw)
  Volumes:
   config-volume:
    Type:      ConfigMap (a volume populated by a ConfigMap)
[...]

PR #33023, opened 2021-06-03
add podSecurityPolicy to the post-delete-hook

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/32878

Problem:
The post-delete-hook pod fails to be scheduled because it does not have and can not use any psp. 

Solution:
Add the missing permission and psp resource. 

Note:
Rancher itself can be installed and uninstalled successfully because it uses the cluster-admin clusterrole. 

Next Step:

- [ ] Backport to the v2.5 branch.

### DIFF
--- a/chart/templates/post-delete-hook-psp.yaml
+++ b/chart/templates/post-delete-hook-psp.yaml
@@ -1,0 +1,32 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ include "rancher.fullname" . }}-post-delete
+  labels: {{ include "rancher.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+spec:
+  privileged: false
+  hostNetwork: false
+  hostIPC: false
+  hostPID: false
+  runAsUser:
+    rule: 'RunAsAny'
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'MustRunAs'
+    ranges:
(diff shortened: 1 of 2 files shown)


nickgerace (approved):

chart/templates/post-delete-hook-psp.yaml
```diff
@@ -0,0 +1,31 @@
...
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+spec:
+  privileged: false
+  hostNetwork: false
+  hostIPC: false
+  hostPID: false
+  runAsUser:
+    rule: 'MustRunAsNonRoot'
```
aiyengar2, line 16: Correct me if I'm wrong, but isn't the user a root user?

https://github.com/rancher/rancher/blob/517013e614b0163f24d48cabe72b0f62c91595cb/chart/templates/post-delete-hook-job.yaml#L26

aiyengar2 (changes requested):

chart/templates/post-delete-hook-psp.yaml
jiaqiluo: @aiyengar2  Good Catch! Yes, the pod runs as a root user. 
I have fixed the PSP to fix its rule, and also update the clusterrole to use the specific PSP. 
It worked because it used other PSPs which it can pass in the cluster.

aiyengar2 (approved):
LGTM

nickgerace (dismissed):

cbron (dismissed):

jiaqiluo:
The PR is already approved by 3 reviewers. 
I just rebase the branch, hopefully, the build will pass this time.

nickgerace (approved):

Issue #13666, opened 2018-05-23
After disabling and enabling AD using a different user for test and enable auth , this user is not allowed to log in to rancher server anymore. 


sangeethah:
**Rancher versions: v2.0.1-rc6
**Steps to Reproduce:**

Enable AD using user - "testuser1" for test and enable auth.
Add "testuser2"in the "Site Access" list - "Allow members of Clusters, Projects, plus Authorized Users and Organizations".

Make sure that both "testuser1 and "testuser2" are allowd to access rancher-server.

Disable AD auth.
Enable AD auth using user - "testuser2" for test and enable auth.

Try to log in as "testuser2". Log in fails with following errors seen in logs:

"2018/05/23 16:33:21 [ERROR] API error response 500 for POST /v3-public/activeDirectoryProviders/activedirectory?action=login. Cause: can't find unique user for principal activedirectory_user://CN=<cn>,CN=Users,DC=<dc1>,DC=<dc2>,DC=<dc3>"

maxkorp:
CC @cjellick @deniseschannon 

So the flow here is kind of odd but I've found where it dies.

- Create testuser2. This creates `u-abc` with no AD principal.
- Enable AD.
- Login to testuser2 via AD. This creates `u-xyz` with the AD principal `AD-testuser2`.
- Disable AD.
- Login with testuser2 (`u-abc`).
- Enable AD using the `testuser2` as the AD account. This associates the AD principal `AD-testuser2` with `u-abc`. 

At this point, both `u-abc` and `u-xyz` are associated with the same AD principal, which is where the login fails.

I have a few solutions proposed below. I think the core decision factor here is if  users with the same user name created pre auth, and then logged in via AD are intentionally separate users, or if they really should be bound together. I don't know if that's a constant prescription we make one way or the other, or if that varies from user/install to user/install. 

Potential solutions, including some which may work in concert:
- When you login with AD, if no user already exists with that primary, attempt to automatically bind to an existing local user with the same username (if the user is able to log in). 
- Same as above, but give the user the option to make the binding to the found user
- Some way to manually merge existing accounts or add/remove/transfer principals. I know this something already being talked about, but it directly impacts this bug. 
- When activating AD, if another user is found with the same principal, rather than just blindly binding the principal to the current user, allow the user to remove the principal from the old user (would have to enforce a password set on local login, as well as potentially generating a local principal or login name or whatever), or to log in to AD as a different user. Generally, disallow getting the principal into a bad state.
[...]

PR #33029, opened 2021-06-03
Check that we are not binding a user principal twice

paynejacob:
https://github.com/rancher/rancher/issues/13666

# Problem
When setting up an external authentication provider the user is asked to test the setup with an external user.  On success the principal for the external user is bound to the current user.  If you login with a second external user a new user is created for them.  Disabling and enabling auth with the second external user will bind their principal to the current local account.  The principal is now bound to the local account and the user created when the user authenticated normally.  There is no way to determine what users is the real one and a 500 error is returned.

# Solution 
When a provider is enabled I added a check to see if the principal is currently bound to any user other than the current local user.  If another user is found an error is returned and details are logged.  The user can then delete the conflicting account, login with the conflicting account during setup, or manually remove the principal.  I opted for a cowardly approach because the only way to correct this is to unbind the principal from the existing user and this could have implications on auditing historical data.

# Testing
1. Enable Authentication with testuser1 
2. login with testuser2
3. disable authentication
4. enable authentication and use testuser2
5. Attempt to login with testuser2

Without the patch applied you should get a 500 on step 5.  With the patch applied you should get an error on step 4 and see details in the rancher logs.  I tested this using Keycloak but this error has been reported with Active Directory and Okta as well.  I believe it will affect the following providers:

- azure
- github
- googleoauth
- ldap
- oidc
- keycloak
- Active Directory


pkg/auth/providers/common/usermanager.go
```diff
@@ -149,6 +149,14 @@ func (m *userManager) SetPrincipalOnCurrentUserByUserID(userID string, principal
 		return nil, err
 	}
 
+	// ensure this principal is unique to this user
+	if conflict, err := m.GetUserByPrincipalID(principal.Name); err != nil {
+		return nil, err
+	} else if conflict != nil && conflict.UID != user.UID {
+		logrus.Errorf("Cannot set principal \"%s\" on user \"%s\". Principal already in use on user \"%s\".", principal.Name, user.UID, conflict.UID)
```
superseb, line 156: Not sure if this is the format in here, in RKE we usually use `[%s]` for variables.

pkg/auth/providers/common/usermanager.go
```diff
@@ -149,6 +149,14 @@ func (m *userManager) SetPrincipalOnCurrentUserByUserID(userID string, principal
...
+		return nil, errors.New("cannot set principal on user that is already in use")
```
superseb, line 157: Is the goal to inform the user with this error and then they have to lookup the error in the logs to find what is causing it? Is this for security reasons?

paynejacob: I did separate error messages because this error is returned on a callback with the external provider.  I want to limit external exposures of user information as much as possible.  The user will need to check the logs to debug this anyway.  The log message should have enough information for them to take action.

pkg/auth/providers/common/usermanager.go
paynejacob: I like that better ill switch

superseb (dismissed):

dramich (commented):
This doesn't feel like the right fix. It will technically help tell a user what happened but not stop the underlying issue which is adding the principal onto the original admin account `user-`. With this fix auth can't be turned back on unless a user is deleted or the principal field is edited which seems like a poor user experience. I think we need to consider just not updated the local admin account with the user principal of whoever is turning on auth for the second time

paynejacob:
I think that solution makes more sense.  It is not a great user experience as is because you have to check the log to see the error.  Given the user is already bound to another user I think not binding them makes sense.  The work around we propose make assumptions about what the user want to do if the admin account is already bound.  By just not binding we are letting them set up auth.   If they want to rebind the local user they can take steps on their own to do so.

dramich:
@paynejacob I think we need to understand why we are attempting to save the principal back to the original `user-` account. Are we somehow ending up with the wrong user during testAndApply?

paynejacob:
Tested this on 2.6

1. login as local admin user
2. setup keycloak SAML with testuser1
3. in incognito tab login with keycloak as testuser2
4. disable keycloak auth
5. login as local admin
6. setup keycloak SAML with testuser2
7. 
image

pkg/auth/providers/saml/saml_client.go
```diff
@@ -326,10 +326,13 @@ func (s *Provider) HandleSamlAssertion(w http.ResponseWriter, r *http.Request, a
 	userID := s.clientState.GetState(r, "Rancher_UserID")
 	if userID != "" && rancherAction == testAndEnableAction {
 		user, err := s.userMGR.SetPrincipalOnCurrentUserByUserID(userID, userPrincipal)
-		if err != nil {
+		if err != nil && user == nil {
 			log.Errorf("SAML: Error setting principal on current user %v", err)
 			http.Redirect(w, r, redirectURL+"errorCode=500", http.StatusFound)
 			return
+		} else if err != nil && user != nil {
+			http.Redirect(w, r, redirectURL+"errorCode=422&errorMsg="+url.QueryEscape(err.Error()), http.StatusFound)
```
snasovich, line 334: Did you confirm UI is handling this as expected? It's almost identical to code we already have and `errorMsg` is error ID in there that UI uses to come up with the actual error message and is the actual error message in this case. (Actually glancing through UI code it looks like it will fallback to using errorMsg verbatim if it cannot find it as code). Still, for i18n support we better use code here instead. I'm OK making it a separate enhancement thought and let this one go in as is due to this improvement requiring both UI and backend change.

```
		// UI uses this translation key to get the error message
		http.Redirect(w, r, redirectURL+"errorCode=422&errorMsg="+UITranslationKeyForErrorMessage, http.StatusFound)
```

snasovich (approved):
LGTM, one "nice to have" comment

pkg/auth/providers/saml/saml_client.go
paynejacob: Yes it works with the up, there was not a code for this error message so I sent this for now.  We can open it as a separate issue.

thedadams (approved):

davidnuzik:
Blocks QA from testing https://github.com/rancher/dashboard/issues/3841

snasovich:
@paynejacob , we need this opened against `release/v2.6`. Unfortunately changing branch in GH doesn't work to well for this so I'm asking you to reopen one it for `release/v2.6`.

paynejacob:
closing in favor of https://github.com/rancher/rancher/pull/34805

PR #33034, opened 2021-06-03
Avoid using Helm v2 for lint and indexing

dirkmueller:
We can be modern and use the latest and greatest
for linting and repo building.

### DIFF
--- a/scripts/chart/index
+++ b/scripts/chart/index
@@ -12,4 +12,4 @@ repo_index=${1}
 echo "Getting current index from ${repo_index}"
 curl -f -H 'Cache-Control: max-age=0,no-cache' -H 'Host: releases.rancher.com' "https://c.storage.googleapis.com/server-charts/${repo_index}/index.yaml?$(date +%s%N)" -o build/index.yaml
 
-helm repo index --merge build/index.yaml ./bin/chart/${repo_index}
+helm_v3 repo index --merge build/index.yaml ./bin/chart/${repo_index}
--- a/scripts/chart/test
+++ b/scripts/chart/test
@@ -3,8 +3,7 @@
 echo "-- chart/test --"
 
 # Check for helm
-hash helm >/dev/null 2>&1
-if [[ $? > 0 ]]; then
+if [ -z "$(type -p helm)" ]; then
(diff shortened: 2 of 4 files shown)


brendarearden (approved):
lgtm

scripts/chart/test
```diff
@@ -3,8 +3,7 @@
 echo "-- chart/test --"
 
 # Check for helm
-hash helm >/dev/null 2>&1
-if [[ $? > 0 ]]; then
+if [ -z "$(type -p helm)" ]; then
```
aiyengar2, line 6: Any reason why we can't use helm_v3 here? Plugins should still be supported in Helm 3

scripts/chart/index
```diff
@@ -12,4 +12,4 @@ repo_index=${1}
 echo "Getting current index from ${repo_index}"
 curl -f -H 'Cache-Control: max-age=0,no-cache' -H 'Host: releases.rancher.com' "https://c.storage.googleapis.com/server-charts/${repo_index}/index.yaml?$(date +%s%N)" -o build/index.yaml
 
-helm repo index --merge build/index.yaml ./bin/chart/${repo_index}
+helm_v3 repo index --merge build/index.yaml ./bin/chart/${repo_index}
```
aiyengar2, line 15: Just to clarify, the reason why you are suggesting this change is to ensure that helm_v3 is used instead of helm, which points to helm v2 based on the following code, right?

https://github.com/rancher/rancher/blob/51d074b202ee3445bb87b827fc1f06f5a66627ff/Dockerfile.dapper#L55-L68

aiyengar2 (changes requested):

scripts/chart/test
aiyengar2: tried this out locally and I see that there seems to be an issue with the way rancher/helm-unittest parses the ARCH when used as a plugin for Helm 3... resolving this comment.

aiyengar2:
@dirkmueller I cherry-picked your changes and merged them in https://github.com/rancher/rancher/pull/33487, so I'm closing this PR out in favor of that one.

PR #33035, opened 2021-06-03
No unzip

dirkmueller:
When *.zip consist of exactly one file, you can use gunzip
for it, so we don't need unzip in the container.

Bonus for not leaving a stray *.zip behind inside the container,
saving twice on size.


StrongMonkey (approved):
:+1:

dramich (approved):

Issue #32657, opened 2021-05-10
No logs seen when deploying a feature chart and kubectl shell is broken


sowmyav27:
**On master-head commit id: `8c014efa`**

- Deploy a downstream cluster using Ember UI.
- Access cluster explorer of this cluster
- deploy a feature chart - CIS benchmark via -- Apps and Martketplace or via Cluster Tools
- Deploy chart
- No logs seen

<img width="1325" alt="Screen Shot 2021-05-10 at 11 12 10 AM" src="https://user-images.githubusercontent.com/26032343/117705176-95145500-b180-11eb-866f-1a97e0e8b606.png">

- Error seen in console
```
WebSocket connection to 'wss://<server>/k8s/clusters/c-c5qtq/v1/catalog.cattle.io.operations/cis-operator-system/helm-operation-xmdw6/logs?previous=false&follow=true&timestamps=true&pretty=true&sinceSeconds=1800&sockId=4' failed: 
```
and
<img width="467" alt="Screen Shot 2021-05-10 at 11 13 38 AM" src="https://user-images.githubusercontent.com/26032343/117705350-cab93e00-b180-11eb-9b5e-f09cf98a5860.png">

- kubectl shell is also broken when accessed through Ember UI and dashboard
- Error seen in Ember:
```
WebSocket connection to 'wss://<server>/v3/clusters/c-c5qtq?shell=true' failed: 
```
- Error seen in Dashboard/Cluster explorer UI:
```
socket.js:86 WebSocket connection to 'wss://<server>/v1/management.cattle.io.clusters/c-c5qtq?link=shell&stdout=1&stdin=1&stderr=1&tty=1&command=%2Fbin%2Fsh&sockId=8' failed: 
```

kravciak:
Works fine in master-head (5715b8a)
 - CIS benchmark installation shows logs
 - I can open kubectl shell in both EmberUI & Dashboard

PR #33040, opened 2021-06-04
Fix cluster shell and helm operation

StrongMonkey:
Remove Impersonate-Extra- Header that were sent through proxy header.
This is not allowed in k8s api server as requests can't only have extra
header without users.

### DIFF
--- a/pkg/api/steve/clusters/shell.go
+++ b/pkg/api/steve/clusters/shell.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"net/http/httputil"
+	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/32657

dramich (dismissed):

cmurphy (commented):
LGTM. I was able to get a shell on GKE, AKS, and k3s 1.21 clusters using this. To test it I had to change both the cattle-agent image and the rancher deployment image in the downstream cluster to use this change.

dramich (approved):

PR #33042, opened 2021-06-04
Change to partial token info logging / full token trace logging

snasovich:
(Not more than) half of token is logged at info level now. Full token logging is now at trace level.
Replaces https://github.com/rancher/rancher/pull/32851

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -349,7 +349,8 @@ func run(ctx context.Context) error {
 		if !isConnect() {
 			wsURL += "/register"
 		}
-		logrus.Infof("Connecting to %s with token %s", wsURL, token)
+		logrus.Infof("Connecting to %s with token starting with %s", wsURL, token[:len(token)/2])
+		logrus.Tracef("Connecting to %s with token %s", wsURL, token)
 		remotedialer.ClientConnect(ctx, wsURL, headers, nil, func(proto, address string) bool {
 			switch proto {
 			case "tcp":


StrongMonkey (approved):

cbron (approved):

cmd/agent/main.go
```diff
@@ -349,7 +349,8 @@ func run(ctx context.Context) error {
 		if !isConnect() {
 			wsURL += "/register"
 		}
-		logrus.Infof("Connecting to %s with token %s", wsURL, token)
+		logrus.Infof("Connecting to %s with token starting with %s", wsURL, token[:len(token)/2])
```
dramich, line 352: Late to the party but doing something like this would help so we don't double log
```
logToken = token
if logrus.GetLevel() < logrus.TraceLevel {
    logToken = token[:len(token)/2]
}
```

snasovich: @dramich , good idea. However, I would say logging `"Connecting to %s with token %s"` with partial token could be a bit misleading. We could have it as something like `logToken = token[:len(token)/2] + "<REMAINDER_REDACTED>"` to make it more obvious.
Now, I would imagine in this case we would log at INFO level in both cases. And that's where I'm not sure it's entirely clean to have content of something logged at INFO level affected by logging level set.
Anyway, thank you for feedback, but I'm hoping you're OK with leaving it "as implemented". Let me know by either replying or resolving this convo.

dramich: The check could be used to do any multitude of things, including modifying the message or just logging at trace in general, my main concern is double logging. This is the agent so there aren't a lot of logs to begin with, even at trace, but it should be kept clean if possible.

snasovich: Open to discussing it further, but I would be against logic like "if logging level is DEBUG or below, log stripped down message at INFO level; otherwise, log full message at TRACE level" as it would result in that INFO log message no longer produced when logging is set to TRACE. Basically, I believe that when you switch logging level to more verbose one, the expectation is you will have the exact same messages logged as before, plus more that are coming from more verbose levels. This way if we have someone/something relying on detecting/parsing out these specific INFO messages, they are not affected when more verbose logging is set.
The other option of "if logging is at TRACE, log full token with INFO message" is not great either as it then appears we're including sensitive info in INFO log.
All this said, I think it's fairly common to have multiple log messages logged for a single operations like "INFO - connecting to ..." and "DEBUG - oh by the way, the following credentials were used". Here we're duplicating some info in TRACE message from INFO so it includes everything that's needed to look at it in isolation, without looking into previous message.

Issue #33036, opened 2021-06-03
Keycloak OIDC - API returns 500 on user search autocomplete input fields on Ember


izaac:
**What kind of request is this:**

**Steps to reproduce:**

- Enable Keycloak OIDC from Dashboard
- Have a cluster ready (I used a downstream cluster)
- Navigate to `Cluster Manager` (Legacy Apps)
- Edit downstream cluster or add cluster member page
- Start typing in the autocomplete input/dropdown

**Result:**

500 from the API `/v3/principals?action=search`

**Other details that may be helpful:**

```
2021/06/03 21:28:31 [TRACE] REST UPDATE apis/management.cattle.io/v3//userattributes/user-tcszx
2021/06/03 21:28:31 [TRACE] REST GET apis/management.cattle.io/v3//authconfigs/keycloakoidc
2021/06/03 21:28:31 [ERROR] Panic serving api request:
goroutine 153746 [running]:
runtime/debug.Stack(0xc00d9c9060, 0x488a7c0, 0x82f9f60)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/rancher/norman/api.(*Server).ServeHTTP.func1(0x5b3f420, 0xc001e33d70)
	/go/pkg/mod/github.com/rancher/norman@v0.0.0-20210513204752-e48df26b54bd/api/server.go:175 +0x85
panic(0x488a7c0, 0x82f9f60)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
net/http.(*Client).deadline(0x0, 0x0, 0xc00d9c9338, 0x40dc9b)
	/usr/local/go/src/net/http/client.go:188 +0x22
net/http.(*Client).do(0x0, 0xc002ff2c00, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/http/client.go:603 +0x2bd
net/http.(*Client).Do(...)
	/usr/local/go/src/net/http/client.go:585
github.com/rancher/rancher/pkg/auth/providers/keycloakoidc.(*KClient).getFromKeyCloak(0xc000fbe1b8, 0x0, 0x0, 0xc00f2c7700, 0x3c, 0xc00fb25a40, 0x3c, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/rancher/rancher/pkg/auth/providers/keycloakoidc/keycloak_client.go:173 +0x355
[...]

PR #33048, opened 2021-06-04
Create check for cert and key to avoid error

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/33036

**Problem**
Client created as part of configuring Keycloak is not being used for the http call
The cert and key values are not being checked before attempting to add them to the http client
Http client is nil
Copy paste resulted in auth type of "token" being used

**Solution**
Remove the new client that is being created and use the one that is part of the keycloakClient
Ensure that cert and key are checked prior to attempting to add them to the http client tls settings 
Add code to set the http client when its created. 
Change so auth type of "bearer" is being used.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -9,15 +9,14 @@ import (
 	"strings"
 
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-	"github.com/rancher/rancher/pkg/auth/providers/oidc"
 	"github.com/sirupsen/logrus"
 )
 
 //account defines properties an account in keycloak has
 type account struct {
-	ID            int    `json:"id,omitempty"`
+	ID            string `json:"id,omitempty"`
 	Email         string `json:"email,omitempty"`
-	EmailVerified string `json:"emailVerified,omitempty"`
+	EmailVerified bool   `json:"emailVerified,omitempty"`
 	Username      string `json:"username,omitempty"`
 	Enabled       bool   `json:"enabled,omitempty"`
 	Name          string `json:"firstName,omitempty"`
@@ -27,7 +26,7 @@ type account struct {
 
 //Group defines properties a group in keycloak has
 type Group struct {
-	ID        int     `json:"id,omitempty"`
+	ID        string  `json:"id,omitempty"`
 	Name      string  `json:"name,omitempty"`
 	Subgroups []Group `json:"subGroups,omitempty"`
 }
@@ -35,12 +34,6 @@ type Group struct {
 //KeyCloakClient implements a httpclient for keycloak
 type KeyCloakClient struct {
 	httpClient *http.Client
-}
-
-func newClient(config *v32.OIDCConfig) (*KeyCloakClient, error) {
-	keyCloakClient := &KeyCloakClient{}
-	err := oidc.GetClientWithCertKey(keyCloakClient.httpClient, config.Certificate, config.PrivateKey)
-	return keyCloakClient, err
 }
 
 func (k *KeyCloakClient) searchPrincipals(searchTerm, principalType string, accessToken string, config *v32.OIDCConfig) ([]account, error) {
@@ -161,7 +154,7 @@ func (k *KeyCloakClient) getFromKeyCloak(accessToken, url string) ([]byte, int, 
 	if accessToken == "" {
 		return nil, 500, fmt.Errorf("[keycloak oidc]: GET request failed, missing authorization token")
 	}
-	req.Header.Add("Authorization", "token "+accessToken)
+	req.Header.Add("Authorization", "Bearer "+accessToken)
 	req.Header.Add("Accept", "application/json")
 	resp, err := k.httpClient.Do(req)
 	if err != nil {
@@ -177,5 +170,4 @@ func (k *KeyCloakClient) getFromKeyCloak(accessToken, url string) ([]byte, int, 
 		return b, resp.StatusCode, err
 	}
 	return b, resp.StatusCode, nil
-
 }
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -37,15 +37,10 @@ type KeyCloakClient struct {
 	httpClient *http.Client
 }
 
-func (k *KeyCloakClient) newClient(config *v32.OIDCConfig) (KeyCloakClient, error) {
-	keyCloakClient := KeyCloakClient{}
-	if config.Certificate != "" && config.PrivateKey != "" {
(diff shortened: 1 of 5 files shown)


paynejacob (dismissed):

prachidamle (approved):

pkg/auth/providers/oidc/oidc_client.go
```diff
@@ -22,19 +22,21 @@ func (o *OpenIDCProvider) AddCertKeyToContext(ctx *context.Context, certificate,
...
+			return err
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM([]byte(certificate))
+		httpClient = &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					RootCAs:      caCertPool,
+					Certificates: []tls.Certificate{keyPair},
+				},
```
prachidamle, line 37: We are replacing the httpClient reference by a new one everytime GetClientWithCertKey is called, why do we do this? Also this means you are replacing the k.httpClient reference, this can cause issues if two parallel requests using k.httpClient are made?

cbron: Agree, seems like a one-time setup should happen every time certs are added or changed. Why'd we move away from that ?

brendarearden: I believe I resolved this by only setting the Transport config on the client. Ready for re-review

prachidamle: But it will be set again per request and  httpClient is a common instance. Either we should just create a new httpClient per request or Save the cert&key on the httpClient in the TestAndApply method where you saveOIDCConfig during auth setup.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -170,7 +159,7 @@ func (k *KClient) getFromKeyCloak(accessToken, url string, config *v32.OIDCConfi
 	}
 	req.Header.Add("Authorization", "token "+accessToken)
 	req.Header.Add("Accept", "application/json")
-	resp, err := kHTTPClient.httpClient.Do(req)
+	resp, err := k.httpClient.Do(req)
```
prachidamle, line 162: We should also call defer resp.Body.Close() in here to ensure the http response is closed

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -317,6 +317,7 @@ func (o *OpenIDCProvider) saveOIDCConfig(config *v32.OIDCConfig) error {
 	if err := common.CreateOrUpdateSecrets(o.Secrets, convert.ToString(config.ClientSecret), field, strings.ToLower(config.Type)); err != nil {
 		return err
 	}
+	config.ClientSecret = common.GetName(config.Type, field)
```
cbron, line 320: nit: all of these field vars pointing at different fields looks weird to me, its done that way in other areas and isn't an err, so not mandatory to change, but would be nice if it was `secretField` and `privateKeyField`

pkg/auth/tokens/manager.go
```diff
@@ -630,7 +630,7 @@ var uaBackoff = wait.Backoff{
 
 func (m *Manager) NewLoginToken(userID string, userPrincipal v3.Principal, groupPrincipals []v3.Principal, providerToken string, ttl int64, description string) (v3.Token, string, error) {
 	provider := userPrincipal.Provider
-	if (provider == "github" || provider == "azuread" || provider == "googleoauth") && providerToken != "" {
+	if (provider == "github" || provider == "azuread" || provider == "googleoauth" || provider == "oidc" || provider == "keycloakoidc") && providerToken != "" {
```
cbron, line 633: nit: can you make this commit more descriptive, or add a comment around the code. I've never understood it.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -37,15 +37,10 @@ type KeyCloakClient struct {
...
-	if config.Certificate != "" && config.PrivateKey != "" {
-		err := oidc.GetClientWithCertKey(keyCloakClient.httpClient, config.Certificate, config.PrivateKey)
-		if err != nil {
-			return KeyCloakClient{}, err
-		}
-	}
-	return keyCloakClient, nil
+func newClient(config *v32.OIDCConfig) (*KeyCloakClient, error) {
+	keyCloakClient := &KeyCloakClient{}
+	err := oidc.GetClientWithCertKey(keyCloakClient.httpClient, config.Certificate, config.PrivateKey)
```
cbron, line 42: nit: you could move this func into provider file

cbron (commented):
Makes a lot more sense to me, just a few nits

pkg/auth/tokens/manager.go
brendarearden: I added a comment

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: Moved

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: Changed field var names to `secretField` and `privateKeyField`

pkg/auth/providers/oidc/oidc_client.go
brendarearden: I am not building a httpclient per request and adding the auth just to the transport config as to not override any additional settings.

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -183,4 +170,5 @@ func (k *KeyCloakClient) getFromKeyCloak(accessToken, url string, config *v32.OI
 		return b, resp.StatusCode, err
 	}
 	return b, resp.StatusCode, nil
+
```
prachidamle, line 173: extra newline can be removed

prachidamle (dismissed):
LGTM

pkg/auth/tokens/manager.go
```diff
@@ -630,7 +630,8 @@ var uaBackoff = wait.Backoff{
 
 func (m *Manager) NewLoginToken(userID string, userPrincipal v3.Principal, groupPrincipals []v3.Principal, providerToken string, ttl int64, description string) (v3.Token, string, error) {
 	provider := userPrincipal.Provider
-	if (provider == "github" || provider == "azuread" || provider == "googleoauth") && providerToken != "" {
+	// Providers that use oauth create need to create a secret for storing the access token.
```
cbron, line 633: Was this typo or on purpose ? I don't think `oauth create` is a special thing
```suggestion
	// Providers that use oauth need to create a secret for storing the access token.
```

cbron (approved):

Issue #33053, opened 2021-06-04
Unable to deploy EKS or GKE on hosted Rancher


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
On Rancher v2.5.8 HA setup.
Bring up a hosted Rancher v2.5.8 setup by installing Rancher on an imported k3s cluster (tenant cluster).
Access the hosted Rancher setup as a global admin with full access.
Deploy an EKS or GKE cluster.

**Result:**
Cluster provisioning will fail with: `Failed to deploy eks-operator: can't find system project`

**Other details that may be helpful:**
This is because there is no system project in the local cluster of the tenant cluster. As a result, kev2 operators cannot find the system namespace to deploy into.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.8
- Installation option (single install/HA): HA

aaronRancher:
![Screen Shot 2021-06-04 at 3 36 34 PM](https://user-images.githubusercontent.com/57409052/120869650-c26bdd00-c54b-11eb-9a01-043a3d56af52.png)

aaronRancher:
Thanks @prabalsharma. Are these issues caused by the same problem? Can this be closed as a duplicate?

prachidamle:
@aaronRancher In your case above, was it the global Admin or restrictedAdmin or standard user trying to deploy EKS/GKE?

prachidamle:
This issue does not seem related to restricted admin role because of two tests:
1) EKS provisioning works with restricted admin role on standalone rancher setup using 2.6/master
2) EKS provisioning  is broken for Global admin having all perms as well on v2.5.8 fresh install for hosted Rancher

aaronRancher:
QA to do:
Test EKS, GKE, AKS deployment on hosted Rancher as global admin AND as restricted admin.

thedadams:
2.6 seems to not be affected by this issue (master-head).

aaronRancher:
GKEv2 provisioning on v2.5.10 is currently blocked by: https://github.com/rancher/rancher/issues/33721

aaronRancher:
Tested on Rancher `v2.5-head` `version defd81c` with the following steps:
- Bring up a hosted Rancher setup by installing Rancher on an imported k3s cluster (tenant cluster).
- Log into tenant cluster Rancher.
[...]

Issue #32149, opened 2021-04-15
Global/restricted admin unable to create an EKS cluster in hosted Rancher tenant


dnoland1:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Install Rancher and use helm option "restrictedAdmin" = "true"
2. Create a new EKS cluster - enter AWS creds with permissions to create EKS cluster, can mostly use default options. Click Create.

**Result:**
Cluster fails to provision with error: "Failed to deploy eks-operator: can't find system project "
<img width="1275" alt="image" src="https://user-images.githubusercontent.com/468462/114798450-3824ab00-9d4a-11eb-971a-88b4814ef9b6.png">

**Other details that may be helpful:**
This appears to be a regression. EKS provisioning was failing in a v2.5.0 alpha, but was fixed in the final v2.5.0 release. See https://github.com/rancher/rancher/issues/28764 . Seems to have broke sometime between v2.5.0 and v2.5.7

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
v2.5.7
- Installation option (single install/HA):
HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
K3s
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
2 AWS m5a.large VMs (2vpu 8gb ram)
- Kubernetes version (use `kubectl version`):

```
> kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:23:52Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5+k3s1", GitCommit:"355fff3017b06cde44dbd879408a3a6826fa7125", GitTreeState:"clean", BuildDate:"2021-03-31T06:21:52Z", GoVersion:"go1.15.10", Compiler:"gc", Platform:"linux/amd64"}
```

- Docker version (use `docker version`):
NA, contained

gz#16465

gz#17477

thedadams:
@dnoland1 Are you getting a bunch of "cannot find system project" errors on these clusters in other situations?

dnoland1:
[...]

PR #33109, opened 2021-06-09
Handle System/Default projects for hosted Rancher

thedadams:
If the creatorID annotation was missing from the local cluster object,
then Rancher would skip creating the System and Default projects. After
this change, these projects will get created regardless of the
existence of the creatorID annotation.
    
Also, when `app.EnsureAppProjectName` was called in preparation for
installing an app, it reassigns the namespace to the project if the
namespace annotation is not the expected projectID. This change does not
reassign the cattle-system namespace in such cases. This is to attempt
to keep the cattle-system namespace in the host cluster's tenant System
project.

Issues:
https://github.com/rancher/rancher/issues/33053
https://github.com/rancher/rancher/issues/32149

### DIFF
--- a/pkg/app/utils.go
+++ b/pkg/app/utils.go
@@ -2,15 +2,18 @@ package app
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/rancher/norman/types/slice"
 	"github.com/rancher/rancher/pkg/controllers/managementagent/nslabels"
 	"github.com/rancher/rancher/pkg/controllers/managementuser/helm/common"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	projv3 "github.com/rancher/rancher/pkg/generated/norman/project.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/project"
+	"github.com/rancher/rancher/pkg/settings"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -47,19 +50,21 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 		}
 	}
 
-	// move Namespace into a project
 	expectedAppProjectName := fmt.Sprintf("%s:%s", clusterName, ownedProjectID)
-	appProjectName := ""
-	if projectName, ok := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]; ok {
-		appProjectName = projectName
+	appProjectID := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]
+	if appProjectID != "" && slice.ContainsString(strings.Split(settings.SystemNamespaces.Get(), ","), deployNamespace.Name) {
+		// Don't reassign the system namespaces to another project.
+		return expectedAppProjectName, nil
 	}
-	if appProjectName != expectedAppProjectName {
-		appProjectName = expectedAppProjectName
+
+	// move Namespace into a project
+	if appProjectID != expectedAppProjectName {
+		appProjectID = expectedAppProjectName
 		if deployNamespace.Annotations == nil {
 			deployNamespace.Annotations = make(map[string]string, 2)
 		}
 
-		deployNamespace.Annotations[nslabels.ProjectIDFieldLabel] = appProjectName
+		deployNamespace.Annotations[nslabels.ProjectIDFieldLabel] = appProjectID
 
 		_, err := userNSClient.Update(deployNamespace)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


thedadams:
With this change, the namespaces in the tenant cluster will maintain the projectID annotations of the upstream cluster. Therefore, the namespaces will show up in the System/Default projects in the host cluster and not in the System/Default projects in the tenant cluster.

It should be decided if:
1. This is acceptable; or
2. The project IDs should be overwritten in the tenant cluster; or
3. We find a way to have both project (host and tenant) annotated on the namespace.

thedadams:
During the design meeting, it was determined that the namespaces should stay in the upstream clusters' tenant system project.

pkg/app/utils.go
```diff
@@ -47,13 +47,16 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 		}
 	}
 
-	// move Namespace into a project
 	expectedAppProjectName := fmt.Sprintf("%s:%s", clusterName, ownedProjectID)
-	appProjectName := ""
-	if projectName, ok := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]; ok {
-		appProjectName = projectName
+	appProjectName := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]
+	if deployNamespace.Name == namespace.System && appProjectName != "" {
```
dramich, line 52: This should use the setting to cover all system-namespaces. This only uses the tenant rancher setting though

pkg/app/utils.go
```diff
@@ -47,13 +47,16 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 		}
 	}
 
-	// move Namespace into a project
 	expectedAppProjectName := fmt.Sprintf("%s:%s", clusterName, ownedProjectID)
-	appProjectName := ""
-	if projectName, ok := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]; ok {
-		appProjectName = projectName
+	appProjectName := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]
```
dramich, line 51:
```suggestion
	appProjectId := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]
```

thedadams: Done

pkg/app/utils.go
thedadams: Done

pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -266,14 +266,11 @@ func (m *mgr) createProject(name string, cond condition.Cond, obj runtime.Object
 			return obj, nil
 		}
 
-		creatorID, ok := metaAccessor.GetAnnotations()[creatorIDAnn]
-		if !ok {
-			logrus.Warnf("Cluster %v has no creatorId annotation. Cannot create %s project", metaAccessor.GetName(), name)
-			return obj, nil
-		}
+		annotation := map[string]string{}
```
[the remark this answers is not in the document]
thedadams: Done

thedadams:
https://github.com/rancher/rancher/issues/33417

pkg/app/utils.go
```diff
@@ -47,27 +49,29 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 		}
 	}
 
-	// move Namespace into a project
 	expectedAppProjectName := fmt.Sprintf("%s:%s", clusterName, ownedProjectID)
-	appProjectName := ""
-	if projectName, ok := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]; ok {
-		appProjectName = projectName
+	appProjectID := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]
+	if strings.Contains(settings.SystemNamespaces.Get(), deployNamespace.Name) && appProjectID != "" {
```
dramich, line 54: This check needs to be exact, a contains will have false positives if the namespace is something like `fleet-system` when `cattle-fleet-system` exists in the setting. 
Do the `appProjectID != ""` check first so we don't have to process namespaces if we don't need to.

thedadams: Done.

thedadams:
I increase the timeouts for two tests in `test_default_roles.py` because they were continually failing in arm64 only.

dramich:
> I increase the timeouts for two tests in `test_default_roles.py` because they were continually failing in arm64 only.

2 out of 7 failed from that test, it looks like test_cluster_catalog_templates_access is also an issue. But increasing that timeout is not the fix. If it takes 40 seconds to populate the roles/annotation we have much bigger problems.

tests/integration/suite/test_default_roles.py
```diff
@@ -49,7 +49,7 @@ def test_cluster_create_default_role(admin_mc, cleanup_roles, remove_resource):
     cluster = client.create_cluster(name=random_str())
     remove_resource(cluster)
 
-    wait_for_condition('InitialRolesPopulated', 'True', client, cluster)
+    wait_for_condition('InitialRolesPopulated', 'True', client, cluster, 90)
```
dramich, line 52: Can you drop this change? https://github.com/rancher/rancher/pull/33109#issuecomment-871671074

thedadams: Done.

dramich (approved):

kinarashah (approved):

Issue #32939, opened 2021-05-28
Reduce size of v3.App objects


nickgerace:
Parent issue: https://github.com/rancher/rancher/issues/32418

nickgerace:
Related `kubernetes/kubernetes` upstream issue: https://github.com/rancher/rancher/issues/32419#issuecomment-858136129

nickgerace:
An update on this issue after speaking to @dramich: this issue is still important since the App objects are oversized regardless of the upstream fix ([kubernetes/kubernetes#102749](https://github.com/kubernetes/kubernetes/issues/102749)).

We do not know how large users' apps are across the board. Specifically, our test payloads may not reflect the size of our users' largest apps. If we have objects that are that massive (nearing ~1.3MB) this issue will still persist due to `etcd` constraints, regardless of the upstream problem.

PR #33129, opened 2021-06-10
Compress valuesYaml for v3.App

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/32939

### DIFF
--- a/pkg/api/norman/customization/app/app.go
+++ b/pkg/api/norman/customization/app/app.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
 	v32 "github.com/rancher/rancher/pkg/apis/project.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/app/compression"
 	"github.com/rancher/rancher/pkg/catalog/manager"
 	clusterv3 "github.com/rancher/rancher/pkg/client/generated/cluster/v3"
 	projectv3 "github.com/rancher/rancher/pkg/client/generated/project/v3"
@@ -168,7 +169,11 @@ func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiConte
 			}
 		}
 		if valuesYaml != nil {
-			obj.Spec.ValuesYaml = convert.ToString(valuesYaml)
+			// We convert to a byte slice since we know that the object is YAML string.
+			obj.Spec.ValuesYaml, err = compression.CompressValuesYaml([]byte(convert.ToString(valuesYaml)))
+			if err != nil {
+				return err
+			}
 		} else {
 			obj.Spec.ValuesYaml = ""
 		}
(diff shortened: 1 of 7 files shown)


pkg/api/norman/customization/app/app.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/values"
 	v32 "github.com/rancher/rancher/pkg/apis/project.cattle.io/v3"
+	appPkg "github.com/rancher/rancher/pkg/app"
```
nickgerace, line 15: Need to match this to match `app2` below.

pkg/client/generated/project/v3/zz_generated_app.go
```diff
@@ -43,6 +44,7 @@ type App struct {
 	Answers              map[string]string `json:"answers,omitempty" yaml:"answers,omitempty"`
 	AppRevisionID        string            `json:"appRevisionId,omitempty" yaml:"appRevisionId,omitempty"`
 	AppliedFiles         map[string]string `json:"appliedFiles,omitempty" yaml:"appliedFiles,omitempty"`
+	CompressedValuesYaml string            `json:"compressedValuesYaml,omitempty" yaml:"compressedValuesYaml,omitempty"`
```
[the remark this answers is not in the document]
nickgerace: No longer relevant

pkg/controllers/managementuserlegacy/helm/common/common.go
```diff
@@ -215,8 +216,12 @@ func GenerateAnswerSetValues(app *v3.App, tempDir *HelmPath, extraArgs map[strin
 	// a user-supplied values file will overridden default values.yaml
 	if app.Spec.ValuesYaml != "" {
 		custom := "custom-values.yaml"
-		valuesYaml := filepath.Join(tempDir.FullPath, custom)
-		if err := ioutil.WriteFile(valuesYaml, []byte(app.Spec.ValuesYaml), 0755); err != nil {
+		valuesYamlFileName := filepath.Join(tempDir.FullPath, custom)
+		valuesYaml, err := compression.DecompressValuesYaml(app.Spec.ValuesYaml)
```
StrongMonkey, line 220: The only concern I have is to make sure that `compression.DecompressValuesYaml` should work against both compressed and uncompressed content, as it will break for exiting app that uses uncompressed values.yaml. I guess it is already the case since you check for magic header, just to double check.

StrongMonkey: Or are we migrating existing app.Spec.ValuesYaml into compressed content?

nickgerace: I spoke with @dramich, and we want to use `string` instead of `[]byte` for the values. That was an intentional change since the last time we looked it this.

Since we are using the same field, we need to check if it's actually compressed with GZIP, as you mentioned. That's what the byte slice check is for.

I believe the current code covers that scenario. Do you see any room for improvement?

StrongMonkey (dismissed):
LGTM, Might want to call out that this change won't immediately mitigate the issue until user upgrade their app. I understand that it is probably better to keep the original app untouched in order to introduce less chaos.

jiaqiluo (dismissed):
LGTM

dramich:
@nickgerace Can you provide an example before and after of how the app object looks and also how much size this saves on a large app like gitlab?

nickgerace:
Closing due to upstream fix (https://github.com/kubernetes-sigs/structured-merge-diff/pull/200) and unable to reproduce with GitLab charts as well as user-provided charts.

Issue #32920, opened 2021-05-27
Dev follow ups on keycloak


cbron:
Follow up on: https://github.com/rancher/rancher/issues/30280
For light technical changes.

brendarearden:
Follow up needed: https://github.com/rancher/rancher/blob/dc2dc5744bc0c94a49c24175b064dd96b24c02ef/pkg/auth/providers/keycloakoidc/keycloak_client.go#L178 << this could result in a 200 status code, but a nil body. Need to add a check for error that is returned on this line.

brendarearden:
Follow up needed: 
https://github.com/rancher/rancher/blob/dc2dc5744bc0c94a49c24175b064dd96b24c02ef/pkg/auth/providers/keycloakoidc/keycloak_client.go#L85 << ensure the error is returned for a 401 instead of eating the error

brendarearden:
Follow up needed: 
the keycloak token isn't being refreshed on timeout
```2021/06/15 08:34:38 [ERROR] Error refreshing token principals, skipping: [generic oidc]: not implemented      
(manually logged out & in)```

cbron:
Closing this as it will be tested via the other 2 keycloak issues

PR #33130, opened 2021-06-10
[keycloak oidc] Add nil check, return errors

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/32920

**Problem**
` b, err := ioutil.ReadAll(resp.Body)` does not check for nil error after it
401 errors are being returned nil stead of just returning the error after `k.getFromKeyCloak`

**Solution**
Add nil check 
Return all errors if they exists after `k.getFromKeyCloak` is called.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -48,12 +48,10 @@ func (k *KeyCloakClient) searchPrincipals(searchTerm, principalType string, acce
 		search := URLEncoded(searchURL)
 
 		b, statusCode, err := k.getFromKeyCloak(accessToken, search)
-		if statusCode == 401 {
-			return nil, nil
-		}
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -82,12 +79,9 @@ func (k *KeyCloakClient) searchPrincipals(searchTerm, principalType string, acce
 		search := URLEncoded(searchURL)
 
 		b, statusCode, err := k.getFromKeyCloak(accessToken, search, config)
-		if statusCode == 401 {
-			return nil, nil
-		}
 		if err != nil {
-			logrus.Errorf("[keycloak oidc]: GET url %v received error from github, err: %v", search, err)
-			return accounts, err
+			return accounts, fmt.Errorf("[keycloak oidc]: GET request failed, got status code: %d. url: %s, err: %s",
```
dramich, line 83: Error returns should not be formatted like a log message

prachidamle (dismissed):

dramich (dismissed):

dramich (approved):

Issue #31209, opened 2021-02-09
Rancher v2.5 fleet and rancher-operator namespaces created outside any project


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Deploy Rancher v2.5.5 into an RKE cluster
2. View Projects/Namespaces

**Result:**

Observe cluster-fleet-local-local-*, fleet-clusters-system, fleet-default, fleet-local and rancher-operator-system Namespaces that exist outside of any project, per the screenshot below. Can these be created within the System project by default?

![Screenshot from 2021-02-09 16-50-10](https://user-images.githubusercontent.com/9771530/107393438-18c79280-6afb-11eb-816b-8991e6addc19.png)

**Environment information**
- Rancher v2.5.5

kinarashah:
Links to refer: 
List of system namespaces:  https://github.com/rancher/rancher/blob/master/pkg/settings/setting.go#L63 
Controller adding system annotation to namespace on create: https://github.com/rancher/rancher/blob/master/pkg/controllers/managementuser/rbac/namespace_handler.go#L62

thehejik:
The same thing happens in rancher `2.5.6-rc5`

sowmyav27:
**Reproduced the issue on 2.5.5**

- Deploy a Rancher HA setup with RKE local cluster.
- These namespaces are not in any project
<img width="1284" alt="Screen Shot 2021-06-15 at 1 02 57 PM" src="https://user-images.githubusercontent.com/26032343/122116150-214a1580-cdda-11eb-9036-71252ffe7795.png">

**On master-head commit id: `2b50005ae `**
- Except for fleet-system, all other namespaces have been moved under System Project
<img width="1291" alt="Screen Shot 2021-06-15 at 1 04 18 PM" src="https://user-images.githubusercontent.com/26032343/122116490-8867ca00-cdda-11eb-86c8-7ee58cabd830.png">

sowmyav27:
Blocked because of https://github.com/rancher/rancher/issues/33203

StrongMonkey:
Should be ready to test since #33203 is fixed and ready to test too.

sowmyav27:
Will validate after this issue is fixed - https://github.com/rancher/rancher/issues/33677

sowmyav27:
Waiting for this issue to be fixed - https://github.com/rancher/rancher/issues/33203

[...]

PR #33146, opened 2021-06-11
Add fleet namespaces to system project

StrongMonkey:
https://github.com/rancher/rancher/issues/31209

### DIFF
--- a/tests/integration/suite/test_system_project.py
+++ b/tests/integration/suite/test_system_project.py
@@ -54,7 +54,7 @@ def test_system_namespaces_assigned(admin_cc):
     if loggingNamespace in system_namespaces_names:
         system_namespaces_names.remove(loggingNamespace)
 
-    assert system_namespaces_names == initial_system_namespaces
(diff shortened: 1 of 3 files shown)


kinarashah (approved):
Discussed offline, we can't add these directly to the system-namespaces setting because we want to move all fleet namespaces based on labels so can't hardcode them.

dramich (approved):

tests/integration/suite/test_system_project.py
```diff
@@ -53,8 +53,9 @@ def test_system_namespaces_assigned(admin_cc):
     # will be present in current system_namespaces, removing it
     if loggingNamespace in system_namespaces_names:
         system_namespaces_names.remove(loggingNamespace)
-
-    assert system_namespaces_names == initial_system_namespaces
+    print(initial_system_namespaces)
```
dramich, line 56: Should drop the print statement, the assert can be extended to include a string with the contents of the lists.

dramich (approved):

Issue #31336, opened 2021-02-16
500 error on /v3/kontainerdrivers/rancher-images as a result of non-existent /var/lib/rancher/management-state/catalog-cache directory


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Provision a cluster with RKE and install Rancher via helm
2. Navigate to https://<rancher_url>/v3/kontainerdrivers/rancher-images and refresh the page many times

**Result:**

Observe many requests fail with the following error:

```
{
"baseType": "error",
"code": "ServerError",
"message": "error getting image list for linux platform",
"status": 500,
"type": "error"
}
```

Rancher pod logs show errors of the following format:

```
2021/02/16 09:16:44 [ERROR] API error response 500 for GET /v3/kontainerdrivers/rancher-images. Cause: failed to fetch images from system charts: failed to get chart and version from "management-state/catalog-cache/df6f64f73bb427a3c62ced202ca09db9cf9f2547296217da13a2eb43d5cf5c08": stat management-state/catalog-cache/df6f64f73bb427a3c62ced202ca09db9cf9f2547296217da13a2eb43d5cf5c08: no such file or directory
```

**Other details that may be helpful:**

`/var/lib/rancher/management-state/catalog-cache` appears to exist only on the Rancher leader Pod. As a result requests for `/v3/kontainerdrivers/rancher-images` only succeeded when served by the leader Pod.

**Environment information**
- Reproduced with Rancher v2.4.4, v2.4.13 and v2.5.5

gz#15049

paynejacob:
The image list is generated at request time by looking at the catalog `index.yaml` files and then crawling the chart `values.yml` one by one.  This depends on the rancher pod cloning all the chat repositories locally.  In an HA environment only the leader rancher pod clones these repositories.  When the ingress round robins requests they can end up on pods that do not have the repositories cloned causing a 5XX error.  

Proposed Solutions:
- **Ensure this request is only served by the leader pod.** This could be done by splitting the rancher service into leader and non leader pods and having the ingress route some paths to the leader service.  This solution is somewhat fragile given the list of paths that can only be served by the leader pod would need to be maintained.
- **Clone the charts repositories onto each pod.**  This solution would be the easiest to implement however it will generate significantly more traffic to our charts repositories and could cause issues with charts data getting stale, causing the images endpoint to return different content based on what pod serves the request.
[...]

PR #33151, opened 2021-06-11
Fix/issue 31336 (image list 500)

paynejacob:
https://github.com/rancher/rancher/issues/31336

#  Problem
When users access `/v3/kontainerdrivers/rancher-images` in HA clusters the will receive a 500 if the request is served from any pod other than the master rancher pod.  This endpoint generates a list of images the user needs by crawling for images in the system catalog.  The system catalog is only available on the master pod.

# Solution
When we sync the system catalog we generate the image list and store it in a configuration map.  Then when any pod gets the request for the image list it can look it up in the configuration map. To do this i had to break out catalog/helm and catalog/git to their own package to avoid a cyclic dependency between the image and catalog package.

# Testing
1. install ha rancher on a new cluster
2. `GET /v3/kontainerdrivers/rancher-images` once for every rancher pod in the cluster
3. observe the request never fails.


pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -3,17 +3,18 @@ package kontainerdriver
...
 	"net/http"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/rancher/rancher/pkg/catalog/utils"
+	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
StrongMonkey, line 13: Nit: empty space

pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -154,30 +156,40 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
 		rkeSysImages[k8sVersion] = *rkeSysImgCopy
 	}
 
-	// get system charts path
-	systemCatalog, err := lh.CatalogLister.Get("", utils.SystemLibraryName)
+	var catalogImageList *v1.ConfigMap
+	catalogImageList, err = lh.ConfigMapLister.Get(utils.GetCatalogImageCacheName(utils.SystemLibraryName), metav1.GetOptions{})
 	if err != nil {
-		return httperror.WrapAPIError(err, httperror.ServerError, "error getting system catalog")
+		return
```
StrongMonkey, line 162: We should probably still return wrapped error in order for better messaging?

pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -154,30 +156,40 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
...
-	systemCatalogHash := helmlib.CatalogSHA256Hash(systemCatalog)
-	systemCatalogChartPath := filepath.Join(helmlib.CatalogCache, systemCatalogHash)
 
-	var targetImages []string
+	var targetRkeSysImages []string
+	var targetSysCatalogImages []string
 	switch apiContext.ID {
 	case linuxImages:
-		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", nil, []string{}, rkeSysImages, image.Linux)
+		targetRkeSysImages, _, err = image.GetImages("", "", nil, []string{}, rkeSysImages, image.Linux)
```
StrongMonkey, line 169: Will targetRkeSysImages have anything since you were passing an empty path to `image.GetImages`? If it is empty I guess we don't have to deal with it.

pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -154,30 +156,40 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
...
+		targetSysCatalogImages, _, err = image.ParseCatalogImageListConfigMap(catalogImageList)
+		if err != nil {
+			return httperror.WrapAPIError(err, httperror.ServerError, "error getting image list for linux platform")
+		}
 	}
 
+	var targetImages []string
 	agentImage := settings.AgentImage.Get()
 	targetImages = append(targetImages, img.Mirror(agentImage))
+	targetImages = append(targetImages, targetRkeSysImages...)
```
StrongMonkey, line 191: Same, not sure if targetRkeSysImages will have anything on it since empty path is passed.

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,13 +47,54 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 	}
 	logrus.Debugf("Chart hash comparison for global catalog %v: new -- %v --- current -- %v", catalog.Name, commit, catalog.Status.Commit)
 
+	// ensure the system catalog image cache exists
+	var systemCatalogImageCache *v1.ConfigMap
+	if catalog.Name == utils.SystemLibraryName {
+		systemCatalogImageCacheName := utils.GetCatalogImageCacheName(catalog.Name)
+		systemCatalogImageCache, err = m.ConfigMapLister.Get(systemCatalogImageCacheName, metav1.GetOptions{})
+		if err != nil {
+			if apiErr, ok := err.(*errors.StatusError); !ok || apiErr.Status().Code != 404 {
```
StrongMonkey, line 56: you can use errors.isNotFound() from `k8s.io/apimachinery/pkg/api/errors`

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,13 +47,54 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+				return nil, err
+			}
+		}
+
+		// if the cache does not exist generate it
+		if systemCatalogImageCache.UID == "" {
```
StrongMonkey, line 62: Better to check if error is not found then create, in case systemCatalogImageCache is a nil pointer

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,13 +47,54 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+			systemCatalogImageCache = &v1.ConfigMap{}
+			systemCatalogImageCache.Name = systemCatalogImageCacheName
```
StrongMonkey, line 64: Hmm, do you specify namespace when creating it? I think it should go to `cattle-system` namespace.

pkg/api/norman/server/managementstored/setup.go
```diff
@@ -780,8 +780,9 @@ func KontainerDriver(schemas *types.Schemas, management *config.ScaledContext) {
 		MetadataHandler:       metadataHandler,
 	}
 	lh := kontainerdriver.ListHandler{
-		SysImageLister: management.Management.RkeK8sSystemImages("").Controller().Lister(),
-		SysImages:      management.Management.RkeK8sSystemImages(""),
+		SysImageLister:  management.Management.RkeK8sSystemImages("").Controller().Lister(),
+		SysImages:       management.Management.RkeK8sSystemImages(""),
+		ConfigMapLister: management.Core.ConfigMaps("cattle-system"),
```
StrongMonkey, line 785: This is not a lister. Lister is the cache lister which is `management.Core.ConfigMaps("").Controller().Lister()`. Also you might want to create a client in order to CRUD configmaps.

```
ConfigMap: management.Core.ConfigMaps("")
```

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,13 +47,54 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+			err = image.CreateCatalogImageListConfigMap(systemCatalogImageCache, catalog)
+			if err != nil {
+				return nil, err
+			}
+
+			systemCatalogImageCache, err = m.ConfigMapLister.Create(systemCatalogImageCache)
```
StrongMonkey, line 70: Use client to create configmap object

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,13 +47,54 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+
 	if isUpToDate(commit, catalog) {
 		if setRefreshed(catalog) {
 			m.catalogClient.Update(catalog)
 		}
 		return nil, nil
 	}
 
+	// update the system catalog image cache if it is out of date
+	if catalog.Name == utils.SystemLibraryName {
```
StrongMonkey, line 85: You should combine this logic into the previous if block. So 
1. Check if configmap exists, if not, create configmap with catalog image list.
2. If configmap exists, check if catalog is update to date. If up to date, do nothing. Otherwise retrieve catalog image list and update the configmap.

pkg/catalog/manager/manager.go
```diff
@@ -6,16 +6,18 @@ import (
 	"strings"
 	"time"
 
+	corev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+
```
StrongMonkey, line 10: Nit: empty space

pkg/catalog/manager/manager.go
```diff
@@ -69,6 +72,7 @@ func New(management managementv3.Interface, project projectv3.Interface) *Manage
 		ClusterCatalogLister:  management.ClusterCatalogs("").Controller().Lister(),
 		appRevisionClient:     project.AppRevisions(""),
 		bundledMode:           bundledMode,
+		ConfigMapLister:       core.ConfigMaps("cattle-system"),
```
StrongMonkey, line 75: Same above as this is a client not a clister. Make sure you have client for CUD and lister for R as reading from cache is fast and less load on api server and etcd.

pkg/git/git.go
```diff
@@ -77,8 +77,21 @@ func FormatURL(pathURL, username, password string) string {
...
 
 func CloneWithDepth(path, url, branch string, depth int) error {
-	if err := catUtil.ValidateURL(url); err != nil {
+	if err := ValidateURL(url); err != nil {
 		return err
 	}
 	return runcmd("git", "clone", "-b", branch, "--single-branch", fmt.Sprintf("--depth=%v", depth), url, path)
 }
+
+var (
```
StrongMonkey, line 86: It is better to put var at the begining of the file, right below import.

pkg/image/resolve_export_test.go
```diff
@@ -0,0 +1,10 @@
+package image
+
+// export private functions for testing only
+
+var FetchImagesFromCharts = fetchImagesFromCharts
```
StrongMonkey, line 5: Hmm is this needed? If your test file is on the same package with actual file, I think you can use private functions too.

pkg/helm/setup.go
```diff
@@ -12,8 +12,9 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/rancher/rancher/pkg/git"
```
StrongMonkey, line 15: nit: empty space

pkg/image/resolve.go
```diff
@@ -9,12 +9,14 @@ import (
 	"sort"
 	"strings"
 
+	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
```
StrongMonkey, line 12: empty space

pkg/helm/helm.go
```diff
@@ -18,10 +18,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rancher/rancher/pkg/git"
```
StrongMonkey, line 21: empty space

pkg/image/resolve_export_test.go
paynejacob: The problem I ran into is reslove_test.go depends on kdm creating a cyclic import.  If the tests exist in image_test you cannot import private functions.

pkg/api/norman/customization/kontainerdriver/actionhandler.go
paynejacob: Yes it will have all the rkeSysImages and all of the mirrored rkeSysImages.  We only cache the chart images in the configmap not the system images.

pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -154,30 +156,40 @@ func (lh ListHandler) LinkHandler(apiContext *types.APIContext, next types.Reque
...
 		if err != nil {
 			return httperror.WrapAPIError(err, httperror.ServerError, "error getting image list for linux platform")
 		}
 	case windowsImages:
-		targetImages, _, err = image.GetImages(systemCatalogChartPath, "", nil, []string{}, rkeSysImages, image.Windows)
+		targetRkeSysImages, _, err = image.GetImages("", "", nil, []string{}, rkeSysImages, image.Windows)
 		if err != nil {
 			return httperror.WrapAPIError(err, httperror.ServerError, "error getting image list for windows platform")
 		}
+		targetSysCatalogImages, _, err = image.ParseCatalogImageListConfigMap(catalogImageList)
```
kinarashah, line 182: Let's move out targetSysCatalogImages cuz it's not os dependent

pkg/image/resolve.go
```diff
@@ -341,3 +352,27 @@ func getSourcesList(imageSources map[string]bool) string {
...
+	var windowsImages []string
+	var linuxImages []string
+
+	catalogHash := libhelm.CatalogSHA256Hash(catalog)
+	catalogChartPath := filepath.Join(libhelm.CatalogCache, catalogHash)
+
+	windowsImages, _, err = GetImages(catalogChartPath, "", nil, []string{}, nil, Windows)
+	linuxImages, _, err = GetImages(catalogChartPath, "", nil, []string{}, nil, Linux)
+
+	cm.Data = make(map[string]string, 2)
```
kinarashah, line 366: don't proceed here if err != nil

pkg/image/resolve.go
```diff
@@ -341,3 +352,27 @@ func getSourcesList(imageSources map[string]bool) string {
...
+	cm.Data[osTypeImageListName[Windows]] = strings.Join(windowsImages, imageListDelimiter)
+	cm.Data[osTypeImageListName[Linux]] = strings.Join(linuxImages, imageListDelimiter)
+
+	return
+}
+
+func ParseCatalogImageListConfigMap(cm *v1.ConfigMap) (windowsImageList, linuxImageList []string, err error) {
```
kinarashah, line 373: where's err here?

pkg/image/resolve.go
```diff
@@ -28,11 +29,21 @@ type chart struct {
 	version string
 }
 
+const imageListDelimiter = "\n"
 const (
 	Linux OSType = iota
 	Windows
 )
 
+var osTypeImageListName map[OSType]string
```
kinarashah, line 38: could you put the values in here instead of init?

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,6 +47,45 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+
+		// if the cache exists and is out of date update it
+		if !isUpToDate(commit, catalog) {
+			err = image.CreateCatalogImageListConfigMap(systemCatalogImageCache, catalog)
+			if err != nil {
+				return nil, err
+			}
+
+			// update
+			_, err = m.ConfigMap.Update(systemCatalogImageCache)
```
kinarashah, line 82: hmmm wouldn't we want to do this after the `traverseAndUpdate` is actually updated? would there be issues if you update the list first but then actual catalog refresh errors out?

pkg/catalog/manager/catalog_sync.go
```diff
@@ -43,6 +47,45 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+			systemCatalogImageCache = &v1.ConfigMap{}
+			systemCatalogImageCache.Name = systemCatalogImageCacheName
+			systemCatalogImageCache.Namespace = namespace.System
+			err = image.CreateCatalogImageListConfigMap(systemCatalogImageCache, catalog)
+			if err != nil {
+				return nil, err
+			}
+
+			systemCatalogImageCache, err = m.ConfigMap.Create(systemCatalogImageCache)
+			if err != nil {
```
kinarashah, line 67: we're listing from Lister so that might give NotFound, but the configmap might exist and Create will fail with AlreadyExists  - add check for `err != nil && !errors.IsAlreadyExists`

kinarashah (approved):

superseb (approved):

kinarashah (approved):

PR #33159, opened 2021-06-12
Remove use of awk

dirkmueller:
It appears that only two places use awk, and their uses are trivial to
replace with grep or cut. this would allow to drop awk from the image.

### DIFF
--- a/package/run.sh
+++ b/package/run.sh
@@ -38,7 +38,7 @@ get_address()
         echo ""
     # If given address is a network interface on the system, retrieve configured IP on that interface (only the first configured IP is taken)
     elif [ -n "$(find /sys/devices -name $address)" ]; then
-        echo $(ip addr show dev $address | grep -w inet | awk '{print $2}' | cut -f1 -d/ | head -1)
+        echo $(ip addr show dev $address | grep -oP "(?<=inet ).+(?=\/)" | head -1)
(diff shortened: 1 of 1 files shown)


snasovich:
@dirkmueller , what is the rationale behind removing `awk` use? I believe `awk` is widely used and extracting field data in whitespace-separated output reads easier when `awk` is used over more complex regex with `grep`.
I know `awk` doesn't come standard with SLE BCI base image, but it's a quick install to get it added.

dirkmueller:
@snasovich simply from the principle of "less (code) is more". the first occurrence combined grep, awk and cut to extract a single field from a line of text. `grep` can do that directly, saving awk and cut. Yes, it does use a bit of a complex regular expression, but I don't believe many people will dive into the detailed line very often, readers stop at "ah this gets the ip address". 

the 2nd invocation was just splitting, and `cut` can do that as well. fewer code, fewer dependencies, smaller containers.

ibuildthecloud:
I'd rather just leave the awk usage as it's quite straightforward for anyone to understand or maintain the script.  Honestly I think awk should be in BCI by default.  I personally would expect sed, grep, awk to be in any Linux environment by default.

dirkmueller:
Great, closing this then.

PR #33165, opened 2021-06-13
[2.5] Fix rancher agent docker auth config when using private registry

rawmind0:
Issue #27306
Issue #30628
Issue #33128

Required by PR https://github.com/rancher/rancher/pull/33166

When using private registry with auth, Rancher is creating/updating the `cattle-system/cattle-private-registry` secret containing `.dockerconfigjson`,  to be used by rancher cluster and node agents to be able to pull images. This secret is not working properly due to `.dockerconfigjson` is generated in a bad format. This issue was "hided" due to rancher cluster and node agents can't pull images by themselves config, but at the end they are able to use the cluster level private registry config.

This PR updates the `GeneratePrivateRegistryDockerConfig` function to use proper k8s secret `.dockerconfigjson` format so rancher cluster and node agents are able to pull docker image  by themselves config

Issue #33128, opened 2021-06-10
Kubernetes components are not restarted after updating cloud provider config file


sadhugit:
**RKE version:**
rancher/rke-tools:v0.1.72
**Docker version: (`docker version`,`docker info` preferred)**
Client:
 Debug Mode: false
Server:
 Containers: 7
  Running: 5
  Paused: 0
  Stopped: 2
 Images: 10
 Server Version: 19.03.15
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 269548fa27e0089a8b8278fc4fc781d7f65a939b
 runc version: ff819c7e9184c13b7c2607fe6c30ae19403a7aff
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 3.10.0-1062.12.1.el7.x86_64
 Operating System: CentOS Linux 7 (Core)
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.51GiB

 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
**Operating system and kernel: (`cat /etc/os-release`, `uname -r` preferred)**
Linux uk-ias-kublw06 3.10.0-1062.12.1.el7.x86_64 rancher/rke#1 SMP Tue Feb 4 23:02:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

**Steps to Reproduce:**
1. Configure azure cluster with cloud provider option
2. Change the azure account password
3. Upgrade the K8 version from 1.18.15 -> 1.18.16
4. Upgrade the rancher from 2.4.13 to 2.5.17 and cluster was active after the upgrade.

**Results:**
[...]

PR #33166, opened 2021-06-13
[2.5] Wait for rancher agent daemonset rollout before update cluster

rawmind0:
Issue https://github.com/rancher/rancher/issues/27306
Issue https://github.com/rancher/rancher/issues/30628
Issue https://github.com/rancher/rancher/issues/33128

Requires PR https://github.com/rancher/rancher/pull/33165

The root cause of these issues seems to be related with the way that rancher agents are updated. Eventually, the rancher agent becomes updated in the middle of a node plan execution, making that the node plan execution aborts and the rancher agent is never restarted. 

This PR add the option to check that the `cattle-node-agent` daemonset has been properly updated and rollout, before continue with the cluster update. This is assuring that the rancher agent is updated and running before the cluster provision to avoid its updated during a node plan execution. This also add some reliability to the cluster update process due to if the new rancher agent config is not working properly the cluster will not be updated and still ready and working

Also, added 2 new rancher setting to allow the config if cluster update should wait for rancher agent daemonset:
* `agent-rollout-wait`: Wait for agent rollout. Default: "true" 
* `agent-rollout-timeout`: Timeout waiting for agent rollout. Default: "300s" 

With the PR applied, the cluster update behaviour can be changed based on `agent-rollout-wait` setting:
* `agent-rollout-wait="false"`: Current behaviour
  cluster updated by user -> agent deployment NO wait for rollout (client tunnel) -> cluster update (rke up - eventually, agent rollout occurs in the middle of this execution causing the issue) -> cluster active/error
* `agent-rollout-wait="true"`: Proposed behaviour
  cluster updated by user -> agent deployment wait for rollout (client tunnel) -> cluster update (rke up - agent is already rollout) -> cluster active
  On tests done on this scenario, the cluster update seems safer and "easier" to recover in that way. When update a cluster, if a new agent config is not working, the agent rollout is timeout and the cluster update is stopped. The cluster became on error on ui, but the k8s is perfectly working due to has not been updated. To recover the state, editing the cluster undoing the last changes and saving, makes that the agent is redeployed with previous working config.

### DIFF
--- a/pkg/systemtemplate/template.go
+++ b/pkg/systemtemplate/template.go
@@ -218,6 +218,11 @@ spec:
         secret:
           secretName: cattle-credentials-{{.TokenKey}}
           defaultMode: 320
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0
+      maxSurge: 1
 {{ if .IsRKE }}
 
 ---
@@ -315,7 +320,7 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 25%
+      maxUnavailable: 50%
 
 {{- end }}
 
@@ -413,7 +418,7 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 25%
+      maxUnavailable: 50%
 {{- end }}
 
 {{- if .AuthImage}}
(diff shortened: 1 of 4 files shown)


pkg/systemtemplate/template.go
```diff
@@ -315,7 +320,7 @@ spec:
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 25%
+      maxUnavailable: 50%
```
superseb, line 323: This is to speed up deployment I assume?

pkg/systemtemplate/template.go
```diff
@@ -218,6 +218,11 @@ spec:
         secret:
           secretName: cattle-credentials-{{.TokenKey}}
           defaultMode: 320
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0
+      maxSurge: 1
```
superseb, line 225: This is `25%` by default and rounding down, so effectively we are allowing one additional pod here with this change, right?

pkg/systemtemplate/template.go
rawmind0: Yes, this is just to accelerate the rollout

superseb (approved):

paynejacob (approved):

Issue #33175, opened 2021-06-14
Include output from jailer.sh to log if it fails


snasovich:
**What kind of request is this (question/bug/enhancement/feature request):** Enhancement

**Other details that may be helpful:**
In case `jailer.sh` script fails during Rancher startup, script's output is not included in the log, only FATAL message like `[FATAL] error running the jail command: exit status 1` is produced making it hard to pinpoint what exactly caused this script to fail.
Currently `jailer.sh` script output is added at DEBUG level only when if it succeeds.
It's suggested to log script output in all cases and also switch it to TRACE level to prevent any sensitive information from appearing in logs (sensitive information is acceptable at TRACE level).
Note it's currently not trivial to observe current behavior as `jailer.sh` doesn't fail. It was discovered as part of working on https://github.com/rancher/rancher/issues/33127.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head (cdc12d8a5cc6fda069709a950f9a6dc9e38892e1).

snasovich:
This is now on `master-head`.
@sowmyav27 , what is our approach on testing issues that are not reproducible unless code is modified for it to fail? The only option I see is to produce a custom build with modified `jailer.sh` that fails so when this version is deployed you would see script output is logged. This does sound like overkill for something with such trivial fix (see linked PR).

sowmyav27:
Waiting for info from @snasovich to provide info on how to reproduce/test this

snasovich:
@kravciak , the idea for testing this is to replace `jailer.sh` file with one that fails before running Rancher.

Here is a script that can be used to test it:

```sh
#!/bin/bash
set -e

IMAGE=$1

if [[ $IMAGE = "" ]]
  then
    echo "One argument (image name with tag) must be passed to the script"
    exit 1
fi

CONTAINER_ID=$(docker create --restart=unless-stopped --privileged --pull=always $IMAGE --trace=true)

printf "#!/bin/bash\n\ncp /usr/path-doesnt-exist /usr" > jailer.sh
docker cp jailer.sh $CONTAINER_ID:/usr/bin/jailer.sh

docker start $CONTAINER_ID

docker logs $CONTAINER_ID -f

docker rm $CONTAINER_ID -f
```

If running it passing `rancher/rancher:master-head`, you can see there is TRACE log message with `jailer.sh` script output that allows understanding how exactly the script has failed.

[...]

Issue #33127, opened 2021-06-10
Update Rancher base image to SLE BCI 


snasovich:
**What kind of request is this (question/bug/enhancement/feature request):** Enhancement

**Other details that may be helpful:**

We are moving to using the same base image across all of our images. SUSE has developed an image called SLE BCI (Base Container Image) that we will need to roll out to all images within Rancher. This issue request this image to be used for Rancher.

snasovich:
QA, to test this before we merge it, could you please run validation tests on the following images built from linked PR:
[snasovichsuse/rancher:67c52b45b](https://hub.docker.com/layers/snasovichsuse/rancher/67c52b45b/images/sha256-753c9194e843e410300aa0254795f8ee59fcbdb77afa946347d48a48a32341d4?context=explore)
[snasovichsuse/rancher-agent:67c52b45b](https://hub.docker.com/layers/153530633/snasovichsuse/rancher-agent/67c52b45b/images/sha256-c7fe9e0a43b5ccdbdc2f84fe6e13aab9f465ecb583f31516316ddb4a5745d6c1?context=explore)

snasovich:
@izaac , just tried smoke test using Docker install and it's failing to come up. Please hold on.

snasovich:
Related PR: https://github.com/rancher/rancher/pull/33007

snasovich:
Have a fix for immediate "startup failed" issue, so Docker container build on this image at least starts OK. Here are the updated images:
[snasovichsuse/rancher:dce011aed](https://hub.docker.com/layers/153614202/snasovichsuse/rancher/dce011aed/images/sha256-55477cba9ff9aca01257175ee4b58feea3fa15473978d65057af215f7750a478?context=explore)
[snasovichsuse/rancher-agent:dce011aed](https://hub.docker.com/layers/153614686/snasovichsuse/rancher-agent/dce011aed/images/sha256-b52b00955c6e6beb319af46cf890356ba698f562f9c8845d3cc67065ff620bd1?context=explore)

@izaac , could you please try running tests on these images?

@deniseschannon , JFYI.

izaac:
@snasovich I am getting this error while deploying an RKE1 DO cluster from the Rancher using this image

[...]

PR #33176, opened 2021-06-14
Include output from jailer.sh to log if it fails

snasovich:
`jailer.sh` script output is now logged not only when the script succeeds and is now at TRACE level instead of DEBUG (to prevent potentially sensitive information from being logged).

### DIFF
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -63,16 +63,16 @@ func CreateJail(name string) error {
 
 	cmd := exec.CommandContext(ctx, "/usr/bin/jailer.sh", name)
 	out, err := cmd.CombinedOutput()
+	if len(out) > 0 {
+		logrus.Tracef("CreateJail: output from jail script for [%s]: [%v]", name, string(out))
+	} else {
+		logrus.Tracef("CreateJail: no output from jail script for [%s]", name)
+	}
 	if err != nil {
 		if strings.HasSuffix(err.Error(), "signal: killed") {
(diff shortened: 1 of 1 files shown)


snasovich:
Linked issue: #33175

Note it's currently not trivial to observe previous behavior as `jailer.sh` doesn't fail. It was discovered as part of working on https://github.com/rancher/rancher/issues/33127

superseb:
Isn't it so that not seeing the output means there is no output (length check)? Also, what was the output you needed from the non-error situation that we need to print it? I would expect to print it when `err != nil`? All and all, it doesn't matter that much but logging is not free in the end.

superseb (commented):
^^

snasovich:
> Isn't it so that not seeing the output means there is no output (length check)? Also, what was the output you needed from the non-error situation that we need to print it? I would expect to print it when `err != nil`? All and all, it doesn't matter that much but logging is not free in the end.

@superseb , in my particular case it was `cp` command that failed due to source file not existing. With "before" code, only something like `[FATAL]: error running the jail command: exit status 1` was logged with no information on root cause of the issue. `err` returned from `out, err := cmd.CombinedOutput()` doesn't contain script output; it's `out` that contains combined `stderr` and `stdout` in this case.
Since we were already logging it in case of success, I believe logging it in case of failure (which results in Rancher failing to start) makes sense as logging overhead is not a concern if application is failing to start.

superseb (approved):

Issue #33080, opened 2021-06-07
[vsphere] Add support for secretIds to the vsphere meta apis


codyrancher:
The rke2 cloud credentials are now just naked secrets. In order for the UI to perform rke2 vsphere deployments we now need to support passing secretIds as well as cloudCredentialIds(for RKE1) to all of the vsphere meta apis i.e. `/meta/vsphere/data-centers?cloudCredentialId=cattle-global-data:cc-275sb`.

This is necessary in order for the UI to retrieve all of the relevant vsphere data in order to pre-populate the relevant form fields and to maintain parity with the old UI.

luthermonson:
since this needs to be backwards compatible for rke1/cluster manager UI we will use a new get param of `secretId=<secretname>`

slickwarren:
on master-head (b6bcfad):
* provisioned rke1 cluster using vsphere provider -> pass
* create credentials + node template for rke2 vsphere node provider -> pass
  * was unable to provision cluster, [but issue seems unrelated](https://github.com/rancher/rancher/issues/33078) 
* double checked with dev that no backport is needed for this ticket (related to new UI, not in 2.5.x)

PR #33204, opened 2021-06-15
vSphere Meta API update to support a secret instead of the old v3 CloudCredential

luthermonson:
PR allows for a new secretId param which is the name of the secret the new provisioning rke2 cloud credential is stored in and check if the user has access to it and continue on to use the cloud cred data to access the vsphere api.

https://github.com/rancher/rancher/issues/33080


phillipsj (approved):
LGTM

dramich (approved):

PR #33208, opened 2021-06-15
Add lock to uninstall to avoid launching multiple uninstall pod

StrongMonkey:
https://github.com/rancher/fleet/issues/419

During the `cattle-fleet-system` namespace migration we need to run uninstall to make sure the previous release in old namespace is removed. The uninstall operation can't be run in parrallel due to the issue above.

### DIFF
--- a/pkg/controllers/dashboard/fleetcharts/controller.go
+++ b/pkg/controllers/dashboard/fleetcharts/controller.go
@@ -57,9 +57,12 @@ func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error
 		return setting, nil
 	}
 
+	h.Lock()
 	if err := h.manager.Uninstall(fleetUninstallChart.ReleaseNamespace, fleetUninstallChart.ChartName); err != nil {
+		h.Unlock()
 		return nil, err
 	}
+	h.Unlock()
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/dashboard/fleetcharts/controller.go
```diff
@@ -57,9 +57,11 @@ func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error
 		return setting, nil
 	}
 
+	h.Lock()
 	if err := h.manager.Uninstall(fleetUninstallChart.ReleaseNamespace, fleetUninstallChart.ChartName); err != nil {
 		return nil, err
 	}
+	h.Unlock()
```
kinarashah, line 64: would it better defer so we release the lock in case of error? or is it intentional to keep retrying for the same pod until it succeeds?

StrongMonkey: I think it only needs to lock the uninstall place.... if using defer then we are locking the whole method and lock will only be released after the controller sync, which is not what we want?

kinarashah: I see, adding Unlock() before returning error then? I think it'll be blocked otherwise

StrongMonkey: :+1:

kinarashah (approved):

dramich (approved):

PR #33209, opened 2021-06-16
Check RBAC status of an AKS cluster

thedadams:
It is possible to create an AKS cluster with RBAC disabled. In order to
display a warning message to users when RBAC is disabled, it status of
RBAC is put on the status of the cluster.

Also, the aksListClusters proxy endpoint also includes whether the RBAC
is enabled for each cluster. This would allow a warning message to be
displayed when registering AKS clusters.

Issue:
https://github.com/rancher/rancher/issues/27670


cmurphy (approved):

cbron (approved):

mjura (approved):
LGTM

PR #33232, opened 2021-06-16
Add rancher cluster IP to generated internal certs

StrongMonkey:
Need to add cluster IP to internal cert SAN since we are switching to use cluster IP in the [kubeconfig](https://github.com/rancher/rancher/commit/60275b1f66d2e2c0ae40a9edaa7ca874ac2c23b6#diff-a670104fda3becbf85fd10edd3[…]d40a10b690e2f02fd36a7c23R38-R48). This is due to the change in rancherd that @ibuildthecloud made because it can't resolve cluster DNS on the host for system-agent.

https://github.com/rancher/fleet/issues/420

### DIFF
--- a/pkg/tls/tls.go
+++ b/pkg/tls/tls.go
@@ -101,6 +101,15 @@ func ListenAndServe(ctx context.Context, restConfig *rest.Config, handler http.H
 		CertNamespace: namespace.System,
 		CertName:      "tls-rancher-internal",
 	}
+	clusterIP, err := getClusterIP(core.Core().V1().Service())
+	if err != nil {
+		return err
+	}
+	if clusterIP != "" {
+		serverOptions.TLSListenerConfig = dynamiclistener.Config{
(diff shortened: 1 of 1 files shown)


pkg/tls/tls.go
```diff
@@ -286,6 +295,20 @@ func readConfig(secrets corev1controllers.SecretController, acmeDomains []string
...
 
+func getClusterIP(services corev1controllers.ServiceController) (string, error) {
+	service, err := services.Get(namespace.System, "rancher", metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", nil
+		}
+		return "", err
+	}
+	if service.Spec.ClusterIP == "" {
```
StrongMonkey, line 306: This will cause rancher to restart if service IP is not assigned. But it should be rare since service are always created and IP should be ready when rancher is started.

Issue #33203, opened 2021-06-15
namespaces not in use on the latest rancher are not cleaned up and are created outside any project in local cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
### Upgrade scenario

- Deployed an RKE HA 2.5.8 setup, local cluster is RKE
- Deployed a downstream cluster.
- Upgrade Rancher to master-head commit id: `2b50005`
- These namespaces are seen outside any project in local RKE cluster

<img width="1286" alt="Screen Shot 2021-06-15 at 1 42 42 PM" src="https://user-images.githubusercontent.com/26032343/122120818-bc91b980-cddf-11eb-8394-b0dc0a75aee2.png">

### On a fresh install of master-head commit id: `2b50005`
`fleet-system` namespace is seen which is not in any project
<img width="1291" alt="Screen Shot 2021-06-15 at 1 04 18 PM" src="https://user-images.githubusercontent.com/26032343/122121677-c5cf5600-cde0-11eb-84e7-1040f71a72ae.png">

**Expected Result:**
The namespaces not in use must be cleaned up

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `2b50005`
- Installation option (single install/HA): HA/single

StrongMonkey:
Rancher-operator-system clean up is fixed here https://github.com/rancher/rancher/pull/33212. For fresh-install with fleet-system, there are some limitation but once https://github.com/rancher/rancher/pull/32528 is merged I can address that also.

StrongMonkey:
@sowmyav27 In master `Rancher-operator-system` should be gone after upgrade. I will try to fix the fleet system later now.

sowmyav27:
**On an upgrade from 2.5.8 too master-head commit id: `def25f1`**

- rancher-operator-system namespace is removed
- fleet-system namespace is still seen outside any Project. 
- Other namespaces moved to cattle-system
<img width="1243" alt="Screen Shot 2021-06-16 at 10 32 28 PM" src="https://user-images.githubusercontent.com/26032343/122337148-c7893e80-cef2-11eb-9745-13804fd8c98f.png">

Reopening for fixes fleet-system namespace

StrongMonkey:
[...]

Issue #32560, opened 2021-05-05
catalog refresh does not happen on an upgraded rancher setup for over 10 minutes


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- On 2.5.7 deployed a couple of downstream clusters. (my case it was couple of windows clusters 3 linux and 3 windows nodes)
- upgraded rancher to 2.5.8-rc9
- Fleet agent did not upgrade (it was over 10 minutes)
- Fleet agent was still in 0.3.4

**Other details that may be helpful:**
After a manual refresh of the chart repo, upgraded the fleet agents on both the clusters

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.8-rc19
- Installation option (single install/HA): Single

Jono-SUSE-Rancher:
@sirredbeard - Can you double check and make sure this has nothing to do with the fact that Sowmya was upgrading with the windows clusters/windows nodes? I'm not sure how windows behaves around the fleet agent.

StrongMonkey:
https://github.com/rancher/rancher/pull/33234 This could potentially fix the timing issue.

sowmyav27:
**On 2.5.8 to master-head upgrade - commit id: `a11f947`**

The new fleet-agent is deployed only after about 15 minutes after the upgrade. The old fleet agent is removed meanwhile (logged this issue too track it)

But reopening as it takes about 15 minutes for the new fleet-agent `rancher/fleet-agent:v0.3.6-rc2` to get deployed

cmurphy:
I'm not seeing this issue, @sowmyav27 is this still reproducible? Here is what I tried:

1. Installed a 2.5.8 cluster from rancher-stable with helm:

```
$ helm install rancher rancher-stable/rancher --set hostname=429604a2fec0.ngrok.io --namespace=cattle-system --version=v2.5.8
```

2. Registered a downstream "Other cluster"
3. fleet-agent 0.3.5 is installed in the fleet-system namespace
![image](https://user-images.githubusercontent.com/1220710/125887671-64f4d034-4f54-4b7e-bcc6-7555d8da0dce.png)

4. Built rancher chart from 1b81a0615ddef614e41c8ca53ccb309c447ac350 and upgraded to it:

```
[...]

PR #33234, opened 2021-06-16
Add adopt and change fleet-crd to cattle-fleet-system

StrongMonkey:
https://github.com/rancher/rancher/issues/32560
https://github.com/rancher/rancher/issues/33203

Add force-adopt flag so that it can adopt crd resource that were previously installed in `fleet-system`. This can help migrating fleet-crd chart from `fleet-system` to `cattle-fleet-system`.

### DIFF
--- a/pkg/controllers/dashboard/fleetcharts/controller.go
+++ b/pkg/controllers/dashboard/fleetcharts/controller.go
@@ -10,11 +10,13 @@ import (
 	"github.com/rancher/rancher/pkg/features"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/wrangler"
+	"github.com/rancher/wrangler/pkg/relatedresource"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 var (
 	fleetCRDChart = chartDef{
-		ReleaseNamespace: "fleet-system",
+		ReleaseNamespace: "cattle-fleet-system",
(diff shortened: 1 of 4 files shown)


pkg/controllers/dashboard/fleetcharts/controller.go
```diff
@@ -38,6 +40,15 @@ func Register(ctx context.Context, wContext *wrangler.Context) error {
 	}
 
 	wContext.Mgmt.Setting().OnChange(ctx, "fleet-install", h.onSetting)
+	// watch cluster repo `rancher-charts` and enqueue the setting to make sure the latest fleet is installed after catalog refresh
```
StrongMonkey, line 43: This makes sure that if rancher-charts has changed, we immediately enqueue the setting which triggers fleet chart to to be installed. This makes sure fleet always get the latest version as soon as chart refresh happens.

paynejacob (approved):

dramich:
@StrongMonkey What is using the force adopt flag? Is that something that exists in the helm pod?

StrongMonkey:
@dramich It is a new flag I added to adopt resources outside of helm/helm releases. https://github.com/rancher/helm/commit/f06a6766c5279e7ddcabd8519eb821af263c1cbb. We have to add this to adopt CRD resources that were previous created in a different namespace in order for the migration.

dramich (approved):

Issue #32721, opened 2021-05-13
Disabling multi cluster management breaks rancher -- 404 error when trying to access


slickwarren:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
* deploy rancher 
* go to `global settings` -> `feature flags`
* disable the `multi-cluster-management` flag

**Result:**
rancher gives an error and does not seem to be recoverable

**Other details that may be helpful:**
related to testing this ticket https://github.com/rancher/rancher/issues/32631#issue-879894838

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head (351f7ca04973c1f023653f171c0f621f31bf2ca2)
- Installation option (single install/HA): single

aaronRancher:
https://github.com/rancher/rancher/pull/32798

kinarashah:
https://github.com/rancher/rancher/pull/32949

sowmyav27:
**On master-head commit id: `be03c8b`**

- Deploy a Rancher HA
- from Settings - Features Flag - MCM -- DeActive
- The Rancher server restarts
- User is able to login through the new UI - <rancher server>/dashboard
- When User navigates to Cluster Manager page from Side panel, the user sees a `404 page not found` error. And user must manually `Go back` or must change the page's URL to navigate back to Dashboard login page.

deniseschannon:
MCM feature flag can only be turned off when starting Rancher and not through the settings inside of Rancher. 

MCM feature flag will only be visible in the UI if it is off. Once on, it will not be shown an ability to turn off.

slickwarren:
Tested on master-head (d44698d):
* deploy rancher with MCM Disabled -- deploys successfully 
  * cluster management is disabled
  * ![Screen Shot 2021-06-23 at 5 07 37 PM](https://user-images.githubusercontent.com/16691014/123184689-09f9d080-d449-11eb-8fda-6ace08335559.png)
  * local cluster is working as expected
  * user is able to enable MCM through the UI
![Screen Shot 2021-06-23 at 5 07 30 PM](https://user-images.githubusercontent.com/16691014/123184766-331a6100-d449-11eb-9b90-a62e16bce1e6.png)
[...]

PR #33243, opened 2021-06-17
Create admin user when mcm is disabled and assign cluster-admin

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/32721

PR #33265, opened 2021-06-17
Fix aggregation and UI access in downstream clusters

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/33266

Issue #33126, opened 2021-06-10
Add 2.6 Audit log enhancements to Kubectl access path.


cloudnautique:
The enhancements added in https://github.com/rancher/rancher/issues/31166  need to also support the kubectl access path. When a user uses a kubeconfig file that passes through Rancher the username and IDP username need to be logged in the Rancher audit logs at level 1.  Additionally, they should be added into the kube-apiserver audit logs as well.

prachidamle:
This fix is to support audit logging from kubectl access which use kubeconfig tokens. In case of kubeconfig, the AuthProvider is always set to "local", the userPrincipal is not found and is missing information. Hence the audit log data is not seen.
As a fix, the kubeconfig token is now fixed to carry the correct authProvider going forward.

Another part of the fix is instead reading extraInfo from user's login token, this is now read from userattributes where the extra info is saved after user logs in. This information can be found using the userId extracted from any token.
The userAttributes.Extra is also saved to ClusterUserAttributes.Extra field.

The process of finding the extra attributes for the user will be:

Check if extra attributes are found in userattributes for the token's provider
If the token's provider is empty or is local, then add all extraAttributes into the Audit entry to cover all IDP values
Fallback is read from token if found, or otherwise read from the user object

prachidamle:
Here are some Audit Log samples


1) Rancher API audit Logs

Before fix, using kubeconfig tokens  having "local" set for admin user even if rancher has OpenLDAP enabled
Notice that the "extra" attributes are logged empty

[...]

PR #33268, opened 2021-06-17
Audit log enhancements for Kubectl access path

prachidamle:
https://github.com/rancher/rancher/issues/33126

This fix is to support audit logging from kubectl access which use kubeconfig tokens. In case of kubeconfig, the AuthProvider is always set to "local", the userPrincipal is not found and is missing information. Hence the audit log data is not seen. 
As a fix, the kubeconfig token is now fixed to carry the correct authProvider going forward.

Another part of the fix is instead reading extraInfo from user's login token, this is now read from userattributes where the extra info is saved after user logs in. This information can be found using the userId extracted from any token.
The userAttributes.Extra is also saved to ClusterUserAttributes.Extra field.

The process of finding the extra attributes for the user will be:
- Check if  extra attributes are found in userattributes for the token's provider
- If the token's provider is empty or is local, then add all extraAttributes into the Audit entry to cover all IDP values 
- Fallback is read from token if found, or otherwise read from the user object

### DIFF
--- a/pkg/apis/management.cattle.io/v3/authn_types.go
+++ b/pkg/apis/management.cattle.io/v3/authn_types.go
@@ -92,7 +92,7 @@ type UserAttribute struct {
 	GroupPrincipals map[string]Principals // the value is a []Principal, but code generator cannot handle slice as a value
 	LastRefresh     string
 	NeedsRefresh    bool
-	Extra           map[string]map[string][]string
+	ExtraByProvider map[string]map[string][]string // extra information for the user to print in audit logs, stored per authProvider. example: map[openldap:map[principalid:[openldap_user://uid=testuser1,ou=dev,dc=us-west-2,dc=compute,dc=internal]]]
 }
 
 type Principals struct {
--- a/pkg/apis/management.cattle.io/v3/authn_types.go
+++ b/pkg/apis/management.cattle.io/v3/authn_types.go
@@ -92,6 +92,7 @@ type UserAttribute struct {
 	GroupPrincipals map[string]Principals // the value is a []Principal, but code generator cannot handle slice as a value
 	LastRefresh     string
 	NeedsRefresh    bool
+	Extra           map[string]map[string][]string
 }
 
 type Principals struct {
--- a/pkg/auth/providers/activedirectory/activedirectory_provider.go
+++ b/pkg/auth/providers/activedirectory/activedirectory_provider.go
@@ -229,9 +229,9 @@ func (p *adProvider) getDNAndScopeFromPrincipalID(principalID string) (string, s
 	return externalID, scope, nil
 }
 
-func (p *adProvider) GetUserExtraAttributes(token *v3.Token) map[string][]string {
+func (p *adProvider) GetUserExtraAttributes(userPrincipal v3.Principal) map[string][]string {
 	extras := make(map[string][]string)
-	extras["principalid"] = []string{token.UserPrincipal.Name}
-	extras["username"] = []string{token.UserPrincipal.LoginName}
+	extras["principalid"] = []string{userPrincipal.Name}
+	extras["username"] = []string{userPrincipal.LoginName}
 	return extras
 }
--- a/pkg/auth/providers/activedirectory/activedirectory_provider.go
+++ b/pkg/auth/providers/activedirectory/activedirectory_provider.go
@@ -231,7 +231,11 @@ func (p *adProvider) getDNAndScopeFromPrincipalID(principalID string) (string, s
 
 func (p *adProvider) GetUserExtraAttributes(userPrincipal v3.Principal) map[string][]string {
 	extras := make(map[string][]string)
-	extras["principalid"] = []string{userPrincipal.Name}
-	extras["username"] = []string{userPrincipal.LoginName}
+	if userPrincipal.Name != "" {
+		extras[common.UserAttributePrincipalID] = []string{userPrincipal.Name}
+	}
+	if userPrincipal.LoginName != "" {
+		extras[common.UserAttributeUserName] = []string{userPrincipal.LoginName}
+	}
 	return extras
 }
--- a/pkg/auth/requests/authenticate.go
+++ b/pkg/auth/requests/authenticate.go
@@ -110,6 +110,7 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
 	if token.ClusterName != "" && token.ClusterName != a.clusterRouter(req) {
 		return nil, errors.Wrapf(ErrMustAuthenticate, "clusterID does not match")
 	}
+	logrus.Infof("request token %v", token)
 
 	attribs, err := a.userAttributeLister.Get("", token.UserID)
 	if err != nil && !apierrors.IsNotFound(err) {
@@ -147,9 +148,7 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
 			groups = append(groups, name)
 		}
 	}
-
 	groups = append(groups, user.AllAuthenticated, "system:cattle:authenticated")
-
 	if !strings.HasPrefix(token.UserID, "system:") {
 		go a.userAuthRefresher.TriggerUserRefresh(token.UserID, false)
 	}
@@ -158,11 +157,30 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
 	authResp.User = token.UserID
 	authResp.UserPrincipal = token.UserPrincipal.Name
 	authResp.Groups = groups
-	logrus.Debugf("Auth filter calling provider to get extra info %v", token.AuthProvider)
 
-	extraFromProvider := providers.GetUserExtraAttributes(token.AuthProvider, token)
-	if extraFromProvider != nil {
-		authResp.Extras = extraFromProvider
+	if attribs != nil {
+		if strings.EqualFold(token.AuthProvider, "local") || token.AuthProvider == "" {
+			//gather all extraInfo for all external auth providers present in the userAttributes
+			extraInfo := make(map[string][]string)
+			for provider, extra := range attribs.Extra {
+				if provider != "local" {
+					for key, value := range extra {
+						extraInfo[key] = append(extraInfo[key], value...)
+					}
+				}
+			}
+			authResp.Extras = extraInfo
+		} else {
+			//non-local authProvider is set in token
+			var ok bool
+			authResp.Extras, ok = attribs.Extra[token.AuthProvider]
+			if !ok {
+				extraFromProvider := providers.GetUserExtraAttributes(token.AuthProvider, token.UserPrincipal)
+				if extraFromProvider != nil {
+					authResp.Extras = extraFromProvider
+				}
+			}
+		}
 	}
 	return authResp, nil
 }
--- a/pkg/auth/requests/authenticate.go
+++ b/pkg/auth/requests/authenticate.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/rancher/pkg/auth/providerrefresh"
 	"github.com/rancher/rancher/pkg/auth/providers"
+	"github.com/rancher/rancher/pkg/auth/providers/common"
 	"github.com/rancher/rancher/pkg/auth/tokens"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config"
@@ -110,7 +111,6 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
 	if token.ClusterName != "" && token.ClusterName != a.clusterRouter(req) {
 		return nil, errors.Wrapf(ErrMustAuthenticate, "clusterID does not match")
 	}
-	logrus.Infof("request token %v", token)
 
 	attribs, err := a.userAttributeLister.Get("", token.UserID)
 	if err != nil && !apierrors.IsNotFound(err) {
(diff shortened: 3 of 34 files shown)


pkg/auth/requests/authenticate.go
```diff
@@ -158,11 +157,30 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
...
 	authResp.UserPrincipal = token.UserPrincipal.Name
 	authResp.Groups = groups
-	logrus.Debugf("Auth filter calling provider to get extra info %v", token.AuthProvider)
 
-	extraFromProvider := providers.GetUserExtraAttributes(token.AuthProvider, token)
-	if extraFromProvider != nil {
-		authResp.Extras = extraFromProvider
+	if attribs != nil {
+		if strings.EqualFold(token.AuthProvider, "local") || token.AuthProvider == "" {
+			//gather all extraInfo for all external auth providers present in the userAttributes
```
prachidamle, line 163: For kubectl access via kubeconfig token, the auth provider is always set as "local" in that token. So we gather extraInfo of all external authProviders and log that to audit-log. The user info for "local" auth is already set in the audit logging, what we desire is having the external IDP (IdentityProvider) values. For UI access, the session tokens always have the correct authProvider configured, so only that provider's extra-info will be passed down.

prachidamle:
Discussed with @dramich and @aiyengar2 and following changes/alternatives were identified:
1) This solution is saving extraInfo from user's login token into userAttribute CR and reading it back by using the token's userId. But we need to additionally lookup principals of the user from the token or from userId if extraInfo is not found in userAttribute  - this is necessary to take care of the upgrade case where old tokens/old userAttribute will not have the extraInfo
2) We also should ideally fix kubeconfig token to carry the real authProvider and not set "local"; this will let us use the correct principalId of that provider instead of us adding a list of all principalIds to audit log

pkg/apis/management.cattle.io/v3/authn_types.go
```diff
@@ -92,6 +92,7 @@ type UserAttribute struct {
 	GroupPrincipals map[string]Principals // the value is a []Principal, but code generator cannot handle slice as a value
 	LastRefresh     string
 	NeedsRefresh    bool
+	Extra           map[string]map[string][]string
```
snasovich, line 95: Wondering if `Extra` name with long type like `map[string]map[string][]string` is too generic for something that looks to me like extra fields per provider? Just looking at field name and type it's not clear that the key of the map is the authn provider name.
It was much more clear when type was `map[string][]string` but addition of extra layer makes it more obscure I think.
Or are there limitations for adding more types here?

pkg/auth/providers/activedirectory/activedirectory_provider.go
```diff
@@ -229,9 +229,13 @@ func (p *adProvider) getDNAndScopeFromPrincipalID(principalID string) (string, s
 	return externalID, scope, nil
 }
 
-func (p *adProvider) GetUserExtraAttributes(token *v3.Token) map[string][]string {
+func (p *adProvider) GetUserExtraAttributes(userPrincipal v3.Principal) map[string][]string {
 	extras := make(map[string][]string)
-	extras["principalid"] = []string{token.UserPrincipal.Name}
-	extras["username"] = []string{token.UserPrincipal.LoginName}
+	if userPrincipal.Name != "" {
```
snasovich, line 234: These checks don't seem to be related to the issue addressed. Is it just an extra check added to not add empty data or there is more subtle reason?
Is it pretty common to have one of or both of `userPrincipal.Name`/`userPrincipal.LoginName` empty?

pkg/auth/providers/activedirectory/activedirectory_provider.go
```diff
@@ -229,9 +229,13 @@ func (p *adProvider) getDNAndScopeFromPrincipalID(principalID string) (string, s
...
+		extras["principalid"] = []string{userPrincipal.Name}
```
snasovich, line 235: Somewhat of a nitpick, but it could be a good idea to introduce consts for repeating string literals like `"principalid"` and `"username"`.

pkg/auth/requests/authenticate.go
```diff
@@ -158,15 +156,48 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
 	authResp.User = token.UserID
 	authResp.UserPrincipal = token.UserPrincipal.Name
 	authResp.Groups = groups
-	logrus.Debugf("Auth filter calling provider to get extra info %v", token.AuthProvider)
+	authResp.Extras = getUserExtraInfo(token, u, attribs)
+	logrus.Infof("Extras returned %v", authResp.Extras)
```
snasovich, line 160: This sounds more like Debug level message rather than Info. Same for the other `Infof` below.

pkg/controllers/managementuser/clusterauthtoken/user_attribute.go
```diff
@@ -68,5 +69,13 @@ func compareUserAttributeClusterUserAttribute(userAttribute managementv3.UserAtt
 		lastRefresh:  clusterUserAttribute.LastRefresh,
 		needsRefresh: clusterUserAttribute.NeedsRefresh,
 	}
-	return groups, reflect.DeepEqual(current, old)
+	if !reflect.DeepEqual(current, old) {
+		return groups, false
+	}
+
+	if !reflect.DeepEqual(userAttribute.Extra, clusterUserAttribute.Extra) {
```
snasovich, line 76: Could we just add `Extra` to `userAttributeCompare` instead? I believe `reflect.DeepEqual` handles maps OK.

pkg/apis/management.cattle.io/v3/authn_types.go
prachidamle: @snasovich that map[string][]string is as defined by Kubernetes struct to carry userInfo https://pkg.go.dev/k8s.io/apiserver@v0.20.2/pkg/authentication/user#DefaultInfo

And we need to carry this per authProvider hence another map of provider -> defaultInfo so that we do not have to translate it from some custom type when sending it downstream

pkg/auth/providers/activedirectory/activedirectory_provider.go
prachidamle: yes that data could be empty, especially for kubeconfig tokens. This is to add keys only with non-empty values so that we can detect if a fallback on reading the missing data is needed in here https://github.com/rancher/rancher/pull/33268/files#diff-ccde1c52dd5ce3e2e7e9d8c62b6a7a60eb7f80d86942e367c19586fce6b70c63R191-R196

pkg/auth/providers/activedirectory/activedirectory_provider.go
prachidamle: sure

pkg/apis/management.cattle.io/v3/authn_types.go
snasovich: @prachidamle , thank you for the explanation. `map[string][]string` makes sense since it's what referenced in k8s struct as you linked. I still think the name `Extra` doesn't make it obvious what is used as the key, so maybe we can rename this field to something like `ExtraByProvider` or `ProviderExtra` - provided we can map it to Extra of type `map[string][]string` that is actually logged as "extra" field on audit log entry (I'm not sure what type is actually used to produce audit log entries)?

pkg/user/manager.go
```diff
@@ -6,16 +6,31 @@ import (
...
+	TokenName    string
+	Description  string
+	Kind         string
+	UserName     string
+	AuthProvider string
+	Ttl          *int64
+	Randomize    bool
+}
+
+type ClusterTokenInput struct {
```
dramich, line 19: I think we should drop this type and have a clusterName param, seems cleaner vs wrapping the object above when needing to call `EnsureClusterToken`

pkg/user/manager.go
```diff
@@ -6,16 +6,31 @@ import (
...
+	ClusterName string
+}
+
 type Manager interface {
 	SetPrincipalOnCurrentUser(apiContext *types.APIContext, principal v3.Principal) (*v3.User, error)
 	GetUser(apiContext *types.APIContext) string
-	EnsureToken(tokenName, description, kind, userName string, ttl *int64, randomize bool) (string, error)
-	EnsureClusterToken(clusterName, tokenName, description, kind, userName string, ttl *int64, randomize bool) (string, error)
+	EnsureToken(input TokenInput) (string, error)
+	EnsureClusterToken(input ClusterTokenInput) (string, error)
```
dramich, line 28:
```suggestion
	EnsureClusterToken(clusterName string, input ClusterTokenInput) (string, error)
```

pkg/user/manager.go
```diff
@@ -6,16 +6,31 @@ import (
 	apitypes "k8s.io/apimachinery/pkg/types"
 )
 
+type TokenInput struct {
```
dramich, line 9: Can we move this to `pkg/auth/providers/common/usermanager.go`? This file is just for the interface as there used to be cyclical imports

pkg/auth/requests/authenticate.go
```diff
@@ -158,15 +157,47 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
...
+	if attribs != nil && attribs.Extra != nil && len(attribs.Extra) != 0 {
+		if token.AuthProvider == "" {
+			//gather all extraInfo for all external auth providers present in the userAttributes
+			for _, extra := range attribs.Extra {
+				for key, value := range extra {
+					extraInfo[key] = append(extraInfo[key], value...)
+				}
+			}
+		} else {
+			//authProvider is set in token
```
dramich, line 179: How does this handle upgrades where all tokens were set to `local`?

pkg/auth/requests/authenticate.go
```diff
@@ -158,15 +157,47 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
...
+			var ok bool
+			extraInfo, ok = attribs.Extra[token.AuthProvider]
```
dramich, line 181:
```suggestion
			if extraInfo, ok := attribs.Extra[token.AuthProvider]; ok {
			    return extraInfo
			}
```

pkg/auth/requests/authenticate.go
```diff
@@ -158,15 +157,47 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
...
+	extraInfo := make(map[string][]string)
+
+	if attribs != nil && attribs.Extra != nil && len(attribs.Extra) != 0 {
+		if token.AuthProvider == "" {
+			//gather all extraInfo for all external auth providers present in the userAttributes
+			for _, extra := range attribs.Extra {
+				for key, value := range extra {
+					extraInfo[key] = append(extraInfo[key], value...)
+				}
+			}
```
dramich, line 177: Can return here after the loop since this never sets readFromToken

pkg/auth/requests/authenticate.go
```diff
@@ -158,15 +157,47 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
...
 
-	extraFromProvider := providers.GetUserExtraAttributes(token.AuthProvider, token)
-	if extraFromProvider != nil {
-		authResp.Extras = extraFromProvider
-	}
 	return authResp, nil
 }
 
+func getUserExtraInfo(token *v3.Token, u *v3.User, attribs *v3.UserAttribute) map[string][]string {
+	var readFromToken bool
```
dramich, line 167: This won't be needed with the returns

pkg/auth/requests/authenticate.go
```diff
@@ -158,15 +157,47 @@ func (a *tokenAuthenticator) Authenticate(req *http.Request) (*AuthenticatorResp
...
+		readFromToken = true
+	}
+
+	if readFromToken {
+		extraInfo = providers.GetUserExtraAttributes(token.AuthProvider, token.UserPrincipal)
+		//if principalid is not set in extra, read from user
+		if extraInfo != nil && len(extraInfo[common.UserAttributePrincipalId]) == 0 {
+			extraInfo[common.UserAttributePrincipalId] = u.PrincipalIDs
+		}
+	}
```
dramich, line 196: This should end up being fallback logic and not need the if check

pkg/apis/management.cattle.io/v3/authn_types.go
dramich: I agree with changing the name as the other fields called `Extra` are for an individual provider, `ExtraByProvider` sounds fine. Can a comment be added as well explaining the field? That type in general is intimidating so having an explanation and example would help.

prachidamle: Done

pkg/user/manager.go
prachidamle: Tried that but it still creates a cyclic import
package github.com/rancher/rancher
	imports github.com/rancher/rancher/pkg/data/management
	imports github.com/rancher/rancher/pkg/agent/clean
	imports github.com/rancher/rancher/pkg/controllers/management/auth
	imports github.com/rancher/rancher/pkg/auth/providerrefresh
	imports github.com/rancher/rancher/pkg/auth/providers
	imports github.com/rancher/rancher/pkg/auth/api/secrets
	imports github.com/rancher/rancher/pkg/auth/providers/common
	imports github.com/rancher/rancher/pkg/auth/tokens
	imports github.com/rancher/rancher/pkg/types/config
	imports github.com/rancher/rancher/pkg/user
	imports github.com/rancher/rancher/pkg/auth/providers/common: import cycle not allowed

pkg/auth/requests/authenticate.go
prachidamle: So yeah I think the original idea where if "local" we write down all provider's  info, that will handle the case where all kubeconfig tokens were set to "local". Also for this feature, external IDP value is the goal, not the local values. But we can set all. Thoughts?

pkg/controllers/managementuser/clusterauthtoken/user_attribute.go
```diff
@@ -62,11 +64,17 @@ func compareUserAttributeClusterUserAttribute(userAttribute managementv3.UserAtt
...
 		lastRefresh:  clusterUserAttribute.LastRefresh,
 		needsRefresh: clusterUserAttribute.NeedsRefresh,
+		extra:        clusterUserAttribute.Extra,
 	}
-	return groups, reflect.DeepEqual(current, old)
+	if !reflect.DeepEqual(current, old) {
+		return groups, false
+	}
+
+	return groups, true
```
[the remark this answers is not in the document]
snasovich: @prachidamle , it doesn't seem to be resolved yet.

snasovich (approved):
There is only one trivial comment that doesn't seem to be resolved yet, but otherwise LGTM. Still need @dramich 's approval.

pkg/controllers/managementuser/clusterauthtoken/user_attribute.go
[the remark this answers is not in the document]
prachidamle: @snasovich what do you mean by old code?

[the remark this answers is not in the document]
snasovich: I mean the one before the change. Existing `master` has one line instead of these 5:
```go
return groups, reflect.DeepEqual(current, old)
```

snasovich:
@prachidamle , even if this is approved, please do not merge it to `master` yet as it was decided this should not get into 2.6.0  release. So, until branches are created for release/v2.6 (and likely release/v2.6) we should't put it in `master`.

prachidamle:
@snasovich I have reopened this against release/v2.6, added you as well fro review since you had reviewed and approved it earlier.
Closing this one.

PR #33271, opened 2021-06-18
Complete refetch group principals method

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/33036
https://github.com/rancher/rancher/issues/32920

**Problem**
Access token is being passed as search type resulting in `invalid character 'B'` error
Refetch group principals was not implemented

**Solution** 
Pass in the correct values to `getFromKeyCloakByID` so that the auth token for the request is correctly
Refactor out the new provider and get user info code from the `LoginUser` so it can be reused to refetch the group values


paynejacob (approved):

dramich (approved):

prachidamle (approved):
LGTM

Issue #25487, opened 2020-02-19
Listing many large resources can be slow on the ember UI


virtuman:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug/ possibly enhancement 

**Steps to reproduce (least amount of steps as possible):**
Deploy 30-50 helm releases to one project. 
And a few revisions for each. With every release or iteration , rancher becomes slower and slower and at some point I am no longer able to access cluster via web UI at all, it just shows loading icon and kicks you back out to cluster list page 

**Result:**

**Other details that may be helpful:**
Helm 3 stores state in kubernetes secret. Looks like very large secrets, looks like rancher is trying to pull on all kubernetes secrets when I try to access UI. 

It should be better to exclude helm release secret types, or at least not to include soft of secret and type of helm3 secrets in cluster pollings. 

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.3.5
- Installation option (single install/HA): HA
Rancher is deployed on bare metal k8s cluster and another deployment is ingke

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

gz#11894

gz#11629

cruikshj:
We are experiencing the same issue...  we have mitigated it somewhat by setting the `--history-max` parameter on our `helm upgrade` command to 1 (default is 10).

This is not a fix... just delays the inevitable.

Isakgicu:
For me this happen only when adding public repositories to catalogs with support of helm 3(helm 2 works fine). If add a private repository with helm 3, everything is just fine. Hope this info will help you fix this issue ASAP.

yshch:
This issue makes UI almost unusable in any cluster with large amount of secrets. It doesn't have to be helm releases, any significant amount of large secrets makes the UI unresponsive, because Rancher tries to get **all** secrets every time you select a Project. Worse, if there are enough pending requests(just 3-4 people clicking is enough in my case), the pod starts to fail the probes, dies and starts a long re-election during which the Rancher is basically completely unavailable.


![secrets](https://user-images.githubusercontent.com/8925723/109018352-db0d6100-76d9-11eb-87f5-4090fac9edf0.png)

Ji993:
[...]

PR #33283, opened 2021-06-18
WIP: issue-25487 (slow secrets)

paynejacob:
https://github.com/rancher/rancher/issues/25487

Note: this pr will require a companion frontend issue, I will create it if this solution is accepted.

# Problem
When users have  a large number of helm charts (30+) and multiple releases of these charts there are many large secrets returned when they list secrets. 

# Solution
This pr adds an optional query parameter to only return the latest release of a given chart.  This dramatically decreases the response times for listing secrets.  By default the endpoint will still return all secrets.

# Testing
1. clean install of rancher
2. navigate to any project and observe the load time of the page
3. install many charts into this project `for ((i=1;i<=30;i++)); do helm install test-chart-$i bitnami/nginx; done;`
4. upgrade each chart multiple times `for ((i=1;i<=10;i++)); do for c in $(helm list -o json | jq -r ".[] | .name" ); do helm upgrade $c bitnami/nginx; done done`
5. navigate to the project page again
6. observe only the helm release secrets are the latest ones and the page still loads in a reasonable amount of time

I tested with 30 chart installations, this number was mentioned in the original issue, each with 10 releases, the default max history in helm.  Before the charts were installed `/v3/project/<project id>/namespacedsecrets` would respond in ~ 700 ms.  After installing the charts and generating the released the response times rose to ~15000 ms.  setting `exclude_old_helm_releases=true` dropped the response time to ~2800 ms.  I also tested with returning no releases but found the difference to be negligible.

### DIFF
--- a/pkg/api/norman/store/secret/store.go
+++ b/pkg/api/norman/store/secret/store.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strings"
 
+	"helm.sh/helm/v3/pkg/release"
+
 	"github.com/rancher/norman/store/proxy"
 	"github.com/rancher/norman/store/transform"
 	"github.com/rancher/norman/types"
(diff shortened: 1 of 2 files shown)


superseb:
@paynejacob Is this ready to review or WIP?

paynejacob:
@superseb this pr is on hold while I research how many other endpoints are affected by this.

paynejacob:
Closing,  further investigation has revealed there is a broader more effective solution.

Issue #33222, opened 2021-06-16
Validate Rancher runtime resources supported on k8s 1.22 clusters


snasovich:
Rancher creates resources during runtime (e.g. CRDs). Update any referenced API versions that are no longer supported in k8s 1.22 (see https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22) to ones that are supported.

Following the changes, vanilla Rancher should start on k8s 1.21 with no deprecation warnings about resources to be removed in 1.22. Currently, there are many messages `apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition` logged.

Notes:
* Usages of `apiextensions.k8s.io/v1beta1` need to be changed to `apiextensions.k8s.io/v1` which is supported since k8s 1.16. We also then need to specify the minimum k8s version on our Helm charts. Note that v2.6 release will support k8s 1.18+, so using something that's available since 1.16 is fine.
* Addressing Ingress is specifically out of scope for this issue and is split into its own issue: https://github.com/rancher/rancher/issues/33312

snasovich:
CRDs created by Wrangler are already `v1` and don't trigger the warning.
However, there is still a reference to API removed in v1.22 - https://github.com/rancher/wrangler/blob/57e55381d3ed37a7842b77cca747eccb0890fe32/pkg/generated/controllers/extensions/v1beta1/ingress.go. This will need to be investigated.

Meanwhile, it looks like v1beta1 CRDs are coming from Norman, looking into it.

snasovich:
Norman does indeed create CRDs via deprecated API: https://github.com/rancher/norman/blob/59b3523c3133434a3943f05cbd8cc6409ea976d5/store/crd/init.go#L13

So far, locally changed the code to use `v1` API instead of `v1beta1`, however the most notable change in `v1` API is that `OpenAPIV3Schema` is now required and there doesn't seem to be code that outputs schema in this format in Norman. Such (or very close) code exists in Wrangler, will look into it.

**Task 2:** Schema for CRDs in Norman.

Meanwhile, specified open schema like this:
```go
					Schema: &apiext.CustomResourceValidation {
						OpenAPIV3Schema: &apiext.JSONSchemaProps{
							Type: "object",
							Properties: map[string]apiext.JSONSchemaProps{
								"spec": {
[...]

PR #33315, opened 2021-06-21
[Draft]: k8s v1.22 updates

snasovich:
Initial updates for k8s 1.22

To be updated once https://github.com/rancher/lasso/pull/10 and https://github.com/rancher/norman/pull/401 are merged in.


snasovich:
For https://github.com/rancher/rancher/issues/33222

snasovich:
Referenced PRs are in for a while now, but holding off on this one until branched for 2.6.0 release as we don't want these changes here.
Note that since Lasso/Norman changes are already merged, we need to be careful if referenced versions need to be updated for something else. We're good so far, there are no newer changes in these frameworks.

Issue #33241, opened 2021-06-17
[GKEv2] Unable to provision a default cluster


aaronRancher:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
On Rancher `v2.5.9-rc2`, deploy a GKEv2 cluster through ember UI.
Enter cloud config.
Leave all fields default.
Enter node group name.
Create cluster.

**Result:**
Cluster provisioning fails with `Googleapi: Error 400: Private cluster config can only be set on private clusters., badRequest`
The UI sends the following `privateClusterConfig`:
```
privateClusterConfig: {
enablePrivateEndpoint: false
enablePrivateNodes: false
masterIpv4CidrBlock: null
type: "/v3/schemas/gkePrivateClusterConfig"
}
```

**Other details that may be helpful:**
GKE operator version `gke-operator:v1.0.1`
This was also reproduced on `v2.5.8`, `v2.5.9-rc1`, and `master-head` `version 54d4afc`

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `v2.5.9-rc2`
- Installation option (single install/HA): single install

aaronRancher:
This error is also seen on `v2.4-5747-head` when trying to bring up a GKEv1 cluster.

gaktive:
As of 2021 Jun 17, the following US zones are working with default values on 2.5.8:
us-east1-d
us-east4-a
us-east4-b
us-central1-a
us-central1-b
us-west1-a
us-west3-a
us-west3-b
us-west4-a
us-west4-b

Following US regions are working with default values on 2.5.8:
us-east4
us-central1

I'll create backport issues for 2.4.17 and 2.5.10.

gaktive:
We'll check with Google Support if this is expected. Otherwise, we'll surmise that other regions will slowly acquire the new defaults.

gaktive:
And just to make the issue explicit here since it wasn't already, GKE now has "Field [privateClusterConfig] cannot be nil for non-import cluster" warnings when attempting to apply to private clusters in certain zones. @cmurphy pointed to an issue with the provider, not anything we changed on our end for both GKEv1 & v2 so this setting looks to be rolling out.

gaktive:
PRs up for this which need review.

aaronRancher:
[...]

PR #33318, opened 2021-06-21
Only send PrivateClusterConfig for private clusters

cmurphy:
GKE rolled out a change that denies setting PrivateClusterConfig on
Cluster objects if they are not using private nodes. Update
kontainer-engine (to handle GKEv1) and gke-operator (GKEv2) (TODO) to
conform to this.

Depends on https://github.com/rancher/gke-operator/pull/56

https://github.com/rancher/rancher/issues/33241

### DIFF
--- a/pkg/kontainer-engine/drivers/gke/gke_driver.go
+++ b/pkg/kontainer-engine/drivers/gke/gke_driver.go
@@ -554,6 +554,10 @@ func (s *state) validate() error {
 		return fmt.Errorf("minNodeCount in the NodePool must be >= 1 and <= maxNodeCount")
 	}
 
+	if s.PrivateClusterConfig != nil && s.PrivateClusterConfig.EnablePrivateEndpoint && !s.PrivateClusterConfig.EnablePrivateNodes {
+		return fmt.Errorf("private endpoint requires private nodes")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


rmweir (commented):
LGTM

rmweir (approved):
LGTM

thedadams (approved):

pkg/kontainer-engine/drivers/gke/gke_driver.go
```diff
@@ -554,6 +554,10 @@ func (s *state) validate() error {
 		return fmt.Errorf("minNodeCount in the NodePool must be >= 1 and <= maxNodeCount")
 	}
 
+	if s.PrivateClusterConfig != nil && s.PrivateClusterConfig.EnablePrivateEndpoint && !s.PrivateClusterConfig.EnablePrivateNodes {
```
cbron, line 557: I thought it was the whole object, couldn't you just set `s.PrivateClusterConfig.EnablePrivateEndpoint` to false and not have nodes ? Then this check would pass but GCP would fail.

cmurphy: > and not have nodes

As in, not set `EnablePrivateNodes`?

This check isn't here to fix the bug, this check is so that if a an API user (UI users are prevented from doing this) accidentally sets `EnablePrivateEndpoint` to true but `EnablePrivateNodes` to false, they are alerted to their error. Before this change, they would have been alerted by the returned error from GKE, but since now we're totally ignoring `PrivateClusterConfig` if `EnablePrivateNodes` is false, we need to do that validation on GKE's behalf.

cbron: >  now we're totally ignoring PrivateClusterConfig if EnablePrivateNodes is false

that answers my question

cbron (approved):

Issue #33072, opened 2021-06-07
[CLI] Add support for flag to keep answers value as string


PennyScissors:
Issue: https://github.com/rancher/rancher/issues/26088
Related PR: https://github.com/rancher/rancher/pull/31837
Add support in the rancher CLI for the new "answersSetString" API field so that answers can be forced set to string type.

thehejik:
Tested on `rancher/rancher:master-head 4d41301`

TL;DR works as expected

* downloaded rancher-cli from https://releases.rancher.com/cli2/v2.4.12-rc2/rancher-linux-amd64-v2.4.12-rc2.tar.gz
* logged in to my rancher server by `rancher login https://myrancher --token <atokenhash>` and switched context to my downstream RKE2 cluster and Default project
* Tried to install redis by using rancher cli with following testing flags:
```
./rancher app install --set-string usePassword=false --set-string foo=bar  --set cluster.enabled=false --set usePassword=false redis redis-helm
```
* Checked API json for the installed redis app:

![rancher-cli-answersSetString](https://user-images.githubusercontent.com/12828077/125455261-0ed628ba-69c0-41a1-8e12-1fe27f9e52ff.png)

From the screenshot you can see that the key-pairs from rancher-cli command were correctly assigned to the coresponding groups of answers `answers/answersSetString`.

Close.

PR #33319, opened 2021-06-21
[master] Remove norman required tag for multicluster app answers

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/33072
This has been removed because an answer can now contain either values or valuesSetString.


brendarearden (approved):

brendarearden:
Is there a reason it was required to begin with? Maybe there is a default value needed in place of the required? What happens if both are empty?

PennyScissors:
> Is there a reason it was required to begin with? Maybe there is a default value needed in place of the required? What happens if both are empty?

I think it was required because it wouldn't make sense to have answers without any values.

PR #33321, opened 2021-06-22
Decode 1.21+ dynamic service token and lookup static token for peers

ibuildthecloud:


brandond:
This was for https://github.com/rancher/rancher/issues/32298#issuecomment-825284303

PR #33338, opened 2021-06-22
Conceal sensitive info in audit logs

thedadams:


pkg/auth/audit/audit.go
```diff
@@ -208,3 +212,41 @@ func isExist(array []string, key string) bool {
...
+
+	for key := range m {
+		if _, ok := m[key].(string); ok {
+			if a.keysToConcealRegex.MatchString(key) {
+				m[key] = redacted
+			}
+		}
+	}
+
+	newBody, err := json.Marshal(m)
```
rmweir, line 247: You could track whether anything has been concealed and if not just return the input body without marshaling.

pkg/auth/audit/audit.go
```diff
@@ -208,3 +212,41 @@ func isExist(array []string, key string) bool {
...
+		}
+	}
+
+	for key := range m {
+		if _, ok := m[key].(string); ok {
+			if a.keysToConcealRegex.MatchString(key) {
+				m[key] = redacted
+			}
+		}
+	}
```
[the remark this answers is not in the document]
thedadams: I haven't seen any such requests. Do we think we should implementation this change just in case?

pkg/auth/audit/audit.go
thedadams: I made this change.

pkg/auth/audit/audit.go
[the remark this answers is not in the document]
thedadams: I updated and added a test case for nested maps.

pkg/auth/audit/audit.go
```diff
@@ -208,3 +212,56 @@ func isExist(array []string, key string) bool {
...
+
+func (a *auditLog) concealMap(m map[string]interface{}) bool {
+	var changed bool
+	for key := range m {
+		if _, ok := m[key].(string); ok {
+			if a.keysToConcealRegex.MatchString(key) {
+				changed = true
+				m[key] = redacted
+			}
+		} else if nested, ok := m[key].(map[string]interface{}); ok && a.concealMap(nested) {
```
rmweir, line 260: Now that nested maps are being accounted for, would performance be impacted? Should it be retested?

thedadams: It was actually faster: 1.599742ms 549.326µs

rmweir (dismissed):
LGTM

pkg/auth/audit/audit.go
```diff
@@ -208,3 +212,56 @@ func isExist(array []string, key string) bool {
...
 	return false
 }
+
+func (a *auditLog) concealSensitiveData(requestURI string, body []byte) []byte {
+	var m map[string]interface{}
+	if err := json.Unmarshal(body, &m); err != nil {
+		return body
+	}
+
+	var changed bool
```
cbron, line 222: would be good to add comment that this block for secrets and below `conceal...` is for non-secrets

pkg/auth/audit/audit.go
```diff
@@ -208,3 +212,56 @@ func isExist(array []string, key string) bool {
...
+	if strings.HasSuffix(requestURI, "secrets") {
```
cbron, line 223: This seems fragile if QSP gets added, or someone adds a trailing slash. Grabbing any URI with secret in it makes more sense to me. Worst case we grab too much, but its not a big performance hit.

pkg/rancher/rancher.go
```diff
@@ -172,8 +172,11 @@ func New(ctx context.Context, clientConfg clientcmd.ClientConfig, opts *Options)
 	}
 
 	auditLogWriter := audit.NewLogWriter(opts.AuditLogPath, opts.AuditLevel, opts.AuditLogMaxage, opts.AuditLogMaxbackup, opts.AuditLogMaxsize)
-	auditFilter := audit.NewAuditLogMiddleware(auditLogWriter)
 	aggregationMiddleware := aggregation.NewMiddleware(ctx, wranglerContext.Mgmt.APIService(), wranglerContext.TunnelServer)
```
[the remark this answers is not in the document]
thedadams: Just the error check. I switched it back.

pkg/auth/audit/filter.go
```diff
@@ -8,23 +8,48 @@ import (
...
-func NewAuditLogMiddleware(auditWriter *LogWriter) func(http.Handler) http.Handler {
+func NewAuditLogMiddleware(auditWriter *LogWriter) (func(http.Handler) http.Handler, error) {
+	r, err := constructKeyConcealRegex()
 	return func(next http.Handler) http.Handler {
 		return &auditHandler{
-			next:        next,
-			auditWriter: auditWriter,
+			next:            next,
+			auditWriter:     auditWriter,
+			sanitizingRegex: r,
```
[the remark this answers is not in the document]
thedadams: Changed.

pkg/auth/audit/audit.go
thedadams: Done.

pkg/auth/audit/audit.go
thedadams: Switched to `strings.Contains`

cbron (approved):

PR #33367, opened 2021-06-24
Add k3s hardended cluster configuration

ShylajaDevadiga:


tests/validation/tests/v3_api/resource/terraform/k3s/master/policy.yaml
```diff
@@ -0,0 +1,186 @@
...
+  podSelector:
+    matchLabels:
+      k8s-app: kube-dns
+  policyTypes:
+  - Ingress
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: cis1.5-compliant-psp
```
rancher-max, line 22: We call this cis1.5, but it is actually also valid for 1.6. I actually like the name of this and don't think we need to change it, but want to call that out.

rancher-max (approved):
LGTM

bmdepesa (approved):
LGTM

PR #33377, opened 2021-06-25
Only watch specific configmaps to reduce traffic

ibuildthecloud:


pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -28,16 +32,38 @@ type handler struct {
...
-	userContext.Core.ConfigMaps("kube-system").AddHandler(ctx, "snapshotbackpopulate", h.OnChange)
+
+	// We want to watch two specific objects, not all config maps.  So we setup a custom controller
+	// to just watch those names.
+	clientFactory, err := client.NewSharedClientFactory(&userContext.RESTConfig, nil)
+	if err != nil {
+		return err
+	}
+
+	for secretName := range snapshotNames {
```
snasovich, line 49: Nitpick, but it should probably be `configMapName` here.

snasovich:
I see it's WIP, but I've ran some testing of it locally by adding extra logging to `OnChange`.
First, `OnChange` is not firing up on Config Map changes we shouldn't be monitoring which is great.
However, after creating and modifying both `k3s-etcd-snapshots` and `rke2-etcd-snapshots` Config Maps it looks like only changes to one trigger OnChange. And in my limited testing one time it was only `k3s-etcd-snapshots` that was triggering OnChange and in the other testing it was only `rke2-etcd-snapshots`. I'm not sure why it's happening, but it is happening.

snasovich:
Since we're filtering on subscribe, it's probably a good idea to remove check in OnChange (`if configMap.Namespace != "kube-system" || !snapshotNames[configMap.Name]`).

ibuildthecloud:
@snasovich Sorry, I accidentally merged this PR without the follow-ups to you comments.  Those fixes are in https://github.com/rancher/rancher/pull/33577 now.

snasovich:
> @snasovich Sorry, I accidentally merged this PR without the follow-ups to you comments. Those fixes are in #33577 now.

@ibuildthecloud , thank you. Just for my understanding, was the fix to behavior I observed in not seeing OnChange triggered by one or the other config map to start controller factories sequentially vs. via goroutine?

Also, any specific reason check `if configMap.Namespace != "kube-system" || !snapshotNames[configMap.Name]` was left in OnChange code (my last comment)?

Issue #33398, opened 2021-06-28
Additional CA certs not recognized in SLE image.


thedadams:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
1. Deploy Rancher behind a proxy with an additional CA certificate. (see [the docs](https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/chart-options/#additional-trusted-cas).
2. Go to Cluster Manager -> Catalogs

**Result:**
There should be errors regarding untrusted self-signed certificates.

**Expected Result:**
There should be no errors and the additional CA added as in the docs should be trusted.

**Other details that may be helpful:**
This is not an issue in 2.5.x because the base image is Ubuntu.

If one were to exec into the Rancher containers and run `c_rehash`, then the error will go away. Adding this command to https://github.com/rancher/rancher/blob/master/package/entrypoint.sh before starting Rancher is one option.

snasovich:
@mvarlese & @dirkmueller , any thoughts / recommendations on fix approach since it's related to https://github.com/rancher/rancher/pull/33007 ?

mvarlese:
I see the patch to run c_rehash has been approved but not merged yet.

I apologize but I am not too familiar with the Rancher stack. That said, considered SLE installs certificates in a different folder than what Ubuntu does, I think it makes sense to run c_rehash to create the sym links.

dirkmueller:
The recommended way is to use `/etc/pki/trust/anchors/` as an installation path for custom system wide trusted CAs. running `update-ca-certificates` will do the necessary logic to refresh the bundles and add links. 

c_rehash runs might work but if for whatever reason `update-ca-certificates` (pki tools) will be executed your rehash run is lost.

thedadams:
Thanks for the input, @dirkmueller. I have followed the documentation you posted.

prabalsharma:
@thedadams is testing https://github.com/rancher/rancher/issues/31846 good enough for this (along with checking Cluster Manager -> Catalogs) ? or do you have some specific test case in mind?

prabalsharma:
I guess checking Catalogs is good enough test case for this bug.

prabalsharma:
tested on 9e0d1b2

[...]

PR #33424, opened 2021-06-29
Run update-ca-certificates before launching Rancher

thedadams:
The new SLE image requires update-ca-certificates be run before
additional trusted CAs are recognized. Therefore, the command is
run prior to launch Rancher.
    
In addition, the location of the mounted certificate is changed to the
location specified in the docs: /etc/pki/trust/anchors/

Finally, the jailer also copies the certificate information to /etc/ssl
instead of /etc/ssl/certs because the Go libraries expect the certs to
be there.

Issue:
https://github.com/rancher/rancher/issues/33398

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -151,11 +151,7 @@ spec:
 {{ toYaml .Values.resources | indent 10 }}
         volumeMounts:
 {{- if .Values.additionalTrustedCAs }}
-        - mountPath: /etc/ssl/certs/ca-additional.pem
-          name: tls-ca-additional-volume
-          subPath: ca-additional.pem
-          readOnly: true
-        - mountPath: /etc/rancher/ssl/ca-additional.pem
+        - mountPath: /etc/pki/trust/anchors/ca-additional.pem
           name: tls-ca-additional-volume
           subPath: ca-additional.pem
           readOnly: true
--- a/package/entrypoint.sh
+++ b/package/entrypoint.sh
@@ -6,4 +6,7 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -x "$(command -v update-ca-certificates)" ]; then
+  update-ca-certificates
+fi
 exec tini -- rancher --http-listen-port=80 --https-listen-port=443 --audit-log-path=${AUDIT_LOG_PATH} --audit-level=${AUDIT_LEVEL} --audit-log-maxage=${AUDIT_LOG_MAXAGE} --audit-log-maxbackup=${AUDIT_LOG_MAXBACKUP} --audit-log-maxsize=${AUDIT_LOG_MAXSIZE} "${@}"
--- a/package/jailer.sh
+++ b/package/jailer.sh
@@ -11,7 +11,7 @@ fi
 
 # Build the jail directory structure
 mkdir -p /opt/jail/$NAME/dev
-mkdir -p /opt/jail/$NAME/etc/ssl/certs
+mkdir -p /opt/jail/$NAME/etc/ssl
 mkdir -p /opt/jail/$NAME/usr/bin
 mkdir -p /opt/jail/$NAME/management-state/node/nodes
 mkdir -p /opt/jail/$NAME/var/lib/rancher/management-state/bin
@@ -27,7 +27,7 @@ fi
 
 cp -r /lib /opt/jail/$NAME
 cp -r /usr/lib /opt/jail/$NAME/usr
-cp /var/lib/ca-certificates/ca-bundle.pem /opt/jail/$NAME/etc/ssl/certs
+cp /var/lib/ca-certificates/ca-bundle.pem /opt/jail/$NAME/etc/ssl
 cp /etc/resolv.conf /opt/jail/$NAME/etc/
 cp /etc/passwd /opt/jail/$NAME/etc/
 cp /etc/hosts /opt/jail/$NAME/etc/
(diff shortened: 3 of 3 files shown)


package/entrypoint.sh
```diff
@@ -6,4 +6,7 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -x "$(command -v c_rehash)" ]; then
+  c_rehash
+fi
```
nickgerace, line 11: I agree with the check here for local development, but shouldn't we fail if the executable is not found or in PATH?

thedadams: If a user doesn't set `additionalTrustedCAs` to `true`, then we don't want to fail because `c_rehash` isn't on the `PATH`.

nickgerace (approved):

package/entrypoint.sh
```diff
@@ -6,4 +6,7 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -x "$(command -v c_rehash)" ]; then
```
jiaqiluo, line 9: what if the command `c_rehash` exists but is not executable? Currently, we just ignore it.

jiaqiluo (changes requested):

package/entrypoint.sh
thedadams: The binary is included when we install `openssl` in the image, so I think it will be executable if it is there.

jiaqiluo: Ok, so we do not need to concern about it.

jiaqiluo (dismissed):
LGTM

dirkmueller:
I suggest to read https://github.com/rancher/rancher/issues/33398#issuecomment-871163998 before merging this.

thedadams:
The implementation has been updated according to @dirkmueller comment.

jiaqiluo (approved):

nickgerace (dismissed):

chart/templates/deployment.yaml
```diff
@@ -151,11 +151,7 @@ spec:
 {{ toYaml .Values.resources | indent 10 }}
         volumeMounts:
 {{- if .Values.additionalTrustedCAs }}
-        - mountPath: /etc/ssl/certs/ca-additional.pem
-          name: tls-ca-additional-volume
-          subPath: ca-additional.pem
-          readOnly: true
-        - mountPath: /etc/rancher/ssl/ca-additional.pem
```
cbron, line 158: Why are we removing the rancher one entirely ? Going back through old issues I see: 
> The difference in behaviour is the result of a change in the rancher deployment object in the helm chart in v2.2 vs 2.1, in which the custom CA file ca-additional.pem is now mounted at /etc/rancher/ssl/ca-additional.pem vs /etc/ssl/certs/ca-additional.pem (rancher/server-chart@7920ce4#diff-e3ed5b4174b5e628950137770a3e94d0). This change was made to rationalise certificate location into a single, non-OS dependent directory. However, pending updates to the git-based catalog design (15330) this breaks git-based custom catalogs from git servers presenting a cert signed with a custom CA.

So we were setting them in both places on purpose. 

If we do go with this change, it must be in the docs or guides. We'll have to open issue and find/replace.

superseb: The rancher one is there because we override the certificate location using ```ENV SSL_CERT_DIR /etc/rancher/ssl``` in the image. The other one is for the `git` binary.

thedadams: Now that we are using the SLE image, the [docs](https://documentation.suse.com/suse-caasp/4.5/html/caasp-admin/_troubleshooting_3.html#_error_x509_certificate_signed_by_unknown_authority) indicate that we should use the `/etc/pki/trust/anchors` directory. It seems that `update-ca-certificates` doesn't respect the `SSL_CERT_DIR` environment variable.

All that is to say that since we were setting the `SSL_CERT_DIR` to the custom directory and that is no longer being used, then we can remove it.

superseb: `SSL_CERT_DIR` is to override the system location for certificates when using Golang, so this can be removed if it is verified that Golang binaries (`rancher` it self but also `rancher-machine` still use the added certificates to validate connections)

thedadams: The `update-ca-certificates` command adds the certs in `/etc/pki/trust/anchors/` to `/etc/ssl/ca-bundle.pem` which is used by Go: https://golang.org/src/crypto/x509/root_linux.go

cbron: Did you test that docker machine still works with it ? I realize its also golang but we should do a smoke test.

I'm still not sold on this being all thats needed. How do the following still work: 
* Jailer: https://github.com/rancher/rancher/blob/master/package/jailer.sh#L46-L48
* RancherD (though this may be getting removed): https://github.com/rancher/rancher/blob/master/cmd/rancherd/chart/templates/daemonset.yaml#L98-L102
* RKE2: https://github.com/rancher/rancher/blob/master/pkg/controllers/provisioningv2/rke2/machineprovision/template.go#L193-L198

cbron (changes requested):

package/jailer.sh
```diff
@@ -11,7 +11,7 @@ fi
 
 # Build the jail directory structure
 mkdir -p /opt/jail/$NAME/dev
-mkdir -p /opt/jail/$NAME/etc/ssl/certs
+mkdir -p /opt/jail/$NAME/etc/ssl
```
cbron, line 14: > Finally, the jailer also copies the certificate information to /etc/ssl instead of /etc/ssl/certs because the Go libraries expect the certs to be there.

Confused why this change affects go libraries. This is all within the jail, the certs themselves are still being pulled out of the same place.

cbron (commented):
One question, and also: 
* Still need to smoke test rancher-machine
* @dramich should approve jailer code

package/jailer.sh
thedadams: The crypto package looks looks in different directories based on the OS: https://golang.org/src/crypto/x509/root_linux.go

Even though the run of `rancher-machine` is jailed, it will still look in the corresponding jailed directory for the certs.

thedadams:
`rancher-machine` doesn't work: https://github.com/rancher/rancher/issues/33556

package/jailer.sh
cbron: Ahh, so we are actually moving this because the crypto library looks in a different spot.

cbron (approved):
lgtm if my last assumption is correct

PR #33437, opened 2021-06-30
Updated kustomize config to manage io.cattle.field/appId label on apps deployment

rawmind0:
Forward https://github.com/rancher/rancher/pull/33436

### DIFF
--- a/pkg/controllers/managementuserlegacy/helm/common/common.go
+++ b/pkg/controllers/managementuserlegacy/helm/common/common.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


paynejacob (dismissed):

paynejacob (approved):

superseb (approved):

herrbpl:
So whenever I install app not from Rancher UI (aka, via helm directly), appId field is not there, and should not. After upgrading app version in Rancher, everything breaks with "MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable" message.

herrbpl:
This applies to deployments as well, please add this change for deployments as well.

PR #33452, opened 2021-07-01
Set aggregation prefix to /k8s/clusters/name when applicable

gitlawr:
Related issue: https://github.com/harvester/harvester/issues/922

On the master head, we support Rancher API extension in the following way:
1. Create the following APIService object in a downstream cluster.
```
kind: APIService
apiVersion: management.cattle.io/v3
metadata:
  name: harvester
spec:
  secretName: harvester-aggregation
  secretNamespace: harvester-system
  pathPrefixes:
    - /v1/harvester
```
2. Start the aggregation client: https://github.com/gitlawr/harvester/blob/apiservice-v1-harvester/pkg/server/types.go#L267
3. Then the extension API can be accessed via `<SERVER_URL>/k8s/clusters/name/v1/harvester`

Problem:
The X-API-URL-Prefix header is added as /v1/harvester in https://github.com/rancher/rancher/blob/92b822fb2e1437a7bacb7a5a65ab5fc02608a7f3/pkg/api/steve/aggregation/aggregation.go#L103

Harvester itself is a steve server which parses the prefix header in https://github.com/rancher/apiserver/blob/f943376c4b42cd1c38acc4bd085aa7bf44305b64/pkg/urlbuilder/base.go#L14

As a result, the /v1/harvester path is duplicated and the resource link becomes `https://<server-url>/v1/harvester/v1/harvester/some.apigroup.resources`

Solution:
Change the aggregation prefix to /k8s/clusters/name according to the request URL path.

### DIFF
--- a/pkg/api/steve/aggregation/aggregation.go
+++ b/pkg/api/steve/aggregation/aggregation.go
@@ -3,6 +3,7 @@ package aggregation
 import (
 	"context"
 	"net/http"
+	"regexp"
 	"sort"
 	"sync"
 	"time"
@@ -15,6 +16,10 @@ import (
 	"github.com/rancher/wrangler/pkg/relatedresource"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/rest"
+)
+
+var (
+	clusterPrefixRegexp = regexp.MustCompile(`^/k8s/clusters/[^/]+`)
 )
 
 type aggregationHandler struct {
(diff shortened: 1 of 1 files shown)


nickgerace (approved):
LGTM for general logic. Will need more reviews

paynejacob (approved):
This makes sense, we should ensure that the old value is not hard coded somewhere else to avoid a regression.

dramich (approved):

Issue #31638, opened 2021-03-09
Global DNS app gets uninstalled/removed immediately after install due to OwnerReferences cleanup in k8s v1.20


abhi1693:
**What kind of request is this (question/bug/enhancement/feature request):**
I have deployed a new cluster and have added the correct Cloudflare DNS provider using the UI. After a few seconds, I see that a new application is deployed in the `local` cluster and then immediately removed by the rancher. I have tried adding new API keys and the result is still the same.

**Steps to reproduce (least amount of steps as possible):**
1. Go to global DNS provider and add Cloudflare details
2. Observe the local cluster

**Result:**
The external DNS app should not be deleted on its own.

**Other details that may be helpful:**
Logs from rancher
```

[storage] 2021/03/09 09:08:44 getting release history for "systemapp-controllergateway"
[tiller] 2021/03/09 09:08:44 uninstall: Release not loaded: systemapp-controllergateway
W0309 09:09:23.731519      10 warnings.go:80] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
2021/03/09 09:09:24 [INFO] deleteSecretData: map[cattle-global-data:2jcpqfj8qzzlfmnq6nrn8pfwvlq97xh6gsx688djn7q9zft27925td:true]
W0309 09:10:11.196045      10 warnings.go:80] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
2021/03/09 09:10:12 [INFO] Installing chart using helm version: rancher-helm
[main] 2021/03/09 09:10:12 Starting Tiller v2.16.8-rancher1 (tls=false)
[main] 2021/03/09 09:10:12 GRPC listening on :58543
[main] 2021/03/09 09:10:12 Probes listening on :45187
[main] 2021/03/09 09:10:12 Storage driver is ConfigMap
[main] 2021/03/09 09:10:12 Max history per release is 10
[tiller] 2021/03/09 09:10:12 getting history for release systemapp-controllergateway
[storage] 2021/03/09 09:10:12 getting release history for "systemapp-controllergateway"
Release "systemapp-controllergateway" does not exist. Installing it now.
[tiller] 2021/03/09 09:10:12 preparing install for systemapp-controllergateway
[storage] 2021/03/09 09:10:12 getting release history for "systemapp-controllergateway"
[...]

PR #33471, opened 2021-07-02
Remove cross namespace ownerRef, handle app delete

prachidamle:
https://github.com/rancher/rancher/issues/31638

Fix is to not use the cross namespace ownerreference (external-dns app in systemProject namespace points to GlobalDNSProvider in cattle-global-data namespace) that causes k8s 1.20 to enforce deletion of the external-dns app.

Instead of relying on the ownerRef for deletion of the app after GlobalDNSProvider is removed, fix is to handle the delete event in the controller to remove the app.

### DIFF
--- a/pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
+++ b/pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
 )
 
 const (
@@ -63,7 +64,7 @@ func newGlobalDNSProviderCatalogLauncher(ctx context.Context, mgmt *config.Manag
 //sync is called periodically and on real updates
 func (n *ProviderCatalogLauncher) sync(key string, obj *v3.GlobalDnsProvider) (runtime.Object, error) {
 	if obj == nil || obj.DeletionTimestamp != nil {
-		return nil, nil
+		return nil, n.deleteExternalDNSApp(key)
 	}
 	metaAccessor, err := meta.Accessor(obj)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
```diff
@@ -276,17 +266,37 @@ func (n *ProviderCatalogLauncher) createUpdateExternalDNSApp(obj *v3.GlobalDnsPr
 	return nil, nil
 }
 
-func (n *ProviderCatalogLauncher) getProviderIfAlreadyRunning(obj *v3.GlobalDnsProvider) (*pv3.App, error) {
+func (n *ProviderCatalogLauncher) deleteExternalDNSApp(key string, obj *v3.GlobalDnsProvider) (runtime.Object, error) {
+	var err error
```
snasovich, line 270: Maybe I'm missing something obvious here, but why do we have to declare `err` here?

pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
```diff
@@ -276,17 +266,37 @@ func (n *ProviderCatalogLauncher) createUpdateExternalDNSApp(obj *v3.GlobalDnsPr
...
+	_, globalDNSproviderName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return obj, err
+	}
+	//check if provider already running for this GlobalDNSProvider.
+	existingApp, err := n.getProviderIfAlreadyRunning(globalDNSproviderName)
+	if err != nil {
+		return nil, err
```
snasovich, line 278: Do we want to `return obj, err` instead here? Is the idea that we return `nil` if deletion was successful and `obj` otherwise? If that's not the case, we could simply remove `obj` parameter from this func as we don't need it here (and it's quite likely `nil` anyway) and we could change calling function to `return obj, n.deleteExternalDNSApp(key)`

snasovich (changes requested):

pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
prachidamle: done, will return ``` return obj, nil ```

pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
prachidamle: cleaned up

pkg/controllers/managementlegacy/globaldns/globaldnsprovider_catalog_launcher.go
prachidamle: update: we should return nil, nil in this case since there is no update.

snasovich (approved):

dramich (commented):
The pr doesn't remove any existing owner references so if a user upgrades to 1.20 aren't the apps still going to get removed?

prachidamle:
> The pr doesn't remove any existing owner references so if a user upgrades to 1.20 aren't the apps still going to get removed?

@dramich The periodic sync will take care of this. A recreation of the app will happen during the periodic sync since no app is found. Won't the periodic sync run every hour or so?
 Ofcourse we can also document that customers can force recreation of the app if they want it immediate.

abhi1693:
@prachidamle I was the guy who raised the issue for which this fix is being worked on. From your comment, does this mean that there might be a downtime of up to an hour if we don't manually edit the DNS entries again?

snasovich:
@abhi1693 , thank you for following up on this issue. Prachi is out this week, but per her response, I believe your understanding is correct.
@dramich , please correct me if I'm wrong here. Also, do you see a way to address this so there is no temporary removal of Global DNS App following k8s upgrade?

dramich:
We discussed this last week and on startup the sync handler will check if the app has an owner reference and remove it

snasovich:
@dramich , thank you.
@abhi1693 , just to expand on the above, the plan is to update this PR with additional changes that will ensure OwnerReferences is removed on existing installations so that when k8s is upgraded to v1.20 the app is not deleted.
ETA for the change is early next week when Prachi is back.

abhi1693:
@snasovich Thanks for the update

prachidamle:
@abhi1693 

- For the case where the setups are already using k8s 1.20 and the rancher-external-dns app has gotten deleted, the solution is to upgrade to rancher v2.6. Upon upgrade, rancher's startup sync will recreate the deleted app without the problematic owner reference. This sync runs immediately on a restart/upgrade and then runs periodically every hour. The sync that runs at the startup after upgrading to v2.6 will recreate the apps.

This PR is updated further to handle the second case:
- The setups which are using K8s version less than 1.20 and not yet seeing any app deletion:  After they are upgraded to use rancher v2.6, upon upgrade rancher controller will drop the ownerRef if found. So when these setups move to k8s 1.20, there is no app delete triggered.

dramich (approved):

Issue #33490, opened 2021-07-06
[2.4] gke v1 - getting error 500 for us-central1 region


prabalsharma:
**What kind of request is this (question/bug/enhancement/feature request):** Bug

**Steps to reproduce (least amount of steps as possible):** 
Create a GKE us-central1 cluster on 2.4-head

**Result:**
`Googleapi: Error 500: Internal error encountered., backendError`

**Expected Result:**

**Other details that may be helpful:**
I was able to create us-central1 GKE cluster from GCP

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
- Installation option (single install/HA):

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
(paste the output here)
```

- Docker version (use `docker version`):

```
(paste the output here)
```

prabalsharma:
I was able to create central1 regional clusters on 2.5 and 2.6 head

sowmyav27:
Happens for GKE version`1.18.20-gke.501` seen on UI.

prabalsharma:
was able to provision a GKE cluster on us-central1 region with the default k8s version (v1.18.20-gke.501)

PR #33499, opened 2021-07-07
Accept zone or region for GKE versions

cmurphy:
Without this patch, /meta/gkeVersions only accepts 'zone' as a location
parameter and so only returns kubernetes versions available in a given
zone. The available versions in a zone is not necessarily the same as
the versions available for a region, even if a zone is within the
region. This change updates the /meta/gkeVersions route to accept either
'zone' or 'region' and return the appropriate list.

Example usage:

  GET /meta/gkeVersions?projectId=my-project&cloudCredentialId=cattle-global-data:cc-abcde?region=us-central1

  GET /meta/gkeVersions?projectId=my-project&cloudCredentialId=cattle-global-data:cc-abcde?zone=us-central1-f

Backport of https://github.com/rancher/rancher/pull/33498

https://github.com/rancher/rancher/issues/33490

### DIFF
--- a/pkg/api/norman/customization/gke/listers.go
+++ b/pkg/api/norman/customization/gke/listers.go
@@ -139,8 +139,14 @@ func listServiceAccounts(ctx context.Context, cap *Capabilities) ([]byte, int, e
 }
 
 func listVersions(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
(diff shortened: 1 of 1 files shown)


rmweir (approved):
LGTM

thedadams (approved):

cbron (approved):

PR #33520, opened 2021-07-07
Fix Connection header

cmurphy:


snasovich:
LGTM, @dramich please take a peek at it. Most specifically the location of this new middleware in the chain. It looks like it doesn't matter where it is in relation to other middlewares, but want you to confirm. Thanks!

dramich (approved):

Issue #33423, opened 2021-06-29
Remove experimental rancherd asset


deniseschannon:
Currently, Rancher ships with a rancherd asset which was an experimental way of installing Rancher introduced in Rancher v2.5. 

As we are looking to define the exact use case and implementation of this asset, we have decided it needs to be updated and re-architected and are looking to remove the experimental ability of this installation.

sowmyav27:
To validate:

On an RC, verify the RancherD assets is not generated.

cbron:
Only deleting asset, nothing else

jiaqiluo:
The build process should not publish any rancherd assets now in the master branch

cbron:
Assigning @rmweir for dev-validate

cbron:
Assigning @brendarearden for dev-validate

deniseschannon:
Available to test in v2.6.0-rc1

brendarearden:
Validation: PASSED

Steps to Reproduce: 

1. Go to a previous release tag, in this case [v2.5.9](https://github.com/rancher/rancher/releases/tag/v2.5.9). 
2. Check the assets list for `rancherd-amd64` and `rancherd-amd64.tar.gz` 
3. See that the assets are available in the Assets list.

![Screen Shot 2021-08-05 at 9 02 14 AM](https://user-images.githubusercontent.com/19455540/128382832-b17ad5f6-48a7-4013-b64b-96f3eb2f5596.png)

Steps to Test: 

1. Go to the [v2.6.0-rc1](https://github.com/rancher/rancher/releases/tag/v2.6.0-rc1) tag.
2. Check the assets list for `rancherd-amd64` and `rancherd-amd64.tar.gz` 
3. See that the assets are no longer list in the Assets list. 

![Screen Shot 2021-08-05 at 9 14 54 AM](https://user-images.githubusercontent.com/19455540/128384570-acdd562d-e05b-49d3-a5fb-e84cfc77d9a1.png)

PR #33525, opened 2021-07-07
remove rancherd assets from the build process

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/33423

This PR is limited to only stop publishing the rancerd assets: https://github.com/rancher/rancher/issues/33423#issuecomment-875999466

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -195,25 +195,6 @@ steps:
     event:
     - push
 
-- name: bin-publish
-  image: plugins/gcs
-  settings:
-    acl:
-    - allUsers:READER
-    cache_control: "public,no-cache,proxy-revalidate"
-    source: bin/rancherd
-    target: releases.rancher.com/rancherd/${DRONE_BRANCH}/rancherd
-    token:
-      from_secret: google_auth_key
-  when:
-    ref:
-      include:
-      - "refs/heads/master"
-      - "refs/heads/release/v*"
-    event:
-    - push
-
-
 - name: docker-publish
   image: plugins/docker
   settings:
@@ -305,7 +286,6 @@ steps:
     - sha256
     files:
     - "bin/rancher-*"
-    - "bin/rancherd-*"
     prerelease: true
     title: "Pre-release ${DRONE_TAG}"
     note: ./bin/rancher-components.txt
@@ -326,7 +306,6 @@ steps:
     - sha256
     files:
     - "bin/rancher-*"
-    - "bin/rancherd-*"
   when:
     event:
     - tag


thedadams:
@jiaqiluo Can we also delete the `rancherd` related things from `.drone.yml`?
https://github.com/rancher/rancher/blob/e7045cac6dde4d0fe7fa31219217fa7b207e14e7/.drone.yml#L198-L214
https://github.com/rancher/rancher/blob/e7045cac6dde4d0fe7fa31219217fa7b207e14e7/.drone.yml#L308
https://github.com/rancher/rancher/blob/e7045cac6dde4d0fe7fa31219217fa7b207e14e7/.drone.yml#L329

jiaqiluo:
@thedadams    `rancherd` is removed from from` .drone.yml`

jiaqiluo:
@cbron @thedadams  Besides those two files, `rancherD` is also mentioned and used in multiple files under https://github.com/rancher/rancher/tree/master/cmd/rancherd and https://github.com/rancher/rancher/tree/master/scripts

thedadams:
@jiaqiluo I think the only one that should be removed is here: https://github.com/rancher/rancher/blob/d9ebb64616bacd2e9a1cfa35e95a914bf24fbd82/scripts/build#L6-L7

I don't think we should build `rancherd` anymore. If it turns out that we want to do that again, then we can add these lines back.

scripts/package
```diff
@@ -69,14 +69,3 @@ if [ ! -d $CHART_REPO_DIR ]; then
     rm -f $INDEX_PATH $CHART_REPO_DIR/assets/index.yaml
 fi
 
-if [ ${ARCH} == amd64 ]; then
-    # Move this out of ARCH check for local dev on non-amd64 hardware.
-    TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go $SYSTEM_CHART_REPO_DIR $CHART_REPO_DIR/assets $IMAGE $AGENT_IMAGE $RUNTIME_IMAGE
```
rmweir, line 74: Isn't this still needed for generating image text files?

rmweir (changes requested):

scripts/package
jiaqiluo: good catch!

scripts/package
```diff
@@ -15,7 +15,7 @@ cd $(dirname $0)/../package
 
 ../scripts/create-components-file.sh
 
-cp ../bin/rancher.yaml ../bin/rancher-namespace.yaml ../bin/rancher ../bin/agent ../bin/data.json ../bin/k3s-airgap-images.tar .
```
jiaqiluo, line 18: `../bin/rancher.yaml` and  `../bin/rancher-namespace.yaml` are generated by the `build-rancherd` script.

thedadams (dismissed):

thedadams (approved):

.drone.yml
```diff
@@ -195,25 +195,6 @@ steps:
...
-      from_secret: google_auth_key
-  when:
-    ref:
-      include:
-      - "refs/heads/master"
-      - "refs/heads/release/v*"
-    event:
-    - push
-
-
```
cbron, line 216: This one is debatable, since all we really need to do is remove it from the release. But I think for now removing it is the right thing, so if people are using automation to go find it then they can't anymore.

cbron (approved):
Saw convo above about stopping building of it too. Some of this code is getting repurposed, but I'm not sure how much. For now let's leave any other cleanup to a later time, and use this PR only to block usage of it. 
Lgtm, wait for Ricky to merge.

rmweir (approved):
LGTM

Issue #33534, opened 2021-07-08
Unable to update k8s on Imported AKS cluster when using Location other than `eastus`


thehejik:
**Rancher Server Setup**
- Rancher version: `rancher/rancher:master-head cf5a75f`
- Installation option (Docker install/Helm Chart): any
- Proxy/Cert Details: -

**Information about the Cluster**
- Kubernetes version: any AKS k8s
- Cluster Type (Local/Downstream): custom local, Imported AKS (default values but region changed to `westeurope`)

**Describe the bug**
When MS `Kubernetes services` aka AKS is deployed from Azure Portal UI in different Location than `eastus` (eg to existing  Resource Group in `westeurope` location the cluster cannot be updated to newer k8s over rancher due problem with hardcoded location. On attempt to update from rancher the status of the cluster is `Updating` with following Error message: 

`Failed to update cluster: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="InvalidResourceLocation" Message="The resource 'thehejik-aks-forimport' already exists in location 'westeurope' in resource group 'thehejik-rancher'. A resource with the same name cannot be created in location 'eastus'. Please select a new resource name."`

Problem is that `Location` field when doing `Edit Cluster` is probably hardcoded to `eastus` value, it's inactive and cannot be changed:
![imported_aks_update](https://user-images.githubusercontent.com/12828077/124949525-ee2e2700-e011-11eb-90a7-122ad6eb190c.png)

**To Reproduce**
* deploy a AKS in Azure Portal in different Location than `eastus` - I was deploying into my own Resource Group assigned to `westeurope` location
* Import the AKS cluster into Rancher 
* Once the cluster is Active perform `Edit Cluster`, change k8s version and Apply

**Result**
Update will fail with the error message above.

**Other info**
The same works ok for Hosted AKS

cmurphy:
Backend PR merged, will be ready to test when UI issue https://github.com/rancher/rancher/issues/33540 is complete

thehejik:
Tested on `rancher/rancher:master-head e83ac8c` with `rancher/aks-operator:v1.0.1-rc15`

The location is now correctly fetched for the imported AKS server living in `norwayeast` azure location on attempt to update the cluster from rancher.
![location](https://user-images.githubusercontent.com/12828077/125811453-5b5f5da1-8e81-44ab-aeca-585922060080.png)

[...]

PR #33539, opened 2021-07-08
Add Location to /meta/aksClusters list

cmurphy:
Add the AKS cluster's location to the list of cluster objects returned
by /meta/aksClusters. This can be used by the UI to set the
ResourceLocation field when importing a cluster.

https://github.com/rancher/rancher/issues/33534

### DIFF
--- a/pkg/api/norman/customization/aks/listers.go
+++ b/pkg/api/norman/customization/aks/listers.go
@@ -221,6 +221,7 @@ type clustersResponseBody struct {
 	ResourceGroup string `json:"resourceGroup"`
 	ClusterName   string `json:"clusterName"`
 	RBACEnabled   bool   `json:"rbacEnabled"`
+	Location      string `json:"location"`
 }
 
 func listClusters(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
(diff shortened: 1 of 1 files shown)


pkg/api/norman/customization/aks/listers.go
```diff
@@ -248,6 +249,7 @@ func listClusters(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
 					return nil, http.StatusInternalServerError, fmt.Errorf("could not parse virtual network ID")
 				}
 				tmpCluster.ResourceGroup = match[1]
+				tmpCluster.Location = to.String(cluster.Location)
```
thedadams, line 252: Could we put this below line 244 so the location gets set regardless of whether `cluster.ID != nil`?

thedadams (changes requested):

pkg/api/norman/customization/aks/listers.go
cmurphy: done

thedadams (approved):

mjura (approved):
LGTM

rmweir (commented):
LGTM

rmweir (approved):
LGTM

Issue #33517, opened 2021-07-07
aks v2 - list of VNET should be based on region selected by a user


prabalsharma:
**Rancher Server Setup**
- Rancher version: master-3e44415
- Installation option (Docker install/Helm Chart): Docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details: none

**Information about the Cluster**
- Kubernetes version:
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):
Hosted = AKS

**Describe the bug**

On cluster creation, if a user selects Azure CNI in network option, Rancher populates a list of VNETs from all regions. This should not be the case, as a user might select a VNET from a wrong region, resulting in error.
**To Reproduce**
 

1. Create AKS cluster
2. Select cluster location - EastUS
3. Select network plugin - Azure CNI
4. Check dropdown list of Virtual Network - Subnet (Subnet CIDR)
You will be able to see VNETs from all regions (which should not be the case as the cluster location has already been selected)
5. Select a VNET from a region different than your cluster location. Eg. WestUS

**Result**
```Error failed to create cluster: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="InvalidParameter" Message="The Agentpool agentpool's vnet Prabal-rg-vnet and the cluster are not in the same region. Vnet region is westus while cluster region is eastus."```
**Expected Result**

**Screenshots**

**Additional context**
<!--Add any other context about the problem here.—>

cmurphy:
Backend PR merged, will be ready to test when UI issue https://github.com/rancher/rancher/issues/33542 is complete

prabalsharma:
passed on 44bb5b4

PR #33541, opened 2021-07-08
Filter by region in /meta/aksVirtualNetworks

cmurphy:
This change optionally enables filtering by region when listing ASK
virtual networks. It is backwards compatible: if region is not set, it
defaults to the old behavior of listing all virtual networks. Also adds
"location" as an attribute returned in the list of vnets. This isn't
functionally necessary for the UI but it's nice for developers to be
able to validate the results.

https://github.com/rancher/rancher/issues/33517


thedadams (approved):

mjura (approved):
LGTM

rmweir (approved):
LGTM

Issue #32874, opened 2021-05-25
rke2 cluster: etcd snapshot  cluster creation fails, if etcd snapshot creation is enabled while creating a cluster


anupama2501:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. From cluster explorer --> create a cluster RKE2 DO (1 node all roles) and from edit as yaml, enable etcd snapshot and click on save.
2. On the cluster we see the following error and the cluster creation is stuck 
`failed to find node to perform etcd snapshot`

**Result:**
While creating a cluster if we edit yaml and provide etcd snapshot details, eg., name, once the node is up and running snapshot should be taken but the cluster creation fails with the below error.

**Other details that may be helpful:**
Some error logs on the rancher server:
```
 rkecluster fleet-default/test-anu: refreshing etcd create state
2021/05/25 16:47:53 [DEBUG] DesiredSet - No change(2) cluster.x-k8s.io/v1alpha4, Kind=Cluster fleet-default/test-anu for rke-cluster fleet-default/test-anu
2021/05/25 16:47:53 [DEBUG] DesiredSet - No change(2) rke.cattle.io/v1, Kind=RKEBootstrapTemplate fleet-default/test-anu-bootstrap-template for rke-cluster fleet-default/test-anu
2021/05/25 16:47:53 [ERROR] error syncing 'fleet-default/test-anu': handler planner: failed to find node to perform etcd snapshot, requeuing
2021/05/25 16:47:53 [DEBUG] Bootstrap provider is not ready, requeuing
2021/05/25 16:47:53 [DEBUG] DesiredSet - No change(2) rke-node.cattle.io/v1, Kind=DigitaloceanMachineTemplate fleet-default/test-anu-nodepool-pool1 for rke-cluster fleet-default/test-anu
2021/05/25 16:47:53 [ERROR] error syncing 'fleet-default/test-anu': handler planner: failed to find node to perform etcd snapshot, requeuing
2021/05/25 16:47:53 [DEBUG] DesiredSet - No change(2) cluster.x-k8s.io/v1alpha4, Kind=MachineDeployment fleet-default/test-anu-nodepool-pool1 for rke-cluster fleet-default/test-anu
2021/05/25 16:47:53 [DEBUG] Unable to retrieve Node status, missing NodeRef

```

**Environment information**
[...]

Issue #32391, opened 2021-04-27
RKE2 Provisioning: Day 2 Ops - On Demand Etcd Snapshot and Restore (Local Snapshots)


deniseschannon:
Support for Local snapshot and Restore Use case

```
apiVersion: provisioning.cattle.io/v1
kind: Cluster
metadata:
  name: #string
  annotations:
    {}
#    key: string
  labels:
    {}
#    key: string
  namespace: fleet-default
spec:
  cloudCredentialSecretName: do
  kubernetesVersion: v1.21.0-alpha2+rke2r1
  rkeConfig:

#    etcdSnapshotCreate:
#      name: string
#      nodeName: string
#      s3:
#        bucket: string
#        cloudCredentialName: string
#        endpoint: string
#        endpointCA: string
#        folder: string
#        region: string
#        skipSSLVerify: boolean
```

anupama2501:
On master-744c23307d61bd5e0951a1a6c76f32335f8011fa-head
1. From cluster explorer, created a node driver cluster and from edit as yaml and provided the following config
etcdSnapshotCreate:
      name: test-anu
Cluster creation has failed with the error:
`failed to find node to perform etcd snapshot`
2. From cluster explorer, created a node driver cluster and once the node is active from edit as yaml and provided the following config
etcdSnapshotCreate:
    etcdSnapshotCreate:
      name: anu-snapshot      

<img width="519" alt="Screen Shot 2021-05-25 at 1 58 24 PM" src="https://user-images.githubusercontent.com/38144301/119567540-4b6f6100-bd61-11eb-84a8-2e67439eea3d.png">

anupama2501:
Re-opening the ticket for testcase1: 

> From cluster explorer, created a node driver cluster and from edit as yaml and provided the following config
etcdSnapshotCreate:
name: test-anu
Cluster creation has failed with the error:
failed to find node to perform etcd snapshot

anupama2501:
Keeping this issue in to-test for further validations and creating new tickets for issues seen with etcd snapshot rke2

deniseschannon:
@anupama2501 The etcd object is changing as of yesterday, so please hold off on testing.

deniseschannon:
@anupama2501 Never mind, this yaml is still valid and can be tested using master.

anupama2501:
On master-head 7d5a854
1. Create a multinode cluster 1 node cp + worker and 3 etcd nodes
[...]

PR #33564, opened 2021-07-10
Wait for etcd snapshot if node is not provisioned

thedadams:
If a user created a cluster with etcdSnapshotCreate set, then the
planner would try to create a snapshot before the node was provisioned.
Now, the planner will wait until the node is provisioned before
attempting to take a snapshot.

Issues:
https://github.com/rancher/rancher/issues/32874
https://github.com/rancher/rancher/issues/32391

### DIFF
--- a/pkg/provisioningv2/rke2/planner/etcdcreate.go
+++ b/pkg/provisioningv2/rke2/planner/etcdcreate.go
@@ -108,6 +108,10 @@ func (e *etcdCreate) createPlan(controlPlane *rkev1.RKEControlPlane, snapshot *r
 }
 
 func (e *etcdCreate) Create(controlPlane *rkev1.RKEControlPlane, clusterPlan *plan.Plan) error {
+	if !Provisioned.IsTrue(controlPlane) && controlPlane.Status.ETCDSnapshotCreatePhase == "" {
+		return nil
+	}
+
 	if controlPlane.Spec.ETCDSnapshotCreate == nil {
 		return e.resetEtcdCreateState(controlPlane)
 	}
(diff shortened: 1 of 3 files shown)


pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -65,6 +66,8 @@ const (
 	SecretTypeMachinePlan = "rke.cattle.io/machine-plan"
 
 	authnWebhookFileName = "/var/lib/rancher/%s/kube-api-authn-webhook.yaml"
+
+	Provisioned = condition.Cond("Provisioned")
```
thedadams, line 70: This was moved to avoid an import cycle.

pkg/provisioningv2/rke2/planner/etcdcreate.go
```diff
@@ -108,6 +108,10 @@ func (e *etcdCreate) createPlan(controlPlane *rkev1.RKEControlPlane, snapshot *r
 }
 
 func (e *etcdCreate) Create(controlPlane *rkev1.RKEControlPlane, clusterPlan *plan.Plan) error {
+	if !Provisioned.IsTrue(controlPlane) && controlPlane.Status.ETCDSnapshotCreatePhase == "" {
```
snasovich, line 111: What would be the situation when controlPlane is not Provisioned, but `controlPlane.Status.ETCDSnapshotCreatePhase` is not empty and why do we want to proceed further?
(Not questioning this, just wondering for my understanding)

paynejacob (approved):

pkg/provisioningv2/rke2/planner/etcdcreate.go
thedadams: Once a snapshot is created, the plan for the `controlPlane` will be updated, causing it to go into a non-provisioned state. In that case, we want to continue.

This change only stops the snapshot create from starting when the `controlPlane` is in a non-provisioned state.

snasovich (approved):

dramich (approved):

Issue #32117, opened 2021-04-14
Errors on master-head rancher server on assigning a role to a user


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Bring up a single node master-head
- deploy a downstream cluster
- Add a user user1 as a standard user, `Unable to fetch user info` is seen on the User. 
- `Principal not found: Internal Server Error` Error is seen when trying to add this user1 to the downstream cluster's project as PM/PO

Errors in rancher logs:
```
2021/04/14 03:22:12 [ERROR] failed on subscribe alertmanager: NotFound 404: the server could not find the requested resource (get alertmanagers.meta.k8s.io)
2021/04/14 03:22:17 [ERROR] failed on subscribe prometheusRule: NotFound 404: the server could not find the requested resource (get prometheusrules.meta.k8s.io)
2021/04/14 03:22:17 [ERROR] failed on subscribe alertmanager: NotFound 404: the server could not find the requested resource (get alertmanagers.meta.k8s.io)
2021/04/14 03:22:17 [ERROR] failed on subscribe serviceMonitor: NotFound 404: the server could not find the requested resource (get servicemonitors.meta.k8s.io)
2021/04/14 03:22:17 [ERROR] failed on subscribe prometheus: NotFound 404: the server could not find the requested resource (get prometheuses.meta.k8s.io)
I0414 03:22:31.391843      23 topology_manager.go:221] [topologymanager] RemoveContainer - Container ID: 2af7b84953c924494ac4468c9361005369dff661484ad0b1655de30be8725ec2
2021/04/14 03:22:42 [ERROR] Unknown error: users.management.cattle.io "" not found
2021/04/14 03:22:50 [INFO] Creating new GlobalRoleBinding for GlobalRoleBinding grb-mrxwv
2021/04/14 03:22:50 [INFO] [mgmt-auth-grb-controller] Creating clusterRoleBinding for globalRoleBinding grb-mrxwv for user u-c74d9 with role cattle-globalrole-user
2021/04/14 03:22:50 [ERROR] Unknown error: users.management.cattle.io "" not found
2021/04/14 03:22:50 [ERROR] Unknown error: users.management.cattle.io "" not found
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head commit id: `bb8d5a2 `
- Installation option (single install/HA): single

[...]

PR #33566, opened 2021-07-12
Fix API redirect for paths with //

snasovich:
For #32117

Currently requests to `/principals/local%3A%2F%2Fu-lr7ts?sort=name` (`/principals/local://u-lr7ts?sort=name`) respond with redirect (301) to `/principals/local:/u-lr7ts?sort=name` which in turn responds with 500 (error message is `users.management.cattle.io "" not found`).

The fix is to change all mux-es in the API handler chain to not redirect requests containing multiple `/`s which is a default mux behavior.

Note that this is a regression as v2.5.8 release handles such requests correctly.
Also, note browser caches 301 HTTP responses, so if testing the fix on users the issue was reproduced previously, make sure to clear browser's cache beforehand.

**Edit after additional investigation:**
Only URL-encoded requests (like `/principals/local%3A%2F%2Fu-lr7ts?sort=name` and NOT `/principals/local://u-lr7ts?sort=name`) are handled without redirect in v2.5.8.
However, both return redirect on `master`. This is due to additional `aggregationMiddleware` added to Rancher's HTTP handler chain (to handle `/fleet/webhook` route). It may be an option to move this handler down the chain after the one that handles API, however unlike other handlers using mux, this handler did not set `mux.UseEncodedPath`. So, to fix issue, a call to `mux.UseEncodedPath` was added.

### DIFF
--- a/pkg/api/steve/aggregation/aggregation.go
+++ b/pkg/api/steve/aggregation/aggregation.go
@@ -65,6 +65,7 @@ func (h *aggregationHandler) next(notFound http.Handler) http.Handler {
 
 func (h *aggregationHandler) setEntries(routes []routeEntry) {
 	mux := mux.NewRouter()
+	mux.UseEncodedPath()
 	for _, entry := range routes {
 		if entry.prefix != "" {
 			mux.PathPrefix(entry.prefix).Handler(h.makeHandler(entry.uuid))


dramich (commented):
Where did the regression happen? I don't think we should be doing this change to all routers because it most likely has much wider impacts and should be limited to just the route/handler the deals with principals. But if it was working in 2.5.8 we should understand why it stopped working now. Did we introduce a middleware that catches the principals before we handle the request?

snasovich:
> Where did the regression happen? I don't think we should be doing this change to all routers because it most likely has much wider impacts and should be limited to just the route/handler the deals with principals. But if it was working in 2.5.8 we should understand why it stopped working now. Did we introduce a middleware that catches the principals before we handle the request?

@dramich , let me dig deeper into this and see why 2.5.8 works fine there. Meanwhile, converted to draft.
I agree updating all mux-es in handlers is not ideal, so maybe there is a better fix.

snasovich:
@dramich , thank you for suggesting to investigate the regression further. I believe I found the root cause and changed the fix accordingly. Please review the updated description for details.
It's ready for review.

dramich (approved):

PR #33577, opened 2021-07-12
Various fixes

ibuildthecloud:


pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -92,11 +95,12 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (*corev1.Con
...
 
-func configMapToSnapshots(configMap *corev1.ConfigMap) (result []rkev1.ETCDSnapshot, _ error) {
-	for _, v := range configMap.Data {
+func (h *handler) configMapToSnapshots(configMap *corev1.ConfigMap) (result []rkev1.ETCDSnapshot, _ error) {
+	for k, v := range configMap.Data {
 		file := &snapshotFile{}
 		if err := json.Unmarshal([]byte(v), file); err != nil {
-			return nil, err
+			logrus.Errorf("invalid non-json value in %s/%s for key %s in cluster %s", configMap.Namespace, configMap.Name, k, h.clusterName)
+			return nil, nil
```
snasovich, line 103: Is it intended to return `nil` error in this case? If that's the case the func never returns non-nil error and it can be removed from the return.

Issue #33337, opened 2021-06-22
Keycloak OIDC - can't access with users members of nested groups with Site access restricted to specific users/groups


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Enable Keycloak OIDC on latest `2.6 master-head` build.
- On site access use Restricted access to specify which users and groups are allowed to login
- Add a group that is a nested under a parent group

Example here is:

![Screen Shot 2021-06-22 at 4 11 34 PM](https://user-images.githubusercontent.com/27659/123011077-a00ff700-d374-11eb-98e2-d2ce5862594c.png)

`nestedautogroupnested2` it has member `testuserx` and the parent group is `nestedautogroupnested1`

**Result:**

`testuserx` is not able to login

**Expected Result:**

- Adding `nestedautogroupnested2` or `nestedautogroupnested1` to allowed groups should allow `testuserx` to login

**Other details that may be helpful:**

- users members of groups that are not nested are able to login
- `testuserx` is able to login with `Allow all valid users` site access.

Other detail here is the `nestedautogroupnested2` groupname is not shown in the UI if we refresh the site access page.
Non-nested groups are shown properly (`izaacgroup`)

![Screen Shot 2021-06-22 at 2 58 38 PM](https://user-images.githubusercontent.com/27659/123011412-2a585b00-d375-11eb-8282-b53a98dc89da.png)

**Environment information**
- Rancher version: `2.6 master-head (06/22/2021)` _d44698d_
- Installation option: HA RKE1

https://github.com/rancher/rancher/issues/30280

izaac:
Rancher version: `2.6 master-head (07/07/2021)` _cf5a75f_
Keycloak version: `3.3.0.Final`

I see two things happening on site access and nested groups.

## 1
First behavior is I add a nested group and then I reload the site access by either navigating to other Dashboard page or login back in. The nested group can't be retrieved.

Error response is:

`{"baseType":"error","code":"ServerError","message":"json: cannot unmarshal array into Go value of type keycloakoidc.account","status":500,"type":"error"}`


## 2
The second behavior, which is probably related to the above issue is that I can't login with users member of a group that has a parent group allowed to access Rancher

`nestedgroup1` -> Parent
`nestedgroup2` -> Child of `nestedgroup1`
`testuserx` is member of `nestedgroup2` and `nestedgroup1` is allowed to login from the Auth Site Access page.

Resulting in `testuserx` mnot able to login.

[...]

Issue #33314, opened 2021-06-21
Keycloak OIDC - Session for the currently logged admin is expiring before time making the user search to fail


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Login as Admin 
- Stay logged in for some time
- Enable Keycloak OIDC
- Try searching for users in the Ember Members input fields (Cluster Members, Project Members, Edit Cluster)

**Result:**
I am seeing these after some time (or if the admin has been logged for some time)  and try to do a user search from ember forms:

`user-65l85` is the local `admin`

```
2021/06/21 19:29:10 [DEBUG] Triggering auth refresh on user-65l85
2021/06/21 19:29:10 [DEBUG] Skipping refresh for user-65l85 due to max-age
```

API returns:

`{"baseType":"error","code":"ServerError","message":"invalid character 'B' looking for beginning of value","status":500,"type":"error"}`

**Expected Result:**

Session shouldn't expire before time and the search users fields should continue to work

**Other details that may be helpful:**

The search works again if I log out and log back in.
I also tried to reproduce this with OpenLDAP, the session stays and the user keeps

**Environment information**
- Rancher version: `2.6 master-head (06/21/2021)` _4d18400_
- Installation option: HA RKE1

Related issue: https://github.com/rancher/rancher/issues/33036

izaac:
Rancher version: `2.6 master-head (07/08/2021)` _cf5a75f_


Is the refresh token request on-demand?

Keycloak seems to have a 30 minutes `SSO Session Idle` of 30 minutes by default. 

If I stay longer that that time idle in the current logged in session without searching and then attempt a user/group search I hit the `500` described in this issue not only with admin but with any user that's logged in and idle.

Rancher has a `auth-user-session-ttl-minutes`  that defaults to `960` or 16 hours.

If I match Keycloak's `SSO Session Idle` with Rancher's session TTL I won't hit this issue. But then using defaults makes refresh token fix not relevant if it's on-demand if we hit the `SSO Session Idle` first we will end with a 500 request and the user will have to request a new session (log out and log in again).

Unless there's a way to request Keycloak's `Refresh Token` before we hit the `SSO Session Idle` the session Idle will be bumped automatically for the users instead of trying to request a new refresh token on-demand. As far as I understand from [Keycloak's Docs](https://wjw465150.gitbooks.io/keycloak-documentation/content/server_admin/topics/sessions/timeouts.html).

`Refresh token requests will also bump the idle timeout.`

[...]

PR #33580, opened 2021-07-13
[Keycloak oidc] Various Fixes 

brendarearden:
**Issues** 
https://github.com/rancher/rancher/issues/33337
https://github.com/rancher/rancher/issues/33314
https://github.com/rancher/dashboard/issues/3406 Scope needed to be space separated and allow for "openid" scope to be passed in instead of set in the client always. 
Certificate and Key were not being added to the context in all situations

**Solution**
Groups were not showing up on refresh because the Keycloak api relies on IDs, but we only have the name for groups. To resolve this, I am using the basic group search endpoint and then filtering down the result to only the value that has a matching name. group names are unique in keycloak, even if its a nested group. Note: since we only return 1 principal at a time for groups, we are not able to add all of the children groups if you add the parent group. 

With the newer version of Keycloak, scopes are now being validated - we were previously using comma separated scopes, but these were also being used in the UI (https://github.com/rancher/dashboard/pull/3434/files) so we switched to a space seperated string and I only add the openid scope value if it is not present

A new context was being created, but the old context was being passed around so the context with the updated values was never being used. I am now grabbing the new context and passing that value through to the other functions correctly. 

Added unauthorized error for when the token expires so that a proper error message is logged and there is no attempt in the client to parse the response.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -130,6 +130,9 @@ func getSubGroups(group Group) []Group {
 func (k *KeyCloakClient) getFromKeyCloakByID(principalID, principalType string, config *v32.OIDCConfig) (account, error) {
 	var searchResult account
 
+	if principalID == "" {
+		return account{}, fmt.Errorf("[keycloak oidc]: cannot perfom search with empty principalID")
+	}
 	sURL, err := getSearchURL(config.Issuer)
 	if err != nil {
 		return account{}, nil
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -64,29 +64,52 @@ func (k *KeyCloakClient) searchPrincipals(searchTerm, principalType string, conf
 		}
 	}
 	if principalType == "" || principalType == GroupType {
-		var groups []Group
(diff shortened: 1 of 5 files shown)


pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -64,31 +64,54 @@ func (k *KeyCloakClient) searchPrincipals(searchTerm, principalType string, conf
...
+		logrus.Errorf("[keycloak oidc]: GET request failed, got status code: %d. url: %s, err: %s",
+			statusCode, search, err)
+		return accounts, err
+	}
+	if err := json.Unmarshal(b, &groups); err != nil {
+		logrus.Errorf("[keycloak oidc]: received error unmarshalling search results, err: %v", err)
+		return accounts, err
+	}
+	for _, g := range groups {
+		accounts = append(accounts, account{ID: g.ID, Name: g.Name, Type: GroupType})
```
paynejacob, line 97: are groups limited to a depth of 2?

paynejacob (dismissed):

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: https://github.com/rancher/rancher/blob/122aa56e845eec8c0be37717556dd7e1ebebe16a/pkg/auth/providers/keycloakoidc/keycloak_client.go#L98 I added this limit so at most there would be a limit of 100 subgroups per group/subgroup

dramich (approved):

PR #33581, opened 2021-07-13
Remove legacy kev2 apps before install

rmweir:
**Problem:**
Prior, the old eks-operator app remained and deleting
it would cause the eks-operator deployment to get removed.
This caused both apps to be deleted and edits to no longer
work. Editing an EKS cluster would not cause the app to
redeploy either.

**Solution:**
Now, the old eks-operator app is deleted
before installing the new app. Old eks-operator app refers
to the app of type apps.project.cattle.io. Now, in v2.6+
an app of type apps.catalog.cattle.io is installed instead.

**Issue:**
https://github.com/rancher/rancher/issues/31592

### DIFF
--- a/pkg/controllers/dashboard/hostedcluster/controller.go
+++ b/pkg/controllers/dashboard/hostedcluster/controller.go
@@ -2,16 +2,22 @@ package hostedcluster
 
 import (
 	"context"
+	"fmt"
 	"os"
 
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/catalogv2/system"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/chart"
+	controllerv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
+	controllerprojectv3 "github.com/rancher/rancher/pkg/generated/controllers/project.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/namespace"
+	"github.com/rancher/rancher/pkg/project"
+	"github.com/rancher/rancher/pkg/ref"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/wrangler"
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cmurphy (dismissed):

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -108,6 +132,28 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
...
+	systemProject, err := project.GetSystemProject(localCluster, h.projectCache)
+	if err != nil {
+		return err
+	}
+
+	systemProjectID := ref.Ref(systemProject)
+	_, systemProjectName := ref.Parse(systemProjectID)
+
+	legacyOperatorAppName := fmt.Sprintf(legacyOperatorAppNameFormat, provider)
+	_, err = h.apps.Get(systemProjectName, legacyOperatorAppName, metav1.GetOptions{})
```
thedadams, line 145: Can we `Delete` here and check for `IsNotFound` error?

thedadams (changes requested):

pkg/controllers/dashboard/hostedcluster/controller.go
rmweir: @thedadams I think the best thing to do here is probably to get it off the cache, is that okay with you? Will update.

cmurphy (approved):

thedadams (approved):

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -108,6 +133,28 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
...
+	_, err = h.appCache.Get(systemProjectName, legacyOperatorAppName)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// legacy app doesn't exist, no-op
+			return nil
+		}
+		return err
+	}
+
+	return h.apps.Delete(systemProjectName, legacyOperatorAppName, &metav1.DeleteOptions{})
```
cbron, line 155: Summary from our chart: This will remove the system-chart, we need to be positive this won't delete or interrupt previous cluster objects/resources and that they get smoothly transitioned. Need to make that clear for QA to test.

cbron (approved):
lgtm as long as we communicate to QA

pkg/controllers/dashboard/hostedcluster/controller.go
rmweir: To be clear, deleting the chart does delete resources associated with it except for the CRD. It should be inconsequential since it does not delete the cluster and the resources will be installed with the new chart.

PR #33629, opened 2021-07-14
default the password to admin rather than a random value

paynejacob:
Randomizing the admin password makes it difficult to setup rancher.  Users can still specify the initial password when the setup rancher with `CATTLE_BOOTSTRAP_PASSWORD` or `bootstrapPassword` in the helm chart.


paynejacob:
this needs to be reworked

PR #33646, opened 2021-07-15
added CATTLE_BOOTSTRAP_PASSWORD to each docker/ha install of rancher

slickwarren:
master-head now requires a password on setup, which is effecting our automation

now, there is an argument for CATTLE_BOOTSTRAP_PASSWORD which is set in each of our cases


slickwarren:
closing in favor of https://github.com/rancher/rancher/pull/34646

PR #33650, opened 2021-07-15
[Validation] go test automation frame work

igomez06:
GO automation framework

Currently a single test provisioning a Digital ocean cluster. With a command to create a Digital Ocean Cloud credential. An initial architecture design encapsulating rancher rest commands and requests.


tests/go-validation/Dockerfile.v1api
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.16
```
izaac, line 1: Are minor golang versions usually stable or do we prefer to stay in an specific minor otherwise?

tests/go-validation/Dockerfile.v1api
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR $WORKSPACE/tests/go-validation
+
+COPY [".", "$WORKSPACE"]
+
+RUN cd $WORKSPACE/tests/go-validation && \
+    go get gotest.tools/gotestsum && \
+    go mod download
+
+RUN CGO_ENABLED=0
```
izaac, line 17: Nit: new line at EOL.

tests/go-validation/v1apitests/provisioning_test.go
```diff
@@ -0,0 +1,92 @@
...
+	clusterName := baseDOClusterName + utils.RandStringBytes(defaultRandStringLength)
+
+	generatedPoolName := fmt.Sprintf("nc-%s-pool1", clusterName)
+
+	machinePoolConfig := rancherrest.NewMachinePoolConfig(generatedPoolName, rancherrest.DOKind, namespace, rancherrest.DOPoolType, "ubuntu-20-04-x64", "sfo3", "s-2vcpu-4gb")
+
+	t.Logf("Creating DO machine pool config %s", generatedPoolName)
+	podConfigClient := client.NewPodConfigClient(rancherrest.DOResourceConfig)
+
+	machinConfigResult, err := podConfigClient.CreateMachineConfigPool(machinePoolConfig)
```
izaac, line 44: Should this be `machineConfigResult`?

tests/go-validation/v1apitests/provisioning_test.go
```diff
@@ -0,0 +1,92 @@
...
+	_, err = client.CreateCloudCredential(doCloudCred)
+	require.NoError(t, err)
+	t.Log("Cloud Credential was created successfully")
+
+	//randominize name
+	clusterName := baseDOClusterName + utils.RandStringBytes(defaultRandStringLength)
+
+	generatedPoolName := fmt.Sprintf("nc-%s-pool1", clusterName)
+
+	machinePoolConfig := rancherrest.NewMachinePoolConfig(generatedPoolName, rancherrest.DOKind, namespace, rancherrest.DOPoolType, "ubuntu-20-04-x64", "sfo3", "s-2vcpu-4gb")
```
izaac, line 39: If we're using `us-east-2` as default Rancher server region I think it would be best to default to DO's `ny3` to reduce latency.

tests/go-validation/rancherrest/machine_pool.go
```diff
@@ -0,0 +1,58 @@
...
+	machineConfig.Object["userdata"] = ""
+	return machineConfig
+}
+
+func (p *PodConfigClient) CreateMachineConfigPool(machinePoolConfig *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	podResult, err := p.Namespace(machinePoolConfig.GetNamespace()).Create(p.Ctx, machinePoolConfig, metav1.CreateOptions{})
+	return podResult, err
+}
+
+func MachinePoolSetup(controlPlaneRole, etcdRole, workerRole bool, poolName string, quantity int32, machineConfig *unstructured.Unstructured) apisV1.RKEMachinePool {
```
izaac, line 44: Should we add types for `controlPlaneRole` and `etcdRole` to be consistent?

tests/go-validation/v1apitests/provisioning_test.go
```diff
@@ -0,0 +1,92 @@
...
+	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	"github.com/rancher/rancher/tests/go-validation/rancherrest"
+	"github.com/rancher/rancher/tests/go-validation/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	digitalOceanCloudCredentialName = "docloudcredential"
+	namespace                       = "fleet-default"
```
izaac, line 17: do we need to specify the `fleet-default` namespace here? Just curious I know that the new API is using fleet for provisioning but im not familiar how is that working internally.

tests/go-validation/v1apitests/provisioning_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	t.Logf("Checking status of cluster %s", v1Cluster.ClusterName)
+	//check cluster status
+	ready, err := cluster.CheckClusterStatus(clusterName)
+	assert.NoError(t, err)
+	assert.True(t, ready)
+
+	if utils.RancherCleanup() {
+		err = ClearDigitalOceanCluster(client, doCloudCred, cluster, clusterName)
+		require.NoError(t, err)
```
izaac, line 76: Can there be a case where the `ClearCluster` removes the cluster from Rancher but still the resources are not removed? Can we wait somehow to check the resources are actually cleared? 

I'm curious because removing manually from Dashboard the reference gets almost immediately removed but the resource removal takes more time to clear the resources.

izaac (changes requested):
A couple questions and nit picks here and there. Overall looks good.

tests/go-validation/build.sh
```diff
@@ -0,0 +1,30 @@
...
+set -x
+set -eu
+
+DEBUG="${DEBUG:-false}"
+RKE_VERSION="${RKE_VERSION:-v1.0.2}"
+RANCHER_HELM_VERSION="${RANCHER_HELM_VERSION:-v3.6.0}"
+KUBECTL_VERSION="${KUBECTL_VERSION:-v1.16.6}"
+CLI_VERSION="${CLI_VERSION:-v2.4.5}"
+SONOBUOY_VERSION="${SONOBUOY_VERSION:-0.18.2}"
+TERRAFORM_VERSION="${TERRAFORM_VERSION:-0.12.10}"
```
git-ival, line 12: It looks like not all of these env vars are used, we may want to remove any unused ones. We should also verify that these are the minimum versions we should be targetting.

tests/go-validation/Dockerfile.v1api
igomez06: It depends. 1.16 is the latest most stable version right now. 1.17 is currently unstable. Stable releases is something we should keep an eye on, in case we want to upgrade. For example versions after 1.13 only support module dependency while versions before would use a vendor file for dependencies.

tests/go-validation/v1apitests/provisioning_test.go
igomez06: yes it should be

tests/go-validation/v1apitests/provisioning_test.go
igomez06: Absolutely I'll make the change

tests/go-validation/rancherrest/machine_pool.go
igomez06: What do you mean Izaac? Have different setup methods?

tests/go-validation/v1apitests/provisioning_test.go
igomez06: Yes we do, we have specify "fleet-default"

tests/go-validation/v1apitests/provisioning_test.go
igomez06: We can do that I can have different clean up methods for cleaning up resources vs clusters. I did it this way, because I create a cloud credential before every test. I'll do whatever makes the most sense. I can have a check to see if a cloud cred exists beforehand, if that works better.

igomez06: In terms of actually checking the resources like DO, maybe that's a future thing ... but I'll do whatever the team wants. I'm sure I would need to setup a DO api client for example

tests/go-validation/build.sh
igomez06: I think we should just remove them for now, since I'm not using these en vars at all.

tests/go-validation/v1apitests/provisioning_test.go
igomez06: I also think something we should think about is creating enums for region, and others as well, so we don't have raw strings.

tests/go-validation/Dockerfile.v1api
igomez06: fixed

tests/go-validation/v1apitests/provisioning_test.go
igomez06: fixed

tests/go-validation/build.sh
igomez06: removed them

dramich:
@igomez06 Add me back on after you have gone through peer reviews first

tests/go-validation/v1apitests/provisioning_test.go
git-ival: My 2 cents: I agree that we would want to verify that the resources are actually removed, but I think it's fine as-is for a first pass.

tests/go-validation/rancherrest/machine_pool.go
izaac: I'm just noting we're using types explicitly for `workerRole`, `poolName` and not for `controlPlaneRole` and `etcdRole`

tests/go-validation/Dockerfile.v1api
izaac: Makes sense.

izaac (approved):
LGTM. Architecturally looks good but I would rely on a more experience review of the golang code.

git-ival (dismissed):
LGTM as well. I agree with Izaac, it would be good to get a review from someone more experienced with golang to comment on the code quality and structure.

tests/go-validation/rancherrest/machine_pool.go
igomez06: Ah I see what you're saying, sorry. This is a go convention thing where if you have multiple parameters with the same type you don't need to explicitly state the type for every one.

izaac: Ah!! conventions, Ok i get it.

tests/go-validation/v1apitests/provisioning_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	t.Logf("Creating DO machine pool config %s", generatedPoolName)
+	podConfigClient := client.NewPodConfigClient(rancherrest.DOResourceConfig)
+
+	machineConfigResult, err := podConfigClient.CreateMachineConfigPool(machinePoolConfig)
+	require.NoError(t, err)
+
+	t.Logf("Successfully created DO machine pool %s", generatedPoolName)
+
+	provistioningClient, err := rancherrest.NewProvisioningClient()
```
git-ival, line 49: Shouldn't this be "provisioningClient"?

igomez06: Yes hahaha. Typo, I'll fix it

jiaqiluo (approved):
lgtm

bmdepesa:
Closing in favor of #33901

Issue #33404, opened 2021-06-28
Passing --address and --internal-address while creating custom cluster is not set on the node.


ShylajaDevadiga:
Build from master-7a2fc74

Describe the issue:
--address and --internal-address is not stored. in config and therefore not set on the node. While passing --node-name is stored in config and set to the node.

Steps to reproduce:

- Click on Cluster Management->Create-Custom
- Enter Name to the the cluster-> Create
- Click on the name of the cluster in the next window
- Click on Show Advanced
- Pass Node-name, Node Public IP,  Node Private IP
- Copy the registration command and execute it on the node.
```
curl --insecure -fL https://<IP>/system-agent-install.sh | sudo  sh -s - --server https://<IP> --token <TOKEN> --ca-checksum <CHECKSUM> --etcd --controlplane --worker --address <Public IP> --internal-address <Private IP> --node-name shysrv1
```
Config yaml output
```
{
  "agent-token": "<TOKEN>",
  "cni": "calico",
  "etcd-snapshot-retention": 5,
  "etcd-snapshot-schedule-cron": "* */5 * * *",
  "node-label": [
    "rke.cattle.io/machine=e876f39c-7a92-4a2d-aa46-18e04862a88d"
  ],
  "node-name": "shysrv4",
  "server": "https://<SERVER PRIVATE IP>:9345",
  "token": "<TOKEN>"
}
```
Confirmed from the debug results, headers are correctly passed on correctly
```
X-Cattle-Node-Name: shysrv1 -H X-Cattle-Address: <PUBLIC IP> -H X-Cattle-Internal-Address: <PRIVATE IP>
```

**Other details that may be helpful:**
Node-name is set on the node as seen the kubectl output
```
kubectl get node
NAME      STATUS   ROLES                       AGE   VERSION
shysrv1   Ready    control-plane,etcd,master   86m   v1.21.2+rke2r1
```

**Environment information**
- Rancher version : 7a2fc74
- Installation option : Single/HA

**Cluster information**
- Cluster type : Custom
- Machine type : AWS

**Additional info:**
On DO InternalIP is set to Public IP
Describe the node
```
etcd.k3s.cattle.io/node-address: <Public IP>
InternalIP:  <Public IP>
```

thedadams:
Leaving a note here: I have an implementation of this. I need to update the new provisioning tests when I return to complete this issue.

catastrophe-brandon:
After running the registration command I see the following (`b395238`):

[...]

PR #33653, opened 2021-07-16
Add node-ip and node-eternal-ip to config

thedadams:
If the --address and/or --internal-address flags were passed to the
system-agent-installer for a custom node, then result was ignored by
Rancher.

After this change, Rancher will look for these values in the config
secret and pass them to the config when bootstrapping the node.

Issue:
https://github.com/rancher/rancher/issues/33404

### DIFF
--- a/pkg/provisioningv2/rke2/configserver/custom.go
+++ b/pkg/provisioningv2/rke2/configserver/custom.go
@@ -35,13 +35,7 @@ func (r *RKE2ConfigServer) findMachineByClusterToken(req *http.Request) (string,
 		return "", "", err
 	}
 
-	data := map[string]interface{}{}
-	for k, v := range req.Header {
-		if strings.HasPrefix(k, headerPrefix) {
-			data[strings.ToLower(strings.TrimPrefix(k, headerPrefix))] = v
-		}
-	}
-	data["id"] = machineID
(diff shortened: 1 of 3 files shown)


pkg/provisioningv2/rke2/configserver/custom.go
```diff
@@ -35,13 +35,7 @@ func (r *RKE2ConfigServer) findMachineByClusterToken(req *http.Request) (string,
 		return "", "", err
 	}
 
-	data := map[string]interface{}{}
-	for k, v := range req.Header {
-		if strings.HasPrefix(k, headerPrefix) {
-			data[strings.ToLower(strings.TrimPrefix(k, headerPrefix))] = v
-		}
-	}
-	data["id"] = machineID
```
luthermonson, line 44: still need to set the machineid?

thedadams: It was already being set above: the header is `X-Cattle-Id` which is handled by the `TrimPrefix` and `ToLower` calls.

luthermonson: Gtg then

luthermonson (approved):

ibuildthecloud (approved):

Issue #32414, opened 2021-04-27
v2 Provisioning: RBAC - Support standard user


deniseschannon:
Currently, RKE2 provisioning works for admin, but needs to extend support for standard user.

We should also test that cluster owners/members or project owners/members don't have any issues with provisioning or using clusters/projects and the resources inside. 

This needs to be tested with the feature flag on/off for import and custom cluster use cases to confirm that the workflow is right depending on the RKE2 feature flag.

deniseschannon:
Rancher side of creating CRTBs has been merged, but there are still some webhook validation issues as well as these issues:

* If you create the cluster and are not the admin, you don't have privileges to the cluster as the creatorID is not associated with the user. 
* Fields in the cluster can reference objects that a user don't have access to like secrets, nodepoolconfig, cluster name fields.

There should be no additional work around PRTBs as project stuff is isolated after a cluster IS provisioned and isn't affected by the new provisioning framework.

dramich:
Changes made/required:

- [x] Add permissions for a standard user to created provisioning cluster and rke-machine-config
- [ ] Validate get permissions on referenced secret/cloud cred on provisioning cluster create
- [ ] Give user permissions to get/update/delete the cloud cred they create (this goes into a generic namespace so perms have to be per user for their specific secret)

dramich:
Opening new issues for things that still need to be finished, but moving this to QA so they can get in on the RBAC party. 
https://github.com/rancher/rancher/issues/33997
https://github.com/rancher/rancher/issues/33998

dramich:
There is a current bug with UI that users can't always see the provisioning cluster in cluster management. This can be verified by going to /v1/provisioning.cattle.io and if the cluster shows then it's the UI bug. 
https://github.com/rancher/dashboard/issues/3626

sowmyav27:
**On 2.6-head commit id: `b43e4d9`**

- Validated usecases for custom rke2, import clusters and nodepool rke2 clusters for Cluster owner, Cluster member, Project owner and project member. 
- Validated usecases when RKE2 flag was disabled for these users.
- Separate bugs logged for issues found.
- Closing this issue

PR #33663, opened 2021-07-16
Add secret controller for cloud cred

dramich:
Problem:
The secret created for cloud creds is created in a namespace a user
doesn't normally have access to

Solution:
Add permissions for the secret in the namespace based off the creatorID
anno the webhook will add when the secret is created

Reordered user roles by apiGroup and added the perm `addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch").` which is required for provisioning v2 cluster templates

https://github.com/rancher/rancher/issues/32414

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -96,7 +96,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
 	github.com/rancher/aks-operator v1.0.1-rc14
-	github.com/rancher/apiserver v0.0.0-20210519053359-f943376c4b42
+	github.com/rancher/apiserver v0.0.0-20210716201731-fcd925c99ba8
 	github.com/rancher/channelserver v0.5.1-0.20210618172430-5cbefd383369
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -96,7 +96,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
 	github.com/rancher/aks-operator v1.0.1-rc14
-	github.com/rancher/apiserver v0.0.0-20210519053359-f943376c4b42
+	github.com/rancher/apiserver v0.0.0-20210716201731-fcd925c99ba8
```
cmurphy, line 99: This change is bringing in something that breaks the UI for non-admin users. UI console error is `Loading error Error: Cannot read property 'startsWith' of undefined`, rancher log error `[ERROR] Error during subscribe websocket: close sent`

dramich: Yup, it's intentional. UI has some work to do now.

cmurphy (approved):

PR #33680, opened 2021-07-19
Add Secrets OnChange handler to AKS controller

cmurphy:
Add a handler to watch Azure cloud credential secrets and inject a
tenantID if one does not exist. The handler uses rancher/machine and the
subscription ID to look up the tenantID from Azure. This makes providing
the tenant ID optional.

Depends on https://github.com/rancher/machine/pull/138

https://github.com/rancher/rancher/issues/33097


pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -67,9 +73,36 @@ func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.M
 		DynamicClient:        aksCCDynamicClient,
 		ClientDialer:         mgmtCtx.Dialer,
 		Discovery:            wContext.K8s.Discovery(),
-	}}
+	},
+		wContext.Core.Secret(),
```
rmweir, line 77: Write out the field name for the struct now that it isn't just one embedded field, will be easier to read.

rmweir (changes requested):

pkg/controllers/management/aks/aks_cluster_handler.go
cmurphy: done

rmweir (dismissed):
LGTM

go.mod
```diff
@@ -45,9 +46,9 @@ replace (
 )
 
 require (
-	github.com/Azure/azure-sdk-for-go v50.0.1-0.20210114072321-4a06a7dc9c3c+incompatible
-	github.com/Azure/go-autorest/autorest v0.11.16
-	github.com/Azure/go-autorest/autorest/adal v0.9.11-0.20210111195520-9fc88b15294e
+	github.com/Azure/azure-sdk-for-go v55.7.0+incompatible
```
cbron, line 49: We should do a quick scan to make sure this bump doesn't have wider impact.

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -53,23 +59,51 @@ func Register(ctx context.Context, wContext *wrangler.Context, mgmtCtx *config.M
...
+	if ok && string(tenantIDBytes) != "" {
+		return cred, nil
+	}
+	tenantID, err := azureutil.FindTenantID(context.Background(), azure.PublicCloud, string(subscriptionID))
+	if err != nil {
+		return cred, err
+	}
+	cred = cred.DeepCopy()
+	cred.Data["azurecredentialConfig-tenantId"] = []byte(tenantID)
+	return e.secretClient.Update(cred)
```
cbron, line 106: Would be curious if UI could create a race condition here. Like user creates it and tries to use it before tenant exists. Talking offline and we are assuming not.

cbron (approved):
Needs machine change obviously.

cbron (commented):
Meant to just comment, need to change machine still. Otherwise lgtm.

cmurphy:
merged rancher/machine change, updated go.mod

cbron (approved):

Issue #33144, opened 2021-06-11
Manually created duplicate rolebindings are not cleaned up when Rancher is upgraded to 2.5.8


anupama2501:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
Before upgrade rancher version: v2.5.6
After rancher upgrade version: v2.5.8

1. On v2.5.6, create a node driver DO cluster with 3 nodes with 1 role in each of the nodes (cp, worker, etcd)
2. Create a user1 from global --> security --> users
3. Create a project P1 in the cluster created above and add user1 as a project owner and project member in P1.
4. By executing the command grab the role template of the role binding for the above user for project member

```
apiVersion: "rbac.authorization.k8s.io/v1"
kind: RoleBinding
metadata:
 labels:
  "cattle.io/creator": norman
  "p-krzs2_prtb-vpn2999": "membership-binding-owner"
 name: "rolebinding-k48nc"
 namespace: "c-gh6nj"
roleRef:
 apiGroup: "rbac.authorization.k8s.io"
 kind: Role
 name: "p-krzs2-projectmember"
subjects:
- apiGroup: "rbac.authorization.k8s.io"
  kind: User
  name: "u-9bj7w"
```
5. Create 10k duplicates for the above rb by changing the name ` name: "rolebinding-k48nc"`
6. verify the number of duplicates created are more than or equal to 10k 
7.  Upgrade rancher server to the version v2.5.8, and the manually created duplicates are not deleted:

From Rancher server logs:
```
2021/06/10 21:15:54 [INFO] Total PRTB duplicate clusterRoleBindings 1, roleBindings 6
2021/06/10 21:15:54 [INFO] Done Duplicate CRB and RB cleanup
2021/06/10 21:15:54 [INFO] Rancher startup complete
```

Duplicates on the cluster:
```
DEBU[0000] rb: kube-system/system::leader-locking-kube-scheduler has multiple subjects
DEBU[0001] crb: /system:public-info-viewer has multiple subjects
DEBU[0001] groups of duplicate role bindings: 1
DEBU[0001] total duplicate role bindings: 10436
DEBU[0001] no dupe crbs found
```

**Result:**
The duplicates that were created manually should be deleted

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.5.8
- Installation option (single install/HA): Single

**Cluster information**
[...]

PR #33681, opened 2021-07-19
CRBs/RBs cleanup fix

snasovich:
On certain setups, initial CRBs/RBs cleanup runs just after additional Cluster Role Template Bindings are created. They do not immediately have a label indicating they are from Rancher v2.5+ (it's set by a controller so it happens asynchronously). This results in duplicates queried by pre-v2.5 label style, so no duplicates are found in these cases.
As a fix, we're adding a 3-minute delay to starting actual cleanup. It virtually eliminates a chance of running into this issue and moves deduplication later helping ease load on initial Rancher startup after the upgrade. 3 minutes was chosen as it seems "not too short, not too long".

While at it, added extra debug logging to help troubleshoot any other potential issues in deduplication code as well as fixed error handling bug (`multierror.Append` returns aggregated error so it needs to be reassigned).

For https://github.com/rancher/rancher/issues/33144

### DIFF
--- a/pkg/agent/clean/binding.go
+++ b/pkg/agent/clean/binding.go
@@ -144,6 +144,7 @@ func (bc *bindingsCleanup) clean() error {
 }
 
 func (bc *bindingsCleanup) cleanCRTB(newLabel bool, crtbs []apiv3.ClusterRoleTemplateBinding) error {
+	logrus.Debugf("Cleaning up duplicates for %v CRTBs", len(crtbs))
 	var objectMetas []metav1.ObjectMeta
 	for _, crtb := range crtbs {
 		objectMetas = append(objectMetas, crtb.ObjectMeta)
@@ -153,6 +154,7 @@ func (bc *bindingsCleanup) cleanCRTB(newLabel bool, crtbs []apiv3.ClusterRoleTem
 }
 
 func (bc *bindingsCleanup) cleanPRTB(newLabel bool, prtbs []apiv3.ProjectRoleTemplateBinding) error {
+	logrus.Debugf("Cleaning up duplicates for %v PRTBs", len(prtbs))
 	var objectMetas []metav1.ObjectMeta
 	for _, prtb := range prtbs {
 		objectMetas = append(objectMetas, prtb.ObjectMeta)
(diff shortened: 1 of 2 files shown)


snasovich:
This is targeting v2.6.1 but we may want to get it into v2.6.0 as well depending on the timing. That's why it's not draft.

prachidamle (commented):
LGTM

prachidamle (dismissed):

pkg/agent/clean/binding.go
```diff
@@ -171,36 +173,40 @@ func (bc *bindingsCleanup) cleanObjectDuplicates(bindingType string, newLabel bo
 	for _, meta := range objMetas {
 		labels := createLabelSelectors(newLabel, meta, bindingType)
 		for _, label := range labels {
+			logrus.Debugf("%v %v label:%v, Checking CRB/RB duplicates...", bindingUpper, meta.Name, label)
```
dramich, line 176: Can this log be switch around along with the one below 
```suggestion
			logrus.Debugf("Checking CRB/RB duplicates for: %v %v label:%v", bindingUpper, meta.Name, label)
```

pkg/data/management/bindings_cleanup.go
```diff
@@ -23,10 +24,13 @@ func CleanupDuplicateBindings(scaledContext *config.ScaledContext, wContext *wra
 		// config map already exists, check if the cleanup key is found
 		if _, ok := adminConfig.Data["DedupeBindingsDone"]; ok {
 			//cleanup has been run already, nothing to do here
-			logrus.Info("bindings cleanup already ran before, not calling again")
+			logrus.Info("Bindings cleanup already ran before, not calling again")
 			return
 		}
-		// run cleanup
+		// run cleanup after delay to greatly reduce the chance of acting of not updated CRTBs/PRTBs
```
dramich, line 30:
```suggestion
		// run cleanup after delay to give other controllers a chance to create CRTB/PRTBs and ease the load on the API at startup
```

pkg/agent/clean/binding.go
snasovich: I was going for consistency with the existing log message (well, with added comma to help separate things).
```
logrus.Infof("%v %v label:%v Duplicates: CRB:%v RB:%v", bindingUpper, meta.Name, label, CRBduplicates, RBDupes)
```

So we can either keep this one as is or change all 3 log messages . I'm fine either way, let me know what you would prefer.

P.S. On a side note, it would be great if we did structured logging.

pkg/data/management/bindings_cleanup.go
snasovich: Fixed, thank you for the suggestion

pkg/agent/clean/binding.go
dramich: Switch the calls to use logrus.WithFields if you want some structure to these, which would probably help. The original logs were mainly for the job and consumption through there so I would prefer them to be easier to read now that this runs inside rancher as well. So log message first then data is my preference.

snasovich: That would be odd to have a single place where we're using structured logging, so I've switched the messages as you suggested with the actual message first followed by the data.

dramich (approved):

snasovich:
Following the discussion with QA converted to draft as the target for this is indeed 2.6.1 so we don't want it merged yet.

Issue #33694, opened 2021-07-12
Unable to create custom rke2 cluster with specific names


kravciak:
**Bug description**

I am not able to create custom rke2 cluster with specific names.
After creation screen is stack at `Loading...`, trying to access it from cluster list has the same result.
```
# UI error
Clusters.management.cattle.io "mkravec-oracle" not found

# rancher container log
[ERROR] error syncing 'fleet-default/mkravec-oracle': handler cluster-create: clusters.management.cattle.io "mkravec-oracle" not found, requeuing
```

![Screenshot from 2021-07-12 13-57-21](https://user-images.githubusercontent.com/538604/125283899-26d34680-e319-11eb-8e22-8f0adcc930d4.png)

**How to reproduce**

Dashboard -> Cluster management -> Create (rke2) -> Custom -> `name` -> Create
```bash
# does not work with: 
mkravec-oracle
mkravec-xxxxxx
xxxxxxc-xxxxxx
xxxxxc-xxxxx

# works fine with
xxxxxz-xxxxx
xxxxc-xxxx
```

**Rancher setup**
kubernetes: v1.21.2+rke2r1
rancher: single node docker 
version: master-head (3ae6f43)

kravciak:
This does not seem right, I can create `local` cluster, but not `c-local`.

![Screenshot from 2021-07-29 21-55-11](https://user-images.githubusercontent.com/538604/127560384-ced1463d-2649-4efa-86cb-b2d46fc8dc06.png)

Tested on master-head (37561eb)

deniseschannon:
Since we expect "local" and "c-XXXXX" cluster names to be reserved for clusters created in the provisioning v1 framework, this will currently cause issues. 

For now, can we add a UI check to not allow those names? @vincent99

kravciak:
Could we also add some information about name reservations and error message?
Currently screen is stack at `Loading...` for 30s and then returns `clusters.management.cattle.io "c-xxxxx" not found` error.

deniseschannon:
Created this UI ticket for the UI validation: https://github.com/rancher/dashboard/issues/3585

@kravciak By adding in UI validation, the error message should indicate what names can't be used.

ronhorton:
I was able to create custom clusters with the provided names. I did not receive the error; I also went on to register my clusters and see them become active. In addition, I created an infrastructure provider cluster using a provided name and did not get an error.

thedadams:
[...]

PR #33695, opened 2021-07-20
Do not match substrings of cluster name

ibuildthecloud:
https://github.com/rancher/rancher/issues/33694

### DIFF
--- a/pkg/controllers/provisioningv2/cluster/controller.go
+++ b/pkg/controllers/provisioningv2/cluster/controller.go
@@ -37,7 +37,7 @@ const (
 )
 
 var (
-	mgmtNameRegexp = regexp.MustCompile("c-[a-z0-9]{5}|local")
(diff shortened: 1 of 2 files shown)


pkg/controllers/provisioningv2/cluster/controller.go
```diff
@@ -37,7 +37,7 @@ const (
 )
 
 var (
-	mgmtNameRegexp = regexp.MustCompile("c-[a-z0-9]{5}|local")
+	mgmtNameRegexp = regexp.MustCompile("^c-[a-z0-9]{5}|local$")
```
kravciak, line 40: I guess this matches `^c-[a-z0-9]{5}` and `local$`. We want `^c-[a-z0-9]{5}$` and `^local$`, right?

```suggestion
	mgmtNameRegexp = regexp.MustCompile("^(c-[a-z0-9]{5}|local)$")
```

ibuildthecloud: good catch, thank!

Issue #33474, opened 2021-07-02
Can't download kubeconfig from downstream clusters when kubeconfig-generate-token is set to false


izaac:
**What kind of request is this:**
bug

**Steps to reproduce:**

- Install Rancher 2.6 `master-head` 
- Provision a downstream cluster (I used RKE1 DO)
- Set `Global Settings` -> `kubeconfig-generate-token` to `false`
- Go to Cluster Management and from the downstream cluster click `Download Kubeconfig`

**Result:**

Nothing happens on the UI but the API returns

`{"baseType":"error","code":"ServerError","message":"resource name may not be empty","status":500,"type":"error"}`

**Expected Result:**

kubeconfig YAML downloads successfully

**Other details that may be helpful:**

- This is not reproducible on 2.5.8
- This is reproducible from Dashboard and Ember

**Environment information**
- Rancher version: `2.6 master-head (07/02/2021)` _82b212d_
- Installation option: RKE1 (1.3.0-rc8) HA

Related issue: https://github.com/rancher/dashboard/issues/3044

sowmyav27:
**On master-head commit id: `2e6926b`**

- Install Rancher master-head
- Provision a downstream cluster - RKE1 DO
- Set Global Settings -> kubeconfig-generate-token to false
- Go to Cluster Management and from the downstream cluster click Download Kubeconfig 
- User is able to download and use the kubeconfig file

PR #33696, opened 2021-07-20
Do not request cluster token if token setting is false

StrongMonkey:
We can not request and create cluster token if cluster local endpoint is
enable but generating token setting is false. The token will be
initialized through CLI.

https://github.com/rancher/rancher/issues/33474


rmweir (approved):
LGTM

dramich (approved):

PR #33699, opened 2021-07-20
Automation - Change vm type for AKS

izaac:
Just bump the vm type to get more resources for the AKS local Rancher cluster.


sowmyav27 (approved):
LGTM

PR #33702, opened 2021-07-20
Check hash annotation

rmweir:


rmweir:
Issue has been created for making it so the feature flag is not one way and can be toggled: https://github.com/rancher/rancher/issues/33703.

dramich (approved):

PR #33719, opened 2021-07-21
[keycloak oidc] Add nested group support, fix 500 error on nil response

brendarearden:
**Issue(s)**
https://github.com/rancher/rancher/issues/33337
https://github.com/rancher/rancher/issues/33314
https://github.com/rancher/rancher/issues/33782

**Problems**
1. A user's groups do not include parent groups
2. When token is expired,  the resp from a get request is nil so panic occurs when resp.statuscode is hit
3. Scope has empty scopes in it occasionally causing invalid scope error
4. Logging vs error messages were very inconsistent

**Solution**
1. Add support for additional mapper to be added to the keycloak client, so when the userInfo is returned, the full path results are included in the response. The full paths are then parsed and included in a user's group principals
2. The statusCode was not being used in a meaningful way, so I removed it. The error that is returned includes error code. 
3. Remove empty values from scope slice
4. Made logging and error messages consistent within keycloak oidc and generic oidc

**Design Discussion**
Two designs were discussed for the nested group fix. 
- Adding a mapper to the keycloak client so that the full group path would be returned on the oidc userInfo endpoint and the full path would be parsed to get the parent groups
- Adding a AdminServiceAccount, whose credentials would be used to call the keycloak /users/{userid}/groups endpoint to get the group information and the full path would be parsed to get the parent groups

After additional discussion with original submitters of the PRD, the user who is configuring the keycloak oidc auth provider in rancher will have access to manage the client in keycloak. This means it is preferable to go the mapper route as Non Person Entity Accounts are not the preferable implementation. If this needs to undergo further discussion in a larger design meeting, we can delay the nested group implementation till after 2.6.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -48,10 +48,9 @@ func (k *KeyCloakClient) searchPrincipals(searchTerm, principalType string, conf
 		searchURL := fmt.Sprintf("%s/%ss?search=%s", sURL, UserType, searchTerm)
 		search := URLEncoded(searchURL)
 
-		b, statusCode, err := k.getFromKeyCloak(search)
+		b, err := k.getFromKeyCloak(search)
 		if err != nil {
-			logrus.Errorf("[keycloak oidc]: GET request failed, got status code: %d. url: %s, err: %s",
-				statusCode, search, err)
+			logrus.Errorf("[keycloak oidc]: GET request failed. url: %s, err: %s", search, err)
 			return accounts, err
 		}
 		if err := json.Unmarshal(b, &userAccounts); err != nil {
@@ -83,10 +82,9 @@ func (k *KeyCloakClient) groupSearch(searchTerm string, sURL string) ([]account,
 	searchURL := fmt.Sprintf("%s/%ss?search=%s", sURL, GroupType, searchTerm)
 	search := URLEncoded(searchURL)
 
-	b, statusCode, err := k.getFromKeyCloak(search)
+	b, err := k.getFromKeyCloak(search)
 	if err != nil {
-		logrus.Errorf("[keycloak oidc]: GET request failed, got status code: %d. url: %s, err: %s",
-			statusCode, search, err)
+		logrus.Errorf("[keycloak oidc]: GET request failed. url: %s, err: %s", search, err)
 		return accounts, err
 	}
 	if err := json.Unmarshal(b, &groups); err != nil {
@@ -141,10 +139,9 @@ func (k *KeyCloakClient) getFromKeyCloakByID(principalID, principalType string, 
 	if principalType == UserType {
 		searchURL := fmt.Sprintf("%s/%ss/%s", sURL, principalType, principalID)
 		search := URLEncoded(searchURL)
-		b, statusCode, err := k.getFromKeyCloak(search)
+		b, err := k.getFromKeyCloak(search)
 		if err != nil {
-			return account{}, fmt.Errorf("[keycloak oidc]: GET request failed, got status code: %d. url: %s, err: %s",
-				statusCode, search, err)
+			return account{}, fmt.Errorf("[keycloak oidc]: GET request failed. url: %s, err: %s", search, err)
 		}
 		if err := json.Unmarshal(b, &searchResult); err != nil {
 			logrus.Errorf("[keycloak oidc]: received error unmarshalling search results, err: %v", err)
@@ -184,31 +181,31 @@ func URLEncoded(str string) string {
 	return u.String()
 }
 
-func (k *KeyCloakClient) getFromKeyCloak(url string) ([]byte, int, error) {
+func (k *KeyCloakClient) getFromKeyCloak(url string) ([]byte, error) {
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
-		return nil, 500, err
+		return nil, err
 	}
 	req.Header.Add("Accept", "application/json")
 	resp, err := k.httpClient.Do(req)
 	if err != nil {
 		logrus.Errorf("[keycloak oidc]: received error from keycloak: %v", err)
-		return nil, resp.StatusCode, err
+		return nil, err
 	}
 	defer resp.Body.Close()
 	b, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return b, resp.StatusCode, err
+		return b, err
 	}
 	switch resp.StatusCode {
 	case 200:
 	case 201:
 	case 403:
-		return b, resp.StatusCode, httperror.NewAPIError(httperror.PermissionDenied, "access denied")
+		return b, httperror.NewAPIError(httperror.PermissionDenied, "access denied")
 	case 401:
-		return b, resp.StatusCode, httperror.NewAPIError(httperror.Unauthorized, "invalid token")
+		return b, httperror.NewAPIError(httperror.Unauthorized, "invalid token")
 	default:
-		return b, resp.StatusCode, err
+		return b, err
 	}
-	return b, resp.StatusCode, nil
+	return b, nil
 }
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -184,31 +181,31 @@ func URLEncoded(str string) string {
...
+		return nil, err
 	}
 	defer resp.Body.Close()
 	b, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return b, resp.StatusCode, err
+		return b, err
 	}
 	switch resp.StatusCode {
 	case 200:
```
paynejacob, line 201: why have these if they are handled by the default case?

paynejacob (dismissed):

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -413,18 +392,47 @@ func (o *OpenIDCProvider) getUserInfo(ctx *context.Context, config *v32.OIDCConf
...
 		Endpoint:     endpoint,
 		RedirectURL:  config.RancherURL,
-		Scopes:       configScopes,
+		Scopes:       finalScopes,
+	}
+}
+
+func (o *OpenIDCProvider) getGroupsFromClaimInfo(claimInfo claimInfo) []v3.Principal {
+	var groupPrincipals []v3.Principal
+
```
prachidamle, line 418: Is there a possibility of nil claimInfo?

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -382,26 +370,17 @@ func (o *OpenIDCProvider) getUserInfo(ctx *context.Context, config *v32.OIDCConf
...
 			return userInfo, oauth2Token, err
 		}
-	}
-
-	rawToken, ok := oauth2Token.Extra("id_token").(string)
-	if !ok {
-		rawToken, ok = oauth2Token.Extra("access_token").(string)
-		if !ok {
+		_, err = verifier.Verify(updatedContext, oauth2Token.AccessToken)
+		if err != nil {
```
prachidamle, line 380: Why are we replacing this code of reading oauth2Token.Extra("access_token").(string)? Was this oauth2Token.AccessToken always present?

pkg/auth/providers/keycloakoidc/keycloak_client.go
brendarearden: This error catch is for the errors with the ReadAll function. The switch uses the response code, if this error catch was not here, it would be possible for the ReadAll function to return an error because it had issues reading the body, but the response code could be 200, in which case the error would get eaten.

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: The struct does not have "omitEmpty" on any of the fields, so at most it would be a bunch of empty values. I check the error when unmarshaling values into the struct so that should keep it from being nil.

pkg/auth/providers/oidc/oidc_provider.go
brendarearden: TLDR; It was alway present so it seems like a cleaner implementation

The example provided from the go-oidc library shows this: 

```
// Extract the ID Token from OAuth2 token.
    rawIDToken, ok := oauth2Token.Extra("id_token").(string)
    if !ok {
        // handle missing token
    }

    // Parse and verify ID Token payload.
    idToken, err := verifier.Verify(ctx, rawIDToken)
    if err != nil {
        // handle error
    }
```
which is what I did initially for the access token. But, as I was refactoring for nested groups, I noticed that if you look at the struct of the oauth2Token, it does have a field for AccessToken, pulling from the access_token json key: 

```
type Token struct {
	// AccessToken is the token that authorizes and authenticates
	// the requests.
	AccessToken string `json:"access_token"`

	// TokenType is the type of token.
	// The Type method returns either this or "Bearer", the default.
	TokenType string `json:"token_type,omitempty"`

	// RefreshToken is a token that's used by the application
	// (as opposed to the user) to refresh the access token
	// if it expires.
	RefreshToken string `json:"refresh_token,omitempty"`

	// Expiry is the optional expiration time of the access token.
	//
	// If zero, TokenSource implementations will reuse the same
	// token forever and RefreshToken or equivalent
	// mechanisms for that TokenSource will not be used.
	Expiry time.Time `json:"expiry,omitempty"`

	// raw optionally contains extra metadata from the server
	// when updating a token.
	raw interface{}
}
```
so I thought it was a cleaner implementation. We are also not using the id_token at any point, so I removed that check all-together.

prachidamle: Thank you for the explanation :)

prachidamle (approved):

dramich:
@brendarearden Is there a docs issue related to this that explains the changes required on the auth provider side to make this work? And does this cover the case a user does not enable that functionality (do we fail gracefully/fall back to current no nesting logic)

brendarearden:
@dramich we do have a docs issue and pr currently open. I have added a comment on the docs instructions to make sure the additional mappers are included in the setup steps: https://github.com/rancher/docs/pull/3358/files#r675247601. Do you think the docs need to cover the different scenarios I have listed below as well, so that they understand how it fails?  


### In terms of failing gracefully: 

**If the Group Mapper or the Full Path Group mapper are not added:**

There will be no group principals associated with any user from keycloak. 
The user that is setting up the auth provider will be able to login. 
Users with access to search will still be able to search users AND groups, but adding groups will not authorize others to login. The user would have to add specific user principals for others to be able to login. 

**If only the Group Mapper is added**

This will result in a flat hierarchy for groups (no nested group support), meaning if a user is able to search users and groups to add access for, they will have to add all groups specifically that they want to have access

**If only the Full Group Path Mapper is added**

This will allow nested group support. If a user is able to search users and groups, if they add a group that has children, keycloak users that belong to the child and/or to the parent group will have access

If both group mappers are added, it defaults to nested group support.

dramich:
> Users with access to search will still be able to search users AND groups, but adding groups will not authorize others to login. The user would have to add specific user principals for others to be able to login.

This is a bad user experience and should probably follow the saml path of not returning anything if we can't do anything with groups. Can we tell if the required mappers are not setup?

brendarearden:
@dramich so after doing some testing, it would appear that if neither of the mappers are configured, the claiminfo returned from keycloak will look like this:

```
claimInfo {
Subject: "0945090349509"
Name: "Brenda Test"
PreferredUsername: ""
GivenName: ""
FamilyName: ""
Email: "random@random.com"
EmailVerified: "false"
Groups: nil
FullGroupPath: nil
}
```
If the mapper is configured, but the user does not have groups, it shows up like this
```
claimInfo {
....
Groups: []string
FullGroupPath: nil
}
```

A solution I can use: I can check if both of those values are nil, add a flag on the `OIDCConfig` type for `groupSearch: bool` and set is false so that when the SearchPrincipals method is called I have something to check against before adding groups to the search result

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -99,15 +100,16 @@ func (o *OpenIDCProvider) LoginUser(ctx context.Context, oauthLoginInfo *v32.OID
...
 	userPrincipal.Me = true
-
-	for _, group := range claimInfo.Groups {
-		groupPrincipal := o.groupToPrincipal(group)
-		groupPrincipal.MemberOf = true
-		groupPrincipals = append(groupPrincipals, groupPrincipal)
-
+	groupPrincipals = o.getGroupsFromClaimInfo(claimInfo)
+	//setting a bool for group search flag
+	if claimInfo.Groups == nil && claimInfo.FullGroupPath == nil {
```
brendarearden, line 107: This is flag is only affected on enable and edit for the auth provider, but I cannot move it to the testAndApply func because the claim information cannot be returned from the LoginUser func.  the framework is expect LoginUser to return only specific values.

dramich: Add this comment as a comment in the code for future generations

brendarearden: After fact checking myself, I was able to move this. I was thinking of AuthenticateUser, which was using LoginUser. I was able to ignore the claimsInfo being returned from LoginUser in the AuthenticateUser func so that its not an issue :D

dramich (approved):

dramich (approved):

PR #33720, opened 2021-07-22
Lookup azure tenant id  if not provided

cmurphy:
* Use rancher/machine FindTenantID to look up the tenant ID if not found in the cloud credential secret in /meta/aks* proxy handlers.

* Reverts https://github.com/rancher/rancher/pull/33680

Depends on https://github.com/rancher/aks-operator/pull/44

https://github.com/rancher/rancher/issues/33097


pkg/api/norman/customization/aks/handler.go
```diff
@@ -209,7 +218,10 @@ func (h *handler) getCredentialsFromBody(req *http.Request, cap *Capabilities) (
 		return http.StatusBadRequest, fmt.Errorf("invalid subscriptionId")
 	}
 	if cap.TenantID == "" {
-		return http.StatusBadRequest, fmt.Errorf("invalid tenantId")
+		cap.TenantID, err = azureutil.FindTenantID(context.Background(), azure.PublicCloud, string(cap.SubscriptionID))
```
cmurphy, line 221: Going to undo this part, AKSv1 should always have the tenant ID passed in the request body

pkg/api/norman/customization/aks/handler.go
```diff
@@ -154,6 +165,33 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
...
+			cc.Annotations[tenantIDAnnotation] = cap.TenantID
+			cc.Annotations[tenantIDTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+			// Ignore errors when updating the secret object. If the secret cannot be updated
+			// (perhaps due to a conflict error), the tenant ID will be re-fetched next time
+			// the proxy request is made and an update will be attempted again.
+			_, err := h.secretClient.Update(cc)
+			logrus.Debugf("encountered error while updating secret, ignoring: %v", err)
+		}
+	}
+
```
rmweir, line 194: I think putting this in it's own function would help with the nesting/readability.

rmweir (changes requested):
Are multiple requests made rapidly or in parallel from the UI that might make the handler use FindTenant ID multiple times because they all get the cloud credential before it's been updated?

pkg/api/norman/customization/aks/handler.go
```diff
@@ -154,6 +165,33 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
...
+			cap.TenantID, err = azureutil.FindTenantID(context.Background(), azure.PublicCloud, string(cap.SubscriptionID))
+			if err != nil {
+				return httperror.ServerError.Status, err
+			}
+			cc.Annotations[tenantIDAnnotation] = cap.TenantID
+			cc.Annotations[tenantIDTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
+			// Ignore errors when updating the secret object. If the secret cannot be updated
+			// (perhaps due to a conflict error), the tenant ID will be re-fetched next time
+			// the proxy request is made and an update will be attempted again.
+			_, err := h.secretClient.Update(cc)
```
rmweir, line 190: Error should be nil checked and only printed out if it is not nil.

rmweir (changes requested):

pkg/api/norman/customization/aks/handler.go
```diff
@@ -154,6 +165,33 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
...
+		tenantAnno, timestamp := annotations[tenantIDAnnotation], annotations[tenantIDTimestampAnnotation]
+		if tenantAnno != "" && timestamp != "" {
+			parsedTime, err := time.Parse(time.RFC3339, timestamp)
+			if err != nil {
+				return httperror.ServerError.Status, err
+			}
+			if parsedTime.Add(tenantIDTimeout).After(time.Now().UTC()) {
+				cap.TenantID = tenantAnno
+			}
+		}
```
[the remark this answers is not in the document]
cmurphy: The timestamp is the whole point, we want to refresh it in case it has changed. If we weren't refreshing it we might as well store it once in the secret data from a controller.

cmurphy (commented):
> Are multiple requests made rapidly or in parallel from the UI that might make the handler use FindTenant ID multiple times because they all get the cloud credential before it's been updated?

The UI will make around 5 requests all at once

pkg/api/norman/customization/aks/handler.go
cmurphy: done

pkg/api/norman/customization/aks/handler.go
cmurphy: fixed

go.mod
```diff
@@ -11,6 +11,7 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20200712062324-13d1f37d2d77
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc10
+	github.com/rancher/aks-operator => github.com/cmurphy/aks-operator v0.0.0-20210724225226-8d6257d22eaa // FIXME
```
cbron, line 14: todo

pkg/api/norman/customization/aks/handler.go
```diff
@@ -242,3 +260,33 @@ func handleErr(writer http.ResponseWriter, errorCode int, originalErr error) {
 	writer.WriteHeader(errorCode)
 	writer.Write(asJSON)
 }
+
+func getCachedTenantID(secretClient v1.SecretInterface, subscriptionID string, secret *v1.Secret) (string, error) {
```
cbron, line 264: Weren't we going to reuse the aks-operator one ?

pkg/api/norman/customization/aks/handler.go
```diff
@@ -154,6 +165,13 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 	cap.ClientID = string(cc.Data["azurecredentialConfig-clientId"])
 	cap.ClientSecret = string(cc.Data["azurecredentialConfig-clientSecret"])
 
+	if cap.TenantID == "" {
+		cap.TenantID, err = getCachedTenantID(h.secretClient, cap.SubscriptionID, cc)
+		if err != nil {
+			logrus.Debugf("encountered error while updating secret, ignoring: %v", err)
```
cbron, line 171: What if this error is around timestamp parsing or a timeout calling azure ? If we can't get the tenantID at all we should error out entirely, if we get the tenantID and error we can then just log.

pkg/api/norman/customization/aks/handler.go
```diff
@@ -22,6 +25,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	tenantIDAnnotation          = "cluster.management.cattle.io/azure-tenant-id"
+	tenantIDTimestampAnnotation = "cluster.management.cattle.io/azure-tenant-id-created-at"
+	tenantIDTimeout             = time.Hour
```
cbron, line 31: On the previous conversation, we could make this 24 hours. If it ever changes it would be very rare. But I don't think 1 hour really hurts either. If anything its going to be that first flurry of requests that could cause a problem.

cbron (commented):
Can you squash the two code commits into one. The revert and wrangler generate can be one.

pkg/apis/go.mod
```diff
@@ -2,7 +2,10 @@ module github.com/rancher/rancher/pkg/apis
 
 go 1.14
 
-replace k8s.io/client-go => k8s.io/client-go v0.21.0
+replace (
+	github.com/rancher/aks-operator => github.com/cmurphy/aks-operator v0.0.0-20210726212725-acf9782735a0 // FIXME
+	k8s.io/client-go => k8s.io/client-go v0.21.0
```
thedadams, line 7: Can this be `v0.21.2` to be consistent?

thedadams (commented):
One minor comment right now.

pkg/apis/go.mod
cmurphy: done

pkg/api/norman/customization/aks/handler.go
cbron: Fair enough. Removing the timeout isn't a terrible idea but there are downsides. Let's just leave at an hour.

pkg/api/norman/customization/aks/handler.go
cbron: Was referencing this: https://github.com/rancher/aks-operator/pull/44#discussion_r675221614
You could just move the update logic out of there. You'd actually be able to sort through errors easier that way, and only ignore the update errors and not the query errors.

cmurphy: good point, done

thedadams (approved):

cbron (commented):
lgtm, just needs aks changes vendored and ricky review

pkg/api/norman/customization/aks/handler.go
```diff
@@ -154,6 +164,24 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 	cap.ClientID = string(cc.Data["azurecredentialConfig-clientId"])
 	cap.ClientSecret = string(cc.Data["azurecredentialConfig-clientSecret"])
 
+	if cap.TenantID == "" {
+		cap.TenantID, err = aks.GetCachedTenantID(cap.SubscriptionID, cc)
+		if err != nil {
+			return httperror.ServerError.Status, err
+		}
+		cc.Annotations[tenantIDAnnotation] = cap.TenantID
+		cc.Annotations[tenantIDTimestampAnnotation] = time.Now().UTC().Format(time.RFC3339)
```
rmweir, line 173: Shouldn't this only be refreshed if FindTenantID was used in GetCachedTenantID why refresh the annotations if it was retrieved from the annotations?

rmweir (changes requested):
Since you are getting the cc off a lister isn't it possible the requests happen quickly enough to where it doesn't see the annotation on the secret? I would just be curious to check it in logs once the UI is using it. Could you add a debug line above azureutil.FindTenantID(ctx, azure.PublicCloud, subscriptionID) in aks-operator's `GetCachedTenantID` function so we can check this later? Or if the UI is up can you add a log in dev and make sure that the UI requests only cause one request to azure and otherwise grabs it from the cache?

pkg/api/norman/customization/aks/handler.go
cmurphy: Fixed in aks-operator

cmurphy:
> Since you are getting the cc off a lister isn't it possible the requests happen quickly enough to where it doesn't see the annotation on the secret?

Yes this is almost guaranteed to happen. This is kind of forced by the design.

> I would just be curious to check it in logs once the UI is using it. Could you add a debug line

done in aks-operator

cmurphy:
> Yes this is almost guaranteed to happen. This is kind of forced by the design.

Actually in practice I think this will be safe, for now. UI just makes one request to /meta/aksLocations to look up the region before moving on to the next screen, so there will be enough time between that request and the user clicking the "next" button for the annotation to be populated. If the UI ever changes this to combine the full view then this might cause a problem.

thedadams (dismissed):

rmweir (approved):
LGTM, be sure to update aks-operator in go.mod.

cmurphy:
Fixed go.mod

thedadams (approved):

cbron (approved):

PR #33722, opened 2021-07-22
fix selectFirstEntry

kinarashah:


PennyScissors (approved):

rmweir (approved):
LGTM, newline could be added back to end of file though

pkg/image/resolve_test.go
```diff
@@ -495,12 +496,13 @@ func flatStringSlice(slices ...[]string) []string {
 }
 
 func selectFirstEntry(rkeSystemImages map[string]rketypes.RKESystemImages) rketypes.RKESystemImages {
-	sortedRkeSystemImages := make([]rketypes.RKESystemImages, 0, len(rkeSystemImages))
-	for _, rkeSystemImage := range rkeSystemImages {
-		sortedRkeSystemImages = append(sortedRkeSystemImages, rkeSystemImage)
+	k8sVersions := make([]string, 0, len(rkeSystemImages))
```
snasovich, line 499: If all we need is value at lowest key, I think it can be done simpler without sort, like this:

```
	if len(rkeSystemImages) == 0 {
		return rketypes.RKESystemImages{}
	}
	lowestVersion := ""
	for version := range rkeSystemImages {
		if lowestVersion == "" || version < lowestVersion {
			lowestVersion = version
		}
	}
	return rkeSystemImages[lowestVersion]
```

Also, I'm not sure what exactly is in the keys of that map, but is it acceptable to sort lexicographically to find smallest version? Since "11.0" < "8.0"

PennyScissors: @kinarashah @snasovich I have a PR up that changes this test to a unit test approach rather than reading the data.json to get the k8s info and this function is removed, so I think this is okay as long as it fixes the issue we are seeing right now.

snasovich: Certainly not worth splitting hairs in this case.

kinarashah: @snasovich k8s versions is a small list of 3-5 versions, latest patch version per minor version. For instance, 1.19.x, 1.20.x and 1.21.x. So would be acceptable to sort in this case while we do use "mcuadros/go-version" for kdm logic where we filter data. 

Agree on the linear traverse vs sort part. But going to merge this as is based on @PennyScissors's comment.

Issue #32797, opened 2021-05-19
make ingress webhook image configurable


kinarashah:
with k8s 1.21, we introduced new ingress template for nginx ingress 0.46 which has hardcoded webhook image `docker.io/jettech/kube-webhook-certgen:v1.5.1` 
https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/templates/nginx_v0.46.go#L589 

We need to:
- add a new field IngressWebhookImage to RKE's system images 
- mirror this image to `rancher/mirrored-kube-webhook-certgen` in `images-mirror` repo 
- use `rancher/mirrored-*` image as the default in RKE system images
- update RKE code to pass the image 
- update KDM template to change from hardcoded image to configurable image passed by RKE 
- confirm `rancher-images.txt` contains this new image

https://github.com/rancher/rancher/issues/31567

izaac:
Rancher version: `2.6 master-head (07/22/2021)` 
RKE `1.3.0-rc9`
Build your own valid certs install

I'm getting cert errors during Rancher install:

`Error: Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io": Post "https://ingress-nginx-controller-admission.ingress-nginx.svc:443/networking/v1beta1/ingresses?timeout=10s": dial tcp 10.43.158.127:443: connect: connection refused`

Logs from nginx controller pods `nginx-ingress-controller-xxxxx`

```
I0722 16:45:02.589122       6 nginx.go:296] "Starting NGINX process"
I0722 16:45:02.589356       6 leaderelection.go:243] attempting to acquire leader lease ingress-nginx/ingress-controller-leader-nginx...
I0722 16:45:02.589667       6 nginx.go:316] "Starting validation webhook" address=":8443" certPath="/usr/local/certificates/cert" keyPath="/usr/local/certificates/key"
I0722 16:45:02.589915       6 controller.go:146] "Configuration changes detected, backend reload required"
I0722 16:45:02.605175       6 status.go:84] "New leader elected" identity="nginx-ingress-controller-244jl"
I0722 16:45:02.681363       6 controller.go:163] "Backend successfully reloaded"
I0722 16:45:02.681468       6 controller.go:174] "Initial sync, sleeping for 1 second"
[...]

PR #33750, opened 2021-07-23
Automation - RKE HA ingress check before installing Rancher

izaac:
It seems that starting with rke 1.3 and k8s 1.21 the ingress-nginx resources take a bit longer to get into an active state, making some of the certificate install types to fail.

This check will make the test wait until the resources are active.
Related: https://github.com/rancher/rancher/issues/32797#issuecomment-885091240


sowmyav27 (approved):
LGTM

slickwarren (approved):

Issue #33625, opened 2021-07-14
Add Windows RKE2 Registration Command


luthermonson:
We need to 
* auto add labels, add `--label cattle.io/os=windows` for windows 
* auto add labels, add `--label cattle.io/os=linux` for linux 
* add the install.ps1 to rancher and get the url working like the install.sh and pass that url back in the reg token
* validate that node labels work in ps1 file and the xheader is properly sent to setup the node

during testing confirm that machine label selectors work for config and try to set kubelet-args etc. and see if the config propagates to the downstream node. if it doesn't work figure out why and create a new issue for it.

sirredbeard:
Outstanding PR https://github.com/rancher/rancher/pull/33759

slickwarren:
on master-head (256a0eb) I ran into this issue when running the new windows registration command:
```
                                                                                                                                                                                                C:\Users\Administrator>                                                                                                 C:\Users\Administrator>curl.exe --insecure -fL https://<ip>/rke2-agent-install.ps1 -o install.ps1 && powershell.exe -ExecutionPolicy Bypass -Command "./install.ps1 -Server https://<ip> -Label 'cattle.io/os=windows' -Token <token> -Worker -CaChecksum --ca-checksum <checksum>"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13609    0 13609    0     0  13609      0 --:--:-- --:--:-- --:--:-- 45821
INFO: Using default agent configuration directory C:/etc/rancher/agent


    Directory: C:\etc\rancher


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----          8/5/2021  12:44 AM                agent
INFO: Using default agent var directory C:/etc/rancher/agent
Test-CaCheckSum : Value from https://<ip>cacerts does not look like an x509 certificate, exited with
At C:\Users\Administrator\install.ps1:411 char:9
+         Test-CaCheckSum
+         ~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Test-CaCheckSum

Test-CaCheckSum : Retrieved cacerts:
At C:\Users\Administrator\install.ps1:411 char:9
+         Test-CaCheckSum
+         ~~~~~~~~~~~~~~~
[...]

PR #33759, opened 2021-07-26
Add Rke2 Install Command for Windows Nodes

luthermonson:
This will add functionality to run rke2 custom clusters with windows nodes. Note the UI will not show the custom windows command without the cluster being a v1.21 + calico as cni.
* Adds rke2 windows node command
* Adds install.ps1 endpoint for serving the install script to the downstream node
* Adds a new /v3/connect/config-yaml custom api end point to be used with rke2 windows installations
  * Endpoint will make a new node in the cluster and ping the machine plan until it has the rke2 config, it then returns that as json for the rke2 windows node to consume to boot a node.

https://github.com/rancher/rancher/issues/33625


pkg/api/steve/additionalapi.go
```diff
@@ -19,11 +19,14 @@ import (
...
 func AdditionalAPIsPreMCM(config *wrangler.Context) func(http.Handler) http.Handler {
+	connectHandler := configserver.New(config)
 	if features.RKE2.Enabled() {
 		mux := gmux.NewRouter()
 		mux.UseEncodedPath()
-		mux.Handle("/v3/connect/agent", configserver.New(config))
+		mux.Handle("/v3/connect/agent", connectHandler)
+		mux.Handle("/v3/connect/config-yaml", connectHandler)
 		mux.Handle("/system-agent-install.sh", server.InstallHandler())
+		mux.Handle("/windows-agent-install.ps1", server.WindowsInstallHandler())
```
thedadams, line 29: Can this be combined like the `connectHandler` above?

pkg/controllers/dashboard/clusterregistrationtoken/status.go
```diff
@@ -5,9 +5,10 @@ import (
 	"net/url"
 	"strings"
 
+	util "github.com/rancher/rancher/pkg/cluster"
```
thedadams, line 8: This can be put back.

pkg/provisioningv2/rke2/configserver/server.go
```diff
@@ -256,3 +312,22 @@ func (r *RKE2ConfigServer) getSecret(machineName string, planSA *corev1.ServiceA
 	secret, err := r.secretsCache.Get(planSA.Namespace, planSA.Secrets[0].Name)
 	return planSA.Labels[planSecret], secret, err
 }
+
+func (r *RKE2ConfigServer) getMachinePlanSecret(name, ns string) (secret *v1.Secret, err error) {
+	for i := 0; i < 10; i++ {
+		secret, err = r.secretsCache.Get(name, ns)
+		if err != nil {
+			logrus.Debugf("error waiting for machine plan secret, sleeping 1s and trying again: %s", err)
```
thedadams, line 320: Technically only waiting 500 milliseconds.

pkg/provisioningv2/rke2/configserver/server.go
```diff
@@ -256,3 +312,22 @@ func (r *RKE2ConfigServer) getSecret(machineName string, planSA *corev1.ServiceA
...
+			time.Sleep(500 * time.Millisecond)
+			continue
+		}
+
+		if len(secret.Data) == 0 || string(secret.Data[rolePlan]) == "" {
+			logrus.Debug("machine plan secret data is empty waiting on planner controller, sleeping 1s and trying again\n")
```
thedadams, line 326: Same.

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -801,7 +802,9 @@ func addTaints(config map[string]interface{}, machine *capi.Machine) error {
 	}
 
 	for _, taint := range taints {
-		taintString = append(taintString, fmt.Sprintf("%s=%s:%s", taint.Key, taint.Value, taint.Effect))
+		if taint.Key != "" && taint.Value != "" && taint.Effect != "" {
```
thedadams, line 805: Do we want to return an error if the taint isn't in the valid form?

pkg/provisioningv2/rke2/configserver/server.go
luthermonson: ha!

pkg/provisioningv2/rke2/planner/planner.go
luthermonson: i just dont want "=:" ending up in the list of taints config

pkg/api/steve/additionalapi.go
luthermonson: two different ideas too me... ones a config server returning json and needs controllers/caches/settings etc. and the other returns plain text scripts to be consumed by down stream nodes

rosskirkpat (approved):
LGTM

thedadams (dismissed):

phillipsj (approved):

pkg/api/steve/additionalapi.go
```diff
@@ -12,18 +12,21 @@ import (
...
 )
 
 func AdditionalAPIsPreMCM(config *wrangler.Context) func(http.Handler) http.Handler {
+	connectHandler := configserver.New(config)
 	if features.RKE2.Enabled() {
 		mux := gmux.NewRouter()
 		mux.UseEncodedPath()
-		mux.Handle("/v3/connect/agent", configserver.New(config))
-		mux.Handle("/system-agent-install.sh", server.InstallHandler())
+		mux.Handle(configserver.ConnectAgent, connectHandler)
```
dramich, line 26: nit but I think it's easier to have the string path here, makes it faster to see which handler you want to look at vs having to go to the const then back here

thedadams (dismissed):

pkg/provisioningv2/rke2/installer/server.go
```diff
@@ -0,0 +1,28 @@
...
+func (s *handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	var err error
+	var content []byte
+
+	switch req.URL.Path {
+	case SystemAgentInstallPath:
+		content, err = LinuxInstallScript("", nil, req.Host)
+	case WindowsRke2InstallPath:
+		content, err = WindowsInstallScript("", nil, req.Host)
+	}
```
dramich, line 20: Does this need a base case?

luthermonson: if other urls were added to handle, not possible to get here beyond the two passed... who would ever get here and serve the response? even if it did it'd send an empty string back in plain text, not the end of the world

dramich (approved):

thedadams (approved):

Issue #33385, opened 2021-06-25
Node with worker role only is seen as role All in the UI


ShylajaDevadiga:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**

- Create a custom cluster with one master node with all roles (etcd, cp, worker)
- Wait for master to be active
- Join agent node using Worker role and run registration command on the node.
- Wait for the cluster to be Active
- Click on cluster, select Explore, go to Nodes
- Worker node has role All. Screenshot attached.
![image](https://user-images.githubusercontent.com/56045581/123494135-aa7bfc00-d5d3-11eb-88f3-36decb429330.png)

**Environment information**

Rancher version: master-head commit: c7f3cce

**Installation option (single install/HA):**
Two node rke2 cluster, one master, one agent

**Cluster information**

Cluster type:Custom
Machine type: AWS
Kubernetes version: v1.21.2+rke2r1

ShylajaDevadiga:
Validated using master-head commit id 01665d0, worker node has role as Worker in a 2 node Custom cluster setup.
![image](https://user-images.githubusercontent.com/56045581/127201208-fbcd3151-1f5d-4f41-8aef-d3daa08ab139.png)

Issue #32654, opened 2021-05-10
worker label missing on custom RKE2 all-in-one node deployed from Explorer


thehejik:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**
* Use Cluster Management in Cluster Explorer UI
* Create a Custom RKE2 single-node cluster by using Registration Command with all 3 roles set

**Result:**
The managed rke2 cluster is up after a while but `node-role.kubernetes.io/worker=true` is not set. There are `etcd`,  `control-plane` and `master` role lables set.

**Other details that may be helpful:**
Ember UI is showing roles `etcd, Control Plane` and Explorer UI is showing only `Etcd` - but in both cases there should be role `All` present.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher/rancher:master-head `8c014ef`
- Installation option (single install/HA): any

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Custom RKE2
- Machine type (cloud/VM/metal) and specifications (CPU/memory): any
- Kubernetes version (use `kubectl version`): `v1.21.0-alpha2+rke2r1`

deniseschannon:
Besides adding the label, we also need to taint the nodes to support the worker role.

Note: when adding these taint, this will cause toleration issues to the RKE2 charts.

deniseschannon:
We need to add the context where if the worker role is selected, then add the worker label (`node-role.kubernetes.io/worker=true`):

https://github.com/rancher/rancher/blob/113c35317ea3524a8d6e9da6659c772377d33f2b/pkg/provisioningv2/rke2/planner/planner.go#L554

This issue should be used in conjunction with https://github.com/rancher/rancher/issues/32904

sowmyav27:
**On master-head commit id: `2e6926b`**

- Deployed a 1 node rke2 cluster.
- `All` roles is seen on the cluster details page in Machines tab
- Labels on the node
```
  labels:
    cluster.x-k8s.io/cluster-name: custom
    objectset.rio.cattle.io/hash: <>
    rke.cattle.io/control-plane-role: "true"
    rke.cattle.io/etcd-role: "true"
    rke.cattle.io/init-node: "true"
    rke.cattle.io/machine-id: <>
    rke.cattle.io/worker-role: "true"
```
[...]

PR #33761, opened 2021-07-26
machine worker label sync

kinarashah:
https://github.com/rancher/rancher/issues/33385
https://github.com/rancher/rancher/issues/32654

### DIFF
--- a/pkg/controllers/managementuser/machinerole/machinerole.go
+++ b/pkg/controllers/managementuser/machinerole/machinerole.go
@@ -1,0 +1,57 @@
+package machinerole
+
+import (
+	"context"
+
+	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1alpha4"
+	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	"github.com/rancher/rancher/pkg/types/config"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type handler struct {
+	clusterName string
+	nodes       v1.NodeInterface
+	machines    capicontrollers.MachineCache
+}
+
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/machinerole/machinerole.go
```diff
@@ -0,0 +1,53 @@
...
+	machineName := node.Annotations["cluster.x-k8s.io/machine"]
+	if machineName == "" {
+		return node, nil
+	}
+	machineNS := node.Annotations["cluster.x-k8s.io/cluster-namespace"]
+	if machineNS == "" {
+		return node, nil
+	}
+
+	if _, ok := node.Labels["node-role.kubernetes.io/worker"]; ok {
```
dramich, line 42: This should be the first check

pkg/controllers/managementuser/machinerole/machinerole.go
```diff
@@ -0,0 +1,53 @@
...
+		return node, nil
+	}
+	machine, err := h.machines.Get(machineNS, machineName)
+	if err != nil {
+		return nil, err
+	}
+	if val, exists := machine.Labels["rke.cattle.io/worker-role"]; exists && val == "true" {
```
StrongMonkey, line 48: nit, this should be enough
`machine.Labels["rke.cattle.io/worker-role"] == "true"`

StrongMonkey: probably we should update deepcopy only and not modifying `node` directly. Also no need to run `h.nodes.Update(node)` if the above condition is false?

kinarashah: thanks, forgot we need deepcopy! updating

StrongMonkey (approved):

dramich (approved):

PR #33767, opened 2021-07-26
Also run c_rehash on startup

thedadams:
A previous change moved the ca-additional.pem files to
/etc/pki/trust/anchors/ diretory. However, while this does allow the
additional trusted CAs to be detected in some instances, other
applications (like git) read the SSL_CERT_DIR environment variable.
Since this variable is set in the Rancher container, the additional
trusted CAs should also be put in the referenced directory
(/etc/rancher/ssl) and c_rehash run on startup.

Issues:
https://github.com/rancher/rancher/issues/33398
https://github.com/rancher/rancher/issues/31846#issuecomment-885053870

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -160,6 +160,10 @@ spec:
           name: tls-ca-additional-volume
           subPath: ca-additional.pem
           readOnly: true
+        - mountPath: /etc/rancher/ssl/ca-additional.pem
+          name: tls-ca-additional-volume
+          subPath: ca-additional.pem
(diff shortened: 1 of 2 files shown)


paynejacob (approved):

jiaqiluo (approved):
LGTM

cbron (approved):
lgtm, this reverts us back to what it used to be, besides the other path moving from ` /etc/ssl/certs/` to `/etc/pki/trust...`

thedadams:
Not quite a revert. We didn't need to run c_rehash previously.

Issue #33777, opened 2021-07-27
Verify Rancher server helm chart will be stopped on a k8s 1.22+ cluster


deniseschannon:
We need to verify if we can run Rancher on a k8s 1.22 cluster and if not, we need to update our Rancher server helm chart to limit deploying on a k8s 1.22 chart until it's supported.

snasovich:
I'm 99% sure it won't be able to run on 1.22 as changes to move CRDs to v1 API from v1beta1 APIs are not in.
@cmurphy , could you please verify this and update Rancher Helm charts?

cmurphy:
Rancher will not run on kubernetes 1.22 because of https://github.com/rancher/rancher/issues/33222. Added https://github.com/rancher/rancher/pull/33781 to pin the version.

deniseschannon:
Available to test in v2.6.0-rc1

snasovich:
@thedadams , please validate this one. k8s 1.22 is out so it should be possible to test.

thedadams:
This issue is verified: installing Rancher from the new helm chart produces an error and Rancher is not installed.
![Screen Shot 2021-08-06 at 11 29 57 AM](https://user-images.githubusercontent.com/3682717/128556048-666ac40f-ca43-4b7d-bda3-b9f8244f504f.png)
![Screen Shot 2021-08-06 at 11 31 01 AM](https://user-images.githubusercontent.com/3682717/128556138-9faa71e3-d1bd-420d-9e37-ad8466246169.png)

deniseschannon:
Rancher will extend support to k8s 1.22 with this issue. Due to the dropped APIs, it will take signifcant work to get this supported. 

https://github.com/rancher/rancher/issues/34060

PR #33781, opened 2021-07-27
Set kubeVersion range in rancher charts

cmurphy:
Rancher does not yet run on kubernetes 1.22, so ensure the charts comply
with that.

https://github.com/rancher/rancher/issues/33777

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -3,6 +3,7 @@ name: rancher
 description: Install Rancher Server to manage Kubernetes clusters across providers.
 version: %VERSION%
 appVersion: %APP_VERSION%
+kubeVersion: < 1.22.0-0
 home: https://rancher.com
 icon: https://github.com/rancher/ui/blob/master/public/assets/images/logos/welcome-cow.svg
 keywords:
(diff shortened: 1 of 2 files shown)


chart/Chart.yaml
```diff
@@ -3,6 +3,7 @@ name: rancher
 description: Install Rancher Server to manage Kubernetes clusters across providers.
 version: %VERSION%
 appVersion: %APP_VERSION%
+kubeVersion: >= v1.17.17 < 1.22.0
```
StrongMonkey, line 6: I don't think we typically add min version based on our support matrix. It will fail if user is upgrading from < k8s 1.17. So probably `< 1.21.0-0` should be enough. cc @deniseschannon @cbron

snasovich: +1, was just about to suggest the same. Now, do we need `< 1.22.0-0` or would `< 1.22.0` be enough?

StrongMonkey: I think `< 1.22.0-0` exclude any pre-release too.

cmurphy: done

StrongMonkey (approved):

chart/Chart.yaml
snasovich: @StrongMonkey , just tested it a bit to be sure. `< 1.22.0-0` will actually allow pre-releases from 1.21 and below while `< 1.22.0` will not allow any pre-releases period.
So I think `< 1.22.0-0` is a better fit as it's less restrictive.

image

snasovich (approved):

cbron (approved):
Potential here for conflicts depending on what external customized k8s versions exist. We'll see.

PR #33787, opened 2021-07-27
fixed chart version having leading zeros

paynejacob:
Helm3 now requires semver2 for chart versions.  Leading zeros are prohibited in semver2.  This can cause some builds to fail when the commit hash starts with a zero.


aiyengar2 (approved):
LGTM

dramich (approved):

PR #33789, opened 2021-07-27
Boostrap password tweaks

vincent99:
- If the user supplied their own password, don't set `mustChangePassword` so the UI won't force them to change it.
- Show a more obvious visual block to make it easier to find the bootstrap info among all the noise `docker run ...` spits out
- Use go-template with internal base64 decode instead of piping and requiring base64 to be present in the commands
- Show the password on it's own line separate from the URL.  The UI will show the user instructions to grep it easily if they're already at the login screen but lacking the password.
- Don't log the password if the user provided it (but this was already done an hour ago..)
- If the user provided password is `admin`, the user is opting-in preserve existing 2.5 behavior (Don't show `?setup=admin` because it's redundant, `mustChangePassword=true`).

### DIFF
--- a/chart/templates/NOTES.txt
+++ b/chart/templates/NOTES.txt
@@ -1,13 +1,22 @@
 Rancher Server has been installed.
 
-NOTE: Rancher may take several minutes to fully initialize. Please standby while Certificates are being issued and Ingress comes up.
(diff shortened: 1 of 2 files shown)


StrongMonkey (approved):
:100:

paynejacob (approved):

PR #33791, opened 2021-07-27
Add auth provisioning permissions

dramich:
Pull in the updated api server so lists don't fail with a 403
Remove the owner reference for roleBindings created for the provisioning cluster. The CRTB lives in another namespace so it can't be used as the owner. 
Update the naming convention of the roleBinding to be unique based on the CRTB name

### DIFF
--- a/pkg/controllers/management/authprovisioningv2/binding.go
+++ b/pkg/controllers/management/authprovisioningv2/binding.go
@@ -7,62 +7,63 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


cmurphy (approved):

pkg/controllers/management/authprovisioningv2/binding.go
```diff
@@ -7,62 +7,63 @@ import (
...
+			Name:     crtb.GroupName,
 		})
 	}
 
-	return crbt, h.roleBindingApply.
-		WithOwner(crbt).
+	return crtb, h.roleBindingApply.
+		WithOwner(crtb).
 		WithListerNamespace(cluster.Namespace).
-		WithSetOwnerReference(true, true).
```
dramich, line 65: This has to be dropped, the CRTB and the roleBinding being created are in different namespaces so the roleBinding gets immediately deleted

Issue #33419, opened 2021-06-29
While provisioning rke2 custom cluster, taints have invalid format


ShylajaDevadiga:
Using rancher master-head commit id c7f3cce
While provisioning rke2 custom cluster on AWS/DO, taints fail to be applied on the cluster due to invalid format stored in config
Follow the steps to reproduce to create the cluster.

Value stored in config
```
"node-taint": [
    "key1=value1=NoSchedule:"
  ],
```
Logs:

```
Jun 24 20:49:54 shyubu rke2[1375]: E0624 20:49:54.440717    1375 server.go:156] "Failed to parse kubelet flag" err="invalid argument \"key1=value1=NoSchedule:\" for \"--register-with-taints\" flag: invalid t
aint effect: , unsupported taint effect"
```
Format when taint is applied on a standalone rke2 node as seen in logs is as below
```
--register-with-taints=key1=value1:NoSchedule
```

davidnuzik:
**Reproduction:**
Not required.

**Validation Successful:**
Rancher version: master-head `b39523882` 8/2/21
Rancher cluster type: single-node docker
Docker version: 19.03

Downstream cluster type: rancher-provisioned rke2 custom cluster
Downstream K8s version: v1.21.3-rc1+rke2r2

Validation steps:
1. Provision an rke2 custom cluster in the dashboard UI. Create a separate pool for workers with a taint on it.
![image](https://user-images.githubusercontent.com/45179589/128066232-72b8f3d3-c42d-4b53-9f09-0f3341c3a8fa.png)

During this process, I observed no issues in the UI or in the rancher server logs as reported originally in the issue (issue appears fixed so far).

2. Once the cluster is ready, inspect the tainted worker node in the UI.
I noticed that the node yaml appears correct (snippet below):
```
metadata:
  annotations:
    rke.cattle.io/taints: '[{"key":"testtaint","value":"true","effect":"PreferNoSchedule"}]'
  creationTimestamp: "2021-08-03T18:22:04Z"
```
I also checked the node in the UI and the taint was successfully applied.
Last, I checked the rke2.io/node-args annotation on the worker in this pool with this taint and it looks correct:
```
[
  "agent",
  "--node-label",
  "rke.cattle.io/machine=76a86ae5-439a-42ac-a9aa-7d192f7cfe06",
  "--node-taint",
  "testtaint=true:PreferNoSchedule",
  "--server",
  "https://REDACTED:9345",
  "--token",
  "********"
]
```

[...]

PR #33797, opened 2021-07-28
Correctly parse taints for custom clusters

thedadams:
The taints for custom clusters are passed as `key=value:Effect`.
However, the parsing of the taint strings were expecting something of
the form `key:value:Effect`.

This change correctly parses the taints for custom clusters.

Issue:
https://github.com/rancher/rancher/issues/33419

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/unmanaged/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/unmanaged/controller.go
@@ -13,10 +13,10 @@ import (
 	rocontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	rkecontroller "github.com/rancher/rancher/pkg/generated/controllers/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/provisioningv2/rke2/planner"
+	"github.com/rancher/rancher/pkg/taints"
 	"github.com/rancher/rancher/pkg/wrangler"
 	"github.com/rancher/wrangler/pkg/apply"
 	"github.com/rancher/wrangler/pkg/data"
-	"github.com/rancher/wrangler/pkg/data/convert"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/kv"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/provisioningv2/rke2/unmanaged/controller.go
```diff
@@ -186,26 +185,27 @@ func (h *handler) createMachineObjects(capiCluster *capi.Cluster, machineName st
...
 
 	var taints []corev1.Taint
-	for _, taint := range convert.ToStringSlice(data["taints"]) {
-		for _, taint := range strings.Split(taint, ",") {
-			parts := strings.Split(taint, ":")
+	for _, taint := range data.StringSlice("taints") {
+		for _, taintString := range strings.Split(taint, ",") {
+			parts := strings.Split(taintString, ":")
+			if len(parts) == 2 {
+				// In this case, we check for '=' in the first part to catch strings of the form 'key=value:Effect'
```
snasovich, line 192: Is it a valid use case when taint is not in `key:value=effect` format? The code here suggests we handle other taint formats with fewer parts. Also, the test only covers `key:value=effect` (kudos for adding test at all!).

If it is the only valid format, should be simplify the logic to expect only this format and return error value(s) don't match the expected format?

thedadams: For clarification, we expect the taints to be in `key=value:effect` form.

To your point: if we assume that taints must always be in that form, then the logic can be simplified. I was following what Darren had previously included to be more flexible. I am not sure of the reasons why he wanted this logic to be more flexible.

snasovich: @thedadams , thank you. Makes sense - it's not broken, so we don't need to fix it.

snasovich (dismissed):

pkg/controllers/provisioningv2/rke2/unmanaged/controller.go
ibuildthecloud: What I previous wrote doesn't matter as much as matching the behavior of how rke1 parses and handles taints during custom node add.  From what I gather this is the code that is used https://github.com/rancher/rancher/blob/8ea9351cf0e8b395dc85937317f28a204c8579a5/pkg/taints/utils.go#L79  So we should probably just use that code or match the logic of that.

snasovich: So previous logic only accepts `key=value:effect` format and logs warnings for anything else.
Though it's worth noting that `GetRKETaintsFromString` returns `nil` for invalid formats and `GetRKETaintsFromStrings` doesn't check for `nil`s. :(

snasovich (approved):

ibuildthecloud (approved):

Issue #33458, opened 2021-07-01
[Backport 2.5] Ability to specify network tags for node pools in GKE cluster


snasovich:
Backport to v2.5 for https://github.com/rancher/rancher/issues/33406

**Request description:**
Add network tags when adding node pools to a GKE cluster during provisioning and edit.

**Actual behavior:**
Can't specify tags when working with node pools in a GKE cluster through Rancher.

**Expected behavior:**

- When creating a GKE v2 cluster through the Rancher UI, a user should be able to specify 0 or more network tags to add to the nodes in a node pool. Mimicking the behavior in the GCP console.
- When creating a GKE v2 cluster through the API or TF provider, the user should be able to specify a list of 0 or more network tags.

**Additional notes:**
https://cloud.google.com/vpc/docs/add-remove-network-tags

_JIRA ID: SURE-3037_

cmurphy:
Chart updated in https://github.com/rancher/system-charts/pull/500

prabalsharma:
v2.5-0e587683fd86a24d25fb8d567c946d3edd867000-head

same as https://github.com/rancher/rancher/issues/33406#issuecomment-885902682

prabalsharma:
tested on v2.5-07131035f51f3e264c114e9d9dde65aae37bdf4f-head

prabalsharma:
looks like new argument needs to be added to tfp

```
Error: Unsupported argument

  on terra.tf line 48, in resource "rancher2_cluster" "prabal-tf2":
  48:         tags = ["http-server", "https-server"]

An argument named "tags" is not expected here.
```

prabalsharma:
this is the new tfp issue https://zube.io/rancher/rancher/c/41711

PR #33810, opened 2021-07-28
[v2.5] Update gke-operator to 1.1.1-rc10

cmurphy:
Backport of https://github.com/rancher/rancher/pull/33809

https://github.com/rancher/rancher/issues/33458

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -89,7 +89,7 @@ require (
 	github.com/rancher/apiserver v0.0.0-20210209001659-a17289640582
 	github.com/rancher/dynamiclistener v0.2.1-0.20200910203214-85f32491cb09
 	github.com/rancher/eks-operator v1.0.9
-	github.com/rancher/gke-operator v1.1.1-rc7
+	github.com/rancher/gke-operator v1.1.1-rc10
 	github.com/rancher/kubernetes-provider-detector v0.1.2
 	github.com/rancher/lasso v0.0.0-20210408231703-9ddd9378d08d
 	github.com/rancher/machine v0.15.0-rancher25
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -89,7 +89,7 @@ require (
 	github.com/rancher/apiserver v0.0.0-20210209001659-a17289640582
 	github.com/rancher/dynamiclistener v0.2.1-0.20200910203214-85f32491cb09
 	github.com/rancher/eks-operator v1.0.9
-	github.com/rancher/gke-operator v1.1.1-rc7
+	github.com/rancher/gke-operator v1.1.1-rc10
```
cbron, line 92: Doesn't this need to get updated here ? https://github.com/rancher/system-charts/blob/dev-v2.5/charts/rancher-gke-operator/1.0.400/values.yaml#L7

cmurphy: We could for completeness but it won't technically affect the bug, the bug was when rancher calls BuildUpstreamClusterState

thedadams (approved):

go.mod
cbron: I think we should keep 2.6 and 2.5 on same version since that is what QA is expecting. If there are chart changes we could re-consider, but even then I'd think we should do it.

cmurphy: They are the same version https://github.com/rancher/rancher/pull/33809/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R104

I haven't made a change to feature charts for 2.6, are you asking me to update feature charts and system-charts?

cbron: Oh, yea I think we should keep all 4 (both charts and both rancher lines) on same version of operator. Not split them up.

cmurphy: https://github.com/rancher/charts/pull/1378 https://github.com/rancher/system-charts/pull/510

cbron (approved):

Issue #33687, opened 2021-07-20
Download KubeConfig fails to download when MCM feature flag is disabled


sowmyav27:
**Rancher Server Setup**
- Rancher version: master-head
- Installation option (Docker install/Helm Chart): Docker install

**Information about the Cluster**
- Kubernetes version: v1.20.7+k3s1
- Cluster Type (Local/Downstream): Local

**Describe the bug**
Download KubeConfig fails to download when MCM is disabled

**To Reproduce**
- Deploy a single node docker install of rancher.
```
docker run -d --privileged --restart=unless-stopped -p 80:80 -p 443:443 rancher/rancher:master-head --features=multi-cluster-management=false
```
- Go to the cluster explorer of the local cluster
- Click on Download kubeconfig icon top right of the page
- Respoonse: `404 page not found`**

StrongMonkey:
This is a duplicate issue of https://github.com/rancher/rancher/issues/33711. Can be retested or closed.

catastrophe-brandon:
This worked in the local cluster with commit `d8824a1` on MacOS Chrome.

StrongMonkey:
@catastrophe-brandon I don't think this has worked yet. Did you disable `multi-cluster-management=false`?

catastrophe-brandon:
That was the part I forgot. Sorry for the confusion; reopening.

catastrophe-brandon:
A minor clarification on the results described in the issue: the 404 surfaces in the browser's dev console. In the UI nothing happens when I click on the icon. Here's the error:

[riker.xyz-1627338075594.log](https://github.com/rancher/rancher/files/6881522/riker.xyz-1627338075594.log)

StrongMonkey:
The explorer dashboard is relying on `/v3/clusters` api to generate kubeconfig, which is no longer available when running rancher with multi-cluster-management=false. https://github.com/rancher/rancher/pull/33811 introduce the same api on `v1/management.cattle.io.clusters` which is steve api(used by dashboard. This will require UI to switch to v1 api.

StrongMonkey:
Above change is done in backend. Now UI can request kubeconfig from v1 steve api. For example `/v1/management.cattle.io.clusters/local?action=generateKubeconfig`

PR #33811, opened 2021-07-28
Add v1 steve api for generateKubeconfig api

StrongMonkey:
Need to port over action `generateKubeconfig` from `v3/clusters` to `v1/management.cattle.io.clusters` since `v3/cluster` api will be disabled when running with multi-cluster-management=false and there is no equivalent api to deal with that. 

This requires UI to switch to using `/v1/management.cattle.io.clusters` api to generate kubeconfig. When multi-cluster-management is enabled, it will just redirect to `/v3/clusters` api.

https://github.com/rancher/rancher/issues/33687

### DIFF
--- a/pkg/api/steve/clusters/kubeconfig.go
+++ b/pkg/api/steve/clusters/kubeconfig.go
@@ -1,0 +1,78 @@
+package clusters
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/rancher/pkg/features"
+	"github.com/rancher/rancher/pkg/kubeconfig"
+	"github.com/rancher/rancher/pkg/settings"
+	"github.com/rancher/rancher/pkg/user"
+	"github.com/rancher/wrangler/pkg/schemas/validation"
+	"k8s.io/apiserver/pkg/endpoints/request"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/api/steve/clusters/kubeconfig.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (k kubeconfigDownload) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	apiRequest := types.GetAPIContext(req.Context())
+	if err := apiRequest.AccessControl.CanGet(apiRequest, apiRequest.Schema); err != nil {
+		apiRequest.WriteError(err)
+		return
+	}
+
+	if features.MCM.Enabled() {
+		http.Redirect(rw, req, fmt.Sprintf("/v3/clusters/%s?action=generateKubeconfig", apiRequest.Name), http.StatusFound)
```
dramich, line 30: Is this required, can we handle the token here anyways?

StrongMonkey: Not really, the old logic handles something like localAuthEndpoint which is no longer avaiable if you run MCM=false, so I would rather redirect the response if MCM is enabled.

dramich: Makes sense

dramich (approved):

kinarashah (approved):

Issue #33771, opened 2021-07-27
Cluster auth tokens are not being created when kubeconfig-generate-token setting is false


sowmyav27:
**Rancher Server Setup**
- Rancher version: master-head commit id: `2e6926b`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: 1.21.1
- Cluster Type (Local/Downstream): Downstream - RKE1 DO

**Describe the bug**
Cluster auth tokens are not being created when `kubeconfig-generate-token` setting is false. This issue is related to this https://github.com/rancher/rancher/issues/33474. I believe https://github.com/rancher/rancher/issues/33474 was closed after validation because a step was missing, it can be reproduced as long as the cluster endpoint context is used in the produced kubeconfig.

**To Reproduce**
1. set `kubeconfig-generate-token`  to "false"
2. create rke1 cluster
3. generate kubeconfig for cluster
4. run `kubectl config get-contexts`
5. run `kubectl config use-context <context-name>` where context name is the context using the cluster's endpoint.
6. run a kubectl command such as `kubectl get nodes`

**Results:**
The kubectl command will return an unauthed error and the following error will be repeatedly printed in logs:
```
2021/07/27 00:32:20 [ERROR] error syncing 'kubeconfig-u-9svh2jkkxv': handler cat-token-controller: clusterauthtokens.cluster.cattle.io "cattle-system/kubeconfig-u-9svh2jkkxv" not found, requeuing

```

**Expected:**
The kubectl command should work, and the error may show a few times since the token is created before the clusterauthtoken but should go away. the message will be changed to reflect that the clusterAuthToken has not been created __yet__.

___
### Original issue
**Describe the bug**
Errors seen in rancher logs - `error syncing 'kubeconfig-user-xqz7d.c-tz4pf': handler cat-token-controller` on using kubeconfig file when kubeconfig-generate-token is false

**To Reproduce**

- Deploy Rancher server
- Deploy a downstream RKE1 DO cluster
- Download kubeconfig for DO cluster
- Errors seen in Rancher logs:

```
2021/07/27 00:32:20 [ERROR] error syncing 'kubeconfig-u-9svh2jkkxv': handler cat-token-controller: clusterauthtokens.cluster.cattle.io "cattle-system/kubeconfig-u-9svh2jkkxv" not found, requeuing
[...]

PR #33816, opened 2021-07-28
Fix bug where clusterAuthToken not being created

StrongMonkey:
There seems to be a regression bug in creating clusterAuthToken when refactoring token code. https://github.com/rancher/rancher/commit/a213e857b2d5d5931ae74804532133140ad8068d#diff-3c5d685b5eb2cef5dc750a2d507238457b6b7f72e0965240f6528c74713453a8L34-R39

It is no longer creating clusterAuthToken due to the new logic.

https://github.com/rancher/rancher/issues/33771

### DIFF
--- a/pkg/controllers/managementuser/clusterauthtoken/token.go
+++ b/pkg/controllers/managementuser/clusterauthtoken/token.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"sort"
 
+	"github.com/rancher/rancher/pkg/controllers/managementuser/clusterauthtoken/common"
 	clusterv3 "github.com/rancher/rancher/pkg/generated/norman/cluster.cattle.io/v3"
 	managementv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


dramich (approved):

rmweir (changes requested):
It should be creating it in the api layer https://github.com/rancher/rancher/pull/30853/files#diff-19e1b4bcc3b3129050f1840c3b5afd40923d5e4cfb073cd246da5962bb592ae4R62. This was done due to token hashing. Here is an explanation I wrote for @thedadams awhile ago: `what the handler used to do is it it would get the token and then make a cluster auth token out of it. Now, by this point the token is hashed, so it would just be using the hashed value to create another token which would be incorrect. The cluster auth token then gets hashed, so you end up with this doubled hashed token value. So now the auth token is created when the regular token is created and the raw value is still in memory`.
I'm guessing this doesn't work because steve doesn't use the validators. So we probably need to discuss this further to figure out a solution that will work with token hashing.

StrongMonkey:
@rmweir Ok good to know that I missed this part. Sounds like there is no good way to fix this since it is not possible to know the original token. Maybe we should figure out something otherwise the local auth endpoint feature will likely be broken.

dramich:
@rmweir Should some of your explanation live in that code?

cbron:
@rmweir @StrongMonkey can we close this now ?

rmweir:
@cbron I think so.

PR #33840, opened 2021-07-29
Add CVEs csv as asset

rmweir:
Download v2.6 CVEs csv to include in assets.

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -8,6 +8,7 @@ SYSTEM_CHART_REPO_DIR=build/system-charts
 SYSTEM_CHART_DEFAULT_BRANCH=${SYSTEM_CHART_DEFAULT_BRANCH:-"dev-v2.6"}
 CHART_REPO_DIR=build/charts
 CHART_DEFAULT_BRANCH=${CHART_DEFAULT_BRANCH:-"dev-v2.6"}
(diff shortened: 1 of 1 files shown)


scripts/package
```diff
@@ -73,6 +73,9 @@ if [ ${ARCH} == amd64 ]; then
     # Move this out of ARCH check for local dev on non-amd64 hardware.
     TAG=$TAG REPO=${REPO} go run ../pkg/image/export/main.go $SYSTEM_CHART_REPO_DIR $CHART_REPO_DIR/assets $IMAGE $AGENT_IMAGE $RUNTIME_IMAGE
 
+    # CVEs CSV
+    curl -sL https://raw.githubusercontent.com/rancher/image-scanning/main/docs/_data/cves-v2.6.csv -o rancher-v2.6-cves.csv
```
kinarashah, line 77: nit: might want to use args like how we do for $CHART_REPO_*.

kinarashah (dismissed):

scripts/package
cbron: We should rename that file to `${TAG}_CVE_SCAN_RESULTS` or similar.

kinarashah (approved):

cbron (approved):

Issue #33743, opened 2021-07-22
User is not redirected to Cluster explorer on a rancher upgrade


sowmyav27:
**Rancher Server Setup**
- Rancher version: master-head commit id: `9e0d1b2`
- Installation option (Docker install/Helm Chart): Helm install
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1

**Information about the Cluster**
- Kubernetes version: 1.20.7

**Describe the bug**
User is not redirected to Cluster explorer on a rancher upgrade

**To Reproduce**

- On 2.5.9 , deploy a downstream cluster (not required) in the Ember UI
- Upgrade rancher to master-head.
- User is still in Ember UI on the upgraded setup

**Expected Result**
User should be redirected to Cluster Explorer, since that is the default UI from 2.6.0

Jono-SUSE-Rancher:
I'm paraphrasing what @vincent99 said in our chat as a comment here for whoever picks this up:
The most simple effective solution would bet to drop all existing UI session (tokens) once upon the first startup of 2.6, and anybody who was logged in will go back to the login page and then into Vue from there.

vincent99:
Yes the simplest solution I can think of is to delete all UI session tokens (there's an annotation identifying them as UI), just once, upon first startup of 2.6.

The UI can't really do much on it's own here because the user already has some existing old version loaded and running in their browser, and we haven't gone back into the past to define a way for it to know it should do anything.

snasovich:
PR: https://github.com/rancher/rancher/pull/33849

deniseschannon:
Available to test in v2.6.0-rc1

snasovich:
Sending back to "Working" as we need UI change as UI stores the last path the user was at in browser storage. After they log back in they still get redirected to Ember-based UI because that was their last know path. @vincent99, please let us know when you have an update on this.

PR #33849, opened 2021-07-30
[v2.6.0] ISSUE-33743 delete dashboard tokens on 2.6 upgrade

paynejacob:
https://github.com/rancher/rancher/issues/33743

# PROBLEM
With v2.6.0 rancher is shipping a new frontend.  Any user that is on the old ember ui while rancher is upgraded will not be directed to the new dashboard until they choose to logout.  This could cause some confusion after the upgrade.

# SOLUTION
When we upgrade to v2.6.0 we need to invalidate all the ember ui tokens.  This will force a logout and send all the users to a central point allowing the frontend to redirect them.  When the rancher app starts we will invalidate all the frontend tokens and then create a config map with the current rancher version.  On the next startup we will check for that configmap and as long as the stored version is greater than v2.6.0 we will not invalidate any tokens.

# TESTING

- [x] installed rancher v2.5.8 on single node cluster with helm and logged into the ember ui.  Redeploy the helm chart with an image with these changes that thinks it is v2.6.0. Tokens were invalidated.  After that the rancher pod was restarted and no tokens were invalidated. 
- [x] installed rancher v2.5.8 on an ha cluster with helm and logged into the ember ui.  Redeploy the helm chart with an image with these changes that thinks it is v2.6.0. Tokens were invalidated.  After that the rancher pod was restarted and no tokens were invalidated.
- [x] installed rancher v2.5.8 with fake v2.6.0 image on single node cluster with helm and logged into the ember ui.  Restart the rancher pods. Tokens were not invalidated.
- [x] tested docker run with container restart.

### DIFF
--- a/pkg/rancher/migrations.go
+++ b/pkg/rancher/migrations.go
@@ -1,0 +1,75 @@
+package rancher
+
+import (
+	"github.com/mcuadros/go-version"
+	"github.com/rancher/rancher/pkg/auth/tokens"
+	v3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
+	rancherversion "github.com/rancher/rancher/pkg/version"
+	controllerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	cattleNamespace          = "cattle-system"
+	forceUpgradeLogoutConfig = "forceupgradelogout"
+	rancherVersionKey        = "rancherVersion"
+)
+
+// forceUpgradeLogout will delete all dashboard tokens forcing a logout.  This is useful when there is a major frontend
+// upgrade and we want all users to be sent to a central point.  This function will check for the `forceUpgradeLogoutConfig`
+// configuration map and only run if the last migrated version is lower than the given `migrationVersion`.
+func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, tokenController v3.TokenController, migrationVersion string) error {
+	cm, err := configMapController.Cache().Get(cattleNamespace, forceUpgradeLogoutConfig)
+	if err != nil && !k8serror.IsNotFound(err) {
+		return err
+	}
+
+	// if this is the first ever migration initialize the configmap
+	if cm == nil {
+		cm = &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      forceUpgradeLogoutConfig,
+				Namespace: cattleNamespace,
+			},
+			Data: make(map[string]string, 1),
+		}
+	}
+
+	// we only want to migrate if the previous installation was earlier than the migration version
+	if lastMigration, ok := cm.Data[rancherVersionKey]; ok {
+		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
+			return nil
+		}
+	}
+
+	logrus.Infof("Detected %s upgrade, forcing logout for all web users", migrationVersion)
+
+	// list all tokens that were created for the dashboard
+	allTokens, err := tokenController.Cache().List(labels.SelectorFromSet(labels.Set{tokens.TokenKindLabel: "session"}))
+	if err != nil {
+		logrus.Error("Failed to list tokens for upgrade forced logout")
+		return err
+	}
+
+	// log out all the dashboard users forcing them to be redirected to the login page
+	for _, token := range allTokens {
+		err = tokenController.Delete(token.ObjectMeta.Name, &metav1.DeleteOptions{})
+		if err != nil && !k8serror.IsNotFound(err) {
+			logrus.Errorf("Failed to delete token [%s] for upgrade forced logout", token.Name)
+		}
+	}
+
+	// record the migration being completed
+	cm.Data[rancherVersionKey] = rancherversion.Version
+	if cm.ObjectMeta.GetResourceVersion() != "" {
+		_, err = configMapController.Update(cm)
+	} else {
+		_, err = configMapController.Create(cm)
+	}
+
+	return err
+}
--- a/pkg/rancher/rancher.go
+++ b/pkg/rancher/rancher.go
@@ -225,9 +225,12 @@ func (r *Rancher) Start(ctx context.Context) error {
 		if err := dashboarddata.Add(ctx, r.Wrangler, localClusterEnabled(r.opts), r.opts.AddLocal == "false", r.opts.Embedded); err != nil {
 			return err
 		}
-		return r.Wrangler.StartWithTransaction(ctx, func(ctx context.Context) error {
-			return dashboard.Register(ctx, r.Wrangler)
-		})
+
+		if err := r.Wrangler.StartWithTransaction(ctx, func(ctx context.Context) error { return dashboard.Register(ctx, r.Wrangler) }); err != nil {
+			return err
+		}
+
+		return forceUpgradeLogout(r.Wrangler.Core.ConfigMap(), r.Wrangler.Mgmt.Token(), "v2.6.0")
 	})
 
 	if err := r.authServer.Start(ctx, false); err != nil {


pkg/rancher/migrations.go
```diff
@@ -0,0 +1,79 @@
...
+const (
+	cattleNamespace        = "cattle-system"
+	upgradeMigrationConfig = "migrated"
+	forceLogoutKey = "forcedLogout"
+)
+
+// forceUpgradeLogout will delete all dashboard tokens forcing a logout.  This is useful when there is a major frontend
+//upgrade and we want all users to be sent to a central point.  This function will check for the `upgradeMigrationConfig`
+// configuration map and only run if the last migrated version is lower than the given `migrationVersion`.
+func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, tokenController v3.TokenController, migrationVersion string) (err error) {
```
dramich, line 25: Don't do naked returns on a large function, it makes following the logic more complicated and easier to miss what is actually going to get returned.

pkg/rancher/rancher.go
```diff
@@ -225,9 +225,14 @@ func (r *Rancher) Start(ctx context.Context) error {
...
 		}
-		return r.Wrangler.StartWithTransaction(ctx, func(ctx context.Context) error {
+		err := r.Wrangler.StartWithTransaction(ctx, func(ctx context.Context) error {
 			return dashboard.Register(ctx, r.Wrangler)
 		})
+		if err != nil {
+			return err
+		}
+
+		return forceUpgradeLogout(r.Wrangler.Core.ConfigMap(), r.Wrangler.Mgmt.Token(), "v2.6.0")
```
dramich, line 235:
```suggestion
		return forceUpgradeLogout(r.Wrangler.Core.ConfigMap().Cache(), r.Wrangler.Mgmt.Token().Cache(), "v2.6.0")
```

pkg/rancher/rancher.go
```diff
@@ -225,9 +225,14 @@ func (r *Rancher) Start(ctx context.Context) error {
 		if err := dashboarddata.Add(ctx, r.Wrangler, localClusterEnabled(r.opts), r.opts.AddLocal == "false", r.opts.Embedded); err != nil {
 			return err
 		}
-		return r.Wrangler.StartWithTransaction(ctx, func(ctx context.Context) error {
+		err := r.Wrangler.StartWithTransaction(ctx, func(ctx context.Context) error {
```
dramich, line 228: This can be done inline following the call pattern above

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,79 @@
...
+			logrus.Debug("failed to delete tokens for upgrade forced logout")
+			return
+		}
+	}
+
+	// record the migration being completed
+	upgradeMigrationConfigMap.Data[forceLogoutKey] = rancherversion.Version
+	_, err = configMapController.Create(upgradeMigrationConfigMap)
+	if k8serror.IsAlreadyExists(err) { // if someone else created the configmap we can just get it and check our value
+		return forceUpgradeLogout(configMapController, tokenController, migrationVersion)
```
dramich, line 75: I don't think the recursive call is needed, if someone else already created it then the cleanup was already done so just exit

paynejacob: This was a bit of forward thinking that we may want to use the migration configmap for other upgrade actions in the future.  I can rename it to `forcedLogout` if we want to make it single use instead.

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		Data: upgradeMigrationConfigMap.Data,
+	}
+
+	// we only want to migrate if the previous installation was earlier than the migration version
+	if lastMigration, ok := upgradeMigrationConfigMap.Data[forceLogoutKey]; ok {
+		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
+			return
+		}
+	}
```
dramich, line 51: This section is awkward, get the config map, check the value and exit. Removing the data defeats the purpose of getting the existing version and also doesn't allow k8s to block rancher from overwriting the same change. 
If the configmap doesn't exist then create the object with the value as expected. If it does exist then update the value and save. 
This is also moving to a cache so a deepcopy will be required
There is no update call so if it exists the create will always fail.

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+const (
+	cattleNamespace          = "cattle-system"
+	forceUpgradeLogoutConfig = "forceUpgradeLogout"
+	rancherVersionKey        = "rancherVersion"
+)
+
+// forceUpgradeLogout will delete all dashboard tokens forcing a logout.  This is useful when there is a major frontend
+//upgrade and we want all users to be sent to a central point.  This function will check for the `forceUpgradeLogoutConfig`
```
snasovich, line 23: Major nitpick, but please add space `// upgrade` so it aligns with other lines.

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+				Namespace: cattleNamespace,
+			},
+			Data: map[string]string{},
+		}
+	}
+
+	// we only want to migrate if the previous installation was earlier than the migration version
+	if lastMigration, ok := cm.Data[rancherVersionKey]; ok {
+		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
+			return err
```
snasovich, line 45: `return nil` is much clearer

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+				Name:      forceUpgradeLogoutConfig,
+				Namespace: cattleNamespace,
+			},
+			Data: map[string]string{},
+		}
+	}
+
+	// we only want to migrate if the previous installation was earlier than the migration version
+	if lastMigration, ok := cm.Data[rancherVersionKey]; ok {
+		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
```
snasovich, line 44: I'm sure you've checked this, but does `version.Compare` return false if `lastMigration` is `nil`?

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+			return err
+		}
+	}
+
+	logrus.Infof("detected %s upgrade, forcing logout for all web users", migrationVersion)
```
snasovich, line 49: Another nit, but log messages generally start with a capital letter (applicable to this and other logging below)

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// list all tokens that were created for the dashboard
+	allTokens, err := tokenController.Cache().List(labels.SelectorFromSet(labels.Set{tokens.TokenKindLabel: "session"}))
+	if err != nil {
+		logrus.Debug("failed to list tokens for upgrade forced logout")
```
snasovich, line 54: Recommend switching this to Error message. If debug logging is not enabled, even if returned error is logged by caller func, it may be hard to understand what it relates to. Same is applicable to another log message below.

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+
+	// log out all the dashboard users forcing them to be redirected to the login page
+	for i := 0; i < len(allTokens); i++ {
+		err = tokenController.Delete(allTokens[i].ObjectMeta.Name, &metav1.DeleteOptions{})
+		if err != nil && !k8serror.IsNotFound(err) {
+			logrus.Debug("failed to delete tokens for upgrade forced logout")
+			return err
```
snasovich, line 63: If there was an error invalidating a single token, I think it's best to proceed with invalidating the remaining ones vs. stopping the process. The next opportunity to process will happen at restart only which would generally be too late.

pkg/rancher/rancher.go
snasovich: @paynejacob , I don't see this addressed. Was it discussed with @dramich and determined the change is not needed?

snasovich (changes requested):

pkg/rancher/migrations.go
paynejacob: it will never be nil because we check if the key exists on line 43.

pkg/rancher/rancher.go
paynejacob: It was addressed in the body of the function.  I need read and write operations so I use the cache internally for the read operations.  If I pass the cache here I would not be able to write changes.

pkg/rancher/migrations.go
snasovich: `cm` will not be nil. But if it is, we create it with empty `Data` map. So `cm.Data[rancherVersionKey]` will be empty string (sorry, I misspoke about `lastMigration` being `nil`). So does `version.Compare` return `false` if second version is empty?

pkg/rancher/rancher.go
snasovich: Ah, I see. Thank you for the explanation.

pkg/rancher/migrations.go
paynejacob: `if lastMigration, ok := cm.Data[rancherVersionKey]; ok {` checks if the given key exists.  If it is an empty data map `ok` will be `false` and we will never get to the point where we compare the version.

snasovich: D'oh... Not sure how I missed this. Need more coffee I guess. :)

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,76 @@
...
+		logrus.Error("Failed to list tokens for upgrade forced logout")
+		return err
+	}
+
+	// log out all the dashboard users forcing them to be redirected to the login page
+	for i := 0; i < len(allTokens); i++ {
+		err = tokenController.Delete(allTokens[i].ObjectMeta.Name, &metav1.DeleteOptions{})
+		if err != nil && !k8serror.IsNotFound(err) {
+			logrus.Error("Failed to delete token for upgrade forced logout")
+			logrus.Debugf("Failed to delete token [%s] for upgrade forced logout", allTokens[i].Name)
```
snasovich, line 63: If the concern is that token's `Name` contains sensitive info, we need to reduce logging to trace here. If that's not the case, we may want to just include the name into Error log message and log once instead of twice.

snasovich (approved):
A single comment about logging to address, otherwise LGTM.

snasovich (dismissed):
LGTM. @dramich, would be great if you could review it today. Thank you.

snasovich (approved):

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,75 @@
...
+		return err
+	}
+
+	// log out all the dashboard users forcing them to be redirected to the login page
+	for i := 0; i < len(allTokens); i++ {
+		err = tokenController.Delete(allTokens[i].ObjectMeta.Name, &metav1.DeleteOptions{})
+		if err != nil && !k8serror.IsNotFound(err) {
+			logrus.Errorf("Failed to delete token [%s] for upgrade forced logout", allTokens[i].Name)
+		}
+	}
```
dramich, line 64: Why is this not a `range` loop?

pkg/rancher/migrations.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// if this is the first ever migration initialize the configmap
+	if cm == nil {
+		cm = &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      forceUpgradeLogoutConfig,
+				Namespace: cattleNamespace,
+			},
+			Data: map[string]string{},
```
dramich, line 38: Maps should be initialized with `make()` unless they have data as part of that initialization

snasovich: @dramich , isn't this purely preference? https://stackoverflow.com/a/31064737/16564280
There are 100+ places in rancher/rancher code that initialize empty map with `map[string]string{}` syntax.

snasovich (approved):

pkg/rancher/migrations.go
dramich: It's a rancher coding standard and best practice, I don't review every PR that goes into rancher

dramich (approved):

PR #33879, opened 2021-08-02
Force to install charts when cluster is changed

StrongMonkey:
Addressed an issue where aks/eks/gke chart is not redeployed/upgraded when cluster is changed. https://github.com/rancher/rancher/issues/31592

This has potential overhead calls to check helm releases because cluster is changed frequently and will be triggering controller code very frequently.

### DIFF
--- a/pkg/catalogv2/system/system.go
+++ b/pkg/catalogv2/system/system.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
@@ -12,9 +13,12 @@ import (
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/rancher/rancher/pkg/api/steve/catalog/types"
 	catalog "github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/catalogv2/content"
 	"github.com/rancher/rancher/pkg/catalogv2/helmop"
+	mgmtcontrollers "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	corev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	"github.com/rancher/rancher/pkg/settings"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/merr"
 	"github.com/sirupsen/logrus"
@@ -52,30 +56,35 @@ type desired struct {
 }
 
 type Manager struct {
-	ctx              context.Context
-	operation        *helmop.Operations
-	content          *content.Manager
-	restClientGetter genericclioptions.RESTClientGetter
-	pods             corecontrollers.PodClient
-	desiredCharts    map[desiredKey]map[string]interface{}
-	sync             chan desired
-	syncLock         sync.Mutex
+	ctx                   context.Context
+	operation             *helmop.Operations
+	content               *content.Manager
+	restClientGetter      genericclioptions.RESTClientGetter
+	pods                  corecontrollers.PodClient
+	desiredCharts         map[desiredKey]map[string]interface{}
+	sync                  chan desired
+	syncLock              sync.Mutex
+	refreshIntervalChange chan struct{}
+	settings              mgmtcontrollers.SettingController
 }
 
 func NewManager(ctx context.Context,
 	restClientGetter genericclioptions.RESTClientGetter,
 	contentManager *content.Manager,
 	ops *helmop.Operations,
-	pods corecontrollers.PodClient) (*Manager, error) {
+	pods corecontrollers.PodClient,
+	settings mgmtcontrollers.SettingController) (*Manager, error) {
 
 	m := &Manager{
-		ctx:              ctx,
-		operation:        ops,
-		content:          contentManager,
-		restClientGetter: restClientGetter,
-		pods:             pods,
-		sync:             make(chan desired, 10),
-		desiredCharts:    map[desiredKey]map[string]interface{}{},
+		ctx:                   ctx,
+		operation:             ops,
+		content:               contentManager,
+		restClientGetter:      restClientGetter,
+		pods:                  pods,
+		sync:                  make(chan desired, 10),
+		desiredCharts:         map[desiredKey]map[string]interface{}{},
+		refreshIntervalChange: make(chan struct{}, 1),
+		settings:              settings,
 	}
 
 	return m, nil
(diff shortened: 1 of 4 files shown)


StrongMonkey:
With the current framework, it will try to upgrade the chart every 15 minutes. Seemin-efficient that we need to check and redeploy the helm chart once cluster is changed. cc @cbron @rmweir

cmurphy:
This does not seem to trigger when a KEv2 cluster is deleted, should it? If I delete the apps, and if they do not automatically get recreated right away (sometimes the controller sync recreates them), and then I delete a cluster, it goes ahead and deletes the cluster object and removes the finalizers without the operator ever getting reinstalled and without deleting the cluster from the provider.

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -76,7 +76,7 @@ func Register(ctx context.Context, wContext *wrangler.Context) error {
 }
 
 func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster, error) {
-	if cluster == nil || cluster.DeletionTimestamp != nil {
```
StrongMonkey, line 79: @cmurphy This should address your concern.

cmurphy (approved):

StrongMonkey:
This PR has been modified according to discussion with @rmweir and @cbron 
1. Now all the system feature charts(kev2 operator/fleet/webhook) are synced and being up-to-date every 15 minutes(which is controlled by a setting called `system-feature-chart-refresh-minutes`). We no longer sync kev2 operator based on cluster change event, as it will eventually be picked up by the latest operator through this auto-update mechanism. 
2. If any kev2 operator chart is uninstalled manually, it will be redeployed on next cluster edit/deletion event. It should not redeploy kev2 operator chart if it has been manually removed and there is no corresponding kev2 clusters.
cc @rmweir @cbron

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -134,6 +138,23 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
...
 
+// check helm release secrets for aks/eks/gke operator chart, if it has been uninstalled, removed it in m.manager.desiredChart
+// so that we don't automatically redeploy it unless there is an AKS/EKS/GKE cluster trigerring it
+func (h handler) onSecretChange(key string, obj *corev1.Secret) (*corev1.Secret, error) {
+	if obj == nil {
+		ns, name := kv.Split(key, "/")
+		if ns == namespace.System {
+			// the name will follow the format sh.helm.release.v1.rancher-eks-operator-crd.v1
+			parts := strings.Split(name, ".")
+			if len(parts) == 6 {
```
cmurphy, line 149: Couldn't this accidentally match other things? Maybe this could be changed to a regex match to match on `/sh.helm.release.v1.rancher-.*-operator(-crd)?.v1/`

StrongMonkey: Oh you are right, regex might be overkill. I can do a name matching.

cmurphy (changes requested):

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -134,6 +138,33 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
...
+				}
+			}
+		}
+	}
+	return obj, nil
+}
+
+func isOperatorChartRelease(name string) bool {
+	switch name {
+	case AksCrdChart.ReleaseNamespace, AksChart.ReleaseNamespace, EksCrdChart.ReleaseNamespace, EksChart.ReleaseNamespace, GkeChart.ReleaseNamespace, GkeCrdChart.ReleaseNamespace:
```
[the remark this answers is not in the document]
StrongMonkey: whoops my bad, fixed it.

cmurphy (dismissed):

rmweir (approved):
LGTM

StrongMonkey:
@cmurphy @rmweir Force pushed to resolve conflicts.

cmurphy (dismissed):

pkg/catalogv2/system/system.go
```diff
@@ -84,14 +93,27 @@ func NewManager(ctx context.Context,
 func (m *Manager) Start(ctx context.Context) {
 	m.ctx = ctx
 	go m.runSync()
+
+	m.settings.OnChange(ctx, "system-feature-chart-refresh", m.onSetting)
```
cbron, line 97: I forget, these all fire on startup yes ? Looks like it starts the interval change below, and that should fire it off the bat and then wait the interval ?

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -134,6 +138,33 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 	return cluster, nil
 }
 
+// check helm release secrets for aks/eks/gke operator chart, if it has been uninstalled, removed it in m.manager.desiredChart
+// so that we don't automatically redeploy it unless there is an AKS/EKS/GKE cluster trigerring it
```
cbron, line 142: nit
```suggestion
// so that we don't automatically redeploy it unless there is an AKS/EKS/GKE cluster triggering it
```

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -134,6 +138,33 @@ func (h handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 	return cluster, nil
 }
 
+// check helm release secrets for aks/eks/gke operator chart, if it has been uninstalled, removed it in m.manager.desiredChart
```
cbron, line 141: nit
```suggestion
// check helm release secrets for aks/eks/gke operator chart, if it has been uninstalled then remove it in m.manager.desiredChart
```

pkg/controllers/dashboard/hostedcluster/controller.go
```diff
@@ -58,25 +61,26 @@ type handler struct {
 	appCache     controllerprojectv3.AppCache
 	apps         controllerprojectv3.AppController
 	projectCache controllerv3.ProjectCache
-	secrets      v1.SecretCache
+	secrets      v1.SecretController
```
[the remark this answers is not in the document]
StrongMonkey: It is also used to watch secrets onChange.

pkg/catalogv2/system/system.go
StrongMonkey: This doesn't fireup on restart though. On restart, everything will be empty(since it is memory) and controller kicks in to add desiredCharts and then manager will reinstall/upgrade all the charts. This is just for watching the settings so that when setting changes, we change the time ticker here. https://github.com/rancher/rancher/pull/33879/files/90c2409439f32318c0552a5836a83c696fa75f36#diff-96f7551d1bcafc092b3502857d33f73f576c753e3a4311b8660fabc446b383b0R115-R116

cbron: This was wrong line to add that comment. So the manager `runSync` method down there on line 109 is what will fire at startup.

cbron (approved):

Issue #33852, opened 2021-07-30
Keycloak OIDC - Improve logging for refresh token


izaac:
**Rancher Server Setup**
- Rancher version: `2.6 master-head` _94bca99_ or later
- Installation option : Any
- Proxy/Cert Details: `byo-valid` (but shouldn't matter for this issue)

**Describe the bug**
- When a refresh token request if made to extend the `Access Token Lifespan`, there's no indication in the Rancher server trace logs that this was done.

**To Reproduce**
- Enable Keycloak OIDC
- Enable Rancher server trace logging
- In the keycloak OIDC backend make sure to use a low value of `Access Token Lifespan`
- Wait more than the time value of the Token Lifespan
- Do a search in Auth Provider site access members input or in a cluster RBAC members add input.

**Result**

No logs about Keycloak actions are observed in the logs.
When a token is expired I see Errors logged but not for other tokens/sessions actions like token refresh

**Expected Result**
- Trace logs about session refresh in the Rancher server logs

cbron:
Assigning ryan for dev-validate.

cbron:
Ready for dev-validate

ryansann:
This has been validated on rancher `v2.6-head (f6120e0)` 

Here are is the output that shows the refresh log:
![Screen Shot 2021-08-09 at 1 00 40 PM](https://user-images.githubusercontent.com/8824599/128767156-e42101f9-c03e-4b98-b009-492d8da005b9.png)

PR #33896, opened 2021-08-03
Add oidc log message to indicate token refresh

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/33852

**Problem**
There is nothing in the logs to indicate if the original token was being used, or if a fresh token had been requested

**Solution**
Add a log message that indicates a refresh will be attempted if the token is no longer valid

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_provider.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_provider.go
@@ -59,6 +59,12 @@ func newClient(config *v32.OIDCConfig, token *oauth2.Token) (*KeyCloakClient, er
 		return nil, err
 	}
 	oauthConfig := oidc.ConfigToOauthConfig(provider.Endpoint(), config)
+	// Valid will return false if access token is expired
+	if !token.Valid() {
+		// since token is not valid, the TokenSource func used in the Client func will attempt to refresh the access token
+		// if the refresh token has not expired
+		logrus.Debugf("[keycloak oidc] newClient: attempting to refresh access token")
+	}
 	keyCloakClient := &KeyCloakClient{
 		httpClient: oauthConfig.Client(ctx, token),
 	}
--- a/pkg/auth/providers/oidc/oidc_provider.go
+++ b/pkg/auth/providers/oidc/oidc_provider.go
@@ -381,6 +381,13 @@ func (o *OpenIDCProvider) getUserInfo(ctx *context.Context, config *v32.OIDCConf
 			return userInfo, oauth2Token, err
 		}
 	}
+	// Valid will return false if access token is expired
+	if !oauth2Token.Valid() {
+		// since token is not valid, the TokenSource func will attempt to refresh the access token
+		// if the refresh token has not expired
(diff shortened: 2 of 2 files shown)


pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -59,6 +59,12 @@ func newClient(config *v32.OIDCConfig, token *oauth2.Token) (*KeyCloakClient, er
 		return nil, err
 	}
 	oauthConfig := oidc.ConfigToOauthConfig(provider.Endpoint(), config)
+	// Valid will return false if access token is expired
+	if !token.Valid() {
+		// since token is not valid, the TokenSource func used in the Client func will attempt to refresh the access token
+		// if the refresh token has not expired
+		logrus.Debugf("[keycloak oidc] newClient: attempting to refresh access token")
+	}
```
prachidamle, line 67: Where does the token get refreshed? Does the OIDC library handle that internally?

brendarearden: Anytime TokenSource is used, its part of the oauth2 library to perform a refresh if needed. So after I log the message, there is use of the func TokenSource.  I am open to suggestions on a better way to determine refresh! 

https://github.com/rancher/rancher/blob/0952b35cc2b86d72d432f4834b2fa8ca1dbe85d8/pkg/auth/providers/oidc/oidc_provider.go#L384
https://github.com/golang/oauth2/blob/2e8d9340160224d36fd555eaf8837240a7e239a7/clientcredentials/clientcredentials.go#L71

https://github.com/rancher/rancher/blob/0952b35cc2b86d72d432f4834b2fa8ca1dbe85d8/pkg/auth/providers/keycloakoidc/keycloak_provider.go#L63
https://github.com/golang/oauth2/blob/2e8d9340160224d36fd555eaf8837240a7e239a7/clientcredentials/clientcredentials.go#L60

prachidamle: Does that mean the token we carry if expired get refreshed internally, but we never get the refreshed token back? So we keep sending back the expired token?

brendarearden: https://github.com/rancher/rancher/blob/edc77f8e255087257987d0920a84430ebb61da91/pkg/auth/providers/googleoauth/goauth_provider.go#L406 it looks similar to what we are doing here in the google provider. But you are correct, this means that once the initial token expires, every time a call is made, a refresh will occur.

prachidamle: I wonder if we can call first to get refreshed token https://github.com/golang/oauth2/blob/2e8d9340160224d36fd555eaf8837240a7e239a7/oauth2.go#L298
If a different token is returned, we update it on our side as well as log that token refreshed message

cbron: Going to make a separate issue for this so we can fix that later. 
https://github.com/rancher/rancher/issues/33916

cbron (approved):

prachidamle (approved):

PR #33903, opened 2021-08-03
Update ldap library to v3

cmurphy:
Notes:
- the v3 major release coincided with an update to use Go modules[1] but
  is an otherwise backwards-compatible release. It was not related to
  compatibility with the LDAP V3 protocol, the ldap.v2 package was
  already compatible with the LDAP V3 protocol.
- the import path changed shortly after the v3 release[2], but it is the
  same package as before

[1] https://github.com/go-ldap/ldap/pull/247
[2] https://github.com/go-ldap/ldap/commit/c135faacbf0f8457d9cb53203f4a8baf1d3a31ab


dramich (approved):

dramich:
@cmurphy I am approving with the assumption that testing was done with an ldap provider to verify the upgrade hasn't caused any regressions.

cmurphy:
@dramich yes I tested this with an LDAP provider

PR #33907, opened 2021-08-03
Add impersonation namespace to system namespaces

cmurphy:
Backport of https://github.com/rancher/rancher/pull/33701

### DIFF
--- a/tests/integration/suite/test_system_project.py
+++ b/tests/integration/suite/test_system_project.py
@@ -54,7 +54,7 @@ def test_system_namespaces_assigned(admin_cc):
     if loggingNamespace in system_namespaces_names:
         system_namespaces_names.remove(loggingNamespace)
 
-    assert system_namespaces_names == initial_system_namespaces
+    assert initial_system_namespaces.issubset(system_namespaces_names)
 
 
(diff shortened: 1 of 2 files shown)


prachidamle (approved):

prachidamle:
We dint need this test change of v2.6?

cmurphy:
@prachidamle it passed without it :woman_shrugging:

tests/integration/suite/test_system_project.py
```diff
@@ -11,6 +11,7 @@
                                  "kube-public",
                                  "cattle-global-data",
                                  "cattle-global-nt",
+                                 "cattle-impersonation-system",
```
prachidamle, line 14: @cmurphy Maybe can there be a timing issue involved?
- Do we create this "cattle-impersonation-system" namespace during startup always? If yes then this change is needed in 2.6 also?
- Or if it depends on some other usecase to get created, are there chances that some testcase created it? Example this test handles the logging-namespace and removes it from the comparison.

cmurphy: I'm investigating. It looks like this test is run on the local cluster and the namespace shouldn't actually be created there, so we shouldn't need this.

prachidamle (approved):

PR #31783, opened 2021-03-22
Automation python pre-commit configuration

izaac:
https://github.com/rancher/qa-tasks/issues/34

- Added the Python pre-commit framework configuration file for files only under the `tests/validation` path.
- configuration files for `autopep8`, `flake8` and `black` under the `tests/validation`

Basic usage:
1. Clone the Rancher repository
2. Go to the `tests/validation` folder
3. `pip install -r requirements_v3api.txt` (or use your favorite environment manager)
4. `pre-commit install --install-hooks`
5. `pre-commit autoupdate` 

After this the pre-commit framework will detect changes in python files under `tests/validation` when a `git commit` is detected in the local repository.

```
git commit -m "test"
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check for added large files..............................................Passed
Check for merge conflicts................................................Passed
Detect AWS Credentials...................................................Passed
Detect Private Key.......................................................Passed
Fix End of Files.........................................................Passed
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted tests/validation/tests/v3_api/test_sa.py
All done! ✨ 🍰 ✨
1 file reformatted.

flake8...................................................................Passed
autopep8.................................................................Passed
```

A change was detected by `black` and applied to the file in the commit to fix it. Let's try  `git add` and `git commit` again with the applied changes to run the checks again.

```
 git commit -m "test2"
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check for added large files..............................................Passed
Check for merge conflicts................................................Passed
Detect AWS Credentials...................................................Passed
Detect Private Key.......................................................Passed
Fix End of Files.........................................................Passed
black....................................................................Passed
flake8...................................................................Passed
autopep8.................................................................Passed
[precommit-test 171771247] test2
 1 file changed, 1 deletion(-)
```

These hooks might prove challenging when applied to large files like `common.py` for the first time as lots of formatting warnings/errors could be detected and fixes applied automatically. Over time this will allow us to focus on the test logic and not much on formatting and coding style.

For the `pytest` version bump I tested that it kept working by running an HA test locally.

PR #33914, opened 2021-08-03
WIP: Add hostname prefix validation to ha deploy

catastrophe-brandon:
When running the HA deploy job, I included a hyphen in `RANCHER_HOSTNAME_PREFIX` value. This caused a bunch of stuff to run and do unnecessary work long before the test finally failed. It makes sense that we would want to validate configuration parameters such as RANCHER_HOSTNAME_PREFIX before doing any actual provisioning, deployments, etc.


catastrophe-brandon:
If this ever gets merged, https://github.com/rancher/rancher/pull/31783 should go first. A lot of the formatting changes here are a side effect of running black on the code.

PR #33943, opened 2021-08-04
Add generation field to ETCDSnapshotCreate

thedadams:
Previously, the UI would put a blank ETCDSnapshotCreate object on the
cluster spec when the user wanted to create an on-demand snapshot. This
caused an issue where multiple on-demand snapshots could not be created
this way because the backend would not detect any changes (the
ETCDSnapshotCreate object would remain blank).

After this change, the UI will update the generation field when a new
on-demand snapshot should be created and the backend will detect this.

Issue:
https://github.com/rancher/rancher/issues/32391

### DIFF
--- a/pkg/apis/rke.cattle.io/v1/etcd.go
+++ b/pkg/apis/rke.cattle.io/v1/etcd.go
@@ -16,6 +16,8 @@ type ETCDSnapshotCreate struct {
 	Name     string          `json:"name,omitempty"`
 	NodeName string          `json:"nodeName,omitempty"`
 	S3       *ETCDSnapshotS3 `json:"s3,omitempty"`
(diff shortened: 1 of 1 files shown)


jiaqiluo:
@thedadams  I have 2 questions:
- Is this specifically for RKE2 clusters? How about RKE1 clusters? 
- Does it affect the restoring of existing etcd snapshot in which ETCDSnapshotCreate is blank?

thedadams:
@jiaqiluo
1. Only for RKE1 clusters
2. Restoring an etcd snapshot where the ETCDSnapshotCreate is blank will set the generation to 0, which is a valid value.

jiaqiluo (approved):
LGTM

PR #33945, opened 2021-08-04
Add some validation for RANCHER_HOSTNAME_PREFIX

catastrophe-brandon:
Do the following to check if RANCHER_HOSTNAME_PREFIX is valid when automation jobs execute:
1. convert the input value to all lower case
2. If a hyphen is found, throw an error


catastrophe-brandon:
For context, this is just some minimal enforcement of restrictions described in the ha_deploy job.

image

Issue #33912, opened 2021-08-02
Unable to view resources in downstream clusters when my user is associated with a group from an external auth provider


deniseschannon:
Rancher Users Setup:

Rancher with GH enabled
Add `rancher` org as a group to the setup and given the role "admins"
Create a RKE2 cluster.

If I search for PodSecurityPolicies through cluster explorer, I can find the list page of all the policies in a cluster, but when I click on one of the them, the page never loads and gets stuck in "Loading..."

thedadams:
I am not able to reproduce this issue. After deploying a cluster and navigating to the PSP page, I can see the details of all PSPs.
![Screen Shot 2021-08-04 at 12 33 29 PM](https://user-images.githubusercontent.com/3682717/128243884-216fb8f0-9dff-4140-8132-d6dd98074b7e.png)

deniseschannon:
I updated the title and think this is more generic and I just happened to stumble upon this resource, but my user is associated to a group which seems to be causing failures for loading the resource in a downstream cluster

```
apiVersion: v1
code: 403
details:
  kind: groups
  name: github_org://9343010
kind: Status
message: ‘groups “github_org://9343010” is forbidden: User “system:serviceaccount:cattle-impersonation-system:cattle-impersonation-u-vnds56pccy”
  cannot impersonate resource “groups” in API group “” at the cluster scope’
metadata: {}
reason: Forbidden
status: Failure
```

deniseschannon:
With v2.6-head `0ab00d7`, validated with the setup that I was having issues and upgraded to the latest version. 

I am able to now view detailed PSPs.

deniseschannon:
@cmurphy Are there other use cases that I should attempt to validate?

bac-w:
I comfirm this issue
https://rancher.host/k8s/clusters/c-m-cf666213/apis/
```{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "groups \"openldap_group://CN=Admins,OU=K8s,DC=dc,DC=lan\" is forbidden: User \"system:serviceaccount:cattle-impersonation-system:cattle-impersonation-user-kv99d\" cannot impersonate resource \"groups\" in API group \"\" at the cluster scope",
  "reason": "Forbidden",
  "details": {
    "name": "openldap_group://CN=Admins,OU=K8s,DC=dc,DC=lan",
    "kind": "groups"
  },
  "code": 403
}
```

[...]

PR #33957, opened 2021-08-04
Update impersonation role based on user attributes

cmurphy:
Problem 1: When the cluster role for the impersonation service account
was created by the role template binding controller, it only added rules
for two basic authenticated groups and optionally a group derived from
the role template binding, and ignored potential groups the user is
associated with in an external provider. This meant that a user logging
in through an external provider would be blocked from impersonating
their service account due to insufficient permissions.

Solution 1: This change updates the role template binding controller to
get the true known groups for the user from the attribute lister.

Problem 2: If the user's Extras attributes changed, such as what happens
when a local user becomes associated with an external provider, the
roles' rules were never be updated. This meant that even though the
groups from the provider were up to date in the role rules, the
userextras/username may be mismatched. It also means that if the user's
groups in the external provider changed at some point after the initial
role creation, they would become mismatched with the allowed groups in
the role.

Solution 2: In the impersonation package, check that the groups and user
extras in the UserInfo object match the rules for groups and extras in
the role, and update them if they do not match. When the setup is run
from the role template binding controller, the user data will come from
the role template, and when the setup is run from the proxy server, the
user data comes from the request token.

https://github.com/rancher/rancher/issues/33912

### DIFF
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -2,6 +2,7 @@ package impersonation
 
 import (
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/rancher/rancher/pkg/types/config"
@@ -29,34 +30,40 @@ func New(user user.Info, clusterContext *config.UserContext) Impersonator {
 	return Impersonator{user: user, clusterContext: clusterContext}
 }
 
-func (i *Impersonator) SetUpImpersonation() (*corev1.ServiceAccount, error) {
-	sa, err := i.checkServiceAccount()
+func (i *Impersonator) SetUpImpersonation() error {
+	rules := i.rulesForUser()
+	role, err := i.checkAndUpdateRole(rules)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	if sa != nil {
-		return sa, nil
+	if role != nil {
+		return nil
 	}
 	err = i.createNamespace()
 	if err != nil {
-		return nil, err
+		return err
 	}
-	role, err := i.createRole()
+	role, err = i.createRole(rules)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	sa, err = i.createServiceAccount(role)
+	sa, err := i.createServiceAccount(role)
 	if err != nil {
-		return nil, err
+		return err
 	}
 	err = i.createRoleBinding(role, sa)
 	if err != nil {
-		return nil, err
+		return err
 	}
 	return i.waitForServiceAccount(sa)
 }
 
-func (i *Impersonator) GetToken(sa *corev1.ServiceAccount) (string, error) {
+func (i *Impersonator) GetToken() (string, error) {
+	name := ImpersonationPrefix + i.user.GetUID()
+	sa, err := i.clusterContext.Core.ServiceAccounts("").Controller().Lister().Get(impersonationNamespace, name)
+	if err != nil {
+		return "", err
+	}
 	if len(sa.Secrets) == 0 {
 		return "", fmt.Errorf("service account is not ready")
 	}
@@ -70,18 +77,6 @@ func (i *Impersonator) GetToken(sa *corev1.ServiceAccount) (string, error) {
 		return "", fmt.Errorf("error getting token: invalid secret object")
 	}
 	return string(token), nil
-}
-
-func (i *Impersonator) checkServiceAccount() (*corev1.ServiceAccount, error) {
-	name := ImpersonationPrefix + i.user.GetUID()
-	sa, err := i.clusterContext.Core.ServiceAccounts("").Controller().Lister().Get(impersonationNamespace, name)
-	if apierrors.IsNotFound(err) {
-		return nil, nil
-	}
-	if err != nil {
-		return nil, err
-	}
-	return sa, nil
 }
 
 func (i *Impersonator) createServiceAccount(role *rbacv1.ClusterRole) (*corev1.ServiceAccount, error) {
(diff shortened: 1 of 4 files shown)


pkg/impersonation/impersonation.go
```diff
@@ -182,10 +183,44 @@ func (i *Impersonator) createRole() (*rbacv1.ClusterRole, error) {
 			// in case cache isn't synced yet, use raw client
 			return i.clusterContext.RBAC.ClusterRoles("").Get(name, metav1.GetOptions{})
 		}
+		return r, nil
+	}
+	if !reflect.DeepEqual(rules, r.Rules) {
+		return i.updateRoleRules(r, rules[1], rules[2], rules[3])
```
ibuildthecloud, line 189: is there anything that guarentees that rules[2] and result[3] won't result in a panic?

pkg/impersonation/impersonation.go
```diff
@@ -182,10 +183,44 @@ func (i *Impersonator) createRole() (*rbacv1.ClusterRole, error) {
...
+				role.Rules[i].ResourceNames = addToListIfNotPresent(rn, role.Rules[i].ResourceNames)
+			}
+		}
+		if role.Rules[i].Resources[0] == "userextras/username" {
+			for _, rn := range extrasUsernameRule.ResourceNames {
+				role.Rules[i].ResourceNames = addToListIfNotPresent(rn, role.Rules[i].ResourceNames)
+			}
+		}
+	}
+	return i.clusterContext.RBAC.ClusterRoles("").Update(role)
```
ibuildthecloud, line 212: Since this is additive and not making it equal, will that result in a situation in which each proxy API call you call update?  Basically first call `rules(A,B) != rules(C)` so you add C, then next call `rules(A, B, C) != rules(C)` so you don't modify the rule but still do an Update call?

pkg/impersonation/impersonation.go
cmurphy: nope. fixed.

pkg/impersonation/impersonation.go
cmurphy: fixed

pkg/impersonation/impersonation.go
```diff
@@ -131,44 +132,44 @@ func (i *Impersonator) createNamespace() error {
...
+	}
+	if usernames, ok := extras["username"]; ok {
+		rules = append(rules, rbacv1.PolicyRule{
+			Verbs:         []string{"impersonate"},
+			APIGroups:     []string{"authentication.k8s.io"},
+			Resources:     []string{"userextras/username"},
+			ResourceNames: usernames,
+		})
+	}
+	role, err := i.clusterContext.RBAC.ClusterRoles("").Controller().Lister().Get("", name)
```
snasovich, line 169: Technically not part of the change itself, but it looks like `err` is not handled fully here - if `err` is not `IsNotFound` but is not `nil` we proceed as if there was no error.
Suggesting additional error handling to be added.

snasovich: Actually, disregard the "technically not part of the change itself" note. Previous code covered this error handling with `return r, err` at the end of func.

pkg/impersonation/impersonation.go
```diff
@@ -34,17 +35,17 @@ func (i *Impersonator) SetUpImpersonation() (*corev1.ServiceAccount, error) {
 	if err != nil {
 		return nil, err
 	}
+	role, err := i.checkAndCreateRole()
```
snasovich, line 38: It's weird how this breaks in-between error handling of errors during checking for SA and returning SA if it already exists. Could we move check/creation of role before we check for SA for more logical flow? Maybe there is a reason it has to be this way, but I'm failing to see it since `sa` is not used here.

cmurphy: I refactored this to use the role as the existence check so we don't need to have this double resource check.

pkg/impersonation/impersonation.go
cmurphy: fixed in the refactor

pkg/impersonation/impersonation.go
```diff
@@ -182,10 +162,46 @@ func (i *Impersonator) createRole() (*rbacv1.ClusterRole, error) {
...
+	if usernames, ok := extras["username"]; ok {
+		rules = append(rules, rbacv1.PolicyRule{
+			Verbs:         []string{"impersonate"},
+			APIGroups:     []string{"authentication.k8s.io"},
+			Resources:     []string{"userextras/username"},
+			ResourceNames: usernames,
+		})
+	}
+	return rules
+}
```
snasovich, line 204: Nit - add empty line before next func.

snasovich (dismissed):
LGTM (1 super minor nit).

@dramich @ryansann , please review this one with priority today as Colleen is out on Monday so we want this merged today if it's good to go.

snasovich (approved):

pkg/impersonation/impersonation.go
```diff
@@ -29,34 +30,39 @@ func New(user user.Info, clusterContext *config.UserContext) Impersonator {
 	return Impersonator{user: user, clusterContext: clusterContext}
 }
 
-func (i *Impersonator) SetUpImpersonation() (*corev1.ServiceAccount, error) {
-	sa, err := i.checkServiceAccount()
+func (i *Impersonator) SetUpImpersonation() error {
+	role, err := i.checkAndUpdateRole()
```
dramich, line 34: Can we create the rules here since we know we need them in either case? That saves us from creating them twice when we have to fall through to create

pkg/impersonation/impersonation.go
```diff
@@ -182,8 +162,45 @@ func (i *Impersonator) createRole() (*rbacv1.ClusterRole, error) {
 			// in case cache isn't synced yet, use raw client
 			return i.clusterContext.RBAC.ClusterRoles("").Get(name, metav1.GetOptions{})
 		}
+		return role, nil
+	}
+	return role, err
+}
+
+func (i *Impersonator) rulesForUser() []rbacv1.PolicyRule {
```
dramich, line 170: I like this vs the inline we were doing before.

pkg/impersonation/impersonation.go

dramich (approved):

snasovich (approved):

Issue #33950, opened 2021-08-04
Rancher will crash when attempt to create an RKE2 cluster with s3 bucket information


davidnuzik:
master-head `2b156fd2b`
single-node rancher server running in docker container
docker 19.03

**Steps:**
1. Create any RKE2 cluster which specifies s3 etcd backup information.
2. Rancher will crash very shortly after attempting to create this rke2 cluster (Clicking the Create button in the UI). In my case I tried to create an rke2 cluster with the Digital Ocean provider with 3 etcd nodes, 2 control plane, and 3 workers. I provided s3 bucket info like:
![image](https://user-images.githubusercontent.com/45179589/128260268-9c2c0e0a-4bb6-498b-98b5-9ea708d91bc4.png)

**Relevant logs:**

```
2021/08/04 21:55:52 [INFO] [mgmt-cluster-rbac-delete] Creating namespace c-m-2220cbc6                                                                                                                                                                                         
2021/08/04 21:55:52 [INFO] [mgmt-cluster-rbac-delete] Creating Default project for cluster c-m-2220cbc6                                                                                                                                                                       
2021/08/04 21:55:52 [INFO] [mgmt-project-rbac-create] Creating namespace p-sn8tv                                                                                                                                                                                              
2021/08/04 21:55:52 [INFO] [mgmt-cluster-rbac-delete] Creating System project for cluster c-m-2220cbc6                                                                                                                                                                        
2021/08/04 21:55:52 [INFO] [mgmt-cluster-rbac-delete] Updating cluster c-m-2220cbc6                                                                                                                                                                                           
2021/08/04 21:55:52 [INFO] [mgmt-project-rbac-create] Creating namespace p-sx4rf                                                                                                                                                                                              
2021/08/04 21:55:52 [INFO] [mgmt-project-rbac-create] Creating creator clusterRoleTemplateBinding for user user-ng5qz for cluster c-m-2220cbc6                                                                                                                                
2021/08/04 21:55:52 [INFO] [mgmt-project-rbac-create] Creating creator projectRoleTemplateBinding for user user-ng5qz for project p-sn8tv                                                                                                                                     
2021/08/04 21:55:52 [INFO] [mgmt-auth-crtb-controller] Creating role/clusterRole c-m-2220cbc6-clusterowner
2021/08/04 21:55:53 [INFO] [mgmt-auth-crtb-controller] Creating clusterRoleBinding for membership in cluster c-m-2220cbc6 for subject user-ng5qz
[...]

PR #33962, opened 2021-08-05
Fix S3 support for etcd snapshots

thedadams:
Previously, enabling S3 config for etcd snapshots caused Rancher to
panic because of an uninitialized struct.

Additionally, Rancher expected the cloud credential for the S3
credentials to either be of a certain type or in a certain namespace.
This is no longer needed and the check for this has been removed.

Issue:
https://github.com/rancher/rancher/issues/33950

### DIFF
--- a/pkg/provisioningv2/rke2/planner/s3args.go
+++ b/pkg/provisioningv2/rke2/planner/s3args.go
@@ -7,7 +7,6 @@ import (
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1/plan"
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2/machineprovision"
-	namespaces "github.com/rancher/rancher/pkg/namespace"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/kv"
 )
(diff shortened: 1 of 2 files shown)


paynejacob (approved):

brendarearden (approved):

pkg/provisioningv2/rke2/planner/s3args.go
```diff
@@ -88,11 +88,11 @@ func getS3Credential(secretCache corecontrollers.SecretCache, namespace, name, r
 	}
 
 	secret, err := machineprovision.GetCloudCredentialSecret(secretCache, namespace, name)
-	if err != nil {
-		return result, fmt.Errorf("failed to lookup etcdSnapshotCloudCredentialName: %w", err)
-	}
-	if secret.Type != "provisioning.cattle.io/cloud-credential" && secret.Namespace != namespaces.GlobalNamespace {
-		return result, fmt.Errorf("expected secret of type provisioning.cattle.io/cloud-credential, got [%s]", secret.Type)
+	if err != nil || (secret.Type != "provisioning.cattle.io/cloud-credential" && secret.Namespace != namespaces.GlobalNamespace) {
```
cbron, line 91: > Support is added for checking for a secret of the expected type or in the expected namespace.

You said "or" in the description but its using && here. Which is correct ? I'm confused by this conditional, couldn't the err exist and be a not found err but then you'd still check the secret.type ? 

Darren commented in other thread and it may not matter anymore.

thedadams: The secret needs to be of the expected type or in the expected namespace. So the negation of that is not the expected type and not in the expected namespace.

Also if the error is not nil, then the or will be true and we won't check the secret type.

thedadams: Now that I see Darren's comment then this doesn't matter. I will adjust it based on his comment.

cbron: Ahh right, was reading it backwards.

cbron (approved):

Issue #33861, opened 2021-07-31
AKSv2 cloud cred test functionality


cbron:
`/meta/aksLocations` requires tenant for AKSv1. We need to adjust or add a new endpoint that doesn't require tenant so AKSv2 can use it.

prabalsharma:
tested https://github.com/rancher/dashboard/issues/3661#issuecomment-894468272

PR #33970, opened 2021-08-05
Add new helper method /meta/aksCheckCredentials

cmurphy:
Add a new POST endpoint /meta/aksCheckCredentials that accepts AKS
credentials, minus tenant ID, as a POST request, and checks whether they
are valid. This allows the UI to validate the user's input when a cloud
credential or cluster is being created before the cloud credential
exists.

https://github.com/rancher/rancher/issues/33861

### DIFF
--- a/pkg/api/norman/customization/aks/handler.go
+++ b/pkg/api/norman/customization/aks/handler.go
@@ -1,6 +1,7 @@
 package aks
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -10,6 +11,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/gorilla/mux"
 	"github.com/rancher/aks-operator/pkg/aks"
+	"github.com/rancher/machine/drivers/azure/azureutil"
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
@@ -62,6 +64,20 @@ func NewAKSHandler(scaledContext *config.ScaledContext) http.Handler {
 func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	writer.Header().Set("Content-Type", "application/json")
 
+	resourceType := mux.Vars(req)["resource"]
+
+	if resourceType == "aksCheckCredentials" {
+		if req.Method != http.MethodPost {
+			handleErr(writer, http.StatusMethodNotAllowed, fmt.Errorf("use POST for this endpoint"))
+			return
+		}
+		if errCode, err := h.checkCredentials(req); err != nil {
+			handleErr(writer, errCode, err)
+			return
+		}
+		return
+	}
+
 	capa := &Capabilities{}
 
 	if credID := req.URL.Query().Get("cloudCredentialId"); credID != "" {
@@ -82,8 +98,6 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	var serialized []byte
 	var errCode int
 	var err error
-
-	resourceType := mux.Vars(req)["resource"]
 
 	switch resourceType {
 	case "aksVersions":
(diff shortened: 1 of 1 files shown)


pkg/api/norman/customization/aks/handler.go
```diff
@@ -126,6 +140,52 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+	}
+
+	if cred.SubscriptionID == "" {
+		return http.StatusBadRequest, fmt.Errorf("invalid subscriptionId")
+	}
+	if cred.ClientID == "" {
+		return http.StatusBadRequest, fmt.Errorf("invalid clientId")
+	}
+	if cred.ClientSecret == "" {
+		return http.StatusBadRequest, fmt.Errorf("invalid clientSecret")
```
rmweir, line 161: Since we haven't validated the values at this point and they could still be valid after this point maybe just say, "Must provide ..."

pkg/api/norman/customization/aks/handler.go
```diff
@@ -126,6 +140,52 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+		cred.AuthBaseURL = azure.PublicCloud.ActiveDirectoryEndpoint
+	}
+	client, err := NewSubscriptionServiceClient(cred)
+	if err != nil {
+		return http.StatusBadRequest, err
+	}
+	_, err = client.Get(ctx, cred.SubscriptionID)
+	if err != nil {
+		return http.StatusUnauthorized, err
+	}
```
rmweir, line 184: Would findTenantID succeed if this were to fail? Would it make sense to do this check first?

rmweir (changes requested):

pkg/api/norman/customization/aks/handler.go
cmurphy: No, this has to be the final step. This is just a dummy API call to test whether the credentials in the client are working. FindTenantID uses the subscription ID but no other credentials to look up the tenant ID, whereas this is using an authenticated client to make a generic GET request, and I happened to choose the subscriptions API to use as the test.

pkg/api/norman/customization/aks/handler.go
cmurphy: done

rmweir (approved):
LGTM, but I'm curious if anyone would be in favor of replacing most Errors from AKS api with generic auth errors so information cannot be extracted. Probably doesn't matter since anyone can just use the SDK to bypass us. @cbron

cmurphy:
> Probably doesn't matter since anyone can just use the SDK to bypass us

Yeah when I was working on this I found the errors from Azure through the SDK were very verbose, they tell you exactly whether the subscription is missing or the format is invalid or the client ID doesn't exist or the secret for it is unauthorized or whatever.

pkg/api/norman/customization/aks/handler.go
```diff
@@ -126,6 +140,52 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	if cred.TenantID == "" {
+		cred.TenantID, err = azureutil.FindTenantID(ctx, azure.PublicCloud, cred.SubscriptionID)
+		if err != nil {
+			return http.StatusBadRequest, err
+		}
+	}
+	if cred.BaseURL == "" {
+		cred.BaseURL = azure.PublicCloud.ResourceManagerEndpoint
```
cbron, line 172: One potential downfall of this whole optional tenant system is azure gov cloud. But we can cross that bridge when we get to it.

pkg/api/norman/customization/aks/handler.go
```diff
@@ -126,6 +140,52 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+	}
+	if cred.ClientSecret == "" {
+		return http.StatusBadRequest, fmt.Errorf("must provide clientSecret")
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	if cred.TenantID == "" {
+		cred.TenantID, err = azureutil.FindTenantID(ctx, azure.PublicCloud, cred.SubscriptionID)
+		if err != nil {
+			return http.StatusBadRequest, err
```
cbron, line 168: Lets make this error custom about getting the tenant id

thedadams (dismissed):

pkg/api/norman/customization/aks/handler.go
```diff
@@ -126,6 +140,52 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+	}
+	if cred.BaseURL == "" {
+		cred.BaseURL = azure.PublicCloud.ResourceManagerEndpoint
+	}
+	if cred.AuthBaseURL == "" {
+		cred.AuthBaseURL = azure.PublicCloud.ActiveDirectoryEndpoint
+	}
+	client, err := NewSubscriptionServiceClient(cred)
+	if err != nil {
+		return http.StatusBadRequest, err
```
cbron, line 179: On the errors convo, agree that its not huge concern, but long term I think we should obscure the errors more. And for this specific method we should make errors related to whether the creds are valid or not, so at this point and below just say "Invalid creds". The checks above this point about not passing enough params or invalid body are ok.

cbron (commented):
I like that we bypass other checks entirely.

pkg/api/norman/customization/aks/handler.go
cmurphy: done

pkg/api/norman/customization/aks/handler.go
cmurphy: done

cbron (approved):

Issue #33983, opened 2021-08-05
Double checksum flags are being added on RKE2 Windows Registration Command


phillipsj:
If this is a UI issue, please file an issue at https://github.com/rancher/dashboard/

**Rancher Server Setup**
- Rancher version: 2.6rc
- Installation option (Docker install/Helm Chart): docker

**Information about the Cluster**

RKE2 Custom cluster

**Describe the bug**
An extra and incorrect CA checksum flag is being added to the windows registration command.

```
curl.exe --insecure -fL https://<ip>/rke2-agent-install.ps1 -o install.ps1 
powershell.exe -ExecutionPolicy Bypass -Command "./install.ps1 -Server https://<ip> -Label 'cattle.io/os=windows' -Token <token> -Worker -CaChecksum  --ca-checksum <checksum>"
```

**To Reproduce**

Create an RKE2 custom cluster and register a server node, then look at the windows agent command.

slickwarren:
this issue has been fixed on master-head(61339fb)/2.6-head(ba95a58)
`curl.exe --insecure -fL https://<IP>/rke2-agent-install.ps1 -o install.ps1 && powershell.exe -ExecutionPolicy Bypass -Command "./install.ps1 -Server https://<IP> -Label 'cattle.io/os=windows' -Token <token> -Worker -CaChecksum <checksum>"

PR #33985, opened 2021-08-05
Correcting CA checksum flag for RKE2 windows registration command.

phillipsj:
An extra and incorrect CA checksum flag is being added to the windows registration command.

Issue: https://github.com/rancher/rancher/issues/33983

### DIFF
--- a/pkg/controllers/dashboard/clusterregistrationtoken/status.go
+++ b/pkg/controllers/dashboard/clusterregistrationtoken/status.go
@@ -37,11 +37,12 @@ func (h *handler) isRKE2(clusterID string) bool {
 }
 
 func (h *handler) assignStatus(crt *v32.ClusterRegistrationToken) (v32.ClusterRegistrationTokenStatus, error) {
-	ca := systemtemplate.CAChecksum()
+	checksum := systemtemplate.CAChecksum()
+	ca := ""
 	caWindows := ""
-	if ca != "" {
-		ca = " --ca-checksum " + ca
-		caWindows = " -CaChecksum " + ca
+	if checksum != "" {
+		ca = " --ca-checksum " + checksum
+		caWindows = " -CaChecksum " + checksum
 	}
 
 	token := crt.Status.Token


luthermonson (approved):

pkg/controllers/dashboard/clusterregistrationtoken/status.go
```diff
@@ -37,11 +37,12 @@ func (h *handler) isRKE2(clusterID string) bool {
 }
 
 func (h *handler) assignStatus(crt *v32.ClusterRegistrationToken) (v32.ClusterRegistrationTokenStatus, error) {
-	ca := systemtemplate.CAChecksum()
+	checksum := systemtemplate.CAChecksum()
+	ca := ""
```
nickgerace, line 41: What if the checksum is empty? Does the empty CA get consumed? If so, what implications does that have?

luthermonson: it ends up as an empty string at the end of the command, look at %s%s stuff in the sprintf format

nickgerace: I see that, but I meant more of the technical implications.  Just want to make sure because it looks like in the way that it's currently written, `ca` and `caWindows` will _always_ be assigned a non-empty value.

phillipsj: how? ca and caWindows will only be assigned a value if checksum isn't empty.

luthermonson: they were assigned to empty strings before. why would it always be non-empty? cachecksum sometimes `== ""` if we do --no-cacerts on rancher. the only fix here is to not add the suffix to the caWindows of `--ca-checksum` because ca was getting overwritten right above it.

nickgerace: Misread the original function that was called for CA assignment. LGTM thanks for the explanation

nickgerace (approved):

PR #33996, opened 2021-08-05
Add provisioning cluster permissions on PRTB

dramich:
Problem:
When a user is added to a project through a PRTB they are not able to
see the provisioning cluster object

Solution:
Add a handler to create a role and binding in the provisioning cluster
namespace granting permissions to get the cluster object.
This adds an owner reference to the cluster for the role and binding
with a cleanup in the PRTB lifecycle on remove to remove the binding
when the PRTB is deleted.

https://github.com/rancher/rancher/issues/32414

### DIFF
--- a/pkg/controllers/management/auth/crtb_handler.go
+++ b/pkg/controllers/management/auth/crtb_handler.go
@@ -6,12 +6,10 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
+	"github.com/rancher/rancher/pkg/controllers/management/authprovisioningv2"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	pkgrbac "github.com/rancher/rancher/pkg/rbac"
-	"github.com/rancher/wrangler/pkg/apply"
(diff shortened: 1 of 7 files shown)


cmurphy (dismissed):

thedadams (dismissed):

pkg/controllers/management/auth/crtb_handler.go
```diff
@@ -92,7 +89,7 @@ func (c *crtbLifecycle) Remove(obj *v3.ClusterRoleTemplateBinding) (runtime.Obje
 		return nil, err
 	}
 
-	err := c.removeAuthV2Permissions(obj)
+	err := c.mgr.removeAuthV2Permissions("auth-prov-v2-crtb-rolebinding", obj)
```
luthermonson, line 92: maybe this should be a constant in `authprovisioningv2` package?

luthermonson: could say similar for any of the `auth-prov-v2` prefixed strings this PR adds

kinarashah (dismissed):

luthermonson (approved):

Issue #33986, opened 2021-08-05
Update the Rancher post-delete hook. It's leaving some fleet, remnants on latest 2.6 rc1


izaac:
If this is a UI issue, please file an issue at https://github.com/rancher/dashboard/

**Rancher Server Setup**
- Rancher version:`2.6.0-rc1`
- Installation option:
   - HA RKE1 k8s 1.21 - CIS
- Proxy/Cert Details: N/A

**Describe the bug**

It seems that after doing `helm uninstall rancher -n cattle-system` fleet is not getting removed now.
This used to work on https://github.com/rancher/rancher/issues/30924

**To Reproduce**
- Install Rancher `v2.6.0-rc1`
- Then uninstall it with `helm uninstall rancher -n cattle-system`

**Result**

```
NAMESPACE             NAME                                      READY   UP-TO-DATE   AVAILABLE   AGE
cattle-fleet-system   deployment.apps/fleet-agent               1/1     1            1           13m
cattle-fleet-system   deployment.apps/fleet-controller          1/1     1            1           13m
cattle-fleet-system   deployment.apps/gitjob                    1/1     1            1           13m

NAMESPACE             NAME                                                 DESIRED   CURRENT   READY   AGE
cattle-fleet-system   replicaset.apps/fleet-agent-7c848bb648               0         0         0       13m
cattle-fleet-system   replicaset.apps/fleet-agent-b7688c69b                1         1         1       13m
cattle-fleet-system   replicaset.apps/fleet-controller-57c9c866f6          1         1         1       13m
cattle-fleet-system   replicaset.apps/gitjob-5778966b7c                    1         1         1       13m

NAMESPACE             NAME                                           READY   STATUS      RESTARTS   AGE
cattle-fleet-system   pod/fleet-agent-b7688c69b-8kmhr                1/1     Running     0          13m
cattle-fleet-system   pod/fleet-controller-57c9c866f6-8z22k          1/1     Running     0          13m
cattle-fleet-system   pod/gitjob-5778966b7c-gplrp                    1/1     Running     0          13m
```

**Expected Result**
Fleet also uninstalled

**Additional context**
https://github.com/rancher/rancher/issues/30924

cbron:
Believe we just need to add `cattle-fleet-system` here: https://github.com/rancher/rancher/blob/release/v2.6/chart/values.yaml#L148-L150

@nickgerace can confirm we want those all removed, I am under impression we do, and this bug was just part of namespace switch on fleet side.

[...]

PR #34001, opened 2021-08-05
add cattle-fleet-system to the namespace list

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/33986

`cattle-fleet-system` is the new namespace Fleet is installed in rancher v2.6, we need to add it to the `namespaceList` such that the post-delete-hook job will uninstall apps in this namespace 

`fleet-system` is removed because:
- it will not exist in a fresh installation of rancher v2.6
- and it will be cleaned up by Rancher in the case of upgrading Rancher from v2.5 to v2.6 

TODO:

- [ ] forward port to the master branch

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -145,7 +145,7 @@ postDelete:
     repository: %POST_DELETE_IMAGE_NAME%
     tag: %POST_DELETE_IMAGE_TAG%
   namespaceList:
-    - fleet-system
+    - cattle-fleet-system
     - cattle-system
     - rancher-operator-system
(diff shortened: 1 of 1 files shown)


StrongMonkey (approved):

nickgerace (approved):

cbron (approved):
As long as we are ok removing fleet-system this is ok

nickgerace:
@cbron this is fine because Fleet upgrades _should_ handle `fleet-system` cleanup. That being said, we do not delete `fleet-system` at the time, but the controller perform cleanup, which involves deleting components like the `fleet-agent`.

Issue #33991, opened 2021-08-05
Too many v3.Nodes created for RKE2/K3S cluster provisioned in Rancher


thedadams:
**Rancher Server Setup**
- Rancher version: 2.6-head
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: 1.19.6
- Cluster Type (Local/Downstream): Local
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE2/K3S

**Describe the bug**

After creating an RKE2/K3S cluster in Rancher, there are two many v3.Nodes in Rancher for the cluster.
**To Reproduce**
 
1. Provision an RKE2/K3S cluster from Rancher with 3 nodes with all roles.
2. Wait for the cluster to become active.

**Result**
Look at the old (Ember) UI. The number of nodes is likely to be greater than 3.

**Expected Result**

There should be exactly 3 v3.Nodes created for the cluster.

**Screenshots**

Vue UI:
![Screen Shot 2021-08-05 at 1 24 41 PM](https://user-images.githubusercontent.com/3682717/128416344-0410158a-a417-4021-8035-d8f40016409a.png)
Ember UI:
![Screen Shot 2021-08-05 at 1 25 19 PM](https://user-images.githubusercontent.com/3682717/128416373-8642826b-f4e2-40a5-a432-e0b617e74a8f.png)

**Additional context**
<!--Add any other context about the problem here.—>
This seems to only happen on HA installs (i.e. > 1 Rancher pod)

sowmyav27:
Dup of this - https://github.com/rancher/rancher/issues/33020

Issue #33020, opened 2021-06-03
Local cluster in a HA setup shows 4 nodes when only 3 nodes are in the cluster


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):**

**Steps to reproduce (least amount of steps as possible):** bug

**Result:**

- Deploy an RKE cluster using 1.3.0-rc3 - 3 nodes all roles.
- Deploy rancher on this using self signed certs (installation steps from rancher docs)
- When the rancher server comes up, local cluster is seen having 4 nodes instead of 3
- View in API of the local cluster in Ember UI shows 4 nodes.

**Other details that may be helpful:**

- This does _NOT_ happen always
- This is _not_ seen on 2.5.8

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): master-head
- Installation option (single install/HA): HA

Jono-SUSE-Rancher:
@sowmyav27 - Does doing anything with the certs have anything to do with this? Does this happen if you don't use self signed certs?
Can you confirm which specific commit of master you're on? Sorry this is just really hard to siphon into a bucket since it doesn't happen all the time and it doesn't happen in 2.5.8 which means it's a regression on master that we need to look at.

izaac:
I've sometimes seen this with k3s imported clusters, it doesn't happen all the time.

Steps:

- Create a single instance (CentOS 7.9, or CentOS 8.3)
- Install k3s `curl -sfL https://get.k3s.io | sh -`
- Import the cluster on Rancher

I sometimes see 1 or 2 duplicates of the same node on Ember only.

izaac:
![Screen Shot 2021-06-10 at 3 55 26 PM](https://user-images.githubusercontent.com/27659/121607228-85e52900-ca04-11eb-9258-69a778646b2a.png)

izaac:
![Screen Shot 2021-06-10 at 3 55 26 PM](https://user-images.githubusercontent.com/27659/121607283-9d241680-ca04-11eb-913f-7dc740c89367.png)

git-ival:
I've also encountered this during scale testing with k3s imported clusters on the updated UI.

In the screenshots below my downstream clusters should only be reporting 1 node and the local cluster should only be reporting 4 nodes. (The cluster health check errors are probably not relevant to this issue, they are an artifact of the scale testing).

[...]

PR #34002, opened 2021-08-06
Set PeerManager to run in clustered mode

thedadams:
Previously, the PeerManager in the wrangler context was not set. This
caused the cluster controllers to run on every node instead of in a
clustered/shard mode. The symptom of this was that too many v3.Nodes
were getting created when deploying RKE2/K3S clusters in an HA setup.
    
Now, the PeerManger will be set and the cluster controllers will run as
expected.

Issues:
https://github.com/rancher/rancher/issues/33991
https://github.com/rancher/rancher/issues/33020

### DIFF
--- a/pkg/wrangler/context.go
+++ b/pkg/wrangler/context.go
@@ -321,6 +321,7 @@ func NewContext(ctx context.Context, clientConfig clientcmd.ClientConfig, restCo
 		CachedDiscovery:         cache,
 		RESTMapper:              restMapper,
 		leadership:              leadership,
+		PeerManager:             peerManager,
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
Why is this logic running multiple nodes? The controllers are supposed to only run in one node in HA

thedadams:
Thanks @ibuildthecloud. I didn't understand that that should be the case. I will refocus my investigation to determining why that is.

thedadams:
@ibuildthecloud I found the reason why the the cluster controllers were running on every node. Please see the description for the explanation.

Issue #34033, opened 2021-08-07
login token for kubeconfig with rancher cli not being properly generated


rmweir:
If this is a UI issue, please file an issue at https://github.com/rancher/dashboard/

**Rancher Server Setup**
- Rancher version:
master-head (da1adad)

**Information about the Cluster**
- Kubernetes version:
local: v1.21.3+k3s1
- Downstream:
DO: v1.21.3

**Describe the bug**

When token has expired the kubectl with embedded CLI prompts for the token twice.

**To Reproduce**
 
1. provision DO cluster
2. set kubeconfig-generate-token setting to "false"
3. set kubeconfig-token-ttl-minutes to "1"
4. use generate kubeconfig
5. make requests like `kubectl get nodes` using kubeconfig, login on first attempt
6. after a little less than a minutes start trying to run the command again

**Result**
Will be prompt for token twice, this is because the first token returned was not actually generated causing rancher to return a 401 and the rancher cli to be ran again. The second time works because the original token has successfully deleted which is what is needed before the token can be created again.

Found this while investigating https://github.com/rancher/rancher/issues/33771.

**Expected Result**
Should be prompted to login once, and command should be successful. This should work for both contexts on the kubeconfig.

**Screenshots**

**Additional context**
<!--Add any other context about the problem here.—>

sowmyav27:
**Reproduced the issue on 2.6.0-rc2**

- provision DO cluster
- set kubeconfig-generate-token setting to "false"
- set kubeconfig-token-ttl-minutes to "1"
- use generate kubeconfig
- make requests like kubectl get nodes using kubeconfig, login on first attempt
- after a little less than a minutes start trying to run the command again
- User is prompted to login two times.

PR #34034, opened 2021-08-07
No gen token2

rmweir:
**Problems:**
1. When token has expired the kubectl with embedded CLI prompts for the token twice. Cause:
Sometimes create fails despite the delete call being made
prior. This is because it is in the process of being deleted.
Already exists errors were being ignored causing the caller
to receive a 201 even though the token was not created. It
was possible that the cluster auth token was created.
2. Cluster auth tokens are not being created when kubeconfig-generate-token setting is false.

**Solutions:**
1. Now, create will be retried if an already exists error is encountered. Controller error message now explains that cluster auth token cannot be found yet, insinuating that it is not abnormal if it is found the issue will be resolved.
2. Now, the api used by the rancher CLI attempts to create a cluster auth token.

**Notes:**
Some considerations around problem 1:
Updating the token was considered but this came with its own issues. Mainly, token expiration is calculated in multiple places as being creation timestamp + ttl. Making any changes to this widens the surface area of the fix significantly. This request is just for login tokens with kubeconfig response type.
Why retry create, doesn't that add to the roundtrip time? These are mainly used for the rancher cli and is a request that would only need to be made once a token has expired. The default ttl for these tokens is 16 hours.

**Issues:**
https://github.com/rancher/rancher/issues/33771
https://github.com/rancher/rancher/issues/34033

### DIFF
--- a/pkg/auth/providers/common/usermanager.go
+++ b/pkg/auth/providers/common/usermanager.go
@@ -278,8 +278,21 @@ func (m *userManager) EnsureClusterToken(clusterName, tokenName, description, ki
 		return "", err
 	}
 	logrus.Infof("Creating token for user %v", userName)
-	token, err = m.tokens.Create(token)
-	if err != nil && !apierrors.IsAlreadyExists(err) {
+	err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+		// Backoff was added here because it is possible the token is the process of deleting.
+		// This should cause the create to retry until the delete is finished.
+		newToken, err := m.tokens.Create(token)
+		if err != nil {
+			if apierrors.IsAlreadyExists(err) {
+				return false, nil
+			}
+			return false, err
+		}
+		token = newToken
+		return true, nil
+	})
+
+	if err != nil {
 		return "", err
 	}
 
--- a/pkg/auth/providers/publicapi/login.go
+++ b/pkg/auth/providers/publicapi/login.go
@@ -3,10 +3,12 @@ package publicapi
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
@@ -34,6 +36,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
@@ -215,7 +218,8 @@ func (h *loginHandler) createLoginToken(request *types.APIContext) (v3.Token, st
 		}
 
 		if err := h.createClusterAuthTokenIfNeeded(token, tokenValue); err != nil {
-			logrus.Infof("Failed to create cluster auth token for cluster [%s], cluster auth endpoint may fail: %v", token.ClusterName, err)
+			return v3.Token{}, "", "", httperror.NewAPIError(httperror.ServerError,
+				fmt.Sprintf("Failed to create cluster auth token for cluster [%s], cluster auth endpoint may fail: %v", token.ClusterName, err))
 		}
 		return *token, tokenValue, responseType, nil
 	}
@@ -225,7 +229,7 @@ func (h *loginHandler) createLoginToken(request *types.APIContext) (v3.Token, st
 }
 
 // createClusterAuthTokenIfNeeded checks if local cluster auth endpoint is enabled. If it is, a cluster auth token
-// is enabled.
+// is created.
 func (h *loginHandler) createClusterAuthTokenIfNeeded(token *v3.Token, tokenValue string) error {
 	clusterID := token.ClusterName
 	if clusterID == "" {
@@ -272,9 +276,22 @@ func (h *loginHandler) createClusterAuthTokenIfNeeded(token *v3.Token, tokenValu
 		}
 	}
 
-	if _, err = clusterContext.Cluster.ClusterAuthTokens("cattle-system").Create(clusterAuthToken); err != nil {
-		return err
+	var backoff = wait.Backoff{
+		Duration: 100 * time.Millisecond,
+		Factor:   1,
+		Jitter:   0,
+		Steps:    7,
 	}
 
-	return nil
+	err = wait.ExponentialBackoff(backoff, func() (bool, error) {
+		if _, err := clusterContext.Cluster.ClusterAuthTokens("cattle-system").Create(clusterAuthToken); err != nil {
+			if errors.IsAlreadyExists(err) {
+				return false, nil
+			}
+			return false, err
+		}
+		return true, nil
+	})
+
+	return err
 }
--- a/pkg/auth/providers/publicapi/login.go
+++ b/pkg/auth/providers/publicapi/login.go
@@ -8,31 +8,32 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


pkg/auth/providers/common/usermanager.go
```diff
@@ -278,8 +278,19 @@ func (m *userManager) EnsureClusterToken(clusterName, tokenName, description, ki
 		return "", err
 	}
 	logrus.Infof("Creating token for user %v", userName)
```
kinarashah, line 280: This was already present, but could we move it to log only if create is successful, otherwise log "Using existing token for user %v"? Would be easier to debug.

pkg/auth/providers/publicapi/login.go
```diff
@@ -208,9 +213,61 @@ func (h *loginHandler) createLoginToken(request *types.APIContext) (v3.Token, st
 		if err != nil {
 			return v3.Token{}, "", "", err
 		}
+
+		if err := h.createClusterAuthTokenIfNeeded(token, tokenValue); err != nil {
```
kinarashah, line 217: remove `IfNeeded` from the name

pkg/auth/providers/publicapi/login.go
```diff
@@ -208,9 +213,61 @@ func (h *loginHandler) createLoginToken(request *types.APIContext) (v3.Token, st
...
+			logrus.Infof("Failed to create cluster auth token for cluster [%s], cluster auth endpoint may fail: %v", token.ClusterName, err)
```
kinarashah, line 218: don't we want to return error too? for cli through kubectl if this actually goes through, when would users know the issue? 
Especially since we're not retrying, this might happen often.

pkg/auth/providers/common/usermanager.go
```diff
@@ -278,8 +278,19 @@ func (m *userManager) EnsureClusterToken(clusterName, tokenName, description, ki
...
+			if apierrors.IsAlreadyExists(err) {
+				return false, nil
+			}
+			return false, err
+		}
+		token = newToken
+		return true, nil
+	})
+
+	if err != nil {
```
cbron, line 293: Linking back to original comment: https://github.com/rancher/rancher/pull/33816#pullrequestreview-718634602

We could add a note here as well about why we added a backoff.

pkg/auth/tokens/token_util.go
```diff
@@ -125,6 +121,14 @@ func GetKubeConfigToken(userName, responseType string, userMGR user.Manager) (*v
...
 }
 
+func extractClusterIDFromResponseType(responseType string) string {
+	responseSplit := strings.SplitN(responseType, "_", 2)
+	if len(responseSplit) != 2 {
+		return ""
+	}
+	return responseSplit[1]
+}
+
```
cbron, line 131: Looks like nothing here actually changed.

pkg/controllers/managementuser/clusterauthtoken/token.go
```diff
@@ -30,11 +31,14 @@ type tokenHandler struct {
 }
 
 func (h *tokenHandler) Create(token *managementv3.Token) (runtime.Object, error) {
+	// clusterAuthToken is no longer created here because it requires the rawValue of the
+	// original token and the token could be hashed. Now, clusterAuthToken is created in
+	// the API layer right after token creation.
```
cbron, line 36: 👍

pkg/auth/providers/publicapi/login.go
cbron: Doesn't that help explain that its not always created ? If this was a cluster with ACE off it wouldn't be created.

pkg/auth/providers/publicapi/login.go
```diff
@@ -208,9 +213,61 @@ func (h *loginHandler) createLoginToken(request *types.APIContext) (v3.Token, st
...
+		}
 		return *token, tokenValue, responseType, nil
 	}
 
 	rToken, unhashedTokenKey, err := h.tokenMGR.NewLoginToken(currUser.Name, userPrincipal, groupPrincipals, providerToken, ttl, description)
 	return rToken, unhashedTokenKey, responseType, err
 }
+
+// createClusterAuthTokenIfNeeded checks if local cluster auth endpoint is enabled. If it is, a cluster auth token
+// is enabled.
```
cbron, line 228:
```suggestion
// is created.
```

pkg/auth/providers/publicapi/login.go
```diff
@@ -208,9 +213,61 @@ func (h *loginHandler) createLoginToken(request *types.APIContext) (v3.Token, st
...
+			return err
+		}
+	}
+
+	if _, err = clusterContext.Cluster.ClusterAuthTokens("cattle-system").Create(clusterAuthToken); err != nil {
+		return err
+	}
+
+	return nil
+}
```
cbron, line 273: I'm confused why this broken now with token hashing. I get that on the kubeconfig generation side, we added it in actions_kubeconfig.go because it has to be done at token generation time before it gets hashed. And we are roughly doing the same thing here, but we weren't creating an auth token here before, so was it already broken ?

pkg/auth/providers/publicapi/login.go
kinarashah: Maybe this is a pattern I am not familiar with, I thought IfNeeded didn't give additional context about ACE. So I was thinking createClusterAuthToken with comments or createClusterAuthTokenForAce - okay with keeping it as is too!

pkg/auth/providers/publicapi/login.go
rmweir: Before a handler (2.5) was taking care of it, after I made changes the handles can't do it anymore and I forgot to add it for login token.

pkg/auth/providers/publicapi/login.go
rmweir: I feel like createClusterAuth means it will definitely be created, which isn't true here.

pkg/auth/providers/publicapi/login.go
rmweir: The problem is that the regular token was created, so unless we clean it up here a 201 is accurate.

pkg/auth/providers/publicapi/login.go
rmweir: ? auth endpoint is a bool on the cluster object, I think enabled makes sense.

pkg/auth/providers/publicapi/login.go
cbron: Thats tough problem. Shouldn't we jitter the create below to ensure thats extreme edge case ? Otherwise we should discuss what to do here.

pkg/auth/providers/publicapi/login.go
cbron: If the _endpoint_ is enabled it _creates_ a token though right ? I don't understand why you would say the token was enabled

rmweir: Ohhh I see, I read the first sentence with "enabled". Good catch.

pkg/auth/providers/common/usermanager.go
rmweir: @kinarashah it either creates the token or it fails, there is no using the existing token since it could be hashed.

kinarashah: Ah. sorry I tend to forget we are now deleting existing tokens and no longer using them.

kinarashah (dismissed):

kinarashah (dismissed):

cbron (approved):

cbron (commented):
Actually real quick can you squash commits, then gtg

cbron (approved):

cbron (approved):

Issue #32719, opened 2021-05-13
snapshotbackpopulate controller handler triggers every second


superseb:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
Run Rancher with loglevel `trace` and add a custom cluster with a single node with all roles and wait for provisioning to complete.

**Result:**
See logs flooded with `snapshotbackpopulate` handler:

```
2021/05/13 19:17:59 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/cattle-controllers
2021/05/13 19:18:00 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/cattle-controllers
2021/05/13 19:18:01 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/k3s
2021/05/13 19:18:01 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/cattle-controllers
2021/05/13 19:18:02 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/cattle-controllers
2021/05/13 19:18:03 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/k3s
2021/05/13 19:18:03 [TRACE] ConfigMapController calling handler snapshotbackpopulate kube-system/cattle-controllers
```

This is because the handler is set on `kube-system` ConfigMap changes and `cattle-controllers` is used for leader election and triggers the handler. Introduced by https://github.com/rancher/rancher/blame/91a6f51f259acd5670ff167d6bcf1916da7406f1/pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go#L37

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): `master-c0508344405fb9bd55198d2196e93fe3fc33ade7-head`
- Installation option (single install/HA): single

Jono-SUSE-Rancher:
@kinarashah is this related to the controller issues we are seeing with the management agent?

snasovich:
@superseb , took initial look into it. Have some thoughts to share and could use some guidance.
[...]

PR #34043, opened 2021-08-07
Create a v1.Cluster indexer on startup.

thedadams:
Previously, the v1.Cluster indexer was created on the management plane,
but not on any of the followers. During sharding for the user management
plane, if the management plane was not selected, then the indexer would
not exist and could not be used. This would cause things like the
snapshotbackpopulate to not function.

Now, the indexer is created on startup so that every Rancher pod can
access it when necessary.

Related issue:
#32719

### DIFF
--- a/pkg/controllers/managementapi/controllers.go
+++ b/pkg/controllers/managementapi/controllers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/managementuser/rbac"
 	"github.com/rancher/rancher/pkg/controllers/managementuser/rbac/podsecuritypolicy"
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy/monitoring"
+	"github.com/rancher/rancher/pkg/controllers/provisioningv2/cluster"
 	"github.com/rancher/rancher/pkg/types/config"
 )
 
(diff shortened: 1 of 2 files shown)


dramich (commented):
This is not a controller so it shouldn't be called controller. The only change needed from the original pr was to move the registration into a function in the same file where the indexer was being registered. Make the below blob into a function that is called basically in the same place as the old pr then add comments that are explicit as to the purpose of the function so it can be used in the future.

```

	if features.RKE2.Enabled() {
		// The v1.Cluster indexer is created here because both management and usermanagement nodes need it.
		r.Wrangler.Provisioning.Cluster().Cache().AddIndexer(cluster.ByCluster, cluster.ByClusterIndex)
	}
```

pkg/rancher/rancher.go
```diff
@@ -221,6 +222,8 @@ func (r *Rancher) Start(ctx context.Context) error {
 		}
 	}
 
+	registerForLeaderAndFollower(r.Wrangler)
```
ibuildthecloud, line 225: You want to register this here: https://github.com/rancher/rancher/blob/b8ced38c3992c74ac365a7291d6424be622bfef9/pkg/controllers/managementapi/controllers.go#L38  This is where we register common indexers for all management nodes.

thedadams: Moved here.

ibuildthecloud (approved):

PR #34069, opened 2021-08-09
Allow 2.6 install of rancher in airgap automation

slickwarren:
Allow 2.6 install of rancher in airgap automation


sowmyav27 (approved):
LGTM

Issue #33979, opened 2021-08-05
RKE2 & RKE1 with calico and project network isolation breaks ingress controller to pod communication


bashofmann:
**Rancher Server Setup**
- Rancher version: v2.5.9 and v2.6.0-rc1
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): v1.21.3+k3s1
- Proxy/Cert Details: no proxy, Let's Encrypt certificate

**Information about the Cluster**
- Kubernetes version: RKE2: v1.21.3+rke2r1 and RKE: v1.20.9-rancher1-1
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): tested with custom and digital ocean infrastructure provider

**Describe the bug**
When you create an RKE1 or RKE2 cluster with calico as a CNI (default for RKE2) and activate project network isolation. The communication between the ingress controller and the Pods does not work anymore.

**To Reproduce**
* Create a multi node RKE2 custom cluster
* Choose default Calico CNI
* Activate project network isolation
* Create a workload (e.g.) an nginx and an ingress
* Try to reach the ingress

**Result**
The nginx ingress controller will get a gateway timeout when communication with the pod:

```
curl 35.158.100.209 -H 'Host: nginx.example'
<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

**Expected Result**
The communication works out of the box.

**Additional context**
The communication works out of the box, if you choose canal as a CNI. Rancher automatically creates a `hn-nodes` NetworkPolicy in every namespace, which should allow communication from the ingress controller. The policy works fine for canal, but not for calico.
If I understand it correctly, Rancher creates the policy from the node's podCIDRs:

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  creationTimestamp: "2021-08-05T13:28:45Z"
  generation: 3
  name: hn-nodes
  namespace: default
  ownerReferences:
  - apiVersion: v1
    kind: Namespace
    name: default
    uid: f6372a0d-28e0-4538-b649-ae8dd62355d8
  resourceVersion: "3157"
  uid: e31fe0a2-0f2c-43de-877b-450fb8b0cd01
spec:
  ingress:
[...]

PR #34082, opened 2021-08-09
[release/v2.6] Fix host network policies for Calico CNI (RKE1/2)

ryansann:
Issue: 
- https://github.com/rancher/rancher/issues/33979

This PR is needed for 2 reasons:
- Ingress traffic from nodes to workloads was not making it through due to misconfigured hn-nodes NetworkPolicies for RKE1/2 clusters deployed with Calico CNI
- Labels not always getting added to Rancher managed NetworkPolicy resources. Added explicit setting of labels so that we always properly select NetworkPolicy resources for deletion when disabling PNI

### DIFF
--- a/pkg/controllers/managementuser/networkpolicy/netpol.go
+++ b/pkg/controllers/managementuser/networkpolicy/netpol.go
@@ -7,9 +7,12 @@ import (
 	"sort"
 
 	"github.com/rancher/rancher/pkg/controllers/managementagent/nslabels"
+	cluster2 "github.com/rancher/rancher/pkg/controllers/provisioningv2/cluster"
+	rancherv1 "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	typescorev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	rnetworkingv1 "github.com/rancher/rancher/pkg/generated/norman/networking.k8s.io/v1"
+	rkecluster "github.com/rancher/rke/cluster"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	knetworkingv1 "k8s.io/api/networking/v1"
@@ -32,6 +35,8 @@ const (
 )
 
 type netpolMgr struct {
+	clusterLister    v3.ClusterLister
+	clusters         rancherv1.ClusterCache
 	nsLister         typescorev1.NamespaceLister
 	nodeLister       typescorev1.NodeLister
 	pods             typescorev1.PodInterface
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/networkpolicy/netpol.go
```diff
@@ -277,7 +356,10 @@ func generateDefaultNamespaceNetworkPolicy(aNS *corev1.Namespace, projectID stri
 		ObjectMeta: v1.ObjectMeta{
 			Name:      defaultNamespacePolicyName,
 			Namespace: aNS.Name,
-			Labels:    labels.Set(map[string]string{nslabels.ProjectIDFieldLabel: projectID}),
+			Labels: map[string]string{
+				nslabels.ProjectIDFieldLabel: projectID,
+				creatorLabel:                 creatorNorman,
```
kinarashah, line 361: are these not getting added again for a few policies?

pkg/controllers/managementuser/networkpolicy/netpol.go
```diff
@@ -236,6 +279,42 @@ func (npmgr *netpolMgr) handleHostNetwork(clusterNamespace string) error {
 	return nil
 }
 
+// getClusterCNI returns the cluster's CNI name if it is found or an api error
+func (npmgr *netpolMgr) getClusterCNI(clusterName string) (string, error) {
+	cluster, err := npmgr.mgmtClusters.Get(clusterName, v1.GetOptions{})
```
kinarashah, line 284: getClusterCNI will be called on each node sync, could we use Listers similar to isNetworkPolicyDisabled check

pkg/controllers/managementuser/networkpolicy/netpol.go
ryansann: Yea- I saw a few cases during testing where these labels weren't added. This resulted in some policies not being cleaned up when disabling PNI. I'm setting explicitly here so that we ensure all rancher managed nps get cleaned up when disabling PNI

pkg/controllers/managementuser/networkpolicy/netpol.go
ryansann: good point, I'll make the change and update

pkg/controllers/managementuser/networkpolicy/netpol.go
kinarashah: 👍🏼 could we double check on all the policies we create and add it for them as well? Good call on testing disabling PNI and catching this.

ryansann: Yep, I made sure all the ones we create are getting the label. I'll do another run through and make sure I got all of them

pkg/controllers/managementuser/networkpolicy/netpol.go
```diff
@@ -236,6 +279,42 @@ func (npmgr *netpolMgr) handleHostNetwork(clusterNamespace string) error {
...
+	if cluster.Spec.RancherKubernetesEngineConfig != nil {
+		return cluster.Spec.RancherKubernetesEngineConfig.Network.Plugin, nil
+	}
+
+	return "", nil
+}
+
+// getRKE2ClusterCNI returns the RKE2 cluster CNI name if it is found, or an api error
+func (npmgr *netpolMgr) getRKE2ClusterCNI(mgmtCluster *v3.Cluster) (string, error) {
+	cluster, err := npmgr.clusters.Get(mgmtCluster.Spec.FleetWorkspaceName, mgmtCluster.Spec.DisplayName, v1.GetOptions{})
```
kinarashah, line 302: I think we can use caches for wrangler clusters too, it's `Wrangler.Provisioning.Cluster().Cache()`

ryansann: Yes, you're right. Addressing this one too 👍

kinarashah (approved):

cbron (commented):
Not very familiar with this area but the approach makes sense, so lgtm

Issue #34074, opened 2021-08-09
rancher-images.txt for rancher 2.5.8 and 2.5.9 is missing docker.io/rancher/kubectl:v1.18.6 -> uninstall crd job fails in image pull due to that


Martin-Weiss:
**Rancher Server Setup**
- Rancher version: 2.5.8 or 2.5.9
- Installation option (Docker install/Helm Chart): helm chart, rke2 1.20.8

**Information about the Cluster**
- Kubernetes version: 1.20.8

**Describe the bug**
Uninstall CRD job does not execute because it wants to use image docker.io/rancher/kubectl:v1.18.6 which is not in the private registry, because it is not part of the rancher-images.txt for 2.5.8 and 2.5.9

**To Reproduce**
Air gapped setup
Only mirror the images from rancher-images.txt
Install a CRD
Uninstall a CRD -> job fails in image pull

**Result**
Uninstall CRD does not work

**Expected Result**
All required images should be in rancher-images.txt so that the CRD uninstall job works also in airgapped..

cbron:
@Martin-Weiss can you clarify which chart you are trying to uninstall, is it rancher itself ? Starting in 2.5.8 the rancher chart doesn't use `rancher/kubectl` for the uninstall, it uses the shell image from setting.go. 
You can see that here: `wget https://releases.rancher.com/server-charts/stable/rancher-2.5.8.tgz` -> look in values.yaml `postDelete.image.repository`. That is `rancher/shell:v0.1.6` which is in the airgap list. 

Was this an upgrade or a fresh install ? Are there any more error logs ?

Martin-Weiss:
The customer deployed the rancher-monitoring-crd 9.4 helm chart to the default namespace in a downstream cluster. When you now go to rancher apps and marketplace in explorer and ask for "uninstall" - the kubectl:v1.18.6 image is used.
I can also see that there is a kubectl:v1.18.0 image in the images.txt so not sure why we have that but not v1.18.6?

cbron:
Moving to 2.5.10, I believe we just need to backport a different fix.

jiaqiluo:
To be clear, The fix is to include the CRD charts when the package script generates the images.txt file. 
But it does not bring the image `rancher/kubectl:v1.18.6` back because the package script runs against only the latest version of each chart and the missing image is not used in the latest version of the monitoring v2 chart anymore.

brendarearden:
available to test in v2.5.10-rc1

[...]

PR #34024, opened 2021-08-06
[release/v2.6] remove the skipping of CRD charts

jiaqiluo:
#33995

remove the skipping of CRD charts which may contain images that need to be in the rancher-images.txt file for airgap

PR #34135, opened 2021-08-10
[release/v2.5] remove the skipping of CRD charts

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/34074
backport of https://github.com/rancher/rancher/pull/34024

remove the skipping of CRD charts which may contain images that need to be in the rancher-images.txt file for airgap

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -56,13 +56,9 @@ if [ ! -d $CHART_REPO_DIR ]; then
     for TGZ_PATH in $LATEST_TGZ_PATHS; do
         TGZ_REL_PATH=$CHART_REPO_DIR/$TGZ_PATH
         TGZ_EXTRACT_PATH=$(dirname $CHART_REPO_DIR/${TGZ_PATH##released/})
-        if [[ $TGZ_PATH == *crd*.tgz ]]; then
-            echo "Skipped CRD: $TGZ_REL_PATH"
-        else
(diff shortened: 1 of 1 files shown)


cbron:
@jiaqiluo can you do a test here so we get the before/after diff again. The feature charts changed a bit between 2.5 and 2.6.

cbron:
We don't need to add PR's to milestones, just issues.

jiaqiluo:
> @jiaqiluo can you do a test here so we get the before/after diff again. The feature charts changed a bit between 2.5 and 2.6.

I have tried to check the diff before/after my fix, but the outputs with/without my fix are the same when running against the feature chart (rancher/charts branch release-v2.5). I think it happens because 1) the package script runs against only the latest version of each chart and 2) all images in CRD charts are included in the non-CRD charts 

But, with my fix, we can see that the package script goes through CRD charts to generate the image list.
```
> cat rancher-images-sources.txt | grep crd 
rancher/kubectl:v1.20.2 fleet-agent:0.3.500,rancher-cis-benchmark:1.0.400,rancher-gatekeeper-crd:3.3.001,rancher-gatekeeper:3.3.001,rancher-monitoring:14.5.100
```
V.S. 
without my fix
```
> cat rancher-images-sources.txt | grep crd 

```

It can also prove that the fix is working.

cbron (approved):

Issue #34078, opened 2021-08-09
On an upgrade of rancher server, fleet agent gets deployed in the new namespace but with older version


sowmyav27:
**Rancher Server Setup**
- Rancher version: upgrade from 2.5.6 to 2.6-head commit id: `ba95a58`
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: Rancher generated

**Information about the Cluster**
- Kubernetes version: 1.20.9
- Cluster Type (Local/Downstream): Downstream RKE1 node driver clusters

**Describe the bug**
On an upgrade of rancher server, fleet agent gets deployed in the new namespace but with older version

**To Reproduce**

- on 2.5.9 deploy a couple of downstream RKE1 clusters
- Upgrade rancher  to 2.6-head 
- Fleet agent gets created in new namespace. But the version is older one - `0.3.5`
- It seems to have upgraded to the newer version in ~ 13-14 mins..

**Expected Result**
On an upgrade, newer version of fleet agent should get dpeloyed.

sowmyav27:
Blocked by https://github.com/rancher/fleet/issues/490

sowmyav27:
**On an upgrade from 2.5.9 to v2.6-head commit id: `c10ef4b0`**

- On 2.5.9 deploy a couple of downstream cluster
- In Continuous Delivery Deploy a Git repo
- It comes Up active and clusters in Active state
- On a rancher upgrade, New Fleet agent is deployed in `cattle-fleet-system` namespace, and the old one is deleted but the `fleet-system` namespace is still available.
- Existing clusters in Continuous delivery tab are Active. Able to deploy new bundles to the downstream clusters.
- Fleet version: `v0.3.6-rc12`

PR #34146, opened 2021-08-10
Use latest rancher-charts when available

nickgerace:
https://github.com/rancher/rancher/issues/34078


pkg/catalogv2/system/system.go
```diff
@@ -106,6 +114,16 @@ func (m *Manager) onSetting(key string, obj *v3.Setting) (*v3.Setting, error) {
 	return obj, nil
 }
 
+func (m *Manager) onTrigger(key string, obj *catalog.ClusterRepo) (*catalog.ClusterRepo, error) {
+	splitKey := strings.Split(key, "/")
```
StrongMonkey, line 118:
```suggestion
      if obj == nil || obj.Deletetimestamp != nil || obj.Name != "rancher-charts" {
        return obj, nil
      }
```

nickgerace: We do not need the key at all then?

StrongMonkey (approved):

kinarashah (approved):

Issue #33897, opened 2021-08-03
RKE2 Cluster Provisioning fails when a cluster with the same name existed before


bashofmann:
**Rancher Server Setup**
- Rancher version: master-head#30574c2
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): K3s v1.21.3+k3s1
- Proxy/Cert Details: No proxy, valid Let's Encrypt certificate

**Information about the Cluster**
- Kubernetes version: v1.21.3-rc1+rke2r2
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): infrastructure provider, tested DO and AWS

**Describe the bug**
RKE2 Cluster Provisioning fails when a cluster with the same name existed before.

**To Reproduce**
* Create a cluster at DigitalOcean with the name `dotest`
* Wait until the cluster has been created successfully
* Delete the cluster
* Create a new cluster at DigitalOcean with the same name `dotest`

**Result**
The second cluster fails with the error message

```
Error while applying agent YAML, it will be retried automatically: exit status 1, error: You must be logged in to the server (the server has asked for the client to provide credentials)
```

**Expected Result**
The second cluster is created successfully.

**Screenshots**
![Bildschirmfoto 2021-08-03 um 16 01 33](https://user-images.githubusercontent.com/243056/128029353-3f149100-a9f1-4762-8387-c82630d9f912.png)

thedadams:
@deniseschannon This is a known issue and should be fixed in https://github.com/rancher/fleet/issues/478

deniseschannon:
This should be available now to-test. @bashofmann Can you take a look at it? It woudl be available with the latest fleet update.

davidnuzik:
We should link the PR to this issue.

davidnuzik:
**My checks PASSED**
 
**Reproduction Steps:**

Not required as I fully understand this issue.
 
**Validation Steps:**

Rancher version: v2.6-head 4def65610 8/13/21 10:50 am Pacific
Rancher cluster type: single-node docker
Docker version: 19.03
 
Downstream cluster type: rke2
Downstream K8s version: v1.21.3-rc5+rke2r2
 
Validation steps:

1. Create an rke2 cluster, mine was named "dave-try-1node".
2. Wait until the cluster has been created successfully.
[...]

PR #34178, opened 2021-08-11
Generate v3.Cluster object name on first pass

thedadams:
Previously, the v3.Cluster object name was deterministic for a given
v1.Cluster based on the name and namespace. This caused an issue with
caches when the a cluster is created, deleted and then created with the
same name: the v3.Cluster object would have the same and would use the
same caches.

Now the v3.Cluster object's name is generated on first pass and then
pulled from the v1.Cluster object's status afterward.

Issue:
https://github.com/rancher/rancher/issues/33897

### DIFF
--- a/pkg/controllers/provisioningv2/cluster/controller.go
+++ b/pkg/controllers/provisioningv2/cluster/controller.go
@@ -2,8 +2,6 @@ package cluster
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"regexp"
 
 	"github.com/rancher/norman/types/convert"
@@ -23,6 +21,7 @@ import (
 	"github.com/rancher/wrangler/pkg/genericcondition"
 	"github.com/rancher/wrangler/pkg/kstatus"
 	"github.com/rancher/wrangler/pkg/name"
+	"github.com/rancher/wrangler/pkg/randomtoken"
 	"github.com/rancher/wrangler/pkg/relatedresource"
 	"github.com/rancher/wrangler/pkg/yaml"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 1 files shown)


paynejacob (approved):

pkg/controllers/provisioningv2/cluster/controller.go
```diff
@@ -225,9 +224,12 @@ func (h *handler) createCluster(cluster *v1.Cluster, status v1.ClusterStatus, sp
 	return h.createNewCluster(cluster, status, spec)
 }
 
-func mgmtClusterName(clusterNamespace, clusterName string) string {
-	hash := sha256.Sum256([]byte(clusterNamespace + "/" + clusterName))
-	return name.SafeConcatName("c", "m", hex.EncodeToString(hash[:])[:8])
+func mgmtClusterName() (string, error) {
```
StrongMonkey, line 227: This is worth to have @ibuildthecloud look at it. I think it has a reason to use hash to determine the name, and there might be code elsewhere that use `clusterNamespace` and `clusterName` to look for v3.Cluster from v1.cluster. Using the random name might break that use case. If not then this change might be fine.

thedadams: @StrongMonkey I did speak to him about this, but yes, I think it is worth him looking at it.

ibuildthecloud: It should be fine.  The only that that matters is that multiple runs of the controller produce the same name so that apply doesn't create/delete the cluster object.  All other code should reference status.ClusterName for the name relationship between prov->mgmt clusters.

ibuildthecloud (approved):

Issue #34107, opened 2021-08-10
system-agent running on downstream cluster cannot communicate back to Rancher server after restore


snasovich:
Found as part of working on https://github.com/rancher/rancher/issues/32599

**Rancher Server Setup**
- Rancher version: `master-head`
- Installation option (Docker install/Helm Chart): `rancher/rancher` code running locally against k3s cluster

**Information about the Cluster**
- Kubernetes version: v1.19.8+k3s1 (upstream), v1.21.3+rke2r1 (downstream)

**Describe the bug**
Following restore of backed up cluster with RKE2-provisioned downstream cluster, `system-agent` running on downstream cluster fails to communicate back to Rancher server as evident by 401s responses to requests it sends to Rancher API. Note that until the agent is restarted, the failing requests are to URLs like `PUT /api/v1/namespaces/fleet-default/secrets/{clustername}-bootstrap-template-{xxxxx}-machine-plan` and `system-agent` logs contain messages pointing that machine plan cannot be updated. If agent is restarted, the requests are to `GET /version` and there are errors `panic: error while connecting to Kubernetes cluster` in `system-agent` logs.
![image](https://user-images.githubusercontent.com/85187633/128797429-510aec49-4be0-41ca-a980-d65955737f37.png)

**To Reproduce**

1. Provision DigitalOcean RKE2 downstream cluster. The issue may not require RKE2-provisioned cluster, but was found with it.
2. Using "Rancher Backup" app perform a backup of the cluster. Note that restore of RKE2-provision clusters is being worked on as part of https://github.com/rancher/rancher/issues/32599 and so far the following needs to be added to Backup/Restore resourceset to backup related objects:
```yaml
   # Added for v2 provisioning
  - apiVersion: apiextensions.k8s.io/v1
    kindsRegexp: .
    resourceNameRegexp: provisioning.cattle.io$|rke-machine-config.cattle.io$|rke-machine.cattle.io$|rke.cattle.io$
  - apiVersion: provisioning.cattle.io/v1
    kindsRegexp: .
  - apiVersion: rke-machine-config.cattle.io/v1
    kindsRegexp: .
  - apiVersion: rke-machine.cattle.io/v1
    kindsRegexp: .
  - apiVersion: rke.cattle.io/v1
    kindsRegexp: .
  - apiVersion: apiextensions.k8s.io/v1
    kindsRegexp: .
    resourceNameRegexp: cluster.x-k8s.io$
[...]

PR #34181, opened 2021-08-11
Add ability to redeploy system-agent

StrongMonkey:
Context: https://github.com/rancher/rancher/issues/34107#issuecomment-899820635

### DIFF
--- a/pkg/apis/provisioning.cattle.io/v1/cluster.go
+++ b/pkg/apis/provisioning.cattle.io/v1/cluster.go
@@ -28,6 +28,8 @@ type ClusterSpec struct {
 	DefaultPodSecurityPolicyTemplateName string          `json:"defaultPodSecurityPolicyTemplateName,omitempty" norman:"type=reference[podSecurityPolicyTemplate]"`
 	DefaultClusterRoleForProjectMembers  string          `json:"defaultClusterRoleForProjectMembers,omitempty" norman:"type=reference[roleTemplate]"`
 	EnableNetworkPolicy                  *bool           `json:"enableNetworkPolicy,omitempty" norman:"default=false"`
+
+	RedeploySystemAgentGeneration int64 `json:"redeploySystemAgentGeneration,omitempty"`
 }
 
 type ClusterStatus struct {
--- a/pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
+++ b/pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
@@ -25,6 +26,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	generationSecretName = "system-agent-upgrade-generation"
 )
 
 type handler struct {
@@ -95,7 +100,7 @@ func (h *handler) OnChange(cluster *rancherv1.Cluster, status rancherv1.ClusterS
 		})
 	}
 
-	resources, err := ToResources(installer(len(cluster.Spec.RKEConfig.MachineSelectorConfig) == 0, secretName))
+	resources, err := ToResources(installer(len(cluster.Spec.RKEConfig.MachineSelectorConfig) == 0, secretName, strconv.Itoa(int(cluster.Spec.RedeploySystemAgentGeneration))))
 	if err != nil {
 		return nil, status, err
 	}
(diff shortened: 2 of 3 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/34107

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
```diff
@@ -148,6 +149,11 @@ func installer(allWorkers bool, secretName string) []runtime.Object {
 		})
 	}
 
+	env = append(env, corev1.EnvVar{
```
snasovich, line 152: Ideally we don't want this ENV at all if `redeploySystemAgentGeneration` is not set because it will result in a change of checksum and redeployment of system-agent if this change gets applied to cluster that was generated before the change. I know we never released this functionality, but it would be cleaner anyway to not add this env if it's not set on the Cluster CR. With current code I believe it would be set to 0.

pkg/apis/provisioning.cattle.io/v1/cluster.go
```diff
@@ -28,6 +28,8 @@ type ClusterSpec struct {
 	DefaultPodSecurityPolicyTemplateName string          `json:"defaultPodSecurityPolicyTemplateName,omitempty" norman:"type=reference[podSecurityPolicyTemplate]"`
 	DefaultClusterRoleForProjectMembers  string          `json:"defaultClusterRoleForProjectMembers,omitempty" norman:"type=reference[roleTemplate]"`
 	EnableNetworkPolicy                  *bool           `json:"enableNetworkPolicy,omitempty" norman:"default=false"`
+
+	RedeploySystemAgentGeneration int64 `json:"redeploySystemAgentGeneration,omitempty"`
```
snasovich, line 32: Alignment

StrongMonkey: This actually passed go fmt.

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
StrongMonkey: yeah, I can change that.

pkg/apis/provisioning.cattle.io/v1/cluster.go
snasovich: Yeah, didn't notice a newline - it's in own block so doesn't have to align.

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
```diff
@@ -211,6 +217,23 @@ func installer(allWorkers bool, secretName string) []runtime.Object {
 			},
 		},
 	}
+
+	if generation != "0" {
```
StrongMonkey, line 221: After talking to @dweomer , we can use secretRef to force job rerun. This doesn't need SUC pr to support that anymore.

pkg/provisioningv2/kubeconfig/manager.go
```diff
@@ -222,7 +222,7 @@ func (m *Manager) GetCRTBForAdmin(cluster *v1.Cluster, status v1.ClusterStatus)
 func (m *Manager) getKubeConfigData(clusterNamespace, clusterName, secretName, managementClusterName string) (map[string][]byte, error) {
 	secret, err := m.secretCache.Get(clusterNamespace, secretName)
 	if err == nil {
-		if secret.Data == nil || secret.Data["token"] == nil || secret.Annotations["objectset.rio.cattle.io/owner-gvk"] != "provisioning.cattle.io/v1, Kind=Cluster" {
```
StrongMonkey, line 225: Checking the annotation seems a little bit problematic during backup and restore as the secret can be created just with no annotation first https://github.com/rancher/rancher/blob/e201d2e02ead6884d789024c7fbacf0819b68363/pkg/provisioningv2/kubeconfig/manager.go#L282. This causes secret to be created and deleted back and forth and somehow the token inside it become invalid. @nickgerace I feel like we can remove this as checking `token` field should suffice.

nickgerace: Ahhh good call. That discrepancy makes sense. Let's do it.

nickgerace (approved):
LGTM with Sergey's convo

jiaqiluo (approved):
LGTM

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
```diff
@@ -211,6 +217,23 @@ func installer(allWorkers bool, secretName string) []runtime.Object {
...
+				Name:      generationSecretName,
+				Namespace: namespaces.System,
+			},
+			StringData: map[string]string{
+				"generation": generation,
+			},
+		})
+	}
+
+	return append(objs, plan)
```
snasovich, line 236: Previous code has `plan` as the first object in the result, this one would have it as last. This may not change anything functionally, but ideally we want to preserve the same order.

snasovich (changes requested):

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagent.go
StrongMonkey: :+1:

snasovich (approved):
LGTM, please get 1 committer approval (@dramich or @kinarashah).

kinarashah (approved):

Issue #34015, opened 2021-08-06
[2.6] Cluster stuck waiting for cluster-agent on docker install rancher behind ngrok


kinarashah:
Steps to reproduce:
- Run docker rancher container behind ngrok. (master-head or v2.6-head) 
- Add a RKE2 cluster. 
- Kube components come up and are running, but cluster is in "Waiting for cluster-agent" 
- Error from rancher logs: 
```
2021/08/06 17:01:41 [INFO] Failed to find providerID for selector rke.cattle.io/machine=64ff49b3-6cbe-4dce-8c3e-9ff0a4ff5d41 in cluster fleet-default/kinara-do-node, machine kinara-do-node-pool1-547ff8b89c-wfsf2: Get "https://ff073ca2ccf1.ngrok.io/k8s/clusters/c-m-585de23a/api/v1/nodes?labelSelector=rke.cattle.io%!F(MISSING)machine%!D(MISSING)64ff49b3-6cbe-4dce-8c3e-9ff0a4ff5d41": x509: certificate signed by unknown authority
```

Refer to https://github.com/rancher/rancher/issues/33878#issuecomment-893603177 for the root cause.

claudioser:
I can reproduce error  trying to provision on AWS with rancher 2.6 install on Azure  RKE1 Cluster

Oats87:
> I can reproduce error trying to provision RKE2 on AWS with Rancher 2.6 install on Azure RKE1 Cluster - build version - rancherImageTag=master-c58f95221d76c6e6a32c89426b300f8afd2ab69a-head
> 
> extract of log:
> 
> ```
> 2021/08/13 10:05:18 [INFO] Failed to find providerID for selector rke.cattle.io/machine=aa6d7bc3-0ae3-4936-acab-7a003311e6d4 in cluster fleet-default/rke2awscentral, machine rke2awscentral-po │
> │ ol1-8c64fd78b-hsfcd: {"Code":{"Code":"Forbidden","Status":403},"Message":"clusters.management.cattle.io \"c-m-1db6b728\" is forbidden: User \"u-pm63ybmili\" cannot get resource \"clusters\" i │
> │ n API group \"management.cattle.io\" at the cluster scope","Cause":null,"FieldName":""} (get nodes)
> ```

The 403 is a transient error that will eventually stop being reported once RBAC is reconciled.

aiyengar2:
Available to test in Rancher 2.6.0-rc3.

#### Steps to reproduce
[...]

PR #34211, opened 2021-08-12
Create rancher service for internal communication to prevent loopback through load balancer

Oats87:
There are certain cases (when running Rancher in a docker container) where we do not want Rancher to attempt to connect via the exposing load balancer. This PR will create a "rancher" service in cattle-system and use that if running an embedded K3s. 

https://github.com/rancher/rancher/issues/34015


brandond (dismissed):

nickgerace (dismissed):

nickgerace:
Assigned committer @kinarashah. Need a committer review before merging

nickgerace (approved):

kinarashah (approved):

Issue #34193, opened 2021-08-12
Restore fails on a rollback in the same rancher setup with error "Failed to find system chart fleet will try again in 5 seconds" and Rancher UI is not accessible


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.5.9 to 2.6.0-rc2 and restore to 2.5.9
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: Rancher generated

**Information about the Cluster**
- Kubernetes version: 1.20.9 (local cluster)
- Cluster Type (Local/Downstream): Downstream cluster - RKE1 custom cluster
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**Describe the bug**
Restore fails on a rollback in the same rancher setup with error "Failed to find system chart fleet will try again in 5 seconds" and Rancher UI is not accessible

**To Reproduce**

- On 2.5.9 HA deploy a downstream cluster RKE1
- Deploy Rancher backups and take backup b1
- Upgrade rancher to 2.6.0-rc2
- Create Projects/namespaces/deployment
- Take backup b2
- Restore to backup b1
- Rancher UI fails to load, 404 error seen on the uI
- Rancher pods throw error 

```
21/08/12 00:07:29 [INFO] Starting catalog controller
2021/08/12 00:07:29 [INFO] Starting project-level catalog controller
2021/08/12 00:07:29 [INFO] Starting cluster-level catalog controller
2021/08/12 00:07:29 [INFO] Refreshing driverMetadata in 1440 minutes
2021/08/12 00:07:29 [ERROR] Failed to find system chart fleet will try again in 5 seconds: configmaps "rancher-charts-0-e5cee34b-2da6-4e53-ad24-af838ac5b2a7" not found

```
- Rancher backup restore has finished restoring.

snasovich:
@sowmyav27 , is this a valid scenario? Backup b1 is taken on 2.5.9 and we're restoring it to cluster running Rancher 2.6.0. Per https://rancher.com/docs/rancher/v2.x/en/backups/v2.5/restoring-rancher/:
> When restoring a backup into a new Rancher setup, the version of the new setup should be the same as the one where the backup is made.
"New Rancher setup" is a bit confusing, but I think it means we shouldn't be using backup taken on earlier version to restore.

[...]

PR #34212, opened 2021-08-12
add checks for no-existing IndexConfigMapNamespace in repo.status

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/34193

Cause: https://github.com/rancher/rancher/issues/34193#issuecomment-897984502

Fix:
 add the logic to check and rebuild the missing IndexConfigMap: if the IndexConfigMap does not exist in the cluster, reset the value of `IndexConfigMapName`, `IndexConfigMapNamespace`, `IndexConfigMapResourceVersion` to `""` such that the method `shouldRefresh` will returns `true` and trigger a rebuild of the index file. 

TODO after merging:

- [ ] backport to 2.5
- [ ] forwardport to master

### DIFF
--- a/pkg/controllers/dashboard/helm/repo.go
+++ b/pkg/controllers/dashboard/helm/repo.go
@@ -20,6 +20,7 @@ import (
 	name2 "github.com/rancher/wrangler/pkg/name"
 	"helm.sh/helm/v3/pkg/repo"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
@@ -33,22 +34,25 @@ var (
 )
 
 type repoHandler struct {
-	secrets      corev1controllers.SecretCache
-	clusterRepos catalogcontrollers.ClusterRepoController
-	configMaps   corev1controllers.ConfigMapClient
-	apply        apply.Apply
+	secrets        corev1controllers.SecretCache
+	clusterRepos   catalogcontrollers.ClusterRepoController
+	configMaps     corev1controllers.ConfigMapClient
+	configMapCache corev1controllers.ConfigMapCache
+	apply          apply.Apply
 }
 
 func RegisterRepos(ctx context.Context,
 	apply apply.Apply,
 	secrets corev1controllers.SecretCache,
 	clusterRepos catalogcontrollers.ClusterRepoController,
-	configMap corev1controllers.ConfigMapController) {
+	configMap corev1controllers.ConfigMapController,
+	configMapCache corev1controllers.ConfigMapCache) {
 	h := &repoHandler{
-		secrets:      secrets,
-		clusterRepos: clusterRepos,
-		configMaps:   configMap,
-		apply:        apply.WithCacheTypes(configMap).WithStrictCaching().WithSetOwnerReference(false, false),
+		secrets:        secrets,
+		clusterRepos:   clusterRepos,
+		configMaps:     configMap,
+		configMapCache: configMapCache,
+		apply:          apply.WithCacheTypes(configMap).WithStrictCaching().WithSetOwnerReference(false, false),
 	}
 
 	catalogcontrollers.RegisterClusterRepoStatusHandler(ctx, clusterRepos,
@@ -75,6 +79,10 @@ func (r *repoHandler) ClusterRepoDownloadEnsureStatusHandler(repo *catalog.Clust
 }
 
 func (r *repoHandler) ClusterRepoDownloadStatusHandler(repo *catalog.ClusterRepo, status catalog.RepoStatus) (catalog.RepoStatus, error) {
+	err := r.ensureIndexConfigMap(repo, &status)
+	if err != nil {
+		return status, err
+	}
 	if !shouldRefresh(&repo.Spec, &status) {
 		r.clusterRepos.EnqueueAfter(repo.Name, interval)
 		return status, nil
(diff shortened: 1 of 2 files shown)


pkg/controllers/dashboard/helm/repo_test.go
```diff
@@ -200,11 +212,39 @@ func TestShouldRefresh(t *testing.T) {
...
+		},
 	}
 
+	mock := configMapMock{}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := shouldRefresh(tt.spec, tt.status)
+			rh := repoHandler{
+				configMaps: mock,
+			}
```
thedadams, line 246: Can this be moved outside the loop so that it is only initialized once?

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -196,27 +202,36 @@ func (r *repoHandler) download(repoSpec *catalog.RepoSpec, status catalog.RepoSt
...
+			// check if the IndexConfigMap exists
+			exist := true
+			_, err = r.configMaps.Get(status.IndexConfigMapNamespace, status.IndexConfigMapName, metav1.GetOptions{})
+			if err != nil && apierrors.IsNotFound(err) {
+				exist = false
+			}
+			if status.Commit == commit && exist {
+				status.DownloadTime = downloadTime
+				return status, nil
+			}
```
thedadams, line 227: I think it would be a good idea to consider the configmap as nonexistent if we get any error at all for the same of updating the `DownloadTime`.

```suggestion
			if err == nil && status.Commit == commit {
				status.DownloadTime = downloadTime
				return status, nil
			}
```

Also, if we get any other error besides `IsNotFound`, should we return the error and let the controller requeue?

thedadams (changes requested):

pkg/controllers/dashboard/helm/repo_test.go
jiaqiluo: fixed.

pkg/controllers/dashboard/helm/repo.go
jiaqiluo: > I think it would be a good idea to consider the configmap as nonexistent if we get any error at all for the same of updating the DownloadTime.

the line for setting the downloadTime is at the bottom of the method for all other cases except the `status.Commit == commit && exist` case and the failure case. 

> Also, if we get any other error besides IsNotFound, should we return the error and let the controller requeue?

Good catch! I added a check for it.

thedadams (approved):
One nit.

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -196,27 +202,40 @@ func (r *repoHandler) download(repoSpec *catalog.RepoSpec, status catalog.RepoSt
...
+				if apierrors.IsNotFound(err) {
+					exist = false
+				} else {
+					return status, err
+				}
+			}
+			if status.Commit == commit && exist {
+				status.DownloadTime = downloadTime
+				return status, nil
+			}
```
thedadams, line 231: nit: the `exists` variable is not needed.
```suggestion
			if err != nil {
				if !apierrors.IsNotFound(err) {
					return status, err
				}
			} else if status.Commit == commit {
				status.DownloadTime = downloadTime
				return status, nil
			}
```

thedadams (approved):
One nit.

pkg/controllers/dashboard/helm/repo.go
jiaqiluo: hmm, your suggestion will miss this case:
 update the downloadTime and return status if configMap exists and "status.Commit == commit"

jiaqiluo: updated.

thedadams (dismissed):

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -32,10 +34,14 @@ var (
...
 
+type configMapGetter interface {
+	Get(namespace, name string, options metav1.GetOptions) (*corev1.ConfigMap, error)
+}
+
 type repoHandler struct {
 	secrets      corev1controllers.SecretCache
 	clusterRepos catalogcontrollers.ClusterRepoController
-	configMaps   corev1controllers.ConfigMapClient
+	configMaps   configMapGetter
```
dramich, line 44: ConfigMapClient is already an interface so this isn't needed. There is also existing work around generating mocks for the interfaces used in the clients.

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -260,6 +275,10 @@ func shouldRefresh(spec *catalog.RepoSpec, status *catalog.RepoStatus) bool {
 	if status.IndexConfigMapName == "" {
 		return true
 	}
+	_, err := r.configMaps.Get(status.IndexConfigMapNamespace, status.IndexConfigMapName, metav1.GetOptions{})
```
dramich, line 278: This needs to use a cache.

dramich (changes requested):
Per our offline discussion this should be updated to do a check in the beginning of `ClusterRepoDownloadStatusHandler` to validate the config map status, and if it doesn't exist cleanup the status object so that the refresh can run again.

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -247,6 +255,23 @@ func (r *repoHandler) download(repoSpec *catalog.RepoSpec, status catalog.RepoSt
 	return status, nil
 }
 
+func (r *repoHandler) checkConfigMap(repo *catalog.ClusterRepo, status *catalog.RepoStatus) error {
```
snasovich, line 258: Suggesting to change func name to something that relays the fact that status may get changed. `ensureConfigMap` maybe?

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -247,6 +255,23 @@ func (r *repoHandler) download(repoSpec *catalog.RepoSpec, status catalog.RepoSt
...
+	if repo.Spec.GitRepo != "" && status.IndexConfigMapName != "" {
+		// check if the IndexConfigMap exists
```
snasovich, line 260: This comment is redundant as code below obviously does what the comment says.

snasovich (approved):
Couple nits, otherwise LGTM

pkg/controllers/dashboard/helm/repo.go
jiaqiluo: the name is updated to `ensureIndexConfigMap`

pkg/controllers/dashboard/helm/repo.go
jiaqiluo: the comment is removed

snasovich (approved):

thedadams (dismissed):

nickgerace (approved):

pkg/controllers/dashboard/helm/repo.go
```diff
@@ -247,6 +255,22 @@ func (r *repoHandler) download(repoSpec *catalog.RepoSpec, status catalog.RepoSt
 	return status, nil
 }
 
+func (r *repoHandler) ensureIndexConfigMap(repo *catalog.ClusterRepo, status *catalog.RepoStatus) error {
```
dramich, line 258: Add a comment explaining the purpose of this function and why we are dropping the fields if the configmap doesn't exist.

jiaqiluo: A comment is added.

dramich (approved):

Issue #34213, opened 2021-08-12
Update xip.io to sslip.io


deniseschannon:
With xip.io no longer available, we'll start using sslip.io.

Note: There will be no migration for anyone using xip.io.

sowmyav27:
On rancher 2.6.0-rc3 UI - "Automatically generate a .sslip.io hostname" is seen when creating an ingress.

PR #34215, opened 2021-08-12
Switch from xip.io to sslip.io

ibuildthecloud:


deniseschannon:
https://github.com/rancher/rancher/issues/34213

Issue #34149, opened 2021-08-10
User is logged out two times and had to login back after a rancher server upgrade to 2.6


sowmyav27:
**Rancher Server Setup**
- Rancher version: master-head commit id: `723f747b`
- Installation option (Docker install/Helm Chart): HELM
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: Rancher generated

**Information about the Cluster**
- Kubernetes version: 1.20.9 (Local cluster) 
- Downstream clusters - GKEv2 and RKE1 Custom cluster

**Describe the bug**
User is logged out two times after a rancher server upgrade to 2.6

**To Reproduce**

- On 2.5.9, HA install, deploy a few downstream clusters
- Upgrade (Helm upgrade) to 2.6-head
- User is redirected to dashboard page but is logged out
- Log back in, and navigate to Home page
- User is logged out one more time.
- After another logging in, user is able to proceed with using Rancher 

**Expected Result**
User be logged out only one time?

paynejacob:
@sowmyav27 did you see any pod restarts on the rancher pods in this setup?

deniseschannon:
@paynejacob This most likely is due to having HA (multiple Rancher server pods) so when Rancher server starts back up, each node is clearing the UI session tokens.

paynejacob:
The migration is only run on the leader pod and is gated by a configuration map after that.  I think what might have happened is the rancher pod crashed for some reason in the middle of the tokens being invalided and on restart it did not see the configmap that says it ran and started again.  I am testing some logic to prevent this.

dramich:
@paynejacob We are seeing this being reproduced by restarting rancher in an HA setup running 2.6-head. cc: @git-ival

paynejacob:
@dramich ill test my solution with this and see if I can still reproduce

dramich:
@paynejacob Repro the issue first, but I am seeing `2021/08/13 22:28:23 [INFO] Detected v2.6.0 upgrade, forcing logout for all web users` in the logs, so definitely hitting this.

paynejacob:
Here is the expected behavior:

anytime a helm upgrade occurs and the versions are the same users should not be logged out.

| dev          | dev   | *-head | < v2.6. 0-rc1 | > v2.6.0-rc1 |
|--------------|-------|--------|---------------|--------------|
| *-head       | false | true   | true          | true         |
[...]

PR #34220, opened 2021-08-12
ensure we only delete old tokens on force logout

paynejacob:

If the leader pod crashes before finishing the force logout or saving the configuration map users could be logged out twice or not at all.  This change ensures that all users will always be logged out on upgrade and any user that logs in after the upgrade will not be logged out again.

### DIFF
--- a/pkg/rancher/migrations.go
+++ b/pkg/rancher/migrations.go
@@ -7,6 +7,7 @@ import (
 	rancherversion "github.com/rancher/rancher/pkg/version"
 	controllerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/mod/semver"
 	v1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -39,9 +40,21 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
 		}
 	}
 
-	// we only want to migrate if the previous installation was earlier than the migration version
+	// we do not migrate in development environments
+	if rancherversion.Version == "dev" {
+		return nil
+	}
+
+	// if no last migration is found we run
 	if lastMigration, ok := cm.Data[rancherVersionKey]; ok {
-		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
+
+		// if a valid sem ver is found we only migrate if the version is less than the target version
+		if semver.IsValid(rancherversion.Version) && version.Compare(migrationVersion, lastMigration, "<") {
+			return nil
+		}
+
+		// if an unknown format is given we migrate any time the current version does not equal the last migration
+		if lastMigration == rancherversion.Version {
 			return nil
 		}
 	}


paynejacob:
I should also note I was unable to reproduce the senario that QA outlines in the ticket.  This is the only logical way I could see users being logged out twice.

pkg/rancher/migrations.go
```diff
@@ -39,15 +45,33 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
...
 			return nil
 		}
 	}
 
 	logrus.Infof("Detected %s upgrade, forcing logout for all web users", migrationVersion)
 
+	// save that we are starting a new migration
+	lastMigrationStart = time.Now().UnixNano()
+	cm.Data[rancherVersionKey] = rancherversion.Version
+	cm.Data[migrationStartedKey] = fmt.Sprintf("%d", lastMigrationStart)
```
snasovich, line 64: Why not `strconv.Itoa`? Same below for completion time.

snasovich (dismissed):
One minor comment, otherwise LGTM

pkg/rancher/migrations.go
```diff
@@ -39,15 +45,33 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
...
-		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
+		if (lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=")) && lastMigrationStart < lastMigrationCompleted {
 			return nil
 		}
 	}
 
 	logrus.Infof("Detected %s upgrade, forcing logout for all web users", migrationVersion)
 
+	// save that we are starting a new migration
+	lastMigrationStart = time.Now().UnixNano()
```
dramich, line 62: You get smaller numbers with just `Unix`
```suggestion
	lastMigrationStart = time.Now().Unix()
```

pkg/rancher/migrations.go
```diff
@@ -39,15 +45,33 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
 		}
 	}
 
-	// we only want to migrate if the previous installation was earlier than the migration version
+	// parse out the start and end time
+	lastMigrationStart, _ := strconv.ParseInt(cm.Data[migrationStartedKey], 10, 64)
```
dramich, line 49: Use the helpers for all the conversions https://pkg.go.dev/strconv#hdr-Numeric_Conversions
`Atoi`
`Itoa`
It keeps it consistent and it's much faster than using sprintf which relies on reflection.

pkg/rancher/migrations.go
paynejacob: I am concerned this will finish in less than a second.

dramich: I highly doubt that, especially if you are calling out to update the config. However, just update your time check to include `=`

chart/templates/secret.yaml
```diff
@@ -6,8 +6,8 @@ metadata:
 type: Opaque
 data:
   {{- $existingSecret := lookup "v1" "Secret" .Release.Namespace "bootstrap-secret" }}
-  {{- if $existingSecret }}
-  bootstrapPassword: {{ $existingSecret.data.bootstrapPassword }}
-  {{- else }}
-  bootstrapPassword: {{ .Values.bootstrapPassword | b64enc | quote }}
-  {{- end }}
\ No newline at end of file
+    {{- if and $existingSecret $existingSecret.data }}
```
StrongMonkey, line 9: Would recommend to put this into a separate PR. If anything go wrong reverting might be easier.

dramich: I agree, separate pr as it's not part of this fix or issue.

paynejacob: https://github.com/rancher/rancher/pull/34249

paynejacob:
Closing this pr as the solution has changed dramatically and this conversation is no longer applicable.

Issue #34172, opened 2021-08-11
500 error in /v3/kontainerdrivers/rancher-images in air-gapped setups only with useBundledSystemChart=true


aiyengar2:
**Rancher Server Setup**
- Rancher version: master-head `9d2e08674b9318e440fb6396c6e168a9c9bdca91`

**Describe the bug**

In an air-gapped setup that uses `useBundledSystemChart` (which means the `SystemCatalog` setting is set to `bundled`), the catalog image-list does not exist, resulting in a 500 error with the following logs on trying to hit `/v3/kontainerdrivers/rancher-images`:

```
[ERROR] API error response 500 for GET /v3/kontainerdrivers/rancher-images. Cause: configmaps "cattle-system/system-library-catalog-image-list" not found
```

**To Reproduce**
 

1. Do a Helm chart install of rancher setting useBundledSystemChart to true. Check if the setting for SystemCatalog was set to bundled
2. Try navigating to RANCHER_URL/v3/kontainerdrivers/rancher-images.

**Result**

500 error

**Expected Result**

System charts should come up

**Screenshots**

**Additional context**

This is not the same as https://github.com/rancher/rancher/issues/31336.

The above issue switched Rancher to storing the list of images from rancher/system-charts in a ConfigMap instead of just storing it in management-state/ to resolve an issue where a non-leader Rancher pod did not have access to the repositories to grab the list of images from, as noted in https://github.com/rancher/rancher/issues/31336#issuecomment-785256694.

This is an issue that affects air-gapped clusters regardless of whether or not a ConfigMap is used. However, as a result, backporting this fix will either involve backporting the fix from https://github.com/rancher/rancher/issues/31336 as well or designing a slightly modified solution for Rancher 2.5 and before.

Also, the proposed solution for this issue would be to check if the SystemCatalog is bundled on startup and, if so, use the system-charts [copied into /var/lib/rancher-data/local-catalogs/system-library as part of building the Docker image](https://github.com/rancher/rancher/blob/522c7764153e4cfee6161895ae44efaa59d611d9/package/Dockerfile#L53) in order to generate the ConfigMap.

anupama2501:
reproduced on v2.6.0-rc2

**helm template:**
```
[...]

PR #34225, opened 2021-08-13
Support pulling system imagelist cache from bundled charts

aiyengar2:
Currently, **for system-charts only**, we use the same scripts we use to generate the rancher-images.txt in order to produce a ConfigMap called `system-library-catalog-image-list` in the `cattle-system` namespace.

This ConfigMap contains a list of images that were extracted from system-charts stored in `management-state/catalog-cache/{HASH-OF-CATALOG}`.

However, for air-gapped setups, the system-charts do not get pulled into this directory. Instead, they only exist as packaged in the Rancher [Dockerfile](https://github.com/rancher/rancher/blob/522c7764153e4cfee6161895ae44efaa59d611d9/package/Dockerfile#L53) under `/var/lib/rancher-data/local-catalogs/system-library`. So the creation of the ConfigMap fails, triggering the 500 error seen on the ticket.

The fix pushed out in this PR modifies the functions that are used to create this ConfigMap to strip it into an exported helper function. This helper function is then added into the `doAddCatalogs` step of a Rancher startup (which is triggered every time the MCM controller starts).

Related Issue: https://github.com/rancher/rancher/issues/34172

### DIFF
--- a/pkg/catalog/manager/catalog_sync.go
+++ b/pkg/catalog/manager/catalog_sync.go
@@ -64,51 +64,68 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
 		return nil, err
 	}
 
-	// ensure the system catalog image cache exists
-	var systemCatalogImageCache *v1.ConfigMap
 	if catalog.Name == utils.SystemLibraryName {
-		var systemCatalogImageCacheCreated bool
-		systemCatalogImageCacheName := utils.GetCatalogImageCacheName(catalog.Name)
-		systemCatalogImageCache, err = m.ConfigMapLister.Get(namespace.System, systemCatalogImageCacheName)
-
-		// if the cache does not exist generate it
-		if err != nil && errors.IsNotFound(err) {
-			logrus.Debug("system catalog image cache configmap not found")
-
-			systemCatalogImageCache = &v1.ConfigMap{}
-			systemCatalogImageCache.Name = systemCatalogImageCacheName
-			systemCatalogImageCache.Namespace = namespace.System
-			err = image.CreateCatalogImageListConfigMap(systemCatalogImageCache, catalog)
-			if err != nil {
-				return nil, err
-			}
-
-			systemCatalogImageCache, err = m.ConfigMap.Create(systemCatalogImageCache)
-			if err != nil && !errors.IsAlreadyExists(err) {
-				return nil, err
-			}
-			systemCatalogImageCacheCreated = true
-			logrus.Debug("system catalog image cache configmap created")
-
-		} else {
-			return nil, err
-		}
-
-		// if the cache exists and is out of date update it
-		if !isUpToDate(commit, catalog) || systemCatalogImageCacheCreated {
-			err = image.CreateCatalogImageListConfigMap(systemCatalogImageCache, catalog)
-			if err != nil {
-				return nil, err
-			}
-
-			// update
-			_, err = m.ConfigMap.Update(systemCatalogImageCache)
-			if err != nil {
-				return nil, err
-			}
-			logrus.Debug("system catalog image cache updated")
-		}
+		// ensure the system catalog image cache exists
+		forceUpdate := !isUpToDate(commit, catalog)
+		return nil, CreateOrUpdateSystemCatalogImageCache(catalog, m.ConfigMap, m.ConfigMapLister, false, forceUpdate)
 	}
 
 	return nil, nil
+}
+
+func CreateOrUpdateSystemCatalogImageCache(systemCatalog *v3.Catalog, configMapInterface v1.ConfigMapInterface, configMapLister v1.ConfigMapLister, bundledMode bool, forceUpdate bool) (err error) {
+	var catalogChartPath string
+	catalogChartPath, err = utils.GetCatalogChartPath(systemCatalog, bundledMode)
+	if err != nil {
+		return err
+	}
+
+	var systemCatalogImageCache *v1.ConfigMap
+	var systemCatalogImageCacheCreated bool
+
+	systemCatalogImageCacheName := utils.GetCatalogImageCacheName(systemCatalog.Name)
+	systemCatalogImageCache, err = configMapLister.Get(namespace.System, systemCatalogImageCacheName)
+
+	// if the cache does not exist generate it
+	if err != nil && errors.IsNotFound(err) {
+		logrus.Debug("system catalog image cache configmap not found")
+
+		systemCatalogImageCache = &v1.ConfigMap{}
+		systemCatalogImageCache.Name = systemCatalogImageCacheName
+		systemCatalogImageCache.Namespace = namespace.System
+
+		err = image.AddImagesToImageListConfigMap(systemCatalogImageCache, catalogChartPath)
+		if err != nil {
+			return
+		}
+
+		systemCatalogImageCache, err = configMapInterface.Create(systemCatalogImageCache)
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return
+		}
+		systemCatalogImageCacheCreated = true
+		logrus.Debug("system catalog image cache configmap created")
+
+	} else if err != nil {
+		return
+	}
+
+	// if the cache exists and is out of date update it
+	if forceUpdate || systemCatalogImageCacheCreated {
+		if err != nil {
+			return err
+		}
+		err = image.AddImagesToImageListConfigMap(systemCatalogImageCache, catalogChartPath)
+		if err != nil {
+			return
+		}
+
+		_, err = configMapInterface.Update(systemCatalogImageCache)
+		if err != nil {
+			return
+		}
+		logrus.Debug("system catalog image cache updated")
+	}
+
+	return
 }
--- a/pkg/catalog/utils/utils.go
+++ b/pkg/catalog/utils/utils.go
@@ -2,9 +2,13 @@ package utils
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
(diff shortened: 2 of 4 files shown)


pkg/catalog/manager/catalog_sync.go
```diff
@@ -64,51 +64,69 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
-			systemCatalogImageCache, err = m.ConfigMap.Create(systemCatalogImageCache)
-			if err != nil && !errors.IsAlreadyExists(err) {
-				return nil, err
-			}
-			systemCatalogImageCacheCreated = true
-			logrus.Debug("system catalog image cache configmap created")
+	return nil, nil
+}
+
+func CreateOrUpdateSystemCatalogImageCache(systemCatalog *v3.Catalog, configMapInterface v1.ConfigMapInterface, configMapLister v1.ConfigMapLister, bundledMode bool, forceUpdate bool) (err error) {
```
aiyengar2, line 76: The original code utilized both a Lister (for Gets) and a Interface (for Create / Update) instead of just doing all calls via the Interface.

I debated making all calls go through the Interface instead to keep the function signature simple but chose to stick with this approach since I did not want to have to reach out to the k8s API in order to check if the ConfigMap exists; instead, looking through Rancher's cache of ConfigMaps via the Lister is sufficient and more efficient IMO.

Feel free to comment if you think I should drop the Lister in favor of just doing calls via the Interface!

pkg/data/management/catalog_data.go
```diff
@@ -193,10 +198,17 @@ func doAddCatalogs(management *config.ManagementContext, name, url, branch, helm
...
+		if obj, err = catalogClient.Create(obj); err != nil {
 			return err
 		}
 	}
 
+	if bundledMode && obj.Name == utils.SystemLibraryName {
+		// force update the catalog cache on every startup
+		configMapInterface := management.Core.ConfigMaps("")
+		configMapLister := configMapInterface.Controller().Lister()
+		return manager.CreateOrUpdateSystemCatalogImageCache(obj, configMapInterface, configMapLister, true, true)
```
aiyengar2, line 210: For this invocation of `CreateOrUpdateSystemCatalogImageCache`, I am specifying that we always do a forceUpdate since, if a user were to change the contents of `/var/lib/rancher-data/local-catalogs/system-library`, they should be able to update the contents of the `/v3/kontainerdriver/rancher-images` simply by restarting Rancher. 

Without this, there would be no way of supporting upgrading the system-charts and seeing the new images pop up in air-gapped setups.

pkg/image/resolve.go
```diff
@@ -349,19 +349,16 @@ func getSourcesList(imageSources map[string]bool) string {
 	return strings.Join(sources, ",")
 }
 
-func CreateCatalogImageListConfigMap(cm *v1.ConfigMap, catalog *v3.Catalog) (err error) {
```
aiyengar2, line 352: This function is no longer used by Rancher, but it is a public facing API. Should I leave this function behind in case someone is using Rancher as a library? Not sure what our policy is with deprecating such exported functions.

pkg/catalog/utils/utils.go
```diff
@@ -79,3 +83,17 @@ func Contains(collection []string, key string) bool {
...
+func GetCatalogChartPath(catalog *v3.Catalog, bundledMode bool) (string, error) {
+	if bundledMode {
+		switch catalog.Name {
+		case "helm3-library", "library", "system-library":
+			return filepath.Join(helmlib.InternalCatalog, catalog.Name), nil
+		case "rancher-charts", "rancher-partner-charts", "rancher-rke2-charts":
+			return filepath.Join(helmlib.InternalCatalog, "v2", catalog.Name), nil
+		default:
+			return "", fmt.Errorf("cannot find bundled catalog chart path for catalog %s", catalog.Name)
+		}
```
aiyengar2, line 96: I didn't need to add all these switch cases, but I did for good measure in case someone wants to use this helper function in the future 😄

pkg/image/resolve.go
PennyScissors: If we aren't using it, I would prefer to remove it since we already export GetImages() and this package can use some cleanup.

aiyengar2: cc: @kinarashah what do you think?

kinarashah: `CreateCatalogImageListConfigMap` was introduced very recently by https://github.com/rancher/rancher/pull/33151 for 2.6. I don't think this has been released yet, was this elsewhere?

aiyengar2: Ah yes I think you are right; we should be safe to remove it then. Thanks @kinarashah @PennyScissors!

aiyengar2:
Need to create backport PR for release/v2.6 once complete, but will wait on approvals before setting up PR.

pkg/catalog/manager/catalog_sync.go
```diff
@@ -64,51 +64,69 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
+	if forceUpdate || systemCatalogImageCacheCreated {
+		if err != nil {
+			return err
+		}
+		err = image.AddImagesToImageListConfigMap(systemCatalogImageCache, catalogChartPath)
+		if err != nil {
+			return
 		}
+
+		// update
```
PennyScissors, line 123: nit: remove this comment, is unnecessary

PennyScissors (dismissed):
lgtm, just one nit

pkg/catalog/manager/catalog_sync.go
aiyengar2: addressed

aiyengar2:
addressed @PennyScissors's nit

pkg/catalog/manager/catalog_sync.go
```diff
@@ -64,51 +64,68 @@ func (m *Manager) Sync(key string, obj *v3.Catalog) (runtime.Object, error) {
...
-			if err != nil {
-				return nil, err
-			}
-			logrus.Debug("system catalog image cache updated")
+	} else if err != nil {
+		return
+	}
+
+	// if the cache exists and is out of date update it
+	if forceUpdate || systemCatalogImageCacheCreated {
```
kinarashah, line 114: what's the use case for updating if systemCatalogImageCacheCreated=true, only forceUpdate=true should suffice?

pkg/data/management/catalog_data.go
kinarashah: This context should also be added as a comment on L207 to explain why we're force updating the catalog cache.

kinarashah: PS: how does `system-library` content affect `kontainerdriver` images? we only cache images from chart in `AddImagesToImageListConfigMap`?

pkg/catalog/manager/catalog_sync.go
kinarashah: Using Lister for Gets and client for Create+Updates sounds good. If the logic is expected to always return object if it exists, we use Interface to Get if Lister returns NotFound but in this case can just use client to Create and no-op if AlreadyExists as you did. 👍🏼

pkg/catalog/manager/catalog_sync.go
aiyengar2: This was copied from the original function's logic in https://github.com/rancher/rancher/blob/fa481ddb14e3a3413ca5d75e709474bfb02a7eab/pkg/catalog/manager/catalog_sync.go#L98

I believe the edge case it catches is if the ConfigMap is created by another Rancher instance during a `Sync` on the catalog manager.

If a ConfigMap were to be created after this line:
https://github.com/rancher/rancher/blob/fa481ddb14e3a3413ca5d75e709474bfb02a7eab/pkg/catalog/manager/catalog_sync.go#L72

Then our Create call in the following line can return an `IsAlreadyExist`:
https://github.com/rancher/rancher/blob/fa481ddb14e3a3413ca5d75e709474bfb02a7eab/pkg/catalog/manager/catalog_sync.go#L86-L87

In this case, if we are doing a forceUpdate, then we should still upgrade the ConfigMap with the latest.

kinarashah: Yeah, this might help resolve this, I think the create through any sync would have already fetched the latest images but makes sense to leave it as is from old logic than remove it.

pkg/data/management/catalog_data.go
aiyengar2: Sure, I will add it as a comment.

aiyengar2: > PS: how does system-library content affect kontainerdriver images? we only cache images from chart in AddImagesToImageListConfigMap?

`kontainerdriver` images are composed of images from 3 sources, as shown in https://github.com/rancher/rancher/blob/fa481ddb14e3a3413ca5d75e709474bfb02a7eab/pkg/api/norman/customization/kontainerdriver/actionhandler.go#L183-L185

1. The agent image
2. RKESystemImages (from KDM)
3. The system-chart images

Prior to 2.6, the contents of the git repo for system-charts in the `management-state` directory on each Rancher container was the source of truth for the latest copy of system-charts. We would `GetImages` on demand when such a request came in.

However, to fix #31336, we made the new source of truth for system-chart images the `system-library` ConfigMap instead, which already contains the output of the `GetImages` command on the system-charts.

Therefore, `AddImagesToImageListConfigMap` gets called in one of two ways:
1. On startup (after my fix) to create the initial ConfigMap and update it based on the bundled contents
2. On update (only in non-bundled Rancher clusters), as a result of the catalog `manager.Sync` if it is detected that the catalog is not up-to-date.

kinarashah (approved):

PennyScissors (approved):

Issue #34187, opened 2021-08-12
"unable to create impersonator account: error getting service account token: serviceaccounts \"cattle-impersonation-system/cattle-impersonation-user-dxx4g\" not found - Error seen for Monitoring and Istio v1 charts and when using downloaded kubeconfig


slickwarren:
**Rancher Server Setup**
- Rancher version: 2.5.9 -> 2.6.0-rc2 upgrade
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): helm install

**Information about the Cluster**
- Kubernetes version: 
- Cluster Type (Local/Downstream): downsteram
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): 

**Describe the bug**

when a user is logged in as admin, all monitoring v1 (v0.2.2) links are not working. 

**To Reproduce**
 
* deploy rancher server v2.5.9
* on a downstream cluster, install monitoring (latest version, v0.2.2)
  * grant a standard user permissions to this cluster i.e. `cluster owner` before the upgrade (see additional context)
  * verify that prometheus/grafana links are working
* upgrade rancher to v2.6.0-rc2

**Result**
`"status":500,"statusText":"","data":"unable to create impersonator account: error getting service account token: serviceaccounts \"cattle-impersonation-system/cattle-impersonation-user-dxx4g\" not found` from both prometheus and grafana. 

**Expected Result**

monitoring v1 should continue to work after an upgrade

**Screenshots**

**Additional context**

* as a standard user that was created + added to the cluster before the upgrade, I am able to access monitoring v1 apps (prom, grafana) after the upgrade.

anupama2501:
I see the same issue while accessing istio `jaeger traces` and `kiali` when logged in as admin.

- Created a downstream cluster on rancher v2.5.9 as a standard user
- installed istio v1.5.901 from cluster --> tools
- jaeger traces and kiali  work fine
- Upgrade rancher server to v2.6.0-rc2
- Access the kiali and jaeger traces as standard user works fine.
- Access the kiali and jaeger traces as an admin **errors** out:
```
unable to create impersonator account: error getting service account token: serviceaccounts "cattle-impersonation-system/cattle-impersonation-user-dxx4g" not found
```

cmurphy:
[...]

Issue #34042, opened 2021-08-07
Not able to deploy clusters when external auth provider is enabled


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `0ab00d`
- Installation option (Docker install/Helm Chart): helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: rancher generated

**Information about the Cluster**
- Kubernetes version: 1.20.9
- Cluster Type (Local/Downstream): Downstream - Windows custom RKE1, k3s Imported

**Describe the bug**
Not able to deploy clusters as a standard user when external auth provider is enabled

**To Reproduce:**
- On 2.6-head, enable GH Auth or AD Auth
- Deploy clusters using standard user token/as an admin
- Windows custom RKE1 and Imported k3s clusters dont come up. Error seen - `Error while applying agent YAML, it will be retried automatically: exit status 1, Error from server (InternalError): an error on the server ("unable to create impersonator account: error getting service account token: serviceaccounts \"cattle-impersonation-system/cattle-impersonation-u-kl3nli645y\" not found") has prevented the request from succeeding`
- Imported RKE2 works fine though
<img width="1273" alt="Screen Shot 2021-08-07 at 11 13 45 AM" src="https://user-images.githubusercontent.com/26032343/128610233-e170dccf-b5e6-465a-afe8-296d9c929144.png">

**Note:**
On 2.6.0-rc1 it works

sowmyav27:
**On 2.6 - head commit id: `4b6af95d` and on master-head commit id: `524aec1`**

- On 2.6-head, enable AD Auth HA setup
- Deploy clusters using standard user token/as an admin - RKE1 custom clusters
- No issues seen, Clusters come up Active

sowmyav27:
**On an upgrade from 2.5.9 to 2.6.0-rc2 , this issue is seen,**

- On 2.5.9 deploy a few downstream clusters
- Upgrade rancher to 2.6.0-rc2
- One of the downstream clusters (Custom RKE1 cluster) is stuck in `Provisioning` state with error `Error while applying agent YAML, it will be retried automatically: exit status 1, Error from server (InternalError): an error on the server ("unable to create impersonator account: error getting service account token: serviceaccounts \"cattle-impersonation-system/cattle-impersonation-u-bdasg4j7sc\" not found") has prevented the request from succeeding`

[...]

PR #34234, opened 2021-08-13
Change impersonation service account retrieval

cmurphy:
This change does three things:
1) Partially revert the refactor done in 8d413413 so that the service
   account only needs to be retrieved once in total if it exists,
   instead of as two steps in SetUpImpersonation and GetToken
2) Fix waitForServiceAccount to use the cache instead of the client
3) Add trace logs to help debug future caching problems

https://github.com/rancher/rancher/issues/34042

### DIFF
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -11,6 +11,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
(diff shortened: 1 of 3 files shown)


pkg/impersonation/impersonation.go
```diff
@@ -30,46 +31,55 @@ func New(user user.Info, clusterContext *config.UserContext) Impersonator {
...
-		return "", err
-	}
+func (i *Impersonator) GetToken(sa *corev1.ServiceAccount) (string, error) {
 	if len(sa.Secrets) == 0 {
 		return "", fmt.Errorf("service account is not ready")
 	}
 	secret := sa.Secrets[0]
 	secretObj, err := i.clusterContext.Core.Secrets("").Controller().Lister().Get(impersonationNamespace, secret.Name)
 	if err != nil {
+		logrus.Tracef("impersonation: error getting service account token %s: %v", secret.Name, err)
```
dramich, line 75: This can be wrapped in a `if logrus.GetLevel() >= logrus.TraceLevel {` so we don't do the list unnecessarily, same for the below sets of trace things.

dramich (approved):

Issue #34235, opened 2021-08-13
Deleting a machine could show an inaccurate status in the UI


thedadams:
**Describe the bug**
When a machine is being deleted, occasionally the machine will have a status with message "waiting for machine to be deleted" in the UI and in Kubernetes. This status is incorrect for a machine that is being deleted.

**To Reproduce**
1. Create a cluster with multiple worker only nodes.
2. After the cluster is active, use the minus-sign in the Machine Pools UI to scale down worker-only nodes.

**Result**
The UI will occasionally show a status of "Waiting for machine to be defined"

**Expected Result**
This status is inaccurate for a machine that is being deleted.

aaronRancher:
This issue is reproduced on Rancher `v2.6.0-rc3` with the following steps:
- Create an RKEv2 cluster with multiple worker only nodes.
- After the cluster is active, use the minus-sign in the Machine Pools UI to scale down worker-only nodes.

Result:
- Node status is `provisioning` UI status is `Waiting for machine to be defined`
![image](https://user-images.githubusercontent.com/57409052/129817775-a840bb4c-4a58-4899-9b90-b096830b0c66.png)

Tested on Rancher `v2.6-head` `version 72986b2`.

- When scaling down the worker nodepool, the node goes into deleting state:
![image](https://user-images.githubusercontent.com/57409052/129817623-a1561520-39f7-4192-90cd-b3c641f15e63.png)
- Then, node goes into `provisioning` state. UI status is `Machine is being deleted`
![image](https://user-images.githubusercontent.com/57409052/129817689-2a9acd0d-b831-4f83-8545-643196236da6.png)

Reopening, as we expect the node to be in `deleting` state when nodepool is scaled down.

snasovich:
Should be addressed with the same fix as https://github.com/rancher/rancher/issues/34264 which is currently being worked on.

thedadams:
Now, instead of the status changing to "Provisioning" (as above), the status should go to Deleted with a message as shown here:
![Screen Shot 2021-08-20 at 4 47 39 PM](https://user-images.githubusercontent.com/3682717/130303288-b5bec7e2-c9c5-4704-b597-b1bee4728ca7.png)

sowmyav27:
**On 2.6-head commit id: `8898bf8`**

- Scale down the node on the UI
- The Node remains in Active state, and get removed from the UI from the cluster.
- There is no status change of the node
- Reopening the issue

PR #34237, opened 2021-08-13
Add new status for machines that are being deleted

thedadams:
When a machine is being deleted, sometime the UI will show a status of
"waiting for machine to be defined." This is inaccurate for a machine
that is being deleted.

This change will display an accurate message in this case.

Issue:
https://github.com/rancher/rancher/issues/34235


cmurphy (approved):

jiaqiluo (approved):
LGTM

dramich (approved):

PR #34239, opened 2021-08-13
[Forwardport master] Create rancher service for internal communication to prevent loopback through load balancer

Oats87:
There are certain cases (when running Rancher in a docker container) where we do not want Rancher to attempt to connect via the exposing load balancer. This PR will create a "rancher" service in cattle-system and use that if running an embedded K3s.

https://github.com/rancher/rancher/issues/34015

Forwardport of https://github.com/rancher/rancher/pull/34211


kinarashah (approved):

kinarashah:
Merging this since it's forward port.

Issue #34227, opened 2021-08-13
Helm fails to upgrade Rancher if bootstrapPassword is not set


jiaqiluo:
**Rancher Server Setup**
- Rancher version: 2.6.0-rc2 
- Installation option (Docker install/Helm Chart): helm 
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): rke1 

Steps:
- use Helm to install rancher 2.6.0-rc2
- wait for Rancher to be active 
- use helm to upgrade Rancher without passing any new values 

Results:
Helm upgrade fails with the following error 
(in my case, I downloaded the chart locally)
```
> 	helm upgrade -n cattle-system  rancher ./rancher --debug
upgrade.go:123: [debug] preparing upgrade for rancher
upgrade.go:459: [debug] copying values from rancher (v1) to new release.
Error: UPGRADE FAILED: error validating "": error validating data: unknown object type "nil" in Secret.data.bootstrapPassword
helm.go:88: [debug] error validating "": error validating data: unknown object type "nil" in Secret.data.bootstrapPassword
helm.sh/helm/v3/pkg/kube.scrubValidationError
	helm.sh/helm/v3/pkg/kube/client.go:622
helm.sh/helm/v3/pkg/kube.(*Client).Build
	helm.sh/helm/v3/pkg/kube/client.go:195
helm.sh/helm/v3/pkg/action.validateManifest
	helm.sh/helm/v3/pkg/action/upgrade.go:466
helm.sh/helm/v3/pkg/action.(*Upgrade).prepareUpgrade
	helm.sh/helm/v3/pkg/action/upgrade.go:240
helm.sh/helm/v3/pkg/action.(*Upgrade).Run
	helm.sh/helm/v3/pkg/action/upgrade.go:124
main.newUpgradeCmd.func2
	helm.sh/helm/v3/cmd/helm/upgrade.go:155
github.com/spf13/cobra.(*Command).execute
	github.com/spf13/cobra@v1.1.3/command.go:852
github.com/spf13/cobra.(*Command).ExecuteC
	github.com/spf13/cobra@v1.1.3/command.go:960
github.com/spf13/cobra.(*Command).Execute
	github.com/spf13/cobra@v1.1.3/command.go:897
main.main
	helm.sh/helm/v3/cmd/helm/helm.go:87
runtime.main
	runtime/proc.go:225
runtime.goexit
	runtime/asm_amd64.s:1371
[...]

PR #34249, opened 2021-08-13
fix error when upgrading helm chart

paynejacob:
https://github.com/rancher/rancher/issues/34227

Fixed helm failing to upgrade when the default password has not yet been initalized.


StrongMonkey (approved):

jiaqiluo (approved):
LGTM

PR #34251, opened 2021-08-13
fixed version checks for force logout

paynejacob:

Fixed how we decide to force a logout. All users will be logged out in ONLY the following cases:
- if the current rancher version is an rc or release and the target version is greater than the last force logout
- if the current rancher version is a `*-head` build and the last force logout version does not match the current version

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -134,6 +134,7 @@ require (
 	github.com/yvasiyarov/gorelic v0.0.7 // indirect
 	github.com/yvasiyarov/newrelic_platform_go v0.0.0-20160601141957-9c099fbc30e9 // indirect
 	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97
+	golang.org/x/mod v0.4.0
 	golang.org/x/net v0.0.0-20210315170653-34ac3e1c2000
 	golang.org/x/oauth2 v0.0.0-20201208152858-08078c50e5b5
 	golang.org/x/sync v0.0.0-20201207232520-09787c993a3a
--- a/pkg/rancher/migrations.go
+++ b/pkg/rancher/migrations.go
@@ -7,6 +7,7 @@ import (
 	rancherversion "github.com/rancher/rancher/pkg/version"
 	controllerv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/mod/semver"
 	v1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


pkg/rancher/migrations.go
```diff
@@ -39,9 +40,21 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
 		}
 	}
 
-	// we only want to migrate if the previous installation was earlier than the migration version
+	// we do not migrate in development environments
+	if rancherversion.Version == "dev" {
+		return nil
+	}
+
+	// if no last migration is found we run
```
snasovich, line 48:
```suggestion
	// if no last migration is found we always run force logout
```

pkg/rancher/migrations.go
```diff
@@ -39,9 +40,21 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
...
 	if lastMigration, ok := cm.Data[rancherVersionKey]; ok {
-		if lastMigration == "dev" || version.Compare(migrationVersion, lastMigration, "<=") {
+
+		// if a valid sem ver is found we only migrate if the version is less than the target version
+		if semver.IsValid(lastMigration) && semver.IsValid(rancherversion.Version) && version.Compare(migrationVersion, lastMigration, "<") {
```
snasovich, line 52: Could still use `<=` here. If versions are the same, we don't migrate. No need to leave it to the next step.

snasovich (approved):
Couple nits, otherwise LGTM

snasovich (approved):
LGTM, still need committer review

kinarashah (commented):
checks based on rancher version look good. I have a question on the logout flow itself tho, we seem to be deleting tokens first followed by storing the rancher version in the configmap. If the update errors out, we'd delete the tokens again, is that expected?

paynejacob:
@kinarashah yes that would be the case.  We had a previous pr to add logic to ensure that didn't happen but it was determined the logic was too complex for such an uncommon situation.  if rancher crashes during this part, it is crashing during the first startup after an upgrade and is likely in an unusable state anyway.

kinarashah:
I was thinking along the lines of error like an update conflict, but in that case we'd have seen the logs twice. I don't see that in the linked issue so looks good for now.

kinarashah (approved):

go.mod
```diff
@@ -134,6 +134,7 @@ require (
 	github.com/yvasiyarov/gorelic v0.0.7 // indirect
 	github.com/yvasiyarov/newrelic_platform_go v0.0.0-20160601141957-9c099fbc30e9 // indirect
 	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97
+	golang.org/x/mod v0.4.0
```
dramich, line 137: This is pulling in a 4th semver package... can an existing be used?

dramich: Or we open an issue to get them consolidated

paynejacob: I would argue we should consolidate, looking at the other ones in here they are all the same.  I can reuse one but I will have to write a wrapper func to validate them.

dramich: Going with the official package seems like the best option anyways.

dramich (approved):

Issue #33800, opened 2021-07-28
Should not delete the cattle-system namespace when deleting the imported k8s cluster when Rancher pod is running in it


guangbochen:
**Rancher Server Setup**
- Rancher version: master-head

**Describe the bug**
When deleting the imported k8s cluster from the Rancher it will delete all the resources and the namespace of the `cattle-system`, but this will cause an issue of the imported k8s cluster if it has extra resources that exist in the same `cattle-system` namespace, e.g, in the Harvester cluster we run embedded Rancher in this namespace which should not get removed by the Rancher.

**To Reproduce**
1. import an existing k8s cluster to the Rancher server.
2. delete the imported k8s cluster
3. the `cattle-system` namespace  and its resources is removed completely 

**Result**
- resources in the same namespace will also get deleted.

**Expected Result**
should keep the other resource and the `cattle-system` namespace, probably only delete the cluster-agent node.

deniseschannon:
QA should test removing an importing cluster and re-importing to ensure a cluster could still be imported.

guangbochen:
verified using the `master-head-08-19` version:
1. imported a downstream Harvester/k8s cluster via the `Importing Existing` option
2. the cluster explorer is ready
3. delete the imported cluster and all resources have remained (the downstream cluster must have embedded rancher installed on the `cattle-system` namespace)
4. reimporting the same cluster again and the cluster explorer still working as expected.

![image](https://user-images.githubusercontent.com/4569037/130049733-db0a1786-0816-4ab1-839f-e2d7068bd078.png)

deniseschannon:
@kravciak Can you do one additional validation? I believe the change we made only keeps the cattle system namespace IF Rancher is running on the imported cluster.

Can you validate that removing a regular imported cluster cleans up the `cattle-system` namespace?

deniseschannon:
@dnoland1 Do you think this change will affect the hosted Rancher use case at all?

dnoland1:
@deniseschannon I don't think it will, but can test. Does "imported k8s cluster" include k3s clusters that have been imported?

deniseschannon:
Yes, I would assume so @dnoland1

kravciak:
@deniseschannon pods are not cleaned from cattle-system namespace after cluster removal

I did:
- run v2.6-head (b3980c2) docker single node rancher
- import k3s cluster `Import Existing -> Generic`
- wait until all pods on k3s are running after import
- remove k3s cluster from dashboard UI

[...]

PR #34294, opened 2021-08-18
Don't cleanup an imported cluster on remove if Rancher is installed

ibuildthecloud:


ibuildthecloud:
@dramich Do you see any issue with this? Basically if Rancher is installed downstream we don't want to wipe out `cattle-system` and delete the downstream installation.

dramich:
My concern would be if we do this that cluster can not be reimported to rancher cleanly. It will have the same issue it did before this where we reference things that don't exist in rancher management.

Issue #34185, opened 2021-08-11
Restricted Admin needs access to new cluster CRDs


deniseschannon:
**Rancher Server Setup**
- Rancher version: v2.6-head `6cb1847`
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details:

**Describe the bug**

Restricted admins can see downstream clusters on the home page, but unable to see any clusters on the cluster management page. 

Based on convo with @vincent99 

**To Reproduce**
 
1. Login as standard user
2. Create a cluster
3. Login as restricted admin.
4. Able to see downstream cluster on home page
5. Go to Cluster Management page

**Result**

Unable to see downstream cluster

**Expected Result**

Able to see all downstream clusters (even if I'm not assigned any membership)

**Screenshots**

**Additional context**

prachidamle:
I moved it to to-test, but there are some issues causing the image not published out.

davidnuzik:
**My checks PASSED**
 
**Reproduction Steps:**

Not required. I understand the issue well.
 
**Validation Environment:**

Rancher version: v2.6-head `70f024e` pulled at 8/24/21 6:13 pm Pacific
Rancher cluster type: single-node docker
Docker version: 20.10
 
Downstream cluster type: rke2
Downstream K8s version: v1.21.3+rke2r2
 
**Validation steps:**

1. Create a standard user and a restricted admin user.
2. Log in as the standard user. Create a cluster.
3. Now, log in as the restricted admin user. Ensure you can see that cluster.

![image](https://user-images.githubusercontent.com/45179589/130711008-787b23af-1073-454c-b472-4d4f71c8818f.png)


Note: I also checked with a cluster created by admin. The restricted admin can indeed see all clusters even those created by the standard user and admin user. Additionally, the standard user can only see the cluster and cloud credentials it created, as expected. The restricted admin can only see it's own cloud credentials, as expected.

PR #34301, opened 2021-08-18
Adding rbac for restrictedAdmin to access Provisioning Clusters

prachidamle:
https://github.com/rancher/rancher/issues/34185

Adding rbac permissions for restrictedAdmin users to be able to access the provisioning Cluster. This has been added to two places:
1) In the cluster sync, to create a rolebinding to the cluster-admin role created for the provisioning cluster for all restrictedAdmin subjects
2) In the GRB sync, when a new restrictedAdmin user is added, we create a rolebinding to the  cluster-admin role created for the provisioning cluster for this user

### DIFF
--- a/pkg/controllers/management/authprovisioningv2/cluster.go
+++ b/pkg/controllers/management/authprovisioningv2/cluster.go
@@ -1,21 +1,41 @@
 package authprovisioningv2
 
 import (
+	"fmt"
 	"reflect"
 
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/rbac"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// OnCluster creates the role required for users assigned through PRTBs to be able to see the
+// OnCluster creates the roles required for users to be able to see/manage the
 // provisioning cluster resource
 func (h *handler) OnCluster(key string, cluster *v1.Cluster) (*v1.Cluster, error) {
-	if cluster == nil || cluster.DeletionTimestamp != nil {
+	if cluster == nil {
 		return cluster, nil
 	}
 
+	if cluster.DeletionTimestamp != nil {
+		return cluster, h.cleanClusterAdminRoleBindings(cluster)
+	}
+
+	err := h.createClusterViewRole(cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	err = h.createClusterAdminRole(cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	return cluster, nil
+}
+
+func (h *handler) createClusterViewRole(cluster *v1.Cluster) error {
 	roleName := clusterViewName(cluster)
 	role := &rbacv1.Role{
 		TypeMeta: metav1.TypeMeta{},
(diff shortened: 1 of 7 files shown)


dramich (commented):
Per offline discussion, validate that the roleBinding is getting deleted when the GRB gets deleted for the restricted-admin.

prachidamle:
> Per offline discussion, validate that the roleBinding is getting deleted when the GRB gets deleted for the restricted-admin.

Added cleanup for the rolebinding created for the provisioning cluster in the current globalrolebinding lifecycle.

pkg/controllers/management/authprovisioningv2/prtb.go
```diff
@@ -39,7 +39,7 @@ func (h *handler) OnPRTB(key string, prtb *v3.ProjectRoleTemplateBinding) (*v3.P
 		// permissions for the provisioning objects won't be created until an
 		// update to the PRTB happens again.
 		logrus.Debugf("[auth-prov-v2-prtb] No provisioning cluster found for cluster %v, enqueuing PRTB %v ", prtb.ClusterName, prtb.Name)
-		h.clusterRoleTemplateBindingController.EnqueueAfter(prtb.Namespace, prtb.Name, 10*time.Second)
```
StrongMonkey, line 42: Is this a bug or intended?

dramich: It's a bug.

StrongMonkey (approved):

pkg/controllers/management/authprovisioningv2/cluster.go
```diff
@@ -1,21 +1,38 @@
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// OnCluster creates the role required for users assigned through PRTBs to be able to see the
+// OnCluster creates the roles required for users to be able to see/manage the
 // provisioning cluster resource
 func (h *handler) OnCluster(key string, cluster *v1.Cluster) (*v1.Cluster, error) {
 	if cluster == nil || cluster.DeletionTimestamp != nil {
-		return cluster, nil
+		return cluster, h.cleanClusterAdminRoleBindings(cluster)
```
dramich, line 19: This needs to be split into two checks, one for nil then a 2nd for delete timestamp or the cleanClusterAdminRoleBindings will panic with nil access.

pkg/controllers/management/authprovisioningv2/cluster.go
```diff
@@ -44,20 +61,107 @@ func (h *handler) OnCluster(key string, cluster *v1.Cluster) (*v1.Cluster, error
...
+	existingRole, err := h.roleCache.Get(cluster.Namespace, roleName)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
+		}
+	}
+	if existingRole == nil {
+		logrus.Infof("Role with ProvisioningClusterAdminName not found %v ", roleName)
+		return nil
+	}
```
dramich, line 146: I would expect the role to be gone, or at least getting deleted so I don't think this part is necessary. The role being gone does not mean the roleBindings are cleaned up

prachidamle: removed this check

pkg/controllers/management/authprovisioningv2/cluster.go
prachidamle: my bad, changed it

dramich (approved):

PR #34303, opened 2021-08-18
[2.6.1] Foward Port: fixed version checks for force logout

paynejacob:

https://github.com/rancher/rancher/pull/34251


cmurphy (approved):
Can't evaluate the code but it looks like a good forwardport :+1:

Issue #34264, opened 2021-08-16
Rancher will spawn dozens of machines in each pool in a certain situation


davidnuzik:
**Summary:**

If you reproduce https://github.com/rancher/rancher/issues/34088#issuecomment-899714663 and delete that cluster, then create a cluster that DOES specify the bucket name on the cluster creation form, and has the same cluster name, then Rancher will spawn dozens (seemingly infinite) number of machines.

This makes me think that _maybe_ if you had a cluster fail to provision the first time (for any reason?), delete it, then create it again with the same name and it's able to provision this time, then you might also encounter this issue. i.e. you might not have to reproduce rancher/rancher#34088 precisely - just have the cluster fail to provision first time, delete, then create again with same name. I am not sure.

While a fix in https://github.com/rancher/rancher/issues/34088 should resolve this problem, what concerns me is the fact that Rancher seems to "flip out" here and create so many machines. This is a peculiar behavior. I am not certain if this should be a blocker for release - probably not, but it might be worth looking into further post 2.6.0 release...

**Environment:**
Rancher version: v2.6-head `f81199694` 8/16/21 9:08 am Pacific
Rancher cluster type: single-node docker
Docker version: 19.03
 
Downstream cluster type: rke2
Downstream K8s version: v1.21.3-rc6+rke2r2

**Steps to Reproduce:**
1. Reproduce https://github.com/rancher/rancher/issues/34088#issuecomment-899714663 but with multiple pools. See the steps below.

1a. Due to [this bug](https://github.com/rancher/dashboard/issues/3847), create an s3-compatible cloud credential first as this will not work during cluster creation. Supply the required key and secret (fake key & secret values pictured but not used). Supply the optional bucket and region. This should have everything we need for when we create a cluster with s3 etcd snapshot/restore capability.
![image](https://user-images.githubusercontent.com/45179589/129608219-e0b56147-38db-48b6-a92e-cecf3c302f7e.png)

Note, the UI appears to send the correct request body for `cloudcredentials` POST:
![image](https://user-images.githubusercontent.com/45179589/129608743-190892c6-29b2-4822-a34a-f1415dce9bb8.png)

The response as well appears correct. The secret is not mentioned probably for security reasons.
![image](https://user-images.githubusercontent.com/45179589/129608867-5a457192-6741-41af-9652-0bde66139997.png)

[...]

PR #34316, opened 2021-08-18
Put deleted status message on InfrastructureReady condition

thedadams:
A previous change mistakenly set the Provision status when a machine was
being deleted. This change now updates the InfrastructureReady condition
when it has been detected that the infrastructure object has been 
deleted.

This is necessary because the CAPI machine controller is waiting for the
infrastructure object to be deleted and this is required so that the 
CAPI controller runs again.

Issues:
https://github.com/rancher/rancher/issues/34235
https://github.com/rancher/rancher/issues/34264

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
+++ b/pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
@@ -32,7 +32,8 @@ import (
 )
 
 const (
-	Provisioned = condition.Cond("Provisioned")
+	Provisioned         = condition.Cond("Provisioned")
+	InfrastructureReady = condition.Cond(capi.InfrastructureReadyCondition)
 )
 
 type handler struct {
@@ -163,6 +164,15 @@ func (h *handler) OnChange(key string, machine *capi.Machine) (*capi.Machine, er
 		return machine, nil
 	}
 
+	status, reason, message, providerID, err := h.getInfraMachineState(machine)
+	if machine.DeletionTimestamp != nil && apierror.IsNotFound(err) {
+		// If the machine is being deleted and the infrastructure machine object is not found,
+		// then update the status of the machine object so the CAPI controller picks it up.
+		return h.setMachineCondition(machine, InfrastructureReady, status, reason, message)
+	} else if err != nil {
+		return machine, err
+	}
+
 	rkeBootstrap, err := h.bootstrapCache.Get(machine.Namespace, machine.Spec.Bootstrap.ConfigRef.Name)
 	if err != nil {
 		return machine, err
@@ -182,11 +192,6 @@ func (h *handler) OnChange(key string, machine *capi.Machine) (*capi.Machine, er
 	}
 
 	if err := h.setJoinURLFromOutput(machine, plan); err != nil {
-		return machine, err
-	}
-
-	status, reason, message, providerID, err := h.getInfraMachineState(machine)
-	if err != nil {
 		return machine, err
 	}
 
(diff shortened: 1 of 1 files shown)


jiaqiluo (approved):
LGTM

paynejacob (dismissed):

dramich:
Why not keep the `Deleting` and update the massage to indicate the machine part. This feels heavy and not totally accurate since the object does still exist.

dramich (commented):
Per offline discussion, ensure we don't change the status/message/condition of the machine if it is being deleted since it's already in the correct state. Revert the previous fix, and find a location to enqueue the machine again after we have done our work on delete.

thedadams:
@dramich I updated the description to try to indicate why these changes are necessary.

snasovich (dismissed):
Code LGTM, leaving it to @dramich to confirm logic is correct.

paynejacob (approved):

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
```diff
@@ -32,7 +32,8 @@ import (
 )
 
 const (
-	Provisioned = condition.Cond("Provisioned")
+	Provisioned         = condition.Cond("Provisioned")
+	InfrastructureReady = condition.Cond("InfrastructureReady")
```
dramich, line 36: Per offline, attempt to use the upstream CAPI consts here

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
```diff
@@ -266,7 +275,7 @@ func (h *handler) getInfraMachineState(capiMachine *capi.Machine) (status corev1
 	machine, err := h.dynamic.Get(gvk, capiMachine.Namespace, capiMachine.Spec.InfrastructureRef.Name)
 	if apierror.IsNotFound(err) {
 		if capiMachine.DeletionTimestamp != nil {
-			return corev1.ConditionUnknown, "MachineDeleted", "machine is being deleted", "", nil
+			return corev1.ConditionFalse, "Deleted", "machine infrastructure is deleted", "", err
```
dramich, line 278: Per offline, attempt to use the upstream CAPI consts here

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
thedadams: There is a Condition constant for `InfrastructureReady`, but not for `Provisioned`.

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
thedadams: Will use the `MachineDelete` reason constant.

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
```diff
@@ -163,6 +164,15 @@ func (h *handler) OnChange(key string, machine *capi.Machine) (*capi.Machine, er
 		return machine, nil
 	}
 
+	status, reason, message, providerID, err := h.getInfraMachineState(machine)
+	if machine.DeletionTimestamp != nil && apierror.IsNotFound(err) {
+		// If the machine is being deleted and the  infrastructure machine object is not found,
```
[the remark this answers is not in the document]
thedadams: Done.

dramich (approved):

Issue #34277, opened 2021-08-17
Grant permissions to cluster operation when creating a cluster through a template


dramich:
Permissions need to be created when a user creates the catalog entry so they can view the `operation` and see the status of the installation process. 

https://github.com/rancher/dashboard/issues/3721

gaktive:
UI components are merged in per https://github.com/rancher/dashboard/issues/3721 -- shall we transfer that ticket to this repo for the backend work?

StrongMonkey:
We have an insteresting problem to solve now...

> Helm install requires to check if resources exist or not, but in rancher you won't have the permission to check until you created cluster

https://github.com/helm/helm/blob/550f250ff90a9cc8f9d6f176d43181f9940c5a7b/pkg/action/install.go#L276

StrongMonkey:
https://github.com/rancher/rancher/pull/34317: Fix the issue where standard user can't see operation and its logs in management cluster, causing failure to view logs after install rke2 templates.

https://github.com/rancher/helm/pull/31: Fix an issue where helm needs to check resources before creating it. It will fail in rancher RBAC model for provisioning v2 cluster and machine configs.

StrongMonkey:
@anupama2501 To test this, you should be able to 
1. Add cluster template through catalog. https://github.com/rancher/dashboard/issues/3721#issuecomment-898592109
2. Login as standard user.
3. Try to create cluster from cluster template. You should be able to create clusters and machineConfigs from cluster template. You should be able to see logs in UI.

anupama2501:
I was able to verify this on v2.6-head b8a44fa

- Added a cluster template in the local --> Apps and market place --> repositories [https://github.com/StrongMonkey/cluster-template-examples.git] 
- Created a cluster as an admin and verified the logs are displayed in the api at `https://<ip>/v1/catalog.cattle.io.operations/fleet-default/helm-operation-dpp9q/logs`

```
helm install --generate-name=true --namespace=fleet-default --timeout=10m0s --values=/home/shell/helm/values-cluster-template-0.0.1.yaml --version=0.0.1 --wait=true /home/shell/helm/cluster-template-0.0.1.tgz
creating 2 resource(s)
beginning wait for 2 resources with timeout of 10m0s
NAME: cluster-template-0-1629765003
LAST DEPLOYED: Tue Aug 24 00:30:04 2021
NAMESPACE: fleet-default
STATUS: deployed
REVISION: 1
TEST SUITE: None

[...]

PR #34317, opened 2021-08-18
Add operation permission to standard user so they can see logs

StrongMonkey:
https://github.com/rancher/rancher/issues/34277

Add operation permission to standard users, as they should be able to use rke2 cluster template(helm chart), install and see helm operation logs.

The implemetation is changed so that when user creates operation through action, the corresponding role and rolebinding that only allow user to get this operation will be created along with operation. This only gives user ability to see operations that were created by itself.

### DIFF
--- a/pkg/catalogv2/helmop/operation.go
+++ b/pkg/catalogv2/helmop/operation.go
@@ -31,9 +31,12 @@ import (
 	data2 "github.com/rancher/wrangler/pkg/data"
 	"github.com/rancher/wrangler/pkg/data/convert"
 	corev1controllers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	rbacv1controllers "github.com/rancher/wrangler/pkg/generated/controllers/rbac/v1"
+	"github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/schemas/validation"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/authentication/user"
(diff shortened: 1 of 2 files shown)


pkg/data/management/role_data.go
```diff
@@ -452,6 +452,7 @@ func addUserRules(role *roleBuilder) *roleBuilder {
 	role.
 		addRule().apiGroups("").resources("secrets").verbs("create").
 		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch").
+		addRule().apiGroups("catalog.cattle.io").resources("operations").verbs("get", "list", "watch").
```
dramich, line 455: Won't this allow a standard user to see any operation?

StrongMonkey: Yes... I can limit it to `fleet-default` namespace, but you are still able to other user's operation.
Unless we go through the route that only who create operation can see it and view logs, but that requires more efforts to implement.

dramich: We spoke offline, but yes, the permissions have to be specific to the user and object.

pkg/catalogv2/helmop/operation.go
```diff
@@ -663,10 +671,68 @@ func (s *Operations) createOperation(ctx context.Context, user user.Info, status
...
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "Role",
+			Name:     role.Name,
+		},
+	}
+
+	if _, err := s.roles.Create(role); err != nil {
```
dramich, line 725: We should check the error for already exists so we don't return it and cause a loop of errors. Same applies below

StrongMonkey: https://github.com/rancher/rancher/blob/19e44aaf2046594f90dbaef8c1883e3d024666c9/pkg/catalogv2/helmop/operation.go#L802 The name is always generated so it is guaranteed to be unique. If it is not unique we should really fail though.

dramich (approved):

Issue #34175, opened 2021-08-11
RKE2 - Disabling etcd backup on a rancher-provisioned rke2 cluster sends wrong flag/config to rke2-server node


davidnuzik:
**Summary:**

If you create an RKE2 cluster with one of the node drivers in Rancher, and disable automatic snapshots in the UI, the wrong flag to disable snapshots is sent to the rke2-server node. This will result in the cluster always being in a "Provisioning" state in the UI.

**Environment:**
Rancher version: v2.6-head 9d2e08674 8/11/21 9:00 am Pacific
Rancher cluster type: single-node docker
Docker version: 19.03
 
Downstream cluster type: rke2
Downstream K8s version: v1.20.3-rc5+rke2r2

**Steps to Reproduce:**
1. Create any RKE2 cluster with a rancher node driver. In my case I made a simple 1-node cluster with the DigitalOcean driver. When you create the cluster, specify etcd > Automatic Snapshots > DISABLE in the cluster configuration section. Then create the cluster:
![image](https://user-images.githubusercontent.com/45179589/129098977-990d8659-9022-4bc8-8a02-dbff667cbf3a.png)

And the UI appears to send correct content in the request body to provision the cluster:
```
{
	"etcd": {
		"disableSnapshots": true,
		"s3": null,
		"snapshotRetention": 5,
		"snapshotScheduleCron": "0 */5 * * *"
	}
}
```
2. Notice the cluster will never come up it's always in a "Provisioning" state.
3. SSH into the RKE2 node and investigate the rke2-server service logs (`journalctl -u rke2-server`) You will see a message like as follows (truncated this to keep it short and to the point).
```
rke2[8562]: time="2021-08-11T20:23:10Z" level=fatal msg="flag provided but not defined: -etcd-disable-snapshot"
```

**Expected Result:**
I expect the backend to send the correct flag to the rke2-server. The correct flag is `--etcd-disable-snapshots` (note there is only one dash provided by rancher and it's missing the 's' in snapshots as well).

**Actual Result:**
The backend sends `-etcd-disable-snapshot` to the rke2-server node, not `--etcd-disable-snapshots`

**Additional Info:**
Full output from journalctl (this repeats forever as systemd keeps restarting the service):
```
Aug 11 20:23:09 dave-try-no-etcd-backup-again-pool1-bsf77 systemd[1]: rke2-server.service: Scheduled restart job, restart counter is at 164.
[...]

PR #34337, opened 2021-08-19
Fix ectd-disable-snapshots flag

thedadams:
The flag for RKE2 uses plural snapshots; Rancher was passing it as
singular. This fix adds to the 's'

Issue:
https://github.com/rancher/rancher/issues/34175


jakefhyde (approved):
LGTM

cmurphy (approved):

Issue #34330, opened 2021-08-19
Fleet - the fleet-agent-local app disappear when reinstalling Rancher into the same cluster


jiaqiluo:
**Rancher Server Setup**
- Rancher version: v2.6.0-rc3
- Installation option (Docker install/Helm Chart): Helm Chart 
   - If Helm Chart, Kubernetes Info:
         Cluster Type (RKE1, RKE2, k3s, EKS, etc): RKE1 
         Node Setup: 1 node, all roles
         Version: v1.20.9
 
**Information about the Cluster**
- Kubernetes version: v1.20.9
- Cluster Type (Local/Downstream): Local
 

**Describe the bug**

The `fleet-agent-local` app disappear when reinstalling Rancher into the same cluster

**To Reproduce**
 
- use helm to install rancher v2.6.0-rc3
- use helm to uninstall rancher 
- use helm to install rancher v2.6.0-rc3 again 
- 
**Result**
- the app `fleet-agent-local` is not installed in the local cluster 

**Expected Result**

 the app `fleet-agent-local` should be  installed in the local cluster

nickgerace:
Spoke with @StrongMonkey @jiaqiluo @cbron: solving this issue has implications for how the uninstall script should work in general. Since Provisioning v2 works in tandem with Fleet and Provisioning v2 CRD(s) are self-installed by Rancher, we may have to clean up specific CRs or crawl `*.cattle.io` APIs to delete self-installed CRDs.

The next steps would require design discussions around how we want the post delete script to work in general, specifically including how sidecar/required apps (e.g. Fleet and Webhook) work on "dirty" clusters.

nickgerace:
This issue is pending a targeted design meeting on the future of the uninstall script and what it should do: https://github.com/rancher/rancher/blob/master/chart/scripts/post-delete-hook.sh

nickgerace:
### Result of the Targeted Design Meeting - 07 Sep 2021
The uninstall script will keep CRD charts for now, which means its design is unchanged. This consensus emerged due to the following:

- The `post-delete-hook` script has worked with general usage in the past.
- If the CRD charts are uninstalled, then you may have to track _other_ resources that need to be deleted after that (@kinarashah).
  - You may inadvertently create another issue with re-installation after uninstalling Rancher.
  - We will have to do this anyway if we pursue the potential solution below.
- This issue has at least one "workaround".

### Potential Solution
[...]

PR #34344, opened 2021-08-19
Delete fleet-crd chart in post delete hook

nickgerace:
This _does_ contain copied code from the other loop, but in order for the UX to be consistent, this PR deletes remaining Fleet resources (in this case, just the Fleet CRD chart) after all other Rancher resources are deleted. The alternative would be to delete the `fleet-crd` chart if found in the main loop, but that might create confusion since no other CRD charts are uninstalled.

Relevant issue: https://github.com/rancher/rancher/issues/34330

### DIFF
--- a/chart/scripts/post-delete-hook.sh
+++ b/chart/scripts/post-delete-hook.sh
@@ -70,6 +70,15 @@ for namespace in ${namespaces}; do
   done
 done
 
+echo "Cleaning up Fleet resources..."
+for namespace in "cattle-fleet-system" "fleet-system"; do
+  for app in $(helm list -n "${namespace}" -q); do
+    if [[ "${app}" = "fleet-crd" ]] && [[ ! $(helm uninstall "${app}" -n "${namespace}") ]]; then
+      failed=("${failed[@]}" "${app}")
+    fi
+  done
+done
+
 echo "------ Summary ------"
 if [[ ${#succeeded[@]} -ne 0 ]]; then
   echo "Succeeded to uninstall the following apps:" "${succeeded[@]}"


nickgerace:
To reviewers: I would be open to transforming the delete logic in a function, but I did not think we needed to. Willing to accept an alternative solution, keeping this in mind: https://github.com/rancher/rancher/pull/34344#issue-975085791

jiaqiluo:
Currently, the script will traverse the following list of namespaces to delete non-crd apps:
https://github.com/rancher/rancher/blob/50c33b29b096ff97048d323daf9fdbda1634f332/chart/values.yaml#L147-L150

A simpler way to achieve the same goal is to NOT skip if the app is  `fleet-crd` 
https://github.com/rancher/rancher/blob/50c33b29b096ff97048d323daf9fdbda1634f332/chart/scripts/post-delete-hook.sh#L35-L37

nickgerace:
> A simpler way to achieve the same goal as your current fix is to NOT skip if the app is fleet-crd
@jiaqiluo my approach was intentional: https://github.com/rancher/rancher/pull/34344#issue-975085791

However, I'd be willing to take the simple approach as well. I just dont want the user to be surprised when they see a single CRD chart get deleted. Perhaps one `echo "--- Exception: CRD chart to be deleted:" could suffice though. What are your thoughts?

nickgerace:
@jiaqiluo tried to minimize this, but realized that the other uninstall function contained a `continue` statement. I think repeating ourselves is fine given the UX context of not wanting to delete a single CRD chart in the middle of skipping all of them.

chart/scripts/post-delete-hook.sh
```diff
@@ -70,6 +70,15 @@ for namespace in ${namespaces}; do
   done
 done
 
+echo "Cleaning up Fleet resources..."
+for namespace in "cattle-fleet-system" "fleet-system"; do
+  for app in $(helm list -n "${namespace}" -q); do
+    if [[ "${app}" = "fleet-crd" ]] && [[ ! $(helm uninstall "${app}" -n "${namespace}") ]]; then
```
jiaqiluo, line 76:
```suggestion
    if [[ "${app}" == "fleet-crd" ]] && [[ ! $(helm uninstall "${app}" -n "${namespace}") ]]; then
```

jiaqiluo (changes requested):

StrongMonkey (approved):

nickgerace:
Closing due to https://github.com/rancher/rancher/issues/34330#issuecomment-904040291

PR #34351, opened 2021-08-20
[2.6] Retrieve the UserContext for cluster dynamically

cmurphy:
When the cluster context is set for the proxy server at initialization,
it doesn't always refresh when a new request comes in. This can cause
the cache to become stale and the proxy server handler can fail to
fetch service accounts from it even though they exist in the downstream
cluster. This change pushes the cluster context getter down to the
request handling layer instead of the server creation layer so that it
is refreshed on every request.

### DIFF
--- a/pkg/clusterrouter/proxy/proxy_server.go
+++ b/pkg/clusterrouter/proxy/proxy_server.go
@@ -28,6 +28,10 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+type ClusterContextGetter interface {
+	UserContext(string) (*config.UserContext, error)
+}
+
 type RemoteService struct {
 	sync.Mutex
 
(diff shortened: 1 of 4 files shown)


pkg/clusterrouter/proxy/proxy_server.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/rancher/rancher/pkg/types/config/dialer"
 	"github.com/rancher/wrangler/pkg/schemas/validation"
+	"github.com/sirupsen/logrus"
```
snasovich, line 22: Not seeing usages of `logrus` here. Leftover from when debug logging was present?

snasovich (dismissed):
LGTM, one small nit

pkg/clusterrouter/proxy/proxy_server.go
cmurphy: Yep, fixed

snasovich (approved):
LGTM
@dramich , please review as soon as you have a chance. We would like to merge it today if it looks good.

dramich (approved):

PR #34362, opened 2021-08-20
[forwardport master] Retrieve the UserContext for cluster dynamically

dramich:
When the cluster context is set for the proxy server at initialization,
it doesn't always refresh when a new request comes in. This can cause
the cache to become stale and the proxy server handler can fail to
fetch service accounts from it even though they exist in the downstream
cluster. This change pushes the cluster context getter down to the
request handling layer instead of the server creation layer so that it
is refreshed on every request.


snasovich (approved):
Changes are the same as https://github.com/rancher/rancher/pull/34351

PR #34369, opened 2021-08-20
[master] Prefix RKE2 Windows node commands with "PowerShell"

rak-phillip:
This prefixes commands for windows registration with `PowerShell` so that the command can be run from either CMD or PowerShell. 

rancher/dashboard#3952


rak-phillip:
@phillipsj this one seems simple enough, but this is my first Rancher PR (I'm usually fiddling around ind dashboard). I see that you've been making changes to this file recently so I figured you might be a good person to review/provide guidance.

dramich:
@rak-phillip I think the issue you referenced is incorrect

phillipsj:
We originally made it that way. The UI explicitly states that the commands need to be executed in PowerShell would require that to be updated too. Has this been tested? We have ran into issues getting the syntax correct previously?

rak-phillip:
@dramich thanks for the heads up - fixed

@phillipsj thanks for the feedback. I originally had the same thought about the dashboard - I think it would suffice to emphasize that the command needs to run in powershell and I'm willing to push back a little more seeing that I'm not the only one who felt that way. I can toggle this PR over to a draft until we get some more info from the rest of the team - it seems like the fix might not be as straightforward as I originally thought.

phillipsj:
Ok, the PR looks good, so if we decide to make the change it will be ready.

phillipsj (approved):
LGTM

PR #34372, opened 2021-08-21
[2.6] Set the source on settings

brendarearden:
**Issue**
When an environment variable is used for settings, we want to know that it was set by an ENV var

**Solution**
Set the existing settings.Source field to "ENV" when an environment variable is used. This will help users know when env vars were used and will help them understand where to make settings changes. 

**Note:** I ran `go generate` and see that there were some files not generated. https://github.com/rancher/rancher/pull/34345 looks like it could've been from this.

### DIFF
--- a/pkg/controllers/dashboardapi/settings/settings.go
+++ b/pkg/controllers/dashboardapi/settings/settings.go
@@ -91,6 +91,7 @@ func (s *settingsProvider) SetAll(settingsMap map[string]settings.Setting) error
 				Default: setting.Default,
 			}
 			if value != "" {
+				newSetting.Source = "env"
 				newSetting.Value = value
 			}
 			if newSetting.Value == "" {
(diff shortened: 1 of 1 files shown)


rmweir (dismissed):
LGTM

pkg/controllers/dashboardapi/settings/settings.go
```diff
@@ -91,6 +91,7 @@ func (s *settingsProvider) SetAll(settingsMap map[string]settings.Setting) error
 				Default: setting.Default,
 			}
 			if value != "" {
+				newSetting.Source = "env"
 				newSetting.Value = value
```
cbron, line 95: Unrelated to your PR, but can we change the `value` variable name to `envValue`. As it is only ever used in that manner.

cbron (commented):
lgtm, one minor improvement. Didn't realize the write code existed under the dashboardAPI folder.

cbron (approved):

rmweir (approved):
LGTM

Issue #34380, opened 2021-08-22
Page not found displayed when deleting a cluster.


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `8898bf8`
- Installation option (Docker install/Helm Chart): HELM
   - If Helm Chart, Kubernetes Info: 
         Cluster Type (RKE1, RKE2, k3s, EKS, etc): RKE1
         Node Setup: 3 nodes all roles
         Version: 1.21.4
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: EC2 RKE2 node driver
- Cluster Type (Local/Downstream): `v1.21.3-rc6+rke2r2`

**Describe the bug**
Page not found displayed when deleting a cluster.

**To Reproduce**

- Deploy a cluster
- Delete the cluster
- Page not found displayed when deleting a cluster.

<img width="747" alt="Screen Shot 2021-08-22 at 3 02 44 PM" src="https://user-images.githubusercontent.com/26032343/130371463-239307a3-f1f0-4052-92de-0aa48d635d4c.png">

deniseschannon:
There was some aggressive changes on path stuff that we reverted and this use case should work now.

sowmyav27:
**On 2.6-head commit id: `b8a44fa`**

- Delete an RKE2 cluster
- No issues seen after.
- Closing this issue.

PR #34382, opened 2021-08-23
[2.6] Don't check user perference or settings for index anymore

deniseschannon:


deniseschannon:
https://github.com/rancher/rancher/issues/34380

Issue #34373, opened 2021-08-21
Unable to provision vSphere RKE2 cluster


PennyScissors:
**Rancher Server Setup**
- Rancher version: v2.6.0-rc5
- Installation option (Docker install/Helm Chart):
   - Cluster Type (RKE1, RKE2, k3s, EKS, etc): RKE2

**Information about the Cluster**
- Kubernetes version: v1.21.3-rc6+rke2r2
- Cluster Type (Local/Downstream): Downstream, Infrastructure Provider vSphere

**Describe the bug**

 When deploying a RKE2 cluster on vSphere I get the following error.
 ```
Failed creating server (VmwarevsphereMachine) in infrastructure provider: CreateError: Failure detected from referenced resource rke-machine.cattle.io/v1, Kind=VmwarevsphereMachine with name "screspo-test-2-pool1-3edb9647-47vv5": could not alter cloud-init file: open /path/to/machine/files/cloud-config: permission denied 
```

**To Reproduce**
 
Create a RKE2 cluster in vSphere

```
apiVersion: provisioning.cattle.io/v1
kind: Cluster
metadata:
  annotations:
    field.cattle.io/creatorId: user-z6h66
  creationTimestamp: "2021-08-20T23:52:05Z"
  finalizers:
  - wrangler.cattle.io/provisioning-cluster-remove
  generation: 2
  managedFields:
  - apiVersion: provisioning.cattle.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .: {}
          v:"wrangler.cattle.io/provisioning-cluster-remove": {}
      f:spec:
        .: {}
        f:cloudCredentialSecretName: {}
        f:kubernetesVersion: {}
        f:rkeConfig:
          .: {}
          f:chartValues:
            .: {}
            f:rke2-calico: {}
          f:etcd:
            .: {}
            f:snapshotRetention: {}
            f:snapshotScheduleCron: {}
          f:localClusterAuthEndpoint: {}
          f:machineGlobalConfig:
            .: {}
            f:cni: {}
            f:disable-kube-proxy: {}
            f:etcd-expose-metrics: {}
            f:profile: {}
          f:machinePools: {}
          f:machineSelectorConfig: {}
          f:registries: {}
          f:upgradeStrategy:
            .: {}
            f:controlPlaneConcurrency: {}
            f:controlPlaneDrainOptions:
              .: {}
              f:timeout: {}
[...]

PR #34400, opened 2021-08-23
Fix provisioning machine files permissions

thedadams:
A previous update changed the machine-provisioning job to run as a
non-root user, as well as updating permissions to volumes to ensure the
new user can access the appropriate files. There was one volume that was
missed, and this change updates the permissions of that volume to allow
the machine-provisioning job to access the files it needs.

Issue:
https://github.com/rancher/rancher/issues/34373


jakefhyde (approved):
LGTM

snasovich (approved):
LGTM

dramich (approved):

PR #34421, opened 2021-08-24
[v2.6] Add RBAC rule for cleanup job user

cmurphy:
The cleanup job service account now needs to be able to look up the
presence of the rancher service in the cluster and fails to perform any
cleanup if it does not have permission. This change adds the appropriate
RBAC rule to the main cluster role for the cleanup job.

### DIFF
--- a/pkg/controllers/management/usercontrollers/controller.go
+++ b/pkg/controllers/management/usercontrollers/controller.go
@@ -210,6 +210,13 @@ func (c *ClusterLifecycleCleanup) createCleanupClusterRole(userContext *config.U
 			APIGroups: []string{"batch"},
 			Resources: []string{"jobs"},
 		},
+		// The job checks for the presence of the rancher service first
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/usercontrollers/controller.go
```diff
@@ -210,6 +210,13 @@ func (c *ClusterLifecycleCleanup) createCleanupClusterRole(userContext *config.U
 			APIGroups: []string{"batch"},
 			Resources: []string{"jobs"},
 		},
+		// The job checks for the presence of the rancher service first
+		rbacV1.PolicyRule{
+			Verbs:         []string{"list", "get"},
```
dramich, line 215: Is list required? The call is a `GET` that checks for the service

cmurphy: Fixed

dramich (approved):

Issue #34115, opened 2021-08-10
rke-ingress-controller-deploy-job fails on rke cluster update


rawmind0:
**If this is a UI issue, please file an issue at https://github.com/rancher/dashboard/**

**Rancher Server Setup**
- Rancher version: v2.6.0-rc1
- Installation option (Docker install/Helm Chart):
   Both

**Information about the Cluster**
- Kubernetes version: v1.21.3-rancher1
- Cluster Type (Local/Downstream): Custom Downstream

**Describe the bug**
Starting from a fully working custom rke downstream cluster. Update the cluster, setting for example a default private registry (in fact, any cluster modification that implies an update at `rke-ingress-controller-deploy-job`, would cause the issue). Downstream cluster is being updated properly (docker images are updated to use private registry), but the `rke-ingress-controller-deploy-job` is failing to be updated.

`rke-ingress-controller-deploy-job` job is trying to update `ingress-nginx-admission-create` and `ingress-nginx-admission-patch` jobs,to set the new private registry image, but k8s jobs are immutable, so the jobs should be recreated instead of patched. Once `ingress-nginx-admission-create` and `ingress-nginx-admission-patch` jobs are manually deleted, `rke-ingress-controller-deploy-job` is finishing fine and the cluster becomes full updated. 

**To Reproduce**

Test deployment env:
* Private docker registry up and running with all required images pushed.
* Deploy rancher 2.6.0-rc1 and create custom rke downstream cluster.
* Once the cluster is fully deployed, edit it, adding private registry and credentials.
* Downstream cluster is almost fully updated properly (images are updated to use private registry), but there is an issue updating `rke-ingress-controller-deploy-job` .  

**Result**
That's the error log got from `rke-ingress-controller-deploy-job` job
```
namespace/ingress-nginx unchanged
serviceaccount/nginx-ingress-serviceaccount unchanged
configmap/nginx-configuration unchanged
clusterrole.rbac.authorization.k8s.io/ingress-nginx unchanged
clusterrolebinding.rbac.authorization.k8s.io/ingress-nginx unchanged
role.rbac.authorization.k8s.io/ingress-nginx unchanged
rolebinding.rbac.authorization.k8s.io/ingress-nginx unchanged
service/ingress-nginx-controller-admission unchanged
daemonset.apps/nginx-ingress-controller configured
validatingwebhookconfiguration.admissionregistration.k8s.io/ingress-nginx-admission configured
serviceaccount/ingress-nginx-admission unchanged
[...]

PR #34424, opened 2021-08-24
Bump rke go mod to v1.3.0-rc17

rawmind0:
Forward issue: https://github.com/rancher/rancher/issues/34115


rawmind0:
At the moment, rancher is already using laster rke master commit.  Forward issue #34115 is already included

PR #34427, opened 2021-08-24
Add RBAC rule for cleanup job user

cmurphy:
The cleanup job service account now needs to be able to look up the
presence of the rancher service in the cluster and fails to perform any
cleanup if it does not have permission. This change adds the appropriate
RBAC rule to the main cluster role for the cleanup job.

Forward port of https://github.com/rancher/rancher/pull/34421


snasovich (approved):
Same as source PR

thedadams (approved):

PR #34428, opened 2021-08-24
bump kubernetes-provider-detector and add rke2-windows and harvester providers

rosskirkpat:
for https://github.com/rancher/kubernetes-provider-detector/commit/2bd6fed84b5f99f7687e42a1d965062ce0c3a579


pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
```diff
@@ -33,10 +33,15 @@ func Register(ctx context.Context,
...
 	if cluster == nil || cluster.Status.Provider != "" {
+		// rke1 windows clusters check
 		// if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
 		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || cluster.Status.Provider == "rke.windows" {
 			return cluster, nil
 		}
+		// rke2 windows clusters check
+		if cluster == nil || cluster.Status.Provider == "rke2.windows" {
+			return cluster, nil
+		}
```
dramich, line 44: Per offline: merge the checks and swap to a endswith check since harvester will probably be coming soon

pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
```diff
@@ -2,12 +2,12 @@ package kubernetesprovider
 
 import (
 	"context"
-
 	detector "github.com/rancher/kubernetes-provider-detector"
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v32 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/wrangler"
 	"k8s.io/client-go/kubernetes"
+	"strings"
```
dramich, line 10: This should be goimport sorted/formatted

pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
```diff
@@ -33,8 +33,10 @@ func Register(ctx context.Context,
 
 func (h *handler) OnChange(key string, cluster *v3.Cluster) (*v3.Cluster, error) {
 	if cluster == nil || cluster.Status.Provider != "" {
-		// if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
-		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || cluster.Status.Provider == "rke.windows" {
+		// rke1 and rke2 windows cluster check
+		// rke1: if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
+		// rke2: if provider is not rke2.windows, continue to detect if it has windows nodes
+		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || strings.HasSuffix(cluster.Status.Provider, "windows") {
```
dramich, line 39:
```suggestion
		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || strings.HasSuffix(cluster.Status.Provider, ".windows") {
```

dramich (approved):
Seems fine to me, @StrongMonkey any concerns with the change in the windows check?

StrongMonkey (approved):

pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
```diff
@@ -33,8 +34,10 @@ func Register(ctx context.Context,
 
 func (h *handler) OnChange(key string, cluster *v3.Cluster) (*v3.Cluster, error) {
 	if cluster == nil || cluster.Status.Provider != "" {
-		// if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
-		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || cluster.Status.Provider == "rke.windows" {
+		// rke1 and rke2 windows cluster check
+		// rke1: if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
+		// rke2: if provider is not rke2.windows, continue to detect if it has windows nodes
+		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || strings.HasSuffix(cluster.Status.Provider, ".windows") {
```
StrongMonkey, line 40: Actually...Do we set `WindowsPreferedCluster` for rke2? Otherwise this will not work

StrongMonkey (changes requested):
The logic we follow at here is

If cluster is created as windows type(by setting .spec.WindowsPreferedCluster) and it is not detected as windows cluster(no windows node shows up), we will try to run detect logic again. 

So if you don't set this field, it will always return once provider is set. So if the cluster is previously set as rke2, it can't be updated.

pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
rosskirkpat: We don't set that for rke2 clusters. do I need to have a second check specifically for rke2 that doesn't check `WindowsPreferedCluster`? something like this:
```
if cluster == nil || cluster.Status.Provider == "rke2.windows" {
```

StrongMonkey: Yeah we need similar logic for that. For rke1 use the existing logic. For rke2 we just have to constantly process it, because at any time if a window node is registered then we treat it as windows cluster. Feel bad that this is getting more complicated.

StrongMonkey: Is there a way to know if a rke2 cluster is going to a windows cluster? any field on api?

deniseschannon: A RKE2 cluster can be a linux cluster *until* you add a windows worker node. It's unlike RKE1 where you define it up front.

pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
```diff
@@ -33,10 +33,15 @@ func Register(ctx context.Context,
 
 func (h *handler) OnChange(key string, cluster *v3.Cluster) (*v3.Cluster, error) {
 	if cluster == nil || cluster.Status.Provider != "" {
-		// if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
+		// rke1 and rke2 windows cluster check
+		// rke1: if cluster has windows enabled and provider is not rke.windows, continue to detect if it has windows nodes
+		// rke2: if provider is not rke2.windows, continue to detect if it has windows nodes
 		if cluster == nil || !cluster.Spec.WindowsPreferedCluster || cluster.Status.Provider == "rke.windows" {
```
StrongMonkey, line 39: This won't work. For rke2 cluster `cluster.Spec.WindowsPreferedCluster` is always false so the first if will always to true, so controller won't detect provider once it is set.

Issue #34404, opened 2021-08-23
Cluster members who is Standard Users needs to be able to see machine pools but not ability to update


deniseschannon:
**Rancher Server Setup**
- Rancher version:v2.6-head - b8a44fa

**Information about the Cluster**
- Kubernetes version: v1.21.3-rc6+rke2r2
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): DO cluster running RKE2

**Describe the bug**

When adding a standard user to a cluster as a cluster member, the cluster config page shows the nodes not grouped in the machine pool.

**To Reproduce**
 
1. Create a cluster
2. Add a standard user as a cluster member
3. log in as standard user
4. Go to cluster management. 
5. Click on the cluster name to view the cluster config

**Result**

The nodes are viewable but they are not in a pool, which is invalid. This is because the standard user, cluster member doesn't have read only view of the machine pools API

**Expected Result**

Expect them to be grouped together in a machine pool.

**Screenshots**

<img width="1192" alt="Screen Shot 2021-08-23 at 4 06 47 PM" src="https://user-images.githubusercontent.com/11410997/130531411-813a1587-e469-4a12-b189-30965d95a1f8.png">

**Additional context**

deniseschannon:
Validated with v2.6-head 70f024e

Both use cases above can see machines in machine pools on the cluster detail page.

<img width="1185" alt="Screen Shot 2021-08-24 at 9 09 55 PM" src="https://user-images.githubusercontent.com/11410997/130725328-10d82f32-924b-4a47-b2d8-89f1bddf28b1.png">

BUT a user base with manage nodes role is unable to scale up or down the machine pool. They have the ability to SSH. Scaling up and down is part of machine pool. 

A user with manage nodes role should have the ability to update the machine pools. 

Note: There is another issue around machine pools that are filed separately:
User base with manage nodes doesn't see the cluster config: https://github.com/rancher/dashboard/issues/4000
Cluster owners/cluster members needing to be able to list cloud credentials: https://github.com/rancher/rancher/issues/34442 & https://github.com/rancher/rancher/issues/34443

PR #34429, opened 2021-08-24
Add machinedeployment permissions

dramich:
Problem:
machinedeployments are required to see the pool machines are associated
with and some roles are missing that permissions.

Solution:
Add permission to roles that also have access to the machine object

https://github.com/rancher/rancher/issues/34404


cmurphy (approved):

snasovich:
Do we need to add the same to Cluster Owner role template as it defines `cluster.x-k8s.io.machines`?

dramich:
> Do we need to add the same to Cluster Owner role template as it defines `cluster.x-k8s.io.machines`?

No, cluster owner perms are mostly for show in there since they already have `**`

snasovich:
> > Do we need to add the same to Cluster Owner role template as it defines `cluster.x-k8s.io.machines`?
> 
> No, cluster owner perms are mostly for show in there since they already have `**`

Ah, I see: `		addRule().apiGroups("*").resources("*").verbs("*").`

snasovich (approved):

Issue #34396, opened 2021-08-23
Need a way to retrieve the externalip for k3s/rke2


rancher-max:
**Rancher Server Setup**
- Rancher version: v2.6-head commit `b8a44fa`
- Installation option (Docker install/Helm Chart): helm - rke1 - v1.21.4 -3 nodes

**Information about the Cluster**
- Kubernetes version: v1.21.3-rc6+rke2r2 and v1.21.3+k3s1 -- should not matter if using rc7 or v1.21.4 
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Node-driver, AWS. DO related.

**Describe the bug**

When bringing up an rke2 or k3s cluster, the `node-external-ip` flag is not set. This means that only InternalIP is being set on the nodes. For DigitalOcean, this InternalIP is actually external, which is fine. For AWS, there is an ExternalIP available but it is not being used at all and only it's private internal ip is being set. This causes an issue for EC2 nodes when specifying Authorized Cluster Endpoint, since the resulting downloaded Kubeconfig file will be using private, internal IPs for the nodes instead of external.

**To Reproduce**
 
1. Create an rke2 or k3s cluster using dashboard UI, and set Authorized Cluster Endpoint to being used.
2. Download the kubeconfig file after it's used

**Result**
Private IPs are added for additional contexts

**Expected Result**

Public IPs should be added instead

**Screenshots**

N/A

**Additional context**

There are annotations set on the nodes: https://github.com/k3s-io/k3s/blob/2705431d9645d128441c578309574cd262285ae6/pkg/cloudprovider/instances.go#L17-L19
However, we do not have the the `external-ip` one being set when we don't provide the `node-external-ip` flag.

Unsure if the proper approach is always setting that on the nodes not, but we need some way to retrieve the external ip (**unless we specifically select the option to use private or internal ips only when bringing up the cluster**) when also selecting ACE.

sowmyav27:
**On 2.6-head commit id: `b8a44fa`**

- Deploy a downstream RKE2 AWS cluster
- Navigate to cluster explorer, Kube nodes page
- External IP/Internal IP  column says None/<internal-ip>

PR #34433, opened 2021-08-25
Get IP address for machine from state secret

thedadams:
In order to properly set external IP address on an RKE2/K3S node, the IP
address must be passed as part of the config. With this change, we pull
the IP address from the machine state secret created during
provisioning.

Issue:
https://github.com/rancher/rancher/issues/34396


paynejacob (dismissed):
The solution makes sense to me.

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -798,14 +800,35 @@ func PruneEmpty(config map[string]interface{}) {
...
+	}
+
+	if secret, err := secrets.Get(machine.Spec.InfrastructureRef.Namespace, name.SafeConcatName(machine.Spec.InfrastructureRef.Name, "machine", "state")); err != nil {
+		return err
+	} else if secret != nil && len(secret.Data["extractedConfig"]) != 0 {
+		if driverConfig, err := nodeconfig.ExtractConfigJSON(base64.StdEncoding.EncodeToString(secret.Data["extractedConfig"])); err != nil {
+			return err
+		} else if len(driverConfig) != 0 && len(convert.ToString(values.GetValueN(driverConfig, "Driver", "IPAddress"))) != 0 {
+			config["node-external-ip"] = convert.ToString(values.GetValueN(driverConfig, "Driver", "IPAddress"))
+		}
```
snasovich, line 826: Is it OK we do nothing if the condition from line 824 is not met (so no error and no `"node-external-ip"` set)? The error return on line 828 does not cover this.

snasovich (approved):

dramich:
I am concerned that this will now block the plan if an error is encountered. This did not return an error before, so if some condition was missed that we always end up in an error state it will never proceed. That seems like a much worse outcome than having the internal IP showing up.

Issue #34365, opened 2021-08-20
Snapshot restoration fails in Rancher downstream RKE2 clusters


Jono-SUSE-Rancher:
Mirror issue of https://github.com/rancher/rke2/issues/1685 - we will need to get the RKE2 team to fix this so we can unblock.

deniseschannon:
Based on the RKE2 investigation, 

It appears as though the cluster is generating a plan for system-agent that includes S3 flags. 1 to enable S3 and one for the bucket. I confirmed with @davidnuzik that the Rancher install is barebones and no S3 configuration for backup/restore.

```sh
root@dave-1node-repro1685-pool1-46807efc-wdbvt:/var/lib/rancher/agent/applied# cat 20210825-020322 | jq -r '.Plan.instructions[4]'
{
  "name": "restore",
  "args": [
    "server",
    "--cluster-reset",
    "--cluster-reset-restore-path=on-demand-dave-1node-repro1685-pool1-46807efc-wdbvt-1629855943",
    "--etcd-s3",
    "--etcd-s3-bucket="
  ],
  "command": "rke2"
}
```

This can be confirmed from the UI by going to this link below. It shows the contents of the generated secret. 

https://143.198.61.244/dashboard/c/local/explorer/secret/fleet-default/dave-1node-repro1685-bootstrap-template-4nspl-machine-plan#data

PR #34441, opened 2021-08-25
[2.6] Only append s3 args if necessary

Oats87:
https://github.com/rancher/rancher/issues/34365
https://github.com/rancher/rke2/issues/1685


deniseschannon (approved):
Offline comment by @ibuildthecloud that this looked good

Issue #34437, opened 2021-08-25
Update provisioning tests to use the correct environment variable for system-agent image


deniseschannon:
Since @Oats87 changed the environment variable for system-agent, we didn't update how it was [used in the provisioning tests](https://github.com/rancher/rancher/blob/master/scripts/provisioning-tests#L28) which resulted in a change that would always pick up the latest system-agent image instead of the one that was in the [package/Dockerfile](https://github.com/rancher/rancher/blob/release/v2.6/package/Dockerfile#L43). 

@Oats87 Can you update the environment variable and see if the original one was used anywhere else in the code?

Oats87:
I don't believe I caused this issue, as I didn't change the environment variable name. It was added in the manner it is in here: https://github.com/rancher/rancher/commit/3abb547f743cfcde471c564b718d01b602faaef0

PR is incoming to update the environment variable.

deniseschannon:
@thedadams @dramich Is there a way to check if the provisioning tests are using the system-agent that's in Dockerfile?

thedadams:
My investigation concluded that the provisioning tests did not properly pay attention to the `CATTLE_SYSTEM_AGENT_VERSION` variable.

The linked PR ensures that the provisioning tests do properly use the `CATTLE_SYSTEM_AGENT_VERSION` and adds a test to prove it.

Once merged, I don't believe that there is anything for QA to do in this case.

PR #34444, opened 2021-08-25
[2.6] Use correct system-agent version in provisioning-tests and bump system-agent

Oats87:


deniseschannon (approved):
Offline comment by @ibuildthecloud that this looked good

PR #34446, opened 2021-08-25
Only append s3 args if necessary

Oats87:
https://github.com/rancher/rancher/issues/34365
https://github.com/rancher/rke2/issues/1685


deniseschannon (approved):
Based on approval on original PR #34441

PR #34523, opened 2021-08-27
Set min webhook version to 1.0.0+up0.2.0

rmweir:


cbron (approved):
good catch, don't forget to forwardport

dramich (approved):

Issue #34500, opened 2021-08-26
Unable to generate kubeconfig after enabling Authorized Cluster Endpoint on RKE2 ec2 clusters


bmdepesa:
rancher/rancher:v2.6.0-rc7

* Create an RKE2 ec2 cluster (custom or nodedriver)
* Edit the cluster to enable Authorized Cluster Endpoint
* Generate the kubeconfig 

Expected:
* Kubeconfig is generated with additional contexts

Actual
* Kubeconfig fails to generate with a 500 when posting
```
POST <url>/v3/clusters/c-m-885g4nfq?action=generateKubeconfig
{"baseType":"error","code":"ServerError","message":"the server could not find the requested resource (post clusterauthtokens.meta.k8s.io)","status":500,"type":"error"}
```
* Disabling ACE allows the kubeconfig to be generated again

bmdepesa:
v2.6-head 8bc6327

The same problem is still occurring. It looks to be mostly unrelated to ACE, but when editing the cluster to enable ACE, the edit is not successful and the cluster goes into a bad state with all of the system workloads stuck `updating`. 

@thedadams has a setup and is investigating.

thedadams:
The problem seen in this issue is not related to ACE. It is this issue: https://github.com/rancher/rancher/issues/34497

I used a setup from @bmdepesa to investigate and I believe I have fixed the linked issue. PRs are in Review.

bmdepesa:
rancher/rancher:v2.6-head 1ae31b1

* Created single and multinode ec2 nodedriver and custom clusters for RKE2 v1.21.4+rke2r2 without ACE enabled
* Generate a kubeconfig for each cluster and check it's functionality
* After clusters are active, edit the cluster to enable ACE
* Each cluster is updated and becomes active
* Kubeconfigs are generated with additional contexts and contexts are functional
* Edit each cluster to disable ACE
* Kubeconfigs are only generated with one context and are functional

mscbpi:
Hi,

- Rancher `v2.6.6` 
- downstream cluster `v1.23.7+rke2r2`
- vSphere provisioner (provisioned with Rancher)
- Created without ACE
- Enabled ACE => Error 500 on `generateKubeconfig`
- ACE works since creating a manual `KUBECONFIG` with IP of a master node and `certificate-authority-data` taken from the master node `/etc/rancher/rke2.yaml` works well...

PR #34524, opened 2021-08-27
Restart cluster controllers on RKE2 ACE change

thedadams:
In order to successfully enable ACE on an RKE2 cluster that was created
with ACE disabled, the cluster controllers must be restarted. This
change allows the controllers to detect when they should be restarted
because of an ACE change.

Issue:
https://github.com/rancher/rancher/issues/34500


rmweir (approved):
LGTM

dramich (approved):

PR #34527, opened 2021-08-28
Fix namespaces not being moved to System project on upgrade

jakefhyde:
Environment
-
Install: `HA`
Rancher version: `v2.5.8` + local changes
Kubernetes version: `v1.20.9`

Changes
-
- Added logic to re assign namespaces to default & system projects if out of date

Notes
-
- If `settings.management.cattle.io/system-namespaces` CRD is updated via `kubectl`, changes are not applied until the project is updated (can be done by editing description in UI and saving). Since nothing currently updates `system-namespaces` besides a rancher upgrade (which almost always in practice will have changes to the system project), may not be an issue.
- Moving `rancher-operator-system` or any other system namespace to another project, will successfully move the namespace (and update is not triggered for the project it is leaving). If the project is updated, it is then moved back. This is probably not desired, needs discussion.


cbron:
This has been moved to 2.5.11. Going to wait to review until 2.5.10 is over.

pkg/controllers/managementuser/rbac/project_handler.go
```diff
@@ -113,21 +124,25 @@ func (p *pLifecycle) ensureNamespacesAssigned(project *v3.Project) error {
...
+	_, err = p.m.workload.Management.Management.Clusters("").Update(cluster)
+
+	return err
 }
 
 func (p *pLifecycle) ensureDefaultNamespaceAssigned(cluster *v3.Cluster, project *v3.Project) error {
-	_, err := v32.ClusterConditionDefaultNamespaceAssigned.DoUntilTrue(cluster.DeepCopy(), func() (runtime.Object, error) {
-		return nil, p.assignNamespacesToProject(project, projectpkg.Default)
+	_, err := v32.ClusterConditionDefaultNamespaceAssigned.DoUntilTrue(cluster, func() (runtime.Object, error) {
+		return cluster, p.assignNamespacesToProject(project, projectpkg.Default)
```
thedadams, line 138: You don't need to return the cluster object here if you aren't doing anything with it afterwards.

pkg/controllers/managementuser/rbac/project_handler.go
```diff
@@ -113,21 +124,25 @@ func (p *pLifecycle) ensureNamespacesAssigned(project *v3.Project) error {
...
 	})
 	return err
 }
 
 func (p *pLifecycle) ensureSystemNamespaceAssigned(cluster *v3.Cluster, project *v3.Project) error {
-	_, err := v32.ClusterConditionSystemNamespacesAssigned.DoUntilTrue(cluster.DeepCopy(), func() (runtime.Object, error) {
-		return nil, p.assignNamespacesToProject(project, projectpkg.System)
+	_, err := v32.ClusterConditionSystemNamespacesAssigned.DoUntilTrue(cluster, func() (runtime.Object, error) {
+		return cluster, p.assignNamespacesToProject(project, projectpkg.System)
```
thedadams, line 145: Same as above.

pkg/controllers/managementuser/rbac/project_handler.go
```diff
@@ -156,6 +171,7 @@ func (p *pLifecycle) assignNamespacesToProject(project *v3.Project, projectName
 		}
 		ns.Annotations[projectIDAnnotation] = fmt.Sprintf("%v:%v", p.m.clusterName, project.Name)
 		if _, err := p.m.workload.Core.Namespaces(p.m.clusterName).Update(ns); err != nil {
+			log.Errorf("projectHandler: assignNamespacesToProject: error adding project id annotation to namespace err=%v", err)
```
thedadams, line 174: The `err` will get logged by the handler if you return it, so it doesn't need to be logged here.

pkg/controllers/managementuser/rbac/project_handler.go
```diff
@@ -4,19 +4,30 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/client-go/util/retry"
```
thedadams, line 7: nit: put this import down with the other `k8s` imports.

thedadams (changes requested):

pkg/controllers/managementuser/rbac/project_handler.go
jakefhyde: Fixed, seems like I shouldn't have force pushed

pkg/controllers/managementuser/rbac/project_handler.go
```diff
@@ -14,9 +14,18 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 )
 
 func newProjectLifecycle(r *manager) *pLifecycle {
+	// set SystemNamespacesAssigned Unknown so that system namespaces are reassigned on startup/upgrade
+	retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		cluster, err := r.workload.Management.Management.Clusters("").Controller().Lister().Get("", r.clusterName)
```
thedadams, line 23: Should probably check this error.

thedadams (changes requested):

Issue #34497, opened 2021-08-26
[RKE2 provisioning] unable to add dedicated worker node to cluster using amazon EC2 as the provider


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.6.0-rc7
- Installation option (Docker install/Helm Chart):helm
   - If Helm Chart, Kubernetes Info:
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):rke1
         Node Setup:
         Version: v1.21.4
- Proxy/Cert Details:valid certs (letsencrypt)

**Information about the Cluster**
- Kubernetes version: v1.21.4-rc1+rke2r1
- Cluster Type (Local/Downstream):downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):infrastructure provider, AWS
Security Group: default (rancher-nodes)

**Describe the bug**

user's dedicated worker node is not registering with the cluster. it is created in the backend, and other node types are registering.

**To Reproduce**
 
* using ec2 as the provider, create 3 separate node pools each with 1 role each
 * wait for the cluster to come to `active`

**Result**
* etcd and control plane nodes come to active, but the worker node is stuck in state `waiting for node to be registered with Kubernetes`

**Expected Result**

all nodes should register with the cluster

**Screenshots**

**Additional context**

* the worker node logs are empty for `rke2-server`
* other providers like DO come up fine with dedicated nodes

sowmyav27:
**on 2.6.0-rc9**

- Deployed an RKE2 AWS cluster - 1 etcd, 1 cp and 1 worker nodes
- Worker node is still stuck in Provisioning state
- `journalctl -u rancher-system-agent` logs

```
Aug 29 05:23:12 ip-172-31-13-77 rancher-system-agent[1348]: time="2021-08-29T05:23:12Z" level=debug msg="[K8s] writing an applied checksum value of ceacbd5533388438e9a509325d945942b91df2fad36a0a504aef1cf3aeea4c65 to the remote plan"
Aug 29 05:23:12 ip-172-31-13-77 rancher-system-agent[1348]: time="2021-08-29T05:23:12Z" level=debug msg="[K8s] Enqueueing after 5.000000 seconds"
Aug 29 05:23:12 ip-172-31-13-77 rancher-system-agent[1348]: time="2021-08-29T05:23:12Z" level=debug msg="[K8s] secret data/string-data did not change, not updating secret"
[...]

Issue #32390, opened 2021-04-27
RKE2 Provisioning: Cluster Options - Cloud Provider Support


deniseschannon:
Ability to specify 

AWS
Azure
~Custom~
"external" (to be tested in https://github.com/rancher/rancher/issues/33611)

This is available in the agentArgs (https://github.com/rancher/rancher/issues/32376) issue, but calling it out specifically to have it specially tested for these use cases.

deniseschannon:
I've created a specific ticket for vsphere since it leverages separate helm charts vs. cloud provider config within the cluster

davidnuzik:
Custom is not an option in the UI so will skip this. There is (None), aws, azure, gcp, vsphere, and external.

davidnuzik:
**My checks FAILED**
 
**Reproduction Steps:**

Not required.
 
**Validation Steps:**

Rancher version: v2.6-head 4b6af95d7 8/9/21 8:46 am Pacific
Rancher cluster type: single-node docker
Docker version: 19.03
 
Downstream cluster type: rke2
Downstream K8s version: v1.21.3-rc4+rke2r2
 
Validation steps:
_**AWS Use Case - FAIL**_
1. Provision an RKE2 cluster with the AWS provider in Rancher. Choose "aws" as the Cloud Provider in the Cluster Configuration section in the UI. Be sure to specify a valid IAM role as well as this is required for the aws cloud provider.
2. The cloud provider "aws" is indeed set on the RKE2 node:
```
root@ip-REDACTED_IP:/etc/rancher/rke2/config.yaml.d# cat 50-rancher.yaml 
{
  "agent-token": "SECRET-REDACTED-TOKEN",
  "cloud-provider-name": "aws",
  "cni": "calico",
  "disable-kube-proxy": false,
  "etcd-expose-metrics": false,
  "etcd-snapshot-retention": 5,
  "etcd-snapshot-schedule-cron": "0 */5 * * *",
  "node-label": [
    "rke.cattle.io/machine=1a3a2bb4-9df3-41b1-92d4-0ad940264532"
  ],
  "protect-kernel-defaults": false,
  "token": "SECRET-REDACTED-TOKEN"
```

Setting up the AWS cloud provider on an rke1 node works as expected - this only doesn't work with RKE2. The cluster will never get out of "Provisioning" state:
![image](https://user-images.githubusercontent.com/45179589/128790971-b2fcaf92-8cfa-40f6-8e56-34a316384e74.png)

[...]

PR #34528, opened 2021-08-28
Be more particular when setting node-external-ip

thedadams:
A previous change set the node-external-ip if the user did not provide
one. Instead, the node-external-ip is only set if a cloud provider is
not provided and if an IP address and private IP address can be
identified and are different. In addition, if the node-external-ip
address is set, then the advertise-address is set to the internal
address to maintain consistency.

Issues:
https://github.com/rancher/rancher/issues/32390
https://github.com/rancher/rancher/issues/34497

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -847,19 +847,23 @@ func addAddresses(secrets corecontrollers.SecretCache, config map[string]interfa
 		return
 	}
 
-	if machine.Annotations["objectset.rio.cattle.io/id"] == "unmanaged-machine" || machine.Spec.InfrastructureRef.Kind == "PodMachine" {
+	if config["cloud-provider-name"] != nil || machine.Annotations["objectset.rio.cattle.io/id"] == "unmanaged-machine" {
 		// Unmanaged machines won't have a state secret,
-		// and PodMachine's IP address is not external
+		// and the cloud provider will handle this, if it is set.
 		return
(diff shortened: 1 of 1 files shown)


pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -847,18 +847,21 @@ func addAddresses(secrets corecontrollers.SecretCache, config map[string]interfa
 		return
 	}
 
-	if machine.Annotations["objectset.rio.cattle.io/id"] == "unmanaged-machine" || machine.Spec.InfrastructureRef.Kind == "PodMachine" {
+	if config["cloud-provider-name"] != nil || machine.Annotations["objectset.rio.cattle.io/id"] == "unmanaged-machine" {
 		// Unmanaged machines won't have a state secret,
-		// and PodMachine's IP address is not external
+		// and if the cloud provider will handle this, if it is set.
```
snasovich, line 852:
```suggestion
		// and the cloud provider will handle this, if it is set.
```
Super-nit, but I think first `if` is not needed in this comment.

snasovich (dismissed):
LGTM, one nit on comment language

kinarashah (dismissed):

pkg/provisioningv2/rke2/planner/planner.go
thedadams: Done.

snasovich (approved):

kinarashah (approved):

PR #34544, opened 2021-08-30
Fix worker config and refactor IP address setting

thedadams:
First, and foremost, this change is to address the worker config getting
an invalid setting: advertise-address is a server only setting.

Secondly, this refactors the code that sets the IP addresses so that
advertise address is set whether or not a cloud-provider has been
chosen. This is to ensure that the internal address is always used for
the apiserver IP.

Issue:
https://github.com/rancher/rancher/issues/34497

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -840,35 +840,37 @@ func PruneEmpty(config map[string]interface{}) {
 }
 
 func addAddresses(secrets corecontrollers.SecretCache, config map[string]interface{}, machine *capi.Machine) {
-	if data := machine.Annotations[InternalAddressAnnotation]; data != "" {
-		config["node-ip"] = data
-	}
-
-	if data := machine.Annotations[AddressAnnotation]; data != "" {
-		config["node-external-ip"] = data
-		// Don't check for an IP address from the secret if it is already provided.
-		return
-	}
-
-	if config["cloud-provider-name"] != nil || machine.Annotations["objectset.rio.cattle.io/id"] == "unmanaged-machine" {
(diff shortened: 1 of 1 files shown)


pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -839,35 +842,22 @@ func PruneEmpty(config map[string]interface{}) {
...
-	secret, err := secrets.Get(machine.Spec.InfrastructureRef.Namespace, name.SafeConcatName(machine.Spec.InfrastructureRef.Name, "machine", "state"))
-	if err == nil && secret != nil && len(secret.Data["extractedConfig"]) != 0 {
-		driverConfig, err := nodeconfig.ExtractConfigJSON(base64.StdEncoding.EncodeToString(secret.Data["extractedConfig"]))
-
-		if err == nil && len(driverConfig) != 0 {
-			ipAddress := convert.ToString(values.GetValueN(driverConfig, "Driver", "IPAddress"))
-			internalIPAddress := convert.ToString(values.GetValueN(driverConfig, "Driver", "PrivateIPAddress"))
+	if config["cloud-provider-name"] == nil {
+		if _, ok := machine.Annotations[InternalAddressAnnotation]; ok {
+			appendToSlice(config, "node-ip", internalIPAddress)
```
snasovich, line 856: Are we actually ever appending a value to a slice in this case (and the one below) or always just setting a value?

snasovich (commented):
Couple comments and one question that we may or may not address depending on the timeline.
Otherwise LGTM as far as the code / basic logic goes - though we definitely need approval from @dramich or @ibuildthecloud so just leaving a generic comment.

ibuildthecloud (approved):

Issue #34546, opened 2021-08-30
Cannot upgrade imported rke2 cluster in rancher 2.6.0-rc9


rancher-max:
**Rancher Server Setup**
- Rancher version: v2.6.0-rc9
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Info:
         Cluster Type (RKE1, RKE2, k3s, EKS, etc): rke1
         Node Setup: 3 nodes all roles
         Version: v1.21.4
- self signed certs

**Information about the Cluster**
- Kubernetes version: v1.19.13+rke2r2
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Imported

**Describe the bug**

I cannot upgrade the cluster through the UI, and the v3 API has the `rke2Config` field empty. I'm not sure what to look for in the v1 api. See screenshots for what I see in the UI.

**To Reproduce**
 
1. Import an older rke2 version cluster (v1.19.13+rke2r2 for example)
2. Attempt to upgrade through the UI

**Result**
No option to upgrade version

**Expected Result**

Should have an option to upgrade the kubernetes version to values associated with KDM

**Screenshots**

<img width="1452" alt="image" src="https://user-images.githubusercontent.com/55997940/131381646-03485790-1a83-456d-9e59-7f32431fccef.png">

**Additional context**

davidnuzik:
I've also reproduced this; I did not see this issue on rc7.

v2.6.0-rc9 (single-node docker install of Rancher)
self-signed certs

rke2 v1.20.10+rke2r1
3 server nodes (all roles)
1 agent node (worker role)

The UI doesn't show any option to upgrade my 1.20.10 cluster.

rancher-max:
https://github.com/rancher/rancher/pull/34549

davidnuzik:
This works for me now.

v2.6-head `a97612b63` pulled at 8/30/21 2:38 pm Pacific
self-signed certs

rke2 v1.20.10+rke2r1
3 server nodes (all roles)

I now see in the UI the option to upgrade:
![image](https://user-images.githubusercontent.com/45179589/131410095-3028833c-5277-4c49-ba9d-08aaaa88541a.png)

I upgraded to v1.21.4+rke2r2 (the only option in the drop down). I used default options (don't drain and 1 cp concurrency, 1 worker concurrency)

[...]

PR #34549, opened 2021-08-30
Distinguish imported RKE2 clusters from provisioning clusters

thedadams:
A previous attempt was made to determine when a v3.Cluster upgrade
should be administered by management or provisioning controllers. That
change was not able to distinguish imported RKE2 clusters which should be
administered by the management controllers.

This change adds an annotation to the v3.Cluster object to better
distinguish these scenarios.

Issue:
https://github.com/rancher/rancher/issues/34546


nickgerace (dismissed):
Approved for general logic flow from reading your PR description.

thedadams:
Tested with imported, custom and machine provisioned clusters to ensure they all work as expected.

cjellick:
did you hit k3s as well as rke2?

thedadams:
@cjellick Yes, all 3 scenarios with both RKE2 and K3S work as expected. Imported clusters are upgraded via `k3sbasedupgrade` and custom and machine provisioned clusters are upgraded via the system-agent.

kinarashah (approved):

snasovich (approved):

Issue #34486, opened 2021-08-26
cluster-agent pod is not deleted from an imported cluster running rancher.


sowmyav27:
Imported 2 k3s clusters, one clean and one with rancher running.
Cattle-system ns is nore deleted after removal from UI on any of them.

After removal from rancher state is:
```
# clean k3s
~> kubectl get pods -n cattle-system
NAME                                    READY   STATUS    RESTARTS   AGE
cattle-cluster-agent-857c8b5694-prstq   1/1     Running   0          25m

# k3s with rancher
~> kubectl get pods -n cattle-system
NAME                                    READY   STATUS      RESTARTS   AGE
cattle-cluster-agent-5bd497d754-74pj8   1/1     Running     0          22m
rancher-64cb65cbf7-xt75l                1/1     Running     0          25m
helm-operation-j5jw2                    0/2     Completed   0          21m
helm-operation-nph7s                    0/2     Completed   0          20m
helm-operation-gj68f                    0/2     Completed   0          20m
rancher-webhook-64f6bf4d54-zrlmv        1/1     Running     0          20m
helm-operation-7smk2                    0/2     Completed   0          20m
```
Tested on v2.6-head (b7c440b)

@cmurphy 's comments
Worked with @kravciak to diagnose the issue, the rancher instance in question is running without `--no-cacerts` and with self-signed certificates, and seems to lose its connection with the cluster agent tunnel, see rancher logs:

```
<snipped>
2021/08/25 19:17:58 [INFO] Creating roleBinding User user-wxsdd Role project-owner
2021/08/25 19:17:58 [INFO] Creating roleBinding User user-wxsdd Role admin
2021/08/25 19:17:58 [INFO] Creating roleBinding User u-xgzgejhop6 Role project-member
2021/08/25 19:17:58 [INFO] Creating roleBinding User u-xgzgejhop6 Role edit
2021/08/25 19:18:17 [INFO] error in remotedialer server [400]: websocket: close 1006 (abnormal closure): unexpected EOF
W0825 19:18:17.978764      32 reflector.go:437] pkg/mod/github.com/rancher/client-go@v0.21.0-rancher.1/tools/cache/reflector.go:168: watch of *v1.ClusterRole ended with: an error on the server ("unable to decode an event from the watch stream: tunnel disconnect") has prevented the request from succeeding
[...]

PR #34557, opened 2021-08-30
Fix imported cluster cleanup race

cmurphy:
There is a race condition where the cluster object on the local cluster
is being deleted, but the cluster agent on the downstream cluster is not
yet cleaned up, and without this change the controller won't get a
kubeconfig for the deleted cluster and won't be able to create the
cleanup job. This change expands the exception made for RKE1 in the rest
config getter to any cluster with the `cluster-scoped-gc` lifecycle
annotation so that even if the cluster is being deleted we can still get
a kubeconfig for it to launch the cleanup job.

https://github.com/rancher/rancher/issues/34486

### DIFF
--- a/pkg/clustermanager/manager.go
+++ b/pkg/clustermanager/manager.go
@@ -254,9 +254,10 @@ func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Con
 		return nil, nil
 	}
 
-	// rke1 custom clusters need cleanup and the cluster delete is being paused by cluster-scoped-gc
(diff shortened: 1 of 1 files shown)


thedadams (approved):

dramich (commented):
This would allow all clusters to do this since all v3 clusters have that annotation. The other issue is we are going to start controllers that are in a deleting state which this used to protect against. In places where we need access to a downstream cluster and it's in a deleting state we should get the RestConfig through a different means. @cmurphy suggested `NewUserContext` so that will be investigated as a viable option.

snasovich:
We're not merging to `master` (at least for now), closed in favor of https://github.com/rancher/rancher/pull/34573

Issue #34561, opened 2021-08-30
[KEv2] Cluster agents on imported hosted provider clusters aren't cleaned up


cmurphy:
**Describe the bug**

Rancher skips cleaning up the cluster agent and cattle-system namespace on imported hosted provider clusters, because the driver gets set to e.g. `GKE` and not to `imported`:

https://github.com/rancher/rancher/blob/662e49eedecb4d3d4782e331a4cff07b37db1ca0/pkg/controllers/management/usercontrollers/controller.go#L74-L78

We should make sure that imported clusters on hosted providers are cleaned up in the same way that other imported clusters are.

**To Reproduce**
 
1. Create an AKS, EKS, or GKE cluster through the provider console.
2. Import the cluster in rancher
3. List the pods in the cluster, note the cattle-cluster-agent pod when the cluster is finished importing
4. Delete the cluster

**Result**

cattle-cluster-agent is still on the cluster

**Expected Result**

A cleanup job should be launched and the cattle-cluster-agent pod and cattle-system namespace should be deleted.

**Screenshots**

**Additional context**
Related to https://github.com/rancher/rancher/issues/34486

izaac:
Rancher version: `v2.6.0` 

Reproduced on this Rancher version. The cluster agents are left in the imported cluster after removing from Rancher.
I will proceed validating the fix on EKS/AKS/GKE

izaac:
Rancher version: `v2.6-head (09/09/2021)` _a6e7905_

I tried GKEv2 and EKSv2 imports, wait until the cluster was in Active state then proceeded to remove it from Rancher. 
Waited 5 minutes and the cluster-agent was still in the cluster along with the `cattle-system` namespace


```
k get po -n cattle-system
NAME                                    READY   STATUS    RESTARTS   AGE
cattle-cluster-agent-67457d5c95-fkmln   1/1     Running   0          10m
cattle-cluster-agent-67457d5c95-g697x   1/1     Running   0          10m

 k get ns
NAME                          STATUS   AGE
cattle-fleet-system           Active   11m
cattle-impersonation-system   Active   11m
cattle-system                 Active   11m
default                       Active   14m
kube-node-lease               Active   14m
kube-public                   Active   14m
kube-system                   Active   14m
local                         Active   10m
```

cmurphy:
I apologize, this issue won't actually be fully fixed until https://github.com/rancher/rancher/pull/34573 is merged, I will move this back to test when that happens.

[...]

PR #34562, opened 2021-08-30
Clean up imported hosted provider clusters

cmurphy:
Without this change, imported KEv2 clusters would skip cleaning up the
cluster agents upon cluster deletion. This was because the driver would
be set to e.g. "GKE" rather than "Imported". This change ensures the
lifecycle controller checks the cluster config for the "imported"
attribute.

https://github.com/rancher/rancher/issues/34561


rmweir (approved):
LGTM

thedadams (approved):

cbron (approved):

snasovich:
We're not merging to `master` (at least for now), closed in favor of https://github.com/rancher/rancher/pull/34574

PR #34573, opened 2021-08-31
Fix cluster cleanup race

cmurphy:
There is a race condition where the cluster object on the local cluster
is being deleted, but the cluster agent on the downstream cluster is not
yet cleaned up, and without this change the controller won't get a
kubeconfig for the deleted cluster and won't be able to create the
cleanup job.

To fix this for RKE1, an exception was made in ToRESTConfig to have it
return a kubeconfig even as the cluster was being deleted, as long as
the driver was RKE1. This doesn't help for the case where the driver is
"Imported" or "RKE2". It's also problematic because it means that
controllers will be unnecessarily started when the cluster is being
deleted.

This change moves the deletion check out of ToRESTConfig and into the
internal controller handling methods, and reworks the imported cluster
cleanup and node cleanup handlers to use clustermanager.ToRESTConfig and
config.NewUserContext, circumventing clustermanager.UserContext which
starts controllers. With the deletion check moved out of ToRESTConfig,
cleanup handlers can safely use it to get credentials for the cluster
and launch cleanup jobs.

https://github.com/rancher/rancher/issues/34486

### DIFF
--- a/pkg/clustermanager/manager.go
+++ b/pkg/clustermanager/manager.go
@@ -117,6 +117,9 @@ func (m *Manager) markUnavailable(clusterName string) {
 }
 
 func (m *Manager) start(ctx context.Context, cluster *v3.Cluster, controllers, clusterOwner bool) (*record, error) {
+	if cluster.DeletionTimestamp != nil {
+		return nil, nil
+	}
 	obj, ok := m.controllers.Load(cluster.UID)
 	if ok {
 		if !m.changed(obj.(*record), cluster, controllers, clusterOwner) {
@@ -251,12 +254,6 @@ func (m *Manager) doStart(rec *record, clusterOwner bool) (exit error) {
 
 func ToRESTConfig(cluster *v3.Cluster, context *config.ScaledContext) (*rest.Config, error) {
 	if cluster == nil {
-		return nil, nil
-	}
-
-	// rke1 custom clusters need cleanup and the cluster delete is being paused by cluster-scoped-gc
-	// allow a rest config since the nodes still exist and we need to start jobs on them for cleanup
-	if cluster.DeletionTimestamp != nil && cluster.Status.Driver != v32.ClusterDriverRKE {
 		return nil, nil
 	}
 
@@ -450,6 +447,8 @@ func (m *Manager) APIExtClient(apiContext *types.APIContext, storageContext type
 	return m.ScaledContext.APIExtClient, nil
 }
 
+// UserContext accepts a cluster name and returns a client for that cluster,
+// starting all controllers for that cluster in the process.
 func (m *Manager) UserContext(clusterName string) (*config.UserContext, error) {
 	cluster, err := m.clusterLister.Get("", clusterName)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


thedadams (approved):
As long as master PR gets merged, this looks good.

cmurphy:
PR to master was closed, this is now the main PR.

See Dan's comment from the original: https://github.com/rancher/rancher/pull/34557#pullrequestreview-743239686

snasovich:
@luthermonson @dramich , just a friendly reminder to review this PR today or tomorrow. Thanks!

pkg/clustermanager/manager.go
```diff
@@ -117,6 +117,9 @@ func (m *Manager) markUnavailable(clusterName string) {
 }
 
 func (m *Manager) start(ctx context.Context, cluster *v3.Cluster, controllers, clusterOwner bool) (*record, error) {
+	if cluster.DeletionTimestamp != nil {
```
dramich, line 120: I don't think we should add this here. `UserContext` is called from a lot of places and so far the only one that seemed to have an issue is for the cleanup which is addressed below. I am concerned this might cause other bugs and affect other flows we aren't aware of or testing.

pkg/clustermanager/manager.go
```diff
@@ -400,6 +397,10 @@ func VerifyIgnoreDNSName(caCertsPEM []byte) (func(rawCerts [][]byte, verifiedCha
 }
 
 func (m *Manager) toRecord(ctx context.Context, cluster *v3.Cluster) (*record, error) {
+	if cluster.DeletionTimestamp != nil {
```
dramich, line 400: Same comment, but this is only called from start so not really needed either way.

pkg/controllers/management/node/cleanup.go
```diff
@@ -36,7 +37,19 @@ func (m *Lifecycle) deleteV1Node(node *v3.Node) (runtime.Object, error) {
...
+	if err != nil {
+		return node, err
+	}
+	if kubeConfig == nil {
+		logrus.Debugf("could not get kubeconfig for cluster %s", cluster.Name)
+		return node, nil
+	}
+	userClient, err := config.NewUserContext(m.clusterManager.ScaledContext, *kubeConfig, cluster.Name)
 	if err != nil {
 		if kerror.IsNotFound(err) {
```
dramich, line 54: This should be moved up to the lister call

dramich (approved):

luthermonson (dismissed):

pkg/clustermanager/manager.go
cmurphy: I added this back, I think it's necessary to fix the issue of unnecessarily starting controllers since it was removed from `ToRESTConfig`. This should have the same behavior since before this it would hit the same check in `ToRESTConfig`. Please re-review.

pkg/clustermanager/manager.go
```diff
@@ -468,6 +467,20 @@ func (m *Manager) UserContext(clusterName string) (*config.UserContext, error) {
 	return record.cluster, nil
 }
 
+// UserContextFromCluster accepts a pointer to a Cluster and returns a client
+// for that cluster. It does not start any controllers.
+func (m *Manager) UserContextFromCluster(cluster *v3.Cluster) (*config.UserContext, error) {
```
snasovich, line 472: Kudos for adding documentation comments.
I'm thinking if we could update func name to better convey the fact it doesn't start controllers. There is not much difference between `UserContext` and `UserContextFromCluster`, especially considering both take some info from cluster (either name or pointer to complete Cluster object). `UserContextNoControllers` or something like this?

dramich (approved):

pkg/clustermanager/manager.go
```diff
@@ -117,6 +117,9 @@ func (m *Manager) markUnavailable(clusterName string) {
 }
 
 func (m *Manager) start(ctx context.Context, cluster *v3.Cluster, controllers, clusterOwner bool) (*record, error) {
+	if cluster.DeletionTimestamp != nil {
+		return nil, nil
+	}
```
dramich, line 122: We should remove this and just revert the change in `ToRESTConfig`, that gets us back to the old logic.

dramich: So remove the `&& cluster.Status.Driver != v32.ClusterDriverRKE` check and the comment

cmurphy: The timestamp check is in `ToRESTConfig` is the whole reason the imported cluster cleanup is failing. If I put it back there then I would have to write a new function to get a rest config for deleting clusters.

dramich: Yup, never mind

PR #34594, opened 2021-09-01
Ensure correct system-agent version used in tests

thedadams:
The provisioning tests should use the correct version of the
system-agent. This change includes a test that checks for the expected
variable in the install script to ensure the correct version is used.

Issue:

### DIFF
--- a/scripts/provisioning-tests
+++ b/scripts/provisioning-tests
@@ -3,11 +3,15 @@ set -e
 
 cd $(dirname $0)/..
 
+eval "$(grep '^ENV CATTLE_SYSTEM_AGENT' package/Dockerfile | awk '{print "export " $2 "=" $3}')"
+
 echo Starting rancher server
 mkdir -p /var/lib/rancher/k3s/agent/images
 grep PodTestImage ./tests/integration/pkg/defaults/defaults.go | cut -f2 -d'"' > /var/lib/rancher/k3s/agent/images/pull.txt
 grep MachineProvisionImage ./pkg/settings/setting.go | cut -f4 -d'"' >> /var/lib/rancher/k3s/agent/images/pull.txt
 touch /tmp/rancher.log
+mkdir -p /usr/share/rancher/ui/assets
+curl -sLf https://github.com/rancher/system-agent/releases/download/${CATTLE_SYSTEM_AGENT_VERSION}/rancher-system-agent-amd64 -o /usr/share/rancher/ui/assets/rancher-system-agent-amd64
 ./scripts/run >/tmp/rancher.log 2>&1 &
 PID=$!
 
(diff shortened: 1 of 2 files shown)


tests/integration/pkg/tests/custom/custom_test.go
```diff
@@ -11,8 +12,25 @@ import (
...
+		t.Fatal(err)
+	}
+	defer clients.Close()
+
+	setting, err := clients.Mgmt.Setting().Get("system-agent-version", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.False(t, setting.Value == "")
```
snasovich, line 30: You could use https://pkg.go.dev/github.com/stretchr/testify/assert#NotEmpty, however, the check below covers this as well I believe. Unless we're trying to account for case when both `os.Getenv("CATTLE_SYSTEM_AGENT_VERSION")` and `setting.Value` end up empty and it should fail the test.

snasovich (commented):
One minor comment to address. Once it's done, feel free to add a committer to review as rest looks good to me.

tests/integration/pkg/tests/custom/custom_test.go
thedadams: Yes, just want to make sure both things are not empty.

snasovich (approved):

scripts/provisioning-tests
```diff
@@ -3,11 +3,15 @@ set -e
...
 cd $(dirname $0)/..
 
+eval "$(grep '^ENV CATTLE_SYSTEM_AGENT' package/Dockerfile | awk '{print "export " $2 "=" $3}')"
+
 echo Starting rancher server
 mkdir -p /var/lib/rancher/k3s/agent/images
 grep PodTestImage ./tests/integration/pkg/defaults/defaults.go | cut -f2 -d'"' > /var/lib/rancher/k3s/agent/images/pull.txt
 grep MachineProvisionImage ./pkg/settings/setting.go | cut -f4 -d'"' >> /var/lib/rancher/k3s/agent/images/pull.txt
 touch /tmp/rancher.log
+mkdir -p /usr/share/rancher/ui/assets
```
cbron, line 13: whats point of this ?

scripts/provisioning-tests
```diff
@@ -3,11 +3,15 @@ set -e
...
+curl -sLf https://github.com/rancher/system-agent/releases/download/${CATTLE_SYSTEM_AGENT_VERSION}/rancher-system-agent-amd64 -o /usr/share/rancher/ui/assets/rancher-system-agent-amd64
```
cbron, line 14: whats point of throwing them in this dir ?

snasovich: That's where system-agent is downloaded from by the downstream nodes. When running locally I needed to download binary there too (well, actually needed to change that path here https://github.com/rancher/rancher/blob/4dd4e3760cda7077de944061fd7187910c1334bd/pkg/settings/setting.go#L81 since `/usr/share` is not easily modifiable on MacOS so changed it to `/usr/local/share/rancher/ui`)

cbron (approved):

PR #34602, opened 2021-09-01
Fixing secrets in automation. 

anupama2501:
Automation scrips have been failing due to a breaking change that was introduced from v2.6
1. Changed the env variable to "environment" and envFrom to "environmentFrom"
2. Verify for v2.5 and below releases, env and envfrom works
3. Verify from 2.6 "environment" and "environmentFrom" works

Jenkins pass log: https://jenkins.int.rancher.io/job/rancher_qa/job/rancher-freeform/100/
https://jenkins.int.rancher.io/job/rancher_qa/job/rancher-freeform/103/

### DIFF
--- a/tests/validation/requirements_v3api.txt
+++ b/tests/validation/requirements_v3api.txt
@@ -21,3 +21,4 @@ requests==2.22.0
 rsa==4.7.2
 websocket-client==0.57.0
 semver==2.13.0
+packaging==20.9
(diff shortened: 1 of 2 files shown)


izaac (approved):
LGTM
Also checked automation jobs passing secrets tests on `v2.5.9` and `v2.6.0`

PR #34606, opened 2021-09-01
[release/v2.5] Update rke to v1.2.12

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/34026


PennyScissors:
> Are we not upgrading to `v1.3.0` for 2.5.x?

AFAIK we don't upgrade rke like that, instead we have a given rke version per rancher version (2.4 -> 1.1, 2.5 -> 1.2, 2.6 -> 1.3).

kinarashah:
@PennyScissors That's correct, release/v1.2 is for 2.5.x.

kinarashah (approved):

Issue #34425, opened 2021-08-24
cluster not recoverable after adding base64 encoded ca.pem to `ca cert bundle` option when provisioning cluster (private registries -> advanced)


slickwarren:
**Rancher Server Setup**
- Rancher version:v2.6.0-rc6
- Installation option (Docker install/Helm Chart):docker
   - If Helm Chart, Kubernetes Info:
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):rke1
         Node Setup:single
         Version:v1.21.3
- Proxy/Cert Details:self signed certs

**Information about the Cluster**
- Kubernetes version:v1.21.3-rke2r2
- Cluster Type (Local/Downstream):downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):digital ocean

**Describe the bug**

when adding a mirror, there is an option for `ca cert bundle`. Upon adding this field and creating the cluster, rancher goes down with a panic in the logs `panic: runtime error: invalid memory address or nil pointer dereference [recovered]`

**To Reproduce**
 
* bring up a registry to use, which has necessary images for the setup. Use valid certs when creating the registry. base64-encode the ca.pem, we will use this when provisioning the cluster
* create a downstream cluster
  * when provisioning, go to the `private registries` section and enter the credentials _as a mirror_ (advanced option).  
  * copy your b64 encoded ca.pem into the `ca cert bundle` section
  * also enter all other relevant information (registry login info, cluster name, etc.)

**Result**
upon creating the cluster, rancher server will fail with this panic in the logs `panic: runtime error: invalid memory address or nil pointer dereference [recovered`
I also have the logs available in additional context section

**Expected Result**

even if this wasn't the correct way to mirror, it should not break all of rancher

**Screenshots**

**Additional context**

[panic_260rc6_34425.txt](https://github.com/rancher/rancher/files/7041954/panic_260rc6_34425.txt)

thedadams:
I don't believe the CA bundle expects a base64 encoded string. I believe pasting the cert file as-is is the way to go.

That didn't cause the problem, but just wanted to mention for future testing.

nickgerace:
This is available to test on v2.6.1-rc1

slickwarren:
hmm, maybe this is a UI issue? When I copy and paste the cert (not base64 encoded) I get this error in the UI:
[...]

PR #34618, opened 2021-09-02
Stop panic if registry CA bundle given without TLS

thedadams:
When provisioning an RKE2/K3S cluster with a private registry, the user
has the option to provide TLS and/or CA bundle. If the user did not
provide a TLS secret, but did provide a CA bundle, Rancher would panic.

This fixes that panic, and a user can now provide TLS secret and/or CA
bundle without issue.

Issue:
https://github.com/rancher/rancher/issues/34425

### DIFF
--- a/pkg/provisioningv2/rke2/planner/registry.go
+++ b/pkg/provisioningv2/rke2/planner/registry.go
@@ -34,20 +34,19 @@ func toRegistryConfig(secrets v1.SecretCache, runtime, namespace string, registr
 
 	for registryName, config := range registry.Configs {
 		registryConfig := &registryConfig{}
-		if config.TLSSecretName == "" && config.InsecureSkipVerify {
+		if config.InsecureSkipVerify || config.TLSSecretName != "" || len(config.CABundle) > 0 {
 			registryConfig.TLS = &tlsConfig{
-				InsecureSkipVerify: true,
+				InsecureSkipVerify: config.InsecureSkipVerify,
 			}
-		} else if config.TLSSecretName != "" {
+		}
+
+		if config.TLSSecretName != "" {
 			secret, err := secrets.Get(namespace, config.TLSSecretName)
 			if err != nil {
 				return nil, nil, err
 			}
 			if secret.Type != corev1.SecretTypeTLS {
 				return nil, nil, fmt.Errorf("secret [%s] must be of type [%s]", config.TLSSecretName, corev1.SecretTypeTLS)
-			}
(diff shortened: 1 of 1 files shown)


cmurphy (approved):

pkg/provisioningv2/rke2/planner/registry.go
```diff
@@ -34,21 +34,20 @@ func toRegistryConfig(secrets v1.SecretCache, runtime, namespace string, registr
...
-		if config.TLSSecretName == "" && config.InsecureSkipVerify {
+		if config.InsecureSkipVerify || config.TLSSecretName != "" || len(config.CABundle) > 0 {
 			registryConfig.TLS = &tlsConfig{
-				InsecureSkipVerify: true,
+				InsecureSkipVerify: config.InsecureSkipVerify,
 			}
-		} else if config.TLSSecretName != "" {
+		}
+
+		if config.TLSSecretName != "" {
```
brendarearden, line 43: Since this used to be an else if, is there any reason why the secret stuff should not be configured if config.InsecureSkipVerify is true?

pkg/provisioningv2/rke2/planner/registry.go
```diff
@@ -34,21 +34,20 @@ func toRegistryConfig(secrets v1.SecretCache, runtime, namespace string, registr
 
 	for registryName, config := range registry.Configs {
 		registryConfig := &registryConfig{}
-		if config.TLSSecretName == "" && config.InsecureSkipVerify {
+		if config.InsecureSkipVerify || config.TLSSecretName != "" || len(config.CABundle) > 0 {
```
brendarearden, line 37: does this need to be in an if check at all if its getting set for all conditions?

pkg/provisioningv2/rke2/planner/registry.go
thedadams: I don't think so. In the original logic, if `config.TLSSecretName != ""`, then it didn't matter what `config.InsecureSkipVerify` was: it was get used on old line 49.

pkg/provisioningv2/rke2/planner/registry.go
thedadams: We only want to set the `registryConfig.TLS` if it has any non-empty config because this gets passed to RKE2/K3S during provisioning. If there is an empty TLS config when RKE2/K3S is started, then command line args validation there will fail.

brendarearden (approved):
lgtm

dramich (approved):

PR #34619, opened 2021-09-02
airgap HA automation

slickwarren:
This includes the following:
* Airgap HA script requires the following vars:
  * RANCHER_HOST_NAME=(short prefix for ha cluster. becomes <RANCHER_HOST_NAME>.qa.rancher.space)
  * RANCHER_SERVER_VERSION=(rancher version. i.e. v2.5.8)
  * RANCHER_HA_CERT_OPTION=byo-valid -- currently only working option. Can submit amended PR to add other options later
* No Auth Private Registry script
  * this is mostly ported directly into the airgap HA script. I think the best option would be to point the airgap HA to this script instead, so it can be built + used on its own.


izaac (approved):
LGTM 🚀

tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -0,0 +1,516 @@
...
+    os.environ.get("RANCHER_BASTION_PASSWORD", ADMIN_PASSWORD)
+RANCHER_EXTERNAL_BASTION = os.environ.get("RANCHER_EXTERNAL_BASTION", "")
+RANCHER_EXTERNAL_HOST_NAME = os.environ.get(
+    "RANCHER_EXTERNAL_HOST_NAME", AG_HOST_NAME)
+REG_HOST_NAME = RANCHER_EXTERNAL_HOST_NAME + "-registry"
+REGISTRY_HOSTNAME = os.environ.get(
+    "REGISTRY_HOSTNAME", REG_HOST_NAME + ".qa.rancher.space")
+
+
+def test_deploy_airgap_rancher(check_hostname_length):
```
izaac, line 51: There's other test that has the same same in `test_airgap.py` we may change this test name to `test_deploy_airgap_rancher_ha`.

PR #34620, opened 2021-09-02
Fixing secrets in automation. 

anupama2501:
Automation scrips have been failing due to a breaking change that was introduced from v2.6

- Changed the env variable to "environment" and envFrom to "environmentFrom"
- Verify for v2.5 and below releases, env and envfrom works
- Verify from 2.6 "environment" and "environmentFrom" works

### DIFF
--- a/tests/validation/tests/v3_api/test_secrets.py
+++ b/tests/validation/tests/v3_api/test_secrets.py
@@ -1,13 +1,13 @@
 import base64
 from rancher import ApiError
 import pytest
+from packaging import version
 
 from .common import *  # NOQA
 
 CLUSTER_NAME = os.environ.get("CLUSTER_NAME", "")
 
 namespace = {"p_client": None, "ns": None, "cluster": None, "project": None}
-
 
 def test_secret_create_all_ns():
 
@@ -623,10 +623,31 @@ def create_project_client(request):
     namespace["project"] = p
     namespace["c_client"] = c_client
 
+
     def fin():
         client = get_user_client()
         client.delete(namespace["project"])
+
     request.addfinalizer(fin)
+
+
+def is_version_greater_than_v25():
+    # Checks if rancher version is greater than v2.5
+    # eg: v2.5-head, v2.5 etc.
+
+    rancher_version = get_setting_value_by_name('server-version')
+
+    version_check = False
+    if rancher_version.startswith('v'):
+        if "head" in rancher_version:
+            rancher_version = ''.join(rancher_version.split("-")[0])
+        else:
+            rancher_version = '.'.join(rancher_version.split(".")[:2])
+
+    if version.parse(rancher_version) > version.parse('v2.5'):
+        version_check = True
+
+    return version_check
 
 
 def validate_workload_with_secret(p_client, workload,
@@ -738,14 +759,27 @@ def create_and_validate_workload_with_secret_as_env_variable(p_client, secret,
     # Create Workload with secret as env variable
     secretName = secret['name']
 
-    environmentdata = [{
-        "source": "secret",
-        "sourceKey": None,
-        "sourceName": secretName
-    }]
+    # Checking if rancher version is 26 and above.
+    # If yes env should be "environment" and envFrom should be "environmentFrom"
+    rancher_version_26 = is_version_greater_than_v25()
+    if rancher_version_26:
+        env_str = "envFrom"
+        environment_data = [{
+            "secretRef": {
+                "name": secretName,
+                "optional": False
+            }
+        }]
+    else:
+        env_str = "environmentFrom"
+        environment_data = [{
+            "source": "secret",
+            "sourceKey": None,
+            "sourceName": secretName
+        }]
     con = [{"name": "test",
             "image": TEST_IMAGE,
-            "environmentFrom": environmentdata}]
+            env_str: environment_data }]
 
     workload = p_client.create_workload(name=name,
                                         containers=con,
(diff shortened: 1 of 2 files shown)


izaac (dismissed):
LGTM 🚀

slickwarren (approved):

tests/validation/tests/v3_api/test_secrets.py
```diff
@@ -623,12 +623,33 @@ def create_project_client(request):
...
 
 
+def rancher_version_check():
+    # Checks if rancher version is greater than v2.6
+    # eg: v2.5-head, v2.5 etc.
+
+    rancher_version = get_setting_value_by_name('server-version')
+
+    version_check = False
+    if rancher_version.startswith('v'):
```
git-ival, line 641: version.parse() should be able to handle version constructs that match [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-scheme). So only handling the "-head" case is necessary:
```
if "head" in rancher_version:
            rancher_version = ''.join(rancher_version.split("-")[0])
OR
# this is more explicit imo
if "head" in rancher_version:
            rancher_version = rancher_version.replace("-head")
```

tests/validation/tests/v3_api/test_secrets.py
```diff
@@ -623,12 +623,33 @@ def create_project_client(request):
...
+        if "head" in rancher_version:
+            rancher_version = ''.join(rancher_version.split("-")[0])
+        else:
+            rancher_version = '.'.join(rancher_version.split(".")[:2])
+
+    if version.parse(rancher_version) > version.parse('v2.5'):
+        version_check = True
+
+    return version_check
```
git-ival, line 650: We should probably just return whatever the version comparison results to:
```
return version.parse(rancher_version) > version.parse('v2.5')
```

tests/validation/tests/v3_api/test_secrets.py
```diff
@@ -623,12 +623,33 @@ def create_project_client(request):
...
+
     def fin():
         client = get_user_client()
         client.delete(namespace["project"])
+
     request.addfinalizer(fin)
 
 
+def rancher_version_check():
+    # Checks if rancher version is greater than v2.6
```
git-ival, line 635: This checks if the rancher version is > v2.5*

It might be useful to allow passing the version to compare against as a parameter? 
e.g. `def rancher_version_check(compare_version='v2.5'): `

That way we compare rancher_version vs compare_version instead of only being able to compare against "v2.5"

tests/validation/tests/v3_api/test_secrets.py
```diff
@@ -623,12 +623,33 @@ def create_project_client(request):
...
 
+
     def fin():
         client = get_user_client()
         client.delete(namespace["project"])
+
     request.addfinalizer(fin)
 
 
+def rancher_version_check():
```
git-ival, line 634: Renaming this function to something like `is_version_greater_than()` would be more explicit and easy to read imo

tests/validation/tests/v3_api/test_secrets.py
anupama2501: @git-ival replacing `"-head"` will result in an error as the v2.5-head versions have commit ID attached. 
eg:`v2.5-47e2c3e7324fb74c643ec877db0272ab500b5097-head` which gives the output as `v2.5-47e2c3e7324fb74c643ec877db0272ab500b5097`.

tests/validation/tests/v3_api/test_secrets.py
anupama2501: fix should return True for any patch versions if they are 2.6 and above eg., v2.6.0, v2.6.1 etc and should return False for any patch versions of 2.5 eg., v2.5.0, v2.5.1,v2.5.9 etc

Fix to change the parameter `environment` to `env` was introduced from v2.6.0 and above. 

Adding v2.5.10 will result in error if the version is v2.5.11 as it returns true and the expected result be False.

tests/validation/tests/v3_api/test_secrets.py
git-ival: That makes sense, but do we need the top-level `if rancher_version.startswith('v'):` ?

tests/validation/tests/v3_api/test_secrets.py
anupama2501: this might not be required as it the comparison is with 'v.5' only

tests/validation/tests/v3_api/test_secrets.py
anupama2501: I added the startswith('v') as we have master-head as well as one of the versions. But now that I think about it, since the master-head version will not be covered anymore, we can remove the startwith('v') validation.

git-ival: Hm, I think that `version.parse()` would error on "master-head" since it doesn't match the expected pattern. If we ever do have to test against master-head we should explicitly check for it and return `True` if so.

anupama2501: I think its ok to have the extra validation for the rancher version (startswith('v'). If we ever come back to master it will not break and no additional changes will be needed.

For```Hm, I think that version.parse() would error on "master-head" ```
I removed it from the previous feedback from Brandon as currently we won't be using it and might switch back to `main` or `master` in future.

git-ival:
LGTM!

git-ival (approved):

bmdepesa:
LGTM

bmdepesa (approved):

Issue #34161, opened 2021-08-11
Rancher 2.6 - updating/upgrading  aks cluster


claudioser:
**Rancher Server Setup**
- Rancher version: Rancher 2.6 rc2
- Installation option (Docker install/Helm Chart): Helm on RKE1
   
**Information about the Cluster**
- Kubernetes version: 
- Cluster Type (Local/Downstream): Downstream 
   - AKS

**Describe the bug**
when trying to upgrade from 1 version ok aks to another, I am getting this error, even though all Im actually modifying on the ui is the actual version desired ,
![image](https://user-images.githubusercontent.com/53047113/128998277-ce0378a1-8c69-4224-8158-6f1de6485962.png)

Furthermore the upgrade never starts or finishes. If I try to edit the cluster again, It thinks its already 1.19.11, even though nothing happens . Error message stays on UI, but cluster functionality is working.

![image](https://user-images.githubusercontent.com/53047113/128998596-5a6c1faa-ccc6-4bf3-b7a7-7f9b795e2e97.png)

**To Reproduce**
 happens on both aks clusters provisioned from rancher or existing aks clusters imported to rancher

**Additional context**
 on a probable similar context, when upgrading get  options for upgrading up to 1.21.2 from UI , but on Azure - AKS, only have up to 1.20.7

![image](https://user-images.githubusercontent.com/53047113/128999094-188a9b54-f645-437e-a7e9-9f4627a0329e.png)

![image](https://user-images.githubusercontent.com/53047113/128999120-9db09277-7753-4836-bfc0-556ba0843ce3.png)

prabalsharma:
**Issue1:** 
I could not repro the issue. I tried it on 2.6-rc4. RKE_VERSION=v1.3.0-rc15
I was able to upgrade clusters k8s version from 1.19.9 to 1.19.11

**Issue2:**
Its a UI issue. Not sure how extra versions are coming up. AFAIK version is is being pulled from Azure. @codyrancher any idea?

codyrancher:
@prabalsharma this looks to be a backend issue. All of the versions in the screenshot are being returned by aksVersions.

![image](https://user-images.githubusercontent.com/55104481/130155066-60433e40-410d-41fe-8d0e-cd4e7d3f8768.png)

prabalsharma:
Thanks Cody.

paynejacob:
waiting on https://github.com/rancher/charts/pull/1421

prabalsharma:
[...]

PR #34624, opened 2021-09-02
Add support for listing update aks versions

paynejacob:
https://github.com/rancher/rancher/issues/34161

Problem:
In AKS you cannot skip minor version updates.  Currently we list all available versions in the upgrade drop down.  This lets users attempt to upgrade to version that is too high causing the cluster to go into a constant error state.

Solution:
 I added the `current_version` parameter to the `/meta/aksVersions` endpoint.  When set the endpoint will only return versions available for upgrade from the current version.  There will need to be a companion dashboard issue for this.  The available versions are derived from the AKS API results.
 
 Testing:
 
 ```bash
 $ curl -s --insecure -H "Authorization: bearer $RANCHER_KEY" "https://localhost:8443/meta/aksVersions?cloudCredentialId=cattle-global-data%3Acc-6dgcx&region=eastus" | jq
[
  "1.18.17",
  "1.18.19",
  "1.19.11",
  "1.19.13",
  "1.20.7",
  "1.20.9",
  "1.21.1",
  "1.21.2"
]
 ```
 
 ```bash
$ curl -s --insecure -H "Authorization: bearer $RANCHER_KEY" "https://localhost:8443/meta/aksVersions?cloudCredentialId=cattle-global-data%3Acc-6dgcx&region=eastus&current_version=1.18.17" | jq
[
  "1.18.19",
  "1.19.11",
  "1.19.13"
]

 ```

### DIFF
--- a/pkg/api/norman/customization/aks/handler.go
+++ b/pkg/api/norman/customization/aks/handler.go
@@ -40,6 +40,7 @@ type Capabilities struct {
 	ClientID         string `json:"clientId"`
 	ClientSecret     string `json:"clientSecret"`
 	ResourceLocation string `json:"region"`
+	CurrentVersion   string `json:"currentVersion"`
 }
 
 // AKS handler lists available resources in Azure API
@@ -239,10 +240,8 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 	if cap.AuthBaseURL == "" {
 		cap.AuthBaseURL = azure.PublicCloud.ActiveDirectoryEndpoint
 	}
-	region := req.URL.Query().Get("region")
-	if region != "" {
-		cap.ResourceLocation = region
-	}
+	cap.ResourceLocation = req.URL.Query().Get("region")
+	cap.CurrentVersion = req.URL.Query().Get("currentVersion")
 
 	return http.StatusOK, nil
 }
--- a/pkg/api/norman/customization/aks/listers.go
+++ b/pkg/api/norman/customization/aks/listers.go
@@ -141,13 +141,24 @@ func listKubernetesVersions(ctx context.Context, cap *Capabilities) ([]byte, int
 
 	var kubernetesVersions []string
 
-	for _, profile := range *orchestrators.Orchestrators {
-		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
-			return nil, http.StatusInternalServerError, fmt.Errorf("unexpected nil orchestrator type or version")
+	// If listing for an upgrade only list the versions available for the current version
+	if cap.CurrentVersion != "" {
+		// find the current version's profile
+		for _, profile := range *orchestrators.Orchestrators {
+			if *profile.OrchestratorType == "Kubernetes" && *profile.OrchestratorVersion == cap.CurrentVersion {
+				// get available upgrade versions for the current version
+				kubernetesVersions, err = kubernetesVersionsFromOrchestratorProfiles(*profile.Upgrades...)
+				if err != nil {
+					return nil, http.StatusInternalServerError, err
+				}
+				break
+			}
 		}
-
-		if *profile.OrchestratorType == "Kubernetes" {
-			kubernetesVersions = append(kubernetesVersions, *profile.OrchestratorVersion)
+	} else {
+		// list all available versions
+		kubernetesVersions, err = kubernetesVersionsFromOrchestratorVersionProfiles(*orchestrators.Orchestrators...)
+		if err != nil {
+			return nil, http.StatusInternalServerError, err
 		}
 	}
 
@@ -325,4 +336,36 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
 	}
 
 	return data, http.StatusOK, err
+}
+
+func kubernetesVersionsFromOrchestratorVersionProfiles(profiles ...containerservice.OrchestratorVersionProfile) ([]string, error) {
+	var kubernetesVersions []string
+
+	for _, profile := range profiles {
+		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
+			return kubernetesVersions, fmt.Errorf("unexpected nil orchestrator type or version")
+		}
+
+		if *profile.OrchestratorType == "Kubernetes" {
+			kubernetesVersions = append(kubernetesVersions, *profile.OrchestratorVersion)
+		}
+	}
+
+	return kubernetesVersions, nil
+}
+
+func kubernetesVersionsFromOrchestratorProfiles(profiles ...containerservice.OrchestratorProfile) ([]string, error) {
+	var kubernetesVersions []string
+
+	for _, profile := range profiles {
+		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
+			return kubernetesVersions, fmt.Errorf("unexpected nil orchestrator type or version")
+		}
+
+		if *profile.OrchestratorType == "Kubernetes" {
+			kubernetesVersions = append(kubernetesVersions, *profile.OrchestratorVersion)
+		}
+	}
+
+	return kubernetesVersions, nil
 }


pkg/api/norman/customization/aks/handler.go
```diff
@@ -40,6 +40,7 @@ type Capabilities struct {
 	ClientID         string `json:"clientId"`
 	ClientSecret     string `json:"clientSecret"`
 	ResourceLocation string `json:"region"`
+	CurrentVersion   string `json:"current_version"`
```
cmurphy, line 43: This should probably be camel case to be consistent with the other parameters (e.g. `clientSecret`)

cmurphy (changes requested):
Looks great to me, one minor consistency issue

paynejacob:
https://github.com/rancher/dashboard/issues/4103

cmurphy (approved):

pkg/api/norman/customization/aks/listers.go
```diff
@@ -326,3 +337,35 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
 
 	return data, http.StatusOK, err
 }
+
+func KubernetesVersionsFromOrchestratorVersionProfiles(profiles ...containerservice.OrchestratorVersionProfile) ([]string, error) {
```
snasovich, line 341: I think these newly introduced funcs should be unexported, especially considering the fact they are called from unexported func. Also, generally, we want exported funcs to be followed by unexported ones (https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering).

Bottom line, if we make these unexported we'll be good.

P.S. Totally unrelated but I wish there was a (reasonable) way to eliminate copy-paste of these 2 funcs (as minimal as it is).

pkg/api/norman/customization/aks/handler.go
```diff
@@ -243,6 +244,10 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 	if region != "" {
 		cap.ResourceLocation = region
 	}
+	currentVersion := req.URL.Query().Get("currentVersion")
```
snasovich, line 247: I know we're following the exact same way `region` is assigned, but really what's the point of `if` check?
```go
cap.CurrentVersion = req.URL.Query().Get("currentVersion")
```
This should be perfectly fine. Even if we consider a hypothetical case where a pointer to `Capabilities` struct with some set fields is passed, I would imagine we would want `CurrentVersion` to be set to empty if `req` doesn't have  `currentVersion` query parameter passed.

Oh, and I just noticed the check on line below checks `region` instead of `currentVersion`. :)

snasovich (changes requested):
Have a couple of comments to address, thank you

pkg/api/norman/customization/aks/listers.go
paynejacob: I agree with unexporting them.  I found that in order to make this one function the type assertion logic made it difficult to read and it just made more sense to duplicate them.

pkg/api/norman/customization/aks/handler.go
paynejacob: I agree there is no need to do this assignment I was just following the pattern above.

pkg/api/norman/customization/aks/listers.go
snasovich: Yup, totally agree. My comment was more about the inability to define an interface for fields in Go, only for funcs. This would have helped in this case.

pkg/api/norman/customization/aks/handler.go
snasovich: 👍  Feel free to follow the pattern if you see some value in it, but if you do we need to have the `if` check fixed.

cmurphy: Oops, I had thought we had gotten rid of these unnecessary empty checks in earlier revisions of this code. I don't think there's value in following the pattern, direct assignment should be fine.

pkg/api/norman/customization/aks/handler.go
```diff
@@ -243,6 +244,10 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
...
+	if region != "" {
```
cmurphy, line 248: missed this

cmurphy (changes requested):

pkg/api/norman/customization/aks/listers.go
paynejacob: Generics are coming

snasovich (approved):

cmurphy (approved):

vincent99:
Did anybody talk to anybody in UI about this?  I'd want to continue to _show_ all the versions, but _disable_ the ones that are too high and tell you why they're disabled.  That way you know in advance there are further upgrades you can do after this one instead of being in the dark.

So it would be preferable to have one call that returns both those pieces of info together instead of making two separate ones and correlating them.

But you can't really do that with the current API, so that's a breaking change, a new endpoint, or 2 round trips. (this is why the top-level of an API response should never be an array; you have nowhere to add more stuff later).

I see it's implemented here by reading the list from somewhere else, but is the rule really as deterministic as "same or +1 semver minor"?  The UI could have just do the filtering on that from the existing list.

snasovich:
@vincent99 , we did not. Thank you for commenting and good point about it being better UX if we show all versions and disable ones that cannot be upgraded to with note on why.

Given the points you've brought up, would going with 2 separate calls be acceptable? I know it's not ideal.

As for the "rule" comment, we probably don't want to duplicate this business logic in our API/UI and "ask" AKS what versions we can upgrade to from a given one - which is what this change does.
Speaking of showing the reason certain version cannot be upgraded to in the UI, I'm wondering if we can have wording that states something like "You cannot upgrade to this version. Most likely, you need to upgrade to an earlier version first." so we're not stating the reason with 100% certainty but simply provide a possible reason.

paynejacob:
@vincent99 I did not reach out to UI before starting this, in the future I will do that first.  With this pr I see a few options going forward.

1. abandon this endpoint and make a new one with top level keys so we can add richer data without breaking the api. e.g. all versions and their upgrades instead of a simple version list.
2. we use this solution and the ui makes 2 calls.
3. The ui does frontend validation for the upgradeable versions  and there is no backend change.

I would favor 1 or 2.  My concern with option 3 is this restriction is not clearly documented by MS, it is a note in their cli docs, and I am concerned there are other restrictions that we will miss.  If they are going to give us a list of what versions we can upgrade to I would rather we just ask them instead of trying to reproduce restrictions that may change in the future.

vincent99:
We can make # 2 work if you really want, but I'd prefer # 1 with something like

```
GET /meta/aksUpgrades?clusterId=c-xxxxx

{
  currentVersion: '1.18.9',
  upgrades: [
    {versions: "1.18.9", enabled: true},
    {version: "1.18.10", enabled: true},
    {version: "1.19.20", enabled: true},
    {version: "1.20.0", enabled: false, reason: 'no_skip'},
  ]
}
```

This:
- Makes the top level a map so additional info can be added later if needed.
- Makes the answer depend directly on the cluster that we're asking for, rather than indirectly via `region` & `currentVersion`.
  - If you find some new reason why upgrade eligibility would vary, you can handle that without changing the API interface again.
- Directly provides a single list of the versions to show that you control.
  - The UI doesn't have to make two calls or decide how to correlate them / what to do in edge cases like if a version comes back in the "upgradeable" list but not the "all" list.
  - The `no_skip` or whatever would be mapped to a translation string that ends up saying something like "v1.20.0 -- Upgrade to a lower version first". in the dropdown.
  - You providing a 'reason' allows a place for there to be other reasons in the future (maybe you know that a 1.22 alpha exists but the user hasn't set the experimental flag or something, who knows).
- (Should only work for clusters the user asking can list)

pkg/api/norman/customization/aks/listers.go
```diff
@@ -326,3 +337,35 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
...
+
+func kubernetesVersionsFromOrchestratorVersionProfiles(profiles ...containerservice.OrchestratorVersionProfile) ([]string, error) {
+	var kubernetesVersions []string
+
+	for _, profile := range profiles {
+		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
+			return kubernetesVersions, fmt.Errorf("unexpected nil orchestrator type or version")
+		}
+
+		if *profile.OrchestratorType == "Kubernetes" {
```
rmweir, line 349: nit: use containerService.Kubernetes constant here.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -326,3 +337,35 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
...
+
+func kubernetesVersionsFromOrchestratorProfiles(profiles ...containerservice.OrchestratorProfile) ([]string, error) {
+	var kubernetesVersions []string
+
+	for _, profile := range profiles {
+		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
+			return kubernetesVersions, fmt.Errorf("unexpected nil orchestrator type or version")
+		}
+
+		if *profile.OrchestratorType == "Kubernetes" {
```
rmweir, line 365: nit: use containerService.Kubernetes constant here.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -141,13 +141,24 @@ func listKubernetesVersions(ctx context.Context, cap *Capabilities) ([]byte, int
...
 	var kubernetesVersions []string
 
-	for _, profile := range *orchestrators.Orchestrators {
-		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
-			return nil, http.StatusInternalServerError, fmt.Errorf("unexpected nil orchestrator type or version")
+	// If listing for an upgrade only list the versions available for the current version
+	if cap.CurrentVersion != "" {
+		// find the current version's profile
+		for _, profile := range *orchestrators.Orchestrators {
+			if *profile.OrchestratorType == "Kubernetes" && *profile.OrchestratorVersion == cap.CurrentVersion {
```
rmweir, line 148: nit: use containerService.Kubernetes constant here.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -326,3 +337,35 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
...
+	}
+
+	return kubernetesVersions, nil
+}
+
+func kubernetesVersionsFromOrchestratorProfiles(profiles ...containerservice.OrchestratorProfile) ([]string, error) {
+	var kubernetesVersions []string
+
+	for _, profile := range profiles {
+		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
```
rmweir, line 361: Is the nil check and error return necessary here? Should we just skip over it if this is the case? Could add a debug log if information could be useful.

rmweir (changes requested):

paynejacob:
@vincent99 that makes sense I think that endpoint is cleaner and we can leave this one as a raw list and deprecate it in the future.  I ll get a new pr up for the new endpoint.

PR #34646, opened 2021-09-03
Fix for docker server deployments: retry password requests

slickwarren:
I've added the argument `-e CATTLE_BOOTSTRAP_PASSWORD=<password>` for each docker run in our automation. 
Also, it turns out that the request just needed to be retried. Jenkins job is passing

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -2084,15 +2084,29 @@ def readDataFile(data_dir, name):
         return f.read()
 
 
-def set_url_password_token(rancher_url, server_url=None):
+def set_url_password_token(rancher_url, server_url=None, version=""):
     """Returns a ManagementContext for the default global admin user."""
     auth_url = \
         rancher_url + "/v3-public/localproviders/local?action=login"
-    r = requests.post(auth_url, json={
-        'username': 'admin',
-        'password': 'admin',
-        'responseType': 'json',
-    }, verify=False)
+    rpassword = 'admin'
+    print(auth_url)
+    if version.find("master") > -1 or version.find("2.6") > -1:
+        rpassword = ADMIN_PASSWORD
+        print("on 2.6 or later")
+    retries = 5
+    for attempt in range(1,retries):
+        try:
+            r = requests.post(auth_url, json={
+            'username': 'admin',
+            'password': rpassword,
+            'responseType': 'json',
+            }, verify=False)
+        except requests.exceptions.RequestException:
+            print("password request failed. Retry attempt: ",
+                  "{} of {}".format(attempt, retries))
+            time.sleep(2)
+        else:
+            break
     print(r.json())
     token = r.json()['token']
     print(token)
--- a/tests/validation/tests/v3_api/test_airgap.py
+++ b/tests/validation/tests/v3_api/test_airgap.py
@@ -558,16 +558,20 @@ def deploy_airgap_rancher(bastion_node):
             '-v ${{PWD}}/privkey.pem:/etc/rancher/ssl/key.pem ' \
             '-e CATTLE_SYSTEM_DEFAULT_REGISTRY={} ' \
             '-e CATTLE_SYSTEM_CATALOG=bundled ' \
+            '-e CATTLE_BOOTSTRAP_PASSWORD={} ' \
             '{}/rancher/rancher:{} --no-cacerts --trace'.format(
                 privileged, bastion_node.host_name, bastion_node.host_name,
-                RANCHER_SERVER_VERSION)
+                ADMIN_PASSWORD, RANCHER_SERVER_VERSION)
     else:
         deploy_rancher_command = \
             'sudo docker run -d {} --restart=unless-stopped ' \
             '-p 80:80 -p 443:443 ' \
             '-e CATTLE_SYSTEM_DEFAULT_REGISTRY={} ' \
+            '-e CATTLE_BOOTSTRAP_PASSWORD={} ' \
             '-e CATTLE_SYSTEM_CATALOG=bundled {}/rancher/rancher:{} --trace'.format(
-                privileged, bastion_node.host_name, bastion_node.host_name,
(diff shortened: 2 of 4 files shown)


tests/validation/tests/v3_api/common.py
```diff
@@ -2084,15 +2084,28 @@ def readDataFile(data_dir, name):
...
-        'password': 'admin',
-        'responseType': 'json',
-    }, verify=False)
+    rpassword = 'admin'
+    print(auth_url)
+    if version.find("master") > -1 or version.find("2.6") > -1:
+        rpassword = ADMIN_PASSWORD
+        print("on 2.6 or later")
+    retries = 4
+    for attempt in range(retries):
```
izaac, line 2097: `password request failed. Retry attempt:  0 of 4`
Nit but it would be more readable if we use `range(1, retries)` with `retries = 5`

We have a `wait_for` in `tests/common` wouldn't that be better to use it?

izaac (changes requested):
Looking good overall, but just a comment on the password retries

tests/validation/tests/v3_api/common.py
```diff
@@ -2084,15 +2084,28 @@ def readDataFile(data_dir, name):
...
+        try:
+            r = requests.post(auth_url, json={
+            'username': 'admin',
+            'password': rpassword,
+            'responseType': 'json',
+            }, verify=False)
+            break
+        except:
```
dramich, line 2105: A naked `except` is a bad pattern, this should name the error expected so if it isn't of that type the exception actually happens. i.e. `except ApiError:`

tests/validation/tests/v3_api/common.py
slickwarren: from what I saw, the `wait_for` function wouldn't handle the error being thrown, it was only for waiting for the callback to be available. I did update the range in the latest version

tests/validation/tests/v3_api/common.py
slickwarren: I've updated this to `except OSError` as this was the original error being thrown. rerun of the jenkins job with the updated code is passing

tests/validation/tests/v3_api/common.py
izaac: You can handle the error in the callback and return `False/None` or a `True` depending on the request result.

tests/validation/tests/v3_api/common.py
izaac: `OSError` is not quite specific for the `requests` library `ConnectionError` would be more appropriate.

tests/validation/tests/v3_api/common.py
```diff
@@ -2084,15 +2084,28 @@ def readDataFile(data_dir, name):
...
+        print("on 2.6 or later")
+    retries = 5
+    for attempt in range(1,retries):
+        try:
+            r = requests.post(auth_url, json={
+            'username': 'admin',
+            'password': rpassword,
+            'responseType': 'json',
+            }, verify=False)
+            break
```
izaac, line 2104: We can move this break to an `else` in the `try` block. That would be more readable.

```python
try:
    r = requests.post(....
except ConnectionError as, ce:
    print(ce.message)
else:
    break
```

izaac (changes requested):
A couple comments

tests/validation/tests/v3_api/common.py
```diff
@@ -2084,15 +2084,29 @@ def readDataFile(data_dir, name):
...
+        print("on 2.6 or later")
+    retries = 5
+    for attempt in range(1,retries):
+        try:
+            r = requests.post(auth_url, json={
+            'username': 'admin',
+            'password': rpassword,
+            'responseType': 'json',
+            }, verify=False)
+        except ConnectionError:
```
git-ival, line 2104: `requests.exceptions.RequestException` is the catchall exception if thats the intent

tests/validation/tests/v3_api/common.py
```diff
@@ -2084,15 +2084,29 @@ def readDataFile(data_dir, name):
...
     auth_url = \
         rancher_url + "/v3-public/localproviders/local?action=login"
-    r = requests.post(auth_url, json={
-        'username': 'admin',
-        'password': 'admin',
-        'responseType': 'json',
-    }, verify=False)
+    rpassword = 'admin'
+    print(auth_url)
+    if version.find("master") > -1 or version.find("2.6") > -1:
```
git-ival, line 2093: Would be good to use `rancher_version_check()` from [this open PR](https://github.com/rancher/rancher/pull/34620) when comparing rancher versions.

tests/validation/tests/v3_api/common.py
```diff
@@ -2084,15 +2084,29 @@ def readDataFile(data_dir, name):
         return f.read()
 
 
-def set_url_password_token(rancher_url, server_url=None):
+def set_url_password_token(rancher_url, server_url=None, version="oldUI"):
```
git-ival, line 2087: Why are we setting version to "oldUI" by default? We may not need this argument at all if we just reference `RANCHER_SERVER_VERSION` directly. (Since we probably want this behavior for all rancher versions >= 2.6.0)

slickwarren: I took out the text of `oldUI` as that isn't important, but RANCHER_SERVER_VERSION isn't available in common.py.

tests/validation/tests/v3_api/common.py
slickwarren: thank you! Updated the latest with this exception.

izaac (approved):
LGTM 🚀

tests/validation/tests/v3_api/common.py
git-ival: Ah I didn't realize we only defined that in `test_custom_host_reg.py`, LGTM then. Though I wonder if we should move `RANCHER_SERVER_VERSION` into common.py

bmdepesa (approved):
LGTM

PR #34667, opened 2021-09-07
[QA-Tests] Add and enhance airgap automation for k3s and rke2

rancher-max:
- Split rke2 and k3s airgap into separate files
- Cleanup helper functions
- Ensure the following airgap scenarios are covered for both:
  - private registry (registries.yaml)
  - system-default-registry (flag that can be passed to use a TLS-secured registry but with no username & password)
  - tarball, with the ability to pass in different types of tarballs for .tar, .gz, and .zst as desired.
  - Additional flags for any extra customization (such as `--airgap-extra-registry` or using different cni's with `--cni=calico`)

### DIFF
--- a/tests/validation/tests/v3_api/scripts/configure.sh
+++ b/tests/validation/tests/v3_api/scripts/configure.sh
@@ -5,7 +5,7 @@ set -eu
 
 DEBUG="${DEBUG:-false}"
 
-env | egrep '^(ARM_|CATTLE_|ADMIN|USER|DO|RANCHER_|AWS_|DEBUG|LOGLEVEL|DEFAULT_|OS_|DOCKER_|CLOUD_|KUBE|BUILD_NUMBER|RKE_|AZURE).*\=.+' | sort > .env
+env | egrep '^(ARM_|CATTLE_|ADMIN|USER|DO|RANCHER_|K3S_|RKE2_|AWS_|DEBUG|LOGLEVEL|DEFAULT_|OS_|DOCKER_|CLOUD_|KUBE|BUILD_NUMBER|RKE_|AZURE).*\=.+' | sort > .env
 
 if [ "false" != "${DEBUG}" ]; then
     cat .env
--- a/tests/validation/tests/v3_api/test_airgap.py
+++ b/tests/validation/tests/v3_api/test_airgap.py
@@ -7,7 +7,7 @@ from lib.aws import AWS_USER
 from .common import (
     ADMIN_PASSWORD, AmazonWebServices, run_command, wait_for_status_code,
     TEST_IMAGE, TEST_IMAGE_NGINX, TEST_IMAGE_OS_BASE, readDataFile,
-    DEFAULT_CLUSTER_STATE_TIMEOUT
+    DEFAULT_CLUSTER_STATE_TIMEOUT, compare_versions
 )
 from .test_custom_host_reg import (
     random_test_name, RANCHER_SERVER_VERSION, HOST_NAME, AGENT_REG_CMD
@@ -16,7 +16,6 @@ from .test_create_ha import (
     set_url_and_password,
     RANCHER_HA_CERT_OPTION, RANCHER_VALID_TLS_CERT, RANCHER_VALID_TLS_KEY
 )
-from .test_import_k3s_cluster import (RANCHER_K3S_VERSION)
 
 PRIVATE_REGISTRY_USERNAME = os.environ.get("RANCHER_BASTION_USERNAME")
 PRIVATE_REGISTRY_PASSWORD = \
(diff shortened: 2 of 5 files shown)


tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -25,7 +24,7 @@
 NUMBER_OF_INSTANCES = int(os.environ.get("RANCHER_AIRGAP_INSTANCE_COUNT", "1"))
 IMAGE_LIST = os.environ.get("RANCHER_IMAGE_LIST", ",".join(
     [TEST_IMAGE, TEST_IMAGE_NGINX, TEST_IMAGE_OS_BASE])).split(",")
-ARCH = os.environ.get("RANCHER_ARCH", "amd64")
+TARBALL_TYPE = os.environ.get("RANCHER_TARBALL_TYPE", "tar.gz")
```
bmdepesa, line 27: Does the `RANCHER_` prefix really fit? Would it be better to change to `K3S_` or `RKE2_` or some other common prefix?

tests/validation/tests/v3_api/test_k3s_airgap.py
```diff
@@ -0,0 +1,216 @@
...
+                          deploy_noauth_bastion_server, copy_certs_to_node,
+                          trust_certs_on_node, add_tarball_to_node)
+
+ARCH = os.environ.get("RANCHER_ARCH", "amd64")
+RANCHER_K3S_VERSION = os.environ.get("RANCHER_K3S_VERSION", "")
+K3S_SERVER_OPTIONS = os.environ.get("RANCHER_K3S_SERVER_OPTIONS", "")
+K3S_AGENT_OPTIONS = os.environ.get("RANCHER_K3S_AGENT_OPTIONS", "")
+
+
+def test_deploy_airgap_k3s_system_default_registry():
```
bmdepesa, line 19: There's lots of re-used code between these tests? Can we parameterize the test with the install method and use common logic between them?

bmdepesa (changes requested):

tests/validation/tests/v3_api/test_airgap.py
rancher-max: We have historically used the `RANCHER_` prefix for k3s and rke2 things. I do think it would be good to update to K3S/RKE2, so I'm making that change for the variables here, but we'll need to refactor quite a lot of code or just make sure future k3s/rke2 automation uses appropriately named variables.

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -441,104 +375,181 @@ def prepare_airgap_node(bastion_node, number_of_nodes):
...
+        'scp -i "{0}.pem" -o StrictHostKeyChecking=no certs/* ' \
+        '{1}@{2}:~/'.format(bastion_node.ssh_key_name, AWS_USER,
+                            ag_node.private_ip_address)
+    bastion_node.execute_command(ag_node_copy_certs)
+
+
+def trust_certs_on_node(bastion_node, ag_node):
+    ag_node_update_certs = \
+        'sudo cp domain.crt ' \
+        '/usr/local/share/ca-certificates/domain.crt && ' \
```
izaac, line 402: I know k3s and rke2 support other linux distros than Ubuntu would this work on RHEL based linuxes?

tests/validation/tests/v3_api/scripts/configure.sh
```diff
@@ -5,7 +5,7 @@ set -eu
 
 DEBUG="${DEBUG:-false}"
 
-env | egrep '^(ARM_|CATTLE_|ADMIN|USER|DO|RANCHER_|AWS_|DEBUG|LOGLEVEL|DEFAULT_|OS_|DOCKER_|CLOUD_|KUBE|BUILD_NUMBER|RKE_|AZURE).*\=.+' | sort > .env
```
izaac, line 8: Any reason behind removing `RKE_`?

izaac (changes requested):
The variables refactoring and encapsulation to reuse code looks good now. 

Just a couple of comments one related to the environment configuration and the other OS specific certificates installs.

tests/validation/tests/v3_api/scripts/configure.sh
rancher-max: Oops.. conflict mistake. I'll fix that

tests/validation/tests/v3_api/test_airgap.py
rancher-max: No this shouldn't work on other ones, but I also haven't gone through the steps to confirm what exactly to do when using this `system-default-registry` airgap method on other distros. 

Maybe we should caveat this by saying it only has been validated against Ubuntu right now and next enhancement should be to support other distros 🤔

izaac: I'm fine if anyone else is by clarifying the support to only Ubuntu here maybe in the in the description and later in the job builder adding a note to the field or something like that?

rancher-max: Added a comment above the function to describe that, and created https://github.com/rancher/qa-tasks/issues/147 for enhancing this

izaac (approved):
From what I understand of airgap installation steps for k3s/rke2
These improvements look good to me now.

bmdepesa (approved):
LGTM

Issue #34669, opened 2021-09-07
[2.5] Backport Missing Permissions in catalogs-manage GlobalRole


ryansann:
Issue:
- https://github.com/rancher/rancher/issues/34592

ryansann:
Testing is the same as [here](https://github.com/rancher/rancher/issues/34592#issuecomment-914694796) though the UI steps are slightly different.

brudnak:
# My checks PASSED

## Reproduction Environment


| Component               | Version / Type                          |
| ----------------------- | --------------------------------------- |
| Rancher version         | v2.5.11                                 |
| Rancher cluster type    | Docker                                  |
| Docker version          | 20.10.7, build 20.10.7-0ubuntu5~20.04.2 |
| Helm version            | v2.16.8-rancher1                        |
| Downstream cluster type | RKE1 Linode                             |
| Downstream K8s version  | v1.20.13                                |
| Browser type            | Google Chrome                           |
| Browser version         | 96.0.4664.110 (Official Build) (x86_64) |

## Reproduction steps

1. Provision a downstream cluster
2. Create a standard user
3. Add the user to the downstream, navigate into the downstream cluster view and add them with the right roles Configure Catalogs
4. Sign in as user, navigate to Apps & Marketplace and try to edit the the chart repo url

**RESULT** 

Unable to edit the chart repo url
  
<hr>
  
## Validation Environment

| Component               | Version / Type                                     |
| ----------------------- | -------------------------------------------------- |
| Rancher version         | v2.5-6ff6d08b210f5ce59f4c8dac817ae63af75a5d76-head |
| Rancher cluster type    | Docker                                             |
| Docker version          | 20.10.7, build 20.10.7-0ubuntu5~20.04.2            |
| Helm version            | v2.16.8-rancher1                                   |
| Downstream cluster type | RKE1 Linode                                        |
| Downstream K8s version  | v1.20.13                                           |
| Browser type            | Google Chrome                                      |
| Browser version         | 96.0.4664.110 (Official Build) (x86_64)            |

## Validation steps

1. Provision a downstream cluster
2. Create a standard user
3. Add the user to the downstream, navigate into the downstream cluster view and add them with the right roles Configure Catalogs
4. Sign in as user, navigate to Apps & Marketplace and try to edit the the chart repo url

**RESULT** 

Successfully able to edit the chart repo url

PR #34670, opened 2021-09-07
[2.5] Backport add clusterrepos to catalogs-manage GlobalRole

ryansann:
Issue:
- https://github.com/rancher/rancher/issues/34669

### DIFF
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -46,7 +46,8 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext) 
 	rb.addRole("Manage Cluster Drivers", "kontainerdrivers-manage").
 		addRule().apiGroups("management.cattle.io").resources("kontainerdrivers").verbs("*")
 	rb.addRole("Manage Catalogs", "catalogs-manage").
-		addRule().apiGroups("management.cattle.io").resources("catalogs", "templates", "templateversions").verbs("*")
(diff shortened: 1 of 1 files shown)


pkg/data/management/role_data.go
```diff
@@ -46,7 +46,8 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext)
 	rb.addRole("Manage Cluster Drivers", "kontainerdrivers-manage").
 		addRule().apiGroups("management.cattle.io").resources("kontainerdrivers").verbs("*")
 	rb.addRole("Manage Catalogs", "catalogs-manage").
-		addRule().apiGroups("management.cattle.io").resources("catalogs", "templates", "templateversions").verbs("*")
+		addRule().apiGroups("management.cattle.io").resources("catalogs", "templates", "templateversions").verbs("*").
+		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("*")
```
cbron, line 50: I'm confused about something. This role grants access to global catalogs, and not clustercatalogs or projectcatalogs.  In the new system there are only clusterrepos, which are the equivalent of both catalogs and clustercatalogs. 
```
Create a standard user, check the "Configure Catalogs" builtin role
Add new user as a member to a cluster (local or downstream)
Log in as user and go to Apps & Marketplace > Chart Repositories in the cluster
```
Doesn't `clustercatalogs-manage` actually make more sense ? Or I could see both.

cbron (changes requested):
The original PR has changed quite a bit and needs re-assessed, this one will need changes also.

Issue #34580, opened 2021-08-31
Rancher 2.6.0 rancher-agent fails to provision on FIPS enabled node


danbarr:
**Rancher Server Setup**
- Rancher version: 2.6.0
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): k3s
- Proxy/Cert Details: No proxy; cert issued by internal CA, CA root cert is properly available at /v3/settings/cacerts

**Information about the Cluster**
- Kubernetes version: 1.20.9
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom OR Infrastructure Provider (tested w/Amazon EC2)

**Describe the bug**
Started testing Rancher 2.6.0 today. I attempted to create a new RKE 1.20.9 cluster using the Amazon EC2 infrastructure provider, with a RHEL 7 base AMI that has FIPS mode enabled. The node successfully installs Docker and launches the rancher-agent container, but it fails to provision with an error, `crypto/fips/fips.c:156: OpenSSL internal error: FATAL FIPS SELFTEST FAILURE`.

Also attempted a Custom cluster and running the generated docker command on a FIPS-enabled node, with the same result.

This worked with the same base AMI as of 2.5.9.

**To Reproduce**
Custom: Prepare a FIPS-enabled node (ex: RHEL 7). Create a custom cluster, and run the generated docker run command on the node. The rancher-agent container will start but immediately enter a restarting loop.

Infrastructure provider: Create a cluster using an infrastructure provider node template. Use a base node image which is FIPS-enabled (ex: a RHEL 7 AMI with FIPS mode enabled). The node will reach the "Waiting to register with Kubernetes" state but go no further.

For either case, connect to the node, run `docker logs <rancher-agent container id>` and observe the error.

**Result**
The rancher-agent container enters a restart loop with the following error (hostname and IP addresses sanitized):
```
INFO: https://rancher.example.com/ping is accessible
INFO: rancher.example.com resolves to 1.1.1.1 2.2.2.2
crypto/fips/fips.c:156: OpenSSL internal error: FATAL FIPS SELFTEST FAILURE
ERROR: The environment variable CATTLE_CA_CHECKSUM is set but there is no CA certificate configured at https://rancher.example.com/v3/settings/cacerts
```

[...]

PR #34694, opened 2021-09-09
[WIP] Add libopenssl1_1-hmac to BCI images

cjellick:
I've not tested this at all, but I wanted to get the PR up as a theoretical way forward, in the hopes someone else will take it over.

Self-tests for OpenSSL were failing because this package was missing.

Assuming this works, would address https://github.com/rancher/rancher/issues/34580

@nikkelma figured this out and POCed this in a simple container.


cjellick:
cc @nikkelma. I edited the comment to add you, but not sure if an edit notifies you.

cjellick:
This PR is not needed if we just bump our BCI based images to `15.3.17.8.1`.

PR #31557, opened 2021-03-02
Adding terraform version param for validation tools

izaac:
Just parameterize the terraform version.
I passed a shellcheck linter on it too.

Tested locally

```
➜  validation git:(terraform_tools_parameter) ✗ tests/v3_api/scripts/build.sh
+ set -eu
+ DEBUG=false
+ RKE_VERSION=v1.0.2
+ RANCHER_HELM_VERSION=v3.3.4
+ KUBECTL_VERSION=v1.16.6
+ CLI_VERSION=v2.4.5
+ SONOBUOY_VERSION=0.18.2
+ TERRAFORM_VERSION=0.13.6
+ JOB_NAME=testizaac1
+ BUILD_NUMBER=1
++ basename testizaac1
+ TRIM_JOB_NAME=testizaac1
+ '[' false '!=' false ']'
+++ dirname tests/v3_api/scripts/build.sh
++ cd tests/v3_api/scripts/../../../
++ pwd
+ cd /Users/izaac/repos/rancher-izaac/tests/validation
+ count=0
+ [[ 3 -gt 0 ]]
+ docker build -q -f Dockerfile.v3api --build-arg CLI_VERSION=v2.4.5 --build-arg RKE_VERSION=v1.0.2 --build-arg RANCHER_HELM_VERSION=v3.3.4 --build-arg KUBECTL_VERSION=v1.16.6 --build-arg SONOBUOY_VERSION=0.18.2 --build-arg TERRAFORM_VERSION=0.13.6 -t rancher-validation-testizaac11 .
sha256:ffc595b19f7c1a019c75672ee59e23fb1a359fb33989231b5a309d439d73b464
+ [[ 0 -eq 0 ]]
+ break

➜  validation git:(terraform_tools_parameter) ✗ docker run rancher-validation-testizaac11 terraform version
Terraform v0.13.6

Your version of Terraform is out of date! The latest version
is 0.14.7. You can update by downloading from https://www.terraform.io/downloads.html
```

PR #34697, opened 2021-09-09
Adding create GKE HA cluster to the create HA test

izaac:
This will require terraform v0.13 for the `gke_auth` module to work. That was prepared for parameterization [here](https://github.com/rancher/rancher/pull/31557). 

`TERRAFORM_VERSION` will be added as a parameter in the test HA job in a separate PR along with the required GKE parameters.

#30956

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -109,6 +109,13 @@ def test_install_rancher_ha(precheck_certificate_options):
                 "--set ingress."
                 "extraAnnotations.\"kubernetes\\.io/ingress\\.class\"=nginx"
             )
+        elif RANCHER_LOCAL_CLUSTER_TYPE == "GKE":
+            create_gke_cluster()
+            install_gke_ingress()
+            extra_settings.append(
+                "--set ingress."
+                "extraAnnotations.\"kubernetes\\.io/ingress\\.class\"=nginx"
+            )
     else:
         write_kubeconfig()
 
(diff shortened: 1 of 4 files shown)


tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -246,33 +248,34 @@ def install_eks_ingress():
...
+        kubectl_ingress = "kubectl get ingress -n cattle-system -o " \
+                          "jsonpath=\"" \
+                          "{.items[0].status.loadBalancer.ingress[0].hostname}\""
+        time.sleep(10)
+        ingress_address = run_command_with_stderr(export_cmd
+                                                  + " && " +
+                                                  kubectl_ingress).decode()
+        AmazonWebServices().upsert_route_53_record_cname(RANCHER_HA_HOSTNAME,
+                                                         ingress_address)
+
```
bmdepesa, line 263: Nit: else if for these comparisons, and maybe a graceful failure if `RANCHER_LOCAL_CLUSTER_TYPE` is an unexpected value

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -489,11 +492,41 @@ def create_aks_cluster():
...
+    print("Creating cluster")
+    tf.init()
+    print(tf.plan(out="gks_plan_server.out"))
+    print("\n\n")
+    print(tf.apply("--auto-approve"))
+    print("\n\n")
+    out_string = tf.output("kube_config", full_value=True)
+    print("GKE KUBECONFIG")
+    print("\n\n")
+    print(out_string)
```
bmdepesa, line 514: Can we also print the base64 encoded kubeconfig? Not 100% necessary but can be useful to avoid formatting issues.

bmdepesa (changes requested):

tests/validation/tests/v3_api/test_create_ha.py
izaac: The base64 gets printed later in the test, you can find it in the output of the automation execution.

tests/validation/tests/v3_api/test_create_ha.py
izaac: I just refactored the code to use `elif` it was called out before the providers code was repetitive that's why I consolidated them in one if, but it makes more sense this way to make changes more easy in case there's a difference in the future between these cluster types.

It's highly unlikely the test will reach this part of the code if a wrong `RANCHER_LOCAL_CLUSTER_TYPE` is set but I've added an else anyways as requested:

```
>           pytest.fail("Wrong RANCHER_LOCAL_CLUSTER_TYPE {}"
                        .format(RANCHER_LOCAL_CLUSTER_TYPE))
E           Failed: Wrong RANCHER_LOCAL_CLUSTER_TYPE GKEX

v3_api/test_create_ha.py:291: Failed
=========================== 1 failed in 0.18 seconds ===========================

Process finished with exit code 1
```

bmdepesa:
LGTM

izaac:
Closing in favor of https://github.com/rancher/rancher/pull/38609
The fork branch was stale after the forked repository removed time ago

PR #34698, opened 2021-09-09
added aksUpgrade versions endpoint

paynejacob:
#34161

Old pr for reference: https://github.com/rancher/rancher/pull/34624

# Problem:
In AKS you cannot skip minor version updates. Currently we list all available versions in the upgrade drop down. This lets users attempt to upgrade to version that is too high causing the cluster to go into a constant error state.

# Solution:
I added the a new endpoint, `/meta/aksUpgrades`. This endpoint will return all available k8s versions similar to `/meta/aksVersions` however each version includes an enabled  flag.  This flag is true if the associated version is a valid upgrade version for the given cluster.  

# Testing

Valid response
```
$ curl -s -H "Authorization: Bearer $RANCHER_TOKEN" "https://<rancher>/meta/aksUpgrades?cloudCredentialId=<creds>&region=eastus&clusterId=<cluster id>" | jq
{
  "currentVersion": "1.20.7",
  "upgrades": [
    {
      "version": "1.18.17",
      "enabled": false
    },
    {
      "version": "1.18.19",
      "enabled": false
    },
    {
      "version": "1.19.11",
      "enabled": false
    },
    {
      "version": "1.19.13",
      "enabled": false
    },
    {
      "version": "1.20.7",
      "enabled": false
    },
    {
      "version": "1.20.9",
      "enabled": true
    },
    {
      "version": "1.21.1",
      "enabled": true
    },
    {
      "version": "1.21.2",
      "enabled": true
    }
  ]
}
```

Error Response
```
$ curl -s -H "Authorization: Bearer $RANCHER_TOKEN" "https://<rancher>/meta/aksUpgrades?cloudCredentialId=<creds>&region=eastus&clusterId=<cluster id>" | jq
{
  "error": "invalid cluster id"
}
```

### DIFF
--- a/pkg/api/norman/customization/aks/handler.go
+++ b/pkg/api/norman/customization/aks/handler.go
@@ -40,6 +40,7 @@ type Capabilities struct {
 	ClientID         string `json:"clientId"`
 	ClientSecret     string `json:"clientSecret"`
 	ResourceLocation string `json:"region"`
+	ClusterID        string `json:"clusterId"`
 }
 
 // AKS handler lists available resources in Azure API
@@ -100,6 +101,13 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	var err error
 
 	switch resourceType {
+	case "aksUpgrades":
+		if serialized, errCode, err = listKubernetesUpgradeVersions(req.Context(), h.clusterLister, capa); err != nil {
+			logrus.Errorf("[aks-handler] error getting kubernetes upgrade versions: %v", err)
+			handleErr(writer, errCode, err)
+			return
+		}
+		writer.Write(serialized)
 	case "aksVersions":
 		if serialized, errCode, err = listKubernetesVersions(req.Context(), capa); err != nil {
 			logrus.Errorf("[aks-handler] error getting kubernetes versions: %v", err)
@@ -239,10 +247,9 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 	if cap.AuthBaseURL == "" {
 		cap.AuthBaseURL = azure.PublicCloud.ActiveDirectoryEndpoint
 	}
-	region := req.URL.Query().Get("region")
-	if region != "" {
-		cap.ResourceLocation = region
-	}
+
+	cap.ResourceLocation = req.URL.Query().Get("region")
+	cap.ClusterID = req.URL.Query().Get("clusterId")
 
 	return http.StatusOK, nil
 }
--- a/pkg/api/norman/customization/aks/listers.go
+++ b/pkg/api/norman/customization/aks/listers.go
@@ -16,6 +16,8 @@ import (
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/mcuadros/go-version"
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
 )
 
 type virtualNetworksResponseBody struct {
@@ -118,6 +120,102 @@ func (s sortableVersion) Swap(a, b int) {
 
 func (s sortableVersion) Less(a, b int) bool {
 	return version.Compare(s[a], s[b], "<")
+}
+
+type KubernetesUpgradeVersion struct {
+	Version string `json:"version"`
+	Enabled bool   `json:"enabled"`
+}
+
+type KubernetesUpgradeVersions []*KubernetesUpgradeVersion
+
+func (s KubernetesUpgradeVersions) Len() int {
+	return len(s)
+}
+
+func (s KubernetesUpgradeVersions) Swap(a, b int) {
+	s[a], s[b] = s[b], s[a]
+}
+
+func (s KubernetesUpgradeVersions) Less(a, b int) bool {
+	return version.Compare(s[a].Version, s[b].Version, "<")
+}
+
+type UpgradeVersionsResponse struct {
+	CurrentVersion string                    `json:"currentVersion"`
+	Upgrades       KubernetesUpgradeVersions `json:"upgrades"`
+}
+
+// listKubernetesUpgradeVersions lists all kubernetes versions listed by AKS Container Service and marks which ones the
+//given cluster can be upgraded to.  A version's `Enabled` flag is true if the cluster can be upgraded to the version
+//in its current state.
+func listKubernetesUpgradeVersions(ctx context.Context, clusterLister v3.ClusterLister, cap *Capabilities) ([]byte, int, error) {
+	var resp UpgradeVersionsResponse
+
+	// load the target cluster, if the cluster is not found we cannot proceed
+	cluster, err := clusterLister.Get("", cap.ClusterID)
+	if err != nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("invalid cluster id")
+	}
+
+	resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+
+	// get the client for aks container service
+	clientContainer, err := NewContainerServiceClient(cap)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	// request a list of orchestrators
+	orchestrators, err := clientContainer.ListOrchestrators(ctx, cap.ResourceLocation, "managedClusters")
+	if err != nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("failed to get orchestrators: %v", err)
+	}
+
+	// ensure the orchestrators are returned
+	if orchestrators.Orchestrators == nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("no version profiles returned: %v", err)
+	}
+
+	var upgradeVersions map[string]bool
+	for _, profile := range *orchestrators.Orchestrators {
+		// check for nil pointers to avoid a panic
+		if profile.OrchestratorType == nil || profile.OrchestratorVersion == nil {
+			logrus.Warning("unexpected nil orchestrator type or version")
+			continue
+		}
+
+		// exclude any non kubernetes types
+		if containerservice.OrchestratorTypes(*profile.OrchestratorType) != containerservice.Kubernetes {
+			continue
+		}
+
+		// exclude any versions older than the current version
+		if version.Compare(*profile.OrchestratorVersion, resp.CurrentVersion, "<") {
+			continue
+		}
+
+		// generate the upgrade map when the current version is found
+		if *profile.OrchestratorVersion == resp.CurrentVersion {
+			upgradeVersions = upgradeableVersionsMap(profile)
+			continue
+		}
+
+		// store this kubernetes version
+		resp.Upgrades = append(resp.Upgrades, &KubernetesUpgradeVersion{Version: *profile.OrchestratorVersion})
+	}
+
+	// enable any version listed in the upgrade versions
+	for _, kubernetesVersion := range resp.Upgrades {
+		if upgradeVersions[kubernetesVersion.Version] {
+			kubernetesVersion.Enabled = true
+		}
+	}
+
+	// sort versions
+	sort.Sort(resp.Upgrades)
+
+	return encodeOutput(resp)
 }
 
 func listKubernetesVersions(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
@@ -325,4 +423,14 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
 	}
 
 	return data, http.StatusOK, err
+}
+
+func upgradeableVersionsMap(upgradeProfile containerservice.OrchestratorVersionProfile) map[string]bool {
+	rval := make(map[string]bool, 0)
+
+	for _, profile := range *upgradeProfile.Upgrades {
+		rval[*profile.OrchestratorVersion] = true
+	}
+
+	return rval
 }


pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +122,94 @@ func (s sortableVersion) Less(a, b int) bool {
...
+}
+
+// listKubernetesUpgradeVersions lists all kubernetes versions listed by AKS Container Service and marks which ones the
+//given cluster can be upgraded to.  A version's `Enabled` flag is true if the cluster can be upgraded to the version
+//in its current state.
+func listKubernetesUpgradeVersions(ctx context.Context, clusterLister v3.ClusterLister, cap *Capabilities) ([]byte, int, error) {
+	// load the target cluster, if the cluster is not found we cannot proceed
+	cluster, err := clusterLister.Get("", cap.ClusterId)
+	if err != nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("failed to load cluster: %v", err)
```
snasovich, line 151: If the cluster is not found, I can get behind 400. I've seen 404 used before, but it's really somewhat of a controversial topic - using 404 for API responses.
However, if there is any other reason, it should probably be 500.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +122,94 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	// get the client for aks container service
+	clientContainer, err := NewContainerServiceClient(cap)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	// request a list of orchestrators
+	orchestrators, err := clientContainer.ListOrchestrators(ctx, cap.ResourceLocation, "managedClusters")
+	if err != nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("failed to get orchestrators: %v", err)
```
snasovich, line 163: Wouldn't 500 make more sense here? There was some error listing orchestrators.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +122,94 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	}
+
+	// ensure the orchestrators are returned
+	if orchestrators.Orchestrators == nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("no version profiles returned: %v", err)
```
snasovich, line 168: What's the reasoning behind 400 here?

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +122,94 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	}
+
+	versions := make(KubernetesUpgradeVersions, 0)
+	upgradeVersions := make(map[string]bool, 0)
+	for _, profile := range *orchestrators.Orchestrators {
+
+		// check for nil pointers to avoid a panic,
```
snasovich, line 175: nit, but I think blank line before comment is unnecessary (same below)
Also, trailing comma in the comment seems unnecessary.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +122,94 @@ func (s sortableVersion) Less(a, b int) bool {
...
+			if *profile.OrchestratorVersion == *cluster.Spec.AKSConfig.KubernetesVersion {
+				upgradeVersions = upgradeableVersionsMap(profile)
+			}
+		}
+	}
+
+	// disable any version not listed in the upgrade versions
+	for _, kubernetesVersion := range versions {
+		if !upgradeVersions[kubernetesVersion.Version] {
+			kubernetesVersion.Enabled = false
```
snasovich, line 197: This may be opinion-based, but I think it's more natural to add all versions with `Enabled=false` first and then mark the ones that can be upgraded to with `Enabled=true`. Feel free to ignore if you disagree.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -326,3 +416,13 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
 
 	return data, http.StatusOK, err
 }
+
+func upgradeableVersionsMap(upgradeProfile containerservice.OrchestratorVersionProfile) map[string]bool {
```
snasovich, line 420: Suggested name `buildUpgradeableVersionsMap` to include verb in func name.

snasovich (changes requested):

pkg/api/norman/customization/aks/handler.go
```diff
@@ -40,6 +40,7 @@ type Capabilities struct {
 	ClientID         string `json:"clientId"`
 	ClientSecret     string `json:"clientSecret"`
 	ResourceLocation string `json:"region"`
+	ClusterId        string `json:"clusterId"`
```
snasovich, line 43: This should be `ClusterID` and actually seems to fail the build:
```
/bin/golangci-lint
--
843 | Running: golangci-lint
844 | pkg/api/norman/customization/aks/listers.go:7: File is not `goimports`-ed (goimports)
845 | v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
846 | "github.com/sirupsen/logrus"
847 | pkg/api/norman/customization/aks/handler.go:43:2: var-naming: struct field ClusterId should be ClusterID (revive)
848 | ClusterId        string `json:"clusterId"`
849 | ^
```

pkg/api/norman/customization/aks/listers.go
paynejacob: I copied this from other parts of the codebase for these endpoints.  I agree a 500 makes more sense but then we are deviating from the behavior of the other endpoints.  I think we should make a ticket to cleanup the code for these endpoints and make the response consistent for all of them.

pkg/api/norman/customization/aks/listers.go
paynejacob: ill flip it and add docs on how to thing about the transformations

snasovich: I'm not sure you mean by "add docs" - if you mean update comments accordingly, then it makes sense. I don't think we need to formally "document" this internal logic elsewhere.

pkg/api/norman/customization/aks/listers.go
snasovich: Ah, I see now. In this case we should favor consistency, so no change is necessary here. For the same reason, I can get behind closing the other 2 comments (including clusterId handling).

pkg/api/norman/customization/aks/handler.go
```diff
@@ -100,6 +101,13 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	var err error
 
 	switch resourceType {
+	case "aksUpgrades":
+		if serialized, errCode, err = listKubernetesUpgradeVersions(req.Context(), h.clusterLister, capa); err != nil {
+			logrus.Errorf("[aks-handler] error getting kubernetes upgrade versions: %v", err)
```
rmweir, line 106: There are a couple things that might be referred to as the aks-handler, could call this aksUpgrades handler or something.

rmweir (changes requested):
Some nits, otherwise LGTM after @snasovich feedback has been addressed.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
+			versions = append(versions, &KubernetesUpgradeVersion{Version: *profile.OrchestratorVersion})
+
+			// generate the upgrade map when the current version is found
+			if *profile.OrchestratorVersion == *cluster.Spec.AKSConfig.KubernetesVersion {
+				upgradeVersions = buildUpgradeableVersionsMap(profile)
+			}
+		}
+	}
+
+	// disable any version not listed in the upgrade versions
```
snasovich, line 193: This comment needs to be updated to match the change in implementation (setting Enabled to `true` vs. `false` as before)

snasovich (changes requested):
One minor comment about comment ;) needs addressing. Otherwise LGTM.

snasovich (approved):

cmurphy (approved):
LGTM, couple of nits but feel free to ignore, one question but I think this is the right approach for now

pkg/api/norman/customization/aks/handler.go
```diff
@@ -100,6 +101,13 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	var err error
 
 	switch resourceType {
+	case "aksUpgrades":
+		if serialized, errCode, err = listKubernetesUpgradeVersions(req.Context(), h.clusterLister, capa); err != nil {
+			logrus.Errorf("[aksUpgrades] error getting kubernetes upgrade versions: %v", err)
```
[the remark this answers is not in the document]
snasovich: This was in response to https://github.com/rancher/rancher/pull/34698#discussion_r706411257
Do you see an issue with using a different tag? Maybe use something like `[aks-handler:aksUpgrades]`?

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
+				upgradeVersions = buildUpgradeableVersionsMap(profile)
+			}
+		}
+	}
+
+	// enable any version listed in the upgrade versions
+	for _, kubernetesVersion := range versions {
+		if upgradeVersions[kubernetesVersion.Version] {
+			kubernetesVersion.Enabled = true
+		}
```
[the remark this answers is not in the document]
snasovich: I don't think we can be sure we're done appending to `versions` by the time we encounter the profile for the current version. So we need to gather all versions first and only then cross-reference it with `upgradeVersions` list.

pkg/api/norman/customization/aks/handler.go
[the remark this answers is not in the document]
cmurphy: It's just inconsistent with the other log messages, if someone is grepping for log messages for this AKS metadata handler they would use `aks-handler` and miss any log messages for this endpoint. I think @rmweir didn't take the other endpoints into account when making that comment. So far it's only this metadata handler using `aks-handler` so it's not likely to be confused with other AKS components.

pkg/api/norman/customization/aks/listers.go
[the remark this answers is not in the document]
paynejacob: It could not because we don't have the the upgradeMap until after that loop completes.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	}
+
+	// sort versions
+	sort.Sort(versions)
+
+	return encodeOutput(struct {
+		CurrentVersion string                    `json:"currentVersion"`
+		Upgrades       KubernetesUpgradeVersions `json:"upgrades"`
```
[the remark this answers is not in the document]
paynejacob: Yes I talked to @vincent99 any reason we give would be a guess, aks does not explain why versions were included or excluded in their response.

cmurphy (approved):

snasovich (dismissed):

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	if err != nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("failed to get orchestrators: %v", err)
+	}
+
+	// ensure the orchestrators are returned
+	if orchestrators.Orchestrators == nil {
+		return nil, http.StatusBadRequest, fmt.Errorf("no version profiles returned: %v", err)
+	}
+
+	versions := make(KubernetesUpgradeVersions, 0)
```
dramich, line 172: A slice doesn't need to be created with `make` and initializing to zero doesn't do anything. So unless you know the len then exclude the 0

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	upgradeVersions := make(map[string]bool, 0)
```
dramich, line 173: Same comment here, 0 isn't needed.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -8,6 +8,9 @@ import (
 	"regexp"
 	"sort"
 
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/sirupsen/logrus"
+
```
dramich, line 13: Needs to be sorted

pkg/api/norman/customization/aks/handler.go
```diff
@@ -239,10 +247,9 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 	if cap.AuthBaseURL == "" {
 		cap.AuthBaseURL = azure.PublicCloud.ActiveDirectoryEndpoint
 	}
-	region := req.URL.Query().Get("region")
-	if region != "" {
-		cap.ResourceLocation = region
-	}
+
+	cap.ResourceLocation = req.URL.Query().Get("region")
```
dramich, line 251: What is the intention of this change? if cap has a value then this would blank it out if it's not in the request

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
 }
 
+type KubernetesUpgradeVersion struct {
+	Version string `json:"version"`
+	Enabled bool   `json:"enabled"`
+}
+
+type KubernetesUpgradeVersions []*KubernetesUpgradeVersion
+
+func (s KubernetesUpgradeVersions) Len() int {
```
dramich, line 133: This shouldn't be needed, it's already a slice

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +123,92 @@ func (s sortableVersion) Less(a, b int) bool {
...
+			if *profile.OrchestratorVersion == *cluster.Spec.AKSConfig.KubernetesVersion {
+				upgradeVersions = buildUpgradeableVersionsMap(profile)
+			}
+		}
+	}
+
+	// enable any version listed in the upgrade versions
+	for _, kubernetesVersion := range versions {
+		if upgradeVersions[kubernetesVersion.Version] {
+			kubernetesVersion.Enabled = true
```
dramich, line 196: If we only care about versions they can upgrade to why return all of them?

dramich:
If you have a 1.19 cluster does the response still show correctly?

pkg/api/norman/customization/aks/listers.go
dramich: And why is this created here when you overwrite it with buildUpgradeableVersionsMap

pkg/api/norman/customization/aks/listers.go
```diff
@@ -326,3 +415,13 @@ func encodeOutput(result interface{}) ([]byte, int, error) {
 
 	return data, http.StatusOK, err
 }
+
+func buildUpgradeableVersionsMap(upgradeProfile containerservice.OrchestratorVersionProfile) map[string]bool {
+	rval := make(map[string]bool, 0)
+
+	for _, profile := range *upgradeProfile.Upgrades {
```
dramich, line 422: If this profile has all the info we need why are we doing all the other processing?

pkg/api/norman/customization/aks/handler.go
snasovich: This was discussed here: https://github.com/rancher/rancher/pull/34624#discussion_r701542177

pkg/api/norman/customization/aks/listers.go
snasovich: This was part of discussion with Vince started by this comment on the original PR: https://github.com/rancher/rancher/pull/34624#issuecomment-912766425 (see few comments below too)

pkg/api/norman/customization/aks/handler.go
dramich: Kinda? My question still stands, if cap has a value and the request doesn't we are going to blank it... then what?

pkg/api/norman/customization/aks/listers.go
dramich: It looks like he wants the available upgrade versions not ALL versions including downgrades. Why would he display downgrade versions in the upgrade screen?

pkg/api/norman/customization/aks/handler.go
rmweir: When I traced this myself it looked like it would not ever have a value as things currently are.

rmweir (approved):
LGTM, after dan's comments are addressed.

pkg/api/norman/customization/aks/handler.go
snasovich: The only place this unexported func is called passes empty `Capabilities` object, so immediate usage is not a concern as there is nothing that could be overwritten.
Overall, it looks to me that the goal of this internal func is to initialize capabilities from request, not to augment existing capabilities passed to it - judging by its current usage. So it would be more fitting for the func to return capabilities rather than take a pointer to one.
Given the current signature, I do see your point and don't mind reinstating this check. Feels like we're splitting hairs here though a bit.

pkg/api/norman/customization/aks/listers.go
snasovich: Are you sure? https://play.golang.org/p/DJJNS35CGzr

Also, it's done for the very much the same situation: https://github.com/rancher/rancher/blob/ef760ce7606af9ef589cc1dc89253acc27609b5b/pkg/api/norman/customization/aks/listers.go#L111

snasovich: Unless `sort` doesn't use `len` for slices already. My bad probably. :)

pkg/api/norman/customization/aks/listers.go
snasovich: Unlike make for maps, make for slices actually requires length and it cannot be omitted.

pkg/api/norman/customization/aks/listers.go
paynejacob: We are implementing this interface https://pkg.go.dev/sort#Interface

pkg/api/norman/customization/aks/listers.go
paynejacob: This profile will not contain future versions that are not supported yet.

paynejacob:
Retested with latest changes

```
GET https://<rancher host>/meta/aksVersions?cloudCredentialId=<creds>&region=<region>

["1.19.11","1.19.13","1.20.7","1.20.9","1.21.1","1.21.2"]
```

```
GET https://<rancher host>/meta/aksUpgrades?cloudCredentialId=< creds >&region=<region>&clusterId=<cluster id>

{"currentVersion":"1.19.13","upgrades":[{"version":"1.20.7","enabled":true},{"version":"1.20.9","enabled":true},{"version":"1.21.1","enabled":false},{"version":"1.21.2","enabled":false}]}
```

pkg/api/norman/customization/aks/handler.go
snasovich: @paynejacob , I see this was not changed. Did you meet with Dan and agree not to address it?

pkg/api/norman/customization/aks/listers.go
```diff
@@ -120,6 +122,100 @@ func (s sortableVersion) Less(a, b int) bool {
...
+	for _, kubernetesVersion := range versions {
+		if upgradeVersions[kubernetesVersion.Version] {
+			kubernetesVersion.Enabled = true
+		}
+	}
+
+	// sort versions
+	sort.Sort(versions)
+
+	return encodeOutput(struct {
```
dramich, line 210: We shouldn't declare a new struct type inline

paynejacob:
Closing this pr due to weird github behavior and also to change the branch

PR #34708, opened 2021-09-11
Allow truncated secrets/configmaps in collection

aiyengar2:
This PR adds the ability to specify an option, `truncateBytes`, to return a list of secrets / configmaps that are truncated. 

If truncated, the secrets will be returned with an `isTruncated` field.

The change has been applied to both the Norman and Steve APIs in this PR.

For the Norman API, the change was applied to `/v3/project/CLUSTER:PROJECT_ID/namespacedsecrets`, `/v3/project/CLUSTER:PROJECT_ID/secrets`, and `/v3/project/CLUSTER:PROJECT_ID/configmaps`.

For the Steve API, the change was applied to `/v1/secrets` as well as `/v1/configmaps`.

Steve PR: https://github.com/rancher/steve/pull/27

Related Issue: https://github.com/rancher/rancher/issues/25487

---

#### Context

The purpose of this PR is to allow the UI to be able to add something like `?truncateBytes=640` in order to ensure that they aren't loading the entire contents of secrets / configmaps on just trying to list resources.

Instead, they will be able to partially load the resources and decide whether they need to collect the full data of the secret / configmap if it has been truncated (which is indicated by the `isTruncated` flag on the secret / configmap) when a user needs a detailed view of the content of a secret / configmap.


aiyengar2:
Pending merge of Steve PR first

Issue #34746, opened 2021-09-15
[Hosted Rancher] Consistent Max CPU load upon upgrade to v2.6.0


nickgerace:
Issues occurs upon hosted, HA upgrade from v2.5.9 to v2.6.0.

Child issues:
- Fleet: https://github.com/rancher/rancher/issues/34776
- Monitoring v1: https://github.com/rancher/rancher/issues/34833

SURE-3301

dnoland1:
Tested with upstream Rancher v2.6.1-rc5 and upgrading downstream Rancher from v2.5.9 to v2.6.1-rc5. CPU Utilization and Load Average were the same before and after the upgrade. Checked the Rancher pod logs (both upstream & downstream) and did not see the repeated "Creating secret..." logs.

Performance graph (upgrade a little after 17:00):
<img width="801" alt="image" src="https://user-images.githubusercontent.com/468462/135183373-f1d0cef7-2322-4cf0-a998-a438f77fec93.png">

PR #34754, opened 2021-09-15
Only add projectID to Fleet NS when system project exists

nickgerace:
Relevant issues:
- This PR partially addresses: https://github.com/rancher/rancher/issues/34746
- This PR primarily addresses: https://github.com/rancher/rancher/issues/34776

## Description

There are two fixes and one "bonus" here. Of the former, one is the actual and the other is preventative.

## Actual
Fleet namespaces will not longer have the projectID annotation if the system project does not exist.
- We've observed handlers using this malformed annotation (e.g. `"local:"` instead of `local:<system-project-id>`).
- When it's removed, the namespace is no longer loaded into the cluster secrets creation sync: https://github.com/rancher/rancher/blob/ebcb2321e254a9d172a130613a1a2d6e06d23248/pkg/controllers/managementuser/secret/secret.go#L102
- Without this PR, an empty string is loaded into the management secrets "List" call, causing significant CPU usage with the following loop running upon every reconciliation of the namespace: https://github.com/rancher/rancher/blob/ebcb2321e254a9d172a130613a1a2d6e06d23248/pkg/controllers/managementuser/secret/secret.go#L110-L127

## Preventative
We will not create management secrets if the projectID is empty.
- This is a safeguard for other namespaces that may have the same bug in the future: https://github.com/rancher/rancher/blob/ebcb2321e254a9d172a130613a1a2d6e06d23248/pkg/controllers/managementuser/secret/secret.go#L105-L108
- We could create and return a new error above (I'd be happy to change it to do so; "testing" it would involve undoing the Fleet change and then re-adding it afterwards for an A/B comparison)

## Bonus
There's an unhandled error that we plan on addressing in another PR.
- For now, I've added it to a trace log so that it's not completely hidden: https://github.com/rancher/rancher/pull/34754/files#diff-51209acaa8140b964a493d6ab61469493b6fc09d9f0583ef3d586cda2fe31849R74-R76

Original comment containing the above: https://github.com/rancher/rancher/pull/34754#issuecomment-921058555

### DIFF
--- a/pkg/controllers/managementuser/secret/secret.go
+++ b/pkg/controllers/managementuser/secret/secret.go
@@ -102,6 +102,10 @@ func (n *NamespaceController) sync(key string, obj *corev1.Namespace) (runtime.O
 	if obj.Annotations[projectIDLabel] != "" {
 		parts := strings.Split(obj.Annotations[projectIDLabel], ":")
 		if len(parts) == 2 {
+			if parts[1] == "" {
+				logrus.Debugf("[NamspaceController|sync] empty project name found in obj.Annotations[projectIDLabel] for cluster: %s", parts[0])
+				return nil, nil
+			}
 			// on the management side, secret's namespace name equals to project name
 			secrets, err := n.managementSecrets.List(parts[1], labels.NewSelector())
 			if err != nil {
--- a/pkg/controllers/managementagent/nslabels/labels.go
+++ b/pkg/controllers/managementagent/nslabels/labels.go
@@ -71,7 +71,9 @@ func (nsh *namespaceHandler) addProjectIDLabelToNamespace(ns *corev1.Namespace, 
 		return fmt.Errorf("cannot add label to nil namespace")
 	}
 	if ns.Labels[ProjectIDFieldLabel] != projectID {
-		nsh.updateProjectIDLabelForSecrets(projectID, ns.Name, clusterID)
+		if err := nsh.updateProjectIDLabelForSecrets(projectID, ns.Name, clusterID); err != nil {
+			logrus.Trace(err)
+		}
 		logrus.Infof("namespaceHandler: addProjectIDLabelToNamespace: adding label %v=%v to namespace=%v", ProjectIDFieldLabel, projectID, ns.Name)
 		nscopy := ns.DeepCopy()
 		if nscopy.Labels == nil {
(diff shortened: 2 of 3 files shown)


StrongMonkey (dismissed):

dramich:
How does this fix the issue seen?

nickgerace:
@dramich there are two fixes and one "bonus" here. Of the former, one is the actual and the other is preventative.

**Actual:** Fleet namespaces will not longer have the projectID annotation if the system project does not exist.
- We've observed handlers using this malformed annotation (e.g. `"local:"` instead of `local:<system-project-id>`).
- When it's removed, the namespace is no longer loaded into the cluster secrets creation sync: https://github.com/rancher/rancher/blob/ebcb2321e254a9d172a130613a1a2d6e06d23248/pkg/controllers/managementuser/secret/secret.go#L102
- What happens is that an empty string is loaded into the management secrets "List" call, causing significant CPU usage with the following loop running upon every reconciliation of the namespace: https://github.com/rancher/rancher/blob/ebcb2321e254a9d172a130613a1a2d6e06d23248/pkg/controllers/managementuser/secret/secret.go#L110-L127

**Preventative:** We will not create management secrets if the projectID is empty.
- This is a safeguard for other namespaces that may have the same bug in the future: https://github.com/rancher/rancher/blob/ebcb2321e254a9d172a130613a1a2d6e06d23248/pkg/controllers/managementuser/secret/secret.go#L105-L108

**Bonus:** There's an unhandled error that we plan on addressing in another PR.
- For now, I've added it to a trace log so that it's not completely hidden: https://github.com/rancher/rancher/pull/34754/files#diff-51209acaa8140b964a493d6ab61469493b6fc09d9f0583ef3d586cda2fe31849R74-R76

pkg/controllers/managementuser/secret/secret.go
```diff
@@ -102,6 +102,10 @@ func (n *NamespaceController) sync(key string, obj *corev1.Namespace) (runtime.O
 	if obj.Annotations[projectIDLabel] != "" {
 		parts := strings.Split(obj.Annotations[projectIDLabel], ":")
 		if len(parts) == 2 {
+			if parts[1] == "" {
+				logrus.Tracef("obj.Annotations[projectIDLabel] has an empty project name for cluster: %s", parts[0])
```
dramich, line 106: This might be better suited to debug, and prepend the message with something to help say where this is coming from i.e. `[NamespaceController]`

But, if you are deleting this annotation anyways I wouldn't expect to hit this after the first start of this version.

nickgerace: I agree. I'll change it to a debug statement and make the message more clear

dramich:
@nickgerace Your 2nd comment should be the original pr comment. That's a nice writeup.

nickgerace:
@dramich Thank you. Doing that now.

dramich (dismissed):

dramich (dismissed):

nickgerace:
@dramich rebased and force pushed.

dramich (approved):

Issue #34473, opened 2021-08-25
when user updates rke-metadata-config on dev-v2.6, refreshing `Kontainer-driver-metadata` does not update available k8s versions


slickwarren:
**Rancher Server Setup**
- Rancher version:v2.6-head(b7c440b)
- Installation option (Docker install/Helm Chart):docker 
   - If Helm Chart, Kubernetes Info:
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):k3s
         Node Setup:single node
         Version:v1.21.3+k3s1
- Proxy/Cert Details:self signed certs

**Describe the bug**

* when the user tries to refresh KDM info via the `refresh Kubernetes Metadata` button (cluster management -> drivers), the response is `null` and the latest versions aren't updated in the rancher setup

**To Reproduce**
 
* ensure there is a mismatch in kdm versions for your 2.6 cluster
* go to cluster management -> drivers and click the `refresh Kubernetes Metadata` button with the console open to view the response

**Result**
response is null, versions are not updated

**Expected Result**

the latest version should be pulled, if any, when the user clicks this button

**Screenshots**

**Additional context**

this is the rke-metadata url:
https://releases.rancher.com/kontainer-driver-metadata/dev-v2.6/data.json

kinarashah:
Synced up @slickwarren. I thought the null `response` showed up in UI but that's not the case, it shows up in response under dev tools. We should fix it to send something meaningful than null tho the action by itself isn't supposed to return anything. 

Also, I'd need to create a custom branch for testing k8s versions refresh because right now all versions from dev-v2.6 are released in release-v2.6.

kinarashah:
@slickwarren Looks like I'm already sending the correct response, 200 OK. The response in the dev tools usually has the object or list of objects to be returned, but don't have anything to return so would be ok in this case. Also, confirmed this with Rancher v2.5.9, it is the same behavior so it's not a regression. 
<img width="606" alt="Screen Shot 2021-10-04 at 3 56 18 PM" src="https://user-images.githubusercontent.com/16870565/135937305-fde9ab51-ae4e-47ba-a2ce-b4c4b7f85081.png">

Also, to test for RKE2 version refresh: 
[...]

PR #34755, opened 2021-09-15
[v2.6] reload rke2/k3s config on kdm refresh action

kinarashah:
https://github.com/rancher/rancher/issues/34473

### DIFF
--- a/pkg/api/norman/customization/kontainerdriver/actionhandler.go
+++ b/pkg/api/norman/customization/kontainerdriver/actionhandler.go
@@ -5,19 +5,18 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
-	"time"
-
-	"github.com/rancher/rancher/pkg/namespace"
 
 	mVersion "github.com/mcuadros/go-version"
 	"github.com/rancher/norman/api/handler"
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/rancher/pkg/catalog/utils"
+	"github.com/rancher/rancher/pkg/channelserver"
 	kd "github.com/rancher/rancher/pkg/controllers/management/kontainerdrivermetadata"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/image"
+	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/settings"
 	rketypes "github.com/rancher/rke/types"
 	img "github.com/rancher/rke/types/image"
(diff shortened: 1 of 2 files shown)


pkg/api/norman/customization/kontainerdriver/actionhandler.go
```diff
@@ -81,20 +80,8 @@ func (a ActionHandler) refresh(apiContext *types.APIContext) error {
...
-	setting, err := a.MetadataHandler.SettingLister.Get("", rkeMetadataConfig)
-	if err != nil {
-		return err
-	}
-
-	if setting.Annotations == nil {
-		setting.Annotations = make(map[string]string)
-	}
-
-	setting.Annotations[forceRefreshAnnotation] = strconv.FormatInt(time.Now().Unix(), 10)
```
kinarashah, line 93: `forceRefreshAnnotation` is no longer used anywhere

StrongMonkey: Do you know how/why it was used before? Is this a problem in 2.5?

StrongMonkey:
Since that action is only for refreshing KDM that is used in rke1, should we really include refreshing k3s/rke2 into KDM or make it a separate api for it? I am open for that discussion.

pkg/api/norman/customization/kontainerdriver/actionhandler.go
kinarashah: It was used earlier to refresh k3s metadata based on this annotation, the logic still exists in 2.5 but no longer in 2.6 https://github.com/rancher/rancher/blob/release/v2.5/pkg/controllers/managementapi/k3smetadata/settings_handler.go#L66

kinarashah:
@StrongMonkey This is a good question and I kept it in the same action because: 
- we haven't separated settings for rke1 and rke2, the url and interval is still read from the same v3 setting `rke-metadata-config`
- Action `refresh` continues to exist on v3/kontainerdriver schema, unrelated to the underlying cluster type 
- would we weird to have 2 different actions for user to go and refresh when the same setting interval works for both handlers

pkg/channelserver/channelserver.go
```diff
@@ -61,12 +62,21 @@ func (d *DynamicInterval) Wait(ctx context.Context) bool {
 				return true
 			}
 			continue
+		case msg := <-action:
+			if msg == "refresh" {
```
StrongMonkey, line 66: Nit: use constant for `refresh` as it is used twice.

StrongMonkey (approved):

pkg/channelserver/channelserver.go
kinarashah: updated

dramich (dismissed):

StrongMonkey (approved):

kinarashah:
Pushed changes again to rebase, no other code change. @StrongMonkey could you re-approve?

StrongMonkey (approved):

Issue #34713, opened 2021-09-13
RKE2 Provisioning: can't delete misconfigured node from node pool/cluster in Rancher


izaac:
**Rancher Server Setup**
- Rancher version: `v2.6-head`
- Installation option: Helm
   - If Helm Chart, Kubernetes Cluster and version: RKE `v1.3.0` k8s `v1.21.4`
- Proxy/Cert Details: bring your own valid certs.

**Information about the Cluster**
- Kubernetes version: `v1.21.4-rke2r2`
- Cluster Type: `RKE2` Downstream
   - If downstream, what type of cluster?: `Custom` Driver OCI

**Describe the bug**

When adding a node if I miss a configuration option or enter an invalid value for one of the options the Node goes into a Provisioning error state.
In this case I missed a configuration option by mistake

`Error setting machine configuration from flags provided: no OCI oci-region specified (--oci-region)`

If I try removing the node I see the node removed from the OCI Cloud but the node Reference in Rancher doesn't get removed.

**To Reproduce**

- Create an RKE2 Downstream cluster I used a Custom OCI where I have to enter the configuration values manually
- Make two pools one with `All` roles 1 node
- A second nodepool for `Worker` roles. 1 node.
- On the worker node do not enter a valid configuration value for like the `region`
- Wait for the error
- Edit the cluster `Worker` node pool with the proper value (of `region` this example).
- Delete the failed node and wait for a new one to come up as `Provisioning`

**Result**

The deleted node will stay in the cluster and won't go away even if we delete the `Worker` nodepool from the above steps.

**Expected Result**

Be able to delete the misconfigured node from the cluster or get it automatically removed from the node pool

**Screenshots**
<img width="983" alt="Screen Shot 2021-09-13 at 11 46 00 AM" src="https://user-images.githubusercontent.com/27659/133139546-987f8571-a0e4-4ec9-87ee-e6b921b858b8.png">

**Additional context**
https://github.com/rancher/rancher/issues/32373

izaac:
Rancher version: `v2.6.1-rc3`

I was able to remove a Node Pool with a failing agent and the agent got removed as expected and a new one was provisioned
Then I tried Editing a misconfigured Node Pool to fix the failed agent. 
  - Misconfigure a Node Pool
  - Wait for the node to fail
  - Edit the Node Pool with the right values.
  - Delete the node and a new one will be provisioned with the right values.

[...]

Issue #34542, opened 2021-08-30
[not always reproducible] RKE2 Nodes are not being removed from backend/provider when deleting infrastructure-provider clusters through rancher


slickwarren:
**Rancher Server Setup**
- Rancher version:v2.6.0-rc6
- Installation option (Docker install/Helm Chart):docker
   - If Helm Chart, Kubernetes Info:
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):
         Node Setup:
         Version:
- Proxy/Cert Details:self signed

**Information about the Cluster**
- Kubernetes version:v1.21.4/1.21.3 (RKE2)
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Infrastructure provider - Azure, DO

**Describe the bug**

Sometimes, nodes are not being deleted from the provider when the cluster is removed through rancher. 

**To Reproduce**
 
* provision a cluster using an infrastructure provider (previously observed with azure, digital ocean)
  * wait for the cluster to come to active
* delete the cluster through rancher
  * wait for rancher to mark the cluster as deleted
* through the provider you chose, ensure your node(s) from the cluster have been deleted

**Result**
sometimes, the nodes are not removed from the provider

**Expected Result**

all nodes should be removed when deleting an infrastructure-provider cluster through rancher

**Screenshots**

**Additional context**

* this has been observed on Azure and Digital Ocean with RKE2
* this has only been observed so far with multi-node clusters

thedadams:
A couple of notes on this:
1. If you create a cluster then delete it before the nodes finish provisioning, then the cluster will be deleted from Rancher and the nodes leftover in the cloud provider. (This will be fixed and should be tested as part of the validation of the fix).
2. As far as the "not always reproducible" status of the issue: previously if the deletion job (to remove the machine from the cloud provider) failed, it would be ignored and never retried. Now the job will at least retry and this issue should be resolved or at least occur much less often.

snasovich:
Suggesting to add improvements to cluster deletion to 2.6.1 release notes. Possibly as one bullet point for all issues mentioned on the PR linked here.

davidnuzik:
**My checks PASSED**
 
**Reproduction Environment:**

Rancher version: v2.6.0
Rancher cluster type: single-node docker install
Docker version: 20.10
 
Downstream cluster type: RKE2
Downstream K8s version: v1.21.4+rke2r2
 
[...]

Issue #33774, opened 2021-07-21
Clusters deleted immediately from Cluster Management page but not from Home page


sowmyav27:
**On master-head commit id: `b6bcfad`**

- Deploy an RKE1 cluster.
- Delete the cluster from cluster management page
- the cluster immediately goes away from this page list.
- Navigate to Home page in dashboard.
- Cluster is seen in `Removing` state - removing resources 

**Expected:**
Cluster should get removed after removing the resources. Behavior seen in Home page is the right way.

Can we defer the provisioning delete until management delete is done ?

cbron:
I think we should bump this one. Potential solution is to not delete the provisioning cluster object (in the backend) until the management cluster object is gone.

deniseschannon:
With the new cluster objects introduced with the new provisioning framework, there are multiple objects that need to be deleted and the UI is reading various cluster objects which are slowly being deleted.

deniseschannon:
Adding a finalizer could fix this, but it could trigger some status updates.

thedadams:
I believe this was fixed by https://github.com/rancher/rancher/pull/34501, but I will double-check before moving to "To Test"

thedadams:
This looks like it was indeed fixed by https://github.com/rancher/rancher/pull/34501

davidnuzik:
**My checks PASSED**
 
**Reproduction Steps:**

Not required.

-----
 
**Validation Environment:**

Rancher version: v2.6-head `a6e7905` pulled 9/9/21 10:37am Pacific
Rancher cluster type: single-node docker install
Docker version: 20.10
 
Downstream cluster type: rke1
Downstream K8s version: v1.21.4-rancher1-1
 
**Validation steps:**

1. Create an RKE1 cluster wait for it to fully come up (active state).
2. Delete the cluster from the cluster management page.
3. It goes into a removing state briefly. Go to the home page this state/status is in sync.
4. Come back to cluster management page wait for it to be removed. Once it is immediately go to home page.
5. It is indeed immediately removed from the home page.

thedadams:
This was reopened because the implementation is changing.

To clarify, deletion of the management cluster object (the one displayed on the Home page) and the provisioning cluster object (the one displayed from the "Cluster Management" page) should now be deleted within 5 seconds of each other regardless of RKE1 or RKE2 clusters.

davidnuzik:
**My checks PASSED**
 
**Reproduction Steps:**

Not required.

-----
 
[...]

PR #34761, opened 2021-09-15
Fix machine deletions

thedadams:
There are a few situations that caused machine deletions to be flakey:

1. There was a race with the machine deletion job and the deletion of
   the state and provision secrets. The state secret was being deleted
   by Kubernetes via an ownerReference. Now, this ownerReference is
   removed so that when the machine deletion job completes, the secrets
   are deleted ensuring they will be around for the machine deletion job.
2. The machine-orphan controller existed to clean up machines that were
   leftover after the CAPI cluster was deleted. This would cause
   problems with deletion, specifically if a the provisioning of a single
   node cluster failed. This controller has been removed, and a
   foregroundDeletion finalizer placed on the MachineDeployments which
   ensures Machines, MachineSets, and MachineDeployments are all cleaned
   up before the CAPI cluster object is removed. Since Machine objects
   are cleaned up before the CAPI cluster is deleted, there are no more
   orphaned machines for this controller to cleanup.
3. It was not possible to delete a node from a provisioned cluster if
   provisioning of the new node failed. This was caused by an issue with
   creating the deletion job when the creation job still existed. Now,
   when running the delete job, we ignore the existing applied objects
   so that the delete job will run if there is a failed creation job
   leftover.
4. Previously, when deleting a cluster from the UI, the cluster would
   disappear from the UI in different places inconsistently. This was
   caused by the two different cluster objects (management and
   provisioning) being deleted at different times. After this change,
   the two cluster objects will be deleted within 5 seconds of each
   other.

Issues:
https://github.com/rancher/rancher/issues/34713
https://github.com/rancher/rancher/issues/34542
https://github.com/rancher/rancher/issues/33774


pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -50,13 +51,11 @@ type handler struct {
 func Register(ctx context.Context, clients *wrangler.Context) {
 	h := &handler{
 		ctx: ctx,
-		apply: clients.Apply.
-			WithSetOwnerReference(true, true).
```
dramich, line 54: Why drop this?

thedadams: When this was set, the infrastructure machine object owned the secrets related to provisioning. When the infrastructure object was deleted, the Kubernetes garbage collector would go through and delete all of its dependents. This included the machine state secret.

Then, occasionally, the state secret wouldn't be mounted when the deletion job ran because it wouldn't exists. The job would fail all while keeping the machines in the cloud provider.

By removing this reference, the Kubernetes garbage collector will not clean up the state secret and the job will run successfully. After the job runs, this call will delete the objects that were applied: https://github.com/rancher/rancher/pull/34761/files#diff-11efa64a6ad5ead8500e161819dd3af981e47ea5546f456a25b7a7e18187c47aR240

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -1197,7 +1201,7 @@ func (p *Planner) generateSecrets(controlPlane *rkev1.RKEControlPlane, fullPlan
 	return controlPlane, secret, nil
 }
 
-func (p *Planner) ensureRKEStateSecret(controlPlane *rkev1.RKEControlPlane, fullPlan *plan.Plan) (string, plan.Secret, error) {
+func (p *Planner) ensureRKEStateSecret(controlPlane *rkev1.RKEControlPlane) (string, plan.Secret, error) {
```
nickgerace, line 1204: I might be missing it, but why was this argument added?

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -302,6 +322,16 @@ func (h *handler) patchStatus(obj runtime.Object, d data.Object, state rkev1.RKE
...
 	}
 
+	if state.JobComplete {
+		// Reset failureMessage and failureReason if they are not provided.
+		if _, ok := statusData["failureMessage"]; !ok {
+			statusData["failureMessage"] = ""
+		}
+		if _, ok := statusData["failureReason"]; !ok {
+			statusData["failureReason"] = ""
+		}
```
nickgerace, line 332: Should we delete the key from the map instead?

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -285,7 +299,13 @@ func (h *handler) run(obj runtime.Object, create bool) (runtime.Object, error) {
...
 	}
 
-	if err := h.apply.WithOwner(obj).ApplyObjects(objs...); err != nil {
+	applier := h.apply.WithOwner(obj)
+	if !create {
+		// If the infrastructure is being deleted, ignore previously applied objects.
+		// If creation failed, this will allow deletion to process.
+		applier = applier.WithIgnorePreviousApplied()
+	}
+	if err := applier.ApplyObjects(objs...); err != nil {
```
nickgerace, line 308: This makes sense with your above comment. Ran into a similar issue with Fleet for imported clusters having old namespaces get deleted after migration.

pkg/controllers/provisioningv2/cluster/remove.go
```diff
@@ -78,9 +100,12 @@ func (h *handler) doClusterRemove(cluster *v1.Cluster) (string, error) {
 			return "", err
 		}
 
-		_, err = h.mgmtClusterCache.Get(cluster.Status.ClusterName)
-		if !apierrors.IsNotFound(err) {
-			return fmt.Sprintf("waiting for cluster [%s] to delete", cluster.Status.ClusterName), nil
+		if h.isLegacyCluster(cluster) {
+			// If this is a legacy cluster (i.e. RKE1 cluster) then we should wait to remove the provisioning cluster until the v3.Cluster is gone.
```
nickgerace, line 104: Why don't we want to wait with a non-legacy cluster? There's still a corresponding v3 Cluster, right?

nickgerace (commented):
Reviewed for general flow, Go code, and high level

pkg/controllers/provisioningv2/cluster/remove.go
thedadams: Yes, but it depends on which cluster object will get deleted first.

For legacy clusters, provisioning cluster object gets cleaned up faster, so the provisioning cluster will wait for the management cluster.

For the new provisioning, the management cluster will get cleaned up faster, so the management cluster will wait for the provisioning cluster.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: Yes. If they don't exist, then the loop below won't overwrite them and the failure information will stay on status. We want to reset failure reason/message to empty if the failure has been resolved.

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -287,6 +287,8 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 				Name:        machinePoolName,
 				Labels:      machineDeploymentLabels,
 				Annotations: machinePool.MachineDeploymentAnnotations,
+				// This ensures that the CAPI cluster is not deleted before the machines are deleted.
+				Finalizers: []string{metav1.FinalizerDeleteDependents},
```
dramich, line 291: How does this block CAPI cluster deletion if we are adding the finalizer to the `MachineDeployment`

thedadams: I added it here because if the MachineDeployment has it then:
1. The MachineDeployment CAPI controller will put in on the MachineSet object. Then, since the MachineDeployment object has it, it will not be removed until all the MachineSet objects it owns are deleted.
2. If the MachineSet object has it then the MachineSet object won’t be removed until all the Machine objects are deleted.
3. The Machine objects will wait for all the infrastructure machine objects to be cleaned up before the Machine object gets removed.

This blocks the CAPI cluster object from being deleted because the controller will loop, look for MachineDeployments that it owns. If it finds any, it will wait to remove its finalizer. If it doesn’t find any, then the controller will remove the finalizer and the cluster object will delete.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -210,22 +215,31 @@ func (h *handler) OnRemove(_ string, obj runtime.Object) (runtime.Object, error)
...
-	job, err := h.jobs.Get(meta.GetNamespace(), getJobName(meta.GetName()))
+	job, err := h.jobs.Get(d.String("metadata", "namespace"), getJobName(infraName))
 	if err != nil {
 		return nil, err
 	}
 
-	if condition.Cond("Failed").IsTrue(job) || job.Status.CompletionTime != nil {
+	if job.Status.CompletionTime != nil {
+		// Delete objects owned by infrastructure object.
+		if err := h.apply.WithOwner(obj).ApplyObjects(); err != nil {
```
dramich, line 240: Do we have an explicit call we can change to? This is not intuitive that it will go delete things because it's empty.

thedadams: There doesn't seem to be any other calls. I added a more explanatory comment here.

dramich (approved):

snasovich (approved):
LGTM from general code perspective; given Dan's approval we should be good to go here :tada:

Issue #33583, opened 2021-07-13
Add built-in Harvester node driver


guangbochen:
**Is your feature request related to a problem? Please describe.**
Add Harvester node driver as built-in node driver of the rancher 2.6

**Describe the solution you'd like**
- Add Harvester node driver as built-in node driver of the rancher 2.6, it should be compatible for both RKE1 and RKE2
- Update the Harvester node driver release URL to `releases.rancher.com/harvester-node-driver/$TAG_VERSION` (currently using `harvester-node-driver.s3.amazonaws.com`)

**Describe alternatives you've considered**
N/A

**Additional context**
https://github.com/rancher/rancher/pull/33531 (merged for testing)
https://github.com/rancher/dashboard/issues/3374
https://github.com/rancher/dashboard/issues/3375

https://github.com/rancher/dashboard/pull/3881

guangbochen:
@futuretea will need to update the harvester node driver of replacing the DV with PVC by https://github.com/harvester/harvester/pull/1066

guangbochen:
close in favor of https://github.com/rancher/rancher/pull/34328

yasker:
Reopen the issue for validation.

slickwarren:
tested on v2.6-head () at the same time as [testing this dashboard issue](https://github.com/rancher/dashboard/issues/3467#issuecomment-915140731):
* it seems that, when the annotation from the linked comment is used, the cluster is not able to be used as a node driver. I don't think this is expected **Reopening for this issue**
  * repro steps: 
  * provision rancher with harvester enabled
  * import a harvester cluster through the virtualization manager
  * in cluster management, create a cluster using harvester node driver rke2
  * create a cloud credential from the imported cluster
    * resources from harvester will not populate; user is unable to select network, image, namespace from the UI (no matching options)

yasker:
@slickwarren I think node provisioning part is pending for the merge of https://github.com/rancher/dashboard/pull/3881 . cc @guangbochen

yasker:
Add https://github.com/rancher/dashboard/pull/3881 to this issue.

slickwarren:
https://github.com/rancher/rancher/issues/34891 related issue

slickwarren:
related issues:
https://github.com/rancher/rancher/issues/34948
https://github.com/rancher/rancher/issues/34946

slickwarren:
tested on v2.6.1-rc7 with harvester v0.3.0-rc1
[...]

PR #34765, opened 2021-09-16
[v2.6] Bump harvester node driver to 0.3.1

futuretea:
Backport #34764

This PR updates Harvester Node Driver to version v0.3.1, which fix volume delete

Related Issue:
https://github.com/rancher/rancher/issues/33583


guangbochen (approved):

nickgerace (approved):
Approved given it's a change in version in hash, but I do wonder if we can move these values to consts or vars at the top file of this file (or another) in the future.

()This PR still requires a commiter's review.)

dramich (approved):

Issue #34671, opened 2021-09-08
helm upgrade/delete intermittently fails with errors like "unable to create impersonator account: error setting up impersonation for user u-xxxxx..Operation cannot be fulfilled on clusterroles.rbac.authorization.k8s.io \"cattle-impersonation-u-8j72f\": the object has been modified; please apply your changes to the latest version and try again."


snasovich:
**Rancher Server Setup**
- Rancher version: 2.6.0
- Installation option (Docker install/Helm Chart): **?**
- Proxy/Cert Details: **?**

**Information about the Cluster**
- Kubernetes version: **?**
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): **?**

**Describe the bug**

Running `helm upgrade` / `helm uninstall` for an unknown application intermittently fails with errors like this:
```
error: unable to recognize "/tmp/kubectl_replace_486731274/resource.stdin": an error on the server ("unable to create impersonator account: error setting up impersonation for user u-8j72f: Operation cannot be fulfilled on clusterroles.rbac.authorization.k8s.io \"cattle-impersonation-u-8j72f\": the object has been modified; please apply your changes to the latest version and try again") has prevented the request from succeeding
```
```
Error: uninstall: Release not loaded: katana-api-dev: query: failed to query with labels: an error on the server ("unable to create impersonator account: error setting up impersonation for user u-8j72f: Operation cannot be fulfilled on clusterroles.rbac.authorization.k8s.io \"cattle-impersonation-u-8j72f\": the object has been modified; please apply your changes to the latest version and try again") has prevented the request from succeeding (get secrets)
```

**Additional context**

Created from the following comment: https://github.com/rancher/rancher/issues/34187#issuecomment-914575796

snasovich:
@andrew-landsverk-win , so far I'm not successful with reproducing it. Tried installing/uninstalling couple charts a few times via Helm v3.6.0 on a single-cluster RKE2 downstream cluster - all successful.
How often do you get these errors? Also, I'm wondering if you could share helm charts you're trying to install as well as details on your downstream cluster and k8s versions on both the downstream cluster and Rancher cluster itself.

andrew-landsverk-win:
[...]

PR #34766, opened 2021-09-16
Fix conflict errors for impersonation cluster role

cmurphy:
Without this patch, requests made on a downstream cluster through the
impersonation proxy can sometimes fail when the impersonation cluster
role fails to update. This is more likely to happen when there are a
high volume of requests. This change addresses the issue in two ways:

1) Reduce the need to update the cluster role by ensuring that the
   resources in the role rules are sorted so that equality comparison is
   meaningful.
2) Retry the update when a conflict is encountered so that a single
   conflict error isn't fatal to the request.

https://github.com/rancher/rancher/issues/34671

### DIFF
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -3,6 +3,7 @@ package impersonation
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"time"
 
 	"github.com/rancher/rancher/pkg/types/config"
@@ -14,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
@@ -27,8 +29,16 @@ type Impersonator struct {
 	clusterContext *config.UserContext
 }
 
-func New(user user.Info, clusterContext *config.UserContext) Impersonator {
-	return Impersonator{user: user, clusterContext: clusterContext}
+func New(userInfo user.Info, group string, clusterContext *config.UserContext) (Impersonator, error) {
+	user, err := getUser(clusterContext, userInfo, group)
+	if err != nil {
+		return Impersonator{}, err
+	}
+
+	return Impersonator{
+		user:           user,
+		clusterContext: clusterContext,
+	}, nil
 }
 
 func (i *Impersonator) SetUpImpersonation() (*corev1.ServiceAccount, error) {
@@ -167,17 +177,27 @@ func (i *Impersonator) createNamespace() error {
 
 func (i *Impersonator) checkAndUpdateRole(rules []rbacv1.PolicyRule) (*rbacv1.ClusterRole, error) {
 	name := ImpersonationPrefix + i.user.GetUID()
-	role, err := i.clusterContext.RBAC.ClusterRoles("").Controller().Lister().Get("", name)
-	if apierrors.IsNotFound(err) {
-		return nil, nil
-	}
+	var role *rbacv1.ClusterRole
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var err error
+		role, err = i.clusterContext.RBAC.ClusterRoles("").Controller().Lister().Get("", name)
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		if !reflect.DeepEqual(role.Rules, rules) {
+			role.Rules = rules
+			role, err = i.clusterContext.RBAC.ClusterRoles("").Update(role)
+			return err
+		}
+		return nil
+	})
 	if err != nil {
 		return nil, err
 	}
-	if !reflect.DeepEqual(role.Rules, rules) {
-		role.Rules = rules
-		return i.clusterContext.RBAC.ClusterRoles("").Update(role)
-	}
+
 	return role, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/impersonation/impersonation.go
```diff
@@ -323,3 +344,86 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
 	}
 	return ret, nil
 }
+
+func getUser(clusterContext *config.UserContext, userInfo user.Info, groupname string) (user.Info, error) {
```
snasovich, line 348: Nit, should be `groupName`

pkg/impersonation/impersonation.go
```diff
@@ -323,3 +344,86 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
...
+	u, err := clusterContext.Management.Management.Users("").Controller().Lister().Get("", userInfo.GetUID())
+	if err != nil {
+		return &user.DefaultInfo{}, err
+	}
+
+	groups := []string{"system:authenticated", "system:cattle:authenticated"}
+	if groupname != "" {
+		groups = append(groups, groupname)
```
snasovich, line 356: We're appending to groups and then using `insertSortedIfNotPresent` for it. Appending passed `groupname` may result in `groups` being no longer sorted and `insertSortedIfNotPresent` assumes passed slice is sorted.

Overall, I'm thinking if it would be simpler and less error-prone if we simply built slices by appending and then sorted and removed duplicates as the last steps (we could also do what `removeEmpty` does if applicable there too)? Just a thought, not necessary to change the implementation this close to code freeze.

pkg/impersonation/impersonation.go
```diff
@@ -323,3 +344,86 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
...
+	user := &user.DefaultInfo{
+		UID:    u.GetName(),
+		Name:   u.Username,
+		Groups: groups,
+		Extra:  extras,
+	}
+	return user, nil
+}
+
+func insertSortedIfNotPresent(item string, list []string) []string {
```
snasovich, line 385: I assume this is for inserting into an already sorted list. I don't think it's working as expected: https://play.golang.org/p/Yko9VTmAqMd

One of those cases where creating unit tests for this func would be really valuable to ensure it's working as expected.

snasovich (changes requested):

pkg/impersonation/impersonation.go

pkg/impersonation/impersonation.go
cmurphy: I decided to remove the insert function and simplified this

pkg/impersonation/impersonation.go
cmurphy: Thanks for catching this, I've simplified this using sort

pkg/impersonation/impersonation.go
```diff
@@ -323,3 +344,84 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
 	}
 	return ret, nil
 }
+
+func getUser(clusterContext *config.UserContext, userInfo user.Info, groupName string) (user.Info, error) {
```
dramich, line 348: I am concerned with how much processing we are adding in front of every API call. This also seems like we are trying to gather info from multiple places. Where is our source of truth for the user?

pkg/impersonation/impersonation.go
```diff
@@ -323,3 +344,84 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
...
+}
+
+func getUserExtras(userInfo user.Info) map[string][]string {
+	extras := userInfo.GetExtra()
+	if extras == nil {
+		return make(map[string][]string)
+	}
+	if _, ok := extras["username"]; ok {
+		extras["username"] = removeEmpty(extras["username"])
+		sort.Strings(extras["username"])
```
snasovich, line 411: I don't believe we need to sort here or below for `extras["principalid"]` as we sort again after adding extra item(s) to these lists in the calling func.

pkg/impersonation/impersonation.go
```diff
@@ -214,6 +234,7 @@ func (i *Impersonator) rulesForUser() []rbacv1.PolicyRule {
 	}}
 
 	if groups := i.user.GetGroups(); len(groups) > 0 {
+		sort.Strings(groups)
```
snasovich, line 237: I think it makes sense to add comments to all these sorts added to explain why they are needed. To someone not familiar with the reason it may seem we do it just so it looks nicer or something trivial.

snasovich (changes requested):

pkg/impersonation/impersonation.go
cmurphy: This function is the source of truth for both the role template controller and the proxy handler. The source of truth for groups is the `UserAttributes` lister. The source of truth for extras is the `User` lister plus the `user.Info` object. Combining these for extras is necessary to address this bug and to avoid regressing on https://github.com/rancher/rancher/issues/33912

pkg/impersonation/impersonation.go

pkg/impersonation/impersonation.go

pkg/impersonation/impersonation.go
cmurphy: We could remove all processing and revert to what we currently have but with retries to fix the conflict errors.

snasovich (approved):
LGTM, still need @dramich to approve

dramich (commented):
After an offline discussion we are going to simplify, trust the data from the token, and add a conflict error retry. There can be future improvements to this once some of the work from Prachi is merged to store all the info needed on userAttributes. 
https://github.com/rancher/rancher/pull/33268

snasovich (dismissed):

cmurphy:
Added a commit with tests, can either squash it, leave it separate or leave it out for now.

snasovich (dismissed):
Main code still LGTM, added some recommendations to improve unit tests added.

snasovich (approved):
LGTM, thank you for improving the tests added.
Also, I see we still have 2 commits, so I'm not sure if you're planning to squash them at merge time automatically or decided to keep them separate.

dramich:
> Also, I see we still have 2 commits, so I'm not sure if you're planning to squash them at merge time automatically or decided to keep them separate.

They should stay as 2 commits, one for the changes and one for the tests.

dramich (approved):

Issue #28776, opened 2020-09-07
Rancher crashes after adding a catalog containing a chart with missing name and version


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Provision a Rancher v2.4.8 instance with a single all role node custom cluster (I used https://github.com/superseb/tf-do-rancher2/)
2. Add a cluster-scoped catalog containing a chart with a missing name and version in the Chart.yaml file (I used https://github.com/axeal/rancher-catalog/tree/noversion)

**Result:**

Rancher containers enters a crash loop with the following error in the logs:

```
2020/09/07 12:09:33 [ERROR] ClusterCatalogController c-79crj/test-catalog [clusterCatalog] failed with : failed to update templates. Multiple error(s) occurred: [CatalogTemplateVersion.management.cattle.io "test-catalog-" is invalid: metadata.name: Invalid value: "test-catalog-": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')];
panic: interface conversion: interface {} is nil, not string

goroutine 9683 [running]:
github.com/rancher/rancher/pkg/api/customization/catalog.(*templateStore).extractVersionLinks(0xc00043d480, 0xc0045cc000, 0xc007db21b0, 0xc0065c5db8)
	/go/src/github.com/rancher/rancher/pkg/api/customization/catalog/store.go:59 +0x5bb
github.com/rancher/rancher/pkg/api/customization/catalog.GetTemplateStore.func1(0xc0045cc000, 0xc0009c6b40, 0xc007db21b0, 0xc005e40380, 0xc0074cefa8, 0xcfa401, 0x101)
	/go/src/github.com/rancher/rancher/pkg/api/customization/catalog/store.go:39 +0x42
github.com/rancher/rancher/vendor/github.com/rancher/norman/store/transform.(*Store).Watch.func1(0xc007db21b0, 0xc0074cefa8)
	/go/src/github.com/rancher/rancher/vendor/github.com/rancher/norman/store/transform/transform.go:58 +0x4e
[...]

PR #34787, opened 2021-09-17
Avoid a panic when a user adds a chart catalog containing a blank Helm chart

genexpr:
Relevant issue: rancher/rancher#28776 

This is relevant only for the old UI. A blank chart is one whose `Chart.yaml` is missing the `version` field. The `name` field is also required, but its lack is already properly handled.

### DIFF
--- a/pkg/api/norman/customization/catalog/store.go
+++ b/pkg/api/norman/customization/catalog/store.go
@@ -16,6 +16,7 @@ import (
 	managementschema "github.com/rancher/rancher/pkg/schemas/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/types/config"
+	"github.com/sirupsen/logrus"
 )
 
 type templateStore struct {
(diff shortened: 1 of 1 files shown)


StrongMonkey (dismissed):
LGTM! we should probably hold merging 2.6 PR as the original issue is not assigned to 2.6.1 milestone and also it has little impact since old UI was deprecated.

nickgerace (dismissed):
LGTM, though I wonder if we should add a trace log here.

StrongMonkey (dismissed):

Tejeev:
@nickgerace, did someone add the tracelog?  If not, can we please?  These are super useful for support

genexpr:
@Tejeev, will add trace logging.

pkg/api/norman/customization/catalog/store.go
```diff
@@ -59,7 +59,10 @@ func (t *templateStore) extractVersionLinks(apiContext *types.APIContext, resour
 			if v, ok := version.(map[string]interface{})["revision"].(int64); ok {
 				revision = strconv.FormatInt(v, 10)
 			}
-			versionString := version.(map[string]interface{})["version"].(string)
+			versionString, ok := version.(map[string]interface{})["version"].(string)
+			if !ok {
+				continue
+			}
```
nickgerace, line 65: Replying your overall PR comment @genexpr: I think our Trace log should not include specific variables because we are dealing with interfaces here. Entirely agree with your concern.

However, we should be careful when adding a "manual" trace log since we do not want to say what we _think_ is happening (unless we clarify our intent). We want to say what is _actually_ happening.

**Likely on the wrong track:**
```go
if !ok {
    logrus.Trace("[templateStore] version information not found in empty CatalogTemplateFieldVersion")
    continue
}
```

**Likely on the right track:**
```go
if !ok {
    logrus.Trace("[templateStore] versionString type assertion failed for CatalogTemplateFieldVersion")
    continue
}
```

_(You have more context to this change, so your message will likely be more "correct" and informative than the above.)_

pkg/api/norman/customization/catalog/store.go
```diff
@@ -52,15 +53,16 @@ func GetTemplateStore(ctx context.Context, managementContext *config.ScaledConte
...
 	if ok {
-		for _, version := range versionMap {
+		for _, version := range versionMaps {
 			revision := ""
 			if v, ok := version.(map[string]interface{})["revision"].(int64); ok {
 				revision = strconv.FormatInt(v, 10)
 			}
 			versionString, ok := version.(map[string]interface{})["version"].(string)
 			if !ok {
+				logrus.Trace("[templateStore] failed version type assertion for CatalogTemplateFieldVersion")
```
nickgerace, line 65: I think this is bit confusing in its wording since "version" is a bit general. Perhaps...
```go
logrus.Trace("[templateStore] failed type assertion for field \"version\" in CatalogTemplateFieldVersion object")
```

nickgerace (changes requested):

nickgerace:
@genexpr please squash your commits as well. I think it makes sense for this "single objective" PR.

nickgerace (approved):
Nice job :) @genexpr

StrongMonkey (approved):

nickgerace (approved):

kinarashah (approved):

StrongMonkey (approved):

Issue #34716, opened 2021-09-14
Services get deleted after importing Harvester to Rancher


guangbochen:
**To Reproduce**
1. Install rancher v2.6.0 using docker run
2. Install Harvester using the master-head ISO image.
3. Import the Harvester cluster to Rancher

**Result:**
Notice that the `harvester-pre-delete` job is triggered. After it completed, kubevirt services except for the virt-operator are deleted.

```
$ kubectl  get po -n harvester-system
NAME                                                    READY   STATUS    RESTARTS   AGE
harvester-7b68989d68-8hkjv                              0/1     Pending   0          16m
harvester-7b68989d68-n5tz7                              1/1     Running   0          16m
harvester-7b68989d68-xwzvq                              0/1     Pending   0          16m
harvester-network-controller-hc8t2                      1/1     Running   0          16m
harvester-network-controller-manager-56f795b468-gnmkg   1/1     Running   0          16m
harvester-network-controller-manager-56f795b468-hgnv9   1/1     Running   0          16m
harvester-webhook-7fb4f8f7f9-djkgm                      0/1     Pending   0          16m
harvester-webhook-7fb4f8f7f9-hrcf5                      0/1     Pending   0          16m
harvester-webhook-7fb4f8f7f9-r5vth                      1/1     Running   0          16m
virt-operator-669cfcc9d8-4gmkz                          1/1     Running   0          16m
```

**Additional context**
Current status of the managedchart:
```
$kubectl get managedchart -n fleet-local harvester -o yaml
apiVersion: management.cattle.io/v3
kind: ManagedChart
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #34798, opened 2021-09-18
Run fleet agent for local cluster in cattle-fleet-local-system

ibuildthecloud:


nickgerace:
Relevant issue: https://github.com/rancher/fleet/issues/531

pkg/data/dashboard/namespace.go
```diff
@@ -11,14 +12,26 @@ import (
 )
 
 func addCattleGlobalNamespaces(ctx context.Context, k8s kubernetes.Interface) error {
+	if features.Fleet.Enabled() {
+		_, err := k8s.CoreV1().Namespaces().Get(ctx, "fleet-local", metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
```
nickgerace, line 17: What if it isn't a `NotFound` error?

deniseschannon:
https://github.com/rancher/rancher/issues/34716

pkg/data/dashboard/namespace.go
ibuildthecloud: Yeah that's not handled, I'll change that.

pkg/data/dashboard/namespace.go
```diff
@@ -11,14 +12,26 @@ import (
...
+			_, err = k8s.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "fleet-local",
+				},
+			}, metav1.CreateOptions{})
+		}
+		if err != nil {
```
nickgerace, line 24: this needs to be `else if`

nickgerace (changes requested):

pkg/data/dashboard/namespace.go
ibuildthecloud: No, because the err is reassigned in the body.  If it's `else if` the err in the body of the iff won't be handled.

pkg/data/dashboard/namespace.go
```diff
@@ -11,14 +12,26 @@ import (
 )
 
 func addCattleGlobalNamespaces(ctx context.Context, k8s kubernetes.Interface) error {
+	if features.Fleet.Enabled() {
+		_, err := k8s.CoreV1().Namespaces().Get(ctx, "fleet-local", metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			_, err = k8s.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
```
StrongMonkey, line 18: err not handled here?

nickgerace: This is fine because it's reset and returned. We know the original error is not found, so we reassign and return that

StrongMonkey: oh it is assigned to err in the next if

StrongMonkey (approved):

pkg/data/dashboard/namespace.go
nickgerace: Ah I see, it is the same as here: https://github.com/rancher/rancher/pull/34798/files#r712303455

nickgerace (approved):

Issue #34801, opened 2021-09-18
Failed to download external drivers with Rancher proxy set


gitlawr:
**Rancher Server Setup**
- Rancher version: v2.6.0
- Installation option (Docker install/Helm Chart): docker
- Proxy/Cert Details: In a private network that can only connect to the internet through a proxy

**Describe the bug**

**To Reproduce**
1. Install Rancher in a private network that can only connect to the internet through a proxy. Set http_proxy via env vars. e.g.,
```
$ sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 \
      -e HTTP_PROXY="http://192.168.0.1:3128" \
      -e HTTPS_PROXY="http://192.168.0.1:3128" \
      -e NO_PROXY="localhost,127.0.0.1,0.0.0.0,10.0.0.0/8,cattle-system.svc,192.168.10.0/24,.svc,.cluster.local,example.com" \
      --privileged rancher/rancher:v2.6.0
```
2. Go to Global -> Cluster Management -> Drivers -> Node Drivers
3. Activate the Harvester node driver
4. Create a new cluster using RKE2 Harvester provider.

**Result**
The provisioning failed.

Logs from the machine-provision pod in the fleet-default namespace:
```
Downloading driver from https://releases.rancher.com/harvester-node-driver/v0.3.0/docker-machine-driver-harvester-amd64.tar.gz
ls: cannot access 'docker-machine-driver-*': No such file or directory
downloaded file failed sha256 checksum
download of driver from https://releases.rancher.com/harvester-node-driver/v0.3.0/docker-machine-driver-harvester-amd64.tar.gz failed
```

**Expected Result**
Be able to download the driver through the http proxy.

noahgildersleeve:
Blocked by https://github.com/rancher/rancher/issues/34891 .

yasker:
No longer blocked due to https://github.com/rancher/rancher/issues/34891 has been fixed.

yasker:
@noahgildersleeve Can you redo the test for this one?

noahgildersleeve:
I tried to redo the test, but I wasn't able to get the Harvester imported into Rancher via virtualization management. I ran the `kubectl` commands and they seemed to be successful, but it wouldn't leave the pending state. I need to import it in order to deploy the RKE2 cluster. This hasn't been an issue in other configurations, but I'm not sure if it's related to it being tested in a virtual machine with a private bridge for testing airgap, or something else.
[...]

PR #34827, opened 2021-09-20
Pass HTTP proxy environment to machine job

thedadams:
The machine job would fail if Rancher was running behind a proxy. This
change passes the proxy environment variables to the job if they are
set.

Issue:
https://github.com/rancher/rancher/issues/34801

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
@@ -2,6 +2,7 @@ package machineprovision
 
 import (
 	"fmt"
+	"os"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -85,6 +86,12 @@ func (h *handler) getArgsEnvAndStatus(typeMeta meta.Type, meta metav1.Object, da
 		Data: map[string][]byte{},
 	}
 
+	if os.Getenv("HTTP_PROXY") != "" || os.Getenv("HTTPS_PROXY") != "" {
```
paynejacob, line 89: is there a reason to not copy them even if they are empty?

thedadams: Only not having blank environment variables passed to the job.

snasovich: > Only not having blank environment variables passed to the job.
Not sure what exactly do you mean by this.

It's a bit weird that we check if any of two is not empty but then setting all 3 as it leaves the possibility of 1 or 2 of them be empty. I would say we either set them regardless or set each only if that specific value is not empty.

thedadams: Updated.

paynejacob (approved):

snasovich (approved):

kinarashah (approved):

PR #31902, opened 2021-04-02
[2.5]Add support for cluster-level registry(rke) for shell and helm

StrongMonkey:

PR #34846, opened 2021-09-21
populate server-version setting with env var for cluster agent

jakefhyde:
Downstream clusters need to have the same `server-version` setting as their upstream in order to view the correct feature charts version for the current rancher installation. The `CATTLE_SERVER_VERSION` environment variable is now passed down through an environment variable when the cluster agent is deployed to ensure that the downstream cluster agent uses the correct version.

Related Issue: https://github.com/rancher/rancher/issues/34875

### DIFF
--- a/pkg/systemtemplate/import.go
+++ b/pkg/systemtemplate/import.go
@@ -44,6 +44,7 @@ type context struct {
 	PrivateRegistryConfig string
 	Tolerations           string
 	ClusterRegistry       string
+	ServerVersion         string
 }
 
 var (
@@ -122,6 +123,7 @@ func SystemTemplate(resp io.Writer, agentImage, authImage, namespace, token, url
 		PrivateRegistryConfig: privateRegistryConfig,
 		Tolerations:           tolerations,
 		ClusterRegistry:       clusterRegistry,
+		ServerVersion:         settings.ServerVersion.Get(),
 	}
 
 	return t.Execute(resp, context)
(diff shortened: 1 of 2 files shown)


StrongMonkey:
@rmweir I think this will have a problem in hosted rancher if we copied the setting. The downstream cluster will have the same rancher version because we copied the setting. Ideally we should pass down the setting to an environment variable in cluster-agent . cc @ibuildthecloud  I think we should stop doing this handler and expose everything through cluster-agent....

rmweir (changes requested):
@StrongMonkey makes a good point. This will need to be discussed before we move forward.

rmweir:
Maybe we could set the default based on upstream value?

StrongMonkey:
We need to stop syncing the setting from upstream to downstream. Instead we should pass down the server version to an environment variable into cluster-agent. Then controller should try to read the environment variable first, and if it is not set it should fallback to settings. There is a similar PR for system-default-registry https://github.com/rancher/rancher/pull/31902/files

rmweir (approved):
LGTM

StrongMonkey (approved):

rmweir:
Should be noted that this would cause the the value to become an env var on the downstream even if it wasn't set via env var for the upstream, which is unlikely to be overwritten unless someone using the downstream changed or removed the envar on the downstream host. I think @jakefhyde and I concluded that this is probably fine for server version.

cbron (approved):

cbron:
Holding off on merging this until we figure out https://github.com/rancher/dashboard/issues/4276

PR #34851, opened 2021-09-21
Add simple password length and username comparison validation

jakefhyde:
Added password validation to make sure the password is at least 12 characters (controlled by `min-password-length` setting), and that the username and password are not equal.

Related Issue: https://github.com/rancherlabs/rancher-security/issues/843


paynejacob (approved):

jakefhyde:
Tests are passing now, I've run ci locally on my machine so if the build fails it is likely a drone issue.

paynejacob (approved):

tests/integration/suite/conftest.py
```diff
@@ -198,8 +198,8 @@ def user_factory(admin_mc, remove_resource):
     """
     def _create_user(globalRoleId='user'):
         admin = admin_mc.client
-        username = random_str()
-        password = random_str()
+        username = random_str() + "username"
+        password = random_str() + "password"
```
cbron, line 202: Whats the point of this ?

pkg/auth/api/user/user_actions.go
```diff
@@ -170,3 +188,22 @@ func (h *Handler) refreshAttributes(actionName string, action *types.Action, req
...
+	if passLen < minPassLen {
+		return errors.Errorf("password must be at least %v characters", minPassLen)
+	}
+
+	if user == pass {
+		return errors.New("username and password are too similar")
+	}
+
+	return nil
+}
```
cbron, line 209: Can you write an integration test for this that errors out ?

pkg/auth/api/user/user_actions.go
```diff
@@ -134,6 +138,20 @@ func (h *Handler) setPassword(actionName string, action *types.Action, request *
 		return errors.New("Invalid password")
 	}
 
+	userID := request.Request.Header.Get("Impersonate-User")
```
cbron, line 141: Is this not in the userData ? even the userID ?

pkg/auth/api/user/user_actions.go
```diff
@@ -134,6 +138,20 @@ func (h *Handler) setPassword(actionName string, action *types.Action, request *
...
+	if userID == "" {
+		return errors.New("can't find user")
+	}
+
+	user, err := h.UserClient.Get(userID, v1.GetOptions{})
```
cbron, line 146: I'm surprised we don't use a userLister in this file. Maybe we can't risk the data not being up to date.

pkg/auth/api/user/user_actions.go
```diff
@@ -170,3 +188,22 @@ func (h *Handler) refreshAttributes(actionName string, action *types.Action, req
...
+func getMinimumPasswordLength() int {
+	return settings.PasswordMinLength.GetInt()
+}
+
+// validatePassword will ensure a password is the minimum required length, and that the password does not contain the username.
+func validatePassword(user string, pass string, minPassLen int) error {
+	passLen := len(pass)
+
+	if passLen < minPassLen {
+		return errors.Errorf("password must be at least %v characters", minPassLen)
```
cbron, line 201: Does this get properly displayed to user ? Is there any UI work that would need to get done ?

tests/integration/suite/conftest.py
jakefhyde: This is to ensure that username and password really are distinct. Although the chances of the following being the same are extremely low, the time segment of a username/password is almost always going to be the same because of int rounding. I'm also almost certain that this did happen while I was testing, but I don't have the error message anymore.
https://github.com/rancher/rancher/blob/7dcdd26018d8815f4dff7db1d5a318141b3f77a4/tests/integration/suite/common.py#L7-L8

pkg/auth/api/user/user_actions.go
jakefhyde: It is, as the `UserFieldUsername` I'll update.

pkg/auth/api/user/user_actions.go
jakefhyde: The error is passed back in a response to the UI. It may make sense to capitalize the error, although I'm not sure if we should format from backend, frontend, etc., may need UI opinion.

cbron: Follow common convention in backend code is fine, I was primarily concerned that the error already displays correctly around the login box.

jakefhyde: error-format
The error displays below the continue button (and appears to resize correctly on chrome at least).

cbron: Lets capitalize password, then looks good

jakefhyde: Ok, I'll fix capitlization and I'm adding here so that the behavior is the same when creating users in "Users and Authentication".

https://github.com/rancher/rancher/blob/1874d541a26721d3ee9d657bf2c93092b1fa54ce/pkg/auth/api/user/user_store.go#L109-L117

cbron: ahh good catch, you should double check if anything else sets it, should be able to track back via code.

paynejacob (approved):

paynejacob (approved):

pkg/auth/api/user/user_actions.go
```diff
@@ -9,9 +9,9 @@ import (
 	"github.com/rancher/norman/parse"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/rancher/pkg/auth/providerrefresh"
-	"github.com/rancher/rancher/pkg/auth/settings"
 	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/settings"
```
snasovich, line 14: Any reason this got moved in imports? I believe it was ordered properly before the change.

pkg/auth/api/user/user_actions.go
```diff
@@ -170,3 +180,22 @@ func (h *Handler) refreshAttributes(actionName string, action *types.Action, req
 func (h *Handler) userCanRefresh(request *types.APIContext) bool {
 	return request.AccessControl.CanDo(v3.UserGroupVersionKind.Group, v3.UserResource.Name, "create", request, nil, request.Schema) == nil
 }
+
+func getMinimumPasswordLength() int {
```
snasovich, line 184: I'd say introducing a separate func for this is overkill.

pkg/auth/api/user/user_actions.go
```diff
@@ -170,3 +180,22 @@ func (h *Handler) refreshAttributes(actionName string, action *types.Action, req
...
+	return settings.PasswordMinLength.GetInt()
+}
+
+// validatePassword will ensure a password is the minimum required length, and that the password does not contain the username.
```
snasovich, line 188: The comment "the password does not contain the username" doesn't match what the func is doing - it's checking if they are not exactly the same.

pkg/auth/api/user/user_actions.go
```diff
@@ -170,3 +180,22 @@ func (h *Handler) refreshAttributes(actionName string, action *types.Action, req
...
+func validatePassword(user string, pass string, minPassLen int) error {
+	passLen := len(pass)
+
+	if passLen < minPassLen {
+		return errors.Errorf("Password must be at least %v characters", minPassLen)
+	}
+
+	if user == pass {
+		return errors.New("Username and password are too similar")
```
snasovich, line 197: The message is rather vague as well. Suggest switching it to "Password cannot be the same as username".

pkg/auth/api/user/user_actions_test.go
```diff
@@ -0,0 +1,81 @@
...
+	tests := []struct {
+		name       string
+		username   string
+		password   string
+		expectsErr bool
+	}{
+		{
+			name:       "password too short",
+			username:   "admin",
+			password:   "admin",
```
snasovich, line 17: Ideally, we don't want these equal here - just have a short password as currently it also matches case when username and password match (making it the same case as below)

pkg/auth/api/user/user_actions_test.go
```diff
@@ -0,0 +1,81 @@
...
+			expectsErr: true,
+		},
+		{
+			name:       "username equals password min length",
+			username:   "passwordpass",
+			password:   "passwordpass",
+			expectsErr: true,
+		},
+		{
+			name:       "password almost contains username begin",
```
snasovich, line 27: These whole "almost contains" should be replaced with a single "almost match" case.

pkg/auth/api/user/user_actions_test.go
```diff
@@ -0,0 +1,81 @@
...
...+HZ8]p%[]DVSbqXpUE]k.[kKp<X^Tsx22d$[.$m>>w==%26j;K\\*eMwH<3yc,w`d&]y7hh{VpMDJNxK#+-6TvKba;^E6gd'HGD!<2SfG8>)uV83wzc)s_j!agMPVq5'$;8q_~H;s.;[?N($r.s#!Q}>8Q:)cezSZwV+-f2&VN{R:nbe3,#8mJ=vJ-r'~g_b<X<Fjj`e(NEaaW93\"ML5/5_&=\\]}xYgg7x%z8NSk\"W}m%-CnkcBn(L7x499Am@93M2y6Y864;Wsm~9X9M6V!4mzA\\#8Z_wuA;B[H@`>+`-B'k!G2}u/WA)!4Fu_)E-fF\\Gt'z\\Q7uys@htugjrwQN5Jb6}rZS3&M;W#8J$.WUrHq+D<BE4S=3aR'p}FCpvUydV(*[4GgGvh;h(K&']zytSDvPhjQ#&[LvCwg>N)(~cq2,u;PW@'M;mQthb\\!W)W]'\"<vtT{8ABc<CM.!b$*zWA3+fQCEh)Z3Yc?z)qWvQdAU;))@zN+`gSPb<Pxnnh/&wA/B~q2')8a4,`=~]`WwHQ<TS5v5ehqhX^fctr~~j.me(+Av'jU<YKPEzkyG;gZ:$ApRTyLVn8x^h]C=-`>H#]u89ujXjZ~mf6T7Kw#z!PSuQE.jS@k`%8Vq!U`DS'9F.D:,WC]~:wDH-gmy%<A4*]vS\\&LJ[QgfG,`S;pz92tAv$Dkr}tW~Kr@<f[jrrUk/<TvEm$xRT5:_Wg&:/LLd$s/KfabHr6?@[<p\"bx!&%#(h\\j3*+Fuu5jLMn=C&&4~$g:3Su.mR%^\"G~\\gpaUDc)s(Ff;jP<';j?tm?2r_Pv~s~?y>,TNZa42?vC;MACq*\"r[8`!SK>][>E~DQ!d&J-{yE>x(v^x<3;['#N:%q-m9MJa]Kh;*y/]T?(;y9`NVsAC\"RL]Zn}uGcQup}t&^sU_kfb*n=gxd+V%&chJQLGqZ`!*qn&7KuKb[E6Zm6W/PkbB7S7q]}mytFH4dLu_PYA7;L)~_Udm.M/,^BuK`n,e+[*jT}@Xz=6~L.7]Y;4SV2\"wP)(b$4%U/uJL9k6GJp3.W^B2J6FYq/{EE<D>*WU5`48aXkav\\?/wWj~+X>8:X;2g\\b'F`\\P'h=pP~y>CdqM5u\"wRc(FvdSx\\E*:[s#,*8y~Ur;Kb?(_\\4H<RJts?@H)fGT?$$7.e/+\"g)N/Cw>a\".",
```
snasovich, line 59: I think these 2 "very large" cases are certainly overkill - we already cover really well with "large" cases as these cases don't cover any different execution paths.

snasovich (changes requested):

pkg/auth/api/user/user_actions.go
jakefhyde: Ah, one is from `pkg/auth/settings`, the other is from `pkg/settings`. Took me a good bit staring at that to figure it out.

pkg/auth/api/user/user_actions_test.go
jakefhyde: I'm just going to replace these with a generate function copied from here:
https://github.com/rancher/rancher/blob/eff74fb877091c2d1651f6f8acaf831235e0a464/pkg/data/management/resetpassword.go#L86-L100 
Normally I'm not too fond of nor deterministic test data, but I think it's fine in this case, we can just update the generate function if we want to ensure different characters are present in the future, or make a testdata folder for really long passwords if determinism becomes important. Also, do we have an official upper bound on passwords? I've tested with a max of 2048 characters which is what the very large input is, but I can cap it at whatever the official supported password length is if it exists.

Issue #34833, opened 2021-09-20
[Hosted Rancher, Monitoring V1, Alerting V1] CPU spike observed when Monitoring / Alerting installed in downstream Rancher's local cluster


aiyengar2:
**Rancher Server Setup**
- Rancher version: v2.5.9 Super Rancher cluster, v2.6.0 Downstream Rancher cluster
- Installation option (Docker install/Helm Chart): Helm Chart for both
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): k3s for both
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: `v1.20.10+k3s1` for both
- Cluster Type (Local/Downstream): Impacted cluster is local to the Downstream Rancher, downstream to the Super Rancher cluster.
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Imported (k3s)

**Describe the bug**

When Monitoring V1 + Alerting V1 are enabled on a local cluster which itself is added as a downstream cluster on another Rancher instance, buggy behaviors during reconcile result in a spike in CPU / memory usage across nodes in the local cluster.

This seems to be caused by a race between the managing Rancher cluster and the Rancher cluster that is being managed since one thinks Monitoring / Alerting should be enabled while the other disagrees.

This is not seen in Rancher 2.5.9 managing Rancher 2.5.9; it is only seen in Rancher 2.5.9 managing Rancher 2.6.0.

**To Reproduce**
 

#### Note on Terminology

To replicate a Hosted Rancher setup, we will have two Ranchers:
- a Super Rancher cluster, which will manage other Ranchers. This will run Rancher 2.5.9
- a Downstream Rancher cluster, which will be added to the Super Rancher cluster as a downstream. This will run Rancher 2.6.0

I will refer to the Super Rancher cluster's local cluster as **superlocal**.

I will refer to the Downstream Rancher cluster's local cluster as **local**.

Both will run the last stable k8s 1.20 release: `v1.20.10+k3s1`

#### Note on Required Resources

I used 2 8GB_4vCPU Digital Ocean instances, where the first instance was used for the superlocal cluster (all roles) and the second instance was used for the local cluster.

I also provisioned DNS records that point to the superlocal host and the local host so that I could use it as the HOSTNAME below.

#### Setup Super Rancher Cluster

- Provision a node
[...]

Issue #34837, opened 2021-09-21
Hosted clusters are not deployed on Rancher Hosted setup after upgrade


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `ba43472` tenant on 2.5.9
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): k3s on k3s
- Proxy/Cert Details: rancher generated

**Information about the Cluster**
- Kubernetes version: v1.20.10+k3s1
- Cluster Type (Local/Downstream): both downstream tenant and hosted rancher local

**Describe the bug**
Hosted clusters are not deployed on Rancher Hosted setup

**To Reproduce**

- Deploy EKS v2/AKS v2
- Clusters are stuck in `Provisioning` state with message `Waiting on eks crd to be initialized`

<img width="432" alt="Screen Shot 2021-09-20 at 11 55 59 PM" src="https://user-images.githubusercontent.com/26032343/134125359-07e79e7d-ff6f-4aad-bef8-52b19b6bc8fc.png">

- Rancher server logs:
```
2021/09/21 06:56:46 [ERROR] error syncing 'c-cnm8d': handler aks-operator-controller: waiting aks crd to be initialized, cluster: c-cnm8d, handler cluster-provisioning-operator: can't find system project, requeuing
--
Mon, Sep 20 2021 11:56:47 pm | 2021/09/21 06:56:47 [ERROR] error syncing 'c-fx4vq': handler eks-operator-controller: waiting eks crd to be initialized, cluster: c-fx4vq, handler cluster-provisioning-operator: can't find system project, requeuing
```

sowmyav27:
**On 2.5.9 hosted rancher running 2.6-head commit id: `10dfee4` tenant rancher**

Provisioning of GKE/EKS/AKS v2 clusters work.

rodriguezmelian:
The solutions is Create a NameSpace cattle-fleet-system if you don´t have in your Cluster.

PR #34860, opened 2021-09-22
Create default and system projects on startup

thedadams:
Currently in v2.5.9, the system and default projects will not be created
in the hosted Rancher setup. Therefore, if the tenant cluster is
upgraded from v2.5.9 to v2.5.10 or to v2.6+, the system and default
projects will not exist. After this change, the system and default
projects will be created on startup if they do no exist.

Issue:
https://github.com/rancher/rancher/issues/34837


pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -271,36 +264,30 @@ type mgr struct {
 }
 
 func (m *mgr) createDefaultProject(obj runtime.Object) (runtime.Object, error) {
-	return m.createProject(project.Default, v32.ClusterConditionconditionDefaultProjectCreated, obj, defaultProjectLabels, defaultProjects)
```
cmurphy, line 274: > Currently in v2.5.9, the system and default projects will not be created
in the hosted Rancher setup

If the project is not created, then what causes this condition `ClusterConditionconditionDefaultProjectCreated` to be true?

thedadams: In 2.5.9 and in the hosted Rancher setup, the local cluster won't have a `creatorID`. So this code will set the condition to `true`: https://github.com/rancher/rancher/blob/release/v2.5.9/pkg/controllers/management/auth/project_cluster_handler.go#L269-L273

cmurphy (dismissed):

snasovich (dismissed):
LGTM based on offline conversation

dramich:
@rmweir This seems familiar, didn't we already do work around this?

dramich (commented):
There are a lot of changes in here but only one line looks to be important, can we split these out to another commit (or a cleanup pr)?
And how does this change fix the issue?

thedadams:
@dramich Separated commits and updated description.

snasovich (dismissed):

pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -271,36 +264,30 @@ type mgr struct {
...
 
 func (m *mgr) createSystemProject(obj runtime.Object) (runtime.Object, error) {
-	return m.createProject(project.System, v32.ClusterConditionconditionSystemProjectCreated, obj, systemProjectLabels, systemProjects)
+	return m.createProject(project.System, v32.ClusterConditionconditionSystemProjectCreated, obj, systemProjectLabels)
 }
 
-func (m *mgr) createProject(name string, cond condition.Cond, obj runtime.Object, labels labels.Set, projectMap map[string]bool) (runtime.Object, error) {
-	return cond.DoUntilTrue(obj, func() (runtime.Object, error) {
+func (m *mgr) createProject(name string, cond condition.Cond, obj runtime.Object, labels labels.Set) (runtime.Object, error) {
+	return cond.Do(obj, func() (runtime.Object, error) {
```
dramich, line 275: Dropping the `DoUntilTrue` makes the condition pointless since it's not used anywhere else, and it also protects from cache misses and needlessly hitting the API. 

Have we considered doing something similar to the token cleanup and doing a one time drop of the condition on upgrade?

thedadams: Yes. I considered the setting the condition(s) to `Unknown` when the `projectLifecycle` controller is created, but that would introduce error handling where there is currently none. You indicated in a previous change I made that introducing error handling where there is none close to a release is a bad idea. I personally prefer this approach.

dramich: Not in controller creation, you are correct that's a bad idea. After rancher is running, similar to token cleanup i.e. https://github.com/rancher/rancher/pull/33849/files

dramich: Or https://github.com/rancher/rancher/blob/release/v2.6/pkg/multiclustermanager/app.go#L220 depending on what you need to do this.

thedadams:
@StrongMonkey On this one, I wanted your review on the moving of the fleet namespaces (the middle commit) to make sure we weren't breaking anything there.

Remember that in the hosted Rancher setup, the default admin is restricted so they don't have access to the local cluster.

pkg/controllers/managementuser/rbac/namespace_handler.go
```diff
@@ -88,7 +88,14 @@ func (n *nsLifecycle) Remove(obj *v1.Namespace) (runtime.Object, error) {
...
 func (n *nsLifecycle) syncNS(obj *v1.Namespace) (bool, error) {
 	// add fleet namespace to system project
-	if isFleetNamespace(obj) {
+	if isFleetNamespace(obj) &&
+		// If this is the local cluster, then only move the namespace to ths system project if the projectIDAnnotation is
+		// empty or beings with "local" (i.e. not c-). If the projectIDAnnotation begins with something other than "local"
+		// then it is likely that local cluster is the tenant cluster in a hosted Rancher setup and the namespace belongs to
+		// the system project for the cluster in the host cluster. Moving it here would only cause the namespace to be
+		// continually moved between projects forever.
+		(n.m.clusterName != "local" || obj.Annotations[projectIDAnnotation] == "" || strings.HasPrefix(obj.Annotations[projectIDAnnotation], "local")) {
```
StrongMonkey, line 97: Is this because we have two controllers running for the same namespace, one is at local cluster and one is at downstream rancher so it will move project back and forth?

StrongMonkey: And in this case, we always make sure the namespace is assigned to the system project in downstream rancher?

thedadams: Yes, that is correct.

StrongMonkey (dismissed):

pkg/rancher/migrations.go
```diff
@@ -86,3 +87,19 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
 
 	return err
 }
+
+func forceSystemAndDefaultProjectCreation(clusterClient v3.ClusterClient) error {
```
snasovich, line 91: Would be nice to add why we're doing what this func is doing. Something along the lines of explanation given here I guess: https://github.com/rancher/rancher/pull/34860#discussion_r714295913.
This would also explain why this logic is in something named `migrations.go`

snasovich (approved):
One "nice to have" comment to address, otherwise LGTM

pkg/rancher/migrations.go
```diff
@@ -86,3 +87,22 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
 
 	return err
 }
+
+// forceSystemAndDefaultProjectCreation will set the correcsponding conditions on the local cluster object,
+// if it exists, to Unknown. This will force the corresponding controller to check that the projects exist
+// and create them, if necessary.
+func forceSystemAndDefaultProjectCreation(clusterClient v3.ClusterClient) error {
+	localCluster, err := clusterClient.Get("local", metav1.GetOptions{})
```
dramich, line 95: We are going to do this on every startup and leader change which I don't think is the intention. forceUpgradeLogout sets a value on a configmap so it's a 1 time operation which I think this should also be.

thedadams: Done.

pkg/rancher/migrations.go
```diff
@@ -76,13 +99,39 @@ func forceUpgradeLogout(configMapController controllerv1.ConfigMapController, to
...
+
+	v32.ClusterConditionconditionDefaultProjectCreated.Unknown(localCluster)
+	v32.ClusterConditionconditionSystemProjectCreated.Unknown(localCluster)
+
+	_, err = clusterClient.Update(localCluster)
+	if err != nil {
+		return err
 	}
 
+	cm.Data[projectsCreatedKey] = "true"
```
dramich, line 135: The configmap isn't being saved

thedadams: Dang it. You're right. Updated.

pkg/app/utils.go
```diff
@@ -47,27 +50,29 @@ func EnsureAppProjectName(userNSClient v1.NamespaceInterface, ownedProjectID, cl
 		}
 	}
 
-	// move Namespace into a project
 	expectedAppProjectName := fmt.Sprintf("%s:%s", clusterName, ownedProjectID)
-	appProjectName := ""
-	if projectName, ok := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]; ok {
-		appProjectName = projectName
+	appProjectID := deployNamespace.Annotations[nslabels.ProjectIDFieldLabel]
+	if appProjectID != "" && slice.ContainsString(strings.Split(settings.SystemNamespaces.Get(), ","), deployNamespace.Name, nil) {
```
dramich, line 55: Should we be using the setting or the annotation since some are added dynamically like fleet

aiyengar2:
This PR partially resolves https://github.com/rancher/rancher/issues/34833 as well since the `EnsureAppProjectName` changes prevent the cattle-prometheus namespace from being flip-flopped too.

pkg/app/utils.go
thedadams: Is there an annotation that would say that this namespace is a system namespace?

dramich: Yes, `management.cattle.io/system-namespace`

dramich: @StrongMonkey How are we marking fleet namespaces as system namespaces

StrongMonkey: We check if label `fleet.cattle.io/managed` is true for namespaces. All the generated namespaces from fleet have this label.
https://github.com/rancher/rancher/blob/4b967ddcf26f87aa5ce1bdf5932d9a1dd5cbbe92/pkg/controllers/managementuser/rbac/namespace_handler.go#L172

thedadams:
After our offline discussion, I made the changes we talked about and tested that:
1. Creating a hosted cluster from the downstream/tenant cluster does not move the cattle-system namespace.
2. Creating an app (in explorer and manager) doesn't touch the system namespaces, including the fleet ones.

pkg/app/utils.go
thedadams: Updated.

dramich (approved):

aiyengar2:
https://github.com/rancher/rancher/issues/34889, https://github.com/rancher/rancher/issues/34746

Issue #34824, opened 2021-09-20
"unable to create impersonator account ... serviceaccounts not found" error after upgrade from 2.5.9 to 2.6.0


cmurphy:
**Rancher Server Setup**
- Rancher version: 2.6.0
- Installation option (Docker install/Helm Chart): ?
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): ?
- Proxy/Cert Details: ?

**Information about the Cluster**
- Kubernetes version: ?
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**Describe the bug**

From https://github.com/rancher/rancher/issues/34671#issuecomment-922883807:

> Not sure if that's exactly the same issue, but it may be:

![image](https://user-images.githubusercontent.com/1220710/134045468-8e83d175-fb17-470d-a06b-795b0eeb88b5.png)

> I just upgraded a Rancher cluster from 2.5.9 to 2.6.0. It now fails to upgrade the agent in the downstream cluster.

> No keycloak is in use, but we use FreeIPA authentication, if that matters.

**To Reproduce**
 

**Result**

**Expected Result**

**Screenshots**

**Additional context**

Created from issue https://github.com/rancher/rancher/issues/34671#issuecomment-922883807. May be related to https://github.com/rancher/rancher/issues/34187.

ankit-ls:
Experiencing the same issue after `2.6.0` upgrade.

After downloading the kubeconfig, and running `kubectl` the following error is being observed:
```
$ k get nodes
Error from server (InternalError): an error on the server ("unable to create impersonator account: error setting up impersonation for user u-qwerty123: failed to get secret for service account: cattle-impersonation-system/cattle-impersonation-u-qwerty123, error: serviceaccounts \"cattle-impersonation-system/cattle-impersonation-u-qwerty123\" not found") has prevented the request from succeeding
```

Though the UI is successfully able to display the nodes:
<img width="1280" alt="Screenshot 2021-09-21 at 3 53 20 PM" src="https://user-images.githubusercontent.com/87482092/134154756-c1a4efac-d634-49b1-a257-a62f8ed7ed5c.png">

cmurphy:
[...]

PR #34861, opened 2021-09-22
Make impersonation service account more resilient

cmurphy:
Currently, the impersonation path checks for the existence of a matching
impersonation clusterrole on the downstream cluster and then assumes
there is an impersonation service account to go with it. If for some
reason there was an interruption in between creating the clusterrole and
creating the service account, the service account won't exist and
further requests will be perpetually stuck. This change ensures that
won't happen by checking for a Not Found error when retrieving the
service account, and continuing to create the needed resources in such
an event. This also fixes an unclear error message to clarify that the
missing resource is the service account, not the secret, and to
distinguish it from another similar error message.

https://github.com/rancher/rancher/issues/34824

### DIFF
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -55,7 +55,11 @@ func (i *Impersonator) SetUpImpersonation() (*corev1.ServiceAccount, error) {
 		return nil, err
 	}
 	if role != nil {
-		return i.getServiceAccount()
+		sa, err := i.getServiceAccount()
+		// in case the role exists but we were interrupted before creating the service account, proceed to create resources
+		if err == nil || !apierrors.IsNotFound(err) {
+			return sa, err
+		}
 	}
 	logrus.Tracef("impersonation: creating impersonation namespace")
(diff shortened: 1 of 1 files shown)


snasovich (commented):
The actual code changes LGTM. Initially, I was concerned that `apierrors.IsNotFound(err)` may not work properly due to returning error wrapped with `fmt.Errorf`, but tracing implementation it looks like it's handled by `apierrors.IsNotFound` properly.

Also, I'm wondering if it's possible we're interrupted right between SA and role binding created. I believe in this case `getServiceAccount` will return non-nil user, but role binding won't be there. Could it be a good idea to ensure RB exists even if we've got SA back?

cmurphy:
The function checks the role first, if the role does not exist then this whole section is skipped https://github.com/rancher/rancher/pull/34861/files#diff-fe913e39db39d26629ffb0d1e83c60678ce415c81af03283120afeec3233f011R57-R63

dramich (approved):

snasovich (approved):

Issue #34863, opened 2021-09-22
[2.6] Add support for cluster-level registry(rke) for shell and helm


StrongMonkey:
Forward-port of https://github.com/rancher/rancher/issues/30735

izaac:
Rancher version: `v2.6.1-rc3`

Validated working on the latest `v2.6.1` RC

![Screen Shot 2021-09-23 at 7 02 51 PM](https://user-images.githubusercontent.com/27659/134607255-7d3e5de9-bd43-4d9d-a763-bee6737b3320.png)



![Screen Shot 2021-09-23 at 7 00 02 PM](https://user-images.githubusercontent.com/27659/134607256-bc94c58b-7558-471f-8390-06fa3cf96c38.png)

PR #34864, opened 2021-09-22
[2.6]Add support for cluster-level registry(rke) for shell and helm

StrongMonkey:
https://github.com/rancher/rancher/issues/34863


kinarashah (approved):
All code files same as original PR. Only missing files are go.mod and go.sum, not required because steve and remotedialer are already updated for release/v2.6.

Issue #34877, opened 2021-09-23
[v2.5.10] Backport: '"unable to create impersonator account ... serviceaccounts not found" error after upgrade from 2.5.9 to 2.6.0'


cmurphy:
Backport issue for https://github.com/rancher/rancher/issues/34824

anupama2501:
this issue is not relevant anymore from Dan and Colleen's comment:
```
cattle-impersonation-system is removed and it now uses a header sanitizing middleware.
All of 2.5 will work this way, 2.6 has the namespace stuff.
All of those changes have been reverted. Those backport issues are not really relevant anymore, all issues related to impersonation in 2.5 won't exist because all changes were reverted.
```

PR #34878, opened 2021-09-23
[v2.5] Make impersonation service account more resilient

cmurphy:
Currently, the impersonation path checks for the existence of a matching
impersonation clusterrole on the downstream cluster and then assumes
there is an impersonation service account to go with it. If for some
reason there was an interruption in between creating the clusterrole and
creating the service account, the service account won't exist and
further requests will be perpetually stuck. This change ensures that
won't happen by checking for a Not Found error when retrieving the
service account, and continuing to create the needed resources in such
an event. This also fixes an unclear error message to clarify that the
missing resource is the service account, not the secret, and to
distinguish it from another similar error message.

Backport of https://github.com/rancher/rancher/pull/34861

https://github.com/rancher/rancher/issues/34877

### DIFF
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -55,7 +55,11 @@ func (i *Impersonator) SetUpImpersonation() (*corev1.ServiceAccount, error) {
 		return nil, err
 	}
 	if role != nil {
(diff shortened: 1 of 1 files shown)


snasovich (approved):
LGTM, same as the source changes

PR #34887, opened 2021-09-23
[Backport 2.5] Avoid a panic when a user adds a chart catalog containing a blank Helm chart

genexpr:
Backport from PR [#34787 ](https://github.com/rancher/rancher/pull/34787)
Original Issue [#28776](https://github.com/rancher/rancher/issues/28776)
Linked backport issue [#34912 ](https://github.com/rancher/rancher/issues/34912)

### DIFF
--- a/pkg/api/norman/customization/catalog/store.go
+++ b/pkg/api/norman/customization/catalog/store.go
@@ -16,6 +16,7 @@ import (
 	managementschema "github.com/rancher/rancher/pkg/schemas/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/types/config"
+	"github.com/sirupsen/logrus"
 )
 
 type templateStore struct {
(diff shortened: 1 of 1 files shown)


superseb (approved):

genexpr:
### QA testing

Use Cluster Manager.
In the context of any cluster, add a new catalog. Give it a name and choose a repository that contains a bad chart definition. A bad chart definition in this case is one whose `Chart.yaml` is missing the `version` field. There is one that is good for testing: https://github.com/axeal/rancher-catalog

Use the above Catalog URL and select a branch that has an invalid `Chart.yaml`. In this case, the branch `noversion`. Then click Create and observe that there are no errors in the cluster. The catalog is added. It only contains one chart, and this chart is invalid. This means that when you try to add an App to a Project, this repository (catalog) you had added will not appear in the list of available catalogs, since it does not have any valid charts.

Add another catalog. Use the same Catalog URL but choose a branch that contains that same chart definition but in a valid form - the `master` branch. When you Create the new catalog, you will also not see any errors. But when you try to add an App, you will see your catalog with its charts (one, in this case) listed in the list of available catalogs.

Issue #34744, opened 2021-09-15
Downstream clusters are deleted upon upgrade that are managed in API-created Fleet workspaces


nickgerace:
This occurs on upgrade from 2.5.9 to 2.6.0. Clusters that are in Fleet workspaces created in GUI survive upgrade. Clusters are in Fleet workspaces create via the API do not survive upgrade (v3 Cluster is deleted and Rancher will deprovision nodes, if applicable).

SURE-3379
SURE-3329

nickgerace:
## QA Validation

This comment contains the reproduction and validation scenarios for this issue.

### Setup

1. Start with Rancher `v2.5.9` (or `v2.5.8`)
1. Create one Fleet workspace through kubectl and call it `fleet-kubectl`
1. Create another Fleet workspace through the GUI and call it `fleet-dashboard`
1. Provision at least three downstream clusters (provider does not matter, but you can vary it to be sure)
1. Move one cluster to one of the new workspaces, and move another cluster to the other
1. Leave the remaining cluster in `fleet-default`
1. Wait for clusters to reach "Active" state

### Reproduction

1. Upgrade to `v2.6.1-rc3` or `v2.6-head`
1. Observe all v3 management clusters that have corresponding v1 provisioning clusters not in `fleet-default` or `fleet-local` being deleted
1. Move cluster from `fleet-default` to either of the new workspaces and observe the v3 management cluster being deleted

### Validation

1. Observe all clusters eventually reaching an "Active" state (may take awhile due to provisioning v2's initial setup which finishing setting up Fleet resources towards the end of the entire upgrade process)
1. Move all three clusters around between the three workspaces (including `fleet-default`) and wait for them to reach an "Active" state (you can do this multiple times)
1. Observe all clusters eventually reaching an "Active" state (you might see the following issue, but it is transient: https://github.com/rancher/rancher/issues/34937)

### Optional Validation

You can create `GitRepo` CR(s) before and after upgrade, but you may have to edit them based on the workspace. It may be best to create `GitRepo` CR(s) _after_ validation as a result.

izaac:
I was not able to reproduce this yet these are my details after following the above documented steps:

- 3 clusters single node all roles RKE1 EC2
- All clusters are assigned to the `fleet-default` fleet workspace by default.
- I created a fleet workspace using kubectl `fleet-kubectl` (see the yaml below)
- I created a fleet workspace using the Dashboard UI `fleet-dashboard`
[...]

PR #34888, opened 2021-09-23
Ensure provisioning cluster migration does not delete management cluster

nickgerace:
Ensure provisioning cluster migration does not delete management cluster
by checking if the FleetWorkspace has changed. This is trigged in the UI
by using "assign to" in Continuous Delivery, which deletes and
re-creates the provisioning cluster as well as its corresponding Fleet
cluster. Additional changes include:

- deleting the provisioning-cluster-specific v3 User due
  to potential v3 User "memory leak" upon change the Fleet cluster's
  FleetWorkspace (v3 Users would be created for each migration, but never
  deleted)
- changing the CRTB name to include the cluster's namespace due to a
  race condition where the old provisioning cluster's deletion could
  cause the CRTB to be deleted due to resource "apply"

Relevant issue: https://github.com/rancher/rancher/issues/34744

Design considerations:
- upgrades from v2.6.0 will still benefit from the CRTB change because they will retain their names until the provisioning cluster is deleted (Wrangler "apply" will clean it up)

### DIFF
--- a/pkg/controllers/provisioningv2/cluster/remove.go
+++ b/pkg/controllers/provisioningv2/cluster/remove.go
@@ -95,16 +95,28 @@ func (h *handler) OnClusterRemove(_ string, cluster *v1.Cluster) (*v1.Cluster, e
 
 func (h *handler) doClusterRemove(cluster *v1.Cluster) (string, error) {
 	if cluster.Status.ClusterName != "" {
-		err := h.mgmtClusters.Delete(cluster.Status.ClusterName, nil)
-		if err != nil && !apierrors.IsNotFound(err) {
-			return "", err
-		}
-
-		if h.isLegacyCluster(cluster) {
-			// If this is a legacy cluster (i.e. RKE1 cluster) then we should wait to remove the provisioning cluster until the v3.Cluster is gone.
-			_, err = h.mgmtClusterCache.Get(cluster.Status.ClusterName)
+		mgmtCluster, err := h.mgmtClusters.Get(cluster.Status.ClusterName, metav1.GetOptions{})
+		if err != nil {
+			// We do nothing if the management cluster does not exist (IsNotFound) because it's been deleted.
 			if !apierrors.IsNotFound(err) {
-				return fmt.Sprintf("waiting for cluster [%s] to delete", cluster.Status.ClusterName), nil
+				return "", err
+			}
+		} else if cluster.Namespace == mgmtCluster.Spec.FleetWorkspaceName {
+			// We only delete the management cluster if its FleetWorkspaceName matches the provisioning cluster's
+			// namespace. The reason: if there's a mismatch, we know that the provisioning cluster needs to be migrated
+			// because the user moved the Fleet cluster (and provisioning cluster, by extension) to another
+			// FleetWorkspace. Ultimately, the aforementioned cluster objects are re-created in another namespace.
+			err := h.mgmtClusters.Delete(cluster.Status.ClusterName, nil)
+			if err != nil && !apierrors.IsNotFound(err) {
+				return "", err
+			}
+
+			if h.isLegacyCluster(cluster) {
+				// If this is a legacy cluster (i.e. RKE1 cluster) then we should wait to remove the provisioning cluster until the v3.Cluster is gone.
+				_, err = h.mgmtClusterCache.Get(cluster.Status.ClusterName)
+				if !apierrors.IsNotFound(err) {
+					return fmt.Sprintf("waiting for cluster [%s] to delete", cluster.Status.ClusterName), nil
+				}
 			}
 		}
 	}
@@ -142,5 +154,5 @@ func (h *handler) doClusterRemove(cluster *v1.Cluster) (string, error) {
 		return fmt.Sprintf("waiting for cluster-api cluster [%s] to delete", cluster.Name), nil
 	}
 
-	return "", nil
+	return "", h.kubeconfigManager.DeleteUser(cluster.Namespace, cluster.Name)
 }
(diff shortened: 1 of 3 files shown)


pkg/provisioningv2/kubeconfig/manager.go
```diff
@@ -131,13 +132,19 @@ func getPrincipalID(clusterNamespace, clusterName string) string {
...
-func (m *Manager) createUser(principalID, userName string) error {
+func (m *Manager) createUser(principalID, userName, clusterName string, clusterUID types.UID) error {
 	_, err := m.userCache.Get(userName)
 	if apierror.IsNotFound(err) {
 		_, err = m.users.Create(&v3.User{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:   userName,
 				Labels: labelsForUser(principalID),
+				OwnerReferences: []metav1.OwnerReference{{
+					APIVersion: "provisioning.cattle.io/v1",
```
StrongMonkey, line 143: Nit: you can use https://github.com/rancher/rancher/blob/28345b9c6463d33b653adc78b16a3922db1e320e/pkg/apis/provisioning.cattle.io/v1/zz_generated_register.go#L35 instead of hardcode group.

pkg/provisioningv2/kubeconfig/manager.go
```diff
@@ -131,13 +132,19 @@ func getPrincipalID(clusterNamespace, clusterName string) string {
...
+					Kind:       "Cluster",
+					Name:       clusterName,
```
StrongMonkey, line 145: I don't think this will work. User is cluster-scoped resource and cluster.provisioning.cattle.io is namespaced scoped. https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/ Cluster-scoped resource can only have cluster-scoped resource as owner.

nickgerace: Perhaps, I should just delete it then in the "mutually assured destruction" cluster removal code then?

pkg/provisioningv2/kubeconfig/manager.go
nickgerace: I do not believe this will work because `APIVersion` requires a string here.

StrongMonkey: hmm, use schemeGroupVersion.String()....

nickgerace: Ah there it is. Changing...

pkg/controllers/provisioningv2/cluster/remove.go
```diff
@@ -142,5 +154,5 @@ func (h *handler) doClusterRemove(cluster *v1.Cluster) (string, error) {
 		return fmt.Sprintf("waiting for cluster-api cluster [%s] to delete", cluster.Name), nil
 	}
 
-	return "", nil
+	return "", h.kubeconfigManager.DeleteUser(cluster.Namespace, cluster.Name)
```
StrongMonkey, line 157: wait... the user name is not the same as cluster.Name. You have to calculate through hash function

StrongMonkey: oh nvm, I see it is a wrapper function

StrongMonkey (approved):

pkg/controllers/provisioningv2/cluster/remove.go
nickgerace: It uses the same pattern as "CreateUser"

pkg/provisioningv2/kubeconfig/manager.go
```diff
@@ -201,15 +212,15 @@ func createSHA256Hash(secretKey string) (string, error) {
...
+func (m *Manager) GetCRTBForClusterOwner(cluster *v1.Cluster, status v1.ClusterStatus) (*v3.ClusterRoleTemplateBinding, error) {
 	if status.ClusterName == "" {
 		return nil, fmt.Errorf("management cluster is not assigned to v1.Cluster")
 	}
 	principalID := getPrincipalID(cluster.Namespace, cluster.Name)
 	userName := getUserNameForPrincipal(principalID)
 	return &v3.ClusterRoleTemplateBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      name.SafeConcatName(status.ClusterName, "owner"),
+			Name:      name.SafeConcatName(status.ClusterName, cluster.Namespace, "owner"),
```
nickgerace, line 223: @dramich: this solved our race condition issue, as a side note

pkg/controllers/provisioningv2/cluster/remove.go
```diff
@@ -95,16 +95,28 @@ func (h *handler) OnClusterRemove(_ string, cluster *v1.Cluster) (*v1.Cluster, e
...
-		err := h.mgmtClusters.Delete(cluster.Status.ClusterName, nil)
-		if err != nil && !apierrors.IsNotFound(err) {
-			return "", err
-		}
-
-		if h.isLegacyCluster(cluster) {
-			// If this is a legacy cluster (i.e. RKE1 cluster) then we should wait to remove the provisioning cluster until the v3.Cluster is gone.
-			_, err = h.mgmtClusterCache.Get(cluster.Status.ClusterName)
+		mgmtCluster, err := h.mgmtClusters.Get(cluster.Status.ClusterName, metav1.GetOptions{})
+		if err != nil {
```
dramich, line 99: Why isn't the error returned?

pkg/controllers/provisioningv2/cluster/remove.go
```diff
@@ -95,16 +95,28 @@ func (h *handler) OnClusterRemove(_ string, cluster *v1.Cluster) (*v1.Cluster, e
...
+			// We do nothing if the management cluster does not exist (IsNotFound) because it's been deleted.
 			if !apierrors.IsNotFound(err) {
-				return fmt.Sprintf("waiting for cluster [%s] to delete", cluster.Status.ClusterName), nil
+				return "", err
+			}
```
nickgerace, line 103: To reviewers: If it's a `NotFound` API error, then this is a NOP

dramich (approved):

Issue #34919, opened 2021-09-24
Can not edit/upgrade the app if the original chart is hidden by the rancher-version annotation 


cbron:
Backend issue for: https://github.com/rancher/dashboard/issues/4276

Look into filtering the chart _only_ from the index.

cbron:
Once the UI changes merge this will be ready to test

cbron:
This can be tested along with https://github.com/rancher/rancher/issues/34875

brendarearden:
### Root cause 
All api calls to the index, info, chart and icons endpoints start with getting the index file. This results in all of those calls happening against a filtered index file. 

### What was fixed, or what changes have occurred 
A `skipFiltering` flag was added so that the filtering of the chart index could be more finely controlled. 

The info endpoint now has `skipFiltering` set to false so that anytime we want to know all of the currently supported charts, charts will get filtered out properly. 

The info, chart and icon endpoints have `skipFiltering` true, because if the chart exist in the repository, we should be able to get the chart info, the chart files and the chart's icon

### Areas or cases that should be tested 
Chart install, edit without upgrading, upgrade, uninstall. This should be done on a chart with and without crds
In rancher v2.6.0 install a chart that is not supported in v2.6.1 (i.e a chart with catalog.cattle.io/rancher-version: "<= 2.6.0-0" in the chart.yaml). After installing, upgrade rancher to v2.6-head, edit your chart installation without upgrading, upgrade and uninstall. 

### What areas could experience regressions ?
All catalog v2 api endpoints had the flag added, so all of the catalog section could. Since we are skipping filtering, it could regress where charts should've been filtered based on rancher-version or kubernetes version but no longer are. 

### Are the repro steps accurate/minimal ?

Yes
https://github.com/rancher/dashboard/issues/4276#issue-1005961299

cbron:
Assigning @jiaqiluo for dev-validate here.

jiaqiluo:
The following validations are done on both the `local` and `downstream` clusters:
- run rancher 2.6.0 and add a downstream cluster 
- add the following GH repo as a repository, it is forked from the release-v2.6 branch  
```
URL:    https://github.com/jakefhyde/charts	
branch: add-min-rancher-version
```
- go to `Charts` -> select only the newly-added repo 
- **Confirmation 1**: for all available charts, all versions show up in the `Chart Versions` list, there is no difference with the released branch
- install the app Istio `1.9.600` into the cluster  
[...]

PR #34921, opened 2021-09-24
Add a skipFilter bool so that not all calls occur on a filtered index

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/34919

**Problem**
rancher-version filtering is causing errors when upgrading to newer versions of rancher where the currently installed chart is not longer supported in the new rancher version. When the user goes to edit/upgrade their install, an error is showing because the chart info is filtered out of the api response. 

**Solution**
Add a flag so that only the index filter charts based on chart annotations. This will allow specific chart get calls to return the chart information if it still exists in the chart repository.

### DIFF
--- a/pkg/catalogv2/content/content_test.go
+++ b/pkg/catalogv2/content/content_test.go
@@ -5,13 +5,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/stretchr/testify/assert"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/repo"
 )
 
-func TestFilterReleases(t *testing.T) {
+func TestFilterReleasesSemver(t *testing.T) {
 	tests := []struct {
 		testName               string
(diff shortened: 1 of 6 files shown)


jakefhyde (approved):
LGTM

jiaqiluo (approved):

pkg/catalogv2/content/content_test.go
```diff
@@ -193,7 +193,199 @@ func TestFilterReleases(t *testing.T) {
...
+		chartVersionAnnotation string
+		rancherVersion         string
+		expectedPass           bool
+	}{
+		{
+			"rancher version in range comparison with `>= <`style comparison",
+			">= 2.5.0-alpha3 <2.6.0",
+			"v2.5.0+123", //SemVer comparisons using constraints without a prerelease comparator will skip prerelease versions
+			true,
+		},
```
cbron, line 218: nit: this is more of a semver test than a functionality test. Since we have all those other semver tests I think you could just have a true/false test here and be good. But otherwise good job adding tests.

cbron (approved):
Not entirely familiar with all cases here but if its been tested via UI then lgtm.

cbron (approved):

Issue #33689, opened 2021-07-20
Secret available for all namespace not respect the labels provided


disaster37:
**Rancher Server Setup**
- Rancher version: 2.5.7
- Installation option (Docker install/Helm Chart): Docker as standalone

**Information about the Cluster**
- Kubernetes version: v1.20.5
- Cluster Type (Local/Downstream): Custom

**Describe the bug**

When create new secret from UI, we can set to use it as `Available to all namespaces in this project
`. And we can set some labels and annotations.

But when get secret on each namespace, we can look that the labels is empty (not duplicated). For annotation it's OK.

**To Reproduce**
Create new secret on default project that contain default namespace. Set option ``Available to all namespaces in this project
` and add some labels and annotations.

Now get secret form kubectl `kubectl get secret/test -n default -o yaml`.
Like you can see, you haven't labels, but only annotations.

**Result**

The labels are lost.

**Expected Result**

Keep labels and annotations provied when cretae secrets from UI.

**Screenshots**

**Additional context**
<!--Add any other context about the problem here.—>

genexpr:
This is relevant for only the Cluster Manager, so just `v2.5`. However, we should add a fix for `v2.6` and then backport to `v2.5`.

genexpr:
## Testing
### Root cause
The labels of a secret aren't created, if a secret is created for all namespaces from the Cluster Manager UI.
We are not copying the labels from the request to the new secret value to be persisted.
### What was fixed, or what changes have occurred
Along with other secret data and metadata, we now properly copy labels from the secret creation request and store a new secret.
### Areas or cases that should be tested
Create a secret for all namespaces from the Cluster Manager UI. Also worth trying in Cluster Dashboard UI, but the issue had not been encountered there.
### What areas could experience regressions?
 
### Steps
Create a new secret from the Cluster Manager UI, set to use it as `Available to all namespaces in this project`. Then set some labels and annotations.
Now read the secret form the integrated shell: `kubectl get secret/test -n default -o yaml`. Or view it in the UI.
Observe that the annotations AND labels are present.

PR #34923, opened 2021-09-24
Include labels on secret create in cluster manager

genexpr:
Issue [#33689 ](https://github.com/rancher/rancher/issues/33689)
This is relevant only for the Cluster Manager UI. Creating a Secret that is `Available to all namespaces in this project` does not copy the labels to a secret available in the `default` namespace.


pkg/controllers/managementuser/secret/secret.go
```diff
@@ -250,12 +250,14 @@ func getNamespacedSecret(obj *corev1.Secret, namespace string) *corev1.Secret {
 	namespacedSecret.Namespace = namespace
 	namespacedSecret.Type = obj.Type
 	namespacedSecret.Annotations = make(map[string]string)
-	copyMap(obj.Annotations, namespacedSecret.Annotations)
+	namespacedSecret.Labels = make(map[string]string)
```
aiyengar2, line 253: I'm not aware enough about the reasons why we might have intentionally omitted copying the labels in the first place, but from what I can tell this LGTM.

@kinarashah any thoughts here? Since it seems like you made the last change in https://github.com/rancher/rancher/commit/413c663813f6b6bf543891c3fd7cce9700a2e213.

aiyengar2 (changes requested):

pkg/controllers/managementuser/secret/secret.go
```diff
@@ -250,12 +250,14 @@ func getNamespacedSecret(obj *corev1.Secret, namespace string) *corev1.Secret {
...
+	copyMap(namespacedSecret.Annotations, obj.Annotations)
+	copyMap(namespacedSecret.Labels, obj.Labels)
 	namespacedSecret.Annotations[userSecretAnnotation] = "true"
 	return namespacedSecret
 }
 
-func copyMap(src map[string]string, dst map[string]string) {
+func copyMap(dst map[string]string, src map[string]string) {
```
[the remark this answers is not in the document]
genexpr: Yes, it's not strictly necessary. Flipping it brings it in line with the loose Go convention of specifying the destination as the first parameter of such functions. Especially since it works with values that contain pointers internally. Like the builtin `copy`, for example.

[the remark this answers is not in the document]
aiyengar2: Hmm okay, that change sounds good to me then

aiyengar2 (approved):
LGTM

pkg/controllers/managementuser/secret/secret.go
kinarashah: @genexpr @aiyengar2 I don't remember why we skipped labels, PR is very old. But we should make sure that labels don't include any keys that controllers check for on sync and run some logic, this could be specific to project level scope and not run for each namespace.

aiyengar2: Hmm @kinarashah I'm not aware of anything that controllers rely on to sync or run some logic, but I'll leave it to other reviewers to perhaps comment. 

I also think that controllers using labels on secrets to execute some action seems like something we should fix rather than something we should prevent in the code here, since they could create that secret via kubectl with the same permissions anyways.

@genexpr if you can identify any such keys in the code (just search for constants or vars that we define that have to do with secrets), I think it might make sense to start a `denylist` here that deletes those labels before persisting it. But unless we can identify such a case, I think the current PR looks good to me still.

kinarashah: @aiyengar2 It's not that the controller would execute some action using labels, it'd be more like what you see in the controller, it relies the projectID from the annotation and if it exists, runs some logic. So it's unrelated to user doing anything. They're rancher keys and not user keys. The idea tho is the same as you mentioned, if in case you find a key is being used for a purpose that isn't required to run for namespacedSecret, we shouldn't save it on this secret. For this, we're on the same page based on your last paragraph and the PR is good otherwise.

kinarashah: I looked for it, and we're good.

aiyengar2: ack, thanks for the explanation! https://github.com/rancher/rancher/pull/34923#discussion_r721765205

Issue #34889, opened 2021-09-23
[Hosted Rancher] resourceQuota controllers error logs on missing projects


aiyengar2:
Same setup as https://github.com/rancher/rancher/issues/34833. Seen in Rancher 2.5.9 and Rancher 2.6.0 for either the super rancher or downstream rancher setups.

**Describe the bug**

Observing the following repeated logs in a Hosted Rancher downstream Rancher logs: 
```
Thu, Sep 23 2021 2:09:30 pm | 2021/09/23 21:09:30 [ERROR] error syncing 'cattle-monitoring-system': handler resourceQuotaSyncController: projects.management.cattle.io "c-vpmfh/p-2f9qt" not found, handler resourceQuotaUsedLimitController: projects.management.cattle.io "c-vpmfh/p-2f9qt" not found, requeuing
Thu, Sep 23 2021 2:09:30 pm | 2021/09/23 21:09:30 [ERROR] error syncing 'fleet-system': handler resourceQuotaSyncController: projects.management.cattle.io "c-vpmfh/p-2f9qt" not found, handler resourceQuotaUsedLimitController: projects.management.cattle.io "c-vpmfh/p-2f9qt" not found, requeuing
Thu, Sep 23 2021 2:09:30 pm | 2021/09/23 21:09:30 [ERROR] error syncing 'cattle-prometheus': handler resourceQuotaSyncController: projects.management.cattle.io "c-vpmfh/p-2f9qt" not found, handler resourceQuotaUsedLimitController: projects.management.cattle.io "c-vpmfh/p-2f9qt" not found, requeuing
```

**To Reproduce**
 

See original ticket.

**Result**

Logs with errors are produced.

**Expected Result**

No errors should be logged.

**Screenshots**

**Additional context**
<!--Add any other context about the problem here.—>

This is likely due to the fact that our resource quota controllers do not have any logic for dealing with projects that are defined within the cluster but not owned by Rancher.

aiyengar2:
Could be part of https://github.com/rancher/rancher/issues/34746.

dnoland1:
Tested with upstream Rancher v2.6.1-rc5 and downstream Rancher v2.5.9 and v2.6.1-rc5. No longer seeing "error syncing..." logs.

PR #34938, opened 2021-09-27
Ignore resourceQuotas and apps in undefined projects

aiyengar2:
This PR modifies the resourceQuota and appState controllers to ignore apps & projects that are not found on trying to update. This is specifically meant to mitigate CPU spikes we see in a Hosted Rancher setup.

Specifically, it encodes the following logic:
- [AppState] If a workload with an annotation that appears to point to a v3 app that is **not** tracked by our management cluster is found, we should ignore updating the app state since it's probably managed by another Rancher.
- [ResourceQuota] If a namespace with an annotation that appears to point to a v3 project that is **not** tracked by our management cluster is found, we should ignore updating the resource quota since it's probably managed by another Rancher.

---

I also made a minor change to use `ref.Parse` instead of an unnecessary helper function.

---

Depends on the fix from https://github.com/rancher/rancher/pull/34860 to resolve the first issue.

Related Issue: https://github.com/rancher/rancher/issues/34833, https://github.com/rancher/rancher/issues/34889, https://github.com/rancher/rancher/issues/34746

### DIFF
--- a/pkg/controllers/managementuser/resourcequota/resource_quota_calculate_used.go
+++ b/pkg/controllers/managementuser/resourcequota/resource_quota_calculate_used.go
@@ -6,8 +6,10 @@ import (
 
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	namespaceutil "github.com/rancher/rancher/pkg/namespace"
+	"github.com/rancher/rancher/pkg/ref"
 	validate "github.com/rancher/rancher/pkg/resourcequota"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	quota "k8s.io/apiserver/pkg/quota/v1"
 	clientcache "k8s.io/client-go/tools/cache"
@@ -44,9 +46,14 @@ func (c *calculateLimitController) calculateResourceQuotaUsedProject(key string,
 }
 
 func (c *calculateLimitController) calculateProjectResourceQuota(projectID string) error {
-	projectNamespace, projectName := getProjectNamespaceName(projectID)
+	projectNamespace, projectName := ref.Parse(projectID)
 	project, err := c.projectLister.Get(projectNamespace, projectName)
 	if err != nil || project.Spec.ResourceQuota == nil {
+		if errors.IsNotFound(err) {
+			// If Rancher is unaware of a project, we should ignore trying to calculate the project resource quota
+			// A non-existent project is likely managed by another Rancher (e.g. Hosted Rancher)
+			return nil
+		}
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


nickgerace (dismissed):

pkg/controllers/managementuser/resourcequota/resource_quota_common.go
```diff
@@ -236,17 +250,6 @@ func getProjectID(ns *corev1.Namespace) string {
...
-func getProjectNamespaceName(projectID string) (string, string) {
-	if projectID == "" {
-		return "", ""
-	}
-	parts := strings.Split(projectID, ":")
-	if len(parts) == 2 {
-		return parts[0], parts[1]
-	}
-	return "", ""
-}
```
aiyengar2, line 248: `ref.Parse` is slightly different than this: https://github.com/rancher/rancher/blob/c068d30eaffa24337b97e16e3a12d22219d0ac6d/pkg/ref/parse.go#L33-L37

However, I don't think the specific edge case that the helper function is catching merits another function that needs to be defined.

On another note, we probably need to clean up the code at some point to consolidate all these one-off helper functions that are doing redundant stuff (e.g. we put labels and annotation stuff for projects all across several controllers rather than referencing a single package that contains the expected label + annotation).

nickgerace: I'd file an issue for that.

aiyengar2: https://github.com/rancherlabs/engineering-improvements/issues/30

nickgerace (dismissed):

pkg/controllers/managementuser/resourcequota/resource_quota_calculate_used.go
```diff
@@ -44,9 +46,14 @@ func (c *calculateLimitController) calculateResourceQuotaUsedProject(key string,
 }
 
 func (c *calculateLimitController) calculateProjectResourceQuota(projectID string) error {
-	projectNamespace, projectName := getProjectNamespaceName(projectID)
+	projectNamespace, projectName := ref.Parse(projectID)
 	project, err := c.projectLister.Get(projectNamespace, projectName)
 	if err != nil || project.Spec.ResourceQuota == nil {
+		if errors.IsNotFound(err) {
+			// If Rancher is unaware of a project, we should ignore trying to calcalate the project resource quota
```
superseb, line 53: there is a typo here but not sure why CI is passing on this

superseb (approved):
just one typo

pkg/controllers/managementuser/resourcequota/resource_quota_calculate_used.go
aiyengar2: updated

nickgerace (approved):

superseb (approved):

PR #34939, opened 2021-09-27
Add cluster ID to error

rmweir:
Make error more useful by including cluster ID.

### DIFF
--- a/pkg/controllers/managementuser/settings/controller.go
+++ b/pkg/controllers/managementuser/settings/controller.go
@@ -2,6 +2,7 @@ package settings
 
 import (
 	"context"
+	"fmt"
 
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/settings"
@@ -20,7 +21,8 @@ var (
 )
 
 type handler struct {
-	apply apply.Apply
+	clusterID string
+	apply     apply.Apply
 }
 
 func Register(ctx context.Context, userContext *config.UserContext) error {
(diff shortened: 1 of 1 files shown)


jiaqiluo (dismissed):
LGTM. We should backport this PR to 2.5 if it is applicable.

pkg/controllers/managementuser/settings/controller.go
```diff
@@ -42,7 +45,7 @@ func (h *handler) onChange(key string, obj *v3.Setting) (runtime.Object, error)
 		return nil, nil
 	}
 
-	return obj, h.apply.
+	err := h.apply.
```
dramich, line 48: This apply should use a cache so we don't have to hit the downstream API i.e. `WithCacheTypes`

snasovich (approved):

pkg/controllers/managementuser/settings/controller.go
```diff
@@ -42,13 +46,18 @@ func (h *handler) onChange(key string, obj *v3.Setting) (runtime.Object, error)
...
 		WithOwner(obj).
+		WithCacheTypes().
```
dramich, line 51: It requires a cache to be passed in

dramich: https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/authprovisioningv2/authprovisioningv2.go#L68

rmweir: Forgot to remove this. I did this above in Register() where I have access to the cache.

snasovich (approved):

dramich (approved):

jiaqiluo (approved):

nickgerace:
https://github.com/rancher/rancher/issues/35045

Issue #34835, opened 2021-09-20
Upgrading rancher server single node install by setting mcm to false `Failed to find system chart fleet` errors are seen in the rancher logs continuously


anupama2501:
**Rancher Server Setup**
- Rancher version: `v2.6.0`
- Installation option (Docker install/Helm Chart): Docker install
   - If Helm Chart, Kubernetes Info: k3s
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):
         Node Setup: 1 node 
         Version: `v1.21.3+k3s1`
- Proxy/Cert Details: `self-signed`

After upgrade, rancher server: `v2.6-head ba43472`

**Describe the bug**
After upgraded rancher server by setting mcm flag to false, errors are seen running continuously in the rancher logs.

**To Reproduce**
1. Installed rancher server via docker `docker run -d --restart=unless-stopped  -p 80:80 -p 443:443 --privileged rancher/rancher:v2.6.0 --features=multi-cluster-management=false`
2. Verified only local cluster is available and no cluster management tab to create clusters.
3. Create a workload in the local cluster and a daemon set
4. Upgrade the rancher server to v2.6-head with the commands
```
docker stop <container_id>
docker create --volumes-from <container_id> --name rancher-data rancher/rancher:v2.6.0
docker run -d --privileged --volumes-from rancher-data --restart=unless-stopped -p 80:80 -p 443:443 rancher/rancher:v2.6-head --features=multi-cluster-management=false
```

**Result**
After a few minutes rancher comes up active. the following logs are seen continuously in the rancher server logs:
```
2021/09/20 21:47:03 [ERROR] Failed to find system chart fleet will try again in 5 seconds: no chart name found
2021/09/20 21:47:08 [ERROR] available chart version (100.0.0+up0.3.6) for fleet is less than the min version (100.0.1+up0.3.7)
```

**Expected Result**
The above error logs shouldn't be seen in the rancher server as the fleet has deployed without any error.

**Screenshots**
<img width="1520" alt="Screen Shot 2021-09-20 at 2 54 55 PM" src="https://user-images.githubusercontent.com/38144301/134081111-52276920-2c57-4cd8-b448-605206007aa6.png">

**Additional context**
Ref: https://github.com/rancher/rancher/issues/34691

**Note**: 
Issue is not seen on a fresh install of v2.6-head when the flag `--features=multi-cluster-management=false`
[...]

PR #34956, opened 2021-09-28
 Fix deadlock when installCharts is stuck in a loop 

ibuildthecloud:


kinarashah:
https://github.com/rancher/rancher/issues/34835

kinarashah (approved):
Validated using `kinarashah/rancher:v2.6.1-rc3` with the fix.

nickgerace (approved):

pkg/catalogv2/system/system.go
```diff
@@ -120,7 +120,10 @@ func (m *Manager) onTrigger(_ string, obj *catalog.ClusterRepo) (*catalog.Cluste
 		return obj, nil
 	}
 
-	m.trigger <- struct{}{}
+	select {
+	case m.trigger <- struct{}{}:
+	default:
+	}
```
thedadams, line 126: Do we want to error or re-enqueue in the `default` case to ensure this eventually gets reconciled?

ibuildthecloud: I don't think so because if this happens the queue already has one item in it.  That chan is of length 1.

cbron: Hmm, thats an interesting point. If the catalog is rancher-charts and its blocked, then default executes and nothing happens until it specifically gets changed/refreshed again.

PR #34976, opened 2021-09-29
Update UI tags and CLI tag for v2.6.1

nickgerace:


kinarashah (approved):
- Confirmed curl for both ui tags work 
- Confirmed there's no other new cli tags

Issue #34876, opened 2021-09-23
Convert sync handler to env variables


cbron:
Continuation of this issue: https://github.com/rancher/rancher/issues/34875

1. Add all settings from sync handler (currently 4) as env vars. 
2. Remove the settings sync handler: https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementuser/settings/controller.go
3. Audit which settings should trigger agent redeployment

jakefhyde:
Are there any concerns regarding updating of a downstream agent when one of these environment variables is changed? For example if I spin up a local and a downstream cluster, and `helm upgrade ... --set env.CATTLE_SERVER_VERSION=v2.5.9`, will the cluster agent be redeployed because it's deployment depends on the variable?

StrongMonkey:
You can check this method. https://github.com/rancher/rancher/blob/5c2933173bc07a1cd260d6e07fd9459243afc45d/pkg/controllers/management/clusterdeploy/clusterdeploy.go#L184

I don't think it will reploy agent if you just change the env variable. We only redeploy acccording certain conditions above. ^^ If you upgrade rancher it will redeploy agent since agent image changes and that will trigger cluster-agent upgrade.

cbron:
Setting up something to discuss.

jakefhyde:
### Root cause

### What was fixed, or what changes have occurred

Settings are no longer synced downstream, and instead passed from the local cluster to the cluster-agent deployment template as environment variables. 
The following settings are used to trigger a redeployment:
- server-version
- install-uuid
- ingress-ip-domain

### Areas or cases that should be tested

Changing server-version, install-uuid, or ingress-ip-domain in the local cluster triggers a redeployment of the downstream agent. Likewise, changing any other setting should **not** trigger a redeployment of the downstream agent.

### What areas could experience regressions?

N/A

### Are the repro steps accurate/minimal?

Deploy a downstream cluster, and then edit one of the settings, either through API or externally (i.e. kubectl, lens, etc.). The cluster agent should be redeployed.

slickwarren:
on 2.6-head (78b25c3) I've tested the following:
`example of how to change one of these vars: go to <your_server_ip>/v3/settings/<your_variable> -> edit -> send request`

* update install-uuid (random name) -> cattle-cluster-agent pod is redeployed -- pass
* update ingres-ip-domain (to nip.io) -> cattle-cluster-agent pod is redeployed -- pass
[...]

PR #34982, opened 2021-09-29
Convert sync handler to env vars

jakefhyde:
Moved the `install-uuid` and `ingress-ip-domain` settings from the settings sync-handler to cluster-agent deployment template. Removed settings sync handler.

Related Issue: https://github.com/rancher/rancher/issues/34876


jakefhyde:
@aiyengar2 @rmweir Do either of you have any input (or know someone who might) in regards to which settings should trigger an agent redployment?

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -239,7 +239,7 @@ func redeployAgent(cluster *v3.Cluster, desiredAgent, desiredAuth string, desire
 	}
 
 	if !reflect.DeepEqual(cluster.Spec.AgentEnvVars, cluster.Status.AppliedAgentEnvVars) {
-		logrus.Infof("clusterDeploy: redeployAgent: redeploy Rancher agents due to agent env vars mismatched for [%s], was [%v] and will be [%v]", cluster.Name, cluster.Spec.AgentEnvVars, cluster.Status.AppliedAgentEnvVars)
+		logrus.Infof("clusterDeploy: redeployAgent: redeploy Rancher agents due to agent env vars mismatched for [%s], was [%v] and will be [%v]", cluster.Name, cluster.Status.AppliedAgentEnvVars, cluster.Spec.AgentEnvVars)
```
snasovich, line 242: Nice catch swapping these 2.
On a side note, I'm wondering if there is a chance for any of these env variables to contain any sensitive information such as passwords? We only allow sensitive data to be present at trace level (FYI - https://rancher.com/docs/rancher/v2.6/en/troubleshooting/logging/)

pkg/settings/setting.go
```diff
@@ -256,3 +257,28 @@ func GetSettingByID(id string) string {
 	}
 	return provider.Get(id)
 }
+
+func MergeSettingsWithAgentEnvVars(envVars []v1.EnvVar) {
```
snasovich, line 261: This won't update the slice passed. Proof: https://play.golang.org/p/4OD4P4u4Sve

It may be a moot point though. Do we actually want `cluster.Spec.AgentEnvVars` updated to now include these select Rancher settings or do we just need a combined list to use further in the call sequence? From what I see it's the latter, so it makes more sense for the function to return a slice.

Alternatively, you can have func like `GetSettingsAsAgentEnvVars` taking no arguments and do appending on the caller's side. It may actually be cleaner.

pkg/settings/setting.go
```diff
@@ -256,3 +257,28 @@ func GetSettingByID(id string) string {
...
+	toCopy := []Setting{
+		ServerVersion,
+		InstallUUID,
+		IngressIPDomain,
+	}
+
+	envVarMap := map[string]v1.EnvVar{}
```
snasovich, line 268: Looks like we need this map to act as a set (we don't need values, just keys). In this case, I would recommend using `map[string]bool`. There is also an option of `map[string]struct{}` but this SO comment sums up nicely why it may be overkill: https://stackoverflow.com/a/57621291

pkg/settings/setting.go
```diff
@@ -256,3 +257,28 @@ func GetSettingByID(id string) string {
...
+		envVarMap[s.Name] = s
+	}
+
+	for _, s := range toCopy {
+		name := GetEnvKey(s.Name)
+		if _, ok := envVarMap[name]; !ok {
+			envVars = append([]v1.EnvVar{
+				{
+					Name:  name,
+					Value: s.Get(),
```
snasovich, line 279: Do we always have non-empty value? If not, do we want to skip adding empty settings?

pkg/systemtemplate/import.go
```diff
@@ -103,7 +102,8 @@ func SystemTemplate(resp io.Writer, agentImage, authImage, namespace, token, url
 		tolerations = templates.ToYAML(taints)
 	}
 
-	if cluster != nil && len(cluster.Spec.AgentEnvVars) > 0 {
+	if cluster != nil {
+		settings.MergeSettingsWithAgentEnvVars(cluster.Spec.AgentEnvVars)
```
snasovich, line 106: Is it possible for none of these settings to have values? If that the case we'll end up with AgentEnvVars being YAML representation of empty array vs. previously it being empty (period). Are we OK with this behavior change?

snasovich (changes requested):

pkg/settings/setting.go
jakefhyde: I really have to get out of the whole "arrays are pointers" thing. I do think this is a good topic for discussion though, we need to be storing them somewhere (preferably relating to the agent) in order to detect if a redeploy is necessary, but if it's the `AgentEnvVars` then we should think about whether or not we want to merge. We can just prepend the environment variables to `AgentEnvVars`, but this affects how they show up in "Edit Config/YAML".

snasovich: Yup. Also, thinking more about it the second suggestion (to have func that returns settings as env variables) may not make sense as then merging logic will move down to the caller. Not sure. But yeah, we need to discuss what you mentioned above.

pkg/settings/setting.go
jakefhyde: I think this depends on current expectations. For example, if you provision a downstream cluster and set `CATTLE_SERVER_VERSION` in agent env vars to empty, should it use the default (`dev`) or should it be considered invalid? Currently on `v2.6-head` if you set `CATTLE_SERVER_VERSION` to empty in downstream config, it uses the following logic to reset to default:

https://github.com/rancher/rancher/blob/7bc7959ea9f2961927d3db85113fa0c37eef1e72/pkg/controllers/dashboardapi/settings/settings.go#L29-L46

thedadams (dismissed):

snasovich (dismissed):

rmweir:
@jakefhyde which configmap? You could also add a settings handler that reenqueues the redeploy handler if one of those settings is changed.

jakefhyde:
@rmweir Disregard configmap nonsense, ended up scrapping that for a different approach. Yeah, adding a settings handler for that purpose makes alot of sense.

pkg/controllers/managementuser/settings/controller.go
```diff
@@ -2,59 +2,44 @@ package settings
...
 	userContext.Management.Management.Settings("").AddHandler(ctx, "copy-settings", h.onChange)
 	return nil
 }
 
-func (h *handler) onChange(key string, obj *v3.Setting) (runtime.Object, error) {
+func (h *handler) onChange(key string, obj *apis.Setting) (runtime.Object, error) {
 	if obj == nil || !toCopy[obj.Name] {
 		return nil, nil
 	}
+	h.cluster.Enqueue("", h.clusterID)
```
thedadams, line 35: Handle the error here.

pkg/controllers/managementuser/settings/controller.go
```diff
@@ -2,59 +2,44 @@ package settings
...
-		ApplyObjects(&v3.Setting{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: obj.Name,
-			},
-			Value:   obj.Value,
-			Default: obj.Default,
-		})
-	if err != nil {
-		return obj, fmt.Errorf("error applying setting object for cluster [%s]: %v", h.clusterID, err)
+func init() {
```
thedadams, line 39: nit: I think having the init function at the top of the file is better.

thedadams (changes requested):

pkg/controllers/managementuser/settings/controller.go
jakefhyde: `Enqueue` is actually a void method, which I found surprising at first but I'm assuming it has to do with the asynchronous nature of en-queuing objects.

thedadams: Oops. Sorry about that.

thedadams (dismissed):

ibuildthecloud:
@jakefhyde Can the settings controller be move to the `/management/`, not `/managementuser`. The important thing here is that managementuser controllers are created 1 for each cluster and there is no reason to do that anymore here.  We only need on controller.

rmweir (changes requested):
Waiting on handler to be moved as requested by @ibuildthecloud.

jakefhyde:
@rmweir Testing the controller changes now, will update when finished.

thedadams (approved):

pkg/controllers/management/settings/controller.go
```diff
@@ -0,0 +1,52 @@
...
+
+	management.Management.Settings("").AddHandler(ctx, "copy-settings", h.onChange)
+}
+
+func (h *handler) onChange(key string, obj *apis.Setting) (runtime.Object, error) {
+	if obj == nil || !toCopy[obj.Name] {
+		return nil, nil
+	}
+
+	clusters, err := h.cluster.Lister().List("", labels.NewSelector())
```
rmweir, line 41: nit: use labels.Everything() instead.

rmweir (approved):
LGTM

rmweir (approved):

Issue #34986, opened 2021-09-29
`rancher/machine` not found in `rancher-images.txt`


nickgerace:
Image `rancher/machine` is not found in `rancher-images.txt` for Rancher `v2.6.1-rc7`.

This image was built-in for v2.5.x. Provisioning v2 now uses that machine image, so it is likely missing from the file as a result of never previously needing to be there.

Relevant issue: https://github.com/rancher/rancher/issues/32403

thedadams:
@nickgerace I moved this to "Waiting for Rancher RC" so we could ensure that the machine image is in the images.txt file.

sowmyav27:
`rancher/machine:v0.15.0-rancher70` is available in rancher-images.txt file in `2.6.1-rc9`

PR #34989, opened 2021-09-29
Ensure rancher-machine image is added to images.txt

thedadams:
The rancher-machine image was mistakenly left out of the images.txt
file. This change will include it.

Issue:
https://github.com/rancher/rancher/issues/34986

### DIFF
--- a/pkg/image/resolve.go
+++ b/pkg/image/resolve.go
@@ -313,6 +313,7 @@ func setRequirementImages(osType OSType, imagesSet map[string]map[string]bool) {
 	switch osType {
 	case Linux:
 		addSourceToImage(imagesSet, settings.ShellImage.Get(), coreLabel)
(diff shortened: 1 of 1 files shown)


thedadams:
I ran `make ci` locally before submitting this to ensure the machine image is included in `rancher-images.txt`.

nickgerace (approved):

pkg/image/resolve.go
```diff
@@ -313,6 +313,7 @@ func setRequirementImages(osType OSType, imagesSet map[string]map[string]bool) {
 	switch osType {
 	case Linux:
 		addSourceToImage(imagesSet, settings.ShellImage.Get(), coreLabel)
+		addSourceToImage(imagesSet, settings.MachineProvisionImage.Get(), coreLabel)
```
rmweir, line 316: Could also be system as I think that's what other RKE related images are under but this works too.

rmweir (approved):
LGTM

Issue #34948, opened 2021-09-28
Services are deploying, but are not working properly (i.e. nodeport)


slickwarren:
**Rancher Server Setup**
- Rancher version:v2.6-head(06e19b0)
- Installation option (Docker install/Helm Chart):docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:n/a

**Information about the Cluster**
- Kubernetes version:any
- Cluster Type (Local/Downstream):downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):any node driver 

**Describe the bug**

* services are deploying, but are not working properly

**To Reproduce**
 
* provision an rke2 cluster using a node driver
* deploy a workload with a service (i.e. nodeport) -- QA has a test for this if you need it

**Result**
service for the workload will deploy, but link is not generating for the service/link is not working through `services` -> `nodeport link`

**see comment below for more details**

**Expected Result**

* services should deploy and work with the workload they are associated with

**Screenshots**

<img width="1467" alt="Screen Shot 2021-09-27 at 10 04 07 PM" src="https://user-images.githubusercontent.com/16691014/135026452-c31373aa-e02f-4cd4-83aa-9786f055b0de.png">
<img width="924" alt="Screen Shot 2021-09-27 at 10 03 49 PM" src="https://user-images.githubusercontent.com/16691014/135026455-e63fe3e5-1310-4526-900a-c2567afb8115.png">

**Additional context**
also tested hostport, nodeport, clusterIP, loadbalancer, none of which are working properly through dashboard.

yaocw2020:
![image](https://user-images.githubusercontent.com/7421463/135230453-901f0d8c-a9a1-4061-b762-8ea3c865a867.png)

There are two services. The first one created automatically when adding the deployment does not work while the second one created manually works. 

The second one has one more selector `workloadID_ngx-nodeport: "true"`, which matches no pod.

yasker:
@slickwarren can you confirm that's due to a wrong selector?

sowmyav27:
**On 2.6.0**

I see nodeport - deployment working. The additional selector as mentioned [here](https://github.com/rancher/rancher/issues/34948#issuecomment-929991464) in this comments is not added to the service that is created.

slickwarren:
yes this is not harvester-specific, I will move this to reopened. it is happening with any rke2 setup

[...]

PR #34993, opened 2021-09-30
add WorkloadID selector only if service is created by norman

kinarashah:
**Issue:** 
Services stopped working in v2.6-head because they have an extra selector `workloadID_`. 
https://github.com/rancher/rancher/issues/34948 

**NOTE:** 
This issue doesn't happen in v2.6.0 because the controller would error out here https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementagent/targetworkloadservice/target_workload_service.go#L123 due to an existing bug. This bug was recently fixed in UI so the logic now proceeds and adds selector the service.  

**Root cause:** 
Vue UI creates services for workloads itself vs backend used to create different type of services in Ember. The default service created for Ember always had Noop annotation https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementagent/workload/workload.go#L249 so this controller wouldn't try to add selector to the service. https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementagent/targetworkloadservice/target_workload_service.go#L115 (This selector is for a specific use case where we could create service pointing to another workload in Ember). 

**Fix:** 
This logic only needs to run for services created by norman for Ember, so look for `"cattle.io/creator":norman` label and return if not present. 

**Validation:**
- Tested that service works with the fix in Vue 
- Tested that service continues to work in Ember

### DIFF
--- a/pkg/controllers/managementagent/targetworkloadservice/target_workload_service.go
+++ b/pkg/controllers/managementagent/targetworkloadservice/target_workload_service.go
@@ -29,6 +29,8 @@ import (
 
 const (
 	WorkloadIDLabelPrefix = "workloadID"
+	creatorLabelKey       = "cattle.io/creator"
+	creatorNorman         = "norman"
 )
 
 var workloadServiceUUIDToWorkloadIDs sync.Map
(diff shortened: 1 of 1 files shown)


nickgerace (approved):

nickgerace (changes requested):
Needs `go fmt`

pkg/controllers/managementagent/targetworkloadservice/target_workload_service.go
```diff
@@ -74,6 +76,15 @@ func (c *Controller) sync(key string, obj *corev1.Service) (runtime.Object, erro
 		return nil, nil
 	}
 
+	// add WorkloadID selector only if service is created by norman
+	if obj.Labels == nil {
```
StrongMonkey, line 80: Nit: it can probably be one liner
```
if obj.Labels == nil || obj.Labels[creatorLabelKey] != creatorNorman {
  return obj, nil
}
```

StrongMonkey (approved):

snasovich (approved):

nickgerace (approved):

Issue #34971, opened 2021-09-29
[CI] Rancher fails to start intermittently, leading to 1 hour test timeouts


nickgerace:
This has been happening much more frequently with recent changes to `release/v2.6` leading to `v2.6.1`. This could reach a point where it becomes a release blocker if not resolved.

```
panic: indexer conflict: map[byPod:{}]

goroutine 3208 [running]:
k8s.io/apimachinery/pkg/util/runtime.Must(...)
	/go/pkg/mod/k8s.io/apimachinery@v0.21.0/pkg/util/runtime/runtime.go:171
github.com/rancher/rancher/pkg/generated/controllers/catalog.cattle.io/v1.(*operationCache).AddIndexer(0xc008e31080, 0x4181464, 0x5, 0x439b4f8)
	/go/src/github.com/rancher/rancher/pkg/generated/controllers/catalog.cattle.io/v1/operation.go:247 +0x127
github.com/rancher/rancher/pkg/controllers/dashboard/helm.RegisterOperations(0x48f4aa8, 0xc008c81240, 0x4959c58, 0xc0016e2420, 0x4942700, 0xc00a0f6460, 0x4940240, 0xc00a0f64d0)
	/go/src/github.com/rancher/rancher/pkg/controllers/dashboard/helm/operations.go:42 +0x1b8
github.com/rancher/rancher/pkg/controllers/dashboard/helm.Register(0x48f4aa8, 0xc008c81240, 0xc000c401a0)
	/go/src/github.com/rancher/rancher/pkg/controllers/dashboard/helm/register.go:22 +0x3c5
github.com/rancher/rancher/pkg/controllers/dashboard.Register(0x48f4aa8, 0xc008c81240, 0xc000c401a0, 0xc008e30d01, 0x39b61a0, 0x69d2b38)
	/go/src/github.com/rancher/rancher/pkg/controllers/dashboard/controller.go:23 +0x5a
github.com/rancher/rancher/pkg/rancher.(*Rancher).Start.func1.1(0x48f4aa8, 0xc008c81240, 0x39b61a0, 0x69d2b38)
	/go/src/github.com/rancher/rancher/pkg/rancher/rancher.go:245 +0x4e
github.com/rancher/rancher/pkg/wrangler.(*Context).StartWithTransaction(0xc000c401a0, 0x48f45d8, 0xc003346a80, 0xc0044c7f58, 0x0, 0x0)
[...]

PR #35001, opened 2021-09-30
Add restart hack to provisioning tests

ibuildthecloud:
Also if the Rancher dies echo the log so we can debug why it failed to start


nickgerace:
https://github.com/rancher/rancher/issues/34971

nickgerace (approved):

PR #35003, opened 2021-09-30
Handle flapping value between rancher and fleet

ibuildthecloud:
When fleet bootstraps the local cluster it uses a non standard secret name
"local-cluster" because the loopback connect is handled different from
any other cluster. Due to a race condition on start it was possible that
Rancher would first create fleet-local/local cluster with no kubeconfig
secret.  If the kubeconfig secret is blank fleet would try to generate
one and would fail for the local cluster.  Instead if rancher creates
the local cluster first ensure that we also set the non standard secret
"local-cluster".


nickgerace (approved):

StrongMonkey (approved):

nickgerace:
https://github.com/rancher/rancher/issues/35004

PR #35005, opened 2021-09-30
Turn Harvester feature off by default

thedadams:


cbron (approved):
Lgtm, as long as this is only place.

thedadams:
I'm a little concerned because the feature doesn't seem to be referenced anywhere else in Rancher's code base.

thedadams:
According to @deniseschannon, this no longer needed.

Issue #34954, opened 2021-09-28
Fleet deploys repo to Harvester Clusters


nwmac:
See: https://zube.io/rancher/dashboard-ui/c/4266

When the user selects 'all clusters' in the UI for the target for a git repo, this deploys to Harvester clusters as well as regular k8s clusters.

The UI creates the resource with `spec.targets = [{ clusterSelector: {} }];` - the backend should not deploy to Harvester clusters in this case.

yasker:
@cloudnautique What do you think we should do for v2.6.1? Disable Fleet to Harvester entirely seems to make sense but I am not sure if it's a blocker for v.2.6.1

Jono-SUSE-Rancher:
@yasker as discussed, I moved this to v2.6.2. I will have @SheilaghM sync up with @StrongMonkey and @nickgerace to see what the best course of action is here.

StrongMonkey:
This is more like a UI change. By default UI can create gitrepos that exclude some labels(there should some labels that are unique to harvester cluster).

StrongMonkey:
Fleet doesn't have knowledge around which cluster it should/should not deploy to, it onlys looks for the targets(labels) that user specifies. It wouldn't make sense for fleet by default to exclude some pre-populate sets of labels in the backend.

yasker:
Harvester baremetal clusters have a special `provider` field shows it's Harvester baremetal cluster. See https://github.com/rancher/rancher/issues/34880#issuecomment-927218920 for details.

yasker:
To make it easier to filter the Harvester cluster like what Daishan did in https://github.com/rancher/rancher/issues/34954#issuecomment-931653234 , we will need to automatically add a label to any cluster with `provider=harvester`, so the filter can be used.

yasker:
As discussed with @StrongMonkey , we will add the following logic in the backend:
1. If a cluster with `provider=""`, no change to the label `cluster.harvesterhci.io` regardless it exists or not.
2. If a cluster with `provider="harvester"`, label `cluster.harvesterhci.io` will be added if it doesn't exist.
3. If a cluster with `provider != "" && provider != "harvester"`, label `cluster.harvesterhci.io` will be removed if it exists.

This should also fix the potential inconsistency issue in the UI since filters are done via the label only at the moment.

StrongMonkey:
[...]

PR #35007, opened 2021-09-30
Add provider labels to v3 cluster

StrongMonkey:
We need to add provider label `provider.cattle.io` to v3 cluster so that we know the type of the cluster based its label. This is used by harvester team to filter out clusters in UI/fleet based on labels. It also provides a way to filter clusters based on provider label without inspecting status.provider.

Related 
https://github.com/rancher/rancher/issues/34954
https://github.com/rancher/rancher/issues/34880
cc @yasker

### DIFF
--- a/pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
+++ b/pkg/controllers/dashboard/kubernetesprovider/kubernetesprovider.go
@@ -10,6 +10,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	providerKey = "provider.cattle.io"
+)
+
 type handler struct {
 	ctx                context.Context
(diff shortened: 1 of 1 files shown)


dramich (commented):
Why are we adding the same data twice? Based on the issues Harvester team is fine using the provider field

StrongMonkey:
@dramich It is for displaying purpose. 
>So before the cluster import completed, the cluster imported from Cluster Management page will be shown in the Cluster Management page, the cluster imported from Virtualization Management page will be shown in the Virtualization Management page.

So when you first create cluster in Virtualization Management, they added the label so that it could show up in UI. But if later the provider is detected and it is not a harvester cluster, it will be moved to elsewhere.

dramich (approved):

Issue #34997, opened 2021-09-30
RKE2 cluster never becomes healthy with certain machine pool names


janeczku:
2.6.1-rc5

**To Reproduce**

1. Create a RKE2 cluster with any node driver (including Harvester)
2. Name the cluster `test-do` 
2. Configure a single machine pool using the following machine pool name:  `harvester-v.0.3.0-rc1-creds-pool1`

**Result**

The cluster remains unhealthy. Opening cluster explorer fails with error 500.

All cluster nodes are provisioned with conflicting (= identical) node names: `test-do-harvester-v0`

**Screenshots**

<img width="1437" alt="Screenshot 2021-09-30 at 11 26 55" src="https://user-images.githubusercontent.com/3813921/135426607-47da14d1-da5e-47fe-b4f4-25cce516bca5.png">

<img width="1507" alt="Screenshot 2021-09-30 at 11 25 17" src="https://user-images.githubusercontent.com/3813921/135426129-cd5c993f-7338-4199-b124-ef02e30e206c.png">

**Root cause**

Rancher node names are not correctly sanitizied and truncated, resulting in node names that are neither unique nor satisfy the Kubernetes requirement Kubernetes of DNS label naming standard (RFC 1123). See https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

thedadams:
This isn't a problem with Kubernetes. It is a problem with cloud-init setting the hostname of the machine to something other than what we want.

thedadams:
For reference: https://github.com/canonical/cloud-init/blob/392c3262dcb4108b42f09e1369d9072502302361/cloudinit/util.py#L1014-L1018

thedadams:
### Root cause
The issue is caused in RKE2 provisioning when creating a node pool with a `.` in the name.

### What was fixed, or what changes have occurred
Now, Rancher will replace `.` with `-` when creating the nodes in the cloud provider. This is because cloud-init will split the name on `.` and set the node's hostname to the first chunk. This would result in all nodes in a node pool having the same name in Kubernetes, which is bad.

### Areas or cases that should be tested
Provisioning an RKE2 cluster from Rancher

### What areas could experience regressions?
Only RKE2 provisioning.

### Are the repro steps accurate/minimal?
Yes:
1. Create an RKE2 cluster with a `.` in the name of its node pool(s)
2. The cluster will provision and go into an error state.
[...]

PR #35020, opened 2021-10-01
Remove periods from machine name passed to machine job

thedadams:
For certain cloud providers, the name of the machine passed to
rancher-machine gets set to the hostname for the machine. This hostname
is then used when deploying RKE2/K3S.

If the hostname of a machine has a period in it, cloud-init will split
the hostname on '.' and set the hostname of the machine to the first
chunk. For example, if the name of the machine is
'my.really.good.machine', then cloud-init will set the hostname to 'my'.
This causes a problem in provisioningv2 if the user puts a period in the
name of a machine pool. Then all the corresponding nodes in that machine
pool will have the same name because of the chunking behavior.

This change will replace all periods in the machine name passed to
rancher-machine to dashes to avoid this problem.

Issue:
https://github.com/rancher/rancher/issues/34997


jakefhyde (approved):
LGTM

PR #35034, opened 2021-10-04
Automation - Airgap fix order of admin password of rancher docker command

izaac:
This works on 2.5 but as we're using now a new variable for the bootstrap password the order matters. So I've reordered the variables order to match the proper parameters on the docker command.


slickwarren (approved):
LGTM

sowmyav27 (approved):
LGTM

PR #35038, opened 2021-10-05
[Forwardport]: Update error

rmweir:
**Original PR:**
https://github.com/rancher/rancher/pull/34939


cbron:
We don't use master branch right now.

PR #35044, opened 2021-10-05
Update roleBinding creation for provisioning clusters

dramich:


cmurphy (commented):
Will this leave behind orphan rolebindings with the old name format on upgrade?

dramich:
> Will this leave behind orphan rolebindings with the old name format on upgrade?

Great question. No, it won't. Apply has its own style of owner reference, so it will go delete the existing ones and replace with the new roleBinding since the names no longer match.

cmurphy (approved):

prachidamle:
Had the same question about existing rolebindings on a setup created with old format. What will cause them to be revisited and deleted? If due to this problem some users have lost access already, how will they gain it back? Do we recreate the bindings on startup?

dramich:
> Had the same question about existing rolebindings on a setup created with old format. What will cause them to be revisited and deleted? If due to this problem some users have lost access already, how will they gain it back? Do we recreate the bindings on startup?

The `apply` being used has its owner owner references so when it reconciles what object you want vs what is found it will cleanup the no longer desired object and create the newly named object. We sync all PRTBs and CRTBs on startup so everything will get done at that time. In most cases it seems users still have another role granting them permissions to see the provisioning cluster but it might not always be true.

Issue #22761, opened 2019-09-10
Duplicate project names


mattmattox:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
- Deploy Rancher v2.2.8
- Create project using UI with name "Test"
- Create another project using UI with name "Test"

**Result:**
It will let you create two projects with the same name.
![image](https://user-images.githubusercontent.com/17145893/64628809-af000200-d3b7-11e9-857d-da00f8c235c6.png)

**Other details that may be helpful:**
::First Project::
"name": "Test",
"namespaceId": null,
"podSecurityPolicyTemplateId": "",
"state": "active",
"transitioning": "no",
"transitioningMessage": "",
"type": "project",
"uuid": "3768af20-d3e1-11e9-9148-0050569e16da"

::Second Project::
"name": "Test",
"namespaceId": null,
"podSecurityPolicyTemplateId": "",
"state": "active",
"transitioning": "no",
"transitioningMessage": "",
"type": "project",
"uuid": "348f2b1d-d3e1-11e9-8783-0050569e8b4b"

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v2.2.8
- Installation option (single install/HA): HA

genexpr:
We can forbid duplicate project names for new projects, thereby breaking the existing behavior, which, as I learned, is deliberate.

PR #35071, opened 2021-10-07
[DNM v2.6.2] On project create, check if a project with a given name exists

maxsokolovsky:
Issue [#22761](https://github.com/rancher/rancher/issues/22761)
The goal of the change is to no longer allow to create a project with a name of an existing project.
Note that names of projects are not case-sensitive, so users are still allowed to create a `default` project in addition to one named `Default`.


genexpr:
@StrongMonkey @nickgerace, I will raise the concern in the issue comments. But I don't know why exactly users would want to create projects with duplicate names, and why we would allow it.

genexpr:
After a discussion in the PR comments, we concluded that the current behavior must remain unchanged, as it is deliberate. Thanks, @vincent99, @StrongMonkey, @nickgerace. Closing the issue and PR.

Issue #34060, opened 2021-08-08
Validate Rancher supports k8s 1.22


deniseschannon:
Due to the dropped APIs, there are significantly more checks needed to validate that k8s 1.22 can be supported. 

**Supporting installing Rancher onto k8s 1.22 clusters**
- [x] Validating runtime resources - https://github.com/rancher/rancher/issues/33222 
- [x] Updating Helm Server chart - https://github.com/rancher/rancher/issues/34058 
- [x] Ingress support - https://github.com/rancher/rancher/issues/33312 and https://github.com/rancher/rancher/issues/30053
[assigned] Dropping Component Status Support - https://github.com/rancher/rancher/issues/34055

**Supporting Downstream k8s 1.22 clusters**
[assigned] Supporting k8s 1.22 into Rancher to support 1.22 downstream clusters (client-go for 1.22) - https://github.com/rancher/rancher/issues/34054
- [x] Adding k8s 1.22 into KDM/RKE - https://github.com/rancher/rancher/issues/33114
[assigned] Adding k8s 1.22 for RKE2/k3s - https://github.com/rancher/rancher/issues/34056
- [x] Validating k8s 1.22 works for EKS - https://github.com/rancher/rancher/issues/33623
- [x] Validating k8s 1.22 works for AKS - https://github.com/rancher/rancher/issues/33622
- [x] Validating k8s 1.22 works for GKE - https://github.com/rancher/rancher/issues/33621

**Supporting Applications in k8s 1.22**
- [x] Continuous Delivery/Fleet - https://github.com/rancher/rancher/issues/33112
- [ ] Apps and Marketplace Apps - https://github.com/rancher/rancher/issues/34057
- [x] Legacy Apps - https://github.com/rancher/rancher/issues/34059

HoseinGhanbari:
unfortunately there is no warning in the official docs noticing people like me that you cannot use the latest version of k3s which is now `v1.22.3+k3s1` to run the latest version of rancher until we, ourselves, reach to this ...
```
[FATAL] the server could not find the requested resource
```

nilathedragon:
Just ran into this when upgrading my RKE2 Cluster... It's a bit inconvenient to have latest stable RKE2 not work with latest stable Rancher.

davidgibbons:
Just to echo the above statements, would have loved a more obvious warning about compatibility before I wasted my night trying to deploy this on the latest stable RKE2

kinarashah:
[...]

PR #35117, opened 2021-10-14
Updated CRDs to use version v1 for k8s 1.22 support

snasovich:
For https://github.com/rancher/rancher/issues/33222

Replaced CRD API from v1beta1 to v1 to support k8s v1.22, including bumping Lasso and Norman versions to include the changes from https://github.com/rancher/lasso/pull/10 and https://github.com/rancher/norman/pull/401
Confirmed that there are no CRD version deprecation warnings during Rancher startup and minimal operation.

The updates also include files in `pkg/controllers/management/test/e2e`, however: 1.) these tests are currently skipped (see https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/test/e2e/authz_test.go#L388); 2.) even after commenting out this line they fail even before these changes are applied. However, confirmed that running these tests no longer outputs any CRD version deprecation warnings (they do prior to applying these changes).

Note that apiextensions.k8s.io/v1 for CRD is supported from k8s v1.16 making it OK to do straight change as Rancher v2.6 release to support k8s 1.18+.


snasovich:
> Might also be worth updating these examples for completeness:
> 
> [`release`/v2.6/pkg/catalog/example](https://github.com/rancher/rancher/tree/release/v2.6/pkg/catalog/example) [`release`/v2.6/pkg/nodeconfig/examples](https://github.com/rancher/rancher/tree/release/v2.6/pkg/nodeconfig/examples)

@cmurphy , following discussion with @dramich these examples are outdated and no longer needed. Good catch though, we don't want additional references to v1beta1 CRDs if we don't have to have them.

snasovich:
> @snasovich Question about `Note that apiextensions.k8s.io/v1 for CRD is supported from k8s v1.16 making it OK to do straight change as Rancher v2.6 release to support k8s 1.18+.` Does this mean k8s 1.16 clusters will start erroring out on 2.6.2 and we expect users to upgrade k8s to >=1.18 before upgrading Rancher?

@kinarashah , technically the requirement is to be on 1.16+. I was just noting that 2.6.x (now) officially supports k8s 1.18+ so we don't need to worry about supporting versions 1.15 and less. I have not touched `kubeVersion` constraint in `Chart.yaml` though yet - I'm sure we will be updating it as we get the rest of 1.22 support items done.

jsbrain:
Is there going to be a pre-release anytime soon or do we need to wait for the v2.6.3?

snasovich:
@jsbrain , the `v2.6-head` image is built off of the latest changes that are merged to `release/v2.6` branch, so you can always use it for testing without waiting for a full release or a release candidate.

However, please note that this change is just one of the changes needed to implement k8s 1.22 support for Rancher. Umbrella issue: https://github.com/rancher/rancher/issues/34060

jsbrain:
@snasovich, oh I see, my bad. Thanks for the heads up!

Issue #34890, opened 2021-09-23
s3 etcd backups taking longer than 5 minutes cause backup loop


nickgerace:
Backups that take longer than 5 minutes will cause Rancher to start a new backup and delete the one that is currently running. Since they all take longer than 5 minutes it will continue this cycle indefinitely.

**Rancher Server Setup**
- Rancher version: 2.6.3

**Repro steps:**
* Have RKE1 provisioned downstream cluster
* Fill etcd with lots of data (e.g. 1 GB+) - enough so it takes more than 5 minutes to complete on all of the etcd nodes.
* Configure recurring snapshots for this downstream cluster, including setting `rancher_kubernetes_engine_config.services.etcd.backup_config.timeout` to high enough value for backups to actually complete on etcd side (e.g. 900 for 15 minutes).
* Wait for a backup to start (depending on interval configured) and wait 5 minutes.

**Actual behavior:**
Since the backup doesn't complete before the sync interval, a new backup will be started. This will repeat indefinitely until a backup is completed in under 5 minutes.

**Expected behavior:**
Backups to complete.

Original description (no longer relevant as fix approach changed per https://github.com/rancher/rancher/issues/34890#issuecomment-943872638):
> The `clusterBackupCheckInterval` is hardcoded to 5 minutes and should be user-changeable via the following
> - Helm (likely)
> - ENV VAR (could be good idea)
> - ConfigMap (eh)
> - etc.
> 
> Moreover, we should considered adding a "maximum retry count" so that the check interval does not infinitely continue.

JIRA ID: SURE-3305

nickgerace:
The "part two" to this issue exists and is listed below, but requires a design discussion and investigation. It is _not_ required for solving this issue.

"Part two" issue: https://github.com/rancher/rancher/issues/34892

bentastic27:
Less intrusive PR that should be far more reliable. Basically a check for running backups instead of bulldozing them and starting over.

https://github.com/rancher/rancher/pull/35132

It doesn't make the clusterBackupCheckInterval malleable, but IMO it makes it far less an issue, perhaps even unnecessary. It appears that changing the check interval it could interfere with the backup interval setting. Let's say that a user sets a backup to run every hour, but they set the check interval to 2 hours. It looks as though that the backup would only run 2 hours if that was the case, so the check interval can be considered resolution of sorts in this case:

By default a backup will run every hour with 5 minutes of slop.

snasovich:
[...]

PR #35120, opened 2021-10-14
Cluster backup check interval bug

bentastic27:
rancher/rancher#34890


bentastic27:
nvm breaks stuff, will try again.

PR #35145, opened 2021-10-15
Fix nil dereference for cluster cleanup

cmurphy:
https://github.com/rancher/rancher/issues/35160


cmurphy:
https://github.com/rancher/rancher/issues/35160

Issue #32045, opened 2021-04-12
Access control cache sees issue when adding namespaces


anupama2501:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Spin a rancher on version 2.4.15
2. Create users u1 and u2 as standard users
3. Create a cluster with 3worker, 1etcd and 1 cp
4. Add a project as admin and add u2 as Project Owner and Project member
5. Upgrade the cluster to the latest v2.5.8-rc2
6. Once the server is active, create a namespace logged in as u2
7. Namespace is stuck in activating
8. Create another namespace and the ns is stuck in activating too. 
9. If we refresh the page, the namespace in step7 appears but namespace in step 8 disappears.

**Result:**
All the created namespaces should appear in the project

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 
Before upgrade: v2.4.15
After upgrade: v2.5.8-rc2
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): Infrastructure provider digital ocean

**Other details that may be helpful:**
Screen recording for the same:

https://user-images.githubusercontent.com/38144301/114331821-bc342400-9af9-11eb-8039-f5bf9c6ced3b.mov

This is not seen on a fresh install of 2.5-head,2.5.7 and not happening always.. 
Also, if we upgrade from 2.4.15 to 2.5-head -- issue is not seen

anupama2501:
@ryansann This is also seen on v2.5-d1083bc63e4ddde70ffadd09bf0133ae5862b5fe-head

1. Create two users u1 as Admin/restricted admin and u2 as standard user
2. Create a cluster logged in as u2 and add a project to u1 to this project with custom roles -- only enable create namespaces
3. Now login as u1 and verify, he has access to all the projects
4. Change the user u1 from admin to standard user logged in as local cluster admin
5. Log in as u1 and create a namespace. We can see the same issue happening 
6. If I verify the namespaces created from kubectl, I see all the created namespaces:

```
ns1               Active   8m55s
ns2               Active   8m45s
ns3               Active   8m31s
ns4               Active   6m13s
ns5               Active   6m2s
ns6               Active   35s
```

If we log out and log in as u1, we can then see all the namespaces

anupama2501:
@ryansann I'm not seeing it on v2.5.7. 
[...]

PR #35172, opened 2021-10-19
[2.5] Fix the problem of handler registration failure caused by inconsistent context

JacieChao:
In Rancher HA mode, each replica will create an [access control cache](https://github.com/rancher/rancher/blob/release/v2.5/pkg/clustermanager/manager.go#L232) for each downstream cluster no matter it's the clusterOwner or not. In access control cache, we start a [handler](https://github.com/rancher/steve/blob/7224dc21013de5499e17cd065dfccef9400bc077/pkg/accesscontrol/role_revision_index.go#L18) here to handle role revisions change when clusterrole/role changed at the downstream cluster. 

For further investigation, I found some inconsistent context to start the access control cache. 
1. https://github.com/rancher/rancher/blob/release/v2.5/pkg/clustermanager/manager.go#L480
2. https://github.com/rancher/rancher/blob/release/v2.5/pkg/clustermanager/manager.go#L457
3. https://github.com/rancher/rancher/blob/release/v2.5/pkg/clustermanager/manager.go#L95

Sometimes it using context.Background() to new AccessControl, but sometimes it using ctx which inherited from wrangler context(This context is a HandlerTransaction context). When it using context.Background() to create an access control cache, the handler could start correctly, so that some replicas got the correct result. If it using ctx which inherited from wrangler context(ReigsterHandler type), the handler in access control cache will never start.

Then I found the HandlerTransaction context for steve controller. 
- When we called [RegisterHandler](https://github.com/rancher/lasso/blob/master/pkg/controller/sharedcontroller.go#L111) here, if the ctx is HandlerTransaction type, will add handler function to [todo list](https://github.com/rancher/lasso/blob/6c6cd7fd6607e280ef3addec9901a8f4dbb12295/pkg/controller/transaction.go#L27), if not, it will start handler function [directly](https://github.com/rancher/lasso/blob/6c6cd7fd6607e280ef3addec9901a8f4dbb12295/pkg/controller/transaction.go#L20).
- If the ctx is HandlerTransaction type, we can call [Commit](https://github.com/rancher/lasso/blob/6c6cd7fd6607e280ef3addec9901a8f4dbb12295/pkg/controller/transaction.go#L37) function to start all handlers here.

So if it using ctx which inherited from wrangler context and has already called Commit function when rancher started, all handlers register after that will never start.

We create a new [HandlerTransaction](https://github.com/rancher/rancher/blob/08ad9f793a2345478a26a514fca021c8e648f2f3/pkg/clustermanager/manager.go#L214) here to start cluster controllers but not using this transaction to create access control. I think it's better to use the same context to register handlers to make sure all handlers can be started consistently.

https://github.com/rancher/rancher/issues/31982

### DIFF
--- a/pkg/clustermanager/manager.go
+++ b/pkg/clustermanager/manager.go
@@ -229,7 +229,7 @@ func (m *Manager) doStart(rec *record, clusterOwner bool) (exit error) {
 		defer close(done)
 
 		logrus.Debugf("[clustermanager] creating AccessControl for cluster %v", rec.cluster.ClusterName)
-		rec.accessControl = rbac.NewAccessControl(rec.ctx, rec.cluster.ClusterName, rec.cluster.RBACw)
+		rec.accessControl = rbac.NewAccessControl(transaction, rec.cluster.ClusterName, rec.cluster.RBACw)
 
 		err := rec.cluster.Start(rec.ctx)
 		if err == nil {


ryansann (dismissed):
I deployed HA validated that this is a problem and then upgraded to a custom image that contains the fix and saw the correct behavior. I also was not able to reproduce #32045 with the new changes, so this looks to have fixed that issue as well. Approving, but we'll need one more review on this.

rmweir (changes requested):
The logic looks fine, however this looks like it was done intentionally. This also is basically "framework related", so my approval is contingent on approval from @ibuildthecloud.

pkg/clustermanager/manager.go
```diff
@@ -229,9 +229,9 @@ func (m *Manager) doStart(rec *record, clusterOwner bool) (exit error) {
 		defer close(done)
 
 		logrus.Debugf("[clustermanager] creating AccessControl for cluster %v", rec.cluster.ClusterName)
-		rec.accessControl = rbac.NewAccessControl(rec.ctx, rec.cluster.ClusterName, rec.cluster.RBACw)
+		rec.accessControl = rbac.NewAccessControl(transaction, rec.cluster.ClusterName, rec.cluster.RBACw)
```
ibuildthecloud, line 232: This is the real fix.

pkg/clustermanager/manager.go
```diff
@@ -229,9 +229,9 @@ func (m *Manager) doStart(rec *record, clusterOwner bool) (exit error) {
...
 
-		err := rec.cluster.Start(rec.ctx)
+		err := rec.cluster.Start(transaction)
```
ibuildthecloud, line 234: This should remain as rec.ctx.

JacieChao: fixed

ibuildthecloud (approved):

ryansann (approved):

rmweir (approved):
LGTM

PR #35200, opened 2021-10-19
Add a dualstack-capable QA image

rancher-max:
This should be built as `ranchertest/mytestcontainer:dualstack`

It can be validated for now using `maxross/mytestcontainer:dualstack`. I ran this in a dualstack capable cluster by deploying a hostport (8081) workload, then curling the ipv6 address like `curl [<redacted>]:8081/name.html`, which was successful. 

The only change between this and our main `ranchertest/mytestcontainer` image is that this specifics `listen [::]:80;` in the nginx.conf server.

Edit: This is now intended as an enhancement to both `ranchertest/mytestcontainer` and `ranchertest/mytestcontainer:unprivileged`. Those images should be retagged and pushed with these changes, and should now be valid when used with only ipv4 or with ipv4+ipv6 dualstack.

Closes: https://github.com/rancher/qa-tasks/issues/144

### DIFF
--- a/tests/validation/tests/Dockerfiles/testcontainer/Dockerfile
+++ b/tests/validation/tests/Dockerfiles/testcontainer/Dockerfile
@@ -1,5 +1,4 @@
 FROM nginx
-MAINTAINER Sangeetha Hariharan "https://github.com/sangeethah"
 
 RUN apt-get update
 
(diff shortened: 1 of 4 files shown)


rancher-max:
Also I should add, for what it's worth, that I can just add this conf file to our other image setups to just allow for dualstack on the main image, but I wasn't sure if we wanted that or not. If so, let me know and I'll just update this to do that and we can update our current images and not use a specific tagged one. It shouldn't cause any issues as I'm able to use my dualstack image on setups that are not dualstack capable as well 🙂

bmdepesa (approved):
LGTM

Issue #34701, opened 2021-09-10
Add support for annotation that prevents helm install and v1 App deletion when deleting v3 App


rmweir:
**Is your feature request related to a problem? Please describe.**
In the long run we'd like to implement a means of migrating an app that was installed in the ember ui to be exclusively manageable in the dashboard UI. Being able to delete the ember app without uninstalling the app's resources or the dashboard app is a first step in that direction and should facilitate a more manual, intermediate process for achieving the migration.

**Describe the solution you'd like**
Support an annotation to a v3 App that indicates a helm uninstall should not be ran on deletion of the v3 App and the v1 App (visible in dashboard) should also NOT be deleted.

**Related issue:**
https://github.com/rancher/rancher/issues/34700

rmweir:
Usually I always add a milestone to issues as a placeholder but for this issue and a related issue I will leave that for @cloudnautique.

rmweir:
This is best tested as the final step of testing https://github.com/rancher/rancher/issues/34700. Once the migration is done, this can be used to delete the norman app without uninstalling the chart or deleting the app in dashboard.

rmweir:
### Steps
__Note:__ This makes the most sense to be done after testing an upgrade for https://github.com/rancher/rancher/issues/34700, once the app has been migrated (meaning cluster repo was created and proper annotations were added to new app's spec.chart.metadata.annotations).

1. Add "cattle.io/skipUninstall": "true" annotation to ember app using kubectl or API UI.
2. Delete the app.

**Expected:**
App is deleted but the underlying resources are not uninstalled by helm.

### Root cause
This is a complementary issue to https://github.com/rancher/rancher/issues/34700. Once the new dashboard apps can be used to edit apps that were once installed through ember apps, there needs to be a way of deleted the legacy ember app without uninstalling the underlying chart.

### What was fixed, or what changes have occurred
Now, an annotation,"cattle.io/skipUninstall": "true",  can be added to prevent the helm uninstall part of the legacy app deletion logic. This can be done using kubectl or API UI in ember.

### Areas or cases that should be tested
Just uninstalling an ember app without the resources actually being uninstalled by helm. It makes the most sense to do this as the last step of testing an upgrade for https://github.com/rancher/rancher/issues/34700.

### What areas could experience regressions?
Any ember app deletion. Should ensure that without the annotation that apps are deleted along with their resources as expected.

### Are the repro steps accurate/minimal?
Yes.

anupama2501:
[...]

Issue #34700, opened 2021-09-10
Cannot upgrade apps with deployments that were created in ember


rmweir:
**Rancher Server Setup**
- Rancher version: v2.6.0
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
     RKE2

**Information about the Cluster**
- Kubernetes version:
v1.21.4+rke2r2
- Cluster Type (Local/Downstream):
  Downstream 
  - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):
    Rancher provisioned AWS cluster

**Describe the bug**

Cannot upgrade apps with deployments if app was created in ember.

**To Reproduce**
 
1. Deploy an old version of a helmv3 app in ember that contains a deployment, such as `aws-event-sources` from helm3-library. Do not use the newest version, must use an app with multiple versions available.
2. Add a cluster repo that uses same url/branch of helm3-library catalog.
3. Install the same version of app from 1 and from the helm3-library clusterrepo using dashboard.
4. Once app from 3 becomes active, open networking tab, update app version, and copy request.
5. Replace name/namespace from request in step 4. with name/namespace of app deployed in step 1.
6. Send request.

**Result**
Encounter error like the following:
`Error: UPGRADE FAILED: cannot patch "aws-event-sourcesember-controller" with kind Deployment: Deployment.apps "aws-event-sourcesember-controller" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"aws-event-sourcesember", "app.kubernetes.io/name":"aws-event-sources"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable`

**Notes**
This is probably happening due to the `io.cattle.field/appId` matchLabel selector on the deployment. It seems as though norman automatically injects this into the deployment and steve is trying to apply a manifest that doesn't include it. Since these are immutable it causes an error. This likely is the case with deployments deployed without apps as well and I believe that is covered by this issue: https://github.com/rancher/rancher/issues/31235.

**Expected Result**

A successful upgrade.

**Additional context**

Step 2 through 5 are a work around for functionality we want to support. We want to support migrating apps from norman to steve. The steps worked on apps such as "alcide-advisor-cronjob", this is because it is mostly just a cronJob which doesn't have this issue.

rmweir:
[...]

PR #35202, opened 2021-10-19
Apps migration

rmweir:
**Problem:**
Customers want to be able to migrate apps.project to dashboard apps (apps.catalog). This is blocked by two issues:
Helm3 apps that were installed by creating an app.project may
have had some of their resources changed with kustomize. Some
of these changes were to matchLabel selectors which are
immutable. This caused any updates to a migrated app to
result in an error because the manifests that were
being applied would try to remove the matchLabels, which
cannot be changed.

Need to be able to delete outdated apps (apps.project) without uninstalling resources.

**Solution:**
Now, migrated apps will continue using
kustomize to make the same changes to the manifests as
before.

Add annotation that skips helm uninstall. Revisions will still
be deleted and app will be removed from namespace's app
annotation. This is instended to be used as part of the
upgrade process from apps.project to apps.catalog.

**Depends on(merge first):**
https://github.com/rancher/shell/pull/16

**Issues:**
https://github.com/rancher/rancher/issues/34700
https://github.com/rancher/rancher/issues/34701

### DIFF
--- a/pkg/controllers/managementuserlegacy/helm/controller.go
+++ b/pkg/controllers/managementuserlegacy/helm/controller.go
@@ -269,29 +269,31 @@ func (l *Lifecycle) Remove(obj *v3.App) (runtime.Object, error) {
 			return obj, err
 		}
 	}
-	tempDirs, err := createTempDir(obj)
-	if err != nil {
-		return obj, err
-	}
-	defer os.RemoveAll(tempDirs.FullPath)
-	tokenName, err := l.writeKubeConfig(obj, tempDirs.KubeConfigFull, true)
-	if err != nil {
-		return obj, err
-	}
-	defer func() {
-		if err := l.systemTokens.DeleteToken(tokenName); err != nil {
-			logrus.Errorf("cleanup for helm token [%s] failed, will not retry: %v", tokenName, err)
+	if !(obj.Annotations["cattle.io/skipUninstall"] == "true") {
+		tempDirs, err := createTempDir(obj)
+		if err != nil {
+			return obj, err
 		}
-	}()
-	// try three times and succeed
-	start := time.Second * 1
-	for i := 0; i < 3; i++ {
-		if err = helmDelete(tempDirs, obj); err == nil {
-			break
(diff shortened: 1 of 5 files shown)


jiaqiluo (dismissed):
LGTM

StrongMonkey (dismissed):

pkg/controllers/managementuserlegacy/helm/controller.go
```diff
@@ -269,29 +269,31 @@ func (l *Lifecycle) Remove(obj *v3.App) (runtime.Object, error) {
...
-	}
-	defer os.RemoveAll(tempDirs.FullPath)
-	tokenName, err := l.writeKubeConfig(obj, tempDirs.KubeConfigFull, true)
-	if err != nil {
-		return obj, err
-	}
-	defer func() {
-		if err := l.systemTokens.DeleteToken(tokenName); err != nil {
-			logrus.Errorf("cleanup for helm token [%s] failed, will not retry: %v", tokenName, err)
+	if !(obj.Annotations["cattle.io/skipUninstall"] == "true") {
```
jiaqiluo, line 272: A couple of  questions here:
1. In which case is this annotation used? In other words,  in which case that the uninstallation should be skipped?  
1. Is this annotation a new or existing one? 
2. Does UI add this annotation to the app?

jiaqiluo (changes requested):

pkg/controllers/managementuserlegacy/helm/controller.go
rmweir: This is used after an app has been fully migrated and a user would like to deleted the old norman app object. The annotation is new. No it is meant to be added manually for now.

jiaqiluo: ok, it sounds like something we need to add to the Rancher document and release notes.

jiaqiluo (approved):
LGTM. 
Updates in the documentation and release notes are needed.

StrongMonkey (approved):

PR #35238, opened 2021-10-21
p0 test for AKS v2 hosted cluster

sgapanovich:
* created a tests/v3_api/test_aks_v2_hosted_cluster.py
  * basic coverage includes 
    * creating azure credentials
    * creating an AKS v2 hosted cluster with default params from UI
    * validating the cluster 
    * deleting the cluster

* also renamed `tests/v3_api/test_hosted_eks_cluster` to `tests/v3_api/test_eks_v2_hosted_cluster.py`
  * now aks v2 and eks v2 hosted clusters have the same pattern 
  * the name was chosen based on the test names `def test_eks/aks_v2_hosted_cluster_create`


sowmyav27 (dismissed):
LGTM

izaac (approved):
I didn't see the renamed EKS hosted cluster file and its tests invoked from on tag jobs yet or from anywhere else. 

I think it's safe to merge unless other reviewers think otherwise.
LGTM

Issue #34886, opened 2021-09-23
[harvester global flag] toggle harvester node driver with the harvester global flag


slickwarren:
**Is your feature request related to a problem? Please describe.**
* Currently harvester node driver is disabled by default. 

**Describe the solution you'd like**
I think we should enable this when harvester global flag is enabled (which is by default).

yasker:
I'd like to enable Harvester node driver by default as well, as long as there is no objection from @cloudnautique .

Also, I don't think we need to do sync between the two flags at this moment.

cloudnautique:
It should be enabled by default. If the user disables the flag, then disable the driver. The admin should still be able to enable the driver manually in the driver management page.

yasker:
Move to v2.6.2 since it's not critical enough for a last-minute squeeze for v.2.6.1.

yasker:
Update from @cloudnautique 

```
I like the idea of enabling both with the harvester flag. It gives an easy all in experience. For toggling it off, we can keep them disconnected.
```

So:
1. If the feature flag was turned on, we enable the Harvester node driver automatically (if it's not enabled).
2. If the feature flag was turned off, nothing will change to the Harvester node driver.
3. Enable/disable the Harvester node driver will not affect the state of the feature flag.

TachunLin:
Verified fixed on Rancher v2.6.3-rc1 with builtin node driver 
Pass the following test scenario. Close this issue and [harvester/harvester#1465](https://github.com/harvester/harvester/issues/1465) accordingly

Test in Rancher v2.6.3-rc1 under airgapped environment


- [x] Harvester feature flag will be enabled by default and turned on harvester node driver accordingly

- [x] If the feature flag was turned off, nothing will change to the Harvester node driver.

- [x] Enable/disable the Harvester node driver will not affect the state of the feature flag.

- [x] If the feature flag was turned on, we enable the Harvester node driver automatically

Issue #1465, opened 2015-07-09
No error when trying to add duplicate services with different request host/path/port to LB services.  


sangeethah:
No error when trying to add duplicate services with different request host/path/port to LB services.  

Steps to reproduce the problem:
Create Balancer service with 2 targets say lb1 with request host-h1 and lb2 with  request host-h2.
Edit Load Balancer and change lb1 to lb2 and save.
After this , LB has only 1 target lb1 with host-h1. 

Expected Behavior:
API should error out  and not accept duplicate services instead of dropping one of the target entries.

sangeethah:
Tested with latest build from master:
When there are multiple service entries with different host/path/port info , they all get saved .

PR #35315, opened 2021-10-28
toggle harvester node driver based on global harvester feature 

shangma:
address https://github.com/rancher/rancher/issues/34886
Harvester issue [#1465](https://github.com/harvester/harvester/issues/1465)

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -36,6 +36,8 @@ ENV TINI_VERSION v0.18.0
 ENV TELEMETRY_VERSION v0.5.16
 ENV DOCKER_MACHINE_LINODE_VERSION v0.1.8
 ENV LINODE_UI_DRIVER_VERSION v0.3.0
+# make sure the version number is consistent with the one at Line 100 of pkg/data/management/machinedriver_data.go
+ENV DOCKER_MACHINE_HARVESTER_VERSION v0.3.2
 ENV CATTLE_KDM_BRANCH ${CATTLE_KDM_BRANCH}
 ENV HELM_VERSION v3.6.0
 ENV KUSTOMIZE_VERSION v3.10.0
@@ -63,6 +65,10 @@ RUN curl -sLf https://github.com/rancher/machine/releases/download/${CATTLE_MACH
     curl -LO https://github.com/linode/docker-machine-driver-linode/releases/download/${DOCKER_MACHINE_LINODE_VERSION}/docker-machine-driver-linode_linux-amd64.zip && \
     unzip docker-machine-driver-linode_linux-amd64.zip -d /opt/drivers/management-state/bin && \
     rm docker-machine-driver-linode_linux-amd64.zip
+
+RUN curl -LO https://releases.rancher.com/harvester-node-driver/${DOCKER_MACHINE_HARVESTER_VERSION}/docker-machine-driver-harvester-amd64.tar.gz && \
+    tar -xf docker-machine-driver-harvester-amd64.tar.gz -C /opt/drivers/management-state/bin && \
+    rm docker-machine-driver-harvester-amd64.tar.gz
 
 ENV TINI_URL_amd64=https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini \
     TINI_URL_arm64=https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-arm64 \
--- a/pkg/controllers/management/feature/feature_handler.go
+++ b/pkg/controllers/management/feature/feature_handler.go
@@ -13,16 +13,18 @@ import (
 )
 
 type handler struct {
-	featuresClient managementv3.FeatureClient
-	tokensLister   managementv3.TokenCache
-	tokenEnqueue   func(string, time.Duration)
+	featuresClient       managementv3.FeatureClient
+	tokensLister         managementv3.TokenCache
+	tokenEnqueue         func(string, time.Duration)
+	nodeDriverController managementv3.NodeDriverController
 }
 
 func Register(ctx context.Context, wContext *wrangler.Context) {
 	h := handler{
-		featuresClient: wContext.Mgmt.Feature(),
-		tokensLister:   wContext.Mgmt.Token().Cache(),
-		tokenEnqueue:   wContext.Mgmt.Token().EnqueueAfter,
+		featuresClient:       wContext.Mgmt.Feature(),
+		tokensLister:         wContext.Mgmt.Token().Cache(),
+		tokenEnqueue:         wContext.Mgmt.Token().EnqueueAfter,
+		nodeDriverController: wContext.Mgmt.NodeDriver(),
 	}
 	wContext.Mgmt.Feature().OnChange(ctx, "feature-handler", h.sync)
 }
@@ -40,7 +42,24 @@ func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
 	if obj.Name == features.TokenHashing.Name() {
 		return obj, h.refreshTokens()
 	}
+
+	if obj.Name == features.Harvester.Name() {
+		return obj, h.toggleHarvesterNodeDriver(obj.Name)
+	}
 	return obj, nil
+}
+
+func (h *handler) toggleHarvesterNodeDriver(harvester string) error {
+	if val := features.GetFeatureByName(harvester).Enabled(); val {
+		m, err := h.nodeDriverController.Cache().Get(harvester)
+		if err != nil {
+			return err
+		}
+		m.Spec.Active = val
+		_, err = h.nodeDriverController.Update(m)
+		return err
+	}
+	return nil
 }
 
 func (h *handler) refreshTokens() error {
--- a/pkg/data/management/machinedriver_data.go
+++ b/pkg/data/management/machinedriver_data.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/features"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/sirupsen/logrus"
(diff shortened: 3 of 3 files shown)


pkg/controllers/management/feature/feature_handler.go
```diff
@@ -40,6 +44,17 @@ func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
...
+
+	if obj.Name == features.Harvester.Name() {
+		lister := h.mgmt.Management.NodeDrivers("").Controller().Lister()
+		cli := h.mgmt.Management.NodeDrivers("")
+		m, _ := lister.Get("", obj.Name)
+		val := feature.GetEffectiveValue(obj)
+		m.Spec.Active = val
+		if _, err := cli.Update(m); err != nil {
+			return obj, err
+		}
```
gitlawr, line 56: Can we wrap these in a standalone function and use `features.GetFeatureByName` instead of `feature.GetEffectiveValue`

shangma: @gitlawr updated. Thx for the input.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -40,9 +43,23 @@ func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
...
+	}
 	return obj, nil
 }
 
+func (h *handler) toggleHarvesterNodeDriver(harvester string) error {
+	lister := h.mgmt.Management.NodeDrivers("").Controller().Lister()
+	cli := h.mgmt.Management.NodeDrivers("")
+	m, _ := lister.Get("", harvester)
+	val := features.GetFeatureByName(harvester).Enabled()
+	m.Spec.Active = val
```
gitlawr, line 58: According to https://github.com/rancher/rancher/issues/34886#issuecomment-931785510, we don't need to deactivate the node driver when the feature flag is turned off.

shangma: @gitlawr updated. Many thanks for the heads-up.

guangbochen (approved):
LGTM

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -40,9 +43,25 @@ func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
...
+
+	if obj.Name == features.Harvester.Name() {
+		return obj, h.toggleHarvesterNodeDriver(obj.Name)
+	}
 	return obj, nil
 }
 
+func (h *handler) toggleHarvesterNodeDriver(harvester string) error {
+	lister := h.mgmt.Management.NodeDrivers("").Controller().Lister()
+	cli := h.mgmt.Management.NodeDrivers("")
```
nickgerace, line 55: Are empty string inputs for the ManagementContext NodeDrivers always going to work? We may also be able to use one NodeDrivers interface and create another one for the lister rather than calling it twice.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -40,9 +43,25 @@ func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
...
+	if val := features.GetFeatureByName(harvester).Enabled(); val {
+		m, _ := lister.Get("", harvester)
```
nickgerace, line 57: We should at least debug log the error here or handle/return it.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -16,13 +17,15 @@ type handler struct {
...
+	mgmt           *config.ManagementContext
 }
 
-func Register(ctx context.Context, wContext *wrangler.Context) {
+func Register(ctx context.Context, wContext *wrangler.Context, management *config.ManagementContext) {
 	h := handler{
 		featuresClient: wContext.Mgmt.Feature(),
 		tokensLister:   wContext.Mgmt.Token().Cache(),
 		tokenEnqueue:   wContext.Mgmt.Token().EnqueueAfter,
+		mgmt:           management,
```
kinarashah, line 28: Common pattern is to have NodeDriverLister/NodeDriverClient here than passing entire management context (look above at tokensLister for example)

pkg/controllers/management/feature/feature_handler.go
nickgerace: This is what I was getting at for the second point: https://github.com/rancher/rancher/pull/35315#discussion_r743008038

pkg/controllers/management/feature/feature_handler.go
shangma: @nickgerace I just updated the code. Many thanks for the input.

pkg/controllers/management/feature/feature_handler.go
shangma: @nickgerace I just updated the code. Many thanks for the input.

pkg/controllers/management/feature/feature_handler.go
shangma: @kinarashah I just updated the code. Many thanks for the input!

nickgerace:
Hi @shangma is this ready for re-review? Totally fine if not! Just want to make sure I didn't miss needing to re-review yet.

shangma:
@nickgerace  yes. it is ready for re-review. Thanks a lot!

nickgerace (approved):

nickgerace:
Will need review from one more Rancher reviewer

package/Dockerfile
```diff
@@ -64,6 +65,10 @@ RUN curl -sLf https://github.com/rancher/machine/releases/download/${CATTLE_MACH
     unzip docker-machine-driver-linode_linux-amd64.zip -d /opt/drivers/management-state/bin && \
     rm docker-machine-driver-linode_linux-amd64.zip
 
+RUN curl -LO https://releases.rancher.com/harvester-node-driver/${DOCKER_MACHINE_HARVESTER_VERSION}/docker-machine-driver-harvester-amd64.tar.gz && \
+    tar -xf docker-machine-driver-harvester-amd64.tar.gz -C /opt/drivers/management-state/bin && \
+    rm docker-machine-driver-harvester-amd64.tar.gz
+
```
thedadams, line 71: Is there are reason to download this here? Since the URL is passed in `machinedriver_data.go`, then it will be downloaded there when necessary.

The answer here is probably "air gap" but I want to make sure I understand what is happening here before approving.

package/Dockerfile
```diff
@@ -36,6 +36,7 @@ ENV TINI_VERSION v0.18.0
 ENV TELEMETRY_VERSION v0.5.16
 ENV DOCKER_MACHINE_LINODE_VERSION v0.1.8
 ENV LINODE_UI_DRIVER_VERSION v0.3.0
+ENV DOCKER_MACHINE_HARVESTER_VERSION v0.3.2
```
thedadams, line 39: If the version is going to be here and in `machinedriver_data.go`, please add a comment here to indicate that the version should be updated there as well.

pkg/data/management/machinedriver_data.go
```diff
@@ -95,7 +96,8 @@ func addMachineDrivers(management *config.ManagementContext) error {
 	if err := addMachineDriver(GoogleDriver, "local://", "", "", nil, false, true, true, management); err != nil {
 		return err
 	}
-	if err := addMachineDriver(HarvesterDriver, "https://releases.rancher.com/harvester-node-driver/v0.3.2/docker-machine-driver-harvester-amd64.tar.gz", "", "f113d826b8bee79abf60f1026b4076befdf735c6675889240c75da6b30f09971", []string{"releases.rancher.com"}, false, false, false, management); err != nil {
+	harvesterEnabled := features.GetFeatureByName(HarvesterDriver).Enabled()
+	if err := addMachineDriver(HarvesterDriver, "https://releases.rancher.com/harvester-node-driver/v0.3.2/docker-machine-driver-harvester-amd64.tar.gz", "", "f113d826b8bee79abf60f1026b4076befdf735c6675889240c75da6b30f09971", []string{"releases.rancher.com"}, harvesterEnabled, harvesterEnabled, false, management); err != nil {
```
thedadams, line 100: If the version is going to be here and in `package/Dockerfile`, please add a comment here to indicate that the version should be updated there as well.

pkg/controllers/management/feature/feature_handler.go
```diff
@@ -40,9 +42,26 @@ func (h *handler) sync(_ string, obj *v3.Feature) (*v3.Feature, error) {
...
 	}
+
+	if obj.Name == features.Harvester.Name() {
+		return obj, h.toggleHarvesterNodeDriver(obj.Name)
+	}
 	return obj, nil
 }
 
+func (h *handler) toggleHarvesterNodeDriver(harvester string) error {
+	if val := features.GetFeatureByName(harvester).Enabled(); val {
```
thedadams, line 53: Should the node driver be disabled when the feature is turned off? In other words, I think it would be better to update the node driver whenever `Active` is different from the feature flag `Enabled` status.

thedadams (changes requested):
There seems to be a gap in the logic here. If the feature flag is enabled, then the node driver gets enabled. However, the following scenarios are not handled:
1. If the feature flag is disabled, then the node driver will be enabled until Rancher restarts.
2. If the node driver is enabled, then the feature flag will not change.
3. If the node driver is disabled, then the feature flag will not change.

Are these gaps intentional? It seems that we should also be handling these as well.

pkg/controllers/management/feature/feature_handler.go
shangma: hey, @thedadams , the node driver can be kept enabled even the global Harvester feature is disabled. Please see the comment [here](https://github.com/rancher/rancher/issues/34886#issuecomment-931785510)

package/Dockerfile
shangma: Hey @thedadams  you are absolutely correct. The reason for downloading it is for air gap environment.

package/Dockerfile
shangma: @thedadams  I have added the comment at both places. The pr is ready for re-review. Thanks a lot for the input!

shangma:
@thedadams  I have added the comment for the version number thing. The pr is ready for re-review. Thanks a lot for the input!

thedadams (approved):

guangbochen:
let us know if anything else is needed for the PR merge, thanks in advance.

Issue #33850, opened 2021-07-30
Use embedded etcd from k3s for single node usage


nickgerace:
This is an exploratory issue to track dropping the etcd binary in favor of embedded etcd from k3s for single node usage.

annablender:
Merged in a PR https://github.com/rancher/rancher/pull/35323. Single node K3s with embedded etcd will be available in 2.6.4. Docs on how to look at the data inside embedded etcd and debug issues with the new database are here How to Use Etcdctl with K3s Embedded Etcd.

cbron:
@annablender can you fill out the QA template here, then @MbolotSuse this might be a good dev-validate for you.

MbolotSuse:
Validated image using manifest sha of `7d3c66bd40f8a0f45add026453ed1e7cec0902e1f31170db7e22977feff7dbed` and config sha of `a50b5a93275dd32e949778b0b20743cdb830e0bfd6d69be481a53a3344543d6a`. This was built in DroneCI from commit `77392a65`, which was merged several days after the target commit `944a4654` (i.e. - testing was completed using `rancher/rancher:v2.6-head` as of ~8:30 am CT, which was also `rancher/rancher:v2.6-77392a65423df8507c27d07d0c0a3e85c4d97a5a-linux-amd64`).

Other images used:
| Image | Image Manifest sha |
|-------|--------------------|
| rancher/rancher:v2.6.2 | e9ea04573c21d9126d76d41c2c1311d52fa6835095410a3487d286776ca424fc |


| Test Case Id | Test Case Overview | Test Case Steps | Test Case Status |
|--------------|--------------------|-----------------|------------------|
| 1 | Ensure ETCD can be interacted with in the container using etcdctl with the [provided instructions](https://github.com/rancher/rancher/wiki/How-to-Use-Etcdctl-with-K3s-Embedded-Etcd) | 1. Run rancher/rancher:v2.6-head. <br/> 2. Exec into the container. <br/> 3. Wait for the k3s cluster to fully come up. <br/> 4. Using the earlier etcdctl instructions, issue a `member list` command and ensure that the output is valid. <br/> 5. Issue a http request `curl -i http://localhost:2379`. Ensure a blank reply. | Pass  |
[...]

PR #35323, opened 2021-10-29
[2.6.3] Rip out etcd binary

annablender:
Issue [33850](https://github.com/rancher/rancher/issues/33850). Exploratory issue that removes the downloaded etcd binary and starts Rancher single node k3s cluster with embedded etcd.

#### Design plan

- Rip etcd binary and env vars out of rancher/rancher Dockerfile.dapper and package/Dockerfile
- Update the Norman API
    - Remove handler function RunETCD that starts etcd
    - Remove —datastore-endpoint
    - Start k3s server with —cluster-init (which will use embedded etcd)

#### What was fixed, or what changes have occurred

- Remove etcd binary and env vars
- Update K3S_URL in Dockerfile to avoid overwriting k3s server url during setup
- Update go.mod to reference changes in Norman API (the k3s server initialization is done in the Norman API and that's where the other half of these changes are)
- Fix bug in the validate script
- Add missing /var/lib/rancher directory

#### Areas or cases that should be tested

- Spin up a custom Rancher image v2.6 and test if 1) the local cluster is stable, 2) a deployment can be created, 3) a downstream cluster can be created. I used a custom image modified for local dev that can be pulled from docker hub _annablender/rancher:k3s-embedded-etcd_.
- Drop into the Rancher container and ping the embedded etcd endpoint "curl -i localhost:2379." If running, you'll establish a connection and receive empty data
- K3s single node upgrade
- Persisted data upgrade
- Backup and restore
- Use etcdctl with flag --endpoints=https://<node ip>:2379 to test embedded etcd

### DIFF
--- a/package/entrypoint.sh
+++ b/package/entrypoint.sh
@@ -6,6 +6,18 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -e /var/lib/rancher/etcd ] && [ ! -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  mkdir -p /var/lib/rancher/k3s/server/db
+  ln -sf /var/lib/rancher/etcd /var/lib/rancher/k3s/server/db/etcd
+  echo -n 'default' > /var/lib/rancher/k3s/server/db/etcd/name
+fi
+if [ -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  k3s server --cluster-init --cluster-reset &> ./k3s-cluster-reset.log
+  if [ $? -ne 0 ]; then
+    echo "ERROR:" && cat ./k3s-cluster-reset.log
+    rm -f /var/lib/rancher/k3s/server/db/reset-flag
+  fi
+fi
 if [ -x "$(command -v update-ca-certificates)" ]; then
   update-ca-certificates
 fi
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -5,7 +5,6 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
 ENV CATTLE_HELM_VERSION v2.16.8-rancher1
 ENV CATTLE_MACHINE_VERSION v0.15.0-rancher73
 ENV CATTLE_K3S_VERSION v1.22.3+k3s1
-ENV CATTLE_ETCD_VERSION v3.5.0
 # version used by helm plugin install script
 ENV CATTLE_HELM_UNITTEST_VERSION v0.1.7-rancher3
 # helm 3 version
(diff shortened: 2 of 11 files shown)


brandond (commented):
LGTM once the norman PR is merged and go.mod updated.

nickgerace:
Will re-review once https://github.com/rancher/norman/pull/406 is merged and the module is updated for this PR

package/entrypoint.sh
```diff
@@ -6,6 +6,12 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -e /var/lib/rancher/etcd ] && [ ! -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  mkdir -p /var/lib/rancher/k3s/server/db
+  ln -sf /var/lib/rancher/etcd /var/lib/rancher/k3s/server/db/etcd
+  echo -n 'default' > /var/lib/rancher/k3s/server/db/etcd/name
+  k3s server --cluster-reset
+fi
```
brandond, line 14:
```suggestion
fi
if [ -e /var/lib/rancher/k3s/server/db/etcd ]; then
  k3s server --cluster-reset
fi
```

gomodtidy.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
brandond, line 1: not sure about this one

brandond (changes requested):

package/entrypoint.sh
annablender: Ahh good point

gomodtidy.sh
annablender: Oh that's a script I wrote for updating go modules! My bad, I'll remove that

package/entrypoint.sh
```diff
@@ -6,6 +6,14 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -e /var/lib/rancher/etcd ] && [ ! -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  mkdir -p /var/lib/rancher/k3s/server/db
+  ln -sf /var/lib/rancher/etcd /var/lib/rancher/k3s/server/db/etcd
+  echo -n 'default' > /var/lib/rancher/k3s/server/db/etcd/name
+fi
+if [ -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  k3s server --cluster-reset &> ./k3s-cluster-reset.log
```
brandond, line 15: Note for other reviewers: this is necessary because K3s's embedded etcd can't be bound to 127.0.0.1, and will get mad if the Docker container IP address changes. This `--cluster-reset` doesn't drop any data from the etcd datastore; it just resets the cluster membership list and ensures that the sole etcd cluster member has the correct IP address.

brandond (approved):

nickgerace (dismissed):

ibuildthecloud:
R.I.P.

nickgerace (dismissed):

brandond:
lets get the conflicts resolved and :ship::it:!

Dockerfile.dapper542220813
```diff
@@ -0,0 +1,94 @@
+FROM registry.suse.com/suse/sle15:15.3
```
superseb, line 1: Can we remove file `Dockerfile.dapper542220813`?

superseb:
@annablender is `annablender/rancher:k3s-embedded-etcd` still the image to test with?

package/entrypoint.sh
```diff
@@ -6,6 +6,14 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
     exit 1
 fi
 rm -f /var/lib/rancher/k3s/server/cred/node-passwd
+if [ -e /var/lib/rancher/etcd ] && [ ! -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  mkdir -p /var/lib/rancher/k3s/server/db
+  ln -sf /var/lib/rancher/etcd /var/lib/rancher/k3s/server/db/etcd
+  echo -n 'default' > /var/lib/rancher/k3s/server/db/etcd/name
+fi
+if [ -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  k3s server --cluster-init --cluster-reset &> ./k3s-cluster-reset.log
```
superseb, line 15: What is the reason to suppress the log here? If this fails, the container log is empty and the user does not know what's going on.

And do we need any counter to when this fails to recover automatically or if this fails, it needs manual intervention and we need to point that out? Or is that covered when we show the log from `--cluster-reset`?

brandond: The log from this operation is quite noisy, and will not normally be something that the user needs to see. We could probably check to see if it succeeds, and cat the output if it fails?

annablender: @superseb Is this ok with you?

superseb: Yeah we dont need more noisy stuff in there but we can't not let the user know whats going on in this case. Exit code check and showing the log is fine.

Does a `cluster-reset` failing for some reason always require manual intervention? So we don't need some sort of auto recovery in that case and just always manual intervention where the user can use the printed log? I'm asking this in relation to how it was, and that was k3s always trying to start (it didn't do a cluster reset afaik)

brandond: The cluster-reset isn't really for k3s, it's for etcd. Since we can't tell k3s to use the loopback address for the cluster member IP, we need to reset the cluster member address to the current container address every time, or etcd will refuse to start.

https://github.com/rancher/rancher/pull/35323#discussion_r747068862

nickgerace (approved):

package/entrypoint.sh
```diff
@@ -6,6 +6,18 @@ if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; the
...
+  if [$? -ne 0]; then
```
brandond, line 16: I'm pretty sure you'll need whitespace around this
```suggestion
  if [ $? -ne 0 ]; then
```

brandond (changes requested):

superseb (approved):

nickgerace (approved):

nickgerace:
If/once @brandond approves, I'll merge.

brandond (approved):

Issue #26295, opened 2020-03-26
Service type makes impossible to deploy with Helm on GKE


vitorfhc:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

1. Follow all steps presented in the [Install Rancher on the Kubernetes Cluster](https://rancher.com/docs/rancher/v2.x/en/installation/k8s-install/helm-rancher/) using LetsEncrypt
2. Rancher will be deployed but the Ingress resource will show an error on GKE UI
> error while evaluating the ingress spec: service "cattle-system/rancher" is type "ClusterIP", expected "NodePort" or "LoadBalancer"

**Result:**
Rancher will not be deployed properly because Ingress is never created

**Other details that may be helpful:**
As far as I know, GKE sets ClusterIP as the default type for a service
#13035 shows the same problem and the solution presented doesn't make sense when you want to create a LetsEncrypt certificate using the cert-manager.

**Screenshot:**
![image](https://user-images.githubusercontent.com/19597793/77697559-7277aa80-6f85-11ea-8796-4ff0db304571.png)

**Possible solution:**
Setting a variable in values.yml which is reflected as the chosen type to the service.

mrickert-bouncex:
Just ran into this issue as well, same error. An ugly workaround is to edit the service in gke services and change it from clusterip to loadbalancer.

vincent99:
We support running the Rancher control plane only on a RKE or K3s cluster, because your ability to backup, restore, scale and configure the underlying etcd is limited or non-existent in hosted providers.

Yes you can make it work in places we don't support, but adding extra options to make an unsupported configuration convenient is not really going to be a priority for us.

daavelar:
Same issue here

amoon78:
Is it fixed on Rancher 2.5 version? Regarding this doc https://rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-k8s/ , it says `As of Rancher v2.5, Rancher can be installed on any Kubernetes cluster`. 
I tired to install rancher 2.5.5 on GKE but, have exactly same issue. 
Is there any available option or variables to have node port or loadbalancer type instead of clusterip?

soudaburger:
Use case: GKE + IAP to protect rancher resources:

Kubernetes cluster is deployed, but service is not configurable to be of type `NodePort` or `LoadBalancer` in order for ingress to be visible as a Google Cloud Global Load Balancer.

[...]

PR #35333, opened 2021-11-01
[2.6] [Helm Chart] Add Service annotation support

charlesarnaudo:
This change adds the ability to specify service annotations for the Rancher service object, allowing users to deploy Rancher in cloud environments as discussed in #26295 and discussed in https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/

### DIFF
--- a/chart/templates/service.yaml
+++ b/chart/templates/service.yaml
@@ -4,6 +4,9 @@ metadata:
   name: {{ template "rancher.fullname" . }}
   labels:
 {{ include "rancher.labels" . | indent 4 }}
+{{- if .Values.service.annotations }}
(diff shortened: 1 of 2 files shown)


edqallen:
Is this going anywhere?   I found my way here from https://github.com/rancher/rancher/issues/26295 as I'm really trying to *avoid* using RKE or K3S, as I'd like to deploy Rancher to a norma K8S cluster that we already have tooling in place for.  The documentation on how to do this is woefully inadequate.

Honestly even looking at the diffs in this push request, it's still not clear to me how I can e.g. override the service type when doing a helm install.  Is it `--set service.type=NodePort`?  `--set serviceType=NodePort`?  Something else?

Issue #35312, opened 2021-10-28
Unable to import AKS clusters from different subscriptions with same name


polivbr:
**Rancher Server Setup**
- Rancher version: 2.6.2
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): AKS
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: 1.20.9
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Hosted - AKS

**Describe the bug**
Cannot import multiple AKS clusters from different subscriptions that have the same name in Azure, even when giving them different names in Rancher.

**To Reproduce**
Import an AKS cluster from one subscription. Attempt to import an AKS cluster from a different subscription that has the same name in Azure as the first cluster.

**Result**
`Bad response statusCode [422]. Status [422 Unprocessable Entity]. Body: [baseType=error, code=InvalidBodyContent, message=cluster already exists for AKS cluster [<cluster name>] in region [centralus]]`

**Expected Result**
The second cluster should import successfully.

polivbr:
hello? can someone from Rancher look at this?

polivbr:
ping

polivbr:
HERE I AM, JUST SHOUTING INTO THE WIND.

I've submitted the ticket and a pull request that address the issue. Not sure what else I need to do to get someone to actually look at this.

mteodori:
same issue for EKS clusters, cannot add multiple clusters with the same name in different AWS accounts

`Error: Bad response statusCode [422]. Status [422 Unprocessable Entity]. Body: [message=cluster already exists for EKS cluster [...] in region [...], baseType=error, code=InvalidBodyContent] from [https://rancher.us.staging.tools.hyland.dev/v3/clusters]`

Erodotos:
Hi we are facing the same issue on AKS. Is anyone facing the same limitation?

PR #35335, opened 2021-11-01
Allow AKS clusters with same name if in different subscriptions

polivbr:
Addresses https://github.com/rancher/rancher/issues/35312


polivbr:
Also requires: https://github.com/rancher/aks-operator/pull/48

Issue #35310, opened 2021-10-28
Fail to upgrade a local RKE2 cluster if node drain is enabled


bk201:
**Rancher Server Setup**
- Rancher version: `2.6.2`
- Installation option (Docker install/Helm Chart): rancherd 
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version:`v1.21.5+rke2r2`
- Cluster Type (Local/Downstream): a local cluster provisioned by rancherd
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**Describe the bug**

I enable node drain options and upgrade a local RKE2 cluster, only the first node has RKE2 upgraded. 

**To Reproduce**
 

Note: this is spotted when developing the upgrade feature in Harvester. I try to create a simple environment to reproduce this issue.

- Use rancherd to provision a RKE2 cluster (1 server + 1 worker).

    On node1 (server):
    ```
    mkdir -p /etc/rancher/rke2/config.yaml.d/
    
    cat > /etc/rancher/rke2/config.yaml.d/90-harvester-server.yaml <<EOF
    disable: rke2-ingress-nginx
    EOF
    
    mkdir -p /etc/rancher/rancherd
    cat > /etc/rancher/rancherd/config.yaml << EOF
    role: cluster-init
    token: somethingrandom
    kubernetesVersion: stable:rke2
    rancherVersion: stable
    rancherValues:
      features: multi-cluster-management=true
    EOF
    curl -fL https://raw.githubusercontent.com/rancher/rancherd/master/install.sh | sh -
    ```
    
    On node2 (worker):
    ```
    mkdir -p /etc/rancher/rancherd
    cat > /etc/rancher/rancherd/config.yaml << EOF
    role: agent
    token: somethingrandom
    kubernetesVersion: stable:rke2
    rancherVersion: stable
    server: https://192.168.121.38:8443
    EOF
    curl -fL https://raw.githubusercontent.com/rancher/rancherd/master/install.sh | sh -
    ```

- After everything is OK, upgrade the local RKE2 cluster by editing the resource:
    ```
    kubectl edit clusters.provisioning.cattle.io local -n fleet-local
    ```

    change `spec` from:
    ```
    spec:
      kubernetesVersion: v1.21.5+rke2r2
      rkeConfig: {}
    ```

    to:

    ```
[...]

PR #35358, opened 2021-11-02
Add better support for upgrading local cluster

thedadams:
When the local cluster is provisioned via rancherd, a hostPort is used.
When Rancher is contacted via the hostPort, the internal-cacerts setting
is used instead of the regular cacerts setting. This was causing issues
with system-agent trying to verify Rancher's CA. After this change, The
install script logic with provide the corresponding CA as necessary.

In addition, when certificates were being regenerated on Kubernetes
upgrade, the clusterIP was being used but not the hostIP. This again was
causing issues with system-agent verifying Rancher's CA because the SAN
did not include the hostIP when hostPort was enabled.

Finally, once machines were "uncordoned" the annotation was never
getting removed. This change will remove the annotation after the node
is uncordoned.

Issue:
https://github.com/rancher/rancher/issues/35310

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
@@ -9,9 +9,12 @@ import (
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1alpha4"
 	rkecontroller "github.com/rancher/rancher/pkg/generated/controllers/rke.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/provisioningv2/rke2/installer"
 	"github.com/rancher/rancher/pkg/provisioningv2/rke2/planner"
+	"github.com/rancher/rancher/pkg/tls"
 	"github.com/rancher/rancher/pkg/wrangler"
+	appcontrollers "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/name"
@@ -43,6 +46,7 @@ type handler struct {
 	secretCache         corecontrollers.SecretCache
 	machineCache        capicontrollers.MachineCache
 	capiClusters        capicontrollers.ClusterCache
+	deploymentCache     appcontrollers.DeploymentCache
 	rkeControlPlanes    rkecontroller.RKEControlPlaneCache
 }
 
@@ -52,6 +56,7 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 		secretCache:         clients.Core.Secret().Cache(),
 		machineCache:        clients.CAPI.Machine().Cache(),
 		capiClusters:        clients.CAPI.Cluster().Cache(),
+		deploymentCache:     clients.Apps.Deployment().Cache(),
 		rkeControlPlanes:    clients.RKE.RKEControlPlane().Cache(),
 	}
 	rkecontroller.RegisterRKEBootstrapGeneratingHandler(ctx,
@@ -109,7 +114,13 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 		}
 
 		hash := sha256.Sum256(secret.Data["token"])
-		data, err := installer.LinuxInstallScript(base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
+
+		hasHostPort, err := h.rancherDeploymentHasHostPort()
+		if err != nil {
+			return nil, err
+		}
+
+		data, err := installer.LinuxInstallScript(context.WithValue(context.Background(), tls.InternalAPI, hasHostPort), base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
 		if err != nil {
 			return nil, err
 		}
@@ -298,4 +309,24 @@ func (h *handler) OnChange(obj *rkev1.RKEBootstrap, status rkev1.RKEBootstrapSta
 
 	result = append(result, objs...)
 	return result, status, nil
+}
+
+func (h *handler) rancherDeploymentHasHostPort() (bool, error) {
+	deployment, err := h.deploymentCache.Get(namespace.System, "rancher")
+	if err != nil {
+		if apierror.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	for _, container := range deployment.Spec.Template.Spec.Containers {
+		for _, port := range container.Ports {
+			if container.Name == "rancher" && port.HostPort != 0 {
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
 }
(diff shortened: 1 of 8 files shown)


pkg/provisioningv2/rke2/configserver/server.go
```diff
@@ -116,11 +117,16 @@ func (r *RKE2ConfigServer) connectAgent(planSecret string, secret *v1.Secret, rw
 	}
 
 	if url == "" {
-		_, pem = settings.InternalServerURL.Get(), settings.InternalCACerts.Get()
+		pem = settings.InternalCACerts.Get()
 		url = fmt.Sprintf("https://%s", req.Host)
 		if strings.TrimSpace(pem) != "" {
 			ca = []byte(pem)
 		}
+	} else if v, ok := req.Context().Value(tls.InternalAPI).(bool); ok && v {
```
thedadams, line 125: For reviewers: this change is copied from here: https://github.com/rancher/rancher/blob/60275b1f66d2e2c0ae40a9edaa7ca874ac2c23b6/pkg/cacerts/handler.go#L36

pkg/provisioningv2/rke2/installer/installer.go
```diff
@@ -69,6 +71,9 @@ func LinuxInstallScript(token string, envVars []corev1.EnvVar, defaultHost strin
 		}
 	}
 	ca := systemtemplate.CAChecksum()
+	if v, ok := ctx.Value(tls.InternalAPI).(bool); ok && v {
```
thedadams, line 74: For reviewers: this is copied from here: https://github.com/rancher/rancher/blob/60275b1f66d2e2c0ae40a9edaa7ca874ac2c23b6/pkg/cacerts/handler.go#L36

pkg/provisioningv2/rke2/installer/installer.go
```diff
@@ -106,6 +111,9 @@ func WindowsInstallScript(token string, envVars []corev1.EnvVar, defaultHost str
 	}
 
 	ca := systemtemplate.CAChecksum()
+	if v, ok := ctx.Value(tls.InternalAPI).(bool); ok && v {
```
thedadams, line 114: For reviewers: this is copied from here: https://github.com/rancher/rancher/blob/60275b1f66d2e2c0ae40a9edaa7ca874ac2c23b6/pkg/cacerts/handler.go#L36

paynejacob (approved):

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -84,7 +84,12 @@ func (h *handler) undrain(machine *capi.Machine) (*capi.Machine, error) {
 
 	c := drain.NewCordonHelper(node)
 
-	if updateRequired := c.UpdateIfRequired(false); !updateRequired {
+	if !c.UpdateIfRequired(false) {
+		if machine.Annotations[planner.UnCordonAnnotation] != "" {
+			machine = machine.DeepCopy()
+			delete(machine.Annotations, planner.UnCordonAnnotation)
```
Oats87, line 90: nice

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -1066,7 +1066,7 @@ func (p *Planner) addConfigFile(nodePlan plan.NodePlan, controlPlane *rkev1.RKEC
 
 	nodePlan.Files = append(nodePlan.Files, plan.File{
 		Content: base64.StdEncoding.EncodeToString(configData),
-		Path:    fmt.Sprintf(ConfigYamlFileName, rancherruntime.GetRuntime(controlPlane.Spec.KubernetesVersion)),
+		Path:    fmt.Sprintf(ConfigYamlFileName, runtime),
```
Oats87, line 1069: nice

pkg/tls/tls.go
```diff
@@ -309,6 +324,54 @@ func getClusterIP(services corev1controllers.ServiceController) (string, error)
 	return service.Spec.ClusterIP, nil
 }
 
+func getHostIPs(deployments appscontrollers.DeploymentController, nodes corev1controllers.NodeController) ([]string, error) {
+	deployment, err := deployments.Get(namespace.System, "rancher", metav1.GetOptions{})
```
Oats87, line 328: Any way we can consolidate the use of `rancher` here into a constant? We're also using it above on line 314.

pkg/tls/tls.go
```diff
@@ -309,6 +324,54 @@ func getClusterIP(services corev1controllers.ServiceController) (string, error)
...
+	nodeIPs := make([]string, 0, len(nodes.Items))
+	for _, node := range nodes.Items {
+		var nodeIP string
+		for _, ip := range node.Status.Addresses {
+			if ip.Type == v1.NodeInternalIP {
+				nodeIP = ip.Address
+				break
+			} else if ip.Type == v1.NodeExternalIP && nodeIP == "" {
+				nodeIP = ip.Address
+			}
```
Oats87, line 364: Is the idea here to prefer internal? My interpretation of this is that once the internal IP is encountered, that is set/appended, and then we break. I'm not sure that's the correct behavior, as a NodeExternal IP is still valid to reach a node for a hostport service.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -109,7 +109,7 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 		}
 
 		hash := sha256.Sum256(secret.Data["token"])
-		data, err := installer.LinuxInstallScript(base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
+		data, err := installer.LinuxInstallScript(context.Background(), base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
```
Oats87, line 112: Does this not mean that when we are provisioning machines via machine, these machines will not get the proper CA checksum populated into the install script?

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
Oats87, line 90: nice

pkg/provisioningv2/rke2/planner/planner.go
Oats87, line 1069: nice

pkg/tls/tls.go
Oats87, line 328: Is there any way we can use a constant for `rancher` as we appear to be using it above as well.

pkg/tls/tls.go
```diff
@@ -309,6 +324,54 @@ func getClusterIP(services corev1controllers.ServiceController) (string, error)
...
+func collectNodeIPs(nodeController corev1controllers.NodeController) ([]string, error) {
+	nodes, err := nodeController.List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	nodeIPs := make([]string, 0, len(nodes.Items))
+	for _, node := range nodes.Items {
+		var nodeIP string
+		for _, ip := range node.Status.Addresses {
```
Oats87, line 358: Is the idea here that we are trying to prefer using the internal node IP? It looks like we take preference to only capturing the internal node IP but I don't think that is fully valid, as the external IP can be used to hit the node as well.

Oats87 (changes requested):
A few clarifications/questions but overall this looks sane/good

pkg/tls/tls.go
thedadams: Yes, you are correct. I will include the external IP as well.

pkg/tls/tls.go
thedadams: Done.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
thedadams: Only if hostPort is set. This is currently the case as well (in other words, this is not a regression). I will see if I can fix this here too.

Oats87 (approved):

PR #35322, opened 2021-10-29
[Backport v2.5] Fix namespaces not being moved to System project on upgrade

jakefhyde:
Backport of #34885, which is a forward port of #34527, addressing some issues that were missed in the original PR (namely, the `retry.RetryOnConflict` logic would never succeed if it failed the first time).

Related Issue: https://github.com/rancher/rancher/issues/32898

Issue #32898, opened 2021-05-26
rancher operator is created rancher-operator-system which is not installed in System project 


sowmyav27:
**What kind of request is this (question/bug/enhancement/feature request):** bug

**Steps to reproduce (least amount of steps as possible):**

- Deploy rancher HA 2.5.8
- The rancher operator chart/pod is seen deployed in rancher-operator-system which is not installed in System project of the local cluster

**Expected Result:**
Rancher operator system namespace should be in thhe System project of the local cluster

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.8
- Installation option (single install/HA): HA

jakefhyde:
Discussion
-
Per discussion, we've defined the following requirements:

- [ ] Fix the `DefaultNamespaceAssigned` and `SystemNamespacesAssigned` conditions not being properly set for system and default project namespace assignment
- [ ] Add the `ensureNamespacesAssigned` check to project updated (pending successful condition set update)
- [ ] Unset the `SystemNamespacesAssigned` conditions on controller registration, which will happen on startup and election ensuring successful update

Notes
-
- Projects are updated on user interaction, as well as cluster sync
- Decided to only move namespaces with no `projectID` annotations (i.e. not currently associated with a project) that should be in the System project (i.e. defined in the `system-namespaces` setting) into the System project
- Only unset the `SystemNamespacesAssigned` on registration, as the Default project only requires the `default` namespace, and is unlikely to change
- The fix to add fleet auto-generated namespaces into the system-namespaces project is 2.6 onward, as is the hook to check if specific updates are required during an upgrade based on the current rancher version

cbron:
Moving this to 2.5.11, don't merge into 2.5 until 2.5.10 goes out.

jmcsagdc:
Verified in v2.5-e014cd0b24c798950cdbcd7d99d4fd83ea529d36-head

1. launch something older. I used 2.5.10
2. add a namespace outside of any project. I created "david"
3. use the api to add Harry as a System project
4. upgrade Rancher to 2.5-head
5. log in and see where david ended up in the Projects/Namespaces

EXPECTED:
david should be in the System project under the local cluster

![postupgrade2 5](https://user-images.githubusercontent.com/15971368/143966347-d1f9fb2a-5be5-40d0-82b4-d61f426febad.jpg)

PR #35359, opened 2021-11-03
[Forwardport v2.6] Fix namespaces not being moved to System project on upgrade (part 2)

jakefhyde:
Forwardport of #35322, which is a backport of #34885, which is a forward port of #34527, addressing some issues that were missed in the original PR. A specific commit was lost as a result of a force push, which actually performed the system project assignment logic.

Related Issue: #32898

### DIFF
--- a/pkg/rancher/migrations.go
+++ b/pkg/rancher/migrations.go
@@ -197,10 +197,8 @@ func forceSystemNamespaceAssignment(configMapController controllerv1.ConfigMapCo
 	}
 
 	for i := range clusterList.Items {
-		c := &clusterList.Items[i]
-
 		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			c, err = clusterClient.Get(c.Name, metav1.GetOptions{})
+			c, err := clusterClient.Get(clusterList.Items[i].Name, metav1.GetOptions{})
 			if err != nil {
 				return err
 			}
--- a/pkg/controllers/managementuser/rbac/project_handler.go
+++ b/pkg/controllers/managementuser/rbac/project_handler.go
@@ -43,7 +43,8 @@ func (p *pLifecycle) Create(project *v3.Project) (runtime.Object, error) {
 }
 
 func (p *pLifecycle) Updated(project *v3.Project) (runtime.Object, error) {
-	return nil, nil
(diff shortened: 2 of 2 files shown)


thedadams (approved):

paynejacob (dismissed):

pkg/rancher/migrations.go
```diff
@@ -197,10 +197,10 @@ func forceSystemNamespaceAssignment(configMapController controllerv1.ConfigMapCo
 	}
 
 	for i := range clusterList.Items {
-		c := &clusterList.Items[i]
+		cluster := &clusterList.Items[i]
 
 		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			c, err = clusterClient.Get(c.Name, metav1.GetOptions{})
+			c, err := clusterClient.Get(cluster.Name, metav1.GetOptions{})
```
jakefhyde, line 203: @thedadams I think this may be a potential fix to the panic, because as it was written previously it would overwrite `c`.

paynejacob (approved):

pkg/rancher/migrations.go
thedadams: The implementation of `Get` never returns a `nil` cluster, even if the cluster is not found.

I think removing the colon would be better here so we don't have two copies of the same cluster object in memory.

thedadams (changes requested):

pkg/rancher/migrations.go
jakefhyde: @thedadams While the returned cluster object is not nil, it is zero initialized on failure which will have the same effect where our retry logic never succeeds. Also I'm omitting the outer scoped `cluster` object entirely, since only the name is required.

thedadams: If I understand correctly, the cluster object returned by `Get` is only zero-valued when an error is returned. Since you are handling the error here, the cluster object is never attempted to be updated. Also, whether you are declaring a new variable or overwriting an existing one, the effect here would be the same.

With that said, I think removing the outer-scoped cluster object has the same effect I was looking for in my previous comment so that is good with me.

thedadams (approved):

paynejacob (approved):

PR #35365, opened 2021-11-03
[release/v2.5] Bump RKE to v1.2.14

PennyScissors:
Bumps RKE version to add new types for NGINX webhook image.


StrongMonkey (approved):
LGTM. The first commit is to remove the additional line that is generated by golang 1.17 since release/2.5 is not fully bumped to 1.17. 

>The go command now understands //go:build lines and prefers them over // +build lines. The new syntax uses boolean expressions, just like Go, and should be less error-prone. As of this release, the new syntax is fully supported, and all Go files should be updated to have both forms with the same meaning. To aid in migration, gofmt now automatically synchronizes the two forms. For more details on the syntax and migration plan, see https://golang.org/design/draft-gobuild.

Issue #35180, opened 2021-10-19
RKE1 and RKE2 Clusters are stuck in Removing state


sowmyav27:
**Rancher Server Setup**
- Rancher version: v2.6.1
- Installation option (Docker install/Helm Chart): docker install
- Proxy/Cert Details: rancher generated

**Information about the Cluster**
- Kubernetes version: `v1.21.5+rke2r2` and `1.21.5`
- Cluster Type (Local/Downstream): RKE2 DO

**Describe the bug**
RKE1 and RKE2 Clusters are stuck in Removing state

**To Reproduce**
- Deployed 2 rke2 DO cluster
- **Disable RKE2 feature in rancher**
- Deploy an RKE1 cluster
- From Cluster Management, delete the clusters
- All the clusters are stuck in removing state 
- RKE2 nodes are not deleted. Logged separate bug - https://github.com/rancher/rancher/issues/35179
- Rancher logs show these logs repeatedly:
```
2021/10/19 15:46:45 [ERROR] error syncing 'c-m-8wdhwsft': handler mgmt-cluster-remove: Index with name by-cluster does not exist, requeuing
2021/10/19 15:46:54 [ERROR] error syncing 'c-kjlxd': handler mgmt-cluster-remove: Index with name by-cluster does not exist, requeuing
```

**Scenario 2**

- Deploy an RKE1 downstream DO cluster
- Disable RKE2 feature flag
- Deploy another RKE1 downstream cluster DO
- Select both the clusters and hit Delete
- The clusters are stuck in removing state
- Rancher logs:
```
2021/10/20 18:00:27 [ERROR] error syncing 'c-6xzkd': handler cluster-watch: namespaces "c-6xzkd" not found, handler mgmt-cluster-remove: Index with name by-cluster does not exist, requeuing
2021/10/20 18:01:39 [ERROR] Error during subscribe websocket: close sent
2021/10/20 18:01:39 [ERROR] Error during subscribe websocket: close sent
2021/10/20 18:01:40 [ERROR] error syncing 'c-2mkz4': handler mgmt-cluster-remove: Index with name by-cluster does not exist, requeuing
2021/10/20 18:01:56 [ERROR] error syncing 'c-6xzkd': handler mgmt-cluster-remove: Index with name by-cluster does not exist, requeuing
2021/10/20 18:03:40 [ERROR] error syncing 'c-2mkz4': handler mgmt-cluster-remove: Index with name by-cluster does not exist, requeuing
[...]

Issue #35179, opened 2021-10-19
Nodes are not deleted from existing RKE2 cluster when clusters are deleted when RKE2 feature flag is disabled


sowmyav27:
**Rancher Server Setup**
- Rancher version: v2.6.1
- Installation option (Docker install/Helm Chart): docker install
- Proxy/Cert Details: rancher generated

**Information about the Cluster**
- Kubernetes version: `v1.21.5+rke2r2` 
- Cluster Type (Local/Downstream): RKE2 DO

**Describe the bug**
RKE2 nodes are not deleted from DO backend when clusters are deleted

**To Reproduce**
- Deployed 2 rke2 DO cluster
- **Disable RKE2 feature in rancher**
- Deploy an RKE1 cluster
- From Cluster Management, delete the clusters
- All the clusters are stuck in removing state and the RKE2 nodes are not deleted from the backend. 
- RKE1 cluster nodes seem to have been deleted from DO backend

ronhorton:
**PASS** Verified in `2.6-head Commit ID 638b29a`

`The verification of this ticket has changed to ensuring that the user cannot disable the RKE2 flag once there are RKE2 clusters deployed`

1. Created 2 RKE2 DO clusters
2. Went to hamburger > global settings > feature flags
3. Disabled rke2 feature flag
4. rke2 flag remains active.

ronhorton:
**Cannot deleted rke2 cluster**
1. created an rke2 cluster
2. attempted to deactivate rke2 feature flag - remains active
3. delete rke1 and rke2 clusters from cluster management page

result: `rke2 is not deleted`
<img width="989" alt="Screen Shot 2021-11-11 at 1 18 17 PM" src="https://user-images.githubusercontent.com/89212856/141351583-23d58c07-f14c-4190-ae82-995984f6a228.png">

ronhorton:
**PASS** Verified in 2.6-head Commit ID 78b25c3

1. Created downstream RKE2 EC2 cluster
2. Created downstream RKE1 EC2 cluster
3. Created downstream RKE1 DO cluster
4. Created downstream RKE2 DO cluster
5. attempted to deactivate RKE2 Feature Flag Cannont deactivate RKE2 feature flag with existing RKE2 cluster(s)
6. Successfully deleted RKE2 clusters
7. Deactivated RKE2 feature flag flag was deactivated as there are no RKE2 clusters
8. Successfully deleted RKE1 clusters

[...]

PR #35372, opened 2021-11-03
Fix disabling RKE2 feature flag

thedadams:
Firstly, the provisioning cluster indexer was mistakenly put behind the RKE2
feature flag. This caused an issue when deleting RKE1/imported clusters
when RKE2 feature flag was turned off.  This change puts it behind the
ProvisioningV2 feature flag.

In addition, the change adds logic to keep track of when the RKE2
feature flag can be disabled. That is, if there are RKE2 provisioned
clusters then a user should not be able to disable the RKE2 feature
flag. An annotation is added to the feature for easy tracking, and the
annotation is checked on startup and added if necessary so that RKE2
provisioning works as expected.

Issues:
https://github.com/rancher/rancher/issues/35179
https://github.com/rancher/rancher/issues/35180

### DIFF
--- a/pkg/controllers/management/clusterprovisioner/provisioner.go
+++ b/pkg/controllers/management/clusterprovisioner/provisioner.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rancher/norman/types/values"
 	apimgmtv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	util "github.com/rancher/rancher/pkg/cluster"
+	"github.com/rancher/rancher/pkg/controllers/management/imported"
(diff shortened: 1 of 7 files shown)


thedadams:
A note here: the webhook version should be bumped once [this](https://github.com/rancher/webhook/pull/59) PR is merged and a new RC is tagged.

paynejacob (approved):

paynejacob (approved):

thedadams:
Based on [this](https://github.com/rancher/webhook/pull/59#pullrequestreview-798067737) comment from Colleen, the implementation was updated. An update to the webhook is no longer needed.

paynejacob (approved):

kinarashah (approved):

PR #35375, opened 2021-11-04
[Validation/Integration] A redesigned test suite to be used for both integration and automation

igomez06:
A redesigned test suite to be used for both integration and automation testing.

Updated go.mod and go.sum files to account for new defaults package needed."

Added new cloud credentials package with client, provider types, and cloud credential operations.

Implemented generator to be able to generate our own test framework clients to use our new clientbase to override api functionality.

Implemented new TestSession struct that is used to track changes made during a test.


igomez06:
Refactored into a new branch

PR #35392, opened 2021-11-05
Add ui-brand as an unauthenticated setting

vincent99:
So the login page can get it


cjellick (approved):
im actually approving this as a test

PR #35401, opened 2021-11-08
Performance fixes

ibuildthecloud:
Refer to the comments of the individual commits.

### DIFF
--- a/pkg/auth/providerrefresh/refresher.go
+++ b/pkg/auth/providerrefresh/refresher.go
@@ -206,6 +206,10 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 			secret := ""
 			if providers.ProvidersWithSecrets[providerName] {
 				secret, err = r.tokenMGR.GetSecret(user.Name, providerName, loginTokens[providerName])
+				if apierrors.IsNotFound(err) {
+					// There is no secret so we can't refresh, just continue to the next attribute
+					return attribs, nil
+				}
 				if err != nil {
 					return nil, err
 				}
--- a/pkg/controllers/managementuser/alert/deployer/upgradeimpl.go
+++ b/pkg/controllers/managementuser/alert/deployer/upgradeimpl.go
@@ -397,12 +397,13 @@ func (l *AlertService) removeFinalizerFromLegacyAlerting() error {
 	}
 
 	for _, v := range oldProjectAlert {
+		if len(v.Finalizers) == 0 {
+			continue
+		}
 		newObj := v.DeepCopy()
 		newObj.SetFinalizers([]string{})
-		if !reflect.DeepEqual(newObj, v) {
-			if _, err = l.oldProjectAlerts.Update(newObj); err != nil {
-				return errors.Wrapf(err, "remove finalizer from legacy projectAlert %s:%s failed", newObj.Namespace, newObj.Name)
-			}
+		if _, err = l.oldProjectAlerts.Update(newObj); err != nil {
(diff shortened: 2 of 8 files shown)


pkg/auth/providerrefresh/refresher.go
```diff
@@ -206,6 +206,10 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 			secret := ""
 			if providers.ProvidersWithSecrets[providerName] {
 				secret, err = r.tokenMGR.GetSecret(user.Name, providerName, loginTokens[providerName])
+				if apierrors.IsNotFound(err) {
+					// There is no secret so we can't refresh, just continue to the next attribute
+					return attribs, nil
```
rmweir, line 211: I think this should be continue instead of a return.

pkg/controllers/managementuser/alert/deployer/upgradeimpl.go
```diff
@@ -397,12 +397,13 @@ func (l *AlertService) removeFinalizerFromLegacyAlerting() error {
 	}
 
 	for _, v := range oldProjectAlert {
+		if len(v.Finalizers) == 0 {
+			continue
+		}
 		newObj := v.DeepCopy()
 		newObj.SetFinalizers([]string{})
-		if !reflect.DeepEqual(newObj, v) {
```
rmweir, line 402: Why is has the DeepEqual been removed?

pkg/controllers/managementuser/secret/secret.go
```diff
@@ -115,6 +119,9 @@ func (n *NamespaceController) sync(key string, obj *corev1.Namespace) (runtime.O
 					continue
 				}
 				namespacedSecret := getNamespacedSecret(secret, obj.Name)
+				if _, err := n.clusterSecretsLister.Get(namespacedSecret.Namespace, namespacedSecret.Name); err == nil {
+					continue
+				}
 				logrus.Infof("Creating secret [%s] into namespace [%s]", namespacedSecret.Name, obj.Name)
 				_, err := n.clusterSecretsClient.Create(namespacedSecret)
 				if err != nil && !errors.IsAlreadyExists(err) {
```
rmweir, line 127: Shouldn't there be a continue in this block as well, instead of a return?

rmweir (changes requested):

pkg/controllers/managementuser/alert/deployer/upgradeimpl.go
ibuildthecloud: Because the len(v.Finalizers) checks the only thing we are modifying.  I believe that check always succeeds because `nil` != `[]string{}`

pkg/auth/providerrefresh/refresher.go
ibuildthecloud: This is to keep the same behavior that is done today.  Currently it will error out and never continue, but causes the handler to infinitely retry.  This way we don't retry anymore.  I'm concerned with continuing as I don't know what the correct behavior is in this situation as we haven't had this before.  Do we keep existing principals or delete them?  I think it's safer to just fix the performance issue and not address the logic issue.

pkg/controllers/managementuser/secret/secret.go
ibuildthecloud: That was the existing logic, I didn't want to change anything except the specific issue of not checking for the existence of the secret.

rmweir (approved):
LGTM. Feedback before was regarding issues that already exist or new code that is intended to minimize changes.

Issue #34914, opened 2021-09-24
RKE2 Provisioning - Nodepool doesn't self heal after correcting it's configuration. Custom Driver.


izaac:
**Rancher Server Setup**
- Rancher version: `v2.6.1-rc3`
- Installation option: Helm HA
   - If Helm Chart, Kubernetes Cluster and version: RKE1 k8s `v1.21.5`
- Proxy/Cert Details: `Bring your own valid`

**Information about the Cluster**
- Kubernetes version: `v1.21.4+rke2r2`
- Cluster Type: `Downstream
   - RKE 2 Cluster Oracle OCI Driver
   - `VM.Standard2.8` Nodes Shape

**Describe the bug**

An RKE2 Node Pool doesn't auto delete a failed node and provision a new node after editing the node pool with correct configuration parameters.

**To Reproduce**

- Create an RKE2 Oracle OCI cluster 
- Configure one node pool with all roles (etcd, control plane, worker) and right parameters.
- Configure a second node pool with worker role and one invalid parameter to make it fail provisioning.
- Wait until it's Provisioning with an error and colored red in the UI
- Edit the cluster Node Pool again and set the correct configuration parameters

**Result**

- The node with the error will stay in the pool and won't be deleted and no node with correct parameters will be provisioned.

I had to manually delete the node

**Expected Result**

- Node pool should auto-heal or time out somehow after some time (or if a new configuration is detected?) deleting the failed node and provisioned a new node with the updated parameters.

**Screenshots**

This is how the node pool looks in the UI after editing the Node Pool parameters with right values. It just hangs there.

![Screen Shot 2021-09-24 at 11 22 24 AM](https://user-images.githubusercontent.com/27659/134722650-53d49a78-3138-458c-a12d-3b46365d6d2c.png)

**Additional context**

If I delete the node manually from the UI then a new node with the Updated/Corrected parameters is provisioned and get active.

https://github.com/rancher/rancher/issues/34713

thedadams:
### Root cause
The root cause was the infrastructure objects were not getting re-enqueued to the controllers.

### What was fixed, or what changes have occurred
The infrastructure objects should be re-enqueued now when edits are made (not just when they are created/deleted).

### Areas or cases that should be tested
Creating clusters with nodes that are misconfigured should be tested:
1. Creating nodes in a cluster that are misconfigured (like in the description).
[...]

Issue #34945, opened 2021-09-28
RKE2 cluster stuck in deleting state on deleting it when one of the node pools failed provisioning


sowmyav27:
SURE-3585

**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `fe57abd`
- Installation option (Docker install/Helm Chart): HELM
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: rancher generated

**Information about the Cluster**
- Kubernetes version: 1.21.5 rke2
- Cluster Type (Local/Downstream): Downstream DO node driver

**Describe the bug**

**To Reproduce**
Deploy an RKE2 cluster - 1 node pool all roles.
Edit cluster - add a node pool - worker and save changes
The node pool is in error state
```
Failed creating server (DigitaloceanMachine) in infrastructure provider: CreateError: Failure detected from referenced resource rke-machine.cattle.io/v1, Kind=DigitaloceanMachine with name "do-test-pool2-27176adf-nmjh5": Running pre-create checks... Creating machine... (do-test-pool2-27176adf-nmjh5) Creating SSH key... (do-test-pool2-27176adf-nmjh5) Creating Digital Ocean droplet... Error creating machine: Error in driver during machine creation: POST https://api.digitalocean.com/v2/droplets: 422 You specified an invalid image for Droplet creation. The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.
```
- This seemed to be because of an invalid image specified.
- On deleting the cluster, the cluster is in `Removing` state  with message `Waiting for machine [do-test-pool2-797f58cbc9-b7rk4] to delete`
- The nodepool that failed provisioning is the only one pending in the machine pools page.

**Expected Result**
The cluster should get deleted, the node stuck in provisioning should also be deleted. (the node has not been created in the backend though in the first place)

snasovich:
@janeczku , thank you for providing this info. However, I'd say this is technically a separate issue as for the original issue we don't even have VMs in the infra provider while we do in your case.
Regardless, the linked PR addresses both issues. To be more precise, it addresses the original issue and provides backe support to address the issue you've outlined. UI ticket https://github.com/rancher/dashboard/issues/4538 was created to track adding support for this new functionality on UI side.

sowmyav27:
**On 2.6-head commit id: `638b29`**

- Deploy an RKE2 cluster - 1 node pool all roles.
[...]

PR #35403, opened 2021-11-08
Add support for deleting machines that failed creation

thedadams:
If a machine failed creation because of a user misconfiguration, then
when the user corrected their mistake a new machine would get created
but the old machine would not get removed. This change introduces
support for deleting machines without user interaction if the machine
failed creation. In addition, this change makes it possible for a user
to force deletion of a machine if the machine cannot be deleted for some
reason.

This change also cleans up the machine-status controller. A type is
created to simplify the return of a few functions.

A special note: conditions are added here to the infrastructure machine
objects. Based on some code here, it seems that they were always
intended to exist but were leftout.

Issue:
https://github.com/rancher/rancher/issues/34914


pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -238,88 +261,82 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
 		if err != nil {
 			return obj, err
 		}
-		if !removed {
-			h.dynamic.EnqueueAfter(obj.GetObjectKind().GroupVersionKind(), d.String("metadata", "namespace"), d.String("metadata", "name"), 5*time.Second)
-			return obj, generic.ErrSkip
+	}
+
+	if _, nodeDrainExcluded := machine.Annotations[capi.ExcludeNodeDrainingAnnotation]; !removed ||
+		(!nodeDrainExcluded && !drainingSucceededCondition.IsTrue(machine) && drainingSucceededCondition.GetReason(machine) != capi.DrainingFailedReason) {
```
Oats87, line 309:
```suggestion
		(!nodeDrainExcluded && drainingSucceededCondition.IsFalse(machine) && drainingSucceededCondition.GetReason(machine) != capi.DrainingFailedReason) {
```

Oats87: Shouldn't this either be True or False?

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -543,3 +567,20 @@ func constructFilesSecret(driver string, config map[string]interface{}) (*corev1
...
+	var createJobCondition summary.Condition
+	for _, c := range summary.GetUnstructuredConditions(d) {
+		if c.Type() == createJobConditionType {
+			createJobCondition = c
+			break
+		}
+	}
+
+	if createJobCondition.Reason() != string(errors.CreateMachineError) &&
+		d.Map("metadata", "annotations")[forceRemoveMachineAnn] != "true" {
```
Oats87, line 581: nit: should we `strings.ToLower` the annotation value?

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -543,3 +567,20 @@ func constructFilesSecret(driver string, config map[string]interface{}) (*corev1
...
+func shouldForceRemoval(job *batchv1.Job, d data.Object) bool {
+	var createJobCondition summary.Condition
+	for _, c := range summary.GetUnstructuredConditions(d) {
+		if c.Type() == createJobConditionType {
+			createJobCondition = c
+			break
+		}
+	}
+
+	if createJobCondition.Reason() != string(errors.CreateMachineError) &&
```
Oats87, line 580: I take it that expected behavior here is that if the `createJobCondition.Reason()` is in fact the `CreateMachineError`, we should fall down to the bottom and return true if the failure threshold has been passed for the create job?

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -543,3 +567,20 @@ func constructFilesSecret(driver string, config map[string]interface{}) (*corev1
...
+		d.Map("metadata", "annotations")[forceRemoveMachineAnn] != "true" {
+		return false
+	}
+
+	return (job.Spec.BackoffLimit == nil && job.Status.Failed > 0) || (job.Spec.BackoffLimit != nil && job.Status.Failed > *job.Spec.BackoffLimit)
```
Oats87, line 585: Is the goal here to look to see if the job has failed? My understanding is there should be a defined job status condition that either marks succeeded or failed. https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/#JobStatus

Also, I may be interpreting this wrong, but `.Failed` should be the number of pods that have "failed", I'd expect that check to be `job.Status.Failed >= *job.Spec.BackoffLimit`. Additionally, people have seemingly encountered problems where a job gets marked as Failed via condition while the `Failed` is < `BackoffLimit`: https://stackoverflow.com/questions/54825671/understanding-backofflimit-in-kubernetes-job

pkg/controllers/provisioningv2/rke2/machinestatus/controller.go
```diff
@@ -0,0 +1,241 @@
...
+			if owner.Kind == "Machine" {
+				h.machines.Enqueue(m.GetNamespace(), owner.Name)
+			}
+		}
+		return obj, nil
+	})
+}
+
+func (h *handler) setJoinURLFromOutput(machine *capi.Machine, nodePlan *plan.Node) error {
+	if nodePlan == nil || !planner.IsEtcdOnlyInitNode(machine) || machine.Annotations[planner.JoinURLAnnotation] != "" {
```
Oats87, line 94: `!planner.IsEtcdOnlyInitNode(machine)`, so if the node is not an etcd-only init node, we just return nil? What if it's a controlplane+etcd init node?

What happens if the join URL changes for this node, i.e. if the node IP changes? We won't update the JoinURLAnnotation?

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
```diff
@@ -1,328 +1,160 @@
...
 
-	return "", "", "", obj.String("spec", "providerID"), nil
-}
-
-type dbinfo struct {
-	Members []member `json:"members,omitempty"`
-}
-type member struct {
-	ClientURLs []string `json:"clientURLs,omitempty"`
+	return &machineStatus{cond: Provisioned, providerID: obj.String("spec", "providerID")}, nil
```
Oats87, line 159: Should we be setting `status` to `corev1.ConditionTrue`?

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: Yes, that is correct.

pkg/controllers/provisioningv2/rke2/machinestatus/controller.go
thedadams: First, I'll not that this code hasn't changed in the PR. It was just moved to a different file.

This function is only meant to handle the case where the init-node is an etcd only node. In every other instance, the join URL is set in the planner.

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
thedadams: No, because if the status is empty here, then [here](https://github.com/rancher/rancher/pull/35403/files#diff-dbeeb0a3acefea09a1c49792078d38eeab261888ec9f34170f24fac122376f4bR208) Rancher will get the status from the plan.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: I will use the condition instead.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: I will switch this.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: I will do this.

thedadams:
https://github.com/rancher/rancher/issues/34945

pkg/controllers/provisioningv2/rke2/machinestatus/controller.go
Oats87: Thanks for the clarification! And sorry, I didn't realize it was just moved haha

pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go
Oats87: OK, thanks!

Oats87 (dismissed):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -187,29 +195,44 @@ func (h *handler) getMachineStatus(job *batchv1.Job, create bool) (rkev1.RKEMach
...
-			} else {
-				reason = string(errors.DeleteMachineError)
-			}
 			return rkev1.RKEMachineStatus{
+				Conditions: []genericcondition.GenericCondition{
+					{
+						Type:    condType,
+						Status:  corev1.ConditionFalse,
+						Reason:  reason,
+						Message: strings.TrimSpace(pod.Status.ContainerStatuses[0].State.Terminated.Message),
```
kinarashah, line 233: Is it obvious that we only need message from the first container status when we're looping and returning only if exit code is non zero? Shouldn't this be `containerStatus.State.Terminated.Message`?

thedadams: @kinarashah Good point. I was just copying what was already there, but your suggestion makes more sense. I updated this.

kinarashah (approved):

Oats87 (approved):

Issue #30053, opened 2020-11-12
Update ingress api when lowest supported k8s version is 1.19


brendarearden:
When installing with `helm template rancher` the Ingress resource was showing the following warning `Warning: extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress` this was resolved for `helm install` in this issue https://github.com/rancher/rancher/issues/29144 but still warns with helm template, unless passing in the apiVersion. 

The resolution added an [if else statement](https://github.com/rancher/rancher/pull/29922/files) which defaults to `networking.k8s.io/v1beta1` as the else. In 1.22+ , `networking.k8s.io/v1beta1` will be unavailable so the if statement needs to be removed and/or the default needs to be `networking.k8s.io/v1`

cbron:
@brendarearden above we wrote: 
> In 1.22+ , networking.k8s.io/v1beta1 will be unavailable

But its only extensions.k8s.io thats removed right ? Is `networking.k8s.io/v1beta1` deprecated anywhere ? I don't see that announcement.

brendarearden:
Based on the [deprecation guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/) I am going to update this card to be: Update ingress when lowest supported k8s version is 1.19. 

This is because the `networking.k8s.io/v1 API version` is only available starting in 1.19. We are currently supporting 1.18 for the 2.6 release so the api version check is still necessary in our rancher helm chart. 

```
Ingress
The extensions/v1beta1 and networking.k8s.io/v1beta1 API versions of Ingress will no longer be served in v1.22.
Migrate manifests and API clients to use the networking.k8s.io/v1 API version, available since v1.19.
All existing persisted objects are accessible via the new API
Notable changes:
spec.backend is renamed to spec.defaultBackend
The backend serviceName field is renamed to service.name
Numeric backend servicePort fields are renamed to service.port.number
String backend servicePort fields are renamed to service.port.name
pathType is now required for each specified path. Options are Prefix, Exact, and ImplementationSpecific. To match the undefined v1beta1 behavior, use ImplementationSpecific.

IngressClass
The networking.k8s.io/v1beta1 API version of IngressClass will no longer be served in v1.22.

Migrate manifests and API clients to use the networking.k8s.io/v1 API version, available since v1.19.
All existing persisted objects are accessible via the new API
No notable changes
```

cbron:
[...]

PR #35426, opened 2021-11-09
Fix ingress template for `helm template` command

cmurphy:
Without this patch, the `helm template` command generates Ingress
resources that use the `networking.k8s.io/v1beta1` APIVersion. This is
because the `helm template` command does not populate the
`Capabilities.APIVersions` set with real available resources, it only
creates a fake set of API versions that does not include resources, so
it does not have `networking.k8s.io/v1/Ingress` in the set. Since the
fake set also does not include `networking.k8s.io/v1beta1/Ingress`, we
can assume in that case that we are running in `helm template` and we
should default to the v1 version.

https://github.com/rancher/rancher/issues/30053


paynejacob (approved):

brendarearden (approved):
LGTM

Issue #35203, opened 2021-10-20
Support SSH for V2 (RKE2) provisioned nodes


snasovich:
Currently Rancher CLI's `ssh` command doesn't support V2 (RKE2) provisioned nodes. For example:
```
./rancher ssh sergey-rke2-for-erd-discussion-sergey-rke2-for-erd-discus-2pgnp
FATA[0001] failed to find SSH key for sergey-rke2-for-erd-discussion-sergey-rke2-for-erd-discus-2pgnp
```

This works fine for RKE1 provisioned clusters.

Rancher CLI should be updated to support SSH to nodes of V2 provisioned clusters.

thedadams:
### Root cause
The CLI was looking for nodes with SSH config associated to them. In provisioningv2, the CLI should look for machines.

### What was fixed, or what changes have occurred
Generated clients were added to Rancher and the CLI will look for machines with the name passed if the node doesn't have SSH config associated with it.

### Areas or cases that should be tested
1. Provision an RKE2 cluster from Rancher.
2. Use Rancher CLI v2.6.0-rc1 or higher to ssh one of the nodes.

### What areas could experience regressions?
Using Rancher CLI to SSH to nodes that were provisioned for RKE1 clusters.
1. Provision an RKE1 cluster from Rancher.
2. Ensure that you can use Rancher CLI v2.6.0-rc1 or higher to SSH to one of the nodes.

### Are the repro steps accurate/minimal?
Yes.

Auston-Ivison-Suse:
**Setup For Reproduction**
- rancher version: v2.6-head (e655fd8) 
- ec2 RKE2 cluster
- rancher CLI version: 2.4.13

**Reproduction Steps**
1. From local terminal using rancher CLI, run:
```
$ rancher login <url> --token <bearer token> 
```
2. You will be prompted to select project, this should be your rke2 cluster
3. Now you can get your node info in the cli by running: 
```
$ rancher nodes
```
4. The following command will give you the failure: 
```
rancher ssh <ID/name>
```

**Results**

Following the above steps you will see "FATA[0001] failed to find SSH key for <ID/name>"

thedadams:
Rancher CLI version 2.6.0-rc2 is necessary for this issue. It contains the fixes necessary.

Auston-Ivison-Suse:
**Further Testing**
-Using the new rc here are the commands I ran
```
[...]

PR #35431, opened 2021-11-09
Machine client fixes

thedadams:
The CAPI machine client uses the ID cluster.x-k8s.io.machine, but the
client generator assumed the ID is the same as the type name (machine in
this case). Therefore, an adjustment has been made to ensure the CAPI
machine client uses the correct ID.

Issue:
https://github.com/rancher/rancher/issues/35203


paynejacob (approved):

rmweir (approved):
LGTM

Issue #20654, opened 2019-06-04
Add functionality to specify limit of app revisions stored


cnuber:
There is currently no limit to the number of apprevision objects that will be stored for an app deployed through the Rancher Catalog.  Over time this can lead to performance issues in the etcd store and likely Rancher itself.  For users that wish to deploy their custom applications as Rancher Catalog Apps this number of objects can increase rapidly with continuous deployment.
**What kind of request is this (question/bug/enhancement/feature request):**
enhancement
**Steps to reproduce (least amount of steps as possible):**
1.  Deploy an app via the Rancher Catalog.
2. Upgrade/Modify the app
**Result:**
A new apprevision object is created for the app.  
**Other details that may be helpful:**
The desired functionality would be to have the ability to set a limit for the number of apprevisions that can be stored for an app when deploying it.  The ability to delete old app revisions from the Rancher UI and CLI is also desired.
**Environment information**
- Rancher version 2.2.3
- Installation option (single install/HA): Single and HA

gz#14990
gz#16168
gz#15715
gz#14921

ebauman:
This brought down a Rancher cluster because of an insane number of app revisions.

"Something" happened to cause a failure in a monitoring deployment in a downstream cluster. I was not able to determine the RCA of that. 
However, the outcome was that the redeploying of our monitoring tools caused 274,000 apprevisions to clog up etcd. This, in turn, caused an OOM issue in kube-apiserver because of all the watches.

We need to cap the apprevisions to prevent this failure scenario.

evhan:
The same happened to us, ~2GB of apprevisions built up for apps that are updated regularly for CI purposes, which brought etcd to its knees. We were surprised to find that old apprevisions are never pruned. Beyond just allowing this to be configured, I think it would be good if there were a default limit in place as well, something like 10 revisions.

abh:
Our cattle-prometheus / monitoring app keeps getting redeployed / changed — over some months it built up ~100k revisions which, yes, also made the cluster unusable until we shutdown rancher and “by hand” got etcd purged of this.

superseb:
Discussed to fix this:
- Add a post action to install/upgrade to fetch existing revisions and remove anything over the max amount of revisions configured (hardcoded or configurable)

genexpr:
Also, will need to establish the rules for validating the max number of stored app revisions. Would zero be allowed?

StrongMonkey:
[...]

PR #35433, opened 2021-11-09
After creating an app revision on install/upgrade, prune old revisions

genexpr:
Issue [#20654](https://github.com/rancher/rancher/issues/20654)
The Helm controller will now limit the number of app revisions per app. By default, up to 10 revisions will be kept in etcd per app.
Users will be able to set the limit themselves. But only on app installation and only by modifying the API request.

This affects new apps and existing apps. For existing apps, a new upgrade/rollback will trigger the pruning of old app revisions. For example, if an app has 200 revisions stored in etcd, then on the next app change, a new revision will be created and 191 of the old ones will be deleted, so Rancher will maintain only the 10 newest revisions (by default) at all times.

### DIFF
--- a/pkg/controllers/managementuser/helm/controller.go
+++ b/pkg/controllers/managementuser/helm/controller.go
@@ -39,6 +39,7 @@ const (
 	creatorIDAnn              = "field.cattle.io/creatorId"
 	MultiClusterAppIDSelector = "mcapp"
 	projectIDFieldLabel       = "field.cattle.io/projectId"
+	defaultMaxRevisionCount   = 10
 )
 
 func Register(ctx context.Context, user *config.UserContext, kubeConfigGetter common.KubeConfigGetter) {
@@ -319,6 +320,7 @@ func (l *Lifecycle) Remove(obj *v3.App) (runtime.Object, error) {
 }
 
 func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath) error {
+	defer l.pruneOldRevisions(obj)
 	err := l.writeKubeConfig(obj, tempDirs.KubeConfigFull, false)
 	if err != nil {
 		return err
@@ -334,6 +336,33 @@ func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath
 		return err
 	}
 	return l.createAppRevision(obj, template, notes, false)
+}
+
+// pruneOldRevisions checks if the total number of app revisions does not exceed the set maximum.
+// If it does, the method deletes the oldest revision(s) to maintain the maximum number of revisions and no more.
+func (l *Lifecycle) pruneOldRevisions(obj *v3.App) {
+	if obj.Spec.MaxRevisionCount < 1 {
+		obj.Spec.MaxRevisionCount = defaultMaxRevisionCount
+	}
+	_, projectName := ref.Parse(obj.Spec.ProjectName)
+	appRevisionClient := l.AppRevisionGetter.AppRevisions(projectName)
+	revisions, err := appRevisionClient.List(metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", appLabel, obj.Name),
+	})
+	if err != nil {
+		logrus.Tracef("[helm-controller] Failed to list revisions for app %s: %v", projectName, err)
+		return
+	}
+
+	if len(revisions.Items) > obj.Spec.MaxRevisionCount {
+		logrus.Tracef("[helm-controller] App %s is exceeding the maximum (%d) number of stored revisions", obj.Name, obj.Spec.MaxRevisionCount)
+		deleteCount := len(revisions.Items) - obj.Spec.MaxRevisionCount
+		for _, revision := range revisions.Items[:deleteCount] {
+			if err := appRevisionClient.Delete(revision.Name, &metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
+				logrus.Tracef("[helm-controller] Failed to delete app revision %s: %v", revision.Name, err)
+			}
+		}
+	}
 }
 
 func (l *Lifecycle) createAppRevision(obj *v3.App, template, notes string, failed bool) error {
--- a/pkg/controllers/managementuser/helm/controller.go
+++ b/pkg/controllers/managementuser/helm/controller.go
@@ -350,7 +350,7 @@ func (l *Lifecycle) pruneOldRevisions(obj *v3.App) {
 		LabelSelector: fmt.Sprintf("%s=%s", appLabel, obj.Name),
 	})
 	if err != nil {
-		logrus.Tracef("[helm-controller] Failed to list revisions for app %s: %v", projectName, err)
+		logrus.Warnf("[helm-controller] Failed to list revisions for app %s: %v", projectName, err)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/helm/controller.go
```diff
@@ -336,6 +338,32 @@ func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath
...
+		LabelSelector: fmt.Sprintf("%s=%s", appLabel, obj.Name),
+	})
+	if err != nil {
+		logrus.Tracef("[helm-controller] Failed to list app revisions for app %s: %v", projectName, err)
+	}
+
+	if len(revisions.Items) >= obj.Spec.MaxRevisionCount {
+		logrus.Tracef("App %s is exceeding the maximum (%d) number of stored revisions", obj.Name, obj.Spec.MaxRevisionCount)
+		deleteCount := len(revisions.Items) - obj.Spec.MaxRevisionCount
+		for _, revision := range revisions.Items[:deleteCount] {
```
nickgerace, line 359: Are we ok with not iterating over the revision lists when `len(revisions.Items) == max`? I see the "exceeding the maximum" log so I am curious.
https://play.golang.org/p/j-ChGHx-hco

pkg/controllers/managementuser/helm/controller.go
```diff
@@ -336,6 +338,32 @@ func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath
...
+		obj.Spec.MaxRevisionCount = defaultMaxRevisionCount
+	}
+	_, projectName := ref.Parse(obj.Spec.ProjectName)
+	appRevisionClient := l.AppRevisionGetter.AppRevisions(projectName)
+	revisions, err := appRevisionClient.List(metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", appLabel, obj.Name),
+	})
+	if err != nil {
+		logrus.Tracef("[helm-controller] Failed to list app revisions for app %s: %v", projectName, err)
+	}
```
nickgerace, line 354: We should probably return here. Maybe we return an error and the parent function chooses to ignore it?

nickgerace (changes requested):

pkg/controllers/managementuser/helm/controller.go
genexpr: A fair point! Yes, I originally did return it, but now just log it for a few reasons.
1. We should handle it in the method's deferred call, which is hard to do cleanly.
2. This error is not likely to happen on its own, without other `appRevisionClient`'s operations not failing too in adjacent controller handlers and bringing attention to itself.
3. This pruning is not critical enough to stop whatever the controller is doing.

But I completely agree with you, although making the parent ignore the error won't really change anything.

nickgerace: Agreed with your points. Regardless, we shouldn't continue the processing here right because those revisions will likely be `nil`, right?

pkg/controllers/managementuser/helm/controller.go
genexpr: Nick, you won't believe it, but I wrote the exact same Playground example before opening the PR to test this Go slice behavior! Wow!
But yeah, now that the pruning is done after a new revision is created, I should update it to just be `len(revisions.Items) > max` and not `>=`.

pkg/controllers/managementuser/helm/controller.go
genexpr: Yes, good catch!

niusmallnan:
@genexpr There was a setting named HelmMaxHistory before: https://github.com/rancher/rancher/pull/28754
Not sure if we can reuse it to avoid user confusion.

genexpr:
@niusmallnan, sure I'd be willing to use an existing setting then. I have searched in the code and issues, and could not find any description or documentation for `HelmMaxHistory`. Is it definitely for limiting the number of app revisions? Or something else/more? Is it set per app or globally?
@brendarearden

pkg/controllers/managementuser/helm/controller.go
genexpr: This behavior is also interesting:
https://play.golang.org/p/eOh1Yh8KqjN

nickgerace (dismissed):

genexpr:
@niusmallnan, about reusing `HelmMaxHistory`: I checked with @StrongMonkey, he confirmed that this setting is for something else. Specifically, the number of Helm release revisions, which is not what we want here, it's an entirely different thing. We want to have control over **app** revisions for each individual app. So, we will keep a separate default constant.

nickgerace (dismissed):

pkg/controllers/managementuser/helm/controller.go
```diff
@@ -336,6 +338,33 @@ func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath
 	return l.createAppRevision(obj, template, notes, false)
 }
 
+// pruneOldRevisions checks if the total number of app revisions does not exceed the set maximum.
+// If it does, the method deletes the oldest revision(s) to maintain the maximum number of revisions and no more.
+func (l *Lifecycle) pruneOldRevisions(obj *v3.App) {
+	if obj.Spec.MaxRevisionCount < 1 {
```
StrongMonkey, line 344: If number is negative, should we skip the prune logic? This provides a way to skip garbage collection.

pkg/controllers/managementuser/helm/controller.go
```diff
@@ -336,6 +338,33 @@ func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath
...
+		obj.Spec.MaxRevisionCount = defaultMaxRevisionCount
+	}
+	_, projectName := ref.Parse(obj.Spec.ProjectName)
+	appRevisionClient := l.AppRevisionGetter.AppRevisions(projectName)
+	revisions, err := appRevisionClient.List(metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", appLabel, obj.Name),
+	})
+	if err != nil {
+		logrus.Tracef("[helm-controller] Failed to list revisions for app %s: %v", projectName, err)
```
StrongMonkey, line 353: Would be better to to warning log in case error happens.

pkg/controllers/managementuser/helm/controller.go
```diff
@@ -336,6 +338,33 @@ func (l *Lifecycle) Run(obj *v3.App, template string, tempDirs *hCommon.HelmPath
...
+		return
+	}
+
+	if len(revisions.Items) > obj.Spec.MaxRevisionCount {
+		logrus.Tracef("[helm-controller] App %s is exceeding the maximum (%d) number of stored revisions", obj.Name, obj.Spec.MaxRevisionCount)
+		deleteCount := len(revisions.Items) - obj.Spec.MaxRevisionCount
+		for _, revision := range revisions.Items[:deleteCount] {
+			if err := appRevisionClient.Delete(revision.Name, &metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
+				logrus.Tracef("[helm-controller] Failed to delete app revision %s: %v", revision.Name, err)
```
StrongMonkey, line 362: Same, warning log is better here.

pkg/controllers/managementuser/helm/controller.go
genexpr: We're just enforcing the default value for `MaxRevisionCount` (which can only be set once on app creation). If the number is negative, and we skip pruning, then new revisions will continue to be created without bounds, right?

StrongMonkey: Yeah if number is negative then we don't purge revisions.

genexpr: We do now, since we set the max to 10 and check how many exist currently. So we can keep the current behavior, right?

StrongMonkey: talked offline, it should be fine

StrongMonkey (approved):

nickgerace (approved):

PR #35434, opened 2021-11-10
Adding fixes for hardened local cluster

sowmyav27:
Was missing sudo in the sysctl commands, and hence the commands never ran on the nodes. And rke up resulted in failure. Kubelet on the nodes were in a state of panic.


jmcsagdc (approved):

izaac (approved):
LGTM

Issue #35413, opened 2021-11-09
[KEv2] Cluster owners should be able to change cloud credential


thedadams:
**Describe the bug**

Currently, a cluster owner of a KEv2 cluster can only change the cloud credential if they have access to the one currently on the cluster AND have access to the one they are trying to change it to.

**To Reproduce**
 
1. As an admin, create an EKS/GKE/AKS cluster
2. Create a standard user
3. Assign the standard user as a cluster owner of the EKS/GKE/AKS cluster
4. Logout as admin and login as the standard user.
5. Edit the cluster and try to change the cloud credential (you may have to create a cloud credential for the standard user).

**Result**
The standard user is not allowed to change the cloud credential.

**Expected Result**

The standard user should be able to do this.

thedadams:
Related UI issue: https://github.com/rancher/dashboard/issues/4529

brudnak:
## Reproduction Environment


| Component               | Version / Type                          |
| ----------------------- | --------------------------------------- |
| Rancher version         | v2.6.2                                  |
| Rancher cluster type    | Docker                                  |
| Docker version          | 20.10.7, build 20.10.7-0ubuntu5~20.04.2 |
| Helm version            | v2.16.8-rancher1                        |
| Downstream cluster type | AWS EKS                                 |
| Downstream K8s version  | 1.21                                    |
| Browser type            | Google Chrome                           |
| Browser version         | 96.0.4664.55 (Official Build) (x86_64)  |
| Cloud Credentials       | AWS Access Token Setup Prior to Steps   |

![01-repro-env](https://user-images.githubusercontent.com/40877149/143289935-8a860524-741d-4199-9753-1a0e0e4f0432.png)

## Reproduction steps

1. Starting from the **default** Rancher homepage `/dashboard/home` & logged in as the default `Admin` user

>Create an EKS cluster

2. Click `hamburger menu` -> click `Create`
3. Click `Amazon EKS`
4. Fill out the form for `Amazon EKS` cluster creation
    1. Cluster Name: `abrudna-qa-eks`
    2. Click `Next: Configure Cluster`
    3. Click `Next: Configure Network`
    4. Select `Custom: Choose from your existing VPC and Subnets`
    5. Click `Next: Select Security Group`
    6. Select desired security group
    7. Node Group Name: `abrudna-qa-node-g`
    8. Click `Create` -> you are then redirected back to the `Clusters` page
    9. Wait for the newly created AWS EKS cluster to have an `Active` state

[...]

Issue #34800, opened 2021-09-18
Cluster owner not able to edit Harvester cluster without accessing cloud credential


guangbochen:
**Rancher Server Setup**
- Rancher version: `v2.6-head`
- Installation option (Docker install/Helm Chart): All works
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details:

**Describe the bug**
Cloud credentials aren't shareable, so if you create a cluster with user A, make user B a cluster owner (or other roles), and then come to the edit screen as user B, you will not find the credential here.

**To Reproduce**
1. login as a cluster-admin user A, and create a new cluster using Harvester/VMWare vSphere RKE2 node driver
2. add another user B to the cluster as a cluster owner and create the cluster
3. login as with a user B
4. click to edit the cluster and it shows `cloud credential no found`, therefore users have to add custom cloud credentials in order to edit this cluster properly
![image](https://user-images.githubusercontent.com/4569037/133878526-8cdcb9db-9b50-44b7-85dd-ea1de966e974.png)

**Result**
The user B may not have access to the VMWare vSphere, but would prefer to edit the assigned cluster as a cluster owner.

**Expected Result**
~~Cloud credentials should be shareable for the cluster owners(but may need some concerns of not expose the credentails)~~ 

**Screenshots**

**Additional context**
as mentioned in the https://github.com/rancher/dashboard/pull/3881

kinarashah:
**There are two ways we can scale up nodes:** 
1) Nodes page - using `(+)` in UI to update node pools 
2) Edit Config/Edit Cluster and increment machine count 

**RKE1:** 
- Cluster owner is able to scale up nodes using both Nodes page and Edit Config. 
- It works for all node drivers, tested on EC2, DO and Vsphere. 
- It works for both v2.5.8, v2.5.9 and v2.6.0. 

**RKE2:**   
- Cluster owner is able to scale up nodes for all node drivers (tested on EC2, DO and Vsphere) **only** using Nodes page

**RKE2: Edit Config behavior** This changes based on the driver: 
**EC2**
On Edit Config, there's an error - `The request must contain the parameter AWSAccessKeyId`. 
- This error doesn't seem to prevent the Edit Cluster. 
- On Save, there's a new error: `"VPC/Subnet" is required` which blocks the save. 
- Edit as YAML works. 

**DigitalOcean** 
[...]

PR #35448, opened 2021-11-10
Allow cloud credential usage for owned clusters

thedadams:
If a user is an owner of a cluster, then they should be allowed to use
the cloud credential associated to that cluster in the meta proxy. This
really only applies to KEv2 and RKE2/K3S provisioned clusters because
those clusters have cloud credentials associated to them.

In addition, if a user is a owner of a cluster, they should also be able
to change the cloud credential associated to a cluster to a cloud
credential they own regardless if they have access to the old cloud
credential.

This PR also fixes an obscure issue when running Rancher in a debugger (see the second commit message for details).

Issues:
https://github.com/rancher/rancher/issues/35413
https://github.com/rancher/rancher/issues/34800

### DIFF
--- a/pkg/api/norman/customization/aks/handler.go
+++ b/pkg/api/norman/customization/aks/handler.go
@@ -16,6 +16,8 @@ import (
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
+	"github.com/rancher/rancher/pkg/controllers/management/cluster"
+	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/namespace"
@@ -47,7 +49,7 @@ type Capabilities struct {
 type handler struct {
 	schemas       *types.Schemas
 	secretsLister v1.SecretLister
-	clusterLister v3.ClusterLister
+	clusterCache  mgmtv3.ClusterCache
 	ac            types.AccessControl
 	secretClient  v1.SecretInterface
 }
@@ -56,7 +58,7 @@ func NewAKSHandler(scaledContext *config.ScaledContext) http.Handler {
 	return &handler{
 		schemas:       scaledContext.Schemas,
 		secretsLister: scaledContext.Core.Secrets(namespace.GlobalNamespace).Controller().Lister(),
-		clusterLister: scaledContext.Management.Clusters("").Controller().Lister(),
+		clusterCache:  scaledContext.Wrangler.Mgmt.Cluster().Cache(),
 		ac:            scaledContext.AccessControl,
 		secretClient:  scaledContext.Core.Secrets(namespace.GlobalNamespace),
 	}
@@ -102,7 +104,7 @@ func (h *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 
 	switch resourceType {
 	case "aksUpgrades":
-		if serialized, errCode, err = listKubernetesUpgradeVersions(req.Context(), h.clusterLister, capa); err != nil {
+		if serialized, errCode, err = listKubernetesUpgradeVersions(req.Context(), h.clusterCache, capa); err != nil {
 			logrus.Errorf("[aks-handler] error getting kubernetes upgrade versions: %v", err)
 			handleErr(writer, errCode, err)
 			return
@@ -255,28 +257,39 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
 }
 
 func (h *handler) clusterCheck(apiContext *types.APIContext, clusterID, cloudCredentialID string) (int, error) {
-	clusterInfo := map[string]interface{}{
-		"id": clusterID,
-	}
-
-	clusterSchema := h.schemas.Schema(&mgmtSchema.Version, client.ClusterType)
-	if err := h.ac.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", apiContext, clusterInfo, clusterSchema); err != nil {
-		return httperror.InvalidBodyContent.Status, fmt.Errorf("cluster not found")
-	}
-
-	cluster, err := h.clusterLister.Get("", clusterID)
-	if err != nil {
-		if httperror.IsNotFound(err) {
-			return httperror.InvalidBodyContent.Status, fmt.Errorf("cluster not found")
+	var (
+		clusters []*v3.Cluster
+		err      error
+	)
+	if clusterID == "" {
+		// If no clusterID is passed, then we check all clusters that the user has access to and are associated to the cloud credential.
+		clusters, err = h.clusterCache.GetByIndex(cluster.ByCloudCredential, cloudCredentialID)
+		if err != nil {
+			return httperror.InvalidBodyContent.Status, err
 		}
-		return httperror.ServerError.Status, err
+		if len(clusters) == 0 {
+			return httperror.InvalidBodyContent.Status, fmt.Errorf("cloud credential not found")
+		}
+	} else {
+		c, err := h.clusterCache.Get(clusterID)
+		if err != nil {
+			return httperror.ServerError.Status, err
+		}
+		clusters = []*v3.Cluster{c}
 	}
 
-	if cluster.Spec.AKSConfig.AzureCredentialSecret != cloudCredentialID {
-		return httperror.InvalidBodyContent.Status, fmt.Errorf("cloud credential not found")
+	for _, c := range clusters {
+		if c.Spec.AKSConfig == nil || c.Spec.AKSConfig.AzureCredentialSecret != cloudCredentialID {
+			continue
+		}
+
+		clusterSchema := h.schemas.Schema(&mgmtSchema.Version, client.ClusterType)
+		if err := h.ac.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", apiContext, map[string]interface{}{"id": c.Name}, clusterSchema); err == nil {
+			return http.StatusOK, nil
+		}
 	}
 
-	return http.StatusOK, nil
+	return httperror.InvalidBodyContent.Status, fmt.Errorf("cloud credential not found")
 }
 
 func (h *handler) getCredentialsFromBody(req *http.Request, cap *Capabilities) (int, error) {
--- a/pkg/api/norman/customization/gke/handler.go
+++ b/pkg/api/norman/customization/gke/handler.go
@@ -11,6 +11,8 @@ import (
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
+	"github.com/rancher/rancher/pkg/controllers/management/cluster"
+	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/namespace"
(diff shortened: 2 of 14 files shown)


cmurphy (dismissed):

pkg/httpproxy/proxy.go
```diff
@@ -230,16 +234,27 @@ func (p *proxy) secretGetter(req *http.Request, cAuth string) SecretGetter {
...
-	if cluster.Spec.EKSConfig.AmazonCredentialSecret != credID {
-		return authorizer.DecisionDeny, nil
+
+	if administratedByProvV2Cluster {
+		provCluster, err := p.provClusters.Get(cluster.Annotations["objectset.rio.cattle.io/owner-namespace"], cluster.Annotations["objectset.rio.cattle.io/owner-name"])
+		if err != nil {
+			return authorizer.DecisionDeny, err
+		}
+
+		if provCluster.Status.ClusterName != clusterID || provCluster.Spec.CloudCredentialSecretName != credID {
```
rmweir, line 253: Is there any validation to prevent someone from adding any credential they like to Spec.CloudCredentialSecretName or changing the Status.ClusterName? If status is a block subschema that should be fine too. ClusterName may be fine if them being able to edit provisioning cluster means they are an owner, similar to norman clusters since to do anything with this they'd need to have access to both clusters anyway.

rmweir (changes requested):

pkg/httpproxy/proxy.go
```diff
@@ -57,7 +59,8 @@ type proxy struct {
 	prefix             string
 	validHostsSupplier Supplier
 	credentials        v1.SecretInterface
-	clusters           v3.ClusterInterface
+	mgmtClusters       v3.ClusterInterface
```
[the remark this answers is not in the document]
thedadams: Done.

pkg/httpproxy/proxy.go
```diff
@@ -57,7 +59,8 @@ type proxy struct {
...
+	provClusters       provv1.ClusterCache
```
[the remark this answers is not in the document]
thedadams: Done.

pkg/httpproxy/proxy.go
thedadams: 1. Status is a sub-object of the provisioning cluster.
2. There is logic [in the webhook](https://github.com/rancher/webhook/blob/master/pkg/resources/validation/cluster/provisioningcluster.go#L67) that validates that the user has access to the cloud credential they are trying to change to.

pkg/httpproxy/proxy.go
```diff
@@ -230,16 +234,27 @@ func (p *proxy) secretGetter(req *http.Request, cAuth string) SecretGetter {
 }
 
 func (p *proxy) checkCluster(req *http.Request, user user.Info, clusterID, credID string) (authorizer.Decision, error) {
```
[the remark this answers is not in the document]
thedadams: Done.

pkg/httpproxy/proxy.go
```diff
@@ -230,16 +234,27 @@ func (p *proxy) secretGetter(req *http.Request, cAuth string) SecretGetter {
...
-	cluster, err := p.clusters.Controller().Lister().Get("", clusterID)
+	cluster, err := p.mgmtClusters.Controller().Lister().Get("", clusterID)
 	if err != nil {
 		return authorizer.DecisionDeny, err
 	}
-	if cluster.Spec.EKSConfig == nil {
+
+	administratedByProvV2Cluster := imported.IsAdministratedByProvisioningCluster(cluster)
+	if (cluster.Spec.EKSConfig == nil || cluster.Spec.EKSConfig.AmazonCredentialSecret != credID) && !administratedByProvV2Cluster {
```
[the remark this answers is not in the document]
thedadams: I added a comment. There is some subtlety here because KEv2 clusters and provisioning v2 clusters are the only clusters that have a cloud credential on them. We don't check GKE and AKS clusters here because they have their own proxy.

thedadams:
The implementation was reworked to allow cloud credential usage if a user has access to any cluster associated to the cloud credential.

pkg/httpproxy/proxy.go
```diff
@@ -229,27 +234,96 @@ func (p *proxy) secretGetter(req *http.Request, cAuth string) SecretGetter {
...
+				continue
+			}
+
+			clusterID = cluster.Status.ClusterName
+		}
+
+		decision, _, err := p.authorizer.Authorize(req.Context(), authorizer.AttributesRecord{
+			User:            user,
+			Verb:            "update",
+			APIGroup:        v3.GroupName,
```
cmurphy, line 299: Isn't this only for management clusters? Do we also need to check authorization on the provisioning cluster?

thedadams: If a user has access to a provisioning cluster, then they will also have access to the corresponding management cluster. So, in the case of a provisioning cluster, we use the `cluster.Status.ClusterName` which is the name of the management cluster.

cmurphy (dismissed):

pkg/httpproxy/proxy.go
```diff
@@ -229,27 +234,96 @@ func (p *proxy) secretGetter(req *http.Request, cAuth string) SecretGetter {
...
+			if (cluster.Spec.EKSConfig == nil || cluster.Spec.EKSConfig.AmazonCredentialSecret != credID) && !administratedByProvV2Cluster {
+				continue
+			}
+
+			if administratedByProvV2Cluster {
+				c, err = p.provClustersCache.GetByIndex(provcluster.ByCluster, cluster.Name)
+				if err != nil {
+					c = nextCluster(mgmtClusters, provClusters, &i)
+				}
+				// If there is no error, then we want to go through the loop again with the provisioning cluster.
```
rmweir, line 281: I suppose this is fine but if there is a provisioning cluster won't it hit it eventually anyway? I think just allowing that to happen would simplify the logic. I think it would also make it to where nextCluster could just be used in the beginning of the loop.

pkg/httpproxy/proxy.go
```diff
@@ -229,27 +234,96 @@ func (p *proxy) secretGetter(req *http.Request, cAuth string) SecretGetter {
...
+		c, err = p.mgmtClustersCache.Get(clusterID)
+		mgmtClusters = []*v3.Cluster{c}
+	}
+	if err != nil || len(mgmtClusters)+len(provClusters) == 0 {
 		return authorizer.DecisionDeny, err
 	}
-	if cluster.Spec.EKSConfig == nil {
-		return authorizer.DecisionDeny, nil
+
+	c = nextCluster(mgmtClusters, provClusters, &i)
```
rmweir, line 267: Is there a reason why there can't just be two for loops instead? I think this logic is likely more complicated than it needs to be so if it can be simplified, I am in favor.

pkg/controllers/managementuser/rbac/handler_base.go
```diff
@@ -428,7 +428,7 @@ type createFn func(objectMeta metav1.ObjectMeta, subjects []rbacv1.Subject, role
 type listFn func(ns string, selector labels.Selector) ([]interface{}, error)
 type convertFn func(i interface{}) (string, string, []rbacv1.Subject)
 
-func (m *manager) ensureBindings(ns string, roles map[string]*v3.RoleTemplate, binding interface{}, client *objectclient.ObjectClient,
+func (m *manager) ensureBindings(ns string, roles map[string]*v3.RoleTemplate, binding metav1.Object, client *objectclient.ObjectClient,
```
rmweir, line 431: Is this necessary? Otherwise it may be out of scope.

rmweir (changes requested):

pkg/api/norman/customization/aks/handler.go
```diff
@@ -255,28 +257,39 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
...
+		clusterSchema := h.schemas.Schema(&mgmtSchema.Version, client.ClusterType)
+		if err := h.ac.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", apiContext, map[string]interface{}{"id": c.Name}, clusterSchema); err != nil {
+			continue
 		}
-		return httperror.ServerError.Status, err
-	}
 
-	if cluster.Spec.AKSConfig.AzureCredentialSecret != cloudCredentialID {
-		return httperror.InvalidBodyContent.Status, fmt.Errorf("cloud credential not found")
+		if c.Spec.AKSConfig != nil && c.Spec.AKSConfig.AzureCredentialSecret == cloudCredentialID {
```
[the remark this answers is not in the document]
thedadams: If the `clusterID` is passed in then we check that specific cluster. We don't know if that cluster has the correct cloud credential on it.

pkg/api/norman/customization/gke/handler.go
```diff
@@ -190,28 +188,39 @@ func (h *handler) getCloudCredential(req *http.Request, cap *Capabilities, credI
...
+			continue
 		}
-		return httperror.ServerError.Status, err
-	}
 
-	if cluster.Spec.GKEConfig.GoogleCredentialSecret != cloudCredentialID {
-		return httperror.InvalidBodyContent.Status, fmt.Errorf("cloud credential not found")
+		if c.Spec.GKEConfig != nil && c.Spec.GKEConfig.GoogleCredentialSecret == cloudCredentialID {
+			return http.StatusOK, nil
+		}
```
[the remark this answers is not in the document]
thedadams: Commented on AKS.

pkg/api/norman/customization/oci/handler.go
```diff
@@ -134,22 +138,34 @@ func (handler *handler) ServeHTTP(writer http.ResponseWriter, req *http.Request)
...
+					var clusters []*prov.Cluster
+					clusters, err = handler.provClusterCache.GetByIndex(provcluster.ByCloudCred, credID)
+					if err != nil || len(clusters) == 0 {
+						return httperror.NotFound.Status, fmt.Errorf("cloud credential not found")
+					}
+					for _, cluster := range clusters {
+						if err = access.ByID(apiContext, &schema.Version, client.ClusterType, cluster.Status.ClusterName, &client.Cluster{}); err != nil || cluster.Spec.CloudCredentialSecretName != credID {
+							continue
+						}
+					}
```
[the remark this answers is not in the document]
thedadams: Yes, you're right. Fixed as you suggested.

pkg/api/norman/customization/vsphere/handler.go
```diff
@@ -145,14 +150,27 @@ func (v *handler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
...
+				var clusters []*prov.Cluster
+				clusters, err = v.provClusterCache.GetByIndex(provcluster.ByCloudCred, id)
+				if err != nil || len(clusters) == 0 {
+					return nil, httperror.NotFound, fmt.Errorf("cloud credential not found")
+				}
+				for _, cluster := range clusters {
+					if err = access.ByID(apiContext, &schema.Version, client.ClusterType, cluster.Status.ClusterName, &client.Cluster{}); err != nil || cluster.Spec.CloudCredentialSecretName != id {
+						continue
+					}
+				}
```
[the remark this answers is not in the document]
thedadams: Yes, you're right. Fixed as you suggested.

pkg/httpproxy/proxy.go
thedadams: The way is written it won't find the provisioning cluster if the `clusterID` is provided. I will update the logic so that this will no longer be the case.

pkg/httpproxy/proxy.go
thedadams: I will update this as you suggested to simplify the logic.

pkg/controllers/managementuser/rbac/handler_base.go
thedadams: See the second commit. There is an issue that you run into when running Rancher locally in Goland and giving owner access to a cluster to another user.

Without this change, the only way to stop Goland from pausing constantly is to recompile.

cmurphy (dismissed):

pkg/api/norman/customization/aks/handler.go
rmweir, line 287: I think these checks, for all the KEv2 providers, can happen prior to the CanDo checks. That way we can skip CanDo if this is the case.

rmweir (changes requested):
I do think that some of this logic could probably be one method and reused over multiple spaces but for time's sake it should be fine how it is. Just one other comment then should be good to go.

pkg/api/norman/customization/aks/handler.go
thedadams: I agree that some of this logic can be consolidated.

For now, I reordered the checks as you suggested.

rmweir (approved):
LGTM

cmurphy (approved):

Issue #33312, opened 2021-06-21
Update Ingress usages to support k8s 1.22


snasovich:
The code contains multiple references to `Ingress` from `extensions/v1beta1` / `networking.k8s.io/v1beta1` which are getting removed in k8s 1.22. Replacement API (`networking.k8s.io/v1`) is available in k8s 1.19+ and since Rancher v2.6 release should still support k8s 1.18, we cannot simply replace it outright.

Note: Support of other objects with removed APIs is https://github.com/rancher/rancher/issues/33222

Some sort of conditional logic will need to be implemented. Initial notes based on discussion with @ibuildthecloud :

> ibuildthecloud
> I'm guess you are going to have to do conditional logic and basically have two code paths.  You can create some wrapper object to abstract ingress so as to not repeat all the code.  Then ideally it would just be which controller is started.  One odd corner case is what do you do when a cluster upgrades from 1.18 to 1.22 as the available apis will dynamically change.
> 
> snasovich
> My understanding is that the existing objects are automagically “upgraded” when they are queried with new API version. Observed it with CRDs created as v1beta1 returned as v1 when querying them with `kubectl get ...`
> 
> ibuildthecloud  12:35 PM
> My point was more around the client logic here.  So the controllers will have to do "if v1.18; then ..."  and the issue is if that check is done once upfront what happens when the API version changes under the hood because of a cluster upgrade.

cbron:
@snasovich this is one of few that had issue already: https://github.com/rancher/rancher/issues/30053
In that one we said just to wait on it until 1.19 is gone, but we could also go after the kube api version conditional. Problem there in past has been that some API versions are non-standard.

snasovich:
@cbron , #30053 sounds specific to Helm charts while this is one is specifically for references in rancher/rancher code. It should be fairly straightforward to conditionally use one or other version of API in Helm charts based either on k8s versions or `.Capabilities.APIVersons.Has` as in https://github.com/rancher/rancher/pull/29922; not so much in Rancher code where is uses specific APIs when creating/querying Ingress objects.
Note the plan is for whoever is assigned to Rancher Helm charts to review and update not just Ingress, but any other APIs that are getting removed in k8s v1.22.

snasovich:
@cbron , clicked "Comment" a bit too early. We may not be able to wait for when we no longer need to support 1.18 as v2.6 (or one of v2.6.x) will need to support k8s from v1.18 to v1.22.
[...]

PR #35357, opened 2021-11-02
bump dependencies for k8s 1.22

kinarashah:
- [x] rancher/client-go v1.22.3-rancher.1 
- [x] rancher/webhook 
- [x] capi 0.4.4 
- [ ] \(Optional) rancher/helm - not blocking 
- [ ] \(Optional) rancher/shell - not blocking 
- [x] update k3s to 1.22.3
- [x] update etcd + grpc to handle `google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.42.0), but does not contain package google.golang.org/grpc/naming` 
- [x] RKE with support for etcd 3.5 and KDM 1.22.3
- [ ] \(Skipped for now) Fix failing Istio tests for 1.22 
- [ ] (Skipped for now) Fix failing tests `https://localhost:8443/v3/projects/local:p-6m86l/ingresses = "the server could not find the requested resource (get ingresses.meta.k8s.io)",` 

**Testing:** 
**Rancher Local: v1.22**
Downstream: 
- [x] RKE1 v1.22.3-rancher1 
- [x] RKE1 v1.21.x 
- [x] RKE1 v1.18.16 
- [x] Imported v1.22.3 
- [x] Imported v1.18.20
- [x] RKE2 v1.22.3

**Rancher Local: v1.18**
Downstream: 
- [x] RKE1 v1.22.3-rancher1 
- [x] RKE1 v1.21.x 
- [x] RKE1 v1.18.20
- [x] Imported v1.22.3 
- [x] Imported v1.18.20
- [ ] RKE2 v1.22.3

PR #35449, opened 2021-11-10
Add support for Ingress for k8s 1.22

cmurphy:
## Add compatibility helpers for ingress controllers

This change adds a package called ingresswrapper which provides
utilities that help make controllers able to use either the
extensions/v1beta1 Ingress API or the networking.k8s.io/v1 Ingress API,
depending on the version of the kubernetes API server the controllers
are running on. Kubernetes 1.18 only supports the extensions/v1beta1
Ingress API, while Kubernetes 1.22 only supports the
networking.k8s.io/v1 Ingress API. The compatibility layer defaults to
using the networking.k8s.io/v1 API when it is available.

The wrapper package provides two main helper objects that the
controllers can use. One is the Ingress interface, which is just a
controller-runtime client Object, which itself is a meta interface for
the apimachinery runtime Object and the meta/v1 Object. It allows direct
access to the object's TypeMeta and ObjectMeta data and can be used when
information from the object's Spec field is not needed. The other is the
CompatIngress struct, which inlines the networking.k8s.io/v1/Ingress
struct and provides methods to convert any Ingress object. This is
useful when the actual Spec of an Ingress object is needed. The data
structure is the same as for a networking.k8s.io/v1/Ingress object, so
when support for Kubernetes 1.18 is eventually removed, we should be
able to simply replace `CompatIngress` with `knetworkingv1.Ingress` in
every controller.

The wrapper also provides interfaces and wrapper methods to help with
setting up controllers.

~~## Update project.cattle.io ingress schema~~

~~This change updates the norman ingress store to use the
networking.k8s.io/v1 Ingress API for storing Ingress objects when
possible, and modifies the project.cattle.io schema for Ingress
resources to use the k8s.io/api/networking/v1 package. This also adds
norman mappers to copy fields where needed, such as from
"defaultBackend" to "backend", so that no matter which storage API is
used, the right data exists in the object. See the deprecation guide[1]
for notes on how the fields are translated.~~

~~[1] https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122~~

https://github.com/rancher/rancher/issues/33312

To do:

- [x] Add compatibility wrapper for norman clients (e.g. `IngressLister` and `IngressInterface`)
- [x] Update project.cattle.io [schema](https://github.com/rancher/rancher/blob/d77f618d20e52ec1788918a8cf0323099d896267/pkg/schemas/project.cattle.io/v3/schema.go#L702-L740)
- [x] Unit tests for ingresswrapper
- [x] Squash commits

Areas that need to be tested:

|                     | managementagent/ingress | managementlegacy/globaldns | managementuserlegacy/approuter | managementuserlegacy/endpoints | managementuserlegacy/globaldns | managementuserlegacy/ingresshostgen |
|---------------------|-------------------------|----------------------------|--------------------------------|--------------------------------|--------------------------------|-------------------------------------|
| Local 1.18 - docker |                         |                            |                                |                                |                                |                                     |
| Local 1.22 - docker |                         |                            |                                |                                |                                |                                     |
| Local 1.18 - helm   |                         |                            |                                |                                |                                |                                     |
| Local 1.22 - helm   |                         |                            |                                |                                |                                |                                     |
| Downstream 1.18     |                         |                            |                                |                                |                                |                                     |
| Downstream 1.22     |                         |                            |                                |                                |                                |                                     |


cmurphy:
I dropped the changes for the project.cattle.io schema, I found it wasn't working on 1.18. I will split it into a separate PR. The controller changes should still work without it.

The project.cattle.io changes should only affect the Ember UI. I'm having a hard time getting it right, and I think it might be safest to not make any changes at all and say that the Ember UI isn't supported for managing ingresses on kubernetes 1.22.

rmweir (changes requested):
Both comments apply to all applicable handlers. @cmurphy @git-ival are there any plans for performance testing this? I think we might want to look at api server metrics and see how much more time is being spent reading and writing ingress objects. Could also look for noticeable cpu spikes.

cmurphy:
I pushed up changes and now github isn't letting me reply directly to comments.

@rmweir

https://github.com/rancher/rancher/pull/35449#discussion_r749782981

> Could the logic contained in sync be put into a single function that accepts the new interface and then we can have two syncs, one for the v1 ingress type and one for the v1beta1 type? That way we don't need to use reflect.ValueOf? The other option may be to do the nil check in the function return by that helper functions CompatSyncV1 and CompatSyncv1Beta1.

`AddHandler` will fail unless the argument matches the controller type, i.e. `workload.Networking.Ingresses("").AddHandler` can only take a sync function whose object argument is type `*knetworkingv1.Ingress`. I also can't move the nil check to the helper function, because in some cases the controller is still supposed to run when the object is nil https://github.com/rancher/rancher/pull/35449/files#diff-cc012132def2a0d86b41b1ac06f83f0529ac56047437405c83a216db3497ce0cL23-L28

https://github.com/rancher/rancher/pull/35449#discussion_r749814127

> In situations like this where we know we are only reading from the object and this is the only operation we are doing with the Ingress object, can we just use compat in the parent function. Here we could use it in the sync handler and then pass it own into at least generateExpectedServices and calls of generateNewService.

I've added this here and a few places where it makes sense.

https://github.com/rancher/rancher/pull/35449#pullrequestreview-806674667

> are there any plans for performance testing this?

We haven't made specific plans yet. The conversion logic will inevitably have performance overhead. At least we'll be able to remove it once we drop support for 1.18.

kinarashah:
@rmweir can't merge changes since changes were requested, could you approve and discuss more with @cmurphy on the issue/another PR? we need this to unblock https://github.com/rancher/rancher/pull/35357

rmweir (approved):
Approved for merge with feedback to possibly be addressed after.

Issue #35445, opened 2021-11-10
Node Templates | Location and Size info is not present on UI (API call does not return them)


sgapanovich:
**Rancher Server Setup**
- Rancher version: v2.6-2708725597f29f23bea92828e34a1355085ad746-head
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details: self-signed

**Information about the Cluster**
- Kubernetes version: v1.21.5

**Describe the bug**

Location and Size for Node Templates are not present due to `GET v3/nodetemplates` not returning cloud config info such as "amazonec2Config"

**To Reproduce**
 
1. Log in as an admin
2. click on Main Menu (hamburger menu)
3. click Cluster Management 
4. expand RKE1 Configuration
5. click Node Templates
6. click Create
  provide with valid credentials and fill out RKE1 template (ec2)
7. Create a template 
8. Refresh the page

**Result**
Location and Size values are not present on UI because v3/nodetemplate call does not return cloud condig

**Expected Result**

Location and Size values are not present on UI

**Screenshots**
2.6-head vs 2.6.1

![26head](https://user-images.githubusercontent.com/40282141/141159930-f533afbd-743a-43ed-8080-8881e74ff912.jpg)

API call does not return location and size in 2.6-head

![ec2 no](https://user-images.githubusercontent.com/40282141/141160197-5ad19726-873e-463a-8f15-0a460ee57056.jpg)

this is how the API call returns location and size info in 2.6.1

![ec2 yes](https://user-images.githubusercontent.com/40282141/141160147-a298350d-0d85-4357-8945-68a502a04402.jpg)

**Additional context**

paynejacob:
the 2.6.1 API response appears identical to the 2.6-head response.  This is likely a dashboard issue.

paynejacob:
the 2.6-head response appears to be missing the amazonec2configuration section.

paynejacob:
### Root cause
The performance improvements to norman we causing some deserialization issues with nodeTemplates on list.
 
### What was fixed, or what changes have occurred
Fixed how norman serializes nodeTemplates on list.
 
### Areas or cases that should be tested
List node templates and ensure all fields in the ui are valid.  I would also compare the responses of `/v3/nodeTemplate` with 2.6-head to 2.6.1 and ensure they are the same structure.
 
### What areas could experience regressions?
-
 
[...]

PR #35451, opened 2021-11-10
[2.6.3] exclude resources that do not represent all their potential fields

paynejacob:
#35445
#35491

# Problem
Some resources in norman have values stored on them that are not part of go struct associated with their GVK.  When norman uses the appropriate go struct it will drop any field not on that go struct causing fields to be missing when listing resources. 

# Solution
Do not inform norman about the struct associated with some GVKs so that it will fall back to the unstructured object and allow the missing fields to be included.

# Testing
I confirmed with a digital ocean node template that the list objects matched the individual get object responses.

### DIFF
--- a/pkg/api/scheme/scheme.go
+++ b/pkg/api/scheme/scheme.go
@@ -1,6 +1,8 @@
 package scheme
 
 import (
+	"fmt"
+
 	istiov1alpha3 "github.com/knative/pkg/apis/istio/v1alpha3"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	monitoringv1alpha1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
@@ -29,11 +31,29 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationv1beta1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1beta1"
 )
 
-var Scheme *runtime.Scheme
+type FilteredScheme struct {
+	*runtime.Scheme
+	excludedGVKs map[schema.GroupVersionKind]struct{}
+}
+
+func (f *FilteredScheme) New(kind schema.GroupVersionKind) (runtime.Object, error) {
+	if _, ok := f.excludedGVKs[kind]; ok {
+		return nil, fmt.Errorf("%s is explicitly excluded from this scheme", kind.String())
+	}
+
+	return f.Scheme.New(kind)
+}
+
+func (f *FilteredScheme) ExcludeGVK(kind schema.GroupVersionKind) {
+	f.excludedGVKs[kind] = struct{}{}
+}
+
+var Scheme *FilteredScheme
 
 var builders = []runtime.SchemeBuilder{
 	ui.SchemeBuilder,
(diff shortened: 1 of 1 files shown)


pkg/api/scheme/scheme.go
```diff
@@ -29,11 +31,29 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationv1beta1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1beta1"
 )
 
-var Scheme *runtime.Scheme
+type FilteredScheme struct {
+	*runtime.Scheme
+	excludedGVKs map[schema.GroupVersionKind]bool
```
thedadams, line 41: nit: I am a fan of the following because it uses less memory:
```suggestion
	excludedGVKs map[schema.GroupVersionKind]struct{}
```

pkg/api/scheme/scheme.go
```diff
@@ -29,11 +31,29 @@ import (
...
+}
+
+func (f *FilteredScheme) New(kind schema.GroupVersionKind) (runtime.Object, error) {
+	if f.excludedGVKs[kind] {
```
thedadams, line 45: ...then this becomes:
```suggestion
	if _, ok := f.excludedGVKs[kind]; ok {
```

pkg/api/scheme/scheme.go
```diff
@@ -29,11 +31,29 @@ import (
...
+		return nil, fmt.Errorf("%s is explicitly excluded from this scheme", kind.String())
+	}
+
+	return f.Scheme.New(kind)
+}
+
+func (f *FilteredScheme) ExcludeGVK(kind schema.GroupVersionKind) {
+	f.excludedGVKs[kind] = true
```
thedadams, line 53: ...and this becomes:
```suggestion
	f.excludedGVKs[kind] = struct{}{}
```

pkg/api/scheme/scheme.go
```diff
@@ -29,11 +31,29 @@ import (
...
 
-var Scheme *runtime.Scheme
+type FilteredScheme struct {
+	*runtime.Scheme
+	excludedGVKs map[schema.GroupVersionKind]bool
+}
+
+func (f *FilteredScheme) New(kind schema.GroupVersionKind) (runtime.Object, error) {
+	if f.excludedGVKs[kind] {
+		return nil, fmt.Errorf("%s is explicitly excluded from this scheme", kind.String())
```
thedadams, line 46: Are there any side effects of returning an error here? Will the call be retried because an error is returned here?

pkg/api/scheme/scheme.go
```diff
@@ -69,10 +89,17 @@ var builders = []runtime.SchemeBuilder{
 }
 
 func init() {
-	Scheme = runtime.NewScheme()
-	if err := addKnownTypes(Scheme); err != nil {
+	Scheme = &FilteredScheme{
+		Scheme:       runtime.NewScheme(),
+		excludedGVKs: make(map[schema.GroupVersionKind]bool),
```
thedadams, line 94: ...oh and this would of course change:
```suggestion
		excludedGVKs: make(map[schema.GroupVersionKind]struct{}),
```

pkg/api/scheme/scheme.go
paynejacob: With how we use it we expect an error when something is not found.  I am just returning a more specific error incase someone is debugging.

pkg/api/scheme/scheme.go
paynejacob: TIL

thedadams (approved):

pkg/api/scheme/scheme.go
```diff
@@ -69,10 +89,17 @@ var builders = []runtime.SchemeBuilder{
...
+		Scheme:       runtime.NewScheme(),
+		excludedGVKs: make(map[schema.GroupVersionKind]struct{}),
+	}
+
+	if err := addKnownTypes(Scheme.Scheme); err != nil {
 		panic(err)
 	}
+
+	// Node templates return values from node drivers that are not referenced on the `NodeTemplate`
+	Scheme.ExcludeGVK(management.SchemeGroupVersion.WithKind("NodeTemplateList"))
```
rmweir, line 102: Are you sure this is the only one? What about kev1 clusters? You can create a kev1 cluster so you can test this by going to this path `<rancher-url>/g/clusters/add/launch/amazoneks`.

rmweir (changes requested):

pkg/api/scheme/scheme.go
snasovich: Alright, added `clusters.management.cattle.io` to the excluded GVKs and confirmed it addresses https://github.com/rancher/rancher/issues/35491. @rmweir , is that the same resource used by KEv1 or there is something else?

@paynejacob , is there any way you audit other types and see if we need to add more exclusions?

paynejacob: I confirmed I can create and edit eks clusters with that endpoint you gave @rmweir

snasovich (approved):

pkg/api/scheme/scheme.go
rmweir: @snasovich oh I think I misunderstood the question earlier, but yes what I'm speaking of is contained in clusters.management.cattle.io. @paynejacob great! Have you checked cloud credentials?

rmweir (changes requested):
Just question about cloud credentials, otherwise LGTM

pkg/api/scheme/scheme.go
paynejacob: cloud credentials are also good

rmweir (approved):
LGTM

thedadams (approved):

PR #35458, opened 2021-11-10
Patch default serviceAccount for new system namespaces

prachidamle:
Patch default serviceAccount for new namespaces that get added to the local cluster by Rancher while deploying a downstream cluster.

The rancher controller will also look at the namespaces having system-namespace annotation ("management.cattle.io/system-namespace") and also those added to the system project, and patch the default serviceAccounts for them so that 5.1.5 test is not broken.


ryansann (approved):
lgtm

StrongMonkey (approved):

kinarashah (approved):

PR #35479, opened 2021-11-12
Fix the issue of missing images in rancher-images.txt

niusmallnan:
If we want to print the array completely in the shell, we should use ${xxx[@]} instead of ${xxx}.

Example:
```
tt=( "1 2 3 4 5" )

echo ${tt}
# output: 1 2 3 4 5

echo ${tt[@]}
# output: 1 2 3 4 5
```
It does not seem to make a difference, however, if we do this:
```
tt=( `echo 1+ 2+ 3+ 4+ 5+ | tr + -` )

echo ${tt}
# output: 1-

echo ${tt[@]}
# output: 1- 2- 3- 4- 5-
```

This happens to cause some images to be missing, as shown in the figure:
截屏2021-11-12 上午11 55 47

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -83,7 +83,7 @@ fi
 
 if [ ${ARCH} == amd64 ]; then
     # Move this out of ARCH check for local dev on non-amd64 hardware.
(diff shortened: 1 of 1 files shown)


rmweir (approved):
LGTM

PennyScissors (approved):

PR #35473, opened 2021-11-11
[2.6.3] bump wrangler

paynejacob:
#34712

see https://github.com/rancher/wrangler/pull/183 for details

PR #35480, opened 2021-11-12
Update to New Signals Package in Wrangler

luthermonson:
Signals changed slightly to accommodate windows service manager being able to directly call a cancel of context. New package exposes a method to send a cancel to the channel and trigger a os.Interupt by hand.


paynejacob:
closing in favor of https://github.com/rancher/rancher/pull/35473

PR #35502, opened 2021-11-15
Add the ability to defer the start of user controllers

ibuildthecloud:
This PR lays the framework to allow user controllers to watch for
conditions before they start their caches/controllers. This is to allow
performance gains if the user is not using any legacy features.

Also included is the ability to limit which namespaces a controller is
watching which will be used by the impersonation logic in a follow on
PR.

https://github.com/rancher/rancher/issues/35503

### DIFF
--- a/pkg/types/config/context.go
+++ b/pkg/types/config/context.go
@@ -2,8 +2,11 @@ package config
 
 import (
 	"context"
+	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
An example of using this change can be seen in this commit: https://github.com/rancher/rancher/pull/35505/commits/930db4f000f08057fba681d5db6bcd9a40e8911f

Issue #32930, opened 2021-05-27
Replace go-version with a library that handles prereleases correctly in catalog rancher check


dramich:
ValidateRancherVersion needs to be updated to use Masterminds/semver (or another alternative library) as go-version does not follow proper semver when comparing pre-releases

jakefhyde:
### Root cause
 
The `go-version` library does not perform proper handling of prereleases. The ordering as per the semver spec is as follows:
```
1.0.0-0 < 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0
```
`go-version` appears to order `-0` before `-alpha`, i.e. numeric prerelease identifiers are given precedence over alphabetical.

### What was fixed, or what changes have occurred

The version range checking library to determine if catalogs are valid was replaced with [`blang/semver`](https://github.com/blang/semver).
 
### Areas or cases that should be tested

Installing and upgrading charts should list the exact same values as before.
 
### What areas could experience regressions?

Improperly formatted version ranges may cause specific charts to no longer show up. For example, a chart with the following values:
```
rancher_min_version: 2.6.4-alpha1
rancher_max_version: 2.6.5-0
```
for `server-version == v2.6.4` would validate with `go-version`, and fail with `blang`. 

Prerelease rancher versions should now actually work correctly, as before a chart with the following values:
```
rancher_min_version: 2.6.4-0
rancher_max_version: 2.6.4
```
for `server-version == v2.6.4-rc1` would have also failed, although we have not utilized a chart with such constraints.

To the best of my knowledge, we have no charts installable with `v2.6-head` that would be subject to either of these possible regressions, but it would be entirely possible for someone to misinterpret the semver spec and assume that `v2.6.5-0 >= v2.6.0`, especially since some libraries like `go-version` seem to reinforce this incorrect behavior.

### Are the repro steps accurate/minimal?

When launching `rancher/rancher:v2.6-head`, the environment variable `CATTLE_SERVER_VERSION=v2.6.3` will allow verification that the exact chart versions for `v2.6.3` are available, as otherwise they would all show up. Additionally, this environment variable can be used to set an `rc` version, or a custom tag. The list of available charts to install and upgrade should be the same.

sgapanovich:
### Test Environment: ###
Rancher version: v2.6-head 
Rancher cluster type: single-node docker install 
[...]

PR #35507, opened 2021-11-15
Swap go-version for blang/semver in catalog version validation

jakefhyde:
Replace `go-version` check with `blang/semver`. 

Elected to use `blang/semver` as opposed to `Masterminds/semver` because the former was already being used in the file, and does handle prerelease correctly (after removing v prefix). Also updated unit tests to include `-0` in min and max versions, as well as add tests for max/min only present.

Related Issue: https://github.com/rancher/rancher/issues/32930

### DIFF
--- a/pkg/catalog/manager/manager.go
+++ b/pkg/catalog/manager/manager.go
@@ -7,13 +7,11 @@ import (
 	"time"
 
 	"github.com/blang/semver"
-	mVersion "github.com/mcuadros/go-version"
 	"github.com/pkg/errors"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/catalog/utils"
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy/helm/common"
 	corev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
-	managementv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	projectv3 "github.com/rancher/rancher/pkg/generated/norman/project.cattle.io/v3"
 	helmlib "github.com/rancher/rancher/pkg/helm"
@@ -52,7 +50,7 @@ type CatalogManager interface {
 	GetSystemAppCatalogID(templateVersionID, clusterName string) (string, error)
 }
 
-func New(management managementv3.Interface, project projectv3.Interface, core corev1.Interface) *Manager {
+func New(management v3.Interface, project projectv3.Interface, core corev1.Interface) *Manager {
 	var bundledMode bool
 	if strings.ToLower(settings.SystemCatalog.Get()) == "bundled" {
 		bundledMode = true
@@ -107,7 +105,7 @@ func (m *Manager) DeleteOldTemplateContent() bool {
 }
 
 func (m *Manager) DeleteBadCatalogTemplates() bool {
-	// Orphaned catalog templates and template versions may exist, remove any where the catalog does not exist
+	// Orphaned catalog templates and template versions may exist, remove anywhere the catalog does not exist
 	errs := m.deleteBadCatalogTemplates()
 	if len(errs) == 0 {
 		return true
@@ -249,8 +247,6 @@ func (m *Manager) ValidateRancherVersion(template *v3.CatalogTemplateVersion, cu
 		// be edited as long as it is not being upgraded/rollbacked to another incompatible version.
 		return nil
 	}
-	rancherMin := template.Spec.RancherMinVersion
-	rancherMax := template.Spec.RancherMaxVersion
 
 	serverVersion := settings.ServerVersion.Get()
 
@@ -259,14 +255,31 @@ func (m *Manager) ValidateRancherVersion(template *v3.CatalogTemplateVersion, cu
 		return nil
 	}
 
-	if rancherMin != "" && !mVersion.Compare(serverVersion, rancherMin, ">=") {
-		return fmt.Errorf("rancher min version not met")
+	serverVersion = strings.TrimPrefix(serverVersion, "v")
+
+	versionRange := ""
+	if template.Spec.RancherMinVersion != "" {
+		versionRange += " >=" + strings.TrimPrefix(template.Spec.RancherMinVersion, "v")
+	}
+	if template.Spec.RancherMaxVersion != "" {
+		versionRange += " <=" + strings.TrimPrefix(template.Spec.RancherMaxVersion, "v")
+	}
+	if versionRange == "" {
+		return nil
+	}
+	constraint, err := semver.ParseRange(versionRange)
+	if err != nil {
+		logrus.Errorf("failed to parse constraint for rancher version %s: %v", versionRange, err)
+		return nil
 	}
 
-	if rancherMax != "" && !mVersion.Compare(serverVersion, rancherMax, "<=") {
-		return fmt.Errorf("rancher max version exceeded")
+	rancherVersion, err := semver.Parse(serverVersion)
+	if err != nil {
+		return err
 	}
-
+	if !constraint(rancherVersion) {
+		return fmt.Errorf("incompatible rancher version [%s] for template [%s]", serverVersion, template.Name)
+	}
 	return nil
 }
 
--- a/pkg/catalog/manager/manager_test.go
+++ b/pkg/catalog/manager/manager_test.go
@@ -53,17 +53,152 @@ func TestLatestAvailableTemplateVersion(t *testing.T) {
 		},
 	}
 
+	templateWithoutMinRancherVersion := &v3.CatalogTemplate{
+		Spec: v3.TemplateSpec{
+			Versions: []v3.TemplateVersionSpec{
+				{
+					ExternalID:        "catalog://?catalog=library&template=artifactory-ha&version=0.12.16",
+					Version:           "0.12.16",
+					RancherMaxVersion: "v2.3.0",
+				},
+				{
+					ExternalID:        "catalog://?catalog=library&template=artifactory-ha&version=0.12.15",
+					Version:           "0.12.15",
+					RancherMaxVersion: "v2.2.0",
+				},
+				{
+					ExternalID:        "catalog://?catalog=library&template=artifactory-ha&version=0.12.14",
+					Version:           "0.12.14",
+					RancherMaxVersion: "v2.1.0",
+				},
+			},
+		},
+	}
+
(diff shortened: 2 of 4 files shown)


paynejacob (dismissed):
as long as we are not introducing another semver library.

paynejacob (approved):

pkg/catalog/manager/manager.go
```diff
@@ -243,24 +241,38 @@ func (m *Manager) ValidateKubeVersion(template *v3.CatalogTemplateVersion, clust
...
+	}
+	constraint, err := semver.ParseRange(versionRange)
+	if err != nil {
+		logrus.Errorf("failed to parse constraint for rancher version %s: %v", versionRange, err)
+		return nil
 	}
 
+	rancherVersion, err := semver.Parse(serverVersion)
+	if err != nil {
+		return err
```
rmweir, line 271: Seems odd to treat these differently, I think it may make sense to log and swallow both since it is probably more annoying to not be able to install a catalog you should be able to install then it is bad that you can install a catalog that you shouldn't be able to install. I do suppose an invalid upgrade could be kind of bad though. @snasovich thoughts?

pkg/catalog/manager/manager_test.go
```diff
@@ -15,20 +15,20 @@ func TestLatestAvailableTemplateVersion(t *testing.T) {
 				{
 					ExternalID:        "catalog://?catalog=library&template=artifactory-ha&version=0.12.16",
 					Version:           "0.12.16",
-					RancherMinVersion: "v2.2.0",
-					RancherMaxVersion: "v2.3.0",
+					RancherMinVersion: "v2.2.0-0",
```
rmweir, line 18: Can you explain what the "-0" is for? I don't think the catalogs we use would include this in the constraints and I don't see you adding it to the versions in ValidateRancherVersion.

tests/integration/suite/test_app.py
```diff
@@ -377,7 +376,6 @@ def test_app_create_validation(admin_mc, admin_pc, custom_catalog,
         app1 = admin_pc.client.create_app(app_data)
         remove_resource(app1)
     assert e.value.error.status == 422
-    assert e.value.error.message == 'rancher max version exceeded'
```
rmweir, line 380: Why remove these? Are there not new error messages that can be checked?

rmweir (changes requested):

pkg/catalog/manager/manager.go
jakefhyde: This was how it is being done for the kubernetes version, but tI think that makes sense since the only way it will fail to parse is if the server version has been manually set.

tests/integration/suite/test_app.py
jakefhyde: Good catch. I've decided to add checks for part of the error message, since formatting the error client side was sort of cumbersome

pkg/catalog/manager/manager_test.go
jakefhyde: Updated, do not need it anymore.

pkg/catalog/manager/manager.go
snasovich: Yes, it matches the following checks so I thought this inconsistency is OK when I reviewed: https://github.com/rancher/rancher/blob/e013338b5eb1271ced81b7ff7ebae093f664d7dc/pkg/catalog/manager/manager.go#L218-L236

I could see some logic in it - if there is the first error, it's an issue with constraints on `CatalogTemplateVersion` so we err on the side of "let's allow to install it" while the second error is when Rancher version is messed up and we block installs against it.

Since I don't know the history of why error handling is inconsistent and there is not a comment that explains it I find it safer to follow the same logic here too.

paynejacob (approved):

snasovich (commented):
I swear I submitted my comment yesterday but now I see it shows as Pending. Maybe my comment is something we can huddle about with @rmweir @paynejacob.

pkg/catalog/manager/manager.go
```diff
@@ -243,24 +241,38 @@ func (m *Manager) ValidateKubeVersion(template *v3.CatalogTemplateVersion, clust
...
+	serverVersion = strings.TrimPrefix(serverVersion, "v")
 
-	if rancherMax != "" && !mVersion.Compare(serverVersion, rancherMax, "<=") {
-		return fmt.Errorf("rancher max version exceeded")
+	versionRange := ""
+	if template.Spec.RancherMinVersion != "" {
+		versionRange += " >=" + strings.TrimPrefix(template.Spec.RancherMinVersion, "v")
+	}
+	if template.Spec.RancherMaxVersion != "" {
+		versionRange += " <=" + strings.TrimPrefix(template.Spec.RancherMaxVersion, "v")
```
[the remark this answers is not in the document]
jakefhyde: Some use `x.y.z-alpha1` and `x.y.99-0` together, like the following:
https://github.com/rancher/system-charts/blob/5a757daa2a82ab4ffeedadbf1ad68d9ec1661524/charts/rancher-logging/0.2.5/questions.yml#L1-L2
whereas others use `x.y.z-rc1` and `x.y.z`, like so:
https://github.com/rancher/system-charts/blob/9d3c12a9704b5501baa87193a07fb9f953c465af/charts/rancher-monitoring/v0.2.1/questions.yml#L1-L2
I think regardless it's safe to say we need more unit and integration tests covering this, because the integration tests at https://github.com/rancher/integration-test-charts only use `x.y.z` for both min and max version. This is also vastly different from the actual catalog charts, as they all use `x.y.z-(alpha/rc)1` (except for one [dynatrace chart](https://github.com/rancher/helm3-charts/blob/136a0a384c31a88393bdc56770ba8c848aa2eac9/charts/dynatrace-operator/v0.2.1/questions.yml#L4)).

Also to answer you're question, yes, blang has no such restriction on prereleases, as the unit tests (not the integration tests) would fail. This is achievable with https://github.com/Masterminds/semver, but would require adding prereleases to the max and min version that do not have it. 

A real concrete example is the following:
```
$ vert ">=2.5.6-rc1,<=2.5.7" 2.5.6-rc1
$ echo $?
1
```
Which comes from the monitoring chart listed above. This works as is right now with blang/semver, which I can only assume is what we want otherwise most of the catalogs would disappear for an rc, but it also works with go-version, so I guess the question is are we going to update all of the charts with min and max version to add `-0` (either in the charts or in the validation check) if there is no prerelease for the given min/max?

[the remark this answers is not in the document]
snasovich: @jakefhyde , thank you for detailed explanation and research on what ranges our charts use.
I've played with previous github.com/mcuadros/go-version based code and new github.com/blang/semver based one and found one mismatch, but I think it's actually correct.
Basically in go-version 2.5.8 or 2.5.8-rc1 fits 2.5.7-alpha1 to 2.5.8-0 range while it doesn't in blang's case - https://play.golang.org/p/hPo4E3ZEbnL. The latter is more correct as the intention of <=2.5.8-0 is to essentially make the upper range not include anything from 2.5.8 release line.

Bottom line I think we're fine as far as the actual code, but it would be nice to update at least unit tests to use realistic ranges we're using rather than ambiguous ranges as we use here https://github.com/jakefhyde/rancher/blob/833a37a62556c448c3599573b107696ea2033e52/pkg/catalog/manager/manager_test.go#L15-L32 as for example 2.1.0 fits 2 ranges.

I know it's a bit beyond what the issue is immediately about, but we want to do the right thing here and have unit tests represent the realistic ranges we use.

pkg/catalog/manager/manager_test.go
```diff
@@ -53,17 +53,172 @@ func TestLatestAvailableTemplateVersion(t *testing.T) {
...
+	testLatestAvailableTemplateVersion(t, "v2.0.1", "0.12.14", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "v2.0.2-beta", "0.12.14", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "v2.1.0", "0.12.15", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "v2.2.5", "0.12.16", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "dev", "0.12.16", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "master", "0.12.16", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "master-head", "0.12.16", templateWithoutMaxRancherVersion)
+	testLatestAvailableTemplateVersion(t, "", "0.12.16", templateWithoutMaxRancherVersion)
+
+	testLatestAvailableTemplateVersion(t, "v2.2.0-alpha1", "0.12.16", templateWithMinPrerelease)
```
snasovich, line 208: There is no test case here that expects 0.12.15. However, it doesn't seem possible as the range for 0.12.16 (`[v2.2.0-alpha1, v2.3.0]`) fully includes the range for 0.12.15 (`[v2.2.0-alpha1, v2.2.0]`) so 0.12.15 won't ever be used.

Now, these ranges don't really match what we're generally using per the examples you provided in your comment https://github.com/rancher/rancher/pull/35507/files#r751808450.

`templateWith99MaxPatchVersion` does match one of these examples, but neither `templateWithMinPrerelease` nor `templateWithMinAndMaxPrerelease` match "x.y.z-rc1 to x.y.(z+1)" example. So I would recommend replacing these 2 sets with new set that has ranges like `[2.2.0-rc1, 2.2.1]` and `[2.2.1-rc1, 2.2.2]` and have a few test cases around these versions.

snasovich (changes requested):
I think unit tests need a bit more work per the comment I have submitted.

pkg/catalog/manager/manager_test.go
jakefhyde: Done, also added more tests to the 99 version to highlight the case for ranges like `>=v2.2.0-alpha1,<=v2.2.99` being followed up by `>=v2.3.0-alpha1,<=v2.3.99` (`v2.3.0-0` should not be accepted by either).

snasovich (approved):
Updated tests LGTM, but we may still want to hold off merging it for 2.6.3 and may want to wait till 2.6.4.
Asked @cbron 's input on this.

paynejacob (dismissed):

pkg/catalog/manager/manager.go
rmweir: @snasovich yeah that makes sense.

rmweir (approved):
LGTM

snasovich:
Decided not to merge it for 2.6.3, so converted to draft to prevent accidental merging.

paynejacob (approved):

snasovich (approved):

Issue #35488, opened 2021-11-12
[Backport] 2.5x - Convert ServerVersion to downstream env var


Jono-SUSE-Rancher:
We need to create a backport version of https://github.com/rancher/rancher/issues/34875 to be included in v2.5x. Please let me know if there are any questions.

jakefhyde:
Do we want to backport #34876 instead? It is the successor to #34875, and redeploys the agent with the settings as environment variables on change, as opposed to setting the settings in the downstream.

Jono-SUSE-Rancher:
Yes that's fine Jake - I didn't know of it's existence and was just backporting the original one so that we'd have it in place. If you think that is the correct one and it will work for what we need, let's do that. 
ccing: @PennyScissors to keep me honest.

jakefhyde:
Issue Template: https://github.com/rancher/rancher/issues/34876#issuecomment-969195463

jmcsagdc:
v2.5-777b5d4d5cffa482e8d348d003d76b22f71756b4-head

In a downstream cluster
1) Use API to update install-uuid to change the uuid a little in a conforming way: 
   Like so "value": "2b18f3a8-1b63-49e6-bf2e-3b7052e29c1b" becomes "value": "3c28f3a8-1b63-49e6-bf2e-3b7052e29c1b"
RESULT:
cattle-cluster-agent pod is redeployed

2) Deploy something like nginx-ingress
    Use API to update ingres-ip-domain (to nip.io)
RESULT:
cattle-cluster-agent pod is redeployed

3) Updating other settings did not cause a redeployment so far, but I'm checking a few more things

jakefhyde:
@jmcsagdc By "Updating other settings", does that include the `server-version`? If so that may be alright, as in the "happy path" the `server-version` is only going to be upgraded through an actual release which should bump the agent image tag and trigger a redeployment anyway.

jmcsagdc:
@jakefhyde as demonstrated, the URL change does not propagate through to the UI even though the API change sticks and the pods redeploy. 

The practical result of this is going *from nip.io* to xip.io or something doesn't actually affect the workload and the original URL continues to be routable while the new URL is not.

![URL-change-redeploy-but-no-update-ui](https://user-images.githubusercontent.com/15971368/147969605-d7e7720f-9f38-43c6-9cc4-76beef1a503c.jpg)

The other things tried here include the labels and annotations in the ingress-ip-domain which did not result in a redeployment.

[...]

PR #35508, opened 2021-11-15
[2.5] convert sync handler to env vars

jakefhyde:
Backport of https://github.com/rancher/rancher/pull/34982 and https://github.com/rancher/rancher/pull/34846

Summary
-
- Downstream clusters need to have the same server-version setting as their upstream in order to view the correct feature charts version for the current rancher installation. The CATTLE_SERVER_VERSION environment variable is now passed down through an environment variable when the cluster agent is deployed to ensure that the downstream cluster agent uses the correct version.

- Moved the install-uuid and ingress-ip-domain settings from the settings sync-handler to cluster-agent deployment template. Moved settings sync handler from managementuser to management and redeploy agent on settings changed.

Related Issue: https://github.com/rancher/rancher/issues/35488


rmweir (approved):
LGTM

Issue #35496, opened 2021-11-13
RKE1 Downstream 1.22 stuck in provisioning


kinarashah:
Create Rancher local cluster 1.22 and Provision a RKE1 Downstream cluster (any provider): 

Cluster is stuck in provisioning, with error msg: 
`2021/11/13 11:20:51 [ERROR] error syncing 'c-25m7f': handler cluster-provisioner-controller: error creating admin role: the server could not find the requested resource, requeuing`
<img width="1424" alt="Screen Shot 2021-11-13 at 11 25 50 AM" src="https://user-images.githubusercontent.com/16870565/141656620-1f04369a-0e65-4b16-b838-668a2d31b1f0.png">

https://github.com/rancher/rancher/issues/34054

kinarashah:
- Available to test with v2.6-head, make sure it's an image after this build https://drone-publish.rancher.io/rancher/rancher/6862. 
- Update rke-metadata-config (Global settings -> rke-metadata-config) URL's branch from `dev-v2.6` to `dev-v2.6-122`
- Make sure `v1.22.3-rancher1-1` shows up as experimental when provisioning RKE1 clusters 
- Normal RKE1 provisioning tests for v1.22.3-rancher1-1

kinarashah:
https://github.com/rancher/rancher/commit/01053b646f5def7e7cadd11b7c5dd9e6d273227d

sowmyav27:
- On `v2.6-f487c8819d726109d8178894b6a4c082d5320b92-head` commit, deployed EC2 node driver and custom clusters using 1.22.3-rancher1-1 k8s version
- Cluster has come up Active. 
- `1.22.3-rancher1-1` does show up as Experimental and is not the default k8s version yet.

PR #35516, opened 2021-11-15
update rbac v1beta1 to v1

kinarashah:
https://github.com/rancher/rancher/issues/35496

### DIFF
--- a/pkg/kontainer-engine/drivers/util/utils.go
+++ b/pkg/kontainer-engine/drivers/util/utils.go
@@ -9,7 +9,7 @@ import (
 	v3 "github.com/rancher/rke/types"
 	"gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/api/rbac/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -47,11 +47,11 @@ func GenerateServiceAccountToken(clientset kubernetes.Interface) (string, error)
 		return "", fmt.Errorf("error creating service account: %v", err)
 	}
 
-	adminRole := &v1beta1.ClusterRole{
+	adminRole := &rbacv1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: clusterAdmin,
 		},
-		Rules: []v1beta1.PolicyRule{
+		Rules: []rbacv1.PolicyRule{
 			{
 				APIGroups: []string{"*"},
 				Resources: []string{"*"},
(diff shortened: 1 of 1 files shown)


pkg/kontainer-engine/drivers/util/utils.go
```diff
@@ -9,6 +9,7 @@ import (
 	v3 "github.com/rancher/rke/types"
 	"gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
```
prachidamle, line 12: Do we need to keep both rbac/v1beta1 and rbac/v1 to support multiple k8s versions?

kinarashah: I could be missing something, but I see both in 1.18 (the lowest version we need to support) - 

v1: https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/api/rbac/v1/types.go 
v1beta1: https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/api/rbac/v1beta1/types.go

ryansann (dismissed):

pkg/kontainer-engine/drivers/util/utils.go
snasovich: JFYI here is a table I've compiled a while ago with old and replacement APIs and k8s versions they are available starting from ([internal SUSE link](https://confluence.suse.com/display/EN/Kubernetes+1.22+Checklist#Kubernetes1.22Checklist-k8s-1.22-removals)).

prachidamle: Both is supported, but I wonder if we can just update everything to v1 if there is nothing tied to v1beta1 strictly? Or maybe I am missing something.

https://github.com/rancher/rancher/blob/f2b697bbc860213fc7720aa106fc6b0b6ecc33d2/pkg/kontainer-engine/drivers/util/utils.go#L129-L131
https://github.com/rancher/rancher/blob/f2b697bbc860213fc7720aa106fc6b0b6ecc33d2/pkg/kontainer-engine/drivers/util/utils.go#L90

snasovich: @prachidamle , ah, misunderstood your original comment. Yes, I think we want to update those occurrences you found to v1 too and get rid of references to `k8s.io/api/rbac/v1beta1` altogether.

kinarashah: Oh! yep I misunderstood too, I thought the question was if updating it to all to `v1` was safe. Will update everything to rbac v1

prachidamle: cool, thanks

snasovich (approved):

prachidamle (approved):

kinarashah:
Closing this because I pulled the commit in an existing PR and  and didn't want to wait on multiple drone builds to pass-https://github.com/rancher/rancher/pull/35357/commits/01053b646f5def7e7cadd11b7c5dd9e6d273227d. 

It's the exact change that was approved here, no net new changes.

PR #35520, opened 2021-11-16
Only start clusterauthtoken controller if ACE is enabled

ibuildthecloud:
The clusterauthtoken controller is only started if v3.Cluster.Spec.LocalClusterAuthEndpoint.Enabled is true.  This controller is responsible for syncing tokens to the downstream clusters for ACE to work properly


rmweir (approved):
LGTM

PR #35521, opened 2021-11-16
Only start networkpolicy user controller if EnabledNetworkPolicy is set

ibuildthecloud:
Network policy controller is only started if v3.Cluster.Spec.EnableNetworkPolicy is true.  The controller is responsible for programing Project Network Policies.


rmweir (approved):
LGTM

PR #35522, opened 2021-11-16
Only enable alert user controller is used

ibuildthecloud:
If the Legacy feature flag is enabled or a ClusterAlert or a Notifier exists then the alert user controller will be started. This controller is responsible for alerts from monitoring v1.


rmweir (approved):
LGTM

PR #35523, opened 2021-11-16
Defer start of approuter user controller

ibuildthecloud:
Only start the approuter user controller if configured to a different API
endpoint.  This feature does not seem to be used anymore as the default
setting points to a service that doesn't exist (https://api.lb.rancher.cloud/v1).
If the setting points to the default we know it doesn't work so only
start the user controller if it's set to some other non-empty string.

The approuter controller will call RancherDNS service and get a hostname
for is the hostname of ingress rule is set to lb.rancher.cloud and the
ingress-ip-domain setting is set to lb.rancher.cloud. Since there is
no know working version of the RancherDNS service the only way you can
test would be to set the setting rdns-base-url to something like
http://example.com and log for errors in rancer like

"udpate fqdn [%s] to server [%s] error"

### DIFF
--- a/pkg/controllers/managementuserlegacy/approuter/approuter.go
+++ b/pkg/controllers/managementuserlegacy/approuter/approuter.go
@@ -3,8 +3,15 @@ package approuter
 import (
 	"context"
 
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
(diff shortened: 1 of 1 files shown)


rmweir (approved):
LGTM

ibuildthecloud:
Had to rebase

rmweir (approved):

PR #35524, opened 2021-11-16
Only start cis v1 user controller if a ClusterScan exists

ibuildthecloud:
Only start CIS v1 controller if a v3.ClusterScan exists.  The CIS controller is responsible for running CIS benchmark scan using v1 framework.


rmweir (approved):
LGTM

PR #35525, opened 2021-11-16
Only start globaldns user controller if GlobalDns object exists

ibuildthecloud:
Only start the globaldns controller if a v3.GlobalDns exists that points to this cluster.  The globaldns controllers is responsible for updating the v3.GlobalDns.Status.Endpoints field for the specific downstream cluster.


rmweir (approved):
LGTM

PR #35526, opened 2021-11-16
Defer start of pspdelete user controller

ibuildthecloud:
Only start pspdelete user controller if DefaultPodSecurityPolicyTemplateName is set on the v3.Cluster and its RKE2 managed cluster.  The pspdelete controller is used specifically for RKE2 to PSP with "psp.rke2.io/global-unrestricted" and "psp.rke2.io/global-restricted" annotations that are created by default for RKE2.

### DIFF
--- a/pkg/controllers/managementuser/pspdelete/pspdelete.go
+++ b/pkg/controllers/managementuser/pspdelete/pspdelete.go
@@ -3,6 +3,7 @@ package pspdelete
 import (
 	"context"
 
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/cluster"
 	provisioningcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	v1beta12 "github.com/rancher/rancher/pkg/generated/norman/policy/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/controllers/managementuser/pspdelete/pspdelete.go
```diff
@@ -22,7 +23,36 @@ type handler struct {
...
+	})
+
+	clusters := userContext.Management.Management.Clusters("")
+	clusterCache := userContext.Management.Wrangler.Provisioning.Cluster().Cache()
+	clusters.AddHandler(ctx, "pspdelete-deferred", func(key string, obj *v3.Cluster) (runtime.Object, error) {
+		if obj != nil &&
+			obj.Name == userContext.ClusterName &&
+			obj.Spec.DefaultPodSecurityPolicyTemplateName != "" {
+
+		}
```
rmweir, line 39: empty block here, guessing you meant to return

pkg/controllers/managementuser/pspdelete/pspdelete.go
```diff
@@ -22,7 +23,36 @@ type handler struct {
...
+		clusters, err := clusterCache.GetByIndex(cluster.ByCluster, userContext.ClusterName)
```
rmweir, line 40: why use indexer instead of just getting it off the cache regularly?

rmweir (changes requested):

pkg/controllers/managementuser/pspdelete/pspdelete.go
rmweir: Answer offline, resolved.

pkg/controllers/managementuser/pspdelete/pspdelete.go
```diff
@@ -22,7 +23,37 @@ type handler struct {
...
+			return obj, nil
+		}
+
+		clusters, err := clusterCache.GetByIndex(cluster.ByCluster, obj.Name)
+		if err != nil || len(clusters) != 1 {
+			return obj, err
+		}
+
+		clusterSpec := clusters[0].Spec
+		if clusterSpec.DefaultPodSecurityPolicyTemplateName != "" && clusters[0].Spec.RKEConfig != nil {
```
rmweir, line 47: nit:
```suggestion
		if clusterSpec.DefaultPodSecurityPolicyTemplateName != "" && clustersSpec.RKEConfig != nil {
```

rmweir (dismissed):
one nit, then LGTM

deniseschannon (approved):

deniseschannon (approved):

PR #35528, opened 2021-11-16
Update project.cattle.io ingress schema

cmurphy:
This change updates the norman ingress store to use the
networking.k8s.io/v1 Ingress API for storing Ingress objects when
possible, and modifies the project.cattle.io schema for Ingress
resources to use the k8s.io/api/networking/v1 package. This also adds
norman mappers to copy fields where needed, such as from
"defaultBackend" to "backend", so that no matter which storage API is
used, the right data exists in the object. See the deprecation guide[1]
for notes on how the fields are translated.

[1] https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

To make this able to work with both 1.18 and 1.22 I needed to add fields to the schema, so the schema now appears to have duplicate alias fields:

image

If everything is working correctly, then the Ember UI and terraform or other clients should be able to work without any changes, it's only when you look at the schema itself that it looks different.

https://github.com/rancher/rancher/issues/33312

### DIFF
--- a/pkg/schemas/mapper/ingress_backend.go
+++ b/pkg/schemas/mapper/ingress_backend.go
@@ -1,0 +1,106 @@
+package mapper
+
+import (
+	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/values"
+)
+
+// These mappers copy data from the networking.k8s.io/v1-style Ingress fields
+// to extensions/v1beta1-style Ingress fields so that when the proxy store
+// serializes them into kubernetes resources, there is no loss of data and the
+// objects are in the expected format for the API in use. The end result is
+// that some data is duplicated on the object, but extraneous fields will be
+// ignored by the proxy store.
+
+// IngressSpec mapper copies defaultBackend (type k8s.io/api/networking/v1.IngressBackend)
+// to backend (type k8s.io/api/extensions/v1beta1.IngressBackend) on Spec.
+type IngressSpec struct{}
+
+func (i IngressSpec) FromInternal(data map[string]interface{}) {
+	if _, ok := data["backend"]; ok && data["defaultBackend"] == nil {
+		data["defaultBackend"] = map[string]interface{}{
+			"targetPort": values.GetValueN(data, "backend", "servicePort"),
+			"serviceId":  values.GetValueN(data, "backend", "serviceName"),
+		}
+		delete(data, "backend")
+	}
+	return
+}
+
+func (i IngressSpec) ToInternal(data map[string]interface{}) error {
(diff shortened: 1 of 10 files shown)


pkg/schemas/mapper/ingress_backend.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return
+}
+
+func (i IngressSpec) ToInternal(data map[string]interface{}) error {
+	if backend, ok := data["backend"]; (!ok || backend == nil) && data["defaultBackend"] != nil {
+		data["backend"] = map[string]interface{}{
+			"servicePort": values.GetValueN(data, "defaultBackend", "targetPort"),
+			"serviceName": values.GetValueN(data, "defaultBackend", "serviceId"),
+		}
```
rmweir, line 26: should defaultBackend be deleted?

pkg/schemas/mapper/ingress_backend.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+func (i IngressBackend) ToInternal(data map[string]interface{}) error {
+	if data != nil {
+		data["servicePort"] = values.GetValueN(data, "targetPort")
+		data["serviceName"] = values.GetValueN(data, "serviceId")
+	}
+	return nil
+}
```
rmweir, line 55: Should the fields the value is extracted from be deleted for any of these?

rmweir (changes requested):

pkg/schemas/mapper/ingress_backend.go
cmurphy: No, I'm intentionally duplicating all these fields. This way when the object is serialized to either a v1 Ingress or a v1beta1 Ingress it has the right attributes.

cmurphy:
Noticed that this wasn't working for services using port names instead of numbers, updated.

pkg/schemas/mapper/ingress_backend.go
cmurphy: I'm open to feedback on a better way to do this, I know this is extremely unpretty.

kinarashah: I think we should comment context explaining the reasoning on why we are duplicating fields. This is easier to infer in other places because of conditional wrapper `ServerSupportsIngressV1` but harder to understand in mapper. Also would help in future if someone else tries to update/remove these.

cmurphy: done

kinarashah (approved):

snasovich:
@rmweir , have your requested changes been addressed? If so, please re-review.

rmweir (approved):
LGTM

PR #35530, opened 2021-11-16
Only start helm controller if v3.App exists

ibuildthecloud:
helm controller is started only if a v3.App for this cluster exists.  The helm controller will install apps from catalog v1.


rmweir (approved):
LGTM

PR #35532, opened 2021-11-16
Only start logging user controller if Cluster/ProjectLogging exists

ibuildthecloud:
The logging controller is responsible for instantiating logging v1 in the downstream clusters.

### DIFF
--- a/pkg/controllers/managementuserlegacy/logging/controller.go
+++ b/pkg/controllers/managementuserlegacy/logging/controller.go
@@ -3,15 +3,36 @@ package logging
 import (
 	"context"
 
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy/logging/configsyncer"
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy/logging/deployer"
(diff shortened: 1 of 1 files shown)


pkg/controllers/managementuserlegacy/logging/controller.go
```diff
@@ -3,15 +3,36 @@ package logging
...
+}
+
+func AddStarter(ctx context.Context, cluster *config.UserContext, starter func() error) {
+	clusterLogging := cluster.Management.Management.ClusterLoggings("")
+	projectLogging := cluster.Management.Management.ProjectLoggings("")
+	clusterLogging.AddClusterScopedHandler(ctx, "logging-deferred", cluster.ClusterName, func(key string, obj *v3.ClusterLogging) (runtime.Object, error) {
+		return obj, starter()
+	})
+	projectLogging.AddClusterScopedHandler(ctx, "logging-deferred", cluster.ClusterName, func(key string, obj *v3.ProjectLogging) (runtime.Object, error) {
+		return obj, starter()
```
rmweir, line 31: Shouldn't these handlers have different names?

rmweir (changes requested):

pkg/controllers/managementuserlegacy/logging/controller.go
rmweir: Resolved offline.

rmweir (approved):
LGTM

PR #35533, opened 2021-11-16
Defer start of monitoring user controller

ibuildthecloud:
Only start monitoring user controller if EnableClusterMonitor on v3.Cluster is true or EnableProjectMonitoring on v3.Project is true. This controller runs the monitoring v1 components.


rmweir (approved):
LGTM

PR #35534, opened 2021-11-16
Only start pipeline user controller if v3.Pipeline exists

ibuildthecloud:
Only start the pipeline controller if a v3.Pipeline exists for the cluster.  The pipeline controller runs all Pipeline functionality in the downstream cluster.


rmweir (approved):
LGTM

PR #35535, opened 2021-11-16
Only start resourcequota controller if quotas are set on project

ibuildthecloud:
Only start the resourcequota controller if ContainerDefaultResourceLimit, ResourceQuota, or NamespaceDefaultResourceQuota is set on a project.  The resourcequota controller is used for programming Project resource quotas


rmweir (approved):
LGTM

PR #35536, opened 2021-11-16
Only start secret controller if more that one secret exists in project ns

ibuildthecloud:
Only start secret controller if a secret is found in the project namespace in the management cluster that doesn't start with "default-token-".  The secret controller is what populates projects secrets that are available to all namespaces in a project.


rmweir (approved):
LGTM, with one nit

rmweir (approved):
LGTM

Issue #34592, opened 2021-09-01
Missing Permissions in catalog management roles


ryansann:
**Rancher Server Setup**
- Rancher version: v2.6.0
- Installation option (Docker install/Helm Chart): Docker install

**Information about the Cluster**
- Cluster Type (Local/Downstream): local

**Describe the bug**

The builtin `catlaogs-manage` GlobalRole is missing permissions for verb: `*` on resource: `clusterrepos`, this permissions grants the ability to manage chart repositories. 

**To Reproduce**
 
- After Rancher starts, run: `kubectl get globalroles catalogs-manage -o yaml` against the local cluster

**Result**
```yaml
rules:
- apiGroups:
  - management.cattle.io
  resources:
  - catalogs
  - templates
  - templateversions
  verbs:
  - '*'
```

**Expected Result**

```yaml
rules:
- apiGroups:
  - management.cattle.io
  resources:
  - catalogs
  - templates
  - templateversions
  verbs:
  - '*'
- apiGroups:
  - catalog.cattle.io
  resources:
  - clusterrepos
  verbs:
  - '*'
```

ryansann:
Testing steps:

Start Rancher v2.6.0 (version without this change)
Create a standard user, check the "Configure Catalogs" builtin role
Add new user as a member to a cluster (local or downstream)
Log in as user and go to Apps & Marketplace > Chart Repositories in the cluster
Note that chart repositories cannot be modified in any way
Upgrade to this PR
Refresh browser, user can modify chart repositories
catalogs-manage GlobalRole:

rules:
- apiGroups:
  - management.cattle.io
  resources:
  - catalogs
  - templates
  - templateversions
  verbs:
  - '*'
- apiGroups:
  - catalog.cattle.io
  resources:
  - clusterrepos
  verbs:
  - '*'

ryansann:
This issue needs to be backported to 2.5

ryansann:
Small change to this. After some discussions, I've also added:
```yaml
- apiGroups:
  - catalog.cattle.io
  resources:
  - clusterrepos
  verbs:
  - '*'
```
to the `clustercatalogs-manage` Role.

The `clusterrepo` permissions are not being added to the `projectcatalogs-manage` role because there is no concept of project level catalogs in dashboard / `catalog.cattle.io`

anupama2501:
Reproduced on v2.6.0

1. Created an rke1 and rke2 node driver clusters
[...]

PR #35538, opened 2021-11-16
Revert display name change for catalogs-manage GlobalRole

ryansann:
Issue:
- https://github.com/rancher/rancher/issues/34592

Dashboard uses translations to pick the display name for the GlobalRoles in the user create view. The display name field of the role resource is only used by Ember UI, so it should remain as it was, i.e. without "Legacy". The `catalogs-manage` global role grants: 
```
rules:
- apiGroups:
  - management.cattle.io
  resources:
  - catalogs
  - templates
  - templateversions
  verbs:
  - '*'
```
The display name still needs to be changed in rancher/dashboard since the above role doesn't apply to dashboard based catalog resources (`*.catalog.cattle.io`).

The permission changes in the [previous PR](https://github.com/rancher/rancher/pull/35265) are still valid.


kinarashah (approved):

brendarearden (approved):
lgtm

PR #35539, opened 2021-11-16
Move endpoint controller to agent

ibuildthecloud:
The endpoint controller is responsible for creating annotations on workload objects so that the UI create links.  This logic is being moved to the agent to avoid watching workload resources


pkg/settings/setting.go
```diff
@@ -47,6 +47,7 @@ var (
 	InstallUUID                       = NewSetting("install-uuid", "")
 	InternalServerURL                 = NewSetting("internal-server-url", "")
 	InternalCACerts                   = NewSetting("internal-cacerts", "")
+	IsRKE                             = NewSetting("is-rke", "")
```
rmweir, line 50: Where is this getting set? Should it be added to the map of settings to pass to downstream?

pkg/controllers/managementagent/endpoints/endpoints.go
```diff
@@ -35,87 +30,38 @@ const (
...
-	ignore := func() bool {
-		return false
-	}
-	if cluster.Spec.Internal {
-		deployments := workload.Apps.Deployments("").Controller().Lister()
-		ignore = func() bool {
-			_, err := deployments.Get(namespace.System, "cattle-cluster-agent")
-			return err == nil
-		}
-	}
```
rmweir, line 58: I don't see any equivalent to this, should we still skip it for local clusters?

rmweir (changes requested):

pkg/node/node.go
```diff
@@ -83,6 +83,29 @@ func GetNodeInternalAddress(node *corev1.Node) string {
...
 
+func GetEndpointV1NodeIP(node *v1.Node) string {
+	externalIP := ""
+	internalIP := ""
+	for _, ip := range node.Status.Addresses {
+		if ip.Type == "ExternalIP" && ip.Address != "" {
+			externalIP = ip.Address
+			break
+		} else if ip.Type == "InternalIP" && ip.Address != "" {
+			internalIP = ip.Address
```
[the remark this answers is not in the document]
ibuildthecloud: There is no break because externalIP if found takes priority over internalIP.

pkg/settings/setting.go
ibuildthecloud: I think I forgot to check in that part, let me look

ibuildthecloud: This is already set here https://github.com/rancher/rancher/blob/release/v2.6/pkg/systemtemplate/template.go#L201

pkg/controllers/managementagent/endpoints/endpoints.go
```diff
@@ -290,15 +229,15 @@ func isMachineReady(machine *managementv3.Node) bool {
...
-	machines, err := machineLister.List(clusterName, labels.NewSelector())
+	nodes, err := nodesLister.List(clusterName, labels.NewSelector())
 	if err != nil {
 		return "", err
 	}
-	for _, machine := range machines {
-		if machine.Spec.Worker && nodehelper.IsMachineReady(machine) {
-			nodePublicIP := getEndpointNodeAddress(machine)
+	for _, node := range nodes {
+		if len(node.Spec.Taints) == 0 {
```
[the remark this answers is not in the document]
ibuildthecloud: I added an is ready check like before and check for the worker annotation

pkg/controllers/managementagent/endpoints/endpoints.go
ibuildthecloud: The logic before was if the user controller was running for the local cluster and rancher is managing this cluster, then don't run.  This is the hosted rancher setup.  Now that we are running the agent we don't need to do this because the agent will not run controllers if the local cluster has rancher installed.  So it's not possible to get two instances of rancher both trying to manage the endpoints

rmweir (approved):
LGTM

PR #35541, opened 2021-11-16
Move ingresshostgen to agent

ibuildthecloud:
Ingresshostgen controller is responsible for programming hostnames dynamically if a user sets the host to sslip.io.


rmweir (approved):
LGTM

PR #35543, opened 2021-11-16
Fix name collision for prtbs

ibuildthecloud:
If two projects in the same cluster have the same user bound to the
same name role they will conflict because the project id is not in the
name of the created rolebinding.  This causes a flapping.


ryansann (approved):
lgtm

deniseschannon:
https://github.com/rancher/rancher/issues/35544

PR #35545, opened 2021-11-16
If a cluster agent is disconnect show disconnected message, not error

ibuildthecloud:
#35546

### DIFF
--- a/pkg/controllers/management/clusterdeploy/clusterdeploy.go
+++ b/pkg/controllers/management/clusterdeploy/clusterdeploy.go
@@ -16,6 +16,7 @@ import (
 	"github.com/rancher/rancher/pkg/auth/tokens"
 	util "github.com/rancher/rancher/pkg/cluster"
 	"github.com/rancher/rancher/pkg/clustermanager"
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -92,6 +93,10 @@ func (cd *clusterDeploy) sync(key string, cluster *v3.Cluster) (runtime.Object,
 		return nil, nil
 	}
 
+	if cluster.Status.Driver == v32.ClusterDriverImported && clusterconnected.Connected.IsFalse(cluster) {
```
rmweir, line 96: Can a log be added here like "cluster [clusterName] deploy skipped: cluster is disconnected."

rmweir (changes requested):

pkg/controllers/management/clusterdeploy/clusterdeploy.go
ibuildthecloud: It will log a lot then.

rmweir (approved):
LGTM

Issue #35547, opened 2021-11-17
Minimize writes to the kontainer-engine v1 state secret


ibuildthecloud:
The state secret is quite large and we would like to avoid updating it when it's not needed.  There are several times that we move the state to Updating and then back to Running and nothing else changes.

git-ival:
Need additional metrics to be exposed in order to properly verify this issue. I believe @rmweir is targetted to expose more metrics after the 2.6.3 release.

git-ival:
Closing with the following:
Able to provision EKS clusters as expected without issue. Additionally, this change has been in-place throughout multiple versions' worth of scale and regression testing without negatively effecting their results.

PR #35548, opened 2021-11-17
Do not set kev1 state to Upgrading during etcdsave and post-check

ibuildthecloud:
The kev1 state secret is quite large and we do extra writes every time
we do an etcd backup or during the post-check phase of provisioning. There
is no disadvantage to just not updating the state and leaving it as Running.

https://github.com/rancher/rancher/issues/35547


rmweir (approved):
LGTM as long as RKE doesn't need the status.

Issue #33916, opened 2021-08-04
Make Keycloak refresh token more efficient


cbron:
https://github.com/rancher/rancher/pull/33896/files#r682168873
Without this there is a window where access and refresh token will expire and quite a lot of calls can be made. But once the access token is reset it will be back to normal.

sgapanovich:
### Test Environment: ###
Rancher version: 2.6-head 4375161 vs 2.6.2
Rancher cluster type: HA
Docker version: 20.10

----------

### Testing: ###
#### Set Up: ####
set up Keycloak (OIDC)
set Access Token Lifespan = 1 minutes (Keycloak site)
enabled debug logs 
to track when Rancher interacts with Keycloak I also used Keycloak admin panel (User -> find your user -> Session): every time a new token / refresh token is requested "Last Access" is updated)
![keycloak](https://user-images.githubusercontent.com/40282141/143153202-d1f6b062-7350-4090-ba94-c778a5f12f64.jpg)

--------

#### Testing: ####
the initial token expires after 1 minute and when I go to User&Authentication -> Auth Provider -> interact with the "Add Member" dropdown
- 2.6.2: a request to Keycloak to get a refresh token is sent on every interaction with the dropdown 
  - Last Access on the Keycloak is updated on every interaction
  - the request is logged in rancher logs (debug logs)
    - ![token](https://user-images.githubusercontent.com/40282141/143152823-afbfa5b5-7f7d-4edc-b017-594b48719f5f.jpg)
- 2.6-head: a request to Keycloak to get a refresh token is sent only when the Access Token Lifespan setting is met (only when 1 minute passed)
  - Last Access on the Keycloak is updated only when a token/refresh token expires (after 1 min between requests)
  - the new messages and a request are logged in rancher logs (debug logs)
    - ![token1](https://user-images.githubusercontent.com/40282141/143152924-90824699-0ba8-4782-bc88-21e4cbbdd5f9.jpg)

----------

Upgraded 2.6.2 with Keycloak user
- the user was able to log in to the cluster
- new refresh token functionality is working as expected
- refresh group memberships is functioning the same way now as the dropdown in Auth Provider 
  - a new token is not requested if the current refresh token is still valid 
- logging out / logging in back refreshes a token

PR #35549, opened 2021-11-17
Ensure keycloak oidc and oidc tokens refresh, and update store

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/33916

**Problem**
Keycloak OIDC and OIDC providers refresh the access token, but do not store the refreshed token so once the initial access token is expired for the session, all calls will get a fresh token. This just causes more calls to happen than need to. 

**Solution**
For Keycloak OIDC, use a different method to get the access token than was being used previously. One the token is fetched, validate if it is different than the one that was fetched from the secret. If it is different, then update the secret to have the new token.

For OIDC, I left the existing calls for the tokens, but I store the token that was passed in as a separate var and validate it against the one that is returned from the existing call. If it is different, then update the secret.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_provider.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_provider.go
@@ -3,6 +3,7 @@ package keycloakoidc
 import (
 	"context"
 	"encoding/json"
+	"reflect"
 	"strings"
 
 	gooidc "github.com/coreos/go-oidc/v3/oidc"
@@ -49,7 +50,8 @@ func (k *keyCloakOIDCProvider) GetName() string {
 	return Name
 }
 
-func newClient(config *v32.OIDCConfig, token *oauth2.Token) (*KeyCloakClient, error) {
+func (k *keyCloakOIDCProvider) newClient(config *v32.OIDCConfig, token v3.Token) (*KeyCloakClient, error) {
+	// creating context for new client and for refreshing oauth token if needed
 	ctx, err := oidc.AddCertKeyToContext(context.Background(), config.Certificate, config.PrivateKey)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


paynejacob (approved):

pkg/auth/providers/keycloakoidc/keycloak_provider.go
```diff
@@ -175,3 +154,32 @@ func (k *keyCloakOIDCProvider) GetPrincipal(principalID string, token v3.Token)
...
+		// since token is not valid, the TokenSource func used in the Client func will attempt to refresh the access token
+		// if the refresh token has not expired
+		logrus.Debugf("[generic oidc] RefeshAndUpdateToken: attempting to refresh access token")
+	}
+	reusedToken, err := oauth2.ReuseTokenSource(oauthToken, oauthConfig.TokenSource(ctx, oauthToken)).Token()
+	if err != nil {
+		return oauthToken, err
+	}
+
+	if reflect.DeepEqual(oauthToken, reusedToken) {
```
prachidamle, line 181: Should this be `!reflect.DeepEqual(oauthToken, reusedToken)` I thought we want to update the token if the new token returned is not same as the one we extracted from the secret?

pkg/auth/providers/oidc/oidc_provider.go
```diff
@@ -387,18 +389,22 @@ func (o *OpenIDCProvider) getUserInfo(ctx *context.Context, config *v32.OIDCConf
...
 	userInfo, err = provider.UserInfo(updatedContext, oauthConfig.TokenSource(updatedContext, oauth2Token))
 	if err != nil {
 		return userInfo, oauth2Token, err
 	}
 	if err := userInfo.Claims(&claimInfo); err != nil {
 		return userInfo, oauth2Token, err
 	}
+	if reflect.DeepEqual(oauth2Token, oldOauth2Token) {
+		o.UpdateToken(oauth2Token, claimInfo.Name)
+	}
```
prachidamle, line 407: Should the token be updated if not equal?

brendarearden: You are correct - I have updated this in my most recent changes.

pkg/auth/providers/keycloakoidc/keycloak_provider.go
brendarearden: You are correct - I have updated this in my most recent changes.

paynejacob (approved):

brendarearden:
I was previously not using the correct username to save the updated token with in the oidc_provider RefetchGroupPrincipals func, so I have changed that to include a lookup to get the user with the principalID. I am not providing a user name / id from the LoginUser func because the the oauthtoken will never be expired at that point so it should not need updated.

prachidamle (dismissed):

paynejacob (approved):

PR #35556, opened 2021-11-17
Move pod aggregration logic from nodesyncer to agent

ibuildthecloud:
The moved logic is responsible for setting the Limits and Requests fields on the v3.Node.Status.

### DIFF
--- a/pkg/controllers/managementagent/podresources/register.go
+++ b/pkg/controllers/managementagent/podresources/register.go
@@ -1,0 +1,146 @@
+package podresources
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/controllers/managementagent/podresources/register.go
```diff
@@ -0,0 +1,146 @@
...
+		// kubectl uses this cache to filter out the pods
+		if pod.Status.Phase == "Succeeded" || pod.Status.Phase == "Failed" {
+			continue
+		}
+		pods = append(pods, pod)
+	}
+	return pods, nil
+}
+
+func getAnnotation(pods []*corev1.Pod) (string, string, error) {
```
rmweir, line 86: nit:
```suggestion
func getPodResourceAnnotations(pods []*corev1.Pod) (string, string, error) {
```

rmweir (changes requested):
1 nit, otherwise LGTM

PR #35557, opened 2021-11-17
Split nsserviceaccount handler between user and agent

ibuildthecloud:
The nsserviceaccount controllers is responsible for setting AutomountServiceAccountToken=false on the default service account in system namespaces (excluding kube-system and default).

### DIFF
--- a/pkg/controllers/managementagent/nsserviceaccount/nssvcaccnt.go
+++ b/pkg/controllers/managementagent/nsserviceaccount/nssvcaccnt.go
@@ -1,0 +1,70 @@
+package nsserviceaccount
+
+import (
+	"context"
+
+	rv1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	"github.com/rancher/rancher/pkg/types/config"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


pkg/controllers/managementagent/nsserviceaccount/nssvcaccnt.go
```diff
@@ -0,0 +1,73 @@
...
+	projectIDAnnotation        = "field.cattle.io/projectId"
+	sysNamespaceAnnotation     = "management.cattle.io/system-namespace"
+	NoDefaultSATokenAnnotation = "management.cattle.io/no-default-sa-token"
+)
+
+type defaultSvcAccountHandler struct {
+	serviceAccountsLister rv1.ServiceAccountLister
+	serviceAccounts       rv1.ServiceAccountInterface
+	projectLister         v3.ProjectLister
+	clusterName           string
```
rmweir, line 25: Doesn't look like these are used, remove if so.

rmweir (changes requested):
I understand if we want to preserve it for metrics or to make it clear it's nearly the same as prior implementation but I think we might want to consider renaming the handlers since they aren't ServiceAccount handlers anymore.

deniseschannon (approved):

pkg/controllers/managementagent/nsserviceaccount/nssvcaccnt.go
ibuildthecloud: removed

rmweir (approved):
LGTM. Discussed offline, name is fine.

PR #35564, opened 2021-11-17
Defer systemimage controller start

ibuildthecloud:
Only start the systemimage controller if alert, logging, or pipeline controllers started.  The system image controller is responsible for upgrading the docker images associated with alert, logging, and pipeline.


rmweir (approved):
LGTM

Issue #35565, opened 2021-11-17
Minimize caches ran on followers


ibuildthecloud:
One rancher pod is dedicated to managing user controllers for a downstream cluster.  All other Rancher pods will be designated a follower for that downstream cluster.  Every follower starts a set of caches.  Today that includes things like Pods which causes a good deal of memory usage and traffic to the Rancher server.  We wish to minimize the caches created to only RBAC related.  Eventually we can get this to zero, but for now we really only need to cache information about RBAC (used by norman API and impersonation logic in the proxy).

git-ival:
Unable to directly verify individual caches.

Ran a Small scale test, results logged here: https://confluence.suse.com/display/EN/Small+Scale+Testing+Results 

Was able to hit the documented supported # of downstreams, but overall resource usage was noticeably higher than the previous 2.6.1 results.

git-ival:
Unable to directly verify individual caches.

Ran a Small scale test as a general regression test. Was able to hit the documented supported # of downstreams, but overall resource usage was noticeably higher than the previous 2.6.1 results.

git-ival:
Closing with the following:
After this change and other related controllers changes a decrease in memory usage is noticeable. Additionally, this change has been in-place for multiple versions' worth of regression and scale-testing efforts without negatively impacting results.

PR #35566, opened 2021-11-17
Minimize the caches ran on followers

ibuildthecloud:
The only controllers that should run on followers for user controllers
are

roles
rolebindings
clusterroles
clusterrolebindings
namespaces

Then specifically for impersonation we run secret and serviceaccount but
only watching the cattle-system-impersonation namespace

#35565

### DIFF
--- a/pkg/controllers/managementuser/controllers.go
+++ b/pkg/controllers/managementuser/controllers.go
@@ -21,7 +21,9 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy"
 	"github.com/rancher/rancher/pkg/features"
 	managementv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/impersonation"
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/controllers.go
```diff
@@ -60,19 +62,30 @@ func Register(ctx context.Context, cluster *config.UserContext, clusterRec *mana
...
+	cluster.KindNamespaces[schema.GroupVersionKind{
+		Version: "v1",
+		Kind:    "Secret",
+	}] = impersonation.ImpersonationNamespace
+	cluster.KindNamespaces[schema.GroupVersionKind{
+		Version: "v1",
+		Kind:    "ServiceAccount",
+	}] = impersonation.ImpersonationNamespace
+
 	cluster.Core.Namespaces("").Controller()
```
rmweir, line 83: I'm guessing namespaces is still need needed for RBAC stuff?

rmweir (approved):
LGTM, as long as namespaces was intentionally left.

Issue #35411, opened 2021-11-08
Clusters stuck in Configuring state when upgrade from Rancher 2.5.11 to v2.6-head


davidnuzik:
**Summary:**

This appears to be very much like https://github.com/rancher/rancher/issues/35330

Install Rancher 2.5.11 and in my case I created multiple imported clusters. I then upgraded to Rancher v2.6-head (`b6d9284`, 11/8/2021). Post-upgrade all my clusters (including the local cluster) are stuck in a "Configuring" state. It looks like all resources are Active except the Management cluster is stuck in the "Configuring" state.

There are no obvious errors in the rancher pod logs or in the fleet pod logs on the local cluster.

-----

**Environment:**
Rancher version: v2.5.11 > v2.6-head  (`b6d9284`, 11/8/2021) upgrade.
Rancher cluster type: HA
 
Downstream cluster type: Several imported clusters each with 1.20.12 kubernetes (one RKE1, one RKE2, one K3S, all single-node clusters)
Downstream K8s version: 1.20.12

**Steps to Reproduce:**
1. Install HA Rancher 2.5.11 and import several clusters, in my case K3S, RKE1, RKE2 each running k8s 1.20.12.
2. Upgrade to latest v2.6-head
 
**Expected Result:**
Downstream clusters eventually come to Active state

**Actual Result:**
All downstream clusters stay in the "Configuring" state.

jakefhyde:
This should be fixed by this PR: https://github.com/rancher/rancher/pull/35359.

davidnuzik:
https://github.com/rancher/rancher/pull/35359 should fix this

davidnuzik:
@jakefhyde the PR is merged - is there another we need? If not, could you move this issue To Test?

ronhorton:
I'm currently testing 2 rancher docker installs, one with the feature flag off during upgrade, the other with the feature flag on. Otherwise, the setups are identical.

first cluster A (rke flag deactivated) has been configuring/provisioning for 4 hours.
second cluster B (rke flag active) has been configuring/provisioning for 25 minutes

- both are 2.6-head Commit ID 638b29a
- do - rke2 - 1 single node clusters
- aws - rke2 - 1 single node cluster; rke1 1 single node cluster

jakefhyde:
### Root cause
 
[...]

PR #35568, opened 2021-11-17
Migrate Namespaces Assigned conditions from project to cluster

jakefhyde:
Description
-
Moved `DefaultNamespacesAssigned` and `SystemNamespacesAssigned` Conditions from cluster objects to associated default and system projects. During an upgrade, the project object may never have been updated (save for project sync), and having the cluster's status depend on the project being updated was causing issues due to the migrations having the wrangler controllers, but the project lifecycle having norman controllers. 

Info
-
Tested on HA  setup a number of times, confirmed issue was not reproducible.
Related Issue: https://github.com/rancher/rancher/issues/35411

### DIFF
--- a/pkg/rancher/migrations.go
+++ b/pkg/rancher/migrations.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 
 	"github.com/mcuadros/go-version"
+	"github.com/rancher/norman/condition"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/auth/tokens"
 	fleetconst "github.com/rancher/rancher/pkg/fleet"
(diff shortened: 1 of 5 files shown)


pkg/rancher/migrations.go
```diff
@@ -191,24 +192,36 @@ func forceSystemNamespaceAssignment(configMapController controllerv1.ConfigMapCo
 		return nil
 	}
 
-	clusterList, err := clusterClient.List(metav1.ListOptions{})
-	if err != nil {
-		return err
-	}
+	applyProjectConditionForNamespaceAssignment := func(label string, condition condition.Cond) error {
```
paynejacob, line 195: can we define this as a regular function and just pass in projectClient

paynejacob (changes requested):

paynejacob (approved):

thedadams (approved):

Issue #35161, opened 2021-10-18
Authorized Cluster Endpoint (ACE) support for imported RKE2/K3S clusters


thedadams:
When provisioning an RKE2/K3S cluster from Rancher, it is possible to use Authorized Cluster Endpoint on the downstream cluster. It would be good to have the same support for imported RKE2/K3S clusters.

A few notes about the support:
1. This must work for RKE2/K3S clusters imported in 2.6.x and those imported in a previous version of Rancher and then Rancher upgraded to 2.6.x
2. Manual steps to alter the config of the downstream RKE2/K3S cluster and deploy the `kube-api-authn-webhook` are expected. These steps should be documented.
3. If a future implementation of ACE requires an update to the `kube-api-authn-webhook`, then this would also have to be done manually.

JIRA ID: SURE-3474

thedadams:
UI issue: https://github.com/rancher/dashboard/issues/4410

davidnuzik:
Note: I filed https://github.com/rancher/rancher/issues/35338 while testing this issue.

davidnuzik:
**My checks PASSED**
Please note, as Donnie mentioned in a prior comment a few manual steps are required. This will get documented via the docs issue he created, also mentioned above (https://github.com/rancher/docs/issues/3612). Further, Donnie mentioned above that [the UI is not yet completed](https://github.com/rancher/dashboard/issues/4410) (as of the time of me writing this), and therefore testing was done via the kuberenetes API on the `local` cluster by modifying the cluster (clusters.provisioning.cattle.io) for each imported cluster to enable localClusterAuthEndpoint

When the UI issue is completed later and ready for test, we will test this again but via the UI. This will also enable us to test supplying an FQDN and Cert. It was not trivial to test the FQDN and Cert fields prior to the UI change, so it was agreed it would make sense to test this only after the UI change is in before release of v2.6.3.

Note, an internal test plan was also created for future reference and was adequately reviewed.
 
**Reproduction Steps:**

Not required. New feature/enhancement.

-----
 
**Validation Environment:**

Rancher version: v2.6-head `cc2b8be` 11/11/21
Rancher cluster type: HA Install
 
Downstream cluster type: imported k3s and imported rke2
Downstream K8s version:  v1.20.12+rke2r1 and  v1.20.12+k3s1 (post rancher upgrade, enable and ensure ace works, upgrade each to  v1.21.6+rke2r1 and  v1.21.6+k3s1 and ensure ACE still works)
 
[...]

PR #35577, opened 2021-11-17
Stop overwriting ACE config for imported RKE2/K3S clusters

thedadams:
When an imported cluster is created, it is created as a provisioningv2
cluster. However, after an imported cluster is recognized as an RKE2/K3S
cluster, then the management cluster is edited. This caused a problem
because the provisioning cluster controllers would overwrite the ACE
config on the management cluster as soon as the management cluster was
edited.

Now, the ACE config is deleted before the provisioning controller saves
the management cluster preventing the ACE config on the management
cluster from being overwritten.

Issue:
https://github.com/rancher/rancher/issues/35161

### DIFF
--- a/pkg/controllers/provisioningv2/cluster/controller.go
+++ b/pkg/controllers/provisioningv2/cluster/controller.go
@@ -189,15 +189,22 @@ func (h *handler) generateCluster(cluster *v1.Cluster, status v1.ClusterStatus) 
 	}
 }
 
-func NormalizeCluster(cluster *v3.Cluster) (runtime.Object, error) {
+func NormalizeCluster(cluster *v3.Cluster, isImportedCluster bool) (runtime.Object, error) {
 	// We do this so that we don't clobber status because the rancher object is pretty dirty and doesn't have a status subresource
 	data, err := convert.EncodeToMap(cluster)
 	if err != nil {
 		return nil, err
 	}
+	spec, _ := data["spec"].(map[string]interface{})
+	if _, ok := spec["localClusterAuthEndpoint"]; ok && isImportedCluster {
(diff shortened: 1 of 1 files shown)


cmurphy (commented):
Rather than deleting the field, could it just not be set in the first place?

https://github.com/rancher/rancher/pull/35577/files#diff-1de92fae2ed9cdb8c91c0bb0692270720b071a7426fc0f0a4ed731b9655f75b9R272-R276

thedadams:
No because `convert.EncodeToMap` uses `json.Marshal` under the hood and the `LocalClusterAuthEndpoint` field doesn't have `omitempty` on it. So whether it is set explicitly or not, it will get overridden if we don't delete it.

cmurphy (approved):

pkg/controllers/provisioningv2/cluster/controller.go
```diff
@@ -189,15 +189,19 @@ func (h *handler) generateCluster(cluster *v1.Cluster, status v1.ClusterStatus)
...
 }
 
-func NormalizeCluster(cluster *v3.Cluster) (runtime.Object, error) {
+func NormalizeCluster(cluster *v3.Cluster, isImportedCluster bool) (runtime.Object, error) {
 	// We do this so that we don't clobber status because the rancher object is pretty dirty and doesn't have a status subresource
 	data, err := convert.EncodeToMap(cluster)
 	if err != nil {
 		return nil, err
 	}
+	spec, _ := data["spec"].(map[string]interface{})
```
snasovich, line 198: Suggesting to add a comment explaining why we do this as it's not trivial.

snasovich (approved):

Issue #34686, opened 2021-09-08
Rancher 2.6 Cannot Login


clayrisser:
I deployed rancher 2.6 and cannot login. I followed the instructions for getting the bootstrap password, and it does not work.

https://rancher.com/docs/rancher/v2.6/en/installation/resources/bootstrap-password/

I get the following error from the Rancher UI.

```
Logging in failed: Check credentials, or your account may not be authorized to log in.
```

I tried deploying rancher with a custom bootstrap password and it also failed with the same login error.

```sh
      helm install rancher rancher-latest/rancher \
        --version ${rancher_version} \
        --namespace cattle-system \
        --set hostname=${hostname} \
        --set ingress.tls.source=letsEncrypt \
        --set bootstrapPassword=Passw0rd \
        --set letsEncrypt.email=${email}
```

tabnul:
I have the same issue, Rancher 2.6 on EKS.
Tried downgrading to 2.5.9 (after helm delete) . No option given to  reset the default password, normally it asks for this.
Helm delete doesnt remove all resources, possibly leaving conflicting resources behind?

tabnul:
After a re-deploy of the entire EKS cluster, and immediately setting the bootstrappassword in the helm values it works correctly.

Not setting it apparently breaks any new installation of rancher on the kubernetes cluster afterwards.

all4innov:
> After a re-deploy of the entire EKS cluster, and immediately setting the bootstrappassword in the helm values it works correctly.
> 
> Not setting it apparently breaks any new installation of rancher on the kubernetes cluster afterwards.

Same conclusion: with EKS version 1.20, it's mandatory to set bootstrap password in helm, else the secret will be empty. 
The autogenerated password is not working well with EKS.

paynejacob:
Changing the bootstrap password secret after will not affect the admin password.  The secret is only used once when the admin user is created.  If the secret is not found rancher will set it for easy reference.  If a password is not set on helm install there can be a delay between the helm install finishing and rancher setting the password.  @clayrisser when you say the linked instructions do not work did you see a generated password in that secret? Did the secret exist?

tabnul:
In my case, even after an hour there was no password , secret was still empty. No bootstrappassword set.

atrauzzi:
Experiencing the same on RKE2.  Password ended up blank, can't figure out how to get things going again, feels like state is pocketed all over the place by Rancher and any failure sends me on a chase to reset it.

[...]

PR #35581, opened 2021-11-17
refactored bootstrap password

paynejacob:
https://github.com/rancher/rancher/issues/34686

# Problem
Occasionally when installing rancher the bootstrap password secret will not be set.

# Solution
I have not been able to consistently reproduce this issue but when I am able to reproduce it the cluster is under extreme load and errors about api timeouts are reported.  Likely the request to set the bootstrap password is failing but because we do not fail on this error rancher continues with creating the admin user loosing the bootstrap password forever.  To fix this I have made storing the bootstrap password a required prerequisite for creating the admin user.  This ensures that if the password is ever not saved the admin user will not be created and rancher will restart.  I also simplified the logic in the helm chart to avoid confusing situations where the bootstrap password is blank.

# Testing
- clean install with no given password
- clean install with given password
- docker run with no given password
- docker run with given password
- upgrade from 2.5.x
- upgrade from 2.5.x with the `--reuse-values` flag
- upgrade from 2.6.x
- run local


pkg/data/management/bootstrap_password.go
```diff
@@ -0,0 +1,57 @@
...
+		generated = true
+		s.StringData[bootstrapPasswordSecretKey], _ = randomtoken.Generate()
+	}
+
+	// persist the password
+	if s.GetResourceVersion() != "" {
+		_, err = lister.Update(ctx, s, metav1.UpdateOptions{})
+	} else {
+		s.Name = bootstrapPasswordSecretName
+		_, err = lister.Create(ctx, s, metav1.CreateOptions{})
```
cmurphy, line 50: Is this needed? It wasn't in the original version, I'm not sure why not.

pkg/data/management/bootstrap_password.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+const (
+	bootstrapPasswordSecretName = "bootstrap-secret"
+	bootstrapPasswordSecretKey  = "bootstrapPassword"
+)
+
+// GetBootstrapPassword reads the bootstrap password from it's secret.  If the secret is not found
+// it will be set from `CATTLE_BOOTSTRAP_PASSWORD` or generated if this is empty as well.
+func GetBootstrapPassword(ctx context.Context, lister corev1.SecretInterface) (string, bool, error) {
```
cmurphy, line 21: A `SecretLister` is a lister, `SecretInterface` is not, it's doing creates and updates.

cmurphy (commented):
I can see what you meant about the original code being hard to follow :monocle_face:

pkg/data/management/bootstrap_password.go
paynejacob: I believe we removed it from the original because of the user set it they should know it. I don't think there is any harm in storing it and it will help debugging in the future because we will know the secret should always get set for new installs.

pkg/data/management/bootstrap_password.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// load the existing secret
+	s, err = lister.Get(ctx, bootstrapPasswordSecretName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return "", false, err
+	}
+
+	// if the bootstrap password is set return it
+	if s.StringData[bootstrapPasswordSecretKey] != "" {
+		return s.StringData[bootstrapPasswordSecretKey], generated, nil
```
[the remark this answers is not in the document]
paynejacob: No this code is only called once on the first start up when the admin user is created.  The bootstrap password is intended to only be used once to set a real password.

cmurphy (approved):

jakefhyde (approved):
LGTM

PR #35582, opened 2021-11-17
Add multi-os support to v2prov rke2 clusters

rosskirkpat:
- enable multi-os support for RKE2 bootstrap calls to the install scripts
- required work for adding Windows support to the existing vSphere node driver in machine 
- related to: https://github.com/rancher/windows/issues/18


rosskirkpat:
Validation that tests successfully pass 
image

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -105,9 +105,14 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
...
 	if err != nil {
 		return nil, err
-
 	}
+
 	for _, secretRef := range sa.Secrets {
+		machine, err := h.machineCache.Get(namespace, name)
+		if machine == nil || err != nil {
+			return nil, err
+		}
```
thedadams, line 114: This is not dependent on the loop. Move it outside the loop so it only happens once (if there are multiple secret references).

phillipsj (approved):
LGTM!

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
```diff
@@ -0,0 +1,245 @@
...
+			a.Nil(err)
+			a.NotNil(secret)
+
+			a.Equal("test", secret.Name)
+			a.Equal("test", secret.Namespace)
+			a.Equal("rke.cattle.io/bootstrap", string(secret.Type))
+			data := string(secret.Data["value"])
+			a.Contains(data, "CATTLE_ROLE_NONE=true")
+			a.Contains(data, fmt.Sprintf("CATTLE_TOKEN=\"%s\"", expectEncodedHash))
+			a.Contains(data, tt.args.body)
```
thedadams, line 172: Can we add a check here to ensure that the correct script (windows/linux) is used?

pkg/provisioningv2/rke2/installer/server.go
```diff
@@ -24,5 +24,8 @@ func (s *handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 	rw.Header().Set("Content-Type", "text/plain")
-	rw.Write(content)
+	_, err = rw.Write(content)
```
phillipsj, line 27: Just havine rw.Write(content) is acceptable here.

phillipsj (changes requested):

pkg/provisioningv2/rke2/installer/server.go
```diff
@@ -24,5 +24,8 @@ func (s *handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		return
+	}
```
thedadams, line 30: We are essentially ignoring the error and trying to make go-vet happy. This is the proper way to do that.
```suggestion
_, _ = rw.Write(content)
```

thedadams (changes requested):

rosskirkpat:
There are still a few outstanding issues:

- fetching the vsphere vm/template guest-os field via govmomi library is taking far too long (rancher was still querying/receiving guest-os for vms after the cluster was created and nodes were trying to join)

- I have been unable to figure out a proper method of fetching a single vsphere vm/template object without having to run through the whole list of vms and templates. Rancher fetches all VMs/templates and returns them for the user to select to clone from. Fetching the guest-os of a single one of those objects requires knowing which specific vsphere object (vm or template) they have chosen. This may not be possible solely in the backend without UI input. The other option is modifying the existing fetch of a vm/template to return the guest-os at the same time. This may be feasible and possibly easier to implement than the desired design that is currently not working.

- adding the `os` field in dynamicschema.go to the vmwarevsphere driver nodeconfigfields is currently causing all vsphere v2prov rke2 nodes to fail to provision(error below). This may be due to the broken state of cattle_dev_mode in rancher and the inability to set a persistent setting for `MachineProvisionImage` prior to Rancher first starting up. Currently, rancher is booting with the default machine image and not using the custom machine image. I have to manually set the desired image by using the v3 api. 

I will be testing if simply the presence of the `vmwarevsphere-os` flag in the vsphere machine driver (even though it's not used) resolves this error.

```console
2022/01/24 22:59:14 [INFO] rkecluster fleet-default/ross: failing bootstrap machine(s) ross-rke2-sles-5d88f78777-pp7hr: failed creating server (VmwarevsphereMachine) in infrastructure provider: CreateError: Failure detected from referenced resource rke-machine.cattle.io/v1, Kind=VmwarevsphereMachine with name "ross-rke2-sles-f1e353d3-nvrwr": arevsphere-cfgparam [--vmwarevsphere-cfgparam option --vmwarevsphere-cfgparam option]				vSphere vm configuration parameters (used for guestinfo) [$VSPHERE_CFGPARAM]
...
flag provided but not defined: -vmwarevsphere-os and join url to be available on bootstrap node
```

- The command being passed to machine:

```console
[[--driver-download-url= --driver-hash= --secret-namespace=fleet-default --secret-name=ross-rke2-sles-d5a7103f-4qnft-machine-state create --driver=vmwarevsphere --custom-install-script=/run/secrets/machine/value --machine-os=linux --vmwarevsphere-cfgparam=disk.enableUUID=TRUE --vmwarevsphere-clone-from=/***/vm/***/ross-sles-prep --vmwarevsphere-cloud-config=/path/to/machine/files/cloud-config --vmwarevsphere-cpu-count=4 --vmwarevsphere-creation-type=template --vmwarevsphere-datacenter=/*** --vmwarevsphere-datastore=/***/datastore/*** --vmwarevsphere-disk-size=35000 --vmwarevsphere-folder=/***/vm --vmwarevsphere-memory-size=8192 --vmwarevsphere-network=/***/network/*** --vmwarevsphere-os=linux --vmwarevsphere-pool=/***/host/***/Resources/***1 --vmwarevsphere-ssh-password=tcuser --vmwarevsphere-ssh-port=22 --vmwarevsphere-ssh-user-group=staff --vmwarevsphere-ssh-user=docker --vmwarevsphere-vcenter-port=443 ross-rke2-sles-d5a7103f-4qnft]]
```

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -253,6 +253,27 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
...
+			result = append(result, &rkev1.RKEBootstrapTemplate{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						rke2.MachineOSLabel: machinePool.OS,
+					},
+				},
+			})
+		} else {
+			// default to linux as the MachineOS if `machinePool.OS` is empty
+			result = append(result, &rkev1.RKEBootstrapTemplate{
```
luthermonson, line 268: the if/else are identical except for the contents of the label. perhaps make the template with the default machine OS and then do an if the os > 0 and set the label and then append the result. turns big if/else into a quick label swap on the same object

rosskirkpat: I believe I addressed this in the latest commit

rosskirkpat:
I have found a potential blocker issue that I PR'd a fix for here: https://github.com/rancher/rancher/pull/36339

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
rosskirkpat: this is now irrelevant due to design changes.

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
rosskirkpat: addressed.

phillipsj (changes requested):

pkg/api/norman/customization/vsphere/getters.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"strings"
+
+	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+)
+
+// vmPath will be passed in from the UI and contains the inventory path to the VM being cloned
```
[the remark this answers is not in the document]
rosskirkpat: Done, also applied this to other functions I added

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -73,8 +76,21 @@ const (
...
 )
 
+// WindowsCheck return a bool based on if the driver is marked as
+// supporting Windows
+func WindowsCheck(driver string) bool {
+	if _, ok := windowsDrivers[driver]; ok {
+		return true
+	}
+	return true
+
```
thedadams, line 91:
```suggestion
	_, ok := windowsDrivers[driver]
	return ok
```

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -38,6 +38,7 @@ const (
 	MachineIDLabel             = "rke.cattle.io/machine-id"
 	MachineNameLabel           = "rke.cattle.io/machine-name"
 	MachineNamespaceLabel      = "rke.cattle.io/machine-namespace"
+	MachineOSAnnotation        = "rke.cattle.io/machine-os"
```
thedadams, line 41: Is there a reason we need this one and `CattleOSLabel`?

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -113,7 +113,15 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 			return nil, err
 		}
 
-		data, err := installer.LinuxInstallScript(context.WithValue(context.Background(), tls.InternalAPI, hasHostPort), base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
+		is := installer.LinuxInstallScript
+		// This is only used by the vmwarevsphere node driver as of Jan 2022
+		// as we add multi-platform support to more node drivers
+		// this should continue to work based on changes in
+		// pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
+		if os, ok := machine.GetAnnotations()[rke2.MachineOSAnnotation]; ok && os == rke2.WindowsMachineOS {
```
thedadams, line 121:
```suggestion
		if os := machine.Annotations[rke2.MachineOSAnnotation]; os == rke2.WindowsMachineOS {
```

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -113,7 +113,15 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 			return nil, err
 		}
 
-		data, err := installer.LinuxInstallScript(context.WithValue(context.Background(), tls.InternalAPI, hasHostPort), base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
+		is := installer.LinuxInstallScript
+		// This is only used by the vmwarevsphere node driver as of Jan 2022
+		// as we add multi-platform support to more node drivers
+		// this should continue to work based on changes in
+		// pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```
thedadams, line 120: This comment is not necessary based on the code here.

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
```diff
@@ -0,0 +1,274 @@
+package bootstrap
```
thedadams, line 1: I am not sure how I feel about this. If my knowledge is correct, this will be the only unit tests in Rancher that mocks any controller interfaces.

I'm not saying that it should be removed, but I am certainly not opposed to removing it.

pkg/controllers/provisioningv2/rke2/dynamicschema/dynamicschema.go
```diff
@@ -4,6 +4,10 @@ import (
 	"context"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
+	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
+
```
thedadams: Remove blank lines.

pkg/controllers/provisioningv2/rke2/dynamicschema/dynamicschema.go
```diff
@@ -159,6 +170,19 @@ func removeKey(fields map[string]schemas.Field, key string) map[string]schemas.F
...
 
+func addField(rFields map[string]schemas.Field, name, newField string) map[string]schemas.Field {
+	newf := rFields
+	if _, ok := newf[newField]; !ok {
+		newf[newField] = schemas.Field{
+			Type:   "string",
+			Create: true,
+			Update: true,
+		}
+		logrus.Debugf("[addField] schema [%s] has new field [%s] with values [\n%v]", name, newField, newf[newField])
```
thedadams: I think this can be removed.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -60,6 +60,11 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		url, hash, cloudCredentialSecretName string
 		jobBackoffLimit                      int32
 		filesSecret                          *corev1.Secret
+		bootstrapOS                          string
```
thedadams, line 63:
```suggestion
		machineOS                          string
```

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -60,6 +60,11 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
...
+		supportedMachineOS                   = map[string]struct{}{
+			"linux":   {},
+			"windows": {},
+		}
```
thedadams, line 67: Can this not be used but use the constants in the `rke2` package?

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -118,7 +128,12 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		cmd = append(cmd, "create",
 			fmt.Sprintf("--driver=%s", driver),
 			fmt.Sprintf("--custom-install-script=/run/secrets/machine/value"))
-
+		if rke2.WindowsCheck(driver) {
+			machineOS, ok := supportedMachineOS[bootstrapOS]
```
thedadams, line 132: I don't see anywhere where `bootstrapOS` is getting set anywhere, so I think it will always be an empty string here.

thedadams: I think this might be why the provisioning tests are failing.

thedadams (changes requested):

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: Fixed, thanks for catching this!

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
rosskirkpat: We have been instructed and requested to write unit tests for code that we are committing to Rancher. This unit test ensures that Windows functionality will be verified alongside Linux.

pkg/controllers/provisioningv2/rke2/dynamicschema/dynamicschema.go
rosskirkpat: resolved

pkg/controllers/provisioningv2/rke2/dynamicschema/dynamicschema.go
rosskirkpat: Thanks - I removed this

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
thedadams: But this code doesn't care about which driver is being used. This code will "work" regardless of which driver is being used. There are other places where the code will break if windows is being used with a driver that doesn't support it. That is a more appropriate place for this comment.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
thedadams: If the annotation doesn't exist, then it won't be the value that you need it to be. It will be empty. Checking both is not necessary.

pkg/controllers/provisioningv2/rke2/common.go
thedadams: So why create a new one and not continue to use the one that already exists?

rosskirkpat: My thoughts behind making a new annotation and not using the label were as follows:
- rke1 and rke2 custom windows clusters use the `cattle.io/os` label for provisioning logic as the OS of the VM is known at the time of registering the node to a cluster
- node driver clusters do not have the `cattle.io/os` applied to them at any point during provisioning and I am unsure if it is currently added on after a machine is finished provisioning. 
- Since these are v2prov node driver pools and leveraging capi machines, I felt the use of an annotation matching the other naming structures of the v2prov labels/annotations was appropriate.

Additionally, this is meant to differentiate and make it easier to filter/search in a cluster for windows machines that are provisioned through v2prov.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
rosskirkpat: understood, thank you for explaining! Fixing this in my next push

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
rosskirkpat: I changed the comment to a single line to indicate which node drivers support multiple OSs and sending a windows install script to. 

> 		// OS-specific logic is only implemented for the vmwarevsphere node driver as of Feb 2022

pkg/controllers/provisioningv2/rke2/common.go
thedadams: I am a fan of keeping this uniform in all these situations, but I am willing to be overruled by a majority.

In addition, we do use labels for somethings like roles.

rosskirkpat:
The provisioning tests are failing due to the dynamicschema changes I made adding an `os` field to the vmwarevsphere nodeconfig that does not yet exist on the machine side.

I am working to get my machine PR merged asap as that will unblock this PR

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -90,6 +94,11 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		},
 	}
 
+	machine, err := h.machines.Get(infraObj.meta.GetNamespace(), infraObj.meta.GetName())
```
thedadams, line 97: The infrastructure object and the CAPI machine don't have the same name. I think you can do:
```suggestion
	machine, err := rke2.GetMachineByOwner(h.machines, infraObj.meta)
```

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -90,6 +94,11 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
...
+	if err != nil && !apierror.IsNotFound(err) {
```
thedadams, line 98: I don't think we want to continue if the machine isn't found.

thedadams (changes requested):

pkg/controllers/provisioningv2/rke2/common.go
rosskirkpat: I will switch from using a new label/annotation to using CattleOSLabel

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -118,7 +126,15 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		cmd = append(cmd, "create",
 			fmt.Sprintf("--driver=%s", driver),
 			fmt.Sprintf("--custom-install-script=/run/secrets/machine/value"))
-
+		if rke2.WindowsCheck(driver) {
+			machineOS, ok := machine.GetAnnotations()[rke2.MachineOSAnnotation]
```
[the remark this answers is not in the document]
rosskirkpat: The error checks have been removed

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: agreed, fixed this

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: I have addressed this and I am now using the capi machine object for additional lookups in `getArgsEnvAndStatus` and also passing the machine object into getSecretData to simplify the number of look ups.

phillipsj (approved):
LGTM!

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -209,8 +200,8 @@ func (h *handler) getSecretData(meta metav1.Object, obj data.Object, create bool
 		cloudCredentialSecretName = oldCredential
 	}
 
-	if cloudCredentialSecretName != "" {
-		secret, err := GetCloudCredentialSecret(h.secrets, meta.GetNamespace(), cloudCredentialSecretName)
+	if cloudCredentialSecretName != "" && machine != nil {
```
thedadams, line 203: This check is not necessary because we check `machine == nil` above.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -134,12 +137,12 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 
 	// cloud-init will split the hostname on '.' and set the hostname to the first chunk. This causes an issue where all
 	// nodes in a machine pool may have the same node name in Kubernetes. Converting the '.' to '-' here prevents this.
-	cmd = append(cmd, strings.ReplaceAll(infraObj.meta.GetName(), ".", "-"))
+	cmd = append(cmd, strings.ReplaceAll(machine.GetName(), ".", "-"))
```
thedadams, line 140: Why is this changing?

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -134,12 +137,12 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
...
 
 	return driverArgs{
 		DriverName:          driver,
-		MachineName:         infraObj.meta.GetName(),
-		MachineNamespace:    infraObj.meta.GetNamespace(),
+		MachineName:         machine.GetName(),
+		MachineNamespace:    machine.GetNamespace(),
```
thedadams, line 145: Why are these changing?

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -119,7 +122,7 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 			fmt.Sprintf("--driver=%s", driver),
 			fmt.Sprintf("--custom-install-script=/run/secrets/machine/value"))
 
-		rancherCluster, err := h.rancherClusterCache.Get(infraObj.meta.GetNamespace(), infraObj.meta.GetLabels()[capi.ClusterLabelName])
+		rancherCluster, err := h.rancherClusterCache.Get(machine.GetNamespace(), machine.GetLabels()[capi.ClusterLabelName])
```
thedadams, line 125: Why is this changing?

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -104,13 +108,12 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
...
 	}
-
-	secretName := rke2.MachineStateSecretName(infraObj.meta.GetName())
+	secretName := rke2.MachineStateSecretName(machine.GetName())
 
 	cmd := []string{
 		fmt.Sprintf("--driver-download-url=%s", url),
 		fmt.Sprintf("--driver-hash=%s", hash),
-		fmt.Sprintf("--secret-namespace=%s", infraObj.meta.GetNamespace()),
+		fmt.Sprintf("--secret-namespace=%s", machine.GetNamespace()),
```
thedadams, line 116: Why is this changing?

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -104,13 +108,12 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		k := strings.ToUpper(envName + "_" + regExHyphen.ReplaceAllString(k, "${1}_${2}"))
 		envSecret.Data[k] = []byte(v)
 	}
-
-	secretName := rke2.MachineStateSecretName(infraObj.meta.GetName())
+	secretName := rke2.MachineStateSecretName(machine.GetName())
```
thedadams, line 111: Why is this changing?

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -73,8 +75,18 @@ const (
...
+	}
 )
 
+// WindowsCheck return a bool based on if the driver is marked as
+// supporting Windows
+func WindowsCheck(driver string) bool {
+	_, ok := windowsDrivers[driver]
+	return ok
+}
+
```
thedadams, line 89: It looks like this is no longer used and can be removed.

thedadams (changes requested):
Most of the comments here are about why we are changing to using the machine name and namespace in various places. The machine and infrastructure machine have different names. We may be introducing regressions in Rancher or the UI by switching using the name from the CAPI machine rather than the infrastructure machine.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: I thought the same but Goland is yelling about this. It's due to the `machine == nil ` being checked alongside `create`. If we checked them separately it would not be an issue.

The first screenshot is with the `machine == nil` check I added commented out and goland complaining. The second screenshot is with my `machine == nil` check present and the `create` bool commented out above. 

image

image

thedadams: Ah, you're right. I'm sorry for the confusion.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: this was a mistake on my end, reverted

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: this was a mistake on my end, reverted

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: this was a mistake on my end, reverted

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: this was a mistake on my end, reverted

pkg/controllers/provisioningv2/rke2/common.go
rosskirkpat: `WindowsCheck` is still used in dynamicschema https://github.com/rosskirkpat/rancher/blob/3ce8e4f7485c3df5bea5e9d1f103be6bed6617cf/pkg/controllers/provisioningv2/rke2/dynamicschema/dynamicschema.go#L59

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
rosskirkpat: mistake on my end, I reverted the change

thedadams (approved):

thedadams (dismissed):

thedadams (approved):

phillipsj (approved):
LGTM

Issue #35593, opened 2021-11-18
[custom cluster] adding 1 node all roles to custom rke2 cluster fails with `unable to create impersonator account: error getting service account token`


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.6-head (936c182)
- Installation option (Docker install/Helm Chart): docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:n/a

**Information about the Cluster**
- Kubernetes version:  v1.21.6+rke2r1

- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): custom (with aws node)

**Describe the bug**

when adding the first node to the custom custer (all roles), node begins to register but fails with an impersonation error. 

**To Reproduce**
 
* deploy a custom cluster on rancher
* run the command from the previous step (all roles should be selected) on a node/vm

**Result**
fails to add node with the following error in rancher:
```
Error while applying agent YAML, it will be retried automatically: exit status 1, Error from server (InternalError): an error on the server ("unable to create impersonator account: error getting service account token: error getting secret: secrets \"cattle-impersonation-system/cattle-impersonation-u-djxin436by-token-9vgqz\" not found") has prevented the request from succeeding
```

**Expected Result**

* custom cluster should allow registration of nodes and nodes should come to an active state

cmurphy:
Is this reliably reproducible?

cmurphy:
Hi @shmutalov, that looks like a different error message, the error message in this bug was `... error getting service account token: error getting secret:...` and the bug was introduced pre-release, while the error in your logs is `...error setting up impersonation for user u-vfc754tywx: Unauthorized```. Could you open up a separate bug with more information about your setup and what was happening when this error occurred?

shmutalov:
@cmurphy I am getting that bug with rancher:latest and `rancher:v2.6-9740d7ab5b988456ad34423d81dba28c9640bb6c-head`. I am trying to create single node instance with known CA issued certificate.

```
[...]

PR #35596, opened 2021-11-18
Always start the Secrets controller

cmurphy:
The impersonation mechanism relies on the Secrets controller being
started on the user context when a proxy request is made. Before
6b2e1359, the controller was incidentally being started as part of
registering other controllers. With 6b2e1359, the certsexpiration
controller was the last to always start it on startup and now only
conditionally starts it for RKE clusters. This change adds back the
Secrets controller startup to the cluster context unconditionally so
that it is ready the first time impersonation.GetToken is called.

https://github.com/rancher/rancher/issues/35593

### DIFF
--- a/pkg/controllers/managementuser/controllers.go
+++ b/pkg/controllers/managementuser/controllers.go
@@ -51,6 +51,8 @@ func Register(ctx context.Context, cluster *config.UserContext, clusterRec *mana
 
 	// register controller for API
 	cluster.APIAggregation.APIServices("").Controller()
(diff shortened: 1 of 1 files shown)


kinarashah (approved):

brendarearden (approved):
LGTM

kinarashah:
Requesting review from @rmweir since he reviewed the original defer PR.

rmweir (approved):
LGTM

PR #35622, opened 2021-11-20
Add drain pre and post hook and ProvisioningGeneration

ibuildthecloud:
PostDrainHooks and PreDrainHook fields are added to the DrainOptions.
The pre drain hook will run after cordon but before drain. The
post drain hook will run after the node has been updated (for example
upgrading rke2) but before the node is uncordoned.

#35622


pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
```diff
@@ -41,6 +41,7 @@ type handler struct {
 	configMapsCache corev1controllers.ConfigMapCache
 	configMaps      corev1controllers.ConfigMapController
 	clusterCache    provisioningcontrollers.ClusterCache
+	namespaceCache  corev1controllers.NamespaceCache
```
thedadams, line 44: This doesn't seem to be used anywhere.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -41,12 +42,12 @@ func (h *handler) OnChange(key string, machine *capi.Machine) (*capi.Machine, er
 	}
 
 	drain := machine.Annotations[planner.DrainAnnotation]
-	if drain != "" {
-		return h.drain(machine, []byte(drain))
+	if drain != "" && machine.Annotations[planner.DrainDoneAnnotation] != drain {
+		return h.drain(machine, drain)
 	}
 
 	if machine.Annotations[planner.UnCordonAnnotation] != "" {
```
thedadams, line 49: Should this check be more explicit?
```suggestion
	if machine.Annotations[planner.UnCordonAnnotation] == drain {
```

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -66,71 +67,109 @@ func (h *handler) k8sClient(machine *capi.Machine) (kubernetes.Interface, error)
...
-}
+	if err := drain.RunCordonOrUncordon(helper, node, false); err != nil {
+		return nil, err
+	}
 
-func (h *handler) drain(machine *capi.Machine, drainData []byte) (*capi.Machine, error) {
-	hash := planner.DrainHash(drainData)
-	if machine.Annotations[planner.DrainDoneAnnotation] == hash {
-		return machine, nil
+	if machine.Annotations[planner.UnCordonAnnotation] != "" {
```
thedadams, line 102: Same as above?
```suggestion
	if machine.Annotations[planner.UnCordonAnnotation] == drain {
```

thedadams: Also, a comment here would be nice. It looks like the uncordon has finished so we are resetting all the annotations so that the next drain happens as expected.

ibuildthecloud: This check is redundant so I've removed it.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
ibuildthecloud: Added a comment explaining above.

thedadams (approved):

Issue #35628, opened 2021-11-21
"Failed to parse cluster status version: Invalid character(s) found in minor number "21+"" Error seen on a fresh install of EKSv2 cluster


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6head commit id `e322f8f`
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1 - 1.22.3-rancher1-1

**Information about the Cluster**
- Kubernetes version: 1.21
- Cluster Type (Local/Downstream): EKS v2 downstream cluster

**Describe the bug**
"Failed to parse cluster status version: Invalid character(s) found in minor number "21+"" Error seen on a fresh install of EKSv2 cluster

**To Reproduce**

- Deploy an EKS v2 cluster. 
- The cluster is in error state in rancher

<img width="1419" alt="Screen Shot 2021-11-21 at 11 44 56 AM" src="https://user-images.githubusercontent.com/26032343/142776657-48250cf7-8784-4a83-81c4-c6f3666580f7.png">

- View in API of the cluster:
```
"version": {
"buildDate": "2021-09-13T14:20:15Z",
"compiler": "gc",
"gitCommit": "5f6d83fe4cb7febb5f4f4e39b3b2b64ebbbe3e97",
"gitTreeState": "clean",
"gitVersion": "v1.21.2-eks-06eac09",
"goVersion": "go1.16.5",
"major": "1",
"minor": "21+",
"platform": "linux/amd64",
"type": "/v3/schemas/i
```

sowmyav27:
**On 2.6-head commit id: `437516`**

Able to deploy an EKS v2 cluster with no errors

Issue #35627, opened 2021-11-21
Panic in rancher server while provisioning an RKE1 cluster


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `e322f8`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: 1.21
- Cluster Type (Local/Downstream): RKE1 DO cluster

**Describe the bug**

**To Reproduce**

- Provision a DO RKE1 cluster
- Rancher restarts with panic
- Logs in rancher server

```
2021/11/21 05:05:46 [INFO] Starting apps/v1, Kind=DaemonSet controller
2021/11/21 05:05:46 [INFO] Starting apps/v1, Kind=StatefulSet controller
2021/11/21 05:05:46 [INFO] Starting batch/v1, Kind=Job controller
2021/11/21 05:05:46 [INFO] Starting batch/v1beta1, Kind=CronJob controller
2021/11/21 05:05:46 [INFO] Starting apps/v1, Kind=ReplicaSet controller
2021/11/21 05:05:46 [INFO] Starting /v1, Kind=ReplicationController controller
2021/11/21 05:05:46 [INFO] Starting /v1, Kind=Service controller
2021/11/21 05:05:46 [INFO] Starting /v1, Kind=Event controller
2021/11/21 05:05:46 [INFO] Starting /v1, Kind=Pod controller
2021/11/21 05:05:46 [ERROR] error syncing 'c-9gkpz/p-b2qkv': handler system-image-upgrade-controller: upgrade cluster c-9gkpz system service logging failed: template system-library-rancher-logging incompatible with rancher version or cluster's [c-9gkpz] kubernetes version, requeuing
2021/11/21 05:05:46 [ERROR] error syncing 'local/p-5tcps': handler system-image-upgrade-controller: upgrade cluster local system service logging failed: template system-library-rancher-logging incompatible with rancher version or cluster's [local] kubernetes version, requeuing
2021/11/21 05:05:46 [ERROR] error syncing 'c-9gkpz/p-b2qkv': handler system-image-upgrade-controller: upgrade cluster c-9gkpz system service logging failed: template system-library-rancher-logging incompatible with rancher version or cluster's [c-9gkpz] kubernetes version, requeuing
[...]

PR #35635, opened 2021-11-22
fix semver parsing of cluster status version

kinarashah:
- Rancher server panics when cluster.status.version is empty
Stats aggregator writes the cluster status, so we can continue relying on status version but return nil if it's not updated yet 
https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/clusterstats/statsaggregator.go#L163
https://github.com/rancher/rancher/issues/35627

- Parsing eks v2 version doesn't work with normal semver parsing because gitVersion minor has "21+".  
Upstream library has parseSemantic, but can't use it directly since it returns numbers + extra, so using the first part https://github.com/kubernetes/apimachinery/blob/v0.22.3/pkg/util/version/version.go#L43-L47 
https://github.com/rancher/rancher/issues/35628

### DIFF
--- a/pkg/controllers/managementuser/healthsyncer/healthsyncer.go
+++ b/pkg/controllers/managementuser/healthsyncer/healthsyncer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"time"
 
@@ -27,6 +28,9 @@ import (
 const (
 	syncInterval = 15 * time.Second
 )
+
+// kubernetes/apimachinery/pkg/util/version/version.go
+var versionMatchRE = regexp.MustCompile(`^\s*v?([0-9]+(?:\.[0-9]+)*)(.*)*$`)
 
 var excludedComponentMap = map[string][]string{
 	"aks":     {"controller-manager", "scheduler"},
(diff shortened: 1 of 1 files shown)


pkg/controllers/managementuser/healthsyncer/healthsyncer.go
```diff
@@ -82,20 +86,29 @@ func (h *HealthSyncer) getComponentStatus(cluster *v3.Cluster) error {
...
-	k8sVersion, err := semver.Parse(fmt.Sprintf("%s.%s.0", cluster.Status.Version.Major, cluster.Status.Version.Minor))
+	if cluster.Status.Version == nil {
+		return nil
+	}
+	parts := versionMatchRE.FindStringSubmatch(cluster.Status.Version.String())
+	if parts == nil {
+		return condition.Error("ComponentStatusFetchingFailure", fmt.Errorf("Failed to parse cluster status version %s",
+			cluster.Status.Version.String()))
+	}
+	k8sVersion, err := semver.Parse(parts[1])
```
paynejacob, line 101: is there any chance of parts having a length of 1?  97 could be `parts == nil || len(parts) < 2`

paynejacob (changes requested):

pkg/controllers/managementuser/healthsyncer/healthsyncer.go
kinarashah: I had a similar question, but they seem to be accessing indices directly, 
https://github.com/kubernetes/apimachinery/blob/v0.22.3/pkg/util/version/version.go#L47

```
	if parts == nil {
		return nil, fmt.Errorf("could not parse %q as version", str)
	}
	numbers, extra := parts[1], parts[2]
```
I'll add a check to be safe.

thedadams (approved):

paynejacob (approved):

Issue #35618, opened 2021-11-19
Unable to provision RKE2 clusters using Amazon cloud provider with k8s 1.22


davidnuzik:
**Summary:**

I'm unable to provision any ec2 RKE2 clusters with Amazon cloud provider option.

-----

**Environment:**
Rancher version: v2.6-head `b736007` (also seen on `9740d7a` build from earlier today) 11/19/2021
Rancher cluster type: single-node docker install and HA helm install both attempted
 
Downstream cluster type: RKE2 w/ ec2 node driver and amazon cloud provider
Downstream K8s version: v1.21.6+rke2r1

**Steps to Reproduce:**
1. Provision an RKE2 cluster with ec2 node driver. Set up amazon cloud provider.
2. The cluster is stuck in provisioning state forever. It is stuck on step `waiting on probes: calico`
3. Try provisioning an ec2 rke2 cluster but without cloud provider. This works.
 
**Expected Result:**
I expected to be able provision an RKE2 cluster using the amazon cloud provider.

**Actual Result:**
I cannot provision an RKE2 cluster using the amazon cloud provider. This was working recently, just yesterday.

**Additional Info:**
I attempted to do my due diligence. I didn't spot anything in rancher server logs of use. When I downloaded the SSH key to remove in to the node and investigate the SSH key does not work (and I don't think I made any mistakes using the rancher provided key).

thedadams:
RKE2 PR: https://github.com/rancher/rke2/pull/2163

thedadams:
It was decided that the best way to fix this issue was through RKE2 (the PR is linked).

There is a work-around for this issue: add the environment variable `CATTLE_MACHINE_PROVISION_IMAGE` to `rancher/machine:v0.15.0-rancher72`. This will not set the hostname on the EC2 nodes and this will allow provisioning to continue. This can be used until the RKE2 fix is available.

NOTE: this is only for RKE2 clusters provisioned from Rancher with the AWS cloud provider enabled. All other provisioning is unaffected.

snasovich:
@thedadams , thank you. Depending on which RKE2 version the fix is going into and whether it's included in 2.6.3 release's KDM, we may want to add this workaround to release notes hence `release-note` label.

rancher-max:
Gave this a quick check using KDM branch `dev-v2.6` on rancher v2.6.3-rc3. I was able to provision a v1.21.7-rc2+rke2r1 cluster using aws cloud provider.

[...]

PR #35638, opened 2021-11-22
Set node-name when AWS cloud-provider is set

thedadams:
The AWS cloud provider expects the node-name to be a very specific
value. This change will ensure that the is the case.

Issue:
https://github.com/rancher/rancher/issues/35618


thedadams:
Closing in favor of [the RKE2 fix](https://github.com/rancher/rke2/pull/2163)

PR #35649, opened 2021-11-23
[WIP 2.6.3] Add system-namespace annotation to local namespace for downstream clusters

prachidamle:


prachidamle:
This is not needed since the hardening scripts will patch the default serviceaccout of this "local" namespace.

PR #35675, opened 2021-11-29
Skip cloud cred check if it is not changed

thedadams:
A cluster owner should be able to edit a cluster without changing the
cloud credential. Therefore, if the edit isn't changing the cloud
credential, then Rancher shouldn't validate that the user has access to
the cloud credential.

After this change, Rancher validates that the user has access to the
cloud credential only if they are trying to update it.

Issue:
https://github.com/rancher/rancher/issues/35413

### DIFF
--- a/pkg/api/norman/customization/cluster/validator.go
+++ b/pkg/api/norman/customization/cluster/validator.go
@@ -359,7 +359,8 @@ func (v *Validator) validateAKSConfig(request *types.APIContext, cluster map[str
 	}
 
 	// check user's access to cloud credential
-	if azureCredential, ok := aksConfig["azureCredentialSecret"].(string); ok {
+	if azureCredential, ok := aksConfig["azureCredentialSecret"].(string); ok && (prevCluster == nil || azureCredential != prevCluster.Spec.AKSConfig.AzureCredentialSecret) {
+		// Only check that the user has access to the credential if the credential is being changed.
 		if err := validateCredentialAuth(request, azureCredential); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/api/norman/customization/cluster/validator.go
```diff
@@ -601,7 +601,11 @@ func validateEKSKubernetesVersion(spec *v32.ClusterSpec, prevCluster *v3.Cluster
 }
 
 // validateCredentialAuth validates that a user has access to the credential they are setting.
-func validateCredentialAuth(request *types.APIContext, credential string) error {
+func validateCredentialAuth(request *types.APIContext, credential string, prevCluster *v32.Cluster) error {
+	if prevCluster != nil && prevCluster.Spec.EKSConfig.AmazonCredentialSecret == credential {
```
snasovich, line 605: It doesn't look right that we're checking `EKSConfig.AmazonCredentialSecret` for GKE/Azure cases as well (this is called from `validateAKSConfig` and `validateGKEConfig`.

snasovich (changes requested):

pkg/api/norman/customization/cluster/validator.go
thedadams: Dang it. What a terrible error. I will fix this.

thedadams: Updated.

cmurphy (commented):
I thought that the way kev2 updates worked was that only the changed fields would be present in the request, so e.g. if we change `kubernetesVersion` then `gkeConfig` would only be `{"kubernetesVersion": "1.21.5-gke.1802"}` and wouldn't include the cloud credential. Is this going to change with the dashboard change https://github.com/rancher/dashboard/issues/4529 ?

thedadams:
@cmurphy You are correct about how the UI handles this. However, when making a request through the API directly (using the "API UI") then the whole object is set with the request. This addresses that scenario.

cmurphy (approved):

rmweir (approved):
LGTM

Issue #35636, opened 2021-11-22
Ingresses not loading for RKE1 downstream clusters on k8s 1.22.3 when HA local cluster is on k8s 1.18


anupama2501:
**Rancher Server Setup**
- Rancher version: v2.6-head e322f8f
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1, v1.18.20, 1.3.3-rc1
- Proxy/Cert Details: Selfsigned

**Information about the Cluster**
- Kubernetes version: v1.22.3-rancher1-1
- Cluster Type (Local/Downstream): Downstream AWS/DO RKE1 node driver 1 worker, 1 etcd, 1 cp

**Describe the bug**
Ingress page is not loading in any of the downstream RKE1 and RKE2 clusters when the k8s version for these clusters is 1.22 and the local HA k8s version is on `v1.18.20`

**To Reproduce**
1. Install rancher server on RKE1 cluster with k8s selected as `v1.18.20`
2. Create a downstream RKE1 node driver/custom EC2 or DO on k8s v1.22.3-rancher1-1
3. Once the cluster is active, navigate to load balancer page from the v3 API

**Result**
ingresses.meta.k8s.io is not found. The following error is displayed in the page:
```
{
"baseType": "error",
"code": "NotFound",
"message": "the server could not find the requested resource (get ingresses.meta.k8s.io)",
"status": 404,
"type": "error"
}
```

**Expected Result**
Expected the ingresses to work on k8s 1.22 downstream clusters

**Additional context**
| k8s local | Downstream K8s| Status|
|--|--|--|
|1.18.20 | 1.22.3| FAIL|
|1.18.20| v1.21.6,v1.18.20| PASS|
|1.22.3|v1.21.6,v1.18.20,1.22.3| PASS|

anupama2501:
Verified on v2.6-head 77392a6

1. Installed rancher on HA server with k8s v1.22.4-rancher1-1
2. Created a downstream node driver AWS rke1 cluster 1 etcd+cp, 1 worker
3. Verified by creating a few ingresses and the ingresses are working

PR #35697, opened 2021-11-30
Dynamically choose Ingress proxy store

cmurphy:
Without this change, the norman store proxy setup chooses the Ingress
API to register based on the local cluster context at startup. It does
not know about other clusters at this point and can't register different
APIs on a cluster-by-cluster basis. This causes a problem when the local
cluster is version 1.18 (does not support networking.k8s.io/v1/Ingress)
and the downstream cluster is 1.22 (does not support
extensions/v1beta1/Ingress) or vice versa, where when the user tries to
navigate to the downstream cluster the wrong Ingress API will have been
registered and the Watch will fail with a 404 error.

This change addresses the problem by first statically registering the
networking.k8s.io/v1/Ingress API into the schema store on the assumption
that this will be used most of the time, and then adding an additional
Store object to be used dynamically during the request when the actual
cluster version is detected.

https://github.com/rancher/rancher/issues/35636
https://github.com/rancher/rancher/issues/35764

### DIFF
--- a/pkg/api/norman/store/ingress/store.go
+++ b/pkg/api/norman/store/ingress/store.go
@@ -1,6 +1,7 @@
 package ingress
 
 import (
+	"context"
 	"crypto/md5"
 	"encoding/base64"
 	"encoding/hex"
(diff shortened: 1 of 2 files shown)


pkg/api/norman/store/ingress/store.go
```diff
@@ -9,44 +10,126 @@ import (
...
+	clusterName := p.ClusterManager.ClusterName(apiContext)
+	return p.ClusterManager.UserContext(clusterName)
+}
+
+// legacyIngressProxyStore returns a norman proxy store object for the
+// extensions/v1beta1/Ingress API that can be used to sidestep the registered
+// networking.k8s.io/v1/Ingress API if the server doesn't support it.
+// TODO: remove when support for k8s 1.18 is dropped
+func (p *Store) legacyIngressProxyStore() types.Store {
+	return proxy.NewProxyStore(
```
kinarashah, line 122: Is there a reason why we need to create a new proxy store on every call vs having it as a field on the Store?

kinarashah (approved):
Getting downstream cluster context on every list/watch is a bit of concern since that would happen for all k8s versions, but I don't know of a better way to handle this. 
Implementation changes LGTM 👍🏼

rmweir (approved):
LGTM

Issue #35694, opened 2021-11-30
Adding support for ingressClassName in Helm installer's Values file


kylep:
**Is your feature request related to a problem? Please describe.**

When Rancher is installed using the Helm method as documented in the [install guide](https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/), the Ingress resource it creates has no `ingressClassName` value in its spec. 

Without this value in the Ingress, if you installed a fresh cluster with `kubeadm` then add the Nginx ingress (`helm install nginx-ingress  nginx-stable/nginx-ingress`), the Rancher page cannot be accessed. Navigating to the Ingress URL returns an HTTP 404 response. The Ingress resource starts working only after injecting the ingressClassName value.

**Describe the solution you'd like**
I'd like to submit a PR that adds a jinja2 section to `ingress.yaml` where if `.Values.ingressClassName` is set in `values.yaml`, `ingressClassName: {{ .Values.ingressClassName }}` will be set inside the `spec:` section. This will provide users with a mechanism to define which Ingress class they would like Rancher to use. This would also be useful when multiple Ingress classes exist so that users could choose between them.

**Describe alternatives you've considered**
The obvious alternative is to patch that value into the Ingress after installing with Helm, but the value will likely get removed by Helm later.

**Additional context**
This issue was discussed and the solution was largely provided by /u/Odonay on this Reddit thread. https://www.reddit.com/r/rancher/comments/qw9nho/is_the_helm_installer_broken/hm6tqm4/ 
I've verified the fix and look forward to opening a PR.

zkl94:
Can we have a follow-up on this issue? Rancher seems like the only helm chart I am using that still lacks support for ingressClassName.

PR #35698, opened 2021-11-30
Issue-35694: Support ingressClassName in Helm Values

kylep:
Relating to issue [GitHub Issue #35694](https://github.com/rancher/rancher/issues/35694) as identified in [this reddit thread](https://www.reddit.com/r/rancher/comments/qw9nho/is_the_helm_installer_broken/)

This PR adds 3 lines of Jinja2 template to the ingress.yaml spec so users may specify a `ingressClassName` in their `values.yaml` file or with `--set`, for example `ingressClassName: nginx` or `--set ingressClassName=nginx`.

This change remedies an issue where the Ingress created by `helm install` returns 404's when no default IngressClass is found by the Ingress resource. This occurs on a fresh `kubeadm` install that uses the `Nginx` ingress controller.

This also enables support for users choosing which of multiple possible Ingress classes to use when installing
with Helm.

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -36,6 +36,9 @@ metadata:
 {{ toYaml .Values.ingress.extraAnnotations | indent 4 }}
 {{- end }}
 spec:
+{{- if .Values.ingress.className }}
+  ingressClassName: {{ .Values.ingress.className }}
+{{- end }}
   rules:
   - host: {{ .Values.hostname }}  # hostname to access rancher server
     http:
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -64,6 +64,9 @@ ingress:
   # configurationSnippet: |
   #   more_set_input_headers "X-Forwarded-Host: {{ .Values.hostname }}";
 
+  # className can be set to specify which IngressClass will be used by the created Ingress
+  className: ""
+
   tls:
     # options: rancher, letsEncrypt, secret
     source: rancher


chart/templates/ingress.yaml
```diff
@@ -36,6 +36,9 @@ metadata:
 {{ toYaml .Values.ingress.extraAnnotations | indent 4 }}
 {{- end }}
 spec:
+{{- if .Values.ingressClassName }}
+  ingressClassName: {{ .Values.ingressClassName }}
```
Oats87, line 40: Rather than using a top-level variable called `ingressClassName`, I think it would make more sense to nest this under the `ingress` field i.e. https://github.com/rancher/rancher/blob/release/v2.6/chart/values.yaml#L55-L70

Also, if we could figure out a good "default value" to populate into `values.yaml` (even if it is "blank") for documentation purposes, I think that would be great as well.

Oats87 (changes requested):

chart/templates/ingress.yaml
kylep: Thanks for the quick feedback! 

Can definitely change `.Values.ingressClassName ` to `.Values.ingress.ingressClassName`. 

I'm not sure if a usable default value is possible since for example Nginx's Ingress uses `nginx` as its value and other Ingress classes probably also use their own name.  Would setting it to null or having it as a comment in the `values.yaml` file be better? Or both so it's a null value with a comment explaining its purpose above it?

thedadams: Keeping it as an empty string by default is my suggestion. I believe that the `if` check here would fail on an empty string and the `ingressClassName` would be left out.

kylep: Sounds good, I'll verify it right now

kylep: Ok so if the user's values.yaml file has
```
ingress:
  ingressClassName: nginx
```

then the `ingressClassName` gets injected as expected. If it is not included, the empty string does indeed cause the `if` check to fail and so it behaves as it did before.

I also added a comment, hope that's ok

chart/values.yaml
```diff
@@ -64,6 +64,9 @@ ingress:
   # configurationSnippet: |
   #   more_set_input_headers "X-Forwarded-Host: {{ .Values.hostname }}";
 
+  # ingressClassName can be set to specify which IngressClass will be used by the created Ingress
+  ingressClassName: ""
```
thedadams, line 68: Change this to `className` so when it is used it will be `ingress.className` and won't "stutter"

thedadams (changes requested):

chart/values.yaml
kylep: Sure thing!

thedadams:
@kylep Looks good. Could you please squash the commits to one and force push?

kylep:
@ thedadams done, I think.

aslafy-z:
Implemented by https://github.com/rancher/rancher/pull/38202
This one should be closed.

Issue #35101, opened 2021-10-12
Replace downloaded K3s binary with its Docker image in the Dockerfile


maxsokolovsky:
SURE-3522
A part of #[33850](https://github.com/rancher/rancher/issues/33850).

Use the `FROM rancher/k3s` instruction in the Dockerfile instead of pulling in the downloaded K3s binary.

slickwarren:
test cases: 
* fresh install (docker) 2.6-head
* docker install 2.6.3 -> upgrade to 2.6-head
* docker install 2.6.3 using volume mounts -> upgrade to 2.6-head
* Standard cluster operations: creating and deleting deployments, services, jobs, secrets, namespaces.
  * in local cluster
  * in downstream cluster

sgapanovich:
### Test Environment: ###
Rancher version: different (see below)
Rancher cluster type: single-node docker install / HA
Docker version: 20.10

Downstream cluster type: n/a
Downstream K8s version: n/a

----------

### Testing: ###

Fresh docker install 2.6-head d540651
-provisioned a rke1 node driver
-local and dowstream cluster manual testing:
--can create/edit/delete deployment, cronjob, service (checked networking), secret, namespace
--can create/delete pod
--can create fleet deployment and services
--can add a custom chart installation (with deployment and services)

Rancher 2.6.3 with 2 downstream clusters (rke1 node driver and a custom cluster)
-ran automation
-upgraded to 2.6-head a140dab
-ran automation
-rolled back to 2.6.3
-ran automation

Rancher 2.6.3 with volume mount and 1 downstream cluster (rke1 node driver)
-ran automation
-upgraded to 2.6-head 3f30288
-ran automation

PR #35699, opened 2021-11-30
[DNM 2.6.4] Replace downloaded K3s binary with rancher/k3s Docker image

genexpr:
SURE-3522
[GitHub Issue](https://github.com/rancher/rancher/issues/35101)

Currently, the Dockerfiles download K3s as a binary that is a self-extracting archive from the main K3s repo. When a container is run, the binary starts extracting various files it needs at runtime (kubectl, cni and containerd dependencies, for example).

This PR modifies the Dockefiles, so that at build time they download the already extracted binaries from a dedicated K3s image, rather than the archive. The unnecessary extraction is then eliminated.


Dockerfile.dapper
```diff
@@ -62,19 +59,59 @@ RUN mkdir /usr/tmp && \
     mv /usr/tmp/helm /usr/bin/helm_v3 && \
     chmod +x /usr/bin/kustomize
 
-RUN curl -sLf ${!K3S_DOWNLOAD_URL} -o /usr/bin/k3s && \
-    chmod +x /usr/bin/k3s && \
-    mkdir -p /go/src/github.com/rancher/rancher/.kube && \
+# Set up K3s: copy the necessary binaries from an archive with extracted version of K3s.
```
brandond, line 62:
```suggestion
# Set up K3s: copy the necessary binaries from the k3s image.
```

Dockerfile.dapper
```diff
@@ -62,19 +59,59 @@ RUN mkdir /usr/tmp && \
...
+    /bin/containerd-shim-runc-v2 \
+    /bin/ethtool \
+    /bin/fuse-overlayfs \
+    /bin/ip \
+    /bin/ipset \
+    /bin/losetup \
+    /bin/pigz \
+    /bin/runc \
+    /bin/slirp4netns \
+    /bin/socat \
```
brandond, line 79: we can actually drop this one; it hasn't been needed since we switched to containerd 1.4

Dockerfile.dapper
```diff
@@ -62,19 +59,59 @@ RUN mkdir /usr/tmp && \
...
+    /bin/containerd \
+    /bin/containerd-shim-runc-v2 \
+    /bin/ethtool \
+    /bin/fuse-overlayfs \
+    /bin/ip \
+    /bin/ipset \
+    /bin/losetup \
+    /bin/pigz \
+    /bin/runc \
+    /bin/slirp4netns \
```
brandond, line 78: you can drop this; it's only needed for rootless

Dockerfile.dapper
```diff
@@ -62,19 +59,59 @@ RUN mkdir /usr/tmp && \
...
+COPY --from=rancher/k3s:v1.22.3-k3s1 \
+    /bin/blkid \
+    /bin/charon \
+    /bin/check-config \
+    /bin/cni \
+    /bin/conntrack \
+    /bin/containerd \
+    /bin/containerd-shim-runc-v2 \
+    /bin/ethtool \
+    /bin/fuse-overlayfs \
```
brandond, line 72: you can drop this; it's only needed for rootless

Dockerfile.dapper
```diff
@@ -62,19 +59,59 @@ RUN mkdir /usr/tmp && \
...
     chmod +x /usr/bin/kustomize
 
-RUN curl -sLf ${!K3S_DOWNLOAD_URL} -o /usr/bin/k3s && \
-    chmod +x /usr/bin/k3s && \
-    mkdir -p /go/src/github.com/rancher/rancher/.kube && \
+# Set up K3s: copy the necessary binaries from an archive with extracted version of K3s.
+COPY --from=rancher/k3s:v1.22.3-k3s1 \
+    /bin/blkid \
+    /bin/charon \
+    /bin/check-config \
```
brandond, line 66: you can drop this; it's only needed for the `k3s check-config` command which won't be used inside the rancher image

brandond (changes requested):

Dockerfile.dapper
genexpr: @brandond, are these and others above not needed in `package/Dockerfile` as well?

genexpr: Also, remove `containerd`? But k3s binaries (k3s, k3s-agent, etc.) are linked to it...

brandond: things removed from here can also be removed from package/Dockerfile.

don't remove containerd, just the specific lines that I left comments on:
```
/bin/check-config
/bin/fuse-overlayfs
/bin/slirp4netns
/bin/socat
```

Dockerfile.dapper
```diff
@@ -62,19 +59,59 @@ RUN mkdir /usr/tmp && \
...
+
+RUN mkdir -p /go/src/github.com/rancher/rancher/.kube && \
     ln -s /etc/rancher/k3s/k3s.yaml /go/src/github.com/rancher/rancher/.kube/k3s.yaml
 
 RUN curl -sLf https://github.com/rancher/k3s/releases/download/${CATTLE_K3S_VERSION}/k3s-images.txt -o /usr/tmp/k3s-images.txt
 
 ENV YQ_URL=https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_${ARCH}
 RUN curl -sLf ${YQ_URL} -o /usr/bin/yq && chmod +x /usr/bin/yq
 
-ENV KUBECTL_URL=https://storage.googleapis.com/kubernetes-release/release/v1.22.3/bin/linux/${ARCH}/kubectl
```
annablender, line 75: If I understand this correctly, the kubectl binary can be removed here because it's packaged in the k3s image. Did you test that it still works?

package/Dockerfile
```diff
@@ -101,13 +98,56 @@ RUN curl ${HELM_URL_V3} | tar xvzf - --strip-components=1 -C /usr/bin && \
     mv /usr/bin/helm /usr/bin/helm_v3 && \
     chmod +x /usr/bin/kustomize
 
+# Set up K3s: copy the necessary binaries from an archive with extracted version of K3s.
```
annablender, line 101: Update like @brandond mentioned

Dockerfile.dapper
genexpr: That's right. It works. I will test again with a few other binaries removed and also test using kubectl directly inside the container.

annablender: Ok great! Can you also check you can ping embedded etcd?

annablender: You can do that using this doc https://github.com/rancher/rancher/wiki/How-to-Use-Etcdctl-with-K3s-Embedded-Etcd. So basically run this cmd

etcdctl --endpoints=https://localhost:2379 \
          --cert="/var/lib/rancher/k3s/server/tls/etcd/server-client.crt" \
          --key="/var/lib/rancher/k3s/server/tls/etcd/server-client.key" \
          --cacert="/var/lib/rancher/k3s/server/tls/etcd/server-ca.crt" member list

And make sure you get a cluster ID and details.

Issue #35680, opened 2021-11-29
Old Catalog API needs to be updated so versionLinks returns the version of the app that is currently installed


deniseschannon:
**Rancher Server Setup**
- Rancher version: v2.6.3 -rc

**Describe the bug**

When a catalog has been maxed out so it's no longer available in any Rancher version, there are issues throughout the UI that relies on the versionLinks API to return what versions the app can be deployed. 

This is fine and expected for fresh installs of Rancher or when we want to figure out which app is eligible to deploy. 

The downside is that when a user has come from an upgraded Rancher setup and already had an app deployed, that version should always be available to continue "upgrading". Currently, there are a various UI errors occurring as the current versionLinks for that app returns no available version.

The proper fix is to have versionLinks return the version of the app if it's currently deployed as a viable upgrade option. 

**To Reproduce**
 
1. Launch 2.5.11
2. Deploy an app from the Cluster Manager UI -> Using the Apps tab within a Project view
3. Upgrade Rancher to 2.6.3-rc2. Note: The legacy feature flag should automatically be enabled
4. Navigate to the Cluster view and inside Legacy section, there is an Apps section.
5. Try to upgrade the existing app 

Note: There are UI issues filed, but this is ultimately an API issue
https://github.com/rancher/dashboard/issues/4628
https://github.com/rancher/dashboard/issues/4627
https://github.com/rancher/dashboard/issues/4655

**Result**

Unable to properly upgrade to the same version of the chart

**Expected Result**

Ability to continue updating to the original version of a deployed charts regardless of min/max version on the chart

**Screenshots**

**Additional context**

In 2.6.3, we are aiming to max out all charts in the old catalogs (library and helm3-library) as most of these are not tested and will not be valid in k8s 1.22. Also, we are moving to new apps&marketplace experience and want to stop allowing people to install charts through Rancher in this method. 

Note: Longhorn doesn't have a max version on it's chart yet which is why it still works.

anupama2501:
Verified on rancher v2.6-head 2a7857d upgraded from 2.5.11
1. Created a downstream linode RKE1 on 2.5.11 with 3 worker,1 etcd and 1 cp
2. From the default project --> installed the apps `docker-registry`,`redis`, `mongodb-replicaset`
Update apps
1. Updated the app by modifying the password of the redis app
2. Verified the update went through without any errors

Upgrade apps
1. Upgraded the version of docker-registry to version `1.9.2` 
2. Verified upgrade went through without any errors.

[...]

PR #35711, opened 2021-12-01
Add currentVersion query parameter for templates

rmweir:
**Problem:**
Versions that are not compatible with current
setup but are already in use are being filtered out. This is required to facilitate the UI, otherwise the version is treated
as though the version doesn't exist, despite being in use.

**Solution:**
A currentVersion query parameter can now be passed when making
templates requests to norman. This parameter will cause any
version matching it to be included in versionLinks. This is
being done so versions that are not compatible with current
setup but are already in use do not have to be filtered out.
This is required to facilitate the UI, otherwise it is treated
as though the version doesn't exist, despite being in use.

**Issue:**
https://github.com/rancher/rancher/issues/35680


brendarearden (approved):
LGTM

vincent99 (approved):

Issue #35646, opened 2021-11-21
Unavailable node is seen Active in Cluster Management --> Machine Pools page


sowmyav27:
**On 2.6-head commit id: `e322f8f`**

- Deploy an RKE1 cluster (I had an 8 node cluster - 3etcd, 2 control and 3 worker nodes and powered one etcd down)
- Power down one of the nodes in the cluster
- the Node in Cluster explorer --> nodes page shows Unavailable
- Machine pool is seen Active.
- Click on the node --> Node is seen Unavailable

**Expected**
Machine pool should not be in Active state

kinarashah:
The issue is stemming from the difference between how norman and steve transform fields for `v3/node (management.cattle.io.nodes)` at api level. 

**Norman**
When we look at v3/node in norman api, conditions from internalNodeStatus get combined with node's status conditions so we can see the `Ready` condition shows up correctly with `Unknown` status for the node - 

<img width="467" alt="Screen Shot 2021-12-01 at 10 34 41 AM" src="https://user-images.githubusercontent.com/16870565/144293299-f83435c2-9039-4ac0-9375-680f9a532534.png">

**Steve**
We don't have transformation logic similar to norman, so there are two different conditions on the v3 node object - 
1. Node status conditions

<img width="423" alt="Screen Shot 2021-12-01 at 10 36 27 AM" src="https://user-images.githubusercontent.com/16870565/144293860-f6d4d973-f7a5-4bc6-b4b3-9a1432e379bd.png">

2. Internal node status conditions which has the Ready condition update 

<img width="565" alt="Screen Shot 2021-12-01 at 10 36 35 AM" src="https://user-images.githubusercontent.com/16870565/144293940-15c03b2a-3ee5-4a12-87b6-e0ac4f4ebcb8.png">

sowmyav27:
**On 2.6-head commit id: `e4db3ee`**

- verified on a custom cluster and an RKE1 DO cluster
- Powered down the node in the backend cloud provider console
- The node showed up as Unavailable in machine Pools page

<img width="845" alt="Screen Shot 2021-12-07 at 5 30 40 PM" src="https://user-images.githubusercontent.com/26032343/145132451-28d127b2-d4c6-40d5-bdf5-412139c5de90.png">
<img width="816" alt="Screen Shot 2021-12-07 at 5 32 47 PM" src="https://user-images.githubusercontent.com/26032343/145132478-b5ae4f72-bb74-4d3d-90ef-28bc331d0d47.png">

PR #35712, opened 2021-12-02
sync internal node ready condition to v3 node condition

kinarashah:
**Issue:**
When powered down, node shows Unavailable from Cluster explorer view but shows up as Active from cluster management view. 

**Solution**
Norman embeds internalNodeStatus.Conditions to node.status.Conditions which causes the ready condition to show up correctly on v3 calls. Syncing it to v3 node.status.conditions for steve v1 node calls to get the same effect. 

https://github.com/rancher/rancher/issues/35646

### DIFF
--- a/pkg/controllers/managementuser/nodesyncer/nodessyncer.go
+++ b/pkg/controllers/managementuser/nodesyncer/nodessyncer.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	cond "github.com/rancher/norman/condition"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	kd "github.com/rancher/rancher/pkg/controllers/management/kontainerdrivermetadata"
 	"github.com/rancher/rancher/pkg/controllers/managementagent/podresources"
@@ -584,6 +585,7 @@ func objectsAreEqual(existing *v3.Node, toUpdate *v3.Node) bool {
 	toUpdateToCompare := resetConditions(toUpdate)
 	existingToCompare := resetConditions(existing)
 	statusEqual := statusEqualTest(toUpdateToCompare.Status.InternalNodeStatus, existingToCompare.Status.InternalNodeStatus)
+	conditionsEqual := reflect.DeepEqual(toUpdateToCompare.Status.Conditions, existing.Status.Conditions)
 	labelsEqual := reflect.DeepEqual(toUpdateToCompare.Status.NodeLabels, existing.Status.NodeLabels) && reflect.DeepEqual(toUpdateToCompare.Labels, existing.Labels)
 	annotationsEqual := reflect.DeepEqual(toUpdateToCompare.Status.NodeAnnotations, existing.Status.NodeAnnotations)
 	specEqual := reflect.DeepEqual(toUpdateToCompare.Spec.InternalNodeSpec, existingToCompare.Spec.InternalNodeSpec)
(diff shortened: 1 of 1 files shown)


pkg/controllers/managementuser/nodesyncer/nodessyncer.go
```diff
@@ -593,11 +595,11 @@ func objectsAreEqual(existing *v3.Node, toUpdate *v3.Node) bool {
 	rolesEqual := toUpdateToCompare.Spec.Worker == existingToCompare.Spec.Worker && toUpdateToCompare.Spec.Etcd == existingToCompare.Spec.Etcd &&
 		toUpdateToCompare.Spec.ControlPlane == existingToCompare.Spec.ControlPlane
 
-	retVal := statusEqual && specEqual && nodeNameEqual && labelsEqual && annotationsEqual && requestsEqual && limitsEqual && rolesEqual
+	retVal := statusEqual && conditionsEqual && specEqual && nodeNameEqual && labelsEqual && annotationsEqual && requestsEqual && limitsEqual && rolesEqual
 	if !retVal {
-		logrus.Debugf("ObjectsAreEqualResults for %s: statusEqual: %t specEqual: %t"+
+		logrus.Debugf("ObjectsAreEqualResults for %s: statusEqual: %t conditionsEqual %t specEqual: %t"+
```
snasovich, line 600: Missing `:` after `conditionsEqual`

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
```diff
@@ -676,6 +678,24 @@ func cleanStatus(machine *v3.Node) {
...
+				LastTransitionTime: condition.LastTransitionTime.String(),
+				LastUpdateTime:     condition.LastHeartbeatTime.String(),
+				Reason:             condition.Reason,
+				Message:            condition.Message,
+			}
+			var exists bool
+			for i, cond := range machine.Status.Conditions {
+				if cond.Type == "Ready" {
+					exists = true
+					machine.Status.Conditions[i] = readyCondition
```
snasovich, line 693: Can we break out of loop here as we've already found `Ready` condition? Not sure if it's theoretically possible to have multiple conditions with the same type.

snasovich (changes requested):
Couple minor requests/questions.

However, on the overall approach - we're copying Ready condition from `internalNodeStatus` up to root status for `v3.Node` objects. Where does the syncing of conditions to `v1.Node` happens? In here I only see `v1.Node` updated in `updateLabels` and `updateAnnotations` none of which would be triggered if conditions on the status is the only thing that has changed. Does v3 -> v1 syncing happens elsewhere?

kinarashah:
@snasovich Thanks for the review, will update review comments. 
About syncing to v1 node, I don't understand why we need to sync it to v1 from v3 node? Here v3 is rancher's management node and v1 is core k8s downstream node. From cluster management view, UI relies on the v3 management node's conditions so we sync the ready condition from k8s core v1 node to v3 management node. Users have the option set labels and annotations on v3 management node, so we need to sync them to k8s core v1 downstream node.

snasovich:
@kinarashah , thank you for the clarifications. 👍 
There was a misunderstanding on my end. I figured UI shows nodes based on the response from `v1/management.cattle.io.nodes` API call and automatically assumed we have `v1` version of Node object somewhere in Rancher CRDs. However that API call returns Node objects from `management.cattle.io/v3` so we're good. All these `v1`s / `v3`s are a bit confusing sometimes.

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
kinarashah: done

snasovich (approved):

PR #35723, opened 2021-12-02
Replace downloaded K3s binary with rancher/k3s Docker image

genexpr:
SURE-3522
[GitHub Issue](https://github.com/rancher/rancher/issues/35101)

Currently, the Dockerfiles download K3s as a binary that is a self-extracting archive from the main K3s repo. When a container is run, the binary starts extracting various files it needs at runtime (kubectl, cni and containerd dependencies, for example).

This PR modifies the Dockefiles, so that at build time they download the already extracted binaries from a dedicated K3s image, rather than the archive. The unnecessary extraction is then eliminated.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,7 +4,7 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
 ENV CATTLE_HELM_VERSION v2.16.8-rancher1
 ENV CATTLE_MACHINE_VERSION v0.15.0-rancher73
-ENV CATTLE_K3S_VERSION v1.21.7+k3s1
+ENV CATTLE_K3S_VERSION v1.22.5+k3s1
 # version used by helm plugin install script
 ENV CATTLE_HELM_UNITTEST_VERSION v0.1.7-rancher3
 # helm 3 version
@@ -40,9 +40,6 @@ ENV HELM_URL_V2_amd64=https://github.com/rancher/helm/releases/download/${CATTLE
     TILLER_URL_amd64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-tiller \
     TILLER_URL_arm64=https://github.com/rancher/helm/releases/download/${CATTLE_HELM_VERSION}/rancher-tiller-arm64 \
     TILLER_URL=TILLER_URL_${ARCH} \
-    K3S_DOWNLOAD_URL_amd64=https://github.com/rancher/k3s/releases/download/${CATTLE_K3S_VERSION}/k3s \
-    K3S_DOWNLOAD_URL_arm64=https://github.com/rancher/k3s/releases/download/${CATTLE_K3S_VERSION}/k3s-arm64 \
-    K3S_DOWNLOAD_URL=K3S_DOWNLOAD_URL_${ARCH} \
     KUSTOMIZE_URL=https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_${ARCH}.tar.gz
 
 RUN curl -sLf ${KUSTOMIZE_URL} | tar -xzf - -C /usr/bin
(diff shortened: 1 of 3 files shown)


annablender (approved):
LGTM, very nice job!

Dockerfile.dapper
```diff
@@ -62,19 +59,55 @@ RUN mkdir /usr/tmp && \
     mv /usr/tmp/helm /usr/bin/helm_v3 && \
     chmod +x /usr/bin/kustomize
 
-RUN curl -sLf ${!K3S_DOWNLOAD_URL} -o /usr/bin/k3s && \
-    chmod +x /usr/bin/k3s && \
-    mkdir -p /go/src/github.com/rancher/rancher/.kube && \
+# Set up K3s: copy the necessary binaries from the K3s image.
+COPY --from=rancher/k3s:v1.22.3-k3s1 \
```
brandond, line 63: Would be really nice to get this up to the latest release, since that's the version of K3s that 2.6.3's KDM will recommend.
```suggestion
COPY --from=rancher/k3s:v1.22.4-k3s1 \
```

package/Dockerfile
```diff
@@ -101,13 +98,52 @@ RUN curl ${HELM_URL_V3} | tar xvzf - --strip-components=1 -C /usr/bin && \
     mv /usr/bin/helm /usr/bin/helm_v3 && \
     chmod +x /usr/bin/kustomize
 
+# Set up K3s: copy the necessary binaries from the K3s image.
+COPY --from=rancher/k3s:v1.22.3-k3s1 \
```
brandond, line 102: same as above
```suggestion
COPY --from=rancher/k3s:v1.22.4-k3s1 \
```

brandond (commented):
last minute version bump!

brandond (approved):

Dockerfile.dapper
```diff
@@ -62,19 +59,55 @@ RUN mkdir /usr/tmp && \
...
+    /bin/containerd-shim-runc-v2 \
+    /bin/ethtool \
+    /bin/ip \
+    /bin/ipset \
+    /bin/losetup \
+    /bin/pigz \
+    /bin/runc \
+    /bin/swanctl \
+    /bin/which \
+    /bin/aux/xtables-legacy-multi \
```
Oats87, line 78: Do we actually need the `iptables`/`xtables` binaries for K3s to run within the Rancher container? 

I have a feeling the explicit use of the legacy binaries may actually cause problems if on a system where the `ip_tables` module is not enabled.

If we actually need the `iptables` binaries, I would recommend just copying the entire `aux` folder out of `/bin/aux` into the `rancher` image so as to preserve the ability to perform iptables module detection (`ip_tables` vs `nftables`) which we do today.

Dockerfile.dapper
```diff
@@ -62,19 +59,55 @@ RUN mkdir /usr/tmp && \
     mv /usr/tmp/helm /usr/bin/helm_v3 && \
     chmod +x /usr/bin/kustomize
 
-RUN curl -sLf ${!K3S_DOWNLOAD_URL} -o /usr/bin/k3s && \
-    chmod +x /usr/bin/k3s && \
-    mkdir -p /go/src/github.com/rancher/rancher/.kube && \
+# Set up K3s: copy the necessary binaries from the K3s image.
+COPY --from=rancher/k3s:v1.22.4-k3s1 \
```
Oats87, line 63: For maximum futureproofing, I would actually think we might want to directly consume the `k3s-root` archive that is where these userspace binaries all come from in the first place. 

Either that, or we could copy the entire `/bin` of the `rancher/k3s` image into a folder like `/usr/bin/k3s-bin` and then set the path in the container to `PATH=$PATH:/usr/bin/k3s-bin` and would have a lot less tech debt regarding binaries we have to maintain. 

There's a whole bunch of convuluted logic in place but more or less it all comes from https://github.com/k3s-io/k3s-root

https://github.com/k3s-io/k3s/blob/master/package/Dockerfile#L3-L5

I suppose we might only actually want a subset of the binaries -- I would be curious to see how you came to determine these are the specific binaries we need in the `rancher/rancher` image (I might just be missing it)

Oats87 (changes requested):
Overall, this optimization looks great; however, I have a few concerns around the "source" of the binaries/placement/what binaries are actually consumed.

Dockerfile.dapper
brandond: Copying symlinks out of Docker images is kinda broken; it copies the file instead of remaining a symlink. We ended up doing this to keep this as a symlink.

Dockerfile.dapper
brandond: That's an interesting idea, but I was trying to get them away from downloading artifacts in favor of `COPY --from` for everything.

Dockerfile.dapper
genexpr: I have been a little concerned about which iptables-related packages to copy over. Currently, in `/bin/aux/`, I see lots of `ip6tables*`, `arptables*`, `ebtables*` symlinks. What if some are needed? In my testing, I have not seen any errors with just having a few select `iptables` and `ip6tables` binaries, but maybe in some scenarios K3s might actually need one that's not been copied over?

Dockerfile.dapper
```diff
@@ -4,7 +4,7 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
 ENV CATTLE_HELM_VERSION v2.16.8-rancher1
 ENV CATTLE_MACHINE_VERSION v0.15.0-rancher73
-ENV CATTLE_K3S_VERSION v1.21.7+k3s1
+ENV CATTLE_K3S_VERSION v1.22.4+k3s1
```
brandond, line 7: time to bump again: 
```suggestion
ENV CATTLE_K3S_VERSION v1.22.5+k3s1
```

Dockerfile.dapper
brandond, line 63:
```suggestion
COPY --from=rancher/k3s:v1.22.5-k3s1 \
```

package/Dockerfile
```diff
@@ -27,7 +27,7 @@ ENV CATTLE_CHART_DEFAULT_BRANCH=$CHART_DEFAULT_BRANCH
 ENV CATTLE_PARTNER_CHART_DEFAULT_BRANCH=$PARTNER_CHART_DEFAULT_BRANCH
 ENV CATTLE_RKE2_CHART_DEFAULT_BRANCH=$RKE2_CHART_DEFAULT_BRANCH
 ENV CATTLE_HELM_VERSION v2.16.8-rancher1
-ENV CATTLE_K3S_VERSION v1.21.7+k3s1
+ENV CATTLE_K3S_VERSION v1.22.4+k3s1
```
brandond, line 30:
```suggestion
ENV CATTLE_K3S_VERSION v1.22.5+k3s1
```

package/Dockerfile
```diff
@@ -27,7 +27,7 @@ ENV CATTLE_CHART_DEFAULT_BRANCH=$CHART_DEFAULT_BRANCH
...
 ENV CATTLE_MACHINE_VERSION v0.15.0-rancher73
 ENV CATTLE_MACHINE_PROVISION_IMAGE rancher/machine:${CATTLE_MACHINE_VERSION}
 ENV CATTLE_ETCD_VERSION v3.4.13
```
brandond, line 33: K3s 1.22 actually embeds etcd 3.5, so we should probably package a similar version of etcdctl
```suggestion
ENV CATTLE_ETCD_VERSION v3.5.1
```

brandond (changes requested):

brandond (approved):

genexpr:
The CI build has been failing consistently for a few days. Not seeing any test failures, but the exit code is 1. The build stages of linux-amd64 and linux-arm64 do have errors in the logs during the Docker image build. Step 27/30 - installing systemd:
`Failed to connect to bus: Host is down`

I've seen this in builds of other people's PRs, so not sure what's going on.

brandond (approved):

genexpr:
@annablender, @Oats87, could you please review when you have a minute?

annablender (commented):
I'd recommend consolidating your changes into 2-3 commits and rebasing with v2.6 so you don't pepper the 2.6 history with merges

I also can't comment on specific lines here. Not sure why

Dockerfile.dapper
```diff
@@ -4,7 +4,7 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
 ENV CATTLE_HELM_VERSION v2.16.8-rancher1
 ENV CATTLE_MACHINE_VERSION v0.15.0-rancher73
-ENV CATTLE_K3S_VERSION v1.22.3+k3s1
+ENV CATTLE_K3S_VERSION v1.22.4+k3s1
```
annablender: Could you help me understand why the k3s version was bumped from 1.22.3, which was where I left off with embedded etcd, to 1.22.5 and back to 1.22.4. What changed?

genexpr: I see it as a bump from 1.22.3 to 1.22.4 to 1.22.5. Not seeing where it was set to 1.22.4 again...

genexpr:
@annablender, yes, I don't plan to merge all those commits without squashing in GitHub first (there is a Squash and merge option on the Merge button). I'll see if I can just reset the branch and only include my changes in a single commit...

Dockerfile.dapper
brandond: @annablender I think you're reviewing one of the intermediate commits. The latest has it bumped to 1.22.5.

brandond:
Looks like we need another rebase to address conflicts. Looks good otherwise.

brandond (approved):

annablender (approved):
Awesome, LGTM

Issue #35696, opened 2021-11-30
When rke2 feature flag is disabled, upgrade to 2.6-head ruins downstream rke2 cluster


ronhorton:
**Rancher Server Setup**
- Rancher version: `2.6-head Commit ID 4375161`
- Installation option (Docker install/Helm Chart): Docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): `RKE2 - DO - v1.21.6+rke2r1`

- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: `v1.21.6+rke2r1`
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? `Downstream RKE2 Digital Ocean cluster`

**Describe the bug**
Created an rke2 cluster in 2.6.2 and deactivate the rke2 feature flag. Upgraded to 2.6-head, rke2 feature flag was set to active, but the downstream cluster goes to an error state. 

**To Reproduce**
1. Create a downstream rke2 cluster in `2.6.2`
2. click `Hamburger menu > global settings > feature flags`
3. Disable rke2 feature flag
4. upgrade to 2.6-head

**Result**
The rke2 feature flag is changed to `active` and locked (ie, cannot disable); the rke2 cluster is in an error state and cannot be recovered. 
**Expected Result**
Rke2 feature flag is set to active and the downstream cluster is active and available. 

**Screenshots**
<img width="969" alt="Screen Shot 2021-11-30 at 12 39 10 PM" src="https://user-images.githubusercontent.com/89212856/144111403-2a352b83-3ae8-4ef1-93ed-c9c55985452c.png">
<img width="1013" alt="Screen Shot 2021-11-30 at 12 38 49 PM" src="https://user-images.githubusercontent.com/89212856/144111416-26f8fde5-3a8c-441d-9a62-2711de1ae274.png">
<img width="1020" alt="Screen Shot 2021-11-30 at 12 50 04 PM" src="https://user-images.githubusercontent.com/89212856/144111594-ec861004-777d-44a8-9ee1-51a9613bf120.png">

thedadams:
@ronhorton Does this eventually rectify itself? The webhook being unavailable is something we have seen before.

thedadams:
I am unable to reproduce this. I tested this with a local dev setup going from v2.6.2 -> 2.6.3-rc3. I also tested this with HA Rancher installed via helm on a K3S cluster. In both cases I:
[...]

PR #35746, opened 2021-12-03
Check rancher-webhook version on Docker container startup

thedadams:
If upgrading a Docker install to Rancher v2.6.3+, then the k3s cluster
that runs in the Docker container has version 1.22+. The issue is that
the rancher-webhook v0.2.1 or less doesn't support Kubernetes 1.22+.
Rancher will try to upgrade rancher-webhook on startup, but it will
fail in this case because the validation webhook (that is called when
the service account is created for the helm pod) fails because the
currently installed rancher-webhook crashed. Therefore, in the Docker
install scenario, rancher-webhook has to be upgrade to a supported
version for Kubernetes 1.22+

Issues:
https://github.com/rancher/rancher/issues/35696
https://github.com/rancher/rancher/issues/35763

### DIFF
--- a/pkg/rancher/rancher.go
+++ b/pkg/rancher/rancher.go
@@ -5,7 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
+	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


Oats87 (dismissed):

Oats87 (approved):

paynejacob:
it feels a little strange editing a helm resource directly but given the chicken and the egg problem it makes sense.

paynejacob (approved):

Issue #35714, opened 2021-12-02
Unable to deploy an RKE2 cluster using v1.21.7-rc3+rke2r1 version


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `46eb9d4`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: `v1.21.7-rc3+rke2r1`
- Cluster Type (Local/Downstream): RKE2 EC2

**Describe the bug**
Unable to deploy an RKE2 cluster using v1.21.7-rc3+rke2r1 version

**To Reproduce**

- Deploy an ec2 cluster - 1 node all roles, 1 node worker node
- The worker node is stuck in `Nodeconditionsfailed` state.
- Worker node logs - `journalctl -eu rke2-agent -f`:

```
Dec 02 01:12:26 sowmyaec2-pool2-cc1a6e42-wlv6c rke2[1589]: time="2021-12-02T01:12:26Z" level=info msg="Failed to update node sowmyaec2-pool2-cc1a6e42-wlv6c: Operation cannot be fulfilled on nodes \"sowmyaec2-pool2-cc1a6e42-wlv6c\": the object has been modified; please apply your changes to the latest version and try again"
Dec 02 01:12:26 sowmyaec2-pool2-cc1a6e42-wlv6c rke2[1589]: time="2021-12-02T01:12:26Z" level=info msg="labels have been set successfully on node: sowmyaec2-pool2-cc1a6e42-wlv6c"
Dec 02 01:12:26 sowmyaec2-pool2-cc1a6e42-wlv6c systemd[1]: Started Rancher Kubernetes Engine v2 (agent).
```

sowmyav27:
**on 2.6-head commit id: `e4db3ee`**

- Deploy an ec2 cluster - 1 node all roles, 1 node worker node using `v1.21.7-rc3+rke2r1` version
- Nodes come up Active on the cluster
- Cluster is also Active

PR #35765, opened 2021-12-05
Fix race on node Ready condition on provision

thedadams:
There is an intermittent issue when provisioning RKE2 clusters where a node may
seem to get stuck in "Ready is False" state. This is a red-herring because the
node is Ready, but the CAPI machine controller is not picking up this change.

In order to force the CAPI machine controller to see that the node is ready,
Rancher will wait to set the providerID on the infrastructure machine until the
Node is Ready. This ensures that the CAPI controller sees the Node in a ready
state.

Issue:
https://github.com/rancher/rancher/issues/35714

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machinenodelookup/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/machinenodelookup/controller.go
@@ -15,6 +15,7 @@ import (
 	"github.com/rancher/rancher/pkg/provisioningv2/rke2/planner"
 	"github.com/rancher/rancher/pkg/provisioningv2/rke2/runtime"
 	"github.com/rancher/rancher/pkg/wrangler"
+	"github.com/rancher/wrangler/pkg/condition"
 	"github.com/rancher/wrangler/pkg/data"
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


Oats87 (approved):

snasovich (approved):

jakefhyde (approved):

pkg/controllers/provisioningv2/rke2/machinenodelookup/controller.go
```diff
@@ -113,7 +114,7 @@ func (h *handler) associateMachineWithNode(_ string, bootstrap *rkev1.RKEBootstr
 
 	nodeLabelSelector := metav1.LabelSelector{MatchLabels: map[string]string{planner.MachineUIDLabel: string(machine.GetUID())}}
 	nodes, err := clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: labels.Set(nodeLabelSelector.MatchLabels).String()})
-	if err != nil || len(nodes.Items) == 0 || nodes.Items[0].Spec.ProviderID == "" {
+	if err != nil || len(nodes.Items) == 0 || nodes.Items[0].Spec.ProviderID == "" || !condition.Cond("Ready").IsTrue(nodes.Items[0]) {
```
jakefhyde, line 117: Just wondering out of curiosity, is there any reason why creating the condition here is preferred over using `NodeConditionReady`?

thedadams: This was the precedent that was set by Darren. I believe the reason is:
1. To keep `v3` separate from `v1`.
2. Related: `NodeContionReady` would be imported at `v3.NodeConditionReady`. So it would look strange to have the on a `v1` object.

PR #35768, opened 2021-12-05
Update size for do images

sgapanovich:
`test_cluster_template` tests fail due to incorrect image sizes passed ("4gb" and "8gb") and cluster would not provision

This PR updates automation to use correct DO images and make clusters provision


sowmyav27 (approved):
LGTM

Issue #35650, opened 2021-11-23
Not able to import rke1 clusters into the rancher server


anupama2501:
**Rancher Server Setup**
- Rancher version: `v2.6-head 44b8030 `
- Installation option (Docker install/Helm Chart): Docker install & HA
   - If Helm Chart, Kubernetes Info: k3s & RKE1 
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):
         Node Setup: 1 node 
         Version: `v1.21.3+k3s1` & `v1.22.3-rancher1-1`
- Proxy/Cert Details: `self-signed`

**Information about the Cluster**
- Kubernetes version:  `v1.21.6` , `v1.22.3` and `v1.18.20`
- Cluster Type (Local/Downstream): `Imported RKE1 cluster 3 nodes all roles`

**Describe the bug**
After importing an rke1 cluster into the rancher server, the cluster is stuck in `Unavailable` state and won't come up active

**To Reproduce**
1. Create an rke1 cluster with any k8s, I tried with k8s 1.22.3, 1.21.6 and 1.18.20
2. Run the curl command against the kubeconfig file of this cluster.

**Result**
Cluster is stuck in `unavailable` state for HA installation and on docker install, it is flapping between `Active` and `Unavailable` state.
Following error was seen on the cluster 
```
Cluster agent is not connected
```
From the cluster agent pod logs, we see the following:
```
time="2021-11-23T22:18:02Z" level=error msg="Failed to connect to proxy. Empty dialer response" error="dial tcp<redacted>:443: i/o timeout"
time="2021-11-23T22:18:02Z" level=error msg="Remotedialer proxy error" error="dial tcp <redacted>:443: i/o timeout"
time="2021-11-23T22:18:12Z" level=info msg="Connecting to wss://<redacted>/v3/connect with token starting with<redacted>"
time="2021-11-23T22:18:12Z" level=info msg="Connecting to proxy" **url="wss://<redacted>/v3/connect"**
```

**Expected Result**
Expected the cluster to come up active 

**Additional context**
Issue is not seen for imported RKE2 clusters

deniseschannon:
Most likely related to https://github.com/rancher/rancher/issues/35546

cbron:
We understand reason its breaking now, just need to find a new fix.

kinarashah:
[...]

PR #35775, opened 2021-12-06
redeploy cluster agent for imported clusters even if connected is false

kinarashah:
**Issue:** 
Not able to import rke1 clusters https://github.com/rancher/rancher/issues/35650

**Problem** 
For the first cluster agent deployed through template generated by cluster import handler, [CATTLE_FEATURES is empty ](https://github.com/rancher/rancher/blob/release/v2.6/pkg/api/norman/customization/clusterregistrationtokens/import.go#L49) so rancher isn't running. Without secret and aggregation controllers running, session with clientKey `"stv-cluster-{clusterName}"` never gets added for clusters.

hasSession will always remain false until cluster agent is redeployed (during redeploy, CATTLE_FEATURES is set) and then connected becomes true. But [we blocked redeploy if connected is false](https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/clusterdeploy/clusterdeploy.go#L96 ) in https://github.com/rancher/rancher/pull/35545, so that's why imported cluster never becomes active. 

**Solution**
Redeploy agent for imported clusters even when cluster condition connected is false. 

**Other solution**
Pass CATTLE_FEATURES for the first agent generated by cluster import handler during cluster registration. I didn't opt for this solution because I don't see any merit in starting to run rancher controllers for this agent when the cluster agent is going to get redeployed by rancher anyway. Also, this logic was recently added for 2.6.3 vs agent has been running without CATTLE_FEATURES since support for it was added. So inclined towards removing this code block.

### DIFF
--- a/pkg/controllers/management/clusterdeploy/clusterdeploy.go
+++ b/pkg/controllers/management/clusterdeploy/clusterdeploy.go
@@ -16,7 +16,6 @@ import (
 	"github.com/rancher/rancher/pkg/auth/tokens"
 	util "github.com/rancher/rancher/pkg/cluster"
 	"github.com/rancher/rancher/pkg/clustermanager"
-	"github.com/rancher/rancher/pkg/controllers/management/clusterconnected"
 	"github.com/rancher/rancher/pkg/features"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/image"
@@ -91,10 +90,6 @@ func (cd *clusterDeploy) sync(key string, cluster *v3.Cluster) (runtime.Object, 
 			return nil, err
 		}
 		return nil, nil
-	}
-
-	if cluster.Status.Driver == v32.ClusterDriverImported && clusterconnected.Connected.IsFalse(cluster) {
-		return cluster, nil
 	}
 
 	original := cluster


cmurphy (approved):

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -93,10 +92,6 @@ func (cd *clusterDeploy) sync(key string, cluster *v3.Cluster) (runtime.Object,
 		return nil, nil
 	}
 
-	if cluster.Status.Driver == v32.ClusterDriverImported && clusterconnected.Connected.IsFalse(cluster) {
-		return cluster, nil
-	}
-
```
thedadams, line 99: Should this check be moved to line 119 below so that the error message doesn't get set in this case? That seems to be the original intention of putting this check here (as the commit message says: https://github.com/rancher/rancher/commit/f82c084cef10ec73264473049055333ec007bbf1)

kinarashah: Thanks @thedadams for calling this out. It's a good point, the original intention was indeed to avoid showing any other errors when the cluster agent is disconnected. But if the agent deploy fails for some reason, this would mean we would return nil and not retry - in that case, we'd see this issue again where imported clusters won't get registered at all. So I experimented with adding AgentDeployed condition check to the removed block for returning: 

`if cluster.Status.Driver == v32.ClusterDriverImported && v32.ClusterConditionAgentDeployedIsTrue(cluster) && clusterconnected.Connected.IsFalse(cluster) {
		return cluster, nil
	}
`

This still blocks out agent redeploys on rancher server upgrades, so I think it'd be safest to leave this as is
- to not block any agent deploys, if agent deploy itself fails, we should still set error and retry 
- once agent deploy succeeds, if for any other reason cluster disconnects for any other reason, we now have a connected condition set to show the msg

thedadams: Thanks for the thorough investigation. What you are saying makes sense to me: if there is an error redeploying the agent, then Rancher should retry.

thedadams (approved):

Issue #35684, opened 2021-11-30
Installing the same helm3 chart in the ember UI after its deleted is updating the existing migrated chart in the explorer UI 


anupama2501:
Setup

Rancher version: 2.6-head 44b8030
Browser type & version: Chrome

**Describe the bug**
Deleting a migrated app from ember UI to explorer UI from the steps mentioned in this [issue](https://github.com/rancher/rancher/issues/34701) is deleting the app from ember UI which will not delete the chart from explorer UI. If we re-install the same app but with a different version, the chart in the explore UI is updated with this version thats installed on ember UI. 

**To Reproduce**
1. Create an rke1 node driver cluster 3 nodes 1 role each
2. From ember UI, install the `aws-event-sources` app in the default project version `v0.1.2`
3. Add the annotations to the app by executing into kubectl shell from spec.chart.metadata.annotations
```
catalog.cattle.io/ui-source-repo: helm3-library
catalog.cattle.io/ui-source-repo-type: cluster
apps.cattle.io/migrated: "true"
```
4. From apps&market --> repositories --> add the repo name: helm3-library | URL: https://git.rancher.io/helm3-charts | Branch: master
5. From the local cluster exec into the kubectl shell and add the annotation `kubectl edit apps.project.cattle.io aws-event-sources -n <ProjectIDAppInstalled>` , add the annotation cattle.io/skipUninstall: "true" under metadata.annotations
6. Delete the aws-event-sources from the downstream cluster --> default project
7. From the ember UI, install the same helm3 chart again but set the version `v0.1.4` in the default project.

**Result**
From explorer UI, the new app is not seen created and the existing app version is updated to v0.1.4

**Expected Result**
Expected it to create a new app under Installed apps from apps & market place and not update the existing app.

**Additional context**
Related : https://github.com/rancher/dashboard/issues/4402

deniseschannon:
I'm not sure why this issue is stuck with the note that it's being transferred, but considering the PR is merged, I'll move it to test.

@anupama2501 I believe this should unblock you from several other issues.

anupama2501:
Verified on v2.6-head ac67f83

1. On 2.5.11, created a downstream cluster with 3 worker, 1etcd +cp - c1
2. Installed aws-event sources in the default project of c1 
3. Upgraded the rancher server to v2.6-head ac67f83
4. Added the annotation for the app 
```
catalog.cattle.io/ui-source-repo: helm3-library
[...]

PR #35779, opened 2021-12-06
Check if matching dashboard app exists

rmweir:
**Problem:**
Prior, if a dashboard app already existed with the same ID in a
namespace that matched the targetNamespace of the
apps.project.cattle.io being created, they would be linked to
the same helm resources. This means changing one changed the
other.

**Solution:**
Now, an apps.project.cattle.io cannot be created if
a apps.catalog.cattle.io already exists with the same ID in
a namespace matching its targetNamespace.

**Issue:**
https://github.com/rancher/rancher/issues/35684

### DIFF
--- a/pkg/api/norman/customization/app/app.go
+++ b/pkg/api/norman/customization/app/app.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 
 	"github.com/rancher/norman/api/access"
 	"github.com/rancher/norman/httperror"
@@ -15,7 +16,7 @@ import (
 	"github.com/rancher/rancher/pkg/catalog/manager"
 	clusterv3 "github.com/rancher/rancher/pkg/client/generated/cluster/v3"
 	projectv3 "github.com/rancher/rancher/pkg/client/generated/project/v3"
-	"github.com/rancher/rancher/pkg/controllers/managementlegacy/compose/common"
+	"github.com/rancher/rancher/pkg/clustermanager"
 	hcommon "github.com/rancher/rancher/pkg/controllers/managementuserlegacy/helm/common"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	pv3 "github.com/rancher/rancher/pkg/generated/norman/project.cattle.io/v3"
@@ -29,10 +30,10 @@ import (
 
 type Wrapper struct {
 	Clusters              v3.ClusterInterface
+	ClusterManager        *clustermanager.Manager
 	CatalogManager        manager.CatalogManager
 	TemplateVersionClient v3.CatalogTemplateVersionInterface
 	TemplateVersionLister v3.CatalogTemplateVersionLister
-	KubeConfigGetter      common.KubeConfigGetter
 	AppGetter             pv3.AppsGetter
 	UserLister            v3.UserLister
 	UserManager           user.Manager
@@ -63,6 +64,14 @@ func Formatter(apiContext *types.APIContext, resource *types.RawResource) {
 }
 
 func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data map[string]interface{}) error {
+	projectID := convert.ToString(data["projectId"])
+	clusterID := strings.Split(projectID, ":")[0]
+	targetNamespace := convert.ToString(data["targetNamespace"])
+	appID := convert.ToString(data["name"])
+	if err := w.ValidateDashboardAppDoesNotExist(clusterID, appID, targetNamespace, request); err != nil {
+		return err
+	}
+
 	externalID := convert.ToString(data["externalId"])
 	if externalID == "" {
 		return nil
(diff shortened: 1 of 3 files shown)


jiaqiluo (dismissed):

pkg/types/config/context.go
```diff
@@ -521,6 +523,12 @@ func NewUserContext(scaledContext *ScaledContext, config rest.Config, clusterNam
 	}
 	context.RBACw = rbacw.Rbac().V1()
 
+	ctlg, err := catalog.NewFactoryFromConfigWithOptions(&context.RESTConfig, &catalog.FactoryOptions{SharedControllerFactory: controllerFactory})
```
aiyengar2, line 526: Just to clarify my own understanding:

If I understand the purpose of this PR correctly, the issue is that an app created by apps and marketplace v2 (an object that is stored on the downstream clusters) can potentially conflict with a user-deployed app created by apps and marketplace v1 (which is validated via norman and stored in the Rancher cluster?).

So we are modifying the user context to add in the ability to query apps & marketplace v2 catalogs, which then allows us to add ClusterManager to the norman wrapper that is used for apps & marketplace v1 apps, so that on seeing a v1 app get created in the local cluster, we can check if a v2 app is already installed in the **downstream** cluster first before updating?

e.g. if I have Monitoring V2 installed under the release name `rancher-monitoring` in the downstream cluster, this prevents Monitoring V1 from being installed since `rancher-monitoring` already exists in the downstream cluster?

Let me know if I'm understanding this incorrectly; if not, the overall logic here looks good to me.

pkg/api/norman/server/managementstored/setup.go
```diff
@@ -532,10 +531,10 @@ func App(schemas *types.Schemas, management *config.ScaledContext, kubeConfigGet
 	schema.Store = store
 	wrapper := app.Wrapper{
 		Clusters:              management.Management.Clusters(""),
+		ClusterManager:        clusterManager,
 		CatalogManager:        management.CatalogManager,
 		TemplateVersionClient: management.Management.CatalogTemplateVersions(""),
 		TemplateVersionLister: management.Management.CatalogTemplateVersions("").Controller().Lister(),
-		KubeConfigGetter:      kubeConfigGetter,
```
aiyengar2, line 538: need to also remove from the wrapper struct right? Since we are effectively replacing KubeConfigGetter with ClusterManager to be able to do more stuff than what the interface allows us to normally do?

pkg/api/norman/customization/app/app.go
```diff
@@ -92,6 +102,33 @@ func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data
...
+
+	appClient := clusterContext.Catalog.V1().App()
+	if _, err := appClient.Get(ns, appID, metav1.GetOptions{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return httperror.NewAPIError(httperror.ServerError, fmt.Sprintf(couldNotValidateErrFormat, err))
+		}
+	} else {
+		return httperror.NewAPIError(httperror.Conflict, fmt.Sprintf("an apps.catalog.cattle.io already exists with"+
+			" ID [%s] in namespace [%s] which matches targetNamespace.", appID, ns))
+	}
```
aiyengar2, line 127: nit: no period in errors and I think we don't need the last phrase.

```suggestion
		return httperror.NewAPIError(httperror.Conflict, fmt.Sprintf("an apps.catalog.cattle.io already exists with"+
			" ID [%s] in namespace [%s]", appID, ns))
	}
```

pkg/api/norman/customization/app/app.go
```diff
@@ -92,6 +102,33 @@ func (w Wrapper) Validator(request *types.APIContext, schema *types.Schema, data
...
 
+// ValidateDashboardAppDoesNotExist validates that a request is not attempting
+// to create an apps.project.cattle.io with an ID that matches an
+// apps.catalog.cattle.io in a namespace that matches its targetNamespace.
+func (w Wrapper) ValidateDashboardAppDoesNotExist(clusterName, appID, ns string, request *types.APIContext) error {
+	if request.Method != http.MethodPost {
+		return nil
+	}
+
+	couldNotValidateErrFormat := "could not validate apps.catalog.cattle.io with same ID and namespace does not exist: %v"
```
aiyengar2, line 113: nit: wording

```suggestion
	couldNotValidateErrFormat := "could not validate whether an apps.catalog.cattle.io with same ID and namespace exists: %v"
```

aiyengar2 (changes requested):

pkg/api/norman/customization/app/app.go
rmweir: I feel like that wouldn't make sense because it implies that the namespace is valid even if it already exists, which it wouldn't be.

pkg/api/norman/server/managementstored/setup.go
rmweir: It's not really being replaced since this wasn't being used at all anyway (although it could be used for what I'm doing), but otherwise yes it's not needed and I have removed from struct.

pkg/types/config/context.go
rmweir: That sounds right.

pkg/api/norman/customization/app/app.go
aiyengar2: discussed offline and was updated

aiyengar2 (approved):
LGTM

Issue #35725, opened 2021-12-02
[2.5] EKS upgrades loop without a default AMI list for v1.19 onwards


deniseschannon:
Backport https://github.com/rancher/rancher/issues/35295

SURE-3342

Auston-Ivison-Suse:
**Setup For Reproduction**
- Rancher version: v2.5.11
- Rancher install: docker
- Downstream clusters: EKS
   - Kubernetes version: 1.18/1.19

**Setup For Validation**
- Rancher version: v2.5-head(07a745a)
- Rancher install: docker
- Downstream clusters: EKS 
   - Kubernetes version: 1.18/1.19

**Steps For Reproduction**

1. Go to  https://rancherserver/g/clusters/add/launch/amazoneks
2. select regions as well as kubernetes versions: us-east2 and k8s: v1.18/v1.19 (One cluster for each version of k8s)
3. Proceed with all other defaults.
4. Create cluster
5. wait for error to show up which will be listed as follows: (this will happen when provisioning the v1.19 cluster and will happening when upgrading the v1.18 cluster to v1.18) 

```
Default ami of region us-east-2 for kubernetes version 1.19 is not set
````

**Steps For Validation**

1. Go to  https://rancherserver/g/clusters/add/launch/amazoneks
2. select regions as well as kubernetes versions: us-east2 and k8s: v1.18/v1.19 (One cluster for each version of k8s)
3. Proceed with all other defaults.
4. Create cluster
5. The v1.19 cluster will become active with no issues and once you upgrade the v1.18 cluster to v1.19 it should also become active with no issues. 

**Results**
The clusters provision correctly and have a default ami set in aws for the v1.19 kubernetes version

**Notes**
It should be noted that https://rancherserver/g/clusters/add/launch/amazoneksv2 allows for clusters to be provisioned in v2.5.11.

throrin19:
Same error in eu-west-1 region

thedadams:
@throrin19 This change was included in v2.5.12, which was just released.

throrin19:
I see that, But How remove the error and correctly launch the update of my cluster ? I try to reedit the cluster and validate but nothing happen

Issue #35295, opened 2021-09-17
EKS upgrades loop without a default AMI list for v1.19 onwards


gaktive:
Spotted in 2.5.9.

Issue description:

When upgrading a downstream EKS cluster (eksv1 provisioned), the AMI list doesn't provide default AMIs for v1.19 and onwards. It's possible to select a version higher than v1.18 in the UI (due to eksv2 dynamic discovery?) and is supported in our matrix.

  * https://github.com/rancher/rancher/blob/release/v2.5/pkg/kontainer-engine/drivers/eks/eks_driver.go

Business impact:
This causes an update/provisioning loop with any planned upgrade not possible.

Troubleshooting steps:

Repro steps:

Create an eksv1 cluster in the UI:

Either directly as v1.19 or higher (provisioning loop)
OR, as v1.18 or lower and later upgrade (update loop)
Note! v2.5.9 defaults to creating new EKS clusters using the ekvs2 provisioning workflow, the eksv1 workflow can be reached in the UI with: https://<rancher URL>/g/clusters/add/launch/amazoneks

Workaround:

Edit the cluster in the API to set the ami field to a valid EKS-Optimized AMI (not possible in the UI) that matches the desired version. Once the request is sent, follow through with another edit/save in the UI to set the client credentials and proceed.

  * https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html

SURE-3342

gaktive:
Clarification:
> That's correct, Rancher will error during provisioning attempting to look up the >=1.19 AMIs in the map. Yes it's the latter, the version is in the list (I assume the same list is presented for eksv1/v2), and if a cluster was provisioned using eksv1 the AMIs aren't defined beyond 1.18.

Error: `This cluster is currently Updating. Default ami of region us-east-1 for kubernetes version 1.19 is not set`

deniseschannon:
Note: We need to add in the AMIs for all EKS versions that are supported in the UI for KEv1. 

To test: Deploy an older KEv1 cluster in 2.4 and upgrade to 2.5-head and upgrade to a 1.19+ cluster.

AMIs need to be added for 1.19, 1.20, 1.21

https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html

We should also confirm that the UI has support to deploy all 3 versions.

deniseschannon:
There is a script that @paraglade has written before to help obtain these AMIs but I don't kmow where it's located.

[...]

PR #35789, opened 2021-12-07
[2.5] Update EKS default AMIs

thedadams:
Issue:
https://github.com/rancher/rancher/issues/35725

Original PR:
https://github.com/rancher/rancher/pull/35788


pkg/kontainer-engine/drivers/eks/eks_driver.go
```diff
@@ -41,125 +41,180 @@ const (
 )
 
 var amiForRegionAndVersion = map[string]map[string]string{
+	"1.21": {
```
cmurphy, line 44: Are you sure we want to support images for > 1.18 in 2.5?

thedadams: According to [this](https://github.com/rancher/rancher/issues/35295#issuecomment-984982454) comment, I believe so.

cmurphy (approved):

jakefhyde (approved):

Issue #35804, opened 2021-12-08
EKS Node Group deletion piloted by Rancher


jonathanribas:
**Rancher Server Setup**
- Rancher version: 2.5.6

**Information about the Cluster**
- Kubernetes version: 1.18
- Cluster Type (Local/Downstream): 
* Hosted = EKS

**Describe the bug**

Yesterday AWS had a serious service outage in North Virginia, so our EKS Node Groups were certainly unhealthy at a time.

Rancher has asked EKS to delete all our node groups. Our services went completely down, we had to run manual commands to make our website back in business.

It's still unstable at that time as Rancher UI inform us the following message even if we have re-created manually Node Groups on AWS side. It

> This cluster is currently Waiting.
> 
> Cluster must have at least one managed nodegroup.

**To Reproduce**

We are not sure how to reproduce but it may be related with unhealthy EKS Node Group.

**Result**

Node Groups are deleted automatically by Rancher. 

Our services are completely down. In order to have them back we need to create a new cluster as the existing one is not editable anymore. It seems to have lost connectivity with cluster on Rancher UI edit cluster.

> This cluster is currently Waiting.
> 
> Cluster must have at least one managed nodegroup.

Some logs at Rancher level:

2021/12/08 09:35:01 [INFO] updating cluster [c-lnl9z], upstream change detected 
2021/12/08 09:35:01 [ERROR] error refreshing EKS cluster [c-lnl9z] upstream state 
2021/12/08 09:35:21 [INFO] update kontainerdriver amazonelasticcontainerservice 
2021/12/08 09:35:21 [INFO] update kontainerdriver azurekubernetesservice 
2021/12/08 09:35:21 [INFO] update kontainerdriver baiducloudcontainerengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver huaweicontainercloudengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver opentelekomcloudcontainerengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver oraclecontainerengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver aliyunkubernetescontainerservice 
2021/12/08 09:35:21 [INFO] update kontainerdriver linodekubernetesengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver tencentkubernetesengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver googlekubernetesengine 
[...]

PR #35810, opened 2021-12-08
Check for nil clusterSpec in kev2 refresher

thedadams:
Previously, when an error was received in the upstream refresher, only the
`upstreamConfig` was nil-checked. However, if retrieving the upstreamSpec
failed, then `upstreamConfig` would not be nil, but the corresponding cluster
config would be. Therefore, in addition to checking the `upstreamConfig` for
nilness, the individual cluster configs are also checked. At least one of them
should be non-nil for syncing to proceed.

Issue:
https://github.com/rancher/rancher/issues/35804

### DIFF
--- a/pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
+++ b/pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
@@ -119,7 +119,6 @@ func getProviderAndReadyStatus(cluster *mgmtv3.Cluster) (string, bool) {
 // getProviderRefreshInterval returns the duration that should pass between
 // refreshing a cluster created by the given cloud provider.
 func getProviderRefreshInterval(provider string) (time.Duration, error) {
-
 	providerRefreshInterval := settings.GetSettingByID(fmt.Sprintf(refreshSettingFormat, strings.ToLower(provider)))
 	if providerRefreshInterval == "" {
 		return 300 * time.Second, nil
@@ -147,7 +146,6 @@ func nextRefreshTime(refreshInterval time.Duration, lastRefreshTime string) (tim
 }
 
 func (c *clusterRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cluster, cloudDriver string) (*mgmtv3.Cluster, error) {
-
 	logrus.Infof("checking cluster [%s] upstream state for changes", cluster.Name)
 
 	// In this call, it is possible to get errors back with non-nil upstreamSpec.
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
```diff
@@ -164,7 +163,7 @@ func (c *clusterRefreshController) refreshClusterUpstreamSpec(cluster *mgmtv3.Cl
 		apimgmtv3.ClusterConditionUpdated.False(cluster)
 		apimgmtv3.ClusterConditionUpdated.Message(cluster, fmt.Sprintf("[Syncing error%s] %s", syncFailed, err.Error()))
 
-		if upstreamConfig == nil {
+		if upstreamConfig == nil || (upstreamConfig.gkeConfig == nil && upstreamConfig.eksConfig == nil && upstreamConfig.aksConfig == nil) {
```
rmweir, line 166: Does DescribeCluster, or comparable functions for non-eks, just return a nil cluster with no error in this case?

rmweir (changes requested):

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
snasovich: Would be nice to update the comment block above to account for these additional checks and the reasons we need it.

thedadams: I think I understand what you are asking and I will do my best to provide an answer:

There are 3 possibilities for what `getComparableUpstreamSpec` returns:
1. non-nil upstreamSpec and nil error -- This indicates success
2. nil upstreamSpec and non-nil error -- This indicates error and no further action should be taken (this is what caused the problem: `upstreamConfig` was not `nil`, but `upstreamConfig.eksConfig` was nil)
3. non-nil upstreamSpec and non-nil error -- This indicates that there is a problem with one of the nodegroups (like the underlying launch template has been deleted and the corresponding nodegroup needs to be recreated). Syncing is still valid in this case, but the error needs to be reported so the user can take action.

rmweir: Oh I did realize we were already checking and writing an error. So in the outage case an error is returned by the eks API?

thedadams: Yes, a 500 error is what was reported.

thedadams: No, because of 3 above. We have to report problems with node groups based on launch templates.

rmweir: Sure we can report the errors but I mean should we be updating the spec ever?

thedadams: If the AWS API reports an error then I don't think we should sync.

If Rancher detects a problem with the node groups (like the launch template was deleted) then I think we should sync. The reason we should in this case is if a launch template is deleted, then the cluster is still in a good state. The only issue is that the node group cannot be edited; it must be recreated and deleted. So, if we don't sync in this case then we could potentially cause problems when a user fixes the deleted launch template.

Consider the following scenario:
1. A user accidentally deletes a launch template on AWS.
2. The user also adds a node group from the EKS console (not from Rancher).
3. The syncer runs and sees the problem with the launch template deleted, but does not sync. The error gets reported so the user sees it in Rancher's UI.
4. The user uses Rancher to recreate the messed up node group. But Rancher doesn't see the new node group the user added from the EKS console because the sync didn't happen. So this new node group gets deleted when the user makes the change in the Rancher UI.

leodotcloud:
@thedadams Couple of requests:
- Can you please share more info on how the outage caused the deletion of node groups?
- If we ever lose connectivity to the Amazon API or if we get any sort of error, can we please bubble this up all the way up to the UI with the specific details?

pkg/controllers/management/clusterupstreamrefresher/cluster_upstream_refresher.go
rmweir: Ah, and in the case the AWS API reported an error the config fields would be nil which is how can differentiate? Another alternative would be to just return nil in getComparableUpstreamSpec instead of an empty config if there's an error right? If I'm understanding this correctly then LGTM.

thedadams: Yes, we could do this error checking in the `getComparableUpstreamSpec`, but we would have to do three error checks there (one for each hosted provider) instead of one here. Also, there would still have to be checks for if "if an error is returned and the config is not nil" for each hosted provider.

In short, what I am trying to say is that even though the error checking is a little complicated here, I think it would be more so if it was moved to `getComparableUpstreamSpec`.

rmweir: Np, was just confirming that was a possible alternative to make sure I am understanding this.

rmweir (approved):
LGTM

cmurphy (approved):

snasovich (approved):

Issue #35809, opened 2021-12-08
[Backport v2.5.12] EKS Node Group deletion piloted by Rancher


rancherbot:
This is a backport issue for https://github.com/rancher/rancher/issues/35804, automatically created via rancherbot by @thedadams

Original issue description:

**Rancher Server Setup**
- Rancher version: 2.5.6

**Information about the Cluster**
- Kubernetes version: 1.18
- Cluster Type (Local/Downstream): 
* Hosted = EKS

**Describe the bug**

Yesterday AWS had a serious service outage in North Virginia, so our EKS Node Groups were certainly unhealthy at a time.

Rancher has asked EKS to delete all our node groups. Our services went completely down, we had to run manual commands to make our website back in business.

It's still unstable at that time as Rancher UI inform us the following message even if we have re-created manually Node Groups on AWS side. It

> This cluster is currently Waiting.
> 
> Cluster must have at least one managed nodegroup.

**To Reproduce**

We are not sure how to reproduce but it may be related with unhealthy EKS Node Group.

**Result**

Node Groups are deleted automatically by Rancher. 

Our services are completely down. In order to have them back we need to create a new cluster as the existing one is not editable anymore. It seems to have lost connectivity with cluster on Rancher UI edit cluster.

> This cluster is currently Waiting.
> 
> Cluster must have at least one managed nodegroup.

Some logs at Rancher level:

2021/12/08 09:35:01 [INFO] updating cluster [c-lnl9z], upstream change detected 
2021/12/08 09:35:01 [ERROR] error refreshing EKS cluster [c-lnl9z] upstream state 
2021/12/08 09:35:21 [INFO] update kontainerdriver amazonelasticcontainerservice 
2021/12/08 09:35:21 [INFO] update kontainerdriver azurekubernetesservice 
2021/12/08 09:35:21 [INFO] update kontainerdriver baiducloudcontainerengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver huaweicontainercloudengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver opentelekomcloudcontainerengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver oraclecontainerengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver aliyunkubernetescontainerservice 
2021/12/08 09:35:21 [INFO] update kontainerdriver linodekubernetesengine 
2021/12/08 09:35:21 [INFO] update kontainerdriver tencentkubernetesengine 
[...]

PR #35811, opened 2021-12-08
[2.5] Fix EKS nodegroup deletion on AWS API error

thedadams:
In the current implementation, an error was not checked, which led to all
nodegroups being deleted when an error was received from the AWS API when
listing nodegroups. Now, the error is checked and no deletion should occur.

In addition, to avoid setting the upstreamSpec to nil on refresh, each config on
 the `upstreamConfig` object is nil-checked. At least one of them should not be
 nil for the refresh to proceed.

Issue:
https://github.com/rancher/rancher/issues/35809

Original PR:
https://github.com/rancher/rancher/pull/35810


snasovich (changes requested):
Same comment(s) as on the https://github.com/rancher/rancher/pull/35810.
Also, super-nit - if we're removing redundant newlines in https://github.com/rancher/rancher/pull/35810, want to do the same here. Or not do it in both PRs.

thedadams:
I added a comment and removed the extra lines.

cmurphy (approved):

snasovich (approved):

rmweir (approved):
LGTM

Issue #35766, opened 2021-12-05
Ingress are not accessible after a rancher server upgrade


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.5.11 to 2.6.3-rc4 upgrade
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: 1.20.12-rancher1-2 rocky Custom cluster RKE1 and v1.17.17 RKE1 DO Node driver cluster

**Describe the bug**
Ingress dont work since they seem to be missing `ingressClassName`

**To Reproduce**

1. On 2.5.11, create downstream cluster
2. Create workload + Ingress
3. Upgrade rancher to 2.6.3-rc4
4. Create workload + Ingress
5. Upgrade k8s versions of the clusters to 1.22.3rancher1-1
6. Create workload + Ingress
7. Ingress created in Step 2 and 4 dont work now. 

- Error in ingress controller:
```
I1205 17:07:25.267710       8 store.go:367] "Ignoring ingress because of error while validating ingress class" ingress="pre-ns1/pre-testingress1" error="ingress does not contain a valid IngressClass"
I1205 17:07:25.267744       8 store.go:367] "Ignoring ingress because of error while validating ingress class" ingress="pre-ns1/pre-testingress2" error="ingress does not contain a valid IngressClass"
```

**Note**

- These failures seem to be inconsistent since [this](https://github.com/rancher/rancher/issues/35764) issue was seen when accessing the ingress (an hour back). But isnt seen anymore.
- The ingresses created in Step 2 ^^^ worked when rancher server was upgraded and Automation was run right after to validate. But doesnt work anymore after k8s upgrade.
- Not able to create an ingress on the DO cluster now. fails with error: 
```
Not able to create an ingress, fails with error "admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: host "sslip.io" and path "/name.html" is already defined in ingress post-k8s-ns1/ing01"
```

superseb:
Can be tested with latest v2.6-head: `fd930f765`

cmurphy:
@superseb https://github.com/rancher/rancher/pull/35818 is still in review, this is not ready to test yet

sowmyav27:
To Validate:
Fresh install checks on 2.6-head

- Deploy 1.22 k8s cluster and make sure the ingress works
- deploy a 1.21 cluster and upgrade to 1.22 and make sure ingress works

[...]

PR #35818, opened 2021-12-09
Restart agent pods after upgrade to 1.22

cmurphy:
When a cluster is created with version 1.18 or lower, various
controllers will start a watch for the extensions/v1beta1/Ingress type.
Without this patch, if that cluster is then upgraded to 1.22, those
watches will start to fail because that API is no longer available, and
the controllers using those watches will stop syncing because they don't
see events for the new Ingress type, networking.k8s.io/v1/Ingress.

This patch adds a check to the clusterstats controller, where the
cluster version status is updated, and annotates the cluster agent
deployment template which forces a restart of the cluster agents. Simply
forcing the agent to be reapplied via the clusterdeploy controller does
not necessarily fix the problem because if there is no change to the
cluster agent deployment template then the pods won't actually be
restarted.

The version checks for an original version of 1.21, rather than 1.18.
This is because it's possible that a cluster was created with version
1.18 or below, then upgraded to version 1.21 or below, where the old
extensions/v1beta1 API would still be available and the controllers
would still work. But then the final upgrade to 1.22 would break the
controllers. So, assume that any upgrade to 1.22 needs the agents
restarted.

https://github.com/rancher/rancher/issues/35766

### DIFF
--- a/pkg/controllers/management/clusterstats/statsaggregator.go
+++ b/pkg/controllers/management/clusterstats/statsaggregator.go
@@ -3,6 +3,7 @@ package clusterstats
 import (
 	"context"
 	"reflect"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -10,8 +11,11 @@ import (
 	"github.com/rancher/rancher/pkg/clustermanager"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config"
+	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 )
@@ -21,6 +25,7 @@ const (
 	nodeRoleControlPlaneHyphen = "node-role.kubernetes.io/control-plane"
 	nodeRoleETCD               = "node-role.kubernetes.io/etcd"
 	nodeRoleMaster             = "node-role.kubernetes.io/master"
+	agentVersionUpgraded       = "agent.cluster.cattle.io/upgraded-v1.22"
 )
 
 type StatsAggregator struct {
@@ -160,7 +165,30 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 		v32.ClusterConditionNoMemoryPressure.False(cluster)
 	}
 
+	var oldVersion int
+	if cluster.Status.Version != nil {
+		oldVersion, err = strconv.Atoi(cluster.Status.Version.Minor)
+		if err != nil {
+			return err
+		}
+	}
 	versionChanged := s.updateVersion(cluster)
+
+	// If the cluster went through an upgrade from <=1.21 to >=1.22, restart
+	// the cluster agent in order to restart controllers that will no longer work
+	// with the new API.
+	if versionChanged {
+		newVersion, err := strconv.Atoi(cluster.Status.Version.Minor)
+		if err != nil {
+			return err
+		}
+		if newVersion >= 22 && oldVersion <= 21 {
+			err := s.restartAgentDeployment(cluster)
+			if err != nil {
+				return err
+			}
+		}
+	}
 
 	if statusChanged(origStatus, &cluster.Status) || versionChanged {
 		_, err = s.Clusters.Update(cluster)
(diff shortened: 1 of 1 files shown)


cmurphy:
Actually my idea won't work, since the controllers that need to be restarted are all in the agent, and the clustermanager is not available in the agent nor are there any cluster controllers running in the agent that can detect when the cluster version has changed, so I think this solution is probably best.

ibuildthecloud (dismissed):

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -403,6 +412,29 @@ func (cd *clusterDeploy) setNetworkPolicyAnn(cluster *v3.Cluster) error {
...
+	if err != nil {
+		return err
+	}
+	if cluster.Annotations[AgentForceRestartAnn] != "true" {
+		return nil
+	}
+	if originalResourceVersion != cd.getClusterAgentResourceVersion(cluster.Name) {
+		// kubectl apply should have already restarted this deployment
+		return nil
+	}
```
rmweir, line 428: Couldn't the resource version be bumped for other reasons? I'm not sure if this is enough to infer this.

rmweir: Would another option be to put the annotation on the agent deployment before updating the cluster, that way it will be removed once the agent is redeployed?

rmweir (changes requested):

pkg/controllers/management/clusterdeploy/clusterdeploy.go
cmurphy: Not sure I parse this, are you saying to put an annotation on the downstream agent deployment resource instead of the local cluster resource, in order to force the agent to be redeployed?

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -403,6 +412,29 @@ func (cd *clusterDeploy) setNetworkPolicyAnn(cluster *v3.Cluster) error {
...
+// restartAgent performs a kubectl rollout restart on the cluster agent deployment
+// if a force restart was requested and if it wasn't already restarted by kubectl apply.
+func (cd *clusterDeploy) restartAgent(originalResourceVersion string, cluster *v3.Cluster) error {
+	kubeConfig, _, err := cd.getKubeConfig(cluster)
+	if err != nil {
+		return err
+	}
+	if cluster.Annotations[AgentForceRestartAnn] != "true" {
+		return nil
+	}
```
thedadams, line 424: nit: move this check to the top of the function.

thedadams (approved):
One nit. Otherwise, looks good.

cmurphy:
@rmweir came up with the idea of annotating the deployment template which automatically causes the pods to restart and doesn't rely on the resource version number, so updated this PR to do that.

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -160,8 +165,31 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 		v32.ClusterConditionNoMemoryPressure.False(cluster)
 	}
 
+	var oldVersion int
+	if cluster.Status.Version != nil {
+		oldVersion, err = strconv.Atoi(cluster.Status.Version.Minor)
+		if err != nil {
+			return err
+		}
+	}
```
thedadams, line 174: Can this be moved inside the if-statement so it only happens when necessary?

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -160,8 +165,31 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
...
 	versionChanged := s.updateVersion(cluster)
 
+	// If the cluster went through an upgrade from <=1.21 to >=1.22, restart
+	// the cluster agent in order to restart controllers that will no longer work
+	// with the new API.
+	if versionChanged {
+		newVersion, err := strconv.Atoi(cluster.Status.Version.Minor)
```
thedadams, line 181: Do we need to check `cluster.Status.Version != nil` here too?

thedadams (commented):
Just two questions. If the answer to both is "no" then looks good to me.

pkg/controllers/management/clusterstats/statsaggregator.go
cmurphy: Do you mean the other if statement on line 180? It has to happen before `s.updateVersion` is called because that is when `cluster.Status.Version` changes, so this is the only place to put it.

pkg/controllers/management/clusterstats/statsaggregator.go
cmurphy: if `s.updateVersion` returns true then `cluster.Status.Version` is guaranteed to be set https://github.com/rancher/rancher/pull/35818/files/4e175ea23b1898d7924c5d52f171eb8dbc48fa37#diff-ae8f3a50a0e7a52898014aa299562a2cf77fb2ef40b5bc348478c6537a304978R213

thedadams (approved):

rmweir (approved):
LGTM

PR #35829, opened 2021-12-09
Validation provisioning test

igomez06:
There provisioning tests added, as well as framework additions.
**Framework:**
    Cleanup boolean (so the end user has the ability to forgo a cleanup)
     rke-machine-config constructor for digital ocean. (Needed to provision a DO cluster)
     machine pool setup helper function (to facilitate creating multiple pools for a cluster)
    name generator package for creating unique names for resources created
    new config for node role inputs for users do input their desired node setup for machine pools for a cluster
**Node Driver Provisioning tests:**
    AWS/Digital Ocean node driver cluster provisioning
    1_Node_all_roles_Admin_User
    1_Node_all_roles_Standard_User
    3_nodes_-_1_role_per_node_Admin_User
    3_nodes_-_1_role_per_node_Standard_User
    Dynamic Input admin user
    Dynamic Input standard user
   1_Node_all_roles_Admin_User
   1_Node_all_roles_Standard_User
   3_nodes_-_1_role_per_node_Admin_User
   3_nodes_-_1_role_per_node_Standard_User
   Dynamic Input admin user
   Dynamic Input standard user

### DIFF
--- a/tests/framework/clients/rancher/client.go
+++ b/tests/framework/clients/rancher/client.go
@@ -42,6 +42,8 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 		RancherConfig: rancherConfig,
 	}
 
+	session.HasCleanupSet = *rancherConfig.Cleanup
+
 	var err error
 	restConfig := newRestConfig(bearerToken, rancherConfig)
 	c.restConfig = restConfig
--- a/tests/framework/clients/rancher/client.go
+++ b/tests/framework/clients/rancher/client.go
@@ -42,7 +42,7 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 		RancherConfig: rancherConfig,
 	}
 
-	session.HasCleanupSet = *rancherConfig.Cleanup
+	session.CleanupEnabled = *rancherConfig.Cleanup
 
 	var err error
 	restConfig := newRestConfig(bearerToken, rancherConfig)
--- a/tests/framework/clients/rancher/config.go
+++ b/tests/framework/clients/rancher/config.go
@@ -6,6 +6,7 @@ type Config struct {
 	Host       string `yaml:"host"`
 	AdminToken string `yaml:"adminToken"`
 	Insecure   *bool  `yaml:"insecure" default:"true"`
+	Cleanup    *bool  `yaml:"cleanup" default:"true"`
 	CAFile     string `yaml:"caFile" default:""`
 	CACerts    string `yaml:"caCerts" default:""`
 }
--- a/tests/framework/extensions/clusters/config.go
+++ b/tests/framework/extensions/clusters/config.go
@@ -1,7 +1,0 @@
-package clusters
-
-const ConfigurationFileKey = "clusters"
-
-type Config struct {
-	NodesAndRoles string `yaml:"nodesAndRoles" default:""`
-}
--- a/tests/framework/extensions/clusters/config.go
+++ b/tests/framework/extensions/clusters/config.go
@@ -1,0 +1,7 @@
+package clusters
+
+const ConfigurationFileKey = "clusters"
+
+type Config struct {
+	NodesAndRoles string `yaml:"nodesAndRoles" default:""`
+}
--- a/tests/framework/pkg/namegenerator/namegenerator.go
+++ b/tests/framework/pkg/namegenerator/namegenerator.go
@@ -13,7 +13,7 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
-func RandStringLowerBytes(n int) string {
+func RandStringLower(n int) string {
 	b := make([]byte, n)
 	for i := range b {
 		b[i] = lowerLetterBytes[rand.Intn(len(lowerLetterBytes))]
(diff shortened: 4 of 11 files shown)


igomez06:
@paynejacob I've made some updates to the framework, would like your thoughts on them. Thank you.

tests/framework/clients/rancher/client.go
```diff
@@ -42,6 +42,8 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 		RancherConfig: rancherConfig,
 	}
 
+	session.HasCleanupSet = *rancherConfig.Cleanup
```
paynejacob, line 45:
```suggestion
	session.Cleanup = *rancherConfig.Cleanup
```

tests/framework/extensions/clusters/config.go
```diff
@@ -0,0 +1,7 @@
+package clusters
+
+const ConfigurationFileKey = "clusters"
```
paynejacob, line 3: I like the idea of this configuration but this seems extremely generic.  Could we move this config to the test suite that uses it and update the key accordingly.

tests/framework/pkg/namegenerator/namegenerator.go
```diff
@@ -0,0 +1,34 @@
...
+
+const lowerLetterBytes = "abcdefghijklmnopqrstuvwxyz"
+const upperLetterBytes = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+const numberBytes = "0123456789"
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+func RandStringLowerBytes(n int) string {
```
paynejacob, line 16: Bytes makes  it sound like the function returns bytes.

```suggestion
func RandStringLower(n int) string {
```

tests/framework/pkg/namegenerator/namegenerator.go
```diff
@@ -0,0 +1,34 @@
...
+
+func RandStringWithCharset(length int, charset string) string {
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[rand.Intn(len(charset))]
+	}
+	return string(b)
+}
+
+func RandStringAllBytes(length int) string {
```
paynejacob, line 32: Bytes makes  it sound like the function returns bytes.

```suggestion
func RandStringAll(length int) string {
```

paynejacob (changes requested):
Over all it looks good. I think we should be careful with adding generic configuration to the extensions.  If configuration is specific to a test is should be defined with the test.

tests/framework/clients/rancher/client.go
igomez06: With the bool name as Cleanup its name collides with our session.Cleanup func name. I think it should be named differently.

igomez06: fixed

tests/framework/extensions/clusters/config.go
igomez06: I agree, I moved the config out to the provisioning validation test suite.

tests/framework/pkg/namegenerator/namegenerator.go
igomez06: fixed

tests/framework/pkg/namegenerator/namegenerator.go
igomez06: fixed

tests/framework/pkg/session/session.go
```diff
@@ -9,9 +9,10 @@ type CleanupFunc func() error
 
 // Session is used to track resources created by tests by having a LIFO queue the keeps track of the delete functions.
 type Session struct {
-	cleanupQueue []CleanupFunc
-	open         bool
-	testingT     *testing.T
+	CleanupEnabled bool
```
paynejacob, line 12: The constructor should default this to true.

paynejacob (changes requested):

tests/framework/pkg/session/session.go
igomez06: fixed

paynejacob (approved):

izaac:
I just wanted to be clear on the scope of this PR before review.

This will make sure the cluster gets a Ready/Active status in Rancher after provisioning and no other validation would be made. Is this correct?

igomez06:
> I just wanted to be clear on the scope of this PR before review.
> 
> This will make sure the cluster gets a Ready/Active status in Rancher after provisioning and no other validation would be made. Is this correct?

Correct

tests/framework/clients/rancher/config.go
```diff
@@ -6,6 +6,7 @@ type Config struct {
 	Host       string `yaml:"host"`
 	AdminToken string `yaml:"adminToken"`
 	Insecure   *bool  `yaml:"insecure" default:"true"`
+	Cleanup    *bool  `yaml:"cleanup" default:"true"`
```
izaac, line 9: If I set the cleanup option to `false` would I end up with all the provisioned clusters in Rancher?
What else are we cleaning up here, users?

I just wanted to understand this part, this PR looks good for the scope defined.

igomez06: If you set the cleanup option to `false` it would not clean anything up. So the clusters would be there, as well as the users.

izaac (approved):
LGTM 🚀

paynejacob (approved):

igomez06:
@paynejacob sorry Jacob one last review, I made some updates for the configs in the machinepools directory. Suggestions made from my team. I would appreciate one last review on this PR before we merge it in. Thanks!

paynejacob (approved):

PR #35832, opened 2021-12-09
Explicitly sort app revisions by creation timestamp

genexpr:
This is a code change for a reopened issue [20654](https://github.com/rancher/rancher/issues/20654).

App revisions on retrieval and listing were sorted by creation timestamp. My code used to rely on that behavior. It appears to have changed, but I don't know how or where. So it's best to explicitly sort app revisions by creation timestamp and delete the oldest revisions, as needed, according to this order.


annablender (approved):
LGTM to me @genexpr, nicely done

Issue #35701, opened 2021-12-01
Deleting provisioned RKE2 cluster could leave jobs/secrets around


thedadams:
**Rancher Server Setup**
- Rancher version: 2.6-head
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): Any
- Proxy/Cert Details: None

**Information about the Cluster**
- Kubernetes version: 1.21
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Provisioned RKE2 cluster with any provider

**Describe the bug**

When deleting a provisioned RKE2 cluster from Rancher, there could be some secrets and a job/pod leftover after the cluster disappears. This is mostly harmless because these items will just be recreated if another cluster is create with the same name and configuration. However, Rancher should clean them up correctly.

**To Reproduce**
 
1. Provision an RKE2 cluster from Rancher
2. Wait for the cluster to become active
3. Delete the cluster
4. Wait for the cluster to disappear from the UI

**Result**
With `kubectl -n fleet-default get jobs` and `kubectl -n fleet-default get secrets` you may see some stuff leftover from the deleted cluster.

**Expected Result**

All the aforementioned items should be cleaned up from Rancher.

**Additional context**

The problem here is that the infrastructure object will go back through the controller after the infrastructure has been cleaned up. This is because a condition will get updated and the controller will pick up the change. The job and secrets will be recreated after they were already deleted and the job will kick off again. The infrastructure object, meanwhile, will be cleaned up and the leftover items never cleaned up (because they only get cleaned up when the infrastructure machine object goes through the delete controller successfully).

The solution (which I have tested) is to short-circuit the infrastructure machine deletion if the `DeleteJob` condition is set (the `DeleteJob` condition will only be set if the delete job completed or failed). Care should be taken to handle the force-delete situation with infrastructure machines.

snasovich:
Likely already fixed in 2.6.3, need to verify in 2.6.4.

jakefhyde:
Haven't been able to reproduce on `v2.6-head` after trying multiple times.

## Before removing cluster
```
❯ kubectl -n fleet-default get jobs
No resources found in fleet-default namespace.
❯ kubectl -n fleet-default get secrets
NAME                                                              TYPE                                          DATA   AGE
[...]

Issue #35823, opened 2021-12-09
No timeout when trying to delete an AWS node when the cloud credential is deleted.


sowmyav27:
**On 2.6-head commit id: `77392a6`**

- Deploy an RKE2 cluster - 1 node all roles and worker node using cloud cred1 (docker install rancher)
- Create cloud cred2
- Edit cluster, change the cloud cred to point to cred2
- Edit cluster, add a worker node
- Delete cred2
- Delete cluster
- Worker node added in step 4 is stuck in removing state. Other nodes are removed.
- No messages as mentioned in the original issue's description seen in the rancher logs
- But the node stays in Deleting stays and does not go into Deletionerror state or failed delete state. (i.e. there is no time out) 
- stays in this state:
<img width="1444" alt="Screen Shot 2021-12-08 at 10 48 28 PM" src="https://user-images.githubusercontent.com/26032343/145348151-41a7ce4a-d173-4969-be4d-ad2a5344f46f.png">

**Expected:**
The delete action should timeout and fail

thedadams:
Since this has been moved, I will comment that this doesn't seem to be a Rancher issue. The deletion job gets created by Rancher as expected, but Kubernetes never creates the pods for the job. This is the same problem as list [here](https://github.com/rancher/rancher/issues/35795).

thedadams:
This is not really a timeout issue. The problem is that Rancher is creating the deletion jobs, deleting it, and recreating it too quickly and Kubernetes does not create the pods for the job.

This has been moved back to 2.6.3 (and other issues moved out) because this issue will not be resolved on its own. It would take some intervention by the user to get the machines to clean up properly.

This issue is intermittent and more likely to occur when a cluster has multiple nodes. Creating a few RKE2 clusters, waiting for them to be active and then deleting them would be enough to test this. If they are all deleted successfully, then the fix should be working.

sowmyav27:
**On 2.6-head commit id: `2a7857d`**

- Deploy an RKE2 cluster - 1 node all roles and worker node using cloud cred1 (docker install rancher)
- Create cloud cred2
- Edit cluster, change the cloud cred to point to cred2
- Edit cluster, add a worker node
- Delete cred2
- Delete cluster
- Worker node added in step 4 is stuck in removing state. Other nodes are removed.
- The machine stays in Deleting state.
- Errors in rancher logs every two minutes:

```
[...]

Issue #35795, opened 2021-12-07
[Intermittent] Machines not deleted for provisioned RKE2 clusters


thedadams:
**Rancher Server Setup**
- Rancher version:
- Installation option (Docker install/Helm Chart): Docker install and Rancher Desktop
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): K3S
- Proxy/Cert Details: None

**Describe the bug**
Deleting a provisioned RKE2 cluster from Rancher may hang waiting for a machine to delete.

**To Reproduce**
1. Provision an RKE2 cluster
2. Wait for said cluster to become active
3. Delete said cluster

**Result**
The delete may hang waiting for a machine to delete

**Expected Result**
The cluster should be successfully deleted.

**Additional context**
The job to delete the infrastructure gets created by the `machineprovision` controller, but the job doesn't do anything. There are no pods created and no events on the job. With this context, it seems to me that this may be a problem with K3S (the local cluster) because if the job gets created successfully, then Kubernetes should create the pods for the job.

One other thing is: at the time of this writing, jobs and secrets are leftover when a cluster is successfully deleted. These jobs and secrets don't affect anything Rancher does (they will be recreated when a new cluster of the same name is created), but their existence may be affecting this issue. This is unlikely, but it may be something to investigate.

thedadams:
There is a workaround for this problem. If a machine get stuck in the deleting state:
1. Find the infrastructure machine object related to the machine. It is on the machine object as `spec.infrastructureRef`
2. Delete the job associated with this infrastructure machine. It will be in the same namespace and its name will be the name of the infrastructure object with `machine-provision` appended.
3. Add a dummy annotation to the infrastructure object. This will cause the infrastructure object to go back through the controllers and the job will be recreated. It is unlikely that the job will hang when it is created a second time.

thedadams:
### Root cause
The root cause here was the deletion job was getting created and then modified too quickly for Kubernetes. The reason for the modification was a secret that is not needed for deletion was getting deleted.

### What was fixed, or what changes have occurred
The secret that is not needed for deletion is left out of the Job spec on delete.

### Areas or cases that should be tested
None, as long as deletion of the machines is successful.

### What areas could experience regressions?
Deletion of RKE2/K3S machines only.

### Are the repro steps accurate/minimal?
Yes.

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head 4d41423
[...]

PR #35847, opened 2021-12-10
Ensure proper job cleanup

thedadams:
There were a few issues related to job cleanup on deletion. Jobs were getting
deleted and recreated too quickly. In addition, jobs were being leftover after
deletion that was exacerbating the problem.

After this change, jobs should not be left over and machine deletion should not
hang because Wrangler will be able to clean up old jobs before creating new
ones.

Issue:
https://github.com/rancher/rancher/issues/35823

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
@@ -50,6 +50,37 @@ const (
 	drainingSucceededCondition = condition.Cond(capi.DrainingSucceededCondition)
 )
 
+type infraObject struct {
+	data     data.Object
+	obj      runtime.Object
+	meta     metav1.Object
+	typeMeta metav1.Type
+}
+
+func newInfraObject(obj runtime.Object) (*infraObject, error) {
+	objMeta, err := meta.Accessor(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := data.Convert(obj.DeepCopyObject())
+	if err != nil {
+		return nil, err
+	}
+
+	typeMeta, err := meta.TypeAccessor(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	return &infraObject{
+		data:     data,
+		obj:      obj,
+		meta:     objMeta,
+		typeMeta: typeMeta,
+	}, nil
+}
+
 type handler struct {
 	ctx                 context.Context
 	apply               apply.Apply
@@ -252,7 +283,7 @@ func getMachineStatusFromPod(pod *corev1.Pod, kind, condType string) rkev1.RKEMa
 	for _, containerStatus := range pod.Status.ContainerStatuses {
 		if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.ExitCode != 0 {
 			failureMessage := strings.TrimSpace(containerStatus.State.Terminated.Message)
-			message := CreatingMachineMessage(kind)
+			message := FailedMachineCreateMessage(kind, reason, failureMessage)
 			if condType != createJobConditionType {
 				message = FailedMachineDeleteMessage(kind, reason, failureMessage)
 			}
@@ -299,12 +330,39 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 		return obj, err
 	}
 
-	objMeta, err := meta.Accessor(obj)
+	infraObj, err := newInfraObject(obj)
 	if err != nil {
-		return nil, err
+		return obj, err
 	}
 
-	clusterName := objMeta.GetLabels()[capi.ClusterLabelName]
+	if !infraObj.data.Bool("status", "jobComplete") && infraObj.data.String("status", "failureReason") == "" {
+		return obj, fmt.Errorf("cannot delete machine %s because create job has not finished", infraObj.meta.GetName())
+	}
+
+	if cond := getCondition(infraObj.data, deleteJobConditionType); cond != nil {
+		job, err := h.jobs.Get(infraObj.meta.GetNamespace(), getJobName(infraObj.meta.GetName()))
+		if apierror.IsNotFound(err) {
+			// If the deletion job condition has been set on the infrastructure object and the deletion job has been removed,
+			// then we don't want to create another deletion job.
+			logrus.Infof("Machine %s %s has already been deleted", infraObj.obj.GetObjectKind().GroupVersionKind(), infraObj.meta.GetName())
+			return obj, nil
+		}
+		if err != nil {
+			return obj, err
+		}
+
+		if shouldCleanupObjects(job, infraObj.data) {
+			// Calling WithOwner(obj).ApplyObjects with no objects here will look for all objects with types passed to
+			// WithCacheTypes above that have an owner label (not owner reference) to the given obj. It will compare the existing
+			// objects it finds to the ones that are passed to ApplyObjects (which there are none in this case). The apply
+			// controller will delete all existing objects it finds that are not passed to ApplyObjects. Since no objects are
+			// passed here, it will delete all objects it finds.
+			return obj, h.apply.WithOwner(obj).ApplyObjects()
+		}
+		return obj, generic.ErrSkip
+	}
+
+	clusterName := infraObj.meta.GetLabels()[capi.ClusterLabelName]
 	if clusterName == "" {
 		return obj, fmt.Errorf("error retrieving the clustername for machine, label key %s does not appear to exist for dynamic machine %s", capi.ClusterLabelName, key)
 	}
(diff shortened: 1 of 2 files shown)


Oats87 (approved):
LGTM

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -590,6 +614,16 @@ func insertOrUpdateCondition(d data.Object, desiredCondition summary.Condition)
 	return true, nil
 }
 
+func getCondition(d data.Object, conditionType string) *summary.Condition {
```
snasovich, line 617: Could use this func in `shouldForceRemoval` implementation.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -639,7 +673,7 @@ func shouldForceRemoval(job *batchv1.Job, d data.Object) bool {
 		return false
 	}
 
-	return condition.Cond("Failed").IsTrue(job)
+	return job.Labels[InfraJobRemove] == "true" && condition.Cond("Failed").IsTrue(job)
```
snasovich, line 676: Could you please explain why this additional condition was added? What specific case does it address?

Also, do we want to do `strings.ToLower` comparison here similar to how we do above for annotation just to be safe?

snasovich (changes requested):
Mostly questions, but there are 1-3 items that probably should be addressed.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -252,7 +283,7 @@ func getMachineStatusFromPod(pod *corev1.Pod, kind, condType string) rkev1.RKEMa
 	for _, containerStatus := range pod.Status.ContainerStatuses {
 		if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.ExitCode != 0 {
 			failureMessage := strings.TrimSpace(containerStatus.State.Terminated.Message)
-			message := CreatingMachineMessage(kind)
+			message := failureMessage
```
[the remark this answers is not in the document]
thedadams: It was changed because the `CreatingMachineMessage` is for actively creating machines i.e. it says "creating machine"

Since the create failed at this point, the message should be changed. I updated the message as you suggested.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -412,55 +452,39 @@ func (h *handler) OnChange(obj runtime.Object) (runtime.Object, error) {
...
 
-func (h *handler) run(obj runtime.Object, objMeta metav1.Object, create bool, jobBackoffLimit int32) (runtime.Object, error) {
-	typeMeta, err := meta.TypeAccessor(obj)
-	if err != nil {
-		return nil, err
-	}
-
-	d, err := data.Convert(obj.DeepCopyObject())
+func (h *handler) run(infraObj *infraObject, create bool, jobBackoffLimit int32) (runtime.Object, error) {
+	typeMeta, err := meta.TypeAccessor(infraObj.obj)
```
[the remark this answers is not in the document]
thedadams: Removed.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -299,12 +330,31 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
 
-	objMeta, err := meta.Accessor(obj)
+	infraObj, err := newInfraObject(obj)
 	if err != nil {
-		return nil, err
+		return obj, err
+	}
+
+	if !infraObj.data.Bool("status", "jobComplete") && infraObj.data.String("status", "failureReason") == "" {
+		return obj, fmt.Errorf("cannot delete machine %s because create job has not finished", infraObj.meta.GetName())
```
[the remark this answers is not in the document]
thedadams: Yes, one of `jobComplete` or `failureMessage` will be set when the create jobs finishes successfully or fails.

These fields are not updated when the delete job starts, so there is no danger in getting stuck here by the delete job.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -299,12 +330,31 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
+			// If the deletion job condition has been set on the infrastructure object and the deletion job has been deleted,
+			// then we don't want to create another deletion job.
+			logrus.Infof("Machine %s %s has already been deleted", infraObj.obj.GetObjectKind().GroupVersionKind(), infraObj.meta.GetName())
+			return obj, nil
+		}
+		if err != nil {
+			return obj, err
+		}
+		// We know the job and condition both exist, so just clean everything up.
+		return obj, h.apply.WithOwner(obj).ApplyObjects()
```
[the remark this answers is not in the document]
thedadams: No, we should check the `shouldForceRemoval` function. This check is added.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: Yes, good point. I will update the implementation.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: Spoke offline about this. I will add the `strings.ToLower`.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -345,63 +403,37 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
+	obj, err := h.run(infraObj, false, 3)
 	if err != nil {
 		return nil, err
 	}
-	job, err := h.jobs.Get(objMeta.GetNamespace(), getJobName(infraName))
-	if err != nil {
-		return nil, err
-	}
-
-	if !job.Status.CompletionTime.IsZero() || shouldForceRemoval(job, d) {
```
thedadams, line 377: For reviewers: This is no longer needed because the check was moved up.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -623,29 +646,33 @@ func constructFilesSecret(driver string, config map[string]interface{}) (*corev1
...
+		return false
 	}
 
 	forceRemoveAnnValue, _ := d.Map("metadata", "annotations")[forceRemoveMachineAnn].(string)
 
-	if createJobCondition.Reason() != string(errors.CreateMachineError) &&
-		strings.ToLower(forceRemoveAnnValue) != "true" {
-		return false
+	if createJobCondition.Reason() == string(errors.CreateMachineError) || strings.ToLower(forceRemoveAnnValue) == "true" {
+		return strings.ToLower(job.Spec.Template.Labels[InfraJobRemove]) == "true" && condition.Cond("Failed").IsTrue(job)
```
thedadams, line 662: This is changed because the previous implementation was incorrect (thanks Sergey).

snasovich: To clarify based on offline discussion - the previous implementation was actually correct and produces the same outputs for the same permutations of input conditions. However, the new implementation matches natural thinking more closely and thus is easier to digest.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
snasovich: Is it OK we've essentially flipped the order? Previously it was `h.run` first and then cleanup of the objects and now it's the other way around.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -623,29 +646,33 @@ func constructFilesSecret(driver string, config map[string]interface{}) (*corev1
...
-	return condition.Cond("Failed").IsTrue(job)
+	return false
 }
 
 func FailedMachineDeleteMessage(kind, failureReason, failureMessage string) string {
 	return fmt.Sprintf("failed deleting server (%s) in infrastructure provider: %s: %s", kind, failureReason, failureMessage)
 }
 
+func FailedMachineCreateMessage(kind, failureReason, failureMessage string) string {
+	return fmt.Sprintf("failed creatinging server (%s) in infrastructure provider: %s: %s", kind, failureReason, failureMessage)
```
snasovich, line 673: A bit too many `ing`s :)
```suggestion
	return fmt.Sprintf("failed creating server (%s) in infrastructure provider: %s: %s", kind, failureReason, failureMessage)
```

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -623,29 +646,33 @@ func constructFilesSecret(driver string, config map[string]interface{}) (*corev1
...
 
-	return condition.Cond("Failed").IsTrue(job)
+	return false
 }
 
 func FailedMachineDeleteMessage(kind, failureReason, failureMessage string) string {
 	return fmt.Sprintf("failed deleting server (%s) in infrastructure provider: %s: %s", kind, failureReason, failureMessage)
 }
 
+func FailedMachineCreateMessage(kind, failureReason, failureMessage string) string {
```
snasovich, line 672: Though why we don't make this `func` unexported, but I see the others like this are used in `rancher/src/github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2/machinestatus/machinestatus.go` - do we need to add handling of this one there too?

snasovich (changes requested):

snasovich:
Also, do I understand it should address https://github.com/rancher/rancher/issues/35795 and https://github.com/rancher/rancher/issues/35701? If that's the case, let's link them to this PR too.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: It's exported because the other, related, functions are also exported. It doesn't need to be used in the spot you mentioned because the status will already be put on the machine status. If this changes, then we will have to manually do it in the same way you mentioned.

thedadams: If you feel strongly that it should be unexported, I can do that. Unless you object, I will keep it as is.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
thedadams: Fixed.

thedadams:
Yes, you are correct that this PR should also fix those issues: https://github.com/rancher/rancher/issues/35795 is essentially the same as the already linked issue and this PR fixes https://github.com/rancher/rancher/issues/35701 sort of by accident.

As far as [this](https://github.com/rancher/rancher/pull/35847#discussion_r767081431) comment goes (not sure why I can't reply to your comment): I did switch the order for `h.run` and checking the job. This is actually better because the infrastructure machine object will be re-enqueued when the job finishes/fails. Then the controller will check the job. Checking the job right after creating it (which was how it was) makes less sense.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
snasovich: "As is" is fine, I just wanted to make sure we didn't need to use it in that spot where the other 3 are used. Resolving.

snasovich (approved):

snasovich:
@thedadams , thank you for the change and clarifications - the changed order makes sense.

PR #35852, opened 2021-12-11
Revert "Fossa seems to be stalling builds, we will re-enable this soon"

cbron:
Reverts rancher/rancher#35793


cbron:
https://github.com/rancher/rancher/pull/35879

PR #35862, opened 2021-12-13
Fix issue w/ upgrade and *-promoted roles

ryansann:


pkg/controllers/managementuser/rbac/prtb_handler.go
```diff
@@ -271,20 +271,31 @@ func (m *manager) checkForGlobalResourceRules(role *v3.RoleTemplate, resource st
 }
 
 // Ensure the clusterRole used to grant access of global resources to users/groups in projects has appropriate rules for the given resource and verbs
-func (m *manager) reconcileRoleForProjectAccessToGlobalResource(resource string, rt *v3.RoleTemplate, newVerbs map[string]bool) (string, error) {
+func (m *manager) reconcileRoleForProjectAccessToGlobalResource(resource string, rt *v3.RoleTemplate, newVerbs map[string]bool, baseRule rbacv1.PolicyRule) (string, error) {
 	clusterRoles := m.workload.RBAC.ClusterRoles("")
 	roleName := rt.Name + "-promoted"
 	if role, err := m.crLister.Get("", roleName); err == nil && role != nil {
 		currentVerbs := map[string]bool{}
+		currentResourceNames := map[string]struct{}{}
```
cbron, line 279: Asked Ryan and there is no point in the `struct{}{}` other than to occupy no memory space. It could have been `map[string]bool`

cbron (approved):

pkg/controllers/managementuser/rbac/prtb_handler.go
ryansann: Yea, values are not needed here, we're just checking existence on keys. And a map is used so ordering of resource names isn't significant

prachidamle (commented):
One comment to follow up as a separate issue - Does this issue apply to all the other resources mentioned in the globalResources list? Will users get access to list all persistentVolumes/clusterrepos/storageclasses ?https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementuser/rbac/prtb_handler.go#L29-L53

prachidamle (approved):

Issue #35797, opened 2021-12-08
Local k3s clusters were not able to be upgraded within Rancher post upgrade from 2.5 to 2.6


deniseschannon:
This seems to be fixed with the latest SUC version introduced in 2.6.3, but this is a explicit request to test this use case to check.

SURE-3539
SURE-3388

Before rancher server upgrade: Rancher version - v2.5.9, k3s v1.20.9+k3s1, SUC: 0.6.2

Post rancher server upgrade: Rancher version - v2.6.2, k3s v1.20.11+k3s2, SUC: 0.7.5

1. Created a k3s HA cluster on the version 1.20.9
2. Installed rancher v2.5.9 on this k3s cluster 
3. Upgraded the k3s version to v1.20.11+k3s2
4. Verified all the nodes were upgraded - SUC version 0.6.2  and the service account name is system-upgrade 
`kubectl get serviceaccount -A | grep -i system-upgrade`
apps & market --> installed apps --> SUC app installed
5. Upgraded the rancher version to v2.6.2
6. Verified all rancher pods were upgraded. SUC version is 0.7.5
7. Upgraded the k3s version in the local cluster from the UI to v1.21.4+k3s2 
8. The initial node, the upgrade was attempted on was stuck in `cordoned`  and the local cluster is stuck in `upgrading` and the service account name is system-upgrade-controller
9. Following error is seen in the cluster events:
Error creating: pods "apply-k3s-master-plan-on-ip-172-31-8-247-with-57eb7803114-9fcec-" is forbidden: error looking up service account cattle-system/system-upgrade: serviceaccount "system-upgrade" not found

snasovich:
Assigning myself for tracking purposes.
@sowmyav27 , please assign QA and test. Thank you.

jmcsagdc:
verified fix in 	v2.6-e62e1ae34eca46bd733b7f67d1cea055395ac45d-head

1. Created a k3s HA cluster on the version 1.20.9
2. Installed rancher v2.5.9 on this k3s cluster
3. Upgraded the k3s version to v1.20.12+k3s1
4. Verified all the nodes were upgraded
5. kubectl get serviceaccount -A | grep -i system-upgrade (present as expected)
6. apps & market --> installed apps --> SUC app installed (yes)
7. Upgraded the rancher version to v2.6.2
8. Verified all rancher pods were upgraded. SUC version is 0.6.2 
9. Upgraded the k3s version in the local cluster from the UI to v1.22.4+k3s1

[...]

PR #35870, opened 2021-12-15
set default serviceAccountName to k3s upgrader plan on updates

kinarashah:
Issue:
Starting Rancher v2.6.0, serviceAccountName for k3s upgrader genericPlan was updated. But genericPlan is used only for creates (generateMasterPlan and generateWorkerPlan). On updates (configureMasterPlan and configureWorkerPlan), it picks the name from existing plan so it gets the old svc account name if the plans were created before Rancher 2.6.

https://github.com/rancher/rancher/issues/35797


dweomer (approved):

cbron:
Testing another use case....

thedadams (approved):

kinarashah:
Tested on downstream k3s clusters and local k3s cluster.

Issue #35700, opened 2021-12-01
Deleting an imported cluster that was unavailable doesn't fully delete it


rancher-max:
**Rancher Server Setup**
- Rancher version: v2.6.3-rc2
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: N/A
- Cluster Type (Local/Downstream): Imported k3s cluster hosted on vsphere nodes
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**Describe the bug**

After deleting an imported cluster that was unavailable, it failed to fully delete. It no longer shows up in the Cluster Management page, but it still shows up in the dashboard home page, in the sidebar for clusters, and in the Advanced -> Mgmt Clusters section and I cannot figure out how to get rid of it.

**To Reproduce**
 
1. Create a cluster and import it into rancher
2. Delete the nodes from the cluster so that rancher sees it as unavailable
3. Delete the cluster from cluster management page in rancher

**Result**
Cluster appears to be deleted but remnants of it hang around

**Expected Result**

The deleted cluster should be fully deleted from all views 

**Screenshots**

Multiple examples in the same setup:
![image (5)](https://user-images.githubusercontent.com/55997940/144149179-417ad27b-8785-4d06-9f5d-bb65be4912bc.png)
![image (4)](https://user-images.githubusercontent.com/55997940/144149182-20fcb101-a842-4939-981b-5759c99fad47.png)
![image (3)](https://user-images.githubusercontent.com/55997940/144149184-ff62ab9d-768c-4812-898d-8f17e8261c68.png)

Trying to delete from Mgmt Clusters gets this error
<img width="1075" alt="image" src="https://user-images.githubusercontent.com/55997940/144149299-5ed8681a-4185-4474-84de-c9ffc52c4cbf.png">

**Additional context**

Jono-SUSE-Rancher:
Seems very similar to https://github.com/rancher/rancher/issues/34650

jakefhyde:
### Root cause
On deletion of a cluster, a specific error was expected, due to trying to start controllers in the downstream. The type of error was changed because these controllers are no longer started when the cluster is being deleted.

### What was fixed, or what changes have occurred
Treat cluster agent is disconnected as an acceptable error and continue with cleaning up the resources in the local cluster. 

### Areas or cases that should be tested
[...]

PR #35876, opened 2021-12-15
Cluster cleanup now ignores cluster agent disconnected error

jakefhyde:
### Background
 On error of cleaning up an imported cluster, the `clusterUnavailable` error was expected, however this was because attempting to get the `UserContext` was always trying to create controllers, which would fail if the deletion timestamp was not nil. Now that this is not necessary, the cleanup is able to use a `UserContext` without creating any controllers, and therefore fails when trying to create a cluster role on the unavailable downstream, thus generating a `url.Error` with `Error() == "cluster agent disconnected"`.

### Changes
This PR fixes the above issue by checking for `cluster agent disconnected` instead of `httperror.ClusterUnavailable`.

## Notes
- It may be necessary to hard refresh (F5) to see these changes in effect, as my imported cluster stayed in the sidebar until doing so (after about 30 minutes).

Related Issue: https://github.com/rancher/rancher/issues/35700

### DIFF
--- a/pkg/controllers/management/usercontrollers/controller.go
+++ b/pkg/controllers/management/usercontrollers/controller.go
@@ -2,14 +2,15 @@ package usercontrollers
 
 import (
 	"context"
+	"errors"
 	"time"
 
-	"github.com/rancher/norman/httperror"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/clustermanager"
 	"github.com/rancher/rancher/pkg/controllers/management/imported"
 	"github.com/rancher/rancher/pkg/controllers/managementagent/nslabels"
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy/helm"
+	"github.com/rancher/rancher/pkg/dialer"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/types/config"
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/usercontrollers/controller.go
```diff
@@ -82,13 +82,8 @@ func (c *ClusterLifecycleCleanup) Remove(obj *v3.Cluster) (runtime.Object, error
...
 	if err != nil {
-		apiError, ok := err.(*httperror.APIError)
-		// If it's not an API error give it back
-		if !ok {
-			return nil, err
-		}
-		// If it's anything but clusterUnavailable give it back
-		if apiError.Code != httperror.ClusterUnavailable {
+		// If it's anything but cluster agent disconnected give it back
+		if !strings.Contains(err.Error(), dialer.ErrAgentDisconnected.Error()) {
```
thedadams, line 86: I think this is a more idiomatic approach.
```suggestion
		if errors.Is(err, dialer.ErrAgentDisconnected) {
```

thedadams (changes requested):

cmurphy:
> however because the controllers are no longer started in this path the api request to create the cleanup cluster role was instead failing with a different error.

Could you explain this more? What does "no longer started in this path" mean, what changed? What's the different error, is that the cluster disconnected error or is it something else we should address?

jakefhyde:
@cmurphy The issue was caused by the change from 
https://github.com/rancher/rancher/blob/26ce1db8dff7cfca12cfafa7198979bdcb346e34/pkg/controllers/management/usercontrollers/controller.go#L101
to
https://github.com/rancher/rancher/blob/bbde38e8cbf1007ba81ce5e68f674948a0033c0d/pkg/controllers/management/usercontrollers/controller.go#L136
because in the former, an attempt to start the controllers is made, but it bailed out with `clusterUnavailable` when it sees the deletion timestamp is not nil when generating the rest config. Should've clarified that starting the controllers actually has nothing to do with it, just the side effect of attempting to start the controllers failing and generating the error that was expected. It's also worth noting that previously, any error besides cluster not found would be discarded and `clusterUnavailable` would be returned instead:
https://github.com/rancher/rancher/blob/26ce1db8dff7cfca12cfafa7198979bdcb346e34/pkg/clustermanager/manager.go#L453-L488
Also yeah, `cluster agent disconnected` is the error as the POST request fails for creating the cluster cleanup role.

cmurphy (approved):
Thanks for the explanation!

thedadams (approved):

Issue #35880, opened 2021-12-15
Metrics requests hang for downstream cluster after project is created


rmweir:
**Rancher Server Setup**
- Rancher version: v2.6.3-rc6

**Information about the Cluster**
- Kubernetes version:
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):
  do, rke v1.21.7

**Describe the bug**

After creating project in downstream cluster, requesting the cluster's metrics will never return.

**To Reproduce**
 
1. provision downstream cluster
2. go to metrics endpoint for cluster: `<rancher-url>/metrics/<cluster-id>`
3. go to cluster exporer for downstream cluster and create project
4. go to metrics endpoint for cluster: `<rancher-url>/metrics/<cluster-id>`

**Result**
request never returns

**Expected Result**

request returns just as it did in step 2

**Additional context**

I think this is related to a wider issue around how we are deferring handlers. For now, to resolve this immediate issue, one of the commits will be revert. After release the underlying problem should be investigated further.

First observed after this commit: https://github.com/rancher/rancher/commit/0f084bf88bd7056686465dfd685152e549379ae9#diff-8b9c5a67f3c8ee7b7b23a1c4c37b181f8609b84f77e5af638fa6d674d09af32dR44

Looking at profile it seems that rancher is sitting in lock for a suspicious amount of time:
![image](https://user-images.githubusercontent.com/19376037/146272105-09d961dc-bd17-45ad-845a-851a3de03c06.png)

aiyengar2:
Steps to reproduce:

- [Rancher] Provisioned 1 node RKE `v1.21.6-rancher1-1` cluster
  - Installed Rancher onto the Rancher RKE cluster via helm install based on Rancher `v2.6.3-rc6`
  - [Downstream Provisioned] Used DO node driver provisioner to provision a RKE1 cluster with `v1.21.7-rancher1-1`
- [Downstream Imported] Provisioned 1 node RKE `v1.21.6-rancher1-1` cluster
  - Imported it into Rancher via Cluster Dashboard UI

On both Downstream Provisioned (`c-nq4l2`) and Downstream Imported (`c-m-kpwp2556`), I took the following steps:
- Try to access metrics (e.g. `https://RANCHER_URL/metrics/CLUSTER_ID`) - No errors expected
- Go to **Cluster Explorer** (not Cluster Manager) and add a Project
[...]

PR #35884, opened 2021-12-16
Don't block controller thread with deferred start

ibuildthecloud:
If two controller goroutines both run deferred start it is possible to deadlock
as one goroutine is holding the start lock and waiting for the controllers
to start and the other gorouting is trying to get the start lock to wait
for the controllers to start.  Since both can not move forward the
controller start is deadlocked.

### DIFF
--- a/pkg/types/config/context.go
+++ b/pkg/types/config/context.go
@@ -242,6 +242,14 @@ type UserContext struct {
 }
 
 func (w *UserContext) DeferredStart(ctx context.Context, register func(ctx context.Context) error) func() error {
+	f := w.deferredStartAsync(ctx, register)
+	return func() error {
+		go f()
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


pkg/types/config/context.go
```diff
@@ -242,6 +242,14 @@ type UserContext struct {
 }
 
 func (w *UserContext) DeferredStart(ctx context.Context, register func(ctx context.Context) error) func() error {
+	f := w.deferredStartAsync(ctx, register)
```
rmweir, line 245: nit: f isn't a very informative function name. Maybe deferredRegister or something.

rmweir (approved):
LGTM, with 1 nit.

rmweir:
https://github.com/rancher/rancher/issues/35880

PR #35901, opened 2021-12-16
Ignore cluster agent restart for local cluster

cmurphy:
There is no cluster agent on the local cluster so it does not need to be
restarted. Attempting to look up the local cluster agent and update it
was failing because the Deployment object returned from Get was non-nil
but empty, so trying to Update that object was invalid. This change 1)
skips the agent restart for local clusters, 2) moves the restart to
after the cluster update so that any errors that occur don't interfere
with saving the version on the cluster object, and 3) fixes the
deployment lookup to check for Not Found rather than nil.

https://github.com/rancher/rancher/issues/35902

### DIFF
--- a/pkg/controllers/management/clusterstats/statsaggregator.go
+++ b/pkg/controllers/management/clusterstats/statsaggregator.go
@@ -174,10 +174,15 @@ func (s *StatsAggregator) aggregate(cluster *v3.Cluster, clusterName string) err
 	}
 	versionChanged := s.updateVersion(cluster)
 
+	if statusChanged(origStatus, &cluster.Status) || versionChanged {
+		_, err = s.Clusters.Update(cluster)
+		return err
+	}
+
 	// If the cluster went through an upgrade from <=1.21 to >=1.22, restart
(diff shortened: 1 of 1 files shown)


kinarashah (approved):

prachidamle:
LGTM

pkg/controllers/management/clusterstats/statsaggregator.go
```diff
@@ -232,19 +232,20 @@ func (s *StatsAggregator) restartAgentDeployment(cluster *v3.Cluster) error {
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
```
thedadams, line 237: nit:
```suggestion
	if apierrors.IsNotFound(err) {
		return nil
	} else if err != nil {
		return err
	}
```

thedadams (approved):
One nit. Feel free to ignore.

prachidamle (approved):

Issue #35904, opened 2021-12-16
Editing/updating existing catalog apps Validation failed in API: rancher max version exceeded


anupama2501:
Information about the Cluster
**Rancher Server Setup**
- Rancher version: v2.6.3-rc6, upgraded from v2.6.2
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1, v1.20.11, 1.2.13
- Proxy/Cert Details: Selfsigned

**Information about the Cluster**
- Kubernetes version: v1.21.7
- Cluster Type (Local/Downstream): `Downstream node driver 3 worker, 1 etcd, 1 cp RKE1`

**Describe the bug**
Not able to update/edit apps that are created in v2.6.2 and below which are compatible with the rancher versions. 

**To Reproduce**
1. Create a rancher server v2.6.2
2. Install any legacy catalog app eg., argo-crd, docker-registry etc. in the default namespace
3. Upgrade the rancher server to v2.6.3-rc6
4. Verify the legacy feature flag is enabled by default. 
5. From downstream cluster --> legacy --> project --> apps select the argo-crd or docker-registry

**Result**
App upgrade fails with the error 
```
Validation failed in API: rancher max version exceeded
```

**Expected Result**
Updating/editing the app should go through as the apps were already installed from the previous version.

**Additional context**
https://github.com/rancher/rancher/issues/35680
https://github.com/rancher/ui/pull/4791

rmweir:
QA Note:
This fix should address updates and rollbacks for both regular apps and multi cluster apps.

brendarearden:
Available to test in rancher/rancher v2.6.3-rc8

anupama2501:
Verified the following on v2.6.2 --> v2.6.3-rc8 and v2.5.11 --> v2.6.3-rc8

-  On 2.6.2, install the apps

1. library apps: argo-cd, docker-registry
2. Legacy apps: monitoring
3. v2 charts: monitoring, istio

- Upgrade the rancher server to v2.6.3-rc8
- Verified from 2.5.11 --> 2.6.3-rc8, legacy flag is enabled by default
- from explorer --> project --> default project --> apps --> Updated the mariadb, mysql, argo-cd, docker-registry
- Verified the upgrade goes through without an error
- Uninstalled the mariadb and argo-cd --> works as expected
- Rolled back redis and verified roll back was successful
[...]

PR #35912, opened 2021-12-17
Accept preexisting incompatible app versions

rmweir:
**Problem:**
Prior, if an app that was installed prior was rendered incompatible
due to a rancher upgrade, updates to the app would fail. This would
happen even if the update was not to the app version.

**Solution:**
Now, apps can
be updated if their current version is incompatible as long as they
are not updating to another incompatible app version.

**Issue:**
https://github.com/rancher/rancher/issues/35904

### DIFF
--- a/pkg/api/norman/customization/app/app.go
+++ b/pkg/api/norman/customization/app/app.go
@@ -169,7 +169,12 @@ func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiConte
 	case "upgrade":
 		externalID := convert.ToString(actionInput["externalId"])
 
-		if err := w.validateChartCompatibility(externalID, clusterName); err != nil {
+		_, _, _, _, currentAppVersion, err := hcommon.SplitExternalID(obj.Spec.ExternalID)
+		if err != nil {
+			httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("could not parse current externalID [%s]: %s",
+				obj.Spec.ExternalID, err.Error()))
+		}
+		if err := w.validateChartCompatibility(externalID, clusterName, currentAppVersion); err != nil {
 			return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
 		}
 
@@ -227,7 +232,12 @@ func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiConte
 			return err
 		}
 
-		if err := w.validateChartCompatibility(appRevision.Status.ExternalID, clusterName); err != nil {
+		_, _, _, _, currentAppVersion, err := hcommon.SplitExternalID(obj.Spec.ExternalID)
+		if err != nil {
+			httperror.NewAPIError(httperror.InvalidBodyContent, fmt.Sprintf("could not parse current externalID [%s]: %s",
+				obj.Spec.ExternalID, err.Error()))
+		}
+		if err := w.validateChartCompatibility(appRevision.Status.ExternalID, clusterName, currentAppVersion); err != nil {
 			return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
 		}
 
@@ -282,7 +292,7 @@ func (w Wrapper) LinkHandler(apiContext *types.APIContext, next types.RequestHan
 	return nil
 }
 
-func (w Wrapper) validateChartCompatibility(externalID, clusterName string) error {
+func (w Wrapper) validateChartCompatibility(externalID, clusterName, currentAppVersion string) error {
 	if externalID == "" {
 		return nil
 	}
@@ -294,5 +304,5 @@ func (w Wrapper) validateChartCompatibility(externalID, clusterName string) erro
 	if err != nil {
 		return err
 	}
-	return w.CatalogManager.ValidateChartCompatibility(template, clusterName)
+	return w.CatalogManager.ValidateChartCompatibility(template, clusterName, currentAppVersion)
 }
--- a/pkg/catalog/manager/manager.go
+++ b/pkg/catalog/manager/manager.go
@@ -45,9 +45,9 @@ type Manager struct {
 }
 
 type CatalogManager interface {
-	ValidateChartCompatibility(template *v3.CatalogTemplateVersion, clusterName string) error
+	ValidateChartCompatibility(template *v3.CatalogTemplateVersion, clusterName, currentAppVersion string) error
 	ValidateKubeVersion(template *v3.CatalogTemplateVersion, clusterName string) error
-	ValidateRancherVersion(template *v3.CatalogTemplateVersion) error
+	ValidateRancherVersion(template *v3.CatalogTemplateVersion, currentAppVersion string) error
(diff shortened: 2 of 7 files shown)


pkg/catalog/manager/manager.go
```diff
@@ -242,7 +242,13 @@ func (m *Manager) ValidateKubeVersion(template *v3.CatalogTemplateVersion, clust
 	return nil
 }
 
-func (m *Manager) ValidateRancherVersion(template *v3.CatalogTemplateVersion) error {
+func (m *Manager) ValidateRancherVersion(template *v3.CatalogTemplateVersion, currentAppVersion string) error {
+	if currentAppVersion != "" && currentAppVersion == template.Spec.Version {
```
aiyengar2, line 246: nit: I think `currentAppVersion` should be a `*string`, where `nil` indicates that a current app version does not exist.

pkg/catalog/manager/manager.go
```diff
@@ -242,7 +242,13 @@ func (m *Manager) ValidateKubeVersion(template *v3.CatalogTemplateVersion, clust
 	return nil
 }
 
-func (m *Manager) ValidateRancherVersion(template *v3.CatalogTemplateVersion) error {
+func (m *Manager) ValidateRancherVersion(template *v3.CatalogTemplateVersion, currentAppVersion string) error {
```
aiyengar2, line 245: Knowing the current version of the chart does not change the fact that this chart is not valid for this version of Rancher; rather, it only indicates that this chart is still **compatible** with your current Rancher setup to allow it to be upgraded as we simply **bypass** the rancher version check in this case.

Therefore, I don't think we should be modifying the definition of `ValidateRancherVersion`; IMO this condition should be relocated to `ValidateChartCompatibility`. ​

This would also allow us to remove the changes from `pkg/api/norman/customization/catalog/store.go` and `pkg/api/norman/customization/multiclusterapp/multi_cluster_app.go`; for `pkg/api/norman/customization/multiclusterapp/store.go`, I think we should introduce this bypass logic directly into code instead.

---

Optionally, you could turn this into a helper function like `func (m *Manager) IsUpgrade(template *v3.CatalogTemplateVersion, currentAppVersion string)` and call it here and in `pkg/api/norman/customization/multiclusterapp/store.go` to avoid redundant code.

pkg/api/norman/customization/app/app.go
```diff
@@ -227,7 +232,7 @@ func (w Wrapper) ActionHandler(actionName string, action *types.Action, apiConte
 			return err
 		}
 
-		if err := w.validateChartCompatibility(appRevision.Status.ExternalID, clusterName); err != nil {
+		if err := w.validateChartCompatibility(appRevision.Status.ExternalID, clusterName, ""); err != nil {
```
aiyengar2, line 235: I'm not too familiar with this particular API, but what would be the outcome of this scenario?

1. User installs App at version x.y.z onto a cluster in Rancher 2.6.3, where this chart has a Rancher max version of 2.6.3
2. User upgrades App in-place (version remains x.y.z) with new values.yaml overrides
3. User upgrades Rancher to 2.6.4
4. User attempts to rollback values.yaml overrides

Would the rollback fail even though the version remained `x.y.z` because the Rancher version is re-evaluated? Or does a rollback strictly mean that you must go back to a previous version (e.g. `x.y.(z-1`)?

aiyengar2 (changes requested):

pkg/catalog/manager/manager.go
rmweir: @aiyengar2 I think have this logic in ValidateRancherVersion makes sense because place are bypassing ValidateChartCompatibility and just using ValidateRancherVersion. I also thing the changes in the store and mcapp are desirable. Mcapp is subject to these same things. It is possible an app would be updated using the API/API UI and in this case we still want this validated, that is why we should leave it in the store. We could use a pointer but I'm not sure if it is wise to use pointers where they are no necessary as it invites another point of failure. An empty string indicating that it isn't being used is not much different than using nil.

pkg/api/norman/customization/app/app.go
rmweir: That is already how it works today, and I think changing that would increase the surface area of this change as the issue is not concerned with rollbacks.

pkg/catalog/manager/manager.go
rmweir: We could use a pointer but I'm not sure if it is wise to use pointers where they are no necessary as it invites another point of failure. An empty string indicating that it isn't being used is not much different than using nil.

aiyengar2: hmm agreed. resolving, no need to change

brendarearden (dismissed):
lgtm! 

nit: I would like to see some test cases around this added in the future but I understand this is a last minute update for release.

pkg/catalog/manager/manager.go
aiyengar2: re:
> @aiyengar2 I think have this logic in ValidateRancherVersion makes sense because place are bypassing ValidateChartCompatibility and just using ValidateRancherVersion.

Any part of the code that is bypassing `ValidateChartCompatibility` in favor of directly calling `ValidateRancherVersion` is effectively running its own custom compatibility check logic (e.g. mcapps). 

And I still think that checking whether the template would trigger an app upgrade (and therefore choosing to bypass the Rancher version check) is independent of checking whether a template is valid in a given Rancher version. Whether it would trigger an app upgrade and require additional checks (such as a Rancher version check) is a concern of whatever is executing the compatibility check, whether it be `ValidateChartCompatibility` or a custom compatibility check directly introduced within the code (like mcapps does).

To facilitate areas in the code like mcapps  that are using custom logic instead of the standard `ValidateChartCompatibility`, I think we should introduce `IsUpgrade(template *v3.CatalogTemplateVersion, currentAppVersion string) bool` to `CatalogManager` that has this check so that mcapps could use it as:

```go
if s.catalogManager.IsUpgrade(template, mcappCurrentTemplateVersion) {
    if err := s.catalogManager.ValidateRancherVersion(template, mcappCurrentTemplateVersion); err != nil {
	return httperror.NewAPIError(httperror.InvalidBodyContent, err.Error())
    }
}
```

I believe this should address the concern in:

> It is possible an app would be updated using the API/API UI and in this case we still want this validated, that is why we should leave it in the store.

This could also be how we call it in `ValidateChartCompatibility` as well.

---

re:
> I also thing the changes in the store and mcapp are desirable.

Maybe I'm missing something, but why would this be desirable?

---

aiyengar2 (changes requested):

pkg/catalog/manager/manager.go
aiyengar2: discussed offline and decided to stick with the current approach on the PR

pkg/api/norman/customization/app/app.go
rmweir: I forgot that rollbacks can be to a template with the same version. I agree that this should be included and have updated.

rmweir:
@brendarearden I agree, tests in the future would be good.

kinarashah (approved):
LGTM

aiyengar2 (approved):

brendarearden (approved):

Issue #35275, opened 2021-10-26
Support for automatic replacement of unreachable nodes in RKE2 machine pool


snasovich:
A new field should be supported on machine pool definition to specify the threshold for how long a node can stay unreachable before it's replaced. TBD if the option should be added to `provisioning.cattle.io.clusters -> spec.machinePools[i]` object or something else.

Then, support for logic utilizing this new field should be added to the V2 provisioning engine. This should be supported on CAPI (Cluster API) side per https://github.com/kubernetes-sigs/cluster-api/issues/1990.

paynejacob:
### Root cause
 
 new feature
 
### What was fixed, or what changes have occurred

Health checks were added to support self healing node pools
 
### Areas or cases that should be tested

- node failures
- upgrades
- scaling up and down
 
### What areas could experience regressions?

- upgrades
- scaling up and down
 
### Are the repro steps accurate/minimal?

see this comment for test cases https://github.com/rancher/rancher/pull/35916#issue-1083643527

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head
Rancher cluster type: HA
Docker version: 20.10

Downstream cluster type:

----------

### Testing: ###

Tested this issue with the following scenarios:
- upgraded k8s version
- forced node failure inside nodeStartupTimeout: node was replaced after nodeStartupTimeout
- forced node failure outside of nodeStartupTimeout: node was replaced within 1 to 2 UnhealthyNodeTimeout periods
- forced multiple node failures exceeding MaxUnhealthy: nodes were marked failed but not replaced. After - manually deleting enough nodes to be below the MaxUnhealthy: the remaining nodes were automatically replaced.

**Result**


k8s v1.21 - Fail

Cluster seems to change state to "Provisioning" but does not seem to be provisioning another node. Cluster is stuck in this state and nothing happens.
- forced node failure inside nodeStartupTimeout: node was **not** replaced after nodeStartupTimeout
![image.png](https://zube.io/files/rancher/956f559962aaa74384cc58d0142fd65c-image.png)

- forced node failure outside of nodeStartupTimeout: node was **not** replaced within 1 to 2 UnhealthyNodeTimeout periods
![image.png](https://zube.io/files/rancher/956f559962aaa74384cc58d0142fd65c-image.png)

[...]

PR #35916, opened 2021-12-17
self-healing machine pools for rke2

paynejacob:
#35275 

# Solution
Using CAPI machine health checks we can replace the machine associated with an unready node after a given period of time.  CAPI health checks require all other components in  a cluster to  have  a ready condition to replace nodes.  Once the cluster's infrastructure is ready, the control plane is initialized and the machine is ready the machine health check controller will start checking the related node's conditions on the interval defined by the node pool's `UnhealthyNodeTimeout`.   If the node  goes unhealthy before the `nodeStartupTimeout` it will be marked as unhealthy but not replaced.  This is also the case if the number of nodes in a pool that are unhealthy exceeds the `MaxUnhealthy` or is not in the `UnhealthyRange`.  If all these conditions are not true the machine will be given the remediation condition and the machine set will replace it.   

# Testing

All node failures were simulated by calling `shutdown -P now` on the node. Theses tests were done with controlplane and standard worker nodes.

Test Cases:
- upgraded k8s version
-  forced node failure inside `nodeStartupTimeout`: node was replaced after `nodeStartupTimeout`
- forced node failure outside of `nodeStartupTimeout`: node was replaced within 1 to 2 `UnhealthyNodeTimeout` periods
- forced multiple node failures exceeding `MaxUnhealthy`: nodes were marked failed but not replaced.  After manually deleting enough nodes to be below the `MaxUnhealthy` the remaining nodes were automatically replaced.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
+++ b/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
@@ -360,9 +360,48 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 		}
 
 		result = append(result, machineDeployment)
+
+		// if a health check timeout was specified create health checks for this machine pool
+		if machinePool.UnhealthyNodeTimeout != nil {
+			hc := deploymentHealthChecks(machineDeployment, machinePool)
+			result = append(result, hc)
+		}
 	}
 
 	return result, nil
+}
+
+// deploymentHealthChecks Health checks will mark a machine as failed if it has any of the conditions below for the duration of the given timeout. https://cluster-api.sigs.k8s.io/tasks/healthcheck.html#what-is-a-machinehealthcheck
+func deploymentHealthChecks(machineDeployment *capi.MachineDeployment, machinePool rancherv1.RKEMachinePool) *capi.MachineHealthCheck {
+	return &capi.MachineHealthCheck{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      machineDeployment.Name,
+			Namespace: machineDeployment.Namespace,
+		},
+		Spec: capi.MachineHealthCheckSpec{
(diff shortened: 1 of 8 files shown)


pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -356,11 +357,50 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 		}
 
 		result = append(result, machineDeployment)
+
+		// if a health check timeout was specified create health checks for this machine pool
+		if machinePool.UnhealthyNodeTimeout != nil {
+			hc := deploymentHealthChecks(machineDeployment, machinePool.NodeStartupTimeout, machinePool.UnhealthyNodeTimeout, machinePool.MaxUnhealthy, machinePool.UnhealthyRange)
```
thedadams, line 363: nit: just pass the `machinePool` to this function to make the signature simpler.

thedadams (commented):
Since this is a draft PR, then I will make this a comment for now. Although, I think this looks good.

jakefhyde (approved):

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -360,11 +360,50 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 		}
 
 		result = append(result, machineDeployment)
+
+		// if a health check timeout was specified create health checks for this machine pool
+		if machinePool.UnhealthyNodeTimeout != nil {
+			hc := deploymentHealthChecks(machineDeployment, machinePool)
+			result = append(result, hc)
```
thedadams, line 367: nit:
```suggestion
			result = append(result, deploymentHealthChecks(machineDeployment, machinePool))
```

thedadams (approved):

Issue #35917, opened 2021-12-17
Ensure SUC charts version is updated


thedadams:
After adding a new version of `system-upgrade-controller` to rancher/charts, there is still the need to update the chart version in `pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go`.

This second step is often forgotten.
1. It should be more obvious where/when this chart version needs to be updated in Rancher.
2. It should be considered whether or not to put in a check in CI that the version is correct/most recent version.

thedadams:
### What was fixed, or what changes have occurred
The SUC chart version was changed to be like rancher-webhook, fleet, etc.

### Areas or cases that should be tested
Deploy an RKE2 cluster and ensure that SUC chart with version 100.0.1 is deployed and the SUC image has version v0.3.1

### What areas could experience regressions?
SUC could not be deployed at all.

### Are the repro steps accurate/minimal?
Yes.

sowmyav27:
Validate usecases:
- Deploy rke2 1.21 cluster and verify SUC is deployed with expected images.
```
- - 3 etcd, 2 control, 3 worker nodes (Custom RKE2 cluster)
- - Cluster should come up Active
```
- Deploy rke2 1.21  cluster and verify SUC is deployed with expected images.
```
- - 3 etcd, 2 control, 3 worker nodes (AWS node driver cluster)
- - Cluster should come up Active
- - upgrade k8s version to 1.22RKE2 version
```

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head c682e30
Rancher cluster type: HA
Docker version: 20.10

Downstream cluster type: RKE2 AWS node driver and RKE2 Custom 

----------

### Testing: ###

Tested this issue with the following steps:

**Custom RKE2 Cluster**
1. Deploy an RKE2 1.21 cluster
2. Verify SUC is deployed with expected images

**AWS node driver cluster**
1. Deploy an RKE2 1.21 cluster
2. Upgrade k8s version to 1.22 RKE2 version
3. Verify SUC is deployed with expected images

**Result** - Pass
The image version should be v0.8.1 according to github
![Screen Shot 2022-02-02 at 5.15.08 PM.png](https://zube.io/files/rancher/7d5ebe8378d6070e2d53f1e227cd77b8-screen-shot-2022-02-02-at-5-15-08-pm.png)

**Custom RKE2 cluster**
Cluster is active and SUC is also running with the expected image
[...]

PR #35920, opened 2021-12-18
Add SUC chart version environment variable

thedadams:
The SUC chart version was buried in the Rancher code. This change make the
version a setting and adds an environment variable that is similar to the other
versions that are needed.

The goal for this change is to make updating this version similar to how other
versions are updated, so it is less likely to be missed.

Issue:
https://github.com/rancher/rancher/issues/35917


jiaqiluo (approved):
LGTM.

I can review it again after the conflict is addressed.

kinarashah (approved):

Issue #35922, opened 2021-12-18
[airgap] unable to provision k3s custom cluster -> images are not being pulled from systemDefaultRegistry


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.6.3-rc7
- Installation option (Docker install/Helm Chart): helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): rke1
- Proxy/Cert Details: valid certs, airgap setup

**Information about the Cluster**
- Kubernetes version: 1.21.7+k3s1
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): custom

**Describe the bug**

custom k3s cluster is not able to register with rancher. Investigation on the node reveals that images are being pulled from docker.io, but should be pulled from the systemDefaultRegistry

**To Reproduce**
 
* deploy an airgap setup
* deploy a custom k3s cluster through rancher (k3s, 1.21.7+k3s1) -- tested with single node deployment 
* run the registration cmd on a private node 

**Result**
node will stay in `provisioning` state in rancher. 
example error from one of the pods from the node:
```
Warning  FailedCreatePodSandBox  23s (x4 over 4m2s)     kubelet            (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "rancher/pause:3.1": failed to pull image "rancher/pause:3.1": failed to pull and unpack image "docker.io/rancher/pause:3.1": failed to resolve reference "docker.io/rancher/pause:3.1": failed to do request: Head "https://registry-1.docker.io/v2/rancher/pause/manifests/3.1": dial tcp 34.198.213.42:443: i/o timeout
```

**Expected Result**

Investigating the node shows that all images being pulled are from docker.io instead of from the registry. 

**Screenshots**

<img width="742" alt="Screen Shot 2021-12-17 at 4 26 53 PM" src="https://user-images.githubusercontent.com/16691014/146622463-d6e22a7e-25f9-487a-acc6-27a42bb38699.png">

**Additional context**

* rke2 custom nodes in the same setup are getting to an `active` state but k3s is not
* `/etc/rancher/k3s/registries.yaml` shows no config:

```
{"configs":{},"mirrors":null}
```

brendarearden:
Available to test in rancher/rancher v2.6.3-rc9

slickwarren:
[...]

PR #35923, opened 2021-12-18
Set system-default-registry for K3S clusters

thedadams:
An if-statement kept the system-default-registry setting from being added to the
config for K3S clusters. This was a mistake. This change will set the
system-default-registry setting regardless of runtime.

Issue:
https://github.com/rancher/rancher/issues/35922


brendarearden (approved):
looks good to me - confirmed that this change will only affect rke2 and k3s cluster types .

Oats87 (approved):

PR #35925, opened 2021-12-18
Leave out bootstrap secret if it is optional

thedadams:
When provisioning a machine, if the bootstrap secret is optional and not found,
then the name will be changed. This causes Wrangler Apply to delete the job and
recreate it with the new name, which in turn may cause the job to hang because
the process of create -> delete -> create is too fast for Kubernetes to handle.

Now, the volume and volume mount are only include if the bootstrap secret is
required. This keeps the job from being deleted and recreated when the bootstrap
secret cannot be found.

In addition, this change also adds tests for deleting clusters. Now, after we ensure
a cluster is created successfully, the cluster is deleted and we ensure the cluster
is cleaned up successfully.

Related Issue:
https://github.com/rancher/rancher/issues/35795

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
@@ -45,7 +45,7 @@ type driverArgs struct {
 	EnvSecret           *corev1.Secret
 	StateSecretName     string
 	BootstrapSecretName string
-	BootstrapOptional   bool
+	BootstrapRequired   bool
 	Args                []string
 }
 
@@ -135,7 +135,7 @@ func (h *handler) getArgsEnvAndStatus(meta metav1.Object, data data.Object, args
 		EnvSecret:           secret,
 		StateSecretName:     secretName,
 		BootstrapSecretName: bootstrapName,
-		BootstrapOptional:   !create,
+		BootstrapRequired:   create,
 		Args:                cmd,
 		RKEMachineStatus: rkev1.RKEMachineStatus{
 			Ready:                     data.String("spec", "providerID") != "" && data.Bool("status", "jobComplete"),
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
@@ -41,27 +41,33 @@ type driverArgs struct {
 
 	DriverName          string
 	ImageName           string
+	MachineName         string
+	MachineNamespace    string
+	MachineGVK          schema.GroupVersionKind
 	ImagePullPolicy     corev1.PullPolicy
 	EnvSecret           *corev1.Secret
+	FilesSecret         *corev1.Secret
 	StateSecretName     string
 	BootstrapSecretName string
 	BootstrapRequired   bool
 	Args                []string
+	BackoffLimit        int32
 }
 
 func MachineStateSecretName(machineName string) string {
 	return name2.SafeConcatName(machineName, "machine", "state")
 }
 
-func (h *handler) getArgsEnvAndStatus(meta metav1.Object, data data.Object, args map[string]interface{}, driver string, create bool) (driverArgs, error) {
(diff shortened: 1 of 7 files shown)


pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -115,30 +121,37 @@ func (h *handler) getArgsEnvAndStatus(meta metav1.Object, data data.Object, args
 			fmt.Sprintf("--driver=%s", driver),
 			fmt.Sprintf("--custom-install-script=/run/secrets/machine/value"))
 
-		rancherCluster, err := h.rancherClusterCache.Get(meta.GetNamespace(), meta.GetLabels()[capi.ClusterLabelName])
+		rancherCluster, err := h.rancherClusterCache.Get(infraObj.meta.GetNamespace(), infraObj.meta.GetLabels()[capi.ClusterLabelName])
 		if err != nil {
 			return driverArgs{}, err
 		}
 		cmd = append(cmd, toArgs(driver, args, rancherCluster.Status.ClusterName)...)
+		jobBackoffLimit = 1
```
snasovich, line 129: From what I see we've used backoff limit of 0 in create situation and now going with 1. Is this intended?

pkg/controllers/provisioningv2/rke2/machineprovision/template.go
```diff
@@ -47,8 +47,22 @@ func (h *handler) objects(ready bool, typeMeta metav1.Type, meta metav1.Object,
 		return []runtime.Object{secret}
 	}
 
-	if args.BootstrapOptional && args.BootstrapSecretName == "" {
-		args.BootstrapSecretName = "not-found"
```
snasovich, line 51: Are there any possible adverse effects stemming from not setting this any more?

tests/integration/pkg/cluster/clusters.go
```diff
@@ -165,6 +168,108 @@ func WaitFor(clients *clients.Clients, c *provisioningv1api.Cluster) (_ *provisi
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, machine := range machines.Items {
+		gvk := schema.FromAPIVersionAndKind(machine.Spec.InfrastructureRef.APIVersion, machine.Spec.InfrastructureRef.Kind)
+		gvr := schema.GroupVersionResource{
+			Group:    gvk.Group,
+			Version:  gvk.Version,
+			Resource: strings.ToLower(gvk.Kind) + "s",
```
snasovich, line 229: Is this acceptable to just append `s` here? I'm assuming we want to get plural name here.

snasovich (commented):
Did not review tests deep enough yet. Have some initial questions/comments already so submitting the review.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
thedadams: No, this is a mistake. Will fix.

pkg/controllers/provisioningv2/rke2/machineprovision/template.go
thedadams: No. The bootstrap secret is used to pass the system-agent install script. It is not used in the deletion scenario.

tests/integration/pkg/cluster/clusters.go
thedadams: Since this is a test, this essentially will only be `podmachines`. I can change it to this explicitly if you think that is better.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
thedadams: Fixed.

tests/integration/pkg/cluster/clusters.go
snasovich: That's fine, thanks for the explanation.

snasovich (approved):

paynejacob (approved):

PR #35931, opened 2021-12-20
Make entrypoint more helpful, and don't immediately exit on error

Oats87:
Signed-off-by: Chris Kim <oats87g@gmail.com>

### DIFF
--- a/package/entrypoint.sh
+++ b/package/entrypoint.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+
 set -e
 
 if [ ! -e /run/secrets/kubernetes.io/serviceaccount ] && [ ! -e /dev/kmsg ]; then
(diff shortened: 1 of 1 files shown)


package/entrypoint.sh
```diff
@@ -12,10 +11,12 @@ if [ -e /var/lib/rancher/etcd ] && [ ! -e /var/lib/rancher/k3s/server/db/etcd ];
   echo -n 'default' > /var/lib/rancher/k3s/server/db/etcd/name
 fi
 if [ -e /var/lib/rancher/k3s/server/db/etcd ]; then
+  echo "INFO: Running k3s server --cluster-init --cluster-reset"
   k3s server --cluster-init --cluster-reset &> ./k3s-cluster-reset.log
   if [ $? -ne 0 ]; then
     echo "ERROR:" && cat ./k3s-cluster-reset.log
     rm -f /var/lib/rancher/k3s/server/db/reset-flag
+    exit $?
```
brandond, line 19: Unless your intent is to return the exit code of `rm -f` I think we need something like:
```suggestion
  k3s server --cluster-init --cluster-reset &> ./k3s-cluster-reset.log
  K3S_EXIT=$?
  if [ $K3S_EXIT -ne 0 ]; then
    echo "ERROR:" && cat ./k3s-cluster-reset.log
    rm -f /var/lib/rancher/k3s/server/db/reset-flag
    exit $K3S_EXIT
```

annablender: Yep agree with this. Thanks for taking a look @brandond

Oats87: true, I rushed this PR up without testing it. I’m not sure this is the best approach in the first place though (getting rid of the set -e) so I need to think about it some more

brandond (approved):

thedadams (approved):

Issue #35274, opened 2021-10-26
"Drain Before Delete" support for RKE2 machine pools


snasovich:
A new flag should be supported on machine pool definition to designate if nodes should be drained before they are deleted. TBD if the option should be added to `provisioning.cattle.io.clusters -> spec.machinePools[i]` object or something else.

Then, (de)provisioning flow should be updated to use the value of this flag to drain nodes before they are deleted if necessary.

jakefhyde:
### Root cause

N/A

### What was fixed, or what changes have occurred

Added backend support for "Drain Before Delete" in rke2 for parity with rke. Until a corresponding UI change, the default behavior will be in line with rke1, which was to have "Drain Before Delete" be false by default.

### Areas or cases that should be tested

Node pools that should not be drained will have the CAPI annotation `machine.cluster.x-k8s.io/exclude-node-draining: "true"`, and node pools that should be drained will have no such annotation. This can be set explicitly in the yaml spec by first configuring a cluster with the UI, then in yaml edit mode under the `RKEMachinePool` spec setting `DrainBeforeDelete: true`
 
### What areas could experience regressions?

N/A

### Are the repro steps accurate/minimal?

N/a

jakefhyde:
This should no longer be blocking https://github.com/rancher/dashboard/issues/4448

Auston-Ivison-Suse:
**Feature Testing**

**Setup For validation**
Rancher version: v2.6-head(7efbeab)
Installation: Helm HA

**Downstream Cluster**
Provisioned Through: RKE2 on EC2 Instances. 3 workers 3 etcd 2 control planes
k8s: v1.21.9

Usecase: Someone has an already provisioned rke2 cluster. They want to "Drain Before Delete" to their nodes. 

**Steps**
1. edit config of rke2 cluster (this cluster shouldn't have Drain Before Delete) 
2. Check mark the Drain Before Delete within the config of each node pull
3. press save Save (might have to refresh the page to get this to work) 

**Result**
The cluster takes 2-3 to bring back up the nodes. When the nodes do come back up the _cloud-controller-manager_ pod for the etcds are unhealthy. 

**Logs of Pod: **

```
I0131 17:20:15.293940 1 serving.go:354] Generated self-signed cert in-memory
W0131 17:20:15.530382 1 authentication.go:419] failed to read in-cluster kubeconfig for delegated authentication: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
[...]

PR #35957, opened 2021-12-22
Add rke2 support for drain before delete

jakefhyde:
## Description

CAPI (cluster API) drains nodes on deletion by default, although it is possible to force a node not to drain by adding the `ExcludeNodeDrainingAnnotation` annotation (e.g.  `machine.cluster.x-k8s.io/exclude-node-draining: true`). This PR adds that annotation to the `MachineTemplateSpec` in the `MachineDeployment`, which will apply to each machine created. This creates parity with RKE1 "Drain Before Delete", by allowing users the option to apply a drainage policy to each machine pool.

Related Issue: #35274


thedadams (changes requested):
Can you add a test in `tests/integration/pkg/tests/machineprovisioning/machine_test.go` for this functionality?

Please reach out if you have questions about the provisioning tests.

jakefhyde:
@thedadams Ok the tests should pass this time, I was using a function that had since been removed from the testing framework.

thedadams (approved):

kinarashah (approved):

Issue #35828, opened 2021-08-25
Actions that cannot be performed by a Cluster member are available on an RKE1 cluster in a rancher setup when RKE2 flag is disabled


sowmyav27:
**On 2.6-head commit id: `b43e4d9`**

Actions that cannot be performed by a Cluster member are available on an RKE1 cluster in a rancher setup when RKE2 flag is disabled
<img width="1126" alt="Screen Shot 2021-08-24 at 11 04 08 PM" src="https://user-images.githubusercontent.com/26032343/130735223-25c71212-db96-459a-ad96-26cf74862442.png">

snasovich:
It's possible that controller(s) that should be tied to V2 provisioning flag are controlled by RKE2 flag instead. We may need to switch which flag enables these controllers.

annablender:
I reproduced this on a local single node DO RKE1 cluster, with the RKE2 flag both enabled and disabled. Both times I saw the `Take Snapshot` and `Rotate Certificates` options for the standard cluster member, so it's safe to assume the RKE2 flag status is unrelated to this bug.

PR in with fix https://github.com/rancher/rancher/pull/35963.

annablender:
## Testing template

#### Root cause

A standard cluster member on an RKE1 cluster was seeing options reserved for a cluster admin because the logic in the Norman API on what actions to make permissible to a user was incorrect.
 
#### What was fixed, or what changes have occurred

- Updated https://github.com/rancher/rancher/blob/release/v2.6/pkg/api/norman/customization/cluster/actionhandler.go
- Updated logic in https://github.com/rancher/rancher/blob/1a361b790d3099e004732ad2f51936456a53cd3c/pkg/api/norman/customization/cluster/formatter.go#L38

#### Areas or cases that should be tested

1. Create a new `user`
2. Create an RKE1 downstream cluster with `user` as a standard cluster member
3. Logout and login as `user`
4. You should only have access to the downstream cluster. View your cluster and make sure your actions drop down for the cluster looks like this
 
![image](https://user-images.githubusercontent.com/12955547/151041790-08dadeb4-90f7-4cf8-b4e8-ecd71cf186ae.png)

#### What areas could experience regressions ?

Rancher UI actions list for an RKE1 cluster if logic is overwritten.
 
#### Are the repro steps accurate/minimal ?

Yes.

Auston-Ivison-Suse:
**Reproduction Setup**
- rancher version: v2.6.3
[...]

PR #35963, opened 2021-12-23
Fix rke1 cluster member action logic

annablender:
Address [GitHub 35828](https://github.com/rancher/rancher/issues/35828)

### DIFF
--- a/pkg/api/norman/customization/cluster/actionhandler.go
+++ b/pkg/api/norman/customization/cluster/actionhandler.go
@@ -39,29 +39,41 @@ type ActionHandler struct {
 	Auth                          requests.Authenticator
 }
 
+func canUpdateCluster(apiContext *types.APIContext) bool {
+	if apiContext == nil {
+		return false
+	}
+	cluster := map[string]interface{}{
+		"id": apiContext.ID,
+	}
+	return canUpdateClusterWithValues(apiContext, cluster)
+}
+
+func canUpdateClusterWithValues(apiContext *types.APIContext, values map[string]interface{}) bool {
+	return apiContext.AccessControl.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", apiContext, values, apiContext.Schema) == nil
+}
+
+func canBackupEtcd(apiContext *types.APIContext) bool {
+	if apiContext == nil {
+		return false
+	}
+	etcdBackupSchema := types.Schema{ID: mgmtclient.EtcdBackupType}
+	backupMap := map[string]interface{}{
+		"namespaceId": apiContext.ID,
+	}
+	return apiContext.AccessControl.CanDo(v3.EtcdBackupGroupVersionKind.Group, v3.EtcdBackupResource.Name, "create", apiContext, backupMap, &etcdBackupSchema) == nil
+}
+
+func canCreateClusterTemplate(sar v1.SubjectAccessReviewInterface, apiContext *types.APIContext) bool {
+	if apiContext == nil {
+		return false
+	}
+	callerID := apiContext.Request.Header.Get(gaccess.ImpersonateUserHeader)
+	canCreateTemplates, _ := CanCreateRKETemplate(callerID, sar)
+	return canCreateTemplates
+}
+
 func (a ActionHandler) ClusterActionHandler(actionName string, action *types.Action, apiContext *types.APIContext) error {
-	canUpdateCluster := func() bool {
-		cluster := map[string]interface{}{
-			"id": apiContext.ID,
-		}
-		return apiContext.AccessControl.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", apiContext, cluster, apiContext.Schema) == nil
-	}
-
-	canBackupEtcd := func() bool {
-		etcdBackupSchema := types.Schema{ID: mgmtclient.EtcdBackupType}
-		// pkg/rbac/access_control.go:55 canAccess checks for the object's ID or namespace. The ns for etcdbackup will be the clusterID
-		backupMap := map[string]interface{}{
-			"namespaceId": apiContext.ID,
-		}
-		return apiContext.AccessControl.CanDo(v3.EtcdBackupGroupVersionKind.Group, v3.EtcdBackupResource.Name, "create", apiContext, backupMap, &etcdBackupSchema) == nil
-	}
-
-	canCreateClusterTemplate := func() bool {
-		callerID := apiContext.Request.Header.Get(gaccess.ImpersonateUserHeader)
-		canCreateTemplates, _ := CanCreateRKETemplate(callerID, a.SubjectAccessReviewClient)
-		return canCreateTemplates
-	}
-
 	switch actionName {
 	case v32.ClusterActionGenerateKubeconfig:
 		return a.GenerateKubeconfigActionHandler(actionName, action, apiContext)
@@ -72,47 +84,47 @@ func (a ActionHandler) ClusterActionHandler(actionName string, action *types.Act
 	case v32.ClusterActionViewMonitoring:
 		return a.viewMonitoring(actionName, action, apiContext)
 	case v32.ClusterActionEditMonitoring:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not access")
 		}
 		return a.editMonitoring(actionName, action, apiContext)
 	case v32.ClusterActionEnableMonitoring:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not access")
 		}
 		return a.enableMonitoring(actionName, action, apiContext)
 	case v32.ClusterActionDisableMonitoring:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not access")
 		}
 		return a.disableMonitoring(actionName, action, apiContext)
 	case v32.ClusterActionBackupEtcd:
-		if !canBackupEtcd() {
+		if !canBackupEtcd(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not backup etcd")
 		}
 		return a.BackupEtcdHandler(actionName, action, apiContext)
 	case v32.ClusterActionRestoreFromEtcdBackup:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not restore etcd backup")
 		}
 		return a.RestoreFromEtcdBackupHandler(actionName, action, apiContext)
 	case v32.ClusterActionRotateCertificates:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not rotate certificates")
 		}
 		return a.RotateCertificates(actionName, action, apiContext)
 	case v32.ClusterActionRotateEncryptionKey:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not rotate encryption key")
 		}
 		return a.RotateEncryptionKey(actionName, action, apiContext)
 	case v32.ClusterActionRunSecurityScan:
 		return a.runCisScan(actionName, action, apiContext)
 	case v32.ClusterActionSaveAsTemplate:
-		if !canUpdateCluster() {
+		if !canUpdateCluster(apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not save the cluster as an RKETemplate")
 		}
-		if !canCreateClusterTemplate() {
+		if !canCreateClusterTemplate(a.SubjectAccessReviewClient, apiContext) {
 			return httperror.NewAPIError(httperror.PermissionDenied, "can not save the cluster as an RKETemplate")
 		}
 		return a.saveAsTemplate(actionName, action, apiContext)
--- a/pkg/api/norman/customization/cluster/formatter.go
+++ b/pkg/api/norman/customization/cluster/formatter.go
@@ -35,22 +35,6 @@ func NewFormatter(schemas *types.Schemas, managementContext *config.ScaledContex
 	return &clusterFormatter
 }
 
-func canUserUpdateCluster(request *types.APIContext, resource *types.RawResource) bool {
-	if request == nil || resource == nil {
-		return false
-	}
-	cluster := map[string]interface{}{
-		"id": resource.ID,
-	}
-	return request.AccessControl.CanDo(
-		v3.ClusterGroupVersionKind.Group,
-		v3.ClusterResource.Name,
-		"update",
-		request,
-		cluster,
-		resource.Schema) == nil
-}
-
 func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawResource) {
 	if convert.ToBool(resource.Values["internal"]) {
 		delete(resource.Links, "remove")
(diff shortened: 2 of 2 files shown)


pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +77,40 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+			resource.AddAction(request, v32.ClusterActionRotateCertificates)
+			if rotateEncryptionKeyEnabled(f.clusterLister, resource.ID) {
+				resource.AddAction(request, v32.ClusterActionRotateEncryptionKey)
+			}
+			if convert.ToBool(resource.Values["enableClusterMonitoring"]) {
+				resource.AddAction(request, v32.ClusterActionDisableMonitoring)
+				resource.AddAction(request, v32.ClusterActionEditMonitoring)
+			} else {
+				resource.AddAction(request, v32.ClusterActionEnableMonitoring)
+			}
```
thedadams, line 104: These fields can be enabled regardless of where the `rancherKubernetesEngineConfig` exists. However, in this change, these are moved inside an if-block so they will only be included for RKE1 clusters. These should be moved outside the if-block so they can be included regardless of whether `rancherKubernetesEngineConfig` field exists for the cluster.

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +77,40 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+			if _, ok := resource.Values["rancherKubernetesEngineConfig"]; ok {
```
thedadams, line 105: This code is already inside an if-block that ensures the `rancherKubernetesEngineConfig` field exists.

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +77,40 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+
 		resource.AddAction(request, v32.ClusterActionExportYaml)
-		resource.AddAction(request, v32.ClusterActionRotateCertificates)
-		if rotateEncryptionKeyEnabled(f.clusterLister, resource.ID) {
-			resource.AddAction(request, v32.ClusterActionRotateEncryptionKey)
-		}
-		if _, ok := values.GetValue(resource.Values, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig"); ok {
+
+		// If a user has the BackupEtcd role/privilege, add it
+		if canUserBackupEtcd(request) {
```
thedadams, line 86: Nice!

thedadams (changes requested):

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -51,6 +52,21 @@ func canUserUpdateCluster(request *types.APIContext, resource *types.RawResource
 		resource.Schema) == nil
 }
 
+func canUserBackupEtcd(apiContext *types.APIContext) bool {
+	etcdBackupSchema := types.Schema{ID: mgmtclient.EtcdBackupType}
+	// pkg/rbac/access_control.go:55 canAccess checks for the object's ID or namespace. The ns for etcdbackup will be the clusterID
```
snasovich, line 57: There is no line 55 in `pkg/rbac/access_control.go:55`. This looks like it's copy-pasted from: https://github.com/rancher/rancher/blob/d3f98065499849a81f14704d7550b24126b8cd5b/pkg/api/norman/customization/cluster/actionhandler.go#L52

This comment is no longer relevant and we definitely just want to have it removed here. Also, recommend removing it from the place I've referenced while we're at it.

In fact, we're actually in the same package, so we don't need to have this copy-pasted and should just reference this newly introduced `canUserBackupEtcd` function in `actionhandler.go` and remove definition of inline `canBackupEtcd` function from `ActionHandler.ClusterActionHandler`.

snasovich (changes requested):
Have couple more comments that need to be addressed in addition to what @thedadams suggested.

snasovich:
Also, please add a comment to the issue explaining how it's not related to RKE2 feature flag.

pkg/api/norman/customization/cluster/formatter.go
annablender: Thank you!

pkg/api/norman/customization/cluster/formatter.go
annablender: Proposal: to make the code cohesive with this change, I'd suggest moving both canUpdateCluster and canUserBackupEtcd into separate functions in ActionHandler.ClusterActionHandler. The one caveat there is that (and please correct if I'm wrong!) canUpdateCluster is called 6+ times in ActionHandler.ClusterActionHandler and it may cause less overhead as an inline func. I was thinking of having both be inline functions in ActionHandler.go and Formatter.go so the code will run faster. Thoughts?

snasovich: I don't see a need to make them have ActionHandler as a receiver as they don't need anything from the handler, so I think they should stay receiver-less.
As far as which file of `cluster` package they need to be defined in is up for debate. Maybe move them to `utils.go` as they will be used in both `actionhandler.go` and `formatter.go`?
Regarding inlining functions for performance gain, I guess the gains will be negligible (if any), so I would recommend favoring reduced copy/paste over potential trivial performance gains.

annablender: Ok, actually. Spoke with @thedadams and since closures aren't that much faster than go functions I'm going to pull canUpdateCluster, canBackupEtcd, and canCreateClusterTemplate into shared standalone functions. I'll remove the inline definitions.

snasovich: Found this - https://stackoverflow.com/questions/45937924/in-go-do-non-capturing-closures-harm-performance
Basically, TLDR is worrying about the cost of function call when we're not talking millions of invocations is probably splitting the hairs. Plus each case is likely different as compiler optimizations may work differently based on specific circumstances. Basically, compilers are smart so we don't need to worry about minutia such as this. :)

pkg/api/norman/customization/cluster/formatter.go
annablender: Thanks for catching this!

pkg/api/norman/customization/cluster/formatter.go
annablender: Sounds good! Let me know if my updates are ok

pkg/api/norman/customization/cluster/actionhandler.go
```diff
@@ -39,29 +38,37 @@ type ActionHandler struct {
...
+		"namespaceId": apiContext.ID,
 	}
+	return apiContext.AccessControl.CanDo(v3.EtcdBackupGroupVersionKind.Group, v3.EtcdBackupResource.Name, "create", apiContext, backupMap, &etcdBackupSchema) == nil
+}
 
-	canCreateClusterTemplate := func() bool {
-		callerID := apiContext.Request.Header.Get(gaccess.ImpersonateUserHeader)
-		canCreateTemplates, _ := CanCreateRKETemplate(callerID, a.SubjectAccessReviewClient)
-		return canCreateTemplates
+func canCreateClusterTemplate(a ActionHandler, apiContext *types.APIContext) bool {
```
thedadams, line 62: I think it would be better to make this a method on the `ActionHandler`:
```suggestion
func (a ActionHandler) canCreateClusterTemplate(apiContext *types.APIContext) bool {
```

pkg/api/norman/customization/cluster/actionhandler.go
```diff
@@ -2,7 +2,6 @@ package cluster
 
 import (
 	"fmt"
-
```
thedadams, line 5: nit: don't remove line.

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +45,43 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+
+		// If a user has the BackupEtcd role/privilege, add it
+		if canBackupEtcd(request) {
 			resource.AddAction(request, v32.ClusterActionBackupEtcd)
-			resource.AddAction(request, v32.ClusterActionRestoreFromEtcdBackup)
 		}
+
+		// If user has permissions to update the cluster
+		if err := request.AccessControl.CanDo(v3.ClusterGroupVersionKind.Group, v3.ClusterResource.Name, "update", request, resource.Values, request.Schema); err == nil {
+			if _, ok := values.GetValue(resource.Values, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig"); ok {
```
thedadams, line 70: Can the signature of `canUpdateCluster` be modified so that it can be used here too?

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +45,43 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+				resource.AddAction(request, v32.ClusterActionRestoreFromEtcdBackup)
+			}
+			resource.AddAction(request, v32.ClusterActionRotateCertificates)
+			if rotateEncryptionKeyEnabled(f.clusterLister, resource.ID) {
+				resource.AddAction(request, v32.ClusterActionRotateEncryptionKey)
+			}
+
+			if val, ok := values.GetValue(resource.Values, "clusterTemplateRevisionId"); ok && val == nil {
+				if err := request.AccessControl.CanDo(v3.ClusterTemplateGroupVersionKind.Group, v3.ClusterTemplateResource.Name, "create", request, resource.Values, request.Schema); err == nil {
```
thedadams, line 79: Can the signature of `canUpdateCluster` be modified so that it can be used here too?

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +45,43 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
 	if _, ok := resource.Values["rancherKubernetesEngineConfig"]; ok {
 		resource.AddAction(request, v32.ClusterActionExportYaml)
-		resource.AddAction(request, v32.ClusterActionRotateCertificates)
-		if rotateEncryptionKeyEnabled(f.clusterLister, resource.ID) {
-			resource.AddAction(request, v32.ClusterActionRotateEncryptionKey)
-		}
-		if _, ok := values.GetValue(resource.Values, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig"); ok {
+
+		// If a user has the BackupEtcd role/privilege, add it
+		if canBackupEtcd(request) {
```
thedadams, line 64: The `v32.ClusterActionBackupEtcd` was previously protected by if-statement below on line 70. This check should be here as well.

thedadams (changes requested):

pkg/api/norman/customization/cluster/actionhandler.go
annablender: Yep! Caught that yesterday when I received golint errors

pkg/api/norman/customization/cluster/formatter.go
annablender: Totally agree, yep

pkg/api/norman/customization/cluster/formatter.go
annablender: This is checking the permissions to see if the user can create a cluster template resource, and we don't have a standalone func for that.

pkg/api/norman/customization/cluster/formatter.go
annablender: @snasovich is good with standalone functions being in actionhandler.go. Ready for re-review

snasovich (approved):

snasovich:
@annablender , looks like backport bot doesn't work on PRs.
And that's fine - we don't need to backport this one to 2.5.x. At this time we're not backporting by default, only on request.

annablender:
Yep, that's the case. @superseb https://github.com/rancherlabs/rancher-it-infra/pull/253 to get management to enable the feature, but sounds good. This will be the only PR for this issue.

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +45,45 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
 	resource.Links["shell"] = shellLink
 	resource.AddAction(request, v32.ClusterActionGenerateKubeconfig)
 	resource.AddAction(request, v32.ClusterActionImportYaml)
+
+	// If user has permissions to update the cluster (regardless of RKE1 or not)
+	if canUpdateCluster(request) {
```
thedadams, line 50: The old implementation will pass `resource.Values` to the `CanDo` function. This implementation passes `map[string]interface{}{"id": request.ID}`.

Are we sure this doesn't introduce potential issues?

thedadams (changes requested):

pkg/api/norman/customization/cluster/formatter.go
annablender: Fixed and retested

thedadams (dismissed):

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +45,45 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
+				resource.AddAction(request, v32.ClusterActionRestoreFromEtcdBackup)
+			}
+			resource.AddAction(request, v32.ClusterActionRotateCertificates)
+			if rotateEncryptionKeyEnabled(f.clusterLister, resource.ID) {
+				resource.AddAction(request, v32.ClusterActionRotateEncryptionKey)
+			}
+
+			if val, ok := values.GetValue(resource.Values, "clusterTemplateRevisionId"); ok && val == nil {
+				if canCreateClusterTemplate(f.SubjectAccessReviewClient, request) {
+					resource.AddAction(request, v32.ClusterActionSaveAsTemplate)
```
snasovich, line 82: Before the change whether `v32.ClusterActionSaveAsTemplate` action is added or not was determined based on  `request.AccessControl.CanDo(v3.ClusterTemplateGroupVersionKind.Group, v3.ClusterTemplateResource.Name, "create", request, resource.Values, request.Schema)` check and now we're calling `canCreateClusterTemplate` which in turn is based on calling `CanCreateRKETemplate` which implementation looks quite different from `CanDo`.

Could you please confirm these 2 approaches are equivalent or we're fixing some discrepancy here?

snasovich (commented):
Have one comment / question.

pkg/api/norman/customization/cluster/formatter.go
annablender: That's interesting. We use rbac in `formatter.go` but our international authenticator on the HTTP request header in `canCreateClusterTemplate.` Looking into it...

annablender: Ok, so looks like the first implementation in `formatter.go` uses a framework in `Steve` and the second one uses built-in Kubernetes. Going to do some debugging to make sure the roles and role bindings that are checked are the same.

paynejacob (dismissed):

pkg/api/norman/customization/cluster/formatter.go
annablender: Synced with @thedadams real quick about this and I think the best course of action is to just revert the call to `canCreateClusterTemplate` in `formatter.go` back to the RBAC call to avoid a regression. Let me know if this looks good

pkg/api/norman/customization/cluster/formatter.go
```diff
@@ -61,16 +45,45 @@ func (f *Formatter) Formatter(request *types.APIContext, resource *types.RawReso
...
-		resource.AddAction(request, v32.ClusterActionRotateCertificates)
-		if rotateEncryptionKeyEnabled(f.clusterLister, resource.ID) {
-			resource.AddAction(request, v32.ClusterActionRotateEncryptionKey)
-		}
+
+		// If a user has the backupetcd role/privilege, add it
 		if _, ok := values.GetValue(resource.Values, "rancherKubernetesEngineConfig", "services", "etcd", "backupConfig"); ok {
-			resource.AddAction(request, v32.ClusterActionBackupEtcd)
-			resource.AddAction(request, v32.ClusterActionRestoreFromEtcdBackup)
+			if canBackupEtcd(request) {
```
thedadams, line 65: nit: this line and the line above can be combined as:
```
ok && canBackupEtcd(request) {
```

thedadams (approved):

thedadams (approved):

paynejacob (approved):

Issue #35489, opened 2021-11-12
CAPI (Cluster API) update to version 1.0+


snasovich:
CAPI version 1.0.0 is available since 10/6/21 (https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.0.0) (in fact there is now 1.0.2 released - https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.0.2).

We should look into updating V2 cluster provisioning framework to be based on this GA version of CAPI.

snasovich:
Want to do it early in the release cycle so there is more time to find any issues potentially caused by this change.
As the change itself, ensure we're not using `v1alpha4` CAPI CRDs and use `v1beta1` everywhere.

thedadams:
These changes have no all been merged.

There should be nothing specific for QA to test here. The myriad of testing that will happen as we approach release will ensure that everything is working properly.

PR #35964, opened 2021-12-24
Bump cluster-api to v1.0.2

thedadams:
In addition to bumping the CAPI dependency, this PR adds conversion webhook
configuration to CAPI CRDs on startup 

Since the CAPI CRD versions have been bumped to v1beta1, we need to provide a
way for the old objects with version v1alpha4 to be upgraded to the new version.
The convert webhook is already running in rancher-webhook, so the conversion
configuration is added to the CRDs on startup.

This conversion configuration cannot be added to the CRD ahead of time because
the CABundle is not known. We rely on the `needacert` controller to recognize
that the conversion configuration needs the CABundle, and it will be added when
the CRD is updated. If the conversion webhook configuration was added
beforehand, then Rancher startup would hang as it tries to watch the CAPI
objects. It would not be able to watch the CAPI objects because doing so would
call the conversion webhook which would fail because of an invalid cert (we
don't know the certificate information). Therefore, this migration is necessary.

NOTE: This PR will not be complete until the [rancher-webhook](https://github.com/rancher/webhook/pull/64) PR is merged, a new RC tag created there, and the rancher-webhook version bumped here.

Issue:
https://github.com/rancher/rancher/issues/35489


pkg/api/norman/customization/monitor/prometheus_manager.go
```diff
@@ -126,7 +126,7 @@ func (q *Queries) Do(querys []*PrometheusQuery) (map[string][]*TimeSeries, error
 }
 
 func (q *Queries) GetLabelValues(labelName string) ([]string, error) {
-	value, _, err := q.api.LabelValues(q.ctx, labelName, []string{}, time.Now().Add(-time.Hour), time.Now())
+	value, _, err := q.api.LabelValues(q.ctx, labelName, nil, time.Now().Add(-time.Hour), time.Now())
```
snasovich, line 129: This change looks irrelevant to the scope of the PR to me.

pkg/rancher/migrations.go
```diff
@@ -228,3 +230,38 @@ func applyProjectConditionForNamespaceAssignment(label string, condition conditi
...
+		}
+
+		crd.Spec.Conversion = &apiextv1.CustomResourceConversion{
+			Strategy: apiextv1.WebhookConverter,
+			Webhook: &apiextv1.WebhookConversion{
+				ClientConfig: &apiextv1.WebhookClientConfig{
+					Service: &apiextv1.ServiceReference{
+						Namespace: "cattle-system",
+						Name:      "webhook-service",
+						Path:      &[]string{"/convert"}[0],
```
snasovich, line 253: This is both ugly and neat at the same time :)

snasovich (changes requested):
One general comment and one question. Otherwise LGTM.
However, the build is failing due to go modules verification issue. It doesn't look to me that it's caused by pending webhook bump, so please take a look.

pkg/api/norman/customization/monitor/prometheus_manager.go
thedadams: Ah yes. This is leftover from my original switch, and should not be changed. Will update.

pkg/provisioningv2/capi/capi.go
```diff
@@ -13,10 +13,15 @@ import (
...
 	clusterv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	clusterv1alpha4 "sigs.k8s.io/cluster-api/api/v1alpha4"
+	clusterv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/controllers"
 	"sigs.k8s.io/cluster-api/controllers/remote"
 	addonsv1alpha3 "sigs.k8s.io/cluster-api/exp/addons/api/v1alpha3"
-	expv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
+	addonsv1alpha4 "sigs.k8s.io/cluster-api/exp/addons/api/v1alpha4"
+	addonsv1beta1 "sigs.k8s.io/cluster-api/exp/addons/api/v1beta1"
+	expvv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
```
jakefhyde, line 22: Should this be `expv1alpha3`? Same question for `alpha4` and `beta1`.

jakefhyde, line 22: Should this be `expv1alpha3`? Same question for `alpha4` and `beta1`.

jakefhyde (changes requested):

pkg/provisioningv2/capi/capi.go
thedadams: Fixed.

pkg/api/norman/customization/monitor/prometheus_manager.go
thedadams: Fixed.

jakefhyde (approved):

pkg/provisioningv2/capi/capi.go
```diff
@@ -26,8 +31,13 @@ func init() {
 	_ = clientgoscheme.AddToScheme(schemes.All)
 	_ = clusterv1alpha3.AddToScheme(schemes.All)
 	_ = clusterv1alpha4.AddToScheme(schemes.All)
+	_ = clusterv1beta1.AddToScheme(schemes.All)
 	_ = expv1alpha3.AddToScheme(schemes.All)
+	_ = expv1alpha4.AddToScheme(schemes.All)
```
kinarashah, line 36: How do we identify we need exp v1alpha4? I looked at `capi-webhooks.yaml` and all entries are using v1beta1.

thedadams: I am not entirely sure how we would verify which schemes we need, especially with all the upgrade scenarios.

kinarashah: Oh okay, so we need to add all of them when bumping capi 👍🏼

kinarashah (dismissed):

thedadams:
Rancher webhook is now bumped here.

package/Dockerfile
```diff
@@ -46,7 +46,7 @@ ENV CATTLE_SYSTEM_AGENT_UPGRADE_IMAGE rancher/system-agent:${CATTLE_SYSTEM_AGENT
 
 # System charts minimal version
 ENV CATTLE_FLEET_MIN_VERSION=100.0.2+up0.3.8
-ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=1.0.2+up0.2.2
+ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=1.0.3+up0.2.4-rc2
```
snasovich, line 49: Doesn't prerelease go before metadata usually? 
https://regex101.com/r/1vI80U/1 - using regex from https://semver.org/

snasovich (commented):
Have one question, otherwise LGTM

package/Dockerfile
thedadams: This is the way rancher-charts works.

thedadams: We are releasing a new version of the chart (1.0.3) and the image contained in this version is 0.2.4-rc2. Once the chart is released as part of rancher/charts release, then the version of the image will be 0.2.4

jiaqiluo (approved):
LGTM 💯

Issue #32803, opened 2021-05-20
Upgrade Plan for RKE2 downstream cluster does not honor system-default-registry / does not work in airgapped scenarios


Martin-Weiss:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

Rancher 2.5.8 or earlier
Imported RKE2 downstream cluster 1.20.5
Air-Gapped deployments

Change the version of the downstream cluster via edit to 1.20.6

The plans created have the prepare and the upgrade image `rancher/rke2-upgade:v.1.20.6-rke2r1` and NOT `<system-default-registry>/rancher/rke2-upgade:v.1.20.6-rke2r1`. This is also not configurable.
Even changing the deployed plan via "patch" does not help as Rancher is overwriting the plans automatically with the wrong settings.

**Result:**

Upgrade Pods fail in pulling the image as they can not reach docker.io

**Other details that may be helpful:**

Registry mirror feature in rke2 can not be used in 1.20.5 or earlier or in 1.19.. as there is only a single on-premise registry where images are in namespaces and the namspace mapping feature in registries.yaml only comes with 1.20.6.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.5.8
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): imported
- Machine type (cloud/VM/metal) and specifications (CPU/memory): VM
- Kubernetes version (use `kubectl version`): 1.20.5

Martin-Weiss:
Some further comments:

There are several changes required in the rke2-server or agent configuration during upgrades.. i.e. from 1.20.5 to 1.20.6 changes for the system-default-registry setting in config.yaml and registries.yaml are required due to changes in the behavior. Also for changes like vsphere in tree to out of tree migration or dependencies on the OS side of things we must have a way to modify / customize the PLANs that are created and used by Rancher for the given downstream cluster.

Is there / What is the way to customize the upgrade plans?

Martin-Weiss:
One more comment:

In airgapped setup we also might have to delete the "old" /var/lib/rancher/rke2/agent/images/rke2-images.linux-amd64.tar.zst during the upgrade - replacing the binary i.e. 1.20.5 -> 1.20.6 and still importing the images from 1.20.5 might not be the best thing...

thedadams:
[...]

PR #35965, opened 2021-12-24
fix the issue that global private registry is not used when upgrading imported RKE2 clusters 

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/32803
Problem:
Images are not pulled from Rancher's global private registry when upgrading the k8s version of RKE2 clusters via Rancher UI, which will fail the upgrade because the cluster's nodes cannot connect to docker.io in the air-gapped environment.

Fixes:
Append the global private registry to the image used in the upgrade plan for RKE2 clusters;
Add 'global.cattle.systemDefaultRegistry:reg' to the default values passed to the app `rancher-rke2-upgrade`

### DIFF
--- a/pkg/controllers/management/k3sbasedupgrade/deployPlans.go
+++ b/pkg/controllers/management/k3sbasedupgrade/deployPlans.go
@@ -11,6 +11,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/management/clusterdeploy"
 	planClientset "github.com/rancher/rancher/pkg/generated/clientset/versioned/typed/upgrade.cattle.io/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/settings"
 	planv1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuserlegacy/helm/common/extra_args_injection.go
```diff
@@ -35,7 +35,7 @@ func injectDefaultRegistry(obj *v3.App) map[string]string {
 		return nil
 	}
 
-	return map[string]string{"systemDefaultRegistry": reg}
+	return map[string]string{"systemDefaultRegistry": reg, "cattle.systemDefaultRegistry": reg}
```
jiaqiluo, line 38: Note: the value will be ignored by the app which values.yaml does not define the field `global.cattle.systemDefaultRegistry`

paynejacob (approved):

pkg/controllers/management/k3sbasedupgrade/deployPlans.go
```diff
@@ -30,7 +31,7 @@ func (h *handler) deployPlans(cluster *v3.Cluster, isK3s, isRke2 bool) error {
 	)
 	switch {
 	case isRke2:
-		upgradeImage = rke2upgradeImage
+		upgradeImage = settings.PrefixPrivateRegistry(rke2upgradeImage)
```
thedadams, line 34: Can we do the same thing for `k3supgradeImage` on Line 40?

jiaqiluo: @thedadams  
I thought about that, too 😄  
I did not add it because 1) it is out of the scope of the linked issue, and 2) it will change the current behavior for importing k3s clusters.  

How about this: I will make a new issue and a PR for the imported k3s cluster for easier tracking and QA?  
Can we get this PR merge fisrt?

paynejacob (approved):

thedadams (approved):

Issue #32284, opened 2021-04-22
In the import cluster, the node should not be deleted when the node crd is deleted


kingsd041:
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
1. Use rke to launch a K8s cluster (2 nodes)
```
~  kubectl get nodes
NAME            STATUS   ROLES                      AGE     VERSION
192.168.64.77   Ready    controlplane,etcd,worker   3m59s   v1.20.5
192.168.64.78   Ready    worker                     3m59s   v1.20.5
```
2. Import the cluster into rancher
3. At this time, you can query the corresponding node crd in rancher
```
root@49130430bb40:/var/lib/rancher# kubectl get nodes.management.cattle.io  -A
NAMESPACE   NAME            AGE
c-qscft     machine-2f5fp   45s
c-qscft     machine-t5lhc   44s
local       machine-c7mp5   15d
```
4. Delete any node crd
```
root@49130430bb40:/var/lib/rancher# kubectl delete nodes.management.cattle.io machine-t5lhc -n c-qscft
node.management.cattle.io "machine-t5lhc" deleted
```
**Result:**
After deleting the node crd in the rancher container, the node imported into the cluster is also deleted. Currently, there is only one node left in the cluster.
```
~ kubectl get nodes
NAME            STATUS   ROLES                      AGE    VERSION
192.168.64.77   Ready    controlplane,etcd,worker   4m5s   v1.20.5
```

In the import cluster, the node should not be deleted when the node crd is deleted

**Other details that may be helpful:**

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): rancher/rancher:v2.5.7
- Installation option (single install/HA): single

thxCode:
https://github.com/rancher/rancher/blob/39d572aff5ccef43ff0bb9432f9e8f9b9b29d234/pkg/systemtemplate/template.go#L91-L107 `cattle-admin` ClusterRole seems to open, perhaps r/r can provide another ClusterRole to maintain the imported cluster, and give an approach to narrow it.

PR #35991, opened 2021-12-30
[2.5] skip remove nodes for non-custom cluster

JacieChao:
We should skip deleting k8s cluster node for non-custom cluster. We add a formatter function to remove delete node actions for non-custom cluster [here](https://github.com/rancher/rancher/blob/release/v2.5/pkg/api/norman/customization/node/node.go#L58) but not skip it in node controller. 

https://github.com/rancher/rancher/issues/32284


ibuildthecloud (approved):

JacieChao:
wait for fix

Issue #36000, opened 2022-01-03
Upgrade Plan for K3s downstream cluster does not honor system-default-registry


jiaqiluo:
When an upgrade of the k8s version in the imported k3s cluster is triggered, the Plan does not use the system-default-registry as the source for pulling images. As a result, the upgrade job and pod fail in pulling images as they cannot reach docker.io in the air-gapped environment. 

The same issue is reported for RKE2 https://github.com/rancher/rancher/issues/32803

jiaqiluo:
### Root cause
When upgrading the k8s version of K3s clusters via Rancher UI, Rancher creates a Plan (which will create a Job later) and installs an app named `rancher-k3s-upgrade` in the K3s cluster (which can be found in System project -> Apps in the old UI), but Rancher does not append the global-system-registry to the images. As a result, the Job and the app try to pull images from DockerHub which fails in the air-gapped environment. 

### What was fixed, or what changes have occurred 
- the system-global-registry is appended to the image when setting the k3supgradeImage in the Plan 

### Areas or cases that should be tested
Upgrading the k8s version of a downstream K3s cluster should succeed in the air-gapped environment. Images should be pulled from the global-system-registry
 
### What areas could experience regressions?
It should not affect anything except upgrading the downstream K3s cluster. 

### Are the repro steps accurate/minimal?
yes.

izaac:
Testing blocked by https://github.com/rancher/rancher/issues/36937

izaac:
Rancher version `v2.6.4-rc13`

Airgap setup with private registry.
k3s cluster configured to use the private registry. Follow the private registry method [here](https://rancher.com/docs/k3s/latest/en/installation/airgap/).

```yaml
mirrors:
  docker.io:
    endpoint:
      - "https://<fqdn>"
configs:
  "<fqdn>":
    auth:
      username: testuser
      password: <REDACTED>
    tls:
      cert_file: /etc/rancher/k3s/cert.pem
      key_file: /etc/rancher/k3s/key.pem
      ca_file: /etc/rancher/k3s/ca.pem
```

Use an older k3s binary like [k8s 1.21](https://github.com/k3s-io/k3s/releases/tag/v1.21.10%2Bk3s1) to upgrade to [k8s 1.22](https://github.com/k3s-io/k3s/releases/tag/v1.22.7%2Bk3s1).

Download the `k3s-images.txt` for each version.
Make sure the registry has the necessary images

[...]

PR #36003, opened 2022-01-03
append the global private registry to the image used in the upgrade plan for K3s clusters

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/36000

Problem:
When an upgrade of the k8s version in the imported k3s cluster is triggered, the Plan does not use the system-default-registry as the source for pulling images. As a result, the upgrade job and pod fail in pulling images as they cannot reach docker.io in the air-gapped environment.

Fixes:
Append the global private registry to the image used in the upgrade plan for RKE2 clusters

### DIFF
--- a/pkg/controllers/management/k3sbasedupgrade/deployPlans.go
+++ b/pkg/controllers/management/k3sbasedupgrade/deployPlans.go
@@ -37,7 +37,7 @@ func (h *handler) deployPlans(cluster *v3.Cluster, isK3s, isRke2 bool) error {
 		Version = cluster.Spec.Rke2Config.Version
 		strategy = cluster.Spec.Rke2Config.ClusterUpgradeStrategy
 	case isK3s:
-		upgradeImage = k3supgradeImage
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/k3sbasedupgrade/deployPlans.go
```diff
@@ -37,7 +37,7 @@ func (h *handler) deployPlans(cluster *v3.Cluster, isK3s, isRke2 bool) error {
 		Version = cluster.Spec.Rke2Config.Version
 		strategy = cluster.Spec.Rke2Config.ClusterUpgradeStrategy
 	case isK3s:
-		upgradeImage = k3supgradeImage
+		upgradeImage = settings.PrefixPrivateRegistry(k3supgradeImage)
```
jiaqiluo, line 40: Note: 
This is exactly what we did for RKE2 at line 34

thedadams (approved):

paynejacob (approved):

PR #36005, opened 2022-01-03
Extensions for core and storage k8s functionality.

igomez06:
Core and storage k8s helpers
**Framework:**
Implemented new client functionality to be able to get a downstream k8s client.
Added helpers for volumes (persistent volumes and persistent volume claims), storage classes, secrets, and config maps

### DIFF
--- a/tests/framework/extensions/configmaps/configmaps.go
+++ b/tests/framework/extensions/configmaps/configmaps.go
@@ -1,0 +1,53 @@
+package configmaps
+
+import (
+	"context"
+
+	"github.com/rancher/rancher/pkg/api/scheme"
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/rancher/rancher/tests/framework/extensions/unstructured"
+	coreV1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var ConfigMapGroupVersionResource = schema.GroupVersionResource{
+	Group:    "",
+	Version:  "v1",
+	Resource: "configmaps",
+}
+
+// CreateConfigMap is a helper function that uses the dynamic client to create a config map on a namespace for a specific cluster.
+// It registers a delete fuction.
+func CreateConfigMap(client *rancher.Client, clusterName, configMapName, description, namespace string, data, labels, annotations map[string]string) (*coreV1.ConfigMap, error) {
+	// ConfigMap object for a namespace in a cluster
+	annotations["field.cattle.io/description"] = description
+	configMap := &coreV1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        configMapName,
+			Annotations: annotations,
+			Namespace:   namespace,
+			Labels:      labels,
+		},
+		Data: data,
+	}
+
+	dynamicClient, err := client.GetDownStreamClusterClient(clusterName)
+	if err != nil {
+		return nil, err
+	}
+
+	configMapResource := dynamicClient.Resource(ConfigMapGroupVersionResource).Namespace(namespace)
+
+	unstructuredResp, err := configMapResource.Create(context.TODO(), unstructured.MustToUnstructured(configMap), metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	newConfig := &coreV1.ConfigMap{}
+	err = scheme.Scheme.Convert(unstructuredResp, newConfig, unstructuredResp.GroupVersionKind())
+	if err != nil {
+		return nil, err
+	}
+	return newConfig, nil
+}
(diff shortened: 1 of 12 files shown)


tests/framework/extensions/secrets/basicAuth.go
```diff
@@ -0,0 +1,24 @@
+package secrets
+
+import (
+	coreV1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewBasicAuthSecret(name, namespace, username, password string) *coreV1.Secret {
+	return &coreV1.Secret{
+		TypeMeta: metav1.TypeMeta{
```
paynejacob, line 10: This is redundant, the k8s client will derive this from the struct type.

tests/framework/extensions/configmaps/configmaps.go
```diff
@@ -0,0 +1,66 @@
...
+	Resource: "configmaps",
+}
+
+// CreateConfigMap is a helper function that uses the dynamic client to create a config map on a namespace for a specific cluster.
+// It registers a delete fuction.
+func CreateConfigMap(client *rancher.Client, clusterName, configMapName, description, namespace string, data, labels, annotations map[string]string) (*coreV1.ConfigMap, error) {
+	// ConfigMap object for a namespace in a cluster
+	annotations["field.cattle.io/description"] = description
+	configMap := &coreV1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
```
paynejacob, line 27: This is redundant, the k8s client will derive this from the struct type.

paynejacob (commented):
The helpers for the corev1 resources seem abstract is there a specific use case these are being created for?

tests/framework/extensions/configmaps/configmaps.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	configMapResource := dynamicClient.Resource(ConfigMapGroupVersionResource).Namespace(namespace)
+
+	unstructuredResp, err := configMapResource.Create(context.TODO(), unstructured.MustToUnstructured(configMap), metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	client.Session.RegisterCleanupFunc(func() error {
```
paynejacob, line 52: Doesn't the dynamic client already do this?

tests/framework/extensions/secrets/secrets.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	secretResource := dynamicClient.Resource(SecretGroupVersionResource).Namespace(namespace)
+
+	unstructuredResp, err := secretResource.Create(context.TODO(), unstructured.MustToUnstructured(secret), metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	client.Session.RegisterCleanupFunc(func() error {
```
paynejacob, line 36: Doesn't the dynamic client already do this?

tests/framework/extensions/storageclasses/awsebs/create.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	storageClassVolumesResource := dynamicClient.Resource(storageclasses.StorageClassGroupVersionResource).Namespace("")
+
+	unstructuredResp, err := storageClassVolumesResource.Create(context.TODO(), unstructuredStorageClass, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	client.Session.RegisterCleanupFunc(func() error {
```
paynejacob, line 50: Doesn't the dynamic client already do this?

tests/framework/extensions/configmaps/configmaps.go
igomez06: Yes this is true, this not needed

tests/framework/extensions/secrets/secrets.go
igomez06: Yes this is true, this not needed

tests/framework/extensions/storageclasses/awsebs/create.go
igomez06: Yes this is true, this not needed

igomez06:
> The helpers for the corev1 resources seem abstract is there a specific use case these are being created for?

Well just to make it easier on the developer when creating these resources, so then they can just make a single create call for these. Instead of having to convert the structs to unstructures every time when wanted create a config map for example. I mean I think having the constructors for some of these brings some value. But I'm open to reimplement.

tests/framework/extensions/secrets/basicAuth.go
igomez06: I don't understand what you mean. There are different types of secrets, my thought process was make that easier to create for the developer.

tests/framework/extensions/configmaps/configmaps.go
igomez06: Do you mean doing the TypeMeta is redundant?

paynejacob (approved):

PR #36032, opened 2022-01-05
Update health check command as per etcdctl v3 API

rishabhmsra:
**Adding this fix for rke_support_matrix job**

**Issue:**
- The etcdctl health check command `ETCDCTL_API=2 etcdctl --endpoints "https://127.0.0.1:2379"  --ca-file /etc/kubernetes/ssl/kube-ca.pem --cert-file  $ETCDCTL_CERT --key-file  $ETCDCTL_KEY cluster-health` fails for k8s 1.22 clusters. Because of which rke_support_matrix jenkins job also fails.
Hence updated the command (to use etcdctl v3) for k8s 1.22 clusters : `ETCDCTL_API=3 etcdctl endpoint health --cluster` and based on it's output changed the assert conditions as well.

- Verified the changes locally using these [steps](https://github.com/rancher/rancher/tree/release/v2.6/tests/validation/tests/rke) and the test case `pytest -k "test_install_config_1"` got passed. **Used RKE v1.3.4-rc1**
- **Note:** For k8s version 1.21 and less, `ETCDCTL_API=2` command will be executed and for higher version `ETCDCTL_API=3` would be executed.

### DIFF
--- a/tests/validation/tests/rke/common.py
+++ b/tests/validation/tests/rke/common.py
@@ -1,9 +1,14 @@
 import time
 import os
 import json
+from packaging import version
 
 k8s_resource_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                "resources/k8s_ymls/")
+k8s_rancher_version = str(os.environ.get("RANCHER_K8S_VERSION").split("-")[0][1:])
+k8s_rancher_version = version.parse(k8s_rancher_version)
+k8s_rancher_version = version.parse(f"{str(k8s_rancher_version.major)}.{str(k8s_rancher_version.minor)}")
+k8s_fixed_version = version.parse("1.21")
 
 # Global expectedimagesdict declared to store the images for a specific
 # k8s Version
@@ -133,11 +138,17 @@ def wait_for_etcd_cluster_health(node, etcd_private_ip=False):
     endpoints = "127.0.0.1"
     if etcd_private_ip:
         endpoints = node.private_ip_address
-    etcd_tls_cmd = (
-        'ETCDCTL_API=2 etcdctl --endpoints "https://'+endpoints+':2379" '
-        ' --ca-file /etc/kubernetes/ssl/kube-ca.pem --cert-file '
-        ' $ETCDCTL_CERT --key-file '
-        ' $ETCDCTL_KEY cluster-health')
+    if k8s_rancher_version <= k8s_fixed_version:
+        etcd_tls_cmd = (
+                'ETCDCTL_API=2 etcdctl --endpoints "https://' + endpoints + ':2379" '
+                 ' --ca-file /etc/kubernetes/ssl/kube-ca.pem --cert-file '
+                 ' $ETCDCTL_CERT --key-file '
+                 ' $ETCDCTL_KEY cluster-health'
+        )
+    else:
+        etcd_tls_cmd = (
+            'ETCDCTL_API=3 etcdctl endpoint health --cluster'
+        )
 
     print(etcd_tls_cmd)
     start_time = time.time()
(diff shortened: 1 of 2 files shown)


tests/validation/tests/rke/common.py
```diff
@@ -133,20 +134,30 @@ def wait_for_etcd_cluster_health(node, etcd_private_ip=False):
     endpoints = "127.0.0.1"
     if etcd_private_ip:
         endpoints = node.private_ip_address
-    etcd_tls_cmd = (
-        'ETCDCTL_API=2 etcdctl --endpoints "https://'+endpoints+':2379" '
-        ' --ca-file /etc/kubernetes/ssl/kube-ca.pem --cert-file '
-        ' $ETCDCTL_CERT --key-file '
-        ' $ETCDCTL_KEY cluster-health')
+    if k8s_rancher_version <= "1.21.8":
```
izaac, line 137: is 1.21.8 the break point of the API change ? Or the reason of hardcoding the version here is due to it being the latest of the 1.21.x version line?

I would set this version in a variable next to the `k8s_rancher_version` instead of here.

Also comparing version strings is a bit risky could we use the `packaging.version.parse`  I think there are a few examples in other tests.

tests/validation/tests/rke/common.py
```diff
@@ -133,20 +134,30 @@ def wait_for_etcd_cluster_health(node, etcd_private_ip=False):
...
 
     print(etcd_tls_cmd)
     start_time = time.time()
     while start_time - time.time() < 120:
         result = node.docker_exec('etcd', "sh -c '" + etcd_tls_cmd + "'")
         print("**RESULT**")
         print(result)
-        if 'cluster is healthy' in result:
-            break
+        if k8s_rancher_version <= "1.21.8":
```
izaac, line 155: Same about the version

tests/validation/tests/rke/common.py
```diff
@@ -356,11 +367,16 @@ def validation_node_roles(nodes, k8s_nodes, etcd_private_ip=False):
                             "Expected to find taint for etcd-only node"
                 # check etcd membership and cluster health
                 result = wait_for_etcd_cluster_health(node, etcd_private_ip)
-                for member in etcd_members:
-                    expect = "got healthy result from https://{}".format(
-                        member)
-                    assert expect in result, result
-                assert 'cluster is healthy' in result, result
+                if k8s_rancher_version <= "1.21.8":
```
izaac, line 370: Same about the version

izaac (changes requested):
Looks good overall, just need a few changes.

rishabhmsra:
Hi @izaac, by mistake Vivek has requested for a review please ignore it. I'll let you know once the changes are done.

tests/validation/tests/rke/common.py
```diff
@@ -4,6 +4,7 @@
 
 k8s_resource_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                "resources/k8s_ymls/")
+k8s_rancher_version = str(os.environ.get("RANCHER_K8S_VERSION").split("-")[0][1:])
```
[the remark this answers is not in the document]
rishabhmsra: Izaac, as per our discussion if this is only for RKE1, would this be still an issue?

tests/validation/tests/rke/common.py
rishabhmsra: Yeah 1.21.8 would not be the correct version for comparison. As 1.22 k8s version is the break point of API change, i'm changing this version from 1.21.x to 1.21.

tests/validation/tests/rke/common.py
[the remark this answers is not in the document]
izaac: This is RKE1 only as Sowmya confirmed. Resolving

izaac (approved):
LGTM

I also reviewed the Jenkins runs results from different k8s versions. 1.22 had console log output coming form the fix logic.

izaac:
Adding @sgapanovich as reviewer

sowmyav27 (approved):
LGTM

sgapanovich:
LGTM as well

sgapanovich (approved):

Issue #35868, opened 2021-12-14
Standard user gets error `Method PUT not supported` when creating an rke2 cluster through  `edit as yaml` 


slickwarren:
**Setup**
- Rancher version: v2.6.3-rc5
- Browser type & version:chrome

**Describe the bug**

when standard user creates a cluster and edits as yaml, the user then has error `Method PUT not supported` and is unable to create the cluster, even if going back to `edit as form`

**To Reproduce**

* login as standard user
* go to `cluster management` -> `create` -> select any rke2 cluster provider -> edit as yaml -> click `create`

**Result**
user is shown `Method PUT not supported` through the UI and is unable to create the cluster

**Expected Result**

standard users should be able to create via `edit as yaml` for any resource that they can create via form through the UI.

**Additional context**
* user is will encounter the same issue if they switch back to `edit as form` and click `create`. So the only way to fix this is to never enter the `edit as yaml` flow. 
* this works fine for an admin user

thedadams:
I have found the problem. The RBAC is setup so that a standard user only has access to update a machine config once a cluster is created with said machine config. When a user clicks to "Edit as YAML" the UI will create machine config without creating the cluster. Then, once the user clicks the "Create" button, the UI will update the machine config before creating the cluster.

It is this update call prior to creating the cluster that is failing in this case. Interestingly, if the standard user successfully provisions a cluster (without clicking the "Edit as YAML" button) and then goes through the above steps, the error will be different because the user will have update access to some machine configs, but not the one they are trying to update.

Ultimately the solution here is to create a controller in Rancher that watches these machine configs and creates roles/rolebindings similar to how nodetemplates are handled in the RKE1 case. Since this work will likely take longer than we have before the 2.6.3 release, then we should release note this and move the issue to the 2.6.4 release.

The release note should state that a standard user will not be able to create a cluster after clicking the "Edit as YAML" button. They will have to start over and create the cluster without clicking this button. The create will be successful otherwise.

slickwarren:
tested on v2.6.4-alpha3 as a standard user:
* provision via `edit via form` -- pass
* provision via `edit as form` switch to `edit via yaml` -- pass
* edit via `edit as form` -- pass
* edit via `edit as form` -> switch to `edit via yaml` -- pass
* edit via `edit as yaml` -- pass
new issue filed from this testing (unrelated to creating via yaml): https://github.com/rancher/rancher/issues/36358

PR #36041, opened 2022-01-05
Add roles and rolebindings when creating rke-machine-configs

thedadams:
Currently, if a standard user creates a rke-machine-config, they won't have
access to modify it until it is associated to a provisioning cluster that
the user owns. This creates issues in the UI when using the "Edit as YAML"
during cluster creation.

Now, roles and rolebindings will be added for rke-machine-configs when they are
created. This also requires changes to rancher-webhook (to add the creatorID).

Issue:
https://github.com/rancher/rancher/issues/35868

### DIFF
--- a/pkg/controllers/management/authprovisioningv2/authprovisioningv2.go
+++ b/pkg/controllers/management/authprovisioningv2/authprovisioningv2.go
@@ -8,8 +8,10 @@ import (
 	"github.com/rancher/lasso/pkg/dynamic"
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/features"
 	mgmtcontrollers "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	provisioningcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/rancher/rancher/pkg/wrangler"
 	"github.com/rancher/wrangler/pkg/apply"
 	apiextcontrollers "github.com/rancher/wrangler/pkg/generated/controllers/apiextensions.k8s.io/v1"
@@ -24,6 +26,7 @@ const (
 )
 
 type handler struct {
+	mgmtCtx                              *config.ManagementContext
 	roleLocker                           locker.Locker
 	roleCache                            rbacv1.RoleCache
 	roleController                       rbacv1.RoleController
@@ -45,7 +48,7 @@ type handler struct {
 	provisioningClusterGVK               schema.GroupVersionKind
 }
 
-func Register(ctx context.Context, clients *wrangler.Context) error {
+func Register(ctx context.Context, clients *wrangler.Context, management *config.ManagementContext) error {
 	clusterGVK, err := gvk.Get(&v1.Cluster{})
 	if err != nil {
 		// this is a build issue if it happens
@@ -53,6 +56,7 @@ func Register(ctx context.Context, clients *wrangler.Context) error {
 	}
 
 	h := &handler{
+		mgmtCtx:                              management,
 		roleCache:                            clients.RBAC.Role().Cache(),
 		roleController:                       clients.RBAC.Role(),
 		roleBindingController:                clients.RBAC.RoleBinding(),
@@ -86,6 +90,9 @@ func Register(ctx context.Context, clients *wrangler.Context) error {
 	clients.Mgmt.ProjectRoleTemplateBinding().OnChange(ctx, "auth-prov-v2-prtb", h.OnPRTB)
 	clients.Provisioning.Cluster().OnChange(ctx, "auth-prov-v2-cluster", h.OnCluster)
 	clients.CRD.CustomResourceDefinition().OnChange(ctx, "auth-prov-v2-crd", h.OnCRD)
+	if features.RKE2.Enabled() {
+		clients.Dynamic.OnChange(ctx, "auth-prov-v2-rke-machine-config", validMachineConfigGVK, h.OnMachineConfigChange)
+	}
 	clients.Provisioning.Cluster().Cache().AddIndexer(byClusterName, func(obj *v1.Cluster) ([]string, error) {
 		return []string{obj.Status.ClusterName}, nil
 	})
--- a/pkg/controllers/management/authprovisioningv2/machineconfigs.go
+++ b/pkg/controllers/management/authprovisioningv2/machineconfigs.go
@@ -1,0 +1,79 @@
+package authprovisioningv2
+
+import (
+	"strings"
+
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/controllers/management/nodetemplate"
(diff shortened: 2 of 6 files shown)


thedadams:
Note that once [this](https://github.com/rancher/webhook/pull/66) PR is merged, the rancher-webhook version will need to be bumped in this PR as well.

pkg/controllers/management/authprovisioningv2/machineconfigs.go
```diff
@@ -0,0 +1,79 @@
...
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/controllers/management/rbac"
+	"github.com/rancher/wrangler/pkg/data"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const ownerBindingsAnno = "ownerBindingsCreated"
```
jiaqiluo, line 16: nit:
The same constant is already declared at `pkg/controllers/management/nodetemplate/nodetemplate.go`, should we just use that?

pkg/controllers/management/authprovisioningv2/machineconfigs.go
```diff
@@ -0,0 +1,79 @@
...
+	annotations := objMeta.GetAnnotations()
+	if annotations != nil && annotations[ownerBindingsAnno] == "true" {
+		return obj, nil
+	}
+
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	creatorID, ok := annotations[rbac.CreatorIDAnn]
+	if !ok {
+		return obj, fmt.Errorf("machine config %s %s has no creatorId annotation", gvk, objMeta.GetName())
+	}
```
jiaqiluo, line 45: In the case of upgrading Rancher which has existing clusters, their `rke-machine-configs` do not have the `creatorId` annotation, right? If we return an error here, does it mean the `role` and `rolebinding` will never be created for those clusters?

pkg/controllers/management/rbac/rbac.go
```diff
@@ -234,7 +234,7 @@ func GetRoleNameAndVerbs(roleAccess string, resourceName string, resourceType st
 	case NodeTemplateResource:
 		resourceName += "-nt-"
 	default:
-		resourceName += "-" + resourceName + "-"
+		resourceName += "-" + resourceType + "-"
```
jiaqiluo, line 237: Good catch!

jiaqiluo (changes requested):

pkg/controllers/management/authprovisioningv2/machineconfigs.go
thedadams: Yes, we should use that.

pkg/controllers/management/authprovisioningv2/machineconfigs.go
thedadams: For now, I don't think an error can be returned here because of this upgrade scenario you laid out. So the controller will just exit in this case.

pkg/controllers/management/nodetemplate/nodetemplate.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	ownerBindingsAnno = "ownerBindingsCreated"
+	OwnerBindingsAnno = "ownerBindingsCreated"
```
jiaqiluo, line 25: Why does GH mark this file as changed?  I do not see anything is changed.

thedadams: The `o` in `ownerBindingsAnno` is now capitalized so it can be used in other packages.

pkg/controllers/management/authprovisioningv2/machineconfigs.go
jiaqiluo: I start to think if the upgrading case is a valid one, as in the description you mentioned that the bug happens only when using the "Edit as YAML" during cluster creation?  If so, then it is ok just to return here.

thedadams: Yes, you are correct. If a user upgrades their version of Rancher and uses the "Edit as YAML" feature, then the controller will still work in this case. It just won't error for existing machine configs that don't have the creatorId annotation.

jiaqiluo (dismissed):
LGTM 💯

pkg/controllers/management/authprovisioningv2/authprovisioningv2.go
```diff
@@ -86,6 +90,9 @@ func Register(ctx context.Context, clients *wrangler.Context) error {
 	clients.Mgmt.ProjectRoleTemplateBinding().OnChange(ctx, "auth-prov-v2-prtb", h.OnPRTB)
 	clients.Provisioning.Cluster().OnChange(ctx, "auth-prov-v2-cluster", h.OnCluster)
 	clients.CRD.CustomResourceDefinition().OnChange(ctx, "auth-prov-v2-crd", h.OnCRD)
+	if features.RKE2.Enabled() {
+		clients.Dynamic.OnChange(ctx, "auth-prov-v2-rke-machine-config", validMachineConfigGVK, h.OnMachineConfigChange)
```
annablender, line 94: How are you calling validMachineConfigGVK without the schema?

pkg/controllers/management/authprovisioningv2/machineconfigs.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/rancher/rancher/pkg/controllers/management/rbac"
+	"github.com/rancher/wrangler/pkg/data"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func validMachineConfigGVK(gvk schema.GroupVersionKind) bool {
+	return gvk.Group == "rke-machine-config.cattle.io" &&
```
annablender, line 17: Will the schema or gvk ever be null here? If so, add a null check

pkg/controllers/management/authprovisioningv2/machineconfigs.go
```diff
@@ -0,0 +1,79 @@
...
+	if obj == nil {
+		return nil, nil
+	}
+
+	objMeta, err := meta.Accessor(obj)
+	if err != nil || !objMeta.GetDeletionTimestamp().IsZero() {
+		return nil, err
+	}
+
+	// if owner bindings annotation is present, the node template is in the proper namespace and has had
```
annablender, line 32: Good comment!

pkg/controllers/management/authprovisioningv2/authprovisioningv2.go
thedadams: In this case, `validMachineConfigGVK` is a function passed to the `OnChange` method. The `OnChange` method will use `validMachineConfigGVK` to determine if this controller wants to look at particular objects.

pkg/controllers/management/authprovisioningv2/machineconfigs.go
thedadams: No. In Go, only pointer types can be `nil`.

thedadams:
rancher-webhook has been bumped and this is ready.

jiaqiluo (approved):

paynejacob (approved):

Issue #36038, opened 2022-01-05
[Backport v2.5] s3 etcd backups taking longer than 5 minutes cause backup loop


rancherbot:
This is a backport issue for https://github.com/rancher/rancher/issues/34890, automatically created via rancherbot by @snasovich

Original issue description:

Backups that take longer than 5 minutes will cause Rancher to start a new backup and delete the one that is currently running. Since they all take longer than 5 minutes it will continue this cycle indefinitely.

**Rancher Server Setup**
- Rancher version: 2.6.3

**Repro steps:**
* Have RKE1 provisioned downstream cluster
* Fill etcd with lots of data (e.g. 1 GB+) - enough so it takes more than 5 minutes to complete on all of the etcd nodes.
* Configure recurring snapshots for this downstream cluster, including setting `rancher_kubernetes_engine_config.services.etcd.backup_config.timeout` to high enough value for backups to actually complete on etcd side (e.g. 900 for 15 minutes).
* Wait for a backup to start (depending on interval configured) and wait 5 minutes.

**Actual behavior:**
Since the backup doesn't complete before the sync interval, a new backup will be started. This will repeat indefinitely until a backup is completed in under 5 minutes.

**Expected behavior:**
Backups to complete.

Original description (no longer relevant as fix approach changed per https://github.com/rancher/rancher/issues/34890#issuecomment-943872638):
> The `clusterBackupCheckInterval` is hardcoded to 5 minutes and should be user-changeable via the following
> - Helm (likely)
> - ENV VAR (could be good idea)
> - ConfigMap (eh)
> - etc.
> 
> Moreover, we should considered adding a "maximum retry count" so that the check interval does not infinitely continue.

JIRA ID: SURE-3305

slickwarren:
tested on rancher v2.5-head (bd840b434201ed280db8a55c46751815a6d1904f):
rancher setup:
* downstream cluster (single node, latest k8s) with backups configured to minio through proxy
  * I have the timeout in the cluster.yaml set to 1200 seconds
minio setup:
* I have an nginx proxy (limited at 10kBps upload/download) to a minio setup. So for a 3mb file that should take ~5 minutes
  * I can confirm this by downloading an existing backup through the proxy to see the speed

Issue:
* when I use this with etcd snapshots, it seems to still have a similar issue on 2.5-head and 2.5.12-rc2 as it does on 2.6.3. 
[...]

PR #36043, opened 2022-01-05
[Backport v2.5] Better handling of still running etcd backups

jakefhyde:
Backport of https://github.com/rancher/rancher/pull/35132
Issue: https://github.com/rancher/rancher/issues/36038

Original PR description:

TLDR: Adds a check for backups that are currently running.

If a user sets the backup timeout to 10 minutes due to a slow S3-alike or similar, the backup will still be running during the next tick of the check interval. The cluster backup sync function will see that the last completed backup was longer than the backup interval (ignoring the one that's currently running) so it'll start a new backup and bulldoze the running one.

My addition basically checks that if since a backup's creation time is less than the backup config's timeout and "completed" is unknown. Then skip until the next iteration.

#34890

### DIFF
--- a/pkg/controllers/management/etcdbackup/etcdbackup.go
+++ b/pkg/controllers/management/etcdbackup/etcdbackup.go
@@ -173,6 +173,24 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
 		return nil
 	}
 
+	// check for backups in progress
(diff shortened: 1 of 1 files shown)


snasovich (approved):
The same changes as the source PR. Good to merge (only 1 "any" approval needed for straightforward backports)

Issue #35955, opened 2021-12-22
Rancher 2.6.2 upgrade 2.6.3 cluster down.


oguzhansrky:
Rancher
RKE 1 v1.21.7-rancher1-1
Docker Version 20.10.12
[486e95fbfe82099b72326d1d70660bb18d9873aa325ce81adbd29e3a0808cca6-json.log](https://github.com/rancher/rancher/files/7765094/486e95fbfe82099b72326d1d70660bb18d9873aa325ce81adbd29e3a0808cca6-json.log)

As I did in previous versions, the existing container and image are removed and the rancher command is run.

I've used this update style at least 100 times and it hasn't been a problem so far.

Use command: 

docker run -d --restart=unless-stopped \
  --name=rancher \
  -v /opt/rancher/var/lib/rancher:/var/lib/rancher \
  -v /opt/rancher/var/log:/var/log \
  -v /opt/rancher/var/lib/cni:/var/lib/cni \
  -v /opt/rancher/var/lib/kubelet:/var/lib/kubelet \
  -p 80:80 -p 443:443 \
  --privileged \
  rancher/rancher:stable

ajacques:
I also encountered an issue from v2.6.2 -> v2.6.3 where all (5) of my clusters were missing from Rancher UI, but the clusters continued running in the background and I had to restore a backup to go back to v2.6.2 to recover them.

oguzhansrky:
Hi,

If I didn't have a spare, I would already sit and cry. :)

I tried it with a clean install in my test environment, it doesn't work on it either.

For your information

pangliang:
same issue !!

mecseid:
Looks the same as https://github.com/rancher/rancher/issues/35956

pangliang:
Delete ${rancher_dir}/k3s/server/db/etcd  and ${rancher_dir}/k3s/server/db/snapshots, Rollback to 2.6.2, Can be solved temporarily

niusmallnan:
It seems that this issue happens in the single node installation.

My setup local-host-data-dir is `/opt/rancher`, which is a rancher volume mountpoint for `/var/lib/rancher`.
```
cd /opt/rancher
mkdir -p k3s/server/db/etcd
cp -rf management-state/etcd/member k3s/server/db/etcd/
```
Then, upgrade your rancher to 2.6.3. This works well for me.

l3system:
For already updated cluster I had to do the following:
stop rancher container
    cd /opt/rancher
[...]

PR #36060, opened 2022-01-06
Update k3s single node etcd migration code

annablender:
Address https://github.com/rancher/rancher/issues/35955

For an upgrade in k3s single node, I updated the etcd migration code to migrate old etcd data to embedded etcd from `/var/lib/rancher/management-state/etcd` to `/var/lib/rancher/k3s/server/db/etcd`, instead of from `/var/lib/rancher/etcd`. The `/var/lib/rancher/etcd` dir gets overwritten if you upgrade a Rancher container that has bind mounted volumes.

Customers were losing downstream clusters on upgrade. Now, if you upgrade single node in Docker all cluster data will be preserved, even if you just update the Rancher version of a container with bind mounted volumes.

### DIFF
--- a/main.go
+++ b/main.go
@@ -202,16 +202,6 @@ func initLogs(c *cli.Context, cfg rancher.Options) {
 	logserver.StartServerWithDefaults()
 }
 
-func migrateETCDlocal() {
-	if _, err := os.Stat("etcd"); err != nil {
-		return
-	}
-
-	// Purposely ignoring errors
-	_ = os.Mkdir("management-state", 0700)
-	_ = os.Symlink("../etcd", "management-state/etcd")
-}
-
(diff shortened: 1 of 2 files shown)


snasovich:
Adding @ibrokethecloud since `migrateETCDlocal` was last added years ago by him - wondering if we're missing any important reasons it was added in the first place.

snasovich:
Adding @ibuildthecloud  since `migrateETCDlocal` was last added years ago by him - wondering if we're missing any important reasons it was added in the first place.

snasovich (approved):
Changes LGTM.
Did not verify it's actually working as it's rather involved with building custom image; however given https://github.com/rancher/rancher/issues/35955#issuecomment-1006947403 it should be working.

paynejacob (approved):

Issue #35750, opened 2021-12-04
Rancher reports cluster as active even though the add-on's are not upgraded as part of k8s version upgrade


SheilaghM:
SURE-3541

**Rancher Server Setup**
- Rancher version:
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version:
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

Issue description:
Customer uses Rancher API call to determine the cluster upgrade state.

curl -s -k -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" \                                                                                                                                    
-X GET \
-H 'Accept: application/json' \
-H 'Content-Type: application/json' \
https://RANCHER_FQDN/v3/clusters/c-xxxx |jq .state
But the state changes to active even before the add-on job completes; especially the critical components like coredns and CNI were not completed. This is not an expected behavior since the timeout add-on jobs will make a partially upgraded cluster. (Affected Cluster ID c-w56h9)

Business impact:
Partially upgraded cluster.

Troubleshooting steps:
They had a situation where the add-on jobs were timed out and none of the add-ons were upgraded. 

The old add-on jobs were in removing state in the UI.
They took the below steps to complete the upgrade

Deleted the removing add-on jobs
Triggered a reconciliation by changing the job timeout
Repro steps:

Old kubelet logs were not available, so not sure how the add-on jobs were stuck in removing state
Workaround:
Is a workaround available and implemented? No
What is the workaround:

Actual behavior:
Cluster state hanged to "Active" even before add-on job completes

Expected behavior:
Cluster state should stay in the "Upgrading" state until all add-on jobs are completed.

Files, logs, traces:

Additional notes:
Logs are attached to SURE ticket.

snasovich:
The issue happens when the previous `*-deploy-job` cannot be deleted for some reason during the upgrade process. A simple way to reproduce it is to manually put a dummy finalizer on some/each of these jobs.

Also, it's important to node that addon deploy jobs are created during the upgrade and NOT cleaned up during the same upgrade, but are kept. Then, when the next upgrade is executed, these jobs are supposed to be deleted prior to creating ones needed for this new upgrade (with the same name). 

[...]

PR #36076, opened 2022-01-07
Bump rke for fix where addon jobs are stuck in removing

jakefhyde:
Related Issue: https://github.com/rancher/rancher/issues/35750

This PR bumps the version of rke to one that includes the fix for clusters reporting as active after an upgrade when removing old addons had failed. Now, a cluster will correctly transition to the `error` state.


rancherbot (approved):

rancherbot (approved):

kinarashah (commented):
I think we have validation for this in our build iirc, but we also need to update `rke` in `pkg/apis/go.mod` and run `go mod tidy` for `pkg/apis` module.

rancherbot (approved):

rancherbot (approved):

snasovich (approved):
LGTM, good to merge once the build succeeds.

Issue #27166, opened 2020-05-20
RKE Template private registry url, user and password overrides do not work


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

1. Create an RKE Template with Private Registry enabled and a default url, user and password, all set to user over-rideable.
2. Create a custom cluster using the RKE template and specify overrides for the url, user and password.
3. Perform `kubectl get clusters <cluster-id> -o yaml` against the Rancher server and observe that spec.answer.values.rancherKubernetesEngineConfig.privateRegistries[0].password, spec.answer.values.rancherKubernetesEngineConfig.privateRegistries[0].url and spec.answer.values.rancherKubernetesEngineConfig.privateRegistries[0].user are all set to the RKE template default, not the user overrides.

**Environment information**
- Reproduced in Rancher v2.4.3, user reported in Rancher v2.3.7

gz#10101

gz#12537

SURE-2029
SURE-2463
SURE-3259

oxr463:
Also occuring in 2.4.5 and 2.4.9 as well.

andreivnst:
Hello,

Also occuring in 2.5.3.

codyrancher:
After diving into this issue I actually think it's a two part problem but related to the same thing.

The front-end wasn't sending answers to the backend because it wasn't handling the bracket notation properly `array[0]`. And I've put together a PR to address this issue.

Unfortunately the cluster instance is showing the answer properly set but still shows `spec.answer.values.rancherKubernetesEngineConfig.privateRegistries[0].url` being the default value that's set in the question. This makes me think that the backend is also having problems with handling the bracket notation. I'm asking around to try and find out if I'm correct about this portion.

Jono-SUSE-Rancher:
This is currently scheduled for 2.6.2 but I just wanted to highlight Cody's comments for @superseb or whoever is going to grab this from the backend. It looks like we'll need both FE/BE on it.

slickwarren:
This is reproducible on 2.6-head (931a0f0):
* Create an RKE Template with Private Registry enabled and a default url, user and password, all set to user over-rideable.
* Create a custom cluster using the RKE template and specify overrides for the url, user and password.
* the run command still has the default URL for the registry
* Perform kubectl get clusters <cluster-id> -o yaml against the Rancher server and observe that 
   * spec.answer.values.rancherKubernetesEngineConfig.privateRegistries[0].password, 
[...]

PR #36084, opened 2022-01-08
Fix the issue where private registry url, user and password overrides do not work when the cluster is  provisioned using a clusterTemplate

jiaqiluo:
Problem:
When the cluster is provisioned using a clusterTemplate,  new values for the private registry are not applied.
It turns out that the key for registry-related questions contains an index, such as [0], because the data type of the field is a list of PrivateRegistry. But the backend does not handle the bracket notation properly when processing those new values.

Fix:
When a registry-related question is detected, generate a list of PrivateRegistry based on the index value and set it to the correct position in the Cluster object.

GH Issue: https://github.com/rancher/rancher/issues/27166

Related UI issue: https://github.com/rancher/dashboard/issues/3940

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -287,6 +288,12 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 
 	defaultedAnswers := make(map[string]string)
 
+	// The key in the map is used to preserve the order of registries
+	registryMap := make(map[int]map[string]interface{})
+	existingRegistries := convert.ToMapSlice(convert.ToMapInterface(existingCluster[managementv3.ClusterSpecFieldRancherKubernetesEngineConfig])[managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries])
+	for i, registry := range existingRegistries {
+		registryMap[i] = registry
+	}
 	for _, question := range clusterTemplateRevision.Spec.Questions {
 		answer, ok := allAnswers[question.Variable]
 		if !ok {
@@ -306,13 +313,37 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers")
 		}
 		keyParts := strings.Split(question.Variable, ".")
-		values.PutValue(dataFromTemplate, val, keyParts...)
+		if strings.HasPrefix(question.Variable, "rancherKubernetesEngineConfig.privateRegistries") {
+			// for example: question.Variable = rancherKubernetesEngineConfig.privateRegistries[0].url
+			index, err := getIndexFromQuestion(question.Variable)
+			if err != nil {
+				return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers to private registry")
+			}
+			key := keyParts[len(keyParts)-1]
+			if _, ok := registryMap[index]; ok {
+				registryMap[index][key] = val
+			} else {
+				registryMap[index] = map[string]interface{}{
+					key: val,
+				}
+			}
+		} else {
+			values.PutValue(dataFromTemplate, val, keyParts...)
+		}
 
 		questionMap, err := convert.EncodeToMap(question)
 		if err != nil {
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error reading clusterTemplate questions")
 		}
 		revisionQuestions = append(revisionQuestions, questionMap)
+	}
+	if len(registryMap) > 0 {
+		registries, err := convertRegistryMapToSliceInOrder(registryMap)
+		if err != nil {
+			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers to private registry")
+		}
+		// save privateRegistries back to rancherKubernetesEngineConfig
+		values.PutValue(dataFromTemplate, registries, managementv3.ClusterSpecFieldRancherKubernetesEngineConfig, managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries)
 	}
 	//save defaultAnswers to answer
 	if allAnswers == nil {
@@ -352,6 +383,35 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 	}
 
 	return dataFromTemplate, nil
+}
+
+func convertRegistryMapToSliceInOrder(registryMap map[int]map[string]interface{}) ([]map[string]interface{}, error) {
+	size := len(registryMap)
+	registries := make([]map[string]interface{}, size)
+	for k, v := range registryMap {
+		if k >= size {
+			return nil, fmt.Errorf("the index %d is out of the bound of the registry list (size of %d)", k, size)
+		}
+		registries[k] = v
+		registries[k][managementv3.PrivateRegistryFieldIsDefault] = false
+	}
+	// set the first registry in the list as the default registry if it exists
+	if len(registries) > 0 {
+		registries[0][managementv3.PrivateRegistryFieldIsDefault] = true
+	}
+	return registries, nil
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	re, err := regexp.Compile(`\[\d+\]`)
+	if err != nil {
+		return 0, err
+	}
+	target := re.FindString(question)
+	if target == "" {
+		return 0, fmt.Errorf("cannot get index from the question: %s", question)
+	}
+	return strconv.Atoi(target[1 : len(target)-1])
 }
 
 func hasTemplate(data map[string]interface{}) bool {


pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -287,6 +287,8 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 
 	defaultedAnswers := make(map[string]string)
 
+	registryConfigPath := fmt.Sprintf("%s.%s", managementv3.ClusterSpecFieldRancherKubernetesEngineConfig, managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries)
```
paynejacob, line 290: I would use strings.Join instead.  `fmt.Sprintf` relies on reflection making it more expensive.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +308,32 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers")
 		}
 		keyParts := strings.Split(question.Variable, ".")
-		values.PutValue(dataFromTemplate, val, keyParts...)
+		if strings.HasPrefix(question.Variable, registryConfigPath) {
+			key := keyParts[len(keyParts)-1]
```
paynejacob, line 312: Is there any validation on the request before this stage? If not there is a possibility for an index error here.

jiaqiluo: There is no validation on this specific field/section before this stage. But since the field comes from `clusterTemplateRevision.Spec.Questions`, unless users are manipulating the questions when creating the clusterTemplateRevision, there should not be an index error. 
However, I will add a check here to avoid the potential error.

thedadams: Regardless of the value of `question.Variable`, it is impossible to have an index error here. From the documentation: `If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.` This means that even if `question.Variable` is empty, then `keyParts` will have at least length 1.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +308,32 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+		}
 
 		questionMap, err := convert.EncodeToMap(question)
 		if err != nil {
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error reading clusterTemplate questions")
 		}
 		revisionQuestions = append(revisionQuestions, questionMap)
 	}
+	if len(existingRegistries) > 0 {
+		existingRegistries[0][managementv3.PrivateRegistryFieldIsDefault] = interface{}(true)
```
thedadams, line 333: Is the cast here necessary?
```suggestion
		existingRegistries[0][managementv3.PrivateRegistryFieldIsDefault] = true
```

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +308,32 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers")
 		}
 		keyParts := strings.Split(question.Variable, ".")
-		values.PutValue(dataFromTemplate, val, keyParts...)
+		if strings.HasPrefix(question.Variable, registryConfigPath) {
+			key := keyParts[len(keyParts)-1]
+			index, err := strconv.Atoi(string(question.Variable[len(registryConfigPath)+1]))
```
thedadams, line 313: Does this work for double-digit indexes? Do we need to support double-digit indexes?

thedadams (changes requested):

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: Technically it could be a two-digit index, but in practice, the index should be never greater than 1 as Rancher UI only allows to set one private registry for your cluster. 
In my option, treating the index as single-dight should be sufficient. What do you think? @thedadams

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: Good point! 
fixd.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: closed this one since it is not an issue.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -287,6 +287,8 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 
 	defaultedAnswers := make(map[string]string)
 
+	registryConfigPath := strings.Join([]string{managementv3.ClusterSpecFieldRancherKubernetesEngineConfig, managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries}, ".")
+	existingRegistries := convert.ToMapSlice(convert.ToMapInterface(existingCluster[managementv3.ClusterSpecFieldRancherKubernetesEngineConfig])[managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries])
```
snasovich, line 291: I wouldn't call it `existingRegistries` as we mutate this slice (or reassign - doesn't matter). The fact is that by assignment time it's not necessarily _existing_ registries. So I would just call it `registries`.

pkg/api/norman/store/cluster/cluster_store.go
snasovich, line 313: This is pretty hacky and requires at least a comment - that we get a digit from `rancherKubernetesEngineConfig.privateRegistries[0].password` this way. It's not trivial by looking at it.
Overall this does look a bit hacky and I was hoping there is a cleaner way to extract index that works not only when there is a single digit.

snasovich (changes requested):
Have couple minor comments to address.

Overall the solution seems to be fixing the issue, but ideally, it would be nice to address the issue in a more generic way to handle array expressions, not just for `privateRegistries`. I realize this approach may be too involved as it doesn't seem we have existing helper methods in norman/wrangler to handle this (e.g. an improved version of `values.PutValue` than handles arrays), so I'm OK with leaving the current approach.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: Update: the PR is updated to handle the index regardless of its many digits.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: The variable is not renamed as it is now used solely for existing registries.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: A new logic is added to handle the index correctly.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +391,26 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	var index int
+	var err error
+	re1, _ := regexp.Compile(`\[\d+\]$`)
+	target := re1.FindString(question)
+	if target != "" {
+		target = target[1 : len(target)-1]
+		index, err = strconv.Atoi(target)
```
paynejacob, line 409: you could remove this error check, the regex will fail if it is not a valid number.  

NIT: this function is a good candidate for some unit tests.

paynejacob (approved):

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +391,26 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 	return dataFromTemplate, nil
 }
 
+func convertRegistryMapToSliceInOrder(registryMap map[int]map[string]interface{}) []map[string]interface{} {
+	registries := make([]map[string]interface{}, len(registryMap))
+	for k, v := range registryMap {
+		registries[k] = v
```
snasovich, line 397: Index `k` may be out of bounds if somehow we have request that contains something like `rancherKubernetesEngineConfig.privateRegistries[7].url` when we only have 3 private registries already set on the cluster. I think previous solution was safer/better as we were just essentially adding it as privateRegistries[3] in this case (as we always appended to the list if index didn't exist).

Still, we're talking hypothetical situations here so this solution feels overly complex for fixing the issue that doesn't exist.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +315,42 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers")
 		}
 		keyParts := strings.Split(question.Variable, ".")
-		values.PutValue(dataFromTemplate, val, keyParts...)
+		if strings.HasPrefix(question.Variable, "rancherKubernetesEngineConfig.privateRegistries") {
+			// for example: question.Variable = rancherKubernetesEngineConfig.privateRegistries[0].url
+			key := keyParts[len(keyParts)-1]
```
snasovich, line 320: `key` is not needed until like 325, so I would recommend moving it just above `registry, ok := registryMap[index]`. Generally, it's a good idea to have variable declaration and usage as close as possible (there are some exceptions if we have like a whole variable declaration block, but it's not the case here).

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +315,42 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+		}
 
 		questionMap, err := convert.EncodeToMap(question)
 		if err != nil {
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error reading clusterTemplate questions")
 		}
 		revisionQuestions = append(revisionQuestions, questionMap)
 	}
+	registries := convertRegistryMapToSliceInOrder(registryMap)
+	// set the first registry in the list as the default registry
```
snasovich, line 345: Do we want to set it to `false` for the remaining ones?
Also, can `registries[0]` be `nil` if `registries` is a non-empty slice? I don't think it is possible. If there are existing registries, there will always be something in `registryMap` at key 0 which is then used to build `registries` and I don't see it ever removed from there. If it's empty to start, we would really be setting zero-index element via `privateRegistries[0]` (if we try to set something rather than `privateRegistries[0]` then `convertRegistryMapToSliceInOrder` will fail per another comment).
In this case we can simply check with registries are non-empty and set `registries[0]` instead of having for-loop.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +391,26 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+		registries[k] = v
+	}
+	return registries
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	var index int
+	var err error
+	re1, _ := regexp.Compile(`\[\d+\]$`)
+	target := re1.FindString(question)
```
snasovich, line 406: If there was no regex match, we would return `(0, nil)` - so caller will think index is 0. I think we should return an error in this case as no match by this regexp means we have malformed request.

snasovich (changes requested):
Left some comments to address.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: fixed bu removing the error check

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +391,26 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+	for k, v := range registryMap {
+		registries[k] = v
+	}
+	return registries
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	var index int
+	var err error
+	re1, _ := regexp.Compile(`\[\d+\]$`)
```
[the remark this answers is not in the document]
jiaqiluo: Fixed by using `regexp.MustCompile` and chaining calls.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: The out-of-bound issue will not occur because we only process the answers to which questions are in `clusterTemplateRevision.Spec.Questions`:
```
for _, question := range clusterTemplateRevision.Spec.Questions {
```

let's say we have 3 private registries defined in the cluster template, then the questions will be 
```
rancherKubernetesEngineConfig.privateRegistries[0].url
rancherKubernetesEngineConfig.privateRegistries[0].username
rancherKubernetesEngineConfig.privateRegistries[0].password
rancherKubernetesEngineConfig.privateRegistries[1].url
rancherKubernetesEngineConfig.privateRegistries[1].username
rancherKubernetesEngineConfig.privateRegistries[1].password
rancherKubernetesEngineConfig.privateRegistries[2].url
rancherKubernetesEngineConfig.privateRegistries[2].username
rancherKubernetesEngineConfig.privateRegistries[2].password
```
If the request contains `rancherKubernetesEngineConfig.privateRegistries[7].url`, it will be dropped because it is not in the above list. 

However, there is a special case: the cluster template is created in a malformed format where the indexes for private registries are not continuous values, which I think should never happen. But even in this case, the current implementation should still be able to process the answers properly.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: Fixed by moving the line to above where it is used

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: I updated the logic to set `isDefault` true for the 1st and false for the rest. 

I checked if registries[i] is nil because I was thinking about the case where the cluster template is created in a malformed format where the indexes for private registries are not continuous values (see my response in another comment).  But you are right that `convertRegistryMapToSliceInOrder` will fail first if that is the case. 

My latest implementation is based on the assumption that the indexes are continuous values( but the order could be random). This brings a general question: should we consider the above-mentioned special case?

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: Good catch!  The `err` has a default value now, so `(0, nil)` will be returned only when 0 is the valid index.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -287,6 +288,14 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 
 	defaultedAnswers := make(map[string]string)
 
+	// The key in the map is used to preserve the order of registries
+	registryMap := make(map[int]map[string]interface{})
+	existingRegistries := convert.ToMapSlice(convert.ToMapInterface(existingCluster[managementv3.ClusterSpecFieldRancherKubernetesEngineConfig])[managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries])
+	if existingRegistries != nil {
```
thedadams, line 294: Don't need this check. If `existingRegistries` is `nil`, then the `for` loop will be skipped (as if `existingRegistries` was empty).

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +315,41 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
-		values.PutValue(dataFromTemplate, val, keyParts...)
+		if strings.HasPrefix(question.Variable, "rancherKubernetesEngineConfig.privateRegistries") {
+			// for example: question.Variable = rancherKubernetesEngineConfig.privateRegistries[0].url
+			index, err := getIndexFromQuestion(question.Variable)
+			if err != nil {
+				return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers to private registry")
+			}
+			key := keyParts[len(keyParts)-1]
+			registry, ok := registryMap[index]
+			if ok {
```
thedadams, line 326: nit:
```suggestion
			if registry, ok := registryMap[index]; ok {
```

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +315,41 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+				registry[key] = val
+				registryMap[index] = registry
```
thedadams, line 328: Can these be combined? Then you wouldn't need the `registry` variable. This would save an extra assignment, but I guess is mostly a style thing. So, consider it a nit.
```suggestion
				registryMap[index][key] = val
```

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +315,41 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
 		questionMap, err := convert.EncodeToMap(question)
 		if err != nil {
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error reading clusterTemplate questions")
 		}
 		revisionQuestions = append(revisionQuestions, questionMap)
 	}
+	registries := convertRegistryMapToSliceInOrder(registryMap)
+	// set the first registry in the list as the default registry
+	for i := 0; i < len(registries); i++ {
+		registries[i][managementv3.PrivateRegistryFieldIsDefault] = interface{}(false)
```
thedadams, line 347: Is this cast necessary? The reason I ask is, as I understand it, casting creates another allocation. Same with the one below.
```suggestion
		registries[i][managementv3.PrivateRegistryFieldIsDefault] = false
```

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +390,24 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 	return dataFromTemplate, nil
 }
 
+func convertRegistryMapToSliceInOrder(registryMap map[int]map[string]interface{}) []map[string]interface{} {
```
thedadams, line 393: This function is not robust enough to handle the example we spoke about "offline." It will panic for this example you provided:
>let’s say len(existingRegistries) = 0 and the data in the api call is
"rancherKubernetesEngineConfig.privateRegistries[3].url":"aaa",
"rancherKubernetesEngineConfig.privateRegistries[3].user":"aaa"
"rancherKubernetesEngineConfig.privateRegistries[3].password":"aaa"

This is because `len(registryMap) == 1`, so the call to `registries[k] = v` will be out of bounds because `k == 3`.

I think handling this properly is equivalent to not using the `map[int]map[string]interface{}` at all. Meaning, that if you can solve this problem, then you would have to use the `map[int]map[string]interface{}` and would be able to continue to use the `[]map[string]interface{}` above.

thedadams (changes requested):

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +390,24 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+	for k, v := range registryMap {
+		registries[k] = v
+	}
+	return registries
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	var index int
+	err := fmt.Errorf("cannot get index from the question: %s", question)
+	target := regexp.MustCompile(`\[\d+\]`).FindString(question)
```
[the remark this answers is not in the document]
jiaqiluo: I switched from `regexp.Compile`  to ` regexp.MustCompile` because the regular expression is hardcoded so it cannot fail to be complied.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: If we do not convert it to an interface, we will have a runtime error saying that an interface{} is expected but a Boolean is provided.

pkg/api/norman/store/cluster/cluster_store.go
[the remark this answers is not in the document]
snasovich: @thedadams , this was changed based on something I commented on before.
https://github.com/rancher/rancher/pull/36084#discussion_r782439144

I'm fine switching back to `regexp.Compile` provided we handle any errors it may return. I generally find using `MustCompile` safe when hardcoded regexps are used based on the documentation note from https://pkg.go.dev/regexp#MustCompile: _MustCompile is like Compile but panics if the expression cannot be parsed._

pkg/api/norman/store/cluster/cluster_store.go
thedadams: So this doesn't apply here? https://go.dev/play/p/L73ZpMddpyx

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: fixed.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: fixed.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: This is discussed offline and we decide to return an error if the index is not consecutive and less than the count of registries.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: This is discussed offline and we decide to return an error if the index is not consecutive and less than the count of registries.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: You are right. The conversion is removed.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: discussed offline and this is properly handled now.

pkg/api/norman/store/cluster/cluster_store.go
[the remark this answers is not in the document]
jiaqiluo: switched back to `regex.Complie` with the proper handle of errors.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: fixed.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +389,30 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 	return dataFromTemplate, nil
 }
 
+func convertRegistryMapToSliceInOrder(registryMap map[int]map[string]interface{}) ([]map[string]interface{}, error) {
+	size := len(registryMap)
+	registries := make([]map[string]interface{}, size)
+	for k, v := range registryMap {
+		if k > size {
```
thedadams, line 396:
```suggestion
		if k >= size {
```

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +313,42 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
 	}
+	registries, err := convertRegistryMapToSliceInOrder(registryMap)
+	if err != nil {
+		return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error processing clusterTemplate answers to private registry")
+	}
+	// set the first registry in the list as the default registry
+	for i := 0; i < len(registries); i++ {
+		registries[i][managementv3.PrivateRegistryFieldIsDefault] = false
+	}
+	registries[0][managementv3.PrivateRegistryFieldIsDefault] = true
```
thedadams, line 348: Can this logic be added to `convertRegistryMapToSliceInOrder` so we only loop through these items once?

At the very least, the loop variable `i` here should start at 1 instead of 0.

thedadams (changes requested):

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: fixed.

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: at first, I thought the `convertRegistryMapToSliceInOrder` function should not modify the content, but now I think it makes more sense to just let it return the finalized data that can be used for the next step, and of course, saves a loop.

thedadams (dismissed):

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +383,33 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+		registries[k] = v
+		registries[k][managementv3.PrivateRegistryFieldIsDefault] = false
+	}
+	// set the first registry in the list as the default registry
+	registries[0][managementv3.PrivateRegistryFieldIsDefault] = true
+	return registries, nil
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	re, err := regexp.Compile(`\[\d+\]`)
```
paynejacob, line 402: I would define this as a var above the function with regex must compile.  The fewer error cases to consiter the better.

paynejacob (approved):

pkg/api/norman/store/cluster/cluster_store.go
jiaqiluo: We have had multiple conversations on using whether `regexp.Compile` or `regexp.MustCompile` in this PR and decided to go with the current implementation.  

I like your suggestion about defining the expression as a var, but I will keep it as it is now to avoid another round of approval.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -306,14 +313,38 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+			values.PutValue(dataFromTemplate, val, keyParts...)
+		}
 
 		questionMap, err := convert.EncodeToMap(question)
 		if err != nil {
 			return nil, httperror.WrapAPIError(err, httperror.ServerError, "Error reading clusterTemplate questions")
 		}
 		revisionQuestions = append(revisionQuestions, questionMap)
 	}
+	if len(registryMap) > 0 {
```
jiaqiluo, line 340: Change 1/2 after previous approval.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +385,35 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+	registries := make([]map[string]interface{}, size)
+	for k, v := range registryMap {
+		if k >= size {
+			return nil, fmt.Errorf("the index %d is out of the bound of the registry list (size of %d)", k, size)
+		}
+		registries[k] = v
+		registries[k][managementv3.PrivateRegistryFieldIsDefault] = false
+	}
+	// set the first registry in the list as the default registry if it exists
+	if len(registries) > 0 {
```
jiaqiluo, line 399: Change 2/2 after previous approval.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -354,6 +385,35 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
...
+		registries[0][managementv3.PrivateRegistryFieldIsDefault] = true
+	}
+	return registries, nil
+}
+
+func getIndexFromQuestion(question string) (int, error) {
+	re, err := regexp.Compile(`\[\d+\]`)
```
paynejacob, line 406: nit: per our discussions this should be moved to a var outside the function created with `MustCompile`.

paynejacob (approved):

snasovich (approved):

PR #36094, opened 2022-01-10
Removing duplicated rule addition for restricted-admin

MbolotSuse:
Removing calls to rolebuilder which overwrote earlier defined
permissions - and a function only used in this process.

### DIFF
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -105,14 +105,6 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext) 
 		addRule().apiGroups("management.cattle.io").resources("cisconfigs").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("cisbenchmarkversions").verbs("get", "list", "watch")
 
-	restrictedAdminRole := addUserRules(rb.addRole("Restricted Admin", "restricted-admin"))
-	restrictedAdminRole.
-		addRule().apiGroups("management.cattle.io").resources("clustertemplates").verbs("*").
-		addRule().apiGroups("management.cattle.io").resources("globalroles", "globalrolebindings").verbs("*").
-		addRule().apiGroups("management.cattle.io").resources("users", "userattribute", "groups", "groupmembers").verbs("*").
-		addRule().apiGroups("management.cattle.io").resources("podsecuritypolicytemplates").verbs("*").
-		addRule().apiGroups("management.cattle.io").resources("fleetworkspaces").verbs("*")
-
 	rb.addRole("User Base", "user-base").
 		addRule().apiGroups("management.cattle.io").resources("preferences").verbs("*").
 		addRule().apiGroups("management.cattle.io").resources("settings").verbs("get", "list", "watch").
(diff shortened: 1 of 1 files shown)


MbolotSuse:
Closing PR. Going to try and fix this another way.

MbolotSuse:
Reopening PR - this method makes the most sense given the 2.5.12 timeline.

prachidamle:
@MbolotSuse so after removing calls to rolebuilder, does the role permissions for "restrictedAdmin" role in 2.5.9 look exactly same as in these for 2.5.12?

MbolotSuse:
> @MbolotSuse so after removing calls to rolebuilder, does the role permissions for "restrictedAdmin" role in 2.5.9 look exactly same as in these for 2.5.12?

Yes - they have the same permissions (on a fresh install).

One major difference to note - 2.5.10+ had the fleet changes included (there's a controller which IIRC should dynamically change the permissions of the restricted admin). So the permissions of a restrictedAdmin in 2.5.10+ will differ from the 2.5.9 over time.

prachidamle:
@MbolotSuse We will need to also include the relevant permissions needed to make this usecase work - "Bind all fleetworkspace except fleet-loal to restricted admin"  https://github.com/rancher/rancher/commit/daf09ba4d88f96cdd03fab6430faed2ee1d4e729

MbolotSuse:
@prachidamle Correct me if I am wrong here, but since https://github.com/rancher/rancher/blob/23f718980f2218ecfdf13d3bfaebee6b44abe593/pkg/data/management/role_data.go#L79 gives `*` verbs for `*` resources on `management.cattle.io` we don't need to specify a `fleetworkspaces` resource correct?

Looking at the linked commit, the permissions added in role data for the restricted admin seem to be broken down into 3 api groups:

`managment.cattle.io` - covered by https://github.com/rancher/rancher/blob/23f718980f2218ecfdf13d3bfaebee6b44abe593/pkg/data/management/role_data.go#L79 
`project.cattle.io` - covered by https://github.com/rancher/rancher/blob/23f718980f2218ecfdf13d3bfaebee6b44abe593/pkg/data/management/role_data.go#L80
`""`,`secrets`, `create` - covered by https://github.com/rancher/rancher/blob/23f718980f2218ecfdf13d3bfaebee6b44abe593/pkg/data/management/role_data.go#L84

Can you clarify which permissions you see as missing?

prachidamle:
@MbolotSuse Correct, that should cover all perms needed. Have we run tests for the usecase broken, Fleet usecase and the upgrade case from 2.5.11 -> 2.5.12?

MbolotSuse:
@prachidamle I've ran a test on fresh-install for the broken use-case. I'll post back once I've got the other test cases ran as well.

ryansann:
The `restriced-admin` global role does not look to be dynamic in 2.5, i.e. the `restrictedadminrbac` controller is not modifying the globalrole / clusterrole for restriced-admin

Also all the rolebuilder logic that is being removed looks to be covered by the prior restricted-admin role building. Pending some testing, this lgtm

ryansann (approved):
Validated fleet use cases on downstream as restricted-admin and saw no issues, approving

MbolotSuse:
Upgrade test case- 
Ran local docker image with version 2.5.11. Imported an existing cluster, and created a restricted admin (couldn't see either of the clusters). Used the [established single node upgrade process](https://rancher.com/docs/rancher/v2.5/en/installation/other-installation-methods/single-node-docker/single-node-upgrades/) to upgrade to a new version using the code in this PR. Restricted admin could now see the clusters and the clusters were still there.

prachidamle (approved):

PR #36099, opened 2022-01-10
Update tags for v2.5.12-rc2

brendarearden:


cbron:
Hard merging to avoid build problems, previous amd64 builds have passed.

Issue #36109, opened 2022-01-11
Files secret not passed to rancher-machine command


thedadams:
**Rancher Server Setup**
- Rancher version: v2.6-head

**Information about the Cluster**
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Provisioned RKE2

**Describe the bug**
When passing cloud-config data to the create job, a files secret is created and things like the cloud-config contents are put into the secret and volumes are created for the job. This construction needs to happen before the `cmd` is created so the file gets passed to the command instead of its contents. A previous change reorder this and broke it.

**To Reproduce**
 
1. Provision an RKE2 cluster and pass cloud-config information. The bug was found with VSphere.

**Result**
The machine will not be created.

**Expected Result**
The machine should be created and the cluster should come up.

slickwarren:
tested using 2.6.4-rc3 in vsphere. 

cloud config used: 
```
#cloud-config
password: password
chpasswd: {expire: False}
sshpwauth: True
```

* provision single node, all roles rke2 cluster (1.22.6-rke2r1) with above cloud config -- fail


Note: I was able to provision the same cluster, without the cloud config, with no issues.

slickwarren:
tested on v2.6-head (024e6fa):
user is now able to add cloud config data to both rke1 and rke2 clusters. Example used with vsphere for 1.23 (for both rke1 and rke2):
```
#cloud-config
users:
  - default
  - name: rancher
    sudo: ALL=(ALL) NOPASSWD:ALL
    lock_passwd: false
    no_ssh_fingerprints: true
    groups: users, sudo
    hashed_passwd: $HASHED_PASSWD
    ssh_authorized_keys:
    - $YOUR_SSH_KEY
ssh_authorized_keys:
  - $YOUR_SSH_KEY
write_files:
- path: /root/CLOUD_INIT_WAS_HERE
```

PR #36110, opened 2022-01-11
Construct files secret before cmd

thedadams:
The constructFilesSecret function will copy the contents of files like
cloud-config to a secret to be passed as a volume to the machine create job. A
previous change reorder these calls, so it did not work correctly. This change
reorders these calls for the intended outcome.

Issue:
https://github.com/rancher/rancher/issues/36109


jakefhyde (approved):

rosskirkpat (approved):
I validated that with this change, I am now seeing `cloud-config:/path/to/machine/files/cloud-config` being passed as an ARG instead of the entire contents of the cloud-config file.

jiaqiluo (approved):

PR #36113, opened 2022-01-11
Fix bug where tokens were always being created for the admin user

jakefhyde:
Previously, the `AsUser` function was using the token client to create a new token for the specified ID, however rancher will create a derived token based on the bearer token, and disregard the token in the request. The framework has been updated to use the login action for parity the python framework.

### DIFF
--- a/tests/framework/clients/rancher/client.go
+++ b/tests/framework/clients/rancher/client.go
@@ -1,10 +1,15 @@
 package rancher
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 
 	"github.com/pkg/errors"
+	"github.com/rancher/norman/httperror"
 	frameworkDynamic "github.com/rancher/rancher/tests/framework/clients/dynamic"
 	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
 	"github.com/rancher/rancher/tests/framework/clients/rancher/provisioning"
@@ -29,7 +34,7 @@ type Client struct {
 	Session       *session.Session
 }
 
-// NewClient is the constructor to the intializing a rancher Client. It takes a bearer token and session.Session. If bearer token is not provided,
+// NewClient is the constructor to the initializing a rancher Client. It takes a bearer token and session.Session. If bearer token is not provided,
 // the bearer token provided in the configuration file is used.
 func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 	rancherConfig := new(Config)
@@ -88,13 +93,47 @@ func clientOpts(restConfig *rest.Config, rancherConfig *Config) *clientbase.Clie
 	}
 }
 
-// AsUser accepts a user object, and then creates a token for said `user`. Then it instantiates and returns a Client using the token created
-func (c *Client) AsUser(user *management.User) (*Client, error) {
-	token := &management.Token{
-		UserID: user.ID,
+// doAction is used to post an action to an endpoint, and marshal the response into the output parameter.
+func (c *Client) doAction(endpoint, action string, body []byte, output interface{}) error {
+	url := "https://" + c.restConfig.Host + endpoint + "?action=" + action
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(body))
+	if err != nil {
+		return err
 	}
 
-	returnedToken, err := c.Management.Token.Create(token)
+	req.Header.Add("Authorization", "Bearer "+c.restConfig.BearerToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := c.Management.APIBaseClient.Ops.Client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 300 {
+		return httperror.NewAPIErrorLong(resp.StatusCode, resp.Status, url)
+	}
+
+	byteContent, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if len(byteContent) > 0 {
+		err = json.Unmarshal(byteContent, output)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	return fmt.Errorf("received empty response")
+}
+
+// AsUser accepts a user object, and then creates a token for said `user`. Then it instantiates and returns a Client using the token created.
+// This function uses the login action, and user must have a correct username and password combination.
+func (c *Client) AsUser(user *management.User) (*Client, error) {
+	returnedToken, err := c.login(user)
 	if err != nil {
 		return nil, err
 	}
@@ -103,7 +142,7 @@ func (c *Client) AsUser(user *management.User) (*Client, error) {
 }
 
 // WithSession accepts a session.Session and instantiates a new Client to reference this new session.Session. The main purpose is to use it
-// when created "sub sessions" when tracking resources created at a test case scope
+// when created "sub sessions" when tracking resources created at a test case scope.
 func (c *Client) WithSession(session *session.Session) (*Client, error) {
 	return NewClient(c.restConfig.BearerToken, session)
 }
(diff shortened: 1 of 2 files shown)


igomez06 (commented):
@jakefhyde I want to understand the issue. Are you attempting to use the `AsUser` function for an admin user? The client created using the bearer token is your admin user, you shouldn't use `AsUser` for an admin user.

jakefhyde:
@igomez06 The description for the function `AsUser` is "accepts a user object, and then creates a token for said `user`", however it was creating a derived token for the user of the bearer token, which was always the admin user. Beforehand, the line `testUser, err := client.AsUser(p.testUser)` would have created a new token, but it would have been for the admin user, so we were effectively creating users, but doing all testing with a new token for the admin user.

The token create handler starts here: https://github.com/rancher/rancher/blob/3fc44ba6ca156581930e1d836334ac48e6b80707/pkg/auth/tokens/api.go#L70-L73

and ends here: https://github.com/rancher/rancher/blob/3fc44ba6ca156581930e1d836334ac48e6b80707/pkg/auth/tokens/manager.go#L109-L118

We can see here in `createDerivedToken`, the provided UserID (from `jsonInput`) is being disregarded for the `tokenAuthValue`, which comes from the request:
https://github.com/rancher/rancher/blob/3fc44ba6ca156581930e1d836334ac48e6b80707/pkg/auth/tokens/manager.go#L274

igomez06:
@jakefhyde 
Yes I see, you are completely correct. `AsUser` is not working as expected. I think your approach makes sense, as well. But a quick question, wouldn't this be an issue with the `createDerivedToken`? Why is it disregarding the `jsonInput` for the `tokenAuthValue` and only using the `jsonInput` for `Description` and `ClusterName`?

jakefhyde:
@igomez06 I agree, it is not very straightforward. Currently looking into why it is structured this way, I'll update when I have more info.

igomez06:
> @igomez06 I agree, it is not very straightforward. Currently looking into why it is structured this way, I'll update when I have more info.

Thank you!

jakefhyde:
@igomez06 Update: this endpoint is used by the UI to create a new token for the currently logged in user via the Account & API Keys page, which is why only `Description`, `ClusterName` and `TTLMillis` are used from the request.

igomez06 (approved):
That makes sense @jakefhyde, your implementation looks good as well.

tests/framework/clients/rancher/client.go
```diff
@@ -88,13 +93,66 @@ func clientOpts(restConfig *rest.Config, rancherConfig *Config) *clientbase.Clie
...
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	return fmt.Errorf("received empty response")
+}
+
+func (c *Client) getToken(user *management.User) (*management.Token, error) {
+	token := &management.Token{}
```
igomez06, line 133: I think we should move these functions outside of `client.go`. Although they are private, I don't think they should belong here. Unless you have a valid reason. I think the implementation is good, I'm just not convince they should be in this file, since it's so specific to getting a token from a created user. May they should go in extensions/users/users.go

tests/framework/extensions/users/users.go
```diff
@@ -96,3 +98,14 @@ func RemoveProjectMember(rancherClient *rancher.Client, user *management.User) e
 	}
 	return rancherClient.Management.ProjectRoleTemplateBinding.Delete(&roleToDelete)
 }
+
+// AsUser accepts a user object, and then creates a token for said `user`. Then it instantiates and returns a Client using the token created.
+// This function uses the login action, and user must have a correct username and password combination.
+func AsUser(s suite.Suite, c *rancher.Client, user *management.User) (*rancher.Client, error) {
+	returnedToken, err := auth.Login(s, c, user)
```
igomez06, line 105: I think this would be better in` tests/framework/clients/rancher/client.go`, as it's returning a client instantiated of the `user` parameter.  `WithSession(session *session.Session)` does something similar, so I think it fits better there. Everything else is good, that is my last nitpick. Good job!

jakefhyde: Awesome thanks! Just had one question about the `project-member` test. Since the testnamespace is created with the testuser client it tries to delete the namespace on cleanup, but it doesn't have the correct rolebinding to be able to do so (and as far as I can tell, [only admin can delete namespaces](https://github.com/rancher/rancher/blob/d8ebe6252ec8e6b6dcd55ae3c592477842a43d93/pkg/data/management/role_data.go#L86-L88)). Should I do anything about this? I was able to remove one of two errors (as `project-members` cannot watch the namespace either), but do you think it would be more user friendly to pass the admin client that we wish to use for cleanup when we delete the namespace (or do you have something more straightforward in mind)?

igomez06: Interesting, I was under the impression, that a project-member (as long as they are a member of that specific project) were allowed to created and delete a namespace. Hmmm ... that seems wrong, I checked a test plan we had and that is the expected behavior

tests/framework/clients/rancher/client.go
```diff
@@ -88,13 +95,47 @@ func clientOpts(restConfig *rest.Config, rancherConfig *Config) *clientbase.Clie
...
+	url := "https://" + c.restConfig.Host + endpoint + "?action=" + action
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(body))
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+c.restConfig.BearerToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := c.Management.APIBaseClient.Ops.Client.Do(req)
```
igomez06, line 109: I just noticed this, I think after the request we should register a delete function, to make sure the token is deleted as well on cleanup.

tests/framework/extensions/auth/auth.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/stretchr/testify/suite"
+
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+)
+
+// Login uses the local authentication provider to authenticate a user and return the subsequent token.
+func Login(s suite.Suite, c *rancher.Client, user *management.User) (*management.Token, error) {
+	if user.Password == "" {
+		s.Fail("Password cannot be empty")
```
igomez06, line 15: Another nitpick sorry, why not just return an error here, instead of failing? We get the error, and fail on the actual test, that way we don't have to pass it the test suite.

tests/framework/extensions/users/users.go
jakefhyde: I just verified that a user can create and delete namespaces with the `project-member` role, not sure why this is any different. I'll update when I have more info, but for now, these are the errors I receive:
```
session.go:47: error calling cleanup function: unknown
session.go:47: error calling cleanup function: namespaces "testnamespace" is forbidden: User "u-c2h7z" cannot delete resource "namespaces" in API group "" in the namespace "testnamespace": RBAC: clusterrole.rbac.authorization.k8s.io "p-hhg9w-namespaces-edit" not found
```
The `unknown` error comes from [here](https://github.com/rancher/rancher/blob/32fa450375fe52720ce2f44db1b243736929ea21/tests/framework/extensions/namespaces/namespaces.go#L70-L73), which is actually a 403 response like this:
```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"namespaces \"testnamespace\" is forbidden: User \"u-6cxpn\" cannot watch resource \"namespaces\" in API group \"\" at the cluster scope: RBAC: clusterrole.rbac.authorization.k8s.io \"p-gtskb-namespaces-edit\" not found","reason":"Forbidden","details":{"name":"testnamespace","kind":"namespaces"},"code":403}
```
Also, I may have to move the login action from `extensions/auth`, or move `AsUser` from `client`, as currently they are creating a cyclic package dependency. Do you have any strong preference one way or the other for how I should approach that?

igomez06: Sounds good Jake thanks for doing that. Move `Login` I think `AsUser` best fits in `client.go` Thank you!

tests/framework/clients/rancher/client.go
paynejacob: an alternative would be to make this a private method given we only use it for the token shenanigans.

tests/framework/extensions/auth/auth.go
```diff
@@ -0,0 +1,34 @@
...
+	"encoding/json"
+
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/stretchr/testify/suite"
+
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+)
+
+// Login uses the local authentication provider to authenticate a user and return the subsequent token.
+func Login(s suite.Suite, c *rancher.Client, user *management.User) (*management.Token, error) {
```
paynejacob, line 13: The client should not depend on extension move this to the client.

paynejacob (changes requested):

paynejacob (approved):

Issue #36120, opened 2022-01-12
[2.5.12] Adding https://git.rancher.io/helm3-charts does not result in any charts being returned


nwmac:
See UI issues: https://github.com/rancher/dashboard/issues/4902 and https://github.com/rancher/dashboard/issues/4812

When you add the chart repo `https://git.rancher.io/helm3-charts` with the name` helm3-library` to Rancher using the dashboard UI, the API call to /v1/management.cattle.io.catalogs/helm3-library returns no charts.

In 2.6 this returns the charts from that repository.

In order to resolve the reported UI issue, the backend needs to provide the same behaviour as 2.6.

aiyengar2:
@sowmyav27 we should also check whether this impacts 2.6 and test use cases around that. From what I can tell, we should be able to reproduce this in any 2.6.3-rcX version.

Filed another PR to resolve this issue in 2.6.4

aiyengar2:
For those who encounter this issue, the root cause of this is that the semver library that Rancher uses to compare the constraint string provided in the `catalog.cattle.io/rancher-version` annotation assumes that a constraint string that has no prerelease version (e.g. `<2.6.3`) always returns false for pre-release versions (e.g. `2.5.12-rc2` would return false for that constraint, although semantically it should return true).

Therefore, this should only impact RC / pre-release versions of Rancher and should not impact any version of Rancher that has been officially released.

brendarearden:
Available to test in rancher/rancher v2.5.12-rc4

anupama2501:
Verified 2.5.12-rc4

1. Created aws-event-sources in the ember UI 0.1.1
2. Migrate the app to the dashboard UI by adding the required annotations and the helm3-library repo from apps & marketplace.
3. Verified the app versions reflected are from the catalog that was added 
4. Upgrade went through fine

tomasnorre:
I'm still getting this error with following command. 

```
helm install rancher rancher-stable/rancher \
  --namespace cattle-system \
  --set hostname=rancher.example.com
```

I'm still a little rusty in helm so might be something that I'm missing, but a pointer would be really helpful.

PR #36124, opened 2022-01-12
Support checking rancher-version constraints against Rancher prereleases

aiyengar2:
The package used to make semver comparisions for catalog v2, e.g. https://github.com/Masterminds/semver, has a known constraint when it comes to working with prerelease versions, as described in https://github.com/Masterminds/semver#working-with-prerelease-versions:

```
SemVer comparisons using constraints without a prerelease comparator will skip prerelease versions. 
```

As a result, if the `catalog.cattle.io/rancher-version` is set to something like `< v2.6.3`, something like `v2.5.9-rcXX` will automatically fail due to the fact that it has a pre-release identifier attached to it (namely `-rcXX`).

Therefore, the fix put forward in this PR first checks to see if the Rancher version is valid for the given constraint; if the output of the validation indicates that it is invalid because the constraint does not match against pre-release versions, it will try to validate against a version with the pre-release omitted.

Related Issue: https://github.com/rancher/rancher/issues/36120

### DIFF
--- a/pkg/catalogv2/content/content.go
+++ b/pkg/catalogv2/content/content.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/url"
+	"strings"
 	"sync"
 
 	"github.com/Masterminds/semver/v3"
@@ -178,13 +179,31 @@ func (c *Manager) filterReleases(index *repo.IndexFile, k8sVersion *semver.Versi
 		logrus.Errorf("failed to parse server version %s: %v", settings.ServerVersion.Get(), err)
 		return index
 	}
+	rancherVersionWithoutPrerelease, err := rancherVersion.SetPrerelease("")
+	if err != nil {
+		logrus.Errorf("failed to remove prerelease from %s: %v", settings.ServerVersion.Get(), err)
+		return index
+	}
 
 	for rel, versions := range index.Entries {
 		newVersions := make([]*repo.ChartVersion, 0, len(versions))
 		for _, version := range versions {
 			if constraintStr, ok := version.Annotations["catalog.cattle.io/rancher-version"]; ok {
 				if constraint, err := semver.NewConstraint(constraintStr); err == nil {
-					if !constraint.Check(rancherVersion) {
+					satisfiesConstraint, errs := constraint.Validate(rancherVersion)
+					// Check if the reason for failure is because it is ignroing prereleases
+					constraintDoesNotMatchPrereleases := false
+					for _, err := range errs {
+						// Comes from error in https://github.com/Masterminds/semver/blob/60c7ae8a99210a90a9457d5de5f6dcbc4dab8e64/constraints.go#L93
+						if strings.Contains(err.Error(), "the constraint is only looking for release versions") {
+							constraintDoesNotMatchPrereleases = true
+							break
+						}
+					}
+					if constraintDoesNotMatchPrereleases {
+						satisfiesConstraint = constraint.Check(&rancherVersionWithoutPrerelease)
+					}
+					if !satisfiesConstraint {
 						continue
 					}
 				} else {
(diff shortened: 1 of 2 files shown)


aiyengar2:
We cannot move to https://github.com/blang/semver because it does not support the additional types of constraint strings in the rancher-version annotation that we already support (e.g. hyphen, wildcard, etc.) as noted in our test cases in https://github.com/rancher/rancher/blob/c775592acc9d909ef205b2cc32b5ab78eadb62f3/pkg/catalogv2/content/content_test.go#L21-L152

pkg/catalogv2/content/content.go
```diff
@@ -178,13 +179,31 @@ func (c *Manager) filterReleases(index *repo.IndexFile, k8sVersion *semver.Versi
...
+		return index
+	}
 
 	for rel, versions := range index.Entries {
 		newVersions := make([]*repo.ChartVersion, 0, len(versions))
 		for _, version := range versions {
 			if constraintStr, ok := version.Annotations["catalog.cattle.io/rancher-version"]; ok {
 				if constraint, err := semver.NewConstraint(constraintStr); err == nil {
-					if !constraint.Check(rancherVersion) {
+					satisfiesConstraint, errs := constraint.Validate(rancherVersion)
```
aiyengar2, line 193: Validate does the same thing as check except it dumps out a list of errors on why the validation failed... so we can use `satisfiesConstraint` the same way we used `constraint.Check(rancherVersion)` but be able to parse through the error to know if we want to fallback to checking without the prerelease identifier.

pkg/catalogv2/content/content.go
```diff
@@ -178,13 +179,31 @@ func (c *Manager) filterReleases(index *repo.IndexFile, k8sVersion *semver.Versi
...
+					// Check if the reason for failure is because it is ignroing prereleases
+					constraintDoesNotMatchPrereleases := false
+					for _, err := range errs {
+						// Comes from error in https://github.com/Masterminds/semver/blob/60c7ae8a99210a90a9457d5de5f6dcbc4dab8e64/constraints.go#L93
+						if strings.Contains(err.Error(), "the constraint is only looking for release versions") {
+							constraintDoesNotMatchPrereleases = true
+							break
+						}
+					}
```
aiyengar2, line 202: Q: Why can't we just always only check the prereleases?
A: If for some reason we need to compare the rancher version `v2.5.12-rc9` against the string `<2.5.12-rc10`, we cannot allow this to return false since that's invalid.

aiyengar2: added test for this.

pkg/catalogv2/content/content.go
```diff
@@ -178,13 +179,31 @@ func (c *Manager) filterReleases(index *repo.IndexFile, k8sVersion *semver.Versi
...
 		for _, version := range versions {
 			if constraintStr, ok := version.Annotations["catalog.cattle.io/rancher-version"]; ok {
 				if constraint, err := semver.NewConstraint(constraintStr); err == nil {
-					if !constraint.Check(rancherVersion) {
+					satisfiesConstraint, errs := constraint.Validate(rancherVersion)
+					// Check if the reason for failure is because it is ignroing prereleases
+					constraintDoesNotMatchPrereleases := false
+					for _, err := range errs {
+						// Comes from error in https://github.com/Masterminds/semver/blob/60c7ae8a99210a90a9457d5de5f6dcbc4dab8e64/constraints.go#L93
+						if strings.Contains(err.Error(), "the constraint is only looking for release versions") {
```
rmweir, line 198: Hopefully this error message doesn't change lol

rmweir (approved):
LGTM, although parsing the error message may end up being fragile.

jakefhyde (approved):
Great work Arvind, do we use wildcard and tilde in such a way that it impacts 2.6? (i.e. do we need to forwardport this)

aiyengar2:
> Great work Arvind, do we use wildcard and tilde in such a way that it impacts 2.6? (i.e. do we need to forwardport this)

I think we should forward port this to 2.6 since regardless we will encounter this issue in 2.6.x RCs.

w.r.t. wildcard and tildes, I'm not aware of any off the top of my head, but we definitely do use hyphens.

PR #36075, opened 2022-01-07
Update remotedialer

ibuildthecloud:
Update to include https://github.com/rancher/remotedialer/pull/38

https://github.com/rancher/rancher/issues/34819

PR #36019, opened 2022-01-04
Bump remotedialer to add wait/resume functionality

thedadams:
Issues:
https://github.com/rancher/rancher/issues/36017
https://github.com/rancher/rancher/issues/36018

PR #36141, opened 2022-01-13
Update remote dialer

thedadams:
Original 2.6 PR: https://github.com/rancher/rancher/pull/36075


rancherbot (approved):

rancherbot (approved):

snasovich (changes requested):
I think we want to bump `steve` as well since it went in as part of https://github.com/rancher/rancher/pull/36019 so to ensure we deliver the exact fix to patch we should have it too.
Please correct me if I'm wrong.

Issue #35999, opened 2022-01-03
Multiple nodes are upgraded at the same time when enabling pre/post-drain hooks during a RKE2 upgrade


bk201:
**Rancher Server Setup**
- Rancher version: `rancher/rancher:v2.6.3-rc4`
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: `v1.21.7+rke2r1`
- Cluster Type (Local/Downstream): local
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**Describe the bug**

RKE2 runtime on multiple nodes are upgraded at the same time but the `controlPlaneConcurrency` and `workerConcurrency` are both `1`.

**To Reproduce**
 
We upgrade the RKE2 cluster in Harvester with the following manifest by modifying the `clusters.provisioning.cattle.io/local` CR:

```
  spec:
    kubernetesVersion: v1.22.4+rke2r2       # upgrade from v1.21.7+rke2r1
    localClusterAuthEndpoint: {}
    rkeConfig:
      chartValues: "null"
      machineGlobalConfig: "null"
      provisionGeneration: 1
      upgradeStrategy:
        controlPlaneConcurrency: "1"
        controlPlaneDrainOptions:
          deleteEmptyDirData: true
          enabled: true
          force: true
          ignoreDaemonSets: true
          postDrainHooks:
          - annotation: harvesterhci.io/post-hook
          preDrainHooks:
          - annotation: harvesterhci.io/pre-hook
          timeout: 0
        workerConcurrency: "1"
        workerDrainOptions:
          deleteEmptyDirData: true
          enabled: true
          force: true
          ignoreDaemonSets: true
          postDrainHooks:
          - annotation: harvesterhci.io/post-hook
          preDrainHooks:
          - annotation: harvesterhci.io/pre-hook
          timeout: 0
```

In the middle of the upgrade, we found node condoning, hooks, and RKE2 upgraded are performed on multiple nodes at the same time. In the environment we tested, it's `harvester-node-2` and `harvester-node-3`:

```
harvester-node-0:~ # kubectl get nodes
NAME               STATUS                     ROLES                       AGE     VERSION
harvester-node-0   Ready                      control-plane,etcd,master   8h      v1.22.4+rke2r2
[...]

PR #36142, opened 2022-01-13
fixed concurrency limit being ignored on upgrade

paynejacob:
#35999 

# Problem
When pre-drain hooks are configured rancher will ignore any concurrency limits set on node upgrade.  

# Solution
The planner was considering machines with stale plans to be unavailable and triggering a drain on them.  The node hooks cause the machine drain controller to send more updates than it would on machine without.  I think this was causing the events to be pushed to the controller with stale plans causing the nodes to be drained even when the concurrency limit had been reached.   Instead the planner checks if the node is explicitly being drained before updating it.   I don't think this problem was unique to clusters with pre drain hooks but I think the hooks caused enough churn to cause this issue reliably. 

# Testing

- single node cluster
- 3 controlplane
- 5 controlplane
- 3 worker

For all these cases I set a concurrency limit of 1 and did not observer the limit exceeded after applying my changes.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -552,11 +552,11 @@ func (p *Planner) reconcile(controlPlane *rkev1.RKEControlPlane, secret plan.Sec
 		} else if !equality.Semantic.DeepEqual(entry.Plan.Plan, plan) {
 			outOfSync = append(outOfSync, entry.Machine.Name)
 			// Conditions
-			// 1. If plan is not in sync then there is no harm in updating it to something else because
-			//    the node will have already been considered unavailable.
+			// 1. If the node is already draining then the plan is out of sync.  There is no harm in updating it if
+			// the node is currently drained.
 			// 2. concurrency == 0 which means infinite concurrency.
 			// 3. unavailable < concurrency meaning we have capacity to make something unavailable
-			if isUnavailable(entry) || concurrency == 0 || unavailable < concurrency {
+			if isInDrain(entry.Machine) || concurrency == 0 || unavailable < concurrency {
 				if !isUnavailable(entry) {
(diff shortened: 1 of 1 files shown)


jakefhyde (approved):

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -552,11 +552,11 @@ func (p *Planner) reconcile(controlPlane *rkev1.RKEControlPlane, secret plan.Sec
...
-			// 1. If plan is not in sync then there is no harm in updating it to something else because
-			//    the node will have already been considered unavailable.
+			// 1. If the node is already draining then the plan is out of sync.  There is no harm in updating it if
+			// the node is currently drained.
 			// 2. concurrency == 0 which means infinite concurrency.
 			// 3. unavailable < concurrency meaning we have capacity to make something unavailable
-			if isUnavailable(entry) || concurrency == 0 || unavailable < concurrency {
+			if isInDrain(entry.Machine) || concurrency == 0 || unavailable < concurrency {
 				if !isUnavailable(entry) {
 					unavailable++
```
jiaqiluo, line 561: In the case that we enter the clause because concurrency == 0 or unavailable < concurrency,  the current entry is available, so should we decrease the count of unavailable? 
Or do I misunderstand the logic here? 

```suggestion
				if !isUnavailable(entry) {
					unavailable--
```

jiaqiluo (changes requested):

pkg/provisioningv2/rke2/planner/planner.go
paynejacob: That is for when the plan is out of sync but the node is already draining.  In that case it was already added to our unavailable count we just want to refresh the plan.

jiaqiluo (approved):

PR #36143, opened 2022-01-13
Run rancher more reliably during testing

Oats87:
On CI runs, K3s was continuously failing with a `time="2022-01-13T22:57:27.410144766Z" level=fatal msg="starting kubernetes: preparing server: init cluster datastore and https: listen tcp :6443: bind: address already in use"`
or something similar. 

In my debugging of this, I have a theory that the `keepalive` loop monitoring Rancher health is not being terminated and is causing problems. Even on a graceful cleanup of Rancher, the keepalive loop is attempting to start a new Rancher, which is then providing conflicting (and unmanaged) rancher instances. 

This PR changes the method of running Rancher for testing.

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -4,10 +4,14 @@ set -e
 cleanup()
 {
     EXIT=$?
-    set +e
+    set +ex
     echo Stopping rancher server
-    kill $PID
-    wait $PID
+    kill $RANCHER_RUN_PID
+    wait $RANCHER_RUN_PID
+    if [ $PID != -1 ]; then
+      kill $PID
+      wait $PID
+    fi
     return $EXIT
 }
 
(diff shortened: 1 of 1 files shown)


rosskirkpat (commented):
LGTM assuming that CI passes. We should probably restart a green CI for this PR a few times to ensure that we don't have another testing issue that needs addressed.

scripts/test
```diff
@@ -20,20 +24,29 @@ if [ ${ARCH} == arm64 ]; then
...
+              echo Rancher died after tests started, aborting
+              exit 1
+            fi
+            PID=-1
+            sleep 5
+        fi
+        if [ "$PID" = "-1" ]; then
+          echo Starting rancher server using run
+          ./scripts/run >/tmp/rancher.log 2>&1 &
+          PID=$!
```
jiaqiluo, line 49: Is it possible to be trapped in the loop of "restarting Rancher -> process dies before starting tests -> restarting rancher -> process dies before starting tests --> restarting Rancher ..." 

If so, we may force it to exit & fail if a max retry count is hit.

jiaqiluo (changes requested):

scripts/test
Oats87: If that is the case, Drone will eventually kill off the job after a timeout interval is hit

I didn't want to artificially limit the amount of startup attempts. That would maybe be a good idea for a future enhancement though

jiaqiluo: Oh, nice. It is good to know that Drone will handle it.

jiaqiluo (approved):
LGTM 💯

luthermonson (approved):

Oats87:
Merging as I had 3 consecutive PR runs:

https://drone-pr.rancher.io/rancher/rancher/20340
https://drone-pr.rancher.io/rancher/rancher/20342
https://drone-pr.rancher.io/rancher/rancher/20344

PR #36167, opened 2022-01-16
Rancher Helm Chart: also add traefik annotations

### LINKED ISSUE #36166: Helm Chart: Also add ingress annotations for traefik, not only for nginx-ingress
**Is your feature request related to a problem? Please describe.**

I installed k3s without traefik.
I then install traefik via helm, by just installing the chart without any values:
```
helm install traefik traefik/traefik -n traefik-system
```

After installation of Rancher via helm I only get a 404 error and the default backend of traefik when connecting via HTTPS.

I found two possible solutions:

The first is to uninstall traefik and install nginx-ingress. Installing without any values is enough, to make this work out of the box. 

The second is to re-install traefik with a values.yaml that contains the following snippet
```
ports:
  websecure:
    tls:
      enabled: true

```

And suddenly everything works.

A third possibility would be to patch the ingress to contain the traefik annotations that can also enable tls. 

I mentioned this to @oskapt and after some digging we agreed that it would be nice if the Rancher helm chart could also add the traefik annotations to the ingress. Having annotations for both nginx-ingress and traefik should not harm, because either one should ignore the other one's annotations AFAIK.

**Describe the solution you'd like**
Both nginx-ingress and traefik annotations get added to the ingress that is being created by the helm chart install (of course keeping the possibilities for the user to influence this through the values).

**Describe alternatives you've considered**
See description above.

**Additional context**
I am currently trying to understand the logic in the ingress template and will try to work on a PR for this.

johanneskastl:
Add traefik annotations to make ingress work with Traefik, without requiring traefik to be installed with values to enable tls

Fixes #36166

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -16,6 +16,7 @@ metadata:
 {{- end }}
 {{- if eq .Values.tls "external" }}
     nginx.ingress.kubernetes.io/ssl-redirect: "false" # turn off ssl redirect for external.
+    traefik.ingress.kubernetes.io/router.tls: "false" # turn off ssl redirect for external.
 {{- else }}
   {{- if ne .Values.ingress.tls.source "secret" }}
     {{- $certmanagerVer :=  split "." .Values.certmanager.version -}}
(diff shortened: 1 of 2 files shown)


johanneskastl:
Downloaded chart via helm, unpacked and copied the modified file into rancher/templates/.

Tested locally with `helm template --set hostname=example.org -n cattle-system ./rancher` with the three different options for TLS (rancher, secret, letsencrypt) and each time the resulting ingress looked valid (to me at least).

Also modified the tests to not error out due to the new traefik annotations.

johanneskastl:
Any news on this?

PR #36179, opened 2022-01-18
[Backport v2.6.3-patch1] Update k3s single node etcd migration code

annablender:
Address #35955

For an upgrade in k3s single node, I updated the etcd migration code to migrate old etcd data to embedded etcd from /var/lib/rancher/management-state/etcd to /var/lib/rancher/k3s/server/db/etcd, instead of from /var/lib/rancher/etcd. The /var/lib/rancher/etcd dir gets overwritten if you upgrade a Rancher container that has bind mounted volumes.

Customers were losing downstream clusters on upgrade. Now, if you upgrade single node in Docker all cluster data will be preserved, even if you just update the Rancher version of a container with bind mounted volumes.


snasovich (changes requested):
Looks like this moved many additional commits, we just need the actual fix moved.

annablender:
New PR created with correct cherry pick https://github.com/rancher/rancher/pull/36208

PR #36201, opened 2022-01-19
Remove need for LB creation in k3s tests

rancher-max:
Validated this is working locally. Will run in freeform as well and send the results to you all privately.


VestigeJ:
IMIO this LGTM
----------------------------------------------------------------------------------------------------------------------------
As I understand it the TLDR would be
```
- Step 1: Creating LBs by default is set to false
- Step 2: If set to true pass that into scripts for TF to use during cluster creation
- Step 3: Create clusters with or without LBs
- Step 4: Profit
```

mdrahman-suse:
LGTM 👍

mdrahman-suse (approved):

ShylajaDevadiga:
LGTM

bmdepesa (approved):

PR #36202, opened 2022-01-19
Prevent recurring backups from running if backup is in progress

jakefhyde:
If etcd is particularly saturated and a snapshot cannot be completed within the timeout, starting a recurring timeout while a snapshot is currently underway will cause the two snapshots to fight for the `etcd-snapshot-once` container. This PR defers creating the recurring snapshot until the user snapshot finishes (in addition to existing recurring snapshots). This PR also fixes an instance where a user can create a snapshot before the initial recurring one.


paynejacob (approved):

jiaqiluo (dismissed):
LGTM except for 1 nit.

jiaqiluo (approved):

thedadams (approved):

PR #36217, opened 2022-01-20
Publish to ECR registry on release

oxr463:
This will publish to ECR in addition to Docker Hub.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -234,6 +234,40 @@ steps:
     event:
     - tag
 
+- name: docker-mirror-tag
+  image: rancher/dapper:v0.5.4
+  environment:
+    AWS_REGISTRY_ID:
+      from_secret: aws_registry_id
+  commands:
+  - "docker tag rancher/rancher:${DRONE_TAG} ${AWS_REGISTRY_ID}.dkr.ecr.us-east-1.amazonaws.com/suse/rancher:${DRONE_TAG}"
+  privileged: true
+  volumes:
+  - name: docker
+    path: /var/run/docker.sock
+  when:
+    ref:
+      include:
+      - "refs/tags/v*"
+    event:
+    - tag
+
+- name: docker-publish-mirror-ecr
+  image: plugins/ecr
+  environment:
+    AWS_REGISTRY_ID:
+      from_secret: aws_registry_id
+  settings:
+    access_key:
+      from_secret: ecr_access_key
+    secret_key:
+      from_secret: ecr_secret_key
+    repo: suse/rancher
+    registry: "${AWS_REGISTRY_ID}.dkr.ecr.us-east-1.amazonaws.com"
+  when:
+    event:
+    - tag
+
 - name: docker-publish-installer
   image: plugins/docker
   settings:


paraglade (approved):
this looks good to me and should be good enough to give a try

oxr463:
We're still waiting on the AWS credentials.

oxr463:
@cbron any reason why we're getting so many merge conflicts? I'm not familiar with the branching strategy for rancher/rancher but I'll try to rebase on the master branch.

cbron:
We don't use master branch, only `release/*`, in this case release/v2.6.

oxr463:
> We don't use master branch, only `release/*`, in this case release/v2.6.

Ah, that explains it. Let me try rebasing on that.

cbron:
Not sure, likely due to the changing of branches. Try cherry-picking commit on fresh branch off release/v2.6.

oxr463:
> Not sure, likely due to the changing of branches. Try cherry-picking commit on fresh branch off release/v2.6.

Now it's even worse... :thinking:

oxr463:
I fixed the merge conflicts by rebasing, thanks to @superseb. I'm not sure why Drone PR is failing now, but this should be ready for testing.

.drone.yml
```diff
@@ -234,6 +234,37 @@ steps:
     event:
     - tag
 
+- name: docker-tag
```
cbron, line 237: Lets name this `docker-mirror-tag` or something, to represent that its different.

.drone.yml
```diff
@@ -234,6 +234,37 @@ steps:
...
+  settings:
+    access_key:
+      from_secret: ecr_access_key
+    secret_key:
+      from_secret: ecr_secret_key
+    repo: suse/rancher
+    registry: "${AWS_REGISTRY_ID}.dkr.ecr.us-east-1.amazonaws.com"
+  when:
+    event:
+    - tag
```
cbron, line 266: We don't want to do this on every tag, just core ones. Meaning no pre-releases, rc's, etc..

.drone.yml
```diff
@@ -234,6 +234,37 @@ steps:
...
+  - "docker tag rancher/rancher:${DRONE_TAG} ${AWS_REGISTRY_ID}.dkr.ecr.us-east-1.amazonaws.com/suse/rancher:${DRONE_TAG}"
+  privileged: true
+  volumes:
+  - name: docker
+    path: /var/run/docker.sock
+  when:
+    event:
+    - tag
+
+- name: docker-publish-ecr
```
cbron, line 252: How about 

```suggestion
- name: docker-publish-mirror-ecr
```

.drone.yml
```diff
@@ -234,6 +234,37 @@ steps:
     event:
     - tag
 
+- name: docker-tag
+  image: rancher/dapper:v0.5.4
+  environment:
+    AWS_REGISTRY_ID:
+      from_secret: aws_registry_id
```
cbron, line 241: Is this necessary to tag ?

.drone.yml
oxr463: How can that be expressed by Drone events?

.drone.yml
oxr463: What do you mean?

.drone.yml
oxr463: This is done.

.drone.yml
oxr463: This is done.

.drone.yml
oxr463: Would this work?

```
  when:
    ref:
      include:
        - "refs/heads/release/v*"
    event:
      - tag
```

cbron: No because we tag everything off the `release/v2.6` branch. The tag `v2.6.3` and `v2.6.3-rc1` are done the same way. You'd have to filter on the tag name.

ryansann: `refs/tags/v*` may work, though this will include `-rcX` tags. @cbron we may need RCs for testing initially, right?

ryansann: if we want to exclude `-rcX` tags, we can use something like:
```
include:
- "refs/tags/v*"
exclude:
- "refs/tags/*-*"
```
(from another rule in `.drone.yml`)

oxr463: I can add that, but how will we test it? Wait for the next official release?

oxr463: @cbron @ryansann any ideas?

cbron: You could leave the excludes out to test an rc, then delete it, as long as you can delete images. Or just wait for final tag and if that doesn't work manually mirror.

cbron:
Due to other complications this feature has been moved back to 2.6.5.

cbron:
This was actually removed from requirements.

Issue #36062, opened 2022-01-07
RKE2 | Worker node is not ready after migrating to a new cluster


sgapanovich:
**Rancher Server Setup**
- Rancher version: 2.6.3
- Installation option (Docker install/Helm Chart): helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): rke1 ha

**Information about the Cluster**
- Kubernetes version: v1.21.7+rke2r1
- Cluster Type (Local/Downstream): downsream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Dowstream RKE2 ec2 8 nodes cluster (3 etcd, 2 cp, 3 workers)

   
**Describe the bug**

After migrating Rancher from docker single node to RKE1 HA:
- cluster is stuck in "Provisioning" status
- worker node is stuck in "Nodeconditionsfailed" status

**To Reproduce**
 
Was discovered while testing https://github.com/rancher/rancher/issues/36057
Issue was found when migrating Rancher from docker install to an HA cluster
1. Create Rancher single node install rancher
2. Provision a RKE2 downstream cluster 
  k8s version: v1.21.7+rke2r1
  setup: 3 etcd, 2 cp, 3 workers
3. Following back up documentation create a backup https://rancher.com/docs/rancher/v2.6/en/backups/back-up-rancher/
4. Create a RKE1 3 node all roles cluster (future rancher server cluster)
5. Following migration documentation migrate the rancher server to the cluster created in step 4. https://rancher.com/docs/rancher/v2.6/en/backups/migrating-rancher/
6. Log in to Rancher server 

**Result**
One worker node is not ready 
Couldn't find any related errors in rancher logs or cattle-cluster-agent

**Expected Result**

RKE2 cluster has all nodes Active after migration

**Screenshots**

![rke2](https://user-images.githubusercontent.com/40282141/148482891-58d29888-d63f-4703-9897-beacce11b769.jpg)

snasovich:
Per @thedadams it's likely the same root cause as https://github.com/rancher/rancher/issues/35822. Should be retested once that issue is fixed.

thedadams:
### Root cause
The root cause of this issue was a race condition in the controllers for Rancher and for CAPI.

### What was fixed, or what changes have occurred
The caches used in CAPI for the Rancher objects were started if the weren't already started.

### Areas or cases that should be tested
RKE2/K3S provisioning:
[...]

Issue #35822, opened 2021-12-09
`v1.21.6+rke2r1` RKE2 cluster is stuck in provisioning state on a Rancher server upgrade when a worker node is stuck in `Nodeconditionsfailed` state


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6.2 to 2.6-head commit id: `77392a6`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: `v1.21.6+rke2r1` RKE2 cluster EC2 node driver
- Cluster Type (Local/Downstream): Downstream

**Describe the bug**
`v1.21.6+rke2r1` RKE2 cluster is stuck in provisioning state on a Rancher server upgrade

**To Reproduce**

- Deploy a 2 node RKE2 cluster - 1 all roles, 2 worker nodes
- Upgrade rancher from 2.6.2 to 2.6-head
- Cluster goes into provisioning state  - logged bug - https://github.com/rancher/rancher/issues/35821
- Worker node is stuck in `Nodeconditionsfailed` state
- And the cluster is stuck in provisioning state
<img width="1452" alt="Screen Shot 2021-12-08 at 9 42 37 PM" src="https://user-images.githubusercontent.com/26032343/145341398-7b16fb33-1609-4c5e-bcce-17fcdf2035f3.png">

**Expected Result:**
On an Upgrade, worker node should not get stuck in `Nodeconditionsfailed` state

Maybe related to: https://github.com/rancher/rancher/issues/35714

**Note:**
Cluster seems to have recovered in ~30 mins

thedadams:
### Root cause
The root cause of this issue was a race condition in the controllers for Rancher and for CAPI.

### What was fixed, or what changes have occurred
The caches used in CAPI for the Rancher objects were started if the weren't already started.

### Areas or cases that should be tested
RKE2/K3S provisioning:
1. Single node all roles.
2. 3 nodes, one role each
3. 1 control plane, 1 etcd and multiple worker nodes

### What areas could experience regressions?
Provisioning and editing

### Are the repro steps accurate/minimal?
Yes.

sowmyav27:
**On 2.6-head commit id: `f904267`**

RKE2 provisioning and edit usecases for the following have been tested
- Single node all roles.
- 3 nodes, one role each
- 1 control plane, 1 etcd and multiple worker nodes

thedadams:
Issue was reopened by zube bot.

Issue #35440, opened 2021-11-10
Rancher server logs error syncing machine-worker-label on imported harvester 0.3.0 cluster


SweBarre:
**Rancher Server Setup**
- Rancher version: v2.6.2
- Installation option (Docker install/Helm Chart): docker install
- Proxy/Cert Details: self signed

**Information about the Cluster**
- Kubernetes version: v1.21.5+rke2r1
- Cluster Type (Local/Downstream): Downstream, harvester v0.3.0
 
**Describe the bug**
The rancher server repeatedly logs the following when importing a newly imported harvester v0.3.0 cluster
```
2021/11/10 08:40:57 [ERROR] error syncing 'node2': handler machine-worker-label: machines.cluster.x-k8s.io "custom-9cebcaf1ae85" not found, requeuing
2021/11/10 08:41:08 [ERROR] error syncing 'node1': handler machine-worker-label: machines.cluster.x-k8s.io "custom-e0a3c84cc108" not found, requeuing
2021/11/10 08:41:17 [ERROR] error syncing 'node3': handler machine-worker-label: machines.cluster.x-k8s.io "custom-752a28e09de7" not found, requeuing
```

**To Reproduce**
Import a harvester v0.3.0 cluster into rancher server

**Additional context**
Rancher server stopped logging the error when labeling the harvester nodes as worker nodes with:
`kubectl label node <nodename> node-role.kubernetes.io/worker=true`

snasovich:
The issue is in https://github.com/rancher/rancher/blame/6ae91f6ea7dee6f54e49291313ac8bb3c7937900/pkg/controllers/managementuser/machinerole/machinerole.go controller. Need to add some checks to run/not run this controller depending on cluster type (possibly not start the controller for Harvester clusters?). Need to see how Harvester clusters are showing up in UI to ensure https://github.com/rancher/rancher/issues/33385 is not applicable here.
@slickwarren has a setup to test.

erSitzt:
Same errors here with Rancher 2.6.3 and Harvester 1.0.0
```
 [ERROR] error syncing 'harvester-test-1': handler machine-worker-label: machines.cluster.x-k8s.io "custom-5345e8d0fea8" not found, requeuing
```

thedadams:
### Root cause
The `machine-worker-label` controller was not checking for a CAPI machine for these custom nodes.

### What was fixed, or what changes have occurred
[...]

PR #36218, opened 2022-01-21
Don't update machines

thedadams:
Rancher was adding labels and annotations to the CAPI machines to track things 
like init nodes. The issue is that this caused the CAPI controllers to look at 
the machines much more often than necessary. In addition, because it is 
difficult for the Wrangler controllers to re-enqueue CAPI machines to the CAPI 
controllers, it created timing/race conditions in certain scenarios. Finally, 
the labels and annotations added by Rancher were for the dynamic planner in 
Rancher, which is not CAPI related.

After this change, the Rancher-specific labels and annotations are moved to the 
machine plan secret. This is a more appropriate place for these items because 
CAPI doesn't know about or care about the machine plan secret.

This change also bumps lasso to fix a problem with the caches used by the CAPI 
controllers. This change allows us to remove the machinestatus controller
because it is no longer necessary. (It was used as a way to re-enqueue CAPI 
machines to the CAPI controllers, but the bump of lasso fixes the underlying 
problem).

Lastly, common items needed by the various RKE2 controllers are moved to a 
central location where they can all be referenced.

Issues:
https://github.com/rancher/rancher/issues/35822
https://github.com/rancher/rancher/issues/36062
https://github.com/rancher/rancher/issues/35440


thedadams:
For reviewers:
I tried to break this up into multiple commits so it is easier to review. I messed this up and the "Move common rke2 items to a common package" commit ended up with two commits squashed. My apologies.

Also, I still need to write the migrations for this, but that is completely separate and will not affect anything of these changes. I wanted to get this up because it is a big change.

Lastly, there is not guarantee that the tests will pass here. I did my best to update them as I went.

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -0,0 +1,163 @@
+package rke2
```
Oats87, line 1: Awesome

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
```diff
@@ -59,16 +55,20 @@ func (h *handler) UpdateSpec(_ string, cluster *v1.RKECluster) (*v1.RKECluster,
 }
 
 func (h *handler) OnChange(obj *v1.RKECluster, status v1.RKEClusterStatus) (v1.RKEClusterStatus, error) {
+	condition := rke2.Ready
```
Oats87, line 58: nit: this variable name is vague and potentially confusing without an IDE

pkg/provisioningv2/rke2/configserver/server.go
```diff
@@ -182,10 +179,16 @@ func (r *RKE2ConfigServer) connectConfigYaml(name, ns string, rw http.ResponseWr
...
+	if err != nil {
+		http.Error(rw, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
 	var content string
 	for _, f := range config["files"].([]interface{}) {
 		f := f.(map[string]interface{})
-		if path, ok := f["path"].(string); ok && path == fmt.Sprintf(planner.ConfigYamlFileName, "rke2") {
+		if path, ok := f["path"].(string); ok && path == fmt.Sprintf(planner.ConfigYamlFileName, rke2.GetRuntime(kubernetesVersion)) {
```
Oats87, line 191: I have a feeling the `rke2` was intentional as we do not support Windows with K3s and `configYaml` is for Windows to my knowledge.

@luthermonson may have more insight

tests/integration/pkg/tests/machineprovisioning/machine_test.go
```diff
@@ -330,20 +331,28 @@ func TestDrain(t *testing.T) {
...
 	var doneHooks int32
 	runHooks := func(machine *capi.Machine) error {
-		return wait.Object(clients.Ctx, clients.CAPI.Machine().Watch, machine, func(obj runtime.Object) (bool, error) {
-			machine := obj.(*capi.Machine)
-			if machine.Annotations[planner.PreDrainAnnotation] != "" &&
-				machine.Annotations[planner.PreDrainAnnotation] != machine.Annotations["test.io/pre-hook1"] {
-				machine.Annotations["test.io/pre-hook1"] = machine.Annotations[planner.PreDrainAnnotation]
-				machine.Annotations["test.io/pre-hook2"] = machine.Annotations[planner.PreDrainAnnotation]
-				_, err := clients.CAPI.Machine().Update(machine)
+		bootstrap, err := clients.RKE.RKEBootstrap().Get(machine.Spec.Bootstrap.ConfigRef.Namespace, machine.Spec.Bootstrap.ConfigRef.Name, metav1.GetOptions{})
```
Oats87, line 334: Can we try to capture transient errors like 500s or 502/503s on these Gets?

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -199,24 +191,11 @@ func (h *handler) assignPlanSecret(machine *capi.Machine, obj *rkev1.RKEBootstra
...
-
-		return h.machineCache.Get(obj.Namespace, ref.Name)
-	}
-	return nil, generic.ErrSkip
+	return []runtime.Object{sa, secret, role, rolebinding}
 }
 
-func (h *handler) getEnvVar(machine *capi.Machine) (result []corev1.EnvVar, _ error) {
-	capiCluster, err := h.capiClusters.Get(machine.Namespace, machine.Spec.ClusterName)
+func (h *handler) getEnvVar(obj *rkev1.RKEBootstrap) (result []corev1.EnvVar, _ error) {
```
Oats87, line 197: very minor nit, but `obj` may be better declared with a more descriptive name for the common passerby

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -0,0 +1,163 @@
...
+
+var (
+	ErrNoMachineOwnerRef = errors.New("no machine owner ref")
+)
+
+func MachineStateSecretName(machineName string) string {
+	return name.SafeConcatName(machineName, "machine", "state")
+}
+
+func GetMachineOwner(machineCache capicontrollers.MachineCache, obj metav1.Object) (*capi.Machine, error) {
```
Oats87, line 72: This function name is confusing to me -- it appears that it gets a machine that owns the passed in `obj`, but when I read the function name I assume it is getting the owner of the machine.

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -45,6 +46,10 @@ func getInfraRef(rkeCluster *rkev1.RKECluster) *corev1.ObjectReference {
 // objects generates the corresponding rkecontrolplanes.rke.cattle.io, clusters.cluster.x-k8s.io, and
 // machinedeployments.cluster.x-k8s.io objects based on the passed in clusters.provisioning.cattle.io object
 func objects(cluster *rancherv1.Cluster, dynamic *dynamic.Controller, dynamicSchema mgmtcontroller.DynamicSchemaCache, secrets v1.SecretCache) (result []runtime.Object, _ error) {
+	if !cluster.DeletionTimestamp.IsZero() {
+		return nil, nil
```
Oats87, line 50: Not really a problem but I feel like this can be short-circuited further upwards. Probably good to have this check in here anyway

pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
```diff
@@ -52,3 +68,55 @@ func (h *handler) OnChange(obj *rkev1.RKEControlPlane, status rkev1.RKEControlPl
 	status.Initialized = Ready.IsTrue(cluster)
 	return status, nil
 }
+
+func (h *handler) updateControlPlaneStatus(cp *rkev1.RKEControlPlane, status rkev1.RKEControlPlaneStatus, previousErr error) (*rkev1.RKEControlPlane, error) {
```
Oats87, line 72: I'm not really sure this needs to be its own function, can we not just move this logic to the `OnRemove` below?

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -22,44 +22,51 @@ import (
...
-		ctx:      ctx,
-		machines: clients.CAPI.Machine(),
-		secrets:  clients.Core.Secret().Cache(),
+		ctx:          ctx,
+		machineCache: clients.CAPI.Machine().Cache(),
+		secrets:      clients.Core.Secret(),
+		secretCache:  clients.Core.Secret().Cache(),
 	}
-	clients.CAPI.Machine().OnChange(ctx, "machine-drain", h.OnChange)
+	clients.Core.Secret().OnChange(ctx, "machine-drain", h.OnChange)
```
Oats87, line 38: We may want to generate the core client/secret thing with a selective indexer to keep down on the noise

pkg/provisioningv2/rke2/planner/drain.go
```diff
@@ -78,28 +70,28 @@ func (p *Planner) drain(oldPlan *plan.NodePlan, newPlan plan.NodePlan, machine *
...
 
-	return machine.Annotations[DrainDoneAnnotation] == optionString, nil
+	return entry.Secret.Annotations[rke2.DrainDoneAnnotation] == optionString, nil
 }
 
-func (p *Planner) setMachineAnnotation(machine *capi.Machine, key, value string) (bool, error) {
-	machine = machine.DeepCopy()
-	if machine.Annotations == nil {
-		machine.Annotations = map[string]string{}
+func (p *Planner) setSecretAnnotation(entry *planEntry, key, value string) (bool, error) {
```
Oats87, line 80: lol, this function is hidden but generic

Oats87 (changes requested):

pkg/controllers/provisioningv2/cluster/remove.go
```diff
@@ -57,8 +43,6 @@ func (h *handler) updateClusterStatus(cluster *v1.Cluster, status v1.ClusterStat
 	if equality.Semantic.DeepEqual(status, cluster.Status) {
 		return cluster, previousErr
 	}
-	cluster = cluster.DeepCopy()
-	cluster.Status = status
 	cluster, err := h.clusters.UpdateStatus(cluster)
```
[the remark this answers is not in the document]
thedadams: I changed the logic. Previously, the changes to the cluster's status were made to `status`. So, if a change was detected, then the `cluster.Status` needed to be updated for the `UpdateStatus` call.

Now, the changes are made to `cluster.Status` and `status` is the "old" status. So, we don't need to update `cluster.Status` because it is the new status.

pkg/controllers/provisioningv2/cluster/remove.go
```diff
@@ -67,96 +51,88 @@ func (h *handler) updateClusterStatus(cluster *v1.Cluster, status v1.ClusterStat
...
-		Removed.Message(status, message)
-		h.clusters.EnqueueAfter(cluster.Namespace, cluster.Name, 5*time.Second)
-		// generic.ErrSkip will mark the cluster as reconciled, but not remove the finalizer.
-		// The finalizer shouldn't be removed until other objects have all been removed.
-		err = generic.ErrSkip
-	}
-	return h.updateClusterStatus(cluster, *status, err)
+	status := cluster.Status
+	cluster = cluster.DeepCopy()
+	return h.updateClusterStatus(cluster, status, rke2.DoRemoveAndUpdateStatus(cluster, h.doClusterRemove(cluster), h.clusters.EnqueueAfter))
```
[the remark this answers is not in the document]
thedadams: `rke2.DoRemoveAndUpdateStatus` doesn't make the call to actually commit the updated status to Kubernetes. That is what `updateClusterStatus` does, if necessary.

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
thedadams: Will update.

pkg/provisioningv2/rke2/configserver/server.go
thedadams: Two thoughts here:
1. Does it matter?
2. I think this will be going away in 2.6.4 because Windows nodes will use the planner.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
thedadams: I was just making it consistent with the rest of the functions in this file. They all call the bootstrap object `obj`.

I will update them all to be more descriptive.

pkg/controllers/provisioningv2/rke2/common.go
thedadams: Will update.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
thedadams: OK. I have a solution for your review.

thedadams:
Force push was a rebase to fix merge conflicts.

pkg/apis/rke.cattle.io/v1/plan/plan.go
```diff
@@ -1,13 +1,15 @@
...
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
 type Plan struct {
-	Nodes    map[string]*Node         `json:"nodes,omitempty"`
-	Machines map[string]*capi.Machine `json:"machines,omitempty"`
-	Cluster  *capi.Cluster            `json:"cluster,omitempty"`
+	Nodes    map[string]*Node          `json:"nodes,omitempty"`
+	Machines map[string]*capi.Machine  `json:"machines,omitempty"`
+	Secrets  map[string]*corev1.Secret `json:"secrets,omitempty"`
```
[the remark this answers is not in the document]
thedadams: After the secret is stored here, we really only care about the labels and annotations on it for the planner. The plan data is in the `Node` struct.

pkg/controllers/provisioningv2/cluster/remove.go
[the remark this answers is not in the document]
Oats87: OK I see what you're doing here. 

What I meant by this, was is it necessary to have an independent function like `updateClusterStatus` instead of just directly calling `h.clusters.UpdateStatus(cluster)` directly if the `UpdateStatus` does not push an update based based on the diff of the objects (so basically, if it also does the equality.Semantic.DeepEqual`

this function confuses me because the passed in `status v1.ClusterStatus` may be different than the `cluster.Status` which would call a `h.clusters.UpdateStatus(cluster)` so it relies on the caller basically call it like `updateClusterStatus(cluster, cluster.status, someErr)`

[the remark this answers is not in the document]
Oats87: I think we can check the server and see if we need to update the status (which I thought `UpdateStatus` would do)

Right now the way this is written the caller has to store a copy of the old status to pass in to see if the status needs to be updated. I think it's a strange/confusing abstraction

pkg/controllers/provisioningv2/cluster/remove.go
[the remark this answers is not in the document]
Oats87: With clarity given here I think I can see what the goal is, i.e. store a copy of the old `Status` and do the remove and update and then update the `cluster.Status` if it differs from the "old" status. This is just not very clear to me

pkg/apis/rke.cattle.io/v1/plan/plan.go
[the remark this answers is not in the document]
Oats87: If that's the case, then should we just store the `labels` and `annotations` in a map of secret names to a struct containing them then? I'm not sure we need to be dragging around all of the secret data around as it can get quite large

pkg/provisioningv2/rke2/configserver/server.go
Oats87: 1. Nah, probably doesn't matter. 
2. Interesting, but makes sense

tests/integration/pkg/tests/machineprovisioning/machine_test.go
Oats87: I can see a situation where Rancher dies during a test and causes retrieving the bootstrap object to fail with a transient error

Oats87 (commented):
Assuming none of the force-pushed files have changed, this overall looks good to me. A few nits here and there and I continued the conversation on a few points.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -140,36 +133,35 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 	return nil, nil
 }
 
-func (h *handler) assignPlanSecret(machine *capi.Machine, obj *rkev1.RKEBootstrap) ([]runtime.Object, error) {
-	secretName := planner.PlanSecretFromBootstrapName(obj.Name)
+func (h *handler) assignPlanSecret(machine *capi.Machine, boostrap *rkev1.RKEBootstrap) []runtime.Object {
```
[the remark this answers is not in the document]
thedadams: Fixed.

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/rancher/wrangler/pkg/generic"
+	"github.com/rancher/wrangler/pkg/name"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	capi "sigs.k8s.io/cluster-api/api/v1beta1"
+	capierrors "sigs.k8s.io/cluster-api/errors"
+)
+
+const (
+	AddressAnnotation = "rke.cattle.io/address"
```
paynejacob, line 21: ❤️

pkg/rancher/migrations.go
```diff
@@ -28,13 +34,52 @@ const (
...
+		}
+	}
+
+	if features.EmbeddedClusterAPI.Enabled() {
+		if err := addWebhookConfigToCAPICRDs(wranglerContext.CRD.CustomResourceDefinition()); err != nil {
+			return err
+		}
+	}
+
+	if features.RKE2.Enabled() {
```
paynejacob, line 67: given this traverses all the management clusters do we want to gate this behind a version check?

thedadams: There is a configmap check in `migrateCAPIMachineLabelsAndAnnotationsToPlanSecret` to ensure this migration only happens once.

Oats87 (commented):
Overall, LGTM

One small nit which is around the struct at the top

Oats87 (approved):
LGTM; pending CI passing

paynejacob (approved):

paynejacob (dismissed):

paynejacob (approved):

Oats87 (approved):

PR #36219, opened 2022-01-21
refactoring provisioning v1 tests to simplify future node drivers/providers

slickwarren:
additional config in go:
```
provisioningInput:
  providers: ["aws", "harvester", "do", ...]
```

this will be a required field in the config file. Tests will provision for each provider listed. Currently supported are aws, do, and harvester.

### DIFF
--- a/tests/v2/validation/provisioning/providers.go
+++ b/tests/v2/validation/provisioning/providers.go
@@ -1,0 +1,63 @@
+package provisioning
+
+import (
+	"fmt"
+
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials"
+	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials/aws"
+	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials/digitalocean"
+	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials/harvester"
+	"github.com/rancher/rancher/tests/framework/extensions/machinepools"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	awsProviderName       = "aws"
+	doProviderName        = "do"
+	harvesterProviderName = "harvester"
+)
+
+type CloudCredFunc func(rancherClient *rancher.Client) (*cloudcredentials.CloudCredential, error)
+type MachinePoolFunc func(generatedPoolName, namespace string) *unstructured.Unstructured
+
+type Provider struct {
+	Name            string
+	MachineConfig   string
+	MachinePoolFunc MachinePoolFunc
+	CloudCredFunc   CloudCredFunc
+}
+
+// CreateProvider returns all machine and cloud credential
+// configs in the form of a Provider struct. Accepts a
+// string of the name of the provider
+func CreateProvider(name string) Provider {
+	switch {
+	case name == awsProviderName:
+		provider := Provider{
+			Name:            name,
+			MachineConfig:   machinepools.AWSResourceConfig,
+			MachinePoolFunc: machinepools.NewAWSMachineConfig,
+			CloudCredFunc:   aws.CreateAWSCloudCredentials,
+		}
+		return provider
+	case name == doProviderName:
+		provider := Provider{
+			Name:            name,
+			MachineConfig:   machinepools.DOResourceConfig,
+			MachinePoolFunc: machinepools.NewDigitalOceanMachineConfig,
+			CloudCredFunc:   digitalocean.CreateDigitalOceanCloudCredentials,
+		}
+		return provider
+	case name == harvesterProviderName:
+		provider := Provider{
+			Name:            name,
+			MachineConfig:   machinepools.HarvesterResourceConfig,
+			MachinePoolFunc: machinepools.NewHarvesterMachineConfig,
+			CloudCredFunc:   harvester.CreateHarvesterCloudCredentials,
+		}
+		return provider
+	default:
+		panic(fmt.Sprintf("Provider:%v not found", name))
+	}
+}
--- a/tests/v2/validation/provisioning/provisioning_node_driver_test.go
+++ b/tests/v2/validation/provisioning/provisioning_node_driver_test.go
@@ -7,9 +7,6 @@ import (
 
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
 	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
-	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials/aws"
-	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials/digitalocean"
-	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials/harvester"
 	"github.com/rancher/rancher/tests/framework/extensions/clusters"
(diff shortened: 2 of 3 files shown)


tests/v2/validation/provisioning/providers.go
```diff
@@ -0,0 +1,59 @@
...
+type MachinePoolFunc func(generatedPoolName, namespace string) *unstructured.Unstructured
+
+type Provider struct {
+	Name            string
+	MachineConfig   string
+	MachinePoolFunc MachinePoolFunc
+	CloudCredFunc   CloudCredFunc
+}
+
+func CreateProvider(name string) Provider {
```
igomez06, line 23: Add a comment above this function explaining what it's used for. In the format of 
`// CreateProvider ...  ` I want to keep all of our exported functions well documented.

tests/v2/validation/provisioning/providers.go
```diff
@@ -0,0 +1,59 @@
...
+		provider := Provider{
+			Name:            name,
+			MachineConfig:   machinepools.DOResourceConfig,
+			MachinePoolFunc: machinepools.NewDigitalOceanMachineConfig,
+			CloudCredFunc:   digitalocean.CreateDigitalOceanCloudCredentials,
+		}
+		return provider
+	//TODO: if no other node driver is found, we should default to custom cluster
+	default:
+		provider := Provider{
```
igomez06, line 51: I disagree, I think the default should be a `panic()` along the lines of a type not existing. 
Something like this maybe:
```
default:
            fmt.Printf("Provider:%s not found", name)
```

tests/v2/validation/provisioning/providers.go
```diff
@@ -0,0 +1,59 @@
...
+type Provider struct {
+	Name            string
+	MachineConfig   string
+	MachinePoolFunc MachinePoolFunc
+	CloudCredFunc   CloudCredFunc
+}
+
+func CreateProvider(name string) Provider {
+	switch {
+	case name == "harvester":
```
igomez06, line 25: I think we should have const() at the top of the this file, with the provider names.
```
const (
  harvesterProviderName = "harvester"
  doProviderName = "do"
  ...
)
```
What are your thoughts?

tests/v2/validation/provisioning/provisioning_node_driver_test.go
```diff
@@ -303,11 +140,11 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2AWSCluster()
 					testSessionClient, err := tt.client.WithSession(testSession)
 					require.NoError(r.T(), err)
 
-					clusterName := AppendRandomString(baseAWSClusterName)
+					clusterName := AppendRandomString("auto-" + provider.Name + "-")
```
igomez06, line 143: I think it would look cleaner if the `"auto-"+ provider.Name +"-"` all happened in `AppendRandomString`, I don't see a problem with that. What do you think?

tests/v2/validation/provisioning/provisioning_node_driver_test.go
```diff
@@ -335,8 +172,10 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2AWSCluster()
 	}
 }
 
-func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2AWSClusterDynamicInput() {
+func (r *RKE2NodeDriverProvisioningTestSuite) Provisioning_RKE2ClusterDynamicInput(provider Provider) {
+
 	nodesAndRoles := NodesAndRolesInput()
+	fmt.Printf("%v", nodesAndRoles)
```
igomez06, line 178: Either remove this print statement or make it a a log statement. Something like this ` r.T().Logf("Node Config: %v", nodesAndRoles) `

tests/v2/validation/provisioning/provisioning_node_driver_test.go
```diff
@@ -403,163 +242,25 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2AWSClusterDyn
...
-					err = wait.WatchWait(result, checkFunc)
-					assert.NoError(r.T(), err)
-					assert.Equal(r.T(), clusterName, clusterResp.Name)
-
-				})
-			}
-		}
+func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning() {
+	for provider_name := range r.providers {
+		fmt.Printf("Starting test for %v\n", r.providers[provider_name])
```
igomez06, line 247: Make this a Log statement, See comment above.

tests/v2/validation/provisioning/provisioning_node_driver_test.go
```diff
@@ -403,163 +242,25 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2AWSClusterDyn
...
+		provider_struct := CreateProvider(r.providers[provider_name])
+		r.Provisioning_RKE2Cluster(provider_struct)
 	}
 }
 
-func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2HarvesterClusterDynamicInput() {
+func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioningDynamicInput() {
 	nodesAndRoles := NodesAndRolesInput()
+	fmt.Printf("%v", nodesAndRoles)
 	if len(nodesAndRoles) == 0 {
```
igomez06, line 256: Why do you have this twice. Do it either here, or in `Provisioning_RKE2ClusterDynamicInput`. If you do it here, then maybe `Provisioning_RKE2ClusterDynamicInput` Takes the nodeConfig Input as a parameter. Pick a place.

tests/v2/validation/provisioning/provisioning_node_driver_test.go
```diff
@@ -403,163 +242,25 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning_RKE2AWSClusterDyn
...
-
-					err = wait.WatchWait(result, checkFunc)
-					assert.NoError(r.T(), err)
-					assert.Equal(r.T(), clusterName, clusterResp.Name)
-
-				})
-			}
-		}
+	for provider_name := range r.providers {
+		fmt.Printf("Starting test for %v\n", r.providers[provider_name])
```
igomez06, line 261: Same as above make this a log statement instead. Actually maybe have all these print/log statements, might be a little much. What does the output look like for a standard test run?

igomez06 (commented):
Overall this looks really good Caleb, good job! Just address the comments, and we'll be good.

tests/v2/validation/provisioning/provisioning_node_driver_test.go
slickwarren: I've removed all of the print statements I had in for now.  I can add some back in later if we decide we need them. Maybe we could have a meeting around when is appropriate to add logging?

igomez06: Sure, about the logging. It was more of a question I'm not against having all these log statements. Just not print statements.

Other note you have 

```
nodesAndRoles := NodesAndRolesInput()
if len(nodesAndRoles) == 0 {
 	r.T().Skip()
}
```

in two spots. I would prefer to just have at the top of `TestProvisioningDynamicInput`

slickwarren: oh, I thought I'd removed that. Just updated the PR

igomez06 (approved):

bmdepesa (approved):
LGTM

PR #36220, opened 2022-01-21
Add sa token support for /metrics

rmweir:
**Problem:**
A service monitor pointing at rancher doesn't work with mounted service account token. This means a user needs to create an api token, then created a secret containing it.  Supporting the SA token will remove that process and also ensure that the used tokens expired since a rancher api token by default does not. Currently the API layer does not appear to support JWT tokens at all.

**Solution:**
Use a token review to auth tokens that have errored for /metrics path.

**Issue:**
https://github.com/rancher/rancher/issues/36234

### DIFF
--- a/pkg/auth/requests/token_review.go
+++ b/pkg/auth/requests/token_review.go
@@ -1,0 +1,59 @@
+package requests
+
+import (
+	"net/http"
+
+	"github.com/rancher/rancher/pkg/auth/tokens"
+	"github.com/rancher/steve/pkg/auth"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	authv1 "k8s.io/client-go/kubernetes/typed/authentication/v1"
+)
+
+type TokenReviewAuth struct {
+	AuthClient authv1.AuthenticationV1Interface
+}
+
+func NewTokenReviewAuth(authClient authv1.AuthenticationV1Interface) auth.Authenticator {
+	return &TokenReviewAuth{
+		AuthClient: authClient,
+	}
+}
+
+// Authenticate attempts to authenticate using given function. If authentication
+// fails AND the endpoint is in allowedPaths, it will attempt to perform a token
+// review to authenticate token and extract user info.
+func (t *TokenReviewAuth) Authenticate(req *http.Request) (user.Info, bool, error) {
+	info, hasAuth := request.UserFrom(req.Context())
+	if info.GetName() != "system:cattle:error" {
+		// auth has succeeded
+		return info, hasAuth, nil
+	}
+
+	tokenReview := &v1.TokenReview{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "authentication.k8s.io/v1",
+			Kind:       "TokenReview",
+		},
+		Spec: v1.TokenReviewSpec{
+			Token: tokens.GetTokenAuthFromRequest(req),
+		},
+	}
+
+	tokenReview, err := t.AuthClient.TokenReviews().Create(req.Context(), tokenReview, metav1.CreateOptions{})
+	if err != nil {
+		logrus.Debugf("tokenReview failed: %v", err)
+		return info, false, nil
+	}
+
+	tokenReviewUserInfo := &user.DefaultInfo{
+		Name:   tokenReview.Status.User.Username,
+		UID:    tokenReview.Status.User.UID,
+		Groups: tokenReview.Status.User.Groups,
+	}
+
+	return tokenReviewUserInfo, true, nil
+}
--- a/pkg/multiclustermanager/routes.go
+++ b/pkg/multiclustermanager/routes.go
@@ -93,8 +93,11 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	// Authenticated routes
 	authed := mux.NewRouter()
 	authed.UseEncodedPath()
-	authed.Use(mux.MiddlewareFunc(auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))))
-	authed.Use(mux.MiddlewareFunc(rbac.NewAccessControlHandler()))
+	impersonatingAuth := auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))
+	accessControlHandler := rbac.NewAccessControlHandler()
+
+	authed.Use(mux.MiddlewareFunc(impersonatingAuth))
+	authed.Use(mux.MiddlewareFunc(accessControlHandler))
 	authed.Use(requests.NewAuthenticatedFilter)
 
 	authed.Path("/meta/{resource:aks.+}").Handler(aks.NewAKSHandler(scaledContext))
@@ -102,7 +105,6 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.Path("/meta/oci/{resource}").Handler(oci.NewOCIHandler(scaledContext))
 	authed.Path("/meta/vsphere/{field}").Handler(vsphere.NewVsphereHandler(scaledContext))
 	authed.Path("/v3/tokenreview").Methods(http.MethodPost).Handler(&webhook.TokenReviewer{})
-	authed.Path("/metrics").Handler(metricsHandler)
 	authed.Path("/metrics/{clusterID}").Handler(metricsHandler)
 	authed.PathPrefix("/k8s/clusters/").Handler(k8sProxy)
 	authed.PathPrefix("/meta/proxy").Handler(metaProxy)
@@ -111,9 +113,20 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.PathPrefix("/v3/token").Handler(tokenAPI)
 	authed.PathPrefix("/v3").Handler(managementAPI)
 
+	// Metrics authenticated route
+	metricsAuthed := mux.NewRouter()
+	metricsAuthed.UseEncodedPath()
+	tokenReviewAuth := auth.ToMiddleware(requests.NewTokenReviewAuth(scaledContext.K8sClient.AuthenticationV1()))
+	metricsAuthed.Use(mux.MiddlewareFunc(tokenReviewAuth.Chain(impersonatingAuth)))
+	metricsAuthed.Use(mux.MiddlewareFunc(accessControlHandler))
+	metricsAuthed.Use(requests.NewAuthenticatedFilter)
+
+	metricsAuthed.Path("/metrics").Handler(metricsHandler)
+
 	unauthed.NotFoundHandler = authed
+	authed.NotFoundHandler = metricsAuthed
 	return func(next http.Handler) http.Handler {
-		authed.NotFoundHandler = next
+		metricsAuthed.NotFoundHandler = next
 		return unauthed
 	}, nil
 }


pkg/auth/requests/token_review.go
```diff
@@ -0,0 +1,78 @@
...
+		// This is to enable prometheus to use
+		// mounted service account tokens to
+		// fetch metrics.
+		"/metrics": true,
+	}
+)
+
+type TokenReviewAuth struct {
+	AuthClient           authv1.AuthenticationV1Interface
+	InitialAuthenticator auth.Authenticator
```
aiyengar2, line 26: I feel like the InitialAuthenticator call does not belong here. I think we should implement AuthenticatorFunc.Chain(authenticator Authenticator) [in rancher/steve](https://github.com/rancher/steve/blob/e7119828b8f9a9ca2cb2620ba650bd0508ca463e/pkg/auth/filter.go#L39-L43) or something similar to allow us to use something like 

```go
impersonateAuthenticator := requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager))
tokenAuthenticator := requests.NewTokenReviewAuth(scaledContext.K8sClient.AuthenticationV1())

authed.Use(mux.MiddlewareFunc(auth.ToMiddleware(impersonateAuthenticator.Chain(tokenAuthenticator))))
```

pkg/auth/requests/token_review.go
```diff
@@ -0,0 +1,78 @@
...
+// Authenticate attempts to authenticate using given function. If authentication
+// fails AND the endpoint is in allowedPaths, it will attempt to perform a token
+// review to authenticate token and extract user info.
+func (t *TokenReviewAuth) Authenticate(req *http.Request) (user.Info, bool, error) {
+	info, hasAuth, err := t.InitialAuthenticator.Authenticate(req)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if info.GetName() != "system:cattle:error" {
```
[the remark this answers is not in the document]
aiyengar2: Also, I don't think this logic works since we don't parse a user to `system:cattle:error` during the Authenticate call; it happens after Authenticate is called in the Middleware function. Therefore, I think the code as it stands right now would never hit this point since it would have failed out with the original error.

I think this needs to be modified to:
```suggestion
	if hasAuth {
```

pkg/auth/requests/token_review.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+		Spec: v1.TokenReviewSpec{
+			Token: tokens.GetTokenAuthFromRequest(req),
+		},
+	}
+
+	tokenReview, err = t.AuthClient.TokenReviews().Create(req.Context(), tokenReview, metav1.CreateOptions{})
+	if err != nil {
+		logrus.Debugf("tokenReview failed: %v", err)
+		return info, hasAuth, nil
```
aiyengar2, line 68:
```suggestion
		return info, false, nil
```

aiyengar2 (changes requested):

pkg/auth/requests/token_review.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	if info.GetName() != "system:cattle:error" {
+		// auth has succeeded
+		return info, hasAuth, nil
+	}
+
+	if !allowedPaths[req.URL.Path] {
```
cmurphy, line 50: Instead of checking paths here, why not just add this to only the /metrics route? https://github.com/rancher/rancher/pull/36220/files#diff-80627fe013488f5294c7e3c058e6231d1f4d622b3e7dd817bbf9db3c88af0cbbR105

pkg/auth/requests/token_review.go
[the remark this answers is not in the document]
rmweir: I think this happens because of the tokenAuthenticator that all requests go through. This code has been tested and when the token is invalid this is the name.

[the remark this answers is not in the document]
rmweir: Nothing gets set as a result of the initial Authenticate because it is not wrapped in the ToMiddleware func which is where that happens. So the info set is from a previous handler. I am changing this in my refactor to be more straight forward.

pkg/auth/requests/token_review.go
rmweir: Changed to do just this.

pkg/auth/requests/token_review.go
rmweir: I think the intended way to do this is to use toMiddleware on both and then chain them, I think I should have been doing that anyway since using that function ensure that the user info from the two functions are applied to the response in order. Fixed to do this.

pkg/auth/requests/token_review.go
```diff
@@ -0,0 +1,78 @@
...
+	authv1 "k8s.io/client-go/kubernetes/typed/authentication/v1"
+)
+
+var (
+	allowedPaths = map[string]bool{
+		// This is to enable prometheus to use
+		// mounted service account tokens to
+		// fetch metrics.
+		"/metrics": true,
+	}
```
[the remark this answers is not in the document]
rmweir: I used @cmurphy suggestion of just routing it to /metrics.

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.PathPrefix("/v3/token").Handler(tokenAPI)
 	authed.PathPrefix("/v3").Handler(managementAPI)
 
+	// Authenticated routes
+	metricsAuthed := mux.NewRouter()
+	metricsAuthed.UseEncodedPath()
+	impersonsatingAuth := auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))
```
cmurphy, line 116:
```suggestion
	impersonatingAuth := auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))
```

pkg/multiclustermanager/routes.go
```diff
@@ -102,7 +102,6 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.Path("/meta/oci/{resource}").Handler(oci.NewOCIHandler(scaledContext))
 	authed.Path("/meta/vsphere/{field}").Handler(vsphere.NewVsphereHandler(scaledContext))
 	authed.Path("/v3/tokenreview").Methods(http.MethodPost).Handler(&webhook.TokenReviewer{})
-	authed.Path("/metrics").Handler(metricsHandler)
 	authed.Path("/metrics/{clusterID}").Handler(metricsHandler)
```
cmurphy, line 105: This would need to apply to this route too right?

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
...
 	authed.PathPrefix("/v3").Handler(managementAPI)
 
+	// Authenticated routes
+	metricsAuthed := mux.NewRouter()
+	metricsAuthed.UseEncodedPath()
+	impersonsatingAuth := auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))
+	tokenReviewAuth := impersonsatingAuth.Chain(auth.ToMiddleware(requests.NewTokenReviewAuth(scaledContext.K8sClient.AuthenticationV1())))
+	metricsAuthed.Use(mux.MiddlewareFunc(tokenReviewAuth))
+	metricsAuthed.Use(mux.MiddlewareFunc(rbac.NewAccessControlHandler()))
+	metricsAuthed.Use(requests.NewAuthenticatedFilter)
```
cmurphy, line 120: I was hoping there would be an easier way to add a middleware to one route without this duplicate code but it looks like I was mistaken

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
...
+
+	metricsAuthed.Path("/metrics").Handler(metricsHandler)
+
 	unauthed.NotFoundHandler = authed
 	return func(next http.Handler) http.Handler {
-		authed.NotFoundHandler = next
+		authed.NotFoundHandler = metricsAuthed
```
cmurphy, line 126: Why is this needed?

rmweir: This chains the routing logic together, first it looks for routes from the unauth handler, if nothing matches then from auth, if nothing matches then from metrics auth. This is also relates to your two above questions. If metrics matches in the auth handler it will not be forwarded to the metrics handler. Using a new auth handler is how we are only using this auth for metrics otherwise we could use less code and just use the auth handler with metrics endpoint but then that performs the metrics auth logic on everything that fails first auth check.

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.PathPrefix("/v3/token").Handler(tokenAPI)
 	authed.PathPrefix("/v3").Handler(managementAPI)
 
+	// Authenticated routes
```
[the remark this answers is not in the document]
rmweir: Changed to "Metrics authenticated route".

cmurphy (approved):

pkg/multiclustermanager/routes.go
aiyengar2: Since this line does not use the `next` var, can we move it outside of this function (just like how we set `unauthed.NotFoundHandler`? I think it only needs to be set once.

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.PathPrefix("/v3/token").Handler(tokenAPI)
 	authed.PathPrefix("/v3").Handler(managementAPI)
 
+	// Metrics authenticated route
+	metricsAuthed := mux.NewRouter()
+	metricsAuthed.UseEncodedPath()
+	impersonatingAuth := auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))
+	tokenReviewAuth := impersonatingAuth.Chain(auth.ToMiddleware(requests.NewTokenReviewAuth(scaledContext.K8sClient.AuthenticationV1())))
+	metricsAuthed.Use(mux.MiddlewareFunc(tokenReviewAuth))
+	metricsAuthed.Use(mux.MiddlewareFunc(rbac.NewAccessControlHandler()))
```
aiyengar2, line 119: Can we reuse the `rbac.NewAccessControlHandler()` from above instead of calling it independently?

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 	authed.PathPrefix("/v3/token").Handler(tokenAPI)
 	authed.PathPrefix("/v3").Handler(managementAPI)
 
+	// Metrics authenticated route
+	metricsAuthed := mux.NewRouter()
+	metricsAuthed.UseEncodedPath()
+	impersonatingAuth := auth.ToMiddleware(requests.NewImpersonatingAuth(sar.NewSubjectAccessReview(clusterManager)))
```
aiyengar2, line 116: can we define this above and use this in https://github.com/rancher/rancher/blob/978e165156362d98a323452049fb04c7fec64f5d/pkg/multiclustermanager/routes.go#L96 as well?

Based on the fact that the logic for [`Chain`](https://github.com/rancher/steve/blob/e7119828b8f9a9ca2cb2620ba650bd0508ca463e/pkg/auth/filter.go#L47-L51) emits a brand new function, we should be able to reuse the same function for impersonating auth in both places rather than defining it twice.

pkg/multiclustermanager/routes.go
```diff
@@ -111,9 +110,21 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
...
+	tokenReviewAuth := impersonatingAuth.Chain(auth.ToMiddleware(requests.NewTokenReviewAuth(scaledContext.K8sClient.AuthenticationV1())))
```
aiyengar2, line 117: Is this the right order? I'm not sure if this is correct or if the following is correct:

```go
auth.ToMiddleware(requests.NewTokenReviewAuth(scaledContext.K8sClient.AuthenticationV1())).Chain(impersonatingAuth)
```

Based on [the logic in Steve](https://github.com/rancher/steve/blob/e7119828b8f9a9ca2cb2620ba650bd0508ca463e/pkg/auth/filter.go#L47-L51) that looks like `return m(middleware(next))`, it seems like we evaluate the `middleware` provided as an argument (e.g. the tokenAuthReview) before evaluating the `m` provided as self (e.g. the impersonatingAuth), whereas what we want is for impersonation to be attempted, followed by tokenAuthReview to be checked.

pkg/auth/requests/token_review.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		logrus.Debugf("tokenReview failed: %v", err)
+		return info, false, nil
+	}
+
+	tokenReviewUserInfo := &user.DefaultInfo{
+		Name:   tokenReview.Status.User.Username,
+		UID:    tokenReview.Status.User.UID,
+		Groups: tokenReview.Status.User.Groups,
+	}
```
aiyengar2, line 56: Do we need to do some check to wait till the status is populated?

aiyengar2 (changes requested):

pkg/auth/requests/token_review.go
rmweir: I believe the status should always be populated https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication:
`The remote service is expected to fill the status field of the request to indicate the success of the login.`

pkg/multiclustermanager/routes.go
rmweir: Effectively I think either way works but you're right this isn't the intended order, changed.

pkg/multiclustermanager/routes.go
rmweir: Thanks to some of Arvind's feedback I was able to reuse certain parts.

pkg/multiclustermanager/routes.go
cmurphy: What about this?

cmurphy (changes requested):

rmweir:
@cmurphy I explained this here https://github.com/rancher/rancher/pull/36220#discussion_r808543272. Basically the metrics endpoint is fully covered by the second router and the request will not be forwarded to that router unless the route is not found in the regular auth router; this is what's going on in the `NotFoundHandler` part.

cmurphy:
@rmweir in that case shouldn't the `/metrics/{clusterID}` route be removed from the regular authed router?

aiyengar2 (dismissed):
LGTM

cmurphy:
> @rmweir in that case shouldn't the /metrics/{clusterID} route be removed from the regular authed router?

I ran this through a debugger and found that querying /metrics/{clusterID} doesn't hit this token review authenticator. I assumed you want both /metrics and /metrics/{clusterID} to be accessible by service accounts, is that incorrect?

cmurphy (dismissed):
Got it, thanks!

cbron (approved):

PR #36235, opened 2022-01-22
fix ssh behind proxy 

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/28411

add the support for provisioning cluster behind a proxy by letting rancher-machine use external ssh binary and adding dependencies to both Rancher image and the Jail directory

another part of the fix goes into rancher/machine: https://github.com/rancher/machine/pull/159

### DIFF
--- a/package/jailer.sh
+++ b/package/jailer.sh
@@ -79,11 +79,17 @@ cp -l /usr/bin/kustomize.sh /opt/jail/$NAME
 # Hard link ssh into the jail
 cp -l /usr/bin/ssh /opt/jail/$NAME/usr/bin
 
+# Hard link nc into the jail
+cp -l /usr/bin/nc /opt/jail/$NAME/usr/bin
+
 # Hard link cat into the jail
 cp -l /bin/cat /opt/jail/$NAME/bin/
 
 # Hard link bash into the jail
 cp -l /bin/bash /opt/jail/$NAME/bin/
+
+# Hard link sh into the jail
+cp -l /bin/sh /opt/jail/$NAME/bin/
 
 # Hard link rm into the jail
 cp -l /bin/rm /opt/jail/$NAME/bin/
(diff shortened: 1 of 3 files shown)


pkg/controllers/management/node/utils.go
```diff
@@ -41,14 +41,14 @@ const (
 
 func buildAgentCommand(node *v3.Node, dockerRun string) []string {
 	drun := strings.Fields(dockerRun)
-	cmd := []string{"--native-ssh", "ssh", node.Spec.RequestedHostname}
+	cmd := []string{"ssh", node.Spec.RequestedHostname}
```
jiaqiluo, line 44: note to reviewers:
by default, rancher-machine uses the external ssh binary found in the running environment.

package/jailer.sh
```diff
@@ -79,12 +79,18 @@ cp -l /usr/bin/kustomize.sh /opt/jail/$NAME
 # Hard link ssh into the jail
 cp -l /usr/bin/ssh /opt/jail/$NAME/usr/bin
 
+# Hard link nc into the jail
```
thedadams, line 82: Do we need to do anything to allow users that are upgrading to Rancher v2.6.4 to be able to use this new functionality?

jiaqiluo: There is no action on the user side for them to start using this new functionality. Of course, they need to set the proxy 
 variables properly when installing Rancher. 

This might be a case that I do not think will happen: users upgrade the Rancher server from a previous version with a non-functional rancher-provisioned downstream cluster. however, since this is kind of a new feature, deleting the broken one then creating a new one will be the solution for this case.

thedadams: OK. I will have to think about whether there is a more elegant way to handle this upgrade scenario.

thedadams: @jiaqiluo Now that I have thought about it, I think it is unlikely that a user will upgrade an air-gapped setup with clusters that are broken like this because this change is also need to successfully provision clusters. Is that correct?

thedadams (approved):

paynejacob (approved):

Issue #35334, opened 2021-11-01
cattle-node-cleanup jobs stick around forever


herrbpl:
**Rancher Server Setup**
- Rancher version: v2.6.1
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): Kubernetes v1.17.4, RKE1
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.17.4
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom
**Describe the bug**
When cluster-autoscaler manages AWS SG-s and removes nodes, jobs prefixed with cattle-node-cleanup remain forever in cluster unless deleted manually, cluster is unable those jobs because underlying node is already terminated. Problem started with upgrade to Rancher 2.6.x from 2.5.5

**To Reproduce**
1. Add custom node (EC2 instance) to custom RKE1 cluster using registration information. 
2. Wait until node is fully operational
3. remove node by terminating in AWS console

**Result**
Node is gone but cleanup job sticks around. 
**Expected Result**
After a while, cluster should notice that node is gone and remove it. There should not be cleanup job sticking around because node is not ever coming back

**Screenshots**
![image](https://user-images.githubusercontent.com/2135514/139689239-c3f87dad-e98d-48a0-93ff-6d58d74af3ff.png)

**Additional context**
This probably also prevents cleanup of machine records, resources for machines stick around for days.
![image](https://user-images.githubusercontent.com/2135514/139689479-3113b64b-63b3-410e-98e5-932c980fb761.png)

ebenezar-mccoy:
In my case, I was performing a series of tests

in one of them:
- I deleted the VM first - I needed to simulate recovery in a situation when VM was unrecoverable
- then I attempted to delete the node via cli and web - I did it several times
- most likely my attempts of deleting an unavailable node created several cattle-node-cleanup jobs, in my case they started one after another 
- after few manual deletions of cattle-node-cleanup* , the node got deleted from my cluster

herrbpl:
On additional node, I think this might be issue related to clusters where windows nodes are in existance. 
Failing job and pod example
JOB:
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  creationTimestamp: "2021-11-19T08:24:12Z"
[...]

PR #36236, opened 2022-01-22
Don't try to cleanup node if it does not exist

thedadams:
If the cleanup job timed out (like if the VM was deleted), then the node
will still exist in the cluster but it will have a Not Ready status.
Therefore, the cleanup job will get created to tolerate this status, but
the pod will never be scheduled because the VM was deleted.

After this change, if the cleanup job times out, then it will be
deleted.

Issue:
https://github.com/rancher/rancher/issues/35334

### DIFF
--- a/pkg/controllers/management/node/cleanup.go
+++ b/pkg/controllers/management/node/cleanup.go
@@ -196,10 +196,7 @@ func (m *Lifecycle) waitUntilJobCompletes(userContext *config.UserContext, job *
 
 		logrus.Infof("[node-cleanup] job %s finished, continuing to delete v3 node", job.Name)
 		return true, nil
-	}); err != nil {
-		if err.Error() == "timed out waiting for the condition" {
-			return nil // gave it 30s and it timed out so move to delete v1 node
-		}
+	}); err != nil && err.Error() != "timed out waiting for the condition" {
 		return err
 	}
 


paynejacob (dismissed):

kinarashah (dismissed):

pkg/controllers/management/node/cleanup.go
```diff
@@ -153,6 +153,13 @@ func (m *Lifecycle) cleanRKENode(node *v3.Node) error {
 		return nil
 	}
 
+	if _, err := userContext.K8sClient.CoreV1().Nodes().Get(m.ctx, node.Status.NodeName, metav1.GetOptions{}); kerror.IsNotFound(err) {
+		logrus.Debugf("node [%s] is already deleted, cannot clean RKE node", node.Status.NodeName)
+		return nil
+	} else if err != nil {
```
Oats87, line 159: out of curiosity, is it possible that (before) we ever got to this point with an unhealthy control plane on the downstream cluster?

jakefhyde (dismissed):

paynejacob:
@thedadams have you done any testing with this? The logic seems sound but I would want to reproduce the cleanup job issue before we merge this.

thedadams:
@paynejacob Good call. Turns out that my original change would not have fixed the problem. The issue is that the node may still exist in the cluster even thought the VM was deleted. This will cause the Job to get created (and the pod may actually get scheduled), but the pod will never run because the VM is gone.

I updated this PR to just delete the job on timeout. So the job will still get created, but it will be cleaned up in cases where the VM was deleted.

jakefhyde (approved):

paynejacob (approved):

PR #36245, opened 2022-01-24
Fix s390x issues

raulcabello:
- fix drone git clone step stuck on s390x
- fix wrong images for drone digests step on s390x
- add missing kubectl on s390x

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -563,7 +563,17 @@ platform:
 node:
   arch: s390x
 
+clone:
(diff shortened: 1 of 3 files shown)


rosskirkpat (changes requested):
Looks good but we should squash the commits before merging.

rosskirkpat (approved):
LGTM, thanks for squashing!

sirredbeard (approved):

Issue #36209, opened 2022-01-20
etcd backups 'thrash' when multiple backups occur at once (i.e. user starts manual backup, and recurring backup starts before manual backup finishes)


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.5.12-rc2
- Installation option (Docker install/Helm Chart): docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): n/a
- Proxy/Cert Details:n/a

**Information about the Cluster**
- Kubernetes version:any version
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): any provider

**User Information**
- What is the role of the user logged in? admin

   
**Describe the bug**

If multiple backups are occurring at once, there is some thrashing between the backups, prolonging the time it takes for a backup to complete. 

**To Reproduce**
 
* provision a downstream cluster and set the backup storage to s3 with valid credentials
 * once the cluster is active, take a manual snapshot at the same time as the first automatic backup occurs

**Result**
* the two backups will compete for the first slot in the queue.  This can result in the first backup taken with an earlier timestamp, but actually taken after the second backup. It also is not optimal as the backups take time to resolve which will be backed up first. 

**Expected Result**

we should queue the 2nd backup and wait for the first backup to complete before attempting any others. 

**Screenshots**

**Additional context**

found when testing https://github.com/rancher/rancher/issues/36038

also worth noting -- with 2 simultaneous backups, they seem to succeed. however 3 or more, only the last 2 backups taken will succeed and the others have errors instead of completing

slickwarren:
see https://github.com/rancher/rancher/issues/36210#issuecomment-1032076311 for testing

timhaneunsoo:
Tested this issue on RKE1 and RKE2. RKE1 seems to be fixed, but RKE2 is blocked by

jakefhyde:
@timhaneunsoo This only affects RKE1, so RKE2 doesn't need to be tested (although RKE2 may still need to be tested for GA, it would be unrelated to this issue)

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head 8ebbb18
Rancher cluster type: HA
Docker version: 20.10

Downstream cluster type: RKE1 node driver

----------

### Testing: ###

Tested this issue with the following:
[...]

PR #36251, opened 2022-01-25
36209 etcd backup thrashing

jakefhyde:
Issue: #36209 
Etcd backups should not run concurrently, as rke/rke-tools expects to use the same container (`etcd-snapshot-once`) for snapshots, and multiple concurrent backups will cause thrashing for these resources. Backups (both manual and recurring) are now queued, and will be popped on cluster sync (5 minute interval), or on creation. Recurring snapshots will not be queued if another recurring snapshot is running or in queue.

### DIFF
--- a/pkg/controllers/management/etcdbackup/etcdbackup.go
+++ b/pkg/controllers/management/etcdbackup/etcdbackup.go
@@ -157,13 +157,13 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
 		return nil
 	}
 
-	clusterBackups, err := c.getRecuringBackupsList(cluster)
+	backups, err := c.getBackupsList(cluster)
 	if err != nil {
 		return err
 	}
 
 	// cluster has no backups, we need to kick a new one.
-	if len(clusterBackups) == 0 {
+	if len(backups) == 0 {
 		logrus.Debugf("[etcd-backup] cluster [%s] has no backups, creating first backup", cluster.Name)
 		newBackup, err := c.createNewBackup(cluster)
 		if err != nil {
@@ -176,7 +176,7 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
 	// check for backups in progress
 	if cluster.Spec.RancherKubernetesEngineConfig != nil {
 		clusterTimeout := time.Duration(cluster.Spec.RancherKubernetesEngineConfig.Services.Etcd.BackupConfig.Timeout) * time.Second
-		for _, clusterBackup := range clusterBackups {
+		for _, clusterBackup := range backups {
 			backupCreationTime, err := getBackupCreatedTime(clusterBackup)
 			if err != nil {
 				logrus.Errorf("[etcd-backup] cluster [%s] backup [%s] is missing creation time: %v", cluster.Name, clusterBackup.Name, err)
@@ -191,8 +191,20 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
 		}
 	}
 
-	newestBackup := clusterBackups[0]
-	for _, clusterBackup := range clusterBackups[1:] {
+	recurringBackups := getRecurringBackupsList(backups)
+	// cluster has no recurring backups, because the user created one before we could.
+	if len(recurringBackups) == 0 {
+		logrus.Debugf("[etcd-backup] cluster [%s] has no backups, creating first backup", cluster.Name)
+		newBackup, err := c.createNewBackup(cluster)
+		if err != nil {
+			return fmt.Errorf("[etcd-backup] error while creating backup for cluster [%s]: %v", cluster.Name, err)
+		}
+		logrus.Debugf("[etcd-backup] cluster [%s] new backup is created: %s", cluster.Name, newBackup.Name)
+		return nil
+	}
+
+	newestBackup := recurringBackups[0]
+	for _, clusterBackup := range recurringBackups[1:] {
 		if getBackupCompletedTime(clusterBackup).After(getBackupCompletedTime(newestBackup)) {
 			newestBackup = clusterBackup
 		}
@@ -212,7 +224,7 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
 	}
 
 	// rotate old backups
-	return c.rotateExpiredBackups(cluster, clusterBackups)
+	return c.rotateExpiredBackups(cluster, recurringBackups)
 }
 
 func (c *Controller) createNewBackup(cluster *v3.Cluster) (*v3.EtcdBackup, error) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -206,13 +232,91 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
...
+			continue
+		}
+		backupCreationTime, err := getBackupCreatedTime(backup)
+		if err != nil {
+			log.Errorf("[etcd-backup] cluster [%s] backup [%s] is missing creation time: %v", cluster.Name, backup.Name, err)
+			continue
+		}
+		// cluster backup is younger than its timeout and completion is unknown
+		// therefore, it's currently running
+		if time.Since(backupCreationTime) < clusterTimeout && rketypes.BackupConditionCompleted.IsUnknown(backup) {
```
annablender, line 281: A backup could be in any state if it's unknown. Do we have a way for the user to tell what state a backup is in if it's unknown, other than logs? Just a thought

annablender (commented):
Had a few comments, lots of changes though! Great work!

pkg/controllers/management/etcdbackup/etcdbackup.go
annablender: We may already have that and I just missed it!

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -157,42 +196,29 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
...
-				continue
-			}
-			// cluster backup is younger than its timeout and completion is unknown
-			// therefore, it's currently running
-			if time.Since(backupCreationTime) < clusterTimeout && !rketypes.BackupConditionCompleted.IsTrue(clusterBackup) {
-				logrus.Debugf("[etcd-backup] cluster [%s] is currently creating a backup, skipping", cluster.Name)
-				return nil
-			}
-		}
+	if anyBackupsRunning(cluster, recurringBackups) || anyBackupsQueued(recurringBackups) {
```
annablender, line 216: Swept over this before! This is the logic I was looking for. Don't create any more recurring backups if there's one already running or queued, so they don't run concurrently. Nice job.

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -206,13 +232,91 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
...
-		logrus.Debugf("[etcd-backup] new backup created: %s", newBackup.Name)
+		log.Debugf("[etcd-backup] new backup created: %s", newBackup.Name)
 	}
 
-	// rotate old backups
-	return c.rotateExpiredBackups(cluster, clusterBackups)
+	return c.rotateExpiredBackups(cluster, recurringBackups)
+}
+
+func (c *Controller) doBackupForCluster(b *v3.EtcdBackup, cluster *v3.Cluster) (runtime.Object, error) {
```
annablender, line 243: Rename to `createBackupForCluster`?

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -135,20 +138,56 @@ func (c *Controller) clusterBackupSync(ctx context.Context, interval time.Durati
...
+		if !next.Spec.Manual {
+			return c.rotateExpiredBackups(cluster, getRecurringBackupsList(backups))
 		}
 	}
+
 	return nil
 }
 
-func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
+func (c *Controller) doRecurringBackup(cluster *v3.Cluster) error {
```
annablender, line 190: Something that's bugging me here is that `doRecurringBackup` doesn't describe _what_ this function does. I think it would improve readability to be more specific, maybe call the function `createRecurringBackup` or `checkRecurringBackup.`

pkg/controllers/management/etcdbackup/etcdbackup.go
jakefhyde: This actually tripped me up for a bit, because a condition has to exist and be set to `"Unknown"` for `IsUnknown` to be true from a [norman perspective](https://github.com/rancher/norman/blob/4feb41eafabded527f156460e70d77947f9a5973/condition/condition.go#L66), which is apparently against the [kubernetes community api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#:~:text=Condition%20status%20values%20may%20be%20True%2C%20False%2C%20or%20Unknown.%20The%20absence%20of%20a%20condition%20should%20be%20interpreted%20the%20same%20as%20Unknown.%20How%20controllers%20handle%20Unknown%20depends%20on%20the%20Condition%20in%20question.). So in lieu of a running condition, setting Completed to Unknown means creation succeeded, and we are not sure if it's successful or failed, we just know that it is running. I think this is probably a bad approach in hindsight, as currently etcd backups only utilize the 3 following states: Creating, Activating, and Active.
queue-4
I think another condition would be better in order to signify that the backups is currently running (I'm almost certain such a condition already exists, just have to figure out which).

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -96,33 +94,38 @@ func (c *Controller) Create(b *v3.EtcdBackup) (runtime.Object, error) {
...
+		}
+		return bObj, nil
 	}
+
 	return b, nil
 }
 
-func (c *Controller) Remove(b *v3.EtcdBackup) (runtime.Object, error) {
-	logrus.Debugf("[etcd-backup] deleting backup %s ", b.Name)
+func (c *Controller) Remove(b *v3.EtcdBackup) (runtime.Object, error) { // todo(jhyde): check started
```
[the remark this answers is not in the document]
jakefhyde: Removed, also made sure to actually do what the todo comment suggested.

pkg/controllers/management/etcdbackup/etcdbackup.go
jakefhyde: @thedadams Do you have any input regarding this? Specifically regarding what to do to make the etcd backup indicate it is progressing from a UI/UX perspective?

annablender: Had a discussion and like @jakefhyde said, the [Norman frontend](https://github.com/rancher/norman/blob/4feb41eafabded527f156460e70d77947f9a5973/condition/condition.go#L66) uses a series of conditions (in other places as well) including Condition.Unknown to indicate a backup is currently running.

My initial concern was: There's no reason to change it unless it could potentially cause a slew of user issues. You know, if a backup fails while it's in 'Unknown' state it won't alert the user until the backup times out correct? It wouldn't be a bug no but it could be an inconvenience if a user has an annoyingly high timeout value set, and then are wondering what happened to their _running_ backup..

So Norman already has seton conditions that wouldn't be ideal to change because it's used in other places in addition to etcd backup, @thedadams correct me if I'm wrong there. The backend conditions just aren't intuitive when you look at them on the backend. etcdbackup.go https://github.com/rancher/rancher/blob/6ade3e8cf06e618d98b1b390c579755504fad2b2/pkg/controllers/management/etcdbackup/etcdbackup.go#L358-L362 does check to see whether a backup failed before timeout and if so, puts it into a Condition.Failed state. This is fine with me

paynejacob (approved):

pkg/controllers/management/etcdbackup/etcdbackup.go
jakefhyde: Ah I think there may be a disconnect here: If a backup fails, it fails instantly (i.e. if the minio endpoint doesn't exist anymore, just exit). This will immediately update the condition to False, and show correctly in the UI. The timeout only comes into play when the backup is taking too long (which is a pretty rare case, most etcd backups complete pretty quickly, however a minio upload on a slow network can cause this), where we consider the backup to be a failure if it goes longer than the timeout, at which point it fails.

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -206,13 +232,91 @@ func (c *Controller) doClusterBackupSync(cluster *v3.Cluster) error {
...
-			return fmt.Errorf("[etcd-backup] error while create new backup for cluster [%s]: %v", cluster.Name, err)
+			return fmt.Errorf("error while create new backup for cluster [%s]: %v", cluster.Name, err)
 		}
-		logrus.Debugf("[etcd-backup] new backup created: %s", newBackup.Name)
+		log.Debugf("[etcd-backup] new backup created: %s", newBackup.Name)
 	}
 
-	// rotate old backups
-	return c.rotateExpiredBackups(cluster, clusterBackups)
+	return c.rotateExpiredBackups(cluster, recurringBackups)
```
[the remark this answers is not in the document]
annablender: LGTM @jakefhyde

PR #36290, opened 2022-01-27
update fossa step in drone

rosskirkpat:
Update the Fossa step in .drone.yml to address an issue with Fossa when it runs on drone-publish


oxr463 (approved):

sirredbeard:
I would like to merge this after we address https://github.com/rancherlabs/eio/issues/768 so that we don't potentially delay a published release, like the 2.5 patch planned for today

luthermonson (approved):

PR #36300, opened 2022-01-27
Remove Thrift as a direct dependency

macedogm:


rancherbot (approved):

PR #36307, opened 2022-01-28
Remove Thrift unused direct dependency

macedogm:


rancherbot (approved):

PR #22610, opened 2019-09-03
Bump vendor for rke and kontainer-driver-metadata

mrajashree:

PR #36312, opened 2022-01-28
Remove unused Thrift direct dependency

macedogm:
```
% go mod why -m github.com/apache/thrift  
# github.com/apache/thrift
(main module does not need module github.com/apache/thrift)
```

The remaining Thrift in [`go.sum`](https://github.com/macedogm/rancher/blob/108719c435c029f5f719c37bd39126a7729b4f47/go.sum#L179) is imported through another dependency, but not shipped:
```
go.opencensus.io@v0.20.1 github.com/apache/thrift@v0.12.0
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,6 @@ module github.com/rancher/rancher
 go 1.16
 
 replace (
-	git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
-
 	github.com/docker/distribution => github.com/docker/distribution v2.7.1+incompatible // oras dep requires a replace is set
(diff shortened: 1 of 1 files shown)


rancherbot (approved):

cmurphy (approved):

cbron (approved):
go-mod-why is clear on both of those items, maybe this dependency was transitive and got removed

cmurphy:
I think it was added accidentally, it doesn't look like it was ever used https://github.com/rancher/rancher/pull/22610

Issue #27750, opened 2020-06-26
Creating a namespace via kubectl will not set Container Default Resource Limit, creating a pod fails due to quota


5nafu:
**What kind of request is this:** bug

**Steps to reproduce (least amount of steps as possible):**
* Create a project in the Rancher UI with Resource Quotas and Container Default Resource Limits set (note the ProjectID)
* Create a new namespace via kubectl with the projectID annotation:
  ```
  apiVersion: v1
  kind: Namespace
  metadata:
    annotations:
      field.cattle.io/projectId: c-jrlpt:p-tjnls
    labels:
      field.cattle.io/projectId: p-tjnls
    name: test-namespace
  ```
* try to create a pod in the namespace:
  ```
  ---
  apiVersion: v1
  kind: Pod
  metadata:
    name: debug
  spec:
    containers:
    - name: count
      image: busybox:1.28
      args: [/bin/sh, -c, 'while true; do echo "Wait ..."; sleep 1; done']
  ```

**Result:**
* A new namespace is created with the `Namespace Default Limit`s from the project, but without `Container Default Resource Limit`:
  ```
  apiVersion: v1
  kind: Namespace
  metadata:
    annotations:
      cattle.io/status: '{[...]}'
      field.cattle.io/projectId: c-jrlpt:p-tjnls
      field.cattle.io/resourceQuota: '{"limit":{"persistentVolumeClaims":"40","requestsStorage":"4Ti","limitsCpu":"64","limitsMemory":"128Gi"}}'
      kubectl.kubernetes.io/last-applied-configuration: |
        [...]
      lifecycle.cattle.io/create.namespace-auth: "true"
    creationTimestamp: "2020-06-26T14:46:45Z"
  [...]
  ```
  I would expect to find the annotation `field.cattle.io/containerDefaultResourceLimit` in the newly created namespace.

* The pod creation is rejected because the quota failed:
  ```
  Error from server (Forbidden): error when creating "pod.yaml": pods "debug" is forbidden: failed quota: default-2vpng: must specify limits.cpu,limits.memory
  ```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): 2.4.4
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):  Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory): metal 24 Cores 126GB RAM
[...]

PR #36317, opened 2022-01-28
Pass container limits from project in a new namespace if it does not override them

maxsokolovsky:
https://github.com/rancher/rancher/issues/27750

To enable the propagation of container default limits from a project to its new namespaces made via `kubectl` directly, we must explicitly set project limits (if defined), if the namespace itself does not provide its own (override) values.

### DIFF
--- a/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
+++ b/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
@@ -450,9 +450,11 @@ func (c *SyncController) getResourceLimitToUpdate(ns *corev1.Namespace) (*corev1
 		return convertPodResourceLimitToLimitRangeSpec(updatedLimit)
 	} else if nsLimit != nil {
 		return convertPodResourceLimitToLimitRangeSpec(nsLimit)
+	} else if projectLimit != nil {
+		return convertPodResourceLimitToLimitRangeSpec(projectLimit)
+	} else {
+		return nil, nil
 	}
-
-	return nil, nil
 }
 
 func completeQuota(existingQuota *v32.NamespaceResourceQuota, defaultQuota *v32.NamespaceResourceQuota) (*v32.NamespaceResourceQuota, error) {


pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -114,6 +115,8 @@ func limitsChanged(existing []corev1.LimitRangeItem, toUpdate []corev1.LimitRang
 }
 
 func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Object, error) {
+	c.ensureContainerDefaultLimitsAnnotationSet(ns)
```
cmurphy, line 118: Calling this function here seems a little out of place to me, because this function is called `CreateResourceQuota` but this is about limit ranges. There's another method `createLimitRange` called just after in the sync handler that might be more appropriate. That way you could also deduplicate some effort by using the result of the `getResourceLimitToUpdate` call to get the default and then calling something similar to `validateAndSetNamespaceQuota` to set it.

cmurphy (changes requested):

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Agreed! Can't we just call the function first thing in the `createLimitRange` method? The new function just sets the annotation, and all the other logic takes care of the actual LimitRange creation. The only thing duplicated is the call to `getProjectContainerDefaultLimit`, which might be OK? Sorry, maybe I am missing something - not seeing how to refactor it more while avoiding restructuring more things.

cmurphy (approved):

MbolotSuse (changes requested):
Got some general feedback here. I don't think that we necessarily need to change the current approach, but I do think that there's some alternatives worth exploring.

First, the current approach seems to be to init a default here (if necessary) so that another controller can propagate the defaults from the project. This other controller seems to be the one at https://github.com/rancher/rancher/blob/release/v2.6/pkg/api/norman/store/namespace/store.go#L159-L162 . It might be a good idea to make changes to this other controller so that if it detects the annotation as missing it also sets the values (i.e. change the other controller so it can handle the "" case as well as the "{}" case). This way - we are more explicit on what each controller should handle, rather than mixing functionality like this PR does.

Secondly, this syncController has code which causes it to ignore cases where there is no limit set on the namespace but there is a limit set on the project: https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go#L426-L456 . I think that this could be changed to allow the use of project limits in the cases where project limits are defined but namespace limits are not defined.

maxsokolovsky:
@MbolotSuse, good points. One thing though is that the namespace store code is not run in v2.6. I believe it is run in v2.5, and we do need to backport the fix for this issue, so we can address it then.

About the second point - I agree, our logic would be more explicit if we just take the project's container limits and use those, instead of setting annotations to empty strings. I tried it - it works the same. 

@cmurphy, what do you think? Instead of modifying the annotation, we would rather have another `else if` clause that checks if the project's container limits exist, and, if yes, return those to the caller. (By then we would know that the namespace itself does not have custom container limits - we would add the last `else if` clause.)
https://github.com/rancher/rancher/blob/19f459a07bdeca8d9eb09e2b72d0fbe7c4cca3aa/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go#L451

cmurphy:
I think this is close to what I was trying to say in https://github.com/rancher/rancher/pull/36317#discussion_r796934394 before we decided offline to use the simpler default annotation, though I don't think I realized that `getResourceLimitToUpdate` doesn't currently work if the ns limit is unset. I think this sounds fine.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -450,9 +450,11 @@ func (c *SyncController) getResourceLimitToUpdate(ns *corev1.Namespace) (*corev1
 		return convertPodResourceLimitToLimitRangeSpec(updatedLimit)
 	} else if nsLimit != nil {
 		return convertPodResourceLimitToLimitRangeSpec(nsLimit)
+	} else if projectLimit != nil {
+		return convertPodResourceLimitToLimitRangeSpec(projectLimit)
+	} else {
+		return nil, nil
 	}
```
MbolotSuse, line 457: Nit: Since you technically return right after this, you don't really need the `else{}` clause. The if block will just fall through to the return case in the situation where we don't have a combo limit, project limit, or namespace limit.

That being said, this is fine since it makes that behavior more explicit, so I don't think it really matters which solution you chose. Up to you - feel free to resolve after you decide what you want to do.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -504,10 +506,6 @@ func completeLimit(existingLimit *v32.ContainerResourceLimit, defaultLimit *v32.
 		}
 	}
 
-	if reflect.DeepEqual(existingLimitMap, newLimitMap) {
-		return nil, nil
-	}
-
```
MbolotSuse, line 510: I'm not exactly sure why you removed this, would you mind elaborating?

Near as I can tell, this caused the code to report a nil 'combined map' in cases where the namespace limit and the combined map (produced by replacing all values in the project limit whose keys are also in the namespace limit) are identical. This would cause the calling function (`getResourceLimitToUpdate`) to use the namespace limit. This new case functions identically, except that we return up the combined map from this function (which is the same as the namespace limit). Was this more of an efficiency change?

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: This is just a matter of style, two `else if` clauses look dangling without the concluding `else`, which is indeed a catch-all clause here, since it returns `nil, nil`.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: This is for the following scenario: there are project's container limits and a namespace's overridden values. This unnecessary map equality check prevents the override from being saved (because nil is returned). Removing this makes sure a namespace's overrides are actually honored.

I tested different scenarios:
- project container limit **defined**, namespaces override **defined**
- project container limit **defined**, namespaces override **undefined**
- project container limit **undefined**, namespaces override **defined**
- project container limit **undefined**, namespaces override **undefined**

MbolotSuse: Here's my confusion - in the case that you described, this function would return nil (since the namespace's values were all that ended up being used). Since there are no combined values, the calling function would then use the values from the namespace. Which is the same as what will be returned from the new function. Correct me if I am wrong there. Would you mind explaining what you see as old/new side for that scenario that you described?

maxsokolovsky: You're right, I just saw a flaw in my tests. Re-tested again with the map equality comparison. Will update the code.

MbolotSuse (approved):
LGTM.

I think that there's a potential edge case with the project limit returning an empty struct (i.e. not Nil - valid obj just empty) but it looks like we have things in place to ensure that value is properly parsed so that this case wouldn't occur.

cmurphy (approved):

PR #36320, opened 2022-01-28
Updated dynamic client

igomez06:
**Framework updates**

- Dynamic client not takes a testing.T object to log the host of the dynamic client.
- Removed GenerateKubeconfig from GetDownStreamClusterClient, and instead just updated the host.


paynejacob (changes requested):
I would simplify this by exposing the Test we already pass to the session.  https://github.com/rancher/rancher/blob/release/v2.6/tests/framework/pkg/session/session.go  We want our tests to push everything through the session.

paynejacob (approved):

bmdepesa (approved):
LGTM

Issue #18877, opened 2019-03-13
container default resource limit from the project level is not propagated to existing namespaces


deniseschannon:
master 3/13

Resource Quota Workflow

- Add resource quota to default project and set a "Namespace Default Limit"
- Any existing namespaces automatically get these quotas set at the namespace level
- Any new namespace gets these quotas set at the namespace level

Container Default Resource Limit
- Add a container default resource limit to a default project
- Any existing namespace does **NOT** get these limit set at the namespace level
- Any new namespace gets these quotas set at the namespace level

Main Difference is around existing namespaces:
For resource quotas, they are automatically propagated to namespaces that exist in your project.

For container default resource limit, you are required to update each existing namespace with these limits. 

This is a problem particularly around when setting CPU/Memory quotas because:

1. When you set a quota for CPU/Memory, you are required to set something on the workload creation. 
2. Even if you set a container default resource limit on the project level, you are also required to set a limit on the namespace level in order for it to be used at the workload creation step.

gz#7610

Tejeev:
I believe this to be more details on this bug: 

**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**

1. Create a project with value(s) set for **project quota(s)**
  ![image](https://user-images.githubusercontent.com/8906089/71269404-26461a00-2347-11ea-9bf0-a9170bb3a165.png)
2. Create a namespace within the project with values set for all **Container Default Resource Limits**
  ![image](https://user-images.githubusercontent.com/8906089/71269504-6b6a4c00-2347-11ea-99cd-c247bee67e59.png)
**Result**
At this point there will be no LimitRange created for the namespace**

**Further steps**

3. Set a single **Project Container Default Resource Limit**:
  ![image](https://user-images.githubusercontent.com/8906089/71269289-e54e0580-2346-11ea-84d9-5c71393d98db.png)

**Result**

The LimitRange is not created until the namespace has a **Container Default Resource Limit** configured.  This LimitRange will only have a value for the resource which has a limit set at the Project level (though the value will correctly be that which is set in the Namespace, not the Project).

```
$ kubectl get limitrange -n 11 28643c69-5f0c-4f05-b17c-b4428abcbe37 -o yaml
apiVersion: v1
kind: LimitRange
metadata:
  creationTimestamp: "2019-12-20T16:30:03Z"
  generateName: default-
  labels:
    cattle.io/creator: norman
[...]

PR #36334, opened 2022-01-31
Allow container default limits propagation in absence of project resource quotas

maxsokolovsky:
https://github.com/rancher/rancher/issues/18877

This allows container default limits defined in a project to be propagated to new namespaces (which don't override them) even if the project does not have resource quotas defined.

This is a quick solution that works, however, I have noticed two things in my tests, but they are insignificant.
1. Sometimes the propagated limits are duplicated on the namespace. This does not affect anything, just looks strange.
2. Having created a new namespace with propagated container default limits, one can see that they are missing in the UI. This could be a UI issue from before.

### DIFF
--- a/pkg/controllers/managementuser/resourcequota/resource_quota_common.go
+++ b/pkg/controllers/managementuser/resourcequota/resource_quota_common.go
@@ -203,7 +203,7 @@ func getProjectContainerDefaultLimit(ns *corev1.Namespace, projectLister v3.Proj
 		return nil, nil
 	}
 	project, err := projectLister.Get(projectNamespace, projectName)
-	if err != nil || project.Spec.ResourceQuota == nil {
+	if err != nil {
 		if errors.IsNotFound(err) {
 			// If Rancher is unaware of a project, we should ignore trying to get the default container limit
 			// A non-existent project is likely managed by another Rancher (e.g. Hosted Rancher)


pkg/controllers/managementuser/resourcequota/resource_quota_common.go
```diff
@@ -203,7 +203,7 @@ func getProjectContainerDefaultLimit(ns *corev1.Namespace, projectLister v3.Proj
 		return nil, nil
 	}
 	project, err := projectLister.Get(projectNamespace, projectName)
-	if err != nil || project.Spec.ResourceQuota == nil {
```
cmurphy, line 206: I'm guessing here, but I think this was here because of copypaste from https://github.com/rancher/rancher/pull/36334/files#diff-19dc1cfc80346c6a1f3908d3850916091cedaa8c6ee9f66fbdcbdcfc6594d739L164 and https://github.com/rancher/rancher/pull/36334/files#diff-19dc1cfc80346c6a1f3908d3850916091cedaa8c6ee9f66fbdcbdcfc6594d739L185 . Unlike https://github.com/rancher/rancher/pull/36334/files#diff-19dc1cfc80346c6a1f3908d3850916091cedaa8c6ee9f66fbdcbdcfc6594d739L172 we're not dereferencing ContainerDefaultResourceLimit so I don't think we need the same nil check here.

cmurphy (approved):

ryansann (approved):

PR #36345, opened 2022-02-01
Fix migrations from CAPI machines to plan secrets

thedadams:
There were some errors and typos in the migration logic for moving
Rancher's labels from the CAPI machines to the bootstrap objects and
secrets. This change will fix these issues and allow migrations to be
successful.


Oats87 (approved):
LGTM!

jakefhyde (approved):

Issue #35343, opened 2021-11-02
Support for RKE2 provisioned cluster certificate rotation


snasovich:
Implement support to rotate certificates on RKE2 provisioned cluster, including an option to rotate certificates for individual service(s) (or all services).

Requires https://github.com/k3s-io/k3s/issues/2636 (and the same for RKE2 - https://github.com/rancher/rke2/issues/570)

snasovich:
Unblocked as the changes are in December 2021 RKE2/k3s release.

paynejacob:
### Root cause
new feature
  
### What was fixed, or what changes have occurred
provisioning v2 clusters now support certificate rotation.
  
### Areas or cases that should be tested
- rotating certs with all permutations of node configurations
- rotating certs on upgrade
- rotating specific service certs
  
### What areas could experience regressions?
- upgrades
  
### Are the repro steps accurate/minimal?
kubectl patch -n fleet-default --patch '{"spec":{"rkeConfig":{"rotateCertificates":{"caCertificates":true,"generation":1}}}}' --type=merge cluster/<cluster name>

kubectl patch -n fleet-default --patch '{"spec":{"rkeConfig":{"rotateCertificates":{"caCertificates":true,"generation":1, "services": [<service name here>]}}}}' --type=merge cluster/<cluster name>

rancher-max:
This appears to just constantly rotate in a loop. My setup is just 3 nodes all roles:
<img width="1443" alt="image" src="https://user-images.githubusercontent.com/55997940/156247274-8689f9f8-c4a5-4e9d-9418-20a72b0b2535.png">

v2.6-head `cf4364d`

Reopening for now and will gather more logs and post them when I have them. In this setup I was using DigitalOcean and it appears the server logs here are unavailable using `journalctl` for some reason:
```
root@maxrp1-maxrp1-49700a55-llzdg:~# journalctl -u rke2-server -f
No journal files were found.
```

rancher-max:
I might've reopened this prematurely. After some time, this seems to have resolved itself and completed. Although there's something strange going on since according to the logs shown below it rotated on one node 5 times, another node once, and another node not at all.
<img width="1465" alt="image" src="https://user-images.githubusercontent.com/55997940/156257814-5d61760a-ac51-4295-b669-287e9d96b3fd.png">

However, when I check how many times it actually rotated, it appears to have done the node from those logs 5 times and both other nodes once:
```
# mk6ff
[...]

PR #36347, opened 2022-02-01
[2.6.4] Support certificate rotation for v2 clusters

paynejacob:
issue: #35343 

# PROBLEM

The certificates on the machines need to be rotated at the user's request.  This may be because they are going to expire or the user has a new CA.  This feature allows users to update all or some of the certificates on their cluster via the rancher API.

# SOLUTION

Using the same certificate rotation object as rke1 clusters You can define what certificates to rotate.  The rotation is controlled by the generation field on the certificate rotation object.  Once certificates have been rotated the controlplane status will be updated with the latest certificate generation.

In my testing I also discovered a bug where clusters would get stuck deleting if they have a bad machine-config.  The fix for this was small and is included in this pr.

# TESTING

- 3 node rke2 cluster with all roles on all nodes
- 3 controlplane, 3 etcd, and 3 worker rke2 cluster

I tested these two cases with node drain enabled and disabled and on k3s clusters as well.

```
kubectl patch -n fleet-default --patch '{"spec":{"rkeConfig":{"rotateCertificates":{"caCertificates":true,"generation":1}}}}' --type=merge cluster/<cluster name>
```


jakefhyde (approved):
LGTM

annablender (approved):
LGTM, long as testing was successful

thedadams:
@paynejacob From these changes, it looks like the cluster/control plane will not have any condition updates to indicate that something is happening with the cluster. Is this the case?

If it is, then I think it would be good to have a way to indicate to the user that certificates are rotating and when rotation is finished.

paynejacob:
Talked about this offline: the cluster goes into an updating state while the plan is applied.

PR #36361, opened 2022-02-02
v2.6.4-rc1

paynejacob:


pkg/apis/rke.cattle.io/v1/zz_generated_deepcopy.go
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
thedadams, line 1: Not sure if this should be included.

thedadams (approved):

pkg/apis/rke.cattle.io/v1/zz_generated_deepcopy.go
paynejacob: it knocks go generate back in sync not sure where it was missed from.

Issue #35191, opened 2021-10-19
Deviation from upstream restricted PSP


SheilaghM:
SURE-3199

**Rancher Server Setup**
- Rancher version: Rancher v2.5.8

**Describe the bug**
runAsUser set to RunAsAny in the restricted PSP which is a deviation from upstream example PSP

**To Reproduce**
Checked the documentation and found below

This policy is based on the Kubernetes example restricted policy.
The upstream example restricted PSP shows below.

  runAsUser:
    # Require the container to run without root privileges.
    rule: 'MustRunAsNonRoot'
But the Rancher restricted PSP is not the same.

**Result**
Default restricted policy allows the container to run processes with root user.

**Expected Result**
Default restricted policy should not allow the container to run processes with root user.
Or there should be a note in documentation if this PSP deviation from the upstream example is intentional.

nobbs:
Still interested in a solution to this issue, as this behavior also occurs in versions 2.6.x, as far as I can tell.

macedogm:
@anupama2501 PR https://github.com/rancher/rancher/pull/36367 is ready for review.

macedogm:
Docs PR https://github.com/rancher/docs/pull/3849

macedogm:
Apologies, the issue was closed when the PR got merged. I'm reopening, so QA can properly validate it.

anupama2501:
Verified test cases on 2.6-head 978e165

1. local k8s `v1.22.6-rancher1-2` and PSP set to restricted [local picks default-psp] - created 1 custom rke1 cluster with restricted-no-root, 1 rke1 node driver and and rke2 node driver cluster. Repeated the steps from the above comment - PASS
2. local psp with no restricted and a downstream restricted-no-root psp.Repeated the steps from the above comment - PASS
3. Local cluster rke1 with restricted PSP - default psp is picked, rancher pods deployed. Updated the default-psp runAsUSer rule to `MustRunAsNoRoot`. Deleted a rancher pod and verified the pods couldn't reconcile
```
Failed (12) | 52 mins ago | Error: container has runAsNonRoot and image will run as root (pod: "rancher-6988c76755-fq57r_cattle-system(2c02d913-d45c-498c-b492-56364f8aa5c5)", container: rancher)
```
4. local cluster set to restricted psp, and created a downstream cluster on restricted-psp. Changed the restricted psp to `restricted-no-root`. Deployed some workload pods with no security context added. Verified the pods errored out.
[...]

PR #36367, opened 2022-02-03
Add restricted-noroot PSP policy to match upstream Kubernetes restricted PSP

macedogm:
Add `restricted-noroot` PSP policy to match upstream Kubernetes `restricted` PSP.

Fix https://github.com/rancher/rancher/issues/35191

### DIFF
--- a/pkg/data/management/podsecuritypolicytemplate_data.go
+++ b/pkg/data/management/podsecuritypolicytemplate_data.go
@@ -13,6 +13,7 @@ import (
 
 const runAsAny = "RunAsAny"
 const mustRunAs = "MustRunAs"
+const mustRunAsNonRoot = "MustRunAsNonRoot"
 
 func addDefaultPodSecurityPolicyTemplates(management *config.ManagementContext) error {
 	pspts := management.Management.PodSecurityPolicyTemplates("")
@@ -49,7 +50,7 @@ func addDefaultPodSecurityPolicyTemplates(management *config.ManagementContext) 
 					Rule: runAsAny,
 				},
 			},
-			Description: "This is the default unrestricted Pod Security Policy Template.  It is the most permissive " +
+			Description: "This is the default unrestricted Pod Security Policy Template. It is the most permissive " +
 				"Pod Security Policy that can be created in Kubernetes and is equivalent to running Kubernetes " +
 				"without Pod Security Policies enabled.",
 		},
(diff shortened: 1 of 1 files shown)


macedogm:
Docs PR https://github.com/rancher/docs/pull/3849

cmurphy (approved):

pkg/data/management/podsecuritypolicytemplate_data.go
```diff
@@ -94,9 +95,54 @@ func addDefaultPodSecurityPolicyTemplates(management *config.ManagementContext)
...
+				RequiredDropCapabilities: []v1.Capability{
+					"ALL",
+				},
+				Volumes: []v1beta1.FSType{
+					"configMap",
+					"emptyDir",
+					"projected",
+					"secret",
+					"downwardAPI",
+					"csi",
```
MbolotSuse, line 117: Couple of questions here:

1. The [upstream PSP](https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/policy/restricted-psp.yaml) includes access to `'ephemeral'` volumes. Would you mind elaborating on why this was excluded from the PSP that we include here?
2. The `restricted` PSP that we also include seems to not include the CSI volume. I may be missing the point of this PSP, but would we want to change this so that the only difference is the runAsUser rule?

MbolotSuse (commented):
Looking for some clarification on the volumes and the use of the `restricted` PSP that we also provide.

pkg/data/management/podsecuritypolicytemplate_data.go
macedogm: Answers:

1.  'ephemeral' volumes were not included in the new PSP, because it’s not supported in k8s 1.18. As soon as we dropped support for 1.18, we can add it.
```
# podsecuritypolicies.policy "restricted-noroot" was not valid:
# * spec.volumes: Unsupported value: "ephemeral": supported values: "*", "awsElasticBlockStore", 
"azureDisk", "azureFile", "cephFS", "cinder", "configMap", "csi", "downwardAPI", "emptyDir", "fc", 
"flexVolume", "flocker", "gcePersistentDisk", "gitRepo", "glusterfs", "hostPath", "iscsi", "nfs", 
"persistentVolumeClaim", "photonPersistentDisk", "portworxVolume", "projected", "quobyte", "rbd", 
"scaleIO", "secret", "storageos", "vsphereVolume"
```

2. The current `restricted` PSP that we include has some deviations from the `restricted` upstream, as you spot, the biggest one being:

```
				RunAsUser: v1beta1.RunAsUserStrategyOptions{
					Rule: runAsAny,
				},
```

Instead of `mustRunAsNonRoot`. We opted for creating a new "restricted" PSP in order to not break current deployment that are using the provided `restricted` PSP and to be as close as possible to the upstream policy.

macedogm: > is this just because we want to leave the current PSP alone and include any upstream changes in the restricted-noroot policy instead?

Yes, that's correct. We don't want to change the current provided `restricted` PSP. So all the changes were made to the new `restricted-noroot` PSP.

MbolotSuse (approved):
Questions were answered - LGTM.

pkg/data/management/podsecuritypolicytemplate_data.go
```diff
@@ -94,9 +95,54 @@ func addDefaultPodSecurityPolicyTemplates(management *config.ManagementContext)
 				},
 				ReadOnlyRootFilesystem: false,
 			},
-			Description: "This is the default restricted Pod Security Policy Template.  It restricts many user " +
+			Description: "This is the default restricted Pod Security Policy Template. It restricts many user " +
 				"actions and does not allow privilege escalation.",
 		},
+		{
+			ObjectMeta: v12.ObjectMeta{
+				Name: "restricted-noroot",
```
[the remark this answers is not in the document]
cbron: Ok we can leave as-is then.

Issue #35436, opened 2021-11-10
Support for RKE2 secrets encryption keys rotation


snasovich:
Implement support to rotate secrets encryption keys. This will  allow the UI equivalent to the following RKE1 UI to be implemented:

Rotate encryption keys:
![image](https://user-images.githubusercontent.com/85187633/141045220-b2f8b75e-ad81-4804-9c3f-23712ec5b51a.png)

This requires https://github.com/rancher/rke2/issues/748 implemented on RKE2 side.

jakefhyde:
## How to rotate encryption keys

Although there will be a UI button to start rotating encryption keys before the `v2.6.4` release, currently the encryption keys can be rotated by adding the following:
```
rotateEncryptionKeys:
  generation: 1
```
under the `rkeConfig` section of a cluster object's yaml.

## Additional test cases

- single node rke2/k3s
- 3 node control plane rke2/k3s
- creating the cluster with `rotateEncryptionKeys.generation` in the spec (cluster should not rotate while provisioning).
- \>5 node control plane

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head
Rancher cluster type: HA
Docker version: 20.10

Downstream cluster type: RKE2 node driver and K3s node driver

----------

### Testing: ###

Tested the following:

Single Node rke2/k3s
1. Create new cluster with encryption enabled (rke2/k3s)
2. Create new cluster and edit cluster and update it and enable encryption (k3s)
3. Disable/Enable encryption (k3s)
4. Add Remove control plane (rke2/k3s)
5. Create a new cluster with custom encryption configuration (rke2/k3s)
6. Create new cluster and edit cluster and update it and enable custom encryption (k3s)
7. Disable encryption (custom encryption) (k3s)
8. Add/Remove control planes (custom encryption) (rke2/k3s)
9. Edit custom encryption (rke2/k3s)
10. Switch from default encryption to custom encryption (rke2/k3s)

The same testing for 3 node control plane rke2/k3s and also for >5 node control plane

Create cluster with `rotateEncryuptionKeys.generation` in the spec

**Result** - Fail

All scenarios failed, spoke with @jakefhyde for all necessary logs and information

Create cluster with `rotateEncryuptionKeys.generation` in the spec - causes worker nodes to be stuck in `Provisioning` state for rke2 and causes etcd nodes to be stuck in `Provisioning` for k3s nodes.

snasovich:
[...]

PR #36373, opened 2022-02-03
Rotate Encryption Keys for RKE2

jakefhyde:
## Description

rke2/k3s support rotating encryption keys for secrets via the cli, which is required for parity with rke1 so v2provisioning can go GA. This is implemented by first calling the correct command on an elected control plane node (in this case, we will use the init node), and then restarting the init node. After the init node has been restarted, all other nodes can be safely restarted. This is according to our [official docs on the rotating encryption secrets with k3s](https://rancher.com/docs/k3s/latest/en/security/secrets_encryption/#high-availability-encryption-key-rotation). If the nodes are out of sync (the usual ran commands/restarted nodes), the planner will alternate between restarting the init node and restarting the other control plane nodes until all nodes match. 

Related Issue: https://github.com/rancher/rancher/issues/35436


pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+	if shouldResetEncryptionStatus(cp) {
+		cp.Status.RotateEncryptionKeysPhase = rkev1.RotateEncryptionKeysPhaseStart
+		_, err := r.rkeControlPlanes.UpdateStatus(cp)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	currentInitNodes := collect(clusterPlan, isInitNode)
```
Oats87, line 50: You shouldn't need to do this, there is a helper function to find the init node in `initnode.go` of the planner

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+			return err
+		}
+		if _, ok := phases[p]; !ok {
+			phases[p] = []*v1.Node{&n}
+		} else {
+			phases[p] = append(phases[p], &n)
+		}
+	}
+
+	// init node changed in between get and list
```
Oats87, line 91: If the init node changes between a get and list we have a huge problem

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil
+	}
+
+	currentInitNodes := collect(clusterPlan, isInitNode)
+	if len(currentInitNodes) != 1 {
+		return fmt.Errorf("cannot rotate encryption keys: cannot find an init node")
+	}
+	initEntry := currentInitNodes[0]
+
+	clientset, err := rke2.ClientSetFromCache(initEntry.Machine.Namespace, initEntry.Machine.Spec.ClusterName, r.rancherClusterCache, r.kubeconfigManager)
```
Oats87, line 56: I'm not convinced we need to/should be generating a clientset in the planner that hits the downstream cluster... can we not scrape data from plan outputs to determine the status of the rotation?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+	// init node changed in between get and list
+	if _, ok := phases[phase]; !ok {
+		initNode, err = clientset.CoreV1().Nodes().Get(context.TODO(), initEntry.Machine.Status.NodeRef.Name, metav1.GetOptions{})
+		phase, err = nodeEncryptionPhase(cp, initNode)
+		if err != nil {
+			return err
+		}
+	}
+
+	if phase == "reencrypt_request" || phase == "reencrypt_active" {
```
Oats87, line 100: these should probably be defined as constants

Oats87 (changes requested):
Stylistically, this is a departure from the established "format" of the way functionality has been implemented in the v2prov planner.

I would be more comfortable if we had uniformity within the planner, the etcdrestore portion is a great example of the way functionality has been implemented today, and makes a very confusing process somewhat less confusing (with adequate comments).

I need to re-familiarize myself with the secrets-encryption feature in Rancher before I can properly scrutinize the procedural logic here; however, I will say that it is extremely important that the secrets encryption process is leaktight/fool proof, otherwise we can end up in very bad situations with non-accessible secrets (thus requiring an etcd snapshot restore etc.)

On first glance, the procedural logic seems to make sense.

Oats87:
I'm not sure that these functions warrant being hidden under their own struct. I had to end up moving the etcd create/etcd restore structs out to being first class in the planner due to the non-accessible variables/functions/etc. and I'm seeing that this might be necessary for this feature as well

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+	initNode, err := clientset.CoreV1().Nodes().Get(context.TODO(), initEntry.Machine.Status.NodeRef.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	phase, err := nodeEncryptionPhase(cp, initNode)
+	if err != nil {
+		return err
+	}
+
+	nodes, err := clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
```
paynejacob, line 70: why not use collect for this?

pkg/apis/rke.cattle.io/v1/controlplane.go
```diff
@@ -44,15 +45,27 @@ var (
 	ETCDSnapshotPhaseFailed   ETCDSnapshotPhase = "Failed"
 )
 
+type RotateEncryptionKeysPhase string
+
+var (
+	RotateEncryptionKeysPhaseStart         RotateEncryptionKeysPhase = "Start"
+	RotateEncryptionKeysPhaseApply         RotateEncryptionKeysPhase = "Apply"
+	RotateEncryptionKeysPhaseRestart       RotateEncryptionKeysPhase = "Restart"
+	RotateEncryptionKeysPhaseRestartOthers RotateEncryptionKeysPhase = "RestartOthers"
```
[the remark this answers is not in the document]
thedadams: This one could be `RotateEncryptionKeysPhaseRestartFollowers` and the one above could be `RotateEncrpytionKeysPhaseRestartLeader`

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -491,6 +491,7 @@ func rkeControlPlane(cluster *rancherv1.Cluster) (*rkev1.RKEControlPlane, error)
 			ETCDSnapshotRestore:      cluster.Spec.RKEConfig.ETCDSnapshotRestore.DeepCopy(),
 			ETCDSnapshotCreate:       cluster.Spec.RKEConfig.ETCDSnapshotCreate.DeepCopy(),
 			RotateCertificates:       cluster.Spec.RKEConfig.RotateCertificates.DeepCopy(),
+			RotateEncryptionKeys:     cluster.Spec.RKEConfig.RotateEncryptionKeys.DeepCopy(),
```
thedadams, line 494: I know this isn't really part of your PR, but can we put an `rkeConfig := cluster.Spec.RKEconfig.DeepCopy()` above and use the pieces here without all the `DeepCopy` calls?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
thedadams: Collect is for our plan stuff, not the `v1.Node`s.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+			Command: "systemctl",
+			Args: []string{
+				"restart",
+				rke2.GetRuntimeServerUnit(cp.Spec.KubernetesVersion),
+			},
+		},
+	}
+}
+
+func (r *encryptionKeyRotation) updatePlan(cp *rkev1.RKEControlPlane, entry *planEntry, node *v1.Node, instructions ...plan.Instruction) error {
```
thedadams, line 249: There is already an `UpdatePlan` method in `store.go`. Can that be used here?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,272 @@
...
+		for k, v := range phases {
+			// skip nodes on same phase as init node, or if all other nodes are the same phase.
+			if k == phase || len(v) == len(nodes.Items) - 1 {
+				continue
+			}
+			// restart nodes that are not in sync
+			for _, entry := range collect(clusterPlan, isControlPlaneAndNotInitNode) {
+				if k == phase {
+					continue
+				}
```
thedadams, line 133: This is checked above. Can this be removed?

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -206,3 +213,17 @@ func CopyMapWithExcludes(destination map[string]string, source map[string]string
 		}
 	}
 }
+
+func ClientSetFromCache(namespace, clusterName string, rancherClusterCache ranchercontrollers.ClusterCache, kubeconfigManager *kubeconfig.Manager) (*kubernetes.Clientset, error) {
```
thedadams, line 217: I believe this code is also used in `machinenodelookup`. Can we use this function there too?

thedadams (changes requested):

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
paynejacob: ah missed that

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -11,6 +11,7 @@ import (
 	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1/plan"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
+
```
Oats87, line 14: nit: remove newline

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1/plan"
+	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
+	"github.com/rancher/wrangler/pkg/generic"
+	"github.com/sirupsen/logrus"
+	capi "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+const (
+	Start             = "start"
```
Oats87, line 18: Given that these constants are under the planner package, can we rename them to indicate that they're for encryption key rotation to help prevent confusion?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+	SecretsEncryptStatusCommand = "secrets-encrypt-status-command"
+)
+
+func (p *Planner) RotateEncryptionKeys(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan) error {
+	if !shouldRotateKeys(cp) {
+		return nil
+	}
+
+	if shouldRestartEncryptionKeyRotation(cp) {
+		clusterPlan.Cluster.Annotations[capi.PausedAnnotation] = "true"
```
Oats87, line 36: Are we safe to start manipulating the passed in `clusterPlan` object without a `DeepCopy`?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+		if err != nil {
+			return p.enqueueAndSkip(cp)
+		}
+		return p.restartEncryptionKeyRotation(cp)
+	}
+	_, _, leader, err := p.findInitNode(cp, clusterPlan)
+	if err != nil {
+		return err
+	}
+	if leader == nil {
```
Oats87, line 48: An error message/log message for this would probably be helpful, otherwise we may run into situations where we don't know why the planner isn't progressing.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	err = p.updateControlPlanePhase(cp, rkev1.RotateEncryptionKeysPhaseStart)
+	if err != nil {
+		return err
+	}
+	return generic.ErrSkip
+}
+
+func (p *Planner) enqueueAndSkip(cp *rkev1.RKEControlPlane) error {
```
Oats87, line 253: Can we put this in a more "central" location for the planner?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+		Args: []string{
+			"restart", rke2.GetRuntimeServerUnit(cp.Spec.KubernetesVersion),
+		},
+		Env: []string{
+			lastKnownLeaderPhaseEnv,
+		},
+	}
+}
+
+func (p *Planner) applyAndWait(cp *rkev1.RKEControlPlane, entry *planEntry, node plan.NodePlan) error {
```
Oats87, line 395: I'm not sure why this function needs to be defined this way, you can probably just directly call `assignAndCheckPlan`

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+		phase, err = phaseFromPeriodic(leader)
+		if err != nil {
+			return p.enqueueIfErrWaiting(cp, err)
+		}
+		apply, err := secretsEncryptInstruction(cp, phase)
+		if err != nil {
+			return p.enqueueIfErrWaiting(cp, err)
+		}
+
+		logrus.Infof("Rotate encryption keys cluster [%s]: running apply command: [%s]", cp.Spec.ClusterName, apply.Args[1]) // todo Debug
```
Oats87, line 211: is this supposed to be a `logrus.Debugf`?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+}
+
+// enqueueIfErrWaiting will enqueue the control plan if err is ErrWaiting, otherwise err is returned.
+// Some plan store functions as well as internal functions can return ErrWaiting, for instance when the plan has been
+// applied, but output is not yet available for periodic status.
+func (p *Planner) enqueueIfErrWaiting(cp *rkev1.RKEControlPlane, err error) error {
+	if err != nil {
+		w := ErrWaiting("")
+		if errors.As(err, &w) {
+			logrus.Infof("Rotate encryption keys cluster [%s]: %s", cp.Spec.ClusterName, w.Error()) // todo Debug
```
Oats87, line 407: nit: debugf?

pkg/provisioningv2/rke2/planner/initnode.go
```diff
@@ -65,35 +65,35 @@ func (p *Planner) findAndDesignateFixedInitNode(rkeControlPlane *rkev1.RKEContro
...
-		return true, entries[0].Metadata.Annotations[rke2.JoinURLAnnotation], generic.ErrSkip
+		return true, entries[0].Metadata.Annotations[rke2.JoinURLAnnotation], entries[0], generic.ErrSkip
 	}
 	logrus.Debugf("rkecluster %s/%s: designated init node %s found", rkeControlPlane.Namespace, rkeControlPlane.Spec.ClusterName, fixedMachineID)
-	return true, entries[0].Metadata.Annotations[rke2.JoinURLAnnotation], nil
+	return true, entries[0].Metadata.Annotations[rke2.JoinURLAnnotation], entries[0], nil
 }
 
 // findInitNode searches the given cluster for the init node. It returns a bool which is whether an init node was
 // found, the init node join URL, and an error for a few conditions, i.e. if multiple init nodes were found or if there
```
Oats87, line 93: Can you update the godoc comment to reflect that the function is now returning the node entry as well for the found init node?

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -216,7 +225,12 @@ func (p *Planner) Process(controlPlane *rkev1.RKEControlPlane) error {
 		return err
 	}
 
-	if err := p.certificateRotation.RotateCertificates(controlPlane, plan); err != nil {
+	if err = p.certificateRotation.RotateCertificates(controlPlane, plan); err != nil {
+		return err
+	}
+
+	err = p.RotateEncryptionKeys(controlPlane, plan)
```
Oats87, line 232:
```suggestion
	if err = p.RotateEncryptionKeys(controlPlane, plan); err != nil {
	    return err
	}
```

Oats87 (changes requested):
On first pass, overall this looks great; however, I think in its current state merging it may cause confusion for the next person to stumble across the procedural logic here.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,450 @@
...
+// restartInstruction generates a restart command for the rke2/k3s server, using the last known leader phase in order to
+// ensure that non-init nodes have a refreshed plan if the leader phase changes. Even if commands were not run on the
+// init node, 'flapping' between restarting the leader and followers is intended until all nodes are on the same status.
+func restartInstruction(cp *rkev1.RKEControlPlane, entry *planEntry, leaderPhase string) plan.OneTimeInstruction {
+	var lastKnownLeaderPhaseEnv string
+	switch entry.Metadata.Labels[rke2.CattleOSLabel] {
+	case windows:
+		lastKnownLeaderPhaseEnv = fmt.Sprintf("$env:LAST_KNOWN_LEADER_PHASE=%s", leaderPhase)
+	default:
+		lastKnownLeaderPhaseEnv = fmt.Sprintf("LAST_KNOWN_LEADER_PHASE=%s", leaderPhase)
```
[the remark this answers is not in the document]
jakefhyde: Updated comment for clarity.

jakefhyde:
The machine health checks definitely interfered with some things (also the default unhealthy node timeout being 0s caused some issues, going to open a UI issue for that), addressed review comments but need to bump down enqueue interval from 10 seconds to 2, and try `killall.sh` again.

paynejacob (dismissed):

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -110,18 +116,19 @@ type roleFilter func(*planEntry) bool
 type Planner struct {
 	ctx                           context.Context
 	store                         *PlanStore
-	rkeControlPlanes              rkecontrollers.RKEControlPlaneClient
+	rkeControlPlanes              rkecontrollers.RKEControlPlaneController
```
Oats87, line 119: Out of curiosity, why are we changing to the `RKEControlPlaneController` here instead of the `RKEControlPlaneClient`?

pkg/provisioningv2/rke2/planner/instructions.go
```diff
@@ -54,7 +54,7 @@ func generateInstallInstruction(controlPlane *rkev1.RKEControlPlane, entry *plan
 
 // addInstallInstructionWithRestartStamp will generate an instruction and append it to the node plan that executes the `run.sh` or `run.ps1`
 // from the installer image based on the control plane configuration. It will generate a restart stamp based on the
-// passed in configuration to determine whether it needs to start/restart the service being managed.
+// passed in configuration to determine whether it needs to encryptionKeyRotationStart/restart the service being managed.
```
Oats87, line 57: nit: does this change to the comment make sense?

pkg/controllers/provisioningv2/rke2/provisioningcluster/controller.go
```diff
@@ -279,7 +279,7 @@ func (h *handler) getRKEControlPlaneForCluster(cluster *rancherv1.Cluster) (*rke
 	}
 
 	cp, err := h.rkeControlPlane.Get(capiCluster.Spec.ControlPlaneRef.Namespace, capiCluster.Spec.ControlPlaneRef.Name)
-	if apierror.IsNotFound(err) && cluster.DeletionTimestamp == nil {
+	if apierror.IsNotFound(err) && cluster.DeletionTimestamp != nil {
```
Oats87, line 282: nice catch

Oats87 (approved):

pkg/provisioningv2/rke2/planner/planner.go
jakefhyde: We can only enqueue with the controller, (which also is a client).

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+	if !expected {
+		return p.restartEncryptionKeyRotation(cp)
+	}
+	periodicPhase, err := encryptionKeyRotationPhaseFromPeriodic(leader)
+	if err != nil {
+		return p.enqueueIfErrWaiting(cp, err)
+	}
+	if phase == EncryptionKeyRotationReencryptRequest || phase == EncryptionKeyRotationReencryptActive {
+		if periodicPhase != EncryptionKeyRotationReencryptRequest && periodicPhase != EncryptionKeyRotationReencryptActive && periodicPhase != EncryptionKeyRotationReencryptFinished {
```
thedadams, line 253:
```suggestion
		if periodicPhase != phase && periodicPhase != EncryptionKeyRotationReencryptFinished 
```

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,440 @@
...
+
+	err = p.encryptionKeyRotationUpdateControlPlanePhase(cp, nextPhase)
+	if err != nil {
+		return err
+	}
+	return generic.ErrSkip
+}
+
+func (p *Planner) encryptionKeyRotationUpdateControlPlanePhase(cp *rkev1.RKEControlPlane, phase rkev1.RotateEncryptionKeysPhase) error {
+	cp.Status.RotateEncryptionKeysPhase = phase
```
thedadams, line 287: Add a check here to see if we need to update the `phase`.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,440 @@
...
+		Name:    SecretsEncryptApplyCommand,
+		Command: rke2.GetRuntimeCommand(cp.Spec.KubernetesVersion),
+		Args: []string{
+			"secrets-encrypt",
+			command,
+		},
+	}, nil
+}
+
+func encryptionKeyRotationStatusInstruction(cp *rkev1.RKEControlPlane) plan.OneTimeInstruction {
```
thedadams, line 356: I think `encryptionKeyRotationStatusOneTimeInstruction` is a more descriptive name.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,440 @@
...
+		Args: []string{
+			"restart", rke2.GetRuntimeServerUnit(cp.Spec.KubernetesVersion),
+		},
+		Env: []string{
+			lastKnownLeaderPhaseEnv,
+		},
+	}
+}
+
+func encryptionKeyRotationFindApplyCommand(entry *planEntry) (plan.OneTimeInstruction, error) {
```
thedadams, line 405: Can we add a comment to this to explain what it does? I also think a better name could be found here, but I am having a hard time coming up with a short one.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -0,0 +1,440 @@
...
+	for _, i := range entry.Plan.Plan.Instructions {
+		if i.Name == SecretsEncryptApplyCommand {
+			return i, nil
+		}
+	}
+	return plan.OneTimeInstruction{}, fmt.Errorf("could not find apply command for [%s]", entry.Machine.Name)
+}
+
+// encryptionKeyRotationExpectedPhase returns the expected secrets-encrypt phase extracted from the leader's apply command present in the plan.
+func encryptionKeyRotationExpectedPhase(leaderPhase string, leader *planEntry) (bool, error) {
```
thedadams, line 418: The name of this function implies that it returns the expected status. It seems that it returns whether the phase is what is expected based on the command.
```suggestion
func encryptionKeyRotationPhaseIsExpectedFromCommand(leaderPhase string, leader *planEntry) (bool, error) {
```

thedadams (changes requested):

pkg/provisioningv2/rke2/planner/initnode.go
```diff
@@ -39,19 +39,19 @@ func (p *Planner) setInitNodeMark(entry *planEntry) error {
 
 // findAndDesignateFixedInitNode is used for rancherd where an exact machine (determined by labeling the
 // rkecontrolplane object) is desired to be the init node
-func (p *Planner) findAndDesignateFixedInitNode(rkeControlPlane *rkev1.RKEControlPlane, plan *plan.Plan) (bool, string, error) {
+func (p *Planner) findAndDesignateFixedInitNode(rkeControlPlane *rkev1.RKEControlPlane, plan *plan.Plan) (bool, string, *planEntry, error) {
```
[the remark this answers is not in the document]
jakefhyde: Unfortunately I wasn't able to implement this change, because `findInitNode` can return true with a `nil` init node to signify that there is an existing init node, with an empty join URL.

thedadams (approved):

Oats87 (approved):

PR #36396, opened 2022-02-04
Node and Node Core Metrics

ryansann:
This PR implements two new prometheus gauge metrics: `cluster_manager_nodes` and `cluster_manager_node_cores`. Both metrics are labeled by `cluster_id` and `provider`.

For example, with a one node management cluster w/ 4 cores, a 3 node downstream RKE cluster w/ 2 cores per node, and a 1 node downstream RKE2 cluster w/ 2 cores, we should see something like this reported in the Rancher `/metrics` endpoint:
```
# HELP cluster_manager_node_cores Number of node cores in ready clusters
# TYPE cluster_manager_node_cores gauge
cluster_manager_node_cores{cluster_id="c-6zt8w",provider="rke"} 6
cluster_manager_node_cores{cluster_id="c-m-w5mpglzf",provider="rke2"} 2
cluster_manager_node_cores{cluster_id="local",provider="k3s"} 4
# HELP cluster_manager_nodes Number of nodes in ready clusters
# TYPE cluster_manager_nodes gauge
cluster_manager_nodes{cluster_id="c-6zt8w",provider="rke"} 3
cluster_manager_nodes{cluster_id="c-m-w5mpglzf",provider="rke2"} 1
cluster_manager_nodes{cluster_id="local",provider="k3s"} 1
``` 

The node/node core metrics are enabled by the `prometheus-metrics` setting (`export CATTLE_PROMETHEUS_METRICS=true`)

Every `nodePollInterval` the `collect` go routine retrieves clusters and their nodes from cache. Then it creates `nodeInfo` objects for each node. The `nodeInfo` objects are then used to report node and node core metrics to the prometheus client

Currently, longhorn nodes aren't counted. 

Issue:
- https://github.com/rancher/rancher/issues/36410

### DIFF
--- a/pkg/metrics/node.go
+++ b/pkg/metrics/node.go
@@ -1,0 +1,179 @@
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	mgmtcontrollers "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	"github.com/rancher/wrangler/pkg/ticker"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	nodeProviderLabel  = "provider"
+	nodeClusterIDLabel = "cluster_id"
+	nodeIsActiveLabel  = "is_active"
+)
+
+var (
+	nodeLabels = []string{nodeProviderLabel, nodeClusterIDLabel, nodeIsActiveLabel}
+	numNodes   = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: "cluster_manager",
+			Name:      "nodes",
+			Help:      "Number of nodes in rancher managed clusters",
+		}, nodeLabels,
+	)
+	numCores = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: "cluster_manager",
+			Name:      "node_cores",
+			Help:      "Number of node cores in rancher managed clusters",
+		}, nodeLabels,
+	)
+)
+
+const (
+	reportInterval = time.Duration(15 * time.Second)
+	logPrefix      = "[prometheus-node-metrics]"
+)
+
+type nodeMetrics struct {
+	nodeCache    mgmtcontrollers.NodeCache
+	clusterCache mgmtcontrollers.ClusterCache
+}
+
+func (m *nodeMetrics) collect(ctx context.Context) {
+	for range ticker.Context(ctx, reportInterval) {
+		logrus.Debugf("%s collecting nodes to report metrics", logPrefix)
+
+		nodes, err := m.nodeCache.List("", labels.Everything())
+		if err != nil {
+			logrus.Errorf("%s couldn't list v3.Nodes: %v", logPrefix, err)
+			continue
+		}
+
+		var infos []*nodeInfo
+		for _, node := range nodes {
+			info, err := m.getNodeInfo(node)
+			if err != nil {
+				logrus.Debugf("%s could not determine node info: %v", logPrefix, err)
+				continue
+			}
+			infos = append(infos, info)
+		}
+
+		setMetrics(infos)
+	}
+
+	logrus.Debugf("%s context cancelled, exiting", logPrefix)
+}
+
+type nodeLabelValues struct {
+	provider  string
+	clusterID string
+	isActive  bool
+}
+
+type nodeInfo struct {
+	coreCount   int
+	labelValues nodeLabelValues
+}
+
+const (
+	unknownProvider = "unknown"
+)
+
+// getNodeInfo uses the v3.Node object and it's corresponding v3.Cluster object to get metric info for a node
+func (m *nodeMetrics) getNodeInfo(node *v3.Node) (*nodeInfo, error) {
+	cluster, err := m.clusterCache.Get(node.ObjClusterName())
+	if err != nil {
+		return nil, fmt.Errorf("error getting cluster %s for node %s, %v", node.ObjClusterName(), node.Name, err)
+	}
+	if cluster == nil {
+		return nil, fmt.Errorf("could not get associated cluster for node: %s/%s", node.Namespace, node.Name)
+	}
+	if cluster.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("associated cluster for node: %s/%s is being deleted", node.Namespace, node.Name)
+	}
+
+	// a node is considered to be active if it is ready and it belongs to a ready cluster
+	isActive := v32.ClusterConditionReady.IsTrue(cluster) && v32.NodeConditionReady.IsTrue(node)
+
+	provider := cluster.Status.Provider
+	if provider == "" {
+		provider = unknownProvider
+	}
+
+	return &nodeInfo{
+		coreCount: getNodeCoreCount(node),
+		labelValues: nodeLabelValues{
+			clusterID: node.ObjClusterName(),
+			provider:  provider,
+			isActive:  isActive,
+		},
+	}, nil
+}
+
+func getNodeCoreCount(node *v3.Node) int {
+	nodeCap := node.Status.InternalNodeStatus.Capacity
+	if nodeCap == nil || nodeCap.Cpu() == nil {
+		return 0
+	}
+
+	cores, ok := nodeCap.Cpu().AsInt64()
+	if !ok {
+		return 0
+	}
+
+	return int(cores)
+}
+
+// setMetrics uses a slice of nodeInfo to update prometheus metrics for nodes and node cores
+func setMetrics(infos []*nodeInfo) {
+	// count nodes and node cores, bucketed by label values
+	nodeGauges := make(map[nodeLabelValues]int, 0)
+	nodeCoreGauges := make(map[nodeLabelValues]int, 0)
+	for _, info := range infos {
+		key := info.labelValues
+		_, ok := nodeGauges[key]
+		if !ok {
+			nodeGauges[key] = 1
+		} else {
+			nodeGauges[key]++
+		}
+
+		cores, ok := nodeCoreGauges[key]
+		if !ok {
+			nodeCoreGauges[key] = info.coreCount
+		} else {
+			nodeCoreGauges[key] = cores + info.coreCount
+		}
+	}
+
+	// use the built nodeGagues map to update prometheus metrics for nodes
+	for k, v := range nodeGauges {
+		l := prometheus.Labels{
+			nodeProviderLabel:  k.provider,
+			nodeClusterIDLabel: k.clusterID,
+			nodeIsActiveLabel:  strconv.FormatBool(k.isActive),
+		}
+		numNodes.With(l).Set(float64(v))
+	}
+
+	// use the build nodeCoreGauges map to update prometheus metrics for node cores
+	for k, v := range nodeCoreGauges {
+		l := prometheus.Labels{
+			nodeProviderLabel:  k.provider,
+			nodeClusterIDLabel: k.clusterID,
+			nodeIsActiveLabel:  strconv.FormatBool(k.isActive),
+		}
+		numCores.With(l).Set(float64(v))
+	}
+}
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -222,10 +222,19 @@ func Register(ctx context.Context, scaledContext *config.ScaledContext) {
 	// Cluster Owner
 	prometheus.MustRegister(clusterOwner)
 
+	// node and node core metrics
+	prometheus.MustRegister(numNodes)
+	prometheus.MustRegister(numCores)
+
 	gc := metricGarbageCollector{
 		clusterLister:  scaledContext.Management.Clusters("").Controller().Lister(),
 		nodeLister:     scaledContext.Management.Nodes("").Controller().Lister(),
 		endpointLister: scaledContext.Core.Endpoints(settings.Namespace.Get()).Controller().Lister(),
+	}
+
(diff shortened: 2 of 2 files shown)


pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+)
+
+const (
+	nodeProviderLabel  = "provider"
+	nodeClusterIDLabel = "cluster_id"
+)
+
+var (
+	nodeLabels = []string{nodeProviderLabel, nodeClusterIDLabel}
```
MbolotSuse, line 25: Clarification question - the original issue (https://github.com/rancher/rancher/issues/36410) specifies that the metrics should also expose status but i don't see us collecting that here - would that be something we want to add?

MbolotSuse (changes requested):
I think the overall structure is good. Most of my concerns revolve around wait/retry periods and 2 label fields (provider and status).

pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	if cluster.Status.Provider == "" {
+		return nil, fmt.Errorf("unrecognized provider for node: %s/%s", node.Namespace, node.Name)
+	}
+
+	return &nodeInfo{
+		coreCount: getNodeCoreCount(node),
+		labelValues: nodeLabelValues{
+			clusterID: node.Namespace,
```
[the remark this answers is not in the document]
ryansann: Sure I'll make this change

pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		return true, nil
+	})
+
+	return nodes, err
+}
+
+type nodeLabelValues struct {
+	provider  string
```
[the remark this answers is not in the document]
ryansann: Yea, so the metrics themselves are labeled by `provider` and `cluster_id`, this is because downstream of the metrics we need to know how many nodes there are per provider, example of a provider label values.
```
gke
aks
eks
rke
rke2
rke.windows
```
The `cluster_id` is there to be able to distinguish management nodes from downstream nodes

pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+func (m *nodeMetrics) listNodes() ([]*v3.Node, error) {
+	var nodes []*v3.Node
+	backoff := wait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   1.5,
+		Steps:    10,
+		Cap:      time.Duration(30 * time.Second),
+	}
+
+	err := wait.ExponentialBackoff(backoff, func() (done bool, err error) {
```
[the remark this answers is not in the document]
ryansann: Good call, I was doing this before I switched to node and cluster caches, I think I can safely remove this since I'm using caches now

pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+	logrus.Debugf("[node-metrics] context cancelled, exiting")
+}
+
+func (m *nodeMetrics) listNodes() ([]*v3.Node, error) {
+	var nodes []*v3.Node
+	backoff := wait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   1.5,
+		Steps:    10,
+		Cap:      time.Duration(30 * time.Second),
```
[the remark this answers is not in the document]
ryansann: See previous comment, will remove backoff retry since caches are being used

pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+		prometheus.GaugeOpts{
+			Subsystem: "cluster_manager",
+			Name:      "node_cores",
+			Help:      "Number of node cores in ready clusters",
+		}, nodeLabels,
+	)
+)
+
+const (
+	nodePollInterval = time.Duration(15 * time.Second)
```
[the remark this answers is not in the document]
ryansann: I considered making the node poll interval a setting, but that I opted not to for a few reasons:
1. The node count and core counts won't change all too often. Scaling clusters, creating new clusters, removing clusters, etc are the operations that would cause these metric values to change
2. The time series functionality for this metric, i.e. seeing how it changed over time is less important than the point in time value
3. User would have control if it's a setting and nothing stopping them from setting it to a very large value

The metrics don't need to be available right away, but they do need to update often enough to give a clear picture of node usage. So for instance we shouldn't wait minutes / hours before returning accurate node and node core count metrics

pkg/metrics/node.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, fmt.Errorf("error getting cluster %s for node %s, %v", node.Namespace, node.Name, err)
+	}
+	if cluster == nil {
+		return nil, nil
+	}
+	if cluster.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	if !v32.ClusterConditionReady.IsTrue(cluster) {
```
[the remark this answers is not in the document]
ryansann: This is something I went back and forth on, but ultimately landed on this as making the most sense.
If this wasn't here and a cluster was unreachable, we would report nodes for that cluster. Since these metrics are basically nodes under management (`cluster_manager_nodes`), I reasoned that those nodes aren't under management for cases like this.

I could see the case for considering nodes from all clusters, but then we'd likely need an additional `status` label or something to indicate whether or not the node / its cluster is active (among other statuses). This complicates the collection a bit as it's hard to nail down a set of fixed statuses for nodes from the cluster/node conditions.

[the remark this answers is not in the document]
ryansann: @rmweir maybe you have an opinion on this?

I could add a boolean label, something like: `is_active` that is true if the `ClusterConditionReady.IsTrue()` and `NodeConditionReady.IsTrue()` and otherwise gets false. Then we'd consider all nodes for metrics

pkg/metrics/node.go
MbolotSuse: Resolving in favor of other discussions on this topic.

pkg/metrics/node.go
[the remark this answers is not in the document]
MbolotSuse: It makes sense to not allow messing with this setting. I was more thinking that we would want to allow users to control the rate so that we don't end up causing usage issues with the endpoints that it is using. But this is infrequent enough to not make an impact, and if we need them at a certain rate it makes sense to not allow adjustments.

pkg/metrics/node.go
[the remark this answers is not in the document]
MbolotSuse: Ok - so the provider is something that we need to meet the requirements as outlined in the issue. Thanks for explaining.

pkg/metrics/node.go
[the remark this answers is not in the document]
ryansann: I added the `is_active` label

pkg/settings/setting.go
```diff
@@ -119,6 +119,7 @@ var (
 	HideLocalCluster                    = NewSetting("hide-local-cluster", "false")
 	MachineProvisionImage               = NewSetting("machine-provision-image", "rancher/machine:v0.15.0-rancher73")
 	SystemFeatureChartRefreshSeconds    = NewSetting("system-feature-chart-refresh-seconds", "900")
+	PrometheusMetrics                   = NewSetting("prometheus-metrics", "false")
```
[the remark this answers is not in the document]
rmweir: using env var also makes it so the deployment needs to restart which makes it more effective, otherwise since you only gate the register, changing the setting value would have no effect anyway. It's also worth noting that once the setting val is set with an env var the value cannot be over written anyway.

rmweir (changes requested):

pkg/metrics/node.go
```diff
@@ -0,0 +1,179 @@
...
+	if cluster == nil {
+		return nil, fmt.Errorf("could not get associated cluster for node: %s/%s", node.Namespace, node.Name)
+	}
+	if cluster.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("associated cluster for node: %s/%s is being deleted", node.Namespace, node.Name)
+	}
+
+	var isActive bool
+	// node conditions aren't reliably updated when node is unreachable so use the cluster ready condition
+	// if the cluster is ready the its nodes are considered to be active
```
MbolotSuse, line 105: Nit:

`then` is misspelled as `the` in `if the cluster is ready the its nodes`

pkg/metrics/node.go
```diff
@@ -0,0 +1,179 @@
...
+	if v32.ClusterConditionReady.IsTrue(cluster) {
+		isActive = true
```
MbolotSuse, line 107: I know that we wanted to avoid making this logic overly complex. That being said, do we want to do some sort of sanity check on the node so that both the node and the cluster are ready? I'm thinking something like:
```
isActive = v32.ClusterConditionReady.IsTrue(cluster) && node.Status.InternalNodeStatus.Phase == types.NodeRunning
```
You could also loop through the node conditions and make sure that there is a condition of type `NodeReady`, but I think that's a bit more complex than you were looking for (looks like conditions are current, so that wouldn't be too bad, but still).

pkg/metrics/node.go
```diff
@@ -0,0 +1,179 @@
...
+		prometheus.GaugeOpts{
+			Subsystem: "cluster_manager",
+			Name:      "node_cores",
+			Help:      "Number of node cores in rancher managed clusters",
+		}, nodeLabels,
+	)
+)
+
+const (
+	reportInterval = time.Duration(1 * time.Minute)
```
MbolotSuse, line 43: Is this interval ok for our AWS integration? Generally, I would think that we would want to know quicker than 1 min if a node went down (10/15 seconds wasn't a bad number). But I don't have a strong reason on why this is too slow, so it's fine if we want to keep it this way.

MbolotSuse (changes requested):
Few minor issues surrounding the time interval and isActive status/condition.

pkg/settings/setting.go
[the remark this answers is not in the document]
ryansann: Ahh ok, good points, I'll change remove the setting and switch back to the `CATTLE_PROMETHEUS_METRICS` env var

pkg/metrics/node.go
ryansann: will fix

pkg/metrics/node.go
ryansann: I initially had that, but the node doesn't set it's ready condition to false when expected.

When I checked both I never saw a false value come through when bringing down nodes removing agents, etc. So in effect this was like not having the label at all (since the value was always `true`)

Worth mentioning that all nodes in a given cluster will either all be active or none with this change though. I think that's acceptable and I'm also getting provider from the cluster object

pkg/metrics/node.go
ryansann: this was not meant to change, thank you for catching that.

pkg/metrics/node.go
ryansann: One additional note on this. When provisioning a cluster, node metrics for nodes in that cluster won't show until the cluster is provisioned. The reason for this is that the `provider` comes from the `cluster.Status.Provider` field, which doesn't get populated until provisioned. If there is no provider, we don't update node metrics. 

Another possible option to include these nodes would be to add an "unknown" provider value count nodes from clusters that don't have a provider in their status (yet)

ryansann: I made the change to include an `unknown` provider label value when the cluster doesn't have the provider value yet. This allows provisioning nodes to be captured in the metrics

ryansann: I've updated this to check the node too

pkg/metrics/node.go
```diff
@@ -0,0 +1,182 @@
...
+type nodeMetrics struct {
+	nodeCache    mgmtcontrollers.NodeCache
+	clusterCache mgmtcontrollers.ClusterCache
+}
+
+func (m *nodeMetrics) collect(ctx context.Context) {
+	for range ticker.Context(ctx, reportInterval) {
+		logrus.Debugf("%s collecting nodes to report metrics", logPrefix)
+
+		nodes, err := m.nodeCache.List("", labels.NewSelector())
```
MbolotSuse, line 56: Question here: does labels.NewSelector() return a selector that matches everything (so same as labels.Everything())? That would be my guess but I'd like to be sure.

MbolotSuse (commented):
Some minor issues/clarifications.

pkg/metrics/node.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if cluster == nil {
+		return nil, fmt.Errorf("could not get associated cluster for node: %s/%s", node.Namespace, node.Name)
+	}
+	if cluster.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("associated cluster for node: %s/%s is being deleted", node.Namespace, node.Name)
+	}
+
+	// a node is considered to be active if it is ready and it belongs to a ready cluster
+	var isActive bool
```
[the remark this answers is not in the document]
ryansann: The default for basic types is the zero value. I think that's pretty basic Go knowledge and will never change. But I do like your suggestion for skipping the if condition all together so I'll update.

pkg/metrics/node.go
ryansann: Yea, but `labels.Everything()` conveys better intent, I'll update

MbolotSuse (approved):
LGTM - Thank you for bearing with me and looking over these changes thoroghly.

rmweir (approved):
LGTM

pkg/metrics/node.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	if cluster == nil {
+		return nil, fmt.Errorf("could not get associated cluster for node: %s/%s", node.Namespace, node.Name)
+	}
+	if cluster.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("associated cluster for node: %s/%s is being deleted", node.Namespace, node.Name)
+	}
+
+	// a node is considered to be active if it is ready and it belongs to a ready cluster
+	isActive := v32.ClusterConditionReady.IsTrue(cluster) && v32.NodeConditionReady.IsTrue(node)
```
cbron, line 107: From last session, this won't actually matter during calculations, but should be fine to present via endpoint.

PR #36398, opened 2022-02-04
Ensure node-cleanup pods are deleted

bentastic27:
A customer is hitting an issue where nodes deleted from Cluster Explorer don't get their cleanup jobs run on them so the jobs-pods just get stuck in pending. The nodes seem to be removed from the cluster before the Job is created so the node scheduling never resolves. I've noticed that the Job gets deleted, but the pod does not.

The problem for them is that they re-add nodes with the same name as part of their patching procedure. So when they do the pod gets scheduled to that node and cleaned. This PR adds a step to clean the nodes belonging to the Job after the fact.

Issue:
https://github.com/rancher/rancher/issues/35334

### DIFF
--- a/pkg/controllers/management/node/cleanup.go
+++ b/pkg/controllers/management/node/cleanup.go
@@ -201,7 +201,7 @@ func (m *Lifecycle) waitUntilJobCompletes(userContext *config.UserContext, job *
 	}
 
 	// remove the job to clean up
-	return userContext.K8sClient.BatchV1().Jobs(job.Namespace).Delete(m.ctx, job.Name, metav1.DeleteOptions{})
+	return userContext.K8sClient.BatchV1().Jobs(job.Namespace).Delete(m.ctx, job.Name, metav1.DeleteOptions{PropagationPolicy: &[]metav1.DeletionPropagation{metav1.DeletePropagationForeground}[0]})
 }
 
 func (m *Lifecycle) createCleanupJob(userContext *config.UserContext, node *v3.Node) (*batchV1.Job, error) {


pkg/controllers/management/node/cleanup.go
```diff
@@ -201,7 +201,15 @@ func (m *Lifecycle) waitUntilJobCompletes(userContext *config.UserContext, job *
...
-	return userContext.K8sClient.BatchV1().Jobs(job.Namespace).Delete(m.ctx, job.Name, metav1.DeleteOptions{})
+	logrus.Infof("[node-cleanup] Cleaning job %s and related pods", job.Name)
+	if err := userContext.K8sClient.BatchV1().Jobs(job.Namespace).Delete(m.ctx, job.Name, metav1.DeleteOptions{}); err == nil {
+		logrus.Debugf("[node-cleanup] Deleting pods for job %s", job.Name)
+		return userContext.K8sClient.CoreV1().Pods(job.Namespace).DeleteCollection(m.ctx, metav1.DeleteOptions{}, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("job-name=%s", job.Name),
+		})
+	} else {
+		return err
+	}
```
thedadams, line 212: It turns out that all you have to do is tell Kubernetes that you want to use `ForegroundDeletion`. Then the pod will get cleaned up before the job does (and the problem with the owner reference disappearing isn't a problem). It a little complicated because of pointers, but here it is.
```suggestion
	return userContext.K8sClient.BatchV1().Jobs(job.Namespace).Delete(m.ctx, job.Name, metav1.DeleteOptions{PropagationPolicy: &[]metav1.DeletionPropagation{metav1.DeletePropagationForeground}[0]})
```

thedadams (changes requested):

pkg/controllers/management/node/cleanup.go
bentastic27: Much cleaner @thedadams. Thanks!

thedadams:
@bentastic27 Can you squash these commits, please?

thedadams (approved):

paynejacob (approved):

Issue #31676, opened 2021-03-12
Adding a clusterrepos.catalog.cattle.io with custom certificate in a dowstream repo is failing.


zaggash:
SURE-3086

**BUG**

**Steps to reproduce (least amount of steps as possible):**
* Start a Rancher cluster in version 2.5.x
* Add a downstream cluster
* Open the Cluster Explorer on the downstream cluster
* Add a custom HTTPS Chart repo with a custom certificate in the Rancher downstream cluster

**Result:**

 * The Chart Repository status stays at `Downloading`.

```
$ k -n cattle-system describe clusterrepos.catalog.cattle.io my-chart
[...]
Status:
  Conditions:
    Last Update Time:                 2021-03-12T10:47:44Z
    Status:                           True
    Type:                             FollowerDownloaded
    Last Update Time:                 2021-03-12T13:39:32Z
    Message:                          Get "https://138.68.84.135:443/index.yaml": x509: certificate signed by unknown authority
    Reason:                           Error
    Status:                           False
    Type:                             Downloaded
[...]
```

**Other details that may be helpful:**

If we add the same Chart Repo in the Rancher cluster it is falling too.

Adding the `tls-additionnal-ca` secret fixes the issue for the Rancher cluster but not for the downstream cluster.

**Workaround**  

* Edit the chart with the UI with "Edit YAML"  
   Then add the key/value `spec.insecureSkipTLSVerify: true`

* We can also add the CA like below.  
  The caBundle key is a base64 encoded DER certificate.  
  `openssl x509 -outform der -in ca.pem | base64 -w0`

Open "Edit YAML":

```
spec:
  forceUpdate: "2021-03-12T12:24:46Z"
  url: https://[chart-DNS-Name]
[...]

PR #36399, opened 2022-02-04
[release/v2.6] Add support in cluster git repos for custom/self-signed certs

PennyScissors:
SURE-3618
Issue: https://github.com/rancher/rancher/issues/31676

Technically, the conversion to PEM is not required since you could encode the contents of a ca.pem file with the header, footer, and line breaks into base64, pass it in as `caBundle` and it will be accepted (git based requires header, footer and line breaks). However, this approach doing the conversion makes setting up git-based repos with a CA bundle match the HTTP-based repo setup from a user standpoint (see [here](https://rancher.com/docs/rancher/v2.6/en/helm-charts/#repositories)). It also allows for re-using the logic to load the chart's icons without any further modifications.

### DIFF
--- a/pkg/catalogv2/git/download.go
+++ b/pkg/catalogv2/git/download.go
@@ -3,6 +3,7 @@ package git
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/pem"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/catalogv2/git/download.go
```diff
@@ -84,14 +85,28 @@ func gitForRepo(secret *corev1.Secret, namespace, name, gitURL string, insecureS
 	if settings.InstallUUID.Get() != "" {
 		headers["X-Install-Uuid"] = settings.InstallUUID.Get()
 	}
+	// convert caBundle to PEM format because git requires correct line breaks, header and footer.
```
superseb, line 88: Do we want to set `insecureSkipTLS` to false if `caBundle` is provided? I think the right thing is to do that if `caBundle` is provided as the expectation is that there is a fully valid certificate chain if provided, not that it still goes through if the user misconfigures this. We could also argue that this is guidance we can provide in the UI but still.

PennyScissors: good catch, it should be set to false

superseb (dismissed):

superseb (approved):

PR #36411, opened 2022-02-05
Don't drain etcd nodes

jakefhyde:
Draining etcd nodes causes issues in RKE2, where some pods are unhealthy. This PR never drains etcd nodes, which is the current approach in RKE1.
Related Issue: #35274

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
+++ b/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
@@ -300,7 +300,8 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 			machineDeploymentLabels[k] = v
 		}
 		machineSpecAnnotations := map[string]string{}
-		if !machinePool.DrainBeforeDelete {
+		// Ignore drain if DrainBeforeDelete is unset or the pool is for etcd nodes
+		if !machinePool.DrainBeforeDelete || machinePool.EtcdRole {
 			machineSpecAnnotations[capi.ExcludeNodeDrainingAnnotation] = "true"
 		}
 


jiaqiluo (dismissed):
LGTM

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -303,6 +303,10 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 		if !machinePool.DrainBeforeDelete {
 			machineSpecAnnotations[capi.ExcludeNodeDrainingAnnotation] = "true"
 		}
+		if machinePool.EtcdRole {
+			logrus.Warnf("Ignoring drain before delete for pool [%s] with etcd role", machinePoolName)
```
thedadams, line 307: This message will be printed a bunch (not just on create). I think a comment here is sufficient. Maybe even combine this if-statement with the one above.

This behavior is something that should go in the docs and be release noted.

thedadams (changes requested):

thedadams (approved):

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -300,7 +300,8 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 			machineDeploymentLabels[k] = v
 		}
 		machineSpecAnnotations := map[string]string{}
-		if !machinePool.DrainBeforeDelete {
+		// Ignore drain before delete etcd nodes
```
jiaqiluo, line 303: hmm, the comment is a bit confusing, could you rephrase it to be more clear? 
```suggestion
		// Ignore drain if DrainBeforeDelete is unset or the pool is for etcd nodes 
```

jiaqiluo (changes requested):

jiaqiluo (approved):

thedadams (approved):

PR #36421, opened 2022-02-07
Added fleet.cattle.io client to the automation framework

igomez06:
**Framework updates**
 Generated fleet client, as well as implemented the hooks for automation cleanup for gitreop and clustergroups.


paynejacob (commented):
Do we need these clients in rancher proper? I though we generated them in the test framework package?

igomez06:
> Do we need these clients in rancher proper? I though we generated them in the test framework package?

So only the v3 clients are generated in our framework. We have been generating the  v1 clients (catalog, and provisioning) in rancher proper from the `rancher/rancher/pkg/codegen/main.go` file, and then extending them to have our session cleanup in the framework. Since we did that with catalog and provisioning, I did the same with fleet. That's really the only reason why.

paynejacob:
@igomez06 how are we using fleet in our testing? this is going to require vendoring fleet to be part of the release process.

paynejacob:
@igomez06 if it were in the framework that would be fine.

igomez06:
@paynejacob Sounds good I'll implement a client that just lives in the framework.

PR #36452, opened 2022-02-08
removing rke2 specific test for getting schedulable nodes

slickwarren:
Jenkins tested. 
Currently, automation special-cases rke2 control plane, but it is no longer the case that CP is schedulable by default. Reverting the code, and looking only for the `NoExecute` and `NoSchedule` taints instead of looking at node roles for schedulability.


rancher-max (approved):
LGTM

bmdepesa (approved):
LGTM

PR #36454, opened 2022-02-09
[Forwardport] skip remove nodes for non-custom cluster

JacieChao:
2.5 issue:
- https://github.com/rancher/rancher/issues/32284

Relevant PR:
- https://github.com/rancher/rancher/pull/35991


JacieChao:
Still wait for fix

PR #36487, opened 2022-02-11
Add log to indicate that dapper download not supported on M1 Macs

thedadams:


Oats87 (approved):

snasovich (approved):

Makefile
```diff
@@ -1,11 +1,16 @@
...
 
 .dapper:
-	@echo Downloading dapper
-	@curl -sL https://releases.rancher.com/dapper/latest/dapper-`uname -s`-`uname -m` > .dapper.tmp
-	@@chmod +x .dapper.tmp
-	@./.dapper.tmp -v
-	@mv .dapper.tmp .dapper
+	@if [[ `uname -s` = "Darwin" && `uname -m` = "arm64" ]]; then\
+		echo "Dapper download is not supported on ARM Macs, you need to build it and add it as .dapper in this directory";\
+		exit 0;\
```
snasovich, line 6: Had a thought after I hit the "Approve" button. Should we exit with non-0 here as there is an error?

PR #36493, opened 2022-02-11
Adding node metrics view capability to cluster members

MbolotSuse:
Changing default permissions for cluster members to include
permission to read node/nodemetrics in the metrics.k8s.io group


guangbochen:
Thanks @MbolotSuse, I think it makes sense to not grant non-core API permissions to the cluster members by default, or at least it should manage a feature toggle if this is something that needs to be implemented on the Rancher side. We will take an investigation of whether the Harvester/Longhorn team can implement/modify the RBAC rules on our side during the installation hook. And can we just have the core node API metrics get supported by the cluster member by default first, thanks?

ryansann (approved):

maxsokolovsky (approved):

PR #36495, opened 2022-02-11
Ec2 node library and custom cluster tests

igomez06:
**Framework updates:**
- EC2 client for the aforementioned creation/deletion of nodes.
- EC2 extension for creating and deleting nodes
- Node library to be used by any provider. Including reading from a config for an already existing external node setup.
- Cluster token registration extension to facilitate getting the registration token

**Validation Tests:**
- Custom cluster tests that are not just dependent ec2 node provider.
- Helpers in the v2/validation/provisioning directory to setup the different kind of node providers. Currently only support EC2 and already existing node configuration information in the config file.

### DIFF
--- a/tests/framework/clients/ec2/client.go
+++ b/tests/framework/clients/ec2/client.go
@@ -1,0 +1,38 @@
+package ec2
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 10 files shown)


izaac (approved):
LGTM. I would rely on Jacob's feedback for a final approval.

PR #36497, opened 2022-02-12
python (v3) automation: updating node.taints for schedulable nodes

slickwarren:
must add `continue` to avoid adding the node after breaking from the first loop. 
Tested:
* manually (via python from client machine) -- pass
* remotely (via jenkins freeform job) -- in progress

### DIFF
--- a/tests/validation/tests/v3_api/common.py
+++ b/tests/validation/tests/v3_api/common.py
@@ -664,9 +664,18 @@ def get_schedulable_nodes(cluster, client=None, os_type=TEST_OS):
     schedulable_nodes = []
     for node in nodes:
         if not node.unschedulable:
-            for tkey, tval in node.taints.items():
-                if tval == "NoExecute" or tval == "NoSchedule":
-                    break
+            shouldSchedule = True
+            # node.taints doesn't exist if the node has no taints. 
+            try:
+                for tval in node.taints:
+                    if str(tval).find("PreferNoSchedule") == -1:
+                        if str(tval).find("NoExecute") > -1 or str(tval).find("NoSchedule") > -1:
+                            shouldSchedule=False
+                            break
+            except AttributeError:
+                pass
+            if not shouldSchedule:
+                continue
             for key, val in node.labels.items():
                 # Either one of the labels should be present on the node
                 if key == 'kubernetes.io/os' or key == 'beta.kubernetes.io/os':


tests/validation/tests/v3_api/common.py
```diff
@@ -664,9 +664,16 @@ def get_schedulable_nodes(cluster, client=None, os_type=TEST_OS):
     schedulable_nodes = []
     for node in nodes:
         if not node.unschedulable:
-            for tkey, tval in node.taints.items():
-                if tval == "NoExecute" or tval == "NoSchedule":
-                    break
+            shouldSchedule = True
+            try:
+                for tval in node.taints:
+                    if str(tval).find("NoExecute") > -1 or str(tval).find("NoSchedule") > -1:
```
rancher-max, line 670: Do we want to disallow scheduling if the value is `PreferNoSchedule` then too?

tests/validation/tests/v3_api/common.py
```diff
@@ -664,9 +664,16 @@ def get_schedulable_nodes(cluster, client=None, os_type=TEST_OS):
...
+                        shouldSchedule=False
+                        break
+            except AttributeError:
```
rancher-max, line 673: I think using an exception to get around this is fine, but could get confusing when doing more development down the road if something else causes this exception. Debugging could become harder than intended.

I wonder if, instead, we can use a nested function to return from. Something like:
```
...
for node in nodes:
  if not node.unschedulable:
    def scheduleBasedOnTaints(taints):
      for val in taints:
        # logic, return False if hits break condition
      return True

    shouldSchedule = scheduleBasedOnTaints(node.taints)
    if not shouldSchedule:
      continue
...
```

I'm not sure if this will work or make things better, but just a suggestion in case we don't want to use the `try, Except` block. I am open to suggestions and happy to approve (pending the `PreferNoSchedule` question above) if we prefer to use the `try, Except` block instead 🙂

slickwarren: the `try/except` block is there to catch what happens when node has no attribute `taints`. In testing, I found that if a node has no taints, its not that `node.taints` is null, but that `node` does not even have a `taints` field anymore. Spoke with Max offline about this, and we will keep the try/except block for now.

tests/validation/tests/v3_api/common.py
slickwarren: No, good catch. Updated in this latest version.

rancher-max (approved):
LGTM

sowmyav27 (approved):

Issue #35073, opened 2021-10-07
RKE2 version does not get refreshed upon a refresh of rke-metadata-config 


sowmyav27:
on v2.6-head (7bc7959):
- Update setting `rke-metadata-config` url to `https://raw.githubusercontent.com/kinarashah/kontainer-driver-metadata/21812d7e35cdc5a8fcca3b470751500633dbdcd1/data/data.json`. 
- Refresh Kontainer-driver-metadata from cluster management -> Drivers page. 
- Confirm both RKE2 and k3s have new versions `v1.21.6` -- pass
- Switch the `rke-metadata-config` back to default, k8s versions should now show only `v1.21.5` -- fail
  - also tried hard refershes, but only the k3s version is changing. RKE2 is not updating back to v1.21.5
<img width="640" alt="Screen Shot 2021-10-07 at 10 12 52 AM" src="https://user-images.githubusercontent.com/16691014/136432371-4cb2c3ef-8518-4379-a6fc-c5d209e9fd74.png">

upon further investigation, there still seems to be an issue with the refresh feature. currently, we seem to be in a `partial` refresh state.  The above behavior is observed in the following case:
* update rke-metadata-config
* refresh kontainer-driver-metadata (`cluster management` -> `drivers` -> button) by clicking 1 time

However, if the user clicks the refresh button a second time without leaving the page, both k3s and rke2 are updated to the versions matching what is in rke-metadata-config

Related issue: https://github.com/rancher/rancher/issues/34473

jiaqiluo:
### Root cause

1. RKE2 metadata update is not invoked when editing the rke-metadata-config setting.
2. RKE2 metadata is not updated all the time when refreshing the Kubernetes Metadata manually via UI. This is a little complicated. Basically, we have two goroutines listening to the same channel, which causes a race condition. Each time a value is sent to the channel, only one of those two goroutines will get the value. (Note: this could happen for either k3s or RKE2 although it is reported for RKE2)

### What was fixed, or what changes have occurred

1. Add the missing call for the RKE2-refreshing function in the handler for KDM metadata.
2. Eliminating the race conditional by changing both the channel and how the goroutines handle the value from the channel.

### Areas or cases that should be tested

1. - Edit the rke-metadata-config setting, conform a refreshing of rke2 and k3s metadata happens.
2. - Refreshing the Kubernetes Metadata manually via UI, conform a refreshing of rke2 and k3s metadata happens.
  
### What areas could experience regressions?
 none
[...]

PR #36519, opened 2022-02-14
Fix issues around editing or updating the KDM 

jiaqiluo:
fix the issue where rke2 versions are not updated when either editing the rke-metadata-config setting or refreshing the Kubernetes Metadata manually via UI. GH-35073

### DIFF
--- a/pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
+++ b/pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/types/convert"
+	"github.com/rancher/rancher/pkg/channelserver"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/kontainerdrivermetadata/metadata_handler.go
```diff
@@ -120,6 +121,8 @@ func (m *MetadataController) sync(key string, setting *v3.Setting) (runtime.Obje
 		m.Settings.Controller().EnqueueAfter(setting.Namespace, setting.Name, time.Minute*time.Duration(interval))
 	}
 
+	// refresh to sync k3s/rke2 releases
+	channelserver.Refresh()
```
kinarashah, line 125: make channel is called (https://github.com/rancher/rancher/blob/release/v2.6/pkg/channelserver/channelserver.go#L124) as a part of https://github.com/rancher/rancher/blob/release%2Fv2.6/pkg/multiclustermanager/routes.go#L69 before controllers are started so this looks safe. 👍🏼

kinarashah (approved):

jakefhyde (approved):

Issue #36294, opened 2022-01-27
Can't remove RKE2 cluster if Harvester Cloud Credential is deleted


FrankYang0529:
**Rancher Server Setup**
- Rancher version: v2.6.3
- Installation option (Docker install/Helm Chart): Docker

**Information about the Cluster**
- Kubernetes version: v1.21.8+rke2r2
- Cluster Type (Local/Downstream): Harvester

**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**

Users can remove cloud credentials if there is an RKE2 cluster based on it. After that, we will get an error message in the `machine-provision-remove` handler.

Also, if users only delete the Harvester cluster CR but remain the cloud credential, we can remove the RKE2 cluster without error. However, associated VMs in the Harvester cluster can't be removed together.

**To Reproduce**
1. Import Harvester Cluster.
2. Create a Cloud Credential.
3. Create an RKE2 cluster based on Harvester.
4. Remove the Harvester cloud credential.
5. Remove the RKE2 cluster.

**Result**
RKE2 cluster status is stuck in `Removing` and there is an error message in the `machine-provision-remove` handler.

```
2022/01/27 03:15:33 [ERROR] error syncing 'fleet-default/rke2-in-aws-pool1-7c09ec84-8jv8n': handler machine-provision-remove: secrets "cc-k4t8n" not found, requeuing
```

**Expected Result**
* Prevent users from deleting Cloud Credentials or Harvester cluster CR if there is an RKE2 cluster based on it.
* Remove cluster-agent in Harvester if users remove Harvester cluster CR in Rancher.

**Related issue**
https://github.com/harvester/harvester/issues/1799

FrankYang0529:
@thedadams, I try this case for AWS. There are also error logs like the following.

```
[ERROR] error syncing 'fleet-default/rke2-in-aws-pool1-7c09ec84-msbq6': handler machine-provision-remove: secrets "cc-dbsgw" not found, requeuing
```

Reproduce steps:
1. Create AWS credential.
2. Create an RKE2 cluster on AWS.
3. After the provisioning job is done, delete the credential.
4. Remove the RKE2 cluster.


https://user-images.githubusercontent.com/4927361/154038942-85d70c67-d88f-4f94-a80c-78518c1cbf87.mov

thedadams:
### Root cause
The root cause was that cloud credentials could be deleted even when there was a provisioning cluster that was using the cloud credential.

[...]

PR #36524, opened 2022-02-15
Disallow deleting cloud credential referenced by provisioning cluster

thedadams:
Currently, it is possible to delete a cloud credential if it is referenced by a
provisioning cluster. This causes issues if a user wants to add or delete nodes
after the cloud credential is gone.

After this change, Rancher will prevent a user from deleting a cloud credential
if it is referenced by a provisioning cluster.

Issue:
https://github.com/rancher/rancher/issues/36294

### DIFF
--- a/pkg/api/norman/customization/cred/store.go
+++ b/pkg/api/norman/customization/cred/store.go
@@ -10,13 +10,15 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/rancher/pkg/api/norman/customization/namespacedresource"
+	"github.com/rancher/rancher/pkg/controllers/provisioningv2/cluster"
+	provv1 "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/namespace"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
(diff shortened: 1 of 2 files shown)


kinarashah (approved):

pkg/api/norman/customization/cred/store.go
```diff
@@ -77,9 +80,16 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
...
 	types.Store
 	NodeTemplateLister v3.NodeTemplateLister
+	ProvClusterCache   provv1.ClusterCache
 }
 
 func (s *Store) Delete(apiContext *types.APIContext, schema *types.Schema, id string) (map[string]interface{}, error) {
+	if provClusters, err := s.ProvClusterCache.GetByIndex(cluster.ByCloudCred, id); err != nil {
+		return nil, err
+	} else if len(provClusters) > 0 {
+		return nil, httperror.NewAPIError(httperror.InvalidAction, fmt.Sprintf("cloud credential is currently referenced by provisioning cluster %s", provClusters[0].Name))
```
jiaqiluo, line 90:
```suggestion
		return nil, httperror.NewAPIError(httperror.InvalidAction, fmt.Sprintf("cloud credential is currently referenced by at least one provisioning cluster %s", provClusters[0].Name))
```

jiaqiluo (changes requested):

pkg/api/norman/customization/cred/store.go
thedadams: The message here is consistent with the node template statement below. I would prefer to keep it as is, for consistency sake.

jiaqiluo: I did not realize that. 
It sounds good to me then.

jiaqiluo (approved):

Issue #36488, opened 2022-02-11
[area/harvester] failure to provision with air-gapped environment


dweomer:
**Rancher Server Setup**
- Rancher version:
  - **2.6.3**
- Installation option (Docker install/Helm Chart):
   - **n/a**
- Proxy/Cert Details:
  - **n/a**

**Information about the Cluster**
- Kubernetes version:
  - **1.22.x**
  - **1.21.x**
- Cluster Type (Local/Downstream):
   - **Downstream :arrow_right: Harvester 1.0.0**

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - **admin**

   
**Describe the bug**
- https://github.com/harvester/harvester/issues/1808

> **Expected behavior**
> 
> The driver binaries of "built-in" node drivers are included in the Rancher release and do not have to be downloaded post-install. 
> 
> This is the case for other node drivers of type "built-in", such as AWS, Azure, etc.  The `url` field of these node drivers  is set to `"url": "local://",`
> 
> **Actual behavior**
> 
> In Rancher 2.6.3 the Harvester node driver is marked as "built-in" driver, yet the `url` field is set to a external URL `https://releases.rancher.com/harvester-node-driver/v0.3.4/docker-machine-driver-harvester-amd64.tar.gz`.
> 
> When provisioning a Harvester RKE2 cluster from Rancher, the provisioning will fail on air-gapped Rancher instances since the driver can't be downloaded from the internet.
> 

**Screenshots**
![image](https://user-images.githubusercontent.com/1033444/153657755-8f3211e8-a25b-487d-a156-4eafd5134df3.png)

**Additional context**
- https://github.com/harvester/harvester/issues/1808

thedadams:
### Root cause
rancher-machine downloads non-built-in machine drivers before creating machines. The problem in air-gapped environments is that the rancher-machine pod doesn't have access to the internet.

### What was fixed, or what changes have occurred
Now, when node drivers are downloaded by Rancher, they are also put in a directory that Rancher serves via HTTP. Since the pod can communicate with the Rancher server (it is in the Rancher cluster), then rancher-machine will download the driver from Rancher instead.

### Areas or cases that should be tested
1. Provisioning Harvester RKE2 clusters in air-gapped environment.
2. Provisioning another RKE2 cluster with non-built-in node driver.

### What areas could experience regressions?
Number 2 above

### Are the repro steps accurate/minimal?
[...]

PR #36532, opened 2022-02-16
Allow node driver download in air-gapped environments

thedadams:
The harvester node driver is bundled in Rancher in order to provide support for
air-gapped installs of Rancher to provision Harvester clusters. However, the
download URL passed to the create job was still the external URL. This wouldn't
allow Harvester clusters to be created in air-gapped environments.

After this change, any enabled node driver will be copied to the appropriate
location so that the create job can download it directly from Rancher.

Issue:


Oats87 (approved):

dweomer (approved):
LGTM @thedadams (although it looks as if you are fixing three separate bugs at once!)

Issue #25443, opened 2020-02-17
kontainer-engine and docker-machine zombie processes


niusmallnan:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
You need to access UI and activate a cluster or node driver, which should be a non-built-in driver.

**Result:**
You need to exec to the rancher-server container, then, you will see the zombie processes like:
```
# ps aux |grep Z
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root        53  0.0  0.0      0     0 ?        Z    08:33   0:00 [kontainer-engin] <defunct>
```

**Other details that may be helpful:**
These "zombie" processes are created when a process has finished, but the parent has not read their exit status via the wait system call.

For the node drivers, when determining whether the driver is executable, `cmd.Wait` is not called after `Start`:
https://github.com/rancher/rancher/blob/7892e90064328cf3128dfcb8da75d389037ec82a/pkg/controllers/management/drivers/dynamic_driver.go#L68-L88

For the cluster drivers, we can look at the [kontainer-engine](https://github.com/rancher/kontainer-engine). We can get the same problem in the `service/service.go` file:
https://github.com/rancher/kontainer-engine/blob/247063a5f1e89bdbd3934075fe9b7b6b846b431e/service/service.go#L444-L475

The simple fix is to use defer:
```
...
err := cmd.Start()
...
defer cmd.Wait()
…
```

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):
 rancher/rancher:v2.3.5
- Installation option (single install/HA):  
 both

I did not submit a PR, as I did not have much time for self-test for the time being, and it may affect communication efficiency due to time zone issues.
It will be very cool if any others can fix it. ping @cjellick ?

gasxia:
Can it be blocked here?

thedadams:
### Root cause
The root cause of this issue is not "waiting" for processes after starting them in Rancher.

### What was fixed, or what changes have occurred
Now, Rancher will wait for the process so that it will not go into zombie status.

### Reproduction steps
Note: It is easiest to do this in Docker, but these steps could also be run in HA on the leader Rancher pod.
1. Deploy Docker install Rancher.
2. Once Rancher starts, activate one of the non-built-in node drivers (like OCI).
[...]

PR #36546, opened 2022-02-16
Wait for commands spawned by Rancher

thedadams:
When enabling node and kontainer-engine drivers, Rancher wasn't waiting for the
command which caused zombie processes in the rancher pod.

After this change, Rancher will wait for these processes and will not cause
zombie processes to appear.

Issue:
https://github.com/rancher/rancher/issues/25443

### DIFF
--- a/pkg/controllers/management/drivers/dynamic_driver.go
+++ b/pkg/controllers/management/drivers/dynamic_driver.go
@@ -80,9 +80,13 @@ func (d *BaseDriver) Executable() error {
 	if err != nil {
 		return fmt.Errorf("Driver %s not found", binaryPath)
 	}
-	err = exec.Command(binaryPath).Start()
+	cmd := exec.Command(binaryPath)
+	err = cmd.Start()
 	if err != nil {
 		return errors.Wrapf(err, "Driver binary %s couldn't execute", binaryPath)
 	}
+
+	// We don't care about the exit code, just want to make sure we can execute the binary
+	_ = cmd.Wait()
 	return nil
 }
--- a/pkg/kontainer-engine/service/service.go
+++ b/pkg/kontainer-engine/service/service.go
@@ -392,6 +392,7 @@ type RunningDriver struct {
 
 	listenAddress string
 	cancel        context.CancelFunc
+	cmd           *exec.Cmd
 }
 
 func (r *RunningDriver) Start() (string, error) {
@@ -458,6 +459,7 @@ func (r *RunningDriver) Start() (string, error) {
 		time.Sleep(5 * time.Second)
 
 		r.listenAddress = listenAddress
+		r.cmd = cmd
 	}
 
 	logrus.Infof("kontainerdriver %v listening on address %v", r.Name, r.listenAddress)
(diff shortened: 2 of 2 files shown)


pkg/controllers/management/drivers/dynamic_driver.go
```diff
@@ -80,9 +80,13 @@ func (d *BaseDriver) Executable() error {
...
 	}
-	err = exec.Command(binaryPath).Start()
+	cmd := exec.Command(binaryPath)
+	err = cmd.Start()
 	if err != nil {
 		return errors.Wrapf(err, "Driver binary %s couldn't execute", binaryPath)
 	}
+
+	// We don't care about the exit code, just want to make sure we can execute the binary
+	_ = cmd.Wait()
```
snasovich, line 90: Should we be concerned that we're blocked here until `cmd` finishes?

paynejacob (approved):

pkg/kontainer-engine/service/service.go
```diff
@@ -491,6 +493,8 @@ func (r *RunningDriver) Stop() {
 		r.Server.Stop()
 	} else {
 		r.cancel()
+		_ = r.cmd.Wait()
```
snasovich, line 496: Nit, but it's a bit more flexible to do this regardless of `r.Builtin` flag and always wait for the command to finish if it's set.
Also, do we have to set it to `nil` as Wait releases resources - https://pkg.go.dev/os/exec#Cmd.Wait ?

snasovich (commented):
Couple questions/comments

pkg/controllers/management/drivers/dynamic_driver.go
thedadams: The command exits (with code 1) immediately because we aren't calling it with any arguments. I can put this in a go-routine if you'd like, but I don't think that is necessary.

pkg/kontainer-engine/service/service.go
thedadams: We can do the first one, but for now there is no `cmd` if `r.Builtin == true`.

I set `r.cmd = nil` so all the memory can be cleaned up not just the resources.

pkg/controllers/management/drivers/dynamic_driver.go
snasovich: That's fine then - thank you for clarifying how it's used here.

pkg/kontainer-engine/service/service.go
snasovich: We're talking hypothetical case here that if we ever start using `cmd` in other cases the suggested approach will minimize the changes needed. I believe it's best if code follows the "actual logic of the function" and in this case the logic of function to stop RunningDriver is to ensure that `cmd` has exited (if it's set), no matter _why_ `cmd` got set.

thedadams: Done.

snasovich (approved):

paynejacob (approved):

PR #36557, opened 2022-02-17
fix health check serialization

paynejacob:
#35275

# Problem
rancher returns an error when parsing the max unhealthy in node pools because int's are parsed as strings.

# Solution
accept maxUnhealthy as a string and parse into to string inside rancher.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
+++ b/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -434,7 +440,7 @@ func deploymentHealthChecks(machineDeployment *capi.MachineDeployment, machinePo
 					Timeout: *machinePool.UnhealthyNodeTimeout,
 				},
 			},
-			MaxUnhealthy:       machinePool.MaxUnhealthy,
+			MaxUnhealthy:       &maxUnhealthy,
```
thedadams, line 443: I think that if the a value for `machinePool.MaxUnhealthy == nil` then `MaxUnhealthy` in the `MachineHealthCheck` should also be `nil`. In this implementation, if `machinePool.MaxUnhealthy == nil`, then `MaxUnhealthy` in the `MachineHealthCheck` is `0`.

Is that intended?

thedadams (changes requested):

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
paynejacob: good catch

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -410,6 +411,12 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 
 // deploymentHealthChecks Health checks will mark a machine as failed if it has any of the conditions below for the duration of the given timeout. https://cluster-api.sigs.k8s.io/tasks/healthcheck.html#what-is-a-machinehealthcheck
 func deploymentHealthChecks(machineDeployment *capi.MachineDeployment, machinePool rancherv1.RKEMachinePool) *capi.MachineHealthCheck {
+	var maxUnhealthy *intstr.IntOrString
+	if machinePool.MaxUnhealthy != nil {
+		mu := intstr.Parse(*machinePool.MaxUnhealthy)
+		maxUnhealthy = &mu
```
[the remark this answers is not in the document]
paynejacob: I was trying to think of a cleaner way but they all seemed bad.  I like this one.

thedadams (dismissed):

jiaqiluo (dismissed):

Issue #36410, opened 2022-02-04
Implement additional metrics for nodes


ryansann:
As part of the Amazon and Cloud building initiative, Rancher needs to expose additional prometheus metrics. The metrics need to track how many nodes there are and how many vCPUs / cores those nodes are using. 

The new prometheus metrics need at least the following labels:
- `cluster_id` - the id of the cluster to which the node belongs
- `type` - the type of node, one of: `k3s, rke, rke2, eks, gke, aks, longhorn, management`
- `status` - availability of the node

NOTE: 
- `type` has changed to `provider`
- `status` has changed to `is_active` and the value is a boolean string, i.e. "true" or "false"

Related to: 
- https://github.com/rancher/rancher/issues/35736

ryansann:
### Changes

- When `CATTLE_PROMETHEUS_METRICS=true` Rancher starts an additional go routine that continuously updates the gauge value for two new prometheus metrics: `cluster_manager_nodes` and `cluster_manger_node_cores`
- The metrics are labeled by: `provider, cluster_id, is_active`
  - `provider` is the kubernetes provider of the cluster to which the node belongs
    - values: `k3s, rke, rke.windows, rke2, gke, aks, eks, unknown`
  - `cluster_id` is the id of the cluster to which the node belongs
  - `is_active` is a boolean that indicates whether a node is active from the perspective of Rancher
    - a node is considered to be active iff it is ready and it belongs to a ready cluster, otherwise it is considered not active

### Notes

- When a cluster and its nodes are provisioning, the metrics for the nodes will be reported, however they will be reported with `provider="unknown",is_active="false"` labels
- A node is considered to be ready if it's kubelet is ready
- Multiple things can make a cluster not ready, some of which are: agent disconnects, upgrades, updates, etc
- Management nodes can be distinguished by `cluster_id="local"`

### Testing

See QA Test Plan

brudnak:
Testing RKE1 blocked for now due to https://github.com/rancher/rancher/issues/36588. Other cases besides RKE1 completed.

brudnak:
# My checks PASSED

## Reproduction Environment

#### Not required.

<hr>
  
## Validation Environment

| Component               | Version / Type                                     |
| ----------------------- | -------------------------------------------------- |
| Rancher version         | v2.6-a49b9d913555d59834a30f6e2ae676f0bd54bba6-head |
| Rancher cluster type    | High Availability                                  |
| Helm version            | v2.16.8-rancher1                                   |
[...]

PR #36576, opened 2022-02-18
Prometheus node metrics gauge value fix

ryansann:
Because the `setMetrics` method accepts the absolute state of what should be reported to the `cluster_manager_nodes, cluster_manager_node_cores` metrics, there could be leftover gauge values from previous calls to `setMetrics`. For this reason we need to first reset the metrics before setting them. 

Issue: 
- https://github.com/rancher/rancher/issues/36410


MbolotSuse (approved):
LGTM. Reset makes sense here rather than doing something more complex to migrate node state.

PR #36585, opened 2022-02-19
Updates to CreateK3DCluster, CreateAndImportK3DCluster, and ImportCluster

igomez06:
**Framework Updates**
- Updated `CreateK3DCluster` to use the updated flags for the newest version of k3d. `"no-hostip"` and `"k3s-server-arg"` are no longer supported. Also `"no-image-volume"` was also causing errors.
- Updated `ImportCluster` to use the `provisioning.cattle.io/v1` api to remove confusion from future users that would need to use two different apis for cluster creating and monitoring. Also updated so the container running the `kubectl apply -f` command uses the insecure command instead.
- Updated CreateAndImportK3DCluster to use the` provisioning.cattle.io/v1` api to monitor and import a cluster.

### DIFF
--- a/tests/framework/extensions/clusters/import.go
+++ b/tests/framework/extensions/clusters/import.go
@@ -2,9 +2,11 @@ package clusters
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/rancher/norman/types"
+	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/rancher/tests/framework/clients/dynamic"
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
 	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
@@ -12,6 +14,7 @@ import (
 	"github.com/rancher/rancher/tests/framework/pkg/wait"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


tests/framework/extensions/clusters/import.go
```diff
@@ -137,7 +140,10 @@ func ImportCluster(client *rancher.Client, cluster *management.Cluster, rest *re
 						{
 							Name:    "kubectl",
 							Image:   "rancher/shell:v0.1.10",
-							Command: []string{"kubectl", "apply", "--filename=" + token.ManifestURL},
+							Command: []string{"/bin/sh", "-c"},
+							Args: []string{
+								fmt.Sprintf("count=0\n until\n [ $status == 0 ] || [$count -eq 4] \n do\n apk --no-cache add curl\n status=$?\n count=$((count+1))\n done;  %s;", token.InsecureCommand),
```
paynejacob, line 145: wget is included in the shell image, it would be faster to use wget.

paynejacob (dismissed):

tests/framework/extensions/clusters/import.go
igomez06: Updated

tests/framework/extensions/clusters/import.go
```diff
@@ -137,7 +141,10 @@ func ImportCluster(client *rancher.Client, cluster *management.Cluster, rest *re
 						{
 							Name:    "kubectl",
 							Image:   "rancher/shell:v0.1.10",
-							Command: []string{"kubectl", "apply", "--filename=" + token.ManifestURL},
+							Command: []string{"/bin/sh", "-c"},
+							Args: []string{
+								fmt.Sprintf("wget -O %s --tries=10 --no-check-certificate %s; kubectl apply -f %s ;", applyFile, token.ManifestURL, applyFile),
```
paynejacob, line 146: why do a format for the apply file?

igomez06: I changed I didn't realize there was a flag for `wget` to pipe out to stdout. How does this look now?

tests/framework/extensions/clusters/import.go
```diff
@@ -121,6 +124,7 @@ func ImportCluster(client *rancher.Client, cluster *management.Cluster, rest *re
 
 	var user int64
 	var group int64
+	// applyFile := "file.yaml"
```
paynejacob, line 127: we should not commit commented code.

paynejacob (changes requested):

tests/framework/extensions/clusters/import.go
igomez06: Sorry missed that, it's fixed now.

paynejacob (approved):

bmdepesa (approved):

Issue #36599, opened 2022-02-22
[Backport v2.5] Pass container limits from project in a new namespace if it does not override them


maxsokolovsky:
Backport of https://github.com/rancher/rancher/issues/27750

brudnak:
## Reproduction Environment

| Component                 | Version / Type                                      |
| ------------------------- | --------------------------------------------------- |
| Rancher version           | 2.5.12                                              |
| Installation option       | Helm high availability (AWS)                        |
| RKE binary version used   | v1.2.16                                             |
| If Helm Chart k8s cluster | v1.20.14                                            |
| Cert Details              | ingress.tls.source=secret (certificates from files) |
| Docker version            | 20.10.7, build f0df350                              |
| Helm version              | v2.16.8-rancher1                                    |
| Downstream cluster type   | RKE1 Linode                                         |
| Downstream K8s version    | v1.20.15-rancher1-2                                 |
| Logged in user role       | Administrator                                       |
| Browser type              | Opera GX                                            |
| Browser version           | LVL3 (core: 84.0.4316.52) (x86_64)                  |

**ADDITIONAL SETUP**

- Helm install commands:

```shell
kubectl create namespace cattle-system
```

```shell
helm install rancher rancher-latest/rancher --version 2.5.12 --namespace cattle-system --set hostname=$OMITTED --set ingress.tls.source=secret
```

```shell
kubectl -n cattle-system create secret tls tls-rancher-ingress --cert=tls.crt --key=tls.key
```

## Reproduction steps

1. Starting from `/g/clusters`
2. Hover over `Global` -> click the downstream RKE1 cluster name -> click `Projects/Namespaces`
3. Create a new project
   - Click `Add Project` -> fill out the project creation form
     - Project Name: `mytestproject`
     - Click `Container Default Resource Limit`
     - CPU Limit: `1000`
     - CPU Reservation: `1000`
     - Memory Limit: `512`
     - Memory Reservation: `512`
     - Click `Create` -> then redirected back to `Projects/Namespaces` page
4. Create a namespace under the newly created project with all default values
   - Click `Add Namespace` under the newly created project -> fill out the namespace creation form
   - Name: `mytestns`
   - Click `Create`
5. Now use `kubectl` commands to inspect the newly created namespace

```shell
kubectl get namespaces
kubectl describe namespace mytestns
```

6. You will see the resource limits listed

7. Now create a new namespace under the project using `kubectl` **AND NOT** the UI

```shell
[...]

PR #36602, opened 2022-02-22
[Backport v2.5] Look up container limits defined in project when making new namespace via kubectl

maxsokolovsky:
Original issues: https://github.com/rancher/rancher/issues/27750 and https://github.com/rancher/rancher/issues/18877
Backport issue: https://github.com/rancher/rancher/issues/36599

This is a backport of two issues fixed for `v2.6.4`.

### DIFF
--- a/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
+++ b/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
@@ -450,9 +450,11 @@ func (c *SyncController) getResourceLimitToUpdate(ns *corev1.Namespace) (*corev1
 		return convertPodResourceLimitToLimitRangeSpec(updatedLimit)
 	} else if nsLimit != nil {
 		return convertPodResourceLimitToLimitRangeSpec(nsLimit)
+	} else if projectLimit != nil {
+		return convertPodResourceLimitToLimitRangeSpec(projectLimit)
+	} else {
+		return nil, nil
 	}
-
-	return nil, nil
 }
 
 func completeQuota(existingQuota *v32.NamespaceResourceQuota, defaultQuota *v32.NamespaceResourceQuota) (*v32.NamespaceResourceQuota, error) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -450,9 +450,11 @@ func (c *SyncController) getResourceLimitToUpdate(ns *corev1.Namespace) (*corev1
 		return convertPodResourceLimitToLimitRangeSpec(updatedLimit)
 	} else if nsLimit != nil {
 		return convertPodResourceLimitToLimitRangeSpec(nsLimit)
+	} else if projectLimit != nil {
+		return convertPodResourceLimitToLimitRangeSpec(projectLimit)
```
MbolotSuse, line 454: Question on this code - 2.5 does not appear to have your changes which changed how the https://github.com/rancher/rancher/commit/fa9806e37e0381f620c0632878986d9a43bc7eaf). That may cause this new code to behave differently than we saw in 2.6 - do we want to backport the other fix as well?

MbolotSuse (changes requested):
Clarification needed on other related fixes and their backport status.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Yes, I had looked at the other issue first, and, strangely, could not reproduce it under v2.5. I will try again though.
In any case, this fix is not really connected with the other issue, so their sequential order is not relevant.

MbolotSuse: The reason why I ask here is that other fix (if I was looking at it right) affected the output of `getProjectContainerDefaultLimit`. So in scenarios where the merged value was nil and the namespace value was nil, it would be using the output of that function - which is different for 2.6 because of that fix. The case where it would matter (nil quota but non-nil default limits) is pretty narrow but I think still worth double checking (at least over github).

maxsokolovsky: Yeah, good point. I will double-check. I just tried v2.5.12 again, and not having resource quotas does not prevent container default limits from being propagated, which is interesting, will investigate.

maxsokolovsky:
We will have to wait from the customers to know if they definitely need these backported to `v2.5.13`, since we no longer add changes of this nature in `v2.5.x`.

PR #36605, opened 2022-02-23
Send Rancher cert to rancher-machine

thedadams:
A previous change allowed rancher-machine to get the driver from Rancher
itself (as opposed to getting it from GitHub). This made air-gapped
environments possible for Harvester. However, this caused an issue if
Rancher was running with a self-signed cert.

This change sends the Rancher cert to the rancher-machine pod so that
the cert is trusted.

Issue:


paynejacob (approved):

jiaqiluo (approved):
LGTM.

Just have 1 question: why do we not run into this issue in the past? Is this the 1st time downloading a non-internal driver in the air gap env in the past?

thedadams:
@jiaqiluo Yes. There was another change to "fix" harvester provisioning in air-gap environments. That fix didn't work if Rancher was using a self-signed certificate. This fixes that problem.

PR #36611, opened 2022-02-23
Prefer stable cert-manager apiversion

lu1as:
Use the stable v1 apiversion for cert-manager issuer when available.

This enables updating the helm-managed issuer resource before upgrading to cert-manager version >= v1.6.0.
Otherwise Helm is not able to upgrade the Rancher Helm release afterwards, because the release references an API version which was removed by applying the CRDs for the new cert-manager version.


fabn:
Why this is not yet merged? it blocks rancher upgrades with cert-manager >= 1.6.0

Issue #36054, opened 2022-01-06
Many "will not create impersonation" errors in logs


cbron:
Follow up from this issue: https://github.com/rancher/rancher/issues/34076 which was fixed in 2.6.0. Since then we have seen errors in 2.6.2 around it. Potentially because system users don't have usernames: 

```
kubectl get users -o custom-columns=Name:metadata.name,Username:username,DisplayName:displayName | head -10
Name           Username   DisplayName
u-pjvstnafg   <none>     User 1
u-qipiwphye   <none>     User 2
u-klujqkr42   <none>     System account for Project p-4lc6c
u-majekn4pz   <none>     System account for Project p-87k61
u-uewiyjhkr   <none>     System account for Cluster c-wxlr3
```

Code: https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementuser/rbac/impersonation_handler.go#L47

SURE-3869
SURE-3874

cmurphy:
### Root cause
 
Rancher was assuming that all role template bindings were for users, when in reality the UserName field of a role template binding might be empty if the binding is actually for a group.

### What was fixed, or what changes have occurred

Check whether the binding is for a user or a group, and only make impersonation accounts for users.
 
### Areas or cases that should be tested

Deploying a downstream cluster with cluster/project owners/members
 
### What areas could experience regressions?
 
No likely regressions

### Are the repro steps accurate/minimal?

Repro steps:

1. Set up an auth provider that has groups in it, I used GitHub
2. Create a downstream cluster, and add as a cluster member a group from the auth provider
3. Once the cluster becomes ready, the warning logs will start to appear

anupama2501:
Reproduced on 2.6.2:
1. On rancher server 2.6.2, enabled openLDAP auth
2. Created a downstream rke1 cluster and added a group, user as cluster owner, cluster member and project owner, project member.
3. Verified the rancher logs and following errors are seen:
```
2022/03/03 05:17:20 [WARNING] could not find user , will not create impersonation account on cluster
2022/03/03 05:17:20 [WARNING] could not find user , will not create impersonation account on cluster
2022/03/03 05:17:21 [WARNING] could not find user , will not create impersonation account on cluster
2022/03/03 05:17:21 [WARNING] could not find user , will not create impersonation account on cluster
```



Verified on v2.6-head 39befda

1. Fresh install rancher 2.6-head - enabled openLDAP auth
2. Created a downstream cluster and added a group, user as a cluster owner/member
[...]

PR #36625, opened 2022-02-23
Fix group handling for impersonation

cmurphy:
When CRTBs and PRTBs are created, they are either created for Users or
Groups, not both. Impersonation only needs to be set up for a user, and
the impersonation account must be able to impersonate the groups that
are found in the user's attributes. This change ensures that we don't
try to create an impersonation account for a group and removes unneeded
code for when an account is being created for a user.

https://github.com/rancher/rancher/issues/36054


ryansann (approved):
lgtm

MbolotSuse (approved):
lgtm.

Issue #35647, opened 2021-11-23
Incorrect resource quota set for namespaces where Project quota is enabled


SheilaghM:
SURE-3695

Rancher Cluster:
Rancher version: v2.5.10 & v.2.6.2
Number of nodes: 3
Node OS version: Ubuntu 20.04.3 LTS

Downstream Cluster:
Number of Downstream clusters: 1 
Node OS: Ubuntu 20.04.3 LTS
RKE/RKE2/K3S version: NA
Kubernetes version: NA
CNI: NA

Longhorn:
Longhorn version:NA
CPU per node: NA
Memory per node: NA
Disk type: HDD/SSD/NVMe
Network bandwidth between the nodes: NA

Other:
Underlying Infrastructure: NA
Any 3rd party software installed on the nodes: NA
-->

Issue description:
The namespace created after utilizing all project limits; the resource quota allocated to new namespace contains quota which is not provided in project quota

Business impact:
Unable to create a new config map on a namespace where no project resource quota assigned for configmap 

Repro steps: 

Create a Project with below resource limits
```  containerDefaultResourceLimit:
    limitsCpu: 200m
    limitsMemory: 128Mi
    requestsCpu: 100m
    requestsMemory: 64Mi
  description: ""
  displayName: quotatest
  enableProjectMonitoring: false
  namespaceDefaultResourceQuota:
    limit:
      limitsCpu: "1000"
      limitsMemory: "2000"
  resourceQuota:
    limit:
      limitsCpu: "1000"
      limitsMemory: "2000"
```

Create a namespace inside the project
The namespace gets a quota with the below settings 
```  spec:
    hard:
      limits.cpu: 1k
      limits.memory: 2k
```
Create another namespace, but this time the new namespace gets below the quota
```  spec:
    hard:
      configmaps: "0"
      limits.cpu: "0"
      limits.memory: "0"
      persistentvolumeclaims: "0"
      pods: "0"
      replicationcontrollers: "0"
      requests.cpu: "0"
      requests.memory: "0"
      requests.storage: "0"
      secrets: "0"
      services: "0"
      services.loadbalancers: "0"
      services.nodeports: "0"
```

Create a new config map on second namespace
```configmaps "test" is forbidden: exceeded quota: default-995sh, requested: configmaps=1, used: configmaps=0, limited: configmaps=0```

Actual behavior:
Unable to create config maps on namespace that doesn't have a project-level configmap quota set

Expected behavior:
[...]

PR #36652, opened 2022-02-25
Create zero quota only for overused resources

maxsokolovsky:
Issue: https://github.com/rancher/rancher/issues/35647

For scenarios where namespaces are created with kubectl:
- If requested quota fits the project, then it is granted and associated with the namespace.
- If requested quota does not fit, then zero quota only for overfilled resources is granted.

It should not matter whether requested quota comes from project defaults or annotation.

The goal is to fix the current behavior where namespaces created with kubectl would receive an all-restrictive resource quota, if the namespace does not fit the project with any resource.
With this change, we will never create the all-restrictive quota (zeros for all resources).

Also, the current behavior with Rancher granting the `min(available, requested)` of a resource is unchanged for requests made with the UI. For requests made with kubectl, a resource amount must be fully available, else zero is granted.

Resources with negative values are now explicitly handled - they are treated as exceeded resources.

There has also been some refactoring done. But more work will be needed in the future to improve the readability of the resource quota sync controller.

### DIFF
--- a/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
+++ b/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
@@ -6,12 +6,8 @@ import (
 	"reflect"
 	"time"
 
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
-	"k8s.io/apimachinery/pkg/runtime"
-
-	"github.com/mitchellh/mapstructure"
 	"github.com/rancher/norman/types/convert"
+	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	namespaceutil "github.com/rancher/rancher/pkg/namespace"
@@ -21,7 +17,9 @@ import (
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	clientcache "k8s.io/client-go/tools/cache"
+	"k8s.io/kubernetes/pkg/kubelet/util/format"
 )
 
 const (
@@ -119,69 +117,66 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
 		return ns, err
 	}
 
-	projectLimit, _, err := getProjectResourceQuotaLimit(ns, c.ProjectLister)
-	if err != nil {
-		return ns, err
-	}
-
-	var quotaSpec *corev1.ResourceQuotaSpec
-	if projectLimit != nil {
-		quotaSpec, err = c.getNamespaceResourceQuota(ns)
-		if err != nil {
-			return ns, err
-		}
-	}
-
-	quotaToUpdate, err := c.getResourceQuotaToUpdate(ns)
+	requestedQuotaLimit, newQuotaSpec, err := c.deriveRequestedResourceQuota(ns)
 	if err != nil {
 		return ns, err
 	}
 
 	operation := "none"
 	if existing == nil {
-		if quotaSpec != nil {
+		if newQuotaSpec != nil {
 			operation = "create"
 		}
 	} else {
-		if quotaSpec == nil {
+		if newQuotaSpec == nil {
 			operation = "delete"
-		} else if quotaToUpdate != "" || !apiequality.Semantic.DeepEqual(existing.Spec.Hard, quotaSpec.Hard) {
+		} else if !apiequality.Semantic.DeepEqual(existing.Spec.Hard, newQuotaSpec.Hard) {
 			operation = "update"
 		}
 	}
 
 	var updated *corev1.Namespace
-	var isFit bool
+	var operationErr error
 	switch operation {
 	case "create":
-		isFit, updated, err = c.validateAndSetNamespaceQuota(ns, quotaToUpdate)
+		isFit, updated, exceeded, err := c.validateAndSetNamespaceQuota(ns, &v32.NamespaceResourceQuota{Limit: *requestedQuotaLimit})
 		if err != nil {
 			return updated, err
 		}
 		if !isFit {
-			// create default "all 0" resource quota
-			quotaSpec, err = getDefaultQuotaSpec()
+			// Create a quota with zeros only for overused resources.
+			limit, err := zeroOutResourceQuotaLimit(requestedQuotaLimit, exceeded)
+			if err != nil {
+				return updated, err
+			}
+
+			newQuotaSpec, err = convertResourceLimitResourceQuotaSpec(limit)
 			if err != nil {
 				return updated, err
 			}
 		}
-		err = c.createDefaultResourceQuota(ns, quotaSpec)
+		operationErr = c.createResourceQuota(ns, newQuotaSpec)
 	case "update":
-		isFit, updated, err = c.validateAndSetNamespaceQuota(ns, quotaToUpdate)
-		if err != nil || !isFit {
-			return updated, err
+		isFit, upd, _, err := c.validateAndSetNamespaceQuota(ns, &v32.NamespaceResourceQuota{Limit: *requestedQuotaLimit})
+		if err != nil {
+			return upd, err
 		}
-		err = c.updateResourceQuota(existing, quotaSpec)
+		if !isFit {
+			updated = upd
+			break
+		}
+		operationErr = c.updateResourceQuota(existing, newQuotaSpec)
 	case "delete":
-		err = c.deleteResourceQuota(existing)
+		operationErr = c.deleteResourceQuota(existing)
 	}
 
 	if updated == nil {
 		updated = ns
 	}
 
-	if err != nil {
-		return updated, err
+	if operationErr != nil {
+		logrus.Errorf("Failed to perform operation %q on namespace %q: %v", operation, ns.Name, operationErr)
+		return updated, operationErr
 	}
 
 	set, err := namespaceutil.IsNamespaceConditionSet(ns, ResourceQuotaInitCondition, true)
@@ -191,7 +186,6 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
 	toUpdate := updated.DeepCopy()
 	namespaceutil.SetNamespaceCondition(toUpdate, time.Second*1, ResourceQuotaInitCondition, true, "")
 	return c.Namespaces.Update(toUpdate)
-
 }
 
 func (c *SyncController) updateResourceQuota(quota *corev1.ResourceQuota, spec *corev1.ResourceQuotaSpec) error {
@@ -244,39 +238,47 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
 	return limitRanger[0], nil
 }
 
-func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (*corev1.ResourceQuotaSpec, error) {
-	limit, err := getNamespaceResourceQuotaLimit(ns)
+// deriveRequestedResourceQuota tries to obtain the new namespace's resource quota limit and its quota spec.
+// It derives it by looking up the requested quota limit. If it's not found, then it looks up the project's default
+// quota for a namespace. If it's also not found, then the method returns nil.
+// If only the requested quota limit exists, then nil returned (no limits).
+// If only the project's default namespace limit exists, then it is returned.
+// If both exist, then the two limits are merged, with requested limits having priority for overlapping resources.
+func (c *SyncController) deriveRequestedResourceQuota(ns *corev1.Namespace) (*v32.ResourceQuotaLimit, *corev1.ResourceQuotaSpec, error) {
+	requested, err := getNamespaceResourceQuotaLimit(ns)
 	if err != nil {
-		return nil, err
-	}
-	if limit == nil {
-		limit = defaultResourceLimit
+		return nil, nil, err
 	}
 
-	return convertResourceLimitResourceQuotaSpec(limit)
+	defaultQuota, err := getProjectNamespaceDefaultQuota(ns, c.ProjectLister)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var quotaLimit *v32.ResourceQuotaLimit
+
+	if requested != nil && defaultQuota == nil {
+		return nil, nil, nil
+	} else if requested == nil && defaultQuota != nil {
+		quotaLimit = &defaultQuota.Limit
+	} else if requested != nil && defaultQuota != nil {
+		quotaLimit, err = completeQuota(requested, &defaultQuota.Limit)
+		if err != nil {
+			return nil, nil, err
+		}
+	} else {
+		// This use case arises when users create a namespace outside any projects.
+		return nil, nil, nil
+	}
+
+	newQuotaSpec, err := convertResourceLimitResourceQuotaSpec(quotaLimit)
+	if err != nil {
+		return nil, nil, err
+	}
+	return quotaLimit, newQuotaSpec, nil
 }
 
-func getDefaultQuotaSpec() (*corev1.ResourceQuotaSpec, error) {
-	return convertResourceLimitResourceQuotaSpec(defaultResourceLimit)
-}
-
-var defaultResourceLimit = &v32.ResourceQuotaLimit{
-	Pods:                   "0",
-	Services:               "0",
-	ReplicationControllers: "0",
-	Secrets:                "0",
-	ConfigMaps:             "0",
-	PersistentVolumeClaims: "0",
-	ServicesNodePorts:      "0",
-	ServicesLoadBalancers:  "0",
-	RequestsCPU:            "0",
-	RequestsMemory:         "0",
-	RequestsStorage:        "0",
-	LimitsCPU:              "0",
-	LimitsMemory:           "0",
-}
-
-func (c *SyncController) createDefaultResourceQuota(ns *corev1.Namespace, spec *corev1.ResourceQuotaSpec) error {
+func (c *SyncController) createResourceQuota(ns *corev1.Namespace, spec *corev1.ResourceQuotaSpec) error {
 	resourceQuota := &corev1.ResourceQuota{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: "default-",
@@ -304,30 +306,34 @@ func (c *SyncController) createDefaultLimitRange(ns *corev1.Namespace, spec *cor
 	return err
 }
 
-func (c *SyncController) validateAndSetNamespaceQuota(ns *corev1.Namespace, quotaToUpdate string) (bool, *corev1.Namespace, error) {
+func (c *SyncController) validateAndSetNamespaceQuota(ns *corev1.Namespace, quotaToUpdate *v32.NamespaceResourceQuota) (bool, *corev1.Namespace, corev1.ResourceList, error) {
 	if ns == nil || ns.DeletionTimestamp != nil {
-		return true, ns, nil
+		return true, ns, nil, nil
 	}
 
 	// get project limit
 	projectLimit, projectID, err := getProjectResourceQuotaLimit(ns, c.ProjectLister)
 	if err != nil {
-		return false, ns, err
+		return false, ns, nil, err
 	}
 
 	if projectLimit == nil {
-		return true, ns, err
+		return true, ns, nil, err
 	}
 
 	updatedNs := ns.DeepCopy()
-	if quotaToUpdate != "" {
+	if quotaToUpdate != nil {
 		if updatedNs.Annotations == nil {
 			updatedNs.Annotations = map[string]string{}
 		}
-		updatedNs.Annotations[resourceQuotaAnnotation] = quotaToUpdate
+		b, err := json.Marshal(quotaToUpdate)
+		if err != nil {
+			return false, ns, nil, err
+		}
+		updatedNs.Annotations[resourceQuotaAnnotation] = string(b)
 		updatedNs, err = c.Namespaces.Update(updatedNs)
 		if err != nil {
-			return false, updatedNs, err
+			return false, updatedNs, nil, err
 		}
 	}
 
@@ -335,92 +341,54 @@ func (c *SyncController) validateAndSetNamespaceQuota(ns *corev1.Namespace, quot
 	mu := validate.GetProjectLock(projectID)
 	mu.Lock()
 	defer mu.Unlock()
-	// get other Namespaces
+
+	// Get other namespaces' limits.
+	nsLimits, err := c.getNamespacesLimits(ns, projectID)
+	if err != nil {
+		return false, updatedNs, nil, err
+	}
+	isFit, exceeded, err := validate.IsQuotaFit(&quotaToUpdate.Limit, nsLimits, projectLimit)
+	if err != nil {
+		return false, updatedNs, nil, err
+	}
+
+	var msg string
+	if !isFit && exceeded != nil {
+		msg = fmt.Sprintf("Resource quota [%v] exceeds project limit", format.ResourceList(exceeded))
+	}
+
+	validated, err := c.setValidated(updatedNs, isFit, msg)
+
+	return isFit, validated, exceeded, err
+}
+
+func (c *SyncController) getNamespacesLimits(ns *v1.Namespace, projectID string) ([]*v32.ResourceQuotaLimit, error) {
 	objects, err := c.NsIndexer.ByIndex(nsByProjectIndex, projectID)
 	if err != nil {
-		return false, updatedNs, err
+		return nil, err
 	}
 	var nsLimits []*v32.ResourceQuotaLimit
 	for _, o := range objects {
 		other := o.(*corev1.Namespace)
-		// skip itself
+		// Skip itself.
 		if other.Name == ns.Name {
 			continue
 		}
 		nsLimit, err := getNamespaceResourceQuotaLimit(other)
 		if err != nil {
-			return false, updatedNs, err
+			return nil, err
 		}
 		nsLimits = append(nsLimits, nsLimit)
 	}
-	nsLimit, err := getNamespaceResourceQuotaLimit(updatedNs)
-	if err != nil {
-		return false, updatedNs, err
-	}
-	isFit, msg, err := validate.IsQuotaFit(nsLimit, nsLimits, projectLimit)
-	if err != nil {
-		return false, updatedNs, err
-	}
-
-	if !isFit && msg != "" {
-		msg = fmt.Sprintf("Resource quota [%v] exceeds project limit ", msg)
-	}
-
-	validated, err := c.setValidated(updatedNs, isFit, msg)
-
-	return isFit, validated, err
-
+	return nsLimits, nil
 }
 
 func (c *SyncController) setValidated(ns *corev1.Namespace, value bool, msg string) (*corev1.Namespace, error) {
-	set, err := namespaceutil.IsNamespaceConditionSet(ns, ResourceQuotaValidatedCondition, value)
-	if set || err != nil {
-		return ns, err
-	}
 	toUpdate := ns.DeepCopy()
-	err = namespaceutil.SetNamespaceCondition(toUpdate, time.Second*1, ResourceQuotaValidatedCondition, value, msg)
-	if err != nil {
+	if err := namespaceutil.SetNamespaceCondition(toUpdate, time.Second*1, ResourceQuotaValidatedCondition, value, msg); err != nil {
 		return ns, err
 	}
 	return c.Namespaces.Update(toUpdate)
-}
-
-func (c *SyncController) getResourceQuotaToUpdate(ns *corev1.Namespace) (string, error) {
-	quota := getNamespaceResourceQuota(ns)
-	defaultQuota, err := getProjectNamespaceDefaultQuota(ns, c.ProjectLister)
-	if err != nil {
-		return "", err
-	}
-
-	// rework after api framework change is done
-	// when annotation field is passed as null, the annotation should be removed
-	// instead of being updated with the null value
-	var updatedQuota *v32.NamespaceResourceQuota
-	if quota != "" && quota != "null" {
-		// check if fields need to be removed or set
-		// based on the default quota
-		var existingQuota v32.NamespaceResourceQuota
-		err := json.Unmarshal([]byte(convert.ToString(quota)), &existingQuota)
-		if err != nil {
-			return "", err
-		}
-		updatedQuota, err = completeQuota(&existingQuota, defaultQuota)
-		if updatedQuota == nil || err != nil {
-			return "", err
-		}
-	}
-
-	var b []byte
-	if updatedQuota == nil {
-		b, err = json.Marshal(defaultQuota)
-	} else {
-		b, err = json.Marshal(updatedQuota)
-	}
-
-	if err != nil {
-		return "", err
-	}
-	return string(b), nil
 }
 
 func (c *SyncController) getResourceLimitToUpdate(ns *corev1.Namespace) (*corev1.LimitRangeSpec, error) {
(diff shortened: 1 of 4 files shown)


pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -244,39 +254,15 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
 	return limitRanger[0], nil
 }
 
-func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (*corev1.ResourceQuotaSpec, error) {
+func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (v32.NamespaceResourceQuota, error) {
```
cmurphy, line 257: Why did the return value change from a pointer to non-pointer?

pkg/resourcequota/quota_validate.go
```diff
@@ -29,33 +26,35 @@ func GetProjectLock(projectID string) *sync.Mutex {
...
-		return false, "", err
+		return false, nil, err
 	}
 
-	allowed, exceeded := quota.LessThanOrEqual(nssResourceList, projectResourceList)
-	if allowed {
-		return true, "", nil
+	_, exceeded := quota.LessThanOrEqual(nssResourceList, projectResourceList)
+	// Include resources with negative values among exceeded resources.
+	exceeded = append(exceeded, quota.IsNegative(nsResourceList)...)
```
cmurphy, line 52: If the UI was working correctly, would there ever be a need for this? Why should values be negative?

pkg/controllers/managementuser/resourcequota/resource_quota_common.go
```diff
@@ -270,3 +270,18 @@ func convertPodResourceLimitToLimitRangeSpec(podResourceLimit *v32.ContainerReso
 	}
 	return limitRangeSpec, err
 }
+
+func zeroOutResourceQuotaLimit(limit *v32.ResourceQuotaLimit) (*v32.ResourceQuotaLimit, error) {
```
cmurphy, line 274: This is only used once so it's not really "common", might be more appropriate to put it in resource_quota_sync.go where the original getDefaultQuotaSpec was.

cmurphy (commented):
I think this looks good, left a couple of minor questions.

pkg/resourcequota/quota_validate.go
maxsokolovsky: Two reasons:
1. I don't know if this gets released before the UI fix, so we could be allowing negative values, even if for a short while.
2. Users can still put negative values in the annotation before submitting a manifest directly via kubectl.

pkg/controllers/managementuser/resourcequota/resource_quota_common.go
maxsokolovsky: Agreed!

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,69 +118,84 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
...
 		if !isFit {
-			// create default "all 0" resource quota
-			quotaSpec, err = getDefaultQuotaSpec()
+			// Create a quota with zeros only for overused resources.
+			limit, err := convertResourceListToLimit(exceeded)
+			if err != nil {
+				return updated, err
+			}
+
+			limit, err = zeroOutResourceQuotaLimit(limit)
```
MbolotSuse, line 165: Given that the `zeroOutResourceQuotaLimit` function is only used here, it might be more helpful to augment it to handle most of the logic that fits within this if statement. 

For example, that function could be changed to take in a limit (i.e. the limit which had excesses) and a list of resource which exceeded the limits (the exceeded values). It could then overwrite the exceeded values and return a properly parsed resource. This could also have the benefit of allowing us to not use completeQuota - which I think is a bit of a dangerous method given that it combines these two values in a very particular way.

MbolotSuse (changes requested):

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -385,42 +387,29 @@ func (c *SyncController) setValidated(ns *corev1.Namespace, value bool, msg stri
...
+	if existingQuota == nil {
+		return defaultQuota, nil
 	}
 
 	// rework after api framework change is done
 	// when annotation field is passed as null, the annotation should be removed
 	// instead of being updated with the null value
 	var updatedQuota *v32.NamespaceResourceQuota
-	if quota != "" && quota != "null" {
+	if existingQuota != nil {
```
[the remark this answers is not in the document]
maxsokolovsky: Good catch!

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -304,72 +293,85 @@ func (c *SyncController) createDefaultLimitRange(ns *corev1.Namespace, spec *cor
 	return err
 }
 
-func (c *SyncController) validateAndSetNamespaceQuota(ns *corev1.Namespace, quotaToUpdate string) (bool, *corev1.Namespace, error) {
+func (c *SyncController) validateAndSetNamespaceQuota(ns *corev1.Namespace, quotaToUpdate *v32.NamespaceResourceQuota) (bool, *corev1.Namespace, corev1.ResourceList, error) {
```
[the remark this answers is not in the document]
maxsokolovsky: Absolutely! Too many return values, and the function is doing too much - I agree.
I would love to simplify this, I just think the main thing here is to remove the updated namespace and all manipulations with it. I wonder if we simply don't need them. I am not yet sure if changing this is dangerous, especially for a relatively hot fix. One thing I see, for example, is how an annotation is set on the namespace with an error informing about insufficient resource(s). Is this really necessary?

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -385,42 +387,29 @@ func (c *SyncController) setValidated(ns *corev1.Namespace, value bool, msg stri
...
-			return "", err
-		}
-	}
-
-	var b []byte
-	if updatedQuota == nil {
-		b, err = json.Marshal(defaultQuota)
-	} else {
-		b, err = json.Marshal(updatedQuota)
+		updatedQuota, err = completeQuota(existingQuota, defaultQuota)
```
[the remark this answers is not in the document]
maxsokolovsky: In the UI, you can't specify namespace limits for things not specified on the project.
In kubectl, you can specify, for example, configMaps via the annotation, and then they become part of the quota limit (if the project does not set a limit on configMaps).

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,69 +118,84 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
...
+	var nsResourceQuota v32.NamespaceResourceQuota
+	if projectLimit, _, err := getProjectResourceQuotaLimit(ns, c.ProjectLister); err == nil && projectLimit != nil {
+		nsResourceQuota, err = c.getNamespaceResourceQuota(ns)
 		if err != nil {
 			return ns, err
 		}
 	}
 
-	quotaToUpdate, err := c.getResourceQuotaToUpdate(ns)
+	quotaToUpdate, err := c.getResourceQuotaToUpdate(ns, &nsResourceQuota)
```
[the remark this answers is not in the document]
maxsokolovsky: Good point! When do we need to look at project's total quota? I think we don't need them (at least in this method). We do care about project's default, but those are fetched inside `getResourceQuotaToUpdate`. So I will remove the code that retrieves total quota - a call to `getProjectResourceQuotaLimit`.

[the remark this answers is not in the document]
MbolotSuse: I would tend to agree with you - not sure why there was an if check here in the first place. I think it didn't end up affecting anything because the merged code re-fetched it anyways, but I'm not 100% sure on that.

maxsokolovsky:
Will squash all commits into one later, before merge - so that reviewers can easily see changes after initial reviews.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -257,12 +243,12 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
 	return limitRanger[0], nil
 }
 
-func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (v32.NamespaceResourceQuota, error) {
+func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (v32.ResourceQuotaLimit, error) {
```
MbolotSuse, line 246: Now that this returns a Limit you may want to change the name of this function to indicate that this is returning a limit (especially given how much we swap between the two in this file).

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -387,27 +373,25 @@ func (c *SyncController) setValidated(ns *corev1.Namespace, value bool, msg stri
 	return c.Namespaces.Update(toUpdate)
 }
 
-func (c *SyncController) getResourceQuotaToUpdate(ns *corev1.Namespace, existingQuota *v32.NamespaceResourceQuota) (*v32.NamespaceResourceQuota, error) {
+func (c *SyncController) getResourceQuotaToUpdate(ns *corev1.Namespace, requestedQuota *v32.ResourceQuotaLimit) (*v32.ResourceQuotaLimit, error) {
```
MbolotSuse, line 376: Similar comment here about names and quotas vs limits.

MbolotSuse (commented):
Small nitpicks around naming.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
MbolotSuse: Resolving - changes made.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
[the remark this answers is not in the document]
MbolotSuse: Resolving - changes made.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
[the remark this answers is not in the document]
MbolotSuse: If the UI prevents you from setting namespace limits for resources which have no limits set in the project than this conversation is more academic than anything else.

That being said, I don' think that a namespace specific limit for a newer type will be included. CompleteQuota won't include these limits unless something is defined in default quota (from the project) that it can override because of how the looping structure works. E.x. below:

1. Create a project - don't define a resourceLimit for configMaps.
2. Create a namespace, define a resourceLimit for configMaps (e.g. 50).
3. completeQuota is called.
4. Since there is no limit for configMaps in the project, no limit for configMaps is added to the combined map.
5. The namespace does not get a resourceLimit for configMaps despite explicitly asking for it.

Thoughts?

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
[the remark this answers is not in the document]
MbolotSuse: I would think that the annotation can be useful for UI display purposes. That being said, feel free to resolve if you don't want to change for the time being - I understand wanting to keep this small for fix purposes.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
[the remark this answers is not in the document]
maxsokolovsky: That's the thing - if you try this via kubectl, then the namespace does get this resource limit for configMaps, when my branch's code is run.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -243,7 +243,7 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
 	return limitRanger[0], nil
 }
 
-func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (v32.ResourceQuotaLimit, error) {
+func (c *SyncController) getNamespaceResourceQuotaLimit(ns *corev1.Namespace) (v32.ResourceQuotaLimit, error) {
```
cmurphy, line 246: I think @MbolotSuse's point is that this went from getting a [NamespaceResourceQuota](https://github.com/rancher/rancher/blob/cf4364d2a5b09a9075910a3f3179ed178d3cde61/pkg/apis/management.cattle.io/v3/resource_quota_types.go#L8) to a [ResourceQuotaLimit](https://github.com/rancher/rancher/blob/cf4364d2a5b09a9075910a3f3179ed178d3cde61/pkg/apis/management.cattle.io/v3/resource_quota_types.go#L12) so the name should actually be `getResourceQuotaLimit`. There is already a function called `getNamespaceResourceQuotaLimit` which is called on the line below so I don't think this will compile.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
cmurphy: >  There were a few places where the code returns nil, nil for the value and error, which the caller may not properly handle.

I disagree with this reasoning. If a method returns a pointer, it is absolutely the caller's responsibility to check for nilness before using it.

> But then I also thought it's fine to keep the change - I return the value itself, rather then the pointer, from the function which creates the value.

A nil value is useful for showing whether the value has been set and whether it makes sense to use the value. In this case, the point of this nil check

https://github.com/rancher/rancher/pull/36652/files#diff-6247a4224a746023f729c575824523aa859221464fb312fc067de65baf55270dR381

is diluted because the object received from 

https://github.com/rancher/rancher/pull/36652/files#diff-6247a4224a746023f729c575824523aa859221464fb312fc067de65baf55270dR121

is never nil, so the default quota will never be used.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -516,3 +468,21 @@ func completeLimit(existingLimit *v32.ContainerResourceLimit, defaultLimit *v32.
...
+		return nil, err
+	}
+
+	for k := range exceeded {
+		resource := string(k)
+		limitMap[resource] = "0"
+	}
+
+	toReturn := &v32.ResourceQuotaLimit{}
+	err = convert.ToObj(limitMap, toReturn)
```
MbolotSuse, line 486: I think that we do something similar to this (convert from a map to an object) in `completeQuota`. There, we use:

```
mapstructure.Decode(newLimitMap, &toReturn)
```
Would you mind elaborating on why you used `convert.ToObj` here rather than `mapstructure.Decode`? I don't have a preference, but I would think that we would want to use the same library in each case here.

MbolotSuse (changes requested):

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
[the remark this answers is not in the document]
MbolotSuse: Talked about this yesterday - conclusion was that this is working because the memory that the function is using to store the computed map is pre-allocated with a deepcopy from the project limit, which causes this merge to happen. Resolving since this was fixed in the latest commits.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,69 +119,65 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
 		return ns, err
 	}
 
-	projectLimit, _, err := getProjectResourceQuotaLimit(ns, c.ProjectLister)
+	requestedQuotaLimit, err := c.getRequestedResourceQuotaLimit(ns)
```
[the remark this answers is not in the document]
maxsokolovsky: So that the method returns both the quota limit and quota spec? Yeah, that would be good.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: I used `convert.ToObj` because it's used in the existing `convertResourceListToLimit` function. There should not be a difference in our use-case, but I can switch to `mapstructure.Decode`.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,69 +119,65 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
...
 			if err != nil {
 				return updated, err
 			}
 		}
-		err = c.createDefaultResourceQuota(ns, quotaSpec)
+		operationErr = c.createResourceQuota(ns, newQuotaSpec)
 	case "update":
-		isFit, updated, err = c.validateAndSetNamespaceQuota(ns, quotaToUpdate)
+		isFit, updated, _, err := c.validateAndSetNamespaceQuota(ns, &v32.NamespaceResourceQuota{Limit: *requestedQuotaLimit})
 		if err != nil || !isFit {
```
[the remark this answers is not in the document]
maxsokolovsky: Agreed!

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -304,72 +289,85 @@ func (c *SyncController) createDefaultLimitRange(ns *corev1.Namespace, spec *cor
...
 	mu.Lock()
 	defer mu.Unlock()
-	// get other Namespaces
+
+	// Get other namespaces' limits.
+	nsLimits, err := c.getNamespacesLimits(ns, projectID)
+	if err != nil {
+		return false, nil, nil, err
+	}
+	nsLimit, err := getNamespaceResourceQuotaLimit(updatedNs)
```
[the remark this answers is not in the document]
maxsokolovsky: Good point!

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -244,39 +239,29 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
 	return limitRanger[0], nil
 }
 
-func (c *SyncController) getNamespaceResourceQuota(ns *corev1.Namespace) (*corev1.ResourceQuotaSpec, error) {
-	limit, err := getNamespaceResourceQuotaLimit(ns)
+func (c *SyncController) getRequestedResourceQuotaLimit(ns *corev1.Namespace) (*v32.ResourceQuotaLimit, error) {
```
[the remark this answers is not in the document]
maxsokolovsky: I think this would look fine, given that we also return the K8s quota spec:
`requestedQuotaLimit, newQuotaSpec, err := c.deriveRequestedResourceQuota(ns)`

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: But then we might as well be not using `convert.EncodeToMap`, since`mapstructure.Decode` can also do this, if I understand correctly...

MbolotSuse: I'm not particularly partial to either method - it just seems like they are doing the same thing and I would rather have us only depend on one library rather than two.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -457,34 +417,26 @@ func (c *SyncController) getResourceLimitToUpdate(ns *corev1.Namespace) (*corev1
 	}
 }
 
-func completeQuota(existingQuota *v32.NamespaceResourceQuota, defaultQuota *v32.NamespaceResourceQuota) (*v32.NamespaceResourceQuota, error) {
-	if defaultQuota == nil {
+func completeQuota(requestedQuota *v32.ResourceQuotaLimit, defaultQuota *v32.ResourceQuotaLimit) (*v32.ResourceQuotaLimit, error) {
```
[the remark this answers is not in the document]
maxsokolovsky: Discussed with @cbron - this is normal behavior. Potentially this can be counter-intuitive for some users.
In the UI, however, users can request fewer configMaps than the default, which they can see.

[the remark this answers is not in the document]
MbolotSuse: IMO it makes a lot of sense to make the default option a safe option (i.e. - if we need you to set a value it should be required to set the value, otherwise, it should always be safe to use the default). 

But that's just my opinion - as long as we are aware of (and ok with) the solution not much more to say. Feel free to resolve.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: I will only use our `convert` package. Looks like `mapstructure` needs a bit more configuration, because it changes the case of the key names in maps, and this breaks things in at least one case. I verified that using our `convert` does not change expected behavior.

pkg/controllers/managementuser/resourcequota/resource_quota_common.go
```diff
@@ -217,7 +217,7 @@ func getProjectContainerDefaultLimit(ns *corev1.Namespace, projectLister v3.Proj
 func getNamespaceResourceQuotaLimit(ns *corev1.Namespace) (*v32.ResourceQuotaLimit, error) {
 	value := getNamespaceResourceQuota(ns)
 	if value == "" {
-		return nil, nil
+		return &v32.ResourceQuotaLimit{}, nil
```
MbolotSuse, line 220: I realize that you are trying to avoid a `nil, nil` return here, but I am a bit uncomfortable with a blank limit being the default for this function. Especially because the newly renamed `deriveRequestedResourceQuota` function does a nil check for this value.

I think that the the `completeQuota` function looks like it can handle a blank default value (it will just not replace anything from the project), but the case of a nil project limit and a nil namespace limit (where we don't want any limits/quotas set) is now going to be different (you will return a zero quota limit and not a nil) for `deriveRequestedResourceQuota`. I would double check that there is no difference between no ResourceQuotaLiimit and a `ResourceQuotaLimit{}`

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -244,39 +237,45 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
...
+	if requested != nil && defaultQuota == nil {
+		quotaLimit = requested
+	} else if requested == nil && defaultQuota != nil {
+		quotaLimit = &defaultQuota.Limit
+	} else if requested != nil && defaultQuota != nil {
+		quotaLimit, err = completeQuota(requested, &defaultQuota.Limit)
+		if err != nil {
+			return nil, nil, err
+		}
+	} else {
```
MbolotSuse, line 267: Is this best represented as an error case? I would think that if you don't specify limits/quotas on the project or namespace we just don't set anything. If we return an error here I would expect to see that logged, which may be disruptive to the customer.

Though I don't think we will hit this case because requested should always be populated.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -244,39 +237,45 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
...
-	ConfigMaps:             "0",
-	PersistentVolumeClaims: "0",
-	ServicesNodePorts:      "0",
-	ServicesLoadBalancers:  "0",
-	RequestsCPU:            "0",
-	RequestsMemory:         "0",
-	RequestsStorage:        "0",
-	LimitsCPU:              "0",
-	LimitsMemory:           "0",
+	newQuotaSpec, err := convertResourceLimitResourceQuotaSpec(requested)
```
MbolotSuse, line 271: Wouldn't we want to convert the quotaLimit (which has gone through the parsing efforts above) rather than the requested value?

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -304,72 +303,81 @@ func (c *SyncController) createDefaultLimitRange(ns *corev1.Namespace, spec *cor
...
 	// validate resource quota
 	mu := validate.GetProjectLock(projectID)
 	mu.Lock()
 	defer mu.Unlock()
-	// get other Namespaces
+
+	// Get other namespaces' limits.
+	nsLimits, err := c.getNamespacesLimits(ns, projectID)
+	if err != nil {
+		return false, nil, nil, err
```
MbolotSuse, line 345: Clarification Question:

Why do you return Nil for the namespace here but return `updatedNs` along with the error in L349?

maxsokolovsky: That's a good catch. The original code returned `updatedNs` in this case.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Yes, this is my mistake! What's interesting is that none of my test scenarios produced incorrect behavior, which is odd...

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Requested may be missing - when you create a namespace via kubectl without the annotation. But to your original point, I think this is a normal error case because we don't know what to do if the merge failed. We can't assume we should just use requested or default on its own.

pkg/controllers/managementuser/resourcequota/resource_quota_common.go
maxsokolovsky: You're right, at this point there is no need to have this non-nil pointer, because the case for it to be explicitly non-nil is moot, since we thoroughly check the value of `requested` and `default` and their combinations.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
MbolotSuse: The case that I am most concerned about would be the nil project and nil namespace limits case. I imagine that this would go as below:
1. The namespace value is nil, so we don't use the first if block.
2. The project value is nil, so we don't use the first if/else block.
3. Since both namespace and project are nil, we don't use the second if/else block.
4. We do use the else block, which returns an error. This returns (and potentially logs/sets) an error, but there's no error. The user at no point (on the project or the namespace) asked for quotas/limits, and we didn't apply a quota/limit.

Let me know if you disagree and I am missing something.

maxsokolovsky: I see, the situation you're describing can happen when a user creates a namespace outside of any project. One can set container resource limits, and those will not get created on the new namespace because we return the error too early.
We can fix this by returning nil for the error in addition to nil for quota limit and quota spec.

maxsokolovsky: Verified - not returning an error in this case allows namespaces outside of projects to be created properly.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -304,72 +306,81 @@ func (c *SyncController) createDefaultLimitRange(ns *corev1.Namespace, spec *cor
...
 	updatedNs := ns.DeepCopy()
-	if quotaToUpdate != "" {
+	if quotaToUpdate != nil {
 		if updatedNs.Annotations == nil {
 			updatedNs.Annotations = map[string]string{}
 		}
-		updatedNs.Annotations[resourceQuotaAnnotation] = quotaToUpdate
+		b, err := json.Marshal(quotaToUpdate)
+		if err != nil {
+			return false, nil, nil, err
```
MbolotSuse, line 331: Nit: in this case, for the second return value, would we want to return the original namespace rather than nil?

I don't think that it really matter much given that the caller really should check for an error before attempting to use any of the values returned from this, but I figured its at least worth discussing.

MbolotSuse, line 331: Nit: in this case, for the second return value, would we want to return the original namespace rather than nil?

I don't think that it really matter much given that the caller really should check for an error before attempting to use any of the values returned from this, but I figured its at least worth discussing.

MbolotSuse (commented):
One minor question.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,68 +117,65 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
...
 	}
 
 	if updated == nil {
 		updated = ns
 	}
 
-	if err != nil {
+	if operationErr != nil {
+		logrus.Tracef("Failed to perform operation %q on namespace %q: %v", operation, ns.Name, operationErr)
 		return updated, err
```
cmurphy, line 179: This is now returning the wrong error
```suggestion
		return updated, operationErr
```

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,68 +117,65 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
...
+		operationErr = c.deleteResourceQuota(existing)
 	}
 
 	if updated == nil {
 		updated = ns
 	}
 
-	if err != nil {
+	if operationErr != nil {
+		logrus.Tracef("Failed to perform operation %q on namespace %q: %v", operation, ns.Name, operationErr)
```
cmurphy, line 178: TRACE messages are the deepest log level and are really only meant to be used for programmer debugging. This looks like a message that wants to be surfaced at regular log levels, right?

```suggestion
		logrus.Errorf("Failed to perform operation %q on namespace %q: %v", operation, ns.Name, operationErr)
```

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -304,72 +305,81 @@ func (c *SyncController) createDefaultLimitRange(ns *corev1.Namespace, spec *cor
...
 	objects, err := c.NsIndexer.ByIndex(nsByProjectIndex, projectID)
 	if err != nil {
-		return false, updatedNs, err
+		return nil, err
 	}
 	var nsLimits []*v32.ResourceQuotaLimit
 	for _, o := range objects {
 		other := o.(*corev1.Namespace)
-		// skip itself
+		// Skip itself.
```
cmurphy, line 372: (nit) unnecessary change

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -244,39 +238,46 @@ func (c *SyncController) getExistingLimitRange(ns *corev1.Namespace) (*corev1.Li
...
 
-	return convertResourceLimitResourceQuotaSpec(limit)
-}
+	var quotaLimit *v32.ResourceQuotaLimit
 
-func getDefaultQuotaSpec() (*corev1.ResourceQuotaSpec, error) {
-	return convertResourceLimitResourceQuotaSpec(defaultResourceLimit)
-}
+	if requested != nil && defaultQuota == nil {
+		quotaLimit = requested
```
cmurphy, line 260: Is this correct? Is this the behavior that happened before this change? It looks like the UI doesn't even let you set a quota on the namespace if the project has no default.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -119,68 +117,65 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
...
 	case "update":
-		isFit, updated, err = c.validateAndSetNamespaceQuota(ns, quotaToUpdate)
-		if err != nil || !isFit {
+		isFit, updated, exceeded, err := c.validateAndSetNamespaceQuota(ns, &v32.NamespaceResourceQuota{Limit: *requestedQuotaLimit})
+		if err != nil {
 			return updated, err
 		}
-		err = c.updateResourceQuota(existing, quotaSpec)
+		if !isFit {
+			operationErr = fmt.Errorf("update quota doesn't fit due to the following resource requests: [%v]", format.ResourceList(exceeded))
```
cmurphy, line 165: This seems like a behavior change that could have some impact. Before, if `isFit` was false on update, then this would return nil and appear successful. Are you sure that this new behavior, where a bad fit now returns an error, is correct behavior? (It would seem to make sense to me but I want to be sure the ramifications have been thought through.)

cmurphy (commented):
In the future it would be better to try to keep cosmetic changes and refactors as a separate commit or separate PR. It is hard to review the logic of a change when it includes so many unrelated changes.

Left comments on some minor flaws and some questions about some of the behavior.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Right, the UI lets you set the quota only for those resources that are found in the project quota. The necessity of this, by the way, is a question for the future.
But this handles both UI and kubectl. In the manifest users can specify anything in the annotation.
And, yes, this behavior is not changing.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Good point. The difference is that now we won't execute the last statements of the method, which set informational annotations on the namespace, like the whether validation of the namespace failed and, if yes, why. But we want that.

maxsokolovsky: Pushed an update so that the annotation is always written to a namespace - now on update the exact exceeded resources are shown.

MbolotSuse (approved):
LGTM - I think that this will move us to a better place.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
cmurphy: This actually does seem to be a change in behavior from my testing. Without this change, if you create a project with no default quota set and a namespace like

```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    field.cattle.io/projectId: local:p-9zk9c
    field.cattle.io/resourceQuota: '{"limit":{"pods":"50","limitsCpu":"1000m"}}'
  name: n4
spec:
```

then no ResourceQuota would be created in the namespace. With this change, there now is one. I don't think we should make any further changes to the quota behavior unrelated to the original bug without further discussion.

maxsokolovsky: Sorry, I misunderstood your original question - I thought you were asking about resources not defined on the project and their quota vs that for resources that ARE defined on the project.

You're right, the behavior does change, I wasn't aware of it, my fault. I believe though that the new behavior is more intuitive, but, as you said, we need to discuss this.
@cbron

maxsokolovsky: It all comes down to the role of the annotation and whether users should use it and rely on it. If we maintain that it should not be used, then the old behavior makes more sense. But if we honor the annotation, then the new behavior should be expected.

maxsokolovsky:
Tested an upgrade from `v2.6.3-patch1` to a Rancher image including my changes and those from the recent upgrade to K8s v1.23. No issues found, existing quotas were not affected.

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Discussed in standup - for this case, reverting to existing behavior. If there is a requested quota but no default on the project, then do not apply quota limits.

cmurphy (approved):

MbolotSuse (approved):

Issue #36626, opened 2022-02-23
Unable to provision rke2 node driver linode cluster errors with - `ls: cannot access 'docker-machine-driver-*': No such file or directory`


anupama2501:
**Rancher Server Setup**
- Rancher version: `v2.6-head 8c785a1`
- Installation option (Docker install/Helm Chart): Docker install
   - If Helm Chart, Kubernetes Info: k3s
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):
         Node Setup: 1 node 
         Version: `v1.21.3+k3s1`
- Proxy/Cert Details: `self-signed`

**Information about the Cluster**
- Kubernetes version: `v1.21.9+rke2r1`
- Cluster Type (Local/Downstream): `Downstream infrastructure provider 1 worker, 1 etcd + cp`
`RKE2`

**Describe the bug**
Provisioning an rke2 node driver cluster fails with the error 
`CreateError: Downloading driver from https://45.33.74.83/assets/docker-machine-driver-linode` 

**To Reproduce**
 Create an rke2 node driver cluster with linode as provider

**Result**
Provisioning fails with the error:
```
Waiting] provisioning bootstrap node(s) anu-linode-ec-767964fb66-77gb7: waiting for failed creating server (LinodeMachine) in infrastructure provider: CreateError: Downloading driver from https://<redacted>/assets/docker-machine-driver-linode ls: cannot access 'docker-machine-driver-*': No such file or directory downloaded file failed sha256 checksum download of driver from https://<redacted>/assets/docker-machine-driver-linode failed, agent to check in and apply initial plan
```

**Expected Result**
Expected the provisioning to go through without any error

jakefhyde:
@thedadams Is this related to another issue you were working on?

thedadams:
@jakefhyde Yep, right now.

noahgildersleeve:
This happens with the node-driver for Harvester as well. Rancher `v2.6.4-rc4` and `v2.6-head` with Harvester `master-bfd6a8a9-head`

anupama2501:
Verified on an upgraded setup from 2.6.2 --> 2.6-head e93a53c

- Created a linode single node all roles -- PASSS
- Created a linode multi node 1 etcd, 1 cp, 3 worker -- PASS
- Created a workload deployment, an SD - works fine

Fresh install checks on 2.6-head e93a53c

- Created a linode single node all roles -- PASSS
[...]

Issue #36547, opened 2022-02-17
Provisioning machine job can't finish


FrankYang0529:
**Rancher Server Setup**
- Rancher version: v2.6.3
- Installation option (Docker install/Helm Chart): Docker

**Information about the Cluster**
- Kubernetes version: v1.22.6+rke2r1
- Cluster Type (Local/Downstream): Infrastructure Provider - Harvester

**User Information**
- What is the role of the user logged in? Admin
   
**Describe the bug**
Try to add node to a RKE2 cluster on Harvester. The Job may get deleted and have OOM issue with following logs:
```
Error creating machine: Error saving host to store after attempting creation: unable to load secret : error getting secret from kubernetes: Unauthorized
The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.
```
Please see details in this thread https://github.com/harvester/harvester/issues/1823

**To Reproduce**
1. Create an RKE2 cluster with image JeOS on Harvester.
2. When provisioning is finished, add a node to the cluster.

**Result**
The new provisioning job may fail and the machine is stuck in `Provisioning` status.

**Expected Result**
The machine can be created successfully.

**Additional context**
https://github.com/harvester/harvester/issues/1823

TachunLin:
The issue is not yet been fixed according to the related issue comment: https://github.com/harvester/harvester/issues/1823#issuecomment-1049501074

TachunLin:
Verified **fixed** on master-088a020f-head. (03/06) with Rancher `v2.6-head`. (`33db2f5`). Close this issue.

 ### Result
* We can scale pool up by clicking + to add node to RKE2 cluster by node driver
![image](https://user-images.githubusercontent.com/29251855/157243631-19d64790-f726-41ca-83f9-80740852035c.png)
![image](https://user-images.githubusercontent.com/29251855/157244484-05552ad7-9b3d-497c-9b97-2c6e9de1b5ce.png)
![image](https://user-images.githubusercontent.com/29251855/157244555-92ab6fdb-2f34-402f-9570-1659006f0bf1.png)

* Can also click - to scale pool down by deleting a node from existing cluster 
![image](https://user-images.githubusercontent.com/29251855/157249545-1a264946-4651-4643-b182-479f6574588d.png)


### Test Information
* Test Environment: 1 node1 harvester on local kvm machine
* Harvester version: master-088a020f-head. (03/06)
* Rancher version: v2.6-head. (33db2f5)
[...]

PR #36658, opened 2022-02-26
Pass Rancher certs to rancher-machine job

thedadams:
Issues:
https://github.com/rancher/rancher/issues/36626
https://github.com/rancher/rancher/issues/36547

## Problem
In order for Harvester provisioning to work in air-gap, the driver needs to be
downloaded from the Rancher server. In addition, if the Rancher server is using
a self-signed certificate, that certificate needs to be passed to the Rancher
job so it can trust it.

## Solution
This change constructs a secret with the Rancher certificate information and
passes it to the rancher-machine job.

## Testing
Provisioned a cluster in Harvester where the node driver is downloaded from a Rancher server using a self-signed certificate.


thedadams:
Note to reviewers: this will have to be updated once [this](https://github.com/rancher/machine/pull/167) change is merged and a new rancher-machine version is available.

pkg/rancher/migrations.go
```diff
@@ -196,43 +191,6 @@ func forceSystemAndDefaultProjectCreation(configMapController controllerv1.Confi
 	return createOrUpdateConfigMap(configMapController, cm)
 }
 
-// copyCAAdditionalSecret will ensure that if a secret named tls-ca-additional exists in the cattle-system namespace
-// then this secret also exists in the fleet-default namespace. This is because the machine creation and deletion jobs
-// need to have access to this secret as well.
-func copyCAAdditionalSecret(secretClient controllerv1.SecretClient) error {
```
thedadams, line 202: This is no longer needed because it will be included in the cert secret passed to the job when necessary.

paynejacob (dismissed):

thedadams:
rancher-machine has been bumped now.

paynejacob (approved):

Oats87 (approved):

Issue #36518, opened 2022-02-14
Unauthorized access on secondary control plane context


MSpencer87:
**Environmental Info:**

Provider: Amazon EC2 
Instance: t3.xlarge
AMI: ubuntu-20-docker-20.07
Tools: RKE v1.3.3
Release: Rancher v2.6.3
Kubernetes Version: v1.21.7 (Default)

Cluster Configuration:

2 Control Plane, 1 etcd, 3 workers

**Describe the bug:**

When provisioning clusters with local cluster authentication endpoint enabled and multiple nodes utilizing explicit control plane roles only one of the secondary contexts is reachable, the other(in a 2 control plane cluster) responed with "Unauthorized"

**Steps To Reproduce:**

1) Provision a node driver Amazon EC2 RKE2 cluster with multiple node pools dedicated to the control plane role
	Pool 0 - 1 Control Plane
	Pool 1 - 1 Control Plane
	Pool 2 - 1 Etcd
	Pool 3 - 1 Worker
	Pool 4 - 1 Worker
	Pool 5 - 1 Worker

2) Enabled Local Cluster Authenticaiton Endpoint, leave FQDN and CACERT blank
	View cluster's YAML and verify localClusterAuthEndpoint = true

3) Download KubeConfig for provisoned cluster view the available contexts (one for each control plane node)
	kubectl config get-contexts

3) Access cluster on each listed context (Unauthorized response)
	kubectl config use-context xxxx-xxxx
	kubectl get pods -A

**Expected behavior:**

An accessible secondary context for each control plane by enabling local cluster auth endpoint

**Actual behavior:**

Consistently unable to reach secondary context in clusters with multiple explicit control plane roles

**Additional context / logs:**

| RKE2 Verison | Single Secondary Context | Multiple Contexts | Notes |
| ------------ | -------------------------| ----------------- | ----- |
| 1.21.9 | Failed | Failed | Showing Enabled in UI |
| 1.21.8+rke2r2 | Passed | Failed | One secondary context available |
| 1.22.6 | Pass | Pass/Failed | Consistently seeing only one secondary context availble, intermittent |
| 1.23.3 | Pass | Pass/Failed | Consistently seeing only one secondary context availble, intermittent |

thedadams:
There was a change made that Only started clusterauthtoken controller if ACE is enabled. This is where I would start investigating this issue.

https://github.com/rancher/rancher/pull/35520

paynejacob:
[...]

PR #36659, opened 2022-02-26
restart daemonsets.apps kube-api-auth to resolve the unauthorized-user issue

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/36518

Problem:
If we create an RKE/RKE2 downstream cluster with multiple control plane nodes and with ClusterAuthEndpoint enabled, then use the secondary contexts in the kubeconfig file downloaded from Rancher UI to access the cluster, we will hit the following error with one of the secondary contexts:
```
error: You must be logged in to the server (Unauthorized)
```
meanwhile, the logs of the `kube-api-auth` pod on this control plane shows the following kind of error:
```
time="2022-02-24T21:11:46Z" level=info msg="  ...looking up token for kubeconfig-user-qmk2kdmjlt"
time="2022-02-24T21:11:46Z" level=error msg="clusteruserattributes.cluster.cattle.io \"cattle-system/user-qmk2k\" not found"
```

Solutions:
Two parts are fixed:
- in rancher/rancher, we add a restart for the pods in the daemonSet `kube-api-auth`,  a restart can mitigate the problem sometimes.
- in rancher/kube-api-auth, we update all dependencies to the latest version. A bug around cache-not-start-properly was fixed in rancher/lasso and that bug could be the cause of this issue.  [PR](https://github.com/rancher/kube-api-auth/pull/11)

Tests:
Both RKE and RKE2 clusters are tested on a Rancher setup with fixes using the customized `kube-api-auth` image.

### DIFF
--- a/pkg/controllers/management/clusterdeploy/clusterdeploy.go
+++ b/pkg/controllers/management/clusterdeploy/clusterdeploy.go
@@ -334,18 +334,33 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
 			return cluster, errors.WithMessage(types.NewErrors(err, errors.New(formatKubectlApplyOutput(string(output)))), "Error while applying agent YAML, it will be retried automatically")
 		}
 		v32.ClusterConditionAgentDeployed.Message(cluster, string(output))
+
+		var restartKubeAPIAuth = true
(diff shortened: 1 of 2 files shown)


paynejacob (changes requested):
I would like to discuss this further, adding a restart based on a log message has a strong smell to me.  We need to fully understand the root cause of this issue and deal with the problem directly.

PR #36670, opened 2022-02-28
feat: bump rke to include external controlplane lb feature

aslafy-z:
See https://github.com/rancher/rke/pull/2853

```
go mod download; go mod tidy; pushd pkg/apis; go mod download; go mod tidy; popd
```


aslafy-z:
No stale

Issue #36660, opened 2022-02-26
Unable to take s3 enabled etcd snapshots on rke1 clusters


anupama2501:
Information about the Cluster
**Rancher Server Setup**
- Rancher version: v2.6-head d0d20a7
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1, v1.22.6, 1.3.7
- Proxy/Cert Details: Selfsigned

**Information about the Cluster**
- Kubernetes version: v1.22.6
- Cluster Type (Local/Downstream): `Downstream node driver 1 worker, 1 etcd, 1 cp RKE1`

**Describe the bug**
Not able to take a snapshot when s3 snapshots are enabled. This happens when a cluster is created with an rke template with s3 snapshots enabled. Happens only with rke1 clusters
Snapshots with the same configs work on 2.6.2 version of rancher.

**To Reproduce**
1. Create an rke1 node driver cluster with s3 enabled.
2. Once the cluster is active, from cluster management --> snapshots --> take a snapshot

**Result**
Snapshot fails with the following errors seen in the rancher logs:
```


2022/02/26 01:22:18 [INFO] [etcd] Snapshots configured to S3 compatible backend at [redacted] to bucket [au-s3bucket] using accesskey [redacted]
--
Fri, Feb 25 2022 5:22:18 pm | 2022/02/26 01:22:18 [INFO] [etcd] Running snapshot save once on host []
Fri, Feb 25 2022 5:22:18 pm | 2022/02/26 01:22:18 [INFO] Image [rancher/rke-tools:v0.1.78] exists on host []
Fri, Feb 25 2022 5:22:18 pm | 2022/02/26 01:22:18 [INFO] Starting container [etcd-snapshot-once] on host [], try #1
Fri, Feb 25 2022 5:22:18 pm | 2022/02/26 01:22:18 [INFO] [etcd] Successfully started [etcd-snapshot-once] container on host []
Fri, Feb 25 2022 5:22:18 pm | 2022/02/26 01:22:18 [INFO] Waiting for [etcd-snapshot-once] container to exit on host []
[...]

PR #36672, opened 2022-02-28
Fix secretmigrator assemblers and s3 snapshots

cmurphy:
An early draft of the secretmigrator controller used resource-specific
keys when creating the secrets, and when that was refactored the
assemblers were never updated, so for example the S3 secret assembler
was looking for a key 'secretKey' in the secret data, when it was
actually stored under the 'credential' key.

This also fixes the etcdbackup controller to use the assembled
BackupConfig, instead of the BackupConfig from the newly fetched Cluster
which does not have the credential on it.

https://github.com/rancher/rancher/issues/36660

### DIFF
--- a/pkg/controllers/management/etcdbackup/etcdbackup.go
+++ b/pkg/controllers/management/etcdbackup/etcdbackup.go
@@ -248,10 +248,6 @@ func (c *Controller) createBackupForCluster(b *v3.EtcdBackup, cluster *v3.Cluste
 	var err error
 	if b.DeletionTimestamp != nil || rketypes.BackupConditionCreated.IsUnknown(b) {
 		b.Spec.Filename = generateBackupFilename(b.Name, cluster.Spec.RancherKubernetesEngineConfig.Services.Etcd.BackupConfig)
-		cluster.Spec, err = secretmigrator.AssembleS3Credential(cluster, cluster.Spec, c.secretLister)
-		if err != nil {
-			return b, err
-		}
 		b.Spec.BackupConfig = *cluster.Spec.RancherKubernetesEngineConfig.Services.Etcd.BackupConfig
 		rketypes.BackupConditionCreated.True(b)
 		// we set ConditionCompleted to Unknown to avoid incorrect "active" state
@@ -350,6 +346,11 @@ func (c *Controller) etcdSaveWithBackoff(b *v3.EtcdBackup) (runtime.Object, erro
 		if err != nil {
 			return b, err
 		}
+		cluster.Spec.RancherKubernetesEngineConfig.Services.Etcd.BackupConfig = b.Spec.BackupConfig.DeepCopy()
+		cluster.Spec, err = secretmigrator.AssembleS3Credential(cluster, cluster.Spec, c.secretLister)
+		if err != nil {
+			return b, err
+		}
 		var inErr error
 		err = wait.ExponentialBackoff(backoff, func() (bool, error) {
 			if inErr = c.backupDriver.ETCDSave(c.ctx, cluster.Name, kontainerDriver, cluster.Spec, snapshotName); inErr != nil {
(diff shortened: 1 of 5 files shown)


pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -350,6 +350,7 @@ func (c *Controller) etcdSaveWithBackoff(b *v3.EtcdBackup) (runtime.Object, erro
 		if err != nil {
 			return b, err
 		}
+		cluster.Spec.RancherKubernetesEngineConfig.Services.Etcd.BackupConfig = &b.Spec.BackupConfig
```
MbolotSuse, line 353: I'm a bit unfamiliar with this code - the following questions encompass a bit more than the code that you changed:
- Is the idea with this change that when the backupDriver does the ETCDSave, it (in this case) was missing the backup config, causing an error? Therefore, in order to fix that, you temporarily (I say temporarily because I don't think that you update the cluster) set the backup config of the cluster to this backupConfig (both on the create/remove functions)?
- Is this functionality exclusive to RKE (I ask b/c this is a property of the RancherKubernetesEngineConfig)?
- The Create function appears to check that something is set in this value before calling these functions using the `isBackupSet` function. Is this change potentially overwriting another value?

MbolotSuse (commented):
Got a couple of clarifying questions.

ryansann (approved):
lgtm

pkg/controllers/management/etcdbackup/etcdbackup.go
cmurphy: > Is the idea with this change that when the backupDriver does the ETCDSave, it (in this case) was missing the backup config

It was not missing the backup config. The credential from the backup config is never stored, and so it needs to be reassembled on the object before using it. We reassemble the object [here](https://github.com/rancher/rancher/blob/58e374f9eaa0e96a76298259734348a377b1a186/pkg/controllers/management/etcdbackup/etcdbackup.go#L251) and the backup config object that we eventually use here comes from that [reassembled cluster object](https://github.com/rancher/rancher/blob/58e374f9eaa0e96a76298259734348a377b1a186/pkg/controllers/management/etcdbackup/etcdbackup.go#L255).

> I say temporarily because I don't think that you update the cluster

Yes, we never store the credential on the cluster object any more

> Is this functionality exclusive to RKE

Yes

> The Create function appears to check that something is set in this value before calling these functions using the isBackupSet function. Is this change potentially overwriting another value?

No. For one, `isBackupSet` causes `Create` to [return](https://github.com/rancher/rancher/blob/58e374f9eaa0e96a76298259734348a377b1a186/pkg/controllers/management/etcdbackup/etcdbackup.go#L99) before [this code is called](https://github.com/rancher/rancher/blob/58e374f9eaa0e96a76298259734348a377b1a186/pkg/controllers/management/etcdbackup/etcdbackup.go#L120) if there is nothing set. For another, the value it is setting comes from the [reassembled cluster itself](https://github.com/rancher/rancher/blob/58e374f9eaa0e96a76298259734348a377b1a186/pkg/controllers/management/etcdbackup/etcdbackup.go#L255). What this is doing is overcoming the fact that the original cluster object was dropped and [re-fetched](https://github.com/rancher/rancher/blob/58e374f9eaa0e96a76298259734348a377b1a186/pkg/controllers/management/etcdbackup/etcdbackup.go#L349) in this function.

pkg/controllers/management/secretmigrator/assemblers.go
```diff
@@ -101,7 +101,7 @@ func AssembleSMTPCredential(notifier *apimgmtv3.Notifier, secretLister v1.Secret
 		return &notifier.Spec, err
 	}
 	spec := notifier.Spec.DeepCopy()
-	spec.SMTPConfig.Password = string(smtpSecret.Data["password"])
+	spec.SMTPConfig.Password = string(smtpSecret.Data[secretKey])
```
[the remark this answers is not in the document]
MbolotSuse: Ok - I think I understand this better now. Thank you for explaining.

pkg/controllers/management/etcdbackup/etcdbackup.go
MbolotSuse: Ok - I see. So we re-fetch the cluster, which removes the backupconfig secrets options. Since we set this value on the backup earlier in another function, we just re-fetch it from the object (where we have it saved). 

Do we need to do something similar for the remove call? I don't see an assemble S3 credential call anywhere in the call chain for removing the backup.

ryansann (approved):

PR #36673, opened 2022-02-28
Added support to provisioning rke2 azure node driver clusters

igomez06:
**Framework updates** 
- Helper for creating a cloud credential for azure 
- Helper for creating a machine config for an azure provider.
- Updates to spelling mistakes in comments in the other cloud credential helpers

**Validation updates**
- Added azure to providers
- Updates to method names to adhere to go lang naming conventions

### DIFF
--- a/tests/framework/extensions/machinepools/azure_machine_config.go
+++ b/tests/framework/extensions/machinepools/azure_machine_config.go
@@ -1,0 +1,73 @@
+package machinepools
+
+import (
+	"github.com/rancher/rancher/tests/framework/pkg/config"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	AzureKind                              = "AzureConfig"
+	AzurePoolType                          = "rke-machine-config.cattle.io.azureconfig"
+	AzureResourceConfig                    = "azureconfigs"
+	AzureMachingConfigConfigurationFileKey = "azureMachineConfig"
+)
+
+type AzureMachineConfig struct {
+	AvailabilitySet   string   `json:"availabilitySet" yaml:"availabilitySet"`
+	DiskSize          string   `json:"diskSize" yaml:"diskSize"`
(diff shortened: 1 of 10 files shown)


paynejacob (approved):

tests/framework/extensions/machinepools/azure_machine_config.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+const (
+	AzureKind                              = "AzureConfig"
+	AzurePoolType                          = "rke-machine-config.cattle.io.azureconfig"
+	AzureResourceConfig                    = "azureconfigs"
+	AzureMachingConfigConfigurationFileKey = "azureMachineConfig"
+)
+
+type AzureMachineConfig struct {
```
izaac, line 15: Are we sure these are all the possible parameters the config could receive?
Nit: add comment to the specific upstream configuration

izaac (changes requested):
Looks good so far but I'd prefer to have the `AzureMachineConfig` (and potentially any other config object?) linked to their documentation in case we need a quick reference on what or how to format the values of each object.

tests/framework/extensions/machinepools/azure_machine_config.go
igomez06: Yes these are all the parameters. Is there something I am possibly missing Isaac? I just checked again on rancher, and this is it, unless I'm not seeing something. Other question what do you by "a comment to the specific upstream configuration" like a windows vs linux? Would be better in the actual test?

izaac: Oh I thought this was coming from an upstream library but is it the Rancher API? In that case we're good.

igomez06: Yes this is in Rancher API, ok sounds good thanks for checking.

izaac (approved):
LGTM

PR #36688, opened 2022-03-01
Fix small bug in rke2 airgap creation tests

rancher-max:
Allow the following config param to work correctly during tarball install:
`--cni=multus,<calico|cilium>` 

Before this change, it would only allow multus to be used with canal during tarball install in airgap.


ShylajaDevadiga (dismissed):
LGTM

ShylajaDevadiga (dismissed):

bmdepesa (approved):

ShylajaDevadiga (approved):
LGTM

Issue #36483, opened 2022-02-11
[v2prov] General fortification/reliability enhancements for v2 provisioning framework


Oats87:
The v2prov framework has come quite a long way since its inception. In preparation for the 2.6.4 release, we must ensure that it is robust/reliable/fortified and this is an (unofficial epic) issue to collect all of the PRs for provisioning framework improvements as we find them/create them.

snasovich:
Closing this as individual issues will be created as necessry.

PR #36701, opened 2022-03-01
Handle machineprovision edge cases around jobs

Oats87:
https://github.com/rancher/rancher/issues/36483

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/common.go
+++ b/pkg/controllers/provisioningv2/rke2/common.go
@@ -19,6 +19,7 @@ import (
 	"github.com/rancher/wrangler/pkg/condition"
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/name"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
--- a/pkg/controllers/provisioningv2/rke2/common.go
+++ b/pkg/controllers/provisioningv2/rke2/common.go
@@ -254,7 +254,8 @@ func GetMachineDeletionStatus(machineCache capicontrollers.MachineCache, cluster
 	})
 	for _, machine := range machines {
 		if machine.Status.FailureReason != nil && *machine.Status.FailureReason == capierrors.DeleteMachineError {
-			return "", fmt.Errorf("error deleting machine [%s], machine must be deleted manually", machine.Name)
+			logrus.Infof("[rke2common] Error while deleting machine %s, machine must be deleted manually", machine.Name)
+			continue
 		}
 		return fmt.Sprintf("waiting for machine [%s] to delete", machine.Name), nil
 	}
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
@@ -3,6 +3,7 @@ package machineprovision
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -104,7 +105,7 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		},
 	}
 	machine, err := rke2.GetMachineByOwner(h.machines, infraObj.meta)
-	if err != nil {
+	if err != nil && (create || !errors.Is(err, rke2.ErrNoMachineOwnerRef)) {
 		return driverArgs{}, err
 	}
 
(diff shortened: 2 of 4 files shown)


jakefhyde (dismissed):

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -104,7 +105,7 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		},
 	}
 	machine, err := rke2.GetMachineByOwner(h.machines, infraObj.meta)
-	if err != nil {
+	if err != nil && (create || !errors.Is(err, rke2.ErrNoMachineOwnerRef)) {
```
thedadams, line 108: I'm trying to understand a situation in which we would have an infrastructure machine that has no CAPI Machine owner. The CAPI controller creates the infrastructure machine from the machine.

Is there a bigger problem here?

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -328,7 +328,20 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
+			if job != nil {
+				// enqueue the job to force-reconcile the condition
+				h.jobController.Enqueue(job.Namespace, job.Name)
+			}
+			return obj, fmt.Errorf("cannot delete machine %s because create job has not finished", infraObj.meta.GetName())
+		}
+		// If the job is not found, go ahead and proceed with machine deletion
+		if apierrors.IsNotFound(err) {
+			return obj, nil
+		}
```
thedadams, line 344:
```suggestion
		if apierrors.IsNotFound(err)) {
			// If the job is not found, go ahead and proceed with machine deletion
			return obj, nil
		} else if err != nil {
			return obj, err
		}
		logrus.Debugf("[MachineProvision] create job for %s not finished, job was found and the error was not nil and was not an isnotfound", key)
		logrus.Tracef("[MachineProvision] create job object for %s was %+v", key, job)

		// enqueue the job to force-reconcile the condition
		h.jobController.Enqueue(job.Namespace, job.Name)
		}
		
		logrus.Infof("[MachineProvision] cannot delete machine %s because job %s is not finished", infraObj.meta.GetName(), job.Name)
		return obj, generic.ErrSkip)
		}
```

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
Oats87: The Machine objects and corresponding InfrastructureMachine objects are created by the CAPI MachineSet controller. When the MachineSet controller creates the corresponding InfrastructureMachine, it does not set the owner reference at that time to the Machine.

The CAPI Machine controller is responsible for "adopting" the InfrastructureMachine object object under the Machine. If the MachineDeployment/MachineSet are updated/a reconcile is triggered causing new machines + infrastructuremachines to be spawned but the Machine controller does not adopt the InfrastructureMachine in time before the Machine object goes into a deleting state, we can enter the edge case scenario where the corresponding infrastructuremachine object is not owned by the machine it was created for.

This poses problems as we can enter a "chicken-and-egg" scenario where the infrastructuremachine cannot be deleted due to the rancher machineprovision-remove finalizer, but we won't remove our finalizer because the infrastructure machine object doesn't have the owner references set on it. The MachineDeployment, MachineSet, and Machine will all be blocked on deletion until this condition is cleared.

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -254,7 +255,8 @@ func GetMachineDeletionStatus(machineCache capicontrollers.MachineCache, cluster
 	})
 	for _, machine := range machines {
 		if machine.Status.FailureReason != nil && *machine.Status.FailureReason == capierrors.DeleteMachineError {
-			return "", fmt.Errorf("error deleting machine [%s], machine must be deleted manually", machine.Name)
+			logrus.Infof("[rke2common] Error while deleting machine %s, machine must be deleted manually", machine.Name)
+			continue
```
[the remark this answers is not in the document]
Oats87: For the sake of not having an argument over this, I will revert the behavior to the prior state. We really need to handle this more elegantly IMO.

Oats87:
Force push required as I rebased on the updated `release/v2.6`

kinarashah (dismissed):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -328,6 +328,19 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 	}
 
 	if !infraObj.data.Bool("status", "jobComplete") && infraObj.data.String("status", "failureReason") == "" {
+		job, err := h.getJobFromInfraMachine(infraObj)
+		if apierrors.IsNotFound(err) {
+			// If the job is not found, go ahead and proceed with machine deletion
+			return obj, nil
```
thedadams, line 334: I just realized that we probably want to do the following to ensure there are no extraneous objects left behind:
```suggestion
			return obj, h.apply.WithOwner(obj).ApplyObjects()
```

thedadams (changes requested):

thedadams (approved):

Issue #35899, opened 2021-12-16
Standard user loses access to provisioned RKE2 cluster before cluster is deleted


thedadams:
**Rancher Server Setup**
- Rancher version: 2.6-head
- Installation option (Docker install/Helm Chart): Docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: 1.21.7
- Cluster Type (Local/Downstream): K3S
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE2 provisioned cluster

**Describe the bug**
If a standard user deletes a cluster they created, the roles and rolebindings that gives the user access to the cluster are deleted before the cluster is cleaned up. That means, if deletion fails for some reason, the user will never know because they will no longer have access to this cluster. As an aside, the user will also need to refresh the UI for the cluster to go away. This is because the UI will try to refresh the status of the cluster, but will be unable to.

**To Reproduce**
As a standard user:
1. Create an RKE2 cluster
2. Wait for the cluster to become active.
3. Delete the cluster.

**Result**
The UI will seem to "hang." Refreshing the page will make the cluster disappear from the UI, even if the cluster has not been deleted.

**Expected Result**
The UI should have status updates and the user should not lose access to the cluster until it is cleaned up.

**Additional context**
This was found while investigating [this](https://github.com/rancher/rancher/issues/35868) issue.

jakefhyde:
@thedadams Did you look into this issue? Is there an approach you have in mind, or possibly something you are already working on?

Josh-Diamond:
Ticket #35899 - Test Results

With Docker and 5 node cluster (1 etcd, 1 control plane, 3 worker):

Reproduced on 2.6.3:
Fresh install of rancher 2.6.3
Create a standard user
Log into rancher with standard user
As standard user, create an RKE2 provisioned 5 node custom cluster (k8s v1.21.9+rke2r1)
Delete RKE2 provisioned custom cluster, as standard user
Issue reproduced - UI stuck in “hang” state and does not update status of cluster until page is manually refreshed

<img width="1484" alt="Screen Shot 2022-03-08 at 9 30 54 AM" src="https://user-images.githubusercontent.com/46494969/157498573-611091cd-6ab9-418f-ac4a-25658b54d6b7.png">


[...]

PR #36706, opened 2022-03-01
ensure the rbac finalizer is the last finalizer to be removed

paynejacob:
#35899 

# Problem
When a cluster is deleted the rbac is deleted before the cluster finishes deleting.  This prevents standard users from seeing the resource delete event and can hide clusters that encounter errors deleting.

# Solution
Delete rbac as the last finalizer to ensure the cluster deleted correctly before deleting it.

### DIFF
--- a/pkg/controllers/management/auth/project_cluster_handler.go
+++ b/pkg/controllers/management/auth/project_cluster_handler.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/systemaccount"
 	"github.com/rancher/rancher/pkg/types/config"
+	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/sirupsen/logrus"
 	v12 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


Oats87 (dismissed):

jakefhyde (dismissed):

pkg/controllers/management/auth/project_cluster_handler.go
```diff
@@ -228,6 +228,10 @@ func (l *clusterLifecycle) Updated(obj *v3.Cluster) (runtime.Object, error) {
 }
 
 func (l *clusterLifecycle) Remove(obj *v3.Cluster) (runtime.Object, error) {
+	if len(obj.Finalizers) > 1 {
+		return obj, errors.New("waiting to be the last finalizer")
```
thedadams, line 232: Can we use `generic.ErrSkip` instead so a log doesn't get out and we push retry to its max value.

thedadams (changes requested):

thedadams (approved):

jakefhyde (dismissed):

Oats87 (approved):

thedadams (approved):

Issue #36333, opened 2022-01-31
Creating RKE cluster with rotateEncryptionKey set to true crashes rancher


jiaqiluo:
**Rancher Server Setup**
- Rancher version: 2.6.3
- Installation option (Docker install/Helm Chart): HA, Hosted Rancher 
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1

**Issue description:**
If you create a custom RKE cluster with rotateEncryptionKey set to true, rancher pods will crash and stay in crash loop back-off

**Business impact:**
This can result in an outage of rancher

**Repro steps:**
 * Create cluster
 * Set a name, click Edit as YAML
 * change rotateEncryptionKey: false to rotateEncryptionKey:true
 * Save
 * Run registration command (worker/etcd/cp) on a fresh VM

**Workaround**
 Is workararound available and implemented? yes
 What is the workaround: edit the clusters.management.cattle.io object and set the rotateEncryptionKey to false.

**Actual behavior:**
rancher crashes, logs:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2a5edfe]goroutine 627123 [running]:
github.com/rancher/rke/cluster.InitClusterObject(0x4cd9e60, 0xc013330c60, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc00d55eea0, 0x2e, ...)
	/go/pkg/mod/github.com/rancher/rke@v1.3.3/cluster/cluster.go:705 +0xde
github.com/rancher/rke/cmd.RotateEncryptionKey(0x4cd9e60, 0xc013330c60, 0x0, 0xc00c3953a0, 0xc00c3953b0, 0xc017dcc660, 0x0, 0x0, 0x0, 0x0, ...)
	/go/pkg/mod/github.com/rancher/rke@v1.3.3/cmd/encryption.go:85 +0x185
github.com/rancher/rke/cmd.ClusterUp(0x4cd9e60, 0xc013330c60, 0xc00c3953a0, 0xc00c3953b0, 0xc017dcc660, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/pkg/mod/github.com/rancher/rke@v1.3.3/cmd/up.go:107 +0x35b
[...]

PR #36708, opened 2022-03-01
[2.6.4] improve the error message for encryption key rotation error.

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/36333

Problem:
RKE has the limitation that we cannot enable the Encryption Key Rotation and Secrets Encryption at the same time when editing an existing cluster. Rancher UI shows the following message that lacks details when we do that. 
  

Solution:
1. The error message returned from the underlying RKE is now appended to the message, so users can see exactly what goes wrong and how to resolve the error.  
2. We need bump the rke dependent after  https://github.com/rancher/rke/pull/2854 is merged 

Tests:
Steps:
- provision a one-node RKE1 cluster
- wait for the cluster to be ready 
- edit the cluster -> edit as YAML -> set `rotateEncryptionKey: true` and `secretsEncryptionConfig.enable: true`

Results:
 the following message shows in Rancher UI: 
`Encryption key rotation failed, please restore your cluster from backup if needed: can't rotate encryption keys: Secrets Encryption needs to be enabled first. Please disable Key Rotation and update the cluster`
 

Screen Shot 2022-03-01 at 4 45 24 PM


jiaqiluo:
Closed in favor of a better solution.

PR #36709, opened 2022-03-02
Fix timing issue for encryption key rotation with few control plane nodes

jakefhyde:
Fixed the case where there is only a single control plane node running in rancher. In this configuration, the `rke2-server` does not need to be restarted, only the `kube-apiserver` pod. Also made some general improvements to leader election as well as provisioning, and ~added a forced wait between rotation phases because the leader will get stuck on `reencrypt_active` if the commands are run too quickly. I am currently experimenting with lowering this.~ Additionally, if a control plane requires re-provisioning, it will not happen until encryption key rotation either succeeds or fails. Phases were also culled from `Start`, `RestartLeader`, `VerifyLeaderStage`, `RestartFollowers,` and `ApplyLeader` to `RestartNodes` and `ApplyLeader`, given that the necessary prerequisites for encryption key rotation to be successful (i.e. a functional apiserver and an ability to get the `secrets-encrypt status`) were identified, and constructs put in place to ensure that this is achieved before encryption key rotation will proceed (which could have the potential to cause encryption key rotation to get stuck).

## Update

A bug was also fixed where encryption key rotation would fail if `systemctl restart rke2-server.service` was run on a control plane node that was not also the leader. The leader for running the `secrets-encrypt` command must also be the leader, unless in a split role setup, where any etcd node must be restarted prior to the control plane nodes being restarted in order for the encryption key changes to take effect.   

Related Issue: https://github.com/rancher/rancher/issues/35436


jakefhyde:
@thedadams There is still some weirdness in this PR regarding Conditions on the control plane, as well as general concerns regarding individual machine statuses, but some of those concerns have considerable overlap with this issue https://github.com/rancher/rancher/issues/36993

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -376,8 +376,7 @@ func (p *Planner) Process(controlPlane *rkev1.RKEControlPlane) error {
 		return ErrWaiting(firstIgnoreError.Error())
 	}
 
-	if controlPlane.Spec.RotateEncryptionKeys != nil &&
-		controlPlane.Status.RotateEncryptionKeysGeneration != controlPlane.Spec.RotateEncryptionKeys.Generation {
+	if rke2.Updated.IsUnknown(controlPlane) && rke2.Updated.GetReason(controlPlane) == rke2.RotatingEncryptionKeysReason {
```
annablender, line 379: I'm reading this as- before, if we aren't on the correct rotation key generation then requeue the rotation. After your change, if the cp node has not finished updating because encryption key rotation is in progress, requeue the rotation. This doesn't sound right. Why do we need to requeue if the current rotation just isn't finished? (or help me understand the logic here).

pkg/provisioningv2/rke2/planner/s3args.go
```diff
@@ -53,7 +53,7 @@ func (s *s3Args) ToArgs(s3 *rkev1.ETCDSnapshotS3, controlPlane *rkev1.RKEControl
 		args = append(args, fmt.Sprintf("--%ss3-access-key=%s", s.prefix, s3Cred.AccessKey))
 	}
 	if s3Cred.SecretKey != "" {
-		if s.env {
+		if s.env && false {
```
thedadams, line 56: Good catch!

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -65,10 +110,16 @@ func (p *Planner) rotateEncryptionKeys(cp *rkev1.RKEControlPlane, clusterPlan *p
...
 func shouldRotateEncryptionKeys(cp *rkev1.RKEControlPlane) bool {
-	// The controlplane must be initialized before we rotate anything
-	if !cp.Status.Initialized {
+	// If we haven't started rotating yet, the cluster must be ready before we start
+	if rke2.Updated.IsUnknown(cp) && rke2.Updated.GetReason(cp) == rke2.RotatingEncryptionKeysReason {
+		return true
+	}
+	// Do not stop rotating for any reason, except that encryption key rotation cannot be completed
+
+	if !rke2.Ready.IsTrue(cp) {
```
thedadams, line 119: nit: combine this check with the one below.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -104,29 +155,142 @@ func hasValidRotateEncryptionKeyStatus(cp *rkev1.RKEControlPlane) bool {
...
+			apiserverID := string(leader.Plan.Output[encryptionKeyRotationRestartApiserverCommand])
+			apiserverID = strings.TrimSuffix(apiserverID, "\n")
+			if len(apiserverID) != 13 { // not a valid container id
+				return p.enqueueAndSkip(cp)
+			}
+			cp.Annotations[encryptionKeyApiserverID] = apiserverID
+			_, err = p.rkeControlPlanes.Update(cp)
+			if err != nil {
+				return err
+			}
```
thedadams, line 272: Do we want to return a `generic.ErrSkip` if the update was successful?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -104,29 +155,142 @@ func hasValidRotateEncryptionKeyStatus(cp *rkev1.RKEControlPlane) bool {
...
+		}
+
+		apiserverStatus := string(leader.Plan.PeriodicOutput[encryptionKeyRotationApiserverStatusCommand].Stdout)
+		apiserverStatus = strings.TrimSuffix(apiserverStatus, "\n")
```
thedadams, line 276: nit:
```suggestion
		apiserverStatus := strings.TrimSuffix(string(leader.Plan.PeriodicOutput[encryptionKeyRotationApiserverStatusCommand].Stdout), "\n")
```

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -104,29 +155,142 @@ func hasValidRotateEncryptionKeyStatus(cp *rkev1.RKEControlPlane) bool {
...
+		if apiserverStatus == "" {
+			return p.enqueueAndSkip(cp)
+		}
+		if cp.Annotations[encryptionKeyApiserverID] == apiserverStatus {
+			return p.enqueueAndSkip(cp)
+		}
+		if len(apiserverStatus) != 13 { // not a valid container id
```
thedadams, line 283: Could you add a longer comment explaining this? Is there a different check we could do that is more intuitive here?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -104,29 +155,142 @@ func hasValidRotateEncryptionKeyStatus(cp *rkev1.RKEControlPlane) bool {
...
-		Instructions: []plan.OneTimeInstruction{encryptionKeyRotationRestartInstruction(cp, leader, phase)},
+
+	nodes := collect(clusterPlan, encryptionKeyRotationIsSuitableControlPlane)
+	if len(nodes) == 0 {
+		return p.restartEncryptionKeyRotation(cp) // actually should bail out here, not even a leader
 	}
-	err = assignAndCheckPlan(p.store, fmt.Sprintf("encryption key rotation: [%s]", cp.Status.RotateEncryptionKeysPhase), leader, nodePlan, 0)
-	if err != nil {
-		return p.enqueueIfErrWaiting(cp, err)
+	if len(nodes) == 1 {
```
thedadams, line 234: Does the fix in RKE2 mean we don't have to handle single control-plane nodes differently?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -146,46 +315,49 @@ func (p *Planner) encryptionKeyRotationVerifyLeaderStatus(cp *rkev1.RKEControlPl
...
+	leaderStatus, err := encryptionKeyRotationSecretsEncryptStatusFromPeriodic(leader)
 	if err != nil {
 		return p.restartEncryptionKeyRotation(cp)
 	}
 
-	for _, entry := range collect(clusterPlan, isControlPlaneAndNotInitNode) {
+	isControlPlaneAndNotLeader := func(entry *planEntry) bool {
+		return encryptionKeyRotationIsSuitableControlPlane(entry) &&
+			cp.Annotations[encryptionKeyRotationLeaderAnnotation] != entry.Machine.Name
+	}
```
thedadams, line 326: Can this be a standard function?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -322,27 +496,27 @@ func encryptionKeyRotationPhaseFromOutput(plan *planEntry, output string) (strin
...
-	logrus.Debugf("Rotate encryption Status [%s]: %s", plan.Machine.Name, phase)
-	return phase, nil
+	status := a[1]
+	return status, nil
 }
 
-func encryptionKeyRotationSecretsEncryptInstruction(cp *rkev1.RKEControlPlane, phase string) (plan.OneTimeInstruction, error) {
+// encryptionKeyRotationSecretsEncryptInstruction generates a secrets-encrypt command to run on the leader node given
+// the current secrets-encrypt status.
+func encryptionKeyRotationSecretsEncryptInstruction(cp *rkev1.RKEControlPlane, current string) (plan.OneTimeInstruction, error) {
```
thedadams, line 505:
```suggestion
func encryptionKeyRotationSecretsEncryptInstruction(cp *rkev1.RKEControlPlane, currentPhase string) (plan.OneTimeInstruction, error) {
```

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -427,15 +649,58 @@ func encryptionKeyRotationPhaseIsExpectedFromCommand(leaderPhase string, leader
...
 	return false, fmt.Errorf("could not extract secrets-encrypt command from plan: unknown command [%s]", command)
 }
+
+// encryptionKeyRotationFailed updates the various status objects on the control plane indicating a successful encryption key rotation.
+func (p *Planner) encryptionKeyRotationFinished(cp *rkev1.RKEControlPlane) error {
+	rke2.Updated.True(cp)
+	cp.Status.RotateEncryptionKeysGeneration = cp.Spec.RotateEncryptionKeys.Generation
+	cp.Status.RotateEncryptionKeysPhase = rkev1.RotateEncryptionKeysPhaseDone
+
+	var err error
```
thedadams, line 672: nit: not needed.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -427,15 +649,58 @@ func encryptionKeyRotationPhaseIsExpectedFromCommand(leaderPhase string, leader
...
+	return generic.ErrSkip
+}
+
+// encryptionKeyRotationFailed updates the various status objects on the control plane, allowing the cluster to
+// continue the reconciliation loop. Encryption key rotation will not be restarted again until requested.
+func (p *Planner) encryptionKeyRotationFailed(cp *rkev1.RKEControlPlane) error {
+	rke2.Updated.False(cp)
+	cp.Status.RotateEncryptionKeysPhase = rkev1.RotateEncryptionKeysPhaseFailed
+
+	var err error
```
thedadams, line 688: nit: not needed.

thedadams (changes requested):
A few comments.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
jakefhyde: We still have to treat single control-plane nodes different (restarting kube-apiserver container for single node, and `systemctl restart rke2-server` for multiple). The fix on the rke2 side was for control-plane nodes that did not also have the etcd role.

pkg/provisioningv2/rke2/planner/planner.go
annablender: I think I'm following. So this code

`if rke2.Updated.IsUnknown(controlPlane) && rke2.Updated.GetReason(controlPlane) == rke2.RotatingEncryptionKeysReason`

Basically, checks that if the cp node went down and RotatingEncryptionKeysReason is set (meaning the cp node crashed during encryption key rotation), then requeue the rotation?

annablender (approved):
LGTM, with changes that Donnie requested

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
jakefhyde: I would like to actually, I believe a container ID is always a hex string of 13 characters, but honestly I'd rather use a library but I haven't been able to find one for this purpose. I simply went with the length 13 check because initially I had not set the correct parameters for the command, so it was returning a very big string instead.

thedadams (dismissed):

thedadams (dismissed):

jakefhyde:
Just pushed one final update to re-run plans on followers if encryption key rotation was restarted.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -31,46 +77,46 @@ const (
 // match the init node, and either finishing if the status is reencrypt_finished, or applying the next secrets-encrypt
 // command.
 func (p *Planner) rotateEncryptionKeys(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan) error {
+	// ensure operating on latest control plane object
```
Oats87, line 80: This breaks convention. Additionally, we will be retrieving the latest `rkeControlPlane` from the cache vs. from directly against the API server IIRC, so this isn't even guaranteed to be the latest `rkeControlPlane`. 

If the goal is to always ensure we are only operating on the latest version of the `rkeControlPlane` object, we can store/compare the `resourceVersion` of the objects, although that may pose problems when rollback/recreation of the rkecontrolplane object occurs. We can also simply retrieve the latest `rkeControlPlane` object from the api server and compare resourceVersions that way, but still, not ideal I don't think.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -31,46 +77,46 @@ const (
...
+	leader, err := p.encryptionKeyRotationElectLeader(cp, clusterPlan)
 	if err != nil {
 		return err
 	}
-	if leader == nil {
-		logrus.Error("Could not rotate encryption keys: unable to find init node")
-		return nil
-	}
+
+	logrus.Infof("rkecluster %s/%s current encryption key rotation phase command: [%s]", cp.Namespace, cp.Spec.ClusterName, cp.Status.RotateEncryptionKeysPhase)
```
Oats87, line 99: Should this be an `Info` level message?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -31,46 +77,46 @@ const (
...
 
 	return p.enqueueAndSkip(cp)
 }
 
 // shouldRotateEncryptionKeys `true` if encryption key rotation has been requested or has not finished
 func shouldRotateEncryptionKeys(cp *rkev1.RKEControlPlane) bool {
-	// The controlplane must be initialized before we rotate anything
-	if !cp.Status.Initialized {
-		return false
+	// Do not stop rotating for any reason, except when encryption key rotation cannot be completed
```
Oats87, line 113: I think this function can be consolidated

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
 		cp.Status.RotateEncryptionKeysPhase == rkev1.RotateEncryptionKeysPhaseDone
 }
 
+// restartEncryptionKeyRotation restarts encryption key rotation if it is required, by resetting all related control
+// plane statuses.
 func (p *Planner) restartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) error {
-	err := p.encryptionKeyRotationUpdateControlPlanePhase(cp, rkev1.RotateEncryptionKeysPhaseStart)
+	logrus.Infof("rkecluster %s/%s restarting encryption key rotation, current phase: [%s]", cp.Namespace, cp.Spec.ClusterName, cp.Status.RotateEncryptionKeysPhase)
+
+	cp.Annotations[encryptionKeyRotationRestartTimestampAnnotation] = time.Now().Format(time.RFC3339)
```
Oats87, line 156: We probably need a deepcopy of this `cp` object before we modify it

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
 		return err
 	}
-	return generic.ErrSkip
-}
 
-func (p *Planner) encryptionKeyRotationStart(cp *rkev1.RKEControlPlane, leader *planEntry) error {
-	return p.encryptionKeyRotationGetLeaderStatus(cp, leader, rkev1.RotateEncryptionKeysPhaseRestartLeader)
+	rke2.Updated.Unknown(cp)
+	cp.Status.RotateEncryptionKeysPhase = rkev1.RotateEncryptionKeysPhaseRestartNodes
+	_, err = p.rkeControlPlanes.UpdateStatus(cp) // verify status is also updated
```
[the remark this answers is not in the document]
Oats87: I'm almost thinking we should put that `restartTimestampAnnotation` as a status field so we don't need to pull this 1-2 punch thing.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
-// subsequent runs will begin in the reencrypt_finished phase, and require the applyLeader phase to be reached at least
-// once before completion.
-func (p *Planner) encryptionKeyRotationRestartFollowers(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan, leader *planEntry) error {
-	leaderPhase, err := encryptionKeyRotationPhaseFromPeriodic(leader)
+// encryptionKeyRotationRestartNodes will restart the leader to ensure that the current encryption key rotation status
+// is persisted to the datastore. With a single node control plane, the apiserver container needs to be restarted,
+// however with an HA control plane, the leader must first be restarted then all the followers, as the followers will
+// update their status when reading from the datastore on startup of the rke2/k3s services.
+func (p *Planner) encryptionKeyRotationRestartNodes(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan, leader *planEntry) error {
+	if rke2.GetRuntime(cp.Spec.KubernetesVersion) == rke2.RuntimeRKE2 &&
```
Oats87, line 215: Can we throw this logic under a switch statement between RKE2 and K3s to help ensure that we are more clear about this? Bonus points if we just throw RKE2/K3s under their own functions and have a master function that switches between them.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
+				return err
+			}
+			return p.enqueueAndSkip(cp)
+		}
+
+		apiserverStatus := strings.TrimSuffix(string(leader.Plan.PeriodicOutput[encryptionKeyRotationApiserverStatusCommand].Stdout), "\n")
+		if apiserverStatus == "" {
+			return p.enqueueAndSkip(cp)
+		}
+		if cp.Annotations[encryptionKeyRotationApiserverIDAnnotation] == apiserverStatus {
```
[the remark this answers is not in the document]
Oats87: and the below if

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
+			return p.enqueueAndSkip(cp)
+		}
+		if len(apiserverStatus) != 13 { // not a valid container id
+			return p.enqueueAndSkip(cp)
+		}
+	} else {
+		nodePlan := plan.NodePlan{
```
Oats87, line 270: this seems to be used if we are not RKE2 or if we have an RKE2 cluster with more than 1 suitable controlplane node?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
+		err := assignAndCheckPlan(p.store, fmt.Sprintf("encryption key rotation: [%s]", cp.Status.RotateEncryptionKeysPhase), leader, nodePlan, 0)
+		if isErrWaiting(err) {
+			if strings.HasPrefix(err.Error(), "starting") {
+				logrus.Infof("rkecluster %s/%s restarting %s on %s", cp.Namespace, cp.Spec.ClusterName, rke2.GetRuntimeServerUnit(cp.Spec.KubernetesVersion), leader.Machine.Name)
+			}
+			return p.enqueueAndSkip(cp)
+		}
+	}
+
+	leaderStage, err := encryptionKeyRotationSecretsEncryptStageFromOneTimeStatus(leader)
```
Oats87, line 292: We should probably be more intelligent about this before just restarting the process otherwise we can thrash

pkg/provisioningv2/rke2/planner/store.go
```diff
@@ -360,12 +360,12 @@ func assignAndCheckPlan(store *PlanStore, msg string, server *planEntry, newPlan
...
 		return ErrWaiting(fmt.Sprintf("starting %s", msg))
 	}
-	if !server.Plan.InSync {
-		return ErrWaiting(fmt.Sprintf("waiting for %s", msg))
-	}
 	if server.Plan.Failed {
 		return fmt.Errorf("operation %s failed", msg)
 	}
+	if !server.Plan.InSync {
+		return ErrWaiting(fmt.Sprintf("waiting for %s", msg))
```
Oats87, line 367: Out of curiosity why are you moving this?

Oats87 (changes requested):
I stopped thoroughly reviewing the PR as there are major concerns I have with behavior within it.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
jakefhyde: In this case `rkeControlPlanes` is the `Client`, not the `Cache` so I think this should be the latest, right? Also I agree that it's not necessarily ideal, but from what I can tell there are too many unknowns to do it any other way given the current implementation.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
+	// a new leader was chosen, must restart
+	return nil, generic.ErrSkip
 }
 
-func (p *Planner) encryptionKeyRotationVerifyLeaderStatus(cp *rkev1.RKEControlPlane, leader *planEntry) error {
-	return p.encryptionKeyRotationGetLeaderStatus(cp, leader, rkev1.RotateEncryptionKeysPhaseRestartFollowers)
+// encryptionKeyRotationIsSuitableControlPlane ensures that a control plane node has not been deleted and has a valid
+// node associated with it.
+func encryptionKeyRotationIsSuitableControlPlane(entry *planEntry) bool {
+	return isControlPlane(entry) && entry.Machine.DeletionTimestamp == nil && entry.Machine.Status.NodeRef != nil
```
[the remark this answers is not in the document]
jakefhyde: Would that be the `Ready` condition on the machine?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +143,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
 }
 
+// restartEncryptionKeyRotation restarts encryption key rotation if it is required, by resetting all related control
+// plane statuses.
 func (p *Planner) restartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) error {
-	err := p.encryptionKeyRotationUpdateControlPlanePhase(cp, rkev1.RotateEncryptionKeysPhaseStart)
+	logrus.Infof("rkecluster %s/%s restarting encryption key rotation, current phase: [%s]", cp.Namespace, cp.Spec.ClusterName, cp.Status.RotateEncryptionKeysPhase)
+
+	cp.Annotations[encryptionKeyRotationRestartTimestampAnnotation] = time.Now().Format(time.RFC3339)
+	cp, err := p.rkeControlPlanes.Update(cp) // verify status is also updated
```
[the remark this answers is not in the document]
jakefhyde: Bad copy paste

pkg/provisioningv2/rke2/planner/store.go
jakefhyde: This was moved because I encountered situations where a plan could be not in sync, but also failed (IIRC when a plan's max-failures is changed from unlimited to a finite amount, but not the actual plan itself). This caused the plan to report out of sync, but it would never actually try again.

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -83,6 +84,8 @@ const (
 
 	RoleBootstrap = "bootstrap"
 	RolePlan      = "plan"
+
+	RotatingEncryptionKeysReason = "Rotating encryption keys"
```
thedadams, line 88: Condition Reasons do not usually contain spaces.

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -1085,12 +1084,16 @@ func (p *Planner) enqueueAndSkip(cp *rkev1.RKEControlPlane) error {
...
 			return p.enqueueAndSkip(cp)
 		}
 		return err
 	}
 	return nil
 }
+
+func isErrWaiting(err error) bool {
+	var errWaiting ErrWaiting
+	return errors.As(err, &errWaiting)
```
thedadams, line 1098:
```suggestion
	return errors.Is(err, ErrWaiting)
```

thedadams (changes requested):

pkg/provisioningv2/rke2/planner/planner.go
jakefhyde: Unfortunately I can't use `errors.Is` here because it requires also requires a struct, and we aren't uniformly wrapping a single error variable.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +136,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
+	return p.encryptionKeyRotationRestartService(cp, clusterPlan, leader)
+}
 
-	_, err := encryptionKeyRotationExtractApplyFromPlan(leader)
+func (p *Planner) encryptionKeyRotationRestartK3S(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan, leader *planEntry) error {
+	return p.encryptionKeyRotationRestartService(cp, clusterPlan, leader)
+}
+
+func (p *Planner) encryptionKeyRotationRestartSingleNodeAPIServer(cp *rkev1.RKEControlPlane, leader *planEntry) error {
+	nodePlan := plan.NodePlan{
```
annablender, line 289: If you can find a way to move this code into one helper function too, that'd be great

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -351,55 +513,124 @@ func encryptionKeyRotationSecretsEncryptInstruction(cp *rkev1.RKEControlPlane, p
...
+}
+
+// lastUpdateEnv returns an environment variable in order to force followers to rerun their plans in the vent that a
+// follower must restart however the leader status didn't change. This is possible when a follower has a state "further"
+// than the leader, however it hasn't been restarted.
+func lastUpdateEnv(cp *rkev1.RKEControlPlane) string {
+	return fmt.Sprintf("LAST_UPDATED=%s", cp.Status.RotateEncryptionKeysLastRestart)
+}
+
+// encryptionKeyRotationSecretsEncryptStatusPeriodicInstruction generates a one time instruction which will scrape the secrets-encrypt
```
annablender, line 530: Update comment to match func name

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -351,55 +513,124 @@ func encryptionKeyRotationSecretsEncryptInstruction(cp *rkev1.RKEControlPlane, p
...
 		},
 		Env: []string{
-			lastKnownLeaderPhaseEnv,
+			encryptionKeyRotationLeaderStageEnv(leaderStage),
+			lastUpdateEnv(cp),
 		},
 	}
 }
 
+func encryptionKeyRotationWaitForSecretsEncryptStatus(cp *rkev1.RKEControlPlane, leaderStage string) plan.OneTimeInstruction {
```
annablender, line 584: Add func description

thedadams (approved):

annablender (changes requested):

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +136,309 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
+		return encryptionKeyRotationIsSuitableControlPlane(entry) &&
+			cp.Annotations[encryptionKeyRotationLeaderAnnotation] != entry.Machine.Name
+	}
+}
+
+func (p *Planner) encryptionKeyRotationRestartRKE2(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan, leader *planEntry) error {
+	if len(collect(clusterPlan, encryptionKeyRotationIsSuitableControlPlane)) == 1 {
+		return p.encryptionKeyRotationRestartSingleNodeAPIServer(cp, leader)
+	}
+	return p.encryptionKeyRotationRestartService(cp, clusterPlan, leader)
```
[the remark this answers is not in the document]
jakefhyde: Yup that's right, rke2 single node is different from all the rest.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
jakefhyde: This is the helper function, not sure it can be condensed any further

paynejacob (approved):

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -31,52 +81,46 @@ const (
...
 func shouldRotateEncryptionKeys(cp *rkev1.RKEControlPlane) bool {
-	// The controlplane must be initialized before we rotate anything
-	if !cp.Status.Initialized {
-		return false
-	}
-	// if a spec is not defined there is nothing to do
-	if cp.Spec.RotateEncryptionKeys == nil {
-		return false
+	// Do not stop rotating for any reason, except when encryption key rotation cannot be completed
+	if rke2.Ready.IsUnknown(cp) && rke2.Ready.GetReason(cp) == rke2.RotatingEncryptionKeysReason {
```
thedadams, line 118: If any function during the encryption process returns an error, even an `ErrWaiting`, then this won't be true and, if I understand this code correctly, encryption key rotation will stop.

For example, if an error is returned, then the `Ready` condition will be `false`. If an `ErrWaiting` is returned (like may happen below), then the reason on the `Ready` condition will be `Waiting`.

I know you tested this, but I am concerned that if a error happens for some reason then this whole process might stall.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -290,59 +484,64 @@ func (p *Planner) encryptionKeyRotationUpdateControlPlanePhase(cp *rkev1.RKECont
...
-	output, ok := plan.Plan.PeriodicOutput[SecretsEncryptStatusCommand]
+// encryptionKeyRotationSecretsEncryptStageFromPeriodic will attempt to extract the current stage (secrets-encrypt status) from the
+// plan by parsing the periodic output.
+func encryptionKeyRotationSecretsEncryptStageFromPeriodic(plan *planEntry) (string, error) {
+	output, ok := plan.Plan.PeriodicOutput[encryptionKeyRotationSecretsEncryptStatusCommand]
 	if !ok {
-		return "", ErrWaitingf("Could not extract current status from plan for [%s]: no output for status", plan.Machine.Name)
+		for _, pi := range plan.Plan.Plan.PeriodicInstructions {
+			if pi.Name == encryptionKeyRotationSecretsEncryptStatusCommand {
+				return "", ErrWaitingf("could not extract current status from plan for [%s]: no output for status", plan.Machine.Name)
```
thedadams, line 494: An example of returning `ErrWaiting` which will cause the `Ready` condition to have a reason of `Waiting`.

thedadams (changes requested):

pkg/apis/rke.cattle.io/v1/controlplane.go
```diff
@@ -48,26 +48,26 @@ var (
...
-	Initialized                    bool                                `json:"initialized,omitempty"`
-	AgentConnected                 bool                                `json:"agentConnected,omitempty"`
+	Conditions                      []genericcondition.GenericCondition `json:"conditions,omitempty"`
+	Ready                           bool                                `json:"ready,omitempty"`
+	ObservedGeneration              int64                               `json:"observedGeneration"`
+	CertificateRotationGeneration   int64                               `json:"certificateRotationGeneration"`
+	RotateEncryptionKeysGeneration  int64                               `json:"rotateEncryptionKeysGeneration"`
+	RotateEncryptionKeysPhase       RotateEncryptionKeysPhase           `json:"rotateEncryptionKeysPhase"`
+	RotateEncryptionKeysLastRestart string                              `json:"rotateEncryptionKeysLastRestart"`
+	RotateEncryptionKeysStage       string                              `json:"rotateEncryptionKeysStage"`
```
Oats87, line 65: Can we embed these fields under a struct...?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -1,27 +1,77 @@
...
+	encryptionKeyRotationWaitForSecretsEncryptStatusScriptPath = "/var/lib/rancher/rke2/bin/wait_for_secrets_encrypt_status.sh"
+
+	// This script finds the apiserver container and restarts it.
+	// Update to use --output json from https://github.com/k3s-io/k3s/issues/5138 when implemented in rke2
+	encryptionKeyRotationRestartAPIServerScript = `
+#!/bin/sh
+
+apiServer=$(/var/lib/rancher/rke2/bin/crictl ps --name kube-apiserver | sed "1 d" | awk '{print $1}')
+
+/var/lib/rancher/rke2/bin/crictl stop $apiServer
```
Oats87, line 47: This will not work if a user specifies a custom data-dir for RKE2

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
Oats87: Yes, if `rkeControlPlanes` is the client it will be the latest; however, this is not OK for us to do and we should not introduce behavior that does this within our planner controller.

Oats87 (changes requested):

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
jakefhyde: Ok, I should be able to remove this then. Given the 2 phase system I've been able to reduce it to, running through the entirety of a phase again should effectively be a no-op because the plans are in sync.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
jakefhyde: No, moved to debug. Now that I'm using `rke2.Ready` and checking for `"starting"` in the planner `ErrWaiting` return value, it is alot clearer.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +158,374 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
 	}
 	return generic.ErrSkip
 }
 
-func (p *Planner) encryptionKeyRotationStart(cp *rkev1.RKEControlPlane, leader *planEntry) error {
-	return p.encryptionKeyRotationGetLeaderStatus(cp, leader, rkev1.RotateEncryptionKeysPhaseRestartLeader)
-}
+// encryptionKeyRotationElectLeader returns the current leader if it is valid, otherwise it selects the first suitable
+// control plane it can find.
+func (p *Planner) encryptionKeyRotationElectLeader(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan) (leader *planEntry, init *planEntry, err error) {
```
thedadams, line 189: I won't block merging on this, but I really don't like this.

In this case, we don't gain anything from having named return values in this case.

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -96,192 +158,374 @@ func shouldRestartEncryptionKeyRotation(cp *rkev1.RKEControlPlane) bool {
...
-// subsequent runs will begin in the reencrypt_finished phase, and require the applyLeader phase to be reached at least
-// once before completion.
-func (p *Planner) encryptionKeyRotationRestartFollowers(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan, leader *planEntry) error {
-	leaderPhase, err := encryptionKeyRotationPhaseFromPeriodic(leader)
-	if err != nil {
-		return p.restartEncryptionKeyRotation(cp)
+// encryptionKeyRotationRestartRKE2 restarts the nodes using encryptionKeyRotationRestartServices for an HA cluster, and
+// encryptionKeyRotationIsSuitableControlPlane if there is only one controlplane node.
+func (p *Planner) encryptionKeyRotationRestartRKE2(cp *rkev1.RKEControlPlane, clusterPlan *plan.Plan, leader *planEntry) (string, error) {
+	if len(collect(clusterPlan, encryptionKeyRotationIsSuitableControlPlane)) == 1 {
```
thedadams, line 286: nit, and you can get rid of the switch statement above:
```suggestion
	if rke2.GetRuntime(cp) == rke2.RuntimeRKE2 && len(collect(clusterPlan, encryptionKeyRotationIsSuitableControlPlane)) == 1 {
```

thedadams (approved):
Two comments, but feel free to ignore to get this in.

thedadams (approved):

Oats87 (approved):

Issue #36536, opened 2022-02-16
RKE2 upgrade doesn't work if pre-drain/post-drain hooks are enabled


bk201:
**Rancher Server Setup**
- Rancher version: `v2.6-11a7451ffee557897197fc624af03ac6fd34a9f0-head`
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): `v1.21.7+rke2r1`
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: `v1.21.7+rke2r1`
- Cluster Type (Local/Downstream): Local
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions:

   
**Describe the bug**

Upgrading an RKE2 runtime doesn't seem to work if [pre-drain and post-drain hooks](https://github.com/rancher/rancher/issues/35464) are enabled.

**To Reproduce**
 
- Use rancherd to provision an RKE2 cluster that has an embedded rancher server
- Upgrade the RKE2 cluster by editing the cluster resource:
```
kubectl edit clusters.provisioning.cattle.io local -n fleet-local
```

Update the existing spec from:
```
spec:
  kubernetesVersion: v1.21.7+rke2r1
  localClusterAuthEndpoint: {}
  rkeConfig: {}
```

to

```
spec:
  kubernetesVersion: v1.21.8+rke2r2
  localClusterAuthEndpoint: {}
  rkeConfig:
    upgradeStrategy:
      controlPlaneConcurrency: "1"
      workerConcurrency: "1"
      controlPlaneDrainOptions:
        deleteEmptyDirData: true
        enabled: true
        force: true
        ignoreDaemonSets: true
        preDrainHooks:
          - annotation: harvester.cattle.io/pre-hook
        postDrainHooks:
          - annotation: harvester.cattle.io/post-hook
      workerDrainOptions:
        deleteEmptyDirData: true
        enabled: true
        force: true
        ignoreDaemonSets: true
        preDrainHooks:
          - annotation: harvester.cattle.io/pre-hook
        postDrainHooks:
          - annotation: harvester.cattle.io/post-hook
```

**Result**
[...]

PR #36719, opened 2022-03-02
No race cache migrations

thedadams:
Issue: https://github.com/rancher/rancher/issues/36536

## Problem
Since we were using caches and indexes to try to migrate right after
Rancher starts, we we having issues where the caches and indexes aren't
populated.

## Solution
This change switches to using clients and label selectors for more
consistent results.

## Testing
I tested as in [this comment](https://github.com/rancher/rancher/issues/36536#issuecomment-1056607168) and was able to upgrade the local cluster.

### DIFF
--- a/pkg/rancher/migrations.go
+++ b/pkg/rancher/migrations.go
@@ -5,13 +5,9 @@ import (
 	"github.com/rancher/norman/condition"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/auth/tokens"
-	"github.com/rancher/rancher/pkg/controllers/provisioningv2/cluster"
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	"github.com/rancher/rancher/pkg/features"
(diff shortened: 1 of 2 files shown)


pkg/rancher/migrations.go
```diff
@@ -279,9 +268,8 @@ func addWebhookConfigToCAPICRDs(crdClient controllerapiextv1.CustomResourceDefin
 	return nil
 }
 
-func migrateCAPIMachineLabelsAndAnnotationsToPlanSecret(configMapController controllerv1.ConfigMapController, secretsController controllerv1.SecretController, mgmtClusterCache v3.ClusterCache,
-	provClusterCache provv1.ClusterCache, capiMachineCache capicontrollers.MachineCache, bootstrapClient rkecontrollers.RKEBootstrapClient) error {
-	cm, err := getConfigMap(configMapController, migrateFromMachineToPlanSecret)
+func migrateCAPIMachineLabelsAndAnnotationsToPlanSecret(w *wrangler.Context) error {
```
Oats87, line 271: so much cleaner

paynejacob (approved):

Issue #36506, opened 2022-02-13
Machine drain errors are not exposed  in the UI during cluster upgrade on an RKE2 cluster


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6head commit id: `11a7451`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: `v1.21.9+rke2r1`
- Cluster Type (Local/Downstream): Linode node driver cluster - 1 etcd, 1 cp and 3 worker nodes. Container network - `[multus-canal]`

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Standard user

**Describe the bug**
No error seen on the UI when cluster upgrade fails.

**To Reproduce:**
- Deploy an RKE2 cluster.
- When the cluster is Active, Edit Config --> Modify the Upgrade strategy to
<img width="1203" alt="Screen Shot 2022-02-13 at 3 16 49 PM" src="https://user-images.githubusercontent.com/26032343/153779639-513f1236-d738-496b-b812-63242164b6b0.png">

- Save changes made.
- Edit cluster config again. Change the k8s version to `v1.22.6+rke2r1` and save changes made.
- The cluster goes into Provisioning state. 
- Worker node is stuck in Cordoned state
- Error in rancher Logs:
```
2022/02/13 23:07:29 [ERROR] error syncing 'fleet-default/sowmya-ec2-1-bootstrap-template-b29nk-machine-plan': handler machine-drain: cannot delete Pods with local storage (use --delete-emptydir-data to override): kube-system/rke2-metrics-server-8574659c85-82sqv, requeuing
```

**Expected Result:**
There should be an error displayed on the UI why the cluster upgrade failed. User must not have to log into the server to see the error.

Oats87:
Given the difference in functionality between RKE1 and RKE2/K3s, it is very difficult to interpolate the exact reason a plan failed when updating. 

I think this is functionality that we can add down the line, otherwise, maybe we just add a "plan has failed to apply N number of times, please check rancher-system-agent.service logs" or something to indicate that the plan is failing

sowmyav27:
**On 2.6-head commit id: `024e6fa`**

- Deployed an RKE2 cluster - 3 etcd, 2 cp and 3 worker nodes
- Upgrade strategy is set on the cluster
- Edit cluster --> Upgrade k8s version from 1.22 to 1.23
[...]

PR #36725, opened 2022-03-03
Display drain errors on machine

### LINKED ISSUE #36763: RKE2 | Custom cluster | Cluster stays in Provisioning state after adding a Windows node
**Rancher Server Setup**
- Rancher version: v2.6-head 3659d2f
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.21 and v1.22
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom

  
**Describe the bug**

Cluster stays in Provisioning state after adding a Windows node even though the node is added and machine is created for it

**To Reproduce**
 
1. Create a RKE2 custom cluster (1 etcd, 1 cp, 1 worker)
2. Add a windows node 

**Result**
The cluster stays in the Provisioning state

**Expected Result**

The cluster goes into the Active state after a node is added successfully with a machine created

**Screenshots**

cluster

**Additional context**

Rancher logs:
```
2022/03/04 17:01:00 [INFO] [planner] rkecluster fleet-default/win121: waiting: provisioning worker node(s) custom-249aaa0b4d98
2022/03/04 17:02:35 [INFO] [planner] rkecluster fleet-default/win122: waiting: provisioning worker node(s) custom-5f36b9e03dbd
2022/03/04 17:02:45 [INFO] [planner] rkecluster fleet-default/win122: waiting: provisioning worker node(s) custom-5f36b9e03dbd
2022/03/04 17:07:29 [ERROR] Unknown error: Operation cannot be fulfilled on preferences.management.cattle.io "cluster": the object has been modified; please apply your changes to the latest version and try again
2022/03/04 17:08:58 [INFO] [planner] rkecluster fleet-default/windows: waiting: provisioning worker node(s) custom-fc77ed38b0d9
2022/03/04 17:09:08 [INFO] [planner] rkecluster fleet-default/windows: waiting: provisioning worker node(s) custom-fc77ed38b0d9
2022/03/04 17:10:51 [INFO] [planner] rkecluster fleet-default/win121: waiting: provisioning worker node(s) custom-249aaa0b4d98
2022/03/04 17:11:01 [INFO] [planner] rkecluster fleet-default/win121: waiting: provisioning worker node(s) custom-249aaa0b4d98
```

thedadams:
Issues:
https://github.com/rancher/rancher/issues/36506
https://github.com/rancher/rancher/issues/36782
https://github.com/rancher/rancher/issues/35999
https://github.com/rancher/rancher/issues/36776
https://github.com/rancher/rancher/issues/36763

## Problems
During provisioning and upgrading, there were no statuses being put on the machine objects. Therefore, if there was a problem with a specific machine, one would have to look at the cluster status, find the machine name, and look at that specific machine to attempt to understand what was happening. This change adds the status (to either Provisioned or Updated conditions) so the proper status is shown directly in the UI. (#36782)

Now, that we are properly showing statuses on the machine objects, we can now show errors statuses as well when they occur. That way a user doesn't have to comb through the logs or connect to the VM/node directly. (#36506)

These statuses now make using the upgrade strategy more reliable. Since the statuses are being directly put on the machines, Rancher is able to calculate the unavailable nodes in a more robust way to ensure that, for example, if a control plane node is still upgrading then the worker nodes will not upgrade. (#35999)

Also, consider the case above with etcd nodes. An etcd node is always the bootstrap node. So, when upgrading a cluster, the boostrap node would get upgraded first. After the bootstrap node is upgraded, then the remaining etcd nodes are upgraded. Since Rancher was mistakenly not waiting for nodes to be upgraded before moving onto the next group of nodes, multiple etcd nodes could be upgrading at the same time. This is dangerous and could cause data integrity issues. Thus, the robustness of determining when to move onto the next group of nodes also fixes this problem. For good measure, Rancher will now only upgrade one etcd node at a time. (#36776)

Lastly, Windows nodes report a different kubelet version from linux nodes (they don't have the +rke2rx on the end). A check for kubelet version was added in a different PR to delay upgrading worker nodes before control plane nodes were completed. This change is no longer needed with the above enhancements and this will fix #36763

## Testing
I tested provisioning, upgrading, and deleting of the following types of clusters.
- single node all roles with and without draining enabled
- one node all roles and one node worker-only with and without draining enabled
- 3 etcd nodes 2 control plane nodes and 1 worker node with draining enabled
- one node all roles and one node worker-only with draining enabled, but misconfigured. I ensured that the error appeared in the UI. Then fixed the draining misconfiguration so that draining and upgrading would continue.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
+++ b/pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
@@ -3,12 +3,9 @@ package machinedrain
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"os"
 	"time"
 
-	"github.com/rancher/lasso/pkg/cache"
-	"github.com/rancher/lasso/pkg/controller"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
@@ -20,7 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/util/workqueue"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/kubectl/pkg/drain"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
@@ -40,30 +37,11 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 		secretCache:  clients.Core.Secret().Cache(),
 	}
 
-	// It would be a bad idea to have this handler run for every secret. Therefore, a new cache factory is created just for secrets with the machine plan type.
-	cacheFactory := cache.NewSharedCachedFactory(clients.ControllerFactory.SharedCacheFactory().SharedClientFactory(), &cache.SharedCacheFactoryOptions{
-		DefaultTweakList: func(options *metav1.ListOptions) {
-			options.TypeMeta = metav1.TypeMeta{
-				APIVersion: "v1",
-				Kind:       "Secret",
-			}
-			options.FieldSelector = fmt.Sprintf("type=%s", rke2.SecretTypeMachinePlan)
-		},
-	})
-
-	controllerFactory := controller.NewSharedControllerFactory(cacheFactory, &controller.SharedControllerFactoryOptions{
-		DefaultRateLimiter: workqueue.NewItemExponentialFailureRateLimiter(1*time.Minute, 5*time.Minute),
-		DefaultWorkers:     1,
-	})
-
-	corecontrollers.New(controllerFactory).Secret().OnChange(ctx, "machine-drain", h.OnChange)
-	if err := controllerFactory.Start(ctx, 1); err != nil {
-		panic(err)
-	}
+	clients.Core.Secret().OnChange(ctx, "machine-drain", h.OnChange)
 }
 
 func (h *handler) OnChange(_ string, secret *corev1.Secret) (*corev1.Secret, error) {
-	if secret == nil || secret.DeletionTimestamp != nil || secret.Labels[rke2.MachineNameLabel] == "" {
+	if secret == nil || secret.DeletionTimestamp != nil || secret.Labels[rke2.MachineNameLabel] == "" || secret.Type != rke2.SecretTypeMachinePlan {
 		return secret, nil
 	}
 
@@ -72,21 +50,45 @@ func (h *handler) OnChange(_ string, secret *corev1.Secret) (*corev1.Secret, err
 		return secret, err
 	}
 
+	oldSecret := secret.DeepCopy()
+	defer func() {
+		if secret == nil || secret.Annotations == nil {
+			secret = oldSecret
+		}
+		drainErrorValue, ok := secret.Annotations[rke2.DrainErrorAnnotation]
+		if err == nil && !ok || err != nil && drainErrorValue == err.Error() {
+			// No need to update the machine if the annotation is already set to the correct value
+			return
+		}
+
+		secret = secret.DeepCopy()
+		if err != nil {
+			secret.Annotations[rke2.DrainErrorAnnotation] = err.Error()
+		} else {
+			delete(secret.Annotations, rke2.DrainErrorAnnotation)
+		}
+
+		var updateErr error
+		if secret, updateErr = h.secrets.Update(secret); updateErr != nil && err == nil {
+			err = updateErr
+		} else if updateErr != nil {
+			logrus.Errorf("Failed to update secret %s/%s with drain error annotation: %v", secret.Namespace, secret.Name, updateErr)
+		}
+	}()
+
 	drain := secret.Annotations[rke2.DrainAnnotation]
 	if drain != "" && secret.Annotations[rke2.DrainDoneAnnotation] != drain {
-		return h.drain(secret, machine, drain)
+		secret, err = h.drain(secret, machine, drain)
+	} else if secret.Annotations[rke2.UnCordonAnnotation] != "" {
+		// Only check that it's non-blank.  There is no correlation between the drain and unDrain options, meaning
+		// that the option values do not need to match.  For drain we track the status by doing the drain annotation
+		// and then adding a drain-done annotation with the same value when it's done.  Uncordon is different in that
+		// we want the final state to have no annotations.  So when UnCordonAnnotation is set we run and then when
+		// it's done we delete it.  So there is no knowledge of what the value should be except that it's set.
+		secret, err = h.unDrain(secret, machine, drain)
 	}
 
-	// Only check that it's non-blank.  There is no correlation between the drain and unDrain options, meaning
-	// that the option values do not need to match.  For drain we track the status by doing the drain annotation
-	// and then adding a drain-done annotation with the same value when it's done.  Uncordon is different in that
-	// we want the final state to have no annotations.  So when UnCordonAnnotation is set we run and then when
-	// it's done we delete it.  So there is no knowledge of what the value should be except that it's set.
-	if secret.Annotations[rke2.UnCordonAnnotation] != "" {
-		return h.unDrain(secret, machine, drain)
-	}
-
-	return secret, nil
+	return secret, err
 }
 
 func (h *handler) k8sClient(machine *capi.Machine) (kubernetes.Interface, error) {
@@ -116,9 +118,7 @@ func (h *handler) unDrain(secret *corev1.Secret, machine *capi.Machine, drainDat
 
 	if len(drainOpts.PostDrainHooks) > 0 {
 		if secret.Annotations[rke2.PostDrainAnnotation] != drainData {
-			secret = secret.DeepCopy()
-			secret.Annotations[rke2.PostDrainAnnotation] = drainData
-			return h.secrets.Update(secret)
+			return h.updateSecretAnnotation(secret, rke2.PostDrainAnnotation, drainData)
 		}
 		for _, hook := range drainOpts.PostDrainHooks {
 			if hook.Annotation != "" && secret.Annotations[hook.Annotation] != drainData {
@@ -164,9 +164,7 @@ func (h *handler) drain(secret *corev1.Secret, machine *capi.Machine, drainData 
 
 	if len(drainOpts.PreDrainHooks) > 0 {
 		if secret.Annotations[rke2.PreDrainAnnotation] != drainData {
-			secret = secret.DeepCopy()
-			secret.Annotations[rke2.PreDrainAnnotation] = drainData
-			return h.secrets.Update(secret)
+			return h.updateSecretAnnotation(secret, rke2.PreDrainAnnotation, drainData)
 		}
 		for _, hook := range drainOpts.PreDrainHooks {
 			if hook.Annotation != "" && secret.Annotations[hook.Annotation] != drainData {
@@ -181,9 +179,7 @@ func (h *handler) drain(secret *corev1.Secret, machine *capi.Machine, drainData 
 		}
 	}
 
-	secret = secret.DeepCopy()
-	secret.Annotations[rke2.DrainDoneAnnotation] = drainData
-	return h.secrets.Update(secret)
+	return h.updateSecretAnnotation(secret, rke2.DrainDoneAnnotation, drainData)
 }
 
 func (h *handler) cordon(machine *capi.Machine, drainOpts *rkev1.DrainOptions) error {
@@ -243,4 +239,18 @@ func (h *handler) performDrain(machine *capi.Machine, drainOpts *rkev1.DrainOpti
 	}
 
 	return drain.RunNodeDrain(helper, node.Name)
+}
+
+func (h *handler) updateSecretAnnotation(secret *corev1.Secret, annotation, value string) (*corev1.Secret, error) {
+	var err error
+	return secret, retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		secret, err = h.secrets.Get(secret.Namespace, secret.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		secret = secret.DeepCopy()
+		secret.Annotations[annotation] = value
+		_, err = h.secrets.Update(secret)
+		return err
+	})
 }
--- a/pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
+++ b/pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
@@ -246,7 +246,7 @@ func (h *handler) updateSecretAnnotation(secret *corev1.Secret, annotation, valu
 	var err error
 	return secret, retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		secret, err = h.secrets.Get(secret.Namespace, secret.Name, metav1.GetOptions{})
-		if err != nil {
+		if err != nil || secret.Annotations[annotation] == value {
 			return err
 		}
 		secret = secret.DeepCopy()
--- a/pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
+++ b/pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
@@ -3,6 +3,7 @@ package machinedrain
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"time"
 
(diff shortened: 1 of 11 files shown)


pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -35,35 +33,17 @@ type handler struct {
...
 	h := &handler{
 		ctx:          ctx,
+		machines:     clients.CAPI.Machine(),
 		machineCache: clients.CAPI.Machine().Cache(),
 		secrets:      clients.Core.Secret(),
 		secretCache:  clients.Core.Secret().Cache(),
 	}
 
-	// It would be a bad idea to have this handler run for every secret. Therefore, a new cache factory is created just for secrets with the machine plan type.
-	cacheFactory := cache.NewSharedCachedFactory(clients.ControllerFactory.SharedCacheFactory().SharedClientFactory(), &cache.SharedCacheFactoryOptions{
```
thedadams, line 44: I put this in previously because:
1. I thought I was being clever, and
2. I thought watching all secrets would be bad performance, but we are already doing that in several places.

So that is why I am removing this now.

Oats87:
Are the changes in this PR are causing provisioning times to increase?

paynejacob (approved):

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -72,21 +52,33 @@ func (h *handler) OnChange(_ string, secret *corev1.Secret) (*corev1.Secret, err
 		return secret, err
 	}
 
+	defer func() {
+		if machine == nil ||
+			err == nil && !rke2.NodeDrained.IsFalse(machine) ||
+			err != nil && rke2.NodeDrained.GetMessage(machine) == err.Error() {
```
jiaqiluo, line 58:
```suggestion
		if machine == nil ||
			(err == nil && !rke2.NodeDrained.IsFalse(machine)) ||
			(err != nil && rke2.NodeDrained.GetMessage(machine) == err.Error()) {
```

jiaqiluo: We need the `()`, right?

jiaqiluo (changes requested):

snasovich:
@thedadams , after discussing with QA we're bumping this issue until 2.6.5, so let's not merge this one yet.

pkg/provisioningv2/rke2/planner/drain.go
```diff
@@ -85,5 +92,21 @@ func (p *Planner) undrain(entry *planEntry) (bool, error) {
 		return false, p.store.updatePlanSecretLabelsAndAnnotations(entry)
 	}
 
+	if err := checkForDrainError(entry, "undraining"); err != nil {
+		// This is the only place true and an error is returned to indicate that draining is ongoing, but there is an error.
```
jakefhyde, line 96:
```suggestion
		// This is the only place true and an error is returned to indicate that undraining is ongoing, but there is an error.
```

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -244,3 +240,17 @@ func (h *handler) performDrain(machine *capi.Machine, drainOpts *rkev1.DrainOpti
 
 	return drain.RunNodeDrain(helper, node.Name)
 }
+
+func (h *handler) updateSecretAnnotation(secret *corev1.Secret, annotation, value string) (*corev1.Secret, error) {
+	var err error
+	return secret, retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
Oats87, line 246: I wonder if using a `retry` here can run us into weird edge cases where the drain/uncordon annotations are being manipulated in rapid succession.

If anything, I would almost want to do a `retry.RetryOnConflict` where we validate that the annotations we care about (the drain/uncordon ones) are in the "old" set state before we update them -- we should easily be able to pass in the current values/keys we care about to compare against.

pkg/provisioningv2/rke2/planner/drain.go
```diff
@@ -85,5 +92,21 @@ func (p *Planner) undrain(entry *planEntry) (bool, error) {
...
+		return true, err
+	}
+
+	// The annotations will be removed when undrain is done
 	return entry.Metadata.Annotations[rke2.UnCordonAnnotation] == "", nil
 }
+
+func checkForDrainError(entry *planEntry, drainStep string) error {
```
Oats87, line 104: Can we make this a godoc comment to indicate the scope of this functionality is only for looking for transient connectivity errors?

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -163,6 +165,88 @@ func New(ctx context.Context, clients *wrangler.Context) *Planner {
...
+	if waiting {
+		return ErrWaiting(messagePrefix + atMostThree(machineNames) + detailMessage(machineNames, messages))
+	}
+	return nil
+}
+
+func atMostThree(names []string) string {
+	sort.Strings(names)
+	if len(names) > 3 {
+		return strings.Join(names[:3], ",")
```
Oats87, line 213: Should we add some type of indicator here to indicate that there are in fact more than 3 `names`? like a `and more` or something

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -179,10 +263,10 @@ func (p *Planner) getCAPICluster(controlPlane *rkev1.RKEControlPlane) (*capi.Clu
...
 	logrus.Debugf("[planner] rkecluster %s/%s: attempting to lock %s for processing", controlPlane.Namespace, controlPlane.Name, string(controlPlane.UID))
 	p.locker.Lock(string(controlPlane.UID))
-	defer func() error {
-		logrus.Debugf("[planner] rkecluster %s/%s: unlocking %s", controlPlane.Namespace, controlPlane.Name, string(controlPlane.UID))
-		return p.locker.Unlock(string(controlPlane.UID))
-	}()
+	defer func(namespace, name, uid string) error {
+		logrus.Debugf("[planner] rkecluster %s/%s: unlocking %s", namespace, name, uid)
+		return p.locker.Unlock(uid)
+	}(controlPlane.Namespace, controlPlane.Name, string(controlPlane.UID))
```
Oats87, line 269: so, I'm assuming you added this because you saw a panic here; however, do you know where we are causing the `RKEControlPlane` to be set to nil...? Just for posterity purposes -- I avoided passing the parameters through the func signature because I didn't think we would be hitting a situation where `controlPlane` would become nil.

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -438,70 +510,89 @@ func (p *Planner) reconcile(controlPlane *rkev1.RKEControlPlane, tokensSecret pl
...
+		} else if entry.Machine.Status.NodeInfo != nil && !strings.HasPrefix(controlPlane.Spec.KubernetesVersion, entry.Machine.Status.NodeInfo.KubeletVersion) {
+			outOfSync = append(outOfSync, entry.Machine.Name)
+			messages[entry.Machine.Name] = append(messages[entry.Machine.Name], "waiting for kubelet to update")
+		} else {
+			ready = append(ready, entry.Machine.Name)
 		}
 	}
 
-	if required && len(entries) == 0 {
+	if len(entries) == 0 && tierName != "worker" {
```
Oats87, line 557: this change removes flexibility from the framework.... I suppose we can re-add it if we need it at a later time.

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -163,6 +165,88 @@ func New(ctx context.Context, clients *wrangler.Context) *Planner {
...
+	var waiting bool
+	for _, machineName := range machineNames {
+		machine := clusterPlan.Machines[machineName]
+		if machine == nil {
+			return fmt.Errorf("found unexpected machine %s that is not in cluster plan", machineName)
+		}
+
+		cond = rke2.Provisioned
+		if rke2.Provisioned.IsTrue(machine) {
+			cond = rke2.Updated
```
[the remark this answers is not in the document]
thedadams: After the machine has the `Provisioned` condition set to true, then we will be editing the `Updated` condition. The `Provisioned` condition will be set to true below (line 196).

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -72,21 +50,45 @@ func (h *handler) OnChange(_ string, secret *corev1.Secret) (*corev1.Secret, err
 		return secret, err
 	}
 
+	oldSecret := secret.DeepCopy()
+	defer func() {
+		if secret == nil || secret.Annotations == nil {
```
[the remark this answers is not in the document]
thedadams: The closure doesn't keep the state at the time it was created. So, if one of the functions below returns a `nil` secret, then it will be `nil` when this `defer` runs.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
```diff
@@ -72,21 +50,45 @@ func (h *handler) OnChange(_ string, secret *corev1.Secret) (*corev1.Secret, err
...
+			secret.Annotations[rke2.DrainErrorAnnotation] = err.Error()
+		} else {
+			delete(secret.Annotations, rke2.DrainErrorAnnotation)
+		}
+
+		var updateErr error
+		if secret, updateErr = h.secrets.Update(secret); updateErr != nil && err == nil {
+			err = updateErr
+		} else if updateErr != nil {
+			logrus.Errorf("Failed to update secret %s/%s with drain error annotation: %v", secret.Namespace, secret.Name, updateErr)
```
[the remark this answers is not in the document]
thedadams: Yes, I will take a stab at this.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
thedadams: The reason I had to add this retry is because the `system-agent` is also updating the plan secret every second while the upgrade/drain is happening. I am unsure if removing this will allow the updating of the secret to be successful.

pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -407,6 +407,7 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 					ObjectMeta: capi.ObjectMeta{
 						Labels: map[string]string{
 							capi.ClusterLabelName:           capiCluster.Name,
+							rke2.ClusterNameLabel:           capiCluster.Name,
```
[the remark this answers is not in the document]
thedadams: Yes, but that is already going to happen when a user upgrades to 2.6.4 because of some other changes we made. I am hoping we can get all these types of changes into 2.6.4 so it doesn't have to happen again.

pkg/provisioningv2/rke2/planner/drain.go
thedadams: Yes, I will add a comment.

pkg/provisioningv2/rke2/planner/planner.go
thedadams: Yes I can do that.

pkg/provisioningv2/rke2/planner/planner.go
thedadams: You're right, but I was just trying to simplify the signature of `reconcile`. I can switch it back if you'd like.

pkg/provisioningv2/rke2/planner/planner.go
thedadams: The method `generateSecrets` on line 281 could return a `nil` `controlPlane`.

I can switch this back and ensure that method doesn't return a `nil` `controlPlane`, but this seemed safer to me so that if any other situation might get us into a `nil` `controlPlane` (now or in the future), then this protects us.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
Oats87: I'm not saying we should remove the retry, I'm saying we should validate the before settings of the annotations being what we expect before we manipulate the latest secret we retrieve.

thedadams: Ah. I misunderstood. Yes, I will add this.

pkg/provisioningv2/rke2/planner/drain.go
Oats87: I understand the functionality here. The name of the function is vague.

pkg/provisioningv2/rke2/planner/planner.go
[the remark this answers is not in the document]
Oats87: I'm not so sure this behavior makes sense to me where we start manipulating different conditions based on other conditions -- this is confusing for both debug and development purposes IMO.

Were we doing this before?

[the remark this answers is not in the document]
thedadams: Are you suggesting that we only modify the `Provisioned` condition and leave out the `Updated` condition altogether?

pkg/provisioningv2/rke2/planner/planner.go
Oats87: No, that makes sense, the change here makes sense. I've been spending too much time between posix vs. golang recently.

pkg/provisioningv2/rke2/planner/planner.go
Oats87: I would either prefer we use a const for the `tierName` or switch it back to where it was before

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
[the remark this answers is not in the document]
Oats87: Makes sense

pkg/provisioningv2/rke2/planner/planner.go
thedadams: I will switch it back.

pkg/provisioningv2/rke2/planner/planner.go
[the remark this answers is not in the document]
thedadams: I added a commit to only use the `Provisioned` condition. Let me know what you think.

jakefhyde (dismissed):

pkg/provisioningv2/rke2/planner/drain.go
thedadams: Fixed

paynejacob (approved):

pkg/provisioningv2/rke2/planner/planner.go
jiaqiluo: Hi @thedadams , if a nil controlPlane is returned on line 281, will the defer function panic because it tries to access the namespace, name, and UID of the nil controlPlane?

jiaqiluo (changes requested):

pkg/provisioningv2/rke2/planner/planner.go
thedadams: No, that is why I passed in the things we need from the control plane as arguments. By passing them as arguments to the deferred function, they are frozen in time.

When we were not passing them as arguments, then what you are saying is possible: line 281 returns a nil control plane, then the deferred function tries to use the control plane and panics.

jiaqiluo: Got it!
Thanks for the explanation.

thedadams:
I had to force push to resolve conflicts.

pkg/controllers/provisioningv2/rke2/machinedrain/machinedrain.go
Oats87: Sorry if I was unclear, this is actually not exactly what I meant. 

In this specific controller, we are looking for and manipulate a couple of annotations, i.e. `rke2.DrainDoneAnnotation`, `rke2.PreDrainAnnotation`, `rke2.PostDrainAnnotation`, `rke2.DrainErrorAnnotation`, `rke2.UncordonAnnotation`, etc. 

What I worry about is that this controller is taking a specific version of the secret, it is then processing the "input" annotations i.e. `rke2.DrainAnnotation` and then setting a bunch of other annotations based on that input. 

The issue with the `retry` here is that, let's say we have a situation where the `rke2.DrainAnnotation` is set, then immediately updated, the "input" secret is no longer the same as what we are currently processing and the output i.e. result of the event is going to have whatever specific annotations we're trying to set on the secret. Because the secret will have been updated while the secret is being processed through this controller, we will get a conflict and will be unable to set the annotation, but will continuously retry to set that annotation reflecting the "old" condition even though that has been updated at this point.

What I am thinking is before allowing the retry to continue, we should validate that our input, i.e. `rke2.DrainAnnotation`, `hook.Annotaiton`, etc. are all what we expect them to be every time we retrieve the secret from the client before setting our "result" annotations. This is to help prevent unnecessary thrashing of the controller, and keeps us closer to the conventional model of "re-enqueue if conflict"

Oats87 (commented):
Other than the drain annotation set functionality, this all looks good to me.

paynejacob (approved):

jakefhyde (approved):

Oats87 (approved):

jiaqiluo (approved):
LGTM 💯

Issue #36505, opened 2022-02-13
Issue with RKE2 upgrade with Upgrade strategy set


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6head commit id: `11a7451`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: `v1.21.9+rke2r1`
- Cluster Type (Local/Downstream): Linode node driver cluster - 1 etcd, 1 cp and 3 worker nodes. Container network - `[multus-canal]`

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Standard user
   
**Describe the bug**
Issue with RKE2 upgrade with Upgrade strategy set

**To Reproduce**

- Deploy an RKE2 cluster.
- When the cluster is Active, Edit Config --> Modify the Upgrade strategy to

<img width="1180" alt="Screen Shot 2022-02-13 at 2 49 34 PM" src="https://user-images.githubusercontent.com/26032343/153778785-51f8f5a2-2911-4dab-9eb9-6e12d7c9e041.png">

- Save changes made.
- Edit cluster config again. Change the k8s version to `v1.22.6+rke2r1` and save changes made.
- The cluster goes into Provisioning state. 
- In Cluster explorer --> Nodes page to the cluster details page on Cluster Management, the nodes state do not show `Upgrading` like how it did in Ember UI for RKE1 clusters. Logged a separate bug for this - https://github.com/rancher/dashboard/issues/5110
- Notice the following in Cluster explorer --> Nodes page:
- Etcd node's k8s changes to 1.22 first
- The worker nodes get cordoned state. 
- Then the Control plane node's state changes and the k8s version upgrades to 1.22.6+rke2r1
- The worker node's k8s version changes
- Rancher server logs:
```
2022/02/13 22:21:29 [INFO] rkecluster fleet-default/sowmya-linode-multus-canal: provisioning bootstrap node(s) sowmya-linode-multus-canal-pool1-7b59f5c454-hkg9z: waiting for probes: etcd
....
2022/02/13 22:22:02 [INFO] rkecluster fleet-default/sowmya-linode-multus-canal: provisioning control plane node(s) sowmya-linode-multus-canal-pool2-78869f96f5-9mn84: waiting for probes: kube-apiserver
[...]

PR #36726, opened 2022-03-03
Wait for version upgrades before proceeding with more upgrades

thedadams:
Issue: https://github.com/rancher/rancher/issues/36505

## Problem
Currently, when upgrading an RKE2 cluster, the planner will go onto the next
machine type before the upgrade finishes. For example, the planner will move
onto upgrade worker nodes before the control plane nodes are completely
upgraded. That is because the plan was considered in-sync after the system-agent
applied it and not after the upgrade completed.

## Solution
Now, a plan will remain in-sync at the point the system-agent has applied it,
but the machine will not be considered to be in-sync until the version of the
node has been updated.

Because of this change, we must include the CAPI machines as "related resources"
so that the planner runs when the node version has been updated. This should
speed up provisioning slightly because the planner will kick off as soon as a
node has the ready state instead of waiting for other things to enqueue the
planner.

## Testing
1. Provision an RKE2 cluster from Rancher with multiple machine pools with different roles. It is easiest if the upgrade strategy is set to drain worker nodes so you can see when that starts.
2. After the cluster is active, upgrade the Kubernetes version of the cluster.
3. Watch the nodes to ensure that nothing happens to the worker nodes until the control plane nodes are upgraded completely.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/planner/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/planner/controller.go
@@ -15,6 +15,7 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
 type handler struct {
@@ -37,9 +38,17 @@ func Register(ctx context.Context, clients *wrangler.Context, planner *planner.P
 					Name:      clusterName,
 				}}, nil
 			}
+		} else if machine, ok := obj.(*capi.Machine); ok {
+			clusterName := machine.Labels[rke2.ClusterNameLabel]
+			if clusterName != "" {
+				return []relatedresource.Key{{
+					Namespace: machine.Namespace,
+					Name:      clusterName,
+				}}, nil
+			}
 		}
 		return nil, nil
-	}, clients.RKE.RKEControlPlane(), clients.Core.Secret())
+	}, clients.RKE.RKEControlPlane(), clients.Core.Secret(), clients.CAPI.Machine())
 }
 
 func (h *handler) OnChange(cluster *rkev1.RKEControlPlane, status rkev1.RKEControlPlaneStatus) (rkev1.RKEControlPlaneStatus, error) {
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -444,7 +444,7 @@ func (p *Planner) reconcile(controlPlane *rkev1.RKEControlPlane, tokensSecret pl
 					draining = append(draining, entry.Machine.Name)
 				}
 			}
-		} else if !entry.Plan.InSync {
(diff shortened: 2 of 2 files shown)


pkg/controllers/provisioningv2/rke2/planner/controller.go
```diff
@@ -37,6 +38,14 @@ func Register(ctx context.Context, clients *wrangler.Context, planner *planner.P
...
+			clusterName := machine.Labels[rke2.ClusterNameLabel]
+			if clusterName != "" {
+				return []relatedresource.Key{{
+					Namespace: machine.Namespace,
+					Name:      clusterName,
+				}}, nil
+			}
 		}
 		return nil, nil
 	}, clients.RKE.RKEControlPlane(), clients.Core.Secret())
```
Oats87, line 51: I believe we need to pass in a `clients.CAPI.Machine()` `ControllerWrapper` into this function in order for the above logic to work

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -444,7 +444,7 @@ func (p *Planner) reconcile(controlPlane *rkev1.RKEControlPlane, tokensSecret pl
 					draining = append(draining, entry.Machine.Name)
 				}
 			}
-		} else if !entry.Plan.InSync {
+		} else if !entry.Plan.InSync || entry.Machine.Status.NodeInfo == nil || entry.Machine.Status.NodeInfo.KubeletVersion != controlPlane.Spec.KubernetesVersion {
```
Oats87, line 447:
```suggestion
		} else if !entry.Plan.InSync || (entry.Machine.Status.NodeInfo != nil && entry.Machine.Status.NodeInfo.KubeletVersion != controlPlane.Spec.KubernetesVersion) {
```

Oats87: I'm thinking this may slow down initial provisioning unnecessarily, but I could be wrong

Oats87 (changes requested):

pkg/provisioningv2/rke2/planner/planner.go
thedadams: I don't think it will because we won't move forward until the CAPI Machine object reports that the downstream node is ready. That won't happen until this field is populated.

In other words, the machine will be in the `outOfSync` slice instead of the `nonReady` slice in the planner.

Oats87: While that is true for the bootstrap node (because we scrape the joinURL off of the NodeInfo for all except for an etcd-only init node), I'm not sure that is actually the case for the rest of the nodes.

The `nonReady` slice gets rendered and returned into an `errIgnore` rather than an `ErrWaiting`. Looking at the planner, we will move on if there is an `errIgnore`

thedadams: OK. Will update.

Oats87 (approved):

jakefhyde (approved):

paynejacob (approved):

Issue #36690, opened 2022-02-28
Unable to edit imported AKS clusters


David-VTUK:
**Setup**
- Rancher version:  v2.6.3 
- Browser type & version: Firefox 97.0.1 (64-bit)

**Describe the bug**
Unable to edit the options of an imported AKS cluster 

**To Reproduce**

1. Create an AKS cluster from the Azure console.
2. Add credentials to Rancher
3. Navigate to `Cluster Management -> Import Existing -> Azure AKS -> Import cluster`
4. Once the cluster has been imported, Navigate to `Cluster Management -> Edit Config -> Select credentials and Configure Cluster`. 

**Result**

An empty warning bar is displayed but no cluster details are shown:

![image](https://user-images.githubusercontent.com/5892615/156033555-da9ab417-5770-44a7-86e8-3286832ed657.png)

Browser tools indicate this API call returned a `502 Bad Gateway` message:

GET https://RANCHERURL/meta/aksUpgrades?cloudCredentialId=cattle-global-data:cc-REF&region=REGION&clusterId=CLUSTERID

**Expected Result**
To edit existing, imported AKS clusters

**Additional context**
* AKS Clusters *created* by Rancher using the same credentials can have their details edited
* AKS version  `v1.21.9`

FerencoVonMatterhorn:
Our Team is also experiencing this issue on Version 2.6.3 😶

annablender:
## Testing template

#### Root cause

Rancher does not manage fields for imported clusters that the user does not set in the cluster config or launch template. So, when the user imported an AKS cluster Rancher didn't care about the kubernetes version set on the upstream AKS config. That caused a panic when `kubernetesVersion` was accessed.
 
#### What was fixed, or what changes have occurred

I added checks to make sure if `kubernetesVersion` isn't set, use the current version set in `aksStatus` https://github.com/rancher/rancher/blob/428ec692b0336a3bc54b5c68e29c4acd6ee9857b/pkg/api/norman/customization/aks/listers.go#L161.

#### Areas or cases that should be tested

* Create an AKS cluster in Azure portal
* Import the AKS cluster using the Rancher UI
* Go to Cluster Management -> Edit Config for your imported cluster and check that you can edit the configuration and then save it, and the cluster updates
 
#### What areas could experience regressions ?
 
N/A

#### Are the repro steps accurate/minimal ?

Yes.

thedadams:
Reproduced the issue on Rancher v2.6.3:
[...]

PR #36735, opened 2022-03-03
Add kubernetes version check to aksUpgrades endpoint

annablender:
Issue: https://github.com/rancher/rancher/issues/36690 

 
## Problem
 

 
We never check whether `kubernetesVersion` is set in the `aksConfig` and if it's nil, the `aksUpgrades` endpoint was failing to return a list of valid Kubernetes upgrade versions.

## Solution
 

I added a check to make sure if `kubernetesVersion` isn't set, use the current version set in `aksStatus`.
 
## Testing
 

* Create an AKS cluster in Azure portal
* Import the AKS cluster using the Rancher UI
* Go to Cluster Management -> Edit Config for your imported cluster and check that you can edit the configuration and then save it, and the cluster updates

### DIFF
--- a/pkg/api/norman/customization/aks/listers.go
+++ b/pkg/api/norman/customization/aks/listers.go
@@ -158,7 +158,14 @@ func listKubernetesUpgradeVersions(ctx context.Context, clusterLister mgmtv3.Clu
 		return nil, http.StatusBadRequest, fmt.Errorf("invalid cluster id")
 	}
 
-	resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	if cluster.Spec.AKSConfig.KubernetesVersion != nil {
+		resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	} else {
+		if cluster.Status.AKSStatus.UpstreamSpec == nil || cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion == nil {
+			return nil, http.StatusBadRequest, fmt.Errorf("kubernetes version of the cluster cannot be determined")
+		}
+		resp.CurrentVersion = *cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion
+	}
 
 	// get the client for aks container service
 	clientContainer, err := NewContainerServiceClient(cap)


paynejacob (approved):

pkg/api/norman/customization/aks/listers.go
```diff
@@ -158,7 +158,11 @@ func listKubernetesUpgradeVersions(ctx context.Context, clusterLister mgmtv3.Clu
 		return nil, http.StatusBadRequest, fmt.Errorf("invalid cluster id")
 	}
 
-	resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	if cluster.Spec.AKSConfig.KubernetesVersion != nil {
+		resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	} else {
+		resp.CurrentVersion = *cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion
```
jiaqiluo, line 164: Is it possible for `*cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion` to be nil ? 
If it is not, the fix looks good to me.

jiaqiluo (changes requested):

pkg/api/norman/customization/aks/listers.go
annablender: @thedadams I don't think so right? Will the kubernetes version in the status always be set?

jiaqiluo (dismissed):

jiaqiluo (approved):

pkg/api/norman/customization/aks/listers.go
```diff
@@ -158,7 +158,17 @@ func listKubernetesUpgradeVersions(ctx context.Context, clusterLister mgmtv3.Clu
...
-	resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	if cluster.Spec.AKSConfig.KubernetesVersion != nil {
+		resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	} else {
+		if cluster.Status.AKSStatus.UpstreamSpec == nil {
+			return nil, http.StatusBadRequest, fmt.Errorf("no upstream spec found")
+		}
+		if cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion == nil {
+			return nil, http.StatusBadRequest, fmt.Errorf("no kubernetes version found in upstream spec")
+		}
```
thedadams, line 169: The user doesn't know that the upstreamSpec exists, so telling them that it (or something on it) doesn't exist won't make sense to them.

I think something like "kubernetes version of the cluster cannot be determined" in both cases is better.

thedadams (changes requested):

jiaqiluo (dismissed):

jakefhyde (approved):

pkg/api/norman/customization/aks/listers.go
```diff
@@ -158,7 +158,17 @@ func listKubernetesUpgradeVersions(ctx context.Context, clusterLister mgmtv3.Clu
...
-	resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	if cluster.Spec.AKSConfig.KubernetesVersion != nil {
+		resp.CurrentVersion = *cluster.Spec.AKSConfig.KubernetesVersion
+	} else {
+		if cluster.Status.AKSStatus.UpstreamSpec == nil {
+			return nil, http.StatusBadRequest, fmt.Errorf("kubernetes version of the cluster cannot be determined")
+		}
+		if cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion == nil {
+			return nil, http.StatusBadRequest, fmt.Errorf("kubernetes version of the cluster cannot be determined")
+		}
```
thedadams, line 169: Can we combine these?
```suggestion
		if cluster.Status.AKSStatus.UpstreamSpec == nil ||
		cluster.Status.AKSStatus.UpstreamSpec.KubernetesVersion == nil{
			return nil, http.StatusBadRequest, fmt.Errorf("kubernetes version of the cluster cannot be determined")
		}
```

thedadams (changes requested):

jakefhyde (approved):

thedadams (approved):

jiaqiluo (approved):

PR #36739, opened 2022-03-03
made certificate rotation idempotent

paynejacob:
#35343 

# Problem
If the planner runs the rotation plan more than once per generation  a server's certificate will be rotated multiple times

# Solution
track the generation on each node and only rotate certificates if the generation is different.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/certificaterotation.go
+++ b/pkg/provisioningv2/rke2/planner/certificaterotation.go
@@ -1,7 +1,9 @@
 package planner
 
 import (
+	"encoding/base64"
 	"fmt"
+	"strconv"
 
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1/plan"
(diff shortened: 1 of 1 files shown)


Oats87 (changes requested):

pkg/provisioningv2/rke2/planner/certificaterotation.go
```diff
@@ -86,11 +97,32 @@ func rotateCertificatesPlan(controlPlane *rkev1.RKEControlPlane, rotation *rkev1
...
+	if len(rotation.Services) > 0 {
+		for _, service := range rotation.Services {
+			args = append(args, "-s", service)
+		}
+	}
+
 	return plan.NodePlan{
+		Files: []plan.File{
+			{
+				Content: base64.StdEncoding.EncodeToString([]byte(idempotentRotateScript)),
```
annablender, line 118: Question - this script is meant to be run programmatically right? Or manually? If it's the former(which I think it is), where do you pass the current and target gen in as args?

annablender (commented):
Otherwise, LGTM syntax/logic wise though I'm not too familiar with cert stuff

pkg/provisioningv2/rke2/planner/certificaterotation.go
jakefhyde: The file is copied, but it is executed through the planner as part of the `OneTimeInstruction.Args`, so it's easy to miss: https://github.com/rancher/rancher/blob/b1b5da4362e5a4057c085cdae0ec644d929d3cfd/pkg/provisioningv2/rke2/planner/certificaterotation.go#L104

annablender: I see, thanks!

jakefhyde (approved):

annablender (dismissed):

Oats87 (approved):

jakefhyde (approved):

PR #36741, opened 2022-03-03
Pin itsdangerous integration test dependency

cbron:
itsdangerous 2.1.0 removed its json module, so pin it to the last known
working version 2.0.1.


rmweir (approved):
LGTM

Issue #36732, opened 2022-03-02
v2prov RKE2 clusters cannot be provisioned when using an egress proxy and setting Agent Environment Variables in the Rancher UI


rosskirkpat:
**Setup**
- Rancher version: v2.6-e93a53c35bd038d4cf0d0d96897e957ffef150c9-head
-- | --
Helm | v2.16.8-rancher1
Machine | v0.15.0-rancher79

- Browser type & version: Firefox 97.0.1 (64-bit)

**Describe the bug**
The environment variables HTTP_PROXY, HTTPS_PROXY, NO_PROXY are not being set in the custom cluster registration commands for either Linux or Windows

**To Reproduce**
- provision custom rke2 cluster through Rancher UI with proxy vars set in the Agent Environment Variables tab

**Result**
The registration commands do not have the environment variables HTTP_PROXY, HTTPS_PROXY, NO_PROXY set for the retrieval of the installation scripts when behind an egress proxy. This causes the installation scripts to be unable to be retrieved from Rancher when an egress proxy is configured/in place between downstream nodes and Rancher itself.

**Expected Result**
The registration commands should be fully functional when nodes are behind an egress proxy with the proper Agent environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY) set in the Rancher UI

**Notes**
Registration commands should have the environment variables HTTP_PROXY, HTTPS_PROXY, NO_PROXY set to allow the installation scripts to be retrieved from Rancher when an egress proxy.

For Windows, they would need to be inserted as shown below:
`PowerShell $env:HTTP_PROXY="XXX"; $env:HTTPS_PROXY="YYY"; $env:NO_PROXY="ZZZ"; curl.exe --insecure -fL https://RANCHER_URL/wins-agent-install.ps1 -o install.ps1; ...`

For linux, the proxy variables need to be set before calling curl to enable the install script to be fetched from Rancher:
`HTTP_PROXY="XXX" HTTPS_PROXY="YYY" NO_PROXY="ZZZ" curl -fL https://RANCHER_URL/system-agent-install.sh | ... `

**Screenshots**
redacting snapshot
![image](https://user-images.githubusercontent.com/55704795/156456934-38f1d536-02f0-48e1-bcd2-4dc4e0228733.png)

phillipsj:
![proxy-working-for-linux](https://user-images.githubusercontent.com/1568133/157123040-7e4cae31-95b5-4fc1-b2cf-e67a5b281abd.png)
[...]

PR #36746, opened 2022-03-03
Improving envvar handling for custom cluster registration command.

phillipsj:
This adds the envvars before the cURL command since proxy ENVVARS needs to be consumed by the cURL command. In addition this adds the ability to handle Linux, Docker, and PowerShell environment variables.

* https://github.com/rancher/rancher/issues/36732

Signed-off-by: Jamie Phillips <jamie.phillips@suse.com>

### DIFF
--- a/pkg/controllers/dashboard/clusterregistrationtoken/agentvars.go
+++ b/pkg/controllers/dashboard/clusterregistrationtoken/agentvars.go
@@ -1,0 +1,37 @@
+package clusterregistrationtoken
+
+import (
+	"fmt"
+	"strings"
+
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+)
+
+type EnvType int
+
+const (
+	Linux EnvType = iota
+	PowerShell
+	Docker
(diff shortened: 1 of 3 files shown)


pkg/controllers/dashboard/clusterregistrationtoken/agentvars.go
```diff
@@ -0,0 +1,36 @@
...
+	Docker
+)
+
+func AgentEnvVars(cluster *v3.Cluster, envType EnvType) string {
+	var agentEnvVars []string
+	if cluster == nil {
+		return ""
+	}
+	for _, envVar := range cluster.Spec.AgentEnvVars {
+		if envVar.Value != "" {
```
luthermonson, line 24: nit: do an == and a continue instead of wrapping the switch

phillipsj:
This all checks out and is working correctly.
proxy-working-for-linux
proxy-working-for-windows

luthermonson (approved):

Oats87 (approved):

PR #36780, opened 2022-03-07
Use v2.6-head as default rancher-agent image

thedadams:
The master-head version of the rancher-agent is very old and does not support
Kubernetes v1.22+. This isn't a problem with "official" Rancher images because
the CATTLE_AGENT_IMAGE variable gets set to the proper values on build. However,
it does cause problems with local dev Rancher instances unless the
aforementioned environment variable is set.

This change uses the v2.6-head version of the agent to avoid problems in the
future.

### DIFF
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -23,7 +23,7 @@ var (
 	provider       Provider
 	InjectDefaults string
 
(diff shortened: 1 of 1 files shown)


Oats87 (approved):
Approving based on the fact that our default branch is now `release/v2.6`

phillipsj (approved):
LGTM!

PR #36793, opened 2022-03-08
add windows v2prov provisioning tests + vSphere extension

rosskirkpat:
- adds hybrid rke2 v2prov provisioning tests for node driver clusters
- adds hybrid rke2 v2prov provisioning tests for custom clusters
- add vsphere extension to the new framework
- add hybrid cluster support for the vsphere extension

### DIFF
--- a/tests/framework/extensions/machinepools/machinepools.go
+++ b/tests/framework/extensions/machinepools/machinepools.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"strconv"
 
+	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
+
 	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
 	corev1 "k8s.io/api/core/v1"
@@ -30,7 +32,7 @@ func CreateMachineConfig(resource string, machinePoolConfig *unstructured.Unstru
 }
 
 // NewRKEMachinePool is a constructor that sets up a apisV1.RKEMachinePool object to be used to provision a cluster.
-func NewRKEMachinePool(controlPlaneRole, etcdRole, workerRole bool, poolName string, quantity int32, machineConfig *unstructured.Unstructured) apisV1.RKEMachinePool {
+func NewRKEMachinePool(controlPlaneRole, etcdRole, workerRole bool, poolName string, machineOS string, quantity int32, machineConfig *unstructured.Unstructured) apisV1.RKEMachinePool {
 	machineConfigRef := &corev1.ObjectReference{
 		Kind: machineConfig.GetKind(),
 		Name: machineConfig.GetName(),
@@ -43,6 +45,7 @@ func NewRKEMachinePool(controlPlaneRole, etcdRole, workerRole bool, poolName str
 		NodeConfig:       machineConfigRef,
 		Name:             poolName,
 		Quantity:         &quantity,
+		MachineOS:        machineOS,
 	}
 }
 
@@ -61,10 +64,16 @@ func NewRKEMachinePool(controlPlaneRole, etcdRole, workerRole bool, poolName str
 // 	"worker":       false,
 // },
 // }
-func RKEMachinePoolSetup(nodeRoles []map[string]bool, machineConfig *unstructured.Unstructured) []apisV1.RKEMachinePool {
-	machinePools := []apisV1.RKEMachinePool{}
+func RKEMachinePoolSetup(nodeRoles []map[string]bool, hasWindows bool, machineConfig *unstructured.Unstructured) []apisV1.RKEMachinePool {
+	var machinePools []apisV1.RKEMachinePool
 	for index, roles := range nodeRoles {
-		machinePool := NewRKEMachinePool(roles["controlplane"], roles["etcd"], roles["worker"], "pool"+strconv.Itoa(index), 1, machineConfig)
+		machinePool := NewRKEMachinePool(roles["controlplane"], roles["etcd"], roles["worker"], "pool"+strconv.Itoa(index), rke2.DefaultMachineOS, 1, machineConfig)
+		if hasWindows {
+			machinePool2019 := NewRKEMachinePool(false, false, true, "windows-pool-2019"+strconv.Itoa(index), rke2.WindowsMachineOS, 1, machineConfig)
+			machinePool2022 := NewRKEMachinePool(false, false, true, "windows-pool-2022"+strconv.Itoa(index), rke2.WindowsMachineOS, 1, machineConfig)
+			machinePools = append(machinePools, machinePool2019, machinePool2022)
+
+		}
 		machinePools = append(machinePools, machinePool)
 	}
 
--- a/tests/framework/extensions/machinepools/vsphere_machine_config.go
+++ b/tests/framework/extensions/machinepools/vsphere_machine_config.go
@@ -1,0 +1,84 @@
+package machinepools
+
+import (
+	"github.com/rancher/rancher/tests/framework/pkg/config"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	VSphereKind                                    = "VmwarevsphereConfig"
+	VSpherePoolType                                = "rke-machine-config.cattle.io.vmwarevsphereconfig"
+	VSphereResourceConfig                          = "vmwarevsphereconfigs"
+	VmwareVsphereMachingConfigConfigurationFileKey = "vmwarevsphereMachineConfig"
+)
+
+type VSpherePublicCredentials struct {
+	Username    string `json:"username" yaml:"username"`
+	VCenter     string `json:"vcenter" yaml:"vcenter"`
+	VCenterPort string `json:"vcenterPort" yaml:"vcenterPort"`
+}
+
+type VSpherePrivateCredentials struct {
+	// TODO: use sha256sum
+	Password string `json:"password" yaml:"password"`
+}
+
+type VSphereMachineConfig struct {
+	DiskSize         string   `json:"diskSize" yaml:"diskSize"`
+	CPUCount         string   `json:"cpuCount" yaml:"cpuCount"`
+	MemorySize       string   `json:"memorySize" yaml:"memorySize"`
+	NetworkName      string   `json:"networkName" yaml:"networkName"`
+	CreationType     string   `json:"creationType" yaml:"creationType"`
+	SSHUser          string   `json:"sshUser" yaml:"sshUser"`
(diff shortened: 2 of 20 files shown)


tests/framework/extensions/machinepools/machinepools.go
```diff
@@ -61,10 +64,13 @@ func NewRKEMachinePool(controlPlaneRole, etcdRole, workerRole bool, poolName str
 // 	"worker":       false,
 // },
 // }
-func RKEMachinePoolSetup(nodeRoles []map[string]bool, machineConfig *unstructured.Unstructured) []apisV1.RKEMachinePool {
-	machinePools := []apisV1.RKEMachinePool{}
+func RKEMachinePoolSetup(nodeRoles []map[string]bool, hasWindows bool, machineConfig *unstructured.Unstructured) []apisV1.RKEMachinePool {
+	var machinePools []apisV1.RKEMachinePool
 	for index, roles := range nodeRoles {
-		machinePool := NewRKEMachinePool(roles["controlplane"], roles["etcd"], roles["worker"], "pool"+strconv.Itoa(index), 1, machineConfig)
+		machinePool := NewRKEMachinePool(roles["controlplane"], roles["etcd"], roles["linuxWorker"], "pool"+strconv.Itoa(index), rke2.DefaultMachineOS, 1, machineConfig)
```
igomez06, line 70: When if we're change `worker` to `linuxWorker` we need to update this in our existing tests. Is it super necessary to rename this?

tests/v2/validation/provisioning/provisioning_node_driver_linux_test.go
```diff
@@ -0,0 +1,203 @@
...
+	require.NoError(r.T(), err)
+
+	cloudCredential, err := provider.CloudCredFunc(client)
+	require.NoError(r.T(), err)
+
+	allNodeRolesLinux := []map[string]bool{
+		{
+			"controlplane": true,
+			"etcd":         true,
+			"worker":       true,
```
igomez06, line 38: This is what I'm referring to with the linuxWorker key over in `tests/framework/extensions/machinepools/machinepools.go`

igomez06 (commented):
So far so good, I like the direction we're going, especially with mixed os test design.

rosskirkpat:
> Also since `hasWindows` is the only thing different, maybe more stuff I'm not sure, but why have the distinction of `ProvisioningRKE2ClusterMixedOS` vs `ProvisioningRKE2ClusterLinuxOnly` especially since the bool is coming from the config file? This goes for the different setups as well. Just curious about the decision

In the context of provisioning v2 and node drivers, we only have Rancher support for RKE2 Windows nodes on vSphere. k3s windows nodes are not supported across any cluster type. We have backlogged tasks to add Windows support for rke2 to multiple cloud providers as well as the harvester node driver.

My design vision was that by breaking the tests out separately now, we are reducing the complexity of maintaining and adding windows tests to each extension/cluster type as it gains windows support.

tests/framework/extensions/machinepools/machinepools.go
rosskirkpat: that is a typo, I was testing out a different method and realized it would not work without a massive rewrite that ultimately made no sense. I will get that fixed.

tests/v2/validation/provisioning/provisioning_node_driver_linux_test.go
rosskirkpat: linuxWorker was a remnant of a prior approach. I have switched it back to worker

rosskirkpat:
Due to the architecture of the framework, I won't be able to add userdata that runs the cluster registration command to replace the ssh step after nodes are provisioned. 

@igomez06 @bmdepesa @slickwarren This PR is now ready for review.

tests/v2/validation/provisioning/provisioning_node_driver_common_test.go
```diff
@@ -0,0 +1,94 @@
...
+	user := &management.User{
+		Username: testuser,
+		Password: "rancherrancher123!",
+		Name:     testuser,
+		Enabled:  &enabled,
+	}
+
+	newUser, err := users.CreateUserWithRole(client, user, "user")
+	require.NoError(r.T(), err)
+
```
igomez06, line 55: Needs `newUser.Password = user.Password`

tests/v2/validation/provisioning/provisioning_node_driver_common_test.go
```diff
@@ -0,0 +1,94 @@
...
```
igomez06, line 89: Needs `newUser.Password = user.Password`. And also what's the difference between this setup and the linux only setup?

tests/v2/validation/provisioning/provisioning_node_driver_hybrid_test.go
```diff
@@ -99,26 +60,19 @@ func (r *RKE2NodeDriverProvisioningTestSuite) ProvisioningRKE2Cluster(provider P
 	}
 
 	tests := []struct {
-		name      string
-		nodeRoles []map[string]bool
-		client    *rancher.Client
+		name       string
+		nodeRoles  []map[string]bool
+		hasWindows bool
```
igomez06, line 65: So if hasWindows an argument here, why is there a need to have two setups for either hybrid or linux only?

tests/framework/extensions/nodes/ec2/ec2.go
```diff
@@ -149,3 +179,203 @@ func getSSHKeyName(sshKeyName string) string {
...
+	if err != nil {
+		return "", err
+	}
+	localPEM := filepath.Join(user.HomeDir, sshPath, aws.StringValue(input.KeyName), ".pem")
+	err = os.WriteFile(localPEM, []byte(convert.ToString(sensitivePEM)), 0400)
+	if err != nil {
+		os.Remove(localPEM)
+		return "", err
+	}
+
```
igomez06, line 209: We should cleanup the PEMKey creation, similar to how we delete the nodes using the session cleanup.

tests/v2/validation/provisioning/config.go
```diff
@@ -19,20 +18,21 @@ type Config struct {
...
 	Providers          []string `json:"providers" yaml:"providers"`
 	NodeProviders      []string `json:"nodeProviders" yaml:"nodeProviders"`
+	ClusterHasWindows  bool     `json:"clusterHasWindows" yaml:"clusterHasWindows"`
 }
 
-func AppendRandomString(baseClusterName string) string {
-	clusterName := "auto-" + baseClusterName + "-" + namegenerator.RandStringLower(defaultRandStringLength)
-	return clusterName
-}
+//func AppendRandomString(baseClusterName string) string {
```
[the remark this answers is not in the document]
rosskirkpat: I had moved it as part of trying to shoehorn userdata in, reverting this change.

tests/framework/extensions/machinepools/vsphere_machine_config.go
```diff
@@ -0,0 +1,87 @@
...
+	VCenter     string `json:"vcenter" yaml:"vcenter"`
+	VCenterPort string `json:"vcenterPort" yaml:"vcenterPort"`
+}
+
+type VSpherePrivateCredentials struct {
+	// TODO: use sha256sum
+	Password string `json:"password" yaml:"password"`
+}
+
+// TODO: use dynamic client to create cloud credential secret for vsphere
```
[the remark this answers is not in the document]
rosskirkpat: It does not, removing.

tests/framework/pkg/nodes/nodes.go
```diff
@@ -19,6 +19,7 @@ type Node struct {
 	SSHUser         string `json:"sshUser" yaml:"sshUser"`
 	SSHKeyName      string `json:"sshKeyName" yaml:"sshKeyName"`
 	SSHKey          []byte
+	NodePassword    string `json:"nodePassword,omitempty" yaml:"nodePassword,omitempty"`
```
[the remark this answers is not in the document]
rosskirkpat: Windows nodes require runtime configuration to enable ssh and also add a public key to allow ssh without a password. I have not found the userdata to be 100% reliable which is why having the node password available as a fallback is a worthwhile idea (to me).

tests/framework/extensions/nodes/ec2/ec2.go
rosskirkpat: I added a call to the cleanupPEM function in this err block

tests/framework/extensions/nodes/ec2/ec2.go
```diff
@@ -1,83 +1,96 @@
...
-		},
+	var listOfInstanceIds []*string
+
+	// Create Linux Nodes
+	runInstancesInput, err := createNodesCommon(client, numOfInstances, false, "")
+	if err != nil {
+		return nil, err
 	}
 
 	reservation, err := ec2Client.SVC.RunInstances(runInstancesInput)
```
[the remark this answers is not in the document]
rosskirkpat: Windows clusters require all three roles to be available on a linux node(s)

tests/framework/pkg/nodes/nodes.go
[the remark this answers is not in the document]
rosskirkpat: Actually, upon second review, I have removed the node passwords

tests/framework/extensions/nodes/ec2/ec2.go
```diff
@@ -149,3 +173,118 @@ func getSSHKeyName(sshKeyName string) string {
...
+		var (
+			windowsOwner            []string
+			windows2019FilterValues []string
+			windows2022FilterValues []string
+		)
+		windowsOwner = append(windowsOwner, "amazon")
+		windowsFilter := &ec2.Filter{}
+
+		switch windowsVersion {
+		case "2019":
```
igomez06, line 203: Would it make sense to have "2019" and "2022" as consts rather than having them hardcoded in the switch case?

igomez06 (commented):
Overall it looks good Ross, I just want two things. I want clarification on if the setup function would run as they are not the standard SetupSuite() style of function name. More of a curiosity. And the other if we could get a Jenkins run and see it pass. I think it would be good to go!

tests/v2/validation/provisioning/custom_cluster_hybrid_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var ClusterID string
+
+func (c *CustomClusterProvisioningTestSuite) setupHybrid() {
+	c.SetupSuiteHybrid()
+}
```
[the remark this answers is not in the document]
igomez06: Curious as well since this setupHyrbid is a private function. Just need clarification thank you.

igomez06:
Also Ross you should rebase, as the provisioning validation tests have slightly changed structure

PR #36805, opened 2022-03-08
adding an isWindows check to rke2 planner to block calico probe checks

luthermonson:
Calico on windows does not boot up the liveness check on 9099 which the planner probes rely on. We need to block this probe for windows only.


pkg/provisioningv2/rke2/planner/planner_test.go
```diff
@@ -128,3 +128,14 @@ func findEnv(s []string, v string) bool {
 	}
 	return false
 }
+
```
phillipsj, line 131: Can you make multiple test cases like here, to add both a true and false? https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/dashboard/clusterregistrationtoken/agentvars_test.go#L11

phillipsj (approved):
LGTM!

Oats87 (approved):

PR #36808, opened 2022-03-08
Implemented fix to address the timing issue with namespace creation

igomez06:
**Background**
When creating a namespace as a standard user project owner/member there were instances when the cleanup would occur before the cluster role for that namespace had been created therefore logging an RBAC error. There would be cases where the cleanup would not be successful and cause failures in the `project_user_test.go` tests. 

**Framework Updates**
To fix it, I put in a watch to make sure that specific cluster role for that namespace had been created before registering the delete function. Also I used the admin client to watch if the namespace had been deleted, as this was causing permissions issues since the cluster role had been deleted. One the thing remains since the namespace delete function (cleanup) is registered twice, the second delete logs this error because of the cluster role no longer existing. `namespaces "<namespace name>" is forbidden: User "<user>" cannot delete resource "namespaces" in API group "" in the namespace "<namespace name>": RBAC: clusterrole.rbac.authorization.k8s.io "<cluster role name>" not found` I am of the opinion to just live with this as it's not causing failures. We can also have the dynamic client keep track of the admin token and then use the admin client to see if that namespace had been deleted to avoid logging the error. I don't like that idea, but I can be persuade to do something to that effect.

### DIFF
--- a/tests/framework/extensions/namespaces/namespaces.go
+++ b/tests/framework/extensions/namespaces/namespaces.go
@@ -3,6 +3,7 @@ package namespaces
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/rancher/rancher/pkg/api/scheme"
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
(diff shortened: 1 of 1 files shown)


tests/framework/extensions/namespaces/namespaces.go
```diff
@@ -51,13 +55,59 @@ func CreateNamespace(client *rancher.Client, namespaceName, containerDefaultReso
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	err = wait.WatchWait(watchTemp, func(event watch.Event) (ready bool, err error) {
+		temp := event.Object
+		aasdf := temp.(*kubeUnstructured.Unstructured)
+		clusterRole := &rbacv1.ClusterRole{}
+		err = scheme.Scheme.Convert(aasdf, clusterRole, aasdf.GroupVersionKind())
```
paynejacob, line 91:
```suggestion
		err = scheme.Scheme.Convert(event.object.(*kubeUnstructured.Unstructured), clusterRole, aasdf.GroupVersionKind())
```

paynejacob (changes requested):

paynejacob (approved):

paynejacob (approved):

PR #36828, opened 2022-03-09
[Automation] - use gp3 as default volume type for ec2 instances

izaac:
Just moving to `gp3` volume type by default in our automation for EC2 instances.

Reference: https://aws.amazon.com/ebs/general-purpose/


bmdepesa (approved):
LGTM

sowmyav27 (approved):

Issue #36825, opened 2022-03-09
RKE1 | Can't provision a 1.20.16 k8s cluster


sgapanovich:
**Rancher Server Setup**
- Rancher version: v2.6-head 47c2238
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1

**Information about the Cluster**
- Kubernetes version: 1.20.16
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom / Node driver
   
**Describe the bug**

Can't provision a RKE1 cluster with v1.20.16 k8s version

**To Reproduce**
 
1. Provision a rke1 node driver or custom cluster with v1.20.16-rancher1-1 k8s version'

**Result**
Cluster is not provisioned

Rancher logs:
```
...
2022/03/09 16:10:33 [WARNING] Can't pull Docker image [rancher/hyperkube:v1.20.16-rancher1] on host [REDACTED]: Error response from daemon: manifest for rancher/hyperkube:v1.20.16-rancher1 not found: manifest unknown: manifest unknown
2022/03/09 16:10:33 [WARNING] [pre-deploy] Can't pull Docker image [rancher/hyperkube:v1.20.16-rancher1] on host [REDACTED]: Error response from daemon: manifest for rancher/hyperkube:v1.20.16-rancher1 not found: manifest unknown: manifest unknown
2022/03/09 16:10:33 [ERROR] cluster [c-8q7w4] provisioning: [Error response from daemon: manifest for rancher/hyperkube:v1.20.16-rancher1 not found: manifest unknown: manifest unknown]
2022/03/09 16:10:33 [INFO] kontainerdriver rancherkubernetesengine stopped
2022/03/09 16:10:33 [ERROR] error syncing 'c-8q7w4': handler cluster-provisioner-controller: [Error response from daemon: manifest for rancher/hyperkube:v1.20.16-rancher1 not found: manifest unknown: manifest unknown], requeuing
```

**Expected Result**

Cluster is provisioned with 1.20.16 k8s version

**Screenshots**

![2016](https://user-images.githubusercontent.com/40282141/157483003-15327ed5-96c7-4d8b-b4ee-c39198d41e4d.jpg)

kinarashah:
[...]

PR #36832, opened 2022-03-09
bump rke to v1.3.8-rc6

kinarashah:
https://github.com/rancher/rancher/issues/36825


rancherbot (approved):

Issue #36631, opened 2022-02-24
Drain on delete - Node is stuck in deleting when tried to scale down


anupama2501:
**Rancher Server Setup**
- Rancher version: `v2.6-head 8c785a1`
- Installation option (Docker install/Helm Chart): Docker install
   - If Helm Chart, Kubernetes Info: k3s
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):
         Node Setup: 1 node 
         Version: `v1.21.3+k3s1`
- Proxy/Cert Details: `self-signed`

**Information about the Cluster**
- Kubernetes version: 
- Cluster Type (Local/Downstream): `Downstream infrastructure provider 3 worker, 3 etcd, 2 cp`
`RKE2`

**Describe the bug**
Created a deployment with node scheduling `Run pods on specific node(s)` set to a node n1. The node pool has the feature `drain on delete` enabled. If we try to scale down the node, the node is stuck in `deleting`.

**To Reproduce**
1. Create an rke2 node driver cluster with 3 node pools 1 for each role with 3 worker, 3 etcd, 2 cp nodes
2. Enable `drain on delete` feature for each of the node pools.
3. Create a workload deployment 
4. From the node scheduling tab >> select `Run pods on specific node(s)` and select the worker node w1
5. From cluster management >> machine pools >> select the worker node w1 >> scale down

**Result**
The node is stuck in `deleting` and never gets deleted. 

Following are the docker logs that go in a loop:

```
I0223 23:56:24.800925      34 request.go:665] Waited for 2.55470303s due to client-side throttling, not priority and fairness, request: 
GET:https://<>/k8s/clusters/c-m-qsc62p7f/api/v1/namespaces/default/pods/wk1-5c779f5b-jht54?timeout=10s

[...]

PR #36835, opened 2022-03-10
Add NodeDrainTimeout in MachineTemplateSpec

jakefhyde:
## Description

CAPI supports the `NodeDrainTimeout` field on a machine as the duration draining should be attempted before giving up and deleting the infrastructure machine, which will allow deletion of the node to proceed. If the node drain timeout is 0, the infrastructure machine will not be deleted until the node drains successfully.

Related Issue: https://github.com/rancher/rancher/issues/36631

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
@@ -405,8 +405,7 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 		}
 	}
 
-	if _, nodeDrainExcluded := machine.Annotations[capi.ExcludeNodeDrainingAnnotation]; !removed ||
-		(!nodeDrainExcluded && !drainingSucceededCondition.IsTrue(machine) && drainingSucceededCondition.GetReason(machine) != capi.DrainingFailedReason) {
+	if !removed {
 		if err = h.dynamic.EnqueueAfter(obj.GetObjectKind().GroupVersionKind(), infraObj.meta.GetNamespace(), infraObj.meta.GetName(), 5*time.Second); err != nil {
 			return obj, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go
```diff
@@ -362,6 +362,7 @@ func machineDeployments(cluster *rancherv1.Cluster, capiCluster *capi.Cluster, d
 							},
 						},
 						InfrastructureRef: infraRef,
+						NodeDrainTimeout:  machinePool.DrainBeforeDeleteTimeout,
```
thedadams, line 365: We have this problem where every cluster will be reprovisioned when Rancher is upgraded. I think that if we "upgrade" from v2.6-head to this change, then we will not have this problem because `DrainBeforeDeleteTimeout` is a pointer.

Has this been tested with these changes?

jakefhyde: I haven't tested the upgrade case yet, but I will confirm.

jakefhyde: Confirmed this change does not trigger a redeployment

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -405,8 +405,7 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 		}
 	}
 
-	if _, nodeDrainExcluded := machine.Annotations[capi.ExcludeNodeDrainingAnnotation]; !removed ||
-		(!nodeDrainExcluded && !drainingSucceededCondition.IsTrue(machine) && drainingSucceededCondition.GetReason(machine) != capi.DrainingFailedReason) {
+	if !removed {
```
jiaqiluo, line 408: Can you elaborate a little bit on why we can drop the check for annotation and status?

jiaqiluo (changes requested):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
jakefhyde: We drop the check here because we actually want to reenqueue now regardless of whether or not we are draining, because CAPI will automagically force delete the node for us if all if the `NodeDrainTimeout` is reached.

jiaqiluo (approved):

thedadams (approved):

PR #36838, opened 2022-03-10
[HELM] Bump chart apiVersion as it is not Helm 2 compatible

nicholasmhughes:
The current chart version is not compatible with Helm 2, so the chart apiVersion should be moved to `v2`.

Attempting to template the chart with Helm 2 results in the following error:
```
$ helm template rancher/chart
Error: parse error in "rancher/templates/secret.yaml": template: rancher/templates/secret.yaml:3: function "lookup" not defined
```

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 name: rancher
 description: Install Rancher Server to manage Kubernetes clusters across providers.
 version: %VERSION%


nicholasmhughes:
still relevant

catherineluse:
Rancher does support installing helm 2 charts, but each catalog only supports either helm 2 or helm 3, not both.

The built-in catalogs that come with Rancher use helm 3. So you would need to create a new github repo catalog in Rancher and configure it to use Helm 3. Then install the app from there.

catherineluse:
Rancher 2.5.x+ needs to be installed with Helm 3. The last version of Rancher to support Helm 2 installs was 2.4.x https://rancher.com/docs/rancher/v2.0-v2.4/en/installation/resources/advanced/helm2/

nicholasmhughes:
@catherineluse , I agree that it needs to be installed with Helm 3. Therefore, the chart `apiVersion` should be changed accordingly. ([ref](https://helm.sh/docs/topics/charts/#the-apiversion-field))

nicholasmhughes:
still an issue

nicholasmhughes:
still an issue

nicholasmhughes:
still an issue

M3GH4NN:
@prachidamle @Jono-SUSE-Rancher @aiyengar2  could one of you you take a look at this PR request? or could you point me in the direction of someone who could help? Thanks in advance.

M3GH4NN (approved):

aiyengar2 (approved):
LGTM, apologies for the delay!

aiyengar2:
created a forward port PR for 2.7's rancher chart. Thakn you for your contribution @nicholasmhughes!

PR #36850, opened 2022-03-11
add mkisofs to rancher image and jailer.sh

rosskirkpat:
Fix a bug in 2.6-head where rke1 nodes cannot be provisioned due to the jailer for the cluster missing the `mkisofs` binary. This PR adds mkisofs to the rancher image and creates a hard link into the jail.

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -14,7 +14,7 @@ FROM ${K3S_BUILDER} as k3s_builder
 
 FROM registry.suse.com/suse/sle15:15.3
 
-RUN zypper -n install git-core curl ca-certificates unzip xz gzip sed tar shadow gawk vim netcat-openbsd && \
+RUN zypper -n install git-core curl ca-certificates unzip xz gzip sed tar shadow gawk vim netcat-openbsd mkisofs && \
     zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* && \
     useradd rancher && \
     mkdir -p /var/lib/rancher /var/lib/cattle /opt/jail /opt/drivers/management-state/bin && \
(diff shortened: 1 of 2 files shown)


phillipsj (approved):
LGTM!

jiaqiluo (changes requested):
@rosskirkpat  Do you know when/where this new dependence was introduced into rke?   Could you link the GH issue to the PR or create one if it does not exist?

rosskirkpat:
@jiaqiluo Here's the change to machine that introduced the new dependency for mkisofs https://github.com/rancher/rancher/pull/36850#pullrequestreview-907722249. This is all part of adding windows support to vsphere in v2prov.

This is only an issue with rke and is not seen in rke2 as v2prov does not use the jailer. This bug resulted in all rke1 vsphere provisioning breaking once machine80 was brought into rancher.

jiaqiluo:
@rosskirkpat  thanks for the explanation.  Do you mean that although rke1 vsphere provisioning doe not use mkisofs, it has to be in the jail otherwise rancher-machine will not work?

rosskirkpat:
> @rosskirkpat thanks for the explanation. Do you mean that although rke1 vsphere provisioning doe not use mkisofs, it has to be in the jail otherwise rancher-machine will not work?

With that change, any node (rke1 or rke2) provisioned by the vsphere node driver will now use mkisofs instead of godisk-fs. This is why it needs to be in the jail. RKE2 does not use the jail but RKE1 does which was an oversight on my part as this PR should have come in with the machine bump that brought in the mkisofs changes.

jiaqiluo (approved):
LGTM

luthermonson (approved):

thedadams (approved):

PR #36879, opened 2022-03-14
EKS hosted cluster supported

igomez06:
**Framework Updates**
- Implemented helper for provisioning an EKS hosted cluster
- Implemented config file read for the particulars of an EKS cluster and its node group

**Validation Tests**
- Implemented simple test for an EKS cluster


tests/framework/extensions/clusters/clusters.go
```diff
@@ -22,6 +25,24 @@ func IsProvisioningClusterReady(event watch.Event) (ready bool, err error) {
 	return ready, nil
 }
 
+// IsHostedProvisioningClusterReady is basic check function that would be used for the wait.WatchWait func in pkg/wait.
+// This functions just waits until a hosted cluster becomes ready.
+func IsHostedProvisioningClusterReady(event watch.Event) (ready bool, err error) {
```
paynejacob, line 30: wouldn't this work for any cluster

igomez06: Yes for clusters provisioned using the v3 API, I can denote that instead. Like `IsManagmentProvisioningClusterReady` if you prefer.

slickwarren (approved):

izaac (approved):
LGTM

izaac:
While I was provisioning an EKS today with the legacy Python it occurred to me what would be the default behavior for subnets and vpcs.

The `ekscli` creates a new VPC and there's configuration on how to set existing subnets instead of creating a new VPC and its subnets for the cluster.

But we usually hit the VPC limit really quick as there's a limit of 3 VPCs per region in our QA org. 

@igomez06 could you elaborate on how this script will handle those use cases? Will it create VPCs? In that case i'd prefer to set existing subnets instead as we do in our Python framework.

Issue #36877, opened 2022-03-09
Can't delete a custom RKE1 cluster


sgapanovich:

jiaqiluo:
### Root cause
A recent change for delaying the deletion of RBAC stuff makes nodes never to be deleted. This bug happens only in a custom cluster. 
  
### What was fixed, or what changes have occurred
Trigger the deletion of nodes at the beginning of the cluster-scope garbage collection for custom clusters. By doing so, we unblock the deletion of nodes and meanwhile keep RBACs to be deleted last. 

### Areas or cases that should be tested
Provisioning and deleting of custom clusters, including REK, RKE2, and K3S.

### What areas could experience regressions?
It should not affect node-driver clusters but it will be good to do some sanity tests on provisioning and deleting of node-driver clusters.

### Are the repro steps accurate/minimal?
yes.

sowmyav27:
**verified on 2.6-head commit id: `5bb08b`**

Able to delete custom and node driver RKE1 clusters from Rancher UI.

PR #36889, opened 2022-03-14
Delete Nodes during the cluster-scope garbage collection

jiaqiluo:
**Issue:** [#36877](https://github.com/rancher/rancher/issues/36877)

**Problem:** 
The deletion of `Nodes` is triggered by the deletion of the namespace for holding the cluster's resources: https://github.com/rancher/rancher/blob/520235ade2f0db8a8789b98209f2dad577d4e95e/pkg/controllers/management/auth/project_cluster_handler.go#L249
But the namespace never be deleted because the `Remove` function for the cluster's lifecycle returns before deleting the namespace because the there are always more than two finalizers in the `Cluster` object : https://github.com/rancher/rancher/blob/520235ade2f0db8a8789b98209f2dad577d4e95e/pkg/controllers/management/auth/project_cluster_handler.go#L231-L236

The purpose of checking the number of finalizer is to make sure rbac stuff is deleted last. 

 As the result, the deletion of the cluster is stuck because we never trigger the deletion of nodes.

**Fix:** 
Trigger and wait for nodes to be deleted at the beginning of the cluster-scope garbage collection for custom clusters. By doing so, we unblock the deletion of nodes and meanwhile keep RBACs to be deleted last. 

**Tests:**
I have tested provisioning and removing RKE1 custom clusters as both admin and restricted admin. Clusters were removed successfully.

### DIFF
--- a/pkg/controllers/management/clustergc/cluster_scoped_gc.go
+++ b/pkg/controllers/management/clustergc/cluster_scoped_gc.go
@@ -75,15 +75,15 @@ func (c *gcLifecycle) waitForNodeRemoval(cluster *v3.Cluster) error {
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
-	var wait bool
+	var waitForNodeDelete bool
 	for _, n := range nodes {
 		// trigger the deletion of node for a custom cluster
 		if n.Status.NodeTemplateSpec == nil && n.DeletionTimestamp == nil {
 			_ = c.mgmt.Management.Nodes(n.Namespace).Delete(n.Name, &metav1.DeleteOptions{})
-			wait = true
+			waitForNodeDelete = true
 		}
 	}
-	if wait {
+	if waitForNodeDelete {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/clustergc/cluster_scoped_gc.go
```diff
@@ -75,14 +75,19 @@ func (c *gcLifecycle) waitForNodeRemoval(cluster *v3.Cluster) error {
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
-
+	var wait bool
```
paynejacob, line 78: Über nit: I would rename this waitForNodeDelete

paynejacob (approved):

pkg/controllers/management/clustergc/cluster_scoped_gc.go
jiaqiluo: fixed.

annablender (approved):
Code LGTM

Did you make sure to check that the deletion completes for both an rke1 cluster created in Rancher and a registered rke1 cluster?

paynejacob (approved):

Issue #36860, opened 2022-03-11
Nodes showing incorrect roles in an RKE1 node driver cluster


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6.4-rc8
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: 1.20
- Cluster Type (Local/Downstream): Node driver EC2 0 3 etcd, 2 cp and 3 worker nodes

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) admin

**Describe the bug**
Nodes showing incorrect roles in an RKE1 node driver cluster

**To Reproduce**

- Deployed 2 RKE1 EC2 clusters:
  - 3 etcd, 2 cp and 3 worker nodes
  - 1 etcd node pool, 1 cp nodepool and 1 worker node pool
- On one cluster - one CP node shows all roles
- This node does not have etcd container on the node when sshed into
- Kubectl get nodes, show one cp node has `all roles`
- On another cluster, 1 etcd node shows all roles
- This node does not have control plane components

**Expected Result**
All roles should not be available for etcd ONLY or control plan only node

**Note:**
- validated this on 2.6.3, same cluster node config and this issue is not seen

snasovich:
This needs to be reproduced and investigated; there were no planned changes around this functionality per @kinarashah , so may take "bisecting commits" approach to figure out what broke it.

kinarashah:
I kept running into this issue while trying to reproduce https://github.com/rancher/rancher/issues/36862, maybe the same regression is causing both issues. 
Able to reproduce this issue consistently with - 

**Rancher server version**: `>=v2.6.4-rc5` 
**Install type**: Docker single node install 

**Steps:**
- Provision a downstream RKE1 ec2 node driver cluster with 3 different machine pools having 3 etcd, 2 cp and 2 worker nodes respectively 
- Initially, the nodes have correct roles so wait for provisioning to complete 
- After the cluster is active, one of the etcd only/control plane only roles will have incorrect `All` roles

sowmyav27:
**On 2.6-head commit id: `5bb08b3`**

- deployed an RKE1 - 3 etcd, 2 cp and 3 wworker nodes cluster
- No All roles node is seen

[...]

PR #36908, opened 2022-03-15
Added deepcopy when modifying node roles for nonWorker

jakefhyde:
https://github.com/rancher/rancher/issues/36860

Sometimes the `cluster` parameter will be updated in the apiserver (outside of this function), so a `DeepCopy()` is required to prevent modifying the cluster (and thus rewriting the first node to have all roles).


thedadams (approved):

kinarashah (approved):
@jakefhyde Could you also link the commit that changed this for future tracking?

jakefhyde:
Issue originated from this commit: https://github.com/rancher/rancher/commit/66d24a7cd870561da94b2339de4430d91f1d0826

PR #36909, opened 2022-03-15
adding wins agent setting pin so binary base url works

luthermonson:
Problem: the wins installer never had an env set for the wins-agent-version. This resulted in the install.ps1 not getting the env var set so that we use a pinned version from the assets server and pulling latest from github.

Fix: add the env var, fix so $env: is prepended to the binaryUrl and add a test


thedadams (dismissed):

rosskirkpat (approved):
LGTM

thedadams (approved):

PR #36912, opened 2022-03-15
make rancher/machine repo a configurable param

rosskirkpat:
This change allows validation of rancher/machine changes without needing to copy in the `rancher-machine` binary from your local machine. 

A dev can push their custom `rancher-machine-${ARCH}.tar.gz` to their fork and Rancher would now process that tar.gz and replace the embedded `rancher-machine` binary in the jail with their custom binary.

### DIFF
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -2,6 +2,9 @@ package jailer
 
 import (
 	"context"
+	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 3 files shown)


pkg/jailer/jailer.go
```diff
@@ -25,6 +28,14 @@ func CreateJail(name string) error {
 		return os.MkdirAll(path.Join(BaseJailPath, name), 0700)
 	}
 
+	if os.Getenv("MACHINE_REPO") != "rancher/machine" {
```
luthermonson, line 31: probably should switch to a new setting and use CATTLE_MACHINE_REPO and get the setting value here so it's more standardized

PR #36928, opened 2022-03-16
adding csi proxy to ui assets + version setting to control it

luthermonson:
Problem: if airgapped and you try to install CSI Proxy you will be met with an outbound call to an azure endpoint to download the node in wins. This needs to be configured so we can use the ui assets server in rancher to allow the wins install script to get configured to point at the rancher assets server to pull CSI Proxy.

Fix: Add two settings for proxy url and version, default version to be the azure endpoint and pass v1.0.0 in the dockerfile to lock the version in the assets server.

This is a sister PR to https://github.com/rancher/rancher/pull/36909 which got wins.exe into the ui assets server.


phillipsj (commented):
LGTM

rosskirkpat (approved):
LGTM!

thedadams (approved):

PR #36932, opened 2022-03-17
update dynamiclistener to released v0.3.3

steved:
Issue: #37766
  
# Problem

When creating a Rancher cluster with a `server_url` that would cause an overflow of the TLS cert annotation, an rke cluster could not be provisioned:
```
2022/03/17 00:09:46 [INFO] failed to create TLS cert for: 127.0.0.1, Secret "serving-cert" is invalid: metadata.annotations: Invalid value: "listener.cattle.io/cn-internal-abcdefghij-rancher-lb-int-123456789.us-west-2.elb.amazonaws.com": name part must be no more than 63 characters
```

# Solution

This PR updates the dynamiclistener dependency to the latest ~commit of v0.3.1~ release in order to pick up the following change: https://github.com/rancher/dynamiclistener/pull/49
  
# Testing

Create a Rancher deployment with a `server_url` setting that is longer than 63 characters.
 
### Manual Testing

Deployed a Rancher instance internally that is frontend by an ELB.
 
### Automated Testing
_TBD_
 
## QA Testing Considerations
_TBD_
  
_TBD_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -92,7 +92,7 @@ require (
 	github.com/rancher/aks-operator v1.0.5
 	github.com/rancher/apiserver v0.0.0-20220223185512-c4e289f92e46
 	github.com/rancher/channelserver v0.5.1-0.20220405170618-28c9b37deff1
-	github.com/rancher/dynamiclistener v0.3.2
+	github.com/rancher/dynamiclistener v0.3.3
 	github.com/rancher/eks-operator v1.1.3
 	github.com/rancher/fleet/pkg/apis v0.0.0-20210918015053-5a141a6b18f0
 	github.com/rancher/gke-operator v1.1.3
--- a/go.mod
+++ b/go.mod
@@ -92,7 +92,7 @@ require (
 	github.com/rancher/aks-operator v1.0.5
 	github.com/rancher/apiserver v0.0.0-20220223185512-c4e289f92e46
 	github.com/rancher/channelserver v0.5.1-0.20220405170618-28c9b37deff1
-	github.com/rancher/dynamiclistener v0.3.1-0.20210616080009-9865ae859c7f
+	github.com/rancher/dynamiclistener v0.3.2
 	github.com/rancher/eks-operator v1.1.3
 	github.com/rancher/fleet/pkg/apis v0.0.0-20210918015053-5a141a6b18f0
 	github.com/rancher/gke-operator v1.1.3
--- a/go.sum
+++ b/go.sum
@@ -1249,8 +1249,8 @@ github.com/rancher/channelserver v0.5.1-0.20220405170618-28c9b37deff1/go.mod h1:
 github.com/rancher/client-go v1.23.3-rancher2 h1:99raSqSdtNF1aQCJ45r/1t8KZmld6AkA4+nsfalnA98=
 github.com/rancher/client-go v1.23.3-rancher2/go.mod h1:47oMd+YvAOqZM7pcQ6neJtBiFH7alOyfunYN48VsmwE=
 github.com/rancher/dynamiclistener v0.2.1-0.20200714201033-9c1939da3af9/go.mod h1:qr0QfhwzcVCR+Ao9WyfnE+jmOpfEAdRhXtNOZGJ3nCQ=
-github.com/rancher/dynamiclistener v0.3.1-0.20210616080009-9865ae859c7f h1:4csSA3WprYZk0uQIaYguJeHv2LM5zOCKxiBd18pBvEk=
-github.com/rancher/dynamiclistener v0.3.1-0.20210616080009-9865ae859c7f/go.mod h1:xbxyiF1/UaPrOrFUOZeYMQW82RPgZIe5UfiT9CyE3Ks=
+github.com/rancher/dynamiclistener v0.3.2 h1:gX1U7V1ifLsmRM86DkvrZQfPq9WBT8EatKieyIbQYSY=
+github.com/rancher/dynamiclistener v0.3.2/go.mod h1:QwTpy+drx4gvPMefrrUUKpVaWiy74O7vNvkwBXJ+s3E=
 github.com/rancher/eks-operator v1.1.3 h1:x1c0T0mXjKA3HHbbQsfx9k8xFHsNo0GSYGO/5yQOgjo=
 github.com/rancher/eks-operator v1.1.3/go.mod h1:N0zzDJ+xvv6BkXRMjIBlFObOmud5H1sSOccG0FS+DJM=
 github.com/rancher/fleet/pkg/apis v0.0.0-20210918015053-5a141a6b18f0 h1:2A92uT1+4GUdZolQg5wfOSLa6RFRkfpnc3Iw7+Z/Bk8=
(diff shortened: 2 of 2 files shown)


steved:
cc @brandond

brandond (dismissed):
👍🏻

steved:
@brandond @luthermonson with 2.6.4 released, I've rebased the branch. thanks for your consideration!

luthermonson:
@steved sorry forgot to get this when you had it rebased. pls fix the conflict again and we'll merge

steved:
My turn to apologize, I was on vacation! @luthermonson PTAL.

steved:
Bump

brandond:
I just tagged another release if you want to bump it again..

steved:
@brandond done!

brandond (approved):

steved:
@brandond this change fixes the test issues (which seem legit, the certs are never saved to k8s): https://github.com/rancher/dynamiclistener/compare/master...steved:steved/fix-rancher. The line was added in your change, and I'm having a hard time figuring out where the informer sync is supposed to be.

brandond:
@steved do you have a PR for that change that we can discuss on? The point of that change is to ensure that the controller has initialized before we try to use it. Are you saying that the rancher is somehow running the controller without an informer? Or is that just true during the tests?

steved:
Thanks for the explanation, I missed that `start.All`! I think Rancher is calling `ListenAndServe` which ends up in a [Load](https://github.com/rancher/dynamiclistener/blob/b1d65efb6f8309327ed7beb41463056379ec18eb/server/server.go#L187) call bypassing that init.

I've opened https://github.com/rancher/dynamiclistener/pull/58, but it doesn't sound like that's the correct fix potentially.

brandond:
It looks like rancher is passing in the secrets controller at https://github.com/rancher/rancher/blob/dfe44512d12d5674607be1848e53ce83f5b67e6c/pkg/tls/tls.go#L68 by way of https://github.com/rancher/rancher/blob/dfe44512d12d5674607be1848e53ce83f5b67e6c/pkg/tls/tls.go#L242-L243 but for some reason that Secrets controller isn't getting started by Rancher so the informer sync check never returns true?

steved:
That looks right, yep! What do you think of 526876e3ade2b80b831b68d05fbd479af8a41fd1?

brandond:
I suspect that'll do it, lets see!

Once we get this squared away I want to go back and add some tests to dynamiclistener, the two entry points used by rancher and k3s/rke2 to set up the stores and listener are different enough that they need proper testing so that changes for one don't keep breaking the other.

brandond:
OK, so I was poking at this on the side and it looks like `start.All` needs to be called AFTER listenAndServe. If you try to start it before getting at any of the controllers, the start is a no-op, which makes sense I guess. There still appears to be another issue, where using ListenAndServe won't preload the certificate, and the first handshake to the endpoint fails. Using `dynamiclistener.NewListener` somehow orders the startup differently; I'm still poking at it to figure out how.

A minimal reproducer (adapted from something I was playing with for setting up GRPC servers behind DynamicListener) is available at https://github.com/brandond/dynamicgrpc/blob/main/main.go - just export KUBECONFIG pointing it at a running cluster and it'll set up a basic server on 8080/8443.

brandond:
Ah, I see. The change I made requires there be some non-Kubernetes storage available, but ListenAndServe does not provide that, and rancher does not set one up itself.

Moving start.All after the listenAndServe, and having `SetupListener` set `Storage: memory.New()`  from `"github.com/rancher/dynamiclistener/storage/memory"` should work around that issue here in Rancher, until we can fix it in dynamiclistener.

steved:
Hm, I see a default being set in ListenAndServe:
https://github.com/rancher/dynamiclistener/blob/b1d65efb6f8309327ed7beb41463056379ec18eb/server/server.go#L127

I also totally missed that the core controller was being started already:
https://github.com/rancher/rancher/blob/a7ae781e8c84aa73a9d0bc6ccfd79ca512579af0/pkg/tls/tls.go#L145

Currently, my understanding of the order of events is:
1. `dynamiclistener.Listener#Accept` is called, which tries to initialize the cert: 
  https://github.com/rancher/dynamiclistener/blob/b1d65efb6f8309327ed7beb41463056379ec18eb/listener.go#L265
  https://github.com/rancher/dynamiclistener/blob/b1d65efb6f8309327ed7beb41463056379ec18eb/listener.go#L378
2. `core` hasn't started, so the k8s controller skips the creation of the cert in `saveInK8s`
3. the only other mechanism to create the cert (besides `kubernetes#init`) is the `OnChange` watcher which cannot be triggered because it only watches for changes on an existing secret or some change in the listener triggering an update of the SANs/CN

(sorry for the CI noise, somehow didn't notice I had vendored deps and that's why my tests passed locally 🤦 )

brandond:
Yeah, I'm walking my way back through this and I think I see where the problem is coming from; it has something to do with how the storage is set up - I believe it only works properly if the outer storage is kubernetes, not memory or file.

steved:
> Explicitly setting the storage to Memory makes it work better, but the cert still doesn't get written back into kubernetes, which is why the tests are failing. I think this'll need to be fixed on the dynamiclistener side - I can get a PR up tomorrow AM, or I can throw some suggestions into your existing PR if you prefer.

I think I've exhausted my understanding of the best way to fix this :) Would be really great if you could put up a PR, I appreciate it.

brandond:
My apologie, you were right all along - it was that informer sync check that was breaking things. I was just having a hard time convincing myself of the root cause, and failed to replicate it because my simple reproducer wasn't setting any initial SANs in the `TLSListenerConfig` - which turns out to cause some important different behavior.

K3s is fine with the current behavior because the secret gets updated a bunch of times during cluster startup to add different hostnames/IPs and it was *eventually* getting written to Kubernetes, but in Rancher dynamiclistener responds to a pretty static list of hostnames and the secret just gets set once. That single write was not being written through to Kubernetes, which caused other things to fail because the secret wasn't getting created as expected.

https://github.com/rancher/dynamiclistener/pull/59

brandond:
@steved can you try again with v0.3.3?

steved:
@brandond updated and successfully tested internally! Thanks for the help!

It appears the last CI run failed because of an intermittent connection error, but I don't have permission to retry it.

brandond:
Restarted; lets see if it flakes on windows again.

steved:
@brandond I see some somewhat similar test flakiness in release/v2.6. I'm unable to reproduce the test error (haven't tried the rke provisioning one, but the server logs don't look super relevant).

brandond:
I kicked CI a few times and it appears to be green now @steved

brandond (approved):

steved:
Bump, thanks!

cbron:
Before we merge, can someone fill this out please so QA knows what to test here. And what issue is this linked with, is it 37766 ? If not can someone make one. 

```
Issue: <!-- link the issue or issues this PR resolves here -->
<!-- If your PR depends on changes from another pr link them here and describe why they are needed on your solution section. -->
  
# Problem
<!-- Describe the root cause of the issue you are resolving. This may include what behavior is observed and why it is not desirable. If this is a new feature describe why we need this feature and how it will be used. -->
  
# Solution
<!-- Describe what you changed to fix the issue. Relate your changes back to the original issue / feature and explain why this addresses the issue. -->
  
# Testing
<!-- Note: Confirm if the repro steps in the GitHub issue are valid, if not, please update the issue with accurate repro steps. -->
 
## Engineering Testing
### Manual Testing
<!-- Describe what manual testing you did (if no testing was done, explain why). -->
 
### Automated Testing
<!--If you added/updated unit/integration/validation tests, describe what cases they cover and do not cover. -->
 
## QA Testing Considerations
<!-- Highlight areas or (additional) cases that QA should test -->
  
<!-- Dedicated section to specifically call out any areas that with higher chance of regressions caused by this change, include estimation of probability of regressions -->
```

steved:
@cbron I gave it my best shot, but I'm not sure about some of the testing consideration sections.

cbron:
@steved that's ok, where is it though ?

steved:
Sorry, edited the PR description.

maxsokolovsky (approved):

steved:
bump @cbron @brandond

brandond:
I think @cbron or someone else from the RM side should merge this.

KevinJoiner (approved):

Issue #36096, opened 2022-01-10
Intermittent user auth errors relating to "impersonation" ClusterRoles


armsnyder:
**Rancher Server Setup**
- Rancher version: v2.6.2
- Installation option (Docker install/Helm Chart): Helm chart on EKS cluster
- Proxy/Cert Details: Using nginx ingress-controller behind AWS NLB, with the nginx ingress handling TLS

**Information about the Cluster**
- Kubernetes version: v1.19.13
- Cluster Type (Local/Downstream): Imported (cluster is an EKS cluster)

**User Information**
- What is the role of the user logged in? Admin.
   
**Describe the bug**

A couple users have noticed intermittent authentication errors using the Kubernetes API. It is especially prevalent when browsing the rancher-monitoring Grafana due to number of API requests but occurs in the terminal as well.

The error message is:

```
Templating
Template variable service failed userextras.authentication.k9s.io "redacted@redacted.com" is forbidden: User "system:serviceaccount:cattle-impersonation-system:cattle-impersonation-u-redacted" cannot impersonate resource "userextras/username" in API group "authentication.k8s.io" at the cluster scope.
```

```json
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "userextras.authentication.k8s.io \"okta_user://redacted@redacted.com\" is forbidden: User \"system:serviceaccount:cattle-impersonation-system:cattle-impersonation-u-redacted\" cannot impersonate resource \"userextras/principalid\" in API group \"authentication.k8s.io\" at the cluster scope",
  "reason": "Forbidden",
  "details": {
    "name": "okta_user://redacted@redacted.com",
    "group": "authentication.k8s.io",
    "kind": "userextras"
  },
  "code": 403
}
```

**To Reproduce**

Set up Rancher using the Okta SAML2 auth provider using Rancher v2.6.2. Some users will experience this issue.

**Additional context**

We spent time investigating and discovered that the `cattle-impersonation-system:cattle-impersonation-u-redacted` ClusterRole mentioned in the error message had some empty `resourceNames` in the rules. The ClusterRole looked like:

```yaml
...
- apiGroups:
  - authentication.k8s.io
  resourceNames:
  - ""
  resources:
  - userextras/principalid
  verbs:
  - impersonate
- apiGroups:
  - authentication.k8s.io
  resourceNames:
  - ""
  resources:
  - userextras/username
  verbs:
  - impersonate
 ```

[...]

Issue #34822, opened 2021-09-20
Use `UserAttribute.ExtraByProvider` to populate impersonation role rules


cmurphy:
Followup from https://github.com/rancher/rancher/issues/34671

Depends on https://github.com/rancher/rancher/pull/33268. That PR will introduce a unified way to get a user's extra attributes from their auth provider, which we will be able to leverage to simplify how we generate the clusterrole rules related to user extras.

samjustus:
SURE-4361

cmurphy:
Moving to test, but I think this can just be tested via https://github.com/rancher/rancher/issues/36096

brudnak:
Tested via  https://github.com/rancher/rancher/issues/36096#issuecomment-1103222428

PR #36951, opened 2022-03-17
[2.6.5] Use UserAttributes for impersonation

cmurphy:
The impersonation module checks for Extra attributes in the user.Info
object and adds principal IDs and usernames from an auth provider to the
impersonation ClusterRole rules. Without this patch, this only works
when a user makes a request, because the request includes the
Extra data in the token, but does not work for the role template binding
controllers, which only send the user's UID and groups. The result is
that when a user sends a request, the ClusterRole is updated to include
the Extra data from the request, but after some time, the CRTB and PRTB
controllers will change it back.

With d9ba4d the user's principal IDs and usernames from the auth
provider should now be populated in the UserAttributes for that user, so
we can rely on that instead of whatever data may be in the user.Info
object. This way the data will be correct no matter whether this is
called from the request proxy or the role template binding controller.

https://github.com/rancher/rancher/issues/34822
https://github.com/rancher/rancher/issues/36096

### DIFF
--- a/pkg/controllers/managementuser/rbac/impersonation_handler.go
+++ b/pkg/controllers/managementuser/rbac/impersonation_handler.go
@@ -8,38 +8,6 @@ import (
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
-func (m *manager) getUser(username, groupname string) (user.Info, error) {
-	u, err := m.userLister.Get("", username)
-	if err != nil {
-		return &user.DefaultInfo{}, err
-	}
-	groups := []string{"system:authenticated", "system:cattle:authenticated"}
-	if groupname != "" {
-		groups = append(groups, groupname)
-	}
-	attribs, err := m.userAttributeLister.Get("", username)
-	if err != nil && !apierrors.IsNotFound(err) {
-		return &user.DefaultInfo{}, err
-	}
-	if attribs != nil {
-		for _, gps := range attribs.GroupPrincipals {
-			for _, principal := range gps.Items {
-				groups = append(groups, principal.Name)
-			}
-		}
-	}
-	user := &user.DefaultInfo{
-		UID:    u.GetName(),
-		Name:   u.Username,
-		Groups: groups,
-		Extra:  map[string][]string{"username": []string{u.Username}},
-	}
-	if len(u.PrincipalIDs) > 0 {
-		user.Extra["principalid"] = u.PrincipalIDs
-	}
-	return user, nil
-}
-
 func (m *manager) ensureServiceAccountImpersonator(username string) error {
 	logrus.Debugf("ensuring service account impersonator for %s", username)
 	i, err := impersonation.New(&user.DefaultInfo{UID: username}, m.workload)
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"time"
 
+	authcommon "github.com/rancher/rancher/pkg/auth/providers/common"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/sirupsen/logrus"
@@ -252,7 +253,7 @@ func (i *Impersonator) rulesForUser() []rbacv1.PolicyRule {
 		})
 	}
 	extras := i.user.GetExtra()
-	if principalids, ok := extras["principalid"]; ok {
+	if principalids, ok := extras[authcommon.UserAttributePrincipalID]; ok {
 		rules = append(rules, rbacv1.PolicyRule{
 			Verbs:         []string{"impersonate"},
 			APIGroups:     []string{"authentication.k8s.io"},
@@ -260,7 +261,7 @@ func (i *Impersonator) rulesForUser() []rbacv1.PolicyRule {
 			ResourceNames: principalids,
 		})
 	}
-	if usernames, ok := extras["username"]; ok {
+	if usernames, ok := extras[authcommon.UserAttributeUserName]; ok {
 		rules = append(rules, rbacv1.PolicyRule{
 			Verbs:         []string{"impersonate"},
 			APIGroups:     []string{"authentication.k8s.io"},
@@ -361,11 +362,21 @@ func (i *Impersonator) getUser(userInfo user.Info) (user.Info, error) {
 	}
 
 	groups := []string{"system:authenticated", "system:cattle:authenticated"}
+	extras := make(map[string][]string)
 	attribs, err := i.userAttributeLister.Get("", userInfo.GetUID())
 	if err != nil && !apierrors.IsNotFound(err) {
 		return &user.DefaultInfo{}, err
 	}
-	if attribs != nil {
+	if attribs == nil { // system users do not have userattributes, but principalid and username are on the user
+		// See https://github.com/rancher/rancher/blob/7ce603ea90ca656f5baa29b0149c19c8d7f73e8f/pkg/auth/requests/authenticate.go#L185-L194
+		// If the extras are not in userattributes, use displayName and principalIDs from the user.
+		if u.DisplayName != "" {
+			extras[authcommon.UserAttributeUserName] = []string{u.DisplayName}
+		}
+		if len(u.PrincipalIDs) > 0 {
+			extras[authcommon.UserAttributePrincipalID] = u.PrincipalIDs
+		}
+	} else { // real users have groups and extras in userattributes
 		for _, gps := range attribs.GroupPrincipals {
 			for _, groupPrincipal := range gps.Items {
 				if !isInList(groupPrincipal.Name, groups) {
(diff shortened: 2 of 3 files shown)


ryansann (approved):
lgtm

cmurphy:
Seeing some strange errors with this, will need to take a closer look

pkg/impersonation/impersonation.go
```diff
@@ -373,12 +374,23 @@ func (i *Impersonator) getUser(userInfo user.Info) (user.Info, error) {
 				}
 			}
 		}
+		for _, exs := range attribs.ExtraByProvider {
```
MbolotSuse, line 377: I think I understand the point of this code, but I have a few questions on the functionality:
1) Why do we single out the "username" and "principalid" extras? I get that these are important extras, but the prior version of the code seems to have captured all extra values. (L380 in prior version)
2) I do see a couple of places where we get the extras from the UserInfo directly. Is this the "canonical way" that we would want this to happen?
3) Instead of replacing all of the extras from the userInfo would we instead want to merge the two maps to get all the extras possible (although I'm kind of thinking that ideally we would want these values to be equal, so really it shouldn't matter)?

MbolotSuse (commented):
Got some questions - overall I think I understand the purpose, but I'm not sure about how this interacts with other areas and why some of the filtering was done.

pkg/controllers/managementuser/rbac/impersonation_handler.go
```diff
@@ -8,38 +8,6 @@ import (
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
-func (m *manager) getUser(username, groupname string) (user.Info, error) {
```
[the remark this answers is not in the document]
cmurphy: Yes, I split this into a separate commit with its own explanation https://github.com/rancher/rancher/pull/36951/commits/1226d517d82f6dd309dc27e0d0d8d2492e6d34d5

pkg/impersonation/impersonation.go
cmurphy: 1. `username` and `principalid` are sent with the user's request and included in the impersonation headers, they are not chosen by us. An example would look something like this: `curl -H "Impersonate-User: user-abcde" -H "Impersonate-Group: cattle:authenticated" -H "Impersonate-Extra-PrincipalID: "github://abc" -H "Impersonate-Extra-Username: cmurphy"`, and if the ClusterRole does not include rules for those extra fields, the request will be denied due to missing RBAC rules.
2. This is the crux of the issue, the extras may not be included in the UserInfo object, depending on whether this is called from a role template binding controller or from the proxy request handler. It's more consistent to get the data from the UserAttribute object than to trust the UserInfo object.
3. With this change, if any extras are included in UserInfo they would be redundant with what is stored in the UserAttribute object. If that were not true, then the rules in the ClusterRole would flap depending on whether the caller included complete data in the UserInfo object.

MbolotSuse: Ok - I think I have a better understanding after reading some of the other areas of the code. Let me rephrase here to see if I have a good understanding:

1. We use the usename/principalid as part of the cluster role that we create for the impersonation logic. So we have to extract those values because we need them in the `rulesForUser()` call, but we don't need any other extras, so we don't bother extracting those.
2. This new way of getting extras (from the attributes rather than from the user info) is the preferred way and should be used from now on. I am assuming that we will eventually get around to switching the other areas that use this code at some point in the future.
3. We expect the userInfo to have a subset of what's in the userAttributes. Anything more would be a separate defect and not part of this PR's scope.

Let me know if any of this is incorrect. But I think that this makes sense to me now.

pkg/impersonation/impersonation.go
```diff
@@ -361,24 +361,45 @@ func (i *Impersonator) getUser(userInfo user.Info) (user.Info, error) {
 	}
 
 	groups := []string{"system:authenticated", "system:cattle:authenticated"}
+	extras := make(map[string][]string)
 	attribs, err := i.userAttributeLister.Get("", userInfo.GetUID())
 	if err != nil && !apierrors.IsNotFound(err) {
 		return &user.DefaultInfo{}, err
 	}
-	if attribs != nil {
+	if attribs == nil { // system users do not have userattributes, but principalid and username are on the user
```
MbolotSuse, line 369: My concern with this if statement is that it links system users to not having userattributes at all. I don't know if we can take it for granted that this will always be true. In addition, if there's a change made which gives system users a user attribute but doesn't give a user attribute for these values (username/principal id) this code will fail.

One other way that you could do this would be to check if we found any username/ by going through the attributes, and if we didn't, use the values from the user. Below is what I am thinking:
```
if attribs != nil{
  // get attributes using current code
}
if usernames, ok := extras["username"]; !ok || len(usernames) == 0 {
    if u.Username != ""{
        extras["principalid"] = []string{u.PrincipalIDs}
    }else if u.DisplayName != ""{
        extras["username"] = []string{u.Username}
    }
}

if principalids, ok := extras["principalid"]; !ok || len(principalids) == 0{
    extras["principalid"] = u.PrincipalIDs
}
```
There's also the case where the attributes did not include the username/principalIDs which could be found on the user - for that case you'd have to loop through what you did find in the attribs and ensure that the username of the user and principalIDs from the user are in the overall master list. But that's a good bit of extra code/processing - interesting in hearing if you think it is necessary.

MbolotSuse (commented):
1 Question on combining the values found from the attribs with the values present on the User (not user info - the actual user that we get back from k8s).

pkg/impersonation/impersonation.go
cmurphy: > but we don't need any other extras, so we don't bother extracting those.

There are no other extras, there are only these two.

> This new way of getting extras (from the attributes rather than from the user info) is the preferred way and should be used from now on.

It would depend on what it is being used for and what is creating the user info object. The user info object isn't a static state stored in etcd, it's an object that any component can create on the fly.

> We expect the userInfo to have a subset of what's in the userAttributes.

With this change, the only requirement for the user info object is that it has the user's UID. That is used to look up the User and UserAttributes.

pkg/impersonation/impersonation.go
cmurphy: > My concern with this if statement is that it links system users to not having userattributes at all. I don't know if we can take it for granted that this will always be true.

In my testing I found this to be the case. UserAttributes are only created when a real life user logs in and gets a token, and this does not happen for system users. @prachidamle can you comment on whether this is reliable?

> In addition, if there's a change made which gives system users a user attribute but doesn't give a user attribute for these values (username/principal id) this code will fail.

The code will not fail, it checks whether there are non-empty values for the username and principal ID. If they are not included in the UserAttributes, then they are not needed.

> One other way that you could do this would be to check if we found any username/ by going through the attributes, and if we didn't, use the values from the user
> There's also the case where the attributes did not include the username/principalIDs which could be found on the user 

If UserAttributes exists, it will be more complete than what is on the User because it will include data from the auth provider. There is no need to do extra processing.

pkg/impersonation/impersonation.go
MbolotSuse: > There are no other extras, there are only these two.

Correct me if I am wrong, but isn't the extras just arbitrary key(string), value (string) pairs that the auth provider thought was important? My source was the [k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/authentication/), but I might be misreading what is there. I would think that this could include any number of values outside of just principalid and username.

> It would depend on what it is being used for and what is creating the user info object. The user info object isn't a static state stored in etcd, it's an object that any component can create on the fly.

Thanks for explaining. I think that I understand. I'm thinking in light of this PR - we would want the ultimate source of the extras (anywhere else) to be the user attributes. But that's out of scope here - I don't mean to get us off track.

pkg/impersonation/impersonation.go
cmurphy: > Let's say that there is one user attribute created for a system user 

These aren't created for system users. They are created when a real user logs in: https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/tokens/manager.go#L648-L659

> Suppose that this attribute doesn't have the principal Id or the username in it. Since we got a user attribute back, this code won't use the values that are present for the system user on the user object.

The UserAttributes object, if it exists, is the authoritative source for user attributes, since it includes data from the auth provider.

> Is there anything that makes that guarantee

Rancher always adds the principal ID and username for every auth provider:

https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/providers/github/github_provider.go#L412-L421
https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/providers/saml/saml_provider.go#L399-L408
https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/providers/ldap/ldap_provider.go#L373-L382

pkg/impersonation/impersonation.go
cmurphy: > Correct me if I am wrong, but isn't the extras just arbitrary key(string), value (string) pairs that the auth provider thought was important?

It is not arbitrary for Rancher, it is always principal ID and username, examples:

https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/providers/github/github_provider.go#L412-L421
https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/providers/saml/saml_provider.go#L399-L408
https://github.com/rancher/rancher/blob/c2d8e324552a9287a726e0cb7df7fe5971169e9e/pkg/auth/providers/ldap/ldap_provider.go#L373-L382

Remember that these are not arbitrary headers coming from the user, these are requests that are proxied through Rancher so Rancher controls the headers.

pkg/impersonation/impersonation.go
```diff
@@ -361,24 +361,45 @@ func (i *Impersonator) getUser(userInfo user.Info) (user.Info, error) {
...
+		if u.Username != "" {
+			extras["username"] = []string{u.Username}
+		} else if u.DisplayName != "" {
+			extras["username"] = []string{u.DisplayName}
+		}
+		if len(u.PrincipalIDs) > 0 {
+			extras["principalid"] = u.PrincipalIDs
```
MbolotSuse, line 376: Small Item - In light of [#35325](https://github.com/rancher/rancher/pull/35325), would you mind changing the hardcoded strings here (and in L371, L373, L387, L388, L391, L393, etc.) to use a constant? I see that pkg/auth/providers/common/provider.go has two for these values, but you can also just define them at the top of the file if you don't want to import that (or can't for some reason).

pkg/impersonation/impersonation.go
cmurphy: Changed this to only look for DisplayName for the `username` extra field, resolving

pkg/impersonation/impersonation.go

MbolotSuse (dismissed):
LGTM

ryansann (approved):

pkg/impersonation/impersonation.go
prachidamle: I think this concern is already resolved, but yes the UserAttributes are only created when a real life user logs in and a token is created for that user. For system users having local auth, the UserAttributes are not created.

prachidamle (approved):

ryansann (approved):

PR #36962, opened 2022-03-18
Updated Catalog client to support chart installation in downstream clusters

igomez06:
**Background:**
Currently in the framework when attempting to install a chart we would just hit this endpoint `v1/catalog.cattle.io.clusterrepos` as a result chart installation was only supported in local clusters.

**Framework updates:**
Updated Catalog client to support chart installation in downstream clusters with the prefix `/k8s/clusters/<clusterid>`. Updated the ranchermonitoring extension to account for this new change.

### DIFF
--- a/tests/framework/clients/rancher/client.go
+++ b/tests/framework/clients/rancher/client.go
@@ -157,6 +157,19 @@ func (c *Client) WithSession(session *session.Session) (*Client, error) {
 	return NewClient(c.restConfig.BearerToken, session)
 }
 
+// GetClusterCatalogClient is a function that takes a clusterID and instanitates a catalog client to directly communicate with that specific cluster.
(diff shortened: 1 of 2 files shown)


tests/framework/clients/rancher/client.go
```diff
@@ -157,6 +149,19 @@ func (c *Client) WithSession(session *session.Session) (*Client, error) {
 	return NewClient(c.restConfig.BearerToken, session)
 }
 
+// Catalog is a function that takes a clusterID and instanitates a catalog client to directly communicate with that specific cluster.
+func (c *Client) Catalog(clusterID string) (*catalog.Client, error) {
```
paynejacob, line 153: I would keep the original client incase we want to install charts on local.  This could be renamed "GetClusterCatalogClient"

paynejacob (approved):

bmdepesa (approved):

PR #37013, opened 2022-03-23
set local k3s to 1.22.7

kinarashah:
1.23 is experimental for 2.6.4, so setting default to 1.22. Changing packaged binaries back to containerd because it was
an update specific to 1.23 k3s.


kinarashah:
https://github.com/rancher/rancher/issues/37067

thedadams (approved):

Issue #37023, opened 2022-03-24
Rancher panic if helm release secret is empty


gprossliner:
**Rancher Server Setup**
- Rancher version:
v2.6.3-patch2

- Installation option (Docker install/Helm Chart):
Helm RKEv1

**Information about the Cluster**
- Kubernetes version:
v1.20.5

- Cluster Type (Local/Downstream):
vsphere provisioned 

**User Information**
cluster-admin role

**Describe the bug**
We observed the following panic in the rancher log:

```
E0323 15:14:59.049352 33 runtime.go:78] Observed a panic: runtime.boundsError{x:3, y:0, signed:true, code:0x2} (runtime error: slice bounds out of range [:3] with capacity 0)
goroutine 3517 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x41ef9e0, 0xc00e977b18)
/go/pkg/mod/k8s.io/apimachinery@v0.22.3/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
/go/pkg/mod/k8s.io/apimachinery@v0.22.3/pkg/util/runtime/runtime.go:48 +0x86
panic(0x41ef9e0, 0xc00e977b18)
/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/rancher/rancher/pkg/catalogv2/helm.decodeHelm3(0x0, 0x0, 0x7f4b7ab89b58, 0xc001049348, 0x6e656d6572697501)
/go/src/github.com/rancher/rancher/pkg/catalogv2/helm/helm3.go:124 +0x2e8
github.com/rancher/rancher/pkg/catalogv2/helm.fromHelm3Data(0x0, 0x0, 0xc005682ab0, 0x5, 0xc0010493c8, 0x0)
/go/src/github.com/rancher/rancher/pkg/catalogv2/helm/helm3.go:23 +0x39
github.com/rancher/rancher/pkg/catalogv2/helm.ToRelease(0x4c91e30, 0xc0036cb1c0, 0xc005682ab0, 0x646e612079656b01, 0x2e7365756c617620, 0x7463656a626f062a)
/go/src/github.com/rancher/rancher/pkg/catalogv2/helm/release.go:74 +0x590
[...]

PR #37025, opened 2022-03-24
Rancher panic if helm release secret is empty

gprossliner:
see #37023


gprossliner:
Closing, because the pull-request is based on a wrong commit

PR #37026, opened 2022-03-24
Rancher panic if helm release secret is empty

gprossliner:
see #37023


gprossliner:
I didn't find the information about PR in general, or on which branch they should be based on specifically. So I based this on the default `release/v2.6` branch.

PR #37041, opened 2022-03-25
pkg/catalogv2/git: fix dropped err

alrs:
This fixes an `err` variable that is passed into a `filepath.Walk()` but then dropped.


alrs:
The CI failure looks like a flake, could someone re-run the tests?

luthermonson:
we can probably merge this after v2.6.5 drops shortly. Thanks for the PR and I'll see if I can get CI/CD to pass.

PR #37042, opened 2022-03-25
Fix unmarsal empty .dockerconfigjson error

FrankYang0529:
If we don't set `.dockerconfigjson` in `registrySecret`, we will have error to unmarshal an empty string.

related issue: https://github.com/harvester/harvester/issues/2046


FrankYang0529:
Close PR, since the issue has been fixed in v2.6.4-rc13.

Issue #37057, opened 2022-03-25
Default Local Admin is demoted to Standard User when Auth Provider is registered


maxsokolovsky:
**Rancher Server Setup**
- Rancher version: v2.6.4-rc5
- Installation option (Docker install/Helm Chart):
   Local and Docker

**Information about the Cluster**
- Kubernetes version: 1.23 and possibly 1.22
- Cluster Type (Local/Downstream): Local

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Admin 

   
**Describe the bug**
When I import an auth provider and authenticate as an admin user from that auth provider, the default local admin user is demoted to a Standard User. Also, there are two DefaultAdmin users instead of one.

This means that users may get locked out of Rancher, if there is a problem with the auth provider, and they had logged out. Because the default admin user does not have admin permissions anymore, users cannot regain control of the cluster. 

**To Reproduce**
1. Register an auth provider, get authenticated as an admin user from the auth provider.
2. Log out.
3. Log back in as the local admin. 
4. Observe that you're suddenly a Standard User, so can't see the local cluster, and there are now two Default Admin users (this can be seen from the view of the auth provider admin user).

**Result**

**Expected Result**
 The local admin should retain admin permissions after an auth provider is registered. And there must be **one** Rancher user with two principals – local and that of the auth provider, like before.

**Screenshots**

<img width="975" alt="Screen Shot 2022-03-26 at 11 42 35" src="https://user-images.githubusercontent.com/17733533/160246892-26a2db5a-73a6-4cb3-ab2e-1438e66dcb23.png">

**Additional context**

MbolotSuse:
With a local cluster of RKE1 version `v1.20.12-rancher1-1` and `v1.22.4-rancher1-1` this is an issue. So this doesn't just affect k8s versions 1.23.

Core issue seems to be related to rancher/rancher#36525 .

[...]

PR #37066, opened 2022-03-26
Adds the hashed-principal-name to the admin user

MbolotSuse:
Addresses rancher/rancher#37057

### DIFF
--- a/pkg/data/management/ensure_admin_user.go
+++ b/pkg/data/management/ensure_admin_user.go
@@ -11,6 +11,7 @@ import (
 	"github.com/urfave/cli"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
 )
 
 func RegisterEnsureDefaultAdminCommand() {
@@ -119,6 +120,22 @@ func createNewAdmin(client v3.Interface, length int) error {
 
 	fmt.Fprintf(os.Stdout, "New default admin user (%v):\n", admin.Name)
 	fmt.Fprintf(os.Stdout, "New password for default admin user (%v):\n%s\n", admin.Name, pass)
+
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		refreshAdmin, err := client.Users("").Get(admin.Name, v1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		userKey, userValue := getAdminPrincipalLabel(admin)
+		newAdmin := refreshAdmin.DeepCopy()
+		newAdmin.ObjectMeta.Labels[userKey] = userValue
+		_, err = client.Users("").Update(newAdmin)
+		return err
+	})
+	if err != nil {
+		return err
+	}
+
 	return err
 }
 
@@ -157,10 +174,19 @@ func ensureAdminIsLabeled(admin *v3.User) bool {
 		changed = current != defaultAdminLabelValue
 	}
 
+	userKey, userValue := getAdminPrincipalLabel(admin)
+	if currentValue, ok := admin.ObjectMeta.Labels[userKey]; ok {
+		// ensure that the admin user also has the labels which are assigned to each user
+		changed = changed || (currentValue != userValue)
+	} else {
+		changed = true
+	}
+
 	if changed {
 		fmt.Fprintf(os.Stdout, "Labeling existing default admin user (%v) as admin\n", admin.Name)
 
 		admin.ObjectMeta.Labels[defaultAdminLabelKey] = defaultAdminLabelValue
+		admin.ObjectMeta.Labels[userKey] = userValue
 	} else {
 		fmt.Fprintf(os.Stdout, "Existing default admin user (%v) already labeled as admin\n", admin.Name)
 	}
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -2,6 +2,7 @@ package management
 
 import (
 	"context"
+	"encoding/base32"
 	"reflect"
 	"sort"
 	"sync"
@@ -9,6 +10,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	localauth "github.com/rancher/rancher/pkg/auth/providers/local"
 	"github.com/rancher/rancher/pkg/features"
 	"github.com/rancher/rancher/pkg/rbac"
 	"github.com/rancher/rancher/pkg/settings"
@@ -540,6 +542,10 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
 		if err != nil && !apierrors.IsAlreadyExists(err) {
 			return "", errors.Wrap(err, "can not ensure admin user exists")
 		}
+		admin, err = addPrincipalLabelToAdmin(admin, management)
+		if err != nil {
+			logrus.Warnf("cannot add principal label to admin, %s", err.Error())
+		}
 		if err == nil {
 			var serverURL string
 			if settings.ServerURL.Get() != "" {
@@ -618,6 +624,28 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
 				logrus.Info("Created default admin user and binding")
 			}
 		}
+	} else {
+		// if we have at least one user, grab the admin user to ensure that it has the Principal label
+		var adminUser *v3.User
+		for _, user := range users.Items {
+			// we could re-list using a label selector, but since we already have a full listing, just use that instead
+			if value, ok := user.Labels[defaultAdminLabelKey]; ok {
+				if value == defaultAdminLabelValue {
+					adminUser = &user
+					break
+				}
+			}
+		}
+		if adminUser == nil {
+			logrus.Warn("Have more than one user, but unable to locate admin user to adjust labels")
+		} else {
+			admin, err := addPrincipalLabelToAdmin(adminUser, management)
+			if err != nil {
+				logrus.Warnf("Unable to add label to adminUser %s", err.Error())
+			} else {
+				adminName = admin.Name
+			}
+		}
 	}
 
 	adminConfigMap := corev1.ConfigMap{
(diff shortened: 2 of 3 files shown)


pkg/data/management/role_data.go
```diff
@@ -637,6 +669,25 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
 	return adminName, nil
 }
 
+func addPrincipalLabelToAdmin(admin *v3.User, management *wrangler.Context) (*v3.User, error) {
+	// admin may not have a principal yet. To get around a complex wait loop, gen the principal name as the provider does
+	principalName := localauth.GetLocalPrincipalID(admin)
+	encodedPrincipalID := base32.HexEncoding.WithPadding(base32.NoPadding).EncodeToString([]byte(principalName))
+	if len(encodedPrincipalID) > 63 {
+		encodedPrincipalID = encodedPrincipalID[:63]
+	}
```
maxsokolovsky, line 678: Since that logging is not necessary, I'd move this inside the main if statement, so that you don't generate the encoded principal ID for nothing, if the label already exists.

MbolotSuse: Will remove the logging in next iteration.

pkg/data/management/role_data.go
```diff
@@ -618,6 +628,28 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
...
+					break
+				}
+			}
+		}
+		if adminUser == nil {
+			logrus.Errorf("Have more than one user, but unable to locate admin user to adjust labels")
+		} else {
+			admin, err := addPrincipalLabelToAdmin(adminUser, management)
+			if err != nil {
+				logrus.Errorf("Unable to add label to adminUser %s", err.Error())
```
maxsokolovsky, line 648: Would we want to also return the error here if unable to add label to Admin User, just like in the case above when there are no users?

MbolotSuse: I've actually changed the above to function the same was as this does. My reasoning is that if we fail to label the user for some reason, we would want to just create the admin in this diminished state. The admin is a bit too vital to not be created because of this, especially when we have an ensure admin script to fix manually if need be. Open to hearing your thoughts though.

pkg/data/management/role_data.go
MbolotSuse: Fixed.

pkg/data/management/role_data.go
```diff
@@ -38,6 +40,10 @@ var (
 	adminCreateLock   sync.Mutex
 )
 
+const (
+	localPrefix = "local://"
```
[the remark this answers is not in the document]
MbolotSuse: Nope - good find, removed.

pkg/data/management/ensure_admin_user.go
```diff
@@ -119,6 +119,15 @@ func createNewAdmin(client v3.Interface, length int) error {
 
 	fmt.Fprintf(os.Stdout, "New default admin user (%v):\n", admin.Name)
 	fmt.Fprintf(os.Stdout, "New password for default admin user (%v):\n%s\n", admin.Name, pass)
+
+	refreshAdmin, err := client.Users("").Get(admin.Name, v1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	userKey, userValue := getAdminPrincipalLabel(admin)
+	newAdmin := refreshAdmin.DeepCopy()
```
maxsokolovsky, line 128: Why the deep copy? In case the update fails, existing user will be unaffected, right?

MbolotSuse: Before updating a resource you do a deep copy to avoid changing the original resource (which I am told causes race conditions).

In this case, I was getting a "object has been changed try again" bug with out the deep copy and differently named var.

MbolotSuse:
## Core issue
This deals with #37057. This issue was caused by the indexer no longer being able to find the admin user by the principal id. In other cases, this code uses a label (`base32encodingOfPrincipal` : `"hashed-principal-name"`) to find the user. The admin user didn't have these labels, and so a second user was created when attempting to login as the admin user after enabling an auth provider (through local means).

## Fix Details
- After creating the admin user, adds the above label to the user
- Label is added after the user is created because the key is based on the name of the user. When creating the admin user, we let k8s choose this value for us, so we have to update the created user (otherwise we won't have the username and can't create the label)
- `ensure_admin_user` is a go file which links a binary in the rancher container that allows you to easily re-create/relabel the admin user. This had to be updated too in case a user accidentally deleted the admin user and tried to recover through the use of this script. 
- I made the local provider's method of principal naming exported so that we can re-use in the various places that need to know the principal id for the purposes of this label's creation

## Notes
- Whatever issue is causing the indexer to not find the user is still unresolved. I

pkg/data/management/ensure_admin_user.go
```diff
@@ -119,6 +120,18 @@ func createNewAdmin(client v3.Interface, length int) error {
 
 	fmt.Fprintf(os.Stdout, "New default admin user (%v):\n", admin.Name)
 	fmt.Fprintf(os.Stdout, "New password for default admin user (%v):\n%s\n", admin.Name, pass)
+
+	//sleep to avoid "object has been changed" condition
+	time.Sleep(3 * time.Second)
```
cmurphy, line 125: This is not a safe way to handle these conflict errors. Follow this pattern instead: https://github.com/rancher/rancher/blob/6f9c19dc3e1daa4880c11dee036899f4ad2b5340/pkg/controllers/management/auth/crtb_handler.go#L230-L242

That said, there is nothing happening between the user create on line 104 and the update on line 134, so the admin object returned from Create should be up to date and safe to act on. But if for some reason that's not the case, use RetryOnConflict.

pkg/data/management/role_data.go
cmurphy: This is called during startup, so if an error is returned here, it will cause the Rancher service to fail to start. IMO that is the correct behavior. As you said, the admin user is too vital. It's best to fail early and let the user handle it.

pkg/data/management/role_data.go
```diff
@@ -618,6 +624,28 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
...
+			// we could re-list using a label selector, but since we already have a full listing, just use that instead
+			if value, ok := user.Labels[defaultAdminLabelKey]; ok {
+				if value == defaultAdminLabelValue {
+					adminUser = &user
+					break
+				}
+			}
+		}
+		if adminUser == nil {
+			logrus.Errorf("Have more than one user, but unable to locate admin user to adjust labels")
```
cmurphy, line 640: If Rancher can stumble on after this, I would demote this to logrus.Warnf. If this is critical (it seems like it is?), then it should return an error here as below.

pkg/data/management/role_data.go
```diff
@@ -637,6 +665,29 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
...
 
+func addPrincipalLabelToAdmin(admin *v3.User, management *wrangler.Context) (*v3.User, error) {
+	// admin may not have a principal yet. To get around a complex wait loop, gen the principal name as the provider does
+	key, value := getAdminPrincipalLabel(admin)
+	var err error
+	if _, ok := admin.Labels[key]; !ok {
+		newAdmin := admin.DeepCopy()
+		newAdmin.Labels[key] = value
+		newAdmin, err = management.Mgmt.User().Update(newAdmin)
+		return newAdmin, err
```
cmurphy, line 676: (nitpick)
```suggestion
		return management.Mgmt.User().Update(newAdmin)
```

pkg/data/management/role_data.go
```diff
@@ -637,6 +665,29 @@ func BootstrapAdmin(management *wrangler.Context) (string, error) {
...
+	}
+	return admin, err
+}
+
+func getAdminPrincipalLabel(admin *v3.User) (string, string) {
+	// admin may not have a principal yet. To get around a complex wait loop, gen the principal name as the provider does
```
cmurphy, line 682: Why wouldn't it have it yet? If there's another component doing that, maybe this logic should be moved there?

cmurphy (changes requested):

pkg/data/management/ensure_admin_user.go
MbolotSuse: I was seeing some errors without the time.sleep - I think there's a controller somewhere that makes edits on the user right after it's created which is causing those issues. I'll migrate to the retryOnConflict method, thanks for the suggestion.

pkg/data/management/role_data.go
MbolotSuse: So in testing, `admin.PrincipalIds` was consistently nil at this point in the boot process. My guess is that the local auth provider is adding the principal when you try to login as the user but I'm not really 100% sure about that. 

That being said, given how we handle other users (adding the labels at create time) I think it makes the most sense to add the labels to this user where it is created rather than trying to stint this in later on. But if you disagree with me let me know and I can look into it further.

pkg/data/management/role_data.go
MbolotSuse: So In this case, the failure to find the admin user only inhibits our ability to add the label that we want. I think that returning an error here (effectively locking the user out on upgrade) would be a bit harsh.

pkg/data/management/role_data.go
MbolotSuse: So I've changed this to warnf for now, but I'm willing to revisit this (and the below suggestion by @cmurphy) if there's significant disagreement.

I'm thinking that while the label is not a good thing to have missing, the system is technically functional without it. Therefore, it's best to warn the user about the failure and continue to move on. I'm also not sure that erroring out here would be helpful - an error in this section would be much more indicative of a failure on our side than anything that the user can actually fix. 

Though in light of the issue which spawned this PR/discussion I'm inclined to agree now that this value is critical, so I'm open to disagreement.

pkg/data/management/ensure_admin_user.go
MbolotSuse: Fixed - thanks again for the suggestion.

pkg/data/management/role_data.go
MbolotSuse: I'll leave this open until we resolve the other discussion along these lines - but I'll keep parity between this part and the above part (i.e. they will work the same way).

pkg/data/management/role_data.go
MbolotSuse: Done.

MbolotSuse:
Closing in favor of #37083 which is more simple and comprehensive. Though we may want to consider adding this functionality at some point in the future.

PR #37096, opened 2022-03-29
Fix clustertemplate condition

cmurphy:


cmurphy:
This should fix the test_clustertemplate.py test failure but there seems to be a separate k3s issue preventing drone from finishing. Even if it doesn't fix the test this is still an important fix.

ryansann (dismissed):

paynejacob (approved):

cmurphy:
https://drone-pr.rancher.io/rancher/rancher/21607/4/2 shows this didn't fully work, added back the retries

thedadams (approved):

kinarashah (approved):

Issue #37001, opened 2022-03-22
RKE2 | Custom cluster | etcd restore leaves etcd node stuck in `Waiting for kubelet to update`


timhaneunsoo:
**Rancher Server Setup**
- Rancher version: v2.6-head 6b336a0
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.21.10
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom RKE2 cluster

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions:
Standard user is cluster owner and restore action was completed by Admin as describe [here](https://github.com/rancher/dashboard/issues/5320)

   
**Describe the bug**

After restoring snapshot, etcd node is stuck in `provisioning` with `Waiting for kubelet to update`
This is reproducible on custom clusters and also node driver clusters
```
[INFO ] provisioning bootstrap node(s) custom-e6ba3dffe8ef: waiting for plan to be applied
[INFO ] provisioning bootstrap node(s) custom-e6ba3dffe8ef: waiting for probes: kubelet
[INFO ] provisioning bootstrap node(s) custom-e6ba3dffe8ef: waiting for kubelet to update
```

After checking the etcd node, I saw that a command for kubelet was running for 1:05
```
[...]

PR #37103, opened 2022-03-29
Be more intelligent when the cluster-agent is disconnected

thedadams:
Issue: https://github.com/rancher/rancher/issues/37001

## Problem
The fundamental issue being fixed here is restoring an RKE2 cluster with 1 
etcd-only node, 1 control plane node, and 1 worker node. Previously, the etcd 
and control plane nodes would be shutdown before the restore happened. The 
cluster-agent would then be rescheduled to the worker node because it wouldn't 
be updated. This caused some issues because it looked like the agent was 
connected, but Rancher couldn't use it in any way because the api-service was 
shutdown. This would cause the restore to hang because the etcd-only node's 
kubelet version was not getting updated.

## Solution
After this change, all nodes in the cluster are shutdown before the restore 
happens. This caused the cluster agent to disconnect, as desired.

However, this caused another problem: since the CAPI clients and caches connect 
through the cluster-agent, the caches would get out of sync and the kubelet 
version was not being updated on the CAPI machines for a long time (several 
minutes). This artificially slowed down the restore on all types of RKE2 
clusters.

To fix this issue, a different client is passed to the CAPI cluster cache 
reconciler that would cause the cache to be refreshed when the cluster-agent 
disconnected.

## Testing
Provisioning, updating, and restoring an RKE2 cluster with 1 etcd-only, 1 cp, and 1 worker node.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
+++ b/pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
@@ -5,10 +5,13 @@ import (
 	"fmt"
 	"time"
 
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+	provcluster "github.com/rancher/rancher/pkg/controllers/provisioningv2/cluster"
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
 	mgmtcontrollers "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
+	provcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	rkecontrollers "github.com/rancher/rancher/pkg/generated/controllers/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/wrangler"
 	"github.com/rancher/wrangler/pkg/relatedresource"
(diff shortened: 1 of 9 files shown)


pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
```diff
@@ -32,25 +36,39 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 
 	rkecontrollers.RegisterRKEControlPlaneStatusHandler(ctx, clients.RKE.RKEControlPlane(),
 		"", "rke-control-plane", h.OnChange)
-	relatedresource.Watch(ctx, "rke-control-plane-trigger", func(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
```
thedadams, line 35: This didn't work because the management cluster and the control plane don't have the same name and namespace.

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -551,6 +548,9 @@ func (p *Planner) reconcile(controlPlane *rkev1.RKEControlPlane, tokensSecret pl
 		} else if !kubeletVersionUpToDate(controlPlane, entry.Machine) {
 			outOfSync = append(outOfSync, entry.Machine.Name)
 			messages[entry.Machine.Name] = append(messages[entry.Machine.Name], "waiting for kubelet to update")
+		} else if tierName == "control plane" && !controlPlane.Status.AgentConnected {
```
paynejacob, line 551: is there supposed to be a space in control plane?

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -950,8 +951,8 @@ func noRole(entry *planEntry) bool {
 	return !isEtcd(entry) && !isControlPlane(entry) && !isWorker(entry)
 }
 
-func anyRole(_ *planEntry) bool {
-	return true
+func anyRole(entry *planEntry) bool {
```
paynejacob, line 954: why change this?

pkg/provisioningv2/rke2/planner/planner.go
thedadams: Yes. I will update this to use `string` constants to be more understandable.

pkg/provisioningv2/rke2/planner/planner.go
thedadams: I was think that `anyRole` means that the node has a role, but we don't care.

paynejacob: When I made this the intent was for the results to be unfiltered. given the possible scenarios these will behave the same way.

paynejacob (dismissed):

Oats87 (approved):

paynejacob (approved):

Issue #37107, opened 2022-03-30
FAILED test_clustertemplate.py::test_disable_template_revision


cmurphy:
test_disable_template_revision is intermittently failing.

Example: https://drone-pr.rancher.io/rancher/rancher/21628/4/2

```
1822	E       rancher.ApiError: (ApiError(...), "ServerError : failed to set enabled flag to false on clusterTemplateRevision\n\t{'baseType': 'error', 'code': 'ServerError', 'message': 'failed to set enabled flag to false on clusterTemplateRevision', 'status': 500, 'type': 'error'}"
```

It has something to do with https://github.com/rancher/rancher/pull/37084 or https://github.com/rancher/rancher/pull/37089 interacting badly with https://github.com/rancher/rancher/blob/release/v2.6/pkg/api/norman/customization/clustertemplate/cluster_template.go#L146

PR #37111, opened 2022-03-30
[2.5] Fix clustertemplate condition and add retries

cmurphy:
Add retries to the clustertemplaterevision action handler in case it
gets caught trying to update the CTR at the same time as the
secretmigrator controller.

Backport https://github.com/rancher/rancher/pull/37096


thedadams (approved):

cmurphy:
Already merged in https://github.com/rancher/rancher/pull/37216

PR #37130, opened 2022-03-30
Automation - use SSD for the bastion/registry for airgap tests which is disk intensive

izaac:
Keep default `gp3` for everything else.


rancher-max (commented):
I know sometimes when updating volume type in AWS we have to specifically set the throughput too. If this works I'm fine with just this change, but it might help to also include throughput to set also just in case

izaac:
> I know sometimes when updating volume type in AWS we have to specifically set the throughput too. If this works I'm fine with just this change, but it might help to also include throughput to set also just in case

I understand what you're saying but that's not possible with `io2` 
Screen Shot 2022-03-30 at 5 01 16 PM

rancher-max (dismissed):
Ah perfect! In that case LGTM 😄

rancher-max (approved):
LGTM! 2 comments:
1. It might be helpful to submit a PR to jjb to allow for customizing `AWS_BASTION_INSTANCE_TYPE` in case we don't care about the speeds. (not sure what the price difference exactly is here)
2. Did you get benchmarks by any chance for this of the speed differences? Or some estimations of what you're seeing in your local testing?

izaac:
It's faster regular run took 1hr 8 min vs  1hr 34min in the regular.

slickwarren (approved):

slickwarren:
awaiting approval

Issue #37079, opened 2022-03-28
[Forwardport v2.6] Add option to disable node-cleanup jobs


rancherbot:
This is a forwardport issue for https://github.com/rancher/rancher/issues/37065, automatically created via rancherbot by @snasovich

Original issue description:

For the installations that do not require node cleanup to be done, it's beneficial to have an option to remove creation of node-cleanup job so it doesn't slow down or introduces potential failure points into node deletion process (for custom clusters).

At this time a global setting seems acceptable (i.e. it doesn't have to be controlled per cluster)

SURE-4289

thedadams:
### Root cause
Many users delete their VMs before or immediately after deleting the nodes in Rancher for a custom RKE1 cluster. This can cause issues with the `cattle-cleanup-node` job.

### What was fixed, or what changes have occurred
Therefore a feature flag is added to disable the node cleanup functionality.

### Areas or cases that should be tested
1. Start Rancher with the `--features rke1-custom-node-cleanup=false`
2. Create an RKE1 custom cluster.
3. Provision two custom nodes: one with all roles and one with worker-only role.
4. When both nodes are active, delete the worker-only node.
5. A job should NOT be created in the downstream cluster and the node should disappear from the UI more quickly than usual.

### What areas could experience regressions?
If one does not delete the VM shortly after deleting the node from Rancher, there is a chance that it will re-join the cluster because the cleanup is not run on the downstream cluster.

### Are the repro steps accurate/minimal?
Yes.

Josh-Diamond:
Ticket #37079 - Test Results

Reproduced on rancher `v2.6.3-patch2`, which may be reviewed in ticket #37065 

Verified on rancher `v2.6-6955cebc208a925e3e6f42a3a7f5d24dcd7d2585-head`:
1. Fresh install of rancher `v2.6-head` with the following flag included `--features rke1-custom-node-cleanup=false`
2. Provision an RKE1 Custom downstream cluster with AT LEAST one dedicated worker node (we will remove this node later)
3. Watch jobs on downstream cluster by running `kubectl get jobs -w`
4. Delete the dedicated worker node, mentioned in step 2
5. Verified - `cattle-node-cleanup` job is NOT called; node is deleted

Screenshots:

`Before deletion`:
[...]

Issue #37063, opened 2022-03-26
Pods can still hang when deleting cattle-node-cleanup job


snasovich:
A previous fix was implemented to try to cleanup the cattle-cleanup-node jobs and pods. That fix worked in some cases, but it is still possible for the pod to get stuck in a terminating state.

A fix should be included to force delete the pods to ensure that pods get deleted, which, in turns, allows the job to be cleaned up.

Reproduction steps:
1. Create an RKE1 custom cluster with Kubernetes version v1.21.10
2. Join one node with all roles and one node with worker-only role.
3. After all nodes are active, turn off the VM. Don't delete the node.
4. Wait for the node to become unavailable in the Rancher UI.
5. Delete the node from the Rancher UI.

Result:
The node may hang and not be cleaned up. You will see the `cattle-cleanup-node` job and a pod in the downstream cluster stuck in the deleting or terminating phase.

Expected result:
The node gets cleaned up and the job and pod are deleted from the downstream cluster.

SURE-4289

thedadams:
### Root cause
The node-cleanup job completely ignored the private registry information for a cluster.

### What was fixed, or what changes have occurred
The cattle-cleanup-node job now created with private registry information similar to the cattle-cluster-agent.

### Areas or cases that should be tested
1. Create a custom RKE1 cluster with Kubernetes version v1.21.10 (I don't know if the Kubernetes version is important, but I was able to reproduce this issue with that version).
2. Provision three custom nodes: one with all roles and two with a worker-only role.
3. Once all nodes are active, turn the VM off one of the worker nodes.
4. Once this worker node is unavailable in the Rancher UI, delete the node from the Rancher UI. It should be deleted.
5. Without turning the VM off, delete the second worker-only node. It should be deleted.

### What areas could experience regressions?
Deleting nodes from custom RKE1 clusters.

### Are the repro steps accurate/minimal?
Yes

Josh-Diamond:
Ticket #37063 - Test Results - [Duplicate of [#35148](https://github.com/rancher/rancher/issues/35148)]

With Docker on single-node instance:

Reproduction Failure on rancher `v2.6-9ef92c2168bd91ce6b05734731d9c2c21c42a5f4-head`:
1. Unable to reproduce; I have experienced this error first hand and am able to ensure quality verification provided below


Verified on rancher `v2.6-6955cebc208a925e3e6f42a3a7f5d24dcd7d2585-head`:
1. Fresh install of rancher `v2.6-head`
[...]

Issue #36981, opened 2022-03-21
cattle-node-cleanup doesn't pull from private registry


jiaqiluo:
**Rancher Server Setup**
- Rancher version: 2.6.2 & 2.6.3
- Installation option (Docker install/Helm Chart): any
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):  
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: any 
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions:  

   
**Describe the bug**

When deleting a node from a custom cluster,  a `cattle-node-cleanup` job is created and it pulls images from DockerHub and not the configured private registry for the cluster.
As the result, the job gets stuck in an airgap environment and holds a finalizer on the machine object, which can cause issues if the hardware is re-used.

**To Reproduce**
 
- Create a custom cluster of 3+ nodes and specify a private registry
- Delete one of the nodes and monitor the job that gets created to clean up the node

**Result**
A `cattle-node-cleanup` job is created and it pulls images from DockerHub and not the configured private registry for the cluster.

**Expected Result**

The job should pull images from the configured private registry for the cluster.
 
SURE-3888
SURE-4289

thedadams:
### Root cause
The node-cleanup job completely ignored the private registry information for a cluster.

### What was fixed, or what changes have occurred
The cattle-cleanup-node job now created with private registry information similar to the cattle-cluster-agent.

### Areas or cases that should be tested
1. Create a custom RKE1 cluster with Kubernetes version v1.21.10 (I don't know if the Kubernetes version is important, but I was able to reproduce this issue with that version).
2. Provision three custom nodes: one with all roles and two with a worker-only role.
3. Once all nodes are active, turn the VM off one of the worker nodes.
4. Once this worker node is unavailable in the Rancher UI, delete the node from the Rancher UI. It should be deleted.
5. Without turning the VM off, delete the second worker-only node. It should be deleted.

### What areas could experience regressions?
Deleting nodes from custom RKE1 clusters.

### Are the repro steps accurate/minimal?
Yes

timhaneunsoo:
### Test Environment: ###
[...]

Issue #37065, opened 2022-03-26
[v2.6.3-patch3] Add option to disable node-cleanup jobs


snasovich:
For the installations that do not require node cleanup to be done, it's beneficial to have an option to remove creation of node-cleanup job so it doesn't slow down or introduces potential failure points into node deletion process (for custom clusters).

At this time a global setting seems acceptable (i.e. it doesn't have to be controlled per cluster)

SURE-4289

thedadams:
### Root cause
Many users delete their VMs before or immediately after deleting the nodes in Rancher for a custom RKE1 cluster. This can cause issues with the `cattle-cleanup-node` job.

### What was fixed, or what changes have occurred
Therefore a feature flag is added to disable the node cleanup functionality.

### Areas or cases that should be tested
1. Start Rancher with the `--features rke1-custom-node-cleanup=false`
2. Create an RKE1 custom cluster.
3. Provision two custom nodes: one with all roles and one with worker-only role.
4. When both nodes are active, delete the worker-only node.
5. A job should NOT be created in the downstream cluster and the node should disappear from the UI more quickly than usual.

### What areas could experience regressions?
If one does not delete the VM shortly after deleting the node from Rancher, there is a chance that it will re-join the cluster because the cleanup is not run on the downstream cluster.

### Are the repro steps accurate/minimal?
Yes.

Josh-Diamond:
Ticket #37065 - Test Results

With Docker on single-node instance:

Reproduced on rancher `v2.6.3-patch2`:
1. Fresh install of rancher `v2.6.3-patch2`
2. Provision an RKE1 Custom downstream cluster with AT LEAST one dedicated worker node (we will remove this node later)
3. Watch jobs on downstream cluster by running `kubectl get jobs -w`
4. Delete the dedicated worker node, mentioned in step 2
5. Reproduced - `cattle-node-cleanup` job is called; node is deleted

Screenshot:
<img width="435" alt="cleanupjobREPRODUCED" src="https://user-images.githubusercontent.com/46494969/161648167-58b181e2-657b-45eb-8630-264ab5133359.png">


Verified on rancher `v2.6.3-patch3-rc1`:
1. Fresh install of rancher `v2.6.3-patch3-rc1` with the following flag included `--features rke1-custom-node-cleanup=false`
2. Provision an RKE1 Custom downstream cluster with AT LEAST one dedicated worker node (we will remove this node later)
[...]

PR #37142, opened 2022-03-31
Private registry cleanup job

thedadams:
Issues:
https://github.com/rancher/rancher/issues/36981
https://github.com/rancher/rancher/issues/37079
https://github.com/rancher/rancher/issues/37063

## Problem
Add feature flag for custom RKE1 node cleanup

It is common for users to cleanup their own custom nodes when deleting
them from Rancher. This has caused issue with jobs and/or pods being
left around or nodes not being deleted properly.

## Solution
This change adds a feature flag to turn the cleanup off globally if a
user doesn't need it.

## Problem
Refactor node-cleanup job implementation

It was determined that some of the code around creating and deleting the
node-cleanup job was producing unexpected results like non-nil errors
with empty error messages. In addition, the cleanup job is being
leftover in some case.

## Solution
This code makes the error messages more descriptive and adds retries to
try to ensure that the cleanup job gets deleted.

## Problem
Add private registry information to node cleanup job

The node cleanup job ignored cluster private registry information. This caused
issues when trying to remove nodes from a custom RKE1 cluster when private
registries were enforced by something like OPA Gatekeeper.

## Solution
Now, just as with cluster agent deployment, private registry information is
added whenever needed.

## Problem
If a pod is stuck because it has been scheduled on a node that node
longer exists, then deleting the job may not be enough to delete this
pod. This would cause the job deletion to hang forever.

## Solution
This change deletes the pod with gradePeriodSeconds of 0 to force its
deletion.

### DIFF
--- a/pkg/controllers/management/node/cleanup.go
+++ b/pkg/controllers/management/node/cleanup.go
@@ -30,6 +30,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+const cleanupPodLabel = "rke.cattle.io/cleanup-node"
+
 func (m *Lifecycle) deleteV1Node(node *v3.Node) (runtime.Object, error) {
 	logrus.Debugf("Deleting v1.node for [%v] node", node.Status.NodeName)
 	if nodehelper.IgnoreNode(node.Status.NodeName, node.Status.NodeLabels) {
@@ -171,7 +173,7 @@ func (m *Lifecycle) cleanRKENode(node *v3.Node) error {
 		return err
 	}
 
-	return m.waitUntilJobDeletes(userContext, job)
+	return m.waitUntilJobDeletes(userContext, node.Name, job)
 }
 
 func (m *Lifecycle) waitForJobCondition(userContext *config.UserContext, job *v1.Job, condition func(*v1.Job, error) bool, logMessage string) error {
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/node/cleanup.go
```diff
@@ -185,64 +198,77 @@ func (m *Lifecycle) waitUntilJobCompletes(userContext *config.UserContext, job *
...
 		}
+
+		return nil, m.waitUntilJobDeletes(userContext, &existingJob.Items[0])
 	}
 
 	meta := metav1.ObjectMeta{
-		GenerateName: "cattle-node-cleanup-",
-		Namespace:    "default",
+		GenerateName: "cattle-cleanup-node-",
+		Namespace:    namespace.System,
```
luthermonson, line 265: i think we should leave this in default for "reasons". there are more cleanup jobs in default that would need to be changed if you stick with this choice.

luthermonson (dismissed):

snasovich (approved):

paynejacob (dismissed):

jakefhyde (approved):

luthermonson (approved):

martyav:
@thedadams I'm a technical writer on the Rancher team -- a user requested documentation for `rke1-custom-node-cleanup` as it was accidentally left out of https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/feature-flags#docusaurus_skipToContent_fallback. Presumably the default value is `true` as it's active on a fresh Rancher install of v2.7.2. Is it considered general availability or experimental?

snasovich:
@martyav , you're correct that this flag defaults to `true` and it was added as part of the linked issue https://github.com/rancher/rancher/issues/37065 that has the rationale on why it got added. Right around the same time we've made multiple fixes to improve the stability of node cleanup jobs and this setting was viewed as "plan B" in case the fixes don't work on some configurations (so far the fixes has been good).

So, the setting does exist and I would not consider it "experimental" per se - just one that we should only recommend changing if the user understands the consequences of disabling it. The node cleanup job was introduced to prevent situations where removed nodes may attempt to rejoin the cluster after being booted up again.

Hope this helps.

cc: @thedadams

martyav:
@snasovich Thank you, was able to update and do a general refresh on the page

Issue #36959, opened 2022-03-18
AKS Basic LoadBalancer can't be used


annablender:
**Rancher Server Setup**
- Rancher version: 2.6.2
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version:
- Cluster Type (Local/Downstream): Local and downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Local cluster 3 nodes, downstream cluster AKS 1.20 provisioned by Rancher. Basic LB selected and no availability zones selected for the node group.
  

**Describe the bug**

AKS clusters cannot be created with the Basic LB sku on 2.6.2. Abbot has noticed that the LB sku dropdown is greyed out until he unclicks all of the availability zones from the first node group, adds a node group, then removes the first again. At this point, it lets him select Basic. Continuing on results in this error:

```
error failed to create cluster: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 – Original Error: Code="SLBRequiredForAvailabilityZone" Message="Standard Load Balancer is required for availability zone." Target="networkProfile.loadBalancerSku"
```

Abbot cannot create clusters without a public IP in front of the LB, apparently Standard ones will always do that.

**To Reproduce**
 

_None_

**Troubleshooting**

We tried editing the cluster yaml after the fact to make everything reflect, but that did not change anything.

SURE-3685

annablender:
We think that what's happening here is an empty slice of Availability Zones is being passed to AKS instead of nil, so AKS thinks there are availability zones even though there aren't. This would cause the error because if availability zones are specified, you have to use a Standard Load Balancer instead of a basic one.

Josh-Diamond:
Ticket #36959 - Test Results

With Docker on single-node instance:

Reproduced on rancher `v2.6.2`:
1. Fresh install of rancher `v2.6.2`
2. Provision a downstream AKS cluster with k8s `v1.20.15` and a `Basic` LB; [un-click availability zones; add node group, delete node group, select LB as `Basic`]
3. Reproduced - Failure to provision downstream AKS cluster

Screenshot:
[...]

PR #37146, opened 2022-03-31
[2.6.5] Fix AKS basic load balancer

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/36959

# Problem
Users were unable to provision AKS clusters using a basic load balancer. When creating an AKS cluster with a basic load balancer, Rancher was setting the node pool `AvailabilityZones` field to an empty list (as availability zones cannot be used with a basic load balancer). However, this caused Azure to mistakenly believe that we _had_ provided some availability zones, preventing the use of a basic load balancer.
 
# Solution
This problem has been resolved by ensuring that any empty `AvailabilityZones` slices are set to `nil`. 
 
# Testing
 I've created several AKS clusters using a basic load balancer and they have provisioned correctly and transitioned into an `active` state without error.

### DIFF
--- a/pkg/controllers/management/aks/aks_cluster_handler.go
+++ b/pkg/controllers/management/aks/aks_cluster_handler.go
@@ -108,7 +108,8 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		}
 
 		// ensure empty availability zones are set to nil
-		if pruneAvailabilityZones(cluster) {
+		cluster, requiresUpdate := pruneAvailabilityZones(cluster)
+		if requiresUpdate {
 			return e.ClusterClient.Update(cluster)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -107,6 +107,12 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			return cluster, err
 		}
 
+		// ensure empty availability zones are set to nil
+		cluster, requiresUpdate := pruneAvailabilityZones(cluster)
+		if requiresUpdate {
```
jakefhyde, line 112: Is there any reason we don't continue here? We could instead:


```
cluster, err = e.clusterClient.Update(cluster)
if err != nil {
  return cluster, err
}
```
but not sure if this choice was intentional

jakefhyde, line 112: Is there any reason we don't continue here? We could instead:


```
cluster, err = e.clusterClient.Update(cluster)
if err != nil {
  return cluster, err
}
```
but not sure if this choice was intentional

HarrisonWAffel: I noticed a pattern throughout the rest of the file wherein any call to `e.ClusterClient.Update` was returned from the controller. I stuck with that pattern to avoid any expected errors. A good example would be lines `87`-`91`, where only the cluster name / driver is updated but the Update call is still returned.

HarrisonWAffel:
After speaking with @thedadams this change is better made within the AKS-operator itself, so I will be closing with PR and raising a new one in that repository

PR #37148, opened 2022-04-01
update url

clamyang:


luthermonson (approved):
thanks for the updated url

luthermonson:
will merge after v2.6.5 releases

Oats87 (approved):

Issue #37088, opened 2022-03-28
Pin KDM data for 2.6.3 patch releases 


kinarashah:
**Problem** 
--- 
KDM was just released for Rancher v2.6.4. If 2.6.3-patch3 is tagged now, it'll build with the latest KDM data (from release-v2.6 branch) - https://github.com/rancher/rancher/blob/v2.6.3-patch2/scripts/build-server#L24 (`CATTLE_KDM_BRANCH` is set to release-v2.6 here https://github.com/rancher/rancher/blob/v2.6.3-patch2/Dockerfile.dapper#L15) 

This means Rancher has the latest copy of KDM data as local data, but vendored RKE (https://github.com/rancher/rancher/blob/v2.6.3-patch2/go.mod#L121) doesn't have the new released data for v2.6.4. Rancher implicitly assumes the build time [local data is the same as the vendored data ](https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/kontainerdrivermetadata/data.go#L69) and sets the [wrong label value on addon CRDs](https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/kontainerdrivermetadata/data.go#L357). 

As a result, when on provisioning a new k8s version released out of band for >=2.6.3 like v1.21.10 - it works as expected for Rancher v2.6.3 but not on Rancher v2.6.3-patch2 because it doesn't use the out of band templates. 

**Solutions** 
---
1) Vendor the latest RKE for Rancher v2.6.3-patch3. This ensures the assumption Rancher is making (local data is same as vendored RKE data). 
2) Pin the KDM data during build time so Rancher v2.6.3-patch3 behavior will be the same as v2.6.3-patch2. 
3) Fix Rancher build time logic to store RKE vendored data in the image and Rancher reads from this data to diff check between RKE and local data. 

**Fix** 
---
For Rancher v2.6.3-patch3, we'll go with approach 2. This is the safest option for now and also ensures the behavior stays same as v2.6.3-patch2. Pin [KDM URL](https://github.com/rancher/rancher/blob/v2.6.3-patch2/scripts/build-server#L24) to the specific commit v2.6.3-patch2 was built with. 

Since pinning to a specific commit requires using `https://raw.githubusercontent.com/...` we should create a new branch in KDM, and use that data.json URL which would be using `https://releases.rancher.com/kontainer-driver-metadata/...`. This is safer in case GH goes down and aligns well with our other release artifacts.

[...]

PR #37149, opened 2022-04-01
pin kdm data for v2.6.3 patch releases

kinarashah:
**Problem**
------ 
Local data in rancher needs to be consistent with vendored RKE1 data. Refer to https://github.com/rancher/rancher/issues/37088 for more details. This is handled by our release process for normal release tags but not patch releases which are tagged adhoc. 

**Solution** 
------ 
Pinning KDM data for patch releases. Setting it to v1.3.7, because patch2 uses v1.3.7 and we want the behavior to remain the same as v2.6.3-patch2. Special branch `release-rke-v1.3.7` created here: https://github.com/rancher/kontainer-driver-metadata/pull/879

### DIFF
--- a/scripts/build-server
+++ b/scripts/build-server
@@ -21,7 +21,8 @@ CGO_ENABLED=0 go build -tags k8s \
   -o bin/rancher
 
 if  [ -n "$CATTLE_KDM_BRANCH" ]; then
-    curl -sLf https://releases.rancher.com/kontainer-driver-metadata/${CATTLE_KDM_BRANCH}/data.json > bin/data.json
+    # bypass KDM_BRANCH for patch releases, pinning local data to vendored RKE1 v1.3.7 data
+    curl -sLf https://releases.rancher.com/kontainer-driver-metadata/release-rke-v1.3.7/data.json > bin/data.json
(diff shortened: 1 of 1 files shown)


scripts/build-server
```diff
@@ -21,7 +21,8 @@ CGO_ENABLED=0 go build -tags k8s \
   -o bin/rancher
 
 if  [ -n "$CATTLE_KDM_BRANCH" ]; then
```
superseb, line 23: Shouldn't we be setting `CATTLE_KDM_BRANCH` so there is no confusing what is used? This is still set in multiple places to a different value.

kinarashah: We still want Rancher to start with branch `release-v2.6` in rke-metadata-settings, this is set by our setting logic by reading `CATTLE_KDM_BRANCH` so I didn't change it. 
https://github.com/rancher/rancher/blob/release/v2.6/pkg/settings/setting.go#L244 
https://github.com/rancher/rancher/blob/release/v2.6/pkg/settings/setting.go#L250

superseb (approved):

paynejacob (approved):

Issue #36941, opened 2022-03-17
[RKE2 s3 snapshot] 2 copies of snapshot are created and snapshots are not backed up in S3 bucket


timhaneunsoo:
**Rancher Server Setup**
- Rancher version: v2.6-head
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.21.10+rke2r2
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): AWS EC2

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions:
Admin

   
**Describe the bug**

Snapshot action seems to take two copies of the snapshot, with one being empty. Non of the snapshots are begin backed up in S3 bucket

**To Reproduce**
 
Create RKE2 Cluster
- in etcd configuration, set `Backup Snapshots to S3` to `Enable` and fill out S3 bucket information

**Result**
When taking snapshot, nothing is uploaded to the S3 bucket

**Expected Result**

A backup of the snapshot should be uploaded to the S3 bucket

**Screenshots**

![image](https://user-images.githubusercontent.com/97477402/158838888-fe901d08-09ce-4061-978d-3513d25c5891.png)
![image](https://user-images.githubusercontent.com/97477402/158838922-c77cbbf4-56a7-410b-a2a6-d465c3b404cd.png)

**Additional Context**
Not sure if this is relevant but when I go to `Edit Config` and go to etcd configuration where the fields for the S3 bucket are, the `Backup Snapshots to S3` is set to `Enable` but the rest of the fields are empty
![image](https://user-images.githubusercontent.com/97477402/158839623-40ce2067-3c26-46c0-b1f9-d4f57f57daa3.png)

snasovich:
For snapshots to be stored in S3 we require deeper changes; that's why moving to 2.6.5.

Oats87:
Analyzing this issue, there seems to be a few things causing issues.

Regarding two snapshots being created, this is expected behavior when using S3 snapshots. This is because we have a reference to both a local snapshot and the snapshot in S3. We should likely have a dashboard enhancement to display the snapshot storage location i.e. local vs. S3.

[...]

Issue #36711, opened 2022-03-02
Redundant error in rancher logs `error while creating etcd snapshot`


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `e93a53c`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: `v1.21.10+rke2r1`
- Cluster Type (Local/Downstream): Downstream DO Node driver cluster

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) admin

   
**Describe the bug**

- Deploy an rke2 do cluster - 3 etcd, 3 worker and 1 cp nodes
- Enable recurring snapshot. - retention = 5 and interval = * * * * * (every minute)
- Error constantly seen in Rancher logs:

```
2022/03/02 01:40:13 [ERROR] error syncing 'kube-system/rke2-etcd-snapshots': handler snapshotbackpopulate: rkecluster fleet-default/sowmya-etcd: error while creating etcd snapshot fleet-default/sowmya-etcd-etcd-snapshot-sowmya-etcd-pool1-80a27032-497l-6b1a7: etcdsnapshots.rke.cattle.io "sowmya-etcd-etcd-snapshot-sowmya-etcd-pool1-80a27032-497l-6b1a7" already exists, requeuing
2022/03/02 01:40:14 [INFO] [snapshotbackpopulate] rkecluster fleet-default/sowmya-etcd-2: processing configmap kube-system/rke2-etcd-snapshots
2022/03/02 01:40:14 [ERROR] error syncing 'kube-system/rke2-etcd-snapshots': handler snapshotbackpopulate: rkecluster fleet-default/sowmya-etcd-2: error while creating etcd snapshot fleet-default/sowmya-etcd-2-on-demand-sowmya-etcd-2-pool1-b66908a4-tf9x-f95b7: etcdsnapshots.rke.cattle.io "sowmya-etcd-2-on-demand-sowmya-etcd-2-pool1-b66908a4-tf9x-f95b7" already exists, requeuing
2022/03/02 01:40:27 [INFO] [snapshotbackpopulate] rkecluster fleet-default/sowmya-etcd: processing configmap kube-system/rke2-etcd-snapshots
[...]

PR #37160, opened 2022-04-01
Fix S3 snapshot integration for v2prov framework

Oats87:
This PR addresses the fact that S3 snapshots were not working in the v2prov framework. 

There are a couple of enhancements in this PR:
* Make S3 snapshots actually get taken -- the primary issue here was that the cloud credentials were not being rendered properly into the config file as there was a change made to always render cloud credentials into an environment variable. The issue is that the config file generation logic ignored environment variables, and thus, the secret key would never make it into the config file for RKE2. Additionally, one-time snapshot creation relied on the credentials provided by the config file, and as such, neither recurring or one-time snapshot S3 snapshots could not be taken.
* Add logic to allow for proper clean up of `etcdsnapshots.rke.cattle.io` objects. K3s/RKE2 only list the configured remote backend when `(k3s|rke2) etcd-snapshot list` is invoked, and when S3 mode is enabled, only S3 snapshots were being cleaned up. This PR adds functionality to run `(k3s|rke2) etcd-snapshot list` twice, once to look at local snapshots, and another to look at S3 backed snapshots.
* Adds feedback for one-time snapshot creation, to list which nodes/machines an etcd snapshot is being taken on. This directly addresses the issue 36711

Notably, this does not fully solve 36941, as there is a dashboard enhancement that is tied into that issue.

https://github.com/rancher/rancher/issues/36941
https://github.com/rancher/rancher/issues/36711


paynejacob (approved):

jakefhyde (approved):

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -873,10 +871,16 @@ func (p *Planner) desiredPlan(controlPlane *rkev1.RKEControlPlane, tokensSecret
...
+		nodePlan, err = p.addEtcdSnapshotListLocalPeriodicInstruction(nodePlan, controlPlane)
 		if err != nil {
 			return nodePlan, err
 		}
+		if controlPlane != nil && controlPlane.Spec.ETCD != nil && controlPlane.Spec.ETCD.S3 != nil {
+			nodePlan, err = p.addEtcdSnapshotListS3PeriodicInstruction(nodePlan, controlPlane)
+			if err != nil {
+				return nodePlan, err
+			}
+		}
```
thedadams, line 883: If s3 snapshots are enabled for a cluster, then we will essentially be doing these allocations twice and error checks twice every time.

I like the idea of doing an `if-else` statement here to avoid that.

thedadams (approved):
Two nits. Otherwise, looks good.

Issue #36948, opened 2022-03-17
[RKE2 snapshots] age of existing snapshots as new ones are taken after scaling up etcd nodes


timhaneunsoo:
**Rancher Server Setup**
- Rancher version: v2.6-head d604b7e
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.23.4+rke2r2
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): EC2 aws node driver

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions:
Cluster Owner

   
**Describe the bug**

The age of existing snapshots changes as new ones are taken after scaling up etcd nodes.
The issue [here](https://github.com/rancher/rancher/issues/36697) is found after scaling up etcd nodes [here](https://github.com/rancher/rancher/issues/36794)

**To Reproduce**
 
1. create the RKE2 cluster with 1 etcd node, 2 control plane, 2 workers
2. makes a snapshot
3. restore that snapshot
4. scale etcd nodes to 3
5. make some new snapshots

**Result**
The snapshot taken in step 2 will change age along with the new snapshots made in step 5

**Expected Result**

The age of the snapshot should be preserved

**Screenshots**

![image](https://user-images.githubusercontent.com/97477402/158882051-c6f27f4b-752d-4f70-9b0d-851ed2bd4e92.png)

snasovich:
@Oats87 , could you take a look into it? Is this Rancher issue or RKE2 one?

PR #37195, opened 2022-04-05
Remove usage of snapshotbackpopulate tweaklist for reliability

Oats87:
This should possibly help to resolve https://github.com/rancher/rancher/issues/36948 and other weird issues we are having with the snapshotbackpopulate controller.

Also, will stop the duplicate snapshot create error message for https://github.com/rancher/rancher/issues/36711

### DIFF
--- a/pkg/controllers/managementuser/controllers.go
+++ b/pkg/controllers/managementuser/controllers.go
@@ -38,9 +38,7 @@ func Register(ctx context.Context, mgmt *config.ScaledContext, cluster *config.U
 	windows.Register(ctx, clusterRec, cluster)
 	nsserviceaccount.Register(ctx, cluster)
 	if features.RKE2.Enabled() {
-		if err := snapshotbackpopulate.Register(ctx, cluster); err != nil {
-			return err
-		}
+		snapshotbackpopulate.Register(ctx, cluster)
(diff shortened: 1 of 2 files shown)


pkg/controllers/managementuser/controllers.go
```diff
@@ -38,9 +38,7 @@ func Register(ctx context.Context, mgmt *config.ScaledContext, cluster *config.U
 	windows.Register(ctx, clusterRec, cluster)
 	nsserviceaccount.Register(ctx, cluster)
 	if features.RKE2.Enabled() {
-		if err := snapshotbackpopulate.Register(ctx, cluster); err != nil {
-			return err
-		}
+		snapshotbackpopulate.Register(ctx, cluster)
```
briandowns, line 41: Would there be any value in logging the error?

Oats87: I don't think there are any errors that are potentially thrown downstream of the `Register` function that would bubble up to this point. We originally had the `err` allocation/check because we were creating a custom controller using a shared cache, but as we are not doing this anymore I removed this

thedadams (approved):

briandowns (approved):

paynejacob (approved):

Issue #36702, opened 2022-03-01
Rancher slowness, High Mem and Cpu on pods


samjustus:
SURE-4188

**Rancher Server Setup**
Rancher Cluster:
Rancher version: v2.6.2
Number of nodes: 3
Node OS version: RHEL 7.9

Downstream Cluster:
Number of Downstream clusters: 3
Node OS: RHEL
RKE/RKE2/K3S version: RKE2 v1.21.5+rke2r2
Kubernetes version: v1.2
   
**Describe the bug**
Rancher performance is slow. Login sometimes work, sometimes not.  Rancher cluster consists of three nodes (2 cpus, 24gb ram). Rancher pods were consuming a lot of cpu and memory. Deleting the pods didn't solve the problem . After scaling the cpus to 4 and memory to 48GB (required a reboot) the problem was gone after a while but is back. Occured 2 times so far

**To Reproduce**
n/a

**Result**
1000s of will not create impersonation account on cluster in Rancher pods. See 45758 rolebindings which is abnormal.

**Expected Result**
No slowness in Rancher and no warning logs.

**Additional context**
Rancher nodes are way above hardware requirements (https://rancher.com/docs/rancher/v2.6/en/installation/requirements/#cpu-and-memory).

see logs in jira issue

cmurphy:
Created subissues for this topic:

https://github.com/rancher/rancher/issues/39215
https://github.com/rancher/rancher/issues/39216
https://github.com/rancher/rancher/issues/39217
https://github.com/rancher/rancher/issues/39218
https://github.com/rancher/rancher/issues/39219

samjustus:
closing and will track via above issues

PR #37197, opened 2022-04-05
Removing empty values from role template rules

MbolotSuse:
Role template rules can have emtpy slices in them, unlike the rules used by normal rbac objects (roles, cluster roles). We remove these empty slices in order to ensure a fair comparison with the rbac objects later on.

### DIFF
--- a/pkg/controllers/managementuser/rbac/handler_base.go
+++ b/pkg/controllers/managementuser/rbac/handler_base.go
@@ -2,6 +2,7 @@ package rbac
 
 import (
 	"context"
+	"encoding/json"
 	"reflect"
 	"strings"
 
@@ -336,6 +337,13 @@ func ToLowerRoleTemplates(roleTemplates map[string]*v3.RoleTemplate) {
 				verbs = append(verbs, strings.ToLower(v))
 			}
 			rule.Verbs = verbs
+			newRule, err := removeEmptyFields(rule)
+			if err == nil {
+				rule = newRule
+			} else {
+				// Removal of empty fields is not a critical process - we should continue even if this fails
+				logrus.Warnf("failed to remove empty fields for rules on %s, with error %v", rt.Name, err)
+			}
 			toLowerRules = append(toLowerRules, *rule)
 		}
 		rt.Rules = toLowerRules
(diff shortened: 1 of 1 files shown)


MbolotSuse:
Related to, but does not necessarily resolve, #36702.

## Background
When we create a Role Template, we create Roles/Cluster roles which actually grant the permissions defined in the Role Template. We also have handlers which check to ensure that the downstream roles/cluster roles have the same rules as defined in the Role Template. This way, if you change the Role Template, these permissions are applied to active users (who have an assignment to these Roles). 

## Core Issue
K8s [RBAC types](https://github.com/kubernetes/api/blob/master/rbac/v1/types.go#L49-L69) specify `omitempty` tags on various fields for RBAC rules (APIGroups, Resources, ResourceNames, NonResourceUrls). This causes the value to be `nil` - even if an empty string slice (of length 0) is specified on input. However, we don't appear to define this in the CRDS that we specify in k8s (we do specify inside of rancher, but not inside of k8s). Because of this, when Role Templates are created with empty slices (rather than not-specifyed nil values), our controller sees the downstream roles as always having different rules than the upstream role and is unable to update them to have the same rules. This causes a massive re-sync effort every 10 hours when lasso forces a refresh.

## Solution
The 'minimal code' solution to this is simply to remove the offending fields from the Rules defined in the RoleTemplates. This doesn't change the RoleTemplate object in the K8s api, but just allows us to us `reflect.DeepEqual` to compare the role template rules and the rules that are already present in the downstream RBAC objects.

## Notes
- I would have preferred to have the objects well structured or properly created in the first place. This would have likely required some form of fix script or more complex solution, so I defaulted to what was simple

ryansann (approved):
Discussed this in depth with @MbolotSuse offline. It's worth noting that there could be other places where this issue is happening. There's also probably a bigger tech debt item around passing schemas to CRDs. Both of those are out of scope for this issue and these changes fix the unnecessary resyncs for empty rule fields so it LGTM

maxsokolovsky (approved):
LGTM! One small nitpick is not about your code, but lines directly above it - where resources and verbs are lower-cased. I am not sure why they are not just changed in place. There are these unnecessary new slice allocations, but it's super-minor.

PR #37198, opened 2022-04-05
Enable commitid installs for rke2 and sets rhel user/pass for only rhel7

rancher-max:


VestigeJ (dismissed):
LGTM - it's fun to see GoLang style nil checking in bash (obviously a good practice)

rancher-max:
Made an update to include the `INSTALL_RKE2_METHOD` variable as this is required for rhel-like systems (per the design discussion with the team yesterday). Confirmed it is working in local runs. This is ready for review again.

ShylajaDevadiga (approved):
LGTM

Issue #36757, opened 2022-03-04
Non-admin user does not have backupEtcd as an available action which prevents snapshot / restore snapshot


neillsom:
Related to https://github.com/rancher/dashboard/issues/5011

The issue appears to be that a standard user does not have 'backupEtcd' as an available k8s action

**As admin**
![](https://zube.io/files/rancher/f9cea1372cc29e576695c101de014a9e-admin-with-action.jpg)

**As standard user**
![](https://zube.io/files/rancher/63a0327cfb36602624c0c9b1a961b27a-standarduser-without-action.jpg)

**Non-admin user attempting to take snapshot**
![](https://zube.io/files/rancher/9a630f1bcc5c84a09281a70ccdee3fc2-403-error.jpg)

The UI team can bypass the logic to enable a standard user to have the UI 'take snapshot' and 'restore snapshot' functionality, but these will result in a 403 errors

gaktive:
Based on the description, the 5035 link should point to Dashboard Zube, which is in fact https://github.com/rancher/dashboard/issues/5011

MbolotSuse:
@neillsom What role does the user have on the cluster (i.e. did the user create the cluster and is the cluster owner, or where they assigned a different role after)?

nwmac:
@MbolotSuse This works correctly for RKE2 cflusters, but not for RKE1 - if you try the same again - create a standard user and make them an owner of an RKE1 cluster, they are unable to backup or restore snapshots - they API does not provide the `backupEtcd` action metadata.

MbolotSuse:
@nwmac Just tried this with an RKE1 cluster (v1.23.4) in digital ocean. Same results as above - I'm able to see the snapshots and take a new snapshot after clicking into the cluster. The backend, in this case, appears to be functioning as intended.

Would you mind providing more clarity on what specifically is not working (as well as what the expected result is)?

MbolotSuse:
@nwmac That specific endpoint/property really helps - thank you for providing that detail. I'll take a look into this and see if I can figure out what is going on.

samjustus:
SURE-4359

MbolotSuse:
Validation Template

## Root Cause
The `canBackup` etcd function was always using the `apiContext.ID` as the `namespaceId` to check access to the `etcdbackup` resources. This was nil in some cases, causing a different access check than intended (this function should have checked access to the `etcdbackup` resources in the cluster namespace - e.x. `c-1234`). 

[...]

PR #37208, opened 2022-04-06
Fixing two bugs to allow the etcdbackup action to properly show

MbolotSuse:
Fixed a bug which caused access to be evaluated for the wrong namespace
Fixed a bug which caused a nil resource when checking for permissions/access.


MbolotSuse:
Related to #36757

ryansann (approved):

Issue #36432, opened 2022-02-08
Unable to execute shell for windows pods as admin user


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.6.4-rc2
- Installation option (Docker install/Helm Chart):helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):rke1
- Proxy/Cert Details:valid certs

**Information about the Cluster**
- Kubernetes version: 1.22.6+rke2r1
- Cluster Type (Local/Downstream):downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): custom with windows nodes (version 2022)

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) admin user

   
**Describe the bug**

users are unable to execute shell through rancher on windows pods

**To Reproduce**
 
* on a cluster with windows nodes, deploy a windows workload
* exec into the pod for that windows workload

**Result**
user is unable to execute shell for the windows pod

**Expected Result**
we should be able to exec into windows workloads

**Additional context**
full pod logs -- tbd

rosskirkpat:
Further debugging of this issue is blocked due to:
https://github.com/rancher/rancher/issues/36512
https://github.com/rancher/loglevel/issues/11

Without the ability to set Rancher logging to debug/trace, we are unable to view the relevant errors (or any pertinent log lines) relating to executing a rancher shell for a Windows pod.

sirredbeard:
Does not work with Windows Firewall turned on.

samjustus:
@sirredbeard mike is out - can we move this back to-test?

sirredbeard:
Yes. @samjustus

slickwarren:
tested on v2.6-head (228a585):
* deploy windows cluster (v1.22.8+rke2r1) using windows 2022
* deploy a windows workload
* as admin, shell into the workload -- pass
* as admin, execute commands in the shell (cmd, powershell) -- pass
*  as standard user, cluster owner shell into the workload -- pass
* as standard user, cluster owner execute commands in the shell (cmd, powershell) -- pass
* as base user, project owner shell into the workload -- pass
* as base user, project owner execute commands in the shell (cmd, powershell) -- pass

samjustus:
SURE-4589

PR #37214, opened 2022-04-06
Adding read-nodes permission to project owner

MbolotSuse:


MbolotSuse:
Requested by the windows team to help resolve #36432 .

When attempting to exec into a pod, the UI now needs to know if that pod represents a windows container or a linux container. Presumably this is because it needs to know which shell command (`cmd` vs `sh`) to use to exec into the container. @phillipsj correct me if I am wrong there. 

It attempts to do this by looking at the node. Project owners, in the current state, can't see nodes, and as such can't tell if this is a windows/linux container. This adds the permissions to the project owner to see the node, and adds the `node` resource to the list of objects that is created in the `project-owner-promoted` cluster role (i.e. permissions granted by a cluster role binding which represents non-namespaced resources).

The security concern here would be if we want, in the long term, project owners to be able to see nodes.

phillipsj (approved):
I was close. LGTM!

rmweir (changes requested):
Because of this are project owners privy to the nodes ssh keys or can now do a node exec?

MbolotSuse:
@rmweir 
> Because of this are project owners privy to the nodes ssh keys or can now do a node exec?

I'm not 100% sure what you mean here. 

I think that you were thinking of `node` objects in the `management.cattle.io` api group. This change does not grant access to those resources.

This grants access to node objects in the core api group (i.e. the default node object in k8s). I don't see a ssh key in the [api spec](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/) or in an example object. I also don't see how you would exec into a node without the ssh key/ permission to deploy a daemonset - if you could elaborate there I would appreciate it.

rmweir (approved):

MbolotSuse:
@thedadams One of the questions that came up here was surrounding potential sensitive data in the node objects (keep in mind these are nodes from k8s core api group, not our CRD defined nodes). Would you happen to see any potential issues with giving read (`get`, `list`, `watch`) permissions to project-owners for nodes of clusters that they own projects in?

thedadams (approved):

phillipsj:
@MbolotSuse I just tested the following and it all now works as expected. I created a project and namespace deploying the 
daemonset we have [here](https://github.com/rancher/windows/blob/main/manifests/workloads/daemonset.yaml). I was able to exec into both OSes with the following account types. 

- exec shell on linux workload:
  - as admin -- pass
  - as standard user, cluster owner -- pass
  - as standard user, project owner -- pass
- exec shell on windows workload:
  - as admin -- pass
  - as standard user, cluster owner -- pass
  - as standard user, project owner -- pass

This a screenshot while being a standard user that is a project owner. The failed node call is now fixed and you can exec into a Windows pod just fine.

image

MbolotSuse:
@phillipsj Ok - looks like we are good on our side from a permissions perspective. I need to run a couple of checks to make sure that this doesn't have any negative side effects, but once I make that confirmation we should be clear to merge this in and close out your issue.

The other thing that we will need to do here is have to docs updated so that people know they need read permissions on nodes if they want to exec into windows pods on those nodes. Not everyone uses the standard RBAC roles so this will need to be published so that people who roll their own RBAC setups can deal with this too.

Issue #25557, opened 2020-02-25
Multiple AD Hosts


axeal:
**What kind of request is this (question/bug/enhancement/feature request):**

Enhancement

**Steps to reproduce (least amount of steps as possible):**

Similar to the request https://github.com/rancher/rancher/issues/21382 for LDAP, the user would like to be able to configure a list of multiple AD servers for authentication, such that auth requests failover in the event of an AD server being unavailable.

**Environment information**
- Rancher v2.3

bemeyert:
We could use this feature too.

ZeusThaBomb:
We could use this feature too.

sevenval-admins:
We would be very interested to this feature too.
Can anyone please give us some feedback?

dbrekau:
Are there any plans on this?

mayrstefan:
From reading the code I would say internally this is already supported. I can see the config object has an attribute "servers" which is an array of strings. Also an API test case use a list of servers - with just one element. So this currently just might be a UI limitation.

mayrstefan:
Reading more UI code my guess would be that hostname could already be a comma separated list of servers. Maybe just the documentation is missing

mayrstefan:
I tried to intercept the request and made a list of multiple servers which gave me this result
`{"baseType":"error","code":"InvalidBodyContent","message":"multiple servers not yet supported","status":422,"type":"error"}`

Looking for this message I found https://github.com/rancher/rancher/blob/2708725597f29f23bea92828e34a1355085ad746/pkg/auth/providers/activedirectory/activedirectory_actions.go#L77-L79

This matches the TODO in https://github.com/rancher/rancher/blob/2708725597f29f23bea92828e34a1355085ad746/pkg/auth/providers/common/ldap/ldap_util.go#L42-L46

mayrstefan:
I don't understand how the UI works. My guess would have been, that the UI should already split a comma separated list of hostnames into an array of servers. But it does not. It keeps the string as entered into the hostname field (e.g. "server1,server2,server3")

mayrstefan:
My fault: rancher/ui already supported this by using a comma-separated list in the hostname field. But the current web ui is rancher/dashboard which does not. I also provided a pull request to restore this feature in the new ui.

24367dfa:
This might be solved by https://github.com/rancher/rancher/pull/37215

Issue #5606, opened 2016-08-01
"java.lang.IllegalStateException: Attempt to cancel when process is still transitioning"


donnlee:
I have a single-container service ('grafana/grafana') that seems to be wedged in state "Starting (Scheduling)".
I tried stopping and starting the container (and service) in Rancher webUI.

Also tried restarting rancher-server:
```
docker stop <id_of_rancher-server>
docker start <id_of_rancher-server>
```

"View in API" (for container):
```
{
"id": "1i936",
"type": "container",
```

rancher-server log:

```
2016-08-01 18:34:00,647 ERROR [d34c8d9a-9a10-4560-8fd7-1b04cda130c9:131925] [instance:936] [instance.start] []
[erviceReplay-10] [c.p.e.p.i.DefaultProcessInstanceImpl] Unknown exception java.lang.IllegalStateException: Att
empt to cancel when process is still transitioning
        at io.cattle.platform.engine.process.impl.DefaultProcessInstanceImpl.runDelegateLoop(DefaultProcessInst
anceImpl.java:190) [cattle-framework-engine-0.5.0-SNAPSHOT.jar:na]
        at io.cattle.platform.engine.process.impl.DefaultProcessInstanceImpl.executeWithProcessInstanceLock(Def
aultProcessInstanceImpl.java:157) [cattle-framework-engine-0.5.0-SNAPSHOT.jar:na]
        at io.cattle.platform.engine.process.impl.DefaultProcessInstanceImpl$1.doWithLock(DefaultProcessInstanc
eImpl.java:107) [cattle-framework-engine-0.5.0-SNAPSHOT.jar:na]
        at io.cattle.platform.engine.process.impl.DefaultProcessInstanceImpl$1.doWithLock(DefaultProcessInstanc
eImpl.java:104) [cattle-framework-engine-0.5.0-SNAPSHOT.jar:na]
        at io.cattle.platform.lock.impl.AbstractLockManagerImpl$3.doWithLock(AbstractLockManagerImpl.java:40) [
cattle-framework-lock-0.5.0-SNAPSHOT.jar:na]
<...>
```

Health check for this service:

```
grafana:
  scale: 1
  health_check:
    port: 3000
    interval: 3000
    initializing_timeout: 60000
    unhealthy_threshold: 3
    strategy: recreate
    request_line: GET "/" "HTTP/1.0"
    healthy_threshold: 2
    response_timeout: 2000
```

What else should I do?  Other info I should provide?

---
[...]

PR #37215, opened 2022-04-06
Support multiple ldap or AD servers

mayrstefan:
Support multiple ldap and AD hosts on the API level as requested in #25557

The implementation is not very sophisticated - if we can not connect to a server we go to the next one in the list. This pull requests only enables setting this on the API level. Multiple servers are not supported in the UI yet.


mayrstefan:
I'm surprised that default-linux-amd64 does not build but all other architectures do. Flaky CI system?

Also i didn't find output for test_auth.py. Do we need any magic environment variables to enable this test?

luthermonson:
@slickwarren any chance you know of a way to QA this with two ldap servers? i think this is done as far as code but we need to give it a once over with a QA pass so it's able to be merged.

slickwarren:
we have both AD and azureAD LDAP setup, I think that would be sufficient to use in this case.

tests/validation/tests/v3_api/test_auth.py
```diff
@@ -610,7 +610,7 @@ def enable_ad(username, token, enable_url=CATTLE_AUTH_ENABLE_URL,
         "groupSearchAttribute": "sAMAccountName",
         "nestedGroupMembershipEnabled": nested,
         "port": PORT,
-        "servers": [HOSTNAME_OR_IP_ADDRESS],
+        "servers": [HOSTNAME_OR_IP_ADDRESS, HOSTNAME_OR_IP_ADDRESS],
```
slickwarren, line 613: this looks like duplicated data, is this line correct? I would expect this to take a new variable that is a list itself or something of the sort, instead of the same variable 2x

pkg/auth/providers/common/ldap/ldap_util.go
```diff
@@ -39,35 +39,36 @@ func NewLDAPConn(servers []string, TLS, startTLS bool, port int64, connectionTim
...
-			return nil, fmt.Errorf("Error creating ssl connection: %v", err)
-		}
-	} else if startTLS {
-		lConn, err = ldapv3.Dial("tcp", fmt.Sprintf("%s:%d", server, port))
-		if err != nil {
-			return nil, fmt.Errorf("Error creating connection for startTLS: %v", err)
-		}
-		if err := lConn.StartTLS(tlsConfig); err != nil {
-			return nil, fmt.Errorf("Error upgrading startTLS connection: %v", err)
+	for _, server := range servers {
```
slickwarren, line 45: will we have UI support for this as well?

tests/validation/tests/v3_api/test_auth.py
mayrstefan: Yes, this line is intended like this: the API should accept a list of more than one server. And I did not want to test any error cases when that host is not reachable. I was not brave enough to replace the first list entry with a wrong servername because my local dapper ci tests don't execute that test. And I have no idea how to enable that.

pkg/auth/providers/common/ldap/ldap_util.go
mayrstefan: Yes, the old UI already has the necessary code to accept a comma separated list and split it into a list. For dashboard I had to re-add that feature. See [#5606](https://github.com/rancher/dashboard/pull/5606/files)

mayrstefan: For reference: 
https://github.com/rancher/ui/blob/543324df8cdb9472eee3950ad3e8b9b45401d749/lib/global-admin/addon/security/authentication/activedirectory/controller.js#L102-L111

slickwarren (approved):

luthermonson (approved):
We will likely merge this and pass to QA after v2.6.5 drops shortly. Thanks again @mayrstefan for the community PR!

jiaqiluo (approved):
LGTM

mayrstefan:
v2.6.5 was released two weeks ago. Do we need anything to change?

Jono-SUSE-Rancher:
@jiaqiluo - This looks good and has approvals from everyone, can we have this merged so it unblocks the associated Dashboard issue?

PR #37216, opened 2022-04-06
Fix clustertemplate condition and add retries

MbolotSuse:
Add retries to the clustertemplaterevision action handler in case it
gets caught trying to update the CTR at the same time as the
secretmigrator controller.

(cherry picked from commit 8ee27472bb0e299cb962e724eb165d22c72e1f5e)


rmweir (approved):
LGTM

cmurphy:

PR #37217, opened 2022-04-06
check for UpstreamSpec before accessing it

luthermonson:
If the EKS cluster handler gets a nil UpstreamSpec it still checks for a NodeGroup count. Default to NO upstream node groups and if the data exists then set to true/false based on data provided.

https://github.com/rancher/rancher/issues/35977

### DIFF
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -190,7 +190,11 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		// for rancher to validate its ability to do so.
 		addNgMessage := "Cluster must have at least one managed nodegroup."
 		noNodeGroupsOnSpec := len(cluster.Spec.EKSConfig.NodeGroups) == 0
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -190,7 +190,11 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 		// for rancher to validate its ability to do so.
 		addNgMessage := "Cluster must have at least one managed nodegroup."
 		noNodeGroupsOnSpec := len(cluster.Spec.EKSConfig.NodeGroups) == 0
-		noNodeGroupsOnUpstreamSpec := len(cluster.Status.EKSStatus.UpstreamSpec.NodeGroups) == 0
+		noNodeGroupsOnUpstreamSpec := true // default to true in case UpstreamSpec is nil
```
rmweir, line 193: By this point the upstream spec should already be set, do we know why that isn't happening? If `noNodeGroupsOnUpstreamSpec` is false then the cluster basically gets set to an error state.

rmweir (changes requested):

PR #37230, opened 2022-04-07
Terraform rke2 standard user token fix

annablender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/824 
 
 
## Problem
 

The problem was that Terraform would throw an error for RKE2 clusters provisioned with a standard user token. The cluster was being prevented from becoming active on refresh if the terraform provider couldn't look up the machine config by ID. The root cause was that RBAC was not configured to allow standard users access to the amazon EC2 machine config in `rke-machine-config.cattle.io.`
 
## Solution
 

Fixed RBAC configuration to allow standard users access to machine config V2

* Here https://github.com/annablender/rancher/blob/17b8d9af09aeea7575618f37e8fc3d8f07a87077/pkg/data/management/role_data.go#L55
* And here https://github.com/annablender/rancher/blob/17b8d9af09aeea7575618f37e8fc3d8f07a87077/pkg/data/management/role_data.go#L480
 
## Testing
 

Verify these scenarios work _the first time_ without error
* Provision an RKE2 EC2 cluster via terraform using an admin token
* Provision an RKE2 EC2 cluster via terraform using a standard user token (click the `Create Clusters` box when creating the user. It is not added by default)

### DIFF
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -52,7 +52,7 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext) 
 		addRule().apiGroups("").resources("secrets").verbs("create").
 		addRule().apiGroups("management.cattle.io").resources("cisconfigs").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("cisbenchmarkversions").verbs("get", "list", "watch").
-		addRule().apiGroups("rke-machine-config.cattle.io").resources("*").verbs("create").
+		addRule().apiGroups("rke-machine-config.cattle.io").resources("*").verbs("create", "get").
 		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch")
 
 	rb.addRole("Manage Node Drivers", "nodedrivers-manage").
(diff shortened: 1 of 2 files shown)


pkg/data/management/role_data.go
```diff
@@ -52,7 +52,7 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext)
 		addRule().apiGroups("").resources("secrets").verbs("create").
 		addRule().apiGroups("management.cattle.io").resources("cisconfigs").verbs("get", "list", "watch").
 		addRule().apiGroups("management.cattle.io").resources("cisbenchmarkversions").verbs("get", "list", "watch").
-		addRule().apiGroups("rke-machine-config.cattle.io").resources("*").verbs("create").
+		addRule().apiGroups("rke-machine-config.cattle.io").resources("*").verbs("create", "get").
```
thedadams, line 55: Standard users shouldn't be able to get all machine configs. They should only be able to get the machine configs they create or specifically given access to.

There is a controller in Rancher that gives standard users access to machine configs they specifically create. My guess is that the terraform provider is moving on before this controller has time to grant this access.

thedadams (changes requested):

pkg/data/management/role_data.go
annablender: Ok that makes sense. I was wondering why standard users couldn't GET machine configs.. it's possible. If you run the apply twice, then terraform allows a standard user to create the cluster. Fix here may be to have the provider wait until that access is granted

annablender:
This fix will be in the terraform provider. Closing.

Issue #36993, opened 2022-03-22
Conditions for CAPI Machines needs to be reorganized


thedadams:
Right now, the `Provisioned` condition is used exclusively for reporting statuses. This isn't intuitive and should be changed. For example, when upgrading a cluster, the `Updated` condition should be used instead.

This should be changed in the planner.

**Additional context**
There are several issues that concern this problem. This is meant to be a gerneral issue that encompasses them all. For example:
- https://github.com/rancher/rancher/issues/36504
- https://github.com/rancher/rancher/issues/36689
- https://github.com/rancher/rancher/issues/36782

Oats87:
Thinking about this, I'm wondering if what we should do is just change the condition we update for different errors, i.e. a snapshot creation error should return a new `ErrUpdating` type of error that would affect the `Updated` condition and `ErrWaiting` can remain for `Provisioned`

thedadams:
### Root cause
The planner only used the `Provisioned` condition when configuring nodes.

### What was fixed, or what changes have occurred
Now, the `Reconciled` condition is always used.

In addition, the `Provisioned` condition was always used for the cluster object. Now, the `Provisioned` condition is used until all the machines are created in the infrastructure provider. After that, the `Updated` condition is used.

Lastly, the Home and Cluster Management pages would show different statuses for the cluster. This has been fixed now and both pages should show the same information.

### Areas or cases that should be tested
Provisioned and custom RKE2 clusters: provisioning, upgrading Kubernetes version, snapshot and restoring, and deleting.

### What areas could experience regressions?
Nothing really. The changes were just how the information is put on conditions and displayed in the UI.

### Are the repro steps accurate/minimal?
Yes.

sowmyav27:
@Josh-Diamond Please validate the following before closing the issue

- Provisioned and custom RKE2 clusters: provisioning, upgrading Kubernetes version, snapshot and restoring, and deleting.
- Cluster state when performing these ^^^ actions and the state on Home and Cluster Management pages
- What is the cluster state when user takes a snapshot on an RKE2 cluster
- What is the cluster state when a worker node is deleted in an RKE2 cluster
- Edit cluster and change cluster config. What is the state of the cluster when cluster is upgrading? related to [this](https://github.com/rancher/rancher/issues/36782) issue
```
- Deploy an RKE2 cluster.
- When the cluster is Active, Edit Config --> Modify the Upgrade strategy to
```

Josh-Diamond:
[...]

PR #37249, opened 2022-04-08
Add better condition syncing for v2 provisioning objects

thedadams:
Issue: https://github.com/rancher/rancher/issues/36993

## Problem
There were issues with different pages of the UI showing different status for
the cluster based on which object it was showing (management vs. provisioning).
This change adds better syncing between the cluster objects so they have the
same conditions and statuses.

## Solution
In addition, the planner always used the Provisioned condition when
changing/applying the node plans. This was confusing, especially when a machine
was being updated. This change will use the Reconciled condition instead.

## Testing
Provisioning and upgrading clusters of all types. Making sure that the Home page and Cluster Management page.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -164,7 +164,7 @@ func New(ctx context.Context, clients *wrangler.Context) *Planner {
 	}
 }
 
-func (p *Planner) applyToMachineCondition(clusterPlan *plan.Plan, machineNames []string, messagePrefix string, messages map[string][]string) error {
+func (p *Planner) setMachineConditionStatus(clusterPlan *plan.Plan, machineNames []string, messagePrefix string, messages map[string][]string) error {
 	var waiting bool
 	for _, machineName := range machineNames {
 		machine := clusterPlan.Machines[machineName]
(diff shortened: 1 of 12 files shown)


pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -173,29 +173,34 @@ func (p *Planner) applyToMachineCondition(clusterPlan *plan.Plan, machineNames [
...
-				continue
-			}
+			waiting = true
+			continue
+		}
+
+		// Once a machine is provisioned (i.e. the first time it is bootstrapped and joined to the cluster),
+		// the machine should not go into a provisioning state again. Instead, updating is used.
+		cond := rke2.Provisioned
+		if cond.IsTrue(machine) {
```
Oats87, line 183: I'm not a fan of changing the conditions we're setting based on other conditions.

Oats87 (changes requested):

pkg/provisioningv2/rke2/planner/planner.go
thedadams: @Oats87 Do you have another suggestion?

Oats87: If the goal is to fix a superficial issue where the user may see "provisioning" on a machine in the UI, then I would lean to fixing this at the UI level. 

My understanding is that the requirement here is entirely to solve a superficial issue where a machine says "provisioning". If that is the case, then maybe we just change the condition entirely to `Reconciled` or something along those lines. 

I see no benefit of this change to the backend other than introducing potential bugs/confusion as we copy/trickle conditions around all over the place.

thedadams:
Note to reviewers: an update to this PR will be needed after [this](https://github.com/rancher/wrangler/pull/201) PR is merged.

pkg/provisioningv2/rke2/planner/planner.go
thedadams: Using Reconciled instead.

paynejacob (approved):

Oats87 (approved):

Oats87 (approved):

Issue #37251, opened 2022-04-08
Allow lasso resync time to be configured via env var


cbron:
This will come in advance of larger fix: https://github.com/rancher/rancher/issues/36785
Part of SURE-4048

brandond:
For what it's worth, upstream intentionally adds a random skew to the resync intervals for each individual controller specifically to avoid this sort of thing. I'm not sure why our default would be a fixed interval, since that pretty much guarantees everything will all refresh at once.

rmweir:
To test this, setup controllers as seen here https://github.com/rancher/rancher/wiki/How-to-setup-prometheus-dashboard-for-controllers. Then change value and ensure the difference is observed in the handler frequency graph. Spike interval should match the env var setting. The env var being set to "15" should look like this: 
![image](https://user-images.githubusercontent.com/19376037/162836985-266d14fd-4971-496f-8849-e358d9c0df62.png)

The env var being set to 0 should have almost no spikes. Some spikes may be there do to cronjob-like activity but there will be less handlers and the execution number will probably be smaller.

If spikes are difficult to observe then making many resources of a 1-3 types will help. Something like making 1000 secrets.

brudnak:
# :x: FAILED / REOPENED 

## Validation Environment

| Component                 | Version / Type                                                                                                                                   |
| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
| Rancher version           | v2.5-cb699448e05d294fae461ec1bae039753b56bb9b-head [cb69944](https://github.com/rancher/rancher/commit/cb699448e05d294fae461ec1bae039753b56bb9b) |
| Installation option       | Helm high availability                                                                                                                           |
| RKE binary version used   | v1.2.19                                                                                                                                          |
| If Helm Chart k8s cluster | v1.20.15                                                                                                                                         |
| Cert Details              | ingress.tls.source=secret (certificates from files)                                                                                              |
| Docker version            | 20.10.7, build f0df350                                                                                                                           |
| Helm version              | v2.16.8-rancher1                                                                                                                                 |
| Downstream cluster type   | not applicable                                                                                                                                   |
| Downstream K8s version    | not applicable                                                                                                                                   |
| Logged in user role       | Administrator                                                                                                                                    |
| Browser type              | Google Chrome                                                                                                                                    |
| Browser version           | 100.0.4896.75 (Official Build) (x86_64)                                                                                                          |

**ADDITIONAL SETUP**

- Helm install commands:

```shell
[...]

PR #37270, opened 2022-04-12
Update lasso

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/37251


rancherbot (approved):

PR #37285, opened 2022-04-12
Fixing flaky test TestGetImages

MbolotSuse:
TestGetImages uses selectFirstEntry (which takes in a map) to
find the WindowsRKE image. In the newest version of kdm, this
image will be the same in win/linux. Now we check to see if
it is an image that should be used by both before calling
it win only.

### DIFF
--- a/pkg/image/resolve_test.go
+++ b/pkg/image/resolve_test.go
@@ -274,9 +274,9 @@ func TestGetImages(t *testing.T) {
 	sourcesToAdd := getImageSourcesList(imagesSet)
 	linuxRKEImage := selectFirstEntry(linuxInfo.RKESystemImages).CoreDNS
 	linuxImagesOnly := append(
-		imagesToAdd,                         // from requirement
-		linuxRKEImage,                       // from system
-		"rancher/prom-alertmanager:v0.17.0", // from chart
+		imagesToAdd,                                    // from requirement
+		linuxRKEImage,                                  // from system
+		"rancher/prom-alertmanager:v0.17.0",            // from chart
 		toolsSystemImages.PipelineSystemImages.Jenkins, // from tools
 	)
 	linuxSourcesOnly := append(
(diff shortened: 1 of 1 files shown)


MbolotSuse:
TestGetImages finds the WindowsRKE image that it will use for the test by selecting the first RKE version that it finds in its map. Unfortunately, since maps are unordered, this image can differ from release to release (to demo - print the key used in the `selectFirstEntry` function and observe how it changes from run to run). Since the latest release uses the same image for windows/linux, this needed to be changed to not add the image to the `windowsImageOnly` or  `windowsSourcesOnly` list. To keep the legacy images behaving the same way (in the cases where `selectFirstEntry` chooses an older version), a loop was devised which checks if the winImage is explicitly called out as being for both windows and linux. If it is not, it is added to windowsImage/Sources only. If it is, then it does not add this value to the respective windows only lists since we expect this image to be in both windows and linux.

pkg/image/resolve_test.go
```diff
@@ -274,9 +274,9 @@ func TestGetImages(t *testing.T) {
 	sourcesToAdd := getImageSourcesList(imagesSet)
 	linuxRKEImage := selectFirstEntry(linuxInfo.RKESystemImages).CoreDNS
 	linuxImagesOnly := append(
-		imagesToAdd,                         // from requirement
-		linuxRKEImage,                       // from system
-		"rancher/prom-alertmanager:v0.17.0", // from chart
+		imagesToAdd,                                    // from requirement
+		linuxRKEImage,                                  // from system
+		"rancher/prom-alertmanager:v0.17.0",            // from chart
```
rmweir, line 279: Did the formatter do this automatically?

rmweir (changes requested):
One question about formatting, otherwise LGTM

pkg/image/resolve_test.go
MbolotSuse: Yeah - I can change back if you want.

rmweir: It's fine if this is how the formatter had it.

rmweir (approved):
LGTM

luthermonson (approved):

PR #37290, opened 2022-04-12
Update ingress-nginx chart version and include necessary values for k3s

rancher-max:
Removed the serviceMonitor part because it now requires CRDs to be installed manually, otherwise the helm chart fails to install with the following error: `Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"`. I don't think we need this part so I just removed it.


ShylajaDevadiga (approved):
LGTM

git-ival (approved):
LGTM, I don't think the servicemonitor bit is used for rancher or any apps AFAIK

Issue #37271, opened 2022-04-12
[Forwardport] v2.6: Allow lasso resync time to be configured via env var


rmweir:
**Original Issue:**
https://github.com/rancher/rancher/issues/37251

brudnak:
# :white_check_mark: PASSED

## Reproduction Environment

#### Not required.

<hr>
  
## Validation Environment

| Component                 | Version / Type                                                                                |
| ------------------------- | --------------------------------------------------------------------------------------------- |
| Rancher version           | v2.6-6bb6e45fba97f58443e547e2f57d2807a424467f-head                                            |
| Rancher commit link       | [6bb6e45](https://github.com/rancher/rancher/commit/6bb6e45fba97f58443e547e2f57d2807a424467f) |
| Installation option       | Helm high availability (in AWS)                                                               |
| RKE binary version used   | v1.3.8                                                                                        |
| If Helm Chart k8s cluster | v1.22.7                                                                                       |
| Cert Details              | ingress.tls.source=secret (certificates from files)                                           |
| Docker version            | 20.10.7, build f0df350                                                                        |
| Helm version              | v2.16.8-rancher1                                                                              |
| Downstream cluster type   | not applicable                                                                                |
| Downstream K8s version    | not applicable                                                                                |
| Logged in user role       | Administrator                                                                                 |
| Browser type              | Google Chrome                                                                                 |
| Browser version           | 100.0.4896.127 (Official Build) (x86_64)                                                      |

**ADDITIONAL SETUP**

- Helm install commands:

```shell
kubectl create namespace cattle-system
```

```shell
helm install rancher rancher-latest/rancher \
  --namespace cattle-system \
  --set hostname=$omitted \
  --set ingress.tls.source=secret \
  --set rancherImageTag=v2.6-head
```

```shell
kubectl -n cattle-system create secret tls tls-rancher-ingress \
  --cert=tls.crt \
  --key=tls.key
```

## Validation steps

1. Run Terraform module to create 1000 secrets in the local cluster under a new project
2. Add `CATTLE_PROMETHEUS_METRICS` with value `true` to the rancher environment variable for the rancher deployment
3. Add `CATTLE_RESYNC_DEFAULT` with a value of `5`
4. Add `CATTLE_AGENT_IMAGE` with a value of `rancher/rancher-agent:v2.6-head`
5. Install monitoring against the local cluster `System`
6. Create a service monitor similar to the following
[...]

PR #37294, opened 2022-04-12
Update remotedialer

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/37271


rancherbot (approved):

Issue #37052, opened 2022-03-25
RKE2 clusters - Snapshots are not available on the UI for a cluster with cloud provider enabled


rak-phillip:
**Rancher Server Setup**
- Rancher version: 44013dc
- Installation option: Docker install

**Information about the Cluster**
- Kubernetes version: v1.22.7+rke2r2 
- Cluster Type (Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom

**User Information**
- What is the role of the user logged in? Admin/Cluster
   
**Describe the bug**

    Deploy an RKE2 cluster - AWS ec2 node driver with Cloud provider enabled on the cluster
    take snapshots on the cluster
    Navigate to Cluster Management --> cluster details --> Snapshots tab
    No snapshots are listed
    Snapshots are available when I ssh into the etcd node:

/var/lib/rancher/rke2/server/db/snapshots# ls -ltr
total 38268
-rw------- 1 root root 13058080 Mar 23 05:00 etcd-snapshot-<etcd node name>-1648011600
-rw------- 1 root root 13058080 Mar 23 05:06 on-demand-sowmya-cp-pool1-59523e22-blrvg-1648011980
-rw------- 1 root root 13058080 Mar 23 05:32 on-demand-sowmya-cp-pool1-59523e22-blrvg-1648013523

Note:

    On a DO Node driver cluster, the snapshots are available on the UI.
    On an EC2 node driver with no Cloud provider enabled, snapshots are available on the UI.

**To Reproduce**
 
1. Deploy an RKE2 cluster - AWS ec2 node driver with Cloud provider enabled on the cluster
2. take snapshots on the cluster
3. Navigate to Cluster Management --> cluster details --> Snapshots tab

**Result**
* No snapshots are listed

**Expected Result**

* On an EC2 node driver with Cloud provider enabled, snapshots are listed

**Screenshots**

### RKE2 Amazon EC2 With Cloud Provider

![image](https://user-images.githubusercontent.com/835961/160195559-6d48f5de-1905-44e5-ab9e-88967bd20687.png)

### RKE2 Amazon EC2 No Cloud Provider

![image](https://user-images.githubusercontent.com/835961/160195650-e919d118-d4b2-428f-a40a-32f1626ff409.png)

**Additional context**

[...]

PR #37379, opened 2022-04-15
More elegantly create and manage v2prov snapshots

Oats87:
Today, etcd snapshots are currently created and maintained by the `snapshotbackpopulate` controller. 

It is possible that the `nodeName` of a node changes, which can throw accountability off. This was causing the issue https://github.com/rancher/rancher/issues/37052 to occur.

This PR changes snapshot creation and maintenance behavior to being properly split between two controllers.

There are two cases of snapshots to consider -- snapshots that are local and snapshots that are stored in S3.

In the local case, snapshots will be created by the `plansecret` controller using the minimal amount of information found from the output of the `etcd-snapshot ls` command, and then populated using the `snapshotbackpopulate` controller.

In the case of S3, snapshots will always be created and maintained by the `snapshotbackpopulate` controller.

Notably, in the case of S3, we will always create and maintain snapshots using the `snapshotbackpopulate` controller, but will only delete S3 snapshots using the `plansecret` controller based on the output of the init node periodic list.

In all cases, etcd snapshots should only be deleted by the `plansecret` controller, unless no machine can be found either by label or node for an existing snapshot object.

### DIFF
--- a/pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
+++ b/pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
@@ -65,9 +65,7 @@ func Register(ctx context.Context, userContext *config.UserContext) {
 
 	userContext.Core.ConfigMaps("kube-system").Controller().AddHandler(ctx, "snapshotbackpopulate", h.OnChange)
 	relatedresource.Watch(ctx, "snapshot-reconcile-trigger", func(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
-		logrus.Tracef("[snapshotbackpopulate] triggering reconcile on object %s/%s - cluster: %s configmap: %s", namespace, name, h.v1ClusterName, h.activeConfigMap)
 		if snapshot, ok := obj.(*rkev1.ETCDSnapshot); ok && snapshot.Spec.ClusterName == h.v1ClusterName && h.activeConfigMap != "" {
-			logrus.Tracef("[snapshotbackpopulate] attempting to set related resource key %s", h.activeConfigMap)
 			return []relatedresource.Key{{
 				Namespace: "kube-system",
 				Name:      h.activeConfigMap,
@@ -163,7 +161,7 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
 			// indicate that it should be OK to delete the etcd snapshot object
 			// don't delete the snapshots here because our configmap can be outdated. we will reconcile based on the system-agent output via the periodic output
 			logrus.Debugf("[snapshotbackpopulate] rkecluster %s/%s: updating status missing=true on etcd snapshot %s/%s as it was not found in the actual snapshot config map", cluster.Namespace, cluster.Name, existingSnapshotCR.Namespace, existingSnapshotCR.Name)
-			logrus.Tracef("[snapshotbackpopulate] rkecluster %s/%s: etcd snapshot was %s/%s: %+v", cluster.Namespace, cluster.Name, existingSnapshotCR.Namespace, existingSnapshotCR.Name, existingSnapshotCR)
+			logrus.Tracef("[snapshotbackpopulate] rkecluster %s/%s: etcd snapshot was %s/%s: %v", cluster.Namespace, cluster.Name, existingSnapshotCR.Namespace, existingSnapshotCR.Name, existingSnapshotCR)
 			existingSnapshotCR = existingSnapshotCR.DeepCopy()
 			existingSnapshotCR.Status.Missing = true // a missing snapshot indicates that it was not found in the (rke2|k3s)-etcd-snapshots configmap. This could potentially be a transient situation after an etcd snapshot restore to an older/newer datastore, when new snapshots have not been taken.
 			if existingSnapshotCR, err = h.etcdSnapshots.UpdateStatus(existingSnapshotCR); err != nil && !apierrors.IsNotFound(err) {
@@ -181,7 +179,7 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
 		if snapshot, ok := currentEtcdSnapshotsToKeep[snapshotKey]; ok {
 			// the snapshot CR exists in the management cluster. check to see if the configmap data needs to be set on the snapshot.
 			logrus.Debugf("[snapshotbackpopulate] rkecluster %s/%s: checking to see if etcdsnapshot %s/%s needs to be updated", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name)
-			logrus.Tracef("[snapshotbackpopulate] rkecluster %s/%s: comparing etcd snapshot %s/%s: %+v : %+v", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name, cmGeneratedSnapshot, snapshot)
+			logrus.Tracef("[snapshotbackpopulate] rkecluster %s/%s: comparing etcd snapshot %s/%s: %v : %v", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name, cmGeneratedSnapshot, snapshot)
 			snapshot = snapshot.DeepCopy()
 			var updated bool
 			if !equality.Semantic.DeepEqual(snapshot.SnapshotFile, cmGeneratedSnapshot.SnapshotFile) {
@@ -197,7 +195,7 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
 				if !equality.Semantic.DeepEqual(snapshot.SnapshotFile, originalSnapshotFile) {
 					updated = true
 					logrus.Debugf("[snapshotbackpopulate] rkecluster %s/%s: snapshot %s/%s SnapshotFile contents were different, triggering update", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name)
-					logrus.Tracef("[snapshotbackpopulate] rkecluster %s/%s: snapshot %s/%s SnapshotFile contents %+v vs %+V", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name, originalSnapshotFile, snapshot.SnapshotFile)
+					logrus.Tracef("[snapshotbackpopulate] rkecluster %s/%s: snapshot %s/%s SnapshotFile contents %v vs %v", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name, originalSnapshotFile, snapshot.SnapshotFile)
 				}
 			}
 			if snapshot.Spec.ClusterName != cmGeneratedSnapshot.Spec.ClusterName {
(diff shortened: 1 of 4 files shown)


pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -60,6 +64,17 @@ func Register(ctx context.Context, userContext *config.UserContext) {
 	}
 
 	userContext.Core.ConfigMaps("kube-system").Controller().AddHandler(ctx, "snapshotbackpopulate", h.OnChange)
+	relatedresource.Watch(ctx, "snapshot-reconcile-trigger", func(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
+		logrus.Tracef("[snapshotbackpopulate] triggering reconcile on object %s/%s - cluster: %s configmap: %s", namespace, name, h.v1ClusterName, h.activeConfigMap)
```
thedadams, line 68: I would prefer that this is debug level. This is called by the wrangler controller only, so I don't know how much including the trace would benefit here.

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -60,6 +64,17 @@ func Register(ctx context.Context, userContext *config.UserContext) {
...
+		if snapshot, ok := obj.(*rkev1.ETCDSnapshot); ok && snapshot.Spec.ClusterName == h.v1ClusterName && h.activeConfigMap != "" {
+			logrus.Tracef("[snapshotbackpopulate] attempting to set related resource key %s", h.activeConfigMap)
```
thedadams, line 70: Same as above.

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -71,13 +86,21 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
...
 	if err != nil || len(clusters) != 1 {
 		return configMap, fmt.Errorf("error while retrieving cluster %s from cache via index: %w", h.clusterName, err)
 	}
 
 	cluster := clusters[0]
 
+	if h.v1ClusterName == "" {
+		h.v1ClusterName = cluster.Name
+	}
+
```
thedadams, line 103: Can this be put in the `Register` function by getting the v1 cluster with the index `ByCluster`?

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -71,13 +86,21 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
 		return configMap, nil
 	}
 
+	if h.activeConfigMap == "" {
+		h.activeConfigMap = configMap.Name
+	}
```
thedadams, line 91: Similarly, can this be put in the `Register` function by getting the cluster object with the index and checking which distribution is being used?

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -137,60 +174,107 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
...
-				return configMap, fmt.Errorf("rkecluster %s/%s: error while creating etcd snapshot %s/%s: %w", cluster.Namespace, cluster.Name, cmGeneratedSnapshot.Namespace, cmGeneratedSnapshot.Name, err)
+				return configMap, fmt.Errorf("rkecluster %s/%s: error while creating s3 etcd snapshot %s/%s: %w", cluster.Namespace, cluster.Name, cmGeneratedSnapshot.Namespace, cmGeneratedSnapshot.Name, err)
 			}
 		}
 	}
 	return configMap, nil
 }
 
+func reconcileStringMaps(original map[string]string, new map[string]string, keys []string) (map[string]string, bool) {
+	ret := original
```
thedadams, line 257: I don't think this is necessary because maps are passed by reference and `original` will get the changes as well, even outside this function.

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -137,60 +174,107 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
...
+
+// configMapToSnapshots parses the given configmap and returns a map of etcd snapshots. The snapshotbackpopulate controller will only create snapshots from this map that are located in S3.
+// The snapshots will have 2 or 3 labels, 2 if they are S3 snapshots (cluster name and node name), and 3 if they are local snapshots. They will always have 2 annotations.
 func (h *handler) configMapToSnapshots(configMap *corev1.ConfigMap, cluster *provv1.Cluster) (map[string]rkev1.ETCDSnapshot, error) {
 	result := map[string]rkev1.ETCDSnapshot{}
 	for k, v := range configMap.Data {
 		file := &snapshotFile{}
 		if err := json.Unmarshal([]byte(v), file); err != nil {
-			logrus.Errorf("invalid non-json value in %s/%s for key %s in cluster %s", configMap.Namespace, configMap.Name, k, h.clusterName)
+			logrus.Errorf("invalid non-json value in %s/%s for key %s in cluster %s", configMap.Namespace, configMap.Name, k, cluster.Name)
```
thedadams, line 275: Should we include the `err` here?

pkg/controllers/provisioningv2/rke2/plansecret/plansecret.go
```diff
@@ -58,7 +62,7 @@ func (h *handler) OnChange(key string, secret *corev1.Secret) (*corev1.Secret, e
 		}
 	}
 
-	if v, ok := node.PeriodicOutput["etcd-snapshot-list-s3"]; ok && v.ExitCode == 0 && len(v.Stdout) > 0 {
+	if v, ok := node.PeriodicOutput["etcd-snapshot-list-s3"]; ok && v.ExitCode == 0 && len(v.Stdout) > 0 && secret.Labels[rke2.InitNodeLabel] == "true" {
```
thedadams, line 65: Do we need this check if the command is only sent to the init node?

pkg/controllers/provisioningv2/rke2/plansecret/plansecret.go
```diff
@@ -143,40 +158,93 @@ func (h *handler) reconcileEtcdSnapshotList(secret *corev1.Secret, s3 bool, list
...
+			// delete the etcd snapshot as it is missing
+			logrus.Infof("[plansecret] Deleting etcd snapshot %s/%s", v.Namespace, v.Name)
 			if err := h.etcdSnapshotsClient.Delete(v.Namespace, v.Name, &metav1.DeleteOptions{}); err != nil {
 				return err
 			}
 		}
+		indexedEtcdSnapshots[v.Name] = v
 	}
+
+	if !s3 && machine != nil {
```
thedadams, line 207: It is not possible to get here with `s3 == false` and `machine == nil`.

thedadams (changes requested):

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
Oats87: My understanding with `logrus.Tracef` is that it only prints when the log level is at `trace`. That being said, I don't think it's actually even useful to have this here so I'm going to remove it.

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
Oats87: The indexers/caches may not be synced when `Register` is called. We can artificially slow the controller down from running until those caches are synced, but I think it's fine to just leave this doing what it is doing now.

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
Oats87: We *can*, but I don't know how useful that is. It's an extra hit against the cache, whereas we know we'll only ever process 1 configmap per instance of a cluster with this controller.

pkg/controllers/provisioningv2/rke2/plansecret/plansecret.go
Oats87: Yes.

If an init node is re-elected, the old init node will still have the periodic output even if the instruction is not defined.

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -197,22 +195,20 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
...
 				updated = true
 			}
-			if newLabels, labelsUpdated := reconcileStringMaps(snapshot.Labels, cmGeneratedSnapshot.Labels, []string{rke2.ClusterNameLabel, rke2.NodeNameLabel, rke2.MachineIDLabel}); labelsUpdated {
+			if labelsUpdated := reconcileStringMaps(snapshot.Labels, cmGeneratedSnapshot.Labels, []string{rke2.ClusterNameLabel, rke2.NodeNameLabel, rke2.MachineIDLabel}); labelsUpdated {
 				logrus.Debugf("[snapshotbackpopulate] rkecluster %s/%s: snapshot %s/%s labels did not match", cluster.Namespace, cluster.Name, snapshot.Namespace, snapshot.Name)
-				snapshot.Labels = newLabels
 				updated = true
 			}
-			if newAnnotations, annotationsUpdated := reconcileStringMaps(snapshot.Labels, cmGeneratedSnapshot.Labels, []string{SnapshotNameKey}); annotationsUpdated {
+			if annotationsUpdated := reconcileStringMaps(snapshot.Annotations, cmGeneratedSnapshot.Annotations, []string{SnapshotNameKey}); annotationsUpdated {
```
thedadams, line 210: Good catch! Sorry I didn't catch this myself.

thedadams (approved):

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -95,26 +118,40 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
...
 			if storageLocation != StorageS3 {
 				// check to make sure the machine actually exists for the snapshot
-				listSuccessful, machine, err := rke2.GetMachineFromNode(h.machineCache, existingSnapshotCR.SnapshotFile.NodeName, cluster)
+				var listSuccessful bool
+				var machine *capi.Machine
+
+				if existingSnapshotCR.Labels[rke2.MachineIDLabel] == "" {
+					logrus.Debugf("[snapshotbackpopulate] rkecluster %s/%s: snapshot %s/%s was missing machine ID label: %s", cluster.Namespace, cluster.Name, existingSnapshotCR.Namespace, existingSnapshotCR.Name, rke2.MachineIDLabel)
+					// If the machineID label was not set, fall back to looking up the machine by node name, as this may be a snapshot from an earlier version of Rancher that created local snapshots using the snapshotbackpopulate controller, which means the snapshot should not have the machine ID label.
+					listSuccessful, machine, err = rke2.GetMachineFromNode(h.machineCache, existingSnapshotCR.SnapshotFile.NodeName, cluster)
```
annablender, line 146: Is this ok to do here since the `NodeName` might change? Is there an alternative way to looking up the machine in the list?

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -137,60 +174,107 @@ func (h *handler) OnChange(key string, configMap *corev1.ConfigMap) (runtime.Obj
...
-				snapshot.SnapshotFile.Metadata = md
-				snapshot.SnapshotFile.Message = msg
+				if originalSnapshotFile.Metadata != "" && snapshot.SnapshotFile.Metadata == "" {
+					snapshot.SnapshotFile.Metadata = originalSnapshotFile.Metadata
+				}
+				if originalSnapshotFile.Message != "" && snapshot.SnapshotFile.Message == "" {
+					snapshot.SnapshotFile.Message = originalSnapshotFile.Message
+				}
+				if !equality.Semantic.DeepEqual(snapshot.SnapshotFile, originalSnapshotFile) {
+					updated = true
```
annablender, line 196: `updated = true` could be moved after log messages

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
```diff
@@ -254,6 +322,16 @@ func (h *handler) configMapToSnapshots(configMap *corev1.ConfigMap, cluster *pro
 				Controller:         &[]bool{true}[0],
 				BlockOwnerDeletion: &[]bool{true}[0],
 			}}
+			snapshot.Annotations[StorageAnnotationKey] = StorageS3
+		} else {
```
annablender, line 321: So, this code is executed if the snapshot is local (stored on a node) right? I thought it'd be helpful to have a comment here to make sure that's clear.

pkg/controllers/provisioningv2/rke2/plansecret/plansecret.go
```diff
@@ -143,40 +158,93 @@ func (h *handler) reconcileEtcdSnapshotList(secret *corev1.Secret, s3 bool, list
...
-	if err != nil {
-		return err
-	}
 
 	etcdSnapshots, err := h.etcdSnapshotsCache.List(secret.Namespace, ls)
 	if err != nil {
 		return err
 	}
 
+	// indexedEtcdSnapshots is a map of etcd snapshots that already exist in the management cluster
```
annablender, line 193: `exists` instead of `exist`

pkg/controllers/managementuser/snapshotbackpopulate/snapshotbackpopulate.go
Oats87: This is for backwards compatibility. The alternative way to looking up the machine is 2 lines below,where we look up the machine by the machine ID label on the snapshot.

annablender: Thanks!

thedadams (approved):

Issue #35199, opened 2021-10-19
Implement cache-control headers on responses from Rancher API response for agent nodeplan


SheilaghM:
SURE-2953

**Rancher Server Setup**
CentOs 7.9, Docker 19.03.14
K8s 1.17.17-rancher1-1 , Rancher 2.5.7

... Singular RMC on 2.5.7, 3x Downstream Custom Clusters (Prod/Dev/Test)

**Describe the bug**
... Singular RMC on 2.5.7, 3x Downstream Custom Clusters (Prod/Dev/Test).. All clusters are unhealthy, etcd is down.. We are seeing agents fire up in one cluster and deploy workloads for another cluster, multiple agents running on downstream CP nodes with versions 2.4.12, 2.5.5, and 2.5.7.  We are getting SSL errors across the board, its like they are getting the wrong certs. 

The short version is the customer had clusters where the nodes' were registering to the wrong clusters/environments due to enabling a caching feature option on their F5 load balancer.  A workaround was introduced - see SURE ticket.

Business impact:

Affects all but one of their downstream clusters.  Only their prod cluster is showing as operating in Rancher.

Troubleshooting steps:

Tried rolling back VM snapshots from downstream control-plane node, restarting agents on downstream control-plane node.  Still seeing many cert errors.

**To Reproduce**
troubleshooting uncovered a proxy that had been caching the api calls, they disabled that and their issues went away.. it was determined that setting cache control headers would had been adequate to prevent such a failure in any other environment.

cmurphy:
Some notes on design/implementation based on @superseb 's comment in jira:

> * Design should cover implementing this for all API responses
> * If not, this customer issue should be specifically mentioned to see if we want to implement it for the nodeplan endpoint only or not at all
> * Possible regression/risks of implementing it

### Implementing for all API responses

[...]

PR #37391, opened 2022-04-18
[2.6.6] Set Cache-Control header default on all responses

cmurphy:
Use the apiserver NoCache middleware to set "Cache-Control: no-cache,
no-store, must-revalidate" on all Rancher endpoint responses. This
prevents problems with clients trying to use cached data that is not
properly invalidated when it becomes stale on the server.

Technically, "no-cache" and "must-revalidate" are superfluous when
"no-store" is set[1].

The header is set early in the middleware chain, and can be overridden
by another middleware or by a route handler. This happens, for instance, with
the catalog endpoint response[2].

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_storing
[2] https://github.com/rancher/rancher/blob/6c12bee2d0bee16a2854df5471aaa0cdee5b9bf1/pkg/api/norman/customization/catalog/template.go#L133

https://github.com/rancher/rancher/issues/35199

### DIFF
--- a/pkg/rancher/rancher.go
+++ b/pkg/rancher/rancher.go
@@ -210,6 +210,7 @@ func New(ctx context.Context, clientConfg clientcmd.ClientConfig, opts *Options)
 		Handler: responsewriter.Chain{
 			auth.SetXAPICattleAuthHeader,
 			responsewriter.ContentTypeOptions,
+			responsewriter.NoCache,
 			websocket.NewWebsocketHandler,
 			proxy.RewriteLocalCluster,
 			clusterProxy,


thedadams (approved):

pkg/rancher/rancher.go
```diff
@@ -210,6 +210,7 @@ func New(ctx context.Context, clientConfg clientcmd.ClientConfig, opts *Options)
 		Handler: responsewriter.Chain{
 			auth.SetXAPICattleAuthHeader,
 			responsewriter.ContentTypeOptions,
+			responsewriter.NoCache,
```
MbolotSuse, line 213: One question here:

IIRC the original resource that this was requested for was the nodeplan resources. Near as I can tell, NodePlan is a field of the Node object (so really you would use the node endpoint to get the node and then get the plan from that).

Norman has a customization in place for the node objects which sets cache-control to private - (https://github.com/rancher/rancher/blob/649e3345fcb9c913b1dd2c056ac9cf8ade1ac3e9/pkg/api/norman/customization/node/node.go#L268). Would we want to change this?

Notes on the above:
- That only seems to happen for the links field, so the user would have to do something that interacted with the link to get that override value. I think that this is unlikely (basic get to this endpoint didn't have the header).
- Since the header is set to `private`, I think that it is unlikely that this would cause cache issues (from what I was reading, CDNs and the like would not cache the nodeplan's even if the node response came from this endpoint - it would only be cached by the browser).

All that being said, I'm wondering if the above is something that we should change.

thedadams: The node config/plan is retrieved via the endpoint `/v3/connect/config`, not the node endpoint. In addition, as you pointed out, this is unlikely to cause the caching problem that we are trying to fix here.

MbolotSuse: Thanks for pointing that out. Since this appears to be active before the MCM routes (which is where this occurs) and since that route seems to not override the header, I'm good here.

MbolotSuse (approved):
LGTM - questions are answered.

PR #37399, opened 2022-04-19
Document impersonation package

cmurphy:


pkg/impersonation/impersonation.go
```diff
@@ -185,6 +193,9 @@ func (i *Impersonator) createNamespace() error {
 	return err
 }
 
+// checkAndUpdateRole checks whether the impersonation clusterrole already exists and whether it has the correct rules.
+// If the role does not exist, the function returns and createRole must be called.
```
maxsokolovsky, line 197: The smallest of nitpicks - method, rather than function.

pkg/impersonation/impersonation.go
```diff
@@ -47,6 +52,8 @@ func New(userInfo user.Info, clusterContext *config.UserContext) (Impersonator,
 	return impersonator, nil
 }
 
+// SetUpImpersonation creates a service account on a cluster with a clusterrole and binding allowing it to impersonate a Rancher user.
```
maxsokolovsky, line 55: Binding here means "cluster role binding", right? I'd be explicit about that, since Kubernetes allows to link cluster roles with role bindings OR cluster role bindings.

pkg/impersonation/impersonation.go
MbolotSuse, line 197: Correct me if I am wrong, but the purpose of L197 seems to be that `checkAndUpdateRole` will return an error if the role doesn't exist (rather than creating the role itself). Also, if it does that and you still want a role, you need to call createRole.

One suggestion:
```
If the role does not exist, the function returns an error and does NOT create the role. 
```

pkg/impersonation/impersonation.go
```diff
@@ -47,6 +52,8 @@ func New(userInfo user.Info, clusterContext *config.UserContext) (Impersonator,
...
+// Returns a reference to the service account which can be used to retrieve the account token, or an error if creating any of the resources failed.
```
MbolotSuse, line 56: Not sure that `which can be used to retrieve the account token` is necessary here. Alternatively, if you want to include it, maybe also include a reference to the below `GetToken` function which you would use to get the token. 

Something like:
```
Returns a reference to the service account (can be used by GetToken to retrieve its token), or an error if creating any of the resources failed.
```

pkg/impersonation/impersonation.go
```diff
@@ -20,18 +21,22 @@ import (
 )
 
 const (
-	impersonationLabel     = "authz.cluster.cattle.io/impersonator"
+	impersonationLabel = "authz.cluster.cattle.io/impersonator"
+	// ImpersonationNamespace is the namespace where impersonation service accounts live.
 	ImpersonationNamespace = "cattle-impersonation-system"
-	ImpersonationPrefix    = "cattle-impersonation-"
+	// ImpersonationPrefix is the prefix for impersonation roles, bindings, and service accounts.
```
MbolotSuse, line 27: Phrasing nitpick:
`ImpersonationPrefix is the prefix used for impersonation related resources (roles, bindings, service accounts, etc.)`

MbolotSuse (dismissed):
Several minor nitpicks.

I'm approving since the changes are docs only and I have no accuracy-based comments (more minor stuff like phrasing).

pkg/impersonation/impersonation.go
cmurphy: There is no "etc", there are only these three things. Will leave as is.

pkg/impersonation/impersonation.go
cmurphy: > the purpose of L197 seems to be that checkAndUpdateRole will return an error if the role doesn't exist

Not sure if you mean L197 on the old change or the new change. If you refer to line 205 in the new change, if the role does not exist, we do not return an error. This is what is meant by "If the role does not exist, the function returns and createRole must be called.". I can change that to "If the role does not exist, the method returns nil for the role and createRole must be called."

> Also, if it does that and you still want a role, you need to call createRole.

You always want the role. If the role is not found in this method, you must call createRole.

Will not make semantic changes on this one, this comment accurately describes this method's usage.

pkg/impersonation/impersonation.go
cmurphy: done

pkg/impersonation/impersonation.go
cmurphy: done

pkg/impersonation/impersonation.go
cmurphy: done

MbolotSuse (approved):
LGTM

maxsokolovsky (approved):

PR #37398, opened 2022-04-19
Fix S3 snapshots to work with newer versions of K3s/RKE2

Oats87:
https://github.com/rancher/rancher/issues/34417

Issue #34417, opened 2021-08-24
RKE2 Provisioning: Day 2 Ops - On Demand/Recurring Etcd Snapshot and Restore with Options


deniseschannon:
This is to validate S3 Support for snapshot and restore. 

```
apiVersion: provisioning.cattle.io/v1
kind: Cluster
metadata:
  name: #string
  annotations:
    {}
#    key: string
  labels:
    {}
#    key: string
  namespace: fleet-default
spec:
  cloudCredentialSecretName: do
  kubernetesVersion: v1.21.0-alpha2+rke2r1
  rkeConfig:

#    etcdSnapshotCreate:
#      name: string
#      nodeName: string
#      s3:
#        bucket: string
#        cloudCredentialName: string
#        endpoint: string
#        endpointCA: string
#        folder: string
#        region: string
#        skipSSLVerify: boolean
```

Note: Things changed with https://github.com/rancher/rancher/issues/34088
An s3-compatible cloud credential was created, you can create one and specify when making a cluster, and override defaults (if any are set).

Oats87:
Now that this is merged into `release/v2.6`, here is the summary of changes for this

1. `etcd snapshot` information is no longer stored on the `clusters.provisioning.cattle.io/status` field, they are now stored in dedicated `etcdsnapshot.rke.cattle.io` objects. They are in the same namespace as the cluster, and have a cluster name label (`rke.cattle.io/cluster-name`) on them that can be used to look for them.
2. `name` on the `etcdSnapshotRestore` field now refers to the name of the CR of the `etcdSnapshot` CR. `nodeName` is no longer an option as that is stored in the metadata of the `etcdSnapshot` CR.
3. The `etcdSnapshotRestore` struct has a new format. `restoreRKEConfig` is a string field that (currently) accepts three options, `none`, `kubernetesVersion`, and `all`. They are the same as what RKE1 will do when restoring from snapshot IF the metadata exists. The option should only probably be exposed to restore more than the snapshot if the snapshot `SnapshotFile.metadata` file exists. Generation behavior is the same -- it only needs to be incremented if the same snapshot is being restored more than once. The format is now:
```
type ETCDSnapshotRestore struct {
	// Name refers to the name of the associated etcdsnapshot object
	Name string `json:"name,omitempty"`

	// Changing the Generation is the only thing required to initiate a snapshot restore.
	Generation int `json:"generation,omitempty"`
	// Set to either none (or empty string), all, or kubernetesVersion
[...]

PR #37401, opened 2022-04-19
Fix S3 snapshots to work with newer versions of K3s/RKE2 pt. 2

Oats87:
Follow up of: https://github.com/rancher/rancher/pull/37398

https://github.com/rancher/rancher/issues/34417


snasovich (approved):
LGTM, but I'll be honest I'm not fully understanding the logic here so definitely need @thedadams approval before merging.

thedadams (approved):

Issue #36072, opened 2021-10-19
[Backport v2.6] [BUG] Deleted cluster not removed from UI view until hard refresh (Ghost cluster)


janeczku:
Rancher 2.6.1

1. Login as standard, non-admin user
2. Create a RKE2 cluster
3. In a second browser login as admin user and navigate to Cluster Manager view
4. As the non-admin user delete the cluster 

**Expected:**

When the cluster is deleted it is removed from the view.

**Actual:**

Normal user session: Deleted cluster is not removed from the view until performing a hard browser refresh.
Admin user session:  Deleted cluster is removed from the view immediately

https://user-images.githubusercontent.com/3813921/137932366-12f9f499-0e30-4c17-86b5-b361ba95c5a5.mov

````
2021/10/19 15:03:11 [INFO] rkecluster fleet-default/template-rke2: waiting for at least one bootstrap node
2021/10/19 15:03:11 [INFO] [mgmt-auth-crtb-controller] Deleting roleBinding crb-n4zbcfymqv
2021/10/19 15:03:11 [INFO] [mgmt-auth-crtb-controller] Deleting rolebinding creator-cluster-owner-cluster-owner in namespace p-l9xs6 for crtb creator-cluster-owner
2021/10/19 15:03:11 [INFO] [mgmt-auth-crtb-controller] Deleting rolebinding creator-cluster-owner-cluster-owner in namespace p-hhc6q for crtb creator-cluster-owner
2021/10/19 15:03:12 [INFO] [mgmt-project-rbac-remove] Deleting namespace p-l9xs6
2021/10/19 15:03:12 [INFO] [mgmt-project-rbac-remove] Deleting namespace p-hhc6q
2021/10/19 15:03:16 [ERROR] error syncing 'c-m-v8p56lnd': handler cluster-watch: clusterregistrationtokens.management.cattle.io "default-token" is forbidden: unable to create new content in namespace c-m-v8p56lnd because it is being terminated, requeuing
2021/10/19 15:03:16 [ERROR] error syncing 'c-m-v8p56lnd': handler cluster-watch: clusterregistrationtokens.management.cattle.io "default-token" is forbidden: unable to create new content in namespace c-m-v8p56lnd because it is being terminated, requeuing
[...]

PR #37416, opened 2022-04-20
Changing cluster/project roles so they persist until cluster delete

MbolotSuse:
Changes CRTBs and PRTBs so that they persist until the cluster is
fully deleted. This allows users to see deletion messages as
the cluster is removed and to see the cluster gone on the UI


ryansann (approved):

MbolotSuse:
Related to #36072

PR #37418, opened 2022-04-20
Put more logs in provisioning log

thedadams:
The original intention of the provisioning log was to put all logs related to
cluster provisioning in the log. When we switched to using the `Updated`
condition, this was not the case.

After this change, provisioning all provisioning logs will be included.


jiaqiluo (approved):
LGTM

Oats87 (approved):

PR #37427, opened 2022-04-21
Ensure that Rancher serves the right Windows install script when airgapped

rosskirkpat:
Currently, Rancher is attempting to serve a `wins-install.ps1` script when airgapped and the remote script is not available. This PR addresses this by switching to serving `wins-agent-install.ps1` instead as well as adding unit tests that attempt to mock airgap by testing with an unstarted http test server, to mimic an airgap setup.


phillipsj (approved):

rosskirkpat:
https://github.com/rancher/rancher/issues/37428

thedadams (approved):

PR #37442, opened 2022-04-22
Replace ioutil with io and os

estensen:
`ioutil` was depcrecated in [Go 1.16](https://go.dev/doc/go1.16)


luthermonson (approved):
We will get this merged after v2.6.5 is released. Thanks for the PR!

Oats87:
We should probably merge this at some point

Issue #37432, opened 2022-04-21
Default k8s version switch to 1.23


snasovich:
For 2.6.5 release, we want to switch the default k8s version to 1.23.

The change needs to happen for the following (list may need to be expanded):
* Downstream RKE1 (in https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/k8s_version_info.go)
* Downstream RKE2/k3s (in https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/channels.yaml and https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/channels-rke2.yaml using `appDefaults` section as introduced per https://github.com/rancher/rancher/issues/36827)
* RKE1 itself (?)
* k3s version embedded into Rancher when running in Docker

jiaqiluo:
## Note for updating the default version of RKE
Because rancher/rke pulls the metadata from the URL `https://releases.rancher.com/kontainer-driver-metadata/release-v2.6/data.json`, updating the default version of RKE will happen after the KDM is released. 

ref:
https://github.com/rancher/rke/blob/release/v1.3/codegen/codegen.go

jiaqiluo:
The fix is available in the latest 2.6-head

### Areas or cases that should be tested
- in the docker-install mode, the local cluster should use k3s v1.23.6+k3s1
- when creating a node-driver RKE1/RKE2/K3s cluster, the default version should be the latest 1.23.x

jiaqiluo:
the fix can be validated in [v2.6.5-rc7](https://github.com/rancher/rancher/releases/tag/v2.6.5-rc7)

slickwarren:
tested on v2.6.5-rc8: 
* rke 1 and rke2 clusters now default to v1.23.6
* docker install default version is v1.23.6+k3s1

PR #37453, opened 2022-04-22
[2.6.5] Set local k8s 1.23.6

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/37432
bump the local k3s cluster to v1.23.6-k3s1

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-ARG K3S_BUILDER=rancher/k3s:v1.22.7-k3s1
+ARG K3S_BUILDER=rancher/k3s:v1.23.6-k3s1
 
 # Use binaries from k3s root for s390x as k3s is not available on s390x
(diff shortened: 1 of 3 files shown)


kinarashah (approved):

snasovich (changes requested):
There is now `v1.23.6-k3s1` (https://github.com/k3s-io/k3s/releases/tag/v1.23.6%2Bk3s1) so we should use at least it.

Note that I've confirmed with RKE2 team they are working on releasing `v1.23.6-rke2r2`, but there won't be `v1.23.6-k3s2` so we're safe to go to `v1.23.6-k3s1` and don't need to wait for anything newer for Rancher v2.6.5.

snasovich (approved):

jakefhyde (approved):

annablender (approved):

Issue #37443, opened 2022-04-22
Project level PSP is not respected while creating deployment.


vivek-infracloud:
**Rancher Server Setup**
- Rancher version: v2.6.5-rc2
- Installation option (Docker install/Helm Chart): Docker

**Information about the Cluster**
- Kubernetes version: v1.22.8+rke2r1
- Cluster Type (Local/Downstream): Downstream, Node Driver, AWS.

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
**Admin**
   
**Describe the bug**

PSP set to the project is not respected while creation of pod.
If you set restricted PSP as default to cluster and create a new namespace with unrestricted PSP, the Pod/Deployment creation in the new project fails with following error.

`Pods "nginx-7fbb89c5fd-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].securityContext.allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed]:Deployment does not have minimum availability.
`

**To Reproduce**
 

- Create a rancher server v2.6.5-rcX
- Deploy a AWS node driver RKE2 cluster with default PSP set to restricted.
- Create a new Project in cluster with PSP set to unrestricted.
- Create a new namespace in the newly created Project.
- Try to create a deployment with Privileged flag on in the newly created namespace.

The above mentioned scenario is one. There is another scenario where cluster default PSP is set to unrestricted and the Project level PSP is set to restricted. In this scenario the Deployment creation should fail, however, the deployment creation succeeds. 

**Result**

- Deployment creation fails in first scenario.
- Deployment creation succeeds in the second scenario. 

**Expected Result**

- Deployment creation should succeed in first scenario
- Deployment creation should fail in second scenario.

MbolotSuse:
@snasovich - please feel free to correct any of the RKE2 related information below if it is inaccurate.

# Release Note

## Synopsis
RKE2 clusters which are provisioned by rancher are assumed to always have the Pod Security Policy admission controller enabled. If you manually disable this admission controller through kube-api-server args, rancher may still allow you to set Pod Security Policies for projects. However, these Pod Security Policies will have no effect.

## Resolved
No (the original issue that this is attached to was resolved - this release note item is for the side-effect which is not resolved).

Status: Side-effect may be resolved in a future release.
[...]

PR #37475, opened 2022-04-26
RKE2 provisioned cluster are viewed as always having PSP enabled

KevinJoiner:
Related to #37443


KevinJoiner:
## Root Causes
When creating a new project using a provisioned RKE2 cluster the action to set a PSP does not work.
This is because the [field](https://github.com/rancherlabs/rancher-security/blob/7db4fe4e81939a5f6b420f303478a169cc35588a/pkg/controllers/management/cluster/clustercontroller.go#L224) for determining if a cluster has PSP capability is set only for RKE1 clusters. This field is checked when [creating a new project](https://github.com/rancherlabs/rancher-security/blob/7db4fe4e81939a5f6b420f303478a169cc35588a/pkg/api/norman/customization/project/project_actions.go#L288-L291)  with a PSP set.

## Changes
 - Check if cluster is RKE2 provisioned when determining if we should return an error for the set PSP action on projects.

## Notes
 - [Docs](https://docs.rke2.io/security/policies/) indicate the RKE2 clusters are always provisioned with PSP enabled.
 -  The decision to check for RKE2 provisioned clusters instead of adding the capability in [the controller](https://github.com/rancherlabs/rancher-security/blob/7db4fe4e81939a5f6b420f303478a169cc35588a/pkg/controllers/management/cluster/clustercontroller.go#L224) was to limit affects on other provisioning capabilities.
    -  currently RKE2 clusters do not have any capabilities set in the controller.

pkg/api/norman/customization/project/project_actions.go
```diff
@@ -443,3 +445,8 @@ func getID(id interface{}) (string, error) {
 	split := strings.Split(s, ":")
 	return split[0] + ":" + split[len(split)-1], nil
 }
+
+// isProvisionedRke2Cluster check to see if this is a rancher provisioned rke2 cluster
+func isProvisionedRke2Cluster(cluster *v3.Cluster) bool {
+	return cluster.Status.Provider == "rke2" && imported.IsAdministratedByProvisioningCluster(cluster)
```
thedadams, line 451:
```suggestion
	return cluster.Status.Provider == v32.ClusterDriverRke2 && imported.IsAdministratedByProvisioningCluster(cluster)
```

thedadams (changes requested):
One small change. Otherwise, looks good!

thedadams (approved):

MbolotSuse (approved):
LGTM.

Worked on this with @KevinJoiner offline. The one flaw in this solution is the assumption that RKE2 always has the PSP admission controller enabled. This can end up not being the case if the users manually set the kube api server args.

However, I'm still ok with this solution for the following reasons:
- This code only applies for RKE2 clusters deployed from rancher
- Without this code, project level psps will fail to apply (I think that this is the more common use case then turning off the admission controller and still expecting psps to work)
- It seems that if you want to turn this off, you have to set the option in the yaml manually. So the flaw is for a more advanced use case. 
- We can release note the remaining portion of the issue so that users are aware of the problem.

cmurphy (approved):

Issue #37465, opened 2022-04-25
downstream rke2 cluster goes to `provisioning` state when upgrading rancher from 2.6.4 -> 2.6.5-rc2


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.6.5-rc2
- Installation option (Docker install/Helm Chart): docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: any
- Cluster Type (Local/Downstream): downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): rke2, any cluster type

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom): standard user, cluster owner

   
**Describe the bug**
when upgrading the rancher server from 2.6.4 -> 2.6.5-rc2 the downstream cluster goes into a `provisioning` state. (rke1 clusters go into an `unavailable` state) before coming back to an `active` state

**To Reproduce**
* deploy rancher 2.6.4
* deploy an rke2 downstream cluster
* upgrade rancher to 2.6.5-rc2

**Result**
the downstream cluster goes into a provisioning state with the below logs:
```
[INFO ] configuring bootstrap node(s) custom-59e5e92a772a: waiting for plan to be applied
[INFO ] provisioning done
```
after which, it comes back to an active state. 

**Expected Result**
cluster should not enter `provisioning` state upon upgrading.

snasovich:
For now considering this release-blocker as we really want to try to fix this logic before going GA with RKE2. If the initial attempt to fix it is not successful, we can remove the label and address in 2.6.6.

snasovich:
Transferring to @Oats87 following the closure of previous approach from https://github.com/rancher/rancher/pull/37482

thedadams:
@Oats87 While driving, I realized that the approach we discussed on Slack won't fix the issue without more changes.

We talked about adding a `Remove` field to the `File` in a node plan. This won't work for if the manifest is still updated when the Rancher version is updated. The file still needs to be "blanked" before adding the `Remove = true` field.

timhaneunsoo:
### Test Environment: ###
Rancher version: v2.6-head b919539
Rancher cluster type: HA
Docker version: 20.10

Downstream cluster type: RKE2 node driver cluster

----------

[...]

PR #37482, opened 2022-04-27
Stop deploying agent in planner after agent is deployed

thedadams:
Issue: https://github.com/rancher/rancher/issues/37465

## Problem
The RKE2 planner would include the manifest for the cluster agent in the
plan for a node. This causes the cluster to go into an updating state on
Rancher upgrade, which is unexpected.

## Solution
After this change, the planner will not include the cluster agent
manifest file. Instead, the cluster agent will be deployed via a periodic 
instruction that is only included until the cluster agent is deployed.

## Testing
Updating Rancher and rancher-agent with deployed RKE2 clusters.

### DIFF
--- a/pkg/apis/rke.cattle.io/v1/controlplane.go
+++ b/pkg/apis/rke.cattle.io/v1/controlplane.go
@@ -49,4 +49,5 @@ type RKEControlPlaneStatus struct {
 	ConfigGeneration              int64                               `json:"configGeneration,omitempty"`
 	Initialized                   bool                                `json:"initialized,omitempty"`
 	AgentConnected                bool                                `json:"agentConnected,omitempty"`
+	AgentDeployed                 bool                                `json:"agentDeployed,omitempty"`
 }
--- a/pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
+++ b/pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
@@ -80,6 +80,7 @@ func (h *handler) OnChange(obj *rkev1.RKEControlPlane, status rkev1.RKEControlPl
 	status.Ready = rke2.Ready.IsTrue(cluster)
 	status.Initialized = rke2.Ready.IsTrue(cluster)
 	status.AgentConnected = rke2.AgentConnected.IsTrue(cluster)
+	status.AgentDeployed = rke2.AgentDeployed.IsTrue(cluster)
 	return status, nil
 }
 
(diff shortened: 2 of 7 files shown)


Oats87 (changes requested):
Clarification around the desired behavior of the K3s/RKE2 deploy-controller needs to be made before this can be merged, as originally the intention was to allow for deleting manifests from a cluster by "blanking" a manifest file in the `/var/lib/<runtime>/server/manifests`. From a quick test, this does not appear to work today; however, we don't want to have future versions of K3s/RKE2 suddenly start behaving in this way and causing inadvertent orphaning of downstream clusters through deletion of the `cattle-cluster-agent` deployment.

@briandowns @galal-hussein @brandond if you folks have any insight into this, that would be much appreciated.

thedadams:
An issue has been create in the K3S repo: https://github.com/k3s-io/k3s/issues/5559
And a PR has been submitted to fix the issue pointed out by Chris: https://github.com/k3s-io/k3s/pull/5560

brandond:
Sounds like a reoccurrence of https://github.com/rancher/wrangler/issues/195

thedadams:
The approach of this solution has changed and the k3s issue is no longer apart of the discussion.

thedadams:
The current implementation of this solution is to include a periodic instruction with the plan until the `AgentDeployed` condition is true. This periodic instruction will check for the `cattle-cluster-agent` deployment in the `cattle-system` namespace. If the deployment doesn't exist, then it will try to create it (and not erroring if it is unsuccessful). Once the deployment is created, the periodic instruction will still run for a little bit, but won't try to overwrite the deployment. This is because Rancher's controllers will redeploy the agent as expected and this instruction shouldn't overwrite it.

The only iffy part of this implementation is dealing with clusters upgraded from v2.6.5 to v2.6.6. These older clusters will have the cluster-agent manifest saved to the manifest directory. In order to avoid RKE2/K3S redeploying these old manifests from the directory when the `rke2-server` service is restarted, then files should be deleted. This is done through a one-time instruction.

Right now, this one-time instruction is included with all control plane plans after the agent is deployed, forever. This is to ensure that the file is properly cleaned up. Personally, I wish there was some other way to clean this file up. The other possible solutions that I can think of for this would require significant changes (like updating the plan to delete the file on Rancher upgrade and pausing the `planner` until this plan is successfully applied by the system-agent). I am open to discussion and suggestion on how to make this second part better (and the first part above, but I feel much better about that).

thedadams:
Closing in favor of another implementation.

Issue #36748, opened 2022-03-03
RKE2 | Custom Windows 2022 cluster | Windows worker is "Unavailable" after snapshot restore


sgapanovich:
**Rancher Server Setup**
- Rancher version: v2.6-head 39befda
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE2
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.21.7 and v1.22.10
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Custom

   
**Describe the bug**

Windows worker node(s) becomes "Unavailable" with "Kubelet stopped posting node status." message 

**To Reproduce**
 
1. Provision RKE2 custom Windows 2022 cluster (1 linux etcd + cp, 1 linux worker + 2 Windows 2022 workers)
2. Wait for the cluster to provision
3. Navigate to Cluster Management 
4. Take Snapshot of the etcd (three dots menu on the cluster)
5. Wait for the snapshot to be created 
6. Restore "Restore Snaphot" -> Etcd only

**Result**
Tried 3 different clusters. In two of those only 1 windows node became unavailable but in the third cluster all two workers became unavailable.
Tried having nodes with separate roles (etcd and cp) - same result 

**Expected Result**

All nodes are in active and working state after restore

**Screenshots**

![unavailable](https://user-images.githubusercontent.com/40282141/156667427-71248597-e3e3-4bcd-a9e6-4bcdfc964c24.jpg)

kubelet is not running on the node
when trying to start it - it run for a few seconds
![kubelet](https://user-images.githubusercontent.com/40282141/156667556-9578232d-af63-4f31-8ee9-dfc929bd2e59.jpg)

**Additional context**

**when `kubectl get node`**
```
conditions:
  - lastHeartbeatTime: "2022-03-03T22:10:39Z"
    lastTransitionTime: "2022-03-03T22:12:40Z"
    message: Kubelet stopped posting node status.
    reason: NodeStatusUnknown
    status: Unknown
    type: MemoryPressure
  - lastHeartbeatTime: "2022-03-03T22:10:39Z"
    lastTransitionTime: "2022-03-03T22:12:40Z"
    message: Kubelet stopped posting node status.
    reason: NodeStatusUnknown
    status: Unknown
[...]

PR #37496, opened 2022-04-27
Going to try to skip Windows since it doesn't run a cluster agent.

phillipsj:
This prevents restarting the node because the cluster agent doesn't run on Windows making it not necessary.

* https://github.com/rancher/rancher/issues/36748

### DIFF
--- a/pkg/provisioningv2/rke2/planner/etcdrestore.go
+++ b/pkg/provisioningv2/rke2/planner/etcdrestore.go
@@ -70,7 +70,7 @@ func (p *Planner) runEtcdSnapshotRestorePlan(controlPlane *rkev1.RKEControlPlane
 	return ErrWaiting("failed to find etcd node to restore on")
 }
 
-// generateRestoreEtcdSnapshotPlan returns a node plan that contains instructions to stop etcd, remove the tombstone file (if one exists), then restore etcd in that order.
+// generateEtcdSnapshotRestorePlan returns a node plan that contains instructions to stop etcd, remove the tombstone file (if one exists), then restore etcd in that order.
 func (p *Planner) generateEtcdSnapshotRestorePlan(controlPlane *rkev1.RKEControlPlane, snapshot *rkev1.ETCDSnapshot, tokensSecret plan.Secret, server *planEntry, joinServer string) (plan.NodePlan, error) {
 	if controlPlane.Spec.ETCDSnapshotRestore == nil {
 		return plan.NodePlan{}, fmt.Errorf("ETCD Snapshot restore was not defined")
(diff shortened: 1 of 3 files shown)


rosskirkpat (dismissed):

phillipsj:
Related issue:

* https://github.com/rancher/rancher/issues/37499

pkg/provisioningv2/rke2/planner/etcdrestore.go
```diff
@@ -175,6 +175,9 @@ func (p *Planner) runEtcdRestoreControlPlaneEtcdServiceStop(controlPlane *rkev1.
 	servers := collect(clusterPlan, anyRole)
 	updated := false
 	for _, server := range servers {
+		if server.Machine.Status.NodeInfo.OperatingSystem == windows {
```
thedadams, line 178: Instead of doing this here, can we use a different function besides `anyRole` to the collect function above that filters out Windows nodes?

That would be cleaner, in my opinion.

thedadams: In addition, and I know this isn't part of your change, but could we change the name of the encapsulating function so that it is more descriptive of what is happening here? Specifically, we are dealing with all non-Windows machines at the moment. Not only control plane and etcd.

thedadams (changes requested):

pkg/provisioningv2/rke2/planner/etcdrestore.go
```diff
@@ -151,9 +151,9 @@ func generateCreateEtcdTombstoneInstruction(controlPlane *rkev1.RKEControlPlane)
 	}
 }
 
-// runEtcdRestoreControlPlaneEtcdServiceStop generates service stop plans for every etcd and controlplane node in the cluster and
+// runEtcdRestoreClusterAgentStopPlan generates service stop plans for every etcd/controlplane and all nodes running a cluster agent in the cluster and
 // assigns/checks the plans to ensure they were successful
-func (p *Planner) runEtcdRestoreControlPlaneEtcdServiceStop(controlPlane *rkev1.RKEControlPlane, snapshot *rkev1.ETCDSnapshot, tokensSecret plan.Secret, clusterPlan *plan.Plan) error {
+func (p *Planner) runEtcdRestoreClusterAgentStopPlan(controlPlane *rkev1.RKEControlPlane, snapshot *rkev1.ETCDSnapshot, tokensSecret plan.Secret, clusterPlan *plan.Plan) error {
```
Oats87, line 156: nit: the name of this function should probably be something like `runEtcdRestoreServiceStop`

the name of the plan right now implies it is going to stop the cluster agent, at least to me. While it does this, it also does a bunch of other things.

pkg/provisioningv2/rke2/planner/etcdrestore.go
```diff
@@ -172,7 +172,7 @@ func (p *Planner) runEtcdRestoreControlPlaneEtcdServiceStop(controlPlane *rkev1.
 			return err
 		}
 	}
-	servers := collect(clusterPlan, anyRole)
+	servers := collect(clusterPlan, anyRoleWithoutWindows)
```
thedadams, line 175: nit: for now could this just be `notWindows`?

phillipsj: I already created a notWindows function and this one is more descriptive as its the same as the anyRole but without Windows.

Oats87 (approved):

thedadams (approved):

Issue #37449, opened 2022-04-22
Unable to create a VMware vSphere cluster with private registry


suryatejaboorlu:
**Rancher Server Setup**
- Rancher version: 2.6.4
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster, and version (RKE1, RKE2, k3s, EKS, etc):RKE1 v1.20.5
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: 1.20.15-rancher1-1
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): VMware vSphere

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom): Admin

   
**Describe the bug**
Unable to create a VMware vSphere cluster with a private registry after upgrading the Rancher from 2.6.3 to 2.6.4.  
**To Reproduce**
1. Create a Rancher 2.6.4 cluster
2. Create an RKE template with a private registry and add the private registry ca.crt in the cloud-init file.
3. Try to provision the vSphere cluster with the RKE template.

**Result**

**Expected Result**
Rancher should provision the downstream vSphere cluster as it did in 2.6.3 with the same RKE template.

**Logs**
[1] Unable to find image 'docker-opc-ecv-group-cicd-nexus.rp-ocn.apps.ocn.infra.ftgroup/rancher/rancher-agent:v2.6.4' locally docker: Error response from daemon: Head "[https://docker-opc-ecv-group-cicd-nexus.rp-ocn.apps.ocn.infra.ftgroup/v2/rancher/rancher-agent/manifests/v2.6.4":](https://docker-opc-ecv-group-cicd-nexus.rp-ocn.apps.ocn.infra.ftgroup/v2/rancher/rancher-agent/manifests/v2.6.4%22:) no basic auth credentials. See 'docker run --help'. exit status 125 : exit status 1

**Additional context**
The RKE template included private registry details with authentication, but the docker needs the private registry ca.crt file in /etc/ssl/certs/ca-certificates.crt.rancher of the downstream nodes to authenticate the private registry and pull the images. if we copy the ca.crt file manually, the images are pulled and the cluster is provisioning fine. 

SURE-4527

phillipsj:
Adding this issue:

* https://github.com/rancher/rancher/issues/27166

[...]

PR #37504, opened 2022-04-28
[2.6] Fix the "no basic auth credentials" error when using private registry 

jiaqiluo:
**Issue:** https://github.com/rancher/rancher/issues/37449

**Problem:**
When rancher provisions an RKE node-driver cluster, one of the steps is to authenticate the machine to a private registry if one is defined in the cluster's spec. A [recent change](https://github.com/rancher/rancher/pull/36578) of migrating the private registry's password to a secret makes Rancher think there is no private registry in the cluster's spec, so it skips authenticating step.  As a result, the error "no basic auth credentials" is returned.

**Solution:**
The fix is to add the password to the cluster's spec before the authenticating step. The cluster object is used only within the function, so there should be no negative impact on security.

**Testing:**
Although I cannot get a setup to test the fix by creating a node-driver vSphere cluster, the following tests should indicate the fix is actually working in theory:

Steps:
- prepare a private registry that is installed with self-signed certs in an ec2 instance 
- bake an AMI that trusts the above certs for Docker 
- in another ec2 instance, run rancher 2.6.3
- in local, run rancher with my fix in GoLand
- create a node template, choose AWS, set AMI
- in both setups, create an RKE node-drive cluster, choose AWS EC2, set the private registry URL, username, and password 
 
Results:
- in rancher 2.6.3, hit the `no basic auth credentials` as the issue says 
 

- in my local setup, cluster is provisioned. Notice the following message in the logs which indicates the private registry is detected. 
`2022/04/27 17:39:22 [INFO] [node-controller-rancher-machine] private registry detected, authenticating jiaqi-amiaaaa1 to ec2-x-x-x-x.us-west-1.compute.amazonaws.com`
 

----
Need to back-port the fix to 2.5 and 2.4

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rancher/rancher/pkg/clustermanager"
 	"github.com/rancher/rancher/pkg/controllers/dashboard/clusterregistrationtoken"
 	"github.com/rancher/rancher/pkg/controllers/management/drivers/nodedriver"
+	secretmigrator "github.com/rancher/rancher/pkg/controllers/management/secretmigrator"
 	"github.com/rancher/rancher/pkg/encryptedstore"
 	corev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
@@ -413,6 +414,10 @@ func (m *Lifecycle) deployAgent(nodeDir string, obj *v3.Node) error {
 		return err
 	}
 
+	cluster.Spec, err = secretmigrator.AssembleRKEConfigSpec(cluster, cluster.Spec, m.credLister)
+	if err != nil {
+		return err
+	}
 	err = m.authenticateRegistry(nodeDir, obj, cluster)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -413,6 +414,10 @@ func (m *Lifecycle) deployAgent(nodeDir string, obj *v3.Node) error {
 		return err
 	}
 
+	cluster.Spec, err = secretmigrator.AssembleRKEConfigSpec(cluster, cluster.Spec, m.credLister)
```
thedadams, line 417: We shouldn't be modifying the cluster spec, especially if the cluster gets accidentally saved.

thedadams (changes requested):

pkg/controllers/management/node/controller.go
thedadams: That doesn't it won't be saved in the future. We shouldn't be modifying the spec of objects.

jiaqiluo: Now we make a deep copy of the cluster and modified the coped one, so we can avoid the accidental save by other processes. 
Can you review the PR again?

thedadams (approved):
I have some reservations about this implementation based on modifying the cluster's spec in the Rancher controllers. However, implementing this in a more robust way would change quite a few things.

Deep-copying the cluster object will keep the changes from being persisted in the cache, and, since this is being down in a node controller, there should be no reason for this cluster object to get saved here. Therefore, with the time-crunch, I think this change is safe in its current implementation. Any changes to this controller in the future should ensure that this cluster object is not saved.

annablender (approved):
LGTM as a temporary security fix

Issue #37502, opened 2022-04-28
RKE2 cluster - Upgrade strategy for control plane is not adhered and cluster is stuck in Updating state


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6head commit id: `7db4fe4`
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1 local cluster - 1.23.6

**Information about the Cluster**
- Kubernetes version: `v1.21.10+rke2r2` RKE2 node driver Downstream cluster - 3 etcd, 2 cp and 3 worker node

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) admin

**Describe the bug**
RKE2 cluster - Upgrade strategy for control plane is not adhered and cluster is stuck in Updating state

**To Reproduce**
- Provision a cluster on `v1.21.10+rke2r2` k8s version
- Upgrade to `v1.22.7+rke2r2` k8s version
- Etcd finished upgrading.
- One of the Control planes started upgrading
- then another one also started upgrading at the same time, before the first one came up Active.  --> Upgrade strategy is not being adhered. (by default, Unavailable CP nodes - 1)
- And now cluster is in Updating state stuck.

**Result**
Provisioning Logs:
```


[INFO ] provisioning done
--
5:07:25 pm | [INFO ] draining bootstrap node(s) sowmya-abc-pool1-577b4f76b7-nh8vv: draining node
5:07:39 pm | [INFO ] configuring bootstrap node(s) sowmya-abc-pool1-577b4f76b7-nh8vv: waiting for plan to be applied
5:07:55 pm | [INFO ] configuring bootstrap node(s) sowmya-abc-pool1-577b4f76b7-nh8vv: waiting for probes: kubelet
5:08:15 pm | [INFO ] draining etcd node(s) sowmya-abc-pool1-577b4f76b7-fgcsq: draining node
5:08:17 pm | [INFO ] configuring etcd node(s) sowmya-abc-pool1-577b4f76b7-fgcsq,sowmya-abc-pool1-577b4f76b7-z265n
5:08:53 pm | [INFO ] draining etcd node(s) sowmya-abc-pool1-577b4f76b7-z265n: draining node
[...]

Issue #37501, opened 2022-04-28
RKE2 cluster - Nodes drain automatically after cluster has provisioned 


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6head commit id: `7db4fe4`
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1 local cluster - 1.23.6

**Information about the Cluster**
- Kubernetes version: `v1.21.10+rke2r2` RKE2 node driver Downstream cluster - 3 etcd, 2 cp and 3 worker node AWS

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) admin

   
**Describe the bug**
RKE2 cluster - Nodes drain automatically after cluster has provisioned 

**To Reproduce**

- Provisioned the cluster
- All nodes active, and then one by one worker nodes Drained and came back up Active
- Then followed by etcd and control plane nodes.

**Result**
Provisioning Logs
```


[INFO ] waiting for at least one bootstrap node
--
4:16:58 pm | [INFO ] configuring bootstrap node(s) sowmya-upgrade-3-pool1-55b664d5cb-k9rc7: waiting for agent to check in and apply initial plan
4:17:06 pm | [INFO ] configuring bootstrap node(s) sowmya-upgrade-3-pool1-55b664d5cb-k9rc7: creating server [fleet-default/sowmya-upgrade-3-pool1-3f6f2bdd-c29h4] of kind (Amazonec2Machine) for machine sowmya-upgrade-3-pool1-55b664d5cb-k9rc7 in infrastructure provider, waiting for agent to check in and apply initial plan
4:17:20 pm | [INFO ] configuring bootstrap node(s) sowmya-upgrade-3-pool1-55b664d5cb-k9rc7: waiting for agent to check in and apply initial plan
4:17:46 pm | [INFO ] configuring bootstrap node(s) sowmya-upgrade-3-pool1-55b664d5cb-k9rc7: waiting for probes: etcd, kubelet
4:18:14 pm | [INFO ] configuring bootstrap node(s) sowmya-upgrade-3-pool1-55b664d5cb-k9rc7: waiting for probes: etcd
4:18:22 pm | [INFO ] configuring bootstrap node(s) sowmya-upgrade-3-pool1-55b664d5cb-k9rc7: waiting for plan to be applied
[...]

Issue #37500, opened 2022-04-27
Worker node of RKE2 cluster is stuck in `Reconciling` state on a Linode cluster when `createPrivateIp: true`


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `7db4fe4`
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1

**Information about the Cluster**
- Kubernetes version: `v1.22.9-rc4+rke2r1` --> RKE2 downstream cluster - Linode RKE2 - 1 node all roles and 1 worker node

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Admin

   
**Describe the bug**
Worker node of RKE2 cluster is stuck in `Reconciling` state

**To Reproduce**

- Deploy an RKE2 linode cluster - 1 node all roles and 1 worker node
- Set `createPrivateIp: true` on the machine pools
- The worker node is stuck in Reconciling state
- Server logs

```
Apr 27 20:28:22 sowmya-linode-26head-pool1-029277cb-4nxh7 rke2[5531]: time="2022-04-27T20:28:22Z" level=error msg="runtime core not ready"
Apr 27 20:28:23 sowmya-linode-26head-pool1-029277cb-4nxh7 rke2[5531]: time="2022-04-27T20:28:23Z" level=info msg="Waiting to retrieve kube-proxy configuration; server is not ready: https://127.0.0.1:9345/v1-rke2/readyz: 500 Internal Server Error"
Apr 27 20:28:27 sowmya-linode-26head-pool1-029277cb-4nxh7 rke2[5531]: time="2022-04-27T20:28:27Z" level=info msg="Defragmenting etcd database"
Apr 27 20:28:27 sowmya-linode-26head-pool1-029277cb-4nxh7 rke2[5531]: time="2022-04-27T20:28:27Z" level=info msg="certificate CN=rke2,O=rke2 signed by CN=rke2-server-ca@1651088172: notBefore=2022-04-27 19:36:12 +0000 UTC notAfter=2023-04-27 20:28:27 +0000 UTC"
[...]

PR #37551, opened 2022-04-30
Fix planner address issues

thedadams:
Issues: 
https://github.com/rancher/rancher/issues/37500
https://github.com/rancher/rancher/issues/37501
https://github.com/rancher/rancher/issues/37502

There were several issues related to IP addresses that were causing issues with
RKE2 provisioning and multiple etcd-only nodes.

1. If there were multiple etcd-only nodes, then joinURL may change and be for
another node when the scraping happens. The reason was that the clientURLs would
be listed for each etcd node. When Rancher was scraping it, it was taking the
first one no matter what which may have been for another node.

The fix here is to add an addition periodic instruction to "cat" the db name
file and use it to identify the clientURL with the correct node.

2. The machine plan for the bootstrap node may have been created before the
machine had finished being created and the config saved to the machine state
secret. This would cause the plan to be missing IP addresses. Then, then next
plan would have the IP addresses, causing the cluster to look like it was
re-provisioning. If the user enabled the "drain on upgrade" feature, then the
nodes would all drain and their plan updated.

The fix here is to wait to assign a bootstrap node until we know that the
machine has finished creating, so that the machine state has the IP addresses,
if they are going to be available. Note that Rancher will not error if no IP
addresses are found because there may not be any depending on the provider.

3. Join URLs were being set in multiple places.

The fix here is to only set them in one place.

4. When getting the control plane join URL, Rancher was taking the URL from the
first control plane node that it found a join URL on. This could cause the join
URL to change after the node is bootstrapped.

The first control plane join URL is returned, regardless if it is set. This
ensures that join URL won't change after it is set on initial bootstrapping.

Additionally, the system-agent will not set the probes as failed on the first failure
if the failure threshold is greater than 1. This is proper and expected.

However, if the plan is updated, we don't want to wait for the
failure-threshold. If the probes pass on the first try, then
system-agent will update them as such and Rancher will move on. However,
if the probes fail on the first try after the plan is updated, then
system-agent won't set the probe as successful and Rancher will wait for
successful probes before moving on.

## Testing
Deployed clusters of various types in Linode to ensure the referenced issues were working as expected.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/config.go
+++ b/pkg/provisioningv2/rke2/planner/config.go
@@ -22,6 +22,7 @@ import (
 	"github.com/rancher/wrangler/pkg/kv"
 	"github.com/rancher/wrangler/pkg/yaml"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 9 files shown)


pkg/provisioningv2/rke2/planner/config.go
```diff
@@ -276,21 +277,30 @@ func addToken(config map[string]interface{}, entry *planEntry, tokensSecret plan
...
-func addAddresses(secrets corecontrollers.SecretCache, config map[string]interface{}, entry *planEntry) {
+func addAddresses(secrets corecontrollers.SecretCache, config map[string]interface{}, entry *planEntry) error {
 	internalIPAddress := entry.Metadata.Annotations[rke2.InternalAddressAnnotation]
 	ipAddress := entry.Metadata.Annotations[rke2.AddressAnnotation]
 	internalAddressProvided, addressProvided := internalIPAddress != "", ipAddress != ""
 
-	secret, err := secrets.Get(entry.Machine.Spec.InfrastructureRef.Namespace, rke2.MachineStateSecretName(entry.Machine.Spec.InfrastructureRef.Name))
-	if err == nil && len(secret.Data["extractedConfig"]) != 0 {
+	// If this is a provisioned node (not a custom node), then get the IP addresses from the machine driver config.
+	if entry.Machine.Spec.InfrastructureRef.APIVersion == rke2.RKEMachineAPIVersion {
```
Oats87, line 286: nit: we can probably short circuit retrieving/parsing the config JSON if `internalAddressProvided && addressProvided`

pkg/provisioningv2/rke2/planner/store.go
```diff
@@ -319,17 +319,15 @@ func (p *PlanStore) UpdatePlan(entry *planEntry, plan plan.NodePlan, maxFailures
...
 	}
 
 	rke2.CopyPlanMetadataToSecret(secret, entry.Metadata)
 
-	// if there are no probes, clear the statuses of the probes to prevent false positives
-	if len(plan.Probes) == 0 {
-		delete(secret.Data, "probe-statuses")
-	}
+	// If the plan is being updated, then delete the probe-statuses so their healthy status will be reported as healthy only when they pass.
+	delete(secret.Data, "probe-statuses")
```
Oats87, line 330: With this change, if the probe status success thresholds are not set to `1`, we will have false positives on failures. Not that this matters right now because I don't think we currently have any probes that have successThreshold > 1 but still, something to keep in mind. https://github.com/rancher/system-agent/blob/v0.2.5/pkg/prober/prober.go#L121-L123

Oats87 (approved):
LGTM

Oats87 (approved):

jakefhyde (approved):

pkg/provisioningv2/rke2/planner/config.go
thedadams: Done

Issue #37193, opened 2022-04-05
loadbalance IP not assign to rancher ingress-rule


kittisuw:
rancher v2.6.4
rke version v1.2.19
Kubernetes v1.20.15
ingress-nginx v.4.0.13

I was installed rancher on the RKE cluster as command below I don't know why default ingress rule name "rancher" that rancher create does not assign public IP from nginx-ngress controller.pls help I don't need to add the ingress rule anymore.

[ingress-nginx]
```shell
#Install NGINX Ingress Controller
helm install ingress-nginx ingress-nginx/ingress-nginx \
    --version 4.0.13 \
    --namespace ingress-nginx --create-namespace \
    -f ingress-nginx/internal-ingress.yaml \
    --set controller.replicaCount=2 \
    --set controller.nodeSelector."beta\.kubernetes\.io/os"=linux \
    --set defaultBackend.nodeSelector."beta\.kubernetes\.io/os"=linux \
    --set controller.admissionWebhooks.patch.nodeSelector."beta\.kubernetes\.io/os"=linux
#Check running pod
k get po -n ingress-nginx
k get svc -A
...
NAMESPACE             NAME                                 TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)                      AGE
ingress-nginx         ingress-nginx-controller             LoadBalancer   10.43.50.73     192.168.40.183   80:32070/TCP,443:32263/TCP   33m
...
```
[Racher]
```shell
#Add the Helm Chart Repositorylink
helm repo add rancher-stable https://releases.rancher.com/server-charts/stable
#Update your local Helm chart repository cache
helm repo update
#Create a Namespace for Rancher
kubectl create namespace cattle-system
#Install Rancher with Helm
helm install rancher rancher-stable/rancher \
  --version 2.6.4 \
  --namespace cattle-system \
  --set hostname=rancher.test.co.th \
  --set replicas=3
#Verify that the Rancher Server is Successfully Deployed
kubectl -n cattle-system rollout status deploy/rancher
kubectl -n cattle-system get deploy rancher
```
As command above rancher will create ingress rule below but loadbalance IP not assign
```
k get ing
NAME      CLASS    HOSTS                      ADDRESS   PORTS     AGE
rancher   <none>   rancher.test.co.th             80, 443   11m
```
till I apply config below
```shell
#rancher-ingress.yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: rancher-ing
[...]

PR #37562, opened 2022-05-02
Add optional ingressClassName support to Rancher helm chart

bashofmann:
Even though ingressClassName is an optional field, some ingress controllers are only watching ingresses with a certain ingressClassName and skip ingresses without. For example the upstream ingress-nginx which by default sets `controller.watchIngressWithoutClass=false`.

Addresses https://github.com/rancher/rancher/issues/37193


bashofmann:
Still relevant.

aslafy-z:
Implemented by https://github.com/rancher/rancher/pull/38202
This one should be closed.

PR #37566, opened 2022-05-02
Adding CRD for RancherUserNotifications

MbolotSuse:


MbolotSuse:
## Background

This introduces a new resource type - RancherStatus. The idea behind this resource type is to provide a mechanism for components of rancher to report a status which needs to be displayed to the user. In the case of the csp-adapter for aws, this can be used to inform users that they are out of compliance with the license. 

## Changes
- Introduces a new CRD - RancherStatus
- RancherStatus has the following fields:
  - ComponentName - the name of the component which this status refers to
  - ComponentStatus - if the status indicates an error (false) or standard operation (true). This could also be expanded to include more possible options if desired
  - Reason - the reason that this component is in this status
- RancherStatus is not a namespaced resource
- All typical types of users (base user, standard user, restricted admin) have permission to read these resources ("get", "list", "watch")
- The admin user gets access to these resources by virtue of it's standard cluster role binding

## Notes
A new CRD was chosen over using configmaps for the following reasons:
- Configmaps are namespaced. Granting users access to these resources would require one of two changes:
  - In a controller, we map users who have access to one of the typical user roles to a rolebinding in the target namespace allowing them to read the status map. This has significant performance impacts (one new rb per user + controller overhead)
  - We grant access in a clusterrolebinding to configmaps in all namespaces with a specific name (to limit impact to only the results produced by the CSP adapter). This would create a potentially vulnerability if a configmap name was re-used unknowingly 
  
  
This is set to DNM until the changes to the csp-adapter are nearing complete.

cmurphy (commented):
A couple of general thoughts:

1) Rancher Alerts are deprecated, the legacy feature has to be enabled in order to see them and a deprecation notice is visible on the page. We should not be adding a new kind of Alert. It's also not clear to me what an Alert has to do with supportconfigs. Alerts are for monitoring specific resources and taking action, like sending an email notification, when certain conditions are met. Can we do this with either a completely new CRD or as an action on v3.Cluster (where it would only work on the local cluster)? I haven't yet looked into how this might work with the steve aggregation API, if you haven't already ruled that out I'll try to dig into whether that could apply here.

2) I thought the reason we were adding a CRD and a norman action was both to simplify authentication (:white_check_mark:) but also to avoid using the nonResourceURL RBAC check. But I still see that as part of the the authorize method.

MbolotSuse:
@cmurphy 

1.

 From what I could see, the Alert objects were deprecated but they were replaced by the AlertRule objects (i.e. ProjectAlerts were superseded by ProjectAlertRules). If I'm wrong there please let me know. 

As far as the purpose here - the overall goal is to allow the CSP adapter to signal to the end user that they haven't bought enough licenses and are out of compliance with their support contract. This has the requirement of being visible to all users (admin, restricted admin, standard user, and base user). I thought that the best way to support this would be a new CRD, which would be non-namespaced (and so we could easily add a read permission to the global roles). The reason that I thought about a generic CRD was that this could be useful in other areas of the code (i.e. a serviceAccount used for auth expired or was about to expire - throw an alert so users know to fix it). 

As far as what this has to do with the SupportConfig endpoint - the endpoint does not interact with the alerts at all. Right now the SupportConfig stuff is closely tied with the csp adapter (and they have some interoperability as you can see in this PR). I can separate this out into another commit or another PR if you would like.

2. 

You are correct. I was told that I should not add another CRD if it can be avoided, and was told not to use norman as we are trying to move away from it to steve. Without the ability to create a CRD or add a norman action, I did not have anything that would allow me to control access to this endpoint outside of a non-resource url. Basically, I can either add a CRD/Action (I might be able to do this with Steve but I'm not 100% sure on that), or I can avoid adding a CRD but then authz has to use a non-resource url (at least that's what I see - open to other ideas). If you prefer the CRD route let me know and I can try to get a team discussion going - that's what I originally implemented before being told to not do that.

cmurphy:
> From what I could see, the Alert objects were deprecated but they were replaced by the AlertRule objects (i.e. ProjectAlerts were superseded by ProjectAlertRules). If I'm wrong there please let me know.

The only way I can see to create a v3.ClusterAlertRule is to enable the legacy feature, go to Legacy -> Alerts, and go to Add Rule, which calls /v3/clusteralertrules. So I assumed these were part of the deprecated feature. I could be wrong here. If there's some existing agreement on using AlertRules for CSP then I won't push back.

> As far as the aggregation api goes I did look into that in the context of point 2 (support config), but not in the context of alerts. How were you thinking we could use that for alerts?
...
> As far as what this has to do with the SupportConfig endpoint - the endpoint does not interact with the alerts at all. Right now the SupportConfig stuff is closely tied with the csp adapter (and they have some interoperability as you can see in this PR). I can separate this out into another commit or another PR if you would like.

Sorry, yes that was my confusion. We talked about the other CSP PR in the context of the supportconfig generator, so I thought this was tied in with that. If these are completely separate ideas then a separate PR would be very helpful.

> You are correct. I was told that I should not add another CRD if it can be avoided,

Sorry again, my confusion was that this PR was adding a new type and also adding the supportconfig route and I thought they were related to each other. I agree it would be best to avoid adding a CRD for this supportconfig case if possible.

I had a quick look at some of the steve code and there are a couple of ideas there. The AdditionalAPIs code could be a good place for the new route to live, it seems to be a kind of home for misfit endpoints:

https://github.com/rancher/rancher/blob/c4d59fa88e2313e1446ba542a21c39e15a17f375/pkg/api/steve/additionalapi.go#L42

There are also some interesting authz ideas that could be borrowed from, where the existence of some token or secret authorizes the user for an action, and there's no need for a k8s-style SAR or Role:

https://github.com/rancher/rancher/blob/c4d59fa88e2313e1446ba542a21c39e15a17f375/pkg/api/steve/aggregation/auth.go#L50
https://github.com/rancher/rancher/blob/c4d59fa88e2313e1446ba542a21c39e15a17f375/pkg/api/steve/proxy/auth.go#L34

If you don't think those apply here then I'm in favor of the nonResourceURL SAR.

MbolotSuse:
> The only way I can see to create a v3.ClusterAlertRule is to enable the legacy feature, go to Legacy -> Alerts, and go to Add Rule, which calls /v3/clusteralertrules. So I assumed these were part of the deprecated feature. I could be wrong here. If there's some existing agreement on using AlertRules for CSP then I won't push back.

No existing agreement. Here's what I will do - I'll change the CRD to a "RancherUserNotification" crd, and with the following fields:

```
Component string // the component that was the source of the notification
Message string // the message for this notification
```
Let me know if you have any objections to the above.

> Sorry, yes that was my confusion. We talked about the other CSP PR in the context of the supportconfig generator, so I thought this was tied in with that. If these are completely separate ideas then a separate PR would be very helpful.

Sorry for the confusion - the supportconfig has been tied in with the CSP adapter functionality from the start (same delivery timeline and CSP adapter needs supportconfig to work properly) even though they are separate. I should have been more clear about the separation.

For clarity's sake, I will open another PR. The current PR will be for RancherNotifications. The other PR will be for the supportconfig endpoint.

> I had a quick look at some of the steve code and there are a couple of ideas there. The AdditionalAPIs code could be a good place for the new route to live, it seems to be a kind of home for misfit endpoints:

I originally intended to use these paths - however, they are not protected by rancher's auth system (the github endpoint has it's own custom method of auth). If I put it here I would need to re-implement our authn functionality to some degree, so I would prefer to avoid that. 

I had not seen the other two examples before now. I need to give them a closer look - that being said, both examples seem to be using a different type of token (i.e. serviceAccount and clusterRegistrationToken) to grant access. That wouldn't really work here - I need to allow users to access the endpoint directly. I could proxy the endpoint through rancher and attach a service account, but that would probably require this to not be inside of rancher and would likely result in a more complex solution.

cmurphy:
:+1: to everything

pkg/multiclustermanager/routes.go
```diff
@@ -68,6 +69,7 @@ func router(ctx context.Context, localClusterEnabled bool, tunnelAuthorizer *mcm
 
 	channelserver := channelserver.NewHandler(ctx)
 
+	supportConfigGenerator := supportconfigs.NewGeneratorHandler(scaledContext)
```
cbron, line 72: Pretty sure scaled context is correct here, since any of the rancher pods could get the api hit, not just the leader.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+func (h *GeneratorHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	err := h.authorize(request)
+	if err != nil {
+		switch err.Error() {
+		case unauthenticatedError:
+			writer.WriteHeader(http.StatusUnauthorized)
+			_, _ = writer.Write([]byte(err.Error()))
```
cbron, line 51: Lot of this style of code exists in Norman already, though given this is a standalone API endpoint it's not necessary to use it: 
https://github.com/rancher/rancher/blob/release/v2.6/pkg/api/norman/customization/pipeline/pipeline.go#L68

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if !response.Status.Allowed {
+		return fmt.Errorf(forbiddenError)
+	}
+	return nil
+}
+
+// generateSupportConfig produces an io.Reader with the supportConfig ready to be returned using a http.ResponseWriter
+// Returns an err if it can't get the support config (note: does not return an err if it can't find the csp config)
+func (h *GeneratorHandler) generateSupportConfig() (io.Reader, error) {
```
cbron, line 105: Long term we should include pod logs in here. Should be able to read in with k8s client. Not necessary right now.

cbron (commented):
Agree with movement away from AlertRule, worried that could be confusing or misconstrued. Or if we do ever end up converting that type it would break this. 
Hey if it's not a hassle can you push your generated code into a secondary commit, that way you can look through commits easier to review.

MbolotSuse:
@cmurphy I've separated these out as requested and adjusted the CRDs as discussed. This PR now only adds the the RancherUserNotification Crd. I also took @cbron 's suggestion (thanks) to separate the commits into a non-generated and generated commit. I will update the description of the PRs accordingly.

cbron (approved):

cmurphy (approved):

Issue #37147, opened 2022-04-01
Generating a kubeconfig for downstream cluster fails when cluster-agent is disconnected, even when token hashing is disabled


thedadams:
**Is your feature request related to a problem? Please describe.**
When ACE is enabled for a downstream cluster, generating a kubeconfig for the cluster will fail when the cluster-agent is disconnected. This is due to a change for token hashing. However, generating a kubeconfig when the cluster-agent is disconnected will still fail if token hashing is not enabled.

**Describe the solution you'd like**
If token hashing is not enabled, then generating a kubeconfig for a cluster with ACE enabled should not fail.

**Additional context**
SURE-3334

thedadams:
### Root cause
Rancher would try to create a ClusterAuthToken in the downstream cluster in the API layer. This would cause API to return a bad status code if the cluster was not active.

### What was fixed, or what changes have occurred
Now, when token hashing is disabled (which it is by default), the ClusterAuthToken is not created in the API layer. Instead, it is created by a controller later when the cluster is ready.

### Areas or cases that should be tested
1. Start Rancher with token hashing disabled (it is disabled by default)
2. Provision an RKE1 cluster from Rancher
3. While the cluster is provisioning, create a kubeconfig for the cluster. This will have to be done through the API.
4. The kubeconfig should be successfully generated.

1. Start Rancher and enable token hashing. This can be enabled after Rancher starts by editing the feature `kubectl edit feature token-hashing` or when starting Rancher by passing `--features token-hashing=true`
2. Provision an RKE1 cluster from Rancher.
3. Wait for the cluster to be active.
4. Download a kubeconfig for the cluster (this should be possible through the UI).

All contexts should be tested when the clusters are active in both cases.

### What areas could experience regressions?
Generating kubeconfigs for clusters when token hashing is enabled in the Rancher cluster.

### Are the repro steps accurate/minimal?
Yes.

markusewalker:
Verified this has been addressed on `v2.6-head` https://github.com/rancher/rancher/commit/25381f9b0a1868e25ea6b0a33ac9e545257d2b80.

**ENVIRONMENT DETAILS**
- Rancher HA: Helm Install
- Rancher versions
     - Reproduced version: `v2.6.4`
     - Verified version: `v2.6-head-25381f9`
- Browser: Chrome

**TEST RESULT**
PASS

**REPRODUCTION STEPS**
1. Setup Rancher  `v2.6.4` and navigate to Rancher UI in a browser.
[...]

PR #37568, opened 2022-05-03
Stop creating ClusterAuthToken on kubeconfig generation if token hashing disabled

thedadams:
Issue: https://github.com/rancher/rancher/issues/37147

## Problem
A previous change added token hashing implementation. When token hashing is
enabled, then ClusterAuthToken must be created in the downstream cluster when
the kubeconfig is generated. This means that the kubeconfig cannot be generated
until the cluster is active. If token hashing is disabled, then this requirement
is not needed and the kubeconfig can be generated without requiring that the
ClusterAuthToken be created in the downstream cluster.

## Solution
This change will check the token hashing feature flag when generating a
kubeconfig and will not create the ClusterAuthToken. In addition, the
controller that would create the ClusterAuthToken in the downstream cluster is
restored to create the ClusterAuthToken if token hashing is disabled.

## Testing
Creating kubeconfigs for clusters with token-hashing enabled and disabled with the cluster in various states (Provisioning, Updating, Active).

### DIFF
--- a/pkg/api/norman/customization/cluster/actions_kubeconfig.go
+++ b/pkg/api/norman/customization/cluster/actions_kubeconfig.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rancher/rancher/pkg/auth/tokens"
 	mgmtclient "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	"github.com/rancher/rancher/pkg/controllers/managementuser/clusterauthtoken/common"
+	"github.com/rancher/rancher/pkg/features"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/kubeconfig"
 	"github.com/rancher/rancher/pkg/settings"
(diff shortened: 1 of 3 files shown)


pkg/api/norman/customization/cluster/actions_kubeconfig.go
```diff
@@ -64,26 +65,28 @@ func (a ActionHandler) GenerateKubeconfigActionHandler(actionName string, action
...
+					return err
+				}
+				clusterAuthToken, err := common.NewClusterAuthToken(token, tokenValue)
+				if err != nil {
+					return err
+				}
+
+				if _, err = clusterClient.Cluster.ClusterAuthTokens("cattle-system").Create(clusterAuthToken); err != nil {
+					return err
+				}
```
rmweir, line 89: Can you put this into a single function for readability. Can do L75-89 instead if preferred.

rmweir (changes requested):

pkg/api/norman/customization/cluster/actions_kubeconfig.go
thedadams: Updated.

rmweir (approved):
LGTM

jakefhyde (approved):

Issue #36607, opened 2022-02-23
Upgrade of downstream cluster does not start if other (completed) plans exist in the cluster


Martin-Weiss:
We are using system-upgrade-controller to patch and upgrade the OS and due to that our downstream clusters have system-upgrade-controller and plans deployed, already before upgrading RKE2. The "other plans and other system-upgrade-controllers are NOT in the cattle-system namespace2.

It seems that rancher looks into ALL namespaces for Plans 

https://github.com/rancher/rancher/blob/d5a943b427776922c6f163623738bbda3e12e9b2/pkg/controllers/management/k3sbasedupgrade/deployPlans.go#L58

and if it finds one - we see this error in the rancher logs:

`022/02/23 08:10:58 [ERROR] error syncing 'c-jrh4v': handler k3s-upgrade-controller: an empty namespace may not be set when a resource name is provided, requeuing`

-> any idea why Rancher checks for plans in "all" namespaces?

-> could we get this "fixed" in a way that Rancher just looks into "cattle-system" namespace for other plans? 

SURE-4141

jakefhyde:
Moved to Blocked because the issue does not seem reproducible with _just_ completed plans in separate namespaces, waiting for additional info.

markusewalker:
Verified this has been resolved in `v2.6-head` https://github.com/rancher/rancher/commit/0e2fdc8060bc74c278d7f043c7846297577fb92c.

**ENVIRONMENT DETAILS**
- Rancher HA
- Rancher version
     - Reproduced version: `v2.6.4`
     - Verified version: `v2.6-head https://github.com/rancher/rancher/commit/0e2fdc8060bc74c278d7f043c7846297577fb92c`
- Browser: Chrome

**TEST RESULT**
PASS

**REPRODUCTION STEPS**
1. Setup Rancher `v2.6.4` and navigated to Rancher UI in a browser.
2. Created a standard user and logged in to Rancher as that user.
3. On a separate client machine, provisioned an k3s cluster at version v1.21.12+k3s1.
4. Ran the following command to install the system-upgrade-controller: `kubectl apply -f https://github.com/rancher/system-upgrade-controller/releases/latest/download/system-upgrade-controller.yaml`
5. Configured a server and agent plan into the resulting `system-upgrade` namespace; verified the K8s cluster was updated to the version specified in the plan.
6. Navigated back to Rancher and imported the provisioned k3s cluster.
[...]

PR #37576, opened 2022-05-03
[2.6.6] Only check cattle-system for existing plans

jakefhyde:
### Root cause

Rancher was checking all namespaces for plans, and trying to update ones not currently managed by rancher in order to force rancher to manage them. By reusing the same client (created using `metav1.NamespaceAll`, the generated client (wrangler uses the kubernetes client generator) uses the namespace of the client in the API request, which is always empty so the object cannot be updated. 

### What was fixed, or what changes have occurred

Now rancher will create a separate client for the namespace it finds existing plans in, and disables all plans not managed by rancher.

### Areas or cases that should be tested

Creating plans manually in any namespace should result in them being disabled, unless they are explicitly managed by rancher.
 
### What areas could experience regressions?

Currently, all non rancher-managed plans will be disabled once an upgrade is started, due to uncertainty around what other plans could exist and possible implications they may have during an upgrade.

### Are the repro steps accurate/minimal?

- Deploy suc in a custom cluster in any namespace other than cattle-system
- Create a plan in this namespace
- When the custom cluster is upgraded, all other plans should be disabled

Related Issue: https://github.com/rancher/rancher/issues/36607

SURE-4141

### DIFF
--- a/pkg/controllers/management/k3sbasedupgrade/deployPlans.go
+++ b/pkg/controllers/management/k3sbasedupgrade/deployPlans.go
@@ -73,7 +73,7 @@ func (h *handler) deployPlans(cluster *v3.Cluster, isK3s, isRke2 bool) error {
 			}
 			plan.Spec.NodeSelector.MatchExpressions = append(plan.Spec.NodeSelector.MatchExpressions, lsr)
 
-			_, err = planClient.Update(context.TODO(), &plan, metav1.UpdateOptions{})
+			_, err = planConfig.Plans(plan.Namespace).Update(context.TODO(), &plan, metav1.UpdateOptions{})
 			if err != nil {
 				return err
 			}


jiaqiluo (approved):
LGTM

jiaqiluo:
@jakefhyde  Have you got this issue reproduced?

jakefhyde:
@jiaqiluo Yes, it looks like trying to use the same client created with "" to specify all namespaces will succeed with getting and listing, but fail with other CRUD commands due to how the kubernetes api request gets generated, and this appears to be the expected behavior, but as far as I can tell, there is no really need for us to disable all plans in other namespaces on upgrade.

jiaqiluo:
@jakefhyde  OK, that sounds good to me. I just want to confirm we have steps to reproduce the bug for QA. 
 If possible, could you update the description to follow the PR template?

pkg/controllers/management/k3sbasedupgrade/deployPlans.go
```diff
@@ -55,7 +55,7 @@ func (h *handler) deployPlans(cluster *v3.Cluster, isK3s, isRke2 bool) error {
 	if err != nil {
 		return err
 	}
-	planClient := planConfig.Plans(metav1.NamespaceAll)
+	planClient := planConfig.Plans("cattle-system")
```
thedadams, line 58: The original intent of this seems to have been to disable all plans in all namespaces. I agree that is seems weird that this may have been necessary, but I don't want to second-guess that choice.

Could another possible solution be to get all upgrade plans, but do something like `planConfig.Plans(plan.Namespace).Update(context.TODO(), &plan, metav1.UpdateOptions{})` when updating them on line 76? This is the way the `planClient` is changed on line 128, for example.

Perhaps, even, this line could be removed and line 60 could be updated similarly to get all plans.

thedadams (changes requested):

pkg/controllers/management/k3sbasedupgrade/deployPlans.go
jakefhyde: Yeah that works. I think you're right, the original intent is lost here and someone could be relying on this behavior that we aren't aware of.

jakefhyde: Yeah that works. I think you're right, the original intent is lost here and someone could be relying on this behavior that we aren't aware of.

thedadams (approved):

jiaqiluo (approved):

thedadams (approved):

Issue #37587, opened 2022-05-04
[RKE2] When etcd components certs are rotated, all the nodes - etcd/control-plane/worker nodes drain/reconcile after the certs are rotated


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id: `a31f5b861`
- Installation option (Docker install/Helm Chart): Helm install
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1

**Information about the Cluster**
- Kubernetes version: 1.22-RKE2
- Cluster Type (Local/Downstream): AWS node driver RKE2 - 3 etcd, 2 cp and 3 worker nodes

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) std user - cluster owner

   
**Describe the bug**
[RKE2] When etcd components certs are rotated, all the nodes - etcd/control-plane/worker nodes drain/reconcile after the certs are rotated

**To Reproduce**
- Deploy an rke2 node driver down stream cluster
- Rotate certs on the cluster, select etcd component from the dropdown.
- certificate rotation happen on all the nodes including control plane and worker nodes as seen in cluster's provisioning logs
- All nodes drain/reconcile/uncordon after this 
- Provisioning Logs
```
[INFO ] waiting for [sowm-upgrade-pool1-7ccff5f4b4-pfvv5] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool1-7ccff5f4b4-s94z2] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool1-7ccff5f4b4-znwl4] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool2-c75996d95-44pq6] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool2-c75996d95-z8d4g] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool3-784b8c5fdc-msffp] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool3-784b8c5fdc-rbxmp] certificate rotation
[INFO ] waiting for [sowm-upgrade-pool3-784b8c5fdc-zf8kk] certificate rotation
[INFO ] draining bootstrap node(s) sowm-upgrade-pool1-7ccff5f4b4-s94z2: draining node
[...]

PR #37589, opened 2022-05-04
certificate rotation only rotates certificates on applicable nodes

paynejacob:
#37587 
#37588

# Problem
When rotating certificates on k3s or rke2 all nodes would be updated regardless of if they were running the relevant service.

# Solution
Only call certificate rotation on nodes when the specified services are related to the node's roles.

# Testing

I tested rotating etcd and only saw etcd and all role nodes updated.  There is a unit test to confirm the behavior for other combinations.


jiaqiluo (approved):
LGTM 💯

jiaqiluo (approved):

thedadams (approved):

Issue #29306, opened 2020-10-01
[RFE] Update Azure AD API from ADAL to MSAL


oxr463:
**What kind of request is this (question/bug/enhancement/feature request):**

Bug

**Steps to reproduce (least amount of steps as possible):**

Configure Azure AD for auth, (See: https://rancher.com/docs/rancher/v2.x/en/admin-settings/authentication/azure-ad).

**Result:**

Rancher is using the deprecated Azure AD graph endpoint, https://graph.windows.net.

There is a new "v2" API for this [[1]], and there is a blog post describing what has changed [[2]].

We need Rancher to upgrade its internal AzureAD auth libraries to use the new MSAL API. The current ADAL-based API is completely deprecated and no longer receives security or feature updates, which could potentially pose a security risk.

**Other details that may be helpful:**

N/A

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): Latest
- Installation option (single install/HA): Either

[1]: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-migration
[2]: https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363

gz#12523

miraculixx:
This may help to assess the priority:

![image](https://user-images.githubusercontent.com/859611/96864815-4bc11a00-1469-11eb-892e-cb297c10e14b.png)

The Learn more link is https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363

oxr463:
> Starting June 30th, 2022, we will end support for ADAL and Azure AD Graph and will no longer provide technical support or security updates. Apps using Azure AD Graph after this time will no longer receive responses from the Azure AD Graph endpoint. Apps using ADAL on existing OS versions will continue to work after this time but will not get any technical support or security updates.

Looks like we still have a couple of years before they stop providing security updates.

dramich:
The go MSAL library is not ready for public preview yet and has no releases so we can not update at this time. Once it's considered stable we can start the transition. 
https://github.com/AzureAD/microsoft-authentication-library-for-go

WolfspiritM:
[...]

PR #37592, opened 2022-05-04
Add a flow for working with Microsoft Graph API to Azure AD auth provider

maxsokolovsky:
Issue https://github.com/rancher/rancher/issues/29306

We must update Rancher's AzureAD auth provider to use the new Microsoft Graph API to access users and groups in Active Directory. Microsoft will end support of the existing, old AzureAD Graph API, by end of 2022.
If we don't do this, users' configurations of AzureAD auth provider will not work, which means users may not be able to log in, or admins may not be able to set up AzureAD as Rancher's auth provider.

This PR does not remove the old authentication and authorization flow for the Azure AD Graph API. It adds code for the new flow, so Rancher can work with the new Microsoft Graph API. This means Rancher supports both Graph APIs. But this is temporary. When Microsoft shut down the old API, we will remove the code that works with the old Azure AD Graph API.

The old flow of authentication and authorization, which users can use before they choose to update Rancher, has not been changed.

When users choose to migrate Rancher to use the new Microsoft Graph API, the backend does the following:
1. Update the GraphEndpoint field of the `azuread` AuthConfig resource in the database.
1. Delete all users' secrets containing access tokens to the old Graph API.
1. Purge the in-memory group cache.

The value of GraphEndpoint on the Auth Config is the state. It's the only piece of data Rancher uses to decide which flow to use, which authentication library and which SDK to use.

### Architectural changes
For the new flow, we use application permissions instead of the delegated permissions.
https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#permission-types
This means we will make requests against the Graph API on behalf of the one service account for all users. The service account is the app configured in Azure. The access token for the service account is stored in the secret `azuread-access-token` in the 
`cattle-global-data` namespace. It is therefore shared among all users who authenticate, perform user and group search, etc.
The token is refreshed every hour; not in the background, but on the next fetch when it has expired.


pkg/auth/providers/azure/azure_actions.go
```diff
@@ -100,3 +106,52 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
+	// It used the Delegated permission types, which required Rancher to retrieve and store all users' credentials to
+	// the Graph API, rather than credentials of just the service account (the application defined in Azure).
+	defer ap.deleteUserAccessTokens()
+	defer groupCache.Purge()
+
+	c, err := ap.getAzureConfigK8s()
+	if err != nil {
+		return err
+	}
+	if authProviderEnabled(c) {
```
MbolotSuse, line 126: So if I am reading this right, you have to be active under the old format to use this migrate endpoint. Is there some other part of functionality to make sure that new users in 2.6.6+ only can use the new graph api?

pkg/auth/providers/azure/clients/adclient/ad_graph_client.go
```diff
@@ -1,28 +1,58 @@
...
+func (c azureADGraphClient) GetGroup(id string) (interface{}, error) {
+	return c.groupClient.Get(context.Background(), id)
+}
+
+func (c azureADGraphClient) ListGroups(filter string) (interface{}, error) {
+	return c.groupClient.List(context.Background(), filter)
+}
+
+func (c azureADGraphClient) ListGroupMemberships(id string, params interface{}) ([]string, error) {
+	p, ok := params.(graphrbac.UserGetMemberGroupsParameters)
```
MbolotSuse, line 37: It looks like the old client is using one parameter - `SecurityEnabledOnly`. You could probably skip providing this as an arg and hardcode it in this function. 

This would have the following benefits:
1) Puts the logic related to the old api in the client specific to the old api (rather than having an old api filter in the more generic provider).
2) Prevents you from having to do a type cast here.
3) Reduces the number of required args to call this function.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,154 +217,154 @@ func (ap *azureProvider) TransformToAuthProvider(
...
+		return nil, errors.New("wrong type of group result")
 	}
-	return principals, nil
 }
 
-func (ap *azureProvider) newAzureClient(secret string) (*azureClient, error) {
-	config, err := ap.getAzureConfigK8s()
-	if err != nil {
-		return nil, err
+func (ap *azureProvider) newClient(cfg *v32.AzureADConfig, secret string) (clients.AzureClient, error) {
```
MbolotSuse, line 351: You may want to consider making a wrapper around this function - I saw the pattern of "if old get user token, pass in secret" at least twice. Maybe something like:

```
func (ap *azureProvider) newClient(cfg *v32.AzureAdConfig, token v3.Token) (*clients.AzureClient, error){
        var secret string
	if authProviderOld(cfg) {
		secret, err = ap.tokenMGR.GetSecret(token.UserID, Name, []*v3.Token{&token})
		        return ap.newAzureADGraphClient(cfg, secret)
		}
	}

	return ap.newMSGraphClient(cfg)
```

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,154 +217,154 @@ func (ap *azureProvider) TransformToAuthProvider(
...
-	}
-	if !allowed {
-		return v3.Principal{}, nil, "", httperror.NewAPIError(httperror.Unauthorized, "unauthorized")
+		return nil, err
 	}
 
-	providerToken, err := azureClient.marshalTokenJSON()
-	if err != nil {
-		return v3.Principal{}, nil, "", err
+	switch u := users.(type) {
```
MbolotSuse, line 272: I would recommend adjusting the relevant functions of the clients to return a list of principals rather than an interface{} which forces the provider to do the conversion for them.

pkg/auth/providers/azure/clients/client.go
```diff
@@ -0,0 +1,14 @@
+package clients
+
+// AzureClient specifies the subset of operations that a real client would delegate to some SDK for accessing
+// one of the two APIs to work with Active Directory resources - Azure AD Graph and Microsoft Graph.
+type AzureClient interface {
```
MbolotSuse, line 5: A general note on this interface:

You have a lot of methods here which return `interface{}` objects. Generally speaking, I think it's best to avoid return types like this. It's difficult to consume when you call the function because you need to be aware about the internal implementation details of the function rather than just being able to accept/process a return type.

In several of these cases, I think that you can replace the `interface{}` with more specific objects (namely principals). You'll have to move a lot of the typecasting/conversion from the provider and into the client, but I think it's doable and would improve readability here substantially.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -422,102 +428,103 @@ func (ap *azureProvider) getAzureConfigK8s() (*v32.AzureADConfig, error) {
...
 
-		g.Go(func() error {
-			groupObj, err := azureClient.groupClient.Get(context.Background(), gp)
+		tasksManager.Go(func() error {
+			defer func() {
+				if err := recover(); err != nil {
+					logrus.Errorf("[AZURE_PROVIDER] Panic while fetching/caching a group: %v", err)
+				}
+			}()
+			// This is inefficient for a collection of msgraph.Group. This is temporary - until support for Azure AD Graph is removed.
```
MbolotSuse, line 464: You might be able to get around this by caching the groups in the client when they are queried.

E.x. 
1. call to azureClient.ListGroupMemberships -> []string, but sets azureClient.Groups = []group
2. Call to azureClient.GetGroup(groupID) checks azureClient.Groups to see if it already listed this group before.

This way, the interface to the provider remains the same, but you still don't have to go all the way to the api. You probably will need to use a map[string]group rather than a slice, but I think that you get the picture.

MbolotSuse (changes requested):
Got a couple of areas where I think that the structure could be approved.

In general, I think that there's some logic in the provider which is specific to the old/new flow that we could put in the clients instead, which would allow for increased readability and a more compact provider.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,154 +217,154 @@ func (ap *azureProvider) TransformToAuthProvider(
...
-	providerToken, err := azureClient.marshalTokenJSON()
-	if err != nil {
-		return v3.Principal{}, nil, "", err
+	switch u := users.(type) {
+	case *[]msgraph.User:
+		return ap.msGraphUsersToPrincipals(u, token)
+	case graphrbac.UserListResultPage:
+		return ap.azureADGraphUsersToPrincipals(u, token)
+	default:
+		return nil, errors.New("wrong type of user result")
```
rmweir, line 278: Use a more descriptive error message that may make sense to end user or support.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,154 +217,154 @@ func (ap *azureProvider) TransformToAuthProvider(
...
+	for _, group := range groups.Values() {
+		p, err := groupToPrincipal(group)
+		if err != nil {
+			return nil, err
+		}
+		p.MemberOf = ap.tokenMGR.IsMemberOf(token, *p)
+		principals = append(principals, *p)
 	}
 	return principals, nil
 }
```
rmweir, line 332: By using interfaces and combining userToPrincipal and groupToPrincipal could you combine these into 1 or 2 functions? This would help code maintenance by making it so if a change is made, it is made for all types at once since any change would probably apply to all of them.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,154 +217,154 @@ func (ap *azureProvider) TransformToAuthProvider(
...
 
-func (ap *azureProvider) searchGroups(client *azureClient, name string, token v3.Token) ([]v3.Principal, error) {
+func (ap *azureProvider) searchGroups(client clients.AzureClient, name string, token v3.Token) ([]v3.Principal, error) {
```
rmweir, line 334: Change function name to be more descriptive or add function comment.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -598,3 +574,130 @@ func (ap *azureProvider) GetUserExtraAttributes(userPrincipal v3.Principal) map[
...
+	logrus.Debug("[AZURE_PROVIDER] Completed token swap with AzureAD")
+
+	oid, err := extractFieldFromJWT(azureClient.AccessToken(), "oid")
+	if err != nil {
+		return v3.Principal{}, nil, "", err
+	}
+
+	logrus.Debug("[AZURE_PROVIDER] Started getting user info from AzureAD")
+
+	u, err := azureClient.GetUser(oid.(string))
```
rmweir, line 595: With this always be safe? To ensure safety you can do something like `id, _ := oid.(string)` above instead.

pkg/auth/providers/azure/clients/msclient/ms_graph_client.go
```diff
@@ -0,0 +1,199 @@
...
+		names = append(names, *g.ID)
+	}
+	return names, nil
+}
+
+type customAuthResult struct {
+	AccessToken    string    `json:"access_token,omitempty"`
+	ExpiresOn      time.Time `json:"expires_on,omitempty"`
+	GrantedScopes  []string  `json:"granted_scopes,omitempty"`
+	DeclinedScopes []string  `json:"declined_scopes,omitempty"`
```
rmweir, line 68: Can these be changed to camel case instead? That is what we use for json properties.

pkg/auth/providers/azure/conversions.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func userToPrincipal(user interface{}) (*v3.Principal, error) {
+	var id, displayName, loginName string
+
+	switch u := user.(type) {
+	case *msgraph.User:
+		id = *u.ID
+		displayName = *u.DisplayName
+		loginName = *u.UserPrincipalName
```
rmweir, line 19: Are these ever nil?

pkg/auth/providers/azure/conversions.go
```diff
@@ -0,0 +1,67 @@
...
+	case graphrbac.User:
+		id = *u.ObjectID
+		displayName = *u.DisplayName
+		loginName = *u.UserPrincipalName
```
rmweir, line 23: Are these ever nil?

pkg/auth/providers/azure/conversions.go
```diff
@@ -0,0 +1,67 @@
...
+	default:
+		return nil, errors.New("wrong type of user")
```
rmweir, line 25: Make this error message more descriptive.

rmweir (changes requested):
Please replace all single/two letter variable names. Review whether certain errors should be returned as API errors with specific errors codes, otherwise 500 will be used by default.

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -100,3 +106,47 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
+	// Provider name for Azure AD is the main key on secret data. This allows to identify the secrets to be deleted.
+	const key = Name
+	for _, secret := range secrets.Items {
+		if _, keyPresent := secret.Data[key]; keyPresent {
+			err := ap.secrets.DeleteNamespaced(tokens.SecretNamespace, secret.Name, &metav1.DeleteOptions{})
+			if err != nil {
+				logrus.Errorf("failed to delete secret %s:%s - %v", tokens.SecretNamespace, secret.Name, err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
maxsokolovsky: `Name` is a constant, the name of the auth provider - `azuread` in this case.

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -100,3 +106,47 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 
 	return ap.tokenMGR.CreateTokenAndSetCookie(user.Name, userPrincipal, groupPrincipals, providerToken, 0, "Token via Azure Configuration", request, userExtraInfo)
 }
+
+// migrateToMicrosoftGraph represents the migration of the registered Azure AD auth provider
+// from the deprecated Azure AD Graph API to the Microsoft Graph API.
+// It modifies the existing auth config value in the database, so that it has a new endpoint to the new API.
+func (ap *azureProvider) migrateToMicrosoftGraph() error {
+	defer ap.deleteUserAccessTokens()
```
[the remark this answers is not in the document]
maxsokolovsky: I had thought about this, and I think it's fine. Deleting them does not hurt - users will just have to re-login.

pkg/auth/providers/azure/azure_actions.go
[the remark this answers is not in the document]
maxsokolovsky: Yes, this code is run if you have Rancher 2.6.6+, and you have Azure AD provider set up.
If you have the new set up, then the code simply re-saves the auth config with no changes (can avoid the single-time re-save, let me know what you think).
If you have the old setup, then the code re-saves the auth config with an actual endpoint change.

In principle, you could run this multiple times, if you know about the API action, but to no effect.
The UI must only show the button (which will run the action) if you run 2.6.6+ AND the auth provider has the old setup.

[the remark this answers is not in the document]
MbolotSuse: Would you mind clarifying where this is run on 2.6.6+? IIRC this migrate is just an action - someone has the manually request a migration through an API call/UI button for this to be triggered right?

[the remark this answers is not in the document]
maxsokolovsky: Right, only when the action is run, which is deliberately triggered.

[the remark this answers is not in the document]
MbolotSuse: Ok - so let's say that I start a fresh setup on 2.6.6 - what ensures that the endpoints in my azuread auth config are for MSGraph and not AZGraph?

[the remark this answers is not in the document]
maxsokolovsky: The UI will offer the new endpoint on registration. It will send it to the backend, and Rancher will save it on the auth config object.

pkg/auth/providers/azure/clients/msclient/ms_graph_client.go
```diff
@@ -0,0 +1,199 @@
...
+	return g, err
+}
+
+func (c azureMSGraphClient) ListGroups(filter string) (interface{}, error) {
+	groups, _, err := c.groupClient.List(context.Background(), odata.Query{Filter: filter})
+	return groups, err
+}
+
+func (c azureMSGraphClient) ListGroupMemberships(id string, _ interface{}) ([]string, error) {
+	groups, _, err := c.userClient.ListGroupMemberships(context.Background(), id, odata.Query{})
```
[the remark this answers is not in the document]
maxsokolovsky: You can't specify it with a call to one of its clients. Also, from what I can tell, it's false by default anyways, which is what we want.

pkg/auth/providers/azure/conversions.go
maxsokolovsky: Since I handle errors in previous calls, when the user/group is initially fetched, I don't expect nils.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -68,30 +69,33 @@ func (ap *azureProvider) GetName() string {
 	return Name
 }
 
-func (ap *azureProvider) AuthenticateUser(
-	ctx context.Context, input interface{},
-) (v3.Principal, []v3.Principal, string, error) {
+func (ap *azureProvider) AuthenticateUser(ctx context.Context, input interface{}) (v3.Principal, []v3.Principal, string, error) {
```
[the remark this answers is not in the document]
maxsokolovsky: Adjusted, now delegating to the client.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -68,30 +69,33 @@ func (ap *azureProvider) GetName() string {
...
+	if err != nil {
+		return nil, err
+	}
+	azureClient, err := ap.newClient(cfg, secret)
 	if err != nil {
 		return nil, err
 	}
 
-	oid, err := parseJWTforField(azureClient.accessToken(), "oid")
+	oid, err := extractFieldFromJWT(azureClient.AccessToken(), "oid")
```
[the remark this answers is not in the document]
maxsokolovsky: Good point, and looks like it has been this way before, in the old code.

pkg/auth/providers/azure/config.go
```diff
@@ -0,0 +1,30 @@
...
+	globalMSGraphEndpoint      = "https://graph.microsoft.com"
+	chinaAzureADGraphEndpoint  = "https://graph.chinacloudapi.cn/"
+	chinaMSGraphEndpoint       = "https://microsoftgraph.chinacloudapi.cn"
+)
+
+func authProviderEnabled(config *v32.AzureADConfig) bool {
+	return config.Enabled && config.GraphEndpoint != ""
+}
+
+func authProviderOld(config *v32.AzureADConfig) bool {
```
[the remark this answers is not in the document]
maxsokolovsky: I did try to rearrange a few things to read the graph endpoint from the auth config from the database, compute a field `isGraphEndpointDeprecated`, and store it on the auth provider value, which lives in memory. And then simply read it on every login or access token request.
However, there were two difficulties:
1. When the auth provider value is being initialized, the `azuread` auth config does not yet exist in the database, so I can't initially read the value from there. Some significant refactoring will be required to change this. I could in principle write some hook that would set that variable later, say on login or other similar action. But, again, a lot of careful work might be needed here.
1. It still might be hard to keep the value in memory on the auth provider and graph config endpoint on the auth config in the database synchronized, especially since we (secretly) allow users to go back by simply editing the graph endpoint straight on the db item. Catching that and updating the value in memory is then really hard, if possible.

I wouldn't say this is all terribly inefficient. First, the fetching of the auth config has been happening all the time in the old code, there just has not been this additional string comparison of the endpoints. And this string comparison is not a particularly expensive operation.
The good thing is that we will remove this check of whether the endpoint is deprecated soon, so this won't be an issue at all.

pkg/auth/providers/azure/conversions.go
maxsokolovsky: I will add another nil check inside this function, just in case.

maxsokolovsky: Since we do exhaustive type checking here, if a user or group is nil, the default branch of the switch statement will be executed.

pkg/auth/providers/azure/azure_actions.go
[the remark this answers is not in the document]
MbolotSuse: To Ricky's point here - I would expect that using the upgrade button would be a no-op on an upgraded provider. However, that's not really true. At minimum we are going to purge the group cache as well.

I don't think that this really represents an issue since both of the functions here are relatively benign, but this is something to think about.

pkg/auth/providers/azure/clients/conversions.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
+	"github.com/manicminer/hamilton/msgraph"
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const Name = "azuread"
```
rmweir, line 16: Change this to something like `AzureADProviderName`.

pkg/auth/providers/azure/clients/group_cache.go
```diff
@@ -0,0 +1,69 @@
+package clients
+
+import (
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	"time"
+
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+)
```
rmweir, line 10: Packages should be divided between standard and otherwise.

pkg/auth/providers/azure/azure_provider.go
rmweir: Still should change this.

rmweir (changes requested):

pkg/auth/providers/azure/clients/group_cache.go
maxsokolovsky: Yup, missed in this file, I wish Goland would watch this and sort automatically. There is a way to do it, just needs tinkering with the settings.

rmweir (dismissed):
LGTM

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,61 +202,27 @@ func (ap *azureProvider) TransformToAuthProvider(
...
-	var mem bool
-	params := graphrbac.UserGetMemberGroupsParameters{
-		SecurityEnabledOnly: &mem,
+	var useMSGraph bool
+	if !graphEndpointDeprecated(config) {
+		useMSGraph = true
 	}
-
-	userGroups, err := azureClient.userClient.GetMemberGroups(context.Background(), *user.ObjectID, params)
+	azureAuthenticator := clients.NewAuthenticator(useMSGraph)
```
MbolotSuse, line 218: This is mostly related to my questions that I have posed in other places, but I think it's worth restating here:

1) Use of "Authenticator" here is confusing - is is a client same as elsewhere.
2) Why don't we initialize the client using the NewClient method that we have in the provider? It would prevent us from doing the Re-init client calls that we do in the client's login user methods.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -68,114 +63,111 @@ func (ap *azureProvider) GetName() string {
...
+	cfg, err := ap.getAzureConfigK8s()
+	if err != nil {
+		return v3.Principal{}, err
+	}
 
-	secret, err := ap.tokenMGR.GetSecret(token.UserID, Name, []*v3.Token{&token})
-	if err != nil && !apierrors.IsNotFound(err) {
-		return princ, err
+	var secret string
+	if graphEndpointDeprecated(cfg) {
```
MbolotSuse, line 161: Nit:

You do this sort of "Is depracted/ if so fetch secret from token" call in the SearchPrincipals call too - might be worth just passing in the token to the "new client" method and having that function/method do this for us.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -0,0 +1,303 @@
...
+
+// AccessTokenCache is responsible for reading (replacing) the access token from some storage (a secret in the database,
+// in this case) and writing it (exporting) to storage. The Microsoft Graph SDK is responsible for verifying and calling
+// the cache methods when needed, so this type simply implements the interface for the cache that the SDK requires.
+type AccessTokenCache struct {
+	Secrets corev1.SecretInterface
+}
+
+func (c AccessTokenCache) Replace(cache cache.Unmarshaler, key string) {
+	secret, err := common.ReadFromSecret(c.Secrets, "cattle-global-data:azuread-access-token", "access-token")
```
MbolotSuse, line 254: You may want to consider using `common.SecretsNamespace + ":azuread-access-token"` to stay in pairity with the way that this functions.

MbolotSuse (commented):
In general: Good work, there was a lot to consider here and some outside design decisions that made this tough to implement.

There are a couple of things that I think need a second look.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// AccessTokenCache is responsible for reading (replacing) the access token from some storage (a secret in the database,
+// in this case) and writing it (exporting) to storage. The Microsoft Graph SDK is responsible for verifying and calling
+// the cache methods when needed, so this type simply implements the interface for the cache that the SDK requires.
+type AccessTokenCache struct {
+	Secrets corev1.SecretInterface
+}
+
+func (c AccessTokenCache) Replace(cache cache.Unmarshaler, key string) {
```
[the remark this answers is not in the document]
maxsokolovsky: Discussed offline - just to document - Replace is load from database into cache, Export is store the cached value in the database.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -0,0 +1,237 @@
...
+	logrus.Debug("[AZURE_PROVIDER] Started token swap with AzureAD")
+
+	azureClient, err := NewClientFromCode(credential.Code, config)
+	if err != nil {
+		return v3.Principal{}, nil, "", err
+	}
+
+	logrus.Debug("[AZURE_PROVIDER] Completed token swap with AzureAD")
+
+	oid, err := ExtractFieldFromJWT(azureClient.AccessToken(), "oid")
```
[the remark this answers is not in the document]
maxsokolovsky: Checked - this token is for the actual user attempting to login.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return *groups.Value, nil
+}
+
+func (c azureADGraphClient) LoginUser(credential *v32.AzureADLogin, config *v32.AzureADConfig, _ corev1.SecretInterface) (v3.Principal, []v3.Principal, string, error) {
+	logrus.Debug("[AZURE_PROVIDER] Started token swap with AzureAD")
+
+	azureClient, err := NewClientFromCode(credential.Code, config)
```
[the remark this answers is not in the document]
maxsokolovsky: Good point, I chose the second option. The first is hard to implement, since each new clients takes its own set of arguments, so it's hard to make a single constructor for them.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -0,0 +1,303 @@
...
+	// Return an empty string for the provider token, so that it does not get saved in a secret later, like users'
+	// access tokens are stored in secrets in the old Azure AD Graph flow.
+	return userPrincipal, groupPrincipals, "", nil
+}
+
+var errMSGraphNoPermissions = fmt.Errorf("missing required permissions from Microsoft Graph: need %s",
+	strings.Join(msGraphRequiredPermissions, ", "))
+
+var msGraphRequiredPermissions = []string{
+	"Group.Read.All",
```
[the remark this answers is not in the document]
maxsokolovsky: We are to require app registrations to have `User.Read.All` and `Group.Read.All` permissions.

Currently, for Azure AD Graph, we require the following (from the docs):
```
    Access the directory as the signed-in user
    Read directory data
    Read all groups
    Read all users’ full profiles
    Read all users’ basic profiles
    Sign in and read user profile
```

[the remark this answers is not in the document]
MbolotSuse: I'll resolve this since this is technically in line with what we do on the old side.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var principals []v3.Principal
+	for _, u := range users.Values() {
+		principal, err := UserToPrincipal(u)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
maxsokolovsky: It is done, but outside the client, after the principal has been fetched and converted properly.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -77,7 +77,7 @@ func (ap *azureProvider) RefetchGroupPrincipals(principalID string, secret strin
 	if err != nil {
 		return nil, err
 	}
-	azureClient, err := ap.newAzureClient(cfg, secret)
+	azureClient, err := ap.newAzureClientFromSecret(cfg, secret)
```
MbolotSuse, line 80: This function seems to always use the old side code. We still need to be able to call this function (RefetchGroupPrincipals) in the nightly refresh if we are using the upgraded provider. 

In the current form, I would expect new side code to straight up not work during the nightly refresh, which would be bad.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -81,55 +81,53 @@ func (c azureADGraphClient) ListGroupMemberships(id string) ([]string, error) {
...
 
 	return userPrincipal, groupPrincipals, providerToken, nil
 }
 
-// NewClientFromCode sets up the SPT, user and group client using a code
-func NewClientFromCode(code string, config *v32.AzureADConfig) (AzureClient, error) {
-	ac := &azureADGraphClient{}
-
+// setInternalFields configures the SPT, user, and group client using a code.
+func (c *azureADGraphClient) setInternalFields(code string, config *v32.AzureADConfig) error {
```
MbolotSuse, line 127: Near as I can tell, you really have two separate constructors here:
- One constructor takes in an auth code and returns a constructed client
- One constructor takes in a token and returns a constructed client

I would recommend creating this as a separate constructor and then using that constructor when the provider's login method is called. There's also a fair amount of duplicated logic between these two constructors - I would consider moving that common logic into a common function.

pkg/auth/providers/azure/clients/conversions.go
```diff
@@ -15,52 +15,56 @@ import (
 
 const Name = "azuread"
 
-// UserToPrincipal attempts to convert a raw user from Graph SDK to a Principal.
-// The user may be of type returned by Azure AD Graph API (deprecated and to be removed soon)
-// or Microsoft Graph API.
-func UserToPrincipal(user interface{}) (v3.Principal, error) {
-	var id, displayName, loginName string
+func (c azureADGraphClient) userToPrincipal(user interface{}) (v3.Principal, error) {
```
MbolotSuse, line 18: Now that this method is on the client you should not have to use an interface{} as the type of the first arg. Would recommend changing this to type `graphrbac.User`.

similar comment for the rest of the functions in the file - we can use real types in the params now.

pkg/auth/providers/azure/clients/conversions.go
```diff
@@ -15,52 +15,56 @@ import (
...
+	p := v3.Principal{
+		ObjectMeta:    metav1.ObjectMeta{Name: Name + "_group://" + *g.ObjectID},
+		DisplayName:   *g.DisplayName,
+		PrincipalType: "group",
+		Provider:      Name,
+	}
+	return p, nil
+}
 
+func (c azureMSGraphClient) userToPrincipal(user interface{}) (v3.Principal, error) {
```
MbolotSuse, line 46: Nit:

I would recommend moving these functions to their respective client files and deleting the conversions file. There's no real reason to mix methods for different structs in this file.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -151,14 +153,19 @@ func msGraphTokenHasPermissions(token string) bool {
 	if !ok {
 		return false
 	}
-	roleNames := make([]string, len(roles))
+	permissionPresent := make(map[string]bool)
+	for _, p := range msGraphRequiredPermissions {
+		permissionPresent[p] = false
```
MbolotSuse, line 158: Nit:

You could just check if the permission is not present in the map at the end rather than re-initing to false here.

Marking this as a nit because the efficiency loss is so minor.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -151,14 +153,19 @@ func msGraphTokenHasPermissions(token string) bool {
...
+	}
 	for _, r := range roles {
 		if name, ok := r.(string); ok {
-			roleNames = append(roleNames, name)
+			if slice.ContainsString(msGraphRequiredPermissions, name) {
```
MbolotSuse, line 162: Nit:

Rather than checking if the permission is one of our requiredPermissions, you could just throw it in the map. This would let us skip the "slice search and string comparison" part of this step. As with the above nit, this is relatively minor.

MbolotSuse (changes requested):
Couple of small nits and one significant functionality issue.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return *groups.Value, nil
+}
+
+func (c *azureADGraphClient) LoginUser(credential *v32.AzureADLogin, config *v32.AzureADConfig, _ corev1.SecretInterface) (v3.Principal, []v3.Principal, string, error) {
+	logrus.Debug("[AZURE_PROVIDER] Started token swap with AzureAD")
+
+	err := c.setInternalFields(credential.Code, config)
```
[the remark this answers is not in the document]
maxsokolovsky: Because there are really two constructors for the client, which is not pretty. Each needs different arguments. One is for the login. Another is for when a secret with an access token is fetched from the database.

pkg/auth/providers/azure/azure_provider.go
maxsokolovsky: Good catch, updating to return the appropriate client based on the Graph endpoint - now it always returns the old client.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -153,19 +152,14 @@ func msGraphTokenHasPermissions(token string) bool {
 	if !ok {
 		return false
 	}
-	permissionPresent := make(map[string]bool)
-	for _, p := range msGraphRequiredPermissions {
-		permissionPresent[p] = false
-	}
+	permissionSet := make(map[string]struct{})
```
MbolotSuse, line 155: Question for you here - is this some go-ism (the struct{}) that I'm not aware of which essentially creates a set of values?

MbolotSuse (commented):
One question on a go-ism I don't quite understand.

pkg/auth/providers/azure/clients/ms_graph_client.go
maxsokolovsky: Yeah, that's a quick way to make a set, simpler than a map of things to bools, a common pattern.

MbolotSuse (commented):
Minor comment on setting state in the login function. Still need to do an E2E review of what is here.

pkg/auth/providers/azure/clients/ad_graph_client.go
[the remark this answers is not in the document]
MbolotSuse: Closing in favor of more current discussion on the same topic.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -0,0 +1,321 @@
...
+
+	// Acquire the OID just to verify the user.
+	oid, err := oidFromAuthCode(credential.Code, config)
+	if err != nil {
+		return v3.Principal{}, nil, "", err
+	}
+	logrus.Debug("[AZURE_PROVIDER] Completed token swap with AzureAD")
+
+	// Now try to acquire the service account's (not the user's) access token.
+	err = c.setInternalFields(config, AccessTokenCache{Secrets: secrets})
```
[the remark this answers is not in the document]
MbolotSuse: Closing this in favor of more current discussion on the same topic.

pkg/auth/providers/azure/azure_provider.go
MbolotSuse: Fixed - resolving.

pkg/auth/providers/azure/clients/client.go
```diff
@@ -29,41 +25,15 @@ type AzureClient interface {
...
-	}
-
-	var data map[string]interface{}
-	err = json.Unmarshal(decoded, &data)
-	if err != nil {
-		return "", fmt.Errorf("error unmarshaling token")
-	}
-	if _, ok := data[fieldID]; !ok {
-		return "", fmt.Errorf("missing field %s from token", fieldID)
+func NewAzureClient(config *v32.AzureADConfig, credential *v32.AzureADLogin, useDeprecatedAzureADClient bool) (AzureClient, error) {
```
MbolotSuse, line 28: A couple of things here:
1) The msgraph client is still not initialized - so anything using the new flow will still need to have the `setInternalFields` called, which is not preferable. 
2) This is using a non-exported function of the adGraphClient (configureServicePrinciaplToken) rather than using an exported Constructor, which seems odd to me. 

I would recommend the following:
1) Initialize the MSGraph client using an exported constructor.
2) Create a constructor for AzureADGraphClient, even if it's just a wrapper around configureServicePrincipalToken.

I would also like to note that we have a `newAzureClient` method in the provider - I'm kind of thinking that method and this one overlap in functionality but don't seem to know about each other. Could you elaborate on why we have these as separate methods.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -228,11 +219,10 @@ func (c azureADGraphClient) userToPrincipal(user graphrbac.User) (v3.Principal,
 }
 
 func (c azureADGraphClient) groupToPrincipal(group graphrbac.ADGroup) (v3.Principal, error) {
```
MbolotSuse, line 221: Nit: Any reason why this and (and the corresponding user method, both here and in the new client) are `c Client` rather than `c *Client`?

MbolotSuse (commented):
Mostly minor nitpicks, some outstanding issues around client construction.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,61 +187,27 @@ func (ap *azureProvider) TransformToAuthProvider(
...
 		return v3.Principal{}, nil, "", err
 	}
-
-	groupPrincipals, err := ap.userGroupsToPrincipals(azureClient, userGroups)
-	if err != nil {
-		return v3.Principal{}, nil, "", err
-	}
-
 	testAllowedPrincipals := config.AllowedPrincipalIDs
 	if test && config.AccessMode == "restricted" {
```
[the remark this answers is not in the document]
MbolotSuse: This is being called by the actions, where test is set to true, so we can't remove this.

pkg/auth/providers/azure/clients/ad_graph_client.go
MbolotSuse: Standard go practice - the receiver is not modified so we don't need this to be a pointer.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -299,9 +248,49 @@ func (c AccessTokenCache) Export(cache cache.Marshaler, key string) {
 	}
 }
 
-func NewMSGraphClientFromAuthConfig(config *v32.AzureADConfig, cache cache.ExportReplace) (*azureMSGraphClient, error) {
+// NewMSGraphClient returns a client of the Microsoft Graph API. It attempts to get an access token to the API.
+// It first tries to fetch the token from the refresh token, if the access token is found in the database.
+// If that fails, it tries to acquire it directly from the auth provider with the credential (application secret in Azure).
+func NewMSGraphClient(config *v32.AzureADConfig, cache cache.ExportReplace) (*azureMSGraphClient, error) {
```
MbolotSuse, line 254: Nit:

Since the cache is only used here, it may make sense for you to accept a secret interface and then construct the cache as part of this method call. That being said, this isn't a big deal since you should be using the constructor in the client rather than calling this directly.

MbolotSuse (changes requested):
Few minor issues.

pkg/auth/providers/azure/azure_provider.go
MbolotSuse: Resolving - we are leaving as is in order to not significantly mangle the old side functionality.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -68,114 +63,96 @@ func (ap *azureProvider) GetName() string {
...
-func (ap *azureProvider) AuthenticateUser(
-	ctx context.Context, input interface{},
-) (v3.Principal, []v3.Principal, string, error) {
+func (ap *azureProvider) AuthenticateUser(ctx context.Context, input interface{}) (v3.Principal, []v3.Principal, string, error) {
 	login, ok := input.(*v32.AzureADLogin)
 	if !ok {
 		return v3.Principal{}, nil, "", errors.New("unexpected input type")
 	}
-	return ap.loginUser(login, nil, false)
+	loginUser, groupPrincipals, providerToken, err := ap.loginUser(login, nil, false)
```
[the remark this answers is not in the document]
MbolotSuse: Change made, resolving.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,61 +187,27 @@ func (ap *azureProvider) TransformToAuthProvider(
...
 
-func (ap *azureProvider) loginUser(
-	azureCredential *v32.AzureADLogin,
-	config *v32.AzureADConfig,
-	test bool,
-) (v3.Principal, []v3.Principal, string, error) {
+func (ap *azureProvider) loginUser(azureCredential *v32.AzureADLogin, config *v32.AzureADConfig, test bool) (v3.Principal, []v3.Principal, string, error) {
 	var err error
-
 	if config == nil {
```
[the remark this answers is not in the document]
MbolotSuse: This is being called by another function (the action `testAndApply`), and as such, needs to remain it's own function.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -68,114 +63,96 @@ func (ap *azureProvider) GetName() string {
...
-func (ap *azureProvider) SearchPrincipals(
-	name string,
-	principalType string,
-	token v3.Token,
-) ([]v3.Principal, error) {
-	var princ []v3.Principal
-
-	secret, err := ap.tokenMGR.GetSecret(token.UserID, Name, []*v3.Token{&token})
-	if err != nil && !apierrors.IsNotFound(err) {
+func (ap *azureProvider) SearchPrincipals(name string, principalType string, token v3.Token) ([]v3.Principal, error) {
```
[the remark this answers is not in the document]
MbolotSuse: Resolving change made.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -530,51 +402,36 @@ func formAzureRedirectURL(config map[string]interface{}) string {
...
-	parsed["provider"] = pparts[0]
-	parsed["type"] = pparts[1]
-
-	return parsed, nil
+	return allowed, nil
 }
 
-// samePrincipal verifies if they are the same principal
-func samePrincipal(me v3.Principal, other v3.Principal) bool {
+func samePrincipal(me *v3.Principal, other *v3.Principal) bool {
```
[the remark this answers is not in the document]
MbolotSuse: Resolving, change made.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -211,61 +187,27 @@ func (ap *azureProvider) TransformToAuthProvider(
...
-		return v3.Principal{}, nil, "", err
+	var useMSGraph bool
+	if !graphEndpointDeprecated(config) {
+		useMSGraph = true
 	}
-
-	logrus.Debug("[AZURE_PROVIDER] Completed token swap with AzureAD")
-
-	oid, err := parseJWTforField(azureClient.accessToken(), "oid")
+	azureClient := clients.NewAzureClient(useMSGraph)
```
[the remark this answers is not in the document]
MbolotSuse: Resolving in favor of other discussions.

pkg/auth/providers/azure/clients/client.go
MbolotSuse: Resolving, change made.

pkg/auth/providers/azure/clients/client.go
```diff
@@ -23,17 +22,22 @@ type AzureClient interface {
...
-		logrus.Debug("[AZURE_PROVIDER] Started token swap with AzureAD")
-		if err := c.configureServicePrincipalToken(config, credential.Code); err != nil {
+		return NewADGraphClientFromCredential(config, credential)
+	}
+	return NewMSGraphClient(config, AccessTokenCache{Secrets: secrets})
+}
+
+func NewAzureClientFromSecret(config *v32.AzureADConfig, useDeprecatedAzureADClient bool, secret string, secrets corev1.SecretInterface) (AzureClient, error) {
+	if useDeprecatedAzureADClient {
+		azureToken, err := unmarshalADALToken(secret)
```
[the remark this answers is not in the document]
MbolotSuse: Resolving, change made.

pkg/auth/providers/azure/clients/ms_graph_client.go
MbolotSuse: Resolving, change made.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -279,86 +212,62 @@ func (ap *azureProvider) loginUser(
...
-	if err != nil {
-		return nil, err
-	}
-
-	client, err := newClientToken(config, azureToken)
-	if err != nil {
-		return nil, err
+func (ap *azureProvider) newAzureClientFromToken(cfg *v32.AzureADConfig, token *v32.Token) (clients.AzureClient, error) {
+	var secret string
+	var err error
```
MbolotSuse, line 262: Nit: You don't need to declare err here since it's only used in the if.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -268,28 +263,9 @@ func (ap *azureProvider) newAzureClientFromToken(cfg *v32.AzureADConfig, token *
...
-	azureToken, err := unmarshalADALToken(secret)
-	if err != nil {
-		return nil, err
-	}
-	return clients.NewAzureADGraphClientFromToken(cfg, azureToken)
-}
-
-func (ap *azureProvider) newMSGraphClient(cfg *v32.AzureADConfig) (clients.AzureClient, error) {
-	return clients.NewMSGraphClientFromAuthConfig(cfg, clients.AccessTokenCache{Secrets: ap.secrets})
+	return clients.NewMSGraphClient(cfg, clients.AccessTokenCache{Secrets: ap.secrets})
```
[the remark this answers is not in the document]
MbolotSuse: Resolving, change made.

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -0,0 +1,234 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type azureADGraphClient struct {
+	servicePrincipal adal.ServicePrincipalToken
+	userClient       graphrbac.UsersClient
+	groupClient      graphrbac.GroupsClient
+}
+
+func (c *azureADGraphClient) GetUser(id string) (v3.Principal, error) {
```
rmweir, line 22: Added comments to all exported functions.

rmweir (changes requested):

MbolotSuse (commented):
I left a minor nitpick, but you have fixed so many of those at this point that I think it's not important.

LGTM - I'll issue the actual approval once we get the commits squashed into the final format.

pkg/auth/providers/azure/clients/group_cache.go
```diff
@@ -0,0 +1,64 @@
...
+			p, ok := principal.(v3.Principal)
+			if !ok {
+				logrus.Errorf("failed to convert a cached group to principal")
+				continue
+			}
+			groupPrincipals[j] = p
+			continue
+		}
+
+		tasksManager.Go(func() error {
```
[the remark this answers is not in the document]
maxsokolovsky: This has been here before, I am not changing this. I assume it had been designed like this to cache many groups in parallel.

rmweir (dismissed):
LGTM

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -422,105 +333,55 @@ func (ap *azureProvider) getAzureConfigK8s() (*v32.AzureADConfig, error) {
...
+	if current == secret {
 		return nil
 	}
 
-	return ap.tokenMGR.UpdateSecret(token.UserID, token.AuthProvider, stringNew)
+	return ap.tokenMGR.UpdateSecret(token.UserID, token.AuthProvider, current)
 }
 
 func formAzureRedirectURL(config map[string]interface{}) string {
+	authEndpoint, ok := config["authEndpoint"].(string)
```
MbolotSuse, line 374: Rather than trying to parse this out of the map (which is tricky and could lead to issues since authEndpoint is being used as the source of truth for deprecation here, but not in the case of the the commonly used method), you could use `mapstructure.decode()` to turn this into a well structured AzureConfig object and then use the `graphEndpointDeprecated` function, which is more consistent with our approaches elsewhere. 

The one thing that I am unsure about would be the danger of trying to convert here - it might be that the configmap isn't a valid authconfig. But I'm not 100% sure on that.

```suggestion
	azureConfig := &v32.AzureADConfig{}
	err := mapstructure.Decode(config, azureConfig)
	if err == nil && !graphEndpointDeprecated(azureConfig) {
		return fmt.Sprintf(
			"%s?client_id=%s&redirect_uri=%s&response_type=code&scope=user.read",
			azureConfig.AuthEndpoint,
			azureConfig.ApplicationID,
			azureConfig.RancherURL,
		)
	}
	return fmt.Sprintf(
		"%s?client_id=%s&redirect_uri=%s&resource=%s",
		config["authEndpoint"],
		config["applicationId"],
		config["rancherUrl"],
		config["graphEndpoint"],
	)
```

MbolotSuse (changes requested):
One suggestion on the new endpoint selection logic.

MbolotSuse (approved):
LGTM.

I would like to request that QA run some test cases around disabling the provider and then making sure that we don't do any further interaction with azure, but outside of that I don't have any issues here.

pkg/auth/providers/azure/config.go
```diff
@@ -0,0 +1,40 @@
...
+	globalMSGraphEndpoint      = "https://graph.microsoft.com"
+	chinaAzureADGraphEndpoint  = "https://graph.chinacloudapi.cn/"
+	chinaMSGraphEndpoint       = "https://microsoftgraph.chinacloudapi.cn"
+)
+
+func authProviderEnabled(config *v32.AzureADConfig) bool {
+	return config.Enabled && config.GraphEndpoint != ""
+}
+
+func graphEndpointDeprecated(config *v32.AzureADConfig) bool {
```
rmweir, line 21: Looks like you only use a single string field of config:
```suggestion
func graphEndpointDeprecated(endpoint string) bool {
```

rmweir (changes requested):

pkg/auth/providers/azure/config.go
rmweir: The purpose of changing the parameter type is not to make more sense with the function name. It's to only pass in the fields necessary, and avoid passing pointers, when it is convenient to do so. If it is changed to what I suggested then the current name works, and a comment can clarify that it is not the only way to determine this but the way we use.

maxsokolovsky: Adjusted.

MbolotSuse (approved):
LGTM

rmweir (approved):
LGTM

Issue #37641, opened 2022-05-09
[RKE2] Snapshot restore does not work on a cloud provider enabled cluster


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head commit id `628b1b9`
- Installation option (Docker install/Helm Chart): docker install

**Information about the Cluster**
- Kubernetes version: `v1.22.9+rke2r2`
- Cluster Type (Local/Downstream): AWS RKE2 node driver - 3 etcd, 2 cp, 3 worker nodes

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) std user

   
**To Reproduce**

- Deploy an AWS cloud provider enabled RKE2 downstream cluster
- As a std user take a snapshot
- Restore from one of the snapshots
- The cluster is stuck in `Updating` state with error: `Error while designating init node during control plane/etcd stop: rkecluster fleet-default/sowmya-test: init node sowmya-test-pool1-7f7ce995-gjgg8 was not found during designation`
- `sowmya-test-pool1-7f7ce995-gjgg8` is NOT available

Oats87:
Areas of potential regression are in etcd snapshot restore in general.

sowmyav27:
**On 2.6-head commit id: `f40a8a7`**

- Deploy an AWS cloud provider enabled RKE2 downstream cluster
- As a std user take a snapshot
- Restore from one of the snapshots
- Cluster is seen stuck in `Updating` state with error. `Error while designating init node during control plane/etcd stop: rkecluster fleet-default/sowmyatest: init node sowmyatest-pool1-2f621161-77tps was not found during designation`
- `sowmyatest-pool1-2f621161-77tps` is not available on the cluster.

sowmyav27:
**On 2.6-head commit id: `5f5fb23`**

- Deploy an AWS cloud provider enabled RKE2 downstream cluster
- As a std user take a snapshot
- Restore from one of the snapshots
- Cluster restores and comes back Active

PR #37649, opened 2022-05-09
Properly correlate local snapshot to node when cloudprovider changes …

Oats87:

Upon etcd snapshot restore, a local snapshot node name was blindly trusted as the node to designate. This is not a valid procedure, as we must ensure we correlate the snapshot to the correct machine rather than the node that is listed from the config map in the snapshot. 

This is primarily due to the cloud provider being able to change node names, which can cause problems down the line for restore.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/etcdrestore.go
+++ b/pkg/provisioningv2/rke2/planner/etcdrestore.go
@@ -48,7 +48,11 @@ func (p *Planner) runEtcdSnapshotRestorePlan(controlPlane *rkev1.RKEControlPlane
 		}
 	} else {
 		logrus.Infof("rkecluster %s/%s re-electing specific init node for etcd snapshot restore", controlPlane.Namespace, controlPlane.Spec.ClusterName)
-		joinServer, err = p.designateInitNode(controlPlane, clusterPlan, snapshot.SnapshotFile.NodeName)
+		listSuccessful, machine, err := rke2.GetMachineByID(p.machinesCache, snapshot.Labels[rke2.MachineIDLabel], controlPlane.Namespace, controlPlane.Name)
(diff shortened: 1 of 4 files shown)


jiaqiluo (approved):
LGTM

pkg/provisioningv2/rke2/planner/etcdrestore.go
```diff
@@ -48,7 +48,11 @@ func (p *Planner) runEtcdSnapshotRestorePlan(controlPlane *rkev1.RKEControlPlane
 		}
 	} else {
 		logrus.Infof("rkecluster %s/%s re-electing specific init node for etcd snapshot restore", controlPlane.Namespace, controlPlane.Spec.ClusterName)
-		joinServer, err = p.designateInitNode(controlPlane, clusterPlan, snapshot.SnapshotFile.NodeName)
+		listSuccessful, machine, err := rke2.GetMachineByID(p.machinesCache, snapshot.Labels[rke2.MachineIDLabel], controlPlane.Namespace, controlPlane.Name)
+		if !listSuccessful || machine == nil || machine.Spec.InfrastructureRef.Name == "" || err != nil {
+			return fmt.Errorf("unable to retrieve nodeName for machine on snapshot: %s/%s", snapshot.Namespace, snapshot.Name)
```
thedadams, line 53:
```suggestion
			return fmt.Errorf("unable to retrieve nodeName for machine on snapshot %s/%s: %v", snapshot.Namespace, snapshot.Name, err)
```

thedadams (changes requested):

pkg/provisioningv2/rke2/planner/etcdrestore.go
Oats87: done

thedadams (approved):

jiaqiluo (approved):

PR #37666, opened 2022-05-10
designate proper init node in etcd restore stop plan

Oats87:

Init node designation is done in two places during an etcd restore, and only one was covered. 

In the event that a snapshot is restored for an init node that doesn't match node names anymore (cloud provider changed node name), we must correlate the node name based on the machine infra.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/etcdrestore.go
+++ b/pkg/provisioningv2/rke2/planner/etcdrestore.go
@@ -166,13 +166,13 @@ func (p *Planner) runEtcdRestoreServiceStop(controlPlane *rkev1.RKEControlPlane,
 	var err error
 	isS3 := snapshot.SnapshotFile.S3 != nil
 	if !isS3 { // In the event that we are restoring a local snapshot, we need to reset our initNode
-		if snapshot.SnapshotFile.NodeName != "" && snapshot.SnapshotFile.NodeName != "s3" {
-			joinServer, err = p.designateInitNode(controlPlane, clusterPlan, snapshot.SnapshotFile.NodeName)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/provisioningv2/rke2/planner/etcdrestore.go
```diff
@@ -166,13 +166,13 @@ func (p *Planner) runEtcdRestoreServiceStop(controlPlane *rkev1.RKEControlPlane,
...
 	if !isS3 { // In the event that we are restoring a local snapshot, we need to reset our initNode
-		if snapshot.SnapshotFile.NodeName != "" && snapshot.SnapshotFile.NodeName != "s3" {
-			joinServer, err = p.designateInitNode(controlPlane, clusterPlan, snapshot.SnapshotFile.NodeName)
-			if err != nil {
-				return fmt.Errorf("error while designating init node during control plane/etcd stop: %w", err)
-			}
-		} else {
-			return fmt.Errorf("error attempting to run etcd snapshot restore plan: either s3 info or nodename must be designated (was: %s)", snapshot.SnapshotFile.NodeName)
+		listSuccessful, machine, err := rke2.GetMachineByID(p.machinesCache, snapshot.Labels[rke2.MachineIDLabel], controlPlane.Namespace, controlPlane.Name)
+		if !listSuccessful || machine == nil || machine.Spec.InfrastructureRef.Name == "" || err != nil {
```
jakefhyde, line 170: Nit: is there a way we can improve the error message in the event that `err  == nil`?

Oats87: Sure, but I would rather address that later on rather than have to run through an entire CI run again.

jakefhyde (approved):

thedadams (approved):

PR #37671, opened 2022-05-10
use machine ID directly when designating init node during etcd restore and make init node waiting message less vague

Oats87:

Upon attempts to correct forced init node designation during etcd restore, the wrong name was used (machine name vs. node name). This PR corrects this behavior, so that the machine ID is properly used.

This PR also makes the waiting message when no potential init nodes are available a little less vague, using the term `viable` rather than `possible`.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/etcdrestore.go
+++ b/pkg/provisioningv2/rke2/planner/etcdrestore.go
@@ -47,12 +47,8 @@ func (p *Planner) runEtcdSnapshotRestorePlan(controlPlane *rkev1.RKEControlPlane
 			return err
 		}
 	} else {
-		logrus.Infof("rkecluster %s/%s re-electing specific init node for etcd snapshot restore", controlPlane.Namespace, controlPlane.Spec.ClusterName)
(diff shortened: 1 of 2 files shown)


pkg/provisioningv2/rke2/planner/etcdrestore.go
```diff
@@ -61,8 +57,8 @@ func (p *Planner) runEtcdSnapshotRestorePlan(controlPlane *rkev1.RKEControlPlane
 
 	for _, server := range servers {
 		if isS3 ||
-			(server.Machine.Status.NodeRef != nil &&
-				server.Machine.Status.NodeRef.Name == snapshot.SnapshotFile.NodeName) {
+			(server.Machine.Labels[rke2.MachineIDLabel] != "" && snapshot.Labels[rke2.MachineIDLabel] != "" &&
```
thedadams, line 60: nit, only one check is enough: 
```suggestion
			(server.Machine.Labels[rke2.MachineIDLabel] != "" &&
```

thedadams (approved):

snasovich (approved):

PR #37693, opened 2022-05-11
Remove 'rc' from webhook min version

maxsokolovsky:


cbron (approved):
Lgtm. These have same commit 6f3a318.

Issue #36845, opened 2022-03-10
Concurrent map read/write panic in authprovisioningv2 controller 


Oats87:
CI run https://drone-publish.rancher.io/rancher/rancher/7459/2/2 failed due to Rancher panic

Analysis of Rancher logs has a panic like:
```

fatal error: concurrent map read and map write

goroutine 30298 [running]:
runtime.throw({0x4030875, 0x3f})
	/usr/lib64/go/1.17/src/runtime/panic.go:1198 +0x71 fp=0xc007927648 sp=0xc007927618 pc=0x437131
runtime.mapaccess2_faststr(0x3780da0, 0xc006347701, {0x3fa056e, 0x4})
	/usr/lib64/go/1.17/src/runtime/map_faststr.go:116 +0x3d4 fp=0xc0079276b0 sp=0xc007927648 pc=0x412eb4
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.NestedFieldNoCopy(0xc006347760, {0xc007927820, 0x1, 0x1})
	/go/src/github.com/rancher/rancher/go/pkg/mod/k8s.io/apimachinery@v0.23.3/pkg/apis/meta/v1/unstructured/helpers.go:61 +0xdf fp=0xc007927738 sp=0xc0079276b0 pc=0xccf67f
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.NestedString(0x0, {0xc007927820, 0x1, 0x1})
	/go/src/github.com/rancher/rancher/go/pkg/mod/k8s.io/apimachinery@v0.23.3/pkg/apis/meta/v1/unstructured/helpers.go:75 +0x35 fp=0xc0079277b0 sp=0xc007927738 pc=0xccf835
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.getNestedString(0xc00a9a4960, {0xc007927820, 0xc004fe7250, 0x3})
	/go/src/github.com/rancher/rancher/go/pkg/mod/k8s.io/apimachinery@v0.23.3/pkg/apis/meta/v1/unstructured/helpers.go:278 +0x1e fp=0xc0079277e0 sp=0xc0079277b0 pc=0xcd0fbe
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.(*Unstructured).GetKind(...)
[...]

PR #37696, opened 2022-05-11
Deepcopy infra machine remove

thedadams:
Issue: https://github.com/rancher/rancher/issues/36845

## Problem
There was an issue where a concurrent map read and write would happen. It is
difficult to diagnose, but this is one place where an unstructured object was
being modified without first deep copying the object.

## Solution
Deep copying the object before changing it should stop the panic.

## Testing
This is an intermittent issue. Previously, the panic would happen on approximately 1 out of 3 CI runs. The CI has run 6 times for this PR and no panic has been seen.

### DIFF
--- a/pkg/auth/tokens/manager.go
+++ b/pkg/auth/tokens/manager.go
@@ -48,7 +48,7 @@ var (
 	toDeleteCookies = []string{CookieName, CSRFCookie}
 )
 
-func RegisterIndexer(ctx context.Context, apiContext *config.ScaledContext) error {
+func RegisterIndexer(apiContext *config.ScaledContext) error {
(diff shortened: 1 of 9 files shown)


jakefhyde (approved):

jakefhyde:
I think I ran into this bug a few times during initial encryption key rotation implementation, glad to see it fixed.

annablender (approved):
Dope, this looks great -- the panic in `role.go` also appears in the logs to happen after the `test-three-nodes-all-roles` when machines are being deleted. So it's very likely this is the where the problem was.

jiaqiluo (approved):

PR #37703, opened 2022-05-12
Added randomization and TTL for all kubeconfig tokens.

KevinJoiner:
Users request kubeconfigs with tokens via the CLI or through the UI.
kubeconfig tokens created through the CLI did not have
randomized names. Thus invalidating any previously created tokens.
Tokens requested via the UI did not have a TTL set creating an
endless list of tokens to store.

### DIFF
--- a/pkg/api/norman/customization/setting/setting.go
+++ b/pkg/api/norman/customization/setting/setting.go
@@ -2,7 +2,6 @@ package setting
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/rancher/norman/api/access"
@@ -82,12 +81,6 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
 	case "auth-user-info-resync-cron":
 		_, err = providerrefresh.ParseCron(newValueString)
 	case "kubeconfig-token-ttl-minutes":
-		generateToken := strings.EqualFold(settings.KubeconfigGenerateToken.Get(), "true")
-		if generateToken {
-			return httperror.NewAPIError(httperror.ActionNotAvailable, fmt.Sprintf("kubeconfig-token-ttl-minutes can be set only if rancher doesn't generate token, "+
-				"disable kubeconfig-generate-token"))
-		}
-
 		var tokenTTL time.Duration
 		tokenTTL, err = tokens.ParseTokenTTL(newValueString)
 		if err == nil {
--- a/pkg/api/steve/clusters/kubeconfig.go
+++ b/pkg/api/steve/clusters/kubeconfig.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/rancher/pkg/auth/requests"
+	"github.com/rancher/rancher/pkg/auth/tokens"
 	"github.com/rancher/rancher/pkg/features"
 	"github.com/rancher/rancher/pkg/kubeconfig"
 	"github.com/rancher/rancher/pkg/settings"
@@ -74,11 +75,16 @@ func (k kubeconfigDownload) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 }
 
 func (k kubeconfigDownload) ensureToken(userName string, req *http.Request) (string, error) {
+	defaultTokenTTL, err := tokens.GetKubeconfigDefaultTokenTTLInMilliSeconds()
+	if err != nil {
+		return "", fmt.Errorf("failed to get default token TTL: %w", err)
+	}
 
 	authToken, err := k.auth.TokenFromRequest(req)
 	if err != nil {
 		return "", err
 	}
+
 	tokenNamePrefix := fmt.Sprintf("kubeconfig-%s", userName)
 	input := user.TokenInput{
 		TokenName:     tokenNamePrefix,
@@ -86,10 +92,10 @@ func (k kubeconfigDownload) ensureToken(userName string, req *http.Request) (str
 		Kind:          "kubeconfig",
 		UserName:      userName,
 		AuthProvider:  authToken.AuthProvider,
-		TTL:           nil,
+		TTL:           defaultTokenTTL,
 		Randomize:     true,
 		UserPrincipal: authToken.UserPrincipal,
 	}
-	token, err := k.userMgr.EnsureToken(input)
-	return token, err
+
+	return k.userMgr.EnsureToken(input)
 }
--- a/pkg/auth/providers/common/usermanager.go
+++ b/pkg/auth/providers/common/usermanager.go
@@ -309,10 +309,11 @@ func (m *userManager) EnsureClusterToken(clusterName string, input user.TokenInp
 	return token.Name + ":" + key, nil
 }
 
+// newTokenForKubeconfig creates a new token for a generated kubeconfig.
 func (m *userManager) newTokenForKubeconfig(clusterName, tokenName, description, kind, userName string, ttl time.Duration, userPrincipal v3.Principal) (string, error) {
 	tokenTTL, err := tokens.ValidateMaxTTL(ttl)
 	if err != nil {
-		return "", fmt.Errorf("failed to validate token ttl %v", err)
+		return "", fmt.Errorf("failed to validate token ttl %w", err)
 	}
 
 	ttlMilli := tokenTTL.Milliseconds()
@@ -324,17 +325,17 @@ func (m *userManager) newTokenForKubeconfig(clusterName, tokenName, description,
 		UserName:      userName,
 		AuthProvider:  userPrincipal.Provider,
 		TTL:           &ttlMilli,
-		Randomize:     false,
+		Randomize:     true,
 		UserPrincipal: userPrincipal,
 	}
-	key, err := m.EnsureClusterToken(clusterName, input)
+	fullToken, err := m.EnsureClusterToken(clusterName, input)
 	if err != nil {
 		return "", err
 	}
-	return key, nil
+	return fullToken, nil
 }
 
-// creates kubeconfig tokens with KubeconfigTokenTTL and regenerates if existing token expired
+// GetKubeconfigToken creates a new token for use in a kubeconfig generated through the CLI.
 func (m *userManager) GetKubeconfigToken(clusterName, tokenName, description, kind, userName string, userPrincipal v3.Principal) (*v3.Token, string, error) {
 
 	tokenTTL, err := tokens.ParseTokenTTL(settings.KubeconfigTokenTTLMinutes.Get())
@@ -347,8 +348,9 @@ func (m *userManager) GetKubeconfigToken(clusterName, tokenName, description, ki
 		return nil, "", err
 	}
 
-	_, createdTokenValue := tokens.SplitTokenParts(fullCreatedToken)
-	token, err := m.tokens.Get(tokenName, v1.GetOptions{})
+	randomizedTokenName, createdTokenValue := tokens.SplitTokenParts(fullCreatedToken)
+
+	token, err := m.tokens.Get(randomizedTokenName, v1.GetOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, createdTokenValue, err
 	}
@@ -364,11 +366,11 @@ func (m *userManager) GetKubeconfigToken(clusterName, tokenName, description, ki
 	token, err = m.tokens.Update(tokenCopy)
 	if err != nil {
 		if !apierrors.IsConflict(err) {
-			return nil, "", fmt.Errorf("getToken: updating token [%s] failed [%v]", tokenName, err)
+			return nil, "", fmt.Errorf("getToken: updating token [%s] failed [%v]", randomizedTokenName, err)
 		}
 
 		err = wait.ExponentialBackoff(backoff, func() (bool, error) {
-			token, err = m.tokens.Get(tokenName, v1.GetOptions{})
+			token, err = m.tokens.Get(randomizedTokenName, v1.GetOptions{})
 			if err != nil {
 				return false, err
 			}
@@ -379,7 +381,7 @@ func (m *userManager) GetKubeconfigToken(clusterName, tokenName, description, ki
 
 				token, err = m.tokens.Update(tokenCopy)
 				if err != nil {
-					logrus.Debugf("getToken: updating token [%s] failed [%v]", tokenName, err)
+					logrus.Debugf("getToken: updating token [%s] failed [%v]", randomizedTokenName, err)
 					if apierrors.IsConflict(err) {
 						return false, nil
 					}
@@ -390,7 +392,7 @@ func (m *userManager) GetKubeconfigToken(clusterName, tokenName, description, ki
 		})
 
 		if err != nil {
-			return nil, "", fmt.Errorf("getToken: retry updating token [%s] failed [%v]", tokenName, err)
+			return nil, "", fmt.Errorf("getToken: retry updating token [%s] failed [%v]", randomizedTokenName, err)
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


pkg/settings/setting.go
```diff
@@ -23,36 +24,38 @@ var (
...
+	HelmVersion                      = NewSetting("helm-version", "dev")
+	HelmMaxHistory                   = NewSetting("helm-max-history", "10")
+	IngressIPDomain                  = NewSetting("ingress-ip-domain", "sslip.io")
+	InstallUUID                      = NewSetting("install-uuid", "")
+	InternalServerURL                = NewSetting("internal-server-url", "")
+	InternalCACerts                  = NewSetting("internal-cacerts", "")
+	IsRKE                            = NewSetting("is-rke", "")
+	JailerTimeout                    = NewSetting("jailer-timeout", "60")
+	KubeconfigGenerateToken          = NewSetting("kubeconfig-generate-token", "true")
+	// Deprecated: On removal use kubeconfig-default-ttl-minutes for all kubeconfigs
```
cbron, line 56: Can this go at end of relevant setting ? Harder to read this way

pkg/settings/setting.go
```diff
@@ -23,36 +24,38 @@ var (
...
-	GithubProxyAPIURL                   = NewSetting("github-proxy-api-url", "https://api.github.com")
-	HelmVersion                         = NewSetting("helm-version", "dev")
-	HelmMaxHistory                      = NewSetting("helm-max-history", "10")
-	IngressIPDomain                     = NewSetting("ingress-ip-domain", "sslip.io")
-	InstallUUID                         = NewSetting("install-uuid", "")
-	InternalServerURL                   = NewSetting("internal-server-url", "")
-	InternalCACerts                     = NewSetting("internal-cacerts", "")
-	IsRKE                               = NewSetting("is-rke", "")
-	JailerTimeout                       = NewSetting("jailer-timeout", "60")
-	KubeconfigGenerateToken             = NewSetting("kubeconfig-generate-token", "true")
```
cbron, line 54: Can you avoid this big reformat on above 20 lines, makes git-blame harder in the future.

pkg/settings/setting.go
```diff
@@ -23,36 +24,38 @@ var (
...
+	IngressIPDomain                  = NewSetting("ingress-ip-domain", "sslip.io")
+	InstallUUID                      = NewSetting("install-uuid", "")
+	InternalServerURL                = NewSetting("internal-server-url", "")
+	InternalCACerts                  = NewSetting("internal-cacerts", "")
+	IsRKE                            = NewSetting("is-rke", "")
+	JailerTimeout                    = NewSetting("jailer-timeout", "60")
+	KubeconfigGenerateToken          = NewSetting("kubeconfig-generate-token", "true")
+	// Deprecated: On removal use kubeconfig-default-ttl-minutes for all kubeconfigs
 	KubeconfigTokenTTLMinutes           = NewSetting("kubeconfig-token-ttl-minutes", "960") // 16 hours
+	KubeconfigDefaultTokenTTLMinutes    = NewSetting("kubeconfig-default-ttl-minutes", "0") // 0 TTL = Forever
```
cbron, line 58: Let's add comment inline to describe what this effects

pkg/settings/setting.go
```diff
@@ -23,36 +24,38 @@ var (
...
-	InternalServerURL                   = NewSetting("internal-server-url", "")
-	InternalCACerts                     = NewSetting("internal-cacerts", "")
-	IsRKE                               = NewSetting("is-rke", "")
-	JailerTimeout                       = NewSetting("jailer-timeout", "60")
-	KubeconfigGenerateToken             = NewSetting("kubeconfig-generate-token", "true")
+	AgentImage                       = NewSetting("agent-image", "rancher/rancher-agent:v2.6-head")
+	AgentRolloutTimeout              = NewSetting("agent-rollout-timeout", "300s")
+	AgentRolloutWait                 = NewSetting("agent-rollout-wait", "true")
+	AuthImage                        = NewSetting("auth-image", v32.ToolsSystemImages.AuthSystemImages.KubeAPIAuth)
+	AuthTokenMaxTTLMinutes           = NewSetting("auth-token-max-ttl-minutes", "0") // never expire
```
cbron, line 31: Can you add comment here too, for point of this one. Or just point people to the `ValidateMaxTTL` func which can have longer description

pkg/auth/providers/common/usermanager.go
```diff
@@ -324,7 +331,7 @@ func (m *userManager) newTokenForKubeconfig(clusterName, tokenName, description,
 		UserName:      userName,
 		AuthProvider:  userPrincipal.Provider,
 		TTL:           &ttlMilli,
-		Randomize:     false,
+		Randomize:     true,
```
cbron, line 334: Can you split up the Randomization and the Max-TTL features here into two different commits, each with a description. That way we can better isolate changes between one and the other.

pkg/auth/providers/common/usermanager.go
```diff
@@ -334,20 +341,14 @@ func (m *userManager) newTokenForKubeconfig(clusterName, tokenName, description,
...
-	}
-
-	fullCreatedToken, err := m.newTokenForKubeconfig(clusterName, tokenName, description, kind, userName, tokenTTL, userPrincipal)
+	fullCreatedToken, err := m.newTokenForKubeconfig(clusterName, tokenName, description, kind, userName, userPrincipal)
 	if err != nil {
 		return nil, "", err
 	}
 
-	_, createdTokenValue := tokens.SplitTokenParts(fullCreatedToken)
+	tokenName, createdTokenValue := tokens.SplitTokenParts(fullCreatedToken)
```
cbron, line 351: I don't understand this change, won't this not use `tokenName` that was passed in ? Or is it actually the same ?

pkg/api/steve/clusters/kubeconfig.go
```diff
@@ -74,22 +75,27 @@ func (k kubeconfigDownload) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
 		AuthProvider:  authToken.AuthProvider,
-		TTL:           nil,
+		TTL:           defaultTokenTTL,
 		Randomize:     true,
 		UserPrincipal: authToken.UserPrincipal,
 	}
-	token, err := k.userMgr.EnsureToken(input)
-	return token, err
+
+	return k.userMgr.EnsureToken(input)
```
cbron, line 100: I'm not as familiar with Steve yet, what path is this from ? Like in what case would this whole file get executed.

pkg/api/norman/customization/setting/setting.go
```diff
@@ -82,12 +81,6 @@ func Validator(request *types.APIContext, schema *types.Schema, data map[string]
 	case "auth-user-info-resync-cron":
 		_, err = providerrefresh.ParseCron(newValueString)
 	case "kubeconfig-token-ttl-minutes":
-		generateToken := strings.EqualFold(settings.KubeconfigGenerateToken.Get(), "true")
-		if generateToken {
-			return httperror.NewAPIError(httperror.ActionNotAvailable, fmt.Sprintf("kubeconfig-token-ttl-minutes can be set only if rancher doesn't generate token, "+
-				"disable kubeconfig-generate-token"))
-		}
-
```
cbron, line 90: Believe we discussed this, but for posterity, why do we want to remove the "helper" here ?

pkg/auth/tokens/manager.go
```diff
@@ -810,10 +811,11 @@ func ParseTokenTTL(ttl string) (time.Duration, error) {
 	return dur, nil
 }
 
+// ValidateMaxTTL will return the duration of the provided TTL or the duration of the max allowed TTL whichever is smaller.
 func ValidateMaxTTL(ttl time.Duration) (time.Duration, error) {
```
rmweir, line 815: I think we should name this since validate implies an error would be returned if it exceeds the MaxTTL setting. May it can be renamed to `ClampWithMaxTTL` or anything else that is more accurate.

rmweir (changes requested):

pkg/settings/setting.go
KevinJoiner: I originally did not reformat those lines, but CI failed because it was not formatted correctly. I can move the settings with comments to the end as you suggested in your previous comment, which should fix this.

pkg/auth/providers/common/usermanager.go
KevinJoiner: We are reassigning `tokenName` because the above call`newTokenForKubeconfig` now suffixes the passed in `tokenName` value with a random string. For clarity, I could have created a new variable called `randomizedTokenName`.
That being said I think it would be better to change the `newTokenForKubeconfig` function to return a token object since we are doing `tokenLister.Get()` for the token directly afterward.

pkg/api/norman/customization/setting/setting.go
KevinJoiner: I had similar thoughts. This code only stopped users from changing the default value of `kubeconfig-token-ttl-minutes` from 16 hours not from using the `kubeconfig-token-ttl-minutes` setting. Furthermore, this is only run if users use the old U, now that setting edits are sent through steve. If we want to keep this code then I would also purpose that we need to add validation to steve.

pkg/api/steve/clusters/kubeconfig.go
KevinJoiner: Download KubeConfig from the UI, but only when MCM is disabled. [Relavent PR](https://github.com/rancher/rancher/pull/33811)

pkg/settings/setting.go
KevinJoiner: Do we want to follow https://go.dev/doc/effective_go#commentary and have our comments above the declaration since they are exported? That way they work with Godoc and IDE helpers. I wouldn't mind adding comments to the other settings as well so things look more uniform.

rmweir (changes requested):

pkg/api/norman/customization/setting/setting.go
cbron: > However, the login token can be used regardless.
>  ...if we think people are using the token outside of using the CLI in the kubeconfig then this makes sense.

Confused by this. You mean if people are generating tokens in the CLI via the token command and outside of the login command ? And they can't change the TTL of those away from the default of 16 hours. 
Though I thought point of this red flag was so to inform people who were thinking it would add TTL to kubeconfigs generated from the UI.

> If we want to keep this code then I would also purpose that we need to add validation to steve.

That's a good point. I was thinking it was still using norman in new UI but it's not.

Overall I'd say we could clear this up with very strong UI messages or docs, and then remove the warning error. But I don't want to remove the warning without those.

pkg/auth/providers/common/usermanager.go
cbron: Ahh ok, so randomizedTokenName is now different from the arg tokenName.

pkg/api/steve/clusters/kubeconfig.go
cbron: @KevinJoiner can you make sure this is noted in the QA testing notes

pkg/settings/setting.go
```diff
@@ -129,12 +125,33 @@ var (
 
 	Rke2DefaultVersion = NewSetting("rke2-default-version", "")
 	K3sDefaultVersion  = NewSetting("k3s-default-version", "")
+
+	// KubeconfigGenerateToken determines whether the UI will return a generate token with kubeconfigs.
+	// If set to false the kubeconfig will contain a command to login to Rancher.
+	KubeconfigGenerateToken = NewSetting("kubeconfig-generate-token", "true")
+
+	// Deprecated: On removal use kubeconfig-default-ttl-minutes for all kubeconfigs.
+	KubeconfigTokenTTLMinutes = NewSetting("kubeconfig-token-ttl-minutes", "960") // 16 hours
```
cbron, line 134: We should still say that this setting applies only when `kubeconfig-generate-token` is set to false, or when user's use the CLI token command. Also that this will override KubeconfigDefaultTokenTTLMinutes (pretty sure that part is true).

pkg/settings/setting.go
```diff
@@ -129,12 +125,33 @@ var (
...
+
+	// KubeconfigDefaultTokenTTLMinutes is the default time to live applied to kubeconfigs created for users.
```
cbron, line 136:
```suggestion
	// KubeconfigDefaultTokenTTLMinutes is the default time to live applied to kubeconfigs created for users, and it will take effect regardless of the kubeconfig-generate-token status.
```

pkg/settings/setting.go
```diff
@@ -129,12 +125,33 @@ var (
...
+	KubeconfigDefaultTokenTTLMinutes = NewSetting("kubeconfig-default-ttl-minutes", "0") // 0 TTL = never expire
+
+	// AuthTokenMaxTTLMinutes is the max allowable time to live for tokens. Excluding those created for UI sessions which is controlled by AuthUserSessionTTLMinutes.
```
cbron, line 139: Isn't `AuthUserSessionTTLMinutes` actually capped by this one ?

pkg/auth/providers/common/usermanager.go
```diff
@@ -309,10 +309,17 @@ func (m *userManager) EnsureClusterToken(clusterName string, input user.TokenInp
 	return token.Name + ":" + key, nil
 }
 
-func (m *userManager) newTokenForKubeconfig(clusterName, tokenName, description, kind, userName string, ttl time.Duration, userPrincipal v3.Principal) (string, error) {
-	tokenTTL, err := tokens.ValidateMaxTTL(ttl)
+// newTokenForKubeconfig creates a new token for a generated kubeconfig.
+func (m *userManager) newTokenForKubeconfig(clusterName, tokenName, description, kind, userName string, userPrincipal v3.Principal) (string, error) {
+	// settings.KubeconfigTokenTTLMinutes is deprecated use tokens.GetKubeconfigDefaultTokenTTLInMilliSeconds() when the setting is removed
```
cbron, line 314: nice comment, very helpful for future us

pkg/settings/setting.go
KevinJoiner: Not currently, One place the session token is set is  [here](https://github.com/rancher/rancher/blob/836b43cec20f4a803f82982f21aa3a1f965275c6/pkg/auth/providers/publicapi/login.go#L109 ) without clamping to `AuthTokenMaxTTLMinutes`. I also ran an end-to-end test to confirm this behavior

pkg/settings/setting.go
KevinJoiner: They are separate due to the fact that KubeconfigDefaultTokenTTLMinutes will not be used for CLI token command until this deprecated setting is removed. That being said I agree we should explicitly state that `KubeconfigTokenTTLMinutes` takes president for tokens created through the CLI or when `kubeconfig-generated-token` is set to false.

pkg/api/norman/customization/setting/setting.go
KevinJoiner: [Dashboard Issue](https://github.com/rancher/dashboard/issues/6120)
[Docs PR](https://github.com/rancher/docs/pull/4140)

cbron (approved):
lgtm

pkg/settings/setting.go
```diff
@@ -261,7 +294,7 @@ func getMetadataConfig() string {
 	return string(ans)
 }
 
-// GetSettingByID returns a setting that is stored with the given id
+// GetSettingByID returns a setting that is stored with the given id.
```
rmweir, line 297: nit: some inconsistency with punctuation. I think normally we only added it if it's multiple lines 🤷 .

rmweir (approved):
One nit, otherwise LGTM.

cbron:
Let's move this over to QA now.

PR #37713, opened 2022-05-13
Adding a new endpoint to generate a supportconfig bundle

MbolotSuse:


pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+	authv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
+)
+
+const (
+	// Endpoint The endpoint that this URL is accessible at - used for routing and for SARs
+	Endpoint             = "/v1/generateSUSERancherSupportConfig"
+	cspAdapterNamespace  = "cattle-csp-system"
+	cspAdapterConfigmap  = "csp-config"
+	tarContentType       = "application/x-tar"
+	unauthenticatedError = "unauthenticaed"
```
cmurphy, line 27:
```suggestion
	unauthenticatedError = "unauthenticated"
```

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+	logrus.Infof("Generating supportconfig")
+	archive, err := h.generateSupportConfig()
+	logrus.Infof("Done Generating supportconfig")
+	if err != nil {
+		logrus.Errorf("Error when generating support config %v", err)
+		writer.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	writer.Header().Set("Content-Type", tarContentType)
+	writer.Header().Set("Content-Disposition", "filename=\"supportConfig.tar\"")
```
cmurphy, line 70: Does this need to include `attachment;`? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+	n, err := io.Copy(writer, archive)
+	logrus.Infof("wrote %v bytes in archive response", n)
+}
+
+// authorize checks to see if the user can get the non-resource url of Endpoint. If not, returns an error, otherwise nil
+func (h *GeneratorHandler) authorize(r *http.Request) error {
```
cmurphy, line 76: Suggestion - it might help with consistency with k8s/rancher to use some of the other `Authorize` methods for inspiration -

https://pkg.go.dev/k8s.io/apiserver/pkg/authorization/authorizer#AuthorizerFunc.Authorize
https://github.com/rancher/rancher/blob/8a31d3cc1bf0830439d4926cd06d8fccfb9c2c54/pkg/api/steve/aggregation/auth.go#L50
https://github.com/rancher/rancher/blob/8a31d3cc1bf0830439d4926cd06d8fccfb9c2c54/pkg/tunnelserver/mcmauthorizer/tunnel.go#L108

where they return multiple values, one to indicate an error and another to indicate an auth decision. This would be cleaner to parse above when deciding what HTTP error to return. Another thing they do is treat missing the token/auth context as the same as a forbidden error instead of a separate unauthenticated error, which also simplifies the code.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		if _, err := tw.Write(configData); err != nil {
+			return nil, err
+		}
+	}
+
+	return &buf, nil
+}
+
+//getCspConfig gets the configmap produced by the csp-adapter
```
cmurphy, line 148:
```suggestion
// getCspConfig gets the configmap produced by the csp-adapter
```

pkg/api/steve/supportconfigs/supportconfigs.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// supportConfigHandler contains the necessary clients/controllers for retrieving a support config
+type supportConfigHandler struct {
+}
+
+// getSupportConfig retrieves a supportConfig
+func (h *supportConfigHandler) getSupportConfig() supportConfig {
+	return supportConfig{
+		SupportEligible: false,
```
cmurphy, line 27: Why is this different? https://github.com/rancher/csp-adapter/blob/d8ef954b0e111203872bec9cf5dc1c5e09d59c93/pkg/supportconfig/generator.go#L31

pkg/api/steve/supportconfigs/supportconfigs.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// supportConfig represents the serializable struct of support related information common to all rancher setups
+type supportConfig struct {
+	// SupportEligible - if this rancher product is eligible for support.
+	SupportEligible bool `json:"support_eligible,omitempty"`
+	// Platform - the platform (arm, x86_64, etc) rancher is running on. As of now, always x86
+	Platform string `json:"platform"`
+	// Product - suse product string identifying rancher - see getRancherProductString
+	Product string `json:"product"`
```
cmurphy, line 17: Do we need these https://github.com/rancher/csp-adapter/blob/d8ef954b0e111203872bec9cf5dc1c5e09d59c93/pkg/supportconfig/generator.go#L32-L38 ?

pkg/api/steve/supportconfigs/supportconfigs.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/rancher/rancher/pkg/version"
+)
+
+// supportConfig represents the serializable struct of support related information common to all rancher setups
+type supportConfig struct {
```
cmurphy, line 11: Is there any reason the types we already had weren't ported over to this? https://github.com/rancher/csp-adapter/blob/release/v2.6/pkg/types/supportconfig.go

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+	cspConfig, err := h.getCspConfig()
+	// csp adapter won't always be installed
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+	for key, value := range cspConfig {
+		// prefer csp's values - CSP config can see rancher's values and the Cloud Providers
```
cmurphy, line 116: Can you explain more about this - I see the merge examples in the PR comments but I don't understand functionally what data is being posted to a configmap and what needs to be merged.

cmurphy (commented):
Leaving some questions inline, I still am not fully ramped up on the background of the CSP adapter so not all of this is obvious to me.

pkg/api/steve/supportconfigs/supportconfigs.go
MbolotSuse: The requirements for this value is that `SupportEligible` is missing from the supportconfig bundle in the standard case. If the csp-adapter is installed, the value should be present and set to true. To support this requirement, I set the supportconfig struct here to have an `omitempty` tag for this value and defaulted it to false (which omits it from the output). Because we merge values from the CSP adapter, the adapter will produce a configmap with this value set to true (making the value true when that value is installed).

The other repo is for the csp adapter. Meaning that the endpoint used to be present only when the adapter was installed. Now that this will be present on any rancher install, this PR needs to be able to handle producing supportconfigs for rancher installs that have the CSP adapter or don't have the csp adapter.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: Here's an overview: 
- There are some values that need to be included in the supportconfig that only the CSP adapter knows. This primarily involves things related to the AWS account that the CSP adapter is installed for (see notes below). When the CSP adapter is not installed, we don't want to use these values. But when they are present, we need to include them.
- In order to support situations where the CSP adapter needs to override a value that is present in the standard supportconfig that rancher produces (such as the support eligible field), this merge process exists to ensure that the values produced by the CSP adapter, which is able to see stuff in k8s and stuff in the cloud provider that it is installed for, are preferred over the ones that rancher produces.

Specific Functionality of the CSP adapter for this support config:
- You can see these fields in code [here](https://github.com/rancher/csp-adapter/blob/88c47fd39dc08ae8daa7d4ecc619cc033f1f9d5a/pkg/manager/types.go). However, keep in mind that from rancher's perspective the CSP adapter can produce anything into this map. Rancher will pick it up and include it while replacing its own values with what is in the map. 
- This includes `SupportEligible bool` - a bool on if this install is eligible for support (always true if CSP adapter is installed)
- Nested Field: CSP - info describing the Cloud Provider setup
  - `Name string` the name of the cloud provider that the csp adapter is using (always `aws` for now)
  - `AcctNumber string` the account number that the CSP adapter is checking for compliance 
- Nested Field: Compliance - info describing the compliance state of rancher
  - `Status string`. If the account is in compliance with the license that it has purchased. Either "Compliant" or "NonCompliant".
  - `Message string`. The message describing why this is in `Status`.

Notes on the CSP adapter:
- The purpose of the CSP adapter (working on better docs, I'm sorry that this isn't better right now) is to track license usage inside of aws. Basically, we want users to be able to purchase support for an arbitrary number of nodes managed by rancher (it's not totally aribitrary but I'll leave those details for the CSP adapter PR). The adapter tracks the usage of these nodes and compares this total to the number that you are entitled to through your license purchases in AWS. If you are using more nodes then you are entitled to, it warns the user and produces output in the supportconfig so that when you open a ticket, support knows if you are outside of compliance. 
- To do this tracking the csp adapter needs to have permissions in the cloud provider. Because of this permissions model and the fact that this functionality is completely optional, we are not including it within core rancher and it is instead a separate chart.

pkg/api/steve/supportconfigs/supportconfigs.go
MbolotSuse: No. These values are provided by the CSP adapter. Rancher lacks the credentials to gather the information on which aws account this is registered to and it lacks the logic to calculate if it is in compliance with the licenses purchased. 

In short - these values are produced by the csp adapter.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: I've removed this to comply with [this comment](https://github.com/rancher/rancher/pull/37713#discussion_r874238725)

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: Good point. I have changed this to return a bool and an error. In addition, I no longer parse for unauthenticated cases. Now if authorized returns false or an error then we still give back a `403`. We do log an error rancher side for the error cases, but that isn't exposed to the user.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: I have made this change.

pkg/api/steve/supportconfigs/supportconfigs.go
cmurphy: In that case it seems like there is a bit of duplication between the Rancher side of this and the CSP adapter side, given that they are using almost-but-not-quite-the-same types. It would be nice if that could be reduced somewhat.

pkg/api/steve/supportconfigs/supportconfigs.go
cmurphy: Is the supportconfig actually meaningful if the csp adapter isn't installed? Could this be simplifed by just returning a 404 if the csp adapter config doesn't exist?

cmurphy (commented):
I've left a few more questions. If there was some documentation or design document I could reference I would be in a better place to review this sanely.

pkg/api/steve/supportconfigs/supportconfigs.go
MbolotSuse: From the perspective of the code, these aren't really using the same types. I've intentionally made this use a map-based merge rather than using a well-defined struct so that the CSP adapter can supply whatever values that it feels are appropriate. 

The csp adapter does already need to import rancher (to create the user notifications), so in theory I could use rancher as the source of what fields are available from an overall perspective. However, I would want to avoid this since it would require changes in rancher's side for new fields that the csp adapter feels are appropriate.

For example, let's say that I port over the entire type that used to be in the csp adapter into core rancher. If the adapter needs to supply a new field (say something specific to the aws license like "Number of Licenses purchased"), I would first need to add the field in rancher, then update the adapter's dependencies, and then I could supply the new field. That seems like a lot of extra work.

The Design I was going for here was for Rancher to define a few things itself and then include whatever arbitrary values the csp adapter supplied. I know that this isn't the clearest model, but it is more flexible and allows the adapter to decide what fields are relevant, rather than requiring rancher to be an overriding source of truth that the adapter relies upon.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: > I'm afraid I'm still missing something. Below it looks like we're taking a JSON file from a configmap, converting that JSON file into a tarball, and delivering that as a response. It looks like the content of this JSON file is a few small pieces of information such as the product and platform names.

You've got it. The csp adapter supplies a few other pieces of information (what aws account you are in and if you have gone over on your node allotment), but this is about the scope of it. 

> There is a SUSE-owned tool called [supportconfig](https://github.com/openSUSE/supportutils/blob/master/bin/supportconfig) that does a deep analysis and collection of a host's configuration and logs. This tool is widely used by various support teams. Is this tool at all related to what is being delivered by this endpoint? If so, what component is running the tool? If not, let's please choose a different name for this endpoint.

To answer your questions directly:
- The suse owned tool is not used/ran in any way for this process.
- I think that the idea here was that rancher needs to be able to supply a supportconfig. As you know, rancher operates at a different part of the infra stack than the OS - so we need to collect different data. This endpoint will allow us to produce data that has equivalent relevancy to us as the supportconfig produces for the OS world.
- PM Explicitly calls this a supportconfig. I would prefer to keep the language between the code and the design consistent. Since I can't change the source documents, I think that it is unwise to maintain a separate name in the codebase for what is called for by the design requirements.

pkg/api/steve/supportconfigs/supportconfigs.go
cmurphy: Okay, this makes more sense in context

pkg/api/steve/supportconfigs/supportconfigs.go
cmurphy: Okay :+1:

cmurphy (commented):
Thanks for the additional context. I have one possible suggestion for the authorization that might be worth addressing before merging this PR if you agree, and a few other nitpicks.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,168 @@
...
+	for name, body := range files {
+		hdr := &tar.Header{
+			Name: name,
+			Mode: 0600,
+			Size: int64(len(body)),
+		}
+		if err := tw.WriteHeader(hdr); err != nil {
+			return nil, err
+		}
+		if _, err := tw.Write(configData); err != nil {
```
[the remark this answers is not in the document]
MbolotSuse: Good catch - I think that this actually would have included a bug if we started to do multiple files (every file would have just had the config data as the value instead of the body as intended). I've fixed accordingly to go along with your suggestion.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,168 @@
...
+	if !response.Status.Allowed {
+		return false, nil
+	}
+	return true, nil
+}
+
+// generateSupportConfig produces an io.Reader with the supportConfig ready to be returned using a http.ResponseWriter
+// Returns an err if it can't get the support config (note: does not return an err if it can't find the csp config)
+func (h *GeneratorHandler) generateSupportConfig() (io.Reader, error) {
+	mergedConfig, err := h.ConfigHandler.getSupportConfigAsMap()
```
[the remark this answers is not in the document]
MbolotSuse: done - changed to `rancherSupportConfig`.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to create sar %s", err)
+	}
+	if !response.Status.Allowed {
+		return false, nil
+	}
+	return true, nil
+}
+
+// generateSupportConfig produces an io.Reader with the supportConfig ready to be returned using a http.ResponseWriter
```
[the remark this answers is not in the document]
MbolotSuse: Done.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,168 @@
...
+// authorize checks to see if the user can get the non-resource url of Endpoint. If not, returns an error, otherwise nil
+func (h *GeneratorHandler) authorize(r *http.Request) (bool, error) {
+	userInfo, ok := request.UserFrom(r.Context())
+	if !ok {
+		return false, fmt.Errorf("unable to extract user info from context")
+	}
+	response, err := h.SubjectAccessReviews.Create(r.Context(), &authzv1.SubjectAccessReview{
+		Spec: authzv1.SubjectAccessReviewSpec{
+			// This auth method prevents collision with future k8s resources and allows this to be added to custom rules
+			NonResourceAttributes: &authzv1.NonResourceAttributes{
```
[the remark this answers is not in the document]
MbolotSuse: > Now that I understand that the entire supportconfig will be contained in one configmap with a predictable name and namespace
One correction here - rancher still produces some values (i.e. arch/product version) on it's own. The adapter does a lot of the 

This would make a bit more sense, but there's a couple of edge cases that make me unsure of this route. The big issue here is that the configmap/csp adapter namespace is only present when the adapter is installed, but we still need to be able to produce a limited supportconfig even without it. I'm unsure about the ability of a user to grant access to a configmap in a namespace which doesn't exist.

The other issue would be controlling access through custom role. Given that our global roles map to cluster roles in the local cluster, I think that it would be challenging to grant access to a custom admin role. You might be able to get away with it by granting access to every configmap with that hardcoded name in every namespace, but that will be challenging for the customer to figure out and would grant more access than intended. 

That being said, I'm flexible on this since this does meet the requirements and is less abusive of the k8s api. Let me know what your thoughts are.

cmurphy (approved):
lgtm

pkg/api/steve/supportconfigs/supportconfigs.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// getSupportConfig retrieves a supportConfig
+func (h *supportConfigHandler) getSupportConfig() supportConfig {
+	return supportConfig{
+		SupportEligible: false,
+		Platform:        h.getRancherPlatform(),
+		Product:         h.getRancherProductString(version.Version),
+	}
+}
```
cbron, line 31: Nit: since primary entrypoint to this method is `getSupportConfigAsMap`, I would put that as first method.

pkg/api/steve/supportconfigs/supportconfigs.go
```diff
@@ -0,0 +1,57 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/rancher/rancher/pkg/version"
+)
+
+// supportConfig represents the serializable struct of support related information common to all rancher setups
+type supportConfig struct {
+	// SupportEligible - if this rancher product is eligible for support.
+	SupportEligible bool `json:"support_eligible,omitempty"`
```
cbron, line 13: Add the extra business logic conditions here in comment. IE: always true if adapter is installed, not present in json at all if false.

pkg/api/steve/supportconfigs/generator.go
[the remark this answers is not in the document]
cbron: One thing that is messing with our design is the idea that any Rancher in the future (non-CSP included) should have the "get support config" button which downloads a support config tar with logs included. So that case makes more sense for SAR route, unless we pre-package this configmap.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,168 @@
...
+				Verb: "get",
+				Path: Endpoint,
```
cbron, line 83: To summarize, this endpoint isn't set or given anywhere, the idea is that no one should have access to this endpoint besides admin, is that right ?

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,168 @@
...
+		SubjectAccessReviews: scaledContext.K8sClient.AuthorizationV1().SubjectAccessReviews(),
+		ConfigHandler:        supportConfigHandler{},
+	}
+}
+
+func (h *GeneratorHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	authorized, err := h.authorize(request)
+	if err != nil {
+		_, _ = writer.Write([]byte("forbidden"))
+		writer.WriteHeader(http.StatusForbidden)
```
cbron, line 48: nit: 401, no user was authorized. The 403 on line 54 seems correct.

pkg/api/steve/supportconfigs/generator.go
cbron: Could we distill this down into some code comments if possible. Not mandatory since we can always look at PR, but makes reading code in IDE faster.

pkg/api/steve/supportconfigs/supportconfigs.go
rmweir, line 13: Change to camelcase if possible as I believe that's what we usually use for json fields.

pkg/api/steve/supportconfigs/supportconfigs.go
```diff
@@ -0,0 +1,57 @@
...
+	// Platform - the platform (arm, x86_64, etc) rancher is running on. As of now, always x86
+	Platform string `json:"platform"`
+	// Product - suse product string identifying rancher - see getRancherProductString
+	Product string `json:"product"`
+}
+
+// supportConfigHandler contains the necessary clients/controllers for retrieving a support config
+type supportConfigHandler struct {
+}
```
rmweir, line 22: Looks like we don't need this struct since it doesn't contain anything. Remove if that is the case.

rmweir (changes requested):

pkg/api/steve/supportconfigs/supportconfigs.go
MbolotSuse: I put this here because future uses of the supportconfig (i.e. retrieving logs) definitely will need to have stuff here (i.e. fetching the logs will require some form of k8s client). Because of this, I would prefer to keep it. Let me know if you feel otherwise.

pkg/api/steve/supportconfigs/supportconfigs.go
MbolotSuse: I'll need to double check the requirements to see if I can change this, but that's a good callout. I will double check and see if we can change this.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: @cbron which parts would you like in the comments? I can't really do a full schema because rancher doesn't know the full schema. I think that the history/naming of the supportconfig might be a bit much. 

I can add: 
- More comments on the fields that rancher does know about (I have some already, if you think they are not sufficient I can add more)
- Comments on on what we expect the supportconfig to do and why we would merge it's results

Does that fit what you were thinking?

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: Correct. The reason that I did it this way is technically you could use K8s rbac to add this to another role if you really wanted to. We don't, but a customer could.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: I originally had a setup which distinguished between 401 (can't get creds), 403 (got creds SAR denied) and 500 (I got an error when trying to auth). @cmurphy Pointed out that at this stage similar functionality returns a 403 as a catch-all.

I can revert this to an earlier setup, but I do think that 403 makes sense in this context given that we are protected by an authn route earlier on. I'll bring this up in standup so we can make sure everyone is on the same page.

cbron: Na, just leave as 403 everywhere. There's also the theory that we should return a 404 everywhere here as to avoid "leaking" a valid endpoint to a person without access, but that's trading security for a poor UX.

pkg/api/steve/supportconfigs/supportconfigs.go
MbolotSuse: @rmweir 

Took a look at this. I don't have any specific documentation on the names of these fields. However, I did not name these fields - this was taken from earlier work that Ryan did. I would prefer not to change this given that there is going to be a program which validates the supportconfig files produced by another area - I'm assuming that Ryan had more information then I do and put this here for a reason. 

In short - the available info that I have indicates that we need a specific schema and does not indicate camel case. I would prefer leave as is rather than making a somewhat risky change in order to meet our naming conventions. Please let me know if you disagree.

MbolotSuse:
Made some changes to how this works in order to support the "rancher is down and we need a config to contact support" use case.

Now, the entirety of the information is contained the configmap produced by the csp adapter. All this endpoint does is retrieve the json from the configmap, zip it up and return it. 

The RBAC check now checks to see if you can get the `csp-config` configmap in `cattle-csp-system`. If you can, you can use this endpoint. If you can't, you get a 403.

If you are authorized to get the configmap but we get a not found when trying to get the configmap, we return a 501 and indicate that you need to install the csp-adapter (now that all of the info is provided by the adapter, you have to install it to get anything from the config).

FAQ:
1. Do we need the endpoint at all now that this is just a configmap (i.e. can't we just get that through norman/steve)?
We don't strictly need the endpoint, but without this the UI team will have to do the bundling/zipping of the configfile. That sounds sub-optimal to me.

cmurphy (approved):
Left some small suggestions but they are just nitpicks, not worth addressing if there is otherwise no new revision needed.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,167 @@
...
+	if !authorized {
+		_, _ = writer.Write([]byte("forbidden"))
+		writer.WriteHeader(http.StatusForbidden)
+		return
+	}
+	logrus.Infof("[%s] Generating supportconfig", logPrefix)
+	archive, err := h.generateSupportConfig()
+	logrus.Infof("[%s] Done Generating supportconfig", logPrefix)
+	if err != nil {
+		if err.Error() == notImplemented {
```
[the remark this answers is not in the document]
MbolotSuse: Good call - I've adjusted this accordingly.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,167 @@
...
+			_, _ = writer.Write([]byte("csp-adapter was not installed, but must be to generate supportconfigs"))
```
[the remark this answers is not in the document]
MbolotSuse: Change made.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,167 @@
...
+			writer.WriteHeader(http.StatusNotImplemented)
+			return
+		}
+		logrus.Errorf("[%s] Error when generating support config %v", logPrefix, err)
```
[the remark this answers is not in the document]
MbolotSuse: Change made.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,167 @@
...
+		_, _ = writer.Write([]byte("Internal Server Error"))
+		writer.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	writer.Header().Set("Content-Type", tarContentType)
+	writer.Header().Set("Content-Disposition", "attachment; filename=\"supportConfig.tar\"")
```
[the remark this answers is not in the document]
MbolotSuse: change made.

cbron (approved):

cmurphy (approved):

MbolotSuse:
Had to push a change to do the following:
- Added the csp-adapter namespace to the system-namespace
- Changed the return error logic to use util.ReturnError() (other way was not displaying in browser right).

This should now be ready to merge - going to re-request one more time (sorry).

cmurphy (approved):

cbron (approved):

cmurphy (approved):

cbron (dismissed):

pkg/api/steve/supportconfigs/supportconfigs.go
rmweir: It seems unnecessary and I'm not sure we should leave code stubs in hopes that it influences future additions. @cbron  or @cmurphy thoughts?

MbolotSuse: Sorry - I did not update this thread. I have removed this struct - this endpoint no longer generates any values itself (it gets it all from the csp configmap). In light of this, I don't think this thread is relevant any longer. If you disagree, let me know.

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -0,0 +1,166 @@
...
+		if _, err := tw.Write(body); err != nil {
+			return nil, err
+		}
+	}
+
+	return &buf, nil
+}
+
+// getCspConfig gets the configmap produced by the csp-adapter
+func (h *GeneratorHandler) getCspConfig() (map[string]interface{}, error) {
```
rmweir, line 143: nit: with acronyms as CSP I think it makes sense to keep them all the same case, here that would be upper.

rmweir (changes requested):
1 nit, otherwise LGTM

MbolotSuse:
Pushed a change to resolve @rmweir's comments, re-requesting reviews.

rmweir (approved):
LGTM

cbron (approved):

PR #37725, opened 2022-05-13
Fix Cluster Templates - private registry support without override option enabled

### LINKED ISSUE #37708: RKE Cluster Templates - fail to update private registry URL when the override option is not enabled
# Update 

Rancher version: v2.6.5

Steps:
- Create an RKE1 cluster template with a private registry configured without enabling any override functionality on the private registry.
- Create a cluster using this cluster template
- Create a cluster template revision and modify the credentials or URL of the private registry, notice that the override functionality is still disabled
- Edit the cluster and use the newly created cluster template revision

Results:
 New private registry settings won't be applied

Expected Results:
The new private registry settings should be applied

Screeshot of the section in the cluster template revision creation page: 
image

---
# Original Post: 

https://github.com/rancher/rancher/pull/37606/files

Repro:

1. Create RKE Template with 1 private registry
2. Apply RKE Template to a Cluster
3. Modify RKE Template and add a second registry or modify the old and make sure override is disabled
4. New Registry won't be applied

Code Issue

Loading old registry details from cluster into registryMap:
https://github.com/rancher/rancher/blob/c4d59fa88e2313e1446ba542a21c39e15a17f375/pkg/api/norman/store/cluster/cluster_store.go#L436-L440

Old registry details from cluster are used to override the new private registry details in dataFromTemplate
Settings from the cluster template revision in dataFromTemplate are gone and old credentials are used
https://github.com/rancher/rancher/blob/c4d59fa88e2313e1446ba542a21c39e15a17f375/pkg/api/norman/store/cluster/cluster_store.go#L518-L525

@jiaqiluo Could you have a look?

zerotens:
Issue: 

 https://github.com/rancher/rancher/issues/37708

# Problem
 

 
For REK clusters that are created from a cluster template revision,  the private registry info ( url, username, or password) from the existing cluster is always used when updating the cluster, this causes the problem in this very specific scenario: if the private registry info is changed in the newly-applied cluster template revision AND those fields are not marked as "allow user override", the new values will never be applied. 

# Solution
 

 
The private registry info from the existing cluster is used only when the "allow user override" is enabled for the private registry info in the cluster template revision; otherwise, apply the values from the revision. 

# Testing

Steps (exactly same to the one in the issue):
- Create an RKE1 cluster template with a private registry configured without enabling any override functionality on the private registry.
- Create a cluster using this cluster template
- Create a cluster template revision and modify the credentials or URL of the private registry, notice that the override functionality is still disabled
- Edit the cluster and use the newly created cluster template revision
 
Results:
New private registry settings are applied 

---

# Original Post

Partially fixes #37708

Repro:

1. Create cluster template with a private registry configured without enabling any override functionality on the private registry.
2. Create a cluster using this cluster template
3. Create a cluster template revision and modify the credentials of the private registry
4. Edit the cluster and use the newly created cluster template revision
5. New private registry settings won't be applied

With this pr the private registry override code won't be applied if override is not enabled in the selected cluster template revision.

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -435,6 +435,7 @@ func loadDataFromTemplate(clusterTemplateRevision *v3.ClusterTemplateRevision, c
 	// The key in the map is used to preserve the order of registries
 	registryMap := make(map[int]map[string]interface{})
 	existingRegistries := convert.ToMapSlice(convert.ToMapInterface(existingCluster[managementv3.ClusterSpecFieldRancherKubernetesEngineConfig])[managementv3.RancherKubernetesEngineConfigFieldPrivateRegistries])
+	privateRegistryOverride := false
 	for i, registry := range existingRegistries {
 		registryMap[i] = registry
 	}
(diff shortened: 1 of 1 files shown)


jiaqiluo (approved):
LGTM

@zerotens 
Thank you for your contribution!

Issue #36492, opened 2022-02-11
RKE1 etcd snapshot failure retention


jakefhyde:
**Rancher Server Setup**
- Rancher version: v2.6-head
 
 **Information about the Cluster**
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE1
   
**Describe the bug**

RKE1 etcd snapshot config specifies a `retention`, which is the maximum number of snapshots retained until rancher starts to delete older snapshots. Currently, there is no check ensuring that rancher is retaining at least `retention` successful snapshots, creating the possibility that rancher has no successful snapshots. Failed snapshots should be retained separately from successful snapshots, so that the user doesn't end up in a situation where the only snapshots are failed ones.

**To Reproduce**
 
- Provision a cluster with a retention of 2 and a snapshot interval of 1 hour.
- Wait for initial etcd snapshot with an s3 endpoint
- Watch as the backup is successful
- Delete the s3 bucket/folder
- Wait for next recurring etcd snapshot
- Snapshot should fail with error status
- Wait for next recurring etcd snapshot
- Watch as last successful snapshot is deleted to make room for failed snapshot

**Result**

The last successful snapshots is deleted, leaving 2 failed snapshots.

**Expected Result**

There are at least `retention` successful snapshots retained (how many failed snapshots should be retained requires discussion)

**Screenshots**

**Additional context**

jiaqiluo:
We should count and keep the successful and failed snapshots separately up to the limit.

jakefhyde:
## Root cause

Previously, failed snapshots were counted towards the total # of recurring backups specified in a cluster's `spec.BackupConfig.Retention`. This meant that for some cluster configurations, it would be possible for backups to start failing (for example, in the event that the s3 credential was no longer authorized), and no usable backups to be available for restore.

## What was fixed, or what changes have occurred

A fix was temporarily put in to retain failed etcd backups indefinitely, separate from what is specified in the `retention` field. Obviously this fix could not be permanent, as it would create significant bloat to etcd after some time, with no way easy for users to delete failed backups. This was fixed by retaining `retention` number of failed recurring backups as well. Failed manual backups are still retained indefinitely, for the same reason that a manually requested backups is more important than a recurring one, and also a less frequent operation.

## Areas or cases that should be tested
[...]

PR #37729, opened 2022-05-13
Added retention for failed snapshots

jakefhyde:
## Background

Previously, retention was limited to the total number of recurring etcd backups, which could leave a cluster without any successful etcd backups if the cluster entered a bad state (for example, if the etcd container was stopped on the downstream node, a backup would never be taken, but it would be counted towards the total). An attempt was made to remediate this by not removing failed recurring backups, temporarily. This PR follows parity with RKE2, by using the retention set for the cluster as the number of successful backups to keep, as well as the number of failed backups to keep (e.x. a retention of 5 means potential of 5 successful and 5 failed backups).

## Additional notes

Some minor refactoring and improvements have been made, mostly notably that we no longer rotate backups in cluster sync, but rather immediately after the backup is attempted.

Related Issue: https://github.com/rancher/rancher/issues/36492

### DIFF
--- a/pkg/controllers/management/etcdbackup/etcdbackup.go
+++ b/pkg/controllers/management/etcdbackup/etcdbackup.go
@@ -162,6 +162,24 @@ func (c *Controller) clusterBackupSync(ctx context.Context, interval time.Durati
 	return nil
 }
 
+type FilterFunc = func(backup *v3.EtcdBackup) bool
+
+// filterBackups will test each backup within backups to determine if the filter validates (i.e. returns true) for the
+// given backup. If any filter is false, the backup is considered invalid.
+func filterBackups(backups []*v3.EtcdBackup, filter ...FilterFunc) []*v3.EtcdBackup {
+	ret := make([]*v3.EtcdBackup, 0)
+OUTER:
+	for _, backup := range backups {
+		for _, f := range filter {
+			if !f(backup) {
+				continue OUTER
+			}
+		}
+		ret = append(ret, backup)
+	}
+	return ret
+}
+
 func (c *Controller) runWaitingBackups(cluster *v3.Cluster) error {
 	if cluster == nil || cluster.DeletionTimestamp != nil {
 		return nil
@@ -204,7 +222,9 @@ func (c *Controller) createRecurringBackup(cluster *v3.Cluster) error {
 	if err != nil {
 		return err
 	}
-	recurringBackups := getRecurringBackups(backups)
+
+	recurringBackups := filterBackups(backups, IsBackupRecurring)
+	chronologicalSort(recurringBackups)
 
 	// cluster has no recurring backups, we need to create initial backup
 	if len(recurringBackups) == 0 {
@@ -241,7 +261,11 @@ func (c *Controller) createRecurringBackup(cluster *v3.Cluster) error {
 		log.Debugf("[etcd-backup] new backup created: %s", newBackup.Name)
 	}
 
-	return c.rotateExpiredBackups(cluster, recurringBackups)
+	return nil
+}
+
+func IsBackupRecurring(backup *v3.EtcdBackup) bool {
+	return !backup.Spec.Manual
 }
 
 func (c *Controller) createBackupForCluster(b *v3.EtcdBackup, cluster *v3.Cluster) (runtime.Object, error) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -162,6 +162,21 @@ func (c *Controller) clusterBackupSync(ctx context.Context, interval time.Durati
...
+func filterBackups(backups []*v3.EtcdBackup, filter ...FilterFunc) []*v3.EtcdBackup {
+	ret := []*v3.EtcdBackup{}
+	for _, backup := range backups {
+		for _, f := range filter {
+			if !f(backup) {
+				continue
+			}
+		}
+		ret = append(ret, backup)
+	}
```
thedadams, line 176: It seems that this implementation will add every backup to the return slice, regardless of the output of the filter functions.

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -162,6 +162,21 @@ func (c *Controller) clusterBackupSync(ctx context.Context, interval time.Durati
 	return nil
 }
 
+type FilterFunc = func(backup *v3.EtcdBackup) bool
+
+func filterBackups(backups []*v3.EtcdBackup, filter ...FilterFunc) []*v3.EtcdBackup {
```
thedadams, line 167: A comment regarding how the filter functions are handled (or-ed versus and-ed) for each backup would be helpful.

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -162,6 +162,21 @@ func (c *Controller) clusterBackupSync(ctx context.Context, interval time.Durati
...
+	ret := []*v3.EtcdBackup{}
```
thedadams, line 168:
```suggestion
	ret := make([]*v3.EtcdBackup, 0)
```

thedadams (changes requested):

pkg/controllers/management/etcdbackup/etcdbackup.go
jakefhyde: Good catch, this actually seems like a good candidate for a unit test

jiaqiluo (approved):
LGTM.
Nit: Please squash commits into one before merging.

jakefhyde:
Had to pull because we removed Windows 20H2 support

jiaqiluo (approved):

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -162,6 +162,24 @@ func (c *Controller) clusterBackupSync(ctx context.Context, interval time.Durati
...
 
+type FilterFunc = func(backup *v3.EtcdBackup) bool
+
+// filterBackups will test each backup within backups to determine if the filter validates (i.e. returns true) for the
+// given backup. If any filter is false, the backup is considered invalid.
+func filterBackups(backups []*v3.EtcdBackup, filter ...FilterFunc) []*v3.EtcdBackup {
+	ret := make([]*v3.EtcdBackup, 0)
+OUTER:
+	for _, backup := range backups {
+		for _, f := range filter {
```
annablender, line 173: Nit: change f to a non single char variable for readability

pkg/controllers/management/etcdbackup/etcdbackup.go
```diff
@@ -400,15 +425,44 @@ func (c *Controller) etcdRemoveSnapshotWithBackoff(b *v3.EtcdBackup) error {
 	})
 }
 
-// rotateExpiredBackups removes backups that are older than the expiration period, while retaining the desired number of etcd backups.
-// This function expects backups to be sorted from newest to oldest. In practice this function should only delete the last backup,
-func (c *Controller) rotateExpiredBackups(cluster *v3.Cluster, backups []*v3.EtcdBackup) error {
+func (c *Controller) rotateSuccessfulBackups(cluster *v3.Cluster) error {
```
annablender, line 428: Nit: Add func descriptions to the funcs (at least the larger ones) that you've added. A comment was deleted here instead of modified

annablender (approved):

pkg/controllers/management/etcdbackup/etcdbackup.go
snasovich: `f` is quite a common name for a "function" variable so I think it's fine to keep it as `f`.

pkg/controllers/management/etcdbackup/etcdbackup.go
jakefhyde: Yes, I'll be flushing out the comments in a future PR (likely the one where I create the integration/validation tests)

PR #37735, opened 2022-05-14
Update k3s and rke2 terraform to use user home directory for scripts

rancher-max:
This allows us to run this automation on sle micro with selinux. Previously, selinux was blocking scripts from the /tmp/ directory due to https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/security-enhanced_linux/sect-security-enhanced_linux-targeted_policy-unconfined_processes.

Also remove the plans because they are unnecessary and just an extra file that gets created.

### DIFF
--- a/tests/validation/tests/v3_api/resource/terraform/k3s/master/instances_server.tf
+++ b/tests/validation/tests/v3_api/resource/terraform/k3s/master/instances_server.tf
@@ -1,3 +1,7 @@
+locals {
+  user_home = "/home/${var.aws_user}"
+}
+
 resource "aws_db_parameter_group" "db-parameters" {
   count                  = (var.cluster_type == "etcd" ? 0 : (var.external_db != "aurora-mysql" ? 1 : 0))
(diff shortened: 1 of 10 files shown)


tests/validation/tests/v3_api/resource/terraform/k3s/master/instances_server.tf
```diff
@@ -63,6 +67,7 @@ resource "aws_instance" "master" {
     user                 = var.aws_user
     host                 = self.public_ip
     private_key          = file(var.access_key)
+    script_path          = "${local.user_home}/terraform_caller.sh"
```
ShylajaDevadiga, line 70: Can we add comments on the purpose and functionality of this script?

rancher-max:
I'm closing this as there are other requirements needed to get this actually working (reboot, then do the next logic in a different terraform block), and as part of it we should do an upgrade in our terraform version, which will happen in k3s/rke2 standalone repositories.

Issue #37752, opened 2022-05-17
Manage Cluster backups role not applied on RKE2 clusters


KevinJoiner:
**Rancher Server Setup**
- Rancher version: v2.6-8a31d3cc1bf0830439d4926cd06d8fccfb9c2c54-head
- Installation option (Docker install/Helm Chart): Docker

**Information about the Cluster**
- Kubernetes version: v1.23.6
- Cluster Type (Local/Downstream): Downstream single node digital ocean RKE2 cluster

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: User with "Manage Cluster Backups" Role

   
**Describe the bug**

"Manage Cluster Backups" custome role does not allow a users to manage backups in a downstream RKE2 cluster

**To Reproduce**
 
- Provision a downstream RKE2 cluster
- Create a new standard user "test-user"
- Add the "test-users" to the cluster with the Custom permission "Manage Cluster Backups" .
- As admin create a snapshot `"Cluster Management" - > "RKE2 cluster Name" -> "Snapshots" -> "Snapshot Now"`
    -  **This step is only need because issue rancher/rancher#37630 **
- Login as "test-user"

**Result**
test-user is unable to view the snapshot tab 

**Expected Result**
test-user can view and create snapshots

KevinJoiner:
## Root Cause
The Manage Cluster Backups role did not have the correct permissions for managing cluster backups on RKE2 clusters.

## What was fixed, or what change has occurred
Added correct permissions for the RKE2 backup CRD `etcdsnapshot.cattle.io` to the built-in roleTemplate

## Areas or cases that should be tested
Manage Cluster Backup role can create and view cluster backups.

1. Verify 'Manage Cluster Backup' role can view and create snapshots
   - Provision a new cluster using RKE2 
   - Create a new User (testUser) 
   - Assign testUser to the custom role 'Manage Cluster Backup'
   - Login as testUser.
   - Verify the snapshots tab is present.
   - Verify the user can view an empty list of snapshots
   - Verify the user can create a new snapshot.
   

## What areas could experience regressions
Creation and viewing of snapshots in an RKE2 provisioned.

## Are the repro steps accurate/minimal?
Yes

samjustus:
SURE-4792

markusewalker:
[...]

Issue #37630, opened 2022-05-06
Snapshots tab is missing for cluster owner that is a standard user


timhaneunsoo:

Jono-SUSE-Rancher:
@samjustus based on the feedback from Nancy, this is almost certainly going to be a Neo issue. Can you prioritize for 2.6.6? We should release note for 2.6.5.

sowmyav27:
**On 2.6head commit id: `628b1b9`**

- Deployed an RKE2 cluster
- Add user1 as standard user.
- Login as user1
- snapshots tab is missing
- As user1 take a manual snapshot
- Snapshot tab and snapshots are available
- Restore option is also available on the snapshot

cbron:
@sowmyav27 @timhaneunsoo aren't your scenario's different ? Original case is for cluster owner / creator, and not a cluster member.

samjustus:
SURE-3770

KevinJoiner:
## Root Cause
After a snapshot is created, it is stored in the fleet-default namespace in the local cluster, and then the specific resource gets a roleBinding to desired roles. When this binding is complete, `/schemas` can correctly inform the UI that the user has permission to manage snapshots. If there are no snapshots taken, there is no binding, and `/schema` is not populated with `etcdsnapshot.rke.cattle.io.` Thus, the users appear not to have permissions to manage snapshots.

## What was fixed, or what change has occurred
We create a roleTemplates with permissions to `etcdsnapshot.cattle.io` in the cluster namespace to fix this issue. Snapshots are not stored in the cluster namespace,  But this allows the UI to see if a user has permissions for the CRD regardless of if any resources are present.

## Areas or cases that should be tested
Test that non-admin users with the cluster-owner role can view and create snapshots for their assigned cluster even if there are no previously created clusters

1. Verify cluster-owner can view and create snapshots
   - Provision a new cluster using RKE2 
   - Create a new User (testUser) 
   - Assign testUser to the role cluster-owner in the newly provisioned cluster.
   - Login as testUser.
   - Verify the snapshots tab is present.
   - Verify the user can view an empty list of snapshots
   - Verify the user can create a new snapshot.

2. Verify users can only view snapshots for their cluster
   - Provision 2 downstream cluster using RKE2 (clusterA and clusterB).
   - Create a new User (testUser) 
   - Assign the cluster-owner in clusterA.
   - As the admin create a snapshot in clusterB.
   - Login as testUser.
[...]

PR #37753, opened 2022-05-17
Fixed snapshot RBAC for RKE2 clusters

KevinJoiner:
Snapshots get stored in the local cluster under the `fleet-default`
namespace. Because of this, roles get access to individual
snapshots. If no snapshots are stored, the users do
not have access to view or create snapshots in the UI.
Users now have access to the `etcdsnapshots.rke.cattle.io`
in the namespace of the cluster. Making the CRD visible in
the schema when no snapshots are present.
#37630

The "Manage Cluster Backup" role also did not have the
`etcdsnapshots.rke.cattle.io` rule in its template.
#37752

### DIFF
--- a/pkg/controllers/management/auth/crtb_handler.go
+++ b/pkg/controllers/management/auth/crtb_handler.go
@@ -52,7 +52,8 @@ var clusterManagmentPlaneResources = map[string]string{
 	"nodepools":                   "management.cattle.io",
 	"notifiers":                   "management.cattle.io",
 	"podsecuritypolicytemplateprojectbindings": "management.cattle.io",
-	"projects": "management.cattle.io",
+	"projects":      "management.cattle.io",
+	"etcdsnapshots": "rke.cattle.io",
 }
 
 type crtbLifecycle struct {
--- a/pkg/data/management/role_data.go
+++ b/pkg/data/management/role_data.go
@@ -229,7 +229,8 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext) 
 		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch")
 
 	rb.addRoleTemplate("Manage Cluster Backups", "backups-manage", "cluster", false, false, false).
-		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("*")
+		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("*").
+		addRule().apiGroups("rke.cattle.io").resources("etcdsnapshots").verbs("get", "list", "watch")
 
 	rb.addRoleTemplate("Manage Navlinks", "navlinks-manage", "cluster", false, false, false).
 		addRule().apiGroups("ui.cattle.io").resources("navlinks").verbs("*")


pkg/data/management/role_data.go
```diff
@@ -229,7 +229,8 @@ func addRoles(wrangler *wrangler.Context, management *config.ManagementContext)
 		addRule().apiGroups("catalog.cattle.io").resources("clusterrepos").verbs("get", "list", "watch")
 
 	rb.addRoleTemplate("Manage Cluster Backups", "backups-manage", "cluster", false, false, false).
-		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("*")
+		addRule().apiGroups("management.cattle.io").resources("etcdbackups").verbs("*").
+		addRule().apiGroups("rke.cattle.io").resources("etcdsnapshots").verbs("get", "list", "watch")
```
cbron, line 233: I don't understand why this is get/list and not `*`, but it's also that way on Cluster Owner ? Maybe @Oats87 can explain.

pkg/controllers/management/auth/crtb_handler.go
```diff
@@ -52,7 +52,8 @@ var clusterManagmentPlaneResources = map[string]string{
 	"nodepools":                   "management.cattle.io",
 	"notifiers":                   "management.cattle.io",
 	"podsecuritypolicytemplateprojectbindings": "management.cattle.io",
-	"projects": "management.cattle.io",
+	"projects":      "management.cattle.io",
+	"etcdsnapshots": "rke.cattle.io",
 }
 
 type crtbLifecycle struct {
```
cbron, line 59: > Because of this, roles get access to individual snapshots. If no snapshots are stored, the users do not have access to view or create snapshots in the UI.

Can you clarify what you mean by "roles get access to snapshots" there ? 

The idea that you grant them this resource in the cluster-specific namespace makes sense. But earlier you said backups are actually available in the fleet-default NS, so does the backup never live in the cluster namespace ? And therefore this is sort of a "fake it" way to get the schema ?

cbron:
Adding someone from provisioning team to review, as this deals more with that system.

pkg/controllers/management/auth/crtb_handler.go
KevinJoiner: You are correct. After a snapshot is created, it is stored in the fleet-default namespace in the local cluster, then the specific resource is added to the role that is bound to cluster-owners. Backups will not live in the cluster namespace, but it allows the schema to know they have access to the CRD `etcdsnapshots.rke.cattle.io`. This lets the UI know the given user has access to create new snapshots when none are available.

In the future, a better solution would be to add the snapshots to the cluster namespace instead of fleet-default, but the current behavior is needed for provisioning with fleet.

rmweir (approved):
LGTM

cbron (commented):
Lgtm, but I'd like someone from provisioning team to chime in.

pkg/data/management/role_data.go
Oats87: We do not necessarily want users being able to delete, update, or patch snapshots manually. These are all supposed to be managed by the controller only, and editing them can lead to unexpected results.

pkg/controllers/management/auth/crtb_handler.go
Oats87: These objects should always live in the namespace that correlates to the provisioning cluster. We have an issue filed to handle this here: https://github.com/rancher/dashboard/issues/5295

Oats87 (approved):

jakefhyde (approved):

cbron (approved):

PR #37772, opened 2022-05-18
update options for optional registry. Also fix bootstrap password.

slickwarren:
This PR allows the HA Airgap job to work with:
* an external/existing registry. Must still be in same subnet where airgap will deploy
  * supports noauth and auth setups. for auth setups, must specify `RANCHER_BASTION_USERNAME`
* prints bootstrap password to jenkins console

Passing jenkins freeform jobs on request. 
Also have a PR for jenkins job builder

RANCHER_HOST_NAME=<6char_name>
RANCHER_BASTION_REGISTRY=<fqdn/url_of_registry>
RKE_VERSION=<v_rke_version_for_rancher>
RANCHER_SERVER_VERSION=<v_rancher_version>
RANCHER_EXTERNAL_BASTION=<aws_bastion_id>
RANCHER_HA_CERT_OPTION=byo-valid

### DIFF
--- a/tests/validation/tests/v3_api/test_airgap_ha.py
+++ b/tests/validation/tests/v3_api/test_airgap_ha.py
@@ -21,7 +21,7 @@ from lib.aws import AWS_USER
 
 AWS_AMI = os.environ.get("AWS_AMI", "ami-012fd49f6b0c404c7")
 DOCKER_COMPOSE_VERSION = os.environ.get("DOCKER_COMPOSE_VERSION", "1.24.1")
-RKE_VERSION = os.environ.get("RKE_VERSION", "v1.3.0")
+RKE_VERSION = os.environ.get("RKE_VERSION", "v1.3.11")
 
 # airgap variables
 NUMBER_OF_INSTANCES = int(os.environ.get("RANCHER_AIRGAP_INSTANCE_COUNT", "3"))
@@ -29,8 +29,6 @@ HOST_NAME = os.environ.get('RANCHER_HOST_NAME', "testsa")
 AG_HOST_NAME = HOST_NAME
 RANCHER_AG_INTERNAL_HOSTNAME = AG_HOST_NAME + "-internal.qa.rancher.space"
 RANCHER_AG_HOSTNAME = AG_HOST_NAME + ".qa.rancher.space"
-RANCHER_HELM_EXTRA_SETTINGS = os.environ.get("RANCHER_HELM_EXTRA_SETTINGS")
-
 
 RESOURCE_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                             'resource')
@@ -44,9 +42,9 @@ PRIVATE_REGISTRY_PASSWORD = \
 RANCHER_EXTERNAL_BASTION = os.environ.get("RANCHER_EXTERNAL_BASTION", "")
 RANCHER_EXTERNAL_HOST_NAME = os.environ.get(
     "RANCHER_EXTERNAL_HOST_NAME", AG_HOST_NAME)
-REG_HOST_NAME = RANCHER_EXTERNAL_HOST_NAME + "-registry"
-REGISTRY_HOSTNAME = os.environ.get(
-    "REGISTRY_HOSTNAME", REG_HOST_NAME + ".qa.rancher.space")
+REG_HOST_NAME = RANCHER_EXTERNAL_HOST_NAME + "-registry" + ".qa.rancher.space"
+REGISTRY_HOSTNAME = os.environ.get("RANCHER_BASTION_REGISTRY", REG_HOST_NAME)
+INSTALL_DEPENDENCIES_FLAG=os.environ.get("INSTALL_DEPENDENCIES_FLAG", "True")
 
 
 def test_deploy_airgap_ha_rancher(check_hostname_length):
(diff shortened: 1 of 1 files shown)


tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -242,7 +270,22 @@ def prepare_airgap_node(bastion_node, number_of_nodes):
                 bastion_node.ssh_key_name, AWS_USER,
                 ag_node.private_ip_address, AWS_USER)
         bastion_node.execute_command(ag_node_update_docker)
-
+        if PRIVATE_REGISTRY_USERNAME is not None:
+            # assuming that we setup the auth-enabled registry from the other
+            # airgap job, which uses self signed certs on the registry. We 
+            # now need to login and move the certs to each downstream node. 
+            ss_certs_command = \
```
izaac, line 277: What does ss mean?

tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -303,9 +352,6 @@ def setup_airgap_rancher(bastion_node, number_of_nodes=NUMBER_OF_INSTANCES):
     new_rancher_version = RANCHER_SERVER_VERSION
     if RANCHER_SERVER_VERSION[0] == 'v':
         new_rancher_version = RANCHER_SERVER_VERSION[1:]
-    extra_settings = []
-    if RANCHER_HELM_EXTRA_SETTINGS:
```
izaac, line 307: Was this extra settings used before? If so why are you removing it now?

tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -159,19 +179,27 @@ def get_registry_resources(external_node):
...
-        'sudo mv rke /usr/local/bin && ' \
-        'sudo snap install helm --classic && ' \
-        'sudo snap install kubectl --classic'.format(RKE_VERSION)
+        'sudo mv rke /usr/local/bin'.format(RKE_VERSION)
     return external_node.execute_command(docker_compose_command)
 
 
+def install_helm(external_node):
+    return external_node.execute_command(
+        'sudo snap install helm --classic')
```
izaac, line 195: snaps are ubuntu only iirc, are we planning to support other OS's in the future? If that's the case I think downloading the binaries would be preferable. 

Otherwise if the plan is to support Ubuntu only this is ok.

tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -97,17 +111,23 @@ def test_deploy_airgap_ha_rancher(check_hostname_length):
...
     auth_url = base_url + "/v3-public/localproviders/local?action=login"
     wait_for_status_code(url=auth_url, expected_code=200)
-    set_url_and_password(base_url, "https://" + RANCHER_AG_INTERNAL_HOSTNAME)
+    get_bootstrap_passwd = "export KUBECONFIG=~/kube_config_config.yaml && " \
+    "/snap/bin/kubectl get secret --namespace cattle-system bootstrap-secret -o " \
+    """go-template='{{.data.bootstrapPassword|base64decode}}{{"\\n"}}'"""
+    bootstrap_passwd = bastion_node.execute_command(get_bootstrap_passwd)[0]
+    print("bootstrap password:", bootstrap_passwd.replace(" ", ""))
+    # currently can't set password in airgap setup without reworking below function
+    # set_url_and_password(base_url, "https://" + RANCHER_AG_INTERNAL_HOSTNAME)
```
izaac, line 130: What's required to make this function work? not a big deal if we're planning to continue manually enter the password but I'm just curious.

izaac (commented):
Just a few comments and questions that I have. Not too much of a deal. Looks good.

tests/validation/tests/v3_api/test_airgap_ha.py
sowmyav27: I think Airgap for now is on ubuntu only if I am not mistaken. HA deploy (non airgap) takes in AMI and user.

tests/validation/tests/v3_api/test_airgap_ha.py
slickwarren: self signed

tests/validation/tests/v3_api/test_airgap_ha.py
slickwarren: there was an attempt at using it, but I tested it and it was not working properly. I tried to debug why, but decided to take it out until we actually need it.

tests/validation/tests/v3_api/test_airgap_ha.py
slickwarren: yes for now we have it for ubuntu only, but that is a good point for future requests.

tests/validation/tests/v3_api/test_airgap_ha.py
slickwarren: I'm actually not sure why it wasn't working in my initial testing. Here were my 2 attempts:
* use the extra_args setting (from earlier thread) to pass in `bootstrap_password=admin` like we do for the regular HA install. However, for whatever reason, since this method uses the `template` approach to install rancher is my hunch, this argument was not allowed and rancher install would fail everytime I tried it. So then I tried
* using the generated bootstrap password, passing it to the `set_url_and_password` function as a new argument, logging in with the bootstrap password then setting it to a proper password. However even this approach did not work and I could not figure out why. I made sure that (in testing that new function argument) the `bootstrap` password worked manually, even though the function would fail in testing with the same setting. 
 
It might be worth more investigation in the future.

tests/validation/tests/v3_api/test_airgap_ha.py
izaac: Oh ok, i see now

tests/validation/tests/v3_api/test_airgap_ha.py
irishgordo: Ahhh that's a good find @slickwarren!
Do you think this might also be somewhat related to:
https://github.com/rancher/rancher/issues/37779 ?

sowmyav27 (approved):

PR #37777, opened 2022-05-19
[WIP] Optimize RTB lifecycle updates

cmurphy:
When there are a large volume of projects and therefore a large volume
of role template bindings, running reflect.DeepEquals on controller
updates, even though there are only on the order of 5-10 rules in each
role template, starts to use a high proportion of memory. This change
introduces a different way of comparing the rules that shifts the
compute and memory burden toward the role template. The role template
controller now computes a checksum representing the rules and sets it as
an annotation on the resource. The RTB update then checks the checksum
to see whether it has changed since the last update, instead of checking
the rules directly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,9 +9,10 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20200712062324-13d1f37d2d77
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.2
+
+	github.com/rancher/norman => github.com/cmurphy/norman v0.0.0-20220607034221-57ba232d48d4
 	github.com/rancher/rancher/pkg/apis => ./pkg/apis
 	github.com/rancher/rancher/pkg/client => ./pkg/client
-
 	github.com/rancher/steve => github.com/cmurphy/steve v0.0.0-20220520222108-78aeca15ab90
 	helm.sh/helm/v3 => github.com/rancher/helm/v3 v3.8.0-rancher1
 	k8s.io/api => k8s.io/api v0.23.3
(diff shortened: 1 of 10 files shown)


cmurphy:
Still testing this, just posting it to give a general picture of what my idea is

MbolotSuse:
One note here: I noticed that you are using sha256 for your hash algorithm. I think that md5 is a bit faster than sha, and since we aren't using this for actual crypto we don't have to worry about md5's collision issues.

I doubt that it would make a substantial difference and wouldn't hold up a PR for that - just something I thought I should bring up.

cmurphy:
@MbolotSuse that's a really good point, the main reason I went with sha256 is that using md5 or other hash functions can trigger some code scanners to flag it as insecure. You're right that this isn't being used for actual crypto and md5 would be fine.

go.mod
```diff
@@ -9,9 +9,10 @@ replace (
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20200712062324-13d1f37d2d77
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.3
-
 	github.com/rancher/rancher/pkg/apis => ./pkg/apis
 	github.com/rancher/rancher/pkg/client => ./pkg/client
+
+	github.com/rancher/steve => github.com/cmurphy/steve v0.0.0-20220520222108-78aeca15ab90
```
cmurphy, line 15: https://github.com/rancher/steve/commit/78aeca15ab905ce089a15ea9a21ac3c1a94d35c4

cmurphy:
The latest commit addresses the issue with `accessSet.Add` showing high memory usage in pprof. With this, the system memory usage (just judging by htop) while generating projects (with https://github.com/cmurphy/scalingscripts/blob/main/genprojects.sh) remains basically flat, as long as the browser isn't open. Not a final solution, but shows where the problem is and proves that it can be solved.

go.mod
cmurphy: https://github.com/cmurphy/steve/tree/optimize-access-for

go.mod
```diff
@@ -10,8 +10,10 @@ replace (
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20200712062324-13d1f37d2d77
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.2
 
+	github.com/rancher/norman => github.com/cmurphy/norman v0.0.0-20220607034221-57ba232d48d4
```
cmurphy, line 13: https://github.com/cmurphy/norman/tree/fail-finalizers

go.mod
```diff
@@ -10,8 +10,10 @@ replace (
...
 	github.com/rancher/rancher/pkg/apis => ./pkg/apis
 	github.com/rancher/rancher/pkg/client => ./pkg/client
+	github.com/rancher/steve => github.com/cmurphy/steve v0.0.0-20220609191147-6ffbae436d0d
```
cmurphy, line 16: https://github.com/cmurphy/steve/tree/optimize-access-for

Issue #37597, opened 2022-05-05
Some RKE2 downstream cluster options are reset to UI defaults on first Edit


snasovich:
**Rancher Server Setup**
- Rancher version: 2.6-head (9de50354f56f69d56f84d324db62db624832486f)
- Installation option (Docker install/Helm Chart): Helm Chart (though this is likely irrelevant)
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): ?
- Proxy/Cert Details: None

**Information about the Cluster**
- Kubernetes version: v1.20.8
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE2 provisioned

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom): Admin
   
**Describe the bug**
Certain RKE2 provisioned downstream cluster options (e.g. Drain nodes on upgrade settings) that were modified during initial cluster creation are reset to their default UI values on the first cluster Edit.

**To Reproduce**
1. Provision an RKE2 node driver cluster in any cloud provider setting "Upgrade Strategy -> Control Plane (or Worker Nodes) -> Drain Nodes" to "No" or leaving it as "Yes" and unchecking "Delete pods using emptyDir volumes" (so they are different than UI default)
	- Note that UI defaults may be changed in the future which may affect bug reproduction using this exact scenario; however it should still be reproducible in general, see more info in "Additional context".
<img width="1525" alt="image" src="https://user-images.githubusercontent.com/85187633/166853604-b32c4936-d0dc-4d31-abfe-5c08f1d7baa5.png">
2. "Edit Config" for this cluster (don't have to wait for it to finish provisioning) and switch to the "Upgrade Strategy" tab.
3. Observe the values of the mentioned fields switch back to defaults:
<img width="1525" alt="image" src="https://user-images.githubusercontent.com/85187633/166853699-59d97a85-a285-4981-b148-79671e9798c9.png">
4. Note that the change happens "behind the scenes", so the user doesn't have to click "Upgrade Strategy" tab for the issue to reproduce. Simply hitting "Save" right away on "Edit Config" screen would save the object with these settings reverted to UI defaults.
5. Another notable item is that if the referenced configuration changes are done on the "Edit Config" screen, the values actually "stick" and subsequent "Edit Config"s perform as expected.

[...]

PR #37785, opened 2022-05-20
Remove `omitempty` tags to ensure fields persist across json marshals

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/37597
 
# Problem
 UI defaults were incorrectly being applied to fields which were equal to a zero value (such as `false`). This resulted in the loss of some initial configurations of the Upgrade Strategy after a subsequent edit of the Configuration. The UI was applying its defaults to fields which it could not find in the `GET` request made to the backend upon attempting to edit the Configuration. 

# Solution
These fields were being dropped by the web-hook upon marshaling due to the `omitempty` tag on the struct `DrainOptions`. This caused those fields to be missing in the subsequent `GET` request made by the UI. This can be avoided by removing the `omitempty` tags.
 
# Testing
 + Begin creating an RKE2 cluster on any infrastructure-provider
 + Under `Cluster Configuration` find the `Upgrade Strategy` tab.
 + Select `yes` for `Drain Nodes` for the `Control Plane` 
 + deselect the `Delete pods using emptyDir volumes`
 + Create the cluster
 + On the Cluster Management screen, select the cluster and click on `edit config`
 + navigate to the `Upgrade Strategy` field and observe that the `Delete pods using emptyDir volumes` box has been not been re-activated, even though it is equal to that fields zero value.


jakefhyde (commented):
Does this cause any functional changes to the backend, or is this solely for UI/terraform?

jakefhyde (approved):

HarrisonWAffel:
@snasovich I could not reproduce this this issue with the other fields under the `Cluster Configuration` section of the UI. It's a bit tricky, since this issue relies heavily on the structure of the JSON `POST`'d to the backend by the UI. Having some fields omitted is actually OK, a lot of the options are more geared towards enabling functionality as opposed to disabling functionality (which drops the field). 

For instance, etcd Automatic snapshots are disabled by default, so dropping that field is acceptable. The same goes for the ACE Option `Authorized Endpoint`. In the instance where you _are_ disabling a field that is enabled by default, such as when disabling a `System Service` (nginx, coreDNS, metrics-server) these items are added to a list and explicitly disabled, as opposed to just setting the field to a zero value.
```
"machineGlobalConfig": {
        ....
        "disable": [
          "rke2-coredns",
          "rke2-ingress-nginx",
          "rke2-metrics-server"
        ],
       .....
},
```

The `DrainOptions` struct used to define the `Upgrade Strategy` tab of the UI is the only struct, from what I can tell, that both has fields enabled by default, and does not explicitly denote them as disabled but instead sets them to the fields zero value. This is why I felt it was appropriate to only drop the `omitempty` tags from this struct, as opposed to removing them from all structs used in the `GET` request for the UI page.

jakefhyde:
@snasovich Seems like actually fixing the root of the problem may require something more involved from a norman perspective that would probably cause some breakage in other places. I think we'll just need to be vigilant for now to make sure we don't run into the same situation again. @HarrisonWAffel does that sound correct?

snasovich:
After chatting with @HarrisonWAffel I'm OK with proceeding with fixing it just for these fields - going with the safer fix.

snasovich (approved):

Issue #37644, opened 2022-05-09
Logging v1 settings n longer visible after update to 2.5.1/2.6.4


SheilaghM:
**Rancher Server Setup**
Rancher Cluster:
Rancher version: 2.5.12 & 2.6.4

k8s: v1.20.14s:

**Information about the Cluster**
Downstream Cluster:
k8s: v1.20.14
Logging v1 v0.2.5
Logging v2 v3.9.400

**Describe the bug**
Issue description:
The Logging v1 settings are not visible within the Rancher v2.6.4 interface, also seen in v2.5.12. The UI is showing the logging for the project as disabled even if the Projectlogging object is available in the cluster. Existing configs are no longer shown in the menu. So if you make your settings save, leave the menu and come back it is empty. Logging v1 Menu is not accessible. Existing configs are no longer shown in the menu. 

**To Reproduce**
**Troubleshooting steps:**
UI is showing the logging for the project as disabled even if the Projectlogging object is available in the cluster
Asked to enabled the legacy feature as well after the upgrade to 2.6.4
Tried on a second donwstream - the issue with logging v1 is there too.

**Repro Steps:**
Make settings->Save->Leave the menu->Come back->Empty

Tested that on two independent rancher instances 
Tested on 2.5.12 & 2.6.4 -> problem is there
Rolling back to 2.5.11 fixes the issue

**Result**
Logging v1 settings are no longer visible for the users via UI. The settings are still visible via API, but that's not really usable for our users

**Expected Result**
Be able to work and save existing configs. Logging v1 settings to be visible in the Rancher 2.6.4 UI.

**Screenshots**
N/A

SURE-4477

niusmallnan:
It should be an issue of Norman API.

To fix this, I think we should do the same thing on the v2.6 branch:
https://github.com/rancher/rancher/issues/36267#issuecomment-1086513498

steffeneichler:
same problem here. 2.5.11 works. 2.5.12 & 2.6.4 buggy,

ronhorton:
The way I'm able to access the logging settings is via the 'install monitoring' href at the top of the cluster page. When I click 'manage' the settings are present. Is that the navigation that was intended?

PR #37181, opened 2022-04-05
Alerts (v1) do not display alert category or description

geethub97:
https://github.com/rancher/rancher/issues/36267

PR #37868, opened 2022-05-31
Excluded GVK of Logging v1 schemes for projectlogging and clusterlogging

eliyamlevy:
Issue: https://github.com/rancher/rancher/issues/37644

### DIFF
--- a/pkg/api/scheme/scheme.go
+++ b/pkg/api/scheme/scheme.go
@@ -106,6 +106,8 @@ func init() {
 	Scheme.ExcludeGVK(management.SchemeGroupVersion.WithKind("NodeTemplateList"))
 	Scheme.ExcludeGVK(management.SchemeGroupVersion.WithKind("ClusterMonitorGraphList"))
 	Scheme.ExcludeGVK(management.SchemeGroupVersion.WithKind("ProjectMonitorGraphList"))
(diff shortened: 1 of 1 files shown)


superseb (changes requested):
Is this a forwardport of https://github.com/rancher/rancher/pull/37181? If so, why are only 2 of the 6 added in this PR?

eliyamlevy:
It's not a forward port but it makes sense to include those as well.

superseb (changes requested):
Now there are 12 ExcludeGVK here and 10 in `release/v2.5`, is `release/v2.5` incomplete?

eliyamlevy:
The two I added are only causing problem in 2.6.x and 2.5.12 as far as the issue goes. Since the chart is deprecated I think it's ok to fix it for the main use case of the customer asking for the fix which is to have it work on the next 2.6 release.

eliyamlevy:
> **Expected Result**
> Be able to work and save existing configs. Logging v1 settings to be visible in the Rancher 2.6.4 UI.

eliyamlevy:
Also for some reason its not showing up on the issue so I'm linking the issue again.

Issue for this PR: [#37644](https://github.com/rancher/rancher/issues/37644)

superseb (approved):

PR #37870, opened 2022-05-31
add debug mode for rancher machine provisioning with rke2

tuxtof:
in rke2 provisioning  mode there is no way to enable debug mode for machine driver.

This PR fix the the problem by checking if global rancher log level is trace and if this is the case set --debug when launching machine provisioning pod


tuxtof:
please merge

PR #37812, opened 2022-05-24
Add support to specify AMI IDs for EC2 node driver

markusewalker:
**PR Description**

This PR adds support for EC2 images to be specified when testing with the Go automation framework. If you choose to not set this optional parameter, then the prior behavior of the default AMI per AWS region still maintains; this simply offers more configuration choice.

Tested with multiple quick start / community AMIs for the following distros:

- Ubuntu 20.04: PASS
- SUSE 15SP3: PASS
- Ubuntu 22.04: PASS
     - Not supported, but the automation should still allow to pass, which it did.

PR #37884, opened 2022-06-01
Add AMI support for EC2 node drivers

markusewalker:
**PR Description**

This PR adds support for EC2 images to be specified when testing with the Go automation framework. If you choose to not set this optional parameter, then the prior behavior of the default AMI per AWS region still maintains; this simply offers more configuration choice.

Tested with multiple quick start / community AMIs for the following distros:

- Ubuntu 20.04: PASS
- SUSE 15SP3: PASS
- Ubuntu 22.04: PASS
     - Not supported, but the automation should still allow to pass, which it did.


sowmyav27:
Merging since prev [PR](https://github.com/rancher/rancher/pull/37812) had approvals.

PR #37893, opened 2022-06-02
add initial codeowners file

rosskirkpat:
As part of addressing some recent issues surrounding changes made that affected CI and were not caught prior to being merged, I am adding a minimal guardrail using a CODEOWNERS file.

PRs will require review and approval by ~~@oxr463, @paraglade,~~ @superseb, @phillipsj , and myself for any changes made to the Drone yml file.

PRs will require review and approval by @phillipsj and myself for any changes to Windows related files.

PRs will require review and approval by @Oats87 and myself for any changes made to the shell scripts for provisioning tests.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,6 @@
+.drone.yml @oxr463 @paraglade @rosskirkpat
+*_windows.go @phillipsj @rosskirkpat
+/package/windows/  @phillipsj @rosskirkpat
+/scripts/windows/  @phillipsj @rosskirkpat
+Dockerfile-windows.dapper  @phillipsj @rosskirkpat
+/scripts/provisioning-test* @Oats87 @rosskirkpat
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
-.drone.yml @oxr463 @paraglade @rosskirkpat
+.drone.yml @phillipsj @rosskirkpat @superseb
 *_windows.go @phillipsj @rosskirkpat
 /package/windows/  @phillipsj @rosskirkpat
 /scripts/windows/  @phillipsj @rosskirkpat


rosskirkpat:
This should be merged after https://github.com/rancher/rancher/pull/37892

phillipsj (dismissed):

superseb:
Minimal guard rails is different from introducing SPOF, adding single persons will create a blocking problem as soon as that person is not working and in general, keep relying on individuals will not make anything more flexible.

Drone build configuration is something that Engineering owns, that EIO is running the infra does not mean that they need to be introduced to review our configuration. And again, a single individual next to EIO that will review is not flexible.

oxr463:
> Minimal guard rails is different from introducing SPOF, adding single persons will create a blocking problem as soon as that person is not working and in general, keep relying on individuals will not make anything more flexible.

I think the goal here is to get more review on changes that impact the pipeline.

> Drone build configuration is something that Engineering owns, that EIO is running the infra does not mean that they need to be introduced to review our configuration. And again, a single individual next to EIO that will review is not flexible.

Agreed. I don't think EIO is the proper owner of this pipeline. However, I do think it would be a good idea for one of the teams to own this.

CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+.drone.yml @oxr463 @paraglade @rosskirkpat
```
oxr463, line 1: I don't think this should be EIO.

oxr463 (changes requested):

CODEOWNERS
rosskirkpat: I agree that it should not be wholly owned by EIO, however, at this time I don't think any Engineering team (or sub-set of a team, like what EIO is), has a clear ownership/knowledge of Drone other than EIO. That was why I included both of you. This is something we should address at a higher level of SMEs and technical ownership of a product/project.

I know that @phillipsj and I have spent a lot of time working in Drone across many projects in Rancher. I will change this to include the two of us and @superseb

PR #37919, opened 2022-06-06
Updated the provisioning client to hit the steve API endpoint.

igomez06:
**Background** 
When using the provisioning client we were not getting the Steve API endpoint but rather the Kubernetes API. So when attempting to do some calls it would fail differently in comparison to the rancher monitoring UI.

**Updates**
- Over wrote most of the API calls to hit the steve API endpoint
- For now just not including watch was we need to monitor the cluster resource using the Kubernetes API
- We get an unknown error during clean up for a standard user, which is expected because the user no longer has the proper RBAC to view the resource it had just deleted
- We also get an expected error when cleaning up the rke-machine-config.cattle.io resource, because now when we delete the cluster using the steve API endpoint, the rke-machine-config.cattle.io resource gets deleted as well.

### DIFF
--- a/tests/framework/clients/rancher/provisioning/cluster.go
+++ b/tests/framework/clients/rancher/provisioning/cluster.go
@@ -2,22 +2,147 @@ package provisioning
 
 import (
 	"context"
+	"time"
 
-	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	scheme "github.com/rancher/rancher/pkg/generated/clientset/versioned/scheme"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
-// Create is ClusterInterface's Create function, that is being overwritten to register its delete function to the session.Session
-// that is being reference.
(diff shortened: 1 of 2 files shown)


izaac:
Is there a test set of tests that make use of these actions, that we could see in action in our freeform?

I understand the CRUD functions but I'm off context of "failing differently than the rancher monitoring UI" so we detected these errors during the design of the monitoring tests?

izaac (commented):
LGTM 🚀 

I reviewed the execution in our CI too.

izaac (approved):

caliskanugur (approved):

tests/framework/clients/rancher/provisioning/cluster.go
```diff
@@ -2,22 +2,147 @@ package provisioning
...
+	scheme "github.com/rancher/rancher/pkg/generated/clientset/versioned/scheme"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
-// Create is ClusterInterface's Create function, that is being overwritten to register its delete function to the session.Session
-// that is being reference.
-func (c *Cluster) Create(ctx context.Context, cluster *apisV1.Cluster, opts metav1.CreateOptions) (*apisV1.Cluster, error) {
+const (
```
anupama2501, line 14: why do we need this constant here?

igomez06: Because this is the Steve api endpoint we need to hit

anupama2501 (approved):
lgtm

Issue #35135, opened 2021-10-15
[2.6.1] Adding repository fails when specifying valid SSH URL


janeczku:
Rancher 2.6.1

**Steps to reproduce:**

Given a Git Server that requires the account name `gitrepos` (in contrast to the more commonly used username `git`) to access Git repositories over SSH:

1. Store Helm chart(s) in a private Git repository
2. Navigate to Apps&Markteplace -> Repositories -> Create
3. Enter the valid SSH URL in the Git Repo URL field: `gitrepos@<GIT-SERVER:<ORG-NAME>/<REPO-NAME>.git`
4. Select branch
5. Enter the SSH private key
6. Click Save

**Expected Result:**

Repo is added and charts are downloaded.

**Actual Result** 

Error is thrown:

> Failed to parse URL ... first path segment cannot contain colon

<img width="1214" alt="Screenshot 2021-10-15 at 11 09 50" src="https://user-images.githubusercontent.com/3813921/137471842-49fff701-fc3a-402f-8324-28e39ada5750.png">

ronhorton:
**Pass** verified in `2.6-head Commit ID 4199cac`

1. generated ssh key
2. added ssh key to git repo
3. app > repositories > create
4. added git repo
5. added public and private keys
6. clicked 'create' and repo became active after ~10 seconds. No errors

![Screen Shot 2022-06-08 at 5.07.40 PM.png](https://zube.io/files/rancher/7371d7a83e98fbf6d4d582cd99328a9d-screen-shot-2022-06-08-at-5-07-40-pm.png)

PR #37923, opened 2022-06-06
Fix git repo SSH URLs with custom usernames

PennyScissors:
Issue: SURE-3512, https://github.com/rancher/rancher/issues/35135

Our existing logic for git repo SSH URL validation only allows URLs with the prefix `git@`, but users may have a username different than `git`. This PR modifies that logic to allow URLs with other usernames such as `gitrepos@`

### DIFF
--- a/pkg/catalogv2/git/download.go
+++ b/pkg/catalogv2/git/download.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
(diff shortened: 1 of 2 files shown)


superseb (commented):
The only concern I can find with this change is that URLs without `.git` were allowed before (and can work), but are now invalid which could cause working repos to not work anymore.

aiyengar2 (approved):

PR #37931, opened 2022-06-07
Add tests for rancher2 TF provider through terratest

Josh-Diamond:
Framework for automating testing on terraform modules using terratest and Go.

To use:
- Create and export configuration specs in config.go, to later reference in tests
- Create a new _test.go file in the `tests` folder and begin writing a test


izaac:
Please squash the commits and one question:

Are those go.mod changes necessary? Could we use the existing versions instead of bumping them?

tests/terratest/functions/getClusterID.go
```diff
@@ -0,0 +1,53 @@
+package functions
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+func GetClusterID(hostURL string, clusterName string, token string) string {
```
igomez06, line 9: This goes for all the functions, you have implemented in the functions package. Why is there a need to do direct rest calls to these resources, rather than using the generated clients? Are you using a token to access these resources in a rancher instance? I mean if you use the clients there would no need for this package to be honest.

igomez06:
I also have the same concerns with the go.mod files.
1. The modifications done in the main rancher/rancher mod file
2. The mod file with the test/terratest package. 
Could you explain a little more on why you need them, and also why it has so many dependencies

tests/terratest/functions/getClusterID.go
Josh-Diamond: If the functions are redundant they don't need to be added.  I'm still getting up to speed with the existing automation framework and have not utilized it yet.  I did not realize the existing framework could be used with these tests, if that's what you're suggesting.  This was just my approach and yes, i'm using a token in the GET headers to authenticate the request.  If we choose to leave out the functions, the tests will need to be refactored.

Josh-Diamond:
@igomez06 I'm new to go, and i can't properly answer your question.  If they are not needed, lets remove it 👍

Josh-Diamond:
@izaac @igomez06 
This is my first PR and i appreciate the feedback.  Please let me know the changes i need to make in order to get this PR in shape for merging.

tests/terratest/tests/aks/aks_test.go
```diff
@@ -0,0 +1,92 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/josh-diamond/rancher-terratest/config"
```
igomez06, line 7: Is josh-diamond your directory in your go path?

Josh-Diamond: oof.. yes, yes it is

Josh-Diamond: truth comes out.. i drag and dropped this project from josh-diamond/rancher-terratest repo where i developed it.. sorry if complexities arose from this..

Josh-Diamond: i have a github.com > josh-diamond  rancher
i was developing this code in github.com > josh-diamond > rancher-terratest
the rancher/rancher repo, lives at github.com > rancher > rancher > tests > terratest

Josh-Diamond: whatever changes need to be made, im still not clear.  Just tell me whatever it is that needs to be done and i'll do my best to accomplish that 🙏

igomez06:
@Josh-Diamond I mentioned this on one of your files, you should never use absolute paths when referencing packages.

You are using terra test to run the tests, is that why you have included a go.mod? Usually you don't want go.mod files within a module (rancher/rancher is a module, but it does happen this repo like the generated clients.). If you could explain the work flow of this is being run I would give you a better explanation. A quick check do you run go mod init/download in your test/terratest package? 

My other objection is you are using direct rest calls to access rancher resources, when this can all be handled using the generated clients. Not saying it has to be from the framework, but the rancher/rancher repo has generated clients to access these resources, this would cut down you code a lot. Please see https://github.com/rancher/rancher/tree/release/v2.6/pkg/client/generated

tests/terratest/functions/getClusterID.go
igomez06: Awesome I shared a link in my other comment, or you can use the framework, you would just import the package. I'm leaning on using the framework do these calls. The framework can be imported to this no problem.

tests/terratest/tests/aks/aks_test.go
igomez06: No problem! So you should have cloned the rancher/rancher repo within a directory of this file structure. `path-to/go/src/github.com/rancher/rancher` Your GOPATH env var should set to `path-to/go`. So then when you are referencing these packages it would the correct path. In this case `"github.com/rancher/rancher/tests/terratest"` so on and so forth.

Josh-Diamond:
@igomez06 As far as my understanding, and i'm super new to go, terratest allows you to use Go's testing on terraform modules.  So ultimately, i believe it is go that is running the tests, and utilizing terratest to grab TF module configurations to test against.  Initially when i began making the repo, i did run go mod init and go mod tidy.. I really don't know the relevance or why that go.mod file is needed. 

As far as the clients, i really just didn't know about it.  I will have to learn more before commenting on that

Josh-Diamond:
@igomez06 So at this point, I should update the import paths from josh-diamond to rancher and delete the go.mod file? What about the go.sum? is that needed or should be removed?

igomez06:
> @igomez06 So at this point, I should update the import paths from josh-diamond to rancher and delete the go.mod file? What about the go.sum? is that needed or should be removed?

Yes make sure it's in the structured I showed you before within the rancher/rancher repo, and remove the go.mod and go.sum, but keep them just in case, and see if you can run your tests without them.

Josh-Diamond:
This PR is a mess.. trying to rebase, having troubles.
@izaac could we go over this cleanup process after tomorrows demo?

tests/terratest/functions/getClusterID.go
Josh-Diamond: There are some functions ending with TF, that would still need to be included in the package, so there is still a need for the package, but I agree, all the api calls could/should be removed

igomez06: For sure, just the rest calls not the terraform stuff

PR #37933, opened 2022-06-07
Fix a linter issue blocking the CI

maxsokolovsky:
Fixing a linter issue that was not caught in the pervious PR CI run due to the `build` stage not having run, giving a fake pass.

There were three instances of a linter error like this: `unexported-return: exported func NewADGraphClientFromCredential returns unexported type *clients.azureADGraphClient, which can be annoying to use (revive)`

We may want to think about this rule - whether it warrants an exception to our golang-ci config.


PennyScissors (approved):

MbolotSuse (approved):
LGTM

PR #37953, opened 2022-06-09
adding vpn to rancher automation

slickwarren:
This PR allows a connection to be made using the docker container running tests through jenkins.  Here's a short explanation of how to use a vpn connection:
### New Variables for VPN:
```
EXTERNAL_ENCODED_VPN=<base64_encoded_ovpn_file>
VPN_ENCODED_LOGIN=(optional variable, but all SUSE VPNs require you to login)<base64_encoded_login_file>
```
decoded login file contents should be in the following format:
```
username
password
```

I've included 2 test files, 1 in go and 1 in python showing that the connection is successful. It is not necessary to run these tests when connecting to the VPN, only that you define the variable(s) mentioned above. 

In this PR, I've also included what I think should be all of the jenkinsfile changes to use the VPN wherever we would like, however there may be other jenkinsfiles that need similar modification that we find through testing. 

### additional notes
* Runs in jenkins available upon request for both go and python, with and without the new variables to show that both new and existing functionality will work. 
  * note that runs in jenkins were not including jenkinsfiles, as currently our jenkins pulls from `rancher/rancher` for those files instead of the testing repo. So runs were modified with the content of the jenkinsfiles you see in this PR using the `replay` function in jenkins. 
* There will be an acompanying PR for jenkins job builder to include the new variables where appropriate.


bmdepesa:
I would suggest creating one new Jenkinsfile for each of Go/Python for tests which need VPN access. As we hope to eventually get a dedicated VPN account, we should not rely on having the VPN environment vars passed in as parameters in the job as that is likely to change. 

Do we actually need the tests included? I'm conflicted as to whether or not they would provide any real value beyond showing that we can connect to the VPN.

bmdepesa:
Overall approach is better, would like to see some additional reviews.

sowmyav27 (approved):

Issue #37681, opened 2022-05-11
Importing a AKS cluster with complicated configuration is not editable from Rancher


thedadams:
**Rancher Server Setup**
- Rancher version: v2.6-head
- Installation option (Docker install/Helm Chart): Any
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details: Any

**Information about the Cluster**
- Kubernetes version: Any
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): AKS

**Describe the bug**
If an AKS cluster is created outside of Rancher with some complicated configuration (specifically `--outbound-type userDefinedRouting`), then importing it into Rancher may work but a user would not be able to make any changes to the AKS cluster through Rancher.

**To Reproduce**
1. Create an AKS cluster with `aksctl` passing some non-standard configuration like `--outbound-type userDefinedRouting`.
2. Import the cluster into Rancher
3. Add a node pool to the cluster from Rancher.

**Result**
An error will occur related to changing a field that cannot be changed.

**Expected Result**
The node pool should successfully be added.

**Additional context**
After some discussion on this issue, it seems that since the AKS cluster configuration through the CLI is so complicated, that the best course of action is that `aks-operator` should pull the cluster state from AKS and update the current cluster state with the configuration `aks-operator` and Rancher know about. For example, `outbound-type` is not part of the configuration that Rancher is aware of, so that will not change when the cluster is update from Rancher.

SURE-4373

annablender:
The root cause of this issue is when rancher imports an AKS cluster, it creates a new v3 mgmt cluster object instead of pulling in already existing fields in Azure. Rancher only manages the fields that it knows about so cluster config fields created using `aksctl` cannot be updated by Rancher. That's why creating a new node pool has no effect. The fix will be to add logic [here](https://github.com/rancher/aks-operator/blob/a32c2bf0c3254301d1257345d4db786fef49162e/pkg/aks/create.go#L151) in the `aks-operator` to retrieve the `ManagedCluster` object from Azure if it exists and if not, create a new one.

annablender:
I was not able to reproduce this issue on `rancher-v2.6` head.

Repro steps
[...]

PR #37970, opened 2022-06-10
Update AKS handler logic so imported clusters don't overwrite upstream fields

annablender:
Issue: https://github.com/rancher/rancher/issues/37681  

 
## Problem
 

When importing AKS cluster, the AKS handler in rancher creates an `AKSClusterConfig` object that is the object both rancher and the operator look at when checking for updates. On import, the AKS handler creates a new, mostly empty `AKSClusterConfig` object for the managed cluster without the majority of any preconfigured data that may already be present for the cluster in AKS. Then, any fields set to nil in the `AKSClusterConfig` object will overwrite data in AKS on the next update after provisioning.
 
## Solution
 

* Updated the AKS handler in rancher to copy the `spec.Status.AKSStatus.UpstreamSpec` to `spec.AKSConfig` so fields from the existing AKS cluster will be pulled into Rancher on import.

* Updated comments to be clearer
 
## Testing
 

This issue requires a lot of testing because it touches all parts of provisioning an AKS cluster. Here's the basic test plan I used to smoke test my changes.

Test plan
1. Provisioned (via the Rancher UI kubernetes v1.23.5)
	a. 1 node AKS cluster
	b. 1 node AKS. Update auth IP range to 0.0.0.0/0 in Azure. Check that no config was overwritten in AKS/change was propogated as expected
	c. 1 node AKS cluster. Update HTTP Application Routing in rancher UI. Check that no config was overwritten in AKS
2. Import
	a. 3 node AKS cluster created via CLI with UDR and firewall. Check that no config was overwritten in AKS (agent pool, network). Update kubernetes version for the cluster configuration
	b. 3 node AKS cluster created via Azure portal

### DIFF
--- a/pkg/controllers/management/aks/aks_cluster_handler.go
+++ b/pkg/controllers/management/aks/aks_cluster_handler.go
@@ -35,9 +35,9 @@ import (
 )
 
 const (
-	aksAPIGroup         = "aks.cattle.io"
-	aksV1               = "aks.cattle.io/v1"
-	enqueueTime         = time.Second * 5
+	aksAPIGroup = "aks.cattle.io"
+	aksV1       = "aks.cattle.io/v1"
+	enqueueTime = time.Second * 5
 )
 
 type aksOperatorController struct {
--- a/pkg/controllers/management/aks/aks_cluster_handler.go
+++ b/pkg/controllers/management/aks/aks_cluster_handler.go
@@ -152,8 +152,8 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 			if cluster.Status.AKSStatus.UpstreamSpec == nil {
 				return e.setInitialUpstreamSpec(cluster)
 			}
-			if !reflect.DeepEqual(cluster.Status.AKSStatus.UpstreamSpec, &cluster.Spec.AKSConfig) {
-				logrus.Infof("found imported cluster [%s], copying upstream state to AKSConfig", cluster.Name)
+			if !reflect.DeepEqual(cluster.Status.AKSStatus.UpstreamSpec, cluster.Spec.AKSConfig) {
+				logrus.Infof("Found imported cluster [%s], copying upstream state to AKSConfig", cluster.Name)
 				cluster = cluster.DeepCopy()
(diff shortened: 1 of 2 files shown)


annablender:
Hi all, these changes have been tested but I'd like some initial feedback on the design/location of code fix before I mark it ready for review. Thanks in advance!

pkg/controllers/management/aks/aks_cluster_handler.go
```diff
@@ -148,11 +145,18 @@ func (e *aksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
 	case "active":
+		// If an imported cluster, set the upstream spec and update the AKSClusterConfig object with fields from the
+		// cluster state. This will prevent configuration in Azure from getting overwritten in later updates. Non
+		// imported clusters will already have upstream spec set.
 		if cluster.Spec.AKSConfig.Imported {
 			if cluster.Status.AKSStatus.UpstreamSpec == nil {
-				// non imported clusters will have already had upstream spec set
 				return e.setInitialUpstreamSpec(cluster)
 			}
+			if !reflect.DeepEqual(cluster.Status.AKSStatus.UpstreamSpec, cluster.Spec.AKSConfig) {
```
rmweir, line 155: Is it okay to use DeepEqual here? And do we want to copy upstream spec to AKSConfig? Both of these bypass the idea of leaving fields nil that are unmanaged by rancher. This is vital to syncing the state from upstream. How does this prevent us from overwriting upstream fields? I would think it would do the opposite.

rmweir (changes requested):

pkg/controllers/management/aks/aks_cluster_handler.go
rmweir: The UI should be showing the fields as follows: upstreamspec, then if any fields on spec are non nil, overwrite the field on upstreamspec with that value.

rmweir (changes requested):

annablender:
Closing this PR because `Status` should not be copied to `Spec` for imported clusters. The rancher UI will already display fields from AKSStatus in the UI so no need to backfill again.

Thank you @rmweir for review

Issue #37947, opened 2022-06-08
Unable to login as an admin to the rancher server when AzureAD auth is enabled post graph endpoint migration 


anupama2501:
**Rancher Server Setup**
- Rancher version: `v2.5.14` upgraded to `v2.6-head f4d09c7`
- Installation option (Docker install/Helm Chart): `Docker install`
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): k3s
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: `v1.23.6`
- Cluster Type (Local/Downstream): Downstream Infrastructure provider
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Admin
   
**Describe the bug**
When we enable AzureAD auth on v2.5.13 and upgrade the rancher server to v2.6-head , login as an admin works. But when we upgrade the graph endpoint login fails when we try to login as an admin.

**To Reproduce**
1. On rancher server v2.5.14, enable Azure AD auth from apps & market >> Auth Providers >> Azure AD
2. Upgrade the rancher server from v2.5.14 >> v2.6-head 
3. Upgrade the graph endpoint via kubectl as follows:
```
k edit authconfigs azuread
Update the following endpoints:
graphEndpoint: https://graph.microsoft.com
authEndpoint: https://login.microsoftonline.com/<tenantID>/oauth2/v2.0/authorize
tokenEndpoint: https://login.microsoftonline.com/<tenanantID>/oauth2/v2.0/token
```
4. From Azure portal, add the additional permissions in the test app
5. Logout and try to login as an admin user

**Result**
Login as admin fails with the error:
```
An error occurred logging in. Please try again.
baseType: "error"
code: "ServerError"
message: "Server error while authenticating"
status: 500
type: "error"
```
Some errors from the rancher logs:
```
2022/06/08 22:05:47 [ERROR] [AZURE_PROVIDER] Error getting group: GroupsClient.BaseClient.Get(): unexpected status 404 with OData error: Request_ResourceNotFound: Resource '<redacted>' does not exist or one of its queried reference-property objects are not present.
[...]

PR #37984, opened 2022-06-13
Update dependencies for Azure AD and include MSGraph groups which have securityEnabled set

maxsokolovsky:
Addresses https://github.com/rancher/rancher/issues/37947 found during the QA testing of https://github.com/rancher/rancher/pull/37592

1. The SDK for Microsoft Graph returns a "ghost" group (one that the user is in according to the `list group memberships` call, but which itself cannot be retrieved via a `get` call or lookup in the Azure portal). This only happens for the Azure administrator.
Because our code sees some group ID, it also tries to fetch the group by ID, which fails, and our code fails to log the Azure admin user in.
I found that this special group is a `Global Administrator` group, which, again, does not exist as a standalone group in the Azure portal. Its securityEnabled value is not set all (the way it is represented in the Go code of the SDK). For all other groups, whether security enabled or not, the value is set to a boolean value, not nil. I think this checking for this in code will be a working way to handle this edge case.

2. Updating all the ADAL dependencies to most recent patch versions - irrelevant to the issue, but good overall for the main issue.


MbolotSuse (approved):
LGTM

PR #37987, opened 2022-06-14
Add support to provision RKE1 clusters

markusewalker:
**PR Description**
PR to allow support for provisioning RKE1 clusters in the Go framework. Currently, this file will work with the following node providers:

- Amazon EC2 (includes support for spot instances + tags based on PR https://github.com/rancher/rancher/pull/37934)
- Azure
- Linode

Additionally, as we plan to use this PR to have further design discussions, I have consciously attempted to not do too much refactoring until after we further discuss.

Locally, I have verified that all above node providers work, but I will provide offline Jenkins runs further showing this as well.


tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
```diff
@@ -0,0 +1,185 @@
...
+							},
+						},
+					}
+
+					clusterResp, err := client.Management.Cluster.Create(cluster)
+					require.NoError(r.T(), err)
+
+					var nodeTemplateResp *nodetemplates.NodeTemplate
+
+					switch {
```
igomez06, line 125: We don't want this switch case here. We need to create a new provider file similar to the current one, but with node template

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
```diff
@@ -0,0 +1,185 @@
...
+					assert.Equal(r.T(), clusterName, clusterResp.Name)
+				})
+			}
+		}
+	}
+}
+
+func (r *RKE1NodeDriverProvisioningTestSuite) TestProvisioning() {
+	for _, providerName := range r.providers {
+		provider := CreateProvider(providerName)
```
igomez06, line 176: We need a new CreateProvider specific to RKE1, this is setting up the rke-machine-config.cattle.io which is specific to RKE2

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
markusewalker: Thanks @igomez06. I'll look into creating a new provider file to do just this and call it within this file. Should we just call it `providers_rke1.go` or how do you think would be best to structure this?

@slickwarren mentioned in the other PR of us separating RKE1/RKE2 to avoid long-term confusion. I had some complications when separating folders out, but I do think that would be a good idea if we can make it work. What do you think?

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
markusewalker: Sounds good, I will look into that along with the other review that you gave.

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
igomez06: @slickwarren what do you think? I think having provisioning/rke1 and provisioning/rke2 makes sense for a directory structure, and having the respective providers. Or maybe for now just have a new provider_rke1.go works for me as well.

slickwarren: I second having different folders for different KEs

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
```diff
@@ -0,0 +1,185 @@
...
+	client, err := rancher.NewClient("", testSession)
+	require.NoError(r.T(), err)
+
+	r.client = client
+
+	enabled := true
+	var testuser = AppendRandomString("testuser-")
+	user := &management.User{
+		Username: testuser,
+		Password: "rancherrancher123!",
```
slickwarren, line 58: I'm sure we do this in other tests, but I feel that we shouldn't expose our test passwords at all. Can we hide these, or just leave them for the user to specify (this would be a new ticket).

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
slickwarren: if you need help with how to do this, see `providers.go` which might be helpful

slickwarren (commented):
Israel's comments pretty much sum up what I have to add.

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
markusewalker: I think it would be better to have the user specify. This is present in our other tests as well, so once we figure out best way to do that, I think we could make the needed adjustments across the board.

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
markusewalker: I refactored into `rke1` and `rke2` and updated the `rke1/providers.go` to specifically handle nodetemplates for RKE1 provisioning; verified locally that this is working fine.

Unsure if `config.go` needs to change at all? I just copied into `rke1` because `go` prefers having all *.go files in the same directory when testing...but the initial concern of folder separation and the switch statement is resolved.

markusewalker:
Following up on this PR to see if there are any changes needed for this PR before we proceed with merging.

tests/v2/validation/provisioning/provisioning_rke1_driver_test.go
caliskanugur: @slickwarren what do you think about having a single `config.go` file as mentioned - would config files change depending on cluster types tests in the future? If yes, please ignore the below explanation.

If it's gonna be the same, @markusewalker, for the `config.go` files, you can use the exported things from `provisioning/config.go` instead of duplicating the file. Only the namespace name needs to be public, the rest already should be public.

slickwarren: good point. I'm not sure if it would need to change in the future, but since they are the same now for rke1 and rke2, I think your approach is best. cc @markusewalker do you agree, and if so could you amend this PR again to reflect the changes?

caliskanugur (changes requested):
Looks good to me, just a small naming comment.

slickwarren (approved):

caliskanugur (approved):

Issue #38011, opened 2022-06-15
SupportConfig Winnowing issues


MbolotSuse:
**Rancher Server Setup**
- Rancher version: 2.6.6-rc1
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): heml

**User Information**
- What is the role of the user logged in? Admin
   
**Describe the bug**
The supportconfig has several issues which need to be resolved in order to meet the requirements of the winnowing service.

- Currently, the tar produced is not extractable using gnu-tar (though it does seem to work with bsd-tar)
- The tar file should have a name indicating that it originated from rancher (supportconfig_rancher.tar)
- For a supportconfig produced for AWS, the csp name should be set to `EC2` to mirror what is produced by the suse support config. To be explicit, the field should look like the following:
```json
{
  "csp": {
      "name": "EC2"
  }
}

```

In the current state, none of these requirements are met.

**To Reproduce**
Download a supportconfig. Observe the incorrect file name. Attempt to extract the file with gnu-tar. Note that the extract fails. Extract the file with bsdtar. Observe that the config.json has the csp name set to aws.

**Result**
The issues noted above are present.

**Expected Result**
The filename is correct, the tar can be extracted using gnu-tar, and the csp name is set to `AWS`

PR #38012, opened 2022-06-15
Winnowing fix

MbolotSuse:
Fix for rancher/rancher#38011

### DIFF
--- a/pkg/api/steve/supportconfigs/generator.go
+++ b/pkg/api/steve/supportconfigs/generator.go
@@ -67,7 +67,7 @@ func (h *GeneratorHandler) ServeHTTP(writer http.ResponseWriter, request *http.R
 		return
 	}
 	writer.Header().Set("Content-Type", tarContentType)
-	writer.Header().Set("Content-Disposition", "attachment; filename=\"supportconfig.tar\"")
+	writer.Header().Set("Content-Disposition", "attachment; filename=\"supportconfig_rancher.tar\"")
 	n, err := io.Copy(writer, archive)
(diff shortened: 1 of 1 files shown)


pkg/api/steve/supportconfigs/generator.go
```diff
@@ -136,7 +136,8 @@ func (h *GeneratorHandler) generateSupportConfig() (io.Reader, error) {
 		}
 	}
 
-	return &buf, nil
+	err = tw.Close()
+	return &buf, err
```
KevinJoiner, line 140: nit: 
```suggestion
	return &buf, tw.Close()
```

KevinJoiner (approved):
LGTM

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: I did think about doing this and this was what I originally had. However, I did not like the idea of having a return which called another function and returned a pre-computed value. I'm ok with one function call serving as the return, but one call and one static value just felt off to me. Let me know if you disagree - definitely see the point of having less LOC.

KevinJoiner: I think it's readable either way,  Which is the only argument to be made here. It's purely a style decision at this point, so I would leave it if you prefer this way. One argument being a static value, should be fine and is similar to what is done [in the bytes package](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/bytes/buffer.go;l=168).

maxsokolovsky (approved):

Issue #37918, opened 2022-06-06
CSP Adapter: Upgrade Strategy


MbolotSuse:
This supports the CSP adapter/AWS integration work.

A strategy needs to be devised which allows the CSP adapter to be upgraded after install. Preferably, this will allow the csp adapter to be upgraded regularly along with rancher without the need for constant user intervention.

Requirements:
- Rancher should specify a minimum version of the csp adapter. 
- When rancher is run, it should check if the adapter is installed. 
  - If the adapter is not installed, rancher should not attempt an upgrade or install
  - If the adapter is installed, rancher should check the values to ensure that the chart is at the minimum version it expects (specified through an environment variable)
    - If the adapter version is at or above the minimum version, rancher should not take any action 
    - If the adapter version is below the minimum version, rancher should upgrade the chart and re-use any user values which were previously provided

If a technical reason is discovered that prevents this from functioning as expected, this reason should be documented in this issue. At minimum, documentation will need to be provided which instructs the user on manually upgrading the csp adapter using the helm cli.

MbolotSuse:
## Validation Template

## Root Cause

There was no logic to automatically upgrade the csp adapter. 

## What was fixed, or what change have occurred

Changes:
- A new setting has been added which specifies the minimum version of the adapter
- A controller has been added which watches this setting
- When the controller is called, if the adapter is installed, it attempts to upgrade the adapter to the specified min version
- When the controller is installed, if the adapter is not installed, it does nothing
- Both the controller and the `v1/generateSUSERancherSupportConfig` endpoints now use helm/chart based logic to determine installation status

## Areas or cases that should be tested

Automatic upgrade. See below for technical procedure steps:
1. Install an older version of the adapter.
2. Upgrade rancher through a helm command (`helm upgrade ...`) to use a new image which specifies a new min version.
3. Ensure that the adapter was upgraded to the requisite minimum version `helm ls -A ...`.
4. Ensure that values which the user specified during the original installation of the chart (aws account number/role name) are not deleted.

No Install:
1. Don't install the adapter.
2. Install rancher.
3. Ensure that rancher doesn't attempt to install or upgrade the adapter. 

No Upgrade:
1. Install the adapter with a version higher than the minimum.
2. Upgrade rancher to use a new minimum version of the adapter which is still lower than the version installed in step 1.
3. Ensure that rancher doesn't attempt to upgrade the adapter.
4. Ensure that the adapter's values appear unchanged.

## What areas could experience regressions

[...]

PR #38020, opened 2022-06-16
Adding automatic upgrade for csp-adapter

MbolotSuse:
Adds the ability for rancher to upgrade the csp-adapter to a
specified minimum version.

### DIFF
--- a/pkg/api/steve/supportconfigs/generator.go
+++ b/pkg/api/steve/supportconfigs/generator.go
@@ -1,3 +1,4 @@
+// Package supportconfigs provides a HTTPHandler to serve supportconfigs. This handler should be registered at Endpoint
 package supportconfigs
 
 import (
@@ -11,6 +12,7 @@ import (
 
 	"github.com/rancher/rancher/pkg/auth/util"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	"github.com/rancher/rancher/pkg/managedcharts/cspadapter"
 	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/sirupsen/logrus"
 	authzv1 "k8s.io/api/authorization/v1"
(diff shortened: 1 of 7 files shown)


MbolotSuse:
Related to rancher/rancher#37918.

## Background

The CSP adapter is a manually installed chart that manages support entitlements purchased through AWS. For more on the adapter, see rancher/docs#4143.

While users do need to install the adapter manually (mostly due to the fact that this only can be used in AWS under certain circumstances), we do want to upgrade the adapter with each new rancher release. This will allow us to push out bug fixes/new features without manual user intervention.

## Changes
- If the adapter is not installed, do nothing
- If the adapter is installed, make sure that the version is equal to the version specified by in our env variables. If it isn't, upgrade the adapter
- When trying to download a supportconfig - instead of looking for the configmap to see if the adapter is installed, look to see if the helm chart is installed. This will help the `generateSUSERancherSupportConfig` endpoint return results even if a user has only partially uninstalled the adapter.

## Notes
- This is intended to mostly mirror similar functionality that we use for the webhook/fleet. You can find that logic for the webhook [here](https://github.com/rancher/rancher/blob/ea017a42dfef70e6407d5cad9e82a74be88c3152/pkg/controllers/dashboard/systemcharts/controller.go#L71). There are a couple of differences, namely:
  - The webhook/fleet are always installed. Because of this, they don't have the extra check to see if this install is necessary. Since we only want to upgrade the adapter if it is installed, we need slightly different logic

pkg/controllers/dashboard/cspadaptercharts/controller.go
```diff
@@ -0,0 +1,51 @@
...
+
+type handler struct {
+	sync.Mutex
+	manager     *system.Manager
+	adapterUtil *cspadapter.ChartUtil
+}
+
+func (h *handler) onSetting(key string, setting *v3.Setting) (*v3.Setting, error) {
+	if setting == nil {
+		return nil, nil
```
maxsokolovsky, line 34: Question: why not return the setting here just as you do below, inside the if statement? Does the caller of `onSetting` handle the `nil, nil` return values properly?

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -54,12 +57,23 @@ func (h *GeneratorHandler) ServeHTTP(writer http.ResponseWriter, request *http.R
...
 	}
+	_, err = h.adapterUtil.GetRelease()
+	if err != nil {
+		if errors.Is(err, cspadapter.ErrNotFound) {
+			// if the adapter is not installed, return a 501 so the user knows to install the adapter
+			util.ReturnHTTPError(writer, request, http.StatusNotImplemented, "csp-adapter must be installed in order to to generate supportconfigs")
+			return
+		}
+		logrus.Warnf("[%s] Error when attempting to determine if adapter is installed, %s", logPrefix, err)
+		util.ReturnHTTPError(writer, request, http.StatusInternalServerError, "internal server error")
```
maxsokolovsky, line 68: Nit: for the last argument, it'd be a good practice to return the text for the HTTP status code using the `StatusText(code int) string` function from the `net/http` package.

maxsokolovsky, line 68: Same here with the error text for common HTTP errors.

pkg/controllers/dashboard/cspadaptercharts/controller.go
```diff
@@ -0,0 +1,51 @@
...
+	adapterRelease, err := h.adapterUtil.GetRelease()
+	if err != nil {
+		if errors.Is(err, cspadapter.ErrNotFound) {
+			// If the adapter isn't installed, don't attempt an upgrade
+			return setting, nil
+		}
+		// if we can't determine the status of the adapter, stop here and attempt to re-evaluate later
+		return setting, err
+	}
+	err = h.manager.Ensure(cspadapter.ChartNamespace, adapterRelease.Name, settings.CSPAdapterMinVersion.Get(), nil, true)
```
maxsokolovsky, line 49: What if the current version is not equal to but greater than the minimum required? Would Rancher downgrade it?

MbolotSuse: This is getting more into the functionality of the Ensure method that we are using here, since that is what decides if the app should be installed/upgraded.

That being said, I believe that it will only upgrade if the desired version (min) is higher than the actual version. See [here](https://github.com/rancher/rancher/blob/952e27f9dd9bf9ebfb2226795fd2d34febcb0fd4/pkg/catalogv2/system/system.go#L409-L431) for the relevant code.

MbolotSuse:
Note: This is now dependent on rancher/charts#1949 due to an upgrade in the min version specified in the dockerfile.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: Good call, Change made.

MbolotSuse: Not sure if this comment was left in the right place, but I updated both L68 and L80 with this.

Good call, change made.

maxsokolovsky (dismissed):

pkg/controllers/dashboard/cspadaptercharts/controller.go
```diff
@@ -0,0 +1,51 @@
...
+		return setting, nil
+	}
+	adapterRelease, err := h.adapterUtil.GetRelease()
+	if err != nil {
+		if errors.Is(err, cspadapter.ErrNotFound) {
+			// If the adapter isn't installed, don't attempt an upgrade
+			return setting, nil
+		}
+		// if we can't determine the status of the adapter, stop here and attempt to re-evaluate later
+		return setting, err
```
KevinJoiner, line 47: nit:
wrap error from external packages

pkg/controllers/dashboard/cspadaptercharts/controller.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	err = h.manager.Ensure(cspadapter.ChartNamespace, adapterRelease.Name, settings.CSPAdapterMinVersion.Get(), nil, true)
+	return setting, err
```
KevinJoiner, line 50: nit:
wrap error from external packages

pkg/api/steve/supportconfigs/generator.go
```diff
@@ -29,41 +30,54 @@ const (
...
+		if errors.Is(err, errNotFound) {
+			util.ReturnHTTPError(writer, request, http.StatusServiceUnavailable, "supportconfig not yet generated, try again later")
 			return
 		}
 		logrus.Errorf("[%s] Error when generating supportconfig %v", logPrefix, err)
-		util.ReturnHTTPError(writer, request, http.StatusInternalServerError, "Internal Server Error")
+		util.ReturnHTTPError(writer, request, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
 		return
 	}
 	writer.Header().Set("Content-Type", tarContentType)
```
KevinJoiner, line 83: This comment is for line 85, I can't leave a line comment because of this [issue](https://github.com/isaacs/github/issues/1655) with Github. 

The error return from `Copy` is not being handled.
```golang
n, err := io.Copy(writer, archive)
```

pkg/controllers/dashboard/cspadaptercharts/controller.go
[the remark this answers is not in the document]
MbolotSuse: I'm going to disagree with you here. This case does not represent an error situation - and as such, I don't want to return an error which indicates that the function has failed. I can change this to:

```go
return setting, nil
```

Which is semantically different, but functionally the same. Or I can combine this and the "this isn't my setting" case, which would change this to 

```go
if setting == nil || setting.Name != settings.CSPAdapterMinVersion.Name{
    return setting, nil
}
```
This approach is taken by [the fleetworkspace controller](https://github.com/rancher/rancher/blob/9bcffaa9cb7259a223280de1c50050a443267362/pkg/controllers/provisioningv2/fleetworkspace/controller.go#L70). Let me know which you prefer, personally I'm leaning toward the second.

pkg/api/steve/supportconfigs/generator.go
MbolotSuse: Good call, will fix.

pkg/managedcharts/cspadapter/cspadapter.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func NewChartUtil(clientGetter genericclioptions.RESTClientGetter) *ChartUtil {
+	return &ChartUtil{
+		restClientGetter: clientGetter,
+	}
+}
+
+// GetRelease finds the release for the CSP adapter. If not found, returns nil, ErrNotFound.
+func (c ChartUtil) GetRelease() (*release.Release, error) {
```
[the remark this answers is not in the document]
MbolotSuse: Good call, will fix.

pkg/settings/setting.go
```diff
@@ -123,6 +123,7 @@ var (
 
 	FleetMinVersion          = NewSetting("fleet-min-version", "")
 	RancherWebhookMinVersion = NewSetting("rancher-webhook-min-version", "")
+	CSPAdapterMinVersion     = NewSetting("csp-adapter-min-version", "")
```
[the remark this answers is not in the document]
MbolotSuse: Not sure that I understand all of this comment - can you elaborate more?

This is grouped with the prior 3 settings because it is closest to them in terms of functionality (each controls the minimum applicable version). I'm ok with adding comments to explain the setting, but I definitely think that it should be grouped with similar settings rather than put in alphabetical order.

pkg/controllers/dashboard/cspadaptercharts/controller.go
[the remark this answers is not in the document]
MbolotSuse: Taking the second approach - resolving.

pkg/settings/setting.go
```diff
@@ -144,11 +148,15 @@ var (
 	// If set to false the kubeconfig will contain a command to login to Rancher.
 	KubeconfigGenerateToken = NewSetting("kubeconfig-generate-token", "true")
 
-	// Deprecated: On removal use kubeconfig-default-ttl-minutes for all kubeconfigs.
 	// KubeconfigTokenTTLMinutes currently is used to set the TTL for kubeconfigs created through the CLI.
 	// This can be done with the token command or via kubectl when kubeconfig-generate-token is false.
 	// This TTL is used regardless of the value of kubeconfig-default-ttl-minutes.
+	//
+	// Deprecated: On removal use kubeconfig-default-ttl-minutes for all kubeconfigs.
```
KevinJoiner, line 155: Thank you for fixing this.

KevinJoiner (dismissed):
LGTM

KevinJoiner (approved):

PR #38021, opened 2022-06-16
Explicitly set Graph Endpoint for new Graph SDK client

maxsokolovsky:
Related to https://github.com/rancher/rancher/issues/29306, there has been an issue found by QA in their tests of China endpoints for Microsoft Azure AD.
Issue: https://github.com/rancher/rancher/issues/38022

The SDK client we use implicitly sets the global Graph Endpoint value in its constructor. This PR explicitly sets the Graph Endpoint necessary at runtime. The global (https://graph.microsoft.com) is not the only Graph Endpoint, since there is also one for China and other regions. The code now passes it from the auth config to this base contractor for the client.


rmweir (approved):
LGTM

PR #38028, opened 2022-06-17
Specify the openid scope for MFA to work with the Azure AD app

maxsokolovsky:
Rancher now sets the proper permission scope for its redirect URL for Azure AD, so that MFA works with a specific app registration in Azure AD.


MbolotSuse (approved):
LGTM. We should release note that the old way now works with MFA because this could suddenly cause MFA to kick in for some people's set up.

rmweir (approved):
LGTM

PR #38043, opened 2022-06-21
Move images from SLE to SLE-BCI

macedogm:
This PRs does:

1. Move  `registry.suse.com/suse/sle15:15.3` images to `registry.suse.com/bci/bci-base:15.3`. Both images are actually the same. This is just to keep consistency with the SLE-BCI base image migration.
2. Move back `agent` image from `registry.suse.com/bci/golang:1.17` to `registry.suse.com/bci/bci-base:15.3`, reverting what was done in [this](https://github.com/rancher/rancher/commit/ff947b4b4fcaa4cfb9af8a77e0ac4b9f0639b380#diff-36cad7767bcc6ab748db81cc23e6437e58558277252b4f90acc76d0a0419850dL5-R5) commit. The `agent` doesn't need to run the Go image, which is 7x bigger. See SURE-4330.
3. Update Helm from `v3.8.0` to `v.3.9.0` to get some security patches.


package/Dockerfile
```diff
@@ -53,7 +53,7 @@ ENV LINODE_UI_DRIVER_VERSION v0.3.0
 # make sure the version number is consistent with the one at Line 100 of pkg/data/management/machinedriver_data.go
 ENV DOCKER_MACHINE_HARVESTER_VERSION v0.4.0
 ENV CATTLE_KDM_BRANCH ${CATTLE_KDM_BRANCH}
-ENV HELM_VERSION v3.8.0
+ENV HELM_VERSION v3.9.0
```
kinarashah, line 56: Could we also update HELM_VERSION in https://github.com/rancher/rancher/blob/release/v2.6/Dockerfile.dapper#L25 and https://github.com/rancher/rancher/blob/release/v2.6/tests/validation/Dockerfile.v3api#L9

kinarashah (approved):

Issue #37969, opened 2022-06-10
Rancher panics try to use local login when azure AD is enabled


anupama2501:
**Rancher Server Setup**
- Rancher version: `v2.5.13` upgraded to `v2.6-head 6e7308f `
- Installation option (Docker install/Helm Chart): Docker install
   - If Helm Chart, Kubernetes Info: k3s
         Cluster Type (RKE1, RKE2, k3s, EKS, etc):
         Node Setup: 1 node 
         Version: `v1.23.6+k3s1
- Proxy/Cert Details: `self-signed`

**Information about the Cluster**
- Kubernetes version: `v1.20.15`
- Cluster Type (Local/Downstream): `Downstream infrastructure provider 1 worker, 1 etcd, 1 cp RKE1 Linode`

**Describe the bug**
Loggin in as a local user when azure AD is enabled causes a rancher panic. Seen on a rancher upgraded from v2.5.13 to v2.6-head.

**To Reproduce**
1. Install rancher on v2.5.13
2. Enable Azure AD with user1 as an admin
3. Create an rke1 downstream node driver cluster with 1 etcd, 1 cp, 1 worker.
4. Add a user user2 as a cluster Owner in the cluster 
5. Update the admin user to user2
6. Upgrade rancher server
7. Upgrade goes through fine
8. Switch the login to local user 

**Result**
Rancher panics:
```
2022/06/10 19:41:01 [ERROR] Error refreshing token principals, skipping: invalid server config. only exactly 1 server is currently supported
E0610 19:41:01.244462      44 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 2921 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x3890740, 0x6fd9020})
	/go/pkg/mod/k8s.io/apimachinery@v0.23.3/pkg/util/runtime/runtime.go:74 +0x7d
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xfffffffe})
	/go/pkg/mod/k8s.io/apimachinery@v0.23.3/pkg/util/runtime/runtime.go:48 +0x75
panic({0x3890740, 0x6fd9020})
	/usr/lib64/go/1.17/src/runtime/panic.go:1038 +0x215
[...]

PR #38061, opened 2022-06-22
Fix panic on refresh; properly parse principal ID before getting a user

maxsokolovsky:
Main issue: https://github.com/rancher/rancher/issues/29306
Direct issues: https://github.com/rancher/rancher/issues/37969 and https://github.com/rancher/rancher/issues/38067

The PR does two things:
1. Sets refresh callbacks on service principle token (SPT) explicitly to nil. This had been accidentally removed when a major refactoring of the old Azure AD client was taking place. The nil-pointer panic happened in the third-party SDK if this refresh callback is not set to nil. It would happen on periodic principal refresh that Rancher does to see if an Azure user known to Rancher still exists in Azure.
 
2. Properly parses the principal ID and extracts the actual user ID for the old flow via Azure AD Graph.

### DIFF
--- a/pkg/auth/providers/azure/azure_provider.go
+++ b/pkg/auth/providers/azure/azure_provider.go
@@ -86,7 +86,11 @@ func (ap *azureProvider) RefetchGroupPrincipals(principalID, secret string) ([]v
 
 	logrus.Debug("[AZURE_PROVIDER] Started getting user info from AzureAD")
 
-	userPrincipal, err := azureClient.GetUser(principalID)
+	parsed, err := clients.ParsePrincipalID(principalID)
+	if err != nil {
+		return nil, err
+	}
+	userPrincipal, err := azureClient.GetUser(parsed["ID"])
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/auth/providers/azure/clients/ad_graph_client.go
+++ b/pkg/auth/providers/azure/clients/ad_graph_client.go
@@ -142,13 +142,11 @@ func NewADGraphClientFromCredential(config *v32.AzureADConfig, credential *v32.A
 		credential.Code,
 		config.RancherURL,
 		config.GraphEndpoint,
-		nil,
 	)
 	if err != nil {
 		return nil, err
 	}
 
-	spt.SetRefreshCallbacks(nil)
 	if err := spt.Refresh(); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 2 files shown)


pkg/auth/providers/azure/azure_provider.go
```diff
@@ -439,3 +440,17 @@ func (ap *azureProvider) GetUserExtraAttributes(userPrincipal v3.Principal) map[
 	}
 	return extras
 }
+
+// userIDFromPrincipalID strips the leading identifier of the principal ID to extract only the ID itself.
+// For example, in a principle ID in the form of azuread_user://a1b2c3, the actual user ID is the a1b2c3 part.
+func userIDFromPrincipalID(principalID string) string {
```
MbolotSuse, line 446: It looks like you already have a method which does a good amount of parsing for you - namely, [ParsePrincipalID](https://github.com/maxsokolovsky/rancher/blob/e848ba6dc91931e43875c88fa5ef89a55f012138/pkg/auth/providers/azure/clients/extract.go#L47). Is there a way that we can leverage this method rather than making an entirely new method, which increases our surface area for bugs in the future?

pkg/auth/providers/azure/clients/ad_graph_client.go
```diff
@@ -187,6 +187,7 @@ func NewAzureADGraphClientFromADALToken(config *v32.AzureADConfig, adalTokenSecr
 	if err != nil {
 		return nil, err
 	}
+	spt.SetRefreshCallbacks(nil)
```
MbolotSuse, line 190: Would you mind elaborating on why this is necessary? From what I can see, you set the callbacks to nil when you construct the client, and I don't see any internal functionality that would result in this being set to not nil. Is this maybe an artifact of `callbacks ...TokenRefreshCallback` getting converted in a weird way?

MbolotSuse (changes requested):
Small comment regarding a re-implemented function.

pkg/auth/providers/azure/clients/ad_graph_client.go
KevinJoiner: The construction of the client does not actually set `refreshCallbacks` to nil in line 145 and line 185. The constructors are variadic functions that take `TokenRefreshCallback` which is a function type. When you pass in nil in the constructors you are creating a list of one nil function.
``` golang
callbacks := []TokenRefreshCallback{nil}
```
instead, we want the list to be nil which is done by not providing any arguments. If you make that change you shouldn't need to call `SetRefreshCallbacks(nil)`

KevinJoiner (changes requested):

pkg/auth/providers/azure/clients/ad_graph_client.go
maxsokolovsky: @KevinJoiner, good call, thanks!

KevinJoiner (approved):
LGTM

MbolotSuse (approved):
LGTM

Issue #38064, opened 2022-05-31
`rke2 secrets-encrypt` commands get stuck when clusters have split etcd & controlplane nodes


jakefhyde:
**Environmental Info:**
RKE2 Version: v1.23.6+rke2r2

Cluster Configuration:

1 etcd, 1 controlplane, 1 worker
3 etcd, 2 controlplane, 3 worker

**Describe the bug:**

On split role clusters (i.e. etcd and controlplane components are split via `--disable` flags), the `rke2-server.service` is stuck restarting following the `rke2 secrets-encrypt rotate` command.

**Steps To Reproduce:**

- Installed RKE2: Rancher provisioned, but also reproducible standalone
- Create a script that does the following:
  - Run `rke2 secrets-encrypt prepare`
  - Scrape `rke2 secrets-encrypt status` to verify that the current stage is `prepare`
  - Restart an etcd node so the bootstrap data containing the encryption key configuration is committed to the datastore
  - Restart the control plane node the commands were run on using `systemctl restart rke2-server`
  - Run `rke2 secrets-encrypt rotate`
  - Scrape `rke2 secrets-encrypt status` to verify that the current stage is `rotate`
  - Restart an etcd node so the bootstrap data containing the encryption key configuration is committed to the datastore
  - Restart the control plane node the commands were run on using `systemctl restart rke2-server`

**Expected behavior:**

Control plane successfully restarts post `rke2 secrets-encrypt rotate`, and allows encrypt key rotation to proceed

**Actual behavior:**

Control plane is stuck in `Activating` state, with the following error message:

```
May 31 18:28:42 jhyde-do-yndaf-pool1-4d7308a6-tf8v2 systemd[1]: rke2-server.service: Scheduled restart job, restart counter is at 6.
May 31 18:28:42 jhyde-do-yndaf-pool1-4d7308a6-tf8v2 systemd[1]: Stopped Rancher Kubernetes Engine v2 (server).
May 31 18:28:42 jhyde-do-yndaf-pool1-4d7308a6-tf8v2 systemd[1]: Starting Rancher Kubernetes Engine v2 (server)...
May 31 18:28:42 jhyde-do-yndaf-pool1-4d7308a6-tf8v2 sh[11530]: + /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service
[...]

PR #38065, opened 2022-06-22
Fixed etcd-only plans always in sync post-prepare

jakefhyde:
Issue: https://github.com/rancher/rancher/issues/38064

 
# Problem

Cluster with etcd only nodes require the etcd nodes to be restarted during encryption key rotation before restarting the control plane nodes.

# Solution

 
There was a bug in rancher's encryption key rotation implementation where etcd nodes would not be restarted on subsequent iterations, because their plans were already in sync. This was fixed by adding an environment variable which utilizes the current control plane encryption key rotation phase to cause the plans to rerun. ~Also added `--force` to plan for `secrets-encrypt` command, so encryption key rotation can be reattempted, otherwise cluster could be stuck forever if something went wrong past prepare.~

# Testing

### Manual Testing

Tested the following cases:
#### v1.23.7-k3s-4dda76b0
configuration | status
-|-
1 node all roles | ✔️
1 etcd, 1 cp, 1 worker | ✔️
3 nodes all roles | ✔️
3 etcd, 2 cp, 3 worker | ✔️
3 etcd, 3 cp, 3 worker | ✔️
3 etcd, 5 cp, 3 worker | ✔️
## QA Testing Considerations

 
This realistically should be tested with varying cloud providers, and varying kubernetes versions. Currently, my testing only covers the 1.23 release line for k3s, however these changes are not in the latest KDM release for rancher, so while this will hit rancher for v2.6.6, encryption key rotation may be enabled OOB.

Encryption key rotation has the potential to break everything, so try to mess with the cluster as much as possible.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
+++ b/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
@@ -84,13 +84,13 @@ generationFile="$dataRoot/generation"
 currentGeneration=$(cat "$generationFile" || echo "")
 
 if [ "$currentGeneration" != "$targetGeneration" ]; then
-  $runtime $@
+	$runtime $@
 else
 	echo "action has already been reconciled to the current generation."
 fi
 
-mkdir -p $dataRoot
-echo $targetGeneration > "$generationFile"
+mkdir -p "$dataRoot"
+echo "$targetGeneration" > "$generationFile"
 `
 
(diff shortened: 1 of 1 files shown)


phillipsj (approved):

rosskirkpat (approved):

phillipsj (changes requested):
Can we split the go mod and sum changes into a separate commit from the code changes?

phillipsj (approved):

annablender (approved):
High level

Oats87 (approved):

PR #38066, opened 2022-06-22
[Automation] - Long Living Envs on Linode

izaac:
This terraform module provides a way to provision a couple of Rancher Hosted instances below a Super Rancher and each with a downstream custom cluster where Rancher v2 Apps are installed. Everything using the Linode Linodes.

- Super Rancher
  - Hosted Rancher 1
    - Custom Cluster 1
  - Hosted Rancher 2
    - Custom Cluster 2

So far only valid certs are supported so the certs should be inside the rancher_hosted module. The certificates are the standard QA certificates we use for automation so they have to be replaced every three months in the nginx load balancers (see nginx configuration template path)

Certificates path: `modules/rancher-hosted/scripts/certs` (automatically created from base64 strings)

- `fullchain.pem`
- `privkey.pem`

The nginx load balancer configuration template path: `modules/rancher-hosted/scripts/nginx/nginx.conf`

A path to the private ssh key using the `ssh_private_key_path` variable is also required, this is for the recollection of the k3s kubeconfig files and k3s token files too. This key should match with the public `ssh_authorized_key` variable provided for the Linode ssh configuration.

A `terraform.tfvars` is required to set the variables easily for the module to work, an example of such file:

```bash
linode_token="<the linode token>"
linode_root_password="<linode strong password>"
linode_resource_prefix="<linode resources prefix"
zone_id="<Route53 zone ID>"
load_balancers_domain="domain to use"
super_load_balancer_subdomain="subdomain for the super Rancher"
hosted1_load_balancer_subdomain="subdomain for the hosted1 Rancher"
hosted2_load_balancer_subdomain="subdomain for the hosted2 Rancher"
ssh_authorized_key="ssh-rsa AAAA..."
rancher_password="<Rancher password> this will be set to all the Ranchers"
rancher_version_super="2.6.5"
rancher_version_hosted1="2.6.5"
rancher_version_hosted2="2.5.14"
k3s_version_super="v1.22.9+k3s1"
k3s_version_hosted1="v1.22.9+k3s1"
k3s_version_hosted2="v1.20.12+k3s1"
ssh_private_key="<base64 private key string>"
fullchain=<base64 certificate string>
privkey=<base64 certificate key string>
```
AWS [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#authentication-and-configuration).
After these requirements are met it'll be just matter of `terraform init` and `terraform apply`


git-ival (changes requested):
I haven't `terraform apply`-ed this yet but I have some minor feedback:

If we aren't pulling any data sources can we delete `tests/validation/tests/v3_api/resource/terraform/longliving/linode/data.tf`?

Also, outputting the urls and other potentially useful information for the Super and Hosted Ranchers might be helpful. Otherwise I think we should delete any empty files, like the `outputs.tf` files currently are.

Will we need/want equivalent terraform modules for RKE1 and RKE2?

izaac:
the linode firewall is optional, you'll be able to access the nodes externally but i can explore the possibility of adding specific or more strict rules to only allow connectivity to certain ports.

PR #38087, opened 2022-06-24
automation framework to test rancher2 TF provider clusters

Josh-Diamond:
with terratest + go; 
refactored to use dynamic clients

### DIFF
--- a/tests/terratest/config/config.go
+++ b/tests/terratest/config/config.go
@@ -1,0 +1,91 @@
+package config
+
+import (
+	"github.com/rancher/rancher/tests/terratest/models"
+)
+
+// Modules
+var Aks                = "aks"
+var K3s                = "k3s"
+var Rke1               = "rke1"
+var Rke2               = "rke2"
+
+// K8s versions
+var AKSK8sVersion1235  = "1.23.5"
+var AKSK8sVersion1226  = "1.22.6"
+var AKSK8sVersion1219  = "1.21.9"
+
+var K3sK8sVersion1237  = "v1.23.7+k3s1"
+var K3sK8sVersion12210 = "v1.22.10+k3s1"
+var K3sK8sVersion12113 = "v1.21.13+k3s1"
+
+var RKE1K8sVersion1237  = "v1.23.7-rancher1-1"
+var RKE1K8sVersion12210 = "v1.22.10-rancher1-1"
+var RKE1K8sVersion12113 = "v1.21.13-rancher1-1"
+
+var RKE2K8sVersion1237  = "v1.23.7+rke2r2"
+var RKE2K8sVersion12210 = "v1.22.10+rke2r2"
+var RKE2K8sVersion12113 = "v1.21.13+rke2r2"
+
+// Nodes3_Etcd1_Cp1_Wkr1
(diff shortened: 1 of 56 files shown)


tests/terratest/config/config.go
```diff
@@ -0,0 +1,91 @@
...
+var K3s                = "k3s"
+var Rke1               = "rke1"
+var Rke2               = "rke2"
+
+// K8s versions
+var AKSK8sVersion1235  = "1.23.5"
+var AKSK8sVersion1226  = "1.22.6"
+var AKSK8sVersion1219  = "1.21.9"
+
+var K3sK8sVersion1237  = "v1.23.7+k3s1"
```
izaac, line 18: Are we going to need to continuously update these k8s values? Could this be added as a configuration file that dynamically loads these values from ?

Could it be the same way the golang framework loads a conf file and is used as a variable in the Jenkins execution.

izaac (commented):
Just a comment, I will defer the Golang review to @igomez06 once he's back from PTO

tests/terratest/config/config.go
Josh-Diamond: Yes, i suppose that would be fine as well.. they are fixed values, so they wouldn't need to be updated per-say, but new versions would need to be declared for future use.  These values are only being referenced within the test files to build the TF  main.tf files, so as long as the test is grabbing the right values, they could come from anywhere.  Once declared, they will always be available to use in any test, so adding new ones will be necessary, but no reason to change versions, unless deprecated, then they could be removed from this config.go file.  

Either way would work fine, i initially thought of this config.go as a "component" config library with defined, reusable values to build tests with.

Josh-Diamond: I feel like its best the way I have it, because if externally provided, i think it would hard to keep consistency between tests being written, like naming conventions and knowing which are needed to run specific tests, as it wouldn't be available for reference anywhere.  You would need to examine the test, to see what the variables are called and then figure out the appropriate values.  With these values centralized and defined in the repo's config.go, it is easy to reference both the key:value in the config, as well as the reference to that value in the tests.

tests/terratest/functions/waitForActiveCluster.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/stretchr/testify/require"
+)
+
+func WaitForActiveCluster(t *testing.T, client *rancher.Client, clusterID string) {
```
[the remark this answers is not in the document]
igomez06: This is the only thing I don't like, as this a busy wait call.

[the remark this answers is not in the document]
Josh-Diamond: Correct, i was not able to get a watch interface or poll implemented but refactored to use dynamic clients.

If this is not adequate, perhaps we can sync up at some point to go over the implementation you're suggesting?
K8s 1.24 support takes priority for me right now, so I won't be available to take this back up until that is completed.

Thanks Israel! Your insight is invaluable to me 🚀

PR #38089, opened 2022-06-24
Add minor file support to v2prov to allow for immediate plan updates

Oats87:
This PR adds functionality to allow v2prov to skip the conventional drain/cordon plan update operations, when a plan change is minor. Minor in this case is that the only change in the plan are files that are considered "minor", as designated by the v2prov engineering team. Currently, the two files that are minor are the etcd-metadata and cluster agent manifest files.

https://github.com/rancher/rancher/issues/37465


jakefhyde (approved):
LGTM

Issue #37965, opened 2022-06-10
fix gouroutine leak when sharedcontroller has been started


samjustus:

samjustus:
[lasso PR](https://github.com/rancher/lasso/pull/30)

JacieChao:
# Problem 

If there's something wrong with downstream clusters and usercluster controller has to restart frequency, will cause goroutine leaks with management controller factory. 
When the usercluster controller starts to handle a downstream cluster, we also start management controllers [here](https://github.com/rancher/rancher/blob/release/v2.5/pkg/types/config/context.go#L490)  where the goroutine leaks problem happens.  

# Solution 

1. Rancher will start a usercluster controller for each downstream cluster and will start management controller factory [here](https://github.com/rancher/rancher/blob/release/v2.5/pkg/types/config/context.go#L490). The management controller factory will start only once because it checked `started` state [here](https://github.com/rancher/lasso/blob/release/v2.5/pkg/controller/controller.go#L153). If there's something wrong with the downstream cluster and usercluster controller has to stop, it canceled the context and skip goroutine [here](https://github.com/rancher/lasso/blob/release/v2.5/pkg/controller/sharedcontroller.go#L97). The management controller factory context is never canceled if Rancher is still running so that each time we start a usercluster controller, the goroutine leaks [here](https://github.com/rancher/lasso/blob/release/v2.5/pkg/controller/sharedcontroller.go#L92). So we also need to check `started` state to avoid creating more goroutine leaks if the controller has started. 

2. It set `started` state in Start() function but never set `unstarted` state for shutting down works when channel done. The controller may fail to start when we restart it. 

# Testing 

## Engineering Testing 

### Manual Testing 

1. Deploy HA Rancher and set up Prometheus dashboard for controllers. Refer to: https://github.com/rancher/rancher/wiki/How-to-setup-prometheus-dashboard-for-controllers 
2. Add an import cluster to Rancher and run the following scripts to test usercluster controller restart. 
``` 
#!/bin/bash

# =============== Usage ===============
# ./test.sh <cluster-id> <limit-times> <interval>
# ./test.sh c-xxxx 100 3
# Need to running on the node which can access to Rancher local cluster(Need to running on the node which can access to Rancher local cluster).
# =======================================

clusterid=$1
limit=$2
interval=$3

n=1
while [ $n -le $limit ]
do
    cluster=$(kubectl get clusters.management.cattle.io $clusterid -o json)
[...]

PR #38091, opened 2022-06-24
Bump lasso version

KevinJoiner:
lasso fix for #37965


rancherbot (approved):

Issue #38090, opened 2022-06-24
Discontinue re-election of init node if fixed init node cannot be found


Oats87:
see: https://github.com/rancher/rancher/issues/37462

In certain cases when an init node was designated by machine ID, Rancher would run into an error state when the machine could not be found.

sowmyav27:
Reopening for QA Test Template

Oats87:
In all honesty, I'm not actually sure how we can really test this effectively, without having the harvester team test this with rancherd. Perhaps we should make this dev-validate.

Josh-Diamond:
Ticket #38090 - Test Results

With HA Helm:

Verified on rancher `v2.6-988b8d1f63c19b13f6afba9de2e34e30cac133a7-head`:
1. Fresh install of rancher `v2.6-head`
2. Provision a downstream RKE2 Custom cluster w/ label [`rke.cattle.io/init-node-machine-id` | `lolnotvalid`]
3. Immediately upon registering node to cluster, `unable to find designated init node matching machine ID lolnotvalid` is observed
4. Verified - Cluster never bootstraps; `unable to find designated init node` is still observed after 5 minutes; No resolve as expected

Screenshots:

`Step 3`
<img width="1647" alt="Screen Shot 2022-07-15 at 11 24 40 AM" src="https://user-images.githubusercontent.com/46494969/179288309-f0329fd2-c073-43d9-97ac-cb77e9352fb3.png">

`Step 4`
<img width="1650" alt="Screen Shot 2022-07-15 at 11 28 43 AM" src="https://user-images.githubusercontent.com/46494969/179288424-e542f442-c315-47f5-9b08-61e1d8de4e95.png">

`Step 4 - after 10+ mins`
<img width="1652" alt="Screen Shot 2022-07-15 at 11 36 16 AM" src="https://user-images.githubusercontent.com/46494969/179289358-b5cb3b33-9743-4c40-956e-902fc12d4e82.png">

---

Regression Consideration - `ensure standard provisioning election of init nodes still works`

Verified on rancher v2.6-988b8d1f63c19b13f6afba9de2e34e30cac133a7-head`:
1. Fresh install of rancher`v2.6-head`
2. Provision a downstream RKE2 Custom cluster WITHOUT adding `rke.cattle.io/init-node-machine-id` label
3. Verified - standard provisioning election of init nodes work as expected; cluster successfully provisions

Screenshots:

`Step 3`
[...]

PR #38092, opened 2022-06-24
[v2prov] don't elect init node if not found and one is designated

Oats87:
https://github.com/rancher/rancher/issues/38090

If an init node was designated, Rancher was continuously trying to re-elect the init node if the init node was not found by machine ID. This was incorrect behavior, and Rancher should not elect an init node if one could not be found as it was designated.


jakefhyde (approved):
LGTM

HarrisonWAffel (approved):

Issue #37250, opened 2022-04-08
Clusters become unreachable via UI and kubectl after some times


Lebvanih:
**Rancher Server Setup**
- Rancher version: 2.6.4
- Installation option (Docker install/Helm Chart): Helm, RKE1, K8S 1.21.10

**Information about the Cluster**
- Kubernetes version: K8S 1.20.9
- Cluster Type (Local/Downstream): Downsteam, the remote cluster was provisioned using docker command. We are able to reproduce this on two different cluster.

**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**
We noticed since the upgrade from 2.5.8 to 2.6.4 that our cluster would become unresponsive via the UI and via Kubectl. However, weirdly enough, if we log via the old ui (removing the /dashboard in the url and add login), we still can't see everything without issue... 

**To Reproduce**
The fastest way we found to reproduce is to spam a few times (10-15, not in parallel) the following command toward the cluster, the cluster is quite busy so we get a bunch of logs doing this:
``` kubectl logs -n ingress-nginx -f <nignx-ingress-pod>``` or ``` kubectl logs -n ingress-nginx <nignx-ingress-pod>```

**Result**
After a few attempts, it will hang, and any command via kubectl will hang (either all the time or once every 2 attempts at the start). The cluster is still displayed as connected and healthy in the UI. 

**Expected Result**
The command shouldn't hang making the interaction with the cluster impossible.
The situation never recover we are forced to kill all the rancher pod to restore the UI + kubectl.

**Additional context**
We are unsure if the kubectl logs is the reason or if it's just a quantity of data that is enough to block something in rancher.. We are wondering if we aren't in a similar case of this one where the routing that handle the forward to the cluster is blocked by a full buffer or something similar. But as the ticket is close, we aren't sure if it could be the same. https://github.com/rancher/rancher/issues/34819#issuecomment-1007600901 

We also see alot of message mentioned in this one: https://github.com/rancher/rancher/issues/36584 but we aren't sure if they are the cause of our outages.

Also note that restarting the agent on the downstream or forcing a full connection interrupt between the agents and the cluster to make them reconnect doesn't fix the issue either, only a full kill of the rancher pod fixes our issue at the moment

SURE-4484

cambierr:
Exact same issue here....

gray314:
the same issue.. =/
[...]

PR #38093, opened 2022-06-24
Bump remotedialer for pause/resume fix

cmurphy:
https://github.com/rancher/rancher/issues/37250


rancherbot (approved):

PR #38122, opened 2022-06-28
Removing hardcoded node pool quantity

thaneunsoo:
Removed hardcoded value for machine pools to allow users to increase node count in the pool. Changed config struct to allow for NodesAndRoles to accept quantity as a value. Adjusted all tests using RKEMachinePoolSetup helper function to adjust to the new structure.


sowmyav27:
@timhaneunsoo Can we squash all commits?

PR #38125, opened 2022-06-29
Make an annotation the only variable in Azure AD auth flow choice

maxsokolovsky:
Related to https://github.com/rancher/rancher/issues/29306

UI issue with background: https://github.com/rancher/dashboard/issues/6236

The main problem is as follows: when admins upgrade Rancher to v2.6.7, while having a custom Azure AD endpoint, they will not see the banner in the UI because the UI (and the backend) only checks if the config has one of the two known old endpoints (for Global and China). Therefore, Rancher will assume it needs to use the new authentication flow via Microsoft Graph, since a custom endpoint (by virtue of being able to hold any value) cannot be known to be deprecated. As a result, admins won't even see the warning about the change and, most importantly, a reminder to add the necessary permissions to the app in Azure. This means Rancher will use the new authentication flow without notifying the admins first, which is dangerous because they will not have prepared the Azure app for that.

We want users of Global, China, Custom endpoints to see the upgrade message and button when admins upgrade Rancher, so there are no surprises and automatic endpoint updates.

This PR makes a change to how Rancher determines if an Azure AD setup uses deprecated endpoints.
Instead of looking at endpoints themselves, Rancher will now put and read a special annotation on the auth config object.

### DIFF
--- a/pkg/auth/providers/azure/azure_actions.go
+++ b/pkg/auth/providers/azure/azure_actions.go
@@ -72,6 +72,16 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 	}
 
 	azureADConfig := &azureADConfigApplyInput.Config
+
+	// This covers the case where users upgrade Rancher to v2.6.7+ without having used Azure AD as the auth provider.
+	// In 2.6.7+, whether Azure AD is later registered or not, Rancher on startup creates the annotation on the template auth config.
+	// But in the case where the auth config had been created on Rancher startup prior to v2.6.7, the annotation would be missing.
+	// This ensures the annotation is set on next auth provider setup attempt.
+	if azureADConfig.ObjectMeta.Annotations == nil {
+		azureADConfig.ObjectMeta.Annotations = make(map[string]string)
+	}
+	azureADConfig.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
+
 	azureLogin := &v32.AzureADLogin{
 		Code: azureADConfigApplyInput.Code,
 	}
@@ -110,7 +120,8 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 
 // migrateToMicrosoftGraph represents the migration of the registered Azure AD auth provider
 // from the deprecated Azure AD Graph API to the Microsoft Graph API.
-// It modifies the existing auth config value in the database, so that it has a new endpoint to the new API.
+// It modifies the existing auth config value in the database, so that it has up-to-date endpoints to the new API.
+// Most importantly, it sets the annotation that specifies that the auth config has been migrated to use the new auth flow.
 func (ap *azureProvider) migrateToMicrosoftGraph() error {
 	defer ap.deleteUserAccessTokens()
 	defer clients.GroupCache.Purge()
@@ -120,7 +131,13 @@ func (ap *azureProvider) migrateToMicrosoftGraph() error {
 		return err
 	}
 	if authProviderEnabled(cfg) {
-		updateAzureADConfig(cfg)
+		updateAzureADEndpoints(cfg)
+
+		if cfg.ObjectMeta.Annotations == nil {
+			cfg.ObjectMeta.Annotations = make(map[string]string)
+		}
+		cfg.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
+
 		err = ap.saveAzureConfigK8s(cfg)
 		if err != nil {
 			return err
--- a/pkg/auth/providers/azure/azure_provider.go
+++ b/pkg/auth/providers/azure/azure_provider.go
@@ -79,7 +79,7 @@ func (ap *azureProvider) RefetchGroupPrincipals(principalID, secret string) ([]v
 	if err != nil {
 		return nil, err
 	}
-	var useDeprecatedAzureADClient = graphEndpointDeprecated(cfg.GraphEndpoint)
+	var useDeprecatedAzureADClient = isConfigDeprecated(cfg)
 	azureClient, err := clients.NewAzureClientFromSecret(cfg, useDeprecatedAzureADClient, secret, ap.secrets)
 	if err != nil {
 		return nil, err
@@ -195,7 +195,7 @@ func (ap *azureProvider) TransformToAuthProvider(
 }
 
 func (ap *azureProvider) loginUser(config *v32.AzureADConfig, azureCredential *v32.AzureADLogin, test bool) (v3.Principal, []v3.Principal, string, error) {
-	var useDeprecatedAzureADClient = graphEndpointDeprecated(config.GraphEndpoint)
+	var useDeprecatedAzureADClient = isConfigDeprecated(config)
 	azureClient, err := clients.NewAzureClientFromCredential(config, useDeprecatedAzureADClient, azureCredential, ap.secrets)
 	if err != nil {
 		return v3.Principal{}, nil, "", err
@@ -264,7 +264,7 @@ func (ap *azureProvider) searchGroupPrincipalsByName(client clients.AzureClient,
 
 func (ap *azureProvider) newAzureClientFromToken(cfg *v32.AzureADConfig, token *v32.Token) (clients.AzureClient, error) {
 	var secret string
-	var deprecated = graphEndpointDeprecated(cfg.GraphEndpoint)
+	var deprecated = isConfigDeprecated(cfg)
 	if deprecated {
 		var err error
 		secret, err = ap.tokenMGR.GetSecret(token.UserID, Name, []*v3.Token{token})
@@ -276,6 +276,8 @@ func (ap *azureProvider) newAzureClientFromToken(cfg *v32.AzureADConfig, token *
 }
 
 func (ap *azureProvider) saveAzureConfigK8s(config *v32.AzureADConfig) error {
+	// Copy the annotations.
+	annotations := config.ObjectMeta.Annotations
 	storedAzureConfig, err := ap.getAzureConfigK8s()
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


pkg/auth/providers/azure/azure_actions.go
```diff
@@ -120,7 +132,14 @@ func (ap *azureProvider) migrateToMicrosoftGraph() error {
...
 	}
 	if authProviderEnabled(cfg) {
-		updateAzureADConfig(cfg)
+		updateAzureADEndpoints(cfg)
+
+		if cfg.ObjectMeta.Annotations == nil {
+			cfg.ObjectMeta.Annotations = map[string]string{GraphEndpointMigratedAnnotation: "true"}
+		} else {
+			cfg.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
+		}
```
KevinJoiner, line 141: nit: because I don't love else statements.
```suggestion
		if cfg.ObjectMeta.Annotations == nil {
			cfg.ObjectMeta.Annotations = map[string]string{}
		}
		cfg.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
```

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -400,6 +415,22 @@ func formAzureRedirectURL(config map[string]interface{}) string {
...
+				if stringValue, ok := v.(string); ok {
+					result[k] = stringValue
+				} else {
+					continue
+				}
+			}
+		}
+	}
+	return result
+}
```
KevinJoiner, line 432: Early returns will help flatten things out and make what is going on easier to read. We also don't nee to `else: continue.`
```suggestion
func extractAnnotations(config map[string]interface{}) map[string]string {
	result := make(map[string]string)

	metadata, ok := config["metadata"].(map[string]interface{})
	if !ok {
		return result
	}

	annotations, ok := metadata["annotations"].(map[string]interface{})
	if !ok {
		return result
	}

	for k, v := range annotations {
		if stringValue, ok := v.(string); ok {
			result[k] = stringValue
		}
	}

	return result
}
```

pkg/auth/providers/azure/config.go
```diff
@@ -19,30 +18,39 @@ const (
...
+	}
+	v, ok := cfg.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation]
+	return !ok || v != "true"
 }
 
-func updateAzureADConfig(c *v32.AzureADConfig) {
+func updateAzureADEndpoints(c *v32.AzureADConfig) {
 	if isConfigForChina(c) {
-		updateConfigForChina(c)
+		updateEndpointsForChina(c)
```
KevinJoiner, line 41: nit: 
If you desired you could use an early return instead of an else.

KevinJoiner (changes requested):

pkg/auth/providers/azure/azure_provider.go
maxsokolovsky: Agreed, will refactor.

pkg/auth/providers/azure/azure_actions.go
maxsokolovsky: Good point!

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -400,6 +415,26 @@ func formAzureRedirectURL(config map[string]interface{}) string {
...
+		return annotations
+	}
+	for k, v := range rawAnnotations {
+		if stringValue, ok := v.(string); !ok {
+			continue
+		} else {
+			annotations[k] = stringValue
+		}
+	}
+	return annotations
```
KevinJoiner, line 435:
```suggestion
	for k, v := range rawAnnotations {
		if stringValue, ok := v.(string); ok {
			annotations[k] = stringValue
		}
	}
	return annotations
```

maxsokolovsky:
Waiting for https://github.com/rancher/dashboard/pull/6237 to be merged, then will rebase and test against the new UI.

KevinJoiner (dismissed):
LGTM

pkg/auth/providers/azure/config.go
```diff
@@ -19,30 +18,39 @@ const (
...
 
 func isConfigForChina(c *v32.AzureADConfig) bool {
 	return strings.HasSuffix(c.GraphEndpoint, ".cn") || strings.HasSuffix(c.GraphEndpoint, ".cn/")
 }
 
-func updateConfigForGlobal(c *v32.AzureADConfig) {
+func updateEndpointsForGlobal(c *v32.AzureADConfig) {
 	if c.GraphEndpoint != globalAzureADGraphEndpoint {
-		logrus.Infoln("Refusing to upgrade because the Graph Endpoint is not deprecated.")
+		logrus.Infof("Refusing to upgrade because the Graph Endpoint %s is not deprecated.", c.GraphEndpoint)
```
MbolotSuse, line 53: Docs:

One note here: We will want to make sure that we include docs for users of custom endpoints so that they understand that mixing custom/standard endpoints are not supported (as well as what steps are required to upgrade a custom endpoint).

MbolotSuse (approved):
A couple of notes but nothing that needs to change. LGTM.

pkg/auth/providers/azure/config.go
maxsokolovsky: Yes, we will say that we're not responsible for any custom invalid endpoints.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -400,6 +415,24 @@ func formAzureRedirectURL(config map[string]interface{}) string {
 	)
 }
 
+func extractAnnotations(config map[string]interface{}) map[string]string {
```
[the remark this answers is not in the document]
maxsokolovsky: I agree, will add a warning in the logs.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -276,6 +276,8 @@ func (ap *azureProvider) newAzureClientFromToken(cfg *v32.AzureADConfig, token *
 }
 
 func (ap *azureProvider) saveAzureConfigK8s(config *v32.AzureADConfig) error {
+	// Copy the annotations.
+	annotations := config.ObjectMeta.Annotations
```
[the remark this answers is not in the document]
maxsokolovsky: Yeah, that's a fair point. I did have the additional boolean flag as an argument for the `saveAzureConfigK8s` method, but I found that it complicates it a bit.

KevinJoiner (approved):

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -400,6 +415,27 @@ func formAzureRedirectURL(config map[string]interface{}) string {
...
+	rawAnnotations, ok := metadata["annotations"].(map[string]interface{})
+	if !ok {
+		return annotations
+	}
+	for k, v := range rawAnnotations {
+		if stringValue, ok := v.(string); ok {
+			annotations[k] = stringValue
+		}
+	}
+	if _, ok := annotations[GraphEndpointMigratedAnnotation]; !ok {
```
MbolotSuse, line 433: I don't think this log condition meets what I was thinking about in https://github.com/rancher/rancher/pull/38125/files#r913802917 and might actually do more harm than good. My main concern here was that you have several places in this function where you return early because you couldn't parse the object into an expected form (like when you look into `config["mateadata"]` or `metadata["annotations"]`). In these cases, you do not return an error but instead return an empty map. It's impossible to distinguish these cases from the cases where there were no annotations, and since there is no log statement, there's no way of telling what happened or why we didn't parse the value.

In addition, this new log statement seems to have awareness of why we are parsing the annotation (to look for the migratedAnnotation). That seems to violate separation of responsibilities - this function doesn't care why we want the annotations (and doesn't use that value at all outside of a log warning), so it shouldn't check for a specific value in the annotations.

In short:
- This still leaves room for "ghost errors" which are not logged.
- Checking for this specific value is out of scope for this function
- Since we expect this value to be missing on legacy configurations, you are going to have a lot of these log statements for a missing annotation

I would recommend one of the following:
- Log the failure at each early return point in the function. If you also want to log in another function (such as isConfigDeprecated - not recommended) that we couldn't find the annotation that's ok, but that should be `trace`/`info` at most, not `warn`. 
- Return an error from this function at the early return points, and handle in the calling function (i.e. formAzureRedirectURL - which can choose to ignore and log `couldn't get annotations, error %s, will use old code`)

MbolotSuse (changes requested):
Outstanding issue on potential "ghost error"

pkg/auth/providers/azure/azure_provider.go
maxsokolovsky: Very good point, will make sure the logging happens before all returns from the function.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -400,6 +415,28 @@ func formAzureRedirectURL(config map[string]interface{}) string {
 	)
 }
 
+func extractAnnotations(config map[string]interface{}) map[string]string {
+	annotations := make(map[string]string)
+	metadata, ok := config["metadata"].(map[string]interface{})
+	if !ok {
+		logrus.Infof("Failed to decode the %q field of the auth config.", "metadata")
```
MbolotSuse, line 422: Nit:

I don't think that it's helpful to use a format here given that the entire statement is a constant. There's arguments here from a readability perspective (why are we formatting something when the arg is a constant) and from an efficiency perspective (instead of just printing out the line, now logrus has to parse the format string, insert the argument, render the statement, and log it out). 

Marked as nit because it's pretty minor.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -400,6 +415,28 @@ func formAzureRedirectURL(config map[string]interface{}) string {
...
+		return annotations
+	}
+	rawAnnotations, ok := metadata["annotations"].(map[string]interface{})
+	if !ok {
+		logrus.Infof("Failed to decode the %q field of the auth config.", "annotations")
```
MbolotSuse, line 427: Nit:

Same comment here on the use of a format string when the only arg is a constant.

MbolotSuse (approved):
Approving Overall from design/implementation side.

I have some small concerns on the log statements, but they are minor and don't necessarily need to be addressed.

pkg/auth/providers/azure/azure_provider.go
maxsokolovsky: I agree, but I did this because of I wanted to give the same form to these TWO logging statements. If there was only one, I would not use the argument.

MbolotSuse: I get that the log statements have the same form, but since the format isn't a variable, we don't get any benefits of having them the same. 

I.E. - if I want to change this, I still need to make a change in two places rather than one. So really they are only the same because the string in two places are the same (just like it would be if the entire log statement was one statement).

pkg/auth/providers/azure/azure_provider.go
maxsokolovsky: Adjusted!

MbolotSuse (approved):
LGTM

PR #38129, opened 2022-06-29
OPA Gatekeeper v2 p0 validation test cases

nataliefyi:
Validation Tests
Implemented:
•	OPA Gatekeeper p0 test cases
o	Helper functions to install Gatekeeper chart
o	Helper to get unstructured List from a cluster resource
o	Helper to parse unstructured list to extract nested values
o	K8srequiredlabels.yaml constraint file to be applied by test
o	P0 validations

### DIFF
--- a/tests/framework/extensions/charts/ranchergatekeeper.go
+++ b/tests/framework/extensions/charts/ranchergatekeeper.go
@@ -1,0 +1,256 @@
+package charts
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/rancher/rancher/pkg/api/steve/catalog/types"
+	catalogv1 "github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	crds "github.com/rancher/rancher/tests/framework/extensions/customresourcedefinitions"
+	"github.com/rancher/rancher/tests/framework/extensions/namespaces"
+	"github.com/rancher/rancher/tests/framework/pkg/wait"
+	"github.com/rancher/rancher/tests/integration/pkg/defaults"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+const (
+	// Namespace that rancher gatekeeper chart is installed in
+	RancherGatekeeperNamespace = "cattle-gatekeeper-system"
+	// Name of the rancher gatekeeper chart
+	RancherGatekeeperName = "rancher-gatekeeper"
+	// Name of rancher gatekeepr crd chart
+	RancherGatekeeperCRDName = "rancher-gatekeeper-crd"
+)
+
+// InstallRancherGatekeeperChart installs the OPA gatekeeper chart
+func InstallRancherGatekeeperChart(client *rancher.Client, installOptions *InstallOptions) error {
+	hostWithProtocol := fmt.Sprintf("https:// %s", client.RancherConfig.Host)
+	gatekeeperChartInstallActionPayload := &payloadOpts{
+		InstallOptions: *installOptions,
+		Name:           RancherGatekeeperName,
+		Host:           hostWithProtocol,
+		Namespace:      RancherGatekeeperNamespace,
+	}
+
+	chartInstallAction := newGatekeeperChartInstallAction(gatekeeperChartInstallActionPayload)
+
+	catalogClient, err := client.GetClusterCatalogClient(installOptions.ClusterID)
+	if err != nil {
+		return err
+	}
+
+	// Cleanup registration
+
+	// register uninstall rancher-gatekeeper as a cleanup function
+	client.Session.RegisterCleanupFunc(func() error {
+		// UninstallAction for when uninstalling the rancher-gatekeeper chart
+		defaultChartUninstallAction := newChartUninstallAction()
+
+		err := catalogClient.UninstallChart(RancherGatekeeperName, RancherGatekeeperNamespace, defaultChartUninstallAction)
+		if err != nil {
+			return err
+		}
+
+		watchAppInterface, err := catalogClient.Apps(RancherGatekeeperNamespace).Watch(context.TODO(), metav1.ListOptions{
+			FieldSelector:  "metadata.name=" + RancherGatekeeperName,
+			TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+		})
+		if err != nil {
+			return err
+		}
+
+		err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
+			if event.Type == watch.Error {
+				return false, fmt.Errorf("there was an error uninstalling rancher gatekeeper chart")
+			} else if event.Type == watch.Deleted {
+				return true, nil
+			}
+			return false, nil
+		})
+		if err != nil {
+			return err
+		}
+
+		dynamicClient, err := client.GetDownStreamClusterClient(installOptions.ClusterID)
+		if err != nil {
+			return err
+		}
+
+		namespaceResource := dynamicClient.Resource(namespaces.NamespaceGroupVersionResource).Namespace("")
+
+		err = namespaceResource.Delete(context.TODO(), RancherGatekeeperNamespace, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+
+		unstructuredCRDList, err := crds.ListCustomResourceDefinitions(client, installOptions.ClusterID, "")
+		if err != nil {
+			return err
+		}
+		CRDSlice := crds.GetCustomResourceDefinitionsListByName(unstructuredCRDList, "gatekeeper")
+		err = crds.BatchDeleteCustomResourceDefinition(client, installOptions.ClusterID, "", CRDSlice)
+		if err != nil {
+			return err
+		}
+
+		adminClient, err := rancher.NewClient(client.RancherConfig.AdminToken, client.Session)
+		if err != nil {
+			return err
+		}
+		adminDynamicClient, err := adminClient.GetDownStreamClusterClient(installOptions.ClusterID)
+		if err != nil {
+			return err
+		}
+		adminNamespaceResource := adminDynamicClient.Resource(namespaces.NamespaceGroupVersionResource).Namespace("")
+
+		watchNamespaceInterface, err := adminNamespaceResource.Watch(context.TODO(), metav1.ListOptions{
+			FieldSelector:  "metadata.name=" + RancherGatekeeperNamespace,
+			TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+		})
+
+		if err != nil {
+			return err
+		}
+
+		return wait.WatchWait(watchNamespaceInterface, func(event watch.Event) (ready bool, err error) {
+			if event.Type == watch.Deleted {
+				return true, nil
+			}
+			return false, nil
+		})
+
+	})
+
+	err = catalogClient.InstallChart(chartInstallAction)
+	if err != nil {
+		return err
+	}
+
+	// wait for chart to be fully deployed
+	watchAppInterface, err := catalogClient.Apps(RancherGatekeeperNamespace).Watch(context.TODO(), metav1.ListOptions{
+		FieldSelector:  "metadata.name=" + RancherGatekeeperName,
+		TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+	})
+	if err != nil {
+		return err
+	}
+
+	err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
+		app := event.Object.(*catalogv1.App)
+
+		state := app.Status.Summary.State
+		if state == string(catalogv1.StatusDeployed) {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// newGatekeeperChartInstallAction is a helper function that returns an array of newChartInstallActions for installing the gatekeeper and gatekeepr-crd charts
+func newGatekeeperChartInstallAction(p *payloadOpts) *types.ChartInstallAction {
+	gatekeeperValues := map[string]interface{}{}
+
+	chartInstall := newChartInstall(p.Name, p.InstallOptions.Version, p.InstallOptions.ClusterID, p.InstallOptions.ClusterName, p.Host, gatekeeperValues)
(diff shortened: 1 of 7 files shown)


caliskanugur (changes requested):
Some questions and comments otherwise it looks good

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,235 @@
...
+			return err
+		}
+
+		clusterRoleResource := dynamicClient.Resource(rbacv1.SchemeGroupVersion.WithResource("clusterroles"))
+		err = clusterRoleResource.Delete(context.TODO(), "rancher-gatekeeper-crd-manager", metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+
+		clusterRoleBindingResource := dynamicClient.Resource(rbacv1.SchemeGroupVersion.WithResource("clusterrolebindings"))
```
igomez06, line 128: Same here Ugur? We don't need this clean up step right?

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,235 @@
...
+//Of note, scheme.scheme.Convert() fails because constraints have custom Kinds that Convert can't parse,
+//marshaling the list of constraints to json, then unmarshaling it into the ConstraintResponse data structure I created was the only method I found capable of getting values from deeply nested UnstructuredLists
+func ParseConstraintList(list *unstructured.UnstructuredList) *ConstraintResponse {
+	jsonConstraint, err := list.MarshalJSON()
+	if err != nil {
+		fmt.Println("error: ", err)
+	}
+
+	var parsedConstraint ConstraintResponse
+	json.Unmarshal([]byte(jsonConstraint), &parsedConstraint)
```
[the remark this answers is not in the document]
igomez06: It indeed does, you should do `err := son.Unmarshal([]byte(jsonConstraint), &parsedConstraint)` and then check the error

igomez06 (commented):
Please address our comments, but overall looks good

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,235 @@
...
+	chartInstallCRD := newChartInstall(p.Name+"-crd", p.InstallOptions.Version, p.InstallOptions.ClusterID, p.InstallOptions.ClusterName, p.Host, gatekeeperValues)
+	chartInstalls := []types.ChartInstall{*chartInstallCRD, *chartInstall}
+
+	chartInstallAction := newChartInstallAction(p.Namespace, p.ProjectID, chartInstalls)
+
+	return chartInstallAction
+}
+
+//GetUnstructuredList helper function that returns an unstructured list of data from a cluster resource
+func GetUnstructuredList(client *rancher.Client, project *management.Project, schema schema.GroupVersionResource) *unstructured.UnstructuredList {
```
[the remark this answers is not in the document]
nataliefyi: It's actually called once on its own to get the list of namespaces to compare against the list of constraint violations, so I've left it separate, but I have moved it to `tests/v2/validation/charts/gatekeeper.go` and made it private along with `parseConstraintList`

caliskanugur (commented):
A few questions and comments otherwise it looks good, excited to see the upgrade test

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,286 @@
...
+
+		namespaceResource := dynamicClient.Resource(namespaces.NamespaceGroupVersionResource).Namespace("")
+
+		err = namespaceResource.Delete(context.TODO(), RancherGatekeeperNamespace, metav1.DeleteOptions{})
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
igomez06: Why are you differentiating between a  not found and just a regular error, but treating them the same? If you are just checking for the error please just use the second error check.

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,286 @@
...
+	RancherGatekeeperNamespace = "cattle-gatekeeper-system"
+	// Name of the rancher gatekeeper chart
+	RancherGatekeeperName = "rancher-gatekeeper"
+	// Name of rancher gatekeepr crd chart
+	RancherGatekeeperCRDName = "rancher-gatekeeper-crd"
+	// namespace that is created without a label
+	RancherDisallowedNamespace = "no-label"
+)
+
+var CustomResourceDefinition = schema.GroupVersionResource{
```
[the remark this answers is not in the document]
igomez06: Completely agree Ugur, have this be its own extensions package rather than having it with charts

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,286 @@
...
+
+			if strings.Contains(CRDname, "gatekeeper") {
+				err := customResourceDefinitionResource.Delete(context.TODO(), CRDname, metav1.DeleteOptions{})
+				if err != nil {
+					return err
+				}
+
+			}
+
+		}
```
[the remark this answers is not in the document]
igomez06: Yes completely agree.

tests/v2/validation/charts/gatekeeper.go
```diff
@@ -0,0 +1,112 @@
...
+
+}
+
+func getAuditTimestamp(client *rancher.Client, project *management.Project) {
+	// sleep until the first audit finishes running.
+	// AuditTimestamp will be empty string until first audit finishes
+	// audit runs every 60 seconds, plus an arbitrary amount of time to set up the audit pod and for the audit itself
+	counter := 0
+	auditTime := ""
+	for auditTime == "" && counter < 5 {
```
igomez06, line 97: I see what you are saying now I recommend using this polling package `k8s.io/apimachinery/pkg/util/wait` and do something like this instead 

```
wait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
                auditList := getUnstructuredList(client, project, Constraint)

 		// parse it so that we can extract individual values
 		parsedAuditList := parseConstraintList(auditList)

 		// extract the timestamp of the last constraint audit
 		auditTime = parsedAuditList.Items[0].Status.AuditTimestamp
		if auditTime == "" {
                    return false, nil  
               }
		
		return true, nil
	})
```

tests/v2/validation/charts/gatekeeper.go
```diff
@@ -0,0 +1,112 @@
...
+func parseConstraintList(list *unstructured.UnstructuredList) *ConstraintResponse {
+	jsonConstraint, err := list.MarshalJSON()
+	if err != nil {
+		fmt.Println("error: ", err)
+	}
+
+	var parsedConstraint ConstraintResponse
+	err = json.Unmarshal([]byte(jsonConstraint), &parsedConstraint)
+	if err != nil {
+		fmt.Println("error: ", err)
```
igomez06, line 84: Why are you printing error and instead of returning them? We shouldn't being using `fmt.Println()` at all.

tests/v2/validation/charts/gatekeeper.go
```diff
@@ -0,0 +1,112 @@
...
+	Items      []Items
+	Kind       string
+	Metadata   interface{}
+}
+
+// GetUnstructuredList helper function that returns an unstructured list of data from a cluster resource
+func getUnstructuredList(client *rancher.Client, project *management.Project, schema schema.GroupVersionResource) *unstructured.UnstructuredList {
+	dynamicClient, err := client.GetDownStreamClusterClient(project.ClusterID)
+	if err != nil {
+		fmt.Println("error: ", err)
```
igomez06, line 61: Why are you printing error and instead of returning them? We shouldn't being using fmt.Println() at all.

tests/v2/validation/charts/gatekeeper.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	unstructured := dynamicClient.Resource(schema).Namespace("")
+
+	unstructuredList, err := unstructured.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		fmt.Println("error: ", err)
```
igomez06, line 68: Why are you printing error and instead of returning them? We shouldn't being using fmt.Println() at all.

igomez06 (commented):
It's looking good, just address comment and it'll be ready for merge

pkg/api/scheme/scheme.go
```diff
@@ -30,6 +30,7 @@ import (
 	rbacv1beta1 "k8s.io/api/rbac/v1beta1"
 	storagev1 "k8s.io/api/storage/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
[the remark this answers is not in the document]
nataliefyi: No I added this scheme, it's used in the cleanup function in `tests/framework/extensions/charts/ranchergatekeeper.go` (line 111)

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -0,0 +1,286 @@
...
+
+		customResourceDefinitionResource := dynamicClient.Resource(CustomResourceDefinition).Namespace("")
+		CRDs, err := customResourceDefinitionResource.List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+
+		for _, unstructuredCRDs := range CRDs.Items {
+			newCRD := &apiextensionsv1.CustomResourceDefinition{}
+			err := scheme.Scheme.Convert(&unstructuredCRDs, newCRD, unstructuredCRDs.GroupVersionKind())
```
igomez06, line 111: So Natalie question, is it necessary to convert your unstructured to the specific CRD? Couldn't you just use the Unstructured object to get the CRDName, and then clean it up that way? There really isn't a need to convert it. I rather not modify the scheme builder just so we can do this. Could be please do that?

tests/framework/extensions/charts/ranchergatekeeper.go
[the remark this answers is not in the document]
nataliefyi: Just to clarify, you're saying separate this block out into three functions, and put them in a package like `tests/framework/extensions/customresourcedefinitions.go` ?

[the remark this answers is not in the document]
caliskanugur: Yes Natalie, partly correct, separating this block into three functions would be great for reusability. 

The naming is mostly personal preferences, so please read below as an example instead of a todo, you can do something like this:

- Put them in a package called `customresourcedefinitions`, and folder `tests/framework/extensions/customresourcedefinitions`.

And then for each helper, my way to do this would be to put them into their files as:

- List - `tests/framework/extensions/customresourcedefinitions/list.go`
- Delete - `tests/framework/extensions/customresourcedefinitions/delete.go`
- A string check helper - `tests/framework/extensions/customresourcedefinitions/customresourcedefinitions.go`

igomez06 (commented):
Looks great. You missed one print statement. but the other changes looks. Once you remove the print statement, I'll approve it.

tests/framework/extensions/customresourcedefinitions/delete.go
```diff
@@ -0,0 +1,45 @@
...
+	dynamicClient, err := client.GetDownStreamClusterClient(clusterID)
+	if err != nil {
+		return err
+	}
+
+	customResourceDefinitionResource := dynamicClient.Resource(CustomResourceDefinitions).Namespace(namespace)
+	_, err = customResourceDefinitionResource.List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
```
caliskanugur, line 21: Is this list and error check required? It seems like the return value isn't used, could you please remove this block if this is the case?

tests/framework/extensions/customresourcedefinitions/delete.go
```diff
@@ -0,0 +1,45 @@
...
+
+// deletes a list of custom resource definitions by name
+func BatchDeleteCustomResourceDefinition(client *rancher.Client, clusterID string, namespace string, list []string) error {
+	dynamicClient, err := client.GetDownStreamClusterClient(clusterID)
+	if err != nil {
+		return err
+	}
+
+	customResourceDefinitionResource := dynamicClient.Resource(CustomResourceDefinitions).Namespace(namespace)
+	_, err = customResourceDefinitionResource.List(context.TODO(), metav1.ListOptions{})
```
caliskanugur, line 35: Same question here. Additionally, missing an error return if the lists are required.

caliskanugur (commented):
A question, but other than that looks really good

tests/framework/extensions/customresourcedefinitions/delete.go
igomez06: Yeah I have the same question what purpose does this list serve?

igomez06 (dismissed):
LGTM

caliskanugur (dismissed):
Looks good to me

tests/v2/validation/charts/k8srequiredlabels.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: constraints.gatekeeper.sh/v1beta1
```
bmdepesa, line 1: Can we move this file to a resources directory?

bmdepesa (changes requested):

tests/framework/extensions/customresourcedefinitions/delete.go
nataliefyi: They've both been removed

tests/v2/validation/charts/k8srequiredlabels.yaml
nataliefyi: Like this? `tests/v2/validation/charts/resources/k8srequiredlabels.yaml`

bmdepesa: That should be fine - I would also prefix the file name with `opa` to signify that's what it's used for.

nataliefyi: Moved it to 'tests/v2/validation/charts/resources/opa-k8srequiredlabels.yaml'

tests/framework/extensions/customresourcedefinitions/delete.go
nataliefyi: removed unnecessary error check

igomez06 (approved):
Looks good

bmdepesa (approved):

PR #38144, opened 2022-06-30
Update `rancher/shell` version to `v0.1.19`

macedogm:
Update `rancher/shell` version to `v0.1.19` and `kustomize` to `v4.5.5`.


pkg/settings/setting.go
```diff
@@ -109,7 +109,7 @@ var (
 	PartnerChartDefaultBranch           = NewSetting("partner-chart-default-branch", "main")
 	RKE2ChartDefaultBranch              = NewSetting("rke2-chart-default-branch", "main")
 	FleetDefaultWorkspaceName           = NewSetting("fleet-default-workspace-name", fleetconst.ClustersDefaultNamespace) // fleetWorkspaceName to assign to clusters with none
-	ShellImage                          = NewSetting("shell-image", "rancher/shell:v0.1.18-rc4")
+	ShellImage                          = NewSetting("shell-image", "rancher/shell:v0.1.19")
```
cbron, line 112: Tagging inside shell seems very odd, we should follow process where we do RC's leading up to a release and then tag a final release in weeks leading up to it. Also why did we go from 18-rc to 19 ? 
https://github.com/rancher/shell/tags

macedogm: My bad. The idea was to bump to a non-rc release, but I mistakenly bump to `v0.1.19` instead of `v0.1.18`. I fixed to a new `rc-5` to follow the proper process.

cbron (approved):

PR #38156, opened 2022-06-30
Update dashboard and ui tags

jakefhyde:


rancherbot (approved):

Issue #37733, opened 2022-05-13
Impersonation resources need more resilience against failed creation or accidental deletion


samjustus:
Ubuntu 
18.04 Nodes, 
Docker 19.03.6, 
RKE1, 
K8s 1.21, 
on-pem VSphere 
Rancher 2.6.3-patch3

The issue is affecting their customers, client wants to mitigate before the upgrade happens in production

samjustus:
the missing piece was the clusterrolebindings, an overloaded API could have missed the creation of these, the script mentioned checks this (dryrun by default) and optionally removes the clusterrole, which then gets recreated along with its corresponding binding ... they have an immediate ability to report on the problematic users and fix them

samjustus:
SURE-4665

sowmyav27:
Reopening for a QA Test Template

brudnak:
# :white_check_mark: PASSED

<hr>
  
## Validation Environment

| Component                 | Version / Type                |
| ------------------------- | ----------------------------- |
| Rancher version           | 2.6.7-rc3 upgraded from 2.6.6 |
| Installation option       | helm high availability        |
| RKE binary version used   | v1.3.11                       |
| If Helm Chart k8s cluster | v1.23.6                       |
| Cert Details              | let's encrypt                 |
| Helm version              | v2.16.8-rancher2              |
| Downstream cluster type   | rke1 linode                   |
| Downstream K8s version    | v1.23.7                       |
| Logged in user role       | administrator                 |
| Browser type              | google chrome                 |
| Browser version           | 103.0.5060.114                |

**ADDITIONAL SETUP**

- Helm command for Rancher **INITIAL 2.6.6** installation:

```shell
helm install rancher rancher-latest/rancher \
  --namespace cattle-system \
  --set hostname=$URL_VAR \
  --set bootstrapPassword=admin \
  --set ingress.tls.source=letsEncrypt \
  --set letsEncrypt.email=$EMAIL_VAR \
  --set letsEncrypt.ingress.class=nginx \
  --version 2.6.6
```

- Upgrade helm command:

```shell
helm upgrade rancher rancher-latest/rancher \
  --namespace cattle-system \
  --set hostname=andy.sbs \
  --set bootstrapPassword=admin \
  --set ingress.tls.source=letsEncrypt \
  --set letsEncrypt.email=brudnak@icloud.com \
  --set letsEncrypt.ingress.class=nginx \
  --version 2.6.7-rc3
```

## Validation steps

[...]

PR #38169, opened 2022-07-01
Check impersonation role bindings in setup

cmurphy:
Impersonation needs three resources: clusterroles, clusterrolebindings,
and serviceaccounts. Without this patch, we assume that if the
clusterrole exists, the clusterrolebinding also exists. In some cases,
if packets were dropped during creation or if the clusterrolebinding was
deliberately or accidentally deleted, the clusterrolebinding might not
exist and a proxy request would fail due to lack of permissions. This
patch ensures the setup is more resilient against missing resources.

https://github.com/rancher/rancher/issues/37733


MbolotSuse (approved):
LGTM

PR #38171, opened 2022-07-01
Extend workload extension helpers for the validation tests

caliskanugur:
**Extensions Updates**

- Implemented:
  - Rancher logging v2 chart installation and its options as a struct
  - Cluster name getter with the given cluster id
  - Create & list ingress and a getter to get a specific ingress with the given name
  - List namespaces, secrets, daemon sets, and deployments
- Refactored:
  - If the `extensions/resource/resources.go` has a helper (such as a getter on top of the list call) move the resource creations to `extensions/resource/create.go`.
  - Lists return <ResourceList> struct that contains a slice of the resource, and that struct can be used to add further methods such as `Names`
  - `ByName` postfix changes all getters if they use the resource name.

**Validation Tests**

- Refactored:
  - Monitoring v2 secret getter rename
  - Istio list looping since the list return a struct rather than a slice of strings

### DIFF
--- a/tests/framework/extensions/charts/charts.go
+++ b/tests/framework/extensions/charts/charts.go
@@ -56,6 +56,11 @@ type RancherMonitoringOpts struct {
 	RKEScheduler         bool
 }
 
+// RancherLoggingOpts is a struct of the required options to install Rancher Logging with desired chart values.
+type RancherLoggingOpts struct {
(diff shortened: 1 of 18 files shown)


igomez06 (approved):
LGTM

izaac (approved):
LGTM

PR #38194, opened 2022-07-05
Bump feature gate for encryption key rotation

jakefhyde:


pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -204,7 +204,7 @@ func encryptionKeyRotationSupported(releaseData *model.Release) (bool, error) {
 		if err != nil {
 			return false, fmt.Errorf("unable to parse semver version for encryption key rotation: %s", featureVersion)
 		}
-		if version.Major == 1 { // we are looking for encryption key rotation 1.x.x
+		if version.Major == 2 { // we are looking for encryption key rotation 2.x.x
```
snasovich, line 207: I think that requires a bit more detailed comment on why we're specifically looking for 2.x.x - like an explanation that 1.x.x was planned for initial support, but we've decided to never actually enable it retroactively for versions of Rancher that required 1.x.x due to critical bugs that needed to be addressed.

Initially, I was concerned it's `==` and not `>=` but I see now that if we ever bump it to `3.0.0` it means it's not backward-compatible. For future reference, we should be very careful of bumping it to `3.0.0` or higher for some later version of RKE2 as it would mean Encryption Key Rotation as it's implemented in 2.6.7 would no longer be available for these versions. There would have to be a very good reason to do this - so for any improvements to this feature on Rancher's side that require RKE2 changes, we would want to bump only minor or even patch.

snasovich (approved):
Just a minor comment to improve a comment. Otherwise LGTM. Feel free to count this approval even if (actually when) it resets on push.

snasovich (approved):

Issue #37971, opened 2022-06-10
Chart for ingress-nginx ingress resource is missing ingressClassName


maxsokolovsky:
A networking.k8s.io/v1/Ingress resource always needs the `ingressClassName` specified if the controller is nginx, otherwise, a new ingress is ignored or the ingress controller pod fails to start. This is relevant for K8s v1.22+.
`"Ignoring ingress because of error while validating ingress class" ingress="cattle-system/rancher" error="ingress does not contain a valid IngressClass"`

brudnak:
Temporary Rancher QA workaround if you come across this issue:

Run `kubectl edit ingress rancher  -n cattle-system` and add a line under spec for `ingressClassName: nginx`

```yaml
spec:
  ingressClassName: nginx
```

maxsokolovsky:
## QA Testing

### Root cause
Rancher may be unavailable to external requesters if the ingress controller
does not validate Rancher's ingress resource when it does not define `ingressClassName` in its definition.
This has been detected with the nginx ingress controller of a version greater than 4.x.

Users whose Rancher instances are backed by RKE clusters do not face this issue because
RKE, when setting up the local cluster, takes care to instruct the nginx ingress controller
to automatically insert the `ingressClassName` with a value of `nginx` if it's missing.

This does not happen with EKS or GKE clusters, so Rancher instances backed by those providers, and taking
advantage of the nginx ingress controller, may face this issue.

### What was fixed, or what changes have occurred
We expose the `ingress.ingressClassName` chart value. By default, its value is an empty string because
we do not make assumptions about the type of an ingress controller that runs in Rancher (nginx, traefik, etc.).

### Areas or cases that should be tested
1. Rancher instance backed by an RKE K8s cluster.
1. Rancher instance backed by a non-RKE K8s cluster, like EKS or GKE, for example.
Both cases need upgrade tests.

### What areas could experience regressions?

### Are the repro steps accurate/minimal?
#### RKE-based Rancher
1. Deploy an HA Rancher instance backed by RKE. Use Helm and choose a non-latest version of Rancher's image.
1. Inspect the Rancher ingress and ensure it does include the `ingressClassName` in its spec definition.
`kubectl get ingress -o yaml -n cattle-system rancher | grep ingressClassName`
1. Ensure that you can access the cluster on its hostname URL.
1. Use Helm to upgrade the Rancher release. Do not specify the `ingressClassName` value to Helm.
[...]

PR #38202, opened 2022-07-06
Add ingressClassName as an optional chart value

maxsokolovsky:
Issue: https://github.com/rancher/rancher/issues/37971

This exposes `ingress.ingressClassName` as a value in the main chart.
For Rancher setups using RKE-provisioned Kubernetes clusters, this changes nothing because RKE ensures an Nginx Ingress controller watches for ingress rules without an ingressClassName specified and sets it.

But for cases where Rancher is installed on non-RKE clusters, this change enables users to specify the class name for the relevant ingress controller (Nginx, Traefik) on the command line to Helm. This will set the value of `ingressClassName` to the name of the ingress controller on the `rancher` ingress resource.


Oats87 (approved):
Nice, welcome change that we've needed for a while.

KevinJoiner (approved):

maxsokolovsky:
@MbolotSuse, yes, I did test this scenario before opening the PR.

I provisioned an EKS cluster and installed ingress-nginx 4.0.18.

With it, Rancher v2.6.5 is not accessible because the ingress controller rejects Rancher's ingress. When I simply upgrade the Rancher release to the custom chart with a newer image and do not specify the `ingress.ingressClassName` value at all, the ingress still does not have the class name, so the controller rejects it again. Only when I upgrade and DO specify the class name, the controller accepts Rancher's ingress definition, and Rancher can be accessed.

So this works as we want. One interesting little detail is that the Rancher image version does not matter, when I upgrade to a different image, helm spins down existing Rancher pods and gradually deploys new ones. But the ingress change is visible immediately, the ingress controller properly receives the request even while the old pod is still available.

maxsokolovsky:
## QA Testing

### Root cause
Rancher may be unavailable to external requesters if the ingress controller
does not validate Rancher's ingress resource when it does not define `ingressClassName` in its definition.
This has been detected with the nginx ingress controller of a version greater than 4.x.

Users whose Rancher instances are backed by RKE clusters do not face this issue because
RKE, when setting up the local cluster, takes care to instruct the nginx ingress controller
to automatically insert the `ingressClassName` with a value of `nginx` if it's missing.

This does not happen with EKS or GKE clusters, so Rancher instances backed by those providers, and taking
advantage of the nginx ingress controller, may face this issue.

### What was fixed, or what changes have occurred
We expose the `ingress.ingressClassName` chart value. By default, its value is an empty string because
we do not make assumptions about the type of an ingress controller that runs in Rancher (nginx, traefik, etc.).

### Areas or cases that should be tested
1. Rancher instance backed by an RKE K8s cluster.
1. Rancher instance backed by a non-RKE K8s cluster, like EKS or GKE, for example.
Both cases need upgrade tests.

### What areas could experience regressions?

### Are the repro steps accurate/minimal?
#### RKE-based Rancher
1. Deploy an HA Rancher instance backed by RKE. Use Helm and choose a non-latest version of Rancher's image.
1. Inspect the Rancher ingress and ensure it does include the `ingressClassName` in its spec definition.
`kubectl get ingress -o yaml -n cattle-system rancher | grep ingressClassName`
1. Ensure that you can access the cluster on its hostname URL.
1. Use Helm to upgrade the Rancher release. Do not specify the `ingressClassName` value to Helm.
1. Ensure the ingress resource still has the class name, and Rancher is still accessible.

#### EKS-based Rancher
1. Provision an EKS Kubernetes cluster.
1. Deploy an ingress-nginx release.
```
helm upgrade --install \
  ingress-nginx ingress-nginx/ingress-nginx \
  --namespace ingress-nginx \
  --set controller.service.type=LoadBalancer \
  --version 4.0.18 \
  --create-namespace
```
1. Deploy Rancher. Use Helm and choose a non-latest version of Rancher's image.
1. Inspect the Rancher ingress and ensure it does NOT include the `ingressClassName` in its spec definition.
1. Ensure that you cannot access the cluster on its hostname URL.
1. Inspect the logs of the ingress-nginx controller pod. Search for errors and find the one that says the controller
refused to allow the ingress resource defined by Rancher, because it does not have the `ingressClassName` defined.
1. Upgrade the Rancher release, make sure the chart is of the version that includes this issue's change. Specify
the `ingress.ingressClassName` as `nginx` on the command line to Helm.
1. Ensure the ingress resource still now the class name, and Rancher is accessible.

Issue #38198, opened 2022-07-06
`couldn't ensure service account impersonator` error seen in Rancher logs


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6head `f40e542`
- Installation option (Docker install/Helm Chart): Helm

**Information about the Cluster**
- Kubernetes version: 1.24.2
- Cluster Type (Local/Downstream): local - 1.24.2 RKE1

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) : Admin

   
**Steps to reproduce:**

- RKE1 1.24 downstream cluster come sup Active, but `timeout getting credential for cluster group` seen in continuous Delivery page for this cluster 
- On cluster management page --> Cluster details --> related resources --> Cluster object is seen in Wait Applied/Error state
- Errors seen in Rancher logs:
```
2022/07/06 03:38:02 [ERROR] error syncing '_all_': handler user-controllers-controller: failed to start user controllers for cluster c-m-mfhzj6w6: ClusterUnavailable 503: cluster not found, requeuing
2022/07/06 03:38:10 [INFO] cri-dockerd is enabled for cluster version [v1.24.2-rancher1-1]
2022/07/06 03:39:12 [INFO] cri-dockerd is enabled for cluster version [v1.24.2-rancher1-1]
2022/07/06 03:39:20 [ERROR] Failed to handle tunnel request from remote address 10.42.1.4:34690 (X-Forwarded-For: 18.224.63.142): response 401: failed authentication
2022/07/06 03:39:21 [INFO] cri-dockerd is enabled for cluster version [v1.24.2-rancher1-1]
```
and
```
2022/07/06 03:35:53 [ERROR] error syncing 'p-8mdtt/creator-project-owner': handler cluster-prtb-sync: couldn't ensure service account impersonator: failed to get secret for service account: cattle-impersonation-system/cattle-impersonation-user-4n6jv, error: timed out waiting for the condition, requeuing
2022/07/06 03:35:53 [ERROR] error syncing 'c-q69rl/creator-cluster-owner': handler cluster-crtb-sync: couldn't ensure service account impersonator: failed to get secret for service account: cattle-impersonation-system/cattle-impersonation-user-4n6jv, error: timed out waiting for the condition, requeuing
```
- This rancher also had a downstream RKE2 EC2 cluster deployed.

jiaqiluo:
**Problem**:
[...]

PR #38203, opened 2022-07-07
fix the issue that waitForServiceAccount times out  

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/38198 

Problem:
The function waitForServiceAccount checks the length of Secrets on the service account to determine if the service account is ready. 
However, starting from k8s 1.24, there is no service account token secret being auto-generated, so the length is always 0. 
As a result, we see the following error in rancher logs 
```
2022/07/06 03:35:53 [ERROR] error syncing 'p-8mdtt/creator-project-owner': handler cluster-prtb-sync: couldn't ensure service account impersonator: failed to get secret for service account: cattle-impersonation-system/cattle-impersonation-user-4n6jv, error: timed out waiting for the condition, requeuing
2022/07/06 03:35:53 [ERROR] error syncing 'c-q69rl/creator-cluster-owner': handler cluster-crtb-sync: couldn't ensure service account impersonator: failed to get secret for service account: cattle-impersonation-system/cattle-impersonation-user-4n6jv, error: timed out waiting for the condition, requeuing
```

Fix:
instead of checking the length of that specific field, we now check the secret we created for the service account. 
Because we always create the secret, the new logic works regardless of the cluster version. 

Tests:
I run the local build on local cluster versions 1.23 and 1.24 and do not see the mentioned error anymore.

### DIFF
--- a/pkg/impersonation/impersonation.go
+++ b/pkg/impersonation/impersonation.go
@@ -351,7 +351,15 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
 		if err != nil {
 			return false, err
 		}
-		if len(ret.Secrets) > 0 {
+		secretName := serviceaccounttoken.ServiceAccountSecretName(sa)
+		secret, err := i.clusterContext.Core.Secrets(ImpersonationNamespace).Get(secretName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		if _, found := secret.Data[corev1.ServiceAccountTokenKey]; found {
 			return true, nil
 		}
 		return false, nil


pkg/impersonation/impersonation.go
```diff
@@ -351,7 +351,16 @@ func (i *Impersonator) waitForServiceAccount(sa *corev1.ServiceAccount) (*corev1
...
-		if len(ret.Secrets) > 0 {
+		secretName := serviceaccounttoken.ServiceAccountSecretName(sa)
+		secret, err := i.clusterContext.Core.Secrets(ImpersonationNamespace).Get(secretName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		_, found := secret.Data[corev1.ServiceAccountTokenKey]
```
Oats87, line 362:
```suggestion
		if _, found := secret.Data[corev1.ServiceAccountTokenKey]; found {
```

Oats87 (approved):
Other than a nit suggested change, LGTM

PR #38244, opened 2022-07-11
Remove deprecated blocks and update service_principal in `test_create_ha.py` test

markusewalker:
## Problem

When running the `test_create_ha.py` function in the Python legacy framework, attempting to run the function while specifying an AKS cluster failed. The failure would include the following error:

```
[WARNING]: error: b'\nError: Unsupported block type\n\n  on aks.tf line 42, in resource "azurerm_kubernetes_cluster" "default":\n  42:   role_based_access_control {\n\nBlocks of type "role_based_access_control" are not expected here.\n\nError: Unsupported block type\n\n  on aks.tf line 46, in resource "azurerm_kubernetes_cluster" "default":\n  46:   addon_profile {\n\nBlocks of type "addon_profile" are not expected here.\n'
(1, '', '\nError: Unsupported block type\n\n  on aks.tf line 42, in resource "azurerm_kubernetes_cluster" "default":\n  42:   role_based_access_control {\n\nBlocks of type "role_based_access_control" are not expected here.\n\nError: Unsupported block type\n\n  on aks.tf line 46, in resource "azurerm_kubernetes_cluster" "default":\n  46:   addon_profile {\n\nBlocks of type "addon_profile" are not expected here.\n')
```

This is because earlier this year, Terraform made the following changes:

- The block role_based_access_control has been deprecated in favor of the (optional) property role_based_access_control_enabled and block azure_active_directory_role_based_access_control.
- addon_profile.kube_dashboard is flat out removed since Kube Dashboard is not supported for K8s versions >= 1.19.
 
## Solution

The solution is to remove the deprecated blocks that are no longer relevant in the code. Additionally, the `service_principal` block needs to have the `client_id` and `client_secret` stored directly in `aks.tf` instead of `variables.tf` as before. Attempting to keep it as is causes failures when running the function with updated versions of Terraform (I used Terraform version 1.2.3 when testing).

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -7,7 +7,7 @@ from .test_import_k3s_cluster import create_multiple_control_cluster
 from .test_import_rke2_cluster import (
     RANCHER_RKE2_VERSION, create_rke2_multiple_control_cluster
 )
-from .test_rke_cluster_provisioning import rke_config
+from .test_rke_cluster_provisioning import AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, rke_config
 
 # RANCHER_HA_KUBECONFIG and RANCHER_HA_HOSTNAME are provided
 # when installing Rancher into a k3s setup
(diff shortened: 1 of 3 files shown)


markusewalker:
@sowmyav27 looks like drone build is failing. I can't restart it despite being logged in, lack of permissions maybe? Also, is there anyone else you want to take a look at this PR?

markusewalker:
@anupama2501 added you for review since you initially noted the issue :) .

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -493,7 +493,17 @@ def create_aks_cluster():
                               'client_id': client_id,
```
izaac, line 493: If we are removing these variables from the terraform files shouldn't these also be removed?

`client_id`
`client_secret`

There are placeholders now that are replaced with the client ID and secret.

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -7,7 +7,7 @@
 from .test_import_rke2_cluster import (
     RANCHER_RKE2_VERSION, create_rke2_multiple_control_cluster
 )
-from .test_rke_cluster_provisioning import rke_config
+from .test_rke_cluster_provisioning import AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, rke_config
```
izaac, line 10: We have environment variables of type Azure Resource Manager (ARM), these should have the same value as the `AZURE_CLIENT_ID` `AZURE_CLIENT_SECRET`

```python
    client_id = os.environ.get('ARM_CLIENT_ID', '')
    client_secret = os.environ.get('ARM_CLIENT_SECRET', '')
```

izaac (changes requested):
Looks good overall, just a couple of comments around variables utilization

tests/validation/tests/v3_api/test_create_ha.py
markusewalker: Would we then have to have to export `ARM_CLIENT_ID` and `ARM_CLIENT_SECRET` instead? I currently source `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET`, should that not be the case?

tests/validation/tests/v3_api/test_create_ha.py
markusewalker: I think those should be safe to remove, I'll test out locally and check.

markusewalker:
Thanks @izaac for your reviews! I updated the PR with the recommendations and am running them through Jenkins. I will send you that job offline once finished.

izaac (approved):
LGTM

slickwarren (approved):

markusewalker:
@sowmyav27 looks like this is ready to be merged whenever you have a free moment. Thanks!

sowmyav27 (approved):

Issue #38160, opened 2022-07-01
Upgrade from k8s v1.23 to v1.24 fails due to restarting kubelet. 


vivek-infracloud:
**Rancher Server Setup**
- Rancher version: v2.6.6 --->  v2.6.7-rc1
- Installation option (Docker install/Helm Chart): Docker

**Information about the Cluster**
- Kubernetes version: v1.23.7-rancher1-1
- Cluster Type (Local/Downstream): Downstream

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom).
Admin
   
**Describe the bug**

Observed this issue while testing upgrade checks for k8s v1.24.
Upgrade from k8s v1.23 to v1.24 fails with kubelet restarting and showing the following error in the log.

**To Reproduce**
1. Create a rancher server v2.6.6
2. Provision a downstream custom cluster v1.23.
3. Upgrade rancher to latest v2.6.7-rcX version.
4. Upgrade downstream cluster v1.23 to v1.24
5. Notice control-plane remains in unavailable state.

**Result**

Cluster upgrade to v1.24 fails. Cluster remains in Updating state with Control plane failed to upgrade error message. 

**Expected Result**

Cluster upgrade from k8s v1.23 to v1.24 succeeds. 

**Screenshots**

Error message on cluster nodes.
![Screenshot from 2022-07-01 13-40-51](https://user-images.githubusercontent.com/83208989/176913847-8042d29b-e685-4f47-ac7d-6fac14ab3ab9.png)

Kubelet on control plane keeps restarting. (possibly on every node, did not check all the nodes).
![Screenshot from 2022-07-01 15-53-20](https://user-images.githubusercontent.com/83208989/176913841-4c95d0c4-5c74-4003-8186-dffc00e06a1f.png)

**Additional context**

kubelet container on Control Plane keeps restarting, and the kubelet container log shows the following error.
`Error: failed to run Kubelet: unable to determine runtime API version: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix: missing address"`

sowmyav27:
This is because `enable_cri_dockerd` is set to false and it fails to provision 1.24 cluster completely. Had an offline discussion with @kinarashah @snasovich. Needs more discussion -  if we must set it to `true` by default on cluster update or not.

kinarashah:
[...]

PR #38254, opened 2022-07-12
default cridockerd to true for >=1.24 unless set by annotation

kinarashah:
**Issue**
Upgrades to 1.24 fail if cri dockerd isn't enabled on the rke config. https://github.com/rancher/rancher/issues/38160 

**Solution**
- Enable cri dockerd by default on creates and updates for >=1.24 clusters. 
- Introduce annotation on cluster `"io.cattle.cluster.cridockerd.enable"` in case user wants to override default behavior. Note that in this case cluster will break if docker runtime isn't available. We don't expect users to use this annotation unless they know what they're doing.


rosskirkpat (approved):

jiaqiluo (approved):
LGTM.

We will need to add the new annotation to the document and release note.

Issue #37549, opened 2022-04-29
rke2/k3s encryption key rotation integration tests


jakefhyde:
Create integration tests for rke2/k3s encryption key rotation which tests the following cases:

- Single controlplane node clusters
- HA clusters
- Single controlplane node clusters where the controlplane node is not the init node (i.e. no nodes with both controlplane and etcd role)
- HA clusters where no controlplane node is the init node (i.e. no nodes with both controlplane and etcd role)
- Clusters "at rest" versus clusters actively creating and manipulating secrets
- Different version of rke2/k3s (unsupported versions should fail)
- Clusters that are not in `start` or `reencrypt_finished` when encryption key rotation is requested should fail

jakefhyde:
Blocked until https://github.com/rancher/rancher/pull/38281 is merged

jakefhyde:
Blocked by https://github.com/rancher/rancher/issues/38283

markusewalker:
**QA TEST PLAN**

**Scenarios**
|Scenario|Test Case|
|---|---|
|1|As a standard user, provision a downstream RKE2 node driver cluster (3 etcd, 2 cp, 3 workers), attach a created secret to a workload, take etcd snapshot, rotate encryption keys|
|2|As a standard user, provision a downstream K3S node driver cluster (2 etcd+cp, 3 workers), attach a created secret to a workload, take etcd snapshot, rotate encryption keys|
|3|As a standard user, provision a downstream RKE2 node driver cluster (1 node, all roles), attach a created secret to a workload, take etcd snapshot, rotate encryption keys|
|4|As a standard user, provision a downstream RKE2 custom cluster (3 etcd, 2 cp, 2 Linux workers+1 Windows worker), attach a created secret to a workload, take etcd snapshot, rotate encryption keys|
|5|As a standard user, provision a downstream K3S node driver cluster (1 etcd, 1 cp, 1 worker), create 10,000 secrets, take etcd snapshot, rotate encryption keys|
|6|Setup Rancher v2.6.13, provision a downstream RKE2 node driver cluster (1 etcd, 1 cp, 1 worker), attach 10 created secrets to 10 unique workloads, upgrade Rancher to v2.7-head, take etcd snapshot, rotate encryption keys

markusewalker:
Validated this against `v2.7-head` https://github.com/rancher/rancher/commit/2e50d9f2f8725dccb40f28c8be2d8c2bb99fa59a. See details below:

ENVIRONMENT DETAILS
[...]

PR #38259, opened 2022-07-12
37549 encryption key rotation validation tests

jakefhyde:
## Issue:  

https://github.com/rancher/rancher/issues/37549 

Implemented encryption key rotation validaiton tests for both a fresh install, and a cluster with 10k secrets. Creating 10k secrets takes a long time, as does reencrypting them (>1 hour in total) so any suggestions for speeding things up are welcome.

## Testing

I've tested the following cases based of the latest 1.24 releases for rke2 & k3s, which succeeded using a custom branch I have used to fix planner thrashing. We should be okay to merge this PR beforehand however, since these tests aren't automated. 

configuration | status
-- | --
1 node all roles | ✔️
1 etcd, 1 cp, 1 worker | ✔️
3 nodes all roles | ✔️
3 etcd, 2 cp, 3 worker | ✔️
3 etcd, 3 cp, 3 worker | ✔️
3 etcd, 5 cp, 3 worker | ✔️


tests/framework/extensions/machinepools/machinepools.go
```diff
@@ -53,7 +55,25 @@ type NodeRoles struct {
...
+		result = append(result, "etcd")
+	}
+	if n.Worker {
+		result = append(result, "worker")
+	}
+
+	return fmt.Sprintf("%d %s", n.Quantity, strings.Join(result, "+"))
+}
+
+// RKEMachinePoolSetup is a helper method that will loop and setup multiple node pools with the defined node roles from the `nodeRoles` parameter
```
annablender, line 76: Looks like lines 76-92 are duplicate code. Why is it commented out? Can it be removed?

tests/v2/validation/provisioning/rke2/rke2_encryption_key_rotation_test.go
```diff
@@ -0,0 +1,190 @@
...
+	require.NoError(r.T(), err)
+
+	r.client = client
+
+	r.clusterName = r.client.RancherConfig.ClusterName
+	r.namespace = r.client.RancherConfig.ClusterName
+}
+
+func (r *RKE2EncryptionKeyRotationTestSuite) TestEncryptionKeyRotationImpl(provider Provider, kubeVersion string, nodesAndRoles []machinepools.NodeRoles, credential *cloudcredentials.CloudCredential) {
+	name := fmt.Sprintf("Provider_%s/Kubernetes_Version_%s/Nodes_%v", provider.Name, kubeVersion, nodesAndRoles)
```
annablender, line 65: Will provider ever be nil?

tests/v2/validation/provisioning/rke2/rke2_encryption_key_rotation_test.go
```diff
@@ -0,0 +1,190 @@
...
+	r.Run(name, func() {
+		testSession := session.NewSession(r.T())
+		defer testSession.Cleanup()
+
+		testSessionClient, err := r.client.WithSession(testSession)
+		require.NoError(r.T(), err)
+
+		clusterName := AppendRandomString(fmt.Sprintf("%s-%s", r.clusterName, provider.Name))
+		generatedPoolName := fmt.Sprintf("nc-%s-pool1-", clusterName)
+		machinePoolConfig := provider.MachinePoolFunc(generatedPoolName, namespace)
```
annablender, line 75: Namespace is never passed into the func. It's a field of the `RKE2EncryptionKeyROtationTestSuite` struct, shouldn't it be `r.namespace?` since I'm assuming you've smoke tested this and it works, I'm probably missing something about Go here

jakefhyde: No you're right it probably needs to be updated to `r.namespace`.

tests/framework/extensions/machinepools/machinepools.go
jakefhyde: It's a comment explaining how it works. Although the comment does seem out of date, I'll update.

tests/v2/validation/provisioning/rke2/rke2_encryption_key_rotation_test.go
jakefhyde: I lied, `namespace` is defined in `tests/v2/validation/provisioning/rke2/config.go` as `fleet-default`, which makes sense.

tests/v2/validation/provisioning/rke2/rke2_encryption_key_rotation_test.go
```diff
@@ -100,35 +104,63 @@ func (r *RKE2EncryptionKeyRotationTestSuite) TestEncryptionKeyRotationImpl(provi
...
+		for i := 0; i < 1000; i++ {
+			s := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: fmt.Sprintf("encryption-key-rotation-test-%d-", i),
+				},
+				Data: map[string][]byte{
+					"key": []byte(namegenerator.RandStringLower(5)),
+				},
+			}
+			_, err = secrets.CreateSecret(r.client, s, clusterName, "default")
```
jakefhyde, line 124: @igomez06 Any idea why this sin't working? It always errors with "the server could not find the requested resource"

igomez06: It's because you are passing the cluster name rather than the clusterID

pkg/codegen/main.go
```diff
@@ -106,7 +106,8 @@ func main() {
 				Types: []interface{}{
 					"./pkg/apis/rke.cattle.io/v1",
 				},
-				GenerateTypes: true,
+				GenerateTypes:   true,
+				GenerateClients: true,
```
igomez06, line 110: This is actually generating the k8's rke client not the rancher one. So this client would be hitting the Kubernetes api endpoint rather than the rancher. One of the last thing's Jacob had suggested was to use generate the client similar to how the management.cattle.io clients are generated using the Norman code generator. I did it for the provisioning.cattle.io client in this PR.

https://github.com/rancher/rancher/pull/38281

igomez06:
I think this PR: https://github.com/rancher/rancher/pull/38281, needs to be approved and merged before merging this one. As the provisioning client has been updated.

tests/v2/validation/provisioning/rke2/rke2_encryption_key_rotation_test.go
jakefhyde: Oh, duh

tests/framework/clients/rancher/rke/client.go
```diff
@@ -0,0 +1,27 @@
...
+type Client struct {
+	rkeClientV1.RkeV1Interface
+	ts *session.Session
+}
+
+
+// NewForConfig creates a new ProvisioningV1Client for the given config. It also takes session.Session as parameter to track the resources
+// the ProvisioningV1Client creates
+func NewForConfig(c *rest.Config, ts *session.Session) (*Client, error) {
+	rkeClient, err := rkeClientV1.NewForConfig(c)
```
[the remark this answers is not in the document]
jakefhyde: AFAIK no, I'm only using it to watch a status field on the controlplane.

tests/framework/extensions/machinepools/machinepools.go
jakefhyde: I have, as a side note I like the `machinepools.NodeRools` mechanism it's really easy to work with and it's really easy to structure the json for in `CATTLE_TEST_CONFIG`.

jakefhyde:
This PR depends on https://github.com/rancher/rancher/pull/38796, so you may see some changes to `encryptionkeyrotation.go`, for the purposes of this PR just ignore them.

igomez06 (commented):
So my question is Jake why are you using the kubeapi to create these v1 resources rather than the Steve client? Since this in validation tests, we want to purely mimic the UI ie. use the Steve client to create v1 resources, not the controllers. Can you see how we did in the cert_rotation test in tests/v2/validation/provisioning/rke2?

jakefhyde:
@igomez06 We are using the kubeapi to create secrets as part of the tests, because the creation of secrets is required to test whether or not encryption key rotation works with clusters with large numbers of secrets, the creation of these resources is not being tested here. We are still using steve to create/update the provisioning cluster.

igomez06:
> @igomez06 We are using the kubeapi to create secrets as part of the tests, because the creation of secrets is required to test whether or not encryption key rotation works with clusters with large numbers of secrets, the creation of these resources is not being tested here. We are still using steve to create/update the provisioning cluster.

@jakefhyde Ah ok, so let me make sure I understand. The 10k secrets is a setup/perquisite of the main test being that  encryption key rotation works with clusters. The secrets part make sense, but you using the controller for updating the provisioned cluster that I pointed out. Shouldn't we use the Steve client to update the cluster, as that is part of the test?

jakefhyde:
@igomez06 Updated, thanks for being patient with this review, it's taking a bit longer than expected but the issues with encryption key rotation for a large # of secrets have been determined so PRs are going to go up soon.

jakefhyde:
Still debugging timeouts

tests/framework/extensions/kubeapi/secrets/create.go
```diff
@@ -8,18 +8,22 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
 )
 
-// CreateSecret is a helper function that uses the dynamic client to create a secret on a namespace for a specific cluster.
-// It registers a delete function.
-func CreateSecret(client *rancher.Client, secret *corev1.Secret, clusterID, namespace string) (*corev1.Secret, error) {
+// CreateSecretForCluster is a helper function that uses the rancher client to create a secret in a namespace for a specific cluster.
+func CreateSecretForCluster(client *rancher.Client, clusterID, namespace string, secret *corev1.Secret) (*corev1.Secret, error) {
```
igomez06, line 15: What's the difference between this func, and other `CreateSecret` func?

igomez06 (commented):
Looks great Jake just had one question

tests/framework/extensions/kubeapi/secrets/create.go
jakefhyde: The difference being that this function just wraps the functionality for creating the dynamic client for the cluster, so that in my encryption key rotation tests I don't have to recreate the client in the creation loop. I don't know if it creates a measurable performance impact, but it takes like an hour to generate 10k secrets so figured it was probably better to do it this way.

igomez06 (approved):
Looks good. We're holding off on test merges until our HA upgrade PR goes in. This requires v2.6 backport correct?

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
```diff
@@ -0,0 +1,270 @@
...
+type RKE2EncryptionKeyRotationTestSuite struct {
+	suite.Suite
+	session     *session.Session
+	client      *rancher.Client
+	config      *provisioning.Config
+	clusterName string
+	namespace   string
+}
+
+const scale = 10000
```
bfbachmann, line 41: Sorry, I lack context here so I'm just curious: why 10k?

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
```diff
@@ -0,0 +1,270 @@
...
+			r.T().Logf("Successfully created %d secrets in namespace default for encryption key rotation for %s", scale, name)
+			// encryption key rotation is capped at 5 secrets per second (10 every 2 seconds), so 10000 secrets will take
+			// 2000 seconds which is ~33 minutes.
+			require.NoError(r.T(), r.rotateEncryptionKeys(clusterResp.ID, 2, 60*60))
+			r.T().Logf("Successfully completed second encryption key rotation for %s", name)
+		})
+	})
+}
+
+func (r *RKE2EncryptionKeyRotationTestSuite) rotateEncryptionKeys(id string, generation, timeout int64) error {
```
bfbachmann, line 148: I see we're passing `timeout` here, but how come you're only using it when watching `RKEControlPlanes` and using `defaults.WatchTimeoutSeconds` when watching `Clusters`? 

Also, this is probably a nitpick, but should `timeout` here be of type `time.Duration` rather than `int64`?

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
```diff
@@ -0,0 +1,270 @@
...
+			kubeProvisioningClient, err := r.client.GetKubeAPIProvisioningClient()
+			require.NoError(r.T(), err)
+
+			result, err := kubeProvisioningClient.Clusters(namespace).Watch(context.TODO(), metav1.ListOptions{
+				FieldSelector:  "metadata.name=" + cluster.ObjectMeta.Name,
+				TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+			})
+			require.NoError(r.T(), err)
+
+			checkFunc := clusters.IsProvisioningClusterReady
```
bfbachmann, line 110: Nit: You're creating this variable but it's only used in one place, which happens to be on the next line.

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
```diff
@@ -0,0 +1,270 @@
...
+	r.client = client
+
+	r.clusterName = r.client.RancherConfig.ClusterName
+	r.namespace = r.client.RancherConfig.ClusterName
+}
+
+func (r *RKE2EncryptionKeyRotationTestSuite) TestEncryptionKeyRotationFreshCluster(provider Provider, kubeVersion string, nodesAndRoles []machinepools.NodeRoles, credential *cloudcredentials.CloudCredential) {
+	name := fmt.Sprintf("Provider_%s/Kubernetes_Version_%s/Nodes_%v", provider.Name, kubeVersion, nodesAndRoles)
+	r.Run(name, func() {
+		r.Run("initial", func() {
```
bfbachmann, line 76: Sorry if this is a stupid question, but what does this `initial` test label refer to? Are we planning on creating more test sub-suites alongside this `initial` one?

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
jakefhyde: context: https://github.com/rancher/rancher/issues/40202

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
jakefhyde: From apimachinery:
```
	TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty" protobuf:"varint,5,opt,name=timeoutSeconds"`
```

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
jakefhyde: I was originally going to separate the provisioning and the rotation, but scrapped that idea, so this "initial" test case can be unwrapped, good catch.

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
jakefhyde: Just copying the logic here: https://github.com/rancher/rancher/blob/10f2cc093c4864a694981338c034acfdd20b6548/tests/v2/validation/provisioning/rke2/custom_cluster.go#L81-L82

Now that I think about it, your comment is probably the more correct way to do it.

tests/framework/extensions/kubeapi/resource.go
```diff
@@ -0,0 +1,16 @@
+package kubeapi
+
+import (
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+func ResourceForClient(client *rancher.Client, clusterName, namespace string, resource schema.GroupVersionResource) (dynamic.ResourceInterface, error) {
```
igomez06, line 9: Can you add a comment here for our generated go docs?

tests/v2/validation/provisioning/k3s/encryption_key_rotation_test.go
```diff
@@ -0,0 +1,246 @@
...
+			provider := CreateProvider(providerName)
+
+			client, err := r.client.WithSession(subSession)
+			require.NoError(r.T(), err)
+
+			cloudCredential, err := provider.CloudCredFunc(client)
+			require.NoError(r.T(), err)
+
+			// provisioning is not considered part of the test
+			id := provisionEnvironment(r.T(), client, r.clusterName, provider, kubernetesVersion, r.config.NodesAndRoles, cloudCredential, r.config.PSACT, r.config.AdvancedOptions)
```
igomez06, line 187: If provisioning is not considered a part of the test would it make sense to have it in its own test package and receive the cluster ID like we do in other tests?
ex) https://github.com/rancher/rancher/blob/a05de31fccb10059447c169f28dcc2068982a6f0/tests/v2/validation/charts/gatekeeper_test.go#L42-L47

igomez06: This way you don't have to duplicate test for k3s as well.

tests/v2/validation/provisioning/rke2/encryption_key_rotation_test.go
```diff
@@ -0,0 +1,245 @@
...
+			provider := CreateProvider(providerName)
+
+			client, err := r.client.WithSession(subSession)
+			require.NoError(r.T(), err)
+
+			cloudCredential, err := provider.CloudCredFunc(client)
+			require.NoError(r.T(), err)
+
+			// provisioning is not considered part of the test
+			id := provisionEnvironment(r.T(), client, r.clusterName, provider, kubernetesVersion, r.config.NodesAndRoles, cloudCredential, r.config.PSACT, r.config.AdvancedOptions)
```
igomez06, line 186: same here

igomez06 (commented):
Looks great, would probably make sense to move this test to its own package, since it's not a provisioning test.

igomez06 (approved):

bfbachmann (approved):

PR #38297, opened 2022-07-15
Update k3s hardened install to be in line with docs

rancher-max:
## Issue
N/A
 
## Problem
Automation for K3s hardened installs did not include audit policy. Also, a recent change to the k3s codebase changed the label for required network policies, so it needs to be updated to allow svclb to work in hardened environments.
 
## Solution

- Update the network policy applied to k3s in automated tests
- Include audit policy file and args required for the check
 
## Testing

I've run this in my local environment to ensure the test runs and works as expected


ShylajaDevadiga (approved):
LGTM

PR #38325, opened 2022-07-19
Adds decompression and error handling to the audit logger.

KevinJoiner:
## Problem
The audit logger silently fails when attempting to marshal gzip data to JSON.
 
## Solution
The audit logger should decode encoded response messages and log errors to Rancher.

### DIFF
--- a/pkg/auth/audit/audit.go
+++ b/pkg/auth/audit/audit.go
@@ -1,9 +1,14 @@
+// Package audit is used to preform audit logging.
 package audit
 
 import (
 	"bytes"
+	"compress/gzip"
+	"compress/zlib"
 	"context"
 	"encoding/json"
+	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"regexp"
@@ -11,7 +16,6 @@ import (
 	"time"
 
 	"github.com/pborman/uuid"
-	"github.com/pkg/errors"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
 	k8stypes "k8s.io/apimachinery/pkg/types"
@@ -19,15 +23,24 @@ import (
 )
 
 const (
-	contentTypeJSON = "application/json"
-	redacted        = "[redacted]"
+	contentTypeJSON     = "application/json"
+	contentEncodingGZIP = "gzip"
+	contentEncodingZLib = "deflate"
+	redacted            = "[redacted]"
 )
 
+// Level represents a desired logging level.
+type Level int
+
 const (
-	levelNull = iota
-	levelMetadata
-	levelRequest
-	levelRequestResponse
+	// LevelNull default value.
+	LevelNull Level = iota
+	// LevelMetadata log request header information.
+	LevelMetadata
+	// LevelRequest log metadata and request body.
+	LevelRequest
+	// LevelRequestResponse log metadata request body and response header and body.
+	LevelRequestResponse
 )
 
 var (
@@ -37,6 +50,8 @@ var (
 	}
 	sensitiveRequestHeader  = []string{"Cookie", "Authorization"}
 	sensitiveResponseHeader = []string{"Cookie", "Set-Cookie"}
+	// ErrUnsupportedEncoding is returned when the response encoding is unsupported
+	ErrUnsupportedEncoding = fmt.Errorf("unsupported encoding")
 )
 
 type auditLog struct {
@@ -64,6 +79,7 @@ type log struct {
 
 var userKey struct{}
 
+// User holds information about the user who caused the audit log
 type User struct {
 	Name  string              `json:"name,omitempty"`
 	Group []string            `json:"group,omitempty"`
@@ -93,6 +109,7 @@ func getUserNameForBasicLogin(body []byte) string {
 	return input.Username
 }
 
+// FromContext gets the user information from the given context.
 func FromContext(ctx context.Context) (*User, bool) {
 	u, ok := ctx.Value(userKey).(*User)
 	return u, ok
@@ -113,7 +130,7 @@ func newAuditLog(writer *LogWriter, req *http.Request, keysToConcealRegex *regex
 
 	contentType := req.Header.Get("Content-Type")
 	loginReq := isLoginRequest(req.RequestURI)
-	if writer.Level >= levelRequest || loginReq {
+	if writer.Level >= LevelRequest || loginReq {
 		if bodyMethods[req.Method] && strings.HasPrefix(contentType, contentTypeJSON) {
 			reqBody, err := readBodyWithoutLosingContent(req)
 			if err != nil {
@@ -125,7 +142,7 @@ func newAuditLog(writer *LogWriter, req *http.Request, keysToConcealRegex *regex
 					auditLog.log.UserLoginName = loginName
 				}
 			}
-			if writer.Level >= levelRequest {
+			if writer.Level >= LevelRequest {
 				auditLog.reqBody = reqBody
 			}
 		}
@@ -139,6 +156,7 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
 	a.log.RequestHeader = filterOutHeaders(reqHeaders, sensitiveRequestHeader)
 	a.log.ResponseHeader = filterOutHeaders(resHeaders, sensitiveResponseHeader)
 	a.log.ResponseCode = resCode
+
 	if a.log.UserLoginName != "" {
 		if a.log.User.Extra == nil {
 			a.log.User.Extra = make(map[string][]string)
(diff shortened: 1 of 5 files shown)


pkg/auth/audit/audit.go
```diff
@@ -148,31 +164,72 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
 	}
 
 	var buffer bytes.Buffer
+
 	alByte, err := json.Marshal(a.log)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to marashal log message: %w", err)
```
cmurphy, line 170:
```suggestion
		return fmt.Errorf("failed to marshal log message: %w", err)
```

pkg/auth/audit/audit.go
```diff
@@ -148,31 +164,72 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
...
+		return nil
+	}
+
+	switch resHeaders.Get("Content-Encoding") {
+	case contentEncodingGZIP:
+		resBody, err = decompressGZIP(resBody)
+	case contentEncodingZLib:
+		resBody, err = decompressZLip(resBody)
+	case "":
+		// do nothing message is  not encoded
```
cmurphy, line 220:
```suggestion
		// do nothing message is not encoded
```

pkg/auth/audit/audit.go
```diff
@@ -266,3 +323,35 @@ func (a *auditLog) concealMap(m map[string]interface{}) bool {
...
+
+func decompress(readCloser io.ReadCloser) ([]byte, error) {
+	rawData, err := ioutil.ReadAll(readCloser)
+	if err != nil {
+		_ = readCloser.Close()
+		return nil, fmt.Errorf("failed to read compressed response: %w", err)
+	}
+
+	if err = readCloser.Close(); err != nil {
+		return rawData, fmt.Errorf("failed to closed reader: %w", err)
```
cmurphy, line 353:
```suggestion
		return rawData, fmt.Errorf("failed to close reader: %w", err)
```

pkg/auth/audit/filter.go
```diff
@@ -65,14 +73,23 @@ func (h auditHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 	h.next.ServeHTTP(wr, req)
 
-	auditLog.write(user, req.Header, wr.Header(), wr.statusCode, wr.buf.Bytes())
+	err = auditLog.write(user, req.Header, wr.Header(), wr.statusCode, wr.buf.Bytes())
+	if err == nil {
+		return
+	}
+	h.errLock.Lock()
+	defer h.errLock.Unlock()
+	if lastSeen, ok := h.errMap[err.Error()]; !ok || time.Since(lastSeen) > errorDebounceTime {
```
cmurphy, line 89: Can you add a comment for why this is needed instead of just logging the error directly?

pkg/auth/audit/audit.go
```diff
@@ -1,33 +1,46 @@
...
+// Level represents a desired logging level
+type Level int
+
 const (
-	levelNull = iota
-	levelMetadata
-	levelRequest
-	levelRequestResponse
+	// LevelNull default value
+	LevelNull Level = iota
```
cmurphy, line 37: Do these need to be exported? I'm not seeing where they are used outside of this package

pkg/auth/audit/audit.go
```diff
@@ -148,31 +164,72 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
...
+func (a *auditLog) writeResponse(buf *bytes.Buffer, resHeaders http.Header, resBody []byte) (err error) {
+	if a.writer.Level < LevelRequestResponse || resHeaders.Get("Content-Type") != contentTypeJSON || len(resBody) == 0 {
+		return nil
+	}
+
+	switch resHeaders.Get("Content-Encoding") {
+	case contentEncodingGZIP:
+		resBody, err = decompressGZIP(resBody)
+	case contentEncodingZLib:
+		resBody, err = decompressZLip(resBody)
```
cmurphy, line 218: When is this type of compression used? I see usage of gzip but not zlip in the k8s source code

KevinJoiner: I didn't see it in the source code, but the request header has `"Accept-Encoding":["gzip, deflate, br"]`. Which means are client should be able to handle it as a valid `Content-Encoding`. Since `zlib` is in the standard library, implementing it along with `gzip` is trivial and protects us from running into the same issue if someone decides to use the `deflate` encoding. Though `br` is also in the list I chose not to add code to handle `Brotli` since it is not widely used or in the standard library. [Golang Brotli proposal ](https://github.com/golang/go/issues/44444).

pkg/auth/audit/audit.go
KevinJoiner: I changed the function signatures of `func NewLogWriter(path string, level Level, maxAge, maxBackup, maxSize int)` to use `audit.Level` instead of an arbitrary `int`. This does not change the functionality of anything, but it seemed odd that to create a new `LogWriter` you have to look into the package source code to know what each int value does.

pkg/auth/audit/filter.go
KevinJoiner: Yes, though I am curious about your thoughts around debouncing the error log messages, such as if it is unneeded complexity. The goal is to not spam rancher with failed log messages if the user does something that will cause all logs to fail, such as providing an invalid log path. I think it is nice to have, but not completely necessary. I also just picked an arbitrary debounce time of 30 seconds.

pkg/auth/audit/audit.go
KevinJoiner: Eagle Eyes  🦅

cmurphy (dismissed):

KevinJoiner:
Added unit test for decompression of response message.

pkg/auth/audit/audit_test.go
```diff
@@ -126,20 +137,178 @@ func Test_concealSensitiveData(t *testing.T) {
...
+			a.NoError(err, "failed to unmarshal")
+			got := logger.concealSensitiveData(test.uri, test.input)
 			var gotMap map[string]interface{}
-			if err := json.Unmarshal(got, &gotMap); err != nil {
-				t.Errorf("error unmarshaling: %v", err)
-			}
-			if !reflect.DeepEqual(want, gotMap) {
-				t.Errorf("concealSensitiveData() = %s, want %s", got, tt.want)
+			err = json.Unmarshal(got, &gotMap)
+			a.NoError(err, "failed tto unmarshal")
```
cmurphy, line 149: (nit)

```suggestion
			a.NoError(err, "failed to unmarshal")
```

pkg/auth/audit/audit_test.go
```diff
@@ -126,20 +137,178 @@ func Test_concealSensitiveData(t *testing.T) {
...
+	for i := range tests {
+		test := tests[i]
+		a.Run(test.name, func() {
+			// write the test to the audit logger
+			err := auditLog.write(nil, req.Header, test.respHeader, test.returnCode, test.respBody)
+
+			// if we are expecting an error check the error is not nil and the correct type
+			if test.Error != nil {
+				a.Require().Error(err, "Expected error of type '%T'.", test.Error)
+				a.ErrorAsf(test.Error, &err, "Expected error of type '%T'.", test.Error)
```
cmurphy, line 256: Aren't these basically the same thing? Do we need both? If they serve different purposes, can we use a different error message for one?

pkg/auth/audit/audit_test.go
```diff
@@ -126,20 +137,178 @@ func Test_concealSensitiveData(t *testing.T) {
...
+		respBody     []byte
+		returnCode   int
+		expectedBody string
+		Error        error
+	}{
+		{
+			name:       "invalid Encoding",
+			respHeader: http.Header{"Content-Type": []string{"application/json"}, "Content-Encoding": []string{"bzip2"}},
+			respBody:   []byte(testString),
+			Error:      fmt.Errorf(""), // any error
```
cmurphy, line 194: There would be a specific error in this case, right? Why not go ahead and test for that?

KevinJoiner: The Error didn't have a type since we created it with this line [`err = fmt.Errorf("unknown content-encoding '%s' for response", resHeaders.Get("Content-Encoding"))`](https://github.com/rancher/rancher/blob/c79aae4abcaa6fb8c5ea20007585b03010037cde/pkg/auth/audit/audit.go#L224). Though I can update it to be a exported error and check for that.

MbolotSuse (changes requested):
Couple of small comments

pkg/auth/audit/audit.go
```diff
@@ -266,3 +323,35 @@ func (a *auditLog) concealMap(m map[string]interface{}) bool {
...
+		return nil, fmt.Errorf("failed to create zlib reader: %w", err)
+	}
+
+	return decompress(zr)
+}
+
+func decompress(readCloser io.ReadCloser) ([]byte, error) {
+	rawData, err := ioutil.ReadAll(readCloser)
+	if err != nil {
+		_ = readCloser.Close()
```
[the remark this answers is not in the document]
KevinJoiner: I realized you can not wrap two errors into the same error, but I can add the close error to the returned error.

pkg/auth/audit/audit.go
```diff
@@ -148,31 +166,74 @@ func (a *auditLog) write(userInfo *User, reqHeaders, resHeaders http.Header, res
...
+	buf.Write(bytes.TrimSuffix(a.concealSensitiveData(a.log.RequestURI, a.reqBody), []byte("\n")))
+}
+
+// writeResponse attempt to write the API response to the log message.
+func (a *auditLog) writeResponse(buf *bytes.Buffer, resHeaders http.Header, resBody []byte) (err error) {
+	if a.writer.Level < LevelRequestResponse || resHeaders.Get("Content-Type") != contentTypeJSON || len(resBody) == 0 {
+		return nil
+	}
+
+	switch resHeaders.Get("Content-Encoding") {
```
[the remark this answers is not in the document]
KevinJoiner: After going through our rancher code, apiserve handles the `content-encoding` and only uses `gzip`. We use a zip when returning key files for nodes, but this does not have the `Content-Type` of `application/json` so we do not need to handle the response. Zip would also be odd to handle since it is used for archiving multiple files.

cmurphy (dismissed):

pkg/auth/audit/filter.go
```diff
@@ -65,14 +73,29 @@ func (h auditHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 	wr := &wrapWriter{ResponseWriter: rw, auditWriter: h.auditWriter, statusCode: http.StatusOK}
 	h.next.ServeHTTP(wr, req)
 
-	auditLog.write(user, req.Header, wr.Header(), wr.statusCode, wr.buf.Bytes())
+	err = auditLog.write(user, req.Header, wr.Header(), wr.statusCode, wr.buf.Bytes())
+	if err == nil {
+		return
+	}
+
+	// Locking after next is called to avoid performance hits on the request.
```
MbolotSuse, line 88: This comment doesn't make sense to me - next isn't called below that I can see.

pkg/auth/audit/filter.go
```diff
@@ -65,14 +73,29 @@ func (h auditHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	h.errLock.Lock()
+	defer h.errLock.Unlock()
+
+	// Only log duplicate error messages at most every errorDebounceTime.
+	// This is to prevent the rancher logs from being flooded with error messages
+	// when the log path is invalid or any other error that will always cause a write to fail.
+	if lastSeen, ok := h.errMap[err.Error()]; !ok || time.Since(lastSeen) > errorDebounceTime {
```
MbolotSuse, line 95: I'm not a huge fan of this logic. IMO, it has the following disadvantages:
- You are maintaining an ever growing map of errors. So if we have a lot of slightly different errors, your map will keep getting bigger.
- The map is never pruned, even when the lastSeen time exceeds the debounce time. So this map will sit in memory and continue to grow over time.

From an overall perspective, I also don't think the issue here needs to be solved. It may be annoying when errors flood the logs, but it ends up being very clear about the fact that there is an error and that you need to fix it. In my opinion, the slight annoyance caused by this doesn't necessitate the overhead of us creating/maintaining a solution. 

But that's just me. I don't think that this is vital enough to hold up your PR (though I do think we will want to prune that map eventually - I could see that causing a non-0 memory uptick), but I did want to bring it up.

pkg/auth/audit/audit_test.go
```diff
@@ -126,20 +139,188 @@ func Test_concealSensitiveData(t *testing.T) {
...
-			}
-			if !reflect.DeepEqual(want, gotMap) {
-				t.Errorf("concealSensitiveData() = %s, want %s", got, tt.want)
+			err = json.Unmarshal(got, &gotMap)
+			a.NoError(err, "failed to unmarshal")
+			a.Equal(want, gotMap, "concealSensitiveData() = %s, want %s", got, test.want)
+		})
+	}
+}
+func (a *AuditTest) TestCompression() {
```
MbolotSuse, line 156: These tests are well-written. However, there doesn't seem to be any consideration for log-level. Given that you did change around the boolean logic, I'd like to see that somewhere if you don't mind.

MbolotSuse (changes requested):
Have a couple of minor concerns:
- Don't love debounce stuff
- One comment that I don't understand
- No tests for log level

Of these, I'm only really concerned about the log level tests. The boolean logic did switch around a bit here, and I would like some tests to make sure that works as expected. Though I'm open to a discussion if you feel differently.

pkg/auth/audit/filter.go
MbolotSuse: Never mind - misread the comment. Is a bit weird that next is called and then we lock w/ comment ~8 lines later, but that's not a big deal, and the misread was more on my failure to read carefully.

MbolotSuse (approved):
LGTM - might be one missing test case, but that isn't critical.

Issue #38215, opened 2022-07-07
Cannot update s3 credentials after cluster has been created.


jakefhyde:
**Rancher Server Setup**
- Rancher version: v2.6.4 - v2.6.7

**Information about the Cluster**
- Cluster Type (Local/Downstream): Downstream
  - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE1

**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**

After the creation of the cluster without setting S3 credentials to take snapshots of the downstream cluster, it's not possible to update the cluster to upload snapshots to S3 buckets. The S3 settings only work when it's set during the creation of the cluster. 

**To Reproduce**
 
- Create a new cluster without setting S3 snapshots;
- After the cluster it's created set the S3 snapshot and try to run a snapshot.
**Result**
Not able to create backups of the downstream cluster and send the backup to S3. Not able to add, update, or remove the secrets.
**Expected Result**

Can add, edit & remove S3 credential in order to enable or disable s3 snapshot capabilities for a cluster. If the cluster was created with S3 credentials, then it is possible to update the secret, but it is much harder to get around if the cluster was created without it.
**Screenshots**

I believe this issue is caused here:

https://github.com/rancher/norman/blob/61b018639b217b33604cae675e189c2afa4cf378/types/convert/merge/merge.go#L11-L20

which is called in `rancher` here:

https://github.com/rancher/rancher/blob/b919539b4869325aadcdb737070c02914a692cb2/pkg/api/norman/store/cluster/cluster_store.go#L841

Before:
![image](https://user-images.githubusercontent.com/33796120/177827299-f7d25ab9-9edd-4173-8f01-eb908d27d3ea.png)

(Service Account Token was purposely spoofed here)

After:
```
value = {interface{} | map[string]interface{}} 
 0 = agentImage -> rancher/rancher-agent:v2.6-head
 1 = aksStatus -> 
 2 = capacity -> 
 3 = gkeStatus -> 
 4 = nodeCount -> 1
 5 = authImage -> rancher/kube-api-auth:v0.1.8
 6 = driver -> rancherKubernetesEngine
 7 = nodeVersion -> 1
 8 = requested -> 
 9 = apiEndpoint -> https://161.35.143.159:6443/
 10 = appliedPodSecurityPolicyTemplateId -> 
[...]

PR #38340, opened 2022-07-19
Move cluster and catalog secrets to Spec

cmurphy:
Norman assumes that no updates to Status happen through the API,
including through the store, and so overwrites any changes[1]. This
means it is impossible to set a new value on Status for clusters or
catalogs upon update, it can only be done on create. This was resulting
in references to secrets being lost.

This change addresses the issue by moving secret values from Status to
Spec for clusters and catalogs.

Implementation notes:
- Notifiers are unaffected because the secret reference is the only
  value in status, since there are no status values in the original
  object it is not overwritten.
- SourceCodeProviderConfigs are unaffected because the secret reference
  was not in status.
- ClusterTemplateRevisions are unaffected because they are never
  updated, only new versions are created.
- The service account token is never updated through the norman store
  and so is unaffected.
- To avoid a complicated migration, unaffected resources are left alone
  even though it introduces inconsistency between the different types of
  resources.
- A getter is added for clusters and catalogs to check both status and
  spec for the secret values. The alternative is to add a migration in a
  controller.

[1] https://github.com/rancher/norman/blob/e21cabdc65b9f556c5cba1a5f2c49297850bdc46/types/convert/merge/merge.go#L13-L15

https://github.com/rancher/rancher/issues/38215

### DIFF
--- a/pkg/api/norman/store/catalog/store.go
+++ b/pkg/api/norman/store/catalog/store.go
@@ -1,11 +1,13 @@
 package catalog
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/values"
 	c "github.com/rancher/rancher/pkg/api/norman/customization/catalog"
 	gaccess "github.com/rancher/rancher/pkg/api/norman/customization/globalnamespaceaccess"
 	"github.com/rancher/rancher/pkg/controllers/management/secretmigrator"
@@ -13,30 +15,53 @@ import (
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/ref"
 	"github.com/rancher/rancher/pkg/settings"
+	"github.com/rancher/rancher/pkg/types/config"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sTypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
 )
 
-const catalogSecretKey = "credentialSecret"
+const (
+	catalogSecrets   = "catalogSecrets"
+	catalogSecretKey = "credentialSecret"
+)
 
 type Store struct {
 	types.Store
-	Users          v3.UserInterface
-	GrbLister      v3.GlobalRoleBindingLister
-	GrLister       v3.GlobalRoleLister
-	secretMigrator *secretmigrator.Migrator
-	clusterLister  v3.ClusterLister
+	Users                v3.UserInterface
+	GrbLister            v3.GlobalRoleBindingLister
+	GrLister             v3.GlobalRoleLister
+	secretMigrator       *secretmigrator.Migrator
+	clusterLister        v3.ClusterLister
+	catalogClient        dynamic.ResourceInterface
+	clusterCatalogClient dynamic.NamespaceableResourceInterface
+	projectCatalogClient dynamic.NamespaceableResourceInterface
 }
 
-func Wrap(store types.Store, users v3.UserInterface, grbLister v3.GlobalRoleBindingLister, grLister v3.GlobalRoleLister, secretLister v1.SecretLister, secrets v1.SecretInterface, clusterLister v3.ClusterLister) types.Store {
+func Wrap(store types.Store, mgmt *config.ScaledContext, users v3.UserInterface, grbLister v3.GlobalRoleBindingLister, grLister v3.GlobalRoleLister, secretLister v1.SecretLister, secrets v1.SecretInterface, clusterLister v3.ClusterLister) types.Store {
+	var catalogClient dynamic.ResourceInterface
+	var clusterCatalogClient dynamic.NamespaceableResourceInterface
+	var projectCatalogClient dynamic.NamespaceableResourceInterface
+	dynamicClient, err := dynamic.NewForConfig(&mgmt.RESTConfig)
+	if err != nil {
+		logrus.Warnf("GetClusterStore error creating K8s dynamic client: %v", err)
+	} else {
+		catalogClient = dynamicClient.Resource(v3.CatalogGroupVersionResource)
+		clusterCatalogClient = dynamicClient.Resource(v3.ClusterCatalogGroupVersionResource)
+		projectCatalogClient = dynamicClient.Resource(v3.ProjectCatalogGroupVersionResource)
+
+	}
 	return &Store{
-		Store:          store,
-		Users:          users,
-		GrbLister:      grbLister,
-		GrLister:       grLister,
-		secretMigrator: secretmigrator.NewMigrator(secretLister, secrets),
-		clusterLister:  clusterLister,
+		Store:                store,
+		Users:                users,
+		GrbLister:            grbLister,
+		GrLister:             grLister,
+		secretMigrator:       secretmigrator.NewMigrator(secretLister, secrets),
+		clusterLister:        clusterLister,
+		catalogClient:        catalogClient,
+		clusterCatalogClient: clusterCatalogClient,
+		projectCatalogClient: projectCatalogClient,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -970,39 +976,54 @@ func (r *Store) migrateSecrets(data map[string]interface{}, currentReg, currentS
...
 	}
 	if s.regSecret != nil {
-		data[registrySecretKey] = s.regSecret.Name
+		values.PutValue(data, s.regSecret.Name, "clusterSecrets", registrySecretKey)
+		if _, ok := data[registrySecretKey]; ok {
+			delete(data, registrySecretKey) // migrate from status to spec
+		}
 		rkeConfig.PrivateRegistries = secretmigrator.CleanRegistries(rkeConfig.PrivateRegistries)
 	}
 	s.s3Secret, err = r.secretMigrator.CreateOrUpdateS3Secret(currentS3, rkeConfig, nil)
```
jakefhyde, line 985: It just occurred to me, but we should probably be setting the owner references on these secrets because anyone that has tried to create them and have it not persist to status just has a bunch of orphaned secrets floating around in `cattle-global-data`.

cmurphy: Good call, I've added that

pkg/apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -186,14 +188,14 @@ type ClusterStatus struct {
...
-	PrivateRegistrySecret                string                      `json:"privateRegistrySecret,omitempty" norman:"nocreate,noupdate"`
-	S3CredentialSecret                   string                      `json:"s3CredentialSecret,omitempty" norman:"nocreate,noupdate"`
-	WeavePasswordSecret                  string                      `json:"weavePasswordSecret,omitempty" norman:"nocreate,noupdate"`
-	VsphereSecret                        string                      `json:"vsphereSecret,omitempty" norman:"nocreate,noupdate"`
-	VirtualCenterSecret                  string                      `json:"virtualCenterSecret,omitempty" norman:"nocreate,noupdate"`
-	OpenStackSecret                      string                      `json:"openStackSecret,omitempty" norman:"nocreate,noupdate"`
-	AADClientSecret                      string                      `json:"aadClientSecret,omitempty" norman:"nocreate,noupdate"`
-	AADClientCertSecret                  string                      `json:"aadClientCertSecret,omitempty" norman:"nocreate,noupdate"`
+	PrivateRegistrySecret                string                      `json:"privateRegistrySecret,omitempty" norman:"nocreate,noupdate"` // deprecated
```
jakefhyde, line 191: If we want the deprecated warnings to show up in an IDE it has to be in the form of `Deprecated: `. I have an example Pr here: https://github.com/rancher/norman/pull/416/files

jakefhyde (approved):
LGTM, just left a comment about updating the deprecated comment. Also, should we add something to `migrations.go` to clean up secrets of the form `cluster-s3backup-` as well as the other secret types that may have been left around without owner references?

pkg/apis/management.cattle.io/v3/cluster_types.go
cmurphy: done

jakefhyde (approved):

pkg/api/norman/store/catalog/store.go
```diff
@@ -1,42 +1,64 @@
...
+	clusterCatalogClient dynamic.NamespaceableResourceInterface
+	projectCatalogClient dynamic.NamespaceableResourceInterface
 }
 
-func Wrap(store types.Store, users v3.UserInterface, grbLister v3.GlobalRoleBindingLister, grLister v3.GlobalRoleLister, secretLister v1.SecretLister, secrets v1.SecretInterface, clusterLister v3.ClusterLister) types.Store {
+func Wrap(store types.Store, mgmt *config.ScaledContext, users v3.UserInterface, grbLister v3.GlobalRoleBindingLister, grLister v3.GlobalRoleLister, secretLister v1.SecretLister, secrets v1.SecretInterface, clusterLister v3.ClusterLister) types.Store {
+	dynamicClient, err := dynamic.NewForConfig(&mgmt.RESTConfig)
+	var catalogClient dynamic.ResourceInterface
+	var clusterCatalogClient dynamic.NamespaceableResourceInterface
+	var projectCatalogClient dynamic.NamespaceableResourceInterface
```
rmweir, line 43: nit: can this go above instantiating dynamicClient?

pkg/api/norman/store/catalog/store.go
```diff
@@ -111,7 +133,12 @@ func (s *Store) Delete(apiContext *types.APIContext, schema *types.Schema, id st
 	_, isProject := existing["projectId"]
 	// global catalogs are owners of the secret so it will automatically be cleaned up
 	if isCluster || isProject {
-		if secretName, ok := existing[catalogSecretKey]; ok {
+		if secretName, ok := values.GetValue(existing, "catalogSecrets", catalogSecretKey); ok {
```
rmweir, line 136: nit: should make "catalogSecrets" into a const to reduce change of mistakes.

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -828,15 +828,24 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
...
-	currentS3Secret, _ := existingCluster[s3SecretKey].(string)
-	currentWeaveSecret, _ := existingCluster[weaveSecretKey].(string)
-	currentVsphereSecret, _ := existingCluster[vsphereSecretKey].(string)
-	currentVcenterSecret, _ := existingCluster[virtualCenterSecretKey].(string)
-	currentOpenStackSecret, _ := existingCluster[openStackSecretKey].(string)
-	currentAADClientSecret, _ := existingCluster[aadClientSecretKey].(string)
-	currentAADCertSecret, _ := existingCluster[aadClientCertSecretKey].(string)
-	allSecrets, err := r.migrateSecrets(data, currentRegSecret, currentS3Secret, currentWeaveSecret, currentVsphereSecret, currentVcenterSecret, currentOpenStackSecret, currentAADClientSecret, currentAADCertSecret)
+	getSecretByKey := func(key string) string {
+		if v, ok := values.GetValue(existingCluster, "clusterSecrets", key); ok && v.(string) != "" {
```
rmweir, line 832: nit: use const

pkg/api/norman/store/cluster/cluster_store.go
```diff
@@ -883,6 +892,113 @@ func (r *Store) Update(apiContext *types.APIContext, schema *types.Schema, data
...
+	owner := metav1.OwnerReference{
+		APIVersion: "management.cattle.io/v3",
+		Kind:       "Cluster",
+		Name:       data["id"].(string),
+		UID:        k8sTypes.UID(data["uuid"].(string)),
+	}
+	if allSecrets.regSecret != nil {
+		err = r.secretMigrator.UpdateSecretOwnerReference(allSecrets.regSecret, owner)
+		if err != nil {
+			logrus.Errorf("cluster store: failed to set %s %s as secret owner", owner.Kind, owner.Name)
```
rmweir, line 957: Should these error be unique for each secret? Like specifying if it was reg, s3, weave, etc.. Otherwise, we can just use a var/const for the error message.

pkg/apis/management.cattle.io/v3/catalog_types.go
```diff
@@ -288,3 +289,24 @@ type ClusterCatalog struct {
 type CatalogRefresh struct {
 	Catalogs []string `json:"catalogs"`
 }
+
+type CatalogSecrets struct {
+	CredentialSecret string `json:"credentialSecret,omitempty" norman:"nocreate,noupdate"`
+}
+
+// GetSecret gets a reference to the private catalog secret, either from the CatalogSecrets field or the Status field.
+// CatalogSecrets is preferred because Status is deprecated.
```
rmweir, line 298: This isn't true for the entire status right? If so, should change wording.

pkg/apis/management.cattle.io/v3/catalog_types.go
```diff
@@ -288,3 +289,24 @@ type ClusterCatalog struct {
...
+	return c.Status.CredentialSecret
+}
+
+func (c *ClusterCatalog) GetSecret() string {
+	return c.Catalog.GetSecret()
+}
+
+func (c *ProjectCatalog) GetSecret() string {
+	return c.Catalog.GetSecret()
+}
```
rmweir, line 312: Since Catalog is embedded in ClusterCatalog and ProjectCatalog these aren't necessary right?

rmweir (changes requested):
I would generally try to use a const whereever a string or error is being reused so that there is only a single point of failure.

jakefhyde (dismissed):
LGTM

rmweir (approved):
LGTM

jakefhyde (approved):

Issue #37228, opened 2022-04-07
[Backport 2.5] Update Azure AD API from ADAL to MSAL


samjustus:
rancher/rancher#29306

gaktive:
@mantis-toboggan-md how does @genexpr's proposal sound to you?

mantis-toboggan-md:
@gaktive I would greatly prefer not to do that, given my inexperience with the Ember UI. I'll ping max to discuss options

anupama2501:
Verified on v2.5-head `444fba4`

Fresh checks standard endpoints

1. Install rancher on v2.5-head
2. Enable Azure AD auth by selecting the standard endpoint using the existing app
3. Verify no message appears in rancher UI for the end point upgrade as the endpoint should be the latest
4. Verify the endpoints configured are the latest by running the command against the local cluster's kubeconfig kubectl describe authconfig azuread | grep graph
5. Create a downstream cluster - c1
6. As an admin, created a projects P1 
7. Add users from Azure AD to the downstream cluster as user1:cluster Owner, user2:cluster member, user3: custom [Create Projects], user4: Project P1 - owner, user5: Project P1 member, user6: Read-only in project P1, user7: custom view workloads in project P1
8. Log in as the new user1 and verify user can list the cluster
9. Run pre-upgrade checks by creating a user token for the cluster owner
10. Verify all the pre-upgrade checks work
11. Login as user2 - create a few workloads, Verify user can list cluster members but cannot delete any member
12. Login as user3 - Verify user3 cannot create projects, cannot create workloads, ingresses etc
13. Login as user4 - Verify user can only list the namespaces and workloads in P1, can create namespace, workloads, ingress in only P1
14. Login as user5 - Verify user can only list the namespaces and workloads in P1, can create namespace, workloads, ingress in only P1 and cannot manage members in other projects
15. Login as user6 - Verify user6 can only view project members, workloads but cannot manage workloads
16. Login as user7 - Verify user6 can only view workloads but cannot manage workloads or list members


Upgrade checks standard points:

1. Install rancher on v2.5.15
2. Enable Azure AD auth by selecting the standard endpoint
3. Verify no message appears in rancher UI for the end point upgrade as the endpoint should be the latest (for v2.5.15)
4. Create 3 downstream clusters - c1, c2
5. Add group1 as restricted admin
6. Add users from Azure AD to the downstream cluster as user1:cluster Owner, user2: cluster member in cluster1
[...]

PR #38364, opened 2022-07-21
[Backport] Azure AD Graph API endpoint migration

maxsokolovsky:
## Issue: 

https://github.com/rancher/rancher/issues/37228
 
## Problem

Users of v2.5.16+, who have Azure AD set up as the auth provider in Rancher, will not be able to log in at the end of 2022, when Microsoft shut down the deprecated Azure AD Graph API, which Rancher uses to work with Azure AD (fetch users, groups, etc.).
 
## Solution

Backport the work that enables a migration for existing Azure AD setups to use the new Microsoft Graph API.

There is one difference between this code and  `v2.6.7` code:
1. `UserExtraAttributes` are no longer added to the arguments of the `ap.tokenMGR.CreateTokenAndSetCookie` method, as they are not supported by the `v2.5` code.
The corresponding method definition has been removed, too.
https://github.com/rancher/rancher/blob/c47178eb1457a7ff1aac135c86d3627dc971b704/pkg/auth/providers/azure/azure_actions.go#L118


pkg/auth/providers/azure/azure_actions.go
```diff
@@ -65,7 +71,17 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 			fmt.Sprintf("Failed to parse body: %v", err))
 	}
 
-	azureADConfig := azureADConfigApplyInput.Config
+	azureADConfig := &azureADConfigApplyInput.Config
+
+	// This covers the case where users upgrade Rancher to v2.6.7+ without having used Azure AD as the auth provider.
+	// In 2.6.7+, whether Azure AD is later registered or not, Rancher on startup creates the annotation on the template auth config.
+	// But in the case where the auth config had been created on Rancher startup prior to v2.6.7, the annotation would be missing.
```
KevinJoiner, line 78:
```suggestion
	// This covers the case where users upgrade Rancher to v2.5.16+ without having used Azure AD as the auth provider.
	// In 2.5.16+, whether Azure AD is later registered or not, Rancher on startup creates the annotation on the template auth config.
	// But in the case where the auth config had been created on Rancher startup prior to v2.5.16, the annotation would be missing.
```

KevinJoiner (changes requested):

pkg/auth/providers/azure/azure_actions.go
maxsokolovsky: Very nice catch, thanks!

KevinJoiner (dismissed):

KevinJoiner (approved):

MbolotSuse (approved):
LGTM.

PR #38383, opened 2022-07-22
Update the Go toolchain version for the Windows Drone pipelines

maxsokolovsky:
default-windows-1809 and 2004 pipelines in Drone use the Go 1.14 version of the toolchain, and we have a 3rd-party package used in release/v2.5 that uses code from the stdlib of Go 1.16. The build fails because Go 1.14 fails to compile that package.
https://drone-pr.rancher.io/rancher/rancher/23039/3/2


MbolotSuse (commented):
Seeing as how newer versions of go are generally backwards compatible (see https://tip.golang.org/doc/go1compat), I'm ok with this PR. That being said, I'm not knowledgeable on the windows side, so I would hold off on merging this until Ross/Jamie weighs in.

MbolotSuse (approved):

phillipsj (approved):

PR #38386, opened 2022-07-22
38277 unable to rotate after rotating once

jakefhyde:
## Problem

Encryption key rotation was failing, but looked like it was succeeding. This was caused by an issue where the follower controlplane nodes have the incorrect (previous) status post restart, and because we have already left the `system-agent` space, there was no way to run subsequent plans to ensure that the phase really was correct. Although this issue rights itself rather quickly (as in before the first status check) for specific cluster configurations (i.e. non split-role), for split-role configurations subsequent restarts would exacerbate the issue, creating a long enough gap that the time between verifying the follower nodes and running the next command was not long enough of an interval for all nodes to have the correct (identical) stage. Because failures are not very well reported, this was making the cluster appear as though encryption key rotation succeeded, when in reality encryption key rotation had failed at a previous step.

I'll update the rest of the description tomorrow.

## Solution

Add a new instruction to verify the follower status if the same as the leader status post restart, and sleep until it is, or 10 seconds have passed (chosen at random, clusters could surpass this timeout but I haven't seen it happen with 1000s of secrets thus far).

Set the `Ready` condition on the controlplane to false, log a warning in the provisioning log, and return an error such that it is very noticeable and the intent is clear. This is also in parity with rke1 which states the following message: `"encryption key rotation failed, please restore your cluster from backup"`.

image

At this point the planner is essentially blocked from moving onwards, however we generally want this to make as much noise as possible: a cluster in this state is NOT usable and any changes made after this point are effectively undefined behavior.
 
### Manual Testing

I ran the automated tests in [this PR](https://github.com/rancher/rancher/pull/38259) for 3 node and 3 etcd, 2 controlplane, 3 workers configurations.

### Automated Testing

The tests in [this PR](https://github.com/rancher/rancher/pull/38259) has been updated to test with a fresh cluster, then add 1000 secrets and rotate again, which should cover both cases.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
+++ b/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
@@ -785,8 +785,8 @@ func encryptionKeyRotationIsCurrentStageAllowed(leaderStage string, currentPhase
 // encryptionKeyRotationFailed updates the various status objects on the control plane, allowing the cluster to
 // continue the reconciliation loop. Encryption key rotation will not be restarted again until requested.
 func (p *Planner) encryptionKeyRotationFailed(cp *rkev1.RKEControlPlane, err error) error {
-	if err := p.setEncryptionKeyRotateState(cp, cp.Spec.RotateEncryptionKeys, rkev1.RotateEncryptionKeysPhaseFailed); err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


jakefhyde:
@Oats87 @jiaqiluo Recreated because `drone/pr` wasn't linking.

jiaqiluo (approved):

PR #38399, opened 2022-07-25
Add missing go generate changes

cmurphy:
go generate for 0bae4c92


rancherbot (approved):

Issue #38397, opened 2022-07-25
[Backport v2.5] Cannot update s3 credentials after cluster has been created.


rancherbot:
This is a backport issue for https://github.com/rancher/rancher/issues/38215, automatically created via rancherbot by @anupama2501

Original issue description:

**Rancher Server Setup**
- Rancher version: v2.6.4 - v2.6.7

**Information about the Cluster**
- Cluster Type (Local/Downstream): Downstream
  - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE1

**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**

After the creation of the cluster without setting S3 credentials to take snapshots of the downstream cluster, it's not possible to update the cluster to upload snapshots to S3 buckets. The S3 settings only work when it's set during the creation of the cluster. 

**To Reproduce**
 
- Create a new cluster without setting S3 snapshots;
- After the cluster it's created set the S3 snapshot and try to run a snapshot.
**Result**
Not able to create backups of the downstream cluster and send the backup to S3. Not able to add, update, or remove the secrets.
**Expected Result**

Can add, edit & remove S3 credential in order to enable or disable s3 snapshot capabilities for a cluster. If the cluster was created with S3 credentials, then it is possible to update the secret, but it is much harder to get around if the cluster was created without it.
**Screenshots**

I believe this issue is caused here:

https://github.com/rancher/norman/blob/61b018639b217b33604cae675e189c2afa4cf378/types/convert/merge/merge.go#L11-L20

which is called in `rancher` here:

https://github.com/rancher/rancher/blob/b919539b4869325aadcdb737070c02914a692cb2/pkg/api/norman/store/cluster/cluster_store.go#L841

Before:
![image](https://user-images.githubusercontent.com/33796120/177827299-f7d25ab9-9edd-4173-8f01-eb908d27d3ea.png)

(Service Account Token was purposely spoofed here)

After:
```
value = {interface{} | map[string]interface{}} 
 0 = agentImage -> rancher/rancher-agent:v2.6-head
 1 = aksStatus -> 
 2 = capacity -> 
 3 = gkeStatus -> 
 4 = nodeCount -> 1
 5 = authImage -> rancher/kube-api-auth:v0.1.8
 6 = driver -> rancherKubernetesEngine
 7 = nodeVersion -> 1
 8 = requested -> 
 9 = apiEndpoint -> https://161.35.143.159:6443/
[...]

PR #38401, opened 2022-07-25
[2.5] Move cluster and catalog secrets to Spec

cmurphy:
Norman assumes that no updates to Status happen through the API,
including through the store, and so overwrites any changes[1]. This
means it is impossible to set a new value on Status for clusters or
catalogs upon update, it can only be done on create. This was resulting
in references to secrets being lost.

This change addresses the issue by moving secret values from Status to
Spec for clusters and catalogs.

Implementation notes:
- Notifiers are unaffected because the secret reference is the only
  value in status, since there are no status values in the original
  object it is not overwritten.
- SourceCodeProviderConfigs are unaffected because the secret reference
  was not in status.
- ClusterTemplateRevisions are unaffected because they are never
  updated, only new versions are created.
- The service account token is never updated through the norman store
  and so is unaffected.
- To avoid a complicated migration, unaffected resources are left alone
  even though it introduces inconsistency between the different types of
  resources.
- A getter is added for clusters and catalogs to check both status and
  spec for the secret values. The alternative is to add a migration in a
  controller.

[1] https://github.com/rancher/norman/blob/e21cabdc65b9f556c5cba1a5f2c49297850bdc46/types/convert/merge/merge.go#L13-L15

Backport of https://github.com/rancher/rancher/pull/38340 and https://github.com/rancher/rancher/pull/38399

https://github.com/rancher/rancher/issues/38397


rmweir (approved):
LGTM

PR #38445, opened 2022-07-27
Expose rancher version endpoint

cmurphy:
Expose the rancher version as an unauthenticated endpoint on
/rancherversion. This now includes a flag indicating whether the build
version is Rancher "plus" or not. The version and "plus" indicator is
set at build time and cannot be overridden by an environment variable at
run time. The endpoint is /rancherversion instead of /version because
the /version endpoint is just passed through to kubernetes and gives the
local cluster kubernetes version. The endpoint is unauthenticated rather
than authenticated because the k8s /version is already exposed[1] as
well as the /v1-{k3s,rke2}-release/* endpoints, so security through
obscurity doesn't buy us much here.

[1] https://github.com/kubernetes/kubernetes/issues/84040

https://github.com/rancher/rancher/issues/38554

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,12 +1,54 @@
+// Package version gathers version data from variables set at build time and runtime
+// and provides access points to retrieve them.
 package version
 
-import "fmt"
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+)
 
 var (
 	Version   = "dev"
 	GitCommit = "HEAD"
 )
 
+const primeEnv = "RANCHER_PRIME"
+
+// Info encapsulates version metadata.
+type Info struct {
+	Version      string
+	GitCommit    string
+	RancherPrime string
+}
+
+// FriendlyVersion returns a human-readable string that can be included in log output.
 func FriendlyVersion() string {
 	return fmt.Sprintf("%s (%s)", Version, GitCommit)
+}
+
+type versionHandler struct {
+	info Info
+}
+
+// NewVersionHandler checks the runtime environment for the RANCHER_PRIME environment variable
+// and uses that along with build-time version information to create an HTTP handler.
+func NewVersionHandler() http.Handler {
+	var rancherPrime = "false"
+	if isPrime, ok := os.LookupEnv(primeEnv); ok && isPrime == "true" {
+		rancherPrime = "true"
+	}
+	return &versionHandler{info: Info{Version, GitCommit, rancherPrime}}
+}
+
+// ServeHTTP handles GET requests for version information.
+func (h *versionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	body, err := json.Marshal(h.info)
+	if err != nil {
+		rw.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	rw.Write(body)
+	rw.WriteHeader(http.StatusOK)
 }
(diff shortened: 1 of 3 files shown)


rmweir (approved):
LGTM

scripts/build-server
```diff
@@ -12,11 +12,13 @@ RKE_VERSION="$(grep -m1 'github.com/rancher/rke' go.mod | awk '{print $2}')"
 
 # Inject Setting values
 DEFAULT_VALUES="{\"rke-version\":\"${RKE_VERSION}\"}"
+RANCHER_PRIME="${RANCHER_PRIME:-false}"
```
cbron, line 15: We won't be able to do this at build time of the go binary. So I believe this needs to go in env var.

cmurphy: It can't be an env var, because then it would be changeable at run time. It has to be done at build time. Why can't we have drone set this at build time?

cmurphy:
Changed to use a runtime environment variable and added tests.

rmweir (dismissed):

KevinJoiner (dismissed):
LGTM

KevinJoiner:
nit: 
code comments

cmurphy:
Added doc comments

KevinJoiner (dismissed):

pkg/version/version.go
```diff
@@ -1,12 +1,54 @@
...
+type versionHandler struct {
+	info Info
+}
+
+// NewVersionHandler checks the runtime environment for the RANCHER_PRIME environment variable
+// and uses that along with build-time version information to create an HTTP handler.
+func NewVersionHandler() http.Handler {
+	var rancherPrime = "false"
+	if isPrime, ok := os.LookupEnv(primeEnv); ok && isPrime == "true" {
+		rancherPrime = "true"
```
cbron, line 40: There was an additional request if this is true, which is to set the setting `ui-brand="suse"`.

cbron: Nevermind, this was changed.

pkg/version/version.go
```diff
@@ -1,12 +1,54 @@
...
+	"net/http"
+	"os"
+)
 
 var (
 	Version   = "dev"
 	GitCommit = "HEAD"
 )
 
+const primeEnv = "RANCHER_PRIME"
```
mattfarina, line 17: For my own education, is there a reason this isn't refixed with `CATTLE_`?

cmurphy: The settings manager uses the `CATTLE_` prefix to look up settings in the environment, so using the prefix would imply that the variable is available as a Rancher setting (though this isn't a 100% rule). Since this isn't a setting, there's no technical reason to use the prefix. This may reach into the realm of personal preference, but `CATTLE_RANCHER_...` feels redundant to me, so I kept it short.

cbron (dismissed):
May change title of env var, but we can do that later.

rmweir (approved):

cmurphy:
needed to rebase

rmweir (approved):

cbron (approved):

pkg/version/version.go
```diff
@@ -1,12 +1,54 @@
...
+
+// ServeHTTP handles GET requests for version information.
+func (h *versionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	body, err := json.Marshal(h.info)
+	if err != nil {
+		rw.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	rw.Write(body)
+	rw.WriteHeader(http.StatusOK)
```
sbstp, line 53: Call order might be inverted. Probably not a huge deal since
```go
	// If WriteHeader has not yet been called, Write calls
	// WriteHeader(http.StatusOK) before writing the data
```

PR #38455, opened 2022-07-28
Refactor RKE1 cluster cleanup and add support for static and dynamic node pools

markusewalker:
## Issues:  
[RKE1 cluster cleanup needs to occur before second cluster creation, not after all clusters are created](https://github.com/rancher/qa-tasks/issues/452)

[rke1/provisioning_node_driver_test.go should add test cases to allow cluster creation and distribute roles amongst three nodes](https://github.com/rancher/qa-tasks/issues/453)

 
## Problem

**Problem 1**
The `rke1/provisioning_ node_driver_test.go` test case had an issue where all of the clusters would be deleted AFTER all of the clusters were provisioned. This is incorrect behavior and we want clusters to clean up immediately after they are successfully provisioned.

This is the behavior the `rke2/provisioning_node_driver_test.go` exhibits, so we should ensure parity.

**Problem 2**
Before this PR, the RKE1 `provisioning_node_driver_test.go` would not allow nodepools to be both static and dynamic when creating cluster; it only was static. This is an issue because this does not encapsulate the test cases we want and what we see in RKE2.

Now, RKE1 has this functionality and more or less, just like RKE2 does, so more parity is being captured here.
 
## Solution

**Solution 1**
Created new `CreateRKE1Cluster` function in `tests/framework/extensions/clusters/clusters.go` that handles creating the cluster AND deleting the cluster. The function makes use of the `GetManagementWatchInterface` function to monitor the cluster's status in order to properly handle this.

**Solution 2**
Refactored the `nodepools.go` file to use new `RKE1NodePoolSetup` function that does the following:

- Creates `nodePoolConfig` variable that is assigned to `management.NodePool`
- The `nodeRoles` is iterated over each of the cluster's roles and a NodePool is created based upon the `nodePoolConfig`; this allows static and dynamic cluster creation

Then, the `provisioning_node_driver_test.go` calls this function instead of creating the actual NodePool in this file itself, providing further abstraction, which is ultimately much cleaner and easier to understand.
 
### Automated Testing

- Local testing on client machine
- Automated Jenkins run showing the success (given to reviewers offline for review)


igomez06 (dismissed):
Looks Good! Thank you Marcus

slickwarren (dismissed):

markusewalker:
@sowmyav27 this PR is good to merge whenever you have a chance to look over :)

markusewalker:
@igomez06 @slickwarren thank you both for your original reviews! I am going to repurpose this PR to include https://github.com/rancher/qa-tasks/issues/453. I have a fix for this issue as well that I verified is working locally. However, this change is a bit bigger, so I want to do more testing next week.

I am going to include the files shortly, but I will be reaching out to you both offline with a freeform job showing the success of the job before we proceed.

tests/v2/validation/provisioning/README.md
```diff
@@ -6,13 +6,25 @@ These are the examples of the configurations needed to run the provisioning pack
...
+      },
+      {
+        "worker": true,
+        "quantity": 2,
+      }
+    ],
+    "kubernetesVersion": ["v1.24.2-rancher1-1"],
     "cni": ["calico"],
-    "providers": ["linode", "aws", "do", "harvester"],
+    "providers": ["linode", "aws", "azure"],
```
slickwarren, line 27: harvester is still supported here, right?

slickwarren (dismissed):

tests/v2/validation/provisioning/README.md
igomez06: It should be right? Markus did you mean to delete this?

tests/v2/validation/provisioning/rke1/config.go
```diff
@@ -0,0 +1,25 @@
+package rke1
```
igomez06, line 1: I think it would be better to just have one config.go file, and have it be in the v2/validation/provisioning directory above rke1 and rke2. That way we don't have two of the same file in each in package. And baseClusterName for each pack (rke1 and rke2) can have the correct prefix. Makes sense?

igomez06 (commented):
Great work, I have one nitpick about the config file, but everything else looks good

tests/v2/validation/provisioning/README.md
markusewalker: Yeah, I had deleted it because currently, we support Linode, AWS, Azure. We don't officially support Harvester yet, so I had removed it, but I can put that back.

tests/v2/validation/provisioning/rke1/config.go
markusewalker: No problem, I can remove it. I had added it in this PR, but I'll revert those changes 👍🏿 .

markusewalker:
@slickwarren @igomez06 thank you both for your feedback, the PR should be good to go now.

slickwarren (approved):

markusewalker:
@slickwarren @igomez06 not sure what happened here, but I believe recent drone failures really caused issues here. I'm going to close this PR and make a new one, since it's just easier that way. It'll be up shortly.

PR #38486, opened 2022-08-01
Bump system-agent to v0.2.10

jakefhyde:
## Problem

Airgap imported clusters could not upgrade the system-agent.

## Solution

Bump system agent to fix which includes pulling uninstall locally when upgrading system-agent for imported clusters in airgap.

### Manual Testing

Tested in QA setup, upgrade succeeded.


cbron (approved):
As long as we don't need to bump wins versions we should be ok. Also I don't know what this is pulling in, but I'm assuming that has been reviewed separately.

snasovich (approved):

PR #38487, opened 2022-08-01
[2.5] Clear existing svc acct token if present

jakefhyde:


HarrisonWAffel (approved):

cmurphy (changes requested):
What about eks_cluster_handler.go, gke_cluster_handler.go, and provisioner.go?

cmurphy (approved):

PR #38493, opened 2022-08-02
additional_test_long-fixes

vivek-infracloud:
## Problem

 
Some of the tests fails while running **rancher-v3_additional_tests_long** jenkins job.

This PR should fix the following tests.

test_rbac_bkp_restore
test_rbac_cert_rotation

## Solution

As per discussion with Izaac, the changes in rancher v2.6.2/v2.6.3 which resulted in the failing of the above tests.
Changing the assert condition and timeout condition as in the PR, fixes the failing tests.


tests/validation/tests/v3_api/common.py
```diff
@@ -1693,6 +1693,7 @@ def wait_for_app_to_active(client, app_id,
         app_data = client.list_app(id=app_id).data
         application = app_data[0]
     while application.state != "active":
+        timeout = start + 300
```
izaac, line 1696: Wasn't this intended for the `wait_for_backup_to_active()`method?

tests/validation/tests/v3_api/test_bkp_restore_local.py
```diff
@@ -102,10 +98,7 @@ def test_rbac_bkp_restore_restore(role):
     user_token2 = rbac_get_user_token_by_role(role)
     user_client2 = get_client_for_token(user_token2)
     user_cluster2 = user_client2.list_cluster(name=CLUSTER_NAME).data[0]
-    with pytest.raises(ApiError) as e:
-        user_cluster2.restoreFromEtcdBackup(etcdBackupId=backup_id)
-    assert e.value.error.status == 403
-    assert e.value.error.code == 'PermissionDenied'
+    assert hasattr(user_cluster2, 'restoreFromEtcdBackup') == False
```
izaac, line 101: This is a good call, im just not sure if this would work on 2.5.x

izaac (commented):
A couple of comments added

tests/validation/tests/v3_api/common.py
vivek-infracloud: Right, this change was for testing another test failure. I'll update PR.

tests/validation/tests/v3_api/test_bkp_restore_local.py
vivek-infracloud: I just updated the PR, this should take care of the difference in rancher v2.5.x and v2.6.x

tests/validation/tests/v3_api/common.py
```diff
@@ -2217,6 +2217,7 @@ def generate_template_global_role(name, new_user_default=False, template=None):
 def wait_for_backup_to_active(cluster, backupname,
                               timeout=DEFAULT_TIMEOUT):
     start = time.time()
+    timeout = start + 300
```
izaac, line 2220: could we change this to 

```python
timeout = start + timeout
```

This way we use the default timeout value instead of a fixed value of 300

izaac (changes requested):
Just a small change request everything else looks good

izaac (approved):
LGTM

anupama2501 (approved):
lgtm

PR #38520, opened 2022-08-03
Refactor RKE1 cluster cleanup and support static/dynamic nodepools

markusewalker:
## Issues:  
[RKE1 cluster cleanup needs to occur before second cluster creation, not after all clusters are created](https://github.com/rancher/qa-tasks/issues/452)

[rke1/provisioning_node_driver_test.go should add test cases to allow cluster creation and distribute roles amongst three nodes](https://github.com/rancher/qa-tasks/issues/453)

 
## Problem

**Problem 1**
The `rke1/provisioning_ node_driver_test.go` test case had an issue where all of the clusters would be deleted AFTER all of the clusters were provisioned. This is incorrect behavior and we want clusters to clean up immediately after they are successfully provisioned.

This is the behavior the `rke2/provisioning_node_driver_test.go` exhibits, so we should ensure parity.

**Problem 2**
Before this PR, the RKE1 `provisioning_node_driver_test.go` would not allow nodepools to be both static and dynamic when creating cluster; it only was static. This is an issue because this does not encapsulate the test cases we want and what we see in RKE2.

Now, RKE1 has this functionality and more or less, just like RKE2 does, so more parity is being captured here.
 
## Solution

**Solution 1**
Created new `CreateRKE1Cluster` function in `tests/framework/extensions/clusters/clusters.go` that handles creating the cluster AND deleting the cluster. The function makes use of the `GetManagementWatchInterface` function to monitor the cluster's status in order to properly handle this.

**Solution 2**
Refactored the `nodepools.go` file to use new `RKE1NodePoolSetup` function that does the following:

- Creates `nodePoolConfig` variable that is assigned to `management.NodePool`
- The `nodeRoles` is iterated over each of the cluster's roles and a NodePool is created based upon the `nodePoolConfig`; this allows static and dynamic cluster creation

Then, the `provisioning_node_driver_test.go` calls this function instead of creating the actual NodePool in this file itself, providing further abstraction, which is ultimately much cleaner and easier to understand.
 
### Automated Testing

- Local testing on client machine
- Automated Jenkins run showing the success (given to reviewers offline for review)

### DIFF
--- a/tests/framework/extensions/clusters/clusters.go
+++ b/tests/framework/extensions/clusters/clusters.go
@@ -11,7 +11,6 @@ import (
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
 	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
 	provisioning "github.com/rancher/rancher/tests/framework/clients/rancher/generated/provisioning/v1"
-	"github.com/rancher/rancher/tests/framework/pkg/namegenerator"
 	"github.com/rancher/rancher/tests/framework/pkg/wait"
 	"github.com/rancher/rancher/tests/integration/pkg/defaults"
 	corev1 "k8s.io/api/core/v1"
@@ -67,12 +66,6 @@ func IsHostedProvisioningClusterReady(event watch.Event) (ready bool, err error)
 	}
 
 	return false, nil
-}
-
-func RKE1AppendRandomString(baseClusterName string) string {
-	const defaultRandStringLength = 5
-	clusterName := "auto-" + baseClusterName + "-" + namegenerator.RandStringLower(defaultRandStringLength)
-	return clusterName
 }
 
 // NewRKE1lusterConfig is a constructor for a v3.Cluster object, to be used by the rancher.Client.Provisioning client.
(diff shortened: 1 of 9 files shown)


slickwarren (approved):

tests/framework/extensions/clusters/clusters.go
```diff
@@ -162,6 +155,49 @@ func NewRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName
...
+
+		watchInterface, err := adminClient.GetManagementWatchInterface(management.ClusterType, opts)
+
+		if err != nil {
+			return err
+		}
+
+		return wait.WatchWait(watchInterface, func(event watch.Event) (ready bool, err error) {
+			if event.Type == watch.Error {
+				return false, fmt.Errorf("there was an error deleting cluster")
```
slickwarren, line 190: is the error returned from `watch` something that could help with debugging why the cluster isn't deleted properly?

slickwarren (approved):

tests/framework/extensions/clusters/clusters.go
igomez06: What do you mean Caleb? WatchWait does return an error if there is indeed one, so 
`return wait.WatchWait()` is the same as 


`err = watit.WatchWait()`
`return err`

tests/v2/validation/provisioning/rke2/config.go
```diff
@@ -13,6 +14,7 @@ const (
 
 type Config struct {
 	NodesAndRoles      []machinepools.NodeRoles `json:"nodesAndRoles" yaml:"nodesAndRoles" default:"[]"`
+	NodesAndRolesRKE1  []nodepools.NodeRoles    `json:"nodesAndRolesRKE1" yaml:"nodesAndRolesRKE1" default:"[]"`
```
igomez06, line 17: Is it necessary to have two versions of this Markus? And also this config file would fit better outside of rke2 and just in v2/validation/provisioning since both rke1 and rke2 both use it.

igomez06 (commented):
Please move config.go out of rke2 and move it the parent directory. Also would that take away the need two have two node role fields?

markusewalker:
> Please move config.go out of rke2 and move it the parent directory. Also would that take away the need two have two node role fields?

Yeah, since RKE1 uses `nodepools` and not `machinepools`, there needs to be two separate fields as both are calling different files with different functionality. I can move out the `config.go` no problem and run a new freeform job to make sure all is good.

tests/v2/validation/provisioning/rke2/config.go
igomez06: Nevermind I like that they are different, I think that works fine. But config.go should definitely go in the parent directory and not in rke2.

tests/framework/extensions/clusters/clusters.go
slickwarren: oh ok got it, thanks!

tests/v2/validation/provisioning/rke2/config.go
markusewalker: Since we're moving config.go to the parent directory, does it make sense to do that for providers.go as well? Or do we think it's best to keep that inside of the `rke2` folder?

igomez06: Yes move that as well. Anything that both rke2/rke1 use.

markusewalker: Never mind, neglected to remember that both `providers.go` are vastly different. Pushed the changes and running freeform jobs for RKE1 and RKE2. @slickwarren @igomez06 I'll ping you both offline once those are done for review.

igomez06 (approved):
LGTM

slickwarren (approved):

bmdepesa (approved):

Issue #38501, opened 2022-08-02
[BUG] Fleet bundle in error state after upgrade rancher from 2.6.6 to 2.6.7-rc5 


jiaqiluo:
**Rancher Server Setup**
- Rancher version:  from 2.6.6 to 2.6.7-rc5 
- Installation option (Docker install/Helm Chart): HA
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1 v1.23.6 
- Proxy/Cert Details: none 

**Information about the Cluster**
- Kubernetes version: RKE2 v1.23.7+rke2r2 
- Cluster Type (Local/Downstream): downstream 
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): one custom cluster, one node-driver cluster, both use AWS EC2 instance 

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Admin
  - If custom, define the set of permissions:

**Describe the bug**

After upgrading rancher from 2.6.6 to 2.6.7-rc5, some of the bundles are in the error state, as the following screen shows. 
Fleet version seen on the local and downstream cluster is `100.0.4+up0.3.10-rc5`

![screencapture-sowmaupg-qa-rancher-space-dashboard-c-local-fleet-fleet-cattle-io-bundle-2022-08-02-11_58_35](https://user-images.githubusercontent.com/6218999/182456764-d9c6e46c-f350-42a3-a78d-8ac10e4c92e3.png)

**To Reproduce**

- HA install Rancher 2.6.6
- provision some downstream clusters, rke1, rke2, node-driver, custom 
- upgrade rancher to 2.6.7-rc5 

**Result**
see the above screenshot 

**Expected Result**

All bundles should upgrade successfully. 

**Additional Info**
![screencapture-sowmaupg-qa-rancher-space-dashboard-c-c-m-lmbn99hj-apps-catalog-cattle-io-app-cattle-system-mcc-sowmya-266-upg-managed-system-upgrade-controller-2022-08-02-12_12_11](https://user-images.githubusercontent.com/6218999/182457458-9bb4b164-dce1-46b9-804c-e7b56ea6f5bc.png)
[...]

Issue #36810, opened 2022-03-09
Shared Label Causes Pods to Show Under Another Deployment in the Cluster Explorer UI 


jiaqiluo:
**Rancher Server Setup**
- Rancher version: 2.6-head 3df76e4
- Installation option (Docker install/Helm Chart): docker install 
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: v1.21.10+rke2r1
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):  Infrastructure Provider, Digital Ocean

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Admin

   
**Describe the bug**

The label `upgrade.cattle.io/controller: system-upgrade-controller` is shared by the pods of the deployment `system-upgrade-controller` and the job `apply-system-agent-upgrader-on-xxx`, and is used as the selector in the deployment.  As the result, the pod for the job shows under the deployment. 

**To Reproduce**

- create an RKE2 cluster 

**Result**
The pods for the job show under the deployment. 

**Expected Result**

The pods for the job should not show under the deployment. 

**Screenshots**

![screencapture-165-232-129-73-dashboard-c-c-m-5nx6t9w6-explorer-apps-deployment-cattle-system-system-upgrade-controller-2022-03-08-17_25_36](https://user-images.githubusercontent.com/6218999/157350103-46a9427f-0c35-499c-bf6d-9f7c032c5f3f.png)

**Additional context**

This issue does not see in the old UI
<img width="1425" alt="image" src="https://user-images.githubusercontent.com/6218999/157350522-ce02ea8d-2227-4416-a8c6-50a1954a6c57.png">

paynejacob:
This issue is in the system upgrade controller.  The job selectors need to be more specific to prevent collisions like this.

HarrisonWAffel:
I could not reproduce this issue on 2.6 head (851ff2f401b2c932869407cb75751806bae72776).

Screenshots:

Deployment list is empty 
![Screen Shot 2022-04-11 at 12 37 57 PM](https://user-images.githubusercontent.com/25062254/162791220-84cf65d1-22f0-4936-a188-455e8045ce66.png)

[...]

PR #38527, opened 2022-08-04
downgrade system-upgrade-controller chart version to 100.0.2+up0.3.2

HarrisonWAffel:
Issues: https://github.com/rancher/rancher/issues/38501 https://github.com/rancher/rancher/issues/36810

This PR downgrades the chart version for the system-upgrade-controller to the last released version of the chart.


jakefhyde (approved):

jiaqiluo (approved):

jiaqiluo:
Note: Because the only change in the new chart 100.0.3+up0.3.2 is adding the labels and it is reverted, so we do not need to release a new version, and therefore switching back to the existing related version.

Issue #38514, opened 2022-08-03
[BUG] Cannot Provision RKE2 Cluster on 2.6-head


HarrisonWAffel:
**Rancher Server Setup**
- Rancher version: v2.6-head
- Installation option (Docker install/Helm Chart): Local IDE, should repro on docker-install as well
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details: 

**Information about the Cluster**
- Kubernetes version: v1.23.8+rke21 and v1.24.2+rke2r1
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Digital Ocean

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Admin

**Describe the bug**
I cannot provision an RKE2 Cluster. After instance creation, the system-agent reports 

```
Aug 03 15:49:03 haffel-test-pool1-05e4fb45-t5d6v rancher-system-agent[1971]: W0803 15:49:03.252401    1971 reflector.go:325] pkg/mod/github.com/rancher/client-go@v1.24.0-rancher1/tools/cache/reflector.go:168: failed to list *v1.Secret: the server has asked for the client to provide credentials (get secrets.meta.k8s.io)
Aug 03 15:49:03 haffel-test-pool1-05e4fb45-t5d6v rancher-system-agent[1971]: E0803 15:49:03.253095    1971 reflector.go:139] pkg/mod/github.com/rancher/client-go@v1.24.0-rancher1/tools/cache/reflector.go:168: Failed to watch *v1.Secret: failed to list *v1.Secret: the server has asked for the client to provide credentials (get secrets.meta.k8s.io)
```

The local rancher server continuously reports 
```
2022/08/03 12:59:18 [DEBUG] no local installation script found, moving on to url: https://raw.githubusercontent.com/rancher/system-agent/v0.2.10/install.sh
2022/08/03 12:59:18 [DEBUG] DesiredSet - No change(2) /v1, Kind=ServiceAccount fleet-default/haffel-124-test-bootstrap-template-f5ddt-machine-bootstrap for rke-machine fleet-default/haffel-124-test-bootstrap-template-f5ddt
[...]

PR #38529, opened 2022-08-04
[WIP] Add clusterrole and clusterrolebinding for v2prov bootstrap service account

Oats87:
Add clusterrole and clusterrolebinding for v2prov bootstrap service account to allow GET on local cluster in order to allow authorization

## Issue: 
https://github.com/rancher/rancher/issues/38514
 
## Problem
As part of a change made in Steve (see: https://github.com/rancher/rancher/pull/38497), v2prov system agents were no longer able to authenticate with the local cluster in order to watch the v2prov plan secrets for cluster provisioning changes, as the service account created for each system agent did not have access to GET the local cluster.

## Solution
The solution is to add a `ClusterRole` and `ClusterRoleBinding` for each `v2prov` bootstrap service account which grants access to the local cluster.
 
## Testing
Test v2provisioning, on both upgrade and greenfield clusters.

### Manual Testing
WIP, this PR is being tested as I develop it.

### Automated Testing
The `provisioning-tests-rke` and `provisioning-tests-k3s` should cover automated testing for this.

We must make sure the system agents that exist for clusters created pre-2.6.7 continue to function after the upgrade is complete. There may be transient errors in the system-agent logs similar to what is found in https://github.com/rancher/rancher/issues/38514 until reconciliation of the service account clusterrole/clusterrolebinding is complete.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
 	rkecontroller "github.com/rancher/rancher/pkg/generated/controllers/rke.cattle.io/v1"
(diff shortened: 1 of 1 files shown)


cbron:
Should not be needed for now.

PR #38534, opened 2022-08-05
Adding enhancement to route for v2 prov

MbolotSuse:
Relies on rancher/steve#56 to go in first.

### DIFF
--- a/pkg/api/steve/proxy/proxy.go
+++ b/pkg/api/steve/proxy/proxy.go
@@ -136,8 +136,10 @@ func (h *Handler) authLocalCluster(router *gmux.Router) func(rw http.ResponseWri
 	return func(rw http.ResponseWriter, req *http.Request) {
 		authed := h.userCanAccessCluster(req, "local")
 		if !authed {
-			rw.WriteHeader(http.StatusUnauthorized)
-			return
+			if req.Context().Value(auth.CattleAuthFailed) != "true" {
+				rw.WriteHeader(http.StatusUnauthorized)
+				return
+			}
 		}
 		router.NotFoundHandler.ServeHTTP(rw, req)
 	}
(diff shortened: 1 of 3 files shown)


MbolotSuse:
Build has been cancelled for now until the PR for steve goes in. This will require steve to be updated to take advantages of the changes made there.

pkg/api/steve/proxy/proxy.go
```diff
@@ -87,8 +87,10 @@ func routeToShellProxy(key, value string, localSupport bool, localCluster http.H
 			if localSupport {
 				authed := proxyHandler.userCanAccessCluster(r, cluster)
 				if !authed {
-					rw.WriteHeader(http.StatusUnauthorized)
-					return
+					if r.Context().Value(auth.CattleAuthFailed) == nil {
```
cbron, line 90: nit here and below: 
```suggestion
					if r.Context().Value(auth.CattleAuthFailed) != "true" {
```

pkg/api/steve/proxy/proxy.go
```diff
@@ -65,7 +65,7 @@ func NewProxyMiddleware(sar v1.AuthorizationV1Interface,
 
 	mux := gmux.NewRouter()
 	mux.UseEncodedPath()
-	mux.PathPrefix("/api").HandlerFunc(proxyHandler.authLocalCluster(mux))
+	mux.PathPrefix("/api").HandlerFunc(proxyHandler.authLocalCluster(localCluster))
```
cbron, line 68: I don't understand this well enough to say whether this is good or bad, but seems noteworthy ?

pkg/api/steve/proxy/proxy.go
rmweir: I agree here that we could be more explicit.

rmweir (changes requested):
Besides @cbron feedback about checking for actual value instead of nil, LGTM

MbolotSuse:
Letting the build go through - this is ready now.

cbron (approved):
lgtm when build passes

rmweir (approved):
LGTM

PR #38537, opened 2022-08-05
Adding enhancement to route for v2 prov

MbolotSuse:
Relies on rancher/steve#57 to go in first.


MbolotSuse:
Build has been cancelled for now until the PR for steve goes in. This will require steve to be updated to take advantages of the changes made there.

Issue #37818, opened 2022-05-25
KEv2 unit/integration tests


Sahota1225:
This ticket is a placeholder for KEv2 testing:

annablender:
Finishing GKE unit tests is blocked by a provisioning issue where the default vpn network does not exist in the new GCP configuration. Need assistance

annablender:
Feedback from SUSE IT is for a GCP project group member to create the default VPC. Will work on that using https://cloud.google.com/vpc/docs/create-modify-vpc-networks

annablender:
If rancher throws an error that _default network does not exist_, follow_ [the docs](https://cloud.google.com/vpc/docs/create-modify-vpc-networks#create-auto-network) to create an auto mode VPC network named `default.` Auto created subnets generated for each region will work for this case. Try reprovisioning the GKE cluster. Thank you also to @cmurphy for her advice here

Sahota1225:
Closing this issue as there is no validation for QA.

PR #38556, opened 2022-08-08
[DNM] KEv2 unit tests for AKS EKS and GKE handlers

annablender:
Issue: https://github.com/rancher/rancher/issues/37818

 
## Problem

Add KEv2 unit tests for cloud providers.
 
## Solution

I wrote KEv2 integration (unit) tests to test code functionality in the rancher/rancher handler for each KEv2 cloud provider (AKS EKS and GKE). The tests are written using the [Go Test Framework](https://pkg.go.dev/testing) and [MockCompose](https://github.com/kelveny/mockcompose). They mock the cluster state and `AKSClusterConfig` objects using test files to simulate reactions to a real cluster, and mock any mockable Interfaces used by the OperatorController. Test files are located in each handler folder and test files are in a `/test` folder in each handler folder. The unit test structure is different from `rancher/rancher/tests` because [Go Test Framework](https://pkg.go.dev/testing) requires test files to be in the same package as the source code.

File structure

```
rancher/rancher
-> pkg/controllers/management/aks
     -> aks_cluster_handler.go
     -> aks_cluster_handler_test.go
     -> tests/
          -> mockdata1.yaml
          -> mockdata2.json
```

## Testing

Git clone rancher/rancher

If you want to test each provider separately

```
go test -v ./pkg/controllers/management/aks
go test -v ./pkg/controllers/management/eks
go test -v ./pkg/controllers/management/gke
```

If you want to run all tests

```
go test -v ./pkg/controllers/management/aks ./pkg/controllers/management/eks ./pkg/controllers/management/gke
```

If you want to check test coverage (current coverage of KEv2: 35.1%, not including mocked sibling funcs so actual coverage is likely around 60-70%)

```
go test (or alias command) -coverprofile=coverage.out
go tool cover -html=coverage.out
```


annablender:
Here is a [branch comparison with master](https://github.com/rancher/rancher/compare/release/v2.6...annablender:rancher:kev2-integration-tests) that might be helpful.

snasovich:
@annablender , are these unit tests? If they are, let's update title and desciption.

pkg/controllers/management/aks/test/buildakscccreateobject.json
```diff
@@ -0,0 +1,69 @@
...
+    "tags": {},
+    "virtualNetwork": null,
+    "virtualNetworkResourceGroup": null
+  },
+  "status": {
+    "failureMessage": "",
+    "phase": "",
+    "rbacEnabled": null
+  }
+}
```
jakefhyde, line 69: Can we remove some of the unneeded values in all of these yaml files? Also I'm assuming these tests don't actually need finalizers or timestamps or valid AKS/EKS/GKE provisioning related data, so we should try to remove as much as we can.

jakefhyde (changes requested):

pkg/controllers/management/aks/aks_cluster_handler_test.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	cluster, err := mockOperatorController.onClusterChange("", &mockCluster)
+
+	// check that cluster is updating node pool
+	if cluster.Status.Conditions[0].Status != "True" ||
+		cluster.Status.Conditions[1].Status != "Unknown" || err != nil {
+		t.Error("provisioned status should be True, updated status should be Unknown and cluster returned successfully: " + err.Error())
+	}
+}
```
jakefhyde, line 118:
```suggestion
	if err != nil {
		t.Errorf("error running onClusterChange: %s", err)
	}
	if !rke2.Provisioned.IsTrue(cluster) || !rke2.Updated.IsUnknown(cluster) {
		t.Errorf("provisioned status should be True, updated status should be Unknown and cluster returned successfully: %s", err)
	}
}
```
I would break up these errors, since an error from `onClusterChange` should not be related to the conditions. I would also use the more idiomatic approach to the conditions (as these tests grow and expand, manually managing condition indices will become more cumbersome). Also using `Errorf("... %s", err)` instead is merely a stylistic choice, but it's one I prefer: it's more concise, usually allows for better git diffs, and the intention is clearer.

jakefhyde (changes requested):

pkg/controllers/management/aks/aks_cluster_handler_test.go
annablender: Sometimes `config, err` returned from a handler means an API call failed yes but it could also mean the cluster isn't in the state it should be so the cluster is returned to wait for updates.

Example:
```
cluster, err := e.CheckCrdReady(cluster, "aks")
	if err != nil {
		return cluster, err
	}
```

But in general yes, I think this is a good idea.

tests/kev2/README.md
```diff
@@ -0,0 +1,56 @@
...
+go test -v ./pkg/controllers/management/gke
+```
+
+If you want to run all tests
+
+```
+go test -v ./pkg/controllers/management/aks ./pkg/controllers/management/eks ./pkg/controllers/management/gke
+```
+
+If you want to check test coverage (current coverage of AKS: 34.8%)
```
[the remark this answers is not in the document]
annablender: I'll not include the %. Especially because the command output for the % will be lower than accurate, because it doesn't include mocked sibling func code being tested.

pkg/controllers/management/aks/aks_cluster_handler_test.go
```diff
@@ -70,8 +77,8 @@ func Test_onClusterChange_Default(t *testing.T) {
 	if err != nil {
 		t.Errorf("error running onClusterChange: %s", err)
 	}
-	if cluster.Status.Conditions[0].Status != "Unknown" {
-		t.Errorf("status should be Unknown and cluster returned successfully")
+	if cluster.Status.Conditions[0].Status != mockConditionUnknown.Status {
```
jakefhyde, line 80: Sorry, I should've clarified, the issue is the lefthand side of the comparison. I don't even know what condition that's supposed to be without looking at the yaml. Something like the following would be alot more idiomatic: 
```
	if !rke2.Provisioned.IsTrue(cluster) || !rke2.Updated.IsUnknown(cluster) {
		t.Errorf("provisioned status should be True, updated status should be Unknown and cluster returned successfully: %s", err)
	}
```

pkg/controllers/management/aks/aks_cluster_handler_test.go
```diff
@@ -8,19 +8,26 @@ import (
...
+	MockAksClusterConfigFilename 		= "test/updateaksclusterconfig.json"
+	MockAksClusterConfigClusterFilename = "test/updateaksclusterconfig.yaml"
+	MockBuildAksCCCreateObjectFilename 	= "test/buildakscccreateobject.json"
 )
 
+var (
+	mockOperatorController mockAksOperatorController // Operator controller with mock interfaces & sibling funcs
 
-var mockOperatorController mockAksOperatorController	// Operator controller with mock interfaces & sibling funcs
+	mockConditionUnknown = v3.ClusterCondition{		 // Mock conditions
```
jakefhyde, line 24: You can remove these as well, checking status itself isn't the issue.

jakefhyde (changes requested):

pkg/controllers/management/aks/aks_cluster_handler_test.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func Test_onClusterChange_UpdateNodePool(t *testing.T) {
+	mockOperatorController = getMockAksOperatorController("update")
+	mockCluster, err := getMockV3Cluster(MockUpdateClusterFilename)
+	if err != nil {
+		t.Errorf("error getting mock v3 cluster: %s", err)
+	}
+
+	cluster, err := mockOperatorController.onClusterChange("", &mockCluster)
```
jiaqiluo, line 126: The cluster read from the file `test/onclusterchange_update.yaml` does not contain the field `status.driver`, so the function `mockOperatorController.onClusterChange` returns the cluster without touching the cluster's node pool or status at the following lines:
https://github.com/rancher/rancher/blob/68215bcdde090854ff28b43b01d1aa0b69611920/pkg/controllers/management/aks/aks_cluster_handler.go#L92 

That means the function does not validte "updateNodePool", right? 
Or am I misunderstanding anything?

jiaqiluo (changes requested):

pkg/controllers/management/aks/aks_cluster_handler_test.go
```diff
@@ -77,7 +68,7 @@ func Test_onClusterChange_Default(t *testing.T) {
 	if err != nil {
 		t.Errorf("error running onClusterChange: %s", err)
 	}
-	if cluster.Status.Conditions[0].Status != mockConditionUnknown.Status {
+	if !IsProvisionedStatusUnknown(cluster) {
```
jakefhyde, line 71: Sorry let me be more explicit, use these specific conditions right there: https://github.com/rancher/rancher/blob/6f86d4a6b0effbaaecadf183ab8888786ca7b6a4/pkg/controllers/provisioningv2/rke2/common.go#L76-L77
Any time you are indexing conditions (like `cluster.Status.Conditions[x]`) is almost certainly not what you should be doing. The `Condition` objects themselves allow you to write `rke2.Provisioned.IsTrue(cluster)` which will do exactly what you want, but the ordering of the conditions no longer matters. The `Condition` object (in this case `Provisioned` just loops all the conditions until it finds itself. Also I'm like 90% positive that those objects are the ones actually used for kev2 hosted provisioning, so it's probably better to test with the condition objects we are using instead of making specific ones for testing. This is also good future proofing in case we have to ever update the test data and add more conditions.

jakefhyde (changes requested):

jakefhyde (approved):
LGTM

snasovich:
@jiaqiluo , could you please review this again and approve if your feedback is now addressed?

jiaqiluo (approved):

annablender:
Closing this PR because the KEv2 unit tests have already been merged into rancher `release/v2.7` [here](https://github.com/rancher/rancher/pull/38889) then back ported to `release/v2.6` [here](https://github.com/rancher/rancher/pull/39273) as part of the branching strategy for moving to the next minor release of Rancher 2.

PR #38558, opened 2022-08-09
Added OnRemove bindings and annotations

steffantucker:
## Issue: 
 https://github.com/rancher/rancher/issues/36072

## Problem
Cluster role was being deleted before the cluster was fully deleted, leading to the final signal not being sent to the frontend saying the cluster was deleted.
 
## Solution
Added an OnRemove handler for the cluster role and cluster binding which re-queues the deletion of the role and binding until the cluster is fully deleted.
 
### Manual Testing
Follow repro steps in issue.

### Automated Testing
Adding tests

### DIFF
--- a/pkg/controllers/management/authprovisioningv2/role.go
+++ b/pkg/controllers/management/authprovisioningv2/role.go
@@ -4,11 +4,13 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/rbac"
 	apiextcontrollers "github.com/rancher/wrangler/pkg/generated/controllers/apiextensions.k8s.io/v1"
+	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/name"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -21,8 +23,11 @@ import (
 )
 
 const (
-	clusterIndexed      = "clusterIndexed"
-	clusterIndexedLabel = "auth.cattle.io/cluster-indexed"
+	clusterIndexed        = "clusterIndexed"
+	clusterIndexedLabel   = "auth.cattle.io/cluster-indexed"
+	clusterNameLabel      = "cluster.cattle.io/name"
+	clusterNamespaceLabel = "cluster.cattle.io/namespace"
+	reenqueueTime         = time.Second * 5
 )
 
 func (h *handler) initializeCRDs(crdClient apiextcontrollers.CustomResourceDefinitionClient) error {
@@ -153,6 +158,92 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
 	return rt, nil
 }
 
+func (h *handler) hasAnnotationsForDeletingCluster(annotations map[string]string, isClusterRole bool) (bool, error) {
+	clusterName, cOK := annotations[clusterNameLabel]
+	clusterNamespace, cnOK := annotations[clusterNamespaceLabel]
+	if !cOK || (!cnOK && !isClusterRole) {
+		// if this role doesn't have a clustername/namespace label it isn't one of the protected roles, so move on
+		return false, nil
+	}
+
+	mgmtCluster, err := h.mgmtClusters.Get(clusterName)
+	if err != nil {
+		if !apierror.IsNotFound(err) {
+			return false, err
+		}
+		// if management cluster was not found, ensure it's nil
+		mgmtCluster = nil
+	}
+
+	cluster, err := h.clusters.Get(clusterNamespace, clusterName)
+	if err != nil {
+		if !apierror.IsNotFound(err) {
+			return false, err
+		}
+		// if cluster was not found, ensure cluster is nil
+		cluster = nil
+	}
+
+	if (cluster != nil && cluster.DeletionTimestamp != nil) || (mgmtCluster != nil && mgmtCluster.DeletionTimestamp != nil) {
+		// if the cluster is being deleted
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func (h *handler) OnRemoveRole(key string, role *rbacv1.Role) (*rbacv1.Role, error) {
+	shouldEnqueue, err := h.hasAnnotationsForDeletingCluster(role.Annotations, false)
+	if err != nil {
+		return role, err
+	}
+	if !shouldEnqueue {
+		return role, nil
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.roleController.EnqueueAfter(role.Namespace, role.Name, reenqueueTime)
+	return role, generic.ErrSkip
+}
+
+func (h *handler) OnRemoveRoleBinding(key string, roleBinding *rbacv1.RoleBinding) (*rbacv1.RoleBinding, error) {
+	shouldEnqueue, err := h.hasAnnotationsForDeletingCluster(roleBinding.Annotations, false)
+	if err != nil {
+		return roleBinding, err
+	}
+	if !shouldEnqueue {
+		return roleBinding, nil
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.roleBindingController.EnqueueAfter(roleBinding.Namespace, roleBinding.Name, reenqueueTime)
+	return roleBinding, generic.ErrSkip
+}
+
+func (h *handler) OnRemoveClusterRole(key string, role *rbacv1.ClusterRole) (*rbacv1.ClusterRole, error) {
+	shouldEnqueue, err := h.hasAnnotationsForDeletingCluster(role.Annotations, true)
+	if err != nil {
+		return role, err
+	}
+	if !shouldEnqueue {
+		return role, nil
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.clusterRoleController.EnqueueAfter(role.Name, reenqueueTime)
+	return role, generic.ErrSkip
+}
+
+func (h *handler) OnRemoveClusterRoleBinding(key string, roleBinding *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, error) {
+	shouldEnqueue, err := h.hasAnnotationsForDeletingCluster(roleBinding.Annotations, true)
+	if err != nil {
+		return roleBinding, err
+	}
+	if !shouldEnqueue {
+		return roleBinding, nil
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.clusterRoleBindingController.EnqueueAfter(roleBinding.Name, reenqueueTime)
+	return roleBinding, generic.ErrSkip
+}
+
 func (h *handler) objects(rt *v3.RoleTemplate, enqueue bool, cluster *v1.Cluster) error {
 	var (
 		matchResults []match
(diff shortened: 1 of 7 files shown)


pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +157,83 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+	if !shouldEnqueue {
+		return roleBinding, err
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.roleBindingController.EnqueueAfter(roleBinding.Namespace, roleBinding.Name, time.Second*5)
+	return roleBinding, generic.ErrSkip
+}
+
+func (h *handler) OnRemoveClusterRole(key string, role *rbacv1.ClusterRole) (*rbacv1.ClusterRole, error) {
+	if role.Annotations == nil {
```
MbolotSuse, line 208: You do this check for the ClusterRole/ClusterRoleBinding functions but not for the Role/Rolebinding functions. It would probably be wise to move this check to the `hasClusterAnnotationsAndDeletionTimestamp` function - that way this is consistent across the various functions.

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +157,83 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
 	return rt, nil
 }
 
+func (h *handler) hasClusterAnnotationsAndDeletionTimestamp(annotations map[string]string) (bool, error) {
```
MbolotSuse, line 160: A general comment on the name of this function - this doesn't really describe what this function is doing. The current name of the function implies that you are validating that a selection of annotations is present in the map - namely a cluster-related annotation and a deletion timestamp.

In reality, this function is to validate (through annotations and some other calls) that you have a binding that is granting access to a deleted cluster.

My suggestion would be to rename this to `isAnnotationsForDeletingCluster` (or something similar). You can also leave a comment above the function indicating the purpose of the function and why we care about this.

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +157,83 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+	clusterName, cOK := annotations[clusterNameLabel]
+	clusterNamespace, cnOK := annotations[clusterNamespaceLabel]
+	if !cOK || !cnOK {
+		// if this role doesn't have a clustername/namespace label it isn't one of the protected roles, so move on
+		return false, nil
+	}
+	cluster, err := h.clusters.Get(clusterNamespace, clusterName)
```
MbolotSuse, line 167: This function currently appears to be using the controllers for the v2 provisioning clusters. There are two different cluster objects - `clusters.provisioning.cattle.io` and `clusters.management.cattle.io`. V1 makes heavier use of the second type of cluster while v2 provisioning makes heavier use of the first type of cluster. The second type, or management clusters, are not namespaced resources, which requires cluster roles/ cluster role bindings for users to access them. 

The code that you have here will interact well with provisioning clusters but will not interact well with v1 clusters, which will cause some issues with RKE1 clusters. I would recommend re-evaluating your approach to do the following:

1. Include a cache object in the handler for the management type clusters
2. Configure the `onRemoveClusterRole` and `onRemoveClusterRoleBinding` methods to use this object to make sure that the cluster still exists. There's a couple of ways that you can do this - either by passing in a "DoesClusterExist" function as a parameter which you construct, using a boolean flag to indicate if the check should be done for mgmt or provisioning, or by just checking both clusters and making sure that neither of them could be found.

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +157,83 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+	return true, nil
+}
+
+func (h *handler) OnRemoveRole(key string, role *rbacv1.Role) (*rbacv1.Role, error) {
+	shouldEnqueue, err := h.hasClusterAnnotationsAndDeletionTimestamp(role.Annotations)
+	if !shouldEnqueue {
+		return role, err
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.roleController.EnqueueAfter(role.Namespace, role.Name, time.Second*5)
```
MbolotSuse, line 193: Nit: Might be good to use a constant for the enqueue time.

MbolotSuse (changes requested):
Some nitpicks and one larger issue with the "does cluster still exist check"

pkg/controllers/management/authprovisioningv2/role.go
steffantucker: That is a very good point

pkg/controllers/management/authprovisioningv2/role.go
steffantucker: I struggled with the name, and expected someone bringing it up, I like your suggestion more than mine.

pkg/controllers/management/authprovisioningv2/role.go
KevinJoiner: @MbolotSuse What are the benefits of checking both cluster CRDs over just the cluster CRD presented to the UI?

KevinJoiner (changes requested):

pkg/controllers/management/authprovisioningv2/role.go
MbolotSuse: Both clusters are presented to the UI, just in different places. Since they both are presented, you need to check both of them. 

This is also leaving aside automation (such as creation of v1 prov vs v2 prov clusters) which would likely need access to the "source of truth" CRD.

KevinJoiner:
I re-ran the pipeline, but it seems to be failing still.

steffantucker:
There's 1 python integration test failing between the 2 builds, and then another for each build, but it's not the same one. Are the integration tests no ran for all OS builds, and that's why they're only failing for those 2? Across all 3 failing tests, it looks like it's failing waiting for the rolebinding to delete in some fashion..

KevinJoiner:
linux-amd64 runs all of the tests the other OS builds run a subset of the test.

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +158,78 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+	}
+	if apierror.IsNotFound(mgmtErr) && apierror.IsNotFound(err) {
+		return false, nil
+	}
+	if cluster == nil && mgmtCluster == nil {
+		// cluster + err both being nil should not happen - but if it does, safe to delete the role
+		return false, nil
+	}
+	// if the err is nil, we can assume that the cluster is not nil
+	if (cluster != nil && cluster.DeletionTimestamp == nil) && (mgmtCluster != nil && mgmtCluster.DeletionTimestamp == nil) {
```
KevinJoiner, line 186: I think you may need a separate case where one cluster is nil and the other cluster is not set for deletion.
```golang
	if (cluster != nil && cluster.DeletionTimestamp == nil ) && mgmtCluster == nil  {
```

KevinJoiner (changes requested):

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +158,78 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+}
+
+func (h *handler) OnRemoveRoleBinding(key string, roleBinding *rbacv1.RoleBinding) (*rbacv1.RoleBinding, error) {
+	shouldEnqueue, err := h.hasAnnotationsForDeletingCluster(roleBinding.Annotations, false)
+	if !shouldEnqueue {
+		return roleBinding, err
+	}
+	// Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
+	h.roleBindingController.EnqueueAfter(roleBinding.Namespace, roleBinding.Name, reenqueueTime)
+	return roleBinding, generic.ErrSkip
```
KevinJoiner, line 210:
```suggestion
	shouldEnqueue, err := h.hasAnnotationsForDeletingCluster(roleBinding.Annotations, false)
         if err != nil {
             return role, err
          }
          if shouldEnqueue{
              // Enqueue to ensure perms remain until full delete. Err skips don't re-enqueue, so we do that manually
	     h.roleBindingController.EnqueueAfter(roleBinding.Namespace, roleBinding.Name, reenqueueTime)
          }
	
	return roleBinding, generic.ErrSkip
```

KevinJoiner: When using the hasAnnnotation function, we should first check the error to verify the validity of the `shouldEnqueue` variable

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +158,78 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+	clusterNamespace, cnOK := annotations[clusterNamespaceLabel]
+	if !cOK || (!cnOK && !isClusterRole) {
+		// if this role doesn't have a clustername/namespace label it isn't one of the protected roles, so move on
+		return false, nil
+	}
+
+	mgmtCluster, mgmtErr := h.mgmtClusters.Get(clusterName)
+	if mgmtErr != nil && !apierror.IsNotFound(mgmtErr) {
+		return false, mgmtErr
+	}
```
KevinJoiner, line 172:
```suggestion
	if mgmtErr != nil {
	 !apierror.IsNotFound(mgmtErr) {
		return false, mgmtErr
	}
	cluster = nil
	}
```

pkg/controllers/management/authprovisioningv2/role.go
```diff
@@ -153,6 +158,78 @@ func (h *handler) OnChange(key string, rt *v3.RoleTemplate) (*v3.RoleTemplate, e
...
+		// cluster + err both being nil should not happen - but if it does, safe to delete the role
+		return false, nil
+	}
+	// if the err is nil, we can assume that the cluster is not nil
+	if (cluster != nil && cluster.DeletionTimestamp == nil) && (mgmtCluster != nil && mgmtCluster.DeletionTimestamp == nil) {
+		// if the cluster isn't being deleted, it wasn't the source of the delete, so we can remove the role
+		return false, nil
+	}
+	return true, nil
+}
```
KevinJoiner, line 191:
```suggestion
	if (cluster != nil && cluster.DeletionTimestamp != nil) || (mgmtCluster != nil && mgmtCluster.DeletionTimestamp != nil) {
		// cluster is being deleted
		return true, nil
	}
	return false, nil
}
```

KevinJoiner (changes requested):

KevinJoiner (changes requested):
Let's Add a test case for the problem we ran into where one cluster is missing, and the other is not being deleted.

KevinJoiner (approved):

MbolotSuse (approved):
LGTM pending some minor concerns over testing/constants.

That being said, we should have QA Validate that:
- The original issue is resolved for users of differing permissions (cluster owner, cluster member, cluster view, project owner in a cluster, ect.)
- There are no left behind roles/rolebindings/cluster roles/clusterrolebindings after the cluster is deleted 
- There are no left behind roles/rolebindings/cluster roles/ clusterrolebindings if you get role of differing permissions (see above) removed from your users. I.e. if you no longer have a CRTB/PRTB to a role, there should be no left over RBAC resources and you should completely lose access.

pkg/controllers/management/authprovisioningv2/role_test.go
```diff
@@ -0,0 +1,307 @@
...
+	funcName := "handler.hasClusterAnnotationsAndDeletionTimestamp()"
+
+	tests := []struct {
+		name        string
+		annotations map[string]string
+		wantErr     bool
+		want        bool
+	}{
+		// Enqueued (returned generic.ErrSkip)
+		{
```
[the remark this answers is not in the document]
steffantucker: These are great test cases, and I should have revisited the tests more thoroughly when I added in the checks for the second cluster.

PR #38563, opened 2022-08-09
Sets the default value of the 'system-default-registry' setting.

KevinJoiner:
## Problem
Rancher needs the ability to set the default value for the system registry that is used at compile time.
 
## Solution
Search for the env during compilation which will be used to set the default value of the setting `system-default-registry`.
If the env is not present, then the setting will be an empty string which is the current default behavior, so no additional logic is needed for that use case.
I replaced the hardcoded JSON string with `jq` so that we can handle and escape malformed input.
 
## Testing
When building with the rancher with `script/build-server` verify that the default value for `system-default-registry` can be set.


cbron:
Superseded by: https://github.com/rancher/rancher/pull/38809

Issue #37324, opened 2022-04-13
Remove Pipelines


cbron:
Pipelines were deprecated in 2.5.0, and therefore will be EOL as of 2.7.0. It should be removed/disabled.

cbron:
All these "Remove" issues should also take into account docs, and other areas that reference them. For instance, pipelines have their own docs page, but they are also referenced in other places, such as the tokens page: https://rancher.com/docs/rancher/v2.6/en/api/api-tokens/

vivek-infracloud:
Based on validations above, closing the issue.

PR #38579, opened 2022-08-10
[2.7.0] Remove Pipelines from rancher

doflamingo721:
## Issue: 
https://github.com/rancher/rancher/issues/37324
 
## Problem
Pipelines were  deprecated in 2.5.0, and therefore will be EOL as of 2.7.0. It should be removed/disabled
 
## Solution
Remove pipeline components from rancher and supporting repos.

### Manual Testing
Provisioned the rancher server after removing pipelines, checked if all the pods were up, deployed a node-driver downstream cluster and installed some apps.

## QA Testing Considerations
Upgrade scenario for the rancher server to the new rancher server which has pipeline removed.


prachidamle (approved):
LGTM

PR #38598, opened 2022-08-12
fix for the failing tests due to timeout

vivek-infracloud:
## Issue: 

Some tests in the v3_additional_tests_long fails with timeout error.

```
wait_for_backup_to_active
wait_for_app_to_active
```
 
This PR serves to fix that by changing timeout condition.

## Solution

 
`timeout = start + timeout` fixes the timeout error.


izaac (commented):
LGTM 
this is a similar fix I reviewed for other methods

izaac (approved):

PR #38599, opened 2022-08-12
Add fix for test_node_annotation flaky tests

rishabhmsra:
## Issue: 

Following test cases fails from the `rancher-v3_additional_tests` jenkins job
- test_node_annotation.test_rbac_node_annotation_add_kubectl[project-owner]
- test_node_annotation.test_rbac_node_annotation_delete_kubectl[project-owner]
- test_node_annotation.test_rbac_node_annotation_edit_kubectl[project-owner]
 
## Problem

The assert condition for `project-owner` role was incorrect, hence during adding, editing and deleting the node annotation it was failing.
- Below is the assert error in jenkins logs:
`assert 'cannot get resource "nodes"' in 'Error from server (Forbidden): nodes "ip-<REDACTED>" is forbidden: User "u-jc7q7" cannot patch resource "nodes" in API group "" at the cluster scope\n'`
- When the role is project-owner the kubectl command fails for patch operation and not for the get operation.

## Solution

 Updated the condition in code so that correct assert condition is executed when the role is project-owner.

### DIFF
--- a/tests/validation/tests/v3_api/test_node_annotation.py
+++ b/tests/validation/tests/v3_api/test_node_annotation.py
@@ -634,7 +634,7 @@ def test_rbac_node_annotation_add_kubectl(role):
                                         annotation_value)
         # cleanup annotation
         delete_node_annotation(annotation_key, node, client)
-    elif role == CLUSTER_MEMBER:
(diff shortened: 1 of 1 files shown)


sowmyav27 (approved):

izaac (approved):
LGTM

PR #38602, opened 2022-08-12
Update charts and kdm to release-v2.5

rmweir:


cbron (commented):
I would make a single PR to migrate everything to the final RC. So that would include UI tags, rke tag, etc... 
That makes things much easier to track, and also revert. 
That being said the 3 things here look correct.

MbolotSuse:
@cbron I've opened #38638 with the UI tags/rke-version/operator bump. Closing this out to focus on that PR.

PR #38609, opened 2022-08-15
Adding create GKE HA cluster to the create HA test

izaac:
This will require terraform v0.13 for the gke_auth module to work. That was prepared for parameterization https://github.com/rancher/rancher/pull/31557.


slickwarren (dismissed):
approving as is based on freeform job runs, and have an enhancement ticket filed for:
* rancher on GKE v1.22+

we also need a jenkinsJobBuilder ticket/PR to use these in the UI

izaac:
@slickwarren we only need the TERRAFORM_VERSION the GKE part is in the env vars

slickwarren (approved):

PR #38612, opened 2022-08-15
Bump rancher-webhook min version to 0.2.6

KevinJoiner:


cbron (approved):
lgtm, but needs to wait for charts pr

PR #38621, opened 2022-08-16
increase timeout

vivek-infracloud:
## Problem

As per the analysis mentioned on the confluence page, **rancher-v3_ontag_az_certification** job fails to remove all the resources created for the testing. 
 
## Solution

Made changes in the timeout parameter.
This PR serves the purpose to fix this issue.


sowmyav27 (dismissed):

izaac (changes requested):
LGTM I just need a freeform run of azure provisioning to check the wait is working.
You can add a print statement too so we're sure that it's waiting and it will be useful for the logs.

I'm requesting changes just to include the print, everything else look good.

PR #38628, opened 2022-08-16
Automation framework to test terraform rancher2 provider clusters with Terratest + go

Josh-Diamond:
## Issue: 
https://github.com/rancher/qa-tasks/issues/386
 
## Problem
Currently no automation tests for rancher2 provider cluster resources;
Intended use is to automate tf rancher2 release check validation
 
## Solution
Created functions that may be called during the test, to allow main.tf to be set, updated, and tested against.
Also created test files using these functions to automate 
 
## Testing
Automation task - no repro steps

### Manual Testing
Manually ran each test, and ensured desired behavior

### Automated Testing
AKS/EKS/RKE1/RKE2/K3s - cover provisioning and scaling up/down

RKE2/K3s - additionally covers k8s upgrades

## QA Testing Considerations
View internal confluence docs on Terratest to understand how to set up and test with this repo
 
None


slickwarren:
Spoke offline with Josh, here is a summary of our conversation:
* we need a new ticket for adding this to our jenkins. All the work here is the bulk work for TF, however it would need to be ran maually now in order for testing. 
* I had some concerns about the following, which he will answer those by providing (internal) documentation and perhaps a KT/meeting once documentation is done:
  * how to use it -- specifically, what to specify when running a TF.go test vs. all other go tests
looks like you have AKS, rke1 and rke2 support. Can I use any provider with rke2? How do I specify credentials? Do I need a reancher setup?
  * how to add to TF:
  * related to bullet in how to use it -> what currently needs to be added? (i.e. custom clusters, other node drivers, etc.)
  * does each test need its own folder? If not, what other tests would go in, say tests/k3s? What else goes in tests/ folder?
  * What does maintenance look like in TF?
  * what is the naming convention? I see a lot of vars in the config.go. What will maintenance look like each release of KDM?
  * How much dependency is there on upstream TF in this PR? What is maintenance like when upstream changes (is it mostly automated, or will we have to manually change stuff when upstream changes?)


I think the above will need to be clarified before merging, this is mostly a note that any testing before Jenkins work is done will need to be done locally.

slickwarren (commented):
awaiting documentation and reasoning behind naming conventions before approval

izaac:
@caliskanugur would you review the config part of this PR ? I know you are working on configuration (parameters or config files iirc) i'd like to have consistency between this and the go framework.

tests/terratest/config/config.go
```diff
@@ -0,0 +1,98 @@
+package config
```
igomez06, line 1: I'm with Izaac, I think it would be better if you read the hardcoded values from a config file, rather than it be hardcoded in go. You can unmarshall form a json or yaml. That way it would give you more flexibility when certain values change in the config file.

tests/terratest/functions/waitForActiveCluster.go
```diff
@@ -0,0 +1,122 @@
...
+
+		return false, nil
+	})
+	require.NoError(t, err)
+	time.Sleep(4 * time.Second)
+	}
+	
+	if module != "aks" {
+			// Check cluster state is active
+	err := wait.Poll(500*time.Millisecond, 30*time.Minute, func() (done bool, err error) {
```
igomez06, line 58: Nice work Josh, much better. Can indent this line though the formatting seems off and it's hard to read since it looks like this whole block of code is under the if statement

tests/terratest/functions/cleanupConfigTF.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"os"
+)
+
+func CleanupConfigTF(module string) {
+
+	f, err := os.Create("../../modules/" + module + "/main.tf")
+
+	if err != nil {
+		fmt.Println(err)
```
igomez06, line 13: Don't print out. If you don't want to return an. error log it instead then. Printing to the command line is not very useful. So either return the error or log the statement. But I would like an explanation first why you aren't returning the error.

Josh-Diamond: Returning the error is ideal.  I will update this @igomez06 👍

tests/terratest/functions/waitForActiveCluster.go
Josh-Diamond: Sure thing!

tests/terratest/functions/waitForActiveCluster.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		if cluster.State == "active" {
+			return true, nil
+		}
+
+		return false, nil
+	})
+	require.NoError(t, err)
+	time.Sleep(4 * time.Second)
```
igomez06, line 120: Why do you have this sleep here?

Josh-Diamond: k3s specifically highlights the need to slow the test down to ensure reliability.  Without this sleep, tests are flakey and not reliable.  By adding this sleep, and allowing rancher server to "breathe", results are consistent and as expected.  Without including this sleep, the following cluster specs grabbed and tested are not always accurate.

igomez06: Sounds good, I figured there was a specific reason just wanted to make sure

PR #38638, opened 2022-08-17
2.5.16 final rc

MbolotSuse:
Version updates/changes for the final 2.5.16 rc.


MbolotSuse:
Still need rancher/charts#2034 to go in before we can merge this.

maxsokolovsky (dismissed):

package/Dockerfile
```diff
@@ -113,8 +113,8 @@ RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
     chmod +x /usr/bin/tini /usr/bin/telemetry /usr/bin/k3s /usr/bin/kubectl /usr/bin/channelserver && \
     mkdir -p /var/lib/rancher-data/driver-metadata
 
-ENV CATTLE_UI_VERSION 2.5.16-rc3
-ENV CATTLE_DASHBOARD_UI_VERSION v2.5.16-rc3
+ENV CATTLE_UI_VERSION 2.5.16-rc4
+ENV CATTLE_DASHBOARD_UI_VERSION v2.5.16-rc4
```
rmweir, line 117: Don't we need the final ui tags to cut the final rc?

rmweir (changes requested):

package/Dockerfile
maxsokolovsky: I don't think we necessarily need the final tags at this point. For v2.5.14 final RC, for example, we had `rc` in UI tags.

maxsokolovsky: Actually, I take that back - the very final RC explicitly did not have `rc` in UI tags. And, Ricky is right, the final RC and release should ideally have the same git commit hash.

MbolotSuse: Final version was published for the UI. I've updated the version in the dockerfile accodingly.

rmweir (approved):
LGTM

maxsokolovsky (approved):

PR #38682, opened 2022-08-22
[2.7.0] Remove v1 istio

doflamingo721:
## Problem

Istio v1 chart was deprecated in 2.5.0, and therefore will be EOL as of 2.7.0. It should be removed/disabled
## Solution

Remove istio v1 components from rancher and supporting repos.
### Manual Testing

Provisioned Rancher server and enabled the legacy feature flag. Select Multicluster apps -> Launch apps, istio should not be visible.
## QA Testing Considerations

 Upgrade scenario for the rancher server to the new rancher server which has v1 istio removed.


pkg/controllers/managementuserlegacy/istio/metricExpression.go
```diff
@@ -1,552 +0,0 @@
...
-func yamlToObject(yml string, obj interface{}) error {
-	jsondata, err := yaml.YAMLToJSON([]byte(yml))
-	if err != nil {
-		return err
-	}
-	return json.Unmarshal(jsondata, obj)
-}
-
-var (
-	IstioMetricsTemplate = `
```
aiyengar2, line 71: should we add in logic to clean up these resources automatically on an upgrade?

pkg/controllers/managementuserlegacy/istio/appHandler.go
```diff
@@ -1,159 +0,0 @@
...
-	if err != nil {
-		return errors.Wrapf(err, "failed to deploy istio metric expression into Cluster %s", clusterName)
-	}
-
-	logrus.Infof("deploy istio metric expression into CLuster %s successfully", clusterName)
-
-	return nil
-}
-
-func isIstioMetricExpressionWithdraw(clusterName string, istioClusterGraphClient mgmtv3.ClusterMonitorGraphInterface, istioMetricsClient mgmtv3.MonitorMetricInterface, obj *v3.App) error {
```
aiyengar2, line 98: if istio v1 is installed, should we trigger something to clean up the helm chart deployment on an upgrade?

aiyengar2 (changes requested):
Removing the controllers LGTM, but the cleanup logic for an upgrade scenario is unclear to me if the user is still using v1 Istio.

pkg/controllers/managementuserlegacy/istio/appHandler.go
mitulshah-suse: @aiyengar2 are we expecting users to manually reinstall the v2 istio chart in case of an upgrade? 
Simply triggering a cleanup might break things. Deleting configurations that were done by users as part of their istio setup could create issues.

prachidamle:
> > Removing the controllers LGTM, but the cleanup logic for an upgrade scenario is unclear to me if the user is still using v1 Istio.
> 
> I assume the user will need to manually migrate from v1 istio to v2 chart. Do we have any direct upgrade paths between the two? If we simply try cleaning up the resources as well for v1 istio, it might cause more problems. The user may end up losing all the istio configurations done on the clusters. While just removing the controllers will force a migration, it will at least not delete configurations. We can have a further discussion on this if needed, but we need to conclude how to proceed on this soon, if it needs to go in 2.7.0 @prachidamle @aiyengar2

Sounds good @mitulshah-suse

prachidamle:
We need to add a testcase to the v1 tools removal in general:
- have a cluster with the v1 tool (cis/istio) installed
- upgrade the cluster to 2.7.0
- sanity test the cluster to deploy apps/check rest of cluster functionality

prachidamle (approved):

prachidamle:
There is not direct upgrade path between v1 -> v2, and upon upgrade to 2.7.0 we shouldn’t try and clean up v1 istio
v1 Istio will be EOL in 2.7.0 and users should move to v2. @aiyengar2 @mitulshah-suse

pkg/controllers/managementuserlegacy/istio/appHandler.go
aiyengar2: noted in https://github.com/rancher/rancher/pull/38682#issuecomment-1252541229

aiyengar2 (approved):

Issue #38659, opened 2022-08-18
Update tests to have standard user passwords generated


igomez06:
Instead of having hard coded passwords for standard users across our tests, refactor to have the passwords generated. 
Create a helper method to do such, using our already existing random name generator.

markusewalker:
PR: https://github.com/rancher/rancher/pull/38710

markusewalker:
Installed.

PR #38710, opened 2022-08-24
Check if serviceAccountTokenSecret exists and remove hardcoded passwords

markusewalker:
## Issue: 

[Create test to check if serviceAccountTokenSecret parameter appears in cluster's YAML](https://github.com/rancher/qa-tasks/issues/458)

[Update tests to have standard user passwords generated](https://github.com/rancher/rancher/issues/38659
) 
## Problem

**Problem 1**
Per the release of 2.6.7, we introduced new parameter `serviceAccountTokenSecret`. This is because previously when looking in a cluster's YAML, the `serviceAccountToken` was fully exposed. Depending on the RBAC, this could be a potential security vulnerability, so `serviceAccountTokenSecret` abstracts the token completely.

As a result, there is a need in our automation to verify that when we provision any cluster that we also check that `serviceAccountTokenSecret` not only exists in the cluster's YAML, but that we verify exactly what it is.

**Problem 2**
In all of our provisioning tests, we were hardcoding the password that the admin and standard user were using to access the supplied Rancher. This isn't good practice as we should not be exposing passwords in our test framework at all; we need to abstract and randomize this as we do with the test user.
 
## Solution

**Solution 1**
`clusters.go` has been given a new function `CheckServiceAccountTokenSecret` which performs a simple check to see if the parameter `serviceAccountTokenSecret` exists in the fully provisioned cluster's YAML. If yes, return the value. If no, return that it does not exist. It does not error out in the negative case, because we should check if it's that specific test that does not contain the `serviceAccountTokenSecret`.

From there, in each of the provisioning tests, added a check after the cluster is validated to be fully stood up and `Active` to then check if the `serviceAccountTokenSecret` exists.

**Solution 2**
Implemented new `passwordgenerator/passwordgenerator.go` folder/file which will reference the `namegenerator` package to create a randomized password. Then, in each provisioning tests, removed the hardcoded password and reference the `GenerateUserPassword` function found in the `passwordgenerator.go` file.
 
### Manual Testing

Manually tested with RKE1/RKE2 node provider and custom cluster provisioning.

### Automated Testing

Jenkins runs will be run after this PR is live and will be passed directly to the PR reviewers.

### DIFF
--- a/tests/framework/extensions/clusters/clusters.go
+++ b/tests/framework/extensions/clusters/clusters.go
@@ -13,6 +13,7 @@ import (
 	provisioning "github.com/rancher/rancher/tests/framework/clients/rancher/generated/provisioning/v1"
 	"github.com/rancher/rancher/tests/framework/pkg/wait"
 	"github.com/rancher/rancher/tests/integration/pkg/defaults"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -66,6 +67,27 @@ func IsHostedProvisioningClusterReady(event watch.Event) (ready bool, err error)
 	}
 
 	return false, nil
+}
+
+// Verify if a serviceAccountTokenSecret exists or not in the cluster.
+func CheckServiceAccountTokenSecret(client *rancher.Client, clusterName string) (success bool, err error) {
+	clusterID, err := GetClusterIDByName(client, clusterName)
+	if err != nil {
+		return false, err
+	}
+
+	cluster, err := client.Management.Cluster.ByID(clusterID)
+	if err != nil {
+		return false, err
+	}
+
+	if cluster.ServiceAccountTokenSecret != "" {
+		logrus.Infof("serviceAccountTokenSecret in this cluster is: %s", cluster.ServiceAccountTokenSecret)
+		return true, nil
+	} else {
+		logrus.Infof("serviceAccountTokenSecret does not exist in this cluster!")
+		return false, nil
+	}
 }
 
 // NewRKE1lusterConfig is a constructor for a v3.Cluster object, to be used by the rancher.Client.Provisioning client.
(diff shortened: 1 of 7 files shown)


sowmyav27 (approved):

slickwarren (approved):

tests/framework/extensions/clusters/clusters.go
```diff
@@ -68,6 +68,27 @@ func IsHostedProvisioningClusterReady(event watch.Event) (ready bool, err error)
...
+		return false, err
+	}
+
+	cluster, err := client.Management.Cluster.ByID(clusterID)
+	if err != nil {
+		return false, err
+	}
+
+	if cluster.ServiceAccountTokenSecret != "" {
+		fmt.Printf("serviceAccountTokenSecret in this cluster is: %s\n", cluster.ServiceAccountTokenSecret)
```
igomez06, line 84: Markus no print statements please. Use this package for logging `logrus.Infof()`

igomez06 (commented):
Overall looks good, but please remove any fmt.Println and use logrus.Info for logging instead.

tests/framework/extensions/clusters/clusters.go
```diff
@@ -68,6 +69,27 @@ func IsHostedProvisioningClusterReady(event watch.Event) (ready bool, err error)
...
+		return false, err
+	}
+
+	cluster, err := client.Management.Cluster.ByID(clusterID)
+	if err != nil {
+		return false, err
+	}
+
+	if cluster.ServiceAccountTokenSecret != "" {
+		logrus.Info("serviceAccountTokenSecret in this cluster is: %s\n", cluster.ServiceAccountTokenSecret)
```
igomez06, line 85: use `Infof` Markus, so you can pass the string like how you are doing now, and also you don't need "\n"

markusewalker: Sure thing, thanks!! Oh okay, I had the `\n` just so that I would push the `===CONT` to the next line but I can remove it

markusewalker: Good to go now. That's a neat package! Was unaware of it....

igomez06 (approved):

slickwarren (approved):

PR #38726, opened 2022-08-25
Fixed Duration Struct in the Generated V1 Provisioning Client

igomez06:
## Issue: 
 When the structs for the v1 Provisioning client were generated, Duration was coming up as empty struct. This was because of the original struct was using protobuf instead of json to unmarshall the API response.
 
## Problem
 When using the provisioning to do a get or list, it would return an unmarshall error. 
 
## Solution
 Fixed the order of importing the appropriate structs. Also included some v3 updates, as the generation code is all together.
 
## Testing
 Locally provisioned a cluster, and did a list call.

### DIFF
--- a/tests/framework/clients/rancher/generated/management/v3/zz_generated_aci_network_provider.go
+++ b/tests/framework/clients/rancher/generated/management/v3/zz_generated_aci_network_provider.go
@@ -1,134 +1,156 @@
 package client
 
 const (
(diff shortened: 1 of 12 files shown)


anupama2501 (approved):

bmdepesa (approved):

caliskanugur (approved):
LGTM

PR #38729, opened 2022-08-25
Update aks_nlb.yml and eks_nlb.yml to fix AKS and EKS local clusters in Jenkins job

markusewalker:
## Issue: 

[Fix AKS and EKS local cluster failures in the rancher-v3_ha_deploy Jenkins job](https://github.com/rancher/qa-tasks/issues/473
) 
## Problem

When specifying AKS or EKS as the local cluster to have a hosted Rancher setup, this fails with internal Jenkins job. GKE, however, succeeds fine without issues. Specifically, the AKS and EKS clusters do provision successfully, but they do not stand up Rancher.

Looking deeper into it, the culprit was that the networking was outdated. GKE works because it uses updated Ingress versions as well as other manifests.
 
## Solution

Copying the `gke_nlb.yml` into both `aks_nlb.yml` and `eks_nlb.yml` resolves this issue. Looking at each provider's respective `nlb.yml` file, there were no differences; they were exactly identical to one another. As such, I have done the same here.

If in the future there are provider-specific issues, we will adjust that providers YAML file as needed.
 
### Manual Testing

Locally, tested that I was able to successfully run the `test_create_ha.py` function with AKS and EKS clusters selected. The clusters provisioned successfully and I was able to access Rancher. Validated that the local cluster was indeed the AKS/EKS cluster.

### Automated Testing

Jenkins freeform jobs will be given offline to the reviewers of this PR.

### DIFF
--- a/tests/validation/tests/v3_api/resource/aks_nlb.yml
+++ b/tests/validation/tests/v3_api/resource/aks_nlb.yml
@@ -1,240 +1,254 @@
-
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: ingress-nginx
   labels:
-    app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/instance: ingress-nginx
-
+    app.kubernetes.io/name: ingress-nginx
+  name: ingress-nginx
 ---
-# Source: ingress-nginx/templates/controller-serviceaccount.yaml
+apiVersion: v1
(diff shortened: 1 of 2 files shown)


izaac (approved):
LGTM

sowmyav27 (approved):

slickwarren (approved):

PR #38737, opened 2022-08-26
refactoring of timeout change for app_to_active.

vivek-infracloud:
## Problem

After changes in timeout parameter the wait_for_app_to_active test kept running for many hours. Hence this PR serves to refactor those changes. The changes do not affect any other test.


izaac (approved):
Reviewed the upgrade job execution and is now finishing in time.
This is only reverting the helper function to its original behavior.

LGTM

PR #38755, opened 2022-08-29
Bump rancher/dashboard to v2.6.8-rc2

aiyengar2:


rancherbot (approved):

prachidamle (approved):

package/Dockerfile
```diff
@@ -162,7 +162,7 @@ RUN curl -sLf ${!TINI_URL} > /usr/bin/tini && \
     mkdir -p /var/lib/rancher-data/driver-metadata
 
 ENV CATTLE_UI_VERSION 2.6.7
-ENV CATTLE_DASHBOARD_UI_VERSION v2.6.7
+ENV CATTLE_DASHBOARD_UI_VERSION v2.6.8-rc2
```
prachidamle, line 165: Do we need to bump CATTLE_UI_VERSION - is it ok to have UI and Dashboard point at different tags?

prachidamle: Confirmed offline by @rak-phillip that these can be different tags since there is no functionality change for UI

rak-phillip: See also `v2.5.14` for a previous example of this

https://github.com/rancher/rancher/commit/52a8de7b63b181188cf4aeeacb59c79545af7312

Issue #38753, opened 2022-08-29
[Azure AD] Properly parse annotations from UI when admins edit Auth Config in the UI


maxsokolovsky:
**Rancher Server Setup**
- Rancher version: v2.6.7

**Describe the bug**
When users edit an existing Azure AD setup by using the UI, Rancher shows an error from Microsoft - `The 'resource' request parameter is not supported.`

Rancher needs to handle a special case to parsing Auth Configs on setup attempts. On Azure AD setup, Rancher reads the Auth Config from the UI. The backend expects to find the annotation that instructs it to use the new authentication flow via Microsoft Graph. For new setups, Rancher always properly picks the new flow because it does not even care about the annotation, since the provider is new, it is not enabled. But if Azure AD is already enabled, and admins try to edit the Auth Config in the UI (thus reconfiguring it), then the backend needs to properly parse the annotation from the UI and use it to pick the new flow.

**To Reproduce**
1. Set up Azure AD in v2.6.7+.
1. Click the Edit Config button and re-enter your config information (the UI requires only the Application Secret to be submitted again).

**Result**
Rancher shows an error from Microsoft - `The 'resource' request parameter is not supported.`

**Expected Result**
Rancher re-saves the configuration without errors.

**Screenshots**

**Additional context**

anupama2501:
QA to validate:


Fresh Install
---
Verify edit azure AD config, add in secret key, save → No error is seen → Log out and login → Add new users → Login as those users
Verify edit config by changing the end points to custom config, save →  No error seen → Log out and login → remove users → Login as those users
Terraform use case?

Upgrade cases
--
Standard end points
Enable AD, Upgrade from 2.5.16 >> 2.6.9, edit the auth configs >>  Logout and relogin >> verify no errors seen
Enable AD, upgrade from 2.6.7 >> 2.6.9, edit the auth configs >>  Logout and relogin >> verify no errors seen.
Enable AD, upgrade from 2.6.7 >> 2.6.9, edit the auth configs >> change it to custom end points >> save >> Logout and relogin >> verify no errors seen

anupama2501:
Reproduced on v2.6.6 >> 2.6-head `e2b5939` and v2.6-head `e2b5939`

Fresh install:
1. Enable azure AD >> Verify new endpoints are updated
2. From auth providers page >> edit the configs >> add in the secret >> Save fails with the error `AADSTS901002: The 'resource' request parameter is not supported.`

maxsokolovsky:
[...]

PR #38757, opened 2022-08-29
Properly parse annotations from UI when admins edit Auth Config in the UI

maxsokolovsky:
Issue: https://github.com/rancher/rancher/issues/38753

This addresses a corner case where the auth config is enabled but is re-registered (by Editing it in the UI).
Rancher now properly parses the annotations sent from the UI and correctly decides when to use the new auth flow.

### DIFF
--- a/pkg/auth/providers/azure/azure_provider.go
+++ b/pkg/auth/providers/azure/azure_provider.go
@@ -392,7 +392,7 @@ func formAzureRedirectURL(config map[string]interface{}) string {
 		// Extract the annotations from the map. This is needed because of the type structure of
 		// the Azure config and the Auth config it embeds. Full deserialization does not work for
 		// fields of the embedded Kubernetes types in this case.
-		ac.ObjectMeta.Annotations = extractAnnotations(config)
+		ac.ObjectMeta.Annotations = extractAnnotationsFromAuthConfig(config)
 		if !isConfigDeprecated(&ac) {
 			// Return the redirect URL for Microsoft Graph.
 			return fmt.Sprintf(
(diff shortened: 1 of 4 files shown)


KevinJoiner (changes requested):
Code LGTM just needs tests.

pkg/auth/providers/azure/azure_provider.go
```diff
@@ -415,23 +415,31 @@ func formAzureRedirectURL(config map[string]interface{}) string {
...
+// In the latter, it's "annotations.[map of annotations]". The function tries to find the annotations in either structure.
+func extractAnnotationsFromAuthConfig(config map[string]interface{}) map[string]string {
+	if metadata, ok := config["metadata"].(map[string]interface{}); ok {
+		return parseAnnotations(metadata)
+	}
+	logrus.Info("Failed to decode the 'metadata' field of the AuthConfig. Attempting to decode 'annotations' at the top level.")
+	return parseAnnotations(config)
+}
+
+func parseAnnotations(m map[string]interface{}) map[string]string {
```
MbolotSuse, line 431: Would you mind changing the parameter here so that it is more clear what specific type of `m` you are expecting passed in? Something like `parseAnnotations(metadata map[string]interface{})` would work nicely.

MbolotSuse, line 431: Would you mind changing the naming of this parameter so that it is more clear what specific type of `m` you are expecting passed in? Something like `parseAnnotations(metadata map[string]interface{})` would work nicely. Might be a bit clunky since this is either a config or a metadata, but something to indicate the general structure of what is expected would be good.

MbolotSuse (changes requested):
As @KevinJoiner pointed out, will need tests.

Outside of that, I've got one nitpick on the name of a function parameter.

Besides those two concerns looks good.

KevinJoiner (approved):

pkg/auth/providers/azure/azure_provider_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"github.com/rancher/norman/api/writer"
+	"github.com/rancher/norman/types"
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	managementschema "github.com/rancher/rancher/pkg/schemas/management.cattle.io/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestConfigureTest inspects the Redirect URL during Azure AD setup.
+func TestConfigureTest(t *testing.T) {
```
MbolotSuse, line 18: There doesn't appear to be a test for our new functionality to extract annotations when they are provided directly. Would you mind adding a test which takes the annotations from the top level?

MbolotSuse: As a side note: there's other test cases that would be good to have here as well like:
- What happens if there's a key for annotations but it isn't of type map[string]interface{}?
- What happen if we have our target key but it's value isn't a string type?

That being said, these aren't the main purpose of this PR so they are more of nice-to-haves.

MbolotSuse (changes requested):
Looking for at least one test for annotations that are at the root level instead of nested (i.e. a test which validates the main goal of the PR).

pkg/auth/providers/azure/azure_provider_test.go
maxsokolovsky: Yes, there is a problem with test setup for this case. With stubbing the Norman schemas for the test. I did spend time on trying to chase that down, but could not figure out why having no `metadata` field causes Norman's schema validation to discard annotations as a superfluous field. This does not happen when the code runs outside of tests.

maxsokolovsky: Met offline to clear up the issue. Thank you, Michael!

pkg/auth/providers/azure/azure_provider_test.go
```diff
@@ -0,0 +1,192 @@
...
+				"applicationId":     "app123",
+				"applicationSecret": "secret123",
+				"rancherUrl":        "https://myrancher.com",
+			},
+			expectedRedirectUrl: "https://login.microsoftonline.com/tenant123/oauth2/authorize?client_id=app123&redirect_uri=https://myrancher.com&resource=https://graph.windows.net/&scope=openid",
+		},
+	}
+
+	var provider azureProvider
+	for _, test := range tests {
```
MbolotSuse, line 183: Nit: There doesn't appear to be tests for the disabled auth config (old or new) as far as the "Get from K8s" case goes.

MbolotSuse (approved):
There are some minor test cases missing (retrieve from k8s but the auth config isn't enabled). I think it's ok though since the relevant functionality is generally covered by the other cases, so I won't hold up the PR for that.

pkg/auth/providers/azure/azure_provider_test.go
maxsokolovsky: I am not sure what the use case would be for fetching a disabled auth config on login.

MbolotSuse: If I'm understanding correctly, that's not what this function is testing. This is testing the `TransformToProvider` method. Rancher might use that only on login, but it's an exported method. So you can't rely on that only being called on login.

KevinJoiner (approved):

MbolotSuse (approved):
LGTM

PR #38759, opened 2022-08-29
remove_psp_in_cis_mode_for_k3s_v1.25

ShylajaDevadiga:

### DIFF
--- a/tests/validation/tests/v3_api/resource/terraform/k3s/master/install_k3s_master.sh
+++ b/tests/validation/tests/v3_api/resource/terraform/k3s/master/install_k3s_master.sh
@@ -20,7 +20,24 @@ fi
 
 if [[ -n "${8}" ]] && [[ "${8}" == *"protect-kernel-defaults"* ]]
 then
-  cat /tmp/cis_masterconfig.yaml >> /etc/rancher/k3s/config.yaml
+  mkdir -p /var/lib/rancher/k3s/server/manifests
+  if [[ "${8}" == *"selinux"* ]]
+  then
+    yum -y install bc
+  fi
+  calc(){ awk "BEGIN { print "$*" }"; }
+  version=`echo ${4}|cut -c2-5`
+  conversion=$(calc $version*1)
+  if (( $(echo "$conversion >= 1.25" | bc -l) ))
+  then
+    sed -i "s/enforce: \"privileged\"/enforce: \"${12}\"/g" /tmp/cluster-level-pss.yaml
+    cat /tmp/cis_v125_masterconfig.yaml >> /etc/rancher/k3s/config.yaml
+    cat /tmp/v125_policy.yaml > /var/lib/rancher/k3s/server/manifests/policy.yaml
+    cat /tmp/cluster-level-pss.yaml > /var/lib/rancher/k3s/server/cluster-level-pss.yaml
+  else
+    cat /tmp/cis_masterconfig.yaml >> /etc/rancher/k3s/config.yaml
+    cat /tmp/policy.yaml > /var/lib/rancher/k3s/server/manifests/policy.yaml
+  fi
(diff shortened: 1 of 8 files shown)


rancher-max (commented):
Is it possible to strip out the PSP stuff from the original `policy.yaml` into its own file, and then instead of having a separate `v125_policy.yaml`, we just use `policy.yaml` for both and then apply the different PSP/PS yaml for v1.24-/v1.25+ depending on installed version?

Also, I wonder if `kube-apiserver-arg` of `'enable-admission-plugins=NodeRestriction,NamespaceLifecycle,ServiceAccount'` needs to include `PodSecurity`? I see it's enabled by default in v1.25 per https://kubernetes.io/docs/concepts/security/pod-security-admission/. It might also be enabled by default as far back as v1.22. It should at least be acceptable as an admission plugin there. 

This will get a little bit more confusing though because in 1.22, 1.23, and 1.24 we should still need to pass the PodSecurityPolicy admission plugin for the PodSecurityPolicy manifests in the policy.yaml to apply successfully I think.

ShylajaDevadiga:
PodSecurityAdmission is enabled by default. So I believe we don't need to include the PodSecurity plugin in the kube-apiserver-arg explicitly and we pass the configuration file used by the Pod Security Admission Controller to set different PodSecurity standards like privileged, baseline or restricted.

tests/validation/tests/v3_api/resource/terraform/k3s/master/install_k3s_master.sh
```diff
@@ -20,16 +20,24 @@ fi
 
 if [[ -n "${8}" ]] && [[ "${8}" == *"protect-kernel-defaults"* ]]
 then
-  cat /tmp/cis_masterconfig.yaml >> /etc/rancher/k3s/config.yaml
+  mkdir -p /var/lib/rancher/k3s/server/manifests
+  if [[ "${4}" == *"v1.25"*  ]]
```
galal-hussein, line 24: we should update this to include any new minor version, i.e " >= 1.25"

galal-hussein (commented):
just one comment

dereknola (approved):

tests/validation/tests/v3_api/resource/terraform/k3s/master/install_k3s_master.sh
```diff
@@ -20,16 +20,31 @@ fi
 
 if [[ -n "${8}" ]] && [[ "${8}" == *"protect-kernel-defaults"* ]]
 then
-  cat /tmp/cis_masterconfig.yaml >> /etc/rancher/k3s/config.yaml
+  mkdir -p /var/lib/rancher/k3s/server/manifests
+  if [[ "${8}" == *"selinux"* ]]
+  then
+    yum -y install bc
```
rancher-max, line 26: We should only install this based on OS, not necessarily if the selinux param is passed in config.

tests/validation/tests/v3_api/resource/terraform/k3s/master/join_k3s_master.sh
```diff
@@ -19,16 +19,31 @@ fi
 
 if [[ -n "${10}" ]] && [[ "${10}" == *"protect-kernel-defaults"* ]]
 then
-  cat /tmp/cis_masterconfig.yaml >> /etc/rancher/k3s/config.yaml
+  mkdir -p /var/lib/rancher/k3s/server/manifests
+  if [[ "${10}" == *"selinux"* ]]
+  then
+    yum -y install bc
```
rancher-max, line 25: Same here: We should only install this based on OS, not necessarily if the selinux param is passed in config.

rancher-max (changes requested):

ShylajaDevadiga:
Closing in favor on new PR

Issue #38699, opened 2022-08-23
[BUG] Unable to interact with downstream clusters after upgrading from 2.5.15 to 2.5.16


junkiebev:
**Rancher Server Setup**
- Rancher version: 2.5.16
- Installation option: Helm Chart
   - RKE1, 1.20.4

**Information about the Cluster**
- Kubernetes version: 1.20.4
- Cluster Type: Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):
* Hosted =AKS

**User Information**
- What is the role of the user logged in? Admin

**Describe the bug**
Post upgrade 2.5.15 to 2.5.16, we are unable to communicate to any downstream AKS clusters (immediate 503 errors) though they appear green in the UI

**To Reproduce**
~Have registered AKS clusters in _any_ version of k8s. Upgrade 2.5.15 to 2.5.16~

1. Run rancher v2.6.3
2. Navigate to <server-url>/g/clusters/add/launch/azureaks
3. Create AKSv1 cluster using the UI.
4. Wait for cluster to become active.
5. Upgrade to v2.6.7.
6. Try to access workloads in system project of AKSv1 cluster. Agent will be outdated.
7. View rancher logs. Should see logs stating it is unavailable.

**Result**
When clicking on the cluster in Cluster Manager, an immediate 503 results
![image](https://user-images.githubusercontent.com/6684332/186246831-616dff65-f8e2-4cf0-b833-584bda50b972.png)

**Expected Result**
Cluster able to be interacted with

**Additional context**
I believe this has to do with changes to the cluster spec in 2.5.16 - all other providers (RKE on Azure, vSphere, custom, etc) clusters work, but it looks like the secrets migrator is tripping over itself.
`kubectl get clusters.management.cattle.io c-CLUSTERID` shows a `status.serviceAccountToken` rather than `status.serviceAccountTokenSecret` and fleet has all sorts of authentication errors.

Rancher support case 00359224 for reference.

junkiebev:
Rancher support provided me a script to fix that - it's in the case I referenced.

rmweir:
This was merged prior to 2.7 branching.

rmweir:
Release note:
* A bug has been fixed that resulted in certain clusters being rendered unavailable after upgrade. This was due to improper migration of certain cluster fields to secrets. The known cluster types this affected were legacy AKS, GKE, and EKS clusters but may have affected other types of clusters as well.

[...]

PR #38760, opened 2022-08-29
Fix migrator

rmweir:
## Issue: 

 https://github.com/rancher/rancher/issues/38699

## Problems

Previously, there were a few problems with how objects were being passed and used. Conditions were being set on clusters, but the change was not committed with a client Update call.

A Do function would pass in one pointer but then return a different pointer value pointing to a more updated cluster object. The code would then proceed to use the outdated object. This would cause the service account token secret migration update to fail here: https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/secretmigrator/clusters.go#L661. This was not a problem for some types of clusters like AKSV2 clusters because there is redundant code for migrating the service account token in their respective sync handlers and the migration code in clusters.go would short circuit before the problematic code was hit (see: https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/aks/aks_cluster_handler.go#L214 and https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/management/secretmigrator/clusters.go#L650).
    
Modifications were made to the object referenced by the pointer passed into the sync function. This is bad since this object will be passed to subsequent handlers in the case that a nil cluster object is returned. Now, the originally referenced object is not modified.

## Solutions

Now, condition changes are followed by an update.

Now, the returned object and object that was passed in are set to have the same pointer value which is up to date.

Now, the originally referenced object is not modified.

## Testing

To reproduce this issue:
1. Run rancher v2.6.3
2. Navigate to `<server-url>/g/clusters/add/launch/azureaks`
3. Create AKSv1 cluster using the UI.
4. Wait for cluster to become active.
5. Upgrade to v2.6.7.
6. Try to access workloads in system project of AKSv1 cluster. Agent will be outdated.
7. View rancher logs. Should see logs stating it is unavailable.

### Manual Testing

Followed the above steps and ensured expected behavior has been restored.

### Automated Testing

Added unit tests to ensure cluster is being updated by client, conditions are being set, secrets are migrated.

## QA Testing Considerations

QA should tests that the upgrade works as expected for AKSV1 and AKSV2 clusters.

### DIFF
--- a/pkg/controllers/management/secretmigrator/clusters.go
+++ b/pkg/controllers/management/secretmigrator/clusters.go
@@ -47,634 +47,25 @@ const (
 var PrivateRegistryQuestion = regexp.MustCompile(`rancherKubernetesEngineConfig.privateRegistries[[0-9]+].password`)
 var VcenterQuestion = regexp.MustCompile(`rancherKubernetesEngineConfig.cloudProvider.vsphereCloudProvider.virtualCenter\[.+\].password`)
 
-func (h *handler) sync(key string, cluster *v3.Cluster) (runtime.Object, error) {
+func (h *handler) sync(_ string, cluster *v3.Cluster) (runtime.Object, error) {
 	if cluster == nil || cluster.DeletionTimestamp != nil {
 		return cluster, nil
 	}
+	// some cluster will have migrated set to true and then short circuit from entire block
 	if apimgmtv3.ClusterConditionSecretsMigrated.IsTrue(cluster) && apimgmtv3.ClusterConditionServiceAccountSecretsMigrated.IsTrue(cluster) {
 		logrus.Tracef("[secretmigrator] cluster %s already migrated", cluster.Name)
 		return cluster, nil
 	}
-	obj, err := apimgmtv3.ClusterConditionSecretsMigrated.Do(cluster, func() (runtime.Object, error) {
-		// privateRegistries
-		if cluster.GetSecret("PrivateRegistrySecret") == "" {
-			logrus.Tracef("[secretmigrator] migrating private registry secrets for cluster %s", cluster.Name)
-			regSecret, err := h.migrator.CreateOrUpdatePrivateRegistrySecret(cluster.GetSecret("PrivateRegistrySecret"), cluster.Spec.RancherKubernetesEngineConfig, cluster)
-			if err != nil {
-				logrus.Errorf("[secretmigrator] failed to migrate private registry secrets for cluster %s, will retry: %v", cluster.Name, err)
-				return nil, err
-			}
-			if regSecret != nil {
-				logrus.Tracef("[secretmigrator] private registry secret found for cluster %s", cluster.Name)
-				cluster.Spec.ClusterSecrets.PrivateRegistrySecret = regSecret.Name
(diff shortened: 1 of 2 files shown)


jakefhyde (dismissed):
LGTM

pkg/controllers/management/secretmigrator/clusters_test.go
```diff
@@ -0,0 +1,175 @@
+package secretmigrator
```
KevinJoiner, line 1: I would add some error-handling test cases. And maybe a couple more for some of the other secrets.

KevinJoiner (changes requested):

pkg/controllers/management/secretmigrator/clusters_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"time"
+
+	apimgmtv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	fakes2 "github.com/rancher/rancher/pkg/generated/norman/core/v1/fakes"
+	"github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3/fakes"
+	rketypes "github.com/rancher/rke/types"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/api/core/v1"
+	apierror "k8s.io/apimachinery/pkg/api/errors"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jakefhyde, line 18: Purely s stylistic choice but I've been naming collisions like this `metav1`

pkg/controllers/management/secretmigrator/clusters.go
```diff
@@ -1051,104 +442,771 @@ func (m *Migrator) Cleanup(secretName string) error {
...
+		clusterCopy, err = h.clusters.Update(clusterCopy)
+		if err != nil {
+			return clusterResult, err
+		}
+		clusterResult = clusterCopy.DeepCopy()
+		// clusterCopy is returned here since it's value will be passed and fields modified without an update
+		return clusterCopy, err
+	})
+	if err != nil {
+		return clusterCopy, err
```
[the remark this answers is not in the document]
rmweir: Good catch. I think something along those lines will do. However, we don't want to return an empty struct. It needs to be nil or up to date. If it's not nil then it will be passed to subsequent handlers.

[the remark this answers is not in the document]
rmweir: Looks like if the object is nil, do just returns the object that was passed in anyway. Due to this I opted into just using the passed in cluster pointer var as the clusterResult var. This will cause the original cluster to be returned in the case that no updates were made. I also change it so even if do returns an err, an update is attempted using the returned obj. This is because Do is written with the intention that the updated will be committed since if an error is returned that means the condition is set to False.

pkg/controllers/management/secretmigrator/clusters_test.go
rmweir: Good point, fixed.

KevinJoiner (approved):
LGTM

jakefhyde (approved):

Issue #16061, opened 2018-10-10
[server-chart] Error deploying chart on GKE (evaluating ingress spec)


deniseschannon:
ngres sevice reports on GKE: Error during sync: error while evaluating the ingress spec: service "rancher/rancher" is type "ClusterIP", expected "NodePort" or "LoadBalancer"

Manual updating the created service rancher/rancher to be type NodePort resolves this

have GKE cluster
deploy helmchart (optionally do some value updates)
check GKE dashboard, and notice the rancher ingress stays in state "creating"
Show detail of said ingress and see the error

robcharlwood:
👍 +1 on this. Specifically I think this is related to running the GKE ingress rather than ``nginx`` or ``traefik``. I have raised a PR to allow that allows the service type to be configured by the user. Default behaviour of ClusterIP remains if not explicitly overridden. 😄

KevinJoiner:
@JonCrowther Can you rerun this test with the new helm setting `service.type` set. The new rancher install command would look like this. `helm install rancher ./build/chart/rancher --namespace cattle-system --create-namespace --set global.cattle.psp.enabled=false --set hostname=jono.cp-dev.rancher.space --set bootstrapPassword=admin --set service.type: NodePort`

anupama2501:
Closing the ticket as the validations look good.

JonCrowther:
# Release Notes
## Synopsis of issue

The default type for `cattle-system/rancher` is ClusterIP. When using the default GKE ingress in a cluster not using VPC-native cluster mode (ie `--no-enable-ip-alias` is set) it is incompatible with the ClusterIP type.

## Resolved (or not):

Resolved by adding a new helm setting `--set service.type: NodePort`

## Status:

resolved in v2.7.5

## Versions:

< v2.7.5

Issue #28961, opened 2020-09-17
Error from server (BadRequest): Upgrade request required with ALB 


autotune:
**What kind of request is this (question/bug/enhancement/feature request):**
BUG

**Steps to reproduce (least amount of steps as possible):**
Set up with ALB in Kubernetes

**Result:**
kubectl exec -it echoheaders-85bbd669fd-v7b2p -- bash

Error from server (BadRequest): Upgrade request required

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported): EKS
- Machine type (cloud/VM/metal) and specifications (CPU/memory): t2.large
- Kubernetes version (use `kubectl version`): 1.14.8

zaggash:
https://github.com/rancher/rancher/issues/17963

maradwan:
https://dev.to/aws-builders/kubectl-execport-forward-with-aws-alb-and-nginx-ingress-controller-4fn0

Issue #17963, opened 2019-02-11
Upgrade request required when using kubectl exec on a cluster when using single install behind Amazon ALB


superseb:
**What kind of request is this (question/bug/enhancement/feature request):**
bug

**Steps to reproduce (least amount of steps as possible):**
* Run Rancher in single install (`docker run`) and put ALB in front

**Result:**
`kubectl exec` fails on managed clusters using embedded kubectl and externally using generated kubeconfig with:

```
Error from server (BadRequest): Upgrade request required
```

**Other details that may be helpful:**
UI itself reports no issues when visited

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI):v2.1.6
- Installation option (single install/HA):single

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):Custom
- Machine type (cloud/VM/metal) and specifications (CPU/memory):EC2 instance

gsoyka:
For what it's worth, we're seeing the same error when using Helm locally as well:
```
~ $ helm version
Client: &version.Version{SemVer:"v2.13.0", GitCommit:"79d07943b03aea2b76c12644b4b54733bc5958d6", GitTreeState:"clean"}
Error: forwarding ports: error upgrading connection: Upgrade request required
```

tylfin:
Trying to solve this for myself, I think you're running into this https://stackoverflow.com/questions/47884666/kubectl-exec-command-failing-with-upgrade-request-required/47890411

elixxx:
> Hello
> 
> We are facing to this problem (ALB in front of rancher. WS works well through the UI but not with kubectl: Upgrade request required ).
> 
> Do you have any information about this issue ? :)
> 
> BR

Having the same Issue with aws ALB

TM-Developers:
hi guys, was there any developments to fix this? Thanks

renperez:
@superseb is there any update on this issue? I am running rancher in eks and using alb ingress and having the same issue. i am running version 2.3.5.

gricuk:
Hi, are there any updates?

s50600822:
I got same issue, verbose logs
`kubectl exec -it --v=9  ... bash`
shows
`69816 helpers.go:219] server response object: [{
  "metadata": {},
  "status": "Failure",
  "message": "Upgrade request required",
  "reason": "BadRequest",
  "code": 400
}]`

kjflaherty:
We have the same issue

[...]

PR #38772, opened 2022-08-30
feat: Allows configuration of the Service

robcharlwood:
## Issue: 
https://github.com/rancher/rancher/issues/16061
 
## Problem
When deploying to GKE using the chart provided service, the load balancer fails to deploy. I specifically noticed this when running the setup below.

* Private GKE cluster
* Using the GKE (Google) Ingress  (Not ``nginx`` or ``trafeik``.)
* Binding the ingress with a static IP address and DNS record already provisioned by terraform using extra annotations (https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-xlb#static_ip_addresses_for_https_load_balancers)
* Using Google Managed TLS Certificates using extra annotations (https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs#setting_up_a_google-managed_certificate)
* Terminating SSL at load balancer

Basically it seems that Google Managed certificates / GKE ingress can only run with NodePort services. Currently the chart defaults to ``ClusterIP`` and doesn't give you any manner of overriding this.

We also had problems applying IAP (Identity Aware Proxy) and service timeouts (google kills connections after 30 seconds by default and this breaks websockets) to our Rancher service. This is because the current rancher ``service.yaml`` file does not allow for customisation of service annotations.
 
## Solution
This PR contains a fix to allow the ``type`` attribute on the service to be provided by the user and to utilise that if it is passed. If it is not passed, the current behaviour will apply and a ``ClusterIP`` will be used.

It also allows a user to pass a block of service annotations to apply to the service. If no service annotations are passed, then the current behaviour will apply an no annotations will be provided for the service.
 
### Manual Testing
* Ran the changes above in a local check out of the chart and deployed successfully using a ``NodePort`` service type.

### Automated Testing
This PR has added helm tests for the feature added.

No regressions should be introduced here since the current behaviours will persist unless specifically overridden.


gecBurton (dismissed):
looks sensible

oncipriani:
This is a **very** welcome change, but it is missing some way to add extra annotations to the Service (like the `ingress.extraAnnotations` field).

This would make it possible to use Amazon NLB without having to create a new Service manually, and would allow a good workaround for some issues:

- https://github.com/rancher/rancher/issues/17963
- https://github.com/rancher/rancher/issues/28961

robcharlwood:
Thanks @oncipriani - I'll add a few extra bits to the PR to make this possible. 👍

robcharlwood:
@oncipriani hey there! So I have made the additional changes requested to allow the service to be configured using user provided annotations. 👍 😄

oncipriani (approved):
I tested this changes using `helm template` with a `values.yml` from our own Rancher installation and these changes are exactly what we needed.

robcharlwood:
@oncipriani @ibuildthecloud - Bump on this please! 😄 👍

robcharlwood:
Any ideas when this might get merged please? 😄  @oncipriani @ibuildthecloud

oncipriani (approved):

oncipriani:
This PR requires at least 1 approving review from someone with write
access to the repository. I have already approved it, but I do not
have write access.

Maybe @jiaqiluo could help?


Em qui., 5 de jan. de 2023 às 04:46, Rob Charlwood
***@***.***> escreveu:

oncipriani:
This PR is still **very** relevant!

rmweir (approved):
LGTM

KevinJoiner (approved):
👍🏾

KevinJoiner:
@robcharlwood We will target this PR for our Q2 release and merge after v2.7.2 is released. Thanks for updating this and adding the test.

robcharlwood:
> @robcharlwood We will target this PR for our Q2 release and merge after v2.7.2 is released. Thanks for updating this and adding the test.

Thank you! and my pleasure! Glad I could help!

Issue #38754, opened 2022-08-29
[Backport v2.6.9][Azure AD] Always check access token from Microsoft Graph


maxsokolovsky:
backport of: https://github.com/rancher/rancher/issues/38880

It would be good to add this guardrail to prevent users from losing access to Rancher via Azure AD or getting their Azure AD setup in an invalid state. This can happen if they don't follow the documentation in setting the necessary permissions for their App registrations in Azure for Rancher v2.6.7+.

On initial Azure AD setup or subsequent login or search attempts, always check the access token to Microsoft Graph if it does not have the necessary permissions.

anupama2501:
Reproduced the issue on v2.6.7
From azure app, removed the permission Microsoft graph permission`Group.Read.All` 
Enable Azure AD from the rancher server, and verify the `azuread-access-token` is created even with missing permissions in the local cluster >> ns: cattle-global-data.

anupama2501:
QA to validate:
Fresh install:
1. Remove the Microsoft graph permissions and enable Azure AD from 2.6-head, verify the local cluster >> cattle-global-data do not cache the secret `azuread-access-token`
2. Enable azure AD with all the required permissions, then remove the configs, edit the azure AD and add in the secret, verify the azure ad config errors out. 
3. Enable azure ad with all the required permissions, verify the secret token is cached
4. Disable azure AD after step3, remove the permissions and verify the secret token is not cached
5. Try enabling azure AD via terraform without the required perms and verify it errors out and no secret is cached. 

Upgrade:
1. From 2.6.7, try to enable azure ad without the permissions, ensure the secret is cached, upgrade rancher and verify the token is deleted. Re try enabling the azure AD config and verify the secret is not cached.

anupama2501:
On v2.6-head 79a5328 after enabling Azure AD with right permissions and then disable it the azuread-access-token is not removed.

Steps:
1. Install rancher on v2.6-head
2. Enable azure AD with all the required microsoft application permissions.
3. Disable the Azure AD auth and notice the secret is not deleted from local cluster >> ns: cattle-global-data >> `azuread-access-token`

re opening the issue as we expect the token to be deleted.

anupama2501:
Verified test cases:

**Migration:**

On v2.6.6 rancher server
1. Enabled azure AD auth by removing the permissions required for migration - auth is enabled 
2. Created 2 downstream clusters - rke1 and rke2 node driver and ran pre upgrade checks
3. Upgraded rancher server
[...]

PR #38776, opened 2022-08-30
Check Azure AD permissions before trying to migrate auth config

maxsokolovsky:
## Problem
Rancher must fail the AuthConfig migration process if the Azure App registrations does not have the necessary permissions configured prior to migration.
 
## Solution
Verify that admins have properly configured an existing app registration's permissions
in the Azure portal before they update their Azure AD auth config to use the new authentication flow.
Fetch the current AuthConfig from the database, update it in-memory to use the new endpoints,
and create a new test Azure client, thereby getting an access token to the Graph API.
Then parse the JWT and inspects the permissions contained within. If the admins had not set those as per docs,
then Rancher won't find the permissions in the test token and will return an error.

I am integrating the code from the https://github.com/rancher/rancher/issues/38754 issue, since these issues are linked, and testing would be easier if these two changes are introduced in a single PR.
1. Immediately throw an error if an access token fetched either from cache or Microsoft lacks permissions.
2. Validate this access token on auth config migration (and thus also do not cache it if any permissions are missing).

### DIFF
--- a/pkg/auth/providers/azure/azure_actions.go
+++ b/pkg/auth/providers/azure/azure_actions.go
@@ -41,6 +41,9 @@ func (ap *azureProvider) actionHandler(actionName string, action *types.Action, 
 	} else if actionName == "testAndApply" {
 		return ap.testAndApply(actionName, action, request)
 	} else if actionName == "upgrade" {
+		if err := ap.checkConfigurationBeforeMigration(); err != nil {
+			return err
+		}
 		return ap.migrateToMicrosoftGraph()
 	}
 
@@ -116,6 +119,28 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 	userExtraInfo := ap.GetUserExtraAttributes(userPrincipal)
 
 	return ap.tokenMGR.CreateTokenAndSetCookie(user.Name, userPrincipal, groupPrincipals, providerToken, 0, "Token via Azure Configuration", request, userExtraInfo)
+}
+
+// checkConfigurationBeforeMigration verifies that admins have properly configured an existing app registration's permissions
+// in the Azure portal before they update their Azure AD auth config to use the new authentication flow.
+// The method fetches the current AuthConfig from the database, then updates it in-memory to use the new endpoints,
+// and creates a new test Azure client, thereby getting an access token to the Graph API.
+// Then it parses the JWT and inspects the permissions contained within. If the admins had not set those as per docs,
+// then Rancher won't find the permissions in the test token and will return an error.
+func (ap *azureProvider) checkConfigurationBeforeMigration() error {
+	cfg, err := ap.getAzureConfigK8s()
+	if err != nil {
+		return err
+	}
+	// Use the future, post-migration, endpoints. This temporarily updated config won't be persisted to the database in this method.
+	updateAzureADEndpoints(cfg)
+
+	azureClient, err := clients.NewMSGraphClient(cfg, ap.secrets)
+	if err != nil {
+		return err
+	}
+	token := azureClient.AccessToken()
+	return clients.EnsureMSGraphTokenHasPermissions(token)
 }
 
 // migrateToMicrosoftGraph represents the migration of the registered Azure AD auth provider
--- a/pkg/auth/providers/azure/azure_actions.go
+++ b/pkg/auth/providers/azure/azure_actions.go
@@ -41,9 +41,6 @@ func (ap *azureProvider) actionHandler(actionName string, action *types.Action, 
 	} else if actionName == "testAndApply" {
 		return ap.testAndApply(actionName, action, request)
 	} else if actionName == "upgrade" {
-		if err := ap.checkConfigurationBeforeMigration(); err != nil {
-			return err
-		}
 		return ap.migrateToMicrosoftGraph()
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -139,21 +136,23 @@ func (c azureMSGraphClient) LoginUser(config *v32.AzureADConfig, credential *v32
...
+		"Group.Read.All",
+		"User.Read.All",
+	}
+	errMSGraphMissingPermissions := fmt.Errorf("missing required Application type permissions from Microsoft Graph: need %s",
+		strings.Join(msGraphRequiredPermissions, ", "))
+
 	field, err := ExtractFieldFromJWT(token, "roles")
 	if err != nil {
-		return false
+		return errMSGraphMissingPermissions
```
KevinJoiner, line 151: `err` here should either be wrapped and returned or logged as info.

KevinJoiner (changes requested):

pkg/auth/providers/azure/clients/ms_graph_client.go
maxsokolovsky: It is wrapped much later and then logged, since this error falls down all the way to HTTP response sending.

MbolotSuse: I think Kevin is referring to the err produced by the `ExtractField` method. Basically, if we can't extract the roles from the JWT, the user will get an error that they don't have the required permissions, and we will have no idea that the reason why was because our processing of the JWT went wrong.

maxsokolovsky: In this case, the processing of the JWT has not gone wrong. This means there are no permissions at all (in this particular case we know it).
If you have only one missing, then the failure will happen later, but the reason the same.
I don't think we need to know which exact permission is missing (although I am open to debate here).
Overall, I don't think any time we return the error from this function we have a JWT processing error.

MbolotSuse: Not sure that I understand properly how we know the jwt processing has not gone wrong. We tried to extract the roles field but got an error - this could be because the token was malformed. How can we be sure that this means the user is missing permissions?

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -232,27 +232,72 @@ type AccessTokenCache struct {
...
 	err = common.CreateOrUpdateSecrets(c.Secrets, string(marshalled), "access-token", "azuread")
 	if err != nil {
-		logrus.Errorf("failed to save the access token in the database: %v", err)
+		logrus.Errorf("[AZUREAD_PROVIDER_CACHE] failed to save the access token in the database: %v", err)
+	}
+}
+
+// extractAccessTokenFromSecret tries to find the JWT with access to the Microsoft Graph API in the JSON value stored
+// in a secret in the database.
+func extractAccessTokenFromSecret(secret []byte) (string, error) {
```
MbolotSuse, line 273: This method seems quite complex. Is there any way that we can give the raw value to an already imported third party library and have it do the parsing of it for us, rather than doing this parsing ourselves as below?

MbolotSuse (changes requested):
Couple of questions, and it needs tests.

pkg/auth/providers/azure/clients/ms_graph_client.go
maxsokolovsky: Yes, I agree, I wonder if there is a third-party library that allows to pass a string for the sequence of keys, how one can do with jq or jsonpath.

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -232,27 +232,72 @@ type AccessTokenCache struct {
...
 }
 
-// Export persists the access token to the secret in the database.
+// Export persists the access token to a secret in the database. Before doing that, it validates the access token
+// to ensure it contains the necessary permissions, so as not to persist a token with wrong or insufficient permissions.
 func (c AccessTokenCache) Export(cache cache.Marshaler, key string) {
 	marshalled, err := cache.Marshal()
 	if err != nil {
-		logrus.Errorf("failed to marshal the access token before saving to the database: %v", err)
+		logrus.Errorf("[AZUREAD_PROVIDER_CACHE] failed to marshal the access token before saving to the database: %v", err)
```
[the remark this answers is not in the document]
maxsokolovsky: Agreed! Also, we seem to use the prefix sporadically. In some statements and also packages I see it used, in other not. Is there some unwritten rule or standard for when to use or not use prefixes like this?

pkg/auth/providers/azure/clients/ms_graph_client.go
maxsokolovsky: Yeah, if it's malformed, then we should indicate so explicitly, will adjust.

KevinJoiner (dismissed):

pkg/auth/providers/azure/clients/ms_graph_client.go
```diff
@@ -282,6 +293,15 @@ func NewMSGraphClient(config *v32.AzureADConfig, secrets corev1.SecretInterface)
 			return nil, err
 		}
 	}
+
+	if err = EnsureMSGraphTokenHasPermissions(ar.AccessToken); err != nil {
+		if deletionErr := tokenCache.Clear(); deletionErr != nil {
+			logrus.Errorf("[%s] failed to delete the cached secret %s with an invalid access token: %v",
```
MbolotSuse, line 299: Typically we don't like to signal the location of cache secrets @cbron any opinion on this error message?

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -118,28 +118,56 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
 	if err != nil {
-		return err
+		return nil, err
 	}
-	if authProviderEnabled(cfg) {
-		updateAzureADEndpoints(cfg)
+	updateAzureADEndpoints(cfg)
+
+	// Try to get a new client, which will fetch a new access token and validate its permissions.
+	_, err = clients.NewMSGraphClient(cfg, ap.secrets)
```
MbolotSuse, line 156: This seems a bit risky to me. Before, I think the client was only constructed when we had an active, enabled config. But here we are constructing even if the config is not active. This seems like it could easily result in a nil pointer due to some value not being filled in that we expect to be filled in. Do you consider this function safe to call for a non-enabled auth config?

MbolotSuse (commented):
Mostly looks good. I have one concern about an overly verbose error message and a concern about an overly risky call to the MS graph client. If you could offer some input on if we are ok with that scenario I would appreciate it.

Outside of that, I would note that this is going to change the behavior of how the migrate function deletes the current cache, which may not be desireable.

I'm open to approving if we are ok with the risk outlined here.

pkg/auth/providers/azure/clients/ms_graph_client.go
cbron: I'm ok with it being in logs, since those should be admin-only anyway. The code is public so if someone really wants to know they can figure it out. As long as we aren't logging any actual data, which it looks like we are not.

pkg/auth/providers/azure/azure_actions.go
cbron: I'd agree. The `authProviderEnabled` check got moved to later in the logic, and it should still be up front.

maxsokolovsky: That's a good point! I checked and had no panics. The upgrade request gets an error on trying to initialize the client, since the secret exchange fails (the secret is an empty string in a disabled config). But I added a check if the auth provider is enabled before trying to initialize a client to the Graph API.

The only small thing is the response code, as I'd prefer to have a 400 in this case, but returning an error from this area of the code always gets caught in a generic 500 response with code. I am not sure how consistent we are in these situations elsewhere...

MbolotSuse: I'm guessing that if you directly return an httperror it will take the code from that error rather than wrapping in a 500 (see the `ActionNotAvailable` error in `actionHandler` and the `InvalidBodyCount` error in `testAndApply`). Would be worth giving it a try if you would prefer a specific 400 error code.

maxsokolovsky: I think `ActionNotAvailable` is good here, but its code is 404, which might seem strange, but better than 500.
There is also `InvalidState 422`, which could be reasonable here.

MbolotSuse (approved):
LGTM assuming that we add tests in a subsequent PR.

KevinJoiner (approved):

PR #38791, opened 2022-09-01
[release/v2.6.8-patch1] Make nodepool controller more reliable

Oats87:
## Problem
The nodepool controller is dangerous and spawns off reconciliation in uncontrolled goroutines. In certain cases, it is possible to see the nodepool controller trigger multiple reconciliation goroutines for the same nodepool concurrently, which can create and/or delete too many nodes. In addition, the nodepool controller was operating off of the node cache which is not ideal, and led to potential creation/deletion of too many nodes. Lastly, the nodepool controller was not properly returning the latest version of the nodepool object, which was causing the nodepool cache to be stale and lead to stale objects being returned. This was particularly exasperated by the fact that the nodepool controller would retrieve the latest object from API for mutation, rather than operating on the cached object.

## Solution
The nodepool controller is no longer using goroutines for reconciliation. This has the side effect of slowing down nodepool provisioning reconciliation (a potential scaling impact), but makes the nodepool controller significantly more reliable.

The nodepool controller now uses the latest version of the node objects from the api server. It also returns the latest mutated object for future reconciliation loops.

### DIFF
--- a/pkg/controllers/management/nodepool/nodepool.go
+++ b/pkg/controllers/management/nodepool/nodepool.go
@@ -30,8 +30,7 @@ var (
 )
 
 const (
-	ReconcileAnnotation  = "nodepool.cattle.io/reconcile"
-	DeleteNodeAnnotation = "nodepool.cattle.io/delete-node"
+	ReconcileAnnotation = "nodepool.cattle.io/reconcile"
 )
 
 type Controller struct {
@@ -44,6 +43,7 @@ type Controller struct {
 	syncmap            map[string]bool
 }
 
+// Register registers the nodepool controller against the managementcontext.
 func Register(ctx context.Context, management *config.ManagementContext) {
 	p := &Controller{
 		NodePoolController: management.Management.NodePools("").Controller(),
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -140,15 +158,32 @@ func (c *Controller) Remove(nodePool *v3.NodePool) (runtime.Object, error) {
...
 
-func (c *Controller) setReconcileAnnotation(nodePool *v3.NodePool, anno string) (*v3.NodePool, error) {
+// nodes returns a slice of node objects listed directly from the Kubernetes API.
+func (c *Controller) nodes(nodePool *v3.NodePool) ([]*v3.Node, error) {
+	nodeList, err := c.Nodes.ListNamespaced(nodePool.Namespace, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	var nodes []*v3.Node
```
kinarashah, line 176: Do we need to create a new list here? since it's a new func, signature could be []*v3.NodeList and loop over list.Items in the logic

kinarashah, line 176: Do we need to create a new list here? since it's a new func, signature could be []*v3.NodeList and loop over list.Items in the logic

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -56,77 +56,95 @@ func Register(ctx context.Context, management *config.ManagementContext) {
...
 	}
+	return nodePool, nil
 }
 
+// Updated reconciles a nodepool if necessary
 func (c *Controller) Updated(nodePool *v3.NodePool) (runtime.Object, error) {
+	if nodePool != nil {
+		logrus.Tracef("[nodepool] Updated called for nodepool %s: %+v", nodePool.Name, nodePool)
+	}
 	obj, err := v32.NodePoolConditionUpdated.Do(nodePool, func() (runtime.Object, error) {
```
kinarashah, line 103: Do we rely on this condition to have the status/error msg? I think it won't reflect on the nodePool because there's no update on the `obj`

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -56,77 +56,95 @@ func Register(ctx context.Context, management *config.ManagementContext) {
...
 		anno, _ := nodePool.Annotations[ReconcileAnnotation]
+		logrus.Debugf("[nodepool] nodepool %s reconcile annotation value was: %s", nodePool.Name, anno)
 		if anno == "" {
-			nodes, err := c.NodeLister.List(nodePool.Namespace, labels.Everything())
+			nodes, err := c.nodes(nodePool)
 			if err != nil {
 				return nodePool, err
 			}
-			go c.deleteBadNodes(nodes)
```
kinarashah, line 99: deleteBadNodes took care of deleting nodes if they failed to came up in 2 mins. once we return from our nodePool logic now, we won't get other nodePool syncs to delete these nodes. is this something we need to account for?

kinarashah: This might break our expected usecase of 
`
After 2-3 minutes the stopped nodes should be removed and replaced from Rancher, additional nodes beyond the stopped count should not be removed or created
`

Oats87: deleteBadNodes the way I interpreted it, set a 2 minute countdown for deletion if the conditions didn’t exist at time of loop reconcile. It did not recheck conditions right before deletion. I figure the node will eventually get deleted, but not at the guaranteed 2 minute mark. I could be wrong on my original interpretation though

pkg/controllers/management/nodepool/nodepool.go
```diff
@@ -56,77 +56,95 @@ func Register(ctx context.Context, management *config.ManagementContext) {
...
+			// This has the potential to race if the reconcile loop takes longer than 120 seconds, but this is a
+			// very rare case as there is nothing in reconcile/needsReconcile that should block. It is more likely that
+			// Rancher will die/panic from leader election loss, and a new instance of Rancher will take over (and thus,
+			// need to clear the annotation) in order to reconcile the pool.
 			pieces := strings.Split(anno, "/")
 			t, err := time.Parse(time.RFC3339, pieces[1])
-			if err != nil || int(time.Since(t)/time.Second) > 20 {
+			if err != nil || int(time.Since(t)/time.Second) > 120 {
+				logrus.Debugf("[nodepool] attempting to clear reconcile annotation for %s", nodePool.Name)
 				nodePool.Annotations[ReconcileAnnotation] = ""
```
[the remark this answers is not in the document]
Oats87: I believe the object passed in from lifecycle is already deepcopied but I could be wrong.

pkg/controllers/management/nodepool/nodepool.go
Oats87: I just stole this function from an old commit. Do you want me to optimize it? It was removed and I am more or less reverting the removal

pkg/controllers/management/nodepool/nodepool.go
kinarashah: ah nvm, I see the check for conditions is now moved to create/check nodes func. That should cover it

pkg/controllers/management/nodepool/nodepool.go
kinarashah: marking resolved to keep old behavior as is for now

pkg/controllers/management/nodepool/nodepool.go
kinarashah: marking resolved to keep old behavior as is for now

kinarashah (approved):

PR #38796, opened 2022-09-01
Fix encryption key rotation for >=10k secrets

jakefhyde:
## Problem

Encryption key rotation would fail with a large number of secrets. This was fixed in part due to https://github.com/k3s-io/k3s/pull/5936, which would was a bug in rke2/k3s that would cause encryption key rotation to finish reencrypting prematurely after 8k secrets. In the rancher implementation, there were issues regarding the planner instructions being unreliable, utilizing periodic instructions to verify that the status was as expected, verifying each phase. However, all controlplane nodes except for the leader would have their stages verified on restart, but the leader would take extra time on startup to have the correct stage. For more saturated clusters, this would cause the leader to not be finished reconciling bootstrap data after restart, and other clusters would not be ready for subsequent restarts, which caused encryption key rotation to get stuck.

## Solution

- Verify  the expected status in the system-agent plans
  - Since encryption key rotation is variable, there is no timeout for verifying the status: a cluster that is stuck is not safe and requires a restore, and it's unknowable whether or not a cluster is stuck
- Removed the killall instruction. The killall instruction was causing issues with clusters that were masked previously due to re-running the plans after a timeout, which would in turn cause encryption key rotation to fail at a later time.
- Also fixed the status instruction grepping the entire output, which could cause issues if a node contained the name of an `secrets-encrypt` stage, like the following example:

```
$ rke2 secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: prepare
Server Encryption Hashes: hash does not match between prepare-abcde and rotate-abcde

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2021-12-10T22:54:38Z
        AES-CBC   aescbckey
```

## Testing

:heavy_check_mark: 
### Manual Testing

I tested manually with a 3 node HA all roles cluster on v1.24.4+rke2r1, which is the earliest release in the v1.24 release line that contains the required fix from the rke2/k3s side: https://github.com/k3s-io/k3s/pull/5936

### Automated Testing

I have created automated tests to run this: https://github.com/rancher/rancher/pull/38259

## QA Testing Considerations

- Rotating encryption keys on a fresh cluster
- Rotating encryption keys a second time on the same cluster after adding 10k secrets (expected to take 45 minutes to an hour).

N/A

### DIFF
--- a/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
+++ b/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
@@ -26,8 +26,7 @@ const (
 	encryptionKeyRotationCommandRotate    = "rotate"
 	encryptionKeyRotationCommandReencrypt = "reencrypt"
 
-	encryptionKeyRotationSecretsEncryptApplyCommand  = "secrets-encrypt-apply"
-	encryptionKeyRotationSecretsEncryptStatusCommand = "secrets-encrypt-status"
+	encryptionKeyRotationSecretsEncryptApplyCommand = "secrets-encrypt-apply"
 
 	encryptionKeyRotationLeaderAnnotation = "rke.cattle.io/encrypt-key-rotation-leader"
 
@@ -393,7 +392,6 @@ func (p *Planner) encryptionKeyRotationRestartNodes(cp *rkev1.RKEControlPlane, c
 // status can be successfully queried, and then gets the status. leaderStage is allowed to be empty if entry is the
 // leader.
 func (p *Planner) encryptionKeyRotationRestartService(cp *rkev1.RKEControlPlane, entry *planEntry) error {
-	//runtime := rke2.GetRuntime(cp.Spec.KubernetesVersion)
 	nodePlan := plan.NodePlan{
 		Files: []plan.File{
 			{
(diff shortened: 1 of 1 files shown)


pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -418,37 +392,30 @@ func (p *Planner) encryptionKeyRotationRestartNodes(cp *rkev1.RKEControlPlane, c
 // encryptionKeyRotationRestartService restarts the server unit on the downstream node, waits until secrets-encrypt
 // status can be successfully queried, and then gets the status. leaderStage is allowed to be empty if entry is the
 // leader.
-func (p *Planner) encryptionKeyRotationRestartService(cp *rkev1.RKEControlPlane, entry *planEntry, scrapeStage bool, leaderStage string) (string, error) {
-	runtime := rke2.GetRuntime(cp.Spec.KubernetesVersion)
+func (p *Planner) encryptionKeyRotationRestartService(cp *rkev1.RKEControlPlane, entry *planEntry) error {
+	//runtime := rke2.GetRuntime(cp.Spec.KubernetesVersion)
```
annablender, line 396: Remove pseudo code comment?

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -758,34 +637,10 @@ func encryptionKeyRotationWaitForSecretsEncryptStatus(cp *rkev1.RKEControlPlane)
...
-		}
-	}
-	return fmt.Errorf("unexpected encryption key rotation stage [%s] for phase [%s]", leaderStage, currentPhase)
-}
-
 // encryptionKeyRotationFailed updates the various status objects on the control plane, allowing the cluster to
 // continue the reconciliation loop. Encryption key rotation will not be restarted again until requested.
 func (p *Planner) encryptionKeyRotationFailed(cp *rkev1.RKEControlPlane, err error) error {
-	if err2 := p.setEncryptionKeyRotateState(cp, cp.Spec.RotateEncryptionKeys, rkev1.RotateEncryptionKeysPhaseFailed); err2 != nil {
+	if err2 := p.setEncryptionKeyRotateState(cp, cp.Spec.RotateEncryptionKeys, rkev1.RotateEncryptionKeysPhaseFailed); err2 != nil && !isErrWaiting(err) {
```
annablender, line 643: Make var more specific instead of `err2` I would do `setRotateStateErr`

annablender (changes requested):

annablender:
Otherwise looks good, long as testing works on your end

Oats87 (changes requested):
Does this still work for older versions of K3s and RKE2 reliably?

Oats87:
What happens if a user upgrades their Rancher in the middle of a secrets encrypt rotation that has the old logic to this new logic?

jakefhyde:
> Does this still work for older versions of K3s and RKE2 reliably?

I tested initially with v1.24.2 which does have the fixes required from https://github.com/k3s-io/k3s/pull/5936, which for >8k secrets would get stuck in the same exact spot. With v1.24.4, the cluster would get stuck at a different spot (restarting the last node). After making changes to the logic to determine which part was causing this new issue, it came down to the killall script, mainly seeing behavior similar to https://github.com/rancher/rke2/issues/3006 but only for restarting the last node after running `reencrypt` (i.e. `secrets-encrypt reencrypt`, wait, restart leader, restart next node, restart last node <- this step would fail).

PR #38809, opened 2022-09-02
Adding environment variable for SUSE to set base registry

mattfarina:
## Problem
SUSE/Rancher needs the ability to set the default registry to pull images from. This is separate from the user setting so that user settings (especially from charts) do not overwrite this default value.
 
## Solution
Add a new environment variable that's the default value for the CATTLE_SYSTEM_DEFAULT_REGISTRY
 
### Manual Testing
N/A

### Automated Testing
An automated test was added to ensure the default value for the system default registry is the same as it was before the change when the new environment variable is not set.

No test was added for when the new environment variable was set because I didn't see a good way to inject it into the current test setup. Pointers welcome if there is a good place to do this.

## QA Testing Considerations
Testing this on a new install and an upgrade would be useful. An upgrade where CATTLE_BASE_REGISTRY starts as "" and moves to a different registry and an upgrade that starts with a specified registry and goes to "".
  
When the new environment variable is not set it will default to "" which is the existing value.

### DIFF
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -142,8 +142,8 @@ var (
 	// RancherWebhookMinVersion is the minimum version of the webhook that rancher will install
 	RancherWebhookMinVersion = NewSetting("rancher-webhook-min-version", "")
 
-	// SystemDefaultRegistry is the default registry set by users when the environment variable CATTLE_SYSTEM_DEFAULT_REGISTRY
-	// is set. The environment variable CATTLE_BASE_REGISTRY is internal and used to set the base registry as set by SUSE.
+	// SystemDefaultRegistry is the default contrainer registry used for images.
+	// The environmental variable "CATTLE_BASE_REGISTRY" controls the default value of this setting.
 	SystemDefaultRegistry = NewSetting("system-default-registry", os.Getenv("CATTLE_BASE_REGISTRY"))
 
 	// Configuration to display a custom fixed banner in the header, footer, or both
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -76,7 +76,6 @@ var (
 	SystemAgentInstallerImage           = NewSetting("system-agent-installer-image", "rancher/system-agent-installer-")
 	SystemAgentUpgradeImage             = NewSetting("system-agent-upgrade-image", "")
 	WinsAgentUpgradeImage               = NewSetting("wins-agent-upgrade-image", "")
-	SystemDefaultRegistry               = NewSetting("system-default-registry", "")
 	SystemNamespaces                    = NewSetting("system-namespaces", "kube-system,kube-public,cattle-system,cattle-alerting,cattle-logging,cattle-prometheus,ingress-nginx,cattle-global-data,cattle-istio,kube-node-lease,cert-manager,cattle-global-nt,security-scan,cattle-fleet-system,cattle-fleet-local-system,calico-system,tigera-operator,cattle-impersonation-system,rancher-operator-system,cattle-csp-adapter-system,calico-apiserver")
 	SystemUpgradeControllerChartVersion = NewSetting("system-upgrade-controller-chart-version", "")
 	TelemetryOpt                        = NewSetting("telemetry-opt", "")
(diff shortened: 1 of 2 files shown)


pkg/settings/setting_test.go
```diff
@@ -29,3 +30,27 @@ func TestIsRelease(t *testing.T) {
 		a.Equal(value, result, fmt.Sprintf("Expected value [%t] for key [%s]. Got value [%t]", value, key, result))
 	}
 }
+
+// TestSystemDefaultRegistryDefault tests that the default registry is either
+// "" (an empty string) or the build time value to ensure backward compatibility.
+// Note, if CATTLE_BASE_REGISTRY is set this test in the testing environment
+// this test will fail.
+func TestSystemDefaultRegistryDefault(t *testing.T) {
+	var expect string
```
KevinJoiner, line 39: We want to test that the value of our new environment variable `CATTLE_BASE_REGISTRY` is used as the default setting for `SystemDefaultRegistry`.
```suggestion
	expect := os.Getenv("CATTLE_BASE_REGISTRY")
```

If we want to ensure backward compatibility, we need an integration test that runs Rancher server and verifies different registries are used when `CATTLE_BASE_REGISTRY` is set and unset.

pkg/settings/setting.go
```diff
@@ -159,6 +158,10 @@ var (
 
 	// RancherWebhookMinVersion is the minimum version of the webhook that rancher will install
 	RancherWebhookMinVersion = NewSetting("rancher-webhook-min-version", "")
+
+	// SystemDefaultRegistry is the default registry set by users when the environment variable CATTLE_SYSTEM_DEFAULT_REGISTRY
+	// is set. The environment variable CATTLE_BASE_REGISTRY is internal and used to set the base registry as set by SUSE.
+	SystemDefaultRegistry = NewSetting("system-default-registry", os.Getenv("CATTLE_BASE_REGISTRY"))
```
KevinJoiner, line 164: This comment is more about Rancher's particular usage of the setting `system-default-registry` and not the variable `SystemDefaultRegistry` itself. I think the information is helpful but is better suited to go into Rancher docs.
> SystemDefaultRegistry is the default registry set by users when the environment variable CATTLE_SYSTEM_DEFAULT_REGISTRY

We can leave this first part off since the fact that this setting is affected by the environmental variable `CATTLE_SYSTEM_DEFAULT_REGISTRY` is an implementation detail of the dashboardapi controller, which is independent of this package.

> The environment variable CATTLE_BASE_REGISTRY is internal and used to set the base registry as set by SUSE.

Our internal usage of the environmental variable `CATTLE_BASE_REGISTRY` is also independent of this package. 
```suggestion
	// SystemDefaultRegistry is the default contrainer registry used for images.
	// The environmental variable "CATTLE_BASE_REGISTRY" controls the default value of this setting.
	SystemDefaultRegistry = NewSetting("system-default-registry", os.Getenv("CATTLE_BASE_REGISTRY"))
```

mattfarina: I am looking to add more code comments to the internals of Rancher. The ratio of comments to code is too low and better explanations within the code will be useful in the future. Thanks for helping to clarify that one.

KevinJoiner: Agreed, I added comments to this file earlier this year for that very reason.

pkg/settings/setting_test.go
```diff
@@ -29,3 +30,27 @@ func TestIsRelease(t *testing.T) {
 		a.Equal(value, result, fmt.Sprintf("Expected value [%t] for key [%s]. Got value [%t]", value, key, result))
 	}
 }
+
+// TestSystemDefaultRegistryDefault tests that the default registry is either
+// "" (an empty string) or the build time value to ensure backward compatibility.
```
cbron, line 35: What do you mean by build time ? Is something at build time setting a var in [InjectDefaults](https://github.com/rancher/rancher/blob/release/v2.6/scripts/build-server#L20) ? If so where is that. All this PR seems to do is add ability to set an ENV var.

pkg/settings/setting_test.go
mattfarina: Updated.

mattfarina:
@cbron @KevinJoiner can you please re-review

pkg/settings/setting_test.go
```diff
@@ -29,3 +31,26 @@ func TestIsRelease(t *testing.T) {
...
+	expect := os.Getenv("CATTLE_BASE_REGISTRY")
+	if InjectDefaults != "" {
+		defaults := map[string]string{}
+		if err := json.Unmarshal([]byte(InjectDefaults), &defaults); err != nil {
+			t.Errorf("Unable to parse InjectDefaults: %v", err)
+		}
+
+		if value, ok := settings["system-default-registry"]; ok {
+			expect = value.Get()
+		}
```
KevinJoiner, line 48:
```suggestion
		if value, ok := defaults["system-default-registry"]; ok {
			expect = value
		}
```
Sorry, I missed this before, but the previous version does not verify that the value from InjectDefaults is being used.

KevinJoiner (changes requested):

pkg/settings/setting_test.go
mattfarina: Great catch. oops on my part. Fixed.

KevinJoiner (dismissed):
LGTM

mattfarina:
@KevinJoiner @cbron can you give this a review. Had to rebase.

KevinJoiner (approved):

Issue #38668, opened 2022-08-19
Hide CIS 1.5 scan profiles from cis benchmark charts


mitulshah-suse:
CIS benchmark 1.5 scan profiles support kubernetes version 1.15 which is not used anywhere in rancher.
So we need to hide those scan profiles in the CIS benchmark chart

vivek-infracloud:
The CIS benchmark scan profile 1.5 is not visible on the cluster.
![Screenshot from 2022-09-20 12-00-30](https://user-images.githubusercontent.com/83208989/191183946-fcd24f12-f5e8-454b-b0f6-990439b17afe.png)

PR #38814, opened 2022-09-06
 Skip flaky tests that are no longer supported

rishabhmsra:
## Issue
From _rancher-v3_additional_tests_ jenkins job below tests were failing due to incompatibility with latest rancher server versions:
- test_catalog_library
- test_cis_scan
- test_logging
- test_monitoring
- test_multi_cluster_app

## Solution

- Added pytest.skip(allow_module_level=True) to skip all the related tests.

### DIFF
--- a/tests/validation/tests/v3_api/test_catalog_library.py
+++ b/tests/validation/tests/v3_api/test_catalog_library.py
@@ -24,6 +24,9 @@ from .common import get_project_client_for_token
 from .common import USER_TOKEN
 from .common import get_user_client
 
+# Skipping all the tests from this module because the library_catalog charts are not supported on latest
+# rancher versions
+pytest.skip(allow_module_level=True)
 
 cluster_info = {"cluster": None, "cluster_client": None,
                 "project": None, "project_client": None,
--- a/tests/validation/tests/v3_api/test_cis_scan.py
+++ b/tests/validation/tests/v3_api/test_cis_scan.py
@@ -25,6 +25,9 @@ from .common import wait_for_cluster_node_count
 from .test_rke_cluster_provisioning import HOST_NAME, \
     POD_SECURITY_POLICY_TEMPLATE, get_cis_rke_config  # NOQA
 
+# Skipping all the tests from this module since CIS 1.5 and CIS 1.4 profile are not supported on latest k8s versions.
+pytest.skip(allow_module_level=True)
+
 scan_results = {
     "rke-cis-1.4": {
         "permissive": {"pass": 63, "skip": 15},
--- a/tests/validation/tests/v3_api/test_multi_cluster_app.py
+++ b/tests/validation/tests/v3_api/test_multi_cluster_app.py
@@ -18,6 +18,8 @@ from .common import delete_node
 import pytest
 import time
 
+# Skipping all the tests from this module because Multi-Cluster Apps have been deprecated as of Rancher v2.5.0
+pytest.skip(allow_module_level=True)
 
 project = {}
 project_detail = {"c0_id": None, "c1_id": None, "c2_id": None,
--- a/tests/validation/tests/v3_api/test_logging.py
+++ b/tests/validation/tests/v3_api/test_logging.py
@@ -12,6 +12,10 @@ from .common import wait_for_app_to_remove
 from .common import CATTLE_TEST_URL
 from .common import USER_TOKEN
 
+# Skipping this test module because library-fluentd-aggregator from library catalog is not supported on latest
+# rancher versions
+pytest.skip(allow_module_level=True)
+
 namespace = {"p_client": None, "ns": None, "cluster": None, "project": None,
              "name_prefix": None, "admin_client": None, "sys_p_client": None}
 fluentd_aggregator_answers = {"defaultImage": "true",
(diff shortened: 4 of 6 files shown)


tests/validation/tests/v3_api/test_catalog_library.py
```diff
@@ -24,6 +24,9 @@
 from .common import USER_TOKEN
 from .common import get_user_client
 
+# Skipping all the tests from this module because the library_catalog charts are not supported on latest
```
izaac, line 27: Do we need to skip if Rancher 2.5.x @sowmyav27 or should we focus on 2.6.x only this is mainly for 2.6 failures due to support issues.

tests/validation/tests/v3_api/test_cis_scan.py
```diff
@@ -25,6 +25,9 @@
 from .test_rke_cluster_provisioning import HOST_NAME, \
     POD_SECURITY_POLICY_TEMPLATE, get_cis_rke_config  # NOQA
 
+# Skipping all the tests from this module since CIS 1.5 and CIS 1.4 profile are not supported on latest k8s versions.
+pytest.skip(allow_module_level=True)
```
izaac, line 29: @sowmyav27 should only the 1.4 scans be disabled here? or both.

izaac (commented):
Just a couple support questions I have but this looks LGTM otherwise.

tests/validation/tests/v3_api/test_cis_scan.py
sowmyav27: Both technically.

tests/validation/tests/v3_api/test_catalog_library.py
sowmyav27: Why are we skipping this @rishabhmsra ?

sowmyav27 (changes requested):

tests/validation/tests/v3_api/test_catalog_library.py
rishabhmsra: @sowmyav27, since `test_catalog_library` tests are all related to v1 charts validation, it's failing because of the compatibility. I've DM'ed you the jenkins build logs for more details.

tests/validation/tests/v3_api/test_cis_scan.py
anupama2501: Since 2.6 supports k8s 1.21 and below shouldn't we have cis 1.5?

tests/validation/tests/v3_api/test_cis_scan.py
```diff
@@ -25,6 +25,9 @@
 from .test_rke_cluster_provisioning import HOST_NAME, \
     POD_SECURITY_POLICY_TEMPLATE, get_cis_rke_config  # NOQA
 
+# Skipping all the tests from this module since CIS 1.5 and CIS 1.4 profile are not supported on latest k8s versions.
```
anupama2501, line 28: just a nit, can we add the k8s version from which we don't support these cis profiles instead of `latest` k8s version here.

tests/validation/tests/v3_api/test_catalog_library.py
anupama2501: IMO we should add a check to skip these tests only if the rancher versions are v2.6 and greater.

tests/validation/tests/v3_api/test_cis_scan.py
rishabhmsra: As per this issue https://github.com/rancher/rancher/issues/38668 CIS 1.5 profile is hidden now, as it supports k8s 1.15.

tests/validation/tests/v3_api/test_cis_scan.py
sowmyav27: Actually test_cis_scan is v1 app install, and i don't think this is supported anymore. Please correct me if I am wrong @rishabhmsra

sowmyav27 (approved):

tests/validation/tests/v3_api/test_catalog_library.py
anupama2501: Checked the convo. sounds good. Thanks Rishabh.

tests/validation/tests/v3_api/test_multi_cluster_app.py
```diff
@@ -18,6 +18,8 @@
 import pytest
 import time
```
ronhorton, line 20: is this a typo? (2.5.0) 
multi-cluster apps are deprecated as of 2.7.0

ronhorton (approved):
chatted with infracloud this morning to answer a couple of questions I had - to familiarize myself with the reason for these changes. 
changes look good.

izaac:
@rishabhmsra we require a forwardport in the 2.7 branch linked to this.

tests/validation/tests/v3_api/test_cis_scan.py
anupama2501: @rishabhmsra if you can confirm if this ^ is the case, we can change the comment based on that.  LGTM otherwise

rishabhmsra: Yes Somwya that's correct.

anupama2501 (approved):
Checked the jenkins job and the run looks good. Please make sure to add in the 2.7 PR as mentioned by Izaac as we have different branches for 2.7 and 2.6 now. 

LGTM

izaac (approved):
LGTM

Issue #37295, opened 2022-04-13
Fails to connect to downstream cluster's nodes when rancher is behind a proxy and nodes do not have public IPs


jiaqiluo:
**Rancher Server Setup**
- Rancher version: 2.6.4
- Installation option (Docker install/Helm Chart): either one 
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1 
- Proxy/Cert Details: Squid proxy 

**Information about the Cluster**
- Kubernetes version: any
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): infrastructure Provider, AWS 

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) Admin
  - If custom, define the set of permissions:

   
**Describe the bug**

The internal `rancher-machine` is used to create nodes when we create a node-driver cluster, but it does not use the `no_proxy` env variable, so when the proxy cannot talk to the nodes in the downstream cluster, rancher will fail to create the cluster and show ssh timeout errors. 

**To Reproduce**
 
- run a squid proxy 
- Helm install rancher with proper values for  `http_proxy` `https_proxy` and `no_proxy` 
- make sure the `no_proxy` value contains the CIDR for the to-be-created downstream cluster's nodes' private IPs. 

Here is the Helm command I used 
```
helm upgrade --install rancher rancher-latest/rancher \
   --namespace cattle-system \
   --version v2.6.4 \
   --set hostname=ec2-xxx1.us-west-1.compute.amazonaws.com \
   --set proxy=http://ec2--xxx2.us-west-1.compute.amazonaws.com:3131 \
   --set noProxy="127.0.0.0/8\,10.0.0.0/8\,172.16.0.0/12\,192.168.0.0/16\,.svc\,.cluster.local\,172.31.0.0/16" 
```

- After rancher runs up, create an ec2 node template that does not have the public IP
- create a downstream cluster by using the node template,  one node with all roles is sufficient 

**Result**
- the node is created, but we will see the following error in rancher UI
`[ERROR] error syncing 'c-pjz46/m-g8ckn': handler node-controller: Error creating machine: Error detecting OS: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded, requeuing`

**Expected Result**

[...]

PR #38835, opened 2022-09-07
Updating proxy to allow node driver provisioning

slickwarren:
## Issue: 
https://github.com/rancher/qa-tasks/issues/486

## Problem

our proxy was not allowing downstream node driver clusters to be created. This was due to missing parameters in the squid.conf 

We also were lacking parameters for a specific use case, where a rancher setup is behind proxy with downstream nodes behind a proxy (see https://github.com/rancher/rancher/issues/37295#issuecomment-1111380535 for more details on the use case)
 
## Solution

* added new rules to squid.conf to open ports for node driver clusters
* added new parameter that appends to NO_PROXY if specified. This should allow for testing the proxy/airgap use case.
 
## Testing

see offline link for more details. Tested with and without the CIDR override setting. Also have supporting jenkinsjob PR ready.


MSpencer87 (approved):
LGTM

izaac (approved):
LGTM
This seems to not impact other jobs.

Issue #38419, opened 2022-07-26
[BUG] Circular RoleTemplate reference causes high CPU usage or crash


MbolotSuse:
**Rancher Server Setup**
- Rancher version: 2.6.6
- Installation option: Docker (crash) or Helm (high cpu) 
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: N/A can occur on any

**User Information**
- What is the role of the user logged in? N/A can occur with any user

**Describe the bug**
If two role templates inherit each other, Rancher enters a state of high cpu usage or outright crashes.

**To Reproduce**
1. Create a role template `test-pods`. Have it grant `get, list, watch`  on `pods`.
2. Create a role template `test-service-accounts`. Have it grant `get, list, watch` on `serviceAccounts`. Have it inherit from `test-pods`.
3. Update `test-pods` to inherit from `test-service-accounts`.

**Result**
Rancher enters a state of high CPU usage or outright crashed.

**Expected Result**
Rancher rejects the request to update `test-pods` to inherit from `test-service-accounts`, or does not enter a state of high cpu usage.

MbolotSuse:
In addition to validation in the webhook, there also needs to be a stop-gap implemented in rancher to ensure that the [gatherRolesRecurse](https://github.com/rancher/rancher/blob/3517d0b2ff5daabe124b6c96b1e2cde40ccb44ba/pkg/controllers/managementuser/rbac/handler_base.go#L355) function does not infinitely recurse. This should involve a soft-limit (<100) after which a log message is output and a hard limit (~1000) after which an error is returned and the role-template is not processed.

cbron:
Either @steffantucker or @MbolotSuse can pick up, it has clear path forward so Tucker just ask michael about it first.

MbolotSuse:
## Validation Template

@steffantucker feel free to add onto this if I miss any details

## Root Cause

When a roleTemplate included a circular reference (i.e. it inherited a role which also inherited it), rancher enters a state of high cpu usage or outright crashes. This is due to the fact that rancher recursively gathers the rules that a roleTemplate provides. Since this is done recursively, circular references will infinitely attempt to gather roles by continuing to recurse over a given set of roles. 

## What was fixed, or what change have occurred

There are two components to this fix:
[...]

PR #38848, opened 2022-09-08
Fixes circular roles prevention filling up logs

steffantucker:
Previous logging would result in a lot of logs when a recursive role was encountered. Addressing re-opening of https://github.com/rancher/rancher/issues/38419

## Issue: 
Logging and errors resulting from circular roles fills up logs.
 
## Problem
The error gets wrapped before returning in a recursive function.
 
## Solution
Just return the error.
 
## Testing
Debug and printed err to confirm it is no longer a big string.

n/a


KevinJoiner (approved):
LGTM 👍

cbron (approved):

PR #38850, opened 2022-09-08
Add sync back to check for user-node-remove, add tests

jakefhyde:
## Problem

The `user-node-remove` handler has been removed, but the finalizer is not being cleaned up.

## Solution

If the node-controller finalizer has already run for the node (or was never present in the first place), the remove handler won't run again. Because of this, we need to add sync back, but ignore the annotation if the object is already being deleted.

### Manual Testing

I tested by setting the finalizer on the object, on a previous version of rancher, deleting the node, then upgrading, which caused the finalizer to be removed correctly.

### DIFF
--- a/pkg/controllers/management/node/cleanup.go
+++ b/pkg/controllers/management/node/cleanup.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/rancher/rancher/pkg/agent/clean"
-	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	util "github.com/rancher/rancher/pkg/cluster"
 	"github.com/rancher/rancher/pkg/dialer"
 	"github.com/rancher/rancher/pkg/features"
@@ -28,9 +28,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-const cleanupPodLabel = "rke.cattle.io/cleanup-node"
+const (
+	cleanupPodLabel                    = "rke.cattle.io/cleanup-node"
+	userNodeRemoveAnnotationPrefix     = "lifecycle.cattle.io/create.user-node-remove_"
+	userNodeRemoveCleanupAnnotationOld = "nodes.management.cattle.io/user-node-remove-cleanup"
+)
 
-func (m *Lifecycle) deleteV1Node(node *v32.Node) (runtime.Object, error) {
+func (m *Lifecycle) deleteV1Node(node *v3.Node) (runtime.Object, error) {
 	logrus.Debugf("Deleting v1.node for [%v] node", node.Status.NodeName)
 	if nodehelper.IgnoreNode(node.Status.NodeName, node.Status.NodeLabels) {
 		logrus.Debugf("Skipping v1.node removal for [%v] node", node.Status.NodeName)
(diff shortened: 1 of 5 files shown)


jiaqiluo (approved):

HarrisonWAffel (dismissed):

pkg/controllers/management/node/cleanup.go
```diff
@@ -393,11 +397,30 @@ func (m *Lifecycle) createCleanupJob(userContext *config.UserContext, cluster *v
...
+	}
+	return m.nodeClient.Update(newObj)
+}
+
 func removeFinalizerWithPrefix(finalizers []string, prefix string) []string {
 	var nf []string
 	for _, finalizer := range finalizers {
 		if strings.HasPrefix(finalizer, prefix) {
-			logrus.Debugf("a finalizer with prefix %s will be removed", prefix)
+			logrus.Debugf("finalizer with prefix %s will be removed", prefix)
```
annablender, line 423: nit: put brackets around resource names `finalizer with prefix [%s] will be removed`

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
```diff
@@ -448,14 +445,18 @@ func (m *nodesSyncer) reconcileAll() error {
...
 	if machine == nil {
 		return m.createNode(node, nodeCache)
 	}
 	return m.updateNode(machine, node)
 }
 
-func (m *nodesSyncer) removeNode(machine *v3.Node) error {
+func (m *nodesSyncer) removeNode(machine *v32.Node) error {
+	if machine.DeletionTimestamp != nil {
+
```
annablender, line 457: nit: remove extra line

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
```diff
@@ -826,7 +827,7 @@ func (m *nodesSyncer) isClusterRestoring() (bool, error) {
 	return false, nil
 }
 
-func determineNodeRoles(machine *v3.Node) {
+func determineNodeRoles(machine *v32.Node) {
```
annablender, line 830: Just curious, but what is the *v32 node obj as opposed to the v3?

annablender (changes requested):

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
jakefhyde:
```
	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
```
using `v3.Node` in this case is bad, because `v3.Node` is just defined as the following:

```
// Deprecated: use v3.Node instead
type Node = v3.Node
```
where in the above example, `v3.Node` is actually `v32.Node`, i.e. all api resources defined in `github.com/rancher/rancher/pkg/generated/norman/` are just deprecated aliases to the real API types. The deprecated warnings also show up in my IDE and I hate them, but not in a "I want to turn my warnings off" kind of way, more like a "why are we still using these deprecated types" kind of way. Some people use `mgmtv3`, or `apimgmtv3`, or `v32`. I'm actually just going to update it to be `apimgmtv3`.

HarrisonWAffel (dismissed):

pkg/controllers/managementuser/nodesyncer/nodessyncer.go
annablender: I see the deprecated warnings too when I use `mgmtv3 Cluster.` The Cluster var is crossed out, not sure why we're still using the generated alias types either

annablender (approved):
Thanks for the quick response! LGTM

jakefhyde:
@HarrisonWAffel @annablender @jiaqiluo Had to change destination branch to release/v2.7 due to the new branching strategy.

HarrisonWAffel (approved):

annablender (approved):

PR #38859, opened 2022-09-09
Update dependencies after branching for v2.7

prachidamle:
Updating the dependencies in rancher/rancher after branches for v2.7 are created.

Versions for UI, DASHBOARD and CLI will be the same as 2.6 until new releases are made to these

Updated RKE to latest release  v1.3.14

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -15,9 +15,7 @@ WORKDIR /var/lib/rancher
 
 ARG ARCH=amd64
 ARG IMAGE_REPO=rancher
-# System-charts will continue to point to dev-v2.6 branch since eventually in 2.7.x the plan is
-# to remove the v1 tools and remove this dependency
-ARG SYSTEM_CHART_DEFAULT_BRANCH=dev-v2.6
+ARG SYSTEM_CHART_DEFAULT_BRANCH=dev-v2.7
 ARG CHART_DEFAULT_BRANCH=dev-v2.7
 ARG PARTNER_CHART_DEFAULT_BRANCH=main
 ARG RKE2_CHART_DEFAULT_BRANCH=main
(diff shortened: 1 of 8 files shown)


aiyengar2 (approved):

package/Dockerfile
```diff
@@ -15,12 +15,14 @@ WORKDIR /var/lib/rancher
 
 ARG ARCH=amd64
 ARG IMAGE_REPO=rancher
+# System-charts will continue to point to dev-v2.6 branch since eventually in 2.7.x the plan is
+# to deprecate the v1 tools and remove this dependency
```
deniseschannon, line 19: I'd update to use the language "remove the v1 tools"

scripts/package
```diff
@@ -5,9 +5,11 @@ source $(dirname $0)/version
 
 ARCH=${ARCH:-"amd64"}
 SYSTEM_CHART_REPO_DIR=build/system-charts
+# System-charts will continue to point to dev-v2.6 branch since eventually in 2.7.x the plan is
+# to deprecate v1 tools and remove this dependency
```
deniseschannon, line 9: update to "to remove v1 tools"

deniseschannon (changes requested):
What about this reference? Should this be set at all based on the comment? 

https://github.com/rancher/rancher/blob/release/v2.7/Dockerfile.dapper#L15

package/Dockerfile
prachidamle: updated

scripts/package
prachidamle: updated

prachidamle:
> What about this reference? Should this be set at all based on the comment?
> 
> https://github.com/rancher/rancher/blob/release/v2.7/Dockerfile.dapper#L15

@deniseschannon updated that reference to dev-v2.7 as well. 
Was not sure if we need it anymore from the comment in the checklist about 2.4 and kdm vendoring; but will keep it updated to not break anything

aiyengar2 (approved):

deniseschannon (approved):

prachidamle:
Validated that the CI drone check has passed, merging in the PR.

PR #38866, opened 2022-09-12
Enable dapper on M1 Macs

moio:
## Problem
`make` currently exists on M1 Macs. Nevertheless, appears to be working fine as of September 2022.

I am unaware of the exact cause why it was blocked before, but it seems to be fixed now
 
## Solution

Revert https://github.com/rancher/rancher/pull/36487
 
## Testing
On an M1 Mac, run `make`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,11 @@
 TARGETS := $(shell ls scripts)
 
 .dapper:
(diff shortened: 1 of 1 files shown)


moio:
An identical patch was merged as https://github.com/rancher/rancher/pull/39228

PR #38867, opened 2022-09-12
Use debugger-friendly compile settings if DEBUG is set

moio:
Issue #39491 

## Problem
In order to use a debugger such as [Delve](https://github.com/go-delve/delve), binaries have to be compiled with debugging symbols and disabling some optimizations.

This PR makes it possible to build binaries and container images in such a way.

I maintain a [companion project to run Delve as an ephemeral container](https://github.com/rancherlabs/delve-debugger) to make attaching a debugger as easy as possible ([instructions](https://github.com/rancherlabs/delve-debugger/blob/master/util/README-rancher.md)).
 
## Solution
Pass the DEBUG variable to build scripts and, if set, make sure `-gcflags="all=-N -l"` is set and `-ldflags -s -w` is not set.

Similar patches have been merged to [fleet](https://github.com/rancher/fleet/pull/961) and [k3s](https://github.com/k3s-io/k3s/pull/6147).
 
## Testing
- `make` should continue to work as before
- `DEBUG=true make` should build as well
- to verify that it is actually possible to attach ad debugger, see [instructions](https://github.com/moio/delve-debugger/blob/master/util/README-rancher.md)

The only place this PR could possibly break is the build system, but care was taken to leave everything unchanged unless the `DEBUG` environment variable is set at `make` call time.


rmweir (dismissed):
LGTM

cmurphy (approved):

rmweir (approved):

PR #38622, opened 2022-08-16
Update `rancher/shell` version to non-rc `v0.1.18`

macedogm:
Update `rancher/shell` version to non-rc `v0.1.18`.

Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>

Issue #38834, opened 2022-09-07
[BUG] proxy container of the helm-operation pod is not terminated and it becomes NotReady


alatui:
**Rancher Server Setup**
- Rancher version: v2.6.8 
- Installation option (Docker install/Helm Chart): Helm Chart
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): v1.23.9+k3s1
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: v1.23.9+k3s1
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Imported

**User Information**
- What is the role of the user logged in? Admin
- If custom, define the set of permissions: N/A

**Describe the bug**
After successfully installing any chart through the Rancher UI (Apps menu) the `proxy` container of the `helm-operation-xxxxxx` pod is not properly terminated and it receives status `NotReady`. 

**To Reproduce**
Install any chart using the Rancher UI and check for the status of the `helm-operation-xxxxx` that was created with `kubectl get pods -n cattle-system | grep helm`

**Result**
The `proxy` container of the `helm-operation-xxxxxx` pod is not properly terminated and it receives status `NotReady`. 

**Expected Result**
Both containers (proxy and helm) of the `helm-operation-xxxxxx` to proper terminate and the pod to receive the status `Completed`

**Screenshots**
<img width="760" alt="Screenshot 2022-09-07 at 18 02 34" src="https://user-images.githubusercontent.com/1859146/188937414-feaabf2d-afdc-489b-bb18-042fc5eccd4a.png">

**Additional context**
I tested using Rancher v2.6.5 and it works fine.
Logs of the `proxy` container:
```
W0907 16:36:08.442267       7 proxy.go:162] Request filter disabled, your proxy is vulnerable to XSRF attacks, please be cautious
Starting to serve on 127.0.0.1:8001
```

raelix:
I'm also hitting this issue

KeithMnemonic:
I am also seeing something similar. My logs show

```
[...]

PR #38870, opened 2022-09-12
[2.7] Use `pkill` instead of `killall` for `helm-operation`

macedogm:
## Problem

After successfully installing any chart through the Rancher UI (Apps menu) the `proxy` container of the `helm-operation-xxxxx` pod is not properly terminated and it receives status `NotReady`. This happens because we migrated the old `rancher/shell` image from Alpine to SLE-BCI (in v2.6.7 - https://github.com/rancher/rancher/pull/38622) and it doesn't contains the `killall` command (from `psmisc` program) used by [Helm operations](https://github.com/rancher/rancher/blob/v2.6.8/pkg/controllers/dashboard/helm/operations.go#L136).

## Solution

This PR modifies [Helm operations](https://github.com/rancher/rancher/blob/v2.6.8/pkg/controllers/dashboard/helm/operations.go#L136) to use the `pkill` command instead of `killall`.

## Testing

To reproduce this issue, install any valid chart from Apps and wait for it to finish the installation. The `proxy` container from `helm-operation-xxxxx` pod will not be terminated.

### Manual Testing

Install any valid chart from Apps and wait for it to finish the installation. All `helm-operation-xxxxx` pods must be properly terminated.

### Automated Testing

Automated testing will be added on a separate PR.

No regression is expected.

### DIFF
--- a/pkg/controllers/dashboard/helm/operations.go
+++ b/pkg/controllers/dashboard/helm/operations.go
@@ -133,7 +133,7 @@ func (o *operationHandler) cleanup(pod *corev1.Pod) error {
 			Stdout:    true,
 			Stderr:    true,
 			Container: "proxy",
-			Command:   []string{"killall", "kubectl"},
+			Command:   []string{"pkill", "-x", "kubectl"},
 		}, scheme.ParameterCodec).Do(o.ctx)
 	return result.Error()
 }


MbolotSuse:
Related to rancher/rancher#38834

MbolotSuse (approved):
LGTM as long as we get tests in another PR

PR #38878, opened 2022-09-12
[2.7] Validate config migration process for Azure AD

maxsokolovsky:
## Problem
Rancher must fail the AuthConfig migration process if the Azure App registrations does not have the necessary permissions configured prior to migration.
 
## Solution
Verify that admins have properly configured an existing app registration's permissions
in the Azure portal before they update their Azure AD auth config to use the new authentication flow.
Fetch the current AuthConfig from the database, update it in-memory to use the new endpoints,
and create a new test Azure client, thereby getting an access token to the Graph API.
Then parse the JWT and inspects the permissions contained within. If the admins had not set those as per docs,
then Rancher won't find the permissions in the test token and will return an error.

I am integrating the code from the https://github.com/rancher/rancher/issues/38754 issue, since these issues are linked, and testing would be easier if these two changes are introduced in a single PR.
1. Immediately throw an error if an access token fetched either from cache or Microsoft lacks permissions.
2. Validate this access token on auth config migration (and thus also do not cache it if any permissions are missing).

### DIFF
--- a/pkg/auth/providers/azure/azure_actions.go
+++ b/pkg/auth/providers/azure/azure_actions.go
@@ -41,6 +41,9 @@ func (ap *azureProvider) actionHandler(actionName string, action *types.Action, 
 	} else if actionName == "testAndApply" {
 		return ap.testAndApply(actionName, action, request)
 	} else if actionName == "upgrade" {
+		if err := ap.checkConfigurationBeforeMigration(); err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


KevinJoiner:
[Backport PR](https://github.com/rancher/rancher/pull/38776)

KevinJoiner (approved):

cbron (approved):
looks like a plain forwardport, should be good to merge

PR #38894, opened 2022-09-13
Refresh ExtrasByProvider from derived tokens

cmurphy:
If a user has no fresh UI login tokens, they may still need to log in using a kubeconfig or API token. Without this change, if the provider refresher happens to have run, ExtraByProvider will become empty, which will cause requests to fail because the impersonation logic depends on extras found in UserAttributes. This change ensures that extras are refreshed either from login tokens or derived tokens, as long as the user still has access to the provider.

## Problem

The problem this PR addresses is that the refresher may set the `ExtraByProvider` field in a `UserAttribute` resource to empty, `{}`. if the user has no active login tokens. Login tokens come from logging into the UI, but some users rarely or never log into the UI, for example automated Jenkins or Terraform users. Impersonation uses `ExtraByProvider` to set impersonation clusterrole rules, so if it is empty the user will run into a "Forbidden" error.

Some users have reported seeing the `ExtraByProvider` field as `null` rather than `{}`, which would have the same effect on user requests. If the refresher is run, this should correct that issue too, but it does not address why the field would get into that state in the first place.
 
## Solution

The refresher needs to also check "derived" tokens, such as the ones from generated kubeconfigs, for user extras. 
 
### Manual Testing

* Create a new user
* Add user as an owner on a non-local cluster
* Log in as that user and generate a kubeconfig for the cluster
* On the local cluster, delete the user's userattributes `kubectl delete userattributes u-XXXX`
* Log out and log back in as admin
* Click "refresh group membership" in the user management page for that user
* User should still be able to authenticate to the downstream cluster with the downloaded kubeconfig
* `ExtraByProvider` in userattribute resource should contain principalid and username for enabled providers

## QA Testing Considerations

* Test with multiple auth providers, especially Shibboleth and other SAML providers
* Test upgrading from v2.6.2 to v2.6-head/v2.7-head
 
* Shibboleth needs special consideration because it is labeled as a refreshable provider in contrast with other SAML providers.
* Ensure that disabling/removing a user in an auth provider deactivates the user in Rancher as before this change

### DIFF
--- a/pkg/auth/providerrefresh/refresher.go
+++ b/pkg/auth/providerrefresh/refresher.go
@@ -148,7 +148,8 @@ func (r *refresher) triggerUserRefresh(userName string, force bool) {
 
 func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttribute, error) {
 	var (
-		derivedTokens         []*v3.Token
+		derivedTokenList      []*v3.Token
+		derivedTokens         map[string][]*v3.Token
 		loginTokenList        []*v3.Token
 		loginTokens           map[string][]*v3.Token
 		canLogInAtAll         bool
@@ -163,6 +164,7 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 	}
 
 	loginTokens = make(map[string][]*v3.Token)
+	derivedTokens = make(map[string][]*v3.Token)
 
 	allTokens, err := r.tokenLister.List("", labels.Everything())
 	if err != nil {
@@ -171,6 +173,7 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 
 	for providerName := range providers.ProviderNames {
 		loginTokens[providerName] = []*v3.Token{}
+		derivedTokens[providerName] = []*v3.Token{}
 	}
 
 	for _, token := range allTokens {
@@ -179,7 +182,8 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 		}
 
 		if token.IsDerived {
-			derivedTokens = append(derivedTokens, token)
+			derivedTokens[token.AuthProvider] = append(derivedTokens[token.AuthProvider], token)
+			derivedTokenList = append(derivedTokenList, token)
 		} else {
 			loginTokens[token.AuthProvider] = append(loginTokens[token.AuthProvider], token)
 			loginTokenList = append(loginTokenList, token)
@@ -233,14 +237,12 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 						if existingPrincipals != nil {
 							newGroupPrincipals = existingPrincipals
 						}
-						continue
+					} else {
+						// In the case that the user explicitly cannot login at all to this provider
+						// (e.g. they no longer exist) we pretend they have no principal with this provider
+						// so that their login tokens get blanked out
+						principalID = ""
 					}
-
-					// In the case that the user explicitly cannot login at all to this provider
-					// (e.g. they no longer exist) we pretend they have no principal with this provider
-					// so that their login tokens get blanked out
-					principalID = ""
-
 				}
 			}
 		}
@@ -251,24 +253,9 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 
 		attribs.GroupPrincipals[providerName] = v32.Principals{Items: newGroupPrincipals}
 
-		if principalID != "" && len(loginTokens[providerName]) > 0 {
-			// A user is 1:1 with its principal for a given provider, no need to get principals from tokens beyond the first one
-			userPrincipal, err := providers.GetPrincipal(principalID, *loginTokens[providerName][0])
-			if err != nil {
-				return nil, err
-			}
-			userExtraInfo := providers.GetUserExtraAttributes(providerName, userPrincipal)
-			if userExtraInfo != nil {
-				if attribs.ExtraByProvider == nil {
-					attribs.ExtraByProvider = make(map[string]map[string][]string)
-				}
-				attribs.ExtraByProvider[providerName] = userExtraInfo
-			}
-		}
-
 		canAccessProvider := false
 
-		if principalID != "" {
+		if principalID != "" && !errorConfirmingLogins {
 			// We want to verify that the user still has rancher access
 			canStillAccess, err := providers.CanAccessWithGroupProviders(providerName, principalID, newGroupPrincipals)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/auth/providerrefresh/refresher.go
```diff
@@ -281,6 +270,27 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 			}
 		}
 
+		if principalID != "" && (len(loginTokens[providerName]) > 0 || (len(derivedTokens[providerName]) > 0 && canAccessProvider)) {
```
MbolotSuse, line 273: I think I understand part of this if condition, but I'm not super sure I understand why we want to check `&& canAccessProvider`. It would seem to me that the below nil pointer issues can be prevented by the first part of this check.

I guess what I am asking is what value do we get out of the above as opposed to:

```go
if principalID != "" && (len(loginTokens[providerName]) > 0 || len(derivedTokens[providerName]) > 0)
```

I'm not saying we need to change it, just would like a better understanding of why the if statement is structured this way.

pkg/auth/providerrefresh/refresher_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+func Test_refreshAttributes(t *testing.T) {
+	tests := []struct {
+		name    string
+		user    *v3.User
+		attribs *v3.UserAttribute
+		tokens  []*v3.Token
+		want    *v3.UserAttribute
+	}{
+		{
```
MbolotSuse, line 28: Nit: 

I think right now we have test cases for:
- No token
- 1 login no derived
- 1 derived no login

It might be good to include a case with 1 login and 1 derived (and assert that it uses what is on the login). Especially since I would expect this to be a relatively common case.

pkg/auth/providerrefresh/refresher_test.go
```diff
@@ -0,0 +1,192 @@
...
+	panic("not implemented")
+}
+func (p *mockLocalProvider) TransformToAuthProvider(authConfig map[string]interface{}) (map[string]interface{}, error) {
+	panic("not implemented")
+}
+func (p *mockLocalProvider) RefetchGroupPrincipals(principalID string, secret string) ([]v3.Principal, error) {
+	return []v3.Principal{}, nil
+}
+func (p *mockLocalProvider) CanAccessWithGroupProviders(userPrincipalID string, groups []v3.Principal) (bool, error) {
+	return true, nil
```
MbolotSuse, line 185: I think that if we keep the if statement in refresher.go as is, we want a test case or two where the user couldn't access the provider, so even though we had tokens we didn't use them and the user principals were empty.

MbolotSuse (changes requested):
The overall approach looks good. I did have a question on one if statement and a couple of test cases that I think would be good to add.

pkg/auth/providerrefresh/refresher.go
cmurphy: Added a comment to explain.

pkg/auth/providerrefresh/refresher_test.go
cmurphy: Added this test

pkg/auth/providerrefresh/refresher_test.go
cmurphy: Added

MbolotSuse (dismissed):
Overall LGTM.

One minor nit: I think this could still use a test case with a user who is disabled but still has an active login token. I'll call this a nit and not needed since from a holistic perspective this case shouldn't occur, but from the perspective of this code it could happen. That being said, I think we have most of what we need here, so I'm ok approving without it.

cmurphy:
Noticed an issue with the shibboleth provider, updated to fix

pkg/auth/providerrefresh/refresher.go
```diff
@@ -281,8 +268,31 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
 			}
 		}
 
+		// Update extras if either the user has an active login token, or an API token/kubeconfig token and is still active in the auth provider.
+		// If the user cannot access the auth provider, the derived tokens are deactivated later and should not be used to determine extra attributes.
```
cbron, line 272: nit: 
```suggestion
		// If the user cannot access the auth provider, the derived tokens are deactivated below and should not be used to determine extra attributes.
```

pkg/auth/providerrefresh/refresher.go
```diff
@@ -281,8 +268,31 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
...
+			userPrincipal, err := providers.GetPrincipal(principalID, token)
+			if err != nil {
+				return nil, err
+			}
+			userExtraInfo := providers.GetUserExtraAttributes(providerName, userPrincipal)
+			if userExtraInfo != nil {
+				if attribs.ExtraByProvider == nil {
+					attribs.ExtraByProvider = make(map[string]map[string][]string)
+				}
+				attribs.ExtraByProvider[providerName] = userExtraInfo
```
cbron, line 290: We should make sure to test this on both an external provider and local.

cbron (approved):
lgtm, thanks for adding tests

pkg/auth/providers/providers.go
```diff
@@ -27,15 +27,15 @@ var (
 	ProviderNames          = make(map[string]bool)
 	ProvidersWithSecrets   = make(map[string]bool)
 	UnrefreshableProviders = make(map[string]bool)
-	providers              = make(map[string]common.AuthProvider)
+	Providers              = make(map[string]common.AuthProvider)
```
rmweir, line 30: Why does this need to be exported?

rmweir (changes requested):

pkg/auth/providers/providers.go
cmurphy: The tests need to set it https://github.com/rancher/rancher/pull/38894/commits/5ebecc2cb448506abaea0e6055cc6692b272650e#diff-669858130ce986f4a18774a2b9b922d11d6cd64d1e8be3c308d8057e007ec0c7R283

We can't just call providers.Configure because there's no ScaledContext in the unit tests.

pkg/auth/providerrefresh/refresher.go
```diff
@@ -251,24 +253,9 @@ func (r *refresher) refreshAttributes(attribs *v3.UserAttribute) (*v3.UserAttrib
...
-					attribs.ExtraByProvider = make(map[string]map[string][]string)
-				}
-				attribs.ExtraByProvider[providerName] = userExtraInfo
-			}
-		}
-
 		canAccessProvider := false
 
-		if principalID != "" {
+		if principalID != "" && !errorConfirmingLogins {
```
cmurphy, line 258: Added to address an issue pointed out by @MbolotSuse which is that we're not 100% sure what the effect of calling `CanAccessWithGroupProviders` would be if we had run into an access error on line 229, so adding this should restore the previous behavior of skipping this section in such a case.

pkg/auth/providerrefresh/refresher.go
cmurphy: fixed

MbolotSuse (approved):
LGTM

rmweir (approved):
LGTM

PR #38868, opened 2022-09-12
Use `pkill` instead of `killall` for `helm-operation`

macedogm:
Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>

PR #38911, opened 2022-09-14
Add validation test for helm-operation pods

macedogm:
## Problem

This PR is adding only the validation tests to check if all `helm-operation-xxxxx` pods have been successfully terminated. The tests was missing from PR https://github.com/rancher/rancher/pull/38868.

After successfully installing any chart through the Rancher UI (Apps menu) the `proxy` container of the `helm-operation-xxxxx` pod is not properly terminated and it receives status `NotReady`. This happens because we migrated the old `rancher/shell` image from Alpine to SLE-BCI (in v2.6.7 - https://github.com/rancher/rancher/pull/38622) and it doesn't contains the `killall` command (from `psmisc` program) used by [Helm operations](https://github.com/rancher/rancher/blob/v2.6.8/pkg/controllers/dashboard/helm/operations.go#L136).

## Solution

This PR modifies [Helm operations](https://github.com/rancher/rancher/blob/v2.6.8/pkg/controllers/dashboard/helm/operations.go#L136) to use the `pkill` command instead of `killall`.

## Testing

To reproduce this issue, install any valid chart from Apps and wait for it to finish the installation. The `proxy` container from `helm-operation-xxxxx` pod will not be terminated.

### Manual Testing

Install any valid chart from Apps and wait for it to finish the installation. All `helm-operation-xxxxx` pods must be properly terminated.

### Automated Testing

**Rancher v2.6.6 test == PASS**
1. Install a test cluster with Rancher `v2.6.6`.
2. Configure env var `CATTLE_TEST_CONFIG` to point to the test cluster.
3. `cd tests/v2/validation/charts`
4. `go test -v`
5. The tests `TestVerifyNoPendingHelmOp` must `PASS`.

**Rancher v2.6.8 test == FAIL**
1. Install a test cluster with Rancher `v2.6.8`.
2. Configure env var `CATTLE_TEST_CONFIG` to point to the test cluster.
3. `cd tests/v2/validation/charts`
4. `go test -v`
5. The tests `TestVerifyNoPendingHelmOp` must `FAIL`.

**Rancher v2.6-head test == PASS**
1. Install a test cluster with Rancher `v2.6-head`.
2. Configure env var `CATTLE_TEST_CONFIG` to point to the test cluster.
3. `cd tests/v2/validation/charts`
4. `go test -v`
5. The tests `TestVerifyNoPendingHelmOp` must `PASS`.

No regression is expected.

### DIFF
--- a/tests/framework/extensions/workloads/podstatus.go
+++ b/tests/framework/extensions/workloads/podstatus.go
@@ -3,12 +3,15 @@ package workloads
 import (
 	"context"
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/rancher/rancher/pkg/api/scheme"
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // PodGroupVersion is the required Group Version for accessing pods in a cluster,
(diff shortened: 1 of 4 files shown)


tests/framework/extensions/workloads/podstatus.go
```diff
@@ -45,15 +48,35 @@ func StatusPods(client *rancher.Client, clusterID string, listOpts metav1.ListOp
...
 }
+
+// WaitPodTerminated is a helper function that uses wait.Poll() to verify if all pods with podName
+// have terminated correctly or if there is still a pod runnnig.
+func WaitPodTerminated(client *rancher.Client, clusterID string, podName string) bool {
+	err := wait.Poll(5*time.Second, 2*time.Minute, func() (done bool, err error) {
+		podResults, _ := StatusPods(client, clusterID, metav1.ListOptions{})
+		for _, pod := range podResults {
+			p := strings.Split(pod, ": ")
+			if strings.HasPrefix(p[1], podName) && strings.HasPrefix(p[2], "Running") {
```
MbolotSuse, line 70: My major concern about this function is that it could be vacuously true. That is to say, if it doesn't find any pod with the given prefix, we are going to succeed and pass the test despite this being broken and not finding any pods. If you have ran a live test and made sure that this is actually finding the pod as expected we can move it forward, but otherwise, it might be good to make sure that len(podResults) > 0 or something similar.

MbolotSuse (changes requested):
Overall very good approach. I think there's a minor improvement that could be made, and I have one outstanding concern on the new `WaitPodTerminated` function.

Issue #38854, opened 2022-09-09
[BUG] Random panic encountered with managementapi/dynamicschema concurrent map read/write


Oats87:
**Rancher Server Setup**
- Rancher version: v2.6.6
- Installation option (Docker install/Helm Chart): Docker install
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

On a fresh `docker run` install of Rancher v2.6.6, I encountered a panic due to a concurrent map read/write. Conveniently, I run my Rancher with trace logs and caught the corresponding trace log for it.

Relevant trace:
```
2022/09/09 01:45:36 [TRACE] ClusterController calling handler mgmt-cluster-rbac-remove _all_
2022/09/09 01:45:36 [INFO] uploading harvestercredentialConfig to credentialconfig schema
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//dynamicschemas/credentialconfig
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//nodedrivers/exoscale
2022/09/09 01:45:36 [TRACE] NodeDriverController calling handler whitelist-proxy google
2022/09/09 01:45:36 [TRACE] NodeDriverController calling handler node-driver-controller google
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//nodedrivers/rackspace
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//nodedrivers/vmwarevsphere
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//nodedrivers/packet
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//nodedrivers/pinganyunecs
2022/09/09 01:45:36 [TRACE] DynamicSchemaController calling handler dynamic-schema credentialconfig
2022/09/09 01:45:36 [TRACE] REST UPDATE apis/management.cattle.io/v3//nodedrivers/pnap
2022/09/09 01:45:36 [TRACE] ClusterController calling handler temporary-credentials local
2022/09/09 01:45:36 [TRACE] ClusterController calling handler restrictedAdminsRBACCluster local
2022/09/09 01:45:36 [TRACE] ClusterController calling handler cluster-secret-migrator local
[...]

PR #38920, opened 2022-09-14
[release/v2.7] DeepCopy nodeSchema after listing from cache

Oats87:
## Issue: 
https://github.com/rancher/rancher/issues/38854
 
## Problem
In certain cases, it is possible to have Rancher panic due to the `machine_driver.go` controller, as it unsafely mutates objects returned from the cache.
 
## Solution
DeepCopy the object before allowing mutation.
 
N/A


snasovich (approved):
Same as https://github.com/rancher/rancher/pull/38919 that has 2 approvals. Merging.

PR #38939, opened 2022-09-15
Add logic to sync generated node role from upstream

annablender:


jakefhyde (approved):
LGTM, pendings eks-operator release

HarrisonWAffel (approved):

pkg/apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -394,6 +394,7 @@ type EKSStatus struct {
 	PrivateRequiresTunnel         *bool                       `json:"privateRequiresTunnel"`
 	ManagedLaunchTemplateID       string                      `json:"managedLaunchTemplateID"`
 	ManagedLaunchTemplateVersions map[string]string           `json:"managedLaunchTemplateVersions"`
+	GeneratedNodeRole             string                      `json:"generatedNodeRole"`
```
rmweir, line 397: Why do we need to add generated noderole to both the cluster status and the EKSClusterConfig object in the operator? If rancher is generating the role then I assume the EKSCC object shouldn't need it and rancher will just write it to NodeRole? I see we do this in other places but do not remember what the purpose was.

rmweir (changes requested):

pkg/apis/management.cattle.io/v3/cluster_types.go
annablender: The original idea here was to design the `GeneratedNodeRole` field the way `ManagedTemplateID` is. The EKSCC object needs the generated node role because it checks for the generated value on `config.Status.GeneratedNodeRole.` If the user didn't specify a custom node role for a node group in the rancher ui, and a generated node role has already been created (which the EKS operator would only know by looking at the EKSCC status), then it creates the node group with that generated role.

rmweir: Oh, I forgot that on the cluster obj we only have the spec and not the status from the ekscc object. Sounds good!

rmweir (approved):

HarrisonWAffel (approved):

rmweir (approved):
LGTM

pkg/controllers/management/eks/eks_cluster_handler.go
```diff
@@ -304,14 +304,24 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
...
 			if err != nil {
 				return cluster, err
 			}
 		}
 
+		generatedNodeRole, _ := status["generatedNodeRole"].(string)
+		if generatedNodeRole != "" && cluster.Status.EKSStatus.GeneratedNodeRole != generatedNodeRole {
+			cluster = cluster.DeepCopy()
+			cluster.Status.EKSStatus.GeneratedNodeRole = generatedNodeRole
+			cluster, err = e.ClusterClient.Update(cluster)
```
jakefhyde, line 319: I'm going to rubber stamp this because the rest of the file is like this, but this is another case where leveraging different controller framework paradigms (like `UpdateStatus`) may be more appropriate.

jakefhyde (approved):

PR #38950, opened 2022-09-16
Support cluster specific registry for managed fleet agent

manno:
Refers to https://github.com/rancher/fleet/issues/948

### DIFF
--- a/pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go
+++ b/pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go
@@ -11,6 +11,7 @@ import (
 	fleetcontrollers "github.com/rancher/rancher/pkg/generated/controllers/fleet.cattle.io/v1alpha1"
 	mgmtcontrollers "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	rocontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/provisioningv2/image"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/rancher/pkg/wrangler"
 	"github.com/rancher/wrangler/pkg/apply"
@@ -125,6 +126,7 @@ func (h *handler) createCluster(cluster *v1.Cluster, status v1.ClusterStatus) ([
 			KubeConfigSecret: clientSecret,
 			AgentEnvVars:     mgmtCluster.Spec.AgentEnvVars,
 			AgentNamespace:   agentNamespace,
+			PrivateRepoURL:   image.GetPrivateRepoURLFromCluster(cluster),
 		},
 	}}, status, nil
 }


mattfarina:
@manno the failed CI attempt was due to a flake in the CI system. This is where CI needs to be re-run.

pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go
```diff
@@ -125,6 +126,7 @@ func (h *handler) createCluster(cluster *v1.Cluster, status v1.ClusterStatus) ([
 			KubeConfigSecret: clientSecret,
 			AgentEnvVars:     mgmtCluster.Spec.AgentEnvVars,
 			AgentNamespace:   agentNamespace,
+			PrivateRepoURL:   image.GetPrivateRepoURLFromCluster(cluster),
```
Oats87, line 129: I believe we will run into an issue with this, as this only will detect a private registry if it is either set as a system default registry or via the machineGlobalData/machineConfigData. It will not detect a cluster-level registry

@HarrisonWAffel just went through the exercise of figuring out which method can be used to find the "correct" private repo -- perhaps we can use the logic from that over here.

Oats87 (changes requested):

pkg/controllers/provisioningv2/fleetcluster/fleetcluster.go
HarrisonWAffel: I think that this is actually OK for RKE2/K3s clusters. When creating a RKE2/K3s cluster via the UI and supplying a cluster-level registry, the URL will always be placed in the `machineSelectorConfig`. This function was also used in the other PR for the cluster agent manifest, and has used to [determine the desired version of the agent ](https://github.com/HarrisonWAffel/rancher/blob/c0ba0f0c94a1e43885a6821a5403c5d600ef1246/pkg/controllers/provisioningv2/cluster/controller.go#L277) for a little while now.

Oats87: @HarrisonWAffel ok, thank you for the clarification.

Do you know what the purpose of the registry field on the spec is then and when that would be set? I'm worried we will have users setting that instead of setting the machineSelectorConfig -- we can't always assume a user will mimic what the UI does.

HarrisonWAffel: @Oats87 Right now the `registries` section is used to hold registry auth configs, and has a similar format to `registries.yaml`, here's what it looks like when created by rancher without tls enabled
```
    registries:
      configs:
        [PRIVATE_REGISTRY_URL]:
          authConfigSecretName: registryconfig-auth-5fqst
          caBundle: null
          insecureSkipVerify: false
          tlsSecretName: null
```
So, while the `registries` section can contain authentication configs for multiple registries, the `machineSelectorConfig` will denote the system-default registry where system-images should be pulled from.

Oats87: Makes sense -- thanks for the clarification!

In this case though, do we need to support authenticating against the private registry being used @manno @mattfarina ?

I'm not sure if Fleet supports setting an image pull secret or something along that path.

mattfarina: @Oats87 Fleet doesn't support auth at the moment. I've updated the existing issue for us to add it. Hopefully for 2.7.1.

Do we need auth for 2.7.0 is more of a question for mgmt.

deniseschannon: What happens if you have a cluster registry with auth? Does it fail?

mattfarina (approved):
Tested with auth using the registries.yaml option in k3s to set a registry with authentication. Fleet is able to install its agent using a custom registry with authentication. Fleet has the custom registry configured on the cluster (which is what this change does) while the auth is handled by the distro. This is what, I think, other feature charts are doing.

Oats87 (approved):
Yep, provided that auth works and we can allow for cluster-level-auth to happen via the `/etc/rancher/<distro>/registries.yaml` file, this should be good. system-default-registry w/ auth is still a pain point, but can be worked around via the `registries.yaml` file as well.

PR #38960, opened 2022-09-16
[release/v2.6] Updating to Fleet 0.4.0-rc3 and the version of Go used for building to 1.19

mattfarina:
Notes:
- This updates some Kubernetes packages to newer patch versions which include numerous security fixes.
- Go is updated to 1.19. The previous minor version, 1.17, is no longer supported by the Go community.
- Fleet updated one dependency to a version that requires Go 1.18 or newer. Updating Rancher to a newer version of Golang brought in a newer version of `gofmt`. This newer version failed with linting in CI due to code comment changes in the way Go works with and looks at them. The resulting fix for linting touched over 180 files. It was automated with `gofmt -e -w .`.
- There are separate commits so the automated formatting changes can be reviewed from the other changes.
- The Go changes here will be brought to the release 2.7 branch once this approved and merged.


aiyengar2 (approved):

Oats87:
@mattfarina Is there any way you can make the name of the PR also include the fact that this is bumping golang to `v1.19`? I understand that the bump to golang is due to the dependency of fleet, but bumping golang is a pretty large change that I don't think we'll want to miss out on release notes when we generate them/PR title skimming.

Oats87 (approved):

kinarashah (approved):
We should also try and update go version in our go.mod, will get that in a separate PR. Merging this to get a rancher RC with fleet `100.1.0+up0.4.0-rc3`.

PR #38972, opened 2022-09-16
[release/v2.7] Updated fields that previously used Duration to use string, to fix an unmarshaling error

igomez06:
## Problem

Since the native Duration type used to generate the appropriate types for the provisioning and cluster clients, doesn't use a native type, and has `protobuf` tag, it doesn't get generated appropriately and cause unmarshaling errors. 
 
## Solution

Updated schemas to change the field types to strings to fix the error.

### DIFF
--- a/tests/framework/pkg/schemas/cluster.x-k8s.io.machines/v1beta1/schema.go
+++ b/tests/framework/pkg/schemas/cluster.x-k8s.io.machines/v1beta1/schema.go
@@ -2,8 +2,7 @@ package schema
 
 import (
 	"github.com/rancher/norman/types"
-	provisioningSchema "github.com/rancher/rancher/tests/framework/pkg/schemas/provisioning.cattle.io/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	m "github.com/rancher/norman/types/mapper"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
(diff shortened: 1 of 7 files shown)


igomez06:
Backport: https://github.com/rancher/rancher/pull/38973

caliskanugur (approved):
LGTM

tests/framework/pkg/schemas/cluster.x-k8s.io.machines/v1beta1/schema.go
```diff
@@ -2,8 +2,7 @@ package schema
 
 import (
 	"github.com/rancher/norman/types"
-	provisioningSchema "github.com/rancher/rancher/tests/framework/pkg/schemas/provisioning.cattle.io/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	m "github.com/rancher/norman/types/mapper"
```
izaac, line 5: This import alias isn't very descriptive but I can't think of a better name `normanTypesMapper` ? 
Im not sure what's the best option to follow conventions (if any) here or just leave it as `m`

izaac (commented):
A comment/question

tests/framework/pkg/schemas/cluster.x-k8s.io.machines/v1beta1/schema.go
igomez06: Yeah I can definitely do that. I'll do `normanMapper`

igomez06: Hey Izaac I can changed this, but this is how the rest of the rancher/rancher repo has the alias for the mapper import, in the other schemas. I think because of the precedent of the rest of the repo, I think it's best to just leave it as `m`

izaac: I didn't know that and is a good point let's leave it as `m` for consistency

bmdepesa (approved):

PR #38973, opened 2022-09-16
[release/v2.6] Updated fields that previously used Duration to use string, to fix an unmarshaling error

igomez06:
## Problem

Since the native Duration type used to generate the appropriate types for the provisioning and cluster clients, doesn't use a native type, and has `protobuf` tag, it doesn't get generated appropriately and cause unmarshaling errors. 
 
## Solution

Updated schemas to change the field types to strings to fix the error.


caliskanugur (approved):
LGTM

izaac (approved):
LGTM

PR #38990, opened 2022-09-17
[release/v2.7] Add specific watch for secrets in rke2configserver to ensure v2prov provisioning timeliness

Oats87:
## Problem
With K8s v1.24+, service account tokens are now created by Rancher if they do not exist (as the token-controller no longer automatically creates service account tokens). Before this PR, Rancher would open a watch for service accounts, relying on the `.secrets` field of the service account to contain the service account token (as this was populated by the token-controller). The secrets field is no longer populated, so opening a watch for service accounts is no longer a valid approach for waiting for service account token.
 
## Solution
Now, Rancher will watch for the service account, then once it finds a valid service account, open a watch for secrets that specifically correspond to the service account.
 
### Manual Testing
Create a v2prov custom cluster in v2.6.8, and observe that running the curl script takes 60+ seconds due to curl timeout after `Generating cattle id`.

After this change, there should be no delay.

### Automated Testing
The v2prov custom integration tests will test this code thoroughly.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,7 +4,7 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH}
 ENV CATTLE_HELM_VERSION v2.16.8-rancher2
 ENV CATTLE_MACHINE_VERSION v0.15.0-rancher95
-ENV CATTLE_K3S_VERSION v1.24.1+k3s1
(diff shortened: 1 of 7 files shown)


jakefhyde (approved):

jakefhyde (dismissed):

HarrisonWAffel (dismissed):

jakefhyde (approved):

snasovich:
@Oats87 , please add commit from https://github.com/rancher/rancher/pull/39098 here as well because it was done on the original 2.6.x issue as well.

Oats87:
@snasovich done, PR updated.

PR #39017, opened 2022-09-19
[release/v2.7] Allow admins to reconfigure existing deprecated Azure AD authconfig

maxsokolovsky:
## Problem
Rancher picks the new auth flow in cases where it should pick the old. Specifically, when admins have a deprecated Azure AD setup in Rancher v2.6.7+ and when they want to simply reconfigure their existing setup (by changing the Application Secret, for example). 
 
## Solution
On `testAndApply`, implement the annotation-setting logic so that it behaves in the following way:
Before setting or not setting an annotation on a new auth config, fetch an existing config from the database. Inspect this config.
1. If it is disabled, then use the new auth flow and thus set the annotation on the new config.
2. If it is enabled and has the annotation, also use the new auth flow and thus set the annotation on the new config.
3. If it is enabled and **does not** have the annotation, use the old auth flow and and thus **do not** set the annotation on the new config.
 
## Testing
1. On v2.6.6 Rancher server, enable Azure AD auth config with standard endpoints.
4. Upgrade the rancher server to v2.6-head.
5. Do not update the endpoints but click on edit config.
6. Input a new Application Secret value and save.

### DIFF
--- a/pkg/auth/providers/azure/azure_actions.go
+++ b/pkg/auth/providers/azure/azure_actions.go
@@ -11,13 +11,10 @@ import (
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-	"github.com/rancher/rancher/pkg/auth/providers/azure/clients"
 	"github.com/rancher/rancher/pkg/auth/providers/common"
-	"github.com/rancher/rancher/pkg/auth/tokens"
 	client "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	managementschema "github.com/rancher/rancher/pkg/schemas/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func (ap *azureProvider) formatter(apiContext *types.APIContext, resource *types.RawResource) {
@@ -73,30 +70,12 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 
 	azureADConfig := &azureADConfigApplyInput.Config
 
-	// Get the current config from the database and, if it is enabled and does not have the new auth flow annotation,
-	// do not set the annotation on the new config.
-	// This covers the case where admins had been using a deprecated Azure AD setup
-	// and must migrate to the new auth config setup after upgrading Rancher.
-	// But they may not want to migrate immediately, and could still want to reconfigure their existing, deprecated config by changing
-	// the Application Secret after rotating it, for example. This way, Rancher allows the old setup to be reconfigured.
 	currentConfig, err := ap.getAzureConfigK8s()
 	if err != nil {
 		logrus.Errorf("Failed to fetch Azure AD Config from Kubernetes: %v", err)
 		return httperror.NewAPIError(httperror.ServerError, "failed to fetch Azure AD Config from Kubernetes")
 	}
-
-	enabled := authProviderEnabled(currentConfig)
-	if !enabled || (enabled && configHasNewFlowAnnotation(currentConfig)) {
-		// This covers the case where admins upgrade Rancher to v2.6.7+ without having used Azure AD as the auth provider.
-		// In 2.6.7+, whether Azure AD is later registered or not, Rancher on startup creates the annotation on the template auth config.
-		// But in the case where the auth config had been created on Rancher startup prior to v2.6.7, the annotation would be missing.
-		// This ensures the annotation is set on initial attempt to set up Azure AD.
-		// This also covers the case where admins want to reconfigure a v2.6.7+ new auth flow setup with a new secret or app.
-		if azureADConfig.ObjectMeta.Annotations == nil {
-			azureADConfig.ObjectMeta.Annotations = make(map[string]string)
-		}
-		azureADConfig.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
-	}
+	migrateNewFlowAnnotation(currentConfig, azureADConfig)
 
 	azureLogin := &v32.AzureADLogin{
 		Code: azureADConfigApplyInput.Code,
@@ -134,79 +113,19 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
 	return ap.tokenMGR.CreateTokenAndSetCookie(user.Name, userPrincipal, groupPrincipals, providerToken, 0, "Token via Azure Configuration", request, userExtraInfo)
 }
 
-// migrateToMicrosoftGraph performs the migration of the registered Azure AD auth provider
-// from the deprecated Azure AD Graph API to the Microsoft Graph API.
-// It modifies the existing auth config value in the database, so that it has up-to-date endpoints to the new API.
-// Most importantly, it sets the annotation that specifies that the auth config has been migrated to use the new auth flow.
-
-// It also verifies that admins have properly configured an existing app registration's permissions
-// in the Azure portal before they update their Azure AD auth config to use the new authentication flow.
-// The method receives the current AuthConfig from the database, then updates it in-memory to use the new endpoints,
-// and creates a new test Azure client, thereby getting an access token to the Graph API.
-// Then it parses the JWT and inspects the permissions contained within. If the admins had not set those as per docs,
-// then Rancher won't find the permissions in the test token and will return an error.
-
-// If validation and applying work, then migrateToMicrosoftGraph deletes all secrets with access tokens to the
-// deprecated Azure AD Graph API.
-func (ap *azureProvider) migrateToMicrosoftGraph() error {
-	cfg, err := ap.updateConfigAndValidatePermissions()
-	if err != nil {
-		return err
-	}
-	if err = ap.applyUpdatedConfig(cfg); err != nil {
-		return err
-	}
-	ap.deleteUserAccessTokens()
-	clients.GroupCache.Purge()
-	return nil
-}
-
-func (ap *azureProvider) updateConfigAndValidatePermissions() (*v32.AzureADConfig, error) {
-	cfg, err := ap.getAzureConfigK8s()
-	if err != nil {
-		return nil, err
-	}
-	if !authProviderEnabled(cfg) {
-		return nil, httperror.NewAPIError(httperror.InvalidState, "the Azure AD auth provider is not enabled")
-	}
-
-	updateAzureADEndpoints(cfg)
-
-	// Try to get a new client, which will fetch a new access token and validate its permissions.
-	_, err = clients.NewMSGraphClient(cfg, ap.secrets)
-	if err != nil {
-		return nil, err
-	}
-	return cfg, nil
-}
-
-func (ap *azureProvider) applyUpdatedConfig(cfg *v32.AzureADConfig) error {
-	if cfg.ObjectMeta.Annotations == nil {
-		cfg.ObjectMeta.Annotations = make(map[string]string)
-	}
-	cfg.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
-	return ap.saveAzureConfigK8s(cfg)
-}
-
-// deleteUserAccessTokens attempts to delete all secrets that contain users' access tokens used for working with
-// the deprecated Azure AD Graph API.
-// It is not possible to filter secrets easily by presence of specific key(s) in the data object. The method fetches all
-// Opaque secrets in the relevant namespace and looks at the target key in the data to find a secret that stores a user's
-// access token to delete.
-func (ap *azureProvider) deleteUserAccessTokens() {
-	secrets, err := ap.secrets.ListNamespaced(tokens.SecretNamespace, metav1.ListOptions{FieldSelector: "type=Opaque"})
-	if err != nil {
-		logrus.Errorf("failed to fetch secrets: %v", err)
+// Check the current auth config and make sure that the proposed one submitted through the API has up-to-date annotations.
+// Rancher relies on GraphEndpointMigratedAnnotation to choose the right authentication flow and Graph API.
+func migrateNewFlowAnnotation(current, proposed *v32.AzureADConfig) {
+	if isConfigDeprecated(current) {
 		return
 	}
-	// Provider name for Azure AD is the main key on secret data. This allows to identify the secrets to be deleted.
-	const key = Name
-	for _, secret := range secrets.Items {
-		if _, keyPresent := secret.Data[key]; keyPresent {
-			err := ap.secrets.DeleteNamespaced(tokens.SecretNamespace, secret.Name, &metav1.DeleteOptions{})
-			if err != nil {
-				logrus.Errorf("failed to delete secret %s:%s - %v", tokens.SecretNamespace, secret.Name, err)
-			}
-		}
+	// This covers the case where admins upgrade Rancher to v2.6.7+ without having used Azure AD as the auth provider.
+	// In 2.6.7+, whether Azure AD is later registered or not, Rancher on startup creates the annotation on the template auth config.
+	// But in the case where the auth config had been created on Rancher startup prior to v2.6.7, the annotation would be missing.
+	// This ensures the annotation is set on initial attempt to set up Azure AD.
+	// This also covers the case where admins want to reconfigure a v2.6.7+ new auth flow setup with a new secret or app.
+	if proposed.ObjectMeta.Annotations == nil {
+		proposed.ObjectMeta.Annotations = make(map[string]string)
 	}
+	proposed.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
 }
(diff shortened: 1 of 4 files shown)


pkg/auth/providers/azure/azure_actions.go
```diff
@@ -73,14 +73,29 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
-	azureADConfig.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] = "true"
+	// Get the current config from the database and, if it is enabled and does not have the new auth flow annotation,
+	// do not set the annotation on the new config.
+	// This covers the case where admins had been using a deprecated Azure AD setup
+	// and must migrate to the new auth config setup after upgrading Rancher.
+	// But they may not want to migrate immediately, and could still want to reconfigure their existing, deprecated config by changing
+	// the Application Secret after rotating it, for example. This way, Rancher allows the old setup to be reconfigured.
+	currentConfig, err := ap.getAzureConfigK8s()
+	if err != nil {
+		return err
```
MbolotSuse, line 84: I think you want to wrap this in an `httperror` so that we have more control over the response to the user.

MbolotSuse, line 84: I think you want to wrap this in an `httperror` so that we have more control over the response to the user.

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -73,14 +73,29 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
+	}
+
+	enabled := authProviderEnabled(currentConfig)
+	if !enabled || (enabled && configHasNewFlowAnnotation(currentConfig)) {
```
MbolotSuse, line 88: The feels excessively complex to me. In the second case, the config is enabled, and it has the annotation. But we are still doing the annotation set, despite the fact that we know this has already been set. 

In short - second case is not needed, this could be:

```go
if !enabled{
```

MbolotSuse (changes requested):
Couple of items on structure/error handling.

Also needs tests, but I assume you were waiting for a review first.

pkg/auth/providers/azure/azure_actions.go
MbolotSuse: Ah - ok. I missed the `currentConfig` part of the statement, my bad.

pkg/auth/providers/azure/azure_actions.go
maxsokolovsky: Errors returned directly from the action (except for login errors which are caught and handled differently) are just returned to the UI with a 500 response code. I tested this particular case, and the user would just see the error's string that is returned here.

MbolotSuse: Thank you for testing that out. I don't think we want user to see the error string returned here. We would want a generic error message and the true error message to be logged server side. 

In short, I hold to what I said above - we should wrap this with a generic error message and a specific response code.

maxsokolovsky: Agreed, but, ideally, I will need to change all the other places where this method is called. Some of them are "user-facing" too.

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -73,14 +73,30 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
+	// Get the current config from the database and, if it is enabled and does not have the new auth flow annotation,
+	// do not set the annotation on the new config.
+	// This covers the case where admins had been using a deprecated Azure AD setup
+	// and must migrate to the new auth config setup after upgrading Rancher.
+	// But they may not want to migrate immediately, and could still want to reconfigure their existing, deprecated config by changing
+	// the Application Secret after rotating it, for example. This way, Rancher allows the old setup to be reconfigured.
+	currentConfig, err := ap.getAzureConfigK8s()
+	if err != nil {
+		logrus.Errorf("Failed to fetch Azure AD Config from the database: %v", err)
+		return httperror.NewAPIError(httperror.ServerError, "failed to fetch Azure AD Config from the database")
```
MbolotSuse, line 85: Nit:

I'm a bit conflicted about this error message. On one hand, it is unique and doesn't expose an arbitrary string, which is good. On the other hand, it gives the user confusing information that wont' help them resolve the issue. 

At minimum, I think it would be good to drop "from the database", since rancher uses k8s as it's database, and this may confuse users into thinking we are running postgresql or something behind the scenes.

MbolotSuse (changes requested):
One nitpick on error message. Looks ok outside of that.

MbolotSuse (commented):
Looks good outside of the tests, LMK when you decide if we are doing that in a separate PR.

maxsokolovsky:
The existing tests should cover this one.

pkg/auth/providers/azure/migrate.go
```diff
@@ -0,0 +1,87 @@
+package azure
```
MbolotSuse, line 1: Speaking generally on this new file: I am personally not a fan of when methods defined for structs are split between several files. It makes it hard to track which methods are for which structs.

But that's just me.

pkg/auth/providers/azure/config.go
```diff
@@ -32,11 +32,12 @@ func isConfigDeprecated(cfg *v32.AzureADConfig) bool {
 	if !cfg.Enabled {
```
MbolotSuse, line 32: I'm starting to dislike the fact that this function checks if the config is enabled or not. Right now we are checking if the config is enabled here for one part of the logic and in the main body of logic (in the provider) in other parts. It creates this weird split where multiple areas of the code are relying on the same piece of logic in different ways. 

I would prefer that we only handle the "Are you enabled or not" check in one place and have the various pieces of logic which rely on it call that one place, rather than what we are doing right now.

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -118,79 +113,23 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
-	}
-	// Provider name for Azure AD is the main key on secret data. This allows to identify the secrets to be deleted.
-	const key = Name
-	for _, secret := range secrets.Items {
-		if _, keyPresent := secret.Data[key]; keyPresent {
-			err := ap.secrets.DeleteNamespaced(tokens.SecretNamespace, secret.Name, &metav1.DeleteOptions{})
-			if err != nil {
-				logrus.Errorf("failed to delete secret %s:%s - %v", tokens.SecretNamespace, secret.Name, err)
-			}
+// Get the current config from the database and, if it is enabled and does not have the new auth flow annotation,
```
MbolotSuse, line 116: Nit:

I would recommend cutting down on the amount of comments here. I would try to convey the purpose of this function in 1-2 lines at most. Right now we are at over 10 - I think people will have a hard time understanding.

Here's a potential replacement:
```go
// Check the current auth config and make sure that the new one proposed through the api has up-to-date annotations
// We rely on GraphEndpointMigratedAnnotation to choose the right auth provider, but we don't want users to be able to request an old auth config
```

pkg/auth/providers/azure/azure_provider_test.go
```diff
@@ -225,3 +228,73 @@ func TestTransformToAuthProvider(t *testing.T) {
...
+func TestProcessConfigOnTestAndApply(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name               string
+		current            *v3.AzureADConfig
+		new                *v3.AzureADConfig
+		annotationExpected bool
+	}{
+		{
```
MbolotSuse, line 241: I think you are missing a test case for if the current config is enabled and has the new flow annotation (in this case we don't want to set the annotation). 

This is also leaving out cases where the new config has an annotation and we make sure that we overrode it and didn't just use what was already there. I would recommend at least one test case where new has an annotation for this value and we forcibly override it to say that it is enabled.

pkg/auth/providers/azure/azure_provider_test.go
```diff
@@ -225,3 +228,73 @@ func TestTransformToAuthProvider(t *testing.T) {
 		})
 	}
 }
+
+func TestProcessConfigOnTestAndApply(t *testing.T) {
```
MbolotSuse, line 232: So this does test your new function - however, test and apply (which is large portion of logic), remains untested. Given that this is a public facing api endpoint, that's probably not what we want long-term. I realize this may be tough, so I'm willing to let it pass if we need to, but I wanted to make sure that it was brought up.

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -118,79 +113,23 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
+// do not set the annotation on the new config.
+// This covers the case where admins had been using a deprecated Azure AD setup
+// and must migrate to the new auth config setup after upgrading Rancher.
+// But they may not want to migrate immediately, and could still want to reconfigure their existing, deprecated config by changing
+// the Application Secret after rotating it, for example. This way, Rancher allows the old setup to be reconfigured.
+func processAuthConfigOnSetup(current, new *v32.AzureADConfig) {
```
MbolotSuse, line 122: Nit: 

Function could have a better name. I think you aren't really processing the config. More you are checking annotations. Maybe something like `migrateNewFlowAnnotation(current, proposed)` might work better?

MbolotSuse (changes requested):
Couple of minor nitpicks on function name and comments.

There's also a test case that I think would be good to add.

pkg/auth/providers/azure/azure_provider_test.go
maxsokolovsky: That one I do have, but I noticed that am missing a case when Azure AD is set up for the first time, and the current one has the annotation, meaning it had been set on Rancher startup.

pkg/auth/providers/azure/config.go
maxsokolovsky: I completely agree, I just did not want to significantly refactor the code, as this might require changes to the structure of package imports. I will try to see if this can be done though.

pkg/auth/providers/azure/migrate.go
maxsokolovsky: I think it's fine, because the migration code is really separate, so it deserves its own file. And it's an important distinct operation/feature of the auth provider, so it's OK to designate it with a file. And the provider already has a lot of methods anyways. The actions file has some methods for it...

pkg/auth/providers/azure/azure_actions.go
maxsokolovsky: Yes, agreed.

pkg/auth/providers/azure/config.go
maxsokolovsky: Ah, never mind, I misread the highlighted function, it's all easier, will adjust.

pkg/auth/providers/azure/azure_provider_test.go
maxsokolovsky: Also, I don't think we should rely on whether the proposed one has the annotation, right? The backend now figures out on its own if the new config must have it or not.

pkg/auth/providers/azure/migrate.go
MbolotSuse: Not sure that I agree that further splitting the code is a good idea.

That being said, this is largely a style item, so I'm going to resolve the discussion.

pkg/auth/providers/azure/azure_provider_test.go
MbolotSuse: > That one I do have, but I noticed that am missing a case when Azure AD is set up for the first time, and the current one has the annotation, meaning it had been set on Rancher startup.

Near as I can tell there are 4 test cases here:
- Current is enabled, and has the annotation set to the right value (set the annotation on proposed)
- Current is enabled, but does not have the annotation set (don't set the annotation on proposed)
- Current is not enabled, but has the new flow annotation (set the annotation on proposed)
- Current is not enabled, and does not have the new flow annotation (set the annotation on proposed)

That's not counting any "part of input is nil" cases. As long as we have those 4, we should be good. I'll give it another look.

> Also, I don't think we should rely on whether the proposed one has the annotation, right? The backend now figures out on its own if the new config must have it or not.

My point here is that you are overwriting what is present on the proposed config. It's possible to change the code to leave what's on the proposed config if the user supplied this annotation (for example if the user changed this to use the old side). If this happened, all of your test cases would still pass, but we wouldn't want that to happen.

pkg/auth/providers/azure/config.go
MbolotSuse: Thanks for looking at this, but the new code isn't really what I was thinking of. 

My issue is that we have this `isConfigDeprecated` function. The job of this function is to determine if we should use the old side client or the new side client. And we use this in other places in the code as the decision maker for this process.

However, with the new code that we have, we now have a separate way of determining which client to use - which is decided by the `migrateNewFlowAnnotation` function. This function makes the decision and then manipulates the proposed config so that `isConfigDeprecated` will work as expected later on. This seems dangerous to me - rather than having one point of failure that determines if we should use old or new config we now have two that work together. In addition, there's no indication that they work together, so this will be trick to debug.

pkg/auth/providers/azure/migrate.go
KevinJoiner: Nit: 
I am also not a fan of splitting up methods and structs. I feel like it hurts readability ,because without an Iquicklyre is no way to easily find the struct definition or other methods the struct is using. In general when I feel the need to create separate files it usually means my struct is getting heavy and can probably be broken up. You don't have to change it for this review, but I think it's worth noting.

pkg/auth/providers/azure/azure_provider_test.go
```diff
@@ -225,3 +228,84 @@ func TestTransformToAuthProvider(t *testing.T) {
...
+					},
+				},
+				GraphEndpoint: "https://graph.microsoft.com",
+			},
+			proposed:           &v3.AzureADConfig{},
+			annotationExpected: true,
+		},
+	}
+
+	for _, test := range tests {
```
KevinJoiner, line 297: If we loop by value and call `Parallel()` test variable will be overridden.
```suggestion
	for i := range tests {
	    test := tests[i]
```

pkg/auth/providers/azure/azure_actions.go
```diff
@@ -118,79 +113,18 @@ func (ap *azureProvider) testAndApply(actionName string, action *types.Action, r
...
-	for _, secret := range secrets.Items {
-		if _, keyPresent := secret.Data[key]; keyPresent {
-			err := ap.secrets.DeleteNamespaced(tokens.SecretNamespace, secret.Name, &metav1.DeleteOptions{})
-			if err != nil {
-				logrus.Errorf("failed to delete secret %s:%s - %v", tokens.SecretNamespace, secret.Name, err)
-			}
+// Check the current auth config and make sure that the proposed one submitted through the API has up-to-date annotations.
+// Rancher relies on GraphEndpointMigratedAnnotation to choose the right authentication flow and Graph API.
+func migrateNewFlowAnnotation(current, proposed *v32.AzureADConfig) {
+	if !isConfigDeprecated(current) {
```
KevinJoiner, line 119: nit:
https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow
```suggestion
	if isConfigDeprecated(current) {
	   return
	}
```

pkg/auth/providers/azure/migrate.go
KevinJoiner: I unresolved just so you can see my comment, but like I said you can leave it how you have it now.

KevinJoiner (changes requested):

pkg/auth/providers/azure/config.go
```diff
@@ -29,14 +29,15 @@ func authProviderEnabled(config *v32.AzureADConfig) bool {
...
-	if !ok || v != "true" {
-		logrus.Tracef("Could not find the %s annotation that specifies whether the Graph Endpoint has been migrated, or its value is not \"true\" - Rancher will use the old endpoint.",
-			GraphEndpointMigratedAnnotation)
-		return true
+	return !configHasNewFlowAnnotation(cfg)
+}
+
+func configHasNewFlowAnnotation(cfg *v32.AzureADConfig) bool {
+	if cfg.ObjectMeta.Annotations != nil {
+		return cfg.ObjectMeta.Annotations[GraphEndpointMigratedAnnotation] == "true"
```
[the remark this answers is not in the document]
maxsokolovsky: Yes, ObjectMeta is always set, I had also verified this.

pkg/auth/providers/azure/config.go
```diff
@@ -29,14 +29,15 @@ func authProviderEnabled(config *v32.AzureADConfig) bool {
 }
 
 func isConfigDeprecated(cfg *v32.AzureADConfig) bool {
-	if !cfg.Enabled {
+	if !authProviderEnabled(cfg) {
```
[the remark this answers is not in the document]
maxsokolovsky: Good point!

pkg/auth/providers/azure/migrate.go
maxsokolovsky: I'd leave it, although I completely agree with you and Michael here. I just think that here this does not hurt, because of how distinct the migrate action is. As I worked, I often searched for functions and methods of this migration process and figured that it does warrant its own file in the end.

KevinJoiner (approved):
LGTM.

MbolotSuse (approved):
LGTM

PR #39018, opened 2022-09-19
[2.6 backport] Allow admins to reconfigure existing deprecated Azure AD authconfig

maxsokolovsky:
## Problem
Rancher picks the new auth flow in cases where it should pick the old. Specifically, when admins have a deprecated Azure AD setup in Rancher v2.6.7+ and when they want to simply reconfigure their existing setup (by changing the Application Secret, for example). 
 
## Solution
On `testAndApply`, implement the annotation-setting logic so that it behaves in the following way:
Before setting or not setting an annotation on a new auth config, fetch an existing config from the database. Inspect this config.
1. If it is disabled, then use the new auth flow and thus set the annotation on the new config.
2. If it is enabled and has the annotation, also use the new auth flow and thus set the annotation on the new config.
3. If it is enabled and **does not** have the annotation, use the old auth flow and and thus **do not** set the annotation on the new config.
 
## Testing
1. On v2.6.6 Rancher server, enable Azure AD auth config with standard endpoints.
4. Upgrade the rancher server to v2.6-head.
5. Do not update the endpoints but click on edit config.
6. Input a new Application Secret value and save.

### DIFF
--- a/pkg/auth/providers/azure/azure_actions.go
+++ b/pkg/auth/providers/azure/azure_actions.go
@@ -11,13 +11,10 @@ import (
 	"github.com/rancher/norman/httperror"
 	"github.com/rancher/norman/types"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-	"github.com/rancher/rancher/pkg/auth/providers/azure/clients"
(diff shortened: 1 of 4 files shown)


MbolotSuse (approved):
LGTM - looks identical to the 2.7 code.

PR #39019, opened 2022-09-19
Add role_based_access_control_enabled to AKS

markusewalker:
## Problem

With previous changes done to the `aks.tf` file, RBAC was not defined to be set to true. As per the AKS documentation, we should ensure that it is enabled to follow best practices.
 
## Solution

Added `role_based_access_control_enabled` parameter to ensure that RBAC is set to true.
 
### Manual Testing

Locally tested that a Rancher server is stood up with AKS as the local cluster. Once done, provisioned an RKE1 cluster and validated that came up successfully.


izaac (approved):
LGTM

sowmyav27 (approved):

Issue #38974, opened 2022-09-16
Implement management.cattle.io schema just Go Automation Framework 


igomez06:
**Is your feature request related to a problem? Please describe.**
Because of a change made to the native eks struct, pr: https://github.com/rancher/eks-operator/pull/59, it is no longer possible to have empty values for tags, loggingTypes, subnets, etc. because of the behavior of golang json marshaling. 

**Describe the solution you'd like**
The solution is for the go automation framework to have it's own schema, where we can change those fields back to pointers.

igomez06:
PR v2.7: https://github.com/rancher/rancher/pull/39034

PR v2.6: https://github.com/rancher/rancher/pull/39035

PR #39035, opened 2022-09-20
[release/v2.6] Implemented our own management.cattle.io schema in the framework

igomez06:
## Issue: 

https://github.com/rancher/rancher/issues/38974
 
## Problem

 Because of a change made to the native eks struct, pr: https://github.com/rancher/eks-operator/pull/59, it is no longer possible to have empty values for tags, loggingTypes, subnets, etc. because of the behavior of golang json marshaling. This is to revert this behavior to be able to have empty arrays and maps in the rest call, since it is allowed by rancher. This also to lay groundwork to update the clients, since some of these api calls are moving to the v1 endpoint.
## Solution
 Created our own schema to be used to generate the management.cattle.io clients in the framework.
 
## Testing
 Ran hosted provider provisioning tests in the Jenkins


izaac (approved):
LGTM

slickwarren (approved):

PR #39044, opened 2022-09-20
Increasing the test watch timeout due to timeouts

mattfarina:
The tests are sometimes timing out which is causing flaky tests. This change increases the timeout period to give the tests longer to complete.

## Problem

The k3s provisioning tests appear to be timing out.
 
## Solution

Increase the timeout.
 
## Testing

This is in the tests.


mattfarina:
Nope, this doesn't fix it as this isn't a timeout being hit. Something else is occurring.

PR #39049, opened 2022-09-20
Update parallelism to reduce server load

mattfarina:
Given the instance sizes we are running the reduced parallelism can help CI have a reduced load at any one time.

## Problem

CI is having issues with timeouts, OOM, or something else in the k3s provisioning tests.

## Solution

 
Reducing the parallelism to 50% of the vCPUs.


mattfarina:
This did not fix the issue. Closing.

PR #39059, opened 2022-09-21
[2.7] Fix registries within the cluster agent manifest for Prov V2

HarrisonWAffel:
## Problem
When deploying RKE2/K3s clusters which have a cluster level private registry configured, the cluster agent manifest did not incorporate that registry. Instead, the global system-default-registry was used. This resulted in images created by the cluster agent (namely the `shell` image when creating the `dashboard-shell` container) being pulled from the incorrect registry, which has implications for clusters which only have access to the cluster level registry. This issue is apart of a larger set of issues regarding the use of cluster level private registries and their precedence over the global system-default-registry. 

This issue only occurs with RKE2/K3s clusters.

## Root Cause
The location of the private registry details differs between the `v3.Cluster` objects for RKE1 and RKE2/K3s. 

This results in the inability to determine if an RKE2/K3s cluster has a cluster level registry configured when generating the cluster agent manifest during the call to `SystemTemplate` within `pkg/systemtemplate/import.go`. 

Due to this, the `settings.SystemDefaultRegistry` was always being applied, causing images to be pulled incorrectly. 
 
## Solution
When creating the `v3.Cluster` object for RKE2/K3s clusters, add a reference to the private registry URL and the secret containing the system-default-registry credentials. The secret is created by the UI, and the registry URL can be found on the `v1.Cluster` object. Then, when generating cluster agent manifests, utilize these references to ensure that cluster level registries for RKE2/K3s are honored. 

These references are held in the `ClusterSecrets` section of the object so as to avoid spec changes as much as possible. A new field has been added to track the URL of the cluster default registry. With these new references, `pkg/controllers/management/clusterdeploy/clusterdeploy.go` and `pkg/systemtemplate/import.go` can determine the cluster level private registry for both RKE1 and RKE2/K3s. 

if there is a better place to put these references, other than the `ClusterSecrets`, let me know. My main intent was to modify the object as little as possible to include these new references. 

 
## Testing
+ Configure two private registries, one with no authentication and another with authentication
   + non-authenticated registry will serve as the global registry 
   + authenticated registry will serve as the cluster level private registry 
+ In the global settings, set the `system-default-registry` to the URL of your unauthenticated registry
+ Begin to create an RKE2/K3s cluster
+ Before creating the RKE2/K3s cluster, go to the `Registries` section of the Cluster configuration screen and supply the authenticated registries URL, username, and password.
+ Create the cluster and wait for it to become available
+ Navigate to the cluster workload page, and select all namespaces 
+ Select the `pods` view to see all running pods 
+ In the top right corner, press the `Kubectl Shell` button to create a new `dashboard-shell` container 
+ Look for the new contain in the `pods` view and see that the image is pulled from the cluster level private registry.

### Manual Testing
I have manually tested the following scenarios, all of which have passed. 

+ RKE1 cluster provisioning with only a global system-default-registry configured
  + all images are pulled from the correct repository, including the `shell` image for the `dashboard-shell` container
+ RKE1 cluster provisioning with both a global system-default-registry and cluster level private registry configured:
  + all images are pulled from the correct repository, including the `shell` image for the `dashboard-shell` container
+ RKE2 & K3s cluster provisioning with only a global system-default-registry configured
  + all images are pulled from the correct repository, including the `shell` image for the `dashboard-shell` container
+ RKE2 & K3s cluster provisioning with both a global system-default-registry and cluster level private registry configured:
  + all images are pulled from the correct repository, including the `shell` image for the `dashboard-shell` container
  + The cluster agent manifest has the correct environment variables and the associated registry secret has the correct configuration. 
+ Scaling RKE1 node pools up and down
+ Scaling RKE2 & K3s node pools up and down 

### Automated Testing
No automated testing was added 

## QA Testing Considerations
The private registries must be loaded with the correct images for the k8's version that you want to use. My testing was done against registries loaded with 2.6.7 images.
 
RKE2/K3s clusters are treated as imported clusters, so additional testing around importing existing clusters should be done to ensure no regressions have been introduced. I've done brief testing of imported clusters with these changes and did not notice any issues.

### DIFF
--- a/pkg/cluster/private_registry.go
+++ b/pkg/cluster/private_registry.go
@@ -34,6 +34,22 @@ func GetPrivateRepo(cluster *v3.Cluster) *rketypes.PrivateRegistry {
 	return nil
 }
 
+// TransformProvV2RegistryCredentialsToDockerConfigJSON transforms a ProvV2 registry secret into a credentialprovider.DockerConfigJSON secret
+func TransformProvV2RegistryCredentialsToDockerConfigJSON(url string, registryCredentials *corev1.Secret) (map[string][]byte, error) {
+	username := string(registryCredentials.Data["username"])
+	password := string(registryCredentials.Data["password"])
+	authConfig := credentialprovider.DockerConfigJSON{
+		Auths: credentialprovider.DockerConfig{
+			url: credentialprovider.DockerConfigEntry{
+				Username: username,
+				Password: password,
+			},
+		},
+	}
+	j, err := json.Marshal(authConfig)
+	return map[string][]byte{".dockerconfigjson": j}, err
+}
+
 func GenerateClusterPrivateRegistryDockerConfig(cluster *v3.Cluster) (string, error) {
 	if cluster == nil {
 		return "", nil
@@ -41,8 +57,8 @@ func GenerateClusterPrivateRegistryDockerConfig(cluster *v3.Cluster) (string, er
 	return GeneratePrivateRegistryDockerConfig(GetPrivateRepo(cluster), nil)
 }
 
-// This method generates base64 encoded credentials for the registry
-func GeneratePrivateRegistryDockerConfig(privateRegistry *rketypes.PrivateRegistry, registrySecret *corev1.Secret) (string, error) {
+// GeneratePrivateRegistryDockerConfig generates base64 encoded credentials for the provided registry
+func GeneratePrivateRegistryDockerConfig(privateRegistry *rketypes.PrivateRegistry, registrySecret []byte) (string, error) {
 	if privateRegistry == nil {
 		return "", nil
 	}
@@ -62,17 +78,15 @@ func GeneratePrivateRegistryDockerConfig(privateRegistry *rketypes.PrivateRegist
 	if registrySecret != nil {
 		privateRegistry = privateRegistry.DeepCopy()
 		dockerCfg := credentialprovider.DockerConfigJSON{}
-		if dockerConfigJSON := registrySecret.Data[".dockerconfigjson"]; len(dockerConfigJSON) > 0 {
-			err := json.Unmarshal(dockerConfigJSON, &dockerCfg)
+		if len(registrySecret) > 0 {
+			err := json.Unmarshal(registrySecret, &dockerCfg) // check to see if registrySecret is in the correct format
 			if err != nil {
 				logrus.Debug("Failed to parse dockerconfig for registry secret: " + err.Error())
 				return "", err
 			}
-		}
-
-		if reg, ok := dockerCfg.Auths[privateRegistry.URL]; ok {
-			privateRegistry.User = reg.Username
-			privateRegistry.Password = reg.Password
+			if _, ok := dockerCfg.Auths[privateRegistry.URL]; ok {
+				return base64.URLEncoding.EncodeToString(registrySecret), nil
+			}
 		}
 	}
 	if privateRegistry.User == "" || privateRegistry.Password == "" {
--- a/pkg/controllers/management/clusterdeploy/clusterdeploy.go
+++ b/pkg/controllers/management/clusterdeploy/clusterdeploy.go
@@ -17,6 +17,7 @@ import (
 	util "github.com/rancher/rancher/pkg/cluster"
 	"github.com/rancher/rancher/pkg/clustermanager"
 	"github.com/rancher/rancher/pkg/features"
+	"github.com/rancher/rancher/pkg/fleet"
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/image"
@@ -289,12 +290,9 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
 	}
 	logrus.Tracef("clusterDeploy: deployAgent: desiredTaints is [%v] for cluster [%s]", desiredTaints, cluster.Name)
 
-	var privateRegistries *corev1.Secret
-	if cluster.GetSecret("PrivateRegistrySecret") != "" {
-		privateRegistries, err = cd.secretLister.Get(namespace.GlobalNamespace, cluster.GetSecret("PrivateRegistrySecret"))
-		if err != nil {
-			return err
-		}
+	privateRegistrySecretData, err := cd.getPrivateRegistrySecretData(cluster)
+	if err != nil {
+		return err
 	}
 
 	if !redeployAgent(cluster, desiredAgent, desiredAuth, desiredFeatures, desiredTaints) {
@@ -312,7 +310,7 @@ func (cd *clusterDeploy) deployAgent(cluster *v3.Cluster) error {
 	}()
 
 	if _, err = v32.ClusterConditionAgentDeployed.Do(cluster, func() (runtime.Object, error) {
-		yaml, err := cd.getYAML(cluster, desiredAgent, desiredAuth, desiredFeatures, desiredTaints, privateRegistries)
+		yaml, err := cd.getYAML(cluster, desiredAgent, desiredAuth, desiredFeatures, desiredTaints, privateRegistrySecretData)
 		if err != nil {
 			return cluster, err
 		}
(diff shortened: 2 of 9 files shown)


pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -520,6 +519,41 @@ func (cd *clusterDeploy) cacheAgentImages(name string) error {
...
+	if registrySecret == "" { // no private registry configured
+		return nil, nil
+	}
+
+	// check for the RKE1 registry secret first
+	privateRegistries, err = cd.secretLister.Get(namespace.GlobalNamespace, cluster.GetSecret("PrivateRegistrySecret"))
+	if err == nil {
+		return privateRegistries, nil
+	}
+	if err != nil && !strings.Contains(err.Error(), "not found") { // ignore secret not found errors as we need to check ProvV2 namespace as well
```
jakefhyde, line 536:
```suggestion
	if err != nil && !apierrors.IsNotFound { // ignore secret not found errors as we need to check ProvV2 namespace as well
```

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -520,6 +519,41 @@ func (cd *clusterDeploy) cacheAgentImages(name string) error {
...
+	var privateRegistries *corev1.Secret
+	var err error
+
+	registrySecret := cluster.GetSecret("PrivateRegistrySecret")
+	if registrySecret == "" { // no private registry configured
+		return nil, nil
+	}
+
+	// check for the RKE1 registry secret first
+	privateRegistries, err = cd.secretLister.Get(namespace.GlobalNamespace, cluster.GetSecret("PrivateRegistrySecret"))
```
jakefhyde, line 532:
```suggestion
	privateRegistries, err = cd.secretLister.Get(namespace.GlobalNamespace, registrySecret)
```

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -520,6 +519,41 @@ func (cd *clusterDeploy) cacheAgentImages(name string) error {
...
+	}
+
+	privateRegistries, err = cd.secretLister.Get(fleet.ClustersDefaultNamespace, cluster.GetSecret("PrivateRegistrySecret"))
+	if err != nil {
+		return nil, err
+	}
+
+	// ProvV2 PrivateRegistrySecret is not in proper .dockerconfigjson format as-is, so we have to transform it into that format
+	// so that the SystemTemplate generates the cluster agent manifest properly.
+	return util.TransformProvV2RegistryCredentialsToDockerConfigJSON(cluster.GetSecret("PrivateRegistryURL"), privateRegistries)
```
jakefhyde, line 554:
```suggestion
	return util.TransformProvV2RegistryCredentialsToDockerConfigJSON(registryURL, privateRegistries)
```

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -520,6 +519,41 @@ func (cd *clusterDeploy) cacheAgentImages(name string) error {
...
+	}
+
+	// This is the only place that ProvV2 stores the system-default-registry URL on the v3.Cluster object.
+	// Without it, we cannot generate the registry credentials (.dockerconfigjson)
+	registryURL := cluster.GetSecret("PrivateRegistryURL")
+	if registryURL == "" {
+		return nil, nil
+	}
+
+	privateRegistries, err = cd.secretLister.Get(fleet.ClustersDefaultNamespace, cluster.GetSecret("PrivateRegistrySecret"))
```
jakefhyde, line 547:
```suggestion
	privateRegistries, err = cd.secretLister.Get(fleet.ClustersDefaultNamespace, registrySecret)
```

jakefhyde (changes requested):
Few minor nits

pkg/controllers/management/clusterdeploy/clusterdeploy.go
HarrisonWAffel: github needs a face-palm emoji

snasovich:
@HarrisonWAffel , please rebase the PR as there are conflicts now.

jakefhyde (changes requested):

pkg/cluster/private_registry.go
```diff
@@ -34,14 +34,32 @@ func GetPrivateRepo(cluster *v3.Cluster) *rketypes.PrivateRegistry {
...
+	authConfig := credentialprovider.DockerConfigJSON{
+		Auths: credentialprovider.DockerConfig{
+			url: credentialprovider.DockerConfigEntry{
+				Username: username,
+				Password: password,
+			},
+		},
+	}
+	j, err := json.Marshal(authConfig)
+	return &corev1.Secret{
```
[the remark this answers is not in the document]
jakefhyde: Either that or actually make the secret a secret.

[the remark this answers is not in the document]
HarrisonWAffel: I had thought about creating a new secret, but felt that it would be duplicating information that we already have in other secrets which is why I opted not to, and instead created one in memory. I am also worried about potential regressions within RKE1, since the two share similar code paths when deploying the agent, and so I tried to keep refactoring of existing behavior to a minimum. I can make this change, but private registry behavior for RKE1 will have to be revalidated as well.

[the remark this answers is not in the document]
HarrisonWAffel: Okay, I've changed things up to just use a `map[string][]byte` instead of the actual secret. I've done a quick smoke test and things seem to be working as expected, let me know if there are other changes related to the refactoring

pkg/cluster/private_registry.go
```diff
@@ -34,15 +34,31 @@ func GetPrivateRepo(cluster *v3.Cluster) *rketypes.PrivateRegistry {
...
 	if cluster == nil {
 		return "", nil
 	}
 	return GeneratePrivateRegistryDockerConfig(GetPrivateRepo(cluster), nil)
 }
 
-// This method generates base64 encoded credentials for the registry
-func GeneratePrivateRegistryDockerConfig(privateRegistry *rketypes.PrivateRegistry, registrySecret *corev1.Secret) (string, error) {
+// GeneratePrivateRegistryDockerConfig generates base64 encoded credentials for the provided registry
+func GeneratePrivateRegistryDockerConfig(privateRegistry *rketypes.PrivateRegistry, registrySecret map[string][]byte) (string, error) {
```
jakefhyde, line 61: If `".dockerconfigjson"` is the only key accessed, you could make the following change `registrySecret map[string][]byte -> dockerConfigJSON string`, this is really a nit though since the original ask was addressed.

pkg/cluster/private_registry.go
```diff
@@ -34,15 +34,31 @@ func GetPrivateRepo(cluster *v3.Cluster) *rketypes.PrivateRegistry {
 	return nil
 }
 
+// TransformProvV2RegistryCredentialsToDockerConfigJSON transforms a ProvV2 registry secret into a credentialprovider.DockerConfigJSON secret
+func TransformProvV2RegistryCredentialsToDockerConfigJSON(url string, registryCredentials *corev1.Secret) (map[string][]byte, error) {
+	username := string(registryCredentials.Data["username"])
+	password := string(registryCredentials.Data["password"])
+	authConfig := credentialprovider.DockerConfigJSON{
```
jakefhyde, line 41: Sorry I didn't catch this first time, but it looks like we create a `credentialprovider.DockerConfigJSON` that we immediately render to a json blob, then later we unmarshal the json, only to create an identical `credentialprovider.DockerConfigJSON` here: https://github.com/HarrisonWAffel/rancher/blob/a8c181e72bd8a7e37caa22ecc34392633115bab0/pkg/cluster/private_registry.go#L97-L105, that we then marshal to more identical json. Seems like alot of back and forth we probably could avoid`

jakefhyde (changes requested):

pkg/cluster/private_registry.go
HarrisonWAffel: Addressed, though instead of a string i've used a byte slice as I would just have to convert the string back into a byte slice to unmarshal the json

jakefhyde (dismissed):

pkg/controllers/management/clusterdeploy/clusterdeploy.go
```diff
@@ -520,6 +518,41 @@ func (cd *clusterDeploy) cacheAgentImages(name string) error {
 	return nil
 }
 
+func (cd *clusterDeploy) getPrivateRegistrySecret(cluster *v3.Cluster) (map[string][]byte, error) {
```
Oats87, line 521: nit: We should probably add a comment outlining the logic in summary form, and maybe call it `getPrivateRegistrySecretData` as it's returning data rather than the secret itself.

pkg/provisioningv2/rke2/planner/agent.go
```diff
@@ -37,5 +40,22 @@ func (p *Planner) generateClusterAgentManifest(controlPlane *rkev1.RKEControlPla
 		return nil, err
 	}
 
-	return systemtemplate.ForCluster(mgmtCluster, tokens[0].Status.Token, taints)
+	var formattedPrivateRegistry map[string][]byte
+	privateRegistrySecret := mgmtCluster.GetSecret("PrivateRegistrySecret")
```
Oats87, line 44: Should we make `PrivateRegistrySecret` and `PrivateRegistryURL` constants? I see them used in multiple places, which introduces the opportunity for mistake.

Oats87 (approved):
Overall, LGTM although I have not spun it up to test it myself.

Few nits, but not important to fix.

pkg/provisioningv2/rke2/planner/agent.go
HarrisonWAffel: Yeah that's a good idea, I'll add one into `cluster_types.go`, but if there's a better place lmk

jakefhyde (dismissed):

HarrisonWAffel:
Build was failing due to linter checks, this should be fixed now

jakefhyde (approved):

PR #39067, opened 2022-09-21
[release/v2.7] Surround ipv6 address in brackets for join URL

jakefhyde:
## Problem

The join URL that control plane nodes would use to join a cluster would always be an ipv4 address, scraped from the etcd leader. The join URL that worker nodes would use would instead be the first controlplane node's last `InternalIP` address (or `ExternalIP` if no suitable `InternalIP` is found) as shown in it's status, like so:
```
status:
  addresses:
  - address: 165.227.189.32
    type: InternalIP
  - address: 2604:a880:800:10::7c6:7001
    type: InternalIP
  - address: jhyde-rke2-test-controlplane-838bba1c-8fhkh
    type: Hostname
```
In this case, the ipv6 address was being used, but it was being rendered as `https://2604:a880:800:10::7c6:7001:9345`, with `9345` being appended as the port. This is an invalid URL, as ipv6 hosts must be enclosed with `[]` like so: `https://[2604:a880:800:10::7c6:7001]:9345`

## Solution

Enclose ipv6 addresses with brackets.

## Testing

Create a 3 node (1 etcd, 1 controlplane, 1 worker) cluster on Digital Ocean with ipv6 enabled. The worker node will get stuck in `waiting for probes: calico, kubelet`

### Manual Testing

Nodes do get stuck in this state, but that's not what this PR is fixing. I need to investigate whether or not it's possible to fix downstream worker nodes by overwriting their plans, since it seems like we purposefully wait for the node plan to reconcile, but the `rke2-server` will become inactive. 

### Automated Testing

I added unit tests to ensure that both hostnames and ipv4 addresses are left untouched.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


jakefhyde:
@Oats87 Do you know why we don't update the plan if it's not equal here? In this case the `joinURL` for the controlplane host changed, but we aren't updating the plan so the workers never come online. Seems like a potential race condition, like if the controlplane host becomes uncreachable before a worker node can apply the plan, it will never stop applying the plan because `maxFailures == -1`.
https://github.com/rancher/rancher/blob/a678d0e12fe21936f09ae0e906bc420e43d7d35c/pkg/provisioningv2/rke2/planner/planner.go#L559-L569

HarrisonWAffel (approved):

Oats87 (approved):
great!!!

HarrisonWAffel (approved):

Oats87 (approved):

Issue #37816, opened 2022-05-25
Cert rotation integration tests for RKE2 


Sahota1225:
This is a placeholder ticket for Cert rotation which tests the following cases: 
create an integration test for rke2/k3s

sowmyav27:
@jakefhyde @igomez06 Do we need a backport to 2.6 for this?

igomez06:
> @jakefhyde @igomez06 Do we need a backport to 2.6 for this?

Definitely, @jakefhyde could do you back port PR? The review will be much faster this time around

jakefhyde:
@sowmyav27 @igomez06 Can do, which milestone should it go into, v2.6.9 or v2.6.10?

sowmyav27:
@jakefhyde 2.6.9? Since the code is already in 2.7, this specific pick must be cherry picked to 2.6

PR #39089, opened 2022-09-22
Added rke.cattle.io client for v2prov validation tests

jakefhyde:
## Issue: 
https://github.com/rancher/rancher/issues/37816

Validation tests for v2prov certificate rotation


tests/v2/validation/provisioning/rke2/cert_rotation_test.go
```diff
@@ -0,0 +1,173 @@
...
+			assert.Equal(r.T(), clusterName, clusterResp.ObjectMeta.Name)
+
+			cluster, err = r.client.Provisioning.Cluster.ByID(clusterResp.ID)
+			require.NoError(r.T(), err)
+			require.NotNil(r.T(), cluster.Status)
+
+			// rotate certs
+			require.NoError(r.T(), r.rotateCerts(clusterName, 1))
+			// verify certs rotated
+
```
HarrisonWAffel, line 98: The comment `// verify certs rotated` makes me think there is some logic missing here, is there? I can see that `r.rotateCerts` waits for the `CertificateRotationGeneration` to equal the expected generation, is that enough to verify they have been rotated? Same question in regards to the below comment `// wait until cluster Updated == True`.

tests/framework/clients/rke/client.go
```diff
@@ -0,0 +1,36 @@
...
+func NewForConfig(c *rest.Config, ts *session.Session) (*Client, error) {
+	rkeClient, err := rkeClientV1.NewForConfig(c)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{rkeClient, ts}, nil
+}
+
+// RKEControlPlanes takes a namespace a returns an RKEControlPlane object that is used for the CRUD of a pkg/apis/rke.cattle.io/v1 RKEControlPlane
```
HarrisonWAffel, line 33: super small nit: `RKEControlPlanes takes a namespace a returns` -> `RKEControlPlanes takes a namespace and returns`

tests/framework/pkg/schemas/rke.cattle.io/v1/schema.go
```diff
@@ -0,0 +1,21 @@
...
+		Group:   "rke.cattle.io",
+		Path:    "/v1",
+	}
+
+	Schemas = RKESchemas(&Version).
+		Init(rkeTypes)
+)
+
+func rkeTypes(schemas *types.Schemas) *types.Schemas {
+	return schemas.MustImport(&Version, v1.RKEControlPlane{})
```
igomez06, line 20: Hi Jake, just checking do you not need to have the `schema.ID` set to `"rke.cattle.io" `?

igomez06 (commented):
Looks good overall just had a question about the schema id

tests/framework/pkg/schemas/rke.cattle.io/v1/schema.go
jakefhyde: Truth be told I'm not sure, this seemed to be the minimal configuration I needed to generate the clients.

jakefhyde: Truth be told I'm not sure, this seemed to be the minimal configuration I needed to generate the clients.

igomez06: Oh ok. I'm just checking, because for the provisioning client for example, I needed to update the schema id to be  able to hit the expect API endpoint. If that is not the case here, it looks good to me. The tests run fine?

igomez06: Sounds good thank you Jake. Looks go to me

igomez06 (dismissed):
LGTM

tests/v2/validation/provisioning/rke2/cert_rotation_test.go
jakefhyde: I was going to add some logic to scrape the certs from the node to verify that they were indeed rotated, but realistically we don't need to do that, verifying the generation is all we need since that would indicate a bug in rke2/k3s, which I think is out of the scope of these tests since a failure to rotate should cause the cert rotation operation to fail/not complete. I've removed the comments.

HarrisonWAffel (approved):

Issue #38345, opened 2022-07-20
[BUG] [2.5]Rancher API Audit Log fails to capture API calls at Audit Level 3


samjustus:
Backport rancher/rancher#38323

KevinJoiner:
## Root Cause
When Audit Level is set to level 3, Rancher attempts to add the response body to the audit message. Rancher then prints the compacted JSON. This audit message would fail if the response body was composed of compressed data since no decompression was performed before attempting to create a valid JSON. The error for this audit message failure was ignored, so the user had no indication of what went wrong.

## What fix or change occurred
The audit logger now checks the content-encoding of the response message to determine if the data is compressed. Audit logger then attempts to decompress this data and append it to the log message.
The audit logger will also log errors that occur during audit writes in Rancher.

## Areas or cases that should be tested.
***While testing, we will have to search the auidtLog for messages, this can be difficult because there are a lot of messages***
***One method is to grep the log file for the action you are looking for example `tail -f /var/log/auditlog/rancher-api-audit.log | grep nginx` will get only log messages that include nginx*** 

1. Start a new rancher cluster with API auditing level 3 enabled
    - via helm `--set auditLog.level=3` 
    - via docker `--env AUDIT_LEVEL=3`
2. Create a new deployment in the local cluster.
3. Verify that the request for the deployment was recorded in the audit log.
   - HA: `kubectl logs -n cattle-system -f rancher-<gen_name> rancher-audit-log`
   - docker: run `cat /var/log/auditlog/rancher-api-audit.log` in docker container.

## What areas could experience regressions
Verify that other audit levels still work as they did before. 

#### Audit level 2 test
1. Start a new rancher cluster with API auditing level 2 enabled
    - via helm `--set auditLog.level=2` 
    - via docker env `AUDIT_LEVEL=2`
2. Create a new deployment in the local cluster (nginx:latest) through the UI. 

3. Verify that the request for the deployment was recorded in the audit log without a `responseBody` section.
   - HA: `kubectl logs -n cattle-system -f rancher-<gen_name> rancher-audit-log`
   - docker: run `cat /var/log/auditlog/rancher-api-audit.log` in docker container.

#### Audit level 1 test
1. Start a new rancher cluster with API auditing level 1 enabled
    - via helm `--set auditLog.level=1` 
    - via docker env `AUDIT_LEVEL=1`
2. Create a new deployment in the local cluster (nginx:latest) through the UI. 
[...]

PR #39123, opened 2022-09-26
[2.5] Adds decompression and error handling to the audit logger.

KevinJoiner:
Issue: https://github.com/rancher/rancher/issues/38345
Backport of this PR: https://github.com/rancher/rancher/pull/38325

### DIFF
--- a/pkg/auth/audit/audit.go
+++ b/pkg/auth/audit/audit.go
@@ -1,9 +1,14 @@
+// Package audit is used to preform audit logging.
 package audit
 
 import (
 	"bytes"
+	"compress/gzip"
+	"compress/zlib"
 	"context"
 	"encoding/json"
+	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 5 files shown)


pkg/auth/audit/audit.go
```diff
@@ -78,6 +96,17 @@ func getUserInfo(req *http.Request) *User {
 	}
 }
 
+func getUserNameForBasicLogin(body []byte) string {
```
MbolotSuse, line 99: This appears to have gotten pulled over in the backport. In 2.6, this appears to be used in [newAuditLog](https://github.com/KevinJoiner/rancher/blob/43660c90aaf2e8b14447eea7d10fab818ab46829/pkg/auth/audit/audit.go#L140). However, 2.5 does not appear to [do to the same](https://github.com/rancher/rancher/blob/96b05b53a48d7c4cc4a9ae5eb4c7353547757af7/pkg/auth/audit/audit.go#L129).

I would recommend one of two paths:
- We port over this new function and the place which uses the function
- We do not port over this function or it's usage at all

LMK what you think.

MbolotSuse (changes requested):
One minor issue that I want to resolve relating to login request functionality that was partially back-ported.

pkg/auth/audit/audit.go
KevinJoiner: I didn't mean to add this function so option b 'We do not port over this function or it's usage at all'

cmurphy (approved):

Issue #38961, opened 2022-09-16
In automation framework rke1 extension doesn't use k8's version parameter


igomez06:
**Is your feature request related to a problem? Please describe.**
In tests/framework/extensions/clusters/clusters.go the rke1 helper ignores the Kubernetes version.

**Describe the solution you'd like**
Make sure the Kubernetes version is used in the function.

https://github.com/rancher/rancher/blob/release/v2.6/tests/framework/extensions/clusters/clusters.go#L110-L140

markusewalker:
Since this is a relatively small change, I'm going to incorporate this with https://github.com/rancher/qa-tasks/issues/461.

markusewalker:
PR: https://github.com/rancher/rancher/pull/39124

markusewalker:
Installed.

PR #39124, opened 2022-09-26
Add RKE1 node scaling support

markusewalker:
## Issue:  

[[RKE1] Scale up/down nodes](https://github.com/rancher/qa-tasks/issues/461)

[In automation framework rke1 extension doesn't use k8's version parameter](https://github.com/rancher/rancher/issues/38961)
 
## Problem

**_Problem 1_**
In the test framework, there is currently no scaling tests for RKE1 or RKE2. Scaling is a fundamental Kubernetes operation and it is crucial that we have routine testing for this.

**_Problem 2_**
In the existing `clusters.go`, there was no field to properly set the Kubernetes version; it was left up to the user-defined `cattle-config.yaml` file. This needed to be properly defined.
 
## Solution

_**Solution 1**_
`nodepools.go` was modified to include new `ScaleWorkerNodePool` that performs the following:

- Creates new worker node pool
- Increases quantity from 1 to 2 to scale upwards to include worker node
- Decreases quantity from 2 back to 1 to scale downwards to remove a worker node
- Deletes the new worker node pool

RKE1 is a bit tricky in terms of terms implementing a `wait.Watch...` function to await nodes in the node pool to be `Active`. Multiple implementations were attempted, but none seemed to be properly working. As such, there are `time.Sleep` set to hardcoded time. This is not ideal and if we can remove that, that is favorable. However, this is the implementation at this time.

_**Solution 2**_
`clusters.go` was given a new field to include the Kubernetes version.
 
### Manual Testing

I performed manual testing with 1 node and 3 node, just as the admin user. This was successful; a full automated run will be ran once the PR is up.

### Automated Testing

Offline Jenkins runs will be conducted and results handed to the reviewers.


tests/framework/extensions/rke1/nodepools/nodepools.go
```diff
@@ -54,3 +60,52 @@ func RKE1NodePoolSetup(client *rancher.Client, nodeRoles []NodeRoles, ClusterID,
...
+	time.Sleep(time.Second * time.Duration(longTimeout))
+
+	nodePoolConfig.Quantity = 2
+
+	updatedPool, err := client.Management.NodePool.Update(nodePool, &nodePoolConfig)
+	if err != nil {
+		return err
+	}
+
+	time.Sleep(time.Second * time.Duration(longTimeout))
```
slickwarren, line 92: As I understand, this is here because we don't have a good `wait.watch` for rke1 at the moment. Could we at least validate that the scaled pool has all nodes active?

markusewalker: Correct to your inquiry. The one solution I thought would work best that I attempted was the following:

- Create node pool struct with only worker node defined
- Create worker node pool
- Create node struct that references the worker node pool that was just created (I was doing this via the struct's `nodePoolID` parameter
- Create a worker node
- Track the status via the `State` parameter found in the worker node struct

The issue with the above is that the worker node does not find its way into the node pool that it is associated with....but that is IMO the simplest way to track if nodes are in a state of `Active`.

markusewalker: As an aside, we have `IsHostedClusterProvisioningReady` in `clusters.go` that I tried to use for both `v3.Node` and `v3.NodePool` objects, but that was not working either.

igomez06: Markus you would need to create your own check func. Do something like this 
`watchInterface, err := r.client.GetManagementWatchInterface(management.NodePoolType, opts)
  require.NoError(r.T(), err)
  `
Then create your own nodePoolReady function. Use `IsHostedClusterProvisioningReady` as reference 

I'm not sure exactly how it would look like because I haven't looked but I assume something like this 

```
func IsNodePoolrReady(event watch.Event) (ready bool, err error) {
	nodePoolUnstructured := event.Object.(*unstructured.Unstructured)
	nodePool := &v3.NodePool{}
	err = scheme.Scheme.Convert(nodePoolUnstructured, nodePool, nodePoolUnstructured())
	if err != nil {
		return false, err
	}
	// check some status or something here. 
	}

	return false, nil
}
```

igomez06: Oh ok coo! Sounds good. What you can also do is this 

```
return kwait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
		resp, err := client.Management.NodePool("id")
	})
```

And get the node pool until that status is active or whatever the condition is like how you described.
import this package `kwait "k8s.io/apimachinery/pkg/util/wait"`

markusewalker: Implemented fix here that removes the time.Sleeps when checking node status. I did keep one as I'm scaling down. This is to allow Rancher some "breathing" room. Without it, it will immediately move to delete the node pool and continue with the next test.

tests/framework/extensions/clusters/clusters.go
```diff
@@ -228,6 +233,8 @@ func CreateRKE1Cluster(client *rancher.Client, rke1Cluster *management.Cluster)
 				return false, fmt.Errorf("there was an error deleting cluster")
 			} else if event.Type == watch.Deleted {
 				return true, nil
+			} else if cluster == nil {
```
igomez06, line 236: What cluster are you referencing here? When would cluster be nil? If cluster is nil then it would return an error at line 218

tests/framework/extensions/rke1/nodepools/nodepools.go
```diff
@@ -54,3 +93,62 @@ func RKE1NodePoolSetup(client *rancher.Client, nodeRoles []NodeRoles, ClusterID,
...
+
+	nodePoolConfig.Quantity = 2
+
+	updatedPool, err := client.Management.NodePool.Update(nodePool, &nodePoolConfig)
+	if err != nil {
+		return err
+	}
+
+	err = wait.Poll(500*time.Millisecond, 30*time.Minute, func() (done bool, err error) {
+		IsNodeReady(client, updatedPool, ClusterID)
```
igomez06, line 133: Same here

tests/framework/extensions/rke1/nodepools/nodepools.go
```diff
@@ -54,3 +93,62 @@ func RKE1NodePoolSetup(client *rancher.Client, nodeRoles []NodeRoles, ClusterID,
...
+		Worker:                  true,
+	}
+
+	nodePool, err := client.Management.NodePool.Create(&nodePoolConfig)
+	if err != nil {
+		return err
+	}
+
+	err = wait.Poll(500*time.Millisecond, 30*time.Minute, func() (done bool, err error) {
+		IsNodeReady(client, nodePool, ClusterID)
```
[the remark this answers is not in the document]
markusewalker: When you call the function by itself without the nested poll, the scaling behavior is incorrect. The node pool gets created and then a worker node is added immediately. But what we want is for the node pool to be created, wait for the node it generates to be `Active` and then a new worker node to be added to that pool.

tests/framework/extensions/clusters/clusters.go
markusewalker: Should be the cluster the testSession generates. Honestly, I added this check here cause `CreateRKE2Cluster` has the exact same check. I noted that when working on this PR, so added same check for parity.

tests/framework/extensions/rke1/nodepools/nodepools.go
```diff
@@ -2,19 +2,58 @@ package rke1
...
+			"clusterId":  ClusterID,
+			"nodePoolId": nodePool.ID,
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodes.Data {
+		if node.State != "active" {
```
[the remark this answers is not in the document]
markusewalker: Good point, I can remove the initial check and see if the behavior is the same. I'll refactor the `active` string to be in an const var instead. Thanks!

tests/framework/extensions/clusters/clusters.go
markusewalker: Got it, line has been removed 👍🏿

igomez06: If it is causing issues, just do the same thing, but with a v3.cluster and check if it's nil

markusewalker:
@igomez06 @slickwarren changes look good to you both?

slickwarren (approved):

PR #39131, opened 2022-09-27
bump ui tags for 2.6.9-rc2

kinarashah:


rancherbot (approved):

PR #39134, opened 2022-09-27
Remove user-node-remove on node remove

jakefhyde:
## Problem

If a node is provisioned on <v2.5.1, rancher is stopped, and the node is deleted; Remove will not be called for the node causing orphaned v1.Nodes which default unmanaged v3.Nodes are then created for.

## Solution

Remove the annotation on sync and v3.Node removal.

### Manual Testing

Tested with a 1 etcd, 1 controlplane, 1 worker. On upgrade from v2.5.1 to v2.7-head, the node is properly cleaned up.

### DIFF
--- a/pkg/controllers/management/node/cleanup.go
+++ b/pkg/controllers/management/node/cleanup.go
@@ -35,19 +35,21 @@ const (
 )
 
 func (m *Lifecycle) deleteV1Node(node *v3.Node) (runtime.Object, error) {
-	logrus.Debugf("Deleting v1.node for [%v] node", node.Status.NodeName)
+	logrus.Debugf("[node-cleanup] Deleting v1.node for [%v] node", node.Status.NodeName)
 	if nodehelper.IgnoreNode(node.Status.NodeName, node.Status.NodeLabels) {
-		logrus.Debugf("Skipping v1.node removal for [%v] node", node.Status.NodeName)
+		logrus.Debugf("[node-cleanup] Skipping v1.node removal for [%v] node", node.Status.NodeName)
 		return node, nil
 	}
 
 	if node.Status.NodeName == "" {
+		logrus.Debugf("[node-cleanup] Skipping v1.node removal for machine [%v] without node name", node.Name)
 		return node, nil
 	}
 
 	cluster, err := m.clusterLister.Get("", node.Namespace)
 	if err != nil {
 		if kerror.IsNotFound(err) {
+			logrus.Debugf("[node-cleanup] Skipping v1.node removal for machine [%v] without cluster [%v]", node.Name, node.Namespace)
 			return node, nil
 		}
 		return node, err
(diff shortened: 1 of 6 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -223,71 +222,80 @@ func (m *Lifecycle) getNodePool(nodePoolName string) (*apimgmtv3.NodePool, error
...
 
-func (m *Lifecycle) Remove(obj *apimgmtv3.Node) (runtime.Object, error) {
-	if obj.Status.NodeTemplateSpec == nil {
-		if err := m.cleanRKENode(obj); err != nil {
-			return obj, err
+func (m *Lifecycle) Remove(machine *apimgmtv3.Node) (obj runtime.Object, err error) {
+	if machine.Annotations[userNodeRemoveCleanupAnnotation] != "true" {
+		// if a node is provisioned on <v2.5.1, rancher is stopped, and the node is deleted; the user-node-remove finalizer
+		// should just be removed here.
+		if machine, err = m.userNodeRemoveCleanup(machine); err != nil {
```
Oats87, line 229: I am slightly concerned that the logic here is checking annotations to decide whether to remove a finalizer. Because we are calling the `userNodeRemoveCleanup` in a `Remove`, the machine's `DeletionTimestamp` will always be not nil, and thus we will constantly be calling `Update` on the object which doesn't make sense to me.

We don't care about the `userNodeRemoveCleanupAnnotation` or finalizer anymore, do we?

jakefhyde: We only update the object if it's not nil, but if that's confusing I can refactor it.

jakefhyde:
@Oats87 I've updated it not update inside the call to `userNodeRemoveCleanup`, and to just remove the finalizer without checking if the annotation was applied for two reasons:
1. It's not a very expensive operation for `Remove`
2. If for any reason the annotation failed to be applied previously (or someone manually removes it), the finalizer won't be removed.

pkg/controllers/management/node/cleanup.go
```diff
@@ -411,9 +411,8 @@ func (m *Lifecycle) userNodeRemoveCleanup(obj *v3.Node) (runtime.Object, error)
 		}
 
 		annos[userNodeRemoveCleanupAnnotation] = "true"
-		newObj.SetAnnotations(annos)
```
jiaqiluo, line 414: Why is this line dropped? 
It makes the logic clear although we know that Map in Go is passed by reference. 😄

jiaqiluo (changes requested):

pkg/controllers/management/node/cleanup.go
jakefhyde: Fixed!

jiaqiluo (approved):
LGTM

Oats87 (dismissed):

jiaqiluo (approved):

Oats87 (approved):

Issue #39107, opened 2022-09-23
[BUG] Invalid groupPrincipals in token objects after 2.6.2 to 2.6.6 upgrade


samjustus:
SURE-5050

**Rancher Server Setup**
Rancher version: 2.6.6
Rancher Cluster:
Rancher version: 2.6.6
Number of nodes: 3
Node OS version: Linux

Downstream Cluster:
Number of Downstream clusters: 3
Node OS: Linux
RKE/RKE2/K3S version: RKE

**Describe the bug**
While creating any Token , there is no Validation happening by default and groupPrincipals is setting as null. When trying to edit that token , it's not allowing to save the token yaml.

========

1. tokens.management.cattle.io "token-8q2tp" was not valid:
2. * <nil>: Invalid value: "The edited file failed validation": unknown object type "nil" in Token.groupPrincipals
=========

**To Reproduce**
Rancher 2.6.6 , Go to Local cluster

1. Create a token ( token-gtxs8 ) either using the kubectl or from UI ( Create API Key from Local cluster)

2. Try to edit the description of token using ; $ kubectl edit token token-gtxs8

3. It won't allow to edit the token and throw the below error.

====

tokens.management.cattle.io "token-gtxs8" was not valid:
* <nil>: Invalid value: "The edited file failed validation": unknown object type "nil" in Token.groupPrincipals
====
4. If you change the groupPrincipals from null to [] as below , and try to edit the token its working

groupPrincipals: []

==
> kubectl edit token token-gtxs8
token.management.cattle.io/token-gtxs8 edited
==

The main concern is its not validating the CRD schema when creating a token first time but its validating while editing the token. Its making lot of manual effort

**Result**
groupPrincipals is set to null while creating a token and there is no initial validation happening, this violates against the CRD schema.

**Expected Result**
While creating a token and there should be the initial validation happening

**Screenshots**
see jira

**Additional context**
What is the workaround:

change the groupPrincipals from null to [] as below , and try to edit the token its working

==
apiVersion: management.cattle.io/v3
authProvider: local
current: false
description: lingo
expired: false
expiresAt: ""
groupPrincipals: []
isDerived: true
kind: Token

crobby:
I've had a quick look at this today.  I think it's a simple fix to have groupPrincipals set correctly on create (currently, it's just absent in the code block that does the create).  I'm still digging into why it's not validated on create, but it is validated on update.

crobby:
https://github.com/rancher/rancher/pull/39147

[...]

PR #39147, opened 2022-09-28
Set GroupPrincipals at token create time

crobby:
In order to avoid having a null GroupPrincipals at token create time, explicitly set the value that is passed to the NewLoginToken call.

## Issue: 
[#39107](https://github.com/rancher/rancher/issues/39107)
 
## Problem

GroupPrincipals is not explicitly set at token creation time, which can lead to if having a null (invalid) value instead of [].
 
## Solution

Updated the NewLoginToken function to use the value for GroupPrincipals that was passed in.
 
## Testing

The provided steps worked to expose the problem and no longer show a problem after this fix is applied.

### Manual Testing

After the change, I was able to deploy a fresh cluster and update the description of a token without needing to also change the value of GroupPrincipals, which was already set to []

### DIFF
--- a/pkg/auth/tokens/manager.go
+++ b/pkg/auth/tokens/manager.go
@@ -668,12 +668,13 @@ func (m *Manager) NewLoginToken(userID string, userPrincipal v3.Principal, group
 	}
 
 	token := &v3.Token{
-		UserPrincipal: userPrincipal,
-		IsDerived:     false,
-		TTLMillis:     ttl,
-		UserID:        userID,
-		AuthProvider:  provider,
-		Description:   description,
+		UserPrincipal:   userPrincipal,
+		GroupPrincipals: groupPrincipals,
+		IsDerived:       false,
+		TTLMillis:       ttl,
+		UserID:          userID,
+		AuthProvider:    provider,
+		Description:     description,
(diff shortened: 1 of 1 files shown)


MbolotSuse (changes requested):
A couple of comments here:
- I'm not sure that we want to add these values on create. This seems to be a legacy field that is no longer filled in our code today (if you take a look, you can see a couple of places that use this field, but very few places which actually update the field). I would be nervous about the impacts of suddenly including this again, as it could cause issues.
- We create tokens in a couple of places, such as [here](https://github.com/rancher/rancher/blob/b3c5e9f37b9b1b1467cf000fa5650842223a360c/pkg/auth/tokens/manager.go#L109). If you intend to go this route, you will need to make sure that you get each area the token is created (or updated, since the issue describes token creation through the kubectl api). If we miss these areas, this issue could re-appear in strange and unique ways, which I would like to avoid.
- I would recommend looking into why we are setting this value, when missing, to `null` as k8s seems to dislike specifically that value. This would be a more comprehensive fix and likely get more to the core of the issue.

crobby:
Closing this request since it seems likely that the ultimate solution will NOT be in this repo.  Instead, the actual fix will likely be in either webhook or wrangler.

PR #39151, opened 2022-09-28
[2.7] Add airgap setup support to the validation charts tests

caliskanugur:
- [x] This PR relies on #39341. And requires updates according to the changes.

**Extensions Updates:**

- Implemented:
  - Default registry and server URL IDs and their values for each of the chart install and update functions
- Refactored:
  - Payload options for each of the charts and the payload options struct to contain the default registry
  - Removed unnecessary double error checks in charts cleanups
  - Gatekeeper zero-value maps

**Tests Updates:**

- Implemented:
  - Istio P0 case Monitoring and Istio charts installation checks
  - Readme for the configuration
- Refactored:
  - Some Istio assertions
  - Removed an unnecessary monitoring helper
  - Gatekeeper typos
  - Moved Istio example application to the resources directory
  - Private helpers error throwing with errors package


caliskanugur:
Backport: #39152

nataliefyi (approved):
LGTM

tests/framework/extensions/charts/ranchergatekeeper.go
```diff
@@ -26,12 +26,22 @@ const (
 
 // InstallRancherGatekeeperChart installs the OPA gatekeeper chart
 func InstallRancherGatekeeperChart(client *rancher.Client, installOptions *InstallOptions) error {
-	hostWithProtocol := fmt.Sprintf("https:// %s", client.RancherConfig.Host)
```
igomez06, line 29: we no longer need to get the host with the protocol?

igomez06 (commented):
Looks good Ugur. I will wait until Steve client is merged in before I approve it.

tests/framework/extensions/charts/ranchergatekeeper.go
caliskanugur: Yes, that's correct. `server-url` v3 setting's value returns the host with the protocol (unless it's overwritten as another string).

caliskanugur:
Updated the v1 steve client changes, ready to review.

Additionally, updated waits to kube poll.

izaac:
Looking good Ugur I just wanted to check what private registry scenarios is this covering.

For non-Airgap does these validations work with:

- `Global Registry` empty. `Cluster Level Registry` set. = use Cluster
- `Global Registry` set. `Cluster Level Registry set.  = use Cluster
- Both empty (default) = public registry (docker.io i think)

For Airgap we usually manual test using the `Global Registry` (systemDefaultRegistry). This is straightforward in your code.

What I'm not sure if the RKE1/RKE2 Custom Clusters can override with the Cluster Level too (like with regular Rancher). I assume that's the case.

igomez06 (commented):
looks good, just want to see some freeform jobs before final approval

izaac:
@caliskanugur what I asked [here](https://github.com/rancher/rancher/pull/39151#issuecomment-1299318761) is not in scope for this change

izaac (approved):
LGTM

igomez06 (approved):
Looks good!

nataliefyi (approved):
LGTM

PR #39157, opened 2022-09-28
Updating UI tags for v2.7.0-rc2

prachidamle:
Updating UI tags for v2.7.0-rc2


rancherbot (approved):

Issue #35587, opened 2021-11-18
Rancher rancher-images.txt includes docker.io/busybox which can not be downloaded due to "too many requests"


Martin-Weiss:
**Rancher Server Setup**

AIRGAPPED

- Rancher version: 2.5.11 and 2.6.2
- Installation option (Docker install/Helm Chart): helm chart / airgapped

**Describe the bug**
When we want to replicate the images based on the list of images from https://github.com/rancher/rancher/releases/download/v2.5.11/rancher-images.txt for airgapped the download of docker.io/busybox fails with "too many requests"

**To Reproduce**
Download docker.io/busybox many times on one day and you will get "too many requests"

**Result**
Airgapped images not complete

**Expected Result**
Image download works

**Additional context**

-> can we add the required image into the rancher namespace so that we do not hit the "too many requests"?

Martin-Weiss:
This is not stale and I just created a Support Case for this..

cbron:
SURE-5044

KevinJoiner:
## What was fixed, or what change has occurred
- audit logging default image is now bci-micro
- bci-busybox is still included for rancher support but not required to run rancher/rancher.
- helm value of busyBoxImage is deprecated and replaced with auditLog.image
- helm value busyBoxImageTag was removed without deprecation since it was never released
- bci-micro is used for audit logging instead of bci-busybox because bci-busybox does not have the required tail option --retry needed to run the sidecar pod.
- helm value auditlog.image will be used by default unless the value busyBoxImage is set.

## Areas or cases that should be tested
Audit logging via HA install.

### Verify audit logs are still visible
1. Install rancher via helm is the following values
    ```
    helm install rancher rancher-latest/rancher \                                                             
    --namespace cattle-system \
    --set hostname=<site_url> \
    --set auditLog.level=2
    ```
2. Log Into rancher and create a new deployment of any kind `nginx:latest`
3. Verify that auditLogs are being produced
    ```
    kubectl logs -n cattle-system -f rancher-<gen_name> rancher-audit-log
    ```
4.  verify that bci-micro image is being used for the rancher sidecar. 
    ```
    kubectl get deployment -n cattle-system rancher -o yaml | grep bci-micro
    ```
   
### Users who upgrade without the busyBoxImage set should automatically be switched to using bci-micro.
1. Install rancher via helm is the following values
    ```
    helm install rancher rancher-stable/rancher \                                                             
    --namespace cattle-system \
    --set hostname=<site_url> \
[...]

PR #39181, opened 2022-09-29
Replace busyBox with bci-busybox and add bci-micro.

KevinJoiner:
## Issue: https://github.com/rancher/rancher/issues/35587

- Replaces the audit logging image `busybox` with `bci-micro`
- `bci-busybox` is still included for rancher support but not required to run rancher/rancher.
- helm value of `busyBoxImage` is deprecated and replaced with `auditLog.image`
- helm value `busyBoxImageTag` was removed without deprecation since it was never released
- `bci-micro` is used for audit logging instead of `bci-busybox` because `bci-busybox` does not have the required `tail` option `--retry` needed to run the sidecar pod.
- helm value `auditlog.image` will be used by default unless the value `busyBoxImage` is set.


chart/README.md
```diff
@@ -171,16 +171,18 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
...
+| `auditLog.image`                         | "registry.suse.com/bci/bci-micro"                                         | ***string*** - Image location for image used to collect audit logs *Note: Available as of v2.7.0*                                                                                                                                                                                       |
+| `aauditLog.imageTag`                     | "15.4.14.3"                                                               | ***string*** - Image tag for image used to collect audit logs *Note: Available as of v2. *Note: Available as of v2.7.0*                                                                                                                                                                 |
```
cmurphy, line 175:
```suggestion
| `auditLog.imageTag`                     | "15.4.14.3"                                                               | ***string*** - Image tag for image used to collect audit logs *Note: Available as of v2. *Note: Available as of v2.7.0*                                                                                                                                                                 |
```

chart/README.md
```diff
@@ -171,16 +171,18 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
...
+| `aauditLog.imagePullPolicy`              | "IfNotPresent"                                                            | ***string*** - Override imagePullPolicy for auditLog images - Never", *"Always", "Never", "IfNotPresent"* *Note: Available as of v2.7.0*                                                                                                                                                |
```
cmurphy, line 176:
```suggestion
| `auditLog.imagePullPolicy`              | "IfNotPresent"                                                            | ***string*** - Override imagePullPolicy for auditLog images - Never", *"Always", "Never", "IfNotPresent"* *Note: Available as of v2.7.0*                                                                                                                                                |
```

cmurphy (commented):
The bci-micro image is 24X larger than the busybox image:

```
$ docker history registry.suse.com/bci/bci-micro:15.4.14.3
IMAGE          CREATED      CREATED BY     SIZE      COMMENT
d981c9ccf19f   3 days ago   KIWI 9.24.36   24.1MB
$ docker history rancher/mirrored-library-busybox:1.34.1
IMAGE          CREATED       CREATED BY                                      SIZE      COMMENT
2bd29714875d   2 weeks ago   /bin/sh -c #(nop)  CMD ["sh"]                   0B        
<missing>      2 weeks ago   /bin/sh -c #(nop) ADD file:53723549e9db3eb5b…   1.24MB
```

Has this tradeoff already been discussed? Is there a smaller image that could serve the same purpose, or a possibility of an updated busybox image?

> bci-busybox is still included for rancher support but not required to run rancher/rancher.

What is it used for exactly, is it just included for convenience? Could we leave it out to pare down on the number of images needed?

chart/values.yaml
```diff
@@ -18,14 +18,16 @@ auditLog:
...
   maxSize: 100
+  image: "registry.suse.com/bci/bci-micro"
+  imageTag: 15.4.14.3
+  imagePullPolicy: "IfNotPresent"
 
 # As of Rancher v2.5.0 this flag is deprecated and must be set to 'true' in order for Rancher to start
 addLocal: "true"
 
 # Image for collecting rancher audit logs.
 # Important: update pkg/image/export/main.go when this default image is changed, so that it's reflected accordingly in rancher-images.txt generated for air-gapped setups.
```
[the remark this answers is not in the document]
KevinJoiner: Yes and the values can be removed since we do not have default values.

chart/tests/deployment_test.yaml
```diff
@@ -62,26 +62,59 @@ tests:
...
       level: 1
-    busyboxImage: my.private.repo:5000/rancher/busybox
-    busyboxImageTag: 1.0.1
+      image: my.private.repo:5000/rancher/busybox
+      imageTag: 1.0.1
   asserts:
   - equal:
       path: spec.template.spec.containers[1].image
       value: my.private.repo:5000/rancher/busybox:1.0.1
+- it: deprecated busyBoxImage should override auditLog.image image when auditLog.destination is sidecar and auditLog.level > 0 and auditLogImage and busyBoxImage are set
```
rmweir, line 85:
```suggestion
- it: deprecated busyboxImage should override auditLog.image image when auditLog.destination is sidecar and auditLog.level > 0 and auditLogImage and busyboxImage are set
```

rmweir (changes requested):

KevinJoiner:
> The bci-micro image is 24X larger than the busybox image:
> 
> Has this tradeoff already been discussed? Is there a smaller image that could serve the same purpose, or a possibility of an updated busybox image?

Our main problem with the BusyBox image is the GPL license used by busybox. BCI images have source code references and licensing handled for us, so I think we are okay with the increased size since we are confident that all binaries in the image are correctly licensed. 
@cbron, let me know if you disagree.

> > bci-busybox is still included for rancher support but not required to run rancher/rancher.
> 
> What is it used for exactly, is it just included for convenience? Could we leave it out to pare down on the number of images needed?

I believe it is only included for convince in air-gapped setups, but @cbron should have a better answer.

rmweir (approved):
LGTM

pkg/image/resolve.go
```diff
@@ -130,7 +130,8 @@ func setRequirementImages(osType OSType, imagesSet map[string]map[string]struct{
 	case Linux:
 		addSourceToImage(imagesSet, settings.ShellImage.Get(), coreLabel)
 		addSourceToImage(imagesSet, settings.MachineProvisionImage.Get(), coreLabel)
-		addSourceToImage(imagesSet, "rancher/mirrored-library-busybox:1.34.1", coreLabel)
+		addSourceToImage(imagesSet, "registry.suse.com/bci/bci-busybox:15.4.11.2", coreLabel)
+		addSourceToImage(imagesSet, "registry.suse.com/bci/bci-micro:15.4.14.3", coreLabel)
```
cbron, line 134: We could use sp3 instead, but sp4 is stable now. So we should instead advance the other images up to sp4.

chart/values.yaml
```diff
@@ -19,18 +19,17 @@ auditLog:
   maxBackup: 1
   maxSize: 100
 
-# As of Rancher v2.5.0 this flag is deprecated and must be set to 'true' in order for Rancher to start
-addLocal: "true"
+  # Image for collecting rancher audit logs.
+  # Important: update pkg/image/export/resolve.go when this default image is changed, so that it's reflected accordingly in rancher-images.txt generated for air-gapped setups.
+  image: "registry.suse.com/bci/bci-micro"
+  imageTag: 15.4.14.3
```
cbron, line 25: Since these are new, we should go with our standard: 
```yaml
image:
   repository:
   tag:
   imagePullPolicy:
```

chart/templates/deployment.yaml
```diff
@@ -191,8 +191,16 @@ spec:
 {{- if eq .Values.auditLog.destination "sidecar" }}
   {{- if gt (int .Values.auditLog.level) 0 }}
       # Make audit logs available for Rancher log collector tools.
-      - image: {{ .Values.busyboxImage }}:{{.Values.busyboxImageTag}}
-        imagePullPolicy: {{ default "IfNotPresent" .Values.busyboxImagePullPolicy }}
+      {{- if .Values.busyboxImage }}
```
cbron, line 194: I was thinking that everyone upgrading would have this set, but I suppose only if you had overridden the default. We should make sure to write upgrade as a test case.

chart/README.md
```diff
@@ -171,16 +171,18 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
...
 | `ingress.extraAnnotations`               | {}                                                                        | ***map*** - additional annotations to customize the ingress                                                                                                                                                                                                                             |
 | `ingress.configurationSnippet`           | " "                                                                       | ***string*** - Add additional Nginx configuration. Can be used for proxy configuration. Note: *Available as of v2.0.15, v2.1.10 and v2.2.4*                                                                                                                                             |
```
cbron, line 187: Need to open a docs issue to update docs as well: https://docs.ranchermanager.rancher.io/reference-guides/installation-references/helm-chart-options#advanced-options

cbron (commented):
Thanks for adding tests.

> Has this tradeoff already been discussed? Is there a smaller image that could serve the same purpose, or a possibility of an updated busybox image?

Yes, we'll need to go with bci-micro because it has the tools we need. bci-busybox doesn't have the right tail package.

> bci-busybox is still included for rancher support but not required to run rancher/rancher.
> What is it used for exactly, is it just included for convenience? Could we leave it out to pare down on the number of images needed?

Good idea trying to pare down images, but we do actually need it for debugging air-gap setups. It's only 14mb, so it's not adding a terrible amount.

chart/README.md
KevinJoiner: Made a PR for the changes https://github.com/rancher/rancher-docs/pull/239

cmurphy (approved):

rmweir (approved):

PR #39186, opened 2022-09-30
Change customLinks default to empty string

Shavindra:
This should be an array instead of an object.


macedogm (approved):

macedogm:
Seems a trivial fix and it has the context in https://github.com/rancher/dashboard/pull/6736#issuecomment-1268308176.

PR #39191, opened 2022-09-30
[v2.7] Add logic and CI checks for image source code origins

HarrisonWAffel:
This PR accompanies the upcoming trusted registry and associated work targeted for 2.7.0. 

This PR introduces a new build artifact for Rancher, named `rancher-origins.txt`. This artifact contains a mapping between all images used in Rancher and their origin repository. This allows users of Rancher to look at the source code of any image being used by an out-of-the box Rancher installation. All images which are produced in `rancher-images.txt` & `rancher-windows-images.txt` are expected to have a URL pointing to their source code repository. 

This PR also introduces an additional CI check to ensure that all images have a URL associated with them, failing builds if images are unaccounted for. If all images are accounted for then the new artifact will be generated and have a similar structure to `rancher-sources.txt` - a space delimited list of images and their source code repository URL's. 

A tool has also been written which will attempt to automatically resolve source code repositories for some images, however it is primarily written to resolve Rancher owned images which share a name with their base repository. A warning will also be printed out which denotes images that are no longer used but are still present within `pkg/image/origins.go`. This tool can be run with the command `dapper check-origins`, and is helpful for addressing any CI errors that may occur due to this new process. 

This mapping of images to source code URL's is maintained in `pkg/image/origins.go`, in the variable titled `OriginMap`. This map does not take into account image versions, and only needs to be updated when net-new images are added to Rancher. When images are no longer used by Rancher they should be removed from `pkg/image/origins.go`, however leaving them in will not cause a build failure (though these unused images _will_ be included in the release artifact). The command `dapper check-origins` will print a warning if it finds any images which are no longer used by rancher.

I've done my best to identify the source code repository for each of the images within the currently released `rancher-images.txt` & `rancher-windows-images.txt`, and have updated the map within `/pkg/image/origins.go`. However, this map should be double-checked to ensure that I have found the correct repository for each image, and if not comment on this PR with the correct URL so that I may update it.


pkg/image/origins.go
```diff
@@ -0,0 +1,306 @@
...
+	// then we cannot continue as the
+	// artifact would be incomplete.
+	if UnknownImages := GatherUnknownImages(allImages); len(UnknownImages) > 0 {
+		return fmt.Errorf("[ERROR] not all images have a source code origin defined. Please provide origin URL's within rancher/pkg/image/origins.go for the following images: \n%s", UnknownImages)
+	}
+
+	// format file
+	fileContents := ""
+	for k, v := range OriginMap {
+		fileContents = fileContents + k + " " + v + "\n"
```
jakefhyde, line 242: QQ: how long does this whole script take to run?

HarrisonWAffel: If you have the image needed for dapper already cached then it takes about 1 minute, most of that time is spent cloning the charts repo and unpacking it so we can get at the images. If you don't have it cached it might take 2-3 minutes, but again that's largely due to dapper setting up and cloning the repositories which can't really be avoided.

pkg/image/origins.go
```diff
@@ -0,0 +1,306 @@
...
+		return ""
+	}
+
+	return split[0]
+}
+
+// UniqueTargetImages finds unique images in a list of targetImages,
+// ignoring the image version
+func UniqueTargetImages(targetImages []string) []string {
+	seenImages := make(map[string]bool)
```
jakefhyde, line 295: I think you can replace this with `map[string]struct{}`, which is idiomatic for set containers.

jakefhyde (changes requested):

pkg/image/origins/main.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return "", err
+	}
+	if r.StatusCode != http.StatusOK {
+		return "", fmt.Errorf(imageNotFound)
+	}
+	return url, nil
+}
+
+func GetWindowsAgentImage() string {
```
jiaqiluo, line 274: The same function is already defined in `pkg/image/export/main.go`. 
How about creating a new file `pkg/image/utilities.go` for all common helper functions?

pkg/image/origins/main.go
```diff
@@ -0,0 +1,284 @@
...
+
+func main() {
+	if err := inner(os.Args[1], os.Args[2], os.Args[3:]); err != nil {
+		panic(err)
+	}
+}
+
+const imageNotFound = "image not found"
+
+func inner(systemChartsPath, chartsPath string, imagesFromArgs []string) error {
```
jiaqiluo, line 37: Most of this function is the same logic as the function `run` in `pkg/image/export/main.go` 
How about extracting the common parts to make a helper function?

scripts/check-origins
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
jiaqiluo, line 1: This file is almost identical to the file `scripts/package`,
how about moving the shared parts into another script file? 

Another reason for the suggestion is that today we need to change the default value of `SYSTEM_CHART_DEFAULT_BRANCH` between `dev-v2.x` and `release-v2.x` in the file `scripts/package` before/after releasing rancher.  If we add this file, it means one more place for us to change the same variable.

jiaqiluo (changes requested):

scripts/check-origins
HarrisonWAffel: I'll move the common code into a shared script, though the `check-origins` script is expected to only be used in a developer workflow, it doesn't contribute to a release or produce any artifacts on its own so I wasn't anticipating needing to update it.

HarrisonWAffel: A new file, `package-env`, was created to handle the common environment variables between the two scripts

pkg/image/origins/main.go
HarrisonWAffel: `utilities.go` now handles the processing of the images and image sources within `GatherTargetImagesAndSources`, which is used by both `img/export/main.go` and `img/origins/main.go`.

pkg/image/origins.go
HarrisonWAffel: I've changed this in two placed, both `UniqueTargetImages` and `CheckForImagesNoLongerBeingUsed`.

pkg/image/origins/main.go
HarrisonWAffel: Created `utilities.go` to handle shared logic 👍

pkg/image/origins.go
HarrisonWAffel: I've update this for both places where we identify unique images

pkg/image/utilities/utilities.go
```diff
@@ -0,0 +1,733 @@
...
+
+	if !img.IsValidSemver(rancherVersion) || strings.HasPrefix(rancherVersion, "dev") || strings.HasPrefix(rancherVersion, "master") || strings.HasSuffix(rancherVersion, "-head") {
+		rancherVersion = settings.RancherVersionDev
+	}
+	rancherVersion = strings.TrimPrefix(rancherVersion, "v")
+
+	// already downloaded in dapper
+	b, err := ioutil.ReadFile(filepath.Join("data.json"))
+	if os.IsNotExist(err) {
+		b, err = ioutil.ReadFile(filepath.Join(os.Getenv("HOME"), "bin", "data.json"))
```
jiaqiluo, line 63: nit: `ioutil.ReadFile` is deprecated, this function simply calls `os.ReadFile`.

pkg/image/export/main.go
```diff
@@ -157,617 +39,29 @@ func run(systemChartsPath, chartsPath string, imagesFromArgs []string) error {
...
 		if err != nil {
 			return err
 		}
 
-		if err := imagesAndSourcesText(arch, imageLists.imagesAndSources); err != nil {
-			return err
-		}
-		err = mirrorScript(arch, imageLists.images)
-		if err != nil {
+		if err := utilities.ImagesAndSourcesText(arch, imageLists.imagesAndSources); err != nil {
```
jiaqiluo, line 47: nit 
```suggestion
		if err = utilities.ImagesAndSourcesText(arch, imageLists.imagesAndSources); err != nil {
```

jiaqiluo (changes requested):

pkg/image/export/main.go
HarrisonWAffel: good catch

jiaqiluo (approved):
LGTM.

@HarrisonWAffel , there is one more follow-up after the PR is merged:
The releasing checklist in Confluence needs to be updated because this PR moves variables from `scripts/package` to `scripts/package-env`. 

You might need to reach out to Jono for all affected files.

pkg/image/origins/main.go
```diff
@@ -0,0 +1,188 @@
...
+			// hardened images conversion, best effort
+			cleanRepo := strings.ReplaceAll(repo, "hardened-", "image-build-")
+			url, err := checkURL("https://www.github.com/rancher/" + cleanRepo)
+			if err != nil && err.Error() != imageNotFound {
+				return err
+			}
+			if err != nil {
+				url = "unknown"
+			}
+			imgToURL[repo] = url
```
jakefhyde, line 160:
```suggestion
		case strings.Contains(repo, "hardened"):
			// hardened images conversion, best effort
			repo = strings.ReplaceAll(repo, "hardened-", "image-build-")
			fallthrough
```
I'm not saying that you have to, just that it's possible.

jakefhyde (approved):

pkg/image/origins/main.go
HarrisonWAffel: Yeah, I had thought about that too its just that some of the hardened images have slightly different names than their base repo's, for instance the `hardened-dns-node-cache` comes from `https://github.com/rancher/image-build-dns-nodecache`, which doesn't have a `-` separating `node` and `cache`.

I'll do a double check of the hardened images tomorrow to make sure I have them all correct, the list still isn't completely finalized

jakefhyde (dismissed):

superseb (commented):
Was it a design choice to have all mappings in the code? There is code to derive the source repository from the repository name but I still see the mappings that match that in the code mappings while these should be accounted for. I believe it is more confusing to have entries in the mappings that can be derived as we call out we have logic for those cases. 

And some images might have `org.opencontainers.image.url` pointing to the source repository, is there a specific reason not to use this as a source for finding out the source? I guess this builds on the previous answer, do we want a full list with origins or auto-detect as much as possible and have the list of exceptions that can't be derived.

pkg/image/origins.go
```diff
@@ -154,6 +160,8 @@ func init() {
 		"mirrored-longhornio-longhorn-ui":                         "https://github.com/longhorn/longhorn-ui",
 		"mirrored-messagebird-sachet":                             "https://github.com/messagebird/sachet",
 		"mirrored-metrics-server":                                 "https://github.com/kubernetes-sigs/metrics-server",
+		"mirrored-minio-mc":                                       "https://github.com/minio/mc",
```
jakefhyde, line 163: Is it feasible to add a check to `convertImagesToRepoUrls` to check for this the same way we do for `hardened`? Seems mirrored images could potentially check `rancher/<REPO>`, then `<ORG>/<REPO>` before setting as unknown.

KevinJoiner (changes requested):
The Team1 managed charts look good. 

Below are some nits because I don't know how short our timeline is for this PR:
- We should have comments on all exported functions.  https://go.dev/doc/comment
- This still needs unit tests.
- error wrapping on errors returned by external packages.

HarrisonWAffel:
@superseb the reason I've included them all in the map is that I'm trying to avoid lengthening the time it takes for CI to run as much as possible. Currently the means in which we automatically resolve image origins is via synchronous `GET` requests, and as more images are added that will add time to each build. I'm open to ideas on how to refine this, I just don't want to extend the time it takes to complete CI due to this artifact.

Also, it may be difficult to find the actual source code of a particular image if we simply state its source as another image, which is why i opted for URL's when possible. ideally, we want to cover all images specified in `rancher-images.txt` and `rancher-windows-images.txt` so that when customers pull these images into their own private repositories they can verify what is being executed.

pkg/image/origins.go
HarrisonWAffel: Good idea, added in latest commit

pkg/image/utilities/utilities.go
```diff
@@ -0,0 +1,732 @@
...
+		"linux":   "rancher-images.txt",
+		"windows": "rancher-windows-images.txt",
+	}
+	sourcesFilenameMap = map[string]string{
+		"linux":   "rancher-images-sources.txt",
+		"windows": "rancher-windows-images-sources.txt",
+	}
+)
+
+func GatherTargetImagesAndSources(systemChartsPath, chartsPath string, imagesFromArgs []string) ([]string, []string, []string, []string, []string, error) {
```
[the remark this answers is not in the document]
HarrisonWAffel: I've added the aggregate type `ImageTargetsAndSources` to handle those slices

KevinJoiner:
When we are doing error wrapping we should use `fmt.Errorf("%w")`
We do not want to use the package `github.com/pkg/errors` it was made obsolete by go 1.13 when they added error wrapping to the standard lib. And the package has since been archived.

pkg/image/origins/main.go
```diff
@@ -0,0 +1,201 @@
...
+		switch {
+		case strings.Contains(repo, "mirrored"):
+			// mirrored images, assumes everything after 'mirrored-'
+			// can be directly converted into a github repository.
+			ownerAndRepository := strings.Split(strings.ReplaceAll(repo, "mirrored-", ""), "-")
+			if len(ownerAndRepository) <= 1 {
+				imgToURL[repo] = "unknown"
+				continue
+			}
+			url, err := checkURL(fmt.Sprintf("https://www.github.com/%s/%s", ownerAndRepository[0], strings.Join(ownerAndRepository[1:], "")))
```
jakefhyde, line 155: `https://www.github.com` just 301s to `https://github.com`. I don't think it will be a measurable time save, but might be worth trying.

KevinJoiner (changes requested):

pkg/image/origins.go
```diff
@@ -233,7 +232,7 @@ var OriginMap = map[string]string{
 // ensures that they are covered by the OriginMap. If so, a file titled `rancher-image-origins.txt` will be written,
 // containing a space delimited list of un-versioned  images and their source code repository.
 func GenerateImageOrigins(linuxImagesFromArgs, targetImages, targetWindowsImages []string) error {
-	logrus.Infof("building rancher-image-origins.txt")
+	log.Printf("building rancher-image-origins.txt")
```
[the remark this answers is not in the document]
HarrisonWAffel: This was done to ensure the log message matches the format of existing log messages, I noticed it didnt when addressing the above comments. 

```
INFO[0000] generating k3s image list...                 
INFO[0006] finished generating k3s image list...        
INFO[0006] generating rke2 image list...                
INFO[0019] finished generating rke2 image list...       
INFO[0019] building rancher-image-origins.txt           // doesn't match the below messages 
2022/10/07 21:57:58 Creating rancher-images.txt
2022/10/07 21:57:58 Creating rancher-images-sources.txt
2022/10/07 21:57:58 Creating rancher-mirror-to-rancher-org.sh
2022/10/07 21:57:58 Creating rancher-save-images.sh
2022/10/07 21:57:58 Creating rancher-load-images.sh
2022/10/07 21:57:58 Creating rancher-windows-images.txt
2022/10/07 21:57:58 Creating rancher-windows-images-sources.txt
2022/10/07 21:57:58 Creating rancher-mirror-to-rancher-org.ps1
2022/10/07 21:57:58 Creating rancher-save-images.ps1
2022/10/07 21:57:58 Creating rancher-load-images.ps1
```

Is now 

```
INFO[0000] generating k3s image list...                 
INFO[0003] finished generating k3s image list...        
INFO[0003] generating rke2 image list...                
INFO[0012] finished generating rke2 image list...       
2022/10/11 15:23:51 building rancher-image-origins.txt
2022/10/11 15:23:51 Creating rancher-images.txt
2022/10/11 15:23:51 Creating rancher-images-sources.txt
2022/10/11 15:23:51 Creating rancher-mirror-to-rancher-org.sh
2022/10/11 15:23:51 Creating rancher-save-images.sh
2022/10/11 15:23:51 Creating rancher-load-images.sh
2022/10/11 15:23:51 Creating rancher-windows-images.txt
2022/10/11 15:23:51 Creating rancher-windows-images-sources.txt
2022/10/11 15:23:51 Creating rancher-mirror-to-rancher-org.ps1
2022/10/11 15:23:51 Creating rancher-save-images.ps1
2022/10/11 15:23:51 Creating rancher-load-images.ps1

```

pkg/image/origins.go
HarrisonWAffel: I added that logic in to the latest commits 👍

KevinJoiner (approved):

jakefhyde (approved):

jiaqiluo (approved):

HarrisonWAffel:
Could someone from @rancher/rancher2-helm-yes-dev please take a look at this PR before EOD PST to check that the URL's I've provided are correct? As tomorrow is code freeze and I've gotten approvals from team 1 and team 2 I'll be merging this PR EOD PST, but would appreciate a +1 from team 3 so that they're aware of the new CI check.

pkg/image/origins.go
```diff
@@ -0,0 +1,319 @@
...
+	"openvswitch":                                             "https://github.com/servicefractal/ovs",
+	"opflex":                                                  "https://github.com/noironetworks/opflex",
+	"opflex-server":                                           "https://github.com/noironetworks/opflex",
+	"pause":                                                   "https://hub.docker.com/r/kubernetes/pause",
+	"prom-alertmanager":                                       "https://github.com/prometheus/alertmanager",
+	"prom-node-exporter":                                      "https://github.com/prometheus/node_exporter",
+	"prom-prometheus":                                         "https://github.com/prometheus/prometheus",
+	"prometheus-auth":                                         "https://github.com/rancher/prometheus-auth",
+	"prometheus-federator":                                    "https://github.com/rancher/prometheus-federator",
+	"pushprox-client":                                         "https://github.com/prometheus-community/PushProx/tree/master/cmd/client",
```
aiyengar2, line 201: This comes from github.com/rancher/PushProx

aiyengar2 (changes requested):
PushProx images have the wrong source, otherwise LGTM

superseb (approved):

KevinJoiner (approved):

PR #39196, opened 2022-09-30
[v2.7] Add K3S provisioning support for node driver/custom clusters

markusewalker:
## Problem

As of now, the Go test framework only has test cases for RKE1 and RKE2. As we look towards pushing K3S to become GA, it is only right that we account for that in our testing as well. This includes node drivers and custom clusters, just as we currently do for RKE1 and RKE2.
 
## Solution

This PR adds support for K3S provisioning for both node drivers and custom clusters. Essentially, majority of the functionality is taken from RKE2, with the exception that CNI is removed, as K3S does not utilize CNI.
 
### Manual Testing

Locally, validated that a 1 node all roles provisions for both the admin and standard user for node provisioned cluster AND custom cluster.

### Automated Testing

Offline Jenkins jobs will be handed to the reviewers prior to approval of the PR.

### DIFF
--- a/tests/framework/extensions/clusters/clusters.go
+++ b/tests/framework/extensions/clusters/clusters.go
@@ -140,8 +140,8 @@ func NewRKE1ClusterConfig(clusterName, cni, kubernetesVersion string, client *ra
 	return clusterConfig
 }
 
-// NewRKE2ClusterConfig is a constructor for a apisV1.Cluster object, to be used by the rancher.Client.Provisioning client.
-func NewRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName, kubernetesVersion string, machinePools []provisioning.RKEMachinePool) *provisioning.Cluster {
+// NewK3SRKE2ClusterConfig is a constructor for a apisV1.Cluster object, to be used by the rancher.Client.Provisioning client.
+func NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName, kubernetesVersion string, machinePools []provisioning.RKEMachinePool) *provisioning.Cluster {
 	//metav1.ObjectMeta
 	objectMeta := &provisioning.ObjectMeta{
 		Name:      clusterName,
@@ -242,9 +242,9 @@ func CreateRKE1Cluster(client *rancher.Client, rke1Cluster *management.Cluster) 
 	return cluster, nil
 }
 
-// CreateRKE2Cluster is a "helper" functions that takes a rancher client, and the rke2 cluster config as parameters. This function
+// CreateK3SRKE2Cluster is a "helper" functions that takes a rancher client, and the rke2 cluster config as parameters. This function
 // registers a delete cluster fuction with a wait.WatchWait to ensure the cluster is removed cleanly.
-func CreateRKE2Cluster(client *rancher.Client, rke2Cluster *provisioning.Cluster) (*provisioning.Cluster, error) {
+func CreateK3SRKE2Cluster(client *rancher.Client, rke2Cluster *provisioning.Cluster) (*provisioning.Cluster, error) {
 	cluster, err := client.Provisioning.Cluster.Create(rke2Cluster)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


tests/framework/extensions/clusters/clusters.go
```diff
@@ -194,6 +194,55 @@ func NewRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName
 	return v1Cluster
 }
 
+// NewK3SClusterConfig is a constructor for a apisV1.Cluster object, to be used by the rancher.Client.Provisioning client.
+func NewK3SClusterConfig(clusterName, namespace, cloudCredentialSecretName, kubernetesVersion string, machinePools []provisioning.RKEMachinePool) *provisioning.Cluster {
```
igomez06, line 198: Is it necessary to have a whole other function to instantiate a K3SClusterConfig? It looks very similar to the RKE2. Why not have a shared function that does this that is private, then have two separate functions for RKE2 and K3S.

tests/framework/extensions/clusters/clusters.go
```diff
@@ -296,3 +345,58 @@ func CreateRKE2Cluster(client *rancher.Client, rke2Cluster *provisioning.Cluster
 
 	return cluster, nil
 }
+
+// CreateK3SCluster is a "helper" functions that takes a rancher client, and the k3s cluster config as parameters. This function
+// registers a delete cluster fuction with a wait.WatchWait to ensure the cluster is removed cleanly.
+func CreateK3SCluster(client *rancher.Client, k3sCluster *provisioning.Cluster) (*provisioning.Cluster, error) {
+	cluster, err := client.Provisioning.Cluster.Create(k3sCluster)
```
igomez06, line 352: Same comment as above how different is k3s vs rke2 in terms of functions?

igomez06 (commented):
Just a few questions on the difference between k3s and Rke2. If the only difference is the k8s version, I don't see a reason to have separate functions and constructors

tests/framework/extensions/clusters/clusters.go
igomez06: I think in this case we should call ie CreateK3SRKE2Cluster, since this is exactly the same as its RKE2 counter part

tests/framework/extensions/clusters/clusters.go
markusewalker: K3s doesn't use cni, so there are some fields that need to be omitted and I figured it's easier to just have a new function.

tests/framework/extensions/clusters/clusters.go
markusewalker: If we want, we can just use the CreateRKE2Cluster function. I only made this new function as to keep things specific to K3s, but not really an issue to use RKE2Cluster

tests/framework/extensions/clusters/clusters.go
igomez06: I think we can have on shared function that the public K3/RKE2 cluster functions would use to setup the cluster

tests/framework/extensions/clusters/clusters.go
igomez06: I would prefer that. We could even just name it CreateK3SRKE2Cluster

tests/framework/extensions/clusters/clusters.go
markusewalker: A bit slow here, so apologies 😄 . By shared function that is private, you mean to not have the shared function exported and then when calling the function for RKE2, just amend the CNI parameter and that's all right?

igomez06: Yes exactly

markusewalker: I had a chance to further look into this and we could actually just have the one cluster as is. Then, in the K3S files, we just leave the cni parameter "" when calling it. Locally tested it and this works as expected and we already do this for `cloudCredentialSecretName` in the when calling it in the node driver and custom cluster test files.

igomez06 (commented):
Looks good, Quick question will this work by have cni be en empty string? If so great job! I'll approve it once we get Jenkins back and can have some test runs

markusewalker:
Yep, I validated locally that it works with the empty string 😄 . We're on the same page, once Jenkins is back up, I'll run automated runs to make sure we're all good.

igomez06 (approved):

slickwarren (approved):

PR #39212, opened 2022-10-03
Exclude `kind/enhancement` from stalebot

snasovich:
Many other `kind/*` labels are excluded from stalebot. Seems logical to also exclude `kind/enhancement`.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -19,8 +19,8 @@ jobs:
           stale-pr-label: 'status/stale'
           days-before-stale: 60
           days-before-close: 14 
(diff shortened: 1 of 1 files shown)


Sahota1225 (approved):

superseb (commented):
I don't think this label was ever excluded, and I believe the reasoning was that an enhancement can be closed unless there is specific interest for it (hence it not being in the list so reporter/watchers get a notification to keep showing interest to keep it open). But I'm fine with adding it if it causes problems for filed issues.

snasovich:
Given the rationale above, closing this issue. We can use other labels when needed to exclude from stalebot.

PR #39225, opened 2022-10-04
go generate

MbolotSuse:
Pre-requisite for #39107. Some of the generated files appear to be generated with an older version of go. This PR aims to generate them with the newest version available and bring them in line with the version currently being used for the project (go 1.19).


KevinJoiner (approved):
LGTM

crobby (approved):
LGTM  This is the same diff set I was seeing earlier.

mattfarina (approved):

PR #39235, opened 2022-10-05
Omit GroupPrincipals if field is empty

crobby:
Added omitempty to the defintion of the GroupPrincipals field in the Token struct.  This will prevent that field from showing GroupPrincipals: null when a token is created without GroupPrincipals set.

## Issue: 

https://github.com/rancher/rancher/issues/39107
 
## Problem

Currently, it is possible for a Token to be generated with the GroupPrincipals field set to null in the yaml. When someone tries to update that token, the effort will fail since null is not a valid value. To make their update stick, they would need to change GroupPrincipals to be [] (or another valid value) as part of their update.
 
## Solution

The GroupPrincipals field is now set to "omitempty" in the definition of the Token struct, so it will just not be present if it is not set rather than defaulting it to null.
 
### Manual Testing

I've ran many clusters and created many tokens with this update. In all cases, the GroupPrincipals field is no longer included in the object and all of the created tokens can be updated without incident.

## QA Testing Considerations

Install Rancher
Verify that all tokens do NOT include GroupPrincipals: null in the yaml. Unless you created a token that specifically has GroupPrincipals, that field will NOT be present at all in the token object.

### DIFF
--- a/pkg/apis/management.cattle.io/v3/authn_types.go
+++ b/pkg/apis/management.cattle.io/v3/authn_types.go
@@ -19,7 +19,7 @@ type Token struct {
 
 	Token           string            `json:"token" norman:"writeOnly,noupdate"`
 	UserPrincipal   Principal         `json:"userPrincipal" norman:"type=reference[principal]"`
-	GroupPrincipals []Principal       `json:"groupPrincipals" norman:"type=array[reference[principal]]"`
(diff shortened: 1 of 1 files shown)


MbolotSuse (approved):
LGTM. Since this for 2.7.1, please refrain from merging until 2.7.0 has released (or double check before merging).

KevinJoiner (approved):

PR #39265, opened 2022-10-10
Remove quotas from namespaces when they are removed from project

maxsokolovsky:
## Problem
When users remove a resource quota from a project, it still exists and is associated with existing namespaces inside the project.
 
## Solution
In these cases, when Rancher computes the new resource quota, it needs to include only those resources on the quota that are defined with some limit value on the project. If the project quota does not define a limit for some resource, then ignore the resource quota requested. Keep the current behavior where the requested amount trumps the default one, if the default exists in the project.

**Note:** this breaks current behavior, one that we discourage and should mention in the docs. Specifically, the updated logic will prevent users from **adding** new quota resource in their override via Kubectl. Rancher will honor only those resource overrides that actually have a corresponding default on the project quota.
 We should encourage users to instead add regular Kubernetes resource quotas for any custom limits for resources in their namespaces.

## Testing
To reproduce, create a new project with a limit on any resource along with a namespace default limit. Then add a namespace to the project, ensuring that the namespace contains a new quota with the default limit. Remove the limit from the project. Observe that the limit is still present on the namespace's quota.

### Automated Testing
Integration tests:
1. Create a project with a default quota for one resource, add a namespace to it.
2. Create a project with a default quota for one resource, add two namespace, each providing a quota override.
3. Create a project with a default quota for one resource, add a namespace to it. Then add another limit on the project and ensure the limit is added to existing namespaces' quotas.
4. **(this issue)**. Create a project with a default quota for one resource, add a namespace to it. Then remove limit on the project and ensure the limit is removed from existing namespaces' quotas.

### DIFF
--- a/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
+++ b/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
@@ -124,11 +124,11 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
 
 	operation := "none"
 	if existing == nil {
-		if newQuotaSpec != nil {
+		if newQuotaSpec != nil && len(newQuotaSpec.Hard) > 0 {
 			operation = "create"
 		}
 	} else {
-		if newQuotaSpec == nil {
+		if newQuotaSpec == nil || len(newQuotaSpec.Hard) == 0 {
 			operation = "delete"
 		} else if !apiequality.Semantic.DeepEqual(existing.Spec.Hard, newQuotaSpec.Hard) {
 			operation = "update"
@@ -167,6 +167,12 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
 		}
 		operationErr = c.updateResourceQuota(existing, newQuotaSpec)
 	case "delete":
+		updatedNs := ns.DeepCopy()
+		delete(updatedNs.Annotations, resourceQuotaAnnotation)
+		updatedNs, err = c.Namespaces.Update(updatedNs)
+		if err != nil {
+			return updatedNs, err
+		}
 		operationErr = c.deleteResourceQuota(existing)
 	}
 
@@ -438,7 +444,10 @@ func completeQuota(requestedQuota *v32.ResourceQuotaLimit, defaultQuota *v32.Res
 		return nil, err
 	}
 	for key, value := range requestedQuotaMap {
-		newLimitMap[key] = value
+		// Only override the values for keys (resources) that actually exist in the project quota.
+		if newLimitMap[key] != nil {
+			newLimitMap[key] = value
+		}
 	}
 
 	toReturn := &v32.ResourceQuotaLimit{}
--- a/tests/v2/integration/resource_quota_test.go
+++ b/tests/v2/integration/resource_quota_test.go
@@ -1,0 +1,268 @@
+package integration
+
+import (
+	"testing"
+	"time"
+
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+	"github.com/rancher/rancher/tests/framework/extensions/namespaces"
+	"github.com/rancher/rancher/tests/framework/extensions/resourcequotas"
+	"github.com/rancher/rancher/tests/framework/pkg/session"
+	"github.com/stretchr/testify/suite"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ResourceQuotaSuite struct {
+	suite.Suite
+	client  *rancher.Client
+	session *session.Session
+}
+
+func (s *ResourceQuotaSuite) TearDownSuite() {
+	s.session.Cleanup()
+}
+
+func (s *ResourceQuotaSuite) SetupSuite() {
+	testSession := session.NewSession(s.T())
+	s.session = testSession
+
+	client, err := rancher.NewClient("", testSession)
+	s.Require().NoError(err)
+	s.client = client
+}
+
+func (s *ResourceQuotaSuite) TestCreateNamespaceWithQuotaInProject() {
+	subSession := s.session.NewSession()
+	defer subSession.Cleanup()
+
+	client, err := s.client.WithSession(subSession)
+	s.Require().NoError(err)
+
+	projectLimit := &management.ResourceQuotaLimit{
+		LimitsCPU: "500m",
+	}
+	namespaceDefaultLimit := &management.ResourceQuotaLimit{
+		LimitsCPU: "200m",
+	}
+
+	projectConfig := &management.Project{
+		ClusterID: "local",
+		Name:      "TestProject",
+		ResourceQuota: &management.ProjectResourceQuota{
+			Limit: projectLimit,
+		},
+		NamespaceDefaultResourceQuota: &management.NamespaceResourceQuota{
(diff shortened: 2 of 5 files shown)


pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -438,7 +438,10 @@ func completeQuota(requestedQuota *v32.ResourceQuotaLimit, defaultQuota *v32.Res
 		return nil, err
 	}
 	for key, value := range requestedQuotaMap {
-		newLimitMap[key] = value
+		// Only override the values for keys (resources) that actually exist in the project quota.
+		if newLimitMap[key] != nil {
```
KevinJoiner, line 442:
```suggestion
		if _, ok := newLimitMap[key]; ok {
```

KevinJoiner (changes requested):

tests/v2/integration/resource_quota_test.go
```diff
@@ -0,0 +1,284 @@
...
+func (s *ResourceQuotaSuite) TearDownSuite() {
+	s.session.Cleanup()
+}
+
+func (s *ResourceQuotaSuite) SetupSuite() {
+	testSession := session.NewSession(s.T())
+	s.session = testSession
+
+	client, err := rancher.NewClient("", testSession)
+	require.NoError(s.T(), err)
```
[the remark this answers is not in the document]
maxsokolovsky: Will adjust, but wondering if there is any functional or semantic difference?

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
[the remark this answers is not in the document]
maxsokolovsky: This won't work because the string keys are always present in the map (created by converting the ResourceQuotaLimit value), but the values may or may not be nil. This would change the logic.

tests/v2/integration/resource_quota_test.go
```diff
@@ -0,0 +1,284 @@
...
+	require.NotNil(s.T(), quotas)
+	require.Lenf(s.T(), quotas.Items, 1, "Expected 1 quota in a new namespace, but got %d", len(quotas.Items))
+
+	limit := &v3.ResourceQuotaLimit{
+		LimitsCPU: "200m",
+	}
+
+	resourceList := quotas.Items[0].Spec.Hard
+	rl, err := v3quotas.ConvertProjectResourceLimitToResourceList(limit)
```
cmurphy, line 79: Instead of changing this function to exported just for use in this test, is it possible to create a native resource list object to compare to rather than converting a v3 object?

tests/v2/integration/resource_quota_test.go
```diff
@@ -0,0 +1,284 @@
...
+			Limit: &management.ResourceQuotaLimit{
+				LimitsCPU: "200m",
+			},
+		},
+	}
+	testProject, err := client.Management.Project.Create(projectConfig)
+	require.NoError(s.T(), err)
+
+	annotations1 := map[string]string{
+		"field.cattle.io/resourceQuota": "{\"limit\":{\"limitsCpu\":\"200m\"}}",
```
cmurphy, line 109: If it's the same as the default, is it really a good test of the override?

tests/v2/integration/resource_quota_test.go
```diff
@@ -0,0 +1,284 @@
...
+	testProject.NamespaceDefaultResourceQuota.Limit.ConfigMaps = ""
+
+	_, err = client.Management.Project.Replace(testProject)
+	require.NoError(s.T(), err)
+
+	time.Sleep(2 * time.Second)
+	quotas, err = resourcequotas.ListResourceQuotas(client, "local", "ns1", metav1.ListOptions{})
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), quotas)
+	require.Lenf(s.T(), quotas.Items, 1, "Expected 1 quota in the namespace, but got %d", len(quotas.Items))
```
cmurphy, line 219: This isn't what I expected. Why aren't we triggering https://github.com/rancher/rancher/blob/acb99e9d0aefb6fcca70963beba3423b8898a49d/pkg/controllers/managementuser/resourcequota/resource_quota_sync.go#L207-L210 ?

tests/v2/integration/resource_quota_test.go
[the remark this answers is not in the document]
KevinJoiner: Currently no, but testify suites like to do some manipulation with the underlying `*testting.T`. Which is why I prefer to call `Require()` instead, but you don't have to change I meant to mark this as a nit.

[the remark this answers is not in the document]
maxsokolovsky: Adjusted all places where an assertion is made.

tests/v2/integration/resource_quota_test.go
maxsokolovsky: For some reason, there is still one nil resource quota, investigating...

maxsokolovsky: I noticed that one can create a regular K8s resource quota without specifying any spec and thus with no limits or requests.
The API server allows it and creates an empty quota with a given name.
What happens here is similar. Perhaps we should delete a quota object if its spec is empty?

tests/v2/integration/resource_quota_test.go
maxsokolovsky: Agreed, fixed!

tests/v2/integration/resource_quota_test.go
cmurphy: > Perhaps we should delete a quota object if its spec is empty?

That's what I was getting at. Why isn't this happening? If it's empty, what could a user possibly need from it?

maxsokolovsky: This is old behavior, by the way. I will see how to delete the quota if its resource list is empty.

maxsokolovsky: OK, an empty quota remains because the annotation is always updated on the namespace. In cases like these, it looks like this: `field.cattle.io/resourceQuota: '{"limit":{}}'`.
Should we just delete the annotation together with the quota object? Is the namespace in a valid state then?

maxsokolovsky: Made an update. If we delete the quota but keep the annotation, then there they are out of sync, so I delete both together. Tested this by adding and removing different resources from the project and seeing how the annotation and quota get updated.

KevinJoiner (dismissed):

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
```diff
@@ -167,6 +167,9 @@ func (c *SyncController) CreateResourceQuota(ns *corev1.Namespace) (runtime.Obje
 		}
 		operationErr = c.updateResourceQuota(existing, newQuotaSpec)
 	case "delete":
+		// To allow annotation deletion.
+		namespaceutil.SetNamespaceCondition(ns, time.Second*1, ResourceQuotaInitCondition, false, "")
```
cmurphy, line 171: It's a little hard to read what this is doing since at first glance the ResourceQuotaInitCondition doesn't have anything to do with deleting and this condition won't ever actually be set on the object in etcd. It would be clearer in my opinion to just go ahead and update the namespace object here, the same way e.g. `validateAndSetNamespaceQuota` calls update. That might be just a personal preference though.

cmurphy (dismissed):

pkg/controllers/managementuser/resourcequota/resource_quota_sync.go
maxsokolovsky: Good point, adjusted!

KevinJoiner (approved):

cmurphy (approved):

PR #39282, opened 2022-10-12
[v2.6] update versions for v2.6.9-rc4

kinarashah:


rancherbot (approved):

Issue #39281, opened 2022-10-12
[BUG] Rancher fails to install on 22.04 Ubuntu with error - `mkdir: cannot create directory '/sys/fs/cgroup/init': Read-only file system`


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head `f10a449b6` or 2.7-head 
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1 using - 1.3.15 or 1.4.0-rc3

**Describe the bug**
Rancher fails to install on 22.04 Ubuntu with error - `mkdir: cannot create directory '/sys/fs/cgroup/init': Read-only file system`

**To Reproduce**

- Deploy RKE1 -3 nodes all roles cluster using ubuntu 22.04 AMI
- Follow rancher docs to install cert manager and Rancher.
- Rancher is in a crashloop
- Error seen on rancher pods: `mkdir: cannot create directory '/sys/fs/cgroup/init': Read-only file system`

**Expected Result**
Rancher should come up successfully.

**Workaround**
- On the instance ran these commands
```
# echo 'GRUB_CMDLINE_LINUX=systemd.unified_cgroup_hierarchy=false' > /etc/default/grub.d/cgroup.cfg
# update-grub
```
- Rebooted the node, before creating a Private AMI from this to then use our automation scripts to install Rancher.
- And rancher 2.6-head provisioned successfully.

kinarashah:
Issue to fix in upstream hasn't been closed yet - https://github.com/moby/moby/issues/42040

https://github.com/moby/moby/issues/42275

kinarashah:
Available to test after https://drone-publish.rancher.io/rancher/rancher/8426 passes

sowmyav27:
Tests using Ubuntu 22.04 AMI:
- [ ] Rancher on RKE1 local cluster with Ubuntu 22.04 AMI with no workaround --> Rancher should come up fine
- [ ] Docker install -- no regression
- [ ] Rancher HA on k3s 1.24 --> Rancher should come up fine

sowmyav27:
**On 2.6-head commit id: `6fabf75`**

- [ ] Rancher on RKE1 local cluster with Ubuntu 22.04 AMI with no workaround --> Rancher should come up fine
- [ ] Docker install -- no regression
- [ ] Rancher HA on k3s 1.24 --> Rancher should come up fine

PR #39286, opened 2022-10-13
[v2.6] enable v2 cgroup nesting only for rancher standalone container

kinarashah:
https://github.com/rancher/rancher/issues/39281


brandond (approved):

brandond:
You could probably also check for the Kubernetes service address environment variable; I'm not sure if that's more or less guaranteed to exist than a service account token.

kinarashah:
@brandond Ya I could use the env var, going with service account to be consistent with our other check (https://github.com/rancher/rancher/blob/release/v2.6/package/entrypoint.sh#L5). 1.24 might have been a concern but both service var and serviceaccount are available in the pod to access the API.

jakefhyde (approved):

PR #39309, opened 2022-10-14
[forwardport] Allow TF to ssh into 22.04 nodes

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/505
 
## Problem

rke2 clusters in jenkins fail to create when using ubuntu 22.04 but not other (older) versions. 
upon further investigation, this was due to an incompatibility with TF when sshing into 22.04 node(s)
 
## Solution

created an elliptic-based ssh key in aws to use. 
 
## Testing

tested in jenkins (see thread) and locally

this PR updates _ALL_ jenkins ssh keys to use the new key. This will require an update in jenkins and all engineers using jenkins to acquire the new key.


markusewalker (dismissed):
LGTM, thanks @slickwarren. Since this will break workflows, we should hold off on merging until Monday to not cause weekend panic 😄

Josh-Diamond (commented):
LGTM ✅ .. Thanks @slickwarren

Josh-Diamond (approved):
oops. i just left a comment above ^^ 😄 Here's approval

igomez06 (dismissed):
LGTM

jameson-mcghee:
LGTM 👍

markusewalker (approved):

PR #39310, opened 2022-10-15
Fix Rancher crash loop in case if Keycloak URL doesn't have /auth/

jamhed:
## Problem
Rancher assumes that Keycloak issuer URL contains `/auth/`
 
## Solution
Handle other cases.

### DIFF
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -171,7 +171,7 @@ func getSearchURL(issuer string) (string, error) {
 	}
 	iss = strings.SplitN(issuer, "/realms/", 2) // keycloak >= 19 doesn't have auth prefix
 	if len(iss) == 2 {
-		return fmt.Sprintf("%s/admin/realms/%s", iss[0], strings.Join(iss[1:], "")), nil
+		return fmt.Sprintf("%s/admin/realms/%s", iss[0], iss[1]), nil
 	}
 	return "", fmt.Errorf("can't parse issuer url")
 }
--- a/pkg/auth/providers/keycloakoidc/keycloak_client.go
+++ b/pkg/auth/providers/keycloakoidc/keycloak_client.go
@@ -165,12 +165,15 @@ func (k *KeyCloakClient) getFromKeyCloakByID(principalID, principalType string, 
 }
 
 func getSearchURL(issuer string) (string, error) {
-	splitIssuer := strings.SplitAfter(issuer, "/auth/")
-	return fmt.Sprintf(
-		"%sadmin/%s",
-		splitIssuer[0],
-		splitIssuer[1],
-	), nil
+	iss := strings.SplitAfter(issuer, "/auth/") // keycloak < 19 has auth prefix
+	if len(iss) == 2 {
+		return fmt.Sprintf("%sadmin/%s", iss[0], iss[1]), nil
+	}
+	iss = strings.SplitN(issuer, "/realms/", 2) // keycloak >= 19 doesn't have auth prefix
+	if len(iss) == 2 {
+		return fmt.Sprintf("%s/admin/realms/%s", iss[0], strings.Join(iss[1:], "")), nil
+	}
+	return "", fmt.Errorf("can't parse issuer url")
 }
 
 // URLEncoded encodes the string


bpedersen2 (commented):
Looks like  a wrorking workaround.
But for a longterm solution probably the config option handling should be changed to give the  
path to `.well-known/openid-configuration` and read the  required endpoints from there.

michael81877:
Since keycloak >= v19 won't work with rancher without this change or without re-adding `/auth` via keycloak's `http-relative-path` config, would it be possible to include this change in a rancher release?

Changing keycloak's `http-relative-path` means you need to change every keycloak client you have configured.

michaelbeutler:
+1

hjones2199:
I suspect this getting merged is fairly critical to quite a few users

pkg/auth/providers/keycloakoidc/keycloak_client.go
```diff
@@ -165,12 +165,15 @@ func (k *KeyCloakClient) getFromKeyCloakByID(principalID, principalType string,
...
-		splitIssuer[0],
-		splitIssuer[1],
-	), nil
+	iss := strings.SplitAfter(issuer, "/auth/") // keycloak < 19 has auth prefix
+	if len(iss) == 2 {
+		return fmt.Sprintf("%sadmin/%s", iss[0], iss[1]), nil
+	}
+	iss = strings.SplitN(issuer, "/realms/", 2) // keycloak >= 19 doesn't have auth prefix
+	if len(iss) == 2 {
+		return fmt.Sprintf("%s/admin/realms/%s", iss[0], strings.Join(iss[1:], "")), nil
```
KevinJoiner, line 174: if `len(iss) == 2` then we don't need `strings.Join(...)`
```suggestion
		return fmt.Sprintf("%s/admin/realms/%s", iss[0], iss[1]), nil
```

KevinJoiner (changes requested):

pkg/auth/providers/keycloakoidc/keycloak_client.go
Roman-Galeev: @KevinJoiner, done.

KevinJoiner (approved):

smolinari:
:+1:  LGTM 

We have this issue too. Switched over to Keycloak SSO authentication in Rancher and Rancher breaks badly with continuous restarts. It causes Rancher to be basically useless. Please merge!

Scott

Roman-Galeev:
> Looks like a wrorking workaround. But for a longterm solution probably the config option handling should be changed to give the path to `.well-known/openid-configuration` and read the required endpoints from there.

@bpedersen2 , is there an issue for that?

jely2002:
Looks like this will fix the issue we're having. Enabling SSO with keycloak OIDC keeps restarting Rancher, making it unusable.

Would be great if this workaround can be merged, as this is critical for us.

moonlsd:
+1 here

Just by configuring OIDC Keycloak against Keycloak 20 in Rancher makes it crash forever - which forced us to manually deal with user provisioning

smolinari:
Hate to rant.... but is anyone working on this repository in a dedicated or serious manner? I mean over 2k issues and almost 100 PRs open. This issue is a fairly significant with no sign of being slated for any future versions or that it has any priority at all. 

Yeah, I see it was approved and someone else self-assigned to also approve, but I think you guys and gals on the Rancher/ Suse team can do better for sure, right? :thinking: 

Scott

MbolotSuse (approved):
LGTM - appears to work for both the Quarkus/Wildfly distributions for basic user/group functionality.

MbolotSuse:
@jamhed Thanks for the PR. Generally, when we finish issues we fill out a template which helps our QA to know what to test. If you could try to fill that out, that would be helpful. I'll talk internally to see what the timeline looks like for getting the code merged/put in a release, and will take care of getting it put in from here. 

## Validation Template

## Root Cause

## What was fixed, or what change have occurred

## Areas or cases that should be tested

## What areas could experience regressions

## Are the repro steps accurate/minimal?

Roman-Galeev:
@MbolotSuse, let me try:

Root cause:
the code in question assumed that keycloak URL is always structured in a certain way ($domain/auth/$realm), and caused crashes due to unhandled error cases.

What was fixed, or what change have occurred:
I've added code to handle both old and new keycloak URL schemas, and error reporting without crashes in case url is different.

Areas or cases that should be tested:
Configuring Keycloak OIDC to work with Rancher.

What areas could experience regressions:
None

Are the repro steps accurate/minimal?
Configure Rancher to work with Keycloak versions < 19 and > 19.

As a side note, I find it hilarious comparing the amount of text I need to write to explain what I did with the actual change.

bpedersen2:
Seems it did not make it into 2.7.1

Roman-Galeev:
@bpedersen2, that's not good, as it seems it affects a lot of people (me included).

KevinJoiner:
@bpedersen2/@Roman-Galeev
v2.7.1 contains security-related issues only.
https://www.suse.com/c/rancher_blog/whats-new-in-ranchers-security-release-only-versions/

PR #39318, opened 2022-10-17
Bump rke docker defaultEngineInstallURL to 20.10

annablender:
Issue: https://github.com/rancher/terraform-provider-rancher2/issues/969

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -45,7 +45,7 @@ import (
 )
 
 const (
-	defaultEngineInstallURL         = "https://releases.rancher.com/install-docker/17.03.2.sh"
+	defaultEngineInstallURL         = "https://releases.rancher.com/install-docker/20.10.sh"
 	amazonec2                       = "amazonec2"
 	userNodeRemoveCleanupAnnotation = "cleanup.cattle.io/user-node-remove"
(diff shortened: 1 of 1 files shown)


jakefhyde (approved):

pkg/controllers/management/node/controller.go
```diff
@@ -45,7 +45,7 @@ import (
 )
 
 const (
-	defaultEngineInstallURL         = "https://releases.rancher.com/install-docker/17.03.2.sh"
```
HarrisonWAffel, line 48: It should be noted that this will only the default docker version of RKE-1 clusters if the docker version has not been specified in the node template. The UI currently defaults to `https://releases.rancher.com/install-docker/20.10.sh`, so this change looks okay to me. We should consider updating this more often, as version `17.03` is no longer within the support matrix for recent rancher releases.

HarrisonWAffel (approved):

pkg/controllers/management/node/controller.go
annablender: Correct, if `install_engine_url` is specified via tf then rancher will use that docker version. I'm assuming it's easier to specify 20.10.x in the [support matrix](https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-6-9/) for ubuntu, but it would be 20.10.17 right now

Issue #38177, opened 2022-07-03
Rancher 2.6.5/2.6.6 liveness crashes the pod and restarts infinitely


Shaked:
**Rancher Server Setup**
- Rancher version: 2.6.5 and 2.6.6
- Installation option (Docker install/Helm Chart):
 Currently `rancher-2.6.6                	v2.6.6` but also tried with `2.6.5` before. Basically following the rancher/latest repo.

**Information about the Cluster**
- Kubernetes version: `v1.20.15+k3s1`
- Cluster Type (Local/Downstream): Local

**User Information**
- What is the role of the user logged in? Admin
   
**Describe the bug**
Rancher tries to start and fails. First I kept seeing Exit Code 137 (NOT OOM). After that, I changed deploy/rancher `livenessProbe` to 

```
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /healthz
            port: 80
            scheme: HTTP
          initialDelaySeconds: 120
          periodSeconds: 30
          successThreshold: 1
          timeoutSeconds: 5
```
The first time failed for: 
```
2022/07/03 10:16:11 [ERROR] failed to start cluster controllers c-kmtbm: context canceled
E0703 10:16:11.471980      32 leaderelection.go:330] error retrieving resource lock kube-system/cattle-controllers: Get "https://10.0.188.63:443/api/v1/namespaces/kube-system/configmaps/cattle-controllers?timeout=15m0s": context canceled
I0703 10:16:11.472029      32 leaderelection.go:283] failed to renew lease kube-system/cattle-controllers: timed out waiting for the condition
E0703 10:16:11.472219      32 leaderelection.go:306] Failed to release lock: resource name may not be empty
I0703 10:16:11.473005      32 trace.go:205] Trace[367625047]: "Reflector ListAndWatch" name:pkg/mod/github.com/rancher/client-go@v1.23.3-rancher2/tools/cache/reflector.go:168 (03-Jul-2022 10:15:15.544) (total time: 55928ms):
Trace[367625047]: [55.928576086s] [55.928576086s] END
2022/07/03 10:16:11 [ERROR] failed to start cluster controllers c-sqmd2: context canceled
2022/07/03 10:16:11 [INFO] Shutting down management.cattle.io/v3, Kind=RoleTemplate workers
2022/07/03 10:16:11 [INFO] Shutting down provisioning.cattle.io/v1, Kind=Cluster workers
[...]

PR #39319, opened 2022-10-17
Add option to override initialDelay and period

steffantucker:
Adds values to the chart for rancher server to override the livenessProbe and readinessProbe's initialDelaySeconds and periodSeconds

## Issue: 
https://github.com/rancher/rancher/issues/38177

 
## Problem

In some situations the liveness probe does not report within the default time, this makes it easier to expand that window.
 
## Solution

Added values to the helm chart that override the default values

### Automated Testing

Added helm tests

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -147,14 +147,30 @@ spec:
           httpGet:
             path: /healthz
             port: 80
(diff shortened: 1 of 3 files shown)


cmurphy (approved):

maxsokolovsky (approved):
LGTM. This can't be merged just yet in `release/v2.7`, right?

PR #39331, opened 2022-10-17
[v2.6] Support RANCHER_HA_HOSTNAME and RANCHER_HA_KUBECONFIG to be set with RANCHER_HA_HARDENED when set to True

markusewalker:
## Problem

When running the `test_install_rancher_ha` function from `test_create_ha.py`, there is a failure when you attempt to have environment variables `RANCHER_HA_HOSTNAME` and `RANCHER_HA_KUBECONFIG` set and `RANCHER_HA_HARDENED` is set to `True`. To get around this, a workaround is to set `RANCHER_HA_HARDENED` to `False` when these variables are present.

This should not happen and disrupts automation when trying to import an RKE1 cluster with the existing automation.
 
## Solution

The variable `rke-cis-1.5` needed to be declared outside of its parental conditionals. This is done so that the variable is always set and not set only in a conditional.

Additionally, there needs to be a new check in the `set_route53_with_ingress` function that looks for `RANCHER_HA_HARDENED.upper() == "TRUE" and RANCHER_LOCAL_CLUSTER_TYPE == "RKE"`. In doing that, when supplying your own RKE1 cluster and EC2 load balancer, Rancher stands up properly with a hardened local RKE1 cluster.
 
### Manual Testing

Validated with manual testing that I do not receive the previous error.

### Automated Testing

### DIFF
--- a/tests/validation/tests/v3_api/test_create_ha.py
+++ b/tests/validation/tests/v3_api/test_create_ha.py
@@ -12,7 +12,7 @@ from .test_rke_cluster_provisioning import AZURE_CLIENT_ID, AZURE_CLIENT_SECRET,
 # RANCHER_HA_KUBECONFIG and RANCHER_HA_HOSTNAME are provided
 # when installing Rancher into a k3s setup
 RANCHER_HA_KUBECONFIG = os.environ.get("RANCHER_HA_KUBECONFIG")
-RANCHER_HA_HARDENED = ast.literal_eval(os.environ.get("RANCHER_HA_HARDENED", "False"))
+RANCHER_HA_HARDENED = os.environ.get("RANCHER_HA_HARDENED", "False")
 RANCHER_HA_HOSTNAME = os.environ.get(
     "RANCHER_HA_HOSTNAME", RANCHER_HOSTNAME_PREFIX + ".qa.rancher.space")
 resource_prefix = RANCHER_HA_HOSTNAME.split(".qa.rancher.space")[0]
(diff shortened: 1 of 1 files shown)


izaac (approved):
LGTM

tests/validation/tests/v3_api/test_create_ha.py
```diff
@@ -140,7 +140,7 @@ def valid_response():
         assert False, "check the logs in console for details"
 
     print_kubeconfig(kubeconfig_path)
-    if RANCHER_HA_HARDENED and RANCHER_LOCAL_CLUSTER_TYPE == "RKE":
+    if (RANCHER_HA_HARDENED.upper() == "TRUE" and RANCHER_LOCAL_CLUSTER_TYPE == "RKE") or (RANCHER_HA_HARDENED.upper() == "TRUE" and RANCHER_LOCAL_CLUSTER_TYPE == "K3S"):
```
sowmyav27, line 143: The steps in prepare_hardened_cluster after that, doesn't apply for k3s cluster. It applies only for RKE1 cluster.

And the condition that should be here is
```
if (RANCHER_HA_HARDENED.upper() == "TRUE" and RANCHER_LOCAL_CLUSTER_TYPE == "RKE") and RANCHER_HA_KUBECONFIG == ""(<blank>) and RANCHER_HA_HOSTNAME=="" (<blank>)
```
Since the hardening of cluster should happen only when harden flag is set, and the local cluster is RKE1, and when user doesn't input values for kubeconfig and hostname. As the default in this job is always
- RANCHER_HA_HARDENED = True
- RANCHER_LOCAL_CLUSTER_TYPE = RKE1

sowmyav27 (changes requested):

sowmyav27 (approved):

izaac (approved):
LGTM

slickwarren (approved):

Issue #39239, opened 2022-10-05
Supporting K8s v1.23 -v1.24


Sahota1225:
We will be supporting 1.23 and 1.24 for 2.7.0. 

- [x]  RKE1 (and standalone) - Make changes in KDM 
- [x] RKE2/k3s - changes in KDM 
- [x] Hosted providers - EKS, AKS, GKE - Make changes accordingly (UI ?) - kev1 vs kev2
- [x] Imported clusters - k3s/RKE2

Josh-Diamond:
Team 2 has reviewed this ticket;
Multiple assignees;
Leaving unassigned for now;

sowmyav27:
**On latest 2.7.0-rc**

### RKE1 (and standalone) 
- [ ] Verify the k8s versions available for standalone `1.4.0-rcx` RKE1 are _only_ `v1.24.6-rancher1-1` and `v1.23.10-rancher1-1`

### RKE2/k3s
- [ ] RKE2 - Verify when provisioning RKE2 clusters from Rancher, the versions available should be _only_  `1.24.6+rke2r1` and `1.23.12+rke2r1`
- [ ] Provisioning a `1.24.6+rke2r1` custom cluster and verify cluster comes up successfully
- [ ] Provisioning a `1.23.6+rke2r1` AWS Node driver cluster and verify cluster comes up successfully
- [ ] K3S - Verify when provisioning k3s clusters from Rancher, the versions available should be _only_  `1.24.6+k3s1` and `1.23.12+ k3s1`
- [ ] Provisioning a `1.24.6+k3s1` custom cluster and verify cluster comes up successfully
- [ ] Provisioning a `1.23.6+k3s1` AWS Node driver cluster and verify cluster comes up successfully

### Imported clusters - k3s/RKE2
- [ ] Import a `1.21.14+k3s1` cluster into rancher. Verify user has only `1.24.6+k3s1` and `1.23.12+ k3s1` versions to upgrade to.
- [ ] Import a `1.21.14+rke2r1` cluster into rancher. Verify user has only 1.24.6+rke2r1` and `1.23.12+rke2r1` versions to upgrade to.

### Hosted providers - EKS, AKS, GKE 
- [ ] Provision an EKS kev2 cluster. Verify k8s version only 1.23+ are available on UI. Verify user is able to provision 1.23 cluster successfully.
[...]

PR #39346, opened 2022-10-18
[2.7] bump the RancherVersionDev to 2.7.99 

jiaqiluo:
## Issue: 
 
https://github.com/rancher/rancher/issues/39239
## Problem

For RKE, RKE1, and K3s, the available versions in the version list are not right when creating a node-drive cluster on **v2.7-head**

## Solution

 
- bump the RancherVersionDev to 2.7.99, so that the available k8s versions can be calculated correctly on 2.7-head and the dev build
- fix the server version for rke2/k3s channel server
- fix tests

### Manual Testing

Built and ran rancher locally and see only the latest  1.23 and 1.24 as the available version for For RKE, RKE1, and K3s

### DIFF
--- a/tests/integration/suite/test_kontainer_drivers.py
+++ b/tests/integration/suite/test_kontainer_drivers.py
@@ -289,7 +289,7 @@ def test_kontainer_driver_links(admin_mc):
     images = get_images(url, token)
     assert "hyperkube" in images
     assert "rke-tools" in images
-    assert "kubelet-pause" in images
+    assert "mirrored-pause" in images
 
 
 def get_images(url, token):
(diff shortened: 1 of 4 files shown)


tests/integration/suite/test_kontainer_drivers.py
```diff
@@ -289,7 +289,7 @@ def test_kontainer_driver_links(admin_mc):
     images = get_images(url, token)
     assert "hyperkube" in images
     assert "rke-tools" in images
-    assert "kubelet-pause" in images
+    assert "mirrored-pause" in images
```
jiaqiluo, line 292: Note to Reviewers:
  1.23 and 1.24 uses `rancher/mirrored-pause` instead of  `rancher/kubelet-pause`

HarrisonWAffel (approved):

jakefhyde (approved):

PR #39351, opened 2022-10-18
[v2.6.10] Add sync back to check for user-node-remove, add tests

jakefhyde:
## Problem

If a node is provisioned on <v2.5.1, rancher is stopped, and the node is deleted; Remove will not be called for the node causing orphaned v1.Nodes which default unmanaged v3.Nodes are then created for.

## Solution

Remove the annotation on sync and v3.Node removal.

### Manual Testing

Tested with a 1 etcd, 1 controlplane, 1 worker. On upgrade from v2.5.1 to v2.6-head, the node is properly cleaned up.

### DIFF
--- a/pkg/controllers/management/node/controller.go
+++ b/pkg/controllers/management/node/controller.go
@@ -6,7 +6,6 @@ import (
 	"encoding/base32"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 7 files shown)


pkg/controllers/management/node/controller.go
```diff
@@ -46,9 +46,10 @@ import (
 )
 
 const (
-	defaultEngineInstallURL       = "https://releases.rancher.com/install-docker/17.03.2.sh"
-	amazonec2                     = "amazonec2"
-	userNodeRemoveFinalizerPrefix = "clusterscoped.controller.cattle.io/user-node-remove_"
+	defaultEngineInstallURL         = "https://releases.rancher.com/install-docker/17.03.2.sh"
```
annablender, line 48: The default version of docker being used for RKE is now 20.10. Not sure if we need to bump this in 2.6 but if we need to I'll backport my PR https://github.com/rancher/rancher/pull/39318.

annablender (approved):
LGTM! Rubber-stamped

HarrisonWAffel (approved):

PR #39360, opened 2022-10-19
[v2.7] Update Ubuntu 18.04 to Ubuntu 20.04

markusewalker:
## Problem

Internal automation is currently using Ubuntu 18.04 as default AMIs. While Ubuntu 18.04 is still actively supported and has not reached EOL, we should update the default AMI to use Ubuntu 20.04.
 
## Solution

Updated the `aws.py` to use an Ubuntu 20.04 AMI that has been validated to successfully work.
 
### Automated Testing

Ran internal automation with updated AMI and each of the tests were successful.


izaac (approved):
LGTM

sowmyav27 (approved):

PR #39375, opened 2022-10-21
add debug mode for rancher machine provisioning with rke2

tuxtof:
## Problem

in rke2 provisioning mode there is no way to enable debug mode for machine driver.

 
## Solution

This PR fix the the problem by checking if global rancher log level is trace and if this is the case set --debug when launching machine provisioning pod

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
@@ -24,6 +24,7 @@ import (
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/kv"
 	name2 "github.com/rancher/wrangler/pkg/name"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


tuxtof:
fix  #39377

snasovich (approved):
LGTM, matches the approach used for RKE1 - https://github.com/rancher/rancher/blob/40932d62cdeb2f4594009436468c444c3378553d/pkg/controllers/management/node/utils.go#L136-L140

sbstp (approved):

Oats87:
@snasovich I do not believe we should be emitting credentials into pods in the `fleet-default` namespace when Rancher trace logs are enabled.

The concern I have is around log aggregation expectations -- if a user is aggregating logs from their local cluster, they may not expect extremely sensitive data to be emitted from pods in a differing namespace from the pod/deployment they are enabling trace logs on (cattle-system vs fleet-default).

I believe if we want to support this type of behavior, we should introduce a specific flag/environment variable that can enable trace logging in the specific machine provision pods, which makes enabling debug logs (and thus, emitting sensitive information in the fleet-default namespace) more explicit than this.

PR #39410, opened 2022-10-25
parameterizing access key name in TF; allowing jenkins global var to be overridden.

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/505
 
## Problem

when updating the key, there were some jobs that still require RSA (windows, specifically).  This was overlooked in the first PR. As a result, I have debugged overriding the key, and this is the resulting PR
 
## Solution

* allow TF access key name to be overridden -> added new TF variable, added respective TF variable to appropriate TF calls
* allow Jenkins global vars to be overridden -> fixed order of env/credential read in jenkinsfile
 
## Testing

local -> Passing
remote -> TBD (see slack)

## QA Testing Considerations

This should be the last of what is necessary to change the key. 
 
This parameterization of the name should help prevent regressions in the future.


mdrahman-suse (dismissed):
LGTM

rancher-max (approved):

Issue #36526, opened 2022-02-15
Cluster provisioning stuck on waiting for API to be available when using RHEL 8.4 AMI on a Node driver RKE1 cluster


timhaneunsoo:
**Rancher Server Setup**
- Rancher version: v2.6-head 11a7451
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): rke1

**Information about the Cluster**
- Kubernetes version: 1.21.8
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE1 EC2 AWS

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Standard User

   
**Describe the bug**

User is trying to deploy RKE1 aws node driver cluster using a public RHEL 8.4 ami. Cluster seems to be stuck on `Waiting for API to be available`
<img width="565" alt="Screen Shot 2022-02-15 at 2 05 44 PM" src="https://user-images.githubusercontent.com/97477402/154138017-36e8c2c4-ecf3-48a4-a0f7-8c547bf180a1.png">
This issue occurred on 2.6.3 v1.21/v1.22 and 2.6-head v1.21/v1.22

**To Reproduce**
 
1. Deploy RKE1 aws node driver cluster using a public RHEL 8.4 ami
2. Check status of deployment

**Result**
Deployment status is stuck on "Waiting" with the message "Waiting for API to be available" below.

**Expected Result**

Deployment is successfully deployed.

**Screenshots**

**Additional context**

The same AMI wasn’t able to finish running the docker install script when attempting to install manually

The kube-apiserver log gives a 503 error
![image](https://user-images.githubusercontent.com/97477402/154138709-5314b3ff-887f-48b9-b383-face3ced622c.png)

The ports seem to be listening, as expected
![image](https://user-images.githubusercontent.com/97477402/154138824-29a0329f-004c-4a67-aa69-e502166c49cf.png)

The cluster agent log on the control plane node gives the following message on the last line
`time="2022-02-15T15:59:51Z" level=info msg="Removing unmanaged agent /elegant_shannon(address)"`

The release version is correct
[...]

PR #39420, opened 2022-10-26
[v2.7] Add Configuration Hooks for ProvV2

HarrisonWAffel:
## Problem
When provisioning nodes using the RHEL operating system the node will fail to fully provision due to two NetworkManager services interfering with the CNI's configuration. This results in the node indefinitely staying in the `Waiting for probes` stage (or some other similar failure condition). This issue was fixed for RKE1 clusters via change in `rancher-machine` (see https://github.com/rancher/rancher/issues/36526). However, this solution is not applicable for ProvV2 as all configuration is done via cloud-init. 
 
## Solution
Allow the user to specify predefined configuration _hooks_ within the yaml of a machine pool. These hooks denote particular commands that will be added to the install script passed to the downstream node. Currently, cloud-init is used to install the rancher system-agent for ProvV2. This PR adds additional commands after the installation of the agent is complete. 

The contents of each hook are defined within Rancher source code, ensuring that arbitrary commands cannot be injected within the cloud-init script. The included hook performs the same configuration step that is performed when provisioning RKE1 clusters using RHEL, which allows the default RHEL AMI to be used without any prior configuration. More hooks can be added as needed, so other pre-configuration issues with particular OS's can be automatically resolved. 

When building the install script, Rancher does not know what particular OS is being used, so these hooks must be manually configured. This differs from the behavior of RKE1 provisioning, however from what I am able to tell it cannot be avoided. 

To use these new hooks the user must modify the YAML of a particular machine pool to denote which hooks should be applied to the install script. For example, the following YAML will disable the `nm-cloud-setup.service` and `nm-cloud-setup.timer` using `systemctl`.

```yaml
machinePools:
    - controlPlaneRole: true
      etcdRole: true
      hooks:
      - disable-network-manager
      machineConfigRef:
        kind: <CONFIG_TYPE>
        name: <CONFIG_NAME>
      name: <POOL_NAME>
      quantity: 1
      unhealthyNodeTimeout: 0s
      workerRole: true
```
 
Specifying this hook will add the following logic to the install script, after the system-agent is installed.
```
sudo systemctl list-units --all | grep -Fq nm-cloud-setup.timer; if [ $? -eq 0 ]; then sudo systemctl disable nm-cloud-setup.timer; fi
sudo systemctl list-units --all | grep -Fq nm-cloud-setup.service; if [ $? -eq 0 ]; then sudo systemctl disable nm-cloud-setup.service; fi
```

Machines must be reprovisioned/recreated if this hook is added to the YAML after a machine has already been created in an infrastructure provider. 

## Testing
+ Begin to create an RKE2 cluster within the Rancher UI
+ Create three machine pools
   + Two of the pools will use a RHEL AMI. One pool should have the `etcd` and `control-plane` roles and the other should just have the `worker` role. 
   + One of the pools will use the default Ubuntu AMI. This pool should just have the `worker` role. 
+ After creating the three pools, click on the `Edit as YAML` button in the lower right hand corner
+ Add the `disable-network-manager` hook to the first pool configured with a RHEL AMI, do not define any hooks for the other two pools
+ Create the cluster after modifying the YAML 
+ Observe that the RHEL pool with the hook defined provisions and becomes available, and that the RHEL pool which does not have the hook defined does not come up or become available. Ensure that the Ubuntu pool comes up and becomes without the need to specify any hooks. 

### Manual Testing
I've done the above testing and have used the new hook to provision a new RHEL 8.6 cluster. I observed it come up and become useable. Machine pools within the same cluster which did not have this hook specified did not come up successfully, as expected. Machine pools which used OS's which did not need any pre-configuration were provisioned and became available as expected.

## QA Testing Considerations
This fix should address many RHEL versions which do not provision successfully by default, so testing should not be limited to just RHEL 8.6. 
 
None that I can think of.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
@@ -9,6 +9,7 @@ import (
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
+	provcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	rkecontroller "github.com/rancher/rancher/pkg/generated/controllers/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/provisioningv2/rke2/installer"
@@ -40,6 +41,7 @@ type handler struct {
 	capiClusters        capicontrollers.ClusterCache
 	deploymentCache     appcontrollers.DeploymentCache
 	rkeControlPlanes    rkecontroller.RKEControlPlaneCache
+	provClusters        provcontrollers.ClusterCache
 	rkeBootstrapClient  rkecontroller.RKEBootstrapClient
 }
 
@@ -52,8 +54,10 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 		capiClusters:        clients.CAPI.Cluster().Cache(),
 		deploymentCache:     clients.Apps.Deployment().Cache(),
 		rkeControlPlanes:    clients.RKE.RKEControlPlane().Cache(),
+		provClusters:        clients.Provisioning.Cluster().Cache(),
 		rkeBootstrapClient:  clients.RKE.RKEBootstrap(),
 	}
+
 	rkecontroller.RegisterRKEBootstrapGeneratingHandler(ctx,
 		clients.RKE.RKEBootstrap(),
 		clients.Apply.
@@ -62,6 +66,7 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 				clients.RBAC.RoleBinding(),
 				clients.CAPI.Machine(),
 				clients.Core.ServiceAccount(),
+				clients.Provisioning.Cluster(),
 				clients.Core.Secret()).
 			WithSetOwnerReference(true, true),
 		"",
@@ -131,7 +136,25 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 	if os := machine.GetLabels()[rke2.CattleOSLabel]; os == rke2.WindowsMachineOS {
 		is = installer.WindowsInstallScript
 	}
-	data, err := is(context.WithValue(context.Background(), tls.InternalAPI, hasHostPort), base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
+
+	// find the machine pool which the current machine belongs to
+	// so that we can utilize any provided hooks.
+	provCluster, err := h.provClusters.Get(namespace, machine.Spec.ClusterName)
+	if err != nil {
+		return nil, err
+	}
+	var hooks []string
+	if config := provCluster.Spec.RKEConfig; config != nil {
+		machinePoolName := machine.ObjectMeta.Labels[rke2.RKEMachinePoolNameLabel]
+		for _, pool := range config.MachinePools {
+			if pool.Name == machinePoolName {
+				hooks = pool.Hooks
+				break
+			}
+		}
+	}
+
+	data, err := is(context.WithValue(context.Background(), tls.InternalAPI, hasHostPort), base64.URLEncoding.EncodeToString(hash[:]), envVars, "", hooks)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
+++ b/pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
@@ -4,11 +4,15 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
+
+	provcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/provisioningv2/rke2/installer"
 	"testing"
 
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	"github.com/rancher/rancher/pkg/settings"
 
+	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	capicontrollers "github.com/rancher/rancher/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
 	"github.com/rancher/rancher/pkg/namespace"
 	v1wa "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
@@ -86,6 +90,22 @@ func (m *machineCacheMock) GetByIndex(indexName, key string) ([]*v1beta1.Machine
 	return args.Get(0).([]*v1beta1.Machine), args.Error(1)
 }
 
+type provisioningClusterCacheMock struct{ mock.Mock }
+
+func (m *provisioningClusterCacheMock) List(namespace string, selector labels.Selector) ([]*provv1.Cluster, error) {
+	return []*provv1.Cluster{}, nil
+}
+func (m *provisioningClusterCacheMock) AddIndexer(indexName string, indexer provcontrollers.ClusterIndexer) {
+	return
+}
+func (m *provisioningClusterCacheMock) GetByIndex(indexName, key string) ([]*provv1.Cluster, error) {
+	return []*provv1.Cluster{}, nil
+}
+func (m *provisioningClusterCacheMock) Get(namespace, name string) (*provv1.Cluster, error) {
+	args := m.Called(namespace, name)
+	return args.Get(0).(*provv1.Cluster), args.Error(1)
+}
+
 func Test_getBootstrapSecret(t *testing.T) {
 	type args struct {
 		secretName    string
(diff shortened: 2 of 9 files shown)


jakefhyde (changes requested):

jakefhyde (commented):
LGTM, pending review from @Oats87

jakefhyde (dismissed):

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -33,27 +35,30 @@ const (
...
-	rkeControlPlanes    rkecontroller.RKEControlPlaneCache
-	rkeBootstrapClient  rkecontroller.RKEBootstrapClient
+	serviceAccountCache     corecontrollers.ServiceAccountCache
+	secretCache             corecontrollers.SecretCache
+	secretClient            corecontrollers.SecretClient
+	machineCache            capicontrollers.MachineCache
+	capiClusters            capicontrollers.ClusterCache
+	deploymentCache         appcontrollers.DeploymentCache
+	rkeControlPlanes        rkecontroller.RKEControlPlaneCache
+	rkeProvisioningClusters v1.ClusterClient
```
jakefhyde, line 45:
```suggestion
	rkeClusters provcontrollers.ClusterClient
```

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -6,14 +6,16 @@ import (
 	"encoding/base64"
 	"errors"
 
+	v1 "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
```
jakefhyde, line 9:
```suggestion
	provcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
```

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
jakefhyde: Do we need the client here or will the cache suffice?

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
```diff
@@ -4,11 +4,15 @@ import (
...
 	"fmt"
+	"github.com/rancher/rancher/pkg/provisioningv2/rke2/installer"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
 	"testing"
 
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	"github.com/rancher/rancher/pkg/settings"
 
+	pv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
```
jakefhyde, line 15:
```suggestion
	provcontrollers "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
```

jakefhyde (changes requested):
Couple nits

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
HarrisonWAffel: I tried to use a cache but it was always missing the cluster object, so a client seemed like the only option. If you have any tips on why that might be I'm all ears, but using `clients.Provisioning.Cluster().Cache()` seemed to be updating fast enough for us to use it here.

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
```diff
@@ -4,11 +4,15 @@ import (
...
 	"fmt"
+
+	provisioningcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/provisioningv2/rke2/installer"
 	"testing"
 
 	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
 	"github.com/rancher/rancher/pkg/settings"
 
+	provcontrollers "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
```
jakefhyde, line 15:
```suggestion
	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
```
My bad, should've clarified this.

pkg/controllers/provisioningv2/rke2/bootstrap/controller_test.go
```diff
@@ -4,11 +4,15 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
+
+	provisioningcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
```
jakefhyde, line 8:
```suggestion
	provcontrollers "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
```

This should probably stay `provcontrollers` so it's clear which import is for the controllers and which is for the types.

jakefhyde (dismissed):

Oats87 (changes requested):

pkg/provisioningv2/rke2/installer/installer.go
```diff
@@ -99,10 +117,11 @@ func LinuxInstallScript(ctx context.Context, token string, envVars []corev1.EnvV
 %s
 
 %s
-`, envVarBuf.String(), binaryURL, server, ca, token, data)), nil
+%s
+`, envVarBuf.String(), binaryURL, server, ca, token, strings.ReplaceAll(string(data), "exit 0", ""), combinedHooks)), nil
```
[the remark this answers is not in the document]
HarrisonWAffel: I've updated it so that the hooks are injected right after the shebang, so we don't mess with the `exit 0` call at all.

jakefhyde:
@Oats87 @HarrisonWAffel We could also add the hooks as a proper field on the bootstrap object: https://github.com/rancher/rancher/blob/77f87c498b1517dcca5ae7fcf707dc8ebe7a95db/pkg/apis/rke.cattle.io/v1/bootstrap.go#L17-L19
Here is an example of how upstream does this:
https://github.com/kubernetes-sigs/cluster-api/blob/c1730d3cb9344bff059d92cc9a18b6b2a6faab04/bootstrap/kubeadm/api/v1alpha4/kubeadmconfig_types.go#L63
https://github.com/kubernetes-sigs/cluster-api/blob/6a4e90cad62de2f4ec0f48ed92f94d348ed41035/bootstrap/kubeadm/internal/controllers/kubeadmconfig_controller.go#L486

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -131,7 +136,25 @@ func (h *handler) getBootstrapSecret(namespace, name string, envVars []corev1.En
 	if os := machine.GetLabels()[rke2.CattleOSLabel]; os == rke2.WindowsMachineOS {
 		is = installer.WindowsInstallScript
 	}
-	data, err := is(context.WithValue(context.Background(), tls.InternalAPI, hasHostPort), base64.URLEncoding.EncodeToString(hash[:]), envVars, "")
+
+	// find the machine pool which the current machine belongs to
+	// so that we can utilize any provided hooks.
+	provCluster, err := h.provClusters.Get(namespace, machine.Spec.ClusterName)
```
jakefhyde, line 142: I feel like directly accessing the provisioning cluster may not be the most appropriate way to do this. It's the simplest, but if we want to stick closer to upstream, we would refactor this to create a bootstrap template per pool: https://github.com/rancher/rancher/blob/release/v2.7/pkg/controllers/provisioningv2/rke2/provisioningcluster/template.go#L235. cc @Oats87

jakefhyde (changes requested):

Oats87:
> @Oats87 @HarrisonWAffel We could also add the hooks as a proper field on the bootstrap object:
> 
> https://github.com/rancher/rancher/blob/77f87c498b1517dcca5ae7fcf707dc8ebe7a95db/pkg/apis/rke.cattle.io/v1/bootstrap.go#L17-L19
> 
> Here is an example of how upstream does this:
> https://github.com/kubernetes-sigs/cluster-api/blob/c1730d3cb9344bff059d92cc9a18b6b2a6faab04/bootstrap/kubeadm/api/v1alpha4/kubeadmconfig_types.go#L63
> https://github.com/kubernetes-sigs/cluster-api/blob/6a4e90cad62de2f4ec0f48ed92f94d348ed41035/bootstrap/kubeadm/internal/controllers/kubeadmconfig_controller.go#L486

Yeah, this is definitely something we could do. I dislike how it's specifically "preKubeadm" i.e. specifically kubeadm, but that doesn't necessarily mean the idea itself isn't bad. 


On a semi-related note, @HarrisonWAffel did we have design discussion where we formally decided on using "hooks"? I think "hook" might be too generic of a field for what these specifically are. they really are "preHook"s.

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
Oats87: Yes, the most ideal manner of fixing this is to create a bootstrap template per pool; however, if we make this change we need to figure out how to make the change in such a way where we do not cause reprovisioning of existing machine pools when we change the name of the bootstrap template, and we need to clean up the old bootstrap templates. If possible, I really would like to clean up that behavior.

I'm not sure if making that change makes the scope of adding these prehooks too big for the given milestone though, but I would much prefer this method. 

cc: @snasovich

Oats87:
@HarrisonWAffel @jakefhyde @snasovich before we merge this PR, we should discuss this in an afterparty item (or more realistically, in a design discussion). @jakefhyde brings up some good points about the implementation of this in our codebase vs. how upstream "would" do it.

Issue #37998, opened 2022-06-15
Once setting is configured with an env var, it only be changed with env var and configuring setting env var to be "" string changes value in rancher api but not in kubernetes


rmweir:
**Rancher Server Setup**
- Rancher version: v2.5.12

**Information about the Cluster**
- Kubernetes version: k3s 1.19 but probably doesn't matter
   
**Describe the bug**

~~Unsetting CATTLE_AGENT_IMAGE before an upgrade does not cause downstream cluster agents to use proper image. They will continue using the value that was previous set for CATTLE_AGENT_IMAGE. This is happening because setting the env var properly sets the setting in rancher. You can check this by running `kubectl get setting agent-image`.~~
The above works as intended but a related issue was found that will be tracked using this issue.

When a setting was configured using an env var of the format
CATTLE_SETTING_NAME, the source was set to "env". That source
would then persist even when the env var was unset in the future.
This prevented users from ever changing a setting value through the
rancher API or UI.

Also, when using env var to set the settings value to empty string, the rancher api will return empty string but the value is not updated in kubernetes. This will cause a change in behavior once the setting is interacted with through means other than norman. This could be moved to webhook but that won't be necessary once this is fixed. Once validation is moved to webhook, kubectl workaround won't be an option either.

There is a work around. You can use `kubectl edit setting <setting-name>`, then set the value and source fields to "" (empty string). Then redeploy rancher. You can also set the value to empty string with kubectl as well.

**To Reproduce**
1. Setup rancher installation.
2. Restart/upgrade rancher using env var such as CATTLE_EKS_REFRESH_CRON to set a setting that is present in the setting UI.
3. Try to change the value in the UI, should not be possible. The option will not be available in UI but even the API should not work for this.

for empty val scenario:
1. Set settings value to something that is a non empty string. You can use env var, rancher, or kubectl.
2. Now set it to empty string, "", using CATTLE_<setting-name> env var.
3. Get the setting from rancher will show the value as "" but if you get the value with kubectl it will be the last non empty string value you set it as.

**Result**
~~Downstream cluster agents are still using rancher/rancher-agent:v2.5.12.~~
The option will not be available in UI but even the API should not work for this.

**Expected Result**

The value of the setting should be the same but you should now be able to change it using the API and UI. Setting the env var value to empty string should now be reflected in kubernetes and in rancher api.

cbron:
Moving this back to 2.7.1 now

brudnak:
[...]

PR #39453, opened 2022-10-28
Remove env on unset

rmweir:
## Issue: 

https://github.com/rancher/rancher/issues/37998

## Problem

Prior, when a setting was configured using an env var of the format
CATTLE_<setting-name>, the source was set to "env". That source
would then persist even when the env var was unset in the future.
This prevented users from ever changing a setting value through the
rancher API or UI.

To clarify: After configuring a setting's value with an env var, that value should persist even once the env var is unset. However, the user should be able to change it once it has been unset.

## Solution

Now, unsetting the env var during an upgrade
will cause the source to be reset. This will allow users to change
the value again as long as the setting is no longer longer set by
an env var at setup.
 
### Manual Testing

Configure setting CATTLE_EKS_REFRESH_CRON using env var. Tried to change value via UI. Was unable to which is expected. Then unset the env var. Should be able to update the env var with UI (should not need to use kubectl), was not able to. After fix I was able to change the env with UI after unsetting the env var or setting the value to empty string "".

### Automated Testing

Added unit tests to cover updating source to "env", unsetting source from "env", updating value when it should, not updating value when it shouldn't, updating default when changed.

## QA Testing Considerations

Setup rancher installation.
Restart/upgrade rancher using env var such as CATTLE_EKS_REFRESH_CRON to set a setting that is present in the setting UI.
Try to change the value in the UI, should not be possible. The option will not be available in UI but even the API should not work for this.
Restart/upgrade rancher unsetting the env var. The value of the setting should be the same but you should now be able to change it using the API and UI.


pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,129 @@
...
+		settingMap[testCase.newSetting.Name] = testCase.newSetting
+		if testCase.envVar != nil {
+			envKey := settings.GetEnvKey(testCase.newSetting.Name)
+			os.Setenv(envKey, *testCase.envVar)
+			defer os.Unsetenv(envKey)
+		}
+
+		settingsETCD[testCase.newSetting.Name] = testCase.existingSetting
+	}
+	testSettingsProvider.SetAll(settingMap)
```
MbolotSuse, line 65: This call ignores the output of `SetAll` - might be good to make sure that this is nil in the cases we want them to be.

pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,129 @@
...
+		finalSetting := settingsETCD[testCase.newSetting.Name]
+		failMsg := fmt.Sprintf("test case failed [%s]: %s", testCase.newSetting.Name, testCase.description)
+		assert.Equal(t, finalSetting.Default, testCase.newSetting.Default, failMsg)
+		assert.True(t, finalSetting.Source == "env" == (testCase.envVar != nil), failMsg)
+		if testCase.envVar != nil {
+			assert.Equal(t, *testCase.envVar, finalSetting.Value, failMsg)
+			continue
+		}
+		if testCase.existingSetting != nil {
+			assert.Equal(t, testCase.existingSetting.Value, finalSetting.Value)
```
MbolotSuse, line 76: Is this universally true? From what I can see, if we have an existing setting that's not from the env, but then we have a setting from the env, the env's value will take precedent and override, making the existing setting's value not the final setting's value.

pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type testCase struct {
+	description     string
+	envVar          *string
+	newSetting      settings.Setting
+	existingSetting *v3.Setting
```
MbolotSuse, line 21: One suggestion here - I've found it helpful to include the expected outcome in test cases. That way it's easier for people to see what you are looking to get out of a test case and how to make a new one if the code changes. You don't have to make this change, just a suggestion.

pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,129 @@
...
+			envKey := settings.GetEnvKey(testCase.newSetting.Name)
+			os.Setenv(envKey, *testCase.envVar)
+			defer os.Unsetenv(envKey)
+		}
+
+		settingsETCD[testCase.newSetting.Name] = testCase.existingSetting
+	}
+	testSettingsProvider.SetAll(settingMap)
+	for _, testCase := range testCases {
+		finalSetting := settingsETCD[testCase.newSetting.Name]
```
MbolotSuse, line 67: Nit: There doesn't appear to be any checks on the setting of testSettingsProvider.fallback. That's ok since that value is internal to the object, but I would recommend considering if we want to assert on the behavior of that value.

pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,129 @@
...
+			assert.Equal(t, testCase.existingSetting.Value, finalSetting.Value)
+			continue
+		}
+		assert.Equal(t, "", finalSetting.Value)
+	}
+}
+
+func populateTestCases() []testCase {
+	testCases := []testCase{
+		{
```
MbolotSuse, line 85: I think these test cases are a good start. There are a few others that also might be good to include:
- Creating a new setting that comes from the env var
- Error on creating a new setting
    - Already exists error?
    - Not already exists error?
- Setting that was not an env var, but now we have an env var for it- https://github.com/rancher/rancher/pull/39453/files#diff-8645d9e3510c3b2f7074d64ef2b316ae3b83a9740f1c60b8a5ab2c92c3483d16L116-L119
- Error when attempting to update the object

These are assuming that we don't want to test the value of fallback. If we end up wanting to do that, then that expands the number of tests a bit more as well.

MbolotSuse (changes requested):
Non-test code LGTM outside of a comment nit.

Have a few comments on the tests - namely, I think it would be a good idea to have a couple more test cases, though I'm open to disagreement.

As far as the overall idea goes - to be honest, I'm not a fan of trying to lock settings based on the source. IMO a better way of this working would be that on boot rancher replaces the settings (in k8s) with what it has in the env (if a setting was specified in env). That would be all - no locking of source, and if you want to make a change after rancher is up, change it in k8s/through rancher the rancher api as an intermediary. But that's a bigger change and what you have here works, so we can stick with this.

pkg/controllers/dashboardapi/settings/settings_test.go
rmweir: Yes, the above checks for this. In this case, envVar would not be nil and would hit continue.

pkg/controllers/dashboardapi/settings/settings_test.go
rmweir: I already have description for the test case so I went ahead changed the description to hopefully explain what the expected behavior is of each.

rmweir:
I'm midway through addressing feedback, so this PR is currently a WIP.

pkg/controllers/dashboardapi/settings/settings_test.go
rmweir: Added checks for this

pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,129 @@
...
+	newSetting      settings.Setting
+	existingSetting *v3.Setting
+}
+
+var settingsETCD = make(map[string]*v3.Setting)
+
+// TestSetAll ensures that SetAll properly updates settings in ETCD. These updates
+// are informed by env vars of the form CATTLE_<SETTING_NAME> and the passed map
+// of settings.Setting objects.
+func TestSetAll(t *testing.T) {
```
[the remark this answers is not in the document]
rmweir: removed

pkg/controllers/dashboardapi/settings/settings.go
```diff
@@ -75,6 +78,10 @@ func (s *settingsProvider) SetIfUnset(name, value string) error {
 	return err
 }
 
+// SetAll iterates through a map of settings.Setting and updates corresponding settings in ETCD
```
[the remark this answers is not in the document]
rmweir: Swapped references to etcd in this file to k8s.

pkg/controllers/dashboardapi/settings/settings_test.go
rmweir: Adding test cases for create returning error and update returning error. Already have a test case for setting with env var that wasn't one before. New setting can't "come from the env var." They should always be created in the file, but added a testcase for a setting that doesn't already exist and configuring with an env var. Added a test case for settings client return AlreadyExists error.

pkg/controllers/dashboardapi/settings/settings.go
```diff
@@ -113,11 +120,15 @@ func (s *settingsProvider) SetAll(settingsMap map[string]settings.Setting) error
...
+			if envOk && obj.Source != "env" {
 				obj.Source = "env"
 				update = true
 			}
-			if envValue != "" && obj.Value != envValue {
+			if !envOk && obj.Source == "env" {
+				obj.Source = ""
+				update = true
+			}
+			if envOk && obj.Value != envValue {
```
MbolotSuse, line 131: I think that this new code has a side effect in cases where there is a setting, with a different value and default, that is specified as `""` in the env. I don't know if this is necessarily a problem given that one of your commits is trying to explicitly support `""` as a setting value, but I did want to bring it up in case this was not intentional.

Consider the following case:
I have a setting which exists with:
`Value` = `"value"`
`Default` = `"default"`

I reboot rancher and specify this setting with an env var. The value from the env is `""`.

Under the old side code, the fallback for this would be `"value"` - since obj.Value is set to something. Under the new side code, obj.Value will be updated to `""` and fallback will be set to `"default"`, since on L135 set check to see if `obj.Value == ""` and fall back to the default case if that is true. I think that under your new approach of treating "" as a full, valid setting value, you probably want to change L135 to use `envOK` instead of `""` like you did for the other values. Although I could be wrong there, and I think that you will have a behavior change either way.

MbolotSuse (commented):
Have one functionality question on the new support for `""` values. 

One general comment on the tests- I had a tough time reading the tests to understand what they did. I won't hold up the PR for this (IMO test readability should not factor into PR approval), but I think it would be worth looking at if you have time. 

I think the tests have the following issues:
- The test cases are defined oddly. There's one section which defines/returns the test cases for the non-error cases, but it doesn't return the error cases. Those are tested separately, directly after the valid test cases, but don't have a table of their own, they just get ran one after another while being constructed directly. I found this difficult to parse and understand what test cases in total were being run and what the desired behavior was.
- The test cases don't define the expected behavior in them. I wrote a bit about this earlier, but each test case only defines what inputs should be given - you derive what outputs are expected rather than stating them explicitly. This makes the code more difficult to understand, and harder to change in the future if the requirements change, for whatever reason.

As said above, I won't hold up the PR for the test readability stuff, but I did feel it was worth noting.

pkg/controllers/dashboardapi/settings/settings.go
rmweir: This was actually on purpose because the transformer in the store makes it effectively this way when the setting is retrieved from the UI. The effective value of a setting should be default if the value is false.  However, I did not put enough the into the fact that this change will effect the backend. Once settings are interacted with a way that isn't settings there will also be a side effect, unless the transformer store stuff is entirely ported to webhook, which this change is trying to prevent.

rmweir: @MbolotSuse I spoke with @cbron and confirmed that this is desired. It will change the value that the backend would receive for the setting, whether that be from client or from fallback. However, the current behavior is much worse. Today someone can set `auth-user-info-max-age-seconds` using the env var to something like "600", then change it again with the env var to "", and in the API will return "" for the value (the ui will show default because of this), while the backend is still using the "600" value. We want to sync these behaviors to be the same and treat setting the value to "" properly. This means fallback being default, and Get() returning default value.

rmweir:
@MbolotSuse I updated one of the test case description to test creating a setting that doesn't exist yet creates the setting in kubernetes. Otherwise, I think that the descriptions to describe the expected behavior. For example, `test a setting that doesn't exist with val and \"env\" source being configured with env var creates setting with env var value and \"env\" source` and `test changing default of existing setting with a value properly updates the default but doesn't change value` describe what the expected behavior is to me. Do you feel that in those example I should be saying explicitly the value I expect on each object?

rmweir:
Also updated some comments around error test cases as they were incorrect.

MbolotSuse (approved):
LGTM

pkg/controllers/dashboardapi/settings/settings_test.go
```diff
@@ -0,0 +1,263 @@
...
+				Source:  "env",
+			},
+		},
+	}
+
+	for index, test := range testCases {
+		settingName := fmt.Sprintf("test%d", index)
+		newSetting := settings.NewSetting(settingName, test.newDefVal)
+		test.newSetting = newSetting
+		if test.existingSetting == nil {
```
maxsokolovsky, line 253: Nit: the if statement with the continue statement could be moved to the top of the loop's body.

maxsokolovsky (approved):

PR #39482, opened 2022-11-02
[v2.7] allow usage of other helm repos to install rancher

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/515
 
## Problem

 currently, rancher helm URL was hard coded. 

## Solution

This PR will allow different URLs for a rancher install to be used.
Introduced `RANCHER_HELM_URL` variable with previous hardcoded value as the default. 
supporting PR in job builder


sowmyav27 (approved):

anupama2501 (approved):
LGTM

Issue #39456, opened 2022-10-28
[BUG] Upgrade from non-community version -> community version of rancher does not reset branding / `ui-brand` persists after upgrade


slickwarren:
**Rancher Server Setup**
- Rancher version: v2.7.0-rc6
- Installation option (Docker install/Helm Chart): docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):n/a
- Proxy/Cert Details:n/a

**Information about the Cluster**
- Kubernetes version: v1.24.4
- Cluster Type (Local/Downstream):local
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): n/a

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: tested with admin user (upgrade use case, must be admin to view settings)

**Describe the bug**

When upgrading from a non-community version of rancher to a community version, the upgrade is successful and the private registry details are reset. However, the non-community version's branding persists.

**To Reproduce**

* deploy a non-community version of rancher
  * ensure that non-community version's registry and branding are set 
* upgrade the setup to a community version of rancher
  * wait for the upgrade to complete

**Result**
the `ui-brand` setting persists after upgrade, and so we see non-community branding after the upgrade. 

**Expected Result**

the `ui-brand` should clear on an upgrade from a non-community version of rancher to a community version of rancher. 

**Screenshots**

**Additional context**

upgrading from a community-version to a non-community version of rancher does properly set the `ui-brand` and registry.

further investigation reveals that the `CATTLE_UI_BRAND` is not set after the upgrade, therefore this is likely a UI bug. Transferring to the UI repo.

cbron:
Making some assumptions, @slickwarren please correct if they are not right: 

1. The upgrade steps here was just switching which chart is used, no other env vars or settings were manually changed. 
2. The non-community env var is `CATTLE_UI_BRAND="suse"`, and above comments seem to convey that the env var is gone, but the setting is there, and the setting is still set to "suse". No other custom ui-brand was set (if it was I imagine this would be fixed). 

So I think this is a known case where once you set the env var, it sticks, and removing the env var does not remove the setting. The `system-default-registry` should actually also still be set, @slickwarren can you confirm ?

slickwarren:
confirmed (also had some offline discussion)

crobby:
[...]

PR #39493, opened 2022-11-03
Set CATTLE_UI_BRAND to empty string for community installs

crobby:
Explicitly setting the CATTLE_UI_BRAND env var to an empty string will force an update of the value at start-time which will allow for proper transitioning between
community and non-community installs of Rancher.

## Issue: 
https://github.com/rancher/rancher/issues/39456
 
## Problem
When upgrading from a non-community version of rancher to a community version, the upgrade is successful and the private registry details are reset. However, the non-community version's branding persists.
 
## Solution
When building the community image, we will now set the CATTLE_BRAND_UI env var to an empty string.  This is different than leaving it unset and will force the value to be updated.
 
## Testing
Manually tested by eng

### Manual Testing
Start a cluster and install Rancher Prime
Note that the logo in the dashboard is the prime logo.
"upgrade" to a non-prime version of rancher (ie:  shut down rancher prime container and start a non-prime rancher)
Verify that the branding logo has changed back to the community logo.
"upgrade" back to Rancher Prime
Note that the logo is the prime logo.

### Automated Testing
N/A

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -169,6 +169,9 @@ ENV CATTLE_CLI_VERSION v2.7.0-rc1
 # Please update the api-ui-version in pkg/settings/settings.go when updating the version here.
 ENV CATTLE_API_UI_VERSION 1.1.9
 
+# Explicitly set the branding to an empty string for non-suse branded installs
+ENV CATTLE_UI_BRAND=
+
 RUN mkdir -p /var/log/auditlog
 ENV AUDIT_LOG_PATH /var/log/auditlog/rancher-api-audit.log
 ENV AUDIT_LOG_MAXAGE 10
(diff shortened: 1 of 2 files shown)


rmweir (changes requested):
I don't think whether env val matches obj default should have any bearing on anything. Default can always be changed. If the value is configured by env var, even if it matches default it should have source set as "env" so that it is locked. My point before was that maybe default should be "_" or "community". If we are going to do the empty string approach I think the change that should be made is in SetAll(). It should be changed so that empty string is not treat the same as unset. This will be done by replacing GetEnv with LookupEnv.

crobby:
Closing in favor of:  https://github.com/rancher/rancher/pull/39453

PR #39497, opened 2022-11-03
[V2.7] Only Regenerate Kubelet Serving Certificate When Necessary  

HarrisonWAffel:
## Problem
A denial of service is encountered when many clusters configured with `generate_serving_certificate` connect to the Rancher server at the same time. This is due to the fact that Rancher cannot intelligently determine if a certificate must be regenerated or not as kubelet certificates are not stored or tracked by the Rancher server. This resulted in Rancher always regenerating the kubelet certificate whenever the node agent reconnects with the Rancher server. At scale, this behavior results in all CPU time being dedicated to generating certificates. Additionally, Rancher was regenerating all of the nodes certificates for each node, resulting in many unused certificates being placed on each node and additional CPU resources being consumed. 
 
## Solution
Before attempting to reconnect with the Rancher server, the node agent will first find and inspect the kubelet serving certificate stored locally and determine if it needs to be regenerated. If the certificate needs to be regenerated, either because the SAN is out of date, the IP address is out of date, or the certificate will expire within three days, then a value is specified in the header of the reconnection attempt which will indicate that the certificate needs to be regenerated.
 
## Testing
when  `generate_serving_certificate` is enabled for the clusters kubelet service, the new kubelet certificate will be generated under these conditions 

1. The certificate does not already exist 
2. The certificate will expire in three days or less 
3. The hostname name included in the certificate does not match the nodes hostname name
4. The IP address included in the certificate does not match the current IP address of the node

Some of these conditions are hard to test, such as the expiry time which is set to 10 years in the future. The two cases below cover the main goal of this ticket. 

---
#### Test Environment / Setup
+ This issue requires that you have the ability to stop/scale down rancher to 0 pods. We are interested in the reconnection process once an existing Rancher server is brought back online. Any setup in which you can restart Rancher will work, but an HA setup will likely be the easiest configuration.
+ Create an RKE1 cluster with three nodes 
  + one of the nodes will have the etcd control plane roles 
  + two of the nodes will only have the worker role 
+ Before creating the cluster, edit the YAML and find the `services` section. Add the `generate_serving_certificate` [as described in this documentation ](https://rancher.com/docs/rke/latest/en/config-options/services/#kubelet)
+ Create the cluster and wait for all nodes to come online. Once they've come online, double check the config YAML and ensure that `generate_serving_certificate` is set to `true`. 

#### 1) Certificate should not regenerate on each reconnection

+ SSH into one of the worker only nodes and navigate to `/etc/kubernetes/ssl`. You should see several `.pem` files with the following naming convention `kube-kubelet-<IP_ADDRESS>.pem` and `kube-kubelet-<IP_ADDRESS>-key.pem`. Mark the time at which these files were created using `ls -l`. Make sure you mark the file which belongs to current node you are SSH'd into 
+ Scale down / restart the rancher server and wait for all nodes to reconnect
+ Ensure that no certificates have been updated by checking the file creation time

#### 2) Certificate should be recreated when deleted
+ SSH into one of the worker only nodes and navigate to `/etc/kubernetes/ssl`. Delete the `kube-kubelet` certificate and key which belongs to the node you are SSH'd into (the file name will include the nodes IP address)
+ scale down / restart the rancher server and wait for all nodes to reconnect 
+ ensure that the deleted files have been recreated and are present on the node
+ Also ensure that no other certificate files have been updated or modified

### Manual Testing
I've done the above steps during development to ensure that the new node agent logic works as desired. 

### Automated Testing
N/A

## QA Testing Considerations
You will need an HA rancher server or any other type of Rancher server which can be restarted.
 
Any features or functionalities which require the proper configuration of the kubelet serving certificate.

### DIFF
--- a/cmd/agent/agent_test.go
+++ b/cmd/agent/agent_test.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"github.com/stretchr/testify/assert"
+	"math/big"
+	"net"
+	"testing"
+	"time"
+)
+
+func TestKubeletCertificateNeedsRegeneration(t *testing.T) {
+
+	correctIPAddress := "167.71.188.113"
+	correctHostName := "cert-test1"
+	testCertExpiryTime := time.Now().AddDate(10, 0, 3)
+
+	type testcase struct {
+		testName      string
+		testIPAddress string
+		testHostName  string
+		testTime      time.Time
+		want          bool
+	}
+
+	testCases := []testcase{
+		{
+			testName:      "All Cert Properties Are Valid",
+			testIPAddress: correctIPAddress,
+			testHostName:  correctHostName,
+			testTime:      time.Now(),
+			want:          false,
+		},
+		{
+			testName:      "Cert Contains Invalid IP Address",
+			testIPAddress: "192.168.1.1",
+			testHostName:  correctHostName,
+			testTime:      time.Now(),
+			want:          true,
+		},
+		{
+			testName:      "Cert Contains Invalid Hostname",
+			testIPAddress: correctIPAddress,
+			testHostName:  "different-hostname",
+			testTime:      time.Now(),
+			want:          true,
+		},
+		{
+			testName:      "Cert Will Expire In Less Than Three Days",
+			testIPAddress: correctIPAddress,
+			testHostName:  correctHostName,
+			testTime:      time.Now().AddDate(10, 0, 1), // ~ 48 hours until expiry
+			want:          true,
+		},
+	}
+	t.Log("generating test certificate...")
+	testCert, err := createTestCert([]net.IP{net.ParseIP(correctIPAddress)}, []string{correctHostName}, testCertExpiryTime)
+	assert.Equal(t, nil, err)
+	t.Log("successfully generated test certificate")
+	for _, c := range testCases {
+		t.Run(c.testName, func(t *testing.T) {
+			got, err := KubeletCertificateNeedsRegeneration(c.testIPAddress, c.testHostName, testCert, c.testTime)
+			assert.Equal(t, nil, err)
+			assert.Equal(t, c.want, got)
+		})
+	}
+}
+
+// createTestCert creates a self-signed certificate for use in tests, it incorporates the given ipAddress's, hostNames, and expiry time
+func createTestCert(ipAddress []net.IP, hostname []string, notAfter time.Time) (tls.Certificate, error) {
+	cert := &x509.Certificate{
+		SerialNumber: big.NewInt(2022),
+		Subject: pkix.Name{
+			Organization:  []string{"Rancher"},
+			Country:       []string{"US"},
+			Province:      []string{""},
+			Locality:      []string{"Green Pastures"},
+			StreetAddress: []string{"123 Cattle Drive"},
+			PostalCode:    []string{"94016"},
+		},
+		NotBefore:    time.Now(),
+		NotAfter:     notAfter,
+		KeyUsage:     x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		SubjectKeyId: []byte{5, 4, 1, 6, 8},
+		DNSNames:     hostname,
+		IPAddresses:  ipAddress,
+	}
+
+	certPriv, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		return tls.Certificate{}, err
+	}
+	privKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		return tls.Certificate{}, err
+	}
+	certBytes, err := x509.CreateCertificate(rand.Reader, cert, cert, &certPriv.PublicKey, privKey)
+	if err != nil {
+		return tls.Certificate{}, err
+	}
+
+	certPem := new(bytes.Buffer)
+	certPrivPem := new(bytes.Buffer)
+	if err = pem.Encode(certPem, &pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: certBytes,
+	}); err != nil {
+		return tls.Certificate{}, err
+	}
+	if err = pem.Encode(certPrivPem, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(certPriv),
+	}); err != nil {
+		return tls.Certificate{}, err
+	}
+
+	return tls.X509KeyPair(certPem.Bytes(), certPrivPem.Bytes())
+}
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/rancher/rancher/pkg/features"
 	"github.com/rancher/rancher/pkg/logserver"
 	"github.com/rancher/rancher/pkg/rkenodeconfigclient"
+	"github.com/rancher/rancher/pkg/rkenodeconfigserver"
 	"github.com/rancher/remotedialer"
 	"github.com/rancher/wrangler/pkg/signals"
 	"github.com/sirupsen/logrus"
@@ -41,7 +42,8 @@ var (
 )
 
 const (
-	Token = "X-API-Tunnel-Token"
+	Token                    = "X-API-Tunnel-Token"
+	KubeletCertValidityLimit = time.Hour * 72
 )
 
 func main() {
@@ -203,9 +205,14 @@ func run(ctx context.Context) error {
 		return err
 	}
 
-	headers := map[string][]string{
+	headers := http.Header{
 		Token:                      {token},
 		rkenodeconfigclient.Params: {base64.StdEncoding.EncodeToString(bytes)},
+	}
+
+	err = KubeletNeedsNewCertificate(headers)
+	if err != nil {
+		return err
 	}
 
 	serverURL, err := url.Parse(server)
@@ -337,6 +344,12 @@ func run(ctx context.Context) error {
 			for {
 				select {
 				case <-time.After(tt):
+					// each time we request a plan we should
+					// check if our cert about to expire
+					err = KubeletNeedsNewCertificate(headers)
+					if err != nil {
+						logrus.Errorf("failed to check validity of kubelet certs: %v", err)
+					}
 					receivedInterval, err := rkenodeconfigclient.ConfigClient(ctx, connectConfig, headers, writeCertsOnly)
 					if err != nil {
 						logrus.Errorf("failed to check plan: %v", err)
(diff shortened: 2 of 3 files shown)


cmd/agent/main.go
```diff
@@ -485,3 +510,75 @@ func reconcileKubelet(ctx context.Context) (bool, error) {
...
+	}
+
+	if !hostnameIncludedInSAN || certIsExpiring || ipAddressHasChanged {
+		headers[RegenerateKubeletCertificate][0] = "true"
+	} else {
+		headers[RegenerateKubeletCertificate][0] = "false"
+	}
+
+	return nil
+}
```
jakefhyde, line 584:
```suggestion
func KubeletNeedsNewCertificate() (bool, error) {
  ipAddress := getIPAddress()
  fileSafeIpAddress := strings.ReplaceAll(ipAddress, ".", "-")
  certFileLocation := fmt.Sprintf("/etc/kubernetes/ssl/kube-kubelet-%s.pem", fileSafeIpAddress)
	certKeyFileLocation := fmt.Sprintf("/etc/kubernetes/ssl/kube-kubelet-%s-key.pem", fileSafeIpAddress)
	
	return needsNewCertificate(ipAddress, certFileLocation, certKeyFileLocation, time.Hour*72)
}

func needsNewCertificate(ipAddress, certFileLocation, certKeyFileLocation string, duration time.Duration) (bool, error) {
	cert, err := tls.LoadX509KeyPair(certFileLocation, certKeyFileLocation)
	if err != nil && !strings.Contains(err.Error(), "no such file") {
		return false, err
	}

	if len(cert.Certificate) == 0 {
		return true, nil
	}
	
	parsedCert, err := x509.ParseCertificate(cert.Certificate[0])
	if err != nil {
		return false, err
	}
	
	if certExpired(parsedCert, duration) {
		return true, nil
	}
	
	currentHostname, err := os.Hostname()
	if err != nil {
		return false, err
	}
	
	if hostnameIncludedInSAN(parsedCert, hostname) {
		return true, nil
	}
	
	if ipChanged(parsedCert, ipAddress) {
		return true, nil
	}
	
	return false, nil
}

func certExpired(cert *x509.Certificate, duration time.Duration) (bool, error) {
	if cert.NotAfter.Sub(time.Now()) < duration {
		return true, nil
	}
	
	return false, nil
}

func hostnameIncludedInSAN(cert *x509.Certificate, hostname string) bool {
	// ensure that the current certificate
	// contains the current hostname
	hostnameIncludedInSAN := false
	for _, name := range cert.DNSNames {
		if name == hostname {
			return true
		}
	}
	return false
}

func ipChanged(cert *x509.Certificate, ipAddress string) bool {
	for _, ip := range parsedCert.IPAddresses {
		if ipAddress == ip.String() {
			return false
		}
	}
	return true
}
```
Overall it looks good, I would just do some minor refactoring to make the code more testable. This would make it possible to add both test certs in `testdata`, as well as unit tests that verify simple cases like the ip changing and the hostname. This also assumes a change to split determining whether to add the header, and actually adding it into 2 functions.

jakefhyde (changes requested):
Just a few nits

cmd/agent/main.go
```diff
@@ -485,3 +510,75 @@ func reconcileKubelet(ctx context.Context) (bool, error) {
...
+func getIPAddress() string {
+	conn, err := net.Dial("udp", "1.3.3.7:1")
+	if err != nil {
+		return ""
+	}
+	defer conn.Close()
+	return conn.LocalAddr().(*net.UDPAddr).IP.String()
+}
+
+func KubeletNeedsNewCertificate(headers map[string][]string) error {
```
sbstp, line 526: Nit: Might be beyond the scope of this PR, but maybe `http.Header` could be used here. I think `headers.Set(RegenerateKubeletCertificate, "true")` would be cleaner than `		headers[RegenerateKubeletCertificate][0] = "true"`

cmd/agent/main.go
```diff
@@ -485,3 +510,75 @@ func reconcileKubelet(ctx context.Context) (bool, error) {
 	}
 	return false, nil
 }
+
+// getIPAddress gets the machines primary outbound IP address by making a UDP
+// request to a random (non-existent) host and taking the local address
+// used to make the request.
+func getIPAddress() string {
+	conn, err := net.Dial("udp", "1.3.3.7:1")
```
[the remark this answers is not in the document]
sbstp: I think that if there's multiple network interfaces in the container it might pick the wrong one but I'm far from an expert in k8s networking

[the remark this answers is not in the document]
HarrisonWAffel: Sounds good, I'll look into other approaches to get the primary outbound IP

cmd/agent/main.go
HarrisonWAffel: Good point!

cmd/agent/main.go
HarrisonWAffel: Alrighty, refactored and tests + test data have been added.

jakefhyde (commented):
Nice job refactoring and adding tests!

cmd/agent/agent_test.go
```diff
@@ -0,0 +1,62 @@
...
+			want:                true,
+		},
+	}
+
+	for _, c := range testCases {
+		got, err := KubeletCertificateNeedsRegeneration(c.testIPAddress, c.testHostName, c.testCertLocation, c.testCertKeyLocation, c.testTime)
+		assert.Equal(t, nil, err)
+		assert.Equal(t, c.want, got)
+	}
+}
```
jakefhyde, line 62:
```suggestion
	tests := []struct {
		name								string
		testCertLocation    string
		testCertKeyLocation string
		testIPAddress       string
		testHostName        string
		testTime            time.Time
		want                bool
	}{
		{
			name:								 "valid cert",
			testCertLocation:    testCertFileLocation,
			testCertKeyLocation: testCertKeyFileLocation,
			testIPAddress:       "167.71.188.113",
			testHostName:        "cert-test1",
			testTime:            time.Date(2022, 11, 3, 12, 1, 1, 1, time.Local),
			want:                false,
		},
		{
			name:								 "bad ip",
			testCertLocation:    testCertFileLocation,
			testCertKeyLocation: testCertKeyFileLocation,
			testIPAddress:       "192.168.1.1",
			testHostName:        "cert-test1",
			testTime:            time.Date(2022, 11, 3, 12, 1, 1, 1, time.Local),
			want:                true,
		},
		{
			name:								 "bad hostname",
			testCertLocation:    testCertFileLocation,
			testCertKeyLocation: testCertKeyFileLocation,
			testIPAddress:       "167.71.188.113",
			testHostName:        "bad-hostname",
			testTime:            time.Date(2022, 11, 3, 12, 1, 1, 1, time.Local),
			want:                true,
		},
		{
			name:								 "30 hours until expired",
			testCertLocation:    testCertFileLocation,
			testCertKeyLocation: testCertKeyFileLocation,
			testIPAddress:       "167.71.188.113",
			testHostName:        "cert-test1",
			testTime:            time.Date(2032, 10, 30, 12, 1, 1, 1, time.Local),
			want:                true,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got, err := KubeletCertificateNeedsRegeneration(tt.testIPAddress, tt.testHostName, tt.testCertLocation, tt.testCertKeyLocation, tt.testTime)
			assert.Equal(t, nil, err)
			assert.Equal(t, tt.want, got)
		}
	}
}
```
This is usually the more idiomatic way to write tests, if you are using goland you should see each entry in `tests` have a green arrow next to it so you can run the tests individually. I think it has to be both an anonymous struct definition as well as wrapped in a `t.Run()`, as I just tried it with a `testcase` struct and it didn't give me any test prompts.

jakefhyde: I wish there was a way to ignore whitespace when making a suggestion.

HarrisonWAffel: Good to know, test is updated to use `t.Run` so we can include test case names

jakefhyde (approved):

cmd/agent/main.go
```diff
@@ -485,3 +512,70 @@ func reconcileKubelet(ctx context.Context) (bool, error) {
 	}
 	return false, nil
 }
+
+func KubeletNeedsNewCertificate(headers http.Header) error {
+	ipAddress := os.Getenv("CATTLE_ADDRESS")
+	fileSafeIpAddress := strings.ReplaceAll(ipAddress, ".", "-")
+	certFileLocation := fmt.Sprintf("/etc/kubernetes/ssl/kube-kubelet-%s.pem", fileSafeIpAddress)
```
Oats87, line 519: I don't remember off the top of my head, but is there a way in RKE to define a different SSL directory? If so, can we use the logic from RKE to determine if the directory is what we expect? There's also a ton of helper logic for this here: https://github.com/rancher/rke/blob/b1afa9911a9111dffafde3c2e8eddce2fbcf1926/pki/util.go#L255

cmd/agent/main.go
```diff
@@ -41,7 +41,9 @@ var (
 )
 
 const (
-	Token = "X-API-Tunnel-Token"
+	Token                        = "X-API-Tunnel-Token"
+	RegenerateKubeletCertificate = "Regenerate-Kubelet-Certificate"
```
Oats87, line 45: nit: a certificate isn't being regenerated if it's a new node on the first run, but I think this is not really important as long as we're consistent.

Can you add a quick comment here just stating what this header does?

Oats87 (changes requested):
Overall, it looks great, just a few minor changes but otherwise, thank you for your work on this!

cmd/agent/testdata/kube-kubelet-167-71-188-113-key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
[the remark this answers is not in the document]
HarrisonWAffel: Okay, I've change the test to auto generate a new certificate on each run. tests should work until the end of time!

cmd/agent/main.go
HarrisonWAffel: I couldn't find an option in RKE which would allow users to change the location of the certs in the downstream node. There is a `--certs-dir` but that flag is for reading custom certs from the machine which is running `rke up`, and doesn't control the location they're added to in the downstream node

cmd/agent/main.go
```diff
@@ -365,6 +384,13 @@ func run(ctx context.Context) error {
 		if !isConnect() {
 			wsURL += "/register"
 		}
+
+		// check if we need a new kubelet cert on reconnection
+		err = KubeletNeedsNewCertificate(headers)
```
Oats87, line 389: I just realized something here -- this will only evaluate the certificate on disk for expiration on connection -- if the certificate is getting close to expiration (long lived agent), will it properly re-generate a new certificate to replace the old  one?

I'm not even sure if the original agent would have done that FWIW, but we should probably fix that. I do worry that swapping the certificate out from under the kubelet might also cause the kubelet to freak out/potentially cause downtime...

HarrisonWAffel: That is a good point. I didn’t see any code relating to proactive refreshing of certs when they’re close to expiring, even on the rancher side. I can’t speak for who implemented it initially, but maybe the expiry time of ten years played a part in that decision. Should we try to handle that case here, or create a new issue for it? 

Plus the agent has been wildly swapping out certificates with the kubelet running so far, so maybe that’s another behavior we want to look at.

HarrisonWAffel: You're right, it looks for a new one every two minutes. I've added a check which ensures the expiry date isn't within three days right before we request the plan, this should handle the long lived connection issue. I'm not sure if we should discuss additional logic to handle the swapping of the cert, so far it looks like swapping it in this way hasn't caused issues for the kubelet.

cmd/agent/main.go
```diff
@@ -337,6 +344,12 @@ func run(ctx context.Context) error {
 			for {
 				select {
 				case <-time.After(tt):
+					// each time we request a plan we should
+					// check if our cert about to expire
+					err = KubeletNeedsNewCertificate(headers)
```
Oats87, line 349: Looks great

Oats87 (approved):
Looks great! This functionality will be super helpful.

Oats87:
@kinarashah would you mind giving this PR a glance?

kinarashah (approved):

jakefhyde (approved):

PR #39501, opened 2022-11-03
[v2.6] allow non-hardened cluster provisioning

slickwarren:
## Issue: 

section. -->
Jenkins would use the hardened rke1 cluster.yaml when hardened == false
 
## Solution

improve logic when checking hardened variable. Currently, we only checked that the variable existed (if hardened: x), likely assuming that the variable was a boolean. However it is a string, so the if statement was always true. 
 
## Testing

tbd


slickwarren:
closing in favor of https://github.com/rancher/rancher/pull/39535#event-7766776671

Issue #21637, opened 2019-07-19
feature request to add/tag clustername/id to rancher server logs


jambajaar:
Rancher version:
v2.2.3

Type:
Enhancement

Description:
Currently in a HA configuration with 3 application servers (min), rancher server pod on any one of the node can handle the client request and the logs are written as expected based on how the pod handles the request.

in troubleshooting scenarios, it would be good if each of the logs are 'tagged' with the cluster name or cluster-id to better distinguish the logs, if multiple requests are being handled for the same type, the logs type and data for the same type of log will be the same across various clusters and extremely hard to distinguish which cluster a particular log belongs to.

for this example snippet, its very hard which particular node/pod these messages are related to?

2019/07/18 20:29:53 [INFO] stdout: Waiting for machine to be running, this may take a few minutes... 
2019/07/18 20:29:53 [INFO] Waiting for machine to be running, this may take a few minutes...

prefixing the clustername/cluster ID will add value to the logs.

gz#5187

SURE-1033

jameson-mcghee:
During validation of this ticket on `v2.7-head` (Commit ID: 914f0f4), I was able to verify that when checking the logs for the "leader" node on an HA Rancher server, all downstream RKE1 cluster logs that are captured in the "leader" node for the `local` cluster will be prefaced by the `Mgmt Cluster` name for the downstream cluster as well as the node name for the downstream cluster's specific node (when applicable). I was also able to verify that modifying the `loglevel` to `DEBUG` or `TRACE` also works as intended. Therefore I am closing this ticket.

![image](https://user-images.githubusercontent.com/13294321/205747124-d3d66aba-87f1-48d7-b2ec-02bc111e3013.png)

PR #39508, opened 2022-11-04
[v2.7.1] Add the cluster-id to log messages during cluster creation

sbstp:
## Issue: 
#21637 
 
## Problem

Add the cluster-id to log messages during cluster creation in order to make trouble shooting easier.
 
## Solution

Add the Node's cluster id to the logs when forwarding messages coming from the driver.
 
### Manual Testing

1. Create a new cluster with the UI and look at the logs produced in stdout/stderr.

### DIFF
--- a/pkg/controllers/management/node/utils.go
+++ b/pkg/controllers/management/node/utils.go
@@ -225,9 +225,9 @@ func (m *Lifecycle) reportStatus(stdoutReader io.Reader, stderrReader io.Reader,
 		if strings.HasPrefix(msg, debugPrefix) {
 			// calls in machine with log.Debug are all prefixed and spammy so only log
 			// under trace and don't add to the v3.NodeConditionProvisioned.Message
-			logrus.Tracef("[node-controller] %v", msg)
+			logrus.Tracef("[node-controller] [cluster-id: %s] %v", node.ObjClusterName(), msg)
 		} else {
-			logrus.Infof("[node-controller] %v", msg)
+			logrus.Infof("[node-controller] [cluster-id: %s] %v", node.ObjClusterName(), msg)
 			v32.NodeConditionProvisioned.Message(node, msg)
 		}
 


pkg/controllers/management/node/utils.go
```diff
@@ -227,7 +227,7 @@ func (m *Lifecycle) reportStatus(stdoutReader io.Reader, stderrReader io.Reader,
 			// under trace and don't add to the v3.NodeConditionProvisioned.Message
 			logrus.Tracef("[node-controller] %v", msg)
```
HarrisonWAffel, line 228: We probably also want to include the new prefix here, for consistency

pkg/controllers/management/node/utils.go
```diff
@@ -227,7 +227,7 @@ func (m *Lifecycle) reportStatus(stdoutReader io.Reader, stderrReader io.Reader,
...
 		} else {
-			logrus.Infof("[node-controller] %v", msg)
+			logrus.Infof("[node-controller] [cluster-id: %s] %v", node.ObjClusterName(), m.clusterManager., msg)
```
HarrisonWAffel, line 230: I think you might have left `m.clusterManager.` in accidentally

HarrisonWAffel (changes requested):

pkg/controllers/management/node/utils.go
sbstp: Indeed. I forgot to add the fix.

pkg/controllers/management/node/utils.go
sbstp: Yeah that makes sense. Updated.

HarrisonWAffel (approved):
LGTM, this will add a prefix to all rancher-machine logs that are produced when creating RKE1 clusters. This still needs one more review from team 2, but should not be merged until v2.7.0 has been released.

sbstp:
> but should not be merged until v2.7.0 has been released.

Should I put the version in the title to v2.7.1 then?

HarrisonWAffel:
@sbstp Yeah that's a good idea. Also, adding `DNM` to issues that should not be merged is good practice, you can always update the title later

sbstp:
Just to be clear, the reason you think it shouldn't go into 2.7.0 is because it's really close to release?

snasovich:
> Just to be clear, the reason you think it shouldn't go into 2.7.0 is because it's really close to release?

Yes, we're well past "bugfix complete" date and are only addressing critical release issues now.

sbstp:
I wasn't able to find a do not merge / DMN label in github labels.

snasovich:
> I wasn't able to find a do not merge / DMN label in github labels.

@sbstp , there is not a formal label for DNM, we're putting it in title, e.g. like `[DNM - v2.7.1] PR title`.

snasovich:
@HarrisonWAffel @sbstp , as mentioned above, this addresses prefixing rancher-machine messages with cluster ID for RKE1 provisioned clusters.
I know the issue is about RKE1, but when it was opened there was no RKE2 provisioning.
Do we already have cluster ID included for rancher-machine messages generated during RKE2 clusters provisioning? If not, what would it take to cover this case as well?

HarrisonWAffel:
@snasovich When provisioning RKE2/K3s clusters Rancher will create a job to run the rancher-machine binary in a container. The logs of that container are not piped to the UI or to the rancher server logs. The naming of this already job denotes the cluster and node pool the new node belongs to, using the following format 

`job.batch/<CLUSTER_NAME>-<POOL_NAME>-<MGMNT_CLUSTER_ID>-<MACHINE_ID>-machine-provision`, e.g.
`job.batch/haffel-test-2-pool1-71cf79c1-rmg56-machine-provision`

snasovich (approved):

Issue #32127, opened 2021-04-14
Project/cluster owners get permissions errors when adding members/groups when PSPs are enabled


bentastic27:
It appears that there needs to be a global permission missing when project/cluster owners add users or groups to clusters/projects when PSPs are enabled

error creating binding: Forbidden 403: podsecuritypolicytemplateprojectbindings.management.cattle.io is forbidden: User "u-b4xv6mhhxu" cannot create resource "podsecuritypolicytemplateprojectbindings" in API group "management.cattle.io" in the namespace "c-7ldp6

Steps to reproduce:

1. enable PSPs on a cluster
2. create project owner and project member users
3. add the project owner user as an owner of a project
4. log in as the project owner and edit the project to add the project member

You'll be faced with that error, but the project member will be added. It doesn't seem to do this when you navigate to the project and use the members page, only when editing the project to add users that way. This same behavior seems to happen when adding cluster members as a cluster owner.

gz#16350
gz#16365

eumel8:
Just wondering that we faced this issue after upgrade from 2.5.11 to 2.6.8. Maybe we missed this permissions in a custome RoleTemplate, but I didn't found it in the built-in roles.

PR #39517, opened 2022-11-07
Adding permission for project owners to modify podsecuritypolicytemplateprojectbindings

crobby:
Added podsecuritypolicytemplateprojectbindings.management.cattle.io permissions to the set of rules that are added to the project-owner rolebinding.

## Issue: 

https://github.com/rancher/rancher/issues/32127

## Problem

If a (non-admin) project owner tried to modify the project's settings, even if the settings succeeded (like modifying resource quotas), the user would still see an error because the UI also makes a call to setpodsecuritypolicytemplateprojectbindings.  The error shown would look similar to this
`error creating binding: Forbidden 403: podsecuritypolicytemplateprojectbindings.management.cattle.io is forbidden: User "u-5tzkm" cannot create resource "podsecuritypolicytemplateprojectbindings" in API group "management.cattle.io" in the namespace "c-82nlk"`
 
## Solution

It was determined that non-admin project owners SHOULD be able to perform this action, so the fix is to add the permission to the project-owner role that is bound to the user.
 
## Testing

The steps in the issue are valid, but an even easier way to see the problem is to skip the setting of resource quotas and to just try to edit the Pod Security Policy dropdown to a new value and then try to save the results.  If the rbac update worked properly, the user will no longer see the error.

### Manual Testing

I performed manual testing of this by modifying the Pod Security Policy and Adding/Removing/Editing resource quotas.  Several permutations of edits/non-edits/deletions/etc were performed over several iterations.

## QA Testing Considerations

I can't think of any special cases beyond the original reproducer and the steps I noted in the manual testing.

### DIFF
--- a/pkg/controllers/management/auth/manager.go
+++ b/pkg/controllers/management/auth/manager.go
@@ -361,7 +361,7 @@ func (m *manager) reconcileClusterMembershipBindingForDelete(roleToKeep, rtbNsAn
 		return rb.RoleRef.Name
 	}
 
-	return m.reconcileMembershipBindingForDelete("", roleToKeep, rtbNsAndName, m.crbIndexer, convert, m.mgmt.RBAC.ClusterRoleBindings("").ObjectClient())
+	return m.reconcileMembershipBindingForDelete("", roleToKeep, rtbNsAndName, convert, true)
 }
 
 // The PRTB has been deleted, either delete or update the project membership binding so that the subject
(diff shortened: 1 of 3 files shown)


KevinJoiner (changes requested):
Missing `prtb_handler_test.go`

crobby:
This PR is no longer needed

eumel8:
@crobby Why is this the case?

crobby:
It was decided that project owners should NOT have the ability to chage PSP for projects that they own.  PSP modifications are considered an admin function by design.

eumel8:
This makes sense. But I got still this error on project member management while adding a new account in Rancher 2.6.9:

```
error creating binding: Forbidden 403: podsecuritypolicytemplateprojectbindings.management.cattle.io is forbidden: User "u-uv5rdttdl4" cannot create resource "podsecuritypolicytemplateprojectbindings" in API group "management.cattle.io" in the namespace "c-fxzb9"
```

My idea was this would solved here.

PR #39536, opened 2022-11-08
Adding fallback env setting for base-ui-brand

crobby:
## Issue: 

https://github.com/rancher/rancher/issues/39456
 
## Problem

We need a flag to indicate whether or not a given install is a Rancher Prime install.  It needs to default to an env var that is set in the Dockerfile, but still be able to be set by users if they prefer to use their own branding.
 
## Solution

Introduce a new fallback env setting, base-ui-brand (not user-facing) that will pick-up the value of an env var (BASE_UI_BRAND) set by the Dockerfile at build time
 
### Manual Testing

Manually tested resultant image to verify that the setting

### DIFF
--- a/pkg/settings/setting.go
+++ b/pkg/settings/setting.go
@@ -153,7 +153,8 @@ var (
 	UIBanners = NewSetting("ui-banners", "{}")
 
 	// UIBrand High level 'brand' value, for example `suse`.
-	UIBrand = NewSetting("ui-brand", "")
+	// Fallback env, not a user-facing setting, used to indicate if this is a Prime install
+	UIBrand = NewSetting("ui-brand", os.Getenv("CATTLE_BASE_UI_BRAND"))
 
 	// UICommunityLinks displays community links in the UI.
 	// Deprecated in favour of UICustomLinks = NewSetting("ui-custom-links", "").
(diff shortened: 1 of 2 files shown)


pkg/settings/setting.go
```diff
@@ -155,6 +155,9 @@ var (
 	// UIBrand High level 'brand' value, for example `suse`.
 	UIBrand = NewSetting("ui-brand", "")
 
+	// Fallback env, not a user-facing setting, used to indicate if this is a Prime install
+	PrimeUIBrand = NewSetting("base-ui-brand", os.Getenv("BASE_UI_BRAND"))
+
 	// UICommunityLinks displays community links in the UI.
 	// Deprecated in favour of UICustomLinks = NewSetting("ui-custom-links", "").
 	UICommunityLinks = NewSetting("ui-community-links", "true")
```
cbron, line 163: Needs one more thing, the setting.go OS.env fallback check.

pkg/settings/setting.go
```diff
@@ -155,6 +155,9 @@ var (
 	// UIBrand High level 'brand' value, for example `suse`.
 	UIBrand = NewSetting("ui-brand", "")
 
+	// Fallback env, not a user-facing setting, used to indicate if this is a Prime install
+	PrimeUIBrand = NewSetting("base-ui-brand", os.Getenv("BASE_UI_BRAND"))
```
cbron, line 159: the getEnv check is correct, but this should be on the `ui-brand` setting, so we still use branding through that

crobby: ah, ok.  I slightly misunderstood you.  Will fix it up

cbron (approved):
lgtm

rmweir (approved):
LGTM

PR #39538, opened 2022-11-08
Unrc gke operator chart for kev2 v2.7.0

annablender:
## Issue: 

Unrc gke operator chart v1.1.5-rc1 -> v1.1.5
AKS and EKS are already unrc


rancherbot (approved):

kinarashah (approved):

PR #39548, opened 2022-11-09
Add extra parameter for source registry

stormqueen1990:
## Changes

* Include an extra parameter for the source registry in both rancher-save-images and rancher-load-images scripts to allow relocation from a customized registry.

## Problem

The current version of these scripts does not allow for specifying a custom source registry. These changes aim to allow users of these scripts to specify such a registry and relocate their images accordingly.
 
## Solution

* Add a new parameter to both the Bash and PowerShell scripts called `--source-registry/-s` to allow users to specify the registry where the images should be pulled from.
* Rename the former `--registry/-r` parameter to `--target-registry/-r` in order to make it clear this is the registry to where the images should be pushed. _Caveat: this is a breaking change and will impact user experience._
* Concatenate the source registry in the save scripts to pull images accordingly, as well as in the load scripts to tag images accordingly.
 
### Manual Testing

I was able to test only the Bash versions of these scripts.

#### Scenario 1
1. Created a rancher-images.txt file with only two images and ran the `rancher-save-images.sh` script, specifying no source registry.
2. Confirmed that the `tar.gz` was generated by `docker save`.
3. Ran the `rancher-load-images.sh` to load said images into a local registry (ran using `registry:2` as instructed in Docker documentation), also specifying no source registry.
4. Checked that the images were correctly loaded into the custom local registry.

#### Scenario 2
1. Created a rancher-images.txt file with only two images and ran the `rancher-save-images.sh` script, this time specifying a source registry.
2. Confirmed that the `tar.gz` was generated by `docker save`.
3. Ran the `rancher-load-images.sh` to load said images into a local registry (ran using `registry:2` as instructed in Docker documentation), specifying the same source registry.
4. Checked that the images were correctly loaded into the custom local registry.

## QA Testing Considerations

* Hosted registry (I did testing solely using my local containerized registry).
* PowerShell versions.


prachidamle:
Changes look good to me - except not sure if we should introduce the breaking change to rename the `--registry` parameter to `--target-registry`

stormqueen1990:
> Changes look good to me - except not sure if we should introduce the breaking change to rename the `--registry` parameter to `--target-registry`

@prachidamle makes sense, I'll revert it.

pkg/image/utilities/utilities.go
```diff
@@ -730,17 +774,27 @@ if ($help)
...
 if (-not (Test-Path $image_list))
 {
-	echo "Could not find '$image_list' file"
+    echo "Could not find '$image_list' file"
     usage
-	exit 1
+    exit 1
+}
+
+if ([string]::IsNull($source_registry))
```
slickwarren, line 782: this is giving me an error

pkg/image/utilities/utilities.go
```diff
@@ -730,17 +774,27 @@ if ($help)
...
+{
+    $source_registry = ""
+}
+elseif (-not [string]::IsNullOrEmpty($source_registry))
+{
+    $source_registry += "/"
```
slickwarren, line 788: this is not working in my setup
Screen Shot 2022-11-09 at 10 30 36 AM

slickwarren (changes requested):

pkg/image/utilities/utilities.go
```diff
@@ -485,14 +495,14 @@ for i in "${linux_images[@]}"; do
...
+        image_name="${target_registry}/${i}"
         ;;
     *)
-        image_name="${reg}/rancher/${i}"
+        image_name="${target_registry}/rancher/${i}"
         ;;
     esac
 
-    docker tag "${i}" "${image_name}${arch_suffix}"
+    docker tag "${source_registry}${i}" "${image_name}${arch_suffix}"
```
aiyengar2, line 505:
```suggestion
    docker tag "${source_registry}/${i}" "${image_name}${arch_suffix}"
```

if `--registry` is expected to not have a `/` at the end of it, so is `--source-registry`

pkg/image/utilities/utilities.go
```diff
@@ -541,15 +558,20 @@ if [[ $help ]]; then
     exit 0
 fi
 
+if [ ! -z "${source_registry}" ]; then
+    source_registry="${source_registry}/"
+fi
```
aiyengar2, line 563: same comment as above; we should either fail out with an invalid source registry or automatically trim the trailing slash if we see it in the input.

also, `target_registry` does not assume a trailing slash so neither should `source_registry`

pkg/image/utilities/utilities.go
```diff
@@ -541,15 +558,20 @@ if [[ $help ]]; then
...
+
 pulled=""
 while IFS= read -r i; do
     [ -z "${i}" ] && continue
+    i="${source_registry}${i}"
```
aiyengar2, line 568: same comment above; assume no trailing slash

```suggestion
    i="${source_registry}/${i}"
```

pkg/image/utilities/utilities.go
```diff
@@ -595,7 +619,11 @@ for ($i = $args.Length; $i -ge 0; $i--)
             $vals = $null
         }
         '^(-r|--registry)$' {
-            $registry = ($vals | Select-Object -First 1)
+            $target_registry = ($vals | Select-Object -First 1)
+            $vals = $null
+        }
+        '^(-r|--source-registry)$' {
```
aiyengar2, line 625: doesn't this imply `-r` is a shorthand for both `target_registry` and `source_registry`? `-r` will never match `--source-registry` here since it will exit above.

pkg/image/utilities/utilities.go
```diff
@@ -622,52 +650,62 @@ if ($help)
...
 Get-Content -Force -Path $image_list | ForEach-Object {
     if ($_) {
         $fullname_image = ('{0}-windows-{1}' -f $_, $os_release_id)
-		echo "Tagging $registry/$fullname_image"
-	
-		switch -regex ($fullname_image)
-		{
-			'.+/.+' {
-				docker tag $fullname_image $registry/$fullname_image
+        $source_image = -join ($source_registry, $fullname_image)
```
aiyengar2, line 691:
```suggestion
        $source_image = -join ($source_registry, "/", $fullname_image)
```

not sure if this is the right syntax (can `-join` take 3 args?) but as mentioned above assume `source_registry` does not have a trailing slash

aiyengar2: also, should we just add this into `$fullname_image`'s definition instead and continue to use that? Why modify the `$source_image`?

aiyengar2 (changes requested):

pkg/image/utilities/utilities.go
stormqueen1990: That will fail if the source registry is empty (i.e. Docker Hub).

pkg/image/utilities/utilities.go
stormqueen1990: Adding a slash here breaks the scenario where source registry is empty (i.e. Docker Hub).

stormqueen1990: The idea is to keep `$fullname_image` as it was originally so we don't need to parse registry information in the script, but rather just prepend the new registry name.

pkg/image/utilities/utilities.go
```diff
@@ -434,6 +441,10 @@ if [[ $help ]]; then
     exit 0
 fi
 
+if [ ! -z "${source_registry}" ]; then
```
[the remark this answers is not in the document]
stormqueen1990: I like omitting the trailing slash better than erroring out. Will update it.

pkg/image/utilities/utilities.go
stormqueen1990: Fixed!

pkg/image/utilities/utilities.go
stormqueen1990: Fixed!

slickwarren (approved):
tested in windows setup, all ps1 scripts are working in the following cases:
* specifying a source registry
* not specifying a source registry

pkg/image/utilities/utilities.go
stormqueen1990: Updated `$fullname_image` as requested.

pkg/image/utilities/utilities.go
```diff
@@ -622,52 +653,67 @@ if ($help)
...
-			}
-			default {
-				docker tag $fullname_image $registry/rancher/$fullname_image
-				if ($?) {
-                    docker push $registry/rancher/$fullname_image
+            }
+            default {
+                $target_image = -join ($target_registry, "rancher/", $fullname_image)
+                echo "Tagging $target_image"
+                docker tag $source_image $target_image
```
aiyengar2, line 711: does `$source_image` here also need to be modified to `-join ($source_registry, "rancher/", $fullname_image)`?

aiyengar2 (changes requested):
one last comment

pkg/image/utilities/utilities.go
stormqueen1990: In this case, the piece that the switch acts upon is already a part of the `$source_image` content:

https://github.com/rancher/rancher/blob/7eff43334bef7fa2423d611a14ed4d33dffeddc5/pkg/image/utilities/utilities.go#L696-L699

stormqueen1990: It cannot be modified as the original image reference already has the `rancher/` part (or not). Hence the `switch` statement to determine whether it should be tagged with `rancher/` or not.

aiyengar2: yes this makes sense, this seems to be how the script has been built for linux as well.

aiyengar2 (dismissed):
LGTM, you can squash commits and merge

pkg/image/utilities/utilities.go
stormqueen1990: Updating this discussion: the use case here is most likely the situation where an image is coming from a Docker Hub library repository, but we need to push it to a private repository (which will most likely require a project to do so).

aiyengar2 (approved):

slickwarren (approved):

prachidamle (commented):
LGTM!

PR #39562, opened 2022-11-09
Do not default OS/release Windows image tags in save/load image scripts

stormqueen1990:
## Problem

The current load and save scripts specify a `-windows-<release>` suffix when pulling Windows images. This does not work when we try to pull from an index instead of an individual image.
 
## Solution

Remove the `-windows-<release>` suffix from the tags so the script can pull images correctly.
 
### Manual Testing

1. In a Windows VM with Docker installed, ran the save script with 3 images in the list.
2. Verified images can be successfully downloaded after the change.
3. Ran the load script and verified images can still be found in the tarball after the change.


pkg/image/utilities/utilities.go
```diff
@@ -565,7 +565,6 @@ docker save $(echo ${pulled}) | gzip --stdout > ${images}
 $script_name = $MyInvocation.InvocationName
 $image_list = "rancher-windows-images.txt"
 $images = "rancher-windows-images.tar.gz"
-$os_release_id = $(Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\' | Select-Object -ExpandProperty ReleaseId)
 
 $registry = $null
 $help = $false
```
slickwarren, line 570: we should remove like 577 and 600-602 as the option won't have any effect

pkg/image/utilities/utilities.go
```diff
@@ -677,7 +676,6 @@ Get-Content -Force -Path $image_list | ForEach-Object {
 $script_name = $MyInvocation.InvocationName
 $image_list = "rancher-windows-images.txt"
 $images = "rancher-windows-images.tar.gz"
-$os_release_id = $(Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\' | Select-Object -ExpandProperty ReleaseId)
 
 $help = $false
```
slickwarren, line 681: we should remove like 686 and 705-707 as the option won't have any effect

slickwarren (changes requested):

pkg/image/utilities/utilities.go
stormqueen1990: Right, that makes sense. I'll update.

pkg/image/utilities/utilities.go
stormqueen1990: Removed.

slickwarren (dismissed):

HarrisonWAffel (approved):

aiyengar2 (changes requested):
@stormqueen1990 given @deniseschannon's comment offline, I believe the ask here is not to entirely remove the `--os-release-id` parameter but rather to avoid defaulting to using `$os_release_id = $(Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\' | Select-Object -ExpandProperty ReleaseId)` when it is not available.

That way, for anyone using that existing command line argument, the behavior will remain the same; however, for those who would like to use the airgap scripts in order to mirror **all** windows manifests (not just ones for the OS release that the private registry is detected to have), the behavior will be to mirror the base OCI artifact list, not a specific image.

Therefore, the fix would be something like changing the `$fullname_image` to using `$_` if `$os_release_id` is an empty string or nil, otherwise using the original logic if it has been set via the command line

aiyengar2 (approved):
LGTM, change the PR title before you merge it so that it notes the right action that is taken by the script though

slickwarren (approved):

stormqueen1990:
Closing this PR as we agreed that this will not fully address the issue at hand.

PR #39570, opened 2022-11-10
[2.6] Add post and pre upgrade resource validation tests

caliskanugur:
**Extensions Updates**
- Implemented:
  - A function to access an ingress externally to `extensions/ingresses/ingresses.go`

**Validation Tests**
- Implemented:
  - Post-upgrade and pre-upgrade resource validation cases
  - Private wait healthy response helper built on top of the access an ingresses helper
  - Private wait ingress hostname update helper built on top of the wait package
  - Private string prefix helpers to get an item and check an item in a slice of strings
  - Private name constructor to generate random suffixes with given names


igomez06 (approved):
Looks good!

Issue #39362, opened 2022-10-19
[BUG] Users are unable to create RKE2 vSphere Node Driver clusters with more than 1 Windows worker node


jameson-mcghee:
**Rancher Server Setup**
- Rancher version: v2.6.5 and v2.6.8
- Installation option (Docker install/Helm Chart): Docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): N/A
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: v1.24.6+rke2r1
- Cluster Type (Local/Downstream): Downstream
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): Node Driver (Infrastructure Provider)

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom): Standard User with Cluster Owner Role
  - If custom, define the set of permissions: N/A

**Describe the bug**
When attempting to create a vSphere Node Driver cluster with more than one Windows worker node or when attempting to scale up a second Windows worker node on an existing cluster, only the first Windows worker node will provision successfully.

**To Reproduce**
_Scenario 1:_
1. On a vSphere Hosted instance of Rancher, attempt to create a vSphere Node Driver cluster with 3 Linux ETCD nodes, 2 Linux Control Plane Nodes, 2 Linux Worker Nodes, and 2 Windows worker nodes

_Scenarios 2:_
1. On a vSphere Hosted instance of Rancher, create a vSphere Node Driver cluster with 3 Linux ETCD nodes, 2 Linux Control Plane Nodes, 2 Linux Worker Nodes, and 1 Windows worker node
2. Once the cluster becomes active, attempt to Scale up a Second Windows worker node

_Scenario 3:_
1. On a vSphere Hosted instance of Rancher, attempt to create a vSphere Node Driver cluster with 3 Linux ETCD nodes, 2 Linux Control Plane Nodes, 2 Linux Worker Nodes, 1 Windows worker node, and 1 Windows worker node (separate node pool) with the same OS Template as the first. 

_Scenario 4:_
[...]

PR #39599, opened 2022-11-15
[DNM v2.7.1] Bump machine to v0.15.0-rancher96

jakefhyde:
Issue: https://github.com/rancher/rancher/issues/39362

 
# Problem

When the vsphere node driver was updated to use `set_hostname` instead of `hostname`, the function used to set this field set it as a list, which would cause setting the hostname to fail.

# Solution

Append `set_hostname` to cloud-init userdata as a single field.

# Testing

### Manual Testing

I tested using a custom built machine image with this fix, and was able to provision over 7 different windows vms with no issue.
### Automated Testing

N/A
## QA Testing Considerations

It is required to set `netbios_host_name_compatibility` to false in the `C:\Program Files\Cloudbase Solutions\Cloudbase-Init\conf\cloudbase-init.conf`.
The vms must not have a hostname set prior to powering them on, otherwise the new hostname will not be used until reboot.
Vsphere templates provided to QA upon request.


HarrisonWAffel (approved):

snasovich (approved):

jakefhyde:
Closing as rancher machine has already been bumped to `v0.15.0-rancher97`

PR #39607, opened 2022-11-15
adding modular helm repo install

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/528
 
## Problem

airgap currently has no way to test various versions of rancher. only what is on rancher-latest 

## Solution

make the install variable
also updated the sed script so that we don't zip/unzip all the images (takes a lot of time and space that isn't necessary)

### DIFF
--- a/tests/validation/tests/v3_api/test_airgap.py
+++ b/tests/validation/tests/v3_api/test_airgap.py
@@ -257,10 +257,10 @@ def add_rancher_images_to_private_registry(bastion_node, push_images=True):
             RANCHER_SERVER_VERSION)
     bastion_node.execute_command(get_images_command)
 
-    # Remove the "docker save" and "docker load" lines to save time
+    # comment out the "docker save" and "docker load" lines to save time
     edit_save_and_load_command = \
-        "sudo sed -i '58d' rancher-save-images.sh && " \
-        "sudo sed -i '76d' rancher-load-images.sh && " \
+        "sudo sed -i -e 's/docker save /# docker/g' rancher-save-images.sh && " \
+        "sudo sed -i -e 's/docker load /# docker/g' rancher-load-images.sh && " \
         "chmod +x rancher-save-images.sh && chmod +x rancher-load-images.sh"
     bastion_node.execute_command(edit_save_and_load_command)
 
--- a/tests/validation/tests/v3_api/test_airgap_ha.py
+++ b/tests/validation/tests/v3_api/test_airgap_ha.py
@@ -29,6 +29,8 @@ HOST_NAME = os.environ.get('RANCHER_HOST_NAME', "testsa")
 AG_HOST_NAME = HOST_NAME
 RANCHER_AG_INTERNAL_HOSTNAME = AG_HOST_NAME + "-internal.qa.rancher.space"
 RANCHER_AG_HOSTNAME = AG_HOST_NAME + ".qa.rancher.space"
+RANCHER_HELM_REPO = os.environ.get("RANCHER_HELM_REPO", "latest")
+RANCHER_HELM_URL = os.environ.get("RANCHER_HELM_URL", "https://releases.rancher.com/server-charts/")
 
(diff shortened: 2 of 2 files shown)


tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -337,16 +339,19 @@ def setup_airgap_rancher(bastion_node, number_of_nodes=NUMBER_OF_INSTANCES):
     assert bastion_node.execute_command(
         '/snap/bin/helm')[0].find("Flags:") > -1, \
         "helm is not installed correctly"
+    
+    repo_name = "rancher-" + RANCHER_HELM_REPO
```
izaac, line 343: Can `stable`, `latest` and `alpha` be allowed ?

tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -226,8 +228,8 @@ def add_rancher_images_to_private_registry(
 
     # Remove the "docker save" and "docker load" lines to save time
```
izaac, line 229: nit. This needs to be changed to `Comment` not remove if this is now commenting the line.

izaac (changes requested):
A couple suggestions.

tests/validation/tests/v3_api/test_airgap_ha.py
slickwarren: yes, this allows them all to be entered by the user.  I will have a supporting jenkins job builder PR, but you can see the same logic in the ha_deploy job.

izaac: I see what's the goal now, to have a select dropdown with those values in the job builder.

izaac (approved):

tests/validation/tests/v3_api/test_airgap_ha.py
```diff
@@ -224,10 +226,10 @@ def add_rancher_images_to_private_registry(
             RANCHER_SERVER_VERSION)
     bastion_node.execute_command(get_images_command)
 
-    # Remove the "docker save" and "docker load" lines to save time
+    # comment out the "docker save" and "docker load" lines to save time
     edit_save_and_load_command = \
-        "sudo sed -i '58d' rancher-save-images.sh && " \
-        "sudo sed -i '76d' rancher-load-images.sh && " \
+        "sudo sed -i -e 's/docker save /# docker/g' rancher-save-images.sh && " \
+        "sudo sed -i -e 's/docker load /# docker/g' rancher-load-images.sh && " \
```
rancher-max, line 232: Same as above

rancher-max (approved):
This LGTM but small nit if you want to fix that (I'm fine either way)

tests/validation/tests/v3_api/test_airgap.py
```diff
@@ -257,10 +257,10 @@ def add_rancher_images_to_private_registry(bastion_node, push_images=True):
             RANCHER_SERVER_VERSION)
     bastion_node.execute_command(get_images_command)
 
-    # Remove the "docker save" and "docker load" lines to save time
+    # comment out the "docker save" and "docker load" lines to save time
     edit_save_and_load_command = \
-        "sudo sed -i '58d' rancher-save-images.sh && " \
-        "sudo sed -i '76d' rancher-load-images.sh && " \
+        "sudo sed -i -e 's/docker save /# docker/g' rancher-save-images.sh && " \
+        "sudo sed -i -e 's/docker load /# docker/g' rancher-load-images.sh && " \
```
rancher-max, line 263: NIT: To truly comment out, we should probably make these:
```
        "sudo sed -i -e 's/docker save /# docker save /g' rancher-save-images.sh && " \
        "sudo sed -i -e 's/docker load /# docker load /g' rancher-load-images.sh && " \
```
otherwise, in the `rancher-save-images.sh` for example, it will be like:
```
# docker$(echo ${pulled}) | gzip --stdout > ${images}
```

PR #39608, opened 2022-11-15
[2.6] updating airgap to allow helm

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/528
 
## Problem

airgap currently has no way to test various versions of rancher. only what is on rancher-latest 

## Solution

make the install variable
also updated the sed script so that we don't zip/unzip all the images (takes a lot of time and space that isn't necessary)
 
## Testing

the scripts were tested locally as part of another issue
the helm changes have been tested as part of the ha_deploy python job, as these changes are the same.


bmdepesa (approved):

izaac:
I've left comments in forward port.

izaac (dismissed):
Reviewed same code in the forward port PR

izaac (approved):
LGTM

PR #39614, opened 2022-11-16
[v2.7] Add support for hardening K3s custom clusters

markusewalker:
## Problem

Currently, the Go test framework allows for provisioning K3s custom clusters. By default, these are non-hardened clusters. In realistic customer environments, hardened clusters are at the forefront, so we need to be able to support this in our testing. This PR addresses this problem by providing the option to harden the custom K3s clusters.
 
## Solution

1. There needed to be an option available to select whether or not the user wishes to harden the cluster. This is done by adding a new `Hardened` bool type in the `config.go` struct.  Inside the user's `cattle-config.yaml`, they will have to define a new `hardened` value and set it to true or false.
2. In the `custom_clusters_test.go` file, new functions `CheckClusterHardening`, `HardenK3SRKE2ClusterConfig` and `UpdateK3SRKE2Cluster` are referenced. We will break each down in the subsequent steps, consequently.
3. `CheckClusterHardening` is found in the new `hardening/harden_nodes.go`. The point of the file is it will check the new `Hardened` value as mentioned in step 1. If user says yes, harden the node following appropriate hardening steps.
4. Once step 3 is finished, `HardenK3SRKE2ClusterConfig` will update the cluster's configuration to harden the cluster.
5. Once step 4 is finished, `UpdateK3SRKE2Cluster` will call upon the Steve client's update function to complete the changes to actually harden the K3s custom cluster.

As of now, this change only goes to the non-dynamic test function. This is because we would have to refactor the nodeRoles to use `machinepools.NodeRoles` and that was getting a bit messy when I was testing this locally. We can always change that if we want.
 
### Manual Testing

Tested against 1 role w/admin and standard user, 3 role w/admin and standard users.

### Automated Testing

### DIFF
--- a/tests/v2/validation/provisioning/config.go
+++ b/tests/v2/validation/provisioning/config.go
@@ -18,6 +18,7 @@ type Config struct {
 	CNIs               []string                 `json:"cni" yaml:"cni"`
 	Providers          []string                 `json:"providers" yaml:"providers"`
 	NodeProviders      []string                 `json:"nodeProviders" yaml:"nodeProviders"`
+	Hardened           bool                     `json:"hardened" yaml:"hardened"`
 }
 
 func AppendRandomString(baseClusterName string) string {
--- a/tests/v2/validation/provisioning/k3s/custom_cluster_test.go
+++ b/tests/v2/validation/provisioning/k3s/custom_cluster_test.go
@@ -10,6 +10,7 @@ import (
 	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
 	v1 "github.com/rancher/rancher/tests/framework/clients/rancher/v1"
 	"github.com/rancher/rancher/tests/framework/extensions/clusters"
+	hardening "github.com/rancher/rancher/tests/framework/extensions/hardening/k3s"
 	"github.com/rancher/rancher/tests/framework/extensions/machinepools"
 	"github.com/rancher/rancher/tests/framework/extensions/tokenregistration"
 	"github.com/rancher/rancher/tests/framework/extensions/users"
@@ -28,6 +29,7 @@ import (
 type CustomClusterProvisioningTestSuite struct {
 	suite.Suite
 	client             *rancher.Client
+	provisioning       *provisioning.Config
 	session            *session.Session
 	standardUserClient *rancher.Client
 	kubernetesVersions []string
@@ -47,6 +49,7 @@ func (c *CustomClusterProvisioningTestSuite) SetupSuite() {
 
 	c.kubernetesVersions = clustersConfig.KubernetesVersions
 	c.nodeProviders = clustersConfig.NodeProviders
+	c.provisioning = clustersConfig
 
 	client, err := rancher.NewClient("", testSession)
 	require.NoError(c.T(), err)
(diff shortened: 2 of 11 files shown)


tests/v2/validation/provisioning/k3s/custom_cluster_test.go
```diff
@@ -90,12 +93,13 @@ func (c *CustomClusterProvisioningTestSuite) ProvisioningK3SCustomCluster(extern
 	tests := []struct {
 		name      string
 		nodeRoles []string
+		hardening *provisioning.Config
```
slickwarren, line 96: can you add this var to the example readme so we know how to use it in the config.yaml?

tests/v2/validation/provisioning/k3s/custom_cluster_test.go
```diff
@@ -159,6 +163,15 @@ func (c *CustomClusterProvisioningTestSuite) ProvisioningK3SCustomCluster(extern
 				clusterToken, err := clusters.CheckServiceAccountTokenSecret(client, clusterName)
 				require.NoError(c.T(), err)
 				assert.NotEmpty(c.T(), clusterToken)
+
+				err = hardening.CheckClusterHardening(client, tt.hardening.Hardened, nodes, tt.nodeRoles)
+				require.NoError(c.T(), err)
+
+				hardenCluster := clusters.HardenK3SRKE2ClusterConfig(clusterName, namespace, "", "", kubeVersion, nil)
```
slickwarren, line 170: I'm curious as to how / why this doesn't run when hardened = false 🤔

slickwarren (commented):
overall LGTM, but couldn't review jenkins yet since it is down.

tests/v2/validation/provisioning/k3s/custom_cluster_test.go
markusewalker: Sure thing, that's a good idea. I'll go and make that change.

tests/v2/validation/provisioning/k3s/custom_cluster_test.go
markusewalker: In the other 2.6 PR, Israel had the same concern as we discussed offline, so I'll definitely take another look into this

markusewalker:
Feedback has been implemented and tested locally. I will provide reviewers with Jenkins test results sometime early next week so that we can proceed forward.

anupama2501 (dismissed):
LGTM

slickwarren (dismissed):
LGTM

markusewalker:
@anupama2501 @slickwarren as mentioned in my [comment](https://github.com/rancher/rancher/pull/39620#issuecomment-1329418695) in the 2.7 PR, I made a small update to a function name. Hence, why a re-request in the review. Nothing in the actual functionality changed from your last approvals.

anupama2501 (approved):

igomez06 (approved):
Looks good, but I forgot to ask, did figure out why this couldn't be added to the dynamic testing? PR the is good as it is now, but just curious.

tests/v2/validation/provisioning/config.go
```diff
@@ -18,6 +18,7 @@ type Config struct {
 	CNIs               []string                 `json:"cni" yaml:"cni"`
 	Providers          []string                 `json:"providers" yaml:"providers"`
 	NodeProviders      []string                 `json:"nodeProviders" yaml:"nodeProviders"`
+	Hardened           bool                     `json:"hardened" yaml:"hardened"`
```
igomez06, line 21: Quick question Markus, if Hardened isn't included in the config file at all, does it default to false?

markusewalker: Correct, I just double-checked to confirm. I removed from the config file and the cluster was not hardened at all.

igomez06: Awesome! Thank you!

izaac (approved):
LGTM

PR #39620, opened 2022-11-16
[v2.6] Add support for hardening K3s custom clusters

markusewalker:
## Problem

Currently, the Go test framework allows for provisioning K3s custom clusters. By default, these are non-hardened clusters. In realistic customer environments, hardened clusters are at the forefront, so we need to be able to support this in our testing. This PR addresses this problem by providing the option to harden the custom K3s clusters.
 
## Solution

1. There needed to be an option available to select whether or not the user wishes to harden the cluster. This is done by adding a new `Hardened` bool type in the `config.go` struct.  Inside the user's `cattle-config.yaml`, they will have to define a new `hardened` value and set it to true or false.
2. In the `custom_clusters_test.go` file, new functions `CheckClusterHardening`, `HardenK3SRKE2ClusterConfig` and `UpdateK3SRKE2Cluster` are referenced. We will break each down in the subsequent steps, consequently.
3. `CheckClusterHardening` is found in the new `hardening/harden_nodes.go`. The point of the file is it will check the new `Hardened` value as mentioned in step 1. If user says yes, harden the node following appropriate hardening steps.
4. Once step 3 is finished, `HardenK3SRKE2ClusterConfig` will update the cluster's configuration to harden the cluster.
5. Once step 4 is finished, `UpdateK3SRKE2Cluster` will call upon the Steve client's update function to complete the changes to actually harden the K3s custom cluster.

As of now, this change only goes to the non-dynamic test function. This is because we would have to refactor the nodeRoles to use `machinepools.NodeRoles` and that was getting a bit messy when I was testing this locally. We can always change that if we want.
 
### Manual Testing

Tested against 1 role w/admin and standard user, 3 role w/admin and standard users.

### Automated Testing

### DIFF
--- a/tests/framework/extensions/clusters/clusters.go
+++ b/tests/framework/extensions/clusters/clusters.go
@@ -221,6 +221,37 @@ func NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretN
 	return v1Cluster
 }
 
+// HardenK3SRKE2ClusterConfig is a constructor for a apisV1.Cluster object, to be used by the rancher.Client.Provisioning client.
+func HardenK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName, kubernetesVersion string, machinePools []apisV1.RKEMachinePool) *apisV1.Cluster {
+	v1Cluster := NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName, kubernetesVersion, machinePools)
+
+	v1Cluster.Spec.RKEConfig.MachineGlobalConfig.Data["kube-apiserver-arg"] = []string{
+		"enable-admission-plugins=NodeRestriction,PodSecurityPolicy,ServiceAccount",
+		"audit-policy-file=/var/lib/rancher/k3s/server/audit.yaml",
+		"audit-log-path=/var/lib/rancher/k3s/server/logs/audit.log",
+		"audit-log-maxage=30",
+		"audit-log-maxbackup=10",
+		"audit-log-maxsize=100",
+		"request-timeout=300s",
+		"service-account-lookup=true",
+	}
+
+	v1Cluster.Spec.RKEConfig.MachineSelectorConfig = []rkev1.RKESystemConfig{
+		{
+			Config: rkev1.GenericMap{
+				Data: map[string]interface{}{
+					"kubelet-arg": []string{
+						"make-iptables-util-chains=true",
+					},
+					"protect-kernel-defaults": true,
+				},
+			},
+		},
+	}
+
+	return v1Cluster
+}
+
 // CreateRKE1Cluster is a "helper" functions that takes a rancher client, and the rke1 cluster config as parameters. This function
 // registers a delete cluster fuction with a wait.WatchWait to ensure the cluster is removed cleanly.
 func CreateRKE1Cluster(client *rancher.Client, rke1Cluster *management.Cluster) (*management.Cluster, error) {
@@ -337,6 +368,47 @@ func CreateK3SRKE2Cluster(client *rancher.Client, rke2Cluster *apisV1.Cluster) (
 			return false, nil
 		})
 	})
+
+	return cluster, nil
+}
+
+// UpdateK3SRKE2Cluster is a "helper" functions that takes a rancher client, old rke2/k3s cluster config, and the new rke2/k3s cluster config as parameters.
+func UpdateK3SRKE2Cluster(client *rancher.Client, cluster *v1.SteveAPIObject, updatedCluster *apisV1.Cluster) (*v1.SteveAPIObject, error) {
+	updateCluster, err := client.Steve.SteveType(ProvisioningSteveResouceType).ByID(cluster.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	updatedCluster.ObjectMeta.ResourceVersion = updateCluster.ObjectMeta.ResourceVersion
+
+	logrus.Infof("Applying cluster YAML hardening changes...")
+	cluster, err = client.Steve.SteveType(ProvisioningSteveResouceType).Update(cluster, updatedCluster)
+	if err != nil {
+		return nil, err
+	}
+
+	err = kwait.Poll(500*time.Millisecond, 5*time.Minute, func() (done bool, err error) {
+		client, err = client.ReLogin()
+		if err != nil {
+			return false, err
+		}
+
+		clusterResp, err := client.Steve.SteveType(ProvisioningSteveResouceType).ByID(cluster.ID)
+		if err != nil {
+			return false, err
+		}
+
+		if clusterResp.ObjectMeta.State.Name == "active" {
+			logrus.Infof("Cluster YAML has successfully been updated!")
+			return true, nil
+		} else {
+			return false, nil
+		}
+	})
+
+	if err != nil {
+		return nil, err
+	}
 
 	return cluster, nil
 }
--- a/tests/framework/extensions/hardening/k3s/harden_nodes.go
+++ b/tests/framework/extensions/hardening/k3s/harden_nodes.go
@@ -1,0 +1,79 @@
+package hardening
+
+import (
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/rancher/rancher/tests/framework/pkg/nodes"
+	"github.com/sirupsen/logrus"
+)
+
+func HardeningNodes(client *rancher.Client, hardened bool, nodes []*nodes.Node, nodeRoles []string) error {
+	for key, node := range nodes {
+		logrus.Infof("Setting kernel parameters on node %s", node.NodeID)
+		_, err := node.ExecuteCommand("sudo setenforce 1")
+		if err != nil {
+			return err
+		}
+		_, err = node.ExecuteCommand("sudo bash -c 'echo vm.panic_on_oom=0 >> /etc/sysctl.conf'")
+		if err != nil {
+			return err
+		}
+		_, err = node.ExecuteCommand("sudo bash -c 'echo kernel.panic=10 >> /etc/sysctl.conf'")
+		if err != nil {
+			return err
+		}
+		_, err = node.ExecuteCommand("sudo bash -c 'echo kernel.panic_on_oops=1 >> /etc/sysctl.conf'")
+		if err != nil {
+			return err
+		}
+		_, err = node.ExecuteCommand("sudo bash -c 'echo kernel.keys.root_maxbytes=25000000 >> /etc/sysctl.conf'")
+		if err != nil {
+			return err
+		}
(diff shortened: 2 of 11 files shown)


tests/framework/extensions/clusters/clusters.go
```diff
@@ -340,3 +436,37 @@ func CreateK3SRKE2Cluster(client *rancher.Client, rke2Cluster *apisV1.Cluster) (
...
+		return nil, err
+	}
+
+	err = kwait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
+		client, err = client.ReLogin()
+		if err != nil {
+			return false, err
+		}
+
+		_, err = client.Steve.SteveType(ProvisioningSteveResouceType).ByID(cluster.ID)
```
igomez06, line 459: Would it be wise to make sure the cluster is in active state as well.
Doing something like 
```
clusterResp err = client.Steve.SteveType(ProvisioningSteveResouceType).ByID(cluster.ID)
clusterResp.ObjectMeta.State.Name == "Active"

```

just checking the condition above.

tests/framework/extensions/hardening/k3s/harden_nodes.go
```diff
@@ -0,0 +1,81 @@
+package hardening
+
+import (
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/rancher/rancher/tests/framework/pkg/nodes"
+	"github.com/sirupsen/logrus"
+)
+
+func CheckClusterHardening(client *rancher.Client, hardened bool, nodes []*nodes.Node, nodeRoles []string) error {
+	if hardened {
```
igomez06, line 10: If this is checking a hardened cluster, why do need this bool check?

igomez06: Would the checks be different from RKE2? Just curious, I do like the design for this.

tests/v2/validation/provisioning/k3s/custom_cluster_test.go
```diff
@@ -159,6 +163,15 @@ func (c *CustomClusterProvisioningTestSuite) ProvisioningK3SCustomCluster(extern
 				clusterToken, err := clusters.CheckServiceAccountTokenSecret(client, clusterName)
 				require.NoError(c.T(), err)
 				assert.NotEmpty(c.T(), clusterToken)
+
+				err = hardening.CheckClusterHardening(client, tt.hardening.Hardened, nodes, tt.nodeRoles)
+				require.NoError(c.T(), err)
+
+				hardenCluster := clusters.HardenK3SRKE2ClusterConfig(clusterName, namespace, "", "", kubeVersion, nil)
+
+				hardenClusterResp, err := clusters.UpdateK3SRKE2Cluster(client, clusterResp, hardenCluster)
```
igomez06, line 172: So basically if hardened is set to false, you will still go through these checks updates? Why don't you check for hardening at the test level, and then check and update accordingly. If it's not hardened.

igomez06 (commented):
Overall it looks good, a few questions that I have wrote comments. My main question is. Is the flow to update the nodes to be hardened, and then the cluster happen after the fact? I'm not too familiar so I'm just asking. We wouldn't hardened the nodes upon creation, and then provision the cluster as such, rather than update? Just a question to make sure this is the flow we want. Otherwise it looks great!!

tests/framework/extensions/hardening/k3s/harden_nodes.go
```diff
@@ -0,0 +1,81 @@
...
+			_, err = node.ExecuteCommand("sudo bash -c 'echo kernel.keys.root_maxbytes=25000000 >> /etc/sysctl.conf'")
+			if err != nil {
+				return err
+			}
+			_, err = node.ExecuteCommand("sudo bash -c 'sysctl -p /etc/sysctl.conf'")
+			if err != nil {
+				return err
+			}
+
+			if nodeRoles[key] == "--etcd --controlplane --worker" || nodeRoles[key] == "--controlplane" {
```
igomez06, line 38: Why wouldn't this work for the dynamic input? Could I get a little clarification. Thank you!

tests/framework/extensions/clusters/clusters.go
markusewalker: Yeah, that's a good idea. I'll make that change so we can make sure it's in a state of `Active` after the changes are applied.

tests/framework/extensions/hardening/k3s/harden_nodes.go
markusewalker: I think Anu is checking if it needs to be different for RKE2, as of now we don't believe it needs to be? I originally had that check to see if the `hardened` value in the user's config.yaml is true. If so, then it would proceed with the hardening , but I can definitely rework this if it seems a bit weird

tests/v2/validation/provisioning/k3s/custom_cluster_test.go
markusewalker: @slickwarren and I were talking of this as well, yeah I'll definitely look into that. Cause although it works for now, there isn't really an explanation behind it. Not a logical one at least lol...

tests/framework/extensions/hardening/k3s/harden_nodes.go
markusewalker: Oh, because the dynamic input uses `machinepools.NodeRoles`. The non-dynamic input defines the node doesn't, so one of two things would have to happen.

Either we convert the non-dynamic input to use `machinepools.NodeRoles` and I modify the file to work for both, or just run only the non-dynamic input. I chose the latter option just cause it was getting a bit hairy to do the former option.

I can definitely take another look at this if we think its better to have both for consistency sake

tests/v2/validation/provisioning/k3s/custom_cluster_test.go
igomez06: Yeah for sure, like why go through these checks and updates if hardened is set to false. Thank you!

tests/framework/extensions/hardening/k3s/harden_nodes.go
igomez06: https://github.com/rancher/rancher/blob/58627999b77516dc1e14e0294a14d9154fd60054/tests/v2/validation/provisioning/k3s/custom_cluster_test.go#L183-L196

markusewalker: Let me take a double-take on that , it's probably the case I was doing too much at once 😄 . Cause yeah, that seems like it should take without issue looking at it further. Thanks man!

igomez06: Thank you!!

markusewalker:
Feedback has been implemented and tested locally. I will provide reviewers with Jenkins test results sometime early next week so that we can proceed forward.

tests/framework/extensions/hardening/k3s/harden_nodes.go
```diff
@@ -0,0 +1,79 @@
...
+		_, err = node.ExecuteCommand("sudo bash -c 'echo kernel.keys.root_maxbytes=25000000 >> /etc/sysctl.conf'")
+		if err != nil {
+			return err
+		}
+		_, err = node.ExecuteCommand("sudo bash -c 'sysctl -p /etc/sysctl.conf'")
+		if err != nil {
+			return err
+		}
+
+		if nodeRoles[key] == "--etcd --controlplane --worker" || nodeRoles[key] == " --controlplane" {
```
anupama2501, line 37: There's a space in the value, should be `--controlplane`. 
`nodeRoles[key] == " --controlplane"`

tests/framework/extensions/clusters/clusters.go
```diff
@@ -221,6 +221,102 @@ func NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretN
 	return v1Cluster
 }
 
+// HardenK3SRKE2ClusterConfig is a constructor for a apisV1.Cluster object, to be used by the rancher.Client.Provisioning client.
+func HardenK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretName, kubernetesVersion string, machinePools []apisV1.RKEMachinePool) *apisV1.Cluster {
```
anupama2501, line 225: I'm just wondering, since we already have a Newk3sRKE2ClusterConfig.  
most of the config looks same except for a few values, should we just leverage the NewK3sRKE2Config instead of adding this function again.

igomez06: Great point. As not have to update all of our provisioning tests. Maybe HardenK3SRKE2ClusterConfig calls NewK3SRKE2ClusterConfig, and just updates machineGlobalConfigMap, and anything else. Which brings to a question I have had. Is the flow to hardened a cluster to do it after fact ie. update the nodes and clusters after they are created?

tests/framework/extensions/hardening/k3s/harden_nodes.go
markusewalker: Yeah, it's actually supposed to have the space funny enough. Since in the `custom_cluster_test.go` test, there is a space in each of the node roles in the dynamic input. So if you don't have a space in this conditional, then it's going to fail.

tests/framework/extensions/clusters/clusters.go
markusewalker: I can definitely look into that, ideally I would like minimal changes to `cluster.go` as well to not overly-complicate it. Yeah, so the flow to harden the cluster is as follows:

1. Harden the nodes after the K3s cluster is created.
2. Update the cluster's configuration to use hardened parameters as defined in step 1.
3. Apply the changes step 2 defined to actually harden the cluster.

markusewalker: Updated the function to leverage `NewK3SRKE2Config`, looks much cleaner! Going to resolve this conversation as a result.

anupama2501 (dismissed):
LGTM

slickwarren (dismissed):
LGTM

tests/framework/extensions/hardening/k3s/harden_nodes.go
```diff
@@ -0,0 +1,79 @@
+package hardening
+
+import (
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	"github.com/rancher/rancher/tests/framework/pkg/nodes"
+	"github.com/sirupsen/logrus"
+)
+
+func CheckClusterHardening(client *rancher.Client, hardened bool, nodes []*nodes.Node, nodeRoles []string) error {
```
igomez06, line 9: I'm not sure this naming makes sense. We're not necessary checking, but rather hardening the cluster. Correct?

markusewalker:
Thanks @igomez06, renamed the function based on your feedback! Both PRs have been updated.

anupama2501 (approved):

izaac (approved):
LGTM

Issue #39515, opened 2022-11-07
vSphere Cloud Credentials name by-cloud-cred does not exist


aleiner:
**Rancher Server Setup**
- Rancher version: 2.7.0-rc9
- Installation option: Helm Chart
   - Kubernetes Cluster and version: RKE2 v1.24.7+rke2r1
- Proxy/Cert Details: Cert-Manager

**Information about the Cluster**
- Kubernetes version: v1.24.7+rke2r1
- Cluster Type: Local

**User Information**
- What is the role of the user logged in?: Admin/Cluster
  - If custom, define the set of permissions:

**Describe the bug**
Error: "Index with name by-cloud-cred does not exist" when attempting to add new cloud credentials for vSphere.

**To Reproduce**
Cluster Management -> Cloud Credentials -> Add credentials -> Save

**Result**
"Index with name by-cloud-cred does not exist".  Credentials do get added, just not with the name provided.

**Expected Result**
Credentials get added with proper name an no error banner.

**Screenshots**
<img width="805" alt="image" src="https://user-images.githubusercontent.com/104371562/200334868-4b14369e-e87f-44ef-836f-913da6c34b7a.png">

**Additional context**

annablender:
@jameson-mcghee I just tried and was also unable to reproduce the issue on a rancher instance v2.6.9 in a vSphere environment. I believe when I saw the issue, my rancher instance was not behind the required vpn which caused connection issues and that error.

jameson-mcghee:
**Reproduction attempt test information:**

**Rancher Server Setup**
- Rancher version: `v2.7.0-rc8` and `v2.6.9`.
- Installation option (Docker install/Helm Chart): Docker
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): N/A
- Proxy/Cert Details: N/A

**Information about the Cluster**
- Kubernetes version: N/A
- Cluster Type (Local/Downstream): N/A

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Admin and Standard

**Steps followed during the attempt to Reproduce**

1. Spin up a Rancher server on `v2.6.9`, and navigate to the Cluster Management --> Cloud Credentials page
2. Attempt to create VMware vSphere credentials with correct credential information

**Note:** I was unable to reproduce the reported issue on our test vSphere environment.

[...]

PR #39622, opened 2022-11-16
[v2.7.1] Ensure ProvV2 indexers are registered for all replicas

HarrisonWAffel:
## Problem
When deleting cloud credentials an error would intermittently appear stating `Index with name by-cloud-cred does not exist`, and the cloud credential would not be deleted. This issue only reproduces in HA setups. 

## Solution
The root cause of this issue is, as the error mentions, a missing indexer in the provisioning cluster cache. This indexer is used to determine if any provisioning clusters are currently using the cloud credential being deleted. The reason that this issue only appears in HA setups is that only the leader pod will actually register the indexer. In cases where there is only one replica, or the Rancher server is being run in docker/IDE, this issue will not reproduce.

The indexer is registered as apart of other prov v2 controllers, which are only active on the leader pod. When the UI makes the request to delete the cloud credential it is not guaranteed that the request will always go to the leader pod.  In cases where the request is routed to a non-leader pod, this error will occur. 

The solution to this is to have all replicas register the indexer upon server start if the provisioning v2 feature is enabled. I've also fixed some casing issues and made the error message returned during a generic cache error more user friendly. 

## Testing
+ Create an HA cluster
+ Create AWS cloud credentials
+ Create an RKE2 cluster on AWS
+ Once the RKE2 cluster has become available, attempt to delete the cloud credentials 
+ Observe that the `index with name by-cloud-cred does not exist` error does not intermittently appear, and instead the message `Cloud credential is currently referenced by provisioning cluster XXX` appears

### Manual Testing
I've built a custom image of Rancher with my changes and used it in an HA setup, I did not see the error appear anymore. 

### Automated Testing
N/A

## QA Testing Considerations
This requires an HA setup. When reproducing the issue the error should appear more frequently when more Rancher replica's are added to the deployment.
 
None I can think of

### DIFF
--- a/pkg/api/norman/customization/cred/store.go
+++ b/pkg/api/norman/customization/cred/store.go
@@ -84,12 +84,15 @@ type Store struct {
 }
 
 func (s *Store) Delete(apiContext *types.APIContext, schema *types.Schema, id string) (map[string]interface{}, error) {
+	// make sure the credential isn't being used by an active RKE2/K3s cluster
 	if provClusters, err := s.ProvClusterCache.GetByIndex(cluster.ByCloudCred, id); err != nil {
-		return nil, err
+		return nil, httperror.NewAPIError(httperror.ServerError, fmt.Sprintf("An error was encountered while attempting to delete the cloud credential: %s", err))
 	} else if len(provClusters) > 0 {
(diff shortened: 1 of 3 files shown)


jakefhyde (approved):

annablender (approved):
LGTM, I believe this fixes https://github.com/rancher/rancher/issues/39515 as well

Issue #39401, opened 2022-10-25
[BUG] After v2.6.8 upgrade, high CPU/load on nodes and database, nodes flapping between ready/unready, excessive "refresh processing" logs


samjustus:
Rancher Cluster:
Rancher version: 2.6.8 (after upgrade from 2.6.6)
Number of nodes: 2
Node OS version: SLES 15 SP4

**Describe the bug**
After upgrade (around 10AM PDT) from v2.6.6 to v2.6.8, CPU/load on nodes and RDS MySQL database has been high. Nodes flapping between Ready/Not Ready

Business impact:
Rancher is availablility is flapping. Unable to meet uptime SLA

Troubleshooting steps:
Looked at Rancher logs. When running in DEBUG, seeing this continuously:

**To Reproduce**
Unknown, happened after v2.6.6 to v2.6.8 upgrade

**Result**
High CPU/load (~ 100%), Rancher unavailable.

**Expected Result**
Rancher running at 10-20% CPU and up.

**Additional context**
logs in jira issue

SURE-5323

rmweir:
Possibly related to https://github.com/rancher/rancher/pull/38193.

MbolotSuse:
Got some additional notes for a test case that we should include for this fix when ready.

This issue can also cause etcd to run out of memory in certain circumstances. Repro steps below:

1. Create a k8s cluster for rancher HA. Ensure that there are 3 nodes, RKE, each node all roles, 16 core CPU 64GB memory (digital ocean slug `g-16vcpu-64gb`) was used in testing.
2. Install rancher 2.6.6. Complete basic setup (provide bootstrap, set new password).
3. Create 6,000 tokens (script available). Tokens are used to slow down the refresh process slightly since the local provider is being used (which can complete refresh quicker than other providers). 
4. Create 700 users and user attributes (script available). Ensure that the user attributes are sufficiently large (I used large annotations to achieve this).
5. Upgrade rancher to 2.6.9 (I used terraform and just changed the version from 2.6.6 -> 2.6.9).
6. Wait and observe etcd memory usage, you will run out of memory soon (you can also increase log verbosity to observe the constant user refresh messages in the logs).

brudnak:
# ✅ PASSED

## Reproduction Environment


| Component                        | Version / Type                           |
| -------------------------------- | ---------------------------------------- |
| Rancher version                  | 2.6.6                                    |
| Installation option              | Helm (high availability)                 |
[...]

PR #39624, opened 2022-11-16
Fix userattribute

rmweir:
## Issue: 

https://github.com/rancher/rancher/issues/39401
 
## Problem

Before, a userattribute handler would check if the ExtraByProvider
field was null and trigger a refresh. ExtraByProvider contains
data that is available only on tokens. When there was no token
associated with the user corresponding to the userattribute, the
handler would fire endlessly. This happened because the field
continued to be null but the refresh process would update the
LastRefreshed field which would cause the userattribute handler
to execute again.

## Solution

Now, the ExtraByProvider check is done on a token
handler. The token handler is aware that the tokens existence
may make populating the field possible and the refresh process,
makes no changes to the token object. This avoids nonstop firing
of handlers.
 
## Testing

1. Create a user in v2.6.2

2. Delete tokens for the user. If you just made it and logged into it then there is probably only one that was made for UI session.

3. Upgrade to v2.6.7 with log level set to debug.

Result with bug:
You will see many logs of the format "[DEBUG] Finished refresh process for <user>". There will also be a performance impact but that is more difficult to measure, so go by logs.

Expected behavior:
Will see much less of the aforementioned logs, probably 1 or 2 per user.

### Manual Testing

I followed the steps above.

### Automated Testing

Added unit tests.

Anything involving auth so making requests after login.

### DIFF
--- a/pkg/controllers/management/auth/token.go
+++ b/pkg/controllers/management/auth/token.go
@@ -36,16 +36,18 @@ func (t *TokenController) sync(key string, obj *v3.Token) (runtime.Object, error
 	// remove legacy finalizers
 	if obj.DeletionTimestamp != nil {
 		finalizers := obj.GetFinalizers()
+		newObj := obj.DeepCopy()
 		for i, finalizer := range finalizers {
 			if finalizer == "controller.cattle.io/cat-token-controller" {
 				finalizers = append(finalizers[:i], finalizers[i+1:]...)
-				newObj := obj.DeepCopy()
+				newObj = obj.DeepCopy()
 				newObj.SetFinalizers(finalizers)
 				var err error
-				obj, err = t.tokens.Update(newObj)
+				newObj, err = t.tokens.Update(newObj)
 				if err != nil {
-					return nil, err
+					return obj, err
 				}
+				obj = newObj
 				break
 			}
 		}
@@ -54,22 +56,24 @@ func (t *TokenController) sync(key string, obj *v3.Token) (runtime.Object, error
 	if obj.TTLMillis != 0 && obj.ExpiresAt == "" {
 		//compute and save expiresAt
 		newObj := obj.DeepCopy()
+		var err error
 		tokenUtil.SetTokenExpiresAt(newObj)
-		if _, err := t.tokens.Update(newObj); err != nil {
-			return nil, err
+		if newObj, err = t.tokens.Update(newObj); err != nil {
+			return obj, err
 		}
+		obj = newObj
 	}
 
 	// trigger corresponding UserAttribute resource to refresh if token potentially
 	// provides new information that is missing from the UserAttribute resource
 	refreshUserAttributes, err := t.userAttributesNeedsRefresh(obj.UserID)
 	if err != nil {
-		return nil, err
+		return obj, err
 	}
 
 	if refreshUserAttributes {
 		if err = t.triggerUserAttributesRefresh(obj.UserID); err != nil {
-			return nil, err
+			return obj, err
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


cmurphy (dismissed):

MbolotSuse (commented):
One comment/question on the use of the `token.UserId` as a means of identifying which user needs a refresh - wondering if we need some sort of `getUserFromToken` method which checks the various user-identifying fields on the tokens and picks one in a priority order.

pkg/controllers/management/auth/token_test.go
```diff
@@ -0,0 +1,210 @@
...
+		assert.Equal(t, testcase.expectedOutputToken, returnToken, fmt.Sprintf("%s", testcase.inputToken.Name))
+		if testcase.inputUserAttribute == nil {
+			continue
+		}
+		returnUserAttribute, _ := testTokenController.userAttributesLister.Get("", testcase.inputUserAttribute.Name)
+		assert.Equal(t, testcase.expectedOutputUserAttribute, returnUserAttribute, fmt.Sprintf("%s", testcase.inputToken.Name))
+	}
+}
+
+func populateTestCases(tokens map[string]*v3.Token, userAttributes map[string]*v3.UserAttribute) []tokenTestCase {
```
MbolotSuse, line 87: One test nit: I think that you got all of the standard test cases, but did not include any error test cases (error when updating after removing finalizer, error when updating after hashing, error when refreshing attribute) and the obj == nil case. Since these are of questionable value (I got an error I return an error), you don't have to add them, but I thought I would leave a comment on it just in case.

MbolotSuse (commented):
One test nitpick.

pkg/controllers/management/auth/token.go
```diff
@@ -49,9 +56,24 @@ func (n *TokenController) sync(key string, obj *v3.Token) (runtime.Object, error
...
-			return nil, err
+		if newObj, err = t.tokens.Update(newObj); err != nil {
+			return obj, err
+		}
+		obj = newObj
+	}
+
+	// trigger corresponding UserAttribute resource to refresh if token potentially
+	// provides new information that is missing from the UserAttribute resource
+	refreshUserAttributes, err := t.userAttributesNeedsRefresh(obj.UserID)
```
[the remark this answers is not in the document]
rmweir: The ExtraByProvider part of the userattributes handler only considers UserID https://github.com/rancher/rancher/blob/release/v2.7/pkg/auth/providerrefresh/refresher.go#L180.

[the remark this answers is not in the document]
MbolotSuse: Good point, thanks for pointing that out.

MbolotSuse (approved):
LGTM

pkg/controllers/management/auth/token_test.go
MbolotSuse: These were all added besides the obj==nil case. Resolving.

MbolotSuse (approved):
LGTM

pkg/controllers/management/auth/token_test.go
```diff
@@ -0,0 +1,373 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type tokenTestCase struct {
+	inputToken                  *v3.Token
+	expectedOutputToken         *v3.Token
+	inputUserAttribute          *v3.UserAttribute
+	expectedOutputUserAttribute *v3.UserAttribute
+	enableHashing               bool
+	description                 string
```
cmurphy, line 25: I don't see this description being used directly by the test, could it be omitted? The description could be added as a comment next to the test case.

rmweir: I instead added an error message when a test case fails using the description. Is that okay?

cmurphy: :+1:

cmurphy (approved):

PR #39631, opened 2022-11-17
Refactor logic for update README script to make it easier to add versions

aiyengar2:


aiyengar2:
Addresses a bug introduced in https://github.com/rancher/rancher/pull/39627 where 2.7.0 was marked latest but 2.6.9 was still also marked latest

superseb (approved):

PR #39633, opened 2022-11-17
Automation to allow rancher2 provider TF cluster resources to be tested with Terratest + Go

Josh-Diamond:
Automation to allow rancher2 provider TF cluster resources to be tested with Terratest + Go:
- refactored away from static .tfvars files to be dynamically created/destroyed with each test; 
- refactored away from config.go to supply configurations via yml

## Issue: 
Automate tests for terraform rancher2 provider clusters' for terraform releases
 
## Problem
No automation exists for this
 
## Solution
Created a framework w/ Go to allow testing automation of rancher2 provider cluster resources
 
## Testing
All tests have been thoroughly tested during development.

GKE module has been implemented, however, when implementing future GKE tests, it will be necessary to observe behavior, and modify functions to account for GKE module. [e.g. `waitForActiveCluster()`]

### Manual Testing
N/A

### Automated Testing
Ran each test, ensuring expected behavior and accurate results

## QA Testing Considerations
N/A
 
None


Josh-Diamond:
To address https://github.com/rancher/qa-tasks/issues/386

tests/terratest/components/aksClusterSpecs1.go
```diff
@@ -0,0 +1,5 @@
+package components
+
+var AKSClusterSpecs1 = `"
```
slickwarren, line 3: are all these generated by go + TF? If not, why can't all these vars be in one file, say per provider?

tests/terratest/functions/set_modules/setEC2RKE1.go
```diff
@@ -0,0 +1,43 @@
...
+	for _, pool := range nodePools {
+		poolNum := strconv.Itoa(num)
+		quantity := strconv.Itoa(pool.Quantity)
+		if pool.Etcd != "true" && pool.Cp != "true" && pool.Wkr != "true" {
+			return false, fmt.Errorf(`no roles selected for pool` + poolNum + `; at least one role is required`)
+		}
+		if pool.Quantity <= 0 {
+			return false, fmt.Errorf(`invalid quantity specified for pool` + poolNum + `; quantity must be greater than 0`)
+		}
+		poolConfig = poolConfig + components.RKE1NodePoolPrefix + poolNum + components.RKE1NodePoolSpecs1 + poolNum + components.RKE1NodePoolSpecs2 + poolNum + `-"` + components.RKE1NodePoolSpecs25 + quantity + components.RKE1NodePoolSpecs3 + pool.Cp + components.RKE1NodePoolSpecs4 + pool.Etcd + components.RKE1NodePoolSpecs5 + pool.Wkr + components.RKE1NodePoolSuffix
```
slickwarren, line 26: (see question about autogenerated files for more context) - The naming of these is not intuitive for me. I would expect some comment about what these are so that another user can come back and understand what they do, as we will need to maintain this part of the code. What is `RKE1NodePoolSpecs25` vs `RKE1NodePoolSpecs2`? even if I go look at those files, I get
```
var RKE1NodePoolSpecs25 = `
  node_template_id = rancher2_node_template.rancher2_node_template.id
  quantity         = ` 
```
and 
```
var RKE1NodePoolSpecs2 = `"
  hostname_prefix  = "${var.hostname_prefix}-tf-pool`
```
or in other words, Whoever has to go back and maintain this shouldn't have to do so much tracking to figure out what the variable does. Is there a way to name these in a more friendly way, or if not at least comment what this is doing?

tests/terratest/functions/set_modules/setHostedModules.go
```diff
@@ -0,0 +1,35 @@
+package functions
```
slickwarren, line 1: The name of this file doesn't match what was set for non-hosted modules. Can you update either this name to not have `modules` in the filename, or update the others to have `modules` in their filenames?

slickwarren (changes requested):
The overall work here is great, Josh! I have a few questions:
* can you add a readme at the root of the tests/terratest/tests/ folder describing what the config.yaml will look like when running the go tests? see https://github.com/rancher/rancher/tree/release/v2.7/tests/v2/validation/provisioning#readme for an example
* the video you posted in slack was very helpful in understanding how it will be used
* I noted this in a separate comment but wanted to talk more about it here. The structure of the tests should follow that of other tests we have if possible.

tests/terratest/functions/set_modules/setHostedModules.go
Josh-Diamond: These names can be updated.

The functions to set the actual module itself will not include `module` in the function name (e.g. `SetAKS()` )

For context, SetConfigTF() is passed in a module, and depending on that module, will call either SetHostedModules() or SetNodeDriverModules()... and they do just that.. SetHostedModules() takes in the module passed in from SetConfigTF()  and depending on that value, will call either SetAKS(), SetEKS(), or SetGKE().  SetNodeDriverModules() do the same, but for Node driver clusters.

`SetConfigTF()`:
- `SetHostedModules()`:
  - `SetAKS()`
  - `SetEKS()`
  - `SetGKE()`
- `SetNodeDriverModules()`:
  - `SetEC2RKE1()`
  - `SetEC2RKE2K3s()`
  - `SetLinodeRKE1()`
  - `SetLinodeRKE2K3s()`   
  
I'm open to changing the names here, and based on your input, would agree we could drop `Modules` for these two functions, and just have them SetHosted() and SetNodeDriver().. although, in a way, its less specific to what the function is actually doing.

Maybe SetHostedModule() and SetNodeDriverModule() would be most specific, as they only set one module at a time 🤔 

Thoughts on naming convention here?

tests/terratest/tests/node_driver/ec2_k3s_kubernetes_upgrade_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	cluster, err := client.Management.Cluster.ByID(clusterID)
+	require.NoError(t, err)
+
+	// Test cluster
+	assert.Equal(t, clusterName, cluster.Name)
+	assert.Equal(t, provider, cluster.Provider)
+	assert.Equal(t, active, cluster.State)
+	assert.Equal(t, clusterConfig.KubernetesVersion, cluster.Version.GitVersion)
+
```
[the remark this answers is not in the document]
Josh-Diamond: Yes, this can absolutely be broken into a function.  This is a great suggestion!  I would love to implement this!

tests/terratest/functions/waitForActiveCluster.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			if cluster.State == "active" {
+				return true, nil
+			}
+
+			return false, nil
+		})
+		require.NoError(t, err)
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
Josh-Diamond: I'm using kwait, but in addition, needed to use sleeps to slow the test execution down, and give rancher server a change to "breathe".  Without the sleeps, the data fetched immediately after is not accurate and the tests will fail.  K3s in particular needed a drastic sleep (10 seconds) in order to be reliable across all tests, and this function is shared by all modules, so this was a tricky piece.  I know its not elegant, and would like suggestions if this needs to be removed.

tests/terratest/functions/waitForActiveCluster.go
```diff
@@ -0,0 +1,143 @@
...
```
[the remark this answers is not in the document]
Josh-Diamond: testing reliability

tests/terratest/tests/hosted/eks_provision_and_scale_test.go
```diff
@@ -0,0 +1,194 @@
...
+		terraform.Destroy(t, terraformOptions)
+		functions.CleanupConfigTF(module)
+		functions.CleanupVarsTF(module)
+	}
+
+	// Deploys [main.tf] infrastructure and sets up resource cleanup
+	defer cleanup()
+	terraform.InitAndApply(t, terraformOptions)
+
+	time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
Josh-Diamond: testing reliability

tests/terratest/functions/waitForActiveCluster.go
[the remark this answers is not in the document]
izaac: I experienced this behavior too building the long living Ranchers

tests/terratest/components/aksClusterSpecs1.go
Josh-Diamond: These were tricky.  These were all created by me, and allows Go to build out the modules main.tf file. 

All files located in the components folder, hold Terraform Rancher2 provider resource blocks.  When SetConfigTF() is called, depending on the module, ultimately a giant string literal is composed of select files from the components folder, are built out and written to modules main.tf file.

In some cases, the entire block was able to be grabbed, and naming convention was appropriate (e.g. `provider.go`)  For some, I needed to break the blocks up, to interpolate values to allow them to be updated.  Those values come from the cattle-config.yaml.  In these instances, I didn't know what to call the blocks, as i tried break them up minimally, so used a number system.

Suggestions to improve this would be appreciated.

tests/terratest/functions/set_modules/setEC2RKE1.go
Josh-Diamond: Id love to discuss this further during PR sync up.  These are essentially building out the main.tf files.  These won't need to be maintained unless the rancher2 provider resource syntax changes.  However, with the growth of this repo, it would be possible to add new fields or values to these blocks, so implementing a better system around this would be ideal.

tests/terratest/functions/set_modules/setHostedModules.go
slickwarren: ok I see.  It makes sense when you lay it out like that. But the file structure doesn't match the function bullets that you laid out, which is why I was confused. everything is in 1 `set_modules` folder, so it was hard to tell that those do different things from looking at it. But this might be fine as-is, lets get another round of review before changing anything.

Josh-Diamond:
w.r.t. `can you add a readme at the root of the tests/terratest/tests/ folder describing what the config.yaml will look like when running the go tests?`

Yes! I have updated internal documentation on this, but will add a README.md with these configurations defined as well!

tests/terratest/functions/waitForActiveCluster.go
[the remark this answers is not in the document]
igomez06: So I'm not a fan of the sleeps either. I think having the data fetches done in a poll instead, would be a much more elegant design and ultimately much more reliable. Does that make sense? That goes for all the sleeps

tests/terratest/components/aksClusterSpecs1.go
Josh-Diamond: Ahhh, re-reading your initial comment, these are only grabbed and referenced by go, so yes, we absolutely could organize this better and place them in files per provider. 👍 Great idea!

tests/terratest/functions/waitForActiveCluster.go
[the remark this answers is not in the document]
Josh-Diamond: Okay, maybe i'll have it additionally check condition for ready to be true, and see how that works out.  I agree, the sleeps are awful, I will work on removing them 👍

Josh-Diamond:
Upon further review by the team, it has been decided to remove static, hardcoded components and set functions, to be replaced with one of the following:
-  [hclwrite](https://pkg.go.dev/github.com/hashicorp/hcl/v2/hclwrite) w/ terratest
- corral - [to provide rancher2 provider terraform resources]

Josh-Diamond:
This PR became a conflicted mess and so i've closed this out.

New draft has been submitted for review, here:
https://github.com/rancher/rancher/pull/39945

PR #39637, opened 2022-11-17
merge some code logic

lucming:


lucming:
https://github.com/rancher/rke/blob/release%2Fv1.4/cluster/plan.go#L106~L156
this is logic of generate `process`,sidekick is the first `process`,so we don't need two loops to restrict the sidekick container to run first

PR #39647, opened 2022-11-18
[v2.6] Prevent user from re-using the bootstrap password

crobby:
Prevent user from re-using the bootstrap password

## Problem

The existing bootstrap password could be reused when the password is changed for the first time.
 
## Solution

We now compare the existing password with the proposed new password, if they match, we reject it.
 
### Manual Testing

Tried re-using bootstrap password as well as other current password, both rejected as designed.

### Automated Testing

Added unit test to verify this funcitonality

## QA Testing Considerations

On the first visit to the dashboard, login as usual to the dashboard with the bootstrap password
When prompted to change the password, try re-using the same (bootstrap) password. It should be rejected with the message, The new password must not be the same as the current password
Try changing the password to a valid password value (either user-typed or the randomly generated password). It should succeed.
From the Account & API page, try changing the password. Try setting it to the same value as the current password. It should be rejected with the message, The new password must not be the same as the current password
From the Account & API page, try changing the password. Set it to a valid value that is NOT the same as the current password. It should succeed.

### DIFF
--- a/pkg/auth/api/user/user_actions.go
+++ b/pkg/auth/api/user/user_actions.go
@@ -95,7 +95,7 @@ func (h *Handler) changePassword(actionName string, action *types.Action, reques
 		return err
 	}
 
-	if err := validatePassword(user.Username, newPass, settings.PasswordMinLength.GetInt()); err != nil {
+	if err := validatePassword(user.Username, currentPass, newPass, settings.PasswordMinLength.GetInt()); err != nil {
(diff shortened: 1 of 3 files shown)


cbron:
In description above can you add a link to the original PR

cbron (approved):

MbolotSuse (approved):
LGTM

PR #39665, opened 2022-11-21
[2.7]Adding additional test cases to the rbac suite

anupama2501:
Same as the problem statement for https://github.com/rancher/rancher/pull/39666 

Task: https://github.com/rancher/qa-tasks/issues/538

### DIFF
--- a/tests/v2/validation/rbac/rbac_test.go
+++ b/tests/v2/validation/rbac/rbac_test.go
@@ -12,6 +12,7 @@ import (
 	v1 "github.com/rancher/rancher/tests/framework/clients/rancher/v1"
 	"github.com/rancher/rancher/tests/framework/extensions/clusters"
 	"github.com/rancher/rancher/tests/framework/extensions/namespaces"
+	"github.com/rancher/rancher/tests/framework/extensions/projects"
 	"github.com/rancher/rancher/tests/framework/extensions/users"
 	"github.com/rancher/rancher/tests/framework/pkg/session"
 	provisioning "github.com/rancher/rancher/tests/v2/validation/provisioning"
@@ -23,14 +24,17 @@ import (
 
 type RBTestSuite struct {
 	suite.Suite
-	client             *rancher.Client
-	standardUser       *management.User
-	standardUserClient *rancher.Client
-	session            *session.Session
-	cluster            *management.Cluster
-	adminProject       *management.Project
-	steveAdminClient   *v1.Client
-	steveStdUserclient *v1.Client
+	client                *rancher.Client
+	standardUser          *management.User
+	standardUserClient    *rancher.Client
+	session               *session.Session
+	cluster               *management.Cluster
+	adminProject          *management.Project
+	steveAdminClient      *v1.Client
+	steveStdUserclient    *v1.Client
+	additionalUser        *management.User
+	additionalUserClient  *rancher.Client
+	standardUserCOProject *management.Project
 }
 
 func (rb *RBTestSuite) TearDownSuite() {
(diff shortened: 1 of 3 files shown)


tests/v2/validation/rbac/rbac_test.go
```diff
@@ -183,6 +187,64 @@ func (rb *RBTestSuite) ValidateRemoveClusterRoles(role string) {
 
 }
 
+func (rb *RBTestSuite) ValidateAddStdUserAsProjectOwner(role string) {
```
caliskanugur, line 190: It seems like this `role` arg isn't in use, for the new private methods.

tests/v2/validation/rbac/rbac_test.go
```diff
@@ -263,6 +325,50 @@ func (rb *RBTestSuite) TestRBAC() {
...
+
+	rb.Run("Additional testcase2 - Validating if member with role "+roleOwner+" can add another standard user as a cluster owner", func() {
+		rb.ValidateAddMemberAsClusterRoles(roleOwner)
+
+	})
+
+	rb.Run("Additional testcase3 - Validating if member with role "+roleOwner+" can add a cluster member as a project owner", func() {
+		rb.ValidateAddCMAsProjectOwner(roleOwner)
+
+	})
```
caliskanugur, line 368: So Anupama, I don't understand the reason that we don't add those steps to the above table (the existing test) with an `if tt.clusterRole == roleOwner` wrapper and include the new runs after it instantiates the additional user in the condition scope. Wouldn't be less test to maintain?

caliskanugur (commented):
A few nits and a question, looks good

tests/v2/validation/rbac/rbac_test.go
```diff
@@ -183,6 +187,64 @@ func (rb *RBTestSuite) ValidateRemoveClusterRoles(role string) {
...
 
+func (rb *RBTestSuite) ValidateAddStdUserAsProjectOwner(role string) {
+
+	//Create project as a cluster Owner
+	createProjectAsCO, err := createProject(rb.client, rb.cluster.ID)
+	require.NoError(rb.T(), err)
+	rb.standardUserCOProject = createProjectAsCO
+
+	//Additional testcase1 Validate if cluster owner can add a user as project owner in a project
+	err = users.AddProjectMember(rb.standardUserClient, rb.standardUserCOProject, rb.additionalUser, "project-owner")
```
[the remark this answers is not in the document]
anupama2501: I agree can be useful for any future automation tasks as well.

tests/v2/validation/rbac/rbac_test.go
anupama2501: additional tests are a few other scenarios which the customers/qa reported. Two reasons why have designed this way:

- The existing tests are the P0 test cases for a cluster rbac. 
- There are 2 more scenarios that pertain to project rbac where we can add them or any other new issues discovered in future in these additional tests. 

So, did not want to have both of these tests combined. Let me know if you don't agree with the way this was designed, we can discuss.

slickwarren (commented):
LGTM

igomez06 (approved):
Looks good! Can you link the backport to this PR?

slickwarren (approved):

tests/v2/validation/rbac/rbac_test.go
```diff
@@ -183,6 +187,65 @@ func (rb *RBTestSuite) ValidateRemoveClusterRoles(role string) {
...
+	assert.Equal(rb.T(), 1, len(clusterList.Data))
+
+	//Add the cluster member to the project created by the cluster Owner
+	err = users.AddProjectMember(rb.standardUserClient, rb.standardUserCOProject, rb.additionalUser, roleProjectOwner)
+	require.NoError(rb.T(), err)
+	userGetProject, err := projects.GetProjectList(rb.additionalUserClient, rb.cluster.ID)
+	assert.Equal(rb.T(), rb.standardUserCOProject.Name, userGetProject.Data[0].Name)
+
+}
+
```
caliskanugur, line 248: So Anupama, it seems like the `role` arguments aren't in use, for the new methods. But in the tests, they are already passed. You can either remove hardcoded const roles in the functions by using the `role` arguments or remove those arguments and use roles hardcoded way. @igomez06 what would be the best way for this case?

tests/v2/validation/rbac/rbac_test.go
```diff
@@ -263,6 +326,50 @@ func (rb *RBTestSuite) TestRBAC() {
...
+		require.NoError(rb.T(), err)
+	})
+
+	rb.Run("Additional testcase1 - Validating if member with role "+roleOwner+" can add another standard user as a project owner", func() {
+		rb.ValidateAddStdUserAsProjectOwner()
+
+	})
+
+	rb.Run("Additional testcase2 - Validating if member with role "+roleOwner+" can add another standard user as a cluster owner", func() {
+		rb.ValidateAddMemberAsClusterRoles(roleOwner)
```
caliskanugur, line 362: Depending above question, this might change

tests/v2/validation/rbac/rbac_test.go
```diff
@@ -263,6 +326,50 @@ func (rb *RBTestSuite) TestRBAC() {
...
+
+	})
+
+	rb.Run("Additional testcase3 - Validating if member with role "+roleOwner+" can add a cluster member as a project owner", func() {
+		rb.ValidateAddCMAsProjectOwner(roleOwner)
```
caliskanugur, line 367: Depending above question, this might change

caliskanugur (commented):
Looks good to me! Only one nitpick from the previous review, also your answer makes sense to me

tests/v2/validation/rbac/rbac_test.go
igomez06: I'm good with either one, if we anticipate the role being dynamic, then let's use the args rather than the consts. But if these test funcs will stay the same, please remove the unused arg. Thank you!

tests/v2/validation/rbac/rbac_test.go
igomez06: Yes depending on what is needed for the tests, updated the names as well, to stay consistent.

tests/v2/validation/rbac/rbac_test.go
anupama2501: Updated based on the feedback

tests/v2/validation/rbac/rbac_test.go
anupama2501: Updated based on the feedback

slickwarren (approved):

igomez06 (approved):
LGTM

caliskanugur (approved):
Looks good to me

PR #39675, opened 2022-11-22
enhance aksVmSizes path to include information on VM feature support

HarrisonWAffel:
## Problem
The API used to gather all Azure VM sizes does not include information on the capabilities of each VM in a given region. This prevents the UI from determining if a given VM supports features such as Availability Zones or Accelerated Networking.
 
## Solution
After discussion, we agreed that the existing API used to gather Azure VM sizes should be enhanced to also return the capabilities of each VM, so that particular UI fields can be enabled or disabled based off of the VM selection and given Azure region. Also, while the URL of the API includes the term `aks`, this API has been used for Azure node driver clusters for some time so I have not changed the name of the path.

 
## Testing
Perform CURL requests against the Azure VM size API, `https://<RANCHER_SERVER_URL>/meta/aksVMSizes?cloudCredentialId=<AZURE_CLOUD_CREDENTIAL>&region=<REGION>`. Observe that the response payload now includes additional information for each VM size.

The payload previously looked like this: 
```json
[
  ...
  "Standard_A1_v2",
  "Standard_A2m_v2",
  "Standard_A2_v2",
  "Standard_A4m_v2",
  ...
]
```

The payload will now have the following structure:
```json
[
  ...
    {
      "Name": "Standard_A2",
      "AcceleratedNetworkingSupported": false,
      "AvailabilityZones": []
    },
    {
      "Name": "Standard_A2m_v2",
      "AcceleratedNetworkingSupported": false,
      "AvailabilityZones": [
        "1",
        "2",
        "3"
      ]
    },
  ...
]
```

### Manual Testing
I've done the above curl requests and have confirmed that the output contains the same list of VM sizes returned from the previous version of the API, and did a smoke test across some of the VM sizes to check that the features in the output are accurate. 

### Automated Testing
None, as we need to use the Azure API / Rancher API to test a majority of the logic. The logic which can be easily tested is fairly trivial.

## QA Testing Considerations

+ UI changes are needed to handle this new payload format, ideally we will merge both changes in tandem so that nothing is broken when creating Azure clusters via the UI. 

+ There may be another change required for RKE1, as that portion of the UI does not currently use this API to get VM sizes.

+ There are ~600 VM sizes for Azure, we probably only want to test ones with different support options / availability zone options

We should check the AKS UI as well, to make sure that the new payload format is handled properly there.

### DIFF
--- a/pkg/api/norman/customization/aks/listers.go
+++ b/pkg/api/norman/customization/aks/listers.go
@@ -8,7 +8,7 @@ import (
 	"regexp"
 	"sort"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-30/compute"
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/skus"
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2020-09-01/containerservice"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-07-01/network"
 	"github.com/Azure/azure-sdk-for-go/services/subscription/mgmt/2020-09-01/subscription"
@@ -34,7 +34,7 @@ type subnet struct {
 
 var matchResourceGroup = regexp.MustCompile("/resource[gG]roups/(.+?)/")
 
-func NewClientAuthorizer(cap *Capabilities) (autorest.Authorizer, error) {
+func NewAzureClientAuthorizer(cap *Capabilities) (autorest.Authorizer, error) {
 	oauthConfig, err := adal.NewOAuthConfig(cap.AuthBaseURL, cap.TenantID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


jakefhyde (dismissed):
LGTM. Was going to suggest adding some tests but it seems like a lot of work.

pkg/api/norman/customization/aks/listers.go
```diff
@@ -373,27 +373,87 @@ func listClusters(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
...
-		return nil, http.StatusBadRequest, fmt.Errorf("failed to get VM sizes: %v", err)
+		return nil, http.StatusInternalServerError, fmt.Errorf("failed to get VM sizes: %v", err)
 	}
 
-	vmSizes := make([]string, 0, len(*vmMachineSizeList.Value))
+	if skuList.Value == nil {
+		return nil, http.StatusNotFound, err
+	}
+
+	var VMSkuInfo []VMSizeInfo
```
sbstp, line 426: nit: should this variable start with a lowercase?

pkg/api/norman/customization/aks/listers.go
```diff
@@ -373,27 +373,87 @@ func listClusters(ctx context.Context, cap *Capabilities) ([]byte, int, error) {
...
+		if resourceSku.Capabilities != nil {
+			for _, skuCapabilities := range *resourceSku.Capabilities {
+				if to.String(skuCapabilities.Name) == AzureAcceleratedNetworkingFeature && to.String(skuCapabilities.Value) == "True" {
+					vm.AcceleratedNetworkingSupported = true
+					break
+				}
+			}
+		}
+
+		if resourceSku.LocationInfo != nil && len(*resourceSku.LocationInfo) > 0 {
```
sbstp, line 447: Seems like these objects use a lot of pointers to maps/slices, i.e. double pointers. Is it just part of the azure sdk? Nothing that can be addressed?

HarrisonWAffel: Yeah. A lot of cloud provider sdks written in go use pointers heavily, so it's nothing we can change on our end.

sbstp (approved):

jiaqiluo (approved):

jakefhyde (approved):

Issue #39710, opened 2022-11-28
test panicked seen in CI for v2 integration test


superseb:
Seems to happen intermittently, but should never fail the step.

```
Running: build-token-generator
=== RUN   TestProjectUserTestSuite
=== RUN   TestProjectUserTestSuite/TestCreateNamespaceProjectMember
=== CONT  TestProjectUserTestSuite
    dynamic.go:25: Dynamic Client Host:localhost:8443
=== CONT  TestProjectUserTestSuite/TestCreateNamespaceProjectMember
    project_user_test.go:82: 
        	Error Trace:	project_user_test.go:82
        	Error:      	Received unexpected error:
        	            	Resource type [namespace] is not creatable
        	            	github.com/rancher/rancher/tests/framework/pkg/clientbase.(*APIOperations).DoCreate
        	            		/go/src/github.com/rancher/rancher/tests/framework/pkg/clientbase/ops.go:190
        	            	github.com/rancher/rancher/tests/framework/clients/rancher/v1.(*SteveClient).Create
        	            		/go/src/github.com/rancher/rancher/tests/framework/clients/rancher/v1/client.go:121
        	            	github.com/rancher/rancher/tests/framework/extensions/namespaces.CreateNamespace
        	            		/go/src/github.com/rancher/rancher/tests/framework/extensions/namespaces/create.go:48
        	            	github.com/rancher/rancher/tests/v2/integration/projects.(*ProjectUserTestSuite).TestCreateNamespaceProjectMember
        	            		/go/src/github.com/rancher/rancher/tests/v2/integration/projects/project_user_test.go:81
        	            	reflect.Value.call
        	            		/usr/lib64/go/1.19/src/reflect/value.go:584
        	            	reflect.Value.Call
        	            		/usr/lib64/go/1.19/src/reflect/value.go:368
        	            	github.com/stretchr/testify/suite.Run.func1
        	            		/go/pkg/mod/github.com/stretchr/testify@v1.7.1/suite/suite.go:158
        	            	testing.tRunner
        	            		/usr/lib64/go/1.19/src/testing/testing.go:1446
        	            	runtime.goexit
        	            		/usr/lib64/go/1.19/src/runtime/asm_amd64.s:1594
        	Test:       	TestProjectUserTestSuite/TestCreateNamespaceProjectMember
    suite.go:63: test panicked: runtime error: invalid memory address or nil pointer dereference
        goroutine 90 [running]:
        runtime/debug.Stack()
[...]

PR #39719, opened 2022-11-28
[v2.6] Fixed intermittent failure with TestCreateNamespaceProjectMember test case

igomez06:
## Issue: 
 
https://github.com/rancher/rancher/issues/39710
 
## Problem
 
There was a bit of a timing issue when adding a `project-member` role to a user in order to create a namespace, where the `project-member` role is not fully set causing a `Resource type [namespace] is not creatable` error. 
## Solution

 I added two extra checks to make sure the ProjectRoleTemplateBinding is successfully created.
## Testing

CI drone tests pass.

### DIFF
--- a/tests/framework/extensions/users/users.go
+++ b/tests/framework/extensions/users/users.go
@@ -1,6 +1,8 @@
 package users
 
 import (
+	"context"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


caliskanugur (dismissed):
Looks good to me

superseb (dismissed):

igomez06:
Pushed up one more check, this is to ensure the service account secret is properly created in order for a project member to create a namespace.

scripts/test
```diff
@@ -89,7 +89,7 @@ echo Running: build-token-generator
 path=$(pwd)
 ./tests/v2/integration/bin/integrationsetup
 export CATTLE_TEST_CONFIG=${path}/config.yaml
-go test -v ./tests/v2/integration/...
+go test -timeout 60m -v ./tests/v2/integration/...
```
superseb, line 92: If this runs longer than a few minutes, we know it's broken right? Should we adjust the timeout to a few minutes? Or was there another reason to make it 60?

superseb (dismissed):

scripts/test
igomez06: I forgot to remove this, I tried this out for a debugging. I'll remove it

superseb (approved):

caliskanugur (approved):

PR #39744, opened 2022-11-30
Add hardening config for rancher provisioned rke2 clusters

anupama2501:
## Problem
Currently, the Go test framework allows for provisioning RKE2 custom clusters. By default, these are non-hardened clusters. In realistic customer environments, hardened clusters are at the forefront, so we need to be able to support this in our testing. This PR addresses this problem by providing the option to harden the custom RKE2 clusters.

## Solution

- There needed to be an option available to select whether or not the user wishes to harden the cluster. This is done by adding a new Hardened bool type in the config.go struct. Inside the user's cattle-config.yaml, they will have to define a new hardened value and set it to true or false.
- In the custom_clusters_test.go file, new functions HardenNodes, HardenK3SRKE2ClusterConfig and UpdateK3SRKE2Cluster are referenced. We will break each down in the subsequent steps, consequently.
- HardenNodes is found in the new hardening/harden_nodes.go. The point of the file is it will check the new Hardened value as mentioned in step 1. If user says yes, harden the node following appropriate hardening steps.
- Once step 3 is finished, HardenK3SRKE2ClusterConfig will update the cluster's configuration to harden the cluster.
- Once step 4 is finished, UpdateK3SRKE2Cluster will call upon the Steve client's update function to complete the changes to actually harden the K3s custom cluster.

### DIFF
--- a/tests/framework/extensions/clusters/clusters.go
+++ b/tests/framework/extensions/clusters/clusters.go
@@ -3,6 +3,7 @@ package clusters
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/rancher/norman/types"
@@ -208,8 +209,7 @@ func NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretN
 		CloudCredentialSecretName: cloudCredentialSecretName,
 		KubernetesVersion:         kubernetesVersion,
 		LocalClusterAuthEndpoint:  localClusterAuthEndpoint,
-
-		RKEConfig: rkeConfig,
+		RKEConfig:                 rkeConfig,
 	}
 
 	v1Cluster := &apisV1.Cluster{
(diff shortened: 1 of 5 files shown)


tests/framework/extensions/clusters/clusters.go
```diff
@@ -225,28 +225,41 @@ func NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredentialSecretN
...
-		"enable-admission-plugins=NodeRestriction,PodSecurityPolicy,ServiceAccount",
-		"audit-policy-file=/var/lib/rancher/k3s/server/audit.yaml",
-		"audit-log-path=/var/lib/rancher/k3s/server/logs/audit.log",
-		"audit-log-maxage=30",
-		"audit-log-maxbackup=10",
-		"audit-log-maxsize=100",
-		"request-timeout=300s",
-		"service-account-lookup=true",
-	}
+	if strings.Contains(kubernetesVersion, "k3s") {
```
markusewalker, line 228: Double-checking...so the default will assume RKE2, correct? In other words, if the `kubernetesVersion` does not include `k3s`, then `rke2` will be assumed as the `kubernetesVersion`, correct?

tests/framework/extensions/hardening/rke2/harden_nodes.go
```diff
@@ -0,0 +1,88 @@
...
+			return err
+		}
+
+
+		if strings.Contains(nodeRoles[key],"--etcd") {
+			_, err = node.ExecuteCommand("sudo useradd -r -c \"etcd user\" -s /sbin/nologin -M etcd -U")
+			if err != nil {
+				return err
+			}
+
```
markusewalker, line 41: Just some nits here 😄 . Can we remove the extra spacing in the `HardeningNodes` function like in this line? Line 43 is another example, just so that we make the code more concise.

tests/framework/extensions/clusters/clusters.go
anupama2501: That is correct. I tried this locally for both k3s and RKE2 cluster and it worked for the both as expected. just to be extra cautious I'll get a k3s jenkins run too so that it works as expected.

igomez06 (approved):
LGTM, I agree with Markus' nits.

tests/v2/validation/provisioning/rke2/custom_cluster_test.go
```diff
@@ -163,6 +166,18 @@ func (c *CustomClusterProvisioningTestSuite) ProvisioningRKE2CustomCluster(exter
 					clusterToken, err := clusters.CheckServiceAccountTokenSecret(client, clusterName)
 					require.NoError(c.T(), err)
 					assert.NotEmpty(c.T(), clusterToken)
+
+					if tt.hardening.Hardened {
+						err = hardening.HardeningNodes(client, tt.hardening.Hardened, nodes, tt.nodeRoles)
+						require.NoError(c.T(), err)
+						hardenCluster := clusters.HardenK3SRKE2ClusterConfig(clusterName, namespace, "", "", kubeVersion, nil)
```
markusewalker, line 173: One last nit-pick 😄 . Cam we have this match the format you have in the dynamic test? You have your code spaced out in a more readable format, can we do the same in the static test?

Everything else lgtm!

markusewalker (approved):

markusewalker (approved):
LGTM

igomez06 (approved):
LGTM

slickwarren (approved):

PR #39757, opened 2022-12-01
Add optional filter on AzureAD auth group memberships

mouellet:
## Problem

Users in AzureAD can be member of hundreds of groups. Prior to MS Graph migration (< v2.6.8), filtering a user's group memberships (i.e. memberOf) was not possible.
 
## Solution

Add an optional filter using [OData filter](https://learn.microsoft.com/en-us/graph/filter-query-parameter) on user's group memberships w/ global setting on auth provider.

Filter example:
`startswith(displayName,'x') or startswith(displayName,'y')`


mouellet:
Still relevant

Stonedge:
Still relevant

PR #39759, opened 2022-12-01
Add `PodSecurityAdmissionConfigurationTemplate` CRD

HarrisonWAffel:
## Problem
Kubernetes v1.25 requires the use of Pod Security Standards and Pod Security Admissions over Pod Security Policies. Users need a way to manage these Pod Security Admissions both through the UI and through a CRD. 
 
## Solution
 
Implement a CRD which represents a Pod Security Admission Configuration, named `PodSecurityAdmissionConfigurationTemplate`. Two functions are also provided to return a preconfigured configuration for both restrictive and permissive configurations. 

If i've missed anything required to unblock the UI team in this PR let me know, these changes follow a similar pattern to other net-new CRD's created within the last year or so. 

## Testing
The new CRD should appear in the output of `kubectl get crd`, it will be named `podsecurityadmissionconfigurationtemplates.management.cattle.io`.

You should be able to apply a yaml the following yaml file and see it in the output of `kubectl get podsecurityadmissionconfigurationtemplates.management.cattle.io -A`

```yaml
apiVersion: management.cattle.io/v3
kind: PodSecurityAdmissionConfigurationTemplate
metadata:
  name: test-admission-config
configuration:
  defaults:
    enforce: "privileged"
    enforce-version: "latest"
    audit: "restricted"
    audit-version: "latest"
    warn: "privileged"
    warn-version: "latest"
  exemptions:
    usernames: ["abc123"]
    runtimeClasses: []
    namespaces: []
```

You should be able to see this new CRD when viewing `https://<RANCHER_SERVER_NAME>/v3`

### Manual Testing
I've done the above steps

### Automated Testing
N/A

N/A


jiaqiluo (approved):
LGTM 💯

pkg/apis/management.cattle.io/v3/pod_security_types.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const PodSecurityAdmissionConfigurationTemplateKind = "PodSecurityAdmissionConfigurationTemplate"
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type PodSecurityAdmissionConfigurationTemplate struct {
```
snasovich, line 12: Since `PodSecurityPolicyTemplate` is in `authz_types.go`, would this make sense to have these types there as well? 
This is really just a question, if there is a rationale for it being in own file, that's fine,

pkg/apis/management.cattle.io/v3/pod_security_types.go
```diff
@@ -0,0 +1,92 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Configuration PodSecurityAdmissionConfigurationTemplateSpec `json:"configuration"`
```
snasovich, line 16: I know it's not part of the requirements from the linked issue, but would adding `description` (we have it on `PodSecurityPolicyTemplate`) be beneficial here to explain the purpose of each PSAConfigTemplate?

pkg/apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -108,21 +108,22 @@ type Cluster struct {
...
-	LocalClusterAuthEndpoint             LocalClusterAuthEndpoint                `json:"localClusterAuthEndpoint,omitempty"`
-	ScheduledClusterScan                 *ScheduledClusterScan                   `json:"scheduledClusterScan,omitempty"`
-	ClusterSecrets                       ClusterSecrets                          `json:"clusterSecrets" norman:"nocreate,noupdate"`
+	DesiredAgentImage                                  string                                  `json:"desiredAgentImage"`
+	DesiredAuthImage                                   string                                  `json:"desiredAuthImage"`
+	AgentImageOverride                                 string                                  `json:"agentImageOverride"`
+	AgentEnvVars                                       []v1.EnvVar                             `json:"agentEnvVars,omitempty"`
+	RancherKubernetesEngineConfig                      *rketypes.RancherKubernetesEngineConfig `json:"rancherKubernetesEngineConfig,omitempty"`
+	DefaultPodSecurityAdmissionConfigurationTemplateId string                                  `json:"defaultPodSecurityAdmissionConfigurationTemplateId,omitempty"`
```
snasovich, line 116: I know it's in the requirements of the linked issue, but would it make more sense to name it `defaultPodSecurityAdmissionConfigurationTemplateName` to match the `defaultPodSecurityPolicyTemplateName` (IMHO it better relays the fact we reference Name and not some sort of ID).

Though I see `defaultPodSecurityPolicyTemplateId` in a few (autogenerated) files in Rancher so not sure why is that. But I clearly see `defaultPodSecurityPolicyTemplateName` on `cluster.provisioning.cattle.io` objects.

pkg/apis/management.cattle.io/v3/pod_security_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+// PodSecurityAdmissionConfigurationTemplateDefaults are applied when a mode label is not set.
+//
+// The level label values must be one of
+// 'privileged' (default), 'baseline', or 'restricted'
+//
+// The version levels must be either 'latest' (default), or a specific version (e.g. 'v1.25')
+type PodSecurityAdmissionConfigurationTemplateDefaults struct {
+	Enforce        string `json:"enforce"`
+	EnforceVersion string `json:"enforce-version" yaml:"enforce-version"`
```
snasovich, line 32: It's somewhat baffling that there are different naming conventions here, like `enforce-version` vs. `runtimeClasses` but that's how it's in upstream. So that's OK we follow it.

pkg/apis/management.cattle.io/v3/pod_security_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+type PodSecurityAdmissionConfigurationTemplateExemptions struct {
+	Usernames      []string `json:"usernames"`
+	RuntimeClasses []string `json:"runtimeClasses"`
+	Namespaces     []string `json:"namespaces"`
+}
+
+func NewPodSecurityAdmissionConfigurationTemplateRestricted() PodSecurityAdmissionConfigurationTemplate {
+	return PodSecurityAdmissionConfigurationTemplate{
+		TypeMeta: metav1.TypeMeta{
```
snasovich, line 47: Do we have to pass `TypeMeta` here? (Trying to avoid hardcoding `APIVersion`)
For comparison, it's skipped for PSPTs: https://github.com/rancher/rancher/blob/60ab459ac61cae96dfc3145e70e5d3afb9fdf7c8/pkg/data/management/podsecuritypolicytemplate_data.go#L18

snasovich (approved):
Generally LGTM, a couple of improvement suggestions + questions.

pkg/apis/management.cattle.io/v3/cluster_types.go
HarrisonWAffel: Yeah, I can make this change - a lot of the naming conventions were taken directly from the spec but having consistency between the fields would look better. The auto-generated files contain `id` simply because this field contains `id`

pkg/apis/management.cattle.io/v3/pod_security_types.go
HarrisonWAffel: Jack and I agreed to put it into a new file, it might be helpful if more helper functions are needed for PSS / PSA.

pkg/apis/management.cattle.io/v3/pod_security_types.go
HarrisonWAffel: I'll follow the convention used by PSPTS, wasn't aware we wanted to avoid hard-coding the `APIVersion`

pkg/apis/management.cattle.io/v3/pod_security_types.go
HarrisonWAffel: After discussing we agreed to add a new field `description` in a similar fashion to how PSP's are implemented (I initially thought this request was to add a comment description to the field).

pkg/apis/management.cattle.io/v3/pod_security_types.go
HarrisonWAffel: I've removed the `TypeMeta` and added the two default templates to `add.go`, they now appear when using `kubectl`

pkg/data/management/podadmissionconfigurationtemplate_data.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/types/config"
+)
+
+func addDefaultPodSecurityAdmissionConfigurationTemplates(management *config.ManagementContext) error {
+	psapts := management.Management.PodSecurityAdmissionConfigurationTemplates("")
+	if _, err := psapts.Create(v3.NewPodSecurityAdmissionConfigurationTemplateRestricted()); err != nil {
+		return err
```
snasovich, line 11: I believe we need to ignore "already exists" errors here, see https://github.com/rancher/rancher/blob/68215bcdde090854ff28b43b01d1aa0b69611920/pkg/data/management/podsecuritypolicytemplate_data.go#L149

Otherwise, we will have these errors bubble up on every Rancher leader change.

Also it's helpful to wrap this error before returning, similar how it's at the linked code (with `%w` instead of `%v` though).

snasovich (changes requested):
One small change requested, otherwise LGTM.

snasovich (approved):

jiaqiluo (dismissed):

snasovich (approved):

Issue #39139, opened 2022-09-27
[BUG] [RKE2] `node auto-replace` does not work on RKE2/k3s rancher provisioned clusters


sowmyav27:
**Rancher Server Setup**
- Rancher version: 2.6-head `2e00266`
- Installation option (Docker install/Helm Chart): Helm
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc): RKE1

**Information about the Cluster**
- Kubernetes version: `v1.23.10+rke2r1`/ `v1.23.10+k3s1`
- Cluster Type (Local/Downstream): RKE2 and k3s

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom) admin

**Describe the bug**
[BUG] [RKE2] `node auto-replace` does not work on RKE2/k3s rancher provisioned clusters

**To Reproduce**
1. Deploy an RKE2 or K3s cluster form Rancher - node driver - 1 etcd/cp and 3 worker nodes
2. In the worker node pool, set Auto replace to 30 or 60 seconds.
3. Once the cluster is up and Active, `Power off` one of the worker nodes in the backend (in the Linode console in my case)
4. The node in Rancher UI is see to have message `Node condition MemoryPressure is Unknown. Node condition DiskPressure is Unknown. Node condition PIDPressure is Unknown. Node condition Ready is Unknown.`
5. The cluster is stuck in Updating state
6. No new node gets provisioned

**Expected Result**
A new node must provision and existing one deleted after the set number of seconds in `Auto Replace` field.

**Note**
Set `nodeStartupTimeout: 0s ` on another cluster and also set Auto replace to 60 seconds. The node did _not_ get deleted/reprovisioned. 

SURE-5262

rancherbot:
@snasovich, Not creating `backport` issue for issue `39139` in repository `rancher` because milestone `2.6.10` does not exist or is not an open milestone

slickwarren:
waiting on design/more info to assign QA

thsnielsen:
@snasovich i have a customer waiting for this feature to put rancher in production. Are we sure+sure+sure it will be in 2.7.1 and i also see a mention of 2.6.10 - is the backport to 2.6.10 comitted?

snasovich:
[...]

PR #39766, opened 2022-12-02
Add functions to find CAPI cluster from objects

jakefhyde:
## Issue: 

Part of groundwork for #39139

## Problem

We need a mechanism to get the owning CAPI cluster or machine objects in controllers to determine if reconcilation is paused.

## Solution

Add functions to get these objects.
 
### Automated Testing

Added unit tests to ensure this works.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/common.go
+++ b/pkg/controllers/provisioningv2/rke2/common.go
@@ -18,13 +18,16 @@ import (
 	rkecontroller "github.com/rancher/rancher/pkg/generated/controllers/rke.cattle.io/v1"
 	"github.com/rancher/rancher/pkg/serviceaccounttoken"
 	"github.com/rancher/wrangler/pkg/condition"
+	"github.com/rancher/wrangler/pkg/data"
 	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/name"
 	corev1 "k8s.io/api/core/v1"
 	apierror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 )
@@ -371,8 +374,8 @@ func CopyPlanMetadataToSecret(secret *corev1.Secret, metadata *plan.Metadata) {
 		secret.Annotations = map[string]string{}
 	}
 
-	CopyMapWithExcludes(secret.Labels, metadata.Labels, nil)
-	CopyMapWithExcludes(secret.Annotations, metadata.Annotations, nil)
+	CopyMap(secret.Labels, metadata.Labels)
+	CopyMap(secret.Annotations, metadata.Annotations)
 }
 
 // CopyMap will copy the items from source to destination. It will only copy items that have keys that start with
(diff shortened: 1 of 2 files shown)


jakefhyde:
@Oats87 Added more tests and comments

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -403,3 +404,76 @@ func SortedKeys(m map[string]interface{}) []string {
...
+
+type Getter[T any] interface {
+	Get(namespace, name string) (*T, error)
+}
+
+var errNilObject = errors.New("cannot get capi cluster for nil object")
+
+// FindCAPIClusterFromLabel takes an obj T and will attempt to find the label denoting which capi cluster it belongs to.
+// If the object is nil, it cannot access to object or type metas, or the label is not present, it returns an error.
+func FindCAPIClusterFromLabel[T any](obj *T, getter Getter[capi.Cluster]) (*capi.Cluster, error) {
```
Oats87, line 416: @jakefhyde is the intention here to be able to retrieve the labels on "any" K8s-like object to look for the CAPI cluster label value?

We have the boilerplate for functionality similar to this (in a generic fashion) with the wrangler data package, see: https://github.com/rancher/wrangler/blob/master/pkg/data/data.go#L29 that operates on runtime.Object -- where are we sourcing the objects that are passed into these functions? I'm not sure that it makes sense to use generics here.

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/common.go
jakefhyde: Is this better?
```go
func FindCAPIClusterFromLabel(obj runtime.Object, getter Getter[capi.Cluster]) (*capi.Cluster, error) {
	data, err := data.Convert(obj)
	if err != nil {
		return nil, err
	}
	if data == nil {
		return nil, errNilObject
	}
	clusterName := data.String("metadata", "labels", capi.ClusterLabelName)
	if clusterName != "" {
		return getter.Get(data.String("metadata", "namespace"), clusterName)
	}
	return nil, fmt.Errorf("%s label not present on %s: %s/%s", capi.ClusterLabelName, obj.GetObjectKind().GroupVersionKind().Kind, data.String("metadata", "namespace"), data.String("metadata", "name"))
}
```
The only reason I was using generics here is because `if obj == nil` is always false (the interface is never nil, but the underlying pointer value is). 

Example usages can be found here: https://github.com/rancher/rancher/blob/8c101f5f63dc103e7ecc5476a08491994a9fc32d/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go#L437
and here:
https://github.com/rancher/rancher/blob/8c101f5f63dc103e7ecc5476a08491994a9fc32d/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go#L448

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -403,3 +406,84 @@ func SortedKeys(m map[string]interface{}) []string {
...
+	Get(namespace, name string) (*capi.Machine, error)
+}
+
+var errNilObject = errors.New("cannot get capi cluster for nil object")
+
+// GetCAPIClusterFromLabel takes a runtime.Object and will attempt to find the label denoting which capi cluster it
+// belongs to.
+// If the object is nil, it cannot access to object or type metas, or the label is not present, it returns an error.
+// If the object has the expected label, it will return the capi cluster object.
+func GetCAPIClusterFromLabel(obj runtime.Object, getter CAPIClusterGetter) (*capi.Cluster, error) {
```
Oats87, line 424: What is the reason for passing in a getter?

Oats87: Like, is there a legitimate reason where the getter will be different? Or is this future-proofing?

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -403,3 +406,84 @@ func SortedKeys(m map[string]interface{}) []string {
 	sort.Strings(keys)
 	return keys
 }
+
+type CAPIClusterGetter interface {
```
Oats87, line 410: I don't think we need this anymore

pkg/controllers/provisioningv2/rke2/common.go
```diff
@@ -403,3 +406,84 @@ func SortedKeys(m map[string]interface{}) []string {
...
+		return nil, err
+	}
+	return cache.Get(namespace, ref.Name)
+}
+
+// getOwnerFromGVK takes a runtime.Object, and will search for a controlling owner reference of kind apiVersion.
+// If the object is nil, it cannot access to object or type metas, the owner reference Kind or APIVersion do not match,
+// or the object could not be found, it returns an error.
+// If the owner reference exists and is valid, it will return the owner reference and the namespace it belongs to.
+func getOwnerFromGVK(groupVersion, kind string, obj runtime.Object) (*metav1.OwnerReference, string, error) {
```
Oats87, line 467: Is there a reason we wouldn't want to publicize this function?

Oats87 (changes requested):

Oats87 (approved):

annablender (approved):
Looks good to me

PR #39768, opened 2022-12-02
[2.6] Add kubernetes version upgrade validation tests

caliskanugur:
**Extensions Updates**
- Refactored:
   - `environmentflags` configuration key to be used as `desiredflags` instead of `use`
- Implemented:
  - A new package called `bundledclusters`:
    - Enables to interact of v1 and v3 clusters as one struct
    - Get, List, and Update `bundledcluster` methods for both v1 and v3
    - A method to update the Kubernetes version of a cluster for all different provider types
    - A method to list available versions for all different provider types
  - Cluster meta struct to contain cluster’s meta for v1 and v3 clusters such as provider type, imported cluster or hosted cluster
  - A listener to watch a cluster in the upgrade state
  - Versions package:
    - List all functions for different cluster types versions
    - List available functions built on a list of all functions with a specified cluster

**Validation Tests**
- Implemented:
  - Private configuration input version, node pool versions for hosted providers, and Kubernetes version validations
  - Private configuration loader with two different options:
    - with a flag to run an upgrade Kubernetes test for all clusters except the local cluster
    - with a slice of structs with given versions for specific clusters

### DIFF
--- a/tests/framework/clients/rancher/v1/client.go
+++ b/tests/framework/clients/rancher/v1/client.go
@@ -82,7 +82,7 @@ func NewClient(opts *clientbase.ClientOpts) (*Client, error) {
 // SteveType is a function that sets the resource type for the SteveClient
 // e.g. accessing the Steve namespace resource
 //
-//	 nameSpaceClient := client.V1.SteveType("namespace")
+//	nameSpaceClient := client.V1.SteveType("namespace")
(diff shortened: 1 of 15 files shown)


igomez06 (approved):
LGTM

markusewalker (approved):

PR #39790, opened 2022-12-06
[2.7] Add additional test cases for AKS

markusewalker:
## Problem

## Solution

1. Scaling AKS node pools
2. Upgrading K8s versions of the AKS cluster and nodes
3. Importing an AKS cluster that was provisioned from Azure into Rancher

After this PR is merged, provisioning private clusters will be a next step, but that will come after this PR has gone through any design changes and is merged.
 
### Manual Testing

### Automated Testing


markusewalker:
Closing this PR and opening a fresh one due to some commit issues being easier to handle by just creating a new PR.

PR #39791, opened 2022-12-06
[v2.6] Include additional AKS tests

markusewalker:
## Problem

## Solution

1. Scaling AKS node pools
2. Refacors the `hosted` folder to have each additional cluster split into it's own folder
     - Long term, this will help in our CI/CD pipelines.

After this PR is merged, provisioning private clusters and importing clusters will be a next step, but that will come after this PR has gone through any design changes and is merged.
 
### Manual Testing

### Automated Testing


markusewalker:
Met offline with @igomez06 @Josh-Diamond @caliskanugur regarding design changes and ultimately what we want future tests to look like. As there will be several changes needed, going to move this into a draft state and bring it back out after.

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/rancher/rancher/tests/framework/extensions/clusters/aks/resources"
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+// ImportAKSCluster is a helper function that imports an AKS cluster into Rancher. It will first create a resoure group
+// and then create a cluster in that resource group. It will then import the cluster into Rancher.
+func ImportAKSCluster(client *rancher.Client, cloudCredential *cloudcredentials.CloudCredential, clusterName string) (*management.Cluster, error) {
+	logrus.Infof("======================================")
+	logrus.Infof("TEST CASE: Importing AKS Cluster")
```
igomez06, line 21: Test cases should never be in extensions. Extensions to facilitate resource set up.

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+)
+
+// CreateAKSCluster creates an AKS cluster in the specified resource group.
+func CreateAKSCluster(aksHostCluster *management.AKSClusterConfigSpec, aksImportCluster *management.AKSImportedClusterConfigSpec, clusterName, subscriptionID, clientID string, credential azcore.TokenCredential) (*runtime.Poller[armcontainerservice.ManagedClustersClientCreateOrUpdateResponse], error) {
```
igomez06, line 14: So Markus, this might have bee lost in discussion, but I thought we agreed that the actual AKS cluster creation would happen instead in corral. That way we aren't supporting multiple node provider SDKs in the framework. Unless others disagree,  I think the actual cluster creation should happen in corral instead. Anyone else please weigh in.

igomez06:
Markus a lot of the design decisions we made were not implemented. ie. separate test cases for scaling, and importing.

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
```diff
@@ -0,0 +1,100 @@
...
+	logrus.Infof("======================================")
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	aksHostCluster := aks.AKSHostClusterConfig(clusterName, cloudCredential.ID)
+	aksImportCluster := aks.AKSImportedClusterConfig(clusterName, cloudCredential.ID)
+	ctx := context.Background()
```
caliskanugur, line 31: From the previous review, I think `context.TODO` is a better fit here, or the context cancellations won't be used by the Azure client cause of `context.Background`, but I guess most of this extension will be changed if corral is used, so please ignore this comment if corral is agreed.

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+
+	params := armcontainerservice.ManagedCluster{
+		Location:   aksImportCluster.Location,
+		Identity:   clusterIdentity,
+		Tags:       map[string]*string{},
+		Properties: properties,
+		SKU:        skuProfile,
+	}
+
+	ctx := context.Background()
```
caliskanugur, line 91: I've got the same context comment here

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// DeleteAKSCluster deletes an AKS cluster in the specified resource group.
+func DeleteAKSCluster(clusterName, subscriptionID string, aksHostCluster *management.AKSClusterConfigSpec, credential azcore.TokenCredential) (*runtime.Poller[armcontainerservice.ManagedClustersClientDeleteResponse], error) {
+	client, err := armcontainerservice.NewManagedClustersClient(subscriptionID, credential, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx := context.Background()
```
caliskanugur, line 103: I've got the same context comment here

tests/framework/clients/rancher/generated/management/v3/zz_generated_aks_imported_cluster_config_spec.go
```diff
@@ -0,0 +1,50 @@
...
+	AKSImortedClusterConfigSpecFieldEnableAutoScaling         = "enableAutoScaling"
+	AKSImortedClusterConfigSpecFieldEnablePublicIP            = "enablePublicIP"
+	AKSImortedClusterConfigSpecFieldEnableRBAC                = "enableRBAC"
+	AKSImortedClusterConfigSpecFieldKubernetesVersion         = "kubernetesVersion"
+	AKSImortedClusterConfigSpecFieldLoadProfileCount          = "loadProfileCount"
+	AKSImortedClusterConfigSpecFieldLocation                  = "location"
+	AKSImortedClusterConfigSpecFieldMaxCount                  = "maxCount"
+	AKSImortedClusterConfigSpecFieldMaxNodeProvisionTime      = "maxNodeProvisionTime"
+	AKSImortedClusterConfigSpecFieldMaxPods                   = "maxPods"
+	AKSImortedClusterConfigSpecFieldMinCount                  = "minCount"
```
caliskanugur, line 18: It seems like these are hardcoded cause of the typos in the const names. Is it the case? 

If this is hardcoded, these things will be gone whenever the v3 generation happens.

tests/framework/clients/rancher/generated/management/v3/zz_generated_aks_imported_cluster_config_spec.go
```diff
@@ -0,0 +1,50 @@
+package client
+
+const (
```
igomez06, line 3: This file does not appear to be generated, did you implement this Markus? All files in clients/rancher/generated/ are generated using the framework/pkg/schemas. Because what will happen is there will be a v3 update, which is going to happen because of the updates in eks, this whole file will be deleted when we generate new structs.

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
markusewalker: Correct, corral was discussed as the solution. I had the solution presented here in short-term as we are prepping for k8s 1.25, that will take highest priority on our end for testing.

It will take a bit of ramp-up time on our end with Corral as well, so I had this in the short-term. I can hold off if we believe that is the best case to have that implemented now rather than later.

markusewalker:
> Markus a lot of the design decisions we made were not implemented. ie. separate test cases for scaling, and importing.

@igomez06 that's my bad, I honestly had forgotten that point when making the changes yesterday afternoon. I'll fix that and push that in a separate commit.

tests/framework/clients/rancher/generated/management/v3/zz_generated_aks_imported_cluster_config_spec.go
markusewalker: Correct, I had made this file to get aside from hardcoding values as we had seen earlier in the import test. That way, a user can just add to the config.yaml file to make it more dynamic, the import test that is.

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
igomez06: Ah ok, well I don't think a short term solution is the answer. We need to decide on how we want to support this. I think it would be best to have it done in corral in order to avoid supporting multiple node provider SDKs, that was brought up as a concern from the dev team, in the initial designing of the framework. But also I can be persuaded otherwise, if multiple people think it would be to do it here.

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
```diff
@@ -149,6 +152,14 @@ func (h *HostedClusterProvisioningTestSuite) TestProvisioningHostedAKSCluster()
...
 			require.NoError(h.T(), err)
 			assert.NotEmpty(h.T(), clusterToken)
+
+			updatedCluster, err := nodepools.ScalingAKSNodePools(client, clusterResp, clusterName, cloudCredential.ID)
+			require.NoError(h.T(), err)
+			assert.Equal(h.T(), clusterName, updatedCluster.Name)
+
+			importedCluster, err := importing.ImportAKSCluster(client, cloudCredential, importedClusterName)
+			require.NoError(h.T(), err)
+			assert.Equal(h.T(), importedClusterName, importedCluster.Name)
```
caliskanugur, line 162: I still think the same with the previous review and design talk. I think this would be better as another test like we have for custom and node provider provisioning RKE1/RKE2/K3s tests. In my opinion, running this test should only cover provisioning AKS. Currently, it covers imported + provisioning + scaling. I don't have a strong opinion on the scaling, but imported and provisioning tests would be better if separated.

tests/framework/clients/rancher/generated/management/v3/zz_generated_aks_imported_cluster_config_spec.go
igomez06: It doesn't belong here, all files in this directory are auto generated, as soon as we run go generate for updates this will be deleted. I guess I need more context to why this file is here. All these structs are for v3 API calls, is this an API call?

markusewalker: Got it, makes sense. No API call, I had placed here as initially I believed I had to do that to allow for the import test to be made dynamic and not hard-coded. But it wounds that is done in `framework/pkg/schemas`. I will take a deeper look into that.

caliskanugur (commented):
There are a few more things that have been discussed in the design meeting to address, generally, it looks good.

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
markusewalker: Thought I had changed that from the review, but guess I missed that too. Thanks for pointing it out, I'll fix that.

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
markusewalker: Yeah, I agree. As mentioned in an earlier comment, I had forgotten to refactor that part when I pushed the changes yesterday. I'll fix that in a separate commit as well and put EKS and GKE in their own respective folders as well while I'm at it.

markusewalker:
Discussed with Israel and Brandon offline. I'm going to move back to draft again, but push changes discussed here. Due to competing priorities with K8s 1.25, this is going to be put on the back-burner.

Once we're ready to come back to this, I'll open this PR back up at that time.

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
markusewalker: Temporarily removing the imported test at this time due to offline conversations with Israel and Brandon. I will come back to that test in a separate PR in the near-future. For now, we will focus on specifically refactoring the hosted cluster test structure and adding scaling node pool support.

igomez06 (approved):
LGTM

caliskanugur (approved):
LGTM

slickwarren (approved):

PR #39793, opened 2022-12-06
[v2.7] Add additional AKS tests

markusewalker:
## Problem

## Solution

1. Scaling AKS node pools
2. Refacors the `hosted` folder to have each additional cluster split into it's own folder
     - Long term, this will help in our CI/CD pipelines.

After this PR is merged, provisioning private clusters and importing clusters will be a next step, but that will come after this PR has gone through any design changes and is merged.
 
### Manual Testing

### Automated Testing


tests/framework/extensions/clusters/aks/resources/create_nodepool.go
```diff
@@ -0,0 +1,34 @@
+package resources
+
+import (
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+)
+
+// CreateNodePool is a helper function that creates a node pool.
+func CreateNodePool() (*management.AKSNodePool, error) {
+	autoScaling := true
+	diskSize := int64(120)
```
Josh-Diamond, line 10: An alternative to hardcoding a default value here, would be to pull these values from the AKSNodePool struct in aks_cluster_config.go.  diskSize, maxCount, maxPods, minCount, nodeCount are already defined there as int64, so it would also avoid this initial type conversion when declaring these.

Josh-Diamond:
This is GREAT Markus!  The logging here is phenomenal 🔥

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
```diff
@@ -0,0 +1,126 @@
...
+var (
+	balanceSimilarNodeGroups = "false"
+	ctx                      = context.Background()
+	defaultRandStringLength  = 5
+	disableLocalAccounts     = false
+	dnsPrefix                = "auto-dns"
+	enableAutoScaling        = true
+	enablePublicIP           = false
+	enableRBAC               = true
+	k8sVersion               = "1.24.6"
```
Josh-Diamond, line 21: Is this a best practice to hard code all these values as defaults?  Perhaps some could be pulled from the config.yml to keep things more dynamic?

tests/framework/extensions/clusters/aks/upgrades/k8s_upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+// UpgradeK8SVersion is a helper function that upgrades the AKS cluster's k8s version
+func UpgradeK8SVersion(client *rancher.Client, cluster *management.Cluster, displayName, cloudCredentialID string) (*management.Cluster, error) {
+	logrus.Infof("======================================")
+	logrus.Infof("TEST CASE: Upgrading K8s Version")
+	logrus.Infof("======================================")
+
+	version_1_24_6 := "1.24.6"
```
Josh-Diamond, line 19: similar thoughts, i think pulling these values from config.yml would keep things more dynamic and flexible.

tests/framework/extensions/clusters/aks/upgrades/k8s_upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+
+			clusterResp, err := client.Management.Cluster.ByID(newCluster.ID)
+			if err != nil {
+				return false, err
+			}
+
+			if *clusterResp.AKSStatus.UpstreamSpec.KubernetesVersion != *updatedPoolsCluster.AKSConfig.KubernetesVersion && clusterResp.State != "active" {
+				return false, err
+			} else if *clusterResp.AKSStatus.UpstreamSpec.KubernetesVersion == *updatedPoolsCluster.AKSConfig.KubernetesVersion && clusterResp.State == "active" {
```
Josh-Diamond, line 60: I wouldn't say this is an issue per-say, but I've always found it best to stick exclusively to "if" statements.  Curious to hear the teams' thoughts around this

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
```diff
@@ -149,6 +153,18 @@ func (h *HostedClusterProvisioningTestSuite) TestProvisioningHostedAKSCluster()
 			clusterToken, err := clusters.CheckServiceAccountTokenSecret(client, clusterName)
 			require.NoError(h.T(), err)
 			assert.NotEmpty(h.T(), clusterToken)
+
+			updatedCluster, err := nodepools.ScalingAKSNodePools(client, clusterResp, clusterName, cloudCredential.ID)
```
Josh-Diamond, line 157: this is so clean! LOVE IT 🔥

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
markusewalker: Possibly. I made the `resources` folder with two particular goals in mind. The first was to eliminate overcrowding the config file with excessive parameters like it does now. The second is that these values are making an AKS cluster from Azure directly, so aside from the K8s version, the defined values should not be left up to the user to tinker with, otherwise the defined tests won't work.

tests/framework/extensions/clusters/aks/upgrades/k8s_upgrade.go
markusewalker: The upgrading tests looks at the highest upgraded version of K8s available in Rancher and tests against that. So we could do that, but it would go against the design of the test.

tests/framework/extensions/clusters/aks/resources/create_nodepool.go
markusewalker: Similar to my comment [below](https://github.com/rancher/rancher/pull/39793#discussion_r1042421079).

igomez06: I agree with Josh, let's not hard code values

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
```diff
@@ -149,6 +153,18 @@ func (h *HostedClusterProvisioningTestSuite) TestProvisioningHostedAKSCluster()
...
+			require.NoError(h.T(), err)
+			assert.Equal(h.T(), clusterName, updatedCluster.Name)
+
+			upgradeK8S, err := upgrades.UpgradeK8SVersion(client, clusterResp, updatedCluster.Name, cloudCredential.ID)
```
igomez06, line 161: I personally don't think upgrading, and importing an aks cluster be part of the main aks test. Shouldn't these all be different tests? What if we are adding more validations, to a provisioned aks cluster. Are we always going to upgrade, then import a cluster? In my opinion these should be different test cases in the test suite. Or a separate test suite. I'm open to hear other's opinions on this.

igomez06 (commented):
Please use Ugur's code for the upgrade portions of your PR, a lot of the issue we have with the hard coded values will go away.

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
markusewalker: Yeah I should have Ugur's changes, I'll take a look. So, do we want to go with putting all values into the config file then? My main concern is excluding the K8s version, the values aren't supposed to change in order to populate an AKS cluster from Azure.

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
markusewalker: I'm open to changing it to be different test cases. I initially had it just like this for further design review, but am open to changing to how we see best for sure.

go.mod
```diff
@@ -160,10 +160,17 @@ require (
 
 require github.com/google/gnostic v0.5.7-v3refs
 
-require github.com/kr/fs v0.1.0 // indirect
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.0.0 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/internal v1.0.0 // indirect
+	github.com/kr/fs v0.1.0 // indirect
+	github.com/pkg/browser v0.0.0-20210115035449-ce105d075bb4 // indirect
```
igomez06, line 167: We adding a few things to the go.mod file. Could explain why we need all of these? I don't want to bloat this file too much as it's rancher/rancher's main go.mod file

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
igomez06: Can you ask Ugur how he did it? I don't want hardcoded values especially here in the extensions, which are to be used by all tests, and these values are specific to a test case. So they definitely don't belong here. Can you see how we do upgrade tests for charts? We are essentially doing the same. What he do there is we make sure we create a chart with a version previous to the latest, that we ensure we can upgrade. That is much more efficient than hard coding it. The rest can be a config entry in a yaml in the actual test directory. This is to avoid having to do a whole PR just to update certain values, which I want to avoid.

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
igomez06: On another note we might need to do separate test suites for all providers. When I wrote this, the framework was still very young, so I'm open to changes. Just some suggestions

tests/framework/extensions/clusters/aks/upgrades/k8s_upgrade.go
markusewalker: On second thought, I think I'll just scrap this upgrade test altogether. Since Ugur already has this changes in, I don't think it makes sense for it to have in multiple locations. Also avoids less work in terms of maintaining as well as you pointed out in an earlier comment @igomez06 .

tests/framework/extensions/clusters/aks/aks_cluster_config.go
```diff
@@ -41,11 +41,13 @@ type AKSClusterConfig struct {
 // AKSNodePool is the configuration needed to an AKS node pool
 type AKSNodePool struct {
 	AvailabilityZones   *[]string `json:"availabilityZones,omitempty" yaml:"availabilityZones,omitempty"`
-	NodeCount           *int64    `json:"nodeCount,omitempty" yaml:"nodeCount,omitempty"`
 	EnableAutoScaling   *bool     `json:"enableAutoScaling,omitempty" yaml:"enableAutoScaling,omitempty"`
 	MaxPods             *int64    `json:"maxPods,omitempty" yaml:"maxPods,omitempty"`
+	MaxCount            *int64    `json:"maxCount,omitempty" yaml:"maxCount,omitempty"`
+	MinCount            *int64    `json:"minCount,omitempty" yaml:"minCount,omitempty"`
```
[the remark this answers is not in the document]
igomez06: Oh nice just curious, I think I forgot to add these, or these fields were created later. Just wanted to make sure these values were in the generated struct was as that is what we are sending to rancher.

tests/framework/extensions/clusters/aks/upgrades/k8s_upgrade.go
markusewalker: Going to remove this test altogether considering Ugur already has upgrades for hosted clusters.

tests/framework/extensions/clusters/aks/resources/create_nodepool.go
markusewalker: Thank you both, I'll rework this so that we simply use the values defined in the config file as opposed to hardcoding here.

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials"
+	"github.com/rancher/rancher/tests/framework/extensions/clusters/aks"
+	"github.com/rancher/rancher/tests/framework/extensions/clusters/aks/resources"
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+var (
+	ctx                     = context.Background()
```
caliskanugur, line 18: Out of curiosity, is this the preferred context specifically for AKS? This won't cancel any context whether the aks client cancels that or not (assuming this context passed to the aks clients). I think context.TODO would be the best fit here if it's not specific. Also for future implementations, I think these should be instantiated in the function scopes, otherwise, the extension function call would overwrite it as they are global variables currently.

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+)
+
+var (
+	balanceSimilarNodeGroups = "false"
+	ctx                      = context.Background()
```
caliskanugur, line 14: I've got the same context question here

tests/framework/extensions/clusters/aks/resources/aks_cluster.go
caliskanugur: I agree with that, that'd be great if we could reduce hardcoded values. Charts and Kubernetes upgrades can be taken as examples, I am open to providing more information on them

tests/framework/extensions/clusters/aks/upgrades/k8s_upgrade.go
caliskanugur: Yes, Markus, there should be now a versions package where you can get different version pools as a slice of strings. In the upgrade test, it takes that string of slices, if the config is not given it takes the latest available for the cluster, or the user still can declare the desired version to upgrade. I am open to talking about this as well, let me know if something seems odd

tests/v2/validation/provisioning/hosted/hosted_provisioning_test.go
caliskanugur: I agree with expanding the tests horizontally. In my opinion, we need the same thing that happened for custom clusters and provisioning clusters. Adding these extra steps will make this hard to combine in the pipelines in long term. F.e. for the HA release if we had two separate tests as creating imported aks and provision aks, then we would be able to glue everything up in CI. The same idea goes for the scaling test as well, but I don't have a strong opinion on that. But creating everything in one test will make this a test where it should've been designed in the CI part.

caliskanugur (commented):
A few nitpicks, and questions. Looks good!

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
```diff
@@ -0,0 +1,105 @@
...
+	aksHostCluster := aks.AKSHostClusterConfig(clusterName, cloudCredential.ID)
+
+	logrus.Infof("Creating cluster %s...", clusterName)
+	azClient, err := resources.CreateAKSCluster(clusterName, cloudCredential.AzureCredentialConfig.SubscriptionID, cloudCredential.AzureCredentialConfig.ClientID, aksHostCluster.ResourceGroup, cred)
+	if err != nil {
+		logrus.Fatalf("FAILURE: Unable to create cluster: %+v", err)
+	} else {
+		_, err = azClient.PollUntilDone(ctx, nil)
+		if err != nil {
+			logrus.Fatalf("FAILURE: Unable to create cluster: %+v", err)
```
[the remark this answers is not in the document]
igomez06: No, fatals completely exit the application, and we should only use them if we have a specific reason. Why wouldn't we just return an error here? Fatal means something drastically broke it exits the program, and therefore none of our cleanup would run.

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
igomez06: agreed this needs to be a context.TODO, or else it would hang since there is no cancel.

tests/framework/extensions/clusters/aks/importing/import_aks_cluster.go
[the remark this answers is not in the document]
markusewalker: That's my bad, that isn't meant to be `fatalf`, I'll go and change that.

go.mod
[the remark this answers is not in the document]
markusewalker: These were added once I started using the azure sdk, unsure if it's a needed dependency. I can further look into that.

markusewalker:
Met offline with @igomez06 @Josh-Diamond @caliskanugur regarding design changes and ultimately what we want future tests to look like. As there will be several changes needed, going to move this into a draft state and bring it back out after.

markusewalker:
Discussed with Israel and Brandon offline. I'm going to move back to draft again, but push changes discussed here. Due to competing priorities with K8s 1.25, this is going to be put on the back-burner.

Once we're ready to come back to this, I'll open this PR back up at that time.

go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.11.27
 	github.com/Azure/go-autorest/autorest/adal v0.9.20
 	github.com/Azure/go-autorest/autorest/to v0.4.1-0.20210111195520-9fc88b15294e
-	github.com/AzureAD/microsoft-authentication-library-for-go v0.5.1
+	github.com/AzureAD/microsoft-authentication-library-for-go v0.7.0
```
igomez06, line 61: Hey Markus, do you know why we have such a significant version bump for this package and the other?

tests/v2/validation/provisioning/hosted/aks/provisioning/hosted_provisioning_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+			err = wait.WatchWait(watchInterface, checkFunc)
+			require.NoError(h.T(), err)
+			assert.Equal(h.T(), clusterName, clusterResp.Name)
+
+			clusterToken, err := clusters.CheckServiceAccountTokenSecret(client, clusterName)
+			require.NoError(h.T(), err)
+			assert.NotEmpty(h.T(), clusterToken)
+
+			updatedCluster, err := ScalingAKSNodePools(client, clusterResp, clusterName, cloudCredential.ID)
```
igomez06, line 105: @caliskanugur @markusewalker  so there will be some copy code, but should the scaling be it's own test case? Ie `(h func (h *HostedClusterProvisioningTestSuite) TestProvisioningHostedAKSClusterNodeScaling()` in case we want to add other validations to the main provisioning test. I would love to hear your opinions.

igomez06 (commented):
Overall looks good, just had some questions that should be very easy to address.

go.mod
markusewalker: Not sure, but initially I had thought it was with some of the import changes. But once I removed the import entries, this version bump still stayed. We could always try to push it back to 0.5.1 and I run a quick test?

tests/v2/validation/provisioning/hosted/aks/provisioning/hosted_provisioning_test.go
markusewalker: I don't mind doing that, makes sense if we elect to add other validations to the main provisioning test. I'm open to whatever the group thinks is best.

go.mod
igomez06: I think that would be great, I would rather dev be in charge of version bumps instead of us doing it unintentionally

tests/v2/validation/provisioning/hosted/aks/provisioning/hosted_provisioning_test.go
caliskanugur: I think that makes sense to me. Code duplication can be prevented by doing this in a private method for the test suite as a cluster provisioning part. 

@markusewalker if you check the upgrade, you can do something like testSingleUpdate, or RBAC tests are also a good example of that. 

This private method I mentioned can be something like that. One cluster provisioning private method of the test suite used in:

- the provisioning test alone
- the scale bits followed after the provisioning test

igomez06: That's a great idea. It's your call Markus, I would be ok with the duplicate for now, with the scaling bit, and come back when you add more test cases, or do it now. Your call. But I do think it should be it's own test case.

go.mod
caliskanugur: I agree with that. That'd be really great to discard all `go.mod` and `go.sum` changes and try it, as these changes might affect other things in the rancher.

tests/v2/validation/provisioning/hosted/aks/provisioning/hosted_provisioning_test.go
markusewalker: @igomez06 @caliskanugur pushed changes that refactors based on Ugur's suggestion. Also, I went and did the same for EKS and GKE as well.

go.mod
markusewalker: Tested locally and it was fine. Unsure why latest commit doesn't show those changes, will try again.

markusewalker:
@igomez06 @caliskanugur pushed latest changes from our last discussion. I'll port them to the backport as well.

tests/v2/validation/provisioning/hosted/aks/scaling/scaling_nodepools.go
```diff
@@ -0,0 +1,102 @@
...
+	np "github.com/rancher/rancher/tests/framework/extensions/clusters/aks/resources"
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+// ScalingAKSNodePools is a helper function that tests scaling of an AKS node pool by adding a new one and then deleting it.
+func ScalingAKSNodePools(client *rancher.Client, oldCluster *management.Cluster, displayName string, cloudCredential *cloudcredentials.CloudCredential) (*management.Cluster, error) {
+	logrus.Infof("======================================")
+	logrus.Infof("TEST CASE: Scaling AKS Node Pools")
+	logrus.Infof("======================================")
```
igomez06, line 19: No a fan of these log statements, as this isn't a test case more of an action

igomez06 (commented):
Looks great!!!! Just one nitpick, but great job!!

tests/v2/validation/provisioning/hosted/aks/scaling/scaling_nodepools.go
markusewalker: Want me to just remove it altogether? My original thought behind was to make it a bit easier to distinguish between test cases. But I guess since now we have it split into different test cases, doesn't really matter too much now huh?

markusewalker: I removed it from previous comment I made. Doesn't really need to be there now that we refactored the test functions anyway.

igomez06: Yes totally agree they don't need to be there at all

igomez06 (dismissed):
Looks good

tests/v2/validation/provisioning/hosted/aks/scaling/scaling_nodepools.go
```diff
@@ -0,0 +1,98 @@
...
+	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
+	"github.com/rancher/rancher/tests/framework/extensions/cloudcredentials"
+	"github.com/rancher/rancher/tests/framework/extensions/clusters/aks"
+	np "github.com/rancher/rancher/tests/framework/extensions/clusters/aks/resources"
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+// ScalingAKSNodePools is a helper function that tests scaling of an AKS node pool by adding a new one and then deleting it.
+func ScalingAKSNodePools(client *rancher.Client, oldCluster *management.Cluster, displayName string, cloudCredential *cloudcredentials.CloudCredential) (*management.Cluster, error) {
```
igomez06, line 16: Markus I just realized something, I don't think this func should be in its own package, since it's a helper not its own test. My last nitpick

markusewalker: Sure thing, I'll remove it and port the change to the backport too.

markusewalker: Alright, made the change to both this and the 2.6 PR. Should be good now 😄 . Thanks for the catch!

caliskanugur:
Looks good, Markus could you provide the freeforms for these changes in our usual channel?

igomez06 (dismissed):
LGTM

caliskanugur (dismissed):

igomez06 (approved):
LGTM

caliskanugur (approved):
LGTM

slickwarren (approved):

Issue #38701, opened 2022-08-23
[Feature] k8s 1.25 Support


snasovich:
Mostly placeholder. Implement k8s 1.25 support. Research is done as per comments on https://github.com/rancher/rancher/issues/37734, also https://github.com/rancher/rancher/issues/38132#issuecomment-1206826017).

SURE-4603

yankees5963:
Would help to get 100% functioning 1.22/23/24 support prior to 1.25

cbron:
@jiaqiluo some design comments:

>  Along with a new action named setpodsecurity on the object:

We should avoid using action's, that was a design pattern influenced by the cluster manager UI. Ideally the new UI forms would just edit the field directly, and then a controller would act on them. We could put some validations in the webhook, but our delineating line is if the object can be created, so for most everything here the failures/errors would come from controller. 

> If PSP is enabled before the upgrade, Rancher will delete related internal resources after the upgrade.

This is interesting point, all the upstream PSPs will get auto-deleted, but our stuff will have to get manually cleaned up. There is cluster/project/ns stuff. Also I'm guessing our controllers may be acting on PSP stuff, and those would have to be able to handle 1.25.

> Move a Namesape from Project A to Project B, or to Non Project

This sounds like same logic we have already for PSPs, we just need a new controller to handle everything.

jiaqiluo:
@cbron  Thank you for the feedback!  
The Design is updated to drop the requirements for action on Project.

ayaranga:
When can we expect k8s 1.25 support in Rancher to be available? Thanks.

sowmyav27:
RKE2/k3s 1.25 support is tracked here - https://github.com/rancher/rancher/issues/39795

Sahota1225:
https://github.com/rancher/rancher/issues/39960 -make sure 39960 is updated before closing this issue

nlahmi:
Hey, is support for 1.25(.5) coming soon?

markusewalker:
The internal test plan has been completed in full and all associated GH issues with this issue and the test plan have been validated and closed. As such, closing this issue.

martyav:
@snasovich Re https://github.com/rancher/rancher/issues/38701#issue-1348453471:

> A new field will be added to represent the pod security admission config being applied to the cluster.
> Its named is default_pod_security_admission_configuration_template_id on Cluster.management.cattle.io/v3 CR
> [...]
[...]

PR #39806, opened 2022-12-07
psact update

HarrisonWAffel:
https://github.com/rancher/rancher/issues/38701

### DIFF
--- a/pkg/api/steve/disallow/disallow.go
+++ b/pkg/api/steve/disallow/disallow.go
@@ -10,6 +10,9 @@ import (
 )
 
 var (
+	allowAll = map[string]bool{
+		"podsecurityadmissionconfigurationtemplates": true,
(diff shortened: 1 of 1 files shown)


pkg/api/steve/disallow/disallow.go
```diff
@@ -27,13 +27,19 @@ var (
 		"templateversions":          true,
 		"tokens":                    true,
 	}
+	mgmtWhiteList = map[string]bool{
```
snasovich, line 30: Somewhat of a not, but I think it's best to name it `allowAll` to match the existing maps naming style and move it to the top of the var block definition.

snasovich (changes requested):
One comment to address. Please also update the description to link to the issue.

snasovich (approved):

PR #39816, opened 2022-12-08
k3s-cert traefik ingress fixes

vivek-infracloud:
## Issue: 

The test_ingress_rule_with_only_path test fails because the way Traefik ingress handles the ingress.
This PR is one of the possible fix for the wrong failure.

k3s clusters uses Traefik ingress manager, hence, the current changes in the code checks if the RANCHER_K3S_VERSION is not empty.
If so, the ingress rule created is slightly different than if the RANCHER_K3S_VERSION is empty.

### DIFF
--- a/tests/validation/tests/v3_api/test_ingress.py
+++ b/tests/validation/tests/v3_api/test_ingress.py
@@ -40,6 +40,7 @@ from .common import DEFAULT_TIMEOUT
 from .common import rbac_get_workload
 from .common import wait_for_ingress_to_active
 from .common import get_setting_value_by_name
+from .test_import_k3s_cluster import RANCHER_K3S_VERSION
 from .test_secrets import is_version_greater_than_v25
 from packaging import version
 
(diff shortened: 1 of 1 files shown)


markusewalker (approved):

izaac (approved):
LGTM 
Please open the backport @vivek-infracloud

tests/validation/tests/v3_api/test_ingress.py
```diff
@@ -332,10 +333,15 @@ def test_ingress_rule_with_only_path():
...
-    p_client.create_ingress(name=name,
+    if RANCHER_K3S_VERSION != "":
+        rule = {"host": host,
+                "paths": [{"workloadIds": [workload.id],
+                           "targetPort": TEST_IMAGE_PORT, "pathType": "Prefix", "path": "/"}]}
+    else:
+        rule = {"host": host,
+                "paths": [{"workloadIds": [workload.id],
+                           "targetPort": TEST_IMAGE_PORT}]}
+        p_client.create_ingress(name=name,
```
anupama2501, line 344: an ingress will only be created when the K3s version is set here. Is that the intention?

vivek-infracloud: No. If the k3s version is set the ingress will be created with `"pathType": "Prefix", "path": "/"` else the ingress will be created without pathType and Prefix.

anupama2501: So, shouldn't this ingress be outside of the else part and be run for both conditions?

vivek-infracloud: Right, I corrected the code.

markusewalker (approved):

anupama2501 (approved):
lgtm

izaac (approved):
LGTM

PR #39843, opened 2022-12-09
added clusterscan type schema back

vardhaman22:
removed cis related types from cluster scan schema 

## Issue: 

 https://github.com/rancher/dashboard/issues/7713
## Problem

 due to clusterscan schema not present old ui calls to apis related to cluster scan is failing.
## Solution

 added clusterscan schema back


vardhaman22:
the issue was fixed by ui changes and these schemas are no longer needed.

PR #39864, opened 2022-12-12
[2.6] Azure AD - remove permissions checks and add an auth config controller

maxsokolovsky:
## Problem and Solution
We want to allow Azure AD admins to be responsible for their own permission set they are ready to grant to an app in Azure.
That way, they can use narrower permissions (like `Directory.Read.All`, instead of `Group.Read.All + User.Read.All`, as both will work with Azure AD actions). However, admins are themselves responsible for making sure they pick those permissions that will fully work with Azure AD in Rancher.

We want to do this to avoid having to maintain a large array of permissions that we check and validate. There are a few permissions that work with all Azure AD actions that Rancher normally performs (logging in, fetching users, groups, etc.). However, customers might have different preferences as to which exact Application permissions to grant to their apps. We don't want to have complicated validating logic, so we instead allow any permissions. If customers choose insufficient permissions, then there will be a failure in Rancher at some point, which they can rectify by quickly disabling or reconfiguring the auth provider and then adjusting their permissions.

Changes proposed:
1. Do not check access token permissions from Microsoft Graph at all. Neither on login, nor on endpoint updates or initial provider setup. This means that any successful attempt to set up Azure AD, for example, would lead to the access token being cached in a secret.
2. Introduce an auth provider controller that would delete the cached access token secret, so that admins who make a mistake in choosing the necessary app permissions in Azure can easily disable the auth provider, fix their permissions, and try again without the cache holding a bad access token for up to an hour before an automatic refresh.


maxsokolovsky:
# QA testing

### Context
There are two secrets to watch, both in the `cattle-global-data` namespace.
1. `azuread-access-token` - the access token for Microsoft Graph API.
2. `azureadconfig-applicationsecret` - the Application of the app in Azure.

### 1. Good token
1. In your Azure app, remove all Microsoft Graph permissions.
Add only the Application permission of type `Directory.Read.All`.
1. Run Rancher v2.7-head, set up Azure AD.
1. No errors.


### 2. Bad token
1. In your Azure app, remove all Microsoft Graph permissions (or simply revoke admin consent).
Youp app does not have permissions for Rancher to successfully run different actions against Azure AD.
1. Run Rancher v2.7-head, set up Azure AD and use your app.
1. There is an error, since on setup, Rancher tries to fetch the user info you logged in as.
1. Check both secrets, make sure they **don't** exist (might need to refresh).
1. Add `Directory.Read.All` permission to the app.
1. Try setting up Azure AD again.
1. No errors.
1. Check the secrets, make sure they exist.

Now try a different set of permissions and ensure they work as before.
1. Revoke consent for the `Directory.Read.All` permission in the app.
1. Add `Group.Read.All` and `User.Read.All`.
1. Disable Azure AD. No errors.
1. Check the `azuread-access-token` and `azureadconfig-applicationsecret` secrets. Make sure both have been deleted.
1. Try setting up Azure AD again.
1. No errors.
1. Check both secrets, make sure they exist.

MbolotSuse (approved):

Issue #39841, opened 2022-12-09
[Enhancement] add probe customization to the Rancher chart


samjustus:
Breakout from: https://github.com/rancher/rancher/issues/38177
Adds values to the chart for rancher server to override the livenessProbe and readinessProbe's initialDelaySeconds and periodSeconds

samjustus:
@cbron  needs PR

jiaqiluo:
Can be validated in Rancher [v2.7.2-rc1](https://github.com/rancher/rancher/releases/tag/v2.7.2-rc1)

brudnak:
# ✅ PASSED

## Reproduction Environment

#### Not required.

<hr>

## Validation Environment


| Component                        | Version / Type           |
| -------------------------------- | ------------------------ |
| Rancher version                  | v2.7.2-rc1               |
| Installation option              | Helm (high availability) |
| RKE binary version used          | v1.4.0                   |
| If Helm Chart k8s cluster        | v1.24.6                  |
| Cert Details                     | Let's Encrypt / `nginx`  |
| Docker version                   | 20.10.7, build f0df350   |
| Helm version                     | v2.16.8-rancher2         |
| Downstream cluster type          | not applicable           |
| Downstream K8s version           | not applicable           |
| Authentication providers enabled | local                    |
| Logged in user role              | administrator            |
| Browser type                     | Google Chrome            |
| Browser version                  | 108.0.5359.124           |


<details>
    <summary>🚨 Additional Validation Setup Details: Click to Expand</summary>

```shell
#!/bin/sh

helm repo update

kubectl create namespace cattle-system

kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.crds.yaml

helm install cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --create-namespace \
  --version v1.7.1

helm install rancher rancher-latest/rancher \
  --namespace cattle-system \
  --set hostname=$URL_VAR \
  --set bootstrapPassword=$PASSWORD_VAR \
  --set ingress.tls.source=letsEncrypt \
  --set letsEncrypt.email=$EMAIL_VAR \
  --set letsEncrypt.ingress.class=nginx \
  --set rancherImageTag=v2.7.2-rc1 \
  --version 2.7.2-rc1 \
  --set livenessProbe.initialDelaySeconds=77 \
  --set livenessProbe.periodSeconds=77 \
  --set readinessProbe.initialDelaySeconds=77 \
[...]

PR #39865, opened 2022-12-12
Adds chart option to configure the initialDelay and periodSeconds of probes

cbron:
## Issue: 

https://github.com/rancher/rancher/issues/39841
 
## Problem

Users need to be able to override probes as sometimes Rancher takes a very long time to start. If they do that manually to the deployment, helm will override it the following upgrade. 
 
## Solution

This PR allows users to set probes via helm directly.
 
### Manual Testing

Ran helm template, helm install, and all the helm unit tests

### Automated Testing

Tests added, originally from: https://github.com/rancher/rancher/pull/39319/commits/fd4caac9e835ad862ce3869b8fd66c9e9bfeafbd

Should not cause any regressions, as this will operate functionally the same.

### DIFF
--- a/chart/tests/README.md
+++ b/chart/tests/README.md
@@ -1,0 +1,33 @@
+
+## local dev testing instructions
+
+Option 1: Full chart CI run with a live cluster
+
+```bash
+./scripts/charts/ci 
+```
+
+Option 2: Test runs against the chart only 
+
+```bash
+# Install plugin if necessary, for version see CATTLE_HELM_UNITTEST_VERSION
+# For a Mac you'll need to download a release and install the darwin binary
+# For linux:
+export ARCH=amd64 export CATTLE_HELM_UNITTEST_VERSION={version} helm plugin install https://github.com/rancher/helm-unittest
+
+# change automated parts of templates
+test_image="rancher/rancher"
+test_image_tag="v2.7.0"
+sed -i -e "s/%VERSION%/${test_image_tag}/g" ./chart/Chart.yaml
+sed -i -e "s/%APP_VERSION%/${test_image_tag}/g" ./chart/Chart.yaml
+sed -i -e "s@%POST_DELETE_IMAGE_NAME%@${test_image}@g" ./chart/values.yaml
+sed -i -e "s/%POST_DELETE_IMAGE_TAG%/${test_image_tag}/g" ./chart/values.yaml
+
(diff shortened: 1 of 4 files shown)


chart/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+# change automated parts of templates
+test_image="rancher/rancher"
+test_image_tag="v2.7.0"
+sed -i -e "s/%VERSION%/${test_image_tag}/g" ./chart/Chart.yaml
+sed -i -e "s/%APP_VERSION%/${test_image_tag}/g" ./chart/Chart.yaml
+sed -i -e "s@%POST_DELETE_IMAGE_NAME%@${test_image}@g" ./chart/values.yaml
+sed -i -e "s/%POST_DELETE_IMAGE_TAG%/${test_image_tag}/g" ./chart/values.yaml
+
+# install plugin if necessary
+helm plugin install --version {version}  https://github.com/rancher/helm-unittest
```
cmurphy, line 15: This didn't work for me, I had to do

```
export ARCH=amd64
export CATTLE_HELM_UNITTEST_VERSION=v0.1.7-rancher4
helm plugin install  https://github.com/rancher/helm-unittest
```

chart/tests/README.md
```diff
@@ -0,0 +1,25 @@
+
+### local dev testing instructions
```
cmurphy, line 2: Couldn't we just run ./scripts/charts/ci ?

chart/tests/README.md
cbron: Looks like that works on a linux box because of how the script is setup: https://github.com/rancher/helm-unittest/blob/master/install-binary.sh#L10

No support for mac in there, but the darwin binary does exist on the release: https://github.com/rancher/helm-unittest/releases/

I'll add both options to readme.

chart/tests/README.md
cbron: Because we use things that require an active cluster in order to run, like `lookup`. My main goal here is to let people run `helm unittest` directly against their tests, without things like a cluster or full chart generation. This will be much quicker, though arguably less real than a full test. So I added both options to the readme.

cmurphy (approved):

crobby (approved):
Could have a couple newlines at the end of the yaml files, otherwise, lgtm

PR #39871, opened 2022-12-12
Add CAPI predicates to RKEBootstrap

jakefhyde:
## Issue: 

Part of groundwork for #39139

## Problem

## Solution

Add missing predicates to rke bootstrap controller, including checking for pausing as well as other required predicates we were missing.

https://cluster-api.sigs.k8s.io/developer/providers/bootstrap.html?highlight=bootstrap#bootstrap-provider-specification

**Note:** there are some predicates which have been added, which do not appear in the upstream docs as requirements, however these predicates (for example, checking if `capiCluster.Status.InfrastructureReady == false`) are added because although we functionally always set `InfrastructureReady = true`, bootstrapping before the cluster is ready is not desired and is subject to change in the future.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
@@ -56,6 +56,9 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 		rkeControlPlanes:    clients.RKE.RKEControlPlane().Cache(),
 		rkeBootstrap:        clients.RKE.RKEBootstrap(),
 	}
+
+	clients.RKE.RKEBootstrap().OnChange(ctx, "rke-bootstrap-cluster-name", h.OnChange)
+
 	rkecontroller.RegisterRKEBootstrapGeneratingHandler(ctx,
 		clients.RKE.RKEBootstrap(),
 		clients.Apply.
@@ -68,7 +71,7 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 			WithSetOwnerReference(true, true),
 		"",
 		"rke-bootstrap",
-		h.OnChange,
+		h.GeneratingHandler,
 		nil)
 
 	relatedresource.Watch(ctx, "rke-bootstrap-trigger", func(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioningv2/rke2/rkebootstrap/controller.go
```diff
@@ -1,10 +1,10 @@
-package bootstrap
```
Oats87, line 1: I would prefer we do not rename packages at this time. It makes maintenance very difficult. We'll be much better off doing this once we've stabilized our code base.

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -276,31 +272,58 @@ func (h *handler) OnChange(bootstrap *rkev1.RKEBootstrap, status rkev1.RKEBootst
 		result []runtime.Object
 	)
 
+	machine, err := rke2.GetOwnerCAPIMachine(bootstrap, h.machineCache)
+	if apierrors.IsNotFound(err) {
```
annablender, line 276: This is fine the way you have it. I feel like it's more `Go style` for error handling but personally, I'd put the `IsNotFound` checks inside the `err != nil` blocks. That makes more logical sense but again I see the efficiency of short circuiting here

annablender (dismissed):

HarrisonWAffel (dismissed):
This looks good and matches the conditions within the CAPI docs

pkg/controllers/provisioningv2/rke2/bootstrap/controller.go
```diff
@@ -20,12 +20,15 @@ import (
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/relatedresource"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierror "k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
Oats87, line 27: this is imported already

Oats87 (changes requested):

HarrisonWAffel (approved):

HarrisonWAffel (approved):

Oats87 (approved):
LGTM!

PR #39872, opened 2022-12-12
Add CAPI predicates to RKEMachine

jakefhyde:
## Issue: 

Part of groundwork for #39139

## Problem

## Solution

Add missing predicates to rke machine controller, including checking for pausing as well as other required predicates we were missing.

https://cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure.html#normal-resource

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
@@ -193,7 +193,8 @@ func (h *handler) getMachineStatus(job *batchv1.Job) (rkev1.RKEMachineStatus, er
 	if job.Spec.Template.Labels[InfraJobRemove] == "true" {
 		condType = deleteJobConditionType
 	}
-	if !job.Status.CompletionTime.IsZero() {
+
+	if condition.Cond("Complete").IsTrue(job) {
 		return rkev1.RKEMachineStatus{
 			Conditions: []genericcondition.GenericCondition{
 				{
@@ -207,9 +208,7 @@ func (h *handler) getMachineStatus(job *batchv1.Job) (rkev1.RKEMachineStatus, er
 			},
 			JobComplete: true,
 		}, nil
-	}
-
-	if condition.Cond("Failed").IsTrue(job) {
+	} else if condition.Cond("Failed").IsTrue(job) {
 		sel, err := metav1.LabelSelectorAsSelector(job.Spec.Selector)
 		if err != nil {
 			return rkev1.RKEMachineStatus{}, err
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
@@ -43,6 +43,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 	capierrors "sigs.k8s.io/cluster-api/errors"
+	capiannotations "sigs.k8s.io/cluster-api/util/annotations"
 )
 
 const (
@@ -91,8 +92,9 @@ type handler struct {
 	jobs                batchcontrollers.JobCache
 	pods                corecontrollers.PodCache
 	secrets             corecontrollers.SecretCache
-	machines            capicontrollers.MachineCache
-	machinesClient      capicontrollers.MachineClient
+	capiClusterCache    capicontrollers.ClusterCache
+	machineCache        capicontrollers.MachineCache
+	machineClient       capicontrollers.MachineClient
 	namespaces          corecontrollers.NamespaceCache
 	nodeDriverCache     mgmtcontrollers.NodeDriverCache
 	dynamic             *dynamic.Controller
@@ -112,8 +114,9 @@ func Register(ctx context.Context, clients *wrangler.Context) {
 		jobController:       clients.Batch.Job(),
 		jobs:                clients.Batch.Job().Cache(),
 		secrets:             clients.Core.Secret().Cache(),
-		machines:            clients.CAPI.Machine().Cache(),
-		machinesClient:      clients.CAPI.Machine(),
+		machineCache:        clients.CAPI.Machine().Cache(),
+		machineClient:       clients.CAPI.Machine(),
+		capiClusterCache:    clients.CAPI.Cluster().Cache(),
 		nodeDriverCache:     clients.Mgmt.NodeDriver().Cache(),
 		namespaces:          clients.Core.Namespace().Cache(),
 		dynamic:             clients.Dynamic,
@@ -161,9 +164,9 @@ func (h *handler) OnJobChange(_ string, job *batchv1.Job) (*batchv1.Job, error) 
 		return job, err
 	}
 
-	infraObj, err := newInfraObject(infraMachine)
+	infra, err := newInfraObject(infraMachine)
 	if err != nil {
-		return nil, err
+		return job, err
 	}
 
 	newStatus, err := h.getMachineStatus(job)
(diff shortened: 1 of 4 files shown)


pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -354,33 +357,39 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
+		logrus.Errorf("[rkemachine] %s/%s: error getting machine by owner reference %v", infra.meta.GetNamespace(), infra.meta.GetName(), err)
 		return obj, err
 	}
 
-	if machine == nil || machine.Status.NodeRef == nil {
+	if machine.Status.NodeRef == nil {
 		// Machine noderef is nil, we should just allow deletion.
-		logrus.Debugf("[MachineProvision] There was no associated K8s node with this etcd dynamicmachine %s. Proceeding with deletion", key)
-		return h.doRemove(infraObj)
+		logrus.Debugf("[rkemachine] There was no associated K8s node with this machine %s. Proceeding with deletion", key)
```
jakefhyde, line 378: I know that this controller isn't registered as the `rke-machine` controller, and I would've updated it to be named that, but:

1. IMO it should've been named that originally
2. I only didn't update it to avoid finalizer migration.

I see 3 potential options:

1. Leave log message prefix as `rkemachine` and handler registration as `machine-provision`
2. Rename handlers and package to `rkemachine` to be in line with the other controllers in this package and do finalizer migration somewhere.
3. Convert log message prefix to `machineprovision`.

Oats87: I would be OK with #1 for now, given the caveat that we know how to grok for the logs for this controller.

annablender (approved):
LGTM, but may need to go over it again in the morning with fresh eyes

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -413,18 +422,65 @@ func (h *handler) doRemove(infraObj *infraObject) (runtime.Object, error) {
 	return obj, generic.ErrSkip
 }
 
+func (h *handler) EnqueueAfter(infra *infraObject, duration time.Duration) {
+	err := h.dynamic.EnqueueAfter(infra.obj.GetObjectKind().GroupVersionKind(), infra.meta.GetNamespace(), infra.meta.GetName(), duration)
+	if err != nil {
+		panic(err) // always programmer error
```
HarrisonWAffel, line 428: Why do we want to `panic` here? Is it just to make the issue really obvious in the logs? Other uses of `h.dynamic.EnqueueAfter` in this file handle the error instead of just panicking

jakefhyde: Opted to log the error instead, its not really any error that can be handled in a meaningful way

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -435,41 +491,44 @@ func (h *handler) OnChange(obj runtime.Object) (runtime.Object, error) {
...
 		if err != nil {
 			return nil, err
 		}
 
 		if failedCreate {
-			logrus.Infof("[MachineProvision] Failed to create infrastructure %s/%s for machine %s, deleting and recreating...", infraObj.meta.GetNamespace(), infraObj.meta.GetName(), dArgs.CapiMachineName)
-			if err = h.machinesClient.Delete(infraObj.meta.GetNamespace(), dArgs.CapiMachineName, &metav1.DeleteOptions{}); err != nil {
-				return infraObj.obj, fmt.Errorf("failed to delete CAPI machine %s: %w", dArgs.CapiMachineName, err)
+			logrus.Infof("[rkemachine] %s/%s: Failed to create infrastructure for machine %s, deleting and recreating...", infra.meta.GetNamespace(), infra.meta.GetName(), dArgs.CapiMachineName)
+			// TODO(jhyde): determine if it's possible to just delete the object instead of the CAPI object
```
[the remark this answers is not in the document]
jakefhyde: Just removed the todo for now, a change like this would probably be too much of a deviation for now.

HarrisonWAffel (dismissed):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -161,9 +164,9 @@ func (h *handler) OnJobChange(_ string, job *batchv1.Job) (*batchv1.Job, error)
 		return job, err
 	}
 
-	infraObj, err := newInfraObject(infraMachine)
+	infra, err := newInfraObject(infraMachine)
 	if err != nil {
-		return nil, err
+		return job, err
```
jakefhyde, line 169: @Oats87 IIRC if `err != nil`, the `OnChange` handler will discard the returned object, so this is merely a stylistic choice.

HarrisonWAffel (approved):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -354,33 +357,39 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
...
 		return obj, err
 	}
 	if apierrors.IsNotFound(err) || !cluster.DeletionTimestamp.IsZero() {
-		return h.doRemove(infraObj)
+		return h.doRemove(infra)
 	}
 
 	removed := true
 	// In the event we are removing an etcd node (as indicated by the etcd-role label on the node), we must safely remove the etcd node from the cluster before allowing machine deprovisioning
-	if val := infraObj.meta.GetLabels()["rke.cattle.io/etcd-role"]; val == "true" {
```
Oats87, line 383: nit: I don't know who put this here (was it me?) but this should probably using a strconv tolower to handle cases where `true` is not all lowercase

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -413,18 +422,65 @@ func (h *handler) doRemove(infraObj *infraObject) (runtime.Object, error) {
 	return obj, generic.ErrSkip
 }
 
+func (h *handler) EnqueueAfter(infra *infraObject, duration time.Duration) {
+	err := h.dynamic.EnqueueAfter(infra.obj.GetObjectKind().GroupVersionKind(), infra.meta.GetNamespace(), infra.meta.GetName(), duration)
+	if err != nil {
```
Oats87, line 427: Should we do something with this error other than just logging it?

Oats87 (approved):
More or less, you can merge this now, but we I did leave a few comments.

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
jakefhyde: I originally had a panic, since this is the summation of `EnqueueAfter`:
https://github.com/rancher/lasso/blob/6ea88ca2fbcc8cadc89c30421b0766e9993c4b7a/pkg/dynamic/controller.go#L349-L357
If this happens, everything is spectacularly broken.

Oats87 (approved):

HarrisonWAffel (approved):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -193,7 +193,8 @@ func (h *handler) getMachineStatus(job *batchv1.Job) (rkev1.RKEMachineStatus, er
 	if job.Spec.Template.Labels[InfraJobRemove] == "true" {
 		condType = deleteJobConditionType
 	}
-	if !job.Status.CompletionTime.IsZero() {
```
jakefhyde, line 193: @Oats87 If we reconcile the job deletion even though it was successful, it would reset the `Ready` condition to false and the infrastructure would never be considered Ready by CAPI.

Oats87: this is the kind of stuff I worry about with us messing with all of these conditions.

jakefhyde: Me too, FWIW I used to be able to reproduce this intermittently on 2.6, so we never fixed this issue, its still just waiting to happen in 2.6 release line.

Oats87 (approved):

HarrisonWAffel (approved):

PR #39873, opened 2022-12-12
[v2.6] Updated create_project_client in test_node_label to have an optional cleanup with the boolean flag RANCHER_CLEANUP_CLUSTER

igomez06:
## Problem

 The tests in tests/validation/tests/v3api/test_node_label.py would clean the clusters after tests. There should be option where we can disable this.
## Solution

 Added the RANCHER_CLEANUP_CLUSTER flag in the clean up portion of that test.
## Testing
 Jenkins runs with both clean up set to true or false.

### DIFF
--- a/tests/validation/tests/v3_api/test_node_label.py
+++ b/tests/validation/tests/v3_api/test_node_label.py
@@ -14,6 +14,7 @@ from .common import get_user_client_and_cluster
 from .common import execute_kubectl_cmd
 from .common import if_test_rbac
 from .common import LINODE_ACCESSKEY
+from .common import RANCHER_CLEANUP_CLUSTER
 from .common import random_name
 from .common import random_test_name
 from .common import rbac_get_user_token_by_role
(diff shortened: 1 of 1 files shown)


izaac (dismissed):
LGTM

anupama2501 (dismissed):
lgtm

tests/validation/tests/v3_api/test_node_label.py
```diff
@@ -923,43 +924,43 @@ def create_project_client(request):
...
-
-    def fin():
-        client = get_user_client()
-        cluster = cluster_node_template["cluster"]
-        if cluster is not None:
-            node_pools_list = client.list_nodePool(clusterId=cluster.id).data
-            # get unique node template ids
-            client.delete(cluster_node_template["cluster"])
-            wait_for_cluster_delete(client, cluster["name"])
+    if RANCHER_CLEANUP_CLUSTER:
```
bmdepesa, line 927: It would be cleaner to have the conditional wrap the `request.addfinalizer(fin)` statement.

bmdepesa (changes requested):

tests/validation/tests/v3_api/test_node_label.py
igomez06: Good call, I updated the conditional.

bmdepesa (approved):

izaac (approved):
LGTM

PR #39875, opened 2022-12-13
Add CAPI predicates to RKECluster

jakefhyde:
## Issue: 

Part of groundwork for #39139

## Problem

## Solution

Add missing predicates to rke cluster controller, including checking for pausing as well as other required predicates we were missing. Also remove sync with the controlplane object for `observedGeneration`.

https://cluster-api.sigs.k8s.io/developer/architecture/controllers/cluster.html?highlight=infrastructure%20cluster#infrastructure-provider


pkg/client/generated/management/v3/zz_generated_node_group.go
```diff
@@ -33,6 +34,7 @@ type NodeGroup struct {
 	LaunchTemplate       *LaunchTemplate   `json:"launchTemplate,omitempty" yaml:"launchTemplate,omitempty"`
 	MaxSize              *int64            `json:"maxSize,omitempty" yaml:"maxSize,omitempty"`
 	MinSize              *int64            `json:"minSize,omitempty" yaml:"minSize,omitempty"`
+	NodeRole             *string           `json:"nodeRole,omitempty" yaml:"nodeRole,omitempty"`
```
Oats87, line 37: Did we miss a go generate or something? Where is this coming from?

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
```diff
@@ -37,38 +38,44 @@ func Register(ctx context.Context, clients *wrangler.Context) {
...
+		return cluster, generic.ErrSkip
+	}
+	if err != nil {
+		logrus.Errorf("[rkecluster] %s/%s: error getting CAPI cluster %v", cluster.Namespace, cluster.Name, err)
+		return cluster, err
+	}
+
+	if capiannotations.IsPaused(capiCluster, cluster) {
+		logrus.Infof("[rkecluster] %s/%s: waiting: CAPI cluster or RKECluster is paused", cluster.Namespace, cluster.Name)
+		h.rkeCluster.EnqueueAfter(cluster.Namespace, cluster.Name, 10*time.Second)
```
Oats87, line 59: Do we really need to re-enqueue after 10 seconds here? This seems rudimentary compared to the tools we have available. We could use related resource to watch for the capi cluster to change...?

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
```diff
@@ -37,38 +38,44 @@ func Register(ctx context.Context, clients *wrangler.Context) {
...
-		conditionToUpdate = rke2.Removed
-	}
-	cp, err := h.rkeControlPlanes.Get(rkeCluster.Namespace, rkeCluster.Name)
-	if err == nil {
-		conditionToUpdate.SetStatus(&status, conditionToUpdate.GetStatus(cp))
-		conditionToUpdate.Reason(&status, conditionToUpdate.GetReason(cp))
-		conditionToUpdate.Message(&status, conditionToUpdate.GetMessage(cp))
-	} else if !apierrors.IsNotFound(err) {
-		return status, err
+	if cluster.Status.Ready != true {
```
Oats87, line 73: Do we not want to do a 1-2 punch for setting controlplane endpoint and then marking the cluster as ready? Although I guess I don't know if we can have a ready cluster when the controlplane endpoint isn't set...?

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
```diff
@@ -37,38 +38,44 @@ func Register(ctx context.Context, clients *wrangler.Context) {
...
-}
-
-func (h *handler) OnChange(rkeCluster *v1.RKECluster, status v1.RKEClusterStatus) (v1.RKEClusterStatus, error) {
-	conditionToUpdate := rke2.Ready
-	if !rkeCluster.DeletionTimestamp.IsZero() {
-		conditionToUpdate = rke2.Removed
-	}
-	cp, err := h.rkeControlPlanes.Get(rkeCluster.Namespace, rkeCluster.Name)
-	if err == nil {
-		conditionToUpdate.SetStatus(&status, conditionToUpdate.GetStatus(cp))
```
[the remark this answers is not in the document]
jakefhyde: The condition here is actually useless, we can remove it if necessary, but neither we nor CAPI check the rkecluster object for this condition for anything.

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
jakefhyde: I considered that, but I chose against it for the following reason:
Our `rkecluster` object is effectively useless, solely existing to satisfy CAPI requirements. Maybe this is future proofing too much, but keeping them separate to me clarifies the intent that they are unrelated, distinct conditions that both have to be set. I'm not opposed to changing it to 1 update, as according to the upstream docs:
> The Cluster controller bubbles up spec.controlPlaneEndpoint and status.ready into status.infrastructureReady from the infrastructureCluster.

pkg/client/generated/management/v3/zz_generated_node_group.go
jakefhyde: It would appear that way, https://github.com/rancher/eks-operator/pull/73 introduced node role but a corresponding go generate was never run after merging into rancher: https://github.com/rancher/rancher/pull/39881

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
jakefhyde: This is what I came up with:
```
	relatedresource.Watch(ctx, "rke-cluster-trigger", func(namespace, name string, obj runtime.Object) ([]relatedresource.Key, error) {
		if capiCluster, ok := obj.(*capi.Cluster); ok && !capiCluster.Spec.Paused {
			return []relatedresource.Key{{
				Namespace: namespace,
				Name:      name,
			}}, nil
		}
		return nil, nil
	}, clients.RKE.RKECluster(), clients.CAPI.Cluster())
```
though I'm not sure this encapsulates what we want either, as we don't really need to enqueue the cluster/bootstrap/machine objects _while_ the CAPI cluster is unpaused, we just want to pause reconcilation _until_ the cluster is unpaused, which may require some wrangler updates.

Oats87: I don't see an issue if the controller reconciles the object based on related resource if the controller isn't going to make a lot of noise/be detrimental to the performance of rancher.

jakefhyde: It shouldn't affect performance, there will be a few extra reconcilations done until the PR that restructures the planner around these predicates goes in, since the CAPI cluster will continuously update the `status.InfrastructureReady` field and a few conditions, but even that is currently negligible. I'll go ahead and update.

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
jakefhyde: I also had to do it this way because you can't use `Update` to update status at the same time.

pkg/controllers/provisioningv2/rke2/rkecluster/controller.go
```diff
@@ -2,73 +2,85 @@ package rkecluster
...
-		conditionToUpdate.Reason(&status, conditionToUpdate.GetReason(cp))
-		conditionToUpdate.Message(&status, conditionToUpdate.GetMessage(cp))
-	} else if !apierrors.IsNotFound(err) {
-		return status, err
+	if len(cluster.Status.Conditions) > 0 || cluster.Status.Ready != true {
+		cluster := cluster.DeepCopy()
+		// the rke2.Ready and rke2.Removed conditions may still be present on the object, remove them if present
+		cluster.Status.Conditions = nil
```
Oats87, line 78: how does rancher react with the conditions on the provisioning cluster being missing?

jakefhyde: This is the infrastructure cluster, not the provisioning cluster, so neither rancher nor CAPI utilize these conditions.

Oats87 (approved):

HarrisonWAffel (approved):
looks good to me, had a huddle with Jake regarding these changes (thanks!)

PR #39901, opened 2022-12-14
Bump steve version for counts-diff changes

MbolotSuse:
Related to rancher/rancher#36681. Since the code for this change was in Steve, Steve's version needs to be bumped to pull the changes in.


rancherbot (approved):

Issue #39734, opened 2022-11-29
[Feature] Adjust Rancher helm chart to allow disabling of PSPs


cbron:
Epic: https://github.com/rancher/rancher/issues/38701
SURE-5624

Need to adjust the rancher chart to handle this PSP existing, which right now you can't disable: https://github.com/rancher/rancher/blob/release/v2.7/chart/templates/post-delete-hook-psp.yaml
There might be other places as well. 
Note that this has nothing to do with feature charts, only the rancher chart itself.

maxsokolovsky:
## QA Testing
Ensure the following configurations produce no errors on downstream cluster provisioning or any unusual error messages in the local cluster's logs. A downstream v1.24 cluster always has PSP's enabled:

local 1.25 downstream 1.25
local 1.25 downstream 1.24
local 1.24 downstream 1.25
local 1.24 downstream 1.24

sowmyav27:
This isn't ready for test yet. this will need a Rancher RC to test

jiaqiluo:
Can be validated in Rancher [v2.7.2-rc1](https://github.com/rancher/rancher/releases/tag/v2.7.2-rc1)

anupama2501:
QA to validate:
1. Install rancher on RKE1 k8s 1.24 hardened cluster - have a hardened PSP 1.24 RKE2 custom cluster - c1 and a non hardened 1.25 k3s node driver cluster - c2 , psp 1.23 rke1 node driver non hardened - c3
2. Uninstall rancher on RKE1 k8s 1.24 local cluster from step1 by running the [cleanup](https://ranchermanager.docs.rancher.com/faq/rancher-is-no-longer-needed) script 
3. Install rancher on k8s 1.25 RKE2 cluster non hardened cluster - Have a hardened PSP 1.24 k3s custom cluster - c1, non hardened 1.25 RKE1 node driver- c2
4. Uninstall rancher on RKE2 local cluster from step3 by running the [cleanup]

sowmyav27:
Wait until this is fixed - https://github.com/rancher/rancher/issues/40327 since this issue talks about reverting the above linked PR

cbron:
Superceded by: https://github.com/rancher/rancher/issues/40327

PR #39920, opened 2022-12-15
[chart] Do not apply manifests or blocks if they reference PSP's in K8s v1.25

maxsokolovsky:
## Issue: https://github.com/rancher/rancher/issues/39734
Part of the epic https://github.com/rancher/rancher/issues/38701
 
## Problem
1. Currently, the chart deploys one PSP resource, which Kubernetes v1.25 will not support.
2. There is also a cluster role created that has a rule that references PSP's - not needed anymore for v1.25.
 
## Solution
For the two cases above, I add a condition that checks the underlying cluster's version. If it's v1.25 or newer, do not render the whole resource (or a relevant YAML block). 
 
### Manual Testing
I tried deploying the chart without these changes onto a v1.25 cluster, and the release deployed, although there were errors on deletion of the release.

### DIFF
--- a/chart/templates/post-delete-hook-cluster-role.yaml
+++ b/chart/templates/post-delete-hook-cluster-role.yaml
@@ -30,9 +30,11 @@ rules:
   - apiGroups: [ "admissionregistration.k8s.io" ]
     resources: [ "validatingwebhookconfigurations", "mutatingwebhookconfigurations" ]
     verbs: [ "get", "list", "delete" ]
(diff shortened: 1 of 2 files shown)


cbron (approved):
Lgtm. I was thinking there was another mechanism tying the PSP and the Service Account together, but I suppose the role being able to use PSPs was all it needed.

crobby (approved):
seems to work, lgtm

PR #39936, opened 2022-12-19
Add validation test for etcd only backup and restore

vivek-infracloud:
## Issue: 
https://github.com/rancher/qa-tasks/issues/405

**Note:** 

This validation test takes around 16 mins to complete which is more than default timeout for the test that's why we need to use -timeout option with value either greater than 16 mins or 0. Similar to below command

`go test providers.go etcd_backup_restore.go etcd_only_backup_restore_test.go -v -timeout 0`

Also it will also requires one k8s version and CNI provider in the cattle config file as per below:
```
provisioningInput:  
   kubernetesVersion: 
    - v1.24.8+rke2r1
  cni:
    - calico
```

### DIFF
--- a/tests/v2/validation/provisioning/rke2/etcd_backup_restore.go
+++ b/tests/v2/validation/provisioning/rke2/etcd_backup_restore.go
@@ -2,26 +2,31 @@ package rke2
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	provisioningV1 "github.com/rancher/rancher/tests/framework/clients/rancher/v1"
 	v1 "github.com/rancher/rancher/tests/framework/clients/rancher/v1"
 	"github.com/rancher/rancher/tests/framework/extensions/clusters"
 	"github.com/rancher/rancher/tests/framework/extensions/machinepools"
 	"github.com/rancher/rancher/tests/framework/extensions/workloads"
+	"github.com/rancher/rancher/tests/framework/extensions/workloads/pods"
 	"github.com/sirupsen/logrus"
 	appv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
+	defaultNamespace             = "default"
 	localClusterName             = "local"
 	wloadBeforeRestore           = "wload-before-restore"
 	ingressName                  = "ingress"
 	wloadServiceName             = "wload-service"
-	wloadAfterRestore            = "wload-after-restore"
+	wloadAfterBackup             = "wload-after-backup"
 	ProvisioningSteveResouceType = "provisioning.cattle.io.cluster"
 )
 
@@ -185,4 +190,44 @@ func createDeployment(deployment *appv1.Deployment, steveclient *v1.Client, clie
 	})
 	return deploymentResp2, err
 
+}
+
+func watchAndWaitForPods(client *rancher.Client, clusterID string) error {
+	logrus.Infof("waiting for all Pods to be up.............")
+	err := kwait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+		steveClient, err := client.Steve.ProxyDownstream(clusterID)
+		if err != nil {
+			return false, nil
+		}
+		pods, err := steveClient.SteveType(pods.PodResourceSteveType).List(nil)
+		if err != nil {
+			return false, nil
+		}
+		isIngressControllerPodPresent := false
+		isKubeControllerManagerPresent := false
+		for _, pod := range pods.Data {
+			podStatus := &corev1.PodStatus{}
+			err = provisioningV1.ConvertToK8sType(pod.Status, podStatus)
+			if err != nil {
+				return false, err
+			}
+			if !isIngressControllerPodPresent && strings.Contains(pod.ObjectMeta.Name, "ingress-nginx-controller") {
+				isIngressControllerPodPresent = true
+			}
+			if !isKubeControllerManagerPresent && strings.Contains(pod.ObjectMeta.Name, "kube-controller-manager") {
+				isKubeControllerManagerPresent = true
+			}
+
+			phase := podStatus.Phase
+			if phase != corev1.PodRunning && phase != corev1.PodSucceeded {
+				return false, nil
+			}
+
+		}
+		if isIngressControllerPodPresent && isKubeControllerManagerPresent {
+			return true, nil
+		}
+		return false, nil
+	})
+	return err
 }
--- a/tests/framework/extensions/clusters/clusters.go
+++ b/tests/framework/extensions/clusters/clusters.go
@@ -12,6 +12,7 @@ import (
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+	kubeProvisioning "github.com/rancher/rancher/tests/framework/clients/provisioning"
 	"github.com/rancher/rancher/tests/framework/clients/rancher"
 	management "github.com/rancher/rancher/tests/framework/clients/rancher/generated/management/v3"
 	v1 "github.com/rancher/rancher/tests/framework/clients/rancher/v1"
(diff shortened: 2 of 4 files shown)


tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,305 @@
...
+	client, err := rancher.NewClient("", testSession)
+	require.NoError(r.T(), err)
+
+	r.client = client
+
+}
+
+func (r *RKE2EtcdOnlySnapshotRestore) EtcdSnapshotRestore(provider *Provider) {
+	initialK8sVersion := r.kubernetesVersions[0]
+	logrus.Infof("running etcd snapshot restore test.............")
```
markusewalker, line 74: nit, but is this log really needed since it's just declaring the test is starting?

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,305 @@
...
+	clusterResp, err := createRKE2NodeDriverCluster(client, provider, clusterName, initialK8sVersion, r.ns, r.cnis[0])
+	require.NoError(r.T(), err)
+	require.Equal(r.T(), clusterName, clusterResp.ObjectMeta.Name)
+	logrus.Infof("rke2Cluster create request successful.............")
+
+	logrus.Infof("creating watch over cluster.............")
+	r.watchAndWaitForCluster(kubeProvisioningClient, clusterName)
+	logrus.Infof("cluster is up and running.............")
+
+	// Get clusterID by clusterName
```
markusewalker, line 98: I think we can do without these comments since you already are doing a `logrus` indicating the action that is taking place anyway. Seems a bit redundant.

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,305 @@
...
+	deploymentResp, err := createDeployment(deploymentBeforeBackup, steveclient, client, clusterID)
+	require.NoError(r.T(), err)
+	require.Equal(r.T(), deploymentBeforeBackup.Name, deploymentResp.ObjectMeta.Name)
+	logrus.Infof("%v is ready.............", deploymentBeforeBackup.Name)
+
+	// creating the ingress1
+	logrus.Infof("creating an ingress.............")
+
+	exactPath := networkingv1.PathTypeExact
+	paths := []networkingv1.HTTPIngressPath{
```
markusewalker, line 136: Can be this abstracted into a function defined in `etcd_backup_restore.go`? Maybe something called `CreateIngressPath`, just to hide the implementation in the actual test.

markusewalker:
@vivek-infracloud the changes look good to me. Offline, can you send us an updated Jenkins job to validate everything is working with the now abstracted function? Thank you!!

markusewalker (approved):
LGTM

vivek-infracloud:
Yes, it would require a backport. Since this code is good for 2.7 I will create backport PR.

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+	client, err := r.client.WithSession(subSession)
+	require.NoError(r.T(), err)
+
+	logrus.Infof("creating kube provisioning client.............")
+	kubeProvisioningClient, err := r.client.GetKubeAPIProvisioningClient()
+	require.NoError(r.T(), err)
+	logrus.Infof("kube provisioning client created.............")
+
+	clusterName := provisioning.AppendRandomString(provider.Name)
```
markusewalker, line 85: I noticed this when reviewing the backport. This should actually be using `namegen` as the backport is using. This is due to Corral changes that recently were introduced.

igomez06: Yes Vivek please rebase your branch, we have had a few incidents with breaking tests because we neglected to rebase our branches for the most recent code.

vivek-infracloud: Updated as per comment. Changes are in the recent commit.

markusewalker (dismissed):
LGTM

igomez06:
One last rebase Vivek, there was a recent merge that affects your branch. Thank you!

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,286 @@
...
+		if totalClusterSnapShots == etcdnodeCount {
+			return true, nil
+		}
+		return false, nil
+	})
+	require.NoError(r.T(), err)
+
+	logrus.Infof("creating a workload(w2, deployment).............")
+	wloadAfterBackupLabels := map[string]string{}
+	wloadAfterBackupLabels["workload.user.cattle.io/workloadselector"] = fmt.Sprintf("apps.deployment-%v-%v", r.ns, wloadAfterRestore)
```
slickwarren, line 176: not sure where the wloadAfterRestore came from ? We haven't restored anything yet, only backed up.

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,286 @@
...
+func (r *RKE2EtcdOnlySnapshotRestore) watchAndWaitForCluster(kubeProvisioningClient *kubeProvisioning.Client, clusterName string) {
+	err := kwait.Poll(5*time.Second, 2*time.Minute, func() (done bool, err error) {
+		clusterResp, err := r.client.Steve.SteveType(ProvisioningSteveResouceType).ByID(r.ns + "/" + clusterName)
+		if err != nil {
+			return false, err
+		}
+		state := clusterResp.ObjectMeta.State.Name
+		return state != "active", nil
+	})
+	require.NoError(r.T(), err)
```
caliskanugur, line 222: I don't understand this extra wait/watch for the active state, the wait/watch below isn't sufficient to wait the cluster to be up?

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,286 @@
...
+			}
+
+		}
+		if isIngressControllerPodPresent && isKubeControllerManagerPresent {
+			return true, nil
+		}
+		return false, nil
+	})
+	require.NoError(r.T(), err)
+}
```
caliskanugur, line 273: In my opinion, this is not a private method of the test suite but more of a private helper that could be in `etcd_restore.go`, seems like it doesn't have any actual test, the only thing that needs to be updated is, returning the error. What do you all think?

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
```diff
@@ -0,0 +1,286 @@
...
+func TestEtcdOnlySnapshotRestore(t *testing.T) {
+	suite.Run(t, new(RKE2EtcdOnlySnapshotRestore))
+}
+
+func (r *RKE2EtcdOnlySnapshotRestore) TestEtcdOnlySnapshotRestore() {
+	logrus.Infof("checking for valid k8s versions and cnis in the configuration....")
+	require.GreaterOrEqual(r.T(), len(r.kubernetesVersions), 1)
+	require.GreaterOrEqual(r.T(), len(r.cnis), 1)
```
caliskanugur, line 281: Another nit, I think using `GreaterOrEqualf` with an error message might be more straightforward to pick up while running the job with wrong values.

caliskanugur (commented):
A few nits and questions, otherwise looks good

tests/v2/validation/provisioning/rke2/etcd_backup_restore.go
```diff
@@ -124,6 +126,21 @@ func createRKE2NodeDriverCluster(client *rancher.Client, provider *Provider, clu
 
 }
 
+func CreateIngressPath(pathType networkingv1.PathType, path string, serviceName string, servicePort int32) networkingv1.HTTPIngressPath {
```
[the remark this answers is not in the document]
igomez06: I agree I think this is a candidate for an extension but calling it` NewIngressPathTemplate()`

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
igomez06: Yeah where is this variable coming from? It doesn't look like it's initialized anywhere, are you get build errors?

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
igomez06: This is after a restore, or other updates to the cluster, he's making string the cluster is in an active state. Which I think can be an extension, since I can see this being used outside of just etcd back up restore testing.

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
igomez06: agreed let's just have strictly validations in the test file, and other logic in the helper file. Does that make sense?

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
igomez06: agreed

tests/v2/validation/provisioning/rke2/etcd_only_backup_restore_test.go
caliskanugur: Ah I see, agreed, definitely a good candidate. In terms of being extensions, ideally, if it could break the poll once it's in the error state that might be even more useful, both for monitoring on the Jenkins side and reducing to waiting the default time out.

markusewalker (approved):

igomez06 (approved):
LGTM

slickwarren (approved):
LGTM

PR #39941, opened 2022-12-19
Do not register podsecuritypolicy controllers for clusters >= v1.25

maxsokolovsky:
## Problem
Rancher may attempt to register various controllers (https://github.com/rancher/rancher/tree/release/v2.7/pkg/controllers/managementuser/rbac/podsecuritypolicy) for a downstream cluster that work with resources (namespaces, cluster roles, and others) for PSP management. If the downstream cluster's version is v1.25 or later, Rancher will have constant controller errors when doing any operation pertaining to PSP resources, since the K8s API server will throw an error when it is asked to fetch an unknown resource. And PSP's are not resources known to K8s API as of v1.25.

There are two conditions which currently allow controllers to be registered:
1. If the downstream cluster has a project that has PSP's configured. Specifically, if there exists a PSPProjectTemplateBinding that has `cluster:project` as its target.
2. If the downstream cluster spec object has some value for `DefaultPodSecurityPolicyTemplateName`.
 
## Solution
1. Do not register the PSP-related controllers, if the downstream cluster version is v1.25 or later.
2. Exit from the controller's sync methods early, if the controllers still run. This happens if a 1.24 cluster with PSPs enabled got upgraded to 1.25 (with PSPs disabled). The controllers would still run, they do not get de-registered. We must not let them run the main logic and exit early. However, if the Rancher pods are restarted, then the controllers are gone, and need to be re-registered. But at that point, the registration function detects a v1.25 cluster and refuses to register the controllers.

I do log the K8s version incompatibly error for a failed controller registration attempt. But I decided not to log the incompatibility error for controller sync attempts, because:
1. This will flood the logs unnecessarily.
2. The sync methods would not run at all once Rancher pods are restarted, because the controllers would not even be registered then.
 
### Manual Testing
Ensure the following configurations produce no errors on downstream cluster provisioning. A downstream v1.24 cluster always has PSP's enabled:

local 1.25 downstream 1.25
local 1.25 downstream 1.24
local 1.24 downstream 1.25
local 1.24 downstream 1.24

#### Other notes
On local 1.25 (did not try this on 1.2.4), I have the following observations.
When I create a 1.24 with PSPs, the cluster compatibility check runs and succeeds.
When I upgrade the cluster to 1.25, the cluster compatibility check does not run.

When I create a 1.24 with PSPs and upgrade it to 1.25 without them, and restart Rancher pods,
the cluster compatibility check runs and fails before controller registration.
It does not run for a new 1.25, nor when I restart Rancher pods.

### DIFF
--- a/pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
+++ b/pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
@@ -2,10 +2,9 @@ package podsecuritypolicy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
-
-	"k8s.io/apimachinery/pkg/runtime"
 
 	v1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
@@ -16,8 +15,9 @@ import (
 	v13 "k8s.io/api/core/v1"
 	v1beta13 "k8s.io/api/policy/v1beta1"
 	rbac "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 )
 
@@ -38,6 +38,7 @@ func RegisterBindings(ctx context.Context, context *config.UserContext) {
 		policyLister:         context.Policy.PodSecurityPolicies("").Controller().Lister(),
 		policies:             context.Policy.PodSecurityPolicies(""),
 		psptLister:           context.Management.Management.PodSecurityPolicyTemplates("").Controller().Lister(),
+		clusterLister:        context.Management.Management.Clusters("").Controller().Lister(),
 		clusterRoleLister:    context.RBAC.ClusterRoles("").Controller().Lister(),
 		clusterRoles:         context.RBAC.ClusterRoles(""),
 		serviceAccountLister: context.Core.ServiceAccounts("").Controller().Lister(),
@@ -45,6 +46,7 @@ func RegisterBindings(ctx context.Context, context *config.UserContext) {
 
 		namespaces:       context.Core.Namespaces(""),
 		namespaceIndexer: namespaceInformer.GetIndexer(),
+		clusterName:      context.ClusterName,
 	}
 
 	context.Management.Management.PodSecurityPolicyTemplateProjectBindings("").
@@ -71,6 +73,8 @@ type lifecycle struct {
 
 	namespaces       v1.NamespaceInterface
 	namespaceIndexer cache.Indexer
+	clusterName      string
+	clusterLister    v3.ClusterLister
 }
 
 func (l *lifecycle) Create(obj *v3.PodSecurityPolicyTemplateProjectBinding) (runtime.Object, error) {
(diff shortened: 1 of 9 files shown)


pkg/controllers/managementuser/rbac/podsecuritypolicy/version.go
```diff
@@ -0,0 +1,31 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var errVersionIncompatible = errors.New("podsecuritypolicies are not available in Kubernetes v1.25 and above")
+var genericErrorString = "failed to check cluster version compatibility for podsecuritypolicy controllers: %v"
```
crobby, line 13: the name "generic" isn't too telling, maybe something more related to the contents of the error string?

pkg/controllers/managementuser/rbac/podsecuritypolicy/version.go
```diff
@@ -0,0 +1,31 @@
...
+
+func checkVersion(clusterName string, clusterLister v3.ClusterInterface) error {
+	cluster, err := clusterLister.Get(clusterName, v1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get cluster [%s]: %w", clusterName, err)
+	}
+	if cluster.Status.Version == nil {
+		return fmt.Errorf("cannot validate Kubernetes version for podsecuritypolicy capability: cluster [%s] status version is not available yet", clusterName)
+	}
+	v, err := strconv.Atoi(cluster.Status.Version.Minor)
```
crobby, line 23: Might not be critical for now, but do we need to future-proof for v 2.X and limit this check to require v1.X?

pkg/controllers/managementuser/rbac/podsecuritypolicy/version.go
```diff
@@ -0,0 +1,31 @@
...
+	"strconv"
+
+	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var errVersionIncompatible = errors.New("podsecuritypolicies are not available in Kubernetes v1.25 and above")
+var genericErrorString = "failed to check cluster version compatibility for podsecuritypolicy controllers: %v"
+
+func checkVersion(clusterName string, clusterLister v3.ClusterInterface) error {
```
crobby, line 15: maybe a comment for this utility function?

crobby (commented):
Seems pretty sane, maybe a test or two could be added as well.

pkg/controllers/managementuser/rbac/podsecuritypolicy/version.go
maxsokolovsky: Good point, I should probably follow other examples of where we check semantic versions, using the `semver` package...

pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
```diff
@@ -90,6 +95,11 @@ func (l *lifecycle) sync(obj *v3.PodSecurityPolicyTemplateProjectBinding) (runti
 		return obj, nil
 	}
 
+	if err := checkClusterVersion(l.clusterName, l.clusters); err != nil && !goerrors.Is(err, errVersionIncompatible) {
```
cmurphy, line 98: If the version check in deferred.go prevents these handlers from being registered, are these in-handler checks necessary?

pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
```diff
@@ -38,13 +39,15 @@ func RegisterBindings(ctx context.Context, context *config.UserContext) {
 		policyLister:         context.Policy.PodSecurityPolicies("").Controller().Lister(),
 		policies:             context.Policy.PodSecurityPolicies(""),
 		psptLister:           context.Management.Management.PodSecurityPolicyTemplates("").Controller().Lister(),
+		clusters:             context.Management.Management.Clusters(""),
```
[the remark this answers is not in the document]
maxsokolovsky: Should I use a cluster lister in other modified controllers in this package?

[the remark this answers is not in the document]
cmurphy: Anywhere that is only doing a read operation on the cluster and not modifying the cluster can use a lister

pkg/controllers/managementuser/rbac/podsecuritypolicy/binding.go
maxsokolovsky: Yes, because registration runs only once - when you provision a remote cluster with PSPs enabled, or when you enable them afterwards in cluster configuration.
That means you could have a remote v1.24 cluster with PSP enabled (so the controllers will be registered and started). When you upgrade the remote cluster to v1.25 (with PSPs disabled), the controllers will not de-register themselves or stop. We don't want them to run, since PSPs are irrelevant on v1.25. Exiting from `sync` early is, unfortunately, the only way, unless there is a clean way to de-register controllers.

cmurphy (approved):

crobby (approved):

Issue #38282, opened 2022-07-14
Code Coverage Percentage for Rancher


igomez06:
Code coverage percentage from units and also external tests. Will need to a build of rancher with code coverage enabled, and run our automation tests against to figure out the code coverage percentage.

PR #39944, opened 2022-12-20
[v2.6] End to End test Code Coverage with Jenkins pipeline and setup

igomez06:
This PR includes code coverage custom binaries for both rancher and agent. As well as, Dockerfiles for both rancher and rancher-agent. Setup scripts for creating the binaries and the docker images. Scripts for setting the Jenkins environment to have a test coverage run for all our tests. New extensions that kill the containers that are running the rancher/agent images to have the coverage reports written, and code to retreive those reports from their pods.

## Issue: 
 https://github.com/rancher/rancher/issues/38282, https://github.com/rancher/qa-tasks/issues/520
 
## Problem
 Need to have a way to measure our code coverage of rancher/rancher from out end to end tests, including all of python tests (validation and integration), and our go automation tests.
 
## Solution

**Overview**
1. Custom test binaries for rancher and rancher-agent. This is so we can encapsulate the rancher/main.go and rancher/cmd/agent/main.go in `TestMain` functions to be able run the binaries as tests to be able to use the `go test -cover` functionality 
2. Custom Dockerfiles for both rancher and rancher-agent to be run in deployments of the local cluster and the downstream clusters provisioned by rancher. 
3. New extensions that work as kubectl exec calls to able to "kill" the container in the its respective pod that is running the custom image (rancher or rancher-agent) and also calls to be able to retrieve said coverage reports from the pods to a local destination.
4. New Jenkins pipeline that runs all of our end to end tests, retrieves the code coverage reports, and publishes the reports.
5. Set up code the sets up the environment for our aforementioned Jenkins pipeline, that uses corral to create a rancher instance, update the rancher instance to use the custom code coverage images, and provision clusters to be used in the end to end tests.
 
## Testing
 code coverage report Jenkins job

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -220,6 +220,7 @@ require (
 	github.com/fatih/color v1.13.0 // indirect
 	github.com/felixge/httpsnoop v1.0.1 // indirect
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
+	github.com/fvbommel/sortorder v1.0.1 // indirect
 	github.com/go-asn1-ber/asn1-ber v1.5.3 // indirect
 	github.com/go-errors/errors v1.0.1 // indirect
 	github.com/go-gorp/gorp/v3 v3.0.2 // indirect
--- a/tests/framework/clients/corral/corral.go
+++ b/tests/framework/clients/corral/corral.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os/exec"
 	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/rancher/tests/framework/pkg/session"
@@ -17,6 +18,18 @@ const (
 	debugFlag       = "--trace"
 	skipCleanupFlag = "--skip-cleanup"
 )
+
+// GetCorralEnvVar gets corral environment variables
+func GetCorralEnvVar(corralName, envVar string) (string, error) {
+	msg, err := exec.Command("corral", "vars", corralName, envVar).CombinedOutput()
+	if err != nil {
+		return "", errors.Wrap(err, "GetCorralEnvVar: "+string(msg))
+	}
+
+	corralEnvVar := string(msg)
+	corralEnvVar = strings.TrimSuffix(corralEnvVar, "\n")
+	return corralEnvVar, nil
+}
 
 // SetupCorralConfig sets the corral config vars. It takes a map[string]string as a parameter; the key is the value and the value the value you are setting
 // For example we are getting the aws config vars to build a corral from aws.
--- a/tests/framework/extensions/codecoverage/codecoverage.go
+++ b/tests/framework/extensions/codecoverage/codecoverage.go
@@ -1,0 +1,185 @@
+package codecoverage
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	apiv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	"github.com/rancher/rancher/tests/framework/clients/rancher"
+	v1 "github.com/rancher/rancher/tests/framework/clients/rancher/v1"
+	"github.com/rancher/rancher/tests/framework/extensions/clusters"
+	"github.com/rancher/rancher/tests/framework/extensions/kubeconfig"
+	"github.com/rancher/rancher/tests/framework/pkg/killserver"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+)
+
+var podGroupVersionResource = corev1.SchemeGroupVersion.WithResource("pods")
+
+const (
+	cattleSystemNameSpace = "cattle-system"
+	localCluster          = "local"
+	rancherCoverFile      = "ranchercoverage"
+	agentCoverFile        = "agentcoverage"
+	outputDir             = "cover"
+)
+
+func checkServiceIsRunning(dynamicClient dynamic.Interface) error {
+	return kwait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
+		_, err = dynamicClient.Resource(podGroupVersionResource).Namespace(cattleSystemNameSpace).List(context.Background(), metav1.ListOptions{})
+		if k8sErrors.IsInternalError(err) || k8sErrors.IsServiceUnavailable(err) {
+			return false, nil
+		} else if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
+}
+
+func killTestServices(client *rancher.Client, clusterID string, podNames []string) error {
+	cmd := []string{
+		"/bin/sh",
+		"-c",
+		fmt.Sprintf("curl -s localhost%s", killserver.KillServerPort),
+	}
+
+	kubeConfig, err := kubeconfig.GetKubeconfig(client, clusterID)
+	if err != nil {
+		return err
+	}
+
+	restConfig, err := (*kubeConfig).ClientConfig()
+	if err != nil {
+		return err
+	}
+
+	for _, podName := range podNames {
+		_, err := kubeconfig.KubectlExec(restConfig, podName, cattleSystemNameSpace, cmd)
+		if err != nil {
+			logrus.Error("error killing pod container %v", err)
+		}
+	}
+
+	return nil
+}
+
+func retrieveCodeCoverageFile(client *rancher.Client, clusterID, coverageFilename string, podNames []string) error {
+	kubeConfig, err := kubeconfig.GetKubeconfig(client, clusterID)
+	if err != nil {
+		return err
+	}
+
+	restConfig, err := (*kubeConfig).ClientConfig()
+	if err != nil {
+		return err
+	}
+
+	for _, podName := range podNames {
+		fileName := fmt.Sprintf("%s%s", podName, coverageFilename)
+		dst := fmt.Sprintf("%s/%s", outputDir, fileName)
+
+		err := kubeconfig.CopyFileFromPod(restConfig, podName, cattleSystemNameSpace, coverageFilename, dst)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// KillRancherTestServicesRetrieveCoverage is a function that kills the rancher service of the local cluster
+// inorder for the code coverage report to be written, and then copies over the coverage reports from the pods
+// to a local destination. The custom code coverage rancher image must be running in the local cluster.
+func KillRancherTestServicesRetrieveCoverage(client *rancher.Client) error {
+	var podNames []string
+	dynamicClient, err := client.GetRancherDynamicClient()
+	if err != nil {
+		return err
+	}
+
+	pods, err := dynamicClient.Resource(podGroupVersionResource).Namespace(cattleSystemNameSpace).List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, pod := range pods.Items {
+		name := pod.GetName()
+		if strings.Contains(name, "rancher") && !strings.Contains(name, "webhook") {
+			podNames = append(podNames, pod.GetName())
+		}
+	}
+
+	err = killTestServices(client, localCluster, podNames)
+	if err != nil {
+		return err
+	}
+
+	err = checkServiceIsRunning(dynamicClient)
+	if err != nil {
+		return err
+	}
+
+	return retrieveCodeCoverageFile(client, localCluster, rancherCoverFile, podNames)
+}
+
+// KillAgentTestServicesRetrieveCoverage is a function that kills the cattle-cluster-agent service of a downstream cluster
+// inorder for the code coverage report to be written, and then copies over the coverage reports from the pods
+// to a local destination. The custom code coverage rancher-agent image must be running in the downstream cluster.
+func KillAgentTestServicesRetrieveCoverage(client *rancher.Client) error {
+	clusters, err := client.Steve.SteveType(clusters.ProvisioningSteveResouceType).ListAll(nil)
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range clusters.Data {
+		clusterStatus := &apiv1.ClusterStatus{}
+		err = v1.ConvertToK8sType(cluster.Status, clusterStatus)
+		if err != nil {
+			return err
+		}
+		clusterID := clusterStatus.ClusterName
+		if clusterID != localCluster {
+			dynamicClient, err := client.GetDownStreamClusterClient(clusterID)
+			if err != nil {
+				logrus.Errorf("could not connect to downstream cluster")
+				continue
+			}
+
+			pods, err := dynamicClient.Resource(podGroupVersionResource).Namespace(cattleSystemNameSpace).List(context.Background(), metav1.ListOptions{})
+			if err != nil {
+				logrus.Errorf("could not list pods")
+				continue
+			}
+
+			var podNames []string
+			for _, pod := range pods.Items {
+				if strings.Contains(pod.GetName(), "cattle-cluster-agent") {
+					podNames = append(podNames, pod.GetName())
+				}
+			}
+
+			err = killTestServices(client, clusterID, podNames)
+			if err != nil {
+				return err
+			}
+
+			err = checkServiceIsRunning(dynamicClient)
+			if err != nil {
+				return err
+			}
+
+			err = retrieveCodeCoverageFile(client, clusterID, agentCoverFile, podNames)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
--- a/tests/framework/pkg/config/config.go
+++ b/tests/framework/pkg/config/config.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"io/ioutil"
 	"os"
 
 	"github.com/creasty/defaults"
@@ -18,7 +17,7 @@ func LoadConfig(key string, config interface{}) {
 		return
 	}
 
-	allString, err := ioutil.ReadFile(configPath)
+	allString, err := os.ReadFile(configPath)
 	if err != nil {
 		panic(err)
 	}
@@ -44,4 +43,41 @@ func LoadConfig(key string, config interface{}) {
 		panic(err)
 	}
 
+}
+
+// UpdateConfig is function that updates the CATTLE_TEST_CONFIG yaml/json that the framework uses.
+func UpdateConfig(key string, config interface{}) {
+	configPath := os.Getenv("CATTLE_TEST_CONFIG")
+
+	if configPath == "" {
+		yaml.Unmarshal([]byte("{}"), config)
+		return
+	}
+
+	// Read json buffer from jsonFile
+	byteValue, err := os.ReadFile(configPath)
+	if err != nil {
+		panic(err)
+	}
+
+	// We have known the outer json object is a map, so we define result as map.
+	// otherwise, result could be defined as slice if outer is an array
+	var result map[string]interface{}
+	err = yaml.Unmarshal(byteValue, &result)
+	if err != nil {
+		panic(err)
+	}
+
+	result[key] = config
+
+	yamlConfig, err := yaml.Marshal(result)
+
+	if err != nil {
+		panic(err)
+	}
+
+	err = os.WriteFile(configPath, yamlConfig, 0644)
+	if err != nil {
+		panic(err)
+	}
 }
--- a/tests/framework/pkg/killserver/killserver.go
+++ b/tests/framework/pkg/killserver/killserver.go
@@ -1,0 +1,47 @@
+package killserver
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	KillServerPort = ":19999"
+)
+
+// KillerServer is struct used to cancel a context of web service, that listens on a specific port.
+type KillServer struct {
+	Server http.Server
+	cancel context.CancelFunc
+}
+
+// NewKillServer initializes a KillServer at a specfic address/port and the cancel context of said web service.
+func NewKillServer(addr string, cancel context.CancelFunc) *KillServer {
+	return &KillServer{
+		Server: http.Server{
+			Addr: addr,
+		},
+		cancel: cancel,
+	}
+}
+
+// Start starts the ListenAndServe of the KillServer server
+func (s *KillServer) Start() {
+	s.Server.Handler = s
+
+	err := s.Server.ListenAndServe()
+	if err != nil {
+		logrus.Errorf("KillServer error: %v", err)
+	}
+}
+
+// ServeHTTP should write reply headers and data to the ResponseWriter
+// and then return.
+func (s *KillServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusOK)
+
+	// cancel the context
+	s.cancel()
+}
--- a/tests/framework/pkg/session/session.go
+++ b/tests/framework/pkg/session/session.go
@@ -1,7 +1,7 @@
 package session
 
 import (
-	"testing"
+	"github.com/sirupsen/logrus"
 )
 
 // CleanupFunc is the type RegisterCleanupFunc accepts
@@ -12,16 +12,14 @@ type Session struct {
 	CleanupEnabled bool
 	cleanupQueue   []CleanupFunc
 	open           bool
-	TestingT       *testing.T
 }
 
 // NewSession is a constructor instantiates a new `Session`
-func NewSession(t *testing.T) *Session {
+func NewSession() *Session {
 	return &Session{
 		CleanupEnabled: true,
 		cleanupQueue:   []CleanupFunc{},
 		open:           true,
-		TestingT:       t,
 	}
 }
 
@@ -44,7 +42,7 @@ func (ts *Session) Cleanup() {
 		for i := len(ts.cleanupQueue) - 1; i >= 0; i-- {
 			err := ts.cleanupQueue[i]()
 			if err != nil {
-				ts.TestingT.Logf("error calling cleanup function: %v", err)
+				logrus.Errorf("error calling cleanup function: %v", err)
 			}
 		}
 		ts.cleanupQueue = []CleanupFunc{}
@@ -53,7 +51,7 @@ func (ts *Session) Cleanup() {
 
 // NewSession returns a `Session` who's cleanup method is registered with this `Session`
 func (ts *Session) NewSession() *Session {
-	sess := NewSession(ts.TestingT)
+	sess := NewSession()
 
 	ts.RegisterCleanupFunc(func() error {
 		sess.Cleanup()
--- a/tests/integration/suite/conftest.py
+++ b/tests/integration/suite/conftest.py
@@ -39,7 +39,8 @@ def get_ip():
 
 
 IP = get_ip()
-SERVER_URL = 'https://' + IP + ':8443'
+SERVER_URL = os.environ.get('CATTLE_TEST_URL', 'https://' + IP + ':8443')
+SERVER_PASSWORD = os.environ.get('RANCHER_SERVER_PASSWORD', 'admin')
 BASE_URL = SERVER_URL + '/v3'
 AUTH_URL = BASE_URL + '-public/localproviders/local?action=login'
 DEFAULT_TIMEOUT = 120
@@ -101,7 +102,7 @@ def admin_mc():
     """Returns a ManagementContext for the default global admin user."""
     r = requests.post(AUTH_URL, json={
         'username': 'admin',
-        'password': 'admin',
+        'password': SERVER_PASSWORD,
         'responseType': 'json',
     }, verify=False)
     protect_response(r)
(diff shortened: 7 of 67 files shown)


tests/v2/codecoverage/setuprancher/main.go
```diff
@@ -0,0 +1,505 @@
...
+		return false, nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	deploymentResource = dynamicClient.Resource(deployments.DeploymentGroupVersionResource).Namespace("cattle-fleet-local-system")
+	err = kwait.Poll(500*time.Millisecond, 10*time.Minute, func() (done bool, err error) {
+		webhookDeployment, err := deploymentResource.Get(context.TODO(), "fleet-agent", metav1.GetOptions{})
```
caliskanugur, line 301: Really small nit, maybe changing these names for the deployments watch might increase readability a bit, maybe calling `deployment` inside the polls? Applicable for below polls as well

tests/framework/pkg/killserver/killserver.go
```diff
@@ -0,0 +1,43 @@
+package killserver
```
caliskanugur, line 1: So I don't have a strong opinion on this, but does this need to be in `pkg` rather than `extensions/codecoverage/killserver`, what are your thoughts on that, do you think this is something generic to be in `pkg` dir?

caliskanugur (commented):
Already looks really good to me, and almost all comments are nits and questions, wanted to point these out before the 2.7 branch. This is my initial review, I'll definitely give it a second-round review

slickwarren (commented):
mostly nits, but a few questions as well. Considering the amount of code changes and additions, very well done!

tests/framework/pkg/session/session.go
```diff
@@ -16,12 +18,11 @@ type Session struct {
 }
 
 // NewSession is a constructor instantiates a new `Session`
-func NewSession(t *testing.T) *Session {
+func NewSession() *Session {
```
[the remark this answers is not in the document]
igomez06: It shouldn't be there I missed that will fix shortly.

tests/v2/validation/provisioning/rke2/custom_cluster_test.go
```diff
@@ -268,6 +268,7 @@ func (c *CustomClusterProvisioningTestSuite) ProvisioningRKE2CustomClusterDynami
 }
 
 func (c *CustomClusterProvisioningTestSuite) TestProvisioningCustomCluster() {
+	c.T().Skip()
```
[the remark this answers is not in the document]
igomez06: This was a debug testing that I had forgotten to remove, I will remove it.

tests/framework/extensions/codecoverage/codecoverage.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		return true, nil
+	})
+}
+
+func killTestServices(client *rancher.Client, clusterID string, podNames []string) error {
+	cmd := []string{
+		"/bin/sh",
+		"-c",
+		"curl -s localhost:19999",
```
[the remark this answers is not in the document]
igomez06: Good point I can change that.

slickwarren:
looking at the jenkins run, I have some further questions:

tests/framework/extensions/codecoverage/codecoverage.go
```diff
@@ -0,0 +1,181 @@
...
+	rancherCoverFile      = "ranchercover.out"
+	agentCoverFile        = "agentcover.out"
+	outputDir             = "cover"
+)
+
+func checkServiceIsRunning(dynamicClient dynamic.Interface) error {
+	return kwait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
+		_, err = dynamicClient.Resource(corev1.SchemeGroupVersion.WithResource("pods")).Namespace(cattleSystemNameSpace).List(context.Background(), metav1.ListOptions{})
+		if k8sErrors.IsInternalError(err) || k8sErrors.IsServiceUnavailable(err) {
+			return false, nil
```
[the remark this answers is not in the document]
igomez06: So if you see I'm specifically checking an InternalError and ServiceUnavailable error, I don't want to return the error because then the poll ends reporting one of the errors above, when we don't want to do that. It is done this way because I am essentially pinging rancher until it's available after updating the rancher image.

tests/v2/codecoverage/Dockerfile.codecoverage
```diff
@@ -0,0 +1,21 @@
...
+ADD rancherlabs/corral-packages "$WORKSPACE"/rancherlabs/corral-packages
+ADD rancher "$WORKSPACE"/rancher
+
+RUN zypper -n install gcc binutils glibc-devel-static ca-certificates git-core wget curl unzip tar vim less file xz gzip sed gawk iproute2 iptables jq
+RUN zypper install -y -f docker && rpm -e --nodeps --noscripts containerd
+RUN zypper install -y python3-tox python3-base python3 libffi-devel libopenssl-devel
+RUN go mod download && \
+    go install gotest.tools/gotestsum@latest && \
+    go install golang.org/x/tools/cmd/cover@latest && \
+    go install github.com/rancherlabs/corral@v1.1.1 && \
```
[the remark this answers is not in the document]
igomez06: Yeah you're right I'll change that.

tests/v2/codecoverage/package/run_agent.sh
```diff
@@ -0,0 +1,259 @@
...
+    if [ -z $address ]; then
+        echo ""
+    # If given address is a network interface on the system, retrieve configured IP on that interface (only the first configured IP is taken)
+    elif [ -n "$(find /sys/devices -name $address)" ]; then
+        echo $(ip addr show dev $address | grep -w inet | awk '{print $2}' | cut -f1 -d/ | head -1)
+    # Loop through cloud provider options to get IP from metadata, if not found return given value
+    else
+        case $address in
+            awslocal)
+                echo $(curl -s http://169.254.169.254/latest/meta-data/local-ipv4)
```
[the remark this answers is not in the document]
igomez06: This is how the `rancher/rancher/package/run.sh` works. This is how the rancher agent image runs in a downstream provisioned cluster

tests/v2/codecoverage/rancherha/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+func main() {
+	results := corral.CorralConfigurations()
+	err := corral.SetupCorralConfig(results.CorralConfigVars)
+	if err != nil {
+		logrus.Fatalf("error setting up corral: %v", err)
+	}
+
+	configPackage := corral.CorralPackagesConfig()
+	path := configPackage.CorralPackagePath["aws-rke2-rancher-calico-v1.23.6-rke2r1-2.6.7"]
```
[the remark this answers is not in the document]
igomez06: Well this essentially a dumby value, I'm updating the rancher image after I have created it. Our custom rancher image, is always updated. The plan is to have a separate Jenkins job to build a new image every week. The scripts to do that already exist. In terms of k8s version, we can possible have this be config entry rather than it being hardcoded, as to keep up with k8s versions. I don't know if that a concern necessarily as are reporting code coverage not really validations if that makes sense.

tests/v2/codecoverage/package/Dockerfile.agenttest
```diff
@@ -0,0 +1,27 @@
...
+    curl -sLf https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl > /usr/bin/kubectl && \
+    chmod +x /usr/bin/kubectl
+
+ENV LOGLEVEL_VERSION v0.1.5
+
+RUN curl -sLf https://github.com/rancher/loglevel/releases/download/${LOGLEVEL_VERSION}/loglevel-${ARCH}-${LOGLEVEL_VERSION}.tar.gz | tar xvzf - -C /usr/bin
+
+ARG VERSION=dev
+LABEL io.cattle.agent true
+ENV AGENT_IMAGE igomez06/rancher-agent:${VERSION}
```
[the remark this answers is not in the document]
igomez06: This is a custom code coverage image.

tests/v2/codecoverage/scripts/build_docker_images.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+set -e
+
+ARCH=${ARCH:-"amd64"}
+REPO=igomez06
```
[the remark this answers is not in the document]
igomez06: I don't have access to the repo, and this was temporary until I could, if you could possibly help with that than we can move our custom image there rather than my own docker repo.

igomez06:
> looking at the jenkins run, I have some further questions:
> 
> * some of the logs are spammy, is there a way to fix this? over 80% of the logs are from about 5 minutes with logs that look like:
> 
> ```
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame sent
>  2022/12/19 16:01:01 (0xc000fcde40) Data frame received for 3
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame handling
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame sent
>  2022/12/19 16:01:01 (0xc000fcde40) Data frame received for 3
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame handling
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame sent
>  2022/12/19 16:01:01 (0xc000fcde40) Data frame received for 3
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame handling
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame sent
>  2022/12/19 16:01:01 (0xc000fcde40) Data frame received for 3
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame handling
>  2022/12/19 16:01:01 (0xc0007e19a0) (3) Data frame sent
> ...
> ```
> 
>  * why have all the subjobs been `aborted`, will they stop on their own?
> * looking at the finished test, is this the current code coverage? If not, what will the user do to find the final number?
> 
> ```
> 
>  ++ go tool cover -func=cover/merged.cov
>  ++ sed -E 's/\s+/ /g'
>  total: (statements) 27.4%
> ```
> 
>  * I see stuff like `github.com/rancher/rancher/pkg/agent/node/node.go:93.2,93.15 1 0` in the file merged.cov, but not sure what the numbers at the end mean. Is there some key?

To answer your questions Caleb, the spamming comes from killing the container of the pods running the code coverage images. I can look into it, but I'm not sure we can, as this is coming from the communication with the pod.
The tests are aborted because there were issues with the python tests hanging, that I have since fixed and added you to review that PR https://github.com/rancher/rancher/pull/39964
The number you showed is the final code coverage.
what you are seeing is not human readable you need to run go tool cover -func= to view them
Also I'm publishing the html report that is more human readable but because of our issues with Jenkins I'm waiting on Brandon to help reinstall the html plugin
```
                // publishHTML (target: [
                 //       allowMissing: false,
                 //       alwaysLinkToLastBuild: true,
                 //       keepAll: true,
                 //       reportDir: ".",
                 //       reportFiles: mergedHTMLFile,
                 //       reportName: "Single-click Code Coverage Test Results"
                 //     ])
```

tests/framework/extensions/codecoverage/codecoverage.go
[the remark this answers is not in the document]
caliskanugur: So this error catch' purpose only to get out of the poll if there are these errors, totally makes sense to me.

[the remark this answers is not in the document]
igomez06: So I return an error then it exits the poll before it's ready, what I can do instead is return an err on an else, just in case another error occurs.

tests/framework/clients/corral/corral.go
```diff
@@ -16,6 +17,19 @@ const (
 	debugFlag = "--debug"
 )
 
+// GetCorralEnvVar
+func GetCorralEnvVar(corralName, envVar string) (string, error) {
+	//corral vars ranchertestcoverage bootstrap_password
```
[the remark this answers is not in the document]
markusewalker: I agree with this and additionally the comment on line 20

tests/framework/extensions/codecoverage/codecoverage.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		clusterID := clusterStatus.ClusterName
+		if clusterID != localCluster {
+			dynamicClient, err := client.GetDownStreamClusterClient(clusterID)
+			if err != nil {
+				logrus.Errorf("could not connect to downstream cluster")
+				continue
+			}
+
+			pods, err := dynamicClient.Resource(corev1.SchemeGroupVersion.WithResource("pods")).Namespace("cattle-system").List(context.Background(), metav1.ListOptions{})
```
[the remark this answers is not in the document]
markusewalker: This is for the cattle-cluster-agent service though, right? So wouldn't it need to be here as well since on line 102, that is for the local cluster and not a downstream cluster like this is doing.

tests/framework/extensions/rke1/nodepools/nodepools.go
```diff
@@ -87,7 +87,7 @@ func NodePoolSetup(client *rancher.Client, nodeRoles []NodeRoles, ClusterID, Nod
 		nodePoolConfig.Etcd = roles.Etcd
 		nodePoolConfig.Worker = roles.Worker
 		nodePoolConfig.Quantity = roles.Quantity
-		nodePoolConfig.HostnamePrefix = "auto-rke1-" + strconv.Itoa(index)
+		nodePoolConfig.HostnamePrefix = "auto-rke1-" + strconv.Itoa(index) + ClusterID
```
markusewalker, line 90: Just needs a rebasing and this will be fixed, but this has been updated in a separate RKE1 PR.

tests/framework/pkg/config/config.go
```diff
@@ -45,3 +44,39 @@ func LoadConfig(key string, config interface{}) {
 	}
 
 }
+
+func UpdateConfig(key string, config interface{}) {
```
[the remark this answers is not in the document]
markusewalker: agreed, unsure if this updating configuration for the local cluster or a downstream cluster?

tests/framework/pkg/session/session.go
```diff
@@ -53,7 +54,7 @@ func (ts *Session) Cleanup() {
 
 // NewSession returns a `Session` who's cleanup method is registered with this `Session`
 func (ts *Session) NewSession() *Session {
-	sess := NewSession(ts.TestingT)
+	sess := NewSession()
```
markusewalker, line 57: Same nit as Ugur above, but seems this was by accident 😄

tests/integration/suite/conftest.py
```diff
@@ -101,7 +102,7 @@ def admin_mc():
     """Returns a ManagementContext for the default global admin user."""
     r = requests.post(AUTH_URL, json={
         'username': 'admin',
```
markusewalker, line 104: not sure if it even matters, but since password is abstracted to `SERVER_PASSWORD`, would we want to do the same but for `SERVER_USERNAME`?

tests/v2/codecoverage/scripts/build_docker_images.sh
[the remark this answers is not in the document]
markusewalker: asking as I'm unsure...did this ever get traction? I agree that we should move this out of personal repos if we can help it.

tests/v2/validation/provisioning/rke2/provisioning_node_driver_test.go
```diff
@@ -348,6 +347,7 @@ func (r *RKE2NodeDriverProvisioningTestSuite) ProvisioningRKE2CNICluster(provide
 }
 
 func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioning() {
+	r.T().Skip()
```
markusewalker, line 350: was this for debugging as well? We plan to remove this function anyway in the RKE2 refactoring so I guess it isn't that important.

tests/integration/suite/conftest.py
igomez06: It will always be admin, as this the main account.

tests/v2/codecoverage/scripts/build_docker_images.sh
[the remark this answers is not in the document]
igomez06: I'm trying to resolve this now. In the long run we do want to move this to our rancher-test docker hub.

go.mod
```diff
@@ -220,6 +220,7 @@ require (
 	github.com/fatih/color v1.13.0 // indirect
 	github.com/felixge/httpsnoop v1.0.1 // indirect
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
+	github.com/fvbommel/sortorder v1.0.1 // indirect
```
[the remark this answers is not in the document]
igomez06: I'm double checking but I don't think we actually need this anymore, it might have been something I needed beforehand when trying different k8s packages.

tests/framework/extensions/codecoverage/codecoverage.go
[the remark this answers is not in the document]
igomez06: What do you mean Markus? The cattle-cluster-agent pods are in the cattle-system namespace as well. I'm not quite following.

tests/framework/pkg/config/config.go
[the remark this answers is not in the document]
igomez06: This is to update to the CATTLE_TEST_CONFIG that the framework uses. I'll put up a comment

tests/framework/pkg/killserver/killserver.go
igomez06: Yes this really outside of the extensions, and more of system level thing I think it belongs pkg for sure

tests/framework/pkg/session/session.go
igomez06: I removed TestingT because we could just use logrus to log things within the framework rather than passing the T just to log

tests/framework/extensions/codecoverage/codecoverage.go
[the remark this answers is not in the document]
markusewalker: Ignore my last comment, please, I was confusing myself 😄

markusewalker (dismissed):
LGTM, awesome work!

go.mod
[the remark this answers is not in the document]
igomez06: `k8s.io/kubectl` needs which weirdly enough hasn't been an issue yet. This was something I added to fix a go.mod issues.

markusewalker (dismissed):

caliskanugur (approved):
Looks good to me

tests/v2/codecoverage/rancherha/main.go
```diff
@@ -0,0 +1,24 @@
...
+	testSession := session.NewSession()
+
+	corralConfig := corral.CorralConfigurations()
+	err := corral.SetupCorralConfig(corralConfig.CorralConfigVars, corralConfig.CorralConfigUser, corralConfig.CorralSSHPath)
+	if err != nil {
+		logrus.Fatalf("error setting up corral: %v", err)
+	}
+	configPackage := corral.CorralPackagesConfig()
+
+	path := configPackage.CorralPackageImages["aws-rke2-rancher-calico-v1.23.6-rke2r1-2.6.7"]
```
izaac, line 19: What can we do to get a predictable package image name here when creating a corral package with different values.

Like adding a section to `CATTLE_TEST_CONFIG`, generate/update a corral package yaml in the `corral-packages` packages folder and then `build_corral_packages.sh`?

tests/v2/codecoverage/setuprancher/main.go
```diff
@@ -0,0 +1,511 @@
...
+	corralName       = "ranchertestcoverage"
+	rancherTestImage = "ranchertest/rancher:v2.6-head"
+	namespace        = "cattle-system"
+	deploymentName   = "rancher"
+	clusterName      = "local"
+	// The json/yaml config key for the corral package to be build ..
+	userClusterConfigsConfigurationFileKey = "userClusterConfig"
+)
+
+type UserClusterConfig struct {
```
izaac, line 56: why add a separate yaml definition for a new config here instead of updating the existing `CATTLE_CONFIG_TEST` ?

tests/v2/codecoverage/setuprancher/main.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	//update userconfig
+	userClusterConfig := UserClusterConfig{}
+	userClusterConfig.Token = standardUserClient.Steve.Opts.TokenKey
+	userClusterConfig.Username = newUser.Name
+	userClusterConfig.AdminPassword = adminPassword
+	userClusterConfig.Clusters = standardClusterNames
+
+	rancherConfig.ClusterName = adminClusterNames[0]
```
izaac, line 150: Shouldn't `ClusterName` be a slice instead of a string so we could set multiple clusters for admin like with userClusterConfig?

tests/v2/codecoverage/setuprancher/main.go
```diff
@@ -0,0 +1,511 @@
...
+	config.UpdateConfig(rancher.ConfigurationFileKey, rancherConfig)
+
+	err = writeToConfigFile(userClusterConfig)
```
izaac, line 153: Why create a new config file if we can update the existing `CATTLE_CONFIG_TEST`?

tests/v2/codecoverage/setuprancher/main.go
```diff
@@ -0,0 +1,511 @@
...
+	})
+	if err != nil {
+		return err
+	} else if len(userList.Data) == 0 {
+		return fmt.Errorf("admin user not found")
+	}
+
+	adminUser := &userList.Data[0]
+	setPasswordInput := management.SetPasswordInput{
+		NewPassword: adminPassword,
```
izaac, line 506: Thanks for adding this to the post Rancher configuration.

And I just wonder if this `adminPassword` should be also included in the `CATTLE_TEST_CONFIG`?

tests/v2/codecoverage/package/run_rancher.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
izaac, line 1: Am I correct assuming most of the code in the scripts of this folder comes from existing scripts in the repo?

izaac (changes requested):
A couple questions and comments.

tests/v2/codecoverage/rancherha/main.go
igomez06: I'm not quite sure what you mean. Like different k8s versions and cnis? Would this be a concern? This is just a dummy value. When we update the deployment the latest rancher code coverage image is used.

tests/v2/codecoverage/setuprancher/main.go
igomez06: Because I'm keeping track of three clusters along with the user tokens used to provision this cluster. This is to glue the python and go code together.

tests/v2/codecoverage/setuprancher/main.go
igomez06: As of right now that isn't the behavior of the framework. Maybe that is something we could implement in the future ,but how it is designed right now. We only need one cluster when testing. I'm indexing the slice because createTestCluster returns a slice

tests/v2/codecoverage/setuprancher/main.go
igomez06: This is a separate config file, this config file is used to glue python and golang test runs. The python tests don't use CATTLE_CONFIG_TEST

tests/v2/codecoverage/setuprancher/main.go
igomez06: If you think we would do this a lot, I can add that. Since this is only to set the admin password after rancher deployment, I think for now we can just keep it how it is.

tests/v2/codecoverage/package/run_rancher.sh
igomez06: Yes, some of them are mine. Like the setup environment ones, but the others I pulled from the main rancher/rancher/scripts dir.

tests/v2/codecoverage/setuprancher/main.go
izaac: If it's by design why do we need an slice?

tests/v2/codecoverage/setuprancher/main.go
izaac: Can't the existing config get updated with this instead of creating a new one ? I'm still not see the reasoning behind creating a new file.

tests/v2/codecoverage/rancherha/main.go
izaac: If we don't care about the cluster Rancher is on, this is ok then.

tests/v2/codecoverage/setuprancher/main.go
izaac: I see where is this coming from now, we're good here.

tests/v2/codecoverage/package/run_rancher.sh
izaac: Cool, thanks for confirming

tests/v2/codecoverage/setuprancher/main.go
igomez06: I'll try to explain better sorry. So we need three clusters to test our python, and go tests. In other words go validations, python v3_need_cluster tests and v3_additional_tests. The latter needs an admin token, user token, and cluster name to run. This config is keep track of what are my clusters I created and the tokens used to create them. We only need this because of the python tests. Does that make sense?

tests/v2/codecoverage/setuprancher/main.go
igomez06: Because of the function `createTestCluster` on line 358

izaac: My bad, what I tried to ask is why the design only considers 1 cluster for admin but there are cases where there are multiple clusters for other users (`userClusterConfig`). The Rancher admin can also create multiple clusters. I am probably confused as why this shouldn't behave as how Rancher does.

Like the case here with `userClusterConfig.Clusters` it's a slice, I know this is only for the python tests and the generated config file that glues the configuration but is not consistent with the design.

tests/v2/codecoverage/setuprancher/main.go
izaac: Python tests doesn't use `CATTLE_CONFIG_TEST` but neither the `userClusterConfig` which in this case is used only to transport data for configuring the Python framework.

The same can't be done just updating `CATTLE_CONFIG_TEST` and use `yq` as done in the jenkinsfile over `userclusterconfig.yaml` ?

I am just curious, if this is just to separate the configuration for clarity im good.

tests/v2/codecoverage/setuprancher/main.go
igomez06: Ah sorry I see what you are saying. Because the python tests need a standard user cluster, and the go tests need a single admin cluster. Isn't that the case?

tests/v2/codecoverage/setuprancher/main.go
igomez06: Simply because we need the distinction between the admin clusters and the standard user clusters.

tests/v2/codecoverage/setuprancher/main.go
izaac: I get it but if we need multiple clusters for go tests what's the steps to follow.

Should we proceed with something like the `userClusterConfig` struct instead of using the `rancher.Config` and create a separate file?

tests/v2/codecoverage/setuprancher/main.go
izaac: I see I thought we could create an additional configuration key in the `CATTLE_CONFIG_TEST` to differentiate, but I guess it's clearer/easier this way.

tests/v2/codecoverage/setuprancher/main.go
igomez06: I guess the boils down to changing the current design. Currently the framework only supports the single cluster in the config file when running validations, if we keep this design then yes in a Jenkins environment I think we would still need the userClusterConfig to update the CATTLE_CONFIG_TEST for multiple jobs.

tests/v2/codecoverage/setuprancher/main.go
igomez06: We don't one of the main designs of the framework is dynamic generate a user token rather than having to manually create

tests/v2/codecoverage/setuprancher/main.go
izaac: I think I am facing this issue with the registries tests. 

I have two clusters besides the local to run the same go tests with different values. I'll try to follow the suggested `userClusterConfig` approach.

igomez06: I think that is a good idea, but also would it make sense to update the framework to support multiple clusters?

tests/v2/codecoverage/setuprancher/main.go
izaac: Ok just to wrap up. Let me know if this is correct.

Like what happens with `rancher.adminToken` in `CATTLE_CONFIG_TEST` where `createAdminToken` is used in go and then assigned to the struct `rancherConfig.AdminToken = token`  to then being applied using `config.UpdateConfig` back to `CATTLE_CONFIG_TEST`

But is not possible to have the same with a new configuration key using a struct like `userClusterConfig`

igomez06: It definitely is possible to have a new configuration key in `CATTLE_CONFIG_TEST` to use something like `userClusterConfig`, the reason I didn't do that was because I felt like code coverage setup was very different from the rest of our testing, and felt like it didn't belong in the `CATTLE_CONFIG_TEST` especially since the main/only reason I have it is to setup our python tests both integration and validation.

izaac (approved):
LGTM

igomez06:
@markusewalker @caliskanugur @izaac hey guys one last review, during hackwork 1.20 Golang was released that finally supported code coverage for applications. With new updates to rancher/wrangler the old method no longer worked so I updated this PR to use the new `go build -cover`.

slickwarren (approved):
lgtm

markusewalker (approved):

caliskanugur (approved):

izaac (approved):

PR #39964, opened 2022-12-21
[v2.6] Fix timeout issue with validate_cluster_state

igomez06:
## Problem
 There was a bug with how the timeout in validate_cluster_state was being set causing an absurdly long timeout. Also there is a chance that an updating state happens after it is checked with this function, so a conditional was added to not fail if the updating state is missed.
 
## Solution
 Removed `timeout = start_time + timeout` since it's just adding to be a super long timeout.
 
## Testing
 Jenkins runs


izaac (approved):
LGTM good catch

slickwarren (approved):
LGTM

vivek-infracloud (approved):
LGTM

PR #39965, opened 2022-12-21
[v2.7] Fix timeout issue with validate_cluster_state 

igomez06:
## Problem
 There was a bug with how the timeout in validate_cluster_state was being set causing an absurdly long timeout. Also there is a chance that an updating state happens after it is checked with this function, so a conditional was added to not fail if the updating state is missed.
 
## Solution
 Removed `timeout = start_time + timeout` since it's just adding to be a super long timeout.
 
## Testing
 Jenkins runs

### DIFF
--- a/tests/validation/tests/v3_api/test_node_label.py
+++ b/tests/validation/tests/v3_api/test_node_label.py
@@ -433,7 +433,8 @@ def test_node_label_custom_add_edit_addnode():
     # cluster will go into updating state
     cluster = validate_cluster_state(client, cluster, True,
                                      intermediate_state="updating",
-                                     nodes_not_in_active_state=[])
+                                     nodes_not_in_active_state=[],
+                                     timeout=600)
 
     node = client.reload(node)
     # Label should be added
@@ -700,7 +701,8 @@ def test_node_label_node_template_delete_api():
     # cluster will go into updating state
     cluster = validate_cluster_state(client, cluster, True,
                                      intermediate_state="updating",
-                                     nodes_not_in_active_state=[])
+                                     nodes_not_in_active_state=[],
+                                     timeout=600)
 
     node = client.reload(node)
     # label should be deleted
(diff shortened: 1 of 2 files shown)


izaac (approved):
LGTM

tests/validation/tests/v3_api/test_node_label.py
```diff
@@ -757,11 +759,13 @@ def test_node_label_custom_edit():
...
     # cluster will go into updating state
+    print("before validate cluster state")
     cluster = validate_cluster_state(client, cluster, True,
                                      intermediate_state="updating",
-                                     nodes_not_in_active_state=[])
+                                     nodes_not_in_active_state=[],
+                                     timeout=600)
     node = client.reload(node)
-
+    print("before validate set on node")
```
anupama2501, line 768: Don't think we need this print statement here.

tests/validation/tests/v3_api/test_node_label.py
```diff
@@ -757,11 +759,13 @@ def test_node_label_custom_edit():
     node_labels[test_label] = new_value
     client.update(node, labels=node_labels)
     # cluster will go into updating state
+    print("before validate cluster state")
```
anupama2501, line 762: we can remove this print statement here?

tests/validation/tests/v3_api/test_node_label.py
igomez06: Yes we don't I'll fix that right now.

tests/validation/tests/v3_api/test_node_label.py
igomez06: Yes I'll will remove

anupama2501 (commented):
Just a nit LGTM otherwise

slickwarren (approved):

anupama2501 (approved):
LGTM

slickwarren (approved):
lgrm

PR #39998, opened 2022-12-23
[2.7] Use `dig` to get external IP address

HarrisonWAffel:
## Problem
When deploying the Rancher node agent no value for `CATTLE_ADDRESS` is provided. `CATTLE_ADDRESS` is used by the node agent to determine the external IP address of the host. Previously, this value was not used by the node agent so its value was unimportant. However, the implementation of #39497 requires that this variable hold the correct external IP address. 

If this environment variable is not provided upon initial deployment of the agent, then the command `ip -o route get 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p'` will be run by the agent on startup in an attempt to determine the external IP address of the host. This works for most providers, however for some (Azure & Linode, and potentially others) it actually returns the private IP address of the host. 

## Solution
Attempt to get the external IP address using the `dig` command. In the scenario where this command does not work, we will fall back to the existing behavior. 

 
### Manual Testing

I've used these changes on a number of providers and have confirmed they have the correct external IP address

## QA Testing Considerations
I do not have access to a Linode account to test this change with, but I do know that this resolves the issue for Azure clusters.
 
Make sure azure node driver clusters still report proper IP addresses, from my testing I have not seen a problem but good to double check.

### DIFF
--- a/package/Dockerfile.agent
+++ b/package/Dockerfile.agent
@@ -7,7 +7,7 @@ ARG ARCH=amd64
 
 ENV KUBECTL_VERSION v1.23.6
 RUN zypper -n install --no-recommends curl ca-certificates jq git-core hostname iproute2 vim-small less \
-	bash-completion acl openssh-clients tar gzip xz gawk sysstat && \
+	bash-completion bind-utils acl openssh-clients tar gzip xz gawk sysstat && \
     zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* && \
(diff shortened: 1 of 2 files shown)


annablender (approved):
Could you provide a list of providers you tested this with? AWS, Amazon and GKE as well right? Did you see the private IP returned in any case with `dig`

jiaqiluo (approved):
LGTM.

jiaqiluo (approved):

annablender (approved):

Issue #39992, opened 2022-12-23
Support the use of PSACT in RKE1 1.25 cluster 


jiaqiluo:
Main issue: https://github.com/rancher/rancher/issues/38701

We want to support using PSACT (Pod Security Admission Configuration Template) in the RKE1 1.25 downstream cluster.

The implementation should be in [rancher/webhook](https://github.com/rancher/webhook)

Generally speaking,  rancher-webhook needs to mutate and validate the cluster spec accordingly when the filed `DefaultPodSecurityAdmissionConfigurationTemplateName` is set or updated/ or removed in the RKE1 1.25 downstream cluster.

jiaqiluo:
waiting for a new rancher-webhook RC

markusewalker:
This issue is covered in the overall K8s 1.25 test plan. Once that internal test plan has been marked as completed, I will close this issue accordingly.

jiaqiluo:
This issue can be validated in the lastest v2.7-head.

markusewalker:
Initial use cases have been validated in the internal test plan. Advanced workflows, not necessarily specific to being able to provision a PSACT enabled RKE1 cluster, that may have issues will be logged in separate issues.

Closing this issue for the above reason.

PR #40006, opened 2022-12-27
Fix Panic Error Caused by Rancher-Webhook

jiaqiluo:
## Issue: 

https://github.com/rancher/rancher/issues/38701
https://github.com/rancher/rancher/issues/40009
https://github.com/rancher/rancher/issues/39992
 
## Problem

Due to the addition of checks for PSA in rancher-webhook, it does more validations on the request for creating/updating a v3 Management cluster and could return an error if the validation fails.  
Previously, rancher-webhook did not return such an error and rancher also did not catch it, so a panic error did not happen before but happens now. 

## Solution

 
Return the error properly. 

## Testing

Steps:
- create an RKE1 cluster in rancher; during the creation, edit the cluster as YAML and add the following 
```
default_pod_security_admission_configuration_template_name: privileged
```
- update the above cluster to change the template name to an non-existing one 

Expected Results:
The update should fail with an error like the following: 
image

Also, there should be no panic error in rancher's logs. 

### Manual Testing

same as the above
### Automated Testing

n/a 

## QA Testing Considerations

 same as the above

cluster creation and update

### DIFF
--- a/pkg/api/norman/store/cluster/cluster_store.go
+++ b/pkg/api/norman/store/cluster/cluster_store.go
@@ -235,7 +235,7 @@ func (r *Store) Create(apiContext *types.APIContext, schema *types.Schema, data 
 		return nil, err
 	}
 
-	//check if template is passed. if yes, load template data
+	// check if template is passed. if yes, load template data
 	if hasTemplate(data) {
 		clusterTemplateRevision, clusterTemplate, err := r.validateTemplateInput(apiContext, data, false)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


kinarashah (dismissed):

cmurphy (changes requested):
Could you add the same fix for clustertemplaterevisions

https://github.com/rancher/rancher/blob/28a84c710edf49a4acedfdc3624d427a36124672/pkg/api/norman/store/clustertemplate/store.go#L159

jiaqiluo:
> Could you add the same fix for clustertemplaterevisions
> 
> https://github.com/rancher/rancher/blob/28a84c710edf49a4acedfdc3624d427a36124672/pkg/api/norman/store/clustertemplate/store.go#L159

@cmurphy  Because the rancher-webhook does not validate the `ClusterTemplateRevision`, the panic will not happen when updating a ClusterTemplateRevision. Do you still want me to add the same fix for ClusterTemplateRevision ?

cmurphy:
@jiaqiluo yes, it is still broken in the same way even if it doesn't affect the webhook

jiaqiluo:
Based on the offline discussion with @cmurphy,  the same "error return" is added to the **Create** method for ClusterTemplte for parity with the logic for ClusterStore. However, rancher-webhook does not trigger the panic error when updating Cluster Template.

cmurphy (approved):

kinarashah (approved):

PR #40010, opened 2022-12-27
Steve pagination and integration tests

cmurphy:
## Issue: https://github.com/rancher/rancher/issues/38427

 
More details in the commit messages. Summary:

Commit 1: add integration tests for the steve API, including some framework changes to make testing more  flexible.

Commit 2: have the integration tests output response samples and create an API document. The idea is to give users automated, validated examples of how to use the steve API.

Commit 3: bump steve and apiserver to include new pagination features. The tests are left unchanged, to show that the new features don't break the existing API.

Commit 4: add tests for the new features.

PRs https://github.com/rancher/rancher/pull/40048 and https://github.com/rancher/rancher/pull/40049 show the intermediate changes passing CI on their own.


tests/framework/clients/rancher/v1/client.go
```diff
@@ -215,6 +232,73 @@ func (c *SteveClient) Delete(container *SteveAPIObject) error {
 	return c.apiClient.Ops.DoResourceDelete(c.steveType, &container.Resource)
 }
 
+func (c *NamespacedSteveClient) Create(container any) (*SteveAPIObject, error) {
```
igomez06, line 235: I want to make sure I understand the purpose of the NamespacedSteveClient. This really more for the usage of the `List` command set correct? Because it looks like the other commands don't change. Next question this for us to have the ability to list resources from a specific namespace? What would be the need to do that? Couldn't we just use the url.Values to set the namespace? I might be wrong just thought I'd ask.

tests/framework/clients/rancher/v1/client.go
```diff
@@ -52,9 +54,15 @@ type SteveClient struct {
...
+type NamespacedSteveClient struct {
+	SteveClient
+	namespace string
+}
+
 type SteveOperations interface {
-	List(opts *types.ListOpts) (*SteveCollection, error)
-	ListAll(opts *types.ListOpts) (*SteveCollection, error)
+	List(params url.Values) (*SteveCollection, error)
+	ListAll(params url.Values) (*SteveCollection, error)
```
igomez06, line 65: Nice! I had noticed our Steve client listing functionality was very restrictive and wouldn't work the same as the v3 client when attempting to put a specific values in the  ListOpts.

tests/framework/pkg/clientbase/ops.go
```diff
@@ -102,21 +102,28 @@ func (a *APIOperations) DoGet(url string, opts *types.ListOpts, respObject inter
 }
 
 func (a *APIOperations) DoList(schemaType string, opts *types.ListOpts, respObject interface{}) error {
+	collectionURL, err := a.GetCollectionURL(schemaType)
```
igomez06, line 105: Make this update doesn't affect our v3/norman API calls?

tests/v2/integration/setup/main.go
```diff
@@ -31,6 +31,7 @@ func main() {
 	rancherConfig.AdminToken = token.Token
 	rancherConfig.Host = hostURL
 	rancherConfig.Cleanup = &cleanup
+	rancherConfig.ClusterName = "local"
```
igomez06, line 34: Is this necessary? I don't want to limit our integration setup to just local, as we have plans of having an imported k3d cluster work in our integration test. I don't think this is necessary.

tests/v2/integration/steveapi/steve_api_test.go
```diff
@@ -0,0 +1,1543 @@
...
+	lastContinueToken string
+	lastRevision      string
+}
+
+func (s *SteveAPITestSuite) TearDownSuite() {
+	s.session.Cleanup()
+}
+
+func (s *SteveAPITestSuite) SetupSuite() {
+	testSession := session.NewSession(s.T())
```
igomez06, line 165: This goes for the namespace creation, secrets, and role creation beforehand. The reason you are using the kubeapi extensions is you want to create these resources with the controllers, and then verify behavior with the Steve client? I'm just curious, and a little description will help understand.

igomez06 (commented):
Awesome work, I just had some questions.

tests/v2/integration/steveapi/steve_api_test.go
```diff
@@ -451,6 +453,246 @@ func (s *SteveAPITestSuite) TestList() {
...
+				{"name": "test4", "namespace": "test-ns-3"},
+				{"name": "test4", "namespace": "test-ns-2"},
+				{"name": "test4", "namespace": "test-ns-5"},
+				{"name": "test5", "namespace": "test-ns-1"},
+				{"name": "test5", "namespace": "test-ns-3"},
+				{"name": "test5", "namespace": "test-ns-4"},
+				{"name": "test5", "namespace": "test-ns-2"},
+				{"name": "test5", "namespace": "test-ns-5"},
+			},
+		},
```
KevinJoiner, line 517: Will this always be the correct order since we are only sorting on metadata.name

KevinJoiner (changes requested):
Looks good could you add just one test that uses a combination of sort, filter and pagination.

tests/v2/integration/steveapi/steve_api_test.go
cmurphy: I want to avoid using the norman API as much as possible since we are phasing it out. I could have created these through steve, but since I don't care about testing how these resources are created, the simplest way to create a resource in kubernetes is to use kubernetes itself, so that was my preference.

tests/v2/integration/steveapi/steve_api_test.go
cmurphy: Theoretically there is no guarantee, but practically there is because the resource order from etcd is deterministic and so the result of the sort is also deterministic. But I agree this is a little wonky, I think I'll change this to ignore the irrelevant  field for these types of cases.

tests/framework/clients/rancher/v1/client.go
igomez06: Ok cool totally understand. One question, maybe I just missed it, but I don't see how the namespace field for the namespace client is being used for the other CRUD operations.

tests/v2/integration/steveapi/steve_api_test.go
igomez06: Sounds good, just was curious. Thank you!

igomez06:
@cmurphy also do we need a v2.6 backport as well?

tests/framework/clients/rancher/v1/client.go
cmurphy: You're right, I wasn't using any of these other operations in my tests so I skipped over actually implementing it, will fix.

cmurphy (commented):
@igomez06 
> also do we need a v2.6 backport as well?

No, this will only be for 2.7

igomez06:
> @igomez06
> 
> > also do we need a v2.6 backport as well?
> 
> No, this will only be for 2.7

Well you did some great work with the Steve client, that I think would be worth to have in our v2.6 branch. Would it be possible to have a v2.6 PR but with just the Steve client changes you made?

cmurphy:
@igomez06 sure thing :+1:

cmurphy:
* rebased on release/v2.7 - since steve was bumped for the warning headers change, this PR no longer includes go.mod changes
* fixed the NamespacedSteveClient operations - only Create needed to be changed, since ByID/Replace/Update/Delete work on the resource's ID which includes the namespace, also added another set of tests just to test these
* @KevinJoiner I added some combo tests

KevinJoiner (approved):

igomez06 (approved):
Looks good.

igomez06:
@cmurphy the PR looks good, just curious when we can expect a PR for the Steve updates for the release/v2.6 branch?

cmurphy:
@igomez06 I'll create that backport now :+1:

PR #40015, opened 2022-12-28
[2.7] Support CATTLE_WHITELIST_ENVVARS in RKE2 provisioning code

sbstp:
This allows passing down environment variables to rancher machine during provisioning.

## Problem

This is an alternative solution to #39375 because there were [concerns](https://github.com/rancher/rancher/pull/39375#issuecomment-1351688797) regarding security. We need to be able to enable debug logging in RKE2.
 
## Solution

Instead of passing the `--debug` flag to rancher machine, we allow RKE2 provisioning to pass down to rancher machine whitelisted environments variables, one of which can be `MACHINE_DEBUG`, which solves the needs of the original issue reporter.
 
### Manual Testing

1. Start rancher with `MACHINE_DEBUG=true` and `CATTLE_WHITELIST_ENVVARS=NO_PROXY,HTTP_PROXY,HTTPS_PROXY,MACHINE_DEBUG` env vars.
2. Create an RKE2 cluster
3. Make sure rancher machine debug logs are displayed

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
+++ b/pkg/controllers/provisioningv2/rke2/machineprovision/args.go
@@ -24,7 +24,6 @@ import (
 	"github.com/rancher/wrangler/pkg/generic"
 	"github.com/rancher/wrangler/pkg/kv"
 	name2 "github.com/rancher/wrangler/pkg/name"
-	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -100,11 +99,7 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 			Name:      name2.SafeConcatName(infraObj.meta.GetName(), "machine", "driver", "secret"),
 			Namespace: infraObj.meta.GetNamespace(),
 		},
-		Data: map[string][]byte{
-			"HTTP_PROXY":  []byte(os.Getenv("HTTP_PROXY")),
-			"HTTPS_PROXY": []byte(os.Getenv("HTTPS_PROXY")),
-			"NO_PROXY":    []byte(os.Getenv("NO_PROXY")),
-		},
+		Data: getWhitelistedEnvVars(),
 	}
 	machine, err := rke2.GetMachineByOwner(h.machines, infraObj.meta)
 	if err != nil && (create || !errors.Is(err, rke2.ErrNoMachineOwnerRef)) {
@@ -132,12 +127,6 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 		fmt.Sprintf("--driver-hash=%s", hash),
 		fmt.Sprintf("--secret-namespace=%s", infraObj.meta.GetNamespace()),
 		fmt.Sprintf("--secret-name=%s", secretName),
-	}
-
-	// only in trace because machine has sensitive details and we can't control who debugs what in there easily
-	if logrus.GetLevel() >= logrus.TraceLevel {
-		// add --debug to pass directly to machine
-		cmd = append(cmd, "--debug")
 	}
 
 	if create {
@@ -350,4 +339,12 @@ func hashFile(path string) (string, error) {
 	}
 
 	return hex.EncodeToString(hasher.Sum(nil)), nil
+}
+
+func getWhitelistedEnvVars() map[string][]byte {
+	result := make(map[string][]byte)
+	settings.IterateWhitelistedEnvVars(func(name, value string) {
+		result[name] = []byte(value)
+	})
+	return result
 }
--- a/pkg/jailer/jailer.go
+++ b/pkg/jailer/jailer.go
@@ -77,20 +77,9 @@ func CreateJail(name string) error {
 	return nil
 }
 
-func WhitelistEnvvars(envvars []string) []string {
-	wl := settings.WhitelistEnvironmentVars.Get()
-	envWhiteList := strings.Split(wl, ",")
-
-	if len(envWhiteList) == 0 {
-		return envvars
-	}
-
-	for _, wlVar := range envWhiteList {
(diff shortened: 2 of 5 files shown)


pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -100,11 +99,7 @@ func (h *handler) getArgsEnvAndStatus(infraObj *infraObject, args map[string]int
 			Name:      name2.SafeConcatName(infraObj.meta.GetName(), "machine", "driver", "secret"),
 			Namespace: infraObj.meta.GetNamespace(),
 		},
-		Data: map[string][]byte{
-			"HTTP_PROXY":  []byte(os.Getenv("HTTP_PROXY")),
-			"HTTPS_PROXY": []byte(os.Getenv("HTTPS_PROXY")),
-			"NO_PROXY":    []byte(os.Getenv("NO_PROXY")),
-		},
```
sbstp, line 107: Maybe we should always insert these in `WhitelistEnvironmentVars` to keep the current behavior. `WhitelistEnvironmentVars` will default to `NO_PROXY,HTTP_PROXY,HTTPS_PROXY` but if you set it manually to like `MACHINE_DEBUG`, you need to remember to put the proxy vars.

jiaqiluo (approved):
The changes LGTM.
We need to update Racher Doc to introduce this new env var and how to use it.
@sbstp Can you also handle the update on the Doc side?

sbstp:
> @sbstp Can you also handle the update on the Doc side?
Yeah I'd like to. I'll ask for some pointers tomorrow to do it.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
snasovich: I think it's fine to keep the current solution as more flexible (user has complete control over which env vars are carried over) and matching the behavior of jailer that is used for RKE1 provisioning.

snasovich (approved):
LGTM

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -351,3 +340,19 @@ func hashFile(path string) (string, error) {
 
 	return hex.EncodeToString(hasher.Sum(nil)), nil
 }
+
+func getWhitelistedEnvVars() map[string][]byte {
```
Oats87, line 344: Is there a specific reason we are duplicating logic already found in the codebase i.e. here: https://github.com/rancher/rancher/blob/d657043de1dc2110562d9035b6c1c6dc9d9232ed/pkg/jailer/jailer.go#L80 ?

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
sbstp: The functions operate on different data structures which is why I decided to copy it and refactor the copy. The `WhitelistEnvvars` function returns a list of string in which each item is a string formatted with `name=value`. The copy I made returns a `map[string][]byte` which is a better fit for the secret used in RKE2.

`WhitelistEnvvars` also takes a parameter which is a seed of already existing `name=value` strings coming from the `*exec.Cmd` object. This parameter is not needed in the RKE2 case. I could have re-used that function and added some extra code to parse a list of `name=value` strings into a `map[string][]byte` but it felt inefficient and ugly to me.

snasovich: In this case the signature of the function you introduced is a lot more natural and can be used instead of the one @Oats87 referenced and have jailer use the function you introduced to range through the returned map and append `name=value` pairs to the slice.
The only question here is where this new function should reside, `setting.go` is one option I guess.

I don't feel too strongly about having to do this, though a perfectionist in me would prefer for it to be done.

sbstp: I think part of the problem is that [Cmd.Env](https://pkg.go.dev/os/exec#Cmd) which is part of the standard library is a list of `name=value` strings. If we want to seed the map of environment variables with initial data from this object, as done in the RKE1 code, we'll have to parse it.

It might be possible to change that functions' signature to use a callback function instead such as,

```go
settings.IterWhitelistedEnvVars(func (name, value string) {
  cmd.Env = append(cmd.Env, name + "=" + value)
})

// ...

settings.IterWhitelistedEnvVars(func (name, value string) {
  secret.Data[name] = []byte(value)
})
```

sbstp:
I pushed a new commit that tries to have more code re-use, let me know what you think

pkg/settings/setting.go
```diff
@@ -437,3 +437,17 @@ func GetRancherVersion() string {
 	}
 	return strings.TrimPrefix(rancherVersion, "v")
 }
+
+// IterWhitelistedEnvVars iterates over the environment variables whitelisted by CATTLE_WHITELIST_ENVVARS.
+// If the variable is whitelisted but unset or empty, the handler function will not be called.
+func IterWhitelistedEnvVars(handler func(name, value string)) {
```
snasovich, line 443: I like this approach. One nit is the function name, I don't see us generally using `Iter` as prefix, so maybe go for full name like `IterateWhitelistedEnvVars` which is not _that_ much longer.

Otherwise LGTM.

pkg/jailer/jailer.go
```diff
@@ -77,20 +77,9 @@ func CreateJail(name string) error {
...
-	}
-
-	for _, wlVar := range envWhiteList {
-		wlVar = strings.TrimSpace(wlVar)
-		if val := os.Getenv(wlVar); val != "" {
-			envvars = append(envvars, wlVar+"="+val)
-		}
-	}
-
+func getWhitelistedEnvVars(envvars []string) []string {
```
snasovich, line 80: Nice that you made it unexported, no reason for it to be exported in `jailer` pkg.

snasovich (dismissed):
LGTM, minor nit.

pkg/settings/setting.go
sbstp: Sure!

Oats87 (approved):

Issue #39996, opened 2022-12-23
Add feature Apps' Namespaces to the list of Exemptions.Namespaces of the Built-in "Restricted" PSACT


jiaqiluo:
Main issue: https://github.com/rancher/rancher/issues/38701

To make Rancher components work properly in a downstream 1.25 cluster, we need to add those namespaces used by our feature APps to the list of `Exemptions.Namespaces` of the built-in "restricted" PSACT.

markusewalker:
This issue is covered in the overall K8s 1.25 test plan. Once that internal test plan has been marked as completed, I will close this issue accordingly.

jiaqiluo:
can be validated in the lastest v2.7-head

sowmyav27:
@jiaqiluo can we add a QA Test Template for this issue?

jiaqiluo:
### Root cause
Rancher components suck as cattle-agent, and feature apps need privileged permissions to work in the downstream clusters. Therefore, we need to put their namespaces in the exception list of the "restricted" PSACT. 

### What was fixed, or what changes have occurred
Add those namespaces to the exception list of the "restricted" PSACT. 

### Areas or cases that should be tested

- provision or update a node-driver RKE1/RKE2/K3S  downstream cluster with `spec.defaultPodSecurityAdmissionConfigurationTemplateName = restricted` 
- install feature Apps into the cluster 

Expected:
- cluster provisioning/updating should succeed
- all rancher components should work in the downstream cluster, and pods run up without errors 
- feature apps should be installed successfully, and pods run up without errors 


### What areas could experience regressions?

- cluster provision
- feature apps 

  
### Are the repro steps accurate/minimal?

yes

jiaqiluo:
@sowmyav27  @markusewalker 
Because the same PSACT is used in both RKE1 and RKE2/K3s clusters, and we can validate this issue with RKE1 downstream cluster as of today, I am going to move this issue back to to-test.

markusewalker:
Validated test cases in the internal test plan pertaining to this issue. Therefore, I will be marking this issue as completed.

NOTE: More thorough testing will occur in this [issue](https://github.com/rancher/rancher/issues/39366).

markusewalker:
Putting test results from the internal K8s 1.25 test plan here for track-ability:

**ENVIRONMENT DETAILS**
- Rancher install: Helm
- Rancher version: `v2.7-head`
- Browser: Chrome

**TEST RESULT**
[...]

PR #39956, opened 2022-12-21
Update system namespaces for Helm charts

stormqueen1990:
## Problem

* Ensure our apps are installable once PSAs are rolled out on Rancher clusters.
 
## Solution

* Mark charts namespaces as system namespaces.

PR #40029, opened 2022-12-29
Update the default restricted PSACT

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/39996

~This PR updates the default `restricted` PSACT so that  its exemption namespaces list matches  the list of SystemNamepaces under Settings~
 
Three changes in this PR:
1. revert https://github.com/rancher/rancher/pull/39956, which was made because we thought we would use the setting SystemNamesapces as the source of the exemption namespaces.
2. Move namespaces under the setting SystemNamespaces to a standalone list for readability.
3. list the namespaces used by feature charts in the PSACT instead of getting them from the settings SystemNamepaces.


pkg/data/management/podadmissionconfigurationtemplate_data.go
```diff
@@ -21,3 +25,51 @@ func addDefaultPodSecurityAdmissionConfigurationTemplates(management *config.Man
...
+				EnforceVersion: "latest",
+				Audit:          "restricted",
+				AuditVersion:   "latest",
+				Warn:           "restricted",
+				WarnVersion:    "latest",
+			},
+			Exemptions: v3.PodSecurityAdmissionConfigurationTemplateExemptions{
+				Usernames:      []string{},
+				RuntimeClasses: []string{},
+				Namespaces:     strings.Split(settings.SystemNamespaces.Get(), ","),
```
Oats87, line 47: I'm not sure this is the best way to do this.... there are a LOT of system namespaces in there that don't necessarily need PSA exemptions.

Oats87 (changes requested):

pkg/data/management/podadmissionconfigurationtemplate_data.go
jiaqiluo: I know it is an overkill, but other opinions are unnecessarily complex:
option 1: hardcode and maintain the list inside the object instead of getting it from settings.SystemNamespaces
option 2:  an extra controller or mutator to add the PSP labels to the namespaces used by rancher and rancher-managed APPs.

Oats87: It's not just a matter of it being overkill, this is borderline a security risk if you ask me. I would be much happier if we had an additional setting for specific PSA-exempt namespaces. This would allow much more granularity. Additionally, we should probably have a cluster-level PSA exemption for namespaces, if the data structure for that won't be too complicated.

Oats87 (changes requested):
I am not comfortable with these changes being introduced in the manner that you are introducing them -- I believe this to be a borderline security risk. We should be much more strategic about the namespaces we grant exemption to, and preferably, with documentation on WHY we want to grant exemption.

pkg/data/management/podadmissionconfigurationtemplate_data.go
jiaqiluo: Based on the offline discussion, we will list the namespaces used by feature charts in the PSACT instead of getting them from the settings.SystemNamepaces.

pkg/data/management/podadmissionconfigurationtemplate_data.go
```diff
@@ -6,13 +6,38 @@ import (
...
+	"rancher-alerting-drivers",   // Alert Driver
+	"cis-operator-system",        // CIS Benchmark
+	"cattle-csp-adapter-system",  // CSP Adapter
+	"cattle-externalip-system",   // External IP Webhook
+	"cattle-gatekeeper-system",   // Gatekeeper
+	"istio-system",               // Istio and Sub-charts
+	"cattle-istio-system",        // Kiali
+	"cattle-logging-system",      // Logging
+	"cattle-windows-gmsa-system", // Windows GMSA
+	"cattle-wins-system",         // Wins Upgrader
```
snasovich, line 30: Wins-upgrader doesn't have 2.7.0 chart either, it's for unsupported RKE1 Windows upgrades

pkg/apis/management.cattle.io/v3/pod_security_types.go
```diff
@@ -41,51 +41,3 @@ type PodSecurityAdmissionConfigurationTemplateExemptions struct {
 	RuntimeClasses []string `json:"runtimeClasses"`
 	Namespaces     []string `json:"namespaces"`
 }
-
-func NewPodSecurityAdmissionConfigurationTemplateRestricted() *PodSecurityAdmissionConfigurationTemplate {
```
jiaqiluo, line 45: Note: 
Those two functions do not need to be in the API package, so I moved to `pkg/data/management/podadmissionconfigurationtemplate_data.go` where the only place they are called.

pkg/data/management/podadmissionconfigurationtemplate_data.go
```diff
@@ -6,13 +6,38 @@ import (
...
+	"rio-system",                 // RIO
```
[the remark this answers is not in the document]
jiaqiluo: this is removed.

pkg/data/management/podadmissionconfigurationtemplate_data.go
jiaqiluo: This is removed.

pkg/data/management/podadmissionconfigurationtemplate_data.go
```diff
@@ -6,13 +6,36 @@ import (
...
+	"cattle-epinio-system",       // Epinio
+	"cattle-fleet-system",        // Fleet
+	"longhorn-system",            // Longhorn
+	"cattle-neuvector-system",    // Neuvector
+	"cattle-monitoring-system",   // Monitoring and Sub-charts
+	"rancher-alerting-drivers",   // Alert Driver
+	"cis-operator-system",        // CIS Benchmark
+	"cattle-csp-adapter-system",  // CSP Adapter
+	"cattle-externalip-system",   // External IP Webhook
+	"cattle-gatekeeper-system",   // Gatekeeper
```
stormqueen1990, line 25: The latest version of the Gatekeeper chart [brings PSAs with it](https://github.com/rancher/charts/blob/8f6c52862117bdf7a6ccfed29a0c1847bfbd441e/charts/rancher-gatekeeper/101.1.0%2Bup3.10.0/values.yaml#L45-L50). If we wish to honour those perhaps we should not add the namespace for Gatekeeper here (unless it is the case that Gatekeeper will fail to install if it's not here).

stormqueen1990 (commented):
I noticed that `cattle-resources-system` is not on the list (the namespace for `rancher-backup`). Is that intentional?

jiaqiluo:
> I noticed that `cattle-resources-system` is not on the list (the namespace for `rancher-backup`). Is that intentional?

Yes, it is intentional. Rancher-backup is for backing up Rancher itself, so it will not be installed in the downstream cluster.

pkg/data/management/podadmissionconfigurationtemplate_data.go
jiaqiluo: I think we still need to include the namespace because:
1. We support PSA in k8s 1.23 and above, so if users can use the PSACT in 1.23/1.24 cluster and install the old version of Gatekeeper, Gatekeeper can still be installed 
2. The latest version of Gatekeeper appends the PS labels to its namespace, which will overwrite the cluster-level default configuration for this namespace.

pkg/settings/setting.go
```diff
@@ -24,47 +24,6 @@ var (
 	provider       Provider
 	InjectDefaults string
 
-	systemNamespaces = []string{
```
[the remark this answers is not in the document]
Oats87: Instead of fully reverting this, can we keep the readability change but omit the namespaces from #39956 ?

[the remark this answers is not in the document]
jiaqiluo: Done. 
Please check the last(3rd) commit for the required changes.

Oats87 (approved):

stormqueen1990 (approved):
LGTM!

Issue #38775, opened 2022-08-30
Add user agent to handlers requests


rmweir:
**Is your feature request related to a problem? Please describe.**
Kube API requests contain little info on where in rancher they were made from.

**Describe the solution you'd like**
Want requests to kube apiserver to have agent that reflects which handler in rancher made them.

**Describe alternatives you've considered**

**Additional context**

samjustus:
LOE s

moio:
If done at the wrangler or lasso levels, this would be useful in fleet too (cc @manno)

PR #40033, opened 2022-12-30
Add useragent to handlers

rmweir:
Some handlers have been configured to use clients with useragents that are named to reference the handlers that use them.

## Issue: 

https://github.com/rancher/rancher/issues/38775

## Problem

We would like to use the new useragent functionality in a small way that only mainly only affects team1 before rolling it out to more parts of rancher.
 
## Solution

Add useragent to some select handlers.

### DIFF
--- a/pkg/controllers/management/auth/register.go
+++ b/pkg/controllers/management/auth/register.go
@@ -56,14 +56,14 @@ func RegisterIndexers(scaledContext *config.ScaledContext) error {
 }
 
 func RegisterEarly(ctx context.Context, management *config.ManagementContext, clusterManager *clustermanager.Manager) {
-	prtb, crtb := newRTBLifecycles(management)
-	gr := newGlobalRoleLifecycle(management)
-	grb := newGlobalRoleBindingLifecycle(management, clusterManager)
+	prtb, crtb := newRTBLifecycles(management.WithAgent("mgmt-auth-crtb-prtb-controller"))
+	gr := newGlobalRoleLifecycle(management.WithAgent(grController))
+	grb := newGlobalRoleBindingLifecycle(management.WithAgent(grbController), clusterManager)
 	p, c := newPandCLifecycles(management)
 	u := newUserLifecycle(management, clusterManager)
-	n := newTokenController(management)
+	n := newTokenController(management.WithAgent(tokenController))
 	ac := newAuthConfigController(ctx, management, clusterManager.ScaledContext)
-	ua := newUserAttributeController(management)
+	ua := newUserAttributeController(management.WithAgent(userAttributeController))
 	s := newAuthSettingController(management)
 	rt := newRoleTemplateLifecycle(management, clusterManager)
 	grbLegacy := newLegacyGRBCleaner(management)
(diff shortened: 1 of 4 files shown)


pkg/controllers/management/auth/register.go
```diff
@@ -56,14 +56,14 @@ func RegisterIndexers(scaledContext *config.ScaledContext) error {
...
 
 func RegisterEarly(ctx context.Context, management *config.ManagementContext, clusterManager *clustermanager.Manager) {
-	prtb, crtb := newRTBLifecycles(management)
-	gr := newGlobalRoleLifecycle(management)
-	grb := newGlobalRoleBindingLifecycle(management, clusterManager)
+	prtb, crtb := newRTBLifecycles(management.WithAgent("mgmt-auth-crtb-prtb-controller"))
+	gr := newGlobalRoleLifecycle(management.WithAgent(grController))
+	grb := newGlobalRoleBindingLifecycle(management.WithAgent(grbController), clusterManager)
 	p, c := newPandCLifecycles(management)
 	u := newUserLifecycle(management, clusterManager)
```
cmurphy, line 63: I know the plan isn't to roll this out everywhere but I'm not clear why some of these were chosen and not others, for instance why do we want this for user attributes but not yet for users?

pkg/controllers/managementuser/rbac/handler_base.go
```diff
@@ -116,11 +118,11 @@ func Register(ctx context.Context, workload *config.UserContext) {
...
-		ProjectLister:       workload.Management.Management.Projects(workload.ClusterName).Controller().Lister(),
+		ProjectLister:       management.Management.Projects(workload.ClusterName).Controller().Lister(),
 	}
 
 	workload.Core.Namespaces("").AddLifecycle(ctx, "namespace-auth", newNamespaceLifecycle(r, sync))
-	workload.Management.Management.RoleTemplates("").AddHandler(ctx, "cluster-roletemplate-sync", newRTLifecycle(r))
+	management.Management.RoleTemplates("").AddHandler(ctx, "cluster-roletemplate-sync", newRTLifecycle(r))
 }
 
 type manager struct {
```
cbron, line 128: Are we only doing 2.6 and not 2.7 for the test run ?

rmweir: I'm planning on doing both and am partway through forward porting to 2.7. Just waiting on this.

pkg/wrangler/context.go
```diff
@@ -456,6 +456,9 @@ func NewContext(ctx context.Context, clientConfig clientcmd.ClientConfig, restCo
 		core:         core,
 		api:          api,
 		crd:          crd,
+		capi:         capi,
+		rke:          rke,
+		rbac:         rbac,
```
cbron, line 461: what's the point of adding these ?

rmweir: It's a fix, I missed them [here](https://github.com/rancher/rancher/pull/39876/files#diff-a386241816deb9f14c0dbac3c18f0642a47992060fe0f582782dd8be506d2c31R151-R164).

cbron (approved):

cmurphy (approved):

PR #39752, opened 2022-12-01
Corral base use cases in Go Framework

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/357
 
## Problem

we want a way to rehydrate an rke2 cluster - specifically the local cluster running rancher
 
## Solution

In order to modify a local cluster, I added a test case for Corral. This also required modifying the dockerfile to install corral. The test suite currently can only handle running corrals that do not depend on eachother in any way, as the order is psudorandom ([see this issue](https://github.com/rancher/qa-tasks/issues/540) for more details)
In the test case, I include 2 corral packages: one to provision the local cluster (which was largely based on packges from the example https://github.com/rancherlabs/corral-packages repo) and another to modify it, replacing x number of server nodes in the cluster. Please see the updated readmes in their respective folder/package

While adding the test for corral. I also added some new variables when using corral in our golang framework. These are also updated in the readme, so please read it for more details (in standalone v2/validation test suite). IMO, the most important reason I added these is that the `corralConfigUser: <string, default="jenkauto">` user is what is used to name resources created in the backend, which is important for cleanup and correct naming practices we should be updating this to our own shortname (i.e. ctw for me). 
 
## Testing

local and remote testing (through jenkins) is done.  Will link privately.

PR #40044, opened 2023-01-03
go + corral initial test case (v2.6)

slickwarren:
## Issue: 

forward port of: https://github.com/rancher/rancher/pull/39752
https://github.com/rancher/qa-tasks/issues/357
 
## Problem

we want a way to rehydrate an rke2 cluster - specifically the local cluster running rancher
 
## Solution

In order to modify a local cluster, I added a test case for Corral. This also required modifying the dockerfile to install corral. The test suite currently can only handle running corrals that do not depend on eachother in any way, as the order is psudorandom ([see this issue](https://github.com/rancher/qa-tasks/issues/540) for more details)
In the test case, I include 2 corral packages: one to provision the local cluster (which was largely based on packges from the example https://github.com/rancherlabs/corral-packages repo) and another to modify it, replacing x number of server nodes in the cluster. Please see the updated readmes in their respective folder/package

While adding the test for corral. I also added some new variables when using corral in our golang framework. These are also updated in the readme, so please read it for more details (in standalone v2/validation test suite). IMO, the most important reason I added these is that the `corralConfigUser: <string, default="jenkauto">` user is what is used to name resources created in the backend, which is important for cleanup and correct naming practices we should be updating this to our own shortname (i.e. ctw for me). 
 
## Testing

local and remote testing (through jenkins) is done.  Will link privately.

### DIFF
--- a/tests/framework/clients/corral/config.go
+++ b/tests/framework/clients/corral/config.go
@@ -13,11 +13,16 @@ const (
 // CorralConfigs is a struct that for necessary corral config environment variables to build a corral
 type CorralConfigs struct {
 	CorralConfigVars map[string]string `json:"corralConfigVars" yaml:"corralConfigVars"`
+	CorralConfigUser string            `json:"corralConfigUser" yaml:"corralConfigUser" default:"jenkauto"`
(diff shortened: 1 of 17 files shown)


igomez06 (approved):
LGTM

jakefhyde (approved):

PR #40063, opened 2023-01-04
[2.7] Provide Internal & External IP addresses to node agent

HarrisonWAffel:
## Problem
When deploying clusters which have `generate_serving_certificates` enabled for the kubelet service, certificate files will be deployed to the downstream node. Depending on the infrastructure provider, these certificates may be named using the public or private IP address of the node. For example, a DO VM will have its certificate files named using the external IP address, while a VM on AWS will have its certificate files named using the private IP address. This is a problem for the node agent, which now needs to interrogate these certificate files to determine if they need to be refreshed. 
 
## Solution
Provide both the public and private IP addresses to the node agent, then check both possible file locations when performing cert validity checks. 
 
## Testing
+ Create a cluster with `generate_serving_certificate` enabled for the kubelet service. 
+ wait for the node to become available 
+ Find the pod running the node agent, and run the following command: `docker logs <NODE_AGENT_CONTAINER> | head`
+ Look in the logs for a line which starts with `INFO: Environment:`, and fine the `CATTLE_ADDRESS` and `CATTLE_INTERNAL_ADDRESS` values. Ensure they equal the actual external and internal IP address of the node

### Manual Testing
I've done the above

I've also run through the existing test steps for the parent issue / PR, and did not observe any refreshes during restarts of rancher or during set periods of time (i.e. as new plans were requested) 

### Automated Testing
n/a

## QA Testing Considerations
Some providers do not set internal IP addresses, in which case the `CATTLE_INTERNAL_ADDRESS` value will be empty
 
This will not introduce any regressions as this logic is only used when refreshing the kubelet serving certificates


package/run.sh
```diff
@@ -172,6 +172,15 @@ if [ -z "$CATTLE_ADDRESS" ]; then
     CATTLE_ADDRESS=$(ip -o route get 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')
 fi
 
+if [ -z "$CATTLE_CATTLE_INTERNAL_ADDRESS" ]; then
```
jiaqiluo, line 175: an extra `CATTLE_`? 
```suggestion
if [ -z "$CATTLE_INTERNAL_ADDRESS" ]; then
```

cmd/agent/main.go
```diff
@@ -552,7 +573,22 @@ func KubeletCertificateNeedsRegeneration(ipAddress, currentHostname string, cert
 		return false, err
 	}
 
-	return !CertificateIncludesHostname(currentHostname, parsedCert) || CertificateIsExpiring(parsedCert, currentTime) || !CertificateIncludesCurrentIP(ipAddress, parsedCert), nil
+	if !CertificateIncludesHostname(currentHostname, parsedCert) {
+		logrus.Tracef("Certificate does not include current hostname, requesting new certificate")
```
jiaqiluo, line 577: nit 
```suggestion
		logrus.Tracef("certificate does not include current hostname, requesting new certificate")
```

jiaqiluo (changes requested):

cmd/agent/main.go
snasovich: These are log messages and generally don't follow the requirement to start with lowercase letter as required by Go standards for _errors_.

HarrisonWAffel: I've changed it to a lower case

snasovich: I see, other log messages introduced here start with lowercase character. I'm fine either way as soon as it's consistent.
Glancing through the code I see we're not consistent across the codebase with some a mix of messages starting with lowercase and uppercase.

Just wanted to make sure we're not misunderstanding https://github.com/golang/go/wiki/CodeReviewComments#error-strings which specifically states:
> This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

jiaqiluo (approved):

cmd/agent/main.go
```diff
@@ -519,11 +519,17 @@ func reconcileKubelet(ctx context.Context) (bool, error) {
 // delivered by Rancher if the generate_serving_certificate property
 // is set to 'true' for the clusters kubelet service.
 func KubeletNeedsNewCertificate(headers http.Header) error {
-	ipAddress := os.Getenv("CATTLE_ADDRESS")
 	currentHostname := os.Getenv("CATTLE_NODE_NAME")
-	fileSafeIPAddress := strings.ReplaceAll(ipAddress, ".", "-")
 
-	cert, err := tls.LoadX509KeyPair(fmt.Sprintf("/etc/kubernetes/ssl/kube-kubelet-%s.pem", fileSafeIPAddress), fmt.Sprintf("/etc/kubernetes/ssl/kube-kubelet-%s-key.pem", fileSafeIPAddress))
+	// RKE will save the certs on the node using either the public or private IP address, depending on the infrastructure provider.
+	// For example, the certs will be stored using the public IP address for VM's on digital ocean, but will use the private IP
```
jakefhyde, line 525: Tangentially related, do we know why RKE uses the public ip for digital ocean (or why AWS uses the private ip)? I'm assuming private is the correct option.

jakefhyde: DO only does so if it the private ip address option is enabled in UI, so that may also be related.

HarrisonWAffel: I checked and it does seem to be the case that digital ocean clusters only report their external IP address in Rancher. Interestingly, for RKE2 clusters only the internal IP address is reported. I'll look to see if this is an existing issue or not, it seems odd though

Issue #40016, opened 2022-12-28
[BUG] Downstream k3s Node driver clusters fail to provision when etcd and controlplane roles are shared by a nodepool


Josh-Diamond:
**Rancher Server Setup**
- Rancher version: `v2.7-28a84c710edf49a4acedfdc3624d427a36124672-head`
- Installation option (Docker install/Helm Chart): Helm w/ RKE1 local `v1.24.9`
- Proxy/Cert Details: `byo-valid`

**Information about the Cluster**
- Kubernetes version: `v1.25.5+k3s1
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? 
   Downstream Node driver [seen w/ AWS + Linode] k3s w/ etcd + cp sharing roles in a nodepool

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
admin/standard user - [seen w/ both]

**Describe the bug**
Downstream cluster fails to provision when etcd and controlplane roles are shared by a nodepool.

**To Reproduce**
1. Fresh install of rancher `v2.7-head`
2. Provision a downstream k3s cluster w/ the following nodepool configurations, [AWS or Linode]:

| Pool | Role(s) | Quantity |
|---|---|---|
| 1. | etcd/controlplane | 3 |
| 2. | worker | 3 |

 4. Reproduced

**Result**

Cluster stuck in `Updating` hang state, reporting `Configuring bootstrap node(s): waiting for cluster agent to connect`

**Expected Result**
Cluster comes up `active`

**Additional context**

Rancher Logs:
```logs
2022/12/28 20:14:37 [INFO] [planner] rkecluster fleet-default/jkeslar-sowmya-anu: waiting: configuring bootstrap node(s) jkeslar-sowmya-anu-pool1-REDACTED: waiting for cluster agent to connect
2022/12/28 20:14:58 [ERROR] error syncing '_all_': handler user-controllers-controller: failed to start user controllers for cluster REDACTED: ClusterUnavailable 503: cluster not found, requeuing
2022/12/28 20:16:15 [ERROR] Error during subscribe websocket: close sent
```

Also seen on `v2.7.0` with k8s:
- `v1.24.7+k3s1`
- `v1.23.14+k3s1`
- `v1.24.9+k3s1`

Oats87:
@Josh-Diamond any info from the downstream node? Logs from K3s and the system agent?

[...]

PR #40066, opened 2023-01-04
Skip adding etcd taint for k3s

jakefhyde:
## Issue: 

#40016

## Problem

K3s cluster system pods + charts do NOT have the correct tolerations to be able to run on a controlplane + etcd node

## Solution

Do not add the etcd taint to etcd nodes if the runtime is k3s and it is also a controlplane node.

### Manual Testing

Tested manually, confirmed etcd taint was not present for split role clusters. Tested upgrade from v2.7.0 v1.24.8+k3s1.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -853,7 +853,7 @@ func PruneEmpty(config map[string]interface{}) {
 }
 
 // getTaints returns a slice of taints for the machine in question
-func getTaints(entry *planEntry) (result []corev1.Taint, _ error) {
+func getTaints(entry *planEntry, cp *rkev1.RKEControlPlane) (result []corev1.Taint, _ error) {
 	data := entry.Metadata.Annotations[rke2.TaintsAnnotation]
 	if data != "" {
 		if err := json.Unmarshal([]byte(data), &result); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -862,7 +862,8 @@ func getTaints(entry *planEntry) (result []corev1.Taint, _ error) {
 	}
 
 	if !isWorker(entry) {
-		if isEtcd(entry) {
+		// k3s charts do not have correct tolerations when the master node is both controlplane and etcd
+		if isEtcd(entry) && (rke2.GetRuntime(cp.Spec.KubernetesVersion) != rke2.RuntimeK3S || !isControlPlane(entry)) {
```
snasovich, line 866: If all we need from CP here is `cp.Spec.KubernetesVersion` why not pass just it (or even `isK3s` flag - but I like former more anyway) and not the whole object (even though it's a pointer).

snasovich: Right, actually thinking more about it passing `string runtime` looks the cleanest to me and matching what we've had before.
I'm not so sure about the "remove in the next release" if the fix has to be on k3s side considering we will be supporting previous k3s patch versions still.
Probably splitting hairs here, so I'm not going to block this based on whether we should pass the whole CP or just string runtime.

snasovich (approved):

Oats87 (approved):

PR #40076, opened 2023-01-05
[v2.7] Adjust object conflict messages seen in provisioning log

HarrisonWAffel:
## Problem
There are intermittent error messages reported in the provisioning log relating to object version conflicts. In all observed scenarios these error messages are benign and resolve themselves automatically. However, the presence of these error messages and subsequent UI display gives the end user an impression that something has failed, when in fact the issue is only temporary.
 
## Solution
Prepend a message stating that this is a transient error, so as to ensure the end user that the provisioning process has not encountered a significant error that needs to be remediated 

## Testing
+ Create a rancher server 
+ provision an RKE2 cluster on any infrastructure provider. Create 3-5 nodes in a single all role node pool
+ observe that the error message is now prepended with `Transient Error Encountered: ...` 

### Manual Testing
I've done the above

### Automated Testing
n/a

## QA Testing Considerations
reproducing the original error is difficult, i've had some success by limiting the resources available to the Rancher server. 

Using 3-5 nodes seems to have this error appear more frequently. 

I've also seen this error occur when scaling up and down, but even when doing this the error is still not consistent. 
 
This change modifies certain log messages, it has no impact on existing functionality outside of what is shown in the UI.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
+++ b/pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
@@ -97,6 +97,10 @@ func (h *handler) recordMessage(provCluster *provv1.Cluster, cm *corev1.ConfigMa
 		return cm, nil
 	}
 
+	if strings.Contains(msg, "the object has been modified; please apply your changes to the latest version and try again") {
+		msg = fmt.Sprintf("Transient error encountered: %s", msg)
+	}
+
 	last := cm.Data["last"]
 	if msg == last {
 		return cm, nil


pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
```diff
@@ -97,6 +97,11 @@ func (h *handler) recordMessage(provCluster *provv1.Cluster, cm *corev1.ConfigMa
 		return cm, nil
 	}
 
+	// conflict errors are handled by further reconciliation without issue, no reason to report them in the UI.
+	if strings.Contains(msg, "the object has been modified; please apply your changes to the latest version and try again") {
```
Oats87, line 101: in certain cases, this error is benign, but we're working behind the scenes to correct controller behavior to stop racing and potentially causing issues with conflicting objects.

In certain cases though, we do want to display this message as there is likely something seriously wrong if conflicts are happening, and it (while not necessarily end user friendly) provides feedback as to why progress may not be made.

HarrisonWAffel: That makes sense! I think the reason the ticket was opened was because it looked 'scary', and I wasn't aware of an scenario where this would indicate a real issue in the server. It makes more sense to keep it in the UI if we expect it to hault provisioning progress. Would it be safe to say that the root cause of this will likely be solved indirectly by other controller improvements?

HarrisonWAffel:
I've updated this PR to append a clarifying message to the end of these error messages, allowing them to be still be shown in the UI but indicating to the user that they are not significant errors or cause for concern

pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
```diff
@@ -97,6 +97,10 @@ func (h *handler) recordMessage(provCluster *provv1.Cluster, cm *corev1.ConfigMa
 		return cm, nil
 	}
 
+	if strings.Contains(msg, "the object has been modified; please apply your changes to the latest version and try again") {
+		msg = fmt.Sprintf("%s: this error may resolve itself, and unless consistently repeated, does not represent a critical error in the provisioning process", msg)
```
Oats87, line 101: It's difficult to both quantify and tell if this message is being "consistently repeated". Perhaps we can just state that it is a transient error, or format it like `Transient error encountered: %s`

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
HarrisonWAffel: Sure. Wording this scenario was a bit difficult, but I think that wording sounds a lot better

snasovich (approved):
LGTM, thank you @Oats87 @HarrisonWAffel for figuring out a good message format.

@HarrisonWAffel , before merging please update the PR description to explain the updated solution (more user-friendly message vs. skipping message)

jiaqiluo (approved):

Oats87 (approved):

PR #40083, opened 2023-01-06
Refactor planner handler and add capi predicates

jakefhyde:
## Problem

This PR fixes node auto replace and encryption key rotation by changing how the control plane's ready status field is set, and fixing planner thrashing respectively.

### Node Auto Replace

1. Delete a node from an Active cluster with Auto Node Replace set to 30 seconds
2. The planner will attempt to reconcile the node, fail, and set the RKEControlPlane (and Provisioning cluster) to `Updated = False` (to indicate there was an error applying the plan).
3. This is then reconciled onto the management cluster here: [https://github.com/rancher/rancher/blob/da810a7523fb4c1b529bee374e9d2574f33ab0c8/p[…]ntrollers/provisioningv2/rke2/provisioningcluster/controller.go](https://github.com/rancher/rancher/blob/da810a7523fb4c1b529bee374e9d2574f33ab0c8/pkg/controllers/provisioningv2/rke2/provisioningcluster/controller.go#L291)
4. This was then copied back from the management cluster onto the controlplane object https://github.com/rancher/rancher/blob/da810a7523fb4c1b529bee374e9d2574f33ab0c8/pkg/controllers/provisioningv2/rke2/provisioningcluster/controller.go#L291
5. When this happens, we set initialized and ready on control plane status to false
6. This was being copied to the infrastructure cluster object here: https://github.com/rancher/rancher/pull/39875/files#r1047940430
These two snippets from the CAPI MachineHealthCheck controller prevent machine health checks from running if either the infrastructure cluster or control plane are not ready/initialized respectively https://github.com/kubernetes-sigs/cluster-api/blob/ccba46ec29e10d8e7c967f242a0a1970dd8dce64/internal/controllers/machinehealthcheck/machinehealthcheck_targets.go#L114-L126

### Node Auto Replace

-- | -- | -- | --

### Encryption Key Rotation

Fix planner thrashing caused when updating an object, discarding it in the cache, updating an objects status again (causing thrashing), and returning that object instead. Now, the signature of all planner related functions that are expected to update the controlplane return the status object and an error, where if the error is of type `ErrWaiting`, the error message is transformed onto the appropriate condition and propagated up to the provisioning cluster.

### Manual Testing

Encryption key rotation was tested manually for a freshly provisioned cluster and works as expected.

ETCD snapshot/restore was also tested, as was certificate rotation, and all performed as expected.

### Automated Testing

Encryption key rotation was also tested with validation tests, and worked as expected for 10k secrets.

## QA Testing Considerations

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/planner/controller.go
+++ b/pkg/controllers/provisioningv2/rke2/planner/controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/rancher/wrangler/pkg/relatedresource"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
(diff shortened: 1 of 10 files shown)


annablender:
@jakefhyde Could you fill out the PR template? Has this been smoke tested?

pkg/controllers/provisioningv2/rke2/planner/controller.go
```diff
@@ -51,37 +52,56 @@ func Register(ctx context.Context, clients *wrangler.Context, planner *planner.P
...
-			// If you are reading this years later we'll just assume we decided not to change the framework.
-			logrus.Errorf("[planner] rkecluster %s/%s: error encountered during plan processing was %v", cluster.Namespace, cluster.Name, err)
-			h.controlPlanes.EnqueueAfter(cluster.Namespace, cluster.Name, 5*time.Second)
+
+	logrus.Debugf("[planner] rkecluster %s/%s: calling planner process", cp.Namespace, cp.Name)
+	status, err := h.planner.Process(cp, status)
+	if err != nil {
+		if planner.IsErrWaiting(err) {
+			logrus.Infof("[planner] rkecluster %s/%s: waiting: %v", cp.Namespace, cp.Name, err)
+			rke2.Reconciled.Message(&status, "reconciling control plane")
```
Oats87, line 74: technically, if we're waiting we could be reconciling the worker nodes as well. this will lead to user confusion imo

pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -29,8 +28,6 @@ const (
 	encryptionKeyRotationSecretsEncryptApplyCommand  = "secrets-encrypt-apply"
 	encryptionKeyRotationSecretsEncryptStatusCommand = "secrets-encrypt-status"
 
-	encryptionKeyRotationLeaderAnnotation = "rke.cattle.io/encrypt-key-rotation-leader"
```
Oats87, line 32: what happens if a user updates from an annotation version of Rancher to a status version of Rancher in the middle of an encryption key rotation?

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -210,179 +207,191 @@ func (p *Planner) setMachineConditionStatus(clusterPlan *plan.Plan, machineNames
...
+
+func atMostThree(names []string) string {
+	sort.Strings(names)
+	if len(names) > 3 {
+		return fmt.Sprintf("%s and %d more", strings.Join(names[:3], ","), len(names)-3)
+	}
+	return strings.Join(names, ",")
+}
+
+func detailMessage(machines []string, messages map[string][]string) string {
```
Oats87, line 365: sp: detail -> detailed?

Oats87 (changes requested):
Overall LGTM; however I have a few functionality concerns (Around upgrades)

HarrisonWAffel (approved):

Oats87 (approved):

PR #40090, opened 2023-01-06
fixing PSP integration tests for 1.25+

slickwarren:
## Issue: 

builds have been failing due to some PSP tests running on the latest (1.25+ does not support PSPs)
 

## Solution

adding a condition to pytest to check the k8s version of the cluster. If the cluster is 1.25 or above, we will not run the tests marked as including PSPs. 
 
## Testing

testing is part of integration -> drone

## QA Testing Considerations

 we should add PSA tests for 1.25+ in the new (golang) integration tests.

### DIFF
--- a/tests/integration/suite/test_cluster_defaults.py
+++ b/tests/integration/suite/test_cluster_defaults.py
@@ -5,6 +5,20 @@ from .common import random_str
 from .conftest import wait_for
 
 
+@pytest.fixture(scope='module')
+def check_cluster_kubernetes_version(admin_mc):
+    """
+       Checks the local cluster's k8s version
+    """
+    client = admin_mc.client
+    cluster = client.by_id_cluster("local")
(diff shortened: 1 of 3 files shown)


jakefhyde (approved):

jiaqiluo (approved):
LGTM, except that the same function is re-defined at multiple files. 
I am not very familiar with the code, but I guess some limitation in the integration framework prevents us from reusing the fixture function.   Other than that, the changes look good to me.

igomez06 (approved):
LGTM

slickwarren:
> LGTM, except that the same function is re-defined at multiple files. I am not very familiar with the code, but I guess some limitation in the integration framework prevents us from reusing the fixture function. Other than that, the changes look good to me.

that's exactly correct, the fixture function(s) can't be imported to other files easily.  So this was the simplest way I could think of at the moment.

Issue #39232, opened 2022-10-05
[BUG] Broken Rancher when migrating from 2.6.6 to 2.6.8


all4innov:
**Rancher Server Setup**
- Rancher version: 2.6.6 to be updated to 2.6.8
- Installation option : Helm chart installation on top of AWS EKS
- Proxy/Cert Details: no proxy / Let's Encrypt Cert

**Information about the Cluster**
- Kubernetes version:  AWS EKS 1.20
- Cluster Type :  downstream imported

**User Information**
- role of the user logged in: rancher admin

**Describe the bug**

When upgrading rancher via helm from version 2.6.6 to 2.6, Rancher is no more working correctly and rancher pod logs contains go panic bug and keep restarting indefenitly. In addition all downstream clusters are no more manageable via rancher.

NB. same problem when migrating from 2.6.6 to 2.6.7.

Here the panic bug from rancher container logs:

```
[INFO] kontainerdriver googlekubernetesengine stopped
[INFO] dynamic schema for kontainerdriver googlekubernetesengine updating
panic: indexer conflict: map[byPod:{}]
goroutine 1932 [running]:
k8s.io/apimachinery/pkg/util/runtime.Must(...)
/go/pkg/mod/k8s.io/apimachinery@v0.24.2/pkg/util/runtime/runtime.go:172
github.com/rancher/rancher/pkg/generated/controllers/catalog.cattle.io/v1.(*operationCache).AddIndexer(0xc018366330, {0x4237b85, 0x5}, 0x44fdfd0)
/go/src/github.com/rancher/rancher/pkg/generated/controllers/catalog.cattle.io/v1/operation.go:247 +0xe5
github.com/rancher/rancher/pkg/controllers/dashboard/helm.RegisterOperations({0x4b80210, 0xc00a86adc0}, {0x4c4e208, 0xc0001a5380}, {0x4c1ead8, 0xc008a1a380}, {0x4c1c2d0, 0xc008a1a3f0})
/go/src/github.com/rancher/rancher/pkg/controllers/dashboard/helm/operations.go:42 +0x1cc
github.com/rancher/rancher/pkg/controllers/dashboard/helm.Register({0x4b80210, 0xc00a86adc0}, 0xc000b4b860)
/go/src/github.com/rancher/rancher/pkg/controllers/dashboard/helm/register.go:22 +0x2ce
[...]

PR #40118, opened 2023-01-10
[DNM investigation needed] Fix panic when SA token secret is empty for eks cluster

alexander-demicev:
## Issue: 

Issue??

Someone apparently saw this here https://github.com/rancher/rancher/issues/41299#issuecomment-1523015726
 
## Problem

 
If an empty secret is returned then attempting to set `cluster.Status.ServiceAccountTokenSecret` will cause panic

## Solution

I added a check for nil pointer that will likely not solve the root cause of the problem but the controller won't panic and return some meaningful error and include debug logs.

### DIFF
--- a/pkg/controllers/management/aks/aks_cluster_handler.go
+++ b/pkg/controllers/management/aks/aks_cluster_handler.go
@@ -193,6 +193,10 @@ func (e *aksOperatorController) onClusterChange(_ string, cluster *apimgmtv3.Clu
 					if err != nil {
 						return cluster, err
 					}
+					if secret == nil {
+						logrus.Debugf("Empty service account token secret returned for cluster [%s]", cluster.Name)
+						return cluster, fmt.Errorf("failed to create or update service account token secret, secret can't be empty")
+					}
 					cluster.Status.ServiceAccountTokenSecret = secret.Name
 					cluster.Status.ServiceAccountToken = ""
 				}
--- a/pkg/controllers/management/eks/eks_cluster_handler.go
+++ b/pkg/controllers/management/eks/eks_cluster_handler.go
@@ -257,6 +257,10 @@ func (e *eksOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 					if err != nil {
 						return cluster, err
 					}
+					if secret == nil {
+						logrus.Debugf("Empty service account token secret returned for cluster [%s]", cluster.Name)
+						return cluster, fmt.Errorf("failed to create or update service account token secret, secret can't be empty")
+					}
 					cluster.Status.ServiceAccountTokenSecret = secret.Name
 					cluster.Status.ServiceAccountToken = ""
 				}
--- a/pkg/controllers/management/gke/gke_cluster_handler.go
+++ b/pkg/controllers/management/gke/gke_cluster_handler.go
@@ -203,6 +203,10 @@ func (e *gkeOperatorController) onClusterChange(key string, cluster *mgmtv3.Clus
 					if err != nil {
 						return cluster, err
 					}
+					if secret == nil {
+						logrus.Debugf("Empty service account token secret returned for cluster [%s]", cluster.Name)
+						return cluster, fmt.Errorf("failed to create or update service account token secret, secret can't be empty")
+					}
 					cluster.Status.ServiceAccountTokenSecret = secret.Name
 					cluster.Status.ServiceAccountToken = ""
 				}


mjura (approved):
LGTM

a-blender (changes requested):
Holding off on approving this PR for now. See comment https://github.com/rancher/rancher/issues/39232#issuecomment-1453701729

alexander-demicev:
@a-blender I totally agree about this issue potentially being serious. We can still merge the PR to fix the panic and investigate further.

a-blender (approved):

snasovich (commented):
One minor nit regarding logging, but overall couple of concerns:
1. The original [issue](https://github.com/rancher/rancher/issues/39232) refers to panic in `panic: indexer conflict: map[byPod:{}]` - so not sure if it's the same issue
2. What will happen when this error is returned? Will the secret migration be retried or what happens?

snasovich:
Added @jakefhyde and @cmurphy to hopefully look into it as well since it has to do with secret migration code.

cmurphy (commented):
If this is a problem for EKS, then it is also potentially a problem for GKE and AKS and should be addressed there as well:

https://github.com/rancher/rancher/blob/85315a04c4975ff8aeea497d97686f24275be791/pkg/controllers/management/gke/gke_cluster_handler.go#L202-L206
https://github.com/rancher/rancher/blob/64c5188a5394f7ef7858ebb6807072ad5abe0e80/pkg/controllers/management/aks/aks_cluster_handler.go#L192-L196

fyi here is the code that returns the secret, in theory there should be no way for the secret and err to both be nil if serviceToken is not nil

https://github.com/rancher/rancher/blob/64c5188a5394f7ef7858ebb6807072ad5abe0e80/pkg/controllers/management/secretmigrator/clusters.go#L236-L254

Was this actually reproduced and validated that this fixes the issue? I agree with @a-blender that it looks like a symptom of a larger issue, and if the service account token secret can't be set on the cluster then the cluster will still be broken, panic or not.

It looks like the issue really covers two separate issues, this doesn't address the issue in the main description with error `panic: indexer conflict: map[byPod:{}]`.

alexander-demicev:
@cmurphy I wasn't able to reproduce this issue

jakefhyde (approved):

a-blender:
@alexander-demicev Like I said before, I think this requires a bit more investigation before merging the PR.

From what I can see, the PR put in is trying to fix the issue https://github.com/rancher/rancher/issues/41299#issuecomment-1523015726 because from the log messages, because the panic is occurring here https://github.com/rancher/rancher/blob/f650ce839db0e2fb755dfff1ce28b005a5f1c0e3/pkg/controllers/management/eks/eks_cluster_handler.go#L259 in the EKS handler but that was a comment made on this issue that does not seem to be related to the original issue.

So basically, this fix is linked to the wrong issue.

I am leaning towards merging this anyway to prevent future panics, but I also recall @jakefhyde and I discussing a similar fix last year so I would try to reproduce this on v2.6-head. If it's not reproducible, this may not be needed and the customer can just upgrade.

alexander-demicev:
@a-blender I totally agree with you. Unfortunately, I can't reproduce this issue and no users are complaining for some time.

alexander-demicev:
@a-blender this one https://github.com/rancher/rancher/issues/41299#issuecomment-1523015726

a-blender:
@alexander-demicev Ok so the EKS cluster handler bug on v2.6.6 -> upgrade to v2.6.8. I'd recommend the following steps for this situation based on the Snow Cow process (doc on Confluence)

* Could you find out if there's already GH issue for that? If so, link it to that issue. If not, create a new issue for it
* Try to repro the issue on v2.6-head
* If you can't, please state that it was fixed somewhere in between v2.6.6->v2.6.8 upgrade and upgrade to v2.6-head. Also try on the latest v2.7-head
* If you can repro on v2.7-head, we can merge this PR.
* If you can also repro on v2.6-head, we need to backport it as well.

This issue https://github.com/rancher/rancher/issues/41299#issuecomment-1453701729 is separate and needs to be investigated separately.

jakefhyde:
I've approved the PR, but a little bit of context behind this is that it was always considered a bug in my opinion if the secret was nil, or an error was not returned. At that point, we are either waiting for the service account token to be generated, or there was an actual problem generating the the token. There weren't really any active maintainers of the hosted providers at the time I wrote that logic, but my intention was to create it such that we should never be without either the corresponding service account token secret, or an error. In hindsight letting it panic probably wasn't the best solution, but realistically that is the direction the investigation should head towards, since the service account token secret never being populated leads to an unmanageable downstream cluster.

a-blender (approved):

furkatgofurov7 (commented):
LGTM

mjura (approved):
LGTM

PR #40132, opened 2023-01-11
Moved control plane ready status to planner

jakefhyde:
## Problem

This PR fixes node auto replace by changing how the control plane's ready status field is set.

### Node Auto Replace

-- | -- | -- | --

### Manual Testing

## QA Testing Considerations

### DIFF
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -370,6 +370,17 @@ func (p *Planner) Process(controlPlane *rkev1.RKEControlPlane) error {
 		return ErrWaiting("waiting for control plane to be available")
 	}
 
+	if controlPlane.Status.Initialized != true || controlPlane.Status.Ready != true {
+		controlPlane = controlPlane.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -370,6 +370,17 @@ func (p *Planner) Process(controlPlane *rkev1.RKEControlPlane) error {
...
 
+	if controlPlane.Status.Initialized != true || controlPlane.Status.Ready != true {
+		controlPlane = controlPlane.DeepCopy()
+		controlPlane.Status.Initialized = true
+		controlPlane.Status.Ready = true
+		_, err := p.rkeControlPlanes.UpdateStatus(controlPlane)
+		if err != nil {
+			return err
+		}
+		return ErrWaiting("marking control plane initialized and ready")
```
Oats87, line 381: We need to return the updated status of the `rkecontrolplane` so that we can update the cache

Oats87 (changes requested):

jakefhyde:
Closing in favor of https://github.com/rancher/rancher/pull/40083

PR #40146, opened 2023-01-12
Restart cluster controllers when remote cluster's major or minor version changes

maxsokolovsky:
## Problem
If a remote cluster is upgraded to a different K8s version, the controllers for the cluster are not restarted. This might make informers watch over resources that are no longer known by the remote cluster's API server. Because of that, there might be lots of informer errors in the local cluster's logs.
 
## Solution
Restart all cluster controllers in the local cluster for the downstream cluster when its K8s version has been upgraded or downgraded to a non-patch version. The controller now stores the current cluster's K8s version as an annotation and compares it to the new version.
 
### Manual Testing
Created a K8s v1.24 remote cluster with PSPs enabled. Then upgraded the cluster to v1.25, saw that there were no errors about PSPs being unknown by the remote cluster.

### DIFF
--- a/pkg/controllers/managementapi/usercontrollers/usercontroller.go
+++ b/pkg/controllers/managementapi/usercontrollers/usercontroller.go
@@ -2,12 +2,14 @@ package usercontrollers
 
 import (
 	"context"
+	"fmt"
 	"hash/crc32"
 	"math"
 	"sort"
 	"sync"
 	"time"
 
+	mVersion "github.com/mcuadros/go-version"
 	v33 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -27,6 +29,8 @@ import (
 var (
 	all = "_all_"
 )
+
+const controllersRestartedAnnotation = "mgmt.cattle.io/controllers-restarted"
 
 func Register(ctx context.Context, scaledContext *config.ScaledContext, clusterManager *clustermanager.Manager) {
 	u := &userControllersController{
@@ -94,8 +98,29 @@ func (u *userControllersController) sync(key string, cluster *v3.Cluster) (runti
 	if key == all {
 		return nil, u.setPeers(nil)
 	}
+
+	//TODO: Check if the cluster has been upgraded in a generic way and always restart on any version upgrade
+	if cluster != nil && cluster.Status.Version != nil && mVersion.Compare(cluster.Status.Version.String(), "v1.25", ">=") && !u.controllersRestarted(cluster) {
+		u.manager.Stop(cluster)
+		err := u.manager.Start(u.ctx, cluster, u.amOwner(u.peers, cluster))
+		if err != nil {
+			return nil, fmt.Errorf("unable to restart controllers for cluster %s: %w", cluster.Name, err)
+		}
+		cluster = cluster.DeepCopy()
+		cluster.Annotations[controllersRestartedAnnotation] = "true"
+		cluster, err = u.clusters.Update(cluster)
+		if err != nil {
+			return nil, fmt.Errorf("unable to update cluster with annotation indicating cluster controllers were restarted: %w", err)
+		}
+	}
+
 	u.clusters.Controller().Enqueue("", all)
 	return cluster, nil
+}
+
+func (u *userControllersController) controllersRestarted(cluster *v3.Cluster) bool {
+	value, ok := cluster.Annotations[controllersRestartedAnnotation]
+	return ok && value == "true"
 }
 
 func (u *userControllersController) setPeers(peers *tpeermanager.Peers) error {
--- a/pkg/controllers/managementapi/usercontrollers/usercontroller.go
+++ b/pkg/controllers/managementapi/usercontrollers/usercontroller.go
@@ -9,13 +9,9 @@ import (
 	"sync"
 	"time"
 
-	mVersion "github.com/mcuadros/go-version"
-	v33 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-
-	"k8s.io/apimachinery/pkg/runtime"
-
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


MbolotSuse (changes requested):
Some small concerns around naming/comments, but this is overall pretty solid, great job re-working this to be a more generic version.

Generic comment: Do we want to add tests to this? I think this would be hard in the current state because clusterManager.Manager is a struct and not an interface, but I think that if you implemented a `clusterControllerRuner` interface in this file that wrapped the start/stop methods it would be pretty easy to write a test that verifies if/when we restart the controllers.

pkg/controllers/managementapi/usercontrollers/usercontroller.go
```diff
@@ -94,10 +97,63 @@ func (u *userControllersController) sync(key string, cluster *v3.Cluster) (runti
...
+	cluster = cluster.DeepCopy()
+	cluster.Annotations[currentClusterVersion] = value
+	cluster, err := u.clusters.Update(cluster)
+	if err != nil {
+		return fmt.Errorf("unable to update cluster %s with annotation indicating its K8s version: %w", cluster.Name, err)
+	}
+	return nil
+}
+
+func getCurrentClusterVersion(cluster *v3.Cluster) (*version.Version, error) {
```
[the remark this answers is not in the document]
maxsokolovsky: Do you mean we don't save the cluster version and the controllers version?

pkg/controllers/managementapi/usercontrollers/usercontroller.go
```diff
@@ -22,12 +20,17 @@ import (
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/version"
 )
 
 var (
 	all = "_all_"
 )
 
+// currentClusterControllersVersion is the version of the controllers that are run in the local cluster for a particular downstream cluster.
+const currentClusterControllersVersion = "mgmt.cattle.io/current-cluster-controllers-version"
```
MbolotSuse, line 32: Nit: You may want to change the prefix ('mgmt.cattle.io') to something else. I choose that in haste, I don't think it's the right long term name.

pkg/controllers/managementapi/usercontrollers/usercontroller_test.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+		Status: apimgmtv3.ClusterStatus{
+			Version: &version.Info{GitVersion: "1.23.4"},
+		},
+	}
+
+	obj, err := controller.sync("", cluster)
+	require.NoError(t, err)
+	require.NotNil(t, obj)
+	assert.NotEmpty(t, obj.(*v3.Cluster).Annotations[currentClusterControllersVersion])
```
MbolotSuse, line 64: Nit: You may want to assert that it's set to a specific value, especially since you do so for other tests.

pkg/controllers/managementapi/usercontrollers/usercontroller.go
```diff
@@ -94,10 +102,65 @@ func (u *userControllersController) sync(key string, cluster *v3.Cluster) (runti
...
+		if err != nil {
+			if cluster, err = u.saveClusterControllersVersionAnnotation(cluster, cluster.Status.Version.String()); err != nil {
+				return nil, err
+			}
+			return cluster, nil
+		}
+
+		newVersion, err := version.ParseSemantic(cluster.Status.Version.String())
+		if err != nil {
+			return cluster, nil
```
MbolotSuse, line 120: I don't think ignoring the error here is the right move. If we can't parse the version of the cluster, we won't restart the controller. This could result in us failing to restart on certain clusters where we need to (i.e. k3s and RKE/RKE2 have odd version semantics).

Since these are called on any update of the cluster and this won't just fix itself, I think logging the issue with a level of error would be a good idea.

pkg/controllers/managementapi/usercontrollers/usercontroller_test.go
```diff
@@ -0,0 +1,139 @@
+package usercontrollers
```
MbolotSuse, line 1: Overall test nit:

You have several different function for the different test cases. You could probably make these into one function and just run through the different cases. Up to you though.

MbolotSuse: I would also recommend adding some error cases (particularly for cases like I can't update the annotation or I can't stop the cluster).

MbolotSuse (changes requested):
Couple of small concerns on error cases and naming (error logging is most important). 

I would also recommend validating the core approach (stop/start after change) with a member of team 2 before we merge.

pkg/controllers/managementapi/usercontrollers/usercontroller.go
```diff
@@ -73,10 +76,15 @@ func Register(ctx context.Context, scaledContext *config.ScaledContext, clusterM
...
+type controllerStarter interface {
+	Start(ctx context.Context, cluster *v3.Cluster, clusterOwner bool) error
+	Stop(cluster *v3.Cluster)
+}
+
 type userControllersController struct {
 	sync.Mutex
 	clustered     bool
-	manager       *clustermanager.Manager
+	manager       controllerStarter
```
MbolotSuse, line 87: Nit: This now has a type of controllerStarter but you are still calling the variable by the name of the struct which implements the interface. This should probably be a `starter`, and  comment should be left above the interface to note:
- What the interface is for
- That clustermanager.Manager implements the interface (optional - this might be nice to have but could technically be a bad practice).

MbolotSuse (approved):
Code LGTM. Please include test cases for a variety of different types of clusters as part of the validation template for this issue.

rmweir (approved):
This is a clean solution considering what it does. Good job!

MbolotSuse (approved):
LGTM

rmweir (approved):

PR #40157, opened 2023-01-13
Adding automation script to validate PSA rbac

anupama2501:
## Issue: 
https://github.com/rancher/qa-tasks/issues/543

## Problem
In this PR we are covering the RBAC cases for the new feature PSA.  
## Solution
PodSecurityAdmission is a new feature in k8s where we allow cluster owners/admins to configure a few labels to capture pod security. Other members such as Project owner/cluster member/Project members/Project read-only permissions should not be able to configure the labels at namespace level. This only fixes the namespace level permissions and not project level permissions.


tests/v2/validation/rbac/rbac.go
```diff
@@ -84,3 +100,61 @@ func createProject(client *rancher.Client, clusterID string) (createProject *man
...
+
+	deploymentResp, err := steveclient.SteveType(workloads.DeploymentSteveType).Create(deployment)
+	if err != nil {
+		return nil, err
+	}
+	err = kwait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+		steveclient, err := client.Steve.ProxyDownstream(clusterID)
+		if err != nil {
+			return false, nil
+		}
```
caliskanugur, line 134: Doesn't this function already use the client argument as proxied to the downstream?

In my opinion, either this part needs a removal or proxied steve client instantiation could happen as the first thing in this helper function after L116.

tests/v2/validation/rbac/rbac.go
```diff
@@ -84,3 +100,61 @@ func createProject(client *rancher.Client, clusterID string) (createProject *man
...
+		for _, statusCondition := range status {
+			if strings.Contains(statusCondition.Message, "forbidden") {
+				err = errors.New(statusCondition.Message)
+				return false, err
+			}
+		}
+
+		if *deployment.Spec.Replicas == deployment.Status.AvailableReplicas {
+			return true, nil
+		}
```
caliskanugur, line 154: Another nit, but it seems like this function is not just a create but `createAndWait` or something like that (please name as you want, don't have a strong opinion), but I believe renaming according to the logic can make the helper function usage clearer

tests/v2/validation/rbac/rbac_psa_test.go
```diff
@@ -0,0 +1,334 @@
...
+	steveNonAdminClient *v1.Client
+	adminNamespace      *v1.SteveAPIObject
+}
+
+func (rb *PSATestSuite) TearDownSuite() {
+	rb.session.Cleanup()
+}
+
+func (rb *PSATestSuite) SetupSuite() {
+	testSession := session.NewSession(rb.T())
```
caliskanugur, line 42: The PR needs a rebase, there have been some changes to this.

tests/v2/validation/rbac/rbac_psa_test.go
```diff
@@ -0,0 +1,334 @@
...
+	case roleMember, roleProjectOwner, roleProjectMember, roleProjectReadOnly:
+		require.Error(rb.T(), err)
+		if role == roleMember || role == roleProjectReadOnly {
+			errMessage := strings.Split(err.Error(), ":")[0]
+			assert.Equal(rb.T(), "Resource type [namespace] is not updatable", errMessage)
+		} else {
+			errStatus := strings.Split(err.Error(), ".")[1]
+			rgx := regexp.MustCompile(`\[(.*?)\]`)
+			errorMsg := rgx.FindStringSubmatch(errStatus)
+			assert.Equal(rb.T(), "403 Forbidden", errorMsg[1])
```
caliskanugur, line 99: Another nit, probably doesn't need an update, but instead this regexp, `assert. Contains` is also another option here.

tests/v2/validation/rbac/rbac_psa_test.go
```diff
@@ -0,0 +1,334 @@
...
+	deleteLabelsNS := rb.ConvertNS(rb.adminNamespace)
+	deleteLabelsNS.Labels = labels
+
+	_, err = rb.steveNonAdminClient.SteveType(namespaces.NamespaceSteveType).Update(rb.adminNamespace, deleteLabelsNS)
+	switch role {
+	case restrictedAdmin, roleOwner:
+		require.NoError(rb.T(), err)
+		expectedLabels := getPSALabels(response, labels)
+		assert.Equal(rb.T(), 0, len(expectedLabels))
+		_, err = createDeployment(rb.steveNonAdminClient, rb.client, rb.cluster.ID, "nginx", "nginx", rb.adminNamespace.Name)
```
caliskanugur, line 120: I think it might be better to store these images and container names as const strings in `yourtest.go` might be useful for future updates. Below has the same update applicable lines.

tests/v2/validation/rbac/rbac_psa_test.go
```diff
@@ -0,0 +1,334 @@
...
+	response, err := rb.steveNonAdminClient.SteveType(namespaces.NamespaceSteveType).Update(rb.adminNamespace, updateNS)
+
+	switch role {
+	case restrictedAdmin, roleOwner:
+		require.NoError(rb.T(), err)
+		expectedLabels := getPSALabels(response, labels)
+		assert.Equal(rb.T(), labels, expectedLabels)
+	case roleMember, roleProjectOwner, roleProjectMember, roleProjectReadOnly:
+		require.Error(rb.T(), err)
+		if role == roleMember || role == roleProjectReadOnly {
```
caliskanugur, line 92: I don't understand this condition in the switch/case, I might be missing something, but wouldn't it be more clear if there was another case for `roleMember, roleProjectReadonly `. If this makes it clearer, below has the same update applicable lines.

tests/v2/validation/rbac/rbac_test.go
```diff
@@ -344,7 +344,7 @@ func (rb *RBTestSuite) TestRBAC() {
 	}
 	for _, tt := range tests {
 		rb.Run("Set up User with Cluster Role "+tt.name, func() {
-			newUser, err := createUser(rb.client)
+			newUser, err := createUser(rb.client, "user")
```
caliskanugur, line 347: I think adding them hard coded is not future-proof like I mentioned above for the deployment, either a global variable or function-scoped variable might be better for future updates for these values.

caliskanugur (commented):
I've got a few nits and questions, but it looks good

tests/v2/validation/rbac/rbac.go
```diff
@@ -10,15 +14,27 @@ import (
...
-
-func createUser(client *rancher.Client) (*management.User, error) {
+const roleProjectReadOnly = "read-only"
+const restrictedAdmin = "restricted-admin"
+const pssRestrictedPolicy = "restricted"
+const pssBaselinePolicy = "baseline"
+const pssPrivilegedPolicy = "privileged"
+const psaWarn = "pod-security.kubernetes.io/warn"
+const psaAudit = "pod-security.kubernetes.io/audit"
+const psaEnforce = "pod-security.kubernetes.io/enforce"
```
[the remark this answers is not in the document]
markusewalker: I agree, it would better to have it in that format of `const (....)`

tests/v2/validation/rbac/rbac.go
```diff
@@ -84,3 +100,61 @@ func createProject(client *rancher.Client, clusterID string) (createProject *man
...
+		if _, found := actualLabels[label]; found {
+			expectedLabels[label] = actualLabels[label]
+		}
+	}
+	return expectedLabels
+
+}
+
+func createDeployment(steveclient *v1.Client, client *rancher.Client, clusterID string, containerName string, image string, namespaceName string) (*v1.SteveAPIObject, error) {
+
```
markusewalker, line 117: just a nit - there's a space here we can remove.

tests/v2/validation/rbac/rbac_psa_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	namespaceObj := &coreV1.Namespace{}
+	err = v1.ConvertToK8sType(getNSSteveObject.JSONResp, namespaceObj)
+	require.NoError(rb.T(), err)
+
+	return namespaceObj
+}
+
+func (rb *PSATestSuite) ValidatePSA(role string) {
+
```
markusewalker, line 70: nit - please remove this extra space 😄

tests/v2/validation/rbac/rbac_psa_test.go
markusewalker: To add on, maybe instead of a switch statement, this is abstracted elsewhere? We may need to make some revisions in the future and that may make it easier to do so instead of adding on here.

tests/v2/validation/rbac/rbac_psa_test.go
```diff
@@ -0,0 +1,334 @@
...
+			rgx := regexp.MustCompile(`\[(.*?)\]`)
+			errorMsg := rgx.FindStringSubmatch(errStatus)
+			assert.Equal(rb.T(), "403 Forbidden", errorMsg[1])
+		}
+	}
+
+}
+
+func (rb *PSATestSuite) ValidateAdditionalPSA(role string) {
+
```
markusewalker, line 168: same nit as above - please remove the extra space

anupama2501:
Closing this PR and creating a new one as I accidentally merged release/v2.7 into my private branch.

tests/v2/validation/rbac/rbac.go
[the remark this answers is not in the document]
anupama2501: Resolving comments here as to make sure these are addressed.

Updated. Yes, it makes sense to declare constants this way.

tests/v2/validation/rbac/rbac.go
anupama2501: yes, good point. Removed the statement where I instantiate steveclient

tests/v2/validation/rbac/rbac_psa_test.go
anupama2501: Yes declared them as const strings in the rbac_psa_test.go

tests/v2/validation/rbac/rbac_test.go
anupama2501: Updated. Added the user as a constant in rbac.go file.

tests/v2/validation/rbac/rbac_psa_test.go
markusewalker: no worries, yeah if it's going to take more work, then let's have that be an enhancement for later.

PR #40177, opened 2023-01-16
 [v2.6] Bump fleet to 0.5.2 rc.1

thardeck:


KevinJoiner:
@thardeck/@mattfarina, What issues are we backporting to v2.6 that this fleet bump solves? I only ask because this would be a minor version bump, and we usually try not to backport new features if they are not needed.

KevinJoiner:
@thardeck Cool, sounds good thanks for putting that together 👍

go.mod
```diff
@@ -107,7 +107,7 @@ require (
 	github.com/rancher/channelserver v0.5.1-0.20220405170618-28c9b37deff1
 	github.com/rancher/dynamiclistener v0.3.4
 	github.com/rancher/eks-operator v1.1.5
-	github.com/rancher/fleet/pkg/apis v0.0.0-20220916041858-eae202f68985
+	github.com/rancher/fleet/pkg/apis v0.0.0-20230116142544-82bfc275d117
```
KevinJoiner, line 110: Should this instead point to the v0.5.2-rc1 ([fc276f5](https://github.com/rancher/fleet/commit/fc276f5505bec0be0da9f6560443495586de566e)) commit instead of the master branch?

KevinJoiner (changes requested):

go.mod
KevinJoiner: The commit hash is fine. This error happens because we have a module (aka a go.mod file) inside another module. Which is not how modules were designed to be used.

KevinJoiner (approved):

manno (approved):

mattfarina (approved):

package/Dockerfile
```diff
@@ -53,7 +53,7 @@ ENV CATTLE_SYSTEM_AGENT_UPGRADE_IMAGE rancher/system-agent:${CATTLE_SYSTEM_AGENT
 ENV CATTLE_SYSTEM_UPGRADE_CONTROLLER_CHART_VERSION 100.0.3+up0.3.3
 
 # System charts minimal version
-ENV CATTLE_FLEET_MIN_VERSION=100.1.0+up0.4.0
+ENV CATTLE_FLEET_MIN_VERSION=100.1.0+up0.5.2-rc.1
```
mattfarina, line 56: `100.1.0+up0.5.2-rc.1` is only updating the build metadata. It's didn't update the version itself which is 100.1.0 before and after the update. The chart version needs to be incremented.

Isn't the version supposed to be [100.2.2+up0.5.2-rc.1](https://github.com/rancher/charts/tree/dev-v2.6/charts/fleet/100.2.2%2Bup0.5.2-rc.1)?

mattfarina (changes requested):

manno:
Note, `100.2.2+up0.5.2-rc.1` is a valid semver, however the `-rc.1` will not be recognized as a pre-release information, but treated as build info instead.

The order is pre-release before build, not the other way around: `<valid semver> ::= <version core> "-" <pre-release> "+" <build>`.

Now, `<build> ::= <dot-separated build identifiers>`, so this version contains the following build identifiers:
 * up0
 * 5
 * 2-rc
 * 1

manno:
Since we now have a `0.5.1` branch for the security fix. Should be bump this to `0.5.10-rc.1`, just in case we ever have to do a security fix to 0.5.1?

mattfarina:
We do security fixes in patch releases of Rancher. The only time we would need to leave room for security fixes like that would be if we have two or more minor releases of Rancher using the same minor release of Fleet when we keep updating Fleet at patch levels in some minor releases and not others that may still get security releases.

For example,
- Rancher 2.5 using Fleet up to 0.3.9. Then stopping updating Fleet but still in the security fix zone.
- Rancher 2.6 using Fleet up to 0.3.11 before going to 0.4.0. 0.3.10 introduced new features (including support for new versions of k8s).

This kind of situation would create a problem when Fleet needs a security release in 2.5. But, this is really born out of not following semver.

I think 2 things can really help this situation:
- Any time there is a new feature, increment the minor version. This is already happening.
- When a new version of k8s is supported treat that as a new feature and increment the minor version.

If these two things are done I don't see an issue coming up where we need to have space between patch versions for a future security release.

mattfarina (approved):

manno (approved):

KevinJoiner (approved):

PR #40190, opened 2023-01-17
Add startup probe

rmweir:
## Issue: 

https://github.com/rancher/rancher/issues/39841

## Problem

 Break out of https://github.com/rancher/rancher/issues/38177.

## Solution

 This gives another way of further delaying readiness probes. This is sometimes necessary now because startup time for rancher has increased since v2.5.

## QA Testing Considerations

 Install chart using --set settings to set `startupProbe.periodSeconds` and/or `startupProbe.failureThreshold`. Just configuring one should be enough. The rest of startupProbe should use sensible defaults. It should work when configuring both as well. Can check the rancher deployment to confirm everything is as expected.

### DIFF
--- a/chart/tests/deployment_test.yaml
+++ b/chart/tests/deployment_test.yaml
@@ -342,3 +342,39 @@ tests:
     - equal:
         path: spec.template.spec.containers[0].readinessProbe.periodSeconds
         value: 60
+- it: should set startupProbe periodSeconds to 60
+  set:
+    startupProbe.periodSeconds: 60
+  asserts:
+    - equal:
+        path: spec.template.spec.containers[0].startupProbe.periodSeconds
+        value: 60
+    - equal:
+        path: spec.template.spec.containers[0].startupProbe.failureThreshold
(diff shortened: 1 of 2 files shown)


cbron (dismissed):
lgtm, just link back to that original issue, which is likely closed now.

chart/tests/deployment_test.yaml
```diff
@@ -342,3 +342,35 @@ tests:
...
+        value: 30
+    - equal:
+        path: spec.template.spec.containers[0].startupProbe.failureThreshold
+        value: 10
+    - equal:
+        path: spec.template.spec.containers[0].startupProbe.httpGet.port
+        value: 80
+    - equal:
+        path: spec.template.spec.containers[0].startupProbe.httpGet.path
+        value: /healthz
```
cbron, line 376: Nit: You could add a "no startup probe exists if you don't specify something" test as well

cbron (approved):

cbron (approved):

PR #40193, opened 2023-01-17
[v2.6] Refactor RKE2/K3s provisioning tests

markusewalker:
## Problem

Recently, we refactored the hosted provisioning tests to do two of the following:

- Rework redundant provisioning code into a private method
- Separate provisioning test and additional test cases into their own, unique test functions

This greatly helps moving forward for internal CI/CD pipeline and overall helps to better organize our ever-growing test framework. As such, the same approach has been made to the RKE2/K3s provisioning tests.
 
## Solution

Like with the hosted provisioning tests, the redundant provisioning code for node provider and custom clusters has been put into a single private method. The now separated unique test functions reference this private method and only keep unique code in their respective functions.

Additionally, I added two new Windows test for specifically the DynamicInput function. This is because before there were no Windows test cases in the DynamicInput, only the static counterpart had Windows tests.

Lastly, I added a check in the custom cluster tests to see if the K8s version is less than 1.25.0. This is because PSPs are officially removed in K8s 1.25, so we need to update our tests to account for this if a cluster version is > 1.25.0.
 
### Manual Testing

Locally tested static and dynamic input functions for both node provider and custom cluster.

### Automated Testing

Jenkins runs will be provided offline to reviewers.

### DIFF
--- a/tests/v2/validation/provisioning/rke2/provisioning_node_driver_test.go
+++ b/tests/v2/validation/provisioning/rke2/provisioning_node_driver_test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/rancher/rancher/tests/framework/extensions/machinepools"
 	"github.com/rancher/rancher/tests/framework/extensions/users"
 	password "github.com/rancher/rancher/tests/framework/extensions/users/passwordgenerator"
-	pods "github.com/rancher/rancher/tests/framework/extensions/workloads/pods"
+	"github.com/rancher/rancher/tests/framework/extensions/workloads/pods"
 	"github.com/rancher/rancher/tests/framework/pkg/config"
 	namegen "github.com/rancher/rancher/tests/framework/pkg/namegenerator"
 	"github.com/rancher/rancher/tests/framework/pkg/session"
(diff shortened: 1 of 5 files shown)


tests/v2/validation/provisioning/rke2/provisioning_node_driver_test.go
```diff
@@ -199,182 +168,56 @@ func (r *RKE2NodeDriverProvisioningTestSuite) ProvisioningRKE2ClusterDynamicInpu
...
 
-					clusterToken, err := clusters.CheckServiceAccountTokenSecret(client, clusterName)
-					require.NoError(r.T(), err)
-					assert.NotEmpty(r.T(), clusterToken)
+	cluster := clusters.NewK3SRKE2ClusterConfig(clusterName, namespace, cni, cloudCredential.ID, kubeVersion, machinePools)
 
-					clusterID, err := clusters.GetClusterIDByName(r.client, clusterName)
-					assert.NoError(r.T(), err)
-					podResults, podErrors := pods.StatusPods(client, clusterID)
-					assert.NotEmpty(r.T(), podResults)
```
igomez06, line 343: Since we're removing this test I still think we should add this check to the main provisioning test. Checking the status pods.

igomez06 (commented):
Looks great just one little suggestion/comment.

tests/v2/validation/provisioning/rke2/provisioning_node_driver_test.go
markusewalker: That's a good idea, I went ahead and put it back here and in the 2.7 forwardport.

thaneunsoo (approved):
lgtm

igomez06 (approved):
Looks good! I would like to hold off on merging this until we have the code coverage branches merged too, but that will be quick.

markusewalker:
Rebased changes based upon latest code coverage PR change. Once drone passes, we're good to merge.

igomez06 (approved):
LGTM

slickwarren (approved):
LGTM

Issue #40197, opened 2023-01-18
[Backport 2.6] RKE(1) unnecessarily causes DOS when generate_serving_certificate is set to true for the kubelet


rancherbot:
This is a backport issue for https://github.com/rancher/rancher/issues/36025, automatically created via rancherbot by @HarrisonWAffel

Original issue description:

When a Rancher installation with a large number of downstream provisioned RKE1 nodes comes back online, it is possible to accidentally/inadvertently DOS the installation.

The overview of this issue is that when a large number of node agents all connect to Rancher at once, it can cause a CPU DOS as Rancher will unnecessarily generate a large number of kubelet serving certificates. This can be seen in the logs if the Rancher log level is set to `debug`.

The code snippet that causes this issue can be seen here: https://github.com/rancher/rancher/blob/v2.6.3/pkg/rkenodeconfigserver/nodeserver.go#L207-L213 

The general gist of the situation is that "expected behavior" is to automatically generate a new kubelet certificate/key for a node agent connection, and have the downstream node agent determine whether or not it needs to apply the new kubelet certificate to the local node. Evidence of this expected behavior can be found here: https://github.com/rancher/rancher/issues/24942#issuecomment-574784471

This behavior should be changed as it is not scalable at large loads.

I would say an easy reproducer for this is to create 30 clusters, each with 10 nodes per cluster. Every cluster will need to be configured with the `kubelet` option `generate_serving_certificate: true` i.e. like: https://rancher.com/docs/rke/latest/en/config-options/services/#kubelet . When Rancher is completely shut down (scaled to 0) then scaled back up, it will be possible 

Depending on the mitigation/change of behavior, this may need to be backported to prior minor release lines of Rancher.

SURE-3089

jameson-mcghee:
During testing on `v2.6.9`, I was able to reproduce a "scaled down" version of this issue when following the Steps to Reproduce. However, during testing on `v2.6-head` (Commit ID: e36364e) I was unable to reproduce this issue. Therefore I am closing this ticket as Done.

**Steps to Reproduce:**
1. Spin up a Rancher server on `v2.6-head` (Commit ID: e36364e)
2. Provision a downstream AWS (or other) Node Driver RKE1 cluster with a node pool with 1 etcd & cp node and a different node pool with 2 workers
  - During provisioning, Edit as YAML, and under `services:` add:
    `kubelet:`
     [indent]`generate_serving_certificate: true`
[...]

PR #40198, opened 2023-01-18
[2.6] Backport kubelet DOS Fix

HarrisonWAffel:
## Issue: https://github.com/rancher/rancher/issues/40197

This PR includes (cleanly) cherry picked commits (48564330, 4840cf78, a8e998e0) from the following PR's 
https://github.com/rancher/rancher/pull/39497
https://github.com/rancher/rancher/pull/40063
https://github.com/rancher/rancher/pull/39998

The problem statement for this back port is a duplicate of the initial issue, as the other PR's were raised in support of that work and did not require independent verification
## Problem
A denial of service is encountered when many clusters configured with `generate_serving_certificate` connect to the Rancher server at the same time. This is due to the fact that Rancher cannot intelligently determine if a certificate must be regenerated or not as kubelet certificates are not stored or tracked by the Rancher server. This resulted in Rancher always regenerating the kubelet certificate whenever the node agent reconnects with the Rancher server. At scale, this behavior results in all CPU time being dedicated to generating certificates. Additionally, Rancher was regenerating all of the nodes certificates for each node, resulting in many unused certificates being placed on each node and additional CPU resources being consumed. 
 
## Solution
Before attempting to reconnect with the Rancher server, the node agent will first find and inspect the kubelet serving certificate stored locally and determine if it needs to be regenerated. If the certificate needs to be regenerated, either because the SAN is out of date, the IP address is out of date, or the certificate will expire within three days, then a value is specified in the header of the reconnection attempt which will indicate that the certificate needs to be regenerated.
 
## Testing
when  `generate_serving_certificate` is enabled for the clusters kubelet service, the new kubelet certificate will be generated under these conditions 

1. The certificate does not already exist 
2. The certificate will expire in three days or less 
3. The hostname name included in the certificate does not match the nodes hostname name
4. The IP address included in the certificate does not match the current IP address of the node

Some of these conditions are hard to test, such as the expiry time which is set to 10 years in the future. The two cases below cover the main goal of this ticket. 

---
#### Test Environment / Setup
+ This issue requires that you have the ability to stop/scale down rancher to 0 pods. We are interested in the reconnection process once an existing Rancher server is brought back online. Any setup in which you can restart Rancher will work, but an HA setup will likely be the easiest configuration.
+ Create an RKE1 cluster with three nodes 
  + one of the nodes will have the etcd control plane roles 
  + two of the nodes will only have the worker role 
+ Before creating the cluster, edit the YAML and find the `services` section. Add the `generate_serving_certificate` [as described in this documentation ](https://rancher.com/docs/rke/latest/en/config-options/services/#kubelet)
+ Create the cluster and wait for all nodes to come online. Once they've come online, double check the config YAML and ensure that `generate_serving_certificate` is set to `true`. 

#### 1) Certificate should not regenerate on each reconnection

+ SSH into one of the worker only nodes and navigate to `/etc/kubernetes/ssl`. You should see several `.pem` files with the following naming convention `kube-kubelet-<IP_ADDRESS>.pem` and `kube-kubelet-<IP_ADDRESS>-key.pem`. Mark the time at which these files were created using `ls -l`. Make sure you mark the file which belongs to current node you are SSH'd into 
+ Scale down / restart the rancher server and wait for all nodes to reconnect
+ Ensure that no certificates have been updated by checking the file creation time

#### 2) Certificate should be recreated when deleted
+ SSH into one of the worker only nodes and navigate to `/etc/kubernetes/ssl`. Delete the `kube-kubelet` certificate and key which belongs to the node you are SSH'd into (the file name will include the nodes IP address)
+ scale down / restart the rancher server and wait for all nodes to reconnect 
+ ensure that the deleted files have been recreated and are present on the node
+ Also ensure that no other certificate files have been updated or modified

### Manual Testing
I've done the above steps during development to ensure that the new node agent logic works as desired. 

### Automated Testing
N/A

## QA Testing Considerations
You will need an HA rancher server or any other type of Rancher server which can be restarted.
 
Any features or functionalities which require the proper configuration of the kubelet serving certificate.

### DIFF
--- a/cmd/agent/agent_test.go
+++ b/cmd/agent/agent_test.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"github.com/stretchr/testify/assert"
+	"math/big"
+	"net"
+	"testing"
+	"time"
+)
+
+func TestKubeletCertificateNeedsRegeneration(t *testing.T) {
+
+	correctIPAddress := "167.71.188.113"
+	correctHostName := "cert-test1"
+	testCertExpiryTime := time.Now().AddDate(10, 0, 3)
+
+	type testcase struct {
+		testName      string
+		testIPAddress string
+		testHostName  string
+		testTime      time.Time
+		want          bool
+	}
+
+	testCases := []testcase{
+		{
+			testName:      "All Cert Properties Are Valid",
+			testIPAddress: correctIPAddress,
+			testHostName:  correctHostName,
+			testTime:      time.Now(),
+			want:          false,
+		},
+		{
+			testName:      "Cert Contains Invalid IP Address",
+			testIPAddress: "192.168.1.1",
+			testHostName:  correctHostName,
+			testTime:      time.Now(),
+			want:          true,
+		},
+		{
+			testName:      "Cert Contains Invalid Hostname",
+			testIPAddress: correctIPAddress,
+			testHostName:  "different-hostname",
+			testTime:      time.Now(),
+			want:          true,
+		},
+		{
+			testName:      "Cert Will Expire In Less Than Three Days",
+			testIPAddress: correctIPAddress,
+			testHostName:  correctHostName,
+			testTime:      time.Now().AddDate(10, 0, 1), // ~ 48 hours until expiry
+			want:          true,
+		},
+	}
+	t.Log("generating test certificate...")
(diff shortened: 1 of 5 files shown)


cmd/agent/agent_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/stretchr/testify/assert"
+	"math/big"
+	"net"
+	"testing"
+	"time"
+)
+
+func TestKubeletCertificateNeedsRegeneration(t *testing.T) {
+
+	correctIPAddress := "167.71.188.113"
```
jakefhyde, line 20: Just for future reference, I like to use either well known ranges `192.168.0.0/16` or the `TEST-NET-1` example range `192.0.2.0/24`: https://stackoverflow.com/a/6778154/8976784

PR #40201, opened 2023-01-18
[v2.6] Auto node replace and encryption key rotation fixes

jakefhyde:
## Problem

This PR fixes node auto replace and encryption key rotation by changing how the control plane's ready status field is set, and fixing planner thrashing respectively.

### Node Auto Replace

1. Delete a node from an Active cluster with Auto Node Replace set to 30 seconds
2. The planner will attempt to reconcile the node, fail, and set the RKEControlPlane (and Provisioning cluster) to `Updated = False` (to indicate there was an error applying the plan).
3. This is then reconciled onto the management cluster here: [https://github.com/rancher/rancher/blob/da810a7523fb4c1b529bee374e9d2574f33ab0c8/p[…]ntrollers/provisioningv2/rke2/provisioningcluster/controller.go](https://github.com/rancher/rancher/blob/da810a7523fb4c1b529bee374e9d2574f33ab0c8/pkg/controllers/provisioningv2/rke2/provisioningcluster/controller.go#L291)
4. This was then copied back from the management cluster onto the controlplane object https://github.com/rancher/rancher/blob/da810a7523fb4c1b529bee374e9d2574f33ab0c8/pkg/controllers/provisioningv2/rke2/provisioningcluster/controller.go#L291
5. When this happens, we set initialized and ready on control plane status to false
6. This was being copied to the infrastructure cluster object here: https://github.com/rancher/rancher/pull/39875/files#r1047940430
These two snippets from the CAPI MachineHealthCheck controller prevent machine health checks from running if either the infrastructure cluster or control plane are not ready/initialized respectively https://github.com/kubernetes-sigs/cluster-api/blob/ccba46ec29e10d8e7c967f242a0a1970dd8dce64/internal/controllers/machinehealthcheck/machinehealthcheck_targets.go#L114-L126

## Solution

- We need a mechanism to get the owning CAPI cluster or machine objects in controllers to determine if reconcilation is paused.
- Add missing predicates to rke cluster controller, including checking for pausing as well as other required predicates we were missing. Also remove sync with the controlplane object for observedGeneration.

[cluster-api.sigs.k8s.io/developer/architecture/controllers/cluster.html?highlight=infrastructure%20cluster#infrastructure-provider](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/cluster.html?highlight=infrastructure%20cluster#infrastructure-provider)
- Add missing predicates to rke bootstrap controller, including checking for pausing as well as other required predicates we were missing.

[cluster-api.sigs.k8s.io/developer/providers/bootstrap.html?highlight=bootstrap#bootstrap-provider-specification](https://cluster-api.sigs.k8s.io/developer/providers/bootstrap.html?highlight=bootstrap#bootstrap-provider-specification)

Note: there are some predicates which have been added, which do not appear in the upstream docs as requirements, however these predicates (for example, checking if capiCluster.Status.InfrastructureReady == false) are added because although we functionally always set InfrastructureReady = true, bootstrapping before the cluster is ready is not desired and is subject to change in the future.
- Add missing predicates to rke machine controller, including checking for pausing as well as other required predicates we were missing.

[cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure.html#normal-resource](https://cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure.html#normal-resource)

### Node Auto Replace

-- | -- | -- | --

### Encryption Key Rotation

Fix planner thrashing caused when updating an object, discarding it in the cache, updating an objects status again (causing thrashing), and returning that object instead. Now, the signature of all planner related functions that are expected to update the controlplane return the status object and an error, where if the error is of type `ErrWaiting`, the error message is transformed onto the appropriate condition and propagated up to the provisioning cluster.

### Manual Testing

Encryption key rotation was tested manually for a freshly provisioned cluster and works as expected.

ETCD snapshot/restore was also tested, as was certificate rotation, and all performed as expected.

### Automated Testing

Encryption key rotation was also tested with validation tests, and worked as expected for 10k secrets.

## QA Testing Considerations


HarrisonWAffel (dismissed):
From what I can tell these seem to be clean cherry picked commits from 2.7, with no additional logic added. I'm not aware of any changes in 2.6 which would impact this PR, so lgtm

HarrisonWAffel (dismissed):
From what I can tell these seem to be clean cherry picked commits from 2.7, with no additional logic added. I can see the probes PR was also added without any additional changes. I'm not aware of any behavior in 2.6 which would impact this PR negatively, so lgtm

jiaqiluo (approved):
This is a backporting PR, LGTM

HarrisonWAffel (approved):

PR #40205, opened 2023-01-18
Refactor machine provision to fix raciness

jakefhyde:
## Problem

The machineprovision controller is racy with updating the infrastructure machine and the job.

This is causing all sorts of other issues, especiall with CI, but it is also the cause of auto node replace not working with etcd nodes.

## Solution

Refactor the machineprovision controller, replacing the competing controllers paradigm with a job controller that only sets the `status.jobName` on the infrastructure machine object, and allowing the infrastructure machine controller to perform all other status updates. 

### Manual Testing

Tested manually with a freshly provisioned single node cluster ~10 times.
Tested manually with a freshly provisioned 3 etcd + 2 controlplane + 3 worker cluster ~10 times. Additionally tested auto node replace, which worked quite well.

### Automated Testing

CI

## QA Testing Considerations

This issue is intermittent, and may not show up during testing. As such, it is recommended to provision a cluster with a high # of nodes multiple times.


HarrisonWAffel (commented):
one nit, looks good over all

HarrisonWAffel (approved):

Oats87:
Please do not merge until I can thoroughly go through this PR with a review.

HarrisonWAffel (approved):
one question and potentially a small nit

pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
```diff
@@ -85,7 +86,15 @@ func (h *handler) OnChange(obj *rkev1.RKEControlPlane, status rkev1.RKEControlPl
...
-	err := rke2.DoRemoveAndUpdateStatus(cp, h.doRemove(cp), h.rkeControlPlaneController.EnqueueAfter)
+
+	provCluster, err := h.provClusterCache.Get(cp.Namespace, cp.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return cp, err
+	}
+
+	// only delete machines from custom clusters, node driver will clean up themselves
+	deleteMachines := len(provCluster.Spec.RKEConfig.MachinePools) == 0
+	err = rke2.DoRemoveAndUpdateStatus(cp, h.doRemove(cp, deleteMachines), h.rkeControlPlaneController.EnqueueAfter)
```
annablender, line 97: Shouldn't we address the err here if its non nil? If we were unable to update the cp Status perhaps we could try again or return?

Suggestion

```
if err != nil || equality.Semantic.DeepEqual(status, cp.Status) {
  return cp, err
}
```

annablender (changes requested):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -317,7 +317,13 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 		return obj, err
 	}
 
-	if !infra.data.Bool("status", "jobComplete") && infra.data.String("status", "failureReason") == "" {
+	// infra was never adopted by a machine, just delete
+	if machineSet, _ := rke2.GetOwnerCAPIMachineSet(infra.obj, h.machineSetCache); machineSet != nil {
+		return obj, nil
```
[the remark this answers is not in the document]
jakefhyde: We only provision a vm in the `OnChange` handler if the controller of this infra machine is the actual machine object, so if it's still controlled by a `MachineSet`, adoption never happened. This is only really possible in edge cases, like our provisioning tests, but is still theoretically possible via automation/multiple administration access if you delete CAPI machines just before deleting your cluster.

pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
jakefhyde: Not quite, because we want to update the status of the controlplane object to expose the machine deletion status in the UI, which we may never do if we always return an error. `DoRemoveAndUpdateStatus` doesn't actually update status, it just populates the expected conditions, which is a confusing name and IMO bad paradigm, but that is something to be fixed next release cycle IMO.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -101,8 +90,10 @@ func (h *handler) getArgsEnvAndStatus(infra *infraObject, args map[string]interf
 		},
 		Data: getWhitelistedEnvVars(),
 	}
-	machine, err := rke2.GetMachineByOwner(h.machineCache, infra.meta)
-	if err != nil && (create || !errors.Is(err, rke2.ErrNoMachineOwnerRef)) {
+	// the owner reference may be deleted if the machine is cleaned up forcefully
+	machineName := infra.meta.GetLabels()[CapiMachineName]
```
Oats87, line 94: Potentially, do we want to "fallback" one way or the other to the owner ref or the label if one or the other doesn't exist? Not an actual problem, just wondering.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
```diff
@@ -173,7 +164,7 @@ func (h *handler) getArgsEnvAndStatus(infra *infraObject, args map[string]interf
 		Args:                cmd,
 		BackoffLimit:        jobBackoffLimit,
 		RKEMachineStatus: rkev1.RKEMachineStatus{
-			Ready:                     infra.data.String("spec", "providerID") != "" && infra.data.Bool("status", "jobComplete"),
+			Ready:                     infra.data.String("spec", "providerID") != "",
```
Oats87, line 167: We reuse job complete -- the same field is used for delete which throws us off.

Oats87: Can we refactor the `jobComplete` to do `createJobComplete` vs `deleteJobComplete`? ready should probably only ever care about the `createJobComplete`, although if the `deleteJobComplete` is being run the ready should be false because it is no longer ready

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -315,6 +317,11 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 		return obj, err
 	}
 
+	// infra was never adopted by a machine, just delete
```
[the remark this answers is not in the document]
Oats87: nit: can we add more context to this comment? Specifically around our observed machineset -> machine owner reference change, I think this would help external folks understand what is going on here more clearly

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -332,9 +338,12 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 			logrus.Tracef("[machineprovision] create job object for %s was %+v", key, job)
 		}
 		return obj, fmt.Errorf("cannot delete machine %s because create job has not finished", infra.meta.GetName())
+	} else if cond == nil {
+		return obj, nil
```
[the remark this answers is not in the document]
Oats87: agree with @HarrisonWAffel , a small comment like:
> If the createJobCondition is not set on this infra object, that means we never actually tried to run a create job, so there should be no infrastructure, proceed with delete

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -409,13 +418,22 @@ func (h *handler) OnRemove(key string, obj runtime.Object) (runtime.Object, erro
 }
 
 func (h *handler) doRemove(infra *infraObject) (runtime.Object, error) {
-	obj, err := h.run(infra, false)
+	state, _, err := h.run(infra, false)
```
Oats87, line 421: for other reviewers, the patch/update below was moved here from the run function

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -425,28 +443,33 @@ func (h *handler) EnqueueAfter(infra *infraObject, duration time.Duration) {
...
 	if err != nil {
-		logrus.Errorf("[machineprovision] %s/%s: error getting machine by owner reference %v", infra.meta.GetNamespace(), infra.meta.GetName(), err)
+		logrus.Errorf("[machineprovision] %s/%s: error getting machine by owner reference: %v", infra.meta.GetNamespace(), infra.meta.GetName(), err)
 		return obj, err
 	}
 
+	// the owner reference may be deleted if the machine is cleaned up forcefully
+	if infra.meta.GetLabels()[CapiMachineName] == "" {
+		infra.data.SetNested(machine.Name, "metadata", "labels", CapiMachineName)
+		return h.dynamic.Update(&unstructured.Unstructured{
```
Oats87, line 468: can we add a comment here stating that we don't want to "continue" logic if we run this Update because we could encounter an error further down and we're trying to keep the cache as up to date as possible

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -476,60 +499,88 @@ func (h *handler) OnChange(obj runtime.Object) (runtime.Object, error) {
...
+	create := infra.meta.GetDeletionTimestamp().IsZero()
+
+	if !create {
+		// do not recreate jobs if about to delete
+		if cond := getCondition(infra.data, deleteJobConditionType); cond != nil && cond.Status() == "True" {
+			return obj, nil
+		}
+	}
+
+	state, failedCreate, err := h.run(infra, create)
```
Oats87, line 511: We shouldn't run delete jobs ever from the OnChange handler, that should solely be in the OnRemove handler

pkg/controllers/provisioningv2/rke2/rkecontrolplane/controlplane.go
```diff
@@ -85,7 +86,15 @@ func (h *handler) OnChange(obj *rkev1.RKEControlPlane, status rkev1.RKEControlPl
...
 	cp = cp.DeepCopy()
-	err := rke2.DoRemoveAndUpdateStatus(cp, h.doRemove(cp), h.rkeControlPlaneController.EnqueueAfter)
+
+	provCluster, err := h.provClusterCache.Get(cp.Namespace, cp.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return cp, err
+	}
+
+	// only delete machines from custom clusters, node driver will clean up themselves
+	deleteMachines := len(provCluster.Spec.RKEConfig.MachinePools) == 0
```
Oats87, line 96: Can we be more specific about looking for custom machines?

Oats87 (changes requested):

pkg/controllers/provisioningv2/rke2/machineprovision/controller.go
```diff
@@ -568,31 +588,27 @@ func (h *handler) run(infra *infraObject, create bool) (rkev1.RKEMachineStatus,
...
 
-func (h *handler) patchStatus(d data.Object, state rkev1.RKEMachineStatus) error {
+// reconcileStatus will update the infra machine's status by updating each field based on the value of status.
+func reconcileStatus(d data.Object, state rkev1.RKEMachineStatus) error {
 	statusData, err := convert.EncodeToMap(state)
 	if err != nil {
 		return err
 	}
 
-	if state.JobComplete {
```
jakefhyde, line 535: To the reviewer: this was removed because it is only relevant when switching from create to delete, so the logic to reset the `failureMessage` and `failureReason` was moved to the `OnRemove` handler.

pkg/controllers/provisioningv2/rke2/machineprovision/args.go
jakefhyde: We could, but if the label is not present, we never reconciled far enough to create infra anyways, so we don't have to.

Oats87 (approved):

HarrisonWAffel (approved):

PR #40206, opened 2023-01-18
Wait for probes during encryption key rotation restart

jakefhyde:
## Problem

With the removal of the killall script, some resources are left over after restarting the `rke2-server/k3s` services during encryption key rotation.

## Solution

Wait for probes after restarting the nodes to ensure the nodes are actually healthy and ready to receive API requests.

### Manual Testing

Tested manually with a freshly provisioned single node cluster.

### Automated Testing

Tested with a freshly provisioned single node cluster, then again with 10k secrets.
Tested with a freshly provisioned 3 etcd + 2 controlplane + 3 worker cluster, then again with 10k secrets.

## QA Testing Considerations

During testing, https://github.com/rancher/rke2/issues/3801 was discovered. This issue is unrelated to rancher, but will not be fixed until a future release of k3s/rke2. If this bug is encountered during encryption key rotation, there is no recourse except for an etcd restore, but this issue occurs intermittently.

### DIFF
--- a/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
+++ b/pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
@@ -141,7 +141,7 @@ func (p *Planner) resetEncryptionKeyRotateState(status rkev1.RKEControlPlaneStat
 
 // rotateEncryptionKeys first verifies that the control plane is in a state where the next step can be derived. If encryption key rotation is required, the corresponding phase and status fields will be set.
 // The function is expected to be called multiple times throughout encryption key rotation, and will set the next corresponding phase based on previous output.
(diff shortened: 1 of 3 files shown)


pkg/provisioningv2/rke2/planner/encryptionkeyrotation.go
```diff
@@ -447,17 +447,52 @@ func (p *Planner) encryptionKeyRotationRestartService(cp *rkev1.RKEControlPlane,
...
+	_, config, err := p.generatePlanWithConfigFiles(cp, tokensSecret, entry, joinServer)
+	if err != nil {
+		return "", status, err
+	}
+
+	probes, err := p.generateProbes(cp, entry, config)
+	if err != nil {
+		return "", status, err
+	}
+	nodePlan.Probes = probes
```
Oats87, line 480: Is there any way we can refactor probe generation to not require this giant preamble every time we need the probes for a node...?

Oats87 (changes requested):

Oats87 (approved):

HarrisonWAffel (approved):

PR #40219, opened 2023-01-19
Update system-agent to v0.3.1 which includes arm64 support for the SUC

sbstp:
## Problem

The system agent was built for arm64 but the system agent update controller was not, which made updates to the system agent impossible.
 
## Solution

This PR updates the version of the system agent in use by rancher, this new version has arm64 builds.

### Manual Testing

* Create a new downstream cluster using arm64 on AWS, no issues coming online.
* Create a cluster on the release/v2.7 branch which has v0.2.13, then stop rancher, switch to this branch and restart rancher. The system agent version should now be v0.3.1 in rancher. **The cluster does not automatically update and I might need help figuring out the exact outcome expected from this.**

* Not sure how rancher updates the system agent when the version changes.


jiaqiluo (approved):

snasovich (approved):
LGTM 🚀

Oats87 (approved):

sbstp:
As far as I can tell from my testing, changing this value does not trigger an update of the downstream clusters. The only place where the setting is read is for the Linux install script, which is only used when a new node is provisioned in the cluster.

In either case I verified that this update does not break existing clusters and it does not prevent new clusters from being created, so everything should be clear for merging.

PR #40236, opened 2023-01-20
[release/v2.7] make drain in provisioning tests aggressive

Oats87:
Issue: https://github.com/rancher/rancher/issues/40267

Make provisioning tests drain super aggressive

### DIFF
--- a/pkg/apis/rke.cattle.io/v1/cluster.go
+++ b/pkg/apis/rke.cattle.io/v1/cluster.go
@@ -70,7 +70,7 @@ type DrainOptions struct {
 	// (even when set to true, kubectl won't delete pods - so setting default to true)
 	IgnoreDaemonSets *bool `json:"ignoreDaemonSets"`
 	// IgnoreErrors Ignore errors occurred between drain nodes in group
-	IgnoreErrors bool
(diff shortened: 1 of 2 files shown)


jakefhyde (approved):

pkg/apis/rke.cattle.io/v1/cluster.go
```diff
@@ -70,7 +70,7 @@ type DrainOptions struct {
 	// (even when set to true, kubectl won't delete pods - so setting default to true)
 	IgnoreDaemonSets *bool `json:"ignoreDaemonSets"`
 	// IgnoreErrors Ignore errors occurred between drain nodes in group
-	IgnoreErrors bool
+	IgnoreErrors bool `json:"ignoreErrors"`
```
snasovich, line 73: Looks like a sane change, but curious if it's related to the issue at hand as we're not setting `IgnoreErrors` in the test below? Maybe we should?

snasovich (approved):
The build has passed on rerun, so approving provided there is an explanation/response to my previous comment.

jakefhyde (approved):

PR #40308, opened 2023-01-25
reenqueue provisioning log config map if cluster not found

jakefhyde:
## Problem

If the provisioning log config map fails to reconcile on it's first iteration, it will never reconcile. 

## Solution

If a config map meets the expected criteria to be the provisioning log, but the provisioning cluster cannot be accessed, reenqueue it for evaluation later. 

## Testing

Due to this being a race condition, testing is quite difficult, as the window between creating a cluster and the config map being reconciled is quite slim.

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
+++ b/pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
@@ -27,14 +27,15 @@ var (
 	clusterRegexp = regexp.MustCompile("^c-m-[a-z0-9]{8}$")
 )
 
-func Register(ctx context.Context, wContext *wrangler.Context) {
+func Register(ctx context.Context, clients *wrangler.Context) {
 	h := &handler{
-		configMapsCache: wContext.Core.ConfigMap().Cache(),
-		configMaps:      wContext.Core.ConfigMap(),
-		clusterCache:    wContext.Provisioning.Cluster().Cache(),
+		configMapsCache: clients.Core.ConfigMap().Cache(),
+		configMaps:      clients.Core.ConfigMap(),
+		clusterCache:    clients.Provisioning.Cluster().Cache(),
 	}
-	wContext.Core.Namespace().OnChange(ctx, "prov-log-namespace", h.OnNamespace)
-	wContext.Core.ConfigMap().OnChange(ctx, "prov-log-configmap", h.OnConfigMap)
+
+	clients.Core.Namespace().OnChange(ctx, "prov-log-namespace", h.OnNamespace)
+	clients.Core.ConfigMap().OnChange(ctx, "prov-log-configmap", h.OnConfigMap)
 }
 
 type handler struct {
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
```diff
@@ -44,17 +45,19 @@ type handler struct {
 }
 
 func (h *handler) OnConfigMap(key string, cm *corev1.ConfigMap) (*corev1.ConfigMap, error) {
-	if cm == nil {
+	if cm == nil || !cm.DeletionTimestamp.IsZero() {
```
annablender, line 48: Nit: I'd prefer `configMap` instead of `cm` I think it reads better

annablender (approved):
Logic LGTM. Were you able to verify that this fixes a bug despite the thin window of testability?

pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
snasovich: There are a few places in Rancher where the same abbreviation is used already, plus it's already existing code and renaming `cm` in func signature would result in a lot of changes making reviewing it harder (unless put in a separate commit).

pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
```diff
@@ -27,14 +27,15 @@ var (
 	clusterRegexp = regexp.MustCompile("^c-m-[a-z0-9]{8}$")
 )
 
-func Register(ctx context.Context, wContext *wrangler.Context) {
+func Register(ctx context.Context, clients *wrangler.Context) {
```
snasovich, line 30: Nit, but not sure why this got renamed? `wContext` is quite an established name across Rancher codebase (regardless of whether I like it or not :P )

snasovich (approved):
LGTM, still want @Oats87 review as well

pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
jakefhyde: This is the only file in this folder that refers to it as `wContext` instead of `clients`, after working solely in these controllers for a month this seemed like the outlier to me.

snasovich: Ah, makes sense. I looked globally where `wClient` is very common.

jakefhyde:
> Logic LGTM. Were you able to verify that this fixes a bug despite the thin window of testability?

I've tested this with the happy path, and verified that if the cluster does not exist, the provisioning map will successfully reenqueue, which should confirm the race condition is no longer present.

pkg/controllers/provisioningv2/rke2/provisioninglog/provisioninglog.go
```diff
@@ -44,17 +45,19 @@ type handler struct {
...
 		return nil, nil
 	}
 	if cm.Name != provisioningLogName || (!clusterRegexp.MatchString(cm.Namespace) && cm.Namespace != "local") {
 		return cm, nil
 	}
 	provCluster, err := h.clusterCache.GetByIndex(clusterindex.ClusterV1ByClusterV3Reference, cm.Namespace)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return cm, err
+	}
 	if apierrors.IsNotFound(err) || len(provCluster) == 0 {
```
jakefhyde, line 58: To the reviewer: if err was not nil, `len(provCluster) == 0` was always true, so we never reenqueued ):

Oats87 (approved):

PR #39782, opened 2022-12-06
[DNM] support 1.25 

jiaqiluo:

PR #40309, opened 2023-01-25
[2.7] Cluster Name Length Fix - Update CAPI Version to 1.2.8

HarrisonWAffel:
## Problem
When provisioning RKE2/K3s clusters using a name containing 63 characters, the cluster would fail to provision and CAPI errors could be seen in the Rancher logs. The root cause of this issue seems to due to [a change in CAPI `v1.2.7`, which placed the name of the MachineSet onto the Machines created from that set, via a label](https://github.com/HarrisonWAffel/cluster-api/commit/4f536d8517d8288be8cfe65dec1082b62f538196#diff-340d8c7320ace7a81632cf023a171b8d22ab2424c2dc0fd84bbd7447a15e6ee2R549-R566). The CAPI version embedded within Rancher was changed from `v1.2.0` to `v1.2.7` [in anticipation of Kubernetes `v1.25` support](https://github.com/rancher/rancher/pull/39782#discussion_r1041680436). Rancher uses the cluster name in various CAPI resources, and that name may have additional text appended to it, exceeding 63 characters. Because the name of this MachineSet can exceed 63 characters, the label placed on the Machine would be invalid and Machine reconciliation would not proceed, preventing the cluster from becoming active. The following errors could be seen rapidly repeating themselves in the Rancher logs 

```
failed to sync MachineSet replicas: failed to clone bootstrap configuration for MachineSet "loremipsumdolorsitametconsectetueradipiscingelitaeneancom-c71cd-6d898ccfc8" in namespace "fleet-default": RKEBootstrap.rke.cattle.io "loremipsumdolorsitametconsectetueradipiscingelitaeneancom-4g8vg" is invalid: metadata.labels: Invalid value: "loremipsumdolorsitametconsectetueradipiscingelitaeneancom-c71cd-6d898ccfc8": must be no more than 63 characters
```

This issue was[ identified and fixed in `v1.2.8` of CAPI](https://github.com/HarrisonWAffel/cluster-api/commit/fca87beb1b65f0724bacddfdb625faa4aeb0ebbe#diff-340d8c7320ace7a81632cf023a171b8d22ab2424c2dc0fd84bbd7447a15e6ee2R562-R567), and now the label value is compressed so as to not exceed 63 characters. 

## Solution
Bump the CAPI version used in rancher to `v1.2.8` so we can pick up this bug fix, resolving reconciliation errors seen for RKE2/K3s. 

I've looked at the change logs for this CAPI version and do not see any breaking changes or significant differences between `v1.2.8` and `v1.2.7`, other than bug fixes. I've been able to create clusters and manage them without issue. The change log can be found here https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.2.8, and if there **is** an issue with `v1.2.8` we can also _downgrade_ the CAPI version to avoid this reconciliation error. At minimum CAPI `v1.2.2` is needed for Kubernetes `v1.25` support.
 
## Testing
+ Use a version of Rancher embedding CAPI `v1.2.8`
+ Create an RKE2 cluster which has a name containing 63 characters
+ Observe that the cluster provisions and becomes active, and that no CAPI reconciliation errors can be seen in the Rancher logs. 

### Manual Testing
I've done the above

### Automated Testing
N/A

## QA Testing Considerations
I discovered a UI bug when testing this, https://github.com/rancher/dashboard/issues/8033, but that bug is not related to the new CAPI version (I witnessed it on 2.6.9 and 2.7.0 which use `v1.2.0`)
 
I've looked at the change logs for this CAPI version and do not see any breaking changes or significant differences between `v1.2.8` and `v1.2.7`. The change log can be found here https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.2.8.


rancherbot (approved):

jiaqiluo (approved):

jakefhyde:
What happens in the upgrade case? If you provision with CAPI v1.2.7 and upgrade to v1.2.8, does the cluster resolve?

HarrisonWAffel:
@jakefhyde upgrade path looks good. I provisioned two clusters on 2.7-head, one with 63 characters in its name and one with only 11 characters. After upgrading to CAPI `v1.2.8` the 'good' cluster did not reconcile or update, and the 'bad' cluster properly provisioned after a few minutes.

PR #40310, opened 2023-01-25
Included k3d cluster creation/import into our integration testing setup.

igomez06:


igomez06:
Backport: https://github.com/rancher/rancher/pull/40310

tests/v2/integration/steveapi/json/user-a_none_fieldSelector=metadata.name=test1.json
```diff
@@ -1,8 +1,7 @@
 {
   "actions": {},
-  "count": 5,
```
cmurphy, line 3: It looks from changes like this that this may have been cherry-picked directly from the 2.6 branch? The responses in 2.7 will be slightly different because of the new 2.7-only features, so it will need to be re-run against a 2.7 cluster before committing.

igomez06: Oops I'll update it and re run

igomez06: Updated.

Dockerfile.dapper
```diff
@@ -19,6 +19,9 @@ RUN zypper -n install gcc binutils glibc-devel-static ca-certificates git-core w
 RUN zypper install -y -f docker && rpm -e --nodeps --noscripts containerd
 
 RUN curl -sLf https://github.com/rancher/machine/releases/download/${CATTLE_MACHINE_VERSION}/rancher-machine-${ARCH}.tar.gz | tar xvzf - -C /usr/bin
+RUN curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
+
+RUN zypper -n in go1.19
```
cmurphy, line 24: The bci image should already be 1.19, why reinstall it?

tests/framework/extensions/users/users.go
```diff
@@ -246,3 +250,23 @@ func RemoveClusterRoleFromUser(rancherClient *rancher.Client, user *management.U
 
 	return rancherClient.Management.ClusterRoleTemplateBinding.Delete(&roleToDelete)
 }
+
+// GetUserIDByName is a helper function that returns the cluster ID by name
```
cmurphy, line 254:
```suggestion
// GetUserIDByName is a helper function that returns the user ID by name
```

tests/v2/integration/steveapi/json/user-a_test-ns-1_limit=3&continue=nondeterministictoken.json
```diff
@@ -57,10 +57,10 @@
     }
   ],
   "links": {
-    "self": "https://localhost:8443/v1/secrets/test-ns-1"
+    "self": "https://rancherurl/v1/secrets/test-ns-1"
   },
   "pagination": {
-    "first": "https://localhost:8443/v1/secrets/test-ns-1",
+    "first": "https://192.168.1.245/v1/secrets/test-ns-1",
```
cmurphy, line 63: This one got missed somehow

tests/v2/integration/steveapi/steve_api_test.go
```diff
@@ -238,14 +253,17 @@ func (s *SteveAPITestSuite) SetupSuite() {
 			Name:     username,
 			Enabled:  &userEnabled,
 		}
-		userObj, err := s.client.Management.User.Create(userObj)
+		userObj, err := users.CreateUserWithRole(client, userObj, "user")
```
cmurphy, line 256: Would "user-base" be sufficient here, since we're adding cluster-member and additional roles? I'm noticing all the JSONs for users b and c now have a "createTypes" field which only happens when users have permission to create resources for that schema, which shouldn't be necessary for these tests.

tests/v2/integration/steveapi/steve_api_test.go
```diff
@@ -1598,6 +1632,34 @@ func formatJSON(obj *clientv1.SteveCollection) ([]byte, error) {
...
+				links["update"] = replaceURL
+			}
+			// replace view
+			replaceLink := links["view"]
+			if replaceLink != nil {
+				replaceURL = urlRegex.ReplaceAllString(replaceLink.(string), defautlUrlString)
+				links["view"] = replaceURL
+			}
+			data[i].(map[string]interface{})["links"] = links
+
```
cmurphy, line 1662: We can simplify all this - on line 1673 (past the end of this diff) do

```
jsonString = urlRegex.ReplaceAllString(jsonString, defautlUrlString)
```

Then all the string replaces in individual parts of the map can be removed.

Dockerfile.dapper
igomez06: This was an accident I'll remove it

tests/v2/integration/steveapi/steve_api_test.go
igomez06: It could be when I was debugging the failures in the first place I thought this was the issue. I can change it back and rerun it.

Dockerfile.dapper
igomez06: Yeah for some reason this line is the v2.6 branch, and I just cherrypicked it. Weird

tests/framework/extensions/users/users.go
igomez06: Updated

tests/v2/integration/steveapi/json/user-a_test-ns-1_limit=3&continue=nondeterministictoken.json
igomez06: updated

tests/v2/integration/steveapi/steve_api_test.go
igomez06: Changed it back

tests/v2/integration/steveapi/steve_api_test.go
igomez06: I completely miss this, made it much simpler, thank you!

cmurphy (approved):

tests/v2/integration/setup/main.go
```diff
@@ -1,51 +1,104 @@
...
+	client, err := rancherClient.NewClient("", testSession)
+	if err != nil {
+		logrus.Fatalf("error creating admin client: %v", err)
+	}
+
+	agentSetting := &v3.Setting{}
+
+	agentSettingResp, err := client.Steve.SteveType("management.cattle.io.setting").ByID("agent-image")
+	if err != nil {
+		logrus.Fatalf("error creating admin client: %v", err)
```
caliskanugur, line 72: Just a nit about error logs, below are the same as well

caliskanugur (commented):
Just a nit, looks good

tests/v2/integration/setup/main.go
igomez06: updated

cmurphy (approved):

caliskanugur (approved):

igomez06:
@superseb I had to make some changes to the Dockerfile in order to be able to run k3d as an import cluster in our integration testing. Would really appreciate a review. Thank you!

scripts/ci
```diff
@@ -12,6 +12,6 @@ fi
 
 ./validate
 ./build
-./test
 ./package
+./test
```
bmdepesa, line 16: Why are we moving the test script to after the package script?

igomez06: This is so we use the `rancher-agent` image built by drone instead of `rancher-agent:latest`. This is so integration tests can written to test against latest changes of rancher-agent

Dockerfile.dapper
```diff
@@ -19,6 +19,7 @@ RUN zypper -n install gcc binutils glibc-devel-static ca-certificates git-core w
 RUN zypper install -y -f docker && rpm -e --nodeps --noscripts containerd
 
 RUN curl -sLf https://github.com/rancher/machine/releases/download/${CATTLE_MACHINE_VERSION}/rancher-machine-${ARCH}.tar.gz | tar xvzf - -C /usr/bin
+RUN curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
```
superseb, line 22: This will work but can lead to unexpected breaking if a new version releases which has issues or is not backwards compatible.

Dockerfile.dapper
```diff
@@ -109,7 +110,7 @@ ENV DAPPER_ENV REPO TAG DRONE_TAG DRONE_COMMIT DRONE_BRANCH DRONE_BUILD_EVENT SY
 ENV DAPPER_SOURCE /go/src/github.com/rancher/rancher/
 ENV DAPPER_OUTPUT ./bin ./dist ./go.mod ./go.sum ./pkg/apis/go.mod ./pkg/apis/go.sum ./pkg/client/go.mod ./pkg/client/go.sum ./scripts/package ./pkg/settings/setting.go ./package/Dockerfile ./Dockerfile.dapper
 ENV DAPPER_DOCKER_SOCKET true
-ENV DAPPER_RUN_ARGS "-v rancher2-go16-pkg-1:/go/pkg -v rancher2-go16-cache-1:/root/.cache/go-build --privileged"
+ENV DAPPER_RUN_ARGS "-v rancher2-go16-pkg-1:/go/pkg -v rancher2-go16-cache-1:/root/.cache/go-build --privileged --add-host=host.docker.internal:172.17.0.1 --network=host"
```
superseb, line 113: There is no description in the issue that explains why this is needed

scripts/test
```diff
@@ -76,19 +76,28 @@ sleep 5
 while ! curl -sf http://localhost:8080/ping; do
     sleep 2
 done
+
 while ! kubectl --kubeconfig /etc/rancher/k3s/k3s.yaml rollout status -w -n cattle-system deploy/rancher-webhook; do
     sleep 2
 done
 #kill $TPID
 
+# get correct agent tag
```
superseb, line 85: Can we do this in `scripts/version`?

Dockerfile.dapper
igomez06: Are you suggesting, downloading the specific version instead?

scripts/test
igomez06: This needs to be fixed regardless as this wouldn't work when releases are tagged.

Dockerfile.dapper
igomez06: My apologies, so the k3d cluster runs in a docker container, so when running k3d in a DinD environment there were issues with binding to the default 0.0.0.0 address. Highlighted by this ticket: https://github.com/k3d-io/k3d/issues/1008 that is the reason for this flag: `--add-host=host.docker.internal:172.17.0.1`

`--network=host` is when we import the k3d cluster into rancher, the k3d cluster need to know its from the IP address of the host machine, again because running in a DinD environment.

scripts/test
igomez06: So I decided to update `scripts/version` with this https://github.com/rancher/rancher/blob/9fcc2f6751e777531b5b5d33726e2f7577596eac/scripts/version#L4-L7

igomez06:
@superseb could I get another review on this? Thank you!

Dockerfile.dapper
```diff
@@ -19,6 +19,7 @@ RUN zypper -n install gcc binutils glibc-devel-static ca-certificates git-core w
 RUN zypper install -y -f docker && rpm -e --nodeps --noscripts containerd
 
 RUN curl -sLf https://github.com/rancher/machine/releases/download/${CATTLE_MACHINE_VERSION}/rancher-machine-${ARCH}.tar.gz | tar xvzf - -C /usr/bin
+RUN curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=v5.4.6 bash
```
superseb, line 22: If you structure this according to the other components (with an env var), we can update it more easily using Renovate

scripts/test
superseb: Hmm, the implementation for `ETCD_UNSUPPORTED_ARCH` basically breaks part of the `scripts/version` script, which is the annoying part. Having this exception works but is not ideal as the whole check for `DIRTY` is that the repository has not been modified. If something modifies `package/Dockerfile` somewhere down the road, we won't be notified which is bad.

Dockerfile.dapper
igomez06: After some testing ` --add-host=host.docker.internal:172.17.0.1` is not necessary. Having `--network=host` is enough to be able to bind the k3d cluster to the local address and then be able to import the k3d cluster into rancher.

scripts/test
igomez06: So I updated the if statement to this instead https://github.com/rancher/rancher/blob/eb9d4f3c310a326763a78a1b5110e7ed0e7fb778/scripts/version#L4-L8
Now I check to see the `ETCD_UNSUPPORTED_ARCH` change instead.

Dockerfile.dapper
```diff
@@ -109,7 +112,7 @@ ENV DAPPER_ENV REPO TAG DRONE_TAG DRONE_COMMIT DRONE_BRANCH DRONE_BUILD_EVENT SY
 ENV DAPPER_SOURCE /go/src/github.com/rancher/rancher/
 ENV DAPPER_OUTPUT ./bin ./dist ./go.mod ./go.sum ./pkg/apis/go.mod ./pkg/apis/go.sum ./pkg/client/go.mod ./pkg/client/go.sum ./scripts/package ./pkg/settings/setting.go ./package/Dockerfile ./Dockerfile.dapper
 ENV DAPPER_DOCKER_SOCKET true
-ENV DAPPER_RUN_ARGS "-v rancher2-go16-pkg-1:/go/pkg -v rancher2-go16-cache-1:/root/.cache/go-build --privileged"
+ENV DAPPER_RUN_ARGS "-v rancher2-go16-pkg-1:/go/pkg -v rancher2-go16-cache-1:/root/.cache/go-build --privileged --network=host"
```
superseb, line 115: If this is the only way to do it, I guess we have to go with it. The runners are protected by a security group so it won't be accessible from the outside. How does this run locally?

scripts/test
superseb: I appreciate the "cleaner" approach but I'd rather fix this properly and just set the environment variable if it is needed and avoid messing with local files (I know this is not your implementation). But I don't now with who this was discussed or what the deadline is for this PR

igomez06:
> @igomez06 Looking at the build log, it logs an enormous amount of things, is this expected? I'd prefer to keep it as little as possible (only log errors). https://drone-pr.rancher.io/rancher/rancher/27287/3/2

Hi @superseb yes that is expected, but only for this PR run. This is coming from these tests https://github.com/rancher/rancher/tree/release/v2.7/tests/v2/integration/steveapi. We made updates to the json files so when the integration tests are ran, it won't modify those files and therefore not have that huge git diff anymore. 

These tests are now ran against the k3d cluster rather than the local cluster, and before the changes it would modify the json files after every test run. So we modified the json to have a generic api/hostname so it will no longer modify them after every run.

Dockerfile.dapper
igomez06: This runs locally no problem without the need of this. This is only the case when running in a DinD environment, simply because the k3d cluster container needs to be able to communicate with its host container in order to be imported to rancher. I can run it outside of a docker container.

scripts/test
igomez06: Would this be a better solution? How about removing this conditional from rancher/scripts/package
https://github.com/rancher/rancher/blob/cc77991e41c3c4062ab115f1b57d817fe1c22fd5/scripts/package#L17-L22

and then putting this conditional in rancher/package/entrypoint.sh
```
if [ ${ETCD_ARCH} != amd64 ]; then
    export ETCD_UNSUPPORTED_ARCH=${ETCD_ARCH}
fi

```

where we set ETCD_ARCH in rancher/package/Dockerfile
`ENV ETCD_ARCH ${ARCH} `

superseb: I'd prefer a combination of ARG and ENV like we partially have to avoid breaking changes in startup scripts. I believe having it empty by default and overriding it should cover both cases but obviously should be tested first.

igomez06: Ok so I updated the if statement 

https://github.com/igomez06/rancher/blob/336245d93df97fda5cfb8755a40d2f8f989c5d80/scripts/package#L17-L22

Plus now the docker build takes an `ETCD_UNSUPPORTED_ARCH` arg

https://github.com/igomez06/rancher/blob/336245d93df97fda5cfb8755a40d2f8f989c5d80/scripts/package#L24

And updated the docker file that will set `ETCD_UNSUPPORTED_ARCH` as an env var, it's an empty value by default

https://github.com/igomez06/rancher/blob/336245d93df97fda5cfb8755a40d2f8f989c5d80/package/Dockerfile#L17

https://github.com/igomez06/rancher/blob/336245d93df97fda5cfb8755a40d2f8f989c5d80/package/Dockerfile#L231

scripts/version
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 shopt -s extglob
 
-if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+UNTRACKED=$(git status --porcelain --untracked-files=no)
+GIT_DIFF=$(git diff)
+if [ -n "$UNTRACKED" ] && [[ "${GIT_DIFF}" != *"+ENV ETCD_UNSUPPORTED_ARCH=arm64"* ]]; then
```
superseb, line 6: this can be reverted

igomez06: fixed

scripts/package
```diff
@@ -15,13 +15,13 @@ cp ../bin/rancher ../bin/agent ../bin/data.json ../bin/k3s-airgap-images.tar .
 cp ../bin/data.json ../bin/rancher-data.json
 
 if [ ${ARCH} == arm64 ]; then
-    sed -i -e '$a\' -e 'ENV ETCD_UNSUPPORTED_ARCH=arm64' Dockerfile
+    ETCD_UNSUPPORTED_ARCH=arm64
 fi
 if [ ${ARCH} == s390x ]; then
-    sed -i -e '$a\' -e 'ENV ETCD_UNSUPPORTED_ARCH=s390x' Dockerfile
+    ETCD_UNSUPPORTED_ARCH==s390x
```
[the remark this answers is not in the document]
igomez06: fixed

scripts/version
```diff
@@ -78,4 +78,4 @@ echo "APP_VERSION: $APP_VERSION"
 echo "SUFFIX: $SUFFIX"
 echo "REPO: $REPO"
 echo "TAG: $TAG"
-echo "AGENT_TAG: $AGENT_TAG"
+echo "AGENT_TAG: $AGENT_TAG"
```
superseb, line 81: This is going to annoy the next person working on this file

superseb:
One newline nit, the `/script` and `/package` parts are fine by me. 

PR needs a rebase due to conflicts.

scripts/version
igomez06: Fixed

superseb (dismissed):

tests/v2/integration/steveapi/steve_api_test.go
```diff
@@ -1533,8 +1552,10 @@ func (s *SteveAPITestSuite) TestList() {
 			jsonResp, err := formatJSON(secretList)
 			require.NoError(s.T(), err)
 			jsonFilePath := filepath.Join(jsonDir, getFileName(test.user, test.namespace, test.query))
-			err = writeResp(csvWriter, test.user, curlURL, jsonFilePath, jsonResp)
-			require.NoError(s.T(), err)
+			if !downStreamClusterRegex.MatchString(curlURL) {
```
igomez06, line 1555: @cmurphy I decided to put this check, so when the test is ran on a downstream cluster it doesn't write to the JSON, that way our drone test run does have a huge git diff at the end. What do you think?

cmurphy: That works for now. fyi this is what I'm working on to have it skip the downstream

https://github.com/rancher/rancher/pull/40536/files#diff-3366c0a89c53a1d567a3c9a95378e6723ffba144299511dd1f6c0907376c75e2R266
https://github.com/rancher/rancher/pull/40536/files#diff-3366c0a89c53a1d567a3c9a95378e6723ffba144299511dd1f6c0907376c75e2R1686-R1694

cmurphy (commented):
needs a rebase

tests/v2/integration/steveapi/steve_api_test.go
igomez06: Ok sounds good, thanks for the review!

igomez06:
@cmurphy @superseb I rebased my PR would appreciate one more review. Thank you!

superseb (dismissed):

cmurphy:
@igomez06 looks like it needs a rebase again :open_mouth:

igomez06:
> @igomez06 looks like it needs a rebase again 😮

Sorry a big PR was just merged. This will be the last rebase I promise :sweat_smile:

markusewalker (approved):

cmurphy (approved):

superseb (approved):

cmurphy:
fyi I sometimes see some instability related to creating the downstream k3d cluster https://drone-pr.rancher.io/rancher/rancher/27931/3/2 ` Failed Cluster Preparation: Failed Network Preparation: failed to create cluster network: docker failed to create new network 'k3d-auto-k3d-cluster-jdivk': Error response from daemon: Failed to program FILTER chain: iptables failed: iptables --wait -I FORWARD -o br-24d47de82769 -j DOCKER: iptables: Resource temporarily unavailable.\n (exit status 4) \n\x1b[31mERRO\x1b[0m[0000] Failed to create cluster >>> Rolling Back    \n\x1b[36mINFO\x1b[0m[0000] Deleting cluster 'auto-k3d-cluster-jdivk'    \n\x1b[31mERRO\x1b[0m[0000] failed to get cluster: No nodes found for given cluster \n\x1b[31mFATA\x1b[0m[0000] Cluster creation FAILED, also FAILED to rollback changes! \n: exit status 1"`

igomez06:
> fyi I sometimes see some instability related to creating the downstream k3d cluster https://drone-pr.rancher.io/rancher/rancher/27931/3/2 ` Failed Cluster Preparation: Failed Network Preparation: failed to create cluster network: docker failed to create new network 'k3d-auto-k3d-cluster-jdivk': Error response from daemon: Failed to program FILTER chain: iptables failed: iptables --wait -I FORWARD -o br-24d47de82769 -j DOCKER: iptables: Resource temporarily unavailable.\n (exit status 4) \n\x1b[31mERRO\x1b[0m[0000] Failed to create cluster >>> Rolling Back \n\x1b[36mINFO\x1b[0m[0000] Deleting cluster 'auto-k3d-cluster-jdivk' \n\x1b[31mERRO\x1b[0m[0000] failed to get cluster: No nodes found for given cluster \n\x1b[31mFATA\x1b[0m[0000] Cluster creation FAILED, also FAILED to rollback changes! \n: exit status 1"`

This looks like a docker networking issue. Does this happen a lot?

cmurphy:
Every once in a while, maybe 1/10 times

igomez06:
> Every once in a while, maybe 1/10 times

I'll take a deeper look and see what I can find. Thanks for the heads up.

PR #40315, opened 2023-01-25
[Backport v2.6] reenqueue provisioning log config map if cluster not found

jakefhyde:
## Problem

If the provisioning log config map fails to reconcile on it's first iteration, it will never reconcile. 

## Solution

If a config map meets the expected criteria to be the provisioning log, but the provisioning cluster cannot be accessed, reenqueue it for evaluation later. 

## Testing

Due to this being a race condition, testing is quite difficult, as the window between creating a cluster and the config map being reconciled is quite slim.


snasovich (approved):
Approving provided @Oats87 approves the main PR (https://github.com/rancher/rancher/pull/40308)

Issue #40327, opened 2023-01-26
Update Rancher's chart to support updated approach for PSPs


MbolotSuse:
Rancher will need to be updated to handle the new approach designed for PSPs. 

The chart will need to be modified include the following option:

```yaml
global:
  cattle:
     psp:
        enabled: true #PSPs enablement should default to true
```

A new file will need to be added called `validate-psp-install.yaml` in the `chart/templates` directory with the following content:

```yaml
#{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
#{{- if .Values.global.cattle.psp.enabled }}
#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
#{{- fail "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." -}}
#{{- end }}
#{{- end }}
#{{- end -}}
```

Any files that render PSPs will need to make that rendering conditional solely on the `global.cattle.psp.enabled` value, like below:

```yaml
{{- if .Values.global.cattle.psp.enabled  -}}
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
...
{{- end }}
```

This approach meets the following requirements:
- PSPs should be used by the chart by default, keeping compatibility with the current 2.7.0 approach. 
- If users attempt to request PSPs  when installing the chart in a cluster which does not support PSPs (for example, a cluster running k8s version 1.25), an error will be presented so that the user knows to change the `global.cattle.psp.enabled` value to `false` before continuing.  

A solution for this should probably start by reverting rancher/rancher#40089 to re-introduce PSPs to the chart, and then continue by applying the above approach.

jiaqiluo:
Can be validated in Rancher [v2.7.2-rc2](https://github.com/rancher/rancher/releases/tag/v2.7.2-rc2)

anupama2501:
QA to validate:

Rancher version | k8s version | psp.enabled flag | hardened jenkins flag | Additional comments | Actual Status
-- | -- | -- | -- | -- | -- |
rancher_version=v2.7-head |  k8s=1.24 | global.cattle.psp.enabled=false | hardened=False | installation goes through |
rancher_version=v2.7-head |  k8s=1.24 |  global.cattle.psp.enabled=true | hardened=True |  installation goes through | 
[...]

PR #40336, opened 2023-01-26
Add PSP references along with setting and logic to handle PSPs

crobby:
https://github.com/rancher/rancher/issues/40327

### DIFF
--- a/chart/templates/post-delete-hook-cluster-role.yaml
+++ b/chart/templates/post-delete-hook-cluster-role.yaml
@@ -33,7 +33,7 @@ rules:
   {{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
   - apiGroups: [ "policy" ]
     resources: [ "podsecuritypolicies" ]
-    verbs: [ "delete", "create" ]
+    verbs: [ "use", "delete", "create" ]
   {{- end }}
   - apiGroups: [ "networking.k8s.io" ]
     resources: [ "ingresses" ]
--- a/chart/templates/post-delete-hook-cluster-role.yaml
+++ b/chart/templates/post-delete-hook-cluster-role.yaml
@@ -30,11 +30,14 @@ rules:
   - apiGroups: [ "admissionregistration.k8s.io" ]
     resources: [ "validatingwebhookconfigurations", "mutatingwebhookconfigurations" ]
     verbs: [ "get", "list", "delete" ]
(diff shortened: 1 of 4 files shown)


chart/templates/post-delete-hook-cluster-role.yaml
```diff
@@ -30,11 +30,11 @@ rules:
   - apiGroups: [ "admissionregistration.k8s.io" ]
     resources: [ "validatingwebhookconfigurations", "mutatingwebhookconfigurations" ]
     verbs: [ "get", "list", "delete" ]
-  {{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
+{{- if .Values.global.cattle.psp.enabled }}
   - apiGroups: [ "policy" ]
     resources: [ "podsecuritypolicies" ]
-    verbs: [ "delete", "create" ]
-  {{- end }}
+    verbs: [ "use", "delete", "create" ]
```
MbolotSuse, line 36: I don't think that this approach quite works. Even if this chart doesn't have PSPs enabled, I could see the post-delete job needing to delete PSPs for other charts which are using PSPs (like a feature chart in the local cluster). You may want to consider separating these into two rules - one for the delete/create permission and one for the use permission, and then only use the if check around the use permission.

MbolotSuse (changes requested):
One concern on the RBAC permissions.

chart/templates/post-delete-hook-cluster-role.yaml
crobby: Makes sense. Done.

chart/templates/validate-psp-install.yaml
```diff
@@ -0,0 +1,5 @@
+#{{- if .Values.global.cattle.psp.enabled }}
+#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+#{{- required "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." "" -}}
```
MbolotSuse, line 3: This recently changed to recommend the use of `fail` instead of `require`. Would you mind changing this here (new content has been updated in the issue)?

MbolotSuse (changes requested):
One small change on the use of `require` vs `fail`

MbolotSuse (approved):
LGTM so long as the CI passes.

maxsokolovsky (approved):

Issue #40335, opened 2023-01-26
[BUG] Can't warm restart docker installed Rancher due to webhook conflicts


Oats87:
**Rancher Server Setup**
- Rancher version: v2.7.1, v2.7-head
- Installation option (Docker install/Helm Chart): Docker install

**Information about the Cluster**
- Kubernetes version: whatever is the K3s version we include

**Describe the bug**
On warm restarts of a Rancher docker-install, Rancher will not start due to a fatal error stemming from the rancher-webhook being unavailable.

**To Reproduce**
Run Rancher as a docker installation
Stop the rancher container
Start the rancher container

**Result**
Rancher will fail to start with logs like:
```
[root@cannon01 ~]# docker logs -f rancher39994 2>&1
INFO: Running k3s server --cluster-init --cluster-reset
2023/01/26 20:17:14 [TRACE] Loglevel set to [trace]
2023/01/26 20:17:14 [INFO] Rancher version 2c236d7b2-dirty (2c236d7b2) is starting
2023/01/26 20:17:14 [INFO] Rancher arguments {ACMEDomains:[] AddLocal:true Embedded:false BindHost: HTTPListenPort:80 HTTPSListenPort:443 K8sMode:auto Debug:false Trace:true NoCACerts:false AuditLogPath:/var/log/auditlog/rancher-api-audit.log AuditLogMaxage:10 AuditLogMaxsize:100 AuditLogMaxbackup:10 AuditLevel:0 Features: ClusterRegistry:}
2023/01/26 20:17:14 [INFO] Listening on /tmp/log.sock
2023/01/26 20:17:14 [INFO] Waiting for server to become available: Get "https://127.0.0.1:6444/version?timeout=15m0s": dial tcp 127.0.0.1:6444: connect: connection refused
2023/01/26 20:17:16 [INFO] Waiting for server to become available: Get "https://127.0.0.1:6444/version?timeout=15m0s": dial tcp 127.0.0.1:6444: connect: connection refused
2023/01/26 20:17:18 [INFO] Waiting for server to become available: Get "https://127.0.0.1:6444/version?timeout=15m0s": dial tcp 127.0.0.1:6444: connect: connection refused
2023/01/26 20:17:20 [INFO] Waiting for server to become available: Get "https://127.0.0.1:6444/version?timeout=15m0s": dial tcp 127.0.0.1:6444: connect: connection refused
2023/01/26 20:17:40 [FATAL] 2 errors occurred:
[...]

PR #40337, opened 2023-01-26
[release/v2.7] Allow for 503 when creating or mutating local cluster

Oats87:
## Issue:
https://github.com/rancher/rancher/issues/40335
 
## Problem
On warm restarts of a Rancher docker install, Rancher cannot start because the rancher-webhook deployment is not running yet.
  
## Solution
Add a retry to the local cluster creation/mutation that will retry if the error received is a 503, which could be that the rancher webhook is unavailable at the time. Since this is a warm restart of Rancher, the webhook deployment will eventually come up and the 503's will stop being emitted.
 
### Manual Testing
Start rancher as a docker install, let it come up, log into it
Restart the rancher install, observe it can still start

### Automated Testing
N/A

### DIFF
--- a/pkg/data/dashboard/cluster_data.go
+++ b/pkg/data/dashboard/cluster_data.go
@@ -1,6 +1,8 @@
 package dashboard
 
 import (
+	"time"
+
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	fleetconst "github.com/rancher/rancher/pkg/fleet"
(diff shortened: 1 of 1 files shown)


brandond (approved):

jakefhyde (dismissed):

Oats87:
@brandond @jakefhyde sorry, had to goimports the file

brandond (approved):

jakefhyde (approved):

Issue #40306, opened 2023-01-25
Update Rancher's default restricted PSACT to support RKE2 profile cis-1.23


jiaqiluo:
**Background**
Rancher allows users to set the `profile` when creating an RKE2 custom cluster. 
When the REK2 cluster version is 1.25, and the profile is set as `cis-1.23`, the RKE2 engine will configure the cluster to use RKE2-builtin [restricted PSA config file](https://github.com/rancher/rke2/blob/0ab4614b13daa5a49e484249a49012918d46ed22/pkg/rke2/psa.go#L40-L60).  

**Problem**
The problem is that the config file does not include rancher's namespaces in its exception list, so Rancher's components can not be deployed into the cluster. Therefore, cluster provision will fail. 

**Solution**
To make the cluster work, we need to do the following:
1. add RKE2's namespaces to the exception list in Rancher's default `restricted` PSACT 
2. When the REK2 cluster version is 1.25, and the profile is set as `cis-1.23`, set the option `pod-security-admission-config-file` with the PSA config file generated from the Rancher's default `restricted` PSACT.  

Note:
We have two ways to overwrite the PSA config file: setting the`pod-security-admission-config-file` option, or the `kube-apiserver-arg` option.  The final implementation will be one of those.

jiaqiluo:
can be validated in the latest 2.7-head

markusewalker:
**QA TEST PLAN**

**Scenarios**
|Scenario|Test Case|
|---|---|
|1|As a standard user, provision a downstream RKE2 v1.25 custom cluster with CIS 1.23, validate cluster and nodes are `Active`|
|2|As a standard user, provision a downstream RKE2 v1.24 custom cluster with CIS 1.6, validate cluster and nodes are `Active`, upgrade cluster to v1.25 and CIS to 1.23, validate cluster and nodes upgrade

markusewalker:
Validated that this is addressed on `v2.7-head` https://github.com/rancher/rancher/commit/ab561bf837ca6051176bf169784edb01d9dee134.

**ENVIRONMENT DETAILS**
- Rancher install: Helm
- Rancher version: `v2.7-head` https://github.com/rancher/rancher/commit/ab561bf837ca6051176bf169784edb01d9dee134
- Browser: Chrome

**TEST RESULT**
|Scenario|Test Case|Result
|---|---|---|
[...]

PR #40352, opened 2023-01-26
add the namespaces from RKE2's default restricted PSA config to Rancher's

jiaqiluo:
## Issue: 

https://github.com/rancher/rancher/issues/40306 
## Problem

 
Rancher allows users to set the profile when creating an RKE2 custom cluster.
When the REK2 cluster version is 1.25, and the profile is set as cis-1.23, the RKE2 engine will configure the cluster to use RKE2-builtin [restricted PSA config file](https://github.com/rancher/rke2/blob/0ab4614b13daa5a49e484249a49012918d46ed22/pkg/rke2/psa.go#L40-L60).
The problem is that the config file does not include rancher's namespaces in its exception list, so Rancher's components can not be deployed into the cluster. Therefore, cluster provision will fail.

## Solution

Instead of using RKE2's default restricted PSA config, we will use Rancher's default restricted PSA config when `profle=cis-1.23` is set on the RKE2 cluster, which means we need to append the namespace in the  RKE2's default restricted PSA config to Rancher's.

## Testing

We only need to confirm that "cis-operator-system" and "tigera-operator" are in the list of exception namespaces in rancher builtin restricted PSACT

### DIFF
--- a/pkg/data/management/podadmissionconfigurationtemplate_data.go
+++ b/pkg/data/management/podadmissionconfigurationtemplate_data.go
@@ -29,6 +29,8 @@ var FeatureAppNS = []string{
 	"cattle-windows-gmsa-system", // Windows GMSA
 	"cattle-sriov-system",        // Sriov
 	"cattle-ui-plugin-system",    // UI Plugin System
+	"cis-operator-system",        // From RKE2 default restricted PSA Config. See https://github.com/rancher/rke2/blob/34633dcc188d3a79744636fe21529ef6f5d64d71/pkg/rke2/psa.go#L58
+	"tigera-operator",            // From RKE2 default restricted PSA Config. See https://github.com/rancher/rke2/blob/34633dcc188d3a79744636fe21529ef6f5d64d71/pkg/rke2/psa.go#L58
 }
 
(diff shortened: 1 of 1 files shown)


pkg/data/management/podadmissionconfigurationtemplate_data.go
```diff
@@ -29,6 +29,8 @@ var FeatureAppNS = []string{
 	"cattle-windows-gmsa-system", // Windows GMSA
 	"cattle-sriov-system",        // Sriov
 	"cattle-ui-plugin-system",    // UI Plugin System
+	"cis-operator-system",        // From RKE2 default restricted PSA Config. See https://github.com/rancher/rke2/blob/release-1.25/pkg/rke2/psa.go#L58
```
snasovich, line 32: Please change the link to permalink (with commit ID) instead of branch.

snasovich (commented):
One nit to address, otherwise LGTM

pkg/data/management/podadmissionconfigurationtemplate_data.go
jiaqiluo: fixed.

snasovich (approved):

Oats87 (approved):

PR #40356, opened 2023-01-27
Updating to Fleet v0.6.0-rc.2

manno:
We don't update the `github.com/rancher/fleet/pkg/apis` go module, since it's using an older version of wrangler (1.0.2) and there are no API changes from 0.6.0-rc.1.


raulcabello (approved):
LGTM

PR #40362, opened 2023-01-27
remove duplication from the list, otherwise kube-apiserver does not work

jiaqiluo:
Issue: #40306

This PR removes duplication from the list, otherwise `kube-apiserver` does not work.


HarrisonWAffel (approved):

jakefhyde (approved):

snasovich (approved):
Ugh, sorry for missing that on the previous review.

PR #40367, opened 2023-01-27
Pass PSP enablement value from rancher chart to webhook chart

maxsokolovsky:


pkg/controllers/dashboard/chart/chart.go
```diff
@@ -12,9 +12,10 @@ import (
 const (
 	// PriorityClassKey is the name of the helm value used for setting the name of the priority class on pods deployed by rancher and feature charts.
 	PriorityClassKey = "priorityClassName"
+	PSPEnablement    = "pspEnablement"
```
crobby, line 15: nit, should probably have a comment for the constant

package/Dockerfile
```diff
@@ -54,7 +54,7 @@ ENV CATTLE_SYSTEM_UPGRADE_CONTROLLER_CHART_VERSION 101.0.0+up0.3.3
 
 # System charts minimal version
 ENV CATTLE_FLEET_MIN_VERSION=101.1.0+up0.6.0-rc.2
-ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=2.0.2+up0.3.2-rc11
+ENV CATTLE_RANCHER_WEBHOOK_MIN_VERSION=2.0.2+up0.3.2-rc12
```
crobby, line 57: rc14 is all the rage nowadays

crobby (commented):
This looks right to me.  Do we need any tests added?

KevinJoiner (changes requested):
Looks good. It should be quick to add on to the current unit tests in charts_test.go

maxsokolovsky:
@KevinJoiner, @crobby, I am a little concerned about this handler code that runs on repo change.
https://github.com/rancher/rancher/blob/9bb8d5674d4f2fe85cca799b07a6d18eee257b9e/pkg/controllers/dashboard/systemcharts/controller.go#L51
https://github.com/rancher/rancher/blob/9bb8d5674d4f2fe85cca799b07a6d18eee257b9e/pkg/controllers/dashboard/systemcharts/controller.go#L74

Not sure I understand what this does, but looks like it will replace one of the inner values map. Thus, PSP inner map would be removed. Should this merge maps instead?

crobby (approved):
lgtm once ci passes

KevinJoiner:
@maxsokolovsky That code is okay because of the [code that follows ](https://github.com/rancher/rancher/blob/9bb8d5674d4f2fe85cca799b07a6d18eee257b9e/pkg/controllers/dashboard/systemcharts/controller.go#L76-#L80)

The desired behavior is to have the system-default-registry set to the setting if the provided chart does not define the global key. 
So your comment is mostly correct, except the chart definition overrides the hardcoded value, which is why the PSP setting persists.

KevinJoiner (approved):

crobby:
Will get this merged...as soon as ci comes back to life.

Issue #40354, opened 2023-01-27
[BUG] Logging in to rancher as an auth user throws 500 error after logging out of rancher


anupama2501:
**Rancher Server Setup**
- Rancher version: v2.7-head 91be24c upgraded from 2.7.0
- Installation option (Docker install/Helm Chart): Docker install
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version: `v1.25.5+k3s1`
- Cluster Type (Local/Downstream): Local 
Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions: Admin/ openldap & FreeIPA auth admin

**Describe the bug**
On upgrading rancher server from 2.7.0 to 2.7-head, auth users are unable to login. This was seen on a fresh install of 2.7-head as well.

**To Reproduce**
1. Install rancher server on v2.7.0
2. Enable openldap/FreeIPA auth
3. Upgrade rancher server to v2.7-head
4. Try to login as an auth admin

**Result**
We get a 500 error whenever we try to login as an auth user:
```{baseType: "error", code: "ServerError", message: "Server error while authenticating", status: 500,…}
baseType
: 
"error"
code
: 
"ServerError"
message
: 
"Server error while authenticating"
status
: 
500
type
: 
"error"
```

Following are the errors seen in docker logs when we try to login and for all the auth users that were added as cluster/project members:
```
2023/01/27 02:47:57 [ERROR] API error response 500 for POST /v3-public/freeIpaProviders/freeipa?action=login. Cause: invalid server config. at least 1 server needs to be configured
2023/01/27 02:48:03 [ERROR] API error response 500 for POST /v3-public/freeIpaProviders/freeipa?action=login. Cause: invalid server config. at least 1 server needs to be configured

2023/01/26 23:55:29 [ERROR] Unknown error: users.management.cattle.io "cn=<redacted>" not found
2023/01/26 23:55:29 [ERROR] Unknown error: users.management.cattle.io "cn=<redacted> not found
2023/01/26 23:55:29 [ERROR] Unknown error: users.management.cattle.io "cn=<redacted>" not found
```

If we login as a local admin and navigate back to the auth page, we see a few details that were previously configured were empty:
[...]

Issue #39908, opened 2022-12-15
[Azure AD] Rancher does not handle an error when getting groups - leading to panic


maxsokolovsky:
**Rancher Server Setup**
- Rancher version: v2.7-head and v2.6-head
- Installation option (Docker install/Helm Chart):
   - If Helm Chart, Kubernetes Cluster and version (RKE1, RKE2, k3s, EKS, etc):
- Proxy/Cert Details:

**Information about the Cluster**
- Kubernetes version:
- Cluster Type (Local/Downstream):
   - If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider):

**User Information**
- What is the role of the user logged in? (Admin/Cluster Owner/Cluster Member/Project Owner/Project Member/Custom)
  - If custom, define the set of permissions:

**Describe the bug**
If admins do not grant sufficient permissions to their apps in Azure, such that Rancher does not have enough permissions to get groups, Rancher panics (although does not crash the server) instead of reporting an error. To the user, this all looks like an expected 500, but for the wrong reason. The panic is in the logs because Rancher does not handle the permissions error.

**To Reproduce**
Grant to your Azure app only `Group.Read.All` permission. Set up Azure AD in Rancher and login. Try searching for principals when adding project members. Observe search fails with a 500, which is fine. But there is a panic in the logs.

**Result**

**Expected Result**

**Screenshots**

**Additional context**

maxsokolovsky:
# QA Testing for v2.6 and v2.7.

1. Go to the Azure portal and set your app's permissions to:
`User.Read.All`
2. Remove any other permissions.
3. Enable Azure AD as the auth provider in Rancher.
4. Notice that you can log in and out fine.
5. Add a new project and search for users to add to it.
6. The search will fail - there will be no results, and a 500 error in the console will appear.
7. Check Rancher logs and ensure there are no panics.

dasarinaidu:
This issue is fixed on  `v2.7-c24fb8b0869a0b445f55b3307c6ed4582e147747-head`

I did not see any panic logs when I disable Group.Read.All

**Steps:**

1. Pre-condition: At AZURE portal make sure the settings

```
Group.Read.All -- Disabled
User.Read.All -- Enabled
```

2. Rancher -> Auth Provider (AZURE)
3. Connect to AZURE
4. Local cluster -> Project/Cluster members -> Try to search for a Group
5. Check for the Network and rancher Logs

**Actual:**
[...]

PR #40378, opened 2023-01-30
Add CRTB, PRTB, GRB, User cleanup when an auth provider is disabled

maxsokolovsky:
## Issue:
The first [commit](https://github.com/rancher/rancher/pull/40378/commits/b208a1c5619125578010c0cf13b82c8da197606e) has the same as the code changes already reviewed for auth provider cleaning up resources after being disabled.

The [second](https://github.com/rancher/rancher/pull/40378/commits/353738f265f3107bf0d096982550a0f3185c5c4c) is just a quick fix to address a small issue https://github.com/rancher/rancher/issues/39908, since it's related to auth providers.

~The [third](https://github.com/rancher/rancher/pull/40378/commits/d1d6ac8d75139106bf3e576325cca79fc751f56f) addresses https://github.com/rancher/rancher/issues/40354. This changes how the auth config annotation is updated. We must use the unstructured client to preserve all the fields of a config when it is updated. When using a regular interface, I was having issues with fields disappearing on update because of the structure of auth config as a type - it's not a regular CRD.~

### DIFF
--- a/pkg/auth/api/secrets/cleanup.go
+++ b/pkg/auth/api/secrets/cleanup.go
@@ -16,7 +16,7 @@ import (
 // CleanupClientSecrets tries to delete common client secrets for each auth provider.
 func CleanupClientSecrets(secretInterface corev1.SecretInterface, config *v3.AuthConfig) error {
 	if config == nil {
-		return fmt.Errorf("cannot delete auth provider resources if its config is nil")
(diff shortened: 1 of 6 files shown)


cmurphy (approved):
Next time, please split commits like the second commit into their own PR, it makes it easier to review and makes it easier for QA to tell what change fixed what issue.

cmurphy (approved):

PR #40379, opened 2023-01-30
[2.6] Add CRTB, PRTB, GRB, User cleanup when an auth provider is disabled

maxsokolovsky:
## Issue:
The first [commit](https://github.com/rancher/rancher/pull/40379/commits/588db18d1ac81602bc27eef9812f79f4d48642b4) has the same as the code changes already reviewed for auth provider cleaning up resources after being disabled.

The [second](https://github.com/rancher/rancher/pull/40379/commits/97fdbe14cacdac975befacf3838d107a99902afc) is just a quick fix to address a small issue https://github.com/rancher/rancher/issues/39908, since it's related to auth providers.

~~The [third](https://github.com/rancher/rancher/pull/40379/commits/d3dc6a7e92a4d9aa05f96ca3dfe85e27bb32d46c) addresses https://github.com/rancher/rancher/issues/40360. This changes how the auth config annotation is updated. We must use the unstructured client to preserve all the fields of a config when it is updated. When using a regular interface, I was having issues with fields disappearing on update because of the structure of auth config as a type - it's not a regular CRD.~~

### DIFF
--- a/pkg/controllers/management/auth/auth_config.go
+++ b/pkg/controllers/management/auth/auth_config.go
@@ -43,7 +43,7 @@ func newAuthConfigController(context context.Context, mgmt *config.ManagementCon
 	controller := &authConfigController{
 		users:            mgmt.Management.Users("").Controller().Lister(),
 		authRefresher:    providerrefresh.NewUserAuthRefresher(context, scaledContext),
-		cleanup:          cleanup.NewCleanupService(mgmt.Core.Secrets("")),
+		cleanup:          cleanup.NewCleanupService(mgmt.Core.Secrets(""), mgmt.Wrangler.Mgmt),
 		authConfigClient: mgmt.Wrangler.Mgmt.AuthConfig(),
 	}
 	return controller
@@ -53,8 +53,8 @@ func (ac *authConfigController) setCleanupAnnotation(obj *v3.AuthConfig, value s
 	if obj.Annotations == nil {
 		obj.Annotations = make(map[string]string)
 	}
-	obj.Annotations[CleanupAnnotation] = value
 	objCopy := obj.DeepCopy()
+	objCopy.Annotations[CleanupAnnotation] = value
 	return ac.authConfigClient.Update(objCopy)
 }
 
--- a/pkg/controllers/management/auth/auth_config_test.go
+++ b/pkg/controllers/management/auth/auth_config_test.go
@@ -69,7 +69,7 @@ func TestCleanupRuns(t *testing.T) {
 			configEnabled:      true,
 			annotationValue:    CleanupRancherLocked,
 			expectCleanup:      false,
-			newAnnotationValue: CleanupRancherLocked,
+			newAnnotationValue: CleanupUnlocked,
 		},
 		{
 			name:               "no cleanup in enabled user_locked auth config",
@@ -91,7 +91,6 @@ func TestCleanupRuns(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			t.Parallel()
 			var service cleanupService
 			controller := authConfigController{
 				cleanup:          &service,
--- a/pkg/auth/api/secrets/cleanup.go
+++ b/pkg/auth/api/secrets/cleanup.go
@@ -16,7 +16,7 @@ import (
 // CleanupClientSecrets tries to delete common client secrets for each auth provider.
 func CleanupClientSecrets(secretInterface corev1.SecretInterface, config *v3.AuthConfig) error {
 	if config == nil {
-		return fmt.Errorf("cannot delete auth provider resources if its config is nil")
+		return fmt.Errorf("cannot delete auth provider secrets if its config is nil")
 	}
 
 	fields, ok := TypeToFields[config.Type]
--- a/pkg/auth/cleanup/service.go
+++ b/pkg/auth/cleanup/service.go
@@ -2,29 +2,220 @@
 package cleanup
 
 import (
+	"errors"
 	"fmt"
+	"strings"
 
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/auth/api/secrets"
+	controllers "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	corev1 "github.com/rancher/rancher/pkg/generated/norman/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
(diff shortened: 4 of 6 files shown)


pkg/controllers/management/auth/auth_config.go
```diff
@@ -33,29 +34,46 @@ type CleanupService interface {
...
+		logrus.Errorf("auth config %s does not have the 'metadata' field", name)
+		return nil, nil
+	}
+	annotations, ok := metadata["annotations"].(map[string]any)
+	if !ok {
+		annotations = make(map[string]any)
+	}
+	annotations[CleanupAnnotation] = value
+	u.SetUnstructuredContent(authConfig)
+	return ac.authConfigs.Update(name, o)
```
KevinJoiner, line 76: You can simplify this by just using `unstructured.Unstructured` instead of `runtime.Unstructured`
```suggestion
	unstructuredObj, ok := o.(*unstructured.Unstructured)
	if !ok {
		logrus.Errorf("auth config %s is not an unstructured value", name)
		return nil, nil
	}
	annotations := unstructuredObj.GetAnnotations()
	if annotations == nil {
		annotations = make(map[string]string)
	}
	annotations[CleanupAnnotation] = value
	unstructuredObj.SetAnnotations(annotations)
	return ac.authConfigs.Update(name, unstructuredObj)
```

KevinJoiner (changes requested):

pkg/controllers/management/auth/auth_config.go
```diff
@@ -33,29 +34,46 @@ type CleanupService interface {
...
 
 type authConfigController struct {
-	users            v3.UserLister
-	authRefresher    providerrefresh.UserAuthRefresher
-	cleanup          CleanupService
-	authConfigClient controllers.AuthConfigClient
+	users         v3.UserLister
+	authRefresher providerrefresh.UserAuthRefresher
+	cleanup       CleanupService
+	authConfigs   objectclient.GenericClient
```
MbolotSuse, line 40: Nit:

Generally not a fan of the name here. Might be good to call this something like "authConfigsUnstructured".

More importantly, there's also no comment here explaining why we need to use an unstructured client for authconfigs instead of the fully-typed authconfig interface. Given that this has already bit us once, I think adding a comment explaining the whole "auth configs have fields that are internal to types and so we need to use the unstructured client to update" situation is a good idea.

pkg/controllers/management/auth/auth_config.go
```diff
@@ -33,29 +34,46 @@ type CleanupService interface {
...
 }
 
 func newAuthConfigController(context context.Context, mgmt *config.ManagementContext, scaledContext *config.ScaledContext) *authConfigController {
 	controller := &authConfigController{
```
MbolotSuse, line 44: I'd recommend leaving the comment here or in the `setCleanup` function indicating the need for an unstructured client vs a normal client.

pkg/controllers/management/auth/auth_config.go
```diff
@@ -33,29 +34,46 @@ type CleanupService interface {
...
+	if err != nil {
+		return nil, err
 	}
-	objCopy := obj.DeepCopy()
-	objCopy.Annotations[CleanupAnnotation] = value
-	return ac.authConfigClient.Update(objCopy)
+	u, ok := o.(runtime.Unstructured)
+	if !ok {
+		logrus.Errorf("auth config %s is not an unstructured value", name)
+		return nil, nil
```
MbolotSuse, line 62: I don't think that returning `nil, nil` here is the right choice. At minimum, the sync method should be the one logging the error and choosing if the controller should re-enqueue.  This return is particularly dangerous because it returns a nil value for both the error and the object - one of those should be filled in. 

From a higher-level perspective here, I think returning nil for an error is deceptive for this function. We asked it to set the annotation, and it wasn't able to because of an unexpected condition. IMO that should result in an error.

pkg/controllers/management/auth/auth_config.go
```diff
@@ -33,29 +34,46 @@ type CleanupService interface {
...
+	}
+	authConfig := u.UnstructuredContent()
+	metadata, ok := authConfig["metadata"].(map[string]any)
+	if !ok {
+		logrus.Errorf("auth config %s does not have the 'metadata' field", name)
+		return nil, nil
```
MbolotSuse, line 68: Same comment here on the use of a `nil, nil` with no error return.

pkg/controllers/management/auth/auth_config.go
```diff
@@ -33,29 +34,46 @@ type CleanupService interface {
...
+	}
+	annotations, ok := metadata["annotations"].(map[string]any)
+	if !ok {
+		annotations = make(map[string]any)
+	}
+	annotations[CleanupAnnotation] = value
```
MbolotSuse, line 74: I think that you are relying on having existing annotations for this to work. If you hit the empty case and need to make a new map, this won't update the annotations on the authconfig, since `annotations` aren't stored anywhere on the auth config. 

You can either:
- Add a line after this one to update `authConfig["metadata"]["annotations"]` to the new annotations (covering both the "no existing annotations" and "existing annotations" cases) - I prefer this option.
- Add a line in the "no existing annotations" case to update `authconfig["metadata"]["annotations"]` to the new empty value

pkg/controllers/management/auth/auth_config_test.go
```diff
@@ -104,10 +100,21 @@ func TestCleanupRuns(t *testing.T) {
...
 			obj, err := controller.sync("test", config)
 			require.NoError(t, err)
 			assert.Equal(t, test.expectCleanup, service.cleanupCalled)
-			assert.Equal(t, test.newAnnotationValue, obj.(*v3.AuthConfig).Annotations[CleanupAnnotation])
+
+			obj, err = controller.authConfigs.Get(config.Name, metav1.GetOptions{})
+			require.NoError(t, err)
+			u, _ := obj.(runtime.Unstructured)
+			authConfig := u.UnstructuredContent()
+			assert.Equal(t, test.newAnnotationValue, authConfig["metadata"].(map[string]any)["annotations"].(map[string]any)[CleanupAnnotation])
```
MbolotSuse, line 117: I don't think this test is making a meaningful assertion now. Before, you were testing the output of sync to make sure that the annotation was added. Now you are just asserting that your mock returns the authConfig that you constructed as part of the test. I would recommend changing this to either:

- Record the new value passed into the update call, and make assertions on that value (Don't like this option)
- Make assertions on the return of sync like before (much better option IMO)

MbolotSuse (changes requested):
Various issues with the final commit, namely:
- We are ignoring errors in at least two places
- The tests no longer make a meaningful assertion
- (Nit) Comments on why we need an unstructured client would be nice.

KevinJoiner (approved):

cmurphy (approved):
Fwiw, the second commit also could and should have been in a separate PR because it doesn't touch any of the same files as the first commit and is unrelated to the issue the first commit addresses.

PR #40403, opened 2023-02-02
[2.6] Remove kontainer-engine and machine drivers from drivers.rancher.cn

niusmallnan:
## Issue: 

SURE-5852
 
### Manual Testing

For the fresh installation, it will not contain these drivers.

When upgrading from the previous Rancher, since these drivers have been stored in CR, they will continue to be retained, which is also conducive to ensuring compatibility.


niusmallnan:
still active

martyav:
@niusmallnan can you re-open, as we still seem to be waiting on this issue and #40402? We still have an open docs PR pertaining to these. Thank you.

PR #40409, opened 2023-02-03
chart: Support custom pathType for ingress

nobiit:
On some ingress classes or some special architectures. People want to customize Ingress's PathType (Because sometimes ImplementationSpecific is quite ambiguous, they want to pass in Prefix for example). Let's give custom permissions to users

## Problem

- User wants to switch Ingress PathType from ImplementationSpecific -> Prefix (Due to ImplementationSpecific crashing or misbehaving)

## Solution

- Supports .Values.ingress.pathType to get custom data from values (if provided) 

## Testing

- Manual check

### Manual Testing

- `helm install rancher chart/ --set ingress.pathType=Prefix`


nobiit:
I am facing this Issue https://github.com/kubernetes/ingress-nginx/issues/9569
 so I need to change PathType

Issue #40384, opened 2023-01-31
Support for updated PSP handling for `system-upgrade-controller` (SUC) from `charts`


snasovich:
`system-upgrade-controller` (SUC) chart (https://github.com/rancher/charts/tree/dev-v2.7/charts/system-upgrade-controller) is a Managed Rancher chart currently used by RKE2/k3s v2prov provisioning framework to upgrade `system-agent` on downstream clusters' nodes. "Managed Rancher chart" here means the chart is automatically installed/uninstalled/upgraded without explicit user interaction.

## The Issue
The chart currently deploys PSP and related resources and even with the recent [change](https://github.com/rancher/charts/pull/2338/files#diff-a607fb89c3ee8edff22b3016e694bdafeb585ffe7dfe82f8474b279ed0a9f7e5R1) to not deploy PSPs if API is not available, there is a potential issue with upgrade flow.

Consider the following scenario:
1. k8s 1.24 RKE2 downstream cluster is provisioned from Rancher
	- This deploys SUC chart to the downstream cluster which in turn creates PSP and related resources from `psp.yaml` file.
	- Helm creates a secret of type "Helm Release" to store objects that got created
2. The downstream cluster is upgraded to k8s 1.25 (on Rancher release 2.7.2)
	- k8s 1.25+ no longer serves PSP API; however that "Helm Release" secret still has a record of PSP resource created
3. If SUC needs to be upgraded/reinstalled, it will fail because `helm uninstall` / `helm upgrade` would fail due to having record of PSP created on install (see https://github.com/helm/helm/issues/11287). SUC upgrade/reinstall will be triggered by Rancher upgrade to version _after_ 2.7.2, but there are probably other cases where it will be triggered.

## Solution Outline
Block (defer) upgrade to 1.25+ until Rancher deploys a version of the SUC chart that has the PSPs disabled, and ensure the chart has actually rolled out before proceeding with the actual upgrade to 1.25+

### Additional Solution Details
Courtesy of @Oats87 
#### v2prov Planner Modifications
- `managesystemagentplan.go` needs to set a condition after it checks the status of the bundle, to see if the bundle is ready: https://fleet.rancher.io/cluster-bundles-state (bonus points if the condition contains a hash or something of the bundle so that we don’t have the possibility of inconsistency if we reuse the condition in the future). It also needs to check for the version of k8s on the cluster object and if it’s >= 1.25, make sure to disable PSP via the chart option detailed below.
[...]

PR #40445, opened 2023-02-07
[2.7] Pass specific values to SUC chart to enable/disable PSPs for imported RKE2/K3s clusters

HarrisonWAffel:
## Problem
When importing RKE2/K3s clusters into Rancher, we allow users to upgrade the Kubernetes version in the UI. Doing so deploys the System Upgrade Controller, which executes a plan that upgrades the RKE2/K3s service. The System Upgrade Controller also deploys `PodSecurityPolicy` resources. This presents a problem when we are upgrading to a version of Kubernetes which no longer supports PSPs. If we continue to use the version of SUC which deployed PSPs after the upgrade to 1.25, or any version that does not support PSPs, we would not be able to delete or modify the controller as the helm operations which deploys it would fail due to an inability to clean-up / modify the PSPs still tracked by the chart. 
 
## Solution
This problem and its solution are somewhat tied to #40384, which defines a similar problem for Rancher provisioned clusters. In either case, we will use an updated version of the SUC chart which allows us to define if we would like to deploy PSP resources or not via the `values.yaml` content.

This PR modifies the updating logic for the SUC which is executed before an upgrade to 1.25 and configures the `values.yaml` content to denote that we want to disable PSPs. This upgrade handler performs the SUC upgrade/redeployment _before_ the clusters Kubernetes version is updated, which results in the PSPs not being tracked by Helm once the upgrade is complete and the ability to further modify the SUC deployment.
 
## Testing
+ Import an RKE2 v1.23.16 cluster into Rancher
    + At this point the SUC is not expected to be deployed
    + We need to use a version > 1.24.x as the SUC and its PSP resources are not deployed until an upgrade is requested. 
+ Upgrade the imported cluster to v1.24.x
    + Observe the SUC is deployed 
    + Observe that related PSP resources are deployed 
        + you can use the following command to easily follow the deployment of PSP resources `watch -n 1 kubectl get psp -A`
+ Upgrade the imported cluster to v1.25.x
    + Observe the SUC is redeployed and that PSP resources are removed at the same time
       + you can use the following command to easily follow the deployment of PSP resources `watch -n 1 kubectl get psp -A`
       + You will know that the cluster is undergoing a k8s upgrade because this command will stop working, reporting a network error as the API server is brought down for an upgrade. If the PSP is gone before this error is shown, then the removal has been successful.
    + Observe that the Kubernetes upgrade happens successfully and the cluster is usable 

### Manual Testing
I've done the above

### Automated Testing
N/A

## QA Testing Considerations
We should also test rolling back the Kubernetes version to ensure that PSP resources are re-added to the cluster once we are on a version that still supports them.
 
There shouldn't be any regressions when updating the cluster version to a Kubernetes version sub 1.25.

### DIFF
--- a/pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
+++ b/pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
@@ -4,12 +4,10 @@ import (
 	"fmt"
 	"strings"
 
-	app2 "github.com/rancher/rancher/pkg/app"
-
+	"github.com/coreos/go-semver/semver"
 	v32 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	v33 "github.com/rancher/rancher/pkg/apis/project.cattle.io/v3"
-
-	"github.com/coreos/go-semver/semver"
+	app2 "github.com/rancher/rancher/pkg/app"
 	v3 "github.com/rancher/rancher/pkg/generated/norman/management.cattle.io/v3"
 	"github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/project"
@@ -19,6 +17,14 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	// PSPAnswersField is passed to the helm --set command and denotes if we want to enable PodSecurityPolicies
+	// when deploying the app, overriding the default value of 'true'.
+	// In clusters >= 1.25 PSP's are not available, however we should
+	// continue to deploy them in sub 1.25 clusters as they are required for cluster hardening.
+	PSPAnswersField = "global.cattle.psp.enabled"
 )
 
 func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster, error) {
@@ -53,6 +59,8 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 		return cluster, nil
 	}
 
+	// Check if the cluster is undergoing a Kubernetes version upgrade, and that
+	// all downstream nodes also need the upgrade
 	isNewer, err := IsNewerVersion(cluster.Status.Version.GitVersion, updateVersion)
 	if err != nil {
 		return cluster, err
@@ -73,6 +81,7 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 		}
 
 	}
+
 	// set cluster upgrading status
 	cluster, err = h.modifyClusterCondition(cluster, planv1.Plan{}, planv1.Plan{}, strategy)
 	if err != nil {
@@ -80,7 +89,7 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 	}
 
 	// create or update k3supgradecontroller if necessary
-	if err = h.deployK3sBasedUpgradeController(cluster.Name, isK3s, isRke2); err != nil {
+	if err = h.deployK3sBasedUpgradeController(cluster.Name, updateVersion, isK3s, isRke2); err != nil {
 		return cluster, err
 	}
 
@@ -94,7 +103,7 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
 
 // deployK3sBaseUpgradeController creates a rancher k3s/rke2 upgrader controller if one does not exist.
 // Updates k3s upgrader controller if one exists and is not the newest available version.
-func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isRke2 bool) error {
+func (h *handler) deployK3sBasedUpgradeController(clusterName, updateVersion string, isK3s, isRke2 bool) error {
 	userCtx, err := h.manager.UserContextNoControllers(clusterName)
 	if err != nil {
 		return err
@@ -130,6 +139,18 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
 		return err
 	}
 
+	// determine what version of Kubernetes we are updating to
+	is125OrAbove, err := Is125OrAbove(updateVersion)
+	if err != nil {
+		return err
+	}
+
+	// if we are using a version above or equal to 1.25 we need to explicitly disable PSPs
+	enablePSPInChart := "false"
+	if !is125OrAbove {
+		enablePSPInChart = "true"
+	}
+
 	appLister := userCtx.Management.Project.Apps("").Controller().Lister()
 	appClient := userCtx.Management.Project.Apps("")
 
@@ -158,9 +179,12 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
 				Description:     "Upgrade controller for k3s based clusters",
 				ExternalID:      latestVersionID,
 				ProjectName:     appProjectName,
+				Answers:         make(map[string]string),
 				TargetNamespace: systemUpgradeNS,
 			},
 		}
+
+		desiredApp.Spec.Answers[PSPAnswersField] = enablePSPInChart
 
 		// k3s upgrader doesn't exist yet, so it will need to be created
 		if _, err = appClient.Create(desiredApp); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -21,6 +19,14 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+const (
+	// ConditionalPSPValues defines the values.yaml content passed to the app. Specifically,
+	// it denotes if PodSecurityPolicies should be deployed with the version of SUC being
+	// deployed / updated. In clusters >= 1.25 PSP's are not available, however we should
+	// continue to deploy them in sub 1.25 clusters as they are required for cluster hardening.
+	ConditionalPSPValues = "global:\n  cattle:\n    systemDefaultRegistry: \"\"\n    psp:\n      enabled: %s\n\nsystemUpgradeController:\n  image:\n    repository: rancher/system-upgrade-controller\n    tag: v0.10.0\nkubectl:\n  image:\n    repository: rancher/kubectl\n    tag: v1.23.3"
```
jakefhyde, line 27:
```suggestion
	ConditionalPSPValues = `global:
	cattle:
	  systemDefaultRegistry: ""
	  psp:
	    enabled: %t
	    systemUpgradeController:
	      image:
	        repository: rancher/system-upgrade-controller
	        tag: v0.10.0
	        kubectl:
	          image:
	          repository: rancher/kubectl
	          tag: v1.23.3`
```

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -130,6 +140,19 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
...
+	if err != nil {
+		return err
+	}
+
+	// if we're creating the app for the first time, and we are using a version
+	// above or equal to 1.25 we need to explicitly disable PSPs
+	pspEnabledValue := "true"
+	if is125OrAbove {
+		pspEnabledValue = "false"
+	}
```
jakefhyde, line 154:
```suggestion
	pspEnabledValue := true
	if is125OrAbove {
		pspEnabledValue = false
	}
```
or potentially:
```
	pspEnabledValue = !is125OrAbove
```

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -177,13 +202,18 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
 			}
 		}
 
-		if app.Spec.ExternalID == latestVersionID {
-			// app is up to date, no action needed
+		desiredApp := app.DeepCopy()
+		externalIDIsCorrect := app.Spec.ExternalID == latestVersionID
+		pspValuesHaveBeenSet := desiredApp.Spec.ValuesYaml == fmt.Sprintf(ConditionalPSPValues, pspEnabledValue)
```
jakefhyde, line 207: We should move `desiredApp := app.DeepCopy()` to directly before `desiredApp` get modified, since it looks like we perform the copy but may not potentially need it.

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -73,14 +81,16 @@ func (h *handler) onClusterChange(key string, cluster *v3.Cluster) (*v3.Cluster,
...
+
 	// set cluster upgrading status
 	cluster, err = h.modifyClusterCondition(cluster, planv1.Plan{}, planv1.Plan{}, strategy)
 	if err != nil {
 		return cluster, err
 	}
 
 	// create or update k3supgradecontroller if necessary
-	if err = h.deployK3sBasedUpgradeController(cluster.Name, isK3s, isRke2); err != nil {
+	err = h.deployK3sBasedUpgradeController(cluster.Name, updateVersion, isK3s, isRke2)
```
jakefhyde, line 92: I would still probably keep this in the format `if err = ...; err != nil {`

jakefhyde (changes requested):

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
HarrisonWAffel: Good catch, updated

jakefhyde (changes requested):
1 nit

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -177,13 +212,18 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
 			}
 		}
 
-		if app.Spec.ExternalID == latestVersionID {
-			// app is up to date, no action needed
+		externalIDIsCorrect := app.Spec.ExternalID == latestVersionID
+		pspValuesHaveBeenSet := app.Spec.ValuesYaml == fmt.Sprintf(ConditionalPSPValues, pspEnabledValue)
```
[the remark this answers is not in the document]
HarrisonWAffel: Good point, also removes the need for `pspEnabledValue` entirely. Fixed

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -160,13 +177,29 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
...
+		} else {
+			desiredApp.Spec.ValuesYaml = fmt.Sprintf(ConditionalPSPValues, "true")
 		}
 
 		// k3s upgrader doesn't exist yet, so it will need to be created
 		if _, err = appClient.Create(desiredApp); err != nil {
 			return err
 		}
 	} else {
+		// we should enforce the use of helm 3 from now on
```
annablender: Setting Helm 3 on the app status is fine, I just see conflicting messages between this and line 181. Do we want to use helm 3 if possible or always regardless of k8s version?

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -177,13 +210,33 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName string, isK3s, isR
 			}
 		}
 
-		if app.Spec.ExternalID == latestVersionID {
-			// app is up to date, no action needed
+		desiredApp := app.DeepCopy()
```
annablender: I'd move this DeepCopy to line 229 because you aren't setting sub values in app until then. You can reference values using `app.` before that

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -141,12 +153,17 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName, updateVersion str
...
-	// determine what version of Kubernetes we are updating for
+	// determine what version of Kubernetes we are updating to
 	is125OrAbove, err := Is125OrAbove(updateVersion)
 	if err != nil {
 		return err
 	}
 
+	// if we're creating the app for the first time, and we are using a version
+	// above or equal to 1.25 we need to explicitly disable PSPs
+
```
annablender, line 164: Nit: remove extra line

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -210,33 +213,18 @@ func (h *handler) deployK3sBasedUpgradeController(clusterName, updateVersion str
...
-		// so if we are sub 1.25 we don't need to define a value at all.
-		if !pspValuesHaveBeenSet {
-			if is125OrAbove {
-				desiredApp.Spec.ValuesYaml = fmt.Sprintf(ConditionalPSPValues, "false")
-			} else {
-				desiredApp.Spec.ValuesYaml = fmt.Sprintf(ConditionalPSPValues, "true")
-			}
-		}
-
+		desiredApp := app.DeepCopy()
```
annablender, line 224: Ok these changes in general look better, I was having the feeling the previous commit got a little wordy haha. Looks pretty good

annablender (changes requested):

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
HarrisonWAffel: Oh sorry, this should be a single commit. I'll fix that

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
HarrisonWAffel: done

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
```diff
@@ -21,6 +19,28 @@ import (
...
+	// it denotes if PodSecurityPolicies should be deployed with the version of SUC being
+	// deployed / updated. In clusters >= 1.25 PSP's are not available, however we should
+	// continue to deploy them in sub 1.25 clusters as they are required for cluster hardening.
+	ConditionalPSPValues = `global:
+  cattle:
+    systemDefaultRegistry: ""
+    psp:
+      enabled: %t
+
+systemUpgradeController:
```
Oats87, line 33: Do we actually need to define this bit of the values file? I believe helm should use the chart values by default

Oats87 (changes requested):

pkg/controllers/management/k3sbasedupgrade/k3sbased_upgrade_handler.go
HarrisonWAffel: I believe we have to specify them all, not because of how helm works but because of how Rancher handles that field when templating the app

https://github.com/rancher/rancher/blob/ea94f0ad4fa168a597da1bc6df74a24faf28507c/pkg/controllers/managementuserlegacy/helm/common.go#L148-L171

https://github.com/rancher/rancher/blob/ea94f0ad4fa168a597da1bc6df74a24faf28507c/pkg/controllers/managementuserlegacy/helm/common/common.go#L289-L297

I'll double check this though, not including them all would be ideal

HarrisonWAffel: I realized that the `app.spec.Answers` field eventually gets used when crafting the `--set` flag, allowing us to just specify that field instead of the entire file. PR has been tested and updated

jakefhyde (approved):

jakefhyde (approved):

Oats87 (approved):

a-blender (approved):

Issue #40326, opened 2023-01-26
Update Webhook chart to support updated approach for PSPs


MbolotSuse:
<s>The chart for rancher/webhook will need to be updated to handle the new approach designed for PSPs. 

The chart will need to be modified include the following option:

```yaml
global:
  cattle:
     psp:
        enabled: false #PSPs enablement should default to true
```

A new file will need to be added called `validate-psp-install.yaml` in the `charts/rancher-webhook/templates` directory with the following content:

```yaml
#{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
#{{- if .Values.global.cattle.psp.enabled }}
#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
#{{- fail "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." -}}
#{{- end }}
#{{- end }}
#{{- end -}}
```

Any files that render PSPs will need to make that rendering conditional solely on the `global.cattle.psp.enabled` value, like below:

```yaml
{{- if .Values.global.cattle.psp.enabled  -}}
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
...
{{- end }}
```
This approach meets the following requirements:
- PSPs should be used by the chart by default, keeping compatibility with the current 2.7.0 approach. 
- If users attempt to request PSPs  when installing the chart in a cluster which does not support PSPs (for example, a cluster running k8s version 1.25), an error will be presented so that the user knows to change the `global.cattle.psp.enabled` value to `false` before continuing.  

A solution for this should probably start by reverting rancher/webhook#171, and then continue by applying the above approach.</s>

After further consideration of how this approach will work with webhooks which are installed in remote clusters, we've adjusted to a new approach outlined in this comment: https://github.com/rancher/rancher/issues/40326#issuecomment-1423238619. 

This approach has the following components:

- The `MutatingWebhookConfiguration` `rancher.cattle.io` will now be declared by the webhook's chart
- All references to the pre-delete job in the webhook chart (which was used solely to delete the above `MutatingWebhookConfiguration`), will be removed, including:
  - The pre-delete configuration values in `values.yaml`
  - The psp configuration values in `values.yaml` (psps were only used for the pre-delete job)
  - The  `validate-psp-install.yaml`
  - Pre-delete cluster role
  - Pre-delete cluster role binding
  - Pre-delete job
[...]

PR #40480, opened 2023-02-09
Forcing adoption of resources for rancher-webhook

MbolotSuse:
Related to #40326. Since the mutating webhook will now be included in the chart definition of the webhook (see rancher/webhook#189), rancher will need to "adopt" the pre-existing mutating webhook configurations that didn't belong to the webhook's chart definition.

Since we only want this to work for the webhook, we only set the force adopt flag to true when the webhook is the chart that we are installing.

### DIFF
--- a/pkg/controllers/dashboard/systemcharts/controller.go
+++ b/pkg/controllers/dashboard/systemcharts/controller.go
@@ -19,7 +19,8 @@ import (
 )
 
 const (
-	repoName = "rancher-charts"
+	repoName         = "rancher-charts"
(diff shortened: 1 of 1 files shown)


MbolotSuse:
This was tested using a local version of the webhook which did not declare the ValidatingAdmissionWebhook as part of it's chart. Rancher was unable to upgrade a release created from this chart without this code fix, but it was able to properly upgrade the chart with this fix.

KevinJoiner (approved):

crobby (approved):
lgtm

PR #40496, opened 2023-02-10
Revert "Add pspEnablement value to rancher ConfigMap and pass it to webhook"

MbolotSuse:
Reverts #40367. The webhook is now being changed to no longer require PSPs (it will still work on hardened clusters). In this change, we removed the option for PSPs from the chart. Therefore, we no longer need to pass down the PSP enablement flag to the webhook.


MbolotSuse:
Related to #40326.

maxsokolovsky (approved):

crobby (approved):

PR #40500, opened 2023-02-10
[2.7] Delay Planner Execution Based Off Of System-Upgrade-Controller Status

HarrisonWAffel:
## Problem
Kubernetes 1.25 has removed support for `PodSecurityPolicy`'s, however several Rancher components deploy them by default via helm. Helm tracks all of the resources it deploys so that it can fully manage them during upgrades or uninstall operations. However, helm does not handle the scenario wherein one of the managed resources no longer has API support for a given Kubernetes version. In such a case, Helm will error out and the deployed chart will become unmanageable, preventing uninstalls/upgrades/any changes at all. This presents a problem for several aspects of Rancher, however this PR simply focuses on the management of the System-Upgrade-Controller. 

## Solution
As described in the linked issue above, this solution has several steps 

1. The system-upgrade-controller chart has been modified to conditionally deploy PSP resources based off of the provided `values.yaml` field `global.cattle.psp.enabled`. 
2. A new condition has been placed on the RKE control plane object which indicates if the currently deployed System-Upgrade-Controller is in a 'ready' state with its PSP's resources enabled or disabled, depending on the Kubernetes version being used 
3. The planner has been modified to pause all plan processing until the System-Upgrade-Controller condition is in a 'ready' state, and has the proper resource configuration deployed for a given Kubernetes version.

## Testing
The following (lengthy) test steps cover both of the linked issues

#### Initial setup / Validation of 40565
<details closed> 

+ Create a single node all roles RKE2 cluster running kubernetes v1.24.x. This can be done on any provider
      + Ensure the name of the cluster is moderately long, e.g. `haffel-test-upgrade-123`. _This is required to validate 40565_
+ Once the cluster has become available take a snapshot 
+ Navigate to the cluster explorer UI and view all pods under all namespaces. Ensure that you can see the `system-upgrade-controller` pod running.
+ Run the following command on the local cluster, `kubectl get bundles.fleet.cattle.io -A` and locate the `system-upgrade-controller` bundle
    + ensure that the bundle is active, and that the name is less than 53 characters long
+ While still in the cluster explorer UI, navigate to `More Resources` and then to `Policy`. Select `PodSecurityPolicies` and ensure toy can see the `system-upgrade-controller` policy. 
+ At this point we can confirm that 40565 has been **resolved**, as we can see the bundle and its resources have been deployed despite the moderately lengthy name

</details>

#### Validation of 40384

<details closed>

+ Using the same cluster, SSH into the all roles node, and execute the following commands so that you may use `kubectl`
    + `export KUBECONFIG=/etc/rancher/rke2/rke2.yaml && alias kubectl=/var/lib/rancher/rke2/bin/kubectl`
+ The goal is to watch the PSP resources active in the cluster so we can ensure they are removed before an upgrade commences. To continuously watch the PSP resources, run the following command ` watch -n 1 /var/lib/rancher/rke2/bin/kubectl get psp -A`
+ Observe that a PodSecurityPolicy exists for the System-Upgrade-Controller 
+ While continuing to observe the active PSP resources, return to the Rancher UI and commence an upgrade to 1.25.x
+ Ensure that the PodSecurityPolicy for the System-Upgrade-Controller is removed _before_ the API server is brought down for the upgrade 
     + You will know that the API server is undergoing an upgrade as the `watch` command executed earlier will report an error communicating with the API server
     + It is important to ensure that the PSP resource is no longer present **_before_** that error is observed 
+ Ensure the cluster is still usable after the upgrade 
+ View the contents of the system-upgrade-controller bundle by running the following command
    + `kubectl get bundles.fleet.cattle.io <YOUR_SUC_BUNDLE_NAME> -n fleet-default -o yaml`
    + Ensure that a PodSecurityPolicy is **_not_** included in the `resourceKey` section. 

</details> 

#### ETCD snapshot and restore test 
<details closed>

+ Restore the 1.25 cluster using the snapshot you took initially on 1.24.x
+ ensure that the system-upgrade-controller pod is present after the cluster has finished restoring
+ ensure that you can once again see the PodSecurityPolicy resources for the System-Upgrade-Controller

</details>

### Manual Testing

I've done the above 

### Automated Testing
N/A

We should ensure that upgrading and downgrading between 1.25 and 1.24 works as expected
We should ensure that manual editing of the bundle works as expected

### DIFF
--- a/pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
+++ b/pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
@@ -1,13 +1,20 @@
 package managesystemagent
 
 import (
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
 	"github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	rancherv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+	"github.com/rancher/rancher/pkg/controllers/provisioningv2/rke2"
+	"github.com/rancher/rancher/pkg/fleet"
 	namespaces "github.com/rancher/rancher/pkg/namespace"
 	"github.com/rancher/rancher/pkg/provisioningv2/image"
 	"github.com/rancher/rancher/pkg/settings"
 	"github.com/rancher/wrangler/pkg/name"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
@@ -17,10 +24,27 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
 		return nil, status, nil
 	}
 
+	currentVersion, err := semver.NewVersion(cluster.Spec.KubernetesVersion)
+	if err != nil {
+		return nil, status, err
+	}
+
+	// indicate to the SUC chart if we want to
+	// install PodSecurityPolicy manifests
+	pspEnabled := false
+	if currentVersion.LessThan(Kubernetes125) {
+		pspEnabled = true
+	}
+
+	// we must limit the output of name.SafeConcatName to at most 48 characters because
+	// a) the chart release name cannot exceed 53 characters, and
+	// b) upon creation of this resource the prefix 'mcc-' will be added to the release name, hence the limiting to 48 characters
+	managedChartName := name.Limit(name.SafeConcatName(cluster.Name, "managed", "system-upgrade-controller"), 48)
+
 	mcc := &v3.ManagedChart{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
-			Name:      name.SafeConcatName(cluster.Name, "managed", "system-upgrade-controller"),
+			Name:      managedChartName,
 		},
 		Spec: v3.ManagedChartSpec{
 			DefaultNamespace: namespaces.System,
@@ -32,6 +56,9 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
 					"global": map[string]interface{}{
 						"cattle": map[string]interface{}{
 							"systemDefaultRegistry": image.GetPrivateRepoURLFromCluster(cluster),
+							"psp": map[string]interface{}{
+								"enabled": pspEnabled,
+							},
 						},
 					},
 				},
@@ -55,4 +82,82 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
 	return []runtime.Object{
 		mcc,
 	}, status, nil
+}
+
+// syncSystemUpgradeControllerStatus queries the managed system-upgrade-controller chart and determines if it is properly configured for a given
+// version of Kubernetes. It applies a condition onto the control-plane object to be used by the planner when handling Kubernetes upgrades.
+func (h *handler) syncSystemUpgradeControllerStatus(obj *rkev1.RKEControlPlane, status rkev1.RKEControlPlaneStatus) (rkev1.RKEControlPlaneStatus, error) {
+	// perform the same name limiting as in the OnChangeInstallSUC controller, but prepend the 'mcc-' prefix that is added when the bundle is created
+	bundleName := fmt.Sprintf("mcc-%s", name.Limit(name.SafeConcatName(obj.Name, "managed", "system-upgrade-controller"), 48))
+	sucBundle, err := h.bundles.Get(fleet.ClustersDefaultNamespace, bundleName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		// if we couldn't find the bundle then we know it's not ready
+		rke2.SystemUpgradeControllerReady.False(&status)
+		// don't return the error, otherwise the status won't be set to 'false'
+		return status, nil
+	}
+
+	if err != nil {
+		return status, err
+	}
+
+	rke2.SystemUpgradeControllerReady.Unknown(&status)
+
+	// determine if the SUC deployment has been rolled out fully, and if there were any errors encountered
+	if sucBundle.Status.Summary.Ready != sucBundle.Status.Summary.DesiredReady {
+		if sucBundle.Status.Summary.ErrApplied != 0 && len(sucBundle.Status.Summary.NonReadyResources) > 0 {
+			nonReady := sucBundle.Status.Summary.NonReadyResources
+			rke2.SystemUpgradeControllerReady.Reason(&status, fmt.Sprintf("Error Encountered Waiting for System Upgrade Controller Deployment To Roll Out: %s", nonReady[0].Message))
+			return status, nil
+		}
+		rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for System Upgrade Controller Deployment roll out")
+		return status, nil
+	}
+
+	// we need to look at the values yaml content to determine if PSPs are enabled
+	// or disabled. We need to wait until SUC is redeployed if we don't see any helm values, as
+	// we expect the PSP value to be explicitly defined as either true or false
+	valuesYamlAvailable := sucBundle.Spec.Helm != nil && sucBundle.Spec.Helm.Values != nil
+	if !valuesYamlAvailable {
+		rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for Upgraded System Upgrade Controller Deployment")
+		return status, nil
+	}
+
+	// look through the values yaml content to determine if 'psp: enabled: true'
+	rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for System Upgrade Controller Bundle Update")
+	data := sucBundle.Spec.Helm.Values.Data
+	global, ok := data["global"].(map[string]interface{})
+	if !ok {
+		return status, nil
+	}
+
+	cattle, ok := global["cattle"].(map[string]interface{})
+	if !ok {
+		return status, nil
+	}
+
+	psp, ok := cattle["psp"].(map[string]interface{})
+	if !ok {
+		return status, nil
+	}
+
+	currentVersion, err := semver.NewVersion(obj.Spec.KubernetesVersion)
+	if err != nil {
+		return status, err
+	}
+
+	// we only want to block an upgrade if PSPs are enabled AND we are on
+	// a version greater than or equal to 1.25.
+	enabled, ok := psp["enabled"].(bool)
+	if !ok {
+		return status, nil
+	}
+
+	if !currentVersion.LessThan(Kubernetes125) && enabled {
+		rke2.SystemUpgradeControllerReady.Reason(&status, "System Upgrade Controller Not Ready")
+		return status, nil
+	}
+
+	rke2.SystemUpgradeControllerReady.True(&status)
+	return status, nil
 }
--- a/pkg/provisioningv2/rke2/planner/planner.go
+++ b/pkg/provisioningv2/rke2/planner/planner.go
@@ -256,6 +256,17 @@ func (p *Planner) Process(cp *rkev1.RKEControlPlane, status rkev1.RKEControlPlan
 		return status, err
 	}
 
+	// we need to make sure the cluster has undergone initial provisioning and bootstrapping before we can enforce this condition,
+	// otherwise the system-upgrade-controller bundle will never become ready and the planner will be indefinitely blocked.
+	// We do this by ensuring the cluster has reconciled, and has a valid control plane join URL. However, it should be noted
+	// that a small window of time does exist when the joinURL has not been set, but the bootstrap node is up.
+	if rke2.Reconciled.IsTrue(cp) && p.store.ClusterHasBeenBootstrapped(plan) && rke2.SystemUpgradeControllerReady.IsFalse(&status) {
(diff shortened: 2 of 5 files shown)


annablender:
Since this PR isn't finished, may be best to convert it to a draft

HarrisonWAffel:
@annablender I need feedback from others before this PR can be finished, hence why it is no longer a draft

pkg/provisioningv2/rke2/planner/planner.go
```diff
@@ -222,6 +222,15 @@ func (p *Planner) Process(cp *rkev1.RKEControlPlane, status rkev1.RKEControlPlan
 		return status, ErrWaitingf("%s/%s: releaseData nil for version %s", cp.Namespace, cp.Name, cp.Spec.KubernetesVersion)
 	}
 
+	// we can't enforce this condition during initial provisioning, all nodes need to come up
```
jakefhyde, line 225: Can you relocate this after the pause check for the CAPI cluster i.e. after the 
```
	if !capiCluster.DeletionTimestamp.IsZero() {
//...
	}
```
block?

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +78,77 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+	sucBundle, sucBundleRetrievalError := h.bundles.Get(fleet.ClustersDefaultNamespace, name.SafeConcatName("mcc", obj.Name, "managed", "system-upgrade-controller"), metav1.GetOptions{})
+	if sucBundleRetrievalError != nil && !errors.IsNotFound(sucBundleRetrievalError) {
+		return status, sucBundleRetrievalError
+	}
+
+	rke2.SystemUpgradeControllerReady.False(&status)
+	// if we couldn't find the bundle then we know it's not ready
+	if sucBundleRetrievalError != nil {
+		return status, nil
+	}
```
jakefhyde, line 94: Generally I would invert these conditions, like so
```suggestion
	sucBundle, err := h.bundles.Get(fleet.ClustersDefaultNamespace, name.SafeConcatName("mcc", obj.Name, "managed", "system-upgrade-controller"), metav1.GetOptions{})
	if errors.IsNotFound(err) {
		// if we couldn't find the bundle then we know it's not ready
		rke2.SystemUpgradeControllerReady.False(&status)
		return status, nil
	}

	if err != nil {
		return status, err
	}
```
Also, we typically set conditions to Unknown while they are in progress, which is sort of opposite from what the current upstream specification is, but it is still established convention.

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +78,77 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+
+	// determine if the SUC deployment has been rolled out fully, and if there were any errors encountered
+	if sucBundle.Status.Summary.Ready != sucBundle.Status.Summary.DesiredReady {
+		if sucBundle.Status.Summary.ErrApplied != 0 && len(sucBundle.Status.Summary.NonReadyResources) > 0 {
+			nonReady := sucBundle.Status.Summary.NonReadyResources
+			rke2.SystemUpgradeControllerReady.Reason(&status, fmt.Sprintf("Error Encountered Waiting for System Upgrade Controller Deployment To Roll Out: %s", nonReady[0].Message))
```
jakefhyde, line 100: Does this indicate retriable errors, we probably want it to differentiate errors vs. in progress.

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +78,77 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+			return status, nil
+		}
+		rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for System Upgrade Controller Deployment To Roll Out...")
```
jakefhyde, line 103:
```suggestion
		rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for System Upgrade Controller Deployment roll out")
```

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +78,77 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+		return status, nil
+	}
+
+	// we need to look at the values yaml content to determine if PSPs are enabled
+	// or disabled. We need to wait until SUC is redeployed if we don't see any helm values, as
+	// we expect the PSP value to be explicitly defined as either true or false
+	valuesYamlAvailable := sucBundle.Spec.Helm != nil && sucBundle.Spec.Helm.Values != nil
+	if !valuesYamlAvailable {
+		rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for Upgraded System Upgrade Controller Deployment...")
```
jakefhyde, line 112:
```suggestion
		rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for Upgraded System Upgrade Controller Deployment")
```

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +78,77 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+		return status, nil
+	}
+
+	// look through the values yaml content to determine if 'psp: enabled: true'
+	rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for System Upgrade Controller Bundle To Be Updated")
```
jakefhyde, line 117:
```suggestion
	rke2.SystemUpgradeControllerReady.Reason(&status, "Waiting for System Upgrade Controller Bundle update")
```

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +78,77 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+
+	// we only want to block an upgrade if PSPs are enabled AND we are on
+	// a version greater than or equal to 1.25.
+	enabled, ok := psp["enabled"].(bool)
+	if !ok {
+		return status, nil
+	}
+
+	if !currentVersion.LessThan(Kubernetes125) && enabled {
+		rke2.SystemUpgradeControllerReady.Reason(&status, "System Upgrade Controller Not Yet Ready")
```
jakefhyde, line 147:
```suggestion
		rke2.SystemUpgradeControllerReady.Reason(&status, "System Upgrade Controller not ready")
```

HarrisonWAffel:
The main question I had with this PR is the blocking of upgrades based off of previous delivery of a plan. I know we deliver an initial plan upon provisioning, so did we mean plans that are currently being being executed? I was a bit confused how we would be able to upgrade at all if we block on if we've ever delivered a plan - likely misunderstanding something about the outline though.

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
HarrisonWAffel: Thats a good point, i'll look more into that today

pkg/provisioningv2/rke2/planner/planner.go
HarrisonWAffel: Okay, its been moved. I wanted to check though, is checking the `Reconciled` condition enough to ensure that we've undergone initial provisioning? I would think so, but I wouldn't be surprised if I've missing something regarding how this status is set during initial provisioning.

jakefhyde:
> The main question I had with this PR is the blocking of upgrades based off of previous delivery of a plan. I know we deliver an initial plan upon provisioning, so did we mean plans that are currently being executed? Or potentially previously delivered plans which contained PSP's? I was a bit confused how we would be able to upgrade at all if we just block on if we've ever delivered a plan - likely misunderstanding something about the outline though.

@Oats87 Correct me if I'm wrong, but it's if any plan that has been delivered to a downstream cluster. We don't have anything built into the framework to know that the plan has been observed and is currently in progress (nor do I think we reasonably could), so delivering a plan containing any PSPs should be the requirement. This is because we need to wait for SUC to upgrade without the PSPs, otherwise the plans will never successfully apply.

HarrisonWAffel:
Okay, I _think_ I've cleared up some of my initial confusion on this requirement after rereading the parent issue more closely and reading Jakes comment. I believe my misunderstanding was how the condition was going to be setup, correct me if I'm wrong but I believe what we are trying to accomplish is more along the lines of 

```
if Reconciled && SUCReady && AnyPlanHasBeenDelivered {
    ErrWait
}
```

If that is the case, what is the best way to determine if a plan has been delivered?  Looking around at the plan struct I would think something like 

```
plan, err := p.Load(cluster, rkeControlPlane)
if err != nil {
	return false, err
}

for _, e := range plan.Nodes {
	if e.AppliedPlan != nil {
		return true, nil
	}
}
```

would suffice, right?

pkg/provisioningv2/rke2/planner/store.go
```diff
@@ -58,6 +58,25 @@ func onlyRKE(machines []*capi.Machine) (result []*capi.Machine) {
...
+	if err != nil {
+		return false, err
+	}
+
+	if plan.Nodes == nil {
+		return false, nil
+	}
+
+	for _, e := range plan.Nodes {
+		if e.AppliedPlan != nil && e.InSync {
```
jakefhyde, line 72: I think in this case, we actually want to check to make sure that the initial plan isn't set. Even though the plan may not have completed, that doesn't necessarily mean that it won't complete in between the time we observe it and the time we act on it, or that there were other changes (i.e. plan may have applied but probes may not yet be complete). We may need to check the secret, convert the `Plan` field to a pointer, or add a function/field (or both) to indicate that a plan was delivered to the node via the secret..

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
HarrisonWAffel: I don't believe we can determine if the error here is retriable or permanent, in either case the same fields are used. Unless we know what types of errors can be retried and check against those I'm not sure how we would differentiate

pkg/provisioningv2/rke2/planner/store.go
HarrisonWAffel: I'm assuming that once the plan secret has been generated it will always be available in the store? In which case we could, like you said, just query the secrets generated for all of the machines and check if there are any. If there are not, can we safely assume the initial secret has not been created yet? 

I've updated the last commit with a different function for checking if plans have been delivered, I still getting to know the planner code base but hopefully its closer to what we're looking for

Oats87: Yeah, any plan that has been delivered should work; however, there is probably an edge case when dealing with a cluster that is being "etcd restored" from nothing, because the initial plan will not actually net a functional cluster. Maybe we need to be a little more intelligent about how we determine to block... need to think about this

HarrisonWAffel:
This PR has been updated to make use of the control-plane `joinURL` when determining if we should block plan executions. This ensures that we do not indefinitely block the planner in cases where need to provide an initial plan before the SUC can enter a ready state.

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -17,10 +24,27 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+	// install PodSecurityPolicy manifests
+	pspEnabled := false
+	if currentVersion.LessThan(Kubernetes125) {
+		pspEnabled = true
+	}
+
+	// we must limit the output of name.SafeConcatName to at most 48 characters because
+	// a) the chart release name cannot exceed 53 characters, and
+	// b) upon creation of this resource the prefix 'mcc-' will be added to the release name, hence the limiting to 48 characters
+	managedChartName := name.Limit(name.SafeConcatName(cluster.Name, "managed", "system-upgrade-controller"), 48)
```
Oats87, line 42: what happened before when we had a super long cluster name...? is this going to break upgrades for super long cluster names?

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
```diff
@@ -56,3 +83,81 @@ func (h *handler) OnChangeInstallSUC(cluster *rancherv1.Cluster, status rancherv
...
+		return status, err
+	}
+
+	rke2.SystemUpgradeControllerReady.Unknown(&status)
+
+	// determine if the SUC deployment has been rolled out fully, and if there were any errors encountered
+	if sucBundle.Status.Summary.Ready != sucBundle.Status.Summary.DesiredReady {
+		if sucBundle.Status.Summary.ErrApplied != 0 && len(sucBundle.Status.Summary.NonReadyResources) > 0 {
+			nonReady := sucBundle.Status.Summary.NonReadyResources
+			rke2.SystemUpgradeControllerReady.Reason(&status, fmt.Sprintf("Error Encountered Waiting for System Upgrade Controller Deployment To Roll Out: %s", nonReady[0].Message))
```
Oats87, line 110: nit: capitalization here is a little wonky, it can be written as a sentence. Not important though

Oats87 (approved):

pkg/controllers/provisioningv2/rke2/managesystemagent/managesystemagentplan.go
HarrisonWAffel: The bundle was never properly deployed, this shouldn't break any existing clusters and should actually result in the proper deployment of the bundle

jakefhyde (approved):

Issue #40494, opened 2023-02-10
[BUG] Rancher Terraform Provider: Unable to remove pod_security_policy from projects


samjustus:
SURE-5795

Terraform Provider for SUSE Rancher 1.22.2

Removing psp seems easy, by just disabling the psp in the cluster config.

pod_security_policy = false 
And then remove the [pod_security_policy_template_id](https://registry.terraform.io/providers/rancher/rancher2/latest/docs/resources/project#pod_security_policy_template_id) of each project. However this does not work and leads to errors when making changes to the projects (e.g. increasing the storage limits):

  # rancher2_project.pr["demo"] will be updated in-place
  ~ resource "rancher2_project" "pr" {
        id                              = "c-xxx:p-xxx"
        name                            = "Demo"
      - pod_security_policy_template_id = "restricted" -> null
        # (6 unchanged attributes hidden)

        # (3 unchanged blocks hidden)
    }
This does not work and terraform apply fails

│ Error: Bad response statusCode [500]. Status [500 Internal Server Error]. 
    Body: [baseType=error, code=ServerError, message=error parse/validate 
     action body: podSecurityPolicyTemplateId=MissingRequired 422: ]
     from [https://rancher.intra/v3/projects/c-xxxx:p-xxxx?action=setpodsecuritypolicytemplate]

│
│   with rancher2_project.pr["demo"],
│   on projects_namespaces.tf line 14, in resource "rancher2_project" "pr":
│   14: resource "rancher2_project" "pr" 
Neither removing the whole attribute pod_security_policy_template_id nor setting it to null or "" works. So how can I disable the psp and remove associated psp from the projects.

Only disabling the pod_security_policy and not removing the psps of the projects, will lead to errors when changing projects (e.g. increasing storage limits).

Error: Bad response statusCode [422]. Status [422 Unprocessable Entity]. 
Body: [baseType=error, code=InvalidAction, 
message=cluster [c-xxxx] does not have Pod Security Policies enabled] 
from [https://rancher.intra/v3/projects/c-xxxx:p-xxxx?action=setpodsecuritypolicytemplat 
While it looks like while on existing projects the pod_security_policy_template_id cannot be removed, I however can create new projects where pod_security_policy_template_id = null. Hence the conclusion is

if psp is disabled, I can create projects where the pod_security_policy_template_id is not set
I can't remove the psp association of existing projects, cause setting pod_security_policy_template_id to null, gives me the error above
[...]

PR #40518, opened 2023-02-13
Handle setpodsecuritypolicytemplate setting to null by norman client

crobby:
## Issue: 
https://github.com/rancher/terraform-provider-rancher2/issues/1006
#40494 

## Problem
The Norman client, used by the terraform provider, has omitempty set for podSecurityPolicyTemplateId.  https://github.com/rancher/rancher/blob/cd8eea1fdf0dd0071bfdcae4f1eb7de2c5b3722f/pkg/client/generated/management/v3/zz_generated_set_pod_security_policy_template_input.go#L9
And that results in an empty request body being sent when trying to unset the value via the terraform provider. 

## Solution
Since fixing the client generation logic would result in a very large number of changes, we have opted to just handle this situation in Rancher by removing the "required" attribute in the schema spec.
 
## Testing
Testing can be done through UI, curls to the api, and most importantly through the terraform provider, as described in the original issue, since that is the only scenario that is resulting in the problem (empty body being sent instead of null)

### Manual Testing
Tested via the UI, via curl, and the terraform provider

### DIFF
--- a/pkg/apis/management.cattle.io/v3/authz_types.go
+++ b/pkg/apis/management.cattle.io/v3/authz_types.go
@@ -193,5 +193,5 @@ func (c *ClusterRoleTemplateBinding) ObjClusterName() string {
 }
 
 type SetPodSecurityPolicyTemplateInput struct {
-	PodSecurityPolicyTemplateName string `json:"podSecurityPolicyTemplateId" norman:"required,type=reference[podSecurityPolicyTemplate]"`
+	PodSecurityPolicyTemplateName string `json:"podSecurityPolicyTemplateId" norman:"type=reference[podSecurityPolicyTemplate]"`
 }


pkg/api/norman/customization/project/project_actions.go
```diff
@@ -269,8 +269,21 @@ func (h *Handler) setPodSecurityPolicyTemplate(actionName string, action *types.
 	request *types.APIContext) error {
 	input, err := handler.ParseAndValidateActionBody(request, request.Schemas.Schema(&managementschema.Version,
 		client.SetPodSecurityPolicyTemplateInputType))
+	// In the empty body scenario, the content length is reported as 2.  We are using the content length
+	// to try and differentiate the ParseAndValidateActionBody failure scenarios.  If the body is just empty,
+	// we will treat it as an attempt to set it to nil, any other failure in parsing/validating will still
+	// result in an error.
+	const emptyBodyContentLength = 2
```
MbolotSuse, line 276: Is this value set to 2 because we expect that the empty body is `"{}"`? If not, would you mind explaining why this is 2?

MbolotSuse (changes requested):
I'm not quite sure that this is the correct place to make this change. Near as I can tell, the original issue is caused because the `PodSecurityPolicyTemplateName` is set to required by the core type here: https://github.com/rancher/rancher/blob/828fe615a30ed0302d39df1a78366ce4f4a8b381/pkg/apis/management.cattle.io/v3/authz_types.go#L196 . 

Norman is comparing the input to the schema and finds that a required value isn't being set here: https://github.com/rancher/norman/blob/34230bb2787c7a809722e041d0ea1c9247ef6c5a/parse/builder/builder.go#L132, which results in the request being rejected. What might be better is if you removed the required tag from the type declaration. You could then update this function to handle an empty name with less guesswork.

pkg/api/norman/customization/project/project_actions.go
crobby: Yeah.  I hate magic numbers like this, but I wasn't sure of another way to distinguish between parsing failing due to empty body vs parsing failure due to something else.

MbolotSuse (approved):

cmurphy (approved):

PR #40534, opened 2023-02-14
Encoded private registry auth config

jakefhyde:
## Problem

During merging of the v2.6.10 backports, a base64 encoding was dropped. 

## Solution

Add the base64 encoding of the private registry auth credentials back.

### Manual Testing

I am still setting up an environment to test this, but I am fairly confident this will work.

### Automated Testing

Updated unit tests to ensure that the base64 encoding is returned.


kinarashah (approved):
LGTM based on [v2.6.9 code](https://github.com/rancher/rancher/blob/v2.6.9/pkg/cluster/private_registry.go#L60)

snasovich (approved):

PR #40543, opened 2023-02-14
Delete webhook-related resources on cluster detachment

maxsokolovsky:
## Problem
Re-importing previously detached clusters on fails on v2.7.2. The problem is that on initial detachment, some cluster-wide webhook resources remain (such as its configurations), so on re-import, Rancher tries to talk to a non-existing webhook service because it detects a configuration. 
 
## Solution
Delete all cluster-wide resources related to rancher-webhook on cluster detachment. Specifically:
1. MutatingConfiguration "rancher.cattle.io"
2. ValidatingConfiguration "rancher.cattle.io"
3. Cluster Role Binding "rancher-webhook"

All the other resources are taken care of by the agent job that already deletes the `cattle-system` namespace.

There are two changes:
1. Change the agent code that performs actual deletion of resources.
1. Change the Rancher code that configures the agent's job permissions - add permissions to delete MutatingConfiguration and ValidatingConfiguration.
 
### Manual Testing
Built a custom Rancher image which configures the job. Also built a custom agent image that runs the job.

### DIFF
--- a/pkg/agent/clean/cluster.go
+++ b/pkg/agent/clean/cluster.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rancher/rancher/pkg/controllers/management/usercontrollers"
 	"github.com/rancher/rancher/pkg/controllers/managementagent/nslabels"
 	"github.com/rancher/rancher/pkg/controllers/managementuserlegacy/helm"
 	"github.com/rancher/rancher/pkg/monitoring"
@@ -118,6 +119,11 @@ func Cluster() error {
 		if err := removeNamespace(ns, client); err != nil {
 			errors = append(errors, err)
 		}
+	}
+
+	webhookErr := cleanupWebhookResources(client)
+	if len(webhookErr) > 0 {
+		errors = append(errors, webhookErr...)
 	}
 
 	nsErr := cleanupNamespaces(client)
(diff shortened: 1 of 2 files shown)


pkg/agent/clean/cluster.go
```diff
@@ -349,6 +354,29 @@ func cleanupRoles(client *kubernetes.Clientset) []error {
 	return errs
 }
 
+func cleanupWebhookResources(client *kubernetes.Clientset) []error {
+	logrus.Info("Starting cleanup of webhook-specific resources")
+	var errs []error
+
+	if !dryRun {
+		err := client.RbacV1().ClusterRoleBindings().Delete(context.TODO(), "rancher-webhook", metav1.DeleteOptions{})
```
MbolotSuse, line 362: Nit:

Generally not a fan of `context.TODO()` in production code. You can use `context.Background()` for the same effect. Marking this as a nit since:
- Other parts of this file use todo
- todo and background are technically the same.

MbolotSuse (commented):
Overall structure lgtm. Got a couple of nits on various parts, but no issues outside of that.

KevinJoiner (changes requested):
@maxsokolovsky would you say the below statement is accurate?

The optimal solution here would be to uninstall the helm chart. But our current design for cluster cleanup does not give us the option of running shell pods to perform helm operations. Since clean-up is formed by a separate execution of the agent binary, which does not run the core rancher code.

maxsokolovsky:
@KevinJoiner, I did have a solution where the Job in the agent uninstalls the release using a Helm client. The problem was that it needs a number of permissions for a few resources. And we would still have this coupling between chart resources in one place and permissions needed to delete them in another.
So I decided to just make this simpler and delete only those resources that are not deleted by other code in the Job. The Job deletes the entire `cattle-system` namespaces, and so the service account and other namespaces resources of the webhook are deleted.

pkg/agent/clean/cluster.go
KevinJoiner: I don't like TODOs either, but I have recently been more accepting of them based on the last part of the function documentation. 
> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

pkg/agent/clean/cluster.go
```diff
@@ -349,6 +354,29 @@ func cleanupRoles(client *kubernetes.Clientset) []error {
 	return errs
 }
 
+func cleanupWebhookResources(client *kubernetes.Clientset) []error {
```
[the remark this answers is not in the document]
KevinJoiner: This can be ignored since we only need to handle non-namespaced resources. Everything else gets removed when the cattle-system namespace is removed.

KevinJoiner (approved):

pkg/agent/clean/cluster.go
```diff
@@ -349,6 +355,29 @@ func cleanupRoles(client *kubernetes.Clientset) []error {
...
+	logrus.Info("Starting cleanup of webhook-specific resources")
+	var errs []error
+
+	if !dryRun {
```
MbolotSuse, line 362: This doesn't follow the pattern of the other function's use of dryRun. Notably, your implementation won't print out the resources from the webhook that it is planning on deleting, which kind of makes the dryRun a bit deceptive.

I'd refer to the below `deleteJob` method for what I mean - you can see that even in a dry run it logs out each job that it intends on removing.

MbolotSuse (changes requested):
One issue with the implementation of dryRun for these resources.

pkg/agent/clean/cluster.go
MbolotSuse: Newest changes technically achieves what I mean.

MbolotSuse (approved):

KevinJoiner (approved):

